PR #92288, opened 2020-06-19
Cleanup tempfiles

zhijianli88:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/cluster/images/conformance/go-runner/tar_test.go
+++ b/cluster/images/conformance/go-runner/tar_test.go
@@ -36,7 +36,7 @@ func TestTar(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	defer os.Remove(tmp)
+	defer os.RemoveAll(tmp)
 
 	if err := os.Mkdir(filepath.Join(tmp, "subdir"), os.FileMode(0755)); err != nil {
 		t.Fatal(err)
--- a/pkg/volume/util/hostutil/hostutil_linux_test.go
+++ b/pkg/volume/util/hostutil/hostutil_linux_test.go
@@ -283,6 +283,7 @@ func TestGetFileType(t *testing.T) {
 
 	for idx, tc := range testCase {
 		path, cleanUpPath, err := tc.setUp()
+		defer os.RemoveAll(cleanUpPath) // RemoveAll can deal with a empty path ""
 		if err != nil {
 			// Locally passed, but upstream CI is not friendly to create such device files
 			// Leave "Operation not permitted" out, which can be covered in an e2e test
@@ -290,9 +291,6 @@ func TestGetFileType(t *testing.T) {
 				continue
 			}
 			t.Fatalf("[%d-%s] unexpected error : %v", idx, tc.name, err)
-		}
-		if len(cleanUpPath) > 0 {
-			defer os.RemoveAll(cleanUpPath)
 		}
 
 		fileType, err := hu.GetFileType(path)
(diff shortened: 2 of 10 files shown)


pkg/volume/util/hostutil/hostutil_linux_test.go
```diff
@@ -283,6 +283,9 @@ func TestGetFileType(t *testing.T) {
 
 	for idx, tc := range testCase {
 		path, cleanUpPath, err := tc.setUp()
+		if len(cleanUpPath) > 0 {
```
SergeyKanzhelev, line 286: theoretically, there is no need for this check: 

> `func RemoveAll(path string) error`
> RemoveAll removes path and any children it contains. It removes everything it can but returns the first error it encounters. ***If the path does not exist, `RemoveAll` returns `nil` (no error).*** If there is an error, it will be of type *PathError.

cluster/images/conformance/go-runner/tar_test.go
```diff
@@ -36,7 +36,7 @@ func TestTar(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	defer os.Remove(tmp)
+	defer os.RemoveAll(tmp)
```
SergeyKanzhelev, line 39: general comment, not specific to this test case: I'm curious if the error output here may help to uncover some issues with test synchronization. For instance, if some routine keeps writing to the file in this folder, having an error log from here may help to uncover this.

SergeyKanzhelev: Like here: https://github.com/kubernetes/kubernetes/blob/cb719f75aac98ca24224c563d80d1035079885a5/staging/src/k8s.io/client-go/util/certificate/certificate_store_test.go#L259-L262

neolit123 (commented):
/ok-to-test

neolit123:
please keep your commits squashed to one.
/release-note-none

pkg/volume/util/hostutil/hostutil_linux_test.go
zhijianli88: It sounds good to me.

cluster/images/conformance/go-runner/tar_test.go
zhijianli88: I also noticed that. AFAIK there are so many places(374/458) don't do the cleanup error check in unit test currently.

i might need another cleanup to do this i guess.

SergeyKanzhelev: Why wouldn't you add it in places you introduces the call to `Remove`? This PR requires so many approvals so doing this clean up twice will be challenging =)

It's not blocking from my perspective though

SergeyKanzhelev:
/assign @dashpole 

for the pkg/kubelet/OWNERS approval

SergeyKanzhelev (approved):
with the suggestion to log error from `RemoveAll` calls

jsafrane:
/approve 
`pkg/volume` changes

zhijianli88:
ping

jsafrane:
@zhijianli88, the PR waits for 1.20 development to open, which should happen around 1.19 release. No action is needed from anyone, just a little patience.

zhijianli88:
thank you very much, good to know this.

pkg/volume/util/hostutil/hostutil_linux_test.go
```diff
@@ -283,6 +283,7 @@ func TestGetFileType(t *testing.T) {
 
 	for idx, tc := range testCase {
 		path, cleanUpPath, err := tc.setUp()
+		defer os.RemoveAll(cleanUpPath) // RemoveAll can deal with a empty path ""
```
lavalamp, line 286: This waits until the end of the function, which is ok I guess.

Even if RemoveAll can deal with an empty path, I think it'd be good to only call this if setUp successfully got a path. It's super annoying to have a bug result in a test deleting something it shouldn't...

lavalamp:
Here is blanket approval:

/approve

lavalamp:
Oh, liggitt already got it :) Not sure why this showed up in my email just now...

Issue #87197, opened 2020-01-14
kube-proxy:1.17 for s390x image is broken?


cheeye:
**What happened**:
When I tried to use kubernetes 1.17 for s390x with kubeadm, I got below messages in the kube-proxy pod. It ended up failure of installing cni like flannel. But if I switch back to kubernetes 1.15, everything works fine without these messages. Checked it with `sig-network` in the k8s slack, there could be something broken in the image kube-proxy:1.17 for s390x. 

```
E0108 13:16:53.070052       1 proxier.go:787] Failed to ensure that filter chain KUBE-EXTERNAL-SERVICES exists: error creating chain "KUBE-EXTERNAL-SERVICES": exit status 2: update-alternatives: error: alternative /usr/sbin/iptables-legacy for iptables not registered; not setting
I0108 13:16:53.070129       1 proxier.go:779] Sync failed; retrying in 30s
W0108 13:17:22.902894       1 iptables.go:508] Could not set up iptables canary mangle/KUBE-PROXY-CANARY: error creating chain "KUBE-PROXY-CANARY": exit status 2: update-alternatives: error: alternative /usr/sbin/iptables-legacy for iptables not registered; not setting
```

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): 1.17
- Cloud provider or hardware configuration: s390x
- OS (e.g: `cat /etc/os-release`):  rhel7.7
- Kernel (e.g. `uname -a`):  3.10.0-1062.9.1.el7.s390x #1 SMP Mon Dec 2 08:33:00 EST 2019 s390x s390x s390x GNU/Linux
- Install tools: kubeadm
- Network plugin and version (if this is a network-related bug): flannel
- Others:  image name `k8s.gcr.io/kube-proxy:v1.17.0`

cheeye:
/sig network

Please let me know if I put the improper sig group name. Thanks!

cheeye:
Please let me know if I put the improper sig group name. Thanks.

athenabot:
/triage unresolved

Comment `/remove-triage unresolved` when the issue is assessed and confirmed.

🤖 I am a bot run by vllry. 👩‍🔬

neolit123:
to enable higher verbosity try adding e.g. `--v=5` to the kube-proxy Pod template in the DaemonSet.

> /usr/sbin/iptables-legacy

[...]

PR #92308, opened 2020-06-19
Build a single multi-arch debian-base image

aledbf:
**What type of PR is this?**

/kind cleanup
/kind feature

Use builds (already enabled) to build images, removing qemu-static-* from the process 

Related to https://github.com/kubernetes/kubernetes/issues/87197#issuecomment-583753009

With this change, the docker image contains a history (right now that is empty)

```release-note
Image debian-base is multi-arch, removing the need to have one image per platform.
```

```docs

```


hack/init-buildx.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
```
BenTheElder, line 1: heh, this is familiar :-)

aledbf: on purpose :)

BenTheElder:
I think the potential problem with this is when we build images on top of it if we can't use buildx then we have to be extra careful to select the architecture matching image (easier if they have a name each), we've recently ripped buildx out of other builds in k/k due to the release pipeline :(

aledbf:
Closing. Thank you for the feedback @BenTheElder

PR #92312, opened 2020-06-19
Make Kubernetes aware of the LoadBalancer behaviour

### LINKED ISSUE #66607: Why kube-proxy add external-lb's address to node local iptables rule?
/kind friction

**What happened**:
I have a `LoadBalancer` type service A of address 1.1.1.1. The external loadbalancer of service A is a TLS decoder, it will convert https requests to http `hostport` and `endpoint`.  But since the kube-proxy add the external-lb's address to local iptables rule. Requests of `https//1.1.1.1` will hijack to local http `endpoints`. Then https request failed.

Kube-proxy don't add external-lb's address to local iptables. And the request will go through external-lb.

**Environment**:
1.10.4
Alibaba Cloud
- OS (e.g. from /etc/os-release):
Centos 7.4
- Kernel (e.g. `uname -a`):
3.10.0-693
- Install tools:
kubeadm

### LINKED ISSUE #79783: LoadBalancer's healthcheck packet will be dropped in ipvs mode
**What happened**:

Hi, I'm from the TKE team. In [TencentCloud TKE](https://intl.cloud.tencent.com/product/tke), LB health check packet source ip is LB's own ip (Service's ExternalIP)，and the LB will not do SNAT, when kube-proxy has ipvs enabled in 1.12，the the LoadBalancer's healthcheck packet to NodePort will be dropped by node, so the LB will think the node is unhealthy.  I tested 1.10 and found no such problem, and because the latest k8s version supported in TKE is 1.12, so I haven't tested k8s higher than 1.12. I didn't find any related issues, so I think the latest version should also have this problem.

I compared 1.10 and 1.12，1.12 will add the service's external ip to the `kube-ipvs0`:
``` bash
$ ip a show kube-ipvs0 | grep -A2 170.106.134.124
    inet 170.106.134.124/32 brd 170.106.134.124 scope global kube-ipvs0
       valid_lft forever preferred_lft forever
```
- `170.106.134.124` is service's external ip (LB's ip)
- LB's ip was added to the dev `kube-ipvs0`

then kernel will add a local route in local table automatically:
``` bash
$ ip route show table local | grep 170.106.134.124
local 170.106.134.124 dev kube-ipvs0  proto kernel  scope host  src 170.106.134.124
```
and in 1.10, there is no LB's ip in dev `kube-ipvs0`, it is because of this difference that the packet cannot reach `cbr0` which source ip is LB's ip in 1.12,  I think it's related to an optimization: packets sent to service's external ip are routed directly to the pod via ipvs, which does not actually pass through the lb, check the ipvs rules in 1.12 we can found:
``` bash
ipvsadm -ln | grep -A1 170.106.134.124
TCP  170.106.134.124:80 rr
  -> 172.16.0.70:80               Masq    1      0          0
```

Check the [reference page](https://stupefied-goodall-e282f7.netlify.com/keps/sig-network/0011-ipvs-proxier/#supporting-loadbalancer-service) we can found kube-proxy will `Iterate LB’s ingress IPs, create an ipvs service whose address corresponding LB’s ingress IP`

You may ask how to prove the lb's healthcheck packet was dropped rather than redirected to the pod, I created a Service with type=LoadBalancer, and externalTrafficPolicy=Local, and the endpoint only have one pod, I captured the cbr0's packet on the node which has the service's pod, and there's no packet from lb has been captured, because the service's pod is only available at this node, if it is forwarded normally, it will must reach the `cbr0`.

If we delete the lb's local route in local table, then it will forward normally:
``` bash
ip route del table local local 170.106.134.124 dev kube-ipvs0  proto kernel  scope host  src 170.106.134.124
```
Or an infinite loop to execute this command to delete the ip in `kube-ipvs0`:
``` bash
$ ip addr del 170.106.134.124/32 dev kube-ipvs0
```

Is it possible that the incoming packet's source ip is found in the kernel's local table, it is considered to be the local ip, and then the kernel will drop the packet?

I guess why other cloud vendors didn't encounter this problem is because the source ip of lb's healthcheck packet is internal vip instead of lb's own ip

**Environment**:
```
Client Version: version.Info{Major:"1", Minor:"12+", GitVersion:"v1.12.4-tke.4", GitCommit:"8433e31fd12cae66e59d634f4e7ab6cc13c93a4a", GitTreeState:"clean", BuildDate:"2019-06-25T05:19:07Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"12+", GitVersion:"v1.12.4-tke.3", GitCommit:"ab6e1c10a35382c2ec70036e0e51c201eb3fc3f8", GitTreeState:"clean", BuildDate:"2019-06-18T12:12:29Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}
```
```
NAME="Ubuntu"
VERSION="16.04.1 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.1 LTS"
VERSION_ID="16.04"
UBUNTU_CODENAME=xenial
```
- Kernel (e.g. `uname -a`):
```
Linux VM-1-3-ubuntu 4.4.0-104-generic #127-Ubuntu SMP Mon Dec 11 12:16:42 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

Sh4d1:
**What type of PR is this?**
/kind feature

Implements https://github.com/kubernetes/enhancements/pull/1392
Fixes #79783
Fixes #66607
Naming for the new field still needs to be agreed on
```release-note
NONE
```

```docs
- [KEP]: https://github.com/kubernetes/enhancements/pull/1392
```

/hold
/cc @thockin 
/cc @andrewsykim


pkg/apis/core/types.go
```diff
@@ -3406,8 +3406,23 @@ type LoadBalancerIngress struct {
 	// (typically AWS load-balancers)
 	// +optional
 	Hostname string
+
+	// RouteType specifies the type of route to use for this ingress
+	// Defaults to `VIP`
```
aojea, line 3411: this has to be defaulted here?
https://github.com/kubernetes/kubernetes/blob/70302466f4deeef5838cc420e770b2f9e2fe1425/pkg/apis/core/v1/defaults.go#L99

pkg/proxy/ipvs/proxier_test.go
```diff
@@ -3892,3 +3892,57 @@ func TestFilterCIDRs(t *testing.T) {
...
+		}),
+	)
+
+	fp.syncProxyRules()
+
+	services, err := fp.ipvs.GetVirtualServers()
+	if err != nil {
+		t.Errorf("Failed to get ipvs services, err: %v", err)
+	}
+	if len(services) != 1 {
```
aojea, line 3945: not familiar with ipvs, but curious about this behaviour, in this test, if `LoadBalancerRouteTypeProxy == VIP` , len(services) should be 2?

pkg/apis/core/types.go
Sh4d1: Since it needs to get populated by the cloud controller manager, I'm not sure it's the right place. Either we don't default it or we loop through the `ingress` once `EnsureLoadBalancer` is called. (Not sure when the `SetDefaults_Service` is called exactly)

pkg/proxy/ipvs/proxier_test.go
Sh4d1: In the default and current behaviour, another virtual server is created for the LB IP iirc

aojea: can we use a table test then?

```go
testCases := []struct {
		svcIP string
                svcLBIP string
                routeType v1.LoadBalancerRouteType
		expectedServices    int
	}{
		// case 0
		{
			svcIP: "10.20.30.41",
			svcLBIP:    "1.2.3.4",
			routeType:    v1.LoadBalancerRouteTypeProxy,
                         expectedServices: 1,
		},
		// case 1
		{
			svcIP: "10.20.30.41",
			svcLBIP:    "1.2.3.4",
			routeType:    v1.LoadBalancerRouteTypeVIP,
                         expectedServices: 2,
		},
		// case 2
		{
			svcIP: "fd00:1234::1",
			svcLBIP:    "2001:db8::1",
			routeType:    v1.LoadBalancerRouteTypeProxy,
                         expectedServices: 1,
		},
		// case 3
		{
			svcIP: "fd00:1234::1",
			svcLBIP:    "2001:db8::1",
			routeType:    v1.LoadBalancerRouteTypeVIP,
                         expectedServices: 2,
		},
		// case 4
		{
			svcIP: "fd00:1234::1",
			svcLBIP:    "2001:db8::1",
			routeType:    "",
                         expectedServices: 2,
		},
```

pkg/proxy/ipvs/proxier.go
```diff
@@ -1295,11 +1295,11 @@ func (proxier *Proxier) syncProxyRules() {
 		}
 
 		// Capture load-balancer ingress.
-		for _, ingress := range svcInfo.LoadBalancerIPStrings() {
-			if ingress != "" {
+		for _, ingress := range svcInfo.LoadBalancerIngress() {
+			if ingress.IP != "" && ingress.RouteType != v1.LoadBalancerRouteTypeProxy {
```
aojea, line 1299: if kube-proxy runs against a previous version of the APIserver this field would not be available, should we check first if the field exists?

pkg/apis/core/types.go
aojea: @thockin and/or @andrewsykim  would know better,
 can you please advice here?

pkg/proxy/iptables/proxier_test.go
```diff
@@ -2506,4 +2506,61 @@ COMMIT
 	assert.NotEqual(t, expectedIPTablesWithSlice, fp.iptablesData.String())
 }
 
+func TestLoadBalancerIngressRouteTypeProxy(t *testing.T) {
+	ipt := iptablestest.NewFake()
+	fp := NewFakeProxier(ipt, false)
+	svcIP := "10.20.30.41"
```
aojea, line 2512: table test as in https://github.com/kubernetes/kubernetes/pull/92312/files/0ad979b5324215cdf50bdb4714173b059490a4f8#r443197212 but using `expectedRule bool` instead of `expectedServices    int`?

pkg/proxy/ipvs/proxier.go
Sh4d1: 🤔 in this case the field would just be an empty string no?

thockin: In general we say that clients can be older than servers but not newer.  If you want to be pedantic, you check for nil and handle it as the default

thockin (commented):
Overall this looks pretty straight-forward.  That said, new fields generally need to be gated and disabled by default, so we can do safe rollbacks.  See pkg/features/kube_features.go

pkg/apis/core/types.go
```diff
@@ -3406,8 +3406,23 @@ type LoadBalancerIngress struct {
...
+	// +optional
```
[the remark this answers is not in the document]
lavalamp: It sounds like you have a union / one of structure. It's best to group fields in substructures, having them all loose is very confusing.

I think throwing an error is the best thing to do now. When we implement union / one of, we will likely have a better answer.

pkg/proxy/ipvs/proxier.go
Sh4d1: @thockin do you rather see a IPMode as a `string` or as a `*string`. If we set a default on create, I think we can safely use a `string` and check of empty string instead of `nil`?

pkg/apis/core/types.go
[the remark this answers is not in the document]
Sh4d1: If we are throwing an error, it's weird to have a default right? 
In thats case, we should either set the filed nullable (and treat the null value as the default VIP mode), or add a new `None` mode or something like that?

pkg/apis/core/types.go
```diff
@@ -3406,8 +3406,23 @@ type LoadBalancerIngress struct {
...
+	RouteType LoadBalancerRouteType
```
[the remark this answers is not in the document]
Sh4d1: `ipMode` sounds good, wdyt?

munnerz:
@Sh4d1 are you able to address Tim's comments above? 😄

Sh4d1:
@munnerz kind of, waiting for some answers on some discussions, but I can push the first modifications 😄

Sh4d1:
@thockin I think I addressed all the comments :)

pkg/features/kube_features.go
```diff
@@ -649,6 +649,11 @@ const (
 	// Add support for the HPA to scale based on metrics from individual containers
 	// in target pods
 	HPAContainerMetrics featuregate.Feature = "HPAContainerMetrics"
+
+	// owner: TBD ?
```
aojea, line 653: who is going to own it?

Sh4d1: That's why it's TDB :smile: it needs to be addressed :sweat_smile: Not sure how owning features gates works :)

aojea: what about you? I mean, somebody has to take care to graduate it, usually the person that implements it, otherwise, if it is stale it eventually will be removed

Sh4d1: I mean I'm up for it, just not sure about the implications :smile:

Sh4d1: But yeah, if it's usually the person that implement's it, I'll do it

Sh4d1: Added mysel then :+1:

pkg/apis/core/v1/defaults.go
```diff
@@ -166,6 +166,17 @@ func SetDefaults_Service(obj *v1.Service) {
...
 		// note: conversion logic handles cases where ClusterIPs is used (but not ClusterIP).
 	}
+
+	if utilfeature.DefaultFeatureGate.Enabled(features.LoadBalancerIPMode) &&
+		obj.Spec.Type == v1.ServiceTypeLoadBalancer {
+
+		for _, ing := range obj.Status.LoadBalancer.Ingress {
+			if ing.IPMode == nil {
+				ipMode := v1.LoadBalancerIPModeVIP
+				ing.IPMode = &ipMode
```
MorrisLaw, line 176: Do we want to re-create this variable every time we come across a LoadBalancer Ingress with an `IPMode` value set to nil?

Maybe we can do it once?
```go
ipMode := v1.LoadBalancerIPModeVIP

for _, ing := range obj.Status.LoadBalancer.Ingress {
	if ing.IPMode == nil {
		ing.IPMode = &ipMode
         }
}
```

_or_

Set the value in one line?
```go
for _, ing := range obj.Status.LoadBalancer.Ingress {
	if ing.IPMode == nil {
		ing.IPMode = &v1.LoadBalancerIPModeVIP
         }
}
```

Wdyt?

MorrisLaw (commented):
It's looking good so far! Left a few comments

pkg/apis/core/v1/defaults.go
Sh4d1: Yeah first one definitely (second one won't work since it's a constant 😄) though in the end I think the compiler is smart enough to optimize it 🤔 but yeah I'll update!

pkg/proxy/service.go
```diff
@@ -168,22 +164,24 @@ func (sct *ServiceChangeTracker) newBaseServiceInfo(port *v1.ServicePort, servic
...
-	var ips []string
+	var allIncorrectIPs []string
 	for _, ing := range service.Status.LoadBalancer.Ingress {
-		ips = append(ips, ing.IP)
-	}
+		correctIPs, incorrectIPs := utilproxy.FilterIncorrectIPVersion([]string{ing.IP}, sct.ipFamily)
 
-	if len(ips) > 0 {
-		correctIPs, incorrectIPs := utilproxy.FilterIncorrectIPVersion(ips, sct.ipFamily)
+		// len is either 1 or 0
```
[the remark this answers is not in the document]
Sh4d1: Maybe I'll explain more yep! But before, the function took a list of ips, coming from all ingress of the lb. Now it calls the function with an array of length 1

pkg/proxy/service.go
```diff
@@ -168,22 +164,25 @@ func (sct *ServiceChangeTracker) newBaseServiceInfo(port *v1.ServicePort, servic
 	}
 
 	// Obtain Load Balancer Ingress IPs
-	var ips []string
+	var allIncorrectIPs []string
 	for _, ing := range service.Status.LoadBalancer.Ingress {
-		ips = append(ips, ing.IP)
-	}
+		// []string{ing.IP} have a len of 1, so len(correctIPs) + len(incorrectIPs) == 1
```
MorrisLaw, line 169: thank you for this explanation here!

MorrisLaw (approved):
This LGTM, but we should probably have a second 👍  from @thockin , @lavalamp  or @aojea since they've done some reviewing too.

staging/src/k8s.io/api/core/v1/types.go
```diff
@@ -3971,13 +3971,30 @@ type LoadBalancerIngress struct {
 	// (typically AWS load-balancers)
 	// +optional
 	Hostname string `json:"hostname,omitempty" protobuf:"bytes,2,opt,name=hostname"`
+
+	// IPMode specifies the IP mode to use for this ingress
+	// Defaults to `VIP` if IP is set, null otherwise
```
thockin, line 3976: Can we rewrite this to be more user-facing?

```
IPMode specifies how the load-balancer's IP behaves.  Setting this to "VIP" indicates that the traffic passing through this load-balancer is delivered with the destination IP and port set to the load-balancer's IP and port.  Setting this to "Proxy" indicates that the load-balancer acts like a proxy, delivering traffic with the destination IP and port set to the node's IP and nodePort or to the pod's IP and targetPort.

This field can only be set when the ip field is also set, and defaults to "VIP" if not specified.
```

pkg/apis/core/types.go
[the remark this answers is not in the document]
thockin: See my comments on v1 types.go

pkg/apis/core/validation/validation.go
```diff
@@ -5983,6 +5987,22 @@ func ValidateLoadBalancerStatus(status *core.LoadBalancerStatus, fldPath *field.
 				allErrs = append(allErrs, field.Invalid(idxPath.Child("ip"), ingress.IP, "must be a valid IP address"))
 			}
 		}
+
+		if utilfeature.DefaultFeatureGate.Enabled(features.LoadBalancerIPMode) {
```
thockin, line 5991: The typical pattern is:

```
if gate-enabled {
  if not-specified {
    error: required field
  }
}
if not nil { // remove when gate is removed
  validate value
}
```

thockin: Also - test?

pkg/proxy/iptables/proxier.go
```diff
@@ -1173,15 +1173,17 @@ func (proxier *Proxier) syncProxyRules() {
...
-					args = append(args[:0],
-						"-A", string(kubeServicesChain),
-						"-m", "comment", "--comment", fmt.Sprintf(`"%s loadbalancer IP"`, svcNameString),
-						"-m", protocol, "-p", protocol,
-						"-d", utilproxy.ToCIDR(net.ParseIP(ingress)),
-						"--dport", strconv.Itoa(svcInfo.Port()),
-					)
-					// jump to service firewall chain
-					writeLine(proxier.natRules, append(args, "-j", string(fwChain))...)
+					if !utilfeature.DefaultFeatureGate.Enabled(features.LoadBalancerIPMode) || ingress.IPMode == nil || *ingress.IPMode == v1.LoadBalancerIPModeVIP {
```
thockin, line 1176: no need to check for nil, I think.  If the gate is enabled we should have a value and if not we do this always

pkg/proxy/ipvs/proxier.go
```diff
@@ -1332,11 +1332,11 @@ func (proxier *Proxier) syncProxyRules() {
 		}
 
 		// Capture load-balancer ingress.
-		for _, ingress := range svcInfo.LoadBalancerIPStrings() {
-			if ingress != "" {
+		for _, ingress := range svcInfo.LoadBalancerIngress() {
+			if ingress.IP != "" && (!utilfeature.DefaultFeatureGate.Enabled(features.LoadBalancerIPMode) || ingress.IPMode == nil || *ingress.IPMode == v1.LoadBalancerIPModeVIP) {
```
thockin, line 1336: same as above - no need to check nil, I think

pkg/proxy/service.go
thockin: Indeed.  This loop is now sufficiently complicated that it should be broken out to a func and have a unit test.

Sh4d1:
@thockin fixed all your review 😄

pkg/apis/core/validation/validation.go
```diff
@@ -5983,6 +5987,26 @@ func ValidateLoadBalancerStatus(status *core.LoadBalancerStatus, fldPath *field.
...
+				allErrs = append(allErrs, field.Required(idxPath.Child("ipMode"), "must be specified when `ip` is set"))
+			}
+		}
+
+		if ingress.IPMode != nil {
+			if len(ingress.IP) == 0 {
+				allErrs = append(allErrs, field.Forbidden(idxPath.Child("ipMode"), "may not be used when `ip` is not set"))
+			}
+
+			switch *ingress.IPMode {
```
thockin, line 6002: this switch can simply be `if !supportedLoadBalancerIPMode.Has(*ingress.IPMode)` I think

thockin:
Just one nit and LGTM

pkg/apis/core/validation/validation.go
Sh4d1: true 😅 fixed

Sh4d1:
@thockin On a cluster upgrade + enablement of the feature gate, the existing service will get a `ipMode` with a value of `nil` right? This will crash `kube-proxy`. Is there a solution for this? 🤔 

As I see it, either we still test for `nil` value in `pkg/proxy` or we use a `string` instead of `*string` (but we'd still have to test for the empty string, well it won't panic at least). I'm not aware of any code path that will modify object on a feature gate enablement

aojea:
> (edit: just tested and as expected `kube-proxy` panics  )

https://github.com/kubernetes/kubernetes/pull/92312#discussion_r443197607

IIRC for the first dual-stack try we ended checking that the field was not nil and seems that Tim suggested this too
https://github.com/kubernetes/kubernetes/pull/92312#discussion_r444572042

Sh4d1:
Yeah but he also said https://github.com/kubernetes/kubernetes/pull/92312#discussion_r517592313 😄 

But yeah, I'll add back the `== nil`!

aojea:
> In general we say that clients can be older than servers but not newer.

he is completely right, but the reality is that (some) users upgrade kube-proxy before apiservers :shrug:

Sh4d1:
Yeah agree on that, but even when upgrading the API server first, the `ipMode` will still be `nil` and still need to be checked in `kube-proxy` right?

aojea:
> Yeah agree on that, but even when upgrading the API server first, the `ipMode` will still be `nil` and still need to be checked in `kube-proxy` right?

my understanding is that the check in kube-proxy is behind a feature gate so:
- If the feature-gate is disabled you don't check that field, no problem
- if the feature gate is enabled on both apiserver and kube-prxy, you are defaulting it in the apiserver so it always have a value, right?

what happen if the feature-gate is enabled in kube-proxy and not in apiserver? it will be nil and it will panic if does not check for nil?
@thockin heeelp! :smile:

Sh4d1:
No I mean if you have a 1.20 cluster, you upgrade to 1.21, you enable the feature gate in the API server, then on kube-proxy. Now the previsouly existing service will still have a `ipMode` value of `nil` and `kube-proxy` will crash if the check isn't here 😅

aojea:
> No I mean if you have a 1.20 cluster, you upgrade to 1.21, you enable the feature gate in the API server, then on kube-proxy. Now the previsouly existing service will still have a `ipMode` value of `nil`

I think that once you enable the the featuregate, the defaulting applies for all operations, also when the apiserver reads from etcd, and here you are setting a default value ... but better wait for Tim or @liggitt , they will explain you correctly ;)

thockin:
> On a cluster upgrade + enablement of the feature gate, the existing service will get a ipMode with a value of nil right? 

No - defaults.go is run on read, too so you should never find a nil.

> edit: just tested and as expected kube-proxy panics 

??  Where?  You should NOT be able read a nil when the gate is enabled (and when IP is set)

> the reality is that (some) users upgrade kube-proxy before apiservers

This is, in general, not supported and not tested, and we make no guarantees.  We skew-test and support the other way - back-rev nodes vs modern control-planes.

pkg/proxy/iptables/proxier.go
thockin: What he said ^^^^

thockin:
I was all ready to LGTM, too :)

Sh4d1:
@thockin okay sorry, fixed by https://github.com/kubernetes/kubernetes/pull/92312/commits/20fc86df258846721019e5a20affd4d33942acf0 I wasn't setting the `IPMode` correctly, my bad 😅

pkg/proxy/iptables/proxier.go
Sh4d1: my bad 😁

thockin:
Thanks!

/lgtm
/approve

Sh4d1:
@thockin so this will get shipped in 1.20, instead of 1.21 😕 should we rather revert? Or create an [exception](https://github.com/kubernetes/sig-release/blob/master/releases/EXCEPTIONS.md) for the kep ? 🤔

sttts:
Please squash next time.

Sh4d1:
> Please squash next time.

@sttts yes my bad, totally screwed up this one

kikisdeliveryservice:
@Sh4d1 as this was not being tracked for milestone and unresponsive to the ping on the issue (https://github.com/kubernetes/enhancements/issues/1860) you must create an exception request.

kikisdeliveryservice:
Also wondering how a provisional kep was implemented?

Sh4d1:
> @thockin so this will get shipped in 1.20, instead of 1.21 😕 should we rather revert? Or create an [exception](https://github.com/kubernetes/sig-release/blob/master/releases/EXCEPTIONS.md) for the kep ? 🤔

Hence my question, I initially thought it will get shipped with 1.21 (And completly forgot the `/hold` would get it merged immediatly, and with my messy commits..)

> Also wondering how a provisional kep was implemented?
Hence https://github.com/kubernetes/enhancements/pull/2134 but indeed it should have been updated to `implementable` before. 

And I completly missed the ping on https://github.com/kubernetes/enhancements/issues/1860 sorry!

I personally see no objection to create an exception for this feature, was waiting on Tim's opinion.

thockin:
I don't mind creating an exception, but this is not trying to end-run around the rules. I lost track of which features were targetted for which release.  I am sorry.

@kikisdeliveryservice No intention to bypass the process, apologies.  There's just a lot of cross-checks and I moved too fast.

If you prefer to revert this, I don't object - we can re-propose it and sit on it (though in fact some of this wil make other PRs easier, so maybe we merge some internal parts and sit on the feature).

kikisdeliveryservice:
Thanks @thockin i've brought this to the attention of the Release Leads and we'll let you know shortly.

jeremyrickard:
Commented on the KEP issue for this, but figured I'd drop in here too 👋 

I think at this point (we're SUPER close to code freeze..), we should revert this and not push for 1.20. There is a KEP PR in flight in k/enhancements that targets 1.21, I think it would be best if we targeted 1.21 for this feature. @thockin maybe doing a revert and landing the internal parts of it would be acceptable, but we're also 2 days away from code freeze so.....

I totally get that ya'll were not trying to bypass the process and things fall through the cracks...we're just kind of out of runway to get the KEP for this updated and start wrangling docs IMO.

liggitt:
The alpha e2e job started failing ingress tests as soon as this was merged, see https://testgrid.k8s.io/google-gce#gci-gce-alpha-enabled-default&width=20

liggitt:
This is the failure message:

> "Ingress.extensions \"e2e-example-inggr5wl\" is invalid: status.loadBalancer.ingress[0].ipMode: Required value: must be specified when `ip` is set"

Which seems to indicate that this broke compatibility with a status update that previously worked

pkg/apis/core/validation/validation.go
```diff
@@ -5983,6 +5987,23 @@ func ValidateLoadBalancerStatus(status *core.LoadBalancerStatus, fldPath *field.
 				allErrs = append(allErrs, field.Invalid(idxPath.Child("ip"), ingress.IP, "must be a valid IP address"))
 			}
 		}
+
+		if utilfeature.DefaultFeatureGate.Enabled(features.LoadBalancerIPMode) {
+			if len(ingress.IP) > 0 && ingress.IPMode == nil {
+				allErrs = append(allErrs, field.Required(idxPath.Child("ipMode"), "must be specified when `ip` is set"))
```
liggitt, line 5993: this is what is failing in the alpha e2e job, so apparently this field is not consistently defaulted. if so, we cannot make it required.

Sh4d1: True, it was only defaulted on `Service`, completly missed the `Ingress`. Reverted in #96454 and I'll make sure to include ingress in the new one!

wilsonehusin:
@Sh4d1 hello! we'd like to remove this entry from 1.20 Release Notes since it was reverted. can you replace the release notes with `NONE`?

cc: @andrewsykim

Sh4d1:
@wilsonehusin oops, yes! Done 😄

pkg/proxy/iptables/proxier.go
```diff
@@ -1175,15 +1175,17 @@ func (proxier *Proxier) syncProxyRules() {
...
-					args = append(args[:0],
-						"-A", string(kubeServicesChain),
-						"-m", "comment", "--comment", fmt.Sprintf(`"%s loadbalancer IP"`, svcNameString),
-						"-m", protocol, "-p", protocol,
-						"-d", utilproxy.ToCIDR(net.ParseIP(ingress)),
-						"--dport", strconv.Itoa(svcInfo.Port()),
-					)
-					// jump to service firewall chain
-					writeLine(proxier.natRules, append(args, "-j", string(fwChain))...)
+					if !utilfeature.DefaultFeatureGate.Enabled(features.LoadBalancerIPMode) || *ingress.IPMode == v1.LoadBalancerIPModeVIP {
```
answer1991, line 1178: why we still need LB forward chain? Is there any rules can jump to LB forward chain if ingress IPMode is Proxy?

PR #92184, opened 2020-06-16
Set better default commands for loading images

dims:
Signed-off-by: Davanum Srinivas <davanum@gmail.com>

**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Do not default to `docker`, check the runtime and use correct command for each runtime.

**Which issue(s) this PR fixes**:
 
Fixes https://github.com/kubernetes/test-infra/issues/17929

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #92314, opened 2020-06-19
Set better default commands for loading images - take 2

dims:

**What type of PR is this?**
/kind bug

Do not default to docker, check the runtime and use correct command for each runtime.

Fixes #

I made a mistake! https://github.com/kubernetes/kubernetes/commit/01183e51f084b1a05420f743bca3011b3925a084#diff-6a44976362555568e0243a480da2e86b need to re-apply what was in https://github.com/kubernetes/kubernetes/pull/92184

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster/gce/gci/configure.sh
+++ b/cluster/gce/gci/configure.sh
@@ -376,9 +376,18 @@ function try-load-docker-image {
   set +e
   local -r max_attempts=5
   local -i attempt_num=1
-  until timeout 30 ${LOAD_IMAGE_COMMAND:-docker load -i} "${img}"; do
+
+  if [[ "${CONTAINER_RUNTIME_NAME:-}" == "docker" ]]; then
+    load_image_command=${LOAD_IMAGE_COMMAND:-docker load -i}
+  elif [[ "${CONTAINER_RUNTIME_NAME:-}" == "containerd" ]]; then
+    load_image_command=${LOAD_IMAGE_COMMAND:-ctr -n=k8s.io images import}
+  else
+    load_image_command="${LOAD_IMAGE_COMMAND:-}"
+  fi
+
+  until timeout 30 ${load_image_command} "${img}"; do
     if [[ "${attempt_num}" == "${max_attempts}" ]]; then
-      echo "Fail to load docker image file ${img} after ${max_attempts} retries. Exit!!"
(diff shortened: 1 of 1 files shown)


BenTheElder:
[borrowing this PR to test something non-blocking, will `/skip` it after verifying]
/test  pull-kubernetes-e2e-kind-canary

cluster/gce/gci/configure.sh
```diff
@@ -376,9 +376,18 @@ function try-load-docker-image {
...
   local -r max_attempts=5
   local -i attempt_num=1
-  until timeout 30 ${LOAD_IMAGE_COMMAND:-docker load -i} "${img}"; do
+
+  if [[ "${CONTAINER_RUNTIME_NAME:-}" == "docker" ]]; then
+    load_image_command=${LOAD_IMAGE_COMMAND:-docker load -i}
+  elif [[ "${CONTAINER_RUNTIME_NAME:-}" == "containerd" ]]; then
+    load_image_command=${LOAD_IMAGE_COMMAND:-ctr -n=k8s.io images import}
+  else
+    load_image_command="${LOAD_IMAGE_COMMAND:-}"
```
BenTheElder, line 385: I think we shouldn't allow this else to succeed if the command is empty? maybe:
```suggestion
    load_image_command="${LOAD_IMAGE_COMMAND:?}"
```
or better yet an explicit check + error message

BenTheElder: the latter is probably best. `:?` will just ensure it was set.

bsdnet: If it is unsupported runtime, it should fail at line 390. 
In theory, failing earlier will be better. But I did not see we
have functions like assert or exit in this file.

bsdnet:
From the latest failures, looks environmental variable load_image_command is not set.

https://storage.googleapis.com/kubernetes-jenkins/logs/ci-containerd-e2e-gci-gce/1275221697632407556/artifacts/bootstrap-e2e-master/kube-master-installation.log

So that means CONTAINER_RUNTIME_NAME is not set properly when calling configure.sh

bsdnet:
I am wondering why ${CONTAINER_RUNTIME_NAME:-}, why not ${CONTAINER_RUNTIME_NAME} if CONTAINER_RUNTIME_NAME is one environmental varaible.

bsdnet:
I have a new PR #92401, but still tried to piece things together :)

Issue #92317, opened 2020-06-19
Use PrepullList in e2e framework


alejandrox1:
Similar to https://github.com/kubernetes/kubernetes/issues/90277 we should switch up from "WhiteList" to "AllowList" in the e2e framework.

**What would you like to be added**:
Within the e2e framework, there is a definition of a set of images to be used for a test suite:
https://github.com/kubernetes/kubernetes/blob/403716626d126c850c233a85977de0e6ce60d8ec/test/e2e/framework/pods.go#L60-L63

We should rename this to `AllowList` here and in all other places where it is used:
```
$ grep -nr ImageWhiteList
e2e/framework/pods.go:60:// ImageWhiteList is the images used in the current test suite. It should be initialized in test suite and
e2e/framework/pods.go:63:var ImageWhiteList sets.String
e2e/framework/pods.go:190:		gomega.Expect(ImageWhiteList.Has(c.Image)).To(gomega.BeTrue(), "Image %q is not in the white list, consider adding it to CommonImageWhiteList in test/e2e/common/util.go or NodeImageWhiteList in test/e2e_node/image_list.go", c.Image)
e2e/common/runtime.go:268:			// Images used for ConformanceContainer are not added into NodeImageWhiteList, because this test is
e2e/common/util.go:60:// CommonImageWhiteList is the list of images used in common test. These images should be prepulled
e2e/common/util.go:64:var CommonImageWhiteList = sets.NewString(
e2e_node/runtime_conformance_test.go:51:			// The following images are not added into NodeImageWhiteList, because this test is
e2e_node/image_list.go:46:// NodeImageWhiteList is a list of images used in node e2e test. These images will be prepulled
e2e_node/image_list.go:48:var NodeImageWhiteList = sets.NewString(
e2e_node/image_list.go:64:// updateImageAllowList updates the framework.ImageWhiteList with
e2e_node/image_list.go:69:	// Union NodeImageWhiteList and CommonImageWhiteList into the framework image white list.
e2e_node/image_list.go:70:	framework.ImageWhiteList = NodeImageWhiteList.Union(commontest.CommonImageWhiteList)
e2e_node/image_list.go:72:	framework.ImageWhiteList.Insert(getNodeProblemDetectorImage())
e2e_node/image_list.go:73:	framework.ImageWhiteList.Insert(getSRIOVDevicePluginImage())
e2e_node/image_list.go:152:	images := framework.ImageWhiteList.List()
```

**Why is this needed**:
[...]

Issue #90277, opened 2020-04-19
Use allowlist and denylist in public documentation


Capstan:
**What would you like to be added**:
The documentation around the APIs, e.g., as seen in field comments in https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/policy/v1beta1/generated.proto, uses "white list" and "whitelist", and I suggest changing this to use allowlist instead.

For future internal code and type names, I suggest avoiding terms like "whilelist" and "blacklist" entirely.

**Why is this needed**:
https://developers.google.com/style/inclusive-documentation#features-and-users describes this pretty well. Associating "white" with good/allowed and "black" with bad/disallowed cannot be disentangled from systemic racism. In contrast, allowlist and denylist exactly and merely describe the functions of these fields and variables.

https://github.com/kubernetes/kubernetes/search?p=2&q=%22whitelist%22+OR+%22white+list%22+OR+%22blacklist%22+OR+%22black+list%22&unscoped_q=%22whitelist%22+OR+%22white+list%22+OR+%22blacklist%22+OR+%22black+list%22

celestehorgan:
I stumbled upon this issue WRT Docs PR https://github.com/kubernetes/website/pull/21591 – 

Is there any plan to follow through on this in the existing codebase? I understand that this is the plan going forward, but I'd like to see a retroactive name change of classes, files, and methods in the current codebase for the following reasons:

- As a developer naming a new feature which may involve an allowlist or denylist, I won't look through existing `kubernetes/kubernetes` issues to find this one. Instead, I'll look at _similar classes and functionality within the codebase_ and follow suit.
- As a PR reviewer, I won't look through existing issues either, especially given that I know whitelist/blacklist are currently in use in the codebase. If I see something different (say allowlist/denylist), I'll suggest the developer follow the codebase convention. 
- As an end user developer reading through the documentation, I might want to double-check the underlying implementation of an allowlist or denylist processor. Given that the documentation uses the terminology allowlist and denylist (or simply allow and deny), as an end user developer my expectation is that the underlying codebase does the same. My first instinct is to search the codebase for "allowlist" and "denylist" rather than white/blacklist.


Finally, a personal objection as a technical writer: 

[...]

PR #92318, opened 2020-06-19
Renamed ImageWhiteList to PrePulledImageList

alejandrox1:
Signed-off-by: alejandrox1 <alarcj137@gmail.com>

**What type of PR is this?**
/kind cleanup

In the spirit of https://github.com/kubernetes/kubernetes/issues/90277
xref https://github.com/kubernetes/kubernetes/issues/92317

> Associating "white" with good/allowed and "black" with bad/disallowed cannot be disentangled from systemic racism. In contrast, allowlist and denylist exactly and merely describe the functions of these fields and variables.

I wanted to propose this change to the e2e node code to get the conversation started and try to improve the naming conventions we use throughout the codebase.

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/util/proxy/transport.go
+++ b/staging/src/k8s.io/apimachinery/pkg/util/proxy/transport.go
@@ -101,15 +101,7 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	resp, err := rt.RoundTrip(req)
 
 	if err != nil {
-		message := fmt.Sprintf("Error trying to reach service: '%v'", err.Error())
-		resp = &http.Response{
-			Header:     http.Header{},
-			StatusCode: http.StatusServiceUnavailable,
-			Body:       ioutil.NopCloser(strings.NewReader(message)),
-		}
-		resp.Header.Set("Content-Type", "text/plain; charset=utf-8")
-		resp.Header.Set("X-Content-Type-Options", "nosniff")
-		return resp, nil
+		return nil, fmt.Errorf("error trying to reach service: %w", err)
 	}
 
 	if redirect := resp.Header.Get("Location"); redirect != "" {
--- a/staging/src/k8s.io/apimachinery/pkg/util/proxy/upgradeaware.go
+++ b/staging/src/k8s.io/apimachinery/pkg/util/proxy/upgradeaware.go
@@ -232,6 +232,12 @@ func (h *UpgradeAwareHandler) ServeHTTP(w http.ResponseWriter, req *http.Request
 	proxy.Transport = h.Transport
 	proxy.FlushInterval = h.FlushInterval
 	proxy.ErrorLog = log.New(noSuppressPanicError{}, "", log.LstdFlags)
+	if h.Responder != nil {
+		// if an optional error interceptor/responder was provided wire it
+		// the custom responder might be used for providing a unified error reporting
+		// or supporting retry mechanisms by not sending non-fatal errors to the clients
+		proxy.ErrorHandler = h.Responder.Error
+	}
 	proxy.ServeHTTP(w, newReq)
 }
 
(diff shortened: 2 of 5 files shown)


alejandrox1:
/sig node
/priority important longterm
for review
/hold

test/e2e_node/e2e_node_suite_test.go
```diff
@@ -179,7 +179,7 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 	// This helps with debugging test flakes since it is hard to tell when a test failure is due to image pulling.
 	if framework.TestContext.PrepullImages {
 		klog.Infof("Pre-pulling images so that they are cached for the tests.")
-		updateImageWhiteList()
+		updateImageAllowList()
```
BenTheElder, line 182: I think both of these are the wrong term https://github.com/kubernetes/kubernetes/issues/92317#issuecomment-646825052, `updateImagePrepullList`?

Is there anywhere we use this as a filter? Or only for pre-pulling.

spiffxp (commented):
Like I said in https://github.com/kubernetes/kubernetes/issues/92317#issuecomment-646825052, I don't think "allowlist" accurately describes what this list is used for (though I am in favor of renaming it)

Maybe `updatePrePulledImagesList` for this function?

test/e2e_node/e2e_node_suite_test.go
alejandrox1: Been staring at /test for a while at it seems to be that it is used only for prepulling.

As per changing the name to `updateImagePrepullList` YES and thank you for the suggestion.
Naming is hard :sweat_smile:

alejandrox1:
/retitle Changed updateImageWhiteList to updatePrePulledImagesList

test/e2e_node/image_list.go
```diff
@@ -61,11 +61,11 @@ var NodeImageWhiteList = sets.NewString(
 	"gcr.io/kubernetes-e2e-test-images/node-perf/tf-wide-deep-amd64:1.0",
 )
 
-// updateImageWhiteList updates the framework.ImageWhiteList with
+// updateImageAllowList updates the framework.ImageWhiteList with
```
BenTheElder, line 64: we should also update the `framework.ImageWhiteList` variable?

test/e2e_node/image_list.go
```diff
@@ -61,11 +61,11 @@ var NodeImageWhiteList = sets.NewString(
...
 // 1. the hard coded lists
 // 2. the ones passed in from framework.TestContext.ExtraEnvs
 // So this function needs to be called after the extra envs are applied.
-func updateImageWhiteList() {
+func updateImageAllowList() {
 	// Union NodeImageWhiteList and CommonImageWhiteList into the framework image white list.
 	framework.ImageWhiteList = NodeImageWhiteList.Union(commontest.CommonImageWhiteList)
```
SergeyKanzhelev, line 70: many occurrences of "white" in comment above and variable name.

test/e2e_node/image_list.go
alejandrox1: Done. I wanted to do it in smaller pieces to have this one as an example but might as well rename ImageWhiteList from the framework now.

test/e2e_node/image_list.go
alejandrox1: Thanks for catching that!
I renamed ImageWhiteList from the framework but im leaving NodeImageWhiteList and CommonImageWhiteList for separate PRs, ref https://github.com/kubernetes/kubernetes/issues/92319 and https://github.com/kubernetes/kubernetes/issues/92320

test/e2e_node/image_list.go
BenTheElder: given how painful it is to get PRs in currently, and how tiny this PR is, I think it might make more sense to combine these.

we should probably name these variables and methods consistently anyhow?

alejandrox1: ah the other 2 issues for node and common images are already being worked on.

on consistent naming: yes I think so. i'm looking at https://github.com/kubernetes/kubernetes/pull/92323 and it seems to be that the image "white lists" are indeed lists of images to prepull so yes to consistent naming.

BenTheElder:
poke for rebase

PR #92258, opened 2020-06-18
Add logic to check if local NVMe SSDs in node boot-up script

SidneyShen:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Current logic would assume all the NVMe disks are data disks and applicable for reformat and mount. This will cause the issue when booting disk is also NVMe disk, which will fail the node boot up. This change will check if any additional NVMe disks are required/specified and skip the reformat step otherwise.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #92324, opened 2020-06-19
Automated cherry pick of #92258: Add logic to check if local NVMe SSDs in node boot-up script

SidneyShen:
Cherry pick of #92258 on release-1.18.

#92258: Add logic to check if local NVMe SSDs in node boot-up script

For details on the cherry pick process, see the [cherry pick requests](https://git.k8s.io/community/contributors/devel/sig-release/cherry-picks.md) page.

### DIFF
--- a/cluster/gce/gci/configure-helper.sh
+++ b/cluster/gce/gci/configure-helper.sh
@@ -334,6 +334,13 @@ function ensure-local-ssds() {
   # The following mounts or symlinks NVMe devices
   get-local-disk-num "nvme" "block"
   local nvmeblocknum="${localdisknum}"
+  get-local-disk-num "nvme" "fs"
+  local nvmefsnum="${localdisknum}"
+  # Check if NVMe SSD specified.
+  if [ "${nvmeblocknum}" -eq "0" ] && [ "${nvmefsnum}" -eq "0" ]; then
+    echo "No local NVMe SSD specified."
+    return
+  fi
   local i=0
   for ssd in /dev/nvme*; do
     if [ -e "${ssd}" ]; then


SidneyShen:
@kubernetes/release-managers
Please help with cherry-pick review

SidneyShen:
@kubernetes/release-managers please help with the review and merge, thanks!

tpepper:
I'd like to understand a little bit more the failure scenario, impact, and test coverage on this cherry pick.  The mainline PR is sparse on detail, doesn't say how common or likely this is.

SidneyShen:
Hi @tpepper, thanks for the comments. The original logic in this script has the environment variable **NODE_LOCAL_SSDS_EXT** , the format of which will be "#,scsi/nvme,block/fs". Users can specify multiple configurations by separating them by a semi-colon ex. "2,scsi,fs;1,nvme,block" is a request for 2 SCSI formatted and mounted SSDs and 1 NVMe block device SSD. However the original code (the snippet I modified) assumes all the NVMe disks are attachable data disks. It did this by scanning all the disk symlink under the path "/dev/nvme*" and tries to format and mount available ones. However, it is possible that one of the NVMe is boot disk, which would ultimately fail the node boot. The current fix is a short-term workaround and straightforward, if user did not specify any NVMe disk via **NODE_LOCAL_SSDS_EXT**, the code should directly return instead return. This did not happen before if the boot disk is not behind NVMe controller or user did not specify the local ssd via **NODE_LOCAL_SSDS_EXT**.

tpepper:
> This did not happen before if the boot disk is not behind NVMe controller or user did not specify the local ssd via **NODE_LOCAL_SSDS_EXT**.

Still unclear is the likelihood/impact.  Eg: Was this always happening but rare?  How did we miss it?  Was it impossible before or unlikely?  Is a new GCP feature causing this to now be possible and likely?

SidneyShen:
> > This did not happen before if the boot disk is not behind NVMe controller or user did not specify the local ssd via **NODE_LOCAL_SSDS_EXT**.
> 
> Still unclear is the likelihood/impact. Eg: Was this always happening but rare? How did we miss it? Was it impossible before or unlikely? Is a new GCP feature causing this to now be possible and likely?

GKE currently only supports the SCSI disk as boot disk, there is no way for user to specify a boot disk behind NVMe controller. This failure did not happen on the existing clusters and regular machine types. As GCP is introducing new machine types using NVMe as boot disk, k8s will fail boot up on this new VMs due to this issue. That's said, this is a new issue for new VM types.

tpepper:
> > > This did not happen before if the boot disk is not behind NVMe controller or user did not specify the local ssd via **NODE_LOCAL_SSDS_EXT**.
> > 
> > 
> > Still unclear is the likelihood/impact. Eg: Was this always happening but rare? How did we miss it? Was it impossible before or unlikely? Is a new GCP feature causing this to now be possible and likely?
> 
> GKE currently only supports the SCSI disk as boot disk, there is no way for user to specify a boot disk behind NVMe controller. This failure did not happen on the existing clusters and regular machine types. As GCP is introducing new machine types using NVMe as boot disk, k8s will fail boot up on this new VMs due to this issue. That's said, this is a new issue for new VM types.

Ok.  This sounds like adding feature support then and is less likely to be something we would cherry pick.

tpepper:
Patch release has discussed and decided to treat this as a feature / non-priority-critical-urgent today for 1.17.  If the incidence increases (eg: this becomes something that can be hit, and it is being hit be test-infra) we can always choose in the future to backport it.

PR #92325, opened 2020-06-19
Add pod and container name in log message when container fails to start

brianpursley:
**What type of PR is this?**
/kind feature

I was investigating e2e test flakes and notice this error message doesn't tell you the pod or container that failed.
This PR adds pod and container name to these two log messages to make them more helpful when trying to read the logs.

None

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kubelet/kuberuntime/kuberuntime_manager.go
+++ b/pkg/kubelet/kuberuntime/kuberuntime_manager.go
@@ -793,9 +793,9 @@ func (m *kubeGenericRuntimeManager) SyncPod(pod *v1.Pod, podStatus *kubecontaine
 			// repetitive log spam
 			switch {
 			case err == images.ErrImagePullBackOff:
-				klog.V(3).Infof("%v start failed: %v: %s", typeName, err, msg)
+				klog.V(3).Infof("%v %+v start failed in pod %v: %v: %s", typeName, spec.container, format.Pod(pod), err, msg)
 			default:
-				utilruntime.HandleError(fmt.Errorf("%v start failed: %v: %s", typeName, err, msg))
+				utilruntime.HandleError(fmt.Errorf("%v %+v start failed in pod %v: %v: %s", typeName, spec.container, format.Pod(pod), err, msg))
 			}
 			return err
 		}


pkg/kubelet/kuberuntime/kuberuntime_manager.go
```diff
@@ -793,9 +793,9 @@ func (m *kubeGenericRuntimeManager) SyncPod(pod *v1.Pod, podStatus *kubecontaine
 			// repetitive log spam
 			switch {
 			case err == images.ErrImagePullBackOff:
-				klog.V(3).Infof("%v start failed: %v: %s", typeName, err, msg)
+				klog.V(3).Infof("%v %q start failed in pod %q: %v: %s", typeName, spec.container.Name, format.Pod(pod), err, msg)
```
SergeyKanzhelev, line 796: the two logs above are using `spec.container`, here you opt in for `spec.container.Name`. Would be nice to make logs consistent unless there is a reason to have it differently

(https://github.com/kubernetes/kubernetes/pull/92325/files#diff-8da51ab012ecbdf2b013ecbcf1c37033R784) and (https://github.com/kubernetes/kubernetes/pull/92325/files#diff-8da51ab012ecbdf2b013ecbcf1c37033R788)

SergeyKanzhelev, line 796: not part of this PR, but would it be nice to customize the log message instead of `%v` for the known `err` object? Would be more human friendly

brianpursley: Thanks, I updated it.

Better to have more information than less when trying to troubleshoot and error.

brianpursley: Sure, what do you have in mind?

Looking at this code, I think all this is doing is changing the log level if it is an image pull backoff error, so it doesn't flood the log with error messages.

pkg/kubelet/kuberuntime/kuberuntime_manager.go
```diff
@@ -793,9 +793,9 @@ func (m *kubeGenericRuntimeManager) SyncPod(pod *v1.Pod, podStatus *kubecontaine
 			// repetitive log spam
 			switch {
 			case err == images.ErrImagePullBackOff:
-				klog.V(3).Infof("%v start failed: %v: %s", typeName, err, msg)
+				klog.V(3).Infof("%v %+v start failed in pod %v: %v: %s", typeName, spec.container, format.Pod(pod), err, msg)
```
SergeyKanzhelev, line 796: It can be something like this. Human readable error first before the long structure describing the image. But again, it is a small suggestion

```suggestion
				klog.V(3).Infof("Image pull failed for %v %+v in pod %v: %s", typeName, spec.container, format.Pod(pod), msg)
```

SergeyKanzhelev (approved):

PR #92328, opened 2020-06-19
Replace evalSymlink() with powershell command for subPath_windows.go

jingxu97:
In golang, evalSymlink() does not work if windows disk driver letter is
not assigned. Replace this function with a powershell command to work
around this issue.

This PR address the issue https://github.com/kubernetes/utils/issues/172

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/volume/util/subpath/subpath_windows.go
+++ b/pkg/volume/util/subpath/subpath_windows.go
@@ -21,6 +21,7 @@ package subpath
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"syscall"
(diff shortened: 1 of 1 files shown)


pkg/volume/util/subpath/subpath_windows.go
```diff
@@ -43,18 +44,36 @@ func NewNSEnter(mounter mount.Interface, ne *nsenter.Nsenter, rootDir string) In
...
+// volume (represented by volumeID) to the given path. In this case, the given path is returned.
+func evalPath(path string) (linkedPath string, err error) {
+	cmd := fmt.Sprintf("Get-Item -Path %s | Select-Object -ExpandProperty Target", path)
+	output, err := exec.Command("powershell", "/c", cmd).CombinedOutput()
+	klog.V(4).Infof("evaluate symlink from %s: %s %v", path, string(output), err)
+	if err != nil {
+		return "", err
+	}
+	linkedPath = strings.TrimSpace(string(output))
+	if strings.HasPrefix(linkedPath, "Volume") || strings.HasPrefix(linkedPath, "\\??\\Volume") {
```
ddebroy, line 58: Do you need the first condition: `strings.HasPrefix(linkedPath, "Volume")` ? You can potentially look for other Windows prefixes like `\\.\Volume` or maybe even `\\` for generic UNC path handling.

jingxu97: Because the output from powershell gives Volumexxxx as output without \\

ddebroy: Ref about various path prefixes: https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats

jingxu97: I updated the code here by adding a volume prefix function. @ddebroy PTAL.

ddebroy:
Thanks! /lgtm

PR #92329, opened 2020-06-20
Add deprecated, deprecationWarning fields to CRDs

liggitt:
**What type of PR is this?**
/kind feature

Implements CRD warnings portion of https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1693-warnings / https://github.com/kubernetes/enhancements/issues/1693

```release-note
CustomResourceDefinitions added support for marking versions as deprecated by setting `spec.versions[*].deprecated` to `true`, and for optionally overriding the default deprecation warning with a `spec.versions[*].deprecationWarning` field.
```

/sig api-machinery
/cc @deads2k 
/area custom-resources


staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/customresource_handler.go
```diff
@@ -323,6 +329,12 @@ func (r *crdHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	if utilfeature.DefaultFeatureGate.Enabled(features.WarningHeaders) {
+		for _, w := range crdInfo.warnings[requestInfo.APIVersion] {
+			warning.AddWarning(req.Context(), "", w)
```
roycaihw, line 334: Dumb question: is it okay for `agent` to be empty here?

roycaihw (commented):
Some nits. LGTM overall

staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/customresource_handler.go
liggitt: yes, when we generate the header, an agent of "" is serialized as `-`. this matches how we produce the deprecated warnings for in-tree types

staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1/types.go
```diff
@@ -174,6 +174,17 @@ type CustomResourceDefinitionVersion struct {
...
+	// When set to true, API requests to this version receive a warning header in the server response.
+	// Defaults to false.
+	// +optional
+	Deprecated bool `json:"deprecated,omitempty" protobuf:"varint,7,opt,name=deprecated"`
+	// deprecationWarning overrides the default warning returned to API clients.
+	// May only be set when `deprecated` is true.
+	// The default warning indicates this version is deprecated and recommends use
+	// of the newest served version of equal or greater stability, if one exists.
+	// +optional
+	DeprecationWarning string `json:"deprecationWarning,omitempty" protobuf:"bytes,8,opt,name=deprecationWarning"`
```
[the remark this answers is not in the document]
liggitt: sigh. I guess I could see it for the deprecationWarning field, since explicitly setting an empty string would ignore your explicit value, which isn't great, I guess.

for `deprecation`, it's a bool, so there will only ever be two values, and we will never default deprecated to `true`, so I think a non-pointer is fine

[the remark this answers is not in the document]
roycaihw: I agree for `deprecation` it's fine. For `deprecationWarning`, I don't see why a user would explicitly want an empty warning message.

The question was more about api conventions instead of real use cases. I don't feel strongly about it

/lgtm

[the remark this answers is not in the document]
liggitt: switched deprecationWarning to a pointer and updated validation to disallow explicit empty string values

liggitt:
/assign @lavalamp 
for API review

staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/types.go
```diff
@@ -177,6 +177,15 @@ type CustomResourceDefinitionVersion struct {
 	// Storage flags the version as storage version. There must be exactly one flagged
 	// as storage version.
 	Storage bool
+	// deprecated indicates this version of the custom resource API is deprecated.
+	// When set to true, API requests to this version receive a warning header in the server response.
+	// Defaults to false.
+	Deprecated bool
```
deads2k, line 183: Why don't we simply require someone to write their own DeprecationWarning and avoid having a bool?  I haven't read all the way down, but I doubt we have a very good stock message on these since versions don't follow kube versions.

liggitt: The normal case of "X is deprecated" or "X is deprecated; use Y instead" is easy to construct, and seemed valuable

staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/customresource_handler.go
```diff
@@ -891,6 +913,25 @@ func (r *crdHandler) getOrCreateServingInfoFor(uid types.UID, name string) (*crd
 	return ret, nil
 }
 
+func defaultDeprecationWarning(deprecatedVersion string, crd apiextensionsv1.CustomResourceDefinitionSpec) string {
```
deads2k, line 916: versions for CRDs mapping to kube versions somehow seems weird.  I'd rather avoid this entire method by requiring the deprecation string.

liggitt: This is following the same rules we already follow when determining the order we present the CRD versions in discovery

Issue #77666, opened 2019-05-09
Kubectl plugins: support global options


aermakov-zalando:
**What would you like to be added**:

If a `kubectl-foo` plugin binary exists, both `kubectl foo -n namespace` and `kubectl -n namespace foo` should work. Currently, only the first command would call the plugin binary, the second one would complain about a missing command.

**Why is this needed**:

Plugins should work correctly when using custom aliases that e.g. operate on a specific namespace.

aermakov-zalando:
@kubernetes/sig-cli-feature-requests

aermakov-zalando:
No idea what would be the best way to handle it. The easiest way would be for `kubectl` to transform global options into command arguments, so the two commands would be equivalent.

liggitt:
See https://github.com/juanvallejo/kubernetes/pull/9#discussion_r191944024 and https://github.com/juanvallejo/kubernetes/pull/9#issuecomment-393571640

cc @juanvallejo

aermakov-zalando:
I'd argue that for global arguments the shape is already known, `kubectl options` even prints it out for you. This would allow aliases to work correctly with plugins so kubectl's UX isn't broken.

Elgarni:
@liggitt @juanvallejo any thoughts on this? I'm thinking to open a PR to further discuss this with more details.

liggitt:
I don't see anything to counter the points raised earlier. Opaque pass through of options to kubectl plugins is a fundamental part of the current design, and I'm not eager to trade that for ambiguous relocation of flags and associated arguments.

Elgarni:
I see your point. Can't we though have both? Here is how I see the problem. 
```bash
kubectl [beforeOptions] plugin-command [afterOptions] 
```
The `beforeOptions` should be one of kubectl global options. Then the whole command should be perceived as follows:
```bash
kubectl plugin-command [afterOptions] [beforeOptions]
```
Moving forward with this implementation, I don't see any conflict with the [plugin design](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cli/0024-kubectl-plugins.md)

What do you think?

liggitt:
Given the following plugins:
```
kubectl-foo
kubectl-foo-bar
```
and the invocation
```
kubectl foo --namespace x bar
```
which of the following should be invoked?
```
kubectl-foo --namespace x bar
kubectl-foo-bar --namespace x
```

[...]

PR #92343, opened 2020-06-20
fix(kubectl): print error if users place flags before plugin name

### LINKED ISSUE #884: kubectl plugins do not work when options placed before plugin name
**Edit**:
The root cause of this issue has nothing to do with strace; it is because calling a kubectl plugin with options before the plugin name causes kubectl to not know how to find the plugin.
```
# works
kubectl my-plugin -n some-namespace

# fails, because kubectl stops looking for a plugin name after it sees the -n
kubectl -n some-namespace my-plugin
```

**What happened**:

`kubectl plugin list` shows that I have some plugins installed:
```
$ kubectl plugin list
The following compatible plugins are available:

/usr/bin/kubectl-pkrizak
/usr/local/bin/kubectl-who_can
```

But calling them does not work:
```
$ kubectl pkrizak
Error: unknown command "pkrizak" for "kubectl"
Run 'kubectl --help' for usage.
$ kubectl who-can
Error: unknown command "who-can" for "kubectl"
Run 'kubectl --help' for usage.
```

However, calling them under strace works somehow:
```
$ strace -f -o /tmp/trace kubectl pkrizak
hello there
$ strace -f -o /tmp/trace kubectl who-can
Error: you must specify two or three arguments: verb, resource, and optional resourceName
```

**What you expected to happen**: Plugins listed in `kubectl plugin list` should be called without having to wrap the command with `strace`.

1. Install a kubectl plugin somewhere in `$PATH`
2. Verify that `kubectl plugin list` shows the plugin
3. Run `kubectl <plugin>` and it returns an error
4. Run `strace <args> kubectl <plugin>` and it works

**Environment**:
- Kubernetes client and server versions (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.3", GitCommit:"2e7996e3e2712684bc73f0dec0200d64eec7fe40", GitTreeState:"clean", BuildDate:"2020-05-20T12:52:00Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.12", GitCommit:"e2a822d9f3c2fdb5c9bfbe64313cf9f657f0a725", GitTreeState:"clean", BuildDate:"2020-05-06T05:09:48Z", GoVersion:"go1.12.17", Compiler:"gc", Platform:"linux/amd64"}
```
- Cloud provider or hardware configuration: n/a
```
NAME="Ubuntu"
VERSION="16.04.6 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.6 LTS"
VERSION_ID="16.04"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```

knight42:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

At present if we call kubectl plugins while putting the flags before its name, like `kubectl -n foo bar`, kubectl would print a confusing error since it consider `bar` as its subcommand:
```
Error: unknown command "bar" for "kubectl"
Run 'kubectl --help' for usage.
```

This patch makes kubectl warn users that they cannot place flags before plugin name.

Before:
```
$ kubectl -n foo echo
Error: unknown command "echo" for "kubectl"
Run 'kubectl --help' for usage.
```

After:
```
$ ./_output/bin/kubectl -n foo echo
Error: flags cannot be placed before plugin name: -n
```

Fixes https://github.com/kubernetes/kubectl/issues/884

https://docs.google.com/document/d/1r0YElcXt6G5mOWxwZiXgGu_X6he3F--wKwg-9UBc29I/edit#heading=h.ji0nwpg7dyuz 
sig-cli has discussed this issue a bit on the sig meeting held on July 15th 2020 and decided not to change the current behavior. After discussion with @soltysh on Slack, we both agree that we could improve the error message so that users are not confused.

```release-note
kubectl: print error if users place flags before plugin name
```

```docs

```

### DIFF
--- a/pkg/kubectl/cmd/cmd.go
+++ b/pkg/kubectl/cmd/cmd.go
@@ -318,7 +318,7 @@ func NewDefaultKubectlCommandWithArgs(pluginHandler PluginHandler, args []string
 		// the specified command does not already exist
 		if _, _, err := cmd.Find(cmdPathPieces); err != nil {
 			if err := HandlePluginCommand(pluginHandler, cmdPathPieces); err != nil {
-				fmt.Fprintf(errout, "%v\n", err)
+				fmt.Fprintf(errout, "Error: %v\n", err)
 				os.Exit(1)
 			}
 		}
@@ -393,13 +393,17 @@ func (h *DefaultPluginHandler) Execute(executablePath string, cmdArgs, environme
 // HandlePluginCommand receives a pluginHandler and command-line arguments and attempts to find
 // a plugin executable on the PATH that satisfies the given arguments.
 func HandlePluginCommand(pluginHandler PluginHandler, cmdArgs []string) error {
-	remainingArgs := []string{} // all "non-flag" arguments
-
-	for idx := range cmdArgs {
-		if strings.HasPrefix(cmdArgs[idx], "-") {
+	var remainingArgs []string // all "non-flag" arguments
+	for _, arg := range cmdArgs {
+		if strings.HasPrefix(arg, "-") {
 			break
 		}
-		remainingArgs = append(remainingArgs, strings.Replace(cmdArgs[idx], "-", "_", -1))
+		remainingArgs = append(remainingArgs, strings.Replace(arg, "-", "_", -1))
+	}
+
+	if len(remainingArgs) == 0 {
+		// the length of cmdArgs is at least 1
+		return fmt.Errorf("flags cannot be placed before plugin name: %s", cmdArgs[0])
 	}
 
 	foundBinaryPath := ""


BenTheElder:
Do built in commands allow flags before the subcommand name?

If so I think instead we should identify the flags and pass them to the plugin as arguments so as to not be inconsistent 🤷‍♂️

knight42:
@BenTheElder Built-in commands only allow global flags before the subcommand name, like `--namespace`, `--context`(for a complete list, you can run `kubectl options`), but these flags may not be supported by arbitrary plugins, I think it might be better to ask users to explicitly set them by putting them after plugin name.

BenTheElder:
> ... but these flags may not be supported by arbitrary plugins ..

so passing them along would create the same failure mode as if the user did `kubectl plugin --flag`. this is working as intended then ...?

> I think it might be better to ask users to explicitly set them by putting them after plugin name.

this means you can't generically run `kubectl --namespace <command>` and ensures that the user has to be aware if a command is a plugin or not / leaks that to the caller.

this seems detrimental for prototyping commands as plugins.

brianpursley:
> Do built in commands allow flags before the subcommand name?
> 
> If so I think instead we should identify the flags and pass them to the plugin as arguments so as to not be inconsistent man_shrugging

Yes. but the catch is that it is unclear how we can identify two-part flags like `--namespace test`, because whether a flag is one-part of two-part is an implementation detail of the command you are running.  We can do it for our own commands because we already have the Cobra command defined with all the valid flags to look for.

Example: Let's say you have a plugin called foo and it has a flag called bar and call it like this:

```
kubectl --bar baz foo
```

Is this supposed to be calling the `foo` plugin using `--bar baz` or is it supposed to call the `baz foo` plugin with the flag `--bar`.

If the caller uses an `=` like this `--bar=baz`, it is more explicit and we could code for that, but kubectl allows for `--bar baz` (without equal sign) too.

So I think where we ended up on the issue discussion was let's provide a better error message for the user, other than "plugin not found" so they can know what is wrong.

Ideally you are right though, we should try to make plugins work the same way as our own internal commands.

HERE IS ONE IDEA...

If you want to try to make this handle args before the plugin, one way might be to iterate through all possible combination to check if any match a plugin, so from the example above: `baz foo`, `baz`, `foo`.  If any of them matches a plugin, assume everything else is an arg and pass it through.

BenTheElder:
@brianpursley but kubectl only allows global flags before a command name, right?

 so we don't need to handle arbitrary flags at all (and shouldn't) and this problem is not actually different versus handling a built-in command from a parsing perspective, we only need to pass the flag values along to the plugin

brianpursley:
> ... Built-in commands only allow global flags before the subcommand name, like `--namespace`, `--context`(for a complete list, you can run `kubectl options`)...

I can do this though...
```
kubectl -o json get pods
```

I don't think it matters which flags.  It is just that kubectl already knows the cobra command definition so it is able to interpret the flags properly no matter where they are in the command.

brianpursley:
> @brianpursley but kubectl only allows global flags before a command name, right?

I don't think so, I can do this fine.
```
kubectl -o json get pods

**Comment by knight42** [comment]
> kubectl -o json get pods

thanks for correcting me, but I guess this is not recommended? If I run
```
$ kubectl -w get pod
Error: unknown command "pod" for "kubectl"

Did you mean this?
        top

Run 'kubectl --help' for usage.
```
I kind of doubt that whether we should support flags before subcommand..

**Comment by brianpursley** [comment]
> I kind of doubt that whether we should support flags before subcommand..

Maybe. It seems to be inconsistent, even for built-in commands.

I think some people, like who opened the issue, use it to create aliases though, and for some like namespace, it seems to work for them (except for plugins of course).

**Comment by knight42** [comment]
yeah it makes sense. so.. I guess we all agree that we should pass all flags to the plugin no matter where they are placed?(if so, I am not sure how to deal with `kubectl --bar baz foo` this case, I would simply consider it as `kubectl --bar=baz foo`)

**Comment by BenTheElder** [comment]
> I think some people, like who opened the issue, use it to create aliases though, and for some like namespace, it seems to work for them (except for plugins of course).

this is the use case that jumped to my mind.

the ambiguity issue doesn't exist for global flags kubectl knows about like --namespace, but for flags the plugin accepts it still would I guess.

**Comment by knight42** [comment]
@BenTheElder @briandorsey I have updated the code to let kubectl pass all the flags to plugin, PTAL

**Comment by zhouya0** [comment]
Interesting issue:)
/ok-to-test
/priority back-log

**Comment by knight42** [comment]
force push to address staticcheck error and rebase again latest master branch

/assign @soltysh

File: pkg/kubectl/cmd/cmd.go [hunk 1]
```diff
@@ -390,21 +390,41 @@ func (h *DefaultPluginHandler) Execute(executablePath string, cmdArgs, environme
...
 }
 
+// skipPrependedFlags iterates over all the arguments and return
+// the flags before plugin name and the remaining arguments
+func skipPrependedFlags(args []string) ([]string, []string) {
+	inFlag := false
+	for idx, arg := range args {
+		switch {
+		case strings.HasPrefix(arg, "-"):
+			if !strings.ContainsRune(arg, '=') {
```
brianpursley, line 400: I think this is OK.

At first, I was thinking what about boolean flags that don't require an arg, like `--all-namespaces=true` vs. `--all-namespaces`.

But it turns out that even with built-in commands in the current version of kubectl, using `--all-namespaces` before the command does not work.

This works:
```
kubectl --all-namespaces=true get pods
```
But this does not...
```
kubectl --all-namespaces get pods
```
Even though this does...
```
kubectl get pods --all-namespaces
```

soltysh:
> thanks for correcting me, but I guess this is not recommended? If I run
> 
> ```
> $ kubectl -w get pod
> Error: unknown command "pod" for "kubectl"
> 
> Did you mean this?
>         top
> 
> Run 'kubectl --help' for usage.
> ```
> 
> I would get this error. I kind of doubt that whether we should support flags before subcommand..

This is because `-w` is expecting `true|false` after it. Intricacies of cobra :shrug: so get is being passed as true, iirc and pod is being recognized as command and thus you're getting an error.

pkg/kubectl/cmd/cmd.go
soltysh: yeah, cobra :wink:

pkg/kubectl/cmd/cmd_test.go
```diff
@@ -73,6 +73,12 @@ func TestKubectlCommandHandlesPlugins(t *testing.T) {
 			expectPlugin:     "plugin/testdata/kubectl-foo",
 			expectPluginArgs: []string{"--bar"},
 		},
+		{
+			name:             "test that flags before plugin name are passed to plugin",
+			args:             []string{"kubectl", "--flag1=bar", "-n", "baz", "foo", "--flag2", "-f3", "qux", "arg"},
```
soltysh, line 78: This does assume that every flag has an argument to it, which is consistent with what we have in kubectl.

soltysh (approved):
/lgtm
/approve

knight42:
quite flaky 😥

soltysh (approved):
/lgtm
/retest

liggitt:
/hold
does this cause the problem described in https://github.com/kubernetes/kubernetes/issues/77666#issuecomment-578227020?

pwittrock:
If the plugin doesn't accept the global flags it will either:

- Fail saying it doesn't recognize the flag
- Silently ignore its value

The plugin may also use the flags differently from how the global flags are documented.  E.g. `-s` and `-v` could mean different things to the plugin.

WDYT of disabling global flags for plugins with an error message saying plugin flags must be provided after the plugin sub command?

liggitt:
> The plugin may also use the flags differently from how the global flags are documented. E.g. -s and -v could mean different things to the plugin.

That's the main issue from kubectl's perspective, since it doesn't know if a plugin treats a given flag `--foo` as having an accompanying value or not, so it doesn't know whether to just relocate `--foo` or to relocate the next value as well `--foo bar`.

> WDYT of disabling global flags for plugins with an error message saying plugin flags must be provided after the plugin sub command?

I'm not quite sure what you're envisioning. What would that look like in the scenario described in https://github.com/kubernetes/kubernetes/issues/77666#issuecomment-577951716? How would we determine a plugin invocation was desired as opposed to a command typo?

knight42:
> does this cause the problem described in #77666 (comment)?

actually I think it is unable to tackle this problem 100% correctly since the plugin is an external program and we cannot know how it handle the flags in advance. given that, in my opinion it might be better to pass all flags to it rather than failing with a confusing error.

soltysh:
I must admit I initially had mixed feeling about it, but after re-reading the threads Jordan pointed out earlier I think it's worthy a broader discussion. I'll comment in the linked issues about the decision sig-cli will make. 
/lgtm cancel
/approve cancel

knight42:
https://docs.google.com/document/d/1r0YElcXt6G5mOWxwZiXgGu_X6he3F--wKwg-9UBc29I/edit#heading=h.ji0nwpg7dyuz
sig-cli has discussed this issue a bit on the sig meeting held on July 15th 2020 and decided not to change the current behavior. After discussion with @soltysh on Slack, we both agree that we could improve the error message so that users are not confused.

zhouya0 (approved):
/lgtm

nikhita:
@soltysh 👋 Can you approve?

/retest

knight42:
force push to get rid of flakes hopefully

pkg/kubectl/cmd/cmd.go
```diff
@@ -393,13 +393,17 @@ func (h *DefaultPluginHandler) Execute(executablePath string, cmdArgs, environme
...
+		if strings.HasPrefix(arg, "-") {
 			break
 		}
-		remainingArgs = append(remainingArgs, strings.Replace(cmdArgs[idx], "-", "_", -1))
+		remainingArgs = append(remainingArgs, strings.Replace(arg, "-", "_", -1))
+	}
+
+	if len(remainingArgs) == 0 {
+		// the length of cmdArgs is at least 1
+		return fmt.Errorf("flags cannot be placed before plugin name: %s", cmdArgs[0])
```
soltysh, line 406: If I invoke `kubectl foo --flag bar` this error will complain about `foo`, but should about `--flag`. I'd suggest to just print the error w/o anything it'll usually look like this:
```
kubectl foo --flag bar
Error: flags cannot be placed before plugin name
```
Alternatively you'd need to look for the first flag, but that's not necessary, imo.

soltysh (changes requested):
/lgtm cancel

soltysh:
This is not needed for 1.19
/milestone clear

pkg/kubectl/cmd/cmd.go
knight42: If the binary `kubectl-foo` does not exist, I think the following output is expected
```
Error: unknown command "foo" for "kubectl"

Did you mean this?
        top

Run 'kubectl --help' for usage.
```

If `kubectl-foo` does exist, kubectl with this patch should work without complaining.

soltysh (approved):
/lgtm
/approve
/hold cancel

PR #92344, opened 2020-06-20
Restore docker focal version in gci nodes

### LINKED ISSUE #90709: Docker repository does not yet support focal fossa
We have some scripts that install docker/containerd etc and are having trouble installing on focal based images as  https://download.docker.com/linux does not yet support focal fossa. So in https://github.com/kubernetes/kubernetes/pull/90702 we have a patch that uses `bionic` debs for focal. 

Additional context, please see ci test failures in https://github.com/kubernetes/kubernetes/issues/90667 and https://github.com/kubernetes/kubernetes/issues/90668.  

This issue is to track fixing the script back up when focal support is added to https://download.docker.com/linux

jesusha123:
**What type of PR is this?**
/kind bug

This change reverts a workaround needed to install docker in ubuntu focal fossa gci nodes.

Fixes #90709

The workaround was installing docker on focal fossa nodes from the bionic repo instead of the focal fossa repo because the focal fossa repo did not exist. Docker team has now created the focal fossa repo, so we no longer need the workaround.

```release-note
NONE
```

### DIFF
--- a/cluster/gce/gci/configure.sh
+++ b/cluster/gce/gci/configure.sh
@@ -423,12 +423,7 @@ function install-docker {
     software-properties-common \
     lsb-release
 
-  # focal repo for docker is not yet available, so we use bonic for now
-  # https://github.com/kubernetes/kubernetes/issues/90709
   release=$(lsb_release -cs)
-  if [ "$release" == "focal" ]; then
-    release="bionic";
-  fi
 
   # Add the Docker apt-repository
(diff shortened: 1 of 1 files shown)


jherrera123:
Taking a quick look at the focal repo packages, it is missing s390x and maybe armhf and ppc64el architectures. However, the modified script is for gce, which according to the doc, only supports Intel and AMD CPUs (x86-64), so it shouldn't affect the test negatively.

Issue #92275, opened 2020-06-18
hypekube kube-proxy: Chain 'KUBE-MARK-DROP' does not exist


ialidzhikov:
**What happened**:
kube-proxy v1.16.11 fails to execute iptables-restore:
```
{"log":"I0618 15:35:56.843819       1 service.go:373] Adding new service port \"kube-system/kube-dns:dns\" at 10.223.0.10:53/UDP\n","stream":"stderr","time":"2020-06-18T15:35:56.844167118Z"}
{"log":"I0618 15:35:56.843832       1 service.go:373] Adding new service port \"kube-system/calico-typha:calico-typha\" at 10.223.28.54:5473/TCP\n","stream":"stderr","time":"2020-06-18T15:35:56.844172318Z"}
{"log":"I0618 15:35:56.843844       1 service.go:373] Adding new service port \"default/kubernetes:https\" at 10.223.0.1:443/TCP\n","stream":"stderr","time":"2020-06-18T15:35:56.844176518Z"}
{"log":"I0618 15:35:56.843857       1 service.go:373] Adding new service port \"kube-system/allow-udp-egress:dummy\" at 10.223.112.137:1234/UDP\n","stream":"stderr","time":"2020-06-18T15:35:56.844181318Z"}
{"log":"I0618 15:35:56.843876       1 service.go:373] Adding new service port \"kube-system/metrics-server:\" at 10.223.80.184:443/TCP\n","stream":"stderr","time":"2020-06-18T15:35:56.844305721Z"}
{"log":"I0618 15:35:56.843890       1 service.go:373] Adding new service port \"kube-system/vpn-shoot:openvpn\" at 10.223.86.14:4314/TCP\n","stream":"stderr","time":"2020-06-18T15:35:56.844312221Z"}
{"log":"I0618 15:35:57.102753       1 proxier.go:1519] Opened local port \"nodePort for kube-system/vpn-shoot:openvpn\" (:32175/tcp)\n","stream":"stderr","time":"2020-06-18T15:35:57.111034942Z"}
{"log":"I0618 15:35:57.102963       1 proxier.go:1519] Opened local port \"nodePort for kube-system/allow-udp-egress:dummy\" (:31061/udp)\n","stream":"stderr","time":"2020-06-18T15:35:57.111055742Z"}
[...]

PR #92281, opened 2020-06-19
add iproute2 to hyperkube base image

waynr:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:

This is a very naive attempt to address one of the problems described in #92272, specifically the need to have the `iproute2` package available in the hyperkube image in order for kubelet to do its thing.

**Which issue(s) this PR fixes**:

Fixes #92272

**Special notes for your reviewer**:

I am also considering adding the following to the hyperkube image:

```
# update-alternatives --set iptables /usr/sbin/iptables-legacy
# update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
# update-alternatives --set arptables /usr/sbin/arptables-legacy
# update-alternatives --set ebtables /usr/sbin/ebtables-legacy
```

Based on docs in https://wiki.debian.org/nftables

I think this change would be justified because it would follow a principle of least surprise for users upgrading to a new patch version of hyperkube since presumably anyone who was using previous versions of hyperkube, like me, were also running it on a system using the default legacy iptables backend.

I am currently working on building a custom hyperkube from this branch to validate that the `iproute2` package install fixes my second problem since I solved the first by updating my VM image to debian 10 (but I can revert that if reviewers agree that it is surprising to change the default iptables backend targeted by hyperkube).

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

Issue #92272, opened 2020-06-18
hyperkube kubelet: KillPodSandboxError ... networkPlugin cni failed to teardown pod ... network: neither iptables nor ip6tables usable


waynr:
**What happened**:

* I attempted to bump the 1.16, 1.17, and 1.18 versions in my team's kubernetes image repo to 1.16.11, 1.17.7, and 1.18.4, respectively.
* The images built successfully but none of the e2e tests we run, including the kubernetes conformance e2e tests, ever actually ran correctly without a clear indication as to why -- ginkgo was just mysteriously succeeding without any tests passed, skipped, or failed. this prompted me to take a closer look a 1.18.4 cluster.
* I expected corednes pods to be in "Running" status but instead found:
```
kube-system   coredns-6b6854dcbf-tqxgn          0/1     Completed   0          5h24m
kube-system   coredns-6b6854dcbf-wztzn          0/1     Completed   0          5h24m
```
* ...and the following event on one of those pods:
```
  Normal  SandboxChanged  98s (x627 over 136m)  kubelet, test-1-18-4-0b8b9360-3fhas  Pod sandbox changed, it will be killed and re-created.
```
* This even prompted me to log on to a worker node and notice repeated messages like the following in the kubelet logs:
```
Jun 18 21:41:53 test-1-18-4-06729635-3f4er docker[1211]: E0618 21:41:53.316852    1291 pod_workers.go:191] Error syncing pod f364dc8e-6e7d-4869-976a-678a3ce913ac ("coredns-6b6854dcbf-2t9cv_kube-system(f364dc8e-6e7d-4869-976a-678a3ce913ac)"), skipping: failed to "KillPodSandbox" for "f364dc8e-6e7d-4869-976a-678a3ce913ac" with KillPodSandboxError: "rpc error: code = Unknown desc = networkPlugin cni failed to teardown pod \"coredns-6b6854dcbf-2t9cv_kube-system\" network: neither iptables nor ip6tables usable"
```

This led me to grep through the kubernetes repo at v1.18.4 for the message `network: neither iptables nor ip6tables usable` but I didn't find anything. So I did the same thing in the containernetworking plugins repo and found the following references:

* https://github.com/containernetworking/plugins/blob/ad10b6fa91aacd720f1f9ab94341a97a82a24965/plugins/meta/portmap/portmap.go#L120
[...]

Issue #92250, opened 2020-06-18
hyperkube: kube-proxy 1.16.11 does not work on CentOS7


dmitry-irtegov:
**What happened**:
We install k8s 1.16.11 on CentOS 7.7.1908 (Core) node using custom installer 
kubelet and CNI binaries are extracted from k8s.gcr.io/hyperkube-amd64:v1.16.11 image.
Everything works, except kube-proxy (and, obviously, anything that depends on it).
It spams log with the message:
```
E0618 09:42:42.811741       7 proxier.go:1418] Failed to execute iptables-restore: exit status 4 (iptables-restore v1.8.2 (nf_tables): 
line 86: CHAIN_USER_DEL failed (Device or resource busy): chain KUBE-SEP-ARQWXEG2ZEM2P3WE
)
```
Service IP are not available from this node.

**What you expected to happen**:
kube-proxy works.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
This does NOT happen on 1.16.11/Ubuntu 18.04 and 1.16.10/Centos 7 on the same OS image.
Looks like iptables native/legacy mode detection is somehow broken on CentOS 7.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.16.11
- Cloud provider or hardware configuration: vSphere, 4 CPU / 8192 MB RAM
- OS (e.g: `cat /etc/os-release`): CentOS Linux release 7.7.1908 (Core)
- Kernel (e.g. `uname -a`): 3.10.0-1062.9.1.el7.x86_64 #1 SMP
- Install tools: custom
- Network plugin and version (if this is a network-related bug):
- Others:

athenabot:
/sig network

These SIGs are my best guesses for this issue. Please comment `/remove-sig <name>` if I am incorrect about one.

🤖 I am a bot run by vllry. 👩‍🔬

justaugustus:
/sig release
/area release-eng
/priority critical-urgent
cc: @kubernetes/sig-network-bugs @kubernetes/release-engineering

athenabot:
/triage unresolved

Comment `/remove-triage unresolved` when the issue is assessed and confirmed.

🤖 I am a bot run by vllry. 👩‍🔬

superseb:
[...]

PR #92354, opened 2020-06-21
hyperkube: Build debian-hyperkube-base v1.1.0 image

justaugustus:
**What type of PR is this?**

/kind regression
/area dependency
/priority critical-urgent

- Use debian-iptables v12.1.0 as base image
  Should ensure kube-proxy behaves the same in hyperkube as it would for
  the standard kube-proxy image.
- Removes the following packages (which are already present in the
  debian-iptables image):
  - conntrack
  - ebtables
  - iptables
  - ipset
  - kmod
  - netbase

Signed-off-by: Stephen Augustus <saugustus@vmware.com>

Attempt to fix https://github.com/kubernetes/kubernetes/issues/92275, https://github.com/kubernetes/kubernetes/issues/92272, https://github.com/kubernetes/kubernetes/issues/92250.
Should supersede https://github.com/kubernetes/kubernetes/pull/92281.

```release-note

```

```docs

```

### DIFF
--- a/build/debian-hyperkube-base/Dockerfile
+++ b/build/debian-hyperkube-base/Dockerfile
@@ -41,19 +41,13 @@ RUN clean-install \
     ca-certificates \
     ceph-common \
     cifs-utils \
-    conntrack \
     e2fsprogs \
     xfsprogs \
-    ebtables \
     ethtool \
     git \
     glusterfs-client \
-    iptables \
-    ipset \
     jq \
-    kmod \
     openssh-client \
-    netbase \
     nfs-common \
     socat \
     udev \
--- a/build/debian-hyperkube-base/Makefile
+++ b/build/debian-hyperkube-base/Makefile
@@ -19,12 +19,12 @@
 
 REGISTRY?=gcr.io/k8s-staging-build-image
 IMAGE?=$(REGISTRY)/debian-hyperkube-base
-TAG?=v1.0.0
+TAG?=v1.1.0
 ARCH?=amd64
 ALL_ARCH = amd64 arm arm64 ppc64le s390x
 
 BASE_REGISTRY?=us.gcr.io/k8s-artifacts-prod/build-image
-BASEIMAGE?=$(BASE_REGISTRY)/debian-base-$(ARCH):v2.1.0
+BASEIMAGE?=$(BASE_REGISTRY)/debian-iptables-$(ARCH):v12.1.0
 CNI_VERSION?=v0.8.6
 
 TEMP_DIR:=$(shell mktemp -d)
--- a/build/dependencies.yaml
+++ b/build/dependencies.yaml
@@ -84,8 +84,6 @@ dependencies:
       match: debian_base_version=
     - path: build/workspace.bzl
       match: tag =
-    - path: build/debian-hyperkube-base/Makefile
-      match: BASEIMAGE\?\=\$\(BASE_REGISTRY\)\/debian-base-\$\(ARCH\)
     - path: build/debian-iptables/Makefile
       match: BASEIMAGE\?\=\$\(BASE_REGISTRY\)\/debian-base-\$\(ARCH\)
     - path: cluster/images/etcd/Makefile
@@ -102,7 +100,7 @@ dependencies:
       match: us\.gcr\.io\/k8s-artifacts-prod\/build-image\/debian-base:v\d+\.\d+\.\d+
 
   - name: "k8s.gcr.io/debian-hyperkube-base"
-    version: 1.0.0
+    version: 1.1.0
     refPaths:
     - path: build/debian-hyperkube-base/Makefile
       match: TAG\?=
@@ -120,6 +118,8 @@ dependencies:
     refPaths:
     - path: build/common.sh
       match: debian_iptables_version=
+    - path: build/debian-hyperkube-base/Makefile
+      match: BASEIMAGE\?\=\$\(BASE_REGISTRY\)\/debian-iptables-\$\(ARCH\):v\d+\.\d+\.\d+
     - path: build/workspace.bzl
       match: tag =
 


justaugustus:
/assign @BenTheElder @danwinship @cblecker 
/area release-eng

cc-ing reporters/commenters:
cc: @ialidzhikov @vpnachev @juliantaylor @majst01 @waynr @superseb @dmitry-irtegov @aojea

aojea:
/lgtm
for the network changes
/hold
for the image dependencies changes, is not something I'm familiar enough to review

saschagrunert (approved):
/lgtm

aojea:
/hold cancel
sascha lgtmed

danwinship:
So where does kubelet get run from when you're using hyperkube? The iptables wrapper script was written under the assumption that kubelet would _not_ use the wrapper script.

Specifically, the iptables wrapper tries to use the same mode that other things on the system are using, but it can only figure that out if _something_ in the base system sets up some iptables rules before anything that uses the iptables wrapper runs. Normally, that something would be kubelet, which always creates a few iptables rules before starting any pods. But if kubelet is also going to end up using the iptables wrapper rather than the system iptables binaries, then we're relying on there being something else in the system that sets up some iptables rules before kubelet runs.

(Though... I guess if kubelet is being run out of the hyperkube image by docker or something like that, then docker would have set up some iptables rules before kubelet runs, so again things would work...)

justaugustus:
Given the conversation in Slack (https://kubernetes.slack.com/archives/C2C40FMNF/p1592857876443000), it seems like this should be reasonable to move forward with?

If so, can I get an `/approve`?
/assign @caseydavenport @thockin @dcbw

thockin:
@danwinship Even if there are no other rules, as long as kubelet and kube-proxy use the same logic, they should make the same choice, right?

IOW, I think this seems OK, but I am anxious about saying that with authority.  You are closer to the impl..

danwinship:
The theoretical failure mode I was worried about is:

1. system boots, /usr/sbin/iptables points to iptables-nft
2. some stuff starts up, but nothing has touched iptables yet
3. kubelet starts, runs iptables wrapper, falls back to iptables-legacy by default because there are no iptables rules at all
4. firewall starts, uses iptables-nft, creates a bunch of rules
5. kube-proxy starts, runs iptables wrapper, sees that there are more nft rules than legacy rules, so uses nft
6. kubelet and kube-proxy are now out of sync

But I guess there's really no obvious way for that to happen because if you're running kubelet out of the hyperkube image, you have to be using docker, in which case 2 would actually be "some stuff starts up, including docker, which creates iptables rules using iptables-nft", so then 3 would be "kubelet starts, sees iptables-nft rules, uses iptables-nft" and everything would work.

Someone responded in the thread Stephen linked to above and he was using docker to start kubelet, but he was also running into a bunch of other problems because the mountpoints he was using for that container ended up breaking iptables 1.8's use of `/etc/alternatives`, etc.

The hyperkube image is already gone in 1.19, right? Maybe the easiest fix here would just be to revert it back to the old debian image, since we don't have to maintain it going forward anyway?

danwinship:
(Suggesting that because of the other problems the slack user was having. If we only care about "picking the right iptables mode", then yes, I think switching from debian-base to debian-iptables should fix that.)

justaugustus:
> Someone responded in the thread Stephen linked to above and he was using docker to start kubelet, but he was also running into a bunch of other problems because the mountpoints he was using for that container ended up breaking iptables 1.8's use of `/etc/alternatives`, etc.

I would consider their downstream mountpoint... "unsupported" is maybe the wrong word, but not the most important thing to solve. I haven't personally seen the mountpoint issues reported elsewhere, so I think it's local to their environment.

> The hyperkube image is already gone in 1.19, right? Maybe the easiest fix here would just be to revert it back to the old debian image, since we don't have to maintain it going forward anyway?

I do wish that was a potential option.
While hyperkube is deprecated from 1.19 onwards, it's still present in the 1.18 --> 1.16 branches, which means we're on the hook to support it until we EOL Kubernetes 1.18.

> (Suggesting that because of the other problems the slack user was having. If we only care about "picking the right iptables mode", then yes, I think switching from debian-base to debian-iptables should fix that.)

Yep, yep, picking the right iptables mode is what I'd like to solve first.
There's a good chance that base image change will resolve the currently open issues.

dmitry-irtegov:
I think there is deeper problem with current iptables selection logic.

There are manifests that still are hardcoded to use iptables-legacy.  At least, dns-node-cache and flannel.

If one of these manifests is used on the cluster, it creates enough legacy rules to force kube-proxy to legacy mode.

On normal startup it is impossible, but think about a pod restart....

BenTheElder:
kube-proxy shouldn't be starting after pods like dns-node-cache and flannel on normal startup, and on pod restart it shouldn't be a problem if the other network tools in use also correctly implement something like this If they use iptables.

If they don't implement this, those tools already left the system broke if the host used iptables / nft not matching them, and kube proxy is matching the host.

BenTheElder:
On restart kube-proxy will also have lots of old kube-proxy rules.

BenTheElder:
This flake is a known problematic test.
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92354/pull-kubernetes-e2e-kind/1275668877891604482

The following flake with the base image though, that one is ??

dmitry-irtegov:
> ... if the other network tools in use also correctly implement something like this If they use iptables.
This makes sence.  But you probably should mention this in release notes and/or docs.   So maintainers of these tools will know what they should do and why.

Also, the recommendation to unload/blacklist ip_tables kernel module might help. So people will know what breaks and why.

BenTheElder:
??
```

1/4992 Tests Failed. | expand_less
-- | --
[sig-node] RuntimeClass should reject a Pod requesting a RuntimeClass with an unconfigured handler expand_less7stest/e2e/common/runtimeclass.go:47 Jun 24 07:08:27.187: Error creating Pod Unexpected error:     <*errors.StatusError \| 0xc001d6e460>: {         ErrStatus: {             TypeMeta: {Kind: "", APIVersion: ""},             ListMeta: {                 SelfLink: "",                 ResourceVersion: "",                 Continue: "",                 RemainingItemCount: nil,             },             Status: "Failure",             Message: "pods \"test-runtimeclass-runtimeclass-1410-unconfigured-handler-\" is forbidden: pod rejected: RuntimeClass \"runtimeclass-1410-unconfigured-handler\" not found",             Reason: "Forbidden",             Details: {                 Name: "test-runtimeclass-runtimeclass-1410-unconfigured-handler-",                 Group: "",                 Kind: "pods",                 UID: "",                 Causes: nil,                 RetryAfterSeconds: 0,             },             Code: 403,         },     }     pods "test-runtimeclass-runtimeclass-1410-unconfigured-handler-" is forbidden: pod rejected: RuntimeClass "runtimeclass-1410-unconfigured-handler" not found occurred test/e2e/framework/pods.go:83 | [sig-node] RuntimeClass should reject a Pod requesting a RuntimeClass with an unconfigured handler expand_less | 7s | test/e2e/common/runtimeclass.go:47 Jun 24 07:08:27.187: Error creating Pod Unexpected error:     <*errors.StatusError \| 0xc001d6e460>: {         ErrStatus: {             TypeMeta: {Kind: "", APIVersion: ""},             ListMeta: {                 SelfLink: "",                 ResourceVersion: "",                 Continue: "",                 RemainingItemCount: nil,             },             Status: "Failure",             Message: "pods \"test-runtimeclass-runtimeclass-1410-unconfigured-handler-\" is forbidden: pod rejected: RuntimeClass \"runtimeclass-1410-unconfigured-handler\" not found",             Reason: "Forbidden",             Details: {                 Name: "test-runtimeclass-runtimeclass-1410-unconfigured-handler-",                 Group: "",                 Kind: "pods",                 UID: "",                 Causes: nil,                 RetryAfterSeconds: 0,             },             Code: 403,         },     }     pods "test-runtimeclass-runtimeclass-1410-unconfigured-handler-" is forbidden: pod rejected: RuntimeClass "runtimeclass-1410-unconfigured-handler" not found occurred test/e2e/framework/pods.go:83
[sig-node] RuntimeClass should reject a Pod requesting a RuntimeClass with an unconfigured handler expand_less | 7s
test/e2e/common/runtimeclass.go:47 Jun 24 07:08:27.187: Error creating Pod Unexpected error:     <*errors.StatusError \| 0xc001d6e460>: {         ErrStatus: {             TypeMeta: {Kind: "", APIVersion: ""},             ListMeta: {                 SelfLink: "",                 ResourceVersion: "",                 Continue: "",                 RemainingItemCount: nil,             },             Status: "Failure",             Message: "pods \"test-runtimeclass-runtimeclass-1410-unconfigured-handler-\" is forbidden: pod rejected: RuntimeClass \"runtimeclass-1410-unconfigured-handler\" not found",             Reason: "Forbidden",             Details: {                 Name: "test-runtimeclass-runtimeclass-1410-unconfigured-handler-",                 Group: "",                 Kind: "pods",                 UID: "",                 Causes: nil,                 RetryAfterSeconds: 0,             },             Code: 403,         },     }     pods "test-runtimeclass-runtimeclass-1410-unconfigured-handler-" is forbidden: pod rejected: RuntimeClass "runtimeclass-1410-unconfigured-handler" not found occurred test/e2e/framework/pods.go:83


```

BenTheElder:
agree re: relnotes @justaugustus

justaugustus:
> ??

Where are you seeing that failure?

> agree re: relnotes @justaugustus

This PR only builds the new image, which is why it doesn't include a release note.
Follow up PRs will actually utilize the new image (user-facing change) and those will include release notes.

BenTheElder:
ack

re the failure: the last kind flake, it's not related to this PR, but it is very ??

Issue #91414, opened 2020-05-25
SecurityContext windowsOptions doesn't work on kind:Deployment


Maellooou:
**What happened**:
I add podSecurityPolicy on my Kubernetes cluster with the parameter RUNASUSER = MustRunAsNonRoot.
I try to create a Deployment with Windows docker image and the securityContext.windowsOptions.runAsUserName

I used the template here : https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/windows/run-as-username-pod.yaml

It works well, but if I add the deployment object like that :

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: run-as-username-demo
  namespace: admin-test
spec:
  selector:
    matchLabels:
      app: run-as-username-demo
  replicas: 1
  template:
    metadata:
      labels:
        app: run-as-username-demo
    spec:
      securityContext:
        windowsOptions:
          runAsUserName: "ContainerUser"
      containers:
      - name: run-as-username-demo
        image: mcr.microsoft.com/windows/servercore:ltsc2019
        command: ["ping", "-t", "localhost"]
```

It doesn't work with the following issue : 

`Error: container has runAsNonRoot and image will run as root`

**Environment**:
- Kubernetes version (use `kubectl version`): v1.16.4
- Cloud provider or hardware configuration: on premise

Thanks for your help

/sig windows

PR #92355, opened 2020-06-21
fix windows container root validate

### LINKED ISSUE #91482: PodSecurityPolicy rule: 'MustRunAsNonRoot' and windowsOptions: runAsUserName not working
Hello, 

**What happened**:
When we apply PodSecurityPolicy with rule:

```
runAsUser:
    # Require the container to run without root privileges.
    rule: 'MustRunAsNonRoot'
```

WIndows containers are still blocked instead we put: 

```
  securityContext:
   windowsOptions:
      runAsUserName: "ContainerUser"
```

the pod is still blocked with logs: 

`  Warning  Failed          1s (x10 over 72s)  kubelet, kubenodewin04  Error: container has runAsNonRoot and image will run as root `

the Container should start

Create Psp disableling Root Users:

```
runAsUser:
    # Require the container to run without root privileges.
    rule: 'MustRunAsNonRoot'
```

Create a pod Windows with runAsUserName:

```
apiVersion: v1
kind: Pod
metadata:
  name: run-as-username-pod-demo
spec:
  securityContext:
    windowsOptions:
      runAsUserName: "ContainerUser"
  containers:
  - name: run-as-username-demo
    image: mcr.microsoft.com/windows/servercore:ltsc2019
    command: ["ping", "-t", "localhost"]
  nodeSelector:
    kubernetes.io/os: windows
```

**Environment**:
- OS (e.g: `cat /etc/os-release`): CentOS Linux 7 (Core) / windows server 2019 version 1809
- Kernel (e.g. `uname -a`):  3.10.0-514.10.2.el7.x86_64
- Others:
/sig windows

wawa0210:
**What type of PR is this?**

/kind bug
/kind feature

From https://github.com/kubernetes/enhancements/blob/master/keps/sig-windows/20190103-windows-node-support.md#v1container

Windows container does not have a root user,but kubelet will still check the user of the windows container when creating the pod.

The current inspection conditions are based on the `RunAsNonRoot` and `RunAsUser` of the Linux container, and the results checked based on these conditions are incorrect

Fixes #91482 #91414 

```release-note
Ignore root user check when windows pod starts
```

```docs

```

/sig windows
/assign @dchen1107

### DIFF
--- a/pkg/kubelet/kuberuntime/security_context_others.go
+++ b/pkg/kubelet/kuberuntime/security_context_others.go
@@ -1,0 +1,51 @@
+// +build !windows
+
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kuberuntime
+
+import (
+	"fmt"
+
+	"k8s.io/api/core/v1"
+	"k8s.io/kubernetes/pkg/securitycontext"
+)
+
+// verifyRunAsNonRoot verifies RunAsNonRoot.
+func verifyRunAsNonRoot(pod *v1.Pod, container *v1.Container, uid *int64, username string) error {
+	effectiveSc := securitycontext.DetermineEffectiveSecurityContext(pod, container)
+	// If the option is not set, or if running as root is allowed, return nil.
+	if effectiveSc == nil || effectiveSc.RunAsNonRoot == nil || !*effectiveSc.RunAsNonRoot {
+		return nil
+	}
+
+	if effectiveSc.RunAsUser != nil {
+		if *effectiveSc.RunAsUser == 0 {
+			return fmt.Errorf("container's runAsUser breaks non-root policy")
+		}
+		return nil
+	}
+
+	switch {
+	case uid != nil && *uid == 0:
+		return fmt.Errorf("container has runAsNonRoot and image will run as root")
+	case uid == nil && len(username) > 0:
+		return fmt.Errorf("container has runAsNonRoot and image has non-numeric user (%s), cannot verify user is non-root", username)
+	default:
+		return nil
+	}
+}
--- a/pkg/kubelet/kuberuntime/security_context_windows.go
+++ b/pkg/kubelet/kuberuntime/security_context_windows.go
@@ -1,0 +1,65 @@
+// +build windows
+
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kuberuntime
+
+import (
+	"fmt"
+	"k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+	"k8s.io/kubernetes/pkg/securitycontext"
+)
+
+var (
+	windowsRootUserName = "ContainerAdministrator"
+)
+
+// verifyRunAsNonRoot verifies RunAsNonRoot on windows.
+// https://github.com/kubernetes/enhancements/blob/master/keps/sig-windows/20190103-windows-node-support.md#v1container
+// Windows does not have a root user, we cannot judge the root identity of the windows container by the way to judge the root(uid=0) of the linux container.
+// According to the discussion of sig-windows, at present, we assume that ContainerAdministrator is the windows container root user,
+// and then optimize this logic according to the best time.
+// https://docs.google.com/document/d/1Tjxzjjuy4SQsFSUVXZbvqVb64hjNAG5CQX8bK7Yda9w
+func verifyRunAsNonRoot(pod *v1.Pod, container *v1.Container, uid *int64, username string) error {
+	effectiveSc := securitycontext.DetermineEffectiveSecurityContext(pod, container)
+	// If the option is not set, or if running as root is allowed, return nil.
+	if effectiveSc == nil || effectiveSc.RunAsNonRoot == nil || !*effectiveSc.RunAsNonRoot {
+		return nil
+	}
+	if effectiveSc.RunAsUser != nil {
+		klog.Warningf("Windows container does not support SecurityContext.RunAsUser, please use SecurityContext.WindowsOptions")
+	}
+	if effectiveSc.SELinuxOptions != nil {
+		klog.Warningf("Windows container does not support SecurityContext.SELinuxOptions, please use SecurityContext.WindowsOptions")
+	}
+	if effectiveSc.RunAsGroup != nil {
+		klog.Warningf("Windows container does not support SecurityContext.RunAsGroup")
+	}
+	if effectiveSc.WindowsOptions != nil {
+		if effectiveSc.WindowsOptions.RunAsUserName != nil {
+			if *effectiveSc.WindowsOptions.RunAsUserName == windowsRootUserName {
+				return fmt.Errorf("container's runAsUser (%s) which will be regarded as root identity and will break non-root policy", username)
+			}
+			return nil
+		}
+	}
+	if len(username) > 0 && username == windowsRootUserName {
+		return fmt.Errorf("container's runAsUser (%s) which will be regarded as root identity and will break non-root policy", username)
+	}
+	return nil
+}
--- a/pkg/kubelet/kuberuntime/BUILD
+++ b/pkg/kubelet/kuberuntime/BUILD
@@ -28,6 +28,8 @@ go_library(
         "labels.go",
         "legacy.go",
         "security_context.go",
+        "security_context_others.go",
+        "security_context_windows.go",
     ],
     importpath = "k8s.io/kubernetes/pkg/kubelet/kuberuntime",
     deps = [
@@ -105,7 +107,8 @@ go_test(
         "kuberuntime_sandbox_test.go",
         "labels_test.go",
         "legacy_test.go",
-        "security_context_test.go",
+        "security_context_others_test.go",
+        "security_context_windows_test.go",
     ],
     embed = [":go_default_library"],
     deps = [
--- a/pkg/kubelet/kuberuntime/security_context.go
+++ b/pkg/kubelet/kuberuntime/security_context.go
@@ -17,8 +17,6 @@ limitations under the License.
 package kuberuntime
 
 import (
-	"fmt"
-
 	v1 "k8s.io/api/core/v1"
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 	"k8s.io/kubernetes/pkg/security/apparmor"
@@ -74,31 +72,6 @@ func (m *kubeGenericRuntimeManager) determineEffectiveSecurityContext(pod *v1.Po
 	synthesized.ReadonlyPaths = securitycontext.ConvertToRuntimeReadonlyPaths(effectiveSc.ProcMount)
 
 	return synthesized
-}
-
-// verifyRunAsNonRoot verifies RunAsNonRoot.
-func verifyRunAsNonRoot(pod *v1.Pod, container *v1.Container, uid *int64, username string) error {
-	effectiveSc := securitycontext.DetermineEffectiveSecurityContext(pod, container)
-	// If the option is not set, or if running as root is allowed, return nil.
-	if effectiveSc == nil || effectiveSc.RunAsNonRoot == nil || !*effectiveSc.RunAsNonRoot {
-		return nil
-	}
-
-	if effectiveSc.RunAsUser != nil {
-		if *effectiveSc.RunAsUser == 0 {
-			return fmt.Errorf("container's runAsUser breaks non-root policy")
-		}
-		return nil
-	}
-
-	switch {
-	case uid != nil && *uid == 0:
-		return fmt.Errorf("container has runAsNonRoot and image will run as root")
-	case uid == nil && len(username) > 0:
-		return fmt.Errorf("container has runAsNonRoot and image has non-numeric user (%s), cannot verify user is non-root", username)
-	default:
-		return nil
-	}
 }
 
 // convertToRuntimeSecurityContext converts v1.SecurityContext to runtimeapi.SecurityContext.
--- a/pkg/kubelet/kuberuntime/security_context_others_test.go
+++ b/pkg/kubelet/kuberuntime/security_context_others_test.go
@@ -1,0 +1,140 @@
+// +build !windows
+
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kuberuntime
+
+import (
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestVerifyRunAsNonRoot(t *testing.T) {
+	pod := &v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			UID:       "12345678",
+			Name:      "bar",
+			Namespace: "new",
+		},
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{
+				{
+					Name:            "foo",
+					Image:           "busybox",
+					ImagePullPolicy: v1.PullIfNotPresent,
+					Command:         []string{"testCommand"},
+					WorkingDir:      "testWorkingDir",
+				},
+			},
+		},
+	}
+
+	rootUser := int64(0)
+	anyUser := int64(1000)
+	runAsNonRootTrue := true
+	runAsNonRootFalse := false
+	for _, test := range []struct {
+		desc     string
+		sc       *v1.SecurityContext
+		uid      *int64
+		username string
+		fail     bool
+	}{
+		{
+			desc: "Pass if SecurityContext is not set",
+			sc:   nil,
+			uid:  &rootUser,
+			fail: false,
+		},
+		{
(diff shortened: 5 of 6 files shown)


pkg/kubelet/kuberuntime/kuberuntime_container.go
```diff
@@ -240,12 +240,16 @@ func (m *kubeGenericRuntimeManager) generateContainerConfig(container *v1.Contai
...
 		return nil, cleanupAction, err
 	}
-
-	// Verify RunAsNonRoot. Non-root verification only supports numeric user.
-	if err := verifyRunAsNonRoot(pod, container, uid, username); err != nil {
-		return nil, cleanupAction, err
+	// https://github.com/kubernetes/enhancements/blob/master/keps/sig-windows/20190103-windows-node-support.md#v1container
+	// Windows does not have a root user, we cannot judge the root identity of the windows container by the way to judge the root of the linux container
+	// SecurityContext.RunAsUser, and SecurityContext.RunAsNonRoot These are Linux-specific attributes and cannot be used on windows containers
+	// Therefore, the judgment logic here does not apply to windows container
```
SergeyKanzhelev, line 246: would it be helpful to report an error back that configuration may be incorrect? Not necessarily by failing this verification, but maybe with the log message?

pkg/kubelet/kuberuntime/kuberuntime_container.go
```diff
@@ -240,12 +240,16 @@ func (m *kubeGenericRuntimeManager) generateContainerConfig(container *v1.Contai
...
+	if goruntime.GOOS != "windows" {
```
SergeyKanzhelev, line 247: would it be possible to add a test?

pkg/kubelet/kuberuntime/kuberuntime_container.go
wawa0210: Both windows container and linux will cover this piece of code, but it only makes sense to execute on linux. Therefore, I feel that there is no need to output some confusing logs on windows

SergeyKanzhelev: why do you think it will be confusing? If these settings were set by user, it may be important to know that settings were ignored. Or the assumption here that those log messages will never be seen by the person who set these configuration settings?

wawa0210: The processing logic of `verifyRunAsNonRoot` is not suitable for windows containers. Many of the judgment fields are based on Linux containers, such as `SecurityContext.RunAsUser`, `image.uid`, `SecurityContext.RunAsNonRoot`. So I have circumvented this logic~ Maybe Windows container should be judged separately and used for user input check. Is this better? Do you think?

jsturtevant: I think it would be cleaner to split it out and judge differently.  Could use file suffix for the implementation as is done for [applyPlatformSpecificContainerConfig](https://github.com/kubernetes/kubernetes/blob/ee53488f19ff2fdd94cf311662e8a62845162871/pkg/kubelet/kuberuntime/kuberuntime_container.go#L277) in the [windows](https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/pkg/kubelet/kuberuntime/kuberuntime_container_windows.go#L37) and [linux](https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/pkg/kubelet/kuberuntime/kuberuntime_container_linux.go#L36) files.  For windows this can be a noop and if later there are privileged containers we would be able to add logic cleanly.

pkg/kubelet/kuberuntime/security_context_others.go
```diff
@@ -0,0 +1,51 @@
+// +build !windows
+
+/*
+Copyright 2016 The Kubernetes Authors.
```
SergeyKanzhelev, line 4: nit 

```suggestion
Copyright 2020 The Kubernetes Authors.
```

pkg/kubelet/kuberuntime/security_context_windows.go
```diff
@@ -0,0 +1,31 @@
+// +build windows
+
+/*
+Copyright 2016 The Kubernetes Authors.
```
SergeyKanzhelev, line 4:
```suggestion
Copyright 2020 The Kubernetes Authors.
```

pkg/kubelet/kuberuntime/security_context_others.go
```diff
@@ -0,0 +1,51 @@
+// +build !windows
```
SergeyKanzhelev, line 1: same in `security_context_test.go`?

pkg/kubelet/kuberuntime/security_context_others.go
wawa0210: done .

pkg/kubelet/kuberuntime/security_context_others.go
```diff
@@ -0,0 +1,51 @@
...
+
+/*
```
SergeyKanzhelev, line 3: the pattern in file naming is `security_context_windows.go` and `security_context.go` In couple places I saw `_unix` or `_linux` suffices. I'd vote for consistency and drop `_other`

pkg/kubelet/kuberuntime/security_context_others.go
wawa0210: I didn't get what you meant.

pkg/kubelet/kuberuntime/security_context_others.go
wawa0210: Here I have a very uncertain thing is that other platforms besides Linux and Windows support this feature, so I chose to use others~~

pkg/kubelet/kuberuntime/security_context_others.go
SergeyKanzhelev: the file security_context_test.go calls into the method you modified. If you compile different version on windows - running these tests on windows would fail. So I suggest to add the same `build !windows` to the file containing tests

SergeyKanzhelev: Also worth renaming the test file with the same pattern as the product code. With `_other` or `_non_windows` suffix.

pkg/kubelet/kuberuntime/security_context_others.go
SergeyKanzhelev: thank you for pointing out. Somehow all examples I saw was not using this pattern. I guess it's ok. If not previous examples my suggestion would be `security_context_non_windows.go`. It may be the most descriptive name

wawa0210: About this, I went to look at some of the existing designs of kubernetes.

- If the platform support is very clear, it is generally `*_linux.go`, `*_windows.go`, `*_unspport.go`. 
- If it is possible to distinguish between windows and others, it is generally `*_others.go`, `*_windows.go`.

I think it’s better to follow the existing naming rules, otherwise it will confuse people , what do you think?

pkg/kubelet/kuberuntime/security_context_others.go
wawa0210: changed.

pkg/kubelet/kuberuntime/security_context_others.go
SergeyKanzhelev: sure, lets keep it. It looks consistent with the pattern. Thank you for looking into it

SergeyKanzhelev (approved):
I still think it may be valuable to log warning if unsupported setting was set for windows. But this is an additional feature to log this warning and PR is ok.

ddebroy:
From a sig-windows perspective, I wonder if it makes sense to bar `runAsUsername` from being `ContainerAdministrator` (https://kubernetes.io/docs/tasks/configure-pod-container/configure-runasusername/). `ContainerAdministrator` does somewhat emulate `root` user in Linux contexts. Let's discuss this is a bit in sig-windows.

dashpole:
@ddebroy let me know the outcome of the sig-windows discussions.  I can approve once sig-windows has decided how to handle this.

jsturtevant:
We [discussed this at sig-windows today](https://docs.google.com/document/d/1Tjxzjjuy4SQsFSUVXZbvqVb64hjNAG5CQX8bK7Yda9w/edit#) and came to the consensus we should be more restrictive initially as it will be hard to constrain later.  Rather than ignore `RunAsUserName` we should block when the user is `ContainerAdministrator`. 

@wawa0210 could you make those updates?

/cc @ddebroy @michmike

wawa0210:
> We [discussed this at sig-windows today](https://docs.google.com/document/d/1Tjxzjjuy4SQsFSUVXZbvqVb64hjNAG5CQX8bK7Yda9w/edit#) and came to the consensus we should be more restrictive initially as it will be hard to constrain later. Rather than ignore `RunAsUserName` we should block when the user is `ContainerAdministrator`.

Of course, I can try to change. Originally I planned to participate in this sig-windows discussion too, unfortunately I missed it due to time constraints. I will go to watch the video on youtube, the details of your discussion, and then make changes

jsturtevant:
thanks! feel free to ask questions in the sig-windows slack channel if you have any

pkg/kubelet/kuberuntime/security_context_windows.go
```diff
@@ -0,0 +1,55 @@
...
+	if effectiveSc.WindowsOptions != nil {
+		if effectiveSc.WindowsOptions.RunAsUserName != nil {
+			if *effectiveSc.WindowsOptions.RunAsUserName == windowsRootUserName {
+				return fmt.Errorf("container's runAsUser (%s) which will be regarded as root identity and will be break non-root policy", username)
+			}
+		}
+
+	}
+	if len(username) > 0 && username == windowsRootUserName {
+		return fmt.Errorf("container's runAsUser (%s) which will be regarded as root identity and will be break non-root policy", username)
```
jsturtevant, line 52:
```suggestion
		return fmt.Errorf("container's runAsUser is (%s) which will be regarded as root identity and will break non-root policy", username)
```

pkg/kubelet/kuberuntime/security_context_windows.go
```diff
@@ -0,0 +1,55 @@
...
+func verifyRunAsNonRoot(pod *v1.Pod, container *v1.Container, uid *int64, username string) error {
+	effectiveSc := securitycontext.DetermineEffectiveSecurityContext(pod, container)
+	// If the option is not set, or if running as root is allowed, return nil.
+	if effectiveSc == nil || effectiveSc.RunAsNonRoot == nil || !*effectiveSc.RunAsNonRoot {
+		return nil
+	}
+	if effectiveSc.WindowsOptions != nil {
+		if effectiveSc.WindowsOptions.RunAsUserName != nil {
+			if *effectiveSc.WindowsOptions.RunAsUserName == windowsRootUserName {
+				return fmt.Errorf("container's runAsUser (%s) which will be regarded as root identity and will be break non-root policy", username)
```
jsturtevant, line 46:
```suggestion
				return fmt.Errorf("container's runAsUser is (%s) which will be regarded as root identity and will break non-root policy", username)
```

jsturtevant:
It would be nice to see an e2e test for this added: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/windows/security_context.go#L15

pkg/kubelet/kuberuntime/security_context_windows.go
```diff
@@ -0,0 +1,55 @@
...
+			}
+		}
```
SergeyKanzhelev, line 48: linux version of this method will return nil here. Meaning that if `WindowsOptions.RunAsUserName` was set to a value that is not root, it will OK that independently from values of `uid ` and `username `. Why is this one different?

pkg/kubelet/kuberuntime/security_context_windows.go
```diff
@@ -0,0 +1,55 @@
...
+
+	}
+	if len(username) > 0 && username == windowsRootUserName {
+		return fmt.Errorf("container's runAsUser (%s) which will be regarded as root identity and will be break non-root policy", username)
+	}
```
SergeyKanzhelev, line 53: Do you want to report a warning or error if `*effectiveSc.RunAsUser` was set to some (any) value? This configuration may indicate an intent to configure security that cannot be satisfied on windows

wawa0210: Adding this warning message is indeed friendly for windows

pkg/kubelet/kuberuntime/security_context_windows.go
wawa0210: Already fixed
The logic of linux container is to judge by the security context of pod and container first, if the user does not input, judge by mirror metadata. Windows reuses this logic

wawa0210:
> It would be nice to see an e2e test for this added: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/windows/security_context.go#L15

I have some doubts about adding e2e test here. I am adjusting `verifyRunAsNonRoot` judgment logic. This function is only to verify whether the current parameters are legal, and does not include the real pod creation logic. I don't know how to write this e2e

michmike:
@jsturtevant and @wawa0210 , is this supposed to go in for v1.19?
i don't see a milestone set

wawa0210:
> @jsturtevant and @wawa0210 , is this supposed to go in for v1.19?
> i don't see a milestone set

I think this pr can be released with 1.19, what is your opinion? @jsturtevant

SergeyKanzhelev (approved):

wawa0210:
@michmike @jsturtevant 

If you agree, I suggest this pr can be released with v1.19

jsturtevant:
Yes it is a bug fix so I think it should be in 1.19

/milestone v1.19

jsturtevant:
@michmike can you help asign the milestone?

wawa0210:
> @michmike can you help asign the milestone?

I think this `verifyRunAsNonRoot` method is not a typical e2e scenario, it just acts as a checker and will not eventually change the pod parameters. Is unit testing enough?

jsturtevant:
> I have some doubts about adding e2e test here. I am adjusting verifyRunAsNonRoot judgment logic. This function is only to verify whether the current parameters are legal, and does not include the real pod creation logic. I don't know how to write this e2e.

I think there are two scenarios for this, which are tested for linux (https://github.com/kubernetes/kubernetes/blob/master/test/e2e/auth/pod_security_policy.go)

- if the psp with `runasroot` is applied the pod should be scheduled if ContainerUser
- if the psp is applied and ContainerAdministrator it should be in `CreateContainerConfigError` state with the message `Error: container has runAsNonRoot `

That said I think the unit tests are probably enough and given the [current state of of psp](https://github.com/kubernetes/enhancements/issues/5) I think this can probably move ahead and we could open a separate issue for the adding end to end tests and add as good-first-issue.  We don't have a ton of those for windows and would be nice to have more for new contributors.

thoughts?

jsturtevant:
It looks like the test `pull-kubernetes-e2e-azure-file-windows` is failing due to the schedule windows pods which your fixing on https://github.com/kubernetes/kubernetes/pull/93063

wawa0210:
> I think there are two scenarios for this, which are tested for linux (https://github.com/kubernetes/kubernetes/blob/master/test/e2e/auth/pod_security_policy.go)
> 
> * if the psp with `runasroot` is applied the pod should be scheduled if ContainerUser
> * if the psp is applied and ContainerAdministrator it should be in `CreateContainerConfigError` state with the message `Error: container has runAsNonRoot `
> 
> That said I think the unit tests are probably enough and given the [current state of of psp](https://github.com/kubernetes/enhancements/issues/5) I think this can probably move ahead and we could open a separate issue for the adding end to end tests and add as good-first-issue. We don't have a ton of those for windows and would be nice to have more for new contributors.
> 
> thoughts?

I agree with you, I will recreate an issue for e2e later, I hope more contributors will participate

liggitt:
since this hasn't been reviewed/approved and we're at the end of 1.19 code freeze, moving to 1.20

/milestone v1.20

liggitt:

immuzz:
SIG Windows Backlog Meeting

Next Step: Need 1.20 Tag and merging. @marosset to work on that

wawa0210:
@jsturtevant @michmike @SergeyKanzhelev 
Can you take the time to help review this pr again?

pkg/kubelet/kuberuntime/security_context_windows.go
```diff
@@ -0,0 +1,65 @@
...
+	effectiveSc := securitycontext.DetermineEffectiveSecurityContext(pod, container)
+	// If the option is not set, or if running as root is allowed, return nil.
+	if effectiveSc == nil || effectiveSc.RunAsNonRoot == nil || !*effectiveSc.RunAsNonRoot {
+		return nil
+	}
+	if effectiveSc.RunAsUser != nil {
+		klog.Warningf("Windows container does not support SecurityContext.RunAsUser, please use SecurityContext.WindowsOptions")
+	}
+	if effectiveSc.SELinuxOptions != nil {
+		klog.Warningf("Windows container does not support SecurityContext.SELinuxOptions, please use SecurityContext.WindowsOptions")
```
dashpole, line 48: This is a big deal, right?  Someone thinks they are running with security options that we aren't applying.  Should this fail more loudly?

If i'm interpreting this correctly, I would suggest emiting an event, or even returning an error here.  Same comment for other unsupported options.

dashpole (commented):
Should setting WindowsOptions on linux containers also fail more loudly?

pkg/kubelet/kuberuntime/security_context_windows.go
SergeyKanzhelev: I think the argumentation here is that these settings were likely set by some plugin. See https://github.com/kubernetes/kubernetes/pull/93475#discussion_r486666465

This is why warning was good enough from my perspective, but I'd like somebody from windows sig to confirm

marosset: Maybe we should add a check to https://github.com/kubernetes/kubernetes/blob/360659102c8ae94a5bee6e9cbbc3d1160db58428/pkg/apis/core/validation/validation.go#L3769 that verifies only only of SecurityContext.SELinuxOptions or SecurityContext.WindowsOptions are set.

From my perspective we should encourage plugin/automation authors to use node selectors if they are going to blanket apply security policy to pods rather than working around that assumption in the kubelet.
@jsturtevant thoughts?

wawa0210: Because these do not cause cri errors, it is just that the features of linux container on windows container cannot take effect, and vice versa. If kubelet fails more loudly, it will affect existing applications (adding wrong WindowsOptions or LinuxOptions). So I think the warning level log may be a way to make sense (may need to be more detailed)

wawa0210: @marosset 

I think this is a very good idea. When apiserver pre-verification, if the user explicitly sets `SecurityContext.SELinuxOptions` or `SecurityContext.WindowsOptions` for verification, if the discussion passes, I can try to open another A pr. 
But kubelet is really responsible for calling cri to create pod logic. Should it also have its own necessary verification logic?

liggitt: > Maybe we should add a check to https://github.com/kubernetes/kubernetes/blob/360659102c8ae94a5bee6e9cbbc3d1160db58428/pkg/apis/core/validation/validation.go#L3769 that verifies only one of SecurityContext.SELinuxOptions or SecurityContext.WindowsOptions are set.

Tightening API validation is not generally possible (see https://github.com/kubernetes/kubernetes/issues/64841). If that combination of settings is never legitimate (is a multi-OS image possible?), we could add that to the list of things we want to warn about (https://github.com/kubernetes/kubernetes/issues/94626)

dashpole: Thanks @liggitt, that addresses my concern.

dashpole: Thanks @liggitt, that addresses my concern.

marosset: This also addresses my concern.

PR #92357, opened 2020-06-21
Introduce feature gates to Scheduler component config

### LINKED ISSUE #84922: introduce FeatureGates to KubeSchedulerConfiguration
introduce FeatureGates to KubeSchedulerConfiguration
/assign
**Why is this needed**:
This threads the --feature-gates flag through the KubeSchedulerConfiguration object

pancernik:
**What type of PR is this?**
/kind api-change

Introduce feature gates to Scheduler CLI flags and component config

Fixes #84922

```release-note
kube-scheduler: a `featureGates` field (mapping from string feature name to boolean enabled state) is added to the v1beta1 config file to allow enabling/disabling feature gates.
```

```docs

```

### DIFF
--- a/cmd/kube-scheduler/app/options/options.go
+++ b/cmd/kube-scheduler/app/options/options.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -151,6 +152,8 @@ func (o *Options) Flags() (nfs cliflag.NamedFlagSets) {
 	fs.StringVar(&o.ConfigFile, "config", o.ConfigFile, "The path to the configuration file. Flags override values in this file.")
 	fs.StringVar(&o.WriteConfigTo, "write-config-to", o.WriteConfigTo, "If set, write the configuration values to this file and exit.")
 	fs.StringVar(&o.Master, "master", o.Master, "The address of the Kubernetes API server (overrides any value in kubeconfig)")
+	fs.Var(cliflag.NewMapStringBool(&o.ComponentConfig.FeatureGates), "feature-gates", "A set of key=value pairs that describe feature gates for alpha/experimental features. "+
+		"Options are:\n"+strings.Join(utilfeature.DefaultFeatureGate.KnownFeatures(), "\n"))
 
 	o.SecureServing.AddFlags(nfs.FlagSet("secure serving"))
 	o.CombinedInsecureServing.AddFlags(nfs.FlagSet("insecure serving"))
@@ -159,7 +162,6 @@ func (o *Options) Flags() (nfs cliflag.NamedFlagSets) {
 	o.Deprecated.AddFlags(nfs.FlagSet("deprecated"), &o.ComponentConfig)
 
 	options.BindLeaderElectionFlags(&o.ComponentConfig.LeaderElection, nfs.FlagSet("leader election"))
-	utilfeature.DefaultMutableFeatureGate.AddFlag(nfs.FlagSet("feature gate"))
 	o.Metrics.AddFlags(nfs.FlagSet("metrics"))
 
 	return nfs
@@ -186,6 +188,11 @@ func (o *Options) ApplyTo(c *schedulerappconfig.Config) error {
 			return err
 		}
 
+		// flag feature gates take precedence over feature gates in config file
+		for name, value := range o.ComponentConfig.FeatureGates {
+			cfg.FeatureGates[name] = value
+		}
+
 		// use the loaded config file only, with the exception of --address and --port. This means that
 		// none of the deprecated flags in o.Deprecated are taken into consideration. This is the old
 		// behaviour of the flags we have to keep.
@@ -208,6 +215,11 @@ func (o *Options) ApplyTo(c *schedulerappconfig.Config) error {
 		}
 	}
 	o.Metrics.Apply()
+
+	if err := utilfeature.DefaultMutableFeatureGate.SetFromMap(c.ComponentConfig.FeatureGates); err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/cmd/kube-scheduler/app/options/options.go
+++ b/cmd/kube-scheduler/app/options/options.go
@@ -188,10 +188,7 @@ func (o *Options) ApplyTo(c *schedulerappconfig.Config) error {
 			return err
 		}
 
-		// flag feature gates take precedence over feature gates in config file
-		for name, value := range o.ComponentConfig.FeatureGates {
-			cfg.FeatureGates[name] = value
-		}
+		flagFeatureGates := o.ComponentConfig.FeatureGates
 
 		// use the loaded config file only, with the exception of --address, --port and --feature-gates. This means that
 		// none of the deprecated flags in o.Deprecated are taken into consideration. This is the old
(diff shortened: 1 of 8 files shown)


alculquicondor:
The title of the PR is misleading. The CLI flag for feature gates already exists.

cmd/kube-scheduler/app/options/options_test.go
```diff
@@ -782,3 +787,139 @@ profiles:
...
+			name:                 "by default no feature gates are set",
+			flags:                []string{},
+			expectedFeatureGates: nil,
+		},
+		{
+			name: "feature gates can be set in command-line flags",
+			flags: []string{
+				"--feature-gates", "NonPreemptingPriority=true,PodDisruptionBudget=true",
+			},
+			expectedFeatureGates: map[string]bool{
```
alculquicondor, line 836: nit: s/expectedFeatureGates/wantFeatureGates

cmd/kube-scheduler/app/options/options_test.go
```diff
@@ -782,3 +787,139 @@ profiles:
...
+				"PodOverhead":           true,
+				"PodDisruptionBudget":   true,
+			},
+		},
+		{
+			name: "unknown feature gates cannot be set in command-line flags",
+			flags: []string{
+				"--feature-gates", "NonExistingFeature=true",
+			},
+			expectedError: "unrecognized feature gate: NonExistingFeature",
```
alculquicondor, line 868: nit: s/expectedError/wantErr

alculquicondor (commented):
/approve
from sig scheduling

cmd/kube-scheduler/app/options/options_test.go
pancernik: I'd leave it `expected` as the other tests here use it as well. I prefer consistency. We can change it outside this PR for the whole file later.

cmd/kube-scheduler/app/options/options.go
```diff
@@ -186,6 +188,11 @@ func (o *Options) ApplyTo(c *schedulerappconfig.Config) error {
 			return err
 		}
 
+		// flag feature gates take precedence over feature gates in config file
+		for name, value := range o.ComponentConfig.FeatureGates {
+			cfg.FeatureGates[name] = value
+		}
+
 		// use the loaded config file only, with the exception of --address and --port. This means that
```
liggitt, line 196: with this change, this comment is no longer accurate. what is the use case for letting CLI feature flags override the config file? that makes a config file pretty difficult to reason about.

cmd/kube-scheduler/app/options/options.go
```diff
@@ -208,6 +215,11 @@ func (o *Options) ApplyTo(c *schedulerappconfig.Config) error {
 		}
 	}
 	o.Metrics.Apply()
+
+	if err := utilfeature.DefaultMutableFeatureGate.SetFromMap(c.ComponentConfig.FeatureGates); err != nil {
+		return err
+	}
```
liggitt, line 221: because feature gates can be checked by almost anything, I would set them as early as possible (before SecureServing.ApplyTo, etc)

cmd/kube-scheduler/app/options/options.go
alculquicondor: I don't expect providers to set feature gates both in the config file AND the flag.

But I would expect providers to use a component config file, but setting feature gates only through the flag. This overriding should simplify adoption of component config and it's consistent with what other binaries are doing (such as kube).

Also see https://github.com/kubernetes/kubernetes/issues/84922#issuecomment-584255604

pancernik: This is an agreed status quo at the moment: https://github.com/kubernetes/kubernetes/issues/84922#issuecomment-577269381 

Other components already do it. It would be confusing to cluster administrators if scheduler behaved differently.

liggitt: This has been contentious every time it has come up, and the current state makes for hard-to-understand configurations. We cannot validate a standalone configuration file without knowing the full scheduler invocation. A configuration file author cannot understand the resulting behavior without also having the full invocation.

liggitt: The "config" flag help needs to be changed... it currently says "The path to the configuration file. Flags override values in this file" but CLI flags do not actually override most of the file values

pancernik: One way or other it will always be the case as long as the same options can be set in two different places.

What is a viable alternative here?

liggitt: > We cannot validate a standalone configuration file without knowing the full scheduler invocation

for example, should the flags be merged and set before calling ValidateKubeSchedulerConfiguration?

liggitt: > One way or other it will always be the case as long as the same options can be set in two different places.

If I'm reading the current code correctly, almost all the flags bound to fields that can be specified via config file are ignored when a config file is provided (we load the config file, then override `c.ComponentConfig = *cfg` with the loaded config, erasing all the values were bound to CLI flags

alculquicondor: It should say something along the lines: "flags not marked as deprecated override values in this file"

cmd/kube-scheduler/app/options/options.go
```diff
@@ -186,7 +188,12 @@ func (o *Options) ApplyTo(c *schedulerappconfig.Config) error {
 			return err
 		}
 
-		// use the loaded config file only, with the exception of --address and --port. This means that
+		// flag feature gates take precedence over feature gates in config file
+		for name, value := range o.ComponentConfig.FeatureGates {
+			cfg.FeatureGates[name] = value
+		}
+
+		// use the loaded config file only, with the exception of --address, --port and --feature-gates. This means that
```
alculquicondor, line 196: This block of code refers to `--address` and `--port`. `--feature-gates` is handled above this comment. For readability, you could do `cliFeatureGates := o.ComponentConfig.FeatureGates` above and do the replacement below.

/hold
But let's hold for the discussion on the issue.

cmd/kube-scheduler/app/options/options.go
pancernik: Regarding the validation, one on line 187 checks if the `KubeSchedulerConfiguration` from the loaded config file is correct.

The validation of the `KubeSchedulerConfiguration` constructed using the flags is done before in `server.go`.

Finally, feature gates are validated when setting in `utilfeature.DefaultMutableFeatureGate`. It can and should be done earlier.  We're planning to have a separate PR to add feature flags validation in component-base which would allow to validate the feature gates before setting them and use in `KubeSchedulerConfiguration`.

liggitt: Yeah, it doesn't look like we make use of feature gates in validation currently, which is good (we should avoid introducing that in the future if possible).

We should update the flag help to reflect the actual behavior.

I still think the CLI flag overlay is more confusing than helpful (I haven't seen compelling use cases in which that was better than just setting the desired configuration in the config file), but as long as we clearly document and test the behavior, I won't block it. Should the config file field documentation call out that it can be overridden via a CLI flag?

alculquicondor: What @liggitt is saying is that we can NOT have such validation as long as the piecewise override from CLI flag happens. Such validation would be inconsistent with the validation done in `SetFromMap`

PR #91319, opened 2020-05-21
cloud/node-controller use InstanceMetadataByProviderID

gongguan:
/kind feature

Ref [#90652](https://github.com/kubernetes/kubernetes/issues/90652)

```release-note
Cloud node-controller use InstancesV2
```

Issue #90652, opened 2020-04-30
Reduce the number of calls to GCE API (for clusters on GCE) coming from cloud node-controller


wojtek-t:
For external cloud provider - Kubelet setting the TaintExternalCloudProvider taint: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_node_status.go#L266

The results in calling `initializeNode` function in cloud node-controller:
https://github.com/kubernetes/kubernetes/blob/5c276a451d64d161b21d70042d93f4c4b8e3201d/pkg/controller/cloud/node_controller.go#L323

This (in a healthy case) sends 4 GCE API calls to:
1. get node addresses: https://github.com/kubernetes/kubernetes/blob/5c276a451d64d161b21d70042d93f4c4b8e3201d/pkg/controller/cloud/node_controller.go#L439 (as part of getNodeModifier call)
2. get instance type:
https://github.com/kubernetes/kubernetes/blob/5c276a451d64d161b21d70042d93f4c4b8e3201d/pkg/controller/cloud/node_controller.go#L452 (as part of getNodeModifiers call)
3. get instance:
https://github.com/kubernetes/kubernetes/blob/5c276a451d64d161b21d70042d93f4c4b8e3201d/pkg/controller/cloud/node_controller.go#L233 (as part of updateNodeAddresses call)
4. get node addresses (again):
https://github.com/kubernetes/kubernetes/blob/5c276a451d64d161b21d70042d93f4c4b8e3201d/pkg/controller/cloud/node_controller.go#L242 (as part of updateNodeAddresses call)

All those requests are read requests and could be replaced by a single call.
We could do that by introducing a simple struct like:
```
type InstanceMetadata struct {
  ProviderID ...
  Type ...
  NodeAddresses ...
}
```
and adding a method (or rather replacing the existing InstanceTypeByProviderID which is used only for this purpose) to Instances cloud-provider interface;
```
func InstanceMetadataByProviderID(...) (*InstanceMetadata, error)
```

With that change, we would be able to replace all that 4 calls with just a single API call.

This would be useful especially for creating large clusters (like 5k nodes), which may saturate the quota for API calls for quite some time.

This is quite significant size of refactoring, though it should be relatively mechanical task to do.

/sig cloud-provider
/sig scalability
/priority important-longterm

/help-wanted

@mborsz @mm4tt - FYI

tedyu:
[...]

PR #92367, opened 2020-06-22
define and implement cloud InstanceV2

gongguan:
/kind feature

Ref #90652

Part of #91319

```release-note
None
```

### DIFF
--- a/staging/src/k8s.io/cloud-provider/cloud.go
+++ b/staging/src/k8s.io/cloud-provider/cloud.go
@@ -49,6 +49,9 @@ type Interface interface {
 	LoadBalancer() (LoadBalancer, bool)
 	// Instances returns an instances interface. Also returns true if the interface is supported, false otherwise.
 	Instances() (Instances, bool)
+	// InstancesV2 is an implementation for instances only used by cloud node-controller now.
+	// Also returns true if the interface is supported, false otherwise.
+	InstancesV2() (InstancesV2, bool)
 	// Zones returns a zones interface. Also returns true if the interface is supported, false otherwise.
 	Zones() (Zones, bool)
 	// Clusters returns a clusters interface.  Also returns true if the interface is supported, false otherwise.
@@ -184,6 +187,17 @@ type Instances interface {
 	InstanceExistsByProviderID(ctx context.Context, providerID string) (bool, error)
 	// InstanceShutdownByProviderID returns true if the instance is shutdown in cloudprovider
 	InstanceShutdownByProviderID(ctx context.Context, providerID string) (bool, error)
+}
+
+// InstancesV2 is an abstract, pluggable interface for sets of instances.
+// Unlike Instances, it is only used by cloud node-controller now.
+type InstancesV2 interface {
+	// InstanceExistsByProviderID returns true if the instance for the given provider exists.
+	InstanceExistsByProviderID(ctx context.Context, providerID string) (bool, error)
+	// InstanceShutdownByProviderID returns true if the instance is shutdown in cloudprovider.
+	InstanceShutdownByProviderID(ctx context.Context, providerID string) (bool, error)
+	// InstanceMetadataByProviderID returns the instance's metadata.
+	InstanceMetadataByProviderID(ctx context.Context, providerID string) (*InstanceMetadata, error)
 }
 
 // Route is a representation of an advanced routing rule.
--- a/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
+++ b/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
@@ -1394,6 +1394,11 @@ func (c *Cloud) Instances() (cloudprovider.Instances, bool) {
 	return c, true
 }
 
+// InstancesV2 returns an implementation of InstancesV2 for Amazon Web Services.
+func (c *Cloud) InstancesV2() (cloudprovider.InstancesV2, bool) {
+	return nil, false
+}
+
 // Zones returns an implementation of Zones for Amazon Web Services.
 func (c *Cloud) Zones() (cloudprovider.Zones, bool) {
 	return c, true
--- a/staging/src/k8s.io/legacy-cloud-providers/azure/azure.go
+++ b/staging/src/k8s.io/legacy-cloud-providers/azure/azure.go
@@ -659,6 +659,11 @@ func (az *Cloud) Instances() (cloudprovider.Instances, bool) {
 	return az, true
 }
 
+// InstancesV2 returns an instancesV2 interface. Also returns true if the interface is supported, false otherwise.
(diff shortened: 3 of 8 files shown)


staging/src/k8s.io/cloud-provider/cloud.go
```diff
@@ -49,6 +49,8 @@ type Interface interface {
 	LoadBalancer() (LoadBalancer, bool)
 	// Instances returns an instances interface. Also returns true if the interface is supported, false otherwise.
 	Instances() (Instances, bool)
+	// InstancesV2 is an external-only implementation for instances. Also returns true if the interface is supported, false otherwise.
```
wojtek-t, line 52: What do you mean by "external-only" ?

staging/src/k8s.io/cloud-provider/cloud.go
```diff
@@ -186,6 +188,18 @@ type Instances interface {
 	InstanceShutdownByProviderID(ctx context.Context, providerID string) (bool, error)
 }
 
+// InstancesV2 is an abstract, pluggable interface for sets of instances.
+// Unlike Instances, it is a external-only implementation which should not be called the components like
```
wojtek-t, line 192: I think I don't understand this sentence.

wojtek-t: @andrewsykim - to help figuring out the best wording here

staging/src/k8s.io/cloud-provider/cloud.go
gongguan: Might it be internal-only? I'd like to ask for @andrewsykim 's original thought.

andrewsykim: Sorry, the comment in the example I shared was not fully thought out. 

"external-only" meant it should only be called by cloud-controller-manager at the moment. Cloud node controller in https://github.com/kubernetes/kubernetes/pull/91319 only runs in cloud-controller-manager and not in kube-controller-manager or kubelet.

gongguan: No sorry, your example helps me a lot to understand your thought. Updated the comments, PTAL

wojtek-t:
/approve
/retest

@andrewsykim - does that look ok to you too?

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
```diff
@@ -1394,6 +1394,11 @@ func (c *Cloud) Instances() (cloudprovider.Instances, bool) {
 	return c, true
 }
 
+// InstancesV2 returns an implementation of InstancesV2 for Amazon Web Services.
+func (c *Cloud) InstancesV2() (cloudprovider.InstancesV2, bool) {
+	return c, true
```
andrewsykim, line 1399: set this to false and let aws OWNERs flip this on?

staging/src/k8s.io/legacy-cloud-providers/azure/azure.go
```diff
@@ -659,6 +659,11 @@ func (az *Cloud) Instances() (cloudprovider.Instances, bool) {
 	return az, true
 }
 
+// InstancesV2 returns an instancesV2 interface. Also returns true if the interface is supported, false otherwise.
+func (az *Cloud) InstancesV2() (cloudprovider.InstancesV2, bool) {
+	return az, true
```
andrewsykim, line 664: set this to false and let Azure OWNERs flip this on?

staging/src/k8s.io/legacy-cloud-providers/gce/gce.go
```diff
@@ -659,6 +659,11 @@ func (g *Cloud) Instances() (cloudprovider.Instances, bool) {
 	return g, true
 }
 
+// InstancesV2 returns an implementation of InstancesV2 for Google Compute Engine.
+func (g *Cloud) InstancesV2() (cloudprovider.InstancesV2, bool) {
+	return g, true
```
andrewsykim, line 664: set this to false and let GCE OWNERs flip this on?

staging/src/k8s.io/legacy-cloud-providers/openstack/openstack_instances.go
```diff
@@ -62,6 +61,16 @@ func (os *OpenStack) Instances() (cloudprovider.Instances, bool) {
...
 }
 
+// Instances returns an implementation of Instances for OpenStack.
+func (os *OpenStack) Instances() (cloudprovider.Instances, bool) {
+	return os.instances()
+}
+
+// InstancesV2 returns an implementation of InstancesV2 for OpenStack.
+func (os *OpenStack) InstancesV2() (cloudprovider.InstancesV2, bool) {
+	return os.instances()
```
andrewsykim, line 71: Ideally this shouldn't call the same thing as Instances

staging/src/k8s.io/legacy-cloud-providers/vsphere/vsphere.go
```diff
@@ -560,6 +560,11 @@ func (vs *VSphere) Instances() (cloudprovider.Instances, bool) {
 	return vs, true
 }
 
+// InstancesV2 returns an implementation of InstancesV2 for vSphere.
+func (vs *VSphere) InstancesV2() (cloudprovider.InstancesV2, bool) {
+	return vs, true
```
andrewsykim, line 565: set this to `false` and let Azure OWNERs flip this on?

andrewsykim (commented):
Overall lgtm, it might be better to leave InstancesV2 off for the in-tree providers and let the individual providers turn it on.

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
gongguan: All set as false, it's a method with low risk.

staging/src/k8s.io/legacy-cloud-providers/gce/gce.go
gongguan: Set as false, while I thought @wojtek-t might want gce use InstancesV2() to reduce api calls, right?

wojtek-t: We can do that when we will be migrating the controller. So I'm fine with leaving false for now.

andrewsykim (commented):
/approve
/lgtm

Thanks for iterating on this @gongguan!

staging/src/k8s.io/legacy-cloud-providers/vsphere/vsphere.go
feiskyer: nice for this, would add InstancesV2 support for Azure.

wojtek-t: It's already added - all built-in providers have this already implemented. It's just a matter of changing this value here.

PR #92372, opened 2020-06-22
ignore "vmdk not found" vsphere errors during unmount (assume success)

breunigs:
**What type of PR is this?**
/kind bug
/sig storage

A vsphere deletion call may succeed before Kubernetes learns about its result. This can happen when the kube-controller-manager loses a leader election or crashes. When the deletion is retried, a "file not found" error is returned. Future retries will get the same error and manual clean up is required. This PR checks for the not found error and treats it as success so that the volume can be cleaned up from Kubernetes side as well. Once applied, this also automatically cleans up volumes that are in "State: Failed" due to this.

```release-note
Treat VSphere "File (vmdk path here) was not found" errors as success during volume deletion
```

```docs
Error classes returned by the API call to delete virtual disks are detailed here:
https://code.vmware.com/apis/196/vsphere/doc/vim.VirtualDiskManager.html#deleteVirtualDisk
`FileNotFound` is a subclass of `FileFault`, which are used for issues specific
to this file or disk. General datastore issues return `InvalidDatastore` errors.
```

### DIFF
--- a/staging/src/k8s.io/legacy-cloud-providers/vsphere/vclib/diskmanagers/vdm.go
+++ b/staging/src/k8s.io/legacy-cloud-providers/vsphere/vclib/diskmanagers/vdm.go
@@ -97,7 +97,7 @@ func (diskManager virtualDiskManager) Delete(ctx context.Context, datacenter *vc
 	}
 	err = task.Wait(ctx)
 	vclib.RecordvSphereMetric(vclib.APIDeleteVolume, requestTime, err)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


andrewsykim:
/assign
/triage accepted

cc @divyenpatel @SandeepPissay

divyenpatel (approved):
/approve

breunigs:
Failing test seems to be flaky: https://github.com/kubernetes/kubernetes/issues/101894

/retest

pacoxu:
/lgtm
The change seems valid to me.

Issue #91236, opened 2020-05-19
 Flaky test: [sig-network] Services should be able to preserve UDP traffic when server pod cycles for a NodePort service


jprzychodzen:
**Which jobs are flaking**:
ci-kubernetes-e2e-gce-scale-correctness

**Which test(s) are flaking**:
[sig-network] Services should be able to preserve UDP traffic when server pod cycles for a NodePort service

**Testgrid link**:
https://k8s-testgrid.appspot.com/sig-release-master-informing#gce-master-scale-correctness

**Reason for failure**:
```
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/network/service.go:1243
May 18 12:22:32.868: pod communication failed
Unexpected error:
    <*errors.errorString | 0xc003faab60>: {
        s: "continuous echo was not able to communicate with initial server pod",
    }
    continuous echo was not able to communicate with initial server pod
occurred
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/network/service.go:1302
```

/sig network
/assign @thockin

athenabot:
/triage unresolved

Comment `/remove-triage unresolved` when the issue is assessed and confirmed.

🤖 I am a bot run by vllry. 👩‍🔬

haircommander:
CRI-O CI is also hitting this failure. We have been unable to reproduce locally. our CI runs on aws

aojea:
hmm, I only see 1 failure out of 4 and seems that the connectivity didn't work at all 🤔 

```
May 18 12:22:16.516: INFO: The status of Pod server-1 is Running (Ready = true)
ï¿½[1mSTEPï¿½[0m: waiting up to 3m0s for service clusterip-test in namespace services-9934 to expose endpoints map[server-1:[80]]
May 18 12:22:16.585: INFO: successfully validated that service clusterip-test in namespace services-9934 exposes endpoints map[server-1:[80]] (68.878243ms elapsed)
ï¿½[1mSTEPï¿½[0m: Sending UDP traffic to NodePort service clusterip-test on node with publicIP 34.75.29.74
ï¿½[1mSTEPï¿½[0m: creating a second pod for the service clusterip-test
May 18 12:22:16.656: INFO: The status of Pod server-2 is Pending, waiting for it to be Running (with Ready = true)
[...]

PR #92375, opened 2020-06-22
Attempt to delfake further conntract e2e flakes

wojtek-t:
Ref https://github.com/kubernetes/kubernetes/issues/91236

All other failures seem to be kind-related.

```release-note
NONE
```

/sig network
/kind flake

### DIFF
--- a/test/e2e/network/service.go
+++ b/test/e2e/network/service.go
@@ -3569,40 +3569,42 @@ func translatePodNameToUID(c clientset.Interface, ns string, expectedEndpoints p
 // validateEndpointsPorts validates that the given service exists and is served by the given expectedEndpoints.
 func validateEndpointsPorts(c clientset.Interface, namespace, serviceName string, expectedEndpoints portsByPodName) error {
 	ginkgo.By(fmt.Sprintf("waiting up to %v for service %s in namespace %s to expose endpoints %v", framework.ServiceStartTimeout, serviceName, namespace, expectedEndpoints))
-	i := 1
-	for start := time.Now(); time.Since(start) < framework.ServiceStartTimeout; time.Sleep(1 * time.Second) {
+	expectedPortsByPodUID, err := translatePodNameToUID(c, namespace, expectedEndpoints)
+	if err != nil {
+		return err
+	}
+
+	i := 0
+	if pollErr := wait.PollImmediate(time.Second, framework.ServiceStartTimeout, func() (bool, error) {
 		ep, err := c.CoreV1().Endpoints(namespace).Get(context.TODO(), serviceName, metav1.GetOptions{})
 		if err != nil {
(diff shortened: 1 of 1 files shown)


test/e2e/network/service.go
```diff
@@ -3569,40 +3569,43 @@ func translatePodNameToUID(c clientset.Interface, ns string, expectedEndpoints p
...
-				return err
-			}
-			framework.Logf("successfully validated that service %s in namespace %s exposes endpoints %v (%v elapsed)",
-				serviceName, namespace, expectedEndpoints, time.Since(start))
-			return nil
-		}
+		i++
 		if i%5 == 0 {
-			framework.Logf("Unexpected endpoints: found %v, expected %v (%v elapsed, will retry)", portsByPodUID, expectedEndpoints, time.Since(start))
+			framework.Logf("Unexpected endpoints: found %v, expected %v, will retry", portsByPodUID, expectedEndpoints)
```
aojea, line 3588: seems we only should go here if `if len(portsByPodUID) != len(expectedEndpoints)` , you removed that check.
However, I don't know if it is useful having a counter to rate limit the error message.
```suggestion
                if i%5 == 0 &&  len(portsByPodUID) != len(expectedEndpoints) {
			framework.Logf("Unexpected endpoints: found %v, expected %v, will retry", portsByPodUID, expectedEndpoints)
			return false, nil
			}
```

wojtek-t: I removed that check, because it's already part of validatePorts. 

Moved the og to validatePorts branch.

test/e2e/network/service.go
```diff
@@ -3569,40 +3569,43 @@ func translatePodNameToUID(c clientset.Interface, ns string, expectedEndpoints p
...
+		return fmt.Errorf("error waithing for service %s in namespace %s to expose endpoints %v", serviceName, namespace, expectedEndpoints)
 	}
+
 	if pods, err := c.CoreV1().Pods(metav1.NamespaceAll).List(context.TODO(), metav1.ListOptions{}); err == nil {
 		for _, pod := range pods.Items {
 			framework.Logf("Pod %s\t%s\t%s\t%s", pod.Namespace, pod.Name, pod.Spec.NodeName, pod.DeletionTimestamp)
 		}
 	} else {
 		framework.Logf("Can't list pod debug info: %v", err)
 	}
```
aojea, line 3607: this block seems to be used only for debugging, hence it should go inside the `err != nil` block that checks if ` wait.PollImmediate` failed

wojtek-t (commented):
@aojea - thanks; PTAL

test/e2e/network/service.go
wojtek-t: done

aojea:
will check the log results to try to diff against previous runs,
but it lgtm

PR #85689, opened 2019-11-27
Export scheduler.Snapshot function

ahg-g:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

CA needs to update the snapshot now that predicates list nodes/pods from the snapshot rather than the scheduler cache. 

For CA to be able to do that, we need to export scheduler.snapshot function. 

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

/cc @alculquicondor

PR #92376, opened 2020-06-22
Cherry pick of #85689: Export scheduler.Snapshot function

towca:
This was added as a workaround for Cluster Autoscaler, but didn't make the cut for release-1.17. Now we want to sync other cherry-pick fixes from release-1.17, but we can't do that without this commit on the branch.

/kind cleanup

### DIFF
--- a/pkg/scheduler/core/generic_scheduler.go
+++ b/pkg/scheduler/core/generic_scheduler.go
@@ -132,8 +132,12 @@ type ScheduleAlgorithm interface {
 	Extenders() []algorithm.SchedulerExtender
 	// GetPredicateMetadataProducer returns the predicate metadata producer. This is needed
 	// for cluster autoscaler integration.
-	// TODO(ahg-g): remove this once CA migrates to creating a Framework instead of a full scheduler.
+	// TODO(#85691): remove this once CA migrates to creating a Framework instead of a full scheduler.
 	PredicateMetadataProducer() predicates.MetadataProducer
+	// Snapshot snapshots scheduler cache and node infos. This is needed
+	// for cluster autoscaler integration.
+	// TODO(#85691): remove this once CA migrates to creating a Framework instead of a full scheduler.
+	Snapshot() error
 }
 
 // ScheduleResult represents the result of one pod scheduled. It will contain
@@ -169,7 +173,7 @@ type genericScheduler struct {
 
 // snapshot snapshots scheduler cache and node infos for all fit and priority
 // functions.
-func (g *genericScheduler) snapshot() error {
+func (g *genericScheduler) Snapshot() error {
 	// Used for all fit and priority funcs.
 	return g.cache.UpdateNodeInfoSnapshot(g.nodeInfoSnapshot)
 }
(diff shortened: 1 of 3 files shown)


MaciekPytel:
Some context: this is a prerequisite to fixing https://github.com/kubernetes/autoscaler/issues/3224. This is a regression introduced when changes to init container handling in scheduler were cherry-picked into release-1.17. As such I think this is higher priority:

/priority crititical-urgent
/kind bug

/lgtm

ahg-g:
I assume CA doesn't need this for 1.18, correct?

towca:
Correct.

MaciekPytel:
@kubernetes/release-managers Can you approve the cherry-pick? 

Note: this is needed for Cluster Autoscaler, which has a separate release schedule. We just need this commit to be on release-1.17 branch to be able to launch a patched version of autoscaler. Given the above would it be possible to merge it before the next scheduled patch release of Kubernetes?

MaciekPytel:
@kubernetes/release-managers Can you help us merge this? It's blocking patch release of Cluster Autoscaler .

tpepper:
While marked as a "cleanup" here, this was marked as a 1.18 feature in https://github.com/kubernetes/kubernetes/pull/85689.  Generally we don't backport features.  The PR and commit message though are short in explaining or justifying the interdependency, so maybe I'm missing something.  Should 1.17 kubernetes need to depend on this patch if relative to a CA 1.17 branch?  Or are you trying to keep a common codebase for CA without branches, don't have runtime feature gates, ...?

towca:
Marking this as "cleanup" was my mistake, but I'd say it's more of a bug - in a sense that this really should've been cherry picked right after it was merged.

Cluster Autoscaler tightly depends on scheduler - we have to import scheduler code in the correct version for each release branch, while maintaining feature parity between branches. This is a huge chore and we have scripts automating it.

For the 1.17 branch, Cluster Autoscaler needs this commit in scheduler code in order to build. We messed up when 1.17 was cut, and it didn't make it in time. Due to how the scripts are written, we're syncing vendor dependencies to a particular ref in kubernetes. So what we did was just sync to that needed commit - a bit past 1.17 branch tip, but with no other delta in scheduler code - it worked fine as a workaround. However, this means that if something was later cherry-picked on the 1.17 branch in scheduler, our scripts wouldn't pick it up, because they were fixed on that one commit and not on the 1.17 branch itself.

Since there were no meaningful cherry-picks to the scheduler code on 1.17 branch, everything was sort-of working. However, now there's been a cherry-pick with a change in scheduler behavior and we need to sync it to our dependencies in Cluster Autoscaler: kubernetes/autoscaler#3224. Our scripts can't handle that if the initial missing scheduler commit is not on the 1.17 branch as well. We could of course do the syncing manually, but this would likely break the automated process in the future, resulting in us having to sync this one branch semi-manually for a long time. This would add a lot of toil to what already is an arduous process.

MaciekPytel:
To clarify - "manual syncing" mentioned in comment above would really have to boil down to Cluster Autoscaler importing kubernetes code and than someone manually modifying that code in vendor/ directory. I don't think we should do that under any circumstances.

And to provide some more context on the commit being cherry-picked: as part of feature work for 1.17 scheduler code was refactored. A side effect of this refactor was making Snapshot() method private, as it was not used by any other code in kubernetes/kubernetes. It is, however, vital for Cluster Autoscaler. All this commit does was make this method public again, so it can be called by autoscaler.

PR #92379, opened 2020-06-22
fix pod status ready condition

### LINKED ISSUE #92378: Pod status ready condition is incorrect
**What happened**:
- kubelet pod status cache ready condition is `true` and node system hangs
- controller-manager update pod status ready condition to `false` because node `NotReady`
- node system recovery and kubelet update node status to `Ready`
- but kubelet doesn't update pod status ready condition to `true`, because cache is `true`

kubelet update pod status ready condition to `true` after node recovery

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

liuxu623:
**What type of PR is this?**
/kind bug

We should update pod status when cache is expired 

Fixes #92378

```release-note
none
```

```docs

```

### DIFF
--- a/pkg/kubelet/status/status_manager.go
+++ b/pkg/kubelet/status/status_manager.go
@@ -388,10 +388,7 @@ func checkContainerStateTransition(oldStatuses, newStatuses []v1.ContainerStatus
 // This method IS NOT THREAD SAFE and must be called from a locked function.
 func (m *manager) updateStatusInternal(pod *v1.Pod, status v1.PodStatus, forceUpdate bool) bool {
 	var oldStatus v1.PodStatus
-	cachedStatus, isCached := m.podStatuses[pod.UID]
-	if isCached {
-		oldStatus = cachedStatus.status
-	} else if mirrorPod, ok := m.podManager.GetMirrorPodByPod(pod); ok {
+	if mirrorPod, ok := m.podManager.GetMirrorPodByPod(pod); ok {
 		oldStatus = mirrorPod.Status
 	} else {
 		oldStatus = pod.Status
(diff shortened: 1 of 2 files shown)


tedyu:
Can you clarify the problem this PR solves (via unit / e2e test) ?

thanks

liuxu623:
> Can you clarify the problem this PR solves (via unit / e2e test) ?
> 
> thanks

I have add some test, thank you for help me review code.

pkg/kubelet/status/status_manager.go
```diff
@@ -431,7 +431,7 @@ func (m *manager) updateStatusInternal(pod *v1.Pod, status v1.PodStatus, forceUp
 	normalizeStatus(pod, &status)
 	// The intent here is to prevent concurrent updates to a pod's status from
 	// clobbering each other so the phase of a pod progresses monotonically.
-	if isCached && isPodStatusByKubeletEqual(&cachedStatus.status, &status) && !forceUpdate {
+	if isCached && isPodStatusByKubeletEqual(&pod.Status, &status) && isPodStatusByKubeletEqual(&cachedStatus.status, &status) && !forceUpdate {
```
SergeyKanzhelev, line 434: is it ok that `updateLastTransitionTime ` will not update the transition time in this case?

SergeyKanzhelev:
@liuxu623 this seems to be the right direction to fix the bug. Would you please take a look at the comment - the change may be a bit bigger to also include updating the last transition time variable.

liuxu623:
@SergeyKanzhelev Hi, I have update commit, please help me review, thank you.

PR #92381, opened 2020-06-22
[WIP]kubeadm will not set --allow-privileged=true in default apiserver manifest

brianpursley:
**What type of PR is this?**
/kind feature

According to the documentation, [the default value for `--allow-privileged` flag of apiserver is false](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/).

But when you use kubeadm to initialize your cluster, it explicitly sets `--allow-privileged=true` in the generated kube-apiserver.yaml.

People should check and be aware of the settings in the configs of their own cluster, but this is a little surprising given that the documented default is false.

I am proposing with this PR that kubeadm should choose the more secure default (false) and require people to explicitly enable `--allow-privileged` on their cluster if they want it.

n/a

```release-note
New clusters initialized using kubeadm no longer have `--allow-privileged=true` in the generated kube-apiserver.yaml.
```

```docs

```


brianpursley:
It seems that changing this breaks kind and some tests that depend on this being set to true.

I'm going to close this PR and turn this into an issue.

PR #92385, opened 2020-06-22
exit probe check during pod deletion

### LINKED ISSUE #52817: liveness/readiness probe is executed and failed while pod is terminated
**What happened**:
liveness/readiness probe fails while pod is terminated. Also it happened only once during the pod termination. The issue started happening after upgrading version  to v1.7 from v1.6.X 

execute `kubectl delete pod nginx-A1` to delete pod, so status of the nginx-podA1 is changed to `Terminating`, right after that it seems Liveness and Readiness Probe is executed and failed, but only once.
Nginx reverse proxy is running in the pod. so I just use `httpGet`method for liveness and readiness

Here is my Deployment config.
``` yaml
   ...
   spec:
      terminationGracePeriodSeconds: 60
        ...
        livenessProbe:
          httpGet:
            path: /healthz
            port: 8080
          initialDelaySeconds: 15
          timeoutSeconds: 3
        readinessProbe:
          httpGet:
            path: /healthz
            port: 8080
          initialDelaySeconds: 15
          timeoutSeconds: 3
```

Here is Events log by `kubectl describe pod nginx-A1`
```
Events:
  FirstSeen	LastSeen	Count	From					SubObjectPath			Type		Reason		Message
  ---------	--------	-----	----					-------------			--------	------		-------
  14s		14s		1	kubelet, *****	spec.containers{dnsmasq}	Normal		Killing		Killing container with id docker://dnsmasq:Need to kill Pod
  9s		9s		1	kubelet, *****	spec.containers{nginx}		Warning		Unhealthy	Liveness probe failed: Get http://100.*.*.*:8080/healthz: dial tcp 100.*.*.*:8080: getsockopt: connection refused
  9s		9s		1	kubelet, *****	spec.containers{nginx}		Warning		Unhealthy	Readiness probe failed: Get http://100.*.*.*:8080/healthz: dial tcp 100.*.*.*:8080: getsockopt: connection refused

```

**Environment**:
- Kubernetes version: 1.7.2

ashleyschuett:
**What type of PR is this?**

/kind bug

If a pod takes longer to delete than it's its health/readiness probe allow a user can get errors that the pods check is erroring even though the pod is being deleted. This can update the pods status field to be inaccurate. 

Fixes #52817

```release-note
none
```

```docs
none
```

### DIFF
--- a/pkg/kubelet/prober/worker.go
+++ b/pkg/kubelet/prober/worker.go
@@ -188,6 +188,13 @@ func (w *worker) doProbe() (keepGoing bool) {
 		return false
 	}
 
+	// We do not need to check status during deletion
+	if w.pod.ObjectMeta.DeletionTimestamp != nil {
+		klog.V(3).Infof("Pod %v %v is being deleted, exiting probe worker",
+			format.Pod(w.pod), status.Phase)
+		return false
+	}
+
 	c, ok := podutil.GetContainerStatus(status.ContainerStatuses, w.container.Name)
 	if !ok || len(c.ContainerID) == 0 {
 		// Either the container has not been created yet, or it was deleted.
--- a/pkg/kubelet/prober/worker_test.go
+++ b/pkg/kubelet/prober/worker_test.go
@@ -58,17 +58,22 @@ func TestDoProbe(t *testing.T) {
 		failedStatus.Phase = v1.PodFailed
 
 		tests := []struct {
-			probe          v1.Probe
-			podStatus      *v1.PodStatus
-			expectContinue bool
-			expectSet      bool
-			expectedResult results.Result
+			probe                v1.Probe
+			podStatus            *v1.PodStatus
+			setDeletionTimestamp bool
+			expectContinue       bool
+			expectSet            bool
+			expectedResult       results.Result
 		}{
 			{ // No status.
 				expectContinue: true,
 			},
 			{ // Pod failed
 				podStatus: &failedStatus,
+			},
+			{ // Pod deletion
+				podStatus:            &runningStatus,
+				setDeletionTimestamp: true,
 			},
 			{ // No container status
 				podStatus:      &otherStatus,
@@ -106,6 +111,10 @@ func TestDoProbe(t *testing.T) {
 			w := newTestWorker(m, probeType, test.probe)
 			if test.podStatus != nil {
 				m.statusManager.SetPodStatus(w.pod, *test.podStatus)
+			}
+			if test.setDeletionTimestamp {
+				now := metav1.Now()
+				w.pod.ObjectMeta.DeletionTimestamp = &now
 			}
 			if c := w.doProbe(); c != test.expectContinue {
 				t.Errorf("[%s-%d] Expected continue to be %v but got %v", probeType, i, test.expectContinue, c)


tedyu:
Is it possible to add a test ?

ashleyschuett:
@tedyu sorry about that. I have now added a test case.

pkg/kubelet/prober/worker.go
```diff
@@ -188,6 +188,13 @@ func (w *worker) doProbe() (keepGoing bool) {
 		return false
 	}
 
+	// We do not need to check status during deletion
+	if w.pod.ObjectMeta.DeletionTimestamp != nil {
```
SergeyKanzhelev, line 192: is it guaranteed that will the traffic already be stopped to this pod when the `DeletionTimestamp ` was set? if not, would it be actually helpful to actually run the readyness probe and take the pod from rotation for the duration of pod termination?

pkg/kubelet/prober/worker_test.go
```diff
@@ -58,18 +58,23 @@ func TestDoProbe(t *testing.T) {
...
+			expectSet            bool
+			expectedResult       results.Result
 		}{
 			{ // No status.
 				expectContinue: true,
 			},
 			{ // Pod failed
 				podStatus: &failedStatus,
 			},
+			{ // Pod deletion
```
SergeyKanzhelev, line 74: I'm curious, why in terminated state test set looks like this:

```
{ // Container terminated
				podStatus:      &terminatedStatus,					podStatus:      &terminatedStatus,
				expectSet:      true,					expectSet:      true,
				expectedResult: results.Failure,					expectedResult: results.Failure,
			},				},
```

Should the terminated state be treated the same as being terminated?

pkg/kubelet/prober/worker.go
matthyx: This is exactly what will happen once `return false` is executed 3 lines below.

SergeyKanzhelev: My question is what mechanism will remove traffic (not the probes calls, regular user traffic) to this pod? Is it done just by `kubectl delete` somehow before setting this `DeletionTimestamp`?

Looking below I see that in case of terminated container we also do this:

```
		if !w.containerID.IsEmpty() {
			w.resultsManager.Set(w.containerID, results.Failure, w.pod)
		}
```

Same comment below for the unit tests. The question is whether ready prob status needs to be changed while pod is being deleted.

ashleyschuett:
pull-kubernetes-e2e-gce - failed test included in latest flakes 
   - [sig-network] EndpointSlice should create Endpoints and EndpointSlices for Pods matching a Service
   - included http://storage.googleapis.com/k8s-metrics/flakes-latest.json

pull-kubernetes-verify - panic while running?
  - `k8s.io/kubernetes/test/e2e/network test/e2e/network/kube_proxy.go:180:4: Failf call has error-wrapping directive %w`
  - looks like it needs to be reran?

pull-kubernetes-integration - seems to have started failing consistently 
  - https://storage.googleapis.com/k8s-gubernator/triage/index.html?test=run_kubectl_request_timeout_tests 

They all look to be flakes?

pigletfly (approved):

ashleyschuett:
what needs to be done here for this to be approved?

matthyx:
> what needs to be done here for this to be approved?

try to ping potential approvers on slack... this is the sad story of sig-node, too many PRs, not enough approvers

dashpole:
I think @SergeyKanzhelev's question above is relevant.  When do services stop sending traffic to pods?  If services stop sending after the deletion timestamp is set, then this change is useful.  If not, then the readiness probe can prevent traffic from going to the terminating pod.

ashleyschuett:
It appears that services stop sending trafic to pods when the deletiontimestap is set, unless the pod has specified that traffic should continue to be sent. https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/endpoint/endpoints_controller.go#L425-L428 

I can update this to check for the annotation / spec field here as well and only return false if both conditions are met? Does that make sense?

matthyx:
I would say yes. Let's do it properly. Thanks

ashleyschuett:
After looking at this closer it seems that the annotation/spec to continue sending sending traffic after pod deletion is only set on the service and not the pod level :/

pkg/kubelet/prober/worker.go
ashleyschuett: Hmm I think the readiness probe failing during deletion does makes sense since it is possible for a service to continue to send traffic to the pod even after deletion. https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/endpoint/endpoints_controller.go#L425-L428 which I was unaware of. 

So maybe this does not fix the bug that I am experiencing, but maybe there's another solution? 

What my system is experiencing is when a pod takes a bit of time to delete it is in the state running, but while clean its self up the readiness probe starts failing (this pod is not behind a service) once the readiness probe fails however it changes the `Phase` of the pod from `Running` to `Pending` even though it has a `deletionTimestamp` set.

My first thought was the readiness probe should not fail since it's being deleted, but that seems to not be the case, however changing the status from `Running` -> `Pending` also does not feel correct.

matthyx:
Ok, so your pr is correct :)

pkg/kubelet/prober/worker.go
dashpole: Moving from Running -> pending is definitely a bug.

SergeyKanzhelev:
@matthyx would it be more correct to mark pod as not ready in this case? Disabling ready probes gives no option for developer to indicate that pod should not accept traffic any longer and it may be unexpected behavior change

pkg/kubelet/prober/worker.go
matthyx: I think `Running` and `Pending` are set by the container runtime... nothing to do with probes.

matthyx:
> @matthyx would it be more correct to mark pod as not ready in this case? Disabling ready probes gives no option for developer to indicate that pod should not accept traffic any longer and it may be unexpected behavior change

I think the code in that PR is correct, we should fail the readiness probe of the container if the pod is being terminated as an optimization. But the rest of the handling should be performed by the same routines that deal with real probe failures, taking into account the `failureThreshold` and other user preferences for instance.

I think Ashley's problem is that without a service acting as the load balancer, nothing prevents the pod from accepting traffic during its shutdown... @ashleyschuett what's the reason for not using one in your case?
There are probably other scenarios where you would want to exclude a pod from the traffic by manipulating tags, and this as well is not covered if you go without a service...

SergeyKanzhelev:
@ashleyschuett can you confirm if this fix actually fixes the problem you experience?

@matthyx I'm not sure I understand your reasoning. My understanding is that this code disables probe checks without setting pod into non-ready state. If the idea is to stop traffic to the pod, setting the state to failure (same as in terminated pod test case) may be a better idea. I do realize we are talking about edge cases when shut down takes a while on the container. I hope that nobody is taking dependency on the currently implemented behavior =).

matthyx:
> If the idea is to stop traffic to the pod, setting the state to failure (same as in terminated pod test case) may be a better idea.

Where and how would you do it?

SergeyKanzhelev:
> > If the idea is to stop traffic to the pod, setting the state to failure (same as in terminated pod test case) may be a better idea.
> 
> Where and how would you do it?

Sorry for delay. By setting Failure result for the readiness probe worker in case `w.pod.ObjectMeta.DeletionTimestamp != nil`.

SergeyKanzhelev:
> @ashleyschuett can you confirm if this fix actually fixes the problem you experience?

Any update? Did you figure out whether `Running -> pending ` transition is happening because of the probe results?

ashleyschuett:
I am unable to reproduce my issue with the `Running -> Pending` during deletion, so I'm unfortunately not sure what caused it :/

matthyx:
Just thinking out loud here... would it make sense to introduce a `shutdownProbe` ? Or maybe some mechanism to make sure we don't kill the container until all the processing is done, at least something better than the grace period we currently have...
@thockin @SergeyKanzhelev what do you think?

pkg/kubelet/prober/worker.go
sjenning: Running -> Pending should be fixed by https://github.com/kubernetes/kubernetes/pull/95364 and https://github.com/kubernetes/kubernetes/pull/95561

SergeyKanzhelev:
Based on @sjenning's [comment](https://github.com/kubernetes/kubernetes/pull/92385#discussion_r508958768) the issue of Running -> Pending is resolved in different PR. 

This PR stops probes execution, but do not address the question on what the actual behavior should be. Should pod be marked not ready or not? I like the idea from @matthyx to resolve this via a termination hook enhancement. As minimum a better explanation on what is expected behavior of readiness and liveness probes must be when termination started - especially from the load balancer perspective in case of a very long termination.

Closing PR for now, please re-open if plan to move it forward.

/close

PR #91939, opened 2020-06-09
CSIStorageCapacity API

pohly:
**What type of PR is this?**
/kind api-change

**What this PR does / why we need it**:

This is the API change for https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1472-storage-capacity-tracking

**Which issue(s) this PR fixes**:

Related-to: https://github.com/kubernetes/enhancements/issues/1472

**Special notes for your reviewer**:

This is the just the API  change. The changes to the volume binder which consume the new storage capacity information will follow separately.

**Does this PR introduce a user-facing change?**:
```release-note
Introduce API for exposing storage capacity information.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

Will be added together with the implementation.

PR #92387, opened 2020-06-22
CSI storage capacity check

pohly:
**What type of PR is this?**
/kind feature

This is the part of https://github.com/kubernetes/enhancements/issues/1472 where the scheduler uses the information for improving pod scheduling of pods with unbound volumes.

The API change itself is getting reviewed in PR #91939 and should be merged first. This PR then just adds one commit with the scheduler change.

```release-note
scheduler: optionally check for available storage capacity before scheduling pods which have unbound volumes (alpha feature with the new `CSIStorageCapacity` feature gate, only works for CSI drivers and depends on support for the feature in a CSI driver deployment)
```

```docs
- [KEP]: https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1472-storage-capacity-tracking
- [Usage]: https://github.com/kubernetes/website/pull/21634
```


pohly:
/hold

Let's merge https://github.com/kubernetes/kubernetes/pull/91939 first, then I'll rebase to resolve the merge conflict in the new commit.

pohly:
The merge conflict was trivial and only affected a function comment. I've rebased this code here, but will do that one more time to get rid of the API change commits once https://github.com/kubernetes/kubernetes/pull/91939 is merged.

pkg/controller/volume/scheduling/scheduler_binder.go
```diff
@@ -849,6 +886,71 @@ func (b *volumeBinder) revertAssumedPVCs(claims []*v1.PersistentVolumeClaim) {
...
+	if !ok {
+		// No capacity to check for.
+		return true, nil
+	}
+
+	// Only enabled for CSI drivers which opt into it.
+	driver, err := b.capacityCheck.CSIDriverInformer.Lister().Get(provisioner)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Either the provisioner is no CSI driver or the driver does not
```
msau42, line 908: typo: "not a CSI driver"

pkg/controller/volume/scheduling/scheduler_binder.go
```diff
@@ -161,17 +169,31 @@ type volumeBinder struct {
 	bindTimeout time.Duration
 
 	translator InTreeToCSITranslator
+
+	capacityCheck *CapacityCheck
```
msau42, line 173: Eventually when this goes GA, I think this structure should be folded into the main struct

pkg/controller/volume/scheduling/scheduler_binder_test.go
```diff
@@ -2010,3 +2130,114 @@ func TestFindAssumeVolumes(t *testing.T) {
...
+		reasons, err := findPodVolumes(testEnv.binder, pod, testNode)
+
+		// Validate
+		if !scenario.shouldFail && err != nil {
+			t.Errorf("returned error: %v", err)
+		}
+		if scenario.shouldFail && err == nil {
+			t.Error("returned success but expected error")
+		}
+		checkReasons(t, reasons, scenario.reasons)
```
msau42, line 2237: Do we want to validate Pod cache in the success case?

pkg/scheduler/framework/plugins/volumebinding/volume_binding.go
```diff
@@ -233,7 +238,14 @@ func New(plArgs runtime.Object, fh framework.FrameworkHandle) (framework.Plugin,
 	pvInformer := fh.SharedInformerFactory().Core().V1().PersistentVolumes()
 	storageClassInformer := fh.SharedInformerFactory().Storage().V1().StorageClasses()
 	csiNodeInformer := fh.SharedInformerFactory().Storage().V1().CSINodes()
-	binder := scheduling.NewVolumeBinder(fh.ClientSet(), nodeInformer, csiNodeInformer, pvcInformer, pvInformer, storageClassInformer, time.Duration(args.BindTimeoutSeconds)*time.Second)
+	var capacityCheck *scheduling.CapacityCheck
+	if utilfeature.DefaultFeatureGate.Enabled(features.CSIStorageCapacity) {
+		capacityCheck = &scheduling.CapacityCheck{
+			CSIDriverInformer:          fh.SharedInformerFactory().Storage().V1().CSIDrivers(),
```
msau42, line 244: I think rbac rules for the scheduler need to be updated with these permissions as well.

pkg/controller/volume/scheduling/scheduler_binder_test.go
```diff
@@ -2010,3 +2130,114 @@ func TestFindAssumeVolumes(t *testing.T) {
 		testEnv.validatePodCache(t, testNode.Name, pod, expectedBindings, nil)
 	}
 }
+
+// TestCapacity covers different scenarios involving CSIStorageCapacity objects.
+// Scenarios without those are covered by TestFindPodVolumesWithProvisioning.
+func TestCapacity(t *testing.T) {
```
msau42, line 2136: Can you also add some integration test cases so we test the full interaction with etcd?

https://github.com/kubernetes/kubernetes/blob/master/test/integration/volumescheduling/volume_binding_test.go

pohly: Okay, done.

pohly (commented):
I added E2E and integration tests and rebased on top of the revised API (moved CSIDriverSpec.StorageCapacity to the end of the struct) and current master (some code conflicts again, nothing major).

pkg/scheduler/framework/plugins/volumebinding/volume_binding.go
msau42: Would the integration test have caught the rbac issue?

The reason to not rely on events wasn't strictly about shortening test time, it was because events are unreliable and could be dropped/garbage collected under high load

pohly: > Would the integration test have caught the rbac issue?

Not sure... let's test it... no. The test completes fine even with the new RBAC permission missing.

> The reason to not rely on events wasn't strictly about shortening test time, it was because events are unreliable and could be dropped/garbage collected under high load.

So one cannot rely on them for the test (i.e. not encountering the event must not be a test failure). But if the event is seen, the test could be stopped early. If slow test execution isn't an issue, then this is probably not worth the complexity.

pkg/controller/volume/scheduling/scheduler_binder.go
msau42: I believe @cofyc has a fix for this in https://github.com/kubernetes/kubernetes/pull/92684

msau42:
Looks like the new integration tests are failing.

Can you also update https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L346 so we can run the new e2e tests in the alpha job?

pkg/controller/volume/scheduling/scheduler_binder.go
```diff
@@ -131,6 +136,9 @@ type SchedulerVolumeBinder interface {
 	// It returns an error when something went wrong or a list of reasons why the node is
 	// (currently) not usable for the pod.
 	//
+	// If the feature is enabled, then it also checks for sufficient storage
```
cofyc, line 139: says `the CSIStorageCapacity feature` explicitly?

pkg/controller/volume/scheduling/scheduler_binder.go
```diff
@@ -836,6 +881,75 @@ func (b *volumeBinder) revertAssumedPVCs(claims []*v1.PersistentVolumeClaim) {
...
+			capacity.Capacity.Value() >= sizeInBytes &&
+			b.nodeHasAccess(node, capacity) {
+			// Enough capacity found.
+			return true, nil
+		}
+	}
+
+	// TODO (?): this doesn't give any information about which pools where considered and why
+	// they had to be rejected. Log that above? But that might be a lot of log output...
+	klog.V(4).Infof("Node %q has no accessible CSIStorageCapacity with enough capacity for a volume of size %d and storage class %q", node.Name, sizeInBytes, storageClass.Name)
```
cofyc, line 934: would it be useful to log the PVC ns/name in the log? knowing which PVC can't be dynamically provisioned in this topology

pkg/controller/volume/scheduling/scheduler_binder.go
```diff
@@ -836,6 +881,75 @@ func (b *volumeBinder) revertAssumedPVCs(claims []*v1.PersistentVolumeClaim) {
...
+			return true, nil
+		}
+		return false, err
+	}
+	if driver.Spec.StorageCapacity == nil || !*driver.Spec.StorageCapacity {
+		return true, nil
+	}
+
+	// Look for a matching CSIStorageCapacity object(s).
+	// TODO (for beta): benchmark this and potentially introduce some kind of lookup structure.
```
cofyc, line 915: some thoughts: indexing by storage class should be valuable as the class field is immutable
we also can index capacity objects by nodes as the labels normally do not change frequently

cofyc (commented):
the scheduler volume binder change LGTM, just a few comments

pkg/controller/volume/scheduling/scheduler_binder.go
```diff
@@ -836,6 +881,75 @@ func (b *volumeBinder) revertAssumedPVCs(claims []*v1.PersistentVolumeClaim) {
...
+			capacity.Capacity != nil &&
+			capacity.Capacity.Value() >= sizeInBytes &&
+			b.nodeHasAccess(node, capacity) {
+			// Enough capacity found.
+			return true, nil
+		}
+	}
+
+	// TODO (?): this doesn't give any information about which pools where considered and why
+	// they had to be rejected. Log that above? But that might be a lot of log output...
```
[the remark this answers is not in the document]
pohly: So shall I remove the TODO?

pohly:
> Looks like the new integration tests are failing.

The only error that I saw in a log looked unrelated and the latest job run was successful.

> Can you also update https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L346 so we can run the new e2e tests in the alpha job?

See https://github.com/kubernetes/test-infra/pull/18156

pkg/apis/storage/types.go
```diff
@@ -424,3 +445,81 @@ type CSINodeList struct {
...
+	metav1.ObjectMeta
+
+	// NodeTopology defines which nodes have access to the storage
+	// for which capacity was reported. If not set, the storage is
+	// not accessible from any node in the cluster. If empty, the
+	// storage is accessible from all nodes.  This field is
+	// immutable.
+	//
+	// +optional
+	NodeTopology *metav1.LabelSelector
```
thockin, line 490: What happens if a candidate node matches more than one capacity?  E.g.

```
CSIStorageCapacity x {
  StorageClassName: standard
  NodeTopology:
    MatchLabels:
      zone: east1
  Capacity: 10Gi
}

CSIStorageCapacity y {
  StorageClassName: standard
  NodeTopology:
    MatchLabels:
      region: east
  Capacity: 1Gi
}
```

In this case,one is not a superset of the other as defined.  I can maybe assume that region is a superset of zone (though that's still pending in a KEP) but storage still allows other topology keys.

what is a consumer of CSIStorageCapacity supposed to do in this case?

thockin (commented):
API LGTM, with one question about conflict resolution in case of overlap.

Ping me when ready for approval overall

pkg/apis/storage/types.go
pohly: That's not really a conflict. If a node has access to both of these and the volume is smaller than 10Gi, then it's fine to go ahead and pick the node. The storage system then gets to decide where exactly it provisions the volume, just as it does now for volumes with immediate binding. What matters is that the volume can be made usable by the chosen node.

A conflict would be when there are two objects with the exact same `StorageClassName` and `NodeTopology` but different capacity. This situation is not allowed and the producer is supposed to avoid it. The consumer could try to detect it, but currently doesn't because it would imply iterating over all objects instead of stopping at the first one - IMHO it's not worth the cost because it really shouldn't happen.

pkg/controller/volume/scheduling/scheduler_binder.go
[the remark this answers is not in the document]
cofyc: it's fine to keep this.

pohly:
Some of the generated files in the PR were stale. Fixed now, tests should pass again.

pohly:
/hold

@jsafrane pointed out an issue in the KEP: the non-namespaced `CSIStorageCapacity` objects cannot be owned by the namespaced CSI driver deployment objects.

We considered making the non-namespaced `CSIDriver` object the owner. But `CSIDriverSpec` is immutable, so currently one has to remove and re-add it when making changes, like enabling the storage capacity tracking check. Then loosing the `CSIStorageCapacity` objects would prevent the smooth rollout and would be surprising.

Proposal: let's just make `CSIStorageCapacity` namespaced. This has no impact for the consumer (it can still list all objects with an empty namespace string) and has the additional benefit that CSI driver deployments can be done so that the driver can only modify CSIStorageCapacity objects in its own namespace. This doesn't prevent the publishing of incorrect information, but at least it then would be obvious from where it came.

One concern I had was whether that then would allow normal users to create those objects in their personal namespace. But @jsafrane confirmed that at least in OpenShift, object types specifically need to be whitelisted. That makes sense, so I assume any locked down cluster with untrusted users will do the same.

@thockin: are you okay with this API change?

I'm going to make the change in this PR right away, to save time. I can always go back if we decide to not do it that way...

pohly:
I've force-pushed the namespace change.

@thockin: please cancel the hold if you agree with this API change (https://github.com/kubernetes/kubernetes/pull/92387#issuecomment-653498857).

jsafrane:
I think that also vanilla Kubernetes gives "namespace editors" (=regular users?) capability to edit only explicitly whitelisted objects:
https://github.com/kubernetes/kubernetes/blob/50d574bf7f4ed24e20fa857642a9eb5e504cdfa2/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go#L268-L278
(there are lot of rules there)

pkg/controller/volume/scheduling/scheduler_binder.go
ahg-g: lets create an issue to track all optimization work and use it in the todo here.

pohly: I've started a TODO list in the enhancement issue: https://github.com/kubernetes/enhancements/issues/1698#issuecomment-654356718

It's also part of the beta graduation criteria in the KEP, but it makes sense to track specific TODOs like this one so that they don't get forgotten.

pkg/controller/volume/scheduling/scheduler_binder.go
```diff
@@ -837,6 +882,76 @@ func (b *volumeBinder) revertAssumedPVCs(claims []*v1.PersistentVolumeClaim) {
...
+	return false, nil
+}
+
+func (b *volumeBinder) nodeHasAccess(node *v1.Node, capacity *storagev1alpha1.CSIStorageCapacity) bool {
+	if capacity.NodeTopology == nil {
+		// Unavailable
+		return false
+	}
+	// Only matching by label is supported.
+	selector, err := metav1.LabelSelectorAsSelector(capacity.NodeTopology)
```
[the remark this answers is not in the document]
pohly: Added to https://github.com/kubernetes/enhancements/issues/1698#issuecomment-654356718

Indexing objects by node should address this (https://github.com/kubernetes/kubernetes/pull/92387/files/99896e4225202e25ba1924458fb104a14dfb9d27#r449032580).

pkg/scheduler/framework/plugins/volumebinding/volume_binding.go
```diff
@@ -254,7 +259,14 @@ func New(plArgs runtime.Object, fh framework.FrameworkHandle) (framework.Plugin,
 	pvInformer := fh.SharedInformerFactory().Core().V1().PersistentVolumes()
 	storageClassInformer := fh.SharedInformerFactory().Storage().V1().StorageClasses()
 	csiNodeInformer := fh.SharedInformerFactory().Storage().V1().CSINodes()
-	binder := scheduling.NewVolumeBinder(fh.ClientSet(), podInformer, nodeInformer, csiNodeInformer, pvcInformer, pvInformer, storageClassInformer, time.Duration(args.BindTimeoutSeconds)*time.Second)
+	var capacityCheck *scheduling.CapacityCheck
+	if utilfeature.DefaultFeatureGate.Enabled(features.CSIStorageCapacity) {
```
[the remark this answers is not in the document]
pohly: This feature check here is only called once when initializing the scheduler, not for every pod.

pohly (commented):
I've force-pushed a version with the TODO comment updated.

pkg/scheduler/framework/plugins/volumebinding/volume_binding.go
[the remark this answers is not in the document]
ahg-g: oh right.

pkg/controller/volume/scheduling/scheduler_binder.go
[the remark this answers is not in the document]
ahg-g: sg, thanks

ahg-g:
scheduler-related changes looks good to me.
/lgtm

msau42:
> We considered making the non-namespaced `CSIDriver` object the owner. But `CSIDriverSpec` is immutable, so currently one has to remove and re-add it when making changes

What if we do make it mutable? It seems odd to make this namespaced considering that all other csi related objects are non-namespaced and I think having a namespaced object generally implies the kind of users we expect to use it.

thockin:
> @jsafrane pointed out an issue in the KEP: the non-namespaced CSIStorageCapacity objects cannot be owned by the namespaced CSI driver deployment objects.
>
> We considered making the non-namespaced CSIDriver object the owner. But CSIDriverSpec is immutable, so currently one has to remove and re-add it when making changes, like enabling the storage capacity tracking check. Then loosing the CSIStorageCapacity objects would prevent the smooth rollout and would be surprising.

Making the deployments own the Capacity feels wrong to me - the CSIDriver seems
like the "correct" coupling, no?  As for mutability, it is possible to make
fields mutable.  Requiring a delete-and-recreate to "fake" mutability seems
kind of awful to me...

> Proposal: let's just make CSIStorageCapacity namespaced. This has no impact for the consumer (it can still list all objects with an empty namespace string) and has the additional benefit that CSI driver deployments can be done so that the driver can only modify CSIStorageCapacity objects in its own namespace. This doesn't prevent the publishing of incorrect information, but at least it then would be obvious from where it came.

The "can only modify CSIStorageCapacity objects in its own namespace" is the
only reason I can think of that justifies this.  That said, it doesn't prevent
driver A from creating a capacity object for a class B is owned by driver B,
does it?

In general, sharing a namespace across 2 non-coordinated actors is likely to
end badly.  In this case, that is the same for the root. Is there any "pointer"
from a CSIDriver to the Namespace that implements that driver?  If there were,
this would be a less awkward idea.

> One concern I had was whether that then would allow normal users to create those objects in their personal namespace. But @jsafrane confirmed that at least in OpenShift, object types specifically need to be whitelisted. That makes sense, so I assume any locked down cluster with untrusted users will do the same.

We rely on the idea that a "normal" user does not have access to EVERY resource
in their namespace, that's true.

> @thockin: are you okay with this API change?

I don't love it.  Tying lifetime to the Deployment doesn't feel right, but
there is an attractive property to not giving drivers access to the root, as
much as possible.  I kind of like the idea that a CSIDriver "knows" which
namespace it "owns".  That way, we can link CSIDriver's driver name to a
namespace which hold CSIStorageCapacity resources for THAT driver alone.  Will
there be more of these sorts of resources?  If there's one, there will probably
be two...

But also, don't fake mutability.

msau42:
> That way, we can link CSIDriver's driver name to a
> namespace which hold CSIStorageCapacity resources for THAT driver alone

Many csi drivers install themselves into kube-system namespace so drivers may still share a namespace. This also makes it easier to integrate into any pre-installed PSPs that give admin-level permissions to anything in kube-system.

pohly:
> What if we do make it [CSIDriver] mutable?

I think this would be useful. But can we make that a goal for the next iteration, i.e. alpha2 in 1.20 or beta?

> It seems odd to make this namespaced considering that all other csi related objects are non-namespaced and I think having a namespaced object generally implies the kind of users we expect to use it.

`CSINode` and `CSIDriver` are created and updated by system actors (cluster admin, kubelet). While CSI drivers have to have quite a few privileges, they are not at that level and the less privileges we give them, the better. Therefore I think the situation is a bit different for `CSIStorageCapacity`.

> The "can only modify CSIStorageCapacity objects in its own namespace" is the
only reason I can think of that justifies this.

That was also my thought: no matter what we end up using as owner, making `CSIStorageCapacity` objects namespaced looks like the better approach.

> That said, it doesn't prevent driver A from creating a capacity object for a class B is owned by driver B,
does it?

Correct. In that sense, the security is still as good (or bad) as it is with cluster-scoped `CSIStorageCapacity` - for now! But we can improve upon that.

> Is there any "pointer"
from a CSIDriver to the Namespace that implements that driver? If there were,
this would be a less awkward idea.

There's no such pointer at the moment. It could be added, and then the consumer could check whether a `CSIStorageCapacity` object is in the right namespace before using it.

pohly:
> >  What if we do make it [CSIDriver] mutable?
>
> I think this would be useful. But can we make that a goal for the next iteration, i.e. alpha2 in 1.20 or beta?

I've added that to the todo list that I started keeping in the feature issue: https://github.com/kubernetes/enhancements/issues/1698#issuecomment-654356718

thockin (commented):
/approve

API approved

pohly:
/hold cancel

We (@msau42, @thockin, myself) discussed namespacing of `CSIStorageCapacity` and decided to go with it for now, with a TODO item in https://github.com/kubernetes/enhancements/issues/1472#issuecomment-655098092 to re-evaluate that choice.

pohly:
/hold

Waiting for CSIStorageCapacity tests in 
pull-kubernetes-e2e-gce-alpha-features to pass.

test/e2e/storage/csi_mock_volume.go
```diff
@@ -944,8 +945,194 @@ var _ = utils.SIGDescribe("CSI mock volume", func() {
 			})
 		}
 	})
+
+	// These tests *only* work on a cluster which has the CSIStorageCapacity feature enabled.
+	ginkgo.Context("CSIStorageCapacity [Feature: CSIStorageCapacity]", func() {
```
msau42, line 950: There shouldn't be a whitespace after "Feature".

Temporarily adding a whitespace in the test job so that we can get a run in today. https://github.com/kubernetes/test-infra/pull/18201

msau42:
/lgtm
/priority important-soon

/hold
Can you update the release note to clarify that this is a new alpha feature with the feature gate name, and that it's only for csi volumes?

test/e2e/storage/csi_mock_volume.go
pohly: Fixed.

pohly:
> Can you update the release note to clarify that this is a new alpha feature with the feature gate name, and that it's only for csi volumes?

Done.

pohly:
/retest

TestSubresourcePatch in pull-kubernetes-integration seems to be flaky.

pohly:
> Waiting for CSIStorageCapacity tests in pull-kubernetes-e2e-gce-alpha-features to pass.

Which they have done in https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92387/pull-kubernetes-e2e-gce-alpha-features/1280744225868091392/

PR #92200, opened 2020-06-16
scheduler: merge Reserve and Unreserve plugins

adtac:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
/kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:

Previously, separate interfaces were defined for Reserve and Unreserve
plugins. However, in nearly all cases, a plugin that allocates a
resource using Reserve will likely want to register itself for Unreserve
as well in order to free the allocated resource at the end of a failed
scheduling/binding cycle. Having separate plugins for Reserve and
Unreserve also adds unnecessary config toil. To that end, this patch
aims to merge the two plugins into a single interface called a
ReservationPlugin that requires implementing both the Reserve and
Unreserve methods.

**Which issue(s) this PR fixes**:

Fixes https://github.com/kubernetes/kubernetes/issues/90724

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Unreserve extension point for scheduler plugins is merged into Reserve extension point
```

PR #83557, opened 2019-10-07
Plugins that reserved must unreserve when another plugin fails to reserve

mrkm4ntr:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:
If one of reserve plugins failed, the plugins already reserved will not be unreserved.
At least this will be a problem when migrating TAVP. #82934 

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #92391, opened 2020-06-22
scheduler: run Unreserve if Reserve fails

adtac:
**What type of PR is this**:

/kind bug

If a reserve plugin's Reserve method returns an error, there could be
previously allocated resources from successfully completed reserve
plugins that must be unallocated by the corresponding Unreserve
operation. Since Unreserve operations are idempotent, this patch runs
the Unreserve operation of ALL reserve plugins when a Reserve operation
fails.

Replaces https://github.com/kubernetes/kubernetes/pull/83557

This depends on https://github.com/kubernetes/kubernetes/pull/92200 - if you'd like to review before that is merged, please only look at the commit https://github.com/kubernetes/kubernetes/pull/92391/commits/59943193d0917a83935f2b40fe478ecb46d8c19e

```release-note
Reserve plugins that fail to reserve will trigger the unreserve extension point
```

```docs

```

### DIFF
--- a/pkg/scheduler/framework/runtime/framework.go
+++ b/pkg/scheduler/framework/runtime/framework.go
@@ -788,7 +788,8 @@ func (f *frameworkImpl) runPostBindPlugin(ctx context.Context, pl framework.Post
 // RunReservePluginsReserve runs the Reserve method in the set of configured
 // reserve plugins. If any of these plugins returns an error, it does not
 // continue running the remaining ones and returns the error. In such a case,
-// the pod will not be scheduled.
+// the pod will not be scheduled and the caller will be expected to call
+// RunReservePluginsUnreserve.
 func (f *frameworkImpl) RunReservePluginsReserve(ctx context.Context, state *framework.CycleState, pod *v1.Pod, nodeName string) (status *framework.Status) {
 	startTime := time.Now()
 	defer func() {
--- a/pkg/scheduler/framework/v1alpha1/interface.go
+++ b/pkg/scheduler/framework/v1alpha1/interface.go
@@ -333,11 +333,14 @@ type ScorePlugin interface {
 type ReservePlugin interface {
 	Plugin
 	// Reserve is called by the scheduling framework when the scheduler cache is
-	// updated.
+	// updated. If this method returns a failed Status, the scheduler will call
+	// the Unreserve method for all enabled ReservePlugins.
 	Reserve(ctx context.Context, state *CycleState, p *v1.Pod, nodeName string) *Status
 	// Unreserve is called by the scheduling framework when a reserved pod was
 	// rejected, an error occurred during reservation of subsequent plugins, or
-	// in a later phase. The Unreserve method implementation must be idempotent.
+	// in a later phase. The Unreserve method implementation must be idempotent
+	// and may be called by the scheduler even if the corresponding Reserve
+	// method for the same plugin was not called.
 	Unreserve(ctx context.Context, state *CycleState, p *v1.Pod, nodeName string)
 }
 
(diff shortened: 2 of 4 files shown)


test/integration/scheduler/framework_test.go
```diff
@@ -1003,30 +1004,39 @@ func TestReservePluginUnreserve(t *testing.T) {
...
-		preBindFail bool
+		name             string
+		failReserve      bool
+		failReserveIndex int
+		failPreBind      bool
 	}{
+		{
+			name:             "fail reserve",
+			failReserve:      true,
+			failReserveIndex: 1,
```
alculquicondor, line 1015: this seems unnecessary

adtac: my intention behind that was to test that a failure in a reserve plugin that's in the middle of the list will still trigger a unreserve in ALL plugins

alculquicondor: got it. looks good.

alculquicondor:
Please rebase and squash

adtac:
Done.

pkg/scheduler/scheduler.go
```diff
@@ -520,6 +520,8 @@ func (sched *Scheduler) scheduleOne(ctx context.Context) {
 	if sts := prof.RunReservePluginsReserve(schedulingCycleCtx, state, assumedPod, scheduleResult.SuggestedHost); !sts.IsSuccess() {
 		sched.recordSchedulingFailure(prof, assumedPodInfo, sts.AsError(), SchedulerError, "")
 		metrics.PodScheduleError(prof.Name, metrics.SinceInSeconds(start))
+		// trigger un-reserve to clean up state associated with the reserved Pod
+		prof.RunReservePluginsUnreserve(schedulingCycleCtx, state, assumedPod, scheduleResult.SuggestedHost)
```
alculquicondor, line 524: We need to document that this is called in the interface for ReservePlugin

adtac: Good point. Done.

pkg/scheduler/framework/v1alpha1/interface.go
```diff
@@ -333,7 +333,8 @@ type ScorePlugin interface {
 type ReservePlugin interface {
 	Plugin
 	// Reserve is called by the scheduling framework when the scheduler cache is
-	// updated.
+	// updated. If this method returns a failed Status, the scheduler will call
+	// the Unreserve method for all enabled ReservePlugins.
 	Reserve(ctx context.Context, state *CycleState, p *v1.Pod, nodeName string) *Status
 	// Unreserve is called by the scheduling framework when a reserved pod was
 	// rejected, an error occurred during reservation of subsequent plugins, or
```
alculquicondor, line 340: Instead of "The Unreserve method implementation must be idempotent." To be more explicit, let's say: "Note that Unreserve might be called even if Reserve for the same plugin was not."

alculquicondor (commented):
@mrkm4ntr anything to add?

pkg/scheduler/framework/v1alpha1/interface.go
adtac: I think it makes sense to include both statements -- I've added a line to say that.

pkg/scheduler/framework/runtime/framework.go
```diff
@@ -788,7 +788,7 @@ func (f *frameworkImpl) runPostBindPlugin(ctx context.Context, pl framework.Post
 // RunReservePluginsReserve runs the Reserve method in the set of configured
 // reserve plugins. If any of these plugins returns an error, it does not
 // continue running the remaining ones and returns the error. In such a case,
-// the pod will not be scheduled.
+// the pod will not be scheduled and Unreserve will be called for all plugins.
```
alculquicondor, line 791: uhmm... this function doesn't actually call Unreserve. The comment could say:
"In such a case, the pod willl not be scheduled and RunReservePluginsUnreserve should be called afterwards"

adtac: ugh, I forgot that I changed my implementation; remember when it used to be such that RunReservePluginsReserve would call RunReservePluginsUnreserve :D

good catch :)

alculquicondor:
/approve

Please squash

adtac:
Squashed.

alculquicondor:
/lgtm

/hold
for @mrkm4ntr. Feel free to unhold when we are closer to code freeze in case of no answer.

mrkm4ntr (approved):
lgtm
/unhold

PR #92392, opened 2020-06-22
Define GetBody of http.Request for automatic retry

tkashem:
**What type of PR is this?**
> /kind api-change

/kind bug

> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Add a unit test that ensures that `rest.Request.body`, if not empty always produces a `http.Request` with a non-nil `GetBody`.

`net/http2` retries a request if it receives a `GOAWAY` frame.  `GOAWAY` frame has `LastStreamID` which represents the last stream identifier the server has processed or is aware of. On the client
side, `net/http2` will set an error (`http2: Transport received Server's graceful shutdown GOAWAY` for any stream that has an identifier higher than that of LastStreamID.

However, if the request body is not empty then net/http2 will retry only when it can reset back the request body to its original state via the optional `req.GetBody`.

This is where it happens in net/http2
https://github.com/golang/go/blob/master/src/net/http/h2_bundle.go#L6991

NONE

-->
```release-note
NONE
```

### DIFF
--- a/staging/src/k8s.io/client-go/rest/request_test.go
+++ b/staging/src/k8s.io/client-go/rest/request_test.go
@@ -2278,3 +2278,72 @@ func TestRequestMaxRetries(t *testing.T) {
 		})
 	}
 }
+
+func TestBodyEnsureHTTPGetBodyIsSet(t *testing.T) {
+	f, err := ioutil.TempFile("", "test-body-")
+	if err != nil {
+		t.Fatalf("failed to create file for request body: %v", err)
+	}
+	defer os.Remove(f.Name())
+
+	if _, err := f.WriteString("foo"); err != nil {
+		t.Fatalf("failed to write to file: %v", err)
+	}
+	f.Close()
+
+	tests := []struct {
+		name  string
+		input interface{}
+	}{
+		{
+			name:  "WithRuntimeObject",
+			input: &v1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "foo"}},
+		},
+		{
+			name:  "WithByteSlice",
+			input: []byte("foo"),
+		},
+		{
+			name:  "WithStringFromFile",
+			input: f.Name(),
+		},
+	}
+
+	base, _ := url.Parse("http://localhost")
+	versionedAPIPath := defaultResourcePathWithPrefix("", "", "", "")
+	client, err := NewRESTClient(base, versionedAPIPath, defaultContentConfig(), nil, nil)
+	if err != nil {
+		t.Fatalf("failed to create a client: %v", err)
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			r := Request{
+				c:          client,
+				pathPrefix: "/path",
+			}
+			r.Body(test.input)
+
+			request, err := http.NewRequest(r.verb, r.URL().String(), r.body)
+			if err != nil {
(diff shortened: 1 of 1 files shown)


sttts:
Great PR!

Can one test the retry behaviour somehow in a unit test?

p0lyn0mial:
great finding @tkashem

/lgtm

tkashem:
@sttts I am looking into unit test/integration test.

fedebongio:
/cc @caesarxuchao 
(just for fyi)

tkashem:
@sttts after further debugging I found out that net/http sets `GetBody` for some type of request body but not all. I think we can keep it as a safeguard to ensure that all net/http requests with a body have a `GetBody` defined.

Also, as far as unit testing is concerned, there is an inherent race condition between the server sending a `GOAWAY` frame with `LastStreamID=N` and the client generating a new stream `P (P>N)` before it receives the `GOAWAY` frame.
- `E1`: server sends a `GOAWAY` frame `LastStreamID=N`
- `E2`: client receives the `GOAWAY` frame and removes the underlying tcp connection from the pool.

The test needs to generate the new stream between `E1` and `E2`. If the test can manage to generate a new stream `P` between `E1` and `E2` then `P` will be ignored by the server. The client will retry `P` on a new tcp connection and `GetBody` will be utilized by net/http. But we don't have the knobs in net/http/httptest to design such a test in a deterministic way.

If `P` is generated after `E2` then it will be tried on a new tcp connection with a `StreamID=1` and the `GOAWAY` frame sent by the server does not have any impact on `P`.

sttts:
Lgtm. But this needs more eyes as it is super central.

sttts:
/lgtm
/approve
/hold

Hold for more eyes.

staging/src/k8s.io/client-go/rest/request.go
```diff
@@ -869,6 +869,27 @@ func (r *Request) request(ctx context.Context, fn func(*http.Request, *http.Resp
...
+		//
+		// This is where it happens in net/http2
+		// https://github.com/golang/go/blob/master/src/net/http/h2_bundle.go#L6991
+		//
+		// Let's set GetBody for our request so net/http2 automatically retry it. It's ideal if
+		// we allow net/http2 to retry a request before we do a retry here.
+		//
+		// net/http sets GetBody for some type of body but not all, so it's good to
+		// ensure that all requests with a body have a GetBody defined.
+		if r.body != nil && req.GetBody == nil {
```
liggitt, line 887: shouldn't this be limited to r.verb == GET or bodies that can be reset? the GetBody implemented here doesn't do anything to reset the body to its original state, which means it is not implemented according to the godoc for GetBody

tkashem: Thanks @liggitt for your feedback. 
The `Body` function sets the `body` of a `rest.Request` object to an instance of `*bytes.Reader` if the input is `Runtime.Object`, `[]byte` or `string`. On the other hand, If body is of type `*bytes.Reader` then `http.NewRequest` will set a `GetBody` for us - https://github.com/golang/go/blob/master/src/net/http/request.go#L881-L887

So I think we are covered, I have reduced my PR to a unit test that ensures that `rest.Request.body`, if not empty always produces a `http.Request` with a non-nil `GetBody`.

PR #92400, opened 2020-06-22
decouple testfiles from framework

spiffxp:
**What type of PR is this?**
/kind cleanup

This drops testfiles.ReadOrDie and updated testfiles.Exists to return an
error, forcing the caller to decide whether to call framework.Fail or do
something else. It's slightly less friendly / more verbose to use now.
But test/e2e/framework/testfiles is back to being a self-contained
package as described in its comments.

```release-note
NONE
```


oomichi (commented):
looks straightforward change.

/retest
/lgtm

PR #75223, opened 2019-03-08
kubelet: include init containers when determining pod QoS

sjenning:
fixes https://github.com/kubernetes/kubernetes/issues/75212

@derekwaynecarr @dashpole 

```release-note
Init container resource requests now impact pod QoS class
```
/sig node

PR #92403, opened 2020-06-23
ensure requests equals to limits in static policy

jfbai:
**What type of PR is this?**

> /kind api-change
/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Static policy allocates CPUs (aligned to `requests`) to init container when requests and limits are not equal.

Fixes #

No

```release-note
No
```

```docs
No
```
/sig node
/assign @klueska

### DIFF
--- a/pkg/kubelet/cm/cpumanager/policy_static.go
+++ b/pkg/kubelet/cm/cpumanager/policy_static.go
@@ -299,6 +299,10 @@ func (p *staticPolicy) guaranteedCPUs(pod *v1.Pod, container *v1.Container) int 
 		return 0
 	}
 	cpuQuantity := container.Resources.Requests[v1.ResourceCPU]
+	cpuLimitsQuantity := container.Resources.Limits[v1.ResourceCPU]
+	if cpuQuantity.Value() != cpuLimitsQuantity.Value() {
+		return 0
+	}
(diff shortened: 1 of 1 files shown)


klueska:
What version of Kubernetes are you running when you see this issue?

As of `v1.16` init containers are included in the decision to mark a pod as part of the `Guaranteed` QoS class. This makes your check for `requests == limits` redundant since the check above the one you are adding already implies that this invariant is true.

Here is the PR that added this support: https://github.com/kubernetes/kubernetes/pull/75223/files

klueska:
@jfbai Could you respond to my question above when you get a chance?
https://github.com/kubernetes/kubernetes/pull/92403#issuecomment-648037633

I'm actually surprised you are would see an issue on master.

jfbai:
> What version of Kubernetes are you running when you see this issue?
> 
> As of `v1.16` init containers are included in the decision to mark a pod as part of the `Guaranteed` QoS class. This makes your check for `requests == limits` redundant since the check above the one you are adding already implies that this invariant is true.
> 
> Here is the PR that added this support: https://github.com/kubernetes/kubernetes/pull/75223/files

Sorry for delay, I've back ported topology manager related changes into internal 1.14 version. Thanks for you information.

PR #92407, opened 2020-06-23
change kubelet metrics name

RainbowMango:
**What type of PR is this?**
/kind feature

https://github.com/kubernetes/kubernetes/blob/3a51aaf8b4baff1242ea30b08d599d913fe7a127/staging/src/k8s.io/component-base/metrics/testutil/promlint.go#L70-L72

Refer [discussion here](https://kubernetes.slack.com/archives/C20HH14P7/p1592661271239300).

```release-note
kubelet: following metrics have been renamed:
kubelet_running_container_count --> kubelet_running_containers
kubelet_running_pod_count --> kubelet_running_pods
```

/priority important-soon

### DIFF
--- a/pkg/kubelet/metrics/metrics.go
+++ b/pkg/kubelet/metrics/metrics.go
@@ -349,7 +349,7 @@ var (
 	RunningPodCount = metrics.NewGauge(
 		&metrics.GaugeOpts{
 			Subsystem:      KubeletSubsystem,
-			Name:           "running_pod_count",
+			Name:           "running_pods",
 			Help:           "Number of pods currently running",
 			StabilityLevel: metrics.ALPHA,
 		},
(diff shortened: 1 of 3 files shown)


RainbowMango:
/assign @vishh @dashpole 
cc @brancz @sftim

sftim:
This is a breaking change, right? I think existing metrics and PromQL could stop working when this ships.

brancz:
These are alpha metrics, so this change is permitted.

pkg/kubelet/metrics/metrics.go
```diff
@@ -349,7 +349,7 @@ var (
 	RunningPodCount = metrics.NewGauge(
```
SergeyKanzhelev, line 349: not critical at all. This name change changes from `pod` (singular) to `pods` (plural) so maybe the same rename can be done for the variable name.

SergeyKanzhelev (commented):
/lgtm

dashpole:
@serathius @kubernetes/sig-node-pr-reviews 
for visibility
/approve

pkg/kubelet/metrics/metrics.go
RainbowMango: Thanks for reminding, I never thought about the variable names, just focusing the metrics...But I guess it's optional.

fedebongio:
/cc @logicalhan 
(for visibility)

PR #92413, opened 2020-06-23
Add protocol buffer generation to types.NamespacedName.

markusthoemmes:
**What type of PR is this?**
/kind feature

This adds protocol buffer generation to `types.NamespacedName` to enable downstream projects to reuse that type in protobuf schemas. Dunno what more to say really :sweat_smile:.

This change could potentially be made smaller if we don't add it to the default list of packages to pull in. I'm happy to change that if reviewers think that way.

```release-note
NONE
```

```docs

```


markusthoemmes:
Talked to @sttts about this (thanks for the quick feedback) and we're **not** going to do this.

Issue #92402, opened 2020-06-23
fix staticcheck failures


BenTheElder:
Many of our packages are failing [staticcheck], we should fix all of them. See previously https://github.com/kubernetes/kubernetes/issues/36858 https://github.com/kubernetes/kubernetes/issues/90208 https://github.com/kubernetes/kubernetes/issues/81657

/help

To fix these, take a look at [hack/.staticcheck_failures] for a list of currently failing packages.

You can check a package by removing it from the list and running `make verify WHAT=staticcheck`.
Once the package is no longer failing, please file a PR including the fixes and removing it from the list.

Before filing your PR, you should run `hack/update-gofmt.sh` and `hack/update-bazel.sh` to auto format your code and update the build.

I recommend keeping PRs scoped down in size for review, fix a package or set of closely related packages, or a certain class of failures, to make it easier for your reviewers to handle. 

We don't need to fix all failures in one PR, but please avoid 100s of single character PRs, as it does cost time & resources to get each PR reviewed, tested, merged etc. :upside_down_face: 

I can help review _some_ of these, but many of them will require other reviewers that own the relevant code.

EDIT: I also don't recommend `/assign` ing to this one, many people are going to need to work on it and we want more people to join in.
You can link to this issue in your PR and comment here that you're working on certain packages to help avoid duplication.
EDIT2: Please do NOT put `Fixes #...` in your Pull Request, despite the template. That will close this PR. Instead put something like `Part of #92402`

I think each of these own code that is currently failing:
/sig testing
/sig storage
/sig api-machinery
/sig architecture
/sig cli
/sig instrumentation
/sig autoscaling
/sig cloud-provider

[staticcheck]: https://staticcheck.io/
[hack/.staticcheck_failures]: https://github.com/kubernetes/kubernetes/blob/master/hack/.staticcheck_failures

skilxn-go:
Sorry, didn't notice this before.

> EDIT: I also don't recommend /assign ing to this one, many people are going to need to work on it and we want more people to join in.

BenTheElder:
thanks :-)
yeah, I think people tend to not work on issues that are already assigned, but I don't think anyone one person is going to get PRs in for all of this, there's lots of room for many to work on this :-)

BenTheElder:
/lifecycle active
I opened a simple one for a package I own, https://github.com/kubernetes/kubernetes/pull/92408
[...]

PR #92421, opened 2020-06-23
fix pkg/controller/deployment staticcheck

wawa0210:
**What type of PR is this?**
/kind cleanup

Part of #92402

```release-note
None
```

```docs

```
/assign @soltysh

### DIFF
--- a/pkg/controller/deployment/sync.go
+++ b/pkg/controller/deployment/sync.go
@@ -98,7 +98,7 @@ func (dc *DeploymentController) checkPausedConditions(d *apps.Deployment) error 
 	}
 
 	var err error
-	d, err = dc.client.AppsV1().Deployments(d.Namespace).UpdateStatus(context.TODO(), d, metav1.UpdateOptions{})
+	_, err = dc.client.AppsV1().Deployments(d.Namespace).UpdateStatus(context.TODO(), d, metav1.UpdateOptions{})
 	return err
 }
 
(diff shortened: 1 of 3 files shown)


foxish (approved):
/lgtm

pkg/controller/deployment/sync.go
```diff
@@ -173,7 +173,7 @@ func (dc *DeploymentController) getNewReplicaSet(d *apps.Deployment, rsList, old
 
 		if needsUpdate {
 			var err error
-			if d, err = dc.client.AppsV1().Deployments(d.Namespace).UpdateStatus(context.TODO(), d, metav1.UpdateOptions{}); err != nil {
+			if _, err = dc.client.AppsV1().Deployments(d.Namespace).UpdateStatus(context.TODO(), d, metav1.UpdateOptions{}); err != nil {
```
soltysh, line 176: This `d` is being used later on in code, so it looks to me this is wrong change.

soltysh (changes requested):

pkg/controller/deployment/sync.go
wawa0210: From the code 

https://github.com/kubernetes/kubernetes/blob/da54185c9eb076a71962ddb5aded24811467dbb3/pkg/controller/deployment/sync.go#L149-L188

Here, the assignment of `d` will only be executed within the condition of `if existingNewRS != nil`, and then return directly to jump out of the function. The code segment used behind `d` is not within this logical scope, and the latter `d` is passed in by the current function

It seems like this is okay to modify

soltysh: Yeah, you're right, thx.

soltysh (approved):
/lgtm
/approve
/retest

PR #92423, opened 2020-06-23
dump conntrack table on e2e conntrack failures

aojea:
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature

/kind flake

```release-note
NONE
```

### DIFF
--- a/test/e2e/network/conntrack.go
+++ b/test/e2e/network/conntrack.go
@@ -17,6 +17,7 @@ limitations under the License.
 package network
 
 import (
+	"context"
 	"fmt"
 	"strings"
 	"time"
@@ -116,6 +117,9 @@ var _ = SIGDescribe("Conntrack", func() {
 	})
 
 	ginkgo.It("should be able to preserve UDP traffic when server pod cycles for a NodePort service", func() {
+		// TODO(#91236): Remove once the test is debugged and fixed.
+		// dump conntrack table for debugging
+		defer dumpConntrack(cs)
 
 		// Create a NodePort service
 		udpJig := e2eservice.NewTestJig(cs, ns, serviceName)
@@ -193,6 +197,9 @@ var _ = SIGDescribe("Conntrack", func() {
 	})
 
 	ginkgo.It("should be able to preserve UDP traffic when server pod cycles for a ClusterIP service", func() {
+		// TODO(#91236): Remove once the test is debugged and fixed.
+		// dump conntrack table for debugging
+		defer dumpConntrack(cs)
 
 		// Create a ClusterIP service
 		udpJig := e2eservice.NewTestJig(cs, ns, serviceName)
@@ -269,3 +276,24 @@ var _ = SIGDescribe("Conntrack", func() {
 		}
 	})
 })
+
+func dumpConntrack(cs clientset.Interface) {
+	// Dump conntrack table of each node for troubleshooting using the kube-proxy pods
+	namespace := "kube-system"
+	pods, err := cs.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil || len(pods.Items) == 0 {
+		framework.Logf("failed to list kube-proxy pods in namespace: %s", namespace)
+		return
+	}
+	cmd := "conntrack -L"
+	for _, pod := range pods.Items {
+		if strings.Contains(pod.Name, "kube-proxy") {
+			stdout, err := framework.RunHostCmd(namespace, pod.Name, cmd)
+			if err != nil {
+				framework.Logf("Failed to dump conntrack table of node %s: %v", pod.Spec.NodeName, err)
+				continue
+			}
+			framework.Logf("conntrack table of node %s: %s", pod.Spec.NodeName, stdout)
+		}
+	}
+}


aojea:
/assign @wojtek-t @BenTheElder 
/hold
Let's see if we can catch a failure and the results is the expected, :)

test/e2e/network/conntrack.go
```diff
@@ -268,4 +269,46 @@ var _ = SIGDescribe("Conntrack", func() {
...
+		}
+	})
 })
+
+// getConntrackTable dumps the node conntrack table
+func getConntrackTable(f *framework.Framework, nodeName string) {
+	// Create a pod to check the conntrack entries on the host node
+	privileged := true
+
+	pod := &v1.Pod{
```
wojtek-t, line 287: do we need to create a pod? Can't we reuse the already existing one (like kube-proxy)?

test/e2e/network/conntrack.go
```diff
@@ -268,4 +269,46 @@ var _ = SIGDescribe("Conntrack", func() {
 			framework.Failf("Failed to connect to backend 2")
 		}
 	})
+
+	ginkgo.AfterEach(func() {
```
wojtek-t, line 273: Is it possible that AfterEach is too late?
Should we consider doing e.g. periodically?

test/e2e/network/conntrack.go
wojtek-t: Actually thinking more about it - maybe we can make the pod that we create priviliged and do that not in AfterEach, but rather before deleting the pod?

wojtek-t: BTW - it happens - it just happened to my on the gce presubmit...

aojea: we need a image that has the `conntrack` userspace tool, 
I think agnhost does not have it, reusing kube-proxy pod will be great, 
how can I get the kube-proxy pod name for kind and gce? is there a consistent way to find them?

test/e2e/network/conntrack.go
aojea: it should be enough, udp timeout for new connectins is 30sec, but if the connection "sees" more packets it extended the timeout to 120sec, so if the conntrack deletion does not flush the stale entries we should see it there
xref https://www.kernel.org/doc/Documentation/networking/nf_conntrack-sysctl.txt
```
nf_conntrack_udp_timeout - INTEGER (seconds)
	default 30

nf_conntrack_udp_timeout_stream - INTEGER (seconds)
	default 120

	This extended timeout will be used in case there is an UDP stream
	detected.
```

test/e2e/network/conntrack.go
wojtek-t: you can probably list pods from kube-system namespace and use the kube-proxy regexp to match the name; not super nice but should work

test/e2e/network/conntrack.go
wojtek-t: I don't have strong opinion here, but if we are unlucky and are missing couple seconds, that different may matter. But we may also start with what you have and if that won't help iterate further.

aojea: yeah, let's try to hit first one  failure in this PR before merging

aojea:
so, it does not want to fail now :rofl: 

/test pull-kubernetes-e2e-gce

test/e2e/network/conntrack.go
aojea: done

test/e2e/network/conntrack.go
aojea: I'm going to run it with a temp commit to see the results, if is ok I'd like to merge because seems is hard to catch a failure in this PR

wojtek-t: Yes - let's merge.
Though, please add a TODO with the issue number to get rid of it once this is debugged and fixed.

test/e2e/network/conntrack.go
```diff
@@ -268,4 +269,26 @@ var _ = SIGDescribe("Conntrack", func() {
...
+		for _, pod := range pods.Items {
+			if strings.Contains(pod.Name, "kube-proxy") {
+				stdout, err := framework.RunHostCmd(namespace, pod.Name, cmd)
+				if err != nil {
+					framework.Logf("Failed to dump conntrack table of node %s: %v", pod.Spec.NodeName, err)
+				}
+				framework.Logf("conntrack table of node %s: %s", pod.Spec.NodeName, stdout)
+			}
+		}
+		// }
```
wojtek-t, line 292: nit: remove

aojea: this is a temp commit

test/e2e/network/conntrack.go
```diff
@@ -268,4 +269,26 @@ var _ = SIGDescribe("Conntrack", func() {
 			framework.Failf("Failed to connect to backend 2")
 		}
 	})
+
+	ginkgo.AfterEach(func() {
+		// if ginkgo.CurrentGinkgoTestDescription().Failed {
+		// Dump conntrack table of each node for troubleshooting using the kube-proxy pods
+		namespace := "kube-system"
+		pods, err := fr.ClientSet.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{})
```
aojea, line 277: why is this panicking?
```
test/e2e/network/conntrack.go:273
Test Panicked
GOROOT/src/runtime/panic.go:212

Panic: runtime error: invalid memory address or nil pointer dereference

Full stack:
k8s.io/kubernetes/test/e2e/network.glob..func1.5()
	test/e2e/network/conntrack.go:277 +0x3e
```

wojtek-t: The only thing that comes to my mind is that fr is nil...

test/e2e/network/conntrack.go
```diff
@@ -117,6 +118,10 @@ var _ = SIGDescribe("Conntrack", func() {
 
 	ginkgo.It("should be able to preserve UDP traffic when server pod cycles for a NodePort service", func() {
 
+		// TODO: skip if test succeeds
```
wojtek-t, line 121: TODO(#91236): Remove once the test is debugged and fixed.

Same below.

aojea: done

test/e2e/network/conntrack.go
```diff
@@ -268,4 +275,26 @@ var _ = SIGDescribe("Conntrack", func() {
 			framework.Failf("Failed to connect to backend 2")
 		}
 	})
+
```
wojtek-t, line 278: nit: remove this line

wojtek-t (commented):
/approve

test/e2e/network/conntrack.go
aojea: done

test/e2e/network/conntrack.go
aojea: ginkgo ?? :man_shrugging:

wojtek-t:
/lgtm

[Assuming test won't fail]

PR #91833, opened 2020-06-05
Structured logging for Apiserver [Issue #1602 Structured logging]

AkashSarda:
```release-note
NONE
```

**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #92424, opened 2020-06-23
Improve performance of http access logs

serathius:
PR fixes performance regressions introduced when migrating `Infof` to `InfoS` https://github.com/kubernetes/kubernetes/pull/91833.  

Fixes:
* Don't call LogArgs if log will not be written due low verbosity
* Create separate slice for hijacked to avoid append on main path
* Shorten log message as this log is to common to be verbose

Improvement in benchmark of WithLogging function (old- current master, new - with this change)
```
name           old time/op    new time/op    delta
WithLogging-4    4.95µs ± 3%    3.52µs ± 1%  -28.80%  (p=0.000 n=10+8)

name           old alloc/op   new alloc/op   delta
WithLogging-4    2.93kB ± 0%    1.22kB ± 0%  -58.45%  (p=0.000 n=10+9)

name           old allocs/op  new allocs/op  delta
WithLogging-4      32.0 ± 0%      20.0 ± 0%  -37.50%  (p=0.000 n=10+10)
```

Difference compared to code before introducing structured logging (old - before, new - with this change)
```
name           old time/op    new time/op    delta
WithLogging-4    3.41µs ± 3%    3.52µs ± 1%   +3.29%  (p=0.000 n=10+8)

name           old alloc/op   new alloc/op   delta
WithLogging-4    1.09kB ± 0%    1.22kB ± 0%  +11.75%  (p=0.000 n=10+9)

name           old allocs/op  new allocs/op  delta
WithLogging-4      16.0 ± 0%      20.0 ± 0%  +25.00%  (p=0.000 n=10+10)
```

/kind cleanup

```release-note
NONE
```

/cc @jkaniuk

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/server/httplog/httplog.go
+++ b/staging/src/k8s.io/apiserver/pkg/server/httplog/httplog.go
@@ -85,7 +85,9 @@ func WithLogging(handler http.Handler, pred StacktracePred) http.Handler {
 		rl := newLogged(req, w).StacktraceWhen(pred)
 		req = req.WithContext(context.WithValue(ctx, respLoggerContextKey, rl))
 
-		defer func() { klog.V(3).InfoS("Received HTTP request", rl.LogArgs()...) }()
+		if klog.V(3).Enabled() {
+			defer func() { klog.InfoS("HTTP", rl.LogArgs()...) }()
+		}
 		handler.ServeHTTP(rl, req)
 	})
 }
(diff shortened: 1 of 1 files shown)


staging/src/k8s.io/apiserver/pkg/server/httplog/httplog.go
```diff
@@ -85,7 +85,11 @@ func WithLogging(handler http.Handler, pred StacktracePred) http.Handler {
 		rl := newLogged(req, w).StacktraceWhen(pred)
 		req = req.WithContext(context.WithValue(ctx, respLoggerContextKey, rl))
 
-		defer func() { klog.V(3).InfoS("Received HTTP request", rl.LogArgs()...) }()
+		defer func() {
```
jkaniuk, line 88: you could replace defer line with if line

jkaniuk: or go back to old Log() method with the check and logging inside

serathius: Do you mean like 
```
if klog.V(3).Enabled() {
			defer func() {klog.InfoS("HTTP", rl.LogArgs()...)}()
		}
```
?

staging/src/k8s.io/apiserver/pkg/server/httplog/httplog.go
```diff
@@ -85,7 +85,9 @@ func WithLogging(handler http.Handler, pred StacktracePred) http.Handler {
 		rl := newLogged(req, w).StacktraceWhen(pred)
 		req = req.WithContext(context.WithValue(ctx, respLoggerContextKey, rl))
 
-		defer func() { klog.V(3).InfoS("Received HTTP request", rl.LogArgs()...) }()
+		if klog.V(3).Enabled() {
+			defer func() {klog.InfoS("HTTP", rl.LogArgs()...)}()
```
mm4tt, line 89:
```
defer klog.InfoS("HTTP", rl.LogArgs()...)
```
?

staging/src/k8s.io/apiserver/pkg/server/httplog/httplog.go
jkaniuk: yes

staging/src/k8s.io/apiserver/pkg/server/httplog/httplog.go
jkaniuk: that would change the moment of running LogArgs, is that indifferent?

serathius: LogArgs measures time for request, we cannot use 
```
defer klog.InfoS("HTTP", rl.LogArgs()...)
```
as this would evaluate ` rl.LogArgs()` immidietly

liggitt:
/test pull-kubernetes-verify

master verify is failing on go1.14 fmt/vet/staticcheck issues, fixed in https://github.com/kubernetes/kubernetes/pull/92438

Issue #56138, opened 2017-11-21
ESIPP [Slow] e2e tests failing on 5000 node cluster jobs


porridge:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

This is a followup from #52495 filed separately now that we're not tracking this as part of 1.9 release. The tests have been disabled with [DisabledForLargeClusters] tag for now, and this fact will be mentioned in the scalability validation report for v1.9.

This issue tracks fixing the test so the tag can be removed and the test ran again in the large cluster.

/assign @MrHohn 
@kubernetes/sig-network-bugs 
@kubernetes/sig-scalability-bugs 
/kind e2e-test-failure

jberkus:
Is this e2e test still broken?

MrHohn:
@jberkus Sorry for the confusion here. Currently this e2e test is disabled on extra large cluster due to some scalability issues with the load balancer related components. At the moment the issues haven't been completely resolved yet.

jberkus:
ok, adding priority so that I know it's not a 1.13 blocker.

/priority important-longterm

yastij:
cc @lbernail - this is why the relevant tests for #72432 are not part of the e2e suite.

MrHohn:
This issue is still valid --- some action items on top of my head:
- Figure out a plan to make service controller properly handle operations in parallel.
- Check on the status for cloud providers (especially GCP, which the test ran against) on handling large number of nodes as LB backends.

/remove-triage unresolved

cblecker:
/reopen
Reopening as #89421 is being reverted in #90171

liggitt:
these tests are flaking in https://testgrid.k8s.io/sig-release-master-informing#gce-master-scale-correctness&width=20

xref https://github.com/kubernetes/kubernetes/issues/90955

liggitt:
cc @wojtek-t, since these were enabled by #90305

wojtek-t:
Yeah - though these failures doesn't seem to be scale related, they seem to be test related to me. 
This basically means nodes without public ips:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/service.go#L2928

It's pretty much the same thing as https://github.com/kubernetes/kubernetes/issues/90047 (just for a different test).
https://github.com/kubernetes/kubernetes/pull/92163 should be addressing these kinds of errors.

[Not a release blocker for sure]

wojtek-t:
[...]

PR #92427, opened 2020-06-23
Improve LB session affinity tests

wojtek-t:
In preparation of enabling those tests for LB session affinity (5k-node) clusters.

Ref: #56138

```release-note
NONE
```

### DIFF
--- a/test/e2e/network/service.go
+++ b/test/e2e/network/service.go
@@ -101,42 +101,81 @@ var (
 // portsByPodName is a map that maps pod name to container ports.
 type portsByPodName map[string][]int
 
+// affinityCheckFromPod returns interval, timeout and function pinging the service and
+// returning pinged hosts for pinging the service from execPod.
+func affinityCheckFromPod(execPod *v1.Pod, serviceIP string, servicePort int) (time.Duration, time.Duration, func() []string) {
+	timeout := AffinityTimeout
+	// interval considering a maximum of 2 seconds per connection
+	interval := 2 * AffinityConfirmCount * time.Second
+
+	serviceIPPort := net.JoinHostPort(serviceIP, strconv.Itoa(servicePort))
+	curl := fmt.Sprintf(`curl -q -s --connect-timeout 2 http://%s/`, serviceIPPort)
+	cmd := fmt.Sprintf("for i in $(seq 0 %d); do echo; %s ; done", AffinityConfirmCount, curl)
+	getHosts := func() []string {
+		stdout, err := framework.RunHostCmd(execPod.Namespace, execPod.Name, cmd)
+		if err != nil {
+			framework.Logf("Failed to get response from %s. Retry until timeout", serviceIPPort)
+			return nil
+		}
+		return strings.Split(stdout, "\n")
+	}
+
+	return interval, timeout, getHosts
+}
+
+// affinityCheckFromTest returns interval, timeout and function pinging the service and
+// returning pinged hosts for pinging the service from the test itself.
+func affinityCheckFromTest(cs clientset.Interface, serviceIP string, servicePort int) (time.Duration, time.Duration, func() []string) {
+	interval := 2 * time.Second
+	timeout := e2eservice.GetServiceLoadBalancerPropagationTimeout(cs)
+
+	params := &e2enetwork.HTTPPokeParams{Timeout: 2 * time.Second}
+	getHosts := func() []string {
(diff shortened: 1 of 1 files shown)


wojtek-t:
`service is not reachable within 2m0s timeout on endpoint affinity-clusterip-transition:80 over TCP protocol` doesn't seem to be related to my change

/retest

danwinship (commented):
It looks like these tests were marked `[DisabledForLargeClusters]` because there were problems with NodePort services in very large clusters... and it doesn't seem like this patch addresses that problem? Why exactly is this needed for re-enabling the tests?

wojtek-t:
> It looks like these tests were marked [DisabledForLargeClusters] because there were problems with NodePort services in very large clusters... and it doesn't seem like this patch addresses that problem? Why exactly is this needed for re-enabling the tests?

We enabled these tests couple weeks ago and they were flaky back then (mostly timeouts). I think that may contribute to improving them (I would like to re-enable them again). So I actually think it may potentially help.

test/e2e/network/service.go
```diff
@@ -110,13 +110,15 @@ func checkAffinity(cs clientset.Interface, execPod *v1.Pod, serviceIP string, se
 	serviceIPPort := net.JoinHostPort(serviceIP, strconv.Itoa(servicePort))
 	curl := fmt.Sprintf(`curl -q -s --connect-timeout 2 http://%s/`, serviceIPPort)
 	cmd := fmt.Sprintf("for i in $(seq 0 %d); do echo; %s ; done", AffinityConfirmCount, curl)
+
 	timeout := AffinityTimeout
+	// interval considering a maximum of 2 seconds per connection
+	interval := 2 * AffinityConfirmCount * time.Second
 	if execPod == nil {
 		timeout = e2eservice.GetServiceLoadBalancerPropagationTimeout(cs)
+		interval = 2 * time.Second
```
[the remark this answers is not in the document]
wojtek-t: @danwinship - would splitting these into two somewhat separate functions address this concern?

test/e2e/network/service.go
```diff
@@ -101,42 +101,87 @@ var (
...
-	if execPod == nil {
-		timeout = e2eservice.GetServiceLoadBalancerPropagationTimeout(cs)
+	var pollingArgs affinityCheckerPollingArgs
+	if execPod != nil {
+		pollingArgs = affinityCheckFromPod(execPod, serviceIP, servicePort)
+	} else {
+		pollingArgs = affinityCheckFromTest(cs, serviceIP, servicePort)
 	}
+
+	interval, timeout, getHosts := pollingArgs()
```
wojtek-t, line 165: @danwinship - thanks a lot for the first pass of review; I refactored this test a bit and hopefully it's much cleaner now.
PTAL

wojtek-t:
@danwinship - friendly ping

test/e2e/network/service.go
```diff
@@ -101,42 +101,87 @@ var (
...
+		if err != nil {
+			framework.Logf("Failed to get response from %s. Retry until timeout", serviceIPPort)
+			return nil
+		}
+		return strings.Split(stdout, "\n")
+	}
+
+	return func() (time.Duration, time.Duration, func() []string) {
+		return interval, timeout, getHosts
+	}
```
danwinship, line 128: weird to have these return a function that returns those arguments, rather than just returning the arguments themselves

test/e2e/network/service.go
```diff
@@ -147,6 +192,7 @@ func checkAffinity(cs clientset.Interface, execPod *v1.Pod, serviceIP string, se
 			return false
 		}
 		if !trackerFulfilled {
+			serviceIPPort := net.JoinHostPort(serviceIP, strconv.Itoa(servicePort))
 			checkAffinityFailed(tracker, fmt.Sprintf("Connection to %s timed out or not enough responses.", serviceIPPort))
```
danwinship, line 196: (you could just remove `serviceIPPort` from the error message... it's obvious from context)

test/e2e/network/service.go
wojtek-t: the function itself is slightly longer then, but I'm fine with changing too.
Done

test/e2e/network/service.go
wojtek-t: done

wojtek-t (commented):
@danwinship - thanks; PTAL

PR #38342, opened 2016-12-08
Make StrategicPatch delete all matching maps in a merging list

mengqiy:
fixes #38332

```release-note
NONE
```

cc: @lavalamp @pwittrock

PR #92437, opened 2020-06-23
delete one key per delete directive in strategicPatch

### LINKED ISSUE #58477: Duplicated environment variable in deployment disappears completely when fixed
/kind bug

**What happened**:
For `deployment` resource:

A container has defined environment variable with name `x` that is duplicated (there are two env vars with the same name, the value is also the same).

When you fix the `deployment` resource descriptor so that environment variable with name `x` appears only once and push it with `kubectl apply`, deployment with no environment variable named `x` is created and therefore no environment variable named `x` is passed to replica set and pods.

After fixing the `deployment`, environment variable with name `x` is defined in the `deployment` once .

1. create deployment with container with duplicated environment variable
1. `kubectl apply` it
1. fix deployment removing one of duplicated environment variable definitions
1. `kubectl apply` it
1. `kubectl get deployment/your-deployment -o yaml` prints deployment without 

nope

**Environment**:
`Client Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.1", GitCommit:"3a1c9449a956b6026f075fa3134ff92f7d55f812", GitTreeState:"clean", BuildDate:"2018-01-04T20:00:41Z", GoVersion:"go1.9.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.1", GitCommit:"3a1c9449a956b6026f075fa3134ff92f7d55f812", GitTreeState:"clean", BuildDate:"2018-01-04T11:40:06Z", GoVersion:"go1.9.2", Compiler:"gc", Platform:"linux/amd64"}`
- Cloud provider or hardware configuration: private Kubernetes cluster
- OS (e.g. from /etc/os-release): N/A
- Install tools: N/A
- Others: N/A

sweeneyb:
attempts to address https://github.com/kubernetes/kubernetes/issues/58477 by only deleting one element per directive

**What type of PR is this?**
/kind bug

A patch to delete an accidental duplicate key in a resource definition ends up deleting all of the keys.  This requires a workaround where a person removes all of the keys, then adds one of them back.  Instead, the merged resource definition should reflect what was submitted into the system.

Fixes #58477

```release-note
Fixed a bug where `kubectl apply` removes all entries when attempting to remove a single duplicated entry in a persisted object
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/util/strategicpatch/patch.go
+++ b/staging/src/k8s.io/apimachinery/pkg/util/strategicpatch/patch.go
@@ -1478,7 +1478,7 @@ func mergeSliceWithSpecialElements(original, patch []interface{}, mergeKey strin
 				mergeValue, ok := typedV[mergeKey]
 				if ok {
 					var err error
-					original, err = deleteMatchingEntries(original, mergeKey, mergeValue)
+					original, err = deleteMatchingEntry(original, mergeKey, mergeValue)
 					if err != nil {
 						return nil, nil, err
 					}
@@ -1501,17 +1501,14 @@ func mergeSliceWithSpecialElements(original, patch []interface{}, mergeKey strin
 	return original, patchWithoutSpecialElements, nil
 }
 
-// delete all matching entries (based on merge key) from a merging list
-func deleteMatchingEntries(original []interface{}, mergeKey string, mergeValue interface{}) ([]interface{}, error) {
-	for {
-		_, originalKey, found, err := findMapInSliceBasedOnKeyValue(original, mergeKey, mergeValue)
-		if err != nil {
-			return nil, err
-		}
+// delete matching entry (based on merge key) from a merging list
+func deleteMatchingEntry(original []interface{}, mergeKey string, mergeValue interface{}) ([]interface{}, error) {
+	_, originalKey, found, err := findMapInSliceBasedOnKeyValue(original, mergeKey, mergeValue)
+	if err != nil {
+		return nil, err
+	}
 
-		if !found {
-			break
-		}
+	if found {
 		// Delete the element at originalKey.
 		original = append(original[:originalKey], original[originalKey+1:]...)
 	}
--- a/staging/src/k8s.io/apimachinery/pkg/util/strategicpatch/patch_test.go
+++ b/staging/src/k8s.io/apimachinery/pkg/util/strategicpatch/patch_test.go
@@ -375,6 +375,27 @@ testCases:
       $patch: replace
     modified:
       other: a
+  - description: de-duplicate entries in a merging list
+    original:
+      mergingList:
+        - name: 1
+        - name: 1
+        - name: 2
+          value: a
+        - name: 3
+        - name: 3
+    twoWay:
+      mergingList:
+        - name: 1
+          $patch: delete
+        - name: 3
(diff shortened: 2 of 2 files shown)


cblecker:
@sweeneyb Looks like some of the failures are unrelated to this PR. I'm having a look at what is going on.

sweeneyb:
@cblecker thanks.  I'll defer retests until tomorrow or I hear something.  I can see one test dying being spurious.  But it's failed a couple of times now.  I don't want to put load on the CI unnecessarily.

cblecker:
The above failure is legitimate. You need to run `hack/update-gofmt.sh`

BenTheElder:
All required tests are passing. 
image

The IPv6 test is soft-blocking, and seems to be having issues unrelated to this PR.

BenTheElder:
> tide Pending — Not mergeable. Needs approved, lgtm labels.

this status is also the relevant one for knowing what is overall required (it could be more obvious...), this PR won't merge until approved and lgtm-ed, but it's already good on the tests.

tide is the kubernetes merge robot.

sweeneyb:
@BenTheElder Thanks for the context.  It is pretty obvious in retrospect.  There's a lot new, here, so I'm learning where to look in a lot of ways.  Thanks again.

BenTheElder:
FWIW I think it's a _lot_ and not anywhere near as obvious as it should be.

mengqiy:
This PR changes the behavior of delete directive`$patch: delete`.
Before: it deletes all entries that match the mergeKey.
After: it deletes the first entry that matches the mergeKey.

The later approach doesn't solve all of the problems.
Suppose user has a list like this:
```
- name: foo
  anotherField: abc
- name: foo
  anotherField: xyz
```
If the user wants to remove
```
- name: foo
  anotherField: xyz
```
there is no great way to express that.
If using SMP, the user needs to delete the 1st entry, delete the 2nd entry and then add the 1st entry back.

I'm not sure if the new behavior is better than the existing behavior.
@lavalamp @liggitt @apelisse WDYT?

Some background:
@lavalamp suggested we should remove all matching entries: https://github.com/kubernetes/community/pull/140#issuecomment-265598220
It is implemented in https://github.com/kubernetes/kubernetes/pull/38342

issssu:
> This PR changes the behavior of delete directive`$patch: delete`. Before: it deletes all entries that match the mergeKey. After: it deletes the first entry that matches the mergeKey.
> 
> The later approach doesn't solve all of the problems. Suppose user has a list like this:
> 
> ```
> - name: foo
>   anotherField: abc
> - name: foo
>   anotherField: xyz
> ```
> 
> If the user wants to remove
> 
> ```
> - name: foo
>   anotherField: xyz
> ```
> 
> there is no great way to express that. If using SMP, the user needs to delete the 1st entry, delete the 2nd entry and then add the 1st entry back.
> 
> I'm not sure if the new behavior is better than the existing behavior. @lavalamp @liggitt @apelisse WDYT?
> 
> Some background: @lavalamp suggested we should remove all matching entries: [kubernetes/community#140 (comment)](https://github.com/kubernetes/community/pull/140#issuecomment-265598220) It is implemented in #38342

maybe we can delete last one, rather than the first one. Because when dulplicate envs exist, the first one is active.

PR #88638, opened 2020-02-28
Update Golang to v1.14.4

justaugustus:
**What type of PR is this?**
/kind cleanup
/area dependency release-eng
/sig release

**What this PR does / why we need it**:
Update Golang to v1.14.4

Signed-off-by: Stephen Augustus <saugustus@vmware.com>

/assign @liggitt @cblecker @dims @BenTheElder @listx 
cc: @kubernetes/release-engineering 
ref: https://github.com/kubernetes/release/issues/1216

**Does this PR introduce a user-facing change?**:

```release-note
Update Golang to v1.14.4
```

Issue #89719, opened 2020-04-01
[Flaky Test] pull-kubernetes-integration / run_kubectl_apply_tests 


ScrapCodes:
**Which jobs are flaking**:
pull-kubernetes-integration
**Which test(s) are flaking**:
run_kubectl_apply_tests 
**Testgrid link**:
https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-integration&include-filter-by-regex=run_kubectl_apply_tests&width=5

**Reason for failure**:
A test has a precondition that no pods should be up, but, previous test's deployment did not fully delete. As a result precondition check fails.

**Anything else we need to know**:
- links to go.k8s.io/triage appreciated
https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&text=run_kubectl_apply_tests
- links to specific failures in spyglass appreciated

https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/88364/pull-kubernetes-integration/1245223371679993859
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/89760/pull-kubernetes-integration/1245616723650940928

ScrapCodes:
Pasting the failed tests' output, for easy reference:

```
 +++ [0401 05:49:00] Testing kubectl apply
apply.sh:30: Successful get pods {{range.items}}{{.metadata.name}}:{{end}}: 
(Bpod/test-pod created
apply.sh:34: Successful get pods test-pod {{.metadata.labels.name}}: test-pod-label
(Bpod "test-pod" deleted
apply.sh:42: Successful get pods {{range.items}}{{.metadata.name}}:{{end}}: 
(Bpod/test-pod created
apply.sh:46: Successful get pods test-pod {{.metadata.labels.name}}: test-pod-label
(Bpod/test-pod configured (dry run)
pod/test-pod configured (server dry run)
pod/test-pod configured
pod "test-pod" deleted
apply.sh:62: Successful get deployments {{range.items}}{{.metadata.name}}:{{end}}: 
(BI0401 05:49:02.156109   52987 controller.go:606] quota admission added evaluator for: deployments.apps
deployment.apps/test-deployment-retainkeys created
I0401 05:49:02.170412   52987 controller.go:606] quota admission added evaluator for: replicasets.apps
[...]

PR #92438, opened 2020-06-23
go1.14 fixup

liggitt:
**What type of PR is this?**
/kind cleanup

follow-up to https://github.com/kubernetes/kubernetes/pull/88638

* Updates required go version to 1.14.4
* Fixes gofmt/govet errors
* Fixes a test/cmd error message check to match what go 1.14 net/http produces
* Updates staticcheck to a level that works with go1.14 (and tolerates newly found staticcheck errors)
* Updates go.mod files to specify go 1.14
* Updates hack/{update-vendor.sh,update-vendor-licenses.sh,lint-dependencies.sh} to work with go 1.14 vendor behavior

```release-note
go1.14.4 is now the minimum version required for building Kubernetes
```

/sig release


hack/lib/golang.sh
```diff
@@ -468,7 +468,7 @@ EOF
   local go_version
   IFS=" " read -ra go_version <<< "$(go version)"
   local minimum_go_version
-  minimum_go_version=go1.13.4
+  minimum_go_version=go1.14.4
```
justaugustus, line 471: Should this always be the latest go version we've bumped to?
If so, can you add an entry in build/dependencies.yaml for this path/regex?

justaugustus (changes requested):
@liggitt -- One blocking question about dependencies.yaml

And then, non-blocking: what's the signal on when/if we should make these changes in staging? Should this be part of the go bump docs?

/hold

hack/lib/golang.sh
BenTheElder: it should, BUT that would require the image bump to go in before the PR, FWIW

justaugustus: Ooof. Chicken and egg! Thanks for clarifying, @BenTheElder!
/hold cancel

liggitt:
/hold

 still working through staticcheck, golang.org/x/... deps, and hack/update-vendor failures

liggitt:
looks like the updated gazelle produces BUILD files with aix/illumos targets CI is unhappy with... dropping back to the previous 1.19.1 level

cblecker (approved):
/lgtm
/approve

hack/lib/golang.sh
BenTheElder: we still have this chicken and egg either way, the question is just which repo we want to merge first :man_shrugging:

test/cmd/request-timeout.sh
```diff
@@ -39,7 +39,7 @@ run_kubectl_request_timeout_tests() {
 
   ## check --request-timeout on 'get pod' with --watch
   output_message=$(kubectl get pod valid-pod --request-timeout=1 --watch --v=5 2>&1)
-  kube::test::if_has_string "${output_message}" 'Timeout exceeded while reading body'
+  kube::test::if_has_string "${output_message}" 'Timeout'
```
BenTheElder, line 42: ugh

BenTheElder:
CI looking better this time.

liggitt:
awesome, platform-specific staticcheck differences... fixed up in https://github.com/kubernetes/kubernetes/compare/b55f384056dfd975b760fb1ae7264e8320dca287..d9bb0b8ee18fe68654ef1a68611f905859b3b7bc

liggitt:
grr
https://github.com/kubernetes/kubernetes/issues/89719

/retest

AkihiroSuda:
Why was support for 1.13 dropped?

BenTheElder:
Kubernetes has long supported only one specific go version.

AkihiroSuda:
I suggest continuing support for Go 1.13 exceptionally.

Go 1.14 introduced a breaking change that requires retrying syscall executions on receiving EINTR: https://golang.org/doc/go1.14#runtime

Because of this change, a lot of runtime components including containerd, runc (<= rc90), Moby, and CNI haven't yet migrated to Go 1.14.
Requiring Go 1.14 as a hard dependency is likely to break downstream build pipelines.

Also, I even doubt that Kubernetes has really completed support for Go 1.14.
I can't find EINTR handlers in the repo (as of  aadaa5d6a917fcb354bce80a0bd1830c95cfbcc3) except in a single function:
```bash
$ git grep EINTR | grep -v ^vendor
pkg/kubelet/eviction/threshold_notifier_linux.go:               if err == unix.EINTR {
```

( Fortunately migration to Go 1.15 would be much smoother, as Go 1.15 can handle most of the EINTR stuffs automatically: https://tip.golang.org/doc/go1.15#os )

liggitt:
Given the support window of go1.13 (it will go out of support before Kubernetes 1.19 is released), I think k8s 1.19 will stay on go1.14+

I opened https://github.com/kubernetes/kubernetes/pull/92611 as an experiment for what it could look like to isolate the bits that actually required go1.14 to let someone who wanted to build with k8s libraries build with go1.13. Since that change appears to be very small, I'd be amenable to keeping that in place until we move to go1.15.

BenTheElder:
Downstream build pipelines should not be depending on a single host go
binary, we make no guarantees about what timeframe we will upgrade go in.
Any such pipeline may again be broken in the future.

The containerized and bazel builds both offer building using go binaries
managed by the project's build.

liggitt:
> Downstream build pipelines should not be depending on a single host go binary, we make no guarantees about what timeframe we will upgrade go in.

I'm less concerned about build pipelines building kubernetes components, and am more sympathetic to something using the published k8s libraries.

PR #92441, opened 2020-06-23
WIP Ensure init and ephemeral containers are reclaimed

derekwaynecarr:
**What type of PR is this?**
/kind bug

Fixes utility methods to ensure init and ephemeral containers are reclaimed.

The `PodResourcesAreReclaimed` is a a seat-belt for the kubelet.  The existing method gates on the cleanup of the pod level cgroup which can only be deleted when its children are deleted, so this should just result in improved logging for debug.

The `podAndContainerResourcesAreTerminal` does gate telling internal status manager that it should terminate the pod, and there was a minor bug in that situation where ephemeral container status not reporting a proper error status when no status could be found. 

As these methods could be called elsewhere in future evolutions of kubelet, it's important they covered all scenarios.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kubelet/kubelet_pods.go
+++ b/pkg/kubelet/kubelet_pods.go
@@ -895,7 +895,11 @@ func (kl *Kubelet) podAndContainersAreTerminal(pod *v1.Pod) (containersTerminal,
 	// fails) or due to external events like deletion or eviction. A terminal pod *should* have no running
 	// containers, but to know that the pod has completed its lifecycle you must wait for containers to also
 	// be terminal.
-	containersTerminal = notRunning(status.ContainerStatuses)
+	containersTerminal = notRunning(status.InitContainerStatuses) && notRunning(status.ContainerStatuses)
+	if utilfeature.DefaultFeatureGate.Enabled(features.EphemeralContainers) {
+		containersTerminal = containersTerminal && notRunning(status.EphemeralContainerStatuses)
+	}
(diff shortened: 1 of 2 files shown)


pkg/kubelet/kubelet_pods.go
```diff
@@ -895,7 +895,11 @@ func (kl *Kubelet) podAndContainersAreTerminal(pod *v1.Pod) (containersTerminal,
 	// fails) or due to external events like deletion or eviction. A terminal pod *should* have no running
 	// containers, but to know that the pod has completed its lifecycle you must wait for containers to also
 	// be terminal.
-	containersTerminal = notRunning(status.ContainerStatuses)
+	containersTerminal = notRunning(status.InitContainerStatuses) && notRunning(status.ContainerStatuses)
+	if utilfeature.DefaultFeatureGate.Enabled(features.EphemeralContainers) {
+		containersTerminal = notRunning(status.EphemeralContainerStatuses)
```
dashpole, line 900: I think we want containersTerminal = containersTerminal & notRunning ...

derekwaynecarr:
/hold 

I want to add unit testing around this, but do not have time to finish this afternoon.

dashpole:
looks good to me now.

@SergeyKanzhelev pod cleanup is an area that could use better unit testing if you are interested.  LMK if you have questions

PR #92442, opened 2020-06-23
Respect grace period when removing mirror pod

### LINKED ISSUE #90999: Mirror pod deletion should respect grace period
**What happened**:
Currently when the mirror pod is deleted, it is deleted without grace period.

When static pod yaml was removed from the manifests directory, @sjenning observed
```
Jul 29 20:54:44 master-2 hyperkube[1526]: I0729 20:54:44.122924    1526 kubelet.go:1904] SyncLoop (REMOVE, "file"): "kube-apiserver-master-2_openshift-kube-apiserver(60d66dd8b81c2a4756d2e2f69ec1e3fc)"
Jul 29 20:54:44 master-2 hyperkube[1526]: I0729 20:54:44.401404    1526 mirror_client.go:93] Deleting a mirror pod "kube-apiserver-master-2_openshift-kube-apiserver" (uid (*types.UID)(nil))
Jul 29 20:54:44 master-2 hyperkube[1526]: I0729 20:54:44.435138    1526 kubelet_pods.go:1695] Orphaned pod "60d66dd8b81c2a4756d2e2f69ec1e3fc" found, but volumes not yet removed.  Reducing cpu to minimum
Jul 29 20:54:44 master-2 hyperkube[1526]: I0729 20:54:44.435544    1526 kubelet.go:1910] SyncLoop (DELETE, "api"): "kube-apiserver-master-2_openshift-kube-apiserver(08b81ce8-b243-11e9-a38f-fa163e8871d1)"
Jul 29 20:54:44 master-2 hyperkube[1526]: I0729 20:54:44.441528    1526 kubelet.go:1904] SyncLoop (REMOVE, "api"): "kube-apiserver-master-2_openshift-kube-apiserver(08b81ce8-b243-11e9-a38f-fa163e8871d1)"
```

The cause has two aspects:
* DeleteOrphanedMirrorPods doesn't consider grace period from pod spec
* cleanupOrphanedPodCgroups prematurely cleans up the underlying cgroup

The desired behavior is that the kubelet honors the pod.spec.terminationGracePeriodSeconds in the yaml file when turning down the container, and only deletes the pod from the APIServer after the container is down.

This would allow the logs to be accessible via the apiserver during pod termination.

* remove static pod of kube-apiserver from /etc/kubernetes/manifests
* monitor the mirror pod in the API and how it goes away immediately
* monitor `crictl ps` and how the container disappears immediately.

Please see the 'Test Strategy' subsection of the description of #91453 for the failure case of e2e test where grace period is not respected.

Without the fix, the new e2e test would fail with the following assertion:
```
[k8s.io] MirrorPodWithGracePeriod when create a mirror pod mirror pod termination should satisfy grace period when static pod is deleted [NodeConformance] expand_less  4s
_output/local/go/src/k8s.io/kubernetes/test/e2e_node/mirror_pod_grace_test.go:54
Unexpected error:
    <*errors.StatusError | 0xc0003b35e0>: {
        ErrStatus: {
            ListMeta: {
                SelfLink: "",
                ResourceVersion: "",
                Continue: "",
                RemainingItemCount: nil,
            },
            Status: "Failure",
            Message: "pods \"static-pod-9ba82792-0187-43a7-be3d-ffb5db011631-tmp-node-e2e-59ba742a-cos-stable-81-12871-103-0\" not found",
            Reason: "NotFound",
            Details: {
                Name: "static-pod-9ba82792-0187-43a7-be3d-ffb5db011631-tmp-node-e2e-59ba742a-cos-stable-81-12871-103-0",
                Group: "",
                Kind: "pods",
                UID: "",
                Causes: nil,
                RetryAfterSeconds: 0,
            },
            Code: 404,
        },
    }
    pods "static-pod-9ba82792-0187-43a7-be3d-ffb5db011631-tmp-node-e2e-59ba742a-cos-stable-81-12871-103-0" not found
occurred
_output/local/go/src/k8s.io/kubernetes/test/e2e_node/mirror_pod_grace_test.go:72
```
Here is the assertion which expects to be able to retrieve the mirror pod within grace period:
```
				pod, err := f.ClientSet.CoreV1().Pods(ns).Get(context.TODO(), mirrorPodName, metav1.GetOptions{})
				framework.ExpectNoError(err)
```

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

tedyu:
**What type of PR is this?**
/kind bug

This PR respects the grace period for mirror pod termination.
Without this fix, the mirror pod would be terminated prior to container(s) having gracefully shut down.
For api server pod, the difference can be as wide as ~70 seconds.

There is no structural modification to pod_manager.

**Overall Structure of the PR**
We use additional map, mirrorPodTerminationMap, to track pod termination in Kubelet. 

The patch refactors deleteOrphanedMirrorPods() so that we wait for the killPod() call to complete before calling podManager.DeleteMirrorPod()
podKiller signals the completion of the killPod() call by setting timestamp in the corresponding entry of mirrorPodTerminationMap.

deleteOrphanedMirrorPods() is moved into Kubelet class because the fulfillment of the functionality is carried out in Kubelet func's.

In the current implementation, across kubelet restart, the grace period may not be respected (since the amount of time waited prior to kubelet restart is not persisted).

**Test Strategy**
On a patched 1.18 master node, verified that moving manifest file out of /etc/kubernetes/manifests/ directory wouldn't result in immediate termination of relevant containers:
```
# crictl ps | grep apiserver
b5c85620df162       65065b8f3de3b401d55f7de1c91bcd7e3f4924b75208ab2109ce71cbd7f2c84d                                                         2 minutes ago       Running             kube-apiserver                                0                   915dfe8935190
467776050a176       09332ba901dec992d16fb1e860b34799aa7219eef076ff5fc196c0738df7baa7                                                         26 minutes ago      Running             openshift-apiserver                           0                   8f92fe1f8c7d2
3c6711f8658dc       5647937ebb1588bed17059dbf86a9d6a889ad18c4ee2b653ae8ec6577b8c153a                                                         30 minutes ago      Running             kube-apiserver-operator                       1                   b6642bc648023
```
After kube-apiserver container completes shutdown (~70 seconds, within the grace period of 135 seconds), the mirror pod is removed.

I introduced new e2e test, test/e2e_node/mirror_pod_grace_period_test.go
Without the fix, the new e2e test would fail with the following assertion:
```
[k8s.io] MirrorPodWithGracePeriod when create a mirror pod mirror pod termination should satisfy grace period when static pod is deleted [NodeConformance] expand_less  4s
_output/local/go/src/k8s.io/kubernetes/test/e2e_node/mirror_pod_grace_test.go:54
Unexpected error:
    <*errors.StatusError | 0xc0003b35e0>: {
        ErrStatus: {
            ListMeta: {
                SelfLink: "",
                ResourceVersion: "",
                Continue: "",
                RemainingItemCount: nil,
            },
            Status: "Failure",
            Message: "pods \"static-pod-9ba82792-0187-43a7-be3d-ffb5db011631-tmp-node-e2e-59ba742a-cos-stable-81-12871-103-0\" not found",
            Reason: "NotFound",
            Details: {
                Name: "static-pod-9ba82792-0187-43a7-be3d-ffb5db011631-tmp-node-e2e-59ba742a-cos-stable-81-12871-103-0",
                Group: "",
                Kind: "pods",
                UID: "",
                Causes: nil,
                RetryAfterSeconds: 0,
            },
            Code: 404,
        },
    }
    pods "static-pod-9ba82792-0187-43a7-be3d-ffb5db011631-tmp-node-e2e-59ba742a-cos-stable-81-12871-103-0" not found
occurred
_output/local/go/src/k8s.io/kubernetes/test/e2e_node/mirror_pod_grace_test.go:72
```
Here is the assertion :
```
				pod, err := f.ClientSet.CoreV1().Pods(ns).Get(context.TODO(), mirrorPodName, metav1.GetOptions{})
				framework.ExpectNoError(err)
```

The test failure (without the fix) was due to the mirror pod having been removed prior to expiration of grace period.

I have also added unit test for Kubelet#HandlePodCleanups() which exercises critical code path such as Kubelet#deleteOrphanedMirrorPods().

Fixes #90999

```release-note
The terminationGracePeriodSeconds from pod spec is respected for the mirror pod.
```

```docs

```


derekwaynecarr:
This helps review approaches side by side.

/assign @derekwaynecarr @sjenning 

Seth - can you take a look at this approach?

derekwaynecarr:
would like @sjenning to provide feedback as well.

/hold

tedyu:
Performed the following test using 1.18 node patched with this PR:
1. move kube-apiserver-pod.yaml out of manifest dir
2. verify with 'crictl ps | grep apise' that the container still runs
3. on Web UI, updated the pod YAML by adding an annotation
```
Operation cannot be fulfilled on pods "kube-apiserver-ip-10-0-137-140.ec2.internal": the object has been modified; please apply your changes to the latest version and try again
```
4. after ~70 seconds, verify that container is gone (via terminal) and pod is deleted (using UI)

Here was relevant portion from kubelet log:
```
Jun 24 04:25:54 ip-10-0-137-140 hyperkube[283398]: I0624 04:25:54.984562  283398 pod_manager.go:271] terminating with grace period 135 for "kube-apiserver-ip-10-0-137-140.ec2.internal_openshift-kube-apiserver" : expires at 2020-06-24 04:28:09.984553293 +0000 UTC m=+220.759681124
...
Jun 24 04:27:05 ip-10-0-137-140 hyperkube[283398]: I0624 04:27:05.585270  283398 kubelet_pods.go:1045] all containers have exited for "openshift-kube-apiserver" / "kube-apiserver-ip-10-0-137-140.ec2.internal" (b8503fdfea61bbef34f94831e73b900b)
```

tedyu:
Performed the following test using patched 1.18 master node:

1. move /etc/kubernetes/manifests/kube-apiserver-pod.yaml to /tmp
2. before kube-apiserver fully shuts down, stop kubelet
3. after a short pause, start kubelet
4. observe the api server mirror pod being terminated
5. move /tmp/kube-apiserver-pod.yaml back to manifest directory
6. observe containers come up and mirror pod run

Here was relevant snippet from kubelet log:
```
Jun 24 16:07:15 ip-10-0-138-125 hyperkube[660885]: W0624 16:07:15.360506  660885 kubelet_pods.go:1036] unable to find pod UID for "kube-apiserver-ip-10-0-138-125.ec2.internal_openshift-kube-apiserver", likely kubelet restarted
Jun 24 16:07:15 ip-10-0-138-125 hyperkube[660885]: I0624 16:07:15.360533  660885 mirror_client.go:123] Deleting a mirror pod "kube-apiserver-ip-10-0-138-125.ec2.internal_openshift-kube-apiserver" (uid (*types.UID)(nil))
```

derekwaynecarr:
thanks for the summary @tedyu , i am reviewing this now.

pkg/kubelet/kubelet_pods.go
```diff
@@ -1020,6 +1021,69 @@ func (kl *Kubelet) removeOrphanedPodStatuses(pods []*v1.Pod, mirrorPods []*v1.Po
...
+		return
+	}
+	// the mirror pod has been removed from api server, clean the entry for bookkeeping
+	kl.podManager.MarkPodTerminated(podFullName)
+}
+
+// deleteOrphanedMirrorPods checks whether grace period for orphaned mirror pod has elapsed.
+// Once the grace period has passed, podManager.DeleteMirrorPod() is called to delete mirror pod
+// from the API server
+func (kl *Kubelet) deleteOrphanedMirrorPods() {
```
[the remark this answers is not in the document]
derekwaynecarr: @tedyu I am wondering the following.

Can the logic in this method, and what we are attempting to track in pod manager be simplified to only call `deleteMirrorPod` if and only if the pod is not in the `podKillingCh`?  If the pod is in the `podKillingCh`, we know it hasn't actually been fully removed?  In this manner, we do not need to track termination phase in pod manager itself?

Am I mistaken, or do you agree that that would be a simplification?

pkg/kubelet/kubelet_pods.go
```diff
@@ -1020,6 +1021,69 @@ func (kl *Kubelet) removeOrphanedPodStatuses(pods []*v1.Pod, mirrorPods []*v1.Po
...
+	podFullNames := kl.podManager.GetOrphanedMirrorPodNames()
```
derekwaynecarr, line 1043: note to self, at this point:

`SyncLoop (REMOVE, "file")` record would have removed the pod from the pod manager
mirror pod record remains in pod manager itself

pkg/kubelet/kubelet_pods.go
```diff
@@ -1020,6 +1021,69 @@ func (kl *Kubelet) removeOrphanedPodStatuses(pods []*v1.Pod, mirrorPods []*v1.Po
...
+	for _, podFullName := range podFullNames {
+		start := time.Now()
+		uidWithGracePeriod, ok := kl.podManager.GetUIDForPodPendingTermination(podFullName)
+		if !ok {
+			klog.Warningf("unable to find pod UID for %q, likely kubelet restarted", podFullName)
+			kl.deleteMirrorPod(podFullName)
+			continue
+		}
+		allContainersExited := true
+		if start.Before(uidWithGracePeriod.Till) {
```
derekwaynecarr, line 1053: i agree that the kubelet restart scenario has no ideal solution, but i wonder if a simpler approach is the following.

could we avoid tracking this state information in pod manager by instead just saying "do not delete a pod if its present in the pod killing channel"?  The pod killing channel is the real thing you could still race against in this approach, and its the actual source of truth internally that the kubelet (pending a restart) still has work to do to remove that pod?


given that we have no checkpoint on pod manager, is it better for now to check if the pod exists in the pod killing channel?  if the pod exists in the pod killing channel, it implicitly means it has not yet been killed?

derekwaynecarr (changes requested):
See the comments throughout.

The core thing I suspect we are racing against is ensuring that between the call to pod manager DeletePod and the async cleanup of mirror pods is the need to check that the pod record itself does not exist in the pod killing channel.

Can you explore that and see if it simplifies code?

Tying the behavior back to the actual sequence of SyncLoop (VERB) commands has been important to get to that root understanding I think.

pkg/kubelet/kubelet_pods.go
tedyu: Derek:
Thanks for the detailed notes and suggestion on simplifying the PR.
Here is my response w.r.t. suggestion on pod killing channel check.

Kubelet#podKiller reads from podKillingCh and calls killPod() on the pod to be killed.
It is unknown how long the pod killing request would stay in the podKillingCh.

My expectation w.r.t. grace period from pod Spec is that the mirror pod should be deleted after this amount of time has elapsed from the time pod manifest is removed. The grace period would become less meaningful if we use a different starting point in time because we may not have much control over the factors that lead to the delayed serving of the pod killing request.

I think in your suggestion above, you meant checking whether the killing set in Kubelet#podKiller contains the pod we are interested in:
```
	killing := sets.NewString()
```
(since I don't know, in golang, how to check whether some request is in the channel without taking it out of the channel and serving it).
Actually I thought about this and didn't choose this route because of the following consideration:

We need to distinguish the following two scenarios:

1. the pod killing request is in podKillingCh and has not been served by the go routine yet
2. the pod killing request has been served by the go routine and is no longer in the killing set:
```
			go func(apiPod *v1.Pod, runningPod *kubecontainer.Pod) {
```
Since the orphan pod cleaner runs periodically, it becomes tricky how to reliably distinguish between the two scenarios (without making the Kubelet#podKiller logic more complicated).

w.r.t. kubelet restart, in the current formation of the PR, we remove the mirror pod after kubelet restart if it is detected that the mere presence of the mirror pod was due to wait (in prior instantiation of the kubelet) for the grace period to expire.

pkg/kubelet/kubelet_pods.go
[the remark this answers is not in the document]
tedyu: Responded in https://github.com/kubernetes/kubernetes/pull/92442#discussion_r447277746
If that response hasn't covered your question here, please let me know.

Thanks

pkg/kubelet/container/testing/fake_runtime.go
```diff
@@ -278,13 +278,44 @@ func (f *FakeRuntime) KillContainerInPod(container v1.Container, pod *v1.Pod) er
...
+}
+
 func (f *FakeRuntime) GetPodStatus(uid types.UID, name, namespace string) (*kubecontainer.PodStatus, error) {
 	f.Lock()
 	defer f.Unlock()
 
 	f.CalledFunctions = append(f.CalledFunctions, "GetPodStatus")
-	status := f.PodStatus
-	return &status, f.Err
+	return copyPodStatus(&f.PodStatus), f.Err
```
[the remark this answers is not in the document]
tedyu: There are two goroutines accessing pod status:

1. Kubelet#deleteOrphanedMirrorPods() calling GetPodStatus
2. the following code in TestDeleteOrphanedMirrorPods:
```
				testKubelet.fakeRuntime.SetPodStatus(orphanPods[0].UID, &podStatus)
```
Returning copy here would avoid data race between the above two accessors.

pkg/kubelet/cm/helpers.go
```diff
@@ -17,10 +17,63 @@ limitations under the License.
...
 	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	evictionapi "k8s.io/kubernetes/pkg/kubelet/eviction/api"
 )
 
+// CGroupPods records the relationship between Pod UID and its cgroup name
+// CGroupPods wraps the mapping between pod UID and cgroup name with a mutex
+// since the test in TestDeleteOrphanedMirrorPods involves goroutine.
+// The mutex prevents data race.
+type CGroupPods struct {
```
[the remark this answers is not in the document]
tedyu: The location of this class changed according to @rphillips suggestion.
I searched the current codebase and it seems helper_stub.go would be a new file.

Let me move it back to container_manager_stub.go

pkg/kubelet/pod/pod_manager.go
```diff
@@ -113,8 +120,9 @@ type basicManager struct {
 	mirrorPodByUID map[kubetypes.MirrorPodUID]*v1.Pod
 
 	// Pods indexed by full name for easy access.
-	podByFullName       map[string]*v1.Pod
-	mirrorPodByFullName map[string]*v1.Pod
+	podByFullName                   map[string]*v1.Pod
+	mirrorPodByFullName             map[string]*v1.Pod
+	podPendingTerminationByFullName map[string]kubetypes.UIDWithGracePeriod
```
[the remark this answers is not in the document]
tedyu: Addressed in the next update.

pkg/kubelet/pod/pod_manager.go
```diff
@@ -77,10 +78,16 @@ type Manager interface {
...
 	// mirror pods, this means deleting from indexes for all non-mirror pods.
 	DeletePod(pod *v1.Pod)
-	// DeleteOrphanedMirrorPods deletes all mirror pods which do not have
-	// associated static pods. This method sends deletion requests to the API
-	// server, but does NOT modify the internal pod storage in basicManager.
-	DeleteOrphanedMirrorPods()
+	// GetOrphanedMirrorPodNames returns names of orphaned mirror pods
+	GetOrphanedMirrorPodNames() []string
+	// GetUIDForPodPendingTermination returns UID for pods pending termination
+	GetUIDForPodPendingTermination(podFullName string) (kubetypes.UIDWithGracePeriod, bool)
```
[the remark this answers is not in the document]
tedyu: Addressed in the next update.

pkg/kubelet/pod/pod_manager.go
```diff
@@ -177,6 +186,7 @@ func isPodInTerminatedState(pod *v1.Pod) bool {
 // lock.
 func (pm *basicManager) updatePodsInternal(pods ...*v1.Pod) {
 	for _, pod := range pods {
+		podFullName := kubecontainer.GetPodFullName(pod)
```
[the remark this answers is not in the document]
tedyu: unintentional: would be restored in the next update.

pkg/kubelet/pod/pod_manager.go
```diff
@@ -77,10 +78,16 @@ type Manager interface {
 	// this means deleting the mappings related to mirror pods.  For non-
 	// mirror pods, this means deleting from indexes for all non-mirror pods.
 	DeletePod(pod *v1.Pod)
```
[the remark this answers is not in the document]
tedyu: Thanks for the reminder, I have updated the comment to reflect the current implementation.

pkg/kubelet/pod/pod_manager.go
```diff
@@ -77,10 +78,16 @@ type Manager interface {
...
-	// DeleteOrphanedMirrorPods deletes all mirror pods which do not have
-	// associated static pods. This method sends deletion requests to the API
-	// server, but does NOT modify the internal pod storage in basicManager.
-	DeleteOrphanedMirrorPods()
+	// GetOrphanedMirrorPodNames returns names of orphaned mirror pods
```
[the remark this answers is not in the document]
tedyu: The concept of orphaned mirror pod refers to mirror pod which is in api server but which is no longer backed by file in static pod manifest source directory.

For GetOrphanedMirrorPodNames(), it seems we don't need to mention graceful termination since it is the responsibility of orphan pod cleaner (located in kubelet_pods.go) to clean up mirror pods (respecting grace period specified for the underlying static pod).

w.r.t. the suggestion of removing relevant podByFullName entry in MarkMirrorPodTerminated, I have looked at the func's which depend on the value for podByFullName entry.

Let's look at the following call in Kubelet#GetAttach:
```
	pod, found := kl.GetPodByFullName(podFullName)
	if !found || (string(podUID) != "" && pod.UID != podUID) {
```
Currently, when the static pod is deleted, GetPodByFullName() would return false for ```found```, returning early.
If we remove podByFullName entry in MarkMirrorPodTerminated, the above call would return true for ```found```,  likely resulting in proceeding to remaining part of Kubelet#GetAttach().

I think we should remove podByFullName entry in DeletePod() because there may be existing code path(s) which depend on pod not being returned from GetPodByFullName() when the pod is undergoing termination.

Please let me know if you agree.

pkg/kubelet/pod/pod_manager.go
```diff
@@ -335,11 +359,34 @@ func (pm *basicManager) getOrphanedMirrorPodNames() []string {
...
+			return true
+		}
 	}
+	return false
+}
+
+func (pm *basicManager) MarkPodTerminated(podFullName string) {
+	pm.lock.Lock()
+	defer pm.lock.Unlock()
+	delete(pm.podPendingTerminationByFullName, podFullName)
```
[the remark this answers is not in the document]
tedyu: I responded in this comment:
https://github.com/kubernetes/kubernetes/pull/92442#discussion_r447334422

Thanks

pkg/kubelet/pod/pod_manager.go
```diff
@@ -77,10 +78,16 @@ type Manager interface {
...
+	// GetOrphanedMirrorPodNames returns names of orphaned mirror pods
+	GetOrphanedMirrorPodNames() []string
+	// GetUIDForPodPendingTermination returns UID for pods pending termination
+	GetUIDForPodPendingTermination(podFullName string) (kubetypes.UIDWithGracePeriod, bool)
+	// IsPodPendingTermination checks whether the UID belongs to pod which is pending termination
+	IsPodPendingTermination(uid types.UID) bool
+	// MarkPodTerminated removes map entry from podPendingTerminationByFullName for the pod
+	MarkPodTerminated(podFullName string)
+	// NumberOfPodsPendingTermination returns the number of pods pending termination
+	NumberOfPodsPendingTermination() int
```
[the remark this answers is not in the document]
tedyu: This is a good suggestion.
I have been thinking of removing this func.

In the updated PR, the assertion is made on the InstrumentedForTestPodManager and this func is no longer needed.

pkg/kubelet/kubelet_pods.go
derekwaynecarr: >Kubelet#podKiller reads from podKillingCh and calls killPod() on the pod to be killed.
>It is unknown how long the pod killing request would stay in the podKillingCh.
>Once the pod killing request is served, we don't know how long the killing of container would take.

The desire is to remove the mirror pod when we know its associated local pod resource is no longer running on the node.  It is right that we do not know how long the pod killing request would stay in the pod killing channel, but this is precisely the reason your code (which is right) gates deletion of the mirror pod until it sees all containers have reported exited.

The flow I see is the following:

1. the pod is removed from pod manager (i.e. its orphaned)
1. the pod is inserted in the pod killing channel (pending deletion)

if the sync loop invokes pod housekeeping before the pod killer acts on the pod in the channel, we will delete the mirror pod prematurely.  this is the core race condition we need to address.

The PR presented attempts to address this situation by making the pod manager aware of the terminating grace period for orphaned mirror pods.  

1. The housekeeping loop checks if the pod is still in its grace period from pod manager 
1. The housekeeping loop checks if the pod container status is terminal

The second check is implicitly ensuring that the pod killer channel has not acted on the pod yet.

Instead of changing the workflow around pod manager and the states it maintains, it seems like we should make the cleanup routine aware that a pod is orphaned AND know if the pod killer channel has completed its task before deleting the mirror pod.

What the changes in your code flow appear to do is try to make pod manager grace period aware, but its still disconnected from the content of the channel which is the actual condition we are racing against.

It's important we understand the sequence of actions from delete to removal from the API server which is why recording the state of interactions in sync loop, state of objects in pod manager, state of objects in pod killing channel are all important to resolving the core issue.

WDYT?

tedyu: The runningPods in Kubelet#HandlePodCleanups() are obtained from kl.runtimeCache, hence they can be slightly out of date.
Currently Kubelet#HandlePodCleanups() interacts with pod killer via sending pod to podKillingCh only.

Some implications of this are:
1. before unwanted pod is killed, Kubelet#HandlePodCleanups() would continue to add the pod to podKillingCh. pod killer would filter out the repetition using the killing set. This repetition seems unnecessary.
2. If we expose the killing set alone, it is still tricky for Kubelet#HandlePodCleanups() to tell whether the pod killer has finished the call to kl.killPod().

In the following sequence of events, T3 and T4 are very close (in the sense that runtimeCache isn't updated).

T1 Kubelet#HandlePodCleanups() finds pod 1 as unwanted, sends it to podKillingCh
T2 pod killer takes pod 1 out of podKillingCh and works on the call to kl.killPod()
T3 call to kl.killPod() call returns to pod killer, pod 1 is taken out of the killing set.
T4 Kubelet#HandlePodCleanups() gathers runningPods (including pod 1, according to runtimeCache) and desiredPods, performs the loop again (sending pod 1 to podKillingCh):
```
	for _, pod := range runningPods {
		if _, found := desiredPods[pod.ID]; !found {
```

If we look at the killing set at T4, we wouldn't know whether the pod termination has completed, without checking the pod status (via containerRuntime.GetPodStatus).

I thought about adding another Set, called toBeKilled, which contain pods in runningPods but not in desiredPods, for pod killer. This would get around the task of checking pod(s) currently in the podKillingCh, but seems inadequate for the scenario described above where the pod is neither in podKillingCh (toBeKilled set) nor in killing set.

Hence I chose to depend on the call to containerRuntime.GetPodStatus in Kubelet#deleteOrphanedMirrorPods which would tell us whether the pod has really terminated.

I will keep thinking about the interactions between the following :
* pod manager
* housekeeping (Kubelet#HandlePodCleanups)
* pod killer
* runtimeCache

thanks

tedyu: From my comment yesterday:
```
My expectation w.r.t. grace period from pod Spec is that the mirror pod should be deleted after this amount of time has elapsed from the time pod manifest is removed.
```
It is responsibility of Kubelet#HandlePodCleanups(Kubelet#deleteOrphanedMirrorPods, to be exact) to clean up orphaned mirror pods since the Kubelet#killPod() call may take longer than grace period to complete.
This means Kubelet#deleteOrphanedMirrorPods needs to know when the deletion of static pod is initiated so that it can keep track of the time before grace period expires.

I thought of introducing a map, podTerminationMap into Kubelet class which keeps the following information:
1 full pod name 
2 the timestamp when static pod deletion is initiated
3 the timestamp when the pod containers are terminated

#2 is used by Kubelet#deleteOrphanedMirrorPods to know when grace period expires.
#3 is filled by pod killer when Kubelet#killPod() returns

Kubelet#deleteOrphanedMirrorPods would iterate podTerminationMap and call mirror client to remove mirror pod when either of the following is true:
* grace period for the underlying pod has passed
* pod killer informs deleteOrphanedMirrorPods() via #3 that the pod has terminated

However we need to coordinate accesses between Kubelet#deleteOrphanedMirrorPods and pod killer to the map so that no map entry for the pod is leaked.

Please let me know if the above is simpler than the formation in this PR.

pkg/kubelet/pod/pod_manager.go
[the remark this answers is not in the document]
derekwaynecarr: i agree that if we evolve pod manager to track pods undergoing termination, then we need to keep the behavior you have.  i think it makes it harder to understand what pod manager is itself doing, and so that is why i was wondering if we could get to a simpler resolution that checks if (mirror,pod) is still present in pod killing channel.

pkg/kubelet/kubelet_pods.go
tedyu: Here was excerpt from kubelet log on a patched 1.18 master node (using patch in the above approach) where kube-apiserver-pod.yaml was moved out of manifest dir.
```
Jun 30 22:38:41 ip-10-0-136-93 hyperkube[768651]: I0630 22:38:41.942971  768651 kubelet_pods.go:1031] checking entry "kube-apiserver-ip-10-0-136-93.ec2.internal_openshift-kube-apiserver", 2020-06-30 22:39:45.76633046 +0000 UTC m=+394.295683069, 2020-06-30 22:38:41.166458121 +0000 UTC m=+329.695810607
Jun 30 22:38:41 ip-10-0-136-93 hyperkube[768651]: I0630 22:38:41.943004  768651 kubelet_pods.go:1040] deleting pod "kube-apiserver-ip-10-0-136-93.ec2.internal_openshift-kube-apiserver", 2020-06-30 22:38:41.166458121 +0000 UTC m=+329.695810607
Jun 30 22:38:41 ip-10-0-136-93 hyperkube[768651]: I0630 22:38:41.943018  768651 mirror_client.go:124] Deleting a mirror pod "kube-apiserver-ip-10-0-136-93.ec2.internal_openshift-kube-apiserver" (uid (*types.UID)(nil))
```
Please note that '2020-06-30 22:38:41.166458121' (in the first line) was the timestamp when pod killer completed the Kubelet#killPod call.
Also note that 2020-06-30 22:39:45.76633046 was the deadline for the grace period - the mirror pod was deleted prior to the deadline since the containers have been destroyed.

tedyu:
[mirror-pod-killer-grace-period.txt](https://github.com/kubernetes/kubernetes/files/4854820/mirror-pod-killer-grace-period.txt)
@derekwaynecarr 
Can you take a look at the above patch ?

TestDeleteOrphanedMirrorPods in kubelet_test.go has been revised to work with new approach
Done cluster testing - see https://github.com/kubernetes/kubernetes/pull/92442#discussion_r448021717
node-e2e passes based on the above patch

If you think the patch is simpler compared to the PR, please let me know whether I should:

1. overwrite this PR with patch or,
2. open another PR with the patch

tedyu:
[mirror-pod-del-check-cont-cgroup.txt](https://github.com/kubernetes/kubernetes/files/4858697/mirror-pod-del-check-cont-cgroup.txt)

This was the earlier version where pod manager was modified.

It shows a previous approach - see comments above for why we opt for the latest approach where pod manager is not modified.

sjenning:
I was out for the past week, so just starting to look at this.  Will update with my feedback today.

pkg/kubelet/kubelet_pods.go
```diff
@@ -1737,6 +1786,14 @@ func (kl *Kubelet) cleanupOrphanedPodCgroups(cgroupPods map[types.UID]cm.CgroupN
 			continue
 		}
 
+		// if the pod is within termination grace period, we shouldn't cleanup the underlying cgroup
+		kl.podKillingLock.Lock()
```
sjenning, line 1790: `podKillingLock` and `mirrorPodTerminationMap` being accessible anywhere in the kubelet is messy.  We need an abstraction on top of it.

sjenning (changes requested):

pkg/kubelet/kubelet_pods.go
```diff
@@ -1137,6 +1181,12 @@ func (kl *Kubelet) podKiller() {
 				lock.Lock()
 				killing.Delete(string(runningPod.ID))
 				lock.Unlock()
+				kl.podKillingLock.Lock()
+				if gracePeriod, ok := kl.mirrorPodTerminationMap[string(runningPod.ID)]; ok {
```
sjenning, line 1185: this can add non-mirror pods to the `mirrorPodTerminationMap` right?

pkg/kubelet/kubelet_pods.go
sjenning: something along the lines of
```
type PodCleanupManager struct {
  podManager
  podKillingLock
  mirrorPodTerminationMap
}

// GracePeriod struct can be made private to this

func RecordGracePeriod
func DeleteOrphanedMirrorPods
func IsPodTerminationPending
```

pkg/kubelet/kubelet_pods.go
tedyu: Please note that we check the presence of UID key in the map.
If the key doesn't exist, there is no entry added.

The removal of the map entry is solely done by kl.deleteOrphanedMirrorPods()

pkg/kubelet/kubelet_pods.go
derekwaynecarr: @tedyu and I discussed something similar this morning.

I think an abstraction that lets us get visibility into the the pods queued up in the pod killer channel is the core need.  So rather that kubelet managing a podKillerCh, it could manage a `PodKiller` and support queue operations, and visibility into contents of the queue.  I see little reason to track a grace period time with pods in that interface for any reason other than logging in debug logs.  We just need to know if a pod or its mirror is in the queue for killing.

sjenning: hold on this for now though. i haven't fully thought it out.

sjenning: sounds like you and @derekwaynecarr where already going down this path :+1:

pkg/kubelet/kubelet_pods.go
```diff
@@ -1722,13 +1772,12 @@ func (kl *Kubelet) GetPortForward(podName, podNamespace string, podUID types.UID
 
 // cleanupOrphanedPodCgroups removes cgroups that should no longer exist.
 // it reconciles the cached state of cgroupPods with the specified list of runningPods
-func (kl *Kubelet) cleanupOrphanedPodCgroups(cgroupPods map[types.UID]cm.CgroupName, activePods []*v1.Pod) {
+func (kl *Kubelet) cleanupOrphanedPodCgroups(pcm cm.PodContainerManager, cgroupPods map[types.UID]cm.CgroupName, activePods []*v1.Pod) {
```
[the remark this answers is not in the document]
tedyu: This is actually used by the enhanced unit test (TestDeleteOrphanedMirrorPods) where using single instance of pcm during one HandlePodCleanups() call allows the assertion to be written.

Also, it seems cleaner using single instance of pcm which reduces garbage for Go runtime.

pkg/kubelet/kubelet_pods.go
```diff
@@ -1129,6 +1170,9 @@ func (kl *Kubelet) podKiller() {
 
 		if !exists {
 			go func(apiPod *v1.Pod, runningPod *kubecontainer.Pod) {
+				if apiPod == nil {
+					apiPod = runningPod.ToAPIPod()
+				}
```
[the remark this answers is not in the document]
tedyu: True.
This was for an earlier version of the patch.

Dropped.

tedyu:
@sjenning @derekwaynecarr 
I updated the PR by extracting mirrorPodTerminationMap related operations into func's

Work toward PodKiller abstraction is under way.

pkg/kubelet/kubelet_pods.go
tedyu: @sjenning @derekwaynecarr 
I have refactored the code by creating PodKiller interface and PodKillerWithChannel struct.

Please take a look.

Thanks

tedyu:
@sjenning @derekwaynecarr
I have made GracePeriod private since it is only used by PodKiller.
I moved the new mirror pod related methods to PodKiller so that the code is more coherent.

We can discuss giving PodKiller a better name.

pkg/kubelet/kubelet_pods.go
tedyu: I have consolidated the locks in PodKillerWithChannel#PerformPodKillingWork() :
```
  	lock := sync.Mutex{}
```
the above lock from existing code and podKillingLock are combined into one lock.

tedyu: Potentially we can experiment with pod killer where the receipt of pod killing requests is via, e.g. slice (instead of channel).
But that requires more validation work and can be done in a subsequent PR.

tedyu: @sjenning @derekwaynecarr 
Please take another look.

Thanks

sjenning:
@tedyu i'm certain we are going to make more changes here.  i'm reviewing this evening.  don't worry about trying to get the test to pass.

sjenning:
In an attempt to cut through everything, I pulled down this PR and just started reducing everything I could.

This was the result https://github.com/sjenning/kubernetes/commit/e69f6f9bdf7df5fc63e1670c0621fe65982cef7a

I cut all the test stuff out of my commit to clean up the review for now.

I would like to see something like this, assuming it still fixes the issue (untested).

tedyu:
In Seth's version, basicManager#getOrphanedMirrorPodNames is kept.
I noticed that the code for removing mirror pod upon kubelet restart is not kept. I assume such code would be added in basicManager#getOrphanedMirrorPodNames

Let me experiment with Seth's version.
Need to get tests compiling:
```
./kubelet_test.go:224:49: not enough arguments in call to "k8s.io/kubernetes/pkg/kubelet/pod".NewBasicPodManager
```

tedyu:
I think in @sjenning 's version, the following call to MarkMirrorPodTerminating was misplaced:
```
@@ -2038,6 +2033,9 @@ func (kl *Kubelet) HandlePodRemoves(pods []*v1.Pod) {
        for _, pod := range pods {
                kl.podManager.DeletePod(pod)
                if kubetypes.IsMirrorPod(pod) {
+                       if _, ok := kl.podManager.GetMirrorPodByPod(pod); ok {
+                               kl.podKiller.MarkMirrorPodTerminating(pod)
```
This is because when static pod is deleted, we need to mark the pod going through graceful termination. Hence the above addition is not effective.

This would be the way to go:
```
		kl.podManager.DeletePod(pod)
		kl.podKiller.MarkMirrorPodTerminating(pod)
```
with which TestDeleteOrphanedMirrorPods passes.

tedyu:
Still need to debug why node-e2e tests fail:
```
[k8s.io] MirrorPod when create a mirror pod should be updated when static pod updated [NodeConformance] expand_more2m16s | [k8s.io] MirrorPod when create a mirror pod should be updated when static pod updated [NodeConformance] expand_more | 2m16s
-- | -- | --
[k8s.io] MirrorPod when create a mirror pod should be updated when static pod updated [NodeConformance] expand_more | 2m16s
[k8s.io] MirrorPod when create a mirror pod should be recreated when mirror pod gracefully deleted [NodeConformance] expand_more2m16s | [k8s.io] MirrorPod when create a mirror pod should be recreated when mirror pod gracefully deleted [NodeConformance] expand_more | 2m16s
[k8s.io] MirrorPod when create a mirror pod should be recreated when mirror pod gracefully deleted [NodeConformance] expand_more | 2m16s
[k8s.io] MirrorPod when create a mirror pod should be recreated when mirror pod forcibly deleted [NodeConformance] expand_more | [k8s.io] MirrorPod when create a mirror pod should be recreated when mirror pod forcibly deleted [NodeConformance] expand_more
[k8s.io] MirrorPod when create a mirror pod should be recreated when mirror pod forcibly deleted [NodeConformance] expand_more
```
Here is the patch based on Seth's change:
[mirror-pod-killer-itf-seth.txt](https://github.com/kubernetes/kubernetes/files/4882391/mirror-pod-killer-itf-seth.txt)

tedyu:
@sjenning 
I have updated the PR with the following changes:

1. dropped the timestamp from gracePeriod struct so that the mirrorPodTerminationMap becomes UID to pod name mapping
2. dropped DeleteExpiredMirrorPod since the removal from mirrorPodTerminationMap is done by the pod killer now

I kept Kubelet#deleteOrphanedMirrorPods() in Kubelet since that's where the pod killer is located - the related code is closer this way.

I noticed you shortened method names. I currently keep the MirrorPodPendingTermination style since it was used in the formation before the pod killer refactoring came about. Let me know your preference.

Thanks

sjenning (changes requested):

pkg/kubelet/kubelet.go
```diff
@@ -2037,6 +2032,7 @@ func (kl *Kubelet) HandlePodRemoves(pods []*v1.Pod) {
 	start := kl.clock.Now()
 	for _, pod := range pods {
 		kl.podManager.DeletePod(pod)
+		kl.podKiller.MarkMirrorPodPendingTermination(pod)
```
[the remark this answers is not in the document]
tedyu: In the next update, I would do the following:

1. use slice instead of channel to track the outstanding pod killing requests
2. func names in pod killer no longer carry 'mirror'
3. podTerminationMap and the killing Set are combined into one map
4. remove the above call in HandlePodRemoves
5. remove Close() func from PodKiller

tedyu:
[mirror-pod-killer-itf-11.txt](https://github.com/kubernetes/kubernetes/files/4886185/mirror-pod-killer-itf-11.txt)

For record: v11 which was modified based on Seth's suggestions.

pkg/kubelet/kubelet.go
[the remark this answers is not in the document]
sjenning: if you don't use a channel, you'll have to have some sort of signalling method to wake up the podKiller goroutine.  what did you have in mind? polling the queue?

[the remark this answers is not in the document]
tedyu: The podKiller goroutine would continue to completion once launched.

I think you are asking how the subsequent pod killing requests are retrieved.
podKillerWithSlice#PerformPodKillingWork() would copy outstanding pod killing requests from the slice field.
If there is no pod killing request at the moment, PerformPodKillingWork() would return.

It would be periodically run:
```
	go wait.Until(kl.podKiller.PerformPodKillingWork, 1*time.Second, wait.NeverStop)
```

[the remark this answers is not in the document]
sjenning: that would introduce up to 1s of latency between a call to `KillPod()` and the podkiller actually starting the goroutine to kill it.  where a channel is more or less instantaneous.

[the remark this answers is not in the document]
tedyu: I switched to using slice because channel doesn't provide 'peek' functionality: we have to fetch the request out of the channel.
For podKillerWithSlice, once pod killing request is sent, we know where to find pod pending termination.

If incurring lower latency is of consideration, we can shorten the period in wait.Until() to, e.g. 500 milliseconds.

I am open to other suggestion on this as well.

tedyu:
[mirror-pod-killer-slice.txt](https://github.com/kubernetes/kubernetes/files/4886439/mirror-pod-killer-slice.txt)

@sjenning 
Here is the podKillerWithSlice implementation.
If you feel it is clearer than current formation, I will update the PR.

I think this gets closer to our initial motivation by using pod killer as source of truth for judging whether pod is going through termination.

I have run this through e2e tests and tests pass.

pkg/kubelet/kubelet.go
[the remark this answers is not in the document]
sjenning: we don't want to poll at any rate when the previous behavior was ~instant.  just keep the channel to avoid regressions.  we want to change _as little as possible_ in order to fix the issue.  strapping a map with redundant data onto the side of the existing data path is preferable to changing the channel based code.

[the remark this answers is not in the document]
tedyu: Looking at podKillerWithChannel#MarkMirrorPodPendingTermination(), it has check which limits the addition to pod which has mirror pod:
```
		if _, ok := pk.getMirrorPodByPod(pod); ok {
```
So the entries in mirrorPodTerminationMap would be for mirror pod. Hence I name the map that way.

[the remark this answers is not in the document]
tedyu: @sjenning 
Let me know if you have more review comments.

Thanks

tedyu:
[mirror-pod-killer-slice-cond.txt](https://github.com/kubernetes/kubernetes/files/4887576/mirror-pod-killer-slice-cond.txt)

@sjenning @derekwaynecarr 

In the above patch, I have done the following:

1. use slice instead of channel to track the outstanding pod killing requests
2. func names in pod killer no longer carry 'mirror'
3. podTerminationMap and the killing Set are combined into one map
4. remove the call to MarkMirrorPodPendingTermination() in HandlePodRemoves since it is not needed
5. remove Close() func from PodKiller

The podKillerWithSlice#PerformPodKillingWork() doesn't return: it waits for the signal (if there is no outstanding pod killing request) that the next pod killing request has arrived. This should give us performance on par with using a channel for receiving pod killing requests.

Methods of potential interest: podKillerWithSlice#PerformPodKillingWork , podKillerWithSlice#KillPod

I have run test suite using this patch. node-e2e tests also pass.

Please see if you want to review this version in more detail.

sjenning:
@tedyu I appreciate your work on this.  Can you push the changes to the PR rather than posting the patch as an attached file?  It makes the review process much easier (most inline comments) and more transparent. Thanks!

tedyu:
@sjenning 
I have updated the PR.

Methods of interest: podKillerWithSlice#PerformPodKillingWork , podKillerWithSlice#KillPod

Please start with the above two func's

pkg/kubelet/kubelet_pods.go
```diff
@@ -1099,46 +1117,170 @@ func (kl *Kubelet) HandlePodCleanups() error {
...
+	pk.podKillingLock.Unlock()
+	return exists
+}
+
+// wait waits for the arrival of subsequent pod killing request
+func (pk *podKillerWithSlice) wait() {
+	pk.podKillingLock.Lock()
+	defer pk.podKillingLock.Unlock()
+	for len(pk.podsToKill) == 0 {
+		pk.cond.Wait()
```
sjenning, line 1245: this `Wait()`s with the `podKillingLock` lock held

tedyu: This is intentional, otherwise golang would panic.
Here is the body of Wait():
```
func (c *Cond) Wait() {
	c.checker.check()
	t := runtime_notifyListAdd(&c.notify)
	c.L.Unlock()
```
Please see the Unlock call.

pkg/kubelet/kubelet_pods.go
```diff
@@ -1099,46 +1117,170 @@ func (kl *Kubelet) HandlePodCleanups() error {
...
+		}
+	}
+	return false
+}
+
+// IsPodPendingTerminationByPodName checks whether the given pod is in grace period of termination
+func (pk *podKillerWithSlice) IsPodPendingTerminationByPodName(podFullname string) bool {
+	pk.podKillingLock.Lock()
+	defer pk.podKillingLock.Unlock()
+	for _, pair := range pk.podsToKill {
```
sjenning, line 1193: `podsToKill` is a subset of the pods in `podTerminationMap`.  we only need to check `podTerminationMap`.

pkg/kubelet/kubelet_pods.go
```diff
@@ -1099,46 +1117,170 @@ func (kl *Kubelet) HandlePodCleanups() error {
...
+}
+
+// IsPodPendingTerminationByUID checks whether the pod for the given uid is pending termination
+func (pk *podKillerWithSlice) IsPodPendingTerminationByUID(uid types.UID) bool {
+	pk.podKillingLock.Lock()
+	defer pk.podKillingLock.Unlock()
+	if _, ok := pk.podTerminationMap[string(uid)]; ok {
+		return true
+	}
+	for _, pair := range pk.podsToKill {
```
sjenning, line 1181: podsToKill is a subset of the pods in podTerminationMap. we only need to check podTerminationMap.

tedyu: Entry is inserted into pk.podTerminationMap when starting to serve pod killing request (which is taken off of podsToKill).
A pod going through termination would either be in podsToKill (queued) or in podTerminationMap (pending termination).
So both checks are needed.

pkg/kubelet/kubelet_pods.go
tedyu: Entry is inserted into pk.podTerminationMap when starting to serve pod killing request (which is taken off of podsToKill).
A pod going through termination would either be in podsToKill (queued) or in podTerminationMap (pending termination).
So both checks are needed.

pkg/kubelet/kubelet_pods.go
sjenning: ok i see

pkg/kubelet/kubelet_pods.go
```diff
@@ -1099,46 +1117,170 @@ func (kl *Kubelet) HandlePodCleanups() error {
...
+	var podsToKill []*kubecontainer.PodPair
+	for true {
+		pk.podKillingLock.Lock()
+		for _, podPair := range pk.podsToKill {
+			podsToKill = append(podsToKill, podPair)
+		}
+		pk.podsToKill = pk.podsToKill[:0]
+		pk.podKillingLock.Unlock()
+
+		if len(podsToKill) == 0 {
```
sjenning, line 1261: we are doubling up on the locking and condition check here.  remove the `wait()` function, inline it here, and move the Unlock on 1259 below this block
```
for len(pk.podsToKill) == 0 {
		pk.cond.Wait()
}
```

tedyu: I was thinking on improving locking for this part of code.

This would be updated in the next push.

pkg/kubelet/kubelet_pods.go
```diff
@@ -1099,46 +1117,170 @@ func (kl *Kubelet) HandlePodCleanups() error {
...
+	// killPod is the func which invokes runtime to kill the pod
+	killPod func(pod *v1.Pod, runningPod *kubecontainer.Pod, status *kubecontainer.PodStatus, gracePeriodOverride *int64) error
+	// getMirrorPodByPod calls pod manager's GetMirrorPodByPod
+	getMirrorPodByPod func(pod *v1.Pod) (*v1.Pod, bool)
+}
+
+// NewPodKiller returns a functional PodKiller
+func NewPodKiller(kl *Kubelet) PodKiller {
+	podKiller := &podKillerWithSlice{
+		podsToKill:        make([]*kubecontainer.PodPair, 0),
```
sjenning, line 1164: adding `podsToKill` plus `cond` has made this code more complex over the channel approach.  please return the the channel method.

tedyu: I have restored to yesterday's formation.

pkg/kubelet/kubelet_pods.go
```diff
@@ -1099,24 +1122,115 @@ func (kl *Kubelet) HandlePodCleanups() error {
...
+		if name == podFullname {
+			return true
+		}
+	}
+	return false
+}
+
+func (pk *podKillerWithChannel) markMirrorPodTerminated(uid string) bool {
+	pk.podKillingLock.Lock()
+	_, ok := pk.mirrorPodTerminationMap[uid]
```
sjenning, line 1199: just call `delete`. don't need to check if it is present.

tedyu:
```
		klog.V(4).Infof("marking pod termination %q", uid)
```
The intention is to have log for this action.
Can we keep the logging ?

sjenning: the logging is fine and will work if the uid is present or not in the map

tedyu: Done.

pkg/kubelet/kubelet.go
```diff
@@ -2041,6 +2036,9 @@ func (kl *Kubelet) HandlePodRemoves(pods []*v1.Pod) {
 			kl.handleMirrorPod(pod, start)
 			continue
 		}
+		if _, ok := kl.podManager.GetMirrorPodByPod(pod); ok {
```
sjenning, line 2039: mirror pods never get to this point due to the `continue` on 2036

tedyu: This call is for pod which has mirror pod.
That's why I place it here after the kubetypes.IsMirrorPod() call.

sjenning:
Ok I think the running code looks good now.  @tedyu please make no further changes to the following files
```
kubelet.go
kubelet_pods.go
pod_manager.go
```
I'll review the tests.
fyi @derekwaynecarr @rphillips

sjenning:
ok, i'm am uneasy about the changes to *stub.go, kubelet_test.go, and kubelet_node_status_test.go.  please pull those changes out for now, except those need to make it compile like https://github.com/kubernetes/kubernetes/pull/92442/files#diff-1e7657ffc53feeb1a99dd257a3ea303dR291)

the e2e test is sufficient for now.

tedyu:
@sjenning 
Unit test code has been taken out.

tedyu:
Tested on patched 1.18 master node.
```
mv kube-apiserver-pod.yaml /tmp/
```
After the manifest is moved out, the container ran for ~70 seconds:
```
# crictl ps | grep apise
531c4ab0e399e       65065b8f3de3b401d55f7de1c91bcd7e3f4924b75208ab2109ce71cbd7f2c84d                                                         13 minutes ago      Running             kube-apiserver                                0                   ccc8c5b2855ee
```
Here is relevant kubelet log:
```
Jul 08 21:56:53 ip-10-0-147-205 hyperkube[79799]: I0708 21:56:53.529796   79799 kubelet.go:1930] SyncLoop (REMOVE, "file"): "kube-apiserver-ip-10-0-147-205.us-west-1.compute.internal_openshift-kube-apiserver(90ab8003b28fbeda6bc89e35f6582ecc)"

Jul 08 21:58:04 ip-10-0-147-205 hyperkube[79799]: I0708 21:58:04.039342   79799 kubelet_pods.go:1196] marking pod termination "90ab8003b28fbeda6bc89e35f6582ecc"
```

sjenning:
Looks good here :+1: 
```
I0709 09:39:59.637011   46821 kubelet.go:1844] SyncLoop (REMOVE, "file"): "mirror-pod-remove-test-127.0.0.1_test(5fda0b7885755b311057b80f00dfae7b)"
I0709 09:39:59.637052   46821 kubelet_pods.go:1207] marking pod pending termination "5fda0b7885755b311057b80f00dfae7b"
I0709 09:39:59.637116   46821 kubelet_pods.go:1242] Killing unwanted pod "mirror-pod-remove-test-127.0.0.1"
I0709 09:39:59.637192   46821 kuberuntime_container.go:635] Killing container "cri-o://51cc27b01ba4983c35c3fa1c0183328777d69afc943d2f57b4da549adbda22b9" with a 20 second grace period
I0709 09:39:59.637350   46821 event.go:291] "Event occurred" [object test/mirror-pod-remove-test-127.0.0.1 kind Pod apiVersion v1 type Normal reason Killing message Stopping container static-pod]="(MISSING)"
I0709 09:40:00.219480   46821 kubelet_pods.go:1852] pod "5fda0b7885755b311057b80f00dfae7b" is pending termination
I0709 09:40:02.221497   46821 kubelet_pods.go:1852] pod "5fda0b7885755b311057b80f00dfae7b" is pending termination
I0709 09:40:02.342587   46821 manager.go:1042] Destroyed container: "/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod5fda0b7885755b311057b80f00dfae7b.slice/crio-1fb618269bc53d46c0a62c84d9012b7b312aa97e733a8ee53f67c9ac0d76285b.scope" (aliases: [k8s_static-pod_mirror-pod-remove-test-127.0.0.1_test_5fda0b7885755b311057b80f00dfae7b_0 1fb618269bc53d46c0a62c84d9012b7b312aa97e733a8ee53f67c9ac0d76285b], namespace: "crio")
I0709 09:40:02.380035   46821 manager.go:1042] Destroyed container: "/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod5fda0b7885755b311057b80f00dfae7b.slice/crio-31b1b6a2fe91eed72890eee0bcc013f6a06049daa8f8a2cd8666787c3b1ed71b.scope" (aliases: [k8s_POD_mirror-pod-remove-test-127.0.0.1_test_5fda0b7885755b311057b80f00dfae7b_0 31b1b6a2fe91eed72890eee0bcc013f6a06049daa8f8a2cd8666787c3b1ed71b], namespace: "crio")
I0709 09:40:04.219709   46821 kubelet_pods.go:1852] pod "5fda0b7885755b311057b80f00dfae7b" is pending termination
I0709 09:40:06.220122   46821 kubelet_pods.go:1852] pod "5fda0b7885755b311057b80f00dfae7b" is pending termination
I0709 09:40:08.219567   46821 kubelet_pods.go:1852] pod "5fda0b7885755b311057b80f00dfae7b" is pending termination
I0709 09:40:10.220060   46821 kubelet_pods.go:1852] pod "5fda0b7885755b311057b80f00dfae7b" is pending termination
I0709 09:40:12.220094   46821 kubelet_pods.go:1852] pod "5fda0b7885755b311057b80f00dfae7b" is pending termination
I0709 09:40:14.219532   46821 kubelet_pods.go:1852] pod "5fda0b7885755b311057b80f00dfae7b" is pending termination
I0709 09:40:16.221687   46821 kubelet_pods.go:1852] pod "5fda0b7885755b311057b80f00dfae7b" is pending termination
I0709 09:40:18.219497   46821 kubelet_pods.go:1852] pod "5fda0b7885755b311057b80f00dfae7b" is pending termination
E0709 09:40:19.685743   46821 remote_runtime.go:563] ReopenContainerLog "51cc27b01ba4983c35c3fa1c0183328777d69afc943d2f57b4da549adbda22b9" from runtime service failed: rpc error: code = Unknown desc = container is not created or running
I0709 09:40:19.711982   46821 kuberuntime_container.go:641] Container "cri-o://51cc27b01ba4983c35c3fa1c0183328777d69afc943d2f57b4da549adbda22b9" exited normally
I0709 09:40:20.219096   46821 kubelet_volumes.go:145] Orphaned pod "5fda0b7885755b311057b80f00dfae7b" found, removing
I0709 09:40:20.219421   46821 mirror_client.go:125] Deleting a mirror pod "mirror-pod-remove-test-127.0.0.1_test" (uid (*types.UID)(nil))
I0709 09:40:20.223785   46821 kubelet_pods.go:1039] deleted pod "mirror-pod-remove-test-127.0.0.1_test"
I0709 09:40:20.223910   46821 kubelet.go:1850] SyncLoop (DELETE, "api"): "mirror-pod-remove-test-127.0.0.1_test(ffab0bbe-f881-4089-b9ce-8b2b5f288447)"
I0709 09:40:20.223932   46821 kubelet.go:1844] SyncLoop (REMOVE, "api"): "mirror-pod-remove-test-127.0.0.1_test(ffab0bbe-f881-4089-b9ce-8b2b5f288447)"
```

PR #92445, opened 2020-06-23
Added script to generate module level dependency graph.

cheftako:
**What type of PR is this?**
/kind documentation

Contains a script which generates a graph/visualization of our module level dependency graph.
Helps understand things like what depends on component_base and what component_base depends on.

Fixes # N/A

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/hack/module-graph.sh
+++ b/hack/module-graph.sh
@@ -1,0 +1,66 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script generates a jpg image of our internal dependency graph.
+# It relies on go mod and dot to do so and should be run from K8s root.
+# TODO: Containerize the script to remove dependency issues with go mod and dot.
+
+# To generate graph with all Kubernetes modules run
+# ./hack/module-graph.sh
+# To generate graph with just staging modules run
+# ./hack/module-graph.sh staging
+
+error_exit()
+{
+	echo "$1" 1>&2
+	exit 1
+}
+
+staging_dependencies()
+{
+	relevant_modules="(k8s.io/kubernetes|$(find staging/src/k8s.io -maxdepth 1 -mindepth 1 | sed -E 's|staging/src/k8s.io|k8s.io|g' | tr '\n' '|' | sed 's#|$##' ))"
+	# Generating lines of the form " k8s_io_kubernetes -> k8s_io_api"
+	# Use only the directories in staging
+	# Trimming away version info
+	# Replacing non DOT (graph description language) characters with underscores
+	# Dedupe lines
+	# Inserting needed arrow.
+	# Indenting the line appropriately
+	go mod graph | grep -E "^${relevant_modules}(@.*| )${relevant_modules}@.*$" | sed -E 's|@\S+ | |g' | sed -E 's|@\S+$||g' | sed -E 's/[\.\/\-]/_/g' | sort | uniq | sed -E 's| | -> |g' | sed -E 's|^|    |g' >> _output/module-dependencies.dot || error_exit "Failed to generate staging dependencies in DOT file"
+}
+
+kubernetes_dependencies()
+{
+	# Generating lines of the form " k8s_io_kubernetes -> k8s_io_api"
+	# Excluding all non Kubernetes dependencies
+	# Trimming away version info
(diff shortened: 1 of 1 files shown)


logicalhan (commented):
/lgtm

Should we err if dot isn't found?

cheftako:
> /lgtm
> 
> Should we err if dot isn't found?

Sure, rudimentary error handling added.

hack/module-graph.sh
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script generates a jpg image of our internal dependency graph.
+# It relies on go mod and dot to do so and should be run from K8s root.
+# TODO: Containerize the script to remove dependency issues with go mod and dot.
+
+echo "digraph module_dependencies {" > hack/module-dependencies.dot
+go mod graph | egrep "^.*k8s.io.*k8s.io.*$" | sed -E 's|@\S+ | |g' | sed -E 's|@\S+$||g' | sed -E 's/[\.\/\-]/_/g' | sed -E 's| | -> |g' | sed -E 's|^|    |g' >> hack/module-dependencies.dot
```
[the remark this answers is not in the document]
cheftako: Ordered comments for this added.

hack/module-graph.sh
```diff
@@ -0,0 +1,24 @@
...
+echo "}" >> hack/module-dependencies.dot
+dot -Gratio=1,1 -Nwidth=2 -Nheight=2 -Nfontsize=48 -Earrowsize=8 -Tjpg hack/module-dependencies.dot -o module-dependencies.jpg
```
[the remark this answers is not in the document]
cheftako: Good point. Switch both intermediary and final file into _output

hack/module-graph.sh
```diff
@@ -0,0 +1,36 @@
...
+}
+
+echo "digraph module_dependencies {" > _output/module-dependencies.dot || error_exit "Failed to open DOT file"
+# Generating lines of the form " k8s_io_kubernetes -> k8s_io_api"
+# Excluding all non Kubernetes dependencies
+# Trimming away version info
+# Replacing non DOT (graph description language) characters with underscores 
+# Inserting needed arrow.
+# Indenting the line appropriately
+go mod graph | egrep "^.*k8s.io.*k8s.io.*$" | sed -E 's|@\S+ | |g' | sed -E 's|@\S+$||g' | sed -E 's/[\.\/\-]/_/g' | sed -E 's| | -> |g' | sed -E 's|^|    |g' >> _output/module-dependencies.dot || error_exit "Failed to generate dependencies in DOT file"
```
liggitt, line 34: since we have some k8s.io/... modules that aren't in staging, something like this would let you limit the graph to just the in-tree modules:

```
relevant_modules="k8s.io/(kubernetes|$(ls staging/src/k8s.io | tr '\n' '|' | sed 's#|$##' ))"
go mod graph | egrep "^${relevant_modules}(@.*| )${relevant_modules}@.*$" | ...
```

liggitt: with that change:
module-dependencies

liggitt: without that change:
module-dependencies

BenTheElder:
/lgtm cancel
(to stop the bot)
legit failure in the current version:

Errors from shellcheck:

In ./hack/module-graph.sh line 34:
go mod graph | egrep "^.*k8s.io.*k8s.io.*$" | sed -E 's|@\S+ | |g' | sed -E 's|@\S+$||g' | sed -E 's/[\.\/\-]/_/g' | sed -E 's| | -> |g' | sed -E 's|^|    |g' >> _output/module-dependencies.dot || error_exit "Failed to generate dependencies in DOT file"
               ^---^ SC2196: egrep is non-standard and deprecated. Use grep -E instead.

For more information:
  https://www.shellcheck.net/wiki/SC2196 -- egrep is non-standard and depreca...

Please review the above warnings. You can test via "./hack/verify-shellcheck.sh"
If the above warnings do not make sense, you can exempt this package from shellcheck
checking by adding it to hack/.shellcheck_failures (if your reviewer is okay with it).

hack/module-graph.sh
cheftako: Little torn. The first graph is definitely simpler, easier to read and nicer looking. Most of the nodes which are removed are leaf nodes and I don't think they are really providing any value. However at least three of the nodes which the change removes are not leaf nodes (utils, kube_openapi, apiserver_network_proxy_konnectivity_client). Tempted to put in flag to let user decide on which. Also found we have duplicate lines which is part of why the first is so messy. Added fix to dedupe lines.

cheftako: module-dependencies

liggitt: I don't feel that strongly, I just thought the goal was to find dependencies between modules that live in k8s.io/kubernetes

cheftako: I was thinking more understanding the dependencies between modules we (kubernetes) own/control.

cheftako:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92445/pull-kubernetes-verify/1276250204529496065 show pull-kubernetes-verify as failed, 31/31 Tests Passed!

Issue #80177, opened 2019-07-15
e2e: remove usage of the DeprecatedMightBeMasterNode function from the e2e testing framework


neolit123:
EDIT: the function was renamed to DeprecatedMightBeMasterNode, but the remove action is still pending.

instead of making assumptions that a node is "master" or not, we should make the choices more granular - e.g. can this node host workloads? we should remove usage of the function `IsMasterNode` from `e2e/framework`

see discussion here:
https://github.com/kubernetes/kubernetes/pull/80021#issuecomment-511507244

part of:
https://github.com/kubernetes/kubernetes/issues/75601

help wanted!

cc @andrewsykim @oomichi @alejandrox1 
/area e2e-test-framework
/sig testing
/kind cleanup

BenTheElder:
cc

neolit123:
> Should IsMasterNode be replaced with something?

that's the tricky part...
there is no sane way in kubernetes to tell if a node hosts a control plane or not.

which means that any calls to IsMasterNode have to be replaced with something that solves the underlying problem - for example are we calling IsMasterNode because we want to know if this node accepts workloads? or are we calling it to determine if there is a controller-manager on the node so that we can scrape metrics from it.

oomichi:
@neolit123 
This issue seems interesting for me, because I am facing a conformance test issue which seems related to this on my local environment.
Even if a node is master, the node can host workload if the taint is removed from the node.
So do we want to use some function which checks a node is tainted instead of current  IsMasterNode()?

oomichi:
Some investigation related to this issue:
Now all 7 callers of `IsMasterNode()` are
```
e2e/framework/metrics/metrics_grabber.go:               if system.IsMasterNode(node.Name) {
e2e/framework/metrics/latencies.go:             if system.IsMasterNode(node.Name) {
e2e/framework/resource_usage_gatherer.go:               if (options.Nodes == MasterNodes) && !system.IsMasterNode(pod.Spec.NodeName) {
e2e/framework/resource_usage_gatherer.go:               if (options.Nodes == MasterAndDNSNodes) && !system.IsMasterNode(pod.Spec.NodeName) && pod.Labels["k8s-app"] != "kube-dns" {
e2e/framework/resource_usage_gatherer.go:               if options.Nodes == AllNodes || system.IsMasterNode(node.Name) || dnsNodes[node.Name] {
e2e/framework/node/resource.go:         if system.IsMasterNode(n.Name) {
[...]

PR #92449, opened 2020-06-24
Use GetReadySchedulableNodes() instead

oomichi:
**What type of PR is this?**

/kind cleanup

GetMasterAndWorkerNodes() calls DeprecatedMightBeMasterNode() which is deprecated function as the name.
This replaces GetMasterAndWorkerNodes() with GetReadySchedulableNodes() as possible to reduce the deprecated function usage.

```release-note
NONE
```


oomichi:
become unnecessary..

/close

PR #92450, opened 2020-06-24
Use worker nodes for WaitForStableCluster()

oomichi:
**What type of PR is this?**

/kind cleanup

WaitForStableCluster() checks all pods run on worker nodes, and the function used to refer master nodes to skip checking controller plane pods.
GetMasterAndWorkerNodes() was used for getting master nodes, but the implementation is not good because it usesDeprecatedMightBeMasterNode().

This makes WaitForStableCluster() refer worker nodes directly to avoid using GetMasterAndWorkerNodes().

```release-note
NONE
```

### DIFF
--- a/test/e2e/scheduling/predicates.go
+++ b/test/e2e/scheduling/predicates.go
@@ -53,7 +53,7 @@ const (
 var localStorageVersion = utilversion.MustParseSemantic("v1.8.0-beta.0")
 
 // variable set in BeforeEach, never modified afterwards
-var masterNodes sets.String
+var workerNodes sets.String
 
 type pausePodConfig struct {
 	Name                              string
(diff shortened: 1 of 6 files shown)


test/e2e/scheduling/predicates.go
```diff
@@ -95,17 +95,14 @@ var _ = SIGDescribe("SchedulerPredicates [Serial]", func() {
 
 		framework.AllNodesReady(cs, time.Minute)
 
-		// NOTE: Here doesn't get nodeList for supporting a master nodes which can host workload pods.
-		masterNodes, _, err = e2enode.GetMasterAndWorkerNodes(cs)
```
Huang-Wei, line 99: Can we remove e2enode.GetMasterAndWorkerNodes(), or mark it as deprecated?

test/e2e/scheduling/predicates.go
```diff
@@ -135,7 +132,7 @@ var _ = SIGDescribe("SchedulerPredicates [Serial]", func() {
 				nodeMaxAllocatable = allocatable.Value()
 			}
 		}
-		WaitForStableCluster(cs, masterNodes)
+		WaitForStableCluster(cs, workerNodes)
```
Huang-Wei, line 135: How about updating the parameter `masterNodes` of WaitForStableCluster() and getScheduledAndUnscheduledPods to `workerNodes`?

Huang-Wei (commented):
Some nits.

test/e2e/scheduling/predicates.go
oomichi: test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go still calls e2enode.GetMasterAndWorkerNodes() and it takes more time for removing that.
It is a good idea to make it as deprecated.
I will update this PR soon.

test/e2e/scheduling/predicates.go
oomichi: Nice point, I forgot to update it (^^;)

PR #84016, opened 2019-10-16
Implementing the storage version API, Part I

caesarxuchao:
/sig-apimachinery
/assign @deads2k @yliaog @lavalamp 
/kind feature
[KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190802-dynamic-coordinated-storage-version.md)

This PR implements:
1. the storage version API.
2. Let the kube-aggregator host the storage version API.
3. kube-apiserver updates the storage version during bootstrap.
4. A filter that blocks certain requests during kube-apiserver bootstrap until the kube-apiserver has updated the storage versions.
5. An integration test that shows the filer works, and that the apiserver works properly after the storage version has been updated.

This PR is already huge, so I'll do the remaining tasks in follow-up PRs. Some major follow-ups are:
1. the filter should block write requests to "apiservices" and "crd".
2. the garbage collection of obsolete storage version objects.
3. a test showing that the API works fine in HA cluster.
4. custom resource handler needs to update the storage version the CR before accepting write requests.

PR #92459, opened 2020-06-24
Add storageversion manager interface

roycaihw:
[KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190802-dynamic-coordinated-storage-version.md). Split off from #84016.

```release-note
NONE
```

/sig api-machinery
/priority important-soon
/cc @caesarxuchao

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
+++ b/staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
@@ -1,0 +1,194 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package storageversion
+
+import (
+	"sync"
+	"sync/atomic"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	apiserverclientset "k8s.io/apiserver/pkg/client/clientset_generated/clientset"
+	"k8s.io/client-go/rest"
+	_ "k8s.io/component-base/metrics/prometheus/workqueue" // for workqueue metric registration
+	"k8s.io/klog"
+)
+
+// ResourceInfo contains the information to register the resource to the
+// storage version API.
+type ResourceInfo struct {
+	Resource metav1.APIResource
+	// We use a standalone Group instead of reusing the Resource.Group
+	// because Resource.Group is often omitted, see the comment on
+	// Resource.Group for why it's omitted.
+	Group                    string
+	EncodingVersion          string
+	DecodableVersions        []string
+	EquivalentResourceMapper runtime.EquivalentResourceRegistry
+}
+
+// Manager records the resources whose StorageVersions need updates, and provides a method to update those StorageVersions.
+type Manager interface {
+	// AddResourceInfo adds ResourceInfo to the manager.
+	AddResourceInfo(resources ...*ResourceInfo)
+	// RemoveResourceInfo removes ResourceInfo from the manager.
+	RemoveResourceInfo(r *ResourceInfo)
+	// UpdatesPending returns if the StorageVersion of a resource is still wait to be updated.
+	UpdatesPending(group, resource string) bool
+
+	// UpdateStorageVersions updates the StorageVersions.
+	UpdateStorageVersions(loopbackClientConfig *rest.Config, apiserverID string)
+	// Completed returns if updating StorageVersions has completed.
+	Completed() bool
+}
+
+var _ Manager = &DefaultManager{}
+
+// NewDefaultManager creates a new DefaultManager.
+func NewDefaultManager() *DefaultManager {
+	s := &DefaultManager{}
+	s.completed.Store(false)
+	s.groupResources = make(map[string]map[string]struct{})
+	s.resources = make(map[*ResourceInfo]struct{})
+	return s
+}
+
+// AddResourceInfo adds ResourceInfo to the manager.
+// This is not thread-safe. It is expected to be called when the apiserver is installing the endpoints, which is done serially.
+func (s *DefaultManager) AddResourceInfo(resources ...*ResourceInfo) {
+	for _, r := range resources {
+		s.resources[r] = struct{}{}
+		s.addGroupResourceFor(r)
+	}
+}
+
+func (s *DefaultManager) addGroupResourceFor(r *ResourceInfo) {
+	gvrs := r.EquivalentResourceMapper.EquivalentResourcesFor(schema.GroupVersionResource{
+		Group:    r.Group,
+		Resource: r.Resource.Name,
+	}, "")
+	for _, gvr := range gvrs {
+		s.addGroupResource(gvr.Group, gvr.Resource)
+	}
+}
+
+func (s *DefaultManager) addGroupResource(group, resource string) {
+	if _, ok := s.groupResources[group]; !ok {
+		s.groupResources[group] = make(map[string]struct{})
+	}
+	s.groupResources[group][resource] = struct{}{}
+}
+
+// RemoveResourceInfo removes ResourceInfo from the manager.
+// It is not safe to call this function concurrently with AddResourceInfo.
+func (s *DefaultManager) RemoveResourceInfo(r *ResourceInfo) {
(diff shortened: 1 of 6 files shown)


roycaihw:
Rebased on top of https://github.com/kubernetes/kubernetes/pull/92064. Only the last 3 commits are new

roycaihw:
cc @ptabor @wojtek-t

/retest

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
```diff
@@ -46,13 +48,15 @@ type ResourceInfo struct {
 type Manager interface {
 	// AddResourceInfo adds ResourceInfo to the manager.
 	AddResourceInfo(resources ...*ResourceInfo)
-	// RemoveResourceInfo removes ResourceInfo from the manager.
-	RemoveResourceInfo(r *ResourceInfo)
+	// MarkCompletedResourceInfo marks updated ResourceInfo as completed.
+	MarkCompletedResourceInfo(r *ResourceInfo)
+	// RecordErrorResourceInfo records latest error updating ResourceInfo.
+	RecordErrorResourceInfo(r *ResourceInfo, err error)
```
caesarxuchao, line 54: Can we unify the two method names? Maybe `RecordResourceInfoSuccess` and `RecordResourceInfoFailure`.

staging/src/k8s.io/apiserver/pkg/storageversion/updater.go
```diff
@@ -82,35 +82,37 @@ func singleUpdate(c Client, apiserverID, resource, encodingVersion string, decod
...
 		return err
 	}
-	_, err = c.Update(context.TODO(), sv, metav1.UpdateOptions{})
+	_, err = c.Update(context.TODO(), updatedSV, metav1.UpdateOptions{})
 	return err
 }
 
-func localUpdateStorageVersion(sv *v1alpha1.StorageVersion, apiserverID, encodingVersion string, decodableVersions []string) {
+func localUpdateStorageVersion(sv *v1alpha1.StorageVersion, apiserverID, encodingVersion string, decodableVersions []string) *v1alpha1.StorageVersion {
+	updatedSV := sv.DeepCopy()
```
caesarxuchao, line 100: Why do we make a deepCopy instead of modifying the original sv?

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
```diff
@@ -168,26 +194,32 @@ func decodableVersions(e runtime.EquivalentResourceRegistry, group string, resou
 
 // UpdateStorageVersions updates the StorageVersions. If the updates are
 // successful, following calls to Completed() returns true.
-func (s *DefaultManager) UpdateStorageVersions(loopbackClientConfig *rest.Config, serverID string) {
-	cfg := rest.AddUserAgent(loopbackClientConfig, "system:kube-apiserver")
-	clientset, err := apiserverclientset.NewForConfig(cfg)
+func (s *DefaultManager) UpdateStorageVersions(kubeAPIServerClientConfig *rest.Config, serverID string) {
+	cfg := rest.AddUserAgent(kubeAPIServerClientConfig, "system:kube-apiserver")
```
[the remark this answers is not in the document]
roycaihw: Appending user-agent is not needed. Removed this line

roycaihw:
This PR is ready for another round of review. PTAL

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
```diff
@@ -0,0 +1,230 @@
...
+type Manager interface {
+	// AddResourceInfo records resources whose StorageVersions need updates
+	AddResourceInfo(resources ...*ResourceInfo)
+	// UpdateStorageVersions tries to update the StorageVersions of the recorded resources
+	UpdateStorageVersions(kubeAPIServerClientConfig *rest.Config, apiserverID string)
+	// PendingUpdate returns true if the StorageVersion of the given resource is still pending update.
+	PendingUpdate(gr schema.GroupResource) bool
+	// LastUpdateError returns the last error hit when updating the storage version of the given resource.
+	LastUpdateError(gr schema.GroupResource) error
+	// Completed returns true if updating StorageVersions has completed.
```
caesarxuchao, line 59: nit: "Completed returns true if updating StorageVersions of all recorded resources has completed."

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
```diff
@@ -0,0 +1,230 @@
...
+	mu sync.RWMutex
+	// managedResourceInfos records the ResourceInfos whose StorageVersions will get updated in the next
+	// UpdateStorageVersions call
+	managedResourceInfos map[*ResourceInfo]struct{}
+	// managedStatus records the status of storage version update for the GroupResources. It allows quick
+	// status lookup for a GroupResource.
+	managedStatus map[schema.GroupResource]*storageVersionStatus
+}
+
+type storageVersionStatus struct {
```
caesarxuchao, line 78: `storageVersionStatus` sounds like the status subresource of the storageVersion API. How about renaming it to `updateStatus`?

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
```diff
@@ -0,0 +1,230 @@
...
+type defaultManager struct {
+	completed atomic.Value
+
+	mu sync.RWMutex
+	// managedResourceInfos records the ResourceInfos whose StorageVersions will get updated in the next
+	// UpdateStorageVersions call
+	managedResourceInfos map[*ResourceInfo]struct{}
+	// managedStatus records the status of storage version update for the GroupResources. It allows quick
+	// status lookup for a GroupResource.
+	managedStatus map[schema.GroupResource]*storageVersionStatus
```
caesarxuchao, line 75: Can you document why we need both `managedResourceInfos` and `managedStatus`, and can't combine them as a `map[*ResourceInfo]*storageVersionStatus`?

(I think it's because one ResourceInfo might expand into multiple GroupResource, but maybe there are other reasons as well).

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
```diff
@@ -0,0 +1,230 @@
...
+func (s *defaultManager) AddResourceInfo(resources ...*ResourceInfo) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	for _, r := range resources {
+		s.managedResourceInfos[r] = struct{}{}
+		s.addManagedStatusLocked(r)
+	}
+}
+
+func (s *defaultManager) addManagedStatusLocked(r *ResourceInfo) {
```
caesarxuchao, line 102: addPendingManagedStatusLocked?

roycaihw:
@caesarxuchao Thanks for the comments. I addressed them. PTAL

caesarxuchao:
lgtm

roycaihw:
Thanks @caesarxuchao!

@deads2k Could you please take a look when you get a chance?

roycaihw:
I squashed the commits

roycaihw:
@deads2k Hi David, could you approve this change when you get a chance? Thanks!

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
```diff
@@ -0,0 +1,231 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	_ "k8s.io/component-base/metrics/prometheus/workqueue" // for workqueue metric registration
+	"k8s.io/klog/v2"
+)
+
+// ResourceInfo contains the information to register the resource to the
+// storage version API.
+type ResourceInfo struct {
+	Resource metav1.APIResource
```
deads2k, line 37: a quick scan of usage suggests that we only use this for its name. Can we instead track a `schema.GroupResource` that contains exactly the info we use?

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
```diff
@@ -0,0 +1,231 @@
...
+func (s *defaultManager) UpdateStorageVersions(kubeAPIServerClientConfig *rest.Config, serverID string) {
+	clientset, err := kubernetes.NewForConfig(kubeAPIServerClientConfig)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("failed to get clientset: %v", err))
+		return
+	}
+	sc := clientset.InternalV1alpha1().StorageVersions()
+
+	s.mu.RLock()
+	resources := s.managedResourceInfos
```
deads2k, line 123: I  haven't played with this in golang in a long time, but doesn't this set `resources` to point to the same map as `s.managedResourceInfos` so that when you iterate on it down below, if the content changes in a different thread you end up with a data race?  I didn't think that this performed a deep copy. 

if the map is small and update is long, a deep copy seems appropriate

roycaihw: good catch! I added a copy

deads2k:
/approve

please squash.

roycaihw:
Squashed. Thanks for reviewing

caesarxuchao:
/lgtm

Thanks @deads2k for the reviews.

Issue #92057, opened 2020-06-12
HostPathType Socket e2e tests hang on containerd


msau42:
**What happened**:
Once we switched over the default runtime to containerd, the slow ci and pull jobs started timing out. The tests, which used to take 13 minutes, jumped to 30 minutes. https://github.com/kubernetes/kubernetes/issues/92045

This increase in test time is only observed in the slow job, so I suspect the slow job is exercising some functionality that may not be exercised as frequently in the other jobs.

**What you expected to happen**:
No performance regression

**How to reproduce it (as minimally and precisely as possible)**:
Run the tests in the slow job with containerd

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

msau42:
@kubernetes/sig-node-bugs @kubernetes/sig-storage-bugs

BenTheElder:
do we know what tests were regressing?

msau42:
I did some analysis in https://github.com/kubernetes/kubernetes/issues/92045, but basically, no single test stood out in terms of increasing the time took. I suspect that all the test cases started taking a few seconds longer, and the sum of all of contributed to the increase in job time.

msau42:
I was mainly eyeballing the graphs, but we probably should get the raw data and do some more detailed analysis on it

msau42:
@kubernetes/sig-scalability-bugs do we have scale runs with the [slow] tests?

msau42:
As an experiment, I changed the runtime back to docker in https://github.com/kubernetes/kubernetes/pull/92056, and the slow job tests went back down to 13 minutes.

msau42:
@mikebrow pointed out that https://github.com/kubernetes-sigs/cri-tools/pull/627 may be relevant

msau42:
@Random-Liu pointed out that cri-tools is not used in the kubelet path, so that particular issue may not be the cause. But maybe we can investigate image pulling as a first start.

FWIW, there are at least 100+ test cases running in this job that create the same container on the same node multiple times: hostpathtype, multivolume, container restart

msau42:
I think I found it. All the HostPath Socket tests are consistently never completing.

jingxu97:
[...]

PR #92465, opened 2020-06-24
Convert exec probe timeout errors into failures

### LINKED ISSUE #92464: exec probe timeout failures cannot trigger container restart
**What happened**:
When using containerd as CRI runtime, a Pod with exec liveness probe cannot be restarted by timeout errors, though the errors were detected and liveness probe events were recorded correctly:
```
  Normal   Created    2m6s                kubelet, k8s-01    Created container test
  Normal   Started    2m6s                kubelet, k8s-01    Started container test
  Warning  Unhealthy  2s (x12 over 112s)  kubelet, k8s-01    Liveness probe errored: rpc error: code = DeadlineExceeded desc = failed to exec in container: timeout 2s exceeded: context deadline exceeded
```
```
Containers:
  test:
    Container ID:  containerd://399a7b6db1b32cc49df9723a9f30217b3d0e9bbc1424272c0a2fdc293ab74bb6
    Image:         nginx
    Image ID:      docker.io/library/nginx@sha256:21f32f6c08406306d822a0e6e8b7dc81f53f336570e852e25fbe1e3e3d0d0133
    Port:          <none>
    Host Port:     <none>
    Command:
      sleep
      3600
    State:          Running
      Started:      Wed, 24 Jun 2020 04:46:06 -0700
    Ready:          True
    Restart Count:  0
    Liveness:       exec [sleep 4] delay=0s timeout=2s period=10s #success=1 #failure=3
    Environment:    <none>
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-24hfm (ro)
```

The reason is the errors other than `exec.ExitError` are converted into `Unknown` Result with error returned, causing kubelet throws away the result.
https://github.com/kubernetes/kubernetes/blob/4de511a4cfdb8e49e804be18245911bf3270fb2f/pkg/probe/exec/exec.go#L61-L70

Like http and tcp probes, exec timeout errors should be converted as `Failure` Result and should trigger restarting container.

1. Use containerd as CRI
2. Create a Pod and make its exec probe timeout
```
apiVersion: v1
kind: Pod
metadata:
  name: testpod
spec:
  restartPolicy: Always
  containers:
  - name: test
    image: nginx
    command:
      - sleep
      - "3600"
    livenessProbe:
      timeoutSeconds: 2
      exec:
        command:
        - sleep
        - "4"
```
The Liveness probe error events will be generated but the container won't be restarted.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others: containerd v1.3.4

tnqn:
**What type of PR is this?**
/kind bug

The exec probe timeout errors are converted into Unknown Result with error returned, causing kubelet throws away the result and the container isn't restarted as expected. Like http and tcp probes, timeout errors should be converted into Failure Result with no error returned.

Fixes #92464

```release-note
Fixed exec probe timeout failures cannot trigger container restart when using containerd as CRI runtime
```

```docs

```

### DIFF
--- a/pkg/probe/exec/exec.go
+++ b/pkg/probe/exec/exec.go
@@ -22,6 +22,8 @@ import (
 	"k8s.io/kubernetes/pkg/kubelet/util/ioutils"
 	"k8s.io/kubernetes/pkg/probe"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/exec"
 )
@@ -59,6 +61,10 @@ func (pr execProber) Probe(e exec.Cmd) (probe.Result, string, error) {
 
 	klog.V(4).Infof("Exec probe response: %q", string(data))
 	if err != nil {
+		// Convert timeout errors into failures.
+		if status.Code(err) == codes.DeadlineExceeded {
+			return probe.Failure, err.Error(), nil
+		}
 		exit, ok := err.(exec.ExitError)
 		if ok {
 			if exit.ExitStatus() == 0 {
--- a/pkg/probe/exec/BUILD
+++ b/pkg/probe/exec/BUILD
@@ -15,6 +15,8 @@ go_library(
         "//pkg/probe:go_default_library",
         "//vendor/k8s.io/klog/v2:go_default_library",
         "//vendor/k8s.io/utils/exec:go_default_library",
+        "//vendor/google.golang.org/grpc/codes:go_default_library",
+        "//vendor/google.golang.org/grpc/status:go_default_library",
     ],
 )
 
@@ -22,7 +24,11 @@ go_test(
     name = "go_default_test",
     srcs = ["exec_test.go"],
     embed = [":go_default_library"],
-    deps = ["//pkg/probe:go_default_library"],
+    deps = [
+        "//pkg/probe:go_default_library",
+        "//vendor/google.golang.org/grpc/codes:go_default_library",
+        "//vendor/google.golang.org/grpc/status:go_default_library",
+    ],
 )
 
 filegroup(
--- a/pkg/probe/exec/exec_test.go
+++ b/pkg/probe/exec/exec_test.go
@@ -22,6 +22,8 @@ import (
 	"strings"
 	"testing"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"k8s.io/kubernetes/pkg/probe"
 )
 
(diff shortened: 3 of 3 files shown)


pkg/probe/exec/exec.go
```diff
@@ -59,6 +61,10 @@ func (pr execProber) Probe(e exec.Cmd) (probe.Result, string, error) {
 
 	klog.V(4).Infof("Exec probe response: %q", string(data))
 	if err != nil {
+		// Convert timeout errors into failures.
+		if status.Code(err) == codes.DeadlineExceeded {
```
SergeyKanzhelev, line 65: this code assumes that timeout will be implemented at grpc layer, not inside the containerD. It introduces tight coupling between with the implementation.

Did you check if `exec.Exited()` was set to `false` in case of timeout?

tnqn: Thanks for the review.
containerd implements the timeout with context timeout and converts the `context.DeadlineExceeded` to grpc `DeadlineExceeded` as response. There is no `ExecSyncResponse` in case of timeout.
https://github.com/containerd/cri/blob/09d6426f33cac217528158ddc6d254ca7d597a7b/pkg/server/container_execsync.go#L171
https://github.com/containerd/containerd/blob/c75180740937d4b2d44b9c1edc1c27b208e66e32/errdefs/grpc.go#L61-L62

I also thought it's too coupled to check grpc error here but didn't find a proper error struct to represent the timeout error.
How about this option:
At rpc client side, convert grpc DeadlineExceeded to `context.DeadlineExceeded` or a new error struct that implements `ExitError`, or the `CodeExitError` if it's acceptable. If `context.DeadlineExceeded`, the `Probe` method needs to convert it to `Failure` result.
https://github.com/kubernetes/kubernetes/blob/4de511a4cfdb8e49e804be18245911bf3270fb2f/pkg/kubelet/cri/remote/remote_runtime.go#L387-L391

SergeyKanzhelev: There was a discussion at sig node meeting lately. And mentioned this comment: https://github.com/kubernetes/kubernetes/issues/92057#issuecomment-644915304 I think it's worth understanding if the timeout was passed correctly to containerd. This may be a better fix.

tnqn: It looks like https://github.com/kubernetes/kubernetes/issues/92057#issuecomment-644915304 was talking about lifecycle event hander, not probe handler?
For probe handler, I think it's passed to container runtime correctly:
https://github.com/kubernetes/kubernetes/blob/e7ca64fbe16d0c4b6c7b36aecde9cd75042b2828/pkg/kubelet/prober/prober.go#L157-L161
https://github.com/kubernetes/kubernetes/blob/e7ca64fbe16d0c4b6c7b36aecde9cd75042b2828/pkg/kubelet/kuberuntime/kuberuntime_container.go#L826-L827
https://github.com/kubernetes/kubernetes/blob/e7ca64fbe16d0c4b6c7b36aecde9cd75042b2828/pkg/kubelet/cri/remote/remote_runtime.go#L382-L387

Besides, the probe failure event in https://github.com/kubernetes/kubernetes/issues/92464#issue-644561356 `Liveness probe errored: rpc error: code = DeadlineExceeded desc = failed to exec in container: timeout 2s exceeded: context deadline exceeded` has the configured timeout `2s`, which was returned by containerd:
https://github.com/containerd/cri/blob/master/pkg/server/container_execsync.go#L171

SergeyKanzhelev: Thank you for all the context. The fact that grpc timeout creeped all the way here confused me. I like your previous comment on converting grpc code back to some non-grpc failure on CRI side. Sounds like a good solution. 

@dchen1107 @yujuhong what do you think?

SergeyKanzhelev:
This PR solves the same problem as the PR that is currently being actively discussed https://github.com/kubernetes/kubernetes/pull/94115/. The agreement from SIG node meeting today was that the code change is only a small part of the fix and we need to actively look at side effects this change may bring to the production payloads. Please join the discussion at the PR: https://github.com/kubernetes/kubernetes/pull/94115/.

/close

PR #92466, opened 2020-06-24
Improve observability of node authorizer

mborsz:
**What type of PR is this?**
/kind feature

It make it possible to check if node authorizer is working and if it's able to process all incoming events.

Fixes #

```release-note
Add node_authorizer_actions_duration_seconds metric that can be used to estimate load to node authorizer.
```

```docs

```

/assign @wojtek-t 
/assign @liggitt


plugin/pkg/auth/authorizer/node/graph.go
```diff
@@ -94,6 +95,8 @@ type namespaceVertexMapping map[string]nameVertexMapping
 type nameVertexMapping map[string]*namedVertex
 
 func NewGraph() *Graph {
+	RegisterMetrics()
```
liggitt, line 98: we should not register global metrics as a side-effect of creating what was previously an isolated object. metrics should be set up in a separate path we are sure is a singleton in the process

plugin/pkg/auth/authorizer/node/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+// NodeAuthorizerSubsystem is a subsystem used by metrics here.
+const NodeAuthorizerSubsystem = "node_authorizer"
+
+var (
+	// NodeAuthorizerActionsDuration is a histogram of duration of graph actions in node authorizer.
+	NodeAuthorizerActionsDuration = metrics.NewHistogramVec(
```
liggitt, line 31: is there a reason to export this? do we want things from other packages referencing this?

plugin/pkg/auth/authorizer/node/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+		&metrics.HistogramOpts{
+			Subsystem:      NodeAuthorizerSubsystem,
+			Name:           "actions_duration_seconds",
+			Help:           "Histogram of duration of graph actions in node authorizer.",
+			StabilityLevel: metrics.ALPHA,
+			// Start with 0.1ms with the last bucket being [~200ms, Inf)
+			Buckets: metrics.ExponentialBuckets(0.0001, 2, 12),
```
liggitt, line 38: cc @logicalhan @kubernetes/sig-instrumentation-pr-reviews for feedback on metric type/name/construction

plugin/pkg/auth/authorizer/node/graph.go
liggitt: an alternative place to instrument this would be in the graph populator (which could take an already-registered `*HistogramVec` parameter), where we wire up informers to call the Add/Update/Delete* methods

wojtek-t: Why can't we simply register it in init() function in this package? There are already multiple places in our codebase doing that, e.g.:
https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/util/metrics.go#L79

liggitt: I'm -1 on adding more places where we register in init()... I consider that an anti-pattern we should get rid of, not extend

mborsz: I copied this pattern from endpointslice controller: https://github.com/kubernetes/kubernetes/blob/eb9bca5f060ab1410a5228dc77026986e32505eb/pkg/controller/endpointslice/endpointslice_controller.go#L93

but I agree, it's quite ugly. I will check what other components do.

wojtek-t: @liggitt - for my own education, why do you consider that anti-pattern? init() guarantees you "exactly once" and guarantees metrics registration if a given package is used.

liggitt: because it registers metrics for anyone that imports the package, which isn't always desired

plugin/pkg/auth/authorizer/node/metrics.go
mborsz: Done

plugin/pkg/auth/authorizer/node/graph.go
mborsz: Other places often use func init() { legacyregistry.MustRegister() }

I migrated to this pattern.

plugin/pkg/auth/authorizer/node/metrics.go
```diff
@@ -0,0 +1,42 @@
...
+		},
+		[]string{"operation"},
+	)
+)
+
+func init() {
```
logicalhan, line 40: Would you mind modifying this to the following signature?

```golang
func RegisterMetrics(r Registry) {
    r.MustRegister(graphActionsDuration)
}
```

And then instead of importing this package to instantiate the metric as a side-effect, we can explicitly invoke the MustRegister function with the metric registry in some boot sequence code-path. This is the pattern we are trying to move towards. We do this currently in the kubelet (https://github.com/kubernetes/kubernetes/blob/v1.19.0-alpha.0/pkg/kubelet/server/server.go#L368-L373). 

You can also just pass in the legacy registry instead of a custom one, if your metrics endpoint is already just using the legacyregistry. This will make it easier for us to get rid of the legacyregistry later on (which we intend to do, since it is deprecated).

plugin/pkg/auth/authorizer/node/graph.go
logicalhan: I added a comment below for an alternative (preferred) pattern.

plugin/pkg/auth/authorizer/node/metrics.go
```diff
@@ -0,0 +1,42 @@
...
+
+var (
+	graphActionsDuration = metrics.NewHistogramVec(
+		&metrics.HistogramOpts{
+			Subsystem:      nodeAuthorizerSubsystem,
+			Name:           "graph_actions_duration_seconds",
+			Help:           "Histogram of duration of graph actions in node authorizer.",
+			StabilityLevel: metrics.ALPHA,
+			// Start with 0.1ms with the last bucket being [~200ms, Inf)
+			Buckets: metrics.ExponentialBuckets(0.0001, 2, 12),
```
logicalhan, line 34: Are you sure you don't need granularity for this metric above 200 milliseconds?

plugin/pkg/auth/authorizer/node/metrics.go
logicalhan: actions seems a bit generic but the metric suffix is in line with guidance (i.e. duration_seconds)

plugin/pkg/auth/authorizer/node/graph.go
mborsz: I used alternative approach suggested below.

plugin/pkg/auth/authorizer/node/metrics.go
mborsz: I changed to this pattern with one diff: legancyregistry is a package and not an object, so I cannot call RegisterMetrics(legacyregistry). I changed the signature to accept function to call, like: RegisterMetrics(legacyregistry.MustRegister). Please let me know if this works.

plugin/pkg/auth/authorizer/node/metrics.go
mborsz: It should. Graph struct is doing purely in-memory operations, so 200 ms for a single call seems to be way too much.

mborsz:
PTAL

ehashman:
@mborsz can we get tests passing for review? At least one of these failures doesn't look like a flake at a glance

wojtek-t:
Agree - this one seems relevant: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/92466/pull-kubernetes-integration/1321459745499910144
@mborsz - can you please take a look?

plugin/pkg/auth/authorizer/node/metrics.go
mborsz: It's not possible to do that this way with legacyregistry:
* I added a node.RegisterMetrics(legacyregistry.MustRegistry) call to node/config (where we decide if we are going to use node authorizer)
* This node/config can be called multiple times in a single binary (e.g. in our tests)
* This triggers panic due to multiple metric registration
* I have to add sync.Once to dedupe metrics registration, but this way, if RegisterMetric will be called twice for multiple metric registries, the second call will be completely ignored leading to not registering this
* I changed this to the pattern that is used everywhere in kube-apiserver for legacyregistry: hardcoding legacyregistry in MustRegister and calling MustRegister on the node/config which seems to be the best option we have currently in kube-apiserver.

mborsz:
Sorry, tests should be fixed now.

mborsz:
Friendly ping

logicalhan (approved):
/lgtm
/approve

plugin/pkg/auth/authorizer/node/metrics.go
logicalhan: Yeah, it's a bit entangled in the apiserver currently.

plugin/pkg/auth/authorizer/node/graph.go
```diff
@@ -334,6 +335,10 @@ func (g *Graph) recomputeDestinationIndex_locked(n graph.Node) {
 //   pvc       -> pod
 //   svcacct   -> pod
 func (g *Graph) AddPod(pod *corev1.Pod) {
+	start := time.Now()
+	defer func() {
+		graphActionsDuration.WithLabelValues("AddPod").Observe(time.Since(start).Seconds())
+	}()
```
wojtek-t, line 341: Why is DeletePod not intstrumented?

wojtek-t (commented):
Just one small comment from me.

plugin/pkg/auth/authorizer/node/graph.go
mborsz: Yes, good point. I added metrics to DeletePod and SetNodeConfigMap.

wojtek-t:
/lgtm
/approve

[Based on Han`s approval from sig-instrumentation perspective]

PR #82416, opened 2019-09-06
Create LoadBalancer in nginx ingress tests

mm4tt:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation

/kind failing-test

> /kind feature
> /kind flake

**What this PR does / why we need it**:
It fixes the nginx ingress tests that is currently failing in the private clusters where nodes don't have public ips, see https://k8s-testgrid.appspot.com/sig-scalability-experiments#gce-private-cluster-correctness.
It achieves that by creating a load balancer in front of the nginx controller and using the [`--publish-service` nginx ingress feature](https://github.com/kubernetes/ingress-nginx/tree/master/docs/examples/static-ip)

**Which issue(s) this PR fixes**:

Fixes https://github.com/kubernetes/kubernetes/issues/77538

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #92467, opened 2020-06-24
Backport #82416 to 1.16: Create LoadBalancer in nginx ingress tests

### LINKED ISSUE #77538: Some [sig-network] tests don't work on "private" clusters
Hello sig-network,

We, sig-scalability, are exploring using "private" clusters (i.e. the ones where nodes don't have public IPs) for our large (up to 5K nodes) scalability tests to avoid consuming expensive public IPs. More details in https://github.com/kubernetes/kubernetes/issues/76374

With a few minor modifications we managed to get an almost passing correctness test suite on a private cluster except the 3 sig-network tests which explicitly assume that nodes are reachable over public IPs. 

**Failing tests** ([testgrid link](https://k8s-testgrid.appspot.com/sig-scalability-experiments#gce-private-cluster-correctness)):
* [sig-network] Firewall rule should have correct firewall rules for e2e cluster
* [sig-network] Loadbalancing: L7 [Slow] Nginx should conform to Ingress spec
* [sig-network] Services should be able to create a functioning NodePort service

Could you take a look at them and advise if it would be possible to rewrite them so they can work both in the "public" and "private" clusters? 

We'd really appreciate your help assistance here.
Let me know if you have any questions.

Thanks,

Matt

jkaniuk:
Backport #82416

**What type of PR is this?**
/kind failing-test

It fixes the nginx ingress tests that is currently failing in the private clusters where nodes don't have public ips, see https://k8s-testgrid.appspot.com/sig-scalability-experiments#gce-private-cluster-correctness.
It achieves that by creating a load balancer in front of the nginx controller and using the [`--publish-service` nginx ingress feature](https://github.com/kubernetes/ingress-nginx/tree/master/docs/examples/static-ip)

Fixes #77538 for 1.16

```
NONE
```

/cc mm4tt
/cc wojtek-t

### DIFF
--- a/test/e2e/framework/ingress/ingress_utils.go
+++ b/test/e2e/framework/ingress/ingress_utils.go
@@ -831,6 +831,7 @@ type NginxIngressController struct {
 	rc     *v1.ReplicationController
 	pod    *v1.Pod
 	Client clientset.Interface
+	lbSvc  *v1.Service
 }
 
 // Init initializes the NginxIngressController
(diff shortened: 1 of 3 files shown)


jkaniuk:
Abandoning after discussion with Wojtek

Issue #92248, opened 2020-06-18
[ERROR] plugin/errors: 2 selasticsearch.k8smon.svc.k8smon.svc.cluster.local. AAAA: dial tcp 10.7.50.50:53: i/o timeout


MarcoPalomo:
Hello /sig Network ,
I use the DNSCache manifest for getting all the dns request cached. I try to understand how to not get theses false requests like "podname.namespace.svc.podname.namespace.svc.cluster.local"

**What happened**:

[INFO] Added back nodelocaldns rule - {filter OUTPUT [-p udp -s 10.45.0.50 --sport 53 -j ACCEPT]}                                 
[INFO] Added back nodelocaldns rule - {raw OUTPUT [-p tcp -d 10.45.0.50 --dport 53 -j NOTRACK]}                                  
[INFO] Added back nodelocaldns rule - {raw OUTPUT [-p udp -d 10.45.0.50 --dport 53 -j NOTRACK]}                                   
[INFO] Added back nodelocaldns rule - {raw OUTPUT [-p tcp -d 10.45.0.50 --dport 8080 -j NOTRACK]}                                 
[INFO] Added back nodelocaldns rule - {raw OUTPUT [-p tcp -s 10.45.0.50 --sport 8080 -j NOTRACK]} 
[ERROR] plugin/errors: 2 selasticsearch.k8smon.svc.k8smon.svc.cluster.local. AAAA: dial tcp 10.7.50.70:53: i/o timeout
[ERROR] plugin/errors: 2 selasticsearch.k8smon.svc.k8smon.svc.cluster.local. AAAA: dial tcp 10.7.50.70:53: i/o timeout
[ERROR] plugin/errors: 2 selasticsearch.k8smon.svc.k8smon.svc.cluster.local. AAAA: dial tcp 10.7.50.50:53: i/o timeout
[ERROR] plugin/errors: 2 selasticsearch.k8smon.svc.k8smon.svc.cluster.local. AAAA: dial tcp 10.7.50.50:53: i/o timeout

**What you expected to happen**:

Cache the DNS request to leverage the coredns workload !

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
data:
  Corefile: |
    cluster.local:53 {
        errors
        cache {
                success 9984 30
                denial 9984 5
        }
        reload
        loop
        bind 10.45.0.50 10.9.9.9
        forward . __PILLAR__CLUSTER__DNS__ {
                force_tcp
        }
[...]

PR #92476, opened 2020-06-24
Fix a preemption bug when pods are listed in pdb.Status.DisruptedPods 

### LINKED ISSUE #92475: Preemption logic on PDB violations
In one recent fix (#91986), we enhanced the preemption logic to respect `pdb.Status.DisruptedPods`. But there is a glitch:

https://github.com/kubernetes/kubernetes/blob/ef8380e6c0ada9ef682d2d3566cbb16b79445233/pkg/scheduler/core/generic_scheduler_test.go#L1486-L1502

- if we changed the DisruptedPods  in L1499 from "a" to "b", the test would fail
- it would also fail if we change the pods (i.e., []Pod{"b", "a"}) ordering in L1496~L1497

This is a problem we need to fix.

---

On the other hand, I'm thinking about the real semantics of `pdb.Status.DisruptedPods`. According to the docs, if a pod is matched by the "DisruptedPods" list, it denotes the pod is about to be evicted by API server as it at least has violated one disruption budget. Given this fact, IMO in scheduler preemption, we should treat this kind of pod as "nonViolating" in all cases - no matter the current `pdb.Status.DisruptionsAllowed` is negative or not. So that the scheduler would treat them as preemption candidates.

/sig scheduling

Huang-Wei:
**What type of PR is this?**

/kind bug
/sig scheduling

- The first commit resolves the 1st problem mentioned in #92248 
- The second commit (will squash later) walks one step further to treat Pods listed in `pdb.Status.DisruptedPods` as "nonViolating" in any case.

Fixes #92475

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/core/generic_scheduler.go
+++ b/pkg/scheduler/core/generic_scheduler.go
@@ -936,11 +936,15 @@ func filterPodsWithPDBViolation(pods []*v1.Pod, pdbs []*policy.PodDisruptionBudg
 				if selector.Empty() || !selector.Matches(labels.Set(pod.Labels)) {
 					continue
 				}
+
+				// Existing in DisruptedPods means it has been processed in API server,
+				// we don't treat it as a violating case.
+				if _, exist := pdb.Status.DisruptedPods[pod.Name]; exist {
+					continue
+				}
 				// Only decrement the matched pdb when it's not in its <DisruptedPods>;
 				// otherwise we may over-decrement the budget number.
-				if _, exist := pdb.Status.DisruptedPods[pod.Name]; !exist {
-					pdbsAllowed[i]--
-				}
+				pdbsAllowed[i]--
 				// We have found a matching PDB.
 				if pdbsAllowed[i] < 0 {
 					pdbForPodIsViolated = true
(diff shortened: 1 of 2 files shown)


Huang-Wei:
This PR is better being reviewed commit by commit. The 1st commit is a must. The 2nd commit is nice to have to resolve this 2nd concern I mentioned in the original issue.

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -936,15 +936,18 @@ func filterPodsWithPDBViolation(pods []*v1.Pod, pdbs []*policy.PodDisruptionBudg
...
 				}
+
+				// Existing in DisruptedPods means it has been processed in API server,
+				// we don't treat it as a violating case.
+				if _, exist := pdb.Status.DisruptedPods[pod.Name]; exist {
+					continue
+				}
+				// Only decrement the matched pdb when it's not in its <DisruptedPods>;
+				// otherwise we may over-decrement the budget number.
+				pdbsAllowed[i]--
```
ahg-g, line 947: wouldn't this make the value go into the minus? I guess it doesn't matter if that happens, right?

ahg-g (commented):
/lgtm

pkg/scheduler/core/generic_scheduler.go
Huang-Wei: Yes, pdbsAllowed[i] can be negative now.

Another style is like this:

```go
if pdbsAllowed[i] <= 0 {
    pdbForPodIsViolated = true
} else {
    pdbsAllowed[i]--
}
```

Not quite make a big difference I think :)

PR #92481, opened 2020-06-24
return nil if empty response get in memcached go-client

### LINKED ISSUE #92480: client-go returns error if serverResourcesForGroupVersion method  return 0 resources
**What happened**:
In client-go package [Memcached](https://github.com/kubernetes/kubernetes/blob/dc0122ca6a7503de3923d5026073882822d5008e/staging/src/k8s.io/client-go/discovery/cached/memory/memcache.go#L228) return error if we get zero resources of group version but ideally it should return nil. 

`serverResourcesForGroupVersion` method should return nil if it gets zero resources  

**Environment**:
- Kubernetes version : v1.18
- OS : Darwin
- Install tools: `client-go : v0.17.4`

yindia:
**What type of PR is this?**
/kind bug

In client-go package [Memcached](https://github.com/kubernetes/kubernetes/blob/dc0122ca6a7503de3923d5026073882822d5008e/staging/src/k8s.io/client-go/discovery/cached/memory/memcache.go#L228) return error if we get zero resources of group version but ideally it should return nil.

Fixes #92480

```release-note
NONE
```

```docs
NO
```

### DIFF
--- a/staging/src/k8s.io/client-go/discovery/cached/memory/memcache.go
+++ b/staging/src/k8s.io/client-go/discovery/cached/memory/memcache.go
@@ -225,7 +225,7 @@ func (d *memCacheClient) serverResourcesForGroupVersion(groupVersion string) (*m
 		return r, err
 	}
 	if len(r.APIResources) == 0 {
-		return r, fmt.Errorf("Got empty response for: %v", groupVersion)
+		return r, nil
 	}
 	return r, nil
 }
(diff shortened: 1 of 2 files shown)


caesarxuchao:
When will there be an empty group version?

Also please add a test case.

evalsocket:
@caesarxuchao Meaning if we get zero `APIResources` for a group version  then we will return nil not error
```
if len(r.APIResources) == 0 { 
		// return r, fmt.Errorf("Got empty response for: %v", groupVersion)
		return r, nil
	}
```

caesarxuchao:
@evalsocket sorry for the late reply. I mean when will server returns 0 APIResrouces for a group? Do you see that in a real cluster?

evalsocket:
@caesarxuchao Yes. We face this issue in the production cluster. And our customer also reported this issue

evalsocket:
@caesarxuchao Any update on issue?

PR #92069, opened 2020-06-12
Update github.com/evanphx/json-patch to e83c0a1c26c884f8ec4fb743f6db9…

tapih:
Signed-off-by: Hiroshi Muraoka <h.muraoka714@gmail.com>

**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
This PR fixes a bug that JSON patch with `kubectl apply` has exponential time complexity for the nesting depth of manifest.

**Which issue(s) this PR fixes**:
Fixes #91615 .

**Special notes for your reviewer**:
The existing PR https://github.com/kubernetes/kubernetes/pull/91622 breaks the compatibility to conform to RFC6902 in [this patch](https://github.com/evanphx/json-patch/pull/85).
This PR takes an intermediate commit to avoid the compatibility break and fixes just #91615.

**Does this PR introduce a user-facing change?**:
```release-note
Fixed a performance issue applying json patches to deeply nested objects
```

PR #92482, opened 2020-06-24
WIP: test https://github.com/evanphx/json-patch/pull/108

liggitt:


blaggacao:
Si there a chance this patch can be temporarily applied to unblock backports of https://github.com/kubernetes/kubernetes/pull/92069?

liggitt:
> Is there a chance this patch can be temporarily applied to unblock backports of #92069?

not without forking the upstream repo (note the pull-kubernetes-dependencies CI failure). we don't diverge from authoritative sources/shas for our dependencies

PR #92485, opened 2020-06-24
Avoid DeprecatedMightBeMasterNode() in e2e metrics

oomichi:
**What type of PR is this?**

/kind cleanup

As its name, DeprecatedMightBeMasterNode is deprecated.
In e2e metrics, the function was used for knowing master node name to get metrics from kube-scheduler and kube-controller-manager pods.
This make e2e metrics get these metrics directly by getting those pod names without calling DeprecatedMightBeMasterNode().

```release-note
NONE
```

### DIFF
--- a/test/e2e/framework/metrics/metrics_grabber.go
+++ b/test/e2e/framework/metrics/metrics_grabber.go
@@ -19,6 +19,7 @@ package metrics
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"sync"
 	"time"
 
@@ -27,7 +28,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
-	"k8s.io/kubernetes/test/e2e/system"
 
 	"k8s.io/klog/v2"
 )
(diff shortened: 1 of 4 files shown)


test/e2e/framework/metrics/metrics_grabber.go
```diff
@@ -61,38 +61,50 @@ type Grabber struct {
...
+		klog.Warningf("Can't find kube-scheduler pod. Grabbing metrics from kube-scheduler is disabled.")
+	}
+	if kubeControllerManager == "" {
 		controllers = false
-		clusterAutoscaler = ec != nil
-		if clusterAutoscaler {
-			klog.Warningf("Master node is not registered. Grabbing metrics from Scheduler, ControllerManager is disabled.")
-		} else {
-			klog.Warningf("Master node is not registered. Grabbing metrics from Scheduler, ControllerManager and ClusterAutoscaler is disabled.")
-		}
```
neolit123, line 95: should we still include the following outside of the `if kubeControllerManager == "" {` branch:

```
if ec == nil {
	klog.Warningf("Did not receive an external client interface. Grabbing metrics from ClusterAutoscaler is disabled.")
}
```

oomichi: @neolit123 Thanks for pointing it out.
That makes sense. I will add the above as you said.

test/e2e/framework/metrics/metrics_grabber.go
```diff
@@ -103,7 +115,8 @@ func NewMetricsGrabber(c clientset.Interface, ec clientset.Interface, kubelets b
 		grabFromKubelets:          kubelets,
 		grabFromScheduler:         scheduler,
 		grabFromClusterAutoscaler: clusterAutoscaler,
-		masterName:                masterName,
+		kubeScheduler:             kubeScheduler,
+		kubeControllerManager:     kubeControllerManager,
 		registeredMaster:          registeredMaster,
```
neolit123, line 120: maybe we should remove the registeredMaster variable and change HasRegisteredMaster too?

```
// HasControllerManagerAndScheduler returns if metrics grabber was able to find the kube-controller-manager and kube-scheduler
func (g *Grabber) HasControllerManagerAndScheduler () bool {
	return kubeScheduler != "" && kubeControllerManager != ""
}
```

the usage is here:
https://github.com/kubernetes/kubernetes/search?q=HasRegisteredMaster&unscoped_q=HasRegisteredMaster

oomichi: Thanks again.
I tend to prefer HasControlPlanePods().
I will update this PR.

test/e2e/framework/metrics/metrics_grabber.go
```diff
@@ -103,14 +113,14 @@ func NewMetricsGrabber(c clientset.Interface, ec clientset.Interface, kubelets b
...
-		registeredMaster:          registeredMaster,
+		kubeScheduler:             kubeScheduler,
+		kubeControllerManager:     kubeControllerManager,
 	}, nil
 }
 
-// HasRegisteredMaster returns if metrics grabber was able to find a master node
-func (g *Grabber) HasRegisteredMaster() bool {
-	return g.registeredMaster
+// HasControlPlanePods returns if metrics grabber was able to find a master node
```
neolit123, line 121: returns true if the metrics grabber was able to find control-plane pods

test/e2e/framework/metrics/metrics_grabber.go
```diff
@@ -151,7 +161,7 @@ func (g *Grabber) GrabFromScheduler() (SchedulerMetrics, error) {
 
 // GrabFromClusterAutoscaler returns metrics from cluster autoscaler
 func (g *Grabber) GrabFromClusterAutoscaler() (ClusterAutoscalerMetrics, error) {
-	if !g.registeredMaster && g.externalClient == nil {
+	if !g.HasControlPlanePods() && g.externalClient == nil {
 		return ClusterAutoscalerMetrics{}, fmt.Errorf("Master's Kubelet is not registered. Skipping ClusterAutoscaler's metrics gathering")
```
neolit123, line 165: i guess we need to replace:
> "Master's Kubelet is not registered. Skipping ClusterAutoscaler's metrics gathering"

with something like:
> Did not find control-plane pods. Skipping ClusterAutoscaler's metrics gathering

"master's kubelet" was misleading originally too.

test/e2e/framework/metrics/metrics_grabber.go
```diff
@@ -61,38 +61,48 @@ type Grabber struct {
...
 func NewMetricsGrabber(c clientset.Interface, ec clientset.Interface, kubelets bool, scheduler bool, controllers bool, apiServer bool, clusterAutoscaler bool) (*Grabber, error) {
-	registeredMaster := false
-	masterName := ""
-	nodeList, err := c.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+
+	kubeScheduler := ""
+	kubeControllerManager := ""
+
+	regKubeScheduler := regexp.MustCompile("kube-scheduler-.*")
+	regKubeControllerManager := regexp.MustCompile("kube-controller-manager-.*")
```
neolit123, line 76: i remembered a discussion around the fact that this is not a reliable way to determine if the namespace has the components in question, because someone might have decided to name them differently in their custom deployment.

but i'd argue that it's much better compared to using the deprecated "master" function and we are not going to return errors for them, but just skip the metrics gathering.

test/e2e/framework/metrics/metrics_grabber.go
oomichi: Thanks, done

test/e2e/framework/metrics/metrics_grabber.go
oomichi: Thanks again, `Did not find control-plane pods. ` will be applied.

neolit123 (commented):
/lgtm
thanks @oomichi

neolit123:
there is one minor complain by govet:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92485/pull-kubernetes-verify/1276264296552075264
```
# k8s.io/kubernetes/test/e2e/framework/metrics
test/e2e/framework/metrics/metrics_grabber.go:83:3: Warning call has possible formatting directive %s
```

oomichi:
oops, sorry about that.

/hold

PR #92489, opened 2020-06-24
Add SIG storage owner aliases

alculquicondor:
/sig scheduling
/sig storage

**What type of PR is this?**

/kind documentation

Update aliases related to SIG storage:

- Remove sig-storage-api-approvers which was not in use
- Merge owners from `pkg/controller/volume` and `pkg/volume` into sig-storage-{approvers,reviewers}
- Add @deads2k to sig-storage-api-reviewers, which are the new reviewers for `staging/src/k8s.io/api/storage` and `pkg/apis/storage` @mbohlool is not added due to inactivity.

Note that this changes give extra ownership to some individuals, given that `pkg/controller/volume` had a subset of `pkg/volume` owners.

And give ownership to `pkg/scheduler/framework/plugins/volumebinding` to sig/storage

```release-note
NONE
```

/cc @msau42 @saad-ali @liggitt @thockin

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -142,8 +142,25 @@ aliases:
     - justaugustus
     - listx
 
+  sig-storage-approvers:
+    - saad-ali
+    - thockin
+    - jsafrane
+    - gnufied
+    - msau42
+  # emeritus:
+  # - rootfs
   sig-storage-reviewers:
     - saad-ali
+    - jsafrane
+    - rootfs
+    - jingxu97
+    - msau42
+    - gnufied
+    - verult
+    - humblec
+  # emeritus:
+  # - davidz627
 
   sig-scheduling-maintainers:
     - alculquicondor
@@ -454,6 +471,7 @@ aliases:
   #   -
 
   sig-storage-api-reviewers:
+    - deads2k
     - saad-ali
     - msau42
     - jsafrane
--- a/pkg/apis/storage/OWNERS
+++ b/pkg/apis/storage/OWNERS
@@ -1,5 +1,8 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
+approvers:
+- sig-storage-api-approvers
 reviewers:
-- deads2k
-- mbohlool
+- sig-storage-api-reviewers
+labels:
+- sig/storage
--- a/pkg/controller/volume/OWNERS
+++ b/pkg/controller/volume/OWNERS
@@ -1,15 +1,6 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
 approvers:
-- jsafrane
-- saad-ali
-- msau42
+- sig-storage-approvers
 reviewers:
-- saad-ali
-- jsafrane
-- jingxu97
-- rootfs
-- gnufied
-- msau42
-- verult
-- davidz627
+- sig-storage-reviewers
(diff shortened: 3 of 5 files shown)


alculquicondor:
And almost forgot @ahg-g

OWNERS_ALIASES
```diff
@@ -361,10 +376,6 @@ aliases:
     - smarterclayton
     - thockin
 
-  sig-storage-api-approvers:
-    - liggitt
-    - thockin
```
liggitt, line 366: please restore this, we use it for targeting subsets of approvers with experience in different areas

OWNERS_ALIASES
```diff
@@ -454,6 +465,7 @@ aliases:
   #   -
 
   sig-storage-api-reviewers:
+    - deads2k
```
liggitt, line 468: can check with @deads2k but I bet he would be fine not being added here

liggitt (commented):
a couple nits, lgtm otherwise

alculquicondor:
I forgot to mention that I looked at `pkg/api/{persistentvolume,persistentvolumeclaim}`, but since the set of people is different, I wasn't sure where to put them.

OWNERS_ALIASES
alculquicondor: Should I also use the alias in the APIs folders for approvers?

OWNERS_ALIASES
```diff
@@ -142,8 +142,23 @@ aliases:
...
+    - msau42
   sig-storage-reviewers:
     - saad-ali
+    - jsafrane
+    - rootfs
+    - jingxu97
+    - msau42
+    - gnufied
+    - verult
+    - davidz627
```
msau42, line 160: Also move to emeritus

ahg-g:
> And give ownership to pkg/scheduler/framework/plugins/volumebinding to sig/storage

what motivated this?

alculquicondor:
> what motivated this?

Visibility. It would add the label, reviewers. The most controversial bit could be that sig-storage-approvers could approve changes that only affect this plugin without sig-scheduling input. If so, we could just not add sig-storage-approvers

ahg-g:
> at only affect this plugin without sig-sched



> > what motivated this?
> 
> Visibility. It would add the label, reviewers. The most controversial bit could be that sig-storage-approvers could approve changes that only affect this plugin without sig-scheduling input. If that's problematic, we could just not add sig-storage-approvers

yes, my comment is about "giving ownership" which equates being approvers. I am ok with adding sig-storage as reviewers, but I think approving changes to the scheduler should continue to be done by sig scheduling maintainers.

OWNERS_ALIASES
```diff
@@ -142,8 +142,23 @@ aliases:
     - justaugustus
     - listx
 
+  sig-storage-approvers:
+    - saad-ali
+    - thockin
+    - jsafrane
+    - rootfs
```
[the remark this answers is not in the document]
alculquicondor: Done. Let's give them some time to ack.

OWNERS_ALIASES
alculquicondor: Done

OWNERS_ALIASES
alculquicondor: I used the alias in pkg/apis/storage and staging/src/k8s.io/api/storage

Let me know if that's undesired.

alculquicondor (commented):
Will squash once we have reached consensus

alculquicondor:
We didn't hear back from tagged people. Anything else from approvers?

alculquicondor:
/hold
let me squash

alculquicondor:
/hold cancel
squashed

msau42:
/lgtm
/retest
Thanks for cleaning this up!

PR #92491, opened 2020-06-24
compile out gofuzz from prod binaries

### LINKED ISSUE #14460: Fuzz() method on IntOrString causes gofuzz linked into prod code
Sigh.  Not a huge deal but need a better answer.

BenTheElder:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes https://github.com/kubernetes/kubernetes/issues/14460

```release-note
A new `nofuzz` go build tag now disables gofuzz support. Release binaries enable this.
```

```docs

```


BenTheElder:
Here's a quick prototype of eliminating the fuzzer from our binaries, closing the oldest open SIG testing bug in this repo.
/cc @dims @liggitt @thockin 

This sort of thing is also unfortunately not available under bazel, we probably care most about this in releases anyhow.
I have not yet audited where else we need to do this, I plan to do that once the approach is cleared.

We do use a similar approach already for the providerless build KEP and dockershim-less building.
However, those tags are not enabled by default. This one is.

We could alternatively have a positive tag, and require our tests enable it, but that's a breaking change for external consumers of types like this. This way the change only affects our binaries.

staging/src/k8s.io/apimachinery/pkg/util/intstr/intstr.go
```diff
@@ -129,21 +128,6 @@ func (IntOrString) OpenAPISchemaType() []string { return []string{"string"} }
 // the OpenAPI spec of this type.
 func (IntOrString) OpenAPISchemaFormat() string { return "int-or-string" }
 
-func (intstr *IntOrString) Fuzz(c fuzz.Continue) {
```
liggitt, line 132: metav1.{MicroTime,Time} both implement this as well... same thing needed there?

liggitt:
if we want this to stick we need a check that the built binaries don't include gofuzz symbols

liggitt:
I'm not very familiar with why gofuzz is problematic to have built into binaries... does it actually cause problems or is this just a representative "should be test-only" dependency? if the latter, do we want fuzz-specific build tags or do we really want a "notest" build tag?

BenTheElder:
As far as I can tell just because it should be test, it's harder than other
test code because for a type to define fuzzing for itself it has to
implement an interface method involving a concrete gofuzz type parameter.

I'm not sure that we have something else like this with other packages.

I'm also not sure how important it is that this stick. I beleive this is
just about avoiding the smell / binary bloat of linking in the fuzzer
runtime used only at (unit) test time in prod.

staging/src/k8s.io/apimachinery/pkg/util/intstr/intstr.go
BenTheElder: yeah, I held on that for feedback on the tag.

staging/src/k8s.io/apimachinery/pkg/util/intstr/instr_fuzz.go
```diff
@@ -0,0 +1,44 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// setting nofuzz in go build -tags will disable linking in the fuzzer code
+// we do this by default at runtime
+// +build !nofuzz
```
thockin, line 19: I'm used to seeing this RIGHT at the top of the file - I would not think to look for it anywhere else.

thockin (commented):
The reason the bug was filed was definitely "we shouldn't compile test code into prod" binaries.  How important is that?  I don't know.  There are other places where the `FakeFoo` is in the same file or same packages as the `Foo`, too.

I don't hate the pattern.  I'm flummoxed why Go doesn't have this handled by default.  The natural approach would be to decorate these as `+build testonly` but that requires anyone who runs `go test` to get it right.  Given that releases are well-controlled, that seems like the right path, so i concur with your approach.

I like the idea of generalizing this to something like not-for-prod, so maybe `+build !for_prod` or something?

staging/src/k8s.io/apimachinery/pkg/util/intstr/instr_fuzz.go
BenTheElder: ack.

BenTheElder:
> I like the idea of generalizing this to something like not-for-prod, so maybe +build !for_prod or something

I don't have super strong opinions on this, but I was thinking the only reason anyone else should toggle this is if they're consuming this code from another repo and either want access to the fuzzing behavior, or what to avoid the dependency.

FWIW `nofuzz` as a name is inspired by gofuzz's `//+build gofuzz` sample code.

> I don't hate the pattern. I'm flummoxed why Go doesn't have this handled by default. The natural approach would be to decorate these as +build testonly but that requires anyone who runs go test to get it right. Given that releases are well-controlled, that seems like the right path, so i concur with your approach.

I should have mentioned, I actually looked into this but it's been brought up in go and shot down, repeatedly. https://github.com/golang/go/issues/21360 https://github.com/golang/go/issues/14668

staging/src/k8s.io/apimachinery/pkg/util/intstr/intstr.go
BenTheElder: done.

liggitt:
this looks reasonable to me

BenTheElder:
/priority backlog
will circle back to this in the 1.20 timeframe

thockin (commented):
I still don't know why Go doesn't have a test build-tag built in.
/lgtm
/approve

Issue #91782, opened 2020-06-04
Scheduler: remove dependency on pkg/apis/core/v1/helper


ingvagabund:
As part of reducing dependency on the main k8s.io/kubernetes project (see https://github.com/kubernetes/kubernetes/issues/89930) and allowing the scheduling framework to be eventually moved under staging repo.

The goal here is the remove dependency on `k8s.io/kubernetes/pkg/apis/core/v1/helper` package. The following functions from `k8s.io/kubernetes/pkg/apis/core/v1/helper` are still imported from within `pkg/scheduler`:
- **GetAvoidPodsFromNodeAnnotations**: `v1.PreferAvoidPodsAnnotationKey` string used in the function is defined under `k8s.io/kubernetes/staging/src/k8s.io/api/core/v1`. `k8s.io/kubernetes/staging/src/k8s.io/api/core/v1` has a bunch of files with methods defined over the core types. Question is what are the condition for adding new functions/methods to the package.
- **GetPersistentVolumeClaimClass**: `v1.BetaStorageClassAnnotation` string used in the function is defined under `k8s.io/kubernetes/staging/src/k8s.io/api/core/v1`.
- **HugePageResourceName**: `v1.ResourceHugePagesPrefix` string used in the function is defined under `k8s.io/kubernetes/staging/src/k8s.io/api/core/v1`.
- **IsAttachableVolumeResourceName**: `v1.ResourceAttachableVolumesPrefix`
- **IsExtendedResourceName**: `v1.DefaultResourceRequestsPrefix`
- **IsHugePageResourceName**: `v1.ResourceHugePagesPrefix` string used in the function is defined under `k8s.io/kubernetes/staging/src/k8s.io/api/core/v1`.
- **IsScalarResourceName**: depends on `IsHugePageResourceName`, `IsExtendedResourceName`, `IsPrefixedNativeResource` (`v1.ResourceDefaultNamespacePrefix`) and `IsAttachableVolumeResourceName` 
- **MatchNodeSelectorTerms**: depends on `k8s.io/kubernetes/staging/src/k8s.io/api/core/v1` bits.
- **NodeSelectorRequirementsAsSelector**: depends on `k8s.io/kubernetes/staging/src/k8s.io/api/core/v1` bits.
- **TolerationsTolerateTaint**: depends on `v1.Toleration` and `v1.Taint`
- **FindMatchingUntoleratedTaint**: depends on `v1.Toleration` and `v1.Taint`

[...]

Issue #89930, opened 2020-04-07
Clean up scheduler dependencies on k8s.io/kubernetes packages


damemi:
To make it easier for out-of-tree developers to import the scheduler framework, among other scheduling helpers and constants, we should explore areas that we can remove dependencies on `k8s.io/kubernetes`. Places that could be easily removed are, for example, using internal APIs where not required or importing unrelated packages for simple helpers which can be trivially implemented. This will clean up the dependency tree for importing scheduler code externally. 

This is related to: [Proposal to migrate scheduler code to staging](https://docs.google.com/document/d/1WO-ixERpqkCSEXEq30YtEH_z_G-BoLKeCbkRJcKq3xA/edit). While the proposal includes all scheduler code, we are only focusing on scheduler framework code. However identifying anywhere in the scheduler code that these dependencies can be cleaned up will be helpful as well.

This will probably be an ongoing cleanup effort, and breaking the work into smaller PRs may actually help with review and preventing code breaking in the process.
/kind cleanup
/sig scheduling

PRs opened so far:
* https://github.com/kubernetes/kubernetes/pull/90105 `pkg/apis/core` @xiaoanyunfei 
  * moving the default scheduler name and references to the external api
* ~~https://github.com/kubernetes/kubernetes/pull/90006~~ `pkg/api/core/v1/helper` @gavinfish (part 1)
  * copies `GetPersistentVolumeClaimClass`, which is used in several other areas around pkg/volume
  * replaced by https://github.com/kubernetes/kubernetes/pull/98433
* https://github.com/kubernetes/kubernetes/issues/91782 Overview issue of `pkg/api/core/v1/helper` @ingvagabund 
* ~~https://github.com/kubernetes/kubernetes/pull/90041 `pkg/api/testing` @tanjunchen~~
* https://github.com/kubernetes/kubernetes/pull/90039 ~~`pkg/scheduler/util`~~ `pkg/api/v1/pod` @yuzhiquan 🔹
  *  move `GetPodPriority` to scheduler utils
  * ~~https://github.com/kubernetes/kubernetes/pull/90703 `pkg/api/v1/pod` @yiyang5055 🔹~~
* https://github.com/kubernetes/kubernetes/pull/90046 `pkg/controller/volume/scheduling/metrics` @tanjunchen 
  * copies metrics registration and definitions around volume binding
* https://github.com/kubernetes/kubernetes/pull/90111 `pkg/features` @gavinfish 🔹
[...]

PR #79330, opened 2019-06-24
publishing: use deps in replace directive in rules

nikhita:
The list of deps in rules for a repo right now is equal to the list of deps
mentioned in the require directive in the repo's `go.mod` file.
This means that we loose all transitive deps mentioned in the replace
directive in `go.mod` files.

So instead use all deps mentioned in the replace directive, and prune
any extra dependencies.

Also add a test for this in `verify-publishing-bot.py`.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

/assign @sttts @dims @liggitt

PR #74879, opened 2019-03-04
remove core repo dependency for `auth reconcile` command

juanvallejo:
/kind cleanup

Addresses part of https://github.com/kubernetes/kubectl/issues/80

```release-note
NONE
```

/sig cli

cc @soltysh @seans3

PR #92507, opened 2020-06-25
Initialize k8s.io/component-helpers staging repository

ingvagabund:
**What type of PR is this?**
/kind feature

Create `k8s.io/component-helpers` staging repository to provide helpers primarily for core components which are still part of `k8s.io/kubernetes`. Yet, to be imported by components no longer located under `k8s.io/kubernetes` or to be moved out of it to avoid code duplication.

Readme of the new staging repository does not contain a complete list of what is and what is not allowed. Although, it's meant primarily for helpers (and possible other code) used by (core) components that is currently (or to be) duplicated.

Opening the PR to start discussion on the topic and attract more interest from the community.

Primary reason for this is to allow the scheduling framework to become free of all `k8s.io/kubernetes` imports (avoiding code duplication) so the framework can be eventually moved under its own repository (presumably `k8s.io/kube-scheduler`) and imported by other projects building on top of the framework. E.g. https://github.com/volcano-sh/volcano, https://github.com/kubernetes-sigs/descheduler, https://github.com/kubernetes-sigs/cluster-capacity.

https://github.com/kubernetes/kubernetes/issues/91782

- root OWNERS file was copy-pasted from k8s.io/apimachinery (there's a bunch of other staging repos which do the same)
- sig-scheduling/OWNERS file was copy-pasted from k8s.io/kube-scheduler since SIG scheduling will be the primary consumer of the helpers
- currently, only a single helper functions is duplicated to discuss what's the right approach here (once agreed on the "right" approach, other helper functions will be moved/duplicated)
- targeted for k8s 1.20

```release-note
Added new k8s.io/component-helpers repository providing shared helper code for (core) components.
```

```docs
TBD
```

**TODO**:
- [x] add validation test complaining about forbidden `k8s.io/` dependencies (other than `k8s.io/apimachinery`, `k8s.io/api`, `k8s.io/client-go`)


ingvagabund:
@liggitt @deads2k @ahg-g

dims:
@ingvagabund i am not sure i like this pattern where we inject the name of the sig AND a "general" library. Will you be able to join the sig-arch / code organization call today?

/area code-organization

ingvagabund:
Given, the helpers are currently moved/duplicated under `sig-scheduling` directory, it's really impossible to talk about removing `pkg/apis/core/v1/helpers`. It will be very strange for non-scheduling components to import `k8s.io/library-go/sig-scheduling` package. Though, eventually, we might find a better location for the helpers to allow non-scheduling related components to import the helpers as well.

ingvagabund:
> @ingvagabund i am not sure i like this pattern where we inject the name of the sig AND a "general" library.

That's still to be discussed. I was following suggestion given by @deads2k and @liggitt.

staging/src/k8s.io/library-go/README.md
```diff
@@ -0,0 +1,23 @@
+# library-go
+
+This repository provides helpers primarily for core components (core components as described in [Create a k8s.io/component-base repo](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cluster-lifecycle/wgs/0032-create-a-k8s-io-component-repo.md#component-definition)) which are required by at least two components living outside
+of the main `k8s.io/kubernetes` package. Yet, still with a high level of abstraction.
```
deads2k, line 4: by at least two separate binaries in k/k.

staging/src/k8s.io/library-go/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+The only allowed kubernetes dependencies are `k8s.io/apimachinery`, `k8s.io/api` and `k8s.io/client-go`.
+
+## Purpose
+
+One of the goals is to provide a better location for helpers currently located under `k8s.io/kubernetes/pkg/apis`. Recent effort of moving [scheduling
+ framework](https://kubernetes.io/docs/concepts/scheduling-eviction/scheduling-framework/) under
+`k8s.io/kube-scheduler` requires duplication of many helper functions
+(see [#91782](https://github.com/kubernetes/kubernetes/issues/91782) for more details).
```
deads2k, line 13: @soltysh provide some verbiage for k/kubectl rbac libraries

staging/src/k8s.io/library-go/SECURITY_CONTACTS
```diff
@@ -0,0 +1,17 @@
...
+# to for triaging and handling of incoming issues.
+#
+# The below names agree to abide by the
+# [Embargo Policy](https://git.k8s.io/security/private-distributors-list.md#embargo-policy)
+# and will be removed and replaced if they violate that agreement.
+#
+# DO NOT REPORT SECURITY VULNERABILITIES DIRECTLY TO THESE NAMES, FOLLOW THE
+# INSTRUCTIONS AT https://kubernetes.io/security/
+
+cjcullen
```
deads2k, line 13: this is a stock list.  I'm willing to be in the list and @liggitt is a glutton for punishment.

staging/src/k8s.io/library-go/sig-scheduling/corev1helpers/README.md
```diff
@@ -0,0 +1,8 @@
+# sig-scheduling/corev1helpers
+
+All helper methods in this package are duplicated from `k8s.io/kubernetes/pkg/apis/core/v1/helper`.
```
deads2k, line 3: this won't age well.  Snip this line.

damemi (commented):
I get that this is just meant to initialize the repo, but since it's moving a function here should it have a 2nd commit removing it from its previous location and updating usages to point here?

staging/src/k8s.io/library-go/SECURITY_CONTACTS
enj: This should be kept in sync with [committee-product-security](https://github.com/nikhita/community/blob/284c616e100a19c184b79267dd838f69815b02c4/OWNERS_ALIASES#L152-L158).

staging/src/k8s.io/library-go/README.md
```diff
@@ -0,0 +1,23 @@
+# library-go
+
+This repository provides helpers primarily for core components (core components as described in [Create a k8s.io/component-base repo](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cluster-lifecycle/wgs/0032-create-a-k8s-io-component-repo.md#component-definition)) which are required by at least two components living outside
+of the main `k8s.io/kubernetes` package. Yet, still with a high level of abstraction.
+
+The only allowed kubernetes dependencies are `k8s.io/apimachinery`, `k8s.io/api` and `k8s.io/client-go`.
```
enj, line 6: Is this being enforced?

enj (commented):
Minor comments.  Really happy to see this moving forward.

staging/src/k8s.io/library-go/README.md
enj: xref:

https://github.com/kubernetes/kubernetes/pull/74879
https://github.com/liggitt/kubernetes/commits/extract-rbac

staging/src/k8s.io/library-go/README.md
enj: (nvm I see it is a TODO at the top)

staging/src/k8s.io/library-go/sig-scheduling/OWNERS
```diff
@@ -0,0 +1,11 @@
+# See the OWNERS docs at https://go.k8s.io/owners
```
enj, line 1: I thought we wanted an import path more like `k8s.io/library-go/scheduling/...` (drop the `sig-` bit)

dims:
@ingvagabund Jan, please see the doc for code-org project (and the meeting video if you have time), we came up with some thoughts/ideas around this issue: https://docs.google.com/document/d/1HtTI0rJEGP_MSf6eO87aCmx_tzpovPAAg7U2Zxwm8FE/edit?ts=5c9d0a4c#heading=h.y3e75ebp5by9

ingvagabund:
> @ingvagabund Jan, please see the doc for code-org project (and the meeting video if you have time),

I see only one video in https://www.youtube.com/playlist?list=PL69nYSiGNLP03VEluzh0wpSRPzgve8kI5. Is there another place where the recordings are shared?

> we came up with some thoughts/ideas around this issue:

Awesome!!! Thank you for that. I will check those out.

soltysh:
> > @ingvagabund Jan, please see the doc for code-org project (and the meeting video if you have time),
> 
> I see only one video in https://www.youtube.com/playlist?list=PL69nYSiGNLP03VEluzh0wpSRPzgve8kI5. Is there another place where the recordings are shared?

There's sig-arch playlist: https://www.youtube.com/playlist?list=PL69nYSiGNLP2m6198LaLN6YahX7EEac5g but the last meeting there is from last week's sig-arch, so I'd love to watch that call too when it gets published.

dims:
@soltysh @ingvagabund i think zoom is slow if it does not get uploaded automatically by monday, i'll dig into why.

staging/src/k8s.io/library-go/README.md
soltysh: from sig-cli pov:

Another example is shared RBAC code which is blocking extracting kubectl to staging (see https://github.com/kubernetes/enhancements/issues/1020). This problem dates all the way back to December 2018 (see SIG-CLI call from December 19, 2018: https://docs.google.com/document/d/1r0YElcXt6G5mOWxwZiXgGu_X6he3F--wKwg-9UBc29I/edit?pli=1). Recently the topic was touched during sig-auth call (see https://docs.google.com/document/d/1woLGRoONE3EBVx-wTb4pvp4CI7tmLZ6lS26VTbosLKM/edit?ts=5ef3be6a#heading=h.etc9yylhln8x).

soltysh:
I support the idea brought in sig-arch code organization about naming strategy, I guess we can still keep clear link through using the sig name just w/o the word itself. Since for now, this will perfectly fit the two use cases we have.

staging/src/k8s.io/library-go/README.md
ingvagabund: Updated `staging/publishing/import-restrictions.yaml` which should do this automatically.

staging/src/k8s.io/library-go/SECURITY_CONTACTS
ingvagabund: Synced

staging/src/k8s.io/library-go/scheduling/OWNERS
```diff
@@ -0,0 +1,11 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+- sig-scheduling-maintainers
+- sttts
+- luxas
+reviewers:
+- sig-scheduling
+- dixudx
```
alculquicondor, line 9: where is this list of extra people coming from?

staging/src/k8s.io/library-go/scheduling/corev1helpers/README.md
```diff
@@ -0,0 +1,6 @@
+# sig-scheduling/corev1helpers
+
+Primarily to be imported by the scheduling framework to allow
```
alculquicondor, line 3: This should be expressed as a criteria, rather than motivation.

Also, can it just be a package godoc?

staging/src/k8s.io/library-go/scheduling/OWNERS
ingvagabund: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kube-scheduler/OWNERS

staging/src/k8s.io/library-go/scheduling/corev1helpers/README.md
ingvagabund: > This should be expressed as a criteria

Descheduler/Cluster capacity are not part of the framework. Yet, they are another suitable candidates for importing the helpers. Although, this is a step toward to have the framework be importable as a go module, it's still important to allow other components to import the code. Otherwise, we might just as well duplicate the code and put it under the framework.

> Also, can it just be a package godoc?

How many people read godoc vs. readme? I'd like the text to be visible so people are aware. Although, how many people read anything but code comments.

ingvagabund:
> @soltysh @ingvagabund i think zoom is slow if it does not get uploaded automatically by monday, i'll dig into why.

@dims still not published

staging/src/k8s.io/library-go/scheduling/corev1helpers/BUILD
```diff
@@ -0,0 +1,34 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
```
knight42, line 1: Is it possible to rename `library-go/scheduling/corev1helpers` to simply `library-go/scheduling/helpers`? 
IMHO, no matter what the group version is, they are all helpers.

dims:
@ingvagabund https://youtu.be/yngBgHeLOW8

ingvagabund:
> @ingvagabund https://youtu.be/yngBgHeLOW8

@dims nice!!! thanks for bringing the topic

knight42:
I just noticed that istio has a similar repo named `pkg`: https://github.com/istio/pkg, which contains "Common utility packages leveraged by other repos". Is it a good idea that we also have such a repo?

staging/src/k8s.io/library-go/scheduling/corev1helpers/BUILD
ingvagabund: Not against that. Right now, the idea is to put here all the helpers that are needed by the scheduling framework and at least other separate component/binary. Though, the version might still help to minimize indirect dependencies when importing `k8s.io/library-go/scheduling/NAME` package. On the other hand we might omit `helpers` in the package name and have `k8s.io/library-go/scheduling/core/v1` instead? Assuming the staging repo gets renamed from `library-go` to e.g. `helpers`.

ingvagabund:
> I just noticed that istio has a similar repo named pkg: https://github.com/istio/pkg, which contains "Common utility packages leveraged by other repos". Is it a good idea that we also have such a repo?

Using `pkg` for a repo name is not a good practice. It's so generic it does not say anything about what's inside. There's https://github.com/kubernetes/utils repo which does it though without any k8s dependency.

ingvagabund:
Given the `library-go` name is not unanimously accepted, I am suggesting alternative names (feel free to suggest more):

| Name |  Notes |
|----------|:-------------:|
| library-go | could be mistaken with client-go |
| api-helpers | helpers are not really for creating helper functions for api types, it's about creating helpers on top of api types |
| base-helpers | base might be too focused, not every helper is supposed to be basic |

staging/src/k8s.io/library-go/scheduling/corev1helpers/README.md
alculquicondor: > This should be expressed as a criteria, rather than motivation.

So I would say something like:

```
Functions in this package should satisfy one of the following:

- Be used by more than one core component (kube-scheduler, kubelet, kube-apiserver, etc.)
- Be used by a core component and another kubernetes project (cluster-autoscaler, descheduler)

And be a scheduling feature.

See initial motivation: ....
```

> How many people read godoc vs. readme? I'd like the text to be visible so people are aware. Although, how many people read anything but code comments.

A code comment is more visible when you are reviewing code changes. After all, the audience of this is developers and reviewers.

alculquicondor:
I vote for component-helpers

liggitt:
api-helpers or component-helpers seems fine to me

staging/src/k8s.io/library-go/scheduling/OWNERS
ingvagabund: Done

staging/src/k8s.io/library-go/scheduling/corev1helpers/README.md
ingvagabund: Moved under `doc.go` with your suggested comment.

ingvagabund:
Staging repo renamed to `k8s.io/component-helpers`.

staging/src/k8s.io/library-go/scheduling/corev1helpers/BUILD
ingvagabund: Going with `k8s.io/component-helpers/scheduling/corev1`. Though, the topic is still open for discussion.

knight42: I am OK with this naming, it looks much better.

staging/publishing/rules.yaml
```diff
@@ -1646,3 +1646,18 @@ rules:
...
+  - source:
+      branch: master
+      dir: staging/src/k8s.io/component-helpers
+    name: master
+    dependencies:
+    - repository: apimachinery
+      branch: master
+    - repository: api
+      branch: master
+    - repository: client-go
```
nikhita, line 1662: client-go would need to be removed from here because component-helper doesn't import it (and client-go isn't in it's go.mod)

ingvagabund: thanks!!!

sttts:
Is this supposed to be consumed by a library (X) outside of k/k, and X is then revendored into kube? This would create a vendoring cycle:

1. e.g. client-go changes in an incompatible way
2. we have to adapt component-helpers.
3. X cannot be changed because it is vendored.
4. X cannot be revendored in a compatible way because (1) and (2) are not finished yet due to (3).

cc @liggitt @dims

ingvagabund:
The plan is to move all the scheduling framework code from k8s.io/kubernetes/pkg/scheduling/framework under k8s.io/kubernetes/staging/src/k8s.io/kube-scheduler and import k8s.io/kube-scheduler in repos outside of k/k. No vendoring back of repos depending in `k8s.io/kube-scheduler` or `k8s.io/component-helpers`.

sttts:
> The plan is to move all the scheduling framework code from k8s.io/kubernetes/pkg/scheduling/framework under k8s.io/kubernetes/staging/src/k8s.io/kube-scheduler and import k8s.io/kube-scheduler in repos outside of k/k.

Ok, this sounds better than what I understood earlier today :+1:

alculquicondor:
Freeze is still in progress, you just have to wait.

damemi:
Bump, now that 1.20 is open is there anything else that was needed for this? I think having it will assist the stability goals of 1.20

staging/src/k8s.io/component-helpers/scheduling/corev1/doc.go
```diff
@@ -0,0 +1,25 @@
...
+*/
+
+// Package corev1 defines functions which should satisfy one of the following:
+//
+// - Be used by more than one core component (kube-scheduler, kubelet, kube-apiserver, etc.)
+// - Be used by a core component and another kubernetes project (cluster-autoscaler, descheduler)
+//
+// And be a scheduling feature.
+//
+// See initial motivation.
```
alculquicondor, line 24: This line seems unnecessary

staging/src/k8s.io/component-helpers/scheduling/corev1/helpers_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+	for i, tc := range testCases {
+		v, err := GetAvoidPodsFromNodeAnnotations(tc.node.Annotations)
+		if err == nil && tc.expectErr {
+			t.Errorf("[%v]expected error but got none.", i)
+		}
+		if err != nil && !tc.expectErr {
+			t.Errorf("[%v]did not expect error but got: %v", i, err)
+		}
+		if !reflect.DeepEqual(tc.expectValue, v) {
```
alculquicondor, line 117: nit: use cmp package.

staging/src/k8s.io/component-helpers/scheduling/corev1/helpers.go
```diff
@@ -0,0 +1,36 @@
...
+
+package corev1
+
+import (
+	"encoding/json"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+// GetAvoidPodsFromNodeAnnotations scans the list of annotations and
```
alculquicondor, line 25: I recall we had other functions that needed to be moved. Are they coming in a followup?

alculquicondor (commented):
/approve
from sig-scheduling

staging/src/k8s.io/component-helpers/scheduling/corev1/helpers.go
damemi: +1 yeah, many of the dependent helpers from https://github.com/kubernetes/kubernetes/issues/89930. I think the idea here was just not to overcomplicate the initial commit

alculquicondor: sgtm

damemi:
Bumping again, is there anything we need to do to (or we're waiting on) to merge this? cc @dims @liggitt @soltysh

dims:
@damemi merging this right now will break stuff (like publishing-bot) as i don't yet see `github.com/kubernetes/component-helpers` 

cc @nikhita (for guidance)

staging/src/k8s.io/component-helpers/go.mod
```diff
@@ -0,0 +1,16 @@
...
+
+require (
+	k8s.io/api v0.0.0
+	k8s.io/apimachinery v0.0.0
+)
+
+replace (
+	k8s.io/api => ../api
+	k8s.io/apimachinery => ../apimachinery
+	k8s.io/component-helpers => ../component-helpers
```
luxas, line 15: this seems unnecessary? just `.` or not at all? I see it's autogenerated but haven't seen it in other places

staging/src/k8s.io/component-helpers/README.md
```diff
@@ -0,0 +1,27 @@
+# component-helpers
+
+This repository provides helpers primarily for core components (core components as described in [Create a k8s.io/component-base repo](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cluster-lifecycle/wgs/0032-create-a-k8s-io-component-repo.md#component-definition)) which are required by at least two separate binaries in kubernetes org.
+Yet, still with a high level of abstraction.
+
+The only allowed kubernetes dependencies are `k8s.io/apimachinery`, `k8s.io/api` and `k8s.io/client-go`.
```
luxas, line 6: Could you please describe how this is different from component-base here?

pkg/apis/core/v1/helper/helpers.go
```diff
@@ -495,19 +494,6 @@ func GetMatchingTolerations(taints []v1.Taint, tolerations []v1.Toleration) (boo
 	return true, result
 }
 
-// GetAvoidPodsFromNodeAnnotations scans the list of annotations and
-// returns the pods that needs to be avoided for this node from scheduling
-func GetAvoidPodsFromNodeAnnotations(annotations map[string]string) (v1.AvoidPods, error) {
```
luxas, line 500: I guess you're planning to move the rest of the things after this initial PR is merged?

luxas (commented):
Hi @ingvagabund and thanks for this helping move the scheduler out of core :clap:!

Sorry if this has been discussed already, but some questions:
What is the difference between this repo and `component-base`? I think I understand the scope somewhat, but I'm not entierly sure on where to draw the line. Was there a KEP detailing this for `component-helpers`, in the same way as for `component-base` (which you linked in the README)?

Also it would be interesting to know why an other repo is made, compared to putting the helpers in `k8s.io/api/core/v1/helper`. Is it because we want to "group" the helper functions by SIG/usage target?

Regardless of place, what will the stability guarantees be now that this is "public" code?
It must never change during the lifetime of the API if stable (excluding major bugs)?

I'm all for moving these out of tree, so thank you for doing the effort :1st_place_medal:

ingvagabund:
@luxas thank you for taking your time reviewing the PR!!!

> What is the difference between this repo and component-base? I think I understand the scope somewhat, but I'm not entirely sure on where to draw the line. Was there a KEP detailing this for component-helpers, in the same way as for component-base (which you linked in the README)?

`component-base` staging repository was considered as a candidate. Thought, the proposal states the purpose of the repository is to "focus on factoring out the code needed for the core Kubernetes components". Our mine goal was to move relevant helpers to a repository that can be imported as a module. So the scheduling framework can be eventually move under `kube-scheduler` staging repo. Given helpers are not needed by the core components, `component-base` turned to be unsuitable.

> Also it would be interesting to know why an other repo is made, compared to putting the helpers in k8s.io/api/core/v1/helper. Is it because we want to "group" the helper functions by SIG/usage target?

API stability. Helpers are not good candidates for becoming part of API (e.g. due to code practices, immaturity of developers in writing APIs, ...).

> Regardless of place, what will the stability guarantees be now that this is "public" code?
It must never change during the lifetime of the API if stable (excluding major bugs)?

The same guarantees as there are for `component-base` or `utils`. The code can still change if there's a pressing reason.

staging/src/k8s.io/component-helpers/go.mod
ingvagabund: Yeah, it's autogenerated. Not much we can change here.

pkg/apis/core/v1/helper/helpers.go
ingvagabund: Yeah, this is an initial helper so the repository is not empty. The relevant rest will follow right after.

nikhita:
> @damemi merging this right now will break stuff (like publishing-bot) as i don't yet see github.com/kubernetes/component-helpers

> cc @nikhita (for guidance)

@ingvagabund @damemi yes, please create [an issue in the k/org repo](https://github.com/kubernetes/org/issues/new?assignees=&labels=area%2Fgithub-repo&template=repo-create.md&title=) and I can get the repo setup. :+1: 

Adding a hold to ensure this doesn't merge before the repo is created:
/hold

staging/src/k8s.io/component-helpers/OWNERS
```diff
@@ -0,0 +1,26 @@
...
+- caesarxuchao
+- cheftako
+- mikedanese
+- liggitt
+- gmarek
+- sttts
+- ncdc
+- logicalhan
+labels:
+- sig/api-machinery
```
neolit123, line 26: i'd recommend starting with a smaller list of owners - e.g. those who "championed" the repository creation idea.
instead of the full list from k/k, as it might feel a bit odd to get pinged by the bots for review on a new repository that you did not know existed?

staging/publishing/rules.yaml
```diff
@@ -1653,3 +1653,16 @@ rules:
...
+  branches:
+  - source:
+      branch: master
+      dir: staging/src/k8s.io/component-helpers
+    name: master
+    dependencies:
+    - repository: apimachinery
+      branch: master
+    - repository: api
+      branch: master
```
sttts, line 1668: client-go too?

nikhita: In the current form, component-helpers does not depend on client-go. Adding client-go here will make the publishing-bot-verify check fail.

We avoid adding extra deps here to avoid additional lines in the published go.mod file (it renders awkwardly) - https://github.com/kubernetes/kubernetes/pull/79330#discussion_r296615365

nikhita:
/hold cancel

Repo has been created - https://github.com/kubernetes/component-helpers

staging/src/k8s.io/component-helpers/OWNERS
ingvagabund: Good point.

@deads2k @soltysh @liggitt @dims would you be interested in being approvers and reviewers?

dims: +1 i can help.

liggitt: that seems ok, though I'd like to see clear top-level packages with sig-specific reviewers/approvers/labels (and drop the root sig/api-machinery label) so that very little actually falls through to the root module approvers

deads2k: > @deads2k @soltysh @liggitt @dims would you be interested in being approvers and reviewers?

Yes, I can help here.  I'd like to see the bar be reasonably high.

ingvagabund: > I'd like to see clear top-level packages with sig-specific reviewers/approvers/labels

Can you more elaborate on that? Are you saying `k8s.io/component-helpers/scheduling` naming is not sufficient or are you referring to a rule to follow for contributors?

liggitt: More of a rule to follow (e.g. adding the expected structure to the readme, where each top-level package has a clear sig-oriented reviewer/approver list, and no code lives in the top-level package)

soltysh: I agree with Jordan and David here. The libraries should be reviewed with great caution and less approvers is better here.

ingvagabund: The list of owners updated to:
```
approvers:
- dims
- deads2k
- liggitt
- soltysh
reviewers:
- dims
- deads2k
- liggitt
- soltysh
```

staging/src/k8s.io/component-helpers/README.md
ingvagabund: Done

ingvagabund:
@liggitt @dims @deads2k @soltysh ready for review. PTAL

staging/src/k8s.io/component-helpers/scheduling/corev1/helpers.go
```diff
@@ -0,0 +1,36 @@
...
+// returns the pods that needs to be avoided for this node from scheduling
+func GetAvoidPodsFromNodeAnnotations(annotations map[string]string) (v1.AvoidPods, error) {
+	var avoidPods v1.AvoidPods
+	if len(annotations) > 0 && annotations[v1.PreferAvoidPodsAnnotationKey] != "" {
```
liggitt, line 29: this is a "core/v1" helper that is deserializing an alpha annotation (`"scheduler.alpha.kubernetes.io/preferAvoidPods"`) on nodes? what is the graduation plan for this annotation and the data in it?

liggitt (commented):
the repo skeleton, import restrictions, and readme lgtm

I have questions about the suitability of the first helper function moved here in terms of maturity and whether something depending on the exported component-helper library will remain correct over time, since that helper is still pulling data out of a serialized annotation. Passing in a Node object and getting back v1.AvoidPods would seem more correct/stable since that could be updated to use a field rather than an alpha annotation.

staging/src/k8s.io/component-helpers/scheduling/corev1/helpers.go
damemi: That is a good question, but since it's being moved from [`pkg/apis/core/v1`](https://github.com/kubernetes/kubernetes/pull/92507/files#diff-f275cc5451de8d0a74dac8072e93d16cL500) I think that's consistent.

I don't know what the normal graduation plan for annotations is, does it differ from the rest of the api the annotation is a part of? I ask since this annotation is `alpha.kubernetes.io` but lives in `core/v1`, does that mean annotations may be versioned separately from their APIs?

liggitt: > since it's being moved from [`pkg/apis/core/v1`](https://github.com/kubernetes/kubernetes/pull/92507/files#diff-f275cc5451de8d0a74dac8072e93d16cL500) I think that's consistent

I'm looking at it from the perspective of external callers. At it's currently location, it shouldn't have any. Before we move it to a published repo intended to be consumed as a library, we should make sure the signature isn't one that will obviously need to be broken in order to keep working long-term, and this one clearly will (it's documentation states that it is serializing data to an annotation in anticipation of becoming a full-fledged field on the node status, which would require a signature change to continue extracting the data from the node properly)

neolit123: > Passing in a Node object and getting back v1.AvoidPods would seem more correct/stable since that could be updated to use a field rather than an alpha annotation.

+1 and a rename to `GetAvoidPodsFromNode()`

liggitt: > Passing in a Node object and getting back v1.AvoidPods would seem more correct/stable since that could be updated to use a field rather than an alpha annotation.

that would also mean the caller that specifically looks at the annotation value for validation purposes will likely need to continue doing so in some form (or this method would need to continue ensuring the annotation is valid if present)

alculquicondor: We are actually planing to deprecate the feature that uses this annotation, because what it achieves can be done with taints #94008

But this will take 3 or 4 releases to get removed.

damemi: Update on this, that label is actually going to be deprecated apparently: https://github.com/kubernetes/kubernetes/issues/94008#issuecomment-704352146

neolit123: does that mean that the new publically exposed API function GetAvoidPodsFromNode* will become obsolete once the feature is removed?

alculquicondor: Correct. But we have other functions that we want to move to this staging repository. I don't think we are offering guarantees on the stability of this library. The guarantee is pretty similar to that of k8s.io/utils

liggitt: > I don't think we are offering guarantees on the stability of this library. The guarantee is pretty similar to that of k8s.io/utils

k8s.io/utils has presubmits in place to make sure the go API it exposes does not break (https://github.com/kubernetes/utils/blob/master/.github/workflows/test.yml)

alculquicondor: In that case, I suggest we pick a different function for this starting PR.

neolit123: if methods are moved to this new repository without stability guarantees, then possibly they may need to reside in specific sub packages e.g. "experimental" and readme files need to clarify that.

ahg-g: >In that case, I suggest we pick a different function for this starting PR.


my thought as well, lets start with something else to unblock this PR, for example [GetPodPriority](https://github.com/kubernetes/kubernetes/blob/708dd2ea7a7765876cfb0a8329c90e92107860c5/pkg/api/v1/pod/util.go#L357)

ingvagabund: `GetAvoidPodsFromNodeAnnotations` replaced with `GetPodPriority`

staging/src/k8s.io/component-helpers/scheduling/corev1/doc.go
alculquicondor: Remove this line.

staging/src/k8s.io/component-helpers/scheduling/corev1/helpers.go
```diff
@@ -0,0 +1,32 @@
...
+*/
+
+package corev1
+
+import (
+	v1 "k8s.io/api/core/v1"
+)
+
+// GetPodPriority returns priority of the given pod.
+func GetPodPriority(pod *v1.Pod) int32 {
```
alculquicondor, line 24: nit: `PodPriority`

ingvagabund:
Thanks for all the reviews so far. Are there other nits or requests for changes? Or, just those two comments @alculquicondor suggested?

liggitt:
/approve
/lgtm
repo mechanics and move lgtm

/hold for @alculquicondor's comments at https://github.com/kubernetes/kubernetes/pull/92507#pullrequestreview-504797597

soltysh (approved):
/lgtm

soltysh:
> /hold for @alculquicondor's comments at [#92507 (review)](https://github.com/kubernetes/kubernetes/pull/92507#pullrequestreview-504797597)

addressed
/hold cancel

ingvagabund:
@nikhita based on https://github.com/kubernetes/kubernetes/issues/56876#issuecomment-707331065, publishing bot was panicking when publishing the new repo for `v0.18.0-alpha.1` tag. Was it fixed? I don't see the same panic in the latest comment.

Issue #92506, opened 2020-06-25
Node authorizer fails to authorize requests right after kube-apiserver restart


mborsz:
**What happened**:

In kubernetes clusters with O(100k) pods, the initialization of watch cache and shared informer inside of kube-apiserver takes 30-50 seconds. Before this happens, node authorizer is not able to accept any request.
When this happens, kubelets are retrying the requests leading to a retry loop and generating a significant load on the master which in effect can significantly slow down a kube-apiserver initialization, ultimately leading to 429 errors and kube-apiserver crashloop.

The information about state of node authorizer (if it's initialized or not) is not available externally and cannot be used by e.g. load balancer to direct traffic instances that are already initialized.

**What you expected to happen**:

We should have a way to determine if the kube-apiserver is ready to serve requests. Currently /healthz, /livez or /readyz endpoints don't export any information that can be used to determine state of node authorizer. 

In multimaster environment, when one of the kube-apiserver replicas is not yet initialized, we should redirect traffic to other replicas, but currently there is no signal we can use to determine if the replica is initialized or not.

**How to reproduce it (as minimally and precisely as possible)**:

Run scalability load test in 3000 node scale. Right before "Scaling objects" step, restart all masters. They will fail to reboot in ~minute window, so will be killed by livenessProbe.

**Anything else we need to know?**:

While in our repro we suffer from node authorizer not being initialized, this likely also affects other controllers working inside of kube-apiserver.

I suggest adding a poststarthook blocking /healthz that will expose the status of all informers in the system. This way we can use that information in load balancer to direct traffic only to already initialized instances. I will send PR for this.

**Environment**:
- Kubernetes version (use `kubectl version`): tested in 1.17 and 1.18
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

/assign @wojtek-t 
/assign @lavalamp 
/cc @mm4tt

wojtek-t:
While #92508 is universally needed imho, I think that eventually we will need a bit more than that.

In the ideal world, I would like to have a post-start-hook that will not only wait for informers to be synced, but in addition to that will wait for the initial state to actually be applied in node-authorizer state. What I mean by that is "ensuring that handlers for all objects from the "initial list" are actually called".
[...]

PR #92508, opened 2020-06-25
Wait for all informers to sync in /readyz

mborsz:
**What type of PR is this?**
/kind bug

Context: https://github.com/kubernetes/kubernetes/issues/92506

In large clusters, informers in kube-apiserver may require significant time (30-50s) to initialize. Before this happens, kube-apiserver is not able to answer some requests (e.g. node authorizer is not able to accept any request).

This PR adds a way to determine if informers in kube-apiserver are already synced. This can be used via /healthz or /readyz to determine if traffic can be sent to that master replica

```release-note
PostStartHook "generic-apiserver-start-informers" waits for all informers to sync
```

```docs

```

/assign @lavalamp

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/server/config.go
+++ b/staging/src/k8s.io/apiserver/pkg/server/config.go
@@ -605,12 +605,17 @@ func (c completedConfig) New(name string, delegationTarget DelegationTarget) (*G
 	}
 
 	genericApiServerHookName := "generic-apiserver-start-informers"
-	if c.SharedInformerFactory != nil && !s.isPostStartHookRegistered(genericApiServerHookName) {
-		err := s.AddPostStartHook(genericApiServerHookName, func(context PostStartHookContext) error {
-			c.SharedInformerFactory.Start(context.StopCh)
-			return nil
-		})
-		if err != nil {
+	if c.SharedInformerFactory != nil {
+		if !s.isPostStartHookRegistered(genericApiServerHookName) {
+			err := s.AddPostStartHook(genericApiServerHookName, func(context PostStartHookContext) error {
+				c.SharedInformerFactory.Start(context.StopCh)
+				return nil
+			})
+			if err != nil {
+				return nil, err
+			}
+		}
+		if err := s.addReadyzChecks(healthz.NewInformerSyncHealthz(c.SharedInformerFactory)); err != nil {
 			return nil, err
 		}
 	}
--- a/staging/src/k8s.io/apiserver/pkg/server/healthz/healthz.go
+++ b/staging/src/k8s.io/apiserver/pkg/server/healthz/healthz.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apiserver/pkg/endpoints/metrics"
 	"k8s.io/apiserver/pkg/server/httplog"
+	"k8s.io/client-go/informers"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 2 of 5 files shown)


staging/src/k8s.io/apiserver/pkg/server/config.go
```diff
@@ -608,6 +608,20 @@ func (c completedConfig) New(name string, delegationTarget DelegationTarget) (*G
...
+			klog.Infof("SharedInformerFactory.WaitForCacheSync finished with: %v", started)
+			var notStarted []string
+			for informType, started := range started {
+				if !started {
+					klog.Warningf("informer for %v failed to sync", informType.String())
+					notStarted = append(notStarted, informType.String())
+				}
+			}
+			if len(notStarted) != 0 {
+				return fmt.Errorf("%d informers not started yet: %v", len(notStarted), notStarted)
```
lavalamp, line 623: Does `WaitForCacheSync` ever return when a cache hasn't synced yet, but the error is transient? I think returning an error here will cause a liveness failure and not just a readiness failure. Even if the error is not transient, then will a restart really help?

I guess I'm saying you might want to take a look at which of healthz/livez/readyz include this.

staging/src/k8s.io/apiserver/pkg/server/config.go
```diff
@@ -608,6 +608,20 @@ func (c completedConfig) New(name string, delegationTarget DelegationTarget) (*G
 	if c.SharedInformerFactory != nil && !s.isPostStartHookRegistered(genericApiServerHookName) {
 		err := s.AddPostStartHook(genericApiServerHookName, func(context PostStartHookContext) error {
 			c.SharedInformerFactory.Start(context.StopCh)
+
+			// Wait for all caches to sync.
+			started := c.SharedInformerFactory.WaitForCacheSync(context.StopCh)
```
lavalamp, line 613: For unit-testability, consider putting this functionality into a separate function?

staging/src/k8s.io/apiserver/pkg/server/config.go
```diff
@@ -608,6 +608,20 @@ func (c completedConfig) New(name string, delegationTarget DelegationTarget) (*G
 	if c.SharedInformerFactory != nil && !s.isPostStartHookRegistered(genericApiServerHookName) {
 		err := s.AddPostStartHook(genericApiServerHookName, func(context PostStartHookContext) error {
```
lavalamp, line 609: Consider adding a second post start hook rather than modify this one. Adding stuff to this one could surprise people who have tuned healthz/livez/readyz. It'll also let people decide e.g. that waiting is optional, but starting is not.

lavalamp (commented):
I think I'd like to see a test of some sort?

lavalamp:
@deads2k might have opinions about the post start hook.

@logicalhan and/or @sttts might have opinions about which of healthz/livez/readyz should care about this post start hook.

staging/src/k8s.io/apiserver/pkg/server/config.go
mborsz: The only case when WaitForCacheSync returns and some informer isn't initialized is when context is cancelled.

I will check how to add this to readyz and not livez

mborsz: The most common cases when WaitForCacheSync fails is either etcd being down, improper etcd configuration or bug in code. None of them can be fixed by restarting kube-apiserver.

Using readiness probe for this purpose sounds reasonable to me.

mborsz:
If informers fail to sync, this likely means unhealthy etcd, improper configuration of kube-apiserver or a bug in the code. I don't think that restarting kube-apiserver may fix any of those.

For that reason I migrated this check to /readyz endpoint that seemed to be the best choice for this kind of check.
I see that 'crd-informer-synced' solves similar problem for crds specifically and it uses poststarthook, which block /livez and /healthz in addition to /readyz. We may want to check if this is really correct.

I pushed a new version with separate readyz check. I'm still working on tests, but will appreciate feedback on code.

mborsz:
@logicalhan @sttts @deads2k Are you with adding this as a /readyz health check only?

@lavalamp I would like to cherry pick this back to old releases, do you know which k8s release adds /readyz endpoint?

deads2k:
> @deads2k might have opinions about the post start hook.

As I recall, we actually had this check in openshift and chose to remove it because it wasn't required for request safety and answering some requests generally worked better than answering no requests and failing a health check.  /readyz would be less worrying from a "death cycle" point of view, but @sttts may still have opinions since he's currently looking at those.

wojtek-t:
> As I recall, we actually had this check in openshift and chose to remove it because it wasn't required for request safety and answering some requests generally worked better than answering no requests and failing a health check. /readyz would be less worrying from a "death cycle" point of view, but @sttts may still have opinions since he's currently looking at those.

It's not required for some requests, but it is required for others (e.g. anything that requires node-authorizer more-or-less). I agree that in the world where we just had healthz/ it wasn't perfect, but putting it just for /readyz actually seems desired to me.

sttts:
Readyz sounds much better than healthz. This is not about being healthy. Would like to hear @logicalhan 's opinion. He reworked the logic there quite a bit few months ago.

staging/src/k8s.io/apiserver/pkg/server/healthz/healthz.go
```diff
@@ -81,6 +82,76 @@ func (l *log) Check(_ *http.Request) error {
...
+// 		if !started {
+// 			klog.Warningf("informer for %v failed to sync", informType.String())
+// 			notStarted = append(notStarted, informType.String())
+// 		}
+// 	}
+// 	if len(notStarted) != 0 {
+// 		return fmt.Errorf("%d informers not started yet: %v", len(notStarted), notStarted)
+// 	}
+// 	return nil
+// }
```
logicalhan, line 129: ?

staging/src/k8s.io/apiserver/pkg/server/healthz/healthz.go
```diff
@@ -81,6 +82,76 @@ func (l *log) Check(_ *http.Request) error {
...
+	if len(notStarted) != 0 {
+		return fmt.Errorf("%d informers not started yet: %v", len(notStarted), notStarted)
+	}
+	return nil
+	// select {
+	// case <-i.done:
+	// 	return i.err.Load().(error)
+	// default:
+	// 	return errors.New("not finished")
+	// }
```
logicalhan, line 152: same?

test/integration/master/kube_apiserver_test.go
```diff
@@ -113,10 +118,10 @@ func TestLivezAndReadyz(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
-	if !endpointReturnsStatusOK(client, "/livez") {
+	if !endpointReturnsStatusOK(t, client, "/livez") {
 		t.Fatalf("livez should be healthy")
 	}
-	if !endpointReturnsStatusOK(client, "/readyz") {
+	if !endpointReturnsStatusOK(t, client, "/readyz?verbose=true") {
```
logicalhan, line 124: the `verbose=true` isn't necessary here, right?

logicalhan (commented):
Yeah `/readyz` sounds appropriate for this. We should probably add a release note stating that we are adding a readyz check which further bifurcates behavior from the deprecated healthz endpoint (as a word of warning).

wojtek-t:
Given Maciek will not be around for the next couple weeks, I forked this PR into https://github.com/kubernetes/kubernetes/pull/92644 and addressed its comments.

Please continue discussion on that other PR (I'm closing this one in factor of the new one).

Issue #28260, opened 2016-06-30
Rework kubeconfig reading/writing to be less complex/racy


cjcullen:
Following up on the discussion from #28197.

Parallel calls to ModifyConfig can do bad things to your kubeconfig file(s). We put together some quick fixes (#28034 & #28232), neither of which were all that satisfying because of the complexity of loading/storing kubeconfig from multiple locations.

@krousey @deads2k @jlowdermilk @ghodss

smarterclayton:
Maybe the real question is - why are we concurrently writing / reading from kubeconfig?  When as a user will that result in a great outcome?

deads2k:
> Maybe the real question is - why are we concurrently writing / reading from kubeconfig? When as a user will that result in a great outcome?

Our ops team hit it too.  For us, it's usually happening when running multiple parallel maintenance/test scripts using the same creds that all create new-projects.  For kube, I think its happening because they're refreshing a token every five multiple from multiple parallel scripts.

smarterclayton:
As mentioned on the other thread, that's implicitly broken (I don't
consider calling writing to the same config to change context , even with
locks, valid).

On Jun 30, 2016, at 5:03 AM, David Eads <notifications@github.com> wrote:

Maybe the real question is - why are we concurrently writing / reading from
kubeconfig? When as a user will that result in a great outcome?

Our ops team hit it too. For us, it's usually happening when running
multiple parallel maintenance/test scripts using the same creds that all
create new-projects. For kube, I think its happening because they're
refreshing a token every five multiple from multiple parallel scripts.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/kubernetes/kubernetes/issues/28260#issuecomment-229638715>,
or mute the thread
<https://github.com/notifications/unsubscribe/ABG_p3-H6oKf9RB5X7v85cHbHqovu3oWks5qQ7CYgaJpZM4JBuOt>
.

smarterclayton:
Possibly one option would be split the token out into its own file if we're rapidly updating it, and reference it from the config file via path.

cjcullen:
Yeah, that was the original plan. Then you are back in the "multiple files for config" world, but maybe that's okay for just the automatically-refreshed token case. For the general case (being able to persist config for a specific auth-provider), it's still nicer to have all config in one place.

smarterclayton:
Absolutely.  Frequently refreshed tokens was not the original use case,
[...]

PR #28232, opened 2016-06-29
Lock all possible kubecfg files at the beginning of ModifyConfig.

cjcullen:
Prevent concurrent calls to ModifyConfig on the same (or overlapping) kubeconfig files.

PR #92513, opened 2020-06-25
Add the ability to opt-out from config lock file

soltysh:
**What type of PR is this?**
/kind cleanup

Back in the early days we thought it was smart to have the lock mechanism to prevent concurrent 
modification of the kubeconfig file (see https://github.com/kubernetes/kubernetes/issues/28260 https://github.com/kubernetes/kubernetes/pull/28232) but currently we are advising users to use multiple files when there's a risk of concurrent access from different clusters. I'm pinging the original commenters of the issue back from then.

This PR adds the ability for a user to decide if he cares about locking through `DISABLE_KUBECONFIG_LOCK`.
It's on by default to maintain backwards compatibility but I'm open to making the locking optional
or even remove it, if majority agrees.

/assign @deads2k @smarterclayton @liggitt 

```release-note
Add the ability to disable kubeconfig file lock through DISABLE_KUBECONFIG_LOCK env var
```

### DIFF
--- a/staging/src/k8s.io/client-go/tools/clientcmd/config.go
+++ b/staging/src/k8s.io/client-go/tools/clientcmd/config.go
@@ -58,6 +58,15 @@ type PathOptions struct {
 	LoadingRules *ClientConfigLoadingRules
 }
 
+var (
+	// UseModifyConfigLock ensures that access to kubeconfig file using ModifyConfig method
+	// is being guarded by a lock file.
+	// This variable is intentionaly made public so other consumers of this library
+	// can modify its default behavior, but be caution when disabling it since
+	// this will make your code not threadsafe.
+	UseModifyConfigLock = true
+)
+
 func (o *PathOptions) GetEnvVarFiles() []string {
 	if len(o.EnvVar) == 0 {
 		return []string{}
@@ -156,15 +165,17 @@ func NewDefaultPathOptions() *PathOptions {
 // that means that this code will only write into a single file.  If you want to relativizePaths, you must provide a fully qualified path in any
 // modified element.
 func ModifyConfig(configAccess ConfigAccess, newConfig clientcmdapi.Config, relativizePaths bool) error {
-	possibleSources := configAccess.GetLoadingPrecedence()
-	// sort the possible kubeconfig files so we always "lock" in the same order
-	// to avoid deadlock (note: this can fail w/ symlinks, but... come on).
-	sort.Strings(possibleSources)
-	for _, filename := range possibleSources {
-		if err := lockFile(filename); err != nil {
-			return err
+	if UseModifyConfigLock {
+		possibleSources := configAccess.GetLoadingPrecedence()
+		// sort the possible kubeconfig files so we always "lock" in the same order
+		// to avoid deadlock (note: this can fail w/ symlinks, but... come on).
+		sort.Strings(possibleSources)
+		for _, filename := range possibleSources {
+			if err := lockFile(filename); err != nil {
+				return err
+			}
+			defer unlockFile(filename)
 		}
-		defer unlockFile(filename)
 	}
 
 	startingConfig, err := configAccess.GetStartingConfig()


staging/src/k8s.io/client-go/tools/clientcmd/config.go
```diff
@@ -58,6 +58,12 @@ type PathOptions struct {
 	LoadingRules *ClientConfigLoadingRules
 }
 
+var (
+	// UseConfigLock ensures that access to kubeconfig files is being guarded by a lock file
+	// Users can opt-out from using it by setting KUBECONFIG_NO_LOCK environment variable
+	UseConfigLock bool = len(os.Getenv("DISABLE_KUBECONFIG_LOCK")) == 0
```
deads2k, line 64: This should actually parse the bool for equivalence to true.  I'd like to prefix with a `KUBE_` on the env var.

You should also explain that this is public because it allows different binaries to directly change the default behavior.

liggitt:
I agree the current locking mechanism is not good, but concurrent modification still seems like a problem we should avoid... corrupting a file in an unrecoverable way seems like really poor behavior.

Have we considered either of these alternatives:
1. switching to a write-to-temp + move approach to atomically replace content
2. using something like flock to lock the specific file we want to write to (rather than locking all in the chain)

soltysh:
> * switching to a write-to-temp + move approach to atomically replace content

That seem overly complicated for this simple use case, imo. 

> * using something like flock to lock the specific file we want to write to (rather than locking all in the chain)

I might be tempted by this approach, only our current [flock implementation](https://github.com/kubernetes/kubernetes/blob/master/pkg/util/flock/flock_other.go) is unix only. We'd probably need to fix that along the way.

But above seem to be orthogonal to the current locking mechanism. This PR is rather exposing both library consumers and kubectl users explicit ability to not to use the locking. I'm perfectly fine with having both, but in separate PRs, if you don't mind @liggitt

staging/src/k8s.io/client-go/tools/clientcmd/config.go
soltysh: Since the only consumers of this envvar are kubectl user, I'm inclined to move that env bit entirely to kubectl.

sftim: @soltysh this now doesn't match the PR description BTW

staging/src/k8s.io/client-go/tools/clientcmd/config.go
```diff
@@ -58,6 +58,13 @@ type PathOptions struct {
 	LoadingRules *ClientConfigLoadingRules
 }
 
+var (
+	// UseConfigLock ensures that access to kubeconfig files is being guarded by a lock file.
+	// This variable is intentionaly made public so other consumers of this library
+	// can modify its default behavior.
+	UseConfigLock = true
```
liggitt, line 65: make the name and godoc for this specific to ModifyConfig (e.g. `UseModifyConfigLock`) and note that it is not threadsafe

liggitt:
> > * switching to a write-to-temp + move approach to atomically replace content
> 
> That seem overly complicated for this simple use case, imo.

Thinking about this more, I think this is exactly what we should do. That would not only protect against corruption from simultaneous writers, but would protect simultaneous readers. Today, they have the potential to read half-written data.

staging/src/k8s.io/client-go/tools/clientcmd/config.go
liggitt: For exposing this in kubectl, I would expect a `--nolock` option to the modifying config commands. That seems like it would work better plumbed into ModifyConfig than as a global var

staging/src/k8s.io/client-go/tools/clientcmd/config.go
soltysh: You still have that ability, only not directly. Only consumers of the library have that ability.

staging/src/k8s.io/client-go/tools/clientcmd/config.go
soltysh: Yeah, I was thinking of a flag, rather than env var.

soltysh:
> Thinking about this more, I think this is exactly what we should do to fix the lock issue. That would not only protect against corruption from simultaneous writers, but would protect simultaneous readers. Today, they have the potential to read half-written data.

Fair point, I'll start working on a followup next week.

PR #92074, opened 2020-06-12
Test pod becoming schedulable when another pod is added or updated

nodo:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
It adds two integration tests. 

The first one is for a pod creation event. The pod we are testing is scheduled because a new pod with affinity his scheduled.

The second one is for a pod update event. The pod we are testing is scheduled because a previously scheduled pod is updated with a new label (which correspond to the affinity of the pod we are testing).

**Which issue(s) this PR fixes**:
Fixes of #91111

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

PR #92514, opened 2020-06-25
Use NodeWrapper to directly initialize nodes with labels

nodo:
**What type of PR is this?**
/kind cleanup

In https://github.com/kubernetes/kubernetes/pull/92074#discussion_r440398314, we discussed the possibility to use `NodeWrapper` in the scheduler integration tests so that we have more flexibility when creating the nodes. For instance we can create nodes directly with a label.

```release-note
NONE
```


nodo:
@alculquicondor @Huang-Wei I have opened this PR as WIP go gather your feedback about this cleanup.

In particular:
* What do you think about keeping `initNode` so that we can easily initialize nodes with certain default?
* I deliberately created `createNodeFromWrapper` to keep this PR simple, however, do you think we should change `createNode` to use `NodeWrapper`?
* Also, if you agree I think we should also change `createNodes` to use `createNodeFromWrapper`, but I wanted to hear from you first.

Huang-Wei:
> - What do you think about keeping initNode and use it in the integration tests, so that we can easily initialize nodes with certain default?

There are just 2 callers using `initNode()`:

```
⇒  ag initNode test/integration/scheduler
test/integration/scheduler/util.go
151:// initNode returns a node with the given resource list and images. If 'res' is nil, a predefined amount of
153:func initNode(name string, res *v1.ResourceList, images []v1.ContainerImage) *v1.Node {
174:	return cs.CoreV1().Nodes().Create(context.TODO(), initNode(name, res, nil), metav1.CreateOptions{})
179:	return cs.CoreV1().Nodes().Create(context.TODO(), initNode(name, res, images), metav1.CreateOptions{})
```

I think it makes more sense to eliminate it by migrating to NodeWrapper.

> - I deliberately created createNodeFromWrapper to keep this PR small in scope, however, do you think we should change createNode to use NodeWrapper as well?
> - Also, if you agree I think we should also change createNodes to use createNodeFromWrapper, but I wanted to hear from you first.

My 2 cents is to use NodeWrapper inside createNode() & createNodes() to craft the Node(s) object first, and then call API server to create them.

pkg/scheduler/testing/wrappers.go
```diff
@@ -390,3 +390,18 @@ func (n *NodeWrapper) Label(k, v string) *NodeWrapper {
 	n.Labels[k] = v
 	return n
 }
+
+func (n *NodeWrapper) Unschedulable(v bool) *NodeWrapper {
+	n.Spec.Unschedulable = v
+	return n
+}
+
+func (n *NodeWrapper) Capacity(capacity v1.ResourceList) *NodeWrapper {
```
Huang-Wei, line 399: I'd suggest using paired key-value string (or just values) instead of v1.ResourceList here. v1.ResourceList is too low-level API, which is friendly for new-users to build, and if you look at the usage of createNode(), there is only one caller specifying a non-nil resourceList.

**Option 1:**

```go
func (n *NodeWrapper) Capacity(pairs ...string) *NodeWrapper {
	res := v1.ResourceList{}
	for i := 0; i < len(pairs); i += 2 {
		k, v := pairs[i], pairs[i+1]
		res[v1.ResourceName(k)] = resource.MustParse(v)
	}
	n.Status.Capacity, n.Status.Allocatable = res, res
	return n
}
```

**Option 2:**

```go
func (n *NodeWrapper) Capacity(cpu, mem string, pods int64) *NodeWrapper {
	res := v1.ResourceList{
		v1.ResourcePods: *resource.NewQuantity(pods, resource.DecimalSI),
	}
	if len(cpu) != 0 {
		res[v1.ResourceCPU] = resource.MustParse(cpu)
	}
	if len(mem) != 0 {
		res[v1.ResourceMemory] = resource.MustParse(mem)
	}
	n.Status.Capacity, n.Status.Allocatable = res, res
	return n
}
```

pkg/scheduler/testing/wrappers.go
```diff
@@ -390,3 +390,18 @@ func (n *NodeWrapper) Label(k, v string) *NodeWrapper {
...
+	n.Status.Capacity = capacity
+	return n
+}
+
+func (n *NodeWrapper) Images(images []v1.ContainerImage) *NodeWrapper {
```
Huang-Wei, line 404: Similar as above, this function should take the burden of building []v1.ContainerImage inside, so users just pass in image names, etc.

test/integration/scheduler/util.go
```diff
@@ -150,33 +151,32 @@ func waitForNodeLabels(cs clientset.Interface, nodeName string, labels map[strin
...
-	}
-	return n
+	return st.MakeNode().
+		Name(name).
+		Unschedulable(false).
+		Capacity(*res).
+		Images(images)
+}
+
+func createNodeFromWrapper(cs clientset.Interface, wrapper *st.NodeWrapper) (*v1.Node, error) {
```
alculquicondor, line 168: I prefer we just call .Obj from the callers.

test/integration/scheduler/util.go
```diff
@@ -150,33 +151,32 @@ func waitForNodeLabels(cs clientset.Interface, nodeName string, labels map[strin
...
+func initNode(name string, res *v1.ResourceList, images []v1.ContainerImage) *st.NodeWrapper {
 	if res == nil {
 		res = &v1.ResourceList{
 			v1.ResourcePods: *resource.NewQuantity(32, resource.DecimalSI),
 		}
 	}
 
-	n := &v1.Node{
-		ObjectMeta: metav1.ObjectMeta{Name: name},
-		Spec:       v1.NodeSpec{Unschedulable: false},
```
alculquicondor, line 163: @Huang-Wei do you know why were we setting this? AFAIK, there is no controller-manager in integration tests that would set this to true, or is there?

test/integration/scheduler/predicates_test.go
```diff
@@ -1155,14 +1152,11 @@ func TestUnschedulablePodBecomesSchedulable(t *testing.T) {
 		{
 			name: "scheduled pod gets updated to match affinity",
 			init: func(cs kubernetes.Interface, ns string) error {
-				node, err := createNode(cs, "node-1", nil)
+				nodeWrapper := initNode("node-1", nil, nil).Label("region", "test")
```
alculquicondor, line 1155: prefer to call st.MakeNode directly and we explicitly set the resourcelist.

pkg/scheduler/testing/wrappers.go
alculquicondor: +1

pkg/scheduler/testing/wrappers.go
alculquicondor: I would prefer Option 1, but accepting map[string]string, instead of the pairs of lists

Huang-Wei: yup, option 1 is also my preference.

test/integration/scheduler/util.go
Huang-Wei: `spec.Unschedulable` should be [deprecated](https://github.com/kubernetes/kubernetes/blob/cb37c0884664fd2f4a2800661cf92d7c32ee9143/pkg/scheduler/framework/plugins/nodeunschedulable/node_unschedulable.go#L60) IMO, so yes, we don't quite need to set it.

pkg/scheduler/testing/wrappers.go
Huang-Wei: nit: `map[v1.ResourceName]string` should be less error-prone.

nodo: Good idea, I will make the change!

test/integration/scheduler/util.go
nodo: Oh I didn't know, I will change it then, thanks!

nodo:
@alculquicondor @Huang-Wei Thanks a lot for the feedback. I have done some changes and cleaned up the rest of the integration tests. Feel free to review it when you have time.

pkg/scheduler/testing/wrappers.go
```diff
@@ -390,3 +391,25 @@ func (n *NodeWrapper) Label(k, v string) *NodeWrapper {
 	n.Labels[k] = v
 	return n
 }
+
+// Capacity sets the capacity and the allocatable resources of the inner node.
+// Each entry in `resources` corresponds to a resource name and its quantity.
+func (n *NodeWrapper) Capacity(resources map[v1.ResourceName]string) *NodeWrapper {
+	res := v1.ResourceList{}
```
alculquicondor, line 398: nit: `res := make(v1.ResourceList)`

test/integration/scheduler/util.go
```diff
@@ -194,6 +167,10 @@ func createNodes(cs clientset.Interface, prefix string, res *v1.ResourceList, nu
 	return nodes[:], nil
 }
 
+func defaultNodeWrapper() *st.NodeWrapper {
```
alculquicondor, line 170: I prefer we don't hide the details in this function.
Each caller should define the resources they need in their tests.
I know this goes against having repeated code, but this is actually a good practice when writing tests.

pkg/scheduler/testing/wrappers.go
```diff
@@ -390,3 +391,25 @@ func (n *NodeWrapper) Label(k, v string) *NodeWrapper {
...
+	for name, value := range resources {
+		res[name] = resource.MustParse(value)
+	}
+	n.Status.Capacity, n.Status.Allocatable = res, res
```
Huang-Wei, line 402: would be good to ensure `v1.ResourcePods` is always present, otherwise some pod scheduling may fail due to "no Pods quota".

FYI: I put it in https://github.com/kubernetes/kubernetes/pull/92571/files?file-filters%5B%5D=.go#diff-1cea28cd0be3cdbab57f5dc287dc98c0R367-R383

pkg/scheduler/testing/wrappers.go
```diff
@@ -390,3 +391,25 @@ func (n *NodeWrapper) Label(k, v string) *NodeWrapper {
...
+	return n
+}
+
+// Images sets the images of the inner node. Each entry in `images` corresponds
+// to an image name and its size in bytes.
+func (n *NodeWrapper) Images(images map[string]int64) *NodeWrapper {
+	containerImages := []v1.ContainerImage{}
```
Huang-Wei, line 409: nit: `var containerImages []v1.ContainerImage`.

test/integration/scheduler/util.go
Huang-Wei: IMO `v1.ResourcePods` is a bit special, unlike zero cpu/mem, a zero Pods number would simply fail a NodeFit plugin in any case. Which is usually not what we want. So having a default `v1.ResourcePods` looks good to me. However, I'd prefer to set the default value on NodesWrapper side - which would also benefit UT (I did that way in another PR: https://github.com/kubernetes/kubernetes/pull/92571/files?file-filters%5B%5D=.go#diff-1cea28cd0be3cdbab57f5dc287dc98c0R367-R383)

If you do so, this function is not needed anymore - you can just replace callers of `defaultNodeWrapper()` directly to `st.MakeNode()`.

Huang-Wei (commented):
Just some nits.

test/integration/scheduler/util.go
alculquicondor: Sg, we can just do `res[pods] = 32` at the beginning of the function, and the loop would override it.

Huang-Wei: Yup, that would save one `if` block.

nodo: Sounds good, I have made the change.

pkg/scheduler/testing/wrappers.go
```diff
@@ -390,3 +391,27 @@ func (n *NodeWrapper) Label(k, v string) *NodeWrapper {
 	n.Labels[k] = v
 	return n
 }
+
+// Capacity sets the capacity and the allocatable resources of the inner node.
+// Each entry in `resources` corresponds to a resource name and its quantity.
```
alculquicondor, line 396: Indicate that sets 32 pods limit by default

test/integration/scheduler/framework_test.go
```diff
@@ -1841,7 +1842,7 @@ func initTestSchedulerForFrameworkTest(t *testing.T, testCtx *testutils.TestCont
 	go testCtx.Scheduler.Run(testCtx.Ctx)
 
 	if nodeCount > 0 {
-		_, err := createNodes(testCtx.ClientSet, "test-node", nil, nodeCount)
+		_, err := createNodes(testCtx.ClientSet, "test-node", st.MakeNode(), nodeCount)
```
alculquicondor, line 1845: Shouldn't MakeNode call `Capacity(nil)` or something like that?

nodo: Totally! My bad, let me fix it.

alculquicondor (commented):
/lgtm

Thanks for the cleanup!

Huang-Wei:
/approve
/retest

Thanks @nodo .

PR #92518, opened 2020-06-25
Update LoadBalancer service on node IP change

Sh4d1:
Signed-off-by: Patrik Cyvoct <patrik@ptrk.io>

**What type of PR is this?**

/kind bug

This PR allows to resync a LoadBalancer service when the node IP changes (changing the floating IP of the node for instance)

```release-note
NONE
```

```docs

```

/sig cloud-provider

### DIFF
--- a/staging/src/k8s.io/cloud-provider/controllers/service/controller.go
+++ b/staging/src/k8s.io/cloud-provider/controllers/service/controller.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/runtime"
@@ -693,6 +694,10 @@ func shouldSyncNode(oldNode, newNode *v1.Node) bool {
 	}
 
 	if !reflect.DeepEqual(oldNode.Labels, newNode.Labels) {
+		return true
+	}
+
+	if !equality.Semantic.DeepEqual(oldNode.Status.Addresses, newNode.Status.Addresses) {
 		return true
 	}
 
--- a/staging/src/k8s.io/cloud-provider/controllers/service/controller.go
+++ b/staging/src/k8s.io/cloud-provider/controllers/service/controller.go
@@ -179,7 +179,11 @@ func New(
 					return
 				}
 
-				s.triggerNodeSync()
+				if !equality.Semantic.DeepEqual(oldNode.Status.Addresses, curNode.Status.Addresses) {
+					s.forceNodeSync()
+				} else {
+					s.triggerNodeSync()
+				}
 			},
 			DeleteFunc: func(old interface{}) {
 				s.triggerNodeSync()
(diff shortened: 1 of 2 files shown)


Sh4d1:
@bowei @thockin could this get merged before 1.19 code freeze?

staging/src/k8s.io/cloud-provider/controllers/service/controller.go
```diff
@@ -673,6 +674,10 @@ func shouldSyncNode(oldNode, newNode *v1.Node) bool {
 		return true
 	}
 
+	if !equality.Semantic.DeepEqual(oldNode.Status.Addresses, newNode.Status.Addresses) {
+		return true
```
andrewsykim, line 678: Even with `shouldSyncNode` returning true here, `nodeSlicesEqualForLB` compares the node list by name so it may not actually trigger a sync.

Sh4d1: @andrewsykim but if equal, `updateLoadBalancerHosts` is still called, so it works (and it works when I tested too :smile: )

andrewsykim: Right, but last time I checked, `updateLoadBalancerHosts` only attempts to retry failed Services. It may have only worked because the controller does a resync of all Services every 100 seconds.

andrewsykim: Or unless your Service was in a failed state before

Sh4d1: True I missed this one https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cloud-provider/controllers/service/controller.go#L713 😅

Is adding the same check before not too expensive ? 🤔

andrewsykim: Which check? Adding all services to `servicesToUpdate`?

andrewsykim: I think it would be too expensive because we would then sync all LBs even when the node slice is equal.

Sh4d1: Nop I meant the deepequal on the address on the same if than the `nodeSlicesEqualForLB`

andrewsykim: Ah.. hmm that one is tricky because not all LB implementations actually check the node address. Which makes me think that the address equal check here may be redundant too. Need to think about this one more.

Sh4d1: Hum true. But I guess nodes changing IP is very rare so it should rarely happen ?

Sh4d1: @andrewsykim up? :smile:

Sh4d1: @andrewsykim double up :smile:

Sh4d1:
@bowei could I get a review? Following the discussion with @andrewsykim I think the nodes IP changes so rarely (and not on all clouds) that it's worth the cost

andrewsykim:
I think my comment https://github.com/kubernetes/kubernetes/pull/92518/files#r459087217 still applies? Even if we `shouldSyncNode` returned true on node address change, `nodeSlicesEqualForLB` doesn't check node addresses so the sync wouldn't happen anyways. Am I missing something here?

Sh4d1:
> I think my comment https://github.com/kubernetes/kubernetes/pull/92518/files#r459087217 still applies? Even if we `shouldSyncNode` returned true on node address change, `nodeSlicesEqualForLB` doesn't check node addresses so the sync wouldn't happen anyways. Am I missing something here?

Yeah but I was proposing adding a check in https://github.com/kubernetes/kubernetes/blob/6352f01e662b089b8a2fb3b367c0cc5d844afd85/staging/src/k8s.io/cloud-provider/controllers/service/controller.go#L629 about the Nodes IP too?

Sh4d1:
@andrewsykim is this solution better for you?

PR #92519, opened 2020-06-25
add validation functions for metav1.Conditions

deads2k:
The last commit adds validation rules. It's best to be as constraining as we think is reasonable, but we should try to make it possible to be consistent with generated schemas.

@liggitt as promised.  See if you agree with the restrictions.

/kind feature
/priority important-soon
@kubernetes/sig-api-machinery-feature-requests 

/hold this still needs tests

@damemi @sttts I tried to make these validation rules easy to port into kubebuilder markers in the types themselves so CRDs will generate cleanly.  Can you confirm these look good.
@mfojtik do these rules correspond to how you've seen them used too?

```release-note
NONE
```

related to https://github.com/kubernetes/enhancements/issues/1623

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
+++ b/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
@@ -18,6 +18,7 @@ package validation
 
 import (
 	"fmt"
+	"regexp"
 	"unicode"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -183,4 +184,79 @@ func ValidateManagedFields(fieldsList []metav1.ManagedFieldsEntry, fldPath *fiel
 		}
 	}
 	return allErrs
+}
+
+func ValidateConditions(conditions []metav1.Condition, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+
+	conditionTypeToFirstIndex := map[string]int{}
+	for i, condition := range conditions {
+		if _, ok := conditionTypeToFirstIndex[condition.Type]; ok {
+			allErrs = append(allErrs, field.Duplicate(fldPath.Index(i).Child("type"), condition.Type))
+		} else {
+			conditionTypeToFirstIndex[condition.Type] = i
+		}
+
+		allErrs = append(allErrs, ValidateCondition(condition, fldPath.Index(i))...)
+	}
+
+	return allErrs
+}
+
+// validConditionStatuses is used internally to check validity and provide a good message
+var validConditionStatuses = sets.NewString(string(metav1.ConditionTrue), string(metav1.ConditionFalse), string(metav1.ConditionUnknown))
+
+const (
+	maxReasonLen  = 1 * 1024
+	maxMessageLen = 32 * 1024
+)
+
+func ValidateCondition(condition metav1.Condition, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+
+	// type is set and is a valid format
+	allErrs = append(allErrs, ValidateLabelName(condition.Type, fldPath.Child("type"))...)
+
+	// status is set and is an accepted value
+	if !validConditionStatuses.Has(string(condition.Status)) {
+		allErrs = append(allErrs, field.NotSupported(fldPath.Child("status"), condition.Status, validConditionStatuses.List()))
+	}
+
+	if condition.ObservedGeneration < 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("observedGeneration"), condition.ObservedGeneration, "must be greater than or equal to zero"))
+	}
+
+	if condition.LastTransitionTime.IsZero() {
+		allErrs = append(allErrs, field.Required(fldPath.Child("lastTransitionTime"), "must be set"))
+	}
+
+	if len(condition.Reason) == 0 {
+		allErrs = append(allErrs, field.Required(fldPath.Child("reason"), "must be set"))
+	} else {
+		for _, currErr := range isValidConditionReason(condition.Reason) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("reason"), condition.Reason, currErr))
+		}
+		if len(condition.Reason) > maxReasonLen {
+			allErrs = append(allErrs, field.TooLong(fldPath.Child("reason"), condition.Reason, maxReasonLen))
+		}
+	}
+
+	if len(condition.Message) > maxMessageLen {
+		allErrs = append(allErrs, field.TooLong(fldPath.Child("message"), condition.Message, maxMessageLen))
+	}
+
+	return allErrs
+}
+
+const conditionReasonFmt string = "[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?"
+const conditionReasonErrMsg string = "a condition reason must start with alphabetic character, optionally followed by a string of alphanumeric characters or '_,:', and must end with an alphanumeric character or '_'"
+
+var conditionReasonRegexp = regexp.MustCompile("^" + conditionReasonFmt + "$")
+
+// isValidConditionReason tests for a string that conforms to rules for condition reasons. This checks the format, but not the length.
+func isValidConditionReason(value string) []string {
+	if !conditionReasonRegexp.MatchString(value) {
+		return []string{validation.RegexError(conditionReasonErrMsg, conditionReasonFmt, "my_name", "MY_NAME", "MyName", "ReasonA,ReasonB", "ReasonA:ReasonB")}
+	}
+	return nil
 }
--- a/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation_test.go
+++ b/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation_test.go
@@ -293,3 +293,136 @@ func TestValidateMangedFieldsValid(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateConditions(t *testing.T) {
+	tests := []struct {
+		name         string
+		conditions   []metav1.Condition
+		validateErrs func(t *testing.T, errs field.ErrorList)
+	}{
+		{
+			name: "bunch-of-invalid-fields",
+			conditions: []metav1.Condition{{
+				Type:               ":invalid",
+				Status:             "unknown",
+				ObservedGeneration: -1,
+				LastTransitionTime: metav1.Time{},
+				Reason:             "invalid;val",
+				Message:            "",
+			}},
+			validateErrs: func(t *testing.T, errs field.ErrorList) {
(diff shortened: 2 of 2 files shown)


staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
```diff
@@ -184,3 +184,65 @@ func ValidateManagedFields(fieldsList []metav1.ManagedFieldsEntry, fldPath *fiel
...
 	return allErrs
 }
+
+func ValidateConditions(conditions []metav1.Condition, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+
+	conditionTypeToFirstIndex := map[string]int{}
+	for i, condition := range conditions {
+		if firstIndex, ok := conditionTypeToFirstIndex[condition.Type]; ok {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("type"), condition.Type, fmt.Sprintf("type was used by %v", fldPath.Index(firstIndex).Child("type"))))
```
liggitt, line 194:
```suggestion
			allErrs = append(allErrs, field.Duplicate(fldPath.Index(i).Child("type"), condition.Type))
```

mfojtik:
@deads2k maybe i would protect having two conditions with same type having different status?

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
```diff
@@ -184,3 +184,65 @@ func ValidateManagedFields(fieldsList []metav1.ManagedFieldsEntry, fldPath *fiel
...
+	}
+
+	return allErrs
+}
+
+// validConditionStatuses is used internally to check validity and provide a good message
+var validConditionStatuses = sets.NewString(string(metav1.ConditionTrue), string(metav1.ConditionFalse), string(metav1.ConditionUnknown))
+
+const (
+	maxReasonLen  = 16 * 1024
```
liggitt, line 209: this seems excessive for a "programmatic identifier"

the longest reasons I could find in-tree were:
```go
GroupVersionParsingFailed
OverlappingBuiltInResource
InstanceDeletionCompleted
UnknownContainerStatuses
```

something like 64 or 128 characters seems like plenty to me

liggitt (commented):
a couple nits, looks reasonable otherwise

I'm unsure about adding kube-builder annotations in types defined in `k8s.io/kubernetes`

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
deads2k: > something like 64 or 128 characters seems like plenty to me

I've seen some longer ones in the hundreds.  Concating during unions as a for instance.  How about 4k?  I want a max value, but I don't want to be constraining to what I've seen already.

deads2k: > I've seen some longer ones in the hundreds. Concating during unions as a for instance. How about 4k? I want a max value, but I don't want to be constraining to what I've seen already.

reduced to 1k

deads2k:
> I'm unsure about adding kube-builder annotations in types defined in `k8s.io/kubernetes`

I think we can draw a reasonable line about adding these tags based on the generator being in kubernetes-sigs.  I would allow additional tags on these types to added as long as they were for generators in kubernetes or kubernetes-sigs.  I would decline PRs for tags for generators that are not sponsored by kubernetes sigs.

mfojtik:
@deads2k regarding to condition API, do we believe that the message in condition provide enough information for users to understand its context or give users clue about possible mitigations?

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
```diff
@@ -184,3 +184,65 @@ func ValidateManagedFields(fieldsList []metav1.ManagedFieldsEntry, fldPath *fiel
...
+	}
+
+	if condition.LastTransitionTime.IsZero() {
+		allErrs = append(allErrs, field.Required(fldPath.Child("lastTransitionTime"), "must be set"))
+	}
+
+	if len(condition.Reason) == 0 {
+		allErrs = append(allErrs, field.Required(fldPath.Child("reason"), "must be set"))
+	} else {
+		for _, currErr := range validation.IsCIdentifier(condition.Reason) {
```
liggitt, line 235: I think `CIdentifier` is reasonable here, but it wouldn't allow the comma-separated reason you described to me

deads2k:
> @deads2k regarding to condition API, do we believe that the message in condition provide enough information for users to understand its context or give users clue about possible mitigations?

Yes.  I think that's the purpose of the message. It's clarity will depend on the author

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
deads2k: > I think `CIdentifier` is reasonable here, but it wouldn't allow the concatenated/delimited reasons you described to me

Expanded for comma and colon which we've seen as delimiters.

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
```diff
@@ -184,3 +185,79 @@ func ValidateManagedFields(fieldsList []metav1.ManagedFieldsEntry, fldPath *fiel
...
+	return allErrs
+}
+
+const conditionReasonFmt string = "[A-Za-z][A-Za-z0-9_,:]*"
+const conditionReasonErrMsg string = "a condition reason must start with alphabetic character, followed by a string of alphanumeric characters or '_,:'"
+
+var conditionReasonRegexp = regexp.MustCompile("^" + conditionReasonFmt + "$")
+
+// IsCIdentifier tests for a string that conforms the definition of an identifier
+// in C. This checks the format, but not the length.
```
liggitt, line 257: doc is out of date

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
```diff
@@ -184,3 +185,79 @@ func ValidateManagedFields(fieldsList []metav1.ManagedFieldsEntry, fldPath *fiel
...
+	}
+
+	if len(condition.Message) > maxMessageLen {
+		allErrs = append(allErrs, field.TooLong(fldPath.Child("message"), condition.Message, maxMessageLen))
+	}
+
+	return allErrs
+}
+
+const conditionReasonFmt string = "[A-Za-z][A-Za-z0-9_,:]*"
```
liggitt, line 251: should we make sure it ends in a non-delimiter?

```
"[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?"
```

deads2k: > should we make sure it ends in a non-delimiter?

Sure.

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
```diff
@@ -184,3 +185,78 @@ func ValidateManagedFields(fieldsList []metav1.ManagedFieldsEntry, fldPath *fiel
...
+
+	if len(condition.Message) > maxMessageLen {
+		allErrs = append(allErrs, field.TooLong(fldPath.Child("message"), condition.Message, maxMessageLen))
+	}
+
+	return allErrs
+}
+
+const conditionReasonFmt string = "[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?"
+const conditionReasonErrMsg string = "a condition reason must start with alphabetic character, followed by a string of alphanumeric characters or '_,:', and must not end with an alphanumeric character or '_'"
```
liggitt, line 252:
```suggestion
const conditionReasonErrMsg string = "a condition reason must start with alphabetic character, followed by a string of alphanumeric characters or '_,:', and must end with an alphanumeric character or '_'"
```

liggitt:
/approve

nit on message, then lgtm

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
```diff
@@ -184,3 +185,78 @@ func ValidateManagedFields(fieldsList []metav1.ManagedFieldsEntry, fldPath *fiel
...
+	// status is set and is an accepted value
+	if !validConditionStatuses.Has(string(condition.Status)) {
+		allErrs = append(allErrs, field.NotSupported(fldPath.Child("status"), condition.Status, validConditionStatuses.List()))
+	}
+
+	if condition.ObservedGeneration < 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("observedGeneration"), condition.ObservedGeneration, "must be greater than or equal to zero"))
+	}
+
+	if condition.LastTransitionTime.IsZero() {
```
sttts, line 229: does this match https://github.com/kubernetes/kubernetes/pull/92660/files#diff-eca3b8d856fa2e661f6da91b61de5e76R1391 ? I think the OpenAPI schema will allow a zero date (not null).

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
```diff
@@ -184,3 +185,78 @@ func ValidateManagedFields(fieldsList []metav1.ManagedFieldsEntry, fldPath *fiel
...
+
+	if len(condition.Message) > maxMessageLen {
+		allErrs = append(allErrs, field.TooLong(fldPath.Child("message"), condition.Message, maxMessageLen))
+	}
+
+	return allErrs
+}
+
+const conditionReasonFmt string = "[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?"
+const conditionReasonErrMsg string = "a condition reason must start with alphabetic character, followed by a string of alphanumeric characters or '_,:', and must end with an alphanumeric character or '_'"
```
sttts, line 252: this doesn't match the regexp. Some "optionally" are missing.

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
deads2k: > does this match https://github.com/kubernetes/kubernetes/pull/92660/files#diff-eca3b8d856fa2e661f6da91b61de5e76R1391 ? I think the OpenAPI schema will allow a zero date (not null).

Ok.  We'll control the ones we can then.  Not allowing zero is correct.

sttts:
Lgtm, with exception of https://github.com/kubernetes/kubernetes/pull/92519/files#r449081321.

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
deads2k: > this doesn't match the regexp. Some "optionally" are missing.

I need help seeing it.  Which optionally do I need to add?

liggitt:
```suggestion
const conditionReasonErrMsg string = "a condition reason must start with alphabetic character, optionally followed by a string of alphanumeric characters or '_,:', and must end with an alphanumeric character or '_'"
```

deads2k:
tagging per comments.

PR #83551, opened 2019-10-06
Move external facing kubelet apis to staging

dims:
**What type of PR is this?**
/kind api-change
/area code-organization

**What this PR does / why we need it**:
`deviceplugin` and `pluginregistration` are external facing APIs. currently we are forcing folks to import them from `k8s.io/kubernetes` which is not optimal. We should instead make them available from the `k8s.io/kubelet` repo to make it easier for them.

**Which issue(s) this PR fixes**:

Related to https://github.com/kubernetes/kubernetes/issues/82437

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
external facing APIs in pluginregistration and deviceplugin packages are now available under k8s.io/kubelet/pkg/apis/
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #92527, opened 2020-06-25
Move pkg/kubelet/apis to k8s.io/kubelet/pkg/apis

michaelbeaumont:
**What type of PR is this?**
/kind api-change
/area code-organization

At the moment, in order to programmatically validate which labels `kubelet` will accept if provided through `--node-labels`, the user needs to import from `k8s.io/kubernetes`, which is not intended for consumption as a module.

```release-note
APIs for kubelet annotations and labels from k8s.io/kubernetes/pkg/kubelet/apis are now available under k8s.io/kubelet/pkg/apis/
```

```docs

```


SergeyKanzhelev:
this PR only shows two files were moved under staging. Is it a problem with visualization and you intended to move all apis or just these two files?

image

michaelbeaumont:
For it's only important that the root `apis` package with the 2 files is moved. I'm indifferent to the rest.

SergeyKanzhelev:
so this PR pairs out nicely with this one: https://github.com/kubernetes/kubernetes/pull/92632/

Would it be better just to move all APIs then? @dims looking at comment here (https://github.com/kubernetes/kubernetes/pull/83551#issuecomment-538796127) I wonder what was the discussion and why the outcome was to move a subset of APIs, not the whole folder?

RenaudWasTaken:
Here is what I see after https://github.com/kubernetes/kubernetes/pull/92632 is merged:
- `resourcemetrics` is alpha and might make sense to move
- `stats` is alpha and might make sense to move
- `podresources` is beta, is "leftover" code that will get moved in cm/podresources as a followup PR
- `config` which already has a place in the staging directory (k8s.io/kubelet/config)

I'm not familiar with the `config` architecture but it definitely seems like it's at two places intentionally. With the `pkg/kubelet/apis` looking like it handles default and schema validation.

Moving these apis might be something that could be done gradually and as request for beta graduation?

SergeyKanzhelev:
@RenaudWasTaken do you think it makes sense to version two files moved with this PR? All other APIs seems to be located in versioned folders.

liggitt:
> Would it be better just to move all APIs then? @dims looking at comment here (#83551 (comment)) I wonder what was the discussion and why the outcome was to move a subset of APIs, not the whole folder?

The goal was to move the API types needed for serializing/deserializing to location that can be referenced outside k8s.io/kubernetes, not to move all validation code used by cmd/kubelet

well_known_annotations_windows.go has a method that uses a feature gate, which makes k8s.io/kubelet depend on k8s.io/apiserver, which is not expected/desired

Validating just `--node-labels` seems insufficient if the goal is to validate a kubelet invocation. Wouldn't cmd/kubelet/app/options/options.go#ValidateKubeletFlags / pkg/kubelet/apis/config/validation/validation.go#ValidateKubeletConfiguration be needed instead?

michaelbeaumont:
It's only about validation of node labels given by users (some more background at https://github.com/weaveworks/eksctl/issues/2363) and the rest of the invocation we just assume to be valid.

PR #92528, opened 2020-06-25
Apiservier staticchecks tests

kensipe:
Signed-off-by: Ken Sipe <kensipe@gmail.com>

**What type of PR is this?**
/kind cleanup

fixes `apiserver` static checks for 
* S1039 unnecessary use of fmt.Sprintf
* S1021 var declaration
* SA2006 value of error not used
* S1019 use of make
* S1023 redundant returns

Part of #92402

the PR would be too large to cover all issues found by static check

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/fieldmanager_test.go
+++ b/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/fieldmanager_test.go
@@ -63,7 +63,6 @@ func (c *fakeObjectConvertor) Convert(in, out, context interface{}) error {
 		out, err = c.converter.Convert(typedValue, c.apiVersion)
 		return err
 	}
-	out = in
 	return nil
 }
 
--- a/staging/src/k8s.io/apiserver/pkg/util/shufflesharding/shufflesharding_test.go
+++ b/staging/src/k8s.io/apiserver/pkg/util/shufflesharding/shufflesharding_test.go
@@ -277,7 +277,7 @@ func TestDealer_DealIntoHand(t *testing.T) {
 		},
 		{
 			"size: 6 cap: 6 slice",
-			make([]int, 6, 6),
+			make([]int, 6),
 			6,
 		},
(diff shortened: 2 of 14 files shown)


staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/fieldmanager_test.go
```diff
@@ -63,7 +63,6 @@ func (c *fakeObjectConvertor) Convert(in, out, context interface{}) error {
 		out, err = c.converter.Convert(typedValue, c.apiVersion)
 		return err
 	}
-	out = in
```
lavalamp, line 66: @apelisse what was this line trying to do?

staging/src/k8s.io/apiserver/pkg/util/shufflesharding/shufflesharding_test.go
```diff
@@ -277,7 +276,7 @@ func TestDealer_DealIntoHand(t *testing.T) {
 		},
 		{
 			"size: 6 cap: 6 slice",
-			make([]int, 6, 6),
```
lavalamp, line 280: hm since this file is making assertions about the slice I think it'd be good to keep the cap set deliberately. Since without this it's not obvious that you won't get a larger cap than len.

lavalamp:
sorry for delay

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/fieldmanager_test.go
apelisse: I think this means: make the conversion a no-op, but that's needed if you want the output to be equal to the input.

kensipe:
thanks for the reviews @apelisse and @lavalamp .   Apologies for missing the last rounds (resulting in the PR closing).  I didn't see a way to reopen... so I recreated the PR #96954

PR #92531, opened 2020-06-26
kube-scheduler: allow deprecated options to be set with configfile

damemi:
**What type of PR is this?**
/kind bug

With `algorithmSource` removed from the v1beta1 ComponentConfig API, the only way to specify a `Policy` config is with the deprecated flags. And even though a user is still using the Policy API for their algorithm config, there may also be configuration options the user wishes to set with the beta CC API through a config file. With the current check in place, it means that the user cannot pass both a ComponentConfig file and a Policy config.

This PR applies deprecated flags related specifically to the `algorithmSource` Policy config, regardless of if a `config-file` is set. This enables a v1 Policy config to be used to configure the algorithm while also using the v1beta1 CC api to configure those beta features which are not available to set otherwise.

Fixes #

```release-note
Scheduler v1 Policy config or algorithm-provider settings can now be passed alongside v1beta1 ComponentConfig to aid transition from Policy to CC
```

```docs

```
/sig scheduling

### DIFF
--- a/cmd/kube-scheduler/app/options/deprecated.go
+++ b/cmd/kube-scheduler/app/options/deprecated.go
@@ -47,12 +47,12 @@ func (o *DeprecatedOptions) AddFlags(fs *pflag.FlagSet, cfg *kubeschedulerconfig
 		return
 	}
 
-	fs.StringVar(&o.AlgorithmProvider, "algorithm-provider", o.AlgorithmProvider, "DEPRECATED: the scheduling algorithm provider to use, one of: "+algorithmprovider.ListAlgorithmProviders())
-	fs.StringVar(&o.PolicyConfigFile, "policy-config-file", o.PolicyConfigFile, "DEPRECATED: file with scheduler policy configuration. This file is used if policy ConfigMap is not provided or --use-legacy-policy-config=true")
-	usage := fmt.Sprintf("DEPRECATED: name of the ConfigMap object that contains scheduler's policy configuration. It must exist in the system namespace before scheduler initialization if --use-legacy-policy-config=false. The config must be provided as the value of an element in 'Data' map with the key='%v'", kubeschedulerconfig.SchedulerPolicyConfigMapKey)
+	fs.StringVar(&o.AlgorithmProvider, "algorithm-provider", o.AlgorithmProvider, "DEPRECATED: the scheduling algorithm provider to use, this sets the default plugins for component config profiles. Choose one of: "+algorithmprovider.ListAlgorithmProviders())
+	fs.StringVar(&o.PolicyConfigFile, "policy-config-file", o.PolicyConfigFile, "DEPRECATED: file with scheduler policy configuration. This file is used if policy ConfigMap is not provided or --use-legacy-policy-config=true. Note: The scheduler will fail if this is combined with Plugin configs")
+	usage := fmt.Sprintf("DEPRECATED: name of the ConfigMap object that contains scheduler's policy configuration. It must exist in the system namespace before scheduler initialization if --use-legacy-policy-config=false. The config must be provided as the value of an element in 'Data' map with the key='%v'. Note: The scheduler will fail if this is combined with Plugin configs", kubeschedulerconfig.SchedulerPolicyConfigMapKey)
 	fs.StringVar(&o.PolicyConfigMapName, "policy-configmap", o.PolicyConfigMapName, usage)
-	fs.StringVar(&o.PolicyConfigMapNamespace, "policy-configmap-namespace", o.PolicyConfigMapNamespace, "DEPRECATED: the namespace where policy ConfigMap is located. The kube-system namespace will be used if this is not provided or is empty.")
-	fs.BoolVar(&o.UseLegacyPolicyConfig, "use-legacy-policy-config", o.UseLegacyPolicyConfig, "DEPRECATED: when set to true, scheduler will ignore policy ConfigMap and uses policy config file")
+	fs.StringVar(&o.PolicyConfigMapNamespace, "policy-configmap-namespace", o.PolicyConfigMapNamespace, "DEPRECATED: the namespace where policy ConfigMap is located. The kube-system namespace will be used if this is not provided or is empty. Note: The scheduler will fail if this is combined with Plugin configs")
+	fs.BoolVar(&o.UseLegacyPolicyConfig, "use-legacy-policy-config", o.UseLegacyPolicyConfig, "DEPRECATED: when set to true, scheduler will ignore policy ConfigMap and uses policy config file. Note: The scheduler will fail if this is combined with Plugin configs")
 
 	fs.BoolVar(&cfg.EnableProfiling, "profiling", cfg.EnableProfiling, "DEPRECATED: enable profiling via web interface host:port/debug/pprof/")
 	fs.BoolVar(&cfg.EnableContentionProfiling, "contention-profiling", cfg.EnableContentionProfiling, "DEPRECATED: enable lock contention profiling, if profiling is enabled")
@@ -87,16 +87,14 @@ func (o *DeprecatedOptions) Validate() []error {
 	return errs
 }
 
-// ApplyTo sets cfg.AlgorithmSource from flags passed on the command line in the following precedence order:
+// ApplyAlgorithmSourceTo sets cfg.AlgorithmSource from flags passed on the command line in the following precedence order:
 //
 // 1. --use-legacy-policy-config to use a policy file.
 // 2. --policy-configmap to use a policy config map value.
 // 3. --algorithm-provider to use a named algorithm provider.
-//
-// This function is only called when no config file is provided.
-func (o *DeprecatedOptions) ApplyTo(cfg *kubeschedulerconfig.KubeSchedulerConfiguration) error {
+func (o *DeprecatedOptions) ApplyAlgorithmSourceTo(cfg *kubeschedulerconfig.KubeSchedulerConfiguration) {
 	if o == nil {
-		return nil
+		return
 	}
 
 	switch {
@@ -122,10 +120,19 @@ func (o *DeprecatedOptions) ApplyTo(cfg *kubeschedulerconfig.KubeSchedulerConfig
 			Provider: &o.AlgorithmProvider,
 		}
 	}
+}
 
-	// Deprecated flags have an effect iff no config file was provided, in which
-	// case this function expects a default KubeSchedulerConfiguration instance,
-	// which has a single profile.
+// ApplyTo sets a default profile plugin config if no config file is specified
+// It also calls ApplyAlgorithmSourceTo to set Policy settings in AlgorithmSource, if applicable.
+// Deprecated flags have an effect iff no config file was provided, in which
+// case this function expects a default KubeSchedulerConfiguration instance,
+// which has a single profile.
+func (o *DeprecatedOptions) ApplyTo(cfg *kubeschedulerconfig.KubeSchedulerConfiguration) {
+	if o == nil {
+		return
+	}
+	// The following deprecated options affect the only existing profile that is
+	// added by default.
 	profile := &cfg.Profiles[0]
 	if len(o.SchedulerName) > 0 {
 		profile.SchedulerName = o.SchedulerName
@@ -136,6 +143,7 @@ func (o *DeprecatedOptions) ApplyTo(cfg *kubeschedulerconfig.KubeSchedulerConfig
 			HardPodAffinityWeight: o.HardPodAffinitySymmetricWeight,
 		},
 	}
+
 	profile.PluginConfig = append(profile.PluginConfig, plCfg)
-	return nil
+	o.ApplyAlgorithmSourceTo(cfg)
 }
--- a/cmd/kube-scheduler/app/options/options.go
+++ b/cmd/kube-scheduler/app/options/options.go
@@ -148,7 +148,13 @@ func newDefaultComponentConfig() (*kubeschedulerconfig.KubeSchedulerConfiguratio
 // Flags returns flags for a specific scheduler by section name
 func (o *Options) Flags() (nfs cliflag.NamedFlagSets) {
 	fs := nfs.FlagSet("misc")
-	fs.StringVar(&o.ConfigFile, "config", o.ConfigFile, "The path to the configuration file. Flags override values in this file.")
+	fs.StringVar(&o.ConfigFile, "config", o.ConfigFile, `The path to the configuration file. The following flags can overwrite fields in this file:
+  --address
+  --port
+  --use-legacy-policy-config
+  --policy-configmap
+  --policy-config-file
+  --algorithm-provider`)
 	fs.StringVar(&o.WriteConfigTo, "write-config-to", o.WriteConfigTo, "If set, write the configuration values to this file and exit.")
 	fs.StringVar(&o.Master, "master", o.Master, "The address of the Kubernetes API server (overrides any value in kubeconfig)")
 
@@ -170,10 +176,8 @@ func (o *Options) ApplyTo(c *schedulerappconfig.Config) error {
 	if len(o.ConfigFile) == 0 {
 		c.ComponentConfig = o.ComponentConfig
 
-		// only apply deprecated flags if no config file is loaded (this is the old behaviour).
-		if err := o.Deprecated.ApplyTo(&c.ComponentConfig); err != nil {
-			return err
-		}
+		// apply deprecated flags if no config file is loaded (this is the old behaviour).
+		o.Deprecated.ApplyTo(&c.ComponentConfig)
 		if err := o.CombinedInsecureServing.ApplyTo(c, &c.ComponentConfig); err != nil {
 			return err
 		}
@@ -186,11 +190,18 @@ func (o *Options) ApplyTo(c *schedulerappconfig.Config) error {
 			return err
 		}
 
-		// use the loaded config file only, with the exception of --address and --port. This means that
-		// none of the deprecated flags in o.Deprecated are taken into consideration. This is the old
-		// behaviour of the flags we have to keep.
 		c.ComponentConfig = *cfg
 
+		// apply any deprecated Policy flags, if applicable
+		o.Deprecated.ApplyAlgorithmSourceTo(&c.ComponentConfig)
+
+		// if the user has set CC profiles and is trying to use a Policy config, error out
+		// these configs are no longer merged and they should not be used simultaneously
+		if !emptySchedulerProfileConfig(c.ComponentConfig.Profiles) && c.ComponentConfig.AlgorithmSource.Policy != nil {
+			return fmt.Errorf("cannot set a Plugin config and Policy config")
+		}
+
+		// use the loaded config file only, with the exception of --address and --port.
 		if err := o.CombinedInsecureServing.ApplyToFromLoadedConfig(c, &c.ComponentConfig); err != nil {
 			return err
 		}
@@ -209,6 +220,15 @@ func (o *Options) ApplyTo(c *schedulerappconfig.Config) error {
 	}
 	o.Metrics.Apply()
 	return nil
+}
+
+// emptySchedulerProfileConfig returns true if the list of profiles passed to it contains only
+// the "default-scheduler" profile with no plugins or pluginconfigs registered
+// (this is the default empty profile initialized by defaults.go)
+func emptySchedulerProfileConfig(profiles []kubeschedulerconfig.KubeSchedulerProfile) bool {
+	return len(profiles) == 1 &&
+		len(profiles[0].PluginConfig) == 0 &&
+		profiles[0].Plugins == nil
 }
 
 // Validate validates all the required options.
--- a/cmd/kube-scheduler/app/options/options_test.go
+++ b/cmd/kube-scheduler/app/options/options_test.go
@@ -722,6 +722,16 @@ profiles:
 			},
 		},
 		{
+			name: "Attempting to set Component Config Profiles and Policy config",
+			options: &Options{
+				ConfigFile: pluginConfigFile,
+				Deprecated: &DeprecatedOptions{
+					PolicyConfigMapName: "bar",
+				},
+			},
+			expectedError: "cannot set a Plugin config and Policy config",
+		},
+		{
 			name: "unknown field",
 			options: &Options{
 				ConfigFile: unknownFieldConfig,


damemi:
/cc @ahg-g 
/priority important-soon
I believe without this the v1 API is broken, or at least unusable when alongside the beta CC options

ahg-g:
/assign @alculquicondor 

I think CC takes precedence over flags, but will defer to Aldo since he re-worked this part.

alculquicondor:
The problem is that Policies and Component Config could conflict and whatever you do to reconcile is confusing.

I prefer we only support Policy through command line flags.

damemi:
> I prefer we only support Policy through command line flags.

That's what this does, it just enables those flags to be set along with a config. And based on the `--config-file` CLI option description I think this is what we should do:
```
--config string                                                                                                                                                                                                                                                
                The path to the configuration file. Flags override values in this file.
```

This implies to me that the `--policy-configmap`, etc, flags would override CC. Which makes sense, given that CC is only in beta, but this allows users to still set the CC options that don't conflict and aids transition from Policy to CC.

alculquicondor:
The problem with such approach is that checking if a configuration is correct requires running the scheduler to incorporate the effects of flags.

In a recent discussion with @liggitt, the suggestion was rather to change the `--config` help string to explicitly tell which flags are still respected (it's only 2 today).

We made an effort to support all the uses of Policy in Component Config so that people can directly migrate. Given that we are now graduating to beta, I prefer we do not introduce more tech debt to the API.

damemi:
Without a change like this, it is currently impossible for users of the Policy API to configure any beta CC features (that don't have flags available) without also migrating their Policy files/configs to a Profile(Plugin) config that can be inserted into the CC. This makes it all-or-nothing: if you want to try any part of the beta API you need to completely migrate off the v1 API (and vice-versa).

If that is the intended behavior then we can clearly document it, but because Component Config covers more than just Plugins it doesn't seem unreasonable to want to use parts of both, keeping in mind that these are different APIs and not just different versions of the same API.

For that reasoning I didn't see this as tech debt but a necessary change to continue v1 support during transition until the v1 API is removed.

alculquicondor:
Could you clarify the feature that is needed that is not available through flags?

damemi:
It's possible that I just missed them, but I couldn't find flags for:
* HealthzBindAddress
* MetricsBindAddress
* DisablePreemption
* PercentageOfNodesToScore
* PodInitialBackoffSeconds
* PodMaxBackoffSeconds

Are any of these intended to be available along with the Policy API? Since they were available together in alpha, it seems weird that the promotion of these features to beta has made them less accessible to people still on the v1 API

ahg-g:
We can probably add those flags to `Policy` to facilitate easier transition.

damemi:
> We can probably the missing flags to `Policy` to facilitate easier transition.

This seems like another option, but won't that add more debt because we will have to consider these options from 2 sources everywhere they're used? It would also introduce more potential points of conflict

alculquicondor:
> This seems like another option, but won't that add more debt because we will have to consider these options from 2 sources everywhere they're used? It would also introduce more potential points of conflict

That's not so hard, because we just need to copy over the data from Policy into the internal component config after processing Policy. And since using Policy implies that Component Config is not used (and flag haven't affected those parameters), it is safe to do so.

ahg-g:
yeah, the copy happens in one function in factory.go which should go away when policy is deprecated: https://github.com/kubernetes/kubernetes/blob/c90427d42e8fe9ef37b61927afd2e12697c2d8a5/pkg/scheduler/factory.go#L229:24

I think this is the cleanest approach.

damemi:
If we want to go with that approach, I'm fine with that, but it seems like extra work for the same result. And since Policy is mostly just equivalent to the Plugins config (which the scheduler only considers one or the other), what could conflict with Component Config? I suppose Extenders would be the only thing, right?

Will we also need to consider that adding these settings to the v1 api binds them to it, meaning we can't change or remove them until Policy is removed? What if we add Policy as a section to the v1beta1 api?

alculquicondor:
> And since Policy is mostly just equivalent to the Plugins config (which the scheduler only considers one or the other), what could conflict with Component Config? I suppose Extenders would be the only thing, right?

The plugins themselves :) Policy translates to a set of plugins and then component config modifies them. It gets tricky to do such intersection in a way that it's clear to users. And it gets worst if you consider multiple scheduling profiles.

alculquicondor:
> Will we also need to consider that adding these settings to the v1 api binds them to it, meaning we can't change or remove them until Policy is removed?

That's fine, we don't expect to remove them from Component Config either.

> What if we add Policy as a section to the v1beta1 api?

That has the same problems as stated in my previous comment.

damemi:
> > And since Policy is mostly just equivalent to the Plugins config (which the scheduler only considers one or the other), what could conflict with Component Config? I suppose Extenders would be the only thing, right?
> 
> The plugins themselves :) Policy translates to a set of plugins and then component config modifies them. It gets tricky to do such intersection in a way that it's clear to users. And it gets worst if you consider multiple scheduling profiles.

Ah, so is this only a problem if the user had configured both a Policy and Plugins, or do some of the other Component Config options modify the plugins as well?

alculquicondor:
> Ah, so is this only a problem if the user had configured both a Policy and Plugins, or do some of the other Component Config options modify the plugins as well?

Correct, the problem is for Policy and Plugins being set together. But the point is that semantics are confusing regardless.

damemi:
> > Ah, so is this only a problem if the user had configured both a Policy and Plugins, or do some of the other Component Config options modify the plugins as well?
> 
> Correct, the problem is for Policy and Plugins being set together. But the point is that semantics are confusing regardless.

Would another feasible option then be to change this check from if `len(o.ConfigFile) == 0` to `if len(c.ComponentConfig.Profiles) == 0`? This would be the same functionality as now, where policy config flags are ignored if a config file is set, except it looks specifically at the section that conflicts with the Policy config

alculquicondor:
> Would another feasible option then be to change this check from if `len(o.ConfigFile) == 0` to `if len(c.ComponentConfig.Profiles) == 0`? This would be the same functionality as now, where policy config flags are ignored if a config file is set, except it looks specifically at the section that conflicts with the Policy config

That's more plausible. But complicates the semantics of `--policy-config-file`. @ahg-g ?

ahg-g:
I am fine with that option if both of you think it is the option with the best tradeoff.

alculquicondor:
@damemi Would you mind crafting a PR that does the above?

alculquicondor:
Ok, let's do this:

1. preserve the basic structure of this PR: allow policy flag when CC is used.
1. fail scheduler startup if both policy and profiles are set
1. Document in the `--config` help the list of parameters that could override CC (3 with this PR, IIRC)

damemi:
> Ok, let's do this:
> 
>     1. preserve the basic structure of this PR: allow policy flag when CC is used.
> 
>     2. fail scheduler startup if both policy and profiles are set
> 
>     3. Document in the `--config` help the list of parameters that could override CC (3 with this PR, IIRC)

This sounds good to me, I'll update the PR with these changes

damemi:
Thank you @ahg-g and @alculquicondor for all of your feedback and help with this problem. Pushed an update to this PR that:
1. returns an error if Profiles are set with deprecated Policy flags
2. adds a test case for this
3. updates the help output for `--config`:
```
--config string           
            The path to the configuration file. The following flags can overwrite fields in this file:
             --use-legacy-policy-config
             --policy-configmap
             --algorithm-provider
```

I haven't manually tested these new changes yet (which I will do soon), so I'll hold until then. Feel free to review the new changes/tag if it looks good though
/hold

damemi:
I had to change how we consider if users have set a profile, because just going by `len(c.Profiles) == 0` doesn't work since [defaults.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/apis/config/v1beta1/defaults.go#L46-L49) initializes an empty profile if none are provided.

I caught that testing it manually, and can say that now it definitely returns an error if Policy and Profiles are both set. I also confirmed that Policy still works with a v1beta1 CC that *doesn't* have any Profiles configured. Unfortunately, my cluster died on me before I could verify that a CC with Profiles configured (and no Policy) still works, but I think CI should be sufficient to test that.

/retest
/hold cancel

ingvagabund (commented):
Just nits

cmd/kube-scheduler/app/options/deprecated.go
```diff
@@ -87,14 +87,20 @@ func (o *DeprecatedOptions) Validate() []error {
...
+
+// ApplyAlgorithmSourceTo sets cfg.AlgorithmSource from flags passed on the command line in the following precedence order:
 //
 // 1. --use-legacy-policy-config to use a policy file.
 // 2. --policy-configmap to use a policy config map value.
 // 3. --algorithm-provider to use a named algorithm provider.
-//
-// This function is only called when no config file is provided.
-func (o *DeprecatedOptions) ApplyTo(cfg *kubeschedulerconfig.KubeSchedulerConfiguration) error {
+func (o *DeprecatedOptions) ApplyAlgorithmSourceTo(cfg *kubeschedulerconfig.KubeSchedulerConfiguration) error {
```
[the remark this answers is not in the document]
damemi: @ingvagabund good catch, updated

cmd/kube-scheduler/app/options/deprecated.go
```diff
@@ -123,9 +129,19 @@ func (o *DeprecatedOptions) ApplyTo(cfg *kubeschedulerconfig.KubeSchedulerConfig
...
-	// case this function expects a default KubeSchedulerConfiguration instance,
-	// which has a single profile.
+	return nil
+}
+
+// ApplyProfileTo sets a default profile plugin config if no config file is specified
+// Deprecated flags have an effect iff no config file was provided, in which
+// case this function expects a default KubeSchedulerConfiguration instance,
+// which has a single profile.
+func (o *DeprecatedOptions) ApplyProfileTo(cfg *kubeschedulerconfig.KubeSchedulerConfiguration) error {
```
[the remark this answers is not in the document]
damemi: @ingvagabund good catch, updated

cmd/kube-scheduler/app/options/options.go
```diff
@@ -169,11 +169,9 @@ func (o *Options) Flags() (nfs cliflag.NamedFlagSets) {
 func (o *Options) ApplyTo(c *schedulerappconfig.Config) error {
 	if len(o.ConfigFile) == 0 {
 		c.ComponentConfig = o.ComponentConfig
+		o.Deprecated.ApplyProfileTo(&c.ComponentConfig)
 
 		// only apply deprecated flags if no config file is loaded (this is the old behaviour).
```
alculquicondor, line 174: Still call the function after the comment, but remove the word "only" from it.

cmd/kube-scheduler/app/options/options.go
```diff
@@ -211,6 +214,16 @@ func (o *Options) ApplyTo(c *schedulerappconfig.Config) error {
...
 }
 
+// emptySchedulerProfileConfig returns true if the list of profiles passed to it contains only
+// the "default-scheduler" profile with no plugins or pluginconfigs registered
+// (this is the default empty profile initialized by defaults.go)
+func emptySchedulerProfileConfig(profiles []kubeschedulerconfig.KubeSchedulerProfile) bool {
+	return len(profiles) == 1 &&
+		len(profiles[0].PluginConfig) == 0 &&
+		profiles[0].Plugins == nil &&
+		profiles[0].SchedulerName == "default-scheduler"
```
alculquicondor, line 224: You should still be able to set a schedulerName

cmd/kube-scheduler/app/options/options.go
```diff
@@ -186,15 +184,20 @@ func (o *Options) ApplyTo(c *schedulerappconfig.Config) error {
 			return err
 		}
 
-		// use the loaded config file only, with the exception of --address and --port. This means that
-		// none of the deprecated flags in o.Deprecated are taken into consideration. This is the old
-		// behaviour of the flags we have to keep.
+		// use the loaded config file only, with the exception of --address and --port.
 		c.ComponentConfig = *cfg
 
+		// if the user has set CC profiles and is trying to use a Policy config, error out
```
alculquicondor, line 190: You could call ApplyAlgorithmSourceTo before this and check that .AlgorithmSource.Policy is empty

cmd/kube-scheduler/app/options/deprecated.go
```diff
@@ -87,16 +87,22 @@ func (o *DeprecatedOptions) Validate() []error {
...
+
+// ApplyAlgorithmSourceTo sets cfg.AlgorithmSource from flags passed on the command line in the following precedence order:
 //
 // 1. --use-legacy-policy-config to use a policy file.
 // 2. --policy-configmap to use a policy config map value.
 // 3. --algorithm-provider to use a named algorithm provider.
-//
-// This function is only called when no config file is provided.
-func (o *DeprecatedOptions) ApplyTo(cfg *kubeschedulerconfig.KubeSchedulerConfiguration) error {
+func (o *DeprecatedOptions) ApplyAlgorithmSourceTo(cfg *kubeschedulerconfig.KubeSchedulerConfiguration) {
```
alculquicondor, line 103: I think I prefer we extract `ApplyAlgorithmSource`, but leave an `ApplyTo` function which calls it.

cmd/kube-scheduler/app/options/deprecated.go
```diff
@@ -87,16 +87,22 @@ func (o *DeprecatedOptions) Validate() []error {
 	return errs
 }
 
-// ApplyTo sets cfg.AlgorithmSource from flags passed on the command line in the following precedence order:
+// PolicyConfig returns whether a legacy policy config has been set
```
alculquicondor, line 90: Document the usage of the policy-related command line flags.

cmd/kube-scheduler/app/options/options.go
```diff
@@ -148,7 +148,7 @@ func newDefaultComponentConfig() (*kubeschedulerconfig.KubeSchedulerConfiguratio
 // Flags returns flags for a specific scheduler by section name
 func (o *Options) Flags() (nfs cliflag.NamedFlagSets) {
 	fs := nfs.FlagSet("misc")
-	fs.StringVar(&o.ConfigFile, "config", o.ConfigFile, "The path to the configuration file. Flags override values in this file.")
+	fs.StringVar(&o.ConfigFile, "config", o.ConfigFile, "The path to the configuration file. The following flags can overwrite fields in this file:\n --use-legacy-policy-config\n --policy-configmap\n --algorithm-provider")
```
alculquicondor, line 151: Don't forget `--adress` and `--port`

cmd/kube-scheduler/app/options/options.go
damemi: done

cmd/kube-scheduler/app/options/options.go
damemi: removed the check for `profiles[0].SchedulerName == "default-scheduler"`

cmd/kube-scheduler/app/options/options.go
damemi: done

also note, that we can't check if `algorithmSource.Provider` is empty because when any config is provided the default algorithmProvider value is [set by conversion to the internal type](https://github.com/kubernetes/kubernetes/blob/ec36ff40016e1ea612ae364b3b550cd31dd04651/pkg/scheduler/apis/config/v1beta1/conversion.go#L53). Not sure if this matters, but I was checking against that before too

cmd/kube-scheduler/app/options/deprecated.go
damemi: done

cmd/kube-scheduler/app/options/deprecated.go
damemi: we actually don't need this function when I made the changes from https://github.com/kubernetes/kubernetes/pull/92531#discussion_r447075338 and https://github.com/kubernetes/kubernetes/pull/92531#discussion_r447076538

cmd/kube-scheduler/app/options/options.go
damemi: added

damemi:
@alculquicondor updated with your feedback (and did another round of some manual testing), please take another look

cmd/kube-scheduler/app/options/options.go
```diff
@@ -148,7 +148,7 @@ func newDefaultComponentConfig() (*kubeschedulerconfig.KubeSchedulerConfiguratio
 // Flags returns flags for a specific scheduler by section name
 func (o *Options) Flags() (nfs cliflag.NamedFlagSets) {
 	fs := nfs.FlagSet("misc")
-	fs.StringVar(&o.ConfigFile, "config", o.ConfigFile, "The path to the configuration file. Flags override values in this file.")
+	fs.StringVar(&o.ConfigFile, "config", o.ConfigFile, "The path to the configuration file. The following flags can overwrite fields in this file:\n --address\n --port\n --use-legacy-policy-config\n --policy-configmap\n --algorithm-provider")
```
alculquicondor, line 151: you missed `--policy-config-file`

cmd/kube-scheduler/app/options/options.go
```diff
@@ -186,11 +184,18 @@ func (o *Options) ApplyTo(c *schedulerappconfig.Config) error {
 			return err
 		}
 
-		// use the loaded config file only, with the exception of --address and --port. This means that
-		// none of the deprecated flags in o.Deprecated are taken into consideration. This is the old
-		// behaviour of the flags we have to keep.
+		// use the loaded config file only, with the exception of --address and --port.
```
alculquicondor, line 187: This comment refers to `o.CombinedInsecureServing.ApplyToFrmLoadedConfig` below

cmd/kube-scheduler/app/options/deprecated.go
```diff
@@ -87,16 +87,14 @@ func (o *DeprecatedOptions) Validate() []error {
 	return errs
```
alculquicondor, line 87: Document in the applicable help strings above that scheduler would fail if the flags are used in combination with scheduling plugins.

alculquicondor (commented):
You are also allowing `--algorithm-provider` to be used alongside CC. Either document that in the release note or revert it.

cmd/kube-scheduler/app/options/options.go
damemi: moved

cmd/kube-scheduler/app/options/deprecated.go
damemi: done

damemi:
> You are also allowing --algorithm-provider to be used alongside CC. Either document that in the release note or revert it.

@alculquicondor yeah that was the problem I ran into when just switching the check to see if AlgorithmSource was set (https://github.com/kubernetes/kubernetes/pull/92531#discussion_r447176624). If any component config is provided (even if it doesn't set plugins), the conversion to internal sets a default algorithm provider. So adding this as a check (ie, `if algorithmSource.Policy != nil || algorithmSource.Provider != nil`) fails when trying to only provide a component config.

Are there any adverse effects you can think of by allowing algorithmProvider and CC? It seems like algorithmProvider takes precedence over CC so we could document that if it isn't already

alculquicondor:
> Are there any adverse effects you can think of by allowing algorithmProvider and CC? It seems like algorithmProvider takes precedence over CC so we could document that if it isn't already

Yes, that should work fine. Please document the semantics in `--algorithm-provider` help string.

damemi:
> > Are there any adverse effects you can think of by allowing algorithmProvider and CC? It seems like algorithmProvider takes precedence over CC so we could document that if it isn't already
> 
> Yes, that should work fine. Please document the semantics in `--algorithm-provider` help string.

Added a note in the help string that this overrides CC profiles

cmd/kube-scheduler/app/options/deprecated.go
```diff
@@ -47,12 +47,12 @@ func (o *DeprecatedOptions) AddFlags(fs *pflag.FlagSet, cfg *kubeschedulerconfig
 		return
 	}
 
-	fs.StringVar(&o.AlgorithmProvider, "algorithm-provider", o.AlgorithmProvider, "DEPRECATED: the scheduling algorithm provider to use, one of: "+algorithmprovider.ListAlgorithmProviders())
-	fs.StringVar(&o.PolicyConfigFile, "policy-config-file", o.PolicyConfigFile, "DEPRECATED: file with scheduler policy configuration. This file is used if policy ConfigMap is not provided or --use-legacy-policy-config=true")
-	usage := fmt.Sprintf("DEPRECATED: name of the ConfigMap object that contains scheduler's policy configuration. It must exist in the system namespace before scheduler initialization if --use-legacy-policy-config=false. The config must be provided as the value of an element in 'Data' map with the key='%v'", kubeschedulerconfig.SchedulerPolicyConfigMapKey)
+	fs.StringVar(&o.AlgorithmProvider, "algorithm-provider", o.AlgorithmProvider, "DEPRECATED: the scheduling algorithm provider to use, this overrides component config profiles. Choose one of: "+algorithmprovider.ListAlgorithmProviders())
```
alculquicondor, line 50: It doesn't override them. It sets the default plugins. So you can still enable/disable plugins from them.

alculquicondor:
/approve

@ahg-g anything to add?

alculquicondor:
oh, don't forget to update the release note with the algorithm provider detail

ahg-g:
/hold

if you still need to update the release notes

damemi:
Thanks, forgot to update the release note. Updated
/hold cancel

changshuchao:
@damemi  i think code below in cmd/kube-scheduler/app/options/options.go is not accurate. 
	```
fs.StringVar(&o.ConfigFile, "config", o.ConfigFile, `The path to the configuration file. The following flags can overwrite fields in this file:
  --address
  --port
  --use-legacy-policy-config
  --policy-configmap
  --policy-config-file
  --algorithm-provider`)
```

alculquicondor:
What is the problem exactly?

changshuchao:
@alculquicondor，After my test, when the --config parameter is set, the following parameters are invalid in the command line: --port, --address, --profiling, --contention-profiling, --kubeconfig,- -kube-api-content-type, --kube-api-qps, --kube-api-burst, --lock-object-namespace, --lock-object-name, --use-legacy-policy-config , --Hard-pod-affinity-symmetric-weight, --leader-elect, --leader-elect-lease-duration, --leader-elect-renew-deadline, --leader-elect-resource-lock,- -leader-elect-resource-name, --leader-elect-resource-namespace, --leader-elect-retry-period。

alculquicondor:
could you clarify if by "invalid" you mean the scheduler errors or if the values are just ignored?

Some of the arguments should be ignored. But you are right, the help line is inaccurate. Could you send a PR to fix?

changshuchao:
@alculquicondor ，yes，it's just not effective. ok，i will send a PR to fix. :)

changshuchao:
@alculquicondor , hi, i send a PR for this question. would you mind review his for me. thanks!  
https://github.com/kubernetes/kubernetes/pull/98211

PR #92532, opened 2020-06-26
Create checkServerVersion Test - 1 Endpoint Coverage

### LINKED ISSUE #92194: Write checkServerVersion Test - 1 Endpoint Coverage
# Progress <code>[6/6]</code>

-   [X] APISnoop org-flow: [checkServerVersionTest.org](https://github.com/cncf/apisnoop/blob/master/tickets/k8s/checkServerVersionTest.org)
-   [X] Test approval issue: [kubernetes/kubernetes#92194](https://github.com/kubernetes/kubernetes/issues/92194)
-   [X] Test pr: kuberenetes/kubernetes [#92532](https://github.com/kubernetes/kubernetes/pull/92532)
-   [x] Two weeks soak start date: [testgrid-link](https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-default&width=5&include-filter-by-regex=should.find.the.server.version) 3 July 2020
-   [x] Two weeks soak end date: 20 July 2020
-   [X] Test promotion pr: kubernetes/kubernetes#[93084](https://github.com/kubernetes/kubernetes/pull/93084)

# Identifying an untested feature Using APISnoop

According to this APIsnoop query, there are still an endpoint which is untested.

```sql-mode
SELECT
  operation_id,
  k8s_action,
  http_method,
  version,
  path,
  description
  -- kind
  -- FROM untested_stable_core_endpoints
  FROM untested_stable_endpoints
  where path like '/version/'
  -- and kind like ''
  -- where description ilike 'get the code version'
  -- and operation_id ilike '%APIGroup'
 ORDER BY operation_id
 LIMIT 25
       ;
```

```example
  operation_id  | k8s_action | http_method | version |   path    |     description      
----------------+------------+-------------+---------+-----------+----------------------
 getCodeVersion |            | get         |         | /version/ | get the code version
(1 row)

```

# API Reference and feature documentation

-   [Kubernetes API Reference Docs](https://kubernetes.io/docs/reference/kubernetes-api/)
-   [Kubernetes API: v1.18](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#)
-   [client-go](https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/)

# The mock test

## Test outline

1.  Get the server version (hits api path /version)

2.  Check that the major version is current

## Test the functionality in Go

```go
package main

import (
  "flag"
  "fmt"
  "k8s.io/apimachinery/pkg/version"
  "k8s.io/client-go/kubernetes"
  "k8s.io/client-go/tools/clientcmd"
  "os"
)

func main() {
  // uses the current context in kubeconfig
  kubeconfig := flag.String("kubeconfig", fmt.Sprintf("%v/%v/%v", os.Getenv("HOME"), ".kube", "config"), "(optional) absolute path to the kubeconfig file")
  flag.Parse()
  config, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
  if err != nil {
    fmt.Println(err)
    return
  }
  // make our work easier to find in the audit_event queries
  config.UserAgent = "live-test-writing"
  // creates the clientset
  ClientSet, _ := kubernetes.NewForConfig(config)

  // TEST BEGINS HERE
  fmt.Println("[status] begin")

  var version *version.Info
  version, err = ClientSet.Discovery().ServerVersion()

  if version.Major != "1" {
    errMsg := "Failed to get major version, got: " + version.Major + "\n"
    os.Stderr.WriteString(errMsg)
    os.Exit(1)
  }
  fmt.Println("Major version:", version.Major)

  // TEST ENDS HERE
  fmt.Println("[status] complete")
}
```

```go
[status] begin
Major version: 1
[status] complete
```

# Verifying increase in coverage with APISnoop

Discover useragents:

```sql-mode
select distinct useragent from audit_event where bucket='apisnoop' and useragent not like 'kube%' and useragent not like 'coredns%' and useragent not like 'kindnetd%' and useragent like 'live%';
```

```example
     useragent     
-------------------
 live-test-writing
(1 row)

```

List endpoints hit by the test:

```sql-mode
select * from endpoints_hit_by_new_test where useragent like 'live%';
```

```example
     useragent     |  operation_id  | hit_by_ete | hit_by_new_test 
-------------------+----------------+------------+-----------------
 live-test-writing | getCodeVersion | f          |               1
(1 row)

```

Display endpoint coverage change:

```sql-mode
select * from projected_change_in_coverage;
```

```example
   category    | total_endpoints | old_coverage | new_coverage | change_in_number 
---------------+-----------------+--------------+--------------+------------------
 test_coverage |             458 |          206 |          207 |                1
(1 row)

```

# Final notes

If a test with these calls gets merged, ****test coverage will go up by 1 points****

This test is also created with the goal of conformance promotion.

---

/sig testing
/sig architecture
/area conformance

riaankleinhans:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it:**
This PR adds a test to test the following untested endpoints:
getCodeVersion

**Which issue(s) this PR fixes:**
Fixes #92194

**Special notes for your reviewer:**
Adds +1 endpoint test coverage (good for conformance)

**Does this PR introduce a user-facing change?:**
NONE

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:**
NONE

/sig testing
/sig architecture
/area conformance


test/e2e/apimachinery/get_code_version.go
```diff
@@ -0,0 +1,44 @@
...
+		version, err := f.ClientSet.Discovery().ServerVersion()
+		framework.ExpectNoError(err, "Fail to access ServerVersion")
+
+		ginkgo.By("Confirm major version")
+		framework.ExpectEqual(version.Major, "1", "unable to find major version")
+		framework.Logf("Major version: %v", version.Major)
+
+		ginkgo.By("Confirm minor version")
+		framework.ExpectEqual(version.Minor, "19+", "unable to find minor version")
+		framework.Logf("Minor version: %v", version.Minor)
```
liggitt, line 42: The hard-coded checks for specific major/minor versions will fail and require changing multiple times every release. Checking the endpoint returns well-structured data, and that the returned values can be parsed would be more reasonable

hh: I would be fine with just ensuring that **version.Major** and **version.Minor** are what you expect.

heyste: Switched to using regexp to check the server version

test/e2e/apimachinery/get_code_version.go
```diff
@@ -0,0 +1,52 @@
+/*
```
oomichi, line 1: How about server_version.go instead of get_code_version.go as the test purpose.

test/e2e/apimachinery/get_code_version.go
```diff
@@ -0,0 +1,52 @@
...
+Copyright 2019 The Kubernetes Authors.
```
oomichi, line 2: nit: s/2019/2020/

oomichi (commented):
https://github.com/kubernetes/kubernetes/blob/209117413f32ec64e800dc32f69055ef2af75d54/test/e2e/network/service.go#L3667 touches the API already, but the existing one doesn't seem main purpose on the test.
It is fine to add separated test like this.
Most part looks good for me, just few comments.

test/e2e/apimachinery/get_code_version.go
Jefftree: +1

test/e2e/apimachinery/get_code_version.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/version"
+	"k8s.io/kubernetes/test/e2e/framework"
+	"regexp"
+
+	"github.com/onsi/ginkgo"
+)
+
+var _ = SIGDescribe("get-code-version", func() {
```
Jefftree, line 27: Similar to the above comment, prefer server version over code version.

test/e2e/apimachinery/get_code_version.go
```diff
@@ -0,0 +1,52 @@
...
+		framework.ExpectNoError(err, "Fail to access ServerVersion")
+
+		ginkgo.By("Confirm major version")
+		re := regexp.MustCompile("[1-9]")
+		framework.ExpectEqual(re.FindString(version.Major), version.Major, "unable to find major version")
+		framework.Logf("Major version: %v", version.Major)
+
+		ginkgo.By("Confirm minor version")
+
+		re = regexp.MustCompile("[^0-9]+")
```
Jefftree, line 44: I'm a bit confused here, is it not possible to encapsulate the minor version in one regex? Also, with the way it is currently written, seems to be just checking if the minor version contains a number.

test/e2e/apimachinery/get_server_version.go
```diff
@@ -0,0 +1,52 @@
+/*
```
oomichi, line 1: Under `test/e2e/` e2e test filenames don't contain verb.
It is nice to rename this to just `server_version.go` without `get_`.

heyste: Thanks for the catch. Fixed now

test/e2e/apimachinery/get_code_version.go
heyste: Fixed now. The original name was based on the `operational_id` from [APIsnoop](https://apisnoop.cncf.io)

oomichi:
Thanks for updating.

/retest
/lgtm

PR #92533, opened 2020-06-26
Stop enabling alpha runtimeclass API

liggitt:
**What type of PR is this?**
/kind cleanup

Stops enabling the alpha runtimeclass API in versions that have the beta API available

```release-note
NONE
```

/cc @tallclair


BenTheElder:
/retest
integration flake looks fun

PR #92534, opened 2020-06-26
[WIP]Created verify-dockerless file

yash97:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
 /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/hack/verify-dockerless.sh
+++ b/hack/verify-dockerless.sh
@@ -1,0 +1,15 @@
+not_have_dockerless=$(find ../pkg/kubelet/dockershim -name "*.go" ! -exec grep -q 'dockerless' {} \; -print)
+if [[ $not_have_dockerless ]]; then
+        echo "Following files do not contain dockerless tag"
+        echo $not_have_dockerless
+fi
+imports_docker=$(find ../{cmd,pkg}/kubelet/ -name "*.go" -exec grep -q 'github.com/docker/docker' {} \; -print | grep -v ../pkg/kubelet/dockershim)
+if [[ $imports_docker ]]; then
+        echo "Following files imports docker outside dockershim"
+        echo $imports_docker
+fi
+imports_dockershim=$(find ../{cmd,pkg}/kubelet/ -name "*.go" -exec grep -q 'kubelet/dockershim' {} \; -print | grep -v dockershim)
+if [[ $imports_dockershim ]]; then
+        echo "Following files imports dockershim outside dockershim folder and kubelet_dockershim.go"
+        echo $imports_dockershim
+fi


yash97:
@mattjmcnaughton  Created Pull request.

hack/verify-dockerless.sh
```diff
@@ -0,0 +1,15 @@
+not_have_dockerless=$(find ../pkg/kubelet/dockershim -name "*.go" ! -exec grep -q 'dockerless' {} \; -print)
```
mattjmcnaughton, line 1: If you look at the scripts in `hack/` you can see that they all share some common boilerplate (specifically lines 1-23 in this [file](https://sourcegraph.com/github.com/kubernetes/kubernetes@master/-/blob/hack/verify-typecheck-dockerless.sh)). Please add that boilerplate here.

hack/verify-dockerless.sh
```diff
@@ -0,0 +1,15 @@
...
+if [[ $not_have_dockerless ]]; then
+        echo "Following files do not contain dockerless tag"
```
mattjmcnaughton, line 3: Since these checks are running as part of CI, manually outputting "violations" isn't sufficient. We also want the script to fail.

Specifically, for this condition statement, we want to check that no files are found.

hack/verify-dockerless.sh
mattjmcnaughton: Also, once you add the boilerplate, the script will be running the KUBE_ROOT directory so you will want to update all paths from `../pkg/kubelet` to `./pkg/kubelet`.

hack/verify-dockerless.sh
```diff
@@ -0,0 +1,15 @@
...
+        echo $not_have_dockerless
+fi
+imports_docker=$(find ../{cmd,pkg}/kubelet/ -name "*.go" -exec grep -q 'github.com/docker/docker' {} \; -print | grep -v ../pkg/kubelet/dockershim)
+if [[ $imports_docker ]]; then
+        echo "Following files imports docker outside dockershim"
+        echo $imports_docker
```
mattjmcnaughton, line 9: Similarly we want to exit w/ failure condition if any files are found.

hack/verify-dockerless.sh
```diff
@@ -0,0 +1,15 @@
...
+fi
+imports_dockershim=$(find ../{cmd,pkg}/kubelet/ -name "*.go" -exec grep -q 'kubelet/dockershim' {} \; -print | grep -v dockershim)
+if [[ $imports_dockershim ]]; then
+        echo "Following files imports dockershim outside dockershim folder and kubelet_dockershim.go"
```
mattjmcnaughton, line 13: Similarly, we want to exit w/ failure condition if any files are found.

One note... it looks like `pkg/kubelet/legacy/logs.go` currently contains `pkg/kubelet/dockershim` in a comment. We'll want to update that file so that its not returned by our grep statement. Should be as simple as modifying the comment.

mattjmcnaughton (commented):
Hi @yash97 - thanks for your work here :)

Off to a great start. Added in guidance for some next steps :)

cc @dims

hack/verify-dockerless.sh
yash97: ok sure will do that

hack/verify-dockerless.sh
yash97: sure

Issue #91073, opened 2020-05-13
Kubelet gets "Timeout: Too large resource version" error from the API server after network outage


rkojedzinszky:
**What happened**:
I have disconnected a node from network for a few minutes. After reconnecting, I keep receiving such error messages from kubelet on the node, even after 15 minutes in reconnected state:

```
May 13 19:44:54 k8s-node04 kubelet[598]: I0513 19:44:54.043005     598 trace.go:116] Trace[1747581308]: "Reflector ListAndWatch" name:object-"kube-system"/"default-token-h8dz9" (started: 2020-05-13 19:44:14.938918643 +0000 UTC m=+81978.107654790) (total time: 39.10398118s):
May 13 19:44:54 k8s-node04 kubelet[598]: Trace[1747581308]: [39.10398118s] [39.10398118s] END
May 13 19:44:54 k8s-node04 kubelet[598]: E0513 19:44:54.043090     598 reflector.go:178] object-"kube-system"/"default-token-h8dz9": Failed to list *v1.Secret: Timeout: Too large resource version: 159128021, current: 159127032
May 13 19:45:16 k8s-node04 kubelet[598]: I0513 19:45:16.944515     598 trace.go:116] Trace[527369896]: "Reflector ListAndWatch" name:k8s.io/client-go/informers/factory.go:135 (started: 2020-05-13 19:44:37.84920601 +0000 UTC m=+82001.017941865) (total time: 39.095209656s):
May 13 19:45:16 k8s-node04 kubelet[598]: Trace[527369896]: [39.095209656s] [39.095209656s] END
May 13 19:45:16 k8s-node04 kubelet[598]: E0513 19:45:16.944595     598 reflector.go:178] k8s.io/client-go/informers/factory.go:135: Failed to list *v1beta1.RuntimeClass: Timeout: Too large resource version: 159128061, current: 159127066
May 13 19:45:23 k8s-node04 kubelet[598]: I0513 19:45:23.959866     598 trace.go:116] Trace[243135295]: "Reflector ListAndWatch" name:k8s.io/kubernetes/pkg/kubelet/kubelet.go:517 (started: 2020-05-13 19:44:44.860565979 +0000 UTC m=+82008.029301834) (total time: 39.099201281s):
May 13 19:45:23 k8s-node04 kubelet[598]: Trace[243135295]: [39.099201281s] [39.099201281s] END
[...]

PR #92537, opened 2020-06-26
Fix bug in reflector not recovering from "Too large resource version"…

wojtek-t:
Ref https://github.com/kubernetes/kubernetes/issues/91073

```release-note
Fix bug in reflector that couldn't recover from "Too large resource version" errors
```

/kind bug

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
+++ b/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
@@ -873,6 +873,9 @@ const (
 	// FieldManagerConflict is used to report when another client claims to manage this field,
 	// It should only be returned for a request using server-side apply.
 	CauseTypeFieldManagerConflict CauseType = "FieldManagerConflict"
+	// CauseTypeResourceVersionTooLarge is used to report that the requested resource version
+	// is newer than the data observed by the API server, so the request cannot be served.
+	CauseTypeResourceVersionTooLarge CauseType = "ResourceVersionTooLarge"
 )
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/staging/src/k8s.io/apiserver/pkg/storage/errors.go
+++ b/staging/src/k8s.io/apiserver/pkg/storage/errors.go
@@ -177,7 +177,12 @@ var tooLargeResourceVersionCauseMsg = "Too large resource version"
 // a minimum resource version that is larger than the largest currently available resource version for a requested resource.
 func NewTooLargeResourceVersionError(minimumResourceVersion, currentRevision uint64, retrySeconds int) error {
 	err := errors.NewTimeoutError(fmt.Sprintf("Too large resource version: %d, current: %d", minimumResourceVersion, currentRevision), retrySeconds)
-	err.ErrStatus.Details.Causes = []metav1.StatusCause{{Message: tooLargeResourceVersionCauseMsg}}
+	err.ErrStatus.Details.Causes = []metav1.StatusCause{
+		{
+			Type:    metav1.CauseTypeResourceVersionTooLarge,
+			Message: tooLargeResourceVersionCauseMsg,
+		},
+	}
 	return err
 }
 
@@ -186,15 +191,5 @@ func IsTooLargeResourceVersion(err error) bool {
 	if !errors.IsTimeout(err) {
 		return false
 	}
-	switch t := err.(type) {
-	case errors.APIStatus:
-		if d := t.Status().Details; d != nil {
-			for _, cause := range d.Causes {
-				if cause.Message == tooLargeResourceVersionCauseMsg {
-					return true
-				}
-			}
-		}
-	}
-	return false
+	return errors.HasStatusCause(err, metav1.CauseTypeResourceVersionTooLarge)
 }
--- a/staging/src/k8s.io/client-go/tools/cache/reflector.go
+++ b/staging/src/k8s.io/client-go/tools/cache/reflector.go
@@ -82,9 +82,9 @@ type Reflector struct {
 	// observed when doing a sync with the underlying store
 	// it is thread safe, but not synchronized with the underlying store
 	lastSyncResourceVersion string
-	// isLastSyncResourceVersionGone is true if the previous list or watch request with lastSyncResourceVersion
-	// failed with an HTTP 410 (Gone) status code.
-	isLastSyncResourceVersionGone bool
+	// isLastSyncResourceVersionUnavailable is true if the previous list or watch request with
+	// lastSyncResourceVersion failed with an "expired" or "too large resource version" error.
+	isLastSyncResourceVersionUnavailable bool
 	// lastSyncResourceVersionMutex guards read/write access to lastSyncResourceVersion
 	lastSyncResourceVersionMutex sync.RWMutex
 	// WatchListPageSize is the requested chunk size of initial and resync watch lists.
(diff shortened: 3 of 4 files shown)


staging/src/k8s.io/apiserver/pkg/storage/tests/cacher_test.go
```diff
@@ -356,10 +356,7 @@ func TestTooLargeResourceVersionList(t *testing.T) {
 
 	result := &example.PodList{}
 	err = cacher.List(context.TODO(), "pods/ns", storage.ListOptions{ResourceVersion: listRV, Predicate: storage.Everything}, result)
-	if !errors.IsTimeout(err) {
```
tedyu, line 359: timeout error is no longer checked ?

wojtek-t: It's checked as part of IsTooLargeResourceVersion.

staging/src/k8s.io/apimachinery/pkg/api/errors/errors.go
```diff
@@ -564,6 +564,21 @@ func IsTimeout(err error) bool {
 	return ReasonForError(err) == metav1.StatusReasonTimeout
 }
 
+// IsTooLargeResourceVersion returns true if the error is a TooLargeResourceVersion error.
+func IsTooLargeResourceVersion(err error) bool {
+	if !IsTimeout(err) {
+		return false
+	}
```
liggitt, line 571: this is a little strange... why does a CauseTypeResourceVersionTooLarge always have to be a timeout?

staging/src/k8s.io/apimachinery/pkg/api/errors/errors.go
```diff
@@ -564,6 +564,21 @@ func IsTimeout(err error) bool {
...
+	if status := APIStatus(nil); errors.As(err, &status) && status.Status().Details != nil {
+		for _, cause := range status.Status().Details.Causes {
+			if cause.Type == metav1.CauseTypeResourceVersionTooLarge {
+				return true
+			}
+		}
+	}
```
liggitt, line 578: could this be simplified to this:
```go
return HasStatusCause(metav1.CauseTypeResourceVersionTooLarge)
```

staging/src/k8s.io/apiserver/pkg/storage/errors.go
```diff
@@ -171,30 +171,10 @@ func NewInternalErrorf(format string, a ...interface{}) InternalError {
...
 }
 
-var tooLargeResourceVersionCauseMsg = "Too large resource version"
-
 // NewTooLargeResourceVersionError returns a timeout error with the given retrySeconds for a request for
 // a minimum resource version that is larger than the largest currently available resource version for a requested resource.
 func NewTooLargeResourceVersionError(minimumResourceVersion, currentRevision uint64, retrySeconds int) error {
 	err := errors.NewTimeoutError(fmt.Sprintf("Too large resource version: %d, current: %d", minimumResourceVersion, currentRevision), retrySeconds)
-	err.ErrStatus.Details.Causes = []metav1.StatusCause{{Message: tooLargeResourceVersionCauseMsg}}
+	err.ErrStatus.Details.Causes = []metav1.StatusCause{{Type: metav1.CauseTypeResourceVersionTooLarge}}
```
liggitt, line 178: should we also continue setting the message?

staging/src/k8s.io/apiserver/pkg/storage/errors.go
```diff
@@ -171,30 +171,10 @@ func NewInternalErrorf(format string, a ...interface{}) InternalError {
...
 	return err
 }
-
-// IsTooLargeResourceVersion returns true if the error is a TooLargeResourceVersion error.
-func IsTooLargeResourceVersion(err error) bool {
```
liggitt, line 185: if we want to backport this, I would probably leave this method here and delegate to `apierrors.IsTooLargeResourceVersion(err)` in old releases, rather than remove an exported method

staging/src/k8s.io/apiserver/pkg/storage/tests/cacher_test.go
liggitt: that seems a bit strange to me... timeout reason and TooLargeResourceVersion cause seem orthogonal to me

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
```diff
@@ -873,6 +873,9 @@ const (
 	// FieldManagerConflict is used to report when another client claims to manage this field,
 	// It should only be returned for a request using server-side apply.
 	CauseTypeFieldManagerConflict CauseType = "FieldManagerConflict"
+	// CauseTypeResourceVersionTooLarge is used to report that resource version is coming
+	// from the future and request cannot be served.
```
liggitt, line 877: "resource version is coming from the future" is confusing... consider rephrasing to something like

> CauseTypeResourceVersionTooLarge is used to report that that requested resource version is newer than the data observed by the API server, so the request cannot be served.

staging/src/k8s.io/client-go/tools/cache/reflector.go
```diff
@@ -288,7 +288,7 @@ func (r *Reflector) ListAndWatch(stopCh <-chan struct{}) error {
 			}
 
 			list, paginatedResult, err = pager.List(context.Background(), options)
-			if isExpiredError(err) {
+			if isExpiredError(err) || apierrors.IsTooLargeResourceVersion(err) {
```
liggitt, line 291: This makes the docs below, and the docs for `setIsLastSyncResourceVersionExpired` misleading, since we're setting that in cases other than for expired errors. Should we rename `setIsLastSyncResourceVersionExpired` to `setIsLastSyncResourceVersionUnavailable` or something that can make sense for both expired and toolarge errors?

staging/src/k8s.io/apimachinery/pkg/api/errors/errors.go
wojtek-t: nice - didn't know about this function

staging/src/k8s.io/apiserver/pkg/storage/errors.go
wojtek-t: done

staging/src/k8s.io/apiserver/pkg/storage/errors.go
wojtek-t: done

staging/src/k8s.io/client-go/tools/cache/reflector.go
wojtek-t: makes sense - done

staging/src/k8s.io/apiserver/pkg/storage/tests/cacher_test.go
wojtek-t: reverted

staging/src/k8s.io/apimachinery/pkg/api/errors/errors.go
wojtek-t: But this is how we create the "TooLargeResourceVersion" error.

Do you suggest dropping it and leaving it rely only on the cause?

wojtek-t (commented):
@liggitt - thanks for the review; PTAL

staging/src/k8s.io/apimachinery/pkg/api/errors/errors.go
wojtek-t: Actually - I played a bit more with the PR and removed this part completely. The cause type being the official type seems to be enough for the purpose of this PR and makes it easier to cherrypick.

wojtek-t (commented):
@liggitt - PTAL

PR #89999, opened 2020-04-09
If firstTimestamp is not set use eventTime when printing event

soltysh:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
This is a follow up to #86557, it looks like it's also possible to have events with empty `firstTimestamp`, in those cases we fallback to `eventTime`. 

**Special notes for your reviewer**:
/assign @sallyom 

**Does this PR introduce a user-facing change?**:
```release-note
If firstTimestamp is not set use eventTime when printing event
```

PR #92542, opened 2020-06-26
Fix v1beta1.Event printing in kubectl

dgrisonnet:
**What type of PR is this?**

/kind bug

This PR intends to fix incorrect data displayed by kubectl v1beta1.Event printing.
The new Events API [deprecates](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/instrumentation/events-redesign.md#backward-compatibility) a few fields that are used by kubectl to print Events:
- FirstTimestamp => DeprecatedFirstTimestamp
- LastTimestamp => DeprecatedLastTimestamp
- Count => DeprecatedCount

For the timestamps, the new API is replacing them with already existing more precise timestamps ([details](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/instrumentation/events-redesign.md#api-changes-goals)):
- FirstTimestamp => EventTime
- LastTimestamp => Series.LastObservedTime

As for the count, it is replaced by the already existing Series.Count.

That being said, there was already an effort to support the new FirstTimestamp alternative in #89999.
However, the LastTimestamp and Count deprecations have not been handled yet. This PR is an attempt to handle these cases.

The first condition added in this PR is handling the new notion of Series brought by the new Event API. Thus, it is making use of the Series.LastObservedTime and Series.Count.

To quote the new [Event API KEP](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/instrumentation/events-redesign.md#changes-in-eventrecorder) on the notion of series:
> - When event is emitted for the first time it's written to the API server without series field set.
> - When isomorphic event is emitted within the threshold from the original one EventRecorder detects the start of the series, updates the Event object, with the Series field set carrying count and sets State to EventSeriesStateOngoing. In the EventRecorder it also creates an entry in activeSeries map with the timestamp of last observed Event in the series.

Programmatically, this translates into this [piece of code](https://github.com/kubernetes/client-go/blob/v0.19.0-beta.2/tools/events/event_broadcaster.go#L162-L173) where we can see that the latest timestamp is now being recorded in Series.LastObservedTime and the count in Series.Count.

As for the second, it is handling the following edge cases from the new notion of Event singleton:
> The biggest change is the semantics of the Event object in case of loops. If Series is nil it means that Event is a singleton, i.e. it happened only once and the semantics is exactly the same as currently in Events with count = 1. If Series is not nil it means that the Event is either beginning or the end of an Event series - equivalence of current Events with count > 1. Events for ongoing series have Series.State set to EventSeriesStateOngoing, while endings have Series.State set to EventSeriesStateFinished. [link](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/instrumentation/events-redesign.md#comparison-between-old-and-new-api)

```release-note
NONE
```
/sig cli
/area kubectl

### DIFF
--- a/pkg/printers/internalversion/printers.go
+++ b/pkg/printers/internalversion/printers.go
@@ -1673,9 +1673,23 @@ func printEvent(obj *api.Event, options printers.GenerateOptions) ([]metav1.Tabl
 	if obj.FirstTimestamp.IsZero() {
 		firstTimestamp = translateMicroTimestampSince(obj.EventTime)
 	}
+
 	lastTimestamp := translateTimestampSince(obj.LastTimestamp)
 	if obj.LastTimestamp.IsZero() {
 		lastTimestamp = firstTimestamp
+	}
+
+	count := obj.Count
+	if obj.Series != nil {
+		lastTimestamp = translateMicroTimestampSince(obj.Series.LastObservedTime)
+		// When a series is created for the first time, its count is set to 1.
+		// However, for a series to be created, there needs to be an isomorphic
+		// singleton event created beforehand. This singleton event is not
+		// counted in the series count which is why one is added here.
+		count = obj.Series.Count + 1
+	} else if count == 0 {
+		// Singleton events don't have a count set in the new API.
+		count = 1
 	}
 
 	var target string
@@ -1694,7 +1708,7 @@ func printEvent(obj *api.Event, options printers.GenerateOptions) ([]metav1.Tabl
 			formatEventSource(obj.Source),
 			strings.TrimSpace(obj.Message),
 			firstTimestamp,
-			int64(obj.Count),
+			int64(count),
 			obj.Name,
 		)
 	} else {
--- a/pkg/printers/internalversion/printers_test.go
+++ b/pkg/printers/internalversion/printers_test.go
@@ -178,11 +178,11 @@ func TestPrintEvent(t *testing.T) {
 				FirstTimestamp: metav1.Time{Time: time.Now().UTC().AddDate(0, 0, -3)},
 				Count:          1,
 				Type:           api.EventTypeWarning,
-				ObjectMeta:     metav1.ObjectMeta{Name: "event3"},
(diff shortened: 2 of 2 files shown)


pkg/printers/internalversion/printers.go
```diff
@@ -1673,11 +1673,24 @@ func printEvent(obj *api.Event, options printers.GenerateOptions) ([]metav1.Tabl
...
 	lastTimestamp := translateTimestampSince(obj.LastTimestamp)
 	if obj.LastTimestamp.IsZero() {
 		lastTimestamp = firstTimestamp
 	}
 
+	count := obj.Count
+	if obj.Series != nil {
+		lastTimestamp = translateMicroTimestampSince(obj.Series.LastObservedTime)
+		// Add the first event to the count.
+		count = obj.Series.Count + 1
```
soltysh, line 1686: If I'm reading the proposal and code correctly, there's no need to add one here.

pkg/printers/internalversion/printers.go
```diff
@@ -1673,11 +1673,24 @@ func printEvent(obj *api.Event, options printers.GenerateOptions) ([]metav1.Tabl
...
+	}
+
+	// Singleton events don't have a count set in the new API.
+	if count == 0 {
```
soltysh, line 1690: This actually could be and else to the above condition, since a singleton will not have `Series` set either.

soltysh (changes requested):
I left you some comments, also please squash changes to a single commit.

pkg/printers/internalversion/printers.go
dgrisonnet: When a series is created for the first time, its [count is set to 1](https://github.com/kubernetes/client-go/blob/v0.19.0-beta.2/tools/events/event_broadcaster.go#L168-L171). However, for a series to be created, there needs to be an isomorphic singleton event created beforehand. This singleton event is not counted in the series count which is why I added one here.

soltysh (approved):
/lgtm
/approve

Issue #92511, opened 2020-06-25
[Failing Test] capg-conformance-v1alpha3-k8s-master (ci-cluster-api-provider-gcp-make-conformance-v1alpha3-k8s-ci-artifacts)


hasheddan:
**Which jobs are failing**:

`capg-conformance-v1alpha3-k8s-master (ci-cluster-api-provider-gcp-make-conformance-v1alpha3-k8s-ci-artifacts)`

**Which test(s) are failing**:

`Overall`

**Since when has it been failing**:

`06-24-20 21:29 CDT`

**Testgrid link**:

https://testgrid.k8s.io/sig-release-master-informing#capg-conformance-v1alpha3-k8s-master

**Reason for failure**:

```
[0;32m    ubuntu-1804: fatal: [default]: FAILED! => {"changed": false, "msg": "Unable to find a checksum for file 'cni-plugins-amd64-v0.8.6.tgz' in 'https://github.com/containernetworking/plugins/releases/download/v0.8.6/cni-plugins-linux-amd64-v0.8.6.tgz.sha256'"}[0m
```

**Anything else we need to know**:

/sig cluster-lifecycle
/cc @kubernetes/ci-signal
/priority critical-urgent
/milestone v1.19

PR #92509, opened 2020-06-25
Fix assignment to nil map in e2e scheduling predicates

hasheddan:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

Ready schedulable nodes are being inserted into an unitialized string
set, causing an assignment to entry in nil map in the underlying data
structure. This initializes the string set before attempting to insert
nodes.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

**Which issue(s) this PR fixes**:

Fixes #92502

**Special notes for your reviewer**:

Bug introduced in https://github.com/kubernetes/kubernetes/pull/92450

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

Issue #92502, opened 2020-06-25
[Failing test] Conformance - GCE- master (ci-kubernetes-gce-conformance-latest)


hkamel:
**Which jobs are failing**:
> Conformance - CE- master (ci-kubernetes-gce-conformance-latest)

**Which test(s) are failing**:
> - Overall
> - TestChanges[sig-scheduling] SchedulerPredicates [Serial] validates resource limits of pods that are allowed to run [Conformance]
>- [sig-scheduling] SchedulerPredicates [Serial] validates that NodeSelector is respected if matching [Conformance]
>- [sig-scheduling] SchedulerPredicates [Serial] validates that NodeSelector is respected if not matching [Conformance]
>- [sig-scheduling] SchedulerPredicates [Serial] validates that there exists conflict between pods with same hostPort and protocol but one using 0.0.0.0 hostIP [Conformance]
>- [sig-scheduling] SchedulerPredicates [Serial] validates that there is no conflict between pods with same hostPort but different hostIP and protocol [Conformance]
>- [sig-scheduling] SchedulerPreemption [Serial] PreemptionExecutionPath runs ReplicaSets to verify preemption running path [Conformance]
>- [sig-scheduling] SchedulerPreemption [Serial] validates basic preemption works [Conformance]
>- [sig-scheduling] SchedulerPreemption [Serial] validates lower priority pod preemption by critical pod [Conformance]

**Since when has it been failing**:
> 06-25 21:37 PDT

**Testgrid link**:
>https://testgrid.k8s.io/sig-release-master-blocking#Conformance%20-%20GCE%20-%20master

**Reason for failure**:
> /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/scheduling/preemption.go:89
Test Panicked
/usr/local/go/src/runtime/map_faststr.go:204

> Panic: assignment to entry in nil map

> Full stack:
k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/sets.String.Insert(0x0, 0xc001406da8, 0x1, 0x1, 0x0)
	/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/sets/string.go:51 +0x78
k8s.io/kubernetes/test/e2e/scheduling.glob..func5.2()
[...]

PR #92545, opened 2020-06-26
Do not ignore unscheduled pods when NodeName not in set of worker nodes

hasheddan:
**What type of PR is this?**

/kind bug

When node scheduling tests were updated to use worker instead of master
nodes the GetPodsScheduled function, which is tasked with getting all
scheduled and not scheduled pods inadvertently was changed to ignore all
pods that have an empty NodeName before checking whether pods had been
scheduled or not. This updates the function to include pods without a
NodeName in the check for unscheduled pods.

Relevant issues for failing tests: https://github.com/kubernetes/kubernetes/issues/92511 https://github.com/kubernetes/kubernetes/issues/92502

Follow up to https://github.com/kubernetes/kubernetes/pull/92509

Original PR: https://github.com/kubernetes/kubernetes/pull/92450

/cc @oomichi @BenTheElder 

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/scheduling/predicates.go
+++ b/test/e2e/scheduling/predicates.go
@@ -1045,23 +1045,20 @@ func translateIPv4ToIPv6(ip string) string {
 // GetPodsScheduled returns a number of currently scheduled and not scheduled Pods on worker nodes.
 func GetPodsScheduled(workerNodes sets.String, pods *v1.PodList) (scheduledPods, notScheduledPods []v1.Pod) {
 	for _, pod := range pods.Items {
-		if workerNodes.Has(pod.Spec.NodeName) {
-			if pod.Spec.NodeName != "" {
-				_, scheduledCondition := podutil.GetPodCondition(&pod.Status, v1.PodScheduled)
-				framework.ExpectEqual(scheduledCondition != nil, true)
-				if scheduledCondition != nil {
-					framework.ExpectEqual(scheduledCondition.Status, v1.ConditionTrue)
-					scheduledPods = append(scheduledPods, pod)
-				}
-			} else {
-				_, scheduledCondition := podutil.GetPodCondition(&pod.Status, v1.PodScheduled)
-				framework.ExpectEqual(scheduledCondition != nil, true)
-				if scheduledCondition != nil {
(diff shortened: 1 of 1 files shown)


hasheddan:
The following tests are currently being affected across a number of jobs:
`[sig-scheduling] SchedulerPredicates [Serial] validates resource limits of pods that are allowed to run [Conformance]`
`[sig-scheduling] SchedulerPredicates [Serial] validates that NodeSelector is respected if not matching [Conformance]`

hasheddan:
This is a good one to look at for more context: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-gce-conformance-latest/1276467007495081987

test/e2e/scheduling/predicates.go
```diff
@@ -1045,23 +1045,20 @@ func translateIPv4ToIPv6(ip string) string {
 // GetPodsScheduled returns a number of currently scheduled and not scheduled Pods on worker nodes.
 func GetPodsScheduled(workerNodes sets.String, pods *v1.PodList) (scheduledPods, notScheduledPods []v1.Pod) {
 	for _, pod := range pods.Items {
-		if workerNodes.Has(pod.Spec.NodeName) {
```
hasheddan, line 1048: Previously this was `if !masterNodes.Has(pod.Spec.NodeName)` so this isn't fully going back to the same functionality. The check for unscheduled nodes could consider master nodes in this case, but from my understanding of this check that would be fine.

Huang-Wei:
/lgtm
/approve

Thanks @hasheddan !

/hold
in case @oomichi wants to take another look.

test/e2e/scheduling/predicates.go
oomichi: Very reasonable change for me.
Why we checked the NodeName is in names of workerNodes before checking the NodeName is empty (^^;)

oomichi (commented):
@hasheddan Thanks for doing this again.

/lgtm
/hold cancel

BenTheElder:
https://github.com/kubernetes/kubernetes/issues/92563
/retest

PR #92549, opened 2020-06-26
Sets desiredcapacity instead of minsize and maxsize on ResizeInstanceGroup

### LINKED ISSUE #92548: Test autoscaling_timer.go incorrectly resets min/max ASG size after test.
**What happened**:
After the test ASG min/max were reset to the desired node count. 

Min/Max shouldn't be reset, desired should.

Run this command twice:
``` make WHAT=test/e2e/e2e.test &&  _output/bin/e2e.test --ginkgo.focus="\[sig-autoscaling\] \[Feature:ClusterSizeAutoscalingScaleUp\] \[Slow\] Autoscaling" --provider=aws --gce-zone="us-west-2" --node-instance-group=eks-44b97165-2d96-674f-8417-c08c5e66d785```

**Environment**:
- Cloud provider or hardware configuration: AWS
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

ellistarn:
**What type of PR is this?**
/kind bug

Currently, after the autoscaling_timer.go test (the only callsite of this function), the ASG is reset with min and max nodes. If the test is ever run again, it fails due to the ASG being unable to scale up or down. This simply resets the Desired count leaving the original min/max intact.

Fixes #92548

### DIFF
--- a/staging/src/k8s.io/legacy-cloud-providers/aws/aws_instancegroups.go
+++ b/staging/src/k8s.io/legacy-cloud-providers/aws/aws_instancegroups.go
@@ -35,8 +35,7 @@ var _ InstanceGroups = &Cloud{}
 func ResizeInstanceGroup(asg ASG, instanceGroupName string, size int) error {
 	request := &autoscaling.UpdateAutoScalingGroupInput{
 		AutoScalingGroupName: aws.String(instanceGroupName),
-		MinSize:              aws.Int64(int64(size)),
-		MaxSize:              aws.Int64(int64(size)),
+		DesiredCapacity:      aws.Int64(int64(size)),
 	}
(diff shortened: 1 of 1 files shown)


nckturner:
Looks like this runs after the test https://github.com/kubernetes/kubernetes/blob/master/test/e2e/autoscaling/autoscaling_timer.go#L78 and setting max and min to the same value prevents further scaling on that ASG.

/lgtm

justinsb:
Looks good, by comparing what gce does

/approve
/lgtm

ellistarn:
Any suggestions on unrelated flakes @nckturner? /retest

BenTheElder:
/retest
^ commands have to be on their own line
-ipv6 is not blocking, only jobs with `Required` marked at the bottom are.
integration and verify are however.

PR #92550, opened 2020-06-26
Change Service.Spec.IPFamily to IPFamilies for forward-compatibility

danwinship:
**What type of PR is this?**
/kind api-change

TL;DR: Instead of leaving the alpha dual-stack Service API unchanged in 1.19 and then breaking-and-improving it in 1.20, I propose to break-but-not-improve the API in 1.19 and then improve-but-not-break it in 1.20, to improve upgradability of dual-stack 1.19 clusters to 1.20.

----

Per https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20180612-ipv4-ipv6-dual-stack.md#services we will be improving the API around Services in dual-stack clusters in 1.20. The plan is to make this change at the start of the 1.20 cycle.

The new API is incompatible with the existing (alpha) API, which means that even where the semantics are compatible, dual-stack-aware Service yaml objects written for a 1.19 dual-stack cluster would not work in a 1.20 dual-stack cluster, and vice versa. However, _with respect to the subset of functionality that already exists in 1.19_, the API difference is trivial, and we could easily expose the existing 1.19 functionality using the proposed 1.20 syntax.

Thus, this PR replaces `ipFamily` with `ipFamilies` as in 1.20, but with the restriction that it can only contain a single value. This means early adopters of dual-stack in 1.19 can create (single-stack) Service yaml objects that will behave the same way in dual-stack 1.19 clusters and dual-stack 1.20 clusters. (This makes dual-stack 1.19 incompatible with dual-stack 1.18, but we had already decided there was going to be an API break; this just moves it sooner.)

Specifically:
- Legacy Services will continue to be defaulted to "single stack of the cluster default family", as in 1.18 dual-stack / proposed 1.20 dual-stack.
- A user who wants to create a single-stack IPv4 service in an IPv6-primary dual-stack cluster, or a single-stack IPv6 service in an IPv4-primary dual-stack cluster, can specify `ipFamilies: ["IPv4"]` or `ipFamilies: ["IPv6"]` to get that behavior.
  - This is also a valid way to specify a single-stack IPv4 or single-stack IPv6 service in 1.20, so the same yaml could be reused after upgrading to 1.20 and get the same behavior.
  - If Services are created in this way in 1.19 and the cluster is then updated to 1.20, the defaulting logic in 1.20 should belatedly initialize `ipFamilyPolicy` to `SingleStack` correctly based on the existing single-valued `ipFamilies`.
- `ipFamilies: ["IPv4", "IPv6"]` or `ipFamilies: ["IPv6", "IPv4"]` will result in a validation error (in 1.19).
- Other than the syntactic change of `IPFamily *IPFamily` → `IPFamilies []IPFamily`, no changes to functionality, defaulting, or validation are made. All functionality that formerly would have been available in 1.19 will still be available in 1.19, and all functionality that formerly would not have been available in 1.19 will still not be available in 1.19.
  - In particular, the handling of pre-existing Headless Services will continue to be sub-optimal, since we are not introducing `ipFamilyPolicy` and not defaulting `ipFamilies` in pre-existing Services. (The problems can be fixed in some cases by manually setting `ipFamilies` on the Service, which again should be handled correctly on upgrade to 1.20.)

```release-note
In clusters using the alpha IPv6DualStack feature, the Service.Spec.IPFamily field has
been replaced with Service.Spec.IPFamilies, for better forward-compatibility with the
planned dual-stack Service API in Kubernetes 1.20. The semantics (in 1.19) are the same
as before, but the field is now a single-valued array. eg, `ipFamily: IPv6` in 1.18
becomes `ipFamilies: [ "IPv6" ]` in 1.19.
```

```docs
[KEP]: https://github.com/kubernetes/enhancements/blob/e9446ed3/keps/sig-network/20180612-ipv4-ipv6-dual-stack.md#services
```
(FIXME: update the link above after the KEP is updated for `preferDualStack`→`ipFamilyPolicy`)

/sig network
/priority important-soon
/assign @thockin @khenidak 
/cc @dcbw


dcbw:
@danwinship can you rebase?

pkg/apis/core/validation/conditional_validation.go
```diff
@@ -40,57 +41,62 @@ func ValidateConditionalService(service, oldService *api.Service, allowedIPFamil
...
 
-// validateIPFamily checks the IPFamily field.
-func validateIPFamily(service, oldService *api.Service, allowedIPFamilies []api.IPFamily) field.ErrorList {
+// validateIPFamilies checks the IPFamilies field.
+func validateIPFamilies(service, oldService *api.Service, allowedIPFamilies []api.IPFamily) field.ErrorList {
 	var errs field.ErrorList
 
 	// specifically allow an invalid value to remain in storage as long as the user isn't changing it, regardless of gate
-	if oldService != nil && oldService.Spec.IPFamily != nil && service.Spec.IPFamily != nil && *oldService.Spec.IPFamily == *service.Spec.IPFamily {
+	if oldService != nil && oldService.Spec.IPFamilies != nil && service.Spec.IPFamilies != nil && reflect.DeepEqual(oldService.Spec.IPFamilies, service.Spec.IPFamilies) {
```
dcbw, line 54: @danwinship do we care of the set between old & new is the same, but just different order? That would be rejected here, but you could argue it's a NOP and should pass validation.

danwinship: it's not a NOP. "dual stack IPv4 primary" and "dual stack IPv6 primary" are different

PR #92554, opened 2020-06-26
Update k8s.io/klog to v2.2.0

serathius:
/kind feature

Ref https://github.com/kubernetes/enhancements/issues/1602

```release-note
NONE
```

/cc @dims

### DIFF
--- a/staging/src/k8s.io/component-base/logs/json/json.go
+++ b/staging/src/k8s.io/component-base/logs/json/json.go
@@ -35,32 +35,23 @@ var (
 	timeNow = time.Now
 )
 
-type noopInfoLogger struct{}
-
-// Enabled func in noopInfoLogger always return false
-func (l *noopInfoLogger) Enabled() bool {
-	return false
-}
-
-func (l *noopInfoLogger) Info(_ string, _ ...interface{}) {}
-
-var disabledInfoLogger = &noopInfoLogger{}
-
-type infoLogger struct {
-	lvl zapcore.Level
+// zapLogger is a logr.Logger that uses Zap to record log.
+type zapLogger struct {
+	// NB: this looks very similar to zap.SugaredLogger, but
+	// deals with our desire to have multiple verbosity levels.
 	l   *zap.Logger
+	lvl zapcore.Level
 }
 
-// implement logr.InfoLogger
-var _ logr.InfoLogger = &infoLogger{}
+// implement logr.Logger
+var _ logr.Logger = &zapLogger{}
 
-// Enabled always return true
-func (l *infoLogger) Enabled() bool {
-	return true
+func (l *zapLogger) Enabled() bool {
+	return l.l.Core().Enabled(l.lvl)
 }
 
 // Info write message to error level log
-func (l *infoLogger) Info(msg string, keysAndVals ...interface{}) {
+func (l *zapLogger) Info(msg string, keysAndVals ...interface{}) {
 	if checkedEntry := l.l.Check(l.lvl, msg); checkedEntry != nil {
 		checkedEntry.Time = timeNow()
 		checkedEntry.Write(l.handleFields(keysAndVals)...)
@@ -70,7 +61,7 @@ func (l *infoLogger) Info(msg string, keysAndVals ...interface{}) {
 // dPanic write message to DPanicLevel level log
 // we need implement this because unit test case need stub time.Now
 // otherwise the ts field always changed
-func (l *infoLogger) dPanic(msg string, keysAndVals ...interface{}) {
+func (l *zapLogger) dPanic(msg string, keysAndVals ...interface{}) {
 	entry := zapcore.Entry{
 		Level:   zapcore.DPanicLevel,
 		Time:    timeNow(),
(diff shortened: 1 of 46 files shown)


dims:
please update go-logr/logr to 0.2.0

yuzhiquan:
> please update go-logr/logr to 0.2.0 and use logr.Logger instead of logr.InfoLogger
> 
> [go-logr/logr@c53dffe](https://github.com/go-logr/logr/commit/c53dffe447a2e465738458d784d8623605ad8827)

Should update [json format]( https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/logs/json/json.go#L55) here, @serathius if deadline allowed, i can work on it.

serathius:
Hey @yuzhiquan, I had a draft prepared when you asked, so I decided to finish this up.
On the other hand I noticed very strange behavior of verbosity. I don't think verbosity is correctly implemented for json logger. 
e.g. infoLogger.V(2) write logs with `v=1`, also verbosity of json logger doesn't match K8s guidelines https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md

Are you interested in looking into it?

yuzhiquan:
> Hey @yuzhiquan, I had a draft prepared when you asked, so I decided to finish this up.
> On the other hand I noticed very strange behavior of verbosity. I don't think verbosity is correctly implemented for json logger.
> e.g. [infoLogger.V(2) write logs with `{v:1}`](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/logs/json/json_test.go#L99), also verbosity of json logger doesn't match [K8s guidelines](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md)
> 
> Are you interested in looking into it?

@serathius 
Sure, i will take Verbose from here.

dims:
/assign @liggitt 

Jordan, this looks ready!

liggitt:
Vendor mechanics lgtm, didn't take a look at the reactionary changes yet. Would like @thockin's ack since he was involved in the interface change (looking at https://github.com/go-logr/logr/commits/master)

Since logr will be a shared dependency now, it would be great if we could get the interface stabilized sooner rather than later.

staging/src/k8s.io/component-base/logs/json/json.go
```diff
@@ -35,32 +35,23 @@ var (
...
-}
-
-func (l *noopInfoLogger) Info(_ string, _ ...interface{}) {}
-
-var disabledInfoLogger = &noopInfoLogger{}
-
-type infoLogger struct {
-	lvl zapcore.Level
+// zapLogger is a logr.Logger that uses Zap to record log.
+type zapLogger struct {
```
thockin, line 39: I am curious why we are embedding this rather than using https://github.com/go-logr/zapr ?

thockin (commented):
One question about the way logr is being used.  As for logr API stability.  I am a bit anxious about that - we haven't gotten a TON of feedback, but what we have gotten indicates the API isn't a problem. There are a few open issues (being able to offset stack frames and k/v pairs being the 2 that I car about most) that could be API impacting.

@DirectXMan12 I wonder if we should put some effort into pinning it down, doing some more examples, and just calling thee things 1.0?

staging/src/k8s.io/component-base/logs/json/json.go
serathius: There are some small differences configuring zap, e.g. EncoderConfig. As codebase was small we decided that it will be easier and faster to first develop it in k/k and propose changes to zapr later.

thockin:
OK

serathius:
> @DirectXMan12  I wonder if we should put some effort into pinning it down, doing some more examples, and just calling thee things 1.0?

Please let me know if I can help.

dims:
@DirectXMan12 @thockin Can we please ship this while you all work on `logr`?

thockin:
I have no objections - logr is pre 1.0, but that isn't different before or after this PR, so ...

thockin (commented):
Thanks!

/lgtm
/approve

vendor/k8s.io/klog/v2/klog.go
```diff
@@ -1320,7 +1320,15 @@ func (v Verbose) Infof(format string, args ...interface{}) {
...
 func (v Verbose) InfoS(msg string, keysAndValues ...interface{}) {
 	if v.enabled {
-		logging.infoS(v.logr, msg, keysAndValues)
+		logging.infoS(v.logr, msg, keysAndValues...)
+	}
+}
+
+// Error is equivalent to the global Error function, guarded by the value of v.
+// See the documentation of V for usage.
+func (v Verbose) Error(err error, msg string, args ...interface{}) {
```
liggitt, line 1329: should this be ErrorS? the signature doesn't match the global Error function as the godoc claims

liggitt:
heh, review collision

/hold for question at https://github.com/kubernetes/kubernetes/pull/92554/files#r447829913

vendor/k8s.io/klog/v2/klog.go
serathius: Yes, this looks like a bug in klog. We are upgrading to klog to pick up a bug fix https://github.com/kubernetes/klog/issues/165 We should be ok with this as currently we are not using klog.V(2).Error.

liggitt: > this looks like a bug in klog

and klog tagged a release with this? that means we can't change the signature until klog/v3, which I hope we never get to

serathius: @dims proposed to add ErrorS and add comment to discourage people from using Error.
I don't think this is blocking. Are we ok to merge?

PR #92555, opened 2020-06-26
Refactor tests in snapshottable and fix #73625

### LINKED ISSUE #73625: E2E storage: test snapshot semantics

https://github.com/kubernetes/kubernetes/pull/69036 adds a test that snapshots can be created and used. It would be good to actually verify that the snapshotting properly handles data (https://github.com/kubernetes/kubernetes/pull/69036/files#r252769254):

1. Write some data
1. Take a snapshot
1. Write more data
1. Create a new volume from snapshot
1. Validate data is the old data

**Why is this needed**:

To ensure that snapshotting really works.

/sig storage
/cc @msau42 @xing-yang @wackxu

AndiLi99:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #73625 

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/storage/testsuites/snapshottable.go
+++ b/test/e2e/storage/testsuites/snapshottable.go
@@ -29,6 +29,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/dynamic"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
@@ -71,11 +72,13 @@ func InitSnapshottableTestSuite() TestSuite {
 		tsInfo: TestSuiteInfo{
 			Name: "snapshottable",
 			TestPatterns: []testpatterns.TestPattern{
-				testpatterns.DynamicSnapshot,
+				testpatterns.DynamicSnapshotDelete,
+				testpatterns.DynamicSnapshotRetain,
 			},
 			SupportedSizeRange: e2evolume.SizeRange{
 				Min: "1Mi",
 			},
+			FeatureTag: "[Feature:VolumeSnapshotDataSource]",
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


AndiLi99:
cla recheck

AndiLi99:
recheck cla

AndiLi99:
recheck cla

msau42:
cc @saikat-royc @mattcary

AndiLi99:
I'm encountering a strange 404 error.
```test/e2e/storage/testsuites/snapshottable.go:222
Jun 29 17:23:29.816: Unexpected error:
    <*errors.StatusError | 0xc001ca74a0>: {
        ErrStatus: {
            ListMeta: {
                SelfLink: "",
                ResourceVersion: "",
                Continue: "",
                RemainingItemCount: nil,
            },
            Status: "Failure",
            Message: "the server could not find the requested resource",
            Reason: "NotFound",
            Details: {
                Name: "",
                Group: "",
                Kind: "",
                UID: "",
                Causes: [
                    {
                        Type: "UnexpectedServerResponse",
                        Message: "404 page not found",
                        Field: "",
                    },
                ],
                RetryAfterSeconds: 0,
            },
            Code: 404,
        },
    }
    the server could not find the requested resource
occurred
test/e2e/storage/testsuites/snapshottable.go:198```
Here is the line: https://github.com/kubernetes/kubernetes/blob/417bda57cb65e4d89dfa5c11553ba47619a59726/test/e2e/storage/testsuites/snapshottable.go#L197
What could be causing this? Perhaps the CRDs are not being registered before the `ginkgo.BeforeEach`?

**Comment by xing-yang** [comment]
Yeah, probably the snapshot CRDs are not registered yet.

Also can you fix the golint error from pull-kubernetes-verify?

**Comment by AndiLi99** [comment]
I just fixed the golint error. Also I added the "[Feature:VolumeSnapshotDataSource]" string to all my it blocks. The lack of it may have been causing the test case to run in the wrong job.

**Comment by AndiLi99** [comment]
this test (https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92555/pull-kubernetes-e2e-gce-storage-snapshot/1277695556403073024/) failed because the way the logging framework deals with long test names places a limit on the  nesting of descriptions using ginkgo contexts.

The filename limit is 255 bytes for most filesystems (https://serverfault.com/questions/9546/filename-length-limits-on-linux) and the library concatenates the contexts and its together.

I created an issue to track this #92622

I will reduce the length of the descriptions for this PR.

**Comment by AndiLi99** [comment]

Is this a flaky test?

**Comment by AndiLi99** [comment]

`pull-kubernetes-e2e-kind — Pod pending timeout.` what is this error?

**Comment by AndiLi99** [comment]
My tests seem to be timing out, should I label them [Slow] or do we need to increase the timeout or the number of parallel nodes running tests?

Or is there another issue that needs to be resolved?

**Comment by AndiLi99** [comment]
It looks like there is a history of these jobs being very close to the edge of the timeout limit: https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce-storage-snapshot

Previous job runs have seen it fail or succeed based on timeout with no underlying code changes.

**Comment by AndiLi99** [comment]
I think I have a better understanding of why the tests keeping timing out and failing.

Prow only runs the tests using the internal drivers found at https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/drivers/csi.go which involves installing and uninstalling the driver at each test run. Since I am testing on my local cluster using the external method there may be differences in the environment causing the timeout. 

https://github.com/kubernetes-csi/external-snapshotter/pull/329 is related.

**Comment by AndiLi99** [comment]
So this job was already failing sometimes due to timeout https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/92387/pull-kubernetes-e2e-gce-storage-snapshot/1280190077598699521. This job was running code unrelated to this PR. 

We've increased the timeout and hopefully it works now.

/test pull-kubernetes-typecheck

**Comment by AndiLi99** [comment]
#92886 flake

**Comment by AndiLi99** [comment]
ping @msau42

**Comment by AndiLi99** [comment]
hit the wrong button

**Comment by msau42** [comment]
Asking @saikat-royc and @mattcary to review first

**Comment by AndiLi99** [comment]
Thank you @msau42.

Thank you @saikat-royc and @mattcary, the PR is ready for review, all jobs passed earlier. I rebased just now to squash some commits so it is still pending and might flake.

File: test/e2e/storage/testsuites/snapshottable.go [hunk 1]
```diff
@@ -109,283 +109,245 @@ func (s *snapshottableTestSuite) DefineTests(driver TestDriver, pattern testpatt
...
+					framework.ExpectNoError(err)
+				})
+
+				// Validate volume snapshot content is not removed
+				DeleteAndWaitSnapshot(dc, vs.GetNamespace(), vs.GetName(), framework.Poll, framework.SnapshotDeleteTimeout)
+				err = utils.WaitForGVRDeletion(dc, SnapshotContentGVR, vscontent.GetName(), 1*time.Second /* poll */, 30*time.Second /* timeout */)
+				framework.ExpectError(err) // should fail deletion check
+			})
+
+			ginkgo.It("should create objects with default", func() {
```
mattcary: what do you mean by "with default" here? Isn't this really just checking that the SnapshotContent and such was created correctly/

test/e2e/storage/testsuites/snapshottable.go
```diff
@@ -109,283 +109,245 @@ func (s *snapshottableTestSuite) DefineTests(driver TestDriver, pattern testpatt
...
+
+				snapshotContentSpec := vscontent.Object["spec"].(map[string]interface{})
+				volumeSnapshotRef := snapshotContentSpec["volumeSnapshotRef"].(map[string]interface{})
+
+				// Check SnapshotContent properties
+				ginkgo.By("checking the SnapshotContent")
+				framework.ExpectEqual(snapshotContentSpec["volumeSnapshotClassName"], vsc.GetName())
+				framework.ExpectEqual(volumeSnapshotRef["name"], vs.GetName())
+				framework.ExpectEqual(volumeSnapshotRef["namespace"], vs.GetNamespace())
+			})
```
mattcary, line 236: Your test ends up creating and extra snapshot compared to the original test, and it also doesn't verify delete/retain works with whatever is happening in the snapshot check---which, granted, shouldn't matter now, but might if something is added.

I wonder if it would be better to pull out the checks in "should create objects with default" into a common function that is run by the delete & retain steps, or put it into the beforeEach step.

mattcary (commented):
Here are my comments on the refactoring commit, I haven't looked at the other one yet.

BenTheElder:
https://github.com/kubernetes/kubernetes/pull/92555#issuecomment-651875959
This means the Kubernetes based CI failed to run the job.

Note that if a job isn't marked Required at the bottom of the page it isn't required, only informational.

test/e2e/storage/testsuites/snapshottable.go
```diff
@@ -109,283 +109,245 @@ func (s *snapshottableTestSuite) DefineTests(driver TestDriver, pattern testpatt
...
+			testVolumeSizeRange := s.GetTestSuiteInfo().SupportedSizeRange
+			driverVolumeSizeRange := dDriver.GetDriverInfo().SupportedSizeRange
+			claimSize, err = getSizeRangesIntersection(testVolumeSizeRange, driverVolumeSizeRange)
+			framework.ExpectNoError(err, "determine intersection of test size range %+v and driver size range %+v", testVolumeSizeRange, driverVolumeSizeRange)
+			pvc = e2epv.MakePersistentVolumeClaim(e2epv.PersistentVolumeClaimConfig{
+				ClaimSize:        claimSize,
+				StorageClassName: &(sc.Name),
+			}, config.Framework.Namespace.Name)
+
+			framework.Logf("In creating storage class object and pvc object for driver - sc: %v, pvc: %v", sc, pvc)
```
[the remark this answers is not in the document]
AndiLi99: Good point, I will remove the extraneous Logf. For future reference, should I prefer logging with By() or Logf?

test/e2e/storage/testsuites/snapshottable.go
```diff
@@ -109,283 +109,245 @@ func (s *snapshottableTestSuite) DefineTests(driver TestDriver, pattern testpatt
...
+			cleanupSteps = append(cleanupSteps, func() {
+				framework.Logf("deleting storage class %s", sc.Name)
+				framework.ExpectNoError(cs.StorageV1().StorageClasses().Delete(context.TODO(), sc.Name, metav1.DeleteOptions{}))
+			})
+
+			ginkgo.By("creating a claim")
+			pvc, err = cs.CoreV1().PersistentVolumeClaims(pvc.Namespace).Create(context.TODO(), pvc, metav1.CreateOptions{})
+			framework.ExpectNoError(err)
+			cleanupSteps = append(cleanupSteps, func() {
+				framework.Logf("deleting claim %q/%q", pvc.Namespace, pvc.Name)
```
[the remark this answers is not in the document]
AndiLi99: Good point, will remove.

test/e2e/storage/testsuites/snapshottable.go
```diff
@@ -109,283 +109,245 @@ func (s *snapshottableTestSuite) DefineTests(driver TestDriver, pattern testpatt
...
-		l.pvc, err = l.cs.CoreV1().PersistentVolumeClaims(l.pvc.Namespace).Create(context.TODO(), l.pvc, metav1.CreateOptions{})
-		framework.ExpectNoError(err)
-		l.cleanupSteps = append(l.cleanupSteps, func() {
-			framework.Logf("deleting claim %q/%q", l.pvc.Namespace, l.pvc.Name)
-			// typically this claim has already been deleted
-			err = l.cs.CoreV1().PersistentVolumeClaims(l.pvc.Namespace).Delete(context.TODO(), l.pvc.Name, metav1.DeleteOptions{})
-			if err != nil && !apierrors.IsNotFound(err) {
-				framework.Failf("Error deleting claim %q. Error: %v", l.pvc.Name, err)
+		// This is a bit of a hacky way to do cleanup.
+		// Don't use after each as they won't be run if test is killed.
```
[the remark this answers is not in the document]
AndiLi99: This comment gets changed in the next commit. My work ended up being refactor -> code new test -> refactor to fix failing tests on CI, so it's a bit confusing.

test/e2e/storage/testsuites/snapshottable.go
AndiLi99: This is actually the wording from the commit before the previous commit. I interpret it to mean as creating objects with default field values. The intention is to just check the creation is done correctly. I think a better wording may be "should create snapshot objects"

test/e2e/storage/testsuites/snapshottable.go
AndiLi99: Please check the next commit. It's confusing because there was more refactoring done after I had added the new test.

AndiLi99:
> [#92555 (comment)](https://github.com/kubernetes/kubernetes/pull/92555#issuecomment-651875959)
> This means the Kubernetes based CI failed to run the job.
> 
> Note that if a job isn't marked Required at the bottom of the page it isn't required, only informational.

Thanks for the clarification.

@mattcary I've rebased to separate the refactor from the new test. It's ready for review now, thanks!

test/e2e/storage/testsuites/snapshottable.go
AndiLi99: Now there is a test pattern which handles "retain" vs "delete" deletion policy. For each It statement it should be testing both policies now. 

Regarding the creating an extra snapshot compared to the original test, I don't believe that is the case, see:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/snapshottable.go#L204

test/e2e/storage/testsuites/snapshottable.go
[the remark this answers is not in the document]
mattcary: Seems like By() is the idiomatic choice here. It doesn't look like framework.Logf provides any extra functionality.

test/e2e/storage/testsuites/snapshottable.go
mattcary: From what I understand, in the original test (https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/snapshottable.go), a snapshot is created in TestSnapshottable, which is run once each on the It()s on lines 189 and 204 when testing delete and retain policies (two snapshots total).

If I'm understanding your refactored code, the delete policy is hoisted up to a testpattern. A snapshot is created in CreateSnapshotResource. That happens before each of the It()s on lines 205 and 218 in https://github.com/kubernetes/kubernetes/blob/9b0dff7686228e2b6f514cdd496e2ae6ab236b9b/test/e2e/storage/testsuites/snapshottable.go. So to test both delete and retain, this code is run through two testpatterns (?) and a total of 4 snapshots are created.

Maybe I've missed something?

test/e2e/storage/testsuites/snapshottable.go
```diff
@@ -109,310 +112,334 @@ func (s *snapshottableTestSuite) DefineTests(driver TestDriver, pattern testpatt
...
+func (sr *SnapshotResource) CleanupResource() error {
+	var err error
+	var cleanupErrs []error
+
+	dc := sr.Config.Framework.DynamicClient
+
+	if sr.Vs != nil {
+		framework.Logf("deleting snapshot %q/%q", sr.Vs.GetNamespace(), sr.Vs.GetName())
+
+		sr.Vs, err = dc.Resource(SnapshotGVR).Namespace(sr.Vs.GetNamespace()).Get(context.TODO(), sr.Vs.GetName(), metav1.GetOptions{})
```
mattcary, line 418: I think this CleanupResource() organization is a big win. But I want to make sure I understand what's going on.

It looks like the semantics have changed here: if a VolumeSnapshot isn't found, the previous test would fail, whereas this one would continue (see https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/snapshottable.go#L335 I think).

Is that because this may not have been called after WaitForSnapshotReady? Or another reason?

test/e2e/storage/testsuites/snapshottable.go
```diff
@@ -236,7 +237,53 @@ func (s *snapshottableTestSuite) DefineTests(driver TestDriver, pattern testpatt
...
+				restoredPod = StartInPodWithVolume(cs, restoredPVC.Namespace, restoredPVC.Name, "restored-pvc-tester", "sleep 300", config.ClientNodeSelection)
+				framework.ExpectNoError(e2epod.WaitForPodRunningInNamespaceSlow(cs, restoredPod.Name, restoredPod.Namespace))
+				cleanupSteps = append(cleanupSteps, func() {
+					StopPod(cs, restoredPod)
+				})
+
+				command = "cat /mnt/test/data"
+				actualData, err := utils.PodExec(f, restoredPod, command)
+				framework.ExpectNoError(err)
+				framework.ExpectEqual(actualData, "hello world")
```
mattcary, line 285: hello world vs junkdata?

At any rate better to put the data written as a string constant.

test/e2e/storage/testsuites/snapshottable.go
AndiLi99: Take a look at https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/snapshottable.go#L321 and https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/snapshottable.go#L264.

I believe both TestSnapshotDeleted and TestSnapshottable create their own snapshot, and snapshot class. So that would be a total of 4 before and after the change.

test/e2e/storage/testsuites/snapshottable.go
AndiLi99: The big idea with CleanupResource is that it's purpose should be to clean up as much as it can to prevent resource leaks and be resilient to weird conditions, instead of acting as a way to test the deletion works.

The semantics of deletion should be captured by the deletion test: https://github.com/kubernetes/kubernetes/blob/9b0dff7686228e2b6f514cdd496e2ae6ab236b9b/test/e2e/storage/testsuites/snapshottable.go#L205. If the semantics of the test there has changed that would be an issue. If the volume snapshot was not found in that test then https://github.com/kubernetes/kubernetes/blob/9b0dff7686228e2b6f514cdd496e2ae6ab236b9b/test/e2e/storage/testsuites/snapshottable.go#L281 would throw an error.

test/e2e/storage/testsuites/snapshottable.go
AndiLi99: Right the data written into the pvc afterward is "junkdata". The expected value should be "hello world", as that is what is there when the snapshot was taken. I agree it should probably be a string constant.

test/e2e/storage/testsuites/snapshottable.go
mattcary: ah, right you are. Thanks.

test/e2e/storage/testsuites/snapshottable.go
mattcary: but line 281 won't fire if the resource is not found---don't we want that to be a test error?

+1 on making CleanupResource robust and doing the testing in the test body, that's very nice organization.

test/e2e/storage/testsuites/snapshottable.go
mattcary: Ah, yes, "hello world" comes from the init function. Yes, definitely a constant should be used to keep things consistent.

test/e2e/storage/testsuites/snapshottable.go
AndiLi99: I suppose you're right. Although CreateSnapshotResource does wait for the snapshot to be ready before it returns, and will error if it doesnt get created. https://github.com/kubernetes/kubernetes/blob/9b0dff7686228e2b6f514cdd496e2ae6ab236b9b/test/e2e/storage/testsuites/snapshottable.go#L332

I suppose implicitly relying on the creation step to guarantee the snapshot exists prior to deletion would be confusing. It may just be easier for the DeleteAndWaitSnapshot method to not ignore `not found` errors.

test/e2e/storage/testsuites/snapshottable.go
AndiLi99: I've added a commit to add a string constant.

test/e2e/storage/testsuites/snapshottable.go
AndiLi99: I changed the behavior of DeleteAndWaitSnapshot to not ignore not found errors, and to return the error instead of failing. The test now checks that the deletion succeeds, so it should catch any not found errors.

test/e2e/storage/testsuites/snapshottable.go
```diff
@@ -109,310 +113,386 @@ func (s *snapshottableTestSuite) DefineTests(driver TestDriver, pattern testpatt
...
+				vs = sr.Vs
+				vscontent = sr.Vscontent
+				vsc = sr.Vsclass
+				cleanupSteps = append(cleanupSteps, func() {
+					framework.ExpectNoError(sr.CleanupResource())
+				})
+			})
+
+			ginkgo.It("should delete the VolumeSnapshotContent according to its deletion policy", func() {
+				err = DeleteAndWaitSnapshot(dc, vs.GetNamespace(), vs.GetName(), framework.Poll, framework.SnapshotDeleteTimeout)
```
saikat-royc, line 209: For a retain test case wont this function return err != nil ? line 210 checks unconditionally ExpectNoError?

saikat-royc: to me this looks simpler (because DeleteAndWaitSnapshot already calls WaitForGVRDeletion()): 

err = DeleteAndWaitSnapshot(...)
switch (policy) {
case delete: framework.ExpectNoError(err)
case retain: framework.ExpectError(err)
}

test/e2e/storage/testsuites/snapshottable.go
```diff
@@ -109,310 +113,386 @@ func (s *snapshottableTestSuite) DefineTests(driver TestDriver, pattern testpatt
...
+				}, config.Framework.Namespace.Name)
+
+				group := "snapshot.storage.k8s.io"
+				dataSourceRef := &v1.TypedLocalObjectReference{
+					APIGroup: &group,
+					Kind:     "VolumeSnapshot",
+					Name:     vs.GetName(),
+				}
+
+				restoredPVC.Spec.DataSource = dataSourceRef
```
saikat-royc, line 267: nits: separate dataSourceRef variable not needed.

test/e2e/storage/testsuites/snapshottable.go
```diff
@@ -109,310 +113,386 @@ func (s *snapshottableTestSuite) DefineTests(driver TestDriver, pattern testpatt
...
+			})
+			ginkgo.It("should restore from snapshot with saved data after modifying source data", func() {
+				var restoredPVC *v1.PersistentVolumeClaim
+				var restoredPod *v1.Pod
+				modifiedMntTestData := fmt.Sprintf("modified data from %s namespace", pvc.GetNamespace())
+
+				ginkgo.By("modifying the data in the source PVC")
+
+				command := fmt.Sprintf("echo '%s' > /mnt/test/data", modifiedMntTestData)
+				RunInPodWithVolume(cs, pvc.Namespace, pvc.Name, "pvc-snapshottable-data-tester", command, config.ClientNodeSelection)
```
saikat-royc, line 252: Cant we use the pod "pvc-snapshottable-tester" created in init() to modify the contents of the PV?
Here we are starting yet another pod "pvc-snapshottable-data-tester" to modify the pv contents.
If we re-use the "pvc-snapshottable-tester" the test timings would be less. Thoughts?

test/e2e/storage/testsuites/snapshottable.go
```diff
@@ -109,310 +113,386 @@ func (s *snapshottableTestSuite) DefineTests(driver TestDriver, pattern testpatt
...
-	} else {
-		framework.Failf("Invalid test config. DeletionPolicy should be either Delete or Retain. DeletionPolicy: %v", sct.DeletionPolicy)
+	case testpatterns.PreprovisionedCreatedSnapshot:
+		// prepare a pre-provisioned VolumeSnapshotContent with certain data
+		// Because this could be run with an external CSI driver, we have no way
+		// to pre-provision the snapshot as we normally would using their API.
+		// We instead dynamically take a snapshot and create another snapshot using
+		// the first snapshot's snapshot handle.
+		ginkgo.Skip("Preprovisioned test not implemented")
+		ginkgo.By("taking a snapshot with deletion policy retain")
```
saikat-royc, line 405: Don't think these are needed. "not implemented" is good enough. We can add details when the test is added.

saikat-royc:
Mostly looks good with minor comments from my side. Good stuff with creating a Create/Cleanup Resource for the snapshots and re-using for each It.

test/e2e/storage/testsuites/snapshottable.go
mattcary: Is there an easy way to talk to the process in that init pod? IIUC it needs to wait until after the snapshot is made before modifying the volume. 

Your point about reducing the timing is good, though, considering all the timeouts we've been having recently. Maybe it would be worth it to create some utility pod infrastructure that could easily communicate back and forth with the test to avoid starting up extra pods.

(all of that out of scope for this PR of course :-)

msau42: We have some tests that do a Pod exec in order to modify data. That may be a method to consider if we don't want to launch a new pod.

AndiLi99: The pod "pvc-snapshottable-tester" created in init() is destroyed in init(), although there's no reason we can't keep it around so that we can reuse it. It looks like it currently takes ten seconds for a pod to become ready, and we would save ten seconds.

I was having trouble finding a way to execute commands in an existing pod, perhaps I can use this https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/util.go#L500 and just ignore the error because I'm not looking for a string in output.

test/e2e/storage/testsuites/snapshottable.go
AndiLi99: True these aren't needed, but it's a scaffolding to remember the steps.

test/e2e/storage/testsuites/snapshottable.go
AndiLi99: Good point

AndiLi99:
Thanks for the reviews!

PR #84678, opened 2019-11-02
Recreate target group when externalTrafficPolicy changes to local

jorge-gasca:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
When the `externalTrafficPolicy` of a service is modified from Cluster to Local, the function ensureTargetGroup will try to change the Target Group's HealthCheckProtocol from TCP to HTTP. This results in an `InvalidConfigurationRequest`. From the [AWS SDK docs](https://docs.aws.amazon.com/sdk-for-go/api/service/elbv2/#ModifyTargetGroupInput): 
```
    // If the protocol of the target group is TCP, you can't modify this setting.
    HealthCheckProtocol *string `type:"string" enum:"ProtocolEnum"
```
This fix adds a check in `ensureLoadBalancerv2` to recreate the target group if the HealthCheckProtocol is changing from TCP to anything else. 

**Which issue(s) this PR fixes**:
Fixes #80996

**Special notes for your reviewer**:
First PR to Kubernetes 👍

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #92556, opened 2020-06-26
AWS NLB/CLB health check config based on service annotations

### LINKED ISSUE #80996: Changing an NLB service from externalTrafficPolicy Cluster to Local results in a stuck Service
**What happened**:

I had an NLB Service object (in AWS naturally) where we decided we needed the source IP address and so changed the ExternalTrafficPolicy from Cluster to Local, but didn't delete the Service object - and didn't want to, because that would break Route53 DNS entries.

The Service started throwing Warning and 400 errors from the AWS API and did not self-heal.
```
  Warning  CreatingLoadBalancerFailed  9m17s               service-controller  Error creating load balancer (will retry): failed to ensure load balancer for service XXX/YYY: Error modifying target group health check: "InvalidConfigurationRequest: You cannot change the health check protocol for a target group with the TCP protocol\n\tstatus code: 400, request id: f8e9446d-b7d3-11e9-886c-3f71b1696d69"
```

The controller should delete out the invalid target group and recreate it in order to configure the NLB as requested.

* Create a LoadBalancer Service annotated as an NLB with default ExternalTrafficPolicy, backed by some running pods.
* Wait for the NLB to be fully provisioned.
* Change the service to have ExternalTrafficPolicy Local.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.13 (eks latest version)
- Cloud provider or hardware configuration: AWS
- OS (e.g: `cat /etc/os-release`): unknown - eks control plane
- Kernel (e.g. `uname -a`): unknown - eks control plane
- Install tools: eksctl
- Network plugin and version (if this is a network-related bug): unknown - eks control plane
- Others:

### LINKED ISSUE #81469: Override NLB health check values based on annotations
**What happened**:
Please refer: https://github.com/kubernetes/cloud-provider-aws/issues/51

I set the proper annotations, as per the documentation in Kubernetes, to the service of `type: LoadBalancer`  using an AWS NLB:

```
service.beta.kubernetes.io/aws-load-balancer-connection-draining-timeout: "30"
service.beta.kubernetes.io/aws-load-balancer-healthcheck-healthy-threshold: "3"
service.beta.kubernetes.io/aws-load-balancer-healthcheck-unhealthy-threshold: "2"
service.beta.kubernetes.io/aws-load-balancer-healthcheck-interval: "5"
service.beta.kubernetes.io/aws-load-balancer-healthcheck-timeout: "5"
```

However, these settings do not have any effect on the code as per the code [here](https://github.com/kubernetes/cloud-provider-aws/blob/8f6e1e3d2cfa20a0deac088b9c4022d433500369/pkg/cloudprovider/providers/aws/aws_loadbalancer.go#L390-L400):

```
                  input := &elbv2.CreateTargetGroupInput{
			VpcId:                      aws.String(vpcID),
			Name:                       aws.String(name),
			Port:                       aws.Int64(mapping.TrafficPort),
			Protocol:                   aws.String("TCP"),
			TargetType:                 aws.String("instance"),
			HealthCheckIntervalSeconds: aws.Int64(30),
			HealthCheckPort:            aws.String("traffic-port"),
			HealthCheckProtocol:        aws.String("TCP"),
			HealthyThresholdCount:      aws.Int64(3),
			UnhealthyThresholdCount:    aws.Int64(3),
		}
```

The configured settings should supersede the default settings.

Create a service of `type: LoadBalancer` with an NLB in AWS and check the Health Check configured in AWS.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.11.9
- Cloud provider or hardware configuration: AWS EKS
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

### LINKED ISSUE #82595: AWS: Network Load Balancer ignores annotations upon creation
**What happened**:
Code for handling annotations for load balancer is executed only upon load balancer modification, not creation in all versions including master branch.

So, for example, it's not possible to enable cross-zone load balancing upon creation. Only create-then-update process works.
- [1.14.6](https://github.com/kubernetes/kubernetes/blob/v1.14.6/pkg/cloudprovider/providers/aws/aws_loadbalancer.go#L344)
- [1.15.3](https://github.com/kubernetes/kubernetes/blob/v1.15.3/staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go#L326)
- [master](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go#L336)

Annotations  should be parsed both upon creation and update of load balancer

Code, that parses annotations should be moved outside `else` block
- [1.14.6](https://github.com/kubernetes/kubernetes/blob/v1.14.6/pkg/cloudprovider/providers/aws/aws_loadbalancer.go#L179)
- [1.15.3](https://github.com/kubernetes/kubernetes/blob/v1.15.3/staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go#L164)
- [master](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go#L174)

or duplicated in `if` block

1. create `Service.spec.type.LoadBalancer` with:
```yaml
annotations:
  service.beta.kubernetes.io/aws-load-balancer-type: "nlb"
  service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: "true"
```
 cross-zone load balancing will be off

2. modify something in `Service` definition — CZLB will be on

**Environment**:
- Kubernetes version (use `kubectl version`): 1.14.0—1.14.6, 1.15.0—1.15.3, master
- Cloud provider or hardware configuration: AWS

### LINKED ISSUE #86624: How do I modify the health check interval of aws NLB TargetGroup
**What happened**:
I created a service (type=loadBalancer, service.beta.kubernetes.io/aws-load-balancer-type: nlb), but I found that can't change HealthCheckIntervalSeconds.

In my opinion, when ingress nginx is updated, it may appear that the pod of ingress nginx has been terminated, but the health-check of NLB has not been completed, and there may be traffic transferred to the pod of ingress nginx that has been terminated. So I want to change the NLB target group HealthCheckIntervalSeconds to 10, to keep them consistent

- 1.ingress nginx
```
          livenessProbe:
            failureThreshold: 3
            httpGet:
              path: /healthz
              port: 10254
              scheme: HTTP
            initialDelaySeconds: 10
            periodSeconds: 10
            successThreshold: 1
            timeoutSeconds: 10
          readinessProbe:
            failureThreshold: 3
            httpGet:
              path: /healthz
              port: 10254
              scheme: HTTP
            periodSeconds: 10
            successThreshold: 1
            timeoutSeconds: 10
```
- 2.nlb targetgroup
image

My service configuration
```
kind: Service
apiVersion: v1
metadata:
  name: ingress-nginx
  namespace: ingress-nginx
  labels:
    app.kubernetes.io/name: ingress-nginx
    app.kubernetes.io/part-of: ingress-nginx
  annotations:
    # Enable PROXY protocol
    service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: "*"
    # by default the type is elb (classic load balancer).
    service.beta.kubernetes.io/aws-load-balancer-type: nlb
    service.beta.kubernetes.io/aws-load-balancer-internal: "true"
spec:
  # this setting is to make sure the source IP address is preserved.
  externalTrafficPolicy: Local
  type: LoadBalancer
  selector:
    app.kubernetes.io/name: ingress-nginx
    app.kubernetes.io/part-of: ingress-nginx
  ports:
    - name: http
      port: 80
      targetPort: http
    - name: https
      port: 443
      targetPort: https
```

I found through [aws_loadbalancer.go](https://github.com/kubernetes/legacy-cloud-providers/blob/master/aws/aws_loadbalancer.go#L532) that it was set to 30s, but I didn't find anything that could be changed.

Excuse me, when I am in Kubernetes through service to create change HealthCheckIntervalSeconds NLB does not support?

**Environment**:
- Kubernetes version (use `kubectl version`): v1.14.6
- Cloud provider or hardware configuration: aws
- OS (e.g: `cat /etc/os-release`): Ubuntu 16.04.6 LTS
- Kernel (e.g. `uname -a`):`Linux ip-172-26-110-194 4.15.0-72-generic #81~16.04.1-Ubuntu SMP Tue Nov 26 16:34:21 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux`
- Install tools:
- Network plugin and version (if this is a network-related bug): aws-cni
- Others:

- `/sig cloud-provider`

### LINKED ISSUE #92321: Make the AWS NLB health check configurable
/sig aws
/kind feature
Additional  annotations for configuring the AWS NLB health checks
**Why is this needed**:
Currently, the health check for AWS NLB uses default values and not configurable. Support additional service annotations to specify the protocol, path, port for the AWS NLB health checks.

kishorj:
**What type of PR is this?**

/kind feature

Currently the AWS NLB uses hardcoded health check configuration. The classic LB also doesn't support configurable protocol/port/path for health check. This PR has support for additional service annotations to specify the health check protocol, port and the path for all supported LB types. The existing annotations for health check configurations are also honored for the NLB.
* `service.beta.kubernetes.io/aws-load-balancer-healthcheck-protocol` to specify the protocol. Default is TCP for Local traffic policy and HTTP for cluster traffic policy. NLB allows TCP/HTTP/HTTPS, ELB allows TCP/SSL/HTTP/HTTPS. 
* `service.beta.kubernetes.io/aws-load-balancer-healthcheck-port` for the port. Defaults to `traffic-port` for Cluster, `healthCheckNodePort` for Local policy. 
* `service.beta.kubernetes.io/aws-load-balancer-healthcheck-path` specify the path for HTTP/HTTPS protocol. Default value is "/" for Cluster mode, "/healthz" for Local mode. If healthcheck protocol is TCP path gets ignored.

NLB will also use the honor the interval and threshold annotations. 

In case of NLB target groups, the protocol and interval configurations are not modifiable. So if changes to those params are detected, the existing target group gets deleted and re-created with the new health check config. 

Fixes #92321
Fixes #80996
Fixes #86624
Fixes #82595
Fixes #81469

Tested via kops that the custom health check config gets honored for NLB/ELB, external traffic policy can be changed from Cluster to Local and vice versa. The AWS NLB doesn't support custom healthcheck timeout for TCP listeners, so default value gets used.
Test summary
* Without any annotations
   - CLB cluster, TCP, traffic-port, default timeouts and thresholds
   - CLB local, HTTP, path /healthz, health check node port, default timeouts and thresholds.
   - NLB cluster, TCP, traffic-port, interval 30, thresholds (healthy/unhealthy) 3
   - NLB local, HTTP, path /healthz, health check node port, interval 10, thresholds 2
   - On modification of external traffic policy, health check of the resultant mode is configured. In case of NLB new target groups get created
* With all health check annotations
    - Health check configuration is based on the annotation values regardless of extrnal traffic policy for both NLB and CLB
    - In case of NLB modification of protocol and interval values result in new target groups
    - In case of NLB, timeout value gets ignored. Timeout is 6 seconds for HTTP, 10 seconds otherwise
* Individual Health check annotations
     - The remaining parameters take the default values
     - CLB cluster, if health check protocol is HTTP, port is the traffic port, path is /, timeout 5s, interval 10s, unhealthy thr 6, healthy thr 2
      - NLB local, if HC protocol is TCP, port is health check nodeport, interval 10s, healthy/unhealthy thresholds 2
      - NLB cluster, if HC protocol is HTTP, port is traffic-port, interval 30, healthy/unhealthy thresholds 3
      - NLB/CLB, path gets ignored for TCP
* Mix of annotations
      - Unspecified parameters take the default value
      - HC annotations get honored switching from Local to Cluster external traffic policy
      - For NLB, port, path and thresholds can be edited on the same target group
      - For NLB, protocol and interval modification results in new target groups

```release-note
Allow configuring AWS LoadBalancer health check protocol via service annotations
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
+++ b/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
@@ -187,9 +187,25 @@ const ServiceAnnotationLoadBalancerBEProtocol = "service.beta.kubernetes.io/aws-
 // For example: "Key1=Val1,Key2=Val2,KeyNoVal1=,KeyNoVal2"
 const ServiceAnnotationLoadBalancerAdditionalTags = "service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags"
 
+// ServiceAnnotationLoadBalancerHealthCheckProtocol is the annotation used on the service to
+// specify the protocol used for the ELB health check. Supported values are TCP, HTTP, HTTPS
+// Default is TCP if externalTrafficPolicy is Cluster, HTTP if externalTrafficPolicy is Local
+const ServiceAnnotationLoadBalancerHealthCheckProtocol = "service.beta.kubernetes.io/aws-load-balancer-healthcheck-protocol"
+
+// ServiceAnnotationLoadBalancerHealthCheckPort is the annotation used on the service to
+// specify the port used for ELB health check.
+// Default is traffic-port if externalTrafficPolicy is Cluster, healthCheckNodePort if externalTrafficPolicy is Local
+const ServiceAnnotationLoadBalancerHealthCheckPort = "service.beta.kubernetes.io/aws-load-balancer-healthcheck-port"
+
+// ServiceAnnotationLoadBalancerHealthCheckPath is the annotation used on the service to
+// specify the path for the ELB health check when the health check protocol is HTTP/HTTPS
+// Defaults to /healthz if externalTrafficPolicy is Local, / otherwise
+const ServiceAnnotationLoadBalancerHealthCheckPath = "service.beta.kubernetes.io/aws-load-balancer-healthcheck-path"
+
 // ServiceAnnotationLoadBalancerHCHealthyThreshold is the annotation used on
 // the service to specify the number of successive successful health checks
-// required for a backend to be considered healthy for traffic.
+// required for a backend to be considered healthy for traffic. For NLB, healthy-threshold
+// and unhealthy-threshold must be equal.
 const ServiceAnnotationLoadBalancerHCHealthyThreshold = "service.beta.kubernetes.io/aws-load-balancer-healthcheck-healthy-threshold"
 
 // ServiceAnnotationLoadBalancerHCUnhealthyThreshold is the annotation used
@@ -3686,6 +3702,91 @@ func (c *Cloud) getSubnetCidrs(subnetIDs []string) ([]string, error) {
 	return cidrs, nil
 }
 
+func parseStringAnnotation(annotations map[string]string, annotation string, value *string) bool {
+	if v, ok := annotations[annotation]; ok {
+		*value = v
+		return true
+	}
+	return false
+}
+
+func parseInt64Annotation(annotations map[string]string, annotation string, value *int64) (bool, error) {
+	if v, ok := annotations[annotation]; ok {
+		parsed, err := strconv.ParseInt(v, 10, 0)
+		if err != nil {
+			return true, fmt.Errorf("failed to parse annotation %v=%v", annotation, v)
+		}
+		*value = parsed
+		return true, nil
+	}
+	return false, nil
+}
+
+func (c *Cloud) buildNLBHealthCheckConfiguration(svc *v1.Service) (healthCheckConfig, error) {
+	hc := healthCheckConfig{
+		Port:               defaultHealthCheckPort,
+		Path:               defaultHealthCheckPath,
+		Protocol:           elbv2.ProtocolEnumTcp,
+		Interval:           defaultNlbHealthCheckInterval,
+		Timeout:            defaultNlbHealthCheckTimeout,
+		HealthyThreshold:   defaultNlbHealthCheckThreshold,
+		UnhealthyThreshold: defaultNlbHealthCheckThreshold,
+	}
+	if svc.Spec.ExternalTrafficPolicy == v1.ServiceExternalTrafficPolicyTypeLocal {
+		path, port := servicehelpers.GetServiceHealthCheckPathPort(svc)
+		hc = healthCheckConfig{
+			Port:               strconv.Itoa(int(port)),
+			Path:               path,
+			Protocol:           elbv2.ProtocolEnumHttp,
+			Interval:           10,
+			Timeout:            10,
+			HealthyThreshold:   2,
+			UnhealthyThreshold: 2,
+		}
+	}
+	if parseStringAnnotation(svc.Annotations, ServiceAnnotationLoadBalancerHealthCheckProtocol, &hc.Protocol) {
+		hc.Protocol = strings.ToUpper(hc.Protocol)
+	}
+	switch hc.Protocol {
+	case elbv2.ProtocolEnumHttp, elbv2.ProtocolEnumHttps:
+		parseStringAnnotation(svc.Annotations, ServiceAnnotationLoadBalancerHealthCheckPath, &hc.Path)
+	case elbv2.ProtocolEnumTcp:
+		hc.Path = ""
+	default:
+		return healthCheckConfig{}, fmt.Errorf("Unsupported health check protocol %v", hc.Protocol)
+	}
+
+	parseStringAnnotation(svc.Annotations, ServiceAnnotationLoadBalancerHealthCheckPort, &hc.Port)
+
+	if _, err := parseInt64Annotation(svc.Annotations, ServiceAnnotationLoadBalancerHCInterval, &hc.Interval); err != nil {
+		return healthCheckConfig{}, err
+	}
+	if _, err := parseInt64Annotation(svc.Annotations, ServiceAnnotationLoadBalancerHCTimeout, &hc.Timeout); err != nil {
+		return healthCheckConfig{}, err
+	}
+	if _, err := parseInt64Annotation(svc.Annotations, ServiceAnnotationLoadBalancerHCHealthyThreshold, &hc.HealthyThreshold); err != nil {
+		return healthCheckConfig{}, err
+	}
+	if _, err := parseInt64Annotation(svc.Annotations, ServiceAnnotationLoadBalancerHCUnhealthyThreshold, &hc.UnhealthyThreshold); err != nil {
+		return healthCheckConfig{}, err
+	}
+
+	if hc.HealthyThreshold != hc.UnhealthyThreshold {
+		return healthCheckConfig{}, fmt.Errorf("Health check healthy threshold and unhealthy threshold must be equal")
+	}
+
+	if hc.Interval != 10 && hc.Interval != 30 {
+		return healthCheckConfig{}, fmt.Errorf("Invalid health check interval '%v', must be either 10 or 30", hc.Interval)
+	}
+
+	if hc.Port != defaultHealthCheckPort {
+		if _, err := strconv.ParseInt(hc.Port, 10, 0); err != nil {
+			return healthCheckConfig{}, fmt.Errorf("Invalid health check port '%v'", hc.Port)
+		}
+	}
+	return hc, nil
+}
+
 // EnsureLoadBalancer implements LoadBalancer.EnsureLoadBalancer
 func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiService *v1.Service, nodes []*v1.Node) (*v1.LoadBalancerStatus, error) {
 	annotations := apiService.Annotations
@@ -3724,11 +3825,10 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
 				FrontendProtocol: string(port.Protocol),
 				TrafficPort:      int64(port.NodePort),
 				TrafficProtocol:  string(port.Protocol),
-
-				// if externalTrafficPolicy == "Local", we'll override the
-				// health check later
-				HealthCheckPort:     int64(port.NodePort),
-				HealthCheckProtocol: elbv2.ProtocolEnumTcp,
+			}
+			var err error
+			if portMapping.HealthCheckConfig, err = c.buildNLBHealthCheckConfiguration(apiService); err != nil {
+				return nil, err
 			}
 
 			certificateARN := annotations[ServiceAnnotationLoadBalancerCertificate]
@@ -3776,15 +3876,6 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
 	}
 
 	if isNLB(annotations) {
-
-		if path, healthCheckNodePort := servicehelpers.GetServiceHealthCheckPathPort(apiService); path != "" {
-			for i := range v2Mappings {
-				v2Mappings[i].HealthCheckPort = int64(healthCheckNodePort)
-				v2Mappings[i].HealthCheckPath = path
-				v2Mappings[i].HealthCheckProtocol = elbv2.ProtocolEnumHttp
-			}
-		}
-
 		// Find the subnets that the ELB will live in
 		subnetIDs, err := c.findELBSubnets(internalELB)
 		if err != nil {
@@ -4041,23 +4132,26 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
 		}
 	}
 
+	// We only configure a TCP health-check on the first port
+	var tcpHealthCheckPort int32
+	for _, listener := range listeners {
+		if listener.InstancePort == nil {
+			continue
+		}
+		tcpHealthCheckPort = int32(*listener.InstancePort)
+		break
+	}
 	if path, healthCheckNodePort := servicehelpers.GetServiceHealthCheckPathPort(apiService); path != "" {
 		klog.V(4).Infof("service %v (%v) needs health checks on :%d%s)", apiService.Name, loadBalancerName, healthCheckNodePort, path)
+		if annotations[ServiceAnnotationLoadBalancerHealthCheckPort] == defaultHealthCheckPort {
+			healthCheckNodePort = tcpHealthCheckPort
+		}
 		err = c.ensureLoadBalancerHealthCheck(loadBalancer, "HTTP", healthCheckNodePort, path, annotations)
 		if err != nil {
 			return nil, fmt.Errorf("Failed to ensure health check for localized service %v on node port %v: %q", loadBalancerName, healthCheckNodePort, err)
 		}
 	} else {
 		klog.V(4).Infof("service %v does not need custom health checks", apiService.Name)
-		// We only configure a TCP health-check on the first port
-		var tcpHealthCheckPort int32
-		for _, listener := range listeners {
-			if listener.InstancePort == nil {
-				continue
-			}
-			tcpHealthCheckPort = int32(*listener.InstancePort)
-			break
-		}
 		annotationProtocol := strings.ToLower(annotations[ServiceAnnotationLoadBalancerBEProtocol])
 		var hcProtocol string
 		if annotationProtocol == "https" || annotationProtocol == "ssl" {
--- a/staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go
+++ b/staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go
@@ -56,10 +56,15 @@ const (
 
 var (
 	// Defaults for ELB Healthcheck
-	defaultHCHealthyThreshold   = int64(2)
-	defaultHCUnhealthyThreshold = int64(6)
-	defaultHCTimeout            = int64(5)
-	defaultHCInterval           = int64(10)
+	defaultElbHCHealthyThreshold   = int64(2)
+	defaultElbHCUnhealthyThreshold = int64(6)
+	defaultElbHCTimeout            = int64(5)
+	defaultElbHCInterval           = int64(10)
+	defaultNlbHealthCheckInterval  = int64(30)
+	defaultNlbHealthCheckTimeout   = int64(10)
+	defaultNlbHealthCheckThreshold = int64(3)
+	defaultHealthCheckPort         = "traffic-port"
+	defaultHealthCheckPath         = "/"
 )
 
 func isNLB(annotations map[string]string) bool {
@@ -76,6 +81,16 @@ func isLBExternal(annotations map[string]string) bool {
 	return false
 }
 
+type healthCheckConfig struct {
+	Port               string
+	Path               string
+	Protocol           string
+	Interval           int64
+	Timeout            int64
+	HealthyThreshold   int64
+	UnhealthyThreshold int64
+}
+
 type nlbPortMapping struct {
 	FrontendPort     int64
 	FrontendProtocol string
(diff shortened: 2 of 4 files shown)


kishorj:
#84678

jsafrane:
/ok-to-test
/assign @M00nF1sh @nckturner 
for AWS load balancer experise

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
```diff
@@ -187,6 +187,21 @@ const ServiceAnnotationLoadBalancerBEProtocol = "service.beta.kubernetes.io/aws-
...
 const ServiceAnnotationLoadBalancerAdditionalTags = "service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags"
 
+// ServiceAnnotationLoadBalancerHealthCheckProtocol is the annotation used on the service to
+// specify the protocol used for the ELB health check. Supported values are TCP, HTTP, HTTPS
+// default is TCP
+const ServiceAnnotationLoadBalancerHealthCheckProtocol = "service.beta.kubernetes.io/aws-load-balancer-healthcheck-protocol"
+
+// ServiceAnnotationLoadBalancerHealthCheckPort is the annotation used on the service to
+// specify the port used for ELB health check.
+// default is traffic-port
```
M00nF1sh, line 197: we should change the comments to match the actual default.
e.g. it's not traffic-port when using clusterTrafficPolicy

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
```diff
@@ -187,6 +187,21 @@ const ServiceAnnotationLoadBalancerBEProtocol = "service.beta.kubernetes.io/aws-
...
+const ServiceAnnotationLoadBalancerHealthCheckPort = "service.beta.kubernetes.io/aws-load-balancer-healthcheck-port"
+
+// ServiceAnnotationLoadBalancerHealthCheckPath is the annotation used on the service to
+// specify the path for the ELB health check when the health check protocol is HTTP/HTTPS
+// default is /
```
M00nF1sh, line 202: we should change the comments to match the actual default.
e.g. it's not / when using LocaltrafficPolicy

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
```diff
@@ -3675,6 +3690,74 @@ func buildListener(port v1.ServicePort, annotations map[string]string, sslPorts
 	return listener, nil
 }
 
+func (c *Cloud) parseNLBHealthCheckAnnotations(portMapping *nlbPortMapping, annotations map[string]string) error {
+	getStringValue := func(annotation string, defaultVal string) string {
```
M00nF1sh, line 3694: humm, maybe better to define some function to make it reusable? similar to https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L3562(however, ideally that should be something like `ParseStringListAnnotation`)

M00nF1sh: Personally i favor interface like below
```
func parseStringAnnotation(annotations map[string]string, annotation string, value *string) bool {
}
func parseInt64Annotation(annotations map[string]string, annotation string, value *int64) bool, error {
}
```
so to use it with default value, you just need to things like 
```
var a int64 = 100
exists, error := parseInt64Annotation(annotations, "some-annotation", &a)
```

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
```diff
@@ -3675,6 +3690,74 @@ func buildListener(port v1.ServicePort, annotations map[string]string, sslPorts
...
+		portMapping.HealthCheckPath = getStringValue(ServiceAnnotationLoadBalancerHealthCheckPath, defaultHealthCheckPath)
+	case elbv2.ProtocolEnumTcp:
+	default:
+		return fmt.Errorf("Invalid Health check protocol %v", portMapping.HealthCheckProtocol)
+	}
+
+	portStr := getStringValue(ServiceAnnotationLoadBalancerHealthCheckPort, defaultHealthCheckPort)
+
+	if portStr != "traffic-port" {
+		i64, err := strconv.ParseInt(portStr, 10, 0)
```
M00nF1sh, line 3725: i64 should be some meaningful variable like portInt

M00nF1sh:
`service.beta.kubernetes.io/aws-load-balancer-healthcheck-port for the port. Defaults to traffic-port for Local, healthCheckNodePort for Cluster policy.` is incorrect, it should be `Defaults to traffic-port for Cluster, healthCheckNodePort for Local policy.`

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
```diff
@@ -3675,6 +3690,74 @@ func buildListener(port v1.ServicePort, annotations map[string]string, sslPorts
...
+	case elbv2.ProtocolEnumHttp, elbv2.ProtocolEnumHttps:
+		portMapping.HealthCheckPath = getStringValue(ServiceAnnotationLoadBalancerHealthCheckPath, defaultHealthCheckPath)
+	case elbv2.ProtocolEnumTcp:
+	default:
+		return fmt.Errorf("Invalid Health check protocol %v", portMapping.HealthCheckProtocol)
+	}
+
+	portStr := getStringValue(ServiceAnnotationLoadBalancerHealthCheckPort, defaultHealthCheckPort)
+
+	if portStr != "traffic-port" {
```
M00nF1sh, line 3724: if it's set to 'traffic-port' explicitly, the healthcheck is still using the nodePort, is it expected? (the effect should be same, but users can see it's a port number instead of 'traffic-port' in aws console

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
```diff
@@ -3763,8 +3851,13 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
 	}
 
 	if isNLB(annotations) {
-
-		if path, healthCheckNodePort := servicehelpers.GetServiceHealthCheckPathPort(apiService); path != "" {
+		customHealthCheck := func(annotations map[string]string) bool {
```
M00nF1sh, line 3854: humm, why not just have a variable here instead of this lambda function.
```
_, customProtocolEnabled := annotations[ServiceAnnotationLoadBalancerHealthCheckProtocol]
_, customPortEnabled  := annotations[ServiceAnnotationLoadBalancerHealthCheckPort]
_, customPathEnabled  := annotations[ServiceAnnotationLoadBalancerHealthCheckPath]
customHealthCheckEnabled = customProtocolEnabled || customPortEnabled || customPathEnabled
```

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
```diff
@@ -3763,8 +3851,13 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
...
+			_, customProtocol := annotations[ServiceAnnotationLoadBalancerHealthCheckProtocol]
+			_, customPort := annotations[ServiceAnnotationLoadBalancerHealthCheckPort]
+			_, customPath := annotations[ServiceAnnotationLoadBalancerHealthCheckPath]
+			return customProtocol || customPort || customPath
+		}
+		if path, healthCheckNodePort := servicehelpers.GetServiceHealthCheckPathPort(apiService); path != "" && !customHealthCheck(annotations) {
 			for i := range v2Mappings {
```
M00nF1sh, line 3861: btw, i know the code is originally structured like this. but can we move this special handling for local traffic mode into parseNLBHealthCheckAnnotations directly?(and rename it like `buildNLBHealthCheckSetting`)
That way it's easier to unit test the behavior.

staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go
```diff
@@ -212,7 +222,26 @@ func (c *Cloud) ensureLoadBalancerv2(namespacedName types.NamespacedName, loadBa
 				frontendPort := mapping.FrontendPort
 				frontendProtocol := mapping.FrontendProtocol
 				nodePort := mapping.TrafficPort
-
+				healthCheckModify := func(mapping nlbPortMapping, tgrp *elbv2.TargetGroup) bool {
```
M00nF1sh, line 225: why we have lambda function like this instead of just define a healthCheckModified variable directly =.=

staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go
```diff
@@ -212,7 +222,26 @@ func (c *Cloud) ensureLoadBalancerv2(namespacedName types.NamespacedName, loadBa
...
+						return true
+					}
+					return false
+				}
+				if listener, ok := actual[frontendPort][frontendProtocol]; ok {
+					if targetGroup, ok := nodePortTargetGroup[nodePort]; ok {
+						if healthCheckModify(mapping, targetGroup) {
+							// If health check configuration changed, delete the existing listener so that a new
+							// listener and target group can get created with the new configuration
+							if err := c.deleteListenerV2(listener); err != nil {
```
M00nF1sh, line 237: this looks hacky...since the listener don't need to be deleted. (only targetGroup need to be recreated)

kishorj:
> `service.beta.kubernetes.io/aws-load-balancer-healthcheck-port for the port. Defaults to traffic-port for Local, healthCheckNodePort for Cluster policy.` is incorrect, it should be `Defaults to traffic-port for Cluster, healthCheckNodePort for Local policy.`

Correct. I will update the description.

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
kishorj: sure, will do

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
kishorj: I'll update it.

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
kishorj: thanks for the recommendation.

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
kishorj: ok

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
kishorj: HealthCheckPort is stored as int64 in nlbPortMapping, so the conversion. I can change it to traffic-port if it is more helpful to the end users.

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
kishorj: sounds good

staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go
kishorj: Most of the health check parameters are immutable, and not included in the target group naming. If any of the health check settings change, the new target group creation woudl fail (due to an existing one with the same name). Let me see what it would take to avoid deleting the listener.

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
```diff
@@ -3675,6 +3691,92 @@ func buildListener(port v1.ServicePort, annotations map[string]string, sslPorts
...
+			HealthyThreshold:   2,
+			UnhealthyThreshold: 2,
+		}
+	}
+	if parseStringAnnotation(svc.Annotations, ServiceAnnotationLoadBalancerHealthCheckProtocol, &hc.Protocol) {
+		hc.Protocol = strings.ToUpper(hc.Protocol)
+	}
+	switch hc.Protocol {
+	case elbv2.ProtocolEnumHttp, elbv2.ProtocolEnumHttps:
+		if parseStringAnnotation(svc.Annotations, ServiceAnnotationLoadBalancerHealthCheckPath, &hc.Path) {
```
M00nF1sh, line 3741: nit, can just be parseStringAnnotation(svc.Annotations, ServiceAnnotationLoadBalancerHealthCheckPath, &hc.Path)

kishorj: correct. I had something else in the condition, forgot to cleanup.

staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go
```diff
@@ -241,10 +253,17 @@ func (c *Cloud) ensureLoadBalancerv2(namespacedName types.NamespacedName, loadBa
...
 					}
 
-					// recreate targetGroup if trafficPort or protocol changed
+					// recreate targetGroup if trafficPort, protocol or HealthCheckProtocol changed
+					healthCheckModified := false
 					targetGroupRecreated := false
 					targetGroup, ok := nodePortTargetGroup[nodePort]
-					if !ok || aws.StringValue(targetGroup.Protocol) != mapping.TrafficProtocol {
+
+					if targetGroup != nil && (!strings.EqualFold(mapping.HealthCheckConfig.Protocol, aws.StringValue(targetGroup.HealthCheckProtocol)) ||
```
M00nF1sh, line 261: why is only Interval and protocol checked here. is only interval and  protocol unmodifiable?

M00nF1sh:
would you help list the test cases you have done. i'd like the following test cases at least
1. without any HC annotation, the behavior for NLB/CLB should remain same for cluster/local mode
2. with all HC annotation, the behavior for NLB/CLB should be expected
3. with each individual HC annotation, the behavior for NLB/CLB should be expected
4. with some combinations of individual HC annotation,  the behavior for NLB/CLB should be expected

staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go
kishorj: correct

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
kishorj: refactored the code. no longer needed.

staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go
kishorj: I was able to avoid the delete by modifying the target group naming logic.

M00nF1sh:
/override pull-kubernetes-e2e-kops-aws
seems pull-kubernetes-e2e-kops-aws is broken due to known issue https://github.com/kubernetes/kops/issues/9526

kishorj:
Unable to override failing test pull-kubernetes-e2e-kops-aws. Closing this pull request, will create a new PR.

PR #92569, opened 2020-06-27
Tolerate slightly different containerd urls

dims:
See https://github.com/kubernetes/test-infra/pull/18089, https://github.com/containerd/containerd/issues/4348 and https://github.com/kubernetes/test-infra/pull/18091, 

**What type of PR is this?**
/kind bug

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster/gce/gci/configure.sh
+++ b/cluster/gce/gci/configure.sh
@@ -490,7 +490,10 @@ function install-containerd-ubuntu {
   # Override to latest versions of containerd and runc
   systemctl stop containerd
   if [[ ! -z "${UBUNTU_INSTALL_CONTAINERD_VERSION:-}" ]]; then
-    curl -fsSL "https://github.com/containerd/containerd/releases/download/${UBUNTU_INSTALL_CONTAINERD_VERSION}/containerd-${UBUNTU_INSTALL_CONTAINERD_VERSION:1}.linux-amd64.tar.gz" | tar --overwrite -xzv -C /usr/
+    # containerd versions have slightly different url(s), so try both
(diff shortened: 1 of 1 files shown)


cluster/gce/gci/configure.sh
```diff
@@ -490,7 +490,10 @@ function install-containerd-ubuntu {
   # Override to latest versions of containerd and runc
   systemctl stop containerd
   if [[ ! -z "${UBUNTU_INSTALL_CONTAINERD_VERSION:-}" ]]; then
-    curl -fsSL "https://github.com/containerd/containerd/releases/download/${UBUNTU_INSTALL_CONTAINERD_VERSION}/containerd-${UBUNTU_INSTALL_CONTAINERD_VERSION:1}.linux-amd64.tar.gz" | tar --overwrite -xzv -C /usr/
+    # containerd versions have slightly different url(s), so try both
+    ( curl -fsSL "https://github.com/containerd/containerd/releases/download/${UBUNTU_INSTALL_CONTAINERD_VERSION}/containerd-${UBUNTU_INSTALL_CONTAINERD_VERSION:1}-linux-amd64.tar.gz" || \
```
aojea, line 494: too hacky :) ??
```
curl -fsSL "https://github.com/containerd/containerd/releases/download/{v,}${UBUNTU_INSTALL_CONTAINERD_VERSION}/containerd-${UBUNTU_INSTALL_CONTAINERD_VERSION}{-,.}linux-amd64.tar.gz"
```
https://unix.stackexchange.com/a/243141

dims: not sure if this will work as if one of the urls fail, the curl process returns a bad exit code which will cause the job to fail. in other words, this will work only if both files can be downloaded i think

BenTheElder:
/lgtm
/approve
Might consider removing this eventually, since it seems upstream is not going to do this again (intentionally at least)

Issue #91038, opened 2020-05-12
[Umbrella] scheduler PostFilter extension point


Huang-Wei:
## Background

Per discussion in the [draft design doc](https://docs.google.com/document/d/1lBZOA-vjo2m52242q4dZ1L1W-RrNyr14F63PfQ5YUr8/view) and [brainstorming issue](https://github.com/kubernetes/kubernetes/issues/90025), we've come out with a good shape on how scheduler preemption logic can be refactored into.

I've done some evaluation work and decided to go with the option that introducing a PostFilter extension point, which is called after Filter phase when there is no node can fit in the incoming Pod. And the old preemption logic will be moved as a PostFilter plugin implementation.

Generally we will introduce a `PreemptHandle` interface which embed all needed operations during preemption logic, specifically a `PreemptHandle` interface will include several sub-interfaces/functions:

- `PluginRunner` interface. It's for abstracting calling RunXYZPlugins during plugin runtime
- `PodNominater` interface with internal nominatedMap (which is related with setting/clearing `.spec.NominatedNodeName` as well as managing internal scheduler status
- `Extenders()` function. It's for interacting with registered extenders.

And then the `PreemptHandle` will be instantiated properly, and passed as an arg to `PostFilter()` function.

Below are the items going forward. Hope they can be finished in this (1.19) release cycle.

### Part 1 - prefactor current code to make PostFilter extension point possible

- [x] Refactor preemption extender logic and move SchedulerExtender interface to framework pkg #91037
- [x] Extract logic related with scheduler nominatedPods to an interface #90356
- [x] Add function NominatedPodsForNode to PodNominator interface #90415
- [x] Add Extender and PluginsRunner to PreemptHandle #90711

[...]

PR #92571, opened 2020-06-27
Polish unit tests of defaultpreemptio plugin

Huang-Wei:
**What type of PR is this?**

/kind cleanup
/sig scheduling

The PR polishes preemption unit tests by:

1. Use a standard *Wrapper to construct pod/node.
1. Remove/simplify unnecessary logic in main testing blocks.

Additionally, this PR also adds a test to cover `PostFilter` of defaultpreemption plugin.

(After this PR, there will be a final PR to expose/rename some internal functions to end this series.)

Part of #91038

```release-note
NONE
```


chendave (commented):
haven't went through the details, but seem like there is a typo in the commit message :)
s/defaultpreemptio plugin/defaultpreemption plugin

Huang-Wei:
@chendave yup, thanks for pointing it out. I will amend the commit msg along with other fixups.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -226,422 +97,569 @@ func mergeObjs(pod *v1.Pod, pods []*v1.Pod) []runtime.Object {
...
+				},
+			},
 			expectedNumFilterCalled: 5,
 		},
 		{
 			name: "mixed priority pods are preempted, pick later StartTime one when priorities are equal",
 			registerPlugins: []st.RegisterPluginFunc{
-				st.RegisterQueueSortPlugin(queuesort.Name, queuesort.New),
 				st.RegisterPluginAsExtensions(noderesources.FitName, noderesources.NewFit, "Filter", "PreFilter"),
-				st.RegisterBindPlugin(defaultbinder.Name, defaultbinder.New),
```
ahg-g, line 357: why did we need to register bind and queue plugins before?

ahg-g (commented):
would have been easier to review if it was broken into multiple PRs: 1) renaming, 2) wrapping objects 3) adding new test etc.

I did take a look and tried to verify, but can you please confirm that non of the tests changed its expectation? otherwise looks good to me.

Huang-Wei:
@ahg-g Thanks for the review. Apologizing for combining several things together. And yes, there are 3 kinds of changes:

1. renaming - in some places, the word "machine" is renamed to "node" as "machine" always reminds me of a bare metal...
1. wrapping objects - the added wrapper functions are just for simplifying the testing code
1. new test - `TestPostFilter` is newly added, it's a new addition.

So overall, all old testing artifacts are 100% kept and their functions remain unchanged. The addition `TestPostFilter` to cover the path of `PostFilter()`.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
Huang-Wei: As `NewFramework` would report missing mandatory plugins (queuesort or bind):

https://github.com/kubernetes/kubernetes/blob/c53ce48d48372c30053a9b67c6d1dee237b5cd69/pkg/scheduler/framework/runtime/framework.go#L307-L315

PR #92578, opened 2020-06-28
Fix scheduler preemt function comment

zhouya0:
**What type of PR is this?**
 /kind cleanup

/cc @Huang-Wei 

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
+++ b/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
@@ -89,9 +89,7 @@ func (pl *DefaultPreemption) PostFilter(ctx context.Context, state *framework.Cy
 
 // preempt finds nodes with pods that can be preempted to make room for "pod" to
 // schedule. It chooses one of the nodes and preempts the pods on the node and
-// returns 1) the node, 2) the list of preempted pods if such a node is found,
-// 3) A list of pods whose nominated node name should be cleared, and 4) any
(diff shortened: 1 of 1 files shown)


pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -89,9 +89,7 @@ func (pl *DefaultPreemption) PostFilter(ctx context.Context, state *framework.Cy
 
 // preempt finds nodes with pods that can be preempted to make room for "pod" to
 // schedule. It chooses one of the nodes and preempts the pods on the node and
-// returns 1) the node, 2) the list of preempted pods if such a node is found,
-// 3) A list of pods whose nominated node name should be cleared, and 4) any
-// possible error.
+// returns 1) the node, 2) any possible error.
```
chendave, line 92: nice catch! just a soft suggestion, it could be more descriptive, maybe something like 1) the node name which is picked up for preemption, fine as it is.

chendave (commented):
Just a soft suggestion.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
zhouya0: ok, changed :)

Huang-Wei:
Thanks @zhouya0 .

/lgtm
/approve

PR #92579, opened 2020-06-28
Change the exception to avoid the cost of preemption

chendave:
node's labels doesn't contain the required topologyKeys in `Constraints`
cannot be resolved by preempting the pods on that pods.

One use case that could easily reproduce the issue is
- set `alwaysCheckAllPredicates` to true.
- one node contains all the required topologyKeys but is failed in predicates
  such as 'taint'.
- another node doesn't hold all the required topologyKeys, and thus return `Unschedulable`
  status code.
- scheduler will try to preempt the pods on the above node with lower priorities.

**What type of PR is this?**
>
/kind bug

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
+++ b/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
@@ -958,7 +958,7 @@ func TestNodesWherePreemptionMightHelp(t *testing.T) {
 			expected: map[string]bool{"node4": true},
 		},
 		{
-			name: "ErrTopologySpreadConstraintsNotMatch should be tried as it indicates that the pod is unschedulable due to topology spread constraints",
+			name: "ErrReasonConstraintsNotMatch should be tried as it indicates that the pod is unschedulable due to topology spread constraints",
 			nodesStatuses: framework.NodeToStatusMap{
 				"node1": framework.NewStatus(framework.Unschedulable, podtopologyspread.ErrReasonConstraintsNotMatch),
 				"node2": framework.NewStatus(framework.UnschedulableAndUnresolvable, nodename.ErrReason),
@@ -970,6 +970,15 @@ func TestNodesWherePreemptionMightHelp(t *testing.T) {
 			name: "UnschedulableAndUnresolvable status should be skipped but Unschedulable should be tried",
 			nodesStatuses: framework.NodeToStatusMap{
 				"node2": framework.NewStatus(framework.UnschedulableAndUnresolvable, ""),
+				"node3": framework.NewStatus(framework.Unschedulable, ""),
+				"node4": framework.NewStatus(framework.UnschedulableAndUnresolvable, ""),
+			},
+			expected: map[string]bool{"node1": true, "node3": true},
+		},
+		{
+			name: "ErrReasonNodeLabelNotMatch should not be tried as it indicates that the pod is unschedulable due to node doesn't have the required label",
+			nodesStatuses: framework.NodeToStatusMap{
+				"node2": framework.NewStatus(framework.UnschedulableAndUnresolvable, podtopologyspread.ErrReasonNodeLabelNotMatch),
 				"node3": framework.NewStatus(framework.Unschedulable, ""),
 				"node4": framework.NewStatus(framework.UnschedulableAndUnresolvable, ""),
 			},
--- a/pkg/scheduler/framework/plugins/podtopologyspread/filtering.go
+++ b/pkg/scheduler/framework/plugins/podtopologyspread/filtering.go
@@ -295,7 +295,7 @@ func (pl *PodTopologySpread) Filter(ctx context.Context, cycleState *framework.C
 		tpVal, ok := node.Labels[c.TopologyKey]
 		if !ok {
 			klog.V(5).Infof("node '%s' doesn't have required label '%s'", node.Name, tpKey)
-			return framework.NewStatus(framework.Unschedulable, ErrReasonConstraintsNotMatch)
+			return framework.NewStatus(framework.UnschedulableAndUnresolvable, ErrReasonNodeLabelNotMatch)
 		}
 
 		selfMatchNum := int32(0)
--- a/pkg/scheduler/framework/plugins/podtopologyspread/filtering_test.go
+++ b/pkg/scheduler/framework/plugins/podtopologyspread/filtering_test.go
@@ -1145,11 +1145,11 @@ func mustConvertLabelSelectorAsSelector(t *testing.T, ls *metav1.LabelSelector) 
 
 func TestSingleConstraint(t *testing.T) {
 	tests := []struct {
-		name         string
-		pod          *v1.Pod
-		nodes        []*v1.Node
-		existingPods []*v1.Pod
-		fits         map[string]bool
+		name           string
+		pod            *v1.Pod
+		nodes          []*v1.Node
+		existingPods   []*v1.Pod
+		wantStatusCode map[string]framework.Code
 	}{
 		{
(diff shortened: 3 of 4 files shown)


chendave:
cc @Huang-Wei  @alculquicondor @ahg-g

pkg/scheduler/framework/plugins/podtopologyspread/filtering.go
```diff
@@ -295,7 +295,7 @@ func (pl *PodTopologySpread) Filter(ctx context.Context, cycleState *framework.C
 		tpVal, ok := node.Labels[c.TopologyKey]
 		if !ok {
 			klog.V(5).Infof("node '%s' doesn't have required label '%s'", node.Name, tpKey)
-			return framework.NewStatus(framework.Unschedulable, ErrReasonConstraintsNotMatch)
+			return framework.NewStatus(framework.UnschedulableAndUnresolvable, ErrReasonConstraintsNotMatch)
```
Huang-Wei, line 298: LGTM. Can you make it UTed?

chendave: create a unittest?

alculquicondor:
+1 to the fix, but, what do you mean by

> scheduler will try to preempt the pods on the above node with lower priorities and loop infinitely.

That sounds like a separate bug. But maybe you just meant that it unnecessarily did a preemption step.

pkg/scheduler/framework/plugins/podtopologyspread/filtering.go
alculquicondor: yes

ahg-g:
sounds good to me.

chendave:
> +1 to the fix, but, what do you mean by
> 
> > scheduler will try to preempt the pods on the above node with lower priorities and loop infinitely.
> 
> That sounds like a separate bug. But maybe you just meant that it unnecessarily did a preemption step.

cause the pod is pending on scheduling, the scheduler will recheck whether there is a node available for binding after a period of time, then the preemption will happen there again.

pkg/scheduler/framework/plugins/podtopologyspread/filtering.go
chendave: done

alculquicondor:
> cause the pod is pending on scheduling, the scheduler will recheck whether there is a node available for binding after a period of time, then the preemption will happen there again.

Well, that's normal behavior for any pending pod. Please remove that note in your PR description so we don't get confused about the bug in the future.

alculquicondor:
/approve

Also fix golint

chendave:
> > cause the pod is pending on scheduling, the scheduler will recheck whether there is a node available for binding after a period of time, then the preemption will happen there again.
> 
> Well, that's normal behavior for any pending pod. Please remove that note in your PR description so we don't get confused about the bug in the future.

done

chendave:
> /approve
> 
> Also fix golint

Done

pkg/scheduler/framework/plugins/podtopologyspread/filtering_test.go
```diff
@@ -1150,6 +1150,7 @@ func TestSingleConstraint(t *testing.T) {
 		nodes        []*v1.Node
 		existingPods []*v1.Pod
 		fits         map[string]bool
+		wantErrCode  map[string]int
```
alculquicondor, line 1153: please do `map[string]status.Code` and use constants instead of numbers.

chendave: Done, pls help to check if that is what you want.

pkg/scheduler/framework/plugins/podtopologyspread/filtering_test.go
```diff
@@ -1424,6 +1452,7 @@ func TestMultipleConstraints(t *testing.T) {
 		nodes        []*v1.Node
 		existingPods []*v1.Pod
 		fits         map[string]bool
+		wantErrCode  map[string]framework.Code
```
alculquicondor, line 1455: Now that I think of it, we only need `wantStatusCode map[string]framework.Code` to replace both `fits` and `wantErrCode`

pkg/scheduler/framework/plugins/podtopologyspread/plugin.go
```diff
@@ -31,6 +31,8 @@ import (
 const (
 	// ErrReasonConstraintsNotMatch is used for PodTopologySpread filter error.
 	ErrReasonConstraintsNotMatch = "node(s) didn't match pod topology spread constraints"
+	// ErrReasonNodeLabelNotMatch is used when the node label doesn't hold the required label.
```
Huang-Wei, line 34:
```suggestion
	// ErrReasonNodeLabelNotMatch is used when the node doesn't hold the required label.
```

pkg/scheduler/framework/plugins/podtopologyspread/filtering_test.go
Huang-Wei: +1. Please help to update `TestSingleConstraint` as well.

pkg/scheduler/framework/plugins/podtopologyspread/plugin.go
```diff
@@ -31,6 +31,8 @@ import (
...
+	ErrReasonNodeLabelNotMatch = "node doesn't have the required label"
```
Huang-Wei, line 35:
```suggestion
	ErrReasonNodeLabelNotMatch = ErrReasonConstraintsNotMatch + " (missing required label)"
```

chendave: Done

pkg/scheduler/framework/plugins/podtopologyspread/plugin.go
chendave: Done

pkg/scheduler/framework/plugins/podtopologyspread/filtering_test.go
chendave: Done

chendave:
Thanks for all those great suggestion! all comments is addressed.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -1016,6 +1016,15 @@ func TestNodesWherePreemptionMightHelp(t *testing.T) {
 			},
 			expected: map[string]bool{"machine1": true, "machine3": true},
 		},
+		{
+			name: "ErrReasonNodeLabelNotMatch should not be tried as it indicates that the pod is unschedulable due to node doesn't have the required label",
+			nodesStatuses: framework.NodeToStatusMap{
+				"machine2": framework.NewStatus(framework.UnschedulableAndUnresolvable, podtopologyspread.ErrReasonNodeLabelNotMatch),
```
Huang-Wei, line 1022: My recent PR polished this whole test a bit. You may need to rebase first, and change "machine" to "node"; otherwise UT may fail.

Huang-Wei (commented):
Just one nit. LGTM otherwise.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
chendave: will do, thanks for the reminder.

chendave:
/hold
for rebase

chendave:
/hold cancel

@Huang-Wei  PTAL?

Huang-Wei:
/lgtm

Thanks @chendave !

PR #91329, opened 2020-05-21
Switch kube-controller-manager to distroless image

dims:
Signed-off-by: Davanum Srinivas <davanum@gmail.com>

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
We did the same for apiserver/scheduler in https://github.com/kubernetes/kubernetes/pull/90674. Withheld this change there to sync up better with sig-storage folks as this was breaking flexvolumes especially bash/shell scripts as the controller-manager no longer has required binaries to run scripts. Please see discussion in issue https://github.com/kubernetes/kubernetes/issues/78737. The plan is outlined there.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
ACTION REQUIRED : Switch core master base images (kube-controller-manager) from debian to distroless. If you need Flex Volumes support using scripts, please build your own image with required packages (like bash)
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

Issue #91253, opened 2020-05-19
FlexVolume tests are failing


andrewsykim:
**Which jobs are failing**:
pull-kubernetes-e2e-gci-gce-ipvs

**Which test(s) are failing**:
[sig-storage] Flexvolumes should be mountable when attachable
[sig-storage] Flexvolumes should be mountable when non-attachable

**Since when has it been failing**:
https://prow.k8s.io/job-history/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gci-gce-ipvs

**Testgrid link**:
N/A since testgrid filtered tests based on [sig-network]

**Reason for failure**:

**Anything else we need to know**:

athenabot:
/triage unresolved

Comment `/remove-triage unresolved` when the issue is assessed and confirmed.

🤖 I am a bot run by vllry. 👩‍🔬

BenTheElder:
see: https://github.com/kubernetes/kubernetes/pull/91257 for potential short-term fix

BenTheElder:
https://github.com/kubernetes/kubernetes/issues/91242 is probably the right issue to track this in
/triage duplicate
/close

andrewsykim:
Thanks @BenTheElder !

aojea:
/reopen

per @msau42 comment this is an IPVS only problem

https://github.com/kubernetes/kubernetes/pull/92584#issuecomment-652521262

BenTheElder:
@aojea please file a new issue.
This issue was about quay.io being down.

aojea:
> @aojea please file a new issue.
> This issue was about quay.io being down.
ack
/close

Issue #91242, opened 2020-05-19
[Failing Test] Storage tests are failing due to quay.io being down


hasheddan:
**Which jobs are failing**:

`skew-cluster-latest-kubectl-stable1-gce`
`gce-cos-master-default`
`gce-cos-master-alpha-features`
`gci-gce-ingress`
`gce-cos-master-reboot`
`gce-cos-master-scalability-100`
`Conformance - GCE - Master`
`integration-master`
`kind-master-parallel`
`kind-ipv6-master-parallel`

**Which test(s) are failing**:

All relying on volume snapshot controller.

**Since when has it been failing**:

`5/18/20 12:38 AM PT`

**Testgrid link**:

https://testgrid.k8s.io/sig-release-master-blocking

**Reason for failure**:

https://quay.io/ is experiencing downtime, causing failure to pull image in volume snapshot controller deployment: https://github.com/kubernetes/kubernetes/blob/6a0e01880ad7ea4c110040c3be417516c0bf6bca/cluster/addons/volumesnapshots/volume-snapshot-controller/volume-snapshot-controller-deployment.yaml#L26

**Anything else we need to know**:

/cc @kubernetes/ci-signal
/milestone v1.19
/priority critical-urgent
/sig testing
/sig storage

hasheddan:
@xing-yang is there an issue open for moving this quay.io image to community infra?

xing-yang:
@hasheddan we haven't talked about it.  Let me open an issue about it.

hasheddan:
Potential fix here: https://github.com/kubernetes-csi/node-driver-registrar/pull/82

BenTheElder:
/assign @dims 
dupes: https://github.com/kubernetes/kubernetes/issues/91253 https://github.com/kubernetes/kubernetes/issues/91255

BenTheElder:
short term work around https://github.com/kubernetes/kubernetes/pull/91257

BenTheElder:
https://prow.k8s.io/?job=pull-kubernetes-e2e-kind lots of green bouncing back
thanks @dims!

Issue #82125, opened 2019-08-29
fix KUBE-MARK-MASQ ownership (kubelet vs kube-proxy)


danwinship:
The `KUBE-MARK-MASQ` and `KUBE-POSTROUTING` chains are currently created by both kubelet and kube-proxy. There have long been comments in the code suggesting that they should be managed only by kubelet, and we should do that.

History:
  - Marking packets to be masqueraded later first appeared in kube-proxy in #12764.
  - #20496 introduced the modern `KUBE-MARK-MASQ` / `KUBE-POSTROUTING` (set up from `pkg/proxy/iptables`)
  - #25604 added `HostPort` support to kubenet, making use of the proxier's `KUBE-MARK-MASQ` chain but adding a comment to the proxier noting that kubelet should be creating that chain and the proxy should just assume that it exists.
  - #30486 required the addition of the `KUBE-MARK-DROP` chain, and was used as an opportunity to start moving `KUBE-MARK-MASQ` to kubelet. But the code to create the chain was not removed from the proxier, so now kubelet and kube-proxy both create it. (But only kubelet creates `KUBE-MARK-DROP`.) A comment was included in the kubelet code: `TODO: clean up this logic in kube-proxy`.
  - #46580 added `pkg/proxy/ipvs`, which copied a lot of iptables code and so now also creates `KUBE-MARK-MASQ` itself (and depends on kubelet creating `KUBE-MARK-DROP`).
  - #78547 changes `KUBE-MARK-MASQ` slightly, which requires changing the rule in both kubelet and the proxies, which is annoying, leading to me filing this bug

`KUBE-MARK-MASQ` is assumed to exist by a few things outside of kubernetes now too; CRI-O vendors the kubelet `HostPort` code and implicitly assumes that the chain exists and works. The CNI portmap plugin doesn't assume it exists by default but can be configured by users to make use of it (`"externalSetMarkChain": "KUBE-MARK-MASQ"`) so some people are likely doing that.

Given that (and the fact that kube-proxy hasn't suffered at all by depending on having kubelet manage `KUBE-MARK-DROP`) it seems like it is correct for `KUBE-MARK-MASQ` to be owned by kubelet, not kube-proxy.

So, I propose:

[...]

PR #92584, opened 2020-06-28
IPVS: kubelet, kube-proxy: unmark packets before masquerading …

### LINKED ISSUE #92553: Some IPVS e2e tests are failing
**Which jobs are failing**:
pull-kubernetes-e2e-gci-gce-ipvs

Not all of these fail on every run:
[sig-network] Networking Granular Checks: Services should function for endpoint-Service: http
[sig-network] Services should allow pods to hairpin back to themselves through services
[sig-storage] Flexvolumes should be mountable when non-attachable
[sig-network] Networking Granular Checks: Services should function for endpoint-Service: udp
[sig-network] Services should have session affinity timeout work for service with type clusterIP [LinuxOnly] [Conformance]
[sig-network] Services should allow pods to hairpin back to themselves through services

Looking through job history, it looks like the last time any runs passed was on May 2.
 
**Testgrid link**:
N/A since testgrid filtered tests based on [sig-network]

This was discovered by @kumarvin123 in https://github.com/kubernetes/kubernetes/pull/90909, it also appears to be affecting other PRs.

/cc @andrewsykim @aojea 
/sig network

aojea:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation

/kind failing-test

> /kind feature
> /kind flake

The sig-network tests started to fail after the double nat fix.
The problem is that the IPVS proxier uses a POSTORUTING for the hairpin traffic that needs to happen before the
kubernetes specific postrouting rules. Prepending this ipvs only masquerade rule solves the issue.

Fixes #92553 

```release-note
NONE
```

```docs

```

> @aojea catching up on this. So we're reverting the double-nat fix for IPVS only since it used ipset and was not impacted?

### DIFF
--- a/pkg/proxy/ipvs/proxier.go
+++ b/pkg/proxy/ipvs/proxier.go
@@ -1787,6 +1787,39 @@ func (proxier *Proxier) writeIptablesRules() {
 		"-j", "ACCEPT",
 	)
 
+	// Install the kubernetes-specific postrouting rules. We use a whole chain for
+	// this so that it is easier to flush and change, for example if the mark
+	// value should ever change.
+	// NB: THIS MUST MATCH the corresponding code in the kubelet
+	writeLine(proxier.natRules, []string{
+		"-A", string(kubePostroutingChain),
+		"-m", "mark", "!", "--mark", fmt.Sprintf("%s/%s", proxier.masqueradeMark, proxier.masqueradeMark),
+		"-j", "RETURN",
+	}...)
+	// Clear the mark to avoid re-masquerading if the packet re-traverses the network stack.
+	writeLine(proxier.natRules, []string{
+		"-A", string(kubePostroutingChain),
+		// XOR proxier.masqueradeMark to unset it
+		"-j", "MARK", "--xor-mark", proxier.masqueradeMark,
+	}...)
+	masqRule := []string{
+		"-A", string(kubePostroutingChain),
+		"-m", "comment", "--comment", `"kubernetes service traffic requiring SNAT"`,
+		"-j", "MASQUERADE",
+	}
+	if proxier.iptables.HasRandomFully() {
+		masqRule = append(masqRule, "--random-fully")
+	}
+	writeLine(proxier.natRules, masqRule...)
+
+	// Install the kubernetes-specific masquerade mark rule. We use a whole chain for
+	// this so that it is easier to flush and change, for example if the mark
+	// value should ever change.
+	writeLine(proxier.natRules, []string{
+		"-A", string(KubeMarkMasqChain),
+		"-j", "MARK", "--or-mark", proxier.masqueradeMark,
+	}...)
+
 	// Write the end-of-table markers.
 	writeLine(proxier.filterRules, "COMMIT")
 	writeLine(proxier.natRules, "COMMIT")
@@ -1845,38 +1878,6 @@ func (proxier *Proxier) createAndLinkeKubeChain() {
 		}
 	}
 
-	// Install the kubernetes-specific postrouting rules. We use a whole chain for
-	// this so that it is easier to flush and change, for example if the mark
-	// value should ever change.
-	// NB: THIS MUST MATCH the corresponding code in the kubelet
-	writeLine(proxier.natRules, []string{
-		"-A", string(kubePostroutingChain),
-		"-m", "mark", "!", "--mark", fmt.Sprintf("%s/%s", proxier.masqueradeMark, proxier.masqueradeMark),
-		"-j", "RETURN",
-	}...)
-	// Clear the mark to avoid re-masquerading if the packet re-traverses the network stack.
-	writeLine(proxier.natRules, []string{
-		"-A", string(kubePostroutingChain),
-		// XOR proxier.masqueradeMark to unset it
-		"-j", "MARK", "--xor-mark", proxier.masqueradeMark,
-	}...)
-	masqRule := []string{
-		"-A", string(kubePostroutingChain),
-		"-m", "comment", "--comment", `"kubernetes service traffic requiring SNAT"`,
-		"-j", "MASQUERADE",
-	}
-	if proxier.iptables.HasRandomFully() {
-		masqRule = append(masqRule, "--random-fully")
-	}
-	writeLine(proxier.natRules, masqRule...)
-
-	// Install the kubernetes-specific masquerade mark rule. We use a whole chain for
-	// this so that it is easier to flush and change, for example if the mark
-	// value should ever change.
-	writeLine(proxier.natRules, []string{
-		"-A", string(KubeMarkMasqChain),
-		"-j", "MARK", "--or-mark", proxier.masqueradeMark,
-	}...)
 }
 
 // getExistingChains get iptables-save output so we can check for existing chains and rules.


aojea:
/assign @andrewsykim @uablrek 

This solves all the test failures reported in #92553 

```
I0628 16:24:49.389] FAIL! -- 757 Passed | 1 Failed | 0 Pending | 4355 Skipped
```

and gives us back to the original situation reported in, with the flexvolume test failure #91253

```
I0628 16:24:49.388] [Fail] [sig-storage] Flexvolumes [It] should be mountable when non-attachable 
I0628 16:24:49.388] test/e2e/framework/volume/fixtures.go:502
```

@danwinship added this fix to solve the double natting problem, I tried to backtrace the problem but seems that IPVS uses ipset too and I wasn't able to find out the problem , so reverting the change only in IPVS will solve the issue

aojea:
/priority urgent-critical
This breaks kube-proxy in IPVS mode

/test pull-kubernetes-e2e-gci-gce-ipvs

andrewsykim:
@aojea catching up on this. So we're reverting the double-nat fix for IPVS only since it used ipset and was not impacted?

aojea:
> @aojea catching up on this. So we're reverting the double-nat fix for IPVS only since it used ipset and was not impacted?

I don't know if it was impacted, sorry, didn't go so far understanding the ipset - iptables relationship, however, with the fix in  IPVS  is broken per the test results.

I think that we should follow up and verify if the double-nat fix is needed for IPVS, but have it working until then.

aojea:
/retest
it works :smile: 
> pull-kubernetes-e2e-gci-gce-ipvs — Job succeeded.

test/e2e/storage/flexvolume.go
```diff
@@ -182,7 +182,7 @@ var _ = utils.SIGDescribe("Flexvolumes", func() {
 		suffix = ns.Name
 	})
 
-	ginkgo.It("should be mountable when non-attachable", func() {
+	ginkgo.It("should be mountable when non-attachable [Feature:Flexvolumes]", func() {
```
Lion-Wei, line 185: I know this is for storage feature, but still feel inappropriate to directly add this label.

aojea: This test is failing constantly, is breaking this job.

It seems a leftover from https://github.com/kubernetes/kubernetes/pull/91329, the linked issue says that these tests should have the feature tag to avoid breaking the e2e suite
https://github.com/kubernetes/kubernetes/issues/78737#issuecomment-627021732

I've added it as a separate commit so the git history is clean and can be reverted independently.

Lion-Wei:
@aojea I think part of the problem is because `RETURN` rule in `KUBE-POSTROUTING` break the loop back logic. 
```
Chain KUBE-POSTROUTING (1 references)
target     prot opt source               destination
RETURN     all  --  0.0.0.0/0            0.0.0.0/0            mark match ! 0x4000/0x4000
MARK       all  --  0.0.0.0/0            0.0.0.0/0            MARK xor 0x4000
MASQUERADE  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service traffic requiring SNAT */
MASQUERADE  all  --  0.0.0.0/0            0.0.0.0/0            /* Kubernetes endpoints dst ip:port, source ip for solving hairpin purpose */ match-set KUBE-LOOP-BACK dst,dst,src
```

However, I think straightly revert the double-nat fix is not a good idea, maybe move `KUBE-LOOP-BACK` up could be an option.

aojea:
> However, I think straightly revert the double-nat fix is not a good idea, maybe move `KUBE-LOOP-BACK` up could be an option.


agree that the best solution is to solve it, but right now IPVS mode is broken and I don't know if the double-nat issue with the random-fully flag is affecting IPVS, but if it does affect, it's only in some specific scenarios, we can see that all the test succeed now.

Honestly, I don't have more time to investigate :sweat_smile: , seems this logic was added here https://github.com/kubernetes/kubernetes/pull/54219 , my suggestion is to merge this, fix the job and IPVS so we can have a good test signal, and open a new Issue to solve/understand the double-nat issue in IPVS.

Lion-Wei:
> > However, I think straightly revert the double-nat fix is not a good idea, maybe move `KUBE-LOOP-BACK` up could be an option.
> 
> agree that the best solution is to solve it, but right now IPVS mode is broken and I don't know if the double-nat issue with the random-fully flag is affecting IPVS, but if it does affect, it's only in some specific scenarios, we can see that all the test succeed now.
> 
> Honestly, I don't have more time to investigate 😅 , seems this logic was added here #54219 , my suggestion is to merge this, fix the job and IPVS so we can have a good test signal, and open a new Issue to solve/understand the double-nat issue in IPVS.

Agree.

danwinship:
There were definitely people using IPVS in various places in the sprawling tangle of bug reports related to the random-fully/checksum problem.

> I think part of the problem is because `RETURN` rule in `KUBE-POSTROUTING` break the loop back logic.

> maybe move KUBE-LOOP-BACK up could be an option.

er, yeah, my fix assumed that kubelet and kube-proxy were creating identical `KUBE-POSTROUTING` rules.

This will hopefully eventually get fixed as part of #82125 but yes, it seems like for now just moving the `KUBE-LOOP-BACK` rule should fix this

danwinship:
Also note that the unmark PR was backported back to 1.18, 1.17 and 1.16

aojea:
> Also note that the unmark PR was backported back to 1.18, 1.17 and 1.16

/hold
that changes everything then, we have to fix it so we can backport it

andrewsykim:
So I guess this also means IPVS is broken on the the v1.18, v1.17 and 1.16 release branches? 

> it seems like for now just moving the KUBE-LOOP-BACK rule should fix this

Do you think you can update this PR to do this?

andrewsykim:
cc @kubernetes/release-managers

aojea:
> So I guess this also means IPVS is broken on the the v1.18, v1.17 and 1.16 release branches?
> 
> > it seems like for now just moving the KUBE-LOOP-BACK rule should fix this
> 
> Do you think you can update this PR to do this?

@andrewsykim I can't understand the logic to move up the  KUBE-LOOP-BACK , I think that for someone more familiar with the code should be easy

andrewsykim:
I think we want to move some of the masq mark rules in `createAndLinkeKubeChain` to `writeIptablesRules`. Not sure yet if that's the right solution but that would move the ordering of the rules so KUBE-LOOP-BACK is first, let's see if the e2es pass with that change first.

andrewsykim:
@aojea actually it might be easier to update KUBE-LOOP-BACK to target KUBE-MARK-MASQ instead of MASQUERADE directly. This should be a one-line change here https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/ipvs/proxier.go#L171

andrewsykim:
^ nvm cause the early return from KUBE-POSTROUTING is still there, but we should still do this in a follow-up PR

rikatz:
Well my bad, I thought that I've tested this also in IPVS but it seems that not :/ Sorry about that

aojea:
> Well my bad, I thought that I've tested this also in IPVS but it seems that not :/ Sorry about that

this is nobodies fault

andrewsykim:
> this is nobodies fault

+100 

I'll add that this only impacts routing of pods back to the originating node so it is really easy to miss in testing. Good thing we had e2es to catch it :)

thockin:
Good catch.  @rikatz these are the sorts of things that just happen.  There are and will always be corner cases that are not as well-exposed as the main paths.  Don't sweat it :)

aojea:
ok, let's see if instead of masquerading directly the traffic, we just detect it and mark it, avoiding the return rule

aojea:
this is how it look like now
```
-A PREROUTING -m comment --comment "kubernetes service portals" -j KUBE-SERVICES
-A OUTPUT -m comment --comment "kubernetes service portals" -j KUBE-SERVICES
-A POSTROUTING -m comment --comment "kubernetes postrouting rules" -j KUBE-POSTROUTING
-A KUBE-FIREWALL -j KUBE-MARK-DROP
-A KUBE-LOAD-BALANCER -j KUBE-MARK-MASQ
-A KUBE-MARK-MASQ -j MARK --set-xmark 0x4000/0x4000
-A KUBE-POSTROUTING -m mark ! --mark 0x4000/0x4000 -j RETURN
-A KUBE-POSTROUTING -j MARK --set-xmark 0x4000/0x0
-A KUBE-POSTROUTING -m comment --comment "kubernetes service traffic requiring SNAT" -j MASQUERADE --random-fully
-A KUBE-SERVICES -m comment --comment "Kubernetes endpoints dst ip:port, source ip for solving hairpin purpose" -m set --match-set KUBE-LOOP-BACK dst,dst,src -j KUBE-MARK-MASQ
-A KUBE-SERVICES ! -s 10.244.0.0/16 -m comment --comment "Kubernetes service cluster ip + port for masquerade purpose" -m set --match-set KUBE-CLUSTER-IP dst,dst -j KUBE-MARK-MASQ
-A KUBE-SERVICES -m addrtype --dst-type LOCAL -j KUBE-NODE-PORT
-A KUBE-SERVICES -m set --match-set KUBE-CLUSTER-IP dst,dst -j ACCEPT
```

aojea:
nope, is not working :/

aojea:
hehe, it works, I know that is not the more elegant fix but it works :smile: 
/hold cancel
/priority critical-urgent

test/e2e/storage/flexvolume.go
andrewsykim: cc @msau42 is this okay to do here?

pkg/proxy/ipvs/proxier.go
```diff
@@ -1845,6 +1844,17 @@ func (proxier *Proxier) createAndLinkeKubeChain() {
 		}
 	}
 
+	// Handle traffic that loops back to the originator with SNAT.
+	// The rules should go before the kubernetes-specific ones
+	if !proxier.ipsetList[kubeLoopBackIPSet].isEmpty() {
```
andrewsykim, line 1849: I don't think this would work because this is before we actually insert entries into the ipset https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/ipvs/proxier.go#L1164-L1169

andrewsykim:
@aojea could we instead move the mark masq rules added in https://github.com/kubernetes/kubernetes/pull/92035 to `writeIptableRules`?

pkg/proxy/ipvs/proxier.go
aojea: the test are passing so it has to be working :sweat_smile: 
but yeah, it seems it can skip one cycle, I think it works because the ipset will have always the coredns endpoints so once it's written it's always going to be there

andrewsykim: Yeah it may be "working" but I don't think this is the correct solution. On line 1081, we reset the ipset entries, but this happens after `createAndLinkeKubeChain` so here we are checking against the ipset from the previous sync.

aojea: right, just was trying to understand why it worked, definitively we can not move ahead with this,.
I'll change it

andrewsykim (commented):
/approve

Thanks @aojea :bow: 

@danwinship can you give this one another review to make sure the mark/unmark logic is correct. Logic is the same, just shifting the ordering so it's after any rules we add to KUBE-POSTROUTING in the IPVS proxy.

pkg/proxy/ipvs/proxier.go
andrewsykim: yeah I'm curious too, I think we can assume it's from the previous sync but not 100% sure  :thinking:

danwinship:
yup, lgtm

andrewsykim:
@aojea are you gonna do the cherry-picks?

Lion-Wei:
Seems like problem perfectly solved. 😄
If we want cherry-pick this pr, maybe leave the flexvolume commit alone. 
/lgtm

test/e2e/storage/flexvolume.go
msau42: The switch the distroless should not have impacted the "non-attachable" test case. This test case is passing fine in the normal k8s job: https://k8s-testgrid.appspot.com/sig-storage-kubernetes#gce&include-filter-by-regex=Flexvolume

So it would be interesting to figure out why it is consistently failing in this job.

msau42:
/hold
sorry for the late response, but adding a feature tag to the flexvolume test is not ideal as it will not run in the default k8s jobs anymore.

aojea:
> /hold
> sorry for the late response, but adding a feature tag to the flexvolume test is not ideal as it will not run in the default k8s jobs anymore.

yeah, my fault, was assuming it was a leftover, let me remove it

aojea:
/hold cancel
flexvolume commit removed

aojea:
@Lion-Wei do you mind lgtm again?

aojea:
/retest
the ipvs job failure is because the flexvolume test only

Lion-Wei:
@aojea Not at all.
/lgtm

aojea:
/retest
:thinking:

andrewsykim:
I thought the flex volume tests were failing because quay.io was down but that seemed to be fixed now https://github.com/kubernetes/kubernetes/issues/91242 ?

andrewsykim:
Maybe we just need to update to use the staging CI image for quay? @BenTheElder might know?

andrewsykim (commented):
/hold cancel

Removing the hold though since only the flex volume tests are failing and this is a known issue

msau42:
I don't think root cause of flex volume test failing in this job is known. It's passing in normal k8s ci, so I suspect it is related to the configuration of this job.

This line from the test log is potentially suspicious and indicates that the test is trying to use IPs.
```
Jul  1 15:50:26.384: INFO: Getting external IP address for e2e-bc0cae53a3-2e794-minion-group-mxnp
```

test/e2e/storage/flexvolume.go
aojea: sorry about that, my mistake

PR #92595, opened 2020-06-29
Fix the bug that kubectl logs does not filter correctly with --since-time at fractions of a second

### LINKED ISSUE #92389: kubectl does not filter correctly with --since-time at fractions of a second
**What happened**:
When using `--timestamps=true` and `--since-time`, `kubectl logs` does not filter correctly, and logs that appear fractions of a second before the specified `since-time` are still printed. It seems that it is only filtering at a whole second level.

`--since-time` should filter at the precise timestamp provided

In this case, `since-time=2020-06-22T13:11:11.749916129Z`, but the first logs printed start at `2020-06-22T13:11:11.661137356Z`
```
$ kubectl logs pod/openshift-kube-scheduler-operator-86cb559677-hb84l --timestamps=true --since-time=2020-06-22T13:11:11.749916129Z
2020-06-22T13:11:11.661137356Z I0622 13:11:11.661090       1 streamwatcher.go:114] Unexpected EOF during watch stream event decoding: unexpected EOF
2020-06-22T13:11:11.661646643Z I0622 13:11:11.661620       1 streamwatcher.go:114] Unexpected EOF during watch stream event decoding: unexpected EOF
2020-06-22T13:11:11.691129252Z I0622 13:11:11.690979       1 streamwatcher.go:114] Unexpected EOF during watch stream event decoding: unexpected EOF
2020-06-22T13:11:11.691567181Z I0622 13:11:11.691537       1 streamwatcher.go:114] Unexpected EOF during watch stream event decoding: unexpected EOF
2020-06-22T13:11:11.692054144Z I0622 13:11:11.692011       1 streamwatcher.go:114] Unexpected EOF during watch stream event decoding: unexpected EOF
2020-06-22T13:11:11.724484516Z I0622 13:11:11.692347       1 streamwatcher.go:114] Unexpected EOF during watch stream event decoding: unexpected EOF
2020-06-22T13:11:11.749916129Z I0622 13:11:11.749877       1 streamwatcher.go:114] Unexpected EOF during watch stream event decoding: unexpected EOF
2020-06-22T13:11:11.750267052Z I0622 13:11:11.749968       1 streamwatcher.go:114] Unexpected EOF during watch stream event decoding: unexpected EOF
```

In my own testing, the nanonseconds were truncated by client-go, but updating the [parameter marshalling](https://github.com/kubernetes/kubernetes/blob/9cd906e9322f75ce48b47613dd9e025c2b06c143/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/time.go#L183) to use `RFC3339Nano` preserved the nanoseconds up to that point. However I still did not see the proper results

**Environment**:
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"19+", GitVersion:"v1.19.0-alpha.1.527+5668dbdc6e60f5", GitCommit:"5668dbdc6e60f54d45f2022fddf8a92359cdcac5", GitTreeState:"clean", BuildDate:"2020-04-13T18:28:17Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.3", GitCommit:"106f511", GitTreeState:"clean", BuildDate:"2020-06-22T04:18:30Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
```
- Kernel (e.g. `uname -a`):
- Install tools:
- Others: https://bugzilla.redhat.com/show_bug.cgi?id=1778137

/sig api-machinery
/sig cli

SataQiu:
**What type of PR is this?**
/kind bug

Fix the bug that `kubectl logs` does not filter correctly with `--since-time` at fractions of a second.
After some investigation, I found that the reasons of the issue were: 
1. `kubectl` lost accuracy when it made a request to `kube-apiserver`
2. `kube-apiserver` also lost accuracy when it made a request to `kubelet`

To solve this problem, we need to fix `kubectl` as well `kube-apiserver`.

Before this patch:
```
[root@centos-01 ~]# ./kubectl logs -f etcd-centos-01 -n kube-system --timestamps=true --since-time=2020-06-29T04:49:18.759915138Z
2020-06-29T04:49:18.013711163Z 2020-06-29 04:49:18.013576 I | mvcc: store.index: compact 3883339
2020-06-29T04:49:18.759915138Z 2020-06-29 04:49:18.759815 I | mvcc: finished scheduled compaction at 3883339 (took 706.127822ms)
2020-06-29T04:52:41.583440288Z 2020-06-29 04:52:41.582861 W | etcdserver: read-only range request "key:\"/registry/namespaces/kube-system\" " with result "range_response_count:1 size:265" took too long (115.855114ms) to execute
...
```

After 
```
[root@centos-01 ~]# ./kubectl logs -f etcd-centos-01 -n kube-system --timestamps=true --since-time=2020-06-29T04:49:18.759915138Z
2020-06-29T04:49:18.759915138Z 2020-06-29 04:49:18.759815 I | mvcc: finished scheduled compaction at 3883339 (took 706.127822ms)
2020-06-29T04:52:41.583440288Z 2020-06-29 04:52:41.582861 W | etcdserver: read-only range request "key:\"/registry/namespaces/kube-system\" " with result "range_response_count:1 size:265" took too long (115.855114ms) to execute
```

Fixes #92389

```release-note
fix the bug that 'kubectl logs' does not filter correctly with '--since-time' at fractions of a second
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/time.go
+++ b/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/time.go
@@ -130,7 +130,7 @@ func (t *Time) UnmarshalQueryParameter(str string) error {
 		return nil
 	}
 
-	pt, err := time.Parse(time.RFC3339, str)
+	pt, err := time.Parse(time.RFC3339Nano, str)
 	if err != nil {
 		return err
 	}
@@ -180,7 +180,7 @@ func (t Time) MarshalQueryParameter() (string, error) {
 		return "", nil
 	}
 
-	return t.UTC().Format(time.RFC3339), nil
+	return t.UTC().Format(time.RFC3339Nano), nil
 }
 
 // Fuzz satisfies fuzz.Interface.
--- a/test/e2e/kubectl/kubectl.go
+++ b/test/e2e/kubectl/kubectl.go
@@ -1415,8 +1415,8 @@ metadata:
 		ginkgo.BeforeEach(func() {
 			ginkgo.By("creating an pod")
 			nsFlag = fmt.Sprintf("--namespace=%v", ns)
-			// Agnhost image generates logs for a total of 100 lines over 20s.
-			framework.RunKubectlOrDie(ns, "run", podName, "--image="+agnhostImage, nsFlag, "--", "logs-generator", "--log-lines-total", "100", "--run-duration", "20s")
+			// Agnhost image generates logs for a total of 100 lines over 30s.
+			framework.RunKubectlOrDie(ns, "run", podName, "--image="+agnhostImage, nsFlag, "--", "logs-generator", "--log-lines-total", "100", "--run-duration", "30s")
 		})
 		ginkgo.AfterEach(func() {
 			framework.RunKubectlOrDie(ns, "delete", "pod", podName, nsFlag)
@@ -1484,6 +1484,51 @@ metadata:
 			olderOut := framework.RunKubectlOrDie(ns, "logs", podName, containerName, nsFlag, "--since=24h")
 			older := len(strings.Split(olderOut, "\n"))
 			gomega.Expect(recent).To(gomega.BeNumerically("<", older), "expected recent(%v) to be less than older(%v)\nrecent lines:\n%v\nolder lines:\n%v\n", recent, older, recentOut, olderOut)
+
+			ginkgo.By("restricting since a specific timestamp")
+			out = framework.RunKubectlOrDie(ns, "logs", podName, containerName, nsFlag, "--timestamps")
+			framework.Logf("got output %q", out)
+			var slow, fast *time.Time
+			for _, line := range strings.Split(out, "\n") {
+				words := strings.Split(line, " ")
+				if len(words) > 0 && len(words[0]) > 0 {
+					timestamp, err := time.Parse(time.RFC3339Nano, words[0])
+					if err != nil {
+						timestamp, err = time.Parse(time.RFC3339, words[0])
+						if err != nil {
+							framework.Failf("expected %q to be RFC3339 or RFC3339Nano", words[0])
+						}
+					}
+					if slow == nil {
+						slow = &timestamp
+					} else {
+						fast = &timestamp
+						if fast.Sub(*slow) < time.Second {
+							break
+						}
+						slow = fast
+					}
+				}
+			}
+			for _, timestamp := range []*time.Time{slow, fast} {
+				if timestamp == nil {
+					break
+				}
+				sinceFlag := fmt.Sprintf("--since-time=%s", timestamp.Format(time.RFC3339Nano))
+				out = framework.RunKubectlOrDie(ns, "logs", podName, containerName, nsFlag, "--timestamps", sinceFlag)
+				lines := strings.Split(out, "\n")
+				gomega.Expect(len(lines)).To(gomega.BeNumerically(">", 0))
+				words := strings.Split(lines[0], " ")
+				gomega.Expect(len(words)).To(gomega.BeNumerically(">", 1))
+				actualTimestamp, err := time.Parse(time.RFC3339Nano, words[0])
+				if err != nil {
+					actualTimestamp, err = time.Parse(time.RFC3339, words[0])
+					if err != nil {
+						framework.Failf("expected %q to be RFC3339 or RFC3339Nano", words[0])
+					}
+				}
+				framework.ExpectEqual(actualTimestamp, *timestamp)
+			}
 		})
 	})
 
--- a/pkg/registry/core/pod/strategy.go
+++ b/pkg/registry/core/pod/strategy.go
@@ -367,7 +367,7 @@ func LogLocation(
 		params.Add("sinceSeconds", strconv.FormatInt(*opts.SinceSeconds, 10))
 	}
 	if opts.SinceTime != nil {
-		params.Add("sinceTime", opts.SinceTime.Format(time.RFC3339))
+		params.Add("sinceTime", opts.SinceTime.Format(time.RFC3339Nano))
 	}
 	if opts.TailLines != nil {
 		params.Add("tailLines", strconv.FormatInt(*opts.TailLines, 10))


damemi (commented):
@SataQiu thank you for finding the solution to this! Could you add a test to verify it?

SataQiu:
> @SataQiu thank you for finding the solution to this! Could you add a test to verify it?

Thanks @damemi
In the added e2e test case, I chose two close timestamps (the difference is less than 1 second) to verify that the log can be filtered correctly.

test/e2e/kubectl/kubectl.go
```diff
@@ -1399,7 +1399,7 @@ metadata:
 			ginkgo.By("creating an pod")
 			nsFlag = fmt.Sprintf("--namespace=%v", ns)
 			// Agnhost image generates logs for a total of 100 lines over 20s.
-			framework.RunKubectlOrDie(ns, "run", podName, "--image="+agnhostImage, nsFlag, "--", "logs-generator", "--log-lines-total", "100", "--run-duration", "20s")
+			framework.RunKubectlOrDie(ns, "run", podName, "--image="+agnhostImage, nsFlag, "--", "logs-generator", "--log-lines-total", "100", "--run-duration", "30s")
```
SataQiu, line 1402: `--run-duration` should big enough to cover all the `kubectl logs` test cases before Pod restart.

soltysh (approved):
/lgtm
/approve
/hold
@SataQiu can you please squash your changes

soltysh:
/assign @deads2k 
for apimachinery approval

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/time.go
```diff
@@ -180,7 +180,7 @@ func (t Time) MarshalQueryParameter() (string, error) {
 		return "", nil
 	}
 
-	return t.UTC().Format(time.RFC3339), nil
+	return t.UTC().Format(time.RFC3339Nano), nil
```
deads2k, line 183: Changing marshalling looks like a compatibility change that could mess with clients reading the values.

deads2k:
/hold

I'm concerned about the marshalling change.  Unmarshalling changes to be more flexible look safe to me, but marshalling will change the outgoing data.  

/assign @liggitt 

for a second opinion.

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/time.go
```diff
@@ -130,7 +130,7 @@ func (t *Time) UnmarshalQueryParameter(str string) error {
 		return nil
 	}
 
-	pt, err := time.Parse(time.RFC3339, str)
```
liggitt, line 133: Old servers will not have this change. Changing clients to serialize in a way old servers cannot deserialize breaks compatibility.

liggitt:
Agree that this is not backwards compatible.

damemi:
Is there a way to make this backward compatible? Without this it seems that the client just can't support `RFC3339Nano` accurately

liggitt:
> Is there a way to make this backward compatible?

The only thing I can think of is plumbing a second parameter all the way through the api server and kubelet, populating both from the client, and dealing with the possibility of old servers ignoring the higher precision parameter... I would tend to think that sub-second log division is not a high enough priority to justify that.

SataQiu:
Golang is very great! I have done a test about this:
```go
package main

import (
	"fmt"
	"time"
)

func main() {
	t := time.Now()

	str := t.UTC().Format(time.RFC3339Nano)
	pt, err := time.Parse(time.RFC3339Nano, str)
	fmt.Printf("got %v, err: %v\n", pt, err)

	str = t.UTC().Format(time.RFC3339)
	pt, err = time.Parse(time.RFC3339, str)
	fmt.Printf("got %v, err: %v\n", pt, err)

	str = t.UTC().Format(time.RFC3339)
	pt, err = time.Parse(time.RFC3339Nano, str)
	fmt.Printf("got %v, err: %v\n", pt, err)

	str = t.UTC().Format(time.RFC3339Nano)
	pt, err = time.Parse(time.RFC3339, str)
	fmt.Printf("got %v, err: %v\n", pt, err)
}

OUTPUT IS:
got 2020-07-01 02:54:38.167651 +0000 UTC, err: <nil>
got 2020-07-01 02:54:38 +0000 UTC, err: <nil>
got 2020-07-01 02:54:38 +0000 UTC, err: <nil>
got 2020-07-01 02:54:38.167651 +0000 UTC, err: <nil>
```
So even the old servers (API servers and kubelets) do not have this change, they can also parse the new format parameter (more accurate).  What do you think? @liggitt @deads2k @damemi

SataQiu:
@damemi That's why if we only change `MarshalQueryParameter()` to use `time.RFC3339Nano`, rebuild the `kubectl` binary (client), but do not replace `kube-apiserver` (server), we will get the logs like before. Even `kubectl` sends a more accurate timestamp, `kube-apiserver` can also parse it correctly, but it will using `t.UTC().Format(time.RFC3339)` to format the timestamp (lost accuracy) and sends to `kubelet`.

damemi:
> @damemi That's why if we only change `MarshalQueryParameter()` to use `time.RFC3339Nano`, rebuild the `kubectl` binary (client), but do not replace `kube-apiserver` (server), we will get the logs like before. Even `kubectl` sends a more accurate timestamp, `kube-apiserver` can also parse it correctly, but it will using `t.UTC().Format(time.RFC3339)` to format the timestamp (lost accuracy) and sends to `kubelet`.

@SataQiu ah, this makes sense. With that and your test in https://github.com/kubernetes/kubernetes/pull/92595#issuecomment-652160810 it seems like when a client receives an unexpected `RFC3339Nano` it is compatible and truncates it to `RFC3339`

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/time.go
SataQiu: see https://github.com/kubernetes/kubernetes/pull/92595#issuecomment-652160810

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/time.go
SataQiu: see https://github.com/kubernetes/kubernetes/pull/92595#issuecomment-652160810

soltysh:
@SataQiu you proved that parsing works which is something I'd expect. The problem here is serialization, iow. you'd need to convert to json and back mixing old and new to ensure this is backwards compatible.

SataQiu:
> @SataQiu you proved that parsing works which is something I'd expect. The problem here is serialization, iow. you'd need to convert to json and back mixing old and new to ensure this is backwards compatible.

Thanks @soltysh  I'll do more checks!

damemi:
bumping, @SataQiu just checking, were you able to verify this with a serialization test as well?

SataQiu:
Hi @damemi Emm... I tested the following scenarios:
```
new kubectl + old apiserver + old kubelet => can work, but this bug is not fixed
new kubectl + new apiserver + old kubelet => can work and this bug is fixed
new kubectl + new apiserver + new kubelet => can work and this bug is fixed
old kubectl + new apiserver + new kubelet => can work, but this bug is not fixed
```
I have no good idea to test the compatibility between different component versions.
Any examples of this?

damemi:
> Hi @damemi Emm... I tested the following scenarios:
> 
> ```
> new kubectl + old apiserver + old kubelet => can work, but this bug is not fixed
> new kubectl + new apiserver + old kubelet => can work and this bug is fixed
> new kubectl + new apiserver + new kubelet => can work and this bug is fixed
> old kubectl + new apiserver + new kubelet => can work, but this bug is not fixed
> ```
> 
> I have no good idea to test the compatibility between different component versions.
> Any examples of this?

Right, you tested these scenarios with `time.Parse` in https://github.com/kubernetes/kubernetes/pull/92595#issuecomment-652160810, but as @soltysh mentioned in https://github.com/kubernetes/kubernetes/pull/92595#issuecomment-652854213 the last thing we'll need to confirm is that when these formatted values are serialized to json, can they be properly deserialized (and vice versa). So basically what you already did, but just to check that it works as decoding JSON values (instead of being directly passed to `Parse`)

SataQiu:
```go
package main

import (
	"encoding/json"
	"fmt"
	"time"
)

func main() {
	t := time.Now()

	{ // 000
		var str string
		buf := make([]byte, 0, len(time.RFC3339Nano)+2)
		buf = append(buf, '"')
		buf = t.UTC().AppendFormat(buf, time.RFC3339Nano)
		buf = append(buf, '"')
		err1 := json.Unmarshal(buf, &str)
		pt, err2 := time.Parse(time.RFC3339Nano, str)
		fmt.Printf("000: got %v, err1: %v, err2: %v\n", pt, err1, err2)
	}

	{ // 001
		var str string
		buf := make([]byte, 0, len(time.RFC3339Nano)+2)
		buf = append(buf, '"')
		buf = t.UTC().AppendFormat(buf, time.RFC3339Nano)
		buf = append(buf, '"')
		err1 := json.Unmarshal(buf, &str)
		pt, err2 := time.Parse(time.RFC3339, str)
		fmt.Printf("001: got %v, err1: %v, err2: %v\n", pt, err1, err2)
	}
	{ // 010
		var str string
		buf := make([]byte, 0, len(time.RFC3339Nano)+2)
		buf = append(buf, '"')
		buf = t.UTC().AppendFormat(buf, time.RFC3339)
		buf = append(buf, '"')
		err1 := json.Unmarshal(buf, &str)
		pt, err2 := time.Parse(time.RFC3339Nano, str)
		fmt.Printf("010: got %v, err1: %v, err2: %v\n", pt, err1, err2)
	}
	{ // 011
		var str string
		buf := make([]byte, 0, len(time.RFC3339Nano)+2)
		buf = append(buf, '"')
		buf = t.UTC().AppendFormat(buf, time.RFC3339)
		buf = append(buf, '"')
		err1 := json.Unmarshal(buf, &str)
		pt, err2 := time.Parse(time.RFC3339, str)
		fmt.Printf("011: got %v, err1: %v, err2: %v\n", pt, err1, err2)
	}
	{ // 100
		var str string
		buf := make([]byte, 0, len(time.RFC3339)+2)
		buf = append(buf, '"')
		buf = t.UTC().AppendFormat(buf, time.RFC3339Nano)
		buf = append(buf, '"')
		err1 := json.Unmarshal(buf, &str)
		pt, err2 := time.Parse(time.RFC3339Nano, str)
		fmt.Printf("100: got %v, err1: %v, err2: %v\n", pt, err1, err2)
	}
	{ // 101
		var str string
		buf := make([]byte, 0, len(time.RFC3339)+2)
		buf = append(buf, '"')
		buf = t.UTC().AppendFormat(buf, time.RFC3339Nano)
		buf = append(buf, '"')
		err1 := json.Unmarshal(buf, &str)
		pt, err2 := time.Parse(time.RFC3339, str)
		fmt.Printf("101: got %v, err1: %v, err2: %v\n", pt, err1, err2)
	}
	{ // 110
		var str string
		buf := make([]byte, 0, len(time.RFC3339)+2)
		buf = append(buf, '"')
		buf = t.UTC().AppendFormat(buf, time.RFC3339)
		buf = append(buf, '"')
		err1 := json.Unmarshal(buf, &str)
		pt, err2 := time.Parse(time.RFC3339Nano, str)
		fmt.Printf("110: got %v, err1: %v, err2: %v\n", pt, err1, err2)
	}
	{ // 111
		var str string
		buf := make([]byte, 0, len(time.RFC3339)+2)
		buf = append(buf, '"')
		buf = t.UTC().AppendFormat(buf, time.RFC3339)
		buf = append(buf, '"')
		err1 := json.Unmarshal(buf, &str)
		pt, err2 := time.Parse(time.RFC3339, str)
		fmt.Printf("111: got %v, err1: %v, err2: %v\n", pt, err1, err2)
	}
}

OUTPUT:
000: got 2020-07-15 02:35:07.082923 +0000 UTC, err1: <nil>, err2: <nil>
001: got 2020-07-15 02:35:07.082923 +0000 UTC, err1: <nil>, err2: <nil>
010: got 2020-07-15 02:35:07 +0000 UTC, err1: <nil>, err2: <nil>
011: got 2020-07-15 02:35:07 +0000 UTC, err1: <nil>, err2: <nil>
100: got 2020-07-15 02:35:07.082923 +0000 UTC, err1: <nil>, err2: <nil>
101: got 2020-07-15 02:35:07.082923 +0000 UTC, err1: <nil>, err2: <nil>
110: got 2020-07-15 02:35:07 +0000 UTC, err1: <nil>, err2: <nil>
111: got 2020-07-15 02:35:07 +0000 UTC, err1: <nil>, err2: <nil>
```

@damemi Did you mean these tests?

damemi:
@SataQiu yeah, https://github.com/kubernetes/kubernetes/pull/92595#issuecomment-658511541 looks like it shows these formats are compatible when serialized. @soltysh @liggitt do you agree?

liggitt:
Given metav1.Time is a very widely used field, and that serialized data is consumed by non-go clients, this does not seem like a safe change to me. Is there a compelling reason sub-second log times are needed?

PR #92597, opened 2020-06-29
Unify the scoring process for `PodTopologySpread` plugin

chendave:
Current IMPL assumes all the node has hostname label defined and with different value
but highlight `hostname` out make the flow of scoring unnecessary complicated and hard
to understand.

**What type of PR is this?**
>
/kind cleanup

/sig scheduling

Fixes #

```release-note
NONE
```

```docs

```


ahg-g:
/hold

Treating hostname differently is intentional, it is done as an optimization.

Huang-Wei:
Yup, it's intentional.

> Current IMPL assumes all the node has hostname label defined and with different value
but highlight hostname out make the flow of scoring unnecessary complicated and hard
to understand.

`kubernetes.io/hostname` is a standard label most cloud and on-premise providers would adopt, so we can assume that. If this label is not present, current implementation won't apply extra overhead either.

PR #92603, opened 2020-06-29
Do not raise exception if unscheduled Pod status is unknown

hasheddan:
**What type of PR is this?**

/kind bug

Currently when checking for unscheduled pods an exception will be raised
if a pod is not scheduled and the status is unknown. This update modifies
the logic to skip pods with an unknown status when determining which
pods have been scheduled and which have not.

Follow up to #92509 and #92450 

See https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-gce-conformance-latest/1277468902246322177 for details on failure.

It might be worth following this up with separating the logic out into separate functions as the use-case in

https://github.com/kubernetes/kubernetes/blob/1cbda2493d2744af0d0e52b008976af0fc1d3a5e/test/e2e/scheduling/predicates.go#L933

is slightly different than in

https://github.com/kubernetes/kubernetes/blob/1cbda2493d2744af0d0e52b008976af0fc1d3a5e/test/e2e/scheduling/framework.go#L44

/cc @oomichi @Huang-Wei @BenTheElder 

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/scheduling/predicates.go
+++ b/test/e2e/scheduling/predicates.go
@@ -1052,15 +1052,8 @@ func GetPodsScheduled(workerNodes sets.String, pods *v1.PodList) (scheduledPods,
 				framework.ExpectEqual(scheduledCondition.Status, v1.ConditionTrue)
 				scheduledPods = append(scheduledPods, pod)
 			}
-		} else {
-			_, scheduledCondition := podutil.GetPodCondition(&pod.Status, v1.PodScheduled)
-			framework.ExpectEqual(scheduledCondition != nil, true)
-			if scheduledCondition != nil {
-				framework.ExpectEqual(scheduledCondition.Status, v1.ConditionFalse)
-				if scheduledCondition.Reason == "Unschedulable" {
-					notScheduledPods = append(notScheduledPods, pod)
-				}
-			}
+		} else if pod.Spec.NodeName == "" {
+			notScheduledPods = append(notScheduledPods, pod)
 		}
 	}
 	return


hasheddan:
@ahg-g if you could take a look here it would also be appreciated :)

test/e2e/scheduling/predicates.go
```diff
@@ -1056,7 +1056,6 @@ func GetPodsScheduled(workerNodes sets.String, pods *v1.PodList) (scheduledPods,
 			_, scheduledCondition := podutil.GetPodCondition(&pod.Status, v1.PodScheduled)
 			framework.ExpectEqual(scheduledCondition != nil, true)
 			if scheduledCondition != nil {
-				framework.ExpectEqual(scheduledCondition.Status, v1.ConditionFalse)
```
ahg-g, line 1059: @Huang-Wei any idea why we had this check in the first place? if a pod was never tried, scheduled condition would never be set, which should be a valid notScheduledPod case.

hasheddan: @ahg-g I also was wondering about the following line, it will not add the pod to the list of unscheduled unless its reason is `Unschedulable`. It seems that a pod should also be added to that list if it is schedulable but just no scheduled yet for some reason. I may be missing context though

Huang-Wei: @ahg-g I've no idea either. It looks like that check is incorrect, but seems we don't have points reporting errors on that line.

@hasheddan Are you sure the flake turns out that the condition is Unknown instead of Scheduled? I asked this is b/c I feel like a scheduled Pod on master node can fall into current `else` block. If so, we should rule master pods out by changing `else` to `else if pod.Spec.NodeName == ""`.

hasheddan: @Huang-Wei I can update, but I am not sure if it makes a difference because the next line checks `if scheduledCondition.Reason == "Unschedulable"`. so `Scheduled` pods would be ignored anyway. Is that how you see it as well?

Huang-Wei: If we made it `else if pod.Spec.NodeName == ""`, then it cannot be a `Scheduled` pod. But it's still possible (1) the status is Unknown, (2) the status is nil. So, a bit aggressive, how about changing the whole function to:

```go
// GetPodsScheduled returns a number of currently scheduled and not scheduled Pods on worker nodes.
func GetPodsScheduled(workerNodes sets.String, pods *v1.PodList) (scheduledPods, notScheduledPods []v1.Pod) {
		if pod.Spec.NodeName != "" && workerNodes.Has(pod.Spec.NodeName) {
			_, scheduledCondition := podutil.GetPodCondition(&pod.Status, v1.PodScheduled)
			framework.ExpectEqual(scheduledCondition != nil, true)
			if scheduledCondition != nil {
				framework.ExpectEqual(scheduledCondition.Status, v1.ConditionTrue)
				scheduledPods = append(scheduledPods, pod)
			}
		} else if pod.Spec.NodeName == "" {
			notScheduledPods = append(notScheduledPods, pod)
		}
	}
	return
}
```

hasheddan: @Huang-Wei this LGTM, updating now :)

hasheddan:
@Huang-Wei can you take another look here?

test/e2e/scheduling/predicates.go
Huang-Wei: Thanks. @ahg-g wanna take another look?

Huang-Wei:
/lgtm

Will leave /approve to @ahg-g .

PR #91750, opened 2020-06-04
Fix an issue that a Pod's nominatedNodeName cannot be cleared upon node deletion

Huang-Wei:
**What type of PR is this?**

/kind bug
/sig scheduling

**What this PR does / why we need it**:

This PR tries to fix an issue that a Pod's nominatedNodeName cannot be cleared upon node deletion. Detailed reproducing steps please check #85677 and the integration test of this PR.

**Which issue(s) this PR fixes**:

Fixes #85677

**Special notes for your reviewer**:

A side effect is that upon unexpected internal errors during preemption, the nominatedNode would be returned as "", so the pod's non-nil nominatedNodeName will be cleared. It means its reserved room on a node can be occupied by other pods before it's scheduled next time.
But compared to the issue that its nominatedNode is unable to be cleared, the impact is negligible.

**Does this PR introduce a user-facing change?**:

```release-note
Fixed an issue that a Pod's nominatedNodeName cannot be cleared upon node deletion.
```

PR #92604, opened 2020-06-29
The Pod is eligible to preempt when previous nominanted node is UnschedulableAndUnresolvable 

### LINKED ISSUE #92600: Pod won't be considered for preemption even its nominated node is considered as UnschedulableAndUnresolvable by filters
**What happened**:
When a Pod's nominated node is considered as UnschedulableAndUnresolvable in the filter stage, the scheduler doesn't consider that pod for preemption even there are pods can be preempted on another node.

When the pod's nominated node is considered as 'UnschedulableAndUnresolvable ' (For example, the taint being added by those conditions, https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-nodes-by-condition), the scheduler should try to preemption on other nodes.

- Create low priority pod, which schedule to nodeA
- Create mid priority pod, which schedule to nodeB
- Create high priority pod, which preempt the low priority pod on nodeA, and set high priority pod's nominated node as nodeA
- The nodeA's kubelet is down, the node life cycle controller taint the nodeA with 'node.kubernetes.io/unschedulable' (or taint with disk pressure..etc) before low priority pod finishes the termination.
- The high priority pod is keeping as pending status. The scheduler doesn't try to preempt the mid-priority pod on nodeB

**Environment**:
1.18.0
baremetal
ubuntu1804
- Kernel (e.g. `uname -a`):
5.3.0-46-generic
- Install tools:
- Others:

soulxu:
**What type of PR is this?**

/kind bug
/sig scheduling

This PR is trying to fix the Pod won't be considered for preemption even its nominated node is considered as UnschedulableAndUnresolvable by the filters

Fixes #92600

```release-note
Fixes the pod can be considered for preemption after its previous nominated node become unschedulable and unresolvable.
```

### DIFF
--- a/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
+++ b/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
@@ -106,7 +106,7 @@ func preempt(ctx context.Context, fh framework.FrameworkHandle, state *framework
 		return "", err
 	}
 
-	if !podEligibleToPreemptOthers(pod, fh.SnapshotSharedLister().NodeInfos()) {
+	if !podEligibleToPreemptOthers(pod, fh.SnapshotSharedLister().NodeInfos(), m[pod.Status.NominatedNodeName]) {
 		klog.V(5).Infof("Pod %v/%v is not eligible for more preemption.", pod.Namespace, pod.Name)
 		return "", nil
 	}
@@ -189,13 +189,19 @@ func preempt(ctx context.Context, fh framework.FrameworkHandle, state *framework
 // considered for preemption.
 // We look at the node that is nominated for this pod and as long as there are
 // terminating pods on the node, we don't consider this for preempting more pods.
-func podEligibleToPreemptOthers(pod *v1.Pod, nodeInfos framework.NodeInfoLister) bool {
+func podEligibleToPreemptOthers(pod *v1.Pod, nodeInfos framework.NodeInfoLister, nominatedNodeStatus *framework.Status) bool {
 	if pod.Spec.PreemptionPolicy != nil && *pod.Spec.PreemptionPolicy == v1.PreemptNever {
 		klog.V(5).Infof("Pod %v/%v is not eligible for preemption because it has a preemptionPolicy of %v", pod.Namespace, pod.Name, v1.PreemptNever)
 		return false
 	}
 	nomNodeName := pod.Status.NominatedNodeName
 	if len(nomNodeName) > 0 {
+		// If the pod's nominated node is considered as UnschedulableAndUnresolvable by the filters,
+		// then the pod should be considered for preempting again.
+		if nominatedNodeStatus.Code() == framework.UnschedulableAndUnresolvable {
+			return true
+		}
+
 		if nodeInfo, _ := nodeInfos.Get(nomNodeName); nodeInfo != nil {
 			podPriority := podutil.GetPodPriority(pod)
 			for _, p := range nodeInfo.Pods {
--- a/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
+++ b/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
@@ -889,6 +889,61 @@ func TestPickOneNodeForPreemption(t *testing.T) {
 	}
 }
 
+func TestPodEligibleToPreemptOthers(t *testing.T) {
+	tests := []struct {
+		name                string
+		pod                 *v1.Pod
+		pods                []*v1.Pod
+		nodes               []string
+		nominatedNodeStatus *framework.Status
+		expected            bool
+	}{
+		{
+			name:                "Pod with nominated node",
+			pod:                 st.MakePod().Name("p_with_nominated_node").UID("p").Priority(highPriority).NominatedNodeName("node1").Obj(),
+			pods:                []*v1.Pod{st.MakePod().Name("p1").UID("p1").Priority(lowPriority).Node("node1").Terminating().Obj()},
+			nodes:               []string{"node1"},
+			nominatedNodeStatus: framework.NewStatus(framework.UnschedulableAndUnresolvable, tainttoleration.ErrReasonNotMatch),
+			expected:            true,
+		},
+		{
+			name:                "Pod with nominated node, but without nominated node status",
+			pod:                 st.MakePod().Name("p_without_status").UID("p").Priority(highPriority).NominatedNodeName("node1").Obj(),
+			pods:                []*v1.Pod{st.MakePod().Name("p1").UID("p1").Priority(lowPriority).Node("node1").Terminating().Obj()},
+			nodes:               []string{"node1"},
+			nominatedNodeStatus: nil,
+			expected:            false,
+		},
+		{
+			name:                "Pod without nominated node",
+			pod:                 st.MakePod().Name("p_without_nominated_node").UID("p").Priority(highPriority).Obj(),
+			pods:                []*v1.Pod{},
+			nodes:               []string{},
+			nominatedNodeStatus: nil,
+			expected:            true,
+		},
+		{
+			name:                "Pod with 'PreemptNever' preemption policy",
+			pod:                 st.MakePod().Name("p_with_preempt_never_policy").UID("p").Priority(highPriority).PreemptionPolicy(v1.PreemptNever).Obj(),
+			pods:                []*v1.Pod{},
+			nodes:               []string{},
+			nominatedNodeStatus: nil,
+			expected:            false,
+		},
+	}
+
+	for _, test := range tests {
+		var nodes []*v1.Node
+		for _, n := range test.nodes {
+			nodes = append(nodes, st.MakeNode().Name(n).Obj())
+		}
+		snapshot := internalcache.NewSnapshot(test.pods, nodes)
+		if got := podEligibleToPreemptOthers(test.pod, snapshot.NodeInfos(), test.nominatedNodeStatus); got != test.expected {
+			t.Errorf("expected %t, got %t for pod: %s", test.expected, got, test.pod.Name)
+		}
+	}
+}
+
 func TestNodesWherePreemptionMightHelp(t *testing.T) {
 	// Prepare 4 nodes names.
 	nodeNames := []string{"node1", "node2", "node3", "node4"}
--- a/pkg/scheduler/testing/wrappers.go
+++ b/pkg/scheduler/testing/wrappers.go
@@ -244,6 +244,12 @@ func (p *PodWrapper) StartTime(t metav1.Time) *PodWrapper {
 	return p
 }
 
+// NominatedNodeName sets `n` as the .Status.NominatedNodeName of the inner pod.
+func (p *PodWrapper) NominatedNodeName(n string) *PodWrapper {
+	p.Status.NominatedNodeName = n
+	return p
+}
+
 // PodAffinityKind represents different kinds of PodAffinity.
 type PodAffinityKind int
 


pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -191,13 +191,18 @@ func preempt(ctx context.Context, fh framework.FrameworkHandle, state *framework
...
 		klog.V(5).Infof("Pod %v/%v is not eligible for preemption because it has a preemptionPolicy of %v", pod.Namespace, pod.Name, v1.PreemptNever)
 		return false
 	}
 	nomNodeName := pod.Status.NominatedNodeName
 	if len(nomNodeName) > 0 {
+		if status, ok := m[nomNodeName]; ok {
+			if status.Code() == framework.UnschedulableAndUnresolvable {
+				return true
+			}
+		}
```
Huang-Wei, line 205: I don't think this logic can help resolve the problem thoroughly. There are more cases that this fix doesn't cover:

- The nominatedNode can no longer exist, due to network partition or node shutdown.
- High-priority pod gets scheduled to that node to take the "reserved" room, etc. And also the status.Code()
- Terminating low-priority pods are still occupying the "to-be-releasing" resources.

With that said, ok can be either true or false, also the code can be Unschedulable or UnschedulableAndUnresolvable. So fixing inside `podEligibleToPreemptOthers()` is not a good idea.

A promising solution is to continue the work in #91750 - which was too aggressive, but should be easy to rework a bit to thoroughly considering all meaningful return paths of preemption.

Huang-Wei: > Just mention that the issue shouldn't only about the case of the node being down, it covers all the case of UnschedulableAndUnresolvable. Like the operator taint the node in same time, the pod won't be able to this node anymore, so the pod should be trying to preempt other node.

Ah, I understood your motivation now. Got misled by the issue that you highlighted as "the node gets down"... The solution here is totally irrelevant with the node is down or not, right?

I _guess_ (please confirm) you were trying to say: if the nominated node is somehow infeasible, in particular UnschedulableAndUnresolvable (it can be a new high-priority pod makes the preemptor violate some critical constraints , or the node gets tainted, etc.), we shouldn't return false in `podEligibleToPreemptOthers()` by simply checking whether there are pods being terminated, as that would waste one preemption cycle.

soulxu: > > Just mention that the issue shouldn't only about the case of the node being down, it covers all the case of UnschedulableAndUnresolvable. Like the operator taint the node in same time, the pod won't be able to this node anymore, so the pod should be trying to preempt other node.
> 
> Ah, I understood your motivation now. Got misled by the issue that you highlighted as "the node gets down"... The solution here is totally irrelevant with the node is down or not, right?

Yes, it isn't only about the node down, it should be about the nominated node considered as 'UnschedulableAndUnresolvable'. Sorry for miss leading that.

> 
> I _guess_ (please confirm) you were trying to say: if the nominated node is somehow infeasible, in particular UnschedulableAndUnresolvable (it can be a new high-priority pod makes the preemptor violate some critical constraints , or the node gets tainted, etc.), we shouldn't return false in `podEligibleToPreemptOthers()` by simply checking whether there are pods being terminated, as that would waste one preemption cycle.

Yes, exactly

Huang-Wei: Got it. Then I think this PR is valid, could you reword the original issue and this PR's description?

soulxu: > Got it. Then I think this PR is valid, could you reword the original issue and this PR's description?

Thanks, done

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -108,7 +108,7 @@ func preempt(ctx context.Context, fh framework.FrameworkHandle, state *framework
 		return "", err
 	}
 
-	if !podEligibleToPreemptOthers(pod, fh.SnapshotSharedLister().NodeInfos()) {
+	if !podEligibleToPreemptOthers(pod, fh.SnapshotSharedLister().NodeInfos(), m) {
```
Huang-Wei, line 111: You can just pass a *status: `m[pod.Status.NominatedNodeName]`, which will be evaluated to nil if either pod.Status.NominatedNodeName == "" or m[nnn] does not exist.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -927,6 +927,40 @@ func TestPickOneNodeForPreemption(t *testing.T) {
 	}
 }
 
+func TestPodEligibleToPreemptOthers(t *testing.T) {
```
Huang-Wei, line 930: Can you refer to the format of #92571 to rewrite this test a bit?  In particular, use {Node/Pod}Wrapper to build the API objects.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: And please add some comments to explain the case.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -927,6 +927,40 @@ func TestPickOneNodeForPreemption(t *testing.T) {
...
+			name:  "Pod with nominated node",
+			pod:   &v1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "machine1", UID: types.UID("machine1")}, Spec: v1.PodSpec{Containers: smallContainers}, Status: v1.PodStatus{NominatedNodeName: "machine1"}},
+			pods:  []*v1.Pod{},
+			nodes: []string{"machine1"},
+			nodesStatuses: framework.NodeToStatusMap{
+				"machine1": framework.NewStatus(framework.UnschedulableAndUnresolvable, tainttoleration.ErrReasonNotMatch),
+			},
+			expected: true,
+		},
+	}
```
Huang-Wei, line 949: I'd like to expect another test to cover`expected = false`.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -927,6 +927,40 @@ func TestPickOneNodeForPreemption(t *testing.T) {
...
+		pod           *v1.Pod
+		pods          []*v1.Pod
+		nodes         []string
+		nodesStatuses framework.NodeToStatusMap
+		expected      bool
+	}{
+		{
+			name:  "Pod with nominated node",
+			pod:   &v1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "machine1", UID: types.UID("machine1")}, Spec: v1.PodSpec{Containers: smallContainers}, Status: v1.PodStatus{NominatedNodeName: "machine1"}},
+			pods:  []*v1.Pod{},
```
Huang-Wei, line 942: Remove this field if it's not needed.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -927,6 +927,40 @@ func TestPickOneNodeForPreemption(t *testing.T) {
...
+			nodes: []string{"machine1"},
```
Huang-Wei, line 943: nit: node1 is shorter :)

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -927,6 +927,40 @@ func TestPickOneNodeForPreemption(t *testing.T) {
...
+			expected: true,
+		},
+	}
+
+	for _, test := range tests {
+		var nodes []*v1.Node
+		for _, n := range test.nodes {
+			nodes = append(nodes, makeNode(n, schedutil.DefaultMilliCPURequest*5, schedutil.DefaultMemoryRequest*5))
+		}
+		snapshot := internalcache.NewSnapshot(test.pods, nodes)
```
Huang-Wei, line 956: You don't need to build a snapshot **_just for_** fetching nodeInfos. Just build the nodeInfos directly, something like:

```go
var nodeInfos []*framework.NodeInfo
for _, name := range nodeNames {
	ni := framework.NewNodeInfo()
	ni.SetNode(st.MakeNode().Name(name).Obj())
	nodeInfos = append(nodeInfos, ni)
}
```

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -927,6 +927,40 @@ func TestPickOneNodeForPreemption(t *testing.T) {
...
+		ret := podEligibleToPreemptOthers(test.pod, snapshot.NodeInfos(), test.nodesStatuses)
+		if ret != test.expected {
```
Huang-Wei, line 958: nit:

```go
if got := ...; got != test.expected
```

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
soulxu: got it, will do

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
soulxu: got it, will do after #92571 merge.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
soulxu: Done

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
soulxu: Yes, you are right :)

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
soulxu: Since there are missing unittests for podEligibleToPreemptOther method, I orginially plan to add more tests in another PR, but I think I can add them in this PR with separated commit. Let me know if it is good way your expected.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
soulxu: got it, thanks!

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
soulxu: Done

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
soulxu: sorry, I continue use the snapshot, since the podEligibleToPreemptOthers function is looking for 'framework.NodeInfoLister' which snapshot implement it

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -189,13 +189,19 @@ func preempt(ctx context.Context, fh framework.FrameworkHandle, state *framework
...
+func podEligibleToPreemptOthers(pod *v1.Pod, nodeInfos framework.NodeInfoLister, nominatedNodeStatus *framework.Status) bool {
 	if pod.Spec.PreemptionPolicy != nil && *pod.Spec.PreemptionPolicy == v1.PreemptNever {
 		klog.V(5).Infof("Pod %v/%v is not eligible for preemption because it has a preemptionPolicy of %v", pod.Namespace, pod.Name, v1.PreemptNever)
 		return false
 	}
 	nomNodeName := pod.Status.NominatedNodeName
 	if len(nomNodeName) > 0 {
+		// If the pod's nominated node is considered as UnschedulableAndUnresolvable by the filters,
+		// then the pod should be considered for preempting again.
+		if nominatedNodeStatus != nil && nominatedNodeStatus.Code() == framework.UnschedulableAndUnresolvable {
```
Huang-Wei, line 201:
```suggestion
		if nominatedNodeStatus.Code() == framework.UnschedulableAndUnresolvable {
```

As status.Code() ensure we get a success if the status is nil.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
Huang-Wei: #92571 just got merged. Please rebase to check out the latest format.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
Huang-Wei: SG.

Huang-Wei (commented):
We're almost there. Just one nit.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
soulxu: thanks, let me update

Huang-Wei:
/lgtm
/approve

Thanks @soulxu !

soulxu:
@Huang-Wei appreciate your help!

PR #92605, opened 2020-06-29
fix the wrong function description

colin404:
**What type of PR is this?**

> /kind bug

function description is wrong, need to fix it.

None
None

```release-note
NONE
```

```docs

```


cheftako:
This is one of the failures causing golint to fail for this directory (https://github.com/kubernetes/kubernetes/blob/release-1.18/hack/.golint_failures#L370). Ideally it would be good to remove this directory from the exceptions list and fix all the lint failures for this directory.

PR #92607, opened 2020-06-29
Add CRUD tests for new Events API verbs

chelseychen:
**What type of PR is this?**

/kind api-change

This PR introduces some CRUD e2e tests for new Events API verbs.

Ref https://github.com/kubernetes/enhancements/issues/383

```release-note
NONE
```

```docs
- [KEP]: <https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/383-new-event-api-ga-graduation>
```

### DIFF
--- a/test/e2e/instrumentation/events.go
+++ b/test/e2e/instrumentation/events.go
@@ -1,0 +1,176 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package instrumentation
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"k8s.io/api/core/v1"
+	eventsv1beta1 "k8s.io/api/events/v1beta1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/diff"
+	typedeventsv1beta1 "k8s.io/client-go/kubernetes/typed/events/v1beta1"
+	"k8s.io/kubernetes/test/e2e/framework"
+	"k8s.io/kubernetes/test/e2e/instrumentation/common"
+
+	"github.com/onsi/ginkgo"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func newTestEvent(namespace, name, label string) *eventsv1beta1.Event {
+	someTime := metav1.MicroTime{Time: time.Unix(1505828956, 0)}
+	return &eventsv1beta1.Event{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+			Labels: map[string]string{
+				label: "true",
+			},
+		},
+		Regarding: v1.ObjectReference{
+			Namespace: namespace,
+		},
+		EventTime:           someTime,
+		Note:                "This is " + name,
+		Action:              "Do",
+		Reason:              "Test",
+		Type:                "Normal",
+		ReportingController: "test-controller",
+		ReportingInstance:   "test-node",
+	}
+}
+
+func eventExistsInList(client typedeventsv1beta1.EventInterface, namespace, name string) bool {
+	eventsList, err := client.List(context.TODO(), metav1.ListOptions{
+		LabelSelector: "testevent-constant=true",
+	})
+	framework.ExpectNoError(err, "failed to list events")
+
+	for _, val := range eventsList.Items {
+		if val.ObjectMeta.Name == name && val.ObjectMeta.Namespace == namespace {
+			return true
+		}
+	}
+	return false
+}
+
+var _ = common.SIGDescribe("Events API", func() {
+	f := framework.NewDefaultFramework("events")
+	var client typedeventsv1beta1.EventInterface
+	var clientAllNamespaces typedeventsv1beta1.EventInterface
+
+	ginkgo.BeforeEach(func() {
+		client = f.ClientSet.EventsV1beta1().Events(f.Namespace.Name)
+		clientAllNamespaces = f.ClientSet.EventsV1beta1().Events(metav1.NamespaceAll)
+	})
+
+	/*
+		Release : v1.19
+		Testname: Event resource lifecycle
+		Description: Create an event, the event MUST exist.
+		The event is patched with a new note, the check MUST have the update note.
+		The event is deleted and MUST NOT show up when listing all events.
+	*/
+	ginkgo.It("should ensure that an event can be fetched, patched, deleted, and listed", func() {
+		eventName := "event-test"
+
+		ginkgo.By("creating a test event")
+		_, err := client.Create(context.TODO(), newTestEvent(f.Namespace.Name, eventName, "testevent-constant"), metav1.CreateOptions{})
+		framework.ExpectNoError(err, "failed to create test event")
+
+		ginkgo.By("listing events in all namespaces")
+		foundCreatedEvent := eventExistsInList(clientAllNamespaces, f.Namespace.Name, eventName)
+		framework.ExpectEqual(foundCreatedEvent, true, "failed to find test event in list with cluster scope")
+
+		ginkgo.By("listing events in test namespace")
+		foundCreatedEvent = eventExistsInList(client, f.Namespace.Name, eventName)
+		framework.ExpectEqual(foundCreatedEvent, true, "failed to find test event in list with namespace scope")
+
+		ginkgo.By("getting the test event")
+		testEvent, err := client.Get(context.TODO(), eventName, metav1.GetOptions{})
+		framework.ExpectNoError(err, "failed to get test event")
+
+		ginkgo.By("patching the test event")
+		eventPatchNote := "This is a test event - patched"
+		eventPatch, err := json.Marshal(map[string]interface{}{
+			"note": eventPatchNote,
+		})
+		framework.ExpectNoError(err, "failed to marshal the patch JSON payload")
+
+		_, err = client.Patch(context.TODO(), eventName, types.StrategicMergePatchType, []byte(eventPatch), metav1.PatchOptions{})
+		framework.ExpectNoError(err, "failed to patch the test event")
+
+		ginkgo.By("getting the test event")
+		event, err := client.Get(context.TODO(), eventName, metav1.GetOptions{})
+		framework.ExpectNoError(err, "failed to get test event")
+		// clear ResourceVersion and ManagedFields which are set by control-plane
+		event.ObjectMeta.ResourceVersion = ""
+		testEvent.ObjectMeta.ResourceVersion = ""
+		event.ObjectMeta.ManagedFields = nil
+		testEvent.ObjectMeta.ManagedFields = nil
+		testEvent.Note = eventPatchNote
+		if !apiequality.Semantic.DeepEqual(testEvent, event) {
+			framework.Failf("test event wasn't properly patched: %v", diff.ObjectReflectDiff(testEvent, event))
+		}
+
+		ginkgo.By("deleting the test event")
+		err = client.Delete(context.TODO(), eventName, metav1.DeleteOptions{})
+		framework.ExpectNoError(err, "failed to delete the test event")
+
+		ginkgo.By("listing events in all namespaces")
+		foundCreatedEvent = eventExistsInList(clientAllNamespaces, f.Namespace.Name, eventName)
+		framework.ExpectEqual(foundCreatedEvent, false, "should not have found test event after deletion")
+
+		ginkgo.By("listing events in test namespace")
+		foundCreatedEvent = eventExistsInList(client, f.Namespace.Name, eventName)
+		framework.ExpectEqual(foundCreatedEvent, false, "should not have found test event after deletion")
+	})
+
+	ginkgo.It("should delete a collection of events", func() {
+		eventNames := []string{"test-event-1", "test-event-2", "test-event-3"}
+
+		ginkgo.By("Create set of events")
+		for _, eventName := range eventNames {
+			_, err := client.Create(context.TODO(), newTestEvent(f.Namespace.Name, eventName, "testevent-set"), metav1.CreateOptions{})
+			framework.ExpectNoError(err, "failed to create event")
+		}
+
+		ginkgo.By("get a list of Events with a label in the current namespace")
+		eventList, err := client.List(context.TODO(), metav1.ListOptions{
+			LabelSelector: "testevent-set=true",
+		})
+		framework.ExpectNoError(err, "failed to get a list of events")
+		framework.ExpectEqual(len(eventList.Items), len(eventNames), fmt.Sprintf("unexpected event list: %#v", eventList))
+
+		ginkgo.By("delete a list of events")
+		framework.Logf("requesting DeleteCollection of events")
+		err = client.DeleteCollection(context.TODO(), metav1.DeleteOptions{}, metav1.ListOptions{
+			LabelSelector: "testevent-set=true",
+		})
+		framework.ExpectNoError(err, "failed to delete the test event")
+
+		ginkgo.By("check that the list of events matches the requested quantity")
+		eventList, err = client.List(context.TODO(), metav1.ListOptions{
+			LabelSelector: "testevent-set=true",
+		})
+		framework.ExpectNoError(err, "failed to get a list of events")
+		framework.ExpectEqual(len(eventList.Items), 0, fmt.Sprintf("unexpected event list: %#v", eventList))
+	})
+})
--- a/test/e2e/instrumentation/BUILD
+++ b/test/e2e/instrumentation/BUILD
@@ -7,11 +7,24 @@ load(
 
 go_library(
     name = "go_default_library",
-    srcs = ["imports.go"],
+    srcs = [
+        "events.go",
+        "imports.go",
+    ],
     importpath = "k8s.io/kubernetes/test/e2e/instrumentation",
     deps = [
+        "//staging/src/k8s.io/api/core/v1:go_default_library",
+        "//staging/src/k8s.io/api/events/v1beta1:go_default_library",
+        "//staging/src/k8s.io/apimachinery/pkg/api/equality:go_default_library",
+        "//staging/src/k8s.io/apimachinery/pkg/apis/meta/v1:go_default_library",
+        "//staging/src/k8s.io/apimachinery/pkg/types:go_default_library",
+        "//staging/src/k8s.io/apimachinery/pkg/util/diff:go_default_library",
+        "//staging/src/k8s.io/client-go/kubernetes/typed/events/v1beta1:go_default_library",
+        "//test/e2e/framework:go_default_library",
+        "//test/e2e/instrumentation/common:go_default_library",
         "//test/e2e/instrumentation/logging:go_default_library",
         "//test/e2e/instrumentation/monitoring:go_default_library",
+        "//vendor/github.com/onsi/ginkgo:go_default_library",
     ],
 )
 


chelseychen:
/assign @wojtek-t 
I mostly copied old event tests from test/e2e/apimachinery/events.go. PTAL and let me know if anything needs to be changes. Thanks!

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,191 @@
...
+	f := framework.NewDefaultFramework("events")
+
+	/*
+			   Release : v1.19
+			   Testname: Event resource lifecycle
+			   Description: Create an event, the event MUST exist.
+		           The event is patched with a new message, the check MUST have the update message.
+		           The event is deleted and MUST NOT show up when listing all events.
+	*/
+	framework.ConformanceIt("should ensure that an event can be fetched, patched, deleted, and listed", func() {
```
wojtek-t, line 48: It not yet conformance.

chelseychen: Changed it to `ginkgo.It`

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/onsi/ginkgo"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	eventRetryPeriod  = 1 * time.Second
+	eventRetryTimeout = 1 * time.Minute
+)
+
+var _ = ginkgo.Describe("[sig-instrumentation] Events", func() {
```
wojtek-t, line 38: Let's introduce SIGDescribe the same way we do in other packages, e.g.:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/framework.go

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,191 @@
...
+	f := framework.NewDefaultFramework("events")
+
+	/*
+			   Release : v1.19
```
wojtek-t, line 42: nit: let's reduce the indentation - simply 2 tabs

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,191 @@
...
+		           The event is patched with a new message, the check MUST have the update message.
+		           The event is deleted and MUST NOT show up when listing all events.
+	*/
+	ginkgo.It("should ensure that an event can be fetched, patched, deleted, and listed", func() {
+		eventTestName := "event-test"
+		someTime := metav1.MicroTime{Time: time.Unix(1505828956, 0)}
+
+		ginkgo.By("creating a test event")
+		// create a test event in test namespace
+		_, err := f.ClientSet.EventsV1beta1().Events(f.Namespace.Name).Create(context.TODO(), &v1beta1.Event{
```
wojtek-t, line 54: Let's create a helper function like
```
func testEvent() *v1beta1.Event {
  return &v1beta1.Event{
    // what you have here
  }
}
```

and use it both here and below

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,191 @@
...
+			   Description: Create an event, the event MUST exist.
+		           The event is patched with a new message, the check MUST have the update message.
+		           The event is deleted and MUST NOT show up when listing all events.
+	*/
+	ginkgo.It("should ensure that an event can be fetched, patched, deleted, and listed", func() {
+		eventTestName := "event-test"
+		someTime := metav1.MicroTime{Time: time.Unix(1505828956, 0)}
+
+		ginkgo.By("creating a test event")
+		// create a test event in test namespace
```
wojtek-t, line 53: let's create a variable for the client - then promoting to v1 will be much smaller change, i.e.
```
client := f.ClientSet.EventsV1beta1()
```

and use it everywhere below

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,191 @@
...
+		framework.ExpectNoError(err, "failed to marshal the patch JSON payload")
+
+		_, err = f.ClientSet.EventsV1beta1().Events(f.Namespace.Name).Patch(context.TODO(), eventTestName, types.StrategicMergePatchType, []byte(eventPatch), metav1.PatchOptions{})
+		framework.ExpectNoError(err, "failed to patch the test event")
+
+		ginkgo.By("getting the test event")
+		// get event by name
+		event, err := f.ClientSet.EventsV1beta1().Events(f.Namespace.Name).Get(context.TODO(), eventTestName, metav1.GetOptions{})
+		framework.ExpectNoError(err, "failed to fetch the test event")
+		framework.ExpectEqual(event.Note, eventPatchNote, "test event note does not match patch note")
```
wojtek-t, line 101: We should probably compare the whole event to ensure that other stuff wasn't overwritten?

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,191 @@
...
+			Reason:              "Test",
+			Type:                "Normal",
+			ReportingController: "test-controller",
+			ReportingInstance:   "test-node",
+		}, metav1.CreateOptions{})
+		framework.ExpectNoError(err, "failed to create test event")
+
+		ginkgo.By("listing all events in all namespaces")
+		// get a list of Events in all namespaces to ensure endpoint coverage
+		eventsList, err := f.ClientSet.EventsV1beta1().Events("").List(context.TODO(), metav1.ListOptions{
```
wojtek-t, line 72: s/""/metav1.NamespaceAll/

wojtek-t: same in places below

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,191 @@
...
+		framework.ExpectEqual(event.Note, eventPatchNote, "test event note does not match patch note")
+
+		ginkgo.By("deleting the test event")
+		// delete original event
+		err = f.ClientSet.EventsV1beta1().Events(f.Namespace.Name).Delete(context.TODO(), eventTestName, metav1.DeleteOptions{})
+		framework.ExpectNoError(err, "failed to delete the test event")
+
+		ginkgo.By("listing all events in all namespaces")
+		// get a list of Events in all namespace
+		eventsList, err = f.ClientSet.EventsV1beta1().Events("").List(context.TODO(), metav1.ListOptions{
```
wojtek-t, line 110: let's also test listing from our namespace

test/e2e/instrumentation/events.go
wojtek-t: based on things below it should probably take name as a parameter

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,191 @@
...
+
+		ginkgo.By("delete a list of events")
+		// delete a list of events
+		framework.Logf("requesting DeleteCollection of events")
+		err = f.ClientSet.EventsV1beta1().Events(f.Namespace.Name).DeleteCollection(context.TODO(), metav1.DeleteOptions{}, metav1.ListOptions{
+			LabelSelector: "testevent-set=true"})
+		framework.ExpectNoError(err, "failed to delete the test event")
+
+		ginkgo.By("check that the list of events matches the requested quantity")
+		err = wait.PollImmediate(eventRetryPeriod, eventRetryTimeout, checkEventListQuantity(f, "testevent-set=true", 0))
```
wojtek-t, line 165: instead of introducing a dedicate function, maybe just:

```
pollFn := func() (bool, error) {
  eventList, err := client.Events(f.Namespace.Name).List(context.TODO(), ...)
  if err != nil {
    return false, nil
  }
  return len(eventList.Items) == quantity, nil
}
```

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,191 @@
...
+		framework.ExpectNoError(err, "failed to count required events")
+	})
```
wojtek-t, line 167: and then use it here

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+		Testname: Event resource lifecycle
+		Description: Create an event, the event MUST exist.
+		The event is patched with a new note, the check MUST have the update note.
+		The event is deleted and MUST NOT show up when listing all events.
+	*/
+	ginkgo.It("should ensure that an event can be fetched, patched, deleted, and listed", func() {
+		eventTestName := "event-test"
+
+		ginkgo.By("creating a test event")
+		// create a test event in test namespace
```
wojtek-t, line 90: remove this comment - the "ginkkgo.By" above already says that

in general: avoid comments saying "what am I doing" in cases where it is obvious from the code

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+		_, err := client.Create(context.TODO(), testEvent(eventTestName, "testevent-constant"), metav1.CreateOptions{})
+		framework.ExpectNoError(err, "failed to create test event")
+
+		ginkgo.By("listing all events in all namespaces")
+		// get a list of Events in all namespaces to ensure endpoint coverage
```
wojtek-t, line 95: same here (and probably in other places too)

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+		EventTime:           someTime,
+		Note:                "This is " + name,
+		Reason:              "Test",
+		Type:                "Normal",
+		ReportingController: "test-controller",
+		ReportingInstance:   "test-node",
+	}
+}
+
+func listEvents(name, namespace string, client eventsv1beta1.EventInterface) bool {
```
wojtek-t, line 58: listEvents name is misleading

How about:
```
func eventExistsInList(client eventsv1beta1.EventInterface, namespace, name string) bool {
  ...
}
```

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+	eventsList, err := client.List(context.TODO(), metav1.ListOptions{
+		LabelSelector: "testevent-constant=true",
+	})
+	framework.ExpectNoError(err, "failed list all events")
+
+	foundCreatedEvent := false
```
wojtek-t, line 64: let's remove this variable - see my comments below

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+	for _, val := range eventsList.Items {
+		if val.ObjectMeta.Name == name && val.ObjectMeta.Namespace == namespace {
+			foundCreatedEvent = true
```
wojtek-t, line 67: return true

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+			break
+		}
+	}
+	return foundCreatedEvent
```
wojtek-t, line 71: return false

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+
+		ginkgo.By("creating a test event")
+		// create a test event in test namespace
+		_, err := client.Create(context.TODO(), testEvent(eventTestName, "testevent-constant"), metav1.CreateOptions{})
+		framework.ExpectNoError(err, "failed to create test event")
+
+		ginkgo.By("listing all events in all namespaces")
+		// get a list of Events in all namespaces to ensure endpoint coverage
+		foundCreatedEvent := listEvents(eventTestName, f.Namespace.Name, clientAllNamespaces)
+		framework.ExpectEqual(foundCreatedEvent, true, "unable to find the test event")
```
wojtek-t, line 97: failed to find test event in list with cluster scope

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+
+		ginkgo.By("listing all events in framework namespace")
+		// get a list of Events in framework namespace to ensure endpoint coverage
+		foundCreatedEvent = listEvents(eventTestName, f.Namespace.Name, client)
+		framework.ExpectEqual(foundCreatedEvent, true, "unable to find the test event")
```
wojtek-t, line 102: failed to find test event in list with namespace scope

test/e2e/instrumentation/events.go
wojtek-t: Yeah - it seems all the comments are useless.

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+
+	/*
+		Release : v1.19
+		Testname: Event resource lifecycle
+		Description: Create an event, the event MUST exist.
+		The event is patched with a new note, the check MUST have the update note.
+		The event is deleted and MUST NOT show up when listing all events.
+	*/
+	ginkgo.It("should ensure that an event can be fetched, patched, deleted, and listed", func() {
+		eventTestName := "event-test"
```
wojtek-t, line 87: let's call it eventName

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/onsi/ginkgo"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	eventRetryPeriod  = 1 * time.Second
+	eventRetryTimeout = 1 * time.Minute
+)
+
+func testEvent(name, label string) *v1beta1.Event {
```
wojtek-t, line 40: let's call it `newTestEvent()`

[will let us avoid naming collisions]

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+		framework.ExpectNoError(err, "failed to marshal the patch JSON payload")
+
+		_, err = client.Patch(context.TODO(), eventTestName, types.StrategicMergePatchType, []byte(eventPatch), metav1.PatchOptions{})
+		framework.ExpectNoError(err, "failed to patch the test event")
+
+		ginkgo.By("getting the test event")
+		// get event by name
+		event, err := client.Get(context.TODO(), eventTestName, metav1.GetOptions{})
+		framework.ExpectNoError(err, "failed to fetch the test event")
+		framework.ExpectEqual(event.Note, eventPatchNote, "test event note does not match patch note")
```
wojtek-t, line 119: Sorry - I should have been more clear. Let's make something like:
```
// at the beginning of the test
testEvent := newTestEvent(...)
...
// here
testEvent.Note = eventPatchNote
// clear ResourceVersion field which is on API calls
event.ObjectMeta.ResourceVersion = ""
if !apiequality.Semantic.DeepEqual(testEvent, event) {
  framework.Failf("test event wasn't properly patched: %v", diff.ObjectReflectDiff(testEvent, event)
}
```

[diff is from "k8s.io/apimachinery/pkg/util/diff" ]
```

File: test/e2e/instrumentation/events.go [hunk 22]
```diff
@@ -0,0 +1,182 @@
...
+		// create a test event in test namespace
+		_, err := client.Create(context.TODO(), testEvent(eventTestName, "testevent-constant"), metav1.CreateOptions{})
+		framework.ExpectNoError(err, "failed to create test event")
+
+		ginkgo.By("listing all events in all namespaces")
+		// get a list of Events in all namespaces to ensure endpoint coverage
+		foundCreatedEvent := listEvents(eventTestName, f.Namespace.Name, clientAllNamespaces)
+		framework.ExpectEqual(foundCreatedEvent, true, "unable to find the test event")
+
+		ginkgo.By("listing all events in framework namespace")
```
wojtek-t, line 99: s/framework namespace/test namespace/

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+		// delete original event
+		err = client.Delete(context.TODO(), eventTestName, metav1.DeleteOptions{})
+		framework.ExpectNoError(err, "failed to delete the test event")
+
+		ginkgo.By("listing all events in all namespaces")
+		// get a list of Events in all namespace
+		foundCreatedEvent = listEvents(eventTestName, f.Namespace.Name, clientAllNamespaces)
+		framework.ExpectEqual(foundCreatedEvent, false, "should not have found test event after deletion")
+
+		ginkgo.By("listing all events in framework namespace")
```
wojtek-t, line 136: s/framework namespace/test namespace/

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+		// get a list of Events in framework namespace
+		foundCreatedEvent = listEvents(eventTestName, f.Namespace.Name, client)
+		framework.ExpectEqual(foundCreatedEvent, false, "should not have found test event after deletion")
+	})
+
+	ginkgo.It("should delete a collection of events", func() {
+		eventTestNames := []string{"test-event-1", "test-event-2", "test-event-3"}
```
wojtek-t, line 143: let's call it eventNames

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+
+		ginkgo.By("Create set of events")
+		// create a test event in test namespace
+		for _, eventTestName := range eventTestNames {
+			_, err := client.Create(context.TODO(), testEvent(eventTestName, "testevent-set"), metav1.CreateOptions{})
+			framework.ExpectNoError(err, "failed to create event")
+			framework.Logf("created %v", eventTestName)
```
wojtek-t, line 150: Let's remove - if it wouldn't be created the above would return an error.

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+
+		ginkgo.By("get a list of Events with a label in the current namespace")
+		// get a list of events
+		eventList, err := client.List(context.TODO(), metav1.ListOptions{
+			LabelSelector: "testevent-set=true",
+		})
+		framework.ExpectNoError(err, "failed to get a list of events")
+		framework.ExpectEqual(len(eventList.Items), len(eventTestNames), "event list should have 3 items")
```
wojtek-t, line 159: fmt.Sprintf("unexpected event list: %#v", eventList)

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+	"k8s.io/kubernetes/test/e2e/framework"
+	"k8s.io/kubernetes/test/e2e/instrumentation/common"
+
+	"github.com/onsi/ginkgo"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	eventRetryPeriod  = 1 * time.Second
+	eventRetryTimeout = 1 * time.Minute
```
wojtek-t, line 37: It seems those constants won't be needed.

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+
+		ginkgo.By("delete a list of events")
+		// delete a list of events
+		framework.Logf("requesting DeleteCollection of events")
+		err = client.DeleteCollection(context.TODO(), metav1.DeleteOptions{}, metav1.ListOptions{
+			LabelSelector: "testevent-set=true"})
+		framework.ExpectNoError(err, "failed to delete the test event")
+
+		ginkgo.By("check that the list of events matches the requested quantity")
+		pollFn := func() (bool, error) {
```
wojtek-t, line 169: Actually - you don't need to this. DeleteCollection is synchronous:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/registry/generic/registry/store.go#L1066

You can simply have:
eventList, err := client.List(...)
framework.ExpectNoError(err, "failed to list events")
framework.ExpectEqual(len(eventList.Items), 0, fmt.Sprintf("unexpected event list: %#v", eventList))

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,182 @@
...
+			})
+			if err != nil {
+				return false, nil
+			}
+			return len(eventList.Items) == 0, nil
+		}
+		err = wait.PollImmediate(eventRetryPeriod, eventRetryTimeout, pollFn)
+		framework.ExpectNoError(err, "failed to count required events")
+	})
+
```
wojtek-t, line 181: nit: remove this empty line

wojtek-t (commented):
I added a lot of comments, but they shouldn't be hard to address - I think we're relatively close to have it mergable.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Good to know. Thanks!

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

test/e2e/instrumentation/events.go
chelseychen: Done.

chelseychen:
Most tests failed and looks like there're mainly two things they're complaining about:
1. conformance rule cannot build successfully.
```
/home/prow/go/src/github.com/kubernetes/kubernetes/test/conformance/BUILD:43:1: Getting all conformance spec summaries. //test/conformance:list_conformance_specs failed (Exit 1): bash failed: error executing command 
```
The error is not informative enough, so I couldn't figure out the reason.

2. `panic: runtime error: invalid memory address or nil pointer dereference`
Looks like something wrong with line 69 / 71 in `test/e2e/instrumentation/events.go`, but again - the error is not informative, it didn't tell what is exactly wrong.

@wojtek-t Have you ever seen something like this before?

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,159 @@
...
+		ReportingController: "test-controller",
+		ReportingInstance:   "test-node",
+	}
+}
+
+func eventExistsInList(client eventsv1beta1.EventInterface, namespace, name string) bool {
+	eventsList, err := client.List(context.TODO(), metav1.ListOptions{
+		LabelSelector: "testevent-constant=true",
+	})
+	framework.ExpectNoError(err, "failed list all events")
```
wojtek-t, line 59: nit: "failed listing events"

[you're not listing all, because you use label selector]

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,159 @@
...
+	*/
+	ginkgo.It("should ensure that an event can be fetched, patched, deleted, and listed", func() {
+		eventName := "event-test"
+		testEvent := newTestEvent(eventName, "testevent-constant")
+
+		ginkgo.By("creating a test event")
+		_, err := client.Create(context.TODO(), testEvent, metav1.CreateOptions{})
+		framework.ExpectNoError(err, "failed to create test event")
+
+		ginkgo.By("listing all events in all namespaces")
```
wojtek-t, line 89: nit: "listing events from all namespaces"

[see my comment above for explanation]

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,159 @@
...
+		foundCreatedEvent := eventExistsInList(clientAllNamespaces, f.Namespace.Name, eventName)
+		framework.ExpectEqual(foundCreatedEvent, true, "failed to find test event in list with cluster scope")
+
+		ginkgo.By("listing all events in test namespace")
```
wojtek-t, line 93: nit: "listing events from test namespace"

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,159 @@
...
+		testEvent.Note = eventPatchNote
+		if !apiequality.Semantic.DeepEqual(testEvent, event) {
+			framework.Failf("test event wasn't properly patched: %v", diff.ObjectReflectDiff(testEvent, event))
+		}
+
+		ginkgo.By("deleting the test event")
+		err = client.Delete(context.TODO(), eventName, metav1.DeleteOptions{})
+		framework.ExpectNoError(err, "failed to delete the test event")
+
+		ginkgo.By("listing all events in all namespaces")
```
wojtek-t, line 120: same comments as above here

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,159 @@
...
+		if val.ObjectMeta.Name == name && val.ObjectMeta.Namespace == namespace {
+			return true
+		}
+	}
+	return false
+}
+
+var _ = common.SIGDescribe("Events API", func() {
+	f := framework.NewDefaultFramework("events")
+	client := f.ClientSet.EventsV1beta1().Events(f.Namespace.Name)
```
wojtek-t, line 71: this seem to be causing crashes.
Instead let's do what we generally do in tests:
```
f := framework.NewDefaultFramework("events")
var client eventsv1beta1.EventInterface
var clientAllNamespaces eventsv1beta1.EventInterface

ginkgo.BeforeEach(func() {
  client = f.ClientSet.EventsV1beta1().Events(f.Namespace.Name)
  clientAllNAmespaces = f.ClientSetEventsV1beta1().Events(metav1.NamespaceAll)
})

Comment (wojtek-t) [reply test/e2e/instrumentation/events.go:71]: Not sure yet if this is causing bazel failures too.

File: test/e2e/instrumentation/events.go (hunk 30 above)
Comment (chelseychen) [reply test/e2e/instrumentation/events.go]: Done.

File: test/e2e/instrumentation/events.go (hunk 32 above)
Comment (chelseychen) [reply test/e2e/instrumentation/events.go]: Done.

File: test/e2e/instrumentation/events.go (hunk 33 above)
Comment (chelseychen) [reply test/e2e/instrumentation/events.go]: Done.

File: test/e2e/instrumentation/events.go (hunk 31 above)
Comment (chelseychen) [reply test/e2e/instrumentation/events.go]: Done.

File: test/e2e/instrumentation/events.go (hunk 34 above)
Comment (chelseychen) [reply test/e2e/instrumentation/events.go]: Done. Let's see if the tests pass :)

File: test/e2e/instrumentation/events.go [hunk 35]
```diff
@@ -0,0 +1,169 @@
...
+		})
+		framework.ExpectNoError(err, "failed to marshal the patch JSON payload")
+
+		_, err = client.Patch(context.TODO(), eventName, types.StrategicMergePatchType, []byte(eventPatch), metav1.PatchOptions{})
+		framework.ExpectNoError(err, "failed to patch the test event")
+
+		ginkgo.By("getting the test event")
+		event, err := client.Get(context.TODO(), eventName, metav1.GetOptions{})
+		framework.ExpectNoError(err, "failed to get test event")
+		event.ObjectMeta = testEvent.ObjectMeta
```
chelseychen, line 120: @wojtek-t 
Looks like `client.Create` will add a lot of info into ObjectMeta, which will cause `DeepEqual` fail. So here I skip the comparison of the whole ObjectMeta. I'm not sure if this is the best way to do it. Or maybe it's better to get the event after `client.Create` and compare that with patched event?

chelseychen: Emm I changed my mind - I think it might be better to include ObjectMeta, so I get `testEvent` after creation. Hope this can pass the test. Please correct me if I'm wrong.

test/e2e/instrumentation/events.go
```diff
@@ -0,0 +1,173 @@
...
+		_, err = client.Patch(context.TODO(), eventName, types.StrategicMergePatchType, []byte(eventPatch), metav1.PatchOptions{})
+		framework.ExpectNoError(err, "failed to patch the test event")
+
+		ginkgo.By("getting the test event")
+		event, err := client.Get(context.TODO(), eventName, metav1.GetOptions{})
+		framework.ExpectNoError(err, "failed to get test event")
+		// clear ResourceVersion field which is on API calls
+		event.ObjectMeta.ResourceVersion = ""
+		testEvent.Note = eventPatchNote
+		if !apiequality.Semantic.DeepEqual(testEvent, event) {
```
wojtek-t, line 126: This is causing legitimate failures now.

I missed that testEvent is set in line 107. So what you have to do is to:
- clear ResourceVersion both in event and test Event
- also clear the ManagedField in both

So change to:
```
// clear ResourceVersion and ManagedFields which are set by control-plane
event.ObjectMeta.ResourceVersion = ""
testEvent.ObjectMeta.ResourceVersion = ""
event.ObjectMeta.ManagedFields = nil
testEvent.ObjectMeta.ManagedFields = nil

testEvent.Note = eventPatchNote
if ... (as is now)
```

wojtek-t (commented):
LGTM modulo the problem causing test failures. Will tag once fixed.

/approve

wojtek-t:
/lgtm
/retest

Thanks!

test/e2e/instrumentation/events.go
chelseychen: Done.

Issue #92521, opened 2020-06-25
consider setting GODEBUG=asyncpreemptoff=1


liggitt:
**What happened**:

Kubernetes is poised to ship 1.19 based on go1.14 (as of https://github.com/kubernetes/kubernetes/pull/88638)

Per https://golang.org/doc/go1.14#runtime:

> Goroutines are now asynchronously preemptible. As a result, loops without function calls no longer potentially deadlock the scheduler or significantly delay garbage collection. This is supported on all platforms except windows/arm, darwin/arm, js/wasm, and plan9/*.
> 
> A consequence of the implementation of preemption is that on Unix systems, including Linux and macOS systems, programs built with Go 1.14 will receive more signals than programs built with earlier releases. This means that programs that use packages like syscall or golang.org/x/sys/unix will see more slow system calls fail with EINTR errors. Those programs will have to handle those errors in some way, most likely looping to try the system call again. For more information about this see man 7 signal for Linux systems or similar documentation for other systems.

Based on discussion in reported issues (golang/go#36644, golang/go#38033, golang/go#38836), it appears that some stdlib methods can also encounter EINTR errors more frequently, and do not handle them in go1.14.

A stdlib mitigation [is in place for go1.15](https://go-review.googlesource.com/c/go/+/232862/), but [will not be backported to go1.14](https://github.com/golang/go/issues/39026#issuecomment-627589315) due to impact.

The asynchronous preemption introduced in go1.14 can be disabled by setting `GODEBUG=asyncpreemptoff=1`.

At the very least, we should document this possibility and workaround. If we see evidence that this error is being encountered by our components, we should consider disabling asynchronous preemption in the images/environments/manifests we control until we get to go 1.15.

/sig release cluster-lifecycle
for awareness

/sig node api-machinery scheduling network
owners of long-running components

/sig docs

liggitt:
cc @dims @justaugustus

BenTheElder:
Is it possible for a binary to set this on itself at runtime?

liggitt:
nope, happens before main() is called:

https://github.com/golang/go/blob/master/src/runtime/proc.go#L526-L577

athenabot:
/triage unresolved

Comment `/remove-triage unresolved` when the issue is assessed and confirmed.

🤖 I am a bot run by vllry. 👩‍🔬

justaugustus:
FYI @kubernetes/release-engineering

BenTheElder:
[...]

PR #92611, opened 2020-06-29
Tolerate go 1.13 builds

liggitt:
**What type of PR is this?**
/kind cleanup

* Keep go.mod files at go1.13 for now
* Isolate go1.14-specific bits in build-tagged files

xref https://github.com/kubernetes/kubernetes/pull/92438#issuecomment-650977961 and https://github.com/kubernetes/kubernetes/issues/92521

```release-note
NONE
```

### DIFF
--- a/vendor/modules.txt
+++ b/vendor/modules.txt
@@ -1,18 +1,8 @@
 # bitbucket.org/bertimus9/systemstat v0.0.0-20180207000608-0eeff89b0690 => bitbucket.org/bertimus9/systemstat v0.0.0-20180207000608-0eeff89b0690
-## explicit
 bitbucket.org/bertimus9/systemstat
-# bitbucket.org/bertimus9/systemstat => bitbucket.org/bertimus9/systemstat v0.0.0-20180207000608-0eeff89b0690
 # cloud.google.com/go v0.51.0 => cloud.google.com/go v0.51.0
-# cloud.google.com/go => cloud.google.com/go v0.51.0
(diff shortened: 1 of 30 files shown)


vendor/modules.txt
```diff
@@ -1,18 +1,8 @@
 # bitbucket.org/bertimus9/systemstat v0.0.0-20180207000608-0eeff89b0690 => bitbucket.org/bertimus9/systemstat v0.0.0-20180207000608-0eeff89b0690
```
BenTheElder, line 1: what is the diff in this file about?

liggitt: `go mod vendor` writes more things to this file when the go.mod file indicates go1.14 is in use

BenTheElder: we are using 1.14 in CI etc. though right? won't this just get written back by the next person?

liggitt: it is based on the go directive in the go.mod file, not the go version used

mikedanese: I tend to assume that presubmits will catch incompatibilities with CI (e.g. pull-kubernetes-verify in this case). :crossed_fingers:

dims:
Hmm, i thought the plan was to move to 1.15 golang when it ships. Are we going back?

liggitt:
> Are we going back?

No, just tweaking the go.mod files and published libraries to allow library consumers to continue compiling with go1.13 if they wish, given the 1.14 oddities. I still plan to try to pick up the 1.15 rc when available. 

/hold cancel

dims:
gotcha. thanks @liggitt

PR #90944, opened 2020-05-10
Update and improve ConfigMap resource lifecycle test

riaankleinhans:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
Adds more checks throughout the test:

* created
* modified
* deleted

**Which issue(s) this PR fixes**:
Fixes #90921
Resolves #90957 

[#88034 (comment)](https://github.com/kubernetes/kubernetes/pull/88034#discussion_r384103741)
(https://github.com/kubernetes/kubernetes/pull/88034#issuecomment-626251014)

**Special notes for your reviewer**:
Should fix flakiness

**Does this PR introduce a user-facing change?**:

```
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```
NONE
```

/sig testing

PR #92621, opened 2020-06-29
Simplify ConfigMap lifecycle e2e test

spiffxp:
**What type of PR is this?**
/kind cleanup

Simplifies the ConfigMap lifecycle test targeted for Conformance to avoid using watches

I'm presenting this as an alternative to the current ConfigMap lifecycle e2e test, which was updated via https://github.com/kubernetes/kubernetes/pull/90944 to include more complexity than I believe to be necessary.

I broke this down commit by commit to try and explain why I don't think the extra complexity is needed. See the commit messages for details

```release-note
NONE
```

/cc @hh @riaankl @BobyMCbobs


BobyMCbobs:
Whilst, there's no kind of patching followed by updating in this test, which would lead to failure of updates due to resource version incompatibilities (where the watch comes in handy, to check that the resource has infact updated and has the values it should) - in this regard it might not be needed.
However, I'd suggest that an important part of the watch tooling is confirming that those events happened (from another source) and in their order.

spiffxp:
>in this regard it might not be needed.

Which was the point I was trying to make with this PR.  I'm not sure such complexity is warranted in all cases.

>an important part of the watch tooling is confirming that those events happened (from another source) and in their order.

I personally am not sold on watch events being generated as part of the API contract for kubernetes.  If it is:
- how do we represent that coverage, so we can identify all of the other parts of kubernetes that don't currently have this covered by tests?
- can we please do this coverage using the stock functions provided by client-go's watch tools or via polling? I feel like the copy-pasted-and-subtly-modified version that ended up in framework is not gaining us anything here, and I tried to explain why in the individual commits

Riaankl:
@spiffxp I agree, let's create a ticket to remove the watch tooling and try to solicit tighter feedback in the future.

PR #92631, opened 2020-06-30
add endpoint-coverage.yaml to testdata

zachmandeville:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design

/kind documentation
> /kind failing-test
> /kind feature
> /kind flake

  This PR introduces a new yaml file in test/conformance/testdata that shows all current, non-deprecated api endpoints, whether they are covered by tests and conformance tests, and which tests hit them.  It also includes a list of the
current tests in the e2e suite and, if they are part of conformance, when they were promoted to conformance.

  Currently, one of the main metrics used for measuring conformance coverage is the number of endpoints hit by conformance tests.  If we want to increase coverage by this metric, it is useful to know which endpoints are already covered, so we can focus our test-writing toward the remaining, uncovered endpoints.

  However, The methods for counting covered endpoints is currently obscure.  APISnoop offers it through a webapp visualization, but it could be useful to have it as raw data one could query directly.
  The way APISnoop gets this data is to pull audit logs from an e2e test run and insert them into a postgres database.  It then combines the events with the current open api spec to list the current endpoints and whether they were hit
by tests.  This PR shows the result of taking this APISnoop query and outputting it as json or yaml.  From this one could use the coverage information in their own webapp, or to get results directly through a tool like `jq` or `yq`.

  With this file, one can more quickly build reports or apps that investigate conformance coverage, since these apps would not require their own database or data-processing.  This could be especially useful for prowbots that want to
check a pull request against existing coverage to ensure testing is not reduced.

_Caveat_: This PR brings in the end file, but not the process that creates it.  The intention of the PR is to discuss the usefulness of including this information in conformance/testdata, and the shape this information should take.

Fixes #

```release-note
NONE
```

```docs

```
/hold


zachmandeville:
CLA is signed, but was using incorrect address. Have updated my account accordingly. please recheck, friendly k8s-ci-robot!

neolit123:
@zachmandeville hi, thank you for the PR/idea. my take is that this is definitely useful but the maintenance seems difficult - who or what process is going to keep this file up-to-date?

unless this is automated the file will contain outdated information all the time.

/cc @hh @spiffxp

hh:
We have a container that can create and update this data. There has been some suggestions that while useful it be maintained out of tree. I'll bring it up on the next sig arch meeting.

PR #92643, opened 2020-06-30
conntrack e2e test debug connections

aojea:
adding more debug to the conntrack tests, it turns out some jobs like KIND are not logging the IP addresses of services, pods and endpoints so the conntrack table dump is useless :shrug: 

xref #91236

```release-note
NONE
```

### DIFF
--- a/test/e2e/network/conntrack.go
+++ b/test/e2e/network/conntrack.go
@@ -120,6 +120,7 @@ var _ = SIGDescribe("Conntrack", func() {
 		// TODO(#91236): Remove once the test is debugged and fixed.
 		// dump conntrack table for debugging
 		defer dumpConntrack(cs)
+		defer dumpIptables(cs)
 
 		// Create a NodePort service
 		udpJig := e2eservice.NewTestJig(cs, ns, serviceName)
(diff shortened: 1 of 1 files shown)


aojea:
/assign @wojtek-t 
/hold
only just to see the logs and check that all the info needed is there,

wojtek-t:
/lgtm
/approve

@aojea - please cancel hold when you will be ready

test/e2e/network/conntrack.go
```diff
@@ -273,6 +301,7 @@ var _ = SIGDescribe("Conntrack", func() {
 			framework.ExpectNoError(err)
 			framework.Logf("Pod client logs: %s", logs)
 			framework.Failf("Failed to connect to backend 2")
+
```
wojtek-t, line 304: nit: remove this empty line if you're going to submit

aojea: done

aojea:
pull-kubernetes-e2e-kind-ipv6 — Job failed.

has a conntrack failure, endpoints are published but there is not traffic from the client pod to the new endpoint
We should verify that the iptables rules are there, so the client-pod traffic is DNATed to the new pod
/hold

aojea:
test is passing but I don't fully understand the logic with the NodePort in GCE, the nat reply dst is 10.64.1.1:18222 :thinking: 
That's weird, in KIND jobs is replying to the original IP, is like it is double natting or something

```
grep 10.64.3.190 build-log.txt | grep -v Jul
I0701 11:14:47.120] udp      17 26 src=10.64.3.190 dst=10.40.0.3 sport=12345 dport=31361 [UNREPLIED] src=10.40.0.3 dst=10.64.3.190 sport=31361 dport=12345 mark=0 use=1
I0701 11:14:47.136] udp      17 29 src=10.64.3.190 dst=10.40.0.3 sport=12345 dport=31361 [UNREPLIED] src=10.40.0.3 dst=10.64.3.190 sport=31361 dport=12345 mark=0 use=1
I0701 11:14:47.142] udp      17 27 src=10.64.3.190 dst=10.40.0.3 sport=12345 dport=31361 src=10.64.1.227 dst=10.64.1.1 sport=80 dport=18222 mark=0 use=1
```

aojea:
now that we fixed the iptables lock problem we can start debugging this again
/test all

aojea:
/retest
endpoints tests failures :eyes:

aojea:
sig-cli failures
/retest

saschagrunert (approved):
/lgtm

aojea:
/retest
[sig-storage] PersistentVolumes-local [Volume type: dir-link] Two pods mounting a local volume one after the other should be able

PR #92644, opened 2020-06-30
Wait for all informers to sync in /readyz.

wojtek-t:
Based on https://github.com/kubernetes/kubernetes/pull/92508 by @mborsz 

**What type of PR is this?**
/kind bug

Context: https://github.com/kubernetes/kubernetes/issues/92506

In large clusters, informers in kube-apiserver may require significant time (30-50s) to initialize. Before this happens, kube-apiserver is not able to answer some requests (e.g. node authorizer is not able to accept any request).

This PR adds a way to determine if informers in kube-apiserver are already synced. This check is added to /readyz which can be used to determine if traffic can be sent to particular kube-apiserver.

Ref: https://github.com/kubernetes/kubernetes/issues/92506

```release-note
Extend kube-apiserver /readyz with new "informer-sync" check ensuring that internal informers are synced.
```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/server/config.go
+++ b/staging/src/k8s.io/apiserver/pkg/server/config.go
@@ -605,13 +605,20 @@ func (c completedConfig) New(name string, delegationTarget DelegationTarget) (*G
 	}
 
 	genericApiServerHookName := "generic-apiserver-start-informers"
-	if c.SharedInformerFactory != nil && !s.isPostStartHookRegistered(genericApiServerHookName) {
-		err := s.AddPostStartHook(genericApiServerHookName, func(context PostStartHookContext) error {
-			c.SharedInformerFactory.Start(context.StopCh)
-			return nil
-		})
-		if err != nil {
-			return nil, err
+	if c.SharedInformerFactory != nil {
+		if !s.isPostStartHookRegistered(genericApiServerHookName) {
+			err := s.AddPostStartHook(genericApiServerHookName, func(context PostStartHookContext) error {
+				c.SharedInformerFactory.Start(context.StopCh)
+				return nil
+			})
+			if err != nil {
+				return nil, err
+			}
+			// TODO: Once we get rid of /healthz consider changing this to post-start-hook.
+			err = s.addReadyzChecks(healthz.NewInformerSyncHealthz(c.SharedInformerFactory))
+			if err != nil {
+				return nil, err
+			}
 		}
 	}
 
--- a/staging/src/k8s.io/apiserver/pkg/server/healthz/healthz.go
+++ b/staging/src/k8s.io/apiserver/pkg/server/healthz/healthz.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apiserver/pkg/endpoints/metrics"
 	"k8s.io/apiserver/pkg/server/httplog"
+	"k8s.io/client-go/informers"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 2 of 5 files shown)


wojtek-t:
To copy some context from the original PR:

@deads2k 
```
As I recall, we actually had this check in openshift and chose to remove it because it wasn't required for request safety and answering some requests generally worked better than answering no requests and failing a health check. /readyz would be less worrying from a "death cycle" point of view, but @sttts may still have opinions since he's currently looking at those.
```

@sttts 
```
Readyz sounds much better than healthz. This is not about being healthy. Would like to hear @logicalhan 's opinion. He reworked the logic there quite a bit few months ago.
```

@logicalhan 
```
Yeah /readyz sounds appropriate for this. We should probably add a release note stating that we are adding a readyz check which further bifurcates behavior from the deprecated healthz endpoint (as a word of warning).
```

test/integration/master/kube_apiserver_test.go
```diff
@@ -98,10 +98,15 @@ func TestRun(t *testing.T) {
...
 }
 
-func endpointReturnsStatusOK(client *kubernetes.Clientset, path string) bool {
-	res := client.CoreV1().RESTClient().Get().AbsPath(path).Do(context.TODO())
+func endpointReturnsStatusOK(t *testing.T, client *kubernetes.Clientset, path string) bool {
+	res := client.CoreV1().RESTClient().Get().RequestURI(path).Do(context.TODO())
 	var status int
 	res.StatusCode(&status)
+	raw, err := res.Raw()
+	if err != nil {
```
logicalhan, line 106: If we're going to do this, then we should probably just fully convert this function so that it doesn't return the bool. 

```
if err != nil || status != http.StatusOK {
   t.Fatalf("got %v but wanted 200, error: %v", status, err)
}
```

logicalhan: And then the calls on 121 and 124 don't have to be in conditional form.

staging/src/k8s.io/apiserver/pkg/server/healthz/healthz.go
```diff
@@ -81,6 +82,43 @@ func (l *log) Check(_ *http.Request) error {
...
+	var notStarted []string
+	for informType, started := range started {
+		if !started {
+			klog.Warningf("informer for %v has not synced yet", informType.String())
+			notStarted = append(notStarted, informType.String())
+		}
+	}
+	if len(notStarted) != 0 {
+		return fmt.Errorf("%d informers not started yet: %v", len(notStarted), notStarted)
+	}
```
logicalhan, line 118: this is a huge nit (sorry):

```
	stopCh := make(chan struct{})
	// Close stopCh to force checking if informers are synced now.
	close(stopCh)

	var informersByStarted map[bool][]string
	for informType, started := range i.sharedInformerFactory.WaitForCacheSync(stopCh) {
         informersByStarted[started] = append(informersByStarted[started], informType.String())
	}
	if len(informersByStarted[false]) != 0 {
		return fmt.Errorf("%d informers have not started yet: %v (but %v informers have started)", len(notStarted), informersByStarted[false], informersByStarted[true])
	}

```

logicalhan (commented):
/lgtm 

just a couple minor comments.

wojtek-t:
/hold
until I apply comments

staging/src/k8s.io/apiserver/pkg/server/config.go
```diff
@@ -605,12 +605,17 @@ func (c completedConfig) New(name string, delegationTarget DelegationTarget) (*G
...
+				c.SharedInformerFactory.Start(context.StopCh)
+				return nil
+			})
+			if err != nil {
+				return nil, err
+			}
+		}
+		if err := s.addReadyzChecks(healthz.NewInformerSyncHealthz(c.SharedInformerFactory)); err != nil {
 			return nil, err
 		}
```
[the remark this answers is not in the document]
wojtek-t: done

staging/src/k8s.io/apiserver/pkg/server/healthz/healthz.go
wojtek-t: done

test/integration/master/kube_apiserver_test.go
wojtek-t: changed the function to return error actually

wojtek-t (commented):
Comments applied

logicalhan (commented):
/lgtm

wojtek-t:
@logicalhan sorry - fixed one more typo - if you could re-lgtm

/hold cancel

wojtek-t:
Just fixed the minor compile error in integration test - reaaplying the label.

PR #92650, opened 2020-06-30
breakdown PodSchedulingDuration by number of attempts

ahg-g:
**What type of PR is this?**

/kind feature

Breakdown PodSchedulingDuration by number of attempts, this is useful when monitoring the scheduler to understand how scheduling attempts impact latency.

```release-note
Add attempts label to scheduler's PodSchedulingDuration metric.
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -612,15 +612,23 @@ func (sched *Scheduler) scheduleOne(ctx context.Context) {
 			if klog.V(2).Enabled() {
 				klog.InfoS("Successfully bound pod to node", "pod", klog.KObj(pod), "node", scheduleResult.SuggestedHost, "evaluatedNodes", scheduleResult.EvaluatedNodes, "feasibleNodes", scheduleResult.FeasibleNodes)
 			}
-
 			metrics.PodScheduled(prof.Name, metrics.SinceInSeconds(start))
 			metrics.PodSchedulingAttempts.Observe(float64(podInfo.Attempts))
-			metrics.PodSchedulingDuration.Observe(metrics.SinceInSeconds(podInfo.InitialAttemptTimestamp))
+			metrics.PodSchedulingDuration.WithLabelValues(getAttemptsLabel(podInfo)).Observe(metrics.SinceInSeconds(podInfo.InitialAttemptTimestamp))
 
 			// Run "postbind" plugins.
 			prof.RunPostBindPlugins(bindingCycleCtx, state, assumedPod, scheduleResult.SuggestedHost)
 		}
 	}()
+}
+
+func getAttemptsLabel(p *framework.QueuedPodInfo) string {
+	// We breakdown the pod scheduling duration by attempts capped to a limit
+	// to avoid ending up with a high cardinality metric.
+	if p.Attempts >= 15 {
+		return "15+"
+	}
+	return string(p.Attempts)
 }
 
 func (sched *Scheduler) profileForPod(pod *v1.Pod) (*profile.Profile, error) {
--- a/pkg/scheduler/metrics/metrics.go
+++ b/pkg/scheduler/metrics/metrics.go
@@ -154,7 +154,7 @@ var (
 			StabilityLevel: metrics.ALPHA,
 		}, []string{"work"})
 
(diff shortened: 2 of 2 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -612,10 +612,15 @@ func (sched *Scheduler) scheduleOne(ctx context.Context) {
 			if klog.V(2).Enabled() {
 				klog.InfoS("Successfully bound pod to node", "pod", klog.KObj(pod), "node", scheduleResult.SuggestedHost, "evaluatedNodes", scheduleResult.EvaluatedNodes, "feasibleNodes", scheduleResult.FeasibleNodes)
 			}
-
 			metrics.PodScheduled(prof.Name, metrics.SinceInSeconds(start))
 			metrics.PodSchedulingAttempts.Observe(float64(podInfo.Attempts))
-			metrics.PodSchedulingDuration.Observe(metrics.SinceInSeconds(podInfo.InitialAttemptTimestamp))
+
+			// We breakdown the pod scheduling duration by attempts (capped to a limit).
+			attempts := podInfo.Attempts
```
ahg-g, line 619: @logicalhan do we really need to cap? can I just use attempts as is?

pkg/scheduler/scheduler.go
```diff
@@ -612,10 +612,15 @@ func (sched *Scheduler) scheduleOne(ctx context.Context) {
...
+			if attempts > 50 {
+				attempts = 50
+			}
+			metrics.PodSchedulingDuration.WithLabelValues(string(attempts)).Observe(metrics.SinceInSeconds(podInfo.InitialAttemptTimestamp))
```
alculquicondor, line 623: Could you send a parent PR that changes the variable name `InitialAttemptTimestamp` to `QueueuingTimestamp` or something like that?

ahg-g: We should do that in a follow up PR to allow for patching this PR back if needed.

pkg/scheduler/scheduler.go
ahg-g: while I don't think the number of attempts will have a wide range of values, leaving it unbounded is certainly not a good idea.

ahg-g:
/assign @Huang-Wei 

Aldo is on vacation for the next couple of days.

pkg/scheduler/scheduler.go
Huang-Wei: How about explicitly expressing that internally we have already "bucked" the attempts (we can adjust the buckets like [5,10), [10, 20), [20, +inf): 

```go
metrics.PodSchedulingDuration.WithLabelValues(displayAttempts(podInfo.Attempts)).Observe(metrics.SinceInSeconds(podInfo.InitialAttemptTimestamp))

func displayAttempts(n int) string {
	if n < 5 {
		return string(n)
	} else if n < 20 {
		return "[5, 20)"
	}
	return "[20, +inf)"
}
```

Huang-Wei (commented):
LGTM generally. Just one comment.

pkg/scheduler/scheduler.go
ahg-g: The most important information from this is actually distinguishing pods scheduled from the first attempt, so perhaps we can change the label to first_attempt, and the value is true or false.

Huang-Wei: I'm inclined to the current version as it exposes more info (maybe not that useful for now), while also able to infer whether it's the first attempt or not.

ahg-g: sg, I increased the limit to 15, I think it is small enough cardinality, and will likely cover most cases. We should increase the limit for PodSchedulingAttempts, I think 5 is too low.

Huang-Wei: Shouldn't we display 15 as `[15, +inf)` or something similar? so that users would know it's not literally 15.

ahg-g: changed it to "15+", likely clear enough, sg?

Huang-Wei: SG.

ahg-g:
> /lgtm

sorry, updated the comment, can you lgtm again :)

PR #92651, opened 2020-06-30
[Kubeadm]:Bump CoreDNS to v1.7.0

rajansandeep:
**What type of PR is this?**
/kind bug

Updates CoreDNS to v1.7.0. The default Corefile now includes the `max_concurrent` option, which sets the maximum number of concurrent upstream queries allowed by the `forward` plugin.

Also updates the migration library to 1.0.10 which supports coredns v1.7.0

Fixes #

```release-note
ACTION REQUIRED : In CoreDNS v1.7.0, [metrics names have been changed](https://github.com/coredns/coredns/blob/master/notes/coredns-1.7.0.md#metric-changes) which will be backward incompatible with existing reporting formulas that use the old metrics' names. Adjust your formulas to the new names before upgrading. 

Kubeadm now includes CoreDNS version v1.7.0. Some of the major changes include:
-  Fixed a bug that could cause CoreDNS to stop updating service records.
-  Fixed a bug in the forward plugin where only the first upstream server is always selected no matter which policy is set.
-  Remove already deprecated options `resyncperiod` and `upstream` in the Kubernetes plugin.
-  Includes Prometheus metrics name changes (to bring them in line with standard Prometheus metrics naming convention). They will be backward incompatible with existing reporting formulas that use the old metrics' names.
-  The federation plugin (allows for v1 Kubernetes federation) has been removed.
More details are available in https://coredns.io/2020/06/15/coredns-1.7.0-release/
```

```docs

```

### DIFF
--- a/cmd/kubeadm/app/phases/addons/dns/dns_test.go
+++ b/cmd/kubeadm/app/phases/addons/dns/dns_test.go
@@ -585,7 +585,9 @@ func TestCreateCoreDNSConfigMap(t *testing.T) {
         ttl 30
     }
     prometheus :9153
-    forward . /etc/resolv.conf
+    forward . /etc/resolv.conf {
+        max_concurrent 1000
+    }
     cache 30
     loop
     reload
@@ -623,7 +625,9 @@ func TestCreateCoreDNSConfigMap(t *testing.T) {
         fallthrough in-addr.arpa ip6.arpa
     }
     prometheus :9153
-    forward . /etc/resolv.conf
+    forward . /etc/resolv.conf {
+        max_concurrent 1000
+    }
     k8s_external example.com
     cache 30
     loop
--- a/build/dependencies.yaml
+++ b/build/dependencies.yaml
@@ -22,7 +22,7 @@ dependencies:
       match: k8s.gcr.io/coredns
 
   - name: "coredns-kubeadm"
-    version: 1.6.7
+    version: 1.7.0
     refPaths:
     - path: cmd/kubeadm/app/constants/constants.go
       match: CoreDNSVersion =
--- a/cmd/kubeadm/app/constants/constants.go
+++ b/cmd/kubeadm/app/constants/constants.go
@@ -329,7 +329,7 @@ const (
 	KubeDNSVersion = "1.14.13"
 
 	// CoreDNSVersion is the version of CoreDNS to be deployed if it is used
-	CoreDNSVersion = "1.6.7"
+	CoreDNSVersion = "1.7.0"
 
 	// ClusterConfigurationKind is the string kind value for the ClusterConfiguration struct
 	ClusterConfigurationKind = "ClusterConfiguration"
--- a/cmd/kubeadm/app/phases/addons/dns/manifests.go
+++ b/cmd/kubeadm/app/phases/addons/dns/manifests.go
@@ -320,7 +320,9 @@ data:
            ttl 30
         }
         prometheus :9153
-        forward . {{ .UpstreamNameserver }}
+        forward . {{ .UpstreamNameserver }} {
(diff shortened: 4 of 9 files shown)


BenTheElder:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92651/pull-kubernetes-dependencies/1277983325801156608 is a real failure

cmd/kubeadm/app/phases/addons/dns/dns_test.go
```diff
@@ -660,7 +660,9 @@ func TestCreateCoreDNSConfigMap(t *testing.T) {
         ttl 30
     }
     prometheus :9153
-    forward . /etc/resolv.conf
+    forward . /etc/resolv.conf {
+        max_concurrent 1000
```
neolit123, line 664: should we also mention this change in the release note? i don't have good statistical context on how breaking it is, but i'm going to assume it affects larger clusters with a lot of queries?

neolit123:
last time i've checked, feature freeze was on the 9th of July.
/milestone v1.19
/approve

cmd/kubeadm/app/phases/addons/dns/dns_test.go
chrisohaver: The limit is per instance. The larger a cluster the more CoreDNS instances... So it should not change drastically at scale in theory.  For comparison, kube-dns (via it's default dnsmasq behavior), had a limit of 150 concurrent queries.  More details [here](https://github.com/coredns/deployment/pull/233#issue-438151578). CoreDNS never had a limit before this, it would just OOM if too many queries came in.

neolit123: ok, thanks for the details.

rajansandeep:
Looks like updating deps broke stuff, particularly the update of `gopkg.in/square/go-jose.v2 v2.2.2 => 
gopkg.in/square/go-jose.v2 v2.3.1` results in the error:

```
/b/f/w/pkg/serviceaccount/claims.go:93:27: cannot use jwt.NewNumericDate(now.Add(time.Duration(warnafter) * time.Second)) (type *jwt.NumericDate) as type jwt.NumericDate in assignment
```

Any suggestions on how to proceed?

cmd/kubeadm/app/phases/addons/dns/dns_test.go
chrisohaver: Regarding release notes: CoreDNS 1.7.0 does introduce some backward incompatible changes. Notably, it makes many Prometheus metrics name changes (to bring them in line with standard Prometheus metrics naming convention).  They will be backward incompatible with existing reporting formulas that use the old metrics names.  Less critically, 1.7.0 drops support for a couple of kubernetes plugin options which had already been deprecated and ignored by CoreDNS for several releases (i.e. now instead of ignoring them, CoreDNS will exit with an unknown directive error).

neolit123: i think we should briefly mention that in the k8s release note and even link to the coredns release notes.

rajansandeep: Yes, I’ll update the release notes.

rajansandeep:
@neolit123, I updated the release notes with "action required" for the metrics changes.
Would it be appropriate to include the workaround example in https://github.com/coredns/coredns/pull/3671#issuecomment-604673272 in the notes?

neolit123:
or just link to the example github comment - e.g. "A workaround can be found here - LINK".
@rosti WDYT?

liggitt:
cc @johnbelamaric for coredns breaking changes

/approve
for vendor bump

Issue #68026, opened 2018-08-29
Fix golint failures


dims:
Anyone itching to understand the code and try fixing things in Kubernetes, please help fixing the lint failures

Step 1: Remove one or two packages (related packages) from the list in - https://github.com/kubernetes/kubernetes/blame/master/hack/.golint_failures (note: do not change generated files, the name of defaulting functions, or the name of conversion functions... packages failing linting due to issues in those files will need to be left alone)
Step 2: Run `hack/verify-golint.sh` and look at the logged errors
Step 3: Fix the code to ensure that `hack/verify-golint.sh` runs cleanly.
Step 4: Run `hack/update-bazel.sh` and `hack/update-gofmt.sh` to make sure we fix any problems with gofmt or bazel.
Step 5: File a PR with the change
Step 6: Go to step #1 :)

Please use the tracking spreadsheet [here](https://docs.google.com/spreadsheets/d/1VhU6zCk-vaAnbu_XkgsIq5I7jVnezUquWBMhcIN6AJM/edit#gid=0) For everyone who is going to contribute, 
- please append the packages you picked at the end.
- For existing contributor, please make sure this list is update to date

fengzixu:
@dims  Hey, I would like to resolve this issue. Could you assign it to me ?

dims:
@fengzixu there are a lot of packages so there will be many people working on this. Also we cannot assign issue to folks who are not in the github org for kubernetes. please feel free to pick up a few packages like i mentioned and file a PR.

fengzixu:
OK. I see. I will probably pick up some failures to fix.

fengzixu:
Hi dims. @dims Should i submit my PR before i remove one or two packages (related packages) from the list in - https://github.com/kubernetes/kubernetes/blame/master/hack/.golint_failures ？ Or I first remove the package from list, then I fix the issue and submit a PR?

dims:
@fengzixu in your PR, you will have both the fixes to packages and some lines removed from /.golint_failures

dims:
@fengzixu 

- Say your GOPATH looks like this `GOPATH=/Users/dims/go` 
- Your kubernetes directory should be `/Users/dims/go/src/k8s.io/kubernetes` (`$GOPATH/src/k8s.io/kubernetes`)
- cd to `/Users/dims/go/src/k8s.io/kubernetes` and then run `hack/verify-golint.sh` from that directory.

fengzixu:
[...]

PR #92658, opened 2020-06-30
cleanup: fix golint errors in pkg/kubectl/cmd/convert

AndersonQ:
**What type of PR is this?**
/kind cleanup

Fix golint failures for pkg/kubectl/cmd/convert

Part of #68026

It's my first contribution to Kubernetes.

NONE

Release note:

```release-note
NONE
```

### DIFF
--- a/pkg/kubectl/cmd/convert/convert.go
+++ b/pkg/kubectl/cmd/convert/convert.go
@@ -60,8 +60,8 @@ var (
 		kubectl convert -f . | kubectl create -f -`))
 )
 
-// ConvertOptions have the data required to perform the convert operation
-type ConvertOptions struct {
+// Options have the data required to perform the convert operation
+type Options struct {
 	PrintFlags *genericclioptions.PrintFlags
 	Printer    printers.ResourcePrinter
 
@@ -76,8 +76,8 @@ type ConvertOptions struct {
 	genericclioptions.IOStreams
 }
 
-func NewConvertOptions(ioStreams genericclioptions.IOStreams) *ConvertOptions {
-	return &ConvertOptions{
+func newConvertOptions(ioStreams genericclioptions.IOStreams) *Options {
+	return &Options{
 		PrintFlags: genericclioptions.NewPrintFlags("converted").WithTypeSetter(scheme.Scheme).WithDefaultOutput("yaml"),
 		local:      true,
 		IOStreams:  ioStreams,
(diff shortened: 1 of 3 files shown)


pkg/kubectl/cmd/convert/convert.go
```diff
@@ -60,8 +60,8 @@ var (
 		kubectl convert -f . | kubectl create -f -`))
 )
 
-// ConvertOptions have the data required to perform the convert operation
-type ConvertOptions struct {
```
brianpursley, line 64: I know golint doesn't like this, but just about every kubectl command names its options this way (`<something>Options`):

```
~/go/src/k8s.io/kubernetes/staging/src/k8s.io/kubectl/pkg/cmd (master) $ grep -re "type .*Options struct" | wc -l
71
~/go/src/k8s.io/kubernetes/staging/src/k8s.io/kubectl/pkg/cmd (master) $ grep -re "type Options struct" | wc -l
1
```

I'm not saying it is wrong to fix the lint warnings, but my concern is that instead of having a single consistent way of naming options (that golint happens not to like) that we end up with inconsistent naming, unless we undertake a larger effort to rename all the option structs.

pkg/kubectl/cmd/convert/convert.go
```diff
@@ -76,8 +76,8 @@ type ConvertOptions struct {
 	genericclioptions.IOStreams
 }
 
-func NewConvertOptions(ioStreams genericclioptions.IOStreams) *ConvertOptions {
-	return &ConvertOptions{
+func newConvertOptions(ioStreams genericclioptions.IOStreams) *Options {
```
brianpursley, line 79: What is the reason for changing this from public to private?

pkg/kubectl/cmd/convert/convert.go
AndersonQ: Ok, it makes sense. However it isn't possible to exclude only this name from golinter.
Should we then keep this file on `hack/.golint_failures` and maybe put a comment there saying to keep this file and the others with Options `<something>Options` without for now?

brianpursley: Yeah, I guess golint doesn't have a way to override except at the file level. I'm OK with it the change, but I just sort of feel like we should follow through and change all the options if we change it in one. Maybe I'm thinking too much into it. I wonder if anyone else has any opinions on this...

pkg/kubectl/cmd/convert/convert.go
brianpursley: Agreed, it doesn't seem to be used anywhere else in the project. I guess it comes down to what our strategy is for making kubectl functionality usable by other applications, and do we want to expose this as the interface for creating options.

I will defer to others on the project about whether we want the NewXxxxxxxxxOptions() to be public or not.  Just the the other comment, it seems like they are public everywhere else.

PR #92659, opened 2020-06-30
api: add port protocol default to allow use in CRDs

sttts:
PodSpecs are not embeddable into CRDs today because there are optional map keys, e.g. in ports. This adds the annotation to be interpreted by controller-gen.

> /kind bug

```release-note
NONE
```


kensipe:
@sttts shouldn't this be back ported?

EnriqueL8:
@sttts @kensipe  Is there any plan to merge this?

sttts:
This should be resolved via `+default` markers.

/close

PR #92660, opened 2020-06-30
Add kubebuilder tags to Condition type

damemi:
**What type of PR is this?**
/kind api-change

This picks the changes from https://github.com/kubernetes/kubernetes/pull/92519 and adds kubebuilder validation tags for CRD generation

Fixes #

```release-note
Kubebuilder validation tags are set on metav1.Condition for CRD generation
```

```docs

```

related to https://github.com/kubernetes/enhancements/issues/1623

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto
+++ b/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto
@@ -153,22 +153,32 @@ message Condition {
   // ---
   // Many .condition.type values are consistent across resources like Available, but because arbitrary conditions can be
   // useful (see .node.status.conditions), the ability to deconflict is important.
+  // The regex it matches is (dns1123SubdomainFmt/)?(qualifiedNameFmt)
   // +required
+  // +kubebuilder:validation:Required
+  // +kubebuilder:validation:Pattern=`^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$`
+  // +kubebuilder:validation:MaxLength=316
   optional string type = 1;
 
   // status of the condition, one of True, False, Unknown.
   // +required
+  // +kubebuilder:validation:Required
+  // +kubebuilder:validation:Enum=True;False;Unknown
   optional string status = 2;
 
   // observedGeneration represents the .metadata.generation that the condition was set based upon.
   // For instance, if .metadata.generation is currently 12, but the .status.conditions[x].observedGeneration is 9, the condition is out of date
   // with respect to the current state of the instance.
   // +optional
+  // +kubebuilder:validation:Minimum=0
   optional int64 observedGeneration = 3;
 
   // lastTransitionTime is the last time the condition transitioned from one status to another.
   // This should be when the underlying condition changed.  If that is not known, then using the time when the API field changed is acceptable.
   // +required
+  // +kubebuilder:validation:Required
+  // +kubebuilder:validation:Type=string
+  // +kubebuilder:validation:Format=date-time
   optional Time lastTransitionTime = 4;
 
   // reason contains a programmatic identifier indicating the reason for the condition's last transition.
@@ -177,11 +187,17 @@ message Condition {
   // The value should be a CamelCase string.
   // This field may not be empty.
   // +required
+  // +kubebuilder:validation:Required
+  // +kubebuilder:validation:MaxLength=1024
+  // +kubebuilder:validation:MinLength=1
+  // +kubebuilder:validation:Pattern=`^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$`
   optional string reason = 5;
 
   // message is a human readable message indicating details about the transition.
   // This may be an empty string.
   // +required
+  // +kubebuilder:validation:Required
+  // +kubebuilder:validation:MaxLength=32768
   optional string message = 6;
 }
 
--- a/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
+++ b/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
@@ -1369,19 +1369,29 @@ type Condition struct {
 	// ---
 	// Many .condition.type values are consistent across resources like Available, but because arbitrary conditions can be
 	// useful (see .node.status.conditions), the ability to deconflict is important.
+	// The regex it matches is (dns1123SubdomainFmt/)?(qualifiedNameFmt)
 	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$`
+	// +kubebuilder:validation:MaxLength=316
 	Type string `json:"type" protobuf:"bytes,1,opt,name=type"`
 	// status of the condition, one of True, False, Unknown.
 	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=True;False;Unknown
 	Status ConditionStatus `json:"status" protobuf:"bytes,2,opt,name=status"`
 	// observedGeneration represents the .metadata.generation that the condition was set based upon.
 	// For instance, if .metadata.generation is currently 12, but the .status.conditions[x].observedGeneration is 9, the condition is out of date
 	// with respect to the current state of the instance.
 	// +optional
+	// +kubebuilder:validation:Minimum=0
 	ObservedGeneration int64 `json:"observedGeneration,omitempty" protobuf:"varint,3,opt,name=observedGeneration"`
 	// lastTransitionTime is the last time the condition transitioned from one status to another.
 	// This should be when the underlying condition changed.  If that is not known, then using the time when the API field changed is acceptable.
 	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Format=date-time
 	LastTransitionTime Time `json:"lastTransitionTime" protobuf:"bytes,4,opt,name=lastTransitionTime"`
 	// reason contains a programmatic identifier indicating the reason for the condition's last transition.
 	// Producers of specific condition types may define expected values and meanings for this field,
@@ -1389,9 +1399,15 @@ type Condition struct {
 	// The value should be a CamelCase string.
 	// This field may not be empty.
 	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=1024
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Pattern=`^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$`
 	Reason string `json:"reason" protobuf:"bytes,5,opt,name=reason"`
 	// message is a human readable message indicating details about the transition.
 	// This may be an empty string.
 	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=32768
 	Message string `json:"message" protobuf:"bytes,6,opt,name=message"`
 }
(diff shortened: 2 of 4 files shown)


staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
```diff
@@ -1367,28 +1368,39 @@ type Condition struct {
 	// Many .condition.type values are consistent across resources like Available, but because arbitrary conditions can be
 	// useful (see .node.status.conditions), the ability to deconflict is important.
 	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=Available;Progressing;Degraded
```
deads2k, line 1372: this is a regex in the validation code.

damemi: Yeah, I based this on the comment above the type. The validation code looks like it [uses different regexes conditionally](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L42) which I think is more complex than the regex marker can do alone

liggitt: using an enum here is not correct... far more types than Available, Progressing, Degraded are supported

The regex validation seems to be conditional based on whether the type is qualified or not. I think that can be expressed like this:
```
(DNS1123_PATTERN/)?(QUALIFIED_NAME_PATTERN)
```

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
```diff
@@ -1367,28 +1368,39 @@ type Condition struct {
...
 	Type string `json:"type" protobuf:"bytes,1,opt,name=type"`
 	// status of the condition, one of True, False, Unknown.
 	// +required
+	// +kubebuilder:validation:Required
 	Status ConditionStatus `json:"status" protobuf:"bytes,2,opt,name=status"`
```
liggitt, line 1377: this is an appropriate place for an enum (True, False, Unknown)

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
```diff
@@ -1367,28 +1368,39 @@ type Condition struct {
...
 	// observedGeneration represents the .metadata.generation that the condition was set based upon.
 	// For instance, if .metadata.generation is currently 12, but the .status.conditions[x].observedGeneration is 9, the condition is out of date
 	// with respect to the current state of the instance.
 	// +optional
+	// +kubebuilder:validation:Minimum=0
 	ObservedGeneration int64 `json:"observedGeneration,omitempty" protobuf:"varint,3,opt,name=observedGeneration"`
 	// lastTransitionTime is the last time the condition transitioned from one status to another.
 	// This should be when the underlying condition changed.  If that is not known, then using the time when the API field changed is acceptable.
 	// +required
+	// +kubebuilder:validation:Required
```
liggitt, line 1387: is there a time format available?

damemi: it looks like we can use [type: string and format: date-time](https://book.kubebuilder.io/reference/markers/crd-validation.html)

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
damemi: @liggitt Ah, I'm not great with complex regexes so I wasn't aware if it could handle conditionals

Since dns1123 pattern is `dns1123LabelFmt + "(\\." + dns1123LabelFmt + ")*"` and qualified name fmt is `"(" + qnameCharFmt + qnameExtCharFmt + "*)?" + qnameCharFmt`, I tried substituting these into `(DNS1123_PATTERN/)?(QUALIFIED_NAME_PATTERN)` (evaluated out). This ended up being pretty complex from what I can see, not sure if there's a good way to simplify it

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
```diff
@@ -1346,3 +1346,64 @@ type PartialObjectMetadataList struct {
...
+// }
+type Condition struct {
+	// type of condition in CamelCase or in foo.example.com/CamelCase.
+	// ---
+	// Many .condition.type values are consistent across resources like Available, but because arbitrary conditions can be
+	// useful (see .node.status.conditions), the ability to deconflict is important.
+	// The regex it matches is (dns1123SubdomainFmt/)?(qualifiedNameFmt)
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])
```
sttts, line 1372: this matches `asdfasfas.sadf.asdfas/bla$'_%`.

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
```diff
@@ -1346,3 +1346,64 @@ type PartialObjectMetadataList struct {
...
+	// reason contains a programmatic identifier indicating the reason for the condition's last transition.
+	// Producers of specific condition types may define expected values and meanings for this field,
+	// and whether the values are considered a guaranteed API.
+	// The value should be a CamelCase string.
+	// This field may not be empty.
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=1024
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Pattern=`^[A-Za-z_][A-Za-z0-9_]*$`
```
deads2k, line 1401: this got expanded `[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?`

sttts, line 1401: this doesn't match https://github.com/kubernetes/kubernetes/pull/92519/files#diff-ebee21b567c0f4e3f14894c1192dbd29R251

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
damemi: This is what I got manually expanding the regexes currently used by the validation function (https://github.com/kubernetes/kubernetes/pull/92660#discussion_r448441739) Did I miss something or write that out wrong?

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
damemi: Updated

damemi:
@deads2k @sttts think I addressed the comments here. There was some offline talk about trying to come up with a regex to match specifically non-zero rfc3339, is that still something we need or is a date-time format validation sufficient? (https://github.com/kubernetes/kubernetes/pull/92660#discussion_r448512878)

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
damemi: Updated (https://github.com/kubernetes/kubernetes/pull/92660#discussion_r449065755)

deads2k:
Can you rebase this on master please?  You've got an old version of the validation PR

damemi:
> Can you rebase this on master please? You've got an old version of the validation PR

Ah I did rebase it on master, I needed to rebase it on your branch from the validation PR. Updated and squashed

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
```diff
@@ -1369,29 +1369,44 @@ type Condition struct {
 	// ---
 	// Many .condition.type values are consistent across resources like Available, but because arbitrary conditions can be
 	// useful (see .node.status.conditions), the ability to deconflict is important.
+	// The regex it matches is (dns1123SubdomainFmt/)?(qualifiedNameFmt)
 	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$`
```
deads2k, line 1375: you sure you need the `\\`?  Can you make a proof PR we can play with?

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
```diff
@@ -1369,29 +1369,44 @@ type Condition struct {
 	// ---
 	// Many .condition.type values are consistent across resources like Available, but because arbitrary conditions can be
 	// useful (see .node.status.conditions), the ability to deconflict is important.
+	// The regex it matches is (dns1123SubdomainFmt/)?(qualifiedNameFmt)
 	// +required
+	// +kubebuilder:validation:Required
```
deads2k, line 1374: it's not perfect, but the max length is 253+63=316

deads2k:
also `make update`.  your generated code is out of date

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
damemi: fixed

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
damemi: added

damemi:
@deads2k updated my commit and rebased on yours again

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
```diff
@@ -1369,29 +1369,45 @@ type Condition struct {
...
+	// +kubebuilder:validation:MaxLength=1024
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Pattern=`^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$`
 	Reason string `json:"reason" protobuf:"bytes,5,opt,name=reason"`
 	// message is a human readable message indicating details about the transition.
 	// This may be an empty string.
 	// +required
+	// +kubebuilder:validation:Required
```
deads2k, line 1410: we allow the message to be empty.  See two lines up.  The required bit is about serialization back to clients.

liggitt: allowed to be empty, but the field must be specified or openapi validation will complain (`"message":""`)

deads2k: > This looks correct to me. The value can be an empty string, but the field must be specified or openapi validation will complain (`"message":""`)

Found it in my live test.  The CR validation doesn't give slice numbers, so I was looking at the wrong item.

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto
```diff
@@ -177,11 +187,17 @@ message Condition {
   // The value should be a CamelCase string.
   // This field may not be empty.
   // +required
+  // +kubebuilder:validation:Required
+  // +kubebuilder:validation:MaxLength=1024
+  // +kubebuilder:validation:MinLength=1
+  // +kubebuilder:validation:Pattern=`^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$`
```
smarterclayton, line 193: Allowing commas.  Not a fan....

smarterclayton: Same for colon.  Hrm... gonna have to think about this.

deads2k: > Allowing commas. Not a fan....

Matches the reality of the world I'm afraid.  We have people interested in adopting now that have union reasons with delimiters

smarterclayton: I guess since we said "machine readable" this is the world we live in.

deads2k:
validation is finalized, removing hold.

PR #92667, opened 2020-06-30
Admission webhook warnings

liggitt:
**What type of PR is this?**
/kind feature

part of https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1693-warnings and https://github.com/kubernetes/enhancements/issues/1693

```release-note
Admission webhooks can now return warning messages that are surfaced to API clients, using the `.response.warnings` field in the admission review response.
```

/sig api-machinery
/assign @deads2k @smarterclayton


liggitt:
cc @thockin, since I know you're interested in ongoing use of warnings

staging/src/k8s.io/apiserver/pkg/endpoints/filters/warning.go
```diff
@@ -47,6 +72,11 @@ func (r *recorder) AddWarning(agent, text string) {
 	r.lock.Lock()
 	defer r.lock.Unlock()
 
+	// if we've already exceeded our limit and are already truncating, return early
+	if r.written >= truncateAtTotalRunes && r.truncating {
+		return
```
deads2k, line 77: utilruntime.HandleError() at least once so that we know this is happening server-side?  or a metric perhaps?

deads2k:
that came our very cleanly. One question about how we'll know this is happening, but lgtm otherwise.

/lgtm
/hold in case you want to log that this is happening at some level

staging/src/k8s.io/apiserver/pkg/endpoints/filters/warning.go
liggitt: logged an error once at the point where we flip to truncating

PR #92669, opened 2020-06-30
Separate network proxy flag for apiserver egress and starting pods

Jefftree:
**What type of PR is this?**

/kind feature

Separate network proxy flag for apiserver egress and starting pods. Provides the option to only configure apiserver egress without starting the proxy server and agent pods.

Fixes #

Network proxy [CI job](https://k8s-testgrid.appspot.com/sig-api-machinery-network-proxy) and documentation will need to be updated with the new flag.

Command to start network proxy changes from 

```release-note
Command to start network proxy changes from 'KUBE_ENABLE_EGRESS_VIA_KONNECTIVITY_SERVICE ./cluster/kube-up.sh' to 'KUBE_ENABLE_KONNECTIVITY_SERVICE=true ./hack/kube-up.sh'
```

```docs

```

/assign @cheftako 
/assign @caesarxuchao 

/sig api-machinery

### DIFF
--- a/cluster/gce/config-default.sh
+++ b/cluster/gce/config-default.sh
@@ -518,8 +518,16 @@ WINDOWS_NODE_TAINTS="${WINDOWS_NODE_TAINTS:-node.kubernetes.io/os=win1809:NoSche
 export GCE_PRIVATE_CLUSTER="${KUBE_GCE_PRIVATE_CLUSTER:-false}"
 export GCE_PRIVATE_CLUSTER_PORTS_PER_VM="${KUBE_GCE_PRIVATE_CLUSTER_PORTS_PER_VM:-}"
 
-# Optional: Create apiserver konnectivity server and agent.
-export ENABLE_EGRESS_VIA_KONNECTIVITY_SERVICE="${KUBE_ENABLE_EGRESS_VIA_KONNECTIVITY_SERVICE:-false}"
+# When KUBE_ENABLE_KONNECTIVITY_SERVICE is enabled, the two variables below will
+# will default to true to enable the konnectivity network proxy and start the required pods.
+# Their values can be overridden for more granular control of the proxy.
+
+# Optional: Whether to use konnectivity network proxy for all egress from apiserver.
+export EGRESS_VIA_KONNECTIVITY_SERVICE="${KUBE_ENABLE_KONNECTIVITY_SERVICE:-false}"
+# Optional: Whether to start the konnectivity server and agent pods.
+export RUN_KONNECTIVITY_SERVICE_PODS="${KUBE_ENABLE_KONNECTIVITY_SERVICE:-false}"
+# Proxy Protocol Mode determines the protocol to use to communicate between apiserver and network proxy.
+# Valid options are grpc and http-connect. Default is grpc.
 export KONNECTIVITY_SERVICE_PROXY_PROTOCOL_MODE="${KUBE_KONNECTIVITY_SERVICE_PROXY_PROTOCOL_MODE:-grpc}"
 
 # Optional: Enable Windows CSI-Proxy
--- a/cluster/gce/config-default.sh
+++ b/cluster/gce/config-default.sh
@@ -519,16 +519,16 @@ export GCE_PRIVATE_CLUSTER="${KUBE_GCE_PRIVATE_CLUSTER:-false}"
 export GCE_PRIVATE_CLUSTER_PORTS_PER_VM="${KUBE_GCE_PRIVATE_CLUSTER_PORTS_PER_VM:-}"
 
 # When KUBE_ENABLE_KONNECTIVITY_SERVICE is enabled, the three variables below will
-# will default to true to enable the konnectivity network proxy and start the required pods.
+# default to true to enable the konnectivity network proxy and start the required pods.
 # Their values can be overridden for more granular control of the proxy.
 
 # Optional: Whether to do the setup for the konnectivity service
 # Includes setting up kubeconfig, tokens, egress files, and firewall rules
-export SETUP_KONNECTIVITY_SERVICE="${KUBE_ENABLE_KONNECTIVITY_SERVICE:-false}"
+export PREPARE_KONNECTIVITY_SERVICE="${KUBE_ENABLE_KONNECTIVITY_SERVICE:-false}"
 # Optional: Whether to use konnectivity network proxy for all egress from apiserver.
 export EGRESS_VIA_KONNECTIVITY="${KUBE_ENABLE_KONNECTIVITY_SERVICE:-false}"
 # Optional: Whether to start the konnectivity server and agent pods.
-export RUN_KONNECTIVITY_SERVICE_PODS="${KUBE_ENABLE_KONNECTIVITY_SERVICE:-false}"
+export RUN_KONNECTIVITY_PODS="${KUBE_ENABLE_KONNECTIVITY_SERVICE:-false}"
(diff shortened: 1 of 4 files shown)


caesarxuchao:
/lgtm

@cheftako can you approve? Thanks.

Jefftree:
Added exports to the variables in `cluster/gce/config-default.sh` to pass the CI.

cluster/gce/config-default.sh
```diff
@@ -503,6 +503,14 @@ WINDOWS_NODE_TAINTS="${WINDOWS_NODE_TAINTS:-node.kubernetes.io/os=win1809:NoSche
...
-# Optional: Create apiserver konnectivity server and agent.
-export ENABLE_EGRESS_VIA_KONNECTIVITY_SERVICE="${KUBE_ENABLE_EGRESS_VIA_KONNECTIVITY_SERVICE:-false}"
+# When KUBE_ENABLE_KONNECTIVITY_SERVICE is enabled, the two variables below will
+# will default to true to enable the konnectivity network proxy and start the required pods.
+# Their values can be overridden for more granular control of the proxy.
+
+# Optional: Whether to use konnectivity network proxy for all egress from apiserver.
+export EGRESS_VIA_KONNECTIVITY_SERVICE="${KUBE_ENABLE_KONNECTIVITY_SERVICE:-false}"
+# Optional: Whether to start the konnectivity server and agent pods.
+export RUN_KONNECTIVITY_SERVICE_PODS="${KUBE_ENABLE_KONNECTIVITY_SERVICE:-false}"
```
cheftako, line 513: As this is not just PODS I think it would be better to not name add pods to the name. 'Run konnectivity service' and 'egress via konnectivity' seem like the correct names. It may also make some of the decisions below a bit more clear.

cluster/gce/gci/configure-helper.sh
```diff
@@ -2896,7 +2896,7 @@ function main() {
   if [[ "${ENABLE_APISERVER_INSECURE_PORT:-false}" != "true" ]]; then
     KUBE_BOOTSTRAP_TOKEN="$(secure_random 32)"
   fi
-  if [[ "${ENABLE_EGRESS_VIA_KONNECTIVITY_SERVICE:-false}" == "true" ]]; then
+  if [[ "${EGRESS_VIA_KONNECTIVITY_SERVICE:-false}" == "true" ]]; then
```
cheftako, line 2899: This (the token) is about running the actual konnectivity service not about egress via the service. It determines things like if the kubeconfig for the konnectivity server is setup and we need it setup.

cluster/gce/util.sh
```diff
@@ -2571,7 +2576,7 @@ function create-master() {
     --allow tcp:443 &
 
   echo "Configuring firewall for apiserver konnectivity server"
-  if [[ "${ENABLE_EGRESS_VIA_KONNECTIVITY_SERVICE:-false}" == "true" ]]; then
+  if [[ "${EGRESS_VIA_KONNECTIVITY_SERVICE:-false}" == "true" ]]; then
```
cheftako, line 2579: It seems like this determines if the agents can connect to the server. I believe we want this turned on, otherwise the agents will behave badly (crashloop?)

Jefftree: You mean moving this to `RUN_KONNECTIVITY_SERVICE` instead of keeping it at `ENABLE_EGRESS_VIA_KONNECTIVITY`?

cluster/gce/config-default.sh
Jefftree: ack

cluster/gce/gci/configure-helper.sh
Jefftree: Hmm, we should probably use a third variable (eg: `SETUP_KONNECTIVITY_SERVICE`) to handle this setup phase since it should not be bundled with the apiserver egress flag. 

An alternative is using an enum, but I think it would get a bit messy given that we would have very long enums to describe the combinations of variables currently.

cluster/gce/config-default.sh
Jefftree: I can make the change but why do you mean by this is not just pods?
The goal of this flag is to toggle only the konnectivity server static pod and the konnectivity agent daemonset pods. Is there something missing?

caesarxuchao (commented):
Sorry I'm picking on the names. @cheftako do you have opinions on the names?

Otherwise lgtm.

Jefftree:
Updated, PTAL.

Jefftree:
Tests are probably failing due to needing a rebase. Since we're in code freeze anyway I'll hold off on that for now.

Jefftree:
@cheftako @caesarxuchao Rebased with v1.20. PTAL, Thanks!

cheftako:
Is it ever ok to have EGRESS_VIA_KONNECTIVITY be set to true (i.e. KAS attempts to talk to cluster via ANP) without PREPARE_KONNECTIVITY_SERVICE and RUN_KONNECTIVITY_PODS also being true? It seems like the answer is no (any such configuration would be a broken cluster). If not then we should have guard rails to prevent that misconfiguration.

Jefftree:
> Is it ever ok to have EGRESS_VIA_KONNECTIVITY be set to true (i.e. KAS attempts to talk to cluster via ANP) without PREPARE_KONNECTIVITY_SERVICE and RUN_KONNECTIVITY_PODS also being true? It seems like the answer is no (any such configuration would be a broken cluster). If not then we should have guard rails to prevent that misconfiguration.

kube-up.sh exposes only one variable "KUBE_ENABLE_KONNECTIVITY_SERVICE" which forces all three variables to be the same value so misconfiguration should be unlikely.

cluster/gce/gci/configure-kubeapiserver.sh
```diff
@@ -332,17 +332,19 @@ function start-kube-apiserver {
...
     csc_config_mount="{\"name\": \"cscconfigmount\",\"mountPath\": \"/etc/srv/kubernetes/egress_selector_configuration.yaml\", \"readOnly\": false},"
     csc_config_volume="{\"name\": \"cscconfigmount\",\"hostPath\": {\"path\": \"/etc/srv/kubernetes/egress_selector_configuration.yaml\", \"type\": \"FileOrCreate\"}},"
-    params+=" --egress-selector-config-file=/etc/srv/kubernetes/egress_selector_configuration.yaml"
 
     # UDS socket for communication between apiserver and konnectivity-server
     local default_konnectivity_socket_path="/etc/srv/kubernetes/konnectivity"
     default_konnectivity_socket_vol="{ \"name\": \"konnectivity-socket\", \"hostPath\": {\"path\": \"${default_konnectivity_socket_path}\", \"type\": \"DirectoryOrCreate\"}},"
     default_konnectivity_socket_mnt="{ \"name\": \"konnectivity-socket\", \"mountPath\": \"${default_konnectivity_socket_path}\", \"readOnly\": false},"
   fi
+  if [[ "${EGRESS_VIA_KONNECTIVITY:-false}" == "true" ]]; then
```
cheftako, line 345: A bit wonky moving this out of the PREPARE_KONNECTIVITY_SERVICE switch. If PREPARE_KONNECTIVITY_SERVICE isn't enabled then the file won't exist. OTOH at least it may make the issue obvious in the way it blows up. (I realize your not supposed to enable this without PREPARE_KONNECTIVITY_SERVICE being enabled. However if we were relying on that we wouldn't move this out)

Jefftree:
Rebased. @cheftako can you lgtm again?

Jefftree:
Build failure seems to be related to https://github.com/kubernetes/test-infra/issues/19662

PR #92677, opened 2020-07-01
no need to check nominated pod exist or not as we always delete it first

mlmhl:
**What type of PR is this?**
>

/kind cleanup

When we adding a nominated pod, we always delete it first, see [here](https://github.com/kubernetes/kubernetes/blob/180af4240c42906f072112e7ac7818f4b3574c5a/pkg/scheduler/internal/queue/scheduling_queue.go#L738), so we needn't to check if it exist again.

Fixes #

```release-note
NONE
```

```docs
NONE
```

/sig scheduling

### DIFF
--- a/pkg/scheduler/internal/queue/scheduling_queue.go
+++ b/pkg/scheduler/internal/queue/scheduling_queue.go
@@ -745,12 +745,6 @@ func (npm *nominatedPodMap) add(p *v1.Pod, nodeName string) {
 		}
 	}
 	npm.nominatedPodToNode[p.UID] = nnn
-	for _, np := range npm.nominatedPods[nnn] {
-		if np.UID == p.UID {
-			klog.V(4).Infof("Pod %v/%v already exists in the nominated map!", p.Namespace, p.Name)
-			return
-		}
-	}
 	npm.nominatedPods[nnn] = append(npm.nominatedPods[nnn], p)
 }
 


pkg/scheduler/internal/queue/scheduling_queue.go
```diff
@@ -745,12 +745,6 @@ func (npm *nominatedPodMap) add(p *v1.Pod, nodeName string) {
 		}
 	}
 	npm.nominatedPodToNode[p.UID] = nnn
-	for _, np := range npm.nominatedPods[nnn] {
```
hex108, line 748: The value of `nnn` seems not be same as the value of [nominated node](https://github.com/kubernetes/kubernetes/blob/180af4240c42906f072112e7ac7818f4b3574c5a/pkg/scheduler/internal/queue/scheduling_queue.go#L758)?

mlmhl: If pod exist in `npm.nominatedPods[nnn]`, it should be deleted when `npm.delete(p)` invoked?

chendave: There is a possibility that pod was nominated to a different node other than the `nodeName` passed to `add` function here.

you can see the map is set here, 
```
npm.nominatedPodToNode[p.UID] = nnn
```

while the `nnn` is get from the existing `nominatedPodToNode`
```
nnn, ok := npm.nominatedPodToNode[p.UID]
```

soulxu: > There is a possibility that pod was nominated to a different node other than the `nodeName` passed to `add` function here.

Emm...seems like the npm.delete(p) queries the "nominatedPodToNode" first, "nominatedPodToNode" means there is only one nominated node for the pod. so that should be ok even the pass different nodeName to add function here.

chendave: +1, seems like it is good even they are different.

chendave: The `delete` is deleting the nominated pod of `npm.nominatedPodToNode[p.UID]`,  the check of existence is actually on the node passed in, is there a status one pod is nominated to different nodes?

PR #92681, opened 2020-07-01
Fix a condition when expiring nil VM entry in VMSS cache

### LINKED ISSUE #92680: Azure: Wrong condition when expiring nil VM entry in VMSS cache
**What happened**:
https://github.com/kubernetes/kubernetes/blob/180af4240c42906f072112e7ac7818f4b3574c5a/staging/src/k8s.io/legacy-cloud-providers/azure/azure_vmss_cache.go#L182-L188

From the comment above it should be `if vmEntry.virtualMachine == nil && time.Since(vmEntry.lastUpdate) > 15*time.Minute`

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

/area provider/azure
/sig cloud-provider

ArchangelSDY:
**What type of PR is this?**
/kind bug

Fix a condition when expiring nil VM entry in VMSS cache.

https://github.com/kubernetes/kubernetes/blob/180af4240c42906f072112e7ac7818f4b3574c5a/staging/src/k8s.io/legacy-cloud-providers/azure/azure_vmss_cache.go#L182-L188

From the comment above it should be `&&`.

Fixes #92680

```release-note
Fix a condition when expiring nil VM entry in VMSS cache
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/legacy-cloud-providers/azure/azure_vmss_cache.go
+++ b/staging/src/k8s.io/legacy-cloud-providers/azure/azure_vmss_cache.go
@@ -182,7 +182,7 @@ func (ss *scaleSet) newVMSSVirtualMachinesCache() (*azcache.TimedCache, error) {
 			for name, vmEntry := range oldCache {
 				// if the nil cache entry has existed for 15 minutes in the cache
 				// then it should not be added back to the cache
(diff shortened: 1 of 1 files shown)


feiskyer (commented):
Nice catch! Thanks for the fix.

/area provider/azure
/lgtm
/approve

PR #92682, opened 2020-07-01
Preemption plugin to fetch pod from informer cache

### LINKED ISSUE #92628: Preemption plugin should lookup the pod from informer cache
Currently the preemption plugin gets the latest version of the pod from [the api server directly](https://github.com/kubernetes/kubernetes/blob/88c1386e0cd39e1939a923e44e799bc639ee4275/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go#L103). This is not needed and expensive (costs an api request).  We  should change this to lookup the pod from the informer cache.

/sig scheduling
/priority important-soon
/assign @Huang-Wei

Huang-Wei:
**What type of PR is this?**

/kind feature
/sig scheduling

The old preemption logic calls API server to get a live version of preemptor, which is costy and unnecessary. This PR changes the logic to fetch the pod from podInformer cache.

Fixes #92628

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
+++ b/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
@@ -33,6 +33,7 @@ import (
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	corelisters "k8s.io/client-go/listers/core/v1"
 	policylisters "k8s.io/client-go/listers/policy/v1beta1"
 	extenderv1 "k8s.io/kube-scheduler/extender/v1"
 	podutil "k8s.io/kubernetes/pkg/api/v1/pod"
@@ -52,6 +53,7 @@ const (
 // DefaultPreemption is a PostFilter plugin implements the preemption logic.
 type DefaultPreemption struct {
 	fh        framework.FrameworkHandle
+	podLister corelisters.PodLister
 	pdbLister policylisters.PodDisruptionBudgetLister
 }
 
@@ -66,6 +68,7 @@ func (pl *DefaultPreemption) Name() string {
 func New(_ runtime.Object, fh framework.FrameworkHandle) (framework.Plugin, error) {
 	pl := DefaultPreemption{
 		fh:        fh,
+		podLister: fh.SharedInformerFactory().Core().V1().Pods().Lister(),
 		pdbLister: getPDBLister(fh.SharedInformerFactory()),
 	}
 	return &pl, nil
@@ -109,7 +112,7 @@ func (pl *DefaultPreemption) preempt(ctx context.Context, state *framework.Cycle
 	// It's safe to directly fetch pod here. Because the informer cache has already been
 	// initialized when creating the Scheduler obj, i.e., factory.go#MakeDefaultErrorFunc().
 	// However, tests may need to manually initialize the shared pod informer.
-	pod, err := pl.fh.SharedInformerFactory().Core().V1().Pods().Lister().Pods(pod.Namespace).Get(pod.Name)
+	pod, err := pl.podLister.Pods(pod.Namespace).Get(pod.Name)
 	if err != nil {
 		klog.Errorf("Error getting the updated preemptor pod object: %v", err)
 		return "", err
--- a/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
+++ b/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
@@ -106,8 +106,10 @@ func (pl *DefaultPreemption) preempt(ctx context.Context, state *framework.Cycle
 	nodeLister := pl.fh.SnapshotSharedLister().NodeInfos()
 
 	// 0) Fetch the latest version of <pod>.
-	// TODO(Huang-Wei): get pod from informer cache instead of API server.
-	pod, err := util.GetUpdatedPod(cs, pod)
+	// It's safe to directly fetch pod here. Because the informer cache has already been
+	// initialized when creating the Scheduler obj, i.e., factory.go#MakeDefaultErrorFunc().
+	// However, tests may need to manually initialize the shared pod informer.
+	pod, err := pl.fh.SharedInformerFactory().Core().V1().Pods().Lister().Pods(pod.Namespace).Get(pod.Name)
 	if err != nil {
 		klog.Errorf("Error getting the updated preemptor pod object: %v", err)
 		return "", err
--- a/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
+++ b/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
@@ -218,6 +218,11 @@ func TestPostFilter(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
+			p := DefaultPreemption{
+				fh:        f,
+				podLister: informerFactory.Core().V1().Pods().Lister(),
+				pdbLister: getPDBLister(informerFactory),
+			}
 
 			state := framework.NewCycleState()
 			// Ensure <state> is populated.
@@ -225,10 +230,6 @@ func TestPostFilter(t *testing.T) {
 				t.Errorf("Unexpected PreFilter Status: %v", status)
 			}
 
-			p := DefaultPreemption{
-				fh:        f,
-				pdbLister: getPDBLister(informerFactory),
-			}
 			gotResult, gotStatus := p.PostFilter(context.TODO(), state, tt.pod, tt.filteredNodesStatuses)
 			if !reflect.DeepEqual(gotStatus, tt.wantStatus) {
 				t.Errorf("Status does not match: %v, want: %v", gotStatus, tt.wantStatus)
@@ -1307,6 +1308,7 @@ func TestPreempt(t *testing.T) {
 			// Call preempt and check the expected results.
 			pl := DefaultPreemption{
 				fh:        fwk,
+				podLister: informerFactory.Core().V1().Pods().Lister(),
 				pdbLister: getPDBLister(informerFactory),
 			}
 			node, err := pl.preempt(context.Background(), state, test.pod, make(framework.NodeToStatusMap))
--- a/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
+++ b/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
@@ -85,17 +85,6 @@ var (
 	epochTime6 = metav1.NewTime(time.Unix(0, 6))
 )
 
-func mergeObjs(pod *v1.Pod, pods []*v1.Pod) []runtime.Object {
-	var objs []runtime.Object
-	if pod != nil {
-		objs = append(objs, pod)
-	}
-	for i := range pods {
-		objs = append(objs, pods[i])
-	}
-	return objs
-}
-
 func TestPostFilter(t *testing.T) {
 	onePodRes := map[v1.ResourceName]string{v1.ResourcePods: "1"}
 	tests := []struct {
@@ -110,9 +99,9 @@ func TestPostFilter(t *testing.T) {
 	}{
 		{
 			name: "pod with higher priority can be made schedulable",
-			pod:  st.MakePod().Name("p").UID("p").Priority(highPriority).Obj(),
+			pod:  st.MakePod().Name("p").UID("p").Namespace(v1.NamespaceDefault).Priority(highPriority).Obj(),
 			pods: []*v1.Pod{
-				st.MakePod().Name("p1").UID("p1").Node("node1").Obj(),
+				st.MakePod().Name("p1").UID("p1").Namespace(v1.NamespaceDefault).Node("node1").Obj(),
 			},
 			nodes: []*v1.Node{
 				st.MakeNode().Name("node1").Capacity(onePodRes).Obj(),
@@ -125,9 +114,9 @@ func TestPostFilter(t *testing.T) {
 		},
 		{
 			name: "pod with tied priority is still unschedulable",
-			pod:  st.MakePod().Name("p").UID("p").Obj(),
+			pod:  st.MakePod().Name("p").UID("p").Namespace(v1.NamespaceDefault).Obj(),
 			pods: []*v1.Pod{
-				st.MakePod().Name("p1").UID("p1").Node("node1").Obj(),
+				st.MakePod().Name("p1").UID("p1").Namespace(v1.NamespaceDefault).Node("node1").Obj(),
 			},
 			nodes: []*v1.Node{
 				st.MakeNode().Name("node1").Capacity(onePodRes).Obj(),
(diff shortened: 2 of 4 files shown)


pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -1355,8 +1344,16 @@ func TestPreempt(t *testing.T) {
 	labelKeys := []string{"hostname", "zone", "region"}
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			apiObjs := mergeObjs(test.pod, test.pods)
-			client := clientsetfake.NewSimpleClientset(apiObjs...)
+			client := clientsetfake.NewSimpleClientset()
+			informerFactory := informers.NewSharedInformerFactory(client, 0)
+			podInformer := informerFactory.Core().V1().Pods().Informer()
+			test.pod.Namespace = v1.NamespaceDefault
```
Huang-Wei, line 1350: It's intended to assign a non-nil namespace, as the informer store uses pod name as the key if the namespace is empty:

https://github.com/kubernetes/kubernetes/blob/180af4240c42906f072112e7ac7818f4b3574c5a/staging/src/k8s.io/client-go/tools/cache/store.go#L99-L111

However, podNamespaceLister just concats namespace + "/" + name as the key to fetch Pod:

https://github.com/kubernetes/kubernetes/blob/180af4240c42906f072112e7ac7818f4b3574c5a/staging/src/k8s.io/client-go/listers/core/v1/pod.go#L90-L99

In a real-world case, API has defaulting logic to apply "default" namespace if namespace is nil. But in UT, we should take care of it.

Huang-Wei: It's actually better to apply the namespace directly in the testcases. I will do that either in this PR, or #92571 - depends on which one gets merged first as they conflict with each other.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -98,9 +98,10 @@ func (pl *DefaultPreemption) PostFilter(ctx context.Context, state *framework.Cy
 // using the nominated resources and the nominated pod could take a long time
 // before it is retried after many other pending pods.
 func preempt(ctx context.Context, fh framework.FrameworkHandle, state *framework.CycleState, pod *v1.Pod, m framework.NodeToStatusMap) (string, error) {
-	cs := fh.ClientSet()
-	// TODO(Huang-Wei): get pod from informer cache instead of API server.
-	pod, err := util.GetUpdatedPod(cs, pod)
+	// It's safe to directly fetch pod here. Because the informer cache has already been
+	// initialized when creating the Scheduler obj, i.e., factory.go#MakeDefaultErrorFunc().
+	// However, tests may need to manually initialize the shared pod informer.
+	pod, err := fh.SharedInformerFactory().Core().V1().Pods().Lister().Pods(pod.Namespace).Get(pod.Name)
```
soulxu, line 104: May I ask why we need to get the pod again even from the informer? is it for just update the info of pod in case something changed?

chendave: yeah, How to make sure the cache has the latest version of pod?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
chendave: sound like a bug in client-go.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: The pod passed in is from the schedulingQueue, which may be a dated version, esp. considering that this is a preemption case. However, the API obj in informed cache, IMO in most cases is safe to use, which in theory can be a bit late than the live version, but should be good enough to represent the latest version.

Huang-Wei:
As #92682 is /lgtmed, I will wait for that to be merged first, and then rebase here.

/hold

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
soulxu: got it, thanks

Huang-Wei:
@ahg-g rebased, PTAL.

/hold cancel

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
ahg-g, line 104: To avoid initialization issues in the future, let's stick with the practice of getting the lister in New.

Huang-Wei: It's now already initialized in New, indirectly  - the full call chain is like this: scheduler.go#New() -> factory.go#create() -> factory.go#MakeDefaultErrorFunc().

ahg-g: I guess what I am suggesting is to establish a pattern in all plugins to not access listers directly like this, but rather create a variable in the plugin that we set in New.

Huang-Wei: I suppose you're suggesting:

```go
informerFactory....Lister()
```

in New(), right? Just like what we did for PDBLister.

ahg-g: Here is my suggestion:

```
type DefaultPreemption struct {
	fh framework.FrameworkHandle
        podLister  corelisters.PodLister 
        pdbLister v1beta1.PodDisruptionBudgetLister
}
```

This line changes to:
```
pod, err := pl.podLister.Pods(pod.Namespace).Get(pod.Name)
```

and New:
```
// New initializes a new plugin and returns it.
func New(_ runtime.Object, fh framework.FrameworkHandle) (framework.Plugin, error) {
	var pdbLister v1beta1.PodDisruptionBudgetLister
	if utilfeature.DefaultFeatureGate.Enabled(kubefeatures.PodDisruptionBudget) {
		// A hack to initialize pdbLister in sharedInformerFactory.
		pdbLister = fh.SharedInformerFactory().Policy().V1beta1().PodDisruptionBudgets().Lister()
	}

         pl := DefaultPreemption{
            fh: fh, 
            podLister: SharedInformerFactory().Core().V1().Pods().Lister().Pods(pod.Namespace).Get(pod.Name),
            pdbLister: pdbLister,
         }
	return &pl, nil
}
```
and lastly

```
  func getPodDisruptionBudgets(pdbLister v1beta1.PodDisruptionBudgetLister) ([]*policy.PodDisruptionBudget, error) {
	if pdbLister != nil {
		return pdbLister.List(labels.Everything())
	}
	return nil, nil
}
```

Huang-Wei: I think we discussed it before. My option is that the xyzLister can be fetched from the informerFactory, so we just need to ensure it's initialized properly, and then just keep the informerFactory variable there.

We can move the discussion to #92781.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -97,29 +102,26 @@ func (pl *DefaultPreemption) PostFilter(ctx context.Context, state *framework.Cy
 // other pods with the same priority. The nominated pod prevents other pods from
 // using the nominated resources and the nominated pod could take a long time
 // before it is retried after many other pending pods.
-func preempt(ctx context.Context, fh framework.FrameworkHandle, state *framework.CycleState, pod *v1.Pod, m framework.NodeToStatusMap) (string, error) {
-	// It's safe to directly fetch pod here. Because the informer cache has already been
-	// initialized when creating the Scheduler obj, i.e., factory.go#MakeDefaultErrorFunc().
-	// However, tests may need to manually initialize the shared pod informer.
-	pod, err := fh.SharedInformerFactory().Core().V1().Pods().Lister().Pods(pod.Namespace).Get(pod.Name)
+func (pl *DefaultPreemption) preempt(ctx context.Context, state *framework.CycleState, pod *v1.Pod, m framework.NodeToStatusMap) (string, error) {
```
Huang-Wei, line 105: As preempt() won't be called by plugin developers, making it a method receiver to reduce the number of parameters.

Huang-Wei:
@ahg-g I pushed a commit to explicitly add podLister and pdbLister to struct DefaultPreemption. PTAL.

Huang-Wei:
/hold

`TestPreemptionRaces` failed twice in a row. [1](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92682/pull-kubernetes-integration/1279302413542494209 and [2](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92682/pull-kubernetes-integration/1279302413542494209) both pointed to:

```
E0704 07:04:03.039809  111916 default_preemption.go:108] Error getting the updated preemptor pod object: pod "preemptor-pod" not found
```

I think it's b/c we have 2 PodInformers: one custom one initialized here:

https://github.com/kubernetes/kubernetes/blob/393f9e94e46b17bde4c7285d17bfa44c4d27d513/pkg/scheduler/factory.go#L428

The other is inside sharedInformerFactory. So it's possible the pod received from the custom podInformer gets through schedulingQueue, and all the way to preemption logic, then when it wants to check its latest version in the sharedPodInfomer, it doesn't exist yet.

Huang-Wei:
Possible solutions:

1. still keep 2 podInformers. In the preemption plugin, don't return error if it cannot be fetched from the sharedPodInformer
    - **Pros:** simple and trivial code change
    - **Cons:** pod from the 2 informers can be different, which may cause unexpected bugs.
1. use one podInfomer.
    - a) pass the podInfomer all the way to the initialization of default preemption plugin
        - **Pros:** one podInformer, hence may not see tricky racing bug.
        - **Cons:** code can be ugly, some function signatures need changes, and other in-tree plugins (as well as 3rd party plugins) need to follow the same pattern (use the custom podInformer rather than the sharedPodInformer from sharedInformerFactory)
    - b) modify client-go code to introduce a "informers.WithCustomTweakListOptions" or "informers.WithCustomInformerOptions", so that we can modify the sharedPodInformer in-place within sharedInformerFactory
        - **Pros:** one podInformer, code is clean and changes are transparent to other plugins.
        - **Cons:** needs api-machinery's review

ahg-g:
It would be great if there is a nice way to achieve option 2.b

Huang-Wei:
> It would be great if there is a nice way to achieve option 2.b

Let me chase API experts to see it's achievable in 1.19.

ahg-g:
lets do the pdblister cleanup while waiting for the changes to the pod informer options, the freeze deadline is in a couple of days.

Huang-Wei:
> lets do the pdblister cleanup while waiting for the changes to the pod informer options, the freeze deadline is in a couple of days.

Sure. I will move the related changes out into a new PR.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
soulxu: Should the method 'SharedInformerFactory()' should be removed from 'framework.FrameworkHandle'? Feel like the people who want to write their own plugin are easily making same mistake.

Huang-Wei: > Should the method 'SharedInformerFactory()' should be removed from 'framework.FrameworkHandle'?

Big no... It's the canonical way to interface with API resources.

> Feel like the people who want to write their own plugin are easily making same mistake.

We should document it in our development doc. Ideally in https://github.com/kubernetes-sigs/scheduler-plugins/

soulxu: > > Should the method 'SharedInformerFactory()' should be removed from 'framework.FrameworkHandle'?
> 
> Big no... It's the canonical way to interface with API resources.

Sorry, I'm not clear my thought. Actually I'm thinking of removing the SharedInformerFactory from the framework.Framworkhandle, but add a new parameter to the New() method. As my understanding, we shouldn't allow extension point to use informer factory, Any lister should be getting at the New() method. But yes, even if we just pass the informer factory to the New() method, people still can pass that through the data structure of plugin itself, just like frameworkhandle. In the end, the user still could be using it with wrong way. So probably there needs another more complex solution here. Forgive me if I'm still thinking something wrong, still trying to learn here :)

> 
> > Feel like the people who want to write their own plugin are easily making same mistake.
> 
> We should document it in our development doc. Ideally in https://github.com/kubernetes-sigs/scheduler-plugins/

Yea, that works. just thinking of there should be something in the design to prevent people make mistakes.

pkg/scheduler/scheduler.go
```diff
@@ -309,9 +302,6 @@ func initPolicyFromConfigMap(client clientset.Interface, policyRef *schedulerapi
 
 // Run begins watching and scheduling. It waits for cache to be synced, then starts scheduling and blocked until the context is done.
 func (sched *Scheduler) Run(ctx context.Context) {
-	if !cache.WaitForCacheSync(ctx.Done(), sched.scheduledPodsHasSynced) {
```
ahg-g, line 312: do we need to do something like `informerFactory.WaitForCacheSync(ctx.Done())`

ahg-g (commented):
cool

pkg/scheduler/scheduler.go
```diff
@@ -265,9 +259,8 @@ func New(client clientset.Interface,
 	// Additional tweaks to the config produced by the configurator.
 	sched.StopEverything = stopEverything
 	sched.client = client
-	sched.scheduledPodsHasSynced = podInformer.Informer().HasSynced
```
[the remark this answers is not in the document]
Huang-Wei: As we switched to use informerFactory.HasSynced, it's no needed any more.

We needed it in the first place b/c podInformer is separately created and managed.

BTW: this commit is rebased from #93044. I'm waiting to see if integration tests are impacted.

test/integration/util/util.go
```diff
@@ -371,34 +369,24 @@ func WaitForSchedulerCacheCleanup(sched *scheduler.Scheduler, t *testing.T) {
 func InitTestScheduler(
 	t *testing.T,
 	testCtx *TestContext,
-	setPodInformer bool,
```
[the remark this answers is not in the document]
Huang-Wei: I added a note in "Special notes" section of #93044.

Simply put, all `setPodInformer=true` case does nothing special but creating a podInformer inside, and that podInformer is creating using passed in informerFactory `informerFactory.Core().V1().Pods()`.

test/integration/util/util.go
```diff
@@ -371,34 +369,24 @@ func WaitForSchedulerCacheCleanup(sched *scheduler.Scheduler, t *testing.T) {
...
 func InitTestSchedulerWithOptions(
 	t *testing.T,
 	testCtx *TestContext,
-	setPodInformer bool,
 	policy *schedulerapi.Policy,
 	resyncPeriod time.Duration,
 	opts ...scheduler.Option,
 ) *TestContext {
 	// 1. Create scheduler
 	testCtx.InformerFactory = informers.NewSharedInformerFactory(testCtx.ClientSet, resyncPeriod)
```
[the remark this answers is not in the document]
Huang-Wei: Good catch! cmd/kube-scheduler/app/options/options.go#Config() is not the only place of instantiating a scheduler, I will do a thorough check to init PodInformer and use InfomerFor to update it in-place in informerFactory.

BTW: let's move the discussion to #93044.

pkg/scheduler/scheduler.go
Huang-Wei: The caller already did it via informerFactory.

ahg-g:
@Huang-Wei what is the status of this PR?

Huang-Wei:
> @Huang-Wei what is the status of this PR?

There should be some cleanup work left after rebasing. I will catch up.

PR #92687, opened 2020-07-01
Fixes suggested global-default flow-schema rules

yue9944882:
```release-note
NONE
```

the global-default rules in the suggested configuration are supposed to match all subjects by matching both authenticated and unauthenticated users but the current spec only matches serviceaccount due to oversight. this pull fixes the typo..

note that for now the modification upon mandatory set will be managed/upgrading automatically, however the suggested set doesn't upgrade after creation. i think we should document it somewhere to users how to opt-in the new suggested configurations..

/cc @lavalamp @deads2k @MikeSpreitzer @rngy


MikeSpreitzer (approved):
/LGTM

PR #92692, opened 2020-07-01
fix bug for verify attached volume loop

Pingan2017:
**What type of PR is this?**
/kind bug
/sig storage

for each volume:
https://github.com/kubernetes/kubernetes/blob/a138be8722ebd0ce281029fa747315500a99ffd5/pkg/volume/util/operationexecutor/operation_executor.go#L703-L704
we paased allvolumes each node to func `VerifyVolumesAreAttachedPerNode`
https://github.com/kubernetes/kubernetes/blob/a138be8722ebd0ce281029fa747315500a99ffd5/pkg/volume/util/operationexecutor/operation_executor.go#L758-L760
This will cause each volume to be verified multiple times

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/volume/util/operationexecutor/operation_executor.go
+++ b/pkg/volume/util/operationexecutor/operation_executor.go
@@ -700,6 +700,7 @@ func (oe *operationExecutor) VerifyVolumesAreAttached(
 	volumeSpecMapByPlugin := make(map[string]map[*volume.Spec]v1.UniqueVolumeName)
 
 	for node, nodeAttachedVolumes := range attachedVolumes {
+		needIndividualVerifyVolumes := []AttachedVolume{}
 		for _, volumeAttached := range nodeAttachedVolumes {
 			if volumeAttached.VolumeSpec == nil {
 				klog.Errorf("VerifyVolumesAreAttached: nil spec for volume %s", volumeAttached.VolumeName)
@@ -753,12 +754,12 @@ func (oe *operationExecutor) VerifyVolumesAreAttached(
 				volumeSpecMapByPlugin[pluginName] = volumeSpecMap
 				continue
 			}
-
 			// If node doesn't support Bulk volume polling it is best to poll individually
-			nodeError := oe.VerifyVolumesAreAttachedPerNode(nodeAttachedVolumes, node, actualStateOfWorld)
-			if nodeError != nil {
-				klog.Errorf("VerifyVolumesAreAttached failed for volumes %v, node %q with error %v", nodeAttachedVolumes, node, nodeError)
-			}
+			needIndividualVerifyVolumes = append(needIndividualVerifyVolumes, volumeAttached)
+		}
+		nodeError := oe.VerifyVolumesAreAttachedPerNode(needIndividualVerifyVolumes, node, actualStateOfWorld)
+		if nodeError != nil {
+			klog.Errorf("VerifyVolumesAreAttached failed for volumes %v, node %q with error %v", needIndividualVerifyVolumes, node, nodeError)
 		}
 	}
 


Pingan2017:
ping @msau42 @jingxu97 
   Please take a look, Thx!

pkg/volume/util/operationexecutor/operation_executor.go
```diff
@@ -753,12 +754,12 @@ func (oe *operationExecutor) VerifyVolumesAreAttached(
...
 			}
-
 			// If node doesn't support Bulk volume polling it is best to poll individually
-			nodeError := oe.VerifyVolumesAreAttachedPerNode(nodeAttachedVolumes, node, actualStateOfWorld)
-			if nodeError != nil {
-				klog.Errorf("VerifyVolumesAreAttached failed for volumes %v, node %q with error %v", nodeAttachedVolumes, node, nodeError)
-			}
+			needVerifyVolumes = append(needVerifyVolumes, volumeAttached)
+		}
+		nodeError := oe.VerifyVolumesAreAttachedPerNode(needVerifyVolumes, node, actualStateOfWorld)
```
msau42, line 760: Is it possible to have a unit test for this behavior that could count the number of calls and/or the volumes that were checked?

msau42:
@kubernetes/sig-storage-pr-reviews

pkg/volume/util/operationexecutor/operation_executor.go
Pingan2017: Sorry for delay!
I am afraid can not  count the number  with UT.

Pingan2017:
/retest
ping @kubernetes/sig-storage-pr-reviews 
Could someone help review this pr?

pkg/volume/util/operationexecutor/operation_executor.go
```diff
@@ -753,12 +754,12 @@ func (oe *operationExecutor) VerifyVolumesAreAttached(
 				volumeSpecMapByPlugin[pluginName] = volumeSpecMap
 				continue
 			}
-
 			// If node doesn't support Bulk volume polling it is best to poll individually
-			nodeError := oe.VerifyVolumesAreAttachedPerNode(nodeAttachedVolumes, node, actualStateOfWorld)
-			if nodeError != nil {
-				klog.Errorf("VerifyVolumesAreAttached failed for volumes %v, node %q with error %v", nodeAttachedVolumes, node, nodeError)
-			}
+			needVerifyVolumes = append(needVerifyVolumes, volumeAttached)
```
jingxu97, line 758: @Pingan2017 Good Catch! Thank you for fixing it!

Just a nit, how about needIndividualVerifyVolumes to make the name even more clear?

Pingan2017: @jingxu97  Thanks for your suggestion！
I have fixed it.

jingxu97:
@Pingan2017 for unit test, you might check here to see whether it is possible to add some logic to count. https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/testing/testing.go#L1282

It would be nice to have unit test to avoid future changes to break this fix. But I am ok to merge this one first. Could you create an issue to track adding unit test?

/lgtm

Pingan2017:
@jingxu97  Thx!
/assign @saad-ali

Pingan2017:
> @Pingan2017 for unit test, you might check here to see whether it is possible to add some logic to count. https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/testing/testing.go#L1282
> 
> It would be nice to have unit test to avoid future changes to break this fix. But I am ok to merge this one first. Could you create an issue to track adding unit test?
> 
> /lgtm

I will create an issue to track it.
Thanks!

msau42:
I think it's good practice to always include unit tests as much as possible. From a first glance, I think adding one should be doable using existing tests as an example.

Jing mentioned you'll probably need to implement https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/testing/testing.go#L1282

and there are some examples of populating attachedVolumes like here: https://github.com/kubernetes/kubernetes/blob/a1063c924bfbda6e7d24818858e78fa387c93b72/pkg/volume/util/operationexecutor/operation_executor_test.go#L171

We would be happy to help walk you through it in more detail. Thanks!

Pingan2017:
> I think it's good practice to always include unit tests as much as possible. From a first glance, I think adding one should be doable using existing tests as an example.
> 
> Jing mentioned you'll probably need to implement https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/testing/testing.go#L1282
> 
> and there are some examples of populating attachedVolumes like here:
> 
> https://github.com/kubernetes/kubernetes/blob/a1063c924bfbda6e7d24818858e78fa387c93b72/pkg/volume/util/operationexecutor/operation_executor_test.go#L171
> 
> We would be happy to help walk you through it in more detail. Thanks!

I will try to work with this. It's also a good start for me
Thanks!

jingxu97:
want to make sure this fix is merged before code freeze. Open an issue to track adding unit tests. 

/approve

jingxu97:
I think we need to cherrypick the fix

PR #92696, opened 2020-07-01
skip mismatched webhookAccessor and object

gongguan:
/kind cleanup

Skip building WebhookInvocation when WebhookAccessor and object mismatch, since it's expensive to build WebhookInvocation.

Before
```
BenchmarkShouldCallHookWithComplexSelector-4              942778              1214 ns/op             288 B/op          6 allocs/op
BenchmarkShouldCallHookWithComplexRule-4                   63638             20492 ns/op              64 B/op          1 allocs/op
BenchmarkShouldCallHookWithComplexSelectorAndRule-4        61670             20800 ns/op              65 B/op          1 allocs/op

```
After
```
BenchmarkShouldCallHookWithComplexSelector-4             6707960               218 ns/op              32 B/op          2 allocs/op
BenchmarkShouldCallHookWithComplexRule-4                 5192454               216 ns/op              32 B/op          2 allocs/op
BenchmarkShouldCallHookWithComplexSelectorAndRule-4      5837241               221 ns/op              32 B/op          2 allocs/op
```

```release-note
None
```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/generic/webhook.go
+++ b/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/generic/webhook.go
@@ -141,7 +141,18 @@ func (a *Webhook) ValidateInitialization() error {
 // ShouldCallHook returns invocation details if the webhook should be called, nil if the webhook should not be called,
 // or an error if an error was encountered during evaluation.
 func (a *Webhook) ShouldCallHook(h webhook.WebhookAccessor, attr admission.Attributes, o admission.ObjectInterfaces) (*WebhookInvocation, *apierrors.StatusError) {
-	var err *apierrors.StatusError
+	matches, matchNsErr := a.namespaceMatcher.MatchNamespaceSelector(h, attr)
+	// Should not return an error here for webhooks which do not apply to the request, even if err is an unexpected scenario.
+	if !matches && matchNsErr == nil {
+		return nil, nil
+	}
+
+	// Should not return an error here for webhooks which do not apply to the request, even if err is an unexpected scenario.
+	matches, matchObjErr := a.objectMatcher.MatchObjectSelector(h, attr)
+	if !matches && matchObjErr == nil {
+		return nil, nil
+	}
+
 	var invocation *WebhookInvocation
 	for _, r := range h.GetRules() {
 		m := rules.Matcher{Rule: r, Attr: attr}
@@ -189,15 +200,11 @@ func (a *Webhook) ShouldCallHook(h webhook.WebhookAccessor, attr admission.Attri
 	if invocation == nil {
 		return nil, nil
 	}
-
-	matches, err := a.namespaceMatcher.MatchNamespaceSelector(h, attr)
-	if !matches || err != nil {
-		return nil, err
+	if matchNsErr != nil {
+		return nil, matchNsErr
 	}
-
-	matches, err = a.objectMatcher.MatchObjectSelector(h, attr)
-	if !matches || err != nil {
-		return nil, err
+	if matchObjErr != nil {
+		return nil, matchObjErr
 	}
 
 	return invocation, nil
--- a/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/generic/BUILD
+++ b/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/generic/BUILD
@@ -53,8 +53,11 @@ go_test(
     embed = [":go_default_library"],
     deps = [
         "//staging/src/k8s.io/api/admissionregistration/v1:go_default_library",
+        "//staging/src/k8s.io/api/core/v1:go_default_library",
+        "//staging/src/k8s.io/apimachinery/pkg/api/errors:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/apis/meta/v1:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured:go_default_library",
+        "//staging/src/k8s.io/apimachinery/pkg/labels:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/runtime:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/runtime/schema:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/util/diff:go_default_library",
--- a/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/generic/webhook_test.go
+++ b/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/generic/webhook_test.go
@@ -22,7 +22,10 @@ import (
 	"testing"
 
 	"k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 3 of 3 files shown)


gongguan:
Changed test case because panic will be thrown if NamespaceSelector is nil.
While we can ensure ValidatingWebhook 's NamespaceSelector always not nil, right? Otherwise we can also produce a panic when removing this line https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/generic/webhook_test.go#L114

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/generic/webhook.go
```diff
@@ -142,6 +142,17 @@ func (a *Webhook) ValidateInitialization() error {
 // or an error if an error was encountered during evaluation.
 func (a *Webhook) ShouldCallHook(h webhook.WebhookAccessor, attr admission.Attributes, o admission.ObjectInterfaces) (*WebhookInvocation, *apierrors.StatusError) {
 	var err *apierrors.StatusError
+
+	matches, err := a.namespaceMatcher.MatchNamespaceSelector(h, attr)
+	if !matches || err != nil {
+		return nil, err
+	}
```
liggitt, line 149: This puts a namespace lookup (potentially a live network request) and several error paths ahead of checking if the webhook's rules even apply to this request. I'd like to see benchmark evidence that this is actually cheaper, and consider how to avoid returning an error if the webhook's rules don't even apply to the request.

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/generic/webhook.go
```diff
@@ -142,6 +142,17 @@ func (a *Webhook) ValidateInitialization() error {
...
+
+	matches, err = a.objectMatcher.MatchObjectSelector(h, attr)
+	if !matches || err != nil {
+		return nil, err
```
liggitt, line 153: same comment about not returning an error early here before checking if the webhook even applies to this request

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/generic/webhook.go
gongguan: Updated, result:
BenchmarkShouldCallHook-4         329312              3182 ns/op
BenchmarkShouldCallHook-4         510973              2305 ns/op

gongguan: > how to avoid returning an error if the webhook's rules don't even apply to the request.

I have a question: where is the error happened, when we build invocation WebhookAccessor and Attributes won't change, right?

gongguan: Updated benchmark, make NewValidatingWebhookAccessor outside loop, result is
BenchmarkShouldCallHook-4        1329668               902 ns/op             288 B/op          6 allocs/op
BenchmarkShouldCallHook-4        8379460               140 ns/op              32 B/op          2 allocs/op

@liggitt  Do you think this bench test reasonable?

gongguan: > consider how to avoid returning an error if the webhook's rules don't even apply to the request.

Do you mean MatchNamespaceSelector returns an error while invocation(rules don't even apply to the request) is nil? We want to continue next hook but returned an error. Yeah, I haven't caught this scenario.

This just hide the error(mostly a get request, other occurred error might be logic bug) when no rules applied to request, once rules applied, error would happen, right?

liggitt: > Do you mean MatchNamespaceSelector returns an error while invocation(rules don't even apply to the request) is nil? We want to continue next hook but returned an error. Yeah, I haven't caught this scenario.

Exactly. We should not return an error here for webhooks which do not even apply to the incoming request.

liggitt: > This puts a namespace lookup (potentially a live network request) ... ahead of checking if the webhook's rules even apply to this request.

this is still outstanding as well

gongguan:
Close it because this can't avoid returning an error if the webhook's rules don't even apply to the request.

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/generic/webhook.go
gongguan: Yeah, it can't avoid returning an error if the webhook's rules don't even apply to the request, so I close it.

gongguan: > Exactly. We should not return an error here for webhooks which do not even apply to the incoming request.

@liggitt addressed it. PTAL.

gongguan:
address @liggitt https://github.com/kubernetes/kubernetes/pull/92696#discussion_r451012290 and reopen it

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/generic/webhook.go
gongguan: Sorry, it's not right. it will break https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/generic/webhook.go#L175

gongguan:
pod pending timeout

gongguan:
> Exactly. We should not return an error here for webhooks which do not even apply to the incoming request.

@liggitt I have addressed your comment and reopen this pr, PTAL.

liggitt:
in determining if this is a beneficial change, benchmark-wise, there are a few scenarios we should consider:

* webhook config with complex selector (namespace or object or both) and simple matching rule
* webhook config with simple or absent selector and complex matching rules
* webhook config with complex selector (namespace or object or both) and complex matching rules

Comparing the before/after for all three scenarios will help determine whether there are cases where this change actually decreases performance

gongguan:
> in determining if this is a beneficial change, benchmark-wise, there are a few scenarios we should consider:

Do you mean I should add at least three benchmark tests to cover these scenarios ?

liggitt:
> @liggitt Do you mean I should add at least three benchmark tests to cover these scenarios ?

yes, extend the current benchmark to run those three scenarios

gongguan:
@liggitt I have extended current benchmark tests, the improvement is more obvious:

Before
```
BenchmarkShouldCallHookWithComplexSelector-4              942778              1214 ns/op             288 B/op          6 allocs/op
BenchmarkShouldCallHookWithComplexRule-4                   63638             20492 ns/op              64 B/op          1 allocs/op
BenchmarkShouldCallHookWithComplexSelectorAndRule-4        61670             20800 ns/op              65 B/op          1 allocs/op

```
After
```
BenchmarkShouldCallHookWithComplexSelector-4             6707960               218 ns/op              32 B/op          2 allocs/op
BenchmarkShouldCallHookWithComplexRule-4                 5192454               216 ns/op              32 B/op          2 allocs/op
BenchmarkShouldCallHookWithComplexSelectorAndRule-4      5837241               221 ns/op              32 B/op          2 allocs/op
```

liggitt:
/lgtm
/approve

it makes sense that a complex selector is faster at short-circuiting than a complex list of rules

gongguan:
Thanks for reviewing @liggitt 

/sig scalability
/cc @wojtek-t

wojtek-t:
nice!

PR #92705, opened 2020-07-01
e2e pod logs: use hierarchical directory structure

### LINKED ISSUE #92622: E2E test: support longer test names
Currently test names are formed by concatenating the ginkgo context's together. The testname is used to create a directory. This creates a limit on the length of the the testname, as most operating systems can only support a filename of up to 255 bytes. This concatenating is done by the logging library (I think)

A way to break up the long test names by ginkgo context.
**Why is this needed**:
This test (https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92555/pull-kubernetes-e2e-gce-storage-snapshot/1277695556403073024/) failed on PR #92555 

Current workarounds are simple by shortening the test names. However this is not desirable if we wish to support more nesting of ginkgo contexts in order to reduce boilerplate testing code by sharing common setup. This is also not desirable because it limits the ability to communicate the intent of the .

pohly:
**What type of PR is this?**
/kind failing-test

When using the entire test name as file name, the name became too
long (> 256 characters, which wasn't supported by all file systems)
and the artifact directory got cluttered.

The original reason (a limitation in Gubernator) no longer applies
because Spyglass is used now for log viewing.

Fixes #92622

```release-note
NONE
```

/sig storage
/cc @msau42

### DIFF
--- a/test/e2e/storage/testsuites/base.go
+++ b/test/e2e/storage/testsuites/base.go
@@ -578,14 +578,24 @@ func StartPodLogs(f *framework.Framework, driverNamespace *v1.Namespace) func() 
 		to.LogWriter = ginkgo.GinkgoWriter
 	} else {
 		test := ginkgo.CurrentGinkgoTestDescription()
+		// Clean up each individual component text such that
+		// it contains only characters that are valid as file
+		// name.
 		reg := regexp.MustCompile("[^a-zA-Z0-9_-]+")
+		var components []string
+		for _, component := range test.ComponentTexts {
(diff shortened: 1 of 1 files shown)


pohly:
/hold

We need to check what the resulting artifact directory now looks like.

pohly:
/hold cancel

IMHO the new layout is okay: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/92705/pull-kubernetes-e2e-gce/1278328852736643072/artifacts/

PR #92706, opened 2020-07-01
Create nodes with labels instead of adding them later

nodo:
**What type of PR is this?**
/kind cleanup

Related to https://github.com/kubernetes/kubernetes/pull/92514, this PR removes the last instances of `AddLabelsToNode` and `waitForNodeLabels` in the scheduler integration test.

```release-note
NONE
```


nodo:
/cc @alculquicondor @Huang-Wei 

This should be the last PR for this kind of cleanup.

Huang-Wei:
@nodo Thanks for the efforts. The test looks neat and way more efficient now!

BTW: don't forget to run bazel update to resolve the CI failure.

nodo:
Thanks @Huang-Wei , bazel is happier now :)

PR #91505, opened 2020-05-27
Consistent Resource Version Semantics for List

jpbetz:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

Implements [KEP: Consistent Resource Version Semantics for List](https://github.com/kubernetes/enhancements/pull/1606).
**Which issue(s) this PR fixes**:

Fixes #91503

**Special notes for your reviewer**:

This one is much easier to review if you review each commit individually.

This is a direct implementation of the KEP.  But I'm not 100% happy with the parameter names, so naming suggestions welcome.

proto after_roundtrip diff (from https://github.com/kubernetes/kubernetes/pull/91505#issuecomment-643053155):

```
diff -u \
   <(tail -c +5 v1.17.0/core.v1.ListOptions.pb | protoc --decode_raw) \
   <(tail -c +5 v1.17.0/core.v1.ListOptions.after_roundtrip.pb | protoc --decode_raw)
diff -u \
   <(tail -c +5 v1.18.0/core.v1.ListOptions.pb | protoc --decode_raw) \
   <(tail -c +5 v1.18.0/core.v1.ListOptions.after_roundtrip.pb | protoc --decode_raw)
```

```diff
@@ -11,6 +11,7 @@
   7: 5339971464584210463
   8: "5"
   9: 0
+  10: ""
 }
 3: ""
 4: ""
```

```diff
@@ -11,6 +11,7 @@
   7: 5339971464584210463
   8: "5"
   9: 0
+  10: ""
 }
 3: ""
 4: ""
```

**Does this PR introduce a user-facing change?**:

```release-note
Add `ResourceVersionMatch` paramater to `ListOptions` to provide consistent resource version semantics for list requests regardless of pagination.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:
- [KEP: Consistent Resource Version Semantics for List]( https://github.com/kubernetes/enhancements/tree/5de5c201a360e99faa0d64ce6f4f5b61d7cfc1cd/keps/sig-api-machinery/20200309-consistent-resource-versions-semantics)

/sig api-machinery
/priority important-longterm

PR #92713, opened 2020-07-01
Deflake TestListResourceVersion0

liggitt:
**What type of PR is this?**
/kind flake

Fixes https://storage.googleapis.com/k8s-gubernator/triage/index.html?test=TestListResourceVersion0

```release-note
NONE
```

/cc @jpbetz @wojtek-t 
/sig api-machinery

### DIFF
--- a/test/integration/apiserver/BUILD
+++ b/test/integration/apiserver/BUILD
@@ -54,6 +54,7 @@ go_test(
         "//staging/src/k8s.io/apimachinery/pkg/runtime/serializer/protobuf:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/runtime/serializer/streaming:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/types:go_default_library",
+        "//staging/src/k8s.io/apimachinery/pkg/util/wait:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/watch:go_default_library",
         "//staging/src/k8s.io/apiserver/pkg/features:go_default_library",
         "//staging/src/k8s.io/apiserver/pkg/util/feature:go_default_library",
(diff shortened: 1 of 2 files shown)


wojtek-t:
/lgtm

/hold
@liggitt - feel free to hold cancel, but I'm wondering what if we know why it started flaking couple days ago given it wasn't before?

liggitt:
that's a good question... the PR flake rate looks much more suspect: https://storage.googleapis.com/k8s-gubernator/triage/index.html?date=2020-07-01&pr=1&test=TestListResourceVersion0

liggitt:
first recent flake was observed in https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/91505/pull-kubernetes-integration/1276457193071185921 on https://github.com/kubernetes/kubernetes/pull/91505

wojtek-t:
> that's a good question... the PR flake rate looks much more suspect: https://storage.googleapis.com/k8s-gubernator/triage/index.html?date=2020-07-01&pr=1&test=TestListResourceVersion0

to clarify: this is absolutely a valid fix; I'm just wondering why we weren't observing those flakes before.
My guess as things that in theory might have affected it are:
https://github.com/kubernetes/kubernetes/pull/91505 [as something that was touching the semantic]
https://github.com/kubernetes/kubernetes/pull/92503 [as something that may make processing inside watchcache a bit slower]

I think that the later may cause those infrequent flakes between 6/26 and 6/29. But I don't have any hypothesis what happened around 6/29 afternoon...

liggitt:
yeah, https://github.com/kubernetes/kubernetes/pull/91505 merged on 6/27

liggitt:
I looked through commits around 6/29 and don't see anything relevant. I'm going to go with the theory that our often-seen resource-constrained CI nodes got hit with CPU use (which we've seen many times) and made this worse starting on the 29th.

/hold cancel

PR #85439, opened 2019-11-19
introduce validation for labels‘ value

zouyee:
Signed-off-by: Zou Nengren <zouyee1989@gmail.com>

**What type of PR is this?**
 /kind bug

**What this PR does / why we need it**:

Validate the key is invalid due to its length, or sequence of characters. 

**Which issue(s) this PR fixes**:

Fixes #85428
xref https://bugzilla.redhat.com/show_bug.cgi?id=1760807

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #64907, opened 2018-06-08
WIP - introduce ratcheting validation mechanism

liggitt:
**What this PR does / why we need it**:

Introduces a ratcheting validation mechanism, as described in https://github.com/kubernetes/kubernetes/issues/64841#issuecomment-395141509

Tightening validation on existing data cannot be done in a way that prevents existing stored objects from being updated.

This PR introduces the following mechanism:
* identify granular additional validations (usually to fix fatal errors that always result in persisted, yet broken objects)
* apply those validations on create to prevent new broken objects from being created
* apply those validations on update **only if the existing object also passes the validation**
  * this prevents valid objects from becoming invalid
  * this avoids breaking the ability to update existing invalid objects

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #64841 
Fixes #57510 - flex driver name
Fixes #58477 - duplicate envvar names
Fixes #54567, #64011 - invalid memory quantities (xref #63426)
xref #52936 - storage medium
xref #60934 - duplicate pvc claimName

**Special notes for your reviewer**:
/assign @lavalamp @deads2k 

**Release note**:
```release-note
Validation has tightened when creating new pods (or new objects containing pod templates):
* flex volume driver names must now be valid
* emptyDir medium must now be valid
* duplicate envvar names are not allowed
* duplicate volumes referencing the same PVC claimName are not allowed
* memory limits and requests must be an integer value
```

Issue #59984, opened 2018-02-16
Validation on update and delete


enj:
As it stands today, changes to validation logic can break the API if they tighten the requirements for an object.  The general case for this is:

1. Create object `X`
2. Upgrade to new server version
3. Try to update object `X` (add a label for example)
4. The update fails because of validation errors

Specifically, an object already stored within the API is considered invalid.

Another similar issue can occur on a delete which requires an update (i.e. a finalizer needs to be removed).  Since the object cannot be updated in-place, it cannot be deleted without first "fixing" it to match the current validation requirements.

This has the potential to break controllers or any autonomous process.  It makes changes to validation hard to review because you have to understand how preexisting objects will behave under the new validation.  In openshift, this constantly breaks the storage migration process which does no-op updates to all objects in the API.

A possible fix for the no-op update case would be to _not_ run validation when the input and output objects are exactly the same.

For the more complex update cases and the delete case, we could do something similar to `IsOnlyMutatingGCFields` to determine if validation can be safely skipped.

@kubernetes/sig-api-machinery-misc

smarterclayton:
I think that makes sense.  However, is the no-op case really special for update?  Any other update (including controllers) would be broken. 

`IsOnlyMutatingGCFields` is a bit of a hack - is it the exception rather than the rule?

sttts:
> Specifically, an object already stored within the API is considered invalid.

Tightened validation and existing invalid objects is a serious problem in every cluster. The only question is whether we keep going during migration and graceful/with-initializer deletion, or we fail early and hard.

> is the no-op case really special for update? Any other update (including controllers) would be broken.

I tend to prefer "fail early" and answer that it is not special. Instead, we should delete the object from the system ASAP, possibly with a good migration UI that warns or asks for confirmation. But as controllers are broken otherwise, allowing mirgration just postpones the error to later and it will be much harder to debug and understand why objects stops behaving as expected.

enj:
> we fail early and hard

Considering this stuff fails upgrades for our clusters regularly as it is, failing hard is unacceptable IMO.

> we should delete the object from the system ASAP, possibly with a good migration UI that warns or asks for confirmation

That is not how I see the problem at all - deleting data is not a solution.

I consider an object stored in the API to be, by definition, valid.  Thus validation does not check "is the object valid?", it checks if "the change to the object is valid."

`any object already stored in the API` + `any valid change to that object` == `no validation error` (you only validate the entire object on create)

[...]

PR #92714, opened 2020-07-01
Add label selector value validation

damemi:
**What type of PR is this?**
/kind bug

Deployments (and pods) are currently able to be created even if they set labelselectors with invalid values. This causes undefined behavior that is inconsistent across versions and usually presents as a "scheduler" bug (when the scheduler affinity plugins fail to parse the invalid selector and return an error)

For example, currently if you try to create a pod with a preferred antiaffinity and an invalid selector, the scheduler assumes an invalid selector will never be created and [throws away the error](https://github.com/kubernetes/kubernetes/blob/ff1d6b3746d8586090d6909079005aecb3019f03/pkg/scheduler/framework/v1alpha1/types.go#L129-L132), essentially scheduling the pod without any affinity.

In 1.18, the scheduler fails to schedule the invalid pod (but really, this is only by chance). During `PreScore` affinity calculations for the pod it calls `LabelSelectorAsSelector` and throws out the error, failing to store a PreScore cycle state. The scheduling cycle then actually continues, and only fails because the `Score` plugin does not find the PreScore state.

From what I can tell, this issue goes back at least as far as 1.11, which has the weirdest behavior. In that version, the scheduler successfully schedules the pods and then every *new* pod after that fails with the error that it has an invalid label selector (even if the new pods have no selector). This is because the scheduler used to parse existing pod's affinity labels, but not incoming pod's labels.

Of course we probably don't need to backport this fix all the way to 1.11, but it shows that this has been a longstanding issue. Ultimately there is a standard regex for label values and pods should not be created if they violate this. It is only by luck that the errors are currently "caught" by the scheduler, and as seen here refactors of the scheduling plugins may eliminate any handling of these errors (this is currently the case).

This is the output of this change:
```
$ cat ~/hello-pod.yaml 
apiVersion: apps/v1
kind: Deployment
metadata:
  name: hello-pod
  labels:
    app: hello-pod
spec:
  replicas: 2
  selector:
    matchLabels:
      app: hello-pod
  template:
    metadata:
      labels:
        app: hello-pod
      name: hello-pod
    spec:
      affinity:
        podAntiAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
            - podAffinityTerm:
                labelSelector:
                  matchExpressions:
                    - key: app.kubernetes.io/instance
                      operator: In
                      values:
                        - '{{ .Release.Name }}'
                topologyKey: kubernetes.io/hostname
              weight: 100
      containers:
        - image: "docker.io/ocpqe/hello-pod"
          name: hello-pod
$ kubectl create -f ~/hello-pod.yaml 
The Deployment "hello-pod" is invalid: spec.template.spec.affinity.podAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution[0].podAffinityTerm.matchExpressions.matchExpressions[0]: Invalid value: "{{ .Release.Name }}": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

Fixes #

```release-note
Label selector values are now validated at creation time
```

```docs

```

/sig scheduling

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
+++ b/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
@@ -45,6 +45,12 @@ func ValidateLabelSelectorRequirement(sr metav1.LabelSelectorRequirement, fldPat
 	case metav1.LabelSelectorOpIn, metav1.LabelSelectorOpNotIn:
 		if len(sr.Values) == 0 {
 			allErrs = append(allErrs, field.Required(fldPath.Child("values"), "must be specified when `operator` is 'In' or 'NotIn'"))
+		} else {
+			for _, v := range sr.Values {
+				for _, msg := range validation.IsValidLabelValue(v) {
+					allErrs = append(allErrs, field.Invalid(fldPath, v, msg))
+				}
+			}
 		}
 	case metav1.LabelSelectorOpExists, metav1.LabelSelectorOpDoesNotExist:
 		if len(sr.Values) > 0 {
--- a/pkg/apis/apps/validation/validation_test.go
+++ b/pkg/apis/apps/validation/validation_test.go
@@ -2141,6 +2141,26 @@ func TestValidateDeployment(t *testing.T) {
 	}}
 	errorCases["ephemeral containers not allowed"] = invalidEphemeralContainersDeployment
 
+	invalidAntiAffinity := validDeployment()
+	invalidAntiAffinity.Spec.Template.Spec.Affinity = &api.Affinity{
+		PodAntiAffinity: &api.PodAntiAffinity{
+			PreferredDuringSchedulingIgnoredDuringExecution: []api.WeightedPodAffinityTerm{
+				{
+					Weight: 100,
+					PodAffinityTerm: api.PodAffinityTerm{
+						TopologyKey: "kubernetes.io/hostname",
+						LabelSelector: &metav1.LabelSelector{
+							MatchExpressions: []metav1.LabelSelectorRequirement{
+								{Key: "app.kubernetes.io/instance", Operator: metav1.LabelSelectorOpIn, Values: []string{"{{ .Release.Name }}"}},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	errorCases["a valid label must be an empty string or consist of alphanumeric characters"] = invalidAntiAffinity
+
 	for k, v := range errorCases {
 		errs := ValidateDeployment(v)
 		if len(errs) == 0 {


damemi:
/assign @liggitt 
Could you please check that this change makes sense? It's surprising to me that it has been missing for so long so I'm suspicious

damemi:
@ahg-g @alculquicondor cc for scheduling, maybe you can confirm whether we should be checking this in our plugins or the API server. If it's on us, we have pretty inconsistent behavior

alculquicondor:
+1 to fixing this in kube-apiserver side. However, deferring to API-reviewers to judge if this breaking change is reasonable.

If this is not, in the scheduler side, the only thing we can do is mark the pods as unschedulable, which is also a breaking change (pods that used to get scheduled, don't get scheduled).

Is this the only instance of label selectors not being validated? What about the `.spec.selector` of the Deployment.

damemi:
> +1 to fixing this in kube-apiserver side. However, deferring to API-reviewers to judge if this breaking change is reasonable.
> 
> If this is not, in the scheduler side, the only thing we can do is mark the pods as unschedulable, which is also a breaking change (pods that used to get scheduled, don't get scheduled).
> 
> Is this the only instance of label selectors not being validated? What about the `.spec.selector` of the Deployment.

@alculquicondor `ValidateLabelSelector` is used in [a lot of places](https://github.com/kubernetes/kubernetes/search?q=validatelabelselector&unscoped_q=validatelabelselector), but yes it is also called for `.spec.selector` of Deployments (https://github.com/kubernetes/kubernetes/blob/c369cf187ea765c0a2387f2b39abe6ed18c8e6a8/pkg/apis/apps/validation/validation.go#L477)

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
```diff
@@ -45,6 +45,12 @@ func ValidateLabelSelectorRequirement(sr metav1.LabelSelectorRequirement, fldPat
 	case metav1.LabelSelectorOpIn, metav1.LabelSelectorOpNotIn:
 		if len(sr.Values) == 0 {
 			allErrs = append(allErrs, field.Required(fldPath.Child("values"), "must be specified when `operator` is 'In' or 'NotIn'"))
+		} else {
+			for _, v := range sr.Values {
+				for _, msg := range validation.IsValidLabelValue(v) {
+					allErrs = append(allErrs, field.Invalid(fldPath, v, msg))
+				}
```
liggitt, line 52: /hold

standard questions when we want to tighten validation:
* does it prevent a no-op update of existing persisted objects that would fail this new validation? (would block storage migration tasks)
* does it prevent deletion of existing persisted objects with finalizers that would fail this new validation? (update is required to remove finalizers to allow deletion)
* does it prevent controllers from updating status of existing persisted objects that would fail this new validation?

we could consider ratcheting validation:
* apply this tightened validation on create
* apply this tightened validation on update where the old object passed this validation (don't let a valid object become invalid)
* skip this tightened validation on update where the old object fails this validation

c.f. https://github.com/kubernetes/kubernetes/issues/59984

ahg-g:
+1 to performing the check in the API server. I think Jordan's suggestion of doing that on create only is the safest path forward.

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
lavalamp: Yeah +1 to these points.

damemi:
Yeah, this would affect existing objects (esp. showing that creating and scheduling these objects has been possible through different versions) so I think starting with just create validation is fine

@liggitt How do you change validation just for create? This affects any types that can create a pod (eg deployment, replicaset) so maybe just adding the validation to Pod creation will cover all cases, though I'm not sure.

(also for reference it looks like there was another PR to attempt to fix this that didn't get any traction https://github.com/kubernetes/kubernetes/pull/85439. I'm hoping we can fix it because this can cause cluster-breaking problems in some cases)

damemi:
Based on the number of resources that call `ValidateLabelSelector`[1], it seems like in order to effectively implement this tighter validation only on create would require something like https://github.com/kubernetes/kubernetes/pull/64907. On top of that though, these ratcheting checks would need to be wired *through* many of the resources' Validate functions since they don't all directly call `ValidateLabelSelector` (or we move just this check up to the resources' main Validate functions, which seems weird because it should live in `ValidateLabelSelectorRequirement` and not, for example, `ValidateDeployment` imo). And we need to either exclude this check from Updates or ignore it if the existing object doesn't pass.

Was there any more motive to add a change like in that PR? I see it's closed now, but there should be a way to tighten validation when bugs are found in it like this one, without breaking migration.

[1] podpreset, clusterrole, volumes (pods), networkpolicy, csistoragecapacity, job, mutatingwebhookconfiguration, daemonset, deployment, statefulset, replicaset, and pdb from a quick look

ahg-g:
This is probably a more significant problem for pods than any other resource, pods with invalid selector could get scheduled and so preventing the scheduling of new pods. If we can't get this fixed through api-server validation, we should validate in the scheduler and patch back (perhaps we should do that anyways).

damemi:
> This is probably a more significant problem for pods than any other resource, pods with invalid anti-affinity selector could get scheduled and so preventing the scheduling of new pods. If we can't get this fixed through api-server validation, we should validate in the scheduler and patch back (perhaps we should do that anyways).

Yeah, this mostly affects pods so we can probably stop most of the problems by just blocking pod creation. If that's sufficient and some minor inconsistency is ok we can go that route. I'm fine with that

I don't think we should try to fix it in the scheduler. Like I said in my original comment, how we implement scheduler plugins can change which could cause inconsistency in this, potentially just re-introducing it in a later version. API validation should remain relatively consistent and it's where resource validating should be done.

ahg-g:
@liggitt can we get the validation done only for pods on creation for now? we also want to patch it back to 1.18 and 1.17 if we can.

liggitt:
> @liggitt can we get the validation done only for pods on creation for now?

How do controllers behave when their pod creation attempts are rejected? Do they back off? Do they surface error status?

> we also want to patch it back to 1.18 and 1.17 if we can.

I would not anticipate tightening validating in patch releases.

ahg-g:
> > @liggitt can we get the validation done only for pods on creation for now?
> 
> How do controllers behave when their pod creation attempts are rejected? Do they back off? Do they surface error status?

I am not sure, we should certainly check that, but we already fail creation of the pod for other reasons, do you expect this to be different?

> 
> > we also want to patch it back to 1.18 and 1.17 if we can.
> 
> I would not anticipate tightening validating in patch releases.

damemi:
> We should do this in the scheduler then, at least for anti-affinity because of its significant impact.

I'm ok with fixing it in the scheduler for backports (old versions are unlikely to change much 🙂)

ahg-g:
> > We should do this in the scheduler then, at least for anti-affinity because of its significant impact.
> 
> I'm ok with fixing it in the scheduler for backports (old versions are unlikely to change much 🙂)

do you want to do that? we can do it in PreFilter for the incoming pod in 1.19 and 1.18.

damemi:
Sure, and from 1.20 on hopefully we can resolve this in the api

damemi:
@ahg-g I'm going to open a separate pr for the scheduler backports (to keep this one for reference) but we have some options for how to add it.

Technically, we already check if the labelSelector is valid when we parse the affinity terms in [`NewPodInfo`](https://github.com/kubernetes/kubernetes/blob/ff1d6b3746d8586090d6909079005aecb3019f03/pkg/scheduler/framework/v1alpha1/types.go#L155-L158) (see [`getAffinityTerms`](https://github.com/kubernetes/kubernetes/blob/ff1d6b3746d8586090d6909079005aecb3019f03/pkg/scheduler/framework/v1alpha1/types.go#L110-L114) which calls [`newAffinityTerm`](https://github.com/kubernetes/kubernetes/blob/ff1d6b3746d8586090d6909079005aecb3019f03/pkg/scheduler/framework/v1alpha1/types.go#L91-L99)). If `newAffinityTerm` hits an error parsing the LabelSelector, we just ignore it and continue scheduling the pod.

So our options to fix this are:
1) Actually bubble up the error from `newAffinityTerm`. This seems like the right place to do it but it adds a lot of error handling (anywhere that calls `NewPodInfo`, which includes [`(*NodeInfo).AddPod`](https://github.com/kubernetes/kubernetes/blob/ff1d6b3746d8586090d6909079005aecb3019f03/pkg/scheduler/framework/v1alpha1/types.go#L457) which bubbles up to `NewNodeInfo`, and continues).
2) Add the check manually in `PreFilter`/`PreScore` for the podaffinity plugins. This is redundant, but cleaner and has less risk of side effects in unrelated areas

I'm leaning toward option 2 because I don't think this needs to be validated everywhere, only in the plugins that are using it. Wanted your opinion though because option 1 seems like the "right" way. wdyt?

damemi:
I went ahead and opened a new PR using option 2 from https://github.com/kubernetes/kubernetes/pull/92714#issuecomment-667227790. It's the simplest approach and blocks the pod as early as possible in the plugin. Please take a look

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
alculquicondor: We ended up marking these pods as unschedulable back in 1.19 #93660

I wonder if we are in a better position to improve the validation in apiserver now. We could certainly ratchet the validation. 

Back in the day the question was whether it was acceptable to break the creation of ReplicaSets, but since these pods are unschedulable, I doubt that this would break any production services.

alculquicondor:
Had a discussion with API folks and it seems safe to proceed with adding this validation (with ratcheting), given https://github.com/kubernetes/kubernetes/pull/92714#discussion_r993739386

alculquicondor:
Actually, I thought this was just affecting the PodSpec, but it affects multiple other places. #99139 should cover it.

PR #92717, opened 2020-07-01
add helpers for managing conditions

deads2k:
adds basic find/set/remove/status for conditions

/priority important-soon
@kubernetes/sig-api-machinery-pr-reviews 

```release-note
NONE
```

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/api/meta/conditions.go
+++ b/staging/src/k8s.io/apimachinery/pkg/api/meta/conditions.go
@@ -1,0 +1,101 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package meta
+
+import (
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SetStatusCondition sets the corresponding condition in conditions to newCondition.
+// conditions must be non-nil.
+// 1. if the condition of the specified type already exists (all fields of the existing condition are updated to
+//    newCondition, LastTransitionTime is set to now if the new status differs from the old status)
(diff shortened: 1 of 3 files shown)


staging/src/k8s.io/apimachinery/pkg/api/meta/conditions.go
```diff
@@ -0,0 +1,80 @@
...
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SetStatusCondition sets the corresponding condition in conditions to newCondition.
+func SetStatusCondition(conditions *[]metav1.Condition, newCondition metav1.Condition) {
+	if conditions == nil {
+		conditions = &[]metav1.Condition{}
+	}
```
liggitt, line 13: if nil is passed in, isn't the effect of this function lost?

staging/src/k8s.io/apimachinery/pkg/api/meta/conditions.go
```diff
@@ -0,0 +1,80 @@
+package meta
+
+import (
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SetStatusCondition sets the corresponding condition in conditions to newCondition.
```
liggitt, line 9: document that `conditions` must be non-nil, and document what happens in the following cases:
1. if the condition of the specified type already exists (all fields of the existing condition are updated to `newCondition`, LastTransitionTime is set to now if the new status differs from the old status)
2. if a condition of the specified type does not exist (LastTransitionTime is set to now() if unset, and newCondition is appended)

staging/src/k8s.io/apimachinery/pkg/api/meta/conditions.go
```diff
@@ -0,0 +1,80 @@
...
+
+	existingCondition.Reason = newCondition.Reason
+	existingCondition.Message = newCondition.Message
+}
+
+// RemoveStatusCondition removes the corresponding conditionType from conditions.
+func RemoveStatusCondition(conditions *[]metav1.Condition, conditionType string) {
+	if conditions == nil {
+		conditions = &[]metav1.Condition{}
+	}
```
liggitt, line 40: same question, if nil, can't we just return early?

staging/src/k8s.io/apimachinery/pkg/api/meta/conditions.go
```diff
@@ -0,0 +1,80 @@
...
+	newConditions := []metav1.Condition{}
```
liggitt, line 41:
```suggestion
	newConditions := make([]metav1.Condition, 0, len(conditions)-1)
```

liggitt:
a couple nits and doc requests, mostly lgtm

liggitt:
/approve 

nit on boilerplate header, lgtm otherwise

deads2k:
verify fixed.  Tagging per comment.

PR #92718, opened 2020-07-01
[kube-up]: Bump CoreDNS to v1.7.0

rajansandeep:
**What type of PR is this?**
/kind bug

Updates CoreDNS to v1.7.0. The default Corefile now includes the max_concurrent option, which sets the maximum number of concurrent upstream queries allowed by the forward plugin.

Also updates the migration library to 1.0.10 which supports coredns v1.7.0

Fixes #

```release-note
Kube-up now includes CoreDNS version v1.7.0. Some of the major changes include:
-  Fixed a bug that could cause CoreDNS to stop updating service records.
-  Fixed a bug in the forward plugin where only the first upstream server is always selected no matter which policy is set.
-  Remove already deprecated options `resyncperiod` and `upstream` in the Kubernetes plugin.
-  Includes Prometheus metrics name changes (to bring them in line with standard Prometheus metrics naming convention). They will be backward incompatible with existing reporting formulas that use the old metrics' names.
-  The federation plugin (allows for v1 Kubernetes federation) has been removed.
More details are available in https://coredns.io/2020/06/15/coredns-1.7.0-release/
```

```docs

```

### DIFF
--- a/cluster/addons/dns/coredns/coredns.yaml.base
+++ b/cluster/addons/dns/coredns/coredns.yaml.base
@@ -73,7 +73,9 @@ data:
             ttl 30
         }
         prometheus :9153
-        forward . /etc/resolv.conf
+        forward . /etc/resolv.conf {
+            max_concurrent 1000
+        }
         cache 30
         loop
         reload
@@ -118,7 +120,7 @@ spec:
         kubernetes.io/os: linux
       containers:
       - name: coredns
-        image: k8s.gcr.io/coredns:1.6.7
+        image: k8s.gcr.io/coredns:1.7.0
         imagePullPolicy: IfNotPresent
         resources:
           limits:
--- a/build/dependencies.yaml
+++ b/build/dependencies.yaml
@@ -12,7 +12,7 @@ dependencies:
       match: cniVersion[\t\n\f\r ]*=
 
   - name: "coredns-kube-up"
-    version: 1.6.7
+    version: 1.7.0
     refPaths:
     - path: cluster/addons/dns/coredns/coredns.yaml.base
       match: k8s.gcr.io/coredns
--- a/cluster/addons/dns/coredns/coredns.yaml.in
+++ b/cluster/addons/dns/coredns/coredns.yaml.in
@@ -73,7 +73,9 @@ data:
             ttl 30
         }
         prometheus :9153
-        forward . /etc/resolv.conf
+        forward . /etc/resolv.conf {
+            max_concurrent 1000
+        }
         cache 30
         loop
         reload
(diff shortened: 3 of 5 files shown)


MrHohn (approved):
/lgtm
/approve
/cc @prameshj

cluster/addons/dns/coredns/coredns.yaml.base
```diff
@@ -73,7 +73,9 @@ data:
             ttl 30
         }
         prometheus :9153
-        forward . /etc/resolv.conf
+        forward . /etc/resolv.conf {
+            max_concurrent 1000
```
bowei, line 77: @prameshj 

We've had people run into this # of concurrents as an issue for DNS masq.
Is there a link or reference (embed as a comment) as to how this value was determined?

bowei (commented):
one comment

cluster/addons/dns/coredns/coredns.yaml.base
chrisohaver: Here was my reasoning: https://github.com/coredns/deployment/pull/233#issue-438151578

prameshj: My understanding was also that since this only applies to forwarded queries, cached responses and "cluster.local" queries would continue to be handled. Thanks for the explanation above. Good to know that the client will get a SERVFAIL rather than timeout after no response.

I am curious, how the memory overhead of 2 MB was measured?Any test that queried 1000 unique external names in parallel? From https://golang.org/doc/go1.2#stack_size, if a single goroutine took up 4KB, this could use upto 4MB?

chrisohaver: > My understanding was also that since this only applies to forwarded queries, cached responses and "cluster.local" queries would continue to be handled.

Yes, that's correct.

> From https://golang.org/doc/go1.2#stack_size, if a single goroutine took up 4KB, this could use upto 4MB?

I was basing it on 2k per go-routine https://golang.org/doc/go1.4#runtime.  Nonetheless, 2MB is admittedly an arbitrary number.  The default memory limit per CoreDNS Pod is 170Mi, so there is _a lot_ of headroom before an OOM happens.

If we are not comfortable picking number at this time, we can just leave the it out.  When people see OOMs due to query spikes, they can add this manually.

prameshj: Based on past issues that have been reported, most of the parallel queries are related to searchpath expansion and will have the "cluster.local" suffix. This will be handled by the  "kubernetes" plugin, and hence unaffected by the 1000 connection limit. It seems reasonable to add this new limit to prevent OOMs of coreDNS pods. @bowei what do you think?

bowei: sgtm

/lgtm
/approve

rajansandeep:
Can the milestone tags be added to this PR?
/priority important-soon

nikhita:
> Can the milestone tags be added to this PR?

/assign @onlydole 
release team lead

PR #92720, opened 2020-07-01
kubeadm: add --port=0 for kube-controller-manager and kube-scheduler

### LINKED ISSUE #2202: kubeadm serves kube-scheduler and kube-controller metrics insecurely
kubeadm serves kube-scheduler and kube-controller manager metrics insecurely outside of localhost, as reported here:
https://kubernetes.slack.com/archives/C2P1JHS2E/p1593237397449300

i need to double check this myself, but it feels like our --bind-address=127.0.0.1 is not sufficient to disable that.

for example:
`curl http://public-ip:10252/metrics`

flag refs:
https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/
https://kubernetes.io/docs/reference/command-line-tools-reference/kube-scheduler/

neolit123:

Kubeadm setup of kube-controller-manager and kube-scheduler is
lacking the --port=0 option which caused the component to enable
the insecure port by default and serve insecurely on the default
node interface.

Add --port=0 by default to both components. Users are still allowed
the explicitly set the flag (via extraArgs), which allows them
to override this default kubeadm behavior and enable the insecure port.

NOTE: the flag is deprecated and should be removed from kubeadm manifests
once it's removed from core.

Fixes kubernetes/kubeadm#2202
xref https://github.com/kubernetes/kubeadm/issues/2207

NONE

```release-note
kubeadm: add the deprecated flag --port=0 to kube-controller-manager and kube-scheduler manifests to disable insecure serving. Without this flag the components by default serve (e.g. /metrics) insecurely on the default node interface (controlled by --address). Users that wish to override this behavior and enable insecure serving can pass a custom --port=X via kubeadm's "extraArgs" mechanic for these components.
```

```docs

```

### DIFF
--- a/cmd/kubeadm/app/phases/controlplane/manifests.go
+++ b/cmd/kubeadm/app/phases/controlplane/manifests.go
@@ -322,6 +322,7 @@ func getControllerManagerCommand(cfg *kubeadmapi.ClusterConfiguration) []string 
 	caFile := filepath.Join(cfg.CertificatesDir, kubeadmconstants.CACertName)
 
 	defaultArguments := map[string]string{
+		"port":                             "0",
 		"bind-address":                     "127.0.0.1",
 		"leader-elect":                     "true",
 		"kubeconfig":                       kubeconfigFile,
@@ -392,6 +393,7 @@ func getControllerManagerCommand(cfg *kubeadmapi.ClusterConfiguration) []string 
 func getSchedulerCommand(cfg *kubeadmapi.ClusterConfiguration) []string {
 	kubeconfigFile := filepath.Join(kubeadmconstants.KubernetesDir, kubeadmconstants.SchedulerKubeConfigFileName)
 	defaultArguments := map[string]string{
+		"port":                      "0",
 		"bind-address":              "127.0.0.1",
 		"leader-elect":              "true",
 		"kubeconfig":                kubeconfigFile,
--- a/cmd/kubeadm/app/phases/controlplane/manifests_test.go
+++ b/cmd/kubeadm/app/phases/controlplane/manifests_test.go
@@ -588,6 +588,7 @@ func TestGetControllerManagerCommand(t *testing.T) {
 			},
 			expected: []string{
 				"kube-controller-manager",
+				"--port=0",
 				"--bind-address=127.0.0.1",
 				"--leader-elect=true",
 				"--kubeconfig=" + kubeadmconstants.KubernetesDir + "/controller-manager.conf",
@@ -612,6 +613,7 @@ func TestGetControllerManagerCommand(t *testing.T) {
 			},
 			expected: []string{
 				"kube-controller-manager",
+				"--port=0",
 				"--bind-address=127.0.0.1",
 				"--leader-elect=true",
 				"--kubeconfig=" + kubeadmconstants.KubernetesDir + "/controller-manager.conf",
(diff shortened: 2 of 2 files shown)


SataQiu (approved):
/lgtm
/retest

/hold 
feel free to cancel when ready

rosti (approved):
Thanks @neolit123 !
/lgtm

This fixes kubernetes/kubeadm#2202 (not #2202).

scoopex:
I upgraded vom kuberntes 1.16 over 1.17.9 to 1.18.6.
It seems that my cluster runs fine, but is see the following bad component status (failing healthchecks)

```
# kubectl get componentstatuses 
NAME                 STATUS      MESSAGE                                                                                     ERROR
scheduler            Unhealthy   Get http://127.0.0.1:10251/healthz: dial tcp 127.0.0.1:10251: connect: connection refused   
controller-manager   Unhealthy   Get http://127.0.0.1:10252/healthz: dial tcp 127.0.0.1:10252: connect: connection refused   
etcd-1               Healthy     {"health":"true"}                                                                           
etcd-0               Healthy     {"health":"true"}                                                                           
etcd-2               Healthy     {"health":"true"}
```

I doesn't seem that your change is a good idea, because this breaks the "kubectl get componentstatuses".

Modifying the deployment yamls for scheduler and controller-manager (/etc/kubernetes/manifests/kube-scheduler.yaml, /etc/kubernetes/manifests/kube-controller-manager.yaml) on every controller node, adding the missing ports in the "—port=0" an applying the change configuration get component status back to work properly.
But i am almost sure that this is not the way to solve that correctly.

Nevertheless component status not seems to be a really good verification mechanism, because it simply returns  the status of the local scheduler or controller-manager but return the status of all etcd instances.
Is there a better possibility to monitor these components?

neolit123:
> I doesn't seem that your change is a good idea, because this breaks the "kubectl get componentstatuses".

"componentstatuses" should not be used in general.
but this PR should fix that problem:
https://github.com/kubernetes/kubernetes/pull/93171

> Nevertheless component status not seems to be a really good verification mechanism, because it simply returns the status of the local scheduler or controller-manager but return the status of all etcd instances.
Is there a better possibility to monitor these components?

it also doesn't work for multiple etcd nodes.
the alternative is to use kubectl (to probe control-plane Pods) / etcdctl (to probe the etcd cluster).

scoopex:
Thank you, very much. 
Breaking existing mechanisms without announcing that and without having a replacement is not a good idea :-)
The mentioned PR https://github.com/kubernetes/kubernetes/pull/93208 seems to help.

neolit123:
the change was announced in the 1.16.13 changelog:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.16.md#bug-or-regression
but we missed the requirement for the additional change in https://github.com/kubernetes/kubernetes/pull/93208

liruishan:
@neolit123 ，I have a question, if --port=0 is set, then the corresponding service is meaningless：
[root@master1 manifests]# kubectl describe svc -n kube-system kube-controller-manager
Name:              kube-controller-manager
Namespace:         kube-system
Labels:            k8s-app=kube-controller-manager
Annotations:       prometheus.io/port: 10252
                   prometheus.io/scrape: true
Selector:          component=kube-controller-manager
Type:              ClusterIP
IP:                None
IPFamily:          IPv4
Port:              http-metrics  10252/TCP
TargetPort:        10252/TCP
Endpoints:         192.168.1.36:10252
Session Affinity:  None
Events:            <none>

neolit123:
The service should not be present in newer k8s versions I assume.

PR #71174, opened 2018-11-17
Restore bootstrap in the background with fix to preserve kubeadm behavior

smarterclayton:
This restores the revert of #69890 due to a failure in kubeadm.

fixes https://github.com/kubernetes/kubernetes/issues/68686

kubeadm depends on being able to give the kubelet a high powered cert as `--kubeconfig` on the master and expects the kubelet to rotate that cert.  For now, preserve the behavior, even though this doesn't improve security on the masters (being able to read root owned files off the master is not a critical part of our threat model).   Add comments and a test to ensure this is clearly required.

```release-note
When a kubelet is using --bootstrap-kubeconfig and certificate rotation, it no longer waits for bootstrap to succeed before launching static pods.
```

PR #92727, opened 2020-07-01
avoid Kubelet request API Server anonymously before client certificate bootstrap finished

answer1991:
**What type of PR is this?**

/kind bug

Kubelet client certificate manager bootstrap certificate in background. Before the client certificate bootstrap finished, Kubelet always try to connect to API Server anonymously.

Thanks to all most Kubelet components writtern as controller mode,  Kubelet will re-try immediately if API Server reject the Kubelet anonymous request. And in most situation, Kubelet will connect to API Server successfully if certificate bootstrap finished and work normally.

However, API Server receives abnormal requests and may misleading cluster admin.

None

None

```release-note
avoid Kubelet request API Server anonymously before client certificate bootstrap finished
```

```docs

```

### DIFF
--- a/cmd/kubelet/app/server.go
+++ b/cmd/kubelet/app/server.go
@@ -874,6 +874,18 @@ func buildKubeletClientConfig(ctx context.Context, s *options.KubeletServer, nod
 		klog.V(2).Info("Starting client certificate rotation.")
 		clientCertificateManager.Start()
 
+		// we should wait certificate manager to finish client certificate bootstrap, otherwise Kubelet always try to connect
+		// to API Server anonymously before client certificate bootstrap finished.
+		if err := wait.PollImmediateUntil(100*time.Millisecond, func() (done bool, err error) {
+			if clientCertificateManager.Current() != nil {
+				return true, nil
+			}
+
+			return false, nil
+		}, ctx.Done()); err != nil {
+			return nil, nil, fmt.Errorf("bootstrap Kubelet client certificate timeout: %v", err)
+		}
+
 		return transportConfig, closeAllConns, nil
 	}
 


alejandrox1 (commented):
This looks good!
/ok-to-test
/priority important-longterm

answer1991:
/retest

I think the fix is a little tricky. We may consider add a new function in certificate manager like `CertificateBootstrapDone()` to tell us if certificate bootstrap done.

cmd/kubelet/app/server.go
```diff
@@ -870,6 +870,18 @@ func buildKubeletClientConfig(s *options.KubeletServer, nodeName types.NodeName)
 		klog.V(2).Info("Starting client certificate rotation.")
 		clientCertificateManager.Start()
 
+		// we should wait certificate manager to finish client certificate bootstrap, otherwise Kubelet always try to connect
+		// to API Server anonymously before client certificate bootstrap finished.
+		if err := wait.PollImmediate(100*time.Millisecond, 5*time.Minute, func() (done bool, err error) {
```
answer1991, line 875: After #92786 merged, I would like use context to poll certificate ready

answer1991:
/hold 

After #92786 merged, I will change it to use context to poll certificate ready.

answer1991:
/hold cancel
/assign @liggitt 

Could you help review this? thanks

cmd/kubelet/app/server.go
```diff
@@ -874,6 +874,18 @@ func buildKubeletClientConfig(ctx context.Context, s *options.KubeletServer, nod
 		klog.V(2).Info("Starting client certificate rotation.")
 		clientCertificateManager.Start()
 
+		// we should wait certificate manager to finish client certificate bootstrap, otherwise Kubelet always try to connect
+		// to API Server anonymously before client certificate bootstrap finished.
+		if err := wait.PollImmediateUntil(100*time.Millisecond, func() (done bool, err error) {
```
liggitt, line 879: I'm pretty sure blocking here would prevent a self-hosted kubelet (where the kube-apiserver was run as a static pod by the kubelet) from ever initializing. The function explicitly logs "Client rotation is on, will bootstrap in background"

liggitt: see https://github.com/kubernetes/kubernetes/pull/71174/files#diff-e0160f895346669c2e4495acfd858ea7R759

answer1991:
@liggitt Thanks for taking a look.

I found if `RotateCertificates` is disable(`--rotate-certificates==false`), then Kubelet will block in initializing before certificate bootstrap finished. see:

https://github.com/kubernetes/kubernetes/blob/70e09f2c24f170f985e4a6d7cf7101e32c97ea2a/cmd/kubelet/app/server.go#L880-L885

liggitt:
> I found if `RotateCertificates` is disable(`--rotate-certificates==false`), then Kubelet will block in initializing before certificate bootstrap finished.

That's a good catch, but we need a fix that moves that scenario (`--rotate-certificates==false --bootstrap-kubeconfig=...`) into a background goroutine, not one that fails and returns an error that would prevent kubelet startup. We clearly also need a test case around that scenario.

PR #92729, opened 2020-07-01
Fix searching for licenses in single-segment modules

### LINKED ISSUE #92728: LICENSES/vendor/go.opencensus.io/LICENSE is not present
**What happened**:

Running `hack/update-vendor-licenses.sh` manually includes this file in LICENSES.

Running `hack/verify-vendor-licenses.sh` complains if it is present.

The file was incorrectly dropped in https://github.com/kubernetes/kubernetes/pull/91366/files#r448584903

This needs to be resolved before 1.19 ships

/priority important-soon
/milestone v1.19
/cc @munnerz @thockin

liggitt:
/kind bug

Fixes #92728

```release-note
NONE
```

/cc @dims

### DIFF
--- a/hack/update-vendor-licenses.sh
+++ b/hack/update-vendor-licenses.sh
@@ -197,14 +197,14 @@ for PACKAGE in $(go list -m -json all | jq -r .Path | sort -f); do
     continue
   fi
   # Skip a directory if 1) it has no files and 2) all the subdirectories contain a go.mod file.
-  if [[ -z "$(find "${DEPS_DIR}/${PACKAGE}" -mindepth 1 -maxdepth 1 -type f)" ]]; then
+  if [[ -z "$(find "${DEPS_DIR}/${PACKAGE}/" -mindepth 1 -maxdepth 1 -type f)" ]]; then
       misses_go_mod=false
       while read -d "" -r SUBDIR; do
           if [[ ! -e "${SUBDIR}/go.mod" ]]; then
               misses_go_mod=true
               break
           fi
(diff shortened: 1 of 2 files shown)


munnerz:
😅 

/lgtm

dims:
good one.

thanks @liggitt 

/approve
/lgtm

hack/update-vendor-licenses.sh
```diff
@@ -197,14 +197,14 @@ for PACKAGE in $(go list -m -json all | jq -r .Path | sort -f); do
...
-  if [[ -z "$(find "${DEPS_DIR}/${PACKAGE}" -mindepth 1 -maxdepth 1 -type f)" ]]; then
+  if [[ -z "$(find "${DEPS_DIR}/${PACKAGE}/" -mindepth 1 -maxdepth 1 -type f)" ]]; then
       misses_go_mod=false
       while read -d "" -r SUBDIR; do
           if [[ ! -e "${SUBDIR}/go.mod" ]]; then
               misses_go_mod=true
               break
           fi
-      done < <(find "${DEPS_DIR}/${PACKAGE}" -mindepth 1 -maxdepth 1 -type d -print0)
+      done < <(find "${DEPS_DIR}/${PACKAGE}/" -mindepth 1 -maxdepth 1 -type d -print0)
```
munnerz, line 207: A heads up that this does technically change the value of `SUBDIR` in the loop (although in an inconsequential way) because `find` will *always* append a `/` to the given path: https://github.com/kubernetes/kubernetes/pull/92735/files#r448662042

munnerz:
Really couldn't think of the most appropriate sig for this, and it needs merging, so...

/sig architecture
? 🤷‍♂️

PR #91366, opened 2020-05-22
vendor: update google/cadvisor and opencontainers/runc

giuseppe:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

vendor a new version of github.com/opencontainers/runc and adapt the callers to the new API to inherit fixes for cgroup v2

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs
- [KEP]: https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20191118-cgroups-v2.md
```

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

PR #92735, opened 2020-07-01
Update hack/update-vendor-licenses.sh to handle unused indirect transitive dependencies

munnerz:

https://github.com/kubernetes/kubernetes/pull/91366/commits/78d295d168b84e9380e9c73982b2bdf2addd349f changed update-vendor-licenses.sh to skip failing on a module if it is empty and all of its subdirectories contain `go.mod` files.

Most of the time, this works. However, if this script is run against a vendor structure like so:

```
$ tree vendor/github.com/cpuguy83/go-md2man
vendor/github.com/cpuguy83/go-md2man
└── v2
    ├── LICENSE.md
    └── md2man
        ├── md2man.go
        └── roff.go
```

It will fail, because `go mod vendor` has pruned the `go.mod` file from the `vendor/github.com/cpuguy83/go-md2man/v2` directory as there are no imported go files in this directory (only in a subpackage of the module).

This PR replaces the previously added check with a more exhaustive one, which instead of relying on a `go.mod` file to determine whether a package is a module, uses the output of `go list -m`.

This has resolved issues in `jetstack/cert-manager` (where we also use this script) when bumping `k8s.io` dependencies to `v0.19.x`, which has a dependency on both `github.com/cpuguy83/go-md2man` (via `cmd/genman`) and `github.com/cpuguy83/go-md2man/v2` (via `github.com/spf13/cobra@1.0.0`) (demonstrated here when bumping our own dependency on cobra to 1.0.0: https://github.com/jetstack/cert-manager/pull/3060)

```release-note
NONE
```

/kind cleanup
/cc @liggitt @thockin

### DIFF
--- a/hack/update-vendor-licenses.sh
+++ b/hack/update-vendor-licenses.sh
@@ -186,6 +186,45 @@ if [ -f "${LICENSE_ROOT}/LICENSE" ]; then
   mv "${TMP_LICENSE_FILE}" "${TMP_LICENSES_DIR}/LICENSE"
 fi
 
+# only_contains_submodules will check whether a package is empty of files
+# and its subdirectories are all modules, which indicates we do not need
+# a LICENSE for this module as it is an indirect transitive dependency.
+only_contains_submodules () {
+  package="$1"
+  if [[ -z "$(find "${DEPS_DIR}/${package}/" -mindepth 1 -maxdepth 1 -type f)" ]]; then
+    # If the package does not contain files, check whether all subdirectories
+    # only contain modules.
+    while read -d "" -r SUBDIR; do
+      # Because we have a trailing '/' on the 'find' command's argument, we
+      # must strip off two '/' from the end of the package path before the
+      # sub-packages name.
+      subpackage="${package}/${SUBDIR#"${DEPS_DIR}/${package}//"}"
+
+      # If the subdirectory is already a module, don't recurse into deeper
+      # subdirectories of this subpackage.
+      if go list -m "${subpackage}"; then
+        continue
+      fi
+
+      # If the subdirectory is not a module, check if the subdirectory only_contains_submodules
+      if only_contains_submodules "${subpackage}"; then
+        continue
+      fi
+
+      # If the subdirectory is not a module, and its subdirectories contain files and
+      # are not themselves (or their parents) modules, then this package does not
+      # contain only submodules
+      return 1
+    done < <(find "${DEPS_DIR}/${package}/" -mindepth 1 -maxdepth 1 -type d -print0)
+  else
+    # If the package contains files, it does not only contain submodules.
+    return 1
+  fi
+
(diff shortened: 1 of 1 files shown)


hack/update-vendor-licenses.sh
```diff
@@ -186,8 +186,60 @@ if [ -f "${LICENSE_ROOT}/LICENSE" ]; then
...
+# a LICENSE for this module as it is an indirect transitive dependency.
+only_contains_submodules () {
+  package="$1"
+  if [[ -z "$(find "${DEPS_DIR}/${package}/" -mindepth 1 -maxdepth 1 -type f)" ]]; then
+    # If the package does not contain files, check whether all subdirectories
+    # only contain modules.
+    while read -d "" -r SUBDIR; do
+      # Because we have a trailing '/' on the 'find' command's argument, we
+      # must strip off two '/' from the end of the package path before the
+      # sub-packages name.
```
munnerz, line 213: The reason for actually ending the `find` command with a `/` (meaning we have to strip two of them off) is explained here: https://github.com/kubernetes/kubernetes/pull/92729

hack/update-vendor-licenses.sh
```diff
@@ -186,6 +186,45 @@ if [ -f "${LICENSE_ROOT}/LICENSE" ]; then
...
+      subpackage="${package}/${SUBDIR#"${DEPS_DIR}/${package}//"}"
+
+      # If the subdirectory is already a module, don't recurse into deeper
+      # subdirectories of this subpackage.
+      if go list -m "${subpackage}"; then
```
munnerz, line 205: This returns "not a known dependency" (with a non-zero exit code) if the given package is not a module, so it's okay to do the check without matching against the output of `go list -m all`

munnerz:
This can either replace #92729 or I can rebase on top of it after merge. This change includes the change to add a `/` to the end of arguments to `find` as well in the process of fixing this issue.

liggitt:
> This can either replace #92729 or I can rebase on top of it after merge.

I'd prefer to recover before a rewrite

hack/update-vendor-licenses.sh
liggitt, line 205: how expensive is this? I'm wary of running this many many times

munnerz: A previous iteration of this captured and compared the output from the one big `go list` call - I can revert back to that if needed.

That said, this line is only reached if we encounter an empty module, which is not common (i.e. happens once in `jetstack/cert-manager` and does not happen at all in `kubernetes/kubernetes`), so it doesn't add any time onto the *current* execution time.

munnerz:
/priority important-longterm
/sig architecture

if someone has a better sig for this, please go ahead and adjust 😅

liggitt:
/uncc
/cc @dims

trying to distribute out reviews

PR #91637, opened 2020-06-01
Adding new EndpointSlice Mirroring Controller

robscott:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
This adds an EndpointSlice mirroring controller to mirror custom Endpoints to EndpointSlices as [described by the KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190603-EndpointSlice-API.md#endpointslicemirroring-controller). 

**Does this PR introduce a user-facing change?**:
```release-note
Custom Endpoints are now mirrored to EndpointSlices by a new EndpointSliceMirroring controller.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

- [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190603-endpointslices/README.md)
- [Enhancement Issue](https://github.com/kubernetes/enhancements/issues/752) 

**Test Coverage**
Progress on required test coverage from the KEP.

**Unit Tests**
* Stale EndpointSlices managed by EndpointSliceMirroring controller are cleaned up by the controller when it starts up.
  _Covered by controller tests, primarily an issue for garbage collector since these are owned by Endpoints resources._

* Only Endpoints with the skip label set to "true" will be mirrored by the EndpointSliceMirroring controller.
  This has changed a bit, but is _covered by controller tests._

* Endpoints transitioning between values of the skip label should result in corresponding EndpointSlices being created or deleted.
  _Covered by controller tests, could potentially use additional coverage in e2e or integration tests_

* Mirrored EndpointSlices will:
  * Include the appropriate labels, endpoints, and ports of the corresponding Endpoints resource.
    _Covered by unit tests._
    
  * Refer to the original Endpoints resource with an OwnerReference and a kubernetes.io/service-name label.
    _Covered by unit tests._ 

  * Include a endpointslice.kubernetes.io/managed-by label set to endpointslicemirroring-controller.k8s.io.
    _Covered by unit tests._ 

* Endpoints with dual stack Endpoints should be represented with multiple EndpointSlices, at least 1 per IP family.
  _Covered by unit tests, could potentially add some scale with unit or integration tests._

* Endpoints with multiple subsets should be represented with multiple EndpointSlices, at least 1 per subset.
  _Covered by unit tests, could potentially add some scale with unit or integration tests._

* Endpoints with multiple subsets and IP families should be represented with multiple EndpointSlices, at least 1 per subset and IP family.
[...]

PR #92736, opened 2020-07-02
Graduating EndpointSliceProxying to beta for Linux

robscott:
**What type of PR is this?**
/kind feature

This graduates the `EndpointSliceProxying` feature gate to beta for Linux and adds a new alpha `WindowsEndpointSliceProxying` feature gate to guard the Windows implementation. This means that kube-proxy on Linux will consume EndpointSlices instead of Endpoints by default.

I spent some time looking for the best way to have different default values for the same feature gate depending on OS and everything I could find felt messy. This is dependent on this [PR](https://github.com/kubernetes/kubernetes/pull/91637) getting in (approved but not yet merged).

```release-note
kube-proxy now consumes EndpointSlices instead of Endpoints by default on Linux. A new alpha `WindowsEndpointSliceProxying` feature gate allows the feature to be enabled on Windows.
```

- [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190603-endpointslices/README.md)
- [Enhancement Issue](https://github.com/kubernetes/enhancements/issues/752) 

/sig network
/priority important-soon

### DIFF
--- a/cmd/kube-proxy/app/server_windows.go
+++ b/cmd/kube-proxy/app/server_windows.go
@@ -174,7 +174,7 @@ func newProxyServer(config *proxyconfigapi.KubeProxyConfiguration, cleanupAndExi
 		OOMScoreAdj:         config.OOMScoreAdj,
 		ConfigSyncPeriod:    config.ConfigSyncPeriod.Duration,
 		HealthzServer:       healthzServer,
-		UseEndpointSlices:   utilfeature.DefaultFeatureGate.Enabled(features.EndpointSliceProxying),
+		UseEndpointSlices:   utilfeature.DefaultFeatureGate.Enabled(features.WindowsEndpointSliceProxying),
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


aojea:
so ... once this is beta kube-proxy will use endpoint slices instead of endpoint, right?

Because that will mean that we'll need to adapt some tooling from the e2e test framework

robscott:
@aojea That's completely right. Let me know if I can help with updating e2e tests. For what it's worth, with EndpointSlice mirroring any custom Endpoints should get mirrored automatically, and any normal Endpoints would naturally have corresponding EndpointSlices created. So as long as all controllers are running, I'm hoping the required number of changes will be minimal, but I'm not too familiar with the tests that might need too change.

aojea:
> @aojea That's completely right. Let me know if I can help with updating e2e tests. For what it's worth, with EndpointSlice mirroring any custom Endpoints should get mirrored automatically, and any normal Endpoints would naturally have corresponding EndpointSlices created. So as long as all controllers are running, I'm hoping the required number of changes will be minimal, but I'm not too familiar with the tests that might need too change.

based in your comment there is no rush, just opening an issue for not forgetting and tracking the effort will be enough.

kumarvin123 (approved):
Looks good

thockin:
Thanks!

/lgtm
/approve

PR #92739, opened 2020-07-02
Add fail-open audit logs to validating admission webhook

### LINKED ISSUE #92738: Write audit logs when failing open a validating admission webhook request
The validating admission webhook, when the failing policy is set to Ignore, will fail open a webhook request. Consider write an audit log when a request is allowed by failing open.

**Why is this needed**:
Fail open events SHOULD be audit logged for integrity. For us specifically, we are currently using the image policy webhook which exports an audit log when failed open, and as we'd like to move to validating admission webhook, we'd want the same feature parity.

cnphil:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

In validating admission webhook, writes an audit log entry per each occurrence of webhook fail open event.  Refer to issue #92738 for more details.

Fixes #92738

```release-note
An audit log entry will be generated when a validating admission webhook is failing open.
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go
+++ b/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go
@@ -55,6 +55,10 @@ const (
 	PatchAuditAnnotationPrefix = "patch.webhook.admission.k8s.io/"
 	// MutationAuditAnnotationPrefix is a prefix for presisting webhook mutation existence in audit annotation.
 	MutationAuditAnnotationPrefix = "mutation.webhook.admission.k8s.io/"
+	// MutationAnnotationFailedOpenKeyPrefix in an annotation indicates
+	// the mutating webhook failed open when the webhook backend connection
+	// failed or returned an internal server error.
+	MutationAuditAnnotationFailedOpenKeyPrefix string = "failed-open." + MutationAuditAnnotationPrefix
 )
 
 var encodingjson = json.CaseSensitiveJSONIterator()
@@ -134,7 +138,9 @@ func (a *mutatingDispatcher) Dispatch(ctx context.Context, attr admission.Attrib
 		if reinvokeCtx.IsReinvoke() {
 			round = 1
 		}
-		changed, err := a.callAttrMutatingHook(ctx, hook, invocation, versionedAttr, o, round, i)
+
+		annotator := newWebhookAnnotator(versionedAttr, round, i, hook.Name, invocation.Webhook.GetConfigurationName())
+		changed, err := a.callAttrMutatingHook(ctx, hook, invocation, versionedAttr, annotator, o, round, i)
 		ignoreClientCallFailures := hook.FailurePolicy != nil && *hook.FailurePolicy == admissionregistrationv1.Ignore
 		rejected := false
 		if err != nil {
@@ -168,6 +174,9 @@ func (a *mutatingDispatcher) Dispatch(ctx context.Context, attr admission.Attrib
 		if callErr, ok := err.(*webhookutil.ErrCallingWebhook); ok {
 			if ignoreClientCallFailures {
 				klog.Warningf("Failed calling webhook, failing open %v: %v", hook.Name, callErr)
+
+				annotator.addFailedOpenAnnotation()
+
 				utilruntime.HandleError(callErr)
 
 				select {
@@ -198,9 +207,8 @@ func (a *mutatingDispatcher) Dispatch(ctx context.Context, attr admission.Attrib
 
 // note that callAttrMutatingHook updates attr
 
-func (a *mutatingDispatcher) callAttrMutatingHook(ctx context.Context, h *admissionregistrationv1.MutatingWebhook, invocation *generic.WebhookInvocation, attr *generic.VersionedAttributes, o admission.ObjectInterfaces, round, idx int) (bool, error) {
+func (a *mutatingDispatcher) callAttrMutatingHook(ctx context.Context, h *admissionregistrationv1.MutatingWebhook, invocation *generic.WebhookInvocation, attr *generic.VersionedAttributes, annotator *webhookAnnotator, o admission.ObjectInterfaces, round, idx int) (bool, error) {
 	configurationName := invocation.Webhook.GetConfigurationName()
-	annotator := newWebhookAnnotator(attr, round, idx, h.Name, configurationName)
 	changed := false
 	defer func() { annotator.addMutationAnnotation(changed) }()
 	if attr.Attributes.IsDryRun() {
@@ -338,20 +346,32 @@ func (a *mutatingDispatcher) callAttrMutatingHook(ctx context.Context, h *admiss
 }
 
 type webhookAnnotator struct {
-	attr                  *generic.VersionedAttributes
-	patchAnnotationKey    string
-	mutationAnnotationKey string
-	webhook               string
-	configuration         string
+	attr                    *generic.VersionedAttributes
+	failedOpenAnnotationKey string
+	patchAnnotationKey      string
+	mutationAnnotationKey   string
+	webhook                 string
+	configuration           string
 }
 
 func newWebhookAnnotator(attr *generic.VersionedAttributes, round, idx int, webhook, configuration string) *webhookAnnotator {
 	return &webhookAnnotator{
-		attr:                  attr,
-		patchAnnotationKey:    fmt.Sprintf("%sround_%d_index_%d", PatchAuditAnnotationPrefix, round, idx),
-		mutationAnnotationKey: fmt.Sprintf("%sround_%d_index_%d", MutationAuditAnnotationPrefix, round, idx),
-		webhook:               webhook,
-		configuration:         configuration,
+		attr:                    attr,
+		failedOpenAnnotationKey: fmt.Sprintf("%sround_%d_index_%d", MutationAuditAnnotationFailedOpenKeyPrefix, round, idx),
+		patchAnnotationKey:      fmt.Sprintf("%sround_%d_index_%d", PatchAuditAnnotationPrefix, round, idx),
+		mutationAnnotationKey:   fmt.Sprintf("%sround_%d_index_%d", MutationAuditAnnotationPrefix, round, idx),
+		webhook:                 webhook,
+		configuration:           configuration,
+	}
+}
+
+func (w *webhookAnnotator) addFailedOpenAnnotation() {
+	if w.attr == nil || w.attr.Attributes == nil {
+		return
+	}
+	value := w.webhook
+	if err := w.attr.Attributes.AddAnnotation(w.failedOpenAnnotationKey, value); err != nil {
+		klog.Warningf("failed to set failed open annotation for mutating webhook key %s to %s: %v", w.failedOpenAnnotationKey, value, err)
 	}
 }
 
--- a/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/testing/testcase.go
+++ b/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/testing/testcase.go
@@ -264,6 +264,18 @@ func ConvertToMutatingTestCases(tests []ValidatingTest, configurationName string
 				break
 			}
 		}
+		// Change annotation keys for Validating's fail open to Mutating's fail open.
+		failOpenAnnotations := map[string]string{}
+		for key, value := range t.ExpectAnnotations {
+			if strings.HasPrefix(key, "failed-open.validating.webhook.admission.k8s.io/") {
+				failOpenAnnotations[key] = value
+			}
+		}
+		for key, value := range failOpenAnnotations {
+			newKey := strings.Replace(key, "failed-open.validating.webhook.admission.k8s.io/", "failed-open.mutation.webhook.admission.k8s.io/", 1)
+			t.ExpectAnnotations[newKey] = value
+			delete(t.ExpectAnnotations, key)
+		}
 		r[i] = MutatingTest{t.Name, ConvertToMutatingWebhooks(t.Webhooks), t.Path, t.IsCRD, t.IsDryRun, t.AdditionalLabels, t.SkipBenchmark, t.ExpectLabels, t.ExpectAllow, t.ErrorContains, t.ExpectAnnotations, t.ExpectStatusCode, t.ExpectReinvokeWebhooks}
 	}
 	return r
@@ -408,6 +420,11 @@ func NewNonMutatingTestCases(url *url.URL) []ValidatingTest {
 
 			SkipBenchmark: true,
 			ExpectAllow:   true,
+			ExpectAnnotations: map[string]string{
+				"failed-open.validating.webhook.admission.k8s.io/round_0_index_0": "internalErr A",
+				"failed-open.validating.webhook.admission.k8s.io/round_0_index_1": "internalErr B",
+				"failed-open.validating.webhook.admission.k8s.io/round_0_index_2": "internalErr C",
+			},
 		},
 		{
 			Name: "match & fail (but disallow because fail close on nil FailurePolicy)",
@@ -502,8 +519,9 @@ func NewNonMutatingTestCases(url *url.URL) []ValidatingTest {
 				ObjectSelector:          &metav1.LabelSelector{},
 				AdmissionReviewVersions: []string{"v1beta1"},
 			}},
-			SkipBenchmark: true,
-			ExpectAllow:   true,
+			SkipBenchmark:     true,
+			ExpectAllow:       true,
+			ExpectAnnotations: map[string]string{"failed-open.validating.webhook.admission.k8s.io/round_0_index_0": "nilResponse"},
 		},
 		{
 			Name: "absent response and fail closed",
--- a/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go
+++ b/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go
@@ -38,6 +38,16 @@ import (
 	utiltrace "k8s.io/utils/trace"
 )
 
+const (
+	// ValidatingAuditAnnotationPrefix is a prefix for keeping noteworthy
+	// validating audit annotations.
+	ValidatingAuditAnnotationPrefix = "validating.webhook.admission.k8s.io/"
+	// ValidatingAuditAnnotationFailedOpenKeyPrefix in an annotation indicates
+	// the validating webhook failed open when the webhook backend connection
+	// failed or returned an internal server error.
+	ValidatingAuditAnnotationFailedOpenKeyPrefix = "failed-open." + ValidatingAuditAnnotationPrefix
+)
+
 type validatingDispatcher struct {
 	cm     *webhookutil.ClientManager
 	plugin *Plugin
@@ -92,7 +102,7 @@ func (d *validatingDispatcher) Dispatch(ctx context.Context, attr admission.Attr
 	errCh := make(chan error, len(relevantHooks))
 	wg.Add(len(relevantHooks))
 	for i := range relevantHooks {
-		go func(invocation *generic.WebhookInvocation) {
+		go func(invocation *generic.WebhookInvocation, idx int) {
 			defer wg.Done()
 			hook, ok := invocation.Webhook.GetValidatingWebhook()
 			if !ok {
@@ -127,6 +137,12 @@ func (d *validatingDispatcher) Dispatch(ctx context.Context, attr admission.Attr
 			if callErr, ok := err.(*webhookutil.ErrCallingWebhook); ok {
 				if ignoreClientCallFailures {
 					klog.Warningf("Failed calling webhook, failing open %v: %v", hook.Name, callErr)
+
+					key := fmt.Sprintf("%sround_0_index_%d", ValidatingAuditAnnotationFailedOpenKeyPrefix, idx)
+					value := hook.Name
+					if err := versionedAttr.Attributes.AddAnnotation(key, value); err != nil {
+						klog.Warningf("Failed to set admission audit annotation %s to %s for validating webhook %s: %v", key, value, hook.Name, err)
+					}
 					utilruntime.HandleError(callErr)
 					return
 				}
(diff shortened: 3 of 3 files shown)


cnphil:
Note that in parallel in looking into if it's feasible / worth it to add an integration test for this in `kubernetes/test/integration/apiserver/admissionwebhook/`, right now no tests there cover audit logs for validating admission webhook.

IMO, this PR alone with the included unit test is ready to be reviewed and merged at this point.

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go
```diff
@@ -126,6 +133,11 @@ func (d *validatingDispatcher) Dispatch(ctx context.Context, attr admission.Attr
 			if callErr, ok := err.(*webhookutil.ErrCallingWebhook); ok {
 				if ignoreClientCallFailures {
 					klog.Warningf("Failed calling webhook, failing open %v: %v", hook.Name, callErr)
+
+					key := hook.Name + "/" + ValidatingWebhookFailedOpenKeySuffix
+					if err := versionedAttr.Attributes.AddAnnotation(key, "true"); err != nil {
+						klog.Warningf("Failed to set admission audit annotation %s to %s for validating webhook %s: %v", key, "true", hook.Name, err)
```
liggitt, line 139: we're already logging a warning on the fail open, no need to log again that we couldn't annotate the audit event

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go
```diff
@@ -37,6 +37,13 @@ import (
 	utiltrace "k8s.io/utils/trace"
 )
 
+const (
+	// ValidatingWebhookFailedOpenKeySuffix in an annotation indicates the
+	// validating webhook failed open when the webhook backend connection
+	// failed or returned an internal server error.
+	ValidatingWebhookFailedOpenKeySuffix string = "failed-open"
```
liggitt, line 44: is there a reason to limit this to validating webhooks and not annotate for mutating webhooks that fail open as well?

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go
cnphil: Sure, will remove this line.

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go
```diff
@@ -126,6 +133,11 @@ func (d *validatingDispatcher) Dispatch(ctx context.Context, attr admission.Attr
 			if callErr, ok := err.(*webhookutil.ErrCallingWebhook); ok {
 				if ignoreClientCallFailures {
 					klog.Warningf("Failed calling webhook, failing open %v: %v", hook.Name, callErr)
+
+					key := hook.Name + "/" + ValidatingWebhookFailedOpenKeySuffix
```
liggitt, line 137: this name would collide with a `failed-open` audit annotation returned by the webhook... that does not seem desireable

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go
cnphil: no particular reason, it's just that GCP Binauthz only uses the validating webhook.  I can try to add the annotation to mutating webhooks.

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go
cnphil: Are you referring to the domain name or the `failed-open` suffix?

For the domain name, the webhook name was put there, instead of a k8s domain name, to distinguish on which webhook we failed open on, as multiple webhooks can be configured.

For the suffix, this is mainly to be backwards compatible with the ImagePolicyWebhook annotation key we've been using (https://github.com/kubernetes/kubernetes/blob/323f34858de18b862d43c40b2cced65ad8e24052/plugin/pkg/admission/imagepolicy/admission.go#L58). And IMO, fail-open is something that _should_ be handled by the client but not the webhook, so I would imagine it's fairly unlikely the webhook to write a failed-open audit annotation?

liggitt: > Are you referring to the domain name or the `failed-open` suffix?

The combination of the two... using the webhook name as the domain name means our `failed-open` annotation suffix is in a domain I would expect to be under the control of the webhook, not used by the kube-apiserver machinery.

cnphil: I changed to use "validation.webhook.admission.k8s.io/failed-open" and "mutation.webhook.admission.k8s.io/failed-open"

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go
cnphil: Added fail open log to mutation webhook as well.

cnphil:
All comments resolved, please take another look :)

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go
```diff
@@ -167,6 +171,8 @@ func (a *mutatingDispatcher) Dispatch(ctx context.Context, attr admission.Attrib
 		if callErr, ok := err.(*webhookutil.ErrCallingWebhook); ok {
 			if ignoreClientCallFailures {
 				klog.Warningf("Failed calling webhook, failing open %v: %v", hook.Name, callErr)
+
+				versionedAttr.Attributes.AddAnnotation(MutationAuditAnnotationPrefix  + MutationAuditAnnotationFailedOpenKeySuffix, "true")
```
roycaihw, line 175: `MutationAuditAnnotationPrefix` is a domain used for tracking webhook mutation existence ("webhook A was called, but did it mutated the API request or not?") purpose. You can add a separate domain for fail-open annotations

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go
```diff
@@ -126,6 +136,8 @@ func (d *validatingDispatcher) Dispatch(ctx context.Context, attr admission.Attr
 			if callErr, ok := err.(*webhookutil.ErrCallingWebhook); ok {
 				if ignoreClientCallFailures {
 					klog.Warningf("Failed calling webhook, failing open %v: %v", hook.Name, callErr)
+
+					versionedAttr.Attributes.AddAnnotation(ValidationAuditAnnotationPrefix  + ValidationAuditAnnotationFailedOpenKeySuffix, "true")
```
roycaihw, line 140: There can be multiple fail-open webhook calls for one request and this doesn't distinguish them

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go
cnphil: Yeah makes sense.  Any suggestions on what domain name to use?  (Not super familiar with the annotation naming convention here).

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go
cnphil: Sure will append the hook's name in the key.

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go
roycaihw: Maybe `fail-open.webhook.admission.k8s.io/`, or `{error/failure}.webhook.admission.k8s.io/` if we want to generalize this annotation in future

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go
roycaihw: Also there are some things to watch out: 

1. apiserver validates the [annotation keys](https://github.com/kubernetes/kubernetes/blob/e45a598a5ca3e722eedc2d491b0c82fd7e77dc24/staging/src/k8s.io/apiserver/pkg/admission/attributes.go#L207) with some length limit (253-char prefix and 63-char name), while [webhook name](https://github.com/kubernetes/kubernetes/blob/e45a598a5ca3e722eedc2d491b0c82fd7e77dc24/pkg/apis/admissionregistration/validation/validation.go#L213) has a 253-char limit
2. webhook names are not validated for [uniqueness](https://github.com/kubernetes/kubernetes/blob/e45a598a5ca3e722eedc2d491b0c82fd7e77dc24/staging/src/k8s.io/apiserver/pkg/admission/configuration/validating_webhook_manager.go#L87-L95) (UIDs are), and apiserver doesn't allow [overwrite](https://github.com/kubernetes/kubernetes/blob/e45a598a5ca3e722eedc2d491b0c82fd7e77dc24/staging/src/k8s.io/apiserver/pkg/admission/attributes.go#L155-L157) existing annotation key

I think in reality it should be rare to have webhooks with super long names, or even with duplicated names. But just watch out that the annotation could fail silently if that happened, given that the code ignores the returned error

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go
cnphil: Picked `<hook-name>.failure.webhook.admission.k8s.io` as the new key.

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go
cnphil: Thanks for the watch outs and pointers, really helpful.

I decided to prepend the hook's name in the front, making it `<hook-name>.failure.webhook.admission.k8s.io`, which would resolve the distinguish-ness issue.  I also considered make separate annotation key/value pairs for the hook names, e.g. `failure.webhook.admission.k8s.io/failed-open-hook-{1..N}` -> `<hook-name>`, but it might be too boilerplate. Let me know what you think.

Also I added a klog.Warning if the annotation was not added successfully (e.g. name was too long due to the reasons you mentioned above)

cnphil:
Addressed the comments by changed the annotation key scheme. Please take another look! ;)

cnphil:
Fixed a gofmt linting issue.  Rerunning some tests.

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go
```diff
@@ -54,6 +54,13 @@ const (
 	PatchAuditAnnotationPrefix = "patch.webhook.admission.k8s.io/"
 	// MutationAuditAnnotationPrefix is a prefix for presisting webhook mutation existence in audit annotation.
 	MutationAuditAnnotationPrefix = "mutation.webhook.admission.k8s.io/"
+	// FailureAuditAnnotationPrefix is a prefix for keeping noteworthy
+	// failure audit annotations.
+	FailureAuditAnnotationPrefix = "failure.webhook.admission.k8s.io/"
```
roycaihw, line 59: the prefix would be indistinguishable if there are one validating webhook and one mutating webhook sharing the same name (but having overlapping rules), but that could be rare as well. I don't feel strongly about it

cnphil:
Forced pushed to reword a commit message to remove an '@' there, which caused a test failure on the commit message itself.

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go
cnphil: Ack.  I prefer just to leave it as is, I think it's a pretty rare situation.

liggitt:

cnphil:
Ping for an approval label :)

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go
```diff
@@ -167,6 +174,15 @@ func (a *mutatingDispatcher) Dispatch(ctx context.Context, attr admission.Attrib
 		if callErr, ok := err.(*webhookutil.ErrCallingWebhook); ok {
 			if ignoreClientCallFailures {
 				klog.Warningf("Failed calling webhook, failing open %v: %v", hook.Name, callErr)
+
+				key := hook.Name + "." + FailureAuditAnnotationPrefix + FailureAuditAnnotationFailedOpenKeySuffix
+				value := "true"
+				if err := versionedAttr.Attributes.AddAnnotation(key, value); err != nil {
+					// In extreme rare cases, super long `hook.Name` could make the key
```
liggitt, line 181: why would we not use the same key computation as newWebhookAnnotator which handles multiple invocations, conflicting names, avoids making the webhook name part of the key and avoids length issues?

cnphil:
/reopen

/remove-lifecycle rotten

I'm back actively working on getting this PR merged.

mikedanese:
Please squash commits.

/lgtm

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go
liggitt: Constructing an annotation key that simply doesn't work for some valid webhook names isn't reliable (this goes for both mutating and validation webhooks).

A webhook name must pass this validation:

* https://github.com/kubernetes/kubernetes/blob/14df10f60bbd68216f8ab9d3f30d1f938d7b659e/pkg/apis/admissionregistration/validation/validation.go#L310

* https://github.com/kubernetes/kubernetes/blob/14df10f60bbd68216f8ab9d3f30d1f938d7b659e/pkg/apis/admissionregistration/validation/validation.go#L262

* https://github.com/kubernetes/kubernetes/blob/14df10f60bbd68216f8ab9d3f30d1f938d7b659e/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L78-L90

That means a webhook name can be up to 253 characters and must pass `IsDNS1123Subdomain`.

An audit annotation key must pass this validation:

* https://github.com/kubernetes/kubernetes/blob/14df10f60bbd68216f8ab9d3f30d1f938d7b659e/staging/src/k8s.io/apiserver/pkg/admission/attributes.go#L141-L142

* https://github.com/kubernetes/kubernetes/blob/14df10f60bbd68216f8ab9d3f30d1f938d7b659e/staging/src/k8s.io/apiserver/pkg/admission/attributes.go#L202-L211

* https://github.com/kubernetes/kubernetes/blob/14df10f60bbd68216f8ab9d3f30d1f938d7b659e/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L42-L71

That means an audit annotation must have `$prefix/$suffix`, $prefix must pass `IsDNS1123Subdomain`, and $suffix cannot be longer than 63 characters.

That means the only place in an audit annotation key the webhook name can go is to be used as the entire $prefix. However:

* Using `$name/failed-open` would conflict with a `failed-open` audit annotation returned by the webhook itself
* This also claims use of a prefix not under Kubernetes' control (the point of namespacing the annotations is to avoid conflicts between uncoordinated parties).

If you want a stable audit annotation key under Kubernetes' control, something like the following would work:

* `mutating.webhook.admission.k8s.io/failed-open`
* `validating.webhook.admission.k8s.io/failed-open`

The value for those annotations could be the names of the webhooks that failed open.

mikedanese:
Actually, cancelling until https://github.com/kubernetes/kubernetes/pull/92739/files#r633906487 is resolved.

/lgtm cancel

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go
```diff
@@ -168,6 +172,12 @@ func (a *mutatingDispatcher) Dispatch(ctx context.Context, attr admission.Attrib
 		if callErr, ok := err.(*webhookutil.ErrCallingWebhook); ok {
 			if ignoreClientCallFailures {
 				klog.Warningf("Failed calling webhook, failing open %v: %v", hook.Name, callErr)
+
+				key := MutationAuditAnnotationPrefix + MutationAuditAnnotationFailedOpenKeySuffix
```
liggitt, line 176: Audit annotations are a string[string] map. Using a constant key like this means that if multiple webhooks fail open, only the first one will get recorded in an audit annotation.

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/testing/testcase.go
```diff
@@ -408,6 +408,11 @@ func NewNonMutatingTestCases(url *url.URL) []ValidatingTest {
 
 			SkipBenchmark: true,
 			ExpectAllow:   true,
+			ExpectAnnotations: map[string]string{
+				"validating.webhook.admission.k8s.io/failed-open": "internalErr A",
+				"validating.webhook.admission.k8s.io/failed-open": "internalErr B",
+				"validating.webhook.admission.k8s.io/failed-open": "internalErr C",
```
liggitt, line 414: this unit test demonstrates the problem with this approach... a map can't have multiple entries for the same key

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go
```diff
@@ -127,6 +137,12 @@ func (d *validatingDispatcher) Dispatch(ctx context.Context, attr admission.Attr
 			if callErr, ok := err.(*webhookutil.ErrCallingWebhook); ok {
 				if ignoreClientCallFailures {
 					klog.Warningf("Failed calling webhook, failing open %v: %v", hook.Name, callErr)
+
+					key := ValidatingAuditAnnotationPrefix + ValidatingAuditAnnotationFailedOpenKeySuffix
```
liggitt, line 141: same issue here with a constant key and multiple entries

cnphil:
I realized using a constant key for all fail opens doesn't work, which liggitt pointed out in the last commit.  I changed the format for the logs a bit to:
key: "failed-open.validating.webhook.admission.k8s.io/round_%d_index_%d"
value:  "%s" (webhook's name)
(and same format but different string for mutation webhooks)

This will resolve the constant key issue by appending the webhook numerical indices to the key, I prepended "failed-open" to the domain name in the key to signify this is a fail open log, and put the webhook's name in the annotation value to help customer consistently identify which webhook failed open in the logs.

Hope this resolves your concern, and let me know if you want a different format for the annotation.

Thanks for the review and please take another look! :)

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go
```diff
@@ -38,6 +38,16 @@ import (
...
 )
 
+const (
+	// ValidatingAuditAnnotationPrefix is a prefix for keeping noteworthy
+	// validating audit annotations.
+	ValidatingAuditAnnotationPrefix = "validating.webhook.admission.k8s.io/"
+	// ValidatingAuditAnnotationFailedOpenKeyPrefix in an annotation indicates
+	// the validating webhook failed open when the webhook backend connection
+	// failed or returned an internal server error.
+	ValidatingAuditAnnotationFailedOpenKeyPrefix = "failed-open."
```
liggitt, line 48:
```suggestion
	ValidatingAuditAnnotationFailedOpenKeyPrefix = "failed-open."+ValidatingAuditAnnotationPrefix
```

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go
```diff
@@ -55,6 +55,10 @@ const (
 	PatchAuditAnnotationPrefix = "patch.webhook.admission.k8s.io/"
 	// MutationAuditAnnotationPrefix is a prefix for presisting webhook mutation existence in audit annotation.
 	MutationAuditAnnotationPrefix = "mutation.webhook.admission.k8s.io/"
+	// MutationAnnotationFailedOpenKeyPrefix in an annotation indicates
+	// the mutating webhook failed open when the webhook backend connection
+	// failed or returned an internal server error.
+	MutationAuditAnnotationFailedOpenKeyPrefix string = "failed-open."
```
liggitt, line 61:
```suggestion
	MutationAuditAnnotationFailedOpenKeyPrefix string = "failed-open."+MutationAuditAnnotationPrefix
```

cnphil:
Thanks for the review @liggitt .  I made the suggested changes and resolved the comments.  Please take another look :)

liggitt:
/hold

certs used in unit tests expired. fixup in https://github.com/kubernetes/kubernetes/pull/102224

staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go
```diff
@@ -168,6 +174,9 @@ func (a *mutatingDispatcher) Dispatch(ctx context.Context, attr admission.Attrib
 		if callErr, ok := err.(*webhookutil.ErrCallingWebhook); ok {
 			if ignoreClientCallFailures {
 				klog.Warningf("Failed calling webhook, failing open %v: %v", hook.Name, callErr)
+
+				annotator.addFailedOpenAnnotation()
```
liggitt, line 178: I think this should move into the `default` case below so we only record the fail-open when we're actually going to continue... if the request timed out and we're going to reject with a timeout error, a fail-open is misleading here

PR #88503, opened 2020-02-25
Adding AppProtocol to Service and Endpoints Ports

robscott:
**What type of PR is this?**
/kind api-change

**What this PR does / why we need it**:
This adds the `AppProtocol` field to Service and Endpoints resources, gated by a new `ServiceAppProtocol` feature gate.

**Does this PR introduce a user-facing change?**:
```release-note
AppProtocol is a new field on Service and Endpoints resources, enabled with the ServiceAppProtocol feature gate.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:
- KEP: https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20191227-app-protocol.md
- Enhancement Issue: https://github.com/kubernetes/enhancements/issues/1507

/sig network
/priority important-soon
/cc @liggitt 
/assign @thockin

PR #92744, opened 2020-07-02
Implement ServiceSpec.AllocateLoadBalancerNodePorts

uablrek:
**What type of PR is this?**

/kind feature
/sig network

Implements [KEP 1864-disable-lb-node-ports](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1864-disable-lb-node-ports)

**NOTE**; This is a Work In Progress and shall **not** be merged (yet).

```release-note
Automatic allocation of NodePorts for services with type LoadBalancer can now be disabled by setting the (new) parameter
Service.spec.allocateLoadBalancerNodePorts=false. The default is to allocate NodePorts for services with type LoadBalancer which is the existing behavior.
```

- [KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1864-disable-lb-node-ports)

```docs

```


uablrek:
/cc @andrewsykim 

The `AllocateLoadBalancerNodePorts` still defaults to "false" but I wanted to have a PR in place asap for discussions and as a place-holder.

rikatz:
/cc @rikatz 

Wanna take a look later :)

uablrek:
The allocateLoadBalancerNodePorts field is visible in output from `kubectl get svc -o json ...` *only* when the value is "true". If the value is false (the current default which will be changed) the field is hidden;
```
$ kubectl get -o json service/nodeport-disabled-3-ipv6 | jq .spec
{
  "allocateLoadBalancerNodePorts": true,
  "clusterIP": "fd00:4000::a567",
  "externalTrafficPolicy": "Cluster",
  "ipFamily": "IPv6",
  "ports": [
    {
      "nodePort": 30657,
      "port": 5001,
      "protocol": "TCP",
      "targetPort": 5001
    }
  ],
  "selector": {
    "app": "mconnect-daemonset"
  },
  "sessionAffinity": "None",
  "type": "LoadBalancer"
}
```
But;
```
kubectl get -o json service/nodeport-disabled-2-ipv6 | jq .spec
{
  "clusterIP": "fd00:4000::bf4",
  "externalTrafficPolicy": "Cluster",
  "ipFamily": "IPv6",
  "ports": [
    {
      "nodePort": 31734,
      "port": 5001,
      "protocol": "TCP",
      "targetPort": 5001
    }
  ],
  "selector": {
    "app": "mconnect-daemonset"
  },
  "sessionAffinity": "None",
  "type": "LoadBalancer"
}
```
This service does not have allocateLoadBalancerNodePorts defined.

@andrewsykim Is this something we can affect?

Ideal would be that allocateLoadBalancerNodePorts:true will be hidden once "true" becomes the default. Then users not caring about the function would not eve see it.

uablrek:
A possibility may be to rename the field; "disableLoadBalancerNodePorts" with a default of "false".

uablrek:
Came across this, see the "Use pointers" comment;
https://stackoverflow.com/questions/37756236/json-golang-boolean-omitempty/

Seem to give you a 3-state var; true, false, unset.

Would it be feasible to use a pointer in the ServiceSpec struct?
```
	// allocateLoadBalancerNodePorts defines if NodePorts shall be automatically for services with type LoadBalancer.
	// Default is "true". It may be set to "false" in clusters that don't have an external load-balancer that uses NodePorts.
	// This is often the case in private clouds that for instance uses "metallb".
	AllocateLoadBalancerNodePorts *bool `json:"allocateLoadBalancerNodePorts,omitempty" protobuf:"bytes,17,opt,name=allocateLoadBalancerNodePorts"`
```

uablrek:
Tried it, works like a charm. I'll do the switch.

This also allows the admission hook to be more intelligent; if unset, (nil) set a default, if set, leave it.

uablrek:
I know that actually checking the flag shall not go into alpha according to the KEP but I could not resist commiting an update that seem to be sufficient;
```
$ kubectl get svc
NAME                      TYPE           CLUSTER-IP        EXTERNAL-IP   PORT(S)          AGE
mconnect-daemonset-ipv4   LoadBalancer   12.0.85.138       <pending>     5001:32633/TCP   2m40s
mconnect-daemonset-ipv6   LoadBalancer   fd00:4000::407c   <pending>     5001:32350/TCP   2m40s
nodeport-disabled-ipv4    LoadBalancer   12.0.93.198       <pending>     5001/TCP         2m16s
nodeport-disabled-ipv6    LoadBalancer   fd00:4000::7f51   <pending>     5001/TCP         2m16s
nodeport-enabled-ipv4     LoadBalancer   12.0.88.112       <pending>     5001:30090/TCP   2m16s
nodeport-enabled-ipv6     LoadBalancer   fd00:4000::9f8e   <pending>     5001:32449/TCP   2m16s
```
In this case the `allocateLoadBalancerNodePorts` flag is `nil` (unset) in the "mconnect-daemonset" svcs, expicitly set to `false` in "nodeport-disabled" and explicitly set to `true` in "nodeport-enabled".

I let the reviewers take a look and see if I have missed anything, and I can revert the commit if needed.

uablrek:
Tests keep on failing. When searching for "allocateLoadBalancerNodePorts" in the `pull-kubernetes-bazel-test` logs, this was repeated 3 times;
```
    TestCompatibility/core.v1.Service/HEAD: compatibility.go:302: json differs
    TestCompatibility/core.v1.Service/HEAD: compatibility.go:303:   strings.Join({
          	... // 76 identical lines
          	`    "topologyKeys": [`,
          	`      "29"`,
        - 	"    ]",
        + 	"    ],",
        + 	`    "allocateLoadBalancerNodePorts": true`,
          	"  },",
          	`  "status": {`,
          	... // 10 identical lines
          }, "\n")
        
    TestCompatibility/core.v1.Service/HEAD: compatibility.go:308: yaml differs
    TestCompatibility/core.v1.Service/HEAD: compatibility.go:309:   strings.Join({
          	... // 30 identical lines
          	`  uid: "7"`,
          	"spec:",
        + 	"  allocateLoadBalancerNodePorts: true",
          	`  clusterIP: "24"`,
          	"  externalIPs:",
          	... // 31 identical lines
          }, "\n")
```
I suspect I have missed some file generation or that test-data must be updated.

@andrewsykim Can you please have a look and see what I have missed?

In previous failed test there was printouts liks "run; hack/..." which was very helpful. Thanks a lot to whoever added them.

andrewsykim:
```
TestCompatibility/core.v1.Service/HEAD: compatibility.go:329: if the diff is expected because of a new type or a new field, re-run with UPDATE_COMPATIBILITY_FIXTURE_DATA=true to update the compatibility data
```
^ saw this in the logs and there's a hack script to update this `hack/update-generated-api-compatibility-data.sh`.

uablrek:
I can't relate the failing tests to the PR;

#### pull-kubernetes-e2e-kind

14 failed tests. No match on "allocatel", but a few noteport tests.

#### pull-kubernetes-verify

I can't find what fails in this test(?!)

#### pull-kubernetes-kubemark-e2e-gce-big

Some of these;
```
 W0710 16:16:16.424] ERROR: (gcloud.compute.ssh) [/usr/bin/ssh] exited with return code [255]. 
```

rikatz:
I'll try to see if a second opinion gets something here but right now, as we're almost in the Code Freeze time is this something that might be overloading the test suite, specially from gce-big? I might be speaking non-sense things, but who knows :man_shrugging: 

I've seen an issue with tide yesterday, so I'm 'afraid' this might be related, as it seems to be some test/merge queue for the code freeze

rikatz:
It seems that pull-kubernetes-verify is failing a lot: https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-verify

It appears that the error is because of some sort of timeout (you can see that other jobs with more than 2 hours failed almost the same way).

uablrek:
@rikatz  Thanks for the help I really appreciate it.

The "pull-kubernetes-verify" does not provide an ERROR or FAIL (my search patterns). It may simply be terminated since it executes >2h?

My next plan is to remove the handling of the flag. It is really trivial but should not really go into the first stage according to the KEP. I don't think it helps though :disappointed: 

The API is changed to I fear that I have still missed some update of generated files or test-data.

rikatz:
By the [Job definition](https://prow.k8s.io/prowjob?prowjob=ca4cac7a-c2b6-11ea-83ba-96398ae925ba) yes, the timeout is 2h. I'm still trying to understand all this test-infra world also :) 

I'll try to fork the changes here this weekend and maybe verify in some other PR that changes the API if and what process was followed :)

By know, I remember that one that Rob Scott has introduced the protocol name (https://github.com/kubernetes/kubernetes/pull/88503) but as far I've checked here everything is almost in the same way :)

Thank you for making this happen @uablrek it's going to be very helpful in our env

uablrek:
/retest

pull-kubernetes-conformance-kind-ga-only-parallel has no failed tests. Another premature termination?

uablrek:
@andrewsykim I think the last failing test is a flaky test and that this PR is really ready for review.

I removed the implementation of the flag when I rebased. The tests result were improved but I don't know if that's the reason. In any case this is according to step 1 in the KEP. The attribute is added, nothing else.

I will go on a short trip now and will be back friday. I will still check the status but be less active.

pkg/apis/core/types.go
```diff
@@ -3658,6 +3658,13 @@ type ServiceSpec struct {
 	// If this is not specified or empty, no topology constraints will be applied.
 	// +optional
 	TopologyKeys []string
+
+	// allocateLoadBalancerNodePorts defines if NodePorts shall be
+	// automatically for services with type LoadBalancer.  Default is
```
andrewsykim, line 3663: "automatically allocated"

pkg/apis/core/types.go
```diff
@@ -3658,6 +3658,13 @@ type ServiceSpec struct {
...
+	// "true". It may be set to "false" in clusters that don't have an
```
andrewsykim, line 3664: It may be set to "false" if the cluster load-balancer does not rely on NodePorts.

andrewsykim (commented):
really minor comments about API docs

staging/src/k8s.io/api/core/v1/types.go
```diff
@@ -4137,6 +4137,11 @@ type ServiceSpec struct {
 	// If this is not specified or empty, no topology constraints will be applied.
 	// +optional
 	TopologyKeys []string `json:"topologyKeys,omitempty" protobuf:"bytes,16,opt,name=topologyKeys"`
+
+	// allocateLoadBalancerNodePorts defines if NodePorts shall be automatically for services with type LoadBalancer.
+	// Default is "true". It may be set to "false" in clusters that don't have an external load-balancer that uses NodePorts.
+	// This is often the case in private clouds that for instance uses "metallb".
```
andrewsykim, line 4143: This needs to match the comment from the internal type

andrewsykim (commented):
/assign @thockin 
/milestone v1.20 

This is just the API field addition without any logic to allow for rollback

andrewsykim:
@uablrek I think we may be able to implement the actual feature in v1.20 with an alpha feature gate and then in v1.21 it can be GA and on by default. Not sure if this is standard for new API fields, will defer to @thockin on that.

thockin (commented):
This also needs to be feature gated so we can do 1 release where it is allowed to exist but not usable by default (to allow for rollbacks).

andrewsykim:
> This also needs to be feature gated so we can do 1 release where it is allowed to exist but not usable by default (to allow for rollbacks).

Is there precedent for adding the API field without implementing the feature (hence no feature gate) or do we always want a feature gate for any new fields to core APIs? Feature gate seems overkill for this but I'm not sure what's standard for this.

uablrek:
Is it even possible to featureGate a declaration-only? Can the field be made "unseen" in protocols (generated files) so the new field is never visible "on-wire" if the featureGate is disabled?

andrewsykim (commented):
@uablrek yes it should be possible, we have some examples of this for dual-stack https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/core/service/strategy.go#L98-L119

thockin:
Even when gated off, the field is visible in the API docs, it is just wiped before storing.

staging/src/k8s.io/api/core/v1/types.go
thockin: +1

Also - mark it as +optional

uablrek:
I have aligned the comments and will check feature-gates soon.

Sorry for the delay, I have been on vacation 2 weeks. Back now.

uablrek:
@andrewsykim What shall I call the feature-gate?
To name it "AllocateLoadBalancerNodePorts" doesn't feel right.

andrewsykim:
ServiceLBWithoutNodePorts?

uablrek:
"ServiceLBNodePortControl"? I would like to avoid pro/con words like "Allocate" or "Without".

andrewsykim:
`ServiceLBNodePortControl` or `ServiceLBNodePortToggle` works for me. We can pick one and revisit naming once the implementation is more complete, maybe we'll think of better names along the way

uablrek:
https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/core/service/strategy.go#L98-L119 seems to inject sane init-values. For AllocateLoadBalancerNodePorts the default `nil` initiates by go is fine so I don't see any need for something similar in this PR.

There is another place here; https://github.com/kubernetes/kubernetes/blob/db28b0239a1d09e46507e0b320f0b4ffa1cbc8b6/pkg/registry/core/service/storage/rest.go#L430-L433

I must think a bit on this.

pkg/features/kube_features.go
```diff
@@ -645,6 +645,12 @@ const (
 	//
 	// Disables Accelerator Metrics Collected by Kubelet
 	DisableAcceleratorUsageMetrics featuregate.Feature = "DisableAcceleratorUsageMetrics"
+
+	// owner: @andrewsykim @uablrek
+	// alpha: v1.20
+	//
+	// Allows control if NodePorts shall be created for services with type: LoadBalancer
```
andrewsykim, line 652: We probably want to call out the specific field in Service that this is enabling

pkg/registry/core/service/storage/rest.go
```diff
@@ -200,8 +200,16 @@ func (rs *REST) Create(ctx context.Context, obj runtime.Object, createValidation
 	defer nodePortOp.Finish()
 
 	if service.Spec.Type == api.ServiceTypeNodePort || service.Spec.Type == api.ServiceTypeLoadBalancer {
-		if err := initNodePorts(service, nodePortOp); err != nil {
-			return nil, err
+		if utilfeature.DefaultFeatureGate.Enabled(features.ServiceLBNodePortControl) {
```
andrewsykim, line 203: I don't think we need the feature gate check here. The `AllocateLoadBalancerNodePorts` field will be nil-ed out in `PrepareForCreate` or `PrepareForUpdate` if the gate was disabled so we can just check for nil field here.

pkg/apis/core/types.go
```diff
@@ -3658,6 +3658,11 @@ type ServiceSpec struct {
 	// If this is not specified or empty, no topology constraints will be applied.
 	// +optional
 	TopologyKeys []string
+
+	// allocateLoadBalancerNodePorts defines if NodePorts shall be automatically
+	// allocated for services with type LoadBalancer.  Default is "true". It may be
+	// set to "false" if the cluster load-balancer does not rely on NodePorts.
```
andrewsykim, line 3664: +optional here too

uablrek:
Thanks for the clearification, I see it now;
> if the gate is disabled

This is what I missed. Of course if the user first enable the featureGate, use it, and then disable it the default value must be restored.

uablrek:
Updated after review, rebased and force-push.

I will test the feature-gate enable/use/disable logic and check https://github.com/kubernetes/kubernetes/blob/db28b0239a1d09e46507e0b320f0b4ffa1cbc8b6/pkg/registry/core/service/storage/rest.go#L430-L433

Then this is about finshed I hope.

pkg/registry/core/service/strategy.go
```diff
@@ -205,6 +210,17 @@ func topologyKeysInUse(svc *api.Service) bool {
 	return len(svc.Spec.TopologyKeys) > 0
 }
 
+// returns true if svc.Spec.ServiceIPFamily field is in use
+func allocateLoadBalancerNodePortsInUse(svc *api.Service) bool {
+	if svc == nil {
+		return false
+	}
+	if svc.Spec.AllocateLoadBalancerNodePorts != nil && !*svc.Spec.AllocateLoadBalancerNodePorts {
```
andrewsykim, line 218: I _think_ the "in use" check should only check if the field is not nil and not the actual bool value.

pkg/registry/core/service/storage/rest.go
```diff
@@ -199,7 +199,8 @@ func (rs *REST) Create(ctx context.Context, obj runtime.Object, createValidation
 	nodePortOp := portallocator.StartOperation(rs.serviceNodePorts, dryrun.IsDryRun(options.DryRun))
 	defer nodePortOp.Finish()
 
-	if service.Spec.Type == api.ServiceTypeNodePort || service.Spec.Type == api.ServiceTypeLoadBalancer {
+	if service.Spec.Type == api.ServiceTypeNodePort ||
+		(service.Spec.Type == api.ServiceTypeLoadBalancer && (service.Spec.AllocateLoadBalancerNodePorts == nil || *service.Spec.AllocateLoadBalancerNodePorts)) {
```
andrewsykim, line 203: I think we also need to gate `updateNodePorts` on line 432 

https://github.com/kubernetes/kubernetes/blob/cd6c4921fc11d00dd60324341061a057da9cb54e/pkg/registry/core/service/storage/rest.go#L430-L434

andrewsykim (commented):
@uablrek can we add some unit tests for the allocate logic? There's a few examples from dual-stack you can reference from [TestServiceRegistryCreate](https://github.com/kubernetes/kubernetes/blob/e3a0ca27314ca60ebe22eacaeaba158b8c494e3a/pkg/registry/core/service/storage/rest_test.go#L305)

pkg/registry/core/service/strategy.go
uablrek: The reasoning was that if a user sets AllocateLoadBalancerNodePorts=true (a no-op) when the feature-gate is enabled then the AllocateLoadBalancerNodePorts would still be visible when doing "kubectl get service -o json ..." even when the feature-gate is disabled. With the current condition it would become hidden again.

pkg/registry/core/service/storage/rest.go
uablrek: Yes, I have commented that myself. But I want to make sure that it is possible to remove the nodePort once the AllocateLoadBalancerNodePorts is set to "false" for a service.

pkg/registry/core/service/strategy.go
andrewsykim: But when the gate is disabled we would nil out the field in line 190 so it would be hidden again right?

pkg/registry/core/service/storage/rest.go
andrewsykim: I might be remembering incorrectly, but I think in the KEP we said if a node port is already set and `AllocateLoadBalancerNodePorts=false`, then we should do nothing and the user needs to update the Service themselves to remove the allocated node port. This means `updateNodePorts` should not run when `AllocateLoadBalancerNodePorts=false`, this leaves the node port as-is but also allows user to remove it without the allocator re-allocating it on removal.

pkg/registry/core/service/strategy.go
uablrek: No, it would not be set to nil since the "inUse" function will return true unless it is already nil. Compare to the `serviceIPFamilyInUse` function which will make sure the value is not set to `nil` unless it already is `nil` which makes only the svc == nil check useful.

pkg/registry/core/service/storage/rest.go
uablrek: You are absolutely right. I did not understand the update function fully.

uablrek: I will fix this and test it but first I must re-base/re-generate (sigh...)

pkg/registry/core/service/strategy.go
andrewsykim: Oh I see -- so if the gate is disabled but AllocateLoadBalanceNodePort was set to a value that is not the default (false) we shouldn't nil it. 

But even if it was explicitly set to `true` shouldn't we leave it instead of setting it to nil? The current logic would return `false` here if `AllocateLoadBalanceNodePort = true` which would result in a nil field on line 190.

andrewsykim: In the case of dual-stack, we only check if IPFamily is nil, we don't care what the value is.

https://github.com/kubernetes/kubernetes/blob/db28b0239a1d09e46507e0b320f0b4ffa1cbc8b6/pkg/registry/core/service/strategy.go#L194-L196

uablrek: > so if the gate is disabled but AllocateLoadBalanceNodePort was set to a value that is not the default (false) we shouldn't nil it.

Yes. Compare with the `topologyKeysInUse` function. The value is not set to nil if there are defined values.

The `serviceIPFamilyInUse()` on the other hand is useless. It returns true only if the value is != nil and the negated condition `!serviceIPFamilyInUse(oldSvc)` at line 179 make sure that the value is only set to nil it is _already nil_. I think the intended function would be to return true (indicating "in-use") only if the ipFamily was different from the "main" family (default). But this will change soon anyway when the ipFamilies becomes an array.

andrewsykim: > Yes. Compare with the topologyKeysInUse function. The value is not set to nil if there are defined values.

Right, so I think we're on the same page in that we shouldn't set the field to nil if it is already set. But I think when the field is set to `true`, we should leave the field as `true` as opposed to setting it to nil. In order to do this, we need to change this if condition to:

```go
if svc.Spec.AllocateLoadBalancerNodePorts != nil {
    return true
}
```

Otherwise on rollback we would drop the field that was set. Although behaviourally there is no difference, it would still be incorrect to drop that field. If `nil` and `true` that different behaviors (not in this case, but true for other fields), we would only do a `nil` check here as well.

uablrek: With that condition the function is useless since it _will_ only set the field to nil if it already is nil. Can I remove the function entirely?

andrewsykim: Don't think it hurts to have this function just to be consistent with other fields. Maybe it can be simplified to:
```
func allocateLoadBalancerNodePortsInUse(svc *api.Service) bool {
	if svc == nil {
		return false
	}

        return svc.Spec.AllocateLoadBalancerNodePorts != nil
}
```

pkg/registry/core/service/strategy.go
```diff
@@ -205,6 +210,17 @@ func topologyKeysInUse(svc *api.Service) bool {
 	return len(svc.Spec.TopologyKeys) > 0
 }
 
+// returns true if svc.Spec.ServiceIPFamily field is in use
```
andrewsykim, line 213: `svc.Spec.AllocateLoadBalancerNodePort`

pkg/registry/core/service/strategy.go
uablrek: I'll update with this as you suggest soon. There may be some case I have missed and it is in line with other similar functions.

I will just rebase and re-generate...

uablrek:
Updated after review. Unit-tests added in `rest_test.go`. The weird initialization of a *bool comes from https://stackoverflow.com/questions/28817992/how-to-set-bool-pointer-to-true-in-struct-literal :smile:

uablrek:
I have tested locally to set the value with;
```
cat > /tmp/patch.yaml <<EOF
spec:
  allocateLoadBalancerNodePorts: false
EOF
kubectl patch svc mconnect-daemonset-ipv4 --patch="$(cat /tmp/patch.yaml)"
```
and then set `nodePort=0` and the nodePort disappears as expected.

andrewsykim:
@uablrek we also have a utils package for that https://github.com/kubernetes/utils/blob/master/pointer/pointer.go#L77-L80

uablrek:
Hm, looks better. But I do not update immadiately.

pkg/registry/core/service/storage/rest_test.go
```diff
@@ -1421,6 +1421,94 @@ func TestServiceRegistryExternalService(t *testing.T) {
...
+	ctx := genericapirequest.NewDefaultContext()
+	defer featuregatetesting.SetFeatureGateDuringTest(t, utilfeature.DefaultFeatureGate, features.ServiceLBNodePortControl, true)()
+
+	storage, registry, server := NewTestREST(t, nil, singleStackIPv4)
+	defer server.Terminate(t)
+
+	svc := &api.Service{
+		ObjectMeta: metav1.ObjectMeta{Name: "alloc-false"},
+		Spec: api.ServiceSpec{
+			AllocateLoadBalancerNodePorts: &[]bool{false}[0],
```
andrewsykim, line 1434: nit: uiltpointer.BoolPtr(false)

pkg/registry/core/service/storage/rest_test.go
```diff
@@ -1421,6 +1421,94 @@ func TestServiceRegistryExternalService(t *testing.T) {
...
+	ctx := genericapirequest.NewDefaultContext()
+	defer featuregatetesting.SetFeatureGateDuringTest(t, utilfeature.DefaultFeatureGate, features.ServiceLBNodePortControl, true)()
+
+	storage, registry, server := NewTestREST(t, nil, singleStackIPv4)
+	defer server.Terminate(t)
+
+	svc := &api.Service{
+		ObjectMeta: metav1.ObjectMeta{Name: "alloc-false"},
+		Spec: api.ServiceSpec{
+			AllocateLoadBalancerNodePorts: &[]bool{true}[0],
```
andrewsykim, line 1478: nit: uiltpointer.BoolPtr(true)

pkg/registry/core/service/storage/rest_test.go
```diff
@@ -1421,6 +1421,94 @@ func TestServiceRegistryExternalService(t *testing.T) {
...
+	for i := range serviceNodePorts {
+		nodePort := serviceNodePorts[i]
+		// Release the node port at the end of the test case.
+		storage.serviceNodePorts.Release(nodePort)
+	}
+}
+
+func TestAllocateLoadBalancerNodePortsTrue(t *testing.T) {
+	ctx := genericapirequest.NewDefaultContext()
+	defer featuregatetesting.SetFeatureGateDuringTest(t, utilfeature.DefaultFeatureGate, features.ServiceLBNodePortControl, true)()
```
andrewsykim, line 1470: I think we also want a test case where the gate is false and so setting `AllocateLoadBalancerNodePorts` is a no-op.

pkg/registry/core/service/storage/rest_test.go
```diff
@@ -1421,6 +1421,94 @@ func TestServiceRegistryExternalService(t *testing.T) {
 	}
 }
 
+func TestAllocateLoadBalancerNodePortsFalse(t *testing.T) {
```
andrewsykim, line 1424: I'm wondering if this and TestAllocateLoadBalancerNodePortsTrue could be the same unit test with a test table with 4 cases:
* allocate=true, gate on
* allocate false, gate on
* allocate=true, gate off
* allocate=fase, gate off

andrewsykim: So something like:

```go
func TestAllocateLoadBalancerNodePorts(t *testing.T) {
   testcases := []struct{
        name                          string
        svc                             *api.Service
        expectNodePorts        bool
        allocateNodePortGate bool
   }{
       ....
   }

   for _, testcase := range testcases {
       t.Run(testcase.name, func(t *testing.T) {
             ....
       })     
   }
}

```

pkg/registry/core/service/strategy.go
andrewsykim: I think this comment was missed. Should be `AllocateLoadBalancerNodePort` and not `ServiceIPFamily` here.

pkg/registry/core/service/strategy.go
```diff
@@ -205,6 +210,17 @@ func topologyKeysInUse(svc *api.Service) bool {
...
+func allocateLoadBalancerNodePortsInUse(svc *api.Service) bool {
+	if svc == nil {
+		return false
+	}
+	if svc.Spec.AllocateLoadBalancerNodePorts != nil {
```
andrewsykim, line 218: nit: this can be simplified to just:
```
if svc == nil {
   return false
}

return svc.Spec.AllocateLoadBalancerNodePorts != nil
```

uablrek:
I have been forced to do other things for a while. I will update before monday.

uablrek:
@andrewsykim Updated after review.

andrewsykim (commented):
One more thing, I think we want to add some logic in [ValidateService](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L4143) to reject `allocateLoadBalancerNodePorts` being set if the Service Type is not LoadBalancer.

staging/src/k8s.io/api/core/v1/types.go
```diff
@@ -4137,6 +4137,12 @@ type ServiceSpec struct {
 	// If this is not specified or empty, no topology constraints will be applied.
 	// +optional
 	TopologyKeys []string `json:"topologyKeys,omitempty" protobuf:"bytes,16,opt,name=topologyKeys"`
+
+	// allocateLoadBalancerNodePorts defines if NodePorts shall be automatically
+	// allocated for services with type LoadBalancer.  Default is "true". It may be
+	// set to "false" if the cluster load-balancer does not rely on NodePorts.
+	// +optional
```
andrewsykim, line 4144: I think here we want to mention that this field will be dropped if the alpha feature gate is not enabled. See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#alpha-field-in-existing-api-version for more details on this.

uablrek:
@andrewsykim I have got a problem; `hack/update-generated-protobuf.sh` does no longer update any files(!?). Do you have any idea on what could have happened. Only change I am aware of is go -> 1.15 which was required. I am grateful for any hint.

uablrek:
`hack/update-generated-protobuf.sh` still does not work, I have no idea why. After reverse-engineering the scripts I found this work-around;
```
./build/run.sh hack/update-generated-protobuf-dockerized.sh k8s.io/api/core/v1
```

andrewsykim:
Sorry for the delay, I'll review this again either today or tomorrow.

uablrek:
@andrewsykim No worries. I just got the file generation working today, so there is actually no delay :smile:

But I am a bit unsure if I shall write an issue on the hack/update-generated-protobuf.sh problem. I asked on #sig-api-machinery on slack but no one else seem to have this problem.

andrewsykim:
Can you try it again with a clean git repo (`git clean -xfd`)?

uablrek:
I have done that and even re-cloned a clean repo and updated only the `type.go` files. Nothing helps. The script collects an awfully long list of packages and I wonder it is too long and becomes truncated some place. My work-around is to call the "dockerized" script with only the package I _know_ is the one I have altered.

andrewsykim:
We recently upgraded to go v1.15, are you on v1.15?

uablrek:
Yes. I even tried to downgrade to go go1.14.4. Didn't help.

andrewsykim:
Hmmm yeah not sure then -- should probably open a separate issue then

uablrek:
Yeah, I'll do that. It doesn't hurt. Any idea what sig?

andrewsykim:
Yeah probably API Machinery

uablrek:
@andrewsykim Can you please re-review?

I found the generated-file problem BTW. A script name "bash" that called the real bash was in the path. This was hard to track down since the symptoms were so misleading.

After review I will fix the remaining comments and re-generate before merge.

pkg/apis/core/validation/validation.go
```diff
@@ -4338,6 +4338,12 @@ func ValidateService(service *core.Service, allowAppProtocol bool) field.ErrorLi
 		}
 	}
 
+	if service.Spec.AllocateLoadBalancerNodePorts != nil {
```
andrewsykim, line 4341: I think we can make this into a single line and it's probably slightly more efficient to do the Type check first:
```
if service.Spec.Type != core.ServiceTypeLoadBalancer &&  service.Spec.AllocateLoadBalancerNodePorts != nil  {
     allErrs = append(allErrs, field.Forbidden(specPath.Child("allocateLoadBalancerNodePorts"), "may only be used when `type` is 'LoadBalancer'")) 
}
```

andrewsykim (commented):
LGTM -- left a really minor comment in validation logic 

/assign @thockin

pkg/apis/core/validation/validation.go
andrewsykim: Actually, I don't think order really matters

andrewsykim:
Pulled this PR locally and did some testing, seems to be working great. One thing I ran into though, when using `externalTrafficPolicy: Local` should we still allocate `healthCheckNodePort` if `allocateLoadBalancerNodePorts` is false?
```
$ kubectl get svc nginx-lb -o yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"Service","metadata":{"annotations":{},"labels":{"app":"nginx"},"name":"nginx-lb","namespace":"default"},"spec":{"allocateLoadBalancerNodePorts":false,"externalTrafficPolicy":"Local","ports":[{"port":80,"protocol":"TCP"}],"selector":{"app":"nginx"},"type":"LoadBalancer"}}
  creationTimestamp: "2020-09-03T02:30:19Z"
  labels:
    app: nginx
  name: nginx-lb
  namespace: default
  resourceVersion: "605"
  selfLink: /api/v1/namespaces/default/services/nginx-lb
  uid: 8f5d2322-9291-4815-ba3e-372649122e14
spec:
  allocateLoadBalancerNodePorts: false
  clusterIP: 10.106.10.207
  externalTrafficPolicy: Local
  healthCheckNodePort: 31321
  ports:
  - port: 80
    protocol: TCP
    targetPort: 80
  selector:
    app: nginx
  sessionAffinity: None
  type: LoadBalancer
status:
  loadBalancer: {}

```

pkg/apis/core/validation/validation.go
uablrek: I'll make it one-line but keep the order. I think like 98% of K8s cluster will never set AllocateLoadBalancerNodePorts and then it make sense to check that first. For the remaining 2% of us it's a great feature though :smile:

uablrek:
> should we still allocate healthCheckNodePort if allocateLoadBalancerNodePorts is false?

A very good question. I am unsure. For us an important reason for disabling NodePorts is security; a bunch of ports in the 30000- range should not show up on port-scans. So from that point of view, "no". But then again apps should be movable and setting allocateLoadBalancerNodePorts for all type:loadBalancer services with a web-hook (that I plan to do) shall not trash for existing apps, so from that point of view, then "yes" I guess.

But I am not the right person to tell since I have not any experience on how healthCheckNodePort are used.

pkg/apis/core/validation/validation.go
uablrek: Unless of course GCE/GKE will use it since they don't load-balance to nodeports (or so I've heard).

uablrek:
Updated after review. Final version?

andrewsykim:
> But I am not the right person to tell since I have not any experience on how healthCheckNodePort are used.

`healthCheckNodePort` opens up a serving port on kube-proxy to indicate whether a specific node has endpoints for that Service. This way an external load balancer can use this port to check which backends have Service Pods and route traffic accordingly. I think since this isn't opening up any ports into the application, it should be fine to continue to allocate it even if `allocateLoadBalancerNodePorts` is false. Also I think `healthCheckNodePort` is still generically useful for health checking even if the actual LB implementation is VIP-based and doesn't rely on NodePorts for LB traffic. So I think the current PR as-is is fine.

uablrek:
@andrewsykim @thockin Can you please make a final (I hope) re-review?

uablrek:
@andrewsykim The v1.20 release cycle has started. Can this PR be merged?

andrewsykim (commented):
This LGTM, needs a review from an api-approver (@thockin)

uablrek:
@thockin Can you please take a look at this PR for the API update before the v1.20 freeze.

thockin:
@uablrek I am sorry I missed this one - github email switched me down from "assigned" to "mention".  :(  So broken.  I will look at this in the next couple weeks - I have made a lot of free time to review the backlog.  

Sorry again.

pkg/registry/core/service/storage/rest.go
```diff
@@ -427,7 +428,8 @@ func (rs *REST) Update(ctx context.Context, name string, objInfo rest.UpdatedObj
 		releaseNodePorts(oldService, nodePortOp)
 	}
 	// Update service from any type to NodePort or LoadBalancer, should update NodePort.
-	if service.Spec.Type == api.ServiceTypeNodePort || service.Spec.Type == api.ServiceTypeLoadBalancer {
+	if service.Spec.Type == api.ServiceTypeNodePort ||
+		(service.Spec.Type == api.ServiceTypeLoadBalancer && (service.Spec.AllocateLoadBalancerNodePorts == nil || *service.Spec.AllocateLoadBalancerNodePorts)) {
```
thockin, line 432: Rather than test for nil, you can add a small block to pkg/apis/core/v1/defaults.go which checks for nil and sets it "true" if the gate allows it to be set.

Then here you can do something like:

```
allocateNodePorts := shouldAllocateNodePorts(service)
```

and

```
func shouldAllocateNodePorts(service) bool {
    if gate disabled {
        return true
    }
    return service.spec.AllocateLoadBalancerNodePorts
}
```

thockin: Update and Create, of course

pkg/apis/core/validation/validation.go
```diff
@@ -4338,6 +4338,10 @@ func ValidateService(service *core.Service, allowAppProtocol bool) field.ErrorLi
 		}
 	}
 
+	if service.Spec.AllocateLoadBalancerNodePorts != nil && service.Spec.Type != core.ServiceTypeLoadBalancer {
```
thockin, line 4341: We don't usually do gates in validation, but I think it's appropriate to add a gated check here that if the gate is enabled a value is required.  We only allow nil if the gate is disabled (see other comments on using defaults.go)

uablrek: I would like to keep `nil` as default. Main reasons;

* On beta when default is true for the feature-gw no new attribute will pop up on a `kubectl get svc`. I guess like 95% of users will not set this attribute and don't want to see it.
* A web-hook was proposed as the solution for automatic setting of this parameter. If the default is `true` the hook can't tell if this is a default (should be set to false) or explicitly set to `true` by the used (should be left as-is).

Or will the web-hook be called before "default.go"? And will values with default values also be invisible on a `kubectl get svc`?

If so, default=true is fine. But then, is there a reson for the attribute to be a pointer? I have read that this is recommended for API attributes someplace.

thockin: > no new attribute will pop up on a kubectl get svc. I guess like 95% of users will not set this attribute and don't want to see it.

We don't want to hide fields that are actually set.  Whether the user knows it or not, we are setting a value.  Most of the time, when we have done this we have regretted it.

> A web-hook was proposed as the solution for automatic setting of this parameter. If the default is true the hook can't tell if this is a default (should be set to false) or explicitly set to true by the used (should be left as-is).

This is a general problem and this is the 3rd or 4th time it has come up this month.  https://github.com/kubernetes/kubernetes/pull/95269 for another.  @apelisse @liggitt we need to either solve this generically or we decide that we are not handling it at all.

@uablrek the problem with leaving it nil is that we lose information.  At defaults.go time we know which API version was used to create the resource and we can apply version-specific default values.  In v1 we default to `true` but maybe in v2 we don't.  If we leave it nil we have no idea.  We can set this much later in REST, but before we really open that can of worms, let's discuss more.

I think it's best to do it the "normal" way (defaults.go) and let the defaulting vs webhooks debate be had in parallel. :)

Making it a pointer makes it possible to know AT ALL (defaults.go or later) whether the user specified `false` or didn't specify anything at all (Go's zero value).  `bool` fields that default to `true` (generally: any scalar type whose default != Go's zero value AND whose zero-value is in-bounds) are the one case where apointer is absolutely required.

uablrek: Ok, I'll update (and re-base). Should the value remain `nil` if the feature-gate is not set?

andrewsykim: > Should the value remain nil if the feature-gate is not set?

Yes, it should be `nil` when the feature gate is disabled. And even when the feature gate is enabled, I think `allocateLoadBalancerNodePorts` should be nil when the type is **not** LoadBalancer.

thockin: Good point Andrew.

uablrek: > it's appropriate to add a gated check here that if the gate is enabled a value is required. We only allow nil if the gate is disabled (see other comments on using defaults.go)

A non-nil is ensured in (the updated) defaults.go if the gate is enabled. I did not alter the check.

I can add a check that the value is _not_ set unless the gate is enabled, but it seems unnecessary.

uablrek:
> Rather than test for nil, you can add a small block to pkg/apis/core/v1/defaults.go which checks for nil and sets it "true" if the gate allows it to be set.

Cool! I learn something new every day :smile:

Updated after review.

The check in pkg/apis/core/validation/validation.go is not updated, please see https://github.com/kubernetes/kubernetes/pull/92744#discussion_r517168361

uablrek:
In `pkg/registry/core/service/storage/rest.go` the pointer is de-referenced without a check;
```go
func shouldAllocateNodePorts(service *api.Service) bool {
	if utilfeature.DefaultFeatureGate.Enabled(features.ServiceLBNodePortControl) {
		return *service.Spec.AllocateLoadBalancerNodePorts
	}
	return true
}
```
Is this a risk? A non-nil should be ensured in `defaults.go` but what if the gate is enabled with a bunch of services with nil values?

Will the defaults be set on read? I recall that Kal added somethig like that for dual-stack. Have I got that right? If so everything should be ok.

I will try to make a test myself, but it may take some time.

uablrek:
I tested;

* Start cluster with ServiceLBNodePortControl=false
* Create some loadBalancer services
* Re-start kube-apiserver with ServiceLBNodePortControl=true
* `kubectl get svc -o yaml ...`

It works without any nil-pointer exceptions and `allocateLoadBalancerNodePorts: true`.

Guess that answers my question above.

uablrek:
Guess this should be "v1.21"?
```go
	// owner: @andrewsykim @uablrek
	// alpha: v1.20
	//
	// Allows control if NodePorts shall be created for services with "type: LoadBalancer" by defining the spec.AllocateLoadBalancerNodePorts field (bool)
	ServiceLBNodePortControl featuregate.Feature = "ServiceLBNodePortControl"
```

uablrek:
"pull-kubernetes-verify" fails;

> /home/prow/go/src/k8s.io/kubernetes/api/openapi-spec is out of date. Please run hack/update-openapi-spec.sh

But if I run "hack/update-openapi-spec.sh" no files are updated! So I have nothing to update.

What can be wrong?

thockin:
Regarding openapi - there's a bug somewhere.  `make clean` and re-run the generator.  :(

pkg/registry/core/service/storage/rest.go
```diff
@@ -231,7 +231,8 @@ func (rs *REST) Create(ctx context.Context, obj runtime.Object, createValidation
 	nodePortOp := portallocator.StartOperation(rs.serviceNodePorts, dryrun.IsDryRun(options.DryRun))
 	defer nodePortOp.Finish()
 
-	if service.Spec.Type == api.ServiceTypeNodePort || service.Spec.Type == api.ServiceTypeLoadBalancer {
+	if service.Spec.Type == api.ServiceTypeNodePort ||
```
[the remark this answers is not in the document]
thockin: Should we reconsider the case of changing this from true to false in an update?  Maybe that *should* clear the NodePort values.

pkg/registry/core/service/strategy.go
```diff
@@ -179,6 +179,16 @@ func dropServiceDisabledFields(newSvc *api.Service, oldSvc *api.Service) {
 	if !utilfeature.DefaultFeatureGate.Enabled(features.ServiceTopology) && !topologyKeysInUse(oldSvc) {
 		newSvc.Spec.TopologyKeys = nil
 	}
+
+	// Clear AllocateLoadBalancerNodePorts if ServiceLBNodePortControl is not enabled
```
thockin, line 183: Why are there 2 blocks doing the same thing here?

pkg/registry/core/service/strategy.go
```diff
@@ -179,6 +179,11 @@ func dropServiceDisabledFields(newSvc *api.Service, oldSvc *api.Service) {
 	if !utilfeature.DefaultFeatureGate.Enabled(features.ServiceTopology) && !topologyKeysInUse(oldSvc) {
 		newSvc.Spec.TopologyKeys = nil
 	}
+
```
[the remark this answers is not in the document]
thockin: I don't see this change in here?

thockin (commented):
close, just a few outstanding.  Interested in thoughts about clearing NodePorts if someone changes this flag.  It would be simpler and maybe less surprising?

pkg/registry/core/service/strategy.go
uablrek: Ooops, I did not see that it was already there in some form :blush: Removing the second.

uablrek:
Updated after re-review;

* The duplicate tests removed from pkg/registry/core/service/strategy.go
* The pkg/registry/core/service/storage/rest_test.go (failed before) now test that nil causes an error not an exception

uablrek:
>  Interested in thoughts about clearing NodePorts if someone changes this flag. It would be simpler and maybe less surprising?

I am unsure. It is kind of a safeguard if someone is "experimenting" on a live cluster for example. Now you must go to some length to actually cause damage.

I wrote a small (83 loc) program to clear nodePorts; https://gist.github.com/uablrek/1436ef3fe30314d6e396e54a31128031

The opposite works already; If the flag is false->true NodePorts are (re)created. I think that is the most important.

andrewsykim:
>  Interested in thoughts about clearing NodePorts if someone changes this flag. It would be simpler and maybe less surprising?

There were some discussions on this during the KEP review as well (see https://github.com/kubernetes/enhancements/pull/1514#issuecomment-643628314). I think the biggest concern was that if a user accidentally turned off node ports and wanted to turn it back on, it would be disruptive since new node ports would have to be allocated. Requiring users to explicilty set node ports to 0 to disable port forwarding seemed safer. Maybe we're being too cautious though.

andrewsykim:
@uablrek can you rebase please?

pkg/apis/core/types.go
```diff
@@ -3696,6 +3696,15 @@ type ServiceSpec struct {
 	// If this is not specified or empty, no topology constraints will be applied.
 	// +optional
 	TopologyKeys []string
+
+	// allocateLoadBalancerNodePorts defines if NodePorts shall be automatically
```
andrewsykim, line 3700: nit: s/shall/will

pkg/apis/core/validation/validation_test.go
```diff
@@ -11055,6 +11055,13 @@ func TestValidateServiceCreate(t *testing.T) {
 			},
 			numErrs: 1,
 		},
+		{
+			name: "Use AllocateLoadBalancerNodePorts when type is not loadBalancer",
```
andrewsykim, line 11059: nit: s/loadBalancer/LoadBalancer

pkg/apis/core/validation/validation_test.go
```diff
@@ -13447,6 +13454,13 @@ func TestValidateServiceUpdate(t *testing.T) {
 			appProtocolEnabled: true,
 			numErrs:            1,
 		},
+		{
+			name: "Set AllocateLoadBalancerNodePorts when type is not loadBalancer",
```
andrewsykim, line 13458: nit: s/loadBalancer/LoadBalancer

staging/src/k8s.io/api/core/v1/types.go
```diff
@@ -4223,6 +4223,15 @@ type ServiceSpec struct {
 	// wiped when updating a service to type ExternalName.
 	// +optional
 	IPFamilyPolicy *IPFamilyPolicyType `json:"ipFamilyPolicy,omitempty" protobuf:"bytes,17,opt,name=ipFamilyPolicy,casttype=IPFamilyPolicyType"`
+
+	// allocateLoadBalancerNodePorts defines if NodePorts shall be automatically
```
andrewsykim, line 4227: nit: s/shall/will

andrewsykim:
> If you rebase you will find a new function dropTypeDependentFields which can be updated to wipe this field when changing type

@uablrek I think this comment from Tim still needs to be addressed. Just FYI code freeze is EOD Nov 12

pkg/registry/core/service/strategy.go
```diff
@@ -185,6 +185,11 @@ func dropServiceDisabledFields(newSvc *api.Service, oldSvc *api.Service) {
 			ing.IPMode = nil
 		}
 	}
+
+	// Clear AllocateLoadBalancerNodePorts if ServiceLBNodePortControl if not enabled
+	if !utilfeature.DefaultFeatureGate.Enabled(features.ServiceLBNodePortControl) {
```
andrewsykim, line 190: This check used to check for `!allocateLoadBalancerNodePortsInUse` but it seems like we dropped it, was that intentional?

pkg/registry/core/service/strategy.go
```diff
@@ -376,6 +389,11 @@ func dropTypeDependentFields(newSvc *api.Service, oldSvc *api.Service) {
 		newSvc.Spec.HealthCheckNodePort = 0
 	}
 
+	// AllocateLoadBalancerNodePorts may only be set for type LoadBalancer
+	if newSvc.Spec.Type != api.ServiceTypeLoadBalancer {
```
andrewsykim, line 393: I think here we also need to check whether `allocatedLoadBalancerNodePorts` was trying to be set by `oldSvc`, otherwise this wouldn't fail validation checks.

andrewsykim: So something like

```
// If a user is switching to a type that doesn't need allocatedLoadBalancerNodePorts AND they did not change
// this field, it is safe to drop it. 
if oldSvc.Spec.Type == api.ServiceTypeLoadBalancer && newSvc.Spec.Type != api.ServiceTypeLoadBalancer {
   if oldSvc.Spec.AllocateLoadBalancerNodePorts == newSvc.Spec.AllocatedLoadBalancerNodePorts {
       newSvc.Spec.AllocateLoadBalancerNodePorts = nil
   }
}
```

andrewsykim: Can we also add a test case for this in TestDropTypeDependentFields?

pkg/registry/core/service/strategy.go
uablrek: Yes, since the allocateLoadBalancerNodePorts must be nil if the feature-gate is not set. At least that's how I have understood it.

pkg/registry/core/service/strategy.go
uablrek: Why all the extra checks? If type != LoadBalancer the AllocateLoadBalancerNodePorts _must_ be nil no matter what.

uablrek: I also find the conditions hard to follow. If the type is changed from LoadBalancer but newSvc sets AllocateLoadBalancerNodePorts to true or false it will remain != nil?

pkg/registry/core/service/strategy.go
andrewsykim: > allocateLoadBalancerNodePorts must be nil if the feature-gate is not set

This and if an existing resource has the field already set. So this check should still check if `oldSvc` was using this field

thockin: Yeah this needs to preserve the field if it was already set in oldSvc.  The rationale is that a user might upgrade to 1.21, use the field, temporarily roll back to 1.20 (where it is gated) - the field value should be preserved so that when they roll forward it is still present.

If this is the only issue I will approve but we need a followup to fix this

pkg/registry/core/service/strategy.go
thockin: Yes, we only want to auto-clear fields if we don't need it AND the user is not trying to change it.  If they are trying to change it then we should let validation fail.

thockin (commented):
The 2 comments I have can be a followup.

/lgtm
/approve

thockin:
uggh, needs rebase but is otherwise approved

kikisdeliveryservice:
As per: https://github.com/kubernetes/enhancements/issues/1864#issuecomment-726542281

Please get this rebased and re-LGTM ASAP

Thanks!
Kirsten

uablrek:
@andrewsykim @thockin Thanks for your guidance and patiance in my first API update :+1:  I hope this will be the last update in this PR. I will immediate create a new PR for the followup's when it's merged.

I have rebased, but not updated the review items since;

> The 2 comments I have can be a followup.

I have (temporary) removed the unused function `allocateLoadBalancerNodePortsInUse()` from `pkg/registry/core/service/strategy.go:210` since it broke the static tests. It will be re-inserted in the followup. Beside that and the rebase conflicts the code is not altered. But if possible please check so I have not messed up the conflicts. I have checked myself of course but...

andrewsykim (commented):
/lgtm

Summarizing some of the follow-up items in case they were lost in review:
1. CreateStrategy shouldn't drop `spec.allocateLoadBalancerNodePorts` if an exisitng Service was using it
2. dropTypeDependentFields should check if `spec.allocateLoadBalancerNodePorts` is changing when the service type also changes
3. unit tests for dropTypeDependentFields when type changes from LoadBalancer to something else.

pkg/registry/core/service/strategy.go
thockin: Re-skim: This will need a followup, still.

pkg/registry/core/service/strategy.go
thockin: as above, followup needed

thockin (commented):
Thanks!

/lgtm
/approve

PR #92745, opened 2020-07-02
Use Pause image for DisruptionController tests

hakman:
**What type of PR is this?**

/kind failing-test

DisruptionController e2e tests are failing on ARM64 due to an issue with the EchoServer image for this architecture. I discussed this with @claudiubelu and we agreed that a new EchoServer build will have to be done to fix it.
https://testgrid.k8s.io/kops-misc#kops-aws-arm64-ci

After investigating a bit, I discovered that these tests don't require EchoServer, but any image that keeps the pod running. Even the container name is `busybox`.

This PR replaces the EchoServer image used in DisruptionController tests with BusyBox, to unblock testing on ARM64.

n/a

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/apps/disruption.go
+++ b/test/e2e/apps/disruption.go
@@ -461,8 +461,8 @@ func createPodsOrDie(cs kubernetes.Interface, ns string, n int) {
 			Spec: v1.PodSpec{
 				Containers: []v1.Container{
 					{
-						Name:  "busybox",
-						Image: imageutils.GetE2EImage(imageutils.EchoServer),
+						Name:  "donothing",
+						Image: imageutils.GetPauseImageName(),
 					},
 				},
 				RestartPolicy: v1.RestartPolicyAlways,
@@ -506,8 +506,8 @@ func waitForPodsOrDie(cs kubernetes.Interface, ns string, n int) {
 
 func createReplicaSetOrDie(cs kubernetes.Interface, ns string, size int32, exclusive bool) {
 	container := v1.Container{
-		Name:  "busybox",
-		Image: imageutils.GetE2EImage(imageutils.EchoServer),
+		Name:  "donothing",
+		Image: imageutils.GetPauseImageName(),
 	}
 	if exclusive {
 		container.Ports = []v1.ContainerPort{


test/e2e/apps/disruption.go
```diff
@@ -459,8 +459,9 @@ func createPodsOrDie(cs kubernetes.Interface, ns string, n int) {
 			Spec: v1.PodSpec{
 				Containers: []v1.Container{
 					{
-						Name:  "busybox",
-						Image: imageutils.GetE2EImage(imageutils.EchoServer),
+						Name:    "busybox",
+						Image:   imageutils.GetE2EImage(imageutils.BusyBox),
+						Command: []string{"sleep", "100000"},
```
spiffxp, line 464: If all we need is a running container, perhaps we could use pause?
```suggestion
						Name:    "donothing",
						Image:   imageutils.GetPauseImageName(),
```

test/e2e/apps/disruption.go
```diff
@@ -504,8 +505,9 @@ func waitForPodsOrDie(cs kubernetes.Interface, ns string, n int) {
 
 func createReplicaSetOrDie(cs kubernetes.Interface, ns string, size int32, exclusive bool) {
 	container := v1.Container{
-		Name:  "busybox",
-		Image: imageutils.GetE2EImage(imageutils.EchoServer),
+		Name:    "busybox",
+		Image:   imageutils.GetE2EImage(imageutils.BusyBox),
+		Command: []string{"sleep", "100000"},
```
spiffxp, line 510: Same here
```suggestion
		Name:    "donothing",
		Image:   imageutils.GetPauseImageName(),
```

spiffxp (commented):
Some suggestions

test/e2e/apps/disruption.go
hakman: Good point. Thanks for the suggestion. Let's see. :)

hakman: Tests look good with Pause.

hakman:
Using the Pause image instead of EchoServer works nicely also. Thanks for the suggestion @spiffxp.
Test failures were unrelated.

hakman:
Thanks @spiffxp @BenTheElder !

BenTheElder:
I suspect this caused https://github.com/kubernetes/kubernetes/issues/92977

There's been a sudden large uptick in flakes for `DisruptionController should block an eviction until the PDB is updated to allow it` in the past day.

PR #92752, opened 2020-07-02
Cut off the cost to run filter plugins when no victim pods are found

chendave:
If no potential victims could be found, there is no need to evaluate the node
again, since its state didn't change.

It's safe to return and thus prevent scheduling from running the filter plugins
again.

**NOTE**:
A node that is filtered out by filter plugins could pass the filter plugins if
there is a change on that node, i.e. pods termination on that node.

Previously, this could be either caught by the normal `schedule` or `preempt` (pods
are terminated when the preemption logic tries to find the nodes and re-evaluate
the filter plugins.)

Actually, this shouldn't be taken care by the preemption, considering the routine
of `schedule` is always running when the interval is "zero", let `schedule`
take care of it will release `preempt` from something irrelevant with the `preemption`.

Due to above reason, couple of testcase as well as the logic of checking the existence
of victim pods are removed as it will never happen after the change.

**What type of PR is this?**
>
/kind feature

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
+++ b/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
@@ -315,13 +315,6 @@ func pickOneNodeForPreemption(nodesToVictims map[string]*extenderv1.Victims) str
 	var minNodes1 []string
 	lenNodes1 := 0
 	for node, victims := range nodesToVictims {
-		if len(victims.Pods) == 0 {
-			// We found a node that doesn't need any preemption. Return it!
-			// This should happen rarely when one or more pods are terminated between
-			// the time that scheduler tries to schedule the pod and the time that
-			// preemption logic tries to find nodes for preemption.
-			return node
-		}
 		numPDBViolatingPods := victims.NumPDBViolations
 		if numPDBViolatingPods < minNumPDBViolatingPods {
 			minNumPDBViolatingPods = numPDBViolatingPods
@@ -488,6 +481,12 @@ func selectVictimsOnNode(
 			}
 		}
 	}
+
+	// No potential victims are found, and so we don't need to evaluate the node again since its state didn't change.
+	if len(potentialVictims) == 0 {
+		return nil, 0, false
+	}
+
 	// If the new pod does not fit after removing all the lower priority pods,
 	// we are almost done and this node is not suitable for preemption. The only
 	// condition that we could check is if the "pod" is failing to schedule due to
--- a/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
+++ b/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
@@ -323,7 +323,7 @@ func TestSelectNodesForPreemption(t *testing.T) {
 			expectedNumFilterCalled: 4,
 		},
 		{
-			name: "a pod that would fit on the machines, but other pods running are higher priority",
+			name: "a pod that would fit on the machines, but other pods running are higher priority, no preemption would happen",
 			registerPlugins: []st.RegisterPluginFunc{
 				st.RegisterPluginAsExtensions(noderesources.FitName, noderesources.NewFit, "Filter", "PreFilter"),
 			},
@@ -334,7 +334,7 @@ func TestSelectNodesForPreemption(t *testing.T) {
 				st.MakePod().Name("p2").UID("p2").Node("node2").Priority(midPriority).Req(largeRes).Obj(),
 			},
 			expected:                map[string]*extenderv1.Victims{},
-			expectedNumFilterCalled: 2,
+			expectedNumFilterCalled: 0,
 		},
 		{
 			name: "medium priority pod is preempted, but lower priority one stays as it is small",
@@ -380,7 +380,7 @@ func TestSelectNodesForPreemption(t *testing.T) {
 					},
 				},
 			},
-			expectedNumFilterCalled: 5,
+			expectedNumFilterCalled: 4,
 		},
 		{
 			name: "mixed priority pods are preempted, pick later StartTime one when priorities are equal",
@@ -404,7 +404,7 @@ func TestSelectNodesForPreemption(t *testing.T) {
 					},
 				},
 			},
-			expectedNumFilterCalled: 5,
+			expectedNumFilterCalled: 4, // no preemption would happen on node2 and no filter call is counted.
 		},
 		{
 			name: "pod with anti-affinity is preempted",
@@ -428,9 +428,8 @@ func TestSelectNodesForPreemption(t *testing.T) {
 							PodAntiAffinityExists("foo", "hostname", st.PodAntiAffinityWithRequiredReq).Obj(),
 					},
 				},
-				"node2": {},
 			},
-			expectedNumFilterCalled: 4,
+			expectedNumFilterCalled: 3, // no preemption would happen on node2 and no filter call is counted.
 		},
 		{
 			name: "preemption to resolve pod topology spread filter failure",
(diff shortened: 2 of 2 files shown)


chendave:
/hold
need to update testcase as well.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -482,6 +482,12 @@ func selectVictimsOnNode(
 			}
 		}
 	}
+
+	// No potential victims are found, so that no room will be released for the "pod" to be scheduled.
```
ahg-g, line 486:
```suggestion
	// No potential victims are found, and so we don't need to evaluate the node again since its state didn't change.
```

ahg-g (commented):
thanks, one nit.

chendave:
@ahg-g , thanks for your review and comments, I found some corner case (manually deleted the pods) which might turn this a invalid PR.

I need to reconsider the validity of the PR, then either reopen or close this.

Huang-Wei (commented):
Thanks @chendave. It's a neat enhancement. We should have caught it earlier :)

Huang-Wei:
> I need to reconsider the validity of the PR, then either reopen or close this.

I do see some tests failed - some are as expected as the "filterNumCalled" needs to be adjusted; while some need more investigation.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
chendave: Done

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -1170,21 +1145,22 @@ func TestPreempt(t *testing.T) {
...
-			name: "One scheduler extender allows only machine1, but it is not interested in given pod, otherwise node1 would have been chosen",
+			name: "One scheduler extender allows only machine1, but it is not interested in given pod, otherwise machine1 would have been chosen",
 			pod:  st.MakePod().Name("p").UID("p").Priority(highPriority).Req(veryLargeRes).PreemptionPolicy(v1.PreemptLowerPriority).Obj(),
 			pods: []*v1.Pod{
-				st.MakePod().Name("p1.1").UID("p1.1").Node("node1").Priority(midPriority).Req(smallRes).Obj(),
-				st.MakePod().Name("p1.2").UID("p1.2").Node("node1").Priority(lowPriority).Req(smallRes).Obj(),
-				st.MakePod().Name("p2.1").UID("p2.1").Node("node2").Priority(midPriority).Req(largeRes).Obj(),
+				st.MakePod().Name("p1.1").UID("p1.1").Node("machine1").Priority(midPriority).Req(smallRes).Obj(),
+				st.MakePod().Name("p1.2").UID("p1.2").Node("machine1").Priority(lowPriority).Req(smallRes).Obj(),
+				st.MakePod().Name("p2.1").UID("p2.1").Node("machine2").Priority(midPriority).Req(largeRes).Obj(),
```
chendave, line 1153: I reverted the "node" to "machine" cause the extender used here is still using "machine" instead of "node", this will make the this testcase invalid without the change here.

https://github.com/kubernetes/kubernetes/blob/865cbf0bdfb4ae0d6cb5029ffacea127cd42e89c/pkg/scheduler/testing/fake_extender.go#L62-L67

As there are still other occurrences of "machine" in other testcase. I will propose another PR to address it instead of mess up this PR.

chendave: cc @Huang-Wei

chendave: see: https://github.com/kubernetes/kubernetes/pull/92819

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -404,7 +404,7 @@ func TestSelectNodesForPreemption(t *testing.T) {
 					},
 				},
 			},
-			expectedNumFilterCalled: 5,
+			expectedNumFilterCalled: 4, //no preemption would happen on node2 and no filter call is counted.
```
Huang-Wei, line 407:
```suggestion
			expectedNumFilterCalled: 4, // no preemption would happen on node2 and no filter call is counted.
```

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -428,9 +428,8 @@ func TestSelectNodesForPreemption(t *testing.T) {
 							PodAntiAffinityExists("foo", "hostname", st.PodAntiAffinityWithRequiredReq).Obj(),
 					},
 				},
-				"node2": {},
 			},
-			expectedNumFilterCalled: 4,
+			expectedNumFilterCalled: 3, //no preemption would happen on node2 and no filter call is counted.
```
Huang-Wei, line 432:
```suggestion
			expectedNumFilterCalled: 3, // no preemption would happen on node2 and no filter call is counted.
```

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -1170,21 +1145,22 @@ func TestPreempt(t *testing.T) {
...
-			nodeNames: []string{"node1", "node2", "node3"},
+			nodeNames: []string{"machine1", "machine2"},
 			extenders: []*st.FakeExtender{
 				{Predicates: []st.FitPredicate{st.Machine1PredicateExtender}, UnInterested: true},
 				{Predicates: []st.FitPredicate{st.TruePredicateExtender}},
 			},
 			registerPlugin: st.RegisterPluginAsExtensions(noderesources.FitName, noderesources.NewFit, "Filter", "PreFilter"),
-			expectedNode:   "node3",
-			expectedPods:   []string{},
+			//sum of priorities of all victims on machine1 is larger than machine2, machine2 is chosen.
```
Huang-Wei, line 1161:
```suggestion
			// sum of priorities of all victims on machine1 is larger than machine2, machine2 is chosen.
```

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -315,13 +315,6 @@ func pickOneNodeForPreemption(nodesToVictims map[string]*extenderv1.Victims) str
 	var minNodes1 []string
 	lenNodes1 := 0
 	for node, victims := range nodesToVictims {
-		if len(victims.Pods) == 0 {
-			// We found a node that doesn't need any preemption. Return it!
-			// This should happen rarely when one or more pods are terminated between
-			// the time that scheduler tries to schedule the pod and the time that
-			// preemption logic tries to find nodes for preemption.
-			return node
-		}
```
Huang-Wei, line 324: Why remove this?

chendave: cause we cannnot got this situation anymore, the case as the comment states would be handled by the normal scheduling process instead of the preemption here.

Huang-Wei: SG. I think this code was introduced long time ago, by then it was not guaranteed we have a consistent cluster view in one scheduling cycle.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
chendave: Done

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
chendave: Done

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
chendave: Done

Huang-Wei:
/approve
leave /lgtm to @ahg-g .

ahg-g:
/lgtm

Thanks!

PR #92765, opened 2020-07-02
Av/aggregator cert reload review

p0lyn0mial:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```

```docs

```


staging/src/k8s.io/client-go/transport/cache.go
```diff
@@ -128,10 +129,23 @@ func tlsConfigKey(c *Config) (tlsCacheKey, error) {
 	if err := loadTLSFiles(c); err != nil {
 		return tlsCacheKey{}, err
 	}
+
+	// since golang forbids comparing functions we should take the address of the certificate instead
```
deads2k, line 133: What if we were to change the entire interface to  require something with a name and we set the name to "dynamic:certfile:keyfile" or more generally, "scheme:something".  It would be more cumbersome and larger, but it would be very explicit.

p0lyn0mial: somehting like `TLSConfig.NewGetCert func() (*tls.Certificate, identity string, error)` ?

PR #92767, opened 2020-07-02
Fix ListZonesInRegion() after client BasePath change

jkaniuk:
**What type of PR is this?**
/kind bug

This path fixes region Regex for listing zones.
Compute client BasePath changed to compute.googleapis.com, but resource URI were left as www.googleapis.com

```release-note
NONE
```

```docs

```
 
/cc @jingxu97
/cc @prameshj

### DIFF
--- a/staging/src/k8s.io/legacy-cloud-providers/gce/gce_zones.go
+++ b/staging/src/k8s.io/legacy-cloud-providers/gce/gce_zones.go
@@ -20,6 +20,7 @@ package gce
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	compute "google.golang.org/api/compute/v1"
@@ -79,7 +80,7 @@ func (g *Cloud) ListZonesInRegion(region string) ([]*compute.Zone, error) {
 	defer cancel()
 
 	mc := newZonesMetricContext("list", region)
-	list, err := g.c.Zones().List(ctx, filter.Regexp("region", g.getRegionLink(region)))
+	list, err := g.c.Zones().List(ctx, filter.Regexp("region", fmt.Sprintf(".*/regions/%s", region)))
 	if err != nil {
 		return nil, mc.Observe(err)
 	}


prameshj:
The other usages are in:

https://github.com/kubernetes/kubernetes/blob/47a9952337dada2cd57f3b2b3aeccac500277318/staging/src/k8s.io/legacy-cloud-providers/gce/gce_instances.go#L82

https://github.com/kubernetes/kubernetes/blob/47a9952337dada2cd57f3b2b3aeccac500277318/staging/src/k8s.io/legacy-cloud-providers/gce/gce_loadbalancer_external.go#L635

https://github.com/kubernetes/kubernetes/blob/47a9952337dada2cd57f3b2b3aeccac500277318/staging/src/k8s.io/legacy-cloud-providers/gce/gce_loadbalancer_internal.go#L877

None of those are using BasePath to create a filter, they are used to create the respective resources, so these seem to be ok.

jingxu97:
> g.service.BasePath

Although this PR could workaround the issue, it is still not clear that whether BasePath is changed. And also I thought filter actually only works if exact region link is passed since regex is actually eq in this case. https://github.com/GoogleCloudPlatform/k8s-cloud-provider/blob/e05e318a0ecdedc5abb4517092c32fc7a48907fa/pkg/cloud/filter/filter.go#L201,

I also found another function trying to get zones with given region.
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/gce/gce.go#L868

It mentioned that filter is not working, but that's three years ago... @bowei

staging/src/k8s.io/legacy-cloud-providers/gce/gce_zones.go
```diff
@@ -79,7 +80,7 @@ func (g *Cloud) ListZonesInRegion(region string) ([]*compute.Zone, error) {
 	defer cancel()
 
 	mc := newZonesMetricContext("list", region)
-	list, err := g.c.Zones().List(ctx, filter.Regexp("region", g.getRegionLink(region)))
+	list, err := g.c.Zones().List(ctx, filter.Regexp("region", fmt.Sprintf(".*/regions/%s", region)))
```
alculquicondor, line 83: why not fix the function `getRegionLink` instead. Also a comment is worth.

jingxu97: we cannot use regexp in getRegionLink, which is also used by other functions.

alculquicondor:
/close

Fixed on #92883

PR #91976, opened 2020-06-10
Implement API changes necessary for recovering from expansion failure

gnufied:
This PR implements necessary API changes for https://github.com/kubernetes/enhancements/issues/1790

It contains following changes:

1. Relaxes api validation to allow shrinking of user requested capacity in PVC. This does not mean a shrink is ever performed. User is only allowed to shrink upto `pvc.status.capacity`
2. Adds a new field in pvc called `pvc.Spec.AllocatedResources` for quota tracking.

```release-note
Allow users to reduce requested PVC size to cancel previously issued volume expansion request
```

PR #92770, opened 2020-07-02
Allow volume expansion to be cancelled 

gnufied:
This includes changes from https://github.com/kubernetes/kubernetes/pull/91976

And it includes changes to allow user to cancel volume expansion altogether if user reduces PVC size to original requested value.

xref kubernetes/enhancements#1790

```release-note
Allow users to cancel volume expansion
```


gnufied:
/sig storage

cc @kubernetes/sig-storage-pr-reviews

pkg/volume/util/operationexecutor/operation_generator.go
```diff
@@ -1428,6 +1428,11 @@ func (og *operationGenerator) GenerateExpandVolumeFunc(
 			}
 
 			klog.Infof("ExpandVolume.UpdatePV succeeded for volume %s", util.GetPersistentVolumeClaimQualifiedName(pvc))
+		} else {
+			cancelled, err := cancelControlPlaneExpansion(pvc, pv, og.kubeClient)
+			if cancelled {
+				return err, err
```
xing-yang, line 1434: Why does it return the same "err" twice here?

pkg/volume/util/operationexecutor/operation_generator.go
```diff
@@ -1681,3 +1691,56 @@ func isDeviceOpened(deviceToDetach AttachedVolume, hostUtil hostutil.HostUtils)
...
+
+func cancelControlPlaneExpansion(pvc *v1.PersistentVolumeClaim, pv *v1.PersistentVolume, client clientset.Interface) (bool, error) {
+	if !utilfeature.DefaultFeatureGate.Enabled(features.RecoverVolumeExpansionFailure) {
+		return false, nil
+	}
+
+	newSize := pvc.Spec.Resources.Requests[v1.ResourceStorage]
+	statusSize := pvc.Status.Capacity[v1.ResourceStorage]
+	pvSize := pv.Spec.Capacity[v1.ResourceStorage]
+	if statusSize.Cmp(newSize) >= 0 && pvSize.Cmp(newSize) >= 0 {
```
xing-yang, line 1703: If newSize > statusSize, it will remain this way?  So after canceling is complete, we'll still see the requested size of PVC is greater than the actual PVC size shown by "kubectl describe pvc"?  The pvSize and statusSize will be the same at the end, right?  Do we need to compare the pvSize and statusSize here?  It seems that there could be different combinations of newSize, statusSize, and pvSize here.  It's good to add comments to explain the possible combinations and expected outcome.

pkg/volume/util/operationexecutor/operation_generator.go
```diff
@@ -1681,3 +1691,56 @@ func isDeviceOpened(deviceToDetach AttachedVolume, hostUtil hostutil.HostUtils)
...
+		klog.Infof("PVC %s already is of requested size", util.GetPersistentVolumeClaimQualifiedName(pvc))
```
xing-yang, line 1704: If newSize > statusSize, then this message is not accurate.

pkg/volume/util/operationexecutor/operation_generator.go
```diff
@@ -1681,3 +1691,56 @@ func isDeviceOpened(deviceToDetach AttachedVolume, hostUtil hostutil.HostUtils)
 	}
 	return deviceOpened, nil
 }
+
+func cancelControlPlaneExpansion(pvc *v1.PersistentVolumeClaim, pv *v1.PersistentVolume, client clientset.Interface) (bool, error) {
```
xing-yang, line 1695: Add comment to this func that bool means we want to cancel.

pkg/volume/util/operationexecutor/operation_generator.go
```diff
@@ -1681,3 +1691,56 @@ func isDeviceOpened(deviceToDetach AttachedVolume, hostUtil hostutil.HostUtils)
...
+		if err != nil {
+			detailedErr := fmt.Errorf("Error marking pvc %s as resized : %v", util.GetPersistentVolumeClaimQualifiedName(pvc), err)
+			return true, detailedErr
+		}
+		return true, nil
+	}
+	return false, nil
+}
+
+func cancelNodeExpansion(pvc *v1.PersistentVolumeClaim, pv *v1.PersistentVolume, client clientset.Interface) (bool, error) {
```
xing-yang, line 1715: Add comment to the func

pkg/volume/util/operationexecutor/operation_generator.go
```diff
@@ -1681,3 +1691,56 @@ func isDeviceOpened(deviceToDetach AttachedVolume, hostUtil hostutil.HostUtils)
...
+	// Just to be safe we should only cancel node expansion of volumes which already have
+	// filesystem resize pending condition
+	if !hasFileSystemResizePendingCondition(pvc) {
+		return false, nil
+	}
+
+	pvSpecCap := pv.Spec.Capacity[v1.ResourceStorage]
+	// if we are here that means PVC status already is of size greater or equal to
+	// PV capacity and hence no node expansion is necessary.
+	err := util.MarkFSResizeFinished(pvc, pvSpecCap, client)
```
xing-yang, line 1729: Is it possible for PVC status to be greater than PV capacity?  I thought PVC status is updated last.  If PVC status is bigger than PV capacity, which one is the real size?

pkg/volume/util/operationexecutor/operation_generator_test.go
```diff
@@ -108,6 +114,149 @@ func TestOperationGenerator_GenerateUnmapVolumeFunc_PluginName(t *testing.T) {
...
+			expectedValue:  false,
+			pv:             getFakePersistentVolume("7G"),
+			pvc:            getFakePersistentVolumeClaim("10G", "7G"),
+		},
+		{
+			name:           "feature:enabled, pv.size is smaller than pvc.size",
+			recoverFeature: true,
+			expectedValue:  false,
+			pv:             getFakePersistentVolume("7G"),
+			pvc:            getFakePersistentVolumeClaim("10G", "20G"),
```
xing-yang, line 151: I wonder when can this happen where the PV capacity is 7G and PVC status is already 20G?

pkg/volume/util/operationexecutor/operation_generator.go
gnufied: if `newSize > statusSize` then there is no reason to mark volume expansion as finished. In that case volume expansion will still be tried.

pkg/volume/util/operationexecutor/operation_generator.go
gnufied: it returns that twice to satisfy contract of all generator functions.

pkg/volume/util/operationexecutor/operation_generator.go
gnufied: In general `pvc.Status.Capacity` should not be greater than `pv.Spec.Capacity` but from volume expansion code flow perspective if that happens then there is no volume expansion needed on the node and expansion operation can be marked as finished.

xing-yang:
CI picked up a golint error:
```
Errors from golint:
pkg/volume/util/operationexecutor/operation_generator.go:1676:10: if block ends with a return statement, so drop this else and outdent its block
```

pkg/volume/util/operationexecutor/operation_generator_test.go
xing-yang: For the completeness (even though it is unlikely to happen), can you also add a test case when PVC requested size < PV capacity < PVC status

pkg/volume/util/operationexecutor/operation_generator.go
gnufied: sure, I will add a comment but if `newSize > statusSize` then volume expansion *should not* be cancelled and it should retried/proceed as before.

gnufied:
@xing-yang thank you for review and thanks for pointing out the CI error. I made a fix.

/retest

pkg/volume/util/operationexecutor/operation_generator.go
gnufied: done.

pkg/volume/util/operationexecutor/operation_generator.go
gnufied: Added

pkg/volume/util/operationexecutor/operation_generator_test.go
gnufied: If pvc requested < pv capacity < pvc status then no volume expansion should be necessary and hence volume expansion operation should be marked as completed.

gnufied: Also added tests for scenario you mentioned though.

pkg/volume/util/operationexecutor/operation_generator.go
```diff
@@ -1443,6 +1443,11 @@ func (og *operationGenerator) GenerateExpandVolumeFunc(
 			}
 
 			klog.Infof("ExpandVolume.UpdatePV succeeded for volume %s", util.GetPersistentVolumeClaimQualifiedName(pvc))
+		} else {
+			cancelled, err := cancelControlPlaneExpansion(pvc, pv, og.kubeClient)
+			if cancelled {
+				return err, err
```
saad-ali, line 1449: If cancel succeeded why return an error? Shouldn't it be the opposite?

pkg/volume/util/operationexecutor/operation_generator.go
saad-ali: These two should not be identical. One is supposed to be a simple error and the other is a detailed error.

Follow the example of:

https://github.com/kubernetes/kubernetes/blob/a26e5881d8aa38aaa0a56c78050fd4ea14611c9f/pkg/volume/util/operationexecutor/operation_generator.go#L332

pkg/volume/util/operationexecutor/operation_generator.go
gnufied: there are three scenarios that this covers:

1. if expansion cancel is successful - `err` will be automatically nil and hence we will be returning nil errors.
2. if cancelled is true but has errors that means - there might be an error updating PV/PVC and hence just api-server update should be retried and hence we have to return errors.
3. if cancelled is false then there was no attempt made to cancel anything and hence volume expansion should proceed as usual.

pkg/volume/util/operationexecutor/operation_generator.go
gnufied: sure, but it is making less sense to return two distinct errors from these function tbh. Unless caller explicitly logs it, by default the detailedError just ends up as being used for recording metrics - https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/util/types/types.go#L45

It is more useful in cases where you do want two distinct messages - shorter version goes to pod/pvc event and longer version goes to logs.

xing-yang:
My comments are addressed.  Looks good to me.

pkg/volume/util/operationexecutor/operation_generator.go
saad-ali: As far as I can tell the expand controller specifies an "event recorder", which if I understand correctly means that events will be generated automatically on errors. The simple error message is used for events, the detailed is used for logs.

It sounds like you are saying that there is no need to make the events messages cleaner or the log messages more detailed (I think attach and other functions will prefix/suffix method information for example)?

pkg/volume/util/operationexecutor/operation_generator.go
saad-ali: Ack.

pkg/volume/util/operationexecutor/operation_generator.go
gnufied: > It sounds like you are saying that there is no need to make the events messages cleaner or the log messages more detailed (I think attach and other functions will prefix/suffix method information for example)?

No! of course not, I am not saying that. What I am trying to say is - `detailedError` is actually not even used for logging anywhere by default. So, there is no point in constructing a detailedError in most generator functions actually, because we are not doing anything special with it.

saad-ali: Ack, we should clean that up - it makes no sense to force return something that is rarely used.

saad-ali (approved):
/lgtm
/approve

No further concerns with this PR.

staging/src/k8s.io/api/core/v1/types.go
```diff
@@ -450,6 +450,13 @@ type PersistentVolumeClaimSpec struct {
 	// More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes#resources
 	// +optional
 	Resources ResourceRequirements `json:"resources,omitempty" protobuf:"bytes,2,opt,name=resources"`
+	// AllocatedResources represents the resources that was requested to satisfy user
+	// request. This field is not meant to be set by the user but is used for tracking
```
deads2k, line 454: what happens if a user does set it?  Is there protection that prevents a user from setting this (I didn't see it right off).  My memory is that users can set fields on PVC.spec.

pkg/apis/core/validation/validation.go
```diff
@@ -1966,18 +1966,53 @@ func ValidatePersistentVolumeClaimUpdate(newPvc, oldPvc *core.PersistentVolumeCl
...
 			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec"), fmt.Sprintf("spec is immutable after creation except resources.requests for bound claims\n%v", specDiff)))
 		}
+
+		var oldAllocatedSize resource.Quantity
+		var newAllocatedSize resource.Quantity
+
+		if oldPvc.Spec.AllocatedResources != nil {
+			oldAllocatedSize = oldPvc.Spec.AllocatedResources.Requests[core.ResourceStorage]
+		}
+		if newPvc.Spec.AllocatedResources != nil {
```
deads2k, line 1994: is the old valid value in this map `core.ResourceStorage`? if so, can we enforce that via validation to prevent accidentally storing more data than we mean to?

staging/src/k8s.io/api/core/v1/types.go
```diff
@@ -450,6 +450,13 @@ type PersistentVolumeClaimSpec struct {
...
+	// volume quota.
+	// This field is alpha-level and is only honored by servers that enable the RecoverVolumeExpansionFailure feature.
+	//
+	// +optional
+	AllocatedResources *ResourceRequirements `json:"allocatedResources,omitempty" protobuf:"bytes,8,opt,name=allocatedResources"`
```
deads2k, line 459: If I've understood the usage, this a value on spec which is derived on updates to spec to track "max value ever requested".  A user could request a different volume size, but it wouldn't impact his quota and I'm not sure it actually impacts the reality of the size allocated for him.

One thing I don't understand is that why this information would not be status information provided when the expansion actually happened.

deads2k:
Based on my questions while reading this, I think I lack the domain expertise to make a good judgment about the API in question.  Please consider my comments non-binding, but storage neophytes may like clarification in the api doc.

staging/src/k8s.io/api/core/v1/types.go
gnufied: if user does set it then - https://github.com/kubernetes/kubernetes/pull/92770/files#diff-88702b41a33bc03bef189034a3c7e7f6R45 will replace/overwrite it with higest observed requested resource.

staging/src/k8s.io/api/core/v1/types.go
gnufied: Hmm.. I am not opposed to move this field to Status. See discussion in KEP - https://github.com/kubernetes/enhancements/pull/1516#discussion_r427017345

This is something I discussed with @liggitt at kubecon and I used in-place pod resizing KEP as a inspiration/starting point to store `allocatedResources` - https://github.com/kubernetes/enhancements/pull/1342/files

pkg/apis/core/validation/validation.go
gnufied: I do not understand this. can you reword it please?

PR #7123, opened 2015-04-21
Propagate pod and container name for log files

satnam6502:
This is a stepping stop towards propagating more useful information when ingesting cluster level logs (from Docker containers). This change modifies both of the Elasticsearch and Cloud Logging collector containers to collect from filenames that have encoded in them the `<podname><container name><Docker container ID>` which can then be used to slice and dice for queries in Elasticsearch (and at some point hopefully used for labels in Cloud Logging). For example `/var/log/containers/mypod_default-mycontainer-22be5c3295a7d2551fed0ce923a9caf61ee43d94f620ae6ca6d1a4daab5c8082.log` would link to `/var/lib/docker/containers/22be5c3295a7d2551fed0ce923a9caf61ee43d94f620ae6ca6d1a4daab5c8082/22be5c3295a7d2551fed0ce923a9caf61ee43d94f620ae6ca6d1a4daab5c8082-josn.log`
@a-robinson @vishh @mr-salty

This PR changes the Kubelet to add a symbolic link to the Docker container log file. @dchen1107 
This change causes an error log message to be written out during the integration test run reporting that a symbolic link could not be created -- but his is probably desirable since we don't want symbolic links being created on the machine that runs the integration tests.

For the moment I am leaving in the mapping to the `/var/lib/docker/containers/...` directory and I shall remove this at a later stage when I am sure this change is working well.

PR #92779, opened 2020-07-03
Return err when create ContainerLogsDir failed

fisherxu:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -1269,7 +1269,7 @@ func (kl *Kubelet) initializeModules() error {
 	// If the container logs directory does not exist, create it.
 	if _, err := os.Stat(ContainerLogsDir); err != nil {
 		if err := kl.os.MkdirAll(ContainerLogsDir, 0755); err != nil {
-			klog.Errorf("Failed to create directory %q: %v", ContainerLogsDir, err)
+			return fmt.Errorf("failed to create directory %q: %v", ContainerLogsDir, err)
 		}
 	}
 


fisherxu:
cc @yujuhong @dashpole @sjenning

dashpole:
This change makes the error fatal for the kubelet.  Can you explain why we want to make that change?

pkg/kubelet/kubelet.go
```diff
@@ -1269,7 +1269,7 @@ func (kl *Kubelet) initializeModules() error {
 	// If the container logs directory does not exist, create it.
 	if _, err := os.Stat(ContainerLogsDir); err != nil {
 		if err := kl.os.MkdirAll(ContainerLogsDir, 0755); err != nil {
-			klog.Errorf("Failed to create directory %q: %v", ContainerLogsDir, err)
+			return fmt.Errorf("Failed to create directory %q: %v", ContainerLogsDir, err)
```
SergeyKanzhelev, line 1272: also [this](https://github.com/golang/go/wiki/CodeReviewComments#error-strings):

```suggestion
			return fmt.Errorf("failed to create directory %q: %v", ContainerLogsDir, err)
```

fisherxu:
@dashpole @SergeyKanzhelev Thanks for review :) If create `ContainerLogsDir` failed, kubelet will output the err log below. So we should make sure the `ContainerLogsDir` created correctly. In addition, all the errs in the func are returned, except this one. :) 
image

pkg/kubelet/kubelet.go
fisherxu: Done.

dashpole:
I looked through the history (this was added in #7123) to see if this was done for a particular reason.  It doesn't look like we chose to log vs return the error intentionally, and I agree with the author that we shouldn't start unless we can create required directories.

/lgtm
/approve

PR #91680, opened 2020-06-02
Generic ephemeral volume API

pohly:
**What type of PR is this?**
/kind api-change

**What this PR does / why we need it**:

Ephemeral inline volumes are useful for a variety of use cases (see https://github.com/kubernetes/enhancements/blob/a080b55b7237dd765ac828e58b7bb30a99755552/keps/sig-storage/1698-generic-ephemeral-volumes/README.md#user-stories), but the current solution is both CSI specific and limited (custom API, no storage capacity tracking, no volume resizing/snapshotting, etc.). For normal storage drivers and volumes, a different approach is needed.

**Which issue(s) this PR fixes**:

Related-To: https://github.com/kubernetes/enhancements/issues/1698

**Special notes for your reviewer**:

As part of the API review, we need to clarify:
- Is it enough that pod security policy can prevent users from using the new volume typo (https://github.com/kubernetes/enhancements/tree/a080b55b7237dd765ac828e58b7bb30a99755552/keps/sig-storage/1698-generic-ephemeral-volumes#risks-and-mitigations)?
- Should `EphemeralVolumeSource` embed a `PersistentVolumeTemplate` (as in the KEP) or is a `PersistentVolumeClaimSpec` sufficient (as in this PR)? This affects how labels are handled for the created PVC. I changed my mind while working on the API change and the implementation...

**Does this PR introduce a user-facing change?**:

```release-note
Introduce API for generic ephemeral inline volumes.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

Will be added together with the actual implementation.

/cc @liggitt 
/sig storage

PR #92784, opened 2020-07-03
generic ephemeral inline volumes

pohly:
**What type of PR is this?**
/kind feature

This is the implementation PR for https://github.com/kubernetes/enhancements/issues/1698 and its API. The API was reviewed and approved in https://github.com/kubernetes/kubernetes/pull/91680#issuecomment-656356694 and will be merged as part of this PR.

Related-to https://github.com/kubernetes/enhancements/issues/1698

Only commits after `'make update' for API` are part of the implementation, everything before is part of [the API PR](https://github.com/kubernetes/kubernetes/pull/91680).

```release-note
Generic ephemeral volumes, a new alpha feature under the `GenericEphemeralVolume` feature gate, provide a more flexible alternative to `EmptyDir` volumes: as with `EmptyDir`, volumes are created and deleted for each pod automatically by Kubernetes. But because the normal provisioning process is used (`PersistentVolumeClaim`), storage can be provided by third-party storage vendors and all of the usual volume features work. Volumes don't need to be empt; for example, restoring from snapshot is supported.
```

```docs
- [KEP]: https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1698-generic-ephemeral-volumes
- [Usage]: https://github.com/kubernetes/website/pull/22438
```


pohly:
/retest

"pull-kubernetes-verify" was aborted after 2 hours, apparently it just ran too slowly.

pkg/apis/core/validation/validation.go
```diff
@@ -2647,21 +2691,31 @@ func validateEphemeralContainers(ephemeralContainers []core.EphemeralContainer,
...
-		// so that new fields will be given consideration prior to inclusion in Ephemeral Containers.
-		specType, specValue := reflect.TypeOf(ec.EphemeralContainerCommon), reflect.ValueOf(ec.EphemeralContainerCommon)
-		for i := 0; i < specType.NumField(); i++ {
-			f := specType.Field(i)
-			if allowedEphemeralContainerFields[f.Name] {
-				continue
-			}
+		// Lifecycle, probes, resources and ports should be disallowed. This is implemented as a list
+		// of allowed fields so that new fields will be given consideration prior to inclusion in Ephemeral Containers.
+		allErrs = append(allErrs, validateFieldAllowList(ec.EphemeralContainerCommon, allowedEphemeralContainerFields, "cannot be set for an Ephemeral Container", idxPath)...)
```
jsafrane, line 2696: Ephemeral Container -> Ephemeral Volume

jsafrane (commented):
I have one question: what should happen, when something deletes PVC created by the new controller.

Right now, nothing really happens - the new controller ignores such event and depending how far the pod went in the process, scheduler / kubelet may report errors or the pod itself may be already running.

The controller will re-create the PVC on the next periodic sync. Is it correct? IMO it should either re-create it every time the PVC is deleted (i.e., react to PVC deletion) or never, even on the periodic sync. 

Would it make sense to check Pod state and re-create the PVC if it's not yet Running?

pkg/controller/volume/attachdetach/util/util.go
```diff
@@ -39,57 +39,67 @@ import (
...
-func CreateVolumeSpec(podVolume v1.Volume, podNamespace string, nodeName types.NodeName, vpm *volume.VolumePluginMgr, pvcLister corelisters.PersistentVolumeClaimLister, pvLister corelisters.PersistentVolumeLister, csiMigratedPluginManager csimigration.PluginManager, csiTranslator csimigration.InTreeToCSITranslator) (*volume.Spec, error) {
+func CreateVolumeSpec(podVolume v1.Volume, pod *v1.Pod, nodeName types.NodeName, vpm *volume.VolumePluginMgr, pvcLister corelisters.PersistentVolumeClaimLister, pvLister corelisters.PersistentVolumeLister, csiMigratedPluginManager csimigration.PluginManager, csiTranslator csimigration.InTreeToCSITranslator) (*volume.Spec, error) {
+	claimName := ""
+	readOnly := false
 	if pvcSource := podVolume.VolumeSource.PersistentVolumeClaim; pvcSource != nil {
+		claimName = pvcSource.ClaimName
+		readOnly = pvcSource.ReadOnly
+	}
+	if ephemeralSource := podVolume.VolumeSource.Ephemeral; ephemeralSource != nil && utilfeature.DefaultFeatureGate.Enabled(features.GenericEphemeralVolume) {
+		claimName = pod.Name + "-" + podVolume.Name
```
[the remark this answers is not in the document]
pohly: I was wondering about that, too. I couldn't figure out where I should put it, though. It is part of the current API which explicitly documents the deterministic naming scheme. So perhaps it actually can go into `staging/src/k8s.io/api/core/v1/types.go`? I don't know whether that would be good or bad practice.

pkg/apis/core/validation/validation.go
pohly: No here. This really is the code for `ec.EphemeralContainerCommon`, I just touched it and moved code into  `validateFieldAllowList` because I wanted to use the same code also for validating allowed fields in `validatePersistentVolumeClaimTemplateObjectMeta`

pohly:
> I have one question: what should happen, when something deletes PVC created by the new controller.
> 
> Right now, nothing really happens - the new controller ignores such event and depending how far the pod went in the  process, scheduler / kubelet may report errors or the pod itself may be already running.
> 
> The controller will re-create the PVC on the next periodic sync. Is it correct?

Correct.

> IMO it should either re-create it every time the PVC is deleted (i.e., react to PVC deletion) or never, even on the periodic sync. Would it make sense to check Pod state and re-create the PVC if it's not yet Running?

I vote for always creating a missing PVC and detecting that situation sooner. The Pod state check also makes sense. Besides "not running", the controller should also ignore pods that are already marked for deletion.

pkg/apis/core/validation/validation.go
jsafrane: Oh, sorry about the noise

pkg/controller/volume/attachdetach/util/util.go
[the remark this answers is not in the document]
pohly: There's very little actual code under `k8s.io/api`, I only found one `String` implementation for a type, so this doesn't look like the right place.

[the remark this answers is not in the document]
jsafrane: There is an attempt for adding new utility functions to a separate repo, https://github.com/kubernetes/kubernetes/pull/92507, not sure it it gets through the review.

pkg/controller/volume/ephemeral/controller.go
```diff
@@ -0,0 +1,276 @@
...
+		queue:      workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "ephemeral_volume"),
+	}
+
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(klog.Infof)
+	eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
+	ec.recorder = eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: "ephemeral_volume"})
+
+	podInformer.Informer().AddEventHandler(kcache.ResourceEventHandlerFuncs{
+		AddFunc: ec.enqueuePod,
```
jsafrane, line 94: Can you please add a comment why the controller does not react to pod update? (because no new (ephemeral) volume can be added after pod creation)

test/e2e/storage/testpatterns/testpattern.go
```diff
@@ -76,11 +78,16 @@ var (
...
-	DefaultFsEphemeralVolume = TestPattern{
-		Name:    "Ephemeral-volume (default fs)",
+	// DefaultFsCSIEphemeralVolume is TestPattern for "CSI Ephemeral-volume (default fs)"
+	DefaultFsCSIEphemeralVolume = TestPattern{
+		Name:    "CSI Ephemeral-volume (default fs)",
 		VolType: CSIInlineVolume,
 	}
+	// DefaultFsGenericEphemeralVolume is TestPattern for "Generic Ephemeral-volume (default fs)"
+	DefaultFsGenericEphemeralVolume = TestPattern{
+		Name:    "Generic Ephemeral-volume (default fs) [Feature: GenericEphemeralVolume]",
```
jsafrane, line 88: No space between `Feature:` and the feature name

pkg/controller/volume/ephemeral/controller.go
pohly: Done. I also commented on pod deletion.

pohly (commented):
Updated version pushed.

pohly:
Looks like this is getting closer to being ready. I asked for enabling the E2E tests in the gce-alpha job here: https://github.com/kubernetes/test-infra/pull/18214

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -275,6 +282,27 @@ func (c *Controller) askAPIServer(pvc *v1.PersistentVolumeClaim) (bool, error) {
 }
 
 func podUsesPVC(pod *v1.Pod, pvc string) bool {
+	// A pod that has a deletionTimestamp and a zero
+	// deletionGracePeriodSeconds
+	// a) has been processed by kubelet and is ready for deletion or
```
msau42, line 282: This is not accurate. deleteGracePeriodSeconds tells kubelet when it can start force terminating the containers. Volume teardown only starts after containers are termianted. So there is an additional time period after the grace period where volume teardown is happening.

And the main signal we have today that volume teardown has finished is the deletion of the Pod object.

Adding this check to removing the finalizer early will introduce races where the volume gets deleted before unmount is complete, potentially causing unmount to hang/fail forever.

We need to think of a different way to get the signal that unmount is finished. And potentially we may still have a race for plugins that support unstage because the logic that kubelet uses to check if the pod resources are cleaned up only checks for pod volume directories, not plugin directories. cc @jsafrane

pohly: I have pushed an updated version where the PVC protection controller change is under a feature gate check. Without the feature, it should behave as before.

I also added a TODO about this questionable check to the code and my tracking list: https://github.com/kubernetes/enhancements/issues/1698#issuecomment-655344680

pohly: I also placed the extension of the PodPVCIndexer under the feature check because `podUsesPVC` is expecting the traditional behavior when the feature is off.

pohly:
Force-pushed once more. "make -j" sometimes re-generates files incorrectly and I accidentally added that when modifying the commit. Fixed.

plugin/pkg/auth/authorizer/rbac/bootstrappolicy/controller_policy.go
```diff
@@ -178,6 +178,17 @@ func buildControllerRoles() ([]rbacv1.ClusterRole, []rbacv1.ClusterRoleBinding)
 		})
 	}
 
+	if utilfeature.DefaultFeatureGate.Enabled(features.GenericEphemeralVolume) {
+		addControllerRole(&controllerRoles, &controllerRoleBindings, rbacv1.ClusterRole{
+			ObjectMeta: metav1.ObjectMeta{Name: saRolePrefix + "ephemeral-controller"},
```
msau42, line 183: Can we name this "ephemeral-volume-controller" to be more specific?

pkg/controller/volume/ephemeral/controller.go
```diff
@@ -0,0 +1,281 @@
...
+	// the API server.
+	kubeClient clientset.Interface
+
+	// pvcLister is the shared PVC lister used to fetch and store PVC
+	// objects from the API server. It is shared with other controllers and
+	// therefore the PVC objects in its store should be treated as immutable.
+	pvcLister  corelisters.PersistentVolumeClaimLister
+	pvcsSynced kcache.InformerSynced
+
+	// same for pods
```
msau42, line 61: What is the same?

pkg/controller/volume/ephemeral/controller.go
```diff
@@ -0,0 +1,281 @@
...
+	podSynced  kcache.InformerSynced
+	podIndexer cache.Indexer
+
+	// recorder is used to record events in the API server
+	recorder record.EventRecorder
+
+	queue workqueue.RateLimitingInterface
+}
+
+// NewController ephemerals the pvs
```
msau42, line 72: Can you clarify this comment more? I'm not sure I understand what it means

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -274,13 +292,48 @@ func (c *Controller) askAPIServer(pvc *v1.PersistentVolumeClaim) (bool, error) {
 	return false, nil
 }
 
-func podUsesPVC(pod *v1.Pod, pvc string) bool {
+func (c *Controller) podUsesPVC(pod *v1.Pod, pvc *v1.PersistentVolumeClaim) bool {
+	// The following text is based on how pod shutdown was
+	// initially described to me. During PR review, @msau42 pointed out
```
msau42, line 297: Prefer not to put my name into the comment. Linking to an issue # to track is fine.

plugin/pkg/auth/authorizer/rbac/bootstrappolicy/controller_policy.go
pohly: Renamed.

pkg/controller/volume/ephemeral/controller.go
pohly: As for PVC - I expanded the comments and also explain `podIndexer`.

pkg/controller/volume/ephemeral/controller.go
pohly: You are too polite... what the heck is this? Looks like the result of search/replace. Fixed.

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
pohly: Okay. It was already part of the issues's TODO list, so I am linking to that now.

I also got more paranoid about breaking something and therefore moved the invocation of `podIsShutDown`: instead of being used as a general shortcut for all volume types, it is now only used for ephemeral volumes. So even with the alpha feature enabled, nothing changes for PVCs.

pohly:
CSIStorageCapacity was merged, which ironically caused a merge conflict. Rebased.

msau42:
Discussed with @pohly. The alpha test failures are legitimate and related to late binding. Unsure if we want to block the alpha on this, but since the alpha tests are release blocking, we would have to at least disable them to merge this.

pohly:
> Discussed with @pohly. The alpha test failures are legitimate and related to late binding. Unsure if we want to block the alpha on this, but since the alpha tests are release blocking, we would have to at least disable them to merge this.

I have a fix for that problem (scheduler binder was skipped entirely for ephemeral volumes, which worked unless late binding got involved) and better E2E testing for it. The next alpha job run should be clean.

pohly:
/hold

Let's wait for clean E2E alpha results before merging.

pkg/scheduler/framework/plugins/volumebinding/volume_binding.go
```diff
@@ -61,7 +61,8 @@ func (d *stateData) Clone() framework.StateData {
 // In the Filter phase, pod binding cache is created for the pod and used in
 // Reserve and PreBind phases.
 type VolumeBinding struct {
-	Binder scheduling.SchedulerVolumeBinder
+	Binder                               scheduling.SchedulerVolumeBinder
```
msau42, line 64: Another followup item before beta. We may need changes in other volume related scheduler plugins, like VolumeAttachLimit

pohly: Added to https://github.com/kubernetes/enhancements/issues/1698#issuecomment-655344680

pkg/controller/volume/attachdetach/util/util.go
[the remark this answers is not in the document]
msau42: +1 for a util method to consolidate this logic everywhere

thockin (commented):
Thanks!

/lgtm
/approve

test/e2e/framework/pod/delete.go
```diff
@@ -57,10 +57,17 @@ func DeletePodWithWait(c clientset.Interface, pod *v1.Pod) error {
...
 func DeletePodWithWaitByName(c clientset.Interface, podName, podNamespace string) error {
 	e2elog.Logf("Deleting pod %q in namespace %q", podName, podNamespace)
-	err := c.CoreV1().Pods(podNamespace).Delete(context.TODO(), podName, metav1.DeleteOptions{})
+	deletionPolicy := metav1.DeletePropagationForeground
+	err := c.CoreV1().Pods(podNamespace).Delete(context.TODO(), podName,
+		metav1.DeleteOptions{
+			// If the pod is the owner of some resources (like ephemeral inline volumes),
+			// then we want to be sure that those are also gone before we return.
+			// Blocking pod deletion via metav1.DeletePropagationForeground achieves that.
+			PropagationPolicy: &deletionPolicy,
```
liggitt, line 69: I'm not sure switching this to foreground propagation for all existing callers was a good idea... it puts GC in the critical path and requires kube-controller-manager to be running (which I'm not sure it does in all e2e jobs, xref https://github.com/kubernetes/kubernetes/issues/93302#issuecomment-661980020)

PR #92786, opened 2020-07-04
make Kubelet bootstrap certificate signal aware

answer1991:
**What type of PR is this?**

/kind bug

When I work on #92727, I found Kubelet is not signal aware when bootstrapping certificate. Which means we can only use twice signal to kill Kubelet(through os.Exit(1)) when the certificate bootstrap is in progress. However, when kube-controller-manager or connection to kube-apiserver has some problem, the progress of certificate bootstrap may take a little long time.

None

Add a new function `SetupSignalContext` which is similar to `SetupSignalHandler`, but returned is `context.Context`. `context.Context` is more standard than stop chan(`<-chan struct{}`) as a parameter to call an async or long-term function, and I found same new function receive `context.Context` rather than `<-chan struct{}`. It's easy to get `<-chan struct{}` from `context.Context`(use `Context.Done()`), but it's not easy to build `context.Context` from `<-chan struct{}`(may cost an new goroutine).

```release-note
make Kubelet bootstrap certificate signal aware
```

```docs

```

### DIFF
--- a/pkg/kubelet/certificate/bootstrap/bootstrap.go
+++ b/pkg/kubelet/certificate/bootstrap/bootstrap.go
@@ -105,7 +105,7 @@ func LoadClientConfig(kubeconfigPath, bootstrapPath, certDir string) (certConfig
 // The kubeconfig at bootstrapPath is used to request a client certificate from the API server.
 // On success, a kubeconfig file referencing the generated key and obtained certificate is written to kubeconfigPath.
 // The certificate and key file are stored in certDir.
-func LoadClientCert(kubeconfigPath, bootstrapPath, certDir string, nodeName types.NodeName) error {
+func LoadClientCert(ctx context.Context, kubeconfigPath, bootstrapPath, certDir string, nodeName types.NodeName) error {
 	// Short-circuit if the kubeconfig file exists and is valid.
 	ok, err := isClientConfigStillValid(kubeconfigPath)
 	if err != nil {
@@ -156,11 +156,11 @@ func LoadClientCert(kubeconfigPath, bootstrapPath, certDir string, nodeName type
 		}
 	}
 
-	if err := waitForServer(*bootstrapClientConfig, 1*time.Minute); err != nil {
+	if err := waitForServer(ctx, *bootstrapClientConfig, 1*time.Minute); err != nil {
 		klog.Warningf("Error waiting for apiserver to come up: %v", err)
 	}
 
-	certData, err := requestNodeCertificate(bootstrapClient, keyData, nodeName)
+	certData, err := requestNodeCertificate(ctx, bootstrapClient, keyData, nodeName)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


alejandrox1:
@answer1991 you have to be an org member to apply an "ok-to-test" label
/ok-to-test

answer1991:
@alejandrox1 thanks

staging/src/k8s.io/apiserver/pkg/server/signal.go
```diff
@@ -44,6 +45,26 @@ func SetupSignalHandler() <-chan struct{} {
 	return stop
 }
 
+// SetupSignalContext is same as SetupSignalHandler, but a context.Context is returned.
```
liggitt, line 48: the definition of these two methods will drift... can we change SetupSignalHandler to:

```go
func SetupSignalHandler() <-chan struct{} {
  return SetupSignalContext().Done()
}
```

staging/src/k8s.io/apiserver/pkg/server/signal.go
```diff
@@ -44,6 +45,26 @@ func SetupSignalHandler() <-chan struct{} {
...
+// Only one of SetupSignalContext and SetupSignalHandler should be called, and only can
+// be called once.
```
liggitt, line 50: put the same note on SetupSignalHandler docs

pkg/kubelet/certificate/bootstrap/bootstrap.go
```diff
@@ -278,15 +278,15 @@ func verifyKeyData(data []byte) bool {
...
+func waitForServer(ctx context.Context, cfg restclient.Config, deadline time.Duration) error {
 	cfg.NegotiatedSerializer = scheme.Codecs.WithoutConversion()
 	cfg.Timeout = 1 * time.Second
 	cli, err := restclient.UnversionedRESTClientFor(&cfg)
 	if err != nil {
 		return fmt.Errorf("couldn't create client: %v", err)
 	}
 
-	ctx, cancel := context.WithTimeout(context.TODO(), deadline)
+	ctx, cancel := context.WithTimeout(ctx, deadline)
```
liggitt, line 289: also pass to the `cli.Get().AbsPath("/healthz").Do(context.TODO())` call below?

staging/src/k8s.io/apiserver/pkg/server/signal.go
answer1991: sure, already done

staging/src/k8s.io/apiserver/pkg/server/signal.go
answer1991: done

pkg/kubelet/certificate/bootstrap/bootstrap.go
answer1991: done

fedebongio:
Thank you Jordan for taking a look!

PR #92793, opened 2020-07-04
Fix throttling issues when Azure VM computer name prefix is different from VMSS name

### LINKED ISSUE #92733: Azure provider uses VMSS API instead of metadata service when computer name prefix differs from VMSS name
**What happened**:
If computer name prefix does not match VMSS name, then Azure cloud provider will use VMSS API instead of using metadata service data. This leads to throttling, because every affected kubelet is making API requests.

Azure cloud provider should use metadata service data instead of calling VMSS API when instance is current.

Set computer name prefix of the VMSS to something different than default. You should observe Azure provider making requests to the VMSS API rather than using metadata service, which leads to throttling if you have multiple kubelets with this bug.

This is bug was introduced by #89138 - `isCurrentInstance` returns false even if instance is current, because it expects
node name to match VMSS name + instance id, however node name is derived from the computer-name-prefix which can be different than VMSS name. Node name should be matched against `{computer-name-prefix}{base-36-instance-id}`, not against `{vmss-name}{base-36-instance-id}`.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.18.x and higher
- Cloud provider or hardware configuration: azure

/sig cloud-provider

feiskyer:
**What type of PR is this?**

/kind bug
/area provider/azure
/sig cloud-provider

If computer name prefix does not match VMSS name, then Azure cloud provider will use VMSS API instead of using metadata service data. This leads to throttling, because every affected kubelet is making API requests. This is because computer name prefix is not included in IMDS, hence NodeName  can't be constructed without VMSS API. 

This PR fixes the issue by getting the NodeName from hostname directly. Because Windows container doesn't support hostNetwork, environment variable "NODE_NAME" would be used as a workaround for Windows containers (e.g. out-of-tree cloud provider).

Fixes #92733

```release-note
Fix throttling issues when Azure VM computer name prefix is different from VMSS name
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/legacy-cloud-providers/azure/azure_instances.go
+++ b/staging/src/k8s.io/legacy-cloud-providers/azure/azure_instances.go
@@ -21,7 +21,7 @@ package azure
 import (
 	"context"
 	"fmt"
-	"strconv"
+	"os"
 	"strings"
 
 	v1 "k8s.io/api/core/v1"
@@ -36,6 +36,10 @@ const (
 	vmPowerStateStopped      = "stopped"
 	vmPowerStateDeallocated  = "deallocated"
 	vmPowerStateDeallocating = "deallocating"
+
+	// nodeNameEnvironmentName is the environment variable name for getting node name.
+	// It is only used for out-of-tree cloud provider.
+	nodeNameEnvironmentName = "NODE_NAME"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


staging/src/k8s.io/legacy-cloud-providers/azure/azure_instances.go
```diff
@@ -310,17 +314,20 @@ func (az *Cloud) InstanceMetadataByProviderID(ctx context.Context, providerID st
...
-			if err != nil {
-				return false, fmt.Errorf("failed to parse VMSS instanceID %q: %v", instanceID, err)
-			}
-			metadataVMName = fmt.Sprintf("%s%06s", ssName, strconv.FormatInt(instance, 36))
+		metadataVMName, err = os.Hostname()
+		if err != nil {
+			return false, err
+		}
+
+		// Use name from env variable "NODE_NAME" if it is set.
```
andyzhangx, line 327: in the PR, said `HOST_NAME`, while it's `NODE_NAME` here? also, is there code change needed on aks & aks-engine, etc.?

andyzhangx: /hold 
hold a while for clarification, thanks.

feiskyer: fixed, should be "NODE_NAME". No changes are required from other components.

PR #92797, opened 2020-07-04
Return a FitError when PreFilter fails with unschedulable status

### LINKED ISSUE #91904: Early filtering at PreFilter extension point
Update the semantics of ```PreFilter``` extension point to allow it to act as an early filtering stage. PreFilter can be used to reject pods before evaluating each node at the Filter extension point. This is useful to quickly reject unschedulable pods in cases that don't depend on the particular node (e.g., pod requesting unbound immediate PersistentVolumeClaims).

A non ```Success``` PreFilter gets converted to Error: https://github.com/kubernetes/kubernetes/blob/5dbe3e308af1598536cda3b785fb05b30505990f/pkg/scheduler/core/generic_scheduler.go#L398

We should change that and support proper rejection of the pod at PreFilter. The only acceptable unschedulable status for now should be  ```UnschedulableAndUnresolvable```.

/sig scheduling

ahg-g:
**What type of PR is this?**

/kind feature

PreFilter can return an unschedulable status, for example VolumeBinding plugin can return unschedulable at prefilter if any of the volumes with immediate VolumeBindingMode is not yet provisioned.

Fixes #91904

-->
```release-note
NONE
```
/assign @Huang-Wei

### DIFF
--- a/pkg/scheduler/core/generic_scheduler.go
+++ b/pkg/scheduler/core/generic_scheduler.go
@@ -151,13 +151,13 @@ func (g *genericScheduler) Schedule(ctx context.Context, prof *profile.Profile, 
 	}
 
 	startPredicateEvalTime := time.Now()
-	filteredNodes, filteredNodesStatuses, err := g.findNodesThatFitPod(ctx, prof, state, pod)
+	feasibleNodes, filteredNodesStatuses, err := g.findNodesThatFitPod(ctx, prof, state, pod)
 	if err != nil {
 		return result, err
 	}
 	trace.Step("Computing predicates done")
 
-	if len(filteredNodes) == 0 {
+	if len(feasibleNodes) == 0 {
 		return result, &FitError{
 			Pod:                   pod,
 			NumAllNodes:           g.nodeInfoSnapshot.NumNodes(),
@@ -170,16 +170,16 @@ func (g *genericScheduler) Schedule(ctx context.Context, prof *profile.Profile, 
 
 	startPriorityEvalTime := time.Now()
 	// When only one node after predicate, just use it.
-	if len(filteredNodes) == 1 {
+	if len(feasibleNodes) == 1 {
 		metrics.DeprecatedSchedulingAlgorithmPriorityEvaluationSecondsDuration.Observe(metrics.SinceInSeconds(startPriorityEvalTime))
 		return ScheduleResult{
-			SuggestedHost:  filteredNodes[0].Name,
+			SuggestedHost:  feasibleNodes[0].Name,
 			EvaluatedNodes: 1 + len(filteredNodesStatuses),
 			FeasibleNodes:  1,
 		}, nil
 	}
 
-	priorityList, err := g.prioritizeNodes(ctx, prof, state, pod, filteredNodes)
+	priorityList, err := g.prioritizeNodes(ctx, prof, state, pod, feasibleNodes)
 	if err != nil {
 		return result, err
 	}
@@ -192,8 +192,8 @@ func (g *genericScheduler) Schedule(ctx context.Context, prof *profile.Profile, 
 
 	return ScheduleResult{
 		SuggestedHost:  host,
-		EvaluatedNodes: len(filteredNodes) + len(filteredNodesStatuses),
-		FeasibleNodes:  len(filteredNodes),
+		EvaluatedNodes: len(feasibleNodes) + len(filteredNodesStatuses),
+		FeasibleNodes:  len(feasibleNodes),
 	}, err
 }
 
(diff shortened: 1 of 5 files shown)


alculquicondor:
lgtm, but leaving it to @Huang-Wei

pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -677,6 +677,43 @@ func TestGenericScheduler(t *testing.T) {
...
+				st.RegisterPreFilterPlugin(
+					"FakePReFilter",
+					st.NewFakePreFilterPlugin(framework.NewStatus(framework.Error, "injected error status")),
+				),
+				st.RegisterBindPlugin(defaultbinder.Name, defaultbinder.New),
+			},
+			nodes:         []string{"1", "2"},
+			pod:           &v1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "test-prefilter", UID: types.UID("test-prefilter")}},
+			expectedHosts: nil,
+			wErr:          fmt.Errorf("prefilter plugin \"FakePreFilter\" failed for pod \"test-prefilter\": injected error status"),
```
Huang-Wei, line 715:
```suggestion
			wErr:          fmt.Errorf(`prefilter plugin "FakePreFilter" failed for pod "test-prefilter": injected error status`),
```

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -253,23 +253,36 @@ func (g *genericScheduler) numFeasibleNodesToFind(numAllNodes int32) (numNodes i
...
+			// needed to avoid this copy.
+			allNodes, err := g.nodeInfoSnapshot.NodeInfos().List()
+			if err != nil {
+				return nil, nil, err
+			}
+			for _, n := range allNodes {
+				filteredNodesStatuses[n.Node().Name] = s
+			}
+			return nil, filteredNodesStatuses, nil
+		}
```
Huang-Wei, line 272: nit, in terms of "indenting negative path":

```go
	if !s.IsSuccess() {
		if !s.IsUnschedulable() {
			return nil, nil, s.AsError()
		}
		// All nodes will have the same status. Some non trivial refactoring is
		// needed to avoid this copy.
		allNodes, err := g.nodeInfoSnapshot.NodeInfos().List()
		if err != nil {
			return nil, nil, err
		}
		for _, n := range allNodes {
			filteredNodesStatuses[n.Node().Name] = s
		}
		return nil, filteredNodesStatuses, nil
	}
```

Huang-Wei (commented):
Thanks @ahg-g . Some nits.

pkg/scheduler/core/generic_scheduler.go
ahg-g: done

pkg/scheduler/core/generic_scheduler_test.go
ahg-g: done

pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -677,6 +677,43 @@ func TestGenericScheduler(t *testing.T) {
 			expectedHosts: nil,
 			wErr:          nil,
 		},
+		{
+			name: "test prefilter plugin returning Unschedulable status",
+			registerPlugins: []st.RegisterPluginFunc{
+				st.RegisterQueueSortPlugin(queuesort.Name, queuesort.New),
+				st.RegisterPreFilterPlugin(
+					"FakePReFilter",
```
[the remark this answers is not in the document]
ahg-g: done

alculquicondor (commented):
still lgtm

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -253,23 +253,37 @@ func (g *genericScheduler) numFeasibleNodesToFind(numAllNodes int32) (numNodes i
 // Filters the nodes to find the ones that fit the pod based on the framework
 // filter plugins and filter extenders.
 func (g *genericScheduler) findNodesThatFitPod(ctx context.Context, prof *profile.Profile, state *framework.CycleState, pod *v1.Pod) ([]*v1.Node, framework.NodeToStatusMap, error) {
+	filteredNodesStatuses := make(framework.NodeToStatusMap)
+
 	// Run "prefilter" plugins.
 	s := prof.RunPreFilterPlugins(ctx, state, pod)
 	if !s.IsSuccess() {
-		return nil, nil, s.AsError()
+		if !s.IsUnschedulable() {
```
Huang-Wei, line 261: I rethink it a bit.

The primary benefit of returning a FitError (instead of Error) is to continue preemption. Given that current preemption logic doesn't proceed upon `UnschedulableAndUnresolvable` status:

https://github.com/kubernetes/kubernetes/blob/21953d15ea48972f20a8de29d58bd5ce6d913914/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go#L224-L228

(but we can argue that custom preemption plugin may proceed even with `UnschedulableAndUnresolvable` status)

So maybe we can optimize the logic here:

```go
if s.Code() == UnschedulableAndUnresolvable {
```

ahg-g: Not only to execute preemption but also proper metric reporting, so I think we should keep it this way and optimize later the fact that all nodes will have the same status.

alculquicondor: The question here is whether PreFilter plugins could return `Unschedulable`, as opposed to only `UnschedulableAndUnresolvable`.

I don't think we need to limit them.

Huang-Wei: > so I think we should keep it this way and optimize later the fact that all nodes will have the same status.

That for sure has an improvement room, like simply using a `*` to represent all nodes.

> I don't think we need to limit them.

+1. I believe it's possible that some PreFilter plugin may return `Unschedulable` as well.

PR #92801, opened 2020-07-04
feat(informer-gen): add option WithCustomInformer 

### LINKED ISSUE #92792: Request to add informers.WithCustomListOptions to SharedInformerFactory
The k8s scheduler needs to watch a number of API objects such as Pods, Nodes, Services, ReplicaSet, PDBs, etc. So SharedInformerFactory is a perfect fit to watch/list objects among its codebase.

One problem we're facing is that we want to keep the ListOption as default for all objects except for Pods - for Pods, we want to exclude Pods whose `.status.phase` equals to Failed or Succeded. However, in current api-machinery code, there is only `informers.WithTweakListOptions`, which unfortunately applied the custom option to all listers. So does it make sense to add a `informers.WithCustomListOptions` to SharedInformerFactory? so that it's possible to apply a custom ListOptions for **PodInformer only**.
(BTW: an alternative is to add a "informers.WithCustomInformer" so caller is able to modify the sharedXYZInformer in place)

Ref: https://github.com/kubernetes/kubernetes/pull/92682#issuecomment-653734850

/kind feature
/sig api-machinery
/sig scheduling

knight42:

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

The k8s scheduler needs to watch running pods and now there are two PodInformers, a cutomized one in scheduler and the default one in `sharedInformerFactory`. The k8s scheduler hope to use only one PodInformer to eliminate unexpected bugs.

Ref: https://github.com/kubernetes/kubernetes/pull/92682#issuecomment-653734850

Fixes #92792

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apiextensions-apiserver/examples/client-go/pkg/client/informers/externalversions/factory.go
+++ b/staging/src/k8s.io/apiextensions-apiserver/examples/client-go/pkg/client/informers/externalversions/factory.go
@@ -75,6 +75,14 @@ func WithNamespace(namespace string) SharedInformerOption {
 	}
 }
 
+// WithCustomInformer sets a custom informer for the given informer type.
+func WithCustomInformer(obj v1.Object, informer cache.SharedIndexInformer) SharedInformerOption {
+	return func(factory *sharedInformerFactory) *sharedInformerFactory {
+		factory.informers[reflect.TypeOf(obj)] = informer
+		return factory
+	}
+}
+
 // NewSharedInformerFactory constructs a new instance of sharedInformerFactory for all namespaces.
 func NewSharedInformerFactory(client versioned.Interface, defaultResync time.Duration) SharedInformerFactory {
 	return NewSharedInformerFactoryWithOptions(client, defaultResync)
--- a/staging/src/k8s.io/code-generator/cmd/informer-gen/generators/factory.go
+++ b/staging/src/k8s.io/code-generator/cmd/informer-gen/generators/factory.go
@@ -142,6 +142,14 @@ func WithNamespace(namespace string) SharedInformerOption {
 	}
 }
 
+// WithCustomInformer sets a custom informer for the given informer type.
+func WithCustomInformer(obj {{.object|raw}}, informer {{.cacheSharedIndexInformer|raw}}) SharedInformerOption {
+	return func(factory *sharedInformerFactory) *sharedInformerFactory {
+		factory.informers[reflect.TypeOf(obj)] = informer
+		return factory
+	}
+}
+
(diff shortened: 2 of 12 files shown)


staging/src/k8s.io/code-generator/cmd/informer-gen/generators/factory.go
```diff
@@ -142,6 +142,14 @@ func WithNamespace(namespace string) SharedInformerOption {
...
 }
 
+// WithCustomInformer sets a custom informer for the given informer type.
+func WithCustomInformer(obj {{.object|raw}}, informer {{.cacheSharedIndexInformer|raw}}) SharedInformerOption {
+	return func(factory *sharedInformerFactory) *sharedInformerFactory {
+		factory.informers[reflect.TypeOf(obj)] = informer
+		return factory
+	}
+}
+
```
knight42, line 152: for reviewer: these lines are the major modification

staging/src/k8s.io/apiextensions-apiserver/examples/client-go/pkg/client/informers/externalversions/factory.go
```diff
@@ -75,6 +75,14 @@ func WithNamespace(namespace string) SharedInformerOption {
...
 }
 
+// WithCustomInformer sets a custom informer for the given informer type.
+func WithCustomInformer(obj v1.Object, informer cache.SharedIndexInformer) SharedInformerOption {
+	return func(factory *sharedInformerFactory) *sharedInformerFactory {
+		factory.informers[reflect.TypeOf(obj)] = informer
+		return factory
+	}
+}
+
```
sttts, line 85: can't the same be done already today?

```
factory.InformerFor(obj, func(versioned.Interface, time.Duration) cache.SharedIndexInformer {
  return informer
})
```

sttts: Not that calling `InformerFor` is praticularly nice. I actually prefer the option func style.

knight42: @sttts you are right, I have overlooked that. so.. do you think it is worhwhile adding this option func?

Huang-Wei: Thanks @sttts. That perfectly fits my requirement.

I will leave it to you to decide whether to continue the "WithCustomInformer" idea or not.

knight42:
close for inactivity
/close

Issue #92799, opened 2020-07-04
"KUBE_MAX_PD_VOLS" is ineffective starting from k8s 1.17.


hardikdr:
**What happened**: Setting `KUBE_MAX_PD_VOLS ` as an environment variable in kube-scheduler doesn't configure custom-limits for max-attachable volumes. Related doc: https://kubernetes.io/docs/concepts/storage/storage-limits/#custom-limits.

I tried setting `KUBE_MAX_PD_VOLS` to 30 for `c5d.metal` based node in AWS, but kube-scheduler restricts the number of volumes to default (25) for the node. 

This seems to be an issue specifically starting from k8s 1.17+. 
- I have been able to configure the max-attachable volumes limit via env-var with the hack of setting the `AttachVolumeLimit` feature-gate to `false` for k8s 1.16.

**What you expected to happen**: I expected `KUBE_MAX_PD_VOLS`  env-var to overwrite the max-attachable volumes for a kubernetes node. 

**How to reproduce it (as minimally and precisely as possible)**: 
1. Spawn a k8s cluster in AWS with k8s 1.17 or 1.18. 
2. Set the  `KUBE_MAX_PD_VOLS` env-var in kube-scheduler deployment to 3.
3. Try attaching more than 3 persistent-volumes/EBS-disks to any k8s node via statefulset.
4. The scheduler should block the attachment of the 4th volume on a node, but it doesn't.

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):  
```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.0", GitCommit:"641856db18352033a0d96dbc99153fa3b27298e5", GitTreeState:"clean", BuildDate:"2019-03-25T15:53:57Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.4", GitCommit:"c96aede7b5205121079932896c4ad89bb93260af", GitTreeState:"clean", BuildDate:"2020-06-17T11:33:59Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```
- Cloud provider or hardware configuration: AWS [though the issue is cloud-provider agnostic]

hardikdr:
cc @paurosello

liggitt:
Do you have the exact envvar you're setting? 

[...]

PR #92806, opened 2020-07-05
[1.17] Respect maxPDVols env-var against node allocatable

hardikdr:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: This PR respects the maxPDVols against the max-attachable-volumes limits set on the node-allocatable. This seems to be a regression after disabling the `AttachVolumeLimit` feature-gate in kube-scheduler. 
- Refer the code-path from 1.16 [here](https://github.com/kubernetes/kubernetes/blob/release-1.16/pkg/scheduler/algorithm/predicates/predicates.go#L536-L546).

Fixes https://github.com/kubernetes/kubernetes/issues/92799

```release-note
Respect maxPDVols env-var against node allocatable
```

```docs

```

### DIFF
--- a/pkg/scheduler/algorithm/predicates/predicates.go
+++ b/pkg/scheduler/algorithm/predicates/predicates.go
@@ -476,7 +476,7 @@ func (c *MaxPDVolumeCountChecker) predicate(pod *v1.Pod, meta Metadata, nodeInfo
 	maxAttachLimit := c.maxVolumeFunc(node)
 
 	volumeLimits := nodeInfo.VolumeLimits()
-	if maxAttachLimitFromAllocatable, ok := volumeLimits[c.volumeLimitKey]; ok {
+	if maxAttachLimitFromAllocatable, ok := volumeLimits[c.volumeLimitKey]; ok && getMaxVolLimitFromEnv() < 0 {
 		maxAttachLimit = int(maxAttachLimitFromAllocatable)
 	}
 


hardikdr:
cc @paurosello

hardikdr:
Ping @hex108 @wgliang

pkg/scheduler/algorithm/predicates/predicates.go
```diff
@@ -476,7 +476,7 @@ func (c *MaxPDVolumeCountChecker) predicate(pod *v1.Pod, meta Metadata, nodeInfo
 	maxAttachLimit := c.maxVolumeFunc(node)
 
 	volumeLimits := nodeInfo.VolumeLimits()
-	if maxAttachLimitFromAllocatable, ok := volumeLimits[c.volumeLimitKey]; ok {
+	if maxAttachLimitFromAllocatable, ok := volumeLimits[c.volumeLimitKey]; ok && getMaxVolLimitFromEnv() < 0 {
```
paurosello, line 479: Looks fine, it would only be overwritten if the environment variable is not set right?

hardikdr: Yes, when the env-var is not set, maxAttachLimit will be overwritten by the value found in the NodeAllocatable.

bertinatto: @hardikdr, I think you have to submit the PR against the master branch and once it's merged there, backport the patch all the way down to v1.17 (you may check the guidelines to be sure).

Also, it seems like the limit from the Node always had precedence over the limit from the env var. I don't know if this is an oversight or the intended behaviour. I'll try to find this out and keep you in the loop.

In addition to that, we are going to need unit tests for this change if we decide to go with it.

hardikdr: @bertinatto Thanks for looking into it. 
I think this change is only applicable on 1.17,  the code-path is completely changed starting from 1.18. There, in 1.18, also we need to fix the working of the env-var, but I don't think it'll exactly be the same change as here or even this trivial. 

>Also, it seems like the limit from the Node always had precedence over the limit from the env var.
I think the env-var should have the highest priority, it's purpose is to overwrite existing values. My best guess would be, it's an oversight :)

I'll add the unit-tests, thanks.

bertinatto: @hardikdr 

> @bertinatto Thanks for looking into it.
> I think this change is only applicable on 1.17, the code-path is completely changed starting from 1.18. There, in 1.18, also we need to fix the working of the env-var, but I don't think it'll exactly be the same change as here or even this trivial.

I'm not sure about the guidelines, but I think we need to change the code in master first anyway, then cherry-pick the change to older releases. It won't be an automated cherry-pick for sure.

The predicates code migrated to scheduler plugins in v1.18, however, the code we are interested is the same.

In v1.17: https://github.com/kubernetes/kubernetes/blob/d1c2f63bd4fc89d80e9bb6d21b5c1c24d241ad25/pkg/scheduler/algorithm/predicates/predicates.go#L475-L481
In master:  https://github.com/kubernetes/kubernetes/blob/633ab1ca61981608ffd0cfee61d40d0b6399680f/pkg/scheduler/framework/plugins/nodevolumelimits/non_csi.go#L250-L255

hardikdr: Perfect, thanks I'll also open a PR on master then.

bertinatto:
CC @gnufied

PR #89486, opened 2020-03-25
[WIP] Server-Side Apply: Status Wiping/Reset Fields implementation

kwiesmueller:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

Implements https://github.com/kubernetes/enhancements/pull/1123
Adds the ResetFieldsProvider interface and implements it for all strategies.

Requires https://github.com/kubernetes-sigs/structured-merge-diff/pull/133 to merge.

**Which issue(s) this PR fixes**:

Fixes #75564

**Does this PR introduce a user-facing change?**:

```release-note
FieldManager no longer owns fields that get reset before the object is persisted (e.g. "status wiping").
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs
- [KEP]: https://github.com/kubernetes/enhancements/pull/1123
```

/sig api-machinery
/wg apply
/cc @apelisse @jennybuckley @julianvmodesto

PR #92661, opened 2020-06-30
Bump to SMD PR#166 to pick up unsetting field changes and add integration tests

jpbetz:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Picks up https://github.com/kubernetes-sigs/structured-merge-diff/pull/166. See also the [Design Proposal](https://docs.google.com/document/d/11gUQY7Qvs1m53RTWzDzbOcDUZYMNRQCBQbTukpxmpSg/edit#heading=h.jp61o1y37pei).

Fixes #92661

**Special notes for your reviewer**:

Vendor change:

```
hack/pin-dependency.sh sigs.k8s.io/structured-merge-diff/v3 43c19bbb7fba66396555aeebbec8227d024220ab
hack/update-vendor.sh
```

Note that SMD has comprehensive testing (https://github.com/kubernetes-sigs/structured-merge-diff/pull/166). The additional integration tests here are to ensure that specific end-to-end cases we are about are working as expected.

**Does this PR introduce a user-facing change?**:

```release-note
Server-side apply behavior has been regularized in the case where a field is removed from the applied configuration. Removed fields which have no other owners are deleted from the live object, or reset to their default value if they have one. Safe ownership transfers, such as the transfer of a `replicas` field from a user to an HPA without resetting to the default value are documented in [Transferring Ownership](https://kubernetes.io/docs/reference/using-api/api-concepts/#transferring-ownership)
```

- [ ] Add 'transferring-ownership' documentation (https://github.com/kubernetes/kubernetes/issues/92844)

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

- [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/0006-apply.md)
- [Design Proposal](https://docs.google.com/document/d/11gUQY7Qvs1m53RTWzDzbOcDUZYMNRQCBQbTukpxmpSg/edit#heading=h.jp61o1y37pei)

/cc @apelisse 
/priority important-soon

PR #92809, opened 2020-07-05
 [WIP] Server-Side Apply: Status Wiping/Reset Fields implementation

### LINKED ISSUE #75564: server-side apply takes ownership of fields that don't get set due to status wiping
The API server wipes status changes on calls to the normal endpoints (not /status) at the storage registry layer (right next to etcd).  This means that calls to `<resource>/<name>` can't change status.  However, if you send a server-side apply patch, the server-side apply logic still takes ownership of any fields you *try* to set, despite the fact that they don't get set.

Reproduced on HEAD (8af1bf313efeeafb94c01ce04965ad8fe9cea21a).

Reproducer: 
- Send a server-side apply patch setting any field in status to `/api/v1/namespaces/<ns>/pods/<name>`
- Observe that ownership of that field has been taken
- Observe that the field isn't set

cc #73723 
cc @jennybuckley @apelisse

kwiesmueller:
**What type of PR is this?**

/kind bug

This is a simpler take on https://github.com/kubernetes/kubernetes/pull/89486

Implements https://github.com/kubernetes/enhancements/pull/1123
Adds the ResetFieldsProvider interface and implements it for all strategies.

Fixes #75564

```release-note
FieldManager no longer owns fields that get reset before the object is persisted (e.g. "status wiping").
```

```docs
- [KEP]: https://github.com/kubernetes/enhancements/pull/1123
```

/sig api-machinery
/wg api-expression
/assign @apelisse

### DIFF
--- a/pkg/api/persistentvolume/util.go
+++ b/pkg/api/persistentvolume/util.go
@@ -20,6 +20,7 @@ import (
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	api "k8s.io/kubernetes/pkg/apis/core"
 	"k8s.io/kubernetes/pkg/features"
+	"sigs.k8s.io/structured-merge-diff/v3/fieldpath"
 )
 
 // DropDisabledFields removes disabled fields from the pv spec.
@@ -29,6 +30,15 @@ func DropDisabledFields(pvSpec *api.PersistentVolumeSpec, oldPVSpec *api.Persist
 		if pvSpec.CSI != nil {
 			pvSpec.CSI.ControllerExpandSecretRef = nil
 		}
+	}
+}
+
+// AddDisabledFieldsTo to the provided set of resetFields
+// This should be called with the resetFields for all resources containing a Pod
+func AddDisabledFieldsTo(resetFields *fieldpath.Set) {
+	if !utilfeature.DefaultFeatureGate.Enabled(features.ExpandCSIVolumes) {
+		// TODO: make sure this path is correct, the api type has some inlines
+		resetFields.Insert(fieldpath.MakePathOrDie("spec", "csi", "controllerExpandSecretRef"))
 	}
 }
 
--- a/pkg/api/persistentvolume/util.go
+++ b/pkg/api/persistentvolume/util.go
@@ -20,7 +20,7 @@ import (
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	api "k8s.io/kubernetes/pkg/apis/core"
 	"k8s.io/kubernetes/pkg/features"
-	"sigs.k8s.io/structured-merge-diff/v3/fieldpath"
+	"sigs.k8s.io/structured-merge-diff/v4/fieldpath"
 )
 
 // DropDisabledFields removes disabled fields from the pv spec.
--- a/pkg/api/pod/util.go
+++ b/pkg/api/pod/util.go
@@ -24,7 +24,7 @@ import (
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	api "k8s.io/kubernetes/pkg/apis/core"
 	"k8s.io/kubernetes/pkg/features"
-	"sigs.k8s.io/structured-merge-diff/v3/fieldpath"
+	"sigs.k8s.io/structured-merge-diff/v4/fieldpath"
 )
 
 // ContainerType signifies container type
--- a/pkg/api/pod/util.go
+++ b/pkg/api/pod/util.go
@@ -304,6 +304,13 @@ func DropDisabledTemplateFields(podTemplate, oldPodTemplate *api.PodTemplateSpec
 	dropDisabledFields(podSpec, podAnnotations, oldPodSpec, oldPodAnnotations)
 }
 
+// AddDisabledTemplateFieldsTo the provided set of resetFields
+// This should be called with the resetFields for all resources containing a PodTemplate
+// To add the fields at the right location, pass a prefixed fieldset
+func AddDisabledTemplateFieldsTo(resetFields *fieldpath.Set) {
+	addDisabledFieldsTo(resetFields)
+}
+
 // DropDisabledPodFields removes disabled fields from the pod metadata and spec.
 // This should be called from PrepareForCreate/PrepareForUpdate for all resources containing a Pod
 func DropDisabledPodFields(pod, oldPod *api.Pod) {
--- a/pkg/api/pod/util.go
+++ b/pkg/api/pod/util.go
@@ -24,6 +24,7 @@ import (
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	api "k8s.io/kubernetes/pkg/apis/core"
 	"k8s.io/kubernetes/pkg/features"
+	"sigs.k8s.io/structured-merge-diff/v3/fieldpath"
 )
 
 // ContainerType signifies container type
@@ -328,6 +329,13 @@ func DropDisabledPodFields(pod, oldPod *api.Pod) {
 	dropPodStatusDisabledFields(podStatus, oldPodStatus)
 }
 
+// AddDisabledFields to the provided set of resetFields
+// This should be called with the resetFields for all resources containing a Pod
+func AddDisabledFields(to *fieldpath.Set) {
+	addDisabledStatusFields(to)
+	addDisabledFields(to)
+}
+
 // dropPodStatusDisabledFields removes disabled fields from the pod status
 func dropPodStatusDisabledFields(podStatus *api.PodStatus, oldPodStatus *api.PodStatus) {
 	// trim PodIPs down to only one entry (non dual stack).
@@ -336,6 +344,12 @@ func dropPodStatusDisabledFields(podStatus *api.PodStatus, oldPodStatus *api.Pod
 		if len(podStatus.PodIPs) != 0 {
 			podStatus.PodIPs = podStatus.PodIPs[0:1]
 		}
+	}
+}
+
+func addDisabledStatusFields(to *fieldpath.Set) {
+	if !utilfeature.DefaultFeatureGate.Enabled(features.IPv6DualStack) {
+		to.Insert(fieldpath.MakePathOrDie("status", "podIPs"))
 	}
 }
 
@@ -437,6 +451,32 @@ func dropDisabledFields(
 		podSpec.SetHostnameAsFQDN = nil
 	}
 
+}
+
+// dropDisabledFields removes disabled fields from the pod metadata and spec.
+func addDisabledFields(to *fieldpath.Set) {
+	if !utilfeature.DefaultFeatureGate.Enabled(features.TokenRequestProjection) {
+		// TODO: add the actual fieldpath here, this is just a probably wrong dummy based on
+		// podSpec.Volumes[i].Projected.Sources[j].ServiceAccountToken
+		to.Insert(fieldpath.MakePathOrDie("spec", "volumes", "projected", "sources", "serviceAccountToken"))
+	}
+	// TODO: add all the fields from podutil.DropDisabledPodFields (a lot)
+
+	// TODO: how to handle cases like this
+	if !utilfeature.DefaultFeatureGate.Enabled(features.AppArmor) /* && !appArmorInUse(oldPodAnnotations) */ {
+		// for k := range podAnnotations {
+		// 	if strings.HasPrefix(k, v1.AppArmorBetaContainerAnnotationKeyPrefix) {
+		// 		delete(podAnnotations, k)
+		// 	}
+		// }
+	}
+
+	if !utilfeature.DefaultFeatureGate.Enabled(features.Sysctls) {
+		// if podSpec.SecurityContext != nil {
+		// 	podSpec.SecurityContext.Sysctls = nil
+		// }
+		to.Insert(fieldpath.MakePathOrDie("spec", "securityContext", "sysctls"))
+	}
 }
 
 // dropDisabledRunAsGroupField removes disabled fields from PodSpec related
--- a/pkg/api/pod/util.go
+++ b/pkg/api/pod/util.go
@@ -19,7 +19,7 @@ package pod
 import (
 	"strings"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	api "k8s.io/kubernetes/pkg/apis/core"
@@ -329,11 +329,11 @@ func DropDisabledPodFields(pod, oldPod *api.Pod) {
 	dropPodStatusDisabledFields(podStatus, oldPodStatus)
 }
 
-// AddDisabledFields to the provided set of resetFields
+// AddDisabledFieldsTo the provided set of resetFields
 // This should be called with the resetFields for all resources containing a Pod
-func AddDisabledFields(to *fieldpath.Set) {
-	addDisabledStatusFields(to)
-	addDisabledFields(to)
+func AddDisabledFieldsTo(resetFields *fieldpath.Set) {
+	addDisabledStatusFieldsTo(resetFields)
+	addDisabledFieldsTo(resetFields)
 }
 
 // dropPodStatusDisabledFields removes disabled fields from the pod status
@@ -347,9 +347,9 @@ func dropPodStatusDisabledFields(podStatus *api.PodStatus, oldPodStatus *api.Pod
 	}
 }
 
-func addDisabledStatusFields(to *fieldpath.Set) {
+func addDisabledStatusFieldsTo(resetFields *fieldpath.Set) {
 	if !utilfeature.DefaultFeatureGate.Enabled(features.IPv6DualStack) {
-		to.Insert(fieldpath.MakePathOrDie("status", "podIPs"))
+		resetFields.Insert(fieldpath.MakePathOrDie("status", "podIPs"))
 	}
 }
 
@@ -453,12 +453,12 @@ func dropDisabledFields(
 
 }
 
-// dropDisabledFields removes disabled fields from the pod metadata and spec.
-func addDisabledFields(to *fieldpath.Set) {
+// addDisabledFields removes disabled fields from the pod metadata and spec.
+func addDisabledFieldsTo(resetFields *fieldpath.Set) {
 	if !utilfeature.DefaultFeatureGate.Enabled(features.TokenRequestProjection) {
 		// TODO: add the actual fieldpath here, this is just a probably wrong dummy based on
 		// podSpec.Volumes[i].Projected.Sources[j].ServiceAccountToken
-		to.Insert(fieldpath.MakePathOrDie("spec", "volumes", "projected", "sources", "serviceAccountToken"))
+		resetFields.Insert(fieldpath.MakePathOrDie("spec", "volumes", "projected", "sources", "serviceAccountToken"))
 	}
 	// TODO: add all the fields from podutil.DropDisabledPodFields (a lot)
 
@@ -475,7 +475,7 @@ func addDisabledFields(to *fieldpath.Set) {
 		// if podSpec.SecurityContext != nil {
 		// 	podSpec.SecurityContext.Sysctls = nil
 		// }
-		to.Insert(fieldpath.MakePathOrDie("spec", "securityContext", "sysctls"))
+		resetFields.Insert(fieldpath.MakePathOrDie("spec", "securityContext", "sysctls"))
 	}
 }
 
--- a/pkg/registry/apps/daemonset/strategy.go
+++ b/pkg/registry/apps/daemonset/strategy.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/kubernetes/pkg/apis/apps"
 	"k8s.io/kubernetes/pkg/apis/apps/validation"
 	corevalidation "k8s.io/kubernetes/pkg/apis/core/validation"
+	"sigs.k8s.io/structured-merge-diff/v3/fieldpath"
 )
 
 // daemonSetStrategy implements verification logic for daemon sets.
@@ -64,6 +65,25 @@ func (daemonSetStrategy) DefaultGarbageCollectionPolicy(ctx context.Context) res
 // NamespaceScoped returns true because all DaemonSets need to be within a namespace.
 func (daemonSetStrategy) NamespaceScoped() bool {
 	return true
+}
+
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (daemonSetStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{
+		"apps/v1": fieldpath.NewSet(
+			fieldpath.MakePathOrDie("status"),
+			fieldpath.MakePathOrDie("spec", "templateGeneration"),
+		),
+	}
+
+	// TODO: make sure this works and is the best way to do this
+	specStr, templateStr := "spec", "template"
+	pod.AddDisabledTemplateFieldsTo(fields["apps/v1"].
+		WithPrefix(fieldpath.PathElement{FieldName: &specStr}).
+		WithPrefix(fieldpath.PathElement{FieldName: &templateStr}))
+
+	return fields
 }
 
 // PrepareForCreate clears the status of a daemon set before creation.
@@ -168,6 +188,16 @@ type daemonSetStatusStrategy struct {
 
 // StatusStrategy is the default logic invoked when updating object status.
 var StatusStrategy = daemonSetStatusStrategy{Strategy}
+
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (daemonSetStatusStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	return map[fieldpath.APIVersion]*fieldpath.Set{
+		"apps/v1": fieldpath.NewSet(
+			fieldpath.MakePathOrDie("spec"),
+		),
+	}
+}
 
 func (daemonSetStatusStrategy) PrepareForUpdate(ctx context.Context, obj, old runtime.Object) {
 	newDaemonSet := obj.(*apps.DaemonSet)
--- a/pkg/registry/apps/daemonset/strategy.go
+++ b/pkg/registry/apps/daemonset/strategy.go
@@ -34,7 +34,7 @@ import (
 	"k8s.io/kubernetes/pkg/apis/apps"
 	"k8s.io/kubernetes/pkg/apis/apps/validation"
 	corevalidation "k8s.io/kubernetes/pkg/apis/core/validation"
-	"sigs.k8s.io/structured-merge-diff/v3/fieldpath"
+	"sigs.k8s.io/structured-merge-diff/v4/fieldpath"
 )
 
 // daemonSetStrategy implements verification logic for daemon sets.
--- a/pkg/registry/apps/deployment/strategy.go
+++ b/pkg/registry/apps/deployment/strategy.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/kubernetes/pkg/apis/apps"
 	"k8s.io/kubernetes/pkg/apis/apps/validation"
 	corevalidation "k8s.io/kubernetes/pkg/apis/core/validation"
+	"sigs.k8s.io/structured-merge-diff/v3/fieldpath"
 )
 
 // deploymentStrategy implements behavior for Deployments.
@@ -66,6 +67,24 @@ func (deploymentStrategy) DefaultGarbageCollectionPolicy(ctx context.Context) re
 // NamespaceScoped is true for deployment.
 func (deploymentStrategy) NamespaceScoped() bool {
 	return true
+}
+
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (deploymentStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{
+		"apps/v1": fieldpath.NewSet(
+			fieldpath.MakePathOrDie("status"),
+		),
+	}
+
+	// TODO: make sure this works and is the best way to do this
+	specStr, templateStr := "spec", "template"
+	pod.AddDisabledTemplateFieldsTo(fields["apps/v1"].
+		WithPrefix(fieldpath.PathElement{FieldName: &specStr}).
+		WithPrefix(fieldpath.PathElement{FieldName: &templateStr}))
+
+	return fields
 }
 
 // PrepareForCreate clears fields that are not allowed to be set by end users on creation.
@@ -147,6 +166,17 @@ type deploymentStatusStrategy struct {
 
 // StatusStrategy is the default logic invoked when updating object status.
 var StatusStrategy = deploymentStatusStrategy{Strategy}
+
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (deploymentStatusStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	return map[fieldpath.APIVersion]*fieldpath.Set{
+		"apps/v1": fieldpath.NewSet(
+			fieldpath.MakePathOrDie("spec"),
+			fieldpath.MakePathOrDie("metadata", "labels"),
+		),
+	}
+}
 
 // PrepareForUpdate clears fields that are not allowed to be set by end users on update of status
 func (deploymentStatusStrategy) PrepareForUpdate(ctx context.Context, obj, old runtime.Object) {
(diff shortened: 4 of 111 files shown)


staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/fieldmanager.go
```diff
@@ -77,8 +78,8 @@ func NewFieldManager(f Manager) *FieldManager {
 
 // NewDefaultFieldManager creates a new FieldManager that merges apply requests
 // and update managed fields for other types of requests.
-func NewDefaultFieldManager(models openapiproto.Models, objectConverter runtime.ObjectConvertor, objectDefaulter runtime.ObjectDefaulter, objectCreater runtime.ObjectCreater, kind schema.GroupVersionKind, hub schema.GroupVersion) (*FieldManager, error) {
-	f, err := NewStructuredMergeManager(models, objectConverter, objectDefaulter, kind.GroupVersion(), hub)
+func NewDefaultFieldManager(models openapiproto.Models, objectConverter runtime.ObjectConvertor, objectDefaulter runtime.ObjectDefaulter, objectCreater runtime.ObjectCreater, kind schema.GroupVersionKind, hub schema.GroupVersion, resetFields rest.ResetFields) (*FieldManager, error) {
```
apelisse, line 81: We're really starting to have a lot of parameters, we'll have to think about it next time we add one.

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/structuredmerge.go
```diff
@@ -55,15 +57,16 @@ func NewStructuredMergeManager(models openapiproto.Models, objectConverter runti
 		groupVersion:    gv,
 		hubVersion:      hub,
 		updater: merge.Updater{
-			Converter: internal.NewVersionConverter(typeConverter, objectConverter, hub),
+			Converter:     internal.NewVersionConverter(typeConverter, objectConverter, hub),
+			IgnoredFields: resetFields,
```
apelisse, line 61: What do you think about the discrepancy between the names? Do we ignore the fields or do we reset them? ;-)

vendor/sigs.k8s.io/structured-merge-diff/v3/fieldpath/set.go
```diff
@@ -94,6 +94,22 @@ func (s *Set) Difference(s2 *Set) *Set {
 	}
 }
 
+// RecursiveDifference returns a Set containing elements which:
```
apelisse, line 97: Just FYI, but @jpbetz is also updating SMD these days: #92661

pkg/registry/core/namespace/storage/storage.go
```diff
@@ -67,6 +67,7 @@ func NewREST(optsGetter generic.RESTOptionsGetter) (*REST, *StatusREST, *Finaliz
 		CreateStrategy:      namespace.Strategy,
 		UpdateStrategy:      namespace.Strategy,
 		DeleteStrategy:      namespace.Strategy,
+		ResetFieldsStrategy: namespace.Strategy,
```
apelisse, line 70: I think the changes in this file make sense

staging/src/k8s.io/apiserver/pkg/registry/rest/rest.go
```diff
@@ -349,3 +350,12 @@ type StorageVersionProvider interface {
 	// list of kinds the object might belong to.
 	StorageVersion() runtime.GroupVersioner
 }
+
+// ResetFields maps versions to the sets of fields that will be reset by the storage strategy.
+type ResetFields map[fieldpath.APIVersion]*fieldpath.Set
```
apelisse, line 355: I find this a little confusing, and I suspect it could lead to more confusion later.

`rest` is pulling SMD, and then `fieldmanager` is pulling `rest`, so you end-up having `rest` squeezed in the middle, that seems a little wrong. Can we have this type defined in `fieldmanager` (or don't we have a smd equivalent?)?

apelisse (commented):
I think this is the right path, thanks Kevin!

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/structuredmerge.go
kwiesmueller: Both names make sense at their respective places. Update and therefore the fieldManager really ignore them, but the strategies reset or wipe them.
I'm not sure if ignoredFields would fit in the strategy/storage context as the logic does not ignore but overwrite the users intention.

staging/src/k8s.io/apiserver/pkg/registry/rest/rest.go
kwiesmueller: I agree, I didn't add a type in SMD I think, but it makes sense to have one in the fieldManager.
The reason I put it into rest is, that rest is already imported in most places anyways.

kwiesmueller: And that the type defines the fields reset by the rest strategy

apelisse: > And that the type defines the fields reset by the rest strategy

It's fine since the requirement is coming from somewhere else.

Squeezing `rest` between `fieldmanager` and `smd` would be unfortunate, especially just for a type definition.

apelisse: Ideally, this would have come from SMD (https://github.com/kubernetes-sigs/structured-merge-diff/blob/master/merge/update.go#L34)

apelisse: Or just don't define a type :-)

staging/src/k8s.io/apiserver/pkg/registry/rest/rest.go
```diff
@@ -349,3 +350,12 @@ type StorageVersionProvider interface {
...
+
+// ResetFieldsProvider is an optional interface that a storage object can
+// implement if it wishes to provide the fields reset by its strategies.
+type ResetFieldsProvider interface {
```
apelisse, line 359: Curious why you called that "provider" rather than "strategy"? Specifically since you call it "Strategy" in the variables somewhere else?

pkg/api/pod/util.go
```diff
@@ -328,6 +329,13 @@ func DropDisabledPodFields(pod, oldPod *api.Pod) {
 	dropPodStatusDisabledFields(podStatus, oldPodStatus)
 }
 
+// AddDisabledFields to the provided set of resetFields
```
lavalamp, line 332: Does `to` mean `resetFields`? I recommend making this documentation a little clearer or changing the parameter name, even though it's very clever right now... maybe TOO clever :)

pkg/registry/core/namespace/storage/storage.go
```diff
@@ -294,6 +297,10 @@ func (r *REST) StorageVersion() runtime.GroupVersioner {
 	return r.store.StorageVersion()
 }
 
+// ResetFields implements rest.ResetFieldsProvider
+func (r *REST) ResetFields() rest.ResetFields {
```
lavalamp, line 301: "Reset" is a verb and that makes this read like a command instead of a Getter. Can we think of a noun? 

"GetResetFields" is a slight improvement :/

How about "UnchangeableFields" for this concept?

staging/src/k8s.io/apiserver/pkg/endpoints/installer.go
```diff
@@ -262,6 +263,17 @@ func (a *APIInstaller) registerResourceHandlers(path string, storage rest.Storag
 		isCreater = true
 	}
 
+	var resetFields rest.ResetFields
+	if a.group.OpenAPIModels != nil && utilfeature.DefaultFeatureGate.Enabled(features.ServerSideApply) {
+		if resetFieldsProvider, isResetFieldsProvider := storage.(rest.ResetFieldsProvider); isResetFieldsProvider {
+			resetFields = resetFieldsProvider.ResetFields()
+			// TODO(kwiesmueller): remove debug logging
```
lavalamp, line 270: remove debug logging :)

lavalamp (commented):
Mostly thoughts on how to make this more understandable...

pkg/api/persistentvolume/util.go
```diff
@@ -32,6 +33,15 @@ func DropDisabledFields(pvSpec *api.PersistentVolumeSpec, oldPVSpec *api.Persist
 	}
 }
 
+// AddDisabledFieldsTo to the provided set of resetFields
+// This should be called with the resetFields for all resources containing a Pod
+func AddDisabledFieldsTo(resetFields *fieldpath.Set) {
+	if !utilfeature.DefaultFeatureGate.Enabled(features.ExpandCSIVolumes) {
+		// TODO: make sure this path is correct, the api type has some inlines
```
kwiesmueller, line 40: TODO

pkg/api/pod/util.go
```diff
@@ -19,11 +19,12 @@ package pod
 import (
 	"strings"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
kwiesmueller, line 22: not sure why this happend, have to teach my editor to stop doing this

- [ ] @kwiesmueller remove this change

pkg/api/pod/util.go
```diff
@@ -447,6 +468,32 @@ func dropDisabledFields(
...
 }
 
+// addDisabledFields removes disabled fields from the pod metadata and spec.
+func addDisabledFieldsTo(resetFields *fieldpath.Set) {
+	if !utilfeature.DefaultFeatureGate.Enabled(features.TokenRequestProjection) {
+		// TODO: add the actual fieldpath here, this is just a probably wrong dummy based on
+		// podSpec.Volumes[i].Projected.Sources[j].ServiceAccountToken
+		resetFields.Insert(fieldpath.MakePathOrDie("spec", "volumes", "projected", "sources", "serviceAccountToken"))
+	}
+	// TODO: add all the fields from podutil.DropDisabledPodFields (a lot)
```
kwiesmueller, line 478: TODO: add all the right ignores for podutil

pkg/registry/apps/daemonset/strategy.go
```diff
@@ -66,6 +67,25 @@ func (daemonSetStrategy) NamespaceScoped() bool {
...
+// and should not be modified by the user.
+func (daemonSetStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{
+		"apps/v1": fieldpath.NewSet(
+			fieldpath.MakePathOrDie("status"),
+			fieldpath.MakePathOrDie("spec", "templateGeneration"),
+		),
+	}
+
+	// TODO: make sure this works and is the best way to do this
```
kwiesmueller, line 80: TODO: make sure template fields are added correctly
@apelisse do you know a better way to do this? Is this even correct?

pkg/registry/apps/deployment/strategy.go
```diff
@@ -68,6 +69,24 @@ func (deploymentStrategy) NamespaceScoped() bool {
...
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (deploymentStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{
+		"apps/v1": fieldpath.NewSet(
+			fieldpath.MakePathOrDie("status"),
+		),
+	}
+
+	// TODO: make sure this works and is the best way to do this
```
kwiesmueller, line 81: Rel: pkg/registry/apps/daemonset/strategy.go

pkg/registry/apps/replicaset/strategy.go
```diff
@@ -74,6 +75,24 @@ func (rsStrategy) NamespaceScoped() bool {
...
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (rsStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{
+		"apps/v1": fieldpath.NewSet(
+			fieldpath.MakePathOrDie("status"),
+		),
+	}
+
+	// TODO: make sure this works and is the best way to do this
```
kwiesmueller, line 87: Rel: pkg/registry/apps/daemonset/strategy.go

pkg/registry/apps/statefulset/strategy.go
```diff
@@ -65,6 +66,24 @@ func (statefulSetStrategy) NamespaceScoped() bool {
...
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (statefulSetStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{
+		"apps/v1": fieldpath.NewSet(
+			fieldpath.MakePathOrDie("status"),
+		),
+	}
+
+	// TODO: make sure this works and is the best way to do this
```
kwiesmueller, line 78: Rel: pkg/registry/apps/daemonset/strategy.go

pkg/registry/batch/job/strategy.go
```diff
@@ -72,6 +73,28 @@ func (jobStrategy) NamespaceScoped() bool {
...
+		"batch/v1": fieldpath.NewSet(
+			fieldpath.MakePathOrDie("status"),
+		),
+	}
+
+	if !utilfeature.DefaultFeatureGate.Enabled(features.TTLAfterFinished) {
+		fields["batch/v1"].Insert(fieldpath.MakePathOrDie("spec", "ttlSecondsAfterFinished"))
+	}
+
+	// TODO: make sure this works and is the best way to do this
```
kwiesmueller, line 89: Rel: pkg/registry/apps/daemonset/strategy.go

pkg/registry/core/namespace/strategy.go
```diff
@@ -131,6 +153,17 @@ func (namespaceFinalizeStrategy) ValidateUpdate(ctx context.Context, obj, old ru
 	return validation.ValidateNamespaceFinalizeUpdate(obj.(*api.Namespace), old.(*api.Namespace))
 }
 
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (namespaceFinalizeStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	return map[fieldpath.APIVersion]*fieldpath.Set{
+		"v1": fieldpath.NewSet(
+			fieldpath.MakePathOrDie("status"),
+			fieldpath.MakePathOrDie("spec.finalizers"),
```
kwiesmueller, line 162: This is not correct. Will fix.

- [ ] @kwiesmueller fix path

pkg/registry/core/node/strategy.go
```diff
@@ -58,6 +59,29 @@ func (nodeStrategy) NamespaceScoped() bool {
...
+		),
+	}
+
+	if !utilfeature.DefaultFeatureGate.Enabled(features.DynamicKubeletConfig) {
+		fields["v1"].Insert(fieldpath.MakePathOrDie("spec", "configSource"))
+		fields["v1"].Insert(fieldpath.MakePathOrDie("status", "config"))
+	}
+
+	if !utilfeature.DefaultFeatureGate.Enabled(features.IPv6DualStack) {
+		// TODO: probably not correct, as reset only resets to the first entry
```
kwiesmueller, line 77: What should we do with fields like this?
Just live with it not being perfect for now and ignore/not ignore?

pkg/registry/core/persistentvolumeclaim/strategy.go
```diff
@@ -49,6 +50,19 @@ func (persistentvolumeclaimStrategy) NamespaceScoped() bool {
 	return true
 }
 
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (persistentvolumeclaimStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{
+		"v1": fieldpath.NewSet(
+			fieldpath.MakePathOrDie("status"),
+			// TODO: add fields reset by pvcutil.DropDisabledFields
```
kwiesmueller, line 59: TODO: add podutil ignores

pkg/registry/core/pod/strategy.go
```diff
@@ -66,6 +67,23 @@ func (podStrategy) NamespaceScoped() bool {
 	return true
 }
 
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (podStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{
+		"v1": fieldpath.NewSet(
+			fieldpath.MakePathOrDie("status"),
+			// TODO: add fields reset by podutil.DropDisabledPodFields
```
kwiesmueller, line 76: TODO: remove comment

pkg/registry/core/replicationcontroller/strategy.go
```diff
@@ -73,6 +74,24 @@ func (rcStrategy) NamespaceScoped() bool {
...
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (rcStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{
+		"v1": fieldpath.NewSet(
+			fieldpath.MakePathOrDie("status"),
+		),
+	}
+
+	// TODO: make sure this works and is the best way to do this
```
kwiesmueller, line 86: Rel: pkg/registry/apps/daemonset/strategy.go

staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/customresource_handler.go
```diff
@@ -845,6 +845,7 @@ func (r *crdHandler) getOrCreateServingInfoFor(uid types.UID, name string) (*crd
 				reqScope.Creater,
 				reqScope.Kind,
 				reqScope.HubGroupVersion,
+				nil,
```
kwiesmueller, line 848: TODO: custom resources need to be implemented

kwiesmueller:
FYI: The kitchen recommends consuming this PR one commit at a time.

I'm doing a commit per resource being implemented, that should make it a bit easier to review the mass of changes necessary.

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/fieldmanager.go
```diff
@@ -77,8 +77,8 @@ func NewFieldManager(f Manager) *FieldManager {
 
 // NewDefaultFieldManager creates a new FieldManager that merges apply requests
 // and update managed fields for other types of requests.
-func NewDefaultFieldManager(models openapiproto.Models, objectConverter runtime.ObjectConvertor, objectDefaulter runtime.ObjectDefaulter, objectCreater runtime.ObjectCreater, kind schema.GroupVersionKind, hub schema.GroupVersion) (*FieldManager, error) {
-	f, err := NewStructuredMergeManager(models, objectConverter, objectDefaulter, kind.GroupVersion(), hub)
+func NewDefaultFieldManager(models openapiproto.Models, objectConverter runtime.ObjectConvertor, objectDefaulter runtime.ObjectDefaulter, objectCreater runtime.ObjectCreater, kind schema.GroupVersionKind, hub schema.GroupVersion, resetFields map[fieldpath.APIVersion]*fieldpath.Set) (*FieldManager, error) {
```
apelisse, line 80: I did mention it somewhere on @nodo's PR, but we're crossing the limit where the number of parameter here actually makes sense. Considering that tests have different ways of building that and some fields are sometimes left empty depending on what we're trying to do, this seems like the pattern we initially used (and made sense at that time) is not suited anymore.

Since this is strictly orthogonal, I think it's fine to leave it like this for now (this will still conflict with Andrea's PR but we should absolutely go over that after and find a better pattern (builder pattern?).

kwiesmueller: Couldn't agree more.
Didn't want to bloat this pr with it, but yeah builder might be an option. Not sure yet.
Most params are mandatory, but maybe they could be gut into smaller pieces as well.

vendor/modules.txt
```diff
@@ -2501,7 +2501,9 @@ sigs.k8s.io/kustomize/pkg/transformers
 sigs.k8s.io/kustomize/pkg/transformers/config
 sigs.k8s.io/kustomize/pkg/transformers/config/defaultconfig
 sigs.k8s.io/kustomize/pkg/types
+# sigs.k8s.io/structured-merge-diff/v3 => sigs.k8s.io/structured-merge-diff/v3 v3.0.1-0.20200706213357-43c19bbb7fba
```
kwiesmueller, line 2504: not sure why this was here after update-vendors.sh

- [ ] make sure this belongs here or remove it/it's cause

liggitt: https://github.com/kubernetes/kubernetes/pull/93867 moved to v4. Perhaps there are a few references to v3 lingering in this PR?

kwiesmueller: Found it, thanks.

pkg/registry/certificates/certificates/strategy.go
```diff
@@ -50,6 +51,23 @@ func (csrStrategy) NamespaceScoped() bool {
 	return false
 }
 
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (csrStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{
```
kwiesmueller, line 57: Certificates seem to be a special case as they are basically immutable after creation.
I added both status and spec to the resetFields as update will reset them.

pkg/registry/certificates/certificates/strategy.go
```diff
@@ -235,6 +270,23 @@ type csrApprovalStrategy struct {
 
 var ApprovalStrategy = csrApprovalStrategy{Strategy}
 
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (csrApprovalStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{
+		"certificates.k8s.io/v1": fieldpath.NewSet(
+			fieldpath.MakePathOrDie("spec"),
+			fieldpath.MakePathOrDie("status", "certificate"),
```
kwiesmueller, line 279: This rule is not explicit enough (but it does the job for now).
Technically we would need a path that says "reset everything in this path, but conditions"
As a csr status only has conditions and the cert, we should be fine, but i'm not completely satisfied.

pkg/registry/certificates/certificates/strategy.go
```diff
@@ -140,6 +158,23 @@ type csrStatusStrategy struct {
...
 
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (csrStatusStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{
+		"certificates.k8s.io/v1": fieldpath.NewSet(
+			fieldpath.MakePathOrDie("spec"),
+			fieldpath.MakePathOrDie("status", "conditions"),
+		),
+		"certificates.k8s.io/v1beta1": fieldpath.NewSet(
```
kwiesmueller, line 169: right now we ignore all conditions, but the reset code only resets some things

pkg/registry/storage/volumeattachment/strategy.go
```diff
@@ -47,6 +48,22 @@ func (volumeAttachmentStrategy) NamespaceScoped() bool {
 	return false
 }
 
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (volumeAttachmentStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{
+		"storage.k8s.io/v1": fieldpath.NewSet(
```
kwiesmueller, line 55: note: v1beta1 is missing here on purpose as status is allowed for modification there (see line 78)

staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/customresource_handler.go
```diff
@@ -884,6 +881,19 @@ func (r *crdHandler) getOrCreateServingInfoFor(uid types.UID, name string) (*crd
 			SelfLinkPathPrefix: selfLinkPrefix,
 			SelfLinkPathSuffix: "/status",
 		}
+
+		if utilfeature.DefaultFeatureGate.Enabled(features.ServerSideApply) {
```
kwiesmueller, line 885: I'm still working on crds. The test still fails for both crds and crs. Not sure why.
The resetFields get added correctly, maybe i messed up something here.
If somebody has time, please take a look. I'll continue here as soon as i can.

staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/status_strategy.go
```diff
@@ -32,6 +33,25 @@ func NewStatusStrategy(strategy customResourceStrategy) statusStrategy {
 	return statusStrategy{strategy}
 }
 
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (a statusStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{
+		"apiextensions.k8s.io/v1": fieldpath.NewSet(
```
kwiesmueller, line 40: i'm not even sure if this is correct.
we are in the customresource strategy, shouldn't the version be the respective CR version and not apiextensions.k8s.io/v1?

staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/strategy.go
```diff
@@ -72,6 +73,23 @@ func (a customResourceStrategy) NamespaceScoped() bool {
 	return a.namespaceScoped
 }
 
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (a customResourceStrategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{}
+
+	if a.status != nil {
```
kwiesmueller, line 81: same here

> i'm not even sure if this is correct.
> we are in the customresource strategy, shouldn't the version be the respective CR version and not apiextensions.k8s.io/v1?

but, we only add the resetFields if status is a subresource, that plays along with https://github.com/kubernetes/kubernetes/pull/92809/files/28a5e20d8d489c4845c1ccc55e8223aca7476a6d..d98641976458e0e4ead9eac950e4d5fdd4714d6d#diff-34aedeefa7784abbe7531bae68833ba6R95

staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresourcedefinition/strategy.go
```diff
@@ -48,6 +50,21 @@ func (strategy) NamespaceScoped() bool {
 	return false
 }
 
+// GetResetFields returns the set of fields that get reset by the strategy
+// and should not be modified by the user.
+func (strategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
+	fields := map[fieldpath.APIVersion]*fieldpath.Set{
+		"apiextensions.k8s.io/v1": fieldpath.NewSet(
```
kwiesmueller, line 57: this should work, i don't know why crds don't work yet.
the test fails like this:
```
    --- FAIL: TestApplyResetFields/apiextensions.k8s.io/v1,_Resource=customresourcedefinitions (0.02s)
        resetFields_test.go:138: Failed to apply: Apply failed with 1 conflict: conflict with "apply.test" using apiextensions.k8s.io/v1: .status.conditions
    --- FAIL: TestApplyResetFields/apiextensions.k8s.io/v1beta1,_Resource=customresourcedefinitions (0.03s)
        resetFields_test.go:138: Failed to apply: Apply failed with 1 conflict: conflict with "apply.test" using apiextensions.k8s.io/v1: .status.conditions
```
Note the conflict, it's with `apply.test` not the actual manager in the test. So something else is going on.

staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresourcedefinition/strategy.go
```diff
@@ -140,18 +157,30 @@ func (statusStrategy) NamespaceScoped() bool {
...
 	newObj.Spec = oldObj.Spec
 
 	// Status updates are for only for updating status, not objectmeta.
-	// TODO: Update after ResetObjectMetaForStatus is added to meta/v1.
-	newObj.Labels = oldObj.Labels
-	newObj.Annotations = oldObj.Annotations
-	newObj.OwnerReferences = oldObj.OwnerReferences
-	newObj.Generation = oldObj.Generation
-	newObj.SelfLink = oldObj.SelfLink
+	metav1.ResetObjectMetaForStatus(&newObj.ObjectMeta, &newObj.ObjectMeta)
```
kwiesmueller, line 183: Looks like i was able to resolve a TODO. Hope it's okay i did it on the way in this PR.

test/integration/apiserver/apply/resetFields_test.go
```diff
@@ -130,59 +129,19 @@ func TestApplyResetFields(t *testing.T) {
 					t.Fatal(err)
 				}
 
-				True := true
-				obj, err := dynamicClient.
+				// an unforced apply should not conflict if the resetfields are set correctly
```
kwiesmueller, line 132: please correct me on this.
I changed it, because Certs made some trouble.
As they don't own anything on create, my test failed with "create manager not found".
I played around with managedFields decoding and doing intersects etc. but nothing really worked. Then i thought about this and so far it looks good.
Is this a reasonable assumption?

kwiesmueller (commented):
Everything except CRDs and the Panda CR are passing the test now.
I think i still didn't add resetFields to all resources, but the test doesn't complain.
The logging still prints some debug stuff showing which resources do not return resetFields, so i'll stay on it.

Could use a second pair of eyes for CRDs i think.
And the PodUtil stuff still need "some" work.
Also do we want to add other subresources as well already? I did some where it fit, but e.g. /scale should be doable?

apelisse:
> Also do we want to add other subresources as well already? I did some where it fit, but e.g. `/scale` should be 

`/scale` is an entirely different object, and we have some other problems with that. I don't think it alter the managedFields today, does it?

test/integration/apiserver/apply/resetFields_test.go
apelisse: Thanks for writing this test. Why would it conflict in the first place if you're applying the same status?

kwiesmueller: Because I'm applying with a different fieldManager

kwiesmueller:
Yeah, you're right. Scale is a bigger problem.

test/integration/apiserver/apply/resetFields_test.go
apelisse: Right, that should work

test/integration/apiserver/apply/resetFields_test.go
```diff
@@ -89,6 +89,13 @@ func TestApplyResetFields(t *testing.T) {
 					t.Skip()
 				}
 
+				// TODO: fix this test for customresourcedefinitions
```
kwiesmueller, line 92: that's kind of an issue...
CRDs should be fine, but their status gets changed before the test finishes, so we get a conflict for .status.conditions
As the test now uses conflicts as a measure if everything is alright, that doesn't work anymore.
Need a better test or another solution to this.

staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresourcedefinition/strategy.go
kwiesmueller: Figured it out, test skips crds for now :-/
https://github.com/kubernetes/kubernetes/pull/92809#discussion_r479782666

The crd status gets changed by the system before the test applies its own status. So we get a conflict.

pkg/api/pod/util.go
```diff
@@ -439,6 +460,32 @@ func dropDisabledFields(
 
 }
 
+// addDisabledFields removes disabled fields from the pod metadata and spec.
+func addDisabledFieldsTo(resetFields *fieldpath.Set) {
+	if !utilfeature.DefaultFeatureGate.Enabled(features.TokenRequestProjection) {
+		// TODO: add the actual fieldpath here, this is just a probably wrong dummy based on
+		// podSpec.Volumes[i].Projected.Sources[j].ServiceAccountToken
```
kwiesmueller, line 467: @apelisse can we address fields like this?
The PathElement only knows fieldnames, keys, values, and index but we want to say "every index" or "every key".
Don't think we can do this right now.
This happens in a few cases.

apelisse: Good point, I don't think we have a solution for that with the fieldset. Let me think about it.

test/integration/apiserver/apply/resetFields_test.go
apelisse: Can we find another field that doesn't get updated, to check for conflicts?

kwiesmueller:
closing as this was done in another PR now

PR #92815, opened 2020-07-06
Bypass PreFilter in ServiceAfffinity if AffinityLabels arg is not present

Huang-Wei:
**What type of PR is this?**

/kind feature
/sig scheduling

The arg `AffinityLabels` in ServiceAfffinity plugin controls the Filtering logic:

https://github.com/kubernetes/kubernetes/blob/865cbf0bdfb4ae0d6cb5029ffacea127cd42e89c/pkg/scheduler/framework/plugins/serviceaffinity/service_affinity.go#L229-L232

So apparently, we should run PreFilter only when that arg is present (not a default configuration); otherwise, every scheduling cycle would waste time on ServiceAfffinity's PreFitler function.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/framework/plugins/serviceaffinity/service_affinity_test.go
+++ b/pkg/scheduler/framework/plugins/serviceaffinity/service_affinity_test.go
@@ -500,6 +500,9 @@ func TestPreFilterStateAddRemovePod(t *testing.T) {
 				p := &ServiceAffinity{
 					sharedLister:  snapshot,
 					serviceLister: fakeframework.ServiceLister(test.services),
+					args: config.ServiceAffinityArgs{
+						AffinityLabels: []string{"region", "zone"},
+					},
 				}
 				cycleState := framework.NewCycleState()
 				preFilterStatus := p.PreFilter(context.Background(), cycleState, test.pendingPod)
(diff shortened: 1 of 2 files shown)


pkg/scheduler/framework/plugins/serviceaffinity/service_affinity_test.go
```diff
@@ -500,6 +500,9 @@ func TestPreFilterStateAddRemovePod(t *testing.T) {
 				p := &ServiceAffinity{
 					sharedLister:  snapshot,
 					serviceLister: fakeframework.ServiceLister(test.services),
+					args: config.ServiceAffinityArgs{
+						AffinityLabels: []string{"region", "zone"},
```
chendave, line 504: `AffinityLabels` here is not nil, is it possible to test the new behavior (where the AffinityLabels is not present?)

alculquicondor: I think Dave's request is to add a new test for the other scenario.

Huang-Wei: I can add it, but I doubt whether it's worth it. In a real-world case, without `AffinityLabels`, the logic won't proceed to Add/Remove Pod, as the PreFilterExtensions returns nil.

It's like taking about whether we should add tests for preemption that some nodes are feasible - the prerequisite of preemption is there are no feasible nodes.

PR #89667, opened 2020-03-30
Check for sandboxes before deleting the pod from apiserver

kmala:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
This PR will make sure that all the sandboxes are removed from the runtime before deleting the Pod from the apiserver. Since kubelet doesn't interact with the networking plugins, this will make sure that all the networking resources are deleted before removing the pod from the apiserver.

**Which issue(s) this PR fixes**:

Fixes #88543

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs
```

PR #92817, opened 2020-07-06
Check for sandboxes before deleting the pod from apiserver

### LINKED ISSUE #88543: Kubelet deletes the pod before cleaning the CNI resources
**What would you like to be added**: I would like to make kubelet wait for CNI resources to be completely cleaned before deleting the pod from the apiserver. Currently it just waits for containers to be deleted and volumes to be unmounted.

**Why is this needed**:
There are couple of reason why this would make sense:
1) This will make sure that apiserver would provide the true view of the system. Because when pod is not cleaned from CNI because of any issue, then it's better for apiserver to show that the pod is still not terminated properly instead of cleaning up from etcd and relying on the GC to take care of it.
2) Currently the CNI's rely on the apiserver for additional metadata during creation and cleaning up since the kubelet doesn't pass additional metadata to the CNI which was discussed [here](https://github.com/kubernetes/kubernetes/issues/69882) and with apiserver deleting the pod before CNI is successful means CNI has to store kubernetes metadata separately.

kmala:
**What type of PR is this?**
/kind bug

This PR is to commit #89667 back into master which was reverted because of performance issue related to deletion of pods. The reason for the performance issue was because sandboxes are removed only during GC and hence waiting for the removal of sandboxes to remove the pod will be increased. In order to fix the performance issue added a new sandbox deleter which is similar to container deleter which gets triggered upon container removal event.

Fixes #88543

```release-note
Users will see increase in time for deletion of pods and also guarantee that removal of pod from api server  would mean deletion of all the resources from container runtime.
```

```docs

```


pkg/kubelet/BUILD
```diff
@@ -25,6 +25,7 @@ go_library(
         "kubelet_resources.go",
         "kubelet_volumes.go",
         "pod_container_deletor.go",
+        "pod_sandbox_deleter.go",
```
SergeyKanzhelev, line 28: nit: in order to match the previous name:

```suggestion
        "pod_sandbox_deletor.go",
```

pkg/kubelet/BUILD
```diff
@@ -176,6 +177,7 @@ go_test(
         "kubelet_volumes_linux_test.go",
         "kubelet_volumes_test.go",
         "pod_container_deletor_test.go",
+        "pod_sandbox_deleter_test.go",
```
SergeyKanzhelev, line 180: same as above - `deletor`

pkg/kubelet/container/runtime.go
```diff
@@ -114,6 +114,8 @@ type Runtime interface {
 	GetContainerLogs(ctx context.Context, pod *v1.Pod, containerID ContainerID, logOptions *v1.PodLogOptions, stdout, stderr io.Writer) (err error)
 	// Delete a container. If the container is still running, an error is returned.
 	DeleteContainer(containerID ContainerID) error
+	// Delete a sandbox. If the container is still running, an error is returned.
```
SergeyKanzhelev, line 117: linter would prefer to start the comment with the name of the method:
```suggestion
	// DeleteSandbox deletes a sandbox. If the container is still running, an error is returned.
```

pkg/kubelet/container/runtime.go
```diff
@@ -309,6 +310,8 @@ type Status struct {
 	ID ContainerID
 	// Name of the container.
 	Name string
+	// ID of the sandbox to which this container belongs.
+	PodSandboxId string
```
SergeyKanzhelev, line 314: nit: ID must be both uppercased:

```suggestion
	PodSandboxID string
```

pkg/kubelet/kubelet.go
```diff
@@ -680,6 +680,7 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 	}
 	klet.containerGC = containerGC
 	klet.containerDeletor = newPodContainerDeletor(klet.containerRuntime, integer.IntMax(containerGCPolicy.MaxPerPodContainer, minDeadContainerInPod))
+	klet.sandboxDeleter = newPodSandboxDeleter(klet.containerRuntime)
```
SergeyKanzhelev, line 683: same `deleter` vs `deletor`. Let's make consistent one way or another

pkg/kubelet/kuberuntime/kuberuntime_gc.go
```diff
@@ -161,26 +161,13 @@ func (cgc *containerGC) removeOldestNSandboxes(sandboxes []sandboxGCInfo, toRemo
 	// Remove from oldest to newest (last to first).
 	for i := len(sandboxes) - 1; i >= numToKeep; i-- {
 		if !sandboxes[i].active {
-			cgc.removeSandbox(sandboxes[i].id)
+			if err := cgc.manager.DeleteSandbox(sandboxes[i].id); err != nil {
+				klog.Errorf("Failed to remove sandbox %q: %v", sandboxes[i].id, err)
```
SergeyKanzhelev, line 165: see https://github.com/golang/go/wiki/CodeReviewComments#error-strings

```suggestion
				klog.Errorf("failed to remove sandbox %q: %v", sandboxes[i].id, err)
```

pkg/kubelet/kuberuntime/kuberuntime_sandbox.go
```diff
@@ -301,3 +301,15 @@ func (m *kubeGenericRuntimeManager) GetPortForward(podName, podNamespace string,
 	}
 	return url.Parse(resp.Url)
 }
+
+// DeleteSandbox removes the sandbox by sandboxID..
```
SergeyKanzhelev, line 305:
```suggestion
// DeleteSandbox removes the sandbox by sandboxID.
```

pkg/kubelet/kuberuntime/kuberuntime_sandbox.go
```diff
@@ -301,3 +301,15 @@ func (m *kubeGenericRuntimeManager) GetPortForward(podName, podNamespace string,
...
+func (m *kubeGenericRuntimeManager) DeleteSandbox(sandboxID string) error {
+	klog.V(4).Infof("Removing sandbox %q", sandboxID)
+	// the stop sandbox is called as part of kill pod but the error is ignored. So,
```
SergeyKanzhelev, line 308: comment says the error is ignored, but the statement below actually returns an error. Is it intentional and I'm missing something in this comment?

pkg/kubelet/pod_sandbox_deleter.go
```diff
@@ -0,0 +1,83 @@
...
+func (a sandboxStatusByCreatedList) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+func (a sandboxStatusByCreatedList) Less(i, j int) bool {
+	return a[i].CreatedAt > a[j].CreatedAt
+}
+
+func newPodSandboxDeleter(runtime kubecontainer.Runtime) *podSandboxDeleter {
+	buffer := make(chan string, sandboxDeletionBufferLimit)
+	go wait.Forever(func() {
+		for {
+			id := <-buffer
```
dashpole, line 50: nit: you can just do `for id := range buffer`

pkg/kubelet/kubelet.go
kmala: spellchecker suggests to use `deleter` and hence used it. Is it fine to change the existing one to `deleter` ?

pkg/kubelet/kubelet.go
```diff
@@ -1888,6 +1892,9 @@ func (kl *Kubelet) syncLoopIteration(configCh <-chan kubetypes.PodUpdate, handle
 				klog.V(4).Infof("SyncLoop (PLEG): ignore irrelevant event: %#v", e)
 			}
 		}
+		if e.Type == pleg.ContainerRemoved {
+			kl.deletePodSandbox(e.ID)
```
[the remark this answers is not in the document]
kmala: deletePodSandbox internally removes sandboxes only if all the containers are removed.

pkg/kubelet/kuberuntime/kuberuntime_gc.go
kmala: since we are logging the error here, it is starting with capital letter.

pkg/kubelet/kubelet.go
SergeyKanzhelev: I like deleter over deletor too. I wouldn't rename as part of this PR, perhaps a follow up clean up.

SergeyKanzhelev: My biggest concern is when code does not look consistent

pkg/kubelet/kubelet.go
```diff
@@ -2212,6 +2219,16 @@ func (kl *Kubelet) fastStatusUpdateOnce() {
 	}
 }
 
+func (kl *Kubelet) deletePodSandbox(podID types.UID) {
+	if podStatus, err := kl.podCache.Get(podID); err == nil {
+		toKeep := 1
```
dashpole, line 2224: do we need to do pruning here?   Or should we simplify everything so we just handle the case where we are deleting everything?

kmala: i didn't understand the comment exactly but the logic here is from kuberuntime gc https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_gc.go#L314

pkg/kubelet/kubelet.go
kmala: Sure, will create a follow up PR

pkg/kubelet/kubelet.go
dashpole: Basically, if the pod is still running, we clean up all sandboxes except the one that is running.  If it isn't, we remove them all.

pkg/kubelet/kubelet.go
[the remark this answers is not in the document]
SergeyKanzhelev: Maybe it can be called `deletePodSandboxIfEmpty` or something since its not clear from this logic

pkg/kubelet/kuberuntime/kuberuntime_sandbox.go
```diff
@@ -301,3 +301,15 @@ func (m *kubeGenericRuntimeManager) GetPortForward(podName, podNamespace string,
 	}
 	return url.Parse(resp.Url)
 }
+
+// DeleteSandbox removes the sandbox by sandboxID.
+func (m *kubeGenericRuntimeManager) DeleteSandbox(sandboxID string) error {
+	klog.V(4).Infof("Removing sandbox %q", sandboxID)
+	// stop sandbox is called as part of kill pod function but the error is ignored. So,
```
SergeyKanzhelev, line 308: comment got lost:

comment says the error is ignored, but the statement below actually returns an error. Is it intentional and I'm missing something in this comment?

kmala: error from stopsandbox called as part of killpod function is ignored and hence we are calling stopsandbox again as part of delete sandbox.

SergeyKanzhelev: I understand this comment now. Thank you!

pkg/kubelet/pod_sandbox_deleter.go
```diff
@@ -0,0 +1,82 @@
...
+	if toKeep > 0 {
+		sort.Sort(sandboxStatusByCreatedList(sandboxStatuses))
+	}
+
+	for i := len(sandboxStatuses) - 1; i >= toKeep; i-- {
+		if _, ok := sandboxIDs[sandboxStatuses[i].Id]; !ok && sandboxStatuses[i].State != runtimeapi.PodSandboxState_SANDBOX_READY {
+			select {
+			case p.worker <- sandboxStatuses[i].Id:
+			default:
+				klog.Warningf("Failed to issue the request to remove sandbox %v", sandboxStatuses[i].Id)
```
SergeyKanzhelev, line 78: so if there are 20 (`sandboxDeletionBufferLimit = 20`) sandboxes are being queued for deletion, this one will not be deleted and re-tried? Is it intended?

dashpole: The other pod sandbox gc routines will clean this up if this happens.  This change is to trigger sandbox cleanup quickly to make pod deletion time quicker.

SergeyKanzhelev:
Few more comments:

1. Release note (or release note "none"). I'd guess it may worth mentioning the bug fix and potential slow down of pod deletion.
2. In the associated bug @dashpole mentions that PR will need a e2e test to demonstrate the fix. See [here](https://github.com/kubernetes/kubernetes/issues/88543#issuecomment-603412994). I'm not sure how strongly it is needed.

pkg/kubelet/pod_sandbox_deleter.go
SergeyKanzhelev: thank you! This makes sense. But than the comment https://github.com/kubernetes/kubernetes/pull/92817/files#r450545856 about postponing pruning goes against this optimization, is it?

dashpole: My other comment is saying that we could choose to run this logic _only_ during deletion, rather than when sandboxes are recreated.  The "other" gc I mention above is here: https://github.com/kubernetes/kubernetes/blob/21953d15ea48972f20a8de29d58bd5ce6d913914/pkg/kubelet/kuberuntime/kuberuntime_gc.go#L408.  That logic will still do pruning

kmala:
> Few more comments:
> 
> 1. Release note (or release note "none"). I'd guess it may worth mentioning the bug fix and potential slow down of pod deletion.

Added release notes.

> 2. In the associated bug @dashpole mentions that PR will need a e2e test to demonstrate the fix. See [here](https://github.com/kubernetes/kubernetes/issues/88543#issuecomment-603412994). I'm not sure how strongly it is needed.

We had discussed about this during my previous PR but couldn't a way to demonstrate it using e2e and hence didn't implement it but just wrote unit tests.

kmala:
@SergeyKanzhelev @dashpole  if you don't have anymore comments, can you please approve

dashpole:
/lgtm
/approve
It doesn't look like any of our tests interact with the CRI directly, and I don't think we should block fixing this bug on improving the test framework.

kmala:
@dashpole @derekwaynecarr @SergeyKanzhelev  can you please approve, if you don't have any comments.

dashpole:
This is just waiting for the 1.19 code freeze to lift.  It already has lgtm and approval labels.

You can see the "tide" check below, it says "Pending — Not mergeable. Must be in milestone v1.19.", which means PRs are currently limited to 1.19-tagged PRs.

pkg/kubelet/pod_sandbox_deleter.go
```diff
@@ -0,0 +1,82 @@
...
+	}, 0)
+
+	return &podSandboxDeleter{
+		worker: buffer,
+	}
+}
+
+// deleteSandboxesInPod issues sandbox deletion requests for all inactive sandboxes after sorting by creation time
+// and skipping toKeep number of sandboxes
+func (p *podSandboxDeleter) deleteSandboxesInPod(podStatus *kubecontainer.PodStatus, toKeep int) {
```
krmayankk, line 63: when does a pod have many sandboxes ?

kmala: Whenever the sandbox container is not in active state, a new sandbox is created and if deletion of stopped sandbox container is failing for some reason, there could be multiple sandboxes which are pending deletion.

PR #92330, opened 2020-06-20
fix: don't use docker config cache if it's empty

andyzhangx:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
fix: don't use docker config cache is it's empty
In this PR
 - don't use docker config cache is it's empty
 - return empty docker config if it's using Managed Identity and image is not from ACR

**Which issue(s) this PR fixes**:

Fixes #92326

**Special notes for your reviewer**:
On a MSI enabled cluster, when non-ACR docker image is pulled first, the cache credential will only cache anonymous access credential, managed identity credential won't be fetched in the next 5min until cache expired.

**Does this PR introduce a user-facing change?**:

```release-note
fix: don't use docker config cache if it's empty
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

/kind bug
/assign @weinong  
/priority important-soon
/sig cloud-provider
/area provider/azure

PR #90425, opened 2020-04-23
fix: Allow automatic ACR auth in private Azure clouds

DavidParks8:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Provides a fix to allow a cluster in a private Azure cloud to authenticate to ACR in the same cloud.

**Which issue(s) this PR fixes**:
Fixes #90412

**Does this PR introduce a user-facing change?**:
NONE

```release-note
Provides a fix to allow a cluster in a private Azure cloud to authenticate to ACR in the same cloud.
```

PR #75587, opened 2019-03-22
Refactor AWS credential provider

tiffanyfay:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
- Adds init to aws cred provider for a bug fix
- Removes excess ecrLazyProvide and moves main functionality into Provide. The next PR will remove LazyProvide completely as no provider is using it.
- Uses ECR API's expiration instead of setting one globally
- Doesn't cache if credentials for 12 hours if they aren't found which fixes a bug below
- Breaks credential provider's dependency on cloud provider so cloud provider can be moved out of tree. 
- Creates only one provider. Also adds repoToPull to Provide interface.
- AWS cred provider can also now get ECR credentials even without being on an EC2 instance -- it only needs AWS credentials to be set

**Which issue(s) this PR fixes**:

Fixes https://github.com/kubernetes/kubernetes/issues/37962
Fixes https://github.com/google/go-containerregistry/issues/355
Needed for #69585

**Special notes for your reviewer**:
The Provide interface change to add repoToPull is needed for ECR to be able to get the registryID and region without using Cloud Provider. There are then used with the ECR API to get credentials. LazyProvide was added for AWS and no other provider uses it, nor are we using it anymore, so it will be removed in a future PR. This PR requires #75585 to be merged.

**Does this PR introduce a user-facing change?**:

```release-note
The AWS credential provider can now obtain ECR credentials even without the AWS cloud provider or being on an EC2 instance. Additionally, AWS credential provider caching has been improved to honor the ECR credential timeout.
```
/sig aws
/sig cloud-provider
/cc @yastij @andrewsykim @mcrute @nckturner @cheftako @jonjohnsonjr
/assign @liggitt @justinsb
/milestone v1.15
/priority important-soon

PR #92818, opened 2020-07-06
fix pull image error from multiple ACRs using azure managed identity

### LINKED ISSUE #92326: Pulling Azure Container Registry image using Managed Service Identity may fail
**What happened**:
When Managed Service Identity (MSI) is used to pull images from Azure Container Registry (ACR), it may fail with `Failed to pull image "<redacted>.azurecr.io/<redacted>": rpc error: code = Unknown desc = Error response from daemon: Get https://<redacted>.azurecr.io/v2/<redacted>: unauthorized: authentication required` even though the MSI has reader access to the ACR. It will recover in 1-2 minutes.

It should not have any unauthorized error.

1. run a k8s deployment using public docker image
2. immediately run a k8s deployment using ACR image
3. observe the error in get events
```
66s         Normal    Scheduled           Pod          Successfully assigned default/nginx-deployment-7fd6966748-txsdm to aks-agentpool-17585647-vmss000001
65s         Normal    Pulling             Pod          Pulling image "nginx:1.14.2"
61s         Normal    Pulled              Pod          Successfully pulled image "nginx:1.14.2"
58s         Normal    Created             Pod          Created container nginx
58s         Normal    Started             Pod          Started container nginx
66s         Normal    Scheduled           Pod          Successfully assigned default/nginx-deployment-7fd6966748-xmk6h to aks-agentpool-17585647-vmss000000
65s         Normal    Pulling             Pod          Pulling image "nginx:1.14.2"
61s         Normal    Pulled              Pod          Successfully pulled image "nginx:1.14.2"
58s         Normal    Created             Pod          Created container nginx
58s         Normal    Started             Pod          Started container nginx
66s         Normal    SuccessfulCreate    ReplicaSet   Created pod: nginx-deployment-7fd6966748-xmk6h
66s         Normal    SuccessfulCreate    ReplicaSet   Created pod: nginx-deployment-7fd6966748-txsdm
66s         Normal    SuccessfulCreate    ReplicaSet   Created pod: nginx-deployment-7fd6966748-bxb62
66s         Normal    ScalingReplicaSet   Deployment   Scaled up replica set nginx-deployment-7fd6966748 to 3
46s         Normal    Scheduled           Pod          Successfully assigned default/ubuntu-56b4744f8b-6mcpk to aks-agentpool-17585647-vmss000000
4s          Normal    Pulling             Pod          Pulling image "<redacted>.azurecr.io/<redacted>"
30s         Warning   Failed              Pod          Failed to pull image "<redacted>.azurecr.io/<redacted>": rpc error: code = Unknown desc = Error response from daemon: Get https://<redacted>.azurecr.io/v2/<redacted>: unauthorized: authentication required
30s         Warning   Failed              Pod          Error: ErrImagePull
16s         Normal    BackOff             Pod          Back-off pulling image "<redacted>.azurecr.io/<redacted>"
16s         Warning   Failed              Pod          Error: ImagePullBackOff
```

Kubelet log:
```
Jun 19 22:36:33 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:33.421665   29650 provider.go:124] Refreshing cache for provider: *credentialprovider.defaultDockerConfigProvider
Jun 19 22:36:33 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:33.422075   29650 config.go:137] looking for config.json at /var/lib/kubelet/config.json
Jun 19 22:36:33 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:33.422119   29650 config.go:137] looking for config.json at /config.json
Jun 19 22:36:33 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:33.422130   29650 config.go:137] looking for config.json at /.docker/config.json
Jun 19 22:36:33 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:33.422137   29650 config.go:137] looking for config.json at /.docker/config.json
Jun 19 22:36:33 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:33.422148   29650 config.go:107] looking for .dockercfg at /var/lib/kubelet/.dockercfg
Jun 19 22:36:33 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:33.422177   29650 config.go:107] looking for .dockercfg at /.dockercfg
Jun 19 22:36:33 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:33.422186   29650 config.go:107] looking for .dockercfg at /.dockercfg
Jun 19 22:36:33 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:33.422194   29650 config.go:107] looking for .dockercfg at /.dockercfg
Jun 19 22:36:33 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:33.422202   29650 provider.go:94] Unable to parse Docker config file: couldn't find valid .dockercfg after checking in [/var/lib/kubelet   /]
Jun 19 22:36:33 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:33.422210   29650 provider.go:124] Refreshing cache for provider: *azure.acrProvider
Jun 19 22:36:33 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:33.422232   29650 azure_credentials.go:189] try to provide secret for image docker.io/library/nginx
Jun 19 22:36:33 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:33.422244   29650 azure_credentials.go:194] image(docker.io/library/nginx) is not from ACR, skip MSI authentication
Jun 19 22:36:33 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:33.422262   29650 kuberuntime_image.go:47] Pulling image "nginx:1.14.2" without credentials
Jun 19 22:36:33 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:33.421975   29650 event.go:258] Event(v1.ObjectReference{Kind:"Pod", Namespace:"default", Name:"nginx-deployment-7fd6966748-xmk6h", UID:"128464dd-b9d3-4e54-a3f9-77d7af6e3be2", APIVersion:"v1", ResourceVersion:"158317", FieldPath:"spec.containers{nginx}"}): type: 'Normal' reason: 'Pulling' Pulling image "nginx:1.14.2"
Jun 19 22:36:37 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:37.096673   29650 kube_docker_client.go:345] Stop pulling image "nginx:1.14.2": "Status: Downloaded newer image for nginx:1.14.2"

Jun 19 22:36:54 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:54.067201   29650 event.go:258] Event(v1.ObjectReference{Kind:"Pod", Namespace:"default", Name:"ubuntu-56b4744f8b-6mcpk", UID:"d9d46d4b-7d40-4c10-82c4-07a36a788d7b", APIVersion:"v1", ResourceVersion:"158402", FieldPath:"spec.containers{ubuntu}"}): type: 'Normal' reason: 'Pulling' Pulling image "<redacted>.azurecr.io/<redacted>"
Jun 19 22:36:54 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:54.175261   29650 kubelet.go:1975] SyncLoop (housekeeping)
Jun 19 22:36:54 aks-agentpool-17585647-vmss000000 kubelet[29650]: E0619 22:36:54.310523   29650 remote_image.go:113] PullImage "<redacted>.azurecr.io/<redacted>" from image service failed: rpc error: code = Unknown desc = Error response from daemon: Get https://<redacted>.azurecr.io/v2/fundamental/test/manifests/v0.0.12: unauthorized: authentication required
Jun 19 22:36:54 aks-agentpool-17585647-vmss000000 kubelet[29650]: E0619 22:36:54.310655   29650 kuberuntime_manager.go:779] container start failed: ErrImagePull: rpc error: code = Unknown desc = Error response from daemon: Get https://<redacted>.azurecr.io/v2/fundamental/test/manifests/v0.0.12: unauthorized: authentication required
Jun 19 22:36:54 aks-agentpool-17585647-vmss000000 kubelet[29650]: E0619 22:36:54.310682   29650 pod_workers.go:190] Error syncing pod d9d46d4b-7d40-4c10-82c4-07a36a788d7b ("ubuntu-56b4744f8b-6mcpk_default(d9d46d4b-7d40-4c10-82c4-07a36a788d7b)"), skipping: failed to "StartContainer" for "ubuntu" with ErrImagePull: "rpc error: code = Unknown desc = Error response from daemon: Get https://<redacted>.azurecr.io/v2/fundamental/test/manifests/v0.0.12: unauthorized: authentication required"
Jun 19 22:36:54 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:54.310974   29650 event.go:258] Event(v1.ObjectReference{Kind:"Pod", Namespace:"default", Name:"ubuntu-56b4744f8b-6mcpk", UID:"d9d46d4b-7d40-4c10-82c4-07a36a788d7b", APIVersion:"v1", ResourceVersion:"158402", FieldPath:"spec.containers{ubuntu}"}): type: 'Warning' reason: 'Failed' Failed to pull image "<redacted>.azurecr.io/<redacted>": rpc error: code = Unknown desc = Error response from daemon: Get https://<redacted>.azurecr.io/v2/fundamental/test/manifests/v0.0.12: unauthorized: authentication required
Jun 19 22:36:54 aks-agentpool-17585647-vmss000000 kubelet[29650]: I0619 22:36:54.311026   29650 event.go:258] Event(v1.ObjectReference{Kind:"Pod", Namespace:"default", Name:"ubuntu-56b4744f8b-6mcpk", UID:"d9d46d4b-7d40-4c10-82c4-07a36a788d7b", APIVersion:"v1", ResourceVersion:"158402", FieldPath:"spec.containers{ubuntu}"}): type: 'Warning' reason: 'Failed' Error: ErrImagePull
```
When MSI is configured in kubelet and a non-ACR docker image is pulled, the code [here](https://github.com/kubernetes/kubernetes/blob/v1.15.10/pkg/credentialprovider/azure/azure_credentials.go#L194) will skip the MSI authentication flow and [return an empty docker registry cfg](https://github.com/kubernetes/kubernetes/blob/v1.15.10/pkg/credentialprovider/azure/azure_credentials.go#L213) assuming it's anonymous ACR auth. The resulting docker registry cfg is [cached in the credential provider](https://github.com/kubernetes/kubernetes/blob/v1.15.10/pkg/credentialprovider/provider.go#L120). So any image pull from ACR will use the staled docker registry cfg and got unauthorized authentication until the cache is refreshed

**Environment**:
- Kubernetes version (use `kubectl version`): 1.15.10
- Cloud provider or hardware configuration: AKS with MSI

andyzhangx:
**What type of PR is this?**
/kind bug

This PR continues with PR(https://github.com/kubernetes/kubernetes/pull/92330), fixed the pull image error with azure managed identity, does not change the behavior of service principal.

This PR would check credential cache first when pull image with azure managed identity
 - if login server(e.g. `foo.azurecr.io`) is already in the `d.cacheDockerConfig` cache, then it would continue with cache if not expired, 
 - if login server is not in the cache and it's an `*.azurecr.*` registry, then it would not use cache and invoke `Provide` directly.

Fixes #92326

```release-note
fix: pull image error from multiple ACRs using azure managed identity
```

```docs
fix: pull image error from multiple ACRs using azure managed identity
```

/kind bug
/assign @feiskyer
/priority important-soon
/sig cloud-provider
/area provider/azure

### DIFF
--- a/pkg/credentialprovider/azure/azure_credentials.go
+++ b/pkg/credentialprovider/azure/azure_credentials.go
@@ -51,6 +51,7 @@ const (
 var (
 	containerRegistryUrls = []string{"*.azurecr.io", "*.azurecr.cn", "*.azurecr.de", "*.azurecr.us"}
 	acrRE                 = regexp.MustCompile(`.*\.azurecr\.io|.*\.azurecr\.cn|.*\.azurecr\.de|.*\.azurecr\.us`)
+	azureCRName           = ".azurecr."
 )
 
 // init registers the various means by which credentials may
@@ -58,9 +59,8 @@ var (
 func init() {
 	credentialprovider.RegisterCredentialProvider("azure",
 		&credentialprovider.CachingDockerConfigProvider{
-			Provider:    NewACRProvider(flagConfigFile),
-			Lifetime:    1 * time.Minute,
-			ShouldCache: func(d credentialprovider.DockerConfig) bool { return len(d) > 0 },
+			Provider: NewACRProvider(flagConfigFile),
+			Lifetime: 1 * time.Minute,
 		})
 }
 
@@ -195,7 +195,7 @@ func (a *acrProvider) Provide(image string) credentialprovider.DockerConfig {
 		Email:    dummyRegistryEmail,
 	}
 
-	if a.config.UseManagedIdentityExtension {
+	if a.config != nil && a.config.UseManagedIdentityExtension {
 		if loginServer := a.parseACRLoginServerFromImage(image); loginServer == "" {
 			klog.V(4).Infof("image(%s) is not from ACR, skip MSI authentication", image)
 		} else {
@@ -218,7 +218,7 @@ func (a *acrProvider) Provide(image string) credentialprovider.DockerConfig {
 
 		// Handle the custom cloud case
 		// In clouds where ACR is not yet deployed, the string will be empty
-		if a.environment != nil && strings.Contains(a.environment.ContainerRegistryDNSSuffix, ".azurecr.") {
+		if a.environment != nil && strings.Contains(a.environment.ContainerRegistryDNSSuffix, azureCRName) {
 			customAcrSuffix := "*" + a.environment.ContainerRegistryDNSSuffix
 			hasBeenAdded := false
 			for _, url := range containerRegistryUrls {
@@ -302,4 +302,24 @@ func (a *acrProvider) parseACRLoginServerFromImage(image string) string {
 	}
 
 	return ""
+}
+
+// UseCache decide whether to use cache according to image and cache content
+func (a *acrProvider) UseCache(image string, config credentialprovider.DockerConfig) bool {
+	if a.config != nil && !a.config.UseManagedIdentityExtension {
+		return true
+	}
+
+	loginServer := a.parseACRLoginServerFromImage(image)
+	if loginServer == "" {
+		return true
+	}
+
+	for k := range config {
+		if strings.EqualFold(loginServer, k) {
+			return true
+		}
+	}
+	// ACR loginServer not found, should not use cache
+	return false
 }
--- a/pkg/credentialprovider/provider.go
+++ b/pkg/credentialprovider/provider.go
@@ -37,6 +37,10 @@ type DockerConfigProvider interface {
 	// implementation depends on information in the image name to return
 	// credentials; implementations are safe to ignore the image.
 	Provide(image string) DockerConfig
+
+	// decide whether to use cache according to image and cache content
+	// returns true(by default) if use cache per image and DockerConfig
+	UseCache(image string, config DockerConfig) bool
 }
 
 // A DockerConfigProvider that simply reads the .dockercfg file
@@ -57,10 +61,6 @@ func init() {
 type CachingDockerConfigProvider struct {
 	Provider DockerConfigProvider
 	Lifetime time.Duration
-
-	// ShouldCache is an optional function that returns true if the specific config should be cached.
-	// If nil, all configs are treated as cacheable.
-	ShouldCache func(DockerConfig) bool
 
 	// cache fields
 	cacheDockerConfig DockerConfig
@@ -84,6 +84,11 @@ func (d *defaultDockerConfigProvider) Provide(image string) DockerConfig {
 	return DockerConfig{}
 }
 
+// UseCache decide whether to use cache according to image and cache content
+func (d *defaultDockerConfigProvider) UseCache(image string, config DockerConfig) bool {
+	return true
+}
+
 // Enabled implements dockerConfigProvider
 func (d *CachingDockerConfigProvider) Enabled() bool {
 	return d.Provider.Enabled()
@@ -94,16 +99,31 @@ func (d *CachingDockerConfigProvider) Provide(image string) DockerConfig {
 	d.mu.Lock()
 	defer d.mu.Unlock()
 
+	useCache := true
 	// If the cache hasn't expired, return our cache
 	if time.Now().Before(d.expiration) {
-		return d.cacheDockerConfig
+		useCache = d.Provider.UseCache(image, d.cacheDockerConfig)
+		if useCache {
+			return d.cacheDockerConfig
+		}
 	}
 
 	klog.V(2).Infof("Refreshing cache for provider: %v", reflect.TypeOf(d.Provider).String())
 	config := d.Provider.Provide(image)
-	if d.ShouldCache == nil || d.ShouldCache(config) {
+	if useCache {
 		d.cacheDockerConfig = config
-		d.expiration = time.Now().Add(d.Lifetime)
+	} else {
+		klog.V(5).Infof("there is cache missing, merge two cache tables for provider: %v", reflect.TypeOf(d.Provider).String())
+		for k, v := range config {
+			d.cacheDockerConfig[k] = v
+		}
 	}
-	return config
+
+	d.expiration = time.Now().Add(d.Lifetime)
+	return d.cacheDockerConfig
+}
+
+// UseCache decide whether to use cache according to image and cache content
+func (d *CachingDockerConfigProvider) UseCache(image string, config DockerConfig) bool {
+	return true
 }
--- a/pkg/credentialprovider/aws/aws_credentials.go
+++ b/pkg/credentialprovider/aws/aws_credentials.go
@@ -104,6 +104,11 @@ func (p *ecrProvider) Provide(image string) credentialprovider.DockerConfig {
 	return cfg
 }
 
+// UseCache decide whether to use cache according to image and cache content
+func (p *ecrProvider) UseCache(image string, config credentialprovider.DockerConfig) bool {
+	return true
+}
+
 // getFromCache attempts to get credentials from the cache
 func (p *ecrProvider) getFromCache(parsed *parsedURL) (credentialprovider.DockerConfig, bool) {
 	cfg := credentialprovider.DockerConfig{}
--- a/pkg/credentialprovider/azure/BUILD
+++ b/pkg/credentialprovider/azure/BUILD
@@ -32,6 +32,7 @@ go_test(
     srcs = ["azure_credentials_test.go"],
     embed = [":go_default_library"],
     deps = [
+        "//pkg/credentialprovider:go_default_library",
         "//vendor/github.com/Azure/azure-sdk-for-go/services/containerregistry/mgmt/2019-05-01/containerregistry:go_default_library",
         "//vendor/github.com/Azure/go-autorest/autorest/azure:go_default_library",
         "//vendor/github.com/Azure/go-autorest/autorest/to:go_default_library",
--- a/pkg/credentialprovider/azure/azure_credentials_test.go
+++ b/pkg/credentialprovider/azure/azure_credentials_test.go
@@ -26,6 +26,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/containerregistry/mgmt/2019-05-01/containerregistry"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/to"
+	"k8s.io/kubernetes/pkg/credentialprovider"
 
 	"github.com/stretchr/testify/assert"
 )
@@ -197,6 +198,140 @@ func TestParseACRLoginServerFromImage(t *testing.T) {
 	for _, test := range tests {
 		if loginServer := provider.parseACRLoginServerFromImage(test.image); loginServer != test.expected {
 			t.Errorf("function parseACRLoginServerFromImage returns \"%s\" for image %s, expected \"%s\"", loginServer, test.image, test.expected)
+		}
+	}
+}
+
+func TestUseCache(t *testing.T) {
+	configSP := `
+    {
+        "aadClientId": "foo",
+        "aadClientSecret": "bar"
+	}`
+	configMI := `
+    {
+        "useManagedIdentityExtension": true
+    }`
+	result := []containerregistry.Registry{
+		{
+			Name: to.StringPtr("foo"),
+			RegistryProperties: &containerregistry.RegistryProperties{
+				LoginServer: to.StringPtr("*.azurecr.io"),
+			},
+		},
+	}
+	fakeClient := &fakeClient{
+		results: result,
+	}
+
+	providerSP := &acrProvider{
+		registryClient: fakeClient,
+	}
+	providerSP.loadConfig(bytes.NewBufferString(configSP))
+	providerSP.environment = &azure.Environment{
+		ContainerRegistryDNSSuffix: ".azurecr.my.cloud",
+	}
+
+	providerMI := &acrProvider{
+		registryClient: fakeClient,
+	}
+	providerMI.loadConfig(bytes.NewBufferString(configMI))
+	providerMI.environment = &azure.Environment{
+		ContainerRegistryDNSSuffix: ".azurecr.my.cloud",
+	}
+
+	defaultConfigEntry := credentialprovider.DockerConfigEntry{
+		Username: "",
+		Password: "",
+		Email:    dummyRegistryEmail,
+	}
+
+	// create cfg for service principal
+	cfgSP := credentialprovider.DockerConfig{}
+	for _, url := range containerRegistryUrls {
+		cred := &credentialprovider.DockerConfigEntry{
+			Username: "user",
+			Password: "password",
+			Email:    dummyRegistryEmail,
+		}
+		cfgSP[url] = *cred
+	}
+	cfgSP["*.azurecr.*"] = defaultConfigEntry
+
+	// create cfg for managed identity
+	cfgMI := credentialprovider.DockerConfig{}
+	cfgMI["foo.azurecr.io"] = defaultConfigEntry
+	cfgMI["foo.azurecr.my.cloud"] = defaultConfigEntry
+
+	tests := []struct {
+		desc           string
+		provider       credentialprovider.DockerConfigProvider
+		image          string
+		cfg            credentialprovider.DockerConfig
+		expectedResult bool
+	}{
+		{
+			desc:           "use cache for SP when it's an ACR image",
+			provider:       providerSP,
+			image:          "foo.azurecr.io/bar/image:version",
+			cfg:            cfgSP,
+			expectedResult: true,
+		},
+		{
+			desc:           "use cache for SP when it's a docker image",
+			provider:       providerSP,
+			image:          "docker.io/library/busybox:latest",
+			cfg:            cfgSP,
+			expectedResult: true,
+		},
+		{
+			desc:           "use cache for MI when ACR login server is already in the cache",
+			provider:       providerMI,
+			image:          "foo.azurecr.io/bar/image:version",
+			cfg:            cfgMI,
+			expectedResult: true,
+		},
+		{
+			desc:           "use cache for MI when ACR custom login server is already in the cache",
+			provider:       providerMI,
+			image:          "foo.azurecr.my.cloud/bar/image:version",
+			cfg:            cfgMI,
+			expectedResult: true,
+		},
+		{
+			desc:           "use cache for MI when it's a docker image",
+			provider:       providerMI,
+			image:          "docker.io/library/busybox:latest",
+			cfg:            cfgMI,
+			expectedResult: true,
+		},
+		{
+			desc:           "use cache for MI when it's an unknown image",
+			provider:       providerMI,
+			image:          "unknown",
+			cfg:            cfgMI,
+			expectedResult: true,
+		},
+		{
+			desc:           "not use cache for MI when there is a new ACR login server",
+			provider:       providerMI,
+			image:          "foo2.azurecr.io/bar/image:version",
+			cfg:            cfgMI,
+			expectedResult: false,
+		},
+		{
+			desc:           "not use cache for MI when there is a new custom ACR login server",
+			provider:       providerMI,
+			image:          "foo2.azurecr.my.cloud/bar/image:version",
+			cfg:            cfgMI,
+			expectedResult: false,
+		},
+	}
+
+	for _, test := range tests {
+		result := test.provider.UseCache(test.image, test.cfg)
+		if result != test.expectedResult {
+			t.Errorf("test case(%s) UseCache(%s) returned with %v, not equal to %v", test.desc, test.image, result, test.expectedResult)
 		}
 	}
 }
--- a/pkg/credentialprovider/gcp/metadata.go
+++ b/pkg/credentialprovider/gcp/metadata.go
@@ -159,6 +159,11 @@ func (g *dockerConfigKeyProvider) Provide(image string) credentialprovider.Docke
 	return credentialprovider.DockerConfig{}
 }
 
+// UseCache decide whether to use cache according to image and cache content
+func (g *dockerConfigKeyProvider) UseCache(image string, config credentialprovider.DockerConfig) bool {
+	return true
+}
+
 // Provide implements DockerConfigProvider
 func (g *dockerConfigURLKeyProvider) Provide(image string) credentialprovider.DockerConfig {
 	// Read the contents of the google-dockercfg-url key and load a .dockercfg from there
@@ -178,6 +183,11 @@ func (g *dockerConfigURLKeyProvider) Provide(image string) credentialprovider.Do
 	}
 
 	return credentialprovider.DockerConfig{}
+}
+
+// UseCache decide whether to use cache according to image and cache content
+func (g *dockerConfigURLKeyProvider) UseCache(image string, config credentialprovider.DockerConfig) bool {
+	return true
 }
 
 // runWithBackoff runs input function `f` with an exponential backoff.
@@ -297,4 +307,9 @@ func (g *containerRegistryProvider) Provide(image string) credentialprovider.Doc
 		cfg[k] = entry
 	}
 	return cfg
+}
+
+// UseCache decide whether to use cache according to image and cache content
+func (g *containerRegistryProvider) UseCache(image string, config credentialprovider.DockerConfig) bool {
+	return true
 }
--- a/pkg/credentialprovider/keyring_test.go
+++ b/pkg/credentialprovider/keyring_test.go
@@ -469,6 +469,11 @@ func (d *testProvider) Provide(image string) DockerConfig {
 	return DockerConfig{}
 }
 
+// UseCache implements dockerConfigProvider
+func (d *testProvider) UseCache(image string, config DockerConfig) bool {
+	return true
+}
+
 func TestProvidersDockerKeyring(t *testing.T) {
 	provider := &testProvider{
 		Count: 0,


andyzhangx:
cc @weinong

pkg/credentialprovider/azure/azure_credentials.go
```diff
@@ -301,3 +301,26 @@ func (a *acrProvider) parseACRLoginServerFromImage(image string) string {
...
+
+// shouldUseCache decide whether to use cache according to image and cache content
+func shouldUseCache(image string, d credentialprovider.DockerConfig) bool {
+	if len(d) > 2 {
+		// for service principal(len=5), always use cache
+		return true
+	}
+
+	match := acrRE.FindAllString(image, -1)
+	if len(match) != 1 {
```
northtyphoon, line 313: Do you need to handle the similar custom cloud case (eg, AzureStack) as parseACRLoginServerFromImage?

andyzhangx: @northtyphoon I think yes, do you know whether ACR login server would have `.azurecr.` in AzureStack, custom cloud case needs cloud provider specific environment, don't want to change the common provider interface, otherwise it may get rejected.

northtyphoon: /cc: @xizhamsft

weinong:
LGTM. but i have same concern as @northtyphoon on private cloud support.

pkg/credentialprovider/azure/azure_credentials.go
feiskyer: I think checking `.azurecr.` should be OK. We have added such logic in #90425.

andyzhangx: it's done, PTAL, thanks. Also updated the unit test cases.

feiskyer (commented):
/lgtm
/approve

feiskyer:
@deads2k could you help to have a look?

andyzhangx:
/assign @liggitt 
this is a refine of #92330, PTAL, thanks.

pkg/credentialprovider/azure/azure_credentials.go
```diff
@@ -301,3 +302,26 @@ func (a *acrProvider) parseACRLoginServerFromImage(image string) string {
 
 	return ""
 }
+
+// shouldUseCache decide whether to use cache according to image and cache content
+func shouldUseCache(image string, d credentialprovider.DockerConfig) bool {
+	if len(d) > 2 {
+		// for service principal(len=5), always use cache
```
liggitt, line 309: how is len=5 related to len > 2?

pkg/credentialprovider/azure/azure_credentials.go
```diff
@@ -301,3 +302,26 @@ func (a *acrProvider) parseACRLoginServerFromImage(image string) string {
...
+		return true
+	}
+
+	fields := strings.Split(image, "/")
+	if len(fields) <= 1 || !strings.Contains(fields[0], azureCRName) {
```
liggitt, line 314: presence of a `.azurecr.` segment somewhere in the image doesn't seem correct... that would mean `bogus.azurecr.example.com` would match

pkg/credentialprovider/azure/azure_credentials.go
```diff
@@ -301,3 +302,26 @@ func (a *acrProvider) parseACRLoginServerFromImage(image string) string {
...
+		// not ACR image, use cache
+		return true
+	}
+
+	loginServer := fields[0]
+	for k := range d {
+		if strings.EqualFold(loginServer, k) {
```
liggitt, line 321: are the credential keys always just the server, or do they sometimes include additional port/path info?

liggitt: does this handle port defaulting correctly (e.g. no port == 443)?

pkg/credentialprovider/provider.go
```diff
@@ -58,9 +58,9 @@ type CachingDockerConfigProvider struct {
 	Provider DockerConfigProvider
 	Lifetime time.Duration
 
-	// ShouldCache is an optional function that returns true if the specific config should be cached.
-	// If nil, all configs are treated as cacheable.
-	ShouldCache func(DockerConfig) bool
+	// ShouldUseCache is an optional function that returns true(by default) if use cache from specific config.
+	// If nil, should also use cache
+	ShouldUseCache func(string, DockerConfig) bool
```
liggitt, line 63: name the parameters, the meaning of `string` is not clear

pkg/credentialprovider/azure/azure_credentials.go
andyzhangx: credential key is always the server, e.g. `foo.azurecr.io`, no port/path included

pkg/credentialprovider/azure/azure_credentials.go
```diff
@@ -301,3 +302,26 @@ func (a *acrProvider) parseACRLoginServerFromImage(image string) string {
 
 	return ""
 }
+
+// shouldUseCache decide whether to use cache according to image and cache content
+func shouldUseCache(image string, d credentialprovider.DockerConfig) bool {
```
liggitt, line 307: It's not clear to me why it is correct for the caching decision to use the image as an input. If this returns true, the results are cached and used for the next call to Provide(), regardless of the next image passed in.

pkg/credentialprovider/azure/azure_credentials.go
andyzhangx: I think that's ok, that would waste one cache in every pull image operation.
since there is some secret azure ACR registry site would also contain *.azurecr.*, so the azurecr suffix endpoint could not be listed here, use a generic *.azurecr.* is a safe way to make sure it won't miss those secret registry sites

pkg/credentialprovider/azure/azure_credentials.go
andyzhangx: every time when `Provide(image string)` is invoked, it would use image to check whether `d credentialprovider.DockerConfig` contains server in `image`, since Azure Managed Identity uses different token per registry server, that's the reason why I worked out this PR.

pkg/credentialprovider/azure/azure_credentials.go
andyzhangx: for managed identity case, `len(d) = 2`, that's a tick, let me add one more comment

pkg/credentialprovider/azure/azure_credentials.go
liggitt: > Azure Managed Identity uses different token per registry server

if Provide is returning credentials that are only usable against the input image's registry, it doesn't seem like they should be cached for all subsequent calls to Provide, regardless of image (which is how CachingDockerConfigProvider currently works)

andyzhangx: I would say it depends on image, e.g. `foo.azurecr.io/image:ver`, `foo.azurecr.io/image2:ver`, these two images should use cache (otherwise it may cause more API to get registry token). In most cases, user would use one registry to store most of images, then we should use cache; while in other cases, if user use 2+ registries, we sometimes don't use cache.

pkg/credentialprovider/provider.go
andyzhangx: fixed and also updated the func `ShouldUseCache` comments

pkg/credentialprovider/azure/azure_credentials.go
andyzhangx: added comment

pkg/credentialprovider/azure/azure_credentials.go
liggitt: > I would say it depends on image, e.g. `foo.azurecr.io/image:ver`, `foo.azurecr.io/image2:ver`, these two images should use cache (otherwise it may cause one more API to get registry token).

The way this is currently implemented seems like a roundabout way to get per-image caching.

Would it make more sense to provide a cache key function that determines the cache key for a given image, orthogonal to ShouldCache? Something like:

```diff
 // Enabled implements dockerConfigProvider
@@ -94,16 +100,50 @@ func (d *CachingDockerConfigProvider) Provide(image string) DockerConfig {
        d.mu.Lock()
        defer d.mu.Unlock()
 
+       key := ""
+       if d.CacheKey != nil {
+               key = d.CacheKey(image)
+       }
+
        // If the cache hasn't expired, return our cache
-       if time.Now().Before(d.expiration) {
-               return d.cacheDockerConfig
+       if time.Now().Before(d.expiration[key]) {
+               return d.cacheDockerConfig[key]
        }
 
        klog.V(2).Infof("Refreshing cache for provider: %v", reflect.TypeOf(d.Provider).String())
        config := d.Provider.Provide(image)
+       now := time.Now()
        if d.ShouldCache == nil || d.ShouldCache(config) {
-               d.cacheDockerConfig = config
-               d.expiration = time.Now().Add(d.Lifetime)
+               if d.cacheDockerConfig == nil {
+                       d.cacheDockerConfig = map[string]DockerConfig{}
+               }
+               d.cacheDockerConfig[key] = config
+               if d.expiration == nil {
+                       d.expiration = map[string]time.Time{}
+               }
+               d.expiration[key] = now.Add(d.Lifetime)
        }
+
+       c.cleanupIfNeeded_locked()
+
        return config
 }
+
+// cleanupIfNeeded_locked removes expired credentials from the cache if needed.
+// must be called with mu already locked.
+func (c *CachingDockerConfigProvider) cleanupIfNeeded_locked() {
+       // skip cleanup if we have very few credentials or our cleanup time has not yet arrived
+       if len(d.cacheDockerConfig) < 100 || now.Before(d.nextCleanupTime) {
+               return
+       }
+
+       // remove expired credentials
+       for key := range d.cacheDockerConfig {
+               if now.Before(d.expiration[key]) {
+                       delete(d.expiration, key)
+                       delete(d.cacheDockerConfig, key)
+               }
+       }
+       // set our next cleanup time
+       d.nextCleanupTime = now.Add(d.Lifetime)
+}
```

andyzhangx: @liggitt thanks for the suggestion. There are two identities on Azure now: Service Principal(SP) and Managed Identity(MI), and we only want to make this change happen in MI since for SP, use cache is always ok. So based on your suggestion, we also need to add identity check here:
```
if time.Now().Before(d.expiration[key]) {
```

That would make the implementation too cloud provider specific, since this feature is going to retire in the future(use a separator credential provider for every cloud provider), I think it may be better to make this change as small as possible, otherwise in common code, we need to handle too cloud provider specific thing.

Let me know if you have any questions, or ping me in slack to make it quick, thanks!

liggitt: > we only want to make this change happen in MI

could that be determined by looking at `a.config.UseManagedIdentityExtension`? would a custom CacheKey impl like this work without adding anything cloud-specific to CachingDockerConfigProvider?

```go
func(image string) (cacheKey string) {
  if !a.config.UseManagedIdentityExtension {
    // not using managed identity, use the same credentials for all images
    return ""
  }
  // cache credentials per loginServer
  return a.parseACRLoginServerFromImage(image)
}
```

andyzhangx: that would make the implementation and logic very complicated, shall we go with this PR? Thanks.

liggitt: The current state of the PR seems fragile to me, for several reasons:
1.  It parses the image differently (splitting on / vs parseACRLoginServerFromImage) for purposes of determining if the returned credentials have an entry for the registry.
2. It pins to a credential set length to detect if managed identity is in use, instead of checking the credential provider config
3. It decreases cache hits when multiple registries are used, and it's not clear to me how impactful that is, performance-wise. 
4. It seems to use false-positive methods for determining if a given image is in a given domain (contains azureCRName)

andyzhangx: @liggitt thanks for the review. I moved `ShouldUseCache` func to provider interface, now the logic is more clear:
within one minute cache expiration time, if `d.provider.ShouldUseCache(image, config)` return false (only for Azure Managed Identity), it would merge the two cache maps. This PR has addressed point. 1-4, and only use one expiration date, make the logic more clear, PTAL, thanks!

deads2k:
> /unassign
> /assign @deads2k

Based on my read of the existing code, it seems extremely unusual to have a `CachingDockerConfigProvider` which looks up a new value based on an `image` and does not use the `image` as a key for its cache.  That seems bugged since the API implies that the returned value can vary based on the input.  Before changing this interface further, I would like to see the cache implementation clarified by either not requesting based on image or by caching based on image.  It looked like the bug slipped in here https://github.com/kubernetes/kubernetes/pull/75587/files#diff-82e3d28f25a0fa876c9517c8e8814ca4R115 in 1.15

Let's fix that first caching bug first before changing the `ShouldUseCache` calculation.

andyzhangx:
> > /unassign
> > /assign @deads2k
> 
> Based on my read of the existing code, it seems extremely unusual to have a `CachingDockerConfigProvider` which looks up a new value based on an `image` and does not use the `image` as a key for its cache. That seems bugged since the API implies that the returned value can vary based on the input. Before changing this interface further, I would like to see the cache implementation clarified by either not requesting based on image or by caching based on image. It looked like the bug slipped in here https://github.com/kubernetes/kubernetes/pull/75587/files#diff-82e3d28f25a0fa876c9517c8e8814ca4R115 in 1.15
> 
> Let's fix that first caching bug first before changing the `ShouldUseCache` calculation.

@deads2k the cache is per registry, not per image. e.g. if image is `foo.azurecr.io/image1:v1`, then cache would be `foo.azurecr.io`, when `foo.azurecr.io/image1:v2` comes, it hit the cache, while when `foo2.azurecr.io/image1:v1` comes cache missed. This PR would fix such cache missing issue. I think we can not do a per image cache, which may be too expensive.

deads2k:
> @deads2k the cache is per registry, not per image. e.g. if image is `foo.azurecr.io/image1:v1`, then cache would be `foo.azurecr.io`, when `foo.azurecr.io/image1:v2` comes, it hit the cache, while when `foo2.azurecr.io/image1:v1` comes cache missed. This PR would fix such cache missing issue. I think we can not do a per image cache, which may be too expensive.

Then I would suggest refactoring the code to pass the registry (not the image) to the provider and rewrite the cache to index based on the registry (instead of nothing).  Constructing values based on input and then caching for all possible inputs is a bug.

liggitt:
> Then I would suggest refactoring the code to pass the registry (not the image) to the provider

The image is expected to be opaque, I think...

> rewrite the cache to index based on the registry (instead of nothing).

I agree the cache needs to be indexed by whatever portion of the input the provider found significant. My suggestion of a CacheKey function at https://github.com/kubernetes/kubernetes/pull/92818#discussion_r451653866 would allow that, I think

@andyzhangx, what are your plans for this PR? master will be opening for 1.20 soon, and only blocker bugs for 1.19 are merging at this point.

pkg/credentialprovider/azure/azure_credentials.go
andyzhangx: @liggitt could you review on this PR again? I have followed up with points 1-4, Thanks. 
the cache is per registry, not per image. e.g. if image is `foo.azurecr.io/image1:v1`, then cache would be `foo.azurecr.io`, when `foo.azurecr.io/image1:v2` comes, it hit the cache, while when `foo2.azurecr.io/image1:v1` comes cache missed. This PR would fix such cache missing issue. I think we can not do a per image cache, which may be too expensive.

liggitt: having a `ShouldUseCache(image, config) bool` method combined with a cache that does not differentiate by image does not make sense... 

> the cache is per registry, not per image

That is not reflected in a single boolean return value. I think the signature should be something like 
```
CacheKey(image string) string
```

I also think cache-related methods (e.g. CacheKey or ShouldUseCache) makes more sense as an option to CachingDockerConfigProvider, since it is only used by that and is not inherent to the provider

kbruner:
@andyzhangx @liggitt Can we push to 1.20? We need to clear out open PRs for 1.19 unless they are critical release blockers.

liggitt:
if #92326 is not a regression from 1.18.x, then this doesn't seem release blocking to me

liggitt:
based on the fact that https://github.com/kubernetes/kubernetes/pull/92330 has cherry picks all the way back to [1.16](https://github.com/kubernetes/kubernetes/pull/93314), this doesn't seem like a regression

andyzhangx:
> if #92326 is not a regression from 1.18.x, then this doesn't seem release blocking to me

@liggit thanks, I will try to fix this issue on 1.20

MonzElmasry:

feiskyer:
yes, I think we're still targeting 1.20 for this change.

andyzhangx:
I have modified this PR a little, here are the things I want to clarify:
 - Current provider cache([cacheDockerConfig: map[string]DockerConfigEntry](https://github.com/kubernetes/kubernetes/blob/2570a9d94ced1954edfe4cbf3b84200f126ab717/pkg/credentialprovider/config.go#L50)) is a map (`<registry host name, credential-token>`), e.g. for aws, it's <`foo.ecr.amazonaws.com`,  credential-token>
 - When there are two images (from different ECR repos) pulling requests coming in parallel, currently kubelet will always use the first ECR repo credential:
, e.g. When there are following two images pulling requests coming:
```
foo1.ecr.amazonaws.com/image1:v1
foo2.ecr.amazonaws.com/image2:v1
``` 

The `cacheDockerConfig` could be <`foo1.ecr.amazonaws.com`,  credential-token>, and when `foo2.ecr.amazonaws.com/image2:v1` comes in, it would still return <`foo1.ecr.amazonaws.com`,  credential-token> in current implementation since it would return with cache without image check directly
https://github.com/kubernetes/kubernetes/blob/2570a9d94ced1954edfe4cbf3b84200f126ab717/pkg/credentialprovider/provider.go#L97-L100

So I think this is definitely a bug! This PR just leverages existing `cacheDockerConfig` to solve this bug

cc @andrewsykim since you are already working on external credential provider.

Let me know what I can do to merge this PR into 1.20, thanks everyone!

andrewsykim:
> The d.cacheDockerConfig could be <foo1.ecr.amazonaws.com, credential-token1>, and when foo2.ecr.amazonaws.com/image2:v1 comes in, it would still return <foo1.ecr.amazonaws.com, credential-token1> in current implementation since it would return with cache without image check directly:

Last I checked this wasn't the case, the ECR plugin caches on a per-registry basis and doesn't use CachingDockerProvider for that reason, see:

https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/pkg/credentialprovider/aws/aws_credentials.go#L148-L156

andrewsykim:
The externalized plugin is going to support caching credentials on a per registry basis at the API level by specifying a `cacheKeyType`.

andyzhangx:
FYI @chewong `pull-kubernetes-e2e-aks-engine-azure` test deployment failed

andyzhangx:
> > The d.cacheDockerConfig could be <foo1.ecr.amazonaws.com, credential-token1>, and when foo2.ecr.amazonaws.com/image2:v1 comes in, it would still return <foo1.ecr.amazonaws.com, credential-token1> in current implementation since it would return with cache without image check directly:
> 
> Last I checked this wasn't the case, the ECR plugin caches on a per-registry basis and doesn't use CachingDockerProvider for that reason, see:
> 
> https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/pkg/credentialprovider/aws/aws_credentials.go#L148-L156

hi @andrewsykim I meant in the beginning(first image pulling time), there is only `<foo1.ecr.amazonaws.com, credential-token1>` in ECR cache, and when `foo2.ecr.amazonaws.com/image2:v1` comes in, it would return the cache directly, at that time, `<foo2.ecr.amazonaws.com, credential-token2>` is not cached, that's the case.

if both `foo1.ecr.amazonaws.com` and `foo2.ecr.amazonaws.com` images are pulled on that node before, it won't hit such issue, I suppose following code will be executed by every cloud provider, is that correct?

https://github.com/kubernetes/kubernetes/blob/2570a9d94ced1954edfe4cbf3b84200f126ab717/pkg/credentialprovider/provider.go#L93-L103

andrewsykim:
> if both foo1.ecr.amazonaws.com and foo2.ecr.amazonaws.com images are pulled on that node before, it won't hit such issue, I suppose following code will be executed by every cloud provider, is that correct?

No because the ECR provider does not use CachingDockerConfigProvider:

https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/pkg/credentialprovider/aws/aws_credentials.go#L43-L48

andyzhangx:
> > if both foo1.ecr.amazonaws.com and foo2.ecr.amazonaws.com images are pulled on that node before, it won't hit such issue, I suppose following code will be executed by every cloud provider, is that correct?
> 
> No because the ECR provider does not use CachingDockerConfigProvider:
> 
> https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/pkg/credentialprovider/aws/aws_credentials.go#L43-L48

@andrewsykim thanks for the point, I worked out another PR(https://github.com/kubernetes/kubernetes/pull/96355) to move the fix to Azure component only, that could be more clear.

/hold

andyzhangx:
close this PR now since https://github.com/kubernetes/kubernetes/pull/96355 is already addressing the issue, thanks for review and comments, that really help!

andyzhangx:
this fix "pull image auth error from **multiple**(>1) ACRs using azure managed identity" has been merged in 1.17.14, 1.18.12, 1.19.4, for existing clusters with MSI, wait for a few more tries or using single ACR could workaround

PR #92824, opened 2020-07-06
fix: license issue in blob disk feature

### LINKED ISSUE #92704: Potential license conflict
I ran an OSS license scan of a codebase that has this project as a dependency and came across an interesting hit.

This go-vhd dependency specified here:
https://github.com/kubernetes/legacy-cloud-providers/blob/1bd57a2af8f3a65a82398a87ce5a99f855e277f1/go.mod#L22
has a snippet of code from StackOverflow which is license under creative commons sharealike. Here's the code in question:
https://github.com/rubiojr/go-vhd/blob/0bfd3b39853cdde5762efda92289f14b0ac0491b/vhd/util.go#L59-L73

I'm not a license expert, so I don't know if there are any real issues/conflicts here or not, but wanted to share to see what you all think. 

Here's SO's license info for reference:
https://creativecommons.org/licenses/by-sa/4.0/

My understanding is that the potential issue would be with this clause:
> ShareAlike — If you remix, transform, or build upon the material, you must distribute your contributions under the same license as the original.

which could potentially require go-vhd and projects that depend on it to use the Creative Commons license as well.

andyzhangx:
**What type of PR is this?**
/kind cleanup

This PR fixed the potential license issue mentioned in 92704.

The problematic code was removed upstream - https://github.com/rubiojr/go-vhd/issues/3, https://github.com/rubiojr/go-vhd/pull/4, https://github.com/rubiojr/go-vhd/pull/5

Fixes #92704

cc @nikhita @dims 
```
hack/update-vendor.sh
```

```release-note
fix: license issue in blob disk feature
```

```docs
none
```


vendor/github.com/rubiojr/go-vhd/vhd/util.go
```diff
@@ -55,19 +52,3 @@ func uuidToBytes(uuid string) []byte {
 
 	return h
 }
-
```
andyzhangx, line 58: @dims that code with license issue is removed in this PR

vendor/github.com/rubiojr/go-vhd/vhd/vhd.go
```diff
@@ -324,8 +327,14 @@ func (vhd *VHD) PrintExtraHeader() {
 	fmtField("Parent timestamp", fmt.Sprintf("%s", t))
 
 	fmtField("Reserved", hexs(header.Reserved[:]))
-	parentName := utf16BytesToString(header.ParentUnicodeName[:],
-		binary.BigEndian)
+	parentNameBytes, _, err := transform.Bytes(
+		unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM).NewDecoder(),
```
liggitt, line 331: looks like this only handles BigEndian... whereas utf16BytesToString claimed to handle big or little endian... is BigEndian always correct here?

vendor/github.com/rubiojr/go-vhd/vhd/vhd.go
```diff
@@ -324,8 +327,14 @@ func (vhd *VHD) PrintExtraHeader() {
...
+		header.ParentUnicodeName[:],
+	)
+	if err != nil {
+		panic(err)
```
liggitt, line 335: where does this data come from? does this open up a DOS vector where invalid data can cause the controller manager to panic?

vendor/github.com/rubiojr/go-vhd/vhd/vhd.go
andyzhangx: cc @rubiojr

andyzhangx:
@chewong @jackfrancis Are you aware of this test failure?
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92824/pull-kubernetes-e2e-aks-engine-azure/1280130243327168512

```
• Failure [74.430 seconds]
[k8s.io] [sig-node] Events
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/framework/framework.go:592
  should be sent by kubelets and the scheduler about pods scheduling and running  [Conformance] [It]
  /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/framework/framework.go:597
  Jul  6 13:56:56.969: Unexpected error:
      
      <*errors.errorString | 0xc00009de50>: {
          s: "timed out waiting for the condition",
      }
      timed out waiting for the condition
  occurred
  /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/node/events.go:97
------------------------------
{"msg":"FAILED [k8s.io] [sig-node] Events should be sent by kubelets and the scheduler about pods scheduling and running  [Conformance]","total":-1,"completed":9,"skipped":184,"failed":1,"failures":["[k8s.io] [sig-node] Events should be sent by kubelets and the scheduler about pods scheduling and running  [Conformance]"]}
skipped 207 lines unfold_more
[Fail] [k8s.io] [sig-node] Events [It] should be sent by kubelets and the scheduler about pods scheduling and running  [Conformance] 
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/node/events.go:97
Ran 285 of 5208 Specs in 437.211 seconds
FAIL! -- 284 Passed | 1 Failed | 0 Pending | 4923 Skipped
```

vendor/github.com/rubiojr/go-vhd/vhd/vhd.go
rubiojr: Yes, only BigEndian is required, according to the VHD specs. Even if the function that was removed handled both, it was hardcoded to BigEndian previously also, so no behavior change here.

vendor/github.com/rubiojr/go-vhd/vhd/vhd.go
rubiojr: > does this open up a DOS vector where invalid data can cause the controller manager to panic?

I'm not familiar with Kubernetes so I can't answer that but:

* The data comes from a VHD header, so a malformed VHD header could trigger that. This helper doesn't really validate the VHD header format in any way so it'd be trivial to cause that even without the changes reviewed here.
* It's a helper function used by the CLI (`go-vhd info`) to print VHD header values to stdout.

Hope that helps.

liggitt: that's helpful, thanks. I double checked and kube doesn't call PrintInfo() or PrintExtraHeader() anywhere

jackfrancis:
@andyzhangx Not aware, but this change makes sense

andyzhangx:
@dims rebased, PTAL, thanks.

PR #46266, opened 2017-05-23
kube-proxy: ratelimit runs of iptables by sync-period flags

thockin:
This bounds how frequently iptables can be synced.  It will be no more often than every 10 seconds and no less often than every 1 minute, by default.

@timothysc FYI

@dcbw @freehan FYI

PR #36332, opened 2016-11-07
Default kube-proxy to the old behavior for proxier sync.

MrHohn:
Fix #36281.

This PR defaults `minSyncPeriod` to 0 and makes kube-proxy fall back to the old behavior to fix the immediate problem #36266.

@bprashanth

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kubernetes/36332)

PR #92836, opened 2020-07-06
kube-proxy iptables min-sync-period default 1sec

aojea:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup

> /kind deprecation
> /kind design
> /kind documentation

/kind failing-test
> /kind feature

/kind flake

Currently kube-proxy defaults the min-sync-period for
iptables to 0. However, as explained by Dan Winship,

"With minSyncPeriod: 0, you run iptables-restore 100 times.
 With minSyncPeriod: 1s , you run iptables-restore once.
 With minSyncPeriod: 10s , you also run iptables-restore once,
 but you might have to wait 10 seconds first"

Fixes #

This is causing issues in KIND jobs, when there are multiple endpoints updates, the kube-proxy fails to acquire the iptables lock

```release-note
kube-proxy iptables min-sync-period defaults to 1 sec. Previously, it was 0.
```

```docs

```

### DIFF
--- a/pkg/proxy/apis/config/scheme/testdata/KubeProxyConfiguration/after/v1alpha1.yaml
+++ b/pkg/proxy/apis/config/scheme/testdata/KubeProxyConfiguration/after/v1alpha1.yaml
@@ -21,7 +21,7 @@ hostnameOverride: ""
 iptables:
   masqueradeAll: false
   masqueradeBit: 14
-  minSyncPeriod: 0s
+  minSyncPeriod: 1s
   syncPeriod: 30s
 ipvs:
   excludeCIDRs: null
--- a/pkg/proxy/apis/config/scheme/testdata/KubeProxyConfiguration/roundtrip/default/v1alpha1.yaml
+++ b/pkg/proxy/apis/config/scheme/testdata/KubeProxyConfiguration/roundtrip/default/v1alpha1.yaml
@@ -21,7 +21,7 @@ hostnameOverride: ""
 iptables:
   masqueradeAll: false
   masqueradeBit: 14
-  minSyncPeriod: 0s
+  minSyncPeriod: 1s
   syncPeriod: 30s
 ipvs:
   excludeCIDRs: null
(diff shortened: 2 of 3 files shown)


danwinship:
@BenTheElder pointed out that there had been a default originally (`2s`), but it was removed in #36332. But it was reverted because it was causing serious problems, and following links from there, it seems like this was just because the original implementation of `minSyncPeriod` was broken. (eg, it blocked the informer thread while waiting until it was allowed to do an update, rather than doing the updates in a different thread)

When Tim wrote `BoundedFrequencyRunner` in #46266 he set a `10s` default for certain cluster templates/setup scripts (eg `cluster/gce/gci/configure-helper.sh`) but didn't change the command-line/config default.

I think `1s` is a much better default value than either `0s` or `10s`; it works much more smoothly than `0s` in the case where you get a bunch of updates all at once (eg, deleting a large deployment), but unlike `10s` it's not _noticeably_ slower when you have steady-but-not-overwhelming updates.

Is it the very best possible value for every cluster? Would `2s` be better than `1s`? I dunno. But `10s`, while maybe appropriate for some clusters, seems way too high as a default, given that most(?) people are currently surviving with `0s`. (That said, this PR does not change the `minSyncPeriod` for any of the cases that are currently setting it to `10s`, since they're overriding the default anyway.)

/lgtm

thockin:
It's been a while, but ISTR deciding that 10s was a compromise
decision.  We had lots of reports of people whose iptables runs were
4-5 seconds each, so setting 10s and burst=2 targetted 5s.  If you
have a steady stream of changes, this will devolve into 1 run per 10s.

I didn't want to do a dynamic controller without REALLY needing to.

If we're going to lower this, we should probably track how long
iptables runs are actually taking and use that to guide this value
dynamically, so we can't pile up requests.  We also don't want to run
iptables-restore continuously.

danwinship:
> We had lots of reports of people whose iptables runs were 4-5 seconds each

And most of that time was probably spent waiting to get the iptables lock. We've made huge improvements to the iptables situation in the last few years. I don't think old data points are valid any more. But I don't currently have any new data points to offer...

> If we're going to lower this

Note that this PR does not lower the value for anyone; it _raises_ the default value from `0` to `1s` for people who are not currently explicitly setting it.

thockin (commented):
I think I considered this but thought maybe it would break someone.  In hindsight, that's pretty clearly the wrong choice.

/lgtm
/approve

PR #92838, opened 2020-07-06
Fix memory leak in endpointSliceTracker

### LINKED ISSUE #92837: endpointSliceTracker memory leak
**What happened**:
The endpointSliceTracker creates a set of resource versions for each service, the resource versions in the set could be deleted when endpointslices are deleted, but the set and its key in the map is never deleted, leading to memory leak.
https://github.com/kubernetes/kubernetes/blob/a26e5881d8aa38aaa0a56c78050fd4ea14611c9f/pkg/controller/endpointslice/endpointslice_tracker.go#L38
```
Showing top 10 nodes out of 59
      flat  flat%   sum%        cum   cum%
     32770 21.66% 21.66%      32770 21.66%  k8s.io/kubernetes/pkg/controller/endpointslice.(*endpointSliceTracker).relatedResourceVersions
     32768 21.66% 43.33%      32768 21.66%  k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/apis/meta/v1.(*ObjectMeta).Unmarshal
     16384 10.83% 54.16%      16384 10.83%  fmt.Errorf
     16384 10.83% 64.99%      16384 10.83%  k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/runtime/schema.GroupVersion.String
     16384 10.83% 75.82%      16384 10.83%  k8s.io/kubernetes/vendor/k8s.io/client-go/rest/watch.NewDecoder
     10930  7.23% 83.05%      10930  7.23%  k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/runtime.(*Scheme).AddKnownTypeWithName
     10923  7.22% 90.27%      10923  7.22%  k8s.io/kubernetes/vendor/k8s.io/api/core/v1.(*ObjectReference).Unmarshal
      5462  3.61% 93.88%      16386 10.83%  k8s.io/kubernetes/pkg/controller/endpointslice.(*endpointSliceTracker).Update
      5461  3.61% 97.49%       5461  3.61%  k8s.io/kubernetes/vendor/golang.org/x/net/http2.awaitRequestCancel
      2521  1.67% 99.16%       2521  1.67%  k8s.io/kubernetes/pkg/controller/garbagecollector.(*concurrentUIDToNode).Write
```

The set of resource versions should be delete from endpointSliceTracker when the service is deleted.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.18.5

tnqn:
**What type of PR is this?**
/kind bug

endpointSliceTracker creates a set of resource versions for each service, the resource versions in the set could be deleted when endpointslices are deleted, but the set and its key in the map is never deleted, leading to memory leak.

This patch deletes the set if the service is deleted, and stops initializing an empty set when "read-only" methods "Has" and "Stale" are called.

Fixes #92837

```release-note
Fixed memory leak in endpointSliceTracker
```

```docs

```

### DIFF
--- a/pkg/controller/endpointslice/endpointslice_tracker.go
+++ b/pkg/controller/endpointslice/endpointslice_tracker.go
@@ -51,8 +51,11 @@ func (est *endpointSliceTracker) Has(endpointSlice *discovery.EndpointSlice) boo
 	est.lock.Lock()
 	defer est.lock.Unlock()
 
-	rrv := est.relatedResourceVersions(endpointSlice)
-	_, ok := rrv[endpointSlice.Name]
+	rrv, ok := est.relatedResourceVersions(endpointSlice)
+	if !ok {
+		return false
+	}
+	_, ok = rrv[endpointSlice.Name]
 	return ok
 }
 
@@ -63,7 +66,10 @@ func (est *endpointSliceTracker) Stale(endpointSlice *discovery.EndpointSlice) b
 	est.lock.Lock()
 	defer est.lock.Unlock()
 
-	rrv := est.relatedResourceVersions(endpointSlice)
+	rrv, ok := est.relatedResourceVersions(endpointSlice)
+	if !ok {
+		return true
+	}
 	return rrv[endpointSlice.Name] != endpointSlice.ResourceVersion
 }
 
@@ -73,8 +79,21 @@ func (est *endpointSliceTracker) Update(endpointSlice *discovery.EndpointSlice) 
 	est.lock.Lock()
 	defer est.lock.Unlock()
 
-	rrv := est.relatedResourceVersions(endpointSlice)
+	rrv, ok := est.relatedResourceVersions(endpointSlice)
+	if !ok {
+		rrv = endpointSliceResourceVersions{}
+		est.resourceVersionsByService[getServiceNN(endpointSlice)] = rrv
+	}
 	rrv[endpointSlice.Name] = endpointSlice.ResourceVersion
+}
+
+// DeleteService removes the set of resource versions tracked for the Service.
+func (est *endpointSliceTracker) DeleteService(namespace, name string) {
+	est.lock.Lock()
+	defer est.lock.Unlock()
+
+	serviceNN := types.NamespacedName{Name: name, Namespace: namespace}
+	delete(est.resourceVersionsByService, serviceNN)
 }
 
 // Delete removes the resource version in this endpointSliceTracker for the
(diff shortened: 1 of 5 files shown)


pkg/controller/endpointslice/endpointslice_tracker_test.go
```diff
@@ -47,30 +48,35 @@ func TestEndpointSliceTrackerUpdate(t *testing.T) {
 		checksParam *discovery.EndpointSlice
 		expectHas   bool
 		expectStale bool
+		expectedRRV int
```
robscott, line 51: This naming is confusing to me. Maybe `expectResourceVersions` or something similarly pluralized to indicate this is about a quantity of resource versions and not the value of the resource version?

pkg/controller/endpointslice/endpointslice_tracker.go
```diff
@@ -73,33 +79,45 @@ func (est *endpointSliceTracker) Update(endpointSlice *discovery.EndpointSlice)
 	est.lock.Lock()
 	defer est.lock.Unlock()
 
-	rrv := est.relatedResourceVersions(endpointSlice)
+	rrv, _ := est.relatedResourceVersions(endpointSlice, true)
```
robscott, line 82: Maybe instead of adding the extra bool param here, we could just add a `if !ok` and do the create here.

pkg/controller/endpointslice/endpointslice_tracker_test.go
robscott: Alternatively if there's a reliable way to predict resource versions it might be good to actually ensure that this has the appropriate value for the resource version instead of just checking the quantity.

robscott (commented):
Thanks for finding and fixing this! This mostly lgtm, just had a couple nits.

pkg/controller/endpointslice/endpointslice_tracker.go
tnqn: Done

pkg/controller/endpointslice/endpointslice_tracker_test.go
tnqn: Thanks for the suggestion, updated to compare the entire resource versions map.

tnqn (commented):
Thanks @robscott for your quick review!

pkg/controller/endpointslice/endpointslice_tracker.go
```diff
@@ -73,33 +79,42 @@ func (est *endpointSliceTracker) Update(endpointSlice *discovery.EndpointSlice)
...
 	serviceNN := getServiceNN(endpointSlice)
 	vers, ok := est.resourceVersionsByService[serviceNN]
-
-	if !ok {
-		vers = endpointSliceResourceVersions{}
-		est.resourceVersionsByService[serviceNN] = vers
-	}
-
-	return vers
+	return vers, ok
```
robscott, line 117: Maybe just return on the line above - `return est.resourceVersionsByService[serviceNN]`

tnqn: It's a single value when returning the map value, would cause compilation error "not enough arguments to return".

robscott: Woops, somehow I hadn't run into that before, thanks for the explanation!

robscott:
Thanks!

/lgtm
/retest
/assign @freehan

freehan (approved):
/lgtm
/approve

PR #92839, opened 2020-07-06
AWS cloudprovider allow external management of LB types nlb-ip and external

kishorj:
**What type of PR is this?**
/kind feature

Currently, for Service type LoadBalancer, the service annotation service.beta.kubernetes.io/aws-load-balancer-type causes creation of a Network Load Balancer if the value is "nlb", and a Classic Load Balancer in all other cases. This PR modifies the AWS cloudprovider to return no-op for nlb-ip and external LB types. Here is the proposed interpretation -

- If value is "nlb" provision a Network Load Balancer

- If value is "nlb-ip" or "external", cloudprovider ignore management of the underlying LB

- For all other cases, provision a Classic Load Balancer

This  provides a mechanism to manage Load Balancers external to the AWS cloud provider if a different behavior is required.

Fixes #

Tested via kops that the AWS cloudprovider provisions NLB/CLB as expected for the supported values and no LB gets created if the annotation aws-load-balancer-type is nlb-ip or external.

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
+++ b/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
@@ -3678,6 +3678,9 @@ func buildListener(port v1.ServicePort, annotations map[string]string, sslPorts 
 // EnsureLoadBalancer implements LoadBalancer.EnsureLoadBalancer
 func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiService *v1.Service, nodes []*v1.Node) (*v1.LoadBalancerStatus, error) {
 	annotations := apiService.Annotations
+	if isLBExternal(annotations) {
+		return nil, cloudprovider.ImplementedElsewhere
+	}
 	klog.V(2).Infof("EnsureLoadBalancer(%v, %v, %v, %v, %v, %v, %v)",
 		clusterName, apiService.Namespace, apiService.Name, c.region, apiService.Spec.LoadBalancerIP, apiService.Spec.Ports, annotations)
 
@@ -3689,7 +3692,6 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
 	if len(apiService.Spec.Ports) == 0 {
 		return nil, fmt.Errorf("requested load balancer with no ports")
 	}
-
 	// Figure out what mappings we want on the load balancer
 	listeners := []*elb.Listener{}
 	v2Mappings := []nlbPortMapping{}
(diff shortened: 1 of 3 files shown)


staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
```diff
@@ -4523,6 +4532,9 @@ func (c *Cloud) UpdateLoadBalancer(ctx context.Context, clusterName string, serv
 	if err != nil {
 		return err
 	}
+	if isLBExternal(service.Annotations) {
```
M00nF1sh, line 4535: this should be the first statement :D

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
```diff
@@ -3690,6 +3690,9 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
 		return nil, fmt.Errorf("requested load balancer with no ports")
 	}
 
+	if isLBExternal(annotations) {
```
M00nF1sh, line 3693: this should be the first statement

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
```diff
@@ -4076,6 +4079,9 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
 // GetLoadBalancer is an implementation of LoadBalancer.GetLoadBalancer
 func (c *Cloud) GetLoadBalancer(ctx context.Context, clusterName string, service *v1.Service) (*v1.LoadBalancerStatus, bool, error) {
 	loadBalancerName := c.GetLoadBalancerName(ctx, clusterName, service)
+	if isLBExternal(service.Annotations) {
+		return nil, false, nil
```
M00nF1sh, line 4083: this should be the first statement

M00nF1sh (commented):
shouldn't "elb" be "clb"? 
https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/introduction.html
alb/nlb/clb are all elb :D

staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
kishorj: done

kishorj:
> shouldn't "elb" be "clb"?
> https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/introduction.html
> alb/nlb/clb are all elb :D

Modified to clb.

adammw:
Great minds think alike, @kishorj I've just been discussing a proposal over on Slack to add a generic annotation for all services to allow other out-of-tree controllers to be used rather than something specific to the AWS NLB implementation, are you able to join in the conversation? 

https://kubernetes.slack.com/archives/C09QYUH5W/p1598558133130400

kishorj:
Thanks, I would definitely join, but I don’t have access to the Kubernetes slack.

Regards,

-Kishor

ltagliamonte-dd:
@adammw @kishorj could we backport this to 1.8? We are using kops and 1.8 is the latest available version.

PR #92842, opened 2020-07-06
Deprecated API request audit annotation

liggitt:
**What type of PR is this?**
/kind feature

Completes beta portion of https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1693-warnings / https://github.com/kubernetes/enhancements/issues/1693

```release-note
Audit events for API requests to deprecated API versions now include a `"k8s.io/deprecated": "true"` audit annotation. If a target removal release is identified, the audit event includes a `"k8s.io/removal-release": "<majorVersion>.<minorVersion>"` audit annotation as well.
```

/sig api-machinery
/cc @deads2k

### DIFF
--- a/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/customresource_handler.go
+++ b/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/customresource_handler.go
@@ -329,10 +329,8 @@ func (r *crdHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	if utilfeature.DefaultFeatureGate.Enabled(features.WarningHeaders) {
-		for _, w := range crdInfo.warnings[requestInfo.APIVersion] {
-			warning.AddWarning(req.Context(), "", w)
-		}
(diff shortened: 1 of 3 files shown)


staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/customresource_handler.go
```diff
@@ -329,10 +332,9 @@ func (r *crdHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	if utilfeature.DefaultFeatureGate.Enabled(features.WarningHeaders) {
```
deads2k, line 332: Why are we no longer feature gating this?

liggitt: moved to construction time, rather than request serve time

deads2k: > moved to construction time, rather than request serve time (`Move feature gate check to crdInfo construction time`)

ok

PR #92627, opened 2020-06-30
Add nodeHasControlPlanePods()

oomichi:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

DeprecatedMightBeMasterNode() has been marked as deprecated and we need to
find alternative way for callers of the function.
In NewResourceUsageGatherer(), the function was called for distinguishing
the specified pods are running on master nodes, and the gatherer gathers
those pods' resource usage.
This adds isNodeHaveControlPlanePods() to gistinguish the specified pods
are running on nodes which are operating control plane pods (kube-scheduler
and kube-controller-manager) and replace callers of DeprecatedMightBeMasterNode()
with this new function as better way.

Ref: https://github.com/kubernetes/kubernetes/issues/80177

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #59761, opened 2018-02-12
Removing Production Code usage from vSphere E2E tests

divyenpatel:
**What this PR does / why we need it**: 

vSphere E2E test cases uses VCP production code for bootstrap logic, example is to get an instance of vSphere struct. Once it gets an instance vSphere struct, it will use this object to call into methods in VCP code like - DiskIsAttached(), DisksAreAttached() etc.

This direct dependency on the VCP production should be removed as any changes to the production code would create a problem in E2E test cases as well.

This PR addressed above stated need.

**Which issue(s) this PR fixes** 
Fixes # https://github.com/vmware/kubernetes/issues/434

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #92845, opened 2020-07-06
Remove DeprecatedMightBeMasterNode()

### LINKED ISSUE #80177: e2e: remove usage of the DeprecatedMightBeMasterNode function from the e2e testing framework

oomichi:
**What type of PR is this?**

/kind cleanup

This removes DeprecatedGetMasterAndWorkerNodes() usage from vsphere e2e test as deprecated function cleanup.
Then all callers of DeprecatedMightBeMasterNode() have been removed.
So this removes DeprecatedMightBeMasterNode() itself also.

Fixes: https://github.com/kubernetes/kubernetes/issues/80177

```release-note
NONE
```


oomichi:
This PR can contain removal of the deprecated function itself if https://github.com/kubernetes/kubernetes/pull/92627 is merged before this.

oomichi:
need to rebase after merging https://github.com/kubernetes/kubernetes/pull/92627

/hold

oomichi:
/cc @neolit123 

This can close long-term issue https://github.com/kubernetes/kubernetes/issues/80177

test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go
```diff
@@ -211,6 +207,8 @@ var _ = utils.SIGDescribe("Storage Policy Based Volume Provisioning [Feature:vsp
 		scParameters[Datastore] = vsanDatastore
 		framework.Logf("Invoking test for SPBM storage policy: %+v", scParameters)
 		kubernetesClusterName := GetAndExpectStringEnvVar(KubernetesClusterName)
+		masterNode, err := getControlPlaneNode(client)
```
neolit123, line 210: we should rename to `controlPlaneNode` here.

test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go
```diff
@@ -341,3 +339,22 @@ func invokeStaleDummyVMTestWithStoragePolicy(client clientset.Interface, masterN
...
+func getControlPlaneNode(client clientset.Interface) (string, error) {
+	regKubeScheduler := regexp.MustCompile("kube-scheduler-.*")
+	regKubeControllerManager := regexp.MustCompile("kube-controller-manager-.*")
+
+	podList, err := client.CoreV1().Pods(metav1.NamespaceSystem).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return "", err
+	}
+	if len(podList.Items) < 1 {
+		return "", fmt.Errorf("Can't find any pods in namespace %s to grab metrics from", metav1.NamespaceSystem)
```
neolit123, line 352: errors should start lowercase.
in this case the message should be different i guess? because we are not getting this node for metrics?

> "could not find kube-scheduler and kube-controller-manager pods in the %s namespace", ....

test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go
```diff
@@ -341,3 +339,22 @@ func invokeStaleDummyVMTestWithStoragePolicy(client clientset.Interface, masterN
...
+	}
+	for _, pod := range podList.Items {
+		if regKubeScheduler.MatchString(pod.Name) || regKubeControllerManager.MatchString(pod.Name) {
+			return pod.Spec.NodeName, nil
+		}
+	}
+	return "", fmt.Errorf("Can't find any nodes where control plane pods are running")
```
neolit123, line 359: > could not find any nodes where control-plane pods are running

test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go
```diff
@@ -341,3 +339,22 @@ func invokeStaleDummyVMTestWithStoragePolicy(client clientset.Interface, masterN
 	framework.ExpectNoError(err)
 	framework.ExpectEqual(isVMPresentFlag, false, errorMsg)
 }
+
+func getControlPlaneNode(client clientset.Interface) (string, error) {
```
neolit123, line 343: with the prior PRs, we now seem to have this similar function in a number of places (but with different error messages, context i guess)
should it have been generalized and placed in a shared location?

test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go
```diff
@@ -211,6 +207,8 @@ var _ = utils.SIGDescribe("Storage Policy Based Volume Provisioning [Feature:vsp
 		scParameters[Datastore] = vsanDatastore
 		framework.Logf("Invoking test for SPBM storage policy: %+v", scParameters)
 		kubernetesClusterName := GetAndExpectStringEnvVar(KubernetesClusterName)
+		masterNode, err := getControlPlaneNode(client)
+		framework.ExpectNoError(err)
 		invokeStaleDummyVMTestWithStoragePolicy(client, masterNode, namespace, kubernetesClusterName, scParameters)
```
neolit123, line 212: i don't really understand why this function really needs a "master" node, though.
it just ends up calling:

```
nodeInfo := TestContext.NodeMapper.GetNodeInfo(masterNode)
```

test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go
```diff
@@ -341,3 +339,22 @@ func invokeStaleDummyVMTestWithStoragePolicy(client clientset.Interface, masterN
 	framework.ExpectNoError(err)
 	framework.ExpectEqual(isVMPresentFlag, false, errorMsg)
 }
+
+func getControlPlaneNode(client clientset.Interface) (string, error) {
+	regKubeScheduler := regexp.MustCompile("kube-scheduler-.*")
+	regKubeControllerManager := regexp.MustCompile("kube-controller-manager-.*")
```
neolit123, line 345: as mentioned above, i don't understand why this test needs a control-plane node (so maybe we have to clear this topic first).

but a meta-question here is - if a node in a certain setup is running the scheduler and the KCM, but not the kube-apiserver (GCP used to do that) can we still assume that this is a "control-plane" node?

test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go
oomichi: OK, let me make error message lowercase.

> "could not find any control-plane pods in the %s namespace", ....

The above error case is for not getting any pods in the namespace before checking the pods are control plane.

test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go
oomichi: done

oomichi (commented):
@neolit123 Thanks for your review.
I will update the PR based on your comments.

test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go
neolit123: ok, SGTM.

test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go
neolit123: @divyenpatel

you added the "master" node related logic here:
https://github.com/kubernetes/kubernetes/pull/59761
the PR and tracking issue did not explain directly why we need to detect if a node is "master" or not.

any comments before we merge the change in this PR?

neolit123:
@oomichi 
this LGTM, thanks.

i will add a hold but feel free to remove it unless @divyenpatel replies within a couple of days on this topic:
https://github.com/kubernetes/kubernetes/pull/92845#discussion_r452463745

/lgtm
/hold

oomichi:
@neolit123 
Thanks for your review again.
It is nice to wait for original author feedback to understand it.

test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go
divyenpatel: @oomichi 

> TBH I am still not sure what is this test purpose.
> Especially what is invokeStaleDummyVMTestWithStoragePolicy() doing.

With in-tree VCP, when the volume is created using storage policy, shadow (dummy) VM is getting created and deleted to apply SPBM policy on the volume.

https://github.com/kubernetes/legacy-cloud-providers/blob/master/vsphere/vclib/diskmanagers/vmdm.go#L157-L161

But when for some reason if VM deletion fails, we just print an error message and continue with provisioning workflow.

We have a routine in VCP which runs every 5 minutes and helps clean up these leftover VMs. 
https://github.com/kubernetes/legacy-cloud-providers/blob/master/vsphere/vsphere_util.go#L382-L414

The test is helping to validate the functionality of this routine.

neolit123: @divyenpatel does it really need a "master" node?

https://github.com/kubernetes/kubernetes/blob/5e5defb9b5df92fccc2a2a8efbb956c6c58ae38f/test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go#L337-L338

is `DataCenterRef` only set for "master" nodes?

divyenpatel: @neolit123 

Basically we need a datacenter specified in the Workspace, thats where dummy VM is getting created.

```
[Global]
user = "administrator@vsphere.local"
password = "password"
port = "443"
insecure-flag = "1"

[VirtualCenter "10.0.1.200"]
datacenters = "DC-1"

[Workspace]
server = "10.0.1.200"
datacenter = "DC-1"
default-datastore = "vsanDatastore"
resourcepool-path = "ClusterNameHere/Resources"
folder = "kubernetes"

[Disk]
scsicontrollertype = pvscsi
```

[Workspace] - Defines the temporary location of shadow VMs provisioned to create volumes

    server is the vCenter Server IP or FQDN
    datacenter - the Datacenter to provision temporary VMs to for volume provisioning

Refer - https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/existing.html

neolit123: are you happy with the change in this PR to change the logic of detecting "master" (control-plane) nodes?

divyenpatel: We do not need to get master node for `invokeStaleDummyVMTestWithStoragePolicy` test. We need to read `vsphere.conf` file supplied in the env variable and get workspace datacenter.

Internally in the test, this is how it is used - https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go#L339-L340

This is wrong. 

The test will be able to validate functionality in the case when the master node is present on the datacenter configured in the workspace.

but since this test issue was already present, I will not block this PR.

@oomichi Can you put a note in the doc for this test- "invokeStaleDummyVMTestWithStoragePolicy assumes master node is present on the datacenter specified in the workspace section of vsphere.conf file"

oomichi: @divyenpatel 
> With in-tree VCP, when the volume is created using storage policy, shadow (dummy) VM is getting created and deleted to apply SPBM policy on the volume.

Thanks for your explanation. I got the point.

> @oomichi Can you put a note in the doc for this test- "invokeStaleDummyVMTestWithStoragePolicy assumes master node is present on the datacenter specified in the workspace section of vsphere.conf file"

Is it fine to put the above note on this PR message?
Or should we put it on `invokeStaleDummyVMTestWithStoragePolicy()` itself?

divyenpatel: would be good to put this as func doc for `invokeStaleDummyVMTestWithStoragePolicy`.

oomichi: @divyenpatel Thanks, I got it.

test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go
```diff
@@ -309,7 +307,9 @@ func invokeInvalidPolicyTestNeg(client clientset.Interface, namespace string, sc
 	return fmt.Errorf("Failure message: %+q", eventList.Items[0].Message)
 }
 
-func invokeStaleDummyVMTestWithStoragePolicy(client clientset.Interface, masterNode string, namespace string, clusterName string, scParameters map[string]string) {
+// invokeStaleDummyVMTestWithStoragePolicy assumes control plane node(aka. master node) is present on the datacenter specified in the workspace section of vsphere.conf file.
```
neolit123, line 310: we should remove references of the word "master" from the code base, so here we can remove "(aka. master node)"

test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go
```diff
@@ -309,7 +307,9 @@ func invokeInvalidPolicyTestNeg(client clientset.Interface, namespace string, sc
...
+// With in-tree VCP, when the volume is created using storage policy, shadow (dummy) VM is getting created and deleted to apply SPBM policy on the volume.
```
neolit123, line 311: AFAIK the word "dummy" can also be considered offensive, but given this is bound deeper in the test, you can leave it for now.

for reference there is a new WG that is being formed to clear some of this language, so i won't be surprised that this test is revisited in the near future:
https://github.com/kubernetes/community/pull/4884

alternatively you could replace the word "dummy" with something like "test" in this PR.

test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go
oomichi: OK, done.

test/e2e/storage/vsphere/vsphere_volume_vsan_policy.go
oomichi: Thanks, that is a nice point.
But if renaming this dummy to the other word like "test", it would be a little difficult to understand what that means because the function contains many "dummy"s.
So I guess it is nice to have another PRs to update such words for the other places.

neolit123 (commented):
/lgtm
/hold cancel

PR #92846, opened 2020-07-06
Remove pvcLister from genericScheduler

Huang-Wei:
**What type of PR is this?**

/kind cleanup
/sig scheduling

With recent code refactoring, PVCLister can be fetched from sharedInformerFactory, so don't need to be a field in genericScheduler.

Fixes #

```release-note
NONE
```


ahg-g:
We should keep the convention of only accessing Lister() only before the informer factory started. I think we should keep the current code of taking a reference of the lister when creating the generic scheduler instance.

Huang-Wei:
> We should keep the convention of accessing Lister() only before the informer factory started. I think we should keep the current code of taking a reference of the lister when creating the generic scheduler instance.

That makes sense.

Given that `podPassesBasicChecks()` (which uses g.pvcLister) is just checking pod's PVC, IMO it's not a must to run at the beginning of Schedule(). Instead, I'd prefer to move it to a new plugin (or merge into an existing general plugin) and it should implement the PreFilter() extension point, then we can keep a pvcLister member field there to ensure it's initialized. WDYT?

ahg-g:
oh, if that is all it does, then we should move it to the VolumeBinding PreFilter plugin.

Huang-Wei:
Exactly, VolumeBinding is a perfect fit.

However, one thing worth noting is that if the pvc's presence check gets moved, `g.snpashot()` will be called unconditionally - this would result in extra overhead for the cases that a Pod's PVC is not present. @ahg-g I think we should hold this for now. And then in 1.20, craft a benchmark test to evaluate the impact. Or, maybe by then, we have a refined flow to avoid the snapshotting overhead.

pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -704,19 +708,14 @@ func TestGenericScheduler(t *testing.T) {
...
 			scheduler := NewGenericScheduler(
 				cache,
 				snapshot,
 				[]framework.Extender{},
-				pvcLister,
 				false,
 				schedulerapi.DefaultPercentageOfNodesToScore)
 			result, err := scheduler.Schedule(context.Background(), prof, framework.NewCycleState(), test.pod)
-			if !reflect.DeepEqual(err, test.wErr) {
+			if err != test.wErr && err.Error() != test.wErr.Error() {
```
chendave, line 718: off the topic though, but this change is nice!
this fix the warning in my IDE:
```avoid using reflect.DeepEqual with errorsdeepequalerrors```

maybe we can do similar thing there,
https://github.com/kubernetes/kubernetes/blob/c46c1c043ee8de4c021c03ddb0fb157d1c1ba3f8/pkg/scheduler/scheduler_test.go#L345

ahg-g:
> Exactly, VolumeBinding is a perfect fit.
> 
> However, one thing worth noting is that if the pvc's presence check gets moved, `g.snpashot()` will be called unconditionally - this would result in extra overhead for the cases that a Pod's PVC is not present. @ahg-g I think we should hold this for now. And then in 1.20, craft a benchmark test to evaluate the impact. Or, maybe by then, we have a refined flow to avoid the snapshotting overhead.

I think we should move to PreFilter, we do many other checks, this is no different. I am not concerned about snapshot overhead in the case of race conditions between pvc and pod creation because snapshot has been optimized compared to the past.

Huang-Wei:
@ahg-g @cofyc this is ready for review. Basically this PR does two things:

- remove pvcLister from genericScheduler; instead, initialize pvcLister inside VolumeBinding plugin
- due to the removal of pvcLister from genericScheduler, podPassesBasicChecks() has to be moved into VolumeBinding#PreFilter.

cofyc:
/lgtm
Great!

ahg-g:
/milestone v1.20

one nit, otherwise looks good to me!

PR #91408, opened 2020-05-25
Add seccomp GA version skew for pods

saschagrunert:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:
    
This adds a new conversion function to the pod strategy to handle the
seccomp version skew strategy:
    
https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20190717-seccomp-ga.md#version-skew-strategy
   
**Which issue(s) this PR fixes**:

None

**Special notes for your reviewer**:

Requires https://github.com/kubernetes/kubernetes/pull/91381

/cc @pjbgf @hasheddan @evrardjp

**Does this PR introduce a user-facing change?**:

```release-note
- Added pod version skew strategy for seccomp profile to synchronize the deprecated annotations with the new API Server fields. Please see the corresponding section [in the KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20190717-seccomp-ga.md#version-skew-strategy) for more detailed explanations. 
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs
- KEP: https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20190717-seccomp-ga.md
```

Issue #91286, opened 2020-05-20
Graduate Seccomp to GA


pjbgf:
Graduate seccomp to GA, as defined on the respective [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20190717-seccomp-ga.md). This is an umbrella issue to track all outstanding work.

### Core Items
- [x] API Server changes https://github.com/kubernetes/kubernetes/pull/91408 (MERGED)
- [x] Kubelet changes https://github.com/kubernetes/kubernetes/pull/91381 (MERGED)
- [x] Mark flag `--seccomp-profile-root` as deprecated https://github.com/kubernetes/kubernetes/pull/91182 (MERGED)
- [x] Update PSP to support seccomp fields https://github.com/kubernetes/kubernetes/pull/92856 (MERGED)
- [x] Document Kubernetes seccomp support https://github.com/kubernetes/website/pull/21278 (MERGED)

### Nice to Have on 1.19
- [x] Add seccomp least privilege for docker sandbox https://github.com/kubernetes/kubernetes/pull/90948 (MERGED)
- [x] Add seccomp least privilege for kuberuntime https://github.com/kubernetes/kubernetes/pull/90949 (MERGED)
- [x] E2E tests review https://github.com/kubernetes/kubernetes/pull/91442 (LGTM)

### Pushed to 1.22
- [x] Add Warning mechanism for deprecated features usage (based off [KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1693-warnings))

/area kubelet
/area kubelet-api
/sig node
/area security
/assign

pjbgf:
@palnabarun just a heads up that we have concluded all the core items for seccomp to go GA. There are still 2 pending PRs that ideally would also go in 1.19, but we are waiting for them to be reviewed: https://github.com/kubernetes/kubernetes/pull/90948 and https://github.com/kubernetes/kubernetes/pull/90949.

kikisdeliveryservice:
https://github.com/kubernetes/kubernetes/pull/92856#issuecomment-655950700 needs to be rebased apparently and isn't just waiting in the merge pool as verify is not passing.

saschagrunert:
@pjbgf I think the warning mechanism was blocked by a not existing API. This is now tracked in https://github.com/kubernetes/kubernetes/issues/94626

pjbgf:
[...]

PR #92856, opened 2020-07-07
Implement PodSecurityPolicy enforcement for seccomp GA

saschagrunert:
**What type of PR is this?**

/kind feature

This implements the necessary pieced for the PodSecurityPolicy
enforcement like described in the appropriate KEP section:

https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20190717-seccomp-ga.md#podsecuritypolicy-enforcement

Refers to #91286 

Based on #91408

```release-note
NONE
```

```docs
- KEP: https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20190717-seccomp-ga.md
```

### DIFF
--- a/pkg/registry/core/pod/strategy.go
+++ b/pkg/registry/core/pod/strategy.go
@@ -587,7 +587,7 @@ func applySeccompVersionSkew(pod *api.Pod) {
 
 	// sync field and annotation
 	if hasField && !hasAnnotation {
-		newAnnotation := seccompAnnotationForField(field)
+		newAnnotation := podutil.SeccompAnnotationForField(field)
 
 		if newAnnotation != "" {
 			if pod.Annotations == nil {
@@ -596,7 +596,7 @@ func applySeccompVersionSkew(pod *api.Pod) {
 			pod.Annotations[v1.SeccompPodAnnotationKey] = newAnnotation
 		}
 	} else if hasAnnotation && !hasField {
-		newField := seccompFieldForAnnotation(annotation)
+		newField := podutil.SeccompFieldForAnnotation(annotation)
 
 		if newField != nil {
 			if pod.Spec.SecurityContext == nil {
@@ -621,7 +621,7 @@ func applySeccompVersionSkew(pod *api.Pod) {
 
 			// sync field and annotation
 			if hasField && !hasAnnotation {
-				newAnnotation := seccompAnnotationForField(field)
+				newAnnotation := podutil.SeccompAnnotationForField(field)
 
 				if newAnnotation != "" {
 					if pod.Annotations == nil {
@@ -630,7 +630,7 @@ func applySeccompVersionSkew(pod *api.Pod) {
 					pod.Annotations[key] = newAnnotation
 				}
 			} else if hasAnnotation && !hasField {
-				newField := seccompFieldForAnnotation(annotation)
+				newField := podutil.SeccompFieldForAnnotation(annotation)
 
 				if newField != nil {
 					if ctr.SecurityContext == nil {
@@ -642,59 +642,4 @@ func applySeccompVersionSkew(pod *api.Pod) {
 
 			return true
 		})
-}
-
-// seccompFieldForAnnotation takes a pod seccomp profile field and returns the
-// converted annotation value
-func seccompAnnotationForField(field *api.SeccompProfile) string {
-	// If only seccomp fields are specified, add the corresponding annotations.
-	// This ensures that the fields are enforced even if the node version
-	// trails the API version
-	switch field.Type {
-	case api.SeccompProfileTypeUnconfined:
-		return v1.SeccompProfileNameUnconfined
-
-	case api.SeccompProfileTypeRuntimeDefault:
-		return v1.SeccompProfileRuntimeDefault
-
-	case api.SeccompProfileTypeLocalhost:
-		if field.LocalhostProfile != nil {
-			return v1.SeccompLocalhostProfileNamePrefix + *field.LocalhostProfile
-		}
-	}
-
-	// we can only reach this code path if the LocalhostProfile is nil but the
-	// provided field type is SeccompProfileTypeLocalhost or if an unrecognized
-	// type is specified
-	return ""
-}
-
-// seccompFieldForAnnotation takes a pod annotation and returns the converted
-// seccomp profile field.
-func seccompFieldForAnnotation(annotation string) *api.SeccompProfile {
-	// If only seccomp annotations are specified, copy the values into the
-	// corresponding fields. This ensures that existing applications continue
-	// to enforce seccomp, and prevents the kubelet from needing to resolve
-	// annotations & fields.
-	if annotation == v1.SeccompProfileNameUnconfined {
-		return &api.SeccompProfile{Type: api.SeccompProfileTypeUnconfined}
-	}
-
-	if annotation == api.SeccompProfileRuntimeDefault || annotation == api.DeprecatedSeccompProfileDockerDefault {
-		return &api.SeccompProfile{Type: api.SeccompProfileTypeRuntimeDefault}
-	}
-
-	if strings.HasPrefix(annotation, v1.SeccompLocalhostProfileNamePrefix) {
-		localhostProfile := strings.TrimPrefix(annotation, v1.SeccompLocalhostProfileNamePrefix)
-		if localhostProfile != "" {
-			return &api.SeccompProfile{
-				Type:             api.SeccompProfileTypeLocalhost,
-				LocalhostProfile: &localhostProfile,
-			}
-		}
-	}
-
-	// we can only reach this code path if the localhostProfile name has a zero
-	// length or if the annotation has an unrecognized value
-	return nil
 }
--- a/pkg/security/podsecuritypolicy/seccomp/strategy.go
+++ b/pkg/security/podsecuritypolicy/seccomp/strategy.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	podutil "k8s.io/kubernetes/pkg/api/pod"
 	api "k8s.io/kubernetes/pkg/apis/core"
 )
 
@@ -83,6 +84,10 @@ func (s *strategy) Generate(annotations map[string]string, pod *api.Pod) (string
 		// Profile already set, nothing to do.
 		return annotations[api.SeccompPodAnnotationKey], nil
 	}
+	if pod.Spec.SecurityContext != nil && pod.Spec.SecurityContext.SeccompProfile != nil {
+		// Profile field already set, translate to annotation
+		return podutil.SeccompAnnotationForField(pod.Spec.SecurityContext.SeccompProfile), nil
+	}
 	return s.defaultProfile, nil
 }
 
@@ -92,6 +97,10 @@ func (s *strategy) ValidatePod(pod *api.Pod) field.ErrorList {
 	allErrs := field.ErrorList{}
 	podSpecFieldPath := field.NewPath("pod", "metadata", "annotations").Key(api.SeccompPodAnnotationKey)
 	podProfile := pod.Annotations[api.SeccompPodAnnotationKey]
+	// if the annotation is not set, see if the field is set and derive the corresponding annotation value
+	if len(podProfile) == 0 && pod.Spec.SecurityContext != nil && pod.Spec.SecurityContext.SeccompProfile != nil {
+		podProfile = podutil.SeccompAnnotationForField(pod.Spec.SecurityContext.SeccompProfile)
+	}
 
 	if !s.allowAnyProfile && len(s.allowedProfiles) == 0 && podProfile != "" {
 		allErrs = append(allErrs, field.Forbidden(podSpecFieldPath, "seccomp may not be set"))
@@ -141,9 +150,19 @@ func (s *strategy) profileAllowed(profile string) bool {
 
 // profileForContainer returns the container profile if set, otherwise the pod profile.
 func profileForContainer(pod *api.Pod, container *api.Container) string {
+	if container.SecurityContext != nil && container.SecurityContext.SeccompProfile != nil {
+		// derive the annotation value from the container field
+		return podutil.SeccompAnnotationForField(container.SecurityContext.SeccompProfile)
+	}
 	containerProfile, ok := pod.Annotations[api.SeccompContainerAnnotationKeyPrefix+container.Name]
 	if ok {
+		// return the existing container annotation
 		return containerProfile
 	}
+	if pod.Spec.SecurityContext != nil && pod.Spec.SecurityContext.SeccompProfile != nil {
+		// derive the annotation value from the pod field
+		return podutil.SeccompAnnotationForField(pod.Spec.SecurityContext.SeccompProfile)
+	}
+	// return the existing pod annotation
 	return pod.Annotations[api.SeccompPodAnnotationKey]
 }
--- a/pkg/security/podsecuritypolicy/seccomp/strategy_test.go
+++ b/pkg/security/podsecuritypolicy/seccomp/strategy_test.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"testing"
 
+	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	api "k8s.io/kubernetes/pkg/apis/core"
 )
@@ -40,6 +41,9 @@ var (
 	}
 	allowSpecific = map[string]string{
 		AllowedProfilesAnnotationKey: "foo",
+	}
+	allowSpecificLocalhost = map[string]string{
+		AllowedProfilesAnnotationKey: v1.SeccompLocalhostProfileNamePrefix + "foo",
 	}
 )
 
@@ -102,9 +106,11 @@ func TestNewStrategy(t *testing.T) {
 }
 
 func TestGenerate(t *testing.T) {
+	bar := "bar"
 	tests := map[string]struct {
 		pspAnnotations  map[string]string
 		podAnnotations  map[string]string
+		seccompProfile  *api.SeccompProfile
 		expectedProfile string
 	}{
 		"no seccomp, no pod annotations": {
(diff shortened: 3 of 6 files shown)


pkg/security/podsecuritypolicy/seccomp/strategy_test.go
```diff
@@ -161,49 +170,169 @@ func TestValidatePod(t *testing.T) {
...
+		},
+		"pod field and no annotations, no required profiles": {
+			pspAnnotations: withoutSeccomp,
+			podAnnotations: nil,
+			seccompProfile: &api.SeccompProfile{
+				Type:             api.SeccompProfileTypeLocalhost,
+				LocalhostProfile: &foo,
+			},
+			expectedError: "Forbidden: seccomp may not be set",
+		},
```
hasheddan, line 292: Does this mean that if a PSP exists and it does not specify seccomp annotations that seccomp may not be used on pods at all?

saschagrunert: Hm, this test cases were there before, so I think the validation should not be triggered at all if seccomp annotations are not available. Not exactly sure about this one…

pjbgf: Very good point @hasheddan. The current documentation says that: 

> seccomp.security.alpha.kubernetes.io/allowedProfileNames - Annotation that specifies which values are allowed for the pod seccomp annotations. Specified as a comma-delimited list of allowed values. Possible values are those listed above, plus * to allow all profiles. **Absence of this annotation means that the default cannot be changed**.

So my take is that when `allowedProfileNames` is set to `nil` all workloads must use the profile set in `defaultProfileName`. If `defaultProfileName` is set to `unconfined`, this combination disables seccomp altogether. It may be worth changing a bit the test description to make that clearer.

pkg/security/podsecuritypolicy/seccomp/strategy.go
```diff
@@ -110,22 +140,55 @@ func (s *strategy) ValidatePod(pod *api.Pod) field.ErrorList {
...
-	if !s.profileAllowed(containerProfile) {
-		msg := fmt.Sprintf("%s is not an allowed seccomp profile. Valid values are %v", containerProfile, s.allowedProfilesString)
-		allErrs = append(allErrs, field.Forbidden(fieldPath, msg))
+	// Annotation based validation
+	if hasAnnotation && !s.profileAllowed(profileAnnotation) {
+		allErrs = append(allErrs, field.Forbidden(annotationFieldPath, s.notAllowedProfileMsg(profileAnnotation)))
+	}
+
+	// Field based validation
+	securityContextFldPath := field.NewPath("pod", "spec", "containers[*]", "securityContext", "seccompProfile", "type")
```
pjbgf, line 167: This may not always be correct, we need to account for `initContainers` too. Ideally this would have a similar interface as `VisitContainersWithPath`, receiving the field path as parameter.

saschagrunert: We do not have something like this at hand, right? Or do you see any utility method which can be used?

pkg/security/podsecuritypolicy/seccomp/strategy_test.go
saschagrunert: Hm alright I just added a note that this combination disables the usage of seccomp.

saschagrunert:
/assign @liggitt

kindly asking for review.

pkg/security/podsecuritypolicy/seccomp/strategy_test.go
hasheddan: @pjbgf @saschagrunert it makes sense to me that absence of `allowedProfileNames` means that you cannot change the default, but it also looks like the absence of both `allowedProfileNames` and `defaultProfileName` means that you cannot use seccomp profiles at all, which feels like it might not be the behavior we want considering that we can't require that these annotations be provided on a PSP like we can for some fields (i.e. `runAsUser`, `fsGroup`, etc.)

pjbgf: So basically seccomp cannot be enabled if the PSP has:

`allowedProfileNames == nil && defaultProfileName == nil`
`allowedProfileNames == nil && defaultProfileName == 'unconfined'`
`allowedProfileNames == 'unconfined' && defaultProfileName == nil`
`allowedProfileNames == 'unconfined' && defaultProfileName == 'unconfined'`

I personally think we are better off documenting this and leaving as it is. 
Then group all changes we need from PSP into a new KEP. What do you think?

liggitt:
Can we base this on https://github.com/kubernetes/kubernetes/pull/91408, and expose and use `seccompAnnotationForField`? I think that would let us do something much more minimal like this:

```diff
diff --git a/pkg/security/podsecuritypolicy/seccomp/strategy.go b/pkg/security/podsecuritypolicy/seccomp/strategy.go
index 0c8d3faf9e5..83d024dce74 100644
--- a/pkg/security/podsecuritypolicy/seccomp/strategy.go
+++ b/pkg/security/podsecuritypolicy/seccomp/strategy.go
@@ -83,6 +83,10 @@ func (s *strategy) Generate(annotations map[string]string, pod *api.Pod) (string
 		// Profile already set, nothing to do.
 		return annotations[api.SeccompPodAnnotationKey], nil
 	}
+	if pod.Spec.SecurityContext != nil && pod.Spec.SecurityContext.SeccompProfile != nil {
+		// Profile field already set, translate to annotation
+		return SeccompAnnotationForField(pod.Spec.SecurityContext.SeccompProfile), nil
+	}
 	return s.defaultProfile, nil
 }
 
@@ -92,6 +96,10 @@ func (s *strategy) ValidatePod(pod *api.Pod) field.ErrorList {
 	allErrs := field.ErrorList{}
 	podSpecFieldPath := field.NewPath("pod", "metadata", "annotations").Key(api.SeccompPodAnnotationKey)
 	podProfile := pod.Annotations[api.SeccompPodAnnotationKey]
+	// if the annotation is not set, see if the field is set and derive the corresponding annotation value
+	if len(podProfile) == 0 && pod.Spec.SecurityContext != nil && pod.Spec.SecurityContext.SeccompProfile != nil {
+		podProfile = SeccompAnnotationForField(pod.Spec.SecurityContext.SeccompProfile)
+	}
 
 	if !s.allowAnyProfile && len(s.allowedProfiles) == 0 && podProfile != "" {
 		allErrs = append(allErrs, field.Forbidden(podSpecFieldPath, "seccomp may not be set"))
@@ -141,9 +149,19 @@ func (s *strategy) profileAllowed(profile string) bool {
 
 // profileForContainer returns the container profile if set, otherwise the pod profile.
 func profileForContainer(pod *api.Pod, container *api.Container) string {
+	if container.SecurityContext != nil && container.SecurityContext.SeccompProfile != nil {
+		// derive the annotation value from the container field
+		return SeccompAnnotationForField(container.SecurityContext.SeccompProfile)
+	}
 	containerProfile, ok := pod.Annotations[api.SeccompContainerAnnotationKeyPrefix+container.Name]
 	if ok {
+		// return the existing container annotation
 		return containerProfile
 	}
+	if pod.Spec.SecurityContext != nil && pod.Spec.SecurityContext.SeccompProfile != nil {
+		// derive the annotation value from the pod field
+		return SeccompAnnotationForField(pod.Spec.SecurityContext.SeccompProfile)
+	}
+	// return the existing pod annotation
 	return pod.Annotations[api.SeccompPodAnnotationKey]
 }
```

pkg/security/podsecuritypolicy/seccomp/strategy_test.go
liggitt: > it also looks like the absence of both `allowedProfileNames` and `defaultProfileName` means that you cannot use seccomp profiles at all, which feels like it might not be the behavior we want

we should not change existing PSP behavior in this PR

hasheddan: @liggitt would updating the PSP to have seccomp fields in spec be possible at all in the future, or is this API considered too mature at this point?

liggitt: > @liggitt would updating the PSP to have seccomp fields in spec be possible at all in the future, or is this API considered too mature at this point?

That's not clear at this point. It's less a question of maturity and more a question about the future of the PSP API and whether it is worth the effort/complexity of migrating an annotation to a field.

hasheddan: Understood, thanks for the clarification!

saschagrunert: Sounds good, let's resolve this discussion for sanity reasons. :)

pkg/security/podsecuritypolicy/seccomp/strategy.go
saschagrunert: This discussion is now obsolete because of the changed implementation approach.

pkg/security/podsecuritypolicy/seccomp/strategy_test.go
```diff
@@ -21,25 +21,31 @@ import (
...
 
+	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	api "k8s.io/kubernetes/pkg/apis/core"
 )
 
 var (
-	withoutSeccomp    = map[string]string{"foo": "bar"}
+	foo               = "foo"
+	bar               = "bar"
```
liggitt, line 31: I would revert this and use the literal strings... the mental overhead of tracking down the values in the variables here isn't really worth it

pkg/security/podsecuritypolicy/seccomp/strategy_test.go
```diff
@@ -21,25 +21,31 @@ import (
...
+	withoutSeccomp    = map[string]string{foo: bar}
```
liggitt, line 32: it's confusing to use foo and bar here... just do `"some-annotation":"some-value"`?

pkg/security/podsecuritypolicy/seccomp/strategy_test.go
```diff
@@ -127,26 +133,26 @@ func TestGenerate(t *testing.T) {
...
+				api.SeccompPodAnnotationKey: bar,
 			},
-			expectedProfile: "bar",
+			expectedProfile: bar,
 		},
 	}
 	for k, v := range tests {
 		s := NewStrategy(v.pspAnnotations)
-		actual, err := s.Generate(v.podAnnotations, nil)
+		actual, err := s.Generate(v.podAnnotations, &api.Pod{})
```
liggitt, line 155: add a `*SeccompProfile` field to the test case and pass that in here, then define the test cases that check what generate returns when the field is set and the annotation isn't:
```suggestion
		actual, err := s.Generate(v.podAnnotations, &api.Pod{
			Spec: api.PodSpec{
				SecurityContext: &api.PodSecurityContext{
					SeccompProfile: v.seccompProfile,
				},
			},
		})
```

pkg/security/podsecuritypolicy/seccomp/strategy_test.go
```diff
@@ -161,57 +167,90 @@ func TestValidatePod(t *testing.T) {
...
 			expectedError:  "",
 		},
+		"valid pod annotations and field, required profiles": {
+			pspAnnotations: allowSpecific,
+			podAnnotations: map[string]string{
+				api.SeccompPodAnnotationKey: foo,
+			},
+			seccompProfile: &api.SeccompProfile{
+				Type:             api.SeccompProfileTypeLocalhost,
+				LocalhostProfile: &foo,
```
liggitt, line 230: using "foo" here is confusing... it has no relationship to the pod-level foo annotation. set to some other value

pkg/security/podsecuritypolicy/seccomp/strategy_test.go
```diff
@@ -161,57 +167,90 @@ func TestValidatePod(t *testing.T) {
...
+		"valid pod field and no annotation, required profiles": {
+			pspAnnotations: allowSpecific,
+			podAnnotations: nil,
+			seccompProfile: &api.SeccompProfile{
+				Type:             api.SeccompProfileTypeLocalhost,
+				LocalhostProfile: &foo,
+			},
+			expectedError: "Forbidden: localhost/foo is not an allowed seccomp profile. Valid values are foo",
+		},
 	}
```
liggitt, line 243: add another testcase where just the field is set and is allowed (`pspAnnotations: allowSpecificLocalhost,`)

saschagrunert: Done :checkered_flag:

pkg/security/podsecuritypolicy/seccomp/strategy_test.go
saschagrunert: Reverted that change as suggested. :+1:

pkg/security/podsecuritypolicy/seccomp/strategy_test.go
saschagrunert: Done :+1:

pkg/security/podsecuritypolicy/seccomp/strategy_test.go
saschagrunert: Reverted that change as suggested. :+1:

pkg/security/podsecuritypolicy/seccomp/strategy_test.go
saschagrunert: Reverted that change.

pkg/security/podsecuritypolicy/seccomp/strategy_test.go
```diff
@@ -102,51 +106,75 @@ func TestNewStrategy(t *testing.T) {
...
 	tests := map[string]struct {
 		pspAnnotations  map[string]string
 		podAnnotations  map[string]string
+		seccompProfile  *api.SeccompProfile
 		expectedProfile string
 	}{
 		"no seccomp, no pod annotations": {
 			pspAnnotations:  withoutSeccomp,
 			podAnnotations:  nil,
+			seccompProfile:  nil,
```
liggitt, line 119: this is nil by default, can omit lines like this

saschagrunert: Fixed :+1:

pkg/security/podsecuritypolicy/seccomp/strategy_test.go
```diff
@@ -158,60 +186,103 @@ func TestGenerate(t *testing.T) {
...
 	tests := map[string]struct {
 		pspAnnotations map[string]string
 		podAnnotations map[string]string
+		seccompProfile *api.SeccompProfile
 		expectedError  string
 	}{
 		"no pod annotations, required profiles": {
 			pspAnnotations: allowSpecific,
 			podAnnotations: nil,
+			seccompProfile: nil,
```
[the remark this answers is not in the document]
saschagrunert: Fixed :+1:

pkg/security/podsecuritypolicy/seccomp/strategy_test.go
```diff
@@ -293,6 +350,14 @@ func TestValidateContainer(t *testing.T) {
...
 		},
+		"valid container field and no annotation, required profiles": {
+			pspAnnotations: allowSpecificLocalhost,
+			seccompProfile: &api.SeccompProfile{
+				Type:             api.SeccompProfileTypeLocalhost,
+				LocalhostProfile: &foo,
+			},
+			expectedError: "",
+		},
 	}
```
liggitt, line 361: add a negative test that makes sure a container localhost profile field of bar would be forbidden

saschagrunert: Added as suggested :+1:

saschagrunert:
I'll fix the linter issue after rebasing on top of the latest master if https://github.com/kubernetes/kubernetes/pull/91408 got merged.

liggitt:
go ahead and fix the lint and import issue and repush. If we can get CI green on this PR it can merge in place of 91408

saschagrunert:
> go ahead and fix the lint and import issue and repush. If we can get CI green on this PR it can merge in place of 91408

Done :+1:

saschagrunert:
/retest
"Pod pending timeout"

saschagrunert:
Rebased on top of the latest master branch

liggitt:
The import-boss failure looks legitimate in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/92856/pull-kubernetes-verify/1281538127155957761

pkg/registry/core/pod/strategy.go
```diff
@@ -644,9 +644,9 @@ func applySeccompVersionSkew(pod *api.Pod) {
 		})
 }
 
-// seccompFieldForAnnotation takes a pod seccomp profile field and returns the
+// SeccompAnnotationForField takes a pod seccomp profile field and returns the
 // converted annotation value
-func seccompAnnotationForField(field *api.SeccompProfile) string {
+func SeccompAnnotationForField(field *api.SeccompProfile) string {
```
liggitt, line 649: Moving this function to a location like https://github.com/kubernetes/kubernetes/blob/master/pkg/api/pod/util.go might simplify the package dependency graph. Run the import verification scripts to make sure our import restrictions are up to date so verify CI will pass.

saschagrunert: Thanks for the hint, moved it into the pod utils and updated the import paths.

liggitt:
Thanks. Did `hack/verify-import-boss.sh` pass locally?

saschagrunert:
> Thanks. Did `hack/verify-import-boss.sh` pass locally?

Unfortunately it does not, we have to find another location for the utility methods:
```
errors in package "k8s.io/kubernetes/test/e2e/framework/providers/kubemark":
the following imports did not match any allowed prefix:
  k8s.io/kubernetes/pkg/api/pod


!!! Error in ./hack/verify-import-boss.sh:46
  Error in ./hack/verify-import-boss.sh:46. '$(kube::util::find-binary "import-boss") --include-test-files=true --verify-only --input-dirs "$(IFS=, ; echo "${packages[*]}")"' exited with status 1
Call stack:
  1: ./hack/verify-import-boss.sh:46 main(...)
Exiting with status 1
```

saschagrunert:
Would `staging/src/k8s.io/apimachinery/pkg/util/seccomp` a sufficient location?

liggitt:
You can add that package to the list of allowed imports for kubemark

PR #92811, opened 2020-07-05
[WIP][TEST} debug kubeproxy iptables lock

aojea:

PR #92860, opened 2020-07-07
iptables: don't do reverse DNS lookups

aojea:
**What type of PR is this?**
/kind bug

the iptables monitor was using iptables -L to list the chains,
without the -n option, so it was trying to do reverse DNS lookups.

However, this is not a big impact in Kubernetes because it's only being used
for checking the CANARY chains that are always empty.

Fixes #

This caused multiple issues on KIND jobs, don't know exactly what it was more evident there, maybe because of the DNS
resolution is slower :shrug: 

```release-note
NONE
```

### DIFF
--- a/pkg/util/iptables/iptables.go
+++ b/pkg/util/iptables/iptables.go
@@ -607,6 +607,9 @@ func (runner *runner) chainExists(table Table, chain Chain) (bool, error) {
 	runner.mu.Lock()
 	defer runner.mu.Unlock()
 
+	trace := utiltrace.New("iptables Monitor CANARY check")
+	defer trace.LogIfLong(2 * time.Second)
+
 	_, err := runner.run(opListChain, fullArgs)
 	return err == nil, err
 }
@@ -617,7 +620,7 @@ const (
 	opCreateChain operation = "-N"
 	opFlushChain  operation = "-F"
 	opDeleteChain operation = "-X"
-	opListChain   operation = "-L"
+	opListChain   operation = "-S"
 	opAppendRule  operation = "-A"
 	opCheckRule   operation = "-C"
 	opDeleteRule  operation = "-D"


aojea:
:/

still locking errors in kind jobs :thinking: 
```
2020-07-07T12:06:20.382538442Z stderr F E0707 12:06:20.368219       1 proxier.go:858] Failed to ensure that filter chain KUBE-SERVICES exists: error creating chain "KUBE-SERVICES": exit status 4: Another app is currently holding the xtables lock; still 4s 100000us time ahead to have a chance to grab the lock...
2020-07-07T12:06:20.38259444Z stderr F Another app is currently holding the xtables lock; still 3s 100000us time ahead to have a chance to grab the lock...
2020-07-07T12:06:20.382603635Z stderr F Another app is currently holding the xtables lock; still 2s 100000us time ahead to have a chance to grab the lock...
2020-07-07T12:06:20.382609803Z stderr F Another app is currently holding the xtables lock; still 1s 100000us time ahead to have a chance to grab the lock...
2020-07-07T12:06:20.382615753Z stderr F Another app is currently holding the xtables lock; still 0s 100000us time ahead to have a chance to grab the lock...
2020-07-07T12:06:20.382624343Z stderr F Another app is currently holding the xtables lock. Stopped waiting after 5s.
```

pkg/util/iptables/iptables.go
```diff
@@ -617,7 +617,7 @@ const (
 	opCreateChain operation = "-N"
 	opFlushChain  operation = "-F"
 	opDeleteChain operation = "-X"
-	opListChain   operation = "-L"
+	opListChain   operation = "-L -n"
```
danwinship, line 620: This won't work... it will pass "`-L -n`" as a single argument, not as two separate arguments.

If you look at [the logs](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/92860/pull-kubernetes-e2e-gce/1280466272315772929/artifacts/e2e-3450d3ab9d-674b9-minion-group-21s9/kube-proxy.log) you'll see that it repeatedly claims that the canary was deleted and the iptables rules need to be recreated... because it mistakes the "syntax error" status code from iptables as meaning "could not list the chain because it doesn't exist".

We should be able to just pass "`-n`" unconditionally (ie, add it in `runContext` when constructing `fullArgs`)

danwinship (changes requested):
doh

BenTheElder:
Nice find :-)

pkg/util/iptables/iptables.go
aojea: > We should be able to just pass "-n" unconditionally (ie, add it in runContext when constructing fullArgs)

it fails

I0707 15:53:41.987391       1 bounded_frequency_runner.go:279] sync-runner: retrying in 59.999931677s
I0707 15:53:41.987600       1 proxier.go:845] Syncing iptables rules
E0707 15:53:41.989259       1 proxier.go:858] Failed to ensure that filter chain KUBE-EXTERNAL-SERVICES exists: error creating chain "KUBE-EXTERNAL-SERVICES": exit status 2: iptables v1.8.3 (legacy): Illegal option `-n' with this command

aojea:
this is still having lock errors 
/hold
/retest

thockin (commented):
Even if this doesn't fix everything it still seems like a good fix, right?  Is it cleaner to put it here or in `chainExists` ?  This seems fine to me - the only reason not might be if we WANTED to do lookups, but I just can't see that happening.

I'll hold in case you want to do try more things, but this one-liner seems good.

/approve
/hold

aojea:
>  Is it cleaner to put it here or in chainExists ?
I've tried adding the `-n` argument in different ways, but were not working as expected, the kind jobs were reporting errors with the locks and the GCE timing out :shrug: 

This seems the best option, now the kind job doesn't report any error about the lock.

Will run it a few times more before unholding to be completely sure

/retest

BenTheElder:
/retest
unrelated CI flakiness (typecheck should not have failed for example, it's just too slow)

aojea:
seems is working better but there is still contention with the iptables lock. I have a PR with more debugging https://github.com/kubernetes/kubernetes/pull/92811, and I can see that the problem happens when there are a lot of updates.
All the iptables operations has the wait flag, that means that all of them will be waiting for the lock during 5 seconds by default and we also run 2 monitor loops (kubelet and kubeproxy). At least the monitor don't need to wait for the lock, because it will retry later, don't know if we can fail fast with other operations too

aojea:
This looks much better now, there are no errors about holding the lock, only the monitor complain because it can not check the iptables canary, but that should not be a problem because it has a retry loop and is not getting in the path of kube-proxy

```
2020-07-08T07:12:57.431984105Z stderr F W0708 07:12:57.431333       1 iptables.go:573] Could not check for iptables canary mangle/KUBE-PROXY-CANARY: exit status 4
2020-07-08T07:13:57.327007289Z stderr F W0708 07:13:57.326787       1 iptables.go:573] Could not check for iptables canary mangle/KUBE-PROXY-CANARY: exit status 4
```

pkg/util/iptables/iptables.go
```diff
@@ -601,13 +601,17 @@ func (runner *runner) Monitor(canary Chain, tables []Table, reloadFunc func(), i
 
 // chainExists is used internally by Monitor; none of the public Interface methods can be
 // used to distinguish "chain exists" from "chain does not exist" with no side effects
+// it does not wait to acquire the iptables lock, avoiding to lock the runner unnecessarely
 func (runner *runner) chainExists(table Table, chain Chain) (bool, error) {
+	iptablesCmd := iptablesCommand(runner.protocol)
 	fullArgs := makeFullArgs(table, chain)
+	fullArgs = append([]string{string(opListChain)}, fullArgs...)
 
 	runner.mu.Lock()
```
aojea, line 610: another  option I think may work is not using the runner lock here and rely on the iptables lock. I prefer this one, this operation should not take long now, we fail fast and still control from the runner the iptables operations. 
also the iptables-nft does not have this external lock , so don't know how this will look like in the future :shrug:

pkg/util/iptables/iptables.go
```diff
@@ -601,13 +601,17 @@ func (runner *runner) Monitor(canary Chain, tables []Table, reloadFunc func(), i
...
 	defer runner.mu.Unlock()
+	klog.V(5).Infof("running iptables: %s %v", iptablesCmd, fullArgs)
+	_, err := runner.exec.Command(iptablesCmd, fullArgs...).CombinedOutput()
```
aojea, line 613: instead of adding an `if op == opListchain` in the `runContext()` method we execute the iptables command directly from here, I think this is cleaner because this `chainExist` method is only used by the monitor

pkg/util/iptables/iptables.go
```diff
@@ -617,7 +621,7 @@ const (
 	opCreateChain operation = "-N"
 	opFlushChain  operation = "-F"
 	opDeleteChain operation = "-X"
-	opListChain   operation = "-L"
+	opListChain   operation = "-S"
```
aojea, line 624: `-S` is better than `-nL` and I think is doing the same, without the dns reverse lookups

aojea:
so, this is my theory:
The `chainExists()` method was taking too long because it was doing DNS reverse lookups. It runs in a loop by the  iptables `Monitor()`, both in kube-proxy and the kubelet.
Because we use the iptables wait option, the iptables operations were waiting to acquire the lock, and holding the runner lock too, adding more delays.
This should not be a problem for the kube-proxy operations because they should run sequentially, but the `Monitor` runs in its own go routine, stepping on the kube-proxy sync operations, if we add that we didn't have a minSyncPeriod in KIND because we use the k8s default, when there are a lot of changes in the endpoints and services this is causing more retries and more contention.

/retest
/hold cancel
PTAL again

pkg/util/iptables/iptables.go
danwinship: So there's some trickiness here which is that iptables-nft 1.8.3 is not consistent with iptables-legacy about error codes in some edge cases. So you need to confirm that `iptables-nft -S NON-EXISTENT-CHAIN` and `iptables-legacy -S NON-EXISTENT-CHAIN` return the same thing... (Originally the monitoring code had used `-F` to check if the chain existed (since it's empty anyway, flushing would be a no-op if the chain existed, but would return an error if it didn't). But that didn't work because iptables-nft didn't actually return an error if the chain didn't exist..)

pkg/util/iptables/iptables.go
danwinship: I don't think there's any need to rewrite this function. If it takes long to run then that's a problem, and it's good that it causes problems in that case, because then we'll notice the problem and fix it.

aojea:
/retest
there are still errors about the lock, but I think that those are related to KIND and its CNI

pkg/util/iptables/iptables.go
danwinship: Actually... I don't think `iptables.runner` needs a mutex at all. None of its own fields are mutable after creation, and we don't need to serialize calls to the iptables binary, because it has its own lock.

pkg/util/iptables/iptables.go
thockin: Minor, but what about a flag to `run` that said Wait or DontWait ?

pkg/util/iptables/iptables.go
aojea: I left it as is, adding a trace so it logs the time to execute

pkg/util/iptables/iptables.go
aojea: seems is consistent, and is what is being used in go-iptables
```
root@kind-control-plane:/# iptables-legacy
iptables v1.8.3 (legacy): no command specified
Try `iptables -h' or 'iptables --help' for more information.
root@kind-control-plane:/# iptables-legacy -t nat -L NON
iptables: No chain/target/match by that name.
root@kind-control-plane:/# echo $?
1
root@kind-control-plane:/# iptables-nft
iptables v1.8.3 (nf_tables): no command specified
Try `iptables -h' or 'iptables --help' for more information.
root@kind-control-plane:/# iptables-nft -t nat -L NON
# Warning: iptables-legacy tables present, use iptables-legacy to see them
iptables: No chain/target/match by that name.
root@kind-control-plane:/# echo $?
1
```

pkg/util/iptables/iptables.go
aojea: it turns out this is not the root cause of the problems, not waiting here alleviates the situation because the runner is not locked for 5 seconds but is not causing the lock to be hold forever 
will left that for another time or if someone want to revisit this, seems that Dan has some ideas :smile: 
https://github.com/kubernetes/kubernetes/pull/92860#discussion_r451556669

aojea:
sorry for the noise, this was a red herring, I updated the description

> However, this is not a big impact in Kubernetes because it's only being used
> for checking the CANARY chains that are always empty.

pkg/util/iptables/iptables.go
danwinship: >     root@kind-control-plane:/# iptables-nft -t nat -L NON

`-L` is not `-S` :slightly_smiling_face:

aojea:
```
root@kind-control-plane:/#  iptables-legacy -t nat -S NON
iptables: No chain/target/match by that name.
root@kind-control-plane:/# echo $?
1
root@kind-control-plane:/# iptables-nft -t nat -S NON
# Warning: iptables-legacy tables present, use iptables-legacy to see them
iptables: No chain/target/match by that name.
root@kind-control-plane:/# echo $?
1
```
:facepalm:

thockin (commented):
/lgtm

BenTheElder:
https://storage.googleapis.com/k8s-gubernator/triage/index.html?test=DisruptionController%20should%20block%20an%20eviction%20until%20the%20PDB%20is%20updated%20to%20allow%20it

this test case broke :|

BenTheElder:
filed https://github.com/kubernetes/kubernetes/issues/92977, all e2e testing seems to be affected by a bad change to `[sig-apps] DisruptionController should block an eviction until the PDB is updated to allow it`

BenTheElder:
test fix just merged, next retest should pick it up (the current run started before).

Issue #88634, opened 2020-02-27
Update scheduler extender API to be better defined


damemi:
As noted in https://github.com/kubernetes/kubernetes/pull/88540, currently the `extender/v1` API in the scheduler is missing a couple things:
* internal version or conversion methods
* typemeta on objects
* json tags on types

Since we can't make these changes to the current version of the type without potentially breaking expectations for existing consumers, this may require bumping to a new api version

Huang-Wei:
One note: if we want to incorporate TypeMeta into the serialized extender APIs, we should ensure "serialized extender API objects" to be passed onto extenders, and also guide users to use newly generated utilities to covert the received objects into v1, or v2 at their extender programs.

To minimize the impact on user side, I'd suggest we only start the refactoring when the fields really need updates (and then bump it to v2). Otherwise, don't bother and keep the current unversioned v1 API.

damemi:
Cross-referencing https://github.com/kubernetes/kubernetes/issues/89396

alculquicondor:
> typemeta on objects

Do we actually need those? Alternatively, do we need to comply to OpenAPI? The extenders API is about IPC and not resources (definition or storage).
What are the precedents to this? The closer I can think of is the communication between kube-apiserver and kubelet around resources usage. IIRC, kubelet offers a GRPC server and some basic json endpoints.

It might be worth asking the API folks for guidance.

alculquicondor:
@liggitt do you know of such precedents?

On the other hand, OpenAPI is probably well-known among kubernetes users.

kerthcet:
Can anyone tell me what unaddressed issues we left here?

alculquicondor:
The extenders API is not something we want to maintain. Unless there is an important use-case that can't be covered with the scheduling framework, I suggest that we close this issue.

kerthcet:
As advised, let us close this issue and reopen if needed. cc @damemi 
/close

PR #92866, opened 2020-07-07
scheduler: Extend ExtenderFilterResult to include UnschedulableAndUnresolvable nodes 

### LINKED ISSUE #91281: scheduler/extender: extend the current `ExtenderFilterResult` struct to include node status
Updated on 13/06/2020:

- ~~[ ] update extender design proposal~~ (not possible, see https://github.com/kubernetes/community/pull/4816)
- [x] migrate it to k/enhancements https://github.com/kubernetes/enhancements/pull/1840
- [x] submit a API change PR https://github.com/kubernetes/enhancements/pull/1850
- [ ] implement it https://github.com/kubernetes/kubernetes/pull/92866

I'd like to extend the current `ExtenderFilterResult` struct to include node status of the failed pod.

https://github.com/kubernetes/kubernetes/blob/9e991415386e4cf155a24b1da15becaa390438d8/staging/src/k8s.io/kube-scheduler/extender/v1/types.go#L86-L97

e.g. 

- add a new map `FailedUnresolvableNodes` to include `UnschedulableAndUnresolvable` nodes 
- or use an extensive struct to include both status and message in the node map (which supersedes current `FailedNodes` map)

**Why is this needed**:

`FailedNodes` is a map of string to string which includes failed message only.

https://github.com/kubernetes/kubernetes/blob/9e991415386e4cf155a24b1da15becaa390438d8/staging/src/k8s.io/kube-scheduler/extender/v1/types.go#L82-L83

The status of the failed node is hardcoded to `Unscheduable` here.

https://github.com/kubernetes/kubernetes/blob/9e991415386e4cf155a24b1da15becaa390438d8/pkg/scheduler/core/generic_scheduler.go#L520-L526

The extender filter can't report that failed nodes are unresolvable like built-in filters.

https://github.com/kubernetes/kubernetes/blob/7ba332a839c844ef4ba0191ddca2ce3601d283af/pkg/scheduler/core/generic_scheduler.go#L1034-L1047

cofyc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Fixes #91281

```release-note
Scheduler extender filter interface now can report unresolvable failed nodes in the new field `FailedAndUnresolvableNodes` of  `ExtenderFilterResult` struct. Nodes in this map will be skipped in the preemption phase. 
```

```docs
- [KEP]: https://github.com/kubernetes/enhancements/pull/1850
```

### DIFF
--- a/pkg/scheduler/core/extender.go
+++ b/pkg/scheduler/core/extender.go
@@ -269,11 +269,13 @@ func convertToNodeNameToMetaVictims(
 
 // Filter based on extender implemented predicate functions. The filtered list is
 // expected to be a subset of the supplied list; otherwise the function returns an error.
-// failedNodesMap optionally contains the list of failed nodes and failure reasons.
+// The failedNodes and failedAndUnresolvableNodes optionally contains the list
+// of failed nodes and failure reasons, except nodes in the latter are
+// unresolvable.
 func (h *HTTPExtender) Filter(
 	pod *v1.Pod,
 	nodes []*v1.Node,
-) ([]*v1.Node, extenderv1.FailedNodesMap, error) {
+) (filteredList []*v1.Node, failedNodes, failedAndUnresolvableNodes extenderv1.FailedNodesMap, err error) {
 	var (
 		result     extenderv1.ExtenderFilterResult
 		nodeList   *v1.NodeList
@@ -287,7 +289,7 @@ func (h *HTTPExtender) Filter(
 	}
 
 	if h.filterVerb == "" {
-		return nodes, extenderv1.FailedNodesMap{}, nil
+		return nodes, extenderv1.FailedNodesMap{}, extenderv1.FailedNodesMap{}, nil
 	}
 
 	if h.nodeCacheCapable {
@@ -310,10 +312,10 @@ func (h *HTTPExtender) Filter(
 	}
 
 	if err := h.send(h.filterVerb, args, &result); err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 	if result.Error != "" {
-		return nil, nil, fmt.Errorf(result.Error)
+		return nil, nil, nil, fmt.Errorf(result.Error)
 	}
 
 	if h.nodeCacheCapable && result.NodeNames != nil {
@@ -322,7 +324,7 @@ func (h *HTTPExtender) Filter(
 			if n, ok := fromNodeName[nodeName]; ok {
 				nodeResult[i] = n
 			} else {
-				return nil, nil, fmt.Errorf(
+				return nil, nil, nil, fmt.Errorf(
 					"extender %q claims a filtered node %q which is not found in the input node list",
 					h.extenderURL, nodeName)
 			}
@@ -334,7 +336,7 @@ func (h *HTTPExtender) Filter(
 		}
 	}
 
-	return nodeResult, result.FailedNodes, nil
+	return nodeResult, result.FailedNodes, result.FailedAndUnresolvableNodes, nil
 }
 
 // Prioritize based on extender implemented priority functions. Weight*priority is added
--- a/pkg/scheduler/core/generic_scheduler.go
+++ b/pkg/scheduler/core/generic_scheduler.go
@@ -345,6 +345,9 @@ func (g *genericScheduler) findNodesThatPassFilters(
 }
 
 func (g *genericScheduler) findNodesThatPassExtenders(pod *v1.Pod, feasibleNodes []*v1.Node, statuses framework.NodeToStatusMap) ([]*v1.Node, error) {
+	// Extenders are called sequentially.
+	// Nodes in original feasibleNodes can be excluded in one extender, and pass on to the next
+	// extender in a decreasing manner.
 	for _, extender := range g.extenders {
 		if len(feasibleNodes) == 0 {
 			break
@@ -352,7 +355,13 @@ func (g *genericScheduler) findNodesThatPassExtenders(pod *v1.Pod, feasibleNodes
 		if !extender.IsInterested(pod) {
 			continue
 		}
-		feasibleList, failedMap, err := extender.Filter(pod, feasibleNodes)
+
+		// Status of failed nodes in failedAndUnresolvableMap will be added or overwritten in <statuses>,
+		// so that the scheduler framework can respect the UnschedulableAndUnresolvable status for
+		// particular nodes, and this may eventually improve preemption efficiency.
+		// Note: users are recommended to configure the extenders that may return UnschedulableAndUnresolvable
+		// status ahead of others.
+		feasibleList, failedMap, failedAndUnresolvableMap, err := extender.Filter(pod, feasibleNodes)
 		if err != nil {
 			if extender.IsIgnorable() {
 				klog.InfoS("Skipping extender as it returned error and has ignorable flag set", "extender", extender, "err", err)
@@ -361,13 +370,28 @@ func (g *genericScheduler) findNodesThatPassExtenders(pod *v1.Pod, feasibleNodes
 			return nil, err
 		}
 
+		for failedNodeName, failedMsg := range failedAndUnresolvableMap {
+			var aggregatedReasons []string
+			if _, found := statuses[failedNodeName]; found {
+				aggregatedReasons = statuses[failedNodeName].Reasons()
+			}
+			aggregatedReasons = append(aggregatedReasons, failedMsg)
+			statuses[failedNodeName] = framework.NewStatus(framework.UnschedulableAndUnresolvable, aggregatedReasons...)
+		}
+
 		for failedNodeName, failedMsg := range failedMap {
+			if _, found := failedAndUnresolvableMap[failedNodeName]; found {
+				// failedAndUnresolvableMap takes precedence over failedMap
+				// note that this only happens if the extender returns the node in both maps
+				continue
+			}
 			if _, found := statuses[failedNodeName]; !found {
 				statuses[failedNodeName] = framework.NewStatus(framework.Unschedulable, failedMsg)
 			} else {
 				statuses[failedNodeName].AppendReason(failedMsg)
 			}
 		}
+
 		feasibleNodes = feasibleList
 	}
 	return feasibleNodes, nil
--- a/pkg/scheduler/core/generic_scheduler_test.go
+++ b/pkg/scheduler/core/generic_scheduler_test.go
@@ -273,6 +273,180 @@ func TestSelectHost(t *testing.T) {
 	}
 }
 
+func TestFindNodesThatPassExtenders(t *testing.T) {
+	tests := []struct {
+		name                  string
+		extenders             []st.FakeExtender
+		nodes                 []*v1.Node
+		filteredNodesStatuses framework.NodeToStatusMap
+		expectsErr            bool
+		expectedNodes         []*v1.Node
+		expectedStatuses      framework.NodeToStatusMap
+	}{
+		{
+			name: "error",
+			extenders: []st.FakeExtender{
+				{
+					Predicates: []st.FitPredicate{st.ErrorPredicateExtender},
+				},
+			},
+			nodes:                 makeNodeList([]string{"a"}),
+			filteredNodesStatuses: make(framework.NodeToStatusMap),
+			expectsErr:            true,
+		},
+		{
+			name: "success",
+			extenders: []st.FakeExtender{
+				{
+					Predicates: []st.FitPredicate{st.TruePredicateExtender},
+				},
+			},
+			nodes:                 makeNodeList([]string{"a"}),
+			filteredNodesStatuses: make(framework.NodeToStatusMap),
+			expectsErr:            false,
+			expectedNodes:         makeNodeList([]string{"a"}),
+			expectedStatuses:      make(framework.NodeToStatusMap),
+		},
+		{
+			name: "unschedulable",
+			extenders: []st.FakeExtender{
+				{
+					Predicates: []st.FitPredicate{func(pod *v1.Pod, node *v1.Node) *framework.Status {
+						if node.Name == "a" {
+							return framework.NewStatus(framework.Success)
+						}
+						return framework.NewStatus(framework.Unschedulable, fmt.Sprintf("node %q is not allowed", node.Name))
+					}},
+				},
+			},
+			nodes:                 makeNodeList([]string{"a", "b"}),
+			filteredNodesStatuses: make(framework.NodeToStatusMap),
+			expectsErr:            false,
+			expectedNodes:         makeNodeList([]string{"a"}),
+			expectedStatuses: framework.NodeToStatusMap{
+				"b": framework.NewStatus(framework.Unschedulable, fmt.Sprintf("FakeExtender: node %q failed", "b")),
+			},
+		},
+		{
+			name: "unschedulable and unresolvable",
+			extenders: []st.FakeExtender{
+				{
+					Predicates: []st.FitPredicate{func(pod *v1.Pod, node *v1.Node) *framework.Status {
+						if node.Name == "a" {
+							return framework.NewStatus(framework.Success)
+						}
+						if node.Name == "b" {
+							return framework.NewStatus(framework.Unschedulable, fmt.Sprintf("node %q is not allowed", node.Name))
+						}
+						return framework.NewStatus(framework.UnschedulableAndUnresolvable, fmt.Sprintf("node %q is not allowed", node.Name))
+					}},
+				},
+			},
+			nodes:                 makeNodeList([]string{"a", "b", "c"}),
+			filteredNodesStatuses: make(framework.NodeToStatusMap),
+			expectsErr:            false,
+			expectedNodes:         makeNodeList([]string{"a"}),
+			expectedStatuses: framework.NodeToStatusMap{
+				"b": framework.NewStatus(framework.Unschedulable, fmt.Sprintf("FakeExtender: node %q failed", "b")),
+				"c": framework.NewStatus(framework.UnschedulableAndUnresolvable, fmt.Sprintf("FakeExtender: node %q failed and unresolvable", "c")),
+			},
+		},
+		{
+			name: "extender may overwrite the statuses",
+			extenders: []st.FakeExtender{
+				{
+					Predicates: []st.FitPredicate{func(pod *v1.Pod, node *v1.Node) *framework.Status {
+						if node.Name == "a" {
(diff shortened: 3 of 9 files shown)


cofyc:
/priority backlog
(not sure if it can go into 1.19)
/sig scheduling
/assign @Huang-Wei

Huang-Wei:
@cofyc As this PR also needs API review, and unfortunately tomorrow is the code freeze for 1.19. Per the original plan (targeting 1.20) in KEP, would you mind holding the review until 1.20?

cofyc:
> @cofyc As this PR also needs API review, and unfortunately tomorrow is the code freeze for 1.19. Per the original plan (targeting 1.20) in KEP, would you mind holding the review until 1.20?

sure

pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -268,6 +269,118 @@ func TestSelectHost(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			var extenders []framework.Extender
+			for ii := range tt.extenders {
+				extenders = append(extenders, &tt.extenders[ii])
+			}
+			scheduler := &genericScheduler{
+				extenders: extenders,
+			}
+			pod := &v1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "1", UID: types.UID("1")}}
+			filteredNodesStatuses := make(framework.NodeToStatusMap)
```
soulxu, line 363: It didn't test the case there is the existing status for the failed node.

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -382,7 +382,19 @@ func (g *genericScheduler) findNodesThatPassExtenders(pod *v1.Pod, feasibleNodes
 			return nil, err
 		}
 
+		for failedNodeName, failedMsg := range failedAndUnresolvableMap {
+			if _, found := statuses[failedNodeName]; !found {
```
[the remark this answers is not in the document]
cofyc: failed nodes in the previous filter phase will not be passed to the extenders, but we still need to aggregate error messages if there are more than one extenders

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -382,7 +382,19 @@ func (g *genericScheduler) findNodesThatPassExtenders(pod *v1.Pod, feasibleNodes
...
+				statuses[failedNodeName] = framework.NewStatus(framework.UnschedulableAndUnresolvable, failedMsg)
+			} else {
+				statuses[failedNodeName].AppendReason(failedMsg)
```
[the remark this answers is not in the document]
cofyc: it makes sense. it's possible that there will be more than one extenders, the `UnschedulableAndUnresolvable` status should take precedence over `Unschedulable`.

pkg/scheduler/core/generic_scheduler_test.go
cofyc: I don't think it's necessary. As you said, failed nodes in filter phase will not be passed to extenders.

But I will add `multiple extenders` scenario. `Unschedulable` status reported by one extender will be overridden by `UnschedulableAndUnresolvable` status from another extender.

pkg/scheduler/core/generic_scheduler.go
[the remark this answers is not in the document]
cofyc: done

pkg/scheduler/core/generic_scheduler_test.go
cofyc: done

cofyc (commented):
@soulxu thanks! PTAL

cofyc:
@Huang-Wei can you take a look recently?

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -344,14 +344,21 @@ func (g *genericScheduler) findNodesThatPassFilters(ctx context.Context, prof *p
...
+	var prevFailedButResolvableNodes map[string]*v1.Node
 	for _, extender := range g.extenders {
 		if len(feasibleNodes) == 0 {
 			break
 		}
 		if !extender.IsInterested(pod) {
 			continue
 		}
-		feasibleList, failedMap, err := extender.Filter(pod, feasibleNodes)
+		// We call extenders sequentially, so unfeasible nodes can be excluded
```
Huang-Wei, line 355:
```suggestion
		// We call extenders sequentially, so infeasible nodes can be excluded
```

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -361,14 +368,45 @@ func (g *genericScheduler) findNodesThatPassExtenders(pod *v1.Pod, feasibleNodes
...
 				statuses[failedNodeName] = framework.NewStatus(framework.Unschedulable, failedMsg)
 			} else {
 				statuses[failedNodeName].AppendReason(failedMsg)
 			}
 		}
-		feasibleNodes = feasibleList
+
+		prevFailedButResolvableNodes = make(map[string]*v1.Node)
+		for _, node := range feasibleNodes {
+			if status, ok := statuses[node.Name]; ok && status.Code() == framework.Unschedulable {
```
Huang-Wei, line 397: How can L396 and L397 happen at the same time? I mean a feasibleNode shouldn't have a Unschedulable or UnschedulableAndUnresolvable status, right?

In the upstream, a failed node won't appear in feasibleNodes, and vice verse - a successful node won't appear in the statusMap:

https://github.com/kubernetes/kubernetes/blob/db9f1e91d977c45fe7df8e1e1bd415a3292995e1/pkg/scheduler/core/generic_scheduler.go#L306-L320

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -361,14 +368,45 @@ func (g *genericScheduler) findNodesThatPassExtenders(pod *v1.Pod, feasibleNodes
...
+				prevFailedButResolvableNodes[node.Name] = node
+			}
+		}
+
+		// Failed but unresolvable nodes may be passed to current extender, we
+		// should exclude them.
```
Huang-Wei, line 403: Same as previous commit, the `feasibleNodes` returned shouldn't contain any failed node (with any failure reason).

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -344,14 +344,21 @@ func (g *genericScheduler) findNodesThatPassFilters(ctx context.Context, prof *p
...
 		if len(feasibleNodes) == 0 {
 			break
 		}
 		if !extender.IsInterested(pod) {
 			continue
 		}
-		feasibleList, failedMap, err := extender.Filter(pod, feasibleNodes)
+		// We call extenders sequentially, so unfeasible nodes can be excluded
+		// for later extenders. However, failed but resolvable nodes must be
+		// passed as they may be marked unresolvable.
```
Huang-Wei, line 357: I doubt if we really want this behavior. Even if we have to do this, isn't in the first round, the _implicit_ infeasible nodes (with Unresolveable code) in statuses should also be included?

pkg/scheduler/core/generic_scheduler.go
Huang-Wei: OK, IIUC, you're passing the unresolvable nodes to feasible nodes just to give the extender a chance to potentially mark it as UnschedulableAndUnresolvable? if so, it's actually breaking the semantics of "feasible nodes", which is not backward compatible IMO - current extender users wouldn't be surprised that they receive an infeasible node in a feasible nodes list.

Huang-Wei (commented):
Thanks @cofyc . I'm concerned with the implementation of `findNodesThatPassExtenders()`.

(Haven't went through tests yet)

pkg/scheduler/core/generic_scheduler.go
ahg-g: I agree with Wei, I think we should either pass all nodes to all extenders or pass the feasible ones only. In the framework we are doing the latter, basically we stop on the first filter that returns Unschedulable status: https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/runtime/framework.go#L506

pkg/scheduler/core/generic_scheduler.go
cofyc: done

pkg/scheduler/core/generic_scheduler.go
cofyc: done

pkg/scheduler/core/generic_scheduler.go
cofyc: sorry for being late (busy on something else)

I've updated the PR, PTAL.

cofyc: created a PR to add a note in the KEP: https://github.com/kubernetes/enhancements/pull/2132/files

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -360,13 +366,30 @@ func (g *genericScheduler) findNodesThatPassExtenders(pod *v1.Pod, feasibleNodes
 			return nil, err
 		}
 
+		for failedNodeName, failedMsg := range failedAndUnresolvableMap {
+			if _, found := statuses[failedNodeName]; !found {
+				statuses[failedNodeName] = framework.NewStatus(framework.UnschedulableAndUnresolvable, failedMsg)
+			} else {
+				if statuses[failedNodeName].Code() != framework.UnschedulableAndUnresolvable {
```
Huang-Wei, line 373: This `if` should be always true, right? If so, how about refactor the whole logic to be:

```go
for failedNodeName, failedMsg := range failedAndUnresolvableMap {
	var aggregatedReasons []string
	if _, found := statuses[failedNodeName]; found {
		aggregatedReasons = statuses[failedNodeName].Reasons()
	}
	aggregatedReasons = append(aggregatedReasons, failedMsg)
	statuses[failedNodeName] = framework.NewStatus(framework.UnschedulableAndUnresolvable, aggregatedReasons...)
}
```

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -350,7 +350,13 @@ func (g *genericScheduler) findNodesThatPassExtenders(pod *v1.Pod, feasibleNodes
 		if !extender.IsInterested(pod) {
 			continue
 		}
-		feasibleList, failedMap, err := extender.Filter(pod, feasibleNodes)
+
+		// We call extenders sequentially, and exclude the node on the first
```
Huang-Wei, line 354: Move part of the comment to the `for` loop (L346)?

```
	// Extenders are called sequentially.
	// Nodes in original feasibleNodes can be excluded in one extender, and pass on to the next
	// extender in a decreasing manner.
```

And then here only comment about failedAndUnresolvableMap vs. failedMap:

```
		// Status of failed nodes in failedAndUnresolvableMap will be added or overwritten in <statuses>,
		// so that the scheduler framework can respect the UnschedulableAndUnresolvable status for
		// particular nodes, and this may eventually improve preemption efficiency.
		// Note: users are recommended to configure the extenders that may return UnschedulableAndUnresolvable
		// status ahead of others.
```

pkg/scheduler/core/generic_scheduler.go
cofyc: make sense, updated

pkg/scheduler/core/generic_scheduler.go
cofyc: seems better!

cofyc:
@Huang-Wei PR is updated and squashed too

pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -269,6 +271,149 @@ func TestSelectHost(t *testing.T) {
...
+			},
+			nodes:         makeNodeList([]string{"a", "b", "c"}),
+			expectsErr:    false,
+			expectedNodes: makeNodeList([]string{"a"}),
+			expectedStatuses: framework.NodeToStatusMap{
+				"b": framework.NewStatus(framework.Unschedulable, fmt.Sprintf("FakeExtender: node %q failed", "b")),
+				"c": framework.NewStatus(framework.UnschedulableAndUnresolvable, fmt.Sprintf("FakeExtender: node %q failed and unresolvable", "c")),
+			},
+		},
+	}
```
Huang-Wei, line 378: Although a little rare, do you think it's necessary to add a case to cover the path that "extender may overwrite the statuses"? i.e., the tuple `{feasibleNodes, statuses}` passed onto the extender(s) is `{[a, b], {c: Unchedulable}}`, extenders may return feasibleNodes as [a], while statuses as {c: UnschedualbeAndUnresolvable, b: Unschedulable}.

Huang-Wei (commented):
One final comment on testcase. LGTM otherwise.

Thanks @cofyc!

pkg/scheduler/core/generic_scheduler_test.go
cofyc: I think it's necessary. The case "extender may overwrite the statuses" is added. PTAL.

Huang-Wei:
/lgtm
/approve

/hold
in case @ahg-g or @alculquicondor want to take another look.

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -360,13 +369,27 @@ func (g *genericScheduler) findNodesThatPassExtenders(pod *v1.Pod, feasibleNodes
...
+			var aggregatedReasons []string
+			if _, found := statuses[failedNodeName]; found {
+				aggregatedReasons = statuses[failedNodeName].Reasons()
+			}
+			aggregatedReasons = append(aggregatedReasons, failedMsg)
+			statuses[failedNodeName] = framework.NewStatus(framework.UnschedulableAndUnresolvable, aggregatedReasons...)
+		}
+
 		for failedNodeName, failedMsg := range failedMap {
+			if _, found := failedAndUnresolvableMap[failedNodeName]; found {
```
alculquicondor, line 382: would this happen?
I can only think of the case where the extender returns both states for the same Node.

cofyc: yes, ignore the nodes in `failedMap` if they are already in `failedAndUnresolvableMap`

alculquicondor: please add a comment that this only happens if the extender returns it in both

cofyc: added, PTAL

alculquicondor (commented):
/lgtm
/hold cancel

Huang-Wei:
This also needs api reviews.

@kubernetes/api-reviewers Can any API expert take a review? Thanks.

pkg/scheduler/core/extender.go
```diff
@@ -287,7 +289,7 @@ func (h *HTTPExtender) Filter(
 	}
 
 	if h.filterVerb == "" {
-		return nodes, extenderv1.FailedNodesMap{}, nil
+		return nodes, nil, nil, nil
```
soulxu, line 292: I'm wondering this would break when loop the failedNodesMap in "genernicSchduler.findNodesThatPassExtenders"

cofyc: Thanks! Fixed.

Huang-Wei:
/approve
on scheduling side

Huang-Wei:
/lgtm
on scheduling side

alculquicondor:
We know that this API doesn't respect openAPI and there is an open issue to fix this in a v2: #88634 

As it stands today, api-change lgtm.

staging/src/k8s.io/kube-scheduler/extender/v1/types.go
```diff
@@ -92,6 +92,10 @@ type ExtenderFilterResult struct {
 	NodeNames *[]string
 	// Filtered out nodes where the pod can't be scheduled and the failure messages
 	FailedNodes FailedNodesMap
+	// Filtered out nodes where the pod can't be scheduled and preemption would
+	// not change anything. The value is the failure message same as FailedNodes.
+	// Nodes specified here takes precedence over FailedNodes.
+	FailedAndUnresolvableNodes FailedNodesMap
```
liggitt, line 98: is it an error to include a node in FailedNodes and not in this map?

liggitt: > ...and preemption would not change anything

is that what unresolveable means? the doc doesn't seem to match the variable name

liggitt, line 98: as a serialized API type, this file is strange... it has no json tags, etc... are the scheduler extender APIs being actively expanded?

ahg-g: > is it an error to include a node in FailedNodes and not in this map?

No. We are practically considering the union, however nodes in this map don't get considered during preemption.

> > ...and preemption would not change anything
> 
> is that what unresolveable means? the doc doesn't seem to match the variable name

Unresolvable is the term used in the scheduler pkg to indicate that preempting pods from a node doesn't make it schedulable for an incoming pod.

ahg-g: It is not "actively" expanded per se, I think this is the first update in a very long time. This is an old API, it precedes the involvement of most active members in sig scheduling, we are aware that it misses json tags and proper versioning, there is an open issue to fix that:  #88634

liggitt: I see. I parsed this as "(Failed Nodes) And (Unresolveable Nodes)" instead of "(Failed And Unresolveable) Nodes". If Unresolveable implies Failed, then would it be clearer to just call this UnresolveableNodes?

liggitt: ok

alculquicondor: this tries to mimic scheduling plugin statuses, which could be "Unschedulable" or "UnschedulableAndUnresolvable".

but I don't have a strong opinion about removing the "Failed" part

liggitt: if this is following an existing pattern in this API, that's fine

cofyc:
conflicts in bazel files are resolved

Huang-Wei:
Kindly ping @liggitt for approval.

liggitt:
/approve
for `staging/src/k8s.io/kube-scheduler/extender/v1/types.go` changes

PR #89629, opened 2020-03-29
Modified subpath configmap mount fails when container restarts

fatedier:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
When a container uses a configmap which is mounted with the subPath option, the configmap is changed and then the container (but not the pod) restarts the mounting of the configmap fails.

Since #82784 is not active and not rebased, i submit this PR to continue this work or someone can provide another solution?

**Which issue(s) this PR fixes**:
Fixes #68211

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
Fix an issue with container restarts using a modified configmap or secret subpath volume mount.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #92869, opened 2020-07-07
Windows: Fixes subpath volume persistance on container restart 

### LINKED ISSUE #90336: Windows: subpath mount changes on a container restart if the environment variable changes
**What happened**:

The test ``[k8s.io] Variable Expansion should not change the subpath mount on a container restart if the environment variable changes [sig-storage][Slow] [Conformance]`` fails on Windows. On container restart, the subpath mount changes.

The test does the following:

- spawns a Pod with an EmptyDir volume mounted in 2 locations: ``/volume_mount`` and ``subpath_mount`` (the 2nd one has a subpath expr set, taken from ``metadata.annotations['mysubpath']``)
- the pod has a liveness probe that checks for the file ``/subpath_mount/test.log``
- the test updates ``metadata.annotations['mysubpath']`` to a different value.
- the test then deletes the file ``/subpath_mount/test.log`` and then waits for the container restart because of the liveness probe.
- the test recreates the ``/subpath_mount/test.log``
- the test asserts that ``/volume_mount/foo/test.log`` exists, and that ``/volume_mount/newsubpath/test.log`` does not.

The test fails on the assertions on Windows.

The container volume subpath should not change on container restart.

    kubectl create -f liveness.yaml

    # check that test.log exists.
    kubectl exec -n subpath-test-ns subpath-test -- ls /subpath_mount

    # create /volume_mount/bar/test.log and /volume_mount/bar/tender
    kubectl exec -n subpath-test-ns subpath-test -- /bin/bash -c "mkdir -p /volume_mount/bar && touch /volume_mount/bar/test.log && touch /volume_mount/bar/tender"

    # update the pod volume subpath.
    kubectl patch pod -n subpath-test-ns subpath-test -p '{"metadata":{"annotations":{"mysubpath":"bar"}}}'

    # trigger the restart by deleting the file checked by the liveness probe.
    kubectl exec -n subpath-test-ns subpath-test -- rm /subpath_mount/test.log

    # check the /subpath_mount contents. It shouldn't contain test.log or tender.
    kubectl exec -n subpath-test-ns subpath-test -- ls /subpath_mount

liveness.yaml:

    apiVersion: v1
    items:
    - apiVersion: v1
      kind: Namespace
      metadata:
        name: subpath-test-ns
    - apiVersion: v1
      kind: Pod
      metadata:
        annotations:
          kubernetes.io/psp: e2e-test-privileged-psp
          mysubpath: foo
        name: subpath-test
        namespace: subpath-test-ns
      spec:
        containers:
        - command:
          - /bin/sh
          - -ec
          - 'sleep 100000'
          env:
          - name: POD_NAME
            valueFrom:
              fieldRef:
                apiVersion: v1
                fieldPath: metadata.annotations['mysubpath']
          image: claudiubelu/busybox:1.29
          imagePullPolicy: Always
          livenessProbe:
            exec:
              command:
              - /bin/sh
              - -c
              - 'cat /subpath_mount/test.log'
            failureThreshold: 1
            initialDelaySeconds: 10
            periodSeconds: 2
            successThreshold: 1
            timeoutSeconds: 1
          name: dapi-container
          volumeMounts:
          - mountPath: /subpath_mount
            name: workdir1
            subPathExpr: $(POD_NAME)
          - mountPath: /volume_mount
            name: workdir1
        initContainers:
        - command:
          - sh
          - -c
          - mkdir -p /volume_mount/foo; touch /volume_mount/foo/test.log
          image: claudiubelu/busybox:1.29
          imagePullPolicy: Always
          name: init-volume-50ab29f4
          volumeMounts:
          - mountPath: /subpath_mount
            name: workdir1
          - mountPath: /volume_mount
            name: workdir1
        volumes:
        - emptyDir: {}
          name: workdir1
    kind: List

/sig windows

**Environment**:
- Kubernetes version (use `kubectl version`): v1.18
- OS (e.g: `cat /etc/os-release`): Windows Server 2019
- Install tools: N/A
- Others: N/A

claudiubelu:
**What type of PR is this?**

/kind bug

/sig windows
/sig storage

/priority important-soon

Pods can have volume subpaths mounted inside the containers. Those subpaths can be based on expressions based on expanded environment variables, which can also be based on Pod annotations.

Kubernetes expects that the mounted volume subpaths to remain consistent across container restarts even if the Pod annotation it was based on changed. This behaviour is checked by the E2E test ``should not change the subpath mount on a container restart if the environment variable change``.

This test fails on Windows because the subpath itself is always mounted (which is based on the Pod Annotation), while on Linux, on the first first, the volume subpath is created and mounted under ``kubelet/pods/<uid>/volume-subpaths/<volume>/<container-name>`` and cached, then on restart, it is going to use the same mount path that already exists, keeping the same subpath mounted in the container.

This commit addresses the issue on Windows, following a similar approach to the Linux implementation.

Refactors some of the code, which can be used on both Linux and Windows.

Fixes #90336

```release-note
Fixed issue on changing volume subpaths based on expanded environment variables on Windows hosts.
```

```docs

```

### DIFF
--- a/pkg/volume/util/subpath/BUILD
+++ b/pkg/volume/util/subpath/BUILD
@@ -3,6 +3,7 @@ load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 go_library(
     name = "go_default_library",
     srcs = [
+        "common.go",
         "subpath.go",
         "subpath_linux.go",
         "subpath_unsupported.go",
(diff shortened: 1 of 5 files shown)


michmike (approved):
/lgtm
/approve
@ddebroy  can you PTAL?

jsturtevant:
I think this PR is no longer relevant due to #89629 which changed the behavior of mounting subpath's on restarts and removed this requirement from the conformance tests.  @claudiubelu  can you confirm?

claudiubelu:
That is correct. The behaviour that this PR was adding is no longer valid in Kubernetes. The test that was verifying this behaviour was also removed in the PR mentioned above. Because of this, we no longer need this, we are already doing what is currently expected regarding volume subpaths on container restart.

Issue #89036, opened 2020-03-11
Optimization for scheduler preemption


hex108:
**What would you like to be added**:

* If a pod's priority is lowest(e.g. there is three priroites in cluster: 10, 20 and 30, and the pod's priority is 10), it could not preempt any other pods, then we could skip trying preemption for it, it will improve schedule performance. 

* We could just try to search a part of nodes instead of all nodes for preemption, the idea is inspired by [Percentage of Nodes to Score
](https://kubernetes.io/docs/concepts/scheduling/scheduler-perf-tuning/#percentage-of-nodes-to-score)

**Why is this needed**:

In our clusters(>3K nodes), preemption occurs frequently, scheduler could only try to schedule less than 10 pods. It could be improved by above ideas.

/sig scheduling
/assign

Huang-Wei:
> If a pod's priority is lowest(e.g. there is three priroites in cluster: 10, 20 and 30, and the pod's priority is 10), it could not preempt any other pods, then we could skip trying preemption for it, it will improve schedule performance.

I can see the point to skip preemption for Pods with default priority (0). However, in your case, Pods with priority 10 can still preempt Pods such as the ones with default priority, how can you manage to skip them?

> We could just try to search a part of nodes instead of all nodes for preemption

This is a valid point. In the next release (1.19), we will extract the hard-coded preemption logic into plugins, which is a good timing to consider introducing some pluginConfig argument like `PercentageOfNodesToTryPreempt`.

hex108:
> I can see the point to skip preemption for Pods with default priority (0). However, in your case, Pods with priority 10 can still preempt Pods such as the ones with default priority, how can you manage to skip them?

The priroties(10, 20, and 30) are just examples. Actually, the priorities could be any interger number and we can set the default priority to any valid value by [setting one priority class's globalDefault value to true](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass). We could list all prioriy classes and find out which is the lowest priroity class.(For optimizaiton, we could record the lowest priority using a variable, and create a prirority informer to wathc priority class add/update/delete event, and update the vairable accordingly.)

> This is a valid point. In the next release (1.19), we will extract the hard-coded preemption logic into plugins, which is a good timing to consider introducing some pluginConfig argument like PercentageOfNodesToTryPreempt.

Glad to know it! Since they are not dependent, maybe we could start the work for `PercentageOfNodesToTryPreempt` in advance?

[...]

PR #92875, opened 2020-07-07
Avoid executing preemption for pods with 0 priority

ahg-g:
**What type of PR is this?**

 /kind feature

0 is the lowest priority, pods with such priority can't preempt any other pods. This PR avoids executing the preemption logic for such pods. This should have a non-negligible positive impact on scheduling latency since 0 is the default pod priority and we expect that most workloads just use that.

Part of #89036

```release-note
NONE
```


ahg-g:
/hold

to understand the negative priority semantics

ahg-g:
negative priority is possible, so the only way to do this is by tracking the lowest priority in the cluster (which in most cases will likely be zero).

/close

hex108:
We are using negative priority now. :)

PR #83426, opened 2019-10-02
Dedupe logging for PD SetUpAt and added a slow SetVolumeOwnership warning

davidz627:
`SetUpAt` errors actually bubble up to the `operation_generator` level and are logged in a much better way with more context. The one-off logs are duplicates and split context in different places.

Also we have a known issue with volume ownership being slow and it would be nice to surface a warning when this feature is being used. This would make debugging the issue much easier.

/kind cleanup
/sig storage
/assign @msau42 @jingxu97 
/priority backlog

```release-note
NONE
```

PR #92878, opened 2020-07-07
cleanup: print warning message only if the function does not finish within 30 seconds

### LINKED ISSUE #90293: kubelet warnings W0420 flooding systemd logs (kubernetes 1.18.2)
**What happened**:
there are many kubelet warnings in the systemd logs

```
Apr 20 11:03:37 lxkubenode01 kubelet[9103]: W0420 11:03:37.275020    9103 volume_linux.go:49] Setting volume ownership for /var/lib/kubelet/pods/63c4a49f-bf23-4b87-989e-102f5fcdb315/volumes/kubernetes.io~secret/seq-token-whpfk and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699
Apr 20 11:03:46 lxkubenode01 kubelet[9103]: W0420 11:03:46.198559    9103 volume_linux.go:49] Setting volume ownership for /var/lib/kubelet/pods/cdb79fa0-4942-4211-9261-a4928f872bd6/volumes/kubernetes.io~secret/prometheus-operator-prometheus-node-exporter-token-snbtf and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699
Apr 20 11:03:47 lxkubenode01 kubelet[9103]: W0420 11:03:47.201212    9103 volume_linux.go:49] Setting volume ownership for /var/lib/kubelet/pods/8cafe45f-2d14-4eb1-8c38-71a54b34f83c/volumes/kubernetes.io~secret/default-token-9hq84 and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699
Apr 20 11:03:50 lxkubenode01 kubelet[9103]: W0420 11:03:50.209770    9103 volume_linux.go:49] Setting volume ownership for /var/lib/kubelet/pods/174c1204-11e5-4297-b71f-1ebd2c961b95/volumes/kubernetes.io~secret/default-token-cwdzw and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699
Apr 20 11:03:50 lxkubenode01 kubelet[9103]: W0420 11:03:50.209791    9103 volume_linux.go:49] Setting volume ownership for /var/lib/kubelet/pods/174c1204-11e5-4297-b71f-1ebd2c961b95/volumes/kubernetes.io~configmap/config-volume and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699
Apr 20 11:03:53 lxkubenode01 kubelet[9103]: W0420 11:03:53.217655    9103 volume_linux.go:49] Setting volume ownership for /var/lib/kubelet/pods/655e1952-1200-4653-bdd6-e4029cb75d2b/volumes/kubernetes.io~configmap/config-volume and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699
Apr 20 11:03:53 lxkubenode01 kubelet[9103]: W0420 11:03:53.217671    9103 volume_linux.go:49] Setting volume ownership for /var/lib/kubelet/pods/2f69fc08-f4c9-447b-8cb2-bf737a2d9136/volumes/kubernetes.io~configmap/app-config and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699
Apr 20 11:03:53 lxkubenode01 kubelet[9103]: W0420 11:03:53.217738    9103 volume_linux.go:49] Setting volume ownership for /var/lib/kubelet/pods/655e1952-1200-4653-bdd6-e4029cb75d2b/volumes/kubernetes.io~secret/default-token-cwdzw and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699
Apr 20 11:03:53 lxkubenode01 kubelet[9103]: W0420 11:03:53.217769    9103 volume_linux.go:49] Setting volume ownership for /var/lib/kubelet/pods/2f69fc08-f4c9-447b-8cb2-bf737a2d9136/volumes/kubernetes.io~secret/geocoding-aspnetcore-token-xrg72 and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699
Apr 20 11:03:53 lxkubenode01 kubelet[9103]: W0420 11:03:53.217805    9103 volume_linux.go:49] Setting volume ownership for /var/lib/kubelet/pods/2f69fc08-f4c9-447b-8cb2-bf737a2d9136/volumes/kubernetes.io~secret/dataprotection-certs and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699
```

I initialized the cluster with kubeadm init (no modifications)

Not to have that many kubelet warnings in the system logs

Environment:

- Kubernetes version (use kubectl version):
```sh
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:56:40Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:48:36Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```
- Cloud provider or hardware configuration: `kubeadm on Ubuntu 18.04.4 LTS (GNU/Linux 4.15.0-96-generic x86_64) amd64`

- OS (e.g: cat /etc/os-release):
```sh
$ cat /etc/os-release
NAME="Ubuntu"
VERSION="18.04.4 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.4 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic
```

- Kernel (e.g. uname -a): `Linux lxkubemaster01 4.15.0-96-generic #97-Ubuntu SMP Wed Apr 1 03:25:46 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`

- Install tools: `kubeadm`

jvanz:
**What type of PR is this?**

/kind cleanup

If a pod has a configmap/secret volume an annoying message shows up in the log approximately every 70 seconds. This happens because the desiredStateOfWorldPopulator sync loop always call the MarkRemountRequired. The function finds the volume plugin and check if the plugin requires mount. Configmap and secret plugins always returns true for that. Thus, the
reconciler code of the volume manager remounts the volume every time. This commit decrease the log level of that message  in the mount function from warning to V4.

Fixes #90293 

```release-note
Fixes the flooding warning messages about setting volume ownership for configmap/secret volumes
```

### DIFF
--- a/pkg/volume/volume_linux.go
+++ b/pkg/volume/volume_linux.go
@@ -23,6 +23,7 @@ import (
 	"syscall"
 
 	"os"
+	"time"
 
 	v1 "k8s.io/api/core/v1"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
@@ -46,7 +47,10 @@ func SetVolumeOwnership(mounter Mounter, fsGroup *int64, fsGroupChangePolicy *v1
 
 	fsGroupPolicyEnabled := utilfeature.DefaultFeatureGate.Enabled(features.ConfigurableFSGroupPolicy)
 
-	klog.Warningf("Setting volume ownership for %s and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699", mounter.GetPath())
+	timer := time.AfterFunc(30*time.Second, func() {
+		klog.Warningf("Setting volume ownership for %s and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699", mounter.GetPath())
+	})
+	defer timer.Stop()
 
 	// This code exists for legacy purposes, so as old behaviour is entirely preserved when feature gate is disabled
 	// TODO: remove this when ConfigurableFSGroupPolicy turns GA.
--- a/pkg/volume/volume_linux.go
+++ b/pkg/volume/volume_linux.go
@@ -46,7 +46,7 @@ func SetVolumeOwnership(mounter Mounter, fsGroup *int64, fsGroupChangePolicy *v1
 
 	fsGroupPolicyEnabled := utilfeature.DefaultFeatureGate.Enabled(features.ConfigurableFSGroupPolicy)
 
-	klog.V(3).Infof("Setting volume ownership for %s and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699", mounter.GetPath())
+	klog.Warningf("Setting volume ownership for %s and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699", mounter.GetPath())
 
 	// This code exists for legacy purposes, so as old behaviour is entirely preserved when feature gate is disabled
 	// TODO: remove this when ConfigurableFSGroupPolicy turns GA.
--- a/pkg/volume/volume_linux.go
+++ b/pkg/volume/volume_linux.go
@@ -46,7 +46,7 @@ func SetVolumeOwnership(mounter Mounter, fsGroup *int64, fsGroupChangePolicy *v1
 
 	fsGroupPolicyEnabled := utilfeature.DefaultFeatureGate.Enabled(features.ConfigurableFSGroupPolicy)
 
-	klog.Warningf("Setting volume ownership for %s and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699", mounter.GetPath())
+	klog.V(3).Infof("Setting volume ownership for %s and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699", mounter.GetPath())
 
 	// This code exists for legacy purposes, so as old behaviour is entirely preserved when feature gate is disabled
 	// TODO: remove this when ConfigurableFSGroupPolicy turns GA.


SergeyKanzhelev:
For the reference, this is PR where the warning was introduced: https://github.com/kubernetes/kubernetes/pull/83426/files

pkg/volume/volume_linux.go
```diff
@@ -46,7 +46,7 @@ func SetVolumeOwnership(mounter Mounter, fsGroup *int64, fsGroupChangePolicy *v1
 
 	fsGroupPolicyEnabled := utilfeature.DefaultFeatureGate.Enabled(features.ConfigurableFSGroupPolicy)
 
-	klog.Warningf("Setting volume ownership for %s and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699", mounter.GetPath())
+	klog.V(3).Infof("Setting volume ownership for %s and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699", mounter.GetPath())
 
 	// This code exists for legacy purposes, so as old behaviour is entirely preserved when feature gate is disabled
```
SergeyKanzhelev, line 51: nit: since you are making change in this file, can you please also fix this spelling issue:

```suggestion
	// This code exists for legacy purposes, so as old behavior is entirely preserved when feature gate is disabled
```

pkg/volume/volume_linux.go
```diff
@@ -46,7 +46,7 @@ func SetVolumeOwnership(mounter Mounter, fsGroup *int64, fsGroupChangePolicy *v1
 
 	fsGroupPolicyEnabled := utilfeature.DefaultFeatureGate.Enabled(features.ConfigurableFSGroupPolicy)
 
-	klog.Warningf("Setting volume ownership for %s and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699", mounter.GetPath())
```
SergeyKanzhelev, line 49: I wonder if it would be better to log this after certain timeout passed instead of decreasing log level

pkg/volume/volume_linux.go
php-coder: I'm not sure that it worth to be fixed BTW: https://ell.stackexchange.com/questions/202548/what-is-the-difference-between-behavior-and-behaviour

pkg/volume/volume_linux.go
jvanz: Thanks! This is my very first PR and I have a question about what's this actually means. Do you believe it's better to have a timeout mechanism witch will print the warning message after `x`  time if this function is called (even if it is called multiples times) or something like that? If so, is there already a mechanism in place to do something like that?

pkg/volume/volume_linux.go
SergeyKanzhelev: thanks for the info. Recommended based on VSCode spell checker =) I guess it uses American English than

pkg/volume/volume_linux.go
SergeyKanzhelev: I'm new too. I looked over the PR introduced this and the related issue and it looks like something that makes many people wonder why the pod has failed to start while it was mounting. Changing the log level will only help troubleshoot the issue when you can repro it consistently and you will need to change the logging level for kubelet. Thus I was thinking if this is really the issue that hard to troubleshoot, adding timeout might be a better solution. I haven't seen this pattern in k8s codebase yet.

jsafrane: Logging the message after X seconds if the fsgroup change has not finished yet is actually a good idea. @gnufied, what do you think?

jsafrane: Alternatively, SetVolumeOwnership could get a parameter that says how noisy it can be - it *should* log the message at higher level for "real" persistent storage (it can be slow) and not log anything for secrets/configmaps.

pkg/volume/volume_linux.go
jvanz: Ok, marking this as resolved.

pkg/volume/volume_linux.go
SergeyKanzhelev: @kubernetes/sig-instrumentation-misc do you have a recommendation on this? Is there pattern to follow on reporting warning in case the method execution takes too long?

serathius: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md

jvanz: I've updated the code to print the warning only after a timeout. I do not know if the 30 seconds is the adequate time period. Let me know if I should change it.

pkg/volume/volume_linux.go
```diff
@@ -46,23 +47,29 @@ func SetVolumeOwnership(mounter Mounter, fsGroup *int64, fsGroupChangePolicy *v1
...
 
 	if skipPermissionChange(mounter, fsGroup, fsGroupChangePolicy) {
 		klog.V(3).Infof("skipping permission and ownership change for volume %s", mounter.GetPath())
+		timer.Stop()
 		return nil
 	}
 
 	return walkDeep(mounter.GetPath(), func(path string, info os.FileInfo, err error) error {
 		if err != nil {
+			timer.Stop()
```
Klaven: I'm not sure you want to call timer.Stop() every time the func gets called. I think the intention here was to call it when the `walkDeep` object was done doing whatever it is that it does. It might be that you wanted to give `walkDeep` a reference too the timer so it can call stop when it's done.

Klaven (commented):
First, welcome too kubernetes! And thanks for putting in the work to get this far! Here are some of my thoughts.

from my point of view this is not a terribly clean solution with a maintenance burden (`timer.Stop()` must be called on return) And this can't be solved by a `defer` because it has to be called outside the scope of this function.

In my opinion it might make since to log this in the function that calls `SetVolumeOwnership()` This puts the responsibility in the hands of the function that makes this call to know when it is done as `SetVolumeOwnership` does not know when it is done because it returns a function.

That being said a flag could also be used instead of a timer.

I personally would also like to see a test function that ensures the timer is eventually stopped, because this is the type of bug you could have a regression with and not even know/see it.

pkg/volume/volume_linux.go
jvanz: Yeah, this is wrong. Thanks for bring my attention for that. I just update the PR removing all the calls and adding a just single `defer` call.

pkg/volume/volume_linux.go
```diff
@@ -46,7 +47,10 @@ func SetVolumeOwnership(mounter Mounter, fsGroup *int64, fsGroupChangePolicy *v1
 
 	fsGroupPolicyEnabled := utilfeature.DefaultFeatureGate.Enabled(features.ConfigurableFSGroupPolicy)
 
-	klog.Warningf("Setting volume ownership for %s and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699", mounter.GetPath())
+	timer := time.AfterFunc(30*time.Second, func() {
+		klog.Warningf("Setting volume ownership for %s and fsGroup set. If the volume has a lot of files then setting volume ownership could be slow, see https://github.com/kubernetes/kubernetes/issues/69699", mounter.GetPath())
+	})
+	defer timer.Stop()
```
Klaven, line 53: I stand corrected. defer should work as `walkDeep` is executed before returning. meaning defer would only happen after. this lgtm now.

Klaven (commented):
defer looks good here I was wrong as I thought walkDeep was returned but it is executed. this is a clean solution. looks good to me.

pkg/volume/volume_linux.go
jvanz: Thanks! :)

liggitt:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92878/pull-kubernetes-node-e2e/1285390922829271040 flake is #93210

jvanz:
Should we set the 1.19 milestone?

ehashman (approved):
/approve

I've also seen this in 1.17, agreed it's annoying.

ehashman:
/milestone v1.20

We're deep into the 1.19 code freeze, this can wait until thaw

liggitt:

msschl:
Can we get this as patch for v1.18.9?

jvanz:
> Can we get this as patch for v1.18.9?

I believe it will be easy to back port the change. It's quite simple. I just do not know what is the process for that and if it will be accepted.

ReillyTevera:
@jvanz [This](https://kubernetes.io/docs/contribute/generate-ref-docs/contribute-upstream/#cherry-picking-your-commit-into-a-release-branch) is the document you need to read. 

Note that the standards for getting into release branches that are already deep into the point releases is higher and is pretty much limited to functionality bugfixes. This is a more cosmetic change and might not be accepted (but it depends on the 1.18 team reviewer you get so I'd still recommend giving it a try). You'd probably have an easier time getting it into 1.19.

ehashman:
Backport applies cleanly to 1.18/1.19 but not to 1.17 (where this was apparently introduced). Opened PRs.

ehashman:
If someone wants to give the 1.17 backport a swing, feel free, but idk if we will get approvals :)

msschl:
A huge thanks to all participants especially to @jvanz @ehashman for your work. Thanks a lot 👍

PR #92888, opened 2020-07-07
Add caller information to apiserver konnectivity egress

Jefftree:
**What type of PR is this?**

/kind feature

Add caller information when making egress requests to the konnectivity network proxy.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/cmd/kube-apiserver/app/server.go
+++ b/cmd/kube-apiserver/app/server.go
@@ -395,7 +395,7 @@ func CreateKubeAPIServerConfig(
 
 		// Use the config.GenericConfig.EgressSelector lookup as the transport used by the "proxy" subresources.
 		networkContext := egressselector.Cluster.AsNetworkContext()
-		dialer, err := config.GenericConfig.EgressSelector.Lookup(networkContext)
+		dialer, err := config.GenericConfig.EgressSelector.Lookup(networkContext, "proxy")
 		if err != nil {
 			return nil, nil, nil, nil, err
 		}
@@ -558,7 +558,7 @@ func BuildAuthorizer(s *options.ServerRunOptions, EgressSelector *egressselector
 	authorizationConfig := s.Authorization.ToAuthorizationConfig(versionedInformers)
 
 	if EgressSelector != nil {
-		egressDialer, err := EgressSelector.Lookup(egressselector.Master.AsNetworkContext())
+		egressDialer, err := EgressSelector.Lookup(egressselector.Master.AsNetworkContext(), "authz-webhook")
 		if err != nil {
 			return nil, nil, err
 		}
--- a/staging/src/k8s.io/apiserver/pkg/server/egressselector/egress_selector.go
+++ b/staging/src/k8s.io/apiserver/pkg/server/egressselector/egress_selector.go
@@ -43,7 +43,7 @@ var directDialer utilnet.DialFunc = http.DefaultTransport.(*http.Transport).Dial
 
 // EgressSelector is the map of network context type to context dialer, for network egress.
 type EgressSelector struct {
-	egressToDialer map[EgressType]utilnet.DialFunc
+	egressToDialer map[EgressType]func(string) utilnet.DialFunc
 }
 
 // EgressType is an indicator of which egress selection should be used for sending traffic.
@@ -68,7 +68,7 @@ type NetworkContext struct {
 }
 
 // Lookup is the interface to get the dialer function for the network context.
-type Lookup func(networkContext NetworkContext) (utilnet.DialFunc, error)
+type Lookup func(networkContext NetworkContext, caller string) (utilnet.DialFunc, error)
 
 // String returns the canonical string representation of the egress type
 func (s EgressType) String() string {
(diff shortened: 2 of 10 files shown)


cmd/kube-apiserver/app/server.go
```diff
@@ -395,7 +395,7 @@ func CreateKubeAPIServerConfig(
 
 		// Use the config.GenericConfig.EgressSelector lookup as the transport used by the "proxy" subresources.
 		networkContext := egressselector.Cluster.AsNetworkContext()
-		dialer, err := config.GenericConfig.EgressSelector.Lookup(networkContext)
+		dialer, err := config.GenericConfig.EgressSelector.Lookup(networkContext, "proxy")
```
Jefftree, line 398: We *could* make these separate constants that are looked up in the egress selector file (similar to Master and Cluster). Not sure if that will improve the clarity, let me know your thoughts.

cheftako:
Is this still a work in progress?

cmd/kube-apiserver/app/server.go
cheftako: +1 I would prefer enums. We can get better type safety that way.

staging/src/k8s.io/apiserver/pkg/server/egressselector/egress_selector.go
```diff
@@ -150,22 +150,22 @@ type grpcProxier struct {
 	tunnel client.Tunnel
 }
 
-func (g *grpcProxier) proxy(addr string) (net.Conn, error) {
+func (g *grpcProxier) proxy(addr, caller string) (net.Conn, error) {
 	return g.tunnel.Dial("tcp", addr)
```
cheftako, line 154: So on the GRPC connection we are just dropping the caller info on the floor? (for now)

staging/src/k8s.io/apiserver/pkg/server/egressselector/egress_selector.go
```diff
@@ -217,26 +217,30 @@ type metricsOptions struct {
...
+			defer trace.LogIfLong(500 * time.Millisecond)
+			start := egressmetrics.Metrics.Clock().Now()
+			proxier, err := d.connector.connect(caller)
+			if err != nil {
+				egressmetrics.Metrics.ObserveDialFailure(d.options.protocol, d.options.transport, egressmetrics.StageConnect)
+				return nil, err
+			}
+			conn, err := proxier.proxy(addr, caller)
+			if err != nil {
+				egressmetrics.Metrics.ObserveDialFailure(d.options.protocol, d.options.transport, egressmetrics.StageProxy)
```
cheftako, line 238: Would probably be good to use the caller as a label on the metric. We need to be careful that labels on metrics don't get large but that doesn't seem an issue for caller. If we introduced an enum we could put a warning on it.

staging/src/k8s.io/apiserver/pkg/server/egressselector/egress_selector.go
```diff
@@ -352,17 +356,17 @@ func NewEgressSelector(config *apiserver.EgressSelectorConfiguration) (*EgressSe
...
 // Lookup gets the dialer function for the network context.
 // This is configured for the Kubernetes API Server at startup.
-func (cs *EgressSelector) Lookup(networkContext NetworkContext) (utilnet.DialFunc, error) {
+func (cs *EgressSelector) Lookup(networkContext NetworkContext, caller string) (utilnet.DialFunc, error) {
 	if cs.egressToDialer == nil {
 		// The round trip wrapper will over-ride the dialContext method appropriately
 		return nil, nil
 	}
-	return cs.egressToDialer[networkContext.EgressSelectionName], nil
+	return cs.egressToDialer[networkContext.EgressSelectionName](caller), nil
```
cheftako, line 371: This really emphasizes that this does not play well with connection/tunnel reuse. What is our plan for reusing a connection for different callers?

Jefftree: We already create a separate tunnel (apiserver <-> proxy server) for each request so this should have no impact on tunnel reuse. (Note that proxy server <-> proxy agent tunnel is always reused/multiplexed) @caesarxuchao performed a performance test with reusing the tunnel between apiserver and proxy server and we didn't observe any noticeable performance improvement while a bunch of complexity was added.

staging/src/k8s.io/apiserver/pkg/server/egressselector/egress_selector.go
Jefftree: This was from a limitation from the way we set up the http connect and grpc tunnels. http connect propagates the user agent on the PROXY function while grpc propagates the user agent on the CONNECT function. L202 captures this propagation for grpc. Since the interface needs to be consistent between the two implementations, I thought it'd be appropriate to pass in the caller as a parameter to both the connect and proxy function and have one of the functions perform a noop.

staging/src/k8s.io/apiserver/pkg/server/egressselector/egress_selector.go
cheftako: Which is a great reason to not make that change now. However there is a difference between saying it doesn't make sense now and it doesn't make sense.

ehashman:
This needs a rebase.

ehashman:
@Jefftree please reopen once you've had a chance to rebase. As this PR is >6mo old with no updates, I am going to close it for now.

/close

PR #92876, opened 2020-07-07
Add pdbLister as a member field of struct DefaultPreemption

Huang-Wei:
**What type of PR is this?**

/kind cleanup
/sig scheduling

**What this PR does / why we need it**:

Per https://github.com/kubernetes/kubernetes/issues/92781#issuecomment-653651001, it's a good practice to keep a xyzLister inside the plugin struct, and initialize that xyzLister in the initialization function. In this way, we use xyzLister (instead of informerFactory....lister()) in the plugin logic as that's guaranteed to have been initialized.

**Which issue(s) this PR fixes**:

Part of #92781

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #92890, opened 2020-07-07
Refactor and expose common preemption functions

Huang-Wei:
**What type of PR is this?**

/kind cleanup
/sig scheduling

This PR divides the whole preemption flow into 5 public functions:

- **PodEligibleToPreemptOthers**: Ensure the preemptor is eligible to preempt other pods. Out-of-tree plugins can either customize or reuse it.
- **FindCandidates:** Calculate the preemption candidates. Out-of-tree plugins may rewrite this in their implementation.
- **CallExtenders:** Interact with extenders. Usually, out-of-tree plugins just call it as-is.
- **SelectBestCandidate:** Pick the best-fit candidate from preemption candidates. Out-of-tree plugins may rewrite this in their implementation.
- **PreNominateCandidate:** Perform preparation work before nominating the candidate winner. Usually, out-of-tree plugins just call it as-is.

Other notable changes:

- `selectNodesForPreemption` is renamed to `dryRunPreemption`
- a `Candidate` interface is introduced. Out-of-tree plugins can have their own implementation.
- map `nodeNameToVictims` is refactored to `[]Candidate` as we used to assume the nominated node name is unique, but with the same nominated node, there could be more than one candidate, so using a slice instead of map to represent all candidates would make more sense.

Part of #91038.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/framework/plugins/defaultpreemption/candidate.go
+++ b/pkg/scheduler/framework/plugins/defaultpreemption/candidate.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package defaultpreemption
+
+import (
+	extenderv1 "k8s.io/kube-scheduler/extender/v1"
+)
+
+// Candidate represents a nominated node on which the preemptor can be scheduled,
+// along with the list of victims that should be evicted for the preemptor to fit the node.
+type Candidate interface {
+	// Victims wraps a list of to-be-preempted Pods and the number of PDB violation.
+	Victims() *extenderv1.Victims
+	// Name returns the target node name where the preemptor gets nominated to run.
+	Name() string
+}
+
+type candidate struct {
+	victims *extenderv1.Victims
+	name    string
+}
+
+// Victims returns s.victims.
+func (s *candidate) Victims() *extenderv1.Victims {
+	return s.victims
+}
+
+// Name returns s.name.
+func (s *candidate) Name() string {
+	return s.name
+}
--- a/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
+++ b/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
 	policylisters "k8s.io/client-go/listers/policy/v1beta1"
 	extenderv1 "k8s.io/kube-scheduler/extender/v1"
 	podutil "k8s.io/kubernetes/pkg/api/v1/pod"
@@ -101,6 +102,10 @@ func (pl *DefaultPreemption) PostFilter(ctx context.Context, state *framework.Cy
 // before it is retried after many other pending pods.
 func (pl *DefaultPreemption) preempt(ctx context.Context, state *framework.CycleState, pod *v1.Pod, m framework.NodeToStatusMap) (string, error) {
 	cs := pl.fh.ClientSet()
+	ph := pl.fh.PreemptHandle()
+	nodeLister := pl.fh.SnapshotSharedLister().NodeInfos()
+
+	// 0) Fetch the latest version of <pod>.
 	// TODO(Huang-Wei): get pod from informer cache instead of API server.
 	pod, err := util.GetUpdatedPod(cs, pod)
 	if err != nil {
@@ -108,17 +113,51 @@ func (pl *DefaultPreemption) preempt(ctx context.Context, state *framework.Cycle
 		return "", err
 	}
 
-	if !podEligibleToPreemptOthers(pod, pl.fh.SnapshotSharedLister().NodeInfos(), m[pod.Status.NominatedNodeName]) {
+	// 1) Ensure the preemptor is eligible to preempt other pods.
+	if !PodEligibleToPreemptOthers(pod, nodeLister, m[pod.Status.NominatedNodeName]) {
 		klog.V(5).Infof("Pod %v/%v is not eligible for more preemption.", pod.Namespace, pod.Name)
 		return "", nil
 	}
-	allNodes, err := pl.fh.SnapshotSharedLister().NodeInfos().List()
+
+	// 2) Find all preemption candidates.
+	candidates, err := FindCandidates(ctx, cs, state, pod, m, ph, nodeLister, pl.pdbLister)
+	if err != nil || len(candidates) == 0 {
+		return "", err
+	}
+
+	// 3) Interact with registered Extenders to filter out some candidates if needed.
+	candidates, err = CallExtenders(ph.Extenders(), pod, nodeLister, candidates)
 	if err != nil {
 		return "", err
 	}
-	if len(allNodes) == 0 {
-		return "", core.ErrNoNodesAvailable
+
+	// 4) Find the best candidate.
+	bestCandidate := SelectCandidate(candidates)
+	if bestCandidate == nil || len(bestCandidate.Name()) == 0 {
+		return "", nil
 	}
+
+	// 5) Perform preparation work before nominating the selected candidate.
+	if err := PrepareCandidate(bestCandidate, pl.fh, cs, pod); err != nil {
+		return "", err
+	}
+
+	return bestCandidate.Name(), nil
+}
+
+// FindCandidates calculates a slice of preemption candidates.
+// Each candidate is executable to make the given <pod> schedulable.
+func FindCandidates(ctx context.Context, cs kubernetes.Interface, state *framework.CycleState, pod *v1.Pod,
+	m framework.NodeToStatusMap, ph framework.PreemptHandle, nodeLister framework.NodeInfoLister,
+	pdbLister policylisters.PodDisruptionBudgetLister) ([]Candidate, error) {
+	allNodes, err := nodeLister.List()
+	if err != nil {
+		return nil, err
+	}
+	if len(allNodes) == 0 {
+		return nil, core.ErrNoNodesAvailable
+	}
+
 	potentialNodes := nodesWherePreemptionMightHelp(allNodes, m)
 	if len(potentialNodes) == 0 {
 		klog.V(3).Infof("Preemption will not help schedule pod %v/%v on any node.", pod.Namespace, pod.Name)
@@ -127,7 +166,7 @@ func (pl *DefaultPreemption) preempt(ctx context.Context, state *framework.Cycle
 			klog.Errorf("Cannot clear 'NominatedNodeName' field of pod %v/%v: %v", pod.Namespace, pod.Name, err)
 			// We do not return as this error is not critical.
 		}
-		return "", nil
+		return nil, nil
 	}
 	if klog.V(5).Enabled() {
 		var sample []string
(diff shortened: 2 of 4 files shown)


Huang-Wei:
/hold
this PR is built on top of #92876.

/assign @ahg-g 
/cc @everpeace @denkensk

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -254,46 +245,97 @@ func selectNodesForPreemption(
...
-				}
-
-				// Replace nodeNameToVictims with new result after preemption. So the
-				// rest of extenders can continue use it as parameter.
-				nodeNameToVictims = newNodeNameToVictims
+// CallExtenders call given <extenders> to filter out <strategies>, and return a filtered strategies.
+// We will only check <strategies> with extenders that support preemption.
+// Extenders which do not support preemption may later prevent preemptor from being scheduled on the nominated
+// node. In that case, scheduler will find a different host for the preemptor in subsequent scheduling cycles.
+func CallExtenders(extenders []framework.Extender, pod *v1.Pod, nodeLister framework.NodeInfoLister,
```
everpeace, line 264: perhaps it would have a more descriptive name?  for example, `CallPreemptExtenders`??

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -254,46 +245,97 @@ func selectNodesForPreemption(
...
+			victims:           victimsMap[nodeName],
+			nominatedNodeName: nodeName,
+		})
+	}
+	return newStrategies, nil
+}
+
+// This function is not applicable for out-of-tree preemption plugins that exercise
+// different strategies on the same nominated node.
+func strategiesToVictimsMap(strategies []Strategy) map[string]*extenderv1.Victims {
```
everpeace, line 311: how about exposing this method because both input/output types are public??  And I think this method is useful for future implementors of custom preemption plugins.

everpeace (commented):
Thanks for pinging me.  I put some comments about naming/method scopes.

ahg-g (commented):
I like this refactoring, but I have an issue with the name Strategy, it took me a while to map such an abstract concept to what we really mean by it.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: SG.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: Ah, I don't want this to be exposed to confuse users as it's specifically needed for defaultpreemption to adapt to the Extender interface.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -256,46 +245,97 @@ func selectNodesForPreemption(
...
+	candidateNode := pickOneNodeForPreemption(victimsMap)
+
+	// Same as strategiesToVictimsMap, this logic is not applicable for out-of-tree
+	// preemption plugins that exercise different strategies on the same nominated node.
+	for _, strategy := range strategies {
+		if candidateNode == strategy.NominatedNodeName() {
+			return strategy
+		}
+	}
+	return strategies[0]
```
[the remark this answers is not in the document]
Huang-Wei: Correct. Updated.

pkg/scheduler/framework/plugins/defaultpreemption/strategy.go
```diff
@@ -0,0 +1,46 @@
...
+package defaultpreemption
+
+import (
+	extenderv1 "k8s.io/kube-scheduler/extender/v1"
+)
+
+// Strategy defines a number of operations to execute a preemption strategy.
+// It's executed by preempting the victims, set the nominatedName for the preemptor,
+// as well as other operations to maintain the nominated Pods properly.
+type Strategy interface {
```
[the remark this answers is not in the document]
Huang-Wei: `Candidate` sounds more descriptive. Updated.

Huang-Wei:
@everpeace @ahg-g Thanks for the review. PTAL.

(this PR should still be /hold until #92876 gets merged.

pkg/scheduler/framework/plugins/defaultpreemption/candidate.go
```diff
@@ -20,27 +20,26 @@ import (
 	extenderv1 "k8s.io/kube-scheduler/extender/v1"
 )
 
-// Strategy defines a number of operations to execute a preemption strategy.
-// It's executed by preempting the victims, set the nominatedName for the preemptor,
-// as well as other operations to maintain the nominated Pods properly.
-type Strategy interface {
+// Candidate represents a nominated node on which the preemptor can be scheduled,
+// along with the cost of preemption a number of victim Pods.
```
ahg-g, line 24:
```suggestion
// along with the list of victims that should be evicted for the pod to fit the node.
```

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -113,36 +113,37 @@ func (pl *DefaultPreemption) preempt(ctx context.Context, state *framework.Cycle
...
+	// 3) Find the best candidate.
+	bestCandidate := SelectBestCandidate(candidates)
+	if bestCandidate == nil || len(bestCandidate.Name()) == 0 {
 		return "", nil
 	}
 
-	// 4) Execute the best-fit PreemptionStrategy.
-	if err := ExecuteStrategy(strategy, pl.fh, cs, pod); err != nil {
+	// 4) Apply the candidate winner.
+	if err := ApplyCandidate(bestCandidate, pl.fh, cs, pod); err != nil {
```
ahg-g, line 135: actually lets name the function `NominateCandidate`

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -582,9 +586,9 @@ func selectVictimsOnNode(
 	return victims, numViolatingVictim, true
 }
 
-// ExecuteStrategy executes the given strategy.
-func ExecuteStrategy(s Strategy, fh framework.FrameworkHandle, cs kubernetes.Interface, pod *v1.Pod) error {
-	for _, victim := range s.Victims().Pods {
+// ApplyCandidate applies the candidate winner.
```
ahg-g, line 589:
```suggestion
// NominateCandidate makes a candidate node the nominee by evicting victim pods. Note that NominatedNodeName is not set in this function, it is set in a single update to the pod that includes updating the pod condition.
```

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -245,33 +246,33 @@ func dryRunPreemption(ctx context.Context, fh framework.PreemptHandle, state *fr
...
 		}
 	}
 	parallelize.Until(ctx, len(potentialNodes), checkNode)
-	return strategies
+	return candidates
 }
 
-// CallExtenders call given <extenders> to filter out <strategies>, and return a filtered strategies.
-// We will only check <strategies> with extenders that support preemption.
+// CallPreemptExtenders call given <extenders> to filter out <candidates>, and return a filtered candidates.
```
ahg-g, line 261:
```suggestion
// CallPreemptExtenders call given <extenders> to select the list of feasible candidates.
```

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -245,33 +246,33 @@ func dryRunPreemption(ctx context.Context, fh framework.PreemptHandle, state *fr
...
+// We will only check <candidates> with extenders that support preemption.
 // Extenders which do not support preemption may later prevent preemptor from being scheduled on the nominated
 // node. In that case, scheduler will find a different host for the preemptor in subsequent scheduling cycles.
-func CallExtenders(extenders []framework.Extender, pod *v1.Pod, nodeLister framework.NodeInfoLister,
-	strategies []Strategy) ([]Strategy, error) {
+func CallPreemptExtenders(extenders []framework.Extender, pod *v1.Pod, nodeLister framework.NodeInfoLister,
```
ahg-g, line 265: I disagree with the name change, I think "`CallPreemptExtenders`" doesn't really read well, also this function is in the preemption plugin, so there is enough context to know what `CallExtenders` mean.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
everpeace: oh, I misunderstood.  this should be private because this method is called only in public methods. 🙇

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -97,26 +100,63 @@ func (pl *DefaultPreemption) PostFilter(ctx context.Context, state *framework.Cy
...
+	// 4) Apply the candidate winner.
+	if err := ApplyCandidate(bestCandidate, pl.fh, cs, pod); err != nil {
+		return "", err
+	}
+
+	return bestCandidate.Name(), nil
+}
+
+// FindCandidates calculates a slice of preemption candidates.
+// Each candidate is executable to make the given <pod> schedulable.
```
chendave, line 143:
```suggestion
// FindCandidates returns a slice of preemption candidates.
// Each candidate is able to make the given <pod> schedulable.
```

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -97,26 +100,63 @@ func (pl *DefaultPreemption) PostFilter(ctx context.Context, state *framework.Cy
...
+
+	// 2) Interact with registered Extenders to filter out some candidates if needed.
+	candidates, err = CallPreemptExtenders(ph.Extenders(), pod, nodeLister, candidates)
 	if err != nil {
 		return "", err
 	}
+
+	// 3) Find the best candidate.
+	bestCandidate := SelectBestCandidate(candidates)
+	if bestCandidate == nil || len(bestCandidate.Name()) == 0 {
```
chendave, line 130: Can we get into this situation?
```len(bestCandidate.Name()) == 0```

File: pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go [hunk 12]
```diff
@@ -97,26 +100,63 @@ func (pl *DefaultPreemption) PostFilter(ctx context.Context, state *framework.Cy
...
+
+	return bestCandidate.Name(), nil
+}
+
+// FindCandidates calculates a slice of preemption candidates.
+// Each candidate is executable to make the given <pod> schedulable.
+func FindCandidates(ctx context.Context, cs kubernetes.Interface, state *framework.CycleState, pod *v1.Pod,
+	m framework.NodeToStatusMap, ph framework.PreemptHandle, nodeLister framework.NodeInfoLister,
+	pdbLister policylisters.PodDisruptionBudgetLister) ([]Candidate, error) {
+	if !podEligibleToPreemptOthers(pod, nodeLister, m[pod.Status.NominatedNodeName]) {
```
denkensk, line 147: I think In terms of logic keeping the `podEligibleToPreemptOthers judgement` in `func preempt` is better than into `FindCandidates`.  WDYT?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
denkensk: How about `FindCandidatesWithExtenders` similar to the previous name ？:smirk:

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
denkensk: I think the `is executable to` is better than `is scheduable for`.  `is executable to` means to  `Make room` for the given pod. smile: WDYT? @chendave

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: Actually this function doesn't really "nominate" - which is done in the ErrorFunc. So how about `PreNominateCandidate`?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: I think `FindCandidatesWithExtenders` is a bit verbose... I'd like to revert to `CallExtenders`.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: Agree with @denkensk - the pod is made schedulable with a cost of evicting victims. "is schedulable" didn't highlight that.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: Usually no, just as an extra safety check.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: SGTM. Users may customize `podEligibleToPreemptOthers` as well.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -97,26 +100,64 @@ func (pl *DefaultPreemption) PostFilter(ctx context.Context, state *framework.Cy
...
 		return "", err
 	}
+
+	// 4) Find the best candidate.
+	bestCandidate := SelectBestCandidate(candidates)
+	if bestCandidate == nil || len(bestCandidate.Name()) == 0 {
+		return "", nil
+	}
+
+	// 5) Perform preparation work before nominating the candidate winner.
```
ahg-g, line 140:
```suggestion
	// 5) Perform preparation work before nominating the selected candidate.
```

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -539,6 +586,38 @@ func selectVictimsOnNode(
 	return victims, numViolatingVictim, true
 }
 
+// PreNominateCandidate does some preparation work before nominating the candidate winner:
+// - Evict the victim pods
+// - Reject the victim pods if they are in waitingPod map
+// - Clear the low-priority pods' nominatedNodeName status if needed
+func PreNominateCandidate(c Candidate, fh framework.FrameworkHandle, cs kubernetes.Interface, pod *v1.Pod) error {
```
ahg-g, line 593: Perhaps `PrepareCandidate`?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -254,46 +247,100 @@ func selectNodesForPreemption(
...
+func candidatesToVictimsMap(candidates []Candidate) map[string]*extenderv1.Victims {
+	m := make(map[string]*extenderv1.Victims)
+	for _, c := range candidates {
+		m[c.Name()] = c.Victims()
+	}
+	return m
+}
+
+// SelectBestCandidate chooses the best-fit candidate from given <candidates> and return it.
+func SelectBestCandidate(candidates []Candidate) Candidate {
```
ahg-g, line 322: just `SelectCandidate`?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: "winner" is informal huh :)

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: Given that the only thing we did after this function is nominating the candidate, I think "PreNominate" is more descriptive than "Prepare".

ahg-g: There is no function or canonical phase named NominateCandidate, and so it might be confusing what the prefix "pre" mean here. I still think PrepareCandidate (which is really what we are doing here, changing the node state to accomodate the pod) is the better option and matches well with the previous phases: find, select and prepare.

Huang-Wei: ok, "prepare" should be good enough.

ahg-g:
Thanks, please squash

Huang-Wei:
> Thanks, please squash

@ahg-g Needs to wait for #92876 to be merged first. (not sure why it stays submerged for days...)

Issue #89701, opened 2020-03-31
Promote kubescheduler.config.k8s.io to beta


alculquicondor:
**What would you like to be added**:
Promote `kubescheduler.config.k8s.io` to Beta by introducing `kubescheduler.config.k8s.io/v1beta1`

/sig scheduling
/wg component-standard

**Why is this needed**:

kube-scheduler has shipped with support for `kubescheduler.config.k8s.io/v1alpha1` since k8s 1.12 and we shipped version `v1alpha2` since k8s 1.18
The most significant changes have been around supporting plugins ([KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/20180409-scheduling-framework.md)) and scheduling profiles ([KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/1451-multi-scheduling-profiles))

We believe it's time to commit to the support of this API by promoting it to beta. This way, more users will be more comfortable incorporating it in their setups.

**Task tracking**:

These are the tasks that need to be finished before 1.19 freeze:

- [x] Write KEP
- [x] Introduce `kubescheduler.config.k8s.io/v1beta1` #91420
- [x] Remove `.bindTimeoutSeconds` once we have a volume binding plugin #83726 #90958
- [x] Merge `Reserve` and `Unreserve` in a single PluginSet #90724

Nice to have:
- [x] Renaming DefaultPodTopologySpread plugin
 #91994
- [x] PostFilter extension point #92012
- [x] Remove case-insensitive parsing of Extenders #65414
- [ ] ~Generate reference docs #88919 kubernetes-sigs/reference-docs#138~
- [x] Add types for scheduling plugin Args #87784
- [ ]  ~Add Feature gates #84922~
- [x] Remove v1alpha1 #89260

alculquicondor:
cc @liggitt @lavalamp @mtaufen @ahg-g @Huang-Wei

ahg-g:
I am not sure " Add types for scheduling plugin Args #87784" is part of this, we can always add those  later, but they are not blockers.

alculquicondor:
I would say having the versioned args is a good sign of maturity of the API. We are also in good track with it, so I wouldn't be worried.

cofyc:
after this is done, do you think it's reasonable to call `Unreserve` if `Reserve` fails?

ahg-g:
> after this is done, do you think it's reasonable to call `Unreserve` if `Reserve` fails?

[...]

PR #92892, opened 2020-07-08
Remove DisablePreemption field from KubeSchedulerConfiguration

Huang-Wei:
**What type of PR is this?**

/kind cleanup
/sig scheduling

With the refactoring of PostFilter extension point, DisablePreemption field in KubeSchedulerConfiguration is no longer needed as that can be deduced from profile's postFilterPlugins.

/ref #91038 #89701.

```release-note
NONE
```

```docs
- [KEP]: https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/785-scheduler-component-config-api
```


pkg/scheduler/scheduler.go
```diff
@@ -481,7 +467,7 @@ func (sched *Scheduler) scheduleOne(ctx context.Context) {
 		// into the resources that were preempted, but this is harmless.
 		nominatedNode := ""
 		if fitError, ok := err.(*core.FitError); ok {
-			if sched.DisablePreemption || !prof.HasPostFilterPlugins() {
+			if !prof.HasPostFilterPlugins() {
 				klog.V(3).Infof("Pod priority feature is not enabled or preemption is disabled by scheduler configuration." +
```
alculquicondor, line 471: Update the log line here. It doesn't seem like Pod priority affect this anymore.

test/integration/scheduler/util.go
```diff
@@ -88,9 +89,19 @@ func initTest(t *testing.T, nsPrefix string, opts ...scheduler.Option) *testutil
 // initTestDisablePreemption initializes a test environment and creates master and scheduler with default
 // configuration but with pod preemption disabled.
 func initTestDisablePreemption(t *testing.T, nsPrefix string) *testutils.TestContext {
+	prof := schedulerconfig.KubeSchedulerProfile{
+		SchedulerName: v1.DefaultSchedulerName,
+		Plugins: &schedulerconfig.Plugins{
+			PostFilter: &schedulerconfig.PluginSet{
+				Disabled: []schedulerconfig.Plugin{
+					{Name: "*"},
```
alculquicondor, line 97: Let's use the name of the plugin, in case we have another default PostFilter plugin in the future.

alculquicondor:
other than the nits, looks good.

/assign @liggitt 
for api review

test/integration/scheduler/util.go
Huang-Wei: Done.

pkg/scheduler/scheduler.go
Huang-Wei: Good catch. Updated.

pkg/scheduler/scheduler.go
```diff
@@ -481,9 +467,8 @@ func (sched *Scheduler) scheduleOne(ctx context.Context) {
 		// into the resources that were preempted, but this is harmless.
 		nominatedNode := ""
 		if fitError, ok := err.(*core.FitError); ok {
-			if sched.DisablePreemption || !prof.HasPostFilterPlugins() {
-				klog.V(3).Infof("Pod priority feature is not enabled or preemption is disabled by scheduler configuration." +
-					" No preemption is performed.")
+			if !prof.HasPostFilterPlugins() {
+				klog.V(3).Infof("No PostFilter plugins are registered, so no preemption will be performed.")
```
liggitt, line 471: This is preexisting, and maybe I'm misunderstanding, but this makes it seem like you have to enable postfilter plugins for preemption to work. That's pretty surprising to me.

liggitt:
API bits look good. Since v1beta1 scheduler config is new this release, the release note doesn't need to say something was removed from v1beta1, this should be added to the docs detailing the differences between v1alpha2 and v1beta1

liggitt:
/approve

for scheduler API change
/retest

pkg/scheduler/scheduler.go
Huang-Wei: > this makes it seem like you have to enable postfilter plugins for preemption to work.

True, it's enabled by default. The background is that scheduler have almost "pluginize" all hard-coded logic into plugins. In 1.19 the defaultpreemption plugin follows that direction.

So for a default scheduler configuration, preemption functions the same as before. But users can still disable preemption by tweaking the config, just like what they did by specifying `DisablePreemption: false`.

Huang-Wei:
Thanks @alculquicondor @liggitt for the review!

> Since v1beta1 scheduler config is new this release, the release note doesn't need to say something was removed from v1beta1, this should be added to the docs detailing the differences between v1alpha2 and v1beta1

Release note removed.

@alculquicondor would you mind adding a bullet to your PR https://github.com/kubernetes/website/pull/21437, in docs/reference/scheduling/config.md#beta-changes?

```
- `.disablePreemption` was removed. Users can disable preemption by disabling the "defaultpreemption" PostFilter plugin.
```

Issue #91506, opened 2020-05-27
Remove the apiserver insecure port


tallclair:
The API server's "insecure-port" has been [deprecated since 1.10](https://github.com/kubernetes/kubernetes/commit/2f175bc43279a1a4552610a42af0a006f4c6fba2), but never had a deadline for removal. Previous efforts to remove the port have been blocked by requirements for http healthz checks (https://github.com/kubernetes/kubernetes/issues/43784).

I propose publishing a deadline to remove the feature in v1.20. This means that in v1.19 we:
1. Update the deprecation notice to indicate that removal will happen in v1.20 (rather than a "future release")
2. Add an [action required] release note to highlight the removal
3. Figure out a path forward to meeting healthz or other requriments

For item 3, there are already out-of-tree solutions available today, such as a healthz-only proxy which can serve the same API on the same port, and authenticates as a user that only has access to healthz.

/cc @justinsb @micahhausler @deads2k @liggitt 
/sig auth api-machinery
/milestone v1.20
/kind cleanup
/priority important-longterm

BenTheElder:
cc @neolit123

liggitt:
I would also consider changing the default in 1.19 to be off (`0`), rather than on (`8080`)

liggitt:
Also, I would recommend leaving the inert flag in place for a few releases even after the insecure serving code is removed. I want to avoid accidents where people stop setting `--insecure-port=0` too early in response to warnings about the flag being removed, and accidentally opt back into the default behavior of enabling the insecure port.

neolit123:
> I would also consider changing the default in 1.19 to be off (0), rather than on (8080).

+1, not enabling the insecure port by default is very reasonable.

knight42:
Currently this is being blocked by #94658, and the fix is ongoing at #94723 .

knight42:
Now #95856 has been merged and apiserver no longer listen on insecure port 🎉 .

[...]

PR #92893, opened 2020-07-08
Disable deprecated insecure port by default

tallclair:
**What type of PR is this?**
/kind deprecation

Disable the deprecated `--insecure-port` by default. The port will be permanently disabled starting in v1.20, and the flag removed in a later release.

For https://github.com/kubernetes/kubernetes/issues/91506

```release-note
kube-apiserver: The `--insecure-port` now defaults to disabled. Set `--insecure-port=8080` to temporarily keep the previous behavior. The flag will be permanently disabled starting in v1.20.
```

/assign @liggitt 
/area security
/milestone v1.19
/sig api-machinery
/sig auth

### DIFF
--- a/cluster/gce/gci/configure-kubeapiserver.sh
+++ b/cluster/gce/gci/configure-kubeapiserver.sh
@@ -76,9 +76,8 @@ function start-kube-apiserver {
   configure-etcd-params params
 
   params+=" --secure-port=443"
-  if [[ "${ENABLE_APISERVER_INSECURE_PORT:-false}" != "true" ]]; then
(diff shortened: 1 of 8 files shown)


tallclair:
Bleh, looks like the node e2e test still depends on the insecure port: https://github.com/kubernetes/kubernetes/blob/21953d15ea48972f20a8de29d58bd5ce6d913914/test/e2e_node/services/apiserver.go#L30

I'll mark this as WIP until it's fixed.

tallclair:
/close

@knight42 is taking over this change. Thanks!

PR #92897, opened 2020-07-08
Fix description of conversion generator

### LINKED ISSUE #92895: Inaccurate declaration of what conversion generator does
**What happened**:
While reviewing work that adds a field to a custom resource defined by code rather than a CRD, I was alerted to a bug in the comment describing what the conversion generator does.  https://github.com/kubernetes/kubernetes/blob/7de3f938c82a7c0538bdfac5bb9b3cad50c30eae/staging/src/k8s.io/code-generator/cmd/conversion-gen/main.go#L34-L38 states that the Convert_... function is generated if it is not present in a non-generated file --- without regard to whether there are any problems in the autoConvert_... function.  But we see at https://github.com/kubernetes/kubernetes/blob/7de3f938c82a7c0538bdfac5bb9b3cad50c30eae/staging/src/k8s.io/code-generator/cmd/conversion-gen/generators/conversion.go#L753 that there is an additional condition that is relevant.

I expected the conversion generator to do what it is documented to do.

**Environment**:
- Kubernetes version (use `kubectl version`): This buggy doc was introduced in #71821
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

MikeSpreitzer:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

The old description failed to note that the `Convert_<pkg1>_<type>_To_<pkg2>_<type>` function is NOT generated if generation of the corresponding `autoConvert_...` function ran into
trouble with a missing or mis-typed field.  This PR revises the description of the conversion
generator to note this exception and also explain the generated registration of conversion functions.

Fixes #92895 

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/code-generator/cmd/conversion-gen/main.go
+++ b/staging/src/k8s.io/code-generator/cmd/conversion-gen/main.go
@@ -30,21 +30,33 @@ limitations under the License.
 // ones named
 //     autoConvert_<pkg1>_<type>_To_<pkg2>_<type>
 // for each such pair of types --- both with (pkg1,pkg2) =
-// (internal,external) and (pkg1,pkg2) = (external,internal).
-// Additionally: if the destination package does not contain one in a
-// non-generated file then a function named
+// (internal,external) and (pkg1,pkg2) = (external,internal).  The
+// generated conversion functions recurse on the structure of the data
+// types.  For structs, source and destination fields are matched up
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:

PR #92909, opened 2020-07-08
Add error if using kubectl run with generator other than run-pod/v1

brianpursley:
**What type of PR is this?**
/kind cleanup

If I try to create a deployment with run, like this:
```
kubectl run foo --image=nginx --generator=deployment/apps.v1
```

With kubectl 1.17, I get a warning and it successfully creates the deployment:
```
kubectl run --generator=deployment/apps.v1 is DEPRECATED and will be removed in a future version. Use kubectl run --generator=run-pod/v1 or kubectl create instead.
deployment.apps/foo created
```

But with kubectl 1.18, I get a warning and at first glance, it *appears* to succeed, but upon closer look, it shows that a pod was created instead of a deployment:
```
Flag --generator has been deprecated, has no effect and will be removed in the future.
pod/foo created
```

This PR puts a check in place to fail if you are attempting to set the generator to something other than `run-pod/v1`.

```release-note
Add validation check to prevent calling kubectl run with unsupported generator
```

```docs

```

/cc @soltysh

### DIFF
--- a/staging/src/k8s.io/kubectl/pkg/cmd/run/run.go
+++ b/staging/src/k8s.io/kubectl/pkg/cmd/run/run.go
@@ -306,6 +306,10 @@ func (o *RunOptions) Run(f cmdutil.Factory, cmd *cobra.Command, args []string) e
 		return err
 	}
 
+	if len(o.Generator) > 0 && o.Generator != generateversioned.RunPodV1GeneratorName {
+		return cmdutil.UsageErrorf(cmd, "generator %q is not supported by run", o.Generator)
+	}
+
 	generators := generateversioned.GeneratorFn("run")
 	generator, found := generators[generateversioned.RunPodV1GeneratorName]
 	if !found {


zhouya0 (approved):

staging/src/k8s.io/kubectl/pkg/cmd/run/run.go
```diff
@@ -173,13 +169,9 @@ func NewCmdRun(f cmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Co
 
 func addRunFlags(cmd *cobra.Command, opt *RunOptions) {
 	cmdutil.AddDryRunFlag(cmd)
-	cmd.Flags().StringVar(&opt.Generator, "generator", opt.Generator, i18n.T("The name of the API generator to use, see http://kubernetes.io/docs/user-guide/kubectl-conventions/#generators for a list."))
-	cmd.Flags().MarkDeprecated("generator", "has no effect and will be removed in the future.")
```
soltysh, line 177: We can't remove then, yet. They were just recently deprecated and that means we'll need to keep it around for quite a while, even if they don't do anything.

soltysh (changes requested):
If you can change this PR to fail when these flags will be used it's ok, but we can't remove them, yet.
/lgtm cancel

staging/src/k8s.io/kubectl/pkg/cmd/run/run.go
brianpursley: Ok, that was my original intention , but then I thought changing to fail would be the same as removing them basically, but I understand.  I will change them to fail of you give an "unsupported value" and update this PR.

brianpursley: OK, PR has been updated to prevent the situation where you try to use run to create a deployment (or other resource) and get a pod instead. PTAL.

PR #91851, opened 2020-06-06
kubectl: honor --cache-dir so that server resources and http-cache data are stored in the same location

SataQiu:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Honor `--cache-dir` so that server resources and http-cache data are stored in the same location:

1) If `--cache-dir` is set and not empty (such as `/foo/http-cache`),
    the discovery cache dir will be `/foo/cache/discovery`
2) If `--cache-dir` is set and empty,
    the discovery cache dir will be the default value (`~/.kube/cache/discovery`)
3) If `--cache-dir` is not set,
    the discovery cache dir will be the default value (`~/.kube/cache/discovery`)

**Which issue(s) this PR fixes**:

Fixes #91846 #78664

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs
NONE
```

PR #92910, opened 2020-07-08
Setup cache dirs properly based on --cache-dir value 

soltysh:
**What type of PR is this?**
/kind cleanup

This is a followup that builds on top of https://github.com/kubernetes/kubernetes/pull/91851 cleaning the remaining bits and setting a single parent cache dir for both http and discovery.

/assign @sallyom @SataQiu 

```release-note
--cache-dir sets cache directory for both http and discovery, defaults to $HOME/.kube/cache
```

### DIFF
--- a/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go
+++ b/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go
@@ -52,12 +52,11 @@ const (
 	flagUsername         = "username"
 	flagPassword         = "password"
 	flagTimeout          = "request-timeout"
-	flagHTTPCacheDir     = "cache-dir"
+	flagCacheDir         = "cache-dir"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go
```diff
@@ -55,7 +55,9 @@ const (
 	flagHTTPCacheDir     = "cache-dir"
```
sallyom, line 55: should this be `flagCacheDir` now? And also update below: https://github.com/kubernetes/kubernetes/pull/92910/files#diff-4e28bfb0f0c95fb4bd5a77134c7ad3f1L256

sallyom (commented):
question about changing flagHTTPCacheDir to flagCacheDir, otherwise this lgtm

deads2k:
client-go change is fine

/approve

SataQiu (commented):
/lgtm

PR #92919, opened 2020-07-08
Update cAdvisor to v0.37.0

dashpole:
**What type of PR is this?**
/kind bug

cAdvisor does a release for each kubernetes release.  This is the release for k8s 1.19.  It has a number of bug-fixes and improvements.

Because we switched to a different cAdvisor interface for serving prometheus metrics, I had to change a bunch of interfaces in the kubelet.  But the behavior is exactly the same.

```release-note
Fix detection of image filesystem, disk metrics for devicemapper, detection of OOM Kills on 5.0+ linux kernels.
```


pkg/kubelet/cadvisor/cadvisor_windows.go
```diff
@@ -58,6 +58,10 @@ func (cu *cadvisorClient) ContainerInfoV2(name string, options cadvisorapiv2.Req
 	return cu.winStatsClient.WinContainerInfos()
 }
 
+func (cu *cadvisorClient) GetRequestedContainersInfo(containerName string, options cadvisorapiv2.RequestOptions) (map[string]*cadvisorapi.ContainerInfo, error) {
+	return nil, nil
```
cheftako, line 62: Why is this also not errUnsupported?

dashpole: This PR switches from using SubcontainerInfo to GetRequestedContainersInfo.  The only difference between the two is that SubcontainerInfo takes a request, and GetRequestedContainersInfo takes options.  This is maintaining the same behavior as SubcontainerInfo, which currently returns `nil, nil` as well.

cheftako: Thanks for explaining :+1:

dashpole:
can I get 1.19 milestone as well?

dashpole:
@dims can you help with the milestone?

liggitt:
This is already in the milestone

dims:
cross my fingers 🤞 !!

dashpole:
ah, my bad

PR #92920, opened 2020-07-08
[test/kubelet]: Fix deadlock in watch manager test

gaurav1086:

**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes a deadlock condition if all the goroutines (10) will fail and try to write to the errCh and block instead of finishing the execution, while the main thread will also block through wait() for all the goroutines to finish before it tries to read from the errCh in the select operation, thus causing a deadlock. A simple illustration: https://play.golang.com/p/x0WpLjw_f1i

The solution is to create a buffered channel upto the size of the total number of the goroutines so they can all finish execution asynchronously.

Fixes #

https://play.golang.com/p/x0WpLjw_f1i

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/test/integration/kubelet/watch_manager_test.go
+++ b/test/integration/kubelet/watch_manager_test.go
@@ -40,6 +40,7 @@ func TestWatchBasedManager(t *testing.T) {
 	server := kubeapiservertesting.StartTestServerOrDie(t, nil, nil, framework.SharedEtcd())
 	defer server.TearDownFn()
 
+	const n = 10
 	server.ClientConfig.QPS = 10000
 	server.ClientConfig.Burst = 10000
 	client, err := kubernetes.NewForConfig(server.ClientConfig)
@@ -65,8 +66,8 @@ func TestWatchBasedManager(t *testing.T) {
 	// create 1000 secrets in parallel
 	t.Log(time.Now(), "creating 1000 secrets")
 	wg := sync.WaitGroup{}
-	errCh := make(chan error, 1)
-	for i := 0; i < 10; i++ {
+	errCh := make(chan error, n)
+	for i := 0; i < n; i++ {
 		wg.Add(1)
 		go func(i int) {
 			defer wg.Done()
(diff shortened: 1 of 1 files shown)


neolit123 (commented):
/lgtm
/sig node

neolit123:
might be worth to include this PR in the 1.19 release?
(`/milestone v1.19`)

test/integration/kubelet/watch_manager_test.go
```diff
@@ -65,7 +65,7 @@ func TestWatchBasedManager(t *testing.T) {
 	// create 1000 secrets in parallel
 	t.Log(time.Now(), "creating 1000 secrets")
 	wg := sync.WaitGroup{}
-	errCh := make(chan error, 1)
+	errCh := make(chan error, 10)
```
SergeyKanzhelev, line 68: nit: worth making it a constant as it is used in 4 places that needs to be coordinated.

gaurav1086: Thanks. Added.

neolit123 (commented):
/lgtm
/approve
/priority important-longterm

neolit123:
/milestone v1.19
this seems appropriate to include in 1.19?

liggitt:
this is a good catch, but only really causes problems when the test would have already failed

moving to 1.20 since we're approaching code thaw and any 1.19 fixes will need cherry picks shortly

/milestone v1.20

liggitt:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92920/pull-kubernetes-node-e2e/1285251532081598464 flake is #93210

justaugustus:
//test pull-kubernetes-e2e-kind-ipv6

justaugustus:
Ooop.

PR #92926, opened 2020-07-09
[client-go] Fix argument ordering for fake scale client update

detiber:
**What type of PR is this?**

/kind bug

Fixes the argument order used when calling testing.NewUpdateSubresourceAction
within the fake scale client. This was causing the generated action to swap the values
of the Namespace and Subresource in the Action.

For more information on release notes see: https://git.k8s.io/community/contributors/guide/release-notes.md
-->
```release-note
NONE
```

### DIFF
--- a/staging/src/k8s.io/client-go/scale/fake/client.go
+++ b/staging/src/k8s.io/client-go/scale/fake/client.go
@@ -60,7 +60,7 @@ func (f *fakeNamespacedScaleClient) Get(ctx context.Context, resource schema.Gro
 
 func (f *fakeNamespacedScaleClient) Update(ctx context.Context, resource schema.GroupResource, scale *autoscalingapi.Scale, opts metav1.UpdateOptions) (*autoscalingapi.Scale, error) {
 	obj, err := f.fake.
-		Invokes(testing.NewUpdateSubresourceAction(resource.WithVersion(""), f.namespace, "scale", scale), &autoscalingapi.Scale{})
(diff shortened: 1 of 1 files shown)


staging/src/k8s.io/client-go/scale/fake/client.go
```diff
@@ -60,7 +60,7 @@ func (f *fakeNamespacedScaleClient) Get(ctx context.Context, resource schema.Gro
 
 func (f *fakeNamespacedScaleClient) Update(ctx context.Context, resource schema.GroupResource, scale *autoscalingapi.Scale, opts metav1.UpdateOptions) (*autoscalingapi.Scale, error) {
 	obj, err := f.fake.
-		Invokes(testing.NewUpdateSubresourceAction(resource.WithVersion(""), f.namespace, "scale", scale), &autoscalingapi.Scale{})
+		Invokes(testing.NewUpdateSubresourceAction(resource.WithVersion(""), "scale", f.namespace, scale), &autoscalingapi.Scale{})
```
detiber, line 63: I went to build some tests using this fake client and was scratching my head as to why the Reactor I was building wasn't behaving as I expected. It turns out that this was previously swapping the namespace and subresource parameters. 

Oddly enough, it doesn't appear to cause any major functionality issues, but if you are trying to do any validation or logic depending on either the Subresource or Namespace fields of the Action you are in for a surprise because they are swapped.

detiber:
/test pull-kubernetes-e2e-gce
I'm assuming flake, because if the e2e test is using the fake scale client we have bigger issues :joy:

Issue #92931, opened 2020-07-09
If volume expansion fails on the node with volume-in-use error then kubelet should not retry nodeExpansion


gnufied:
If CSI volume expansion fails on the node with volume-in-use error then kubelet not should retry node expansion.

/sig storage

PR #92932, opened 2020-07-09
Ensure that for RWX volumes expansion is called on all nodes

### LINKED ISSUE #80918: CSI resize doesn't handle multi-attached volumes quite right
if a CSI volume is attached to more than 1 node, and it is expanded, and if the CSI plugin returns true from ControllerResizeVolume, then the PVC will get marked for node resize, but there's no mechanism to ensure that every node currently using the volume will call the CSI node plugin.

**What happened**: One node called the NodeExpandVolume RPC, the rest did not.

**What you expected to happen**: All nodes should call the NodeExpandVolume RPC.

**How to reproduce it (as minimally and precisely as possible)**: Use a CSI plugin that supports mutli-attach, create a PVC using that CSI plugin, create 2 or more pods using the same PVC on different nodes, then resize the PVC.

**Anything else we need to know?**: There are 2 necessary activities getting combined into a single operation: 1. node-side work required to complete the resize operation, such as resizing of a filesystem and 2. node-side work required to make pods aware of the updated volume size. (1) needs to be done only a single time but needs to be done before the resize can be considered complete. It needs to be done even if the volume is not in use at the time of the resize. (2) needs to be done for every node currently accessing the volume so they all find out about the new size, but if no nodes are accessing the volume, then it could be skipped without harm.

Consider the case of an iSCSI-based plugin implementing raw block volumes. iSCSI supports r/w multi attach as long as it's raw blocks and not filesystem mode. To perform a resize of such a volume, the underlying LUN needs to be resized (on the storage controller) and then the iSCSI initiators all have to rescan their buses to see the new size. In CSI, the iSCSI rescan operation needs to happen in NodeExpandVolume, because this is the only place where such a thing is possible. This is a case of using NodeExpandVolume for purpose (2) above, though, and kubelet only supports purpose (1) currently.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.15.1
- Cloud provider or hardware configuration: n/a
- OS (e.g: `cat /etc/os-release`): Arch Linux
- Kernel (e.g. `uname -a`): 5.2.5-arch1-1-ARCH
- Install tools: n/a
- Network plugin and version (if this is a network-related bug): bridge
- Others: containerd

gnufied:
Also if CSI driver throws volume-in-use error do not retry node expansion.

This PR adds desired size based on PV capacity to desired_state_of_the_world and actual size based on `pvc.Status.Capacity` to actual_state_of_the_world. Actual size is calculated the time when pod enters ASOW. This way if a RWX volume that is node-expanded on one node will not have actual PVC size updated on second node until second node also calls `NodeExpandVolume`. 

Fixes https://github.com/kubernetes/kubernetes/issues/80918 and https://github.com/kubernetes/kubernetes/issues/92931

/kind bug
/sig storage

```release-note
Do not retry NodeExpandVolume if CSI drivers throw volume-in-use error
Call NodeExpandVolume on all RWX nodes
```

### DIFF
--- a/pkg/kubelet/volumemanager/cache/actual_state_of_world.go
+++ b/pkg/kubelet/volumemanager/cache/actual_state_of_world.go
@@ -25,6 +25,7 @@ import (
 	"sync"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/klog/v2"
@@ -76,6 +77,9 @@ type ActualStateOfWorld interface {
 	// attached volumes, an error is returned.
 	SetDeviceMountState(volumeName v1.UniqueVolumeName, deviceMountState operationexecutor.DeviceMountState, devicePath, deviceMountPath string) error
 
+	// SetVolumeClaimSize sets pvc claim size by reading pvc.Status.Capacity
+	SetVolumeClaimSize(volumeName v1.UniqueVolumeName, claimSize resource.Quantity)
+
 	// DeletePodFromVolume removes the given pod from the given volume in the
 	// cache indicating the volume has been successfully unmounted from the pod.
 	// If a pod with the same unique name does not exist under the specified
@@ -108,7 +112,7 @@ type ActualStateOfWorld interface {
 	// volumes, depend on this to update the contents of the volume.
 	// All volume mounting calls should be idempotent so a second mount call for
 	// volumes that do not need to update contents should not fail.
-	PodExistsInVolume(podName volumetypes.UniquePodName, volumeName v1.UniqueVolumeName) (bool, string, error)
+	PodExistsInVolume(vmt VolumeToMount) (bool, string, error)
 
 	// VolumeExistsWithSpecName returns true if the given volume specified with the
 	// volume spec name (a.k.a., InnerVolumeSpecName) exists in the list of
@@ -148,11 +152,6 @@ type ActualStateOfWorld interface {
 	// have no mountedPods. This list can be used to determine which volumes are
 	// no longer referenced and may be globally unmounted and detached.
 	GetUnmountedVolumes() []AttachedVolume
-
-	// MarkFSResizeRequired marks each volume that is successfully attached and
-	// mounted for the specified pod as requiring file system resize (if the plugin for the
-	// volume indicates it requires file system resize).
-	MarkFSResizeRequired(volumeName v1.UniqueVolumeName, podName volumetypes.UniquePodName)
 
 	// GetAttachedVolumes returns a list of volumes that is known to be attached
 	// to the node. This list can be used to determine volumes that are either in-use
@@ -265,6 +264,17 @@ type attachedVolume struct {
 	// deviceMountPath contains the path on the node where the device should
 	// be mounted after it is attached.
 	deviceMountPath string
+
+	// claimStatusSize indicates actual size of the persistent volume claim.
+	// This is calculated from pvc.Status.Capacity. claimStatusSize is recorded
+	// when a pod enters ASOW via VerifyControllerAttachedVolume and before volume is mounted.
+	// In case other nodes expand the volume while this node is mounting the volume, claimStatusSize
+	// still refers to old pvc status size that was recorded when pod entered ASOW.
+	claimStatusSize resource.Quantity
+
+	// volumeInUseErrorForExpansion indicates volume driver has previously returned volume-in-use error
+	// for this volume and volume expansion on this node should not be retried
+	volumeInUseErrorForExpansion bool
 }
 
 // The mountedPod object represents a pod for which the kubelet volume manager
@@ -306,10 +316,6 @@ type mountedPod struct {
 	// volumeGidValue contains the value of the GID annotation, if present.
 	volumeGidValue string
 
-	// fsResizeRequired indicates the underlying volume has been successfully
-	// mounted to this pod but its size has been expanded after that.
-	fsResizeRequired bool
-
 	// volumeMountStateForPod stores state of volume mount for the pod. if it is:
 	//   - VolumeMounted: means volume for pod has been successfully mounted
 	//   - VolumeMountUncertain: means volume for pod may not be mounted, but it must be unmounted
@@ -318,7 +324,15 @@ type mountedPod struct {
 
 func (asw *actualStateOfWorld) MarkVolumeAsAttached(
 	volumeName v1.UniqueVolumeName, volumeSpec *volume.Spec, _ types.NodeName, devicePath string) error {
-	return asw.addVolume(volumeName, volumeSpec, devicePath)
+	claimSize := resource.Quantity{Format: resource.BinarySI}
+	return asw.addVolume(volumeName, volumeSpec, devicePath, claimSize)
+}
+
+// MarkAsAttachedWithSize is same as MarkVolumeAsAttached except we also record pvc size in actualStateOfWorld
+// This is the function that should be used in volumemanager for marking a volume as attached
+func (asw *actualStateOfWorld) MarkAsAttachedWithSize(
+	volumeName v1.UniqueVolumeName, volumeSpec *volume.Spec, _ types.NodeName, devicePath string, claimSize resource.Quantity) error {
+	return asw.addVolume(volumeName, volumeSpec, devicePath, claimSize)
 }
 
 func (asw *actualStateOfWorld) MarkVolumeAsUncertain(
@@ -397,6 +411,22 @@ func (asw *actualStateOfWorld) GetVolumeMountState(volumeName v1.UniqueVolumeNam
 	return podObj.volumeMountStateForPod
 }
 
+// SetVolumeClaimSize sets actual pvc claim size in actualStateOfWorld for volume expansion.
+// Usually claimStatusSize is set when PVC enters actualStateOfWorld of the world but in case of kubelet
+// restart we might need to set claimStatusSize after the fact.
+func (asw *actualStateOfWorld) SetVolumeClaimSize(volumeName v1.UniqueVolumeName, claimSize resource.Quantity) {
+	asw.Lock()
+	defer asw.Unlock()
+
+	volumeObj, ok := asw.attachedVolumes[volumeName]
+	// only set volume claim size if claimStatusSize is zero
+	// this can happen when volume was rebuilt after kubelet startup
+	if ok && volumeObj.claimStatusSize.IsZero() {
+		volumeObj.claimStatusSize = claimSize
+		asw.attachedVolumes[volumeName] = volumeObj
+	}
+}
+
 // addVolume adds the given volume to the cache indicating the specified
 // volume is attached to this node. If no volume name is supplied, a unique
 // volume name is generated from the volumeSpec and returned on success. If a
@@ -404,7 +434,7 @@ func (asw *actualStateOfWorld) GetVolumeMountState(volumeName v1.UniqueVolumeNam
 // volume plugin can support the given volumeSpec or more than one plugin can
 // support it, an error is returned.
 func (asw *actualStateOfWorld) addVolume(
-	volumeName v1.UniqueVolumeName, volumeSpec *volume.Spec, devicePath string) error {
+	volumeName v1.UniqueVolumeName, volumeSpec *volume.Spec, devicePath string, claimSize resource.Quantity) error {
 	asw.Lock()
 	defer asw.Unlock()
 
@@ -442,6 +472,7 @@ func (asw *actualStateOfWorld) addVolume(
 			pluginIsAttachable: pluginIsAttachable,
 			deviceMountState:   operationexecutor.DeviceNotMounted,
 			devicePath:         devicePath,
+			claimStatusSize:    claimSize,
 		}
 	} else {
 		// If volume object already exists, update the fields such as device path
@@ -495,34 +526,6 @@ func (asw *actualStateOfWorld) AddPodToVolume(markVolumeOpts operationexecutor.M
 	return nil
 }
 
-func (asw *actualStateOfWorld) MarkVolumeAsResized(
-	podName volumetypes.UniquePodName,
-	volumeName v1.UniqueVolumeName) error {
-	asw.Lock()
-	defer asw.Unlock()
-
-	volumeObj, volumeExists := asw.attachedVolumes[volumeName]
-	if !volumeExists {
-		return fmt.Errorf(
-			"no volume with the name %q exists in the list of attached volumes",
-			volumeName)
-	}
-
-	podObj, podExists := volumeObj.mountedPods[podName]
-	if !podExists {
-		return fmt.Errorf(
-			"no pod with the name %q exists in the mounted pods list of volume %s",
-			podName,
-			volumeName)
-	}
-
-	klog.V(5).Infof("Volume %s(OuterVolumeSpecName %s) of pod %s has been resized",
-		volumeName, podObj.outerVolumeSpecName, podName)
-	podObj.fsResizeRequired = false
-	asw.attachedVolumes[volumeName].mountedPods[podName] = podObj
-	return nil
-}
-
 func (asw *actualStateOfWorld) MarkRemountRequired(
 	podName volumetypes.UniquePodName) {
 	asw.Lock()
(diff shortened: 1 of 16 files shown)


pkg/kubelet/volumemanager/cache/actual_state_of_world.go
```diff
@@ -397,14 +400,30 @@ func (asw *actualStateOfWorld) GetVolumeMountState(volumeName v1.UniqueVolumeNam
...
+// Usually claimStatusSize is set when PVC enters actualStateOfWorld of the world but in case of kubelet
+// restart we might need to set claimStatusSize after the fact.
+func (asw *actualStateOfWorld) SetVolumeClaimSize(volumeName v1.UniqueVolumeName, claimSize resource.Quantity) {
+	asw.Lock()
+	defer asw.Unlock()
+
+	volumeObj, ok := asw.attachedVolumes[volumeName]
+	// only set volume claim size if claimStatusSize is zero
+	// this can happen when volume was rebuilt after kubelet startup
+	if ok && volumeObj.claimStatusSize.IsZero() {
```
msau42, line 413: What happens if you expand a volume twice?

msau42 (commented):
Can you explain how this change will call expand on all nodes?  What if by the time a Pod enters dsw one node2, node1 has already updated pvc status?

pkg/kubelet/volumemanager/cache/actual_state_of_world.go
gnufied: if you expand a volume twice then only size recorded in DSOW changes until `NodeExpandVolume` is called and at which point size recorded in ASOW s updated as well.

gnufied:
@msau42 updated the description with how it fixes RWX expansion.  This PR only fixes online expansion for RWX volumes if volume was offline on one node and online on another, then `NodeExpandVolume` will be called only on one node. I think that is reasonable behaviour.

pkg/kubelet/volumemanager/populator/desired_state_of_world_populator.go
```diff
@@ -412,10 +403,12 @@ func (dswp *desiredStateOfWorldPopulator) checkVolumeFSResize(
...
-	if volumeRequiresFSResize(pvc, volumeSpec.PersistentVolume) {
-		dswp.actualStateOfWorld.MarkFSResizeRequired(uniqueVolumeName, uniquePodName)
-	}
-	processedVolumesForFSResize.Insert(string(uniqueVolumeName))
+	pvCap := volumeSpec.PersistentVolume.Spec.Capacity.Storage()
+	pvcStatusCap := pvc.Status.Capacity.Storage()
+	dswp.desiredStateOfWorld.UpdatePersistentVolumeSize(uniqueVolumeName, *pvCap)
+
+	// in case the actualStateOfWorld was rebuild after kubelet restart ensure that claimSize is set to accurate value
+	dswp.actualStateOfWorld.SetVolumeClaimSize(uniqueVolumeName, *pvcStatusCap)
```
msau42, line 411: this seems odd. We update asw in dswp? This is because reconstruction doesn't know/have the pvc status? CSI at least keeps a file, so we could use that?

pkg/volume/util/operationexecutor/operation_generator.go
```diff
@@ -1667,6 +1663,11 @@ func (og *operationGenerator) nodeExpandVolume(volumeToMount VolumeToMount, rsOp
 				// On retry, NodeExpandVolume will be called again but do nothing
 				return false, fmt.Errorf("MountVolume.NodeExpandVolume update PVC status failed : %v", err)
 			}
+			ok := actualStateOfWorld.MarkVolumeAsResized(volumeToMount.VolumeName, pvSpecCap)
```
msau42, line 1666: In the readonly case above, or any other case where we return true early, do we need to mark volume as resized? https://github.com/kubernetes/kubernetes/pull/92932/files#diff-450e811a4953f760ff1594ede8b2037eR1637

pkg/kubelet/volumemanager/cache/actual_state_of_world.go
```diff
@@ -265,6 +264,14 @@ type attachedVolume struct {
 	// deviceMountPath contains the path on the node where the device should
 	// be mounted after it is attached.
 	deviceMountPath string
+
+	// claimStatusSize indicates actual size of the persistent volume claim.
+	// This is mainly calculated from pvc.Status.Capacity
```
msau42, line 269: Can you add some more comments here explaining how this is populated/updated, how it addresses the RWX case, and the known races?

pkg/kubelet/volumemanager/cache/actual_state_of_world.go
msau42: Not related to this change, but it may be worth thinking of a scenario where someone increases the size twice in quick succession. Would it be possible that we resize the fs for the first expansion, but when we update the pvc status, it updates the size to the 2nd value? (maybe an e2e test could be valuable too)

pkg/volume/util/operationexecutor/operation_generator.go
gnufied: we don't need to mark them as resized as such because such volumes are not even entered into DSOW's `persistentVolumeSize` field. 

for readonly volumes, we don't track their sizes in DSOW - https://github.com/kubernetes/kubernetes/pull/92932/files#diff-94e527d26427ceb4f990f05f76952656 and hence will not result in resize. I have also added an unit test that covers readonly volumes

pkg/kubelet/volumemanager/cache/actual_state_of_world.go
gnufied: done.

pkg/kubelet/volumemanager/populator/desired_state_of_world_populator.go
gnufied: Yes, that is indeed the case. I would like to this more about recording pvc's status size in json file. Can I do this in a follow up?

pkg/kubelet/volumemanager/cache/actual_state_of_world.go
gnufied: yes If someone increases the size in quick succession and assuming control-plane is finished doing its thing, then size of PV will be second value. But in this case when `NodeExpandVolume` is invoked, it will get 2nd value recorded in PV as newSize (and not the intermediate one) and hence `pvc.Status.Capacity` will be updated with 2nd value.

pkg/volume/util/operationexecutor/operation_generator.go
msau42: Would we still need to update the pvc status to say resizing is finished?

pkg/kubelet/volumemanager/populator/desired_state_of_world_populator.go
msau42: sure can you open an issue to track

pkg/kubelet/volumemanager/cache/actual_state_of_world.go
msau42: Do we need to think about a way we can address this? Such as recording the size before we call node expansion and using that value to update? This is kind of related to the challenge that the driver doesn't actually tell us the size.

pkg/volume/util/operationexecutor/operation_generator.go
gnufied: we can't mark PVC resizing as finished because if PVC was readonly on this node, resizing will never happen whereas it can successfully happen on some other node.

gnufied: Having said that - all the `return true, blah` statements look weird but we are basically double checking scenarios we already checking when a volume enters DSOW for volume expansion. May be that can be cleaned up but it does not hurt to be safe.

pkg/kubelet/volumemanager/populator/desired_state_of_world_populator.go
gnufied: https://github.com/kubernetes/kubernetes/issues/92947

pkg/controller/volume/attachdetach/attach_detach_controller.go
```diff
@@ -376,7 +377,7 @@ func (adc *attachDetachController) populateActualStateOfWorld() error {
 			// volume spec is not needed to detach a volume. If the volume is used by a pod, it
 			// its spec can be: this would happen during in the populateDesiredStateOfWorld which
 			// scans the pods and updates their volumes in the ActualStateOfWorld too.
-			err = adc.actualStateOfWorld.MarkVolumeAsAttached(uniqueName, nil /* VolumeSpec */, nodeName, attachedVolume.DevicePath)
+			err = adc.actualStateOfWorld.MarkVolumeAsAttached(uniqueName, nil /* VolumeSpec */, nodeName, attachedVolume.DevicePath, resource.Quantity{Format: resource.BinarySI})
```
saad-ali, line 380: Something seems wrong here since you are having to pass new params to A/D controller but don't use those new params in A/D controller.

gnufied: yeah I thought about it, it is somewhat ugly. we don't track `pvc.status.capacity` in control-plane for volume expansion in attach-detach controller but we do in volumemanager and hence this parameter is unused in A/D controller.

I could introduce a new interface function but it has to be part of `ActualXXUpdater` interface and A/D controller side of `ASOW` has to provide a dummy implementation.  I am happy to do that instead of this.

msau42:
/hold
to address pending comment

saad-ali:
Taking a deeper look

pkg/controller/volume/attachdetach/attach_detach_controller.go
gnufied: I have a commit that introduces the new interface function called `MarkAsAttachedWithSize` inside `ActualStateOfWorldAttacherUpdater` interface.  I have not pushed it because I am afraid to make CI unhappy at last minute - https://github.com/gnufied/kubernetes/commit/983e410660fcec7bb0fb019f9d68eb4d7eaf333a#diff-24ba5bb87df94b8b68b6fbd13ef972b0R262 

But I am happy to do this in a follow up. The change is mostly cosmetic and does not changes underlying logic of the PR.

gnufied: done

pkg/kubelet/volumemanager/cache/actual_state_of_world.go
```diff
@@ -108,7 +112,7 @@ type ActualStateOfWorld interface {
 	// volumes, depend on this to update the contents of the volume.
 	// All volume mounting calls should be idempotent so a second mount call for
 	// volumes that do not need to update contents should not fail.
-	PodExistsInVolume(podName volumetypes.UniquePodName, volumeName v1.UniqueVolumeName) (bool, string, error)
+	PodExistsInVolume(vmt VolumeToMount) (bool, string, error)
```
saad-ali, line 115: Don't collapse - scope the input for each method down to the bare minimum it needs.

pkg/kubelet/volumemanager/cache/actual_state_of_world.go
```diff
@@ -265,6 +264,17 @@ type attachedVolume struct {
...
 	deviceMountPath string
+
+	// claimStatusSize indicates actual size of the persistent volume claim.
+	// This is calculated from pvc.Status.Capacity. claimStatusSize is recorded
+	// when a pod enters ASOW via VerifyControllerAttachedVolume and before volume is mounted.
+	// In case other nodes expand the volume while this node is mounting the volume, claimStatusSize
+	// still refers to old pvc status size that was recorded when pod entered ASOW.
+	claimStatusSize resource.Quantity
+
+	// volumeInUseErrorForExpansion indicates CSI driver has previously returned volume-in-use error
```
saad-ali, line 275: This code is not CSI specific

pkg/volume/util/operationexecutor/operation_executor.go
```diff
@@ -216,6 +220,9 @@ type ActualStateOfWorldAttacherUpdater interface {
 	// volumes.  See issue 29695.
 	MarkVolumeAsAttached(volumeName v1.UniqueVolumeName, volumeSpec *volume.Spec, nodeName types.NodeName, devicePath string) error
 
+	// MarkAsAttachedWithSize is same as MarkVolumeAsAttached except we also record pvc size in actualStateOfWorld
+	MarkAsAttachedWithSize(volumeName v1.UniqueVolumeName, volumeSpec *volume.Spec, nodeName types.NodeName, devicePath string, claimSize resource.Quantity) error
```
saad-ali, line 224: Move this to `type ActualStateOfWorld interface` in `kubernetes/kubernetes/pkg/kubelet/volumemanager/cache/actual_state_of_world.go`. The methods in `ActualStateOfWorldAttacherUpdater` are common for both A/D controller and kubelet volume manager. Since this only applies to the later we don't need to put it in the common interface.

pkg/volume/util/operationexecutor/operation_executor.go
```diff
@@ -195,13 +195,17 @@ type ActualStateOfWorldMounterUpdater interface {
...
 
 	// GetDeviceMountState returns mount state of the device in global path
 	GetDeviceMountState(volumeName v1.UniqueVolumeName) DeviceMountState
 
 	// GetVolumeMountState returns mount state of the volume for the Pod
 	GetVolumeMountState(volumName v1.UniqueVolumeName, podName volumetypes.UniquePodName) VolumeMountState
+
+	// MarkForInUseExpansionError marks the volume to have in-use error during expansion.
+	// volume expansion must not be retried for this volume
+	MarkForInUseExpansionError(volumeName v1.UniqueVolumeName)
```
saad-ali, line 208: Move this to `type ActualStateOfWorld interface` in `kubernetes/kubernetes/pkg/kubelet/volumemanager/cache/actual_state_of_world.go`. The methods in `ActualStateOfWorldAttacherUpdater` are common for both A/D controller and kubelet volume manager. Since this only applies to the later we don't need to put it in the common interface.

saad-ali (changes requested):

pkg/volume/util/operationexecutor/operation_executor.go
gnufied: We can't move this function into `ActualStateOfWorld` interface of `volumemanager/cache` because - it will cause cyclic dependency.

pkg/volume/util/operationexecutor/operation_executor.go
gnufied: same as above.

pkg/volume/util/operationexecutor/operation_executor.go
gnufied: see - cache.ActualStateWorld already depends on `operationexecutor` package, now if we say operationexecutor will depend on `cache.ActualStateWorld` it will cause cyclic dependency.

pkg/kubelet/volumemanager/cache/actual_state_of_world.go
gnufied: In general I agree but if we keep each argument separate then this function will now need 4 arguments and IMO i would be happy if it were 3 but 4 is tad too much. 

I know in other places we have used https://refactoring.guru/introduce-parameter-object pattern when number of parameters to a function grow too big, for example `MarkVolumeAsMounted` will require 9 arguments if specified separately.

pkg/kubelet/volumemanager/cache/actual_state_of_world.go
gnufied: fixed

nikhita:
👋 Hi, just checking if this is still planned for 1.19 and if this needs to go in before code freeze is lifted? Can this be cherry-picked to the `release-1.19` branch after code freeze is lifted?

gnufied:
@nikhita this PR can be removed from 1.19 milestone. The important bug we wanted to fix has been merged via https://github.com/kubernetes/kubernetes/pull/92986 . I need to rework this PR little bit to what was discussed with @saad-ali offline.

nikhita:
@gnufied Thanks for the update! I've updated the milestone to v1.20. :+1:

sayanchowdhury:
👋🏽 Hello from Bug Triage team! The code freeze is starting November 12th (about 3 weeks from now) and while there is still plenty of time, we want to ensure that each PR has a chance to be merged on time.

droslean:
@gnufied Can you check this? Are you expecting that this PR will land in 1.20?

droslean:
Since there is no activity on this PR for many days, and the code freeze has started, I am moving this to 1.21

/milestone v1.21

shubheksha:
@gnufied do you plan to get this merged?

erismaster:
Hi 👋 Bug Triage here, just a reminder that code freeze is end of day, tomorrow the 9th! PRs need to be merged in before then, or they will require an approved exception to go into the 1.21 milestone.

Also, this PR looks hasn't had much activity lately, as such if we don't hear any updates I will be dropping the milestone until it is active again. Thanks!

erismaster:
This PR has not had much activity for a few releases now, I am going to clear the milestone. If this becomes active again, it can be added into the relevant milestone at that time. Thanks!

/milestone clear

Issue #67730, opened 2018-08-22
Unable use localhost:<nodeport> with IPVS


lbernail:
/kind bug
/sig network
/area ipvs

**What happened**:
Connecting to a service using a nodeport and the loopback address does not work

**What you expected to happen**:
Being able to connect to a nodeport using the loopback (it works with IPVS)

**How to reproduce it (as minimally and precisely as possible)**:
Create a nodeport service and try to connect to it using localhost:<nodeport>

**Anything else we need to know?**:
The IPVS configuration looked fine:
```
sudo ipvsadm -Ln -t 127.0.0.1:32116 --stats
Prot LocalAddress:Port               Conns   InPkts  OutPkts  InBytes OutBytes
  -> RemoteAddress:Port
TCP  127.0.0.1:32116                     1        2        0      120        0
  -> 10.x.0.99:5000                     0        0        0        0        0
  -> 10.x.0.115:5000                    1        2        0      120        0
  -> 10.x.0.130:5000                    0        0        0        0        0
```
The masquerading mark is applied properly
```
Chain KUBE-MARK-MASQ (3 references)
    pkts      bytes target     prot opt in     out     source               destination
       2      120 MARK       all  --  *      *       0.0.0.0/0            0.0.0.0/0            MARK or 0x4000

Chain KUBE-NODE-PORT (1 references)
    pkts      bytes target     prot opt in     out     source               destination
       2      120 KUBE-MARK-MASQ  all  --  *      *       0.0.0.0/0            0.0.0.0/0

Chain KUBE-SERVICES (2 references)
    pkts      bytes target     prot opt in     out     source               destination
       2      120 KUBE-NODE-PORT  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* Kubernetes nodeport TCP port for masquerade purpose */ match-set KUBE-NODE-PORT-TCP dst
```
However the MASQUERADE rule in the KUBE-POSTROUTING is not reached
```
Chain KUBE-POSTROUTING (1 references)
    pkts      bytes target     prot opt in     out     source               destination
       0        0 MASQUERADE  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* kubernetes service traffic requiring SNAT */ mark match 0x4000/0x4000
```

tcpdump does not show any traffic targeting port 5000. It looks like the kernel is dropping connections with source 127.0.0.1 and an external destination before reaching the POSTROUTING chain.

Changing the local route table fixed the issue:
```
[...]

Issue #92708, opened 2020-07-01
[failing test] "Flexvolumes should be mountable when non-attachable" fails when using ubuntu node image


aojea:
The test

[sig-storage] Flexvolumes should be mountable when non-attachable

fail using kube-proxy in IPVS mode

xref: https://github.com/kubernetes/kubernetes/pull/92584#issuecomment-652521262

https://testgrid.k8s.io/sig-network-gce#pull-gci-gce-ipvs

athenabot:
/triage unresolved

Comment `/remove-triage unresolved` when the issue is assessed and confirmed.

🤖 I am a bot run by vllry. 👩‍🔬

andrewsykim:
Still digging into this but no good progress yet. The only difference in the IPVS job vs the standard `pull-kubernetes-e2e-gce ` is the kube-proxy mode and the GCP node image. The former is using `ubuntu` and the latter is using `gci`. 

@msau42 do you know if the difference in node image could possibly lead to this test failure?

andrewsykim:
The failure is the flex client pod failing to succesfully start up and timing out:
```
ï¿½[1mSTEPï¿½[0m: starting flex-client
ï¿½[1mSTEPï¿½[0m: Deleting pod flex-client in namespace flexvolume-2882
Jul 16 12:34:33.805: INFO: Waiting for pod flex-client to disappear
Jul 16 12:34:33.856: INFO: Pod flex-client still exists
Jul 16 12:34:35.856: INFO: Waiting for pod flex-client to disappear
Jul 16 12:34:35.900: INFO: Pod flex-client still exists
Jul 16 12:34:37.856: INFO: Waiting for pod flex-client to disappear
Jul 16 12:34:37.898: INFO: Pod flex-client still exists
Jul 16 12:34:39.856: INFO: Waiting for pod flex-client to disappear
Jul 16 12:34:39.894: INFO: Pod flex-client still exists
Jul 16 12:34:41.856: INFO: Waiting for pod flex-client to disappear
Jul 16 12:34:41.896: INFO: Pod flex-client no longer exists
Jul 16 12:34:41.896: FAIL: Failed to create client pod: timed out waiting for the condition

Full Stack Trace
k8s.io/kubernetes/test/e2e/framework/volume.testVolumeClient(0xc000d1a160, 0xc001a28060, 0xf, 0x4ae1fce, 0x4, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
[...]

PR #91666, opened 2020-06-02
pkg/kubelet: add KUBE-FIREWALL-IN chain, move localnet rule to it

squeed:
@champtar observed that the rule added in #91569 still allowed packets through unidirectionally, meaning UDP with a spoofed source wouldn't be caught. So we need to filter on source interface.

However, KUBE-FIREWALL is run on both OUTPUT and INPUT, so it can't reference interfaces. This PR, thus,
- creates a chain KUBE-FIREWALL-IN
- Add it to INPUT, pass through to KUBE-FIREWALL
- Move the localnet protection rule to KUBE-FIREWALL-IN, and adjust it
  to reference source if, not source address.

Signed-off-by: Casey Callendrello <cdc@redhat.com>

**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #90259

**Special notes for your reviewer**:
You can test this pretty simply using `cnitool` if you don't want to spin up a whole kind cluster.

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

/sig network

PR #91569, opened 2020-05-29
kubelet: block non-forwarded packets from crossing the localhost boundary

squeed:
We set route_localnet so that host-network processes can connect to `127.0.0.1:NodePort` and it still works. This, however, is too permissive.

So, block martians that are not already in conntrack.

**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
As outlined in #90259, blindly setting `route_localnet` is too permissive. We need to ensure that only packets we want can pass the martian boundary.

**Which issue(s) this PR fixes**:
Fixes #90259

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

PR #92938, opened 2020-07-09
Do not set sysctlRouteLocalnet (CVE-2020-8558)

### LINKED ISSUE #92315: CVE-2020-8558: Node setting allows for neighboring hosts to bypass localhost boundary
CVSS Rating:

In typical clusters: medium (5.4) [CVSS:3.1/AV:A/AC:L/PR:N/UI:N/S:U/C:L/I:L/A:N](https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:A/AC:L/PR:N/UI:N/S:U/C:L/I:L/A:N)

In clusters where API server insecure port has not been disabled: high (8.8) [CVSS:3.1/AV:A/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H](https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:A/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H)

A security issue was discovered in `kube-proxy` which allows adjacent hosts to reach TCP and UDP services bound to `127.0.0.1` running on the node or in the node's network namespace. For example, if a cluster administrator runs a TCP service on a node that listens on `127.0.0.1:1234`, because of this bug, that service would be potentially reachable by other hosts on the same LAN as the node, or by containers running on the same node as the service. If the example service on port `1234` required no additional authentication (because it assumed that only other localhost processes could reach it), then it could be vulnerable to attacks that make use of this bug.

The Kubernetes API Server's default insecure port setting causes the API server to listen on `127.0.0.1:8080` where it will accept requests without authentication. Many Kubernetes installers explicitly disable the API Server's insecure port, but in clusters where it is not disabled, an attacker with access to another system on the same LAN or with control of a container running on the master may be able to reach the API server and execute arbitrary API requests on the cluster. This port is deprecated, and will be removed in Kubernetes v1.20.

### Am I vulnerable?

You may be vulnerable if:

- You are running a vulnerable version (see below)
- Your cluster nodes run in an environment where untrusted hosts share the same layer 2 domain (i.e. same LAN) as nodes
- Your cluster allows untrusted pods to run containers with `CAP_NET_RAW` (the Kubernetes default is to allow this capability).
- Your nodes (or hostnetwork pods) run any localhost-only services which do not require any further authentication. To list services that are potentially affected, run the following commands on nodes:
    - `lsof +c 15 -P -n -i4TCP@127.0.0.1 -sTCP:LISTEN`
    - `lsof +c 15 -P -n -i4UDP@127.0.0.1`

    On a master node, an lsof entry like this indicates that the API server may be listening with an insecure port:

```
COMMAND        PID  USER FD   TYPE DEVICE SIZE/OFF NODE NAME
kube-apiserver 123  root  7u  IPv4  26799      0t0  TCP 127.0.0.1:8080 (LISTEN)
```

#### Affected Versions
- kubelet/kube-proxy v1.18.0-1.18.3
- kubelet/kube-proxy v1.17.0-1.17.6
- kubelet/kube-proxy <=1.16.10

### How do I mitigate this vulnerability?
Prior to upgrading, this vulnerability can be mitigated by manually adding an iptables rule on nodes. This rule will reject traffic to 127.0.0.1 which does not originate on the node.

` iptables -I INPUT --dst 127.0.0.0/8 ! --src 127.0.0.0/8 -m conntrack ! --ctstate RELATED,ESTABLISHED,DNAT -j DROP`

Additionally, if your cluster does not already have the API Server insecure port disabled, we strongly suggest that you disable it. Add the following flag to your kubernetes API server command line: `--insecure-port=0`
#### Detection
Packets on the wire with an IPv4 destination in the range 127.0.0.0/8 and a layer-2 destination MAC address of a node may indicate that an attack is targeting this vulnerability.

#### Fixed Versions
Although the issue is caused by `kube-proxy`, the current fix for the issue is in `kubelet` (although future versions may have the fix in `kube-proxy` instead). We recommend  updating both `kubelet` and `kube-proxy` to be sure the issue is addressed.

The following versions contain the fix:
  
- kubelet/kube-proxy master - fixed by #91569
- kubelet/kube-proxy v1.18.4+ - fixed by #92038
- kubelet/kube-proxy v1.17.7+ - fixed by #92039
- kubelet/kube-proxy v1.16.11+ - fixed by #92040

To upgrade, refer to the documentation: https://kubernetes.io/docs/tasks/administer-cluster/cluster-management/#upgrading-a-cluster

## Additional Details
This issue was originally raised in issue #90259 which details how the `kube-proxy` sets `net.ipv4.conf.all.route_localnet=1` which causes the system not to reject traffic to localhost which originates on other hosts.

IPv6-only services that bind to a `localhost` address are not affected. 

There may be additional attack vectors possible in addition to those fixed by #91569 and its cherry-picks. For those attacks to succeed, the target service would need to be UDP and the attack could only rely upon sending UDP datagrams since it wouldn't receive any replies. Finally, the target node would need to have reverse-path filtering disabled for an attack to have any effect. Work is ongoing to determine whether and how this issue should be fixed. See #91666 for up-to-date status on this issue.  

#### Acknowledgements
This vulnerability was reported by János Kövér, Ericsson with additional impacts reported by Rory McCune, NCC Group and Yuval Avrahami and Ariel Zelivansky, Palo Alto Networks.

/area security
/kind bug
/committee product-security
/sig network
/sig node
/area kubelet

lbernail:
**What type of PR is this?**
/kind bug
/sig network
/area ipvs

Don't set sysctlRouteLocalnet when running in IPVS mode to avoid CVE-2020-8558
Accessing a Nodeport from the host network using localhost:Nodeport never worked in IPVS (see #67730)

Fixes #92315 (for IPVS mode)

```release-note
[ACTION REQUIRED] kube-proxy's IPVS proxy mode no longer sets the net.ipv4.conf.all.route_localnet sysctl parameter. Nodes upgrading will have net.ipv4.conf.all.route_localnet set to 1 but new nodes will inherit the system default (usually 0). If you relied on any behavior requiring net.ipv4.conf.all.route_localnet, you must set ensure it is enabled as kube-proxy will no longer set it automatically. This change helps to further mitigate CVE-2020-8558.
```

### DIFF
--- a/pkg/proxy/ipvs/proxier.go
+++ b/pkg/proxy/ipvs/proxier.go
@@ -190,7 +190,6 @@ var ipsetWithIptablesChain = []struct {
 }
 
 // In IPVS proxy mode, the following flags need to be set
-const sysctlRouteLocalnet = "net/ipv4/conf/all/route_localnet"
 const sysctlBridgeCallIPTables = "net/bridge/bridge-nf-call-iptables"
 const sysctlVSConnTrack = "net/ipv4/vs/conntrack"
 const sysctlConnReuse = "net/ipv4/vs/conn_reuse_mode"
@@ -361,11 +360,6 @@ func NewProxier(ipt utiliptables.Interface,
 	nodePortAddresses []string,
 	kernelHandler KernelHandler,
 ) (*Proxier, error) {
-	// Set the route_localnet sysctl we need for
-	if err := utilproxy.EnsureSysctl(sysctl, sysctlRouteLocalnet, 1); err != nil {
-		return nil, err
-	}
-
 	// Proxy needs br_netfilter and bridge-nf-call-iptables=1 when containers
 	// are connected to a Linux bridge (but not SDN bridges).  Until most
 	// plugins handle this, log when config is missing


pkg/proxy/ipvs/proxier.go
```diff
@@ -353,7 +353,7 @@ func NewProxier(ipt utiliptables.Interface,
 	kernelHandler KernelHandler,
 ) (*Proxier, error) {
 	// Set the route_localnet sysctl we need for
```
andrewsykim, line 355: nit: update comment with context?

andrewsykim:
(some failures expected)

thockin:
Isn't #92315 fixed byhttps://github.com/kubernetes/kubernetes/pull/91569 ?

lbernail:
> Isn't #92315 fixed by #91569 ?
@thockin it is yes but the mitigation based on the kubelet patch will take a lot longer to rollout. A patch in kube-proxy is easier to roll out in most environments because you can just update the dameonset.
In addition, in the case of IPVS we don't need this sysctl (as far as I know) so there is no point in setting it

thockin:
I don't object to unsetting this in ipvs.  I'd like to figure out how to properly fix it in iptables mode too.  Reopened #90259 to discuss.

I can't locally repro the attack, so looking for other clever ideas

andrewsykim:
/retest

(some of the test failures are actually expected https://github.com/kubernetes/kubernetes/issues/92708)

tabbysable:
@thockin if you have root on the (linux) attacker machine, the following should suffice to reproduce:

```
ip addr add 127.0.0.2/8 dev lo
ip addr del 127.0.0.1/8 dev lo
ip route add 127.0.0.1/32 via YOUR-TARGET-HERE
sysctl net.ipv4.conf.all.route_localnet=1
```
With more packet gymnastics, it's also possible to reproduce using only CAP_NET_RAW; let me know on Slack or Twitter if you want to see that code before it's released.

tabbysable:
@lbernail I'm sure these things have been discussed, but it's probably good to bring them into the PR for future/public reference...

IMO, the right way to conceptualize CVE-2020-8558 is that it makes localhost-bound services *unexpectedly* available outside the host.

Back in RFC1122, they say that 127/8 addresses "MUST NOT appear outside a host" and more recently in 5735 that they "do not legitimately appear on any network anywhere." When `route_localnet=1` is set, Linux goes out of compliance with RFC5735. We've all taken that RFC so deeply to heart that the consequences of violating it take us by surprise.

A key issue with this fix (and the existing fix from #91569) is that it assumes kube-proxy is the only entity with an interest in setting `route_localnet=1`. End-users may, for reasons they believe are valid, be setting that sysctl themselves.

Forcing the sysctl to zero in IPVS mode gives a great remediation experience for end users who do not need `route_localnet=1`, but it also prevents any user who is using `route_localnet=1` on purpose from upgrading to the newest kube-proxy. Such users will *have to* fork kube-proxy or reimplement whatever clever thing they're doing.

There's the rub: We all have been breathing RFC5735 since we first learned networking. Whatever someone's doing with `route_localnet=1` is probably filled with security holes they don't realize. If we can discourage that practice, we probably should, but that conflicts with the principle of letting users configure kubernetes to meet their own needs.

Adding a knob is painful when your system already has too many knobs, but I think it should be seriously considered. My suggestion, trying to balance all these concerns is as follows:

* Add a commandline option --route-localnet-compatible
* In IPVS mode
    * By default, force `route_localnet=0` as you have proposed
    * With --route-localnet-compatible, leave the `route_localnet` state unchanged.
* In IPTables mode
    * By default, do whatever mitigation is agreed upon. (I favor removing the localhost-nodePorts feature and setting `route_localnet=0`, but the block-non-loopback-127/8-traffic-via-IPTables mitigation is acceptable, too.)
    * With --route-localnet-compatible, do not apply mitigations, because they could impact whatever the end-user is attempting to do. We trust them to apply appropriate mitigations for their situation.
* Deprecate --route-localnet-compatible in 1.20
* Remove all appropriate code (depending on what's chosen WRT localhost-nodePorts in IPTables mode) when enough time has passed that CVE-2020-8558 is no longer a concern.

WDYT?

lbernail:
@tabbysable Thanks a lot for your input (for transparency, we work together and had this discussion on slack and figured it was worth having in public).

A few thoughts:
- Yes this patch and  #91569 could break existing behaviors. We actually have a use case relying on this behavior but are working on changing it (use-case: shared service on the host that we don't want exposed outside of the node. It binds 127.0.0.1 and we redirect a specific IP to localhost with iptables so pods can access it)
- I agree we shouldn't touch a sysctl that was set outside of Kubernetes but here we felt that this would allow for a very easy patch when kube-proxy is running in a daemonset
- I also agree that localhost:nodeport should probably not be supported (but this will be hard to deprecate if users rely on it. We are lucky with IPVS because it never worked)

The flag approach is probably the best as it would allow for simple patching and users could disable it if needed. This is very similar to what we did with `strictARP` (see https://github.com/kubernetes/kubernetes/pull/75295). But of course, if we could avoid adding flags it would be better.

@andrewsykim  We also discussed this on slack but we should probably discuss it here again. What do you think? (also do you know why the test is failing?)

andrewsykim:
/retest

(fixed failing test in https://github.com/kubernetes/test-infra/pull/18339)

andrewsykim:
I agree with @tabbysable that a flag to toggle `route_localnet` makes sense, we may not have use-cases with IPVS but we do with iptables and such a flag should probably apply for both proxy modes anyways. 

At this point though, even if we add a flag to skip setting `route_localnet`, the kubelet patch in https://github.com/kubernetes/kubernetes/pull/91569 already breaks this behavior regardless of what kube-proxy is setting. I think would have to undo the kubelet iptable rules first before we make these changes to kube-proxy?

andrewsykim:
Side note around sysctl flags: maybe it's time we consider a `--sysctl-override` flag that toggles all sysctl overrides from kube-proxy  (default behavior is "true" which overrides sysctls like today). A flag per sysctl is probably not sustainable and an admin that wants to override any of the default sysctl is probably willing to configure all of them on their own. Documentation around all the sysctls we override would help as well.

TabbySableDD:
@andrewsykim Yes, I agree that the kubelet patch from #91569 should be taken out for the same reasons. It would be ideal if that and this could both happen together before 1.19 so that the behavior of 1.19 could be "CVE-2020-8558 is fixed in kube-proxy and can be switched on and off."

How to handle the kubelet patch in maintenance branches where it has already been released is another difficult question. :-(

A "don't touch my sysctls" flag sounds like a good, sustainable solution compared to separate ones per sysctl. The docs update is a very important part, so that users can make an informed decision.

tabbysable:
@andrewsykim Yes, I agree that the kubelet patch from #91569 should be taken out for the same reasons. It would be ideal if that and this could both happen together before 1.19 so that the behavior of 1.19 could be "CVE-2020-8558 is fixed in kube-proxy and can be switched on and off."

How to handle the kubelet patch in maintenance branches where it has already been released is another difficult question. :-(

A "don't touch my sysctls" flag sounds like a good, sustainable solution compared to separate ones per sysctl. The docs update is a very important part, so that users can make an informed decision.

andrewsykim:
@thockin any thoughts on the above?

lbernail:
>  maybe it's time we consider a `--sysctl-override` flag that toggles all sysctl overrides from kube-proxy

Would this mean:
1 - keep all sysctls that don't have a toggle today in kube-proxy and add a global flag controlling the ones with a toggle
2- either set all sysctls which we believe work in most cases or none

Option 2 will be harder because it means if you use the toggle today you have to reconfigure all your hosts to include the sysctls you need

And what would we do with route_localnet in 2 ? Set it to 0 when kube-proxy controls sysctls?

More than happy to help with documenting the IPVS related sysctl as I was involved with quite a few

thockin:
Sorry I missed this before!

Thanks everyone for great comments.  The use-case for localhost host-ports certainly USED TO exist.  I honestly do not know if it does any more.  We could do some outreach to try to figure it out, and use that to inform timelines.

I think I like the general approach.  If you need localhost nodeports, we can provide clear instructions on how to enable this on your own, or we can leave it in kube-proxy without doing it by default.

@tabbysable do you know of concrete reasons people might be setting this (or relying on it)?  The difficult spot here is ripping the kubelet patch out of stable branches have already cut releases.  My inclination is, absent real signal, to leave those and fix this in 1.19.  If anyone screams, we can backport the "fix" but I honestly doubt anyone will scream.

Thoughts?

thockin:
For context, the reason this doesn't work without route_localnet is that when we get to the `-j DNAT` rule (which is terminal) the packet is still src=127.0.0.1, dest=127.0.0.1, and even though we have marked it for MASQUERADE, that won't happen until later.  Before we get a chance to do that MASQUERADE, the packet is dropped (since we can't route src=127/8 to a non 127/8 destination).  I can't find any way to SNAT and DNAT *both* in PREROUTING, and I can't find any way to enable route_localnet "but only between virtual interfaces on this machine".

andrewsykim:
@lbernail can we have this PR updated to reflect the above changes? v1.19 code freeze was two weeks ago so this seems pretty urgent if we want to get it into v1.19.

andrewsykim:
So to summarize I _think_ the changes we want at this point are:
* revert the kubelet iptable rules added in https://github.com/kubernetes/kubernetes/pull/91569
* disable route_localnet by default in both the iptables and ipvs proxier
* add a flag to toggle route_localnet behavior in kube-proxy.

These changes would only be for v1.19 (or v1.20 at this point?) but we may back port to stable branches if we hear feedback from users.

Did I miss anything?

andrewsykim:
cc @joelsmith

tabbysable:
A little blog post and some code to exploit this using CAP_NET_RAW: https://github.com/tabbysable/POC-2020-8558

thockin:
I think we want the equivalent iptables rules in iptables mode with
route_localnet enabled - that's the compatible approach.

andrewsykim:
I think in order to ensure the `route_localnet` behavior is configurable, we would have to remove the iptable rules entirely? Unless the iptable rules are also behind the new flag for route_localnet. I was thinking this new flag would only toggle the sysctl value.

andrewsykim:
Spoke with Tim about this on Slack. I had originally thought the iptables rules in the kubelet were blocking node port traffic on localhost but that isn't the case. With that said, we can probably defer this to v1.20 and figure out the right approach for setting `route_localnet` then. I think disabling it for IPVS is still the right approach but not something we need to urgently add in v1.19 and backport to previous versions.   

/milestone v1.20

lbernail:
If we keep iptables in the kubelet, use-cases redirecting traffic from pods to a service binded on the host localhost will break, so I guess simply setting route_localnet to 0 in IPVS is fine (localhost:nodeport doesn't work with IPVS, and other use-case from pods won't work due to the kubelet patch).

It's simpler and avoids adding a flag, the only risk I see is for users deploying a patched kube-proxy before a patched kubelet (likely because it's easier) and breaking existing use-cases. I think we just need to add a very clear note to the release note. Down the road, I think we should just remove the sysctl in IPVS. What about:
- setting it to 0 on past versions to allow for simple patching
- not setting it all on new versions (1.19 or 1.20?) because we are protected by the kubelet
It is a bit confusing but is probably what would make mitigating the CVE as simple as possible

Another option would be to move the iptables from the kubelet to kube-proxy for 1.19 but it's probably a bit late (so we would have full control on the logic in kube-proxy)

tabbysable:
@thockin I've become convinced (through chatting with Laurent and writing this issue up) that the NodePorts-via-LocalHost use-case is the only valid one.

The use case Laurent had mentioned (offering a sensitive service only to pods on the same host) is better addressed like this:
* Choose a bit of private IP space
* Add that private IP as an IP alias on lo
* Apply iptables rules that block access to that bit of IP space except from your PodCIDR
* Bind your sensitive service to this newly-created IP alias

This plan is equivalent to binding the sensitive service to localhost and setting route_localnet, but safer because "localhost" offers a false sense of security.

I'm therefore convinced that we are OK to discourage users from privately using route_localnet on k8s nodes.

My suggested solution is now:
* In 1.16-1.19, to allow upgrading kube-proxy alone to be a remediation, without changing existing behavior in a patch release:
    * In IPVS mode: set route_localnet=0
    * In IPTables mode: set the same iptables rule as #92315
* In 1.20 (or as soon as the deprecation policy allows, if NotePorts via Localhost is considered a feature subject to that policy)
    * Remove all this code
    * Don't touch route_localnet (breaking NodePorts via Localhost in IPTables mode)
    * Trust the OS to do the right thing WRT route_localnet

This plan would cause in-place 1.20->1.21 upgraders to be protected by the iptables rule until their first reboot, then route_localnet=0 (via OS default) subsequently.

andrewsykim:
The IPVS mode changes proposed by @lbernail and @tabbysable make sense to me:
* explicitly set `route_loclanet=0` on v1.16 - v1.19 since the kernel never allowed localhost:nodeport with ipvs anyways
* starting v1.20 we don't set it at all, which in most cases it will be set to 0 but users can flip it on. 

As for the iptables mode changes, I'm not sure yet but everything proposed seems reasonable at first glance.

I think that means this PR as-is is good to merge once the code freeze is lifted and we can cherry-pick it into the stable branches. I think the current e2e test failure is legit though. I don't see this test failing in other PRs:
```
[sig-auth] Metadata Concealment should run a check-metadata-concealment job to completion
```

andrewsykim:
Looks like "metadata concealment" is something that would actually break on gce with this change since it requires a DNAT to localhost: https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/gci/configure-helper.sh#L174

lbernail:
@andrewsykim well this is exactly the type of use-cases that requires route_localnet:
```
-j DNAT --to-destination 127.0.0.1:988
```
This should also break with the kubelet patch because traffic will be dropped

andrewsykim:
I don't think the kubelet patch in https://github.com/kubernetes/kubernetes/pull/91569 actually breaks this because it only blocks traffic that is not part of an existing forwarded connection. Otherwise `pull-kubernetes-e2e-gce` should be failing.

andrewsykim:
The `! --ctstate RELATED,ESTABLISHED,DNAT` is the check that is allowing DNAT to localhost, as long as the traffic is origination from the local machine.

lbernail:
> The `! --ctstate RELATED,ESTABLISHED,DNAT` is the check that is allowing DNAT to localhost, as long as the traffic is origination from the local machine.

I just discovered `--ctstate DNAT`, interesting

TabbySableDD:
Digging into the failure of `pull-kubernetes-e2e-gci-gce-ipvs` has revealed that metadata-proxy is currently dependent on `route_localnet=1` in its default installation. (I have brought this to their attention.)

I suspect this is only the first of many high-profile deployments that will be broken by kube-proxy forcibly setting `route_localnet=0`.

I fear this means that best *practical* short term plan is:
* In 1.16-1.19, to allow upgrading kube-proxy alone to be a remediation, without changing existing behavior in a patch release:
    * In IPVS mode: leave `route_localnet` alone; Set the same iptables rule as #91569
    * In IPTables mode: set the same iptables rule as #91569
* In 1.20 (or as soon as the deprecation policy allows, if NotePorts via Localhost is considered a feature subject to that policy)
    * Remove all this code from both kube-proxy and kubelet
    * Don't touch route_localnet at all in any mode (breaking NodePorts via Localhost in IPTables mode)
    * Trust the OS to do the right thing WRT route_localnet

mikedanese:
@zshihang is taking a look at metadata-proxy.

zshihang:
could you rebase and retest?

bai:
Hey 👋 Bug Triage here. This PR has not been updated for a while, so I'd like to check on the status. The code freeze is starting 12.11.2020 (about 3-4 weeks from now) and while there is still plenty of time, we want to ensure that each PR has a chance to be merged on time.

tabbysable:
I'll ping Laurent about this in early November, when we hope to have more reasonable schedules. Thanks!

erismaster:
Hi 👋🏽 I'm from the Bug Triage team. We've crossed the Code Freeze for 1.20 release on 12th November. As this PR is tagged with 1.20, I'm sending a final reminder to either move the milestone to 1.21 or clear the milestone.

thockin:
Hi folks!  As 1.20 and 2020 wrap up, I guess it is time to revisit this.

It sounds like the iptables hack is a bandaid, but does close the gap in all cases, right?  

It also sounds like we should really just EOL the nodeport-on-localhost behavior since it doesn't work on v6 or IPVS, right?

But that's a breaking change.  In the short-to-medium term, we can make a flag to opt-in to no-localnet mode, with localnet + iptables bandaid as the default.

The real trick is helping a cluster admin determine if it is safe to set no-localnet.

We could sample conntrack looking for signature records, but that's weak on TCP.

We could add a -j LOG to capture evidence in dmesg and then scrape that and report it   (`iptables -t filter -I OUTPUT -s 127.0.0.0/8 ! -d 127.0.0.0/8 -m conntrack --ctstate NEW -j LOG --log-prefix "LOCALNET:  "`).  Kind of stinks for UDP, maybe we can do better.

Other ideas?

uablrek:
My vote is on;

* Do not _require_ or set this sysctl in any proxier
* Log a security warning if it is on
* Remove entries for localhost for proxy-mode=ipvs (that never have worked, #96879 )
* Remove the ipv6 entries for localhost for proxy-mode=iptables (that never have worked)
* Keep the entries for ipv4 for proxy-mode=iptables for backward compatibility

Let the user decide if the sysctl should be on. It is off by default so any node-reboot will reset it.

The only use-case I have heard (from @thockin ) is for a local docker registry that by-passes security for localhost. The users are most likely very few, but I am sure there are some.

Here is the requirement check;
https://github.com/kubernetes/kubernetes/blob/6e41006e1dc344be71a82586820f89f7752cde30/pkg/proxy/iptables/proxier.go#L101-L107

uablrek:
> Let the user decide if the sysctl should be on.

BTW K3s relies on this (or at least they did). I tried to get ipv6 running ~1.5y ago but failed. I did not understand why at the time;
https://github.com/k3s-io/k3s/issues/284#issuecomment-488588527

danwinship:
> It sounds like the iptables hack is a bandaid, but does close the gap in all cases, right?

#91666 is still open, although [this comment](https://github.com/kubernetes/kubernetes/pull/91666#issuecomment-657079776) suggests maybe the theory behind it was wrong and there is not actually a bug? (I pinged.)

> It also sounds like we should really just EOL the nodeport-on-localhost behavior since it doesn't work on v6 or IPVS, right?

Yeah.

Though, I guess the alternative would be to use the userspace proxy code for localhost connections. This is actually not _quite_ as terrible as it sounds; instead of using `utilproxy.PortOpener` to hold open a no-op socket for each service, we'd create a `userspace.ProxySocket` for each one instead, so if any connection actually reaches that socket (because there was no iptables/ipvs rule to redirect it away), then we could userspace-redirect it instead.

(But also mumble mumble *host*port-on-localhost too. Though maybe it could all be moved into `k8s.io/utils/net/port.go`.)

> The real trick is helping a cluster admin determine if it is safe to set no-localnet.

Yeah but what if it's not? We still have to solve things for those users too, so this doesn't seem like the right direction.

pkg/proxy/ipvs/proxier.go
```diff
@@ -352,8 +352,10 @@ func NewProxier(ipt utiliptables.Interface,
 	nodePortAddresses []string,
 	kernelHandler KernelHandler,
 ) (*Proxier, error) {
-	// Set the route_localnet sysctl we need for
-	if err := utilproxy.EnsureSysctl(sysctl, sysctlRouteLocalnet, 1); err != nil {
+	// Enabling sysctlRouteLocalnet created a security issue (see CVE-2020-8558)
+	// We are purposefully setting it to 0 to allow mitigating the issue on existing clusters
+	// This should not impact existing workloads as accessing localhost:Nodeport from the host does not work with IPVS
+	if err := utilproxy.EnsureSysctl(sysctl, sysctlRouteLocalnet, 0); err != nil {
```
danwinship, line 358: It seems like the IPVS proxier was just setting `route_localnet` to 1 because it copied that code from the iptables proxier where it was actually needed, but IPVS doesn't care at all whether the sysctl is set or not, so the fix is just to _not set it_.

danwinship (changes requested):

pkg/proxy/ipvs/proxier.go
andrewsykim: This sounds accurate. Given IPVS doesn't actually allow localhost:nodeport it's likely not in use. A user could have added iptables rules that relied on kube-proxy setting this in IPVS mode but I don't think that's our problem.

aojea: so, do we remove the sysctl @lbernail ?

andrewsykim:
I am in favor of this and agree on @uablrek's comment above https://github.com/kubernetes/kubernetes/pull/92938#issuecomment-742985382, especially for IPVS since it never really worked. My only remaining questions are:
1. Should we disable for iptables? I'm not sure but it would be nice if we can disable this for iptables and IPVS on the same release.
2. Would this warrant reverting the kubelet patch https://github.com/kubernetes/kubernetes/pull/91569? I think it depends on what we do for iptables proxy. If we disable it for both it seems like we should revert that kubelet patch, if we only disable for IPVS it seems like we need to keep it around for iptables.
3. Should we actually set `sysctlRouteLocalnet` to 0 or stop setting it entirely? I'm in favor of the latter where we stop setting it at all. This allows for users to set it back if they were depending on it for something critical.

pkg/proxy/ipvs/proxier.go
andrewsykim: yes, I think we should stop setting this sysctl, at least this allows users to set it if they needed it for something critical

lbernail:
As discussed with @andrewsykim and @tabbysable : let's not set the sysctl (because we don't need it in IPVS mode). Deploying kube-proxy won't fix the CVE but at least it will not introduce it on new nodes (note that if the version of the kubelet in recent enough, the CVE is mitigated)

andrewsykim:
LGTM but release notes need an update

lbernail:
> LGTM but release notes need an update

Is the description clear enough? I could make it longer but release notes tend to be very short

andrewsykim:
I think this one probably warrants a longer one with an "ACTION REQUIRED" in it, maybe something like:
```
[ACTION REQUIRED] kube-proxy's IPVS proxy mode no longer sets the net.ipv4.conf.all.route_localnet sysctl parameter. Nodes upgrading will have net.ipv4.conf.all.route_localnet set to 1 but new nodes will inherit the system default (usually 0). If you relied on any behavior requiring net.ipv4.conf.all.route_localnet, you must set ensure it is enabled as kube-proxy will no longer set it automatically. This change helps to further mitigate CVE-2020-8558.
```

andrewsykim (commented):
/lgtm

PR #92941, opened 2020-07-09
Don't return proxied redirects to the client

tallclair:
**What type of PR is this?**
/kind bug

Treat all non-101 (switching protocol) responses to upgrade requests as errors. If the server responds with a 200 but the client is expecting an upgrade, the behavior can be unpredictable and lead to a stuck client.

```release-note
CVE-2020-8559 (Medium): Privilege escalation from compromised node to cluster. See https://github.com/kubernetes/kubernetes/issues/92914 for more details.
Action Required: The API Server will no longer proxy non-101 responses for upgrade requests. This could break proxied backends (such as an extension API server) that respond to upgrade requests with a non-101 response code.
```

/sig api-machinery
/milestone v1.19
/priority important-soon
/assign @liggitt @deads2k

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/util/net/http.go
+++ b/staging/src/k8s.io/apimachinery/pkg/util/net/http.go
@@ -453,7 +453,7 @@ redirectLoop:
 
 		// Only follow redirects to the same host. Otherwise, propagate the redirect response back.
 		if requireSameHostRedirects && location.Hostname() != originalLocation.Hostname() {
-			break redirectLoop
+			return nil, nil, fmt.Errorf("hostname mismatch: expected %s, found %s", originalLocation.Hostname(), location.Hostname())
 		}
 
 		// Reset the connection.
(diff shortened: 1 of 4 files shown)


staging/src/k8s.io/apimachinery/pkg/util/net/http.go
```diff
@@ -453,7 +453,7 @@ redirectLoop:
 
 		// Only follow redirects to the same host. Otherwise, propagate the redirect response back.
 		if requireSameHostRedirects && location.Hostname() != originalLocation.Hostname() {
-			break redirectLoop
+			return nil, nil, fmt.Errorf("hostname mismatch: expected %s, found %s", originalLocation.Hostname(), location.Hostname())
```
deads2k, line 456: This change doesn't immediately seem related to the release note, but seems ok.

deads2k:
> that respond to upgrade requests with a non-101 response code.

Based on my read, error responses (greater than 400) are still allowed.  I like this, it does what I need, but I want to make sure it is intentional on your part.  

/lgtm
/approve
/hold

holding in case you want to do something different with errors.  if you're ok passing them back as failures (I like that), feel free to release the hold.

tallclair:
@deads2k yep, that's intentional, for the same reasons you highlighted. Thanks!

/hold cancel

Issue #88752, opened 2020-03-03
Resource consumer controller does not respect configured dns domain


vryzhenkin:
**What happened**:
During execution of HPA e2e part on kubespray deployment environment with non-default domain 7 of 8 HPA tests failed. One passed by accident due it require only downscale.

**What you expected to happen**:
All 8 tests passed

**How to reproduce it (as minimally and precisely as possible)**:
1. Deploy a Kubernetes cluster using any desired tool with any domain setting that differs from cluster.local
2. Execute e2e tests with regex [Feature:HPA]

**Anything else we need to know?**:
Discovered using metrics server 0.3.6.
Domain set to different than default

During normal execution was observed that creating of consume requests doing nothing.
```
Mar  3 01:27:04.782: INFO: RC test-deployment: sending request to consume 325 millicores
Mar  3 01:27:04.782: INFO: ConsumeCPU URL: {https   172.19.116.45:443 /api/v1/namespaces/horizontal-pod-autoscaling-470/services/test-deployment-ctrl/proxy/ConsumeCPU  false durationSec=30&millicores=325&requestSizeMillicores=100 }
Mar  3 01:27:09.542: INFO: waiting for 3 replicas (current: 1)
Mar  3 01:27:29.543: INFO: waiting for 3 replicas (current: 1)
Mar  3 01:27:29.774: INFO: RC test-deployment: sending request to consume 0 MB
Mar  3 01:27:29.774: INFO: ConsumeMem URL: {https   172.19.116.45:443 /api/v1/namespaces/horizontal-pod-autoscaling-470/services/test-deployment-ctrl/proxy/ConsumeMem  false durationSec=30&megabytes=0&requestSizeMegabytes=100 }
Mar  3 01:27:30.052: INFO: RC test-deployment: sending request to consume 0 of custom metric QPS
Mar  3 01:27:30.052: INFO: ConsumeCustomMetric URL: {https   172.19.116.45:443 /api/v1/namespaces/horizontal-pod-autoscaling-470/services/test-deployment-ctrl/proxy/BumpMetric  false delta=0&durationSec=30&metric=QPS&requestSizeMetrics=10 }
```
By idea such requests goes through controller and then - controller sending it to consumers. But consumers continue consume 1Mi MEM and 0 CPU even in case I saw a request from e2e to controller.
```
freerunner@pegassie:[~/metrics-server] (master)$ kubectl top pod -n horizontal-pod-autoscaling-470
[...]

PR #92964, opened 2020-07-10
agnhost image: use actual DNS domain instead of hardcoded cluster.local

vponomaryov:
'agnhost' image uses hardcoded 'cluster.local' value for DNS domain.
It leads to failure of a bunch of HPA tests when test cluster is
configured to use custom DNS domain and there is no alias for
default 'cluster.local' one.
So, fix it by reusing it's own function for reading DNS domain suffixes.

Bumps 'agnhost' image version to 2.24

Signed-off-by: Valerii Ponomarov <kiparis.kh@gmail.com>

**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

```release-note
NONE
```

```docs

```


vponomaryov:
This PR fixes following issue: https://github.com/kubernetes/kubernetes/issues/88752

vponomaryov:
Fixes #88752

vponomaryov:
I see that failed jobs cannot pull docker image with new tag. Obviously, that new image doesn't get created/uploaded, so it is unavailable. How is it solved when helper image logic gets updated?

claudiubelu:
It's a 3-step process.

- do the desired image changes. Basically, don't change the image version in ``test/util/manifest.go``. We'll merge this pull request, which will trigger the Image Promoter process.
- Promote the new image (see more here: https://github.com/kubernetes/kubernetes/tree/master/test/images#making-changes-to-images)
- send a PR with the updated image version in ``test/util/manifest.go``.

vponomaryov:
@claudiubelu Thank you for clarification! Updated PR.

test/images/agnhost/resource-consumer-controller/controller.go
```diff
@@ -214,7 +240,7 @@ func (c *controller) sendConsumeCustomMetric(w http.ResponseWriter, metric strin
 }
 
 func createConsumerURL(suffix string) string {
-	return fmt.Sprintf("http://%s.%s.svc.cluster.local:%d%s", consumerServiceName, consumerServiceNamespace, consumerPort, suffix)
```
aojea, line 217: are other tests depending on this?
should not keep it as default?

vponomaryov: With proposed change we will have working DNS suffix, because it is taken from real DNS config. So, if "cluster.local" is configured, it will be picked up. That's the trick.

aojea: great

vponomaryov:
Is it possible to have it added to the "v1.19" milestone?

test/images/agnhost/resource-consumer-controller/controller.go
```diff
@@ -43,8 +46,31 @@ var (
...
 	consumerServiceName      string
 	consumerServiceNamespace string
+	dnsDomain                string
 )
 
+func getDNSDomain() string {
+	if dnsDomain != "" {
+		return dnsDomain
+	}
+	dnsSuffixList, err := exec.Command("/agnhost", "dns-suffix").Output()
```
claudiubelu, line 56: This will actually break Windows support. You can't run binaries without any extension on Windows, so this automatically fails.

I don't know why you have to do an exec call. Why not just import and use ``agnhost/dns`` module directly?

vponomaryov: Docker image entrypoint for Linux and Windows is the same according to the [1] and [2]. So, looking at it I assume that either agnhost windows image is broken or this code must be working.

I don't have Windows cluster to test it there.

[1] https://github.com/kubernetes/kubernetes/blob/master/test/images/agnhost/Dockerfile
[2] https://github.com/kubernetes/kubernetes/blob/master/test/images/agnhost/Dockerfile_windows

claudiubelu: Docker works a bit differently, so having ``/agnhost`` as an entrypoint on a Windows image is fine. This is backed by the fact that we've been using this image in our CIs for a very long time.

With your changes, this is what it'll be printed out when using the image on a Windows node: ``panic: exec: "/agnhost": file does not exist``. I've already have to investigate this issue for the resource-consumer image.

Nonetheless, my point still stands: why not just import the ``agnhost/dns`` module and use it? It's faster, safer, and it makes more sense.

vponomaryov: I can reuse functions directly. Just keep in mind that it is "private" as of now and cannot be imported from another module. So, I need to make it public (update func name to start with Uppercase letter). Does it work for you?

vponomaryov: @claudiubelu, Ok, uploaded change with the direct func usage. Please, re-review.

liggitt:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92964/pull-kubernetes-node-e2e/1283440500241600512 flake is #93210

vponomaryov:
@claudiubelu , does current state of bugfix work for you? Or something else need to be updated?

claudiubelu:
Indeed, the changes look fine now. I've built my own image with these changes and I've ran the HPA tests which are using ``resource-consumer-controller`` and they've passed on my Windows env.

/lgtm

test/images/agnhost/resource-consumer-controller/controller.go
liggitt: is there a reason not to use `http://%s.%s.svc:%d%s`?

BenTheElder:
Does this cover all places e2e.test assumes cluster.local?

test/images/agnhost/resource-consumer-controller/controller.go
vponomaryov: > is there a reason not to use `http://%s.%s.svc:%d%s`? that also resolves to the service and doesn't require custom DNS suffix lookup

I was not sure whether it is needed for someone to use full DNS name or not. So, kept the behaviour - full DNS name having correct suffix. For HPA tests just "svc" is enough.
I don't mind to simplify the PR to just removing this ".cluster.local" part of DNS. Should I?

vponomaryov:
> Does this cover all places e2e.test assumes cluster.local?

If answering precisely then "no". We have "dns-domain" config option for other places. This one fixes places where "agnhost" image is used.

test/images/agnhost/resource-consumer-controller/controller.go
claudiubelu: No, leave it as full DNS name, otherwise it'll break Windows. :) It doesn't like partially qualified names, and that would break some tests.

liggitt: so... windows doesn't support the kube approach to dns?

vponomaryov: So, if it is a must to keep old behavior then, I guess, we keep it as is and can move on.

claudiubelu:
From what I found, the ``peer-finder`` image also has a ``-domain=cluster.local`` flag, so this PR is in line with that. There is only one other place where ``cluster.local`` is hardcoded, and that's agnhost's resource-consumer-controller subcommand. If that is also made configurable, I think that solves Ben's question.

vponomaryov:
> From what I found, the `peer-finder` image also has a `-domain=cluster.local` flag, so this PR is in line with that. There is only one other place where `cluster.local` is hardcoded, and that's agnhost's resource-consumer-controller subcommand. If that is also made configurable, I think that solves Ben's question.

Looked at `peer-finder` image, and I can say that it doesn't hardcode "cluster.local". In case "domain" config option is not set it autocalculates it using it's own code, not reusing "agnhost" image. It's problem in not having Windows support when "domain" config option is not set, but it is different kind of an issue, IMHO.
This particular change fixes general impossibility to run bunch of tests (HPA tests) when cluster domain is different than "cluster.local" even if "domain" config option is set.

claudiubelu:
> 
> 
> > From what I found, the `peer-finder` image also has a `-domain=cluster.local` flag, so this PR is in line with that. There is only one other place where `cluster.local` is hardcoded, and that's agnhost's resource-consumer-controller subcommand. If that is also made configurable, I think that solves Ben's question.
> 
> Looked at `peer-finder` image, and I can say that it doesn't hardcode "cluster.local". In case "domain" config option is not set it autocalculates it using it's own code, not reusing "agnhost" image. It's problem in not having Windows support when "domain" config option is not set, but it is different kind of an issue, IMHO.
> This particular change fixes general impossibility to run bunch of tests (HPA tests) when cluster domain is different than "cluster.local" even if "domain" config option is set.

I was just mentioning that there is a precedence to what you're adding, so that's good. I didn't realize then that this PR was updating the resource-consumer-controller subcommand. :)

IMO, LGTM.

vponomaryov:
@claudiubelu , so can you please set "/lgtm" label?

vponomaryov:
@justaugustus thanks for adding milestone label!
@claudiubelu , as I understood, this PR requires your "/approve" first, as owner at [1], and only then "/approve" from someone at [2]. 

[1] https://github.com/kubernetes/kubernetes/blob/master/test/images/agnhost/OWNERS
[2] https://github.com/kubernetes/kubernetes/blob/master/build/OWNERS

claudiubelu:
Indeed, I didn't approve it since it also has changes outside of the agnhost image, so my approval wouldn't make this go in anyways.

/approve

liggitt:
cc @kubernetes/sig-network-pr-reviews

liggitt:
/assign @thockin 

would like your $0.02 on the DNS changes here

test/images/agnhost/dns/dns.go
```diff
@@ -28,7 +28,9 @@ const etcHostsFile = "/etc/hosts"
 // nameserver DNS_CLUSTER_IP
 // search test-dns.svc.cluster.local svc.cluster.local cluster.local q53aahaikqaehcai3ylfqdtc5b.bx.internal.cloudapp.net
 // options ndots:5
-func getDNSSuffixList() []string {
+
+// GetDNSSuffixList reads DNS config file and returns the list of configured DNS suffixes
+func GetDNSSuffixList() []string {
```
thockin, line 33: FWIW using strings.Fields() would be safer WRT repeated whitespace

test/images/agnhost/resource-consumer-controller/controller.go
```diff
@@ -43,8 +45,28 @@ var (
 	consumerPort             int
 	consumerServiceName      string
 	consumerServiceNamespace string
+	dnsDomain                string
 )
 
+func getDNSDomain() string {
```
thockin, line 51: Can we please get a comment explaining how this is operating and why?  It's PROBABLY safe, all things considered, but it's a heuristic.

This has come up in other places, and we really ought to expose it through downward API or something like that.

test/images/agnhost/resource-consumer-controller/controller.go
thockin: WRT Windows, this is one of a long list of things that don't work right in Windows which were deemed "acceptable because no Windows user would expect them to work", even if a Kube user would.  Sigh.

test/images/agnhost/agnhost.go
```diff
@@ -51,7 +51,7 @@ import (
 func main() {
 	rootCmd := &cobra.Command{
 		Use:     "app",
-		Version: "2.22",
+		Version: "2.23",
```
thockin, line 54: Why do we need to update the version in 3 places?  Can't we inject it at build time via linker?

thockin (commented):
LGTM, I just want this self-documenting

test/images/agnhost/agnhost.go
vponomaryov: I agree that it is not ok to have such kind of duplication, but I do not implement it here. I just do minimal changes needed to fix the issue.

test/images/agnhost/dns/dns.go
vponomaryov: The only thing that is changed here is that function becomes exported having uppercase first letter. So, I do not understand the point of the comment.

test/images/agnhost/resource-consumer-controller/controller.go
vponomaryov: Maybe it is good to have comments, but even this module has more complex functions without comments.
According to the activity on the bug-report which gets fixed here, no one really cared about it. So, I wonder exactly where it has come up?

test/images/agnhost/dns/dns.go
liggitt: exporting it and starting to use it as a utility method for other packages expands the scope of the code and increases the importance that it be robust/correct

thockin: Strictly speaking this function gets a list of "search paths" and if I were to find "" in that list (which you would if there was a double-space) I would be very surprised.

Is there a test-case for this function?  I'm not going to demand you go write one, unless you want to, but it probably should have a test including values with duplicate whitespace

test/images/agnhost/resource-consumer-controller/controller.go
thockin: Past mistakes do not mean we shouldn't do better.  The function is sufficiently subtle that having an explanation of what it is supposed to be doing and why is justified.  I'm not asking for a book, just a couple sentences. Less than this comment, probably.

test/images/agnhost/agnhost.go
thockin: sure, that's fine - I don't know who owns this sub-component now, so I guess I hope they see this and agree.

test/images/agnhost/dns/dns.go
vponomaryov: @liggitt , @thockin I agree that it would be good to fix this thing, but I disagree that it must be part of this particular change that has one concrete goal - fixing concrete bug. And the change you propose me doing is not related to the bugfix at all.

Then, about unit tests, the fact that I changed the name of the function and unit tests passed means that there are either no tests or they were not run. In anyway, it doesn't get tested on regular basis. So, it has much more TODOs which must be done, but definitely not as part of this change, IMHO.

test/images/agnhost/resource-consumer-controller/controller.go
vponomaryov: Added comment in the code here and in the place of it's call.

johngmyers:
I've always wondered why the cluster domain isn't available through the downward API.

vponomaryov:
@claudiubelu can you please revisit this PR?

fejta (commented):
/approve
the build changes

PR #90783, opened 2020-05-06
Clarify behavior of publishNotReadyAddresses

JeremyOT:
**What type of PR is this?**
/kind documentation

**What this PR does / why we need it**:
The implementation of `service.spec.publishNotReadyAddresses = true` causes pod readiness to be ignored and all endpoints created by the controller to be marked ready. This has always been the behavior and is not expected to change, so it should be discoverable via documentation.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

Issue #85643, opened 2019-11-26
Create ability to do zero downtime deployments when using externalTrafficPolicy: Local


WillPlatnick:
I am using externalTrafficPolicy set to Local for my a LoadBalancer service for an ingress controller on GKE.

Right now, when a pod gets terminated, it is immediately removed from the NodePort service, which stops traffic from routing to the pod (step 5 at https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods). 

The problem is that the GCP Load Balancer doesn't update itself immediately, so it continues to send traffic to the NodePort even though Kubernetes has already removed the pod from the NodePort as part of the termination process. This results in timeouts and an inability to do zero downtime deployments when a node no longer has an active application residing on it when externalTrafficPolicy is set to Local.

I'd like to see an option where we can use Local, but allow for zero-downtime deployments. 

I'm wondering if there could be a configurable option to wait until a preStop hook has finished (or grace period hits) before removing the pod from the NodePort service? With something like this, we could make a preStop hook that can make health checks fail but have the pod continue to serve traffic normally. The preStop hook could then sleep for a certain amount of time while the load balancers gracefully stop sending traffic because the health checks start to fail. Once the preStop hook completes, then it removes the pod from the NodePort. This would allow for graceful draining of outgoing pods.

Or maybe the answer is a pre-PreStop hook that can run before termination officially begins?

athenabot:
/triage unresolved

Comment `/remove-triage unresolved` when the issue is assessed and confirmed.

🤖 I am a bot run by vllry. 👩‍🔬

tuananhnguyen-ct:
Having the same issue. Any graceful termination in the pod itself will not help, since the port at node level stop accepting traffic.

kevinscholz:
Same issue here. I tried to fix in a few months ago but gave up in the end and set `externalTrafficPolicy` to `Cluster` (loosing the client ip but deploy zero-downtime). The issue is that the Google LBs don't get a signal BEFORE a pod is terminated, so only after the not configurable healthcheck timeout the node is deregistered from the LB. Found no way so far to remove the node from the LB before terminating the POD.

thockin:
[...]

PR #92968, opened 2020-07-10
Add Terminating Condition to EndpointSlice

andrewsykim:
**What type of PR is this?**
/kind feature

Adds a `serving` and `terminating` condition to EndpointSlice as proposed in this [KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1672-tracking-terminating-endpoints). 

https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1672-tracking-terminating-endpoints 

```release-note
Add a 'serving' and `terminating` condition to the EndpointSlice API.

`serving` tracks the readiness of endpoints regardless of their terminating state. This is distinct from `ready` since `ready` is only true when pods are not terminating. 
`terminating` is true when an endpoint is terminating. For pods this is any endpoint with a deletion timestamp.
```

```docs

```


andrewsykim (commented):
/assign @thockin @robscott 
/retest

pkg/registry/discovery/endpointslice/strategy.go
```diff
@@ -67,6 +75,12 @@ func (endpointSliceStrategy) PrepareForUpdate(ctx context.Context, obj, old runt
 
 	newEPS.ObjectMeta = ogNewMeta
 	oldEPS.ObjectMeta = ogOldMeta
+
+	if !utilfeature.DefaultFeatureGate.Enabled(features.EndpointSliceTerminatingCondition) {
+		for i := range newEPS.Endpoints {
+			newEPS.Endpoints[i].Conditions.Terminating = nil
```
robscott, line 81: I think we need to maintain an existing value here if it is already set to ensure compatibility with future versions that could set this field.

pkg/controller/endpointslice/utils_test.go
```diff
@@ -223,9 +268,12 @@ func TestPodToEndpoint(t *testing.T) {
 			node: node1,
 			svc:  &svc,
 			expectedEndpoint: discovery.Endpoint{
-				Addresses:  []string{"1.2.3.5"},
-				Conditions: discovery.EndpointConditions{Ready: utilpointer.BoolPtr(true)},
-				Hostname:   &readyPodHostname.Spec.Hostname,
+				Addresses: []string{"1.2.3.5"},
+				Conditions: discovery.EndpointConditions{
+					Ready:       utilpointer.BoolPtr(true),
+					Terminating: utilpointer.BoolPtr(false),
```
robscott, line 274: Are there any instances where we would expect terminating to be nil when the feature gate is enabled? It looks like all of these tests expect true or false values.

pkg/controller/endpointslice/utils_test.go
```diff
@@ -73,18 +77,21 @@ func TestNewEndpointSlice(t *testing.T) {
 }
 
 func TestPodToEndpoint(t *testing.T) {
+	defer featuregatetesting.SetFeatureGateDuringTest(t, utilfeature.DefaultFeatureGate, features.EndpointSliceTerminatingCondition, true)()
```
robscott, line 80: Can we make this something that applies per test case instead of for all of `TestPodToEndpoint`. Setting this to `true` here means we lose all our test coverage for the detault state where the feature gate is disabled. I think it would be helpful to leave the majority of existing tests covering that state and then add test cases that are specific to this feature. As an example, I think the default value for 1.20 is going to be nil but that is not actually covered by the tests here.

pkg/controller/endpointslice/endpointslice_controller_test.go
```diff
@@ -329,16 +332,18 @@ func TestOnEndpointSliceUpdate(t *testing.T) {
 
 // Ensure SyncService handles a variety of protocols and IPs appropriately.
 func TestSyncServiceFull(t *testing.T) {
+	defer featuregatetesting.SetFeatureGateDuringTest(t, utilfeature.DefaultFeatureGate, features.EndpointSliceTerminatingCondition, true)()
```
robscott, line 335: Same comment here, I hate to lose test coverage of existing/default functionality. Until this feature gate is enabled by default I'd prefer to have any test coverage be in addition to existing test coverage.

pkg/apis/discovery/types.go
```diff
@@ -116,6 +116,11 @@ type EndpointConditions struct {
 	// indicates an unknown state. In most cases consumers should interpret this
 	// unknown state as ready.
 	Ready *bool
+
+	// terminating indicates if this endpoint is terminating. Consumers should assume a
+	// nil value indicates the endpoint is not terminating.
```
robscott, line 121: Maybe something closer to the comment for `ready` above?
```suggestion
	// terminating indicates if this endpoint is terminating. A nil value indicates an
	// unknown state. Consumers should interpret this unknown state indicates the
	// endpoint is not terminating.
```

robscott (commented):
Thanks for the work on this!

pkg/registry/discovery/endpointslice/strategy.go
andrewsykim: Those future versions would have the feature gate enabled so we wouldn't hit this code path right?

pkg/controller/endpointslice/utils_test.go
andrewsykim: I don't think there is, at least in the context of endpointslice controller that is only adding pod endpoints. The terminating state would be either `true` or `false` depending on whether `DeletionTimestamp` is nil. Maybe I missed an edge case here though :thinking: 

Will add some tests for when the gate is disabled though

pkg/controller/endpointslice/utils_test.go
andrewsykim: Yup, good cal!

pkg/apis/discovery/types.go
andrewsykim: Will apply

pkg/controller/endpointslice/endpointslice_controller_test.go
```diff
@@ -327,18 +330,103 @@ func TestOnEndpointSliceUpdate(t *testing.T) {
 	assert.Equal(t, 1, esController.queue.Len())
 }
 
-// Ensure SyncService handles a variety of protocols and IPs appropriately.
-func TestSyncServiceFull(t *testing.T) {
+// Ensure SyncService works with ipFamily=IPv4 but with endpoints from multiple families
+func TestSyncService_IPv4(t *testing.T) {
```
andrewsykim, line 334: This is the same test as `TestSyncService_IPv6` below but using `ipFamily=IPv4` on the Service.

pkg/controller/endpointslice/endpointslice_controller_test.go
```diff
@@ -400,6 +488,95 @@ func TestSyncServiceFull(t *testing.T) {
 	}}, slice.Endpoints)
 }
 
+// Ensure SyncService works the EndpointSliceTerminatingConditoin gate enabled
+func TestSyncService_WithEndpointSliceTerminatingConditionEnabled(t *testing.T) {
```
andrewsykim, line 492: This is the same test as `TestSyncService_IPv4` but with the termination gate enabled.

pkg/registry/discovery/endpointslice/strategy.go
robscott: I think the idea here is that we need to be able to safely round trip between API versions. As an example, if a resource was created with a 1.21 API Server, data from that resource should not be dropped/unset by a 1.20 API Server. At least that's what I had to do for [AppProtocol in 1.18](https://github.com/kubernetes/kubernetes/blob/release-1.18/pkg/apis/core/validation/validation.go#L4273-L4283) when it was alpha.

andrewsykim: Okay, this makes sense now, thanks for explaining! :)

andrewsykim: PR updated, based on your feedback

pkg/controller/endpointslice/endpointslice_controller_test.go
```diff
@@ -327,18 +330,103 @@ func TestOnEndpointSliceUpdate(t *testing.T) {
...
 }
 
-// Ensure SyncService handles a variety of protocols and IPs appropriately.
-func TestSyncServiceFull(t *testing.T) {
+// Ensure SyncService works with ipFamily=IPv4 but with endpoints from multiple families
+func TestSyncService_IPv4(t *testing.T) {
+	client, esController := newController([]string{"node-1"}, time.Duration(0))
+	namespace := metav1.NamespaceDefault
+	serviceName := "all-the-protocols"
+	ipv6Family := v1.IPv4Protocol
```
robscott, line 338: This seems wrong? Maybe just `ipFamily`?

pkg/controller/endpointslice/endpointslice_controller_test.go
robscott: Is this similar enough to the IPv4 one that we could just have a single test function with 2 test cases?

pkg/controller/endpointslice/endpointslice_controller_test.go
robscott: Similarly, can we just rework 1 function to support multiple test cases?

robscott (commented):
Implementation LGTM, just had some questions about tests.

pkg/controller/endpointslice/endpointslice_controller_test.go
andrewsykim: Good catch

pkg/controller/endpointslice/endpointslice_controller_test.go
andrewsykim: While I agree, I think putting all the `TestSyncService*` into one unit test with test tables is gonna require a bigger refactor due to the way the pod IPs end up getting filtered in the EndpointSlice resource.

andrewsykim: Happy to do the refactor if you think it's worthwhile

robscott: It looks like each test case would need to define pods, ipFamily, and expected endpoints. It looks like the Service and EndpointPorts are always consistent so we may not need to check them or define them as part of test cases. If I'm understanding the changes in this PR correctly, it looks like the `TestSyncService` func is being repeated 3x with relatively minimal changes, so if that can be pulled back into a single function with multiple test cases I think that would be idea. With that said, I'm probably missing some extra bits of variation between these test functions that might make this more complex. So definitely let me know if I'm missing anything.

andrewsykim: The only other sticking point is that `newPod` has some custom indexing logic and in this test case we have a separate step to override `status.podIPs`. If we use test tables, defining the pods directly in the table would make the most sense but would require some refactoring.

I agree with you that it would be much better, just not sure it's worth the extra effort :thinking:

andrewsykim: Actually scratch that, I'll do the refactor. I'll keep it in a separate commit so we can revert if we find it's changing too many things.

pkg/registry/discovery/endpointslice/strategy.go
```diff
@@ -96,3 +102,38 @@ func (endpointSliceStrategy) ValidateUpdate(ctx context.Context, new, old runtim
...
+
+// dropTerminatingConditionOnUpdate will drop the terminating condition field if the EndpointSliceTerminatingCondition
+// feature gate is disabled unless an existing EndpointSlice object has the field already set. This ensures
+// the field is not dropped on rollback.
+func dropTerminatingConditionOnUpdate(oldEPS, newEPS *discovery.EndpointSlice) {
+	if utilfeature.DefaultFeatureGate.Enabled(features.EndpointSliceTerminatingCondition) {
+		return
+	}
+
+	unsetTerminatingCondition := true
```
thockin, line 126: I feel like we discussed this before.  This is not exactly correct - if any old EP has it set, any new EP can use it.  Given the context, I think this is acceptable, but worth commenting.

thockin (commented):
I will leave Rob to review the controller parts.  API LGTM except some comment nits.

pkg/registry/discovery/endpointslice/strategy.go
andrewsykim: > I feel like we discussed this before. 

Yeah, admittedly this always confuses me though :P 

Added more comments to make this clearer. `dropTerminatingConditionOnUpdate` has some extra comments which should help too.

pkg/apis/discovery/types.go
```diff
@@ -116,6 +116,11 @@ type EndpointConditions struct {
 	// indicates an unknown state. In most cases consumers should interpret this
 	// unknown state as ready.
 	Ready *bool
+
+	// terminating indicates if this endpoint is terminating. A nil value indicates an
```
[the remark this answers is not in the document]
andrewsykim: updated

pkg/apis/discovery/types.go
```diff
@@ -116,6 +116,11 @@ type EndpointConditions struct {
...
+	// unknown state. Consumers should interpret this unknown state indicates the
```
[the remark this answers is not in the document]
andrewsykim: updated

pkg/controller/endpointslice/endpointslice_controller_test.go
andrewsykim: Last commit has refactors a bunch of the TestSyncService* tests into a single one with test tables. There's a few TestSyncService tests I didin't combine like `TestSyncServiceNoSelector` and `TestSyncServicePendingDeletion` since they are edge cases which would be difficult to fit into the generic testing table.

pkg/controller/endpointslice/endpointslice_controller_test.go
```diff
@@ -327,77 +330,494 @@ func TestOnEndpointSliceUpdate(t *testing.T) {
...
+						PodIPs: []v1.PodIP{{
+							IP: "10.0.0.1",
+						}},
+						Conditions: []v1.PodCondition{
+							{
+								Type:   v1.PodReady,
+								Status: v1.ConditionTrue,
+							},
+						},
+					},
```
robscott, line 728: I think it's easier to make sense of what's changing in tests like this if things that are consistent are mostly hidden. As an example, maybe you could use a `standardPodSpec` var since that seems to be irrelevant to all of these tests. Similarly, maybe it would help to add some kind of helper func for generating `ObjectMeta`. Maybe something like this:

```suggestion
					ObjectMeta: standardPodMeta("pod0", nil),
					Spec: standardPodSpec,
					Status: v1.PodStatus{
						PodIP: "10.0.0.1",
						PodIPs: []v1.PodIP{{
							IP: "10.0.0.1",
						}},
						Conditions: []v1.PodCondition{
							{
								Type:   v1.PodReady,
								Status: v1.ConditionTrue,
							},
						},
					},
```

robscott (commented):
Thanks for the work on this! It all LGTM, just had some nits around how tests could be further simplified.

pkg/controller/endpointslice/endpointslice_controller_test.go
```diff
@@ -327,77 +330,494 @@ func TestOnEndpointSliceUpdate(t *testing.T) {
...
+					Name:     utilpointer.StringPtr("udp-example"),
+					Protocol: protoPtr(v1.ProtocolUDP),
+					Port:     utilpointer.Int32Ptr(int32(161)),
+				},
+				{
+					Name:     utilpointer.StringPtr("tcp-example"),
+					Protocol: protoPtr(v1.ProtocolTCP),
+					Port:     utilpointer.Int32Ptr(int32(80)),
+				},
+			},
```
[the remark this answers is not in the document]
andrewsykim: I was thinking that we probably want to add more test cases where both the ServicePort and IPFamily in Service would vary and the ports in Pod would vary. This is why the entire Service and Pods are put into the test table.

pkg/controller/endpointslice/endpointslice_controller_test.go
andrewsykim: I was thinking we would add tests here in the future where the labels could vary between the pods which was why I didn't try to hide any of the specific fields. I think it's safe to remove TypeMeta though.

andrewsykim: For standardPodSpec, I don't think any of the unit tests here use named pod ports. If we added those the pod spec would change as well :thinking:

robscott: 👍 makes sense

pkg/controller/endpointslice/endpointslice_controller_test.go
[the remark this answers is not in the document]
robscott: 👍 makes sense

andrewsykim:
Part of https://github.com/kubernetes/kubernetes/issues/85643

Quentin-M:
@andrewsykim 

IHMO this concern is not really relevant compared to how significant of an impact the lack of proper production support for `externalTrafficPolicy=Local`. As it stands right now, companies running production services on AWS LBs with massive amounts of traffic, and low latency requirements, this is very problematic, as to simply avoid having downtime during roll-outs - `externalTrafficPolicy=Cluster` must be used, which makes traffic bounce node to node for no reason, saturating the NICs. For someone not checking the service endpoint statuses, I don't think that'd really change too much. 

Starting v1.19, with ELB/NLBs, thankfully, the issue may be worked around by using `service.beta.kubernetes.io/aws-load-balancer-target-node-labels` and topological routing. With alb ALB ingress controller, you are currently forced to build a custom target group, custom routing rules, and a second service with topological routing.

andrewsykim:
> IHMO this concern is not really relevant compared to how significant of an impact the lack of proper production support for externalTrafficPolicy=Local. As it stands right now, companies running production services on AWS LBs with massive amounts of traffic, and low latency requirements, this is very problematic, as to simply avoid having downtime during roll-outs - externalTrafficPolicy=Cluster must be used, which makes traffic bounce node to node for no reason, saturating the NICs. For someone not checking the service endpoint statuses, I don't think that'd really change too much.

I definitely understand this concern, and in fact this issue in particular is why we started this work initially (see https://github.com/kubernetes/kubernetes/issues/85643). But I disagree that this concern is not relevant based on the impact of this bug. Taking an ingress controller as an example, if we suddenly started to include terminating endpoints by default, we are introducing a new behavior that may now include not ready endpoints that are not available to receive traffic.

robscott:
This is an interesting one. It’s a little scary in that an upgrade on API/Controller side could potentially break earlier consumers. I think it’s reasonable to expect consumers to honor ready, but if endpoints that previously disappeared or appeared unready are going to appear ready with this approach, that seems problematic. Since there are cases where both `ready` and `terminating` would be set to true, we would end up exposing `ready` endpoints that previously would not have been published. 

Given that background, we'll need to be very careful rolling this out. We would need to guarantee compatibility with kube-proxy running 1 minor version earlier than API/Controller. In that case, we’d need to ensure that kube-proxy could handle/interpret terminating at least one minor release before it was set/published by the controller.

Perhaps more importantly, as you've already pointed out, we'll have to communicate this change broadly to implementations so they can also appropriately prepare for this change. With all of that said, this change does still feel worthwhile, we'll just have to spread it out over several releases with lots of communication along the way.

andrewsykim:
> Given that background, we'll need to be very careful rolling this out. We would need to guarantee compatibility with kube-proxy running 1 minor version earlier than API/Controller. In that case, we’d need to ensure that kube-proxy could handle/interpret terminating at least one minor release before it was set/published by the controller.

This makes sense, and per the KEP both endpointslice controller and kube-proxy will have an alpha feature gate for this behavior.

> Perhaps more importantly, as you've already pointed out, we'll have to communicate this change broadly to implementations so they can also appropriately prepare for this change. With all of that said, this change does still feel worthwhile, we'll just have to spread it out over several releases with lots of communication along the way.

Sgtm!

thockin:
Going back through this - perhaps we should add this field NOW (before EPSlice is GA) and assert that while it is always false today it will eventually be true for some endpoints, and so clients should handle it.

Then this KEP becomes about the act of setting the field to true.

I'm not sure this is significantly better, because of Beta users, thought I feel SLIGHTLY less bad about impacting them.

Honestly, the only REALLY compatible fix is to leave `ready` meaning what it means today (ready and not terminating) and consider it to be the "summary value".  Then add TWO new fields: `terminating` and `accepting` (names TBD).  Old clients stay safe, new clients can use the more granular fields.

Of course, if we ever want to add ANOTHER such, we would have to do it again (e.g. split accepting into 2 new fields).

We could make it self-describing:

```
readiness: [ { name: accepting, polarity: true }, { name: terminating, polarity: false } ]
conditions: [ { name: accepting, value: true }, name: terminating, value: true } ] # <-- NOT READY
```

Blech!  Other ideas?

robscott:
> Going back through this - perhaps we should add this field NOW (before EPSlice is GA) and assert that while it is always false today it will eventually be true for some endpoints, and so clients should handle it.

If we're planning on taking EPSlice to GA in 1.20 wouldn't it be too late to add a field to that API pre-GA?

Is there any way we could do a safe/gradual rollout over several minor releases to avoid adding another API field here? IE in the first release we update consumers to understand terminating if it is set (nil by default) and in a second release we update producers to actually set the field.

thockin:
>> Going back through this - perhaps we should add this field NOW (before EPSlice is GA) and assert that while it is always false today it will eventually be true for some endpoints, and so clients should handle it.
>
>If we're planning on taking EPSlice to GA in 1.20 wouldn't it be too late to add a field to that API pre-GA?

We *could* squeek it by - it will be, by definition, false.  That said, it's
bending the rules for a one-time win.  Anyone who has written code against the
beta API will be broken (silently) when this eventually becomes true.  You can
argue "well, they used the beta API and we documented the new field" but
that's mostly denying reality.

>Is there any way we could do a safe/gradual rollout over several minor releases to avoid adding another API field here? IE in the first release we update consumers to understand terminating if it is set (nil by default) and in a second release we update producers to actually set the field.

How are you going to find every consumer of a GA API in a system a widely used
as this?  As soon as this API hits GA, ready needs to mean what it means today.
Anything else is a breaking change. :(

thockin:
On the breakingness of this change specifically:

Today a consumer of EPSlice will see a pod in the list until it becomes
terminating, at which time it disappears from the list.

If we just added this field (changing "ready" meaning), a consumer
would pod in the list until it becomes terminated, at which time it
disappears from the list.

The important difference is that while it is terminating, the consumer can
at least TRY to gracefully wrap up.  After this change they don't get a
chance.  Yes, they can choose to learn about `terminating`, but we broke
them.  I tried to convince myself it wasn't so bad, but it is.

Sorry...

andrewsykim:
In https://github.com/kubernetes/kubernetes/pull/92968#issuecomment-678328208 I suggested maybe we have the "terminating" condition be opt-in by adding a field in Service like `spec.publishTerminating`, but that would be yet another dimension of complexity we add to Service. Do we think a new field in Service is worth while for this use-case?

thockin:
I don't think so - I honestly don't think 2 new bit is that bad, especially
since the write-cycles will be in sync - we are not tripling the potential
number of writes, just doubling.

pkg/apis/discovery/types.go
```diff
@@ -114,8 +114,21 @@ type EndpointConditions struct {
...
-	// unknown state as ready.
+	// unknown state as ready. For compatibility reasons, ready should never be
+	// "true" for terminating endpoints.
 	Ready *bool
+
+	// accepting is identical to ready except it is set regardless of the terminating
+	// state of endpoints. A ready endpoint that is terminating should set this condition
+	// to true.
+	// +optional
+	Accepting *bool
```
andrewsykim, line 125: Some alternative names for this field:
* Serving
* Receiving
* Healthy
* ???

thockin: I like Serving better than Accepting.  Healthy implies that we are actively health-checking (not yet!)

Open to other ideas...

thockin (commented):
API wise I am good with this. 

@robscott can you review the slice-controller logic and tests?

pkg/apis/discovery/types.go
```diff
@@ -114,8 +114,22 @@ type EndpointConditions struct {
...
 	// indicates an unknown state. In most cases consumers should interpret this
-	// unknown state as ready.
+	// unknown state as ready. For compatibility reasons, ready should never be
+	// "true" for terminating endpoints.
 	Ready *bool
+
+	// accepting is identical to ready except it is set regardless of the terminating
+	// state of endpoints. A ready endpoint that is terminating should set this condition
+	// to true. A nil vaue indicates an unknown state. Consumers should interepret this
+	// unknown state to mean the endpoint is accepting traffic.
```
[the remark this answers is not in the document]
robscott: I think it would be good to address some potential points of confusion here. I think we should update this to cover the following:
- This field may be set less often than `ready`. When it is not set, consumers should defer to `ready`. I'm not sure if that's accurate, just seems like it could be better than assuming `nil == true` here.
- It could be helpful to clarify that this should be set to true if `service.spec.publishNotReadyAddresses` is set to true. Maybe we don't actually want to carry that through to this field though? This could be our opportunity to drop that once and for all...

pkg/apis/discovery/types.go
robscott: Agree, I like `Serving` here.

pkg/apis/discovery/types.go
```diff
@@ -114,8 +114,22 @@ type EndpointConditions struct {
...
+	// +optional
+	Accepting *bool
+
+	// terminating indicates that this endpoint is terminating. A nil value indicates an
+	// unknown state. Consumers should interpret this unknown state to mean that the
+	// endpoint is not terminating.
```
robscott, line 130: Probably not a good idea, but could we update the `Terminating` condition to mean both `Terminating` and `Ready`? We would simply exclude endpoints that were both terminating and not ready in this list.

pkg/controller/endpointslice/utils.go
```diff
@@ -60,7 +62,11 @@ func podToEndpoint(pod *corev1.Pod, node *corev1.Node, service *corev1.Service,
 		}
 	}
 
-	ready := service.Spec.PublishNotReadyAddresses || podutil.IsPodReady(pod)
+	accepting := podutil.IsPodReady(pod)
+	terminating := pod.DeletionTimestamp != nil
+	// For compatibility reasons, "ready" should never be "true" if a pod is terminatng, unless
+	// publishNotReadyAddresses was set.
+	ready := service.Spec.PublishNotReadyAddresses || (accepting && !terminating)
```
robscott, line 69: It seems like we need to call out this difference in the docs for the accepting condition. I generally agree with dropping the `publishNotReadyAddresses` check here for `accepting`, but that may be unexpected for some.

pkg/registry/discovery/endpointslice/strategy_test.go
```diff
@@ -0,0 +1,448 @@
...
+
+	for _, testcase := range testcases {
+		t.Run(testcase.name, func(t *testing.T) {
+			defer featuregatetesting.SetFeatureGateDuringTest(t, utilfeature.DefaultFeatureGate, features.EndpointSliceTerminatingCondition, testcase.terminatingGateEnabled)()
+
+			dropDisabledConditionsOnCreate(testcase.eps)
+			if !apiequality.Semantic.DeepEqual(testcase.eps, testcase.expectedEPS) {
+				t.Logf("actual endpointslice: %v", testcase.eps)
+				t.Logf("expected endpointslice: %v", testcase.expectedEPS)
+				t.Errorf("unexpected EndpointSlice")
```
robscott, line 143: This output seems like it might be a bit confusing if it failed. Maybe a more descriptive error message could help.

pkg/registry/discovery/endpointslice/strategy_test.go
```diff
@@ -0,0 +1,448 @@
...
+		t.Run(testcase.name, func(t *testing.T) {
+			defer featuregatetesting.SetFeatureGateDuringTest(t, utilfeature.DefaultFeatureGate, features.EndpointSliceTerminatingCondition, testcase.terminatingGateEnabled)()
+
+			dropDisabledConditionsOnUpdate(testcase.oldEPS, testcase.newEPS)
+			if !apiequality.Semantic.DeepEqual(testcase.newEPS, testcase.expectedEPS) {
+				t.Logf("actual endpointslice: %v", testcase.newEPS)
+				t.Logf("expected endpointslice: %v", testcase.expectedEPS)
+				t.Errorf("unexpected EndpointSlice")
+			}
+		})
```
robscott, line 446: This output seems like it might be a bit confusing if it failed. Maybe a more descriptive error message could help.

robscott (commented):
Thanks for your work on this!

pkg/controller/endpointslice/utils.go
andrewsykim: > I generally agree with dropping the publishNotReadyAddresses check here for accepting, but that may be unexpected for some.

I was wondering about this too. When this gate goes beta and on by default, `publishNotReadyAddresses` wouldn't really apply anymore (at least with EndpointSlice) since all endpoints are included but with the new conditions. Is this technically a breaking change? I'm in favor of just dropping `publishNotReadyAddresses` in the context of EndpointSlice but I don't know if that's okay.

pkg/apis/discovery/types.go
andrewsykim: Thanks, will update the field to `serving`

pkg/apis/discovery/types.go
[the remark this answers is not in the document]
andrewsykim: > This field may be set less often than ready. When it is not set, consumers should defer to ready. I'm not sure if that's accurate, just seems like it could be better than assuming nil == true here.

Agreed that we should document users should check `Ready` if `Accepting is nil.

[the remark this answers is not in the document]
andrewsykim: > It could be helpful to clarify that this should be set to true if service.spec.publishNotReadyAddresses is set to true. Maybe we don't actually want to carry that through to this field though? This could be our opportunity to drop that once and for all...

I'm in favor of dropping `publishNotReadyAddresses` in the context of EndpointSlice. This behavior doesn't change during alpha though so I was hoping we can come to a conclusion on this before this feature goes Beta and all endpoints regardless of their condition is added to EndpointSlice.

pkg/controller/endpointslice/utils.go
robscott: It looks like the definition of `publishNotReadyAddresses` has [evolved a bit](https://github.com/kubernetes/kubernetes/pull/90783) since the last time I looked at it. Given the recent clarifications here, I think we probably need to continue to support this both for `ready` and `serving`.

andrewsykim: Ah, thanks for digging that up.

pkg/apis/discovery/types.go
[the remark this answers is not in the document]
thockin: I think it's fair to document that publishNotReadyAddresses is ignored for EndpointSlice

pkg/apis/discovery/types.go
thockin: If we did that, why do we have the Ready bit at all?

pkg/apis/discovery/types.go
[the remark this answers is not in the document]
thockin: I missed the later comment.  OK.  So:

if publishNotReadyAddresses - all endpoints are always Ready.  We could still make Serving and Terminating be sane?
else 
  - Ready = Serving && !Terminating
  - Serving means pod is Ready (terminating or not) 
  - Terminating means pod is deleting (ready or not)

Yes?

pkg/apis/discovery/types.go
robscott: I guess I was just trying to see if we could survive without the `Serving` bool. Here's how I understand the rationale for that field:

- Adding the Terminating condition means the endpoints will stick around longer than previously.
- To make that relatively safe, we need the existing `Ready` field to be false.
- Consumers want to know if an endpoint is `Terminating` but still able to handle traffic.
- Our solution to that was to add a `Serving` field that would be true even if an endpoint was `Terminating`.

My question is if we can have a single condition that indicates that an endpoint is terminating and can handle traffic. My argument is that consumers likely don't care about endpoints that are terminating and can't handle traffic so we could just drop them from the list. (I'm not actually sure on that part, maybe consumers do care about those endpoints).

This would be a pretty simple approach:

- `Ready` remains unchanged
- New `Terminating` condition to indicate that an endpoint is terminating but still able to handle traffic. 

Maybe the actual name of the condition might need to be clearer, but if we can safely say that no one cares about the `Terminating=true,Serving=false` condition, it seems unnecessary to add a second condition here.

andrewsykim: > My argument is that consumers likely don't care about endpoints that are terminating and can't handle traffic so we could just drop them from the list. (I'm not actually sure on that part, maybe consumers do care about those endpoints).

While I agree that generally we won't care too much about {!R,T} endpoints, I think it would still be useful to allow consumers to check for this state. In this thread https://github.com/kubernetes/enhancements/pull/1607#discussion_r412581130 I think we agreed that while it's not very common, it's may still be useful for consumers (e.g. kube-proxy) to track this state.

pkg/apis/discovery/types.go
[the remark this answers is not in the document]
andrewsykim: Yeah that makes sense.

andrewsykim:
@robscott comments addressed, PTAL

pkg/apis/discovery/types.go
thockin: Re-reading the discussion linked, we did agree that we should explore all 4 quadrants.

@robscott if `Ready` means "serving && !terminating" and `Terminating` means "serving && terminating", then `!Ready && !Terminating` means "!serving" but that's all we know for sure. There's no representation of "!serving && terminating", right? 

It seems maybe more useful to flatten all Terminating into one.  So `Ready` means "serving && !terminating" and `Terminating` means "terminating".  It seems to me (with no real data) that pods will catch SIGTERM and either stop accepting or keep accepting, but that all pods in a set will do the same thing.  Thus, it seems unlikely that we will have a mix of serving and !serving all terminating, or at least it won't last long.  But even as I type that out, it's all async, so we WILL have that mix, but I think it will be fleeting, and it will either settle at "serving && terminating" or move from towards "!serving && terminating".  It seems unlikelty to move from "!serving" -> "serving".

For the original `Local` use-case, does this matter?  If I have an ingress LB daemonset, the difference between serving and not serving isn't going to matter much (generally 1 instance). 

So I could be convinced to ignore serving within the set of terminating.  It's not what we said before. - is the one less bit and somewhat more cognitive load worhtwhile?

robscott: Thanks for the context here, it was helpful to read through the previous conversation around this. One of the points that stood out to me was:

> Sending traffic to any pod that fails readiness seems like we're breaking API contract. I'm personally in favor of only falling back to {R, T} endpoints

That makes it sound like there's no value in tracking endpoints that are `!Ready && Terminating`. As a variation of what Tim said above, maybe we could use 2 conditions that were defined as the following:

`ready`: `serving && !terminating`
`draining`: `serving && terminating`

Those conditions should never be true at the same time. I'm not sure that `draining` is the appropriate term here, I just wanted to throw out an alternative to `terminating` since that may be too broad for what I'm suggesting.

thockin:
So the implied third state here is `!Ready && !Draining` which can only
mean !serving, regardless of terminating or not.

This is a little different than the discussion, but I am not sure it is
meaningful.  Andrew?

pkg/apis/discovery/types.go
andrewsykim: Reposting Tim's comment https://github.com/kubernetes/kubernetes/pull/92968#issuecomment-720684922 cause email replies from Github doesn't seem to recognize threads:

> So the implied third state here is `!Ready && !Draining` which can only
mean !serving, regardless of terminating or not.

> This is a little different than the discussion, but I am not sure it is
meaningful.  Andrew?

andrewsykim: While I still think there's no real use-case yet for tracking {!R,T} endpoints, I don't think it hurts to leave room in the API to track it. If we went with `ready` and `draining` but in the future we wanted to support {!R, T}, we would have to add two additional conditions (total 4), to not breaking the existing meaning of `draining`.

I also think we want each condition to mean one thing and not have overlapping meaning with respect to terminating state. In an ideal scenario Ready refers to the pod's readiness and Terminating just refers to terminating, but because that's breaking I think the next best thing is to leave `Ready` as-is and add `serving` and `terminating`. Whether we actually account for {!R,T} in kube-proxy is a separate discussion I think.

andrewsykim:
Alternative naming proposal for conditions:
* `ready` = ready && !terminating
* `readiness` = ready (identical to `serving` but more explicitly tied to pod readiness, we would have to document why we have both a Ready and Readiness condition) 
* `terminating` = terminating

Kinda awkward to have a `ready` and `readiness` condition, but at least readiness is more explicit about a Pod's ready condition as opposed to `serving`. We would have to document really well why `readiness` exists and how it differs from `ready`. Thoughts @thockin @robscott?

pkg/apis/discovery/types.go
robscott: >  If we went with ready and draining but in the future we wanted to support {!R, T}, we would have to add two additional conditions (total 4), to not breaking the existing meaning of draining.

I think `draining` as proposed refers to a `{R, T}` state. If in the future we wanted to add a state for `{!R, T}`, I think that could be accomplished with one additional condition (`UnhealthyTerminating` or similar). That condition would also only be true if both `Ready` and `Draining` were false. I agree with you that a condition like that is probably not useful though.

It could actually be easy to explain (and understand) `ready` and `draining`, although this may be an oversimplification:

`ready`: An endpoint is ready for new connections and requests.
`draining`: An endpoint is closing out existing connections. 

On the other hand, I think a third `serving` condition may add unnecessary complexity at this point.

andrewsykim: > draining: An endpoint is closing out existing connections.

I think the misleading thing here is that an endpoint could be closing existing connections AND not be ready. In the context of the kube-proxy issue we're trying to address, `Draining` could mean two things:

1. an endpoint doesn't receive traffic and is closing out existing connections.
2. an endpoint is used as a fallback endpoint if no other endpoints exists.  

I think 2. will be difficult to articulate with only a `Draining` condition.

thockin: > no real use-case yet for tracking {!R,T} endpoints

I'd amend that - there's no real use-case for tracking {!R,T} distinct from {!R,!T}.  Rob's proposal lumps those together.  If we did need to distinguish between them, we could add an "Unready" or "Failing" condition (or a better name) that explicitly means {!R, !T} (IOW - should be ready but isn't).  That leaves the "no conditions" state as implicitly {!R, T}, though I think we'd want to just call that unknown.

We're dangerously close to defining a state machine, though.

thockin: Thinking out loud... I'll go out on a limb a bit - it can be "compatible" to split an existing set of endpoints by changing the meaning of conditions or adding new conditions as long as we believe the change is safe (won't cause things to break).

Simpy adding `terminating` is NOT safe because it first *expands* the set of endpoints to include dangerous ones (endpoints that are imminently going away) and then segments them.  As discussed elsewhere, any client who was looking just at `ready` would suddenly see more endpoints and not safe ones.

That said, once we have expanded the endpoints to "all", we can subset.  For example, if we wanted to distinguish pods that are !ready from pods that are just starting up, we could (in theory) track whether a pod has EVER been ready and, if not call it "starting", thereby subsetting the {!R} set.  Again, this sounds state-machiney, but it's just an example.  That would be a "safe" change because clients who were looking for !R would not be getting anything new.

I am trying to find other hypotheticals that would make this discussion more concrete.  Here's one.  Suppose we define a new pod-condition "flapping" - the pod is flipping ready/unready too often.

How would you think of that?

andrewsykim: I feel that `flapping` and `draining` are describing behaviors instead of conditions. Maybe they are the same things just at different levels of granularity.  A consumer of this API should be able to extract `flapping` by seeing the `serving` condition flip between `true` and `false` repeatedly. In the same way, `draining` should be extracted from checking `terminating` and `serving`. To me at least, EndpointSlice is the lowest denominator here (assuming it is too expensive to watch all pods) so it should try not to be prescriptive to behaviours and only describe states that are deterministic with well defined definitions.

robscott: > Draining could mean two things:
> 1. an endpoint doesn't receive traffic and is closing out existing connections.
> 2. an endpoint is used as a fallback endpoint if no other endpoints exists.

That's a fair point, maybe draining isn't the right word here. Maybe there's another word that can communicate the `{R,T}` state?

> there's no real use-case for tracking {!R,T} distinct from {!R,!T}

Good clarification, thanks!

> That leaves the "no conditions" state as implicitly {!R, T}, though I think we'd want to just call that unknown.

Wouldn't unknown/empty map more naturally to all conditions being false? 

> Suppose we define a new pod-condition "flapping" - the pod is flipping ready/unready too often.

I'd imagine that we would set `ready` to false and optionally add a new `flapping` condition that would also only be true if every other condition for the endpoint was false.

robscott: > I feel that flapping and draining are describing behaviors instead of conditions. 

I'm open to different names here, I just want to avoid adding two new conditions when we only actually care about one very specific combination `{R, T}`. What if we need to add some additional condition in the future that relies on pod readiness being expanded or contracted in some way? Would we need a new condition like `serving` for that expansion as well?

andrewsykim: Really on the fence about this one :thinking: 

While I do think a `serving` condition (or similar) is generically useful to consumers, I agree that having to add two new conditions is not ideal. My gut tells me that consumers would want a way to track pod readiness independent of anything else but I don't really have good evidence or use-cases to back this up. Since we are going to alpha gate this for one release anyways I think I'm okay to stick with one condition for now since it'll be easier to transition to two conditions if we change our minds later.

Regarding naming of the condition, I thought about both `draining` and `cordoning` but both I feel are inaccurate because they're prescriptive to the behavior consumers _should_ be doing as opposed to describing the endpoint status and letting consumers do what they wish with it. I'm thinking the best name for this condition is just `readyTerminating`. Thoughts?

andrewsykim:
Going to update the condition name to `readyTerminating` based on my comment here https://github.com/kubernetes/kubernetes/pull/92968#discussion_r517406788, open to other naming suggestions though.

andrewsykim:
@thockin @robscott collapsed `serving` and `terminating` into a single conditoin `readyTerminating`, PTAL

pkg/apis/discovery/types.go
robscott: I'm not 100% sold on that name, but I don't hate it. I think `readyTerminating` makes sense if EndpointSlices were only ever going to be used to reference Pods. Since we want this to be a more generic API, I think a more generic term here might be helpful. I still don't mind `draining` but can see the hesitation there as it could be seen as too prescriptive. Naming things is hard.

thockin: >  Since we are going to alpha gate this for one release 

At least.  If we get it wrong it could be several.

I really don't like the evolution towards N mutually-exclusive state-conditions.  I think the right answer at this layer is orthogonal.

What about "up" and "terminating" if we don't like Serving?

andrewsykim: > I really don't like the evolution towards N mutually-exclusive state-conditions. I think the right answer at this layer is orthogonal.

I agree.

> What about "up" and "terminating" if we don't like Serving?

I also suggested here https://github.com/kubernetes/kubernetes/pull/92968#issuecomment-720710258 to use `ready` and `readiness`. But I think @robscott's concern is less about naming (up vs serving) but more so that we are adding two conditions for one use-case.

thockin (commented):
I'll leave the EPSlice logic and final LGTM to @robscott but API-wise I am OK with this.

/approve

robscott:
Thanks!

/lgtm

PR #92979, opened 2020-07-11
Relax validation for dnsConfig searches

### LINKED ISSUE #92889: dnsConfig applies RFC-1123 too broadly to searches

RFC-1123 defines restrictions to host names. In Kubernetes, these restrictions are also
applied to object names.

However, for search strings in dnsConfig, names that are perfectly valid according
to [RFC 2181 Section 11](https://tools.ietf.org/html/rfc2181#section-11) are rejected.

An example of this is using underscores in a subdomain:

```yaml
apiVersion: v1
kind: Pod
metadata:
  namespace: default
  name: dns-example
spec:
  containers:
    - name: test
      image: nginx
  dnsPolicy: "None"
  dnsConfig:
    nameservers:
      - 1.2.3.4
    searches:
      - abc_d.example.com
```

Since this is specifying a DNS search string specifically and not the name of an object
in Kubernetes, it seems like this check could be relaxed to allow names that comply with RFC 2181.

Instead this gives the following error:
```
The Pod "dns-example" is invalid: spec.dnsConfig.searches[0]: Invalid value: "abc_d.example.com": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

**Why is this needed**:

Workloads that rely on short names can take time to change, which is why the dnsConfig is useful. However, in our case most of the names we need in the search have an underscore. Our current work-around is to re-write resolv.conf from inside the pod.

sethev:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Fixes #92889

```release-note
Relax validation rules for dnsConfig searches to allow valid DNS names
```

```docs

```

### DIFF
--- a/pkg/apis/core/validation/validation.go
+++ b/pkg/apis/core/validation/validation.go
@@ -31,6 +31,7 @@ import (
 
 	"k8s.io/klog/v2"
 
+	"github.com/miekg/dns"
 	v1 "k8s.io/api/core/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -2930,6 +2931,17 @@ func validateReadinessGates(readinessGates []core.PodReadinessGate, fldPath *fie
 	return allErrs
 }
 
+// Validate domain name, note this is a looser check than ValidateDNS1123Subdomain
+// which does ensure valid DNS names but is more restrictive than DNS.
+func validateDomainName(value string, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	_, d := dns.IsDomainName(value)
+	if !d {
+		allErrs = append(allErrs, field.Invalid(fldPath, value, fmt.Sprintf("not a valid domain name: %s", value)))
+	}
+	return allErrs
+}
+
 func validatePodDNSConfig(dnsConfig *core.PodDNSConfig, dnsPolicy *core.DNSPolicy, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
@@ -2966,7 +2978,7 @@ func validatePodDNSConfig(dnsConfig *core.PodDNSConfig, dnsPolicy *core.DNSPolic
 			if strings.HasSuffix(search, ".") {
 				search = search[0 : len(search)-1]
 			}
-			allErrs = append(allErrs, ValidateDNS1123Subdomain(search, fldPath.Child("searches").Index(i))...)
+			allErrs = append(allErrs, validateDomainName(search, fldPath.Child("searches").Index(i))...)
 		}
 		// Validate options.
 		for i, option := range dnsConfig.Options {
--- a/pkg/apis/core/validation/validation_test.go
+++ b/pkg/apis/core/validation/validation_test.go
@@ -6535,9 +6535,12 @@ func TestValidatePodDNSConfig(t *testing.T) {
 			expectedError: true,
 		},
 		{
-			desc: "invalid search path",
+			desc: "invalid search path: more than 63 characters in DNS label",
 			dnsConfig: &core.PodDNSConfig{
-				Searches: []string{"custom?"},
+				Searches: []string{
+					generateTestSearchPathFunc(2),
+					generateTestSearchPathFunc(64),
+				},
 			},
 			expectedError: true,
 		},
--- a/pkg/apis/core/validation/BUILD
+++ b/pkg/apis/core/validation/BUILD
@@ -41,6 +41,7 @@ go_library(
         "//staging/src/k8s.io/apimachinery/pkg/util/validation:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/util/validation/field:go_default_library",
         "//staging/src/k8s.io/apiserver/pkg/util/feature:go_default_library",
+        "//vendor/github.com/miekg/dns:go_default_library",
         "//vendor/k8s.io/klog/v2:go_default_library",
         "//vendor/k8s.io/utils/net:go_default_library",
     ],


sethev:
Signed the CLA

sethev:
Linked email to account

sethev:
@aojea mind reviewing? Fixes #92889

aojea:
hi @sethev , for new features there is another step, you have to "convince" @thockin first 😄

sethev:
> hi @sethev , for new features there is another step, you have to "convince" @thockin first 😄

_begins sweating profusely_

pkg/apis/core/validation/validation_test.go
```diff
@@ -6535,9 +6535,12 @@ func TestValidatePodDNSConfig(t *testing.T) {
 			expectedError: true,
 		},
 		{
-			desc: "invalid search path",
+			desc: "invalid search path: more than 63 characters in DNS label",
 			dnsConfig: &core.PodDNSConfig{
-				Searches: []string{"custom?"},
```
sethev, line 6540: `custom?` is technically valid in DNS. This test was introduced along with all of the other checks on dnsConfig by af7208047a39694f7015c715466ada48fda1312c, so i think it was an example that would fail the previous check but wasn't added because of a problem. `custom?` is the only test that failed with the new checks.

sethev:
@aojea is there any follow-up needed from me on this?

aojea:
> @aojea is there any follow-up needed from me on this?

you can bring it up in the next sig-network meeting, 
is the end of the release and most people use to be very busy during this period :shrug:

aojea:
> //pkg/master:go_default_test                                    FAILED TO BUILD

you have to run `hack\update-bazel.sh`

sethev:
@aojea thanks for the assist, and apologies for the notification spam. This appears to be passing now.

aojea:
As explained in the issue https://github.com/kubernetes/kubernetes/issues/92889#issuecomment-656679600 , technically this looks correct, however, is not the first time that some things are the way it is because of legacy things, that's we need to defer to Tim :smile: 

/assign @thockin

thockin:
This really wants a KEP.

pkg/apis/core/validation/validation.go
```diff
@@ -2930,6 +2931,17 @@ func validateReadinessGates(readinessGates []core.PodReadinessGate, fldPath *fie
 	return allErrs
 }
 
+// Validate domain name, note this is a looser check than ValidateDNS1123Subdomain
+// which does ensure valid DNS names but is more restrictive than DNS.
+func validateDomainName(value string, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	_, d := dns.IsDomainName(value)
+	if !d {
+		allErrs = append(allErrs, field.Invalid(fldPath, value, fmt.Sprintf("not a valid domain name: %s", value)))
```
thockin, line 2940: For the sake of errors, I think it would be easier to write the validation ourselves.

test length
split on "."
test each length

thockin, line 2940: error messages should be of the form "must be" or "may not be" or similar.

pkg/apis/core/validation/validation.go
```diff
@@ -2966,7 +2978,7 @@ func validatePodDNSConfig(dnsConfig *core.PodDNSConfig, dnsPolicy *core.DNSPolic
 			if strings.HasSuffix(search, ".") {
 				search = search[0 : len(search)-1]
 			}
-			allErrs = append(allErrs, ValidateDNS1123Subdomain(search, fldPath.Child("searches").Index(i))...)
+			allErrs = append(allErrs, validateDomainName(search, fldPath.Child("searches").Index(i))...)
```
thockin, line 2981: We need to distinguish updates of existing values from net-new usage.  For example, if I upgrade my cluster, create a deployment using a previously-invalid value, then downgrade, I should still be able to update my deployment.  But I should not be able to make a new deployment using invalid values unless the gate is enabled.

pkg/apis/core/validation/validation.go
```diff
@@ -2930,6 +2931,17 @@ func validateReadinessGates(readinessGates []core.PodReadinessGate, fldPath *fie
 	return allErrs
 }
 
+// Validate domain name, note this is a looser check than ValidateDNS1123Subdomain
+// which does ensure valid DNS names but is more restrictive than DNS.
+func validateDomainName(value string, fldPath *field.Path) field.ErrorList {
```
thockin, line 2936: this could use a test of its own

thockin (commented):
This needs a feature gate

pkg/apis/core/validation/validation.go
aojea: @thockin  slightly unrelated to the PR, just for learning purposes, how does this work for skewed apiservers?
apiserver A with the feature gate enabled and new version will allow to use the new search domain "my-domain.local", but clients using apiserver B with version-1 will fail, right?

aojea: I've found the answer, https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#changing-the-api
just for reference, 

> The recommended place to do this is in the REST storage strategy's PrepareForCreate/PrepareForUpdate methods:

thockin:
Hi @sethev - is this PR still something you want to pursue?

thockin:
Reopen if we are going to pursue this.

PR #92981, opened 2020-07-11
[staging/client-go] fetchGroupVersionResources: remove unnecessary lock

gaurav1086:

**What type of PR is this?**
>
> /kind api-change
/kind bug
/kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Since the key groupVersion is unique for every goroutine, the operation `failedGroups[groupVersion] = err` is mutually exclusive, therefore there cannot be a data race condition here. So we do not need to lock around map.

Fixes #
N/A
N/A

```release-note
NONE
```

```docs
NONE
```


liggitt:
Maps are not safe to concurrently write to and read from, even if the key being written does not overlap. 

/close

PR #93000, opened 2020-07-12
Skip arch dependent kubectl test for non AMD64 nodes

hakman:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation

/kind failing-test

> /kind feature
> /kind flake

The `[sig-cli] Kubectl client Simple pod [It] should handle in-cluster config` test fails on ARM64 setups because it assumes it copies and uses the `kubectl` binary from the host from which tests are run.
https://testgrid.k8s.io/kops-misc#kops-aws-arm64-latest

In most situations the host running the tests is AMD64 based, so it will fail with something like this:
```
<exec.CodeExitError>: {
        Err: {
            s: "error running /workspace/kubernetes/platforms/linux/amd64/kubectl --server=https://api.e2e-kops-aws-arm64-latest.test-cncf-aws.k8s.io --kubeconfig=/tmp/kops584361883/kubeconfig exec --namespace=kubectl-6690 httpd -- /bin/sh -x -c /tmp/kubectl get pods --v=6 2>&1:\nCommand stdout:\n/tmp/kubectl: line 1: syntax error: unterminated quoted string\n\nstderr:\n+ /tmp/kubectl get pods '--v=6'\ncommand terminated with exit code 2\n\nerror:\nexit status 2",
        },
        Code: 2,
    }
```

n/a

n/a

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/framework/skipper/skipper.go
+++ b/test/e2e/framework/skipper/skipper.go
@@ -213,6 +213,13 @@ func SkipUnlessNodeOSDistroIs(supportedNodeOsDistros ...string) {
 	}
 }
 
+// SkipUnlessNodeOSArchIs skips if the node OS distro is not included in the supportedNodeOsArchs.
+func SkipUnlessNodeOSArchIs(supportedNodeOsArchs ...string) {
+	if !framework.NodeOSArchIs(supportedNodeOsArchs...) {
+		skipInternalf(1, "Only supported for node OS arch %v (not %s)", supportedNodeOsArchs, framework.TestContext.NodeOSArch)
+	}
+}
+
 // SkipIfNodeOSDistroIs skips if the node OS distro is included in the unsupportedNodeOsDistros.
 func SkipIfNodeOSDistroIs(unsupportedNodeOsDistros ...string) {
 	if framework.NodeOSDistroIs(unsupportedNodeOsDistros...) {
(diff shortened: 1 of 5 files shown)


test/e2e/framework/skipper/skipper.go
```diff
@@ -213,6 +213,20 @@ func SkipUnlessNodeOSDistroIs(supportedNodeOsDistros ...string) {
 	}
 }
 
+// SkipUnlessMasterOSArchIs skips if the master OS distro is not included in the supportedMasterOsArchs.
+func SkipUnlessMasterOSArchIs(supportedMasterOsArchs ...string) {
```
oomichi, line 217: This SkipUnlessMasterOSArchIs() is not used in this PR.
How about just implementing SkipUnlessNodeOSArchIs() only without SkipUnlessMasterOSArchIs()?
(I know this can be implemented for consistency like the other functions)

Now we are trying to avoid using code related to master node thing like https://github.com/kubernetes/kubernetes/issues/80177
So if this is not mandatory, it is nice to skip implementing.

oomichi (commented):
/cc @oomichi

test/e2e/framework/skipper/skipper.go
hakman: I have no objection. I actually think at some point some of these things could be read from node labels.

hakman: Done.

spiffxp (approved):
/milestone v1.19
/approve
/lgtm

hakman:
Thanks @nikhita :)

PR #93006, opened 2020-07-13
fix func addAllEventHandlers definition

lixiaobing1:
**What type of PR is this?**
/kind cleanup

func addAllEventHandlers defined as follows:

func addAllEventHandlers(
	sched *Scheduler,
	informerFactory informers.SharedInformerFactory,
	podInformer coreinformers.PodInformer,
) {

which only used one place:
	addAllEventHandlers(sched, informerFactory, podInformer)

so better to define addAllEventHandlers as:

 func addAllEventHandlers(
	sched *Scheduler,
	informerFactory informers.SharedInformerFactory,
	podInformer coreinformers.PodInformer) {

```
    fix func addAllEventHandlers definition

```

### DIFF
--- a/pkg/scheduler/eventhandlers.go
+++ b/pkg/scheduler/eventhandlers.go
@@ -362,8 +362,7 @@ func (sched *Scheduler) skipPodUpdate(pod *v1.Pod) bool {
 func addAllEventHandlers(
 	sched *Scheduler,
 	informerFactory informers.SharedInformerFactory,
-	podInformer coreinformers.PodInformer,
-) {
+	podInformer coreinformers.PodInformer) {
 	// scheduled pod cache
 	podInformer.Informer().AddEventHandler(
 		cache.FilteringResourceEventHandler{


pkg/scheduler/eventhandlers.go
```diff
@@ -362,8 +362,7 @@ func (sched *Scheduler) skipPodUpdate(pod *v1.Pod) bool {
 func addAllEventHandlers(
 	sched *Scheduler,
 	informerFactory informers.SharedInformerFactory,
-	podInformer coreinformers.PodInformer,
-) {
+	podInformer coreinformers.PodInformer) {
```
chendave, line 365: either one should be good.

chendave (commented):
/kind cleanup

pkg/scheduler/eventhandlers.go
lixiaobing1: yeah, but I think it's easy to make a mistake :)

Huang-Wei: I don't think the proposed change is always better than the current style. For instance, with the current style, you can delete a parameter by simply deleting that whole line; while with the proposed change, if you want to delete the last parameter, you have to manually move `){` one line above.

PR #93008, opened 2020-07-13
[staging/apiextensions] StartTestServer: fix goroutine leak

gaurav1086:

**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR fixes the goroutine leak possibly caused since its trying to write to the errCh and if the main routine returns before reading from the channel errCh: 

```
    client, err := kubernetes.NewForConfig(server.GenericAPIServer.LoopbackClientConfig)
    if err != nil {
        return result, fmt.Errorf("failed to create a client: %v", err)
    }  
```

Fixes #
N/A
N/A

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/staging/src/k8s.io/apiextensions-apiserver/pkg/cmd/server/testing/testserver.go
+++ b/staging/src/k8s.io/apiextensions-apiserver/pkg/cmd/server/testing/testserver.go
@@ -151,7 +151,7 @@ func StartTestServer(t Logger, instanceOptions *TestServerInstanceOptions, custo
 		return result, fmt.Errorf("failed to create server: %v", err)
 	}
 
-	errCh := make(chan error)
+	errCh := make(chan error, 1)
 	go func(stopCh <-chan struct{}) {
 		if err := server.GenericAPIServer.PrepareRun().Run(stopCh); err != nil {
 			errCh <- err


gaurav1086:
cc @sttts @liggitt

gaurav1086:
cc @lavalamp

lavalamp:
Do you have evidence that this goroutine actually leaks, or is it just theoretical?

*Please don't fix goroutine leaks by inserting a `, 1`*. This style of fix always turns a goroutine leak into an error (or other) value that is never examined. The goroutine leak is at least visible. The unexamined error is not visible. So, this style of fix makes things worse. There's (thus far in my experience) always a better way. I'm sorry to be a little blunt here. But I have seen a number of these changes from you, and even though I feel like I say some variation of this in every one, I guess I haven't been clear enough: please don't fix things like this, even in first drafts.

gaurav1086:
@lavalamp thanks for the pointers. However this an old PR. I will close this one and re-open another one with an alternative solution. Thanks.

fedebongio:
@gaurav1086 did you reopen this on purpose? I thought it was closed in the past. Could you clarify? 
Thanks!

gaurav1086:
@fedebongio Reopened this mistake. Closing.

PR #93021, opened 2020-07-13
integration test for dynamic certificate reload in the kube-aggregator

p0lyn0mial:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**: is simply adds a missing integration test for dynamic certificate reload in the kube-aggregator

Fixes #

```release-note
NONE
```

```docs

```


test/integration/apiserver/certreload/aggregator_certreload_test.go
```diff
@@ -0,0 +1,341 @@
...
+		t.Fatal(err)
+	}
+	defer cleanUpFn()
+
+	t.Log("STEP 3: Registering the Aggregated API Server and waiting for it to show up in the discovery")
+	if err := registerAggregatedAPIAndWaitForDiscovery(t, kubeClient, aggregatorClient, aggregatorIP, aggregatorPort); err != nil {
+		t.Fatal(err)
+	}
+
+	t.Log("STEP 4: Swapping the proxy certificate and trying to connect to the aggregated API")
```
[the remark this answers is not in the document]
p0lyn0mial: I prefer to check if client->kube->aggregator can talk to each other before swapping the certificates.

test/integration/apiserver/certreload/aggregator_certreload_test.go
```diff
@@ -0,0 +1,341 @@
...
+
+func TestAggregatorCertReload(t *testing.T) {
+	proxyCA, proxyClientKey, proxyClientCert, err := newCAWithClientCert()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var proxyClientKeyPath string
+	var proxyClientCertPath string
+	var servingCertPath string
```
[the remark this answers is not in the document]
p0lyn0mial: I'm initializing proxyClientKeyPath and proxyClientCertPath at the beginning of the test.

I'm initializing servingCertPath inside ModifyServerRunOptions because I don't have access to it. It is populated in kastesting.StartTestServerOrDie function

test/integration/apiserver/certreload/aggregator_certreload_test.go
```diff
@@ -0,0 +1,341 @@
...
+	if err := ioutil.WriteFile(proxyClientCertPath, testutil.EncodeCertPEM(proxyClientCert), 0644); err != nil {
+		return err
+	}
+
+	if err := ioutil.WriteFile(proxyClientKeyPath, proxyClientKey, 0644); err != nil {
+		return err
+	}
+
+	// give it a time to reload the cert
+	time.Sleep(2 * time.Second)
```
alenkacz, line 272: should you rather poll multiple times and wait for the expected result than just wait? From my experience, this usually leads to flaky tests. I know that the reload time is 1s so this SHOULD be enough but I would still rather retry and wait for couple of seconds than this.

p0lyn0mial: sure, good idea. I'll update the PR.

cmd/kube-apiserver/app/testing/testserver.go
```diff
@@ -52,6 +52,9 @@ type TestServerInstanceOptions struct {
 
 	// Enable cert-auth for the kube-apiserver
 	EnableCertAuth bool
+
+	// ModifyServerRunOptions allows for providing a custom function to modify ServerRunOption just before running the server
+	ModifyServerRunOptions func(opts *options.ServerRunOptions)
```
sttts, line 57: I don't think we want this. The whole testing apiserver is meant to be driven through flags. There should not be a generic tweak func.

test/integration/apiserver/certreload/aggregator_certreload_test.go
```diff
@@ -0,0 +1,341 @@
...
+		}
+
+		proxyClientCertPath = filepath.Join(servingCertPath, "proxy-client.pem")
+		if err := ioutil.WriteFile(proxyClientCertPath, testutil.EncodeCertPEM(proxyClientCert), 0644); err != nil {
+			t.Fatal(err)
+		}
+
+		opts.ProxyClientCertFile = proxyClientCertPath
+		opts.ProxyClientKeyFile = proxyClientKeyPath
+		opts.EnableAggregatorRouting = true
```
sttts, line 81: what of these tweaks cannot be done with flags?

test/integration/apiserver/util.go
```diff
@@ -0,0 +1,146 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apiserver/pkg/server/dynamiccertificates"
+	client "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/client-go/util/cert"
+)
+
+func WaitForWardleRunning(t *testing.T, wardleToKASKubeConfig *rest.Config, wardleCertDir string, wardleIP string, wardlePort int) (*rest.Config, error) {
```
sttts, line 37: these are 1:1 copies? Please move them to its own commit. This is hard to review and trace later.

test/integration/apiserver/util.go
```diff
@@ -0,0 +1,146 @@
...
+	adminKubeConfig := createKubeConfig(wardleToKASKubeClientConfig)
+	wardleToKASKubeConfigFile, _ := ioutil.TempFile("", "")
+	if err := clientcmd.WriteToFile(*adminKubeConfig, wardleToKASKubeConfigFile.Name()); err != nil {
+		t.Fatal(err)
+	}
+
+	return wardleToKASKubeConfigFile.Name()
+}
+
+func createKubeConfig(clientCfg *rest.Config) *clientcmdapi.Config {
```
sttts, line 111: is this generic? Doesn't look like. Needs a less generic name.

cmd/kube-apiserver/app/testing/testserver.go
zerodayz: I am going to remove this, as you mentioned we want to drive apiserver through flags.

test/integration/apiserver/certreload/aggregator_certreload_test.go
zerodayz: Removing instead we can use flags to apiserver:

--proxy-client-key-file string
--proxy-client-cert-file string
--enable-aggregator-routing

zerodayz: @sttts I have rewritten the code by using flags but now trying to figure out how to pass the cert-file and key-file files to the testServer when the file paths are by default created inside StartTestServer and we generate cert as part of ModifyServerRunOptions.

Also you have suggested to remove: ModifyServerRunOptions, how do we pass `dynamiccertificates.FileRefreshDuration = 1 * time.Second `as a kube-apiserver argument?

Thank you for your help, I ll also check with @p0lyn0mial  and see what would be the best to do.

sttts: We usually pre-create certs in tests because generation costs entropy and might block or slow down tests. Does that help with the directory problem?

sttts: We definitely don't want a generic and powerful modifier like ModifyServerRunOptions. This will encourage misuse. If there is something super internal that is not exposed via flags (like maybe fresh duration), make it an explicit field  of TestServerInstanceOptions.

zerodayz: > We usually pre-create certs in tests because generation costs entropy and might block or slow down tests. Does that help with the directory problem?

Thank you! I was not aware, meaning I can use key and crt in  testdata folder: https://github.com/kubernetes/kubernetes/tree/master/cmd/kube-apiserver/app/testing/testdata

zerodayz: @sttts I have pre-generated the certs, do you think it makes sense to expose testing.pkgPath? I want to use that one to locate testdata, rather than using my own.

zerodayz: Please review new changes in this commit https://github.com/kubernetes/kubernetes/pull/93021/commits/d30d3903741bf0832d0bf6ca10def65f15b5dcf9

I have pre-generated the certs and removed the ModifyServerRunOptions

cmd/kube-apiserver/app/testing/testserver.go
zerodayz: Please review new changes in this commit https://github.com/kubernetes/kubernetes/pull/93021/commits/d30d3903741bf0832d0bf6ca10def65f15b5dcf9

I have pre-generated the certs and removed the ModifyServerRunOptions

enj:
/remove-sig auth

SIG API machinery owns the aggregator code.  @deads2k PTAL.

dims:
Is this PR still needed, please rebase if so (or we can close it?)

PR #93027, opened 2020-07-13
Add metav1.SetMetaDataLabel func

ialidzhikov:
**What type of PR is this?**
/kind feature

`k8s.io/apimachinery/pkg/apis/meta/v1` has already utility funcs such as `HasAnnotation` and `SetMetaDataAnnotation`. This PR also adds to the same package the funcs `HasLabel` and `SetMetaDataLabel`. My idea is to prevent boilerplate code on user side like:

```golang
	if obj.Labels == nil {
		obj.Labels = make(map[string]string)
	}
	obj.Labels[label] = value
```

for setting label on `ObjectMeta`.
I believe that the same boilerplate exist also in the k/k code base and the newly added funcs can be useful also there.

Fixes #

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/helpers.go
+++ b/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/helpers.go
@@ -201,6 +201,20 @@ func SetMetaDataAnnotation(obj *ObjectMeta, ann string, value string) {
 	obj.Annotations[ann] = value
 }
 
+// HasLabel returns a bool if passed in label exists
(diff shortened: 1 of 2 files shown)


therc:
> I believe that the same boilerplate exist also in the k/k code base and the newly added funcs can be useful also there.

Have you looked into how prevalent this might be in the k/k code, actually?

liggitt:
/lgtm
/approve
/ok-to-test
for symmetry with HasAnnotation/SetMetaDataAnnotation

Issue #85353, opened 2019-11-15
EndpointSliceController does not react to updated or deleted EndpointsSlice objects


liggitt:
**What happened**:
1. Create a service+pod
2. Observe an endpointslice get created with a `endpointslice.kubernetes.io/managed-by=endpointslice-controller.k8s.io` label (as expected
3. Manually modify the endpointslice to change the data, it is not corrected automatically
4. Delete the endpoint slice, it is not recreated automatically
5. Recreate the endpoint slice with a `endpointslice.kubernetes.io/managed-by=custom` label, it is not populated automatically (which is correct)
6. Change the label value to endpointslice-controller.k8s.io, it is still not populated automatically
7. Once I modify the service, the endpoint slice is updated

**What you expected to happen**:

The endpointsslice controller to react to deletions/updates of endpointsslice objects

cc @robscott @thockin

liggitt:
adding to the milestone until @thockin / @robscott triage this

athenabot:
/triage unresolved

Comment `/remove-triage unresolved` when the issue is assessed and confirmed.

🤖 I am a bot run by vllry. 👩‍🔬

thockin:
I don't think these need to be addressed for beta.  They are not worse than existing Endpoints.  Whether we should actually fix those or not is a good question, but I think it's OK to answer that post-beta.

As to the other issues - we need to build confidence in upgrades ASAP and make a call - we can flip back to default-off if we need to.

liggitt:
I do think it's reasonable to expect the controller to react to ownership changes or deletion, but agree this is roughly equivalent to what endpoints do. Lowering severity and moving to 1.18

liggitt:
Actually, thinking through this more, I think there's a significant issue here. The endpointslice controller differs from the endpoint controller in that endpointslices and services are not 1:1. 

The endpoints controller can deterministically create/update/delete the single correct endpoints object in response to every service change. 

The endpointsslice controller has to look up the endpointslices it should reconcile with a label based query. That means that if the cache is not up to date (as it would not be if the same service was synced rapidly), it could miss cleaning up a not-yet-observed endpointslice when reconciling a service update that removed an IP. That would leave a stale endpointsservice present until something else on the service or a pod triggered a reconcile after the cache was up to date. 

/priority important-soon

athenabot:
@robscott
If this issue has been triaged, please comment `/remove-triage unresolved`.

[...]

PR #93030, opened 2020-07-13
Requeue Service after Endpoint Deletion

### LINKED ISSUE #47711: endpoints not showing up for service
**Is this a request for help?** no.

**What keywords did you search in Kubernetes issues before filing this one?** (If you have found any duplicates, you should instead reply there.): service endpoints missing

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): Bug Report

**Kubernetes version** (use `kubectl version`): 1.6.2 

**Environment**:
- **Cloud provider or hardware configuration**: AWS
- **OS** (e.g. from /etc/os-release): 

```
NAME="Container Linux by CoreOS"
ID=coreos
VERSION=1353.8.0
VERSION_ID=1353.8.0
BUILD_ID=2017-05-30-2322
PRETTY_NAME="Container Linux by CoreOS 1353.8.0 (Ladybug)"
ANSI_COLOR="38;5;75"
HOME_URL="https://coreos.com/"
BUG_REPORT_URL="https://issues.coreos.com"
```

- **Kernel** (e.g. `uname -a`): `4.9.24-coreos`
- **Install tools**: manual
- **Others**:

**What happened**:

Service in `kube-system` namespace sees no endpoints that are available

**What you expected to happen**: For it to see available endpoints

**How to reproduce it** (as minimally and precisely as possible):

Not sure I can. It appeared on the latest remove/replace of worker nodes, but was fine before

I have a traefik Ingress controller defined and a service for it:

```
$ kubectl describe service traefik-ingress-controller -n kube-system
Name:			traefik-ingress-controller
Namespace:		kube-system
Labels:			app=traefik-ingress-controller
Annotations:		kubectl.kubernetes.io/last-applied-configuration={"apiVersion":"v1","kind":"Service","metadata":{"annotations":{},"labels":{"app":"traefik-ingress-controller"},"name":"traefik-ingress-controller","nam...
Selector:		name=traefik-ingress-lb
Type:			NodePort
IP:			10.200.177.150
Port:			http	80/TCP
NodePort:		http	30080/TCP
Endpoints:
Port:			https	443/TCP
NodePort:		https	30443/TCP
Endpoints:
Session Affinity:	None
Events:			<none>

$ kubectl -n kube-system get pod --selector=name=traefik-ingress-lb
NAME                               READY     STATUS    RESTARTS   AGE
traefik-ingress-controller-8vlms   1/1       Running   0          1h
traefik-ingress-controller-c81cq   1/1       Running   0          1h
traefik-ingress-controller-s51mq   1/1       Running   0          59m
```

Selectors match, the pods are running just fine, yet the endpoints do not show up.

I suspect if I delete the service and recreate it, it probably will pick it up, but I shouldn't have to, and am concerned it might occur again; rather actually debug it.

How can I check what kube is doing to affiliate it? How often does it check? Where?

swetharepakula:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

When the same service is deleted and quickly recreated, there is a race of whether the endpoints cache is up to date when processing the service creation causing the endpoints controller to wrongly see deleted endpoints as existing. This causes the endpoints to never be recreated.

This PR attempts to fix that race by adding a tracker similar to one in endpoints slice. The endpoints tracker will keep track of what the endpoints controller expects the state is. If any endpoint add, update, or deletion occurs that does not match the tracker's state, the associated service will be reprocessed.

Fixes #47711

```release-note
Endpoint controller requeues service after an endpoint deletion event occurs to confirm that deleted endpoints are undesired to mitigate the effects of an out of sync endpoint cache.
```

```docs

```

### DIFF
--- a/pkg/controller/endpoint/endpoints_controller.go
+++ b/pkg/controller/endpoint/endpoints_controller.go
@@ -110,6 +110,9 @@ func NewEndpointController(podInformer coreinformers.PodInformer, serviceInforme
 	e.podLister = podInformer.Lister()
 	e.podsSynced = podInformer.Informer().HasSynced
 
+	endpointsInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		DeleteFunc: e.onEndpointsDelete,
+	})
 	e.endpointsLister = endpointsInformer.Lister()
 	e.endpointsSynced = endpointsInformer.Informer().HasSynced
 
@@ -284,6 +287,15 @@ func (e *EndpointController) onServiceDelete(obj interface{}) {
 	}
 
 	e.serviceSelectorCache.Delete(key)
+	e.queue.Add(key)
+}
+
+func (e *EndpointController) onEndpointsDelete(obj interface{}) {
+	key, err := controller.KeyFunc(obj)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("Couldn't get key for object %+v: %v", obj, err))
+		return
+	}
 	e.queue.Add(key)
 }
 
--- a/pkg/controller/endpoint/endpoints_controller_test.go
+++ b/pkg/controller/endpoint/endpoints_controller_test.go
@@ -159,6 +159,49 @@ func makeTestServer(t *testing.T, namespace string) (*httptest.Server, *utiltest
 	return httptest.NewServer(mux), &fakeEndpointsHandler
 }
 
+// makeBlockingEndpointDeleteTestServer will signal the blockNextAction channel on endpoint "POST" & "DELETE" requests. All
+// block endpoint "DELETE" requestsi will wait on a blockDelete signal to delete endpoint. If controller is nil, a error will
+// be sent in the response.
+func makeBlockingEndpointDeleteTestServer(t *testing.T, controller *endpointController, endpoint *v1.Endpoints, blockDelete, blockNextAction chan struct{}, namespace string) *httptest.Server {
+
+	handlerFunc := func(res http.ResponseWriter, req *http.Request) {
+		if controller == nil {
+			res.WriteHeader(http.StatusInternalServerError)
+			res.Write([]byte("controller has not been set yet"))
+			return
+		}
+
+		if req.Method == "POST" {
+			controller.endpointsStore.Add(endpoint)
+			blockNextAction <- struct{}{}
+		}
+
+		if req.Method == "DELETE" {
+			go func() {
+				// Delay the deletion of endoints to make endpoint cache out of sync
+				<-blockDelete
+				controller.endpointsStore.Delete(endpoint)
+				controller.onEndpointsDelete(endpoint)
+			}()
+			blockNextAction <- struct{}{}
+		}
+
+		res.WriteHeader(http.StatusOK)
+		res.Write([]byte(runtime.EncodeOrDie(clientscheme.Codecs.LegacyCodec(v1.SchemeGroupVersion), &v1.Endpoints{})))
+	}
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/api/v1/namespaces/"+namespace+"/endpoints", handlerFunc)
+	mux.HandleFunc("/api/v1/namespaces/"+namespace+"/endpoints/", handlerFunc)
+	mux.HandleFunc("/api/v1/namespaces/"+namespace+"/events", func(res http.ResponseWriter, req *http.Request) {})
+	mux.HandleFunc("/", func(res http.ResponseWriter, req *http.Request) {
+		t.Errorf("unexpected request: %v", req.RequestURI)
+		http.Error(res, "", http.StatusNotFound)
+	})
+	return httptest.NewServer(mux)
+
+}
+
 type endpointController struct {
 	*EndpointController
 	podStore       cache.Store
@@ -1954,9 +1997,105 @@ func TestEndpointPortFromServicePort(t *testing.T) {
 	}
 }
 
+// TestMultipleServiceChanges tests that endpoints that are not created because of an out of sync endpoints cache are eventually recreated
+// A service will be created. After the endpoints exist, the service will be deleted and the endpoints will not be deleted from the cache immediately.
+// After the service is recreated, the endpoints will be deleted replicating an out of sync cache. Expect that eventually the endpoints will be recreated.
+func TestMultipleServiceChanges(t *testing.T) {
+	ns := metav1.NamespaceDefault
+	expectedSubsets := []v1.EndpointSubset{{
+		Addresses: []v1.EndpointAddress{
+			{IP: "1.2.3.4", NodeName: &emptyNodeName, TargetRef: &v1.ObjectReference{Kind: "Pod", Name: "pod0", Namespace: ns}},
+		},
+	}}
+	endpoint := &v1.Endpoints{
+		ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: ns, ResourceVersion: "1"},
+		Subsets:    expectedSubsets,
+	}
+
+	controller := &endpointController{}
+	blockDelete := make(chan struct{})
+	blockNextAction := make(chan struct{})
+	stopChan := make(chan struct{})
+	testServer := makeBlockingEndpointDeleteTestServer(t, controller, endpoint, blockDelete, blockNextAction, ns)
+	defer testServer.Close()
+
+	*controller = *newController(testServer.URL, 0*time.Second)
+	addPods(controller.podStore, ns, 1, 1, 0, ipv4only)
+
+	go func() { controller.Run(1, stopChan) }()
+
+	svc := &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: ns},
+		Spec: v1.ServiceSpec{
+			Selector:  map[string]string{"foo": "bar"},
+			ClusterIP: "None",
+			Ports:     nil,
+		},
+	}
+
+	controller.serviceStore.Add(svc)
+	controller.onServiceUpdate(svc)
+	// blockNextAction should eventually unblock once server gets endpoint request.
+	waitForChanReceive(t, 1*time.Second, blockNextAction, "Service Add should have caused a request to be sent to the test server")
+
+	controller.serviceStore.Delete(svc)
+	controller.onServiceDelete(svc)
+	waitForChanReceive(t, 1*time.Second, blockNextAction, "Service Delete should have caused a request to be sent to the test server")
+
+	// If endpoints cache has not updated before service update is registered
+	// Services add will not trigger a Create endpoint request.
+	controller.serviceStore.Add(svc)
+	controller.onServiceUpdate(svc)
+
+	// Ensure the work queue has been processed by looping for up to a second to prevent flakes.
+	wait.PollImmediate(50*time.Millisecond, 1*time.Second, func() (bool, error) {
+		return controller.queue.Len() == 0, nil
+	})
+
+	// Cause test server to delete endpoints
+	close(blockDelete)
+	waitForChanReceive(t, 1*time.Second, blockNextAction, "Endpoint should have been recreated")
+
+	close(blockNextAction)
+	close(stopChan)
+}
+
+func TestEndpointsDeletionEvents(t *testing.T) {
+	ns := metav1.NamespaceDefault
+	testServer, _ := makeTestServer(t, ns)
+	defer testServer.Close()
+	controller := newController(testServer.URL, 0)
+	store := controller.endpointsStore
+	ep1 := &v1.Endpoints{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            "foo",
+			Namespace:       ns,
+			ResourceVersion: "rv1",
+		},
+	}
+
+	// Test Unexpected and Expected Deletes
+	store.Delete(ep1)
+	controller.onEndpointsDelete(ep1)
+
+	if controller.queue.Len() != 1 {
+		t.Errorf("Expected one service to be in the queue, found %d", controller.queue.Len())
+	}
+}
+
 func stringVal(str *string) string {
 	if str == nil {
 		return "nil"
 	}
 	return *str
+}
+
+// waitForChanReceive blocks up to the timeout waiting for the receivingChan to receive
+func waitForChanReceive(t *testing.T, timeout time.Duration, receivingChan chan struct{}, errorMsg string) {
+	timer := time.NewTimer(timeout)
+	select {
+	case <-timer.C:
+		t.Errorf(errorMsg)
+	case <-receivingChan:
+	}
 }


pkg/controller/endpoint/endpoints_controller_test.go
```diff
@@ -1990,6 +1990,200 @@ func TestEndpointPortFromServicePort(t *testing.T) {
...
+	endpoints.onServiceDelete(svc)
+	timer = time.NewTimer(1 * time.Second)
+	select {
+	// BlockcNextAction should eventually unblock once server gets endpoint request.
+	case <-timer.C:
+		t.Errorf("Service Delete should have caused a request to be sent to the test server")
+	case <-blockNextAction:
+	}
+
+	// if endpoints cache has not updated becfore service update is registered
```
aojea, line 2090: nit s/before/becfore/

pkg/controller/endpoint/endpoints_controller_test.go
```diff
@@ -1990,6 +1990,200 @@ func TestEndpointPortFromServicePort(t *testing.T) {
...
+		},
+	}
+
+	// Test Unexpected and Expected Adds
+	if endpoints.queue.Len() != 0 {
+		t.Errorf("Queue should be empty")
+	}
+	store.Add(ep1)
+	endpoints.onEndpointsUpdate(ep1)
+	if endpoints.queue.Len() != 1 {
```
aojea, line 2132: can we have a race here if we assert before the queue is populated?

pkg/controller/endpoint/endpoints_controller_test.go
```diff
@@ -1990,6 +1990,200 @@ func TestEndpointPortFromServicePort(t *testing.T) {
...
+		t.Errorf("Expected one service to be in the queue, found %d", endpoints.queue.Len())
+	}
+	// Clear the queue
+	for endpoints.queue.Len() != 0 {
```
aojea, line 2136: we confirmed that the Len is 1 before, should be possible to have more than 1 at this moment?

pkg/controller/endpoint/endpoints_controller.go
```diff
@@ -309,6 +319,38 @@ func (e *EndpointController) onServiceDelete(obj interface{}) {
 	e.queue.Add(key)
 }
 
+func (e *EndpointController) onEndpointsUpdate(obj interface{}) {
+	ep, ok := obj.(*v1.Endpoints)
+	if !ok {
+		utilruntime.HandleError(fmt.Errorf("Failed getting endpoint from %+v", obj))
```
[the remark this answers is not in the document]
swetharepakula: Fixed in the update here and for the delete function.

pkg/controller/endpoint/endpoints_controller_test.go
swetharepakula: This test does not try to replicate the race, but just makes sure the service is added to the queue to be processed when an unexpected endpoints event comes through. `TestMultipleServiceChanges` covers the race condition.

pkg/controller/endpoint/endpoints_controller_test.go
swetharepakula: That is true, this for loop is a remnant of an earlier iteration of the test. Updated so it just pops the one element.

pkg/controller/endpoint/endpoints_tracker.go
```diff
@@ -0,0 +1,75 @@
...
+limitations under the License.
+*/
+
+package endpoint
+
+import (
+	"sync"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
```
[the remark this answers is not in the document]
swetharepakula: Thanks! The BUILD file should have all the right dependencies now.

pkg/controller/endpoint/endpoints_tracker_test.go
```diff
@@ -0,0 +1,240 @@
...
+import (
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestHandleEndpointTrackerUpdate(t *testing.T) {
+
+	tracker := newEndpointsTracker()
```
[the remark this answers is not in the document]
swetharepakula: I initially envisioned the test where each case built upon the previous. Though for clarity I removed the duplicate test cases and a new tracker is initialized per test case. Now each case is independent of the previous.

pkg/controller/endpoint/endpoints_tracker_test.go
```diff
@@ -0,0 +1,240 @@
...
+		{
+			desc:           "add different service endpoint",
+			updateEndpoint: ep2,
+			checkEndpoint:  ep2,
+			expectHas:      true,
+			expectStale:    false,
+		},
+		{
+			desc:           "check for different resource version, since now added",
+			updateEndpoint: ep2,
```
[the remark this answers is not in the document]
swetharepakula: See above comment.

swetharepakula:
Thanks @aojea for the feedback and guidance! I have made updates based on your suggestions. Please let me know what you think.

aojea:
heh, nice finding, this is somehow breaking leader election in client-go, we need to investigate why @swetharepakula, I pasted below some logs that may give us a hint

```
F0720 19:58:50.404098       8 controllermanager.go:293] leaderelection lost
```

```
2020-07-20T19:35:47.795032275Z stderr F E0720 19:35:47.794880       1 controller.go:152] Unable to remove old endpoints from kubernetes service: StorageError: key not found, Code: 1, Key: /registry/masterleases/172.18.0.3, ResourceVersion: 0, AdditionalErrorMsg:
```
@robscott you have a new TODO for the endpointslice migration (cc @liggitt ) :)

https://github.com/kubernetes/client-go/blob/a1d715839a207876281587dee364fb74f5cc153b/tools/leaderelection/resourcelock/endpointslock.go#L65

liggitt:
> 2020-07-20T19:35:47.795032275Z stderr F E0720 19:35:47.794880       1 controller.go:152] Unable to remove old endpoints from kubernetes service: StorageError: key not found, Code: 1, Key: /registry/masterleases/172.18.0.3, ResourceVersion: 0, AdditionalErrorMsg:


That's not leader election in client-go, that's the kubernetes master endpoints writer.

liggitt:
As noted in https://github.com/kubernetes/kubernetes/issues/85353#issuecomment-554575105, endpoints are 1:1 with their service matched by name. I don't think we should add tracking overhead for endpoints, and should just watch endpoints deletes and requeue the corresponding service name.

swetharepakula:
@aojea, that was very helpful! My understanding now is that the endpoints created by the endpointslock aren't in the tracker so the endpoint controller views it as an undesired endpoint and tries to delete them.

@liggitt Without tracking, are you thinking we should queue services only on delete or all endpoint events? Both situations seem like we would be enqueuing many services unnecessarily.

liggitt:
We've survived without tracking updates for ~5 years. I would start by doing a requeue on delete to fix the known issue. In most cases, the service will not exist and the sync will be very cheap.

swetharepakula:
In that case. I will update the PR with that solution and remove the tracker. Thanks for the suggestion @liggitt !

liggitt:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/93030/pull-kubernetes-node-e2e/1285271293750415360 flake is #93210

swetharepakula:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/93030/pull-kubernetes-e2e-gce-100-performance/1285371796182274048/ flake is #93217

aojea:
> I would start by doing a requeue on delete

let me understand it, the theory is
1. create service
2. delete service
3. create service
4. the endpoint object is deleted but still in cache, so the controller does not create new endpoints

in order to solve it, after 2. we receive a delete endpoint and we requeue it to sync again

with this approach, we may solve 4 but , if the endpoint delete arrives later, can this new sync delete the new endpoints created by 3?

liggitt:
> can this new sync delete the new endpoints created by 3?

the controller is the thing creating the endpoints. if the service is present in its cache at the beginning of its sync method, it will create the endpoints. if the service is absent from the cache at the beginning of the sync method it will delete the corresponding endpoints.

aojea:
thanks for the explanation :smile: 
/retest

pkg/controller/endpoint/endpoints_controller_test.go
```diff
@@ -1954,6 +1954,146 @@ func TestEndpointPortFromServicePort(t *testing.T) {
...
+
+	endpoints.serviceStore.Add(svc)
+	endpoints.onServiceUpdate(svc)
+	timer := time.NewTimer(1 * time.Second)
+	select {
+	// blockNextAction should eventually unblock once server gets endpoint request.
+	case <-timer.C:
+		t.Errorf("Service Add should have caused a request to be sent to the test server")
+	case <-blockNextAction:
+	}
```
robscott, line 2041: Would it be possible to move this logic into a helper function? It looks like it's repeated a few times.

pkg/controller/endpoint/endpoints_controller_test.go
```diff
@@ -1954,6 +1954,146 @@ func TestEndpointPortFromServicePort(t *testing.T) {
...
+		Spec: v1.ServiceSpec{
+			Selector:  map[string]string{"foo": "bar"},
+			ClusterIP: "None",
+			Ports:     nil,
+		},
+	}
+
+	endpoints.serviceStore.Add(svc)
+	endpoints.onServiceUpdate(svc)
+	timer := time.NewTimer(1 * time.Second)
```
robscott, line 2035: Can we always guarantee that 1 second will be enough here?

pkg/controller/endpoint/endpoints_controller_test.go
```diff
@@ -1954,6 +1954,146 @@ func TestEndpointPortFromServicePort(t *testing.T) {
 	}
 }
 
+func TestMultipleServiceChanges(t *testing.T) {
```
robscott, line 1957: There's a lot going on in this test, it might help to have a comment above the function explaining what this test is trying to cover and a high level overview of how it accomplishes that.

pkg/controller/endpoint/endpoints_controller_test.go
```diff
@@ -1954,6 +1954,146 @@ func TestEndpointPortFromServicePort(t *testing.T) {
...
+			{IP: "1.2.3.4", NodeName: &emptyNodeName, TargetRef: &v1.ObjectReference{Kind: "Pod", Name: "pod0", Namespace: ns}},
+		},
+	}}
+	endpoint := &v1.Endpoints{
+		ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: ns, ResourceVersion: "1"},
+		Subsets:    expectedSubsets,
+	}
+
+	var endpointsStore cache.Store
+	var endpoints *endpointController
```
robscott, line 1971:
```suggestion
	var controller *endpointController
```

pkg/controller/endpoint/endpoints_controller_test.go
```diff
@@ -1954,6 +1954,146 @@ func TestEndpointPortFromServicePort(t *testing.T) {
...
+	stopChan := make(chan struct{})
+
+	handlerFunc := func(res http.ResponseWriter, req *http.Request) {
+		if endpointsStore == nil {
+			res.WriteHeader(http.StatusInternalServerError)
+			res.Write([]byte("endpointsStore has not been set yet"))
+			return
+		}
+		if endpoints == nil {
+
```
robscott, line 1983: This feels like an unnecessary new line. On the other hand, a new line before the `if endpoints == nil` line would help match the rest of the structure here.

pkg/controller/endpoint/endpoints_controller_test.go
```diff
@@ -1954,6 +1954,146 @@ func TestEndpointPortFromServicePort(t *testing.T) {
...
+			res.WriteHeader(http.StatusInternalServerError)
+			res.Write([]byte("endpoints"))
```
robscott, line 1985: This could use a more descriptive error message.

pkg/controller/endpoint/endpoints_controller_test.go
```diff
@@ -1954,6 +1954,146 @@ func TestEndpointPortFromServicePort(t *testing.T) {
...
+	}
+
+	mux.HandleFunc("/api/v1/namespaces/"+ns+"/endpoints", handlerFunc)
+	mux.HandleFunc("/api/v1/namespaces/"+ns+"/endpoints/", handlerFunc)
+	mux.HandleFunc("/api/v1/namespaces/"+ns+"/events", func(res http.ResponseWriter, req *http.Request) {})
+	mux.HandleFunc("/", func(res http.ResponseWriter, req *http.Request) {
+		t.Errorf("unexpected request: %v", req.RequestURI)
+		http.Error(res, "", http.StatusNotFound)
+	})
+	testServer := httptest.NewServer(mux)
```
robscott, line 2015: Would it be possible to extract this and the handlerFunc into some kind of helper func?

pkg/controller/endpoint/endpoints_controller_test.go
```diff
@@ -1954,6 +1954,146 @@ func TestEndpointPortFromServicePort(t *testing.T) {
...
+	timer = time.NewTimer(1 * time.Second)
+	select {
+	// blockNextAction should eventually unblock once server gets endpoint request.
+	case <-timer.C:
+		t.Errorf("Service Delete should have caused a request to be sent to the test server")
+	case <-blockNextAction:
+	}
+
+	// if endpoints cache has not updated before service update is registered
+	// Services add will not trigger a Create endpoint request.
```
robscott, line 2054: Nit: The capitalization feels inconsistent here.

pkg/controller/endpoint/endpoints_controller_test.go
```diff
@@ -1954,6 +1954,146 @@ func TestEndpointPortFromServicePort(t *testing.T) {
...
+
+	close(blockNextAction)
+	close(stopChan)
+}
+
+func TestEndpointsDeletionEvents(t *testing.T) {
+	ns := metav1.NamespaceDefault
+	testServer, _ := makeTestServer(t, ns)
+	defer testServer.Close()
+	endpoints := newController(testServer.URL, 0)
```
robscott, line 2078:
```suggestion
	controller := newController(testServer.URL, 0)
```

robscott (commented):
Thanks for the work on this! The actual change to the controller LGTM, just had some suggestions and questions around the tests.

pkg/controller/endpoint/endpoints_controller_test.go
swetharepakula: Moved this logic into a helper function.

pkg/controller/endpoint/endpoints_controller_test.go
swetharepakula: Looking at the other tests, when waiting for the queue to be processed 1 second seemed to be used. But if there is a chance of increased flakiness, I can increase the timeout. What would you suggest?

robscott: 👍 That does seem reasonable. The other tests have not been flaky.

robscott:
Thanks for the work on this!

/lgtm
/assign @freehan

MrHohn (approved):
Thanks for the fix and nice test!
/approve

PR #93047, opened 2020-07-13
Fix staticcheck for integration test framework, examples, and garbage…

helenfeng737:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
 /kind cleanup

Part of https://github.com/kubernetes/kubernetes/issues/92402

Part of issue: https://github.com/kubernetes/kubernetes/issues/92402

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/integration/examples/apiserver_test.go
+++ b/test/integration/examples/apiserver_test.go
@@ -77,6 +77,8 @@ func TestAggregatedAPIServer(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	// use a channel to hand off the error
+	errs := make(chan error, 1)
 	go func() {
 		o := sampleserver.NewWardleServerOptions(os.Stdout, os.Stderr)
 		o.RecommendedOptions.SecureServing.Listener = listener
@@ -90,9 +92,14 @@ func TestAggregatedAPIServer(t *testing.T) {
 			"--kubeconfig", wardleToKASKubeConfigFile,
 		})
 		if err := wardleCmd.Execute(); err != nil {
-			t.Fatal(err)
+			errs <- err
 		}
 	}()
+	// wait for it
+	err := <-errs
+	if err != nil {
+		t.Fatal(err)
+	}
 	directWardleClientConfig, err := waitForWardleRunning(t, kubeClientConfig, wardleCertDir, wardlePort)
 	if err != nil {
 		t.Fatal(err)
@@ -131,7 +138,7 @@ func TestAggregatedAPIServer(t *testing.T) {
 
 	// wait for the unavailable API service to be processed with updated status
 	err = wait.Poll(100*time.Millisecond, 5*time.Second, func() (done bool, err error) {
-		_, err = kubeClient.Discovery().ServerResources()
+		_, _, err = kubeClient.Discovery().ServerGroupsAndResources()
 		hasExpectedError := checkWardleUnavailableDiscoveryError(t, err)
 		return hasExpectedError, nil
 	})
--- a/test/integration/examples/apiserver_test.go
+++ b/test/integration/examples/apiserver_test.go
@@ -77,8 +77,6 @@ func TestAggregatedAPIServer(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	// use a channel to hand off the error
-	errs := make(chan error, 1)
 	go func() {
 		o := sampleserver.NewWardleServerOptions(os.Stdout, os.Stderr)
 		o.RecommendedOptions.SecureServing.Listener = listener
@@ -92,19 +90,14 @@ func TestAggregatedAPIServer(t *testing.T) {
 			"--kubeconfig", wardleToKASKubeConfigFile,
 		})
 		if err := wardleCmd.Execute(); err != nil {
-			errs <- err
+			t.Errorf("Error executing command: %v", err)
 		}
 	}()
-	// wait for it
-	err = <-errs
-	if err != nil {
-		t.Fatal(err)
-	}
 	directWardleClientConfig, err := waitForWardleRunning(t, kubeClientConfig, wardleCertDir, wardlePort)
+
 	if err != nil {
 		t.Fatal(err)
 	}
-
 	// now we're finally ready to test. These are what's run by default now
 	wardleClient, err := client.NewForConfig(directWardleClientConfig)
 	if err != nil {
--- a/test/integration/examples/apiserver_test.go
+++ b/test/integration/examples/apiserver_test.go
@@ -96,7 +96,7 @@ func TestAggregatedAPIServer(t *testing.T) {
 		}
 	}()
 	// wait for it
-	err := <-errs
+	err = <-errs
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/test/integration/framework/test_server.go
+++ b/test/integration/framework/test_server.go
@@ -126,18 +126,11 @@ func StartTestServer(t *testing.T, stopCh <-chan struct{}, setup TestServerSetup
 	if err != nil {
 		t.Fatal(err)
 	}
-	// use a channel to hand off the error
-	errs := make(chan error, 1)
 	go func() {
 		if err := kubeAPIServer.GenericAPIServer.PrepareRun().Run(stopCh); err != nil {
-			errs <- err
+			t.Errorf("Failed to launch generic api server: %v", err)
 		}
 	}()
-	// wait for it
-	err = <-errs
-	if err != nil {
-		t.Fatal(err)
-	}
 
 	// Adjust the loopback config for external use (external server name and CA)
 	kubeAPIServerClientConfig := rest.CopyConfig(kubeAPIServerConfig.GenericConfig.LoopbackClientConfig)
--- a/test/integration/framework/test_server.go
+++ b/test/integration/framework/test_server.go
@@ -126,11 +126,18 @@ func StartTestServer(t *testing.T, stopCh <-chan struct{}, setup TestServerSetup
 	if err != nil {
 		t.Fatal(err)
 	}
+	// use a channel to hand off the error
+	errs := make(chan error, 1)
 	go func() {
 		if err := kubeAPIServer.GenericAPIServer.PrepareRun().Run(stopCh); err != nil {
-			t.Fatal(err)
+			errs <- err
 		}
 	}()
+	// wait for it
+	err = <-errs
+	if err != nil {
+		t.Fatal(err)
+	}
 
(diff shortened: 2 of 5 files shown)


test/integration/examples/apiserver_test.go
```diff
@@ -90,7 +90,7 @@ func TestAggregatedAPIServer(t *testing.T) {
 			"--kubeconfig", wardleToKASKubeConfigFile,
 		})
 		if err := wardleCmd.Execute(); err != nil {
-			t.Fatal(err)
+			t.Errorf("unexpected error: %v", err)
```
neolit123, line 93: could you explain this change?

test/integration/framework/test_server.go
```diff
@@ -128,7 +128,7 @@ func StartTestServer(t *testing.T, stopCh <-chan struct{}, setup TestServerSetup
 	}
 	go func() {
 		if err := kubeAPIServer.GenericAPIServer.PrepareRun().Run(stopCh); err != nil {
-			t.Fatal(err)
+			t.Errorf("unexpected error: %v", err)
```
neolit123, line 131: this too.

test/integration/garbagecollector/garbage_collector_test.go
```diff
@@ -443,7 +443,7 @@ func setupRCsPods(t *testing.T, gc *garbagecollector.GarbageCollector, clientSet
 	rc.ObjectMeta.Finalizers = initialFinalizers
 	rc, err := rcClient.Create(context.TODO(), rc, metav1.CreateOptions{})
 	if err != nil {
-		t.Fatalf("Failed to create replication controller: %v", err)
+		t.Errorf("Failed to create replication controller: %v", err)
```
neolit123, line 446: [1]

test/integration/garbagecollector/garbage_collector_test.go
```diff
@@ -453,18 +453,15 @@ func setupRCsPods(t *testing.T, gc *garbagecollector.GarbageCollector, clientSet
 		pod := newPod(podName, namespace, []metav1.OwnerReference{{UID: rc.ObjectMeta.UID, Name: rc.ObjectMeta.Name}})
 		createdPod, err := podClient.Create(context.TODO(), pod, metav1.CreateOptions{})
 		if err != nil {
-			t.Fatalf("Failed to create Pod: %v", err)
+			t.Errorf("Failed to create Pod: %v", err)
```
neolit123, line 456: [1]

test/integration/garbagecollector/garbage_collector_test.go
```diff
@@ -485,12 +482,12 @@ func setupRCsPods(t *testing.T, gc *garbagecollector.GarbageCollector, clientSet
 			return true, nil
 		})
 		if err != nil {
-			t.Fatalf("failed to observe the expected pods in the GC graph for rc %s", rcName)
+			t.Errorf("failed to observe the expected pods in the GC graph for rc %s", rcName)
```
neolit123, line 485: [1]

test/integration/examples/apiserver_test.go
ZhiFeng1993: I got this: 
```
Errors from staticcheck:
test/integration/examples/apiserver_test.go:80:2: the goroutine calls T.Fatalf, which must be called in the same goroutine as the test (SA2002)
test/integration/examples/apiserver_test.go:93:4: call to T.Fatalf
```
https://github.com/kubernetes/kubernetes/blob/master/test/integration/examples/apiserver_test.go#L78

This function is used by some other function so I assume changing this will not change the test behavior?

neolit123: having another look,
the correct fix in such a case is to capture the potential error from the goroutine and call t.Fatal in the main thread.

@kubernetes/sig-api-machinery-pr-reviews

ZhiFeng1993: @neolit123 thanks for the advice. It makes sense to me. I think it is a better fix. Will update

neolit123:
the latest update seems fine, overall.
i think given this touches api-server related integration tests it should get review from @kubernetes/sig-api-machinery-pr-reviews 
if nobody is responding try asking in #sig-api-machinery on k8s slack.

also we are in code freeze for 1.19 and this should merge after code freeze is lifted.

ZhiFeng1993:
> the latest update seems fine, overall.
> i think given this touches api-server related integration tests it should get review from @kubernetes/sig-api-machinery-pr-reviews
> if nobody is responding try asking in #sig-api-machinery on k8s slack.
> 
> also we are in code freeze for 1.19 and this should merge after code freeze is lifted.

Will do. Thanks for the review.

test/integration/examples/apiserver_test.go
```diff
@@ -90,9 +92,14 @@ func TestAggregatedAPIServer(t *testing.T) {
 			"--kubeconfig", wardleToKASKubeConfigFile,
 		})
 		if err := wardleCmd.Execute(); err != nil {
-			t.Fatal(err)
+			errs <- err
```
knight42, line 95: I think you need to `close(errs)`  if no error occurs, otherwise `err = <-errs` would block indefinitely.

test/integration/garbagecollector/garbage_collector_test.go
```diff
@@ -443,7 +443,7 @@ func setupRCsPods(t *testing.T, gc *garbagecollector.GarbageCollector, clientSet
 	rc.ObjectMeta.Finalizers = initialFinalizers
 	rc, err := rcClient.Create(context.TODO(), rc, metav1.CreateOptions{})
 	if err != nil {
-		t.Fatalf("Failed to create replication controller: %v", err)
+		errs <- err
```
knight42, line 446:
```suggestion
		errs <- fmt.Errorf("Failed to create replication controller: %v", err)
```
it might be better to perserve the original error message, like

test/integration/garbagecollector/garbage_collector_test.go
```diff
@@ -433,7 +433,7 @@ func TestCreateWithNonExistentOwner(t *testing.T) {
 	}
 }
 
-func setupRCsPods(t *testing.T, gc *garbagecollector.GarbageCollector, clientSet clientset.Interface, nameSuffix, namespace string, initialFinalizers []string, options metav1.DeleteOptions, wg *sync.WaitGroup, rcUIDs chan types.UID) {
+func setupRCsPods(t *testing.T, gc *garbagecollector.GarbageCollector, clientSet clientset.Interface, nameSuffix, namespace string, initialFinalizers []string, options metav1.DeleteOptions, wg *sync.WaitGroup, rcUIDs chan types.UID, errs chan error) {
```
knight42, line 436: again, `close(errs)` if no error occurs

test/integration/examples/apiserver_test.go
```diff
@@ -131,7 +138,7 @@ func TestAggregatedAPIServer(t *testing.T) {
 
 	// wait for the unavailable API service to be processed with updated status
 	err = wait.Poll(100*time.Millisecond, 5*time.Second, func() (done bool, err error) {
-		_, err = kubeClient.Discovery().ServerResources()
+		_, _, err = kubeClient.Discovery().ServerGroupsAndResources()
```
liggitt, line 141: why is this changed?

test/integration/garbagecollector/garbage_collector_test.go
```diff
@@ -534,16 +531,43 @@ func TestStressingCascadingDeletion(t *testing.T) {
 	wg.Add(collections * 5)
 	rcUIDs := make(chan types.UID, collections*5)
 	for i := 0; i < collections; i++ {
+		// use a channel to hand off the error
 		// rc is created with empty finalizers, deleted with nil delete options, pods will remain.
-		go setupRCsPods(t, gc, clientSet, "collection1-"+strconv.Itoa(i), ns.Name, []string{}, metav1.DeleteOptions{}, &wg, rcUIDs)
+		errs := make(chan error, 1)
+		go setupRCsPods(t, gc, clientSet, "collection1-"+strconv.Itoa(i), ns.Name, []string{}, metav1.DeleteOptions{}, &wg, rcUIDs, errs)
+		err := <-errs
```
liggitt, line 538: blocking on receiving the error means these all run serially, which defeats the purpose of the test. it would be simpler to change the Fatal to an Error

test/integration/framework/test_server.go
```diff
@@ -126,11 +126,18 @@ func StartTestServer(t *testing.T, stopCh <-chan struct{}, setup TestServerSetup
...
+	// use a channel to hand off the error
+	errs := make(chan error, 1)
 	go func() {
 		if err := kubeAPIServer.GenericAPIServer.PrepareRun().Run(stopCh); err != nil {
-			t.Fatal(err)
+			errs <- err
 		}
 	}()
+	// wait for it
+	err = <-errs
```
liggitt, line 137: blocking here until the server stops defeats the purpose of starting it in a goroutine

test/integration/examples/apiserver_test.go
```diff
@@ -90,9 +92,14 @@ func TestAggregatedAPIServer(t *testing.T) {
 			"--kubeconfig", wardleToKASKubeConfigFile,
 		})
 		if err := wardleCmd.Execute(); err != nil {
-			t.Fatal(err)
+			errs <- err
 		}
 	}()
+	// wait for it
+	err = <-errs
```
liggitt, line 99: blocking here until the server stops defeats the purpose of starting it in a goroutine

test/integration/garbagecollector/garbage_collector_test.go
ZhiFeng1993: Will do.

test/integration/examples/apiserver_test.go
ZhiFeng1993: Thanks. I'll remove the channel and change it to err per @liggitt's suggestion.

test/integration/examples/apiserver_test.go
ZhiFeng1993: Will remove the channel and change it to regular error message.

test/integration/examples/apiserver_test.go
ZhiFeng1993: This is a static check error. `ServerResources` has been deprecated so we'd use `ServerGroupsAndResources` instead.

test/integration/framework/test_server.go
ZhiFeng1993: Will change back to t.Error

test/integration/garbagecollector/garbage_collector_test.go
ZhiFeng1993: Will change back to t.Error

PR #93052, opened 2020-07-14
fix: initial delay in mounting azure disk & file

### LINKED ISSUE #93025: initial delay when starting Pods with Azure File volumes
**What happened**:

When starting pods with AFS volumes, there is an initial delay of five seconds until the pod is transitioning from the "Scheduled" state. The reason for this is that currently the volume mounting happens inside a wait.Poll which will initially wait a specified interval (currently 5 seconds) before execution.

Event log for comparison with and without AFS volume attached (seconds since event):

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

/assign 
/priority important-soon
/sig cloud-provider
/area provider/azure

andyzhangx:
**What type of PR is this?**
/kind bug

This PR uses `wait.PollImmediate` instead of `wait.Poll` function to fix
 - initial delay(5s) when mount azure file
 - initial delay(1s) when mount azure disk

https://github.com/kubernetes/kubernetes/blob/c46c1c043ee8de4c021c03ddb0fb157d1c1ba3f8/staging/src/k8s.io/apimachinery/pkg/util/wait/wait.go#L410-L442

Fixes #93025

```release-note
fix: initial delay in mounting azure disk & file
```

```docs
fix: initial delay in mounting azure disk & file
```

/assign @feiskyer 
/priority important-soon
/sig cloud-provider
/area provider/azure

### DIFF
--- a/pkg/volume/azure_dd/attacher.go
+++ b/pkg/volume/azure_dd/attacher.go
@@ -163,7 +163,7 @@ func (a *azureDiskAttacher) WaitForAttach(spec *volume.Spec, devicePath string, 
 
 	newDevicePath := ""
 
-	err = wait.Poll(1*time.Second, timeout, func() (bool, error) {
+	err = wait.PollImmediate(1*time.Second, timeout, func() (bool, error) {
 		if newDevicePath, err = findDiskByLun(int(lun), io, exec); err != nil {
 			return false, fmt.Errorf("azureDisk - WaitForAttach ticker failed node (%s) disk (%s) lun(%v) err(%s)", nodeName, diskName, lun, err)
 		}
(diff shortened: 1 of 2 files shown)


andyzhangx:
cc @chriserik

feiskyer:
Added to v1.19 since it is serious bug fix
/milestone v1.19

feiskyer (commented):
/lgtm
/approve

liggitt:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/93052/pull-kubernetes-node-e2e/1283730470764285952 flake was #93210

PR #91587, opened 2020-05-29
optimize insecure tls cipher checks within one loop

KielChan:
**What type of PR is this?**

> /kind cleanup

**What this PR does / why we need it**:

Old insecure cipher are checked by two for loops nested and one extra loop to fetch all insecure cipher Info, [here](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/server.go#L1014)
The new one will collect insecure ciphers within one loop to check all ciphers in flag valid or not.
```golang
func TLSCipherSuites(cipherNames []string) ([]uint16, []string, error) {
	if len(cipherNames) == 0 {
		return nil, nil, nil
	}
	ciphersIntSlice := make([]uint16, 0)
	insecureCipherNamesMatched := make([]string, 0)
	possibleCiphers := allCiphers()
	for _, cipher := range cipherNames {
		intValue, ok := possibleCiphers[cipher]
		if !ok {
			return nil, nil, fmt.Errorf("Cipher suite %s not supported or doesn't exist", cipher)
		}
		ciphersIntSlice = append(ciphersIntSlice, intValue)

		if _, ok := insecureCiphers[cipher]; ok {
			insecureCipherNamesMatched = append(insecureCipherNamesMatched, cipher)
		}
	}
	return ciphersIntSlice, insecureCipherNamesMatched, nil
}
```

Then:
```golang
//...
	tlsCipherSuites, insecureCipherNames, err := cliflag.TLSCipherSuites(kc.TLSCipherSuites)
	if err != nil {
		return nil, err
	}

	for _, cipherName := range insecureCipherNames {
		klog.Warningf("Use of insecure cipher '%s' detected.", cipherName)
	}

//...
```

Maybe that is cool than before.

**Which issue(s) this PR fixes**: 
none

Fixes # 
**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #93053, opened 2020-07-14
replace InsecureTLSCiphers function with InsecureTLSCipherNamesUsed

KielChan:
**What type of PR is this?**

> /kind cleanup

Old insecure cipher are checked by two `for` loops nested and one extra loop to fetch all insecure cipher Info.
The new one will collect insecure ciphers within one loop to check all ciphers in flag valid or not, and remove fussy codes by comparing cipher ids in two place: [here](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/server.go#L1012-L1021) and [here](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/secure_serving.go#L66-L72).

It is a duplicate PR like https://github.com/kubernetes/kubernetes/pull/91587. For some reason(CI test or milestone), the former one was closed. 

none

Fixes # 

```release-note

```

```docs

```

### DIFF
--- a/cmd/kubelet/app/server.go
+++ b/cmd/kubelet/app/server.go
@@ -59,7 +59,6 @@ import (
 	"k8s.io/client-go/util/connrotation"
 	"k8s.io/client-go/util/keyutil"
 	cloudprovider "k8s.io/cloud-provider"
-	"k8s.io/component-base/cli/flag"
 	cliflag "k8s.io/component-base/cli/flag"
 	"k8s.io/component-base/configz"
 	"k8s.io/component-base/featuregate"
@@ -1009,15 +1008,8 @@ func InitializeTLS(kf *options.KubeletFlags, kc *kubeletconfiginternal.KubeletCo
 		return nil, err
 	}
 
-	if len(tlsCipherSuites) > 0 {
-		insecureCiphers := flag.InsecureTLSCiphers()
-		for i := 0; i < len(tlsCipherSuites); i++ {
-			for cipherName, cipherID := range insecureCiphers {
-				if tlsCipherSuites[i] == cipherID {
-					klog.Warningf("Use of insecure cipher '%s' detected.", cipherName)
-				}
-			}
-		}
+	for _, cipherName := range cliflag.InsecureTLSCipherNamesUsed(kc.TLSCipherSuites) {
+		klog.Warningf("Use of insecure cipher '%s' detected.", cipherName)
 	}
 
 	minTLSVersion, err := cliflag.TLSVersion(kc.TLSMinVersion)
--- a/staging/src/k8s.io/apiserver/pkg/server/options/serving.go
+++ b/staging/src/k8s.io/apiserver/pkg/server/options/serving.go
@@ -260,6 +260,10 @@ func (s *SecureServingOptions) ApplyTo(config **server.SecureServingInfo) error 
 			return err
 		}
 		c.CipherSuites = cipherSuites
+
+		for _, cipherName := range cliflag.InsecureTLSCipherNamesUsed(s.CipherSuites) {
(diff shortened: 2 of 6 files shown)


KielChan:
@SergeyKanzhelev Could you help me review this PR? : )

cmd/kubelet/app/server.go
```diff
@@ -1009,15 +1008,8 @@ func InitializeTLS(kf *options.KubeletFlags, kc *kubeletconfiginternal.KubeletCo
...
-	if len(tlsCipherSuites) > 0 {
-		insecureCiphers := flag.InsecureTLSCiphers()
-		for i := 0; i < len(tlsCipherSuites); i++ {
-			for cipherName, cipherID := range insecureCiphers {
-				if tlsCipherSuites[i] == cipherID {
-					klog.Warningf("Use of insecure cipher '%s' detected.", cipherName)
-				}
-			}
-		}
+	for _, cipherName := range cliflag.InsecureTLSCipherNamesUsed(kc.TLSCipherSuites) {
```
SergeyKanzhelev, line 1011: it may be even more optimal to return insecure certs list from the `cliflag.TLSCipherSuites`. Eliminating back mapping from IDs to names here is a good thing fro sure

staging/src/k8s.io/apiserver/pkg/server/options/serving.go
```diff
@@ -260,6 +260,10 @@ func (s *SecureServingOptions) ApplyTo(config **server.SecureServingInfo) error
 			return err
 		}
 		c.CipherSuites = cipherSuites
+
+		for _, cipherName := range cliflag.InsecureTLSCipherNamesUsed(s.CipherSuites) {
```
SergeyKanzhelev, line 264: can you add more justification in PR description on why this check is moved here.

KielChan:
Sorry for my delay in responding. I will fix it today.

staging/src/k8s.io/apiserver/pkg/server/options/serving.go
KielChan: The old one will check when secure server prepares to run, for example in [kube-scheduler](https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-scheduler/app/server.go#L178-L185), it will check insecure cipher suites in `tlsConfig` method in `serve` function,
And the new one will check it if insecure cipher suites used when `ApplyTo` options [here](https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-scheduler/app/server.go#L295-L335)  and in `config` method.
It is the same check but the new one will check it when init options rather than check it after all things are ok to run secure server. And the old one checks used cipher ids obtained by `TLSCipherSuites` function, that's not cool.

KielChan: Some components like `kube-apiserver` is as same as secure server mentioned before, its code is [here](https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-apiserver/app/server.go#L454)

KielChan:
@SergeyKanzhelev Could you help me continue to review this PR? : )

staging/src/k8s.io/apiserver/pkg/server/options/serving.go
SergeyKanzhelev: This is just a warning. Not an error that will prevent server to start. So really doesn't really matter if it will be written from config Initialization or actual place where the config will be applied. I'm not very familiar with this initialization sequence. Can it happen that server will not be started because of other reasons after this warning was already written? Will it be any confusing for the customer?

One justification I can suggest is if instead of the new method `InsecureTLSCipherNamesUsed` the list of insecure ones will be returned from the `cliflag.TLSCipherSuites`. This will get rid of one more loop here.

PR #93060, opened 2020-07-14
Update the outdated comment of RotateCertificates for kubelet

fisherxu:
**What type of PR is this?**
/kind cleanup

Now the default value of `RotateCertificates` in kubelet is `true`.

Fixes #

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/staging/src/k8s.io/kubelet/config/v1beta1/types.go
+++ b/staging/src/k8s.io/kubelet/config/v1beta1/types.go
@@ -182,7 +182,7 @@ type KubeletConfiguration struct {
 	// Dynamic Kubelet Config (beta): If dynamically updating this field, consider that
 	// disabling it may disrupt the Kubelet's ability to authenticate with the API server
 	// after the current certificate expires.
-	// Default: false
+	// Default: true
 	// +optional
 	RotateCertificates bool `json:"rotateCertificates,omitempty"`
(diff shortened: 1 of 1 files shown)


staging/src/k8s.io/kubelet/config/v1beta1/types.go
```diff
@@ -182,7 +182,7 @@ type KubeletConfiguration struct {
 	// Dynamic Kubelet Config (beta): If dynamically updating this field, consider that
 	// disabling it may disrupt the Kubelet's ability to authenticate with the API server
 	// after the current certificate expires.
-	// Default: false
+	// Default: true
```
liggitt, line 185: this does not default to true

liggitt:
> Now the default value of `RotateCertificates` in kubelet is `true`.

that is not accurate, kubelet config continues defaulting RotateCertificates to false

fisherxu:
> that is not accurate, kubelet config continues defaulting RotateCertificates to false

Okey, get it, I just found the true default value is set in keadm... btw, I guess keadm has other configs so it can set RotateCertificates to true? @liggitt  

https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/componentconfigs/kubelet.go#L194

liggitt:
sure, kubeadm is free to set non-default configuration values

Issue #84615, opened 2019-10-31
Pull kubectl functions out of e2e test framework utils


MorrisLaw:
**Why is this needed:**
The original PR, [#77592](https://github.com/kubernetes/kubernetes/pull/77592), was made to address issue [#77095](https://github.com/kubernetes/kubernetes/issues/77095). It ended up getting pretty large and difficult to iterate on and difficult for others to review.

I've decided to break this out into smaller PRs, in order to iterate on this task much quicker. The goal is to move all kubectl-related functions out of `test/e2e/framework/util.go`.

alejandrox1:
/priority important-longterm
/lifecycle active
/retitle Pull kubectl functions out of e2e test framework utils

SataQiu:
I'm willing to help! Is there a more detailed division of work? @MorrisLaw

MorrisLaw:
Hey @SataQiu I’ve started work on this already with the original PR. But It was getting too big and I thought it’d be easier to break the PR down into smaller pieces.

I made this issue to track that work..
The first PR which adds a `TestKubeconfig` struct is up now and is important for being able to abstract away the `kubectl-related` functions.

I haven’t broke down a division of work, but, I planned on using my original PR as a reference and breaking that PR out into smaller ones.

SataQiu:
@MorrisLaw I'd be happy to help you if you'd like, and we can do it bit by bit :)

BenTheElder:
Is this done or still active?
It shouldn't both be marked "help wanted" and "lifecycle/active"

oomichi:
I think this issue is important for e2e framework.
and I'd like to help this.
It is nice to remove `help wanted` label at this time.

/remove-help

oomichi:
Oops, I forgot to assign myself.

/assign @oomichi

oomichi:
https://github.com/kubernetes/kubernetes/pull/93213 can solve this issue.
In addition, I am not working for this issue as the above PR.
/unassign @oomichi 
/remove-lifecycle stale

Issue #77095, opened 2019-04-25
Break up and Refactor test/e2e/framework/util.go


andrewsykim:
**What would you like to be added**:
There are a lot of methods in `test/e2e/framework/util.go` that can be cleaned up or moved into it's own package for easier consumption/discovery. In https://github.com/kubernetes/kubernetes/issues/76206 there were some discussions for how to break up this file. The proposed package structure was:

**test/e2e/framework/**
   - [ ] kubectl/ (@MorrisLaw)
            - builder.go
            -  runner.go   
   - [x] log/ (@draveness)
            - logger.go 
   - [x] pod/ (@alejandrox1) 
            - wait.go
            - resource.go
   - [x] node/ (@jiatongw)
            - wait.go
            - resource.go
   - [ ] service/  (@draveness) 
            - resource.go
   - [x] endpoints/ (@draveness) 
            - resource.go
   - [x] resource/ (@alejandrox1) 
            - runtime.go
   - [ ] lifecycle/ (@jiatongw)
            - versions.go
            - state.go
            - restart
- [ ] wrappers.go (@MorrisLaw) 
- [ ] providers.go (@alejandrox1) 

I think each package & top level file can be it's own task. Pick a package above, go through each method in `test/e2e/framework/util.go` and move it to that package/file if it makes sense. The above is just a guideline, feel free to introduce new packages/files where it makes sense. 

**Why is this needed**:
This is a part of https://github.com/kubernetes/kubernetes/issues/76206 & https://github.com/kubernetes/kubernetes/issues/75601

johnSchnake:
There are also ssh utilities that should probably move into their own package; I'll take that one as I think it is also a reasonable step in #69373 which is assigned to me (reducing reliance on ssh in tests). Step 1, figure out where it really bubbles up to, step 2, do something about it.

draveness:
I'd like to take the files as below :)

```go
service/*
endpoints/*
resource/*
```

alejandrox1:
I’d like to take kubectl/* and pod/* :)

WanLinghao:
I would like to take kubectl and log

WanLinghao:
Hi, all. When I try handle cluster thing, I found there are many functions containing k8s-component. Like 
```
CheckNodesReady
RestartControllerManager
WaitForApiserverUp
```
Should we combine node/ and cluster/, maybe something like:

- [ ] cluster/
      - kubelet.go
      - kube-proxy.go
      - controller-manager.go
[...]

PR #93065, opened 2020-07-14
Move LookForStringInLog() from e2e framework

oomichi:
**What type of PR is this?**

/kind cleanup

Both LookForStringInLog() and LookForStringInPodExec() call RunKubectlOrDie() internally, they seem like wrapper functions for RunKubectlOrDie().
We have long-term goal for moving kubectl related functions into e2ekubectl.
So this moves the above functions for the goal as a small step.

Ref: https://github.com/kubernetes/kubernetes/issues/84615
Ref: https://github.com/kubernetes/kubernetes/issues/77095

```release-note
NONE
```


oomichi:
hmm, this makes circular dependency..

/close

PR #93066, opened 2020-07-14
kubelet: add operations count and error count metrics to network plugin manager

AnishShah:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**: This PR adds additional metrics to network plugin manager to report total operations and total operations errors.

```release-note
The following new metrics are available.
* network_plugin_operations_total
* network_plugin_operations_errors_total
```

```docs

```

### DIFF
--- a/pkg/kubelet/dockershim/network/metrics/metrics.go
+++ b/pkg/kubelet/dockershim/network/metrics/metrics.go
@@ -28,9 +28,11 @@ import (
 
 const (
 	// NetworkPluginOperationsKey is the key for operation count metrics.
-	NetworkPluginOperationsKey = "network_plugin_operations"
+	NetworkPluginOperationsKey = "network_plugin_operations_total"
 	// NetworkPluginOperationsLatencyKey is the key for the operation latency metrics.
 	NetworkPluginOperationsLatencyKey = "network_plugin_operations_duration_seconds"
+	// NetworkPluginOperationsErrorsKey is the key for the operations error metrics.
+	NetworkPluginOperationsErrorsKey = "network_plugin_operations_errors_total"
 
 	// Keep the "kubelet" subsystem for backward compatibility.
 	kubeletSubsystem = "kubelet"
@@ -49,6 +51,28 @@ var (
 		},
 		[]string{"operation_type"},
 	)
+
+	// NetworkPluginOperations collects operation counts by operation type.
+	NetworkPluginOperations = metrics.NewCounterVec(
+		&metrics.CounterOpts{
+			Subsystem:      kubeletSubsystem,
+			Name:           NetworkPluginOperationsKey,
+			Help:           "Cumulative number of network plugin operations by operation type.",
+			StabilityLevel: metrics.ALPHA,
+		},
+		[]string{"operation_type"},
+	)
+
+	// NetworkPluginOperationsErrors collects operation errors by operation type.
+	NetworkPluginOperationsErrors = metrics.NewCounterVec(
+		&metrics.CounterOpts{
+			Subsystem:      kubeletSubsystem,
+			Name:           NetworkPluginOperationsErrorsKey,
+			Help:           "Cumulative number of network plugin operation errors by operation type.",
+			StabilityLevel: metrics.ALPHA,
+		},
+		[]string{"operation_type"},
+	)
 )
 
 var registerMetrics sync.Once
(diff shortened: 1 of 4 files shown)


AnishShah:
I signed it

AnishShah:
I signed it

pkg/kubelet/dockershim/network/metrics/metrics.go
```diff
@@ -28,9 +28,11 @@ import (
 
 const (
 	// NetworkPluginOperationsKey is the key for operation count metrics.
-	NetworkPluginOperationsKey = "network_plugin_operations"
+	NetworkPluginOperationsKey = "network_plugin_operations_total"
```
johnbelamaric, line 31: won't changing this cause backwards compatibility issues for existing dashboards and alerts? Or I guess looking below, this was never used before?

Are these names all following the latest guidance from the Prometheus project?

pkg/kubelet/dockershim/network/plugins.go
```diff
@@ -382,45 +382,57 @@ func (pm *PluginManager) podUnlock(fullPodName string) {
...
-	defer recordOperation("get_pod_network_status", time.Now())
+	const operation = "get_pod_network_status"
+	defer recordOperation(operation, time.Now())
 	fullPodName := kubecontainer.BuildPodFullName(podName, podNamespace)
 	pm.podLock(fullPodName).Lock()
 	defer pm.podUnlock(fullPodName)
 
 	netStatus, err := pm.plugin.GetPodNetworkStatus(podNamespace, podName, id)
 	if err != nil {
+		recordError(operation)
```
johnbelamaric, line 403: would it be useful to differentiate the latency metrics on successful vs error operations?

pkg/kubelet/dockershim/network/metrics/metrics.go
AnishShah: Thanks John for the quick review. I don't see `NetworkPluginOperationsKey` being used anywhere in this file or in the kubernetes repo when I used Github-search.

The names do follow the convention from the Prometheus project - https://prometheus.io/docs/practices/naming/#metric-names. 

The reason for change from "network_plugin_operations" to "network_plugin_operations_total" was to follow this convention.

pkg/kubelet/dockershim/network/plugins.go
AnishShah: That would cause some backwards compatibility issue I think. We can always correlated the data from the newly added metrics to see latency for successful and error operations.

pkg/kubelet/dockershim/network/metrics/metrics.go
```diff
@@ -49,6 +51,28 @@ var (
...
+		},
+		[]string{"operation_type"},
+	)
+
+	// NetworkPluginOperationsErrors collects operation errors by operation type.
+	NetworkPluginOperationsErrors = metrics.NewCounterVec(
+		&metrics.CounterOpts{
+			Subsystem:      kubeletSubsystem,
+			Name:           NetworkPluginOperationsErrorsKey,
+			Help:           "Cumulative number of network plugin operations error by operation type.",
```
SergeyKanzhelev, line 71: nit:

```suggestion
			Help:           "Cumulative number of network plugin operation errors by operation type.",
```

AnishShah: Done.

AnishShah:
I don't think the tests are failing due to my changes. I feel like this is a flake.

SergeyKanzhelev:
I looked at previous failure. It's in CNI logic - timeout when it expects failure. It MIGHT be related to the new changes. I didn't have time to look further. Let's see how retest will go.

AnishShah:
@SergeyKanzhelev Thanks for your quick response. Each time I retest, a different test fails. Any idea if pull-kubernetes-integration tests are flaky?

AnishShah:
@SergeyKanzhelev looks like the tests passed

AnishShah:
@dashpole @SergeyKanzhelev @johnbelamaric Is this ready to merge? Can I get an approval?

dashpole:
These metrics need unit tests.  See https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/component-base/metrics/testutil.

dashpole:
This is a reasonable example to follow: https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/internal/queue/scheduling_queue_test.go#L1181

You can use GatherAndCompare to check that the prometheus metrics match what you expect them to be.

AnishShah:
@dashpole Added unit tests for the metrics. PTAL

AnishShah:
/retest

@dashpole Can I get LGTM again? I think the test failed because I forgot to add license on top of plugings_test.go file.

freehan (approved):
/approve
/lgtm

PR #93068, opened 2020-07-14
[wip] expose featuregate names in k8s.io/api

deads2k:
Feature gates for k/k are coupled to the level of kubernetes and different each release.  We have another repo that is similarly coupled and follows the same deprecation lifecycle and has the same set of approvers: k8s.io/api.

1. We can move the feature gate definitions (the names and whether they are on or off by default in a particular version of kube) to k8s.io/api.
2. We can also move the default instance of feature gates, without any registration init logic, to k8s.io/component-base (currently they seem to be accidentally held in k8s.io/apiserver)
3. Binaries can then call `AddCurrentKubernetesFeatureGates` to `k8s.io/component-base/featuregateinstance.DefaultMutableFeatureGates`.

Since we already expose and promise some compatibility on featuregate names, the exposure of names (but not the featuregate instance itself) in k8s.io/api follows our existing guarantees. Anyone relying on them will see compile failures if they go away (naughty people), but at least they will catch them.  If any registration is messed up, missing feature gates result in panics.

This PR is actually fairly close, but since k8s.io/api should not depend on k8s.io/component-base, it would require one addition refactor to make compatible golang interfaces in each location to make the AddAll method palatable to write.  For 1.19 delivery, we could consider quickly writing an adapter in k8s.io/kubernetes/pkg/feature.

/kind cleanup
@kubernetes/sig-api-machinery-misc 
/assign @cheftako @cici37 @liggitt 

```release-note
NONE
```

/hold  this is a poc.


liggitt:
> 1. whether they are on or off by default in a particular version of kube

that seems like it will be misinterpreted by clients to mean "this feature is off" when what it really means is "this feature's default was off in the version of the API library I built against"

I'm not convinced moving all the features we have defined to be externally inspectable like this is great.

Two things we discussed in the [code-organization meeting](https://docs.google.com/document/d/1HtTI0rJEGP_MSf6eO87aCmx_tzpovPAAg7U2Zxwm8FE/edit?ts=5f19d4d5#bookmark=id.yg5hwmvlft1):

* let's set up import guards to prevent accidental use from client-go / cli-runtime / kubectl until we mean to and have a good story for users configuring it
* limiting the number of features exposed via k8s.io/api to ones we expect to be consumed by external components like cloud-controller-manager seems good: exposing default feature enablement values introduces skew issues when consumed by external libraries

> 3\. Binaries can then call `AddCurrentKubernetesFeatureGates`

would that register all features? in code-organization we talked about having an easy registration function clients that don't care about curating can use to just register all exposed features. clients that do care can curate/register specific features.

PR #82653, opened 2019-09-12
Skip publishing OpenAPI for nonstructural schemas

liggitt:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
* Restores compatibility with 1.15 CRDs by not publishing openapi non-structural schemas
* Restores compatibility with 1.14 CRDs by not publishing openapi for structural schemas whose CRDs set `spec.preserveUnknownFields=true`

**Which issue(s) this PR fixes**:
Fixes #82651

**Does this PR introduce a user-facing change?**:
```release-note
Restores compatibility with <=1.15.x custom resources by not publishing OpenAPI for non-structural custom resource definitions
```

/assign @roycaihw @sttts 
/cc @aanm 
/sig api-machinery

PR #79636, opened 2019-07-02
Fix publishing x-kubernetes-preserve-unknown-fields working with kubectl

liggitt:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Fixes published openapi for CRDs with x-kubernetes-preserve-unknown-fields so that kubectl allows sending unknown fields, by dropping items/properties (including apiVersion/kind/metadata properties) from the published v2 schema.

Extracted from https://github.com/kubernetes/kubernetes/pull/79604 for separate merge and picking to 1.15

**Does this PR introduce a user-facing change?**:
```release-note
Fixes a bug in openapi published for custom resources using x-kubernetes-preserve-unknown-fields extensions, so that kubectl will allow sending unknown fields for that portion of the object.
```

/sig api-machinery
/priority important-soon
/cc @sttts @roycaihw

PR #79587, opened 2019-07-01
DO NOT publish openapi specs containing bad types

yue9944882:
for now, we can set arbitrary types for the json schema like:

```yaml
  validation:
    openAPIV3Schema:
      properties:
        spec:
          type: bug // whatever type except for the root element
      type: object
```
this is violating https://tools.ietf.org/html/draft-wright-json-schema-00#section-4.2 and this pull adds a list of supported types and their validation according to the specification.

https://swagger.io/docs/specification/data-models/data-types/

```release-note
NONE
```

/king bug
/sig api-machinery

/cc @sttts

PR #93078, opened 2020-07-14
Show error in status if preserve unknown fields is true for nonstructural schemas

vareti:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

CRDs with `preserveUnknownFields: true` won't be published via OpenAPI and `kubectl` explain will not work. This is intentional. But, nothing in the CRD status suggests that `preserveUnknownFields: true` is the problem. This PR adds an error message when that is the case for non-structural schema.

Fixes #

```release-note
Status of v1beta1 CRDs without "preserveUnknownFields:false" will show violation "spec.preserveUnknownFields: Invalid value: true: must be false"
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apiextensions-apiserver/pkg/controller/nonstructuralschema/nonstructuralschema_controller.go
+++ b/staging/src/k8s.io/apiextensions-apiserver/pkg/controller/nonstructuralschema/nonstructuralschema_controller.go
@@ -90,6 +90,12 @@ func calculateCondition(in *apiextensionsv1.CustomResourceDefinition) *apiextens
 
 	allErrs := field.ErrorList{}
 
+	if in.Spec.PreserveUnknownFields {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "preserveUnknownFields"),
+			in.Spec.PreserveUnknownFields,
+			fmt.Sprint("must be false")))
+	}
+
 	for i, v := range in.Spec.Versions {
 		if v.Schema == nil || v.Schema.OpenAPIV3Schema == nil {
 			continue
--- a/staging/src/k8s.io/apiextensions-apiserver/pkg/controller/nonstructuralschema/nonstructuralschema_controller_test.go
+++ b/staging/src/k8s.io/apiextensions-apiserver/pkg/controller/nonstructuralschema/nonstructuralschema_controller_test.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package nonstructuralschema
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+func Test_calculateCondition(t *testing.T) {
+	tests := []struct {
+		name string
+		args *apiextensionsv1.CustomResourceDefinition
+		want *apiextensionsv1.CustomResourceDefinitionCondition
+	}{
+		{
+			name: "preserve unknown fields is false",
+			args: &apiextensionsv1.CustomResourceDefinition{
+				Spec: apiextensionsv1.CustomResourceDefinitionSpec{
+					PreserveUnknownFields: false,
+				},
+			},
+		},
+		{
+			name: "preserve unknown fields is true",
+			args: &apiextensionsv1.CustomResourceDefinition{
+				Spec: apiextensionsv1.CustomResourceDefinitionSpec{
+					PreserveUnknownFields: true,
+				},
+			},
+			want: &apiextensionsv1.CustomResourceDefinitionCondition{
+				Type:   apiextensionsv1.NonStructuralSchema,
+				Status: apiextensionsv1.ConditionTrue,
+				Reason: "Violations",
+				Message: field.Invalid(field.NewPath("spec", "preserveUnknownFields"),
+					true,
+					fmt.Sprint("must be false")).Error(),
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := calculateCondition(tt.args); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("calculateCondition() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
--- a/staging/src/k8s.io/apiextensions-apiserver/test/integration/validation_test.go
+++ b/staging/src/k8s.io/apiextensions-apiserver/test/integration/validation_test.go
@@ -27,6 +27,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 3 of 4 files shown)


staging/src/k8s.io/apiextensions-apiserver/pkg/controller/nonstructuralschema/nonstructuralschema_controller.go
```diff
@@ -90,6 +90,12 @@ func calculateCondition(in *apiextensionsv1.CustomResourceDefinition) *apiextens
 
 	allErrs := field.ErrorList{}
 
+	if in.Spec.PreserveUnknownFields {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "preserveUnknownFields"),
+			in.Spec.PreserveUnknownFields,
+			fmt.Sprint("preserveUnknownFields must not be true")))
```
sttts, line 96: I believe the error string can be short,i.e. "must not be true". Would the condition still look good?

vareti:
Agreed. Field name is redundant in error message. Condition still looks good even after removing the field.

staging/src/k8s.io/apiextensions-apiserver/pkg/controller/nonstructuralschema/nonstructuralschema_controller.go
vareti: Agreed. Field name is redundant in error message. Condition still looks good even after removing the field.

staging/src/k8s.io/apiextensions-apiserver/test/integration/validation_test.go
```diff
@@ -889,7 +892,12 @@ spec:
 		unexpectedViolations                  []string
 	}
 	tests := []Test{
-		{"empty", "", "", "", nil, nil, nil, nil},
+		{
+			desc: "empty",
+			expectedViolations: []string{
+				"spec.preserveUnknownFields: Invalid value: true: must not be true",
```
sttts, line 898: can you make preserveUnknownFields a test input and add some test with and without?

staging/src/k8s.io/apiextensions-apiserver/test/integration/validation_test.go
```diff
@@ -801,21 +801,24 @@ spec:
...
 		obj, err := apiExtensionClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(context.TODO(), name, metav1.GetOptions{})
 		if err != nil {
 			return false, err
 		}
 		cond = findCRDCondition(obj, apiextensionsv1beta1.NonStructuralSchema)
-		return cond == nil, nil
+		if v := "spec.versions[0].schema.openAPIV3Schema.properties[a].type: Required value: must not be empty for specified object fields"; strings.Contains(cond.Message, v) {
+			return false, nil
+		}
+		return true, nil
```
sttts, line 814: would add another step where you set preserveUnknownFields to false. Then this last violation should also go.

staging/src/k8s.io/apiextensions-apiserver/pkg/controller/nonstructuralschema/nonstructuralschema_controller_test.go
```diff
@@ -0,0 +1,66 @@
...
+					PreserveUnknownFields: true,
+				},
+			},
+			want: &apiextensionsv1.CustomResourceDefinitionCondition{
+				Type:   apiextensionsv1.NonStructuralSchema,
+				Status: apiextensionsv1.ConditionTrue,
+				Reason: "Violations",
+				Message: field.Invalid(field.NewPath("spec", "preserveUnknownFields"),
+					true,
+					fmt.Sprint("must not be true")).Error(),
```
sttts, line 55: `must be false` is better IMO.

liggitt:
aren't there several reasons we'll elide the schema?

* spec.preserveUnknownFields (https://github.com/kubernetes/kubernetes/pull/82653)
* x-kubernetes-preserve-unknown-fields (https://github.com/kubernetes/kubernetes/pull/79636)
* array type missing items schema (https://github.com/kubernetes/kubernetes/pull/94888)
* invalid openapi types (https://github.com/kubernetes/kubernetes/pull/79587)

should we treat those coherently (either by adding/removing a condition for each of those, or a single condition to cover "reasons openapi isn't being published")?

sttts:
@liggitt don't mix non-structural and having constructs that kubectl does not understand or handle well.

This here is about the former. We have a NonStructural condition exactly for that, but forget the global `spec.preserveUnknownFields` to show as violation.

liggitt:
> This here is about the former. We have a NonStructural condition exactly for that, but forget the global `spec.preserveUnknownFields` to show as violation.

ah, ok

staging/src/k8s.io/apiextensions-apiserver/test/integration/validation_test.go
vareti: I added preserveUnknownFields as test input

staging/src/k8s.io/apiextensions-apiserver/test/integration/validation_test.go
vareti: If I set preserveUnknownFields to false, I get a validation error that says the all fields in schema will be pruned which I think is a valid error. Instead, I tried fixing both the validation errors with a patch. Now there will be no violations

staging/src/k8s.io/apiextensions-apiserver/pkg/controller/nonstructuralschema/nonstructuralschema_controller_test.go
vareti: Yes, sounds good. I updated all the references to `must be false`

staging/src/k8s.io/apiextensions-apiserver/test/integration/validation_test.go
```diff
@@ -934,7 +946,9 @@ properties:
  metadata:
    type: object
 `,
-			expectedViolations: []string{},
+			expectedViolations: []string{
+				"spec.preserveUnknownFields: Invalid value: true: must be false",
```
sttts, line 950: would instead set the value to false in all test fixtures by default, instead of changing all the expectations below.

staging/src/k8s.io/apiextensions-apiserver/test/integration/validation_test.go
```diff
@@ -763,45 +768,51 @@ spec:
...
+			return cond != nil, nil
+		})
 		if err != nil {
-			return false, err
+			t.Fatalf("unexpected error waiting for NonStructuralSchema condition: %v", cond)
+		}
+		for _, violation := range expectedViolations {
+			if !strings.Contains(cond.Message, violation) {
+				t.Fatalf("expected violation %q, but got: %v", violation, cond.Message)
+			}
```
sttts, line 788: am still a little unsure and surprised about this big fallout here in the test. Would expect something very local. Isn't there a way to rewrite this test by adding steps only without rewriting what we have?

vareti: I agree the change is big and the same can be achieved by also adding the steps. But the intent was to just refactor the code to avoid duplication. This part of code does not change any of the previous logic and only adds some checking for `preserveUnknownFields `

vareti: made the changes to update the steps locally

staging/src/k8s.io/apiextensions-apiserver/test/integration/validation_test.go
vareti: done!

staging/src/k8s.io/apiextensions-apiserver/test/integration/validation_test.go
```diff
@@ -780,21 +781,25 @@ spec:
...
+		// 2. preserveUnknownFields validation error
+		patch := []byte("[{\"op\":\"add\",\"path\":\"/spec/validation/openAPIV3Schema/properties/a/type\",\"value\":\"int\"}," +
+			"{\"op\":\"replace\",\"path\":\"/spec/preserveUnknownFields\",\"value\":false}]")
+		if _, err = apiExtensionClient.ApiextensionsV1beta1().CustomResourceDefinitions().Patch(context.TODO(), name, types.JSONPatchType, patch, metav1.PatchOptions{}); apierrors.IsConflict(err) {
 			continue
 		}
 		if err != nil {
 			t.Fatalf("unexpected update error: %v", err)
 		}
+		break
```
sttts, line 802: good catch

PR #93063, opened 2020-07-14
Fix no-new-privileges not supported by docker on Windows

wawa0210:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:

/kind bug

**What this PR does / why we need it**:

Fix kubernetes all windows pod creation failure

ref #90948 

This pr does not take into account the Windows scenario.
Windows container does not support `no-new-privileges`. If you directly specify that the container will not be created, an error will be reported directly.
image

**Which issue(s) this PR fixes**:

Fixes  #93061

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

/assign @yujuhong

PR #93080, opened 2020-07-14
Extending DSR loadbalancer feature in winkernel kube-proxy to HNS versions 9.3-9.max, 10.2+

### LINKED ISSUE #93304: Windows HNS version check for DSR feature is failing on RS5, even though it is supported
**What happened**:
I tried to enabled DSR loadbalancers on kube-proxy Windows but there was an error shown
"Platform does not support feature Direct Server Return (DSR)"

I expected kube-proxy to run in DSR mode on a supported Windows build.

Enable DSR mode using --enable-dsr=true --feature-gates="WinDSR=true"

The hcsshim [version check used by winkernel proxier](https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/winkernel/proxier.go#L545) needs to be updated to be inclusive of HNS versions used in Windows Server, version 1809.

**Environment**: Windows Server, version 1809

- Kubernetes version (use `kubectl version`): 1.19

- Cloud provider or hardware configuration: VM

- OS (e.g: `cat /etc/os-release`): Windows Server, version 1809

- Kernel (e.g. `uname -a`): Windows Server, version 1809 with HNS version 9.3

- Install tools: kubeadm

- Network plugin and version (if this is a network-related bug): Flannel host-gw mode
- Others:

elweb9858:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR introduces an updated version of hcsshim, which has the latest version check for DSR loadbalancers, which are used in winkernel kube-proxy. Customers running on older versions of Windows (HNS versions 9.3-9.max, 10.2+) will be able to use the DSR feature with these changes.

Fixes #93304

```release-note
Extended DSR loadbalancer feature in winkernel kube-proxy to HNS versions 9.3-9.max, 10.2+
```

```docs

```


elweb9858:
fyi @keith-mange @sbangari

elweb9858:
fyi @daschott

elweb9858:
@feiskyer I know it's a bit late in the cycle, but we are trying to get this in v1.19. Any chance you could mark this PR with the v1.19 milestone?

michmike (approved):
/lgtm
/approve

pkg/proxy/winkernel/proxier_test.go
```diff
@@ -374,6 +384,67 @@ func TestCreateLoadBalancer(t *testing.T) {
...
+	proxier.setInitialized(true)
+	proxier.syncProxyRules()
+
+	svc := proxier.serviceMap[svcPortName]
+	svcInfo, ok := svc.(*serviceInfo)
+	if !ok {
+		t.Errorf("Failed to cast serviceInfo %q", svcPortName.String())
+
+	} else {
+		if svcInfo.hnsID != guid {
```
kumarvin123, line 441: Can you also make sure that svcInfo has dsr field set to  true

feiskyer:
Michael has already added to v1.19. @elweb9858 could you also add a release note?

kumarvin123 (approved):

pkg/proxy/winkernel/proxier.go
```diff
@@ -1177,7 +1184,7 @@ func (proxier *Proxier) syncProxyRules() {
 			// Try loading existing policies, if already available
 			hnsLoadBalancer, err = hns.getLoadBalancer(
 				hnsEndpoints,
-				loadBalancerFlags{sessionAffinity: sessionAffinityClientIP, isIPv6: proxier.isIPv6Mode},
+				loadBalancerFlags{isDSR: svcInfo.dsr, sessionAffinity: sessionAffinityClientIP, isIPv6: proxier.isIPv6Mode},
```
ksubrmnn, line 1187: Would we want to also check proxier.isDSR for these other loadbalancer types?

elweb9858: I was not planning on altering any functionality surrounding isDSR and preserveDIP, @madhanrm could you weigh in on this?

pkg/proxy/winkernel/proxier.go
```diff
@@ -352,6 +353,11 @@ func (proxier *Proxier) newServiceInfo(port *v1.ServicePort, service *v1.Service
 	if err != nil {
 		preserveDIP = false
 	}
+	dsr := service.Spec.ExternalTrafficPolicy == v1.ServiceExternalTrafficPolicyTypeLocal
+	err = hcn.Slash32EndpointPrefixesSupported()
```
JocelynBerrendonner, line 357: We need to use DSRSupported() here.

elweb9858: @JocelynBerrendonner I am using the Slash32EndpointPrefixesSupported() method here because DSRSupported() does not reflect the backport of the dsr functionality. Some users need this functionality on rs5, which DSRSupported() will inaccurately say is not supported. I can update the version check in hcsshim (assuming the preserveDIP functionality has also been backported) and file a bug to change this version check the next time an hcsshim release is vendored into k8s

JocelynBerrendonner: @elweb9858 : why not do the proper fix in hcsshim today, merge it and pick it up here later today or tomorrow?

JocelynBerrendonner: FWIW, the hcsshim change should be a two lines change in this file: https://github.com/microsoft/hcsshim/blob/27a858bf1651e8204c64b5e6f9979f6c51a0f943/hcn/hcnglobals.go#L41

pkg/proxy/winkernel/proxier.go
```diff
@@ -352,6 +353,11 @@ func (proxier *Proxier) newServiceInfo(port *v1.ServicePort, service *v1.Service
 	if err != nil {
 		preserveDIP = false
 	}
+	dsr := service.Spec.ExternalTrafficPolicy == v1.ServiceExternalTrafficPolicyTypeLocal
```
ksubrmnn, line 356: You need to also check if the feature flag is enabled as DSR is still an alpha feature.

pkg/proxy/winkernel/proxier.go
elweb9858: @JocelynBerrendonner you make a fair point, I just want to make sure we are not breaking existing scenarios that use DSRSupported()

JocelynBerrendonner: They shouldn't. Fixing the Hcsshim should actually help everyone. Let's take the conversation offline.

pkg/proxy/winkernel/proxier.go
elweb9858: @ksubrmnn Isn't that feature gate tied to the preserveDIP implementation of DSR? AFAIK this is a separate implementation based on the existing ExternalTrafficPolicy feature in the service spec. @madhanrm correct me if I'm wrong

ksubrmnn: No it is not. That flag is for DSR, not preserve dip. PreserveDIP is not an implementation of DSR. Rather, it depends on DSR.

However, you are right in that we won't need to guard with a flag for the pre-existing ExternalTrafficPolicy feature. So this is a separate implementation.

For 1.20, you can follow up with a PR to add DSR functionality for the other loadbalancers if the feature flag is enabled.

ksubrmnn: Please rename this variable so that it is not confused with proxier.isDSR

pkg/proxy/winkernel/proxier.go
ksubrmnn: Since vendoring in new dependencies does take time, I think it's acceptable to use the version checks we have today (as long as they work properly) and update the shim for future releases.

pkg/proxy/winkernel/proxier.go
elweb9858: > For 1.20, you can follow up with a PR to add DSR functionality for the other loadbalancers if the feature flag is enabled.

I'm not quite following on this. Could you explain what you mean here?

Also, I will change the variable to localTrafficDSR

ksubrmnn: DSR is its own feature. PreserveDIP depends on DSR.

We have enabled DSR to work for all types of loadbalancers on the platform. However, external traffic policy only applies to loadbalancers of type Loadbalancer and of type NodePort. This PR also needs to be amended to honor that.

To add DSR to the other loadbalancer types, we would need to first check the feature flag. Then we would enable it accordingly.

ksubrmnn: Looks like Kubernetes will block on yaml deployment if the incorrect loadbalancer type is specified with externalTrafficPolicy local so you won't have to worry about that after all.

pkg/proxy/winkernel/proxier.go
ksubrmnn: Let's follow up in a separate PR to use proxier.isDSR for these loadbalancers as well.

pkg/proxy/winkernel/proxier.go
```diff
@@ -1131,7 +1138,7 @@ func (proxier *Proxier) syncProxyRules() {
 
 		hnsLoadBalancer, err := hns.getLoadBalancer(
 			hnsEndpoints,
-			loadBalancerFlags{isDSR: proxier.isDSR, isIPv6: proxier.isIPv6Mode, sessionAffinity: sessionAffinityClientIP},
+			loadBalancerFlags{isDSR: proxier.isDSR || svcInfo.dsr, isIPv6: proxier.isIPv6Mode, sessionAffinity: sessionAffinityClientIP},
```
ksubrmnn, line 1141: I believe that we should not be modifying this loadbalancer as externalTrafficPolicy: Local only affects the nodeport loadbalancer. Same for below.

pkg/proxy/winkernel/proxier.go
JocelynBerrendonner: Thanks @elweb9858 ! The latest iteration looks good.

JocelynBerrendonner (approved):
LGTM

ksubrmnn (approved):

pkg/proxy/winkernel/proxier.go
liggitt: > Let's follow up in a separate PR to use proxier.isDSR for these loadbalancers as well.

would that follow-up PR be required to be in 1.19 as well? as noted, we are past code freeze

ksubrmnn: No that is for 1.20. This PR is now only updating a dependency which contains a fix for Windows 2004

elweb9858: @liggitt this is no longer in scope for this PR and v1.20. Originally was trying to implement externalTrafficPolicy local in winkernel kube-proxy in this PR, but we are going to hold off on that and the followup work mentioned here until v1.20. This PR is now just simply vendoring in a version check update from hcsshim

pkg/proxy/winkernel/proxier.go
elweb9858: Resolving this as out of scope for the PR. We are continuing this conversation offline for future changes

pkg/proxy/winkernel/proxier.go
liggitt: ack

liggitt:
/approve
for dependency mechanics

/hold
would like a clean CI run once https://github.com/kubernetes/kubernetes/pull/93063 merges, then can unhold

liggitt:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/93080/pull-kubernetes-node-e2e/1283460130247020544 flake is #93210

liggitt:
@elweb9858 can you add a link to the bug this is fixing?

liggitt:
@michmike can cancel the hold on this once the two azure jobs pass

liggitt:
opened https://github.com/kubernetes/kubernetes/issues/93295 for CI failure

liggitt:
azure jobs are green

/hold cancel

daschott:
@liggitt here is the bug this is fixing: https://github.com/kubernetes/kubernetes/issues/93304

There are many other bugs that need the DSR mode to be properly resolved, such as https://github.com/Azure/AKS/issues/1375

PR #91399, opened 2020-05-25
multiple IPv6/dual-stack endpoint fixes

danwinship:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Fixes two problems with endpoints in dual-stack and single-stack IPv6 clusters:
1. Fixes #92965; EndpointSliceController was only looking at IPv4 Pod IPs during Pod updates, so it wouldn't notice when a Pod went from having no IPs to having only an IPv6 IP. (May also fix some other problems where certain Pod updates were not noticed in one or both controllers, because of the odd way they were trying to figure out whether a Pod update would result in Endpoint updates. Eg, neither controller would notice if a pod went from single-stack to dual-stack, though it's possible that never happens in practice.)
2. EndpointController now allows headless services to have IPv6 endpoints in single-stack IPv6 and dual stack clusters, as long as they specify `ipFamily: IPv6`. (EndpointSliceController already allowed this.)

However, this preserves the existing behavior (in both controllers) that "legacy" headless services (with no `IPFamily` set) are treated as IPv4-only in all clusters when `IPv6DualStack` is enabled. (Even if the cluster is single-stack IPv6.) ~I had thought we could fix this easily because I thought kube-controller-manager knew the service CIDR(s). But in fact, it usually doesn't; kcm has a `--service-cluster-ip-range` argument, but you only need to specify it if (a) you are using `--allocate-node-cidrs` and (b) your service CIDR is _inside_ the cluster CIDR so you need the Node CIDR allocator to skip over that range. (Srsly???). So anyway, most kcm's are not going to know what the default service CIDR is.~

~There is another approach we could take, which is to take advantage of the fact that the endpoint controller will, shortly after startup, learn whether `kubernetes.default` is IPv4 or IPv6, which is the same as learning whether the default service CIDR is IPv4 or IPv6. But I think we'd need to rearrange the code to ensure that we find that out before queuing any other updates, since if we got an informer notification for a legacy headless service before we got a notification for `kubernetes.default`, we wouldn't be able to process it yet.~

Given that 1.20 will fix the behavior of headless dual-stack services, it's not worth going to any further effort to improve things in 1.19.

Fixes #92965

**Does this PR introduce a user-facing change?**:
```release-note
Fixed the EndpointSliceController to correctly create endpoints for IPv6-only pods.

Fixed the EndpointController to allow IPv6 headless services, if the IPv6DualStack
feature gate is enabled, by specifying `ipFamily: IPv6` on the service. (This already
[...]

PR #93089, opened 2020-07-15
Fixing a bug with EndpointSlice controller IPv6 change detection

### LINKED ISSUE #92965: [Failing Test] kind-ipv6-master-parallel (ci-kubernetes-kind-ipv6-e2e-parallel)
**Which jobs are failing**:

`kind-ipv6-master-parallel (ci-kubernetes-kind-ipv6-e2e-parallel)`

`[sig-network] Services should create endpoints for unready pods` is failing consistently.

`07-09-2020 11:39 CDT`

**Testgrid link**:

https://testgrid.k8s.io/sig-release-master-blocking#kind-ipv6-master-parallel

**Reason for failure**:

```
test/e2e/network/service.go:1979
Jul 10 05:27:25.533: expected un-ready endpoint for Service slow-terminating-unready-pod within 5m0s, stdout: 
test/e2e/network/service.go:2061
```

/cc @kubernetes/ci-signal 
/sig network
/priority critical-urgent
/milestone v1.19

robscott:
**What type of PR is this?**
/kind bug
/kind failing-test

The EndpointSlice controller only considers IPs that match the IP family of the Service when it comes to converting Pods to endpoints. That works well during a reconcile loop when the Service is provided, but does not work during Pod change detection when the Service(s) are unknown. This PR updates that function to ignore IP family altogether if
service.Name is unspecified, something that should only happen during Pod change detection.

This surfaced in 1 specific Kind e2e IPv6 test that involved the initial syncService call happening before the Pod had an IP address assigned. Unlike other tests, this relied on the Pod change functionality working, not just another syncService call eventually resolving things.

This commit also reworks the related unit tests as they were unfortunately not working as expected in their current state.

Fixes https://github.com/kubernetes/kubernetes/issues/92965

I actually was really close to having this fixed yesterday but then went down a DNS rabbit hole due to the test still failing locally for me with the `cluster.local` suffix. I believe that is an issue unique to my local Kind setup, but I'm filing this PR to be sure.

```release-note
Fixed a bug with EndpointSlice controller IPv6 change detection.
```

/sig network
/priority critical-urgent
/cc @aojea @danwinship @hasheddan  
/assign @freehan

### DIFF
--- a/pkg/controller/endpointslice/utils.go
+++ b/pkg/controller/endpointslice/utils.go
@@ -128,12 +128,19 @@ func getEndpointPorts(service *corev1.Service, pod *corev1.Pod) []discovery.Endp
 }
 
 // getEndpointAddresses returns a list of addresses generated from a pod status.
+// There are 2 similar but distinct uses for this function:
+// 1. Get the IPs from a Pod that should be added to EndpointSlices.
+// 2. Get the IPs from a Pod to see if they are different from a previous Pod.
 func getEndpointAddresses(podStatus corev1.PodStatus, service *corev1.Service) []string {
 	addresses := []string{}
 
 	for _, podIP := range podStatus.PodIPs {
 		isIPv6PodIP := utilnet.IsIPv6String(podIP.IP)
-		if isIPv6PodIP == isIPv6Service(service) {
+		// If the service.Name is empty we know this method is only being used
+		// to determine if IP addresses have changed for a Pod and we don't have
+		// access to the full Service. In that case we can't determine if the IP
+		// family matches and it doesn't matter for the purposes of diffing.
+		if isIPv6PodIP == isIPv6Service(service) || service.Name == "" {
 			addresses = append(addresses, podIP.IP)
 		}
 	}
--- a/pkg/controller/endpointslice/endpointslice_controller.go
+++ b/pkg/controller/endpointslice/endpointslice_controller.go
@@ -237,7 +237,7 @@ func (c *Controller) Run(workers int, stopCh <-chan struct{}) {
 	klog.Infof("Starting endpoint slice controller")
 	defer klog.Infof("Shutting down endpoint slice controller")
 
-	if !cache.WaitForNamedCacheSync("endpoint_slice", stopCh, c.podsSynced, c.servicesSynced) {
+	if !cache.WaitForNamedCacheSync("endpoint_slice", stopCh, c.podsSynced, c.servicesSynced, c.nodesSynced, c.endpointSlicesSynced) {
 		return
 	}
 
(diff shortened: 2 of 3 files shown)


pkg/controller/endpointslice/utils.go
```diff
@@ -128,12 +128,19 @@ func getEndpointPorts(service *corev1.Service, pod *corev1.Pod) []discovery.Endp
...
 	addresses := []string{}
 
 	for _, podIP := range podStatus.PodIPs {
 		isIPv6PodIP := utilnet.IsIPv6String(podIP.IP)
-		if isIPv6PodIP == isIPv6Service(service) {
+		// If the service.Name is empty we know this method is only being used
+		// to determine if IP addresses have changed for a Pod and we don't have
+		// access to the full Service. In that case we can't determine if the IP
+		// family matches and it doesn't matter for the purposes of diffing.
+		if isIPv6PodIP == isIPv6Service(service) || service.Name == "" {
```
liggitt, line 143: would it be clearer to pass a nil service than a non-nil dummy service?

`podToEndpoint` already does nil checks for the node object that is passed in... should it do the same for the service object, and then pass nil for node and service in `podEndpointChanged`?

pkg/controller/endpointslice/utils.go
```diff
@@ -128,12 +128,19 @@ func getEndpointPorts(service *corev1.Service, pod *corev1.Pod) []discovery.Endp
...
 			addresses = append(addresses, podIP.IP)
```
liggitt, line 144: this can result in returning an addresses list containing a mix of ipv6 and ipv4 addresses... is that ok?

danwinship:
Compare #91399, particularly [the first commit](https://github.com/kubernetes/kubernetes/pull/91399/commits/ef207c6c59034e05854d7e8999d42773105d7c18). It changes both endpoint controllers so that they don't consider Services/Nodes/etc at all when deciding whether a pod's endpoints have changed, which I think is more correct.

danwinship:
(Sorry, probably should have mentioned #91399 before, but I thought the only actual bug it fixed was "if the secondary IP changes and the primary IP doesn't, we miss the update".)

pkg/controller/endpointslice/utils.go
aojea: I think no, the endpointlsice has an `addressType:` field so each slices only have endpoints from the same IPFamily

hasheddan:
@aojea done :heavy_check_mark:

robscott:
Talked with @danwinship on Slack, sounds like it's reasonable to cherry-pick #91399, so I'll go ahead and close this PR.

/close

liggitt:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/93089/pull-kubernetes-node-e2e/1283211616950685696 flake was #93210

PR #93092, opened 2020-07-15
[apiserver/metrics] reset: fix range issue

gaurav1086:

**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Go uses the same address variable while iterating in a range, so any change done on that address only affects the last value: 
https://play.golang.com/p/_Fym_BNigTo
To fix this, use the original address by accessing through array index.

Fixes #

https://play.golang.com/p/_Fym_BNigTo

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go
+++ b/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go
@@ -256,8 +256,8 @@ func Register() {
 
 // Reset all metrics.
 func Reset() {
-	for _, metric := range metrics {
-		metric.Reset()
+	for i := range metrics {
+		metrics[i].Reset()
 	}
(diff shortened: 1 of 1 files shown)


logicalhan:
I don't think this is doing what you believe this is doing..

https://play.golang.com/p/VB_cEket3xF

logicalhan:
> I don't think this is doing what you believe this is doing..

https://play.golang.com/p/vdJpT16JLeY

logicalhan:
Also a change like this should probably have a test case accompanying it, you want to demonstrate the situation that you believe is faulty and the same test should should that the change does, in fact, fix it.

PR #93103, opened 2020-07-15
WIP - Add URL support in APIService spec

### LINKED ISSUE #92374: Add URL support in APIService

The [aggregation layer](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/) allows to extend the KAS with additional APIs through APIService resources. Currently external extensions API servers can only be referenced through [ServiceReferences](https://github.com/kubernetes/kubernetes/blob/5e0bd8a44c129f513b42a6b86799ef92bfe11ef2/staging/src/k8s.io/kube-aggregator/pkg/apis/apiregistration/v1beta1/types.go#L55) and not by directly specifying a URL. I propose to add URL support in APIService resource consistently with what is done for [Webhook](https://github.com/kubernetes/kubernetes/blob/5e0bd8a44c129f513b42a6b86799ef92bfe11ef2/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1/types.go#L157) configuration.

**Why is this needed**:

Starting from k8s 1.18 it is possible to setup [apiserver-network-proxy](https://github.com/kubernetes-sigs/apiserver-network-proxy) for scenarios where `control` and `cluster` networks are disjoint. The requests destined for the services use the `cluster` context. This means that it is currently not supported to have API extensions running in the `control` network. Adding URL support to APIService could enable having API extension servers on the 'master' network.

References:

- https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190226-network-proxy.md#definitions

- https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190226-network-proxy.md#kubernetes-api-server-outbound-requests

irozzo-1A:
**What type of PR is this?**
/kind feature

Fixes #92374

This PR is not implementing the functionality, but just for getting a review on the proposed API change.

```release-note

```

```docs

```


staging/src/k8s.io/kube-aggregator/pkg/apis/apiregistration/types.go
```diff
@@ -83,6 +83,32 @@ type APIServiceSpec struct {
...
+	//
+	// Attempting to use a user or basic auth e.g. "user:password@" is not
+	// allowed. Fragments ("#...") and query parameters ("?...") are not
+	// allowed, either.
+	//
+	// When egress configuration is given `master` network context is used
+	// to dial API server calls.
+	//
+	// +optional
+	URL *string
```
lavalamp, line 111: The path handling can't be the same as for webhooks, since we proxy to it. What should we do with the path? Treat it as a prefix, i.e. prepend it to the proxied path? Require there not be one?

@liggitt @deads2k

liggitt, line 111: cc @tallclair @cjcullen on allowing API-control of arbitrary proxy destinations for arbitrary requests

irozzo-1A: > The path handling can't be the same as for webhooks, since we proxy to it. What should we do with the path? Treat it as a prefix, i.e. prepend it to the proxied path? Require there not be one?

I was thinking to prepend it, but actually I don't have in mind a use-case where it is actually useful. I agree with @liggitt that requiring no path is probably better.

lavalamp: I can imagine it being useful, and if we ever end up with clients talking to the AA directly, we'd likely start by redirecting them, so we could make that work either way. @deads2k runs a lot of AAs, maybe he can tie-break.

liggitt: > if we ever end up with clients talking to the AA directly, we'd likely start by redirecting them

xref https://github.com/kubernetes/kubernetes/issues/93129 :)

irozzo-1A:
@lavalamp @liggitt I changed not to support path in the URL, I think it's easier introducing support in case it's needed than removing it later.

PR #93108, opened 2020-07-15
fix HandleCrash() order

lixiaobing1:
What type of PR is this?
/kind bug

HandleCrash() is a final handling func, after HandleCrash(), the process will terminate, and the funcs before it will not be handled, thus should put HandleCrash() to the first line:

func (sw *StreamWatcher) receive() {
	defer close(sw.result)
	defer sw.Stop()

```release-note
   fix HandleCrash order
```

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/watch/streamwatcher.go
+++ b/staging/src/k8s.io/apimachinery/pkg/watch/streamwatcher.go
@@ -97,9 +97,9 @@ func (sw *StreamWatcher) stopping() bool {
 
 // receive reads result from the decoder in a loop and sends down the result channel.
 func (sw *StreamWatcher) receive() {
+	defer utilruntime.HandleCrash()
 	defer close(sw.result)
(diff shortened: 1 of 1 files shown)


jpbetz:
/lgtm

This looks correct.

Out of curiosity, was there an actual issue that motivated fixing this or was it just a observed defect while looking over code?

lixiaobing1:
> /lgtm
> 
> This looks correct.
> 
> Out of curiosity, was there an actual issue that motivated fixing this or was it just a observed defect while looking over code?

lixiaobing1:
@jpbetz find it while looking over code,  at the same time I found some code had changed in another place:

func (le *LeaderElector) Run(ctx context.Context) {
	defer func() {
		runtime.HandleCrash()
		le.config.Callbacks.OnStoppedLeading()
	}()

had changed to :

	defer runtime.HandleCrash()
	defer func() {
		le.config.Callbacks.OnStoppedLeading()
	}()

Issue #93112, opened 2020-07-15
[sig-scheduling] IPAM / Scheduler Performance tests fail on timeout


hasheddan:
**Which jobs are failing**:

`integration-master (ci-kubernetes-integration-master)`

**Which test(s) are failing**:

`k8s.io/kubernetes/test/integration/ipamperf.TestMain`
`k8s.io/kubernetes/test/integration/scheduler_perf.TestMain`

**Since when has it been failing**:

`07-12-20 00:15 CDT`

**Testgrid link**:

https://testgrid.k8s.io/sig-release-master-blocking#integration-master

**Reason for failure**:

These tests started running after https://github.com/kubernetes/test-infra/pull/18196 caused `integration-master` to run tests that had been marked to skip unless `--short=false`. This change was made to avoid running some other tests when `--short=true` in https://github.com/kubernetes/kubernetes/pull/92879. We need to establish how to adjust timeouts on these tests such that they will be able to run successfully. They will be disabled in the mean time since they were not running previously.

**Anything else we need to know**:

/cc @kubernetes/ci-signal @liggitt @kubernetes/sig-scheduling-misc 
/milestone v1.19
/priority important-soon
/sig scheduling

ahg-g:
Those are non-critical scheduler perf debugging tests, the running time will increase with each test we add, it will also depend on the machine where this is running. 

I think we should just run them on the vanilla workloads, so it is fine to keep it disabled for now and handle this issue in 1.20

liggitt:
sounds good

Huang-Wei:
+1 to what @ahg-g said above. Additionally, this test was firstly built 5 years ago, the internal mechanics (like using a `for..sleep` loop to calculate the metrics) are not that concise, other aspects also need to be ported to latest scheduler architecture.

So this is not a must for 1.19, let's see how to reintroduce it in 1.20.

ahg-g:
/assign @adtac 

the scheduler_perf tests should run on CI, but only few of them (e.g., only the 500node tests).

ahg-g:
can we rename this issue to only focus on the scheduler and track the test related to IPAM in another issue?

PR #93113, opened 2020-07-15
Skip failing scheduler integration tests until they can be run successfully

hasheddan:
**What type of PR is this?**

/kind failing-test

The IPAM and scheduler performance tests are currently causing
integration-master job to fail because of timeouts. They were not
previously running as part of integration-master, so we can disable them
without loss of test coverage. They should be re-enabled as part of fix
for #93112.

Ref: https://github.com/kubernetes/kubernetes/pull/92879 https://github.com/kubernetes/test-infra/pull/18196

This is a short-term solution to get `integration-master` passing again and should be followed up by fixing #93112 

/cc @kubernetes/ci-signal @kubernetes/sig-scheduling-misc @liggitt @BenTheElder 

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/integration/ipamperf/ipam_test.go
+++ b/test/integration/ipamperf/ipam_test.go
@@ -108,7 +108,8 @@ func logResults(allResults []*Results) {
 }
 
 func TestPerformance(t *testing.T) {
-	if testing.Short() {
+	// TODO (#93112) skip test until appropriate timeout established
+	if testing.Short() || true {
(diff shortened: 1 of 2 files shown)


liggitt:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/93113/pull-kubernetes-node-e2e/1283567145829011456 flake was #93210

PR #76310, opened 2019-04-09
Relax namespace restriction for critical pods

ravisantoshgudimetla:
**What type of PR is this?**

 /kind bug

**What this PR does / why we need it**:

I think, we should relax the restriction on critical pods to be created within `kube-system`, I have outlined an approach which ensures that, initial quota will be restricted to `kube-system` namespace which can later be expanded to other namespaces(https://github.com/kubernetes/kubernetes/pull/76310/files#diff-4e1245d9faac6e47c01c1c068f4de517R199). Let me know, if I have missed something.

**Which issue(s) this PR fixes**:

Fixes https://github.com/kubernetes/kubernetes/issues/76308

**Special notes for your reviewer**:
/cc @bsalamat @vikaschoudhary16 

**Does this PR introduce a user-facing change?**:

```release-note
Critical pods can now be created in namespaces other than kube-system. To limit critical pods to the kube-system namespace, cluster admins should create an admission configuration file limiting critical pods by default, and a matching quota object in the `kube-system` namespace permitting critical pods in that namespace. See https://kubernetes.io/docs/concepts/policy/resource-quotas/#limit-priority-class-consumption-by-default for details.
```
Ref: [document](https://kubernetes.io/docs/concepts/policy/resource-quotas/#limit-priority-class-consumption-by-default)

PR #93121, opened 2020-07-15
kube-up: limit critical pods to kube-system by default

liggitt:
**What type of PR is this?**
/kind cleanup

Implements the approach described in https://github.com/kubernetes/kubernetes/pull/76310#discussion_r344902842 for kube-up (xref https://kubernetes.io/docs/concepts/policy/resource-quotas/#limit-priority-class-consumption-by-default)

```release-note
kube-up: defaults to limiting critical pods to the kube-system namespace to match behavior prior to 1.17
```

/sig scheduling
/cc @ahg-g


liggitt:
moved to addons set so it can be applied in the kube-system namespace

liggitt:
performance failure looks like a pre-existing flake, the same failures exist in runs on other PRs that don't touch any files related to cluster setup

/retest

PR #93125, opened 2020-07-15
[plugin/admission_test] fix range issue in TestPodSecurityContextAdmission

gaurav1086:

**What type of PR is this?**
>
> /kind api-change
 /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Go uses the same address variable while iterating in a range, so use a copy when using its address.
https://play.golang.org/p/wh5_P3NpZXl

Fixes #
N/A
N/A

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/plugin/pkg/admission/securitycontext/scdeny/admission_test.go
+++ b/plugin/pkg/admission/securitycontext/scdeny/admission_test.go
@@ -139,8 +139,8 @@ func TestPodSecurityContextAdmission(t *testing.T) {
 			errorExpected: true,
 		},
 	}
-	for _, test := range tests {
-		pod.Spec.SecurityContext = &test.securityContext
+	for i, test := range tests {
+		pod.Spec.SecurityContext = &tests[i].securityContext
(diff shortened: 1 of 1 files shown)


plugin/pkg/admission/securitycontext/scdeny/admission_test.go
```diff
@@ -139,8 +139,8 @@ func TestPodSecurityContextAdmission(t *testing.T) {
 			errorExpected: true,
 		},
 	}
-	for _, test := range tests {
-		pod.Spec.SecurityContext = &test.securityContext
+	for i, test := range tests {
+		pod.Spec.SecurityContext = &tests[i].securityContext
```
derekwaynecarr, line 143: i am missing why this change is needed?

the assignment and validation happens in each iteration of the loop.

gaurav1086: @derekwaynecarr, my bad. Here's the sample illustration. https://play.golang.org/p/wh5_P3NpZXl . Update the PR description as well.

gaurav1086:
cc @liggitt

gaurav1086:
cc @lavalamp

lavalamp:
The code would be wrong if the reference escaped the loop, but it doesn't, so there's no need to change this.

gaurav1086:
@lavalamp thanks for clarification. closing this.

PR #93127, opened 2020-07-15
[test e2e/pods] fix range address issue in WaitForErrorEventOrSuccess

gaurav1086:

**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Go uses the same address variable while iterating in a range, so use a copy when using its address.
https://play.golang.org/p/wh5_P3NpZXl

Fixes #
N/A
https://play.golang.org/p/wh5_P3NpZXl

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/test/e2e/framework/pods.go
+++ b/test/e2e/framework/pods.go
@@ -237,10 +237,10 @@ func (c *PodClient) WaitForErrorEventOrSuccess(pod *v1.Pod) (*v1.Event, error) {
 		if err != nil {
 			return false, fmt.Errorf("error in listing events: %s", err)
 		}
-		for _, e := range evnts.Items {
+		for i, e := range evnts.Items {
 			switch e.Reason {
 			case killingContainer, failedToCreateContainer, forbiddenReason:
-				ev = &e
+				ev = &evnts.Items[i]
 				return true, nil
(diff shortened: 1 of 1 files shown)


test/e2e/framework/pods.go
```diff
@@ -237,10 +237,10 @@ func (c *PodClient) WaitForErrorEventOrSuccess(pod *v1.Pod) (*v1.Event, error) {
 		if err != nil {
 			return false, fmt.Errorf("error in listing events: %s", err)
 		}
-		for _, e := range evnts.Items {
+		for i, e := range evnts.Items {
 			switch e.Reason {
 			case killingContainer, failedToCreateContainer, forbiddenReason:
-				ev = &e
```
liggitt, line 243: since this was returning immediately, rather than accumulating multiple pointers to various indices, the current use is fine

liggitt:
the current use is not problematic, see https://github.com/kubernetes/kubernetes/pull/93127#discussion_r465230359

if we were accumulating multiple items into a list of return values, this change would be needed

/close

Issue #86448, opened 2019-12-19
CSI drivers need some way to get service account tokens


mikedanese:
CSI drivers need some way to get service account tokens for the pods they are provisioning volumes for. These tokens may be used to authenticate to the Kubernetes API (e.g. the [Certificates API](https://github.com/jetstack/cert-manager-csi), or to authenticate to external services like [vault](https://github.com/hashicorp/secrets-store-csi-driver-provider-vault)). Right now the state of the art is to give the CSI driver acccess to the filesystem. This is suboptimal because:
* It's not straightforward but security critical. Every driver has to re-implement this.
* It violates least privilege. CSI drivers should not be able to read secrets of pods that don't use them.
* The token is not guaranteed to be available (e.g. it won't be if automountServiceAccountToken=false).

The kubelet VolumeHost is wired up to fetch new-style service account tokens:

https://github.com/kubernetes/kubernetes/blob/9fad2e320104cc44d5631203c0a7085dedbaf35c/pkg/volume/plugins.go#L433

We can pretty easily wire up the CSI plugin to pass audience bound tokens to CSI drivers. If we like this approach, then we need an API design for:
* CSI configuration. How do I configure a driver that (1)  needs a service account token and (2) needs the token to have a specific audience?
  * Some options are configuration on the global CSIDriver object or at the pod level.
* CSI runtime: How does the kubelet pass the token to the driver?
  * This will probably look something like the other [attributes](https://github.com/kubernetes/kubernetes/blob/9f841bf42396daf1ed4b7ab47c1e1708bfb028b3/pkg/volume/csi/csi_mounter.go#L220-L228) we pass.

/sig storage
/sig auth

@kubernetes/sig-auth-feature-requests @kubernetes/sig-storage-feature-requests 
@seh @kfox1111 @munnerz

seh:
As an intermediate hack, following a suggestion posed by @tallclair, my CSI driver uses the _TokenRequest_ API to fetch a service account token for the pod into which it's mounting a volume. However, as discussed in the SIG Auth meeting on 11 December 2019, doing so is only really safe if we also use an admission plugin that verifies that the token request both comes from the CSI driver's service account and that the pertinent pod is indeed mounting a volume provided by that CSI driver.

It works, but it's messy.

kfox1111:
should such a check be added to the node admission plugin?

seh:
[...]

PR #93130, opened 2020-07-16
plumb service account token down to csi driver

zshihang:
**What type of PR is this?**
> /kind api-change

KEP: https://github.com/kubernetes/enhancements/pull/1855
Issue: kubernetes/kubernetes#86448
release: https://github.com/kubernetes/enhancements/issues/2047

```release-note
this PR will introduce a feature gate CSIServiceAccountToken with two additional fields in `CSIDriverSpec`.
```

```docs
- [KEP]: https://github.com/kubernetes/enhancements/pull/1855
```


pkg/apis/storage/types.go
```diff
@@ -337,6 +337,35 @@ type CSIDriverSpec struct {
 	//
 	// +optional
 	StorageCapacity *bool
+
+	// ServiceAccountTokens indicates the csi driver needs pods' service account
+	// tokens it is mounting volume for to do necessary authentication. Kubelet
+	// will pass the tokens in VolumeContext in the CSI NodePublishVolume calls.
+	// The CSI driver should parse and validate the following VolumeConext:
+	// "csi.storage.k8s.io/serviceAccount.tokens": {
+	//   'audience-0': {
```
msau42, line 346: Is "audience-0" actually the name of the key, or is it some other value?

pkg/apis/storage/types.go
```diff
@@ -337,6 +337,35 @@ type CSIDriverSpec struct {
...
+	// most one token can ignore Audience.
+	//
+	// +optional
+	// +listType=set
+	ServiceAccountTokens []ServiceAccountToken
+
+	// RequiresRemount indicates the csi driver wants `NodePublishVolume` being
+	// periodically called to reflect any possible change in the mounted volume.
+	// +optional
+	RequiresRemount *bool
```
msau42, line 368: To align with CSI terminology, I would call this something like `RequiresRepublish`. I know the code name is different but I don't think we need to expose internal code function names in the API.

pkg/apis/storage/types.go
```diff
@@ -337,6 +337,35 @@ type CSIDriverSpec struct {
...
+	//
+	// Note: Audience in each ServiceAccountToken should be different and at
+	// most one token can ignore Audience.
+	//
+	// +optional
+	// +listType=set
+	ServiceAccountTokens []ServiceAccountToken
+
+	// RequiresRemount indicates the csi driver wants `NodePublishVolume` being
+	// periodically called to reflect any possible change in the mounted volume.
```
msau42, line 366: How often is it called? Is it tunable?

pkg/apis/storage/types.go
```diff
@@ -337,6 +337,35 @@ type CSIDriverSpec struct {
...
+	// }
+	//
+	// Note: Audience in each ServiceAccountToken should be different and at
+	// most one token can ignore Audience.
+	//
+	// +optional
+	// +listType=set
+	ServiceAccountTokens []ServiceAccountToken
+
+	// RequiresRemount indicates the csi driver wants `NodePublishVolume` being
```
msau42, line 365: Is it required for drivers to enable this if they specify service account tokens?

pkg/apis/storage/types.go
msau42, line 366: Also specify the default

pkg/kubelet/volumemanager/cache/actual_state_of_world.go
```diff
@@ -832,6 +833,25 @@ func (asw *actualStateOfWorld) newAttachedVolume(
 	}
 }
 
+func (asw *actualStateOfWorld) requiresRemount(spec *volume.Spec) bool {
```
msau42, line 836: Should this be implemented somewhere inside https://github.com/kubernetes/kubernetes/blob/0c4e1d04dfce877cf8a8e11b62e81eabf9df62a4/pkg/volume/csi/csi_plugin.go.

The logic in this file is not specific to csi.

pkg/registry/storage/csidriver/strategy_test.go
```diff
@@ -222,116 +244,123 @@ func TestCSIDriverPrepareForUpdate(t *testing.T) {
...
+			Name: "foo",
+		},
+		Spec: storage.CSIDriverSpec{
+			ServiceAccountTokens: []storage.ServiceAccountToken{{Audience: &vault}},
+			RequiresRemount:      &enabled,
+		},
+	}
 
-	var resultEmpty []storage.VolumeLifecycleMode
+	// var resultEmpty []storage.VolumeLifecycleMode
```
msau42, line 266: how come we got rid of these?

pkg/volume/csi/csi_plugin.go
```diff
@@ -182,6 +184,7 @@ func (h *RegistrationHandler) DeRegisterPlugin(pluginName string) {
 
 func (p *csiPlugin) Init(host volume.VolumeHost) error {
 	p.host = host
+	p.serviceAccountTokenGetter = host.GetServiceAccountTokenFunc()
```
msau42, line 187: I guess this was here before, but since this function is only implemented for kubelet, it could be in the KubeletVolumeHost instead of the general VolumeHost.

pkg/apis/storage/types.go
msau42: I think we also need to make clear here that drivers would need to repopulate the content, but not actually do a remount. Because CRI has already bind mounted the mount point, and a new mount point cannot be updated without a container restart.

And in that case, do we also need to pass in an attribute to the driver to tell them that this is a republish vs the 1st time mounting? cc @jsafrane for thoughts.

pkg/volume/csi/csi_mounter.go
```diff
@@ -238,14 +240,15 @@ func (c *csiMountMgr) SetUpAt(dir string, mounterArgs volume.MounterArgs) error
...
 		return volumetypes.NewTransientOperationFailure(log("mounter.SetUpAt failed to assemble volume attributes: %v", err))
 	}
-	if podAttrs != nil {
-		if volAttribs == nil {
-			volAttribs = podAttrs
-		} else {
-			for k, v := range podAttrs {
-				volAttribs[k] = v
-			}
+	volAttribs = mergeMap(volAttribs, podAttrs)
```
msau42, line 243: A side effect that we need to think about is since this function will be called much more frequently for the plugins that enable it, are there any log statements we need to conditionally not log?

pkg/apis/storage/types.go
zshihang: it is a value that user specifies.

pkg/apis/storage/types.go
zshihang: <= 10/s. it is the same frequency as the reconciler loop.
https://github.com/zshihang/kubernetes/blob/ebb6995f3e5fc696596502fbf861fc68c09efe4e/pkg/kubelet/volumemanager/volume_manager.go#L57-L59

pkg/apis/storage/types.go
zshihang: added a note.

pkg/apis/storage/types.go
zshihang: no, it is separate.

pkg/apis/storage/types.go
zshihang: done

pkg/registry/storage/csidriver/strategy_test.go
zshihang: oops. i accidentally commented them out.

pkg/volume/csi/csi_plugin.go
zshihang: sorry, what is issue here? it is available in the general VolumeHost

pkg/kubelet/volumemanager/cache/actual_state_of_world.go
zshihang: done. RequiresRemount wasn't accepting `volume.Spec`.

pkg/volume/csi/csi_mounter.go
zshihang: looking at the SetupAt function which will be called more frequently: all informational logs are V(4) and other logs are all about errors.

pkg/volume/csi/csi_plugin.go
msau42: I meant since this is functionality is specific to kubelet (ie the other controllers don't implement this interface), then we could put it directly in KubeletVolumeHost instead of the generic VolumeHost.

pkg/apis/storage/types.go
msau42: Can we add some comments to which fields are fixed strings vs which are going to depend on user provider values? Maybe we can use the "<foo>" notation to denote which parts are going to be variable.

pkg/apis/storage/types.go
msau42: If they set the token to expire, is this required?

zshihang: the token always expires. we can requires `RequiresRepublish` if `ServiceAccountTokens` is not empty but not vice versa (i.e. user can specify `RequiresRepublish` only). WDYT?

pkg/apis/storage/types.go
```diff
@@ -337,6 +337,41 @@ type CSIDriverSpec struct {
...
+	//     'token': token,
+	//     'expiry': expiry,
+	//   },
+	//   ...
+	// }
+	//
+	// Note: Audience in each ServiceAccountToken should be different and at
+	// most one token can ignore Audience.
+	//
+	// +optional
```
msau42, line 361: Add a comment that this field is alpha-level and requires the feature gate to be set. Ditto everywhere

pkg/volume/csi/csi_mounter.go
msau42: I need to double check the log level for other messages in the remount path.

msau42: Looks like we use log level 5 for logs during remount: https://github.com/kubernetes/kubernetes/blob/e23d83eead3b5ae57731afb0209f4a2aaa4009dd/pkg/kubelet/volumemanager/reconciler/reconciler.go#L271

pkg/volume/csi/csi_plugin.go
zshihang: done

pkg/apis/storage/types.go
zshihang: done

pkg/apis/storage/types.go
zshihang: done

pkg/kubelet/cm/BUILD
```diff
@@ -96,11 +96,6 @@ go_library(
             "//staging/src/k8s.io/client-go/tools/record:go_default_library",
             "//vendor/k8s.io/utils/mount:go_default_library",
         ],
-        "@io_bazel_rules_go//go/platform:illumos": [
```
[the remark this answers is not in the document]
zshihang: yes, looks like there is some issue going on with `bazel_gazelle`. i'll try fix this

pkg/apis/storage/types.go
```diff
@@ -374,6 +409,18 @@ const (
...
 type VolumeLifecycleMode string
 
+// ServiceAccountToken contains parameters of a service account token.
+type ServiceAccountToken struct {
+	// Audience are the intendend audience of the token in "TokenRequestSpec",
+	// it will default to the audiences of kube apiserver if not specified.
+	// +optional
+	Audience *string
+
+	// ExpirationSeconds is the duration of validity of the token in "TokenRequestSpec"
```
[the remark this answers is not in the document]
zshihang: it will default to the default value of ExpirationSeconds in TokenRequest api. added a note.

pkg/kubelet/cm/BUILD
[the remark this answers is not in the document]
zshihang: done

test/e2e/storage/csi_mock_volume.go
```diff
@@ -1165,6 +1169,68 @@ var _ = utils.SIGDescribe("CSI mock volume", func() {
...
 		}
 	})
+
+	ginkgo.Context("CSIServiceAccountToken [Feature:CSIServiceAccountToken]", func() {
+		var (
+			err error
+		)
+		tests := []struct {
+			desc                   string
+			deployClusterRegistrar bool
```
msau42, line 1179: We don't support clusterregistrar anymore. Maybe rename this to "deployCSIDriverObject"?

pkg/apis/storage/types.go
msau42: Yes I think it would be a good idea to require this if they're using serviceaccount tokens.

msau42:
/approve
/label api-review

Mostly lgtm! Just some minor wording comments. We may need to revisit bumping log levels to 5 for remount if it turns out 4 is commonly used.

Will let @liggitt and @mikedanese also review.

test/e2e/storage/csi_mock_volume.go
```diff
@@ -1165,6 +1169,68 @@ var _ = utils.SIGDescribe("CSI mock volume", func() {
 			})
 		}
 	})
+
+	ginkgo.Context("CSIServiceAccountToken [Feature:CSIServiceAccountToken]", func() {
```
msau42, line 1173: For alpha feature tests, we'll need to update test-infra alpha job to run this.

pkg/apis/storage/types.go
zshihang: one thing to mention is that some csi drivers might only need the token once and do not care about whether the token expires later. we probably want to keep the flexibility here cause `RequiresRefresh` incurs extra overhead.

pkg/apis/storage/types.go
```diff
@@ -337,6 +337,47 @@ type CSIDriverSpec struct {
 	//
 	// +optional
 	StorageCapacity *bool
+
+	// ServiceAccountTokens indicates the csi driver needs pods' service account
```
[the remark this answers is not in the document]
zshihang: done

pkg/apis/storage/types.go
```diff
@@ -337,6 +337,47 @@ type CSIDriverSpec struct {
...
+	//
+	// +optional
+	// +listType=set
+	ServiceAccountTokens []ServiceAccountToken
+
+	// RequiresRepublish indicates the csi driver wants `NodePublishVolume`
+	// being periodically called to reflect any possible change in the mounted
+	// volume. This field defaults to false.
+	//
+	// Note: CSI drivers should only modify volumes. Mount point will not be
```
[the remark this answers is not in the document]
zshihang: done

pkg/registry/storage/csidriver/strategy.go
```diff
@@ -93,6 +98,13 @@ func (csiDriverStrategy) PrepareForUpdate(ctx context.Context, obj, old runtime.
 		newCSIDriver := obj.(*storage.CSIDriver)
 		newCSIDriver.Spec.FSGroupPolicy = nil
 	}
+	if old.(*storage.CSIDriver).Spec.ServiceAccountTokens == nil &&
```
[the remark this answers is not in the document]
zshihang: makes sense

pkg/volume/csi/csi_mounter.go
```diff
@@ -332,6 +336,61 @@ func (c *csiMountMgr) podAttributes() (map[string]string, error) {
...
 
+func (c *csiMountMgr) podServiceAccountTokenAttrs() (map[string]string, error) {
+	if c.plugin.serviceAccountTokenGetter == nil {
+		return nil, errors.New("ServiceAccountTokenGetter is nil")
+	}
+
+	csiDriver, err := c.plugin.csiDriverLister.Get(string(c.driverName))
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(4).Infof(log("CSIDriver %q not found, not adding service account token information", c.driverName))
```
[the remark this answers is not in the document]
zshihang: done

test/e2e/storage/csi_mock_volume.go
zshihang: done

pkg/apis/storage/types.go
msau42: Ok I guess if that's a valid use case then we don't need to require it. Can you add a comment to the API that RequiresRefresh is the way for a csi plugin to get the new token after it expires?

test/e2e/storage/csi_mock_volume.go
zshihang: https://github.com/kubernetes/test-infra/pull/19571

pkg/apis/storage/validation/validation.go
```diff
@@ -473,6 +475,44 @@ func validateFSGroupPolicy(fsGroupPolicy *storage.FSGroupPolicy, fldPath *field.
...
+func validateServiceAccountTokens(serviceAccountTokens []storage.ServiceAccountToken, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if serviceAccountTokens != nil && utilfeature.DefaultFeatureGate.Enabled(features.CSIServiceAccountToken) {
+		audiences := make(map[string]bool)
+		for _, serviceAccountToken := range serviceAccountTokens {
+			audience := ""
+			if serviceAccountToken.Audience != nil {
+				audience = *serviceAccountToken.Audience
+				if audience == "" {
+					allErrs = append(allErrs, field.Invalid(fldPath, audience, "empty string is not a valid audience"))
```
mikedanese, line 489: > empty string is not a valid audience

Doesn't "" mean audience is the kube-apiserver elsewhere?

pkg/apis/storage/types.go
```diff
@@ -337,6 +337,48 @@ type CSIDriverSpec struct {
...
+	//
+	// Note: After a successful initial NodePublishVolume call, subsequent calls
+	// to NodePublishVolume should only update the contents of the volume. New
+	// mount points will not be seen by a running container.
+	//
+	// This is an alpha feature and only available when the
+	// CSIServiceAccountToken feature is enabled.
+	//
+	// +optional
+	RequiresRepublish *bool
```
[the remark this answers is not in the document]
msau42: I recall in general, we've had to always make new fields pointers. Is this different for bool types? My understanding is defaulting is done per external version, not internal version. 

If it was not a pointer, would there be problems if you actually wanted the default to be true, but it ended up being defaulted to false by some client that doesn't know about the field?

pkg/apis/storage/validation/validation.go
```diff
@@ -473,6 +475,44 @@ func validateFSGroupPolicy(fsGroupPolicy *storage.FSGroupPolicy, fldPath *field.
 	return allErrs
 }
 
+// validateServiceAccountTokens tests if the Audience in each ServiceAccountToken are different and not empty.
+// Besides, at most one ServiceAccountToken can ignore Audience.
+func validateServiceAccountTokens(serviceAccountTokens []storage.ServiceAccountToken, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if serviceAccountTokens != nil && utilfeature.DefaultFeatureGate.Enabled(features.CSIServiceAccountToken) {
```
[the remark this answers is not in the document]
msau42: Do you see problematic scenarios from having this check? In scenarios like rollback or disablement, then this validation will not run.

pkg/apis/storage/types.go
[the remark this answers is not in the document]
mikedanese: The difference i'm talking about is internal vs external api versions.

We use *bool so we can detect false vs unset in the external type. If the field is always defaulted to either true or false before we convert to the internal, then there's no need for it to be a pointer in the internal type. If it is a pointer in the internal type, you end up with either:

* a bunch of unnecessary nil checking
* dereferences with out nil checking where it's difficult to ever know if the dereference is safe
* some combo

My inclination in the past has been to try to eliminate pointer types during conversion where possible (i.e. where defaulting has occurred). I'm not up to date on the conventions though. @liggitt?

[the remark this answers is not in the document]
liggitt: fields we want omitted in serialization if the feature is disabled have to be pointers+omitempty in the versioned structs

making the internal version a non-pointer means versioned->internal conversion turns a nil into a non-nil, and then internal->versioned produces a non-nil pointer value which will get serialized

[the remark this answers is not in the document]
mikedanese: Ok, seems like this is cleanup we can do when this feature graduates to beta.

pkg/apis/storage/types.go
```diff
@@ -374,6 +416,21 @@ const (
 // More modes may be added in the future.
 type VolumeLifecycleMode string
 
+// ServiceAccountToken contains parameters of a service account token.
+type ServiceAccountToken struct {
```
apelisse, line 420: Maybe I'm misreading, but basically `ServiceAccountToken` is not a service account token?

pkg/apis/storage/types.go
```diff
@@ -337,6 +337,48 @@ type CSIDriverSpec struct {
...
 	StorageCapacity *bool
+
+	// ServiceAccountTokens indicates the CSI driver needs pods' service account
+	// tokens it is mounting volume for to do necessary authentication. Kubelet
+	// will pass the tokens in VolumeContext in the CSI NodePublishVolume calls.
+	// The CSI driver should parse and validate the following VolumeConext:
+	// "csi.storage.k8s.io/serviceAccount.tokens": {
+	//   '<audience>': {
+	//     'token': <token>,
+	//     'expiry':<token>,
```
apelisse, line 348: this comment is wrong

apelisse:
```suggestion
	//     'expiry':<expiry>,
```

pkg/apis/storage/types.go
```diff
@@ -337,6 +337,47 @@ type CSIDriverSpec struct {
...
+	// }
+	//
+	// Note: Audience in each ServiceAccountToken should be different and at
+	// most one token can ignore Audience.
+	//
+	// This is an alpha feature and only available when the
+	// CSIServiceAccountToken feature is enabled.
+	//
+	// +optional
+	// +listType=set
```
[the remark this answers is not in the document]
apelisse: Yeah, that's questionnable. I believe this should be mechanically forbidden.
We don't support them yet in SSA, I don't think we've seen any case where this was needed.
So yes, the question is:
> what is listType=set accomplishing here?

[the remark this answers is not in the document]
zshihang: oh, it is not supported in SSA? reading from https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/0006-apply.md#api-topology that's the way to specify object topology.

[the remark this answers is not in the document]
apelisse: non-primitive types aren't supported yet. We could probably make this work if needed but I'm curious if this is actually doing what you want to do. What are you trying to achieve?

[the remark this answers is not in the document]
apelisse: Actually, doc is correct, it reads:
> Sets: the list is a set (it has to be of a scalar type). [snipped...]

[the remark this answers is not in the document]
zshihang: this list of ServiceAccountToken should be unique. not sure whether we can validate from API side or this marker is just informational.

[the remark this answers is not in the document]
apelisse: If I read properly, `ServiceAccountToken` is made of an optional `Audience` and 
 another optional `ExpirationSeconds`. What is supposed to be unique? Can there be two entries with the same Audience? Or are you trying to have unique Audience/ExpirationSeconds combination?

I still find it hard to reason about this because i find the name `ServiceAccountToken` to be misleading. It's very likely that I'm just misunderstanding, but the naming doesn't help me here :-)

>	// Note: Audience in each ServiceAccountToken should be different and at
>	// most one token can ignore Audience.

Based on that, I understand that you want `ServiceAccountTokens` to be an associative-list (a map) and the key should be `Audience`. We don't allow optional non-defaulted fields to be keys, but you could make audience required and use empty-string for "ignore", would that work?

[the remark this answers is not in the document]
apelisse: Also note that while this is not strictly informational (it's used for merging on server-side apply), at this point you still have to write manual validation for that unfortunately.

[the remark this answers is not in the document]
zshihang: so if using `+patchMergeKey=name`, the audience has to be a required `string` instead of an optional `*string`? wrt naming, probably `TokenRequest` more accurate.

pkg/apis/storage/types.go
zshihang: yea, it contains parameters in TokenRequests. maybe i can rename it to `TokenRequest`. what do you think @liggitt

pkg/apis/storage/types.go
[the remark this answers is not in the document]
apelisse: Slightly unclear, you don't have a `name` field here, so I assume you mean `+patchMergeKey=audience`. In that case, yes audience is required. I'm working on a change that would allow you to use a defaulted empty string for audience while still work as the patchMergeKey.

pkg/apis/storage/types.go
zshihang: done

pkg/apis/storage/validation/validation.go
zshihang: right. empty string is valid actually. done. audience is now required as it is gonna be the key for strategic merge

pkg/apis/storage/types.go
```diff
@@ -337,6 +337,48 @@ type CSIDriverSpec struct {
 	//
```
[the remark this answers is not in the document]
zshihang: done

pkg/apis/storage/validation/validation.go
```diff
@@ -473,6 +475,44 @@ func validateFSGroupPolicy(fsGroupPolicy *storage.FSGroupPolicy, fldPath *field.
...
+		}
+	}
+
+	return allErrs
+}
+
+// validateRequiresRepublish tests if requiresRepublish is set for CSIDriver.
+func validateRequiresRepublish(requiresRepublish *bool, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if requiresRepublish == nil && utilfeature.DefaultFeatureGate.Enabled(features.CSIServiceAccountToken) {
```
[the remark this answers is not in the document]
zshihang: done. removed this function.

pkg/apis/storage/validation/validation.go
```diff
@@ -473,6 +475,44 @@ func validateFSGroupPolicy(fsGroupPolicy *storage.FSGroupPolicy, fldPath *field.
...
+			if serviceAccountToken.Audience != nil {
+				audience = *serviceAccountToken.Audience
+				if audience == "" {
+					allErrs = append(allErrs, field.Invalid(fldPath, audience, "empty string is not a valid audience"))
+					continue
+				}
+			}
+
+			if _, ok := audiences[audience]; ok {
+				allErrs = append(allErrs, field.Duplicate(fldPath, audience))
```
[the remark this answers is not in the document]
zshihang: done

pkg/apis/storage/validation/validation.go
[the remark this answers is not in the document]
zshihang: this is obsolete because empty string is valid now.

pkg/apis/storage/validation/validation.go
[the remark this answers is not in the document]
zshihang: i removed the branch as it is not necessary

pkg/apis/storage/types.go
```diff
@@ -374,6 +416,21 @@ const (
...
+// ServiceAccountToken contains parameters of a service account token.
+type ServiceAccountToken struct {
+	// Audience is the intendend audience of the token in "TokenRequestSpec".
+	// It will default to the audiences of kube apiserver if not specified.
+	//
+	// +optional
+	Audience *string
+
+	// ExpirationSeconds is the duration of validity of the token in "TokenRequestSpec".
+	// It has the same default value of ExpirationSeconds in TokeRequestSpec.
```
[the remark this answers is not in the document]
zshihang: done

pkg/apis/storage/types.go
```diff
@@ -337,6 +337,48 @@ type CSIDriverSpec struct {
...
+
+	// ServiceAccountTokens indicates the CSI driver needs pods' service account
+	// tokens it is mounting volume for to do necessary authentication. Kubelet
+	// will pass the tokens in VolumeContext in the CSI NodePublishVolume calls.
+	// The CSI driver should parse and validate the following VolumeConext:
+	// "csi.storage.k8s.io/serviceAccount.tokens": {
+	//   '<audience>': {
+	//     'token': <token>,
+	//     'expiry':<token>,
+	//   },
```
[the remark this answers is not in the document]
zshihang: done

pkg/apis/storage/types.go
```diff
@@ -337,6 +337,48 @@ type CSIDriverSpec struct {
 	//
 	// +optional
 	StorageCapacity *bool
+
+	// ServiceAccountTokens indicates the CSI driver needs pods' service account
+	// tokens it is mounting volume for to do necessary authentication. Kubelet
+	// will pass the tokens in VolumeContext in the CSI NodePublishVolume calls.
+	// The CSI driver should parse and validate the following VolumeConext:
```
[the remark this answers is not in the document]
zshihang: done

pkg/apis/storage/fuzzer/fuzzer.go
```diff
@@ -87,6 +87,13 @@ var Funcs = func(codecs runtimeserializer.CodecFactory) []interface{} {
 				obj.Spec.FSGroupPolicy = new(storage.FSGroupPolicy)
 				*obj.Spec.FSGroupPolicy = storage.ReadWriteOnceWithFSTypeFSGroupPolicy
 			}
+			if len(obj.Spec.ServiceAccountTokens) == 0 {
+				obj.Spec.ServiceAccountTokens = []storage.ServiceAccountToken{{}}
```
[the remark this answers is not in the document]
zshihang: ugh, i forgot it maybe due to test failure. removed it.

pkg/volume/csi/csi_plugin.go
```diff
@@ -334,8 +340,28 @@ func (p *csiPlugin) CanSupport(spec *volume.Spec) bool {
...
 
-func (p *csiPlugin) RequiresRemount() bool {
-	return false
+func (p *csiPlugin) RequiresRemount(spec *volume.Spec) bool {
+	if !utilfeature.DefaultFeatureGate.Enabled(features.CSIServiceAccountToken) {
+		return false
+	}
+	kletHost, ok := p.host.(volume.KubeletVolumeHost)
+	if ok {
+		if err := kletHost.WaitForCacheSync(); err != nil {
```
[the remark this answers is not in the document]
zshihang: good point. it is actually will block forever. @msau42 this is a interface method that used in many places. i think we need to give a timeout in most cases.

pkg/volume/csi/csi_plugin.go
```diff
@@ -334,8 +340,28 @@ func (p *csiPlugin) CanSupport(spec *volume.Spec) bool {
 	return spec.PersistentVolume != nil && spec.PersistentVolume.Spec.CSI != nil
 }
 
-func (p *csiPlugin) RequiresRemount() bool {
-	return false
+func (p *csiPlugin) RequiresRemount(spec *volume.Spec) bool {
```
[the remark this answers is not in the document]
zshihang: done

staging/src/k8s.io/api/storage/v1beta1/types.go
```diff
@@ -364,6 +364,48 @@ type CSIDriverSpec struct {
...
+	// Note: Audience in each ServiceAccountToken should be different and at
+	// most one token can ignore Audience. To receive a new token after expiry,
+	// RequiresRepublish can be used to trigger NodePublishVolume periodically.
+	//
+	// This is an alpha feature and only available when the
+	// CSIServiceAccountToken feature is enabled.
+	//
+	// +optional
+	// +listType=set
+	ServiceAccountTokens []ServiceAccountToken `json:"serviceAccountTokens,omitempty" protobuf:"bytes,6,opt,name=serviceAccountTokens"`
```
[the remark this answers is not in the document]
zshihang: done

pkg/apis/storage/types.go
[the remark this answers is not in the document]
zshihang: yea, i meant `audience` key. the change would be needed by this PR. when do you the change would be ready?

pkg/apis/storage/types.go
```diff
@@ -374,6 +416,21 @@ const (
 // More modes may be added in the future.
 type VolumeLifecycleMode string
 
+// ServiceAccountToken contains parameters of a service account token.
+type ServiceAccountToken struct {
+	// Audience is the intendend audience of the token in "TokenRequestSpec".
```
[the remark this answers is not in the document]
zshihang: done

pkg/volume/csi/csi_plugin.go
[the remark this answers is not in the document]
msau42: I think instead of calling `WaitForCacheSync` here, which gets called very frequently, it would be better to do this during volume manager initialization in kubelet. It looks like we already have plumbing to wait for all sources to be ready: https://github.com/kubernetes/kubernetes/blob/2ee1003430ea42060b48de9b28ecab2ce846c366/pkg/kubelet/volumemanager/volume_manager.go#L260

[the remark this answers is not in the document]
zshihang: done. cache will be syncing in the initialization. in a future PR, might clean up the places calling `WaitForCacheSync` as it is not necessary.

pkg/volume/plugins.go
```diff
@@ -1027,6 +1027,7 @@ func (pm *VolumePluginMgr) Run(stopCh <-chan struct{}) {
 		// start informer for CSIDriver
 		informerFactory := kletHost.GetInformerFactory()
 		informerFactory.Start(stopCh)
+		informerFactory.WaitForCacheSync(stopCh)
```
msau42, line 1030: Hm, I believe that if kubelet runs in standalone mode, then having `WaitForCacheSync` here is problematic because there's no apiserver.

zshihang: this is guarded by the `if` here
https://github.com/kubernetes/kubernetes/blob/941e9242a50fa8f273a0582eca283176c64e3085/pkg/kubelet/volumemanager/volume_manager.go#L271-L274

msau42: Ah yes, that should protect it, but I wonder if we need to change the ordering of it because right now the reconciler (which is the one that will call into MountVolume) starts before this.

zshihang: done.

zshihang: looks like we should not block on the `WaitForCacheSync` before reconciler because it will timeout the bootstrap:

```
Nov 04 21:48:15 kind-control-plane kubelet[266]: E1104 21:48:15.835859     266 kubelet.go:1650] Unable to attach or mount volumes for pod "kube-apiserver-kind-control-plane_kube-system(b9476abec6e5f541a230b24626dd0325)": unmounted volumes=[k8s-certs usr-local-share-ca-certificates usr-share-ca-certificates ca-certs etc-ca-certificates], unattached volumes=[k8s-certs usr-local-share-ca-certificates usr-share-ca-certificates ca-certs etc-ca-certificates]: timed out waiting for the condition; skipping pod
```

pkg/apis/storage/types.go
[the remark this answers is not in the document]
apelisse: Sorry I missed that comment, as answered privately, we're planning on having that in 1.20.

pkg/volume/plugins.go
msau42: Thanks for experimenting around. I opened up https://github.com/kubernetes/kubernetes/issues/96316.

For this feature, if there isn't any negative effect from the informer being not in sync, then I'm ok with not waiting on it, and letting a retry resolve itself.

staging/src/k8s.io/api/storage/v1beta1/types.go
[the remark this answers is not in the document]
zshihang: right now empty string as key is not allowed. use listType=atomic instead for now until the necessary changes for empty string as key are landed in 1.20.

pkg/apis/storage/types.go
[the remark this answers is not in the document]
zshihang: thank you. for now i'll use `listType=atomic` until the changes land in 1.20. PTAL.

pkg/apis/storage/types.go
zshihang: renamed

test/e2e/storage/csi_mock_volume.go
zshihang: PR merged

zshihang:
/retest

those azure tests have been failing for a while.

zshihang:
ping for review. note the code freeze for 1.20 is Nov 12th.

staging/src/k8s.io/api/storage/v1/types.go
```diff
@@ -344,6 +344,43 @@ type CSIDriverSpec struct {
...
+	//
+	// Note: Audience in each TokenRequest should be different and at
+	// most one token is empty string. To receive a new token after expiry,
+	// RequiresRepublish can be used to trigger NodePublishVolume periodically.
+	//
+	// This is an alpha feature and only available when the
+	// CSIServiceAccountToken feature is enabled.
+	//
+	// +optional
+	// +listType=atomic
```
apelisse, line 368: Looks good to me. I would keep the code that validates that you only have one audience at most, and I would default unspecified audience to empty string so that you can smoothly use audience as a key in a future release without breaking compatibility.

apelisse: Looks like that's what we have!

pkg/volume/csi/csi_mounter.go
```diff
@@ -319,6 +322,56 @@ func (c *csiMountMgr) podAttributes() (map[string]string, error) {
...
+					UID:        c.pod.UID,
+				},
+			},
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		output, _ := json.Marshal(tr.Status)
+		outputs[audience] = string(output)
```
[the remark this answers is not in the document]
zshihang: done.

yes. in the csi_mounter_test.

pkg/volume/csi/csi_mounter.go
```diff
@@ -319,6 +322,56 @@ func (c *csiMountMgr) podAttributes() (map[string]string, error) {
...
+		}
+		return nil, err
+	}
+
+	if len(csiDriver.Spec.TokenRequests) == 0 {
+		return nil, nil
+	}
+
+	outputs := map[string]string{}
+	for _, tokenRequest := range csiDriver.Spec.TokenRequests {
```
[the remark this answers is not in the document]
zshihang: 100ms between between republish calls. the tokens are cached in kubelet. it is using the same handler as projected tokens.

pkg/apis/storage/types.go
```diff
@@ -337,6 +337,43 @@ type CSIDriverSpec struct {
...
+	// Note: Audience in each TokenRequest should be different and at
+	// most one token is empty string. To receive a new token after expiry,
+	// RequiresRepublish can be used to trigger NodePublishVolume periodically.
+	//
+	// This is an alpha feature and only available when the
+	// CSIServiceAccountToken feature is enabled.
+	//
+	// +optional
+	// +listType=atomic
+	TokenRequests []TokenRequest
```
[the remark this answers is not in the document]
zshihang: in case of secrets store csi driver, they require at least 3 audiences for providers: vault, azure, gcp... why we would limit the maximum number? we'd defer this decision to cluster administrators.

pkg/volume/csi/csi_mounter.go
```diff
@@ -319,6 +322,56 @@ func (c *csiMountMgr) podAttributes() (map[string]string, error) {
...
+		var audiences []string
+		audience := tokenRequest.Audience
+		audiences = []string{audience}
```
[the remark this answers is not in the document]
zshihang: good catch. done.

no, there is no such an e2e, but i added a test case in csi_mounter_test to make sure when len(audiences)=0, we received api audiences in the volume attributes. together with the e2e added in this change, we achieved the same effect as "an e2e test that requests a token with audience "" and verifies it is usable against the kube apiserver"

pkg/registry/storage/csidriver/strategy.go
```diff
@@ -55,6 +55,11 @@ func (csiDriverStrategy) PrepareForCreate(ctx context.Context, obj runtime.Objec
 	if !utilfeature.DefaultFeatureGate.Enabled(features.CSIVolumeFSGroupPolicy) {
 		csiDriver.Spec.FSGroupPolicy = nil
 	}
+	if !utilfeature.DefaultFeatureGate.Enabled(features.CSIServiceAccountToken) {
+		csiDriver := obj.(*storage.CSIDriver)
```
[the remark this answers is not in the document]
zshihang: done

pkg/printers/internalversion/printers.go
```diff
@@ -1395,6 +1401,21 @@ func printCSIDriver(obj *storage.CSIDriver, options printers.GenerateOptions) ([
 		}
 		row.Cells = append(row.Cells, storageCapacity)
 	}
+	if utilfeature.DefaultFeatureGate.Enabled(features.CSIServiceAccountToken) {
+		tokenRequests := "<unset>"
+		if obj.Spec.TokenRequests != nil {
+			tokens := []string{}
+			for _, t := range obj.Spec.TokenRequests {
+				tokens = append(tokens, fmt.Sprintf("{audience:%s,expirationSeconds:%d}", t.Audience, *t.ExpirationSeconds))
```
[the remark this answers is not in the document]
zshihang: done

pkg/kubelet/volumemanager/volume_manager.go
```diff
@@ -260,6 +260,11 @@ type volumeManager struct {
 func (vm *volumeManager) Run(sourcesReady config.SourcesReady, stopCh <-chan struct{}) {
 	defer runtime.HandleCrash()
 
+	if vm.kubeClient != nil {
```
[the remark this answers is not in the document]
zshihang: not necessary here but the sooner to populate the csi driver lister cache before desire workload state populator starts the better. racing is still there tho. Michelle opened an issue to figure out the place to start `WaitForCacheSync`.

pkg/apis/storage/validation/validation.go
```diff
@@ -473,6 +474,23 @@ func validateFSGroupPolicy(fsGroupPolicy *storage.FSGroupPolicy, fldPath *field.
...
+	audiences := make(map[string]bool)
+	for i, serviceAccountToken := range serviceAccountTokens {
+		audience := serviceAccountToken.Audience
+		if _, ok := audiences[audience]; ok {
+			allErrs = append(allErrs, field.Duplicate(fldPath.Index(i).Child("audience"), audience))
+			continue
+		}
+		audiences[audience] = true
+	}
+
```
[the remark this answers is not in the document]
zshihang: done for the `ExpirationSeconds `.

the rationale for not setting the max number is that we don't know how many audiences a csi driver would like to have. e.g. secrets store csi driver would require three audiences now and may be more in the future.

pkg/volume/csi/csi_mounter.go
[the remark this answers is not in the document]
liggitt: does that mean we're calling the CSI driver every 100ms?

pkg/volume/csi/csi_mounter_test.go
```diff
@@ -927,3 +936,122 @@ func TestIsCorruptedDir(t *testing.T) {
...
+				},
+				Spec: storagev1.CSIDriverSpec{
+					TokenRequests: []storagev1.TokenRequest{
+						{
+							Audience: "",
+						},
+					},
+				},
+			},
+			wantVolumeContext: map[string]string{"csi.storage.k8s.io/serviceAccount.tokens": `{"":{"token":"test-ns:test-service-account:3600:[api]","expirationTimestamp":null}}`},
```
liggitt, line 980: `"expirationTimestamp":null` is unexpected... shouldn't this be a string value?

pkg/volume/csi/csi_mounter_test.go
```diff
@@ -927,3 +936,122 @@ func TestIsCorruptedDir(t *testing.T) {
...
+				},
+				Spec: storagev1.CSIDriverSpec{
+					TokenRequests: []storagev1.TokenRequest{
+						{
+							Audience: gcp,
+						},
+					},
+				},
+			},
+			wantVolumeContext: map[string]string{"csi.storage.k8s.io/serviceAccount.tokens": `{"gcp":{"token":"test-ns:test-service-account:3600:[gcp]","expirationTimestamp":null}}`},
```
liggitt, line 996: `"expirationTimestamp":null` is unexpected... shouldn't this be a string value?

pkg/volume/csi/csi_mounter.go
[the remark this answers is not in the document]
zshihang: no because every reconciler loop takes ~3 seconds and the reconciler loop calls republish. https://github.com/kubernetes/kubernetes/blob/8be0a29ce027ecc75004c1a4945a87d128c4f45f/pkg/kubelet/volumemanager/reconciler/reconciler.go#L80-L81

pkg/volume/csi/csi_mounter_test.go
zshihang: metav1.Time has the marshallJson to marshall zero value as null and in this test i am doing `tr.Status.ExpirationTimestamp = metav1.Time{}`

i changed it to a non zero value.

pkg/volume/csi/csi_mounter_test.go
zshihang: same above

zshihang:
> >For this feature, if there isn't any negative effect from the informer being not in sync, then I'm ok with not waiting on it, and letting a retry resolve itself.

> is this still outstanding? should WaitForCacheSync be removed?

no, this will be called in a go routine. it is better to have it happen sooner. @msau42

pkg/volume/plugins.go
msau42: `volumeManager.Run` is called in a [go routine](https://github.com/kubernetes/kubernetes/blob/941e9242a50fa8f273a0582eca283176c64e3085/pkg/kubelet/kubelet.go#L1387) so we won't block kubelet boostrapping.

pkg/kubelet/volumemanager/volume_manager.go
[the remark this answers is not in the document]
liggitt: > racing is still there though

what is the impact of losing the race? functional failure or longer startup time?

liggitt:
/approve

API and validation changes look good. I did not review the volume plugin changes.

Needs rebase, and storage reviewers have lgtm on remaining items.

pkg/kubelet/volumemanager/volume_manager.go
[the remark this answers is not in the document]
msau42: @zshihang and I discussed. The impact is NodePublish (mount) may be called without the token. The CSI driver is expected to fail if it expects one. The volume reconciler will retry NodePublish until successful.

Issue #92671, opened 2020-07-01
After the aggregation api is deleted, the deleted api still exists in the kube-apiserver metrics


DesistDaydream:
**What happened**:
After the aggregation api is deleted, the deleted api still exists in the kube-apiserver metrics
```
[root@master-3 ~]# kubectl get apiservices.apiregistration.k8s.io 
NAME                                   SERVICE                         AVAILABLE   AGE
v1.                                    Local                           True        37d
v1.admissionregistration.k8s.io        Local                           True        37d
v1.apiextensions.k8s.io                Local                           True        37d
v1.apps                                Local                           True        37d
v1.authentication.k8s.io               Local                           True        37d
v1.authorization.k8s.io                Local                           True        37d
v1.autoscaling                         Local                           True        37d
v1.batch                               Local                           True        37d
v1.coordination.k8s.io                 Local                           True        37d
v1.monitoring.coreos.com               Local                           True        23h
v1.networking.k8s.io                   Local                           True        37d
v1.rbac.authorization.k8s.io           Local                           True        37d
v1.scheduling.k8s.io                   Local                           True        37d
v1.storage.k8s.io                      Local                           True        37d
v1alpha1.install.istio.io              Local                           True        23h
v1alpha1.rbac.istio.io                 Local                           True        23h
v1alpha2.config.istio.io               Local                           True        12d
v1alpha3.networking.istio.io           Local                           True        23h
v1beta1.admissionregistration.k8s.io   Local                           True        37d
v1beta1.apiextensions.k8s.io           Local                           True        37d
v1beta1.authentication.k8s.io          Local                           True        37d
v1beta1.authorization.k8s.io           Local                           True        37d
v1beta1.batch                          Local                           True        37d
v1beta1.certificates.k8s.io            Local                           True        37d
v1beta1.coordination.k8s.io            Local                           True        37d
v1beta1.custom.metrics.k8s.io          monitoring/prometheus-adapter   True        18h
v1beta1.discovery.k8s.io               Local                           True        37d
v1beta1.events.k8s.io                  Local                           True        37d
v1beta1.extensions                     Local                           True        37d
v1beta1.metrics.k8s.io                 monitoring/prometheus-adapter   True        34h
v1beta1.networking.istio.io            Local                           True        23h
v1beta1.networking.k8s.io              Local                           True        37d
v1beta1.node.k8s.io                    Local                           True        37d
v1beta1.policy                         Local                           True        37d
[...]

PR #91310, opened 2020-05-20
Reset flowcontrol metrics on a DELETE /metrics

tkashem:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:
Priority and Fairness metrics should reset when a a reset request is received (DELETE on /metrics endpoint)

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
None

```release-note
```

PR #93134, opened 2020-07-16
Add reset handler to the instrumentation metric library and expose Reset on the metric registries

logicalhan:
**What type of PR is this?**

/kind cleanup

We've had a few issues involving metric registry reset recently (https://github.com/kubernetes/kubernetes/pull/91310 & https://github.com/kubernetes/kubernetes/issues/92671). This PR will fix re-occurrances of the former but not the latter. 

```release-note
NONE
```

/sig instrumentation
/assign @brancz @RainbowMango

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/server/routes/metrics.go
+++ b/staging/src/k8s.io/apiserver/pkg/server/routes/metrics.go
@@ -17,9 +17,6 @@ limitations under the License.
 package routes
 
 import (
-	"io"
-	"net/http"
-
 	apimetrics "k8s.io/apiserver/pkg/endpoints/metrics"
 	"k8s.io/apiserver/pkg/server/mux"
 	etcd3metrics "k8s.io/apiserver/pkg/storage/etcd3/metrics"
@@ -43,18 +40,7 @@ type MetricsWithReset struct{}
 // Install adds the MetricsWithReset handler
 func (m MetricsWithReset) Install(c *mux.PathRecorderMux) {
 	register()
-	defaultMetricsHandler := legacyregistry.Handler().ServeHTTP
-	c.HandleFunc("/metrics", func(w http.ResponseWriter, req *http.Request) {
-		if req.Method == "DELETE" {
-			apimetrics.Reset()
-			etcd3metrics.Reset()
-			flowcontrolmetrics.Reset()
-
-			io.WriteString(w, "metrics reset\n")
-			return
-		}
-		defaultMetricsHandler(w, req)
-	})
+	c.Handle("/metrics", legacyregistry.HandlerWithReset())
 }
 
 // register apiserver and etcd metrics
--- a/staging/src/k8s.io/component-base/metrics/http.go
+++ b/staging/src/k8s.io/component-base/metrics/http.go
@@ -17,6 +17,7 @@ limitations under the License.
 package metrics
 
 import (
+	"io"
 	"net/http"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -60,4 +61,17 @@ func (ho *HandlerOpts) toPromhttpHandlerOpts() promhttp.HandlerOpts {
 // kind of instrumentation as it is used by the Handler function.
 func HandlerFor(reg Gatherer, opts HandlerOpts) http.Handler {
 	return promhttp.HandlerFor(reg, opts.toPromhttpHandlerOpts())
+}
+
+// HandlerWithReset return an http.Handler with Reset
+func HandlerWithReset(reg KubeRegistry, opts HandlerOpts) http.Handler {
+	defaultHandler := promhttp.HandlerFor(reg, opts.toPromhttpHandlerOpts())
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == http.MethodDelete {
+			reg.Reset()
+			io.WriteString(w, "metrics reset\n")
+			return
+		}
+		defaultHandler.ServeHTTP(w, r)
+	})
 }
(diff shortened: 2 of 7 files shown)


staging/src/k8s.io/apiserver/pkg/server/routes/metrics.go
```diff
@@ -43,18 +40,8 @@ type MetricsWithReset struct{}
...
-			etcd3metrics.Reset()
-			flowcontrolmetrics.Reset()
-
-			io.WriteString(w, "metrics reset\n")
-			return
-		}
-		defaultMetricsHandler(w, req)
-	})
+	defaultMetricsHandler := legacyregistry.HandlerWithReset().ServeHTTP
+	c.HandleFunc("/metrics", defaultMetricsHandler)
```
lavalamp, line 44: Shorten even more: `c.Handle("/metrics", legacyregistry.HandlerWithReset())`

staging/src/k8s.io/component-base/metrics/http.go
```diff
@@ -61,3 +62,16 @@ func (ho *HandlerOpts) toPromhttpHandlerOpts() promhttp.HandlerOpts {
 func HandlerFor(reg Gatherer, opts HandlerOpts) http.Handler {
 	return promhttp.HandlerFor(reg, opts.toPromhttpHandlerOpts())
 }
+
+// HandlerWithReset return an http.Handler with Reset
+func HandlerWithReset(reg KubeRegistry, opts HandlerOpts) http.Handler {
```
lavalamp, line 67: It doesn't feel right for this to be just loose in the package. Could it be a method on `reg`?

staging/src/k8s.io/component-base/metrics/legacyregistry/registry.go
```diff
@@ -83,3 +91,8 @@ func CustomMustRegister(cs ...metrics.StableCollector) {
 		prometheus.MustRegister(c)
 	}
 }
+
+// Reset calls reset on the global registry
```
lavalamp, line 95: Sometimes I do `var Reset = defaultRegistry.Reset()` instead of a function, to make this even more obvious (up to you).

staging/src/k8s.io/component-base/metrics/registry.go
```diff
@@ -129,11 +144,11 @@ type kubeRegistry struct {
 // uniqueness criteria described in the documentation of metric.Desc.
 func (kr *kubeRegistry) Register(c Registerable) error {
 	if c.Create(&kr.version) {
+		defer kr.addResettable(c)
```
lavalamp, line 147: Do you really want to add it to the resettable list if registering it gives an error?

staging/src/k8s.io/component-base/metrics/registry.go
```diff
@@ -266,6 +294,12 @@ func (kr *kubeRegistry) enableHiddenStableCollectors() {
 	kr.CustomMustRegister(cs...)
 }
 
+func (kr *kubeRegistry) Reset() {
```
lavalamp, line 297: Why not take the lock here? At least a read lock?

lavalamp (commented):
I like the tests and that everything will be consistent. It still feels like a pretty arbitrary & undiscoverable way to reset the metrics, but I don't think this PR needs to fix that.

staging/src/k8s.io/apiserver/pkg/server/routes/metrics.go
logicalhan: ack

staging/src/k8s.io/component-base/metrics/http.go
logicalhan: Prometheus decouples these on purpose (I think), because it is possible to reuse a registry with different metric endpoints (and have custom behavior). I would be okay with hanging it off the registry (since I don't we rely off of that behavior in Kubernetes) but I'd feel more comfortable if @brancz ok'd that particular suggestion first.

staging/src/k8s.io/component-base/metrics/registry.go
logicalhan: ack

staging/src/k8s.io/component-base/metrics/http.go
lavalamp: I must not understand what a registry is. Wouldn't all handlers that get passed the same registry show the same data?

staging/src/k8s.io/component-base/metrics/registry.go
logicalhan: It's in a conditional, so it will only happen if Create returned true.

lavalamp: Right, but the function could still return an error overall? (Honestly this function is already problematic, since it can return an error but leave things in a half completed state.)

staging/src/k8s.io/component-base/metrics/http.go
logicalhan: Technically, it doesn't actually take a registry, it takes a prometheus.Gatherer (which is an interface the Registry satisfies), so you could do things with the registry that would produce a different Gatherer.

lavalamp: I see. Except now it takes a KubeRegistry because Gatherer doesn't have a Reset() function? `func HandlerWithReset(reg Gatherer, r Resetter, opts HandlerOpts) http.Handler` as a function signature would make that more obvious.

(no need to change, we've already talked about this longer than it merits)

staging/src/k8s.io/component-base/metrics/legacyregistry/registry.go
logicalhan: ah nice, yeah i made the refactoring for that and some other code sites where that comment is applicable.

staging/src/k8s.io/component-base/metrics/http.go
logicalhan: Yes, I've overloaded the Registry a lot already... :(

staging/src/k8s.io/component-base/metrics/registry.go
logicalhan: It shouldn't error in that spot, but we had to keep the signature consistent. But you are correct about this function being kinda broken, we should kick back an error if the create fails.

PR #91968, opened 2020-06-09
Remove contemplation of invariant violations from delta_fifo.go

MikeSpreitzer:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:
Some comments and code incorrectly contemplated violating the
invariant that a keys is in `f.items` if and only if it is in
`f.queue`.  Changed to not contemplate invariant violations,
and panic if one is stumbled upon.

Also fixed up some comment wording, some of which was discussed late in #91435 

**Which issue(s) this PR fixes**:

Fixes #91709 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #93140, opened 2020-07-16
 ignore check  if the queue item exists

ide-rea:
**What type of PR is this?**
 /kind cleanup
the queue  and  items are strictly 1:1 and the pop operation is protected by lock, so we don't need to check if item exists.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/client-go/tools/cache/fifo.go
+++ b/staging/src/k8s.io/client-go/tools/cache/fifo.go
@@ -294,11 +294,8 @@ func (f *FIFO) Pop(process PopProcessFunc) (interface{}, error) {
 		if f.initialPopulationCount > 0 {
 			f.initialPopulationCount--
 		}
-		item, ok := f.items[id]
-		if !ok {
-			// Item may have been deleted subsequently.
-			continue
-		}
+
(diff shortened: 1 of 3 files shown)


lavalamp:
https://github.com/kubernetes/kubernetes/pull/91968 is already in my queue about this...

staging/src/k8s.io/client-go/tools/cache/fifo.go
```diff
@@ -294,11 +294,8 @@ func (f *FIFO) Pop(process PopProcessFunc) (interface{}, error) {
 		if f.initialPopulationCount > 0 {
 			f.initialPopulationCount--
 		}
-		item, ok := f.items[id]
-		if !ok {
-			// Item may have been deleted subsequently.
-			continue
-		}
+
+		item := f.items[id]
```
mikedanese, line 298: As mentioned, the comment is inccorect. See:

https://github.com/kubernetes/kubernetes/blob/b1b93e30131986319f1d46a62771feb9c0f097ae/staging/src/k8s.io/client-go/tools/cache/fifo.go#L210-L220

PR #93142, opened 2020-07-16
fix typo release to update

lixiaobing1:
What type of PR is this?
/kind cleanup

following error info release should be update:
func (le *LeaderElector) release() bool {
	if err := le.config.Lock.Update(context.TODO(), leaderElectionRecord); err != nil {
		klog.Errorf("Failed to release lock: %v", err)
		return false
	}

```release-note
   fix typo release to update
```

### DIFF
--- a/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go
+++ b/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go
@@ -294,7 +294,7 @@ func (le *LeaderElector) release() bool {
 		LeaderTransitions: le.observedRecord.LeaderTransitions,
 	}
 	if err := le.config.Lock.Update(context.TODO(), leaderElectionRecord); err != nil {
-		klog.Errorf("Failed to release lock: %v", err)
+		klog.Errorf("Failed to update lock: %v", err)
 		return false
 	}
 	le.observedRecord = leaderElectionRecord


staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go
```diff
@@ -294,7 +294,7 @@ func (le *LeaderElector) release() bool {
 		LeaderTransitions: le.observedRecord.LeaderTransitions,
 	}
 	if err := le.config.Lock.Update(context.TODO(), leaderElectionRecord); err != nil {
-		klog.Errorf("Failed to release lock: %v", err)
```
caesarxuchao, line 297: The purpose of the function is to release the lock, so I don't think this is a typo.

lixiaobing1: In my opinion,  the func release() is to release the leader lease,  and the error info is after executing le.config.Lock.Update, I think the error info should be identical to the background.

// release attempts to release the leader lease if we have acquired it.
func (le *LeaderElector) release() bool {

liggitt:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/93142/pull-kubernetes-node-e2e/1283732232065781760 flake is #93210

staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go
caesarxuchao: How about "failed to release lease", I think that's more specific.

Issue #92914, opened 2020-07-08
[RESERVED]


tallclair:
This issue is reserved for a future vulnerability announcement.

/close

tdwyer:
PoC

https://github.com/tdwyer/CVE-2020-8559

tabbysable:
A different approach to exploiting CVE-2020-8559: https://github.com/tabbysable/POC-2020-8559/

tallclair:
@tabbysable Any interest in turning your POC into a Kubernetes E2E test? I was intending to use the overridden redirect feature added in https://github.com/kubernetes/kubernetes/pull/92850 as the basis for the E2E.

tabbysable:
@tallclair that is indeed interesting! I'll put this on my list and try to look at it in the next couple weeks.

PushkarJ:
/label official-cve-feed

(Related to https://github.com/kubernetes/sig-security/issues/1)

PR #93144, opened 2020-07-16
feat(client-go): not follow http redirects by default

### LINKED ISSUE #93129: Allow kubectl (and client-go and other client libs) to not follow redirects
**What would you like to be added**: A field in kubeconfig, and a flag to kubectl to prevent the underlying client from following any redirects.

**Why is this needed**: kubectl generally doesn't need to follow redirects. In most cases, following redirects is potentially dangerous.

[CVE-2020-8559](https://github.com/kubernetes/kubernetes/issues/92914) would not have been an issue if kubectl didn't follow redirects. If there were an option to disallow following redirects, we would have had an immediate mitigation.

knight42:

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This PR make users be able to configure whether the rest client should follow HTTP redirects, and not to follow HTTP redirects by default. It also addresses CVE-2020-8559.

Fixes #93129 

```release-note
The rest client do not follow HTTP redirects by default.
```

```docs

```


knight42:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93144/pull-kubernetes-conformance-kind-ga-only-parallel/1283970033424273408#1:build-log.txt%3A13482
looks like the missing trailing '/' causes redirects in e2e tests

knight42:
another case: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/93144/pull-kubernetes-e2e-gce-ubuntu-containerd/1284011933082587139/#1:build-log.txt%3A52127
```
I0717 06:58:15.143] DEBUG: original url: https://35.230.103.251/logs
I0717 06:58:15.143] redirect location: logs/
```

staging/src/k8s.io/client-go/rest/config_test.go
```diff
@@ -611,7 +611,7 @@ func TestConfigSprint(t *testing.T) {
...
 		c.Transport, fakeWrapperFunc, c.RateLimiter, fakeDialFunc, fakeProxyFunc,
```
yliaog, line 615: by default, FollowRedirect is false, so the change is not strictly needed, right?

staging/src/k8s.io/client-go/rest/client_test.go
```diff
@@ -372,6 +372,63 @@ func TestCreateBackoffManager(t *testing.T) {
...
+	server2 := httptest.NewServer(&utiltesting.FakeHandler{
+		StatusCode: 301,
+		ResponseHeaders: map[string]string{
+			"Location": server1.URL,
+		},
+		T: t,
+	})
+	defer server2.Close()
+
+	c, err := RESTClientFor(&Config{
```
yliaog, line 409: what about a case for UnversionedRESTClientFor?

staging/src/k8s.io/client-go/rest/request.go
```diff
@@ -967,6 +967,10 @@ func (r *Request) DoRaw(ctx context.Context) ([]byte, error) {
 		result.body, result.err = ioutil.ReadAll(resp.Body)
 		glogBody("Response Body", result.body)
 		if resp.StatusCode < http.StatusOK || resp.StatusCode > http.StatusPartialContent {
+			if resp.StatusCode >= 300 && resp.StatusCode < 400 {
+				// DEBUG
```
yliaog, line 977: remove?

staging/src/k8s.io/client-go/rest/request.go
```diff
@@ -1041,6 +1045,11 @@ func (r *Request) transformResponse(resp *http.Response, req *http.Request) Resu
 	switch {
 	case resp.StatusCode == http.StatusSwitchingProtocols:
 		// no-op, we've been upgraded
+
+	case resp.StatusCode >= 300 && resp.StatusCode < 400:
```
yliaog, line 1055: remove?

staging/src/k8s.io/client-go/rest/request.go
```diff
@@ -476,6 +476,12 @@ func (r *Request) URL() *url.URL {
 	// Join trims trailing slashes, so preserve r.pathPrefix's trailing slash for backwards compatibility if nothing was changed
 	if len(r.resourceName) != 0 || len(r.subpath) != 0 || len(r.subresource) != 0 {
 		p = path.Join(p, r.resourceName, r.subresource, r.subpath)
+		// Since Join trims trailing slashes,
+		// if r.subresource is "proxy" and r.subpath is "/",
+		// we need to preserve the trailing slash to keep the original path.
+		if strings.HasSuffix(r.subpath, "/") {
+			p += "/"
```
yliaog, line 483: the comment at line # 476 mentioned "preserve r.pathPrefix's trailing slash", no?

staging/src/k8s.io/client-go/rest/config_test.go
knight42: I think the change is required to pass the test, as we added a new field to `rest.Config` and the test is `fmt.Sprintf`ing it, we need to update the result.

staging/src/k8s.io/client-go/rest/request.go
knight42: will remove it after all tests passed.

staging/src/k8s.io/client-go/rest/request.go
knight42: I think the meaning of line # 476 is that if `r.resourceName`, `r.subpath` and `r.subresource` are all empty, then do not update variable `p` as `path.Join` trims trailing slashes.

test/e2e/framework/pod/resource.go
```diff
@@ -105,6 +105,7 @@ func (r ProxyResponseChecker) CheckAllResponses() (done bool, err error) {
 			Resource("pods").
 			SubResource("proxy").
 			Name(string(pod.Name)).
+			Suffix("/").
```
liggitt, line 108: this is unexpected

test/e2e/network/networking.go
```diff
@@ -133,7 +133,7 @@ var _ = SIGDescribe("Networking", func() {
 			// TODO: test proxy links here
 		}
 		if !framework.ProviderIs("gke", "skeleton") {
-			tests = append(tests, struct{ path string }{path: "/logs"})
+			tests = append(tests, struct{ path string }{path: "/logs/"})
```
liggitt, line 136: this is unexpected

staging/src/k8s.io/client-go/rest/request.go
liggitt, line 483: adding a trailing slash is unexpected

test/e2e/framework/pod/resource.go
knight42: see also https://github.com/kubernetes/kubernetes/pull/93144#issuecomment-659835179

staging/src/k8s.io/client-go/rest/request.go
knight42: see also https://github.com/kubernetes/kubernetes/pull/93144#issuecomment-659835179

test/e2e/network/networking.go
knight42: see also https://github.com/kubernetes/kubernetes/pull/93144#issuecomment-659920421

staging/src/k8s.io/client-go/rest/request.go
liggitt: Needing to do fixup here and in the e2e tests is a good sign that this is more of a breaking change than was expected. blindly appending "/" to all resource/subresource/subpath names is likely to cause problems.

knight42: we would only append '/' if original subpath ends with '/', it is not quite bindly I think. and since `path.Join` trims trailing slashes, simply updating e2e test is not enough. I would like to here your advice on how we fix https://github.com/kubernetes/kubernetes/pull/93144#issuecomment-659835179 here.

liggitt: ah, I see. I think `p = utilnet.JoinPreservingTrailingSlash(p, r.resourceName, r.subresource, r.subpath)` will do what you want here

If our e2e tests are hitting failures because of dropping redirect following, it is likely real-world clients will too.

knight42: > utilnet.JoinPreservingTrailingSlash

Great! I will update the code.

> If our e2e tests are hitting failures because of dropping redirect following, it is likely real-world clients will too.

yes, I am also a bit worried about that, so I leave it to be configurable. Do you think we should follow redirect by default?

liggitt: I think it needs broader discussion. Normally a change like this starts as a proposal where the risks and benefits are discussed and reviewed by a wider audience before jumping into a PR

knight42: ok I will submit a KEP later.

knight42: @liggitt I just noticed that [CVE-2020-8559](https://github.com/kubernetes/kubernetes/issues/92914) has already been solved by #92941 , do you think it is still valuable to work on this?

liggitt: yes, I still think https://github.com/kubernetes/kubernetes/issues/93129 is a good direction to move

PR #93159, opened 2020-07-16
Wait on PD CSI driver deployment

### LINKED ISSUE #139: Validate the driver successfully came up in k8s test
Need to deal with the master DaemonSet pod pending

mattcary:
/kind bug
Fixes https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/139

```release-note
NONE
```

### DIFF
--- a/test/e2e/storage/drivers/csi.go
+++ b/test/e2e/storage/drivers/csi.go
@@ -53,6 +53,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
+	e2edeployment "k8s.io/kubernetes/test/e2e/framework/deployment"
 	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
 	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
@@ -559,7 +560,7 @@ func (g *gcePDCSIDriver) PrepareTest(f *framework.Framework) (*testsuites.PerTes
 		"test/e2e/testing-manifests/storage-csi/external-provisioner/rbac.yaml",
 		"test/e2e/testing-manifests/storage-csi/gce-pd/csi-controller-rbac.yaml",
 		"test/e2e/testing-manifests/storage-csi/gce-pd/node_ds.yaml",
-		"test/e2e/testing-manifests/storage-csi/gce-pd/controller_ss.yaml",
+		"test/e2e/testing-manifests/storage-csi/gce-pd/controller.yaml",
 	}
 
 	cleanup, err := utils.CreateFromManifests(f, driverNamespace, nil, manifests...)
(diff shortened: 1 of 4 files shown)


test/e2e/storage/drivers/csi.go
```diff
@@ -602,6 +606,26 @@ func (g *gcePDCSIDriver) PrepareTest(f *framework.Framework) (*testsuites.PerTes
...
+
+	waitErr := wait.PollImmediate(10*time.Second, csiPDDriverDeployTimeout, func() (bool, error) {
+		_, err := cs.AppsV1().StatefulSets(driverNamespace.Name).Get(context.TODO(), ssName, metav1.GetOptions{})
+		if err != nil && !apierrors.IsNotFound(err) {
+			return false, err
+		}
+		if err != nil {  // apierrors.IsNotFound(err)
+			return false, nil
+		}
+		return true, nil
```
msau42, line 621: This only waits for the statefulset object to exist, which should be immediately true once the create operation returns.

Do we want to wait for it and/or all the pods to be ready?

mattcary: Ah, that's my k8s ignorance. Yes, I think we should wait for the pods.

I can dig to figure out what pods should exist from a SS but if you have a pointer handy so much the better.

msau42: statefulset has `readyReplicas` that seem suitable: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#statefulset-v1-apps

mattcary: Thx for the pointer, used that field.

I was a little paranoid in the ready check, LMK if that's excessive.

mattcary:
Changed the PD CSI driver controller to a deployment, mutatis mutandis, and made the exported function more friendly by passing the driver name as a parameter.

This should permit it to be used externally, eg https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/551

test/e2e/storage/drivers/csi.go
```diff
@@ -602,6 +607,36 @@ func (g *gcePDCSIDriver) PrepareTest(f *framework.Framework) (*testsuites.PerTes
 	}, cleanupFunc
 }
 
+// WaitForPDCSIController waits for all replicas in the deployment at driverNamespace/name to be ready.
+func WaitForPDCSIController(driverNamespace *v1.Namespace, name string, cs clientset.Interface) error {
```
msau42, line 611: Oh if you want to export this, then maybe https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/deployment/wait.go would be a better place for such a util

test/e2e/storage/utils/create.go
```diff
@@ -355,8 +359,14 @@ func patchItemRecursively(f *framework.Framework, driverNamespace *v1.Namespace,
 		}
 	case *v1.Service:
 		PatchNamespace(f, driverNamespace, &item.ObjectMeta.Namespace)
-	case *appsv1.StatefulSet:
+	case *appsv1.Deployment:
```
msau42, line 376: I think csi hostpath uses StatefulSet, so we may need to add Deployment as an additional case.

test/e2e/testing-manifests/storage-csi/gce-pd/controller.yaml
```diff
@@ -25,6 +25,8 @@ spec:
           args:
             - "--v=5"
             - "--csi-address=/csi/csi.sock"
+            - "--leader-election"
+            - "--leader-election-namespace=$(PDCSI_NAMESPACE)"
```
msau42, line 29: Can namespace be set with downward API instead? Then we don't need to patch it in the manifest

test/e2e/storage/drivers/csi.go
mattcary: I was looking at your https://github.com/kubernetes/kubernetes/pull/93120 and trying to do the same thing.

I guess this function as written is general, but I can see in the future adding pd-specific stuff to it. LMK

test/e2e/storage/utils/create.go
mattcary: Yeah, I figured that out when the other tests started failing... :-/

test/e2e/testing-manifests/storage-csi/gce-pd/controller.yaml
mattcary: Good point.

I guess the pd csi driver yaml (from which this was copied) could be updated too

test/e2e/storage/drivers/csi.go
msau42: I would prefer to keep exported functions from k/k as provider-unspecific as possible.

test/e2e/testing-manifests/storage-csi/gce-pd/controller.yaml
mattcary: Oh, the pd csi driver was using the downardAPI. Hunh, I thought $(VAR) was only for make.

mattcary: Arg, the parens are because variables in args aren't evaluated in a bash-like context but some custom thing. Now I see. So many ways to step off the path.

mattcary:
This PR is probably the wrong approach as it doesn't wait for the CSINode objects to be created.

The existing waiting for those is enough AFAIU.

PR #90187, opened 2020-04-16
Implement server-side apply upgrade and downgrade

julianvmodesto:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

##### (1) Allow client-side to server-side apply upgrade.

Ensure that a user can change management of an object from client-side apply to
server-side apply without conflicts.

We do this by reading the last-applied-configuration annotation for client-side apply
to identify if conflicts can be forced because `kubectl apply` owns them.

##### (2) Allow server-side apply to client-side downgrade.

For an object managed with client-side apply, a user may upgrade to
managing the object with server-side apply, then decide to downgrade.

We can support this downgrade by keeping the last-applied-configuration
annotation for client-side apply updated with server-side apply.

This was suggested by @apelisse. This is an alternative to conflicting entirely and failing a downgrade.

See this KEP update: https://github.com/kubernetes/enhancements/pull/1382

**Which issue(s) this PR fixes**:

Related to https://github.com/kubernetes/kubernetes/issues/80916, https://github.com/kubernetes/kubernetes/issues/90146

**Does this PR introduce a user-facing change?**:

```release-note
Support a smooth upgrade from client-side apply to server-side apply without conflicts, as well as support the corresponding downgrade.
```

Issue #89954, opened 2020-04-08
before-first-apply manager prevents resource updates


dulltz:
**What happened**:

This problem is occurred when trying to do server-side apply without force option.
I cannot apply new patch to the resource which was created by client-side apply.
It is because `before-first-apply` manager prevents resource updates by other manager's server-side apply.

What is `before-first-apply` field manager?
Of course it can be patched if the force-conflicts option is added to the request, however is this the intended behavior of kube-apiserver?

**What you expected to happen**:

It is enable to do server-side apply without force option to the resource which was created by client-side apply.

**How to reproduce it (as minimally and precisely as possible)**:

1. Prepare pod manifest. 
```console
$ cat pause.yaml
apiVersion: v1
kind: Pod
metadata:
  name: pause
  namespace: sandbox
spec:
  containers:
  - image: k8s.gcr.io/pause:latest
    name: pause
```
2. Apply it by client-side-apply
```console
$ kubectl apply -f pause.yaml
pod/pause created
```
3. Apply the same manifest by server-side apply
```console
$ kubectl apply -f pause.yaml --server-side
pod/pause serverside-applied
```
4. Prepare the tweaked pod manifest.
```console
$ sed s/pause:latest/pause:3.1/g pause.yaml > pause-3-1.yaml
```
5. Apply the tweaked manifest by server-side apply without `--force-conflicts` option.
```console
$ kubectl apply -f pause-3-1.yaml --server-side
error: Apply failed with 1 conflict: conflict with "before-first-apply" using v1: .spec.containers[name="pause"].image

Please review the fields above--they currently have other managers. Here
are the ways you can resolve this warning:
* If you intend to manage all of these fields, please re-run the apply
  command with the `--force-conflicts` flag.
* If you do not intend to manage all of the fields, please edit your
  manifest to remove references to the fields that should keep their
  current managers.
* You may co-own fields by updating your manifest to match the existing
  value; in this case, you'll become the manager if the other manager(s)
  stop managing the field (remove it from their configuration).

See http://k8s.io/docs/reference/using-api/api-concepts/#conflicts
```

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): 
```console
$ kubectl version
[...]

PR #93162, opened 2020-07-16
Add tests for stopped managed fields tracking

julianvmodesto:
**What type of PR is this?**

/kind cleanup

This test verifies that client-side apply users who are upgrading to server-side-apply don't encounter conflicts for resources that were created with [the `before-first-apply` field manager](https://github.com/kubernetes/kubernetes/blob/36e40fb850293076b415ae3d376f5f81dc897105/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/skipnonapplied.go#L50) (https://github.com/kubernetes/kubernetes/issues/89954). Resources may have this field manager pre-1.18 (https://github.com/kubernetes/kubernetes/commit/a54a52c5de3458bfa5dbe1973d12584f59a5581c).

This is not an issue anymore because we use the last-applied configuration annotation to avoid conflicts now (https://github.com/kubernetes/kubernetes/pull/90187). The last-applied annotation contains any fields that would be owned by the before-first-apply field manager if CSA was used to create the object.

https://github.com/kubernetes/kubernetes/issues/89954

```release-note
NONE
```

/assign @apelisse 
/wg api-expression
/sig api-machinery

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/fieldmanager_test.go
+++ b/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/fieldmanager_test.go
@@ -111,6 +111,7 @@ func NewTestFieldManager(gvk schema.GroupVersionKind, chainFieldManager func(fie
 	f = fieldmanager.NewStripMetaManager(f)
 	f = fieldmanager.NewManagedFieldsUpdater(f)
 	f = fieldmanager.NewBuildManagerInfoManager(f, gvk.GroupVersion())
+	f = fieldmanager.NewProbabilisticSkipNonAppliedManager(f, &fakeObjectCreater{gvk: gvk}, gvk, fieldmanager.DefaultTrackOnCreateProbability)
 	f = fieldmanager.NewLastAppliedManager(f, typeConverter, objectConverter, gvk.GroupVersion())
 	f = fieldmanager.NewLastAppliedUpdater(f)
 	if chainFieldManager != nil {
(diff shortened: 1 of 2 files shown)


staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/fieldmanager_test.go
```diff
@@ -1039,6 +1039,72 @@ spec:
 	}
 }
 
+func TestUpgradeClientSideBeforeFirstApply(t *testing.T) {
+	f := NewDefaultTestFieldManager(schema.FromAPIVersionAndKind("apps/v1", "Deployment"))
```
julianvmodesto, line 1043: I also tried this test by adding the skipNonAppliedManager and it works.

```
	f := NewTestFieldManager(schema.FromAPIVersionAndKind("apps/v1", "Deployment"),
		func(m fieldmanager.Manager) fieldmanager.Manager {
			return fieldmanager.NewSkipNonAppliedManager(
				m,
				&fakeObjectCreater{gvk: schema.GroupVersionKind{Version: "apps/v1", Kind: "Deployment"}},
				schema.GroupVersionKind{},
			)
		})
```

liggitt:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/93162/pull-kubernetes-node-e2e/1283875284247908352 flake is #93210

apelisse:
Before we can merge that, I'd really like to see a PR that fixes the difference between "reset" and "stop tracking".

julianvmodesto:
Hm okay, so i've added an integration test as discussed at the wg meeting where we restart the apiserver with the feature gate off.

It looks like SSA fails, but we don't clear managed fields. Do we expect to clear managed fields, or only stop tracking managed fields?

kwiesmueller:
For a lot of features there are switches in the storage strategies that wipe the respective field if the feature is disabled *and* not present on the old object.
Which to me indicates, disabled stuff usually still stays in storage.
On the other hand, our field is very big and verbose and it might not hurt releasing the burden if people decide to disable the feature entirely.

julianvmodesto:
Ah okay, so we should remove managed fields from storage when the feature is disabled. I can see how this would be important if the feature is re-enabled, which might be dangerous since the field would not be accurate anymore.

julianvmodesto:
@apelisse for the issue around ensuring we stop tracking managed fields for large objects, there's an existing test for `managedFields: []` that I've updated that passes.

julianvmodesto:
Oops, @apelisse pointed out that [we clear managed fields on Update.](https://github.com/kubernetes/kubernetes/blob/release-1.18/staging/src/k8s.io/apiserver/pkg/registry/rest/update.go#L107-L111)

All new tests pass so everything seems to still work :)

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/fieldmanager_test.go
```diff
@@ -1039,6 +1040,145 @@ spec:
...
+	}
+
+	// stop tracking managed fields
+	newObj = &unstructured.Unstructured{Object: map[string]interface{}{}}
+	deployment = []byte(`
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: my-deployment
+  managedFields: [] # stop tracking managed fields
```
apelisse, line 1075: interesting, does this actually stop the tracking?

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/fieldmanager_test.go
```diff
@@ -1039,6 +1040,145 @@ spec:
...
+	if err := f.Apply(newObj, "test_server_side_apply_without_upgrade", false); err != nil {
+		t.Errorf("error applying object: %v", err)
+	}
+	if m := f.ManagedFields(); len(m) == 0 {
+		t.Errorf("expected to start tracking managed fields, but got: %v", m)
+	}
+	if e, a := "test_server_side_apply_without_upgrade", f.ManagedFields()[0].Manager; e != a {
+		t.Fatalf("exected first manager name to be %v, but got %v: %#v", e, a, f.ManagedFields())
+	}
+	if e, a := "before-first-apply", f.ManagedFields()[1].Manager; e != a {
```
apelisse, line 1148: Since you're already checking the length before that, can you make sure that this is not going to crash by checking you have at least 2 items?

julianvmodesto: Ah done

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/fieldmanager_test.go
julianvmodesto: Yes, confirmed also in this integration test:

https://github.com/kubernetes/kubernetes/blob/f2deb2417a6c542c54606ab17376b26ef1552b87/test/integration/apiserver/apply/apply_test.go#L1583-L1665

julianvmodesto: Just a note that it's possible to only clear/stop tracking managed fields like this on an update, but not on a create.

apelisse (approved):
/lgtm
/approve

PR #93087, opened 2020-07-15
Move the StorageVersions types to kube-apiserver

roycaihw:
/kind api-change

Move the internal types to kube-apiserver, and the external types to k8s.io/api. Ref https://github.com/kubernetes/kubernetes/pull/92064#discussion_r447795942

The first three commits are based on https://github.com/kubernetes/kubernetes/pull/93164

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

/sig api-machinery
/priority important-soon
/assign @caesarxuchao

PR #93164, opened 2020-07-16
Use canonical import path in client-gen input-base

roycaihw:
/kind bug

The [parameter](https://github.com/kubernetes/kubernetes/blob/fbbc3ed3595287cc8daf684a39ad9c26a90ca7eb/staging/src/k8s.io/code-generator/cmd/client-gen/args/args.go#L80) `--input-base` in client-gen is used to [compose](https://github.com/kubernetes/kubernetes/blob/fbbc3ed3595287cc8daf684a39ad9c26a90ca7eb/staging/src/k8s.io/code-generator/cmd/client-gen/args/gvpackages.go#L129) the [full qualified package path](https://github.com/kubernetes/kubernetes/blob/fbbc3ed3595287cc8daf684a39ad9c26a90ca7eb/staging/src/k8s.io/code-generator/cmd/client-gen/types/types.go#L47-L48), which is later used with `gengo` to [locate the package source and extract package comments](https://github.com/kubernetes/kubernetes/blob/fbbc3ed3595287cc8daf684a39ad9c26a90ca7eb/staging/src/k8s.io/code-generator/cmd/client-gen/generators/client_generator.go#L282-L283). 

Because `gengo` always [canonicalizes](https://github.com/kubernetes/gengo/blob/c0d492a0f3cab707b841450d5b5919720682da31/parser/parse.go#L854-L862) a package path before [adding it to the universe](https://github.com/kubernetes/gengo/blob/c0d492a0f3cab707b841450d5b5919720682da31/parser/parse.go#L259-L290), the universe doesn't understand non-canonical paths and will [return a empty package](https://github.com/kubernetes/gengo/blob/c0d492a0f3cab707b841450d5b5919720682da31/types/types.go#L267-L285). As a result the package comments were lost and group overrides didn't happen for APIs under `k8s.io/kubernetes/vendor/k8s.io/api`.

It was okay for most APIs because the [old group name](https://github.com/kubernetes/kubernetes/blob/fbbc3ed3595287cc8daf684a39ad9c26a90ca7eb/hack/update-codegen.sh#L47) (e.g. "**flowcontrol**/v1alpha1") is the same as the [override group name](https://github.com/kubernetes/kubernetes/blob/fbbc3ed3595287cc8daf684a39ad9c26a90ca7eb/staging/src/k8s.io/code-generator/cmd/client-gen/generators/client_generator.go#L285) ("**flowcontrol**.apiserver.k8s.io/v1alpha1"). However, for the StorageVersion API the old group name (**apiserverinternal**/v1alpha1) and the override group name (**internal**.apiserver.k8s.io/v1alpha1) are different (ref https://github.com/kubernetes/kubernetes/pull/93087). This caused error  because other generators like [informer-gen](https://github.com/kubernetes/kubernetes/blob/fbbc3ed3595287cc8daf684a39ad9c26a90ca7eb/staging/src/k8s.io/code-generator/cmd/informer-gen/generators/packages.go#L125) honored the package comments. 

```release-note
NONE
```

/sig api-machinery
/cc @caesarxuchao 
/priority important-soon

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -59,7 +59,7 @@ GV_DIRS_CSV=$(IFS=',';echo "${GV_DIRS[*]// /,}";IFS=$)
 
 # This can be called with one flag, --verify-only, so it works for both the
 # update- and verify- scripts.
-${clientgen} --output-base "${KUBE_ROOT}/vendor" --output-package="k8s.io/client-go" --clientset-name="kubernetes" --input-base="k8s.io/kubernetes/vendor/k8s.io/api" --input="${GV_DIRS_CSV}" --go-header-file "${KUBE_ROOT}/hack/boilerplate/boilerplate.generatego.txt" "$@"
+${clientgen} --output-base "${KUBE_ROOT}/vendor" --output-package="k8s.io/client-go" --clientset-name="kubernetes" --input-base="k8s.io/api" --input="${GV_DIRS_CSV}" --go-header-file "${KUBE_ROOT}/hack/boilerplate/boilerplate.generatego.txt" "$@"
 
 listergen_external_apis=()
 kube::util::read-array listergen_external_apis < <(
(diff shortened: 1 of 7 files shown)


caesarxuchao:
cc @sttts

sttts:
I don't see any fixed output of the generators. Where can I see that?

roycaihw:
> I don't see any fixed output of the generators. Where can I see that?

It doesn't affect most of the existing APIs as I explained in https://github.com/kubernetes/kubernetes/pull/93164#issue-450572649. That's why there is no change in the generated clients.

It affects https://github.com/kubernetes/kubernetes/pull/93087. Here is the difference w/ & w/out this change: https://github.com/roycaihw/kubernetes/compare/storage-version/move-types...roycaihw:storage-version/move-types-without-clientgen-fix

liggitt:

Issue #93042, opened 2020-07-13
[Flaky Test] The framework ClientSet must not be nil at this point


MikeSpreitzer:
**Which jobs are flaking**:
pull-kubernetes-e2e-gce

**Which test(s) are flaking**:
CSI  Volumes [Driver: csi-hostpath] [Testpattern: Dynamic PV (default fs)]  subPath should support existing directories when readOnly specified in  the volumeSource 

**Testgrid link**:
https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce

**Reason for failure**:
Unknown.

**Anything else we need to know**:
- https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&text=The%20framework%20ClientSet%20must%20not%20be%20nil%20at%20this%20point
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/91177/pull-kubernetes-e2e-gce/1282707814711562240/

MikeSpreitzer:
Here is the stdout from the failure cited:
```
[BeforeEach] [Testpattern: Dynamic PV (default fs)] subPath
  test/e2e/storage/testsuites/base.go:127
[BeforeEach] [Testpattern: Dynamic PV (default fs)] subPath
  test/e2e/framework/framework.go:174
ï¿½[1mSTEPï¿½[0m: Creating a kubernetes client
Jul 13 17:26:37.008: INFO: >>> kubeConfig: /workspace/.kube/config
ï¿½[1mSTEPï¿½[0m: Building a namespace api object, basename provisioning
ï¿½[1mSTEPï¿½[0m: Binding the e2e-test-privileged-psp PodSecurityPolicy to the default service account in provisioning-4507
ï¿½[1mSTEPï¿½[0m: Waiting for a default service account to be provisioned in namespace
[It] should support existing directories when readOnly specified in the volumeSource
  test/e2e/storage/testsuites/subpath.go:391
ï¿½[1mSTEPï¿½[0m: Building a driver namespace object, basename provisioning
ï¿½[1mSTEPï¿½[0m: Binding the e2e-test-privileged-psp PodSecurityPolicy to the default service account in provisioning-7569
ï¿½[1mSTEPï¿½[0m: Waiting for a default service account to be provisioned in namespace
ï¿½[1mSTEPï¿½[0m: deploying csi-hostpath driver
[...]

PR #88936, opened 2020-03-07
Add nested tracing and endpoint filter tracing to apiserver 

jpbetz:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

Leverage https://github.com/kubernetes/utils/pull/139 to nest traces.

Also add tracing to the filter chain so that http requests are traced comprehensively. Includes adding the authentication and authorization checks as a steps in the HTTP level trace.

Example of the tracing output when an etcd list operation, authn and authz are slow:

```
Trace[1374095871]: "HTTP Request" method:GET,url:/api/v1/namespaces/kube-system/resourcequotas,user-agent:kube-apiserver/v1.19.0 (linux/amd64) kubernetes/482c37a,client:::1 (02-Jun-2020 10:38:00.615) (total time: 3201ms):
Trace[1374095871]: ---"Authenticate check done" 1000ms (10:38:00.615)
Trace[1374095871]: ---"Authorize check done" 1000ms (10:38:00.616)
Trace[1374095871]: ["List" url:/api/v1/namespaces/kube-system/resourcequotas,user-agent:kube-apiserver/v1.19.0 (linux/amd64) kubernetes/482c37a,client:::1 1200ms (10:38:00.616)
Trace[1374095871]:  ["List etcd3" key:/resourcequotas/kube-system,resourceVersion:,limit:0,continue: 1200ms (10:38:00.616)]]
Trace[1374095871]: [3.201532894s] [3.201532894s] END
```

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Trace output in apiserver logs is more organized and comprehensive. Traces are nested, and for all non-long running request endpoints, the entire filter chain is instrumented (e.g. authentication check is included).
```
/sig api-machinery
/priority important-longterm
/cc @SomtochiAma @lavalamp

PR #91634, opened 2020-06-01
Bump k8s.io/utils to 6e3d28 to pick up nested trace

jpbetz:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Picks up nested tracing support (https://github.com/kubernetes/utils/pull/139) which is needed by https://github.com/kubernetes/kubernetes/pull/88936

**Special notes for your reviewer**:

```
hack/pin-dependency.sh k8s.io/utils 24a76e807f40c05aee600673f7f38f88d0ab254f
hack/update-vendor.sh
```

**Does this PR introduce a user-facing change?**:

no

```release-note
NONE
```

/priority important-longterm
/sig api-machinery
/cc @sttts @SomtochiAma @thockin

PR #93170, opened 2020-07-17
Prevent excessive logging in e2e tests of traces added by PR#88936

jpbetz:
**What type of PR is this?**

/kind flake

Disables (for e2e tests) the request tracing that was added in https://github.com/kubernetes/kubernetes/pull/88936 to try to helps with flakes reported in https://github.com/kubernetes/kubernetes/issues/93042

Even though the timeout of the newly added trace is 30s, e2e tests have verbosity set up pretty high, and the way tracing works is that when verbosity is above 4, all traces using "LogIfLong" printed, even if the duration is less than the "logIfLong" threshold.

So as mitigation, this only calls "LogIfLong" for request traces if the threshold has actually been exceeded, which prevents it all request traces from being logged by e2e tests.

```release-note
NONE
```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/endpoints/filters/trace.go
+++ b/staging/src/k8s.io/apiserver/pkg/endpoints/filters/trace.go
@@ -43,6 +43,7 @@ func WithTrace(handler http.Handler, longRunningCheck genericapirequest.LongRunn
 			handler.ServeHTTP(w, req)
 			return
 		}
+		// Set trace as root trace in context for nested tracing
 		ctx, trace := genericapirequest.WithTrace(req.Context(), "HTTP Request",
 			utiltrace.Field{Key: "method", Value: req.Method},
 			utiltrace.Field{Key: "url", Value: req.URL.Path},
@@ -56,8 +57,16 @@ func WithTrace(handler http.Handler, longRunningCheck genericapirequest.LongRunn
 			utiltrace.Field{Key: "user-agent", Value: &internal.LazyTruncatedUserAgent{req}},
 			utiltrace.Field{Key: "client", Value: &internal.LazyClientIP{req}})
 		req = req.Clone(ctx)
-		// Set trace as root trace in context for nested tracing
-		defer trace.LogIfLong(30 * time.Second)
+		defer func(start time.Time) {
+			if time.Since(start) >= 30*time.Second {
+				// To mitigate https://github.com/kubernetes/kubernetes/issues/93042
+				// don't call LogIfLong here unless threshold is exceeded. This avoid printing
+				// this trace unconditionally when verbosity is set above 4, which is the case for
+				// e2e tests. LogIfLong called here instead of Log so that individual steps are
+				// only logged if step threshold is exceeded.
+				trace.LogIfLong(30 * time.Second)
+			}
+		}(time.Now())
 		handler.ServeHTTP(w, req)
 	})
 }


staging/src/k8s.io/apiserver/pkg/endpoints/filters/trace.go
```diff
@@ -56,8 +57,16 @@ func WithTrace(handler http.Handler, longRunningCheck genericapirequest.LongRunn
 			utiltrace.Field{Key: "user-agent", Value: &internal.LazyTruncatedUserAgent{req}},
 			utiltrace.Field{Key: "client", Value: &internal.LazyClientIP{req}})
 		req = req.Clone(ctx)
-		// Set trace as root trace in context for nested tracing
-		defer trace.LogIfLong(30 * time.Second)
+		defer func(start time.Time) {
+			if time.Since(start) >= 30*time.Second {
```
lavalamp, line 61: I'm not sure I follow this, is there like both a step trigger and an overall trigger or something?

staging/src/k8s.io/apiserver/pkg/endpoints/filters/trace.go
```diff
@@ -56,8 +57,16 @@ func WithTrace(handler http.Handler, longRunningCheck genericapirequest.LongRunn
...
+				// To mitigate https://github.com/kubernetes/kubernetes/issues/93042
+				// don't call LogIfLong here unless threshold is exceeded. This avoid printing
+				// this trace unconditionally when verbosity is set above 4, which is the case for
+				// e2e tests. LogIfLong called here instead of Log so that individual steps are
+				// only logged if step threshold is exceeded.
+				trace.LogIfLong(30 * time.Second)
```
lavalamp, line 67: 30s is super long, is this really the minimum? I think the best is to try to log the longest 1% - .1% of calls. That way it's a constant amount of logging and it's the most useful calls to log. Could do this by looking at the latency metric and logging if we're above 99th percentile, or something.

@wojtek-t might have thoughts.

lavalamp: (I'm worried that this will make the trace essentially never show, which also isn't good?)

liggitt: the 30 second threshhold is preexisting IIUC

what is really unexpected to me is that LogIfLong **ignores** the specified duration at high verbosity levels

liggitt: > what is really unexpected to me is that LogIfLong ignores the specified duration at high verbosity levels

it looks like prior to https://github.com/kubernetes/kubernetes/commit/52838be7b22c53050d4e397c221f0b3c3842122c#diff-06ce45a2e81bbfd5d4ce7e3c67d49a84L120 that was not the case ... that seems like it should be reverted

liggitt: https://github.com/kubernetes/utils/pull/139/files#r456819382 was the location of the change in k8s.io/utils

liggitt:
> e2e tests have verbosity set up pretty high, and the way tracing works is that when verbosity is above 4, all traces using "LogIfLong" printed, even if the duration is less than the "logIfLong" threshold.

I think that is a behavior change introduced in https://github.com/kubernetes/kubernetes/pull/91634 ... prior to that, V(4) meant that all steps would be logged for traces which exceeded the threshold. Now, all steps of all traces are logged even if the overall trace did not exceed the threshold. That seems likely to dramatically increase log volume.

justaugustus:
FYI @kubernetes/ci-signal

staging/src/k8s.io/apiserver/pkg/endpoints/filters/trace.go
wojtek-t: I agree with Jordan - the LogIfLong shouldn't ignore the passed threshold. And that is the thing that we should fix instead.

Re logging X% of calls - I'm not convinced. If everything is healthy we probably don't want to trace anything and that may result in significant increase in log volume in large clusters. We may get back to this discussion, but I think that we should first fix the regression Jordan pointed out above.

liggitt: > I agree with Jordan - the LogIfLong shouldn't ignore the passed threshold. And that is the thing that we should fix instead.

fixed in https://github.com/kubernetes/utils/pull/177 and pulled into k/k in https://github.com/kubernetes/kubernetes/pull/93253

jpbetz:
Superseded by https://github.com/kubernetes/kubernetes/pull/93253

PR #93171, opened 2020-07-17
[Follow #92720]: fix component status server address

SataQiu:
**What type of PR is this?**
/kind bug

[Follow #92720]: fix component status server address.
As #92720 has been merged, the insecure ports are unavailable, we should change to use the secure ports instead.
It is not the best solution, but make it work like before. 
The `ComponentStatus` may be removed or refactored in the future.
For more information, see kubernetes/enhancements#553

Ref kubernetes/kubeadm#2222
Ref kubernetes/enhancements#553

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/registry/core/rest/storage_core.go
+++ b/pkg/registry/core/rest/storage_core.go
@@ -17,6 +17,7 @@ limitations under the License.
 package rest
 
 import (
+	"crypto/tls"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


fedebongio:
/priority backlog
there has been some discussion in the triage meeting about the deprecation of the feature

deads2k:
Given that the maintainers of the feature believe the design for this "feature" is backwards and is not generally applicable, I would not expect continued improvements or reviews on the feature.  If its capability continues to degrade, given an ultimate goal of removal, I don't think we mind and want to improve or encourage its usage.

I recommend closure without merge

PR #93176, opened 2020-07-17
Fix for host path test case

yuzhiquan:
**What type of PR is this?**

/kind bug

- if err!=nil,plug may be nil, so need finish test case
- replace string with const variable

Fixes #

```release-note
NONE
```

```docs

```
/priority backlog

### DIFF
--- a/pkg/volume/hostpath/host_path.go
+++ b/pkg/volume/hostpath/host_path.go
@@ -160,7 +160,7 @@ func (plugin *hostPathPlugin) NewDeleter(spec *volume.Spec) (volume.Deleter, err
 
 func (plugin *hostPathPlugin) NewProvisioner(options volume.VolumeOptions) (volume.Provisioner, error) {
 	if !plugin.config.ProvisioningEnabled {
-		return nil, fmt.Errorf("Provisioning in volume plugin %q is disabled", plugin.GetPluginName())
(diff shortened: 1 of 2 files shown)


jsafrane:
The code looks OK, however, can you please change the commit message to something like

> Fixed potential nil dereference in hostpath unit test
>
> And fixed an error message when at it.

Btw, using two commits would be perfectly OK here.

yuzhiquan:
Respond.

yuzhiquan:
@jsafrane should this pr add milestone tag, or wait for code freeze finished?

jsafrane:
I think we can wait for 1.20. It should be open soon.

yuzhiquan:
> I think we can wait for 1.20. It should be open soon.

Why still stuck in `Must be in milestone v1.19.` after 1.19 released? or just lack milestone label?

yuzhiquan:
@justaugustus Hi,Stephen, can you  addres 1.20 label for this pr?

justaugustus:
Sure thing!
/milestone v1.20

yuzhiquan:
> Sure thing!
> /milestone v1.20
> /test pull-kubernetes-e2e-kind-ipv6

Thanks for so quickly reply!

PR #93177, opened 2020-07-17
Use temporary directory when creating temporary file in tests

soltysh:
**What type of PR is this?**
/kind cleanup

/sig cli
/priority backlog

[ioutil.TempFile](https://golang.org/pkg/io/ioutil/#TempFile) creates a temporary file in the directory specified as the first argument. It sometimes is `/tmp` but sometimes not. It's better to be explicit about this being `/tmp`. 

/assign @damemi 

```release-note
NONE
```

### DIFF
--- a/staging/src/k8s.io/kubectl/pkg/cmd/config/config_test.go
+++ b/staging/src/k8s.io/kubectl/pkg/cmd/config/config_test.go
@@ -261,7 +261,7 @@ func TestAdditionalAuth(t *testing.T) {
 }
 
 func TestEmbedClientCert(t *testing.T) {
-	fakeCertFile, _ := ioutil.TempFile("", "")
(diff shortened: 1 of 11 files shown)


damemi (approved):
/lgtm

brianpursley:
I think the change is fine, but I don't see how this will be different than what it does currently.

From https://golang.org/pkg/io/ioutil/#TempFile ...
> If dir is the empty string, TempFile uses the default directory for temporary files (see os.TempDir)

soltysh:
> I think the change is fine, but I don't see how this will be different than what it does currently.

We want to ensure the tests are running in a contained environment, here `/tmp` directory. Without it, according to golang description `ioutil.TempFile` creates the temporary file in the directory specified as the first argument, which being `""` is usually current.

brianpursley (approved):

Issue #85677, opened 2019-11-27
Scheduler does not clear nominatedNodeName on Pod when node is no longer valid


losipiuk:
**What happened**:

Scheduler decided that it wants to schedule Pod P on Node N after preempting some pods from N.
It set the P.nominatedNodeName to N. After a while (before scheduling actually happened) node N was removed from the cluster.
Scheduler did not reset P.nominatedNodeName to empty value.

**What you expected to happen**:

I would expect scheduler to unset P.nominatedNodeName when the node is no longer valid.
The trivial reason for node not being valid any more is it is removed from the cluster. 
But there are more cases to cover. Out of the top of my head I may think of:
- extra Pod was scheduled to node and there is not enough resources any more
  - either scheduled by scheduler
  - or directly as static pod/daemonset
- node became unschedulable

**How to reproduce it (as minimally and precisely as possible)**:

We did not try to reproduce issue manually. Yet following setup should work.
- create cluster with nodes A and B.
- ensure all system pods are running on A. Fill up A so it does not have spare resources.
- add a low priority Pod L with long termination, with requests matching allocatable of B. It should be scheduled to B.
- add a high priority Pod H. It should get B as nominatedNodeName
- kill node B
- the H.nominatedNodeName will still point to B which is no longer in the cluster

**Anything else we need to know?**:

Current behavior does not play well with Cluster Autoscaler which assumes that Pod with nominatedNodeName will be scheduled in a while.
Therefore such pods do not trigger scale-up. If we can have P.nominatedNodeName set incorrectly for unbounded period of time CA will never
provision node for P to run.

**Environment**:
- Kubernetes version (use `kubectl version`):

We observed the behavior during presubmit tests which run k8s built from HEAD of master on 27.11.2019.

Versions logged in test log:
I1125 18:55:11.691] Client Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.0-alpha.0.1183+414202578ba1ad", GitCommit:"414202578ba1ad53a26bc5126b3828a3a410097e", GitTreeState:"clean", BuildDate:"2019-11-25T17:04:18Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #93179, opened 2020-07-17
prefer nominated node - IMPL

### LINKED ISSUE #93013: Honor "NominatedNodeName" during the new scheduling cycle
A option of scheduler used to check the feasibility of `NominatedNodeName` directly instead of loop all the nodes in the cluster after pod preemption is done previously.

**Why is this needed**:
When the preemption is happened a best candidate is chosen from current nodes in the cluster, i.e. the feasibility and prioritization has been thoroughly considered. 

The best candidate is patched to be the pod's field of `NominatedNodeName`, after the new scheduling cycle is started, the "NominatedNodeName" **mostly** like is still the best candidate for the preemptor pod, but in the new scheduling cycle, all nodes are added to check the feasibility again.

Given in the a large cluster with lots of nodes, how it does like this make the efforts unnecessary heaven, we could add a new option so that we can validate the `NominatedNodeName` directly instead of looping all the nodes, and fall back to check other nodes if the `NominatedNodeName` can not pass all the filters again.

pros:
Save scheduling from running all the filter plugins for all the nodes twice.

cons:
There might be other nodes are more suitable for scheduling the pod than the `NominatedNodeName` node, if the are some big changes happened to that **nodes** after the new scheduling cycle but before the `NominatedNodeName` is re-examined.

chendave:
Performance evaluation based on the metric of `scheduler_framework_extension_point_duration_seconds/Filter` in the case of `preemption` when the feature gate is enabled, workloads configuration is 500Nodes, 2000InitPods and 500PodsToSchedule.

xref: https://github.com/kubernetes/kubernetes/pull/96258

indicator  |          before (ms)           |  after  (ms)                      | change
-----------|---------------------------| -------------------------- |---------
Average   | 0.46601329799999947  | 0.3219608919999995     |  -30.9%
Perc50      | 0.2828025477707007    | 0.1457317073170732    |  -48.4%
Perc90      |1.2829493087557604     | 1.0933940774487472    |  -14.8%
Perc99      | 1.4672811059907833    | 1.366742596810934      |  -6.9%

- backup for the legacy approach - component based config and skip the schedule cycle when the nominated node is not fit.
   legacy metric name is `scheduling_algorithm_predicate_evaluation_seconds`, workloads configuration is 500Nodes, 2000InitPods and 500PodsToSchedule.

indicator  |          before (ms)            |  after (ms)                     | change
-----------|---------------------------| --------------------------|---------
Average   |  0.40068560239999973  | 0.20466781999999958 | -48.9%
Perc50      | 0.5357908272610373    | 0.5070993914807302   | -5.3%
Perc90      | 0.9644234890698671     | 0.9127789046653144  | -5.4%
Perc99      | 1.2937062937062935    | 0.9634888438133874   | -25.5%

**What type of PR is this?**
 /kind feature

Fixes #93013 

```release-note
Overall, enable the feature of `PreferNominatedNode` will  improve the performance of scheduling where preemption might frequently happen, but in theory, enable the feature of `PreferNominatedNode`, the pod might not be scheduled to the best candidate node in the cluster.
```

Please use the following format for linking documentation:
-  KEP:  https://github.com/kubernetes/enhancements/pull/2026

### DIFF
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -269,6 +269,16 @@ const (
 	// Enables generic ephemeral inline volume support for pods
 	GenericEphemeralVolume featuregate.Feature = "GenericEphemeralVolume"
 
+	// owner: @chendave
+	// alpha: v1.21
+	//
+	// PreferNominatedNode tells scheduler whether the nominated node will be checked first before looping
+	// all the rest of nodes in the cluster.
+	// Enabling this feature also implies the preemptor pod might not be dispatched to the best candidate in
+	// some corner case, e.g. another node releases enough resources after the nominated node has been set
+	// and hence is the best candidate instead.
+	PreferNominatedNode featuregate.Feature = "PreferNominatedNode"
+
 	// owner: @tallclair
 	// alpha: v1.12
 	// beta:  v1.14
@@ -779,6 +789,7 @@ var defaultKubernetesFeatureGates = map[featuregate.Feature]featuregate.FeatureS
 	GracefulNodeShutdown:                           {Default: false, PreRelease: featuregate.Alpha},
 	ServiceLBNodePortControl:                       {Default: false, PreRelease: featuregate.Alpha},
 	MixedProtocolLBService:                         {Default: false, PreRelease: featuregate.Alpha},
+	PreferNominatedNode:                            {Default: false, PreRelease: featuregate.Alpha},
 
 	// inherited features from generic apiserver, relisted here to get a conflict if it is changed
 	// unintentionally on either side:
--- a/pkg/scheduler/core/generic_scheduler.go
+++ b/pkg/scheduler/core/generic_scheduler.go
@@ -27,7 +27,9 @@ import (
 	"k8s.io/klog/v2"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apiserver/pkg/util/feature"
 	extenderv1 "k8s.io/kube-scheduler/extender/v1"
+	"k8s.io/kubernetes/pkg/features"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
 	"k8s.io/kubernetes/pkg/scheduler/framework/runtime"
 	internalcache "k8s.io/kubernetes/pkg/scheduler/internal/cache"
@@ -196,6 +198,26 @@ func (g *genericScheduler) numFeasibleNodesToFind(numAllNodes int32) (numNodes i
 	return numNodes
 }
 
+func (g *genericScheduler) evaluateNominatedNode(ctx context.Context, pod *v1.Pod, fwk framework.Framework, state *framework.CycleState, filteredNodesStatuses framework.NodeToStatusMap) ([]*v1.Node, error) {
+	nnn := pod.Status.NominatedNodeName
+	nodeInfo, err := g.nodeInfoSnapshot.Get(nnn)
+	if err != nil {
+		return nil, err
+	}
+	node := []*framework.NodeInfo{nodeInfo}
+	feasibleNodes, err := g.findNodesThatPassFilters(ctx, fwk, state, pod, filteredNodesStatuses, node)
+	if err != nil {
+		return nil, err
+	}
+
+	feasibleNodes, err = g.findNodesThatPassExtenders(pod, feasibleNodes, filteredNodesStatuses)
+	if err != nil {
+		return nil, err
+	}
+
+	return feasibleNodes, nil
+}
+
 // Filters the nodes to find the ones that fit the pod based on the framework
 // filter plugins and filter extenders.
 func (g *genericScheduler) findNodesThatFitPod(ctx context.Context, fwk framework.Framework, state *framework.CycleState, pod *v1.Pod) ([]*v1.Node, framework.NodeToStatusMap, error) {
@@ -203,27 +225,38 @@ func (g *genericScheduler) findNodesThatFitPod(ctx context.Context, fwk framewor
 
 	// Run "prefilter" plugins.
 	s := fwk.RunPreFilterPlugins(ctx, state, pod)
+	allNodes, err := g.nodeInfoSnapshot.NodeInfos().List()
+	if err != nil {
+		return nil, nil, err
+	}
 	if !s.IsSuccess() {
 		if !s.IsUnschedulable() {
 			return nil, nil, s.AsError()
 		}
 		// All nodes will have the same status. Some non trivial refactoring is
 		// needed to avoid this copy.
-		allNodes, err := g.nodeInfoSnapshot.NodeInfos().List()
-		if err != nil {
-			return nil, nil, err
-		}
 		for _, n := range allNodes {
 			filteredNodesStatuses[n.Node().Name] = s
 		}
 		return nil, filteredNodesStatuses, nil
 	}
 
-	feasibleNodes, err := g.findNodesThatPassFilters(ctx, fwk, state, pod, filteredNodesStatuses)
+	// "NominatedNodeName" can potentially be set in a previous scheduling cycle as a result of preemption.
+	// This node is likely the only candidate that will fit the pod, and hence we try it first before iterating over all nodes.
+	if len(pod.Status.NominatedNodeName) > 0 && feature.DefaultFeatureGate.Enabled(features.PreferNominatedNode) {
+		feasibleNodes, err := g.evaluateNominatedNode(ctx, pod, fwk, state, filteredNodesStatuses)
+		if err != nil {
+			klog.ErrorS(err, "Evaluation failed on nominated node", "pod", klog.KObj(pod), "node", pod.Status.NominatedNodeName)
+		}
+		// Nominated node passes all the filters, scheduler is good to assign this node to the pod.
+		if len(feasibleNodes) != 0 {
+			return feasibleNodes, filteredNodesStatuses, nil
+		}
+	}
+	feasibleNodes, err := g.findNodesThatPassFilters(ctx, fwk, state, pod, filteredNodesStatuses, allNodes)
 	if err != nil {
 		return nil, nil, err
 	}
-
 	feasibleNodes, err = g.findNodesThatPassExtenders(pod, feasibleNodes, filteredNodesStatuses)
 	if err != nil {
 		return nil, nil, err
@@ -232,22 +265,23 @@ func (g *genericScheduler) findNodesThatFitPod(ctx context.Context, fwk framewor
 }
 
 // findNodesThatPassFilters finds the nodes that fit the filter plugins.
-func (g *genericScheduler) findNodesThatPassFilters(ctx context.Context, fwk framework.Framework, state *framework.CycleState, pod *v1.Pod, statuses framework.NodeToStatusMap) ([]*v1.Node, error) {
-	allNodes, err := g.nodeInfoSnapshot.NodeInfos().List()
-	if err != nil {
-		return nil, err
-	}
-
-	numNodesToFind := g.numFeasibleNodesToFind(int32(len(allNodes)))
+func (g *genericScheduler) findNodesThatPassFilters(
+	ctx context.Context,
+	fwk framework.Framework,
+	state *framework.CycleState,
+	pod *v1.Pod,
+	statuses framework.NodeToStatusMap,
+	nodes []*framework.NodeInfo) ([]*v1.Node, error) {
+	numNodesToFind := g.numFeasibleNodesToFind(int32(len(nodes)))
 
 	// Create feasible list with enough space to avoid growing it
 	// and allow assigning.
 	feasibleNodes := make([]*v1.Node, numNodesToFind)
 
 	if !fwk.HasFilterPlugins() {
-		length := len(allNodes)
+		length := len(nodes)
 		for i := range feasibleNodes {
-			feasibleNodes[i] = allNodes[(g.nextStartNodeIndex+i)%length].Node()
+			feasibleNodes[i] = nodes[(g.nextStartNodeIndex+i)%length].Node()
 		}
 		g.nextStartNodeIndex = (g.nextStartNodeIndex + len(feasibleNodes)) % length
 		return feasibleNodes, nil
@@ -260,7 +294,7 @@ func (g *genericScheduler) findNodesThatPassFilters(ctx context.Context, fwk fra
 	checkNode := func(i int) {
 		// We check the nodes starting from where we left off in the previous scheduling cycle,
 		// this is to make sure all nodes have the same chance of being examined across pods.
-		nodeInfo := allNodes[(g.nextStartNodeIndex+i)%len(allNodes)]
+		nodeInfo := nodes[(g.nextStartNodeIndex+i)%len(nodes)]
 		status := fwk.RunFilterPluginsWithNominatedPods(ctx, state, pod, nodeInfo)
 		if status.Code() == framework.Error {
 			errCh.SendErrorWithCancel(status.AsError(), cancel)
@@ -292,9 +326,9 @@ func (g *genericScheduler) findNodesThatPassFilters(ctx context.Context, fwk fra
 
 	// Stops searching for more nodes once the configured number of feasible nodes
 	// are found.
-	parallelize.Until(ctx, len(allNodes), checkNode)
+	parallelize.Until(ctx, len(nodes), checkNode)
 	processedNodes := int(feasibleNodesLen) + len(statuses)
-	g.nextStartNodeIndex = (g.nextStartNodeIndex + processedNodes) % len(allNodes)
+	g.nextStartNodeIndex = (g.nextStartNodeIndex + processedNodes) % len(nodes)
 
 	feasibleNodes = feasibleNodes[:feasibleNodesLen]
 	if err := errCh.ReceiveError(); err != nil {
--- a/pkg/scheduler/core/generic_scheduler_test.go
+++ b/pkg/scheduler/core/generic_scheduler_test.go
@@ -33,9 +33,12 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/informers"
 	clientsetfake "k8s.io/client-go/kubernetes/fake"
+	featuregatetesting "k8s.io/component-base/featuregate/testing"
 	pvutil "k8s.io/kubernetes/pkg/controller/volume/persistentvolume/util"
+	"k8s.io/kubernetes/pkg/features"
 	schedulerapi "k8s.io/kubernetes/pkg/scheduler/apis/config"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/defaultbinder"
@@ -1196,5 +1199,90 @@ func TestFairEvaluationForNodes(t *testing.T) {
 		if g.nextStartNodeIndex != (i+1)*nodesToFind%numAllNodes {
 			t.Errorf("got %d lastProcessedNodeIndex, want %d", g.nextStartNodeIndex, (i+1)*nodesToFind%numAllNodes)
 		}
+	}
+}
+
+func TestPreferNominatedNodeFilterCallCounts(t *testing.T) {
+	tests := []struct {
+		name                  string
+		feature               bool
+		pod                   *v1.Pod
+		nodeReturnCodeMap     map[string]framework.Code
+		expectedCount         int32
+		expectedPatchRequests int
+	}{
+		{
+			name:          "Enable the feature, pod has the nominated node set, filter is called only once",
+			feature:       true,
+			pod:           st.MakePod().Name("p_with_nominated_node").UID("p").Priority(highPriority).NominatedNodeName("node1").Obj(),
+			expectedCount: 1,
+		},
+		{
+			name:          "Disable the feature, pod has the nominated node, filter is called for each node",
+			feature:       false,
+			pod:           st.MakePod().Name("p_with_nominated_node").UID("p").Priority(highPriority).NominatedNodeName("node1").Obj(),
+			expectedCount: 3,
+		},
+		{
+			name:          "pod without the nominated pod, filter is called for each node",
+			feature:       true,
+			pod:           st.MakePod().Name("p_without_nominated_node").UID("p").Priority(highPriority).Obj(),
+			expectedCount: 3,
+		},
+		{
+			name:              "nominated pod cannot pass the filter, filter is called for each node",
+			feature:           true,
+			pod:               st.MakePod().Name("p_with_nominated_node").UID("p").Priority(highPriority).NominatedNodeName("node1").Obj(),
+			nodeReturnCodeMap: map[string]framework.Code{"node1": framework.Unschedulable},
+			expectedCount:     4,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			defer featuregatetesting.SetFeatureGateDuringTest(t, utilfeature.DefaultFeatureGate, features.PreferNominatedNode, test.feature)()
+			// create three nodes in the cluster.
+			nodes := makeNodeList([]string{"node1", "node2", "node3"})
+			client := &clientsetfake.Clientset{}
+			cache := internalcache.New(time.Duration(0), wait.NeverStop)
+			for _, n := range nodes {
+				cache.AddNode(n)
+			}
+			plugin := st.FakeFilterPlugin{FailedNodeReturnCodeMap: test.nodeReturnCodeMap}
+			registerFakeFilterFunc := st.RegisterFilterPlugin(
+				"FakeFilter",
+				func(_ runtime.Object, fh framework.Handle) (framework.Plugin, error) {
+					return &plugin, nil
+				},
+			)
+			registerPlugins := []st.RegisterPluginFunc{
+				st.RegisterQueueSortPlugin(queuesort.Name, queuesort.New),
+				registerFakeFilterFunc,
+				st.RegisterBindPlugin(defaultbinder.Name, defaultbinder.New),
+			}
+			fwk, err := st.NewFramework(
+				registerPlugins,
+				frameworkruntime.WithClientSet(client),
+				frameworkruntime.WithPodNominator(internalqueue.NewPodNominator()),
+			)
+			if err != nil {
+				t.Fatal(err)
+			}
+			snapshot := internalcache.NewSnapshot(nil, nodes)
+			scheduler := NewGenericScheduler(
+				cache,
+				snapshot,
+				[]framework.Extender{},
+				schedulerapi.DefaultPercentageOfNodesToScore).(*genericScheduler)
+
+			_, _, err = scheduler.findNodesThatFitPod(context.Background(), fwk, framework.NewCycleState(), test.pod)
+
+			if err != nil {
+				t.Errorf("unexpected error: %v", err)
+			}
+			if test.expectedCount != plugin.NumFilterCalled {
+				t.Errorf("predicate was called %d times, expected is %d", plugin.NumFilterCalled, test.expectedCount)
+			}
+		})
 	}
 }
--- a/test/integration/scheduler/preemption_test.go
+++ b/test/integration/scheduler/preemption_test.go
@@ -1314,3 +1314,152 @@ func TestPDBInPreemption(t *testing.T) {
 		})
 	}
 }
+
+func initTestPreferNominatedNode(t *testing.T, nsPrefix string, opts ...scheduler.Option) *testutils.TestContext {
+	testCtx := testutils.InitTestSchedulerWithOptions(t, testutils.InitTestMaster(t, nsPrefix, nil), nil, opts...)
+	testutils.SyncInformerFactory(testCtx)
+	// wraps the NextPod() method to make it appear the preemption has been done already and the nominated node has been set.
+	f := testCtx.Scheduler.NextPod
+	testCtx.Scheduler.NextPod = func() (podInfo *framework.QueuedPodInfo) {
+		podInfo = f()
+		podInfo.Pod.Status.NominatedNodeName = "node-1"
+		return podInfo
+	}
+	go testCtx.Scheduler.Run(testCtx.Ctx)
+	return testCtx
+}
+
+// TestPreferNominatedNode test when the feature of "PreferNominatedNode" is enabled, the overall scheduling logic is not changed.
+// If the nominated node pass all the filters, then preemptor pod will run on the nominated node, otherwise, it will be scheduled
+// to another node in the cluster that ables to pass all the filters.
+// NOTE: This integration test is not intending to check the logic of preemption, but rather a sanity check when the feature is
+// enabled.
+func TestPreferNominatedNode(t *testing.T) {
+	defer featuregatetesting.SetFeatureGateDuringTest(t, utilfeature.DefaultFeatureGate, features.PreferNominatedNode, true)()
+	testCtx := initTestPreferNominatedNode(t, "perfer-nominated-node")
+	t.Cleanup(func() {
+		testutils.CleanupTest(t, testCtx)
+	})
+	cs := testCtx.ClientSet
+	defaultPodRes := &v1.ResourceRequirements{Requests: v1.ResourceList{
+		v1.ResourceCPU:    *resource.NewMilliQuantity(100, resource.DecimalSI),
+		v1.ResourceMemory: *resource.NewQuantity(100, resource.DecimalSI)},
+	}
+	defaultNodeRes := map[v1.ResourceName]string{
+		v1.ResourcePods:   "32",
+		v1.ResourceCPU:    "500m",
+		v1.ResourceMemory: "500",
+	}
+
+	type nodeConfig struct {
+		name string
+		res  map[v1.ResourceName]string
+	}
+
+	tests := []struct {
+		name         string
+		nodes        []*nodeConfig
+		existingPods []*v1.Pod
+		pod          *v1.Pod
+		runnningNode string
+	}{
+		{
+			name: "nominated node released all resource, preemptor is scheduled to the nominated node",
+			nodes: []*nodeConfig{
+				{name: "node-1", res: defaultNodeRes},
+				{name: "node-2", res: defaultNodeRes},
+			},
+			existingPods: []*v1.Pod{
+				initPausePod(&pausePodConfig{
+					Name:      "low-pod1",
+					Namespace: testCtx.NS.Name,
+					Priority:  &lowPriority,
+					NodeName:  "node-2",
+					Resources: defaultPodRes,
+				}),
+			},
+			pod: initPausePod(&pausePodConfig{
+				Name:      "preemptor-pod",
+				Namespace: testCtx.NS.Name,
+				Priority:  &highPriority,
+				Resources: &v1.ResourceRequirements{Requests: v1.ResourceList{
+					v1.ResourceCPU:    *resource.NewMilliQuantity(500, resource.DecimalSI),
+					v1.ResourceMemory: *resource.NewQuantity(200, resource.DecimalSI)},
+				},
+			}),
+			runnningNode: "node-1",
+		},
+		{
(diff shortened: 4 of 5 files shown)


chendave:
/cc @Huang-Wei @alculquicondor @ahg-g 

will try to build a benchmark for this later.

pkg/scheduler/apis/config/types.go
```diff
@@ -106,6 +106,13 @@ type KubeSchedulerConfiguration struct {
 	// Extenders are the list of scheduler extenders, each holding the values of how to communicate
 	// with the extender. These extenders are shared by all scheduler profiles.
 	Extenders []Extender
+
+	// PreemptorOnNominatedNode tells scheduler whether it is allowed to short-circuit to check the
+	// nominated node directly instead of loop over all the nodes in the cluster.
+	// Set this to true implies the preemptor pod might not be dispatched to the best candidate in
+	// some corner case, e.g. another node releases the resources and hence is the best candidate
+	// while the victims got removed from the nominated node.
+	PreemptorOnNominatedNode bool
```
alculquicondor, line 115: I still think this shouldn't be an option. But if people strongly disagree, let's put it in the preemption plugin args.

alculquicondor: Oops, sorry. I guess this is not related to the preemption plugin. I would call it `TryNominatedNodeFirst`

chendave: How about the non-optimal case as what @Huang-Wei 's concern if make it as default?

chendave:
@alculquicondor thanks for your review, let's see whether @Huang-Wei  and @ahg-g  has anything to say about this.

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -278,6 +279,32 @@ func (g *genericScheduler) findNodesThatFitPod(ctx context.Context, prof *profil
...
+			filteredNodesStatuses[nodeInfo.Node().Name] = status
+			// As only the nominated node is validated this time, it's possible there is node in the cluster is already
+			// available for this pod, return the error so that a fresh new scheduling cycle is expected later on.
+			err = fmt.Errorf("Nominated node is not a good candidate anymore, error: %v", err.Error())
+		}
+		// Cleanup the nominated node from the pod so that the node won't affect the future scheduling cycle.
+		if errs := util.ClearNominatedNodeName(prof.ClientSet(), pod); errs != nil {
+			klog.Errorf("Cannot clear 'NominatedNodeName' field of pod %v/%v: %v", pod.Namespace, pod.Name, errs)
+		}
+		return feasibleNodes, filteredNodesStatuses, err
```
soulxu, line 305: a question about do we want to bypass the extender which invoking at line 313?

also there will bypassing some logic inside 'findNodesThatPassFilters. another way, if we call the 'findNodesThatPassFilters' and add new parameter to it about the nodes we want to check, instead of the hardcode all nodes at line 322, then it won't be a problem.

pkg/scheduler/apis/config/types.go
chendave: partially done,  has been renamed to `TryNominatedNodeFirst`

pkg/scheduler/core/generic_scheduler.go
chendave: Done

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -278,7 +279,38 @@ func (g *genericScheduler) findNodesThatFitPod(ctx context.Context, prof *profil
 
 	}
 
-	feasibleNodes, err := g.findNodesThatPassFilters(ctx, prof, state, pod, filteredNodesStatuses)
+	var nodes []*framework.NodeInfo
+	// The "NominatedNodeName" field is set due to the preemption in the previous scheduling cycle, the node
+	// most likely is still the only candidate for the preemptor pod.
+	// Have a try on this node directly when the option tryNominatedNodeFirst is set.
+	tryNominatedNodeFirst := g.tryNominatedNodeFirst && len(pod.Status.NominatedNodeName) > 0
```
chendave, line 286: by default, this option is set to false but would love to hear from you about this option and the default value.

chendave:
https://github.com/kubernetes/kubernetes/blob/43fbe17dc682b703b0b663935a797244378f0aaf/test/integration/scheduler_perf/config/performance-config.yaml#L230-L235 might be the right benchmark needed here, but I found some issue there and currently is still investigating into it.

Huang-Wei:
@alculquicondor Suppose we go with this new option, should the SchedulerConfig version be bumped to v1beta2 in 1.20?

pkg/scheduler/apis/config/types.go
alculquicondor: I think the non-optimal case would be rare.

alculquicondor:
> Aldo Culquicondor Suppose we go with this new option, should the SchedulerConfig version be bumped to v1beta2 in 1.20?

No. New fields are not a problem.

chendave:
Have evaluated with `scheduler_perf`,  the result is updated in the cover message, and more testcase is provided.

@Huang-Wei  @alculquicondor  @ahg-g

pkg/scheduler/apis/config/types.go
chendave: I agree it's rare, maybe we can make it as an option first for a while and deprecate the option after the couple of cycle if there is no negative feedback from end-user?

staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
```diff
@@ -93,6 +93,13 @@ type KubeSchedulerConfiguration struct {
 	// with the extender. These extenders are shared by all scheduler profiles.
 	// +listType=set
 	Extenders []Extender `json:"extenders,omitempty"`
+
+	// TryNominatedNodeFirst tells scheduler whether it is allowed to short-circuit to check the
+	// nominated node directly instead of loop over all the nodes in the cluster.
+	// Set this to true implies the preemptor pod might not be dispatched to the best candidate in
+	// some corner case, e.g. another node releases the resources and hence is the best candidate
+	// while the victims got removed from the nominated node.
+	TryNominatedNodeFirst bool `json:"tryNominatedNodeFirst,omitempty"`
```
alculquicondor, line 102: Let's discuss the need of this option. Could you add an entry for the next SIG meeting?

pkg/scheduler/scheduler.go
```diff
@@ -347,11 +357,13 @@ func updatePod(client clientset.Interface, pod *v1.Pod, condition *v1.PodConditi
...
-	if !podutil.UpdatePodCondition(&podCopy.Status, condition) &&
-		(len(nominatedNode) == 0 || pod.Status.NominatedNodeName == nominatedNode) {
+	if !podutil.UpdatePodCondition(&podCopy.Status, condition) && (pod.Status.NominatedNodeName == nominatedNode) {
 		return nil
 	}
-	if nominatedNode != "" {
+	// Update NominatedNodeName when the nominatedNode has a different value, the case nominatedNode is
+	// empty while the pod has NominatedNodeName set means the NominatedNodeName no longer fit,
+	// make it empty will moves the pod to the active queue. It lets scheduler find another place for the pod.
+	if podCopy.Status.NominatedNodeName != nominatedNode {
```
soulxu, line 366: feel like this fixes the issue #85677 also

chendave: this change is necessary, actually, I found a race condition which is partially addressed by this change. I will give details about the race condition in my next day.

staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
chendave: I will check the schedule and see whether I can attend in person.

pkg/scheduler/scheduler.go
chendave: pls see: https://github.com/kubernetes/kubernetes/issues/93505 for the race condition.

staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
chendave: Added a topic in a agenda, it's a little late for me but I will try my best to attend.

ahg-g: I haven't looked at the details of the PR yet, but +1 to not having this as a config flag. We could use a feature flag to roll this out, but a config flag isn't useful.

pkg/scheduler/apis/config/types.go
chendave: Done

staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
chendave: Done

chendave:
Have been switched to `featuregate` instead of component config.

alculquicondor:
Recall that we also suggested a KEP

chendave:
> Recall that we also suggested a KEP

sure, will do

chendave:
> Recall that we also suggested a KEP

@alculquicondor , this will be hold for an while, the `enhancements` repo is currently under internal IP review process. :(

pkg/features/kube_features.go
```diff
@@ -435,6 +445,12 @@ const (
 	// Expects vSphere CSI Driver to be installed and configured on all nodes.
 	CSIMigrationvSphereComplete featuregate.Feature = "CSIMigrationvSphereComplete"
 
+	// owner: @huffmanca
+	// alpha: v1.19
+	//
+	// Determines if a CSI Driver supports applying fsGroup.
+	CSIVolumeFSGroupPolicy featuregate.Feature = "CSIVolumeFSGroupPolicy"
```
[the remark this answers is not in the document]
chendave: Done

chendave:
/hold

Just a rebase.

alculquicondor:
Note for reviewers: Current plan is to add this feature in 1.21, as per KEP https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/1923-prefer-nominated-node

chendave:
> Note for reviewers: Current plan is to add this feature in 1.21, as per KEP

I will keep this hold until 1.21 is open.

@alculquicondor  @Huang-Wei  @ahg-g , this is ready for review, I'd highly appreciate if you could take a look before the open of 1.21.

alculquicondor:
Please remind me after the freeze. I have low bandwidth at the moment.

Huang-Wei:
> @alculquicondor @Huang-Wei @ahg-g , this is ready for review, I'd highly appreciate if you could take a look before the open of 1.21.

Acked. But I may choose to review it when 1.21 is open.

chendave:
/hold cancel

rebased.

chendave:
@alculquicondor @Huang-Wei  Do you have the bandwidth on this one?

alculquicondor:
Please make sure that the KEP is in implementable state first

chendave:
> Please make sure that the KEP is in implementable state first

Thanks, just update the doc is enough? or something else should I do?

alculquicondor:
Production Readiness Review is now required for 1.21. So you have to add a reviewer for that. But the bar for alpha is not too high.

chendave:
> Production Readiness Review is now required for 1.21. So you have to add a reviewer for that. But the bar for alpha is not too high.

updated.

chendave:
rebased

chendave:
Really sorry for chasing, I saw the Enhancements Freeze is only three weeks away,  https://github.com/kubernetes/sig-release/tree/master/releases/release-1.21

Begin Enhancements Freeze (EOD PST)	Enhancements Lead	Tue February 9th	week 5	master-blocking, master-informing

@ahg-g  @Huang-Wei  @alculquicondor @damemi

alculquicondor:
Will take a look soon, but enhancements freeze is not the same as code freeze, so you are fine.

pkg/features/kube_features.go
```diff
@@ -269,6 +269,16 @@ const (
 	// Enables generic ephemeral inline volume support for pods
 	GenericEphemeralVolume featuregate.Feature = "GenericEphemeralVolume"
 
+	// owner: @chendave
+	// alpha: v1.21
+	//
+	// PreferNominatedNode tells scheduler whether the nominated node will be checked first before looping
+	// all the rest of nodes in the cluster.
+	// Enables this feature also implies the preemptor pod might not be dispatched to the best candidate in
```
alculquicondor, line 277: Enabling

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -203,51 +205,94 @@ func (g *genericScheduler) findNodesThatFitPod(ctx context.Context, fwk framewor
...
+		if len(nominatedNode) != 0 {
+			return nominatedNode, filteredNodesStatuses, nil
+		}
+		// Remove the nominated node from node list and continue to evaluate the rest of the nodes in the cluster.
+		for i, nodeInfo := range allNodes {
+			if nodeInfo.Node().Name == nnn {
+				allNodes = append(allNodes[:i], allNodes[i+1:]...)
+			}
+		}
+		feasibleNodes, err = g.findNodesThatPassFilters(ctx, fwk, state, pod, filteredNodesStatuses, allNodes)
```
alculquicondor, line 257: This code just repeats in the `else` block. Why not just put it outside of the conditional?

chendave: yes, I also don't understand why I repeated the code here.  ;)

chendave: Done

pkg/features/kube_features.go
chendave: Done

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -203,27 +205,60 @@ func (g *genericScheduler) findNodesThatFitPod(ctx context.Context, fwk framewor
...
+		nodeInfo, err := g.nodeInfoSnapshot.Get(nnn)
+		if err != nil {
+			return nil, nil, err
+		}
+		node := []*framework.NodeInfo{nodeInfo}
+		nominatedNode, err := g.findNodesThatPassFilters(ctx, fwk, state, pod, filteredNodesStatuses, node)
+		if err != nil {
+			return nil, nil, err
+		}
+		if len(nominatedNode) != 0 {
```
alculquicondor, line 241: doesn't findNodesThatPassExtenders handle this? we don't have the same check when we filter the rest of the pods.

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -203,27 +205,60 @@ func (g *genericScheduler) findNodesThatFitPod(ctx context.Context, fwk framewor
...
+	// Have a try on this node directly when the field is set and the feature is enabled.
+	preferNominatedNode := len(pod.Status.NominatedNodeName) > 0 && utilfeature.DefaultFeatureGate.Enabled(features.PreferNominatedNode)
+	if preferNominatedNode {
+		nnn := pod.Status.NominatedNodeName
+		nodeInfo, err := g.nodeInfoSnapshot.Get(nnn)
+		if err != nil {
+			return nil, nil, err
+		}
+		node := []*framework.NodeInfo{nodeInfo}
+		nominatedNode, err := g.findNodesThatPassFilters(ctx, fwk, state, pod, filteredNodesStatuses, node)
```
alculquicondor, line 237: call it feasibleNodes for consistency

test/integration/scheduler/preemption_test.go
```diff
@@ -1314,3 +1314,155 @@ func TestPDBInPreemption(t *testing.T) {
...
+			if err != nil {
+				t.Errorf("Cannot schedule Pod %v/%v, error: %v", test.pod.Namespace, test.pod.Name, err)
+			}
+			// Make sure the pod has been scheduled to the right node.
+			if preemptor.Spec.NodeName != test.runnningNode {
+				t.Errorf("Expect pod running on %v, got %v.", test.runnningNode, preemptor.Spec.NodeName)
+			}
+			//cleanup
+			pods = append(pods, preemptor)
+			testutils.CleanupPods(cs, t, pods)
```
alculquicondor, line 1464: call this from a defer

test/integration/scheduler/preemption_test.go
```diff
@@ -1314,3 +1314,155 @@ func TestPDBInPreemption(t *testing.T) {
...
+				initPausePod(&pausePodConfig{
+					Name:      "low-pod1",
+					Namespace: testCtx.NS.Name,
+					Priority:  &lowPriority,
+					NodeName:  "node-2",
+					Resources: defaultPodRes,
+				}),
+			},
+			pod: initPausePod(&pausePodConfig{
+				Name:      "preemptor-pod",
```
alculquicondor, line 1385: The node1 is empty, why would this trigger preemption?

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -203,27 +205,60 @@ func (g *genericScheduler) findNodesThatFitPod(ctx context.Context, fwk framewor
 
 	// Run "prefilter" plugins.
 	s := fwk.RunPreFilterPlugins(ctx, state, pod)
+	allNodes, err := g.nodeInfoSnapshot.NodeInfos().List()
```
alculquicondor, line 208: Is there any unit test we can add for the changes in this file?

test/integration/scheduler/preemption_test.go
chendave: I think we needn't test the logic of preemption, as it is already covered in the "TestPreemption", `initPreemptorTest` and `wrapNext` will set the `.Pod.Status.NominatedNodeName` for the pod so that this testcase will focus on the code change in this PR instead.

I have added some comments on this this to explain what this testcase want to cover.

chendave: node1 is empty and `.Pod.Status.NominatedNodeName` is set to "node1", this is simulating that the resource on node1 has been released, so it's good to be assigned for the pod.

pkg/scheduler/core/generic_scheduler.go
chendave: Done

test/integration/scheduler/preemption_test.go
```diff
@@ -1314,3 +1314,155 @@ func TestPDBInPreemption(t *testing.T) {
...
+	testCtx.Scheduler.NextPod = wrapNext(testCtx.Scheduler.NextPod)
+	go testCtx.Scheduler.Run(testCtx.Ctx)
+	return testCtx
+}
+
+// TestPreferNominatedNode test when the feature of "PreferNominatedNode" is enabled, the overall scheduling logic is not changed.
+// If the nominated node pass all the filters, then preemptor pod will run on the nominated node, otherwise, it will be scheduled
+// to another node in the cluster that ables to pass all the filters.
+// NOTE: This integration test is not intending to check the logic of preemption, but rather a sanity check when the feature is
+// enabled.
```
chendave, line 1339: please checkout the comments here.

test/integration/scheduler/preemption_test.go
chendave: Done

pkg/scheduler/core/generic_scheduler.go
chendave: Sorry,  I don't quite understand this, do you mean "the rest of the nodes"? Shouldn't the same check has been done at line 262?

pkg/scheduler/core/generic_scheduler.go
chendave: The overall logic has been verified by testcase `TestPreferNominatedNode` created in "test/integration/scheduler/preemption_test.go". 
- If the nominated passed all the filters,  the nominated node will be return, otherwise,
- other nodes will be evaluated and return a suitable candidate.

"TestPreferNominatedNode" is truly has not verify the logic of preemption, if you think it's better to test in "generic_scheduler_test.go" or there is any specific change haven't covered by any testcase, please just let me know, I will take care it.

chendave: hmm, that's integration test, as to the unit test for the change like this, output should be the same before and after the change, it is hard to see which aspect should to be test in the unit test.

chendave:
@alculquicondor thanks for the review.

test/integration/scheduler/preemption_test.go
alculquicondor: Oh, I missed the `wrapNext`. Can you put explain the test is simulating the nominatedNode in the comment for TestPreferNominatedNode

test/integration/scheduler/preemption_test.go
```diff
@@ -1314,3 +1314,155 @@ func TestPDBInPreemption(t *testing.T) {
 		})
 	}
 }
+
+// wrapNext wraps the NextPod() method to make it appear the preemption has been done already and the nominated node has been set.
+func wrapNext(f func() *framework.QueuedPodInfo) func() *framework.QueuedPodInfo {
```
alculquicondor, line 1319: `nextPodWithNominatedNodeName`

test/integration/scheduler/preemption_test.go
```diff
@@ -1314,3 +1314,155 @@ func TestPDBInPreemption(t *testing.T) {
...
+	return func() (podInfo *framework.QueuedPodInfo) {
+		podInfo = f()
+		podInfo.Pod.Status.NominatedNodeName = "node-1"
+		return podInfo
+	}
+}
+
+func initPreemptorTest(t *testing.T, nsPrefix string, opts ...scheduler.Option) *testutils.TestContext {
```
alculquicondor, line 1327: Choose a better name. This is not about preemption.

Maybe you can also do:

`test.Scheduler.NextPod = func() {`

So that we don't leave standalone functions that other tests don't need.

pkg/scheduler/core/generic_scheduler.go
alculquicondor: I like the filter call counts :)

pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -1198,3 +1202,91 @@ func TestFairEvaluationForNodes(t *testing.T) {
 		}
 	}
 }
+
+func TestPreferNominatedNodeFilterCallCounts(t *testing.T) {
+	tests := []struct {
+		name                  string
+		feature               featuregate.Feature
```
alculquicondor, line 1209: prefer a boolean and always set the feature gate value.

If a test is meant to have different combinations of feature gates, this makes sense. But this test is specific to this gate.

pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -1198,3 +1202,91 @@ func TestFairEvaluationForNodes(t *testing.T) {
...
+		pod                   *v1.Pod
+		nodeReturnCodeMap     map[string]framework.Code
+		expectedCount         int32
+		expectedPatchRequests int
+	}{
+		{
+			name:              "Enable the feature, pod has the nominated node set, filter is called only once",
+			feature:           features.PreferNominatedNode,
+			pod:               st.MakePod().Name("p_with_nominated_node").UID("p").Priority(highPriority).NominatedNodeName("node1").Obj(),
+			nodeReturnCodeMap: map[string]framework.Code{},
```
alculquicondor, line 1219: can we leave it `nil` as the third test?

test/integration/scheduler/preemption_test.go
chendave: Done

test/integration/scheduler/preemption_test.go
chendave: removed and change to use lambda

pkg/scheduler/core/generic_scheduler_test.go
chendave: Good catch! removed.

pkg/scheduler/core/generic_scheduler_test.go
chendave: Great suggestion, done!

chendave:
@alculquicondor thanks for the review! please let me know if there is anything more I need to address.

chendave:
unrelated

/retest

pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -33,9 +33,12 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
```
alculquicondor, line 36: nit: no need for alias

pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -1198,3 +1201,90 @@ func TestFairEvaluationForNodes(t *testing.T) {
...
+			pod:               st.MakePod().Name("p_with_nominated_node").UID("p").Priority(highPriority).NominatedNodeName("node1").Obj(),
+			nodeReturnCodeMap: map[string]framework.Code{"node1": framework.Unschedulable},
+			expectedCount:     3,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			if test.feature != false {
+				defer featuregatetesting.SetFeatureGateDuringTest(t, utilfeature.DefaultFeatureGate, features.PreferNominatedNode, true)()
```
alculquicondor, line 1244: always set the value.

Otherwise, once we enable the feature by default, the test needs to be updated.

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -203,27 +205,58 @@ func (g *genericScheduler) findNodesThatFitPod(ctx context.Context, fwk framewor
...
+			feasibleNodes, err = g.findNodesThatPassExtenders(pod, feasibleNodes, filteredNodesStatuses)
+		}
+		if err != nil {
+			return nil, nil, err
+		}
+		// Nominated node passes all the filters, scheduler is good to assign this node to the pod.
+		if len(feasibleNodes) != 0 {
+			return feasibleNodes, filteredNodesStatuses, nil
+		}
+		// Remove the nominated node from node list and continue to evaluate the rest of the nodes in the cluster.
```
ahg-g, line 249: That is not necessary, in fact it is more expensive to iterate over all nodes to remove the nominated node vs just evaluate the node again.

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -203,27 +205,58 @@ func (g *genericScheduler) findNodesThatFitPod(ctx context.Context, fwk framewor
...
 			filteredNodesStatuses[n.Node().Name] = s
 		}
 		return nil, filteredNodesStatuses, nil
 	}
 
-	feasibleNodes, err := g.findNodesThatPassFilters(ctx, fwk, state, pod, filteredNodesStatuses)
+	// The "NominatedNodeName" field is set due to the preemption in the previous scheduling cycle, the node
+	// most likely is still the only candidate for the preemptor pod.
+	// Have a try on this node directly when the field is set and the feature is enabled.
+	preferNominatedNode := len(pod.Status.NominatedNodeName) > 0 && utilfeature.DefaultFeatureGate.Enabled(features.PreferNominatedNode)
```
ahg-g, line 227: move this to a separate function, e.g., `evaluateNominatedNode`

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -203,27 +205,58 @@ func (g *genericScheduler) findNodesThatFitPod(ctx context.Context, fwk framewor
...
+	if preferNominatedNode {
+		nnn := pod.Status.NominatedNodeName
+		nodeInfo, err := g.nodeInfoSnapshot.Get(nnn)
+		if err != nil {
```
ahg-g, line 231: Lets not return, just fall back to the old logic. This will be easier to do once this logic is encapsulated in a separate function.

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -203,27 +205,58 @@ func (g *genericScheduler) findNodesThatFitPod(ctx context.Context, fwk framewor
...
 		for _, n := range allNodes {
 			filteredNodesStatuses[n.Node().Name] = s
 		}
 		return nil, filteredNodesStatuses, nil
 	}
 
-	feasibleNodes, err := g.findNodesThatPassFilters(ctx, fwk, state, pod, filteredNodesStatuses)
+	// The "NominatedNodeName" field is set due to the preemption in the previous scheduling cycle, the node
+	// most likely is still the only candidate for the preemptor pod.
+	// Have a try on this node directly when the field is set and the feature is enabled.
```
ahg-g, line 226:
```suggestion
	// "NominatedNodeName" can potentially be set in a previous scheduling cycle as a result of preemption.
	// This node is likely the only candidate the will fit the pod, and hence we try it first before iterating over 
	// all nodes.
```

pkg/scheduler/core/generic_scheduler_test.go
chendave: Done

pkg/scheduler/core/generic_scheduler_test.go
chendave: I have a test where the feature should be disabled in order to compare the behavior before and after the change, is there decent way to do that?

pkg/scheduler/core/generic_scheduler.go
chendave: Done

pkg/scheduler/core/generic_scheduler.go
chendave: Done

pkg/scheduler/core/generic_scheduler.go
chendave: Done

chendave:
@ahg-g @alculquicondor  thanks for the review and comments!

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -196,34 +198,65 @@ func (g *genericScheduler) numFeasibleNodesToFind(numAllNodes int32) (numNodes i
...
-	feasibleNodes, err := g.findNodesThatPassFilters(ctx, fwk, state, pod, filteredNodesStatuses)
+	// "NominatedNodeName" can potentially be set in a previous scheduling cycle as a result of preemption.
+	// This node is likely the only candidate that will fit the pod, and hence we try it first before iterating over all nodes.
+	if len(pod.Status.NominatedNodeName) > 0 && feature.DefaultFeatureGate.Enabled(features.PreferNominatedNode) {
+		feasibleNodes, err := g.evaluateNominatedNode(ctx, pod, fwk, state, filteredNodesStatuses)
+		if err != nil {
+			klog.InfoS("Evaluation failed on nominated node", "err", err)
+		}
+		// Nominated node passes all the filters, scheduler is good to assign this node to the pod.
+		if feasibleNodes != nil && len(feasibleNodes) != 0 {
```
alculquicondor, line 252: `len(feasibleNodes) != 0` is enough. `len` can deal with `nil`

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -196,34 +198,65 @@ func (g *genericScheduler) numFeasibleNodesToFind(numAllNodes int32) (numNodes i
...
 		return nil, filteredNodesStatuses, nil
 	}
 
-	feasibleNodes, err := g.findNodesThatPassFilters(ctx, fwk, state, pod, filteredNodesStatuses)
+	// "NominatedNodeName" can potentially be set in a previous scheduling cycle as a result of preemption.
+	// This node is likely the only candidate that will fit the pod, and hence we try it first before iterating over all nodes.
+	if len(pod.Status.NominatedNodeName) > 0 && feature.DefaultFeatureGate.Enabled(features.PreferNominatedNode) {
+		feasibleNodes, err := g.evaluateNominatedNode(ctx, pod, fwk, state, filteredNodesStatuses)
+		if err != nil {
+			klog.InfoS("Evaluation failed on nominated node", "err", err)
```
alculquicondor, line 249: ErrorS, as errors here are unexpected.

For this reason, I actually don't agree that we should continue testing the rest of the nodes in this case. In any case, we will error again when all nodes are tested.

pkg/scheduler/core/generic_scheduler_test.go
alculquicondor: remove the conditional and do:

```golang
defer featuregatetesting.SetFeatureGateDuringTest(t, utilfeature.DefaultFeatureGate, features.PreferNominatedNode, test.feature)()
```

pkg/scheduler/core/generic_scheduler.go
ahg-g: I treat this as a short cut, if it doesn't work, I would rather always go the full route. The error we may face in the short cut may not always manifest in the full route.

ahg-g:
/approve
/hold

approving the feature gate and the core logic, will leave the lgtm to Aldo.

pkg/scheduler/core/generic_scheduler.go
ahg-g: +1 to using ErrorS

chendave: Done

pkg/scheduler/core/generic_scheduler.go
chendave: Done

pkg/scheduler/core/generic_scheduler_test.go
chendave: Done

chendave:
commit squashed, comments addressed! 
@alculquicondor PTAL, thanks!

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -196,34 +198,65 @@ func (g *genericScheduler) numFeasibleNodesToFind(numAllNodes int32) (numNodes i
...
 		return nil, filteredNodesStatuses, nil
 	}
 
-	feasibleNodes, err := g.findNodesThatPassFilters(ctx, fwk, state, pod, filteredNodesStatuses)
+	// "NominatedNodeName" can potentially be set in a previous scheduling cycle as a result of preemption.
+	// This node is likely the only candidate that will fit the pod, and hence we try it first before iterating over all nodes.
+	if len(pod.Status.NominatedNodeName) > 0 && feature.DefaultFeatureGate.Enabled(features.PreferNominatedNode) {
+		feasibleNodes, err := g.evaluateNominatedNode(ctx, pod, fwk, state, filteredNodesStatuses)
+		if err != nil {
+			klog.ErrorS(err, "Evaluation failed on nominated node")
```
alculquicondor, line 249: Let's add the pod and node

chendave: Done

alculquicondor (commented):
/lgtm
/hold cancel

chendave:
per the requirement, link this to the enhancement issue: https://github.com/kubernetes/enhancements/issues/1923

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -196,34 +198,65 @@ func (g *genericScheduler) numFeasibleNodesToFind(numAllNodes int32) (numNodes i
...
-	feasibleNodes, err := g.findNodesThatPassFilters(ctx, fwk, state, pod, filteredNodesStatuses)
+	// "NominatedNodeName" can potentially be set in a previous scheduling cycle as a result of preemption.
+	// This node is likely the only candidate that will fit the pod, and hence we try it first before iterating over all nodes.
+	if len(pod.Status.NominatedNodeName) > 0 && feature.DefaultFeatureGate.Enabled(features.PreferNominatedNode) {
+		feasibleNodes, err := g.evaluateNominatedNode(ctx, pod, fwk, state, filteredNodesStatuses)
+		if err != nil {
+			klog.ErrorS(err, "Evaluation failed on nominated node", "pod", klog.KObj(pod), "node", pod.Status.NominatedNodeName)
+		}
+		// Nominated node passes all the filters, scheduler is good to assign this node to the pod.
+		if len(feasibleNodes) != 0 {
```
edwardstudy, line 252: @chendave Just curious.  Could we remove nominatedNode from `allNodes`  quickly if this result is empty. After that, this node should not go Filters/Extenders twice.

chendave: the original version was removing this node from the node list to avoid the evaluation again later, but this require some slice manipulation, which might be cost more than just evaluate it again.

chendave: this is true today, as we don't have too much default filtering list, and cost of the filtering is not dramatic.

PR #93180, opened 2020-07-17
add  unit test for GetAddressAndDialer

### LINKED ISSUE #92996: Add unit test for `pkg/kubelet/util/util_unix.go#GetAddressAndDialer`
**What would you like to be added:**
Currently, pkg/kubelet/util/util_unix.go#GetAddressAndDialer does not have unit test coverage. Unless there's a strong reason not to, we'd like public facing functions to have unit test coverage.

**Why is this needed:**
There's a fairly small chance of GetAddressAndDialer being modified, and thus breaking, in the future without this added test coverage. However, it should be relatively straightforward to add the test coverage, and could make a good first issue for a new contributor. There's also the added benefit that it helps keep the test coverage in pkg/kubelet high :) And we never know, it might help catch a bug some day :)

**Extra note**
I think this ticket could be a great candidate for someone who is new to
contributing to sig-node/k8s. *I'd prefer to please give first dibs for picking
this up to any folks who are currently underrepresented in the k8s community.* Please reach out to @mattjmcnaughton on k8s Slack (or post directly in #sig-node and tag me) if you're interested, and I'll assign you the issue. I'm happy to mentor and be helpful in whatever way I can :)

FreeZhang61:
**What type of PR is this?**

/kind feature

```
    to ensure the code be strong,  add unit test for GetAddressAndDialer
```

Fixes [#92996](https://github.com/kubernetes/kubernetes/issues/92996)

### DIFF
--- a/pkg/kubelet/util/util_unix_test.go
+++ b/pkg/kubelet/util/util_unix_test.go
@@ -151,12 +151,20 @@ func TestGetAddressAndDialer(t *testing.T) {
 			expectedAddr: "/tmp/f6.sock",
 		},
 		{
-			endpoint:    "tco:/localhost:9090",
+			endpoint:    "tcp://localhost:9090",
 			expectError: true,
 		},
 		{
 			// The misspelling is intentional to make it error
 			endpoint:    "htta://free-test.com",
+			expectError: true,
+		},
+		{
+			endpoint:    "https://www.youtube.com/",
+			expectError: true,
+		},
+		{
+			endpoint:    "http://www.baidu.com/",
 			expectError: true,
 		},
 	}
--- a/pkg/kubelet/util/util_unix_test.go
+++ b/pkg/kubelet/util/util_unix_test.go
@@ -133,3 +133,41 @@ func TestIsUnixDomainSocket(t *testing.T) {
 		assert.Equal(t, result, test.expectSocket, "Unexpected result from IsUnixDomainSocket: %v for %s", result, test.label)
 	}
 }
+
+func TestGetAddressAndDialer(t *testing.T) {
+	tests := []struct {
+		endpoint     string
+		expectError  bool
+		expectedAddr string
+	}{
+		{
+			endpoint:     "unix:///tmp/s1.sock",
+			expectError:  false,
+			expectedAddr: "/tmp/s1.sock",
+		},
+		{
+			endpoint:     "unix:///tmp/f6.sock",
+			expectError:  false,
+			expectedAddr: "/tmp/f6.sock",
+		},
+		{
+			endpoint:    "tco:/localhost:9090",
+			expectError: true,
+		},
+		{
+			// The misspelling is intentional to make it error
+			endpoint:    "htta://free-test.com",
+			expectError: true,
+		},
+	}
+	for _, test := range tests {
+		// just test addr and err
+		addr, _, err := GetAddressAndDialer(test.endpoint)
+		if test.expectError {
+			assert.NotNil(t, err, "expected error during parsing %s", test.endpoint)
+			continue
+		}
+		assert.Nil(t, err, "expected no error during parsing %s", test.endpoint)
+		assert.Equal(t, test.expectedAddr, addr)
+	}
+}


FreeZhang61:
> **What type of PR is this?**
> 
> /kind flake
> 
> **What this PR does / why we need it**:
> 
> **Which issue(s) this PR fixes**:

pkg/kubelet/util/util_unix_test.go
```diff
@@ -133,3 +133,41 @@ func TestIsUnixDomainSocket(t *testing.T) {
...
+			endpoint:     "unix:///tmp/f6.sock",
+			expectError:  false,
+			expectedAddr: "/tmp/f6.sock",
+		},
+		{
+			endpoint:    "tco:/localhost:9090",
+			expectError: true,
+		},
+		{
+			endpoint:    "htta://free-test.com",
```
kikisdeliveryservice, line 158: should this be "http://..."?

pkg/kubelet/util/util_unix_test.go
```diff
@@ -133,3 +133,41 @@ func TestIsUnixDomainSocket(t *testing.T) {
...
+			expectError: true,
+		},
+	}
+	for _, test := range tests {
+		addr, f, err := GetAddressAndDialer(test.endpoint)
```
kikisdeliveryservice, line 163: maybe sub f for dialer for clarity?

pkg/kubelet/util/util_unix_test.go
```diff
@@ -133,3 +133,41 @@ func TestIsUnixDomainSocket(t *testing.T) {
...
+
+		if test.expectError {
+			assert.NotNil(t, err, "Expect error during parsing %s", test.endpoint)
+			continue
+		}
+		assert.NotNil(t, f, "Expect not nil func")
```
kikisdeliveryservice, line 169: Expect -> Expected

pkg/kubelet/util/util_unix_test.go
```diff
@@ -133,3 +133,41 @@ func TestIsUnixDomainSocket(t *testing.T) {
...
+		assert.Nil(t, err, "Expect no error during parsing %s", test.endpoint)
```
kikisdeliveryservice, line 170: maybe: "Unexpected error %v during parsing %s", err, test.endpoint

pkg/kubelet/util/util_unix_test.go
```diff
@@ -133,3 +133,41 @@ func TestIsUnixDomainSocket(t *testing.T) {
...
+		{
+			endpoint:    "htta://free-test.com",
+			expectError: true,
+		},
+	}
+	for _, test := range tests {
+		addr, f, err := GetAddressAndDialer(test.endpoint)
+
+		if test.expectError {
+			assert.NotNil(t, err, "Expect error during parsing %s", test.endpoint)
```
kikisdeliveryservice, line 166: Expect -> Expected

pkg/kubelet/util/util_unix_test.go
```diff
@@ -133,3 +133,41 @@ func TestIsUnixDomainSocket(t *testing.T) {
 		assert.Equal(t, result, test.expectSocket, "Unexpected result from IsUnixDomainSocket: %v for %s", result, test.label)
 	}
 }
+
+func TestGetAddressAndDialer(t *testing.T) {
+	tests := []struct {
+		endpoint     string
```
kikisdeliveryservice, line 139: Wondering if you need a case that covers using the fallback protocol
see: https://github.com/kubernetes/kubernetes/blob/0a7050f531f8fa4de19e8aa4feb274aee4d8340c/pkg/kubelet/util/util_unix.go#L83

kikisdeliveryservice (commented):
Adding the test is awesome! Just some small comments 😄

pkg/kubelet/util/util_unix_test.go
mattjmcnaughton: I think the misspelling is actually intentional, so that it fails :) But perhaps a comment to that effect both for this test and the one above could be helpful :)

pkg/kubelet/util/util_unix_test.go
mattjmcnaughton: Nit: I think that the "best practice" is for the first letter to be undercase so `Expected -> expected`. We'll want to make this change for all of the `assert` statements.

mattjmcnaughton (commented):
Great progress and great review from @kikisdeliveryservice :) Just a couple of add ons from me.

pkg/kubelet/util/util_unix_test.go
FreeZhang61: > I think the misspelling is actually intentional, so that it fails :) But perhaps a comment to that effect both for this test and the one above could be helpful :)

yeah, it's  intentional, I will add useful comment for it. 😀

pkg/kubelet/util/util_unix_test.go
FreeZhang61: > Wondering if you need a case that covers using the fallback protocol
> see:
> 
> https://github.com/kubernetes/kubernetes/blob/0a7050f531f8fa4de19e8aa4feb274aee4d8340c/pkg/kubelet/util/util_unix.go#L83

This testing  needn't  use  fallback protocol,  because the GetAddressAndDialer func doesn't return protocol  😃

pkg/kubelet/util/util_unix_test.go
FreeZhang61: > undercase

I've changed it

FreeZhang61: > Nit: I think that the "best practice" is for the first letter to be undercase so `Expected -> expected`. We'll want to make this change for all of the `assert` statements.

I've changed it

nikhita:
/assign @sjenning 
for approval

pkg/kubelet/util/util_unix_test.go
```diff
@@ -133,3 +133,41 @@ func TestIsUnixDomainSocket(t *testing.T) {
...
+			expectError:  false,
+			expectedAddr: "/tmp/s1.sock",
+		},
+		{
+			endpoint:     "unix:///tmp/f6.sock",
+			expectError:  false,
+			expectedAddr: "/tmp/f6.sock",
+		},
+		{
+			endpoint:    "tco:/localhost:9090",
```
sjenning, line 154: cloud you add a comment here explaining why this case is expected to fail?

pkg/kubelet/util/util_unix_test.go
```diff
@@ -133,3 +133,41 @@ func TestIsUnixDomainSocket(t *testing.T) {
...
+			expectError: true,
+		},
+		{
+			// The misspelling is intentional to make it error
+			endpoint:    "htta://free-test.com",
+			expectError: true,
+		},
```
sjenning, line 161: could you add a test case for a valid but not unix domain address so we exercise that error case?

FreeZhang61: > could you add a test case for a valid but non-unix domain address so we exercise that error case?
> https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/util/util_unix.go#L87-L89

Of course, I will add it immediately😊

sjenning:
thanks!
/lgtm
/approve

BenTheElder:
/uncc @kubernetes/dep-approvers
(the robots do not respond to edits)
(I'm pretty sure uncc doesn't work on teams anyhow though)

PR #92001, opened 2020-06-10
Include CSIDriver capability to indicate fsGroup support 

huffmanca:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
This PR is the followup to https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1682-csi-driver-skip-permission , and introduces the `SupportsFsGroup` field for CSIDrivers. This allows CSIDrivers to declare if they wish to allow modifying the fsGroup when volumes are mounted.

**Which issue(s) this PR fixes**:
Fixes https://github.com/kubernetes/enhancements/issues/1682

**Special notes for your reviewer**:
I've currently added the field in the v1 and v1beta1 staging APIs; however, this should be an alpha field on the CSIDriver.Spec object. I noticed another field (InlineVolumeSpec) that includes a comment mentioning it's alpha-level.

What's the correct way to include this?

**Does this PR introduce a user-facing change?**:
```release-note
A new alpha-level field, `SupportsFsGroup`, has been introduced for CSIDrivers to allow them to specify whether they support volume ownership and permission modifications. The `CSIVolumeSupportFSGroup` feature gate must be enabled to allow this field to be used. 
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:
```docs
- [KEP]: https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1682-csi-driver-skip-permission
- TODO: Include docs PR
```

PR #93195, opened 2020-07-17
Addresses nitpicks for FSGroupPolicy

huffmanca:
**What type of PR is this?**
/kind cleanup

This is a follow up to https://github.com/kubernetes/kubernetes/pull/92001 , and addresses the feedback items from that PR.

```release-note
NONE
```

```docs

```


huffmanca:
@msau42 

Tagging you here, as this addresses your feedback on https://github.com/kubernetes/kubernetes/pull/92001 .

huffmanca:
@smarterclayton ,

This PR slightly rewords the descriptions from https://github.com/kubernetes/kubernetes/pull/92001 . Would it be possible to review these API changes, since you were the reviewer of 92001?

pkg/apis/storage/types.go
```diff
@@ -282,6 +282,8 @@ type CSIDriverSpec struct {
 	// Refer to the specific FSGroupPolicy values for additional details.
 	// This field is alpha-level, and is only honored by servers
 	// that enable the CSIVolumeFSGroupPolicy feature gate.
+	// Defaults to ReadWriteOnceWithFSType, which will use heuristics to
+	// determine if Kubernetes should modify ownership and permissions of the volume.
```
smarterclayton, line 286: I'm trying to read the definition of this and it's not really exactly clear what it does.  For instance, the name includes `Once`, but it's not clear from the description below what decides "once", and I think it needs to.

huffmanca: I've updated the definition in this section to be more explicit and have it match the description of the type below. This will attempt to modify volume permissions if both a `fstype` is defined and the access modes contain `ReadWriteOnce`.

huffmanca: @smarterclayton ,

Is it possible to take a look at the latest definition to determine if this is clearer? I believe this is ready for another review.

pkg/apis/storage/types.go
```diff
@@ -280,8 +280,12 @@ type CSIDriverSpec struct {
 	// Defines if the underlying volume supports changing ownership and
 	// permission of the volume before being mounted.
 	// Refer to the specific FSGroupPolicy values for additional details.
-	// This field is alpha-level, and is only honored by servers
+	// This field is beta, and is only honored by servers
 	// that enable the CSIVolumeFSGroupPolicy feature gate.
+	// Defaults to ReadWriteOnceWithFSType, which will examine each volume
+	// to determine if Kubernetes should modify ownership and permissions of the volume.
+	// If a fstype is defined and the volume's access mode
+	// contains ReadWriteOnce, then the defined fsGroup will be applied.
```
msau42, line 288: I initially interpreted this to mean that fsgrouppolicy only applies if it's rwo. Can we make it more clear that this criteria only applies to the RWOWithFstype value?

huffmanca: This is currently in the description of the `ReadWriteOnceWithFSTypeFSGroupPolicy`. I've reworded it here to make it more explicit that this only applies to the default policy.

huffmanca:
@liggitt ,

I've pinged @smarterclayton on this one a couple of times, but he's mentioned that it may be best to get a separate API reviewer due to his current workload. This PR has been lingering for some time; would it be possible to review it?

liggitt:
done

/lgtm
/approve
/priority backlog

PR #93196, opened 2020-07-17
Andi li99/preprovisoned snapshot test

### LINKED ISSUE #324: Add E2E tests for pre-provisioned snapshots (both Delete and Retain policies)

AndiLi99:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes minor issues with the snapshotting test suite and adds preprovisioning tests.
Also combines two test cases with overlap to speed up the test suite.

Fixes https://github.com/kubernetes-csi/external-snapshotter/issues/324

```release-note
NONE
```

```docs

```


msau42:
/ok-to-test
/assign @saikat-royc 
cc @mattcary

test/e2e/storage/testsuites/snapshottable.go
```diff
@@ -401,12 +402,70 @@ func CreateSnapshotResource(sDriver SnapshottableTestDriver, config *PerTestConf
...
+
+		ginkgo.By("checking the Snapshot has been deleted")
+		err = utils.WaitForNamespacedGVRDeletion(dc, SnapshotGVR, r.Vs.GetName(), r.Vs.GetNamespace(), framework.Poll, framework.SnapshotDeleteTimeout)
+		framework.ExpectNoError(err)
+
+		err = dc.Resource(SnapshotContentGVR).Delete(context.TODO(), r.Vscontent.GetName(), metav1.DeleteOptions{})
+		framework.ExpectNoError(err)
+
+		ginkgo.By("checking the Snapshot content has been deleted")
+		err = utils.WaitForGVRDeletion(dc, SnapshotContentGVR, r.Vscontent.GetName(), framework.Poll, framework.SnapshotDeleteTimeout)
```
saikat-royc, line 446: We update the vs content object delete policy to "retain" in line 424. Should delete of vs content fail here? i.e framework.ExpectError(err) ? Can you please elaborate?

AndiLi99: The deletion policy can be confusing, I was confused as well. If the deletion policy is retain on vscontent, when vs is deleted vscontent will not be deleted. Further, if the vscontent is manually deleted then the underlying snapshot resource will not be deleted.

AndiLi99: The policy will not prevent vscontent from being deleted.

saikat-royc: I see, can you note this behavior in a comment in the code

test/e2e/storage/testsuites/base.go
```diff
@@ -585,6 +588,64 @@ func getSnapshot(claimName string, ns, snapshotClassName string) *unstructured.U
...
+}
+
+func genShortHash(s string) string {
+	h := sha1.New()
+	h.Write([]byte(s))
+	bs := h.Sum(nil)
+	return fmt.Sprintf("%x", bs)[:7]
+}
+func getPreProvisionedSnapshotContentName(snapshotHandle string) string {
+	return fmt.Sprintf("pre-provisioned-snapcontent-%s", genShortHash(snapshotHandle))
```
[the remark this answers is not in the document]
AndiLi99: I think having the name as it is now helps with debugging (logs). It makes it clear it is a pre-provisioned snapshot content.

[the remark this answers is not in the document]
saikat-royc: I mean pre-provisioned-snapcontent-<uuid>

test/e2e/storage/testsuites/snapshottable.go
AndiLi99: Noted in the code at line 437

test/e2e/storage/testsuites/snapshottable.go
```diff
@@ -401,12 +406,79 @@ func CreateSnapshotResource(sDriver SnapshottableTestDriver, config *PerTestConf
...
+		err = WaitForSnapshotReady(dc, r.Vs.GetNamespace(), r.Vs.GetName(), framework.Poll, framework.SnapshotCreateTimeout)
+		framework.ExpectNoError(err)
+
+		r.Vs, err = dc.Resource(SnapshotGVR).Namespace(r.Vs.GetNamespace()).Get(context.TODO(), r.Vs.GetName(), metav1.GetOptions{})
+
+		snapshotStatus := r.Vs.Object["status"].(map[string]interface{})
+		snapshotContentName := snapshotStatus["boundVolumeSnapshotContentName"].(string)
+		framework.Logf("received snapshotStatus %v", snapshotStatus)
+		framework.Logf("snapshotContentName %s", snapshotContentName)
+		framework.ExpectNoError(err)
```
saikat-royc, line 423: These steps of creating a dynamic snapshot and validation of basic properties is duplicated in both dynamic and pre-provisioned case. Can we create a helper function and use in both the cases?

test/e2e/storage/testsuites/snapshottable.go
```diff
@@ -460,7 +532,7 @@ func (sr *SnapshotResource) CleanupResource() error {
 		}
 	}
 	if sr.Vscontent != nil {
-		framework.Logf("deleting snapshot content %q/%q", sr.Vscontent.GetNamespace(), sr.Vscontent.GetName())
```
saikat-royc, line 463: keep the namespace print?

test/e2e/storage/testsuites/snapshottable.go
AndiLi99: What if we moved the common portion out of the switch and changed the switch statement to an if statement which only checks if it is a prepv test case?

saikat-royc: Yes that works too. Put a comment for the if part saying this is additional step for pre-provisioned snapshot case

saikat-royc (commented):
The changes looks good to me with some minor comments

test/e2e/storage/external/external.go
```diff
@@ -255,6 +255,10 @@ func (d *driverDefinition) SkipUnsupportedTest(pattern testpatterns.TestPattern)
 		if d.SnapshotClass.FromName || d.SnapshotClass.FromFile != "" || d.SnapshotClass.FromExistingClassName != "" {
```
msau42, line 255: Why is the empty string case on L252 mean that it's supported?

test/e2e/storage/external/external.go
```diff
@@ -255,6 +255,10 @@ func (d *driverDefinition) SkipUnsupportedTest(pattern testpatterns.TestPattern)
...
 			supported = true
 		}
+	case testpatterns.PreprovisionedCreatedSnapshot:
```
msau42, line 258: Since this check is exactly the same, can we combine the two cases into one line?

msau42 (commented):
/approve

Just some nits

test/e2e/storage/external/external.go
AndiLi99: git blame shows this commit https://github.com/kubernetes/kubernetes/commit/6644db9914379a4a7b3d3487b41b2010f226e4dc#diff-e5fc8d9911130b421b74b1ebc273f458R206

which is the first commit for this file. I don't know, perhaps @pohly can help.

pohly: It might have been needed for some tests which don't set a type explicitly, but I don't remember exactly. Most of the documentation for `SnapshotType` is just of the variety "XXXType is a type for XXX", which doesn't really help: https://github.com/kubernetes/kubernetes/blob/6644db9914379a4a7b3d3487b41b2010f226e4dc/test/e2e/storage/testpatterns/testpattern.go#L48-L63

PR #91285, opened 2020-05-20
WIP integration test services finalizers

aojea:
Integration test for #90753

PR #90753, opened 2020-05-05
Move IP de-allocation logic for Service into AfterDeleteFunc

MrHohn:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Ref https://github.com/kubernetes/kubernetes/issues/87603#issuecomment-579443444, the service registry currently doesn't handle object deletion in response to removing the last finalizer, specifically it fails to release the allocated IP.

This PR fixes that by moving the release logic into the AfterDelete hook.

**Which issue(s) this PR fixes**:

Fixes #87603

**Special notes for your reviewer**:
/assign @liggitt @thockin 

I'm honestly not sure if I'm using the AfterDelete thing right, so feedbacks welcome :)

**Does this PR introduce a user-facing change?**:

```release-note
Fix a bug that service's clusterIP is not released promptly after deletion with finalizers.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #93199, opened 2020-07-17
Skip repairing clusterIP and nodeport if a service is being deleted

### LINKED ISSUE #87603: service IPs and ports are not released when deleting a service via a finalizer-removing update
**What happened**:

* Created a service including a finalizer
* Triggered deletion of the service
* Removed finalizer
* Service got deleted by apiserver (not visible anymore via kubectl)
* Tried to create service again
* Creation got denied: `The Service "foo" is invalid: spec.ports[0].nodePort: Invalid value: 30003: provided port is already allocated`
  The apiserver logs the following lines prior to this happening:
  ```
  E0128 08:15:36.920788       1 repair.go:145] the node port 30003 for service foo/default is not allocated; repairing
  E0128 08:15:36.920837       1 repair.go:237] the cluster IP 10.0.0.81 for service foo/default is not allocated; repairing
  ```
* After about 10 minutes I'm able to create the service, the apiserver shows the following log lines when it is repairing it:
  ```
  E0128 08:28:51.429642       1 repair.go:184] the node port 30003 appears to have leaked: cleaning up
  E0128 08:28:51.436350       1 repair.go:311] the cluster IP 10.0.0.81 appears to have leaked: cleaning up
  ```

* Service is allowed to get created some seconds after deletion

```
cd $(mktemp -d)
mkdir etcd
docker run -d -p 2379:2379 --name=kube-etcd -v $(pwd)/etcd:/tmp/ --rm k8s.gcr.io/etcd:3.3.15 /usr/local/bin/etcd --data-dir /tmp/etcd --advertise-client-urls=http://0.0.0.0:2379 --listen-client-urls=http://0.0.0.0:2379
docker run -d --net=host --name=kube-apiserver --rm k8s.gcr.io/kube-apiserver:v1.17.2 kube-apiserver --etcd-servers http://127.0.0.1:2379 --insecure-port 8080 --authorization-mode=RBAC

export KUBECONFIG=$(pwd)/kubeconfig
touch $KUBECONFIG
kubectl config set-cluster etcd-local --server=http://localhost:8080
kubectl config set-context etcd-local --cluster=etcd-local
kubectl config use-context etcd-local

cat <<EOF > service.yaml
apiVersion: v1
kind: Service
metadata:
  name: foo
  finalizers:
  - foo.bar/some-finalizer
spec:
  ports:
  - port: 80
    protocol: TCP
    targetPort: 8080
    nodePort: 30003
  selector:
    app: kuard
  type: NodePort

EOF

for i in {1..200}; do
  echo "[$(date +%Y-%m-%d-%H:%M:%S)] # $i"
  kubectl apply -f service.yaml
  kubectl delete svc foo --wait=false
  sleep 1
  kubectl patch svc foo --type='json' -p='[{"op":"remove","path":"/metadata/finalizers"}]'
  kubectl delete svc foo --ignore-not-found
  sleep 1
done
```

Example output:
```
[2020-01-28-08:55:25] # 1                                                                        
service/foo unchanged                                                                                         
service "foo" deleted                                 
service/foo patched                                                      
...
[2020-01-28-08:58:21] # 77
service/foo created
service "foo" deleted
service/foo patched
[2020-01-28-08:58:23] # 78
service/foo created
service "foo" deleted
service/foo patched
[2020-01-28-08:58:26] # 79
The Service "foo" is invalid: spec.ports[0].nodePort: Invalid value: 30003: provided port is already allocated
Error from server (NotFound): services "foo" not found
Error from server (NotFound): services "foo" not found
[2020-01-28-08:58:28] # 80
The Service "foo" is invalid: spec.ports[0].nodePort: Invalid value: 30003: provided port is already allocated
Error from server (NotFound): services "foo" not found
Error from server (NotFound): services "foo" not found
[2020-01-28-08:58:30] # 81
The Service "foo" is invalid: spec.ports[0].nodePort: Invalid value: 30003: provided port is already allocated
Error from server (NotFound): services "foo" not found
Error from server (NotFound): services "foo" not found
...
[2020-01-28-09:07:23] # 5
The Service "foo" is invalid: spec.ports[0].nodePort: Invalid value: 30003: provided port is already allocated
Error from server (NotFound): services "foo" not found
Error from server (NotFound): services "foo" not found
[2020-01-28-09:07:26] # 6
service/foo created
service "foo" deleted
service/foo patched
```

* This does not always happen / this is a flakyness
* Also the problem get's auto-resolved by the apiserver after some time (But this may need about 10 minutes):
  `E0128 08:01:24.562044       1 repair.go:300] the cluster IP 10.0.0.215 may have leaked: flagging for later clean up`

* Background for us here is: we want to run a custom controller for services of `type: LoadBalancer` and want to use a finalizer. We did hit this issue sometimes during dev.

**Environment**:
  ```
  $ kubectl version
  Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.6", GitCommit:"72c30166b2105cd7d3350f2c28a219e6abcd79eb", GitTreeState:"clean", BuildDate:"2020-01-18T23:31:31Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
  Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.2", GitCommit:"59603c6e503c87169aea6106f57b9f242f64df89", GitTreeState:"clean", BuildDate:"2020-01-18T23:22:30Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
  ```
- Cloud provider or hardware configuration: none / locally reproducable / all are affected
  ```
  $ cat /etc/os-release
  NAME="Ubuntu"
  VERSION="18.04.3 LTS (Bionic Beaver)"
  ID=ubuntu
  ID_LIKE=debian
  PRETTY_NAME="Ubuntu 18.04.3 LTS"
  VERSION_ID="18.04"
  HOME_URL="https://www.ubuntu.com/"
  SUPPORT_URL="https://help.ubuntu.com/"
  BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
  PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
  VERSION_CODENAME=bionic
  UBUNTU_CODENAME=bionic
  ```
- Kernel (e.g. `uname -a`):
  ```
  $ uname -a
Linux <hostname hidden> 5.3.0-26-generic #28~18.04.1-Ubuntu SMP Wed Dec 18 16:40:14 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
  ```
- Install tools:
- Others:

MrHohn:
**What type of PR is this?**
/kind bug

Ref https://github.com/kubernetes/kubernetes/pull/90753#issuecomment-660268232, we are leaking clusterIP and nodeport because the repair controller mistakenly repairs them even if a service is being deleted with finalizer.

This PR fixes the repair controller to also check for deletionTimestamp before repairing.

I will work on the E2E test (as requested on https://github.com/kubernetes/kubernetes/pull/90753#issuecomment-626873028, probably adapting #91285) when I get a chance :)

Fixes #87603

/assign @liggitt @bowei 
cc @aojea @thockin 

```release-note
Fix a bug that service's clusterIP and nodeport is not released promptly after deletion with finalizers.
```

```docs

```

### DIFF
--- a/pkg/registry/core/service/ipallocator/controller/repair.go
+++ b/pkg/registry/core/service/ipallocator/controller/repair.go
@@ -22,7 +22,7 @@ import (
 	"net"
 	"time"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/runtime"
@@ -212,8 +212,10 @@ func (c *Repair) runOnce() error {
 
 	// Check every Service's ClusterIP, and rebuild the state as we think it should be.
 	for _, svc := range list.Items {
-		if !helper.IsServiceIPSet(&svc) {
-			// didn't need a cluster IP
+		if svc.ObjectMeta.DeletionTimestamp != nil ||
+			!helper.IsServiceIPSet(&svc) {
+			// Skip repairing if service is being deleted
+			// or didn't need a cluster IP.
 			continue
 		}
 		ip := net.ParseIP(svc.Spec.ClusterIP)
--- a/pkg/registry/core/service/ipallocator/controller/repair_test.go
+++ b/pkg/registry/core/service/ipallocator/controller/repair_test.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"strings"
 	"testing"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -169,6 +170,15 @@ func TestRepairWithExisting(t *testing.T) {
 			ObjectMeta: metav1.ObjectMeta{Namespace: "six", Name: "six"},
 			Spec:       corev1.ServiceSpec{ClusterIP: "None"},
 		},
+		&corev1.Service{ // being deleted with finalizer, skipped
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace:         "seven",
+				Name:              "seven",
+				DeletionTimestamp: &metav1.Time{Time: time.Now()},
+				Finalizers:        []string{"test-finalizer"},
+			},
+			Spec: corev1.ServiceSpec{ClusterIP: "192.168.2.1"},
+		},
 	)
 
 	ipregistry := &mockRangeRegistry{
@@ -188,7 +198,9 @@ func TestRepairWithExisting(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if !after.Has(net.ParseIP("192.168.1.1")) || !after.Has(net.ParseIP("192.168.1.100")) {
+	if !after.Has(net.ParseIP("192.168.1.1")) ||
+		!after.Has(net.ParseIP("192.168.1.100")) ||
+		after.Has(net.ParseIP("192.168.2.1")) {
 		t.Errorf("unexpected ipallocator state: %#v", after)
 	}
 	if free := after.Free(); free != 252 {
--- a/pkg/registry/core/service/portallocator/controller/repair.go
+++ b/pkg/registry/core/service/portallocator/controller/repair.go
@@ -130,6 +130,10 @@ func (c *Repair) runOnce() error {
 	// Check every Service's ports, and rebuild the state as we think it should be.
 	for i := range list.Items {
 		svc := &list.Items[i]
+		if svc.ObjectMeta.DeletionTimestamp != nil {
+			// Skip repairing if service is being deleted.
+			continue
+		}
 		ports := collectServiceNodePorts(svc)
 		if len(ports) == 0 {
 			continue
--- a/pkg/registry/core/service/portallocator/controller/repair_test.go
+++ b/pkg/registry/core/service/portallocator/controller/repair_test.go
@@ -22,6 +22,7 @@ import (
 	"sort"
 	"strings"
 	"testing"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -179,6 +180,17 @@ func TestRepairWithExisting(t *testing.T) {
 				HealthCheckNodePort: 144,
 			},
 		},
+		&corev1.Service{ // being deleted with finalizer, skipped
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace:         "seven",
+				Name:              "seven",
+				DeletionTimestamp: &metav1.Time{Time: time.Now()},
+				Finalizers:        []string{"test-finalizer"},
+			},
+			Spec: corev1.ServiceSpec{
+				Ports: []corev1.ServicePort{{NodePort: 222}},
+			},
+		},
 	)
 
 	registry := &mockRangeRegistry{
@@ -198,7 +210,7 @@ func TestRepairWithExisting(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if !after.Has(111) || !after.Has(122) || !after.Has(133) || !after.Has(144) {
+	if !after.Has(111) || !after.Has(122) || !after.Has(133) || !after.Has(144) || after.Has(222) {
 		t.Errorf("unexpected portallocator state: %#v", after)
 	}
 	if free := after.Free(); free != 97 {


liggitt:
currently, the code in master does the following when a DELETE request is issued for a service:
* processes the deletion
  * if there are no finalizers, removes the service object from etcd
  * if there are finalizers, sets deletionTimestamp and leaves the service present in etcd
* deallocates the clusterIP and node port

In the finalizer case, this leaves a service present in etcd with a clusterIP and node port which are no longer considered allocated, which has several implications:
1. If another delete request is issued for the service, the clusterIP/nodePort will be deallocated again
  a. If no other service grabbed that ip or port in the meantime, this is seen as an error (attempt to free an unallocated thing)
  b. If another service grabbed that ip or port in the meantime, this incorrectly deallocates the IP or port another service is actually using
2. If an update request is issued for the service while it is finalizing, that can exercise deallocate/allocate cases
  a. The deallocate cases have the same failure modes as scenario 1
  b. The allocate cases result in an allocated IP or node associated with a service with a deletionTimestamp. This PR changes the allocator repair loop to undo this eventually.

A fundamental question is whether an IP or port associated with a service that still exists in etcd but which has a deletionTimestamp+finalizers should be considered allocated. If things that program the network still honor that IP or port, it seems like it should be considered allocated.

MrHohn:
> In the finalizer case, this leaves a service present in etcd with a clusterIP and node port which are no longer considered allocated, which has several implications:

Thanks @liggitt for the raising those, definitely something I haven't thought of.

> A fundamental question is whether an IP or port associated with a service that still exists in etcd but which has a deletionTimestamp+finalizers should be considered allocated. If things that program the network still honor that IP or port, it seems like it should be considered allocated.

I was wanting to avoid changing the existing behavior (when to release the IP/port). Given the issue raised, I don't think this PR is the real fix then.  I'm also leaning toward that we should honer the allocated IP/port until the service is removed from etcd.

@bowei @thockin @kubernetes/sig-network-pr-reviews would be great to hear your thought on this, thanks!

liggitt:
> I'm also leaning toward that we should honer the allocated IP/port until the service is removed from etcd.

That seems right at first glance.

I think that means:
* the Delete() method should only deallocate in non-finalizer cases
* all Update() paths that can result in a deletion (when an update is made that removes all remaining finalizers from a service with a deletionTimestamp set) should deallocate

At first glance, the `AfterDelete` hook on the REST storage seems to be called from all of those locations, and we could move the endpoints delete and deallocate into an `AfterDelete` hook. Things to watch out for on that approach are:
* We don't appear to have *any* uses of AfterDelete in the codebase currently (though service's post-deletion deallocation is also unique, so... 🤷 )
* Three of the six calls to finalizeDelete invoke the AfterDelete hook when I'm not sure they should... I think AfterDelete should only be invoked when a storage.Delete call succeeds:

```diff
--- a/staging/src/k8s.io/apiserver/pkg/registry/generic/registry/store.go
+++ b/staging/src/k8s.io/apiserver/pkg/registry/generic/registry/store.go
@@ -433,7 +433,7 @@ func (e *Store) deleteWithoutFinalizers(ctx context.Context, name, key string, o
                // requests to remove all finalizers from the object, so we
                // ignore the NotFound error.
                if storage.IsNotFound(err) {
-                       _, err := e.finalizeDelete(ctx, obj, true)
+                       _, err := e.finalizeDelete(ctx, obj, false)
                        // clients are expecting an updated object if a PUT succeeded,
                        // but finalizeDelete returns a metav1.Status, so return
                        // the object in the request instead.
@@ -881,7 +881,7 @@ func (e *Store) updateForGracefulDeletionAndFinalizers(ctx context.Context, name
                // we should fall through and truly delete the object.
                return nil, false, true, out, lastExisting
        case errAlreadyDeleting:
-               out, err = e.finalizeDelete(ctx, in, true)
+               out, err = e.finalizeDelete(ctx, in, false)
                return err, false, false, out, lastExisting
        default:
                return storeerr.InterpretUpdateError(err, e.qualifiedResourceFromContext(ctx), name), false, false, out, lastExisting
@@ -971,7 +971,7 @@ func (e *Store) Delete(ctx context.Context, name string, deleteValidation rest.V
                if storage.IsNotFound(err) && ignoreNotFound && lastExisting != nil {
                        // The lastExisting object may not be the last state of the object
                        // before its deletion, but it's the best approximation.
-                       out, err := e.finalizeDelete(ctx, lastExisting, true)
+                       out, err := e.finalizeDelete(ctx, lastExisting, false)
                        return out, true, err
                }
                return nil, false, storeerr.InterpretDeleteError(err, qualifiedResource, name)
```

MrHohn:
> We don't appear to have any uses of AfterDelete in the codebase currently (though service's post-deletion deallocation is also unique, so... 🤷 )

Right, my previous implementation (#90753) isn't working 100% as expected and I was having trouble looking for existing usage on this :)

> Three of the six calls to finalizeDelete invoke the AfterDelete hook when I'm not sure they should... I think AfterDelete should only be invoked when a storage.Delete call succeeds:

Thanks, sounds like something I need to double check on if taking that approach.

aojea:
@liggitt unrelated to the PR but related to IMHO the main problem with services,
is there any possibility to not have to use the allocators and use the storage directly?

I asked Daniel Smith in slack and he told me we need "indexes", but I'm very naive in this area :smile: so please bear with me, is that something impossible to do? hard to do? not worth to do?

> can I reject without using the ip allocator the creation of a new service svc2.Spec.ClusterIP = "10.1.2.2" , if I already have one with the same IP svc1.Spec.ClusterIP = "10.1.2.2"
> 
> lavalamp  24 days ago
>..., but the answer is no, we don't have indexes, which would be needed for that

liggitt:
> @liggitt unrelated to the PR but related to IMHO the main problem with services,
> is there any possibility to not have to use the allocators and use the storage directly?

not quite sure what you mean, but if you're asking whether we could just use an in-memory index based on the persisted services to determine which IPs/ports are in use, the answer is no:

 * Two requests creating two different services using IP 1.1.1.1 and nodePort 1234 could arrive to two different API servers at the same time.
* Both servers would think the IP/port was available, and would allow the services to be persisted.
* Then the API servers would observe the service created via the other API server, and you'd have double-allocated IPs+ports.

BenTheElder:
seems like perhaps this should be in-milestone for 1.19? and possibly backported?
/retest

BenTheElder:
> RemoveEndpoints timed out

Hmm

aojea:
I think this fix the bug in the repair logic, but I don't have clear from the comments if this PR will target the finalizers behavior too.

> A fundamental question is whether an IP or port associated with a service that still exists in etcd but which has a deletionTimestamp+finalizers should be considered allocated. If things that program the network still honor that IP or port, it seems like it should be considered allocated.

I think the same, but this seems a quite important change to implement at this stage of the release.
Maybe we can go with the repair fix and address the finalizer behavior in 1.20?

liggitt:
> Maybe we can go with the repair fix and address the finalizer behavior in 1.20?

With the change in this PR, what happens if an update which forces an allocate (either of an IP or a port) is done to the service being deleted?

aojea:
> > Maybe we can go with the repair fix and address the finalizer behavior in 1.20?
> 
> With the change in this PR, what happens if an update which forces an allocate (either of an IP or a port) is done to the service being deleted?

So, if I read it correctly, the repair loops runs every 3 mins and does the following:

1. get the bitmap from etcd with the allocated resources (ports or IPs) (storedAllocator)
2. rebuild the allocator creating a new one (actualAllocator)
3. list current services from the API and loop over them, allocating the ClusterIPs in the new allocator (actualAllocator)
4. compare stored and actual Allocators, to find leaked IPs

I think that this only reduces the possibility of the repair loops to interact with the allocation/deallocation process, not taking into account the service that is going to be deleted in the step 3 and recovering that IP that was already deleted, because if a new service requesting the same IP is created until the repair loop realize it's leaked, it will fail.

In the other hand, there is the situation with Services and how allocate resources that you described here https://github.com/kubernetes/kubernetes/pull/93199#issuecomment-660340444 and https://github.com/kubernetes/kubernetes/issues/87603#issuecomment-579443444. 

That's what I said that maybe they are related topics but can be handled separately, and the later seems complicated due to the amount of time that this was working this way and the importance of the Services in Kubernetes.

What is clear is that we need more coverage on this area, there are many moving parts :smile:

thockin:
@MrHohn - This PR is old but still open.  I assume it is still an issue?

MrHohn:
@thockin Sorry for my lack of updates. If we can get https://github.com/kubernetes/kubernetes/pull/90753 in, this PR will no longer be needed.

PR #93201, opened 2020-07-17
Filters on kubemark nodegroups when selecting node template to enable hollow node clusters that contain real nodes.

### LINKED ISSUE #93200: Support Real and Hollow node configurations in the same cluster.
Kubemark controller should only use nodes for its template that are hollow nodes, rather than any node.

**Why is this needed**:
I want to perf-test the cluster autoscaler using clusters the contain real and hollow nodes. The original integration between cluster autoscaler and kubemark assumed multiple clusters, but I would like to enable using a single cluster for simplicity. I.e. a single node group with kubemark pods and many node groups of hollow kubemark nodes.

ellistarn:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind flake

/kind feature

Kubemark controller detects existing kubemark nodes to determine the node template they use. One consumer is the cluster autoscaler, which uses this node template to create more kubemark nodes. This works well for kubemark setups which only have hollow nodes in a cluster, but for clusters that have both real and hollow nodes, this code can accidentally select real nodes, which causes breaking behavior. 

Fixes #93200

```release-note
Kubemark now supports both real and hollow nodes in a single cluster.
```

```docs

```

### DIFF
--- a/pkg/kubemark/controller.go
+++ b/pkg/kubemark/controller.go
@@ -351,7 +351,8 @@ func (kubemarkController *KubemarkController) runNodeCreation(stop <-chan struct
 }
 
 func (kubemarkCluster *kubemarkCluster) getHollowNodeName() (string, error) {
-	nodes, err := kubemarkCluster.nodeLister.List(labels.Everything())
+	selector, _ := labels.Parse(nodeGroupLabel)
+	nodes, err := kubemarkCluster.nodeLister.List(selector)
 	if err != nil {
 		return "", err
 	}


pkg/kubemark/controller.go
```diff
@@ -355,8 +355,13 @@ func (kubemarkCluster *kubemarkCluster) getHollowNodeName() (string, error) {
 	if err != nil {
 		return "", err
 	}
+	klog.Infof("TESTINGTESTING")
```
ellistarn, line 358: Oops. This will be removed.

pkg/kubemark/controller.go
```diff
@@ -356,7 +356,11 @@ func (kubemarkCluster *kubemarkCluster) getHollowNodeName() (string, error) {
 		return "", err
 	}
 	for _, node := range nodes {
-		return node.Name, nil
+		for key := range node.ObjectMeta.Labels {
```
shyamjvs, line 359: You can directly list using a label-selector in the List() call above instead of filtering the results yourself.

ellistarn: Done

ellistarn:
What are the blockers here? Is 1.19 still frozen?

shyamjvs:
/milestone v1.20

Testing improvement.

PR #93207, opened 2020-07-18
Use local daemonset manifest for installing Nvidia drivers

### LINKED ISSUE #93187: [Failing Test] [sig-scheduling] [Feature:GPUDevicePlugin] run Nvidia GPU Device Plugin tests (ci-kubernetes-e2e-gce-device-plugin-gpu)
**Which jobs are failing**:

`gce-device-plugin-gpu-master (ci-kubernetes-e2e-gce-device-plugin-gpu)`

`[sig-scheduling] [Feature:GPUDevicePlugin] run Nvidia GPU Device Plugin tests`

`07-17-20 02:31 CDT`

**Testgrid link**:

https://testgrid.k8s.io/sig-release-master-blocking#gce-device-plugin-gpu-master

**Reason for failure**:

```
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/scheduling/nvidia-gpus.go:218
Jul 17 07:47:26.988: Timed out after 600.025s.
Expected
    <bool>: false
to be true
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/scheduling/nvidia-gpus.go:169
```

/sig scheduling
/cc @kubernetes/ci-signal @kubernetes/sig-scheduling-misc 
/priority critical-urgent
/milestone v1.19

hasheddan:
**What type of PR is this?**

/kind failing-test

Updates sig-scheduling e2e Nvidia GPU tests to install drivers using
local manifest by default. Currently the DaemonSet is fetched from the
GoogleCloudPlatform/container-enginer-accelerators repo by default.
Using a local manifest allows for manually specifying the image
cos-gpu-installer image rather than always using latest. A remote
manifest can still be fetched by setting
NVIDIA_DRIVER_INSTALLER_DAEMONSET env var.

Fixes #93187 

We may want to move back to specifying `latest` for the `cos-gpu-installer` image, but this will allow us to do that as desired when we want to avoid issues like the one described in #93187. Currently we get the advantage of picking up any changes in https://github.com/GoogleCloudPlatform/container-engine-accelerators, but it appears that repo is primarily for the GKE scenario where the image can be preloaded:
Ref: https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/nvidia-driver-installer/cos/daemonset-preloaded.yaml
Ref: https://cloud.google.com/kubernetes-engine/docs/how-to/gpus#installing_drivers

Looking back at the commit that added the preloaded version (https://github.com/GoogleCloudPlatform/container-engine-accelerators/commit/4fd24fa94e5a392d7560dab3571d9af4441cc144) it seems that the only reason the `daemonset.yaml` manifest is still maintained there is because k/k tests are referencing it. Because the GCP repo is primarily meant for GKE, I think it likely makes more sense to maintain the manifest for testing a cluster created using `kube-up.sh` in the k/k repo. If there is strong desire to continue to consume from the GCP repo, I would be happy to open a PR to pin that image to `gcr.io/cos-cloud/cos-gpu-installer:v20200701` until the issue with the `latest` image is resolved. Other options could be:
- Merging this and then setting `NVIDIA_DRIVER_INSTALLER_DAEMONSET` if we want to switch back to using the remote manifest
- Modifying this PR to continue to use the remote by default, but then setting `NVIDIA_DRIVER_INSTALLER_DAEMONSET` in the job config that runs this test to point to a separate manifest that we maintain (that would look like the one included in this PR)

/cc @Huang-Wei @ahg-g @liggitt 
/sig scheduling
/milestone v1.19
/priority critical-urgent

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/scheduling/nvidia-gpus.go
+++ b/test/e2e/scheduling/nvidia-gpus.go
@@ -22,6 +22,7 @@ import (
 	"regexp"
 	"time"
 
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -36,6 +37,7 @@ import (
 	"k8s.io/kubernetes/test/e2e/framework/providers/gce"
 	e2eresource "k8s.io/kubernetes/test/e2e/framework/resource"
 	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+	e2etestfiles "k8s.io/kubernetes/test/e2e/framework/testfiles"
 	imageutils "k8s.io/kubernetes/test/utils/image"
 
 	"github.com/onsi/ginkgo"
@@ -50,7 +52,6 @@ const (
 
 var (
 	gpuResourceName v1.ResourceName
-	dsYamlURL       string
 )
 
 func makeCudaAdditionDevicePluginTestPod() *v1.Pod {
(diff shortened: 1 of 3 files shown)


test/e2e/scheduling/nvidia-gpus.go
```diff
@@ -128,20 +129,26 @@ func getGPUsAvailable(f *framework.Framework) int64 {
 func SetupNVIDIAGPUNode(f *framework.Framework, setupResourceGatherer bool) *framework.ContainerResourceGatherer {
 	logOSImages(f)
 
+	ds := &appsv1.DaemonSet{}
```
Huang-Wei, line 132: nit: define a bare variable here:

```
var ds *appsv1.DaemonSet
```

and then assign the DaemonSet at L137 and L145.

test/e2e/testing-manifests/scheduling/nvidia-driver-installer.yaml
```diff
@@ -0,0 +1,80 @@
...
+        - name: nvidia-install-dir-host
+          mountPath: /usr/local/nvidia
+        - name: vulkan-icd-mount
+          mountPath: /etc/vulkan/icd.d
+        - name: dev
+          mountPath: /dev
+        - name: root-mount
+          mountPath: /root
+      containers:
+      - image: "gcr.io/google-containers/pause:2.0"
```
Huang-Wei, line 79: nit: we can use latest image `pause:3.2`

hasheddan:
@Huang-Wei thanks for the review! Both updates have been made :+1:

Huang-Wei:
/approve
/lgtm
from scheduling side

hasheddan:
@liggitt would you mind giving this a look?

liggitt:
/sig node

this looks ok as a workaround, can we keep an issue open/assigned to sig-node to resolve the issue with the test image?

PR #92220, opened 2020-06-17
fix the timeline for CRD v1beta1 removal

deads2k:
fixes https://github.com/kubernetes/kubernetes/issues/92110

/kind bug
/priority important-soon
@kubernetes/sig-api-machinery-api-reviews 

```release-note
NONE
```

/assign @liggitt

Issue #93210, opened 2020-07-18
MirrorPodWithGracePeriod test flaking in 5% of pull-kubernetes-node-e2e runs


liggitt:
These tests have flaked consistently since https://github.com/kubernetes/kubernetes/pull/92442 merged: https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&text=not%20found&test=MirrorPodWithGracePeriod

in many cases, they are the only failure in a test run:
* https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-node-e2e/1284519583780179968
* https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-node-e2e/1284343925561102336
* https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/batch/pull-kubernetes-node-e2e/1284519583780179968/

Either the functionality introduced by that PR is not reliable, or the test is faulty.

/sig node
/priority critical-urgent
/assign @tedyu @derekwaynecarr

liggitt:
Looks like this is failing ~5% of node-e2e runs

justaugustus:
FYI @kubernetes/ci-signal

sjenning:
Looks like there might still be a race somewhere looking at the kubelet logs

Success case:
```
kubelet[1948]: I0720 13:17:30.646802    1948 kubelet.go:1823] SyncLoop (REMOVE, "file"): "graceful-pod-bf354966-a220-49e2-b6d7-d33cd5d28c7e-tmp-node-e2e-9fe4e632-cos-77-12371-274-0_mirror-pod-with-grace-period-9360(636522eeddcbefb43f6cc92784f2ba57)"
kubelet[1948]: I0720 13:17:30.646825    1948 kubelet_pods.go:1212] marking pod pending termination "636522eeddcbefb43f6cc92784f2ba57"
kubelet[1948]: I0720 13:17:30.646879    1948 kubelet_pods.go:1247] Killing unwanted pod "graceful-pod-bf354966-a220-49e2-b6d7-d33cd5d28c7e-tmp-node-e2e-9fe4e632-cos-77-12371-274-0"
[...]

PR #93211, opened 2020-07-18
Make CRDList lifecycle consistent with CRD

liggitt:
**What type of PR is this?**
/kind cleanup

xref https://github.com/kubernetes/kubernetes/pull/92220#issuecomment-660491491

```release-note
NONE
```

/assign @deads2k 
/sig api-machinery


liggitt:
#93210 flake in https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/93211/pull-kubernetes-node-e2e/1284547265532268544/

cbandy:
I see one more `+k8s:prerelease-lifecycle-gen:deprecated=1.19` on `ConversionReview` which was introduced in the same commit as the others, 555225eb29124a83cdc5c01ebe01c605d74450c6:

https://github.com/kubernetes/kubernetes/blob/66020d22923055133f5f05ca16fbfb66a79053c5/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1/types.go#L475-L479

I do not know anything about these, but thought I should point it out.

Issue #81245, opened 2019-08-09
Core e2e test framework should not import sub e2e test frameworks


oomichi:
**What happened**:

As https://kubernetes.slack.com/archives/C9NK9KFFW/p1565372337078100 Core e2e test framework(`test/e2e/framework`) should not import sub e2e test frameworks (e.g. `test/e2e/framework/auth`) for avoiding circular dependency.
This issue is for managing such issues.

The following files are in e2e core framework:
- [x] cleanup.go (no dependency to sub packages)
- [x] exec_util.go (no dependency to sub packages)
- [x] expect.go (no dependency to sub packages)
- [x] flake_reporting_util.go (no dependency to sub packages)
- [ ] framework.go (https://github.com/kubernetes/kubernetes/pull/86679 is a part, but still we should have more PRs)
- [x] get-kubemark-resource-usage.go
- [x] google_compute.go (no dependency to sub packages)
- [ ] log.go
- [ ] log_size_monitoring.go
- [ ] nodes_util.go
- [ ] pods.go
- [x] profile_gatherer.go (https://github.com/kubernetes/kubernetes/pull/85304)
- [x] provider.go (no dependency to sub packages)
- [ ] psp.go
- [x] rc_util.go (no dependency to sub packages)
- [ ] resource_usage_gatherer.go
- [x] size.go (no dependency to sub packages)
- [x] skip.go (https://github.com/kubernetes/kubernetes/pull/87031 and https://github.com/kubernetes/kubernetes/issues/87047)
- [x] suites.go (https://github.com/kubernetes/kubernetes/pull/85235)
- [x] test_context.go (no dependency to sub packages)
- [ ] util.go

oomichi:
Now the core framework imports the following sub frameworks:
```
k8s.io/kubernetes/test/e2e/framework/config: 1
k8s.io/kubernetes/test/e2e/framework/kubelet: 3
k8s.io/kubernetes/test/e2e/framework/log: 20
k8s.io/kubernetes/test/e2e/framework/metrics: 7
k8s.io/kubernetes/test/e2e/framework/node: 3
k8s.io/kubernetes/test/e2e/framework/pod: 9
k8s.io/kubernetes/test/e2e/framework/resource: 1
k8s.io/kubernetes/test/e2e/framework/ssh: 5
k8s.io/kubernetes/test/e2e/framework/auth: 1
k8s.io/kubernetes/test/e2e/framework/ginkgowrapper: 1
[...]

PR #93081, opened 2020-07-14
Move MasterUpgrade() out from e2e framework

oomichi:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

MasterUpgrade() is called only from cloud/gcp/cluster_upgrade.go.
And the function depends on GCP, so it is nice to move this function from e2e framework.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #93213, opened 2020-07-18
Move kubectl to e2e framework subpackage

helenfeng737:
**What type of PR is this?**
/kind cleanup

We'd like to make e2e framework core as lean as possible, meaning it shall depend on the fewest number of sub packages. This PR is to move all kubectl related functions into e2e kubectl subpackage and e2e fw core no longer need to depend on e2e kubectl subpackage. 

This PR is part of: https://github.com/kubernetes/kubernetes/issues/81245
In e2e fw core's `util.go` file, 
```
// TODO: Remove the following imports (ref: https://github.com/kubernetes/kubernetes/issues/81245)		
	e2ekubectl "k8s.io/kubernetes/test/e2e/framework/kubectl"
```
I removed e2e fw core's dependency on e2e kubectl subpackage.  

This PR is to move all the kubectl related functions from e2e framework core into e2e kubectl subpackage. Some functions should be moved to e2epod subpackage, I'll address them in the following PRs. Therefore, this PR is all about refactoring kubectl subpackage. 

```release-note
None
```

```docs
None
```


SataQiu (commented):
Thanks 
/ok-to-test

ZhiFeng1993:
cc: @oomichi

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,345 @@
...
+		return err
+	}
+
+	WaitForSSHTunnels(namespace)
+
+	return nil
+}
+
+// MasterUpgrade upgrades master node on GCE/GKE.
+func MasterUpgrade(f *framework.Framework, v string) error {
```
oomichi, line 330: This function is moved to better place with https://github.com/kubernetes/kubernetes/pull/93081
It is nice to wait for merging the other PR and rebase this again.

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,345 @@
...
+	case "gke":
+		return MasterUpgradeGKE(f.Namespace.Name, v)
+	default:
+		return fmt.Errorf("MasterUpgrade() is not implemented for provider %s", framework.TestContext.Provider)
+	}
+}
+
+// MasterUpgradeGCEWithKubeProxyDaemonSet upgrades master node on GCE with enabling/disabling the daemon set of kube-proxy.
+// TODO(mrhohn): Remove this function when kube-proxy is run as a DaemonSet by default.
+func MasterUpgradeGCEWithKubeProxyDaemonSet(v string, enableKubeProxyDaemonSet bool) error {
```
oomichi, line 343: ditto

oomichi (commented):
Nice work, thanks for taking big effort for that.

Just one comment.

test/e2e/framework/kubectl/kubectl_builder.go
ZhiFeng1993: Sure. thanks

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,345 @@
...
+// Add more functions to customize the builder as needed.
+type KubectlBuilder struct {
+	cmd     *exec.Cmd
+	timeout <-chan time.Time
+}
+
+// NewKubectlCommand returns a KubectlBuilder for running kubectl.
+func NewKubectlCommand(namespace string, args ...string) *KubectlBuilder {
+	b := new(KubectlBuilder)
+	tk := NewTestKubeconfig(framework.TestContext.CertDir, framework.TestContext.Host, framework.TestContext.KubeConfig, framework.TestContext.KubeContext, framework.TestContext.KubectlPath, namespace)
```
alejandrox1, line 50: This is a bit of an antipattern that we have because it relies on the magic of the e2e framework which makes things hard to debug/track sometimes.
In terms of code organization, it would also be best to not import the test framework within one of its subpackages (specially if it is for passing this sort of data - later on its used for logging so :shrug: ).

For this, I think, the best thing to do would be to refactor it in such a way that `e2ekubectl.NewTestKubeconfig` was built in the code calling `e2ekubectl` and then the `e2ekubectl.TestKubeconfig` (https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/kubectl/kubectl_utils.go) instance was passed around to this function.
Ultimately changing the signature here to something like

```go
// NewKubectlCommand returns a KubectlBuilder for running kubectl.
func NewKubectlCommand(tk *TestKubeconfig) *KubectlBuilder {
  ...
```

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+// RunKubectlInput is a convenience wrapper over kubectlBuilder that takes input to stdin
+func RunKubectlInput(namespace string, data string, args ...string) (string, error) {
+	return NewKubectlCommand(namespace, args...).WithStdinData(data).Exec()
+}
+
+// RunKubemciWithKubeconfig is a convenience wrapper over RunKubemciCmd
+func RunKubemciWithKubeconfig(args ...string) (string, error) {
+	if framework.TestContext.KubeConfig != "" {
```
alejandrox1, line 179: This kind of data should be explicitly passed as an argument.
Something like
```go
// RunKubemciWithKubeconfig is a convenience wrapper over RunKubemciCmd
func RunKubemciWithKubeconfig(kubeconfig string, args ...string) (string, error) {
  ...
```

Getting data from the framework, secretly, inside of functions is a weird pattern.

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,345 @@
...
+	return RunKubemciCmd(args...)
+}
+
+// RunKubemciCmd is a convenience wrapper over kubectlBuilder to run kubemci.
+// It assumes that kubemci exists in PATH.
+func RunKubemciCmd(args ...string) (string, error) {
+	// kubemci is assumed to be in PATH.
+	kubemci := "kubemci"
+	b := new(KubectlBuilder)
+	args = append(args, "--gcp-project="+framework.TestContext.CloudConfig.ProjectID)
```
alejandrox1, line 191: ProjectID should be passed explicitly as an argument to this function.

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,345 @@
...
+	})
+}
+
+// MasterUpgradeGKE upgrades master node to the specified version on GKE.
+func MasterUpgradeGKE(namespace string, v string) error {
+	framework.Logf("Upgrading master to %q", v)
+	args := []string{
+		"container",
+		"clusters",
+		fmt.Sprintf("--project=%s", framework.TestContext.CloudConfig.ProjectID),
```
alejandrox1, line 311: ProjectID should be explicitly passed as an argument.

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,345 @@
...
+		framework.LocationParamGKE(),
```
alejandrox1, line 312: The result of this called should be explicitly passed as an arguement as well, instead of being done inside of the function.

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,345 @@
...
+		"upgrade",
+		framework.TestContext.CloudConfig.Cluster,
```
alejandrox1, line 314: Cluster should be explicitly passed as an argument.

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,345 @@
...
+		"--master",
+		fmt.Sprintf("--cluster-version=%s", v),
+		"--quiet",
+	}
+	_, _, err := framework.RunCmd("gcloud", framework.AppendContainerCommandGroupIfNeeded(args)...)
```
alejandrox1, line 319: Call to "AppendContainerCommandGroupIfNeeded" should also be done outside.

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,345 @@
...
+	kubemci := "kubemci"
+	b := new(KubectlBuilder)
+	args = append(args, "--gcp-project="+framework.TestContext.CloudConfig.ProjectID)
+
+	b.cmd = exec.Command(kubemci, args...)
+	return b.Exec()
+}
+
+// CreateEmptyFileOnPod creates empty file at given path on the pod.
+// TODO(alejandrox1): move to subpkg pod once kubectl methods have been refactored.
```
alejandrox1, line 198: This comment can be removed :tada:

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,345 @@
...
+
+		l, _ := RunKubectl(ns, "logs", s.Name, fmt.Sprintf("--namespace=%v", ns), "--tail=100")
+		framework.Logf("\nLast 100 log lines of %v:\n%v", s.Name, l)
+	}
+}
+
+// lookForString looks for the given string in the output of fn, repeatedly calling fn until
+// the timeout is reached or the string is found. Returns last log and possibly
+// error if the string was not found.
+// TODO(alejandrox1): move to pod/ subpkg once kubectl methods are refactored.
```
alejandrox1, line 219: This comment can also be removed :-)

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,345 @@
...
+			return
+		}
+	}
+	err = fmt.Errorf("Failed to find \"%s\", last result: \"%s\"", expectedString, result)
+	return
+}
+
+// LookForStringInPodExec looks for the given string in the output of a command
+// executed in a specific pod container.
+// TODO(alejandrox1): move to pod/ subpkg once kubectl methods are refactored.
```
alejandrox1, line 233: This comment can be removed.

alejandrox1 (changes requested):
Thank you for working on this @ZhiFeng1993 !
Left some minor comments on how to refactor the kubectl test utility functions to not hide away they data dependencies from the e2e framework.

test/e2e/framework/kubectl/kubectl_builder.go
ZhiFeng1993: Good point. Will work on it.

test/e2e/framework/kubectl/kubectl_builder.go
ZhiFeng1993: Similar to the previous one, will do.

test/e2e/framework/kubectl/kubectl_builder.go
ZhiFeng1993: Same:)

test/e2e/framework/kubectl/kubectl_builder.go
ZhiFeng1993: :) !

test/e2e/framework/kubectl/kubectl_builder.go
ZhiFeng1993: :) !

test/e2e/framework/kubectl/kubectl_builder.go
ZhiFeng1993: :) !

test/e2e/framework/kubectl/kubectl_builder.go
ZhiFeng1993: Makes sense.

test/e2e/framework/kubectl/kubectl_builder.go
ZhiFeng1993: Makes sense.

test/e2e/framework/kubectl/kubectl_builder.go
ZhiFeng1993: I agree with the idea, but here 'args’, which will be passed to "AppendContainerCommandGroupIfNeeded", is constructed inside this function, then "AppendContainerCommandGroupIfNeeded" adds some other args to the original 'args'. So it's hard to move it outside the function. Also here is using "AppendContainerCommandGroupIfNeeded" consistently with other places(e.g. https://github.com/kubernetes/kubernetes/blob/master/test/e2e/cloud/gcp/cluster_upgrade.go#L672)

Another option is to get rid of this function entirely, since the logic in the function is not complicated and there are only a couple of places being used. What do you think?

alejandrox1: I think the main difference between this instance and https://github.com/kubernetes/kubernetes/blob/master/test/e2e/cloud/gcp/cluster_upgrade.go#L672
Is that framework code is meant to be imported by other test, while the later is a test itself.
The problem with doing it inside of the kubectl package is that it obscures what is going on by relying on the core framework.

---

> Another option is to get rid of this function entirely, since the logic in the function is not complicated and there are only a couple of places being used. What do you think?

I think this would also be a descent option, if for example, the same test package is the only place where it is used.

test/e2e/framework/kubectl/kubectl_builder.go
ZhiFeng1993: Makes sense to me.  Will do.

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,352 @@
...
+		return "", "", fmt.Errorf("timed out waiting for command %v:\nCommand stdout:\n%v\nstderr:\n%v", cmd, cmd.Stdout, cmd.Stderr)
+	}
+	framework.Logf("stderr: %q", stderr.String())
+	framework.Logf("stdout: %q", stdout.String())
+	return stdout.String(), stderr.String(), nil
+}
+
+// RunKubectlOrDie is a convenience wrapper over Builder
+func RunKubectlOrDie(namespace string, args ...string) string {
+	tk := NewTestKubeconfig(framework.TestContext.CertDir, framework.TestContext.Host, framework.TestContext.KubeConfig, framework.TestContext.KubeContext, framework.TestContext.KubectlPath, namespace)
```
alejandrox1, line 152: I think it'd be better to pass this `tk` struct as an argument.
Otherwise we have a dependency in the test framework that is not explicit.

Think it would also be good to refactor how the `NewKubectlCommand` depends on the `tk`s to not have to pass `tk`s are method arguments so much but we could leave that for a follow up PR.

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,352 @@
...
+	return NewKubectlCommand(tk, args...).ExecOrDie(namespace)
+}
+
+// RunKubectl is a convenience wrapper over Builder
+func RunKubectl(namespace string, args ...string) (string, error) {
+	tk := NewTestKubeconfig(framework.TestContext.CertDir, framework.TestContext.Host, framework.TestContext.KubeConfig, framework.TestContext.KubeContext, framework.TestContext.KubectlPath, namespace)
```
alejandrox1, line 158: Same as prev comment: I think it'd be best to not hide the dependency on the test framework and instead pass the `tk`'s as an argument to the method.
Same for all the other RunKubectl* methods in this file.

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,352 @@
...
+func RunKubemciWithKubeconfig(kubeconfig string, projectID string, args ...string) (string, error) {
+	if kubeconfig != "" {
+		args = append(args, "--"+clientcmd.RecommendedConfigPathFlag+"="+kubeconfig)
+	}
+	return RunKubemciCmd(projectID, args...)
+}
+
+// RunKubemciCmd is a convenience wrapper over Builder to run kubemci.
+// It assumes that kubemci exists in PATH.
+func RunKubemciCmd(projectID string, args ...string) (string, error) {
```
alejandrox1, line 191: This is a nice catch with the projectID!

test/e2e/framework/kubectl/kubectl_builder.go
```diff
@@ -0,0 +1,352 @@
...
+		framework.Logf("\nLast 100 log lines of %v:\n%v", s.Name, l)
+	}
+}
+
+// lookForString looks for the given string in the output of fn, repeatedly calling fn until
+// the timeout is reached or the string is found. Returns last log and possibly
+// error if the string was not found.
+// TODO(alejandrox1): move to pod/ subpkg once kubectl methods are refactored.
+func lookForString(expectedString string, timeout time.Duration, fn func() string) (result string, err error) {
+	for t := time.Now(); time.Since(t) < timeout; time.Sleep(framework.Poll) {
```
alejandrox1, line 225: let's define `framework.Poll`  within the package instead of importing it - reduce the amount of code e2ekubectl depends on the framework for.

test/e2e/apimachinery/webhook.go
```diff
@@ -1201,7 +1202,8 @@ func testAttachingPodWebhook(f *framework.Framework) {
 	ginkgo.By("'kubectl attach' the pod, should be denied by the webhook")
 	timer := time.NewTimer(30 * time.Second)
 	defer timer.Stop()
-	_, err = framework.NewKubectlCommand(f.Namespace.Name, "attach", fmt.Sprintf("--namespace=%v", f.Namespace.Name), pod.Name, "-i", "-c=container1").WithTimeout(timer.C).Exec()
+	tk := e2ekubectl.NewTestKubeconfig(framework.TestContext.CertDir, framework.TestContext.Host, framework.TestContext.KubeConfig, framework.TestContext.KubeContext, framework.TestContext.KubectlPath, f.Namespace.Name)
+	_, err = e2ekubectl.NewKubectlCommand(tk, "attach", fmt.Sprintf("--namespace=%v", f.Namespace.Name), pod.Name, "-i", "-c=container1").WithTimeout(timer.C).Exec()
```
alejandrox1, line 1206: just leaving this as a thought (i dont think we should block this PR on this comment) but it may be good to give some thought on the spec for e2ebuectl.
There is a good deal of repetition and tons of verbosity because of how data is retrieved from the framework's test context and some ambiguity on where and how it is actually needed.
Another thing is that these `tk` structs are passed as method arguments when it may be better to associate the methods to these structs directly.

alejandrox1 (commented):
@ZhiFeng1993 thankyou so much for working on this!
And please excuse the delay.
Though on the plus side, the master branch is now open for kube 1.20 and we can merge things again.
Left some general comments for some minor clean up but overall looking good.

alejandrox1:
@ZhiFeng1993 friendly ping.
will you be able to continue with this?

ZhiFeng1993:
@alejandrox1 Hi Jorge. Yes I'll keep working on this. I have been super busy in the past month but I'll get back to this work once I'm done!

ZhiFeng1993:
@alejandrox1 hi Jorge, sorry about the delay- I just finished rebasing this PR. Please take a look when you have time.

PR #93216, opened 2020-07-19
Deflake PreferredVersion e2e test

### LINKED ISSUE #93215: [flaky] Discovery should validate PreferredVersion for each APIGroup
**Which jobs are flaking**:

pull-kubernetes-e2e-gce

Discovery should validate PreferredVersion for each APIGroup

**Testgrid link**:

https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-e2e-gce&include-filter-by-regex=Discovery%20should%20validate%20PreferredVersion%20for%20each%20APIGroup

**Reason for failure**:

The test does not handle API groups which are added or removed dynamically during the run of the test.

Just because an API group is present in `/apis` at the beginning of the test does not mean `/apis/<group>` or `/apis/<group>/<version>` will be present by the end of the test.

This is particularly true for the dynamic/custom-resource API groups registered/unregistered during the e2e test run

/sig api-machinery
/assign @Riaankl

liggitt:
**What type of PR is this?**
/kind flake

Ignores example.com API groups known to be added/removed during the e2e test run.

Fixes #93215

```release-note
NONE
```

/cc @Riaankl

### DIFF
--- a/test/e2e/apimachinery/discovery.go
+++ b/test/e2e/apimachinery/discovery.go
@@ -18,6 +18,8 @@ package apimachinery
 
 import (
 	"context"
+	"strings"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilversion "k8s.io/apimachinery/pkg/util/version"
 	"k8s.io/apiserver/pkg/endpoints/discovery"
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
/lgtm

PR #75215, opened 2019-03-08
Error when users set unsupported PodSecurityContext fields on Windows

benmoss:
**What type of PR is this?**
> /kind bug

**What this PR does / why we need it**:
Right now Windows pods will happily ignore any invalid PodSecurityContext settings a user attempts to apply. This seems like an area ripe for confusion as users begin using Windows containers on Kubernetes and have to discover that it is silently ignoring parts of their Pod spec.

**Special notes for your reviewer**:
As documented in the Windows KEP at https://github.com/kubernetes/enhancements/blob/07f1a3a634af992217a048189276409849790106/keps/sig-windows/20190103-windows-node-support.md#v1podsecuritycontext

**Does this PR introduce a user-facing change?**:
```release-note
Windows pods will error if they use unsupported PodSecurityContext fields
```

/sig windows
cc @PatrickLang @michmike 

Fixes #75132

PR #93220, opened 2020-07-19
ingore apparmor on windows

### LINKED ISSUE #93165: A PSP with apparmor annotations does not allow windows pods to be scheduled
**What happened**:
If I have the a psp with apparmor annotations:

```
apiVersion: {{if IsKubernetesVersionGe "1.16.0"}}policy{{else}}extensions{{end}}/v1beta1
kind: PodSecurityPolicy
metadata:
  name: restricted
  annotations:
    seccomp.security.alpha.kubernetes.io/allowedProfileNames: docker/default
    apparmor.security.beta.kubernetes.io/allowedProfileNames: runtime/default
    seccomp.security.alpha.kubernetes.io/defaultProfileName:  docker/default
    apparmor.security.beta.kubernetes.io/defaultProfileName:  runtime/default
  labels:
    addonmanager.kubernetes.io/mode: Reconcile
#truncated...
```

and I deploy a windows pod it gets blocked:

```
kubectl create -f- <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
  - name: test
    image: e2eteam/busybox:1.29
    command:
      - sleep
      - "3600"
  restartPolicy: Always
  nodeSelector:
    "kubernetes.io/os": windows
EOF
pod "test" created
```

```
kubectl get pods                    
NAME                 READY   STATUS                       RESTARTS   AGE
test                 0/1     Blocked                      0          9m55s

kubectl describe po test
Name:         test                                                                                                                                         
Namespace:    default                                                                                                                          
Priority:     0                                                                                                                                            
Node:         1309k8s01000000/10.240.0.65                                                                                                                  
Start Time:   Thu, 16 Jul 2020 16:39:52 -0700                                                                                                              
Labels:       <none>                                                                                                                                       
Annotations:  container.apparmor.security.beta.kubernetes.io/test: runtime/default                                                                         
              kubernetes.io/psp: restricted                                                                                                                
              seccomp.security.alpha.kubernetes.io/pod: docker/default                                                                                     
Status:       Pending                                                                                                                                      
Reason:       AppArmor                                                                                                                                     
Message:      Cannot enforce AppArmor: binary not compiled for linux    
```
Apparmor check is ignored for windows pods

see above

Should be fixed/documented with work around as per #64801 

**Environment**: 
kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.5", GitCommit:"e6503f8d8f769ace2f338794c914a96fc335df0f", GitTreeState:"clean", BuildDate:"2020-06-26T03:47:41Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.5", GitCommit:"e6503f8d8f769ace2f338794c914a96fc335df0f", GitTreeState:"clean", BuildDate:"2020-06-26T19:51:26Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}

- OS (e.g: `cat /etc/os-release`): windows/unbuntu
- Kernel (e.g. `uname -a`):
- Install tools: aks-engine
- Others:

wawa0210:
**What type of PR is this?**

/kind bug

AppArmor is a Linux kernel security module.and It does not support windows

At present, the startup logic of kubelet does not deal with this logic, which causes apparmor to be running on windows, causing an exception

Fixes #93165

AppArmor is a linux kernel security module  and it does not support other operating systems.

```release-note
Ignore apparmor for non-linux operating systems
```

```docs

```

/sig windows

### DIFF
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -25,6 +25,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	sysruntime "runtime"
 	"sort"
 	"strings"
 	"sync"
@@ -784,8 +785,11 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 		opt(klet)
 	}
 
-	klet.appArmorValidator = apparmor.NewValidator(containerRuntime)
-	klet.softAdmitHandlers.AddPodAdmitHandler(lifecycle.NewAppArmorAdmitHandler(klet.appArmorValidator))
+	if sysruntime.GOOS == "linux" {
+		// AppArmor is a Linux kernel security module and it does not support other operating systems.
+		klet.appArmorValidator = apparmor.NewValidator(containerRuntime)
+		klet.softAdmitHandlers.AddPodAdmitHandler(lifecycle.NewAppArmorAdmitHandler(klet.appArmorValidator))
+	}
 	klet.softAdmitHandlers.AddPodAdmitHandler(lifecycle.NewNoNewPrivsAdmitHandler(klet.containerRuntime))
 	klet.softAdmitHandlers.AddPodAdmitHandler(lifecycle.NewProcMountAdmitHandler(klet.containerRuntime))
 


mattjmcnaughton (commented):
Thanks for taking a look at this issue @wawa0210 !

To start, could we talk through some of the diff options for solving this issue?

One option, which you've already implemented, definitely seems to be gating the `appArmorValidator` on the runtime, and not starting if its windows. However, I'm curious about how this will work on non-Windows/non-Linux operating systems (i.e. FreeBSD and Darwin), which I imagine also doesn't support app armor.

However, I think there may be other options as well - off the top of my head, I think there are at least three.

1) Do some kind of runtime check around whether the system running the `kubelet` supports app armor, and only add the admit handler if it does.

2) Make it so that, if the app armor validator is attempting to run on a non-supported platform, the validator itself is smart enough to fail gracefully.

3) As you mentioned, break out this runtime specific component into a separate file using build tags.

We may find that, after exploring all options, the initially proposed one is the best. Still, I think it would be useful to think through the pros and cons of this option :)

wawa0210:
Thank you so much for such a detailed comment

Before I start, I also thought about different ways to implement it, as you said

> 2. Make it so that, if the app armor validator is attempting to run on a non-supported platform, the validator itself is smart enough to fail gracefully.

I don't recommend using this method, it will judge it like an armor validator at the specific logic level, and then to fail gracefully. But in the future, there will be many features like this, so that the same judgment will be written and executed

> 3. As you mentioned, break out this runtime specific component into a separate file using build tags.

I also want to implement it in this way, but it seems that the current logic is only here. Therefore, I tried to use the first relatively simple way. If there are such features later, maybe it is a better choice to split and refactor.

liggitt:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/93220/pull-kubernetes-node-e2e/1284999243387375616 flake is #93210

dashpole:
This needs guidance from sig-windows.  @benmoss @ddebroy

wawa0210:
@jsturtevant @michmike 
Can you take the time to help review this pr?

pkg/kubelet/kubelet.go
```diff
@@ -781,8 +782,13 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 		opt(klet)
 	}
 
-	klet.appArmorValidator = apparmor.NewValidator(containerRuntime)
-	klet.softAdmitHandlers.AddPodAdmitHandler(lifecycle.NewAppArmorAdmitHandler(klet.appArmorValidator))
+	if sysruntime.GOOS == "linux" {
+		// AppArmor is the linux kernel security module, and does not support non linux
```
neolit123, line 786:
```suggestion
		// AppArmor is a Linux kernel security module and does not support other operating systems.
```

pkg/kubelet/kubelet.go
```diff
@@ -781,8 +782,13 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
...
+		// At present, kubelet.go rarely pays attention to the differences of os. At present, it is recommended to judge by `runtime.GOOS`,
+		// and then if there are more features, it is recommended to separate them into different files for processing.
```
neolit123, line 788: i think we can omit these comments for now.

looking at the kubelet package, there is split between using runtime checks and compile time checks (_someos.go files).
but the decisions here is to the kubelet maintainers.

pkg/kubelet/kubelet.go
wawa0210: done~

pkg/kubelet/kubelet.go
wawa0210: Thank you for your comment. It is a better way to split into different files according to the runtime~ But the logic is relatively simple at present, and it is currently the main entry file of kubelet, so I chose a conservative solution.

Hope to get suggestions from kubelet maintainers

neolit123:
Lgtm

jsturtevant:
This is another place where a warning would be appropriate as discussed in https://github.com/kubernetes/kubernetes/pull/92355#pullrequestreview-486972623 for https://github.com/kubernetes/kubernetes/issues/94626

marosset:
LGTM

michmike (approved):
/lgtm
/approve

dims:
/assign @dashpole @sjenning 

Quick easy one David and Seth. adds a if condition around the admit handler for apparmor

/approve
/lgtm

dashpole:
It feels odd to me that this behavior differs from the behavior we get when trying to run a pod on a [linux node without apparmor](https://github.com/kubernetes/kubernetes/blob/ea5ca74195cbbeb4acefac966cf50edfc5ce1760/pkg/security/apparmor/validate.go#L110).  What do users do today when they have clusters with some nodes that do/don't support AppArmor?  Or do we currently assume AppArmor is either enabled or disabled on all nodes?  What makes windows special in this regard?

I have the same discomfort with this as I had with https://github.com/kubernetes/kubernetes/pull/92355#discussion_r487172260.  The current behavior is that if you ask for AppArmor on your pod, and the node cannot satisfy it, we don't run your pod.  That seems like the most secure behavior, so a misconfigured node or being scheduled to a node you didn't intend doesn't silently result in a security incident.  A warning when creating the object is better, but it sounds like [multi-arch images mean we can't do validation at that point](https://github.com/kubernetes/kubernetes/issues/94626#issuecomment-693027610).

Why isn't the right answer here: "Don't apply an AppArmor SP to your pod if you want to run without AppArmor"?

@tallclair @mikedanese

jsturtevant:
>   Or do we currently assume AppArmor is either enabled or disabled on all nodes? What makes windows special in this regard?

specifically for AppArmor, it would never be enabled on Windows as it is a Linux kernel component.  I don't think there is a misconfigured node scenario in the case where `goos == windows`

> Why isn't the right answer here: "Don't apply an AppArmor SP to your pod if you want to run without AppArmor"?

Important question. It might not be the ccorrect answer but from the other discussions there was a an expectation from windows users that since it is not valid for Windows it wouldn't come into play.

For comparison, I ran a pod with SecurityContext.WindowsOptions and SecurityContext.seLinuxOptions set on a linux node (which didn't have seLinux enabled) and it was scheduled.  Maybe there needs to be another look at this more holistically?

jsturtevant:
thinking on it some more, the multi-arch use case might be the scenario that was the impetus for https://github.com/kubernetes/kubernetes/pull/93475.

@ravisantoshgudimetla could you share more info on your expectations?

benmoss:
This discussion reminds me of https://github.com/kubernetes/kubernetes/pull/75215

> @liggitt: we have existing fields that get ignored in environments where they don't apply (selinux, apparmor/seccomp annotations, etc), and the windows GA behavior in 1.14 is to ignore linux-specific fields, correct? don't the changes in this PR mean that pod specs that could run correctly in 1.14 will now fail?

ravisantoshgudimetla:
> thinking on it some more, the multi-arch use case might be the scenario that was the impetus for #93475.
> 
> @ravisantoshgudimetla could you share more info on your expectations?

To me, ignoring the linux specific security constructs on kubelet side makes sense for now. There are distributions of Kube like OpenShift that apply SELinux options to the pods at the admission and this was the reason for starting #93475. I think, the fundamental problem is we want to distinguish Linux from Windows pods at the scheduling time using nodeSelector and tolerations but we have some api within the pod Spec that can be defaulted at the admission time. Unless we enforce runtime class and have reserved label which says `I am Windows pod` and have that applied to the pod Spec at the admission time, it becomes so hard for us to distinguish and we might have to end up special casing on the kubelet or runtime side as these security specific fields are passed down to them.

wawa0210:
@liggitt 

Looking forward to your comments, thanks

liggitt:
> @liggitt
> 
> Looking forward to your comments, thanks

sorry, didn't realize this was pending my comments... what section specifically?

wawa0210:
> sorry, didn't realize this was pending my comments... what section specifically?

sig-windows agrees that this pr can solve the current scenario where apparmor is not applicable to windows~ but @dashpole seems to have some different suggestions

liggitt:
> I have the same discomfort with this as I had with [#92355 (comment)](https://github.com/kubernetes/kubernetes/pull/92355#discussion_r487172260). The current behavior is that if you ask for AppArmor on your pod, and the node cannot satisfy it, we don't run your pod. That seems like the most secure behavior, so a misconfigured node or being scheduled to a node you didn't intend doesn't silently result in a security incident.

That seems like a reasonable position for non-apparmor linux nodes which could happily unsafely run a linux container you would have thought would be protected by apparmor.

If the container is a multi-arch container that supports windows, I don't think there will be confusion that the windows container runs without apparmor

If we were structuring the pod spec today, we'd likely group things like apparmor, selinux, and seccomp options in a `linuxOptions` substanza of security context. Imagining those linux-only things grouped under a linuxOptions stanza, I don't find it strange for the windows kubelet to ignore them.

dashpole:
I guess it's just me then.  I don't have any powers here anymore, though.
/approve

dims:
/assign @sjenning @derekwaynecarr 

easy PR, already approved by @dashpole :)

Issue #93145, opened 2020-07-16
Support the pod used nodeName schedule, volume controller can binding pvc to pod.


mikechengwei:
**What would you like to be added**:
Support the pod used nodeName schedule, volume controller can binding pvc to pod.
**Why is this needed**:
Now, when we use the nodeName field to run the Pod on a specific node, scheduler will skip update. This causes, the Volume controller to never start binding of the PVC. 
The log like below:
```
I0503 16:12:10.049967       1 event.go:278] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"default", Name:"demo-local-hostpath-busybox-sts-local-hostpath-0", UID:"7f06a345-8551-4f07-bafa-9cad87ace6d1", APIVersion:"v1", ResourceVersion:"2914960", FieldPath:""}): type: 'Normal' reason: 'WaitForFirstConsumer' waiting for first consumer to be created before binding
I0503 16:12:25.050080       1 event.go:278] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"default", Name:"demo-local-hostpath-busybox-sts-local-hostpath-0", UID:"7f06a345-8551-4f07-bafa-9cad87ace6d1", APIVersion:"v1", ResourceVersion:"2914960", FieldPath:""}): type: 'Normal' reason: 'WaitForFirstConsumer' waiting for first consumer to be created before binding
```
The pod binding pvc resources will not start forever.

mikechengwei:
Currently we use deployment with specified pvc to deploy tidbmonitor. Can we change to use statefulset object?  Monitoring is about storage-related services.

mikechengwei:
Hi, can you assign me to do it ? @zhouya0

zhouya0:
Yes, of course. :)
/unassign

mikechengwei:
Thanks, I don’t know if this can be done, so I ask a feature request.

mwennrich:
I think this is related to https://github.com/kubernetes/kubernetes/issues/81725

ahg-g:
pls use node affinity instead of node name. The scheduler doesn't execute the scheduling cycle on pods with `pod.Spec.NodeName` set, those pods are considered "scheduled" already.

mikechengwei:
ok.

Issue #89953, opened 2020-04-08
PVC is not annotated with "volume.kubernetes.io/selected-node" when .spec.nodeName is used


ialidzhikov:
**What happened**:
PVC stays forever in `Pending` state when the Pod that uses the PVC is created with `.spec.nodeName`.

**What you expected to happen**:
I would expect to be able to schedule a Pod on a predefined Node (using `.spec.nodeName`) that use a PVC (kubelet running with `--cloud-provider=external` and CSI migration feature gates are enabled on the corresponding components).

**How to reproduce it (as minimally and precisely as possible)**:
1. The default storage class is the following one:
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
  name: default
provisioner: ebs.csi.aws.com
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer
```

2. Apply a PVC
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: ebs-claim
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 4Gi
```

3. Apply a Pod with `.spec.nodeName` that uses this PVC
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: app
spec:
  containers:
  - name: app
    image: centos
    command: ["/bin/sh"]
    args: ["-c", "while true; do echo $(date -u) >> /data/out.txt; sleep 5; done"]
    volumeMounts:
    - name: persistent-storage
      mountPath: /data
  volumes:
  - name: persistent-storage
    persistentVolumeClaim:
      claimName: ebs-claim
  nodeName: ip-10-222-0-45.eu-west-1.compute.internal
```

5. Ensure that the PVC does not have the `volume.kubernetes.io/selected-node` annotation. 
The Pod stays in `ContainerCreating` with events:
```
$ k describe po app

  Warning  FailedMount  3m40s (x300 over 93m)  kubelet, ip-10-222-0-45.eu-west-1.compute.internal  Unable to attach or mount volumes: unmounted volumes=[persistent-storage], unattached volumes=[persistent-storage default-token-65nsk]: error processing PVC default/ebs-claim-coreos: PVC is not bound
```

The PVC events
```
$ k describe pvc ebs-claim

Events:
  Type    Reason                Age                   From                         Message
  ----    ------                ----                  ----                         -------
[...]

PR #93227, opened 2020-07-19
Add the pod with the specified nodename to the schedule life cycle.

mikechengwei:
**What type of PR is this?**
/kind feature

Now pod specify `nodename` field will skip schedule phase , it will cause the volume binding to not be executed and the pod failed to start. 
I think we shouldn't skip plugins running phase, we can give the feasibleNodes directly at `findNodesThatFitPod` method by `nodename` field:

https://github.com/kubernetes/kubernetes/blob/363c3b89f570142da5c43b1ccbd99629f03e0ee0/pkg/scheduler/core/generic_scheduler.go#L259

I think this is a friendly method to incorporate the specified nodename schedule into the current schedule life cycle.

#93145 

#89953

My design is giving the feasibleNodes directly  in findNodesThatFitPod method  when we specify `nodename` field.

I think although we specified `nodename`, we can't skip the predicate stage, checking node info by `nodename` field   is also required. 

Therefore, the pod started by the specified node name can also be added to the schedule life cycle.

None

None

### DIFF
--- a/pkg/scheduler/core/generic_scheduler.go
+++ b/pkg/scheduler/core/generic_scheduler.go
@@ -280,7 +280,7 @@ func (g *genericScheduler) findNodesThatFitPod(ctx context.Context, prof *profil
 
 	if len(pod.Spec.NodeName) > 0 {
 		nodeInfo, err := g.nodeInfoSnapshot.NodeInfos().Get(pod.Spec.NodeName)
-		if err != nil {
+		if err != nil || nodeInfo == nil {
 			return nil, filteredNodesStatuses, err
 		}
 		return []*v1.Node{nodeInfo.Node()}, filteredNodesStatuses, nil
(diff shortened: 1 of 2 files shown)


mikechengwei:
@liu-cong  Do  you have time to  help me review this pr?

ahg-g:
Unfortunately this PR is not correct. The scheduler by design doesn't insert pods with `pod.Spec.NodeName` set into the queue, those are considered scheduled pods. Use node affinity rather than setting the `NodeName` directly If you want the scheduler to handle the pod.

zhouya0:
@ahg-g I saw there's `nodeName` plugin in filter plugin, like you said, is it really needed? 
https://github.com/kubernetes/kubernetes/blob/1fdd8fb213e0361e8f18b1dd152dddb4c88ad183/pkg/scheduler/algorithmprovider/registry.go#L98

ahg-g:
> @ahg-g I saw there's `nodeName` plugin in filter plugin, like you said, is it really needed?
> https://github.com/kubernetes/kubernetes/blob/1fdd8fb213e0361e8f18b1dd152dddb4c88ad183/pkg/scheduler/algorithmprovider/registry.go#L98

This plugin is a re-write of an old predicte. It is likely not needed, but I would rather keep it in case it addressed race conditions related to updating the NodeName by another controller while the pod is being scheduled.

PR #93222, opened 2020-07-19
Cleanup: fix linting failures for pkg/{kubectl, kubelet} and related packages

ESWZY:
**What type of PR is this?**

/kind cleanup

**What this PR does?**

Fix golint failures for package `pkg/kubectl`, `pkg/kubelet` and related packages.

**Which issue this PR fixes?**

Ref: #68026 [Fix golint failures](https://github.com/kubernetes/kubernetes/issues/68026)

**Special notes for your reviewer**:

Following is the failure kinds I fixed. Because I don't have the right to interpret the code, I skiped some comment and type name golint failures. If you have some good idea, please tell me. Glad to help. 🥰

Additionally, you can inspect the changes of sub-package by corresponding commit.

- Linting failures fixed:

```
XXX.go:X:2: struct field CpuXxx should be CPUXxx
XXX.go:X:3: var xxxJson should be xxxJSON
XXX.go:X:8: const xxxId should be xxxID
xxx_xxx.go:X:1: don't use an underscore in package name
XXX.go:X:2: don't use underscores in Go names; var xxx_xxx should be xxxXxx
XXX.go:X:30: error strings should not be capitalized or end with punctuation or a newline
```

- Linting failures partly fixed:

```
XXX.go:X:2: comment on exported XXX should be of the form "XXX ..."
XXX.go:X:1: exported XXX should have comment or be unexported
XXX.go:X:6: type name will be used as xxx.XxxID by other packages, and that stutters; consider calling this ID
XXX.go:X:1: package comment should be of the form "Package types ..."
XXX.go:295:9: if block ends with a return statement, so drop this else and outdent its block
```

- Linting failures not fixed:

```
XXX.go:X:42: exported func XXX returns unexported type *XX.xXX which can be annoying to use
pkg\kubelet\dockershim\network\plugins.go:49:2: don't use ALL_CAPS in Go names; use CamelCase
XXX.go:X:1: package comment is detached; there should be no blank lines between it and the package statement
```

The involved sub packages are:

- Packages removed from [.golint_failures list](https://github.com/kubernetes/kubernetes/blob/master/hack/.golint_failures):

```
pkg/kubectl/cmd/convert	
pkg/kubelet/apis/config/v1beta1
pkg/kubelet/cri/remote
pkg/kubelet/dockershim/network/hostport/testing	
pkg/kubelet/dockershim/network/kubenet
```

- Packages modified:

```
pkg/kubelet/cm
pkg/kubelet/container/testing
[...]

PR #93231, opened 2020-07-19
Cleanup: fix small golint failures in pkg/kubelet

ESWZY:
**What type of PR is this?**

/kind cleanup

Fix *small* golint failures under package `pkg/kubelet`.

Ref: #68026 [Fix golint failures](https://github.com/kubernetes/kubernetes/issues/68026)

Break up: #93222 [Cleanup: fix linting failures](https://github.com/kubernetes/kubernetes/pull/93222)

The golint failures it fixed mostly related with comment.

To break up another PR, I reverted some of commits. Sorry for a long commit log.

Thanks for your review! If you have any reply, I will respond as soon as possible!


ESWZY:
/assign @vishh

Thanks very much!

SergeyKanzhelev:
please squash commits

pkg/kubelet/container/testing/fake_runtime_helper.go
```diff
@@ -53,7 +53,7 @@ func (f *FakeRuntimeHelper) GetPodDNS(pod *v1.Pod) (*runtimeapi.DNSConfig, error
 		Options:  f.DNSOptions}, f.Err
 }
 
-// This is not used by docker runtime.
+// GeneratePodHostNameAndDomain is not used by docker runtime.
```
SergeyKanzhelev, line 56: please add a description before this note. Something like

```suggestion
// GeneratePodHostNameAndDomain on fake runtime helper returns fake hostname and domain. It is not used by docker runtime.
```

SergeyKanzhelev:
Just realized, this PR changes API method name: `SetDefaults_KubeletConfiguration` -> `SetDefaultsKubeletConfiguration`. That might be worth a release note. Not sure here.

pkg/kubelet/container/testing/fake_runtime_helper.go
ESWZY: Thanks for your review! 

I changed it and also added description for other methods.

SergeyKanzhelev:
/lgtm
/assign @dashpole 
/assign @dims 

(@dims for /hack file)

@dashpole do you think the name change in API  deserve the release note?

dashpole:
Looks like the answer is "no":

```
Unexpected dirty working directory:

  diff --git a/pkg/kubelet/apis/config/v1beta1/zz_generated.defaults.go b/pkg/kubelet/apis/config/v1beta1/zz_generated.defaults.go
  index 320d448931f..73e63fc114d 100644
  --- a/pkg/kubelet/apis/config/v1beta1/zz_generated.defaults.go
  +++ b/pkg/kubelet/apis/config/v1beta1/zz_generated.defaults.go
  @@ -22,17 +22,11 @@ package v1beta1
   
   import (
   	runtime "k8s.io/apimachinery/pkg/runtime"
  -	v1beta1 "k8s.io/kubelet/config/v1beta1"
   )
   
   // RegisterDefaults adds defaulters functions to the given scheme.
   // Public to allow building arbitrary schemes.
   // All generated defaulters are covering - they call all nested defaulters.
   func RegisterDefaults(scheme *runtime.Scheme) error {
  -	scheme.AddTypeDefaultingFunc(&v1beta1.KubeletConfiguration{}, func(obj interface{}) { SetObjectDefaultsKubeletConfiguration(obj.(*v1beta1.KubeletConfiguration)) })
   	return nil
   }
  -
  -func SetObjectDefaultsKubeletConfiguration(in *v1beta1.KubeletConfiguration) {
  -	SetDefaultsKubeletConfiguration(in)
  -}
!!! [0721 14:06:10] Call tree:
!!! [0721 14:06:10]  1: hack/make-rules/../../hack/verify-generated-files.sh:29 kube::util::ensure_clean_working_dir(...)
```

dashpole:
you will probably have to revert changes in pkg/kubelet/apis/config/v1beta1

ESWZY:
> you will probably have to revert changes in pkg/kubelet/apis/config/v1beta1

Thanks for your correction! 

I found this problem, too. But if I don't change this file, it will not pass the ci tests. (more failed tests)

pkg/kubelet/apis/config/v1beta1/defaults.go
```diff
@@ -30,16 +30,16 @@ import (
 	utilpointer "k8s.io/utils/pointer"
 )
 
+// TODO: Move these constants to k8s.io/kubelet/config/v1beta1 instead?
```
kikisdeliveryservice, line 33: If you end up including this TODO, there should be a bit more context ie the "why" so people can quickly understand. Though this might be better tracked in it's own issue as well.

pkg/kubelet/apis/config/v1beta1/defaults.go
```diff
@@ -30,16 +30,16 @@ import (
...
 const (
-	// TODO: Move these constants to k8s.io/kubelet/config/v1beta1 instead?
 	DefaultIPTablesMasqueradeBit = 14
 	DefaultIPTablesDropBit       = 15
 	DefaultVolumePluginDir       = "/usr/libexec/kubernetes/kubelet-plugins/volume/exec/"
 )
 
+// TODO: Move these constants to k8s.io/kubelet/config/v1beta1 instead?
```
kikisdeliveryservice, line 40: Same as above

pkg/kubelet/types/constants.go
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 package types
 
 const (
-	// system default DNS resolver configuration
+	// ResolvConfDefault is system default DNS resolver configuration
```
kikisdeliveryservice, line 20: "is the"

pkg/kubelet/apis/config/v1beta1/defaults.go
ESWZY: Those TODOs are code left by the original developer. Maybe he/she thinks that is a better format.

pkg/kubelet/types/constants.go
ESWZY: Thanks! Fixed.

ESWZY:
Updated commit message only.

kikisdeliveryservice (commented):
> Updated commit message only.

A better format for the commit message (using what you already have):

```
Cleanup: fix small golint failures in pkg/kubelet

Fix golint failures under pkg/kubelet including api\config\v1beta1,
container\testing, pod\testing, premption and types

```

Please see: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#7-commit-message-guidelines for further commit message guidelines.

pkg/kubelet/apis/config/v1beta1/defaults.go
kikisdeliveryservice: Ah this diff confused me

ESWZY:
> > Updated commit message only.
> 
> A better format for the commit message (using what you already have):
> 
> ```
> Cleanup: fix small golint failures in pkg/kubelet
> 
> Fix golint failures under pkg/kubelet including api\config\v1beta1,
> container\testing, pod\testing, premption and types
> ```
> 
> Please see: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#7-commit-message-guidelines for further commit message guidelines.

Thanks! Got it!

SergeyKanzhelev:
@ESWZY If my understanding of the state of this PR is correct - I'd suggest to fix the issues without the change of `.hack` file. So build will pass even with generated files having issues. Later separate PR can fix the hack file.

Do you plan to move it forward?

ESWZY:
> @ESWZY If my understanding of the state of this PR is correct - I'd suggest to fix the issues without the change of `.hack` file. So build will pass even with generated files having issues. Later separate PR can fix the hack file.
> 
> Do you plan to move it forward?

Hello, thanks for your advice! I have withdrawn the change.

SergeyKanzhelev:
> > @ESWZY If my understanding of the state of this PR is correct - I'd suggest to fix the issues without the change of `.hack` file. So build will pass even with generated files having issues. Later separate PR can fix the hack file.
> > Do you plan to move it forward?
> 
> Hello, thanks for your advice! I have withdrawn the change.

my understanding you still need to undo changes in generated files

ESWZY:
> > > @ESWZY If my understanding of the state of this PR is correct - I'd suggest to fix the issues without the change of `.hack` file. So build will pass even with generated files having issues. Later separate PR can fix the hack file.
> > > Do you plan to move it forward?
> > 
> > 
> > Hello, thanks for your advice! I have withdrawn the change.
> 
> my understanding you still need to undo changes in generated files

I appreciate your patient reviews! 

Undone.

SergeyKanzhelev:
/close

please re-open if you plan to work on it. For generated files - they should be modified using `make update` and be part of the PR.

Also some of the fixes are a part of another PR: https://github.com/kubernetes/kubernetes/pull/97374

PR #93218, opened 2020-07-19
[1.17] Update repo-infra to v0.0.8 (to support go1.14.6 and go1.13.14)

justaugustus:
**What type of PR is this?**

/kind cleanup
/area dependency release-eng
/sig testing release

**What this PR does / why we need it**:

https://github.com/kubernetes/repo-infra/releases/tag/v0.0.8 pulls in bazel toolchain updates to support go1.14.6 and go1.13.14.

Needed for the in-flight Golang updates: https://github.com/kubernetes/release/issues/1410

/assign @fejta @BenTheElder @liggitt 
cc: @kubernetes/release-engineering 

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

Cherry pick of https://github.com/kubernetes/kubernetes/pull/91348.
/hold for merge of ^^

**Does this PR introduce a user-facing change?**:

```release-note
- Update bazel to 2.2.0
- Update repo-infra to 0.0.8 (to support go1.14.6 and go1.13.14)
  - Includes:
    - bazelbuild/bazel-toolchains@3.4.0
    - bazelbuild/rules_go@v0.22.8
```

PR #92362, opened 2020-06-21
Add bazel_skylib_workspace to fix make bazel-test 'no matching toolchains found' error

brianpursley:
**What type of PR is this?**
/kind failing-test

**What this PR does / why we need it**:
Fixes the following error that occurs when running `make bazel-test`:
```
ERROR: While resolving toolchains for target //build:code_generation_tests_test_1: no matching toolchains found for types @bazel_skylib//toolchains/unittest:toolchain_type
ERROR: Analysis of target '//build:code_generation_tests_test_1' failed; build aborted: no matching toolchains found for types @bazel_skylib//toolchains/unittest:toolchain_type
INFO: Elapsed time: 2.645s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (2 packages loaded, 4 targets configured)
FAILED: Build did NOT complete successfully (2 packages loaded, 4 targets configured)
make: *** [Makefile:594: bazel-test] Error 1
```

**Which issue(s) this PR fixes**:
Fixes # https://github.com/kubernetes/kubernetes/issues/90739

**Special notes for your reviewer**:
I’m not very familiar with bazel, but according to this link, the lines in this PR need to be added to the workspace:  
https://github.com/bazelbuild/bazel-skylib#troubleshooting

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:
```docs

```

PR #91312, opened 2020-05-20
build: Update to repo-infra@v0.0.5 to support go1.14.3 and go1.13.11

justaugustus:
**What type of PR is this?**

/kind cleanup
/area dependency release-eng
/sig testing release

**What this PR does / why we need it**:

https://github.com/kubernetes/repo-infra/releases/tag/v0.0.5 pulls in bazel toolchain updates to support go1.14.3 and go1.13.11.

Needed for the in-flight Golang updates:
- go1.14.3: https://github.com/kubernetes/release/issues/1216 --> https://github.com/kubernetes/kubernetes/pull/88638
- go1.13.11: https://github.com/kubernetes/release/issues/1257

/assign @fejta @BenTheElder 
cc: @markyjackson-taulia @Verolop @kubernetes/release-engineering 

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #90347, opened 2020-04-21
Update to repo-infra v0.0.4

fejta:
And use repo-infra to configure rules_go and rbe

/kind cleanup

```release-note
NONE
```

PR #93233, opened 2020-07-19
[1.17] Update Golang to v1.13.14

justaugustus:
**What type of PR is this?**

/kind cleanup
/area dependency release-eng
/sig testing release

Update Golang to v1.13.14.

Tracking issue: https://github.com/kubernetes/release/issues/1410

~Includes the bazel updates from https://github.com/kubernetes/kubernetes/pull/93218.
If this goes green without issue, I'll close the previous PR and update the release note here instead.~

Combined cherry pick of https://github.com/kubernetes/kubernetes/pull/89430, https://github.com/kubernetes/kubernetes/pull/90347, https://github.com/kubernetes/kubernetes/pull/91312, bddd52046d8122af98eab4a172cebb45066a78d1, and https://github.com/kubernetes/kubernetes/pull/92362.

```release-note
- Update Golang to v1.13.14
- Update bazel to 2.2.0
- Update repo-infra to 0.0.8 (to support go1.14.6 and go1.13.14)
  - Includes:
    - bazelbuild/bazel-toolchains@3.4.0
    - bazelbuild/rules_go@v0.22.8
```


justaugustus:
`pull-kubernetes-e2e-gce-device-plugin-gpu` is a known flake.
/skip

justaugustus:
Ready for review.
/assign @liggitt @dims @BenTheElder 

liggitt:
/cc @mikedanese @fejta 
can you ack the bazel changes?

liggitt:
verified this matches master bazel bits, and CI passes

/lgtm
/approve

hasheddan (approved):
/lgtm

MaciekPytel:
Shouldn't go version in go.mod also be updated when bumping it here? For release-1.17 go.mod still lists go1.12.

liggitt:
xref https://github.com/golang/go/issues/30791 - it's sort of unclear what that directive does ... I would not change it on a release branch

MaciekPytel:
In this particular case the problem is that release-1.17 branch doesn't compile using go1.12, so that seems particularly wrong. 

I noticed the issue, because Cluster Autoscaler (which heavily relies on importing k/k) uses go.mod to determine which go version it should use when vendoring a particular k8s version. I can change Cluster Autoscaler tooling to allow a manual override easily enough, however, it seems misleading to list go version that doesn't compile.

fejta:
https://golang.org/ref/mod#tmp_12

Issue #75890, opened 2019-03-29
Mark nodes as "NotReady" until critical pods from daemonsets are Ready


lbernail:
**What would you like to be added**:
An extension to node conditions that would mark them NotReady until critical pods from daemonsets are running (maybe a CriticalPodsReady or DaemonsetsReady similar to the NetworkReady condition).

**Why is this needed**:
When using cluster-autoscaler, application pods are immediately scheduled when a node starts and sometimes they start before local pods from daemonsets are Ready, which can be an issue. A few examples:
- kube-proxy, which prevents application from connecting to their dependencies (DNS is the most common problem)
- local dns pod: we deploy a local dns cache as a daemonset and configure pods to use it as a resolver, so resolution fail until this pod is up
- monitoring agent: some applications send metrics very early when they start and require the monitoring agent to be ready
- kube2iam: required by some applications to get aws credentials

/sig scheduling
@kubernetes/sig-scheduling-feature-requests

Joseph-Irving:
We very recently made a controller to deal with this exact issue. https://github.com/uswitch/nidhogg tainting the node if certain daemonset pods haven't been scheduled and are ready yet.

lbernail:
> We very recently made a controller to deal with this exact issue. https://github.com/uswitch/nidhogg tainting the node if certain daemonset pods haven't been scheduled and are ready yet.

@Joseph-Irving Very nice. I'll definitely have a look. It would be great to have it natively in kubernetes but this would work in the meantime (and if the feature is not accepted of course)

yastij:
cc @kubernetes/sig-scheduling-feature-requests @kubernetes/sig-node-feature-requests 

I think this is something worth exploring. This can be tied to priorityClass: mark nodes with a specific condition until pods of a specific priorityClass are running on the node.

Huang-Wei:
I like the idea. We have [pod ready++](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0007-pod-ready%2B%2B.md), why not node ready++?

yastij:
/assign

I’ll brainstorm with @lbernail and see if can come up with a KEP to kickoff discussions.

cizixs:
We have the same feature requirement, and have come with up a design and implementation running in our kubernetes cluster.

And we would love to share and contribute to this.

yastij:
@cizixs - can you share some thoughts and use cases here ?

kevin-wangzefeng:
[...]

PR #93239, opened 2020-07-20
Preferentially pull the the images of critical pod

saberrey:
**What type of PR is this?**
 
/kind feature

This pr implements a method to preferentially pull images of critical pods based on serial pull strategy( --serialize-image-pulls=true ). To mark a Pod as critical, we can set priorityClassName for that Pod to system-cluster-critical or system-node-critical，and we can also prefer to pull the image of the pod in the specified namespaces. 
When a node fails and pods are scheduled to other nodes, we hope that images of critical pods will be pulled first to ensure that important services can started first.

#75890

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kubelet/images/puller.go
+++ b/pkg/kubelet/images/puller.go
@@ -17,6 +17,10 @@ limitations under the License.
 package images
 
 import (
+	"k8s.io/kubernetes/pkg/apis/scheduling"
+	"k8s.io/kubernetes/pkg/util/env"
+	"sort"
+	"strings"
 	"time"
 
 	"k8s.io/api/core/v1"
@@ -31,7 +35,7 @@ type pullResult struct {
 }
 
 type imagePuller interface {
-	pullImage(kubecontainer.ImageSpec, []v1.Secret, chan<- pullResult, *runtimeapi.PodSandboxConfig)
+	pullImage(kubecontainer.ImageSpec, []v1.Secret, chan<- pullResult, *runtimeapi.PodSandboxConfig, *v1.Pod)
 }
 
 var _, _ imagePuller = &parallelImagePuller{}, &serialImagePuller{}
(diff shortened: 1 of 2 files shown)


odinuge (commented):
Hey @saberrey! Even tho this looks like a nice way of solving the issue, I do think this change will require a [KEP](https://github.com/kubernetes/enhancements/tree/master/keps). It should at least be discussed within sig-node, and then a KEP makes it a lot easier and quicker to get consensus about how things should work! You can also write a post on #sig-node on slack and/or join the sig-node meetings. :smile: 

Adding a hold for now.
/hold

pacoxu:
@saberrey if you have no time to work on the KEP, I would like to create one for this. I think this is a nice improvement for cases with --serialize-image-pulls=true especially during node failures. 

- storage pod may start before those pods which would use pv 
- kube-system pods pull image first
 
______________________________________________

BTW, --serialize-image-pulls default is  true. 
```
Pull images one at a time. We recommend *not* changing the default value on nodes that run docker daemon with version < 1.9 or an `aufs` storage backend. Issue #10959 has more details. (DEPRECATED: This parameter should be set via the config file specified by the Kubelet's `--config` flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.)
```
If --serialize-image-pulls=true, a big image pull will make other pods that is scheduled to the same host,  pending for image pulling. If the image is pull from somewhere slow， other pod may start after minutes or even a hour.
- Big image is not good practice, but may effect others.
- docker 1.9 is really old. 

Why not set --serialize-image-pulls to false in the future release ?

pkg/kubelet/images/puller.go
```diff
@@ -70,26 +74,78 @@ func newSerialImagePuller(imageService kubecontainer.ImageService) imagePuller {
...
 }
 
 func (sip *serialImagePuller) processImagePullRequests() {
-	for pullRequest := range sip.pullRequests {
-		imageRef, err := sip.imageService.PullImage(pullRequest.spec, pullRequest.pullSecrets, pullRequest.podSandboxConfig)
-		pullRequest.pullChan <- pullResult{
-			imageRef: imageRef,
-			err:      err,
+	// Set environment variables to identify the namespace that needs to be pulled first.
+	priorityNamespace := env.GetEnvAsStringOrFallback("PRIORITY_NAMESPACES", "kube-system")
```
pacoxu, line 95: priorityNamespace would be another feature in this pr.

`NamespaceSystem is the system namespace` 
You may use NamespaceSystem here.

PR #93243, opened 2020-07-20
Implement TopologyInfo and cpu_ids in podresources interface

AlexeyPerevalov:
/kind api-change

 This PR introduces additional information numaid and cpu_ids list (in [cpuset list format](https://man7.org/linux/man-pages/man7/cpuset.7.html#FORMATS) ).
It implements KEP [1884](https://github.com/kubernetes/enhancements/pull/1884)
 It's necessary for [topology aware scheduling](https://github.com/kubernetes/enhancements/pull/1858), this information will be used in [topology exposing daemon](https://github.com/kubernetes/enhancements/pull/1870).

To test it you need to enable `KubeletPodResources`  feature gate in the command line option or in the KubeletConfiguration (by default it's in the /var/lib/kubelet/config.yaml file)

Also you can use following sample to track down pod resources

```
package main

import (
        "context"
        "fmt"
        "time"

        podresources "k8s.io/kubernetes/pkg/kubelet/apis/podresources"
        podresourcesapi "k8s.io/kubernetes/pkg/kubelet/apis/podresources/v1alpha1"
)

var timeout = 2 * time.Minute

func main() {
        socket := "unix:///var/lib/kubelet/pod-resources/kubelet.sock"
        client, _, err := podresources.GetClient(socket, timeout, 1000)
        if err != nil {
                fmt.Println("Can't create client:", err)
                return
        }
        req := podresourcesapi.ListPodResourcesRequest{}
        resp, err := client.List(context.TODO(), &req)
        if err != nil {
                fmt.Println("Can't receive response:", err)
                return
        }
        for idx := range resp.PodResources {
                podresource := resp.PodResources[idx]
                fmt.Printf("podresource: %v", podresource)
        }
}

```
Issue: https://github.com/kubernetes/enhancements/issues/2043


pkg/kubelet/apis/podresources/server.go
```diff
@@ -29,6 +29,12 @@ type DevicesProvider interface {
 	UpdateAllocatedDevices()
 }
 
+// DevicesProvider knows how to provide the devices used by the given container
+type CPUsProvider interface {
+	GetCPUs(podUID, containerName string) []uint32
+	UpdateAllocatedDevices()
```
swatisehgal, line 35: We should name this more appropriately, maybe UpdateAllocatedCPUs(). Also, where is this method implemented? I assumed that it would be implemented in cpu_manager.go but didn't find its implementation

pkg/kubelet/cm/cpumanager/cpu_manager.go
```diff
@@ -461,3 +465,12 @@ func (m *manager) updateContainerCPUSet(containerID string, cpus cpuset.CPUSet)
 			CpusetCpus: cpus.String(),
 		})
 }
+
+func (m *manager) GetCPUs(podUID, containerName string) []uint32 {
+	cpus := m.state.GetCPUSetOrDefault(string(podUID), containerName)
+	result := []uint32{}
+	for cpu := range cpus.ToSliceNoSort() {
```
swatisehgal, line 472: This line should be corrected to : `for _, cpu := range cpus.ToSliceNoSort() ` The current implementation results in the index values being stored in the result slice (e.g. []int32{0,1,2,3,4....} )

swatisehgal (changes requested):

pkg/kubelet/cm/cpumanager/cpu_manager.go
AlexeyPerevalov: exactly

pkg/kubelet/apis/podresources/server.go
AlexeyPerevalov: Thanks, nice catch, I forget to remove it from here. In DeviceManager this interface is necessary to actualize podDevices state (GetDevice of DeviceManager return devices from podDevices), for CPUManager, where we return cpu directly - UpdateAllocatedDevices interface is not necessary as well as something like podDevices abstraction, since CPUManager uses CheckpointState for such purpose.

swatisehgal: That sounds reasonable!

fromanirh:
(general questions, not for Alexey)
not completely sure why this PR runs on the e2e-gce-device-plugin-gpu - I guess it is required by changes to podresources?
Is the environment still available?

staging/src/k8s.io/kubelet/pkg/apis/podresources/v1alpha1/api.proto
```diff
@@ -39,10 +39,22 @@ message PodResources {
 message ContainerResources {
     string name = 1;
     repeated ContainerDevices devices = 2;
+    repeated uint32 cpu_ids = 3;
```
fromanirh, line 42: plase note it is int64 in the KEP: https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2043-pod-resource-concrete-assigments#proposed-api

which version is correct? this or KEP's?

AlexeyPerevalov: right, int32 was in initial version, but since kernel keeps it as int, it's more safer to keep it as int64. I'll update it soon.

pkg/kubelet/cm/devicemanager/manager.go
```diff
@@ -397,7 +397,7 @@ func (m *ManagerImpl) UpdatePluginResources(node *schedulerframework.NodeInfo, a
 	defer m.mutex.Unlock()
 
 	// quick return if no pluginResources requested
-	if _, podRequireDevicePluginResource := m.podDevices[string(pod.UID)]; !podRequireDevicePluginResource {
+	if _, podRequireDevicePluginResource := m.podDevices.devs[string(pod.UID)]; !podRequireDevicePluginResource {
```
RenaudWasTaken, line 400: This seems to be protected by the manager mutex, but would be nice to be a function call protected by the podDevices object :)

pkg/kubelet/cm/devicemanager/pod_devices.go
```diff
@@ -35,54 +37,79 @@ type deviceAllocateInfo struct {
...
-type podDevices map[string]containerDevices             // Keyed by podUID.
+type podDevices struct {
+	sync.RWMutex
+	devs map[string]containerDevices // Keyed by podUID.
+}
+
+// NewPodDevices is a function that returns object of podDevices type with its own guard
+// RWMutex and a map where key is a pod UID and value contains
+// container devices information of type containerDevices.
+func newPodDevices() podDevices {
```
RenaudWasTaken, line 48: As a rule of thumb `newSomething` should return a pointer :)

RenaudWasTaken (commented):
Thanks for making the lock changes! A few small nits, can you also split the protobuf change in its own commit?

Thanks a lot!

AlexeyPerevalov:
> Thanks for making the lock changes! A few small nits, can you also split the protobuf change in its own commit?
> 
> Thanks a lot!

Do you mean to put staging/src/k8s.io/kubelet/pkg/apis/podresources/v1alpha1/api.pb.go  into Add TopologyInfo and cpu_ids into podresources?

pkg/kubelet/cm/devicemanager/manager.go
AlexeyPerevalov: yes, here and in other places manager's mutex.Lock could be avoided. But manager's mutex still protect m.allocatedDevices, m.healthyDevices, etc, and it could be also factored out to RWmutex or even to lock free structure.

fromanirh:
>     * Commit "Add TopologyInfo and cpu_ids into podresources" should contain the protobug
> 
>     * Commit "Regerate PodResources golang API" (or a message you prefer) should contain the new golang code
> 
> 
> This is the general pattern I've observed in kubernetes, it's pretty nice when it comes to merge conflicts

Thanks, this helps also my PR. I somehow missed this pattern up until now., sorry about that. Rephrasing in my words to doublecheck: the best option is to have three commits:
1. protobuf changes (api.proto)
2. code generation (api.pb.go)
3. implementation proper

RenaudWasTaken:
Yep for the three commits!

pkg/kubelet/cm/devicemanager/manager.go
```diff
@@ -393,11 +393,8 @@ func (m *ManagerImpl) Allocate(pod *v1.Pod, container *v1.Container) error {
 func (m *ManagerImpl) UpdatePluginResources(node *schedulerframework.NodeInfo, attrs *lifecycle.PodAdmitAttributes) error {
 	pod := attrs.Pod
 
-	m.mutex.Lock()
```
RenaudWasTaken, line 396: Unfortunately I think the `m.sanitizeNodeAllocatable` still needs the lock since it seems to be reading from `m.allocatedDevices`

RenaudWasTaken:
Great PR! Ping me as soon as the lock issue is fixed :) !

pkg/kubelet/cm/devicemanager/manager.go
AlexeyPerevalov: yes, I also forgot about isDevicePluginResource

spiffxp:
possible github issue, please link if you run into this on other PR's https://github.com/kubernetes/test-infra/issues/19910

AlexeyPerevalov:
> possible github issue, please link if you run into this on other PR's [kubernetes/test-infra#19910](https://github.com/kubernetes/test-infra/issues/19910)

Yes, it's
I and prow robot can't
git fetch https://github.com/kubernetes/kubernetes.git pull/93243/head
as result no commits after fetch.

pkg/kubelet/cm/devicemanager/manager.go
derekwaynecarr: just noting this appears resolved below

derekwaynecarr (approved):
can you add a node_e2e test that invokes this in a follow-on?

/lgtm
/approve

fromanirh:
> can you add a node_e2e test that invokes this in a follow-on?
> 
> /lgtm
> /approve

@derekwaynecarr I'm adding e2e tests for podresources in my PR (https://github.com/kubernetes/kubernetes/pull/95734/commits/2bda2a711e292789833af001db3beb4bbdb6380f and 
https://github.com/kubernetes/kubernetes/pull/95734/commits/cba703b36487c3d2912a37d271036505873687a8) We already have basic coverage. If works for you I'll prepare a followup PR to make sure this case is covered explicitely (e.g. checking the NUMA ids)

PR #91049, opened 2020-05-13
change pod status not belong to kubelet log level

ks864148379:
change pod status not belong to kubelet log level

**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:

We add some condition on Pod. But kubelt report error log "Kubelet is trying to update pod condition xxx for pod xxx. But it is not owned by kubelet." very often.

We don't have to deal with this error log and I think this log is log level is too high and it make some trouble to us.

PR #93245, opened 2020-07-20
update log level for conditions not owned by kubelet

fatedier:
**What type of PR is this?**
>

/kind cleanup

Currently `PodReadinessGate` feature is GA. `Custom Pod Condition` is normal for this feature.

Kubelet maybe not log this as an error when it needs to update pod conditions.

```release-note
NONE
```

```docs

```


SergeyKanzhelev:
dup of https://github.com/kubernetes/kubernetes/pull/91049/files

Can you please clarify why the fix is to change the log level, not change the logic.

PR #93246, opened 2020-07-20
remove some notes about scheduler/algorithm

lixiaobing1:
What type of PR is this?
/kind cleanup

Path pkg/scheduler/algorithm has been removed, so following notes better to remove:

	// the given node fits the pod. If Filter doesn't return "Success",
	// please refer scheduler/algorithm/predicates/error.go
	// to set error message.

### DIFF
--- a/pkg/scheduler/framework/v1alpha1/interface.go
+++ b/pkg/scheduler/framework/v1alpha1/interface.go
@@ -262,8 +262,7 @@ type FilterPlugin interface {
 	// Filter is called by the scheduling framework.
 	// All FilterPlugins should return "Success" to declare that
 	// the given node fits the pod. If Filter doesn't return "Success",
-	// please refer scheduler/algorithm/predicates/error.go
-	// to set error message.
+	// it will return "Unschedulable", "UnschedulableAndUnresolvable" or "Error".
 	// For the node being evaluated, Filter plugins should look at the passed
 	// nodeInfo reference for this particular node's information (e.g., pods
 	// considered to be running on the node) instead of looking it up in the


pkg/scheduler/framework/v1alpha1/interface.go
```diff
@@ -261,9 +261,7 @@ type FilterPlugin interface {
 	Plugin
 	// Filter is called by the scheduling framework.
 	// All FilterPlugins should return "Success" to declare that
-	// the given node fits the pod. If Filter doesn't return "Success",
-	// please refer scheduler/algorithm/predicates/error.go
-	// to set error message.
+	// the given node fits the pod.
```
chendave, line 264: sgtm, looks like those errors has been defined in each of the plugin instead.

btw, it's would be cool to update those line as well?
https://github.com/kubernetes/kubernetes/blob/43fbe17dc682b703b0b663935a797244378f0aaf/pkg/scheduler/framework/v1alpha1/interface.go#L256

`UnschedulableAndUnresolvable` is one of the code that would return from filter plugin.

damemi (commented):
looks like you need to `./hack/update-gofmt.sh`

alculquicondor:
Also please remove the Release Note, as this is NOT a user facing change.

alculquicondor:
Please change the release note to NONE

damemi:
@lixiaobing1 for future reference NONE means there still needs to be a `release-note` section, with literally `NONE` in it. Like so:

```
```release-note
NONE
.```
```

(added a `.` on the last three backticks so it would show)

pkg/scheduler/framework/v1alpha1/interface.go
```diff
@@ -262,8 +262,7 @@ type FilterPlugin interface {
 	// Filter is called by the scheduling framework.
 	// All FilterPlugins should return "Success" to declare that
 	// the given node fits the pod. If Filter doesn't return "Success",
-	// please refer scheduler/algorithm/predicates/error.go
-	// to set error message.
+	// it will return "Unschedulable" or "Error".
```
alculquicondor, line 265: Actually, you missed "UnschedulableAndUnresolvable". Could you please add it?
Also, use present tense.

lixiaobing1:
> @lixiaobing1 for future reference NONE means there still needs to be a `release-note` section, with literally `NONE` in it. Like so:
> 
> ```
> ```release-note
> NONE
> .```
> ```
> 
> (added a `.` on the last three backticks so it would show)

ok,thx :)

pkg/scheduler/framework/v1alpha1/interface.go
lixiaobing1: ok :)

lixiaobing1:
@damemi, ok thx:)

PR #93252, opened 2020-07-20
scheduler_perf: refactor to allow arbitrary workloads

### LINKED ISSUE #93117: scheduler: support arbitrary workloads in integration benchmarks
Currently the Kubernetes scheduler integration benchmarks support workloads of a particular kind -- simulated nodes are created and then an initial set of pods are scheduled. Once this setup is complete, a second and final set of pods are scheduled and various metrics are measured. I propose improving this to support arbitrary workloads that can support benchmarking heterogeneous mixture of pods, complex scheduling patterns, synchronisation primitives and cluster autoscaling.

Please see this doc for details: https://docs.google.com/document/d/1IxBgOUBTBSMSsX381ASdKuTiZE508YNXa5eVFRMzABk/edit?usp=sharing

/sig scheduling
/kind feature
/cc @alculquicondor @ahg-g @Huang-Wei @ingvagabund

adtac:
**What type of PR is this?**

/sig scheduling
/kind feature

Currently the Kubernetes scheduler integration benchmarks support workloads of a particular kind -- simulated nodes are created and then an initial set of pods are scheduled. Once this setup is complete, a second and final set of pods are scheduled and various metrics are measured. I propose improving this to support arbitrary workloads that can support benchmarking heterogeneous mixture of pods, complex scheduling patterns, synchronisation primitives and cluster autoscaling.

Fixes https://github.com/kubernetes/kubernetes/issues/93117

Work in progress, yet to migrate all existing tests over to the new pattern.

```release-note
NONE
```

https://docs.google.com/document/d/1IxBgOUBTBSMSsX381ASdKuTiZE508YNXa5eVFRMzABk/edit?usp=sharing

### DIFF
--- a/test/integration/framework/perf_utils.go
+++ b/test/integration/framework/perf_utils.go
@@ -58,7 +58,7 @@ func NewIntegrationTestNodePreparerWithNodeSpec(client clientset.Interface, coun
 }
 
 // PrepareNodes prepares countToStrategy test nodes.
-func (p *IntegrationTestNodePreparer) PrepareNodes() error {
+func (p *IntegrationTestNodePreparer) PrepareNodes(nextNodeIndex int) error {
 	numNodes := 0
 	for _, v := range p.countToStrategy {
 		numNodes += v.Count
@@ -103,11 +103,9 @@ func (p *IntegrationTestNodePreparer) PrepareNodes() error {
 	if err != nil {
 		klog.Fatalf("Error listing nodes: %v", err)
 	}
-	index := 0
-	sum := 0
+	index := nextNodeIndex
 	for _, v := range p.countToStrategy {
-		sum += v.Count
-		for ; index < sum; index++ {
+		for i := 0; i < v.Count; i, index = i+1, index+1 {
 			if err := testutils.DoPrepareNode(p.client, &nodes.Items[index], v.Strategy); err != nil {
 				klog.Errorf("Aborting node preparation: %v", err)
 				return err
@@ -119,14 +117,18 @@ func (p *IntegrationTestNodePreparer) PrepareNodes() error {
 
 // CleanupNodes deletes existing test nodes.
 func (p *IntegrationTestNodePreparer) CleanupNodes() error {
+	// TODO(#93794): make CleanupNodes only clean up the nodes created by this
+	// IntegrationTestNodePreparer to make this more intuitive.
 	nodes, err := GetReadySchedulableNodes(p.client)
 	if err != nil {
 		klog.Fatalf("Error listing nodes: %v", err)
 	}
+	var errRet error
 	for i := range nodes.Items {
 		if err := p.client.CoreV1().Nodes().Delete(context.TODO(), nodes.Items[i].Name, metav1.DeleteOptions{}); err != nil {
 			klog.Errorf("Error while deleting Node: %v", err)
+			errRet = err
 		}
 	}
-	return nil
+	return errRet
 }
--- a/test/integration/scheduler_perf/config/performance-config.yaml
+++ b/test/integration/scheduler_perf/config/performance-config.yaml
@@ -1,264 +1,426 @@
-- template:
-    desc: SchedulingBasic
-    initPods:
-    - podTemplatePath: config/pod-default.yaml
-    podsToSchedule:
-      podTemplatePath: config/pod-default.yaml
-  params:
-  - numNodes: 500
-    numInitPods: [500]
-    numPodsToSchedule: 1000
-  - numNodes: 5000
-    numInitPods: [5000]
-    numPodsToSchedule: 1000
-- template:
-    desc: SchedulingPodAntiAffinity
-    nodes:
-      uniqueNodeLabelStrategy:
-        labelKey: kubernetes.io/hostname
-    initPods:
-    - podTemplatePath: config/pod-with-pod-anti-affinity.yaml
-    podsToSchedule:
-      podTemplatePath: config/pod-with-pod-anti-affinity.yaml
-  params:
-  - numNodes: 500
-    numInitPods: [100]
-    numPodsToSchedule: 400
-  - numNodes: 5000
-    numInitPods: [1000]
-    numPodsToSchedule: 1000
-- template:
-    desc: SchedulingSecrets
-    initPods:
-    - podTemplatePath: config/pod-with-secret-volume.yaml
-    podsToSchedule:
-      podTemplatePath: config/pod-with-secret-volume.yaml
-  params:
-  - numNodes: 500
-    numInitPods: [500]
-    numPodsToSchedule: 1000
-  - numNodes: 5000
-    numInitPods: [5000]
-    numPodsToSchedule: 1000
-- template:
-    desc: SchedulingInTreePVs
-    initPods:
-    - persistentVolumeTemplatePath: config/pv-aws.yaml
-      persistentVolumeClaimTemplatePath: config/pvc.yaml
-    podsToSchedule:
-      persistentVolumeTemplatePath: config/pv-aws.yaml
-      persistentVolumeClaimTemplatePath: config/pvc.yaml
-  params:
-  - numNodes: 500
-    numInitPods: [500]
-    numPodsToSchedule: 1000
-  - numNodes: 5000
-    numInitPods: [5000]
-    numPodsToSchedule: 1000
-- template:
-    desc: SchedulingMigratedInTreePVs
-    nodes:
-      nodeTemplatePath: config/node-default.yaml
-      nodeAllocatableStrategy:
-        nodeAllocatable:
-          attachable-volumes-csi-ebs.csi.aws.com: 39
-        csiNodeAllocatable:
-          ebs.csi.aws.com:
-            count: 39
-        migratedPlugins:
-        - "kubernetes.io/aws-ebs"
-    initPods:
-    - persistentVolumeTemplatePath: config/pv-aws.yaml
-      persistentVolumeClaimTemplatePath: config/pvc.yaml
-    podsToSchedule:
-      persistentVolumeTemplatePath: config/pv-aws.yaml
-      persistentVolumeClaimTemplatePath: config/pvc.yaml
-    featureGates:
-      CSIMigration: true
-      CSIMigrationAWS: true
-  params:
-  - numNodes: 500
-    numInitPods: [500]
-    numPodsToSchedule: 1000
-  - numNodes: 5000
-    numInitPods: [5000]
-    numPodsToSchedule: 1000
-- template:
-    desc: SchedulingCSIPVs
-    nodes:
-      nodeTemplatePath: config/node-default.yaml
-      nodeAllocatableStrategy:
-        nodeAllocatable:
-          attachable-volumes-csi-ebs.csi.aws.com: 39
-        csiNodeAllocatable:
-          ebs.csi.aws.com:
-            count: 39
-    initPods:
-    - persistentVolumeTemplatePath: config/pv-csi.yaml
-      persistentVolumeClaimTemplatePath: config/pvc.yaml
-    podsToSchedule:
-      persistentVolumeTemplatePath: config/pv-csi.yaml
-      persistentVolumeClaimTemplatePath: config/pvc.yaml
-  params:
-  - numNodes: 500
-    numInitPods: [500]
-    numPodsToSchedule: 1000
-  - numNodes: 5000
-    numInitPods: [5000]
-    numPodsToSchedule: 1000
-- template:
-    desc: SchedulingPodAffinity
-    nodes:
-      nodeTemplatePath: config/node-default.yaml
-      labelNodePrepareStrategy:
-        labelKey: "failure-domain.beta.kubernetes.io/zone"
-        labelValues: ["zone1"]
-    initPods:
-    - podTemplatePath: config/pod-with-pod-affinity.yaml
-    podsToSchedule:
-      podTemplatePath: config/pod-with-pod-affinity.yaml
-  params:
-  - numNodes: 500
-    numInitPods: [500]
-    numPodsToSchedule: 1000
-  - numNodes: 5000
-    numInitPods: [5000]
-    numPodsToSchedule: 1000
-- template:
-    desc: SchedulingPreferredPodAffinity
-    nodes:
-      uniqueNodeLabelStrategy:
-        labelKey: kubernetes.io/hostname
-    initPods:
-    - podTemplatePath: config/pod-with-preferred-pod-affinity.yaml
-    podsToSchedule:
-      podTemplatePath: config/pod-with-preferred-pod-affinity.yaml
-  params:
-  - numNodes: 500
-    numInitPods: [500]
-    numPodsToSchedule: 1000
-  - numNodes: 5000
-    numInitPods: [5000]
-    numPodsToSchedule: 1000
-- template:
-    desc: SchedulingPreferredPodAntiAffinity
-    nodes:
-      uniqueNodeLabelStrategy:
-        labelKey: kubernetes.io/hostname
-    initPods:
-    - podTemplatePath: config/pod-with-preferred-pod-anti-affinity.yaml
-    podsToSchedule:
-      podTemplatePath: config/pod-with-preferred-pod-anti-affinity.yaml
-  params:
-  - numNodes: 500
-    numInitPods: [500]
-    numPodsToSchedule: 1000
-  - numNodes: 5000
-    numInitPods: [5000]
-    numPodsToSchedule: 1000
-- template:
-    desc: SchedulingNodeAffinity
-    nodes:
-      nodeTemplatePath: config/node-default.yaml
-      labelNodePrepareStrategy:
-        labelKey: "failure-domain.beta.kubernetes.io/zone"
-        labelValues: ["zone1"]
-    initPods:
-    - podTemplatePath: config/pod-with-node-affinity.yaml
-    podsToSchedule:
-      podTemplatePath: config/pod-with-node-affinity.yaml
-  params:
-  - numNodes: 500
-    numInitPods: [500]
-    numPodsToSchedule: 1000
-  - numNodes: 5000
-    numInitPods: [5000]
-    numPodsToSchedule: 1000
-- template:
-    desc: TopologySpreading
-    nodes:
-      nodeTemplatePath: config/node-default.yaml
-      labelNodePrepareStrategy:
-        labelKey: "topology.kubernetes.io/zone"
-        labelValues: ["moon-1", "moon-2", "moon-3"]
-    initPods:
-    - podTemplatePath: config/pod-default.yaml
-    podsToSchedule:
-      podTemplatePath: config/pod-with-topology-spreading.yaml
-  params:
-  - numNodes: 500
-    numInitPods: [1000]
-    numPodsToSchedule: 1000
-  - numNodes: 5000
-    numInitPods: [5000]
-    numPodsToSchedule: 2000
-- template:
-    desc: PreferredTopologySpreading
-    nodes:
-      nodeTemplatePath: config/node-default.yaml
-      labelNodePrepareStrategy:
-        labelKey: "topology.kubernetes.io/zone"
-        labelValues: ["moon-1", "moon-2", "moon-3"]
-    initPods:
-    - podTemplatePath: config/pod-default.yaml
-    podsToSchedule:
-      podTemplatePath: config/pod-with-preferred-topology-spreading.yaml
-  params:
-  - numNodes: 500
-    numInitPods: [1000]
-    numPodsToSchedule: 1000
-  - numNodes: 5000
-    numInitPods: [5000]
-    numPodsToSchedule: 2000
-- template:
-    desc: MixedSchedulingBasePod
-    nodes:
-      nodeTemplatePath: config/node-default.yaml
-      labelNodePrepareStrategy:
-        labelKey: "topology.kubernetes.io/zone"
-        labelValues: ["zone1"]
-    initPods:
-    - podTemplatePath: config/pod-default.yaml
-    - podTemplatePath: config/pod-with-pod-affinity.yaml
-    - podTemplatePath: config/pod-with-pod-anti-affinity.yaml
-    - podTemplatePath: config/pod-with-preferred-pod-affinity.yaml
-    - podTemplatePath: config/pod-with-preferred-pod-anti-affinity.yaml
-    podsToSchedule:
-      podTemplatePath: config/pod-default.yaml
-  params:
-  - numNodes: 500
-    numInitPods: [200, 200, 200, 200, 200]
-    numPodsToSchedule: 1000
-  - numNodes: 5000
-    numInitPods: [2000, 2000, 2000, 2000, 2000]
-    numPodsToSchedule: 1000
-- template:
-    desc: Preemption
-    initPods:
-    - podTemplatePath: config/pod-low-priority.yaml
-    podsToSchedule:
-      podTemplatePath: config/pod-high-priority.yaml
-  params:
-  - numNodes: 500
-    numInitPods: [2000]
-    numPodsToSchedule: 500
-  - numNodes: 5000
-    numInitPods: [20000]
-    numPodsToSchedule: 5000
-- template:
-    desc: Unschedulable
-    skipWaitUntilInitPodsScheduled: true
-    initPods:
-    - podTemplatePath: config/pod-large-cpu.yaml
-    podsToSchedule:
-      podTemplatePath: config/pod-default.yaml
-  params:
-  - numNodes: 500
-    numInitPods: [200]
-    numPodsToSchedule: 1000
-  - numNodes: 5000
-    numInitPods: [200]
-    numPodsToSchedule: 5000
-  - numNodes: 5000
-    numInitPods: [2000]
-    numPodsToSchedule: 5000
+- name: SchedulingBasic
+  workloadTemplate:
+  - opcode: createNodes
+    countParam: $initNodes
+  - opcode: createPods
+    countParam: $initPods
+    podTemplatePath: config/pod-default.yaml
+  - opcode: createPods
+    countParam: $measurePods
+    podTemplatePath: config/pod-default.yaml
+    collectMetrics: true
+  workloads:
+  - name: 500Nodes
+    params:
+      initNodes: 500
+      initPods: 500
+      measurePods: 1000
+  - name: 5000Nodes
+    params:
+      initNodes: 5000
+      initPods: 1000
+      measurePods: 1000
+
+- name: SchedulingPodAntiAffinity
+  workloadTemplate:
+  - opcode: createNodes
+    countParam: $initNodes
+    uniqueNodeLabelStrategy:
+      labelKey: kubernetes.io/hostname
+  - opcode: createPods
+    countParam: $initPods
+    podTemplatePath: config/pod-with-pod-anti-affinity.yaml
+    namespace: sched-setup
+  - opcode: createPods
+    countParam: $measurePods
+    podTemplatePath: config/pod-with-pod-anti-affinity.yaml
+    collectMetrics: true
+    namespace: sched-test
+  workloads:
+  - name: 500Nodes
+    params:
+      initNodes: 500
+      initPods: 100
+      measurePods: 400
+  - name: 5000Nodes
+    params:
+      initNodes: 500
+      initPods: 100
+      measurePods: 400
+
+- name: SchedulingSecrets
+  workloadTemplate:
+  - opcode: createNodes
+    countParam: $initNodes
+  - opcode: createPods
+    countParam: $initPods
+    podTemplatePath: config/pod-with-secret-volume.yaml
+  - opcode: createPods
+    countParam: $measurePods
+    podTemplatePath: config/pod-with-secret-volume.yaml
+    collectMetrics: true
+  workloads:
+  - name: 500Nodes
+    params:
+      initNodes: 500
+      initPods: 500
+      measurePods: 1000
+  - name: 5000Nodes
+    params:
+      initNodes: 5000
+      initPods: 5000
+      measurePods: 1000
+
+- name: SchedulingInTreePVs
+  workloadTemplate:
+  - opcode: createNodes
+    countParam: $initNodes
+  - opcode: createPods
+    countParam: $initPods
+    persistentVolumeTemplatePath: config/pv-aws.yaml
+    persistentVolumeClaimTemplatePath: config/pvc.yaml
+  - opcode: createPods
+    countParam: $measurePods
+    persistentVolumeTemplatePath: config/pv-aws.yaml
+    persistentVolumeClaimTemplatePath: config/pvc.yaml
+    collectMetrics: true
+  workloads:
+  - name: 500Nodes
+    params:
+      initNodes: 500
+      initPods: 500
+      measurePods: 1000
+  - name: 5000Nodes
+    params:
+      initNodes: 5000
+      initPods: 5000
+      measurePods: 1000
+
+- name: SchedulingMigratedInTreePVs
+  featureGates:
+    CSIMigration: true
+    CSIMigrationAWS: true
+  workloadTemplate:
+  - opcode: createNodes
+    countParam: $initNodes
+    nodeTemplatePath: config/node-default.yaml
+    nodeAllocatableStrategy:
+      nodeAllocatable:
+        attachable-volumes-csi-ebs.csi.aws.com: "39"
+      csiNodeAllocatable:
+        ebs.csi.aws.com:
+          count: 39
+      migratedPlugins:
+      - "kubernetes.io/aws-ebs"
+  - opcode: createPods
+    countParam: $initPods
+    persistentVolumeTemplatePath: config/pv-aws.yaml
+    persistentVolumeClaimTemplatePath: config/pvc.yaml
+  - opcode: createPods
+    countParam: $measurePods
+    persistentVolumeTemplatePath: config/pv-aws.yaml
+    persistentVolumeClaimTemplatePath: config/pvc.yaml
+    collectMetrics: true
+  workloads:
+  - name: 500Nodes
+    params:
+      initNodes: 500
+      initPods: 500
+      measurePods: 1000
+  - name: 5000Nodes
+    params:
+      initNodes: 5000
+      initPods: 5000
+      measurePods: 1000
+
+- name: SchedulingCSIPVs
+  workloadTemplate:
+  - opcode: createNodes
+    countParam: $initNodes
+    nodeTemplatePath: config/node-default.yaml
+    nodeAllocatableStrategy:
+      nodeAllocatable:
+        attachable-volumes-csi-ebs.csi.aws.com: "39"
+      csiNodeAllocatable:
+        ebs.csi.aws.com:
+          count: 39
+  - opcode: createPods
+    countParam: $initPods
+    persistentVolumeTemplatePath: config/pv-csi.yaml
+    persistentVolumeClaimTemplatePath: config/pvc.yaml
+  - opcode: createPods
+    countParam: $measurePods
+    persistentVolumeTemplatePath: config/pv-csi.yaml
+    persistentVolumeClaimTemplatePath: config/pvc.yaml
+    collectMetrics: true
+  workloads:
+  - name: 500Nodes
+    params:
+      initNodes: 500
+      initPods: 500
+      measurePods: 1000
+  - name: 5000Nodes
+    params:
+      initNodes: 5000
+      initPods: 5000
+      measurePods: 1000
+
+- name: SchedulingPodAffinity
+  workloadTemplate:
+  - opcode: createNodes
+    countParam: $initNodes
+    nodeTemplatePath: config/node-default.yaml
+    labelNodePrepareStrategy:
+      labelKey: "failure-domain.beta.kubernetes.io/zone"
+      labelValues: ["zone1"]
+  - opcode: createPods
+    countParam: $initPods
+    podTemplatePath: config/pod-with-pod-affinity.yaml
+    namespace: sched-setup
+  - opcode: createPods
+    countParam: $measurePods
+    podTemplatePath: config/pod-with-pod-affinity.yaml
+    namespace: sched-test
+    collectMetrics: true
+  workloads:
+  - name: 500Nodes
+    params:
+      initNodes: 500
+      initPods: 500
+      measurePods: 1000
+  - name: 5000Nodes
+    params:
+      initNodes: 5000
+      initPods: 5000
+      measurePods: 1000
+
+- name: SchedulingPreferredPodAffinity
+  workloadTemplate:
+  - opcode: createNodes
+    countParam: $initNodes
+    uniqueNodeLabelStrategy:
+      labelKey: kubernetes.io/hostname
+  - opcode: createPods
+    countParam: $initPods
+    podTemplatePath: config/pod-with-preferred-pod-affinity.yaml
+    namespace: sched-setup
+  - opcode: createPods
+    countParam: $measurePods
+    podTemplatePath: config/pod-with-preferred-pod-affinity.yaml
+    namespace: sched-test
+    collectMetrics: true
+  workloads:
+  - name: 500Nodes
+    params:
+      initNodes: 500
+      initPods: 500
+      measurePods: 1000
+  - name: 5000Nodes
+    params:
+      initNodes: 5000
+      initPods: 5000
+      measurePods: 1000
+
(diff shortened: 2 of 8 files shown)


adtac:
Marking as ready for review, this is no longer WIP. All existing benchmarks have been ported, but no new benchmarks have been added. Metrics for the new benchmarks mostly match with the data from the in-tree benchmarks for the most part; this is as expected since the benchmarks do the same thing, just in a different way (note: a fraction of the `scheduler_pod_scheduling_duration_seconds` metrics don't match for some reason). See below for the diffs:

- [SchedulingCSIPVs](http://ix.io/2tqD/diff)
- [SchedulingInTreePVs](http://ix.io/2tqE/diff)
- [SchedulingMigratedInTreePVs](http://ix.io/2tqF/diff)
- [SchedulingPodAffinity](http://ix.io/2tqG/diff)
- [SchedulingPreferredPodAffinity](http://ix.io/2tqH/diff)
- [SchedulingPreferredPodAntiAffinity](http://ix.io/2tqI/diff)
- [SchedulingSecrets](http://ix.io/2tqJ/diff)
- [TopologySpreading,PreferredTopologySpreading](http://ix.io/2tqK/diff)

(New benchmarks have shorter `name` fields; use the - / + markers to identify which data corresponds to which benchmark.)

/cc @ahg-g @alculquicondor @Huang-Wei @ingvagabund

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +53,475 @@ var (
...
+	SkipWaitToCompletion bool
+	// Persistent volume settings for the pods to be scheduled. Optional.
 	PersistentVolumeTemplatePath      *string
 	PersistentVolumeClaimTemplatePath *string
 }
 
-// simpleTestCases defines a set of test cases that share the same template (node spec, pod spec, etc)
-// with testParams(e.g., NumNodes) being overridden. This provides a convenient way to define multiple tests
-// with various sizes.
-type simpleTestCases struct {
```
alculquicondor, line 101: I just noticed that this was getting removed.

What if we make `testParams` just a dictionary?  Then each quantity (like `createPodsOp.CreatePods`) is a string, that could be one of the variables, or a plain number (I would just have 2 fields, each for one purpose).

adtac: do you mean something like this?

```
SchedulingBasic:
  500Nodes:
  - CreateNodes: 500
  - CreatePods: 500
```

This wouldn't allow us to have testCase level params (like featureGates, metricsCollectorConfig). Or do you mean to say we should keep the `workloads` field, but within that use maps? I don't mind either, but I like the ordered way of describing things with arrays. With a map, the 5000Nodes case could run before the 500Nodes case; that's probably not an issue, but it feels slightly less intuitive.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +53,475 @@ var (
...
+	Name string
+	// Feature gates to set before running the test. Optional.
 	FeatureGates map[featuregate.Feature]bool
-	// optional, replaces default defaultMetricsCollectorConfig if supplied.
+	// List of metrics to collect. Optional, defaults to
+	// defaultMetricsCollectorConfig if unspecified.
 	MetricsCollectorConfig *metricsCollectorConfig
+	// List of workloads to run under this testCase.
+	Workloads []*workload
+	// TODO(adtac): reduce config toil by having a default pod and node spec per
```
ahg-g, line 69: nit: replace with an issue number.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +53,475 @@ var (
...
+}
+
+// UnmarshalJSON is a custom unmarshaler for the op struct since we don't know
+// which op we're decoding at runtime.
+func (op *op) UnmarshalJSON(b []byte) error {
+	possibleOps := []realOp{
+		&createNodesOp{},
+		&createPodsOp{},
+		&barrierOp{},
+		// TODO(adtac): add a sleep timer op to simulate user action?
```
ahg-g, line 106: ditto, perhaps open a single issue listing all possible improvements.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +53,475 @@ var (
...
+			if err != nil {
+				b.Fatal(err)
+			}
+			if err := nodePreparer.PrepareNodes(numNodes); err != nil {
+				b.Fatal(err)
+			}
+			if numNodes == 0 {
+				// Schedule a cleanup at most once. The CleanupNodes function will list
+				// and delete *all* nodes.
+				// TODO(adtac): make CleanupNodes only clean up its own nodes to make
```
ahg-g, line 266: ditto

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +53,475 @@ var (
...
+			}
+		default:
+			b.Fatalf("invalid op %v", realOp)
+		}
 	}
 
-	// Schedule the main workload
-	if err := createPods(testNamespace, test.PodsToSchedule, clientset); err != nil {
-		b.Fatal(err)
+	if allPodsAsync {
```
ahg-g, line 348: ditto, I think this is not necessary, stop/start from beginning to end of `runWorkload`.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +53,475 @@ var (
...
 	PersistentVolumeClaimTemplatePath *string
 }
 
-// simpleTestCases defines a set of test cases that share the same template (node spec, pod spec, etc)
-// with testParams(e.g., NumNodes) being overridden. This provides a convenient way to define multiple tests
-// with various sizes.
-type simpleTestCases struct {
-	Template testCase
-	Params   []testParams
+func (cpo *createPodsOp) isValid() error {
```
ahg-g, line 180: should we validate that `CollectMetrics` and `SkipWaitToCompletion` can't be true at the same time?

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +53,475 @@ var (
...
+			var namespace string
+			if realOp.Namespace != nil {
+				namespace = *realOp.Namespace
+			} else {
+				namespace = fmt.Sprintf("namespace-%d", opIndex)
+			}
+			var stopCh chan struct{}
+			var collectors []testDataCollector
+			if realOp.CollectMetrics {
+				b.StartTimer()
```
ahg-g, line 282: I don't think we should care too much about when to start/end this timer, we should just make it span the whole benchmark case, from beginning to end. This is because we will look mostly at the metrics (throughput, latency etc.).

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +53,475 @@ var (
...
+			}
+			// Make sure there's at least one CreatePods op with collectMetrics set
+			// to true in each workload. What's the point of running a performance
+			// benchmark if no statistics are collected for reporting?
+			if !w.collectsMetrics() {
+				return fmt.Errorf("%s/%s: none of the ops collect metrics", tc.Name, w.Name)
 			}
-			testCase.PodsToSchedule.Num = p.NumPodsToSchedule
-			testCases = append(testCases, testCase)
+			// TODO(adtac): make sure each workload within a test case has a unique
```
ahg-g, line 496: ditto, reference an enhancement issue.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: I did that originally, but then I added support for them both to be true. The metrics collector will run in the background when the benchmark finishes automatically.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: https://github.com/kubernetes/kubernetes/issues/93792

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: https://github.com/kubernetes/kubernetes/issues/93793

I think having separate issues might be easier from a discussion point-of-view. I don't have a strong case for the timer op for example, so we can close that issue if everyone thinks it'd be pointless without affecting the other issues.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: https://github.com/kubernetes/kubernetes/issues/93794

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: https://github.com/kubernetes/kubernetes/issues/93795

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: Ack, done.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: Done.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +53,458 @@ var (
...
+				}
+				go finish()
+				// We can't stop the timer after the pods actually get scheduled
+				// because a different timer may have been started and stopped by then.
+				// AFAIK, there is no way to add the time elapsed to the benchmark's
+				// timer concurrently without using StartTimer and StopTimer.
+				b.StopTimer()
+			} else {
+				finish()
+				b.StopTimer()
```
ahg-g, line 319: are those two StopTimer calls still necessary? I think not.

test/integration/scheduler_perf/scheduler_perf_test.go
ahg-g: StartTimer call still exist, I am suggesting to remove this.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: Oops, sorry about that, must've missed those two. Thanks for catching! Done.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: What would simpleTestCase.Params and createPodsOp.CreatePodsParam store? can you give a short yaml example, please? I don't fully understand what you're proposing yet, sorry.

alculquicondor: I'm thinking something like:

```yaml
- name: SchedulingBasic
  template:
    - createNodes: initialNodes
    - createPods: initialPods
      podTemplatePath: config/pod-default.yaml
    - createPods: podsToMeasure
      podTemplatePath: config/pod-default.yaml
      collectMetrics: true
  workloads:
  - name: 500Nodes
    params:
      initialNodes: 500
      initialPods: 500
      podsToMeasure: 1000
```

This example is actually not compatible with the structs I proposed earlier, so they need to be adapted. The idea is that workloads (or testCases?) only define values.

adtac: Ah I see what you mean.

While this will certainly reduce verbosity, there is another approach we can take too (https://github.com/kubernetes/kubernetes/issues/93792): since nearly all tests use just one node template and one pod template, we can use have a default template for nodes and pods defined at the root of the test case and simply omit the fields inside `workloads:`.

One concern I have with your proposal is that it becomes difficult to distinguish between string values and param-mapped values for the reader. We can potentially solve that with patterns like `{{initialNodes}}` for template values that must be replaced, though. I'm not too sure about the readability of such an approach though.

alculquicondor: casing would be another solution, something like `INITIAL_NODES`. But it's still an nit. The most important thing is making tests more flexible.

adtac: True. I think there's scope for doing both proposals: allow defining a default node/pod template *and* a workload ops template.

I like this workload templating approach; do the others have any opinions before I implement something like this?

ahg-g: sounds good to me.

adtac: Done.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +54,518 @@ var (
...
+				for _, collector := range collectors {
+					go collector.run(stopCh)
+				}
+			}
+			if err := createPods(namespace, concreteOp, clientset); err != nil {
+				b.Fatal(err)
+			}
+			barrierAndCollect := func() {
+				if err := barrierOne(podInformer, b.Name(), namespace, concreteOp.CreatePods.(int)); err != nil {
+					b.Fatal(err)
```
alculquicondor, line 360: Fatal is not properly supported within routines.

What you can do is pass the errors back through a channel and call `Fatal` from the main routine.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +54,518 @@ var (
...
+			numNodes += concreteOp.CreateNodes.(int)
+
+		case *createPodsOp:
+			var namespace string
+			if concreteOp.Namespace != nil {
+				namespace = *concreteOp.Namespace
+			} else {
+				namespace = fmt.Sprintf("namespace-%d", opIndex)
+			}
+			var stopCh chan struct{}
```
alculquicondor, line 346: Use `context.WithCancel` instead.
Use a parent context for the whole test and call it's cancel function from `b.Cleanup`

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +54,518 @@ var (
...
+			}
+			if concreteOp.SkipWaitToCompletion {
+				// Only record those namespaces that may potentially require barriers
+				// in the future.
+				if _, ok := numPodsScheduledPerNamespace[namespace]; ok {
+					numPodsScheduledPerNamespace[namespace] += concreteOp.CreatePods.(int)
+				} else {
+					numPodsScheduledPerNamespace[namespace] = concreteOp.CreatePods.(int)
+				}
+				go barrierAndCollect()
```
alculquicondor, line 379: why do we collect in this case? wouldn't metrics get mixed with metrics from future steps?

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +54,518 @@ var (
...
+			} else {
+				barrierAndCollect()
+			}
 
-	// Schedule the main workload
-	if err := createPods(testNamespace, test.PodsToSchedule, clientset); err != nil {
-		b.Fatal(err)
+		case *barrierOp:
+			for _, barrier := range concreteOp.Barrier {
```
alculquicondor, line 385: s/barrier/namespace

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +54,518 @@ var (
...
+			}
+			if err := barrier(podInformer, b.Name(), concreteOp.Barrier, numPodsScheduledPerNamespace); err != nil {
+				b.Fatal(err)
+			}
+			// At the end of the barrier, we can be sure that there are no pods
+			// pending scheduling in the namespaces that we just blocked on.
+			if len(concreteOp.Barrier) == 0 {
+				numPodsScheduledPerNamespace = make(map[string]int)
+			} else {
+				for _, namespace := range concreteOp.Barrier {
```
alculquicondor, line 398: just always iterate. It's easier to read and we don't need to optimize further.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +54,518 @@ var (
...
+		klog.Infof("%s: namespace %s: got %d existing pods, want %d", name, namespace, len(scheduled), wantCount)
+		time.Sleep(1 * time.Second)
 	}
-	if pc.PersistentVolumeClaimTemplatePath == nil {
-		return testutils.NewCustomCreatePodStrategy(basePod), nil
+	return nil
+}
+
+// barrier blocks until the all pods in the given namespaces are scheduled.
+func barrier(podInformer coreinformers.PodInformer, name string, namespaces []string, numPodsScheduledPerNamespace map[string]int) error {
```
alculquicondor, line 487: s/barrier/waitUntilPodsScheduled

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +54,518 @@ var (
...
+		scheduled, err := getScheduledPods(podInformer, namespace)
 		if err != nil {
-			return nil, err
+			return err
 		}
+		if len(scheduled) >= wantCount {
+			break
+		}
+		klog.Infof("%s: namespace %s: got %d existing pods, want %d", name, namespace, len(scheduled), wantCount)
+		time.Sleep(1 * time.Second)
```
alculquicondor, line 481: we need a timeout. You can reuse functions in `k8s.io/apimachinery/pkg/util/wait`

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +54,518 @@ var (
...
 	return podCreator.CreatePods()
 }
 
-func getPodStrategy(pc podCase) (testutils.TestPodCreateStrategy, error) {
-	basePod := makeBasePod()
-	if pc.PodTemplatePath != nil {
-		var err error
-		basePod, err = getPodSpecFromFile(pc.PodTemplatePath)
+// barrierOne blocks until all pods in the given namespace are scheduled.
+func barrierOne(podInformer coreinformers.PodInformer, name string, namespace string, wantCount int) error {
```
alculquicondor, line 471: s/barrier/waitUntilPodsScheduledInNamespace

test/integration/scheduler_perf/config/performance-config.yaml
```diff
@@ -1,264 +1,377 @@
...
-    numPodsToSchedule: 1000
-  - numNodes: 5000
-    numInitPods: [200]
-    numPodsToSchedule: 5000
-  - numNodes: 5000
-    numInitPods: [2000]
-    numPodsToSchedule: 5000
+- name: SchedulingBasic
+  workloadTemplate:
+  - createNodes: $initNodes
```
alculquicondor, line 3: I would prefer something like:

```
- op: createNodes
  number: $initNodes
```

Then you can use the string to disambiguate when unmarshaling

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +54,518 @@ var (
...
+}
+
+func isValidParameterizable(val string) bool {
+	return strings.HasPrefix(val, "$")
+}
+
+// createNodesOp defines an op where nodes are created as a part of a workload.
+type createNodesOp struct {
+	// Number of nodes to create. Parameterizable.
+	CreateNodes interface{}
```
alculquicondor, line 152: I would just use 2 members:

```
Count int
CountVar string
```

Then you don't need `isValid`

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +54,518 @@ var (
...
+	}
+	return &cno, (&cno).isValid(false)
+}
+
+// createPodsOp defines an op where pods are scheduled as a part of a workload.
+// The test can block on the completion of this op before moving forward or
+// continue asynchronously.
+type createPodsOp struct {
+	// Number of pods to schedule. Parameterizable.
+	CreatePods interface{}
```
alculquicondor, line 202: Same here for variables

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,258 +54,518 @@ var (
...
+// at the end of each each workload automatically.
+type barrierOp struct {
+	// Namespaces to block on. Required. Empty array signifies that the barrier
+	// should block on all namespaces.
+	Barrier []string
+}
+
+func (bo *barrierOp) isValid(allowParameterizable bool) error {
+	if bo.Barrier == nil {
+		// nil isn't the same as an empty array.
```
alculquicondor, line 265: In general, avoid this distinction.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: Done.

test/integration/framework/perf_utils.go
```diff
@@ -103,8 +103,8 @@ func (p *IntegrationTestNodePreparer) PrepareNodes() error {
 	if err != nil {
 		klog.Fatalf("Error listing nodes: %v", err)
 	}
-	index := 0
-	sum := 0
+	index := startIndex
+	sum := startIndex
 	for _, v := range p.countToStrategy {
 		sum += v.Count
 		for ; index < sum; index++ {
```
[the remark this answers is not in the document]
adtac: Done.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -343,24 +351,24 @@ func runWorkload(b *testing.B, tc *testCase, w *workload) []DataItem {
...
 			if err := createPods(namespace, concreteOp, clientset); err != nil {
-				b.Fatal(err)
+				b.Fatalf("op %d: %v", opIndex, err)
 			}
-			barrierAndCollect := func() {
-				if err := barrierOne(podInformer, b.Name(), namespace, concreteOp.CreatePods.(int)); err != nil {
-					b.Fatal(err)
+			barrierAndCollect := func(idx int) {
+				if err := waitUntilPodsScheduledInNamespace(podInformer, b.Name(), namespace, concreteOp.CreatePods.(int)); err != nil {
+					barrierAndCollectErrors[idx] = err
```
alculquicondor, line 367: you could still use a channel with just one element and use `fmt.Errorf` to indicate the operation index.

See https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/internal/parallelize/error_channel.go
(but don't use that directly as we prefer to keep that internal)

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -303,38 +307,42 @@ func BenchmarkPerfScheduling(b *testing.B) {
 	}
 }
 
-func runWorkload(b *testing.B, tc *testCase, w *workload) []DataItem {
-	var mu sync.Mutex
-	var dataItems []DataItem
+func runWorkload(testCtx context.Context, b *testing.B, tc *testCase, w *workload) []DataItem {
+	workloadCtx, subtestCancel := context.WithCancel(testCtx)
```
alculquicondor, line 311: why do we need this intermediate context?

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -467,37 +480,38 @@ func createPods(namespace string, cpo *createPodsOp, clientset clientset.Interfa
...
 		}
 	}
-
 	for _, namespace := range namespaces {
 		wantCount, ok := numPodsScheduledPerNamespace[namespace]
 		if !ok {
 			return fmt.Errorf("unknown namespace %s", namespace)
 		}
-		if err := barrierOne(podInformer, name, namespace, wantCount); err != nil {
+		if err := waitUntilPodsScheduledInNamespace(podInformer, name, namespace, wantCount); err != nil {
```
alculquicondor, line 514: This is going to have a 600s timeout for each namespace. I think we should have a 600s timeout total. If it's too much refactoring to achieve that, leave it.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -286,14 +288,16 @@ func BenchmarkPerfScheduling(b *testing.B) {
 	}
 
 	dataItems := DataItems{Version: "v1"}
+	benchmarkCtx, benchmarkCancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
adtac: Done.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -303,38 +307,42 @@ func BenchmarkPerfScheduling(b *testing.B) {
 	}
 }
 
-func runWorkload(b *testing.B, tc *testCase, w *workload) []DataItem {
-	var mu sync.Mutex
-	var dataItems []DataItem
+func runWorkload(testCtx context.Context, b *testing.B, tc *testCase, w *workload) []DataItem {
```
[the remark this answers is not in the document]
adtac: Done.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: We don't, done.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -303,38 +307,42 @@ func BenchmarkPerfScheduling(b *testing.B) {
...
+				b.Fatalf("op %d: %v", opIndex, err)
 			}
-			if numNodes == 0 {
+			if nextNodeIndex == 0 {
 				// Schedule a cleanup at most once. The CleanupNodes function will list
 				// and delete *all* nodes.
 				// TODO(#93794): make CleanupNodes only clean up its own nodes to make
 				// this more intuitive?
-				defer nodePreparer.CleanupNodes()
+				b.Cleanup(nodePreparer.CleanupNodes)
```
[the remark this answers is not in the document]
adtac: Actually, let's call cleanup every time after PrepareNodes even though everything but the first CleanupNodes does nothing. It's cleaner this way (no pun intended). Done.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: Nice, we only care about one error, so we can drop errors after the first one without blocking.

test/integration/framework/perf_utils.go
```diff
@@ -103,11 +103,12 @@ func (p *IntegrationTestNodePreparer) PrepareNodes(startIndex int) error {
...
 		klog.Fatalf("Error listing nodes: %v", err)
 	}
-	index := startIndex
-	sum := startIndex
+	index := 0
 	for _, v := range p.countToStrategy {
-		sum += v.Count
-		for ; index < sum; index++ {
+		for i := 0; i < v.Count; i, index = i+1, index+1 {
+			if index < nextNodeIndex {
```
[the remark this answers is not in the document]
adtac: we need some sort of counter to keep track of how many nodes we've prepared across multiple strategies, so there would need to be a sum variable like before, which is what we removed after the previous review iteration :)

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: Added a deadline 1800s deadline per workload that is checked at the start of each op. This way, each workload will take at most 1800s (per workload deadline) + 600s (per op deadline) = 2400s in the worst case, which is a sane limit for some of the large 5000 node tests. LMK what you think.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -307,25 +305,33 @@ func BenchmarkPerfScheduling(b *testing.B) {
 	}
 }
 
-func runWorkload(testCtx context.Context, b *testing.B, tc *testCase, w *workload) []DataItem {
-	workloadCtx, subtestCancel := context.WithCancel(testCtx)
-	b.Cleanup(subtestCancel)
+func runWorkload(b *testing.B, tc *testCase, w *workload) []DataItem {
+	// 1800s should be plenty enough even for the 5000-node tests.
+	ctx, cancel := context.WithTimeout(context.Background(), 1800*time.Second)
```
alculquicondor, line 310: 30 * time.Minute?

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -364,7 +366,10 @@ func runWorkload(testCtx context.Context, b *testing.B, tc *testCase, w *workloa
 			}
 			barrierAndCollect := func(idx int) {
 				if err := waitUntilPodsScheduledInNamespace(podInformer, b.Name(), namespace, concreteOp.CreatePods.(int)); err != nil {
-					barrierAndCollectErrors[idx] = err
+					select {
+					case opErrorChan <- opError{opIndex: idx, err: err}:
```
alculquicondor, line 370: I would just do `fmt.Errorf("op %d: %w", idx, err)` to avoid the extra type.

test/integration/framework/perf_utils.go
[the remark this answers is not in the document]
adtac: nvm, Done.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -418,10 +421,10 @@ func runWorkload(testCtx context.Context, b *testing.B, tc *testCase, w *workloa
...
-	for opIndex, err := range barrierAndCollectErrors {
-		if err != nil {
-			b.Fatalf("op %d: %v", opIndex, err)
-		}
+	select {
+	case err := <-opErrorChan:
+		b.Fatalf("op %d: %v", err.opIndex, err.err)
+	default:
 	}
 	return dataItems
```
alculquicondor, line 429: there are no guarantees here that there is no `barrierAndCollect` still running at this point. Maybe we can add a wait group.

test/integration/scheduler_perf/scheduler_perf_test.go
alculquicondor: Sounds good, but you should be looking at that context somewhere in this function.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -343,24 +351,29 @@ func runWorkload(b *testing.B, tc *testCase, w *workload) []DataItem {
 			} else {
 				namespace = fmt.Sprintf("namespace-%d", opIndex)
 			}
-			var stopCh chan struct{}
 			var collectors []testDataCollector
+			var collectorCtx context.Context
+			var collectorCancel func()
 			if concreteOp.CollectMetrics {
-				stopCh = make(chan struct{})
+				collectorCtx, collectorCancel = context.WithCancel(ctx)
```
alculquicondor, line 358: `defer collectorCancel()` in case `barrierAndCollect` fails before calling it.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,300 +56,567 @@ var (
...
+		case *createNodesOp:
+			nodePreparer, err := getNodePreparer(fmt.Sprintf("node-%d-", opIndex), concreteOp, clientset)
+			if err != nil {
+				b.Fatalf("op %d: %v", opIndex, err)
+			}
+			if err := nodePreparer.PrepareNodes(nextNodeIndex); err != nil {
+				b.Fatalf("op %d: %v", opIndex, err)
+			}
+			b.Cleanup(nodePreparer.CleanupNodes)
+			nextNodeIndex += concreteOp.CreateNodes.(int)
```
alculquicondor, line 345: I heavily dislike the casting here. That's why I prefer the separate members `createNodes int` and `createNodesVar string`

test/integration/scheduler_perf/scheduler_perf_test.go
alculquicondor: I misunderstood what you were doing for empty map.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: Done.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: Done.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: Done/.

test/integration/scheduler_perf/config/performance-config.yaml
adtac: Done.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: Done.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: Done. Unfortunately, there's no PollImmediateWithContext so we can't cancel `waitUntilPodsScheduledInNamespace`.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -51,300 +56,567 @@ var (
...
+			}
+			if concreteOp.SkipWaitToCompletion {
+				// Only record those namespaces that may potentially require barriers
+				// in the future.
+				if _, ok := numPodsScheduledPerNamespace[namespace]; ok {
+					numPodsScheduledPerNamespace[namespace] += concreteOp.CreatePods.(int)
+				} else {
+					numPodsScheduledPerNamespace[namespace] = concreteOp.CreatePods.(int)
+				}
+				go barrierAndCollect(opIndex)
```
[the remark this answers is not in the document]
ahg-g: great, much simpler!

test/integration/scheduler_perf/scheduler_perf_test.go
alculquicondor: You could still use the context in the anonymous function and do the exact same thing you are doing in this function.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -101,7 +101,11 @@ type workload struct {
 }
 
 type workloadParams map[string]interface{}
-func (wp workloadParams) getInt(key string) (int, bool) { n, ok := wp[key].(float64); return int(n), ok }
+
+func (wp workloadParams) getInt(key string) (int, bool) {
+	n, ok := wp[key].(float64)
```
alculquicondor, line 106: unmarshal always does so with float64?

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -362,56 +371,47 @@ func runWorkload(b *testing.B, tc *testCase, w *workload) []DataItem {
...
+					numPodsScheduledPerNamespace[namespace] = concreteOp.Count
 				}
-				go barrierAndCollect(opIndex)
 			} else {
-				barrierAndCollect(opIndex)
+				if err := waitUntilPodsScheduledInNamespace(podInformer, b.Name(), namespace, concreteOp.Count); err != nil {
+					b.Fatalf("op %d: error in waiting for pods to get scheduled: %v", opIndex, err)
+				}
+			}
+			if concreteOp.CollectMetrics {
```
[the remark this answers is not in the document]
adtac: Not necessarily. The initial set of pods we schedule need to be scheduled before we can start the measurement pods, so SkipWaitToCompletion isn't set for either set of pods. However, CollectMetrics is only set for the measurement pods.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: Yes, unfortunately. In fact, with `sigs.k8s.io/yaml`, the YAML is converted to JSON, and the JSON is unmarshalled, so I imagine there's some loss of information in the intermediate steps.

alculquicondor: do we need floats? Can we use `map[string]int`?

adtac: we may need bool/string template vars, which is why the map value is an interface{}

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: Done.

test/integration/scheduler_perf/scheduler_perf_test.go
adtac: Switched to int.

test/integration/scheduler_perf/scheduler_perf_test.go
```diff
@@ -480,8 +476,13 @@ func createPods(namespace string, cpo *createPodsOp, clientset clientset.Interfa
 // waitUntilPodsScheduledInNamespace blocks until all pods in the given
 // namespace are scheduled. Times out after 10 minutes because even at the
 // lowest observed QPS of ~10 pods/sec, a 5000-node test should complete.
-func waitUntilPodsScheduledInNamespace(podInformer coreinformers.PodInformer, name string, namespace string, wantCount int) error {
+func waitUntilPodsScheduledInNamespace(ctx context.Context, podInformer coreinformers.PodInformer, name string, namespace string, wantCount int) error {
 	return wait.PollImmediate(1*time.Second, 10*time.Minute, func() (bool, error) {
+		select {
+		case <-ctx.Done():
+			return true, nil
```
alculquicondor, line 483: `return true, ctx.Err()`?

adtac: No, returning an error would cause wait.PollImmediate to try again after 1 second.

adtac: Actually, nevermind, you're right :D I really thought returning error would poll again.

Done.

alculquicondor:
/approve

If @ahg-g doesn't have anything to add, you can squash

ahg-g:
Looks good, thanks Adi!

One immediate followup is to delete the legacy benchmark (the very first one we had): https://github.com/kubernetes/kubernetes/blob/master/test/integration/scheduler_perf/scheduler_perf_legacy_test.go

adtac:
>One immediate followup is to delete the legacy benchmark (the very first one we had): https://github.com/kubernetes/kubernetes/blob/master/test/integration/scheduler_perf/scheduler_perf_legacy_test.go

ack, will do in a follow up PR :+1:

alculquicondor:
/assign @liggitt 
for test

/priority important-longterm

test/utils/runners.go
```diff
@@ -931,8 +931,8 @@ type CountToStrategy struct {
 }
 
 type TestNodePreparer interface {
-	PrepareNodes() error
-	CleanupNodes() error
+	PrepareNodes(nextNodeIndex int) error
+	CleanupNodes()
```
liggitt, line 935: why was the error dropped? asymmetry on prepare/cleanup is unexpected. is there another way the cleanup can fail the test if unexpected errors are encountered?

adtac: The TestNodePreparer interface's implementations are pending a refactor (see #93794). However, there's no reason why the interface's CleanupNodes can't be left unchanged for now. Done :+1:

test/integration/framework/perf_utils.go
```diff
@@ -119,13 +117,17 @@ func (p *IntegrationTestNodePreparer) PrepareNodes() error {
 
 // CleanupNodes deletes existing test nodes.
 func (p *IntegrationTestNodePreparer) CleanupNodes() error {
+	// TODO(#93794): make CleanupNodes only clean up the nodes created by this
+	// IntegrationTestNodePreparer to make this more intuitive.
 	nodes, err := GetReadySchedulableNodes(p.client)
 	if err != nil {
 		klog.Fatalf("Error listing nodes: %v", err)
+		return err
```
liggitt, line 125: return after a Fatalf isn't going to do anything :-)

test/integration/framework/perf_utils.go
```diff
@@ -119,13 +117,17 @@ func (p *IntegrationTestNodePreparer) PrepareNodes() error {
...
 	}
 	for i := range nodes.Items {
 		if err := p.client.CoreV1().Nodes().Delete(context.TODO(), nodes.Items[i].Name, metav1.DeleteOptions{}); err != nil {
 			klog.Errorf("Error while deleting Node: %v", err)
+			return err
```
liggitt, line 130: continue iterating over nodes and accumulate errors to return so we clean up as much as possible

test/integration/framework/perf_utils.go
adtac: Ah :) Done.

test/integration/framework/perf_utils.go
adtac: Done.

alculquicondor:
/lgtm

nice job @adtac 
Thanks for the conscious review @liggitt

PR #93254, opened 2020-07-20
WIP: Make E2E tests support providerless

lorenz:
/kind feature

This makes it possible to build the E2E test suite without all of the provider-specific dependencies & code. It lowers the Bazel-built binary size from ~130MiB to ~80MiB and reduces the amount of non-K8s files included by around 1000.

This is marked as WIP because it may require a KEP and some additional cleanup (like dealing with the import restriction files).

```release-note
NONE
```


lorenz:
CLA issues should be fixed

test/e2e/network/service.go
```diff
@@ -1191,375 +1185,6 @@ var _ = SIGDescribe("Services", func() {
 		framework.ExpectNoError(err)
 	})
 
-	// TODO: Get rid of [DisabledForLargeClusters] tag when issue #56138 is fixed.
-	ginkgo.It("should be able to change the type and ports of a service [Slow] [DisabledForLargeClusters]", func() {
```
ZhiFeng1993, line 1195: Why getting rid of this test and the following one?

test/e2e/storage/nfs_persistent_volume-disruptive.go
```diff
@@ -91,7 +91,7 @@ var _ = utils.SIGDescribe("NFSPersistentVolumes[Disruptive][Flaky]", func() {
 
 	ginkgo.BeforeEach(func() {
 		// To protect the NFS volume pod from the kubelet restart, we isolate it on its own node.
-		e2eskipper.SkipUnlessNodeCountIsAtLeast(minNodes)
+		e2eskipper.SkipUnlessNodeCountIsAtLeast(2)
```
ZhiFeng1993, line 94: Why do we use magic number 2 here?

PR #93258, opened 2020-07-20
mv TokenRequest and TokenRequestProjection to GA

zshihang:
**What type of PR is this?**

/kind api-change

move `TokenRequest` and `TokenRequestProjection` features to GA in v1.20

```release-note
+ `TokenRequest` and `TokenRequestProjection` features have been promoted to GA. This feature allows generating service account tokens that are not visible in Secret objects and are tied to the lifetime of a Pod object. See https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection for details on configuring and using this feature. The `TokenRequest` and `TokenRequestProjection` feature gates will be removed in v1.21.
+ kubeadm's kube-apiserver Pod manifest now includes the following flags by default "--service-account-key-file", "--service-account-signing-key-file", "--service-account-issuer".
```

```docs
- [KEP]: https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/1205-bound-service-account-tokens/README.md
```


mikedanese:
@kubernetes/sig-auth-pr-reviews

pkg/features/kube_features.go
```diff
@@ -677,8 +679,8 @@ var defaultKubernetesFeatureGates = map[featuregate.Feature]featuregate.FeatureS
 	SupportPodPidsLimit:            {Default: true, PreRelease: featuregate.Beta},
 	SupportNodePidsLimit:           {Default: true, PreRelease: featuregate.Beta},
 	HyperVContainer:                {Default: false, PreRelease: featuregate.Alpha},
-	TokenRequest:                   {Default: true, PreRelease: featuregate.Beta},
-	TokenRequestProjection:         {Default: true, PreRelease: featuregate.Beta},
+	TokenRequest:                   {Default: true, PreRelease: featuregate.GA, LockToDefault: true}, // remove in 1.21
+	TokenRequestProjection:         {Default: true, PreRelease: featuregate.GA, LockToDefault: true}, // remove in 1.21
```
liggitt, line 683: should the presence of the tokenrequest API (and associated startup flag config) be required as part of this? if so, we'll need a conformance test for the tokenrequest API.

zshihang: done

cmd/kubeadm/app/phases/controlplane/manifests.go
```diff
@@ -132,20 +132,22 @@ func CreateStaticPodFiles(manifestDir, patchesDir string, cfg *kubeadmapi.Cluste
...
-		"secure-port":                     fmt.Sprintf("%d", localAPIEndpoint.BindPort),
-		"allow-privileged":                "true",
-		"kubelet-preferred-address-types": "InternalIP,ExternalIP,Hostname",
+		"advertise-address":                localAPIEndpoint.AdvertiseAddress,
+		"insecure-port":                    "0",
+		"enable-admission-plugins":         "NodeRestriction",
+		"service-cluster-ip-range":         cfg.Networking.ServiceSubnet,
+		"service-account-key-file":         filepath.Join(cfg.CertificatesDir, kubeadmconstants.ServiceAccountPublicKeyName),
+		"service-account-signing-key-file": filepath.Join(cfg.CertificatesDir, kubeadmconstants.ServiceAccountPrivateKeyName),
+		"service-account-issuer":           "https://kubernetes.default.svc",
```
liggitt, line 141: will defer to @neolit123 if this should be a constant

plugin/pkg/admission/noderestriction/admission_test.go
```diff
@@ -56,15 +56,12 @@ var (
 func init() {
 	// all features need to be set on all featuregates for the tests.  We set everything and then then the if's below override it.
 	relevantFeatures := map[featuregate.Feature]featuregate.FeatureSpec{
-		features.TokenRequest:            {Default: false},
 		features.CSINodeInfo:             {Default: false},
 		features.ExpandPersistentVolumes: {Default: false},
 	}
-	utilruntime.Must(trEnabledFeature.Add(relevantFeatures))
```
liggitt, line 63: drop `trEnabledFeature` and change test cases which referenced that to not set an explicit featuregate??

test/e2e/auth/service_accounts.go
```diff
@@ -418,6 +419,59 @@ var _ = SIGDescribe("ServiceAccounts", func() {
...
+		mounttestArgs := []string{
+			"mounttest",
+			fmt.Sprintf("--file_content=%v", tokenVolumePath),
+		}
+
+		pod := e2epod.NewAgnhostPod(f.Namespace.Name, podName, volumes, volumeMounts, nil, mounttestArgs...)
+		pod.Spec.RestartPolicy = v1.RestartPolicyNever
+
+		output := []string{
+			fmt.Sprintf("content of file \"%v\": %s", tokenVolumePath, ".+"),
```
liggitt, line 469: `.+` isn't a very strong test... maybe something that shows it was a JWT?

test/e2e/auth/service_accounts.go
```diff
@@ -539,7 +581,7 @@ var _ = SIGDescribe("ServiceAccounts", func() {
 		}
 	})
 
-	ginkgo.It("should support InClusterConfig with token rotation [Slow] [Feature:TokenRequestProjection]", func() {
+	ginkgo.It("should support InClusterConfig with token rotation [Slow]", func() {
```
liggitt, line 584: do you know what test suite this runs in (if any)?

liggitt (commented):
largely LGTM

cmd/kubeadm/app/phases/controlplane/manifests.go
neolit123: it's fine to leave it like so for now, we can refactor later.

test/e2e/auth/service_accounts.go
zshihang: `pull-kubernetes-e2e-kind` will run this when removed the `[Feature:TokenRequestProjection]`.

liggitt: `Slow` tests get skipped in presubmit

https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml#L32

zshihang: right. i meant `ginkgo.It("should set ownership and permission when RunAsUser or FsGroup is present [LinuxOnly] [NodeFeature:FSGroup] [Feature:TokenRequestProjection]"` and the newly added one.

i'll drop `Slow` because it is not that slow when i was testing it.

zshihang: if remove the `Slow` then we might not need the newly added e2e test

liggitt: > i'll drop `Slow` because it is not that slow when i was testing it.

it takes at least 10 minutes to observe a rotation, right? that's... slow

cmd/kubeadm/app/phases/controlplane/manifests.go
```diff
@@ -200,6 +202,7 @@ func getAPIServerCommand(cfg *kubeadmapi.ClusterConfiguration, localAPIEndpoint
 	if cfg.APIServer.ExtraArgs == nil {
 		cfg.APIServer.ExtraArgs = map[string]string{}
 	}
+
```
neolit123, line 205: we can remove this whitespace line.

pkg/kubeapiserver/options/authentication.go
```diff
@@ -464,14 +460,13 @@ func (o *BuiltInAuthenticationOptions) ApplyTo(authInfo *genericapiserver.Authen
 		authInfo.APIAudiences = authenticator.Audiences{o.ServiceAccounts.Issuer}
```
liggitt, line 460: I just noticed the "`" quotes around the example in the service-account-issuer help make it render weirdly.

Drop the "`" quotes wrapping {service-account-issuer}/.well-known/openid-configuration above

test/e2e/auth/service_accounts.go
zshihang: i think it will show up in https://k8s-testgrid.appspot.com/sig-auth-gce#gce-slow after remove this

test/e2e/auth/service_accounts.go
zshihang: done

plugin/pkg/admission/noderestriction/admission_test.go
zshihang: done

cmd/kubeadm/app/phases/controlplane/manifests.go
zshihang: done

cmd/kubeadm/app/phases/controlplane/manifests.go
liggitt, line 141: This value as the default is the last open question I have. Continuing the conversation started in https://github.com/kubernetes/kubernetes/pull/95667/files#r510954650

@liggitt:

> I expected this to be the public address of the API server. I think the test value set in local-up-cluster is not one we'd use in a real server. @mikedanese, do we have good guidance for what to set this to? 
> 
> Because server names can change and people can make mistakes and want to fix them, we also need to make sure it is possible for users to modify the issuer. I think the workflow for that looks like:
> 1. Explicitly set `--api-audiences` to include both the old and new issuer values
> 2. If you have multiple API servers, roll out the `--api-audiences` change
> 3. Modify the `--service-account-issuer` to the new value
> 4. Roll out the updated API manifest
> 
> I don't know how these default values propagate into something user-modifiable. cc @neolit123 for insight there

@neolit123:

> > I expected this to be the public address of the API server
> 
> if this should be the address, then we should set `service-account-issuer` to the same value in:
> `"advertise-address":                localAPIEndpoint.AdvertiseAddress,`
> 
> would it be fine that separate instances of the kube-apiserver in an HA setup to have separate value for `service-account-issuer` (if this is based on the address)?
> 
> > I don't know how these default values propagate into something user-modifiable. cc @neolit123 for insight there
> 
> nowadays, kubeadm allows patching the manifest files before the kubelet starts on control-plane nodes (and creates the static pods), so technically users can patch the manifests with the flag changes they want to. it's also possible to use "phases", i.e. create manifests, edit manifests with text editor, start the pods later.


@zshihang:

> issuer is used as `iss` claim in the token. it would be a URL. https://tools.ietf.org/html/rfc7519#section-4.1.1
> 
> > server names can change and people can make mistakes and want to fix them
> 
> users can specify `api-audiences` if they doing the audiences change.
> 
> also, will also allow users to pass in `service-account-issuer` and the default will still be `https://kubernetes.default.svc`.

liggitt: Digging deeper, I was wrong. We don't currently support changing issuer in a way that allows tokens from the previous issuer to still be considered valid.

I have two concerns with using https://kubernetes.default.svc:

* `svc` is not currently a TLD but could become one in the future, at which point clients that tried to resolve `https://kubernetes.default.svc/.well-known/openid-configuration` would fetch validation data from some random server. This could be mitigated in this PR by defaulting this to `"https://kubernetes.default.svc."+cfg.DNSDomain` which would default to "https://kubernetes.default.svc.cluster.local" which is not resolvable.

* It is not currently possible to change the issuer in a non-disruptive way. This could be mitigated in the future by kube-apiserver by adding support for accepting multiple issuers. It seems likely we'd want to do that.

@neolit123 can you verify that a kubeadm user that wants to override just the `--service-account-issuer` arg is able to do so in a layered way without taking full ownership over the manifests?

neolit123: > svc is not currently a TLD but could become one in the future, at which point clients that tried to resolve https://kubernetes.default.svc/.well-known/openid-configuration would fetch validation data from some random server. This could be mitigated in this PR by defaulting this to "https://kubernetes.default.svc."+cfg.DNSDomain which would default to "https://kubernetes.default.svc.cluster.local" which is not resolvable.

that would be a problem, yes. passing `"https://kubernetes.default.svc.cluster.local"` by default for the time being seems good. the `cluster.local` defaulted DNS field is under [`cfg.Networking.DNSDomain`](https://godoc.org/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta2#Networking)

> @neolit123 can you verify that a kubeadm user that wants to override just the --service-account-issuer arg is able to do so in a layered way without taking full ownership over the manifests?

yes, they could pass this to `kubeadm init` (on cluster creation):
```
apiVersion: ...
kind: ClusterConfiguration
apiServer:
  extraArgs:
    service-account-issuer: "https://some-URL"
```

which would override the default of `https://kubernetes.default.svc.cluster.local`

zshihang: how about just set it as `https://kubernetes.default.svc.cluster.local`? users can overwrite anyway.

liggitt:
@zshihang go ahead and open a WIP docs PR agains the https://github.com/kubernetes/website/tree/dev-1.20 branch to update https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection with the `--api-audiences` flag and the GA status.

zshihang:
regarding the kubeadm, the only diff is the default value for issuer and necessary changes to pass unit test.

liggitt:
/hold cancel

based on https://github.com/kubernetes/kubernetes/pull/93258#issuecomment-719015364 and https://github.com/kubernetes/kubernetes/pull/93258#discussion_r514639235

neolit123:
Lgtm. Thanks for the updates.

kikisdeliveryservice:
Q: Is this related to an enhancement?

neolit123:
@kikisdeliveryservice 

hi, yes this is for:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/1205-bound-service-account-tokens/README.md

@zshihang 

could you please add the `docs` block as seen at the bottom here (the `docs` type is needed around the block):
https://raw.githubusercontent.com/kubernetes/kubernetes/master/.github/PULL_REQUEST_TEMPLATE.md

```docs
- [KEP]: https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/1205-bound-service-account-tokens/README.md
```

also could you please add some sort of a mention about kubeadm in the release-note:
> kubeadm's kube-apiserver Pod manifest now includes the following flags by default "--service-account-key-file", "--service-account-signing-key-file", "--service-account-issuer"

zshihang:
@neolit123 added. thanks.

kikisdeliveryservice:
Hi all,

I don't see any open (or closed) issue 1205 nor is this enhancement tracked for 1.20 in the spreadsheet: http://bit.ly/k8s-1-20-enhancements

Can someone please explain what the tracking issue for this is?  As this was not tracked and there was no underlying tracking issue an exception will need to be filed. See: https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#exceptions

Thanks,
Kirsten
1.20 Enhancements Lead

liggitt:
Apologies for the disconnect. The KEP is at https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/1205-bound-service-account-tokens and was updated with v1.20 plans prior to enhancements freeze in https://github.com/kubernetes/enhancements/pull/1912

The KEP number came from the original PR that converted a pre-KEP design to KEP format (https://github.com/kubernetes/enhancements/pull/1205)

kikisdeliveryservice:
Hi @liggitt 

There is no tracking issue (which is required for all enhancements), and the enhancement is not being tracked by the release team for 1.20. The KEP is also misnumbered as the naming should be <IssueNumber>-<EnhancementTitle> as per the templates found here: https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template

The requirement to have a 1.20 milestoned issue was also mentioned in my email on October 5 sent to k-dev: https://groups.google.com/g/kubernetes-dev/c/1KF6mmO1WsQ/m/IFWMkeKVCwAJ 

Please open an exception request ASAP.

Thanks,
Kirsten

liggitt:
> There is no tracking issue (which is required for all enhancements), and the enhancement is not being tracked by the release team for 1.20. The KEP is also misnumbered as the naming should be - as per the templates found here: https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template

Understood. The template/process has changed since this KEP was created, and work to sweep existing designs/KEPs that predated those changes has been inconsistent. This highlights a gap in the release process that relies on keeping multiple disconnected information sources (kep.yaml files, enhancements issues, tracking spreadsheet, etc) in sync manually.

We'll note the exception.

PR #93260, opened 2020-07-20
Fix ECR provider startup latency

### LINKED ISSUE #92162: aws credentials provider causes ~10s-20s hangup on kubelet's initializing docker runtime
Following commit in 1.15 on aws credentials provider  causes ~10s-20s delay(kubelet main goroutine hangup) on kubelet's initializing docker runtime for nodes of K8s clusters not running on AWS:

[Refactors and fixes bugs in AWS credentialprovider - 11efc013288a](https://github.com/kubernetes/kubernetes/commit/11efc013288a15322a6abc0852a4859f7b0785dc)

We observed **20s delay** on our 1.16 production nodes, while on local clusters, the delay is about **9s to 14s**

This commit failed to check whether the node is running on AWS or not before calling AWS APIs. 

Please add this check like other providers, such as GCP,Azure. Thanks.

GCP: https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/gcp/metadata.go#L146
Azure: https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/azure/azure_credentials.go#L157

# Test results
## add trace to pkg/credentialprovider/aws/aws_credentials.go
```
func (p *ecrProvider) Enabled() bool {
	klog.Info("aws_credential:1")
	sess, err := session.NewSessionWithOptions(session.Options{
		SharedConfigState: session.SharedConfigEnable,
	})
	klog.Info("aws_credential:2")
	if err != nil {
		klog.Errorf("while validating AWS credentials %v", err)
		return false
	}
	klog.Info("aws_credential:3")
	if _, err := sess.Config.Credentials.Get(); err != nil {
		klog.Info("aws_credential:4")
		klog.Errorf("while getting AWS credentials %v", err)
		return false
	}
	klog.Info("aws_credential:5")
	return true
}
```
## our 1.16 production node - 20s delay
```
Jun 16 10:34:59 VM-32-46-ubuntu kubelet[2749059]: I0616 10:34:59.602839 2749059 aws_credentials.go:69] aws_credential:1
Jun 16 10:34:59 VM-32-46-ubuntu kubelet[2749059]: I0616 10:34:59.602926 2749059 aws_credentials.go:73] aws_credential:2
Jun 16 10:34:59 VM-32-46-ubuntu kubelet[2749059]: I0616 10:34:59.602931 2749059 aws_credentials.go:78] aws_credential:3
Jun 16 10:35:19 VM-32-46-ubuntu kubelet[2749059]: I0616 10:35:19.972128 2749059 aws_credentials.go:80] aws_credential:4
Jun 16 10:35:19 VM-32-46-ubuntu kubelet[2749059]: E0616 10:35:19.972153 2749059 aws_credentials.go:81] while getting AWS credentials NoCredentialProviders: no valid providers in chain. Deprecated.
```
### master - 14s
```
I0616 09:50:28.031310  193270 aws_credentials.go:69] aws_credential:1
I0616 09:50:28.031401  193270 aws_credentials.go:73] aws_credential:2
I0616 09:50:28.031408  193270 aws_credentials.go:78] aws_credential:3
I0616 09:50:42.004205  193270 aws_credentials.go:80] aws_credential:4
E0616 09:50:42.004229  193270 aws_credentials.go:81] while getting AWS credentials NoCredentialProviders: no valid providers in chain. Deprecated.
```
### 1.18 - 11s
```
I0616 10:05:18.584844  270482 aws_credentials.go:69] aws_credential:1
I0616 10:05:18.584996  270482 aws_credentials.go:73] aws_credential:2
I0616 10:05:18.585003  270482 aws_credentials.go:78] aws_credential:3
I0616 10:05:29.799597  270482 aws_credentials.go:80] aws_credential:4
E0616 10:05:29.799613  270482 aws_credentials.go:81] while getting AWS credentials NoCredentialProviders: no valid providers in chain. Deprecated.
```
### 1.16 - ~9s
```
I0616 10:15:01.767194   49200 aws_credentials.go:69] aws_credential:1
I0616 10:15:01.767269   49200 aws_credentials.go:73] aws_credential:2
I0616 10:15:01.767273   49200 aws_credentials.go:78] aws_credential:3
I0616 10:15:10.436312   49200 aws_credentials.go:80] aws_credential:4
E0616 10:15:10.436331   49200 aws_credentials.go:81] while getting AWS credentials NoCredentialProviders: no valid providers in chain. Deprecated.
```

nckturner:
### Background
* Before this change, even on non-AWS platforms, the Enabled() check attempts
  to make calls to the metadata endpoint when the session and credentials
  are initialized (in order to determine if the provider should be
  initialized at all).
* This can cause latency because the SDK times out and retries -- up to
  20 seconds of latency has been observed on non-AWS platforms when the
  metadata IP was blocked with an iptables rule.
* Instead, check once if we are running on an EC2 platform, first trying
  to find the EC2 UUID in system files, and second attempting to get
  credentials.
* Add a benchmark test that includes intialization and the credential
  check.

#### GCE, k8s v1.16.1
```
nic@kubernetes-build:~/kubernetes/pkg/credentialprovider/aws$ go test -v -bench=. -run=SetupLatency
goos: linux
goarch: amd64
pkg: k8s.io/kubernetes/pkg/credentialprovider/aws
cpu: Intel(R) Xeon(R) CPU @ 2.30GHz
BenchmarkSetupLatency
E0312 22:29:38.488946   86306 aws_credentials.go:77] while getting AWS credentials NoCredentialProviders: no valid providers in chain. Deprecated.
        For verbose messaging see aws.Config.CredentialsChainVerboseErrors
[...]
BenchmarkSetupLatency-8         1000000000               0.0006955 ns/op
PASS
ok      k8s.io/kubernetes/pkg/credentialprovider/aws    0.032s

```

#### GCE, k8s v1.16.1, with metadata blocked
```
nic@kubernetes-build:~/kubernetes/pkg/credentialprovider/aws$ sudo iptables --insert OUTPUT 1 --destination 169.254.169.254/32 --jump DROP
nic@kubernetes-build:~/kubernetes/pkg/credentialprovider/aws$ go test -v -bench=. -run=SetupLatency
goos: linux
goarch: amd64
pkg: k8s.io/kubernetes/pkg/credentialprovider/aws
cpu: Intel(R) Xeon(R) CPU @ 2.30GHz
BenchmarkSetupLatency
E0312 17:57:50.922473   79323 aws_credentials.go:77] while getting AWS credentials NoCredentialProviders: no valid providers in chain. Deprecated.
        For verbose messaging see aws.Config.CredentialsChainVerboseErrors
BenchmarkSetupLatency-8                1        20308753891 ns/op
PASS
ok      k8s.io/kubernetes/pkg/credentialprovider/aws    20.330s

```

#### GCE, k8s HEAD
```
nic@kubernetes-build:~/kubernetes/pkg/credentialprovider/aws$ go test -v -bench=. -run=SetupLatency
goos: linux
goarch: amd64
pkg: k8s.io/kubernetes/pkg/credentialprovider/aws
cpu: Intel(R) Xeon(R) CPU @ 2.30GHz
BenchmarkSetupLatency
E0312 22:02:59.904691   85420 aws_credentials.go:77] while getting AWS credentials NoCredentialProviders: no valid providers in chain. Deprecated.
        For verbose messaging see aws.Config.CredentialsChainVerboseErrors
[...]
        For verbose messaging see aws.Config.CredentialsChainVerboseErrors
BenchmarkSetupLatency-8         1000000000               0.2403 ns/op
PASS
ok      k8s.io/kubernetes/pkg/credentialprovider/aws    4.542s

```
#### GCE, k8s HEAD, with metadata blocked
```
nic@kubernetes-build:~/kubernetes/pkg/credentialprovider/aws$ sudo iptables --insert OUTPUT 1 --destination 169.254.169.254/32 --jump DROP
nic@kubernetes-build:~/kubernetes/pkg/credentialprovider/aws$ go test -v -bench=. -run=SetupLatency
goos: linux
goarch: amd64
pkg: k8s.io/kubernetes/pkg/credentialprovider/aws
cpu: Intel(R) Xeon(R) CPU @ 2.30GHz
BenchmarkSetupLatency
E0312 22:01:36.837444   85274 aws_credentials.go:77] while getting AWS credentials NoCredentialProviders: no valid providers in chain. Deprecated.
        For verbose messaging see aws.Config.CredentialsChainVerboseErrors
BenchmarkSetupLatency-8                1        6291492407 ns/op
PASS
ok      k8s.io/kubernetes/pkg/credentialprovider/aws    6.309s
```

#### GCE, this commit
```
nic@kubernetes-build:~/kubernetes/pkg/credentialprovider/aws$ go test -v -bench=. -run=SetupLatency
goos: linux
goarch: amd64
pkg: k8s.io/kubernetes/pkg/credentialprovider/aws
cpu: Intel(R) Xeon(R) CPU @ 2.30GHz
BenchmarkSetupLatency
BenchmarkSetupLatency-8         1000000000               0.0000006 ns/op
PASS
ok      k8s.io/kubernetes/pkg/credentialprovider/aws    0.026s
```

#### GCE, this commit, with metadata blocked
```
nic@kubernetes-build:~/kubernetes/pkg/credentialprovider/aws$ sudo iptables --insert OUTPUT 1 --destination 169.254.169.254/32 --jump DROP
nic@kubernetes-build:~/kubernetes/pkg/credentialprovider/aws$ go test -v -bench=. -run=SetupLatency
goos: linux
goarch: amd64
pkg: k8s.io/kubernetes/pkg/credentialprovider/aws
cpu: Intel(R) Xeon(R) CPU @ 2.30GHz
BenchmarkSetupLatency
BenchmarkSetupLatency-8         1000000000               0.0000005 ns/op
PASS
ok      k8s.io/kubernetes/pkg/credentialprovider/aws    0.018s
```

**What type of PR is this?**
>

/kind bug

Reduce latency introduced by the ECR credential provider, even on non-AWS platforms.

Fixes #92162

```release-note
Reduces delay initializing on non-AWS platforms docker runtime.
```

```docs

```

### DIFF
--- a/pkg/credentialprovider/aws/aws_credentials.go
+++ b/pkg/credentialprovider/aws/aws_credentials.go
@@ -20,7 +20,9 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"net/url"
+	"os"
 	"regexp"
 	"strings"
 	"sync"
@@ -38,13 +40,19 @@ import (
 	"k8s.io/kubernetes/pkg/credentialprovider"
 )
 
-var ecrPattern = regexp.MustCompile(`^(\d{12})\.dkr\.ecr(\-fips)?\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.amazonaws\.com(\.cn)?$`)
+var (
+	ecrPattern = regexp.MustCompile(`^(\d{12})\.dkr\.ecr(\-fips)?\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.amazonaws\.com(\.cn)?$`)
+	once       sync.Once
+	isEC2      bool
+)
 
 // init registers a credential provider for each registryURLTemplate and creates
 // an ECR token getter factory with a new cache to store token getters
 func init() {
 	credentialprovider.RegisterCredentialProvider("amazon-ecr",
-		newECRProvider(&ecrTokenGetterFactory{cache: make(map[string]tokenGetter)}))
+		newECRProvider(&ecrTokenGetterFactory{cache: make(map[string]tokenGetter)},
+			ec2ValidationImpl,
+		))
 }
 
 // ecrProvider is a DockerConfigProvider that gets and refreshes tokens
@@ -52,20 +60,74 @@ func init() {
 type ecrProvider struct {
 	cache         cache.Store
 	getterFactory tokenGetterFactory
+	isEC2         ec2ValidationFunc
 }
 
 var _ credentialprovider.DockerConfigProvider = &ecrProvider{}
 
-func newECRProvider(getterFactory tokenGetterFactory) *ecrProvider {
+func newECRProvider(getterFactory tokenGetterFactory, isEC2 ec2ValidationFunc) *ecrProvider {
 	return &ecrProvider{
 		cache:         cache.NewExpirationStore(stringKeyFunc, &ecrExpirationPolicy{}),
 		getterFactory: getterFactory,
+		isEC2:         isEC2,
 	}
 }
 
-// Enabled implements DockerConfigProvider.Enabled. Enabled is true if AWS
-// credentials are found.
+// Enabled implements DockerConfigProvider.Enabled.
 func (p *ecrProvider) Enabled() bool {
+	return true
+}
+
+type ec2ValidationFunc func() bool
+
+// ec2ValidationImpl returns true if we detect
+// an EC2 vm based on checking for the EC2 system UUID, the asset tag (for nitro
+// instances), or instance credentials if the UUID is not present.
+// Ref: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html
+func ec2ValidationImpl() bool {
+	return tryValidateEC2UUID() || tryValidateEC2Creds()
+}
+
+func tryValidateEC2UUID() bool {
+	hypervisor_uuid := "/sys/hypervisor/uuid"
+	product_uuid := "/sys/devices/virtual/dmi/id/product_uuid"
+	asset_tag := "/sys/devices/virtual/dmi/id/board_asset_tag"
+
+	if _, err := os.Stat(hypervisor_uuid); err == nil {
+		b, err := ioutil.ReadFile(hypervisor_uuid)
+		if err != nil {
+			klog.Errorf("error checking if this is an EC2 instance: %v", err)
+		} else if strings.HasPrefix(string(b), "EC2") || strings.HasPrefix(string(b), "ec2") {
+			klog.V(5).Infof("found 'ec2' in uuid %v from %v, enabling legacy AWS credential provider", string(b), hypervisor_uuid)
+			return true
+		}
+	}
+
+	if _, err := os.Stat(product_uuid); err == nil {
+		b, err := ioutil.ReadFile(product_uuid)
+		if err != nil {
+			klog.Errorf("error checking if this is an EC2 instance: %v", err)
+		} else if strings.HasPrefix(string(b), "EC2") || strings.HasPrefix(string(b), "ec2") {
+			klog.V(5).Infof("found 'ec2' in uuid %v from %v, enabling legacy AWS credential provider", string(b), product_uuid)
+			return true
+		}
+	}
+
+	if _, err := os.Stat(asset_tag); err == nil {
+		b, err := ioutil.ReadFile(asset_tag)
+		s := strings.TrimSpace(string(b))
+		if err != nil {
+			klog.Errorf("error checking if this is an EC2 instance: %v", err)
+		} else if strings.HasPrefix(s, "i-") && len(s) == 19 {
+			// Instance ID's are 19 characters plus newline
+			klog.V(5).Infof("found instance ID in %v from %v, enabling legacy AWS credential provider", string(b), asset_tag)
+			return true
+		}
+	}
+	return false
+}
+
+func tryValidateEC2Creds() bool {
 	sess, err := session.NewSessionWithOptions(session.Options{
 		SharedConfigState: session.SharedConfigEnable,
 	})
@@ -77,6 +139,7 @@ func (p *ecrProvider) Enabled() bool {
 		klog.Errorf("while getting AWS credentials %v", err)
 		return false
 	}
+	klog.V(5).Infof("found aws credentials, enabling legacy AWS credential provider")
 	return true
 }
 
@@ -85,12 +148,25 @@ func (p *ecrProvider) Enabled() bool {
 func (p *ecrProvider) Provide(image string) credentialprovider.DockerConfig {
 	parsed, err := parseRepoURL(image)
 	if err != nil {
-		klog.V(3).Info(err)
+		return credentialprovider.DockerConfig{}
+	}
+
+	// To prevent the AWS SDK from causing latency on non-aws platforms, only test if we are on
+	// EC2 or have access to credentials once.  Attempt to do it without network calls by checking
+	// for certain EC2-specific files.  Otherwise, we ask the SDK to initialize a session to see if
+	// credentials are available.  On non-aws platforms, especially when a metadata endpoint is blocked,
+	// this has been shown to cause 20 seconds of latency due to SDK retries
+	// (see https://github.com/kubernetes/kubernetes/issues/92162)
+	once.Do(func() {
+		isEC2 = p.isEC2()
+	})
+
+	if !isEC2 {
 		return credentialprovider.DockerConfig{}
 	}
 
 	if cfg, exists := p.getFromCache(parsed); exists {
-		klog.V(6).Infof("Got ECR credentials from cache for %s", parsed.registry)
+		klog.V(3).Infof("Got ECR credentials from cache for %s", parsed.registry)
 		return cfg
 	}
 	klog.V(3).Info("unable to get ECR credentials from cache, checking ECR API")
--- a/pkg/credentialprovider/aws/aws_credentials_test.go
+++ b/pkg/credentialprovider/aws/aws_credentials_test.go
@@ -154,6 +154,10 @@ func TestParseRepoURLFail(t *testing.T) {
 	}
 }
 
+func isAlwaysEC2() bool {
+	return true
+}
+
 func TestECRProvide(t *testing.T) {
 	registry := "123456789012.dkr.ecr.lala-land-1.amazonaws.com"
 	otherRegistries := []string{
@@ -168,7 +172,7 @@ func TestECRProvide(t *testing.T) {
 			password: password,
 			endpoint: registry,
 		},
-	})
+	}, isAlwaysEC2)
 	keyring := &credentialprovider.BasicDockerKeyring{}
 	keyring.Add(p.Provide(image))
 
@@ -213,7 +217,7 @@ func TestECRProvideCached(t *testing.T) {
 			endpoint:          registry,
 			randomizePassword: true,
 		},
-	})
+	}, isAlwaysEC2)
 	image1 := path.Join(registry, "foo/bar")
 	image2 := path.Join(registry, "bar/baz")
 	keyring := &credentialprovider.BasicDockerKeyring{}
@@ -266,7 +270,7 @@ func TestChinaECRProvide(t *testing.T) {
 			password: password,
 			endpoint: registry,
 		},
-	})
+	}, isAlwaysEC2)
 	keyring := &credentialprovider.BasicDockerKeyring{}
 	keyring.Add(p.Provide(image))
 	// Verify that we get the expected username/password combo for
@@ -310,7 +314,7 @@ func TestChinaECRProvideCached(t *testing.T) {
 			endpoint:          registry,
 			randomizePassword: true,
 		},
-	})
+	}, isAlwaysEC2)
 	image := path.Join(registry, "foo/bar")
 	keyring := &credentialprovider.BasicDockerKeyring{}
 	keyring.Add(p.Provide(image))
@@ -330,4 +334,9 @@ func TestChinaECRProvideCached(t *testing.T) {
 	if creds[0].Password != creds[1].Password {
 		t.Errorf("cached credentials do not match")
 	}
+}
+
+func BenchmarkSetupLatency(b *testing.B) {
+	p := newECRProvider(&ecrTokenGetterFactory{cache: make(map[string]tokenGetter)}, ec2ValidationImpl)
+	_ = p.Enabled()
 }


nckturner:
Need to do some testing.
/hold

therc:
> Need to do some testing.
> /hold

I think this is bringing the code back to my original implementation, so it should work (I checked that it wouldn't introduce delays outside of AWS), unless there's something else that changed substantially in the credential providers. I can approve in the meantime, but please proceed with more testing.

nckturner:
Did some testing on AWS platforms to make sure this change does not affect image pull for both credentialed (ECR) and non-credentialed pulls, and it looks fine.

BenTheElder:
beep boop @nckturner @andrewsykim :upside_down_face:

andrewsykim:
On second thought, this could be misleading because kubelet will always attempt to authenticate to ECR repos but `Provide` will fail if there's no session established. Could `Enabled` instead check the availablility of the metadata server and/or the aws credentials on disk `~/.aws/credentials`?

andrewsykim:
FYI @liggitt

pkg/credentialprovider/aws/aws_credentials.go
```diff
@@ -66,17 +66,6 @@ func newECRProvider(getterFactory tokenGetterFactory) *ecrProvider {
...
-	})
-	if err != nil {
-		klog.Errorf("while validating AWS credentials %v", err)
-		return false
-	}
-	if _, err := sess.Config.Credentials.Get(); err != nil {
-		klog.Errorf("while getting AWS credentials %v", err)
-		return false
-	}
 	return true
```
liggitt, line 69: this means that Provide() will be called for all images for all providers and log v(3) output about the image not being a valid ECR repository URL for non-ECR images

Would it make sense to also:

1. drop the `klog.V(3).Info(err)` output if a non-ECR image is passed to provide... this isn't something we need to log

2. move the current Enabled() check to a once.Do check we call before calling `getFromECR`. If that fails, make future calls to Provider() no-op

pkg/credentialprovider/aws/aws_credentials.go
```diff
@@ -66,17 +66,6 @@ func newECRProvider(getterFactory tokenGetterFactory) *ecrProvider {
 // Enabled implements DockerConfigProvider.Enabled. Enabled is true if AWS
 // credentials are found.
 func (p *ecrProvider) Enabled() bool {
-	sess, err := session.NewSessionWithOptions(session.Options{
```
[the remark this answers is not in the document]
liggitt: agree I'd like to see a cheap zero-network-call check instead of an expensive one

liggitt:
sorry this escaped attention for so long. agree this should be fixed asap.

pkg/credentialprovider/aws/aws_credentials.go
nckturner: I think both of those sound fine to me.

nckturner:
Will address the comments

pkg/credentialprovider/aws/aws_credentials.go
nckturner: The only issue with this is we have to be careful with failures, i.e. we don't want an outage to cause permanent unhealthiness in kubelet.

nckturner:
@andrewsykim 
> On second thought, this could be misleading because kubelet will always attempt to authenticate to ECR repos but `Provide` will fail if there's no session established. Could `Enabled` instead check the availablility of the metadata server and/or the aws credentials on disk `~/.aws/credentials`? Basically a similar check to what already exists but skips the API calls.

Hmm, I'm not sure what assumptions we want to make here.  Metadata isn't guaranteed to be available, and in the cases where its not, we'd need to check everything in the sdk's credential chain (i.e., one possibility is using the 2 env vars: `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN`).  Probably possible but also sounds brittle.

pkg/credentialprovider/aws/aws_credentials.go
liggitt: sure... by moving the existing check to a single run-once that happens only after we are asked to provide credentials for an ECR image, I *think* we do the following:
* maintain current behavior on AWS
* remove lag on non-AWS on a cluster that never tries to pull an ECR image
* limits lag on a non-AWS cluster that tries to pull an ECR image to a single attempt

andrewsykim:
> we'd need to check everything in the sdk's credential chain

Is there a way to configure the client to check everything in the SDK credential chain except for those that require API calls?

nckturner:
@andrewsykim I actually think that its retries to IMDS during the normal that is causing the latency when we try to set up the session. There is logic in the SDK which has to decide between IMDSv1 and IMDSv2, with a fallback mechanism.  There have been issues in the past with the SDK which caused latency when falling back to IMDSv1.  If this is the case, then we might be able to get rid of the latency just adjusting the timeout.  The other possibility is, the SDK sees the service account token and then tries to do an assume-role-with-web-identity call, which it can use to exchange for aws creds.  I'll see if I can pinpoint the latency.

nckturner:
@andrewsykim I repro'd this on a gce instance (although that might have not been necessary, given the repro).  If you drop packets to the metadata endpoint via an iptables rule, the aws sdk gets really unhappy:  

```
root@kubernetes-build:/home/nic/kubernetes/pkg/credentialprovider/aws# sudo iptables -A OUTPUT -p tcp -o ens4 -d 169.254.169.254/32 -j DROP
root@kubernetes-build:/home/nic/kubernetes/pkg/credentialprovider/aws# telnet 169.254.169.254 80
Trying 169.254.169.254...
^C
root@kubernetes-build:/home/nic/kubernetes/pkg/credentialprovider/aws# go test -bench .
E1120 07:51:03.065014    9083 aws_credentials.go:69] aws_credential:1
E1120 07:51:03.065656    9083 aws_credentials.go:73] aws_credential:2
E1120 07:51:03.065686    9083 aws_credentials.go:78] aws_credential:3
E1120 07:51:23.405789    9083 aws_credentials.go:80] aws_credential:4
E1120 07:51:23.405826    9083 aws_credentials.go:81] while getting AWS credentials NoCredentialProviders: no valid providers in chain. Deprecated.
        For verbose messaging see aws.Config.CredentialsChainVerboseErrors
ECR is enabled: false
goos: linux
goarch: amd64
pkg: k8s.io/kubernetes/pkg/credentialprovider/aws
BenchmarkSetupLatency-8                1        20340895511 ns/op
PASS
ok      k8s.io/kubernetes/pkg/credentialprovider/aws    20.358s
root@kubernetes-build:/home/nic/kubernetes/pkg/credentialprovider/aws# 

```

This is the patch I used to test: 
```
nic@kubernetes-build:~/kubernetes/pkg/credentialprovider/aws$ git diff
diff --git a/pkg/credentialprovider/aws/aws_credentials.go b/pkg/credentialprovider/aws/aws_credentials.go
index a650a7da50b..3a06d9ea7b3 100644
--- a/pkg/credentialprovider/aws/aws_credentials.go
+++ b/pkg/credentialprovider/aws/aws_credentials.go
@@ -66,17 +66,22 @@ func newECRProvider(getterFactory tokenGetterFactory) *ecrProvider {
 // Enabled implements DockerConfigProvider.Enabled. Enabled is true if AWS
 // credentials are found.
 func (p *ecrProvider) Enabled() bool {
+       klog.Error("aws_credential:1")
        sess, err := session.NewSessionWithOptions(session.Options{
                SharedConfigState: session.SharedConfigEnable,
        })
+       klog.Error("aws_credential:2")
        if err != nil {
                klog.Errorf("while validating AWS credentials %v", err)
                return false
        }
+       klog.Error("aws_credential:3")
        if _, err := sess.Config.Credentials.Get(); err != nil {
+               klog.Error("aws_credential:4")
                klog.Errorf("while getting AWS credentials %v", err)
                return false
        }
+       klog.Error("aws_credential:5")
        return true
 }
 
diff --git a/pkg/credentialprovider/aws/aws_credentials_test.go b/pkg/credentialprovider/aws/aws_credentials_test.go
index 7161fc136d1..27536acce20 100644
--- a/pkg/credentialprovider/aws/aws_credentials_test.go
+++ b/pkg/credentialprovider/aws/aws_credentials_test.go
@@ -331,3 +331,9 @@ func TestChinaECRProvideCached(t *testing.T) {
                t.Errorf("cached credentials do not match")
        }
 }
+
+func BenchmarkSetupLatency(b *testing.B) {
+       p := newECRProvider(&ecrTokenGetterFactory{cache: make(map[string]tokenGetter)})
+       enabled := p.Enabled()
+       fmt.Printf("ECR is enabled: %v\n", enabled)
+}

```

Without the iptables rule, the benchmark takes about 30 milliseconds:
```
ok      k8s.io/kubernetes/pkg/credentialprovider/aws    0.031s
```
and from a network namespace without any connectivity: 
```
root@kubernetes-build:/home/nic/kubernetes/pkg/credentialprovider/aws# ip netns exec test0 go test -bench .
...
ok      k8s.io/kubernetes/pkg/credentialprovider/aws    0.021s

```
Its about 10 ms faster then the default case, probably because it immediately fails because there is no interface. (20 ms is still a long time though...)

Anyway, this is all to say, I don't think we can rely on the SDK or configure it to only check "the fast things" since instance metadata is pretty central to what it tries to do.  Using `-j REJECT` instead of `-j DROP brings the latency down to 4-6 seconds (depending on version of sdk), but that's still unacceptable. I think I'm leaning toward's @liggitt's solution of a run once check, the only thing I don't like about it is that if the metadata check randomly fails for an AWS kubelet, then that kubelet would have to be restarted (if I'm understanding correctly).

andrewsykim:
> the only thing I don't like about it is that if the metadata check randomly fails for an AWS kubelet, then that kubelet would have to be restarted (if I'm understanding correctly).

Yeah this would be problematic -- a failed connection to metadata server would disable the provider, but we can't necessary panic either,  otherwise kubelets not on EC2 instances would panic too.

liggitt:
> the only thing I don't like about it is that if the metadata check randomly fails for an AWS kubelet, then that kubelet would have to be restarted (if I'm understanding correctly).

it would be nice if that could be improved, but I see that as something for the AWS provider to work on as a follow-up after getting this merged to avoid affecting all non-AWS deployments

andrewsykim:
Having a hard time understanding how the run once check helps here. `Enabled()` is a blocking call and even if we ran it once it would block kubelet startup anyways?

andrewsykim:
> move the current Enabled() check to a once.Do check we call before calling getFromECR. If that fails, make future calls to Provide() no-op

Ah, because we would move the check to `Provide()` and we have `Enabled` always return true.

nckturner:
Well this makes me think we should just have Enabled() return true and rely on the regex to short circuit any non-ecr images, like it was before.  Then, only if the user is actually trying to pull an ECR image, do we actually attempt to set up aws credentials.  This means that the only users who would experience the latency are those who: 1. are trying to pull an ECR image and 2. don't have aws credentials set up and 3. have blocked access to metadata.

liggitt:
Enabled() is called from NewDockerKeyring()

When used with CRI, this is only called once at startup time from NewKubeGenericRuntimeManager(), so AWS nodes with metadata unavailable at kubelet start already had their AWS credential provider disabled.

When used with dockershim, this is called on every image pull from ensureSandboxImageExists()

While limiting the impact to Provide() and scoping to ECR images is clearly better than the current code for dockershim, I don't think we should re-add a check on every pull to kubelets using CRI on non-AWS nodes.

nckturner:
I guess what you're saying is its gross to make "Enabled" meaningless in the CRI code path, and cause every image pull to check with the ecr credential provider, even if its protected by the image uri regex, because it exposes clusters running on other providers to bugs in the ecr credential provider, and could be considered an anti-pattern to execute this provider specific code path when its unnecessary.  

But I would argue if you look at the negatives for each approach, comparing how it affects aws users and non-aws users, the downside to all aws users in the event of a metadata outage at the wrong time is much worse (requires intervention by cluster operators) than the downside of non-aws users being exposed to future code bugs and having code path execute that is unnecessary (unless I'm missing other downsides).  Additionally, this is remediated by the new credential provider framework, so this isn't something permanent, either.  Given that, I'd ask if we can go back to using the regex short-circuit on every CRI image pull.

nckturner:
If we still want to go with a run-once check, maybe we can first check for the "EC2" prefix in each of:
```
sh-4.2$ cat /sys/hypervisor/uuid 
ec2106d7-184d-6e02-7a1b-00fcc21b0741

sh-4.2$ sudo cat /sys/devices/virtual/dmi/id/product_uuid 
EC2106D7-184D-6E02-7A1B-00FCC21B0741
```
and only attempt metadata/credential chain if those fail or or don't exist, which would make most ec2 checks work even during a metadata outage or flake.

liggitt:
Those files look promising. Avoiding perma-failure due to unavailable metadata where one of those files positively indicates we are in ec2 seems reasonable, as long as we get the following characteristics on non-EC2 nodes:

* Enabled() is fast and doesn't make API calls
* Provide(non-ECR image) is fast and doesn't make API calls
* Provide(ECR image) makes a slow failing invocation at most once, then is fast on future invocations and doesn't make API calls

nckturner:
@liggitt I took an initial pass at your comments.

> sure... by moving the existing check to a single run-once that happens only after we are asked to provide credentials for an ECR image, I think we do the following:
> * maintain current behavior on AWS
> * remove lag on non-AWS on a cluster that never tries to pull an ECR image
> * limits lag on a non-AWS cluster that tries to pull an ECR image to a single attempt

As an initial version I did the once.Do in the enabled(), because I think its easier to understand, and the slow call will only happen once on non-AWS kubelets. 

> When used with CRI, this is only called once at startup time

Since it will only be called once at startup time with CRI, I think it would be reasonable, but if you'd prefer I can move it to `Provide()` to account for dockershim.

pkg/credentialprovider/aws/aws_credentials.go
```diff
@@ -63,9 +68,57 @@ func newECRProvider(getterFactory tokenGetterFactory) *ecrProvider {
...
+func tryValidateEC2UUID() bool {
+	hypervisor_uuid := "/sys/hypervisor/uuid"
+	product_uuid := "/sys/devices/virtual/dmi/id/product_uuid"
+	asset_tag := "/sys/devices/virtual/dmi/id/board_asset_tag"
+
+	if _, err := os.Stat(hypervisor_uuid); err == nil {
+		b, err := ioutil.ReadFile(hypervisor_uuid)
+		if err != nil {
+			klog.Errorf("error checking if this is an EC2 instance: %v", err)
+		} else if string(b[:3]) == "EC2" || string(b[:3]) == "ec2" {
```
liggitt, line 93: this will panic if fewer than three bytes are read... prefer `strings.HasPrefix()` (applies everywhere)?

pkg/credentialprovider/aws/aws_credentials.go
```diff
@@ -63,9 +68,57 @@ func newECRProvider(getterFactory tokenGetterFactory) *ecrProvider {
...
+			klog.V(5).Infof("found 'ec2' in uuid %v from %v, enabling legacy AWS credential provider", string(b), product_uuid)
+			return true
+		}
+	}
+
+	if _, err := os.Stat(asset_tag); err == nil {
+		b, err := ioutil.ReadFile(asset_tag)
+		if err != nil {
+			klog.Errorf("error checking if this is an EC2 instance: %v", err)
+		} else if string(b[:2]) == "i-" {
```
liggitt, line 113: I'm not very familiar with this file... does this exist in non-aws environments? is an `i-` prefix really strongly indicative we're in aws?

liggitt:
> the slow call will only happen once on non-AWS kubelets.

there needs to be zero network calls made on a non-AWS kubelet that is not asked to pull an ECR image.

pkg/credentialprovider/aws/aws_credentials.go
```diff
@@ -38,34 +40,93 @@ import (
...
+			klog.V(5).Infof("found 'ec2' in uuid %v from %v, enabling legacy AWS credential provider", string(b), product_uuid)
+			return true
+		}
+	}
+
+	if _, err := os.Stat(asset_tag); err == nil {
+		b, err := ioutil.ReadFile(asset_tag)
+		if err != nil {
+			klog.Errorf("error checking if this is an EC2 instance: %v", err)
+		} else if strings.HasPrefix(string(b), "i-") && len(string(b)) == 20 {
```
ayberk, line 120: Can we check if the trimmed string is 19 characters instead? I am thinking of possibility of having something like `\r\n` as new line for example.

pkg/credentialprovider/aws/aws_credentials.go
```diff
@@ -38,34 +40,93 @@ import (
...
+// an EC2 vm based on checking for the EC2 system UUID, the asset tag (for nitro
+// instances), or instance credentials if the UUID is not present.
+// Ref: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html
+func ec2ValidationImpl() bool {
+	return tryValidateEC2UUID() || tryValidateEC2Creds()
+}
+
+func tryValidateEC2UUID() bool {
+	hypervisor_uuid := "/sys/hypervisor/uuid"
+	product_uuid := "/sys/devices/virtual/dmi/id/product_uuid"
```
ayberk, line 93: Is this the location for `Get the UUID from DMI`?

pkg/credentialprovider/aws/aws_credentials.go
```diff
@@ -38,34 +40,93 @@ import (
...
+			klog.V(5).Infof("found 'ec2' in uuid %v from %v, enabling legacy AWS credential provider", string(b), hypervisor_uuid)
+			return true
+		}
+	}
+
+	if _, err := os.Stat(product_uuid); err == nil {
+		b, err := ioutil.ReadFile(product_uuid)
+		if err != nil {
+			klog.Errorf("error checking if this is an EC2 instance: %v", err)
+		} else if strings.HasPrefix(string(b), "EC2") || strings.HasPrefix(string(b), "ec2") {
```
ayberk, line 110: Assuming this is the UUID from DMI, this won't work for little endian machines (if we care about that):

> In the following example output, the UUID is represented in little-endian format.
> 45E12AEC-DCD1-B213-94ED-01234ABCDEF

pkg/credentialprovider/aws/aws_credentials.go
nckturner: Good call

pkg/credentialprovider/aws/aws_credentials.go
nckturner: Interesting.  It doesn't have to be perfect, we still try to initialize credentials after doing these checks. The advantage to being able to shortcut the credential checks is that in the case the metadata endpoint is down temporarily, we can still verify the machine is on EC2.  Since we're now using a `once.Do()` check, if the endpoint was down we would disable the provider and the user would have to restart kubelet.  With these checks, the operator would only have to restart kubelet if these file checks don't work (e.g. little endian), and the credentials fail to initialize because of an outage.

pkg/credentialprovider/aws/aws_credentials.go
nckturner: Added a length check as well, so if the file contains only "i-" and is the length of an AWS instance ID, which I think is a pretty good indicator.  Since this only enables the ECR credential provider, I don't think a false positive in an obscure case is that bad--the user would still need to have ECR images in podspecs in order for the `Provide` method to actually be called.

pkg/credentialprovider/aws/aws_credentials.go
nckturner: Correct

nckturner:
Ready for review @liggitt

pkg/credentialprovider/aws/aws_credentials.go
```diff
@@ -85,12 +148,26 @@ func (p *ecrProvider) Enabled() bool {
 func (p *ecrProvider) Provide(image string) credentialprovider.DockerConfig {
 	parsed, err := parseRepoURL(image)
 	if err != nil {
-		klog.V(3).Info(err)
+		klog.V(5).Info(err)
```
liggitt, line 151: this should not log anything if the regex simply didn't match

liggitt: mentioned above as:

> drop the klog.V(3).Info(err) output if a non-ECR image is passed to provide... this isn't something we need to log

pkg/credentialprovider/aws/aws_credentials.go
```diff
@@ -38,34 +40,94 @@ import (
...
+		}
+	}
+
+	if _, err := os.Stat(asset_tag); err == nil {
+		b, err := ioutil.ReadFile(asset_tag)
+		s := strings.TrimSpace(string(b))
+		if err != nil {
+			klog.Errorf("error checking if this is an EC2 instance: %v", err)
+		} else if strings.HasPrefix(s, "i-") && len(s) == 19 {
+			// Instance ID's are 19 characters plus newline
```
liggitt, line 122: is this guaranteed or will this expand/change in the future?

pkg/credentialprovider/aws/aws_credentials.go
```diff
@@ -38,34 +40,94 @@ import (
 	"k8s.io/kubernetes/pkg/credentialprovider"
 )
 
-var ecrPattern = regexp.MustCompile(`^(\d{12})\.dkr\.ecr(\-fips)?\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.amazonaws\.com(\.cn)?$`)
+var (
+	ecrPattern = regexp.MustCompile(`^(\d{12})\.dkr\.ecr(\-fips)?\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.amazonaws\.com(\.cn)?$`)
```
liggitt, line 44: verified the regex didn't change, this is just a formatting change to be in a var block

pkg/credentialprovider/aws/aws_credentials.go
nckturner: Very unlikely to change anytime soon because it was increased a few years ago in order to become large enough to account for forseeable future growth.

pkg/credentialprovider/aws/aws_credentials.go
nckturner: Yeah I reduced to V5 because I found it useful for testing but happy to remove it.

nckturner: Removed in latest push.

wongma7:
/lgtm

tx

nckturner:
@liggitt I believe I've addressed your comments, lmk if there's anything else.

PR #93261, opened 2020-07-20
WIP: kubelet: invalidate runtime cache on container start

### LINKED ISSUE #93210: MirrorPodWithGracePeriod test flaking in 5% of pull-kubernetes-node-e2e runs

sjenning:
**What type of PR is this?**
/kind bug

If a pod's container starts and the pod is deleted all within the 2s defaultCachePeriod of the runtime cache, then the grace periods of the containers may not be honored.  This is because `GetPods()` can return a stale `runningPod` with no containers and proceed directly to `StopPodSandbox` without properly terminating the container.

Also see https://github.com/kubernetes/kubernetes/issues/93210#issuecomment-661244774

~~This PR allows a caller to force and update to the cache on paths that require the latest state from the runtime.~~
This PR invalidates the kubelet `runtimeCache` whenever a container start is detect so a subsequent call to `GetPods()` will get up-to-date information.

xref #93210

```release-note
NONE
```
/sig node
/priority important-soon 

I currently haven't proved that this solves the issue, hence WIP.  I'm opening this PR to do some test runs to see if the issue occurs.  In the meantime, I am trying to develop a more concise reproducer to confirm the root-cause/fix.

/cc @liggitt @derekwaynecarr


pkg/kubelet/kubelet.go
```diff
@@ -1657,7 +1657,7 @@ func (kl *Kubelet) deletePod(pod *v1.Pod) error {
 
 	// Runtime cache may not have been updated to with the pod, but it's okay
 	// because the periodic cleanup routine will attempt to delete again later.
-	runningPods, err := kl.runtimeCache.GetPods()
+	runningPods, err := kl.runtimeCache.GetPods(true)
```
dims, line 1660: If this is the only place we are calling `GetPods(true)`, may be just call `getPodsWithTimestamp()` directly? is there a trade off?

sjenning: For now, yes.

`runtimeCache` is a `RuntimeCache` interface and `getPodsWithTimestamp()` isn't accessible and I didn't want to add another method to the interface that did mostly the same thing as `GetPods()` when I could just toggle the behavior with a boolean.

liggitt: how expensive is getting pods and missing the cache hit?

sjenning: The default housekeeping routine runs every 2s.  The cache expiry is also 2s.  So there is a call through the cache every 2s.  This only adds an additional call through the cache into the runtime on each pod delete.

It is expensive for the runtime, but we aren't doing it considerably more after this change.

In fact... know that I think about it... we probably aren't doing any more at all because forcing the update means that we the next call the housekeeping routine would do is cached.  We are just advancing a call that we would have done anyway.  Unless we are doing more that one pod delete every 2s.

liggitt: > Unless we are doing more than one pod delete every 2s.

"rapidly delete lots of pods" is the scenario I'm wondering about. 

> If a pod's container starts and the pod is deleted all within the 2s defaultCachePeriod of the runtime cache, then the grace periods of the containers may not be honored. This is because GetPods() can return a stale runningPod with no containers

is it possible to determine the last time we attempted a start of a container in the pod and pass that to GetPods as a sort of notOlderThan? Would that let us continue using the cached result in most cases, and only skip the cache when needed?

sjenning: maybe there is some way to link PLEG to the runtime cache... i'll look into it

sjenning:
@liggitt ok, complete reworked this.

Now, the PR invalidates the kubelet `runtimeCache` whenever a container start is detect so a subsequent call to `GetPods()` will get up-to-date information.

pkg/kubelet/container/runtime_cache.go
```diff
@@ -30,6 +30,7 @@ var (
 type RuntimeCache interface {
 	GetPods() ([]*Pod, error)
 	ForceUpdateIfOlder(time.Time) error
+	Invalidate()
```
liggitt, line 33: interesting, this is similar in nature to ForceUpdateIfOlder, except:
* I would have expected ForceUpdateIfOlder to be lazy, and it is not
* ForceUpdateIfOlder is also never called :)

sjenning: ha! I noticed that as well.  We could remove it and just use `Invalidate` since the whole mechanism of `ForceUpdateIfOlder` doesn't really make sense.

Caller: Update the cache is it is older that this.
Callee: I would return you the updated data but I only return and error.
Caller: np, I just wanted to create runtime load anyway.  I might call `GetPods()` later.
Callee: ooook....

pkg/kubelet/kubelet.go
```diff
@@ -1850,6 +1850,12 @@ func (kl *Kubelet) syncLoopIteration(configCh <-chan kubetypes.PodUpdate, handle
 			}
 		}
 
+		if e.Type == pleg.ContainerStarted {
```
liggitt, line 1853: remind me where `<-plegCh` events come from? can this still race with a DELETE event from `<-configCh`?

sjenning: They come from here
https://github.com/sjenning/kubernetes/blob/a104fa36d24d95a6cc87889b16e49e4f76abb59c/pkg/kubelet/pleg/generic.go#L188-L302

At first I thought I had just created a loop and that the runtimeCache fed PLEG, but that isn't true.

PLEG calls kuberuntime's `GetPods()` (different from the `runtimeCache`), which calls directly into the actual runtime service via `getKubeletSandboxes`/`getKubeletContainers`.

sjenning:
/assign @dashpole 
This is kinda tricky.  Would appreciate your review, David!

pkg/kubelet/kubelet.go
```diff
@@ -1839,6 +1839,17 @@ func (kl *Kubelet) syncLoopIteration(configCh <-chan kubetypes.PodUpdate, handle
 		kl.sourcesReady.AddSource(u.Source)
 
 	case e := <-plegCh:
+		if e.Type == pleg.ContainerStarted {
```
liggitt, line 1842: talked offline... this doesn't solve the race observing a delete from configCh and oberving a ContainerStarted event from plegCh, but it does solve the issue where a container status reflected in the API (so the kubelet clearly knew the container was running) did not inform graceful deletion

My main question now is what the load impact is for a kubelet that is rapidly creating containers:

* does invalidating here cause any pathological issues for a pod with N containers (where each individual ContainerStarted event invalidates the cache and some other path calls GetPods() per-container, resulting in N cache misses.
* do we only care that GetPods() is super up-to-date in the delete case? would it make sense to use the ForceUpdateIfOlder() method in deletePod() passing in the max status.{init}containers[*].state.running.startedAt or something?

Suggestions if we keep the current approach:

1. a test exercising this would be great if it's possible
2. I'm not sure if it would be possible to move the invalidate to the point where we make the call to create the container or not

/cc @dashpole for kubelet sync loop review expertise as well

liggitt: > * do we only care that GetPods() is super up-to-date in the delete case? would it make sense to use the ForceUpdateIfOlder() method in deletePod()?

a sketch of that approach is at https://github.com/liggitt/kubernetes/commits/force-delete ... that would let us keep the benefits of the runtimeCache in most cases (only rapid start/delete cases would result in cache misses)

dashpole: I like the deletion-only approach.  The only thing I need to check is IIRC there are a few different deletion paths.  We should make sure we actually cover all deletion cases...

dashpole: Looks like we just need to add the update logic in one other location

dashpole:
I prefer @liggitt's approach above.  I think extra container list calls are a concern if there is significant churn

sjenning:
@dashpole agreed.  I've asked @liggitt to open a PR based on https://github.com/liggitt/kubernetes/commit/da1f636514251fa84d89471d1f66c9fb20909d11

sjenning:
closing in favor of https://github.com/kubernetes/kubernetes/pull/93307

PR #91113, opened 2020-05-14
Refactor create deployment and add --port flag 

soltysh:
**What type of PR is this?**
/kind cleanup
/kind feature

**What this PR does / why we need it**:
This is coming from discussions from https://github.com/kubernetes/kubernetes/pull/87077 to add `--port` flag support for `kubectl create deployment` command.

**Special notes for your reviewer**:
/assign @sallyom @zhouya0 

**Does this PR introduce a user-facing change?**:
```release-note
Add --port flag to kubectl create deployment
```

PR #87077, opened 2020-01-10
Remove kubectl run generators

soltysh:
**What type of PR is this?**
/kind api-change
/kind cleanup
/kind deprecation
/priority backlog
/sig cli

**What this PR does / why we need it**:
Over a year ago we've deprecated all the generators from under `kubectl run` command except for the one responsible for generating a pod (see https://github.com/kubernetes/kubernetes/pull/68132). 
This PR removes all the generators and deprecates all the flags towards the goal to make `kubectl run` simpler.

**Special notes for your reviewer**:
/assign @liggitt @seans3 @pwittrock 

**Does this PR introduce a user-facing change?**:
```release-note
Remove all the generators from kubectl run. It will now only create pods. Additionally, deprecates all the flags that are not relevant anymore.
```

/cc @kubernetes/sig-cli-api-reviews @kubernetes/sig-cli-pr-reviews

PR #68132, opened 2018-08-31
Start deprecating all generators in run except for run-pod/v1

soltysh:
**What this PR does / why we need it**:
This was discussed during SIG-CLI meetings over several past months. The direction is that we want to move away from `kubectl run` because it's over bloated and complicated for both users and developers. We want to mimic `docker run` with `kubectl run` so that it *only* creates a pod, and if you're interested in other resources `kubectl create` is the intended replacement. 

This PR starts with deprecating all of the generator except for the pod one.

/assign @juanvallejo 
/sig cli
/milestone v1.12

**Release note**:
```release-note
Deprecate kubectl run generators, except for run-pod/v1
```

PR #93269, opened 2020-07-20
kubectl: add env option to deployment creation

Xercoy:
**What type of PR is this?**
/kind feature

Creating a deployment using kubectl with environment variables was done with `kubectl run`. Since most of the generators have been deprecated, including the one for deployments (#87077), this PR adds functionality to the preferred method of deployment creation, `kubectl create deployment`.

I think this feature would be useful in providing more flexibility for deployments with `kubectl create`. Additionally, resources related to deployment creation with environment variables via `kubectl run` can be updated and remain useful rather than eventually being deprecated.

I did not create an issue/enhancement beforehand but I would be more than happy to do so. Apologies especially in case I had missed any content which deemed that this feature shouldn't be implemented. 

Supporting documentation/PRs:
- [#91113](https://github.com/kubernetes/kubernetes/pull/91113)
- [#87077](https://github.com/kubernetes/kubernetes/pull/87077)
- [#68132](https://github.com/kubernetes/kubernetes/pull/68132)

and also attached in the docs section.

In addition to running and providing tests, I built the kubectl binary and tested that the new flag/feature works as expected; the intent is for its behavior to be similar to how the flag was in `kubectl run`.

Please let me know if any changes or amendments should be made, or if I'm not adhering to any guidelines or processes. Would also be more than happy to refactor any recommendations for changes.

Thank you very much to all reviewers, maintainers, and contributors for the time and effort!

Yes, this will add an optional flag `--env` which expects a key value pair of an environment name and value. This flag can be repeated to provide multiple environment variables. 

```release-note
Added the --env flag to allow environment variables to be provided to deployments made with kubectl create. 
```

```docs
- [#91113](https://github.com/kubernetes/kubernetes/pull/91113)
- [#87077](https://github.com/kubernetes/kubernetes/pull/87077)
- [#68132](https://github.com/kubernetes/kubernetes/pull/68132)
```

### DIFF
--- a/staging/src/k8s.io/kubectl/pkg/cmd/create/create_deployment_test.go
+++ b/staging/src/k8s.io/kubectl/pkg/cmd/create/create_deployment_test.go
@@ -18,6 +18,7 @@ package create
 
 import (
 	"bytes"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"strings"
@@ -31,6 +32,7 @@ import (
 	cmdtesting "k8s.io/kubectl/pkg/cmd/testing"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/scheme"
+	"k8s.io/kubectl/pkg/util"
 )
 
 func TestCreateDeployment(t *testing.T) {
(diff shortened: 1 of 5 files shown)


zhouya0 (commented):
Code lgtm but:
The main goal is to make create command simple as possible, other need to create deployment should use `kubectl apply` instead. Before we add `--port`, `--replicas` flag since they are commonly used in deployment . Not sure we really want to continue add more.

Xercoy:
@zhouya0 , thank you for the quick review! I definitely agree with you in that  `kubectl create deployment` should be as simple as possible. 

For whatever purposes that are required when using `kubectl` to create a deployment, whether it's debugging, testing, learning, or teaching, I agree and say that you're right again in that the previous options that were added affect configurations which are commonly used in deployments:

- `--port`: the ability to specify a port for a deployment

- `--replicas`: the ability to control the number of replicas

-  the ability to provide a command to a deployment

I also think that the ability to provide a deployment with environment variables can also be considered a core configuration of a deployment. I believe that it is commonly used and that the benefits from this feature would outweigh the simplicity that the changes in this PR removes.

The flags listed above enable users to create a deployment which dictates the command, number of replicas, and port. One of the most important pieces of an application's operations are configuration values; many applications need some kind of configuration values to run smoothly. Adding `--env` provides the create command with the ability to create deployments with the information they need to run, if such configuration is necessary.

Manipulating deployments with `kubectl apply` or through Helm charts are definitely the best way to have the most control and customization, but I still think that use cases for `kubectl create` would largely benefit from having `--env`. The create command is much more than capable of creating deployments for images other than `busybox` or `nginx`. This flag would improve on that capability by providing deployments with data that it may need to run.

From my perspective as a user and someone who shares knowledge about Kubernetes, this PR improves how deployments can be created with the create command. 

As a contributor with this PR, the changes make use of utility/helper functions that already exist, includes tests, and adds code changes that matches how other flags are handled, such as `--replicas` and `--port`. In my opinion, this last detail keeps things simple and ensures that refactoring is just as easy. 

While I do agree with you on both points in that the create command should be as simple as possible, and again when you mentioned that the flags mentioned above are commonly used in deployments, adding `--env` and regarding it as commonly a used configuration from the points I mentioned should be considered.

Of course, I defer the decision to validate the need for this PR to you and the consensus of the community. Overall, it's been an awesome learning experience and a great chance to collaborate and make contributions to a project that has given so much to me. Thanks again.

Xercoy:
@zhouya0 while Tide is mentioning that this PR isn't mergeable yet, all tests are passing. Thanks again for the review and enabling tests! For the time being, I'll try to promote visibility but please let me know if there's anything else I can do to help.

staging/src/k8s.io/kubectl/pkg/cmd/create/create_deployment_test.go
```diff
@@ -125,6 +127,85 @@ func TestCreateDeploymentWithReplicas(t *testing.T) {
...
+
+	testCases := []struct {
+		Name string
+		Envs []string
+	}{
+		{
+			"SingleEnvVar",
+			[]string{
+				"FOO=bar",
+				"FIZZ=buzz",
```
Xercoy, line 141: This test case should have only one env var set, will fix.

eddiezane:
Holding this while we explore other options.

/hold

PR #93284, opened 2020-07-21
pass instanceMetadata to updateNodeAddress to reduce api calls

### LINKED ISSUE #90652: Reduce the number of calls to GCE API (for clusters on GCE) coming from cloud node-controller

gongguan:
**What type of PR is this?**
/kind feature

Pass instanceMetadata to updateNodeAddress to reduce api calls.

Ref #90652

```release-note
None
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
+++ b/staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
@@ -155,7 +155,12 @@ func (cnc *CloudNodeController) UpdateNodeStatus(ctx context.Context) {
 	}
 
 	for i := range nodes.Items {
-		cnc.updateNodeAddress(ctx, &nodes.Items[i])
+		instanceMetadata, err := cnc.getInstanceNodeAddresses(ctx, &nodes.Items[i])
+		if err != nil {
+			klog.Errorf("Error getting instance metadata for node addresses: %v", err)
+			continue
+		}
+		cnc.updateNodeAddress(ctx, &nodes.Items[i], instanceMetadata)
 	}
 
 	for _, node := range nodes.Items {
@@ -217,7 +222,7 @@ func (cnc *CloudNodeController) reconcileNodeLabels(nodeName string) error {
 }
 
 // UpdateNodeAddress updates the nodeAddress of a single node
-func (cnc *CloudNodeController) updateNodeAddress(ctx context.Context, node *v1.Node) {
+func (cnc *CloudNodeController) updateNodeAddress(ctx context.Context, node *v1.Node, instanceMetadata *cloudprovider.InstanceMetadata) {
 	// Do not process nodes that are still tainted
 	cloudTaint := getCloudTaint(node.Spec.Taints)
 	if cloudTaint != nil {
@@ -225,34 +230,7 @@ func (cnc *CloudNodeController) updateNodeAddress(ctx context.Context, node *v1.
 		return
 	}
 
-	instanceMetadataGetter := func(providerID string, node *v1.Node) (*cloudprovider.InstanceMetadata, error) {
-		if instancesV2, ok := cnc.cloud.InstancesV2(); instancesV2 != nil && ok {
-			return instancesV2.InstanceMetadata(ctx, node)
-		}
-
-		// If InstancesV2 not implement, use Instances.
-		instances, ok := cnc.cloud.Instances()
-		if !ok {
-			return nil, fmt.Errorf("failed to get instances from cloud provider")
-		}
-
-		nodeAddresses, err := getNodeAddressesByProviderIDOrName(ctx, instances, node.Spec.ProviderID, node.Name)
-		if err != nil {
-			klog.Errorf("Error getting node addresses for node %q: %v", node.Name, err)
-			return nil, err
-		}
-		return &cloudprovider.InstanceMetadata{
-			NodeAddresses: nodeAddresses,
-		}, nil
-	}
-
-	instanceMeta, err := instanceMetadataGetter(node.Spec.ProviderID, node)
-	if err != nil {
-		utilruntime.HandleError(err)
-		return
-	}
-
-	nodeAddresses := instanceMeta.NodeAddresses
+	nodeAddresses := instanceMetadata.NodeAddresses
 	if len(nodeAddresses) == 0 {
 		klog.V(5).Infof("Skipping node address update for node %q since cloud provider did not return any", node.Name)
 		return
@@ -363,9 +341,19 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
 		return
 	}
 
-	// TODO: getNodeModifiersFromCloudProvider and updateNodeAddress both call cloud api to get instanceMetadata,
-	// get instanceMetadata and pass it to getNodeModifiersFromCloudProvider and updateNodeAddress which reduces api calls.
-	nodeModifiers, err := cnc.getNodeModifiersFromCloudProvider(ctx, curNode)
+	providerID, err := cnc.getProviderID(ctx, curNode)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("failed to get provider ID for node %s at cloudprovider: %v", node.Name, err))
+		return
+	}
+
+	instanceMetadata, err := cnc.getInstanceMetadata(ctx, providerID, curNode)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("failed to get instance metadata for node %s: %v", node.Name, err))
+		return
+	}
+
+	nodeModifiers, err := cnc.getNodeModifiersFromCloudProvider(ctx, providerID, curNode, instanceMetadata)
 	if err != nil {
 		utilruntime.HandleError(fmt.Errorf("failed to initialize node %s at cloudprovider: %v", node.Name, err))
 		return
@@ -392,7 +380,7 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
 
 		// After adding, call UpdateNodeAddress to set the CloudProvider provided IPAddresses
 		// So that users do not see any significant delay in IP addresses being filled into the node
-		cnc.updateNodeAddress(ctx, curNode)
+		cnc.updateNodeAddress(ctx, curNode, instanceMetadata)
 
 		klog.Infof("Successfully initialized node %s with cloud provider", node.Name)
 		return nil
@@ -407,86 +395,16 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
 // a node object with provider-specific information.
 // All of the returned functions are idempotent, because they are used in a retry-if-conflict
 // loop, meaning they could get called multiple times.
-func (cnc *CloudNodeController) getNodeModifiersFromCloudProvider(ctx context.Context, node *v1.Node) ([]nodeModifier, error) {
-	var (
-		nodeModifiers []nodeModifier
-		providerID    string
-		err           error
-	)
+func (cnc *CloudNodeController) getNodeModifiersFromCloudProvider(
+	ctx context.Context,
+	providerID string,
+	node *v1.Node,
+	instanceMeta *cloudprovider.InstanceMetadata,
+) ([]nodeModifier, error) {
 
-	if node.Spec.ProviderID == "" {
-		providerID, err = cloudprovider.GetInstanceProviderID(ctx, cnc.cloud, types.NodeName(node.Name))
-		if err == nil {
-			nodeModifiers = append(nodeModifiers, func(n *v1.Node) {
-				if n.Spec.ProviderID == "" {
-					n.Spec.ProviderID = providerID
-				}
-			})
-		} else if err == cloudprovider.NotImplemented {
-			// if the cloud provider being used does not support provider IDs,
-			// we can safely continue since we will attempt to set node
-			// addresses given the node name in getNodeAddressesByProviderIDOrName
-			klog.Warningf("cloud provider does not set node provider ID, using node name to discover node %s", node.Name)
-		} else {
-			// if the cloud provider being used supports provider IDs, we want
-			// to propagate the error so that we re-try in the future; if we
-			// do not, the taint will be removed, and this will not be retried
-			return nil, err
-		}
-	} else {
-		providerID = node.Spec.ProviderID
-	}
-
-	instanceMetadataGetter := func(providerID string, nodeName string, node *v1.Node) (*cloudprovider.InstanceMetadata, error) {
-		if instancesV2, ok := cnc.cloud.InstancesV2(); instancesV2 != nil && ok {
-			return instancesV2.InstanceMetadata(ctx, node)
-		}
-
-		// If InstancesV2 not implement, use Instances.
-		instances, ok := cnc.cloud.Instances()
-		if !ok {
-			return nil, fmt.Errorf("failed to get instances from cloud provider")
-		}
-
-		nodeAddresses, err := getNodeAddressesByProviderIDOrName(ctx, instances, providerID, nodeName)
-		if err != nil {
-			return nil, err
-		}
-
-		instanceType, err := getInstanceTypeByProviderIDOrName(ctx, instances, providerID, nodeName)
-		if err != nil {
-			return nil, err
-		}
-
-		instanceMetadata := &cloudprovider.InstanceMetadata{
-			InstanceType:  instanceType,
-			NodeAddresses: nodeAddresses,
-		}
-
-		zones, ok := cnc.cloud.Zones()
-		if !ok {
-			return instanceMetadata, nil
-		}
-
-		zone, err := getZoneByProviderIDOrName(ctx, zones, providerID, node.Name)
-		if err != nil {
-			return nil, fmt.Errorf("failed to get zone from cloud provider: %v", err)
-		}
-
-		if zone.FailureDomain != "" {
-			instanceMetadata.Zone = zone.FailureDomain
-		}
-
-		if zone.Region != "" {
-			instanceMetadata.Region = zone.Region
-		}
-
-		return instanceMetadata, nil
-	}
-
-	instanceMeta, err := instanceMetadataGetter(providerID, node.Name, node)
-	if err != nil {
-		return nil, err
+	var nodeModifiers []nodeModifier
+	if node.Spec.ProviderID == "" && providerID != "" {
+		nodeModifiers = append(nodeModifiers, func(n *v1.Node) { n.Spec.ProviderID = providerID })
 	}
 
 	// If user provided an IP address, ensure that IP address is found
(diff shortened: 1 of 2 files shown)


wojtek-t:
LGTM, but unfortunately we are in code-freeze now and this needs to wait until we start accepting 1.20 PRs.

I would also like @andrewsykim to take a look.

staging/src/k8s.io/legacy-cloud-providers/gce/gce.go
```diff
@@ -661,7 +661,7 @@ func (g *Cloud) Instances() (cloudprovider.Instances, bool) {
 
 // InstancesV2 returns an implementation of InstancesV2 for Google Compute Engine.
 func (g *Cloud) InstancesV2() (cloudprovider.InstancesV2, bool) {
-	return nil, false
+	return g, true
```
andrewsykim, line 664: This should be done in a separate PR

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -405,7 +405,7 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
 // a node object with provider-specific information.
 // All of the returned functions are idempotent, because they are used in a retry-if-conflict
 // loop, meaning they could get called multiple times.
-func (cnc *CloudNodeController) getNodeModifiersFromCloudProvider(ctx context.Context, node *v1.Node) ([]nodeModifier, error) {
```
andrewsykim, line 408: I don't think it makes sense for this method to be returning `InstanceMetadata`, what was the reason for returning it here?

gongguan: Returning `InstanceMetadata` and delivering it to `updateNodeAddress`, then `updateNodeAddress` won't call cloud apis to get `InstanceMetadata` again.
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go#L393

staging/src/k8s.io/legacy-cloud-providers/gce/gce.go
gongguan: Since this change is tiny, I decided to put it in a separate commit now(this is also easy to revert).
How about my decision?

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
andrewsykim: Could we get `InstanceMetadata` in a separate method before and pass it into `getNodeModifiersFromCloudProvider` instead?

staging/src/k8s.io/legacy-cloud-providers/gce/gce.go
andrewsykim: Prefer separate PR so it's more searchable and to keep the discussions specific to GCE in a separate PR thread

gongguan: ok, I'll do it.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
gongguan: Sure we can, but we should move get providerID outside `getNodeModifiersFromCloudProvider` at the same time.
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go#L415-L436

Is it ok?

andrewsykim: I think so

gongguan: Done, PTAL.

staging/src/k8s.io/legacy-cloud-providers/gce/gce.go
gongguan: Reverted this change.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -405,63 +417,17 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
...
-	var (
-		nodeModifiers []nodeModifier
-		providerID    string
-		err           error
-	)
+func (cnc *CloudNodeController) getNodeModifiersFromCloudProvider(
+	ctx context.Context,
+	node *v1.Node,
+	providerID string,
+	providerIDModifier nodeModifier,
```
andrewsykim, line 424: remove this and just pass in getProviderID directly below?

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -244,13 +244,15 @@ func (cnc *CloudNodeController) updateNodeAddress(ctx context.Context, node *v1.
 		}, nil
 	}
 
-	instanceMeta, err := instanceMetadataGetter(node.Spec.ProviderID, node)
-	if err != nil {
-		utilruntime.HandleError(err)
-		return
+	if instanceMetadata == nil {
+		var err error
+		if instanceMetadata, err = instanceMetadataGetter(node.Spec.ProviderID, node); err != nil {
```
andrewsykim, line 249: remove `instanceMetadataGetter` and call getInstanceMetadata directly?

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -244,13 +244,15 @@ func (cnc *CloudNodeController) updateNodeAddress(ctx context.Context, node *v1.
 		}, nil
 	}
 
-	instanceMeta, err := instanceMetadataGetter(node.Spec.ProviderID, node)
-	if err != nil {
-		utilruntime.HandleError(err)
-		return
+	if instanceMetadata == nil {
```
andrewsykim, line 247: Thinking that the nil case should always result in an error and we should ensure that `getInstanceMetadata` is called prior to calling `updateNodeAddress`.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
gongguan: There is a difference between them: instanceMetadataGetter only get instance nodeAddresses, but getInstanceMetadata get both instance type and nodeAddresses.

gongguan: Maybe an option whether to get instance type make sense.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
gongguan: See https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go#L415-L436
We should handle providerID not set scenario and get this modifier before get instance metadata that's why I move this out and also pass providerID and related modifier.

andrewsykim: The modifer is very simple though:
```
modifier := func(n *v1.Node) {
    if n.Spec.ProviderID == "" {
        n.Spec.ProviderID = providerID
    }
}
````

We can define the modifer here if providerID is "".

File: staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go (hunk 5 above)
Comment (gongguan) [reply staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go]: Updated, maybe better now.

File: staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go (hunk 3 above)
Comment (gongguan) [reply staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go]: Thanks, looks good, simplified your suggestion and updated.

File: staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go [hunk 6]
```diff
@@ -405,63 +393,16 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
...
-
+func (cnc *CloudNodeController) getNodeModifiersFromCloudProvider(
+	ctx context.Context,
+	node *v1.Node,
+	providerID string,
+	instanceMeta *cloudprovider.InstanceMetadata,
+) ([]nodeModifier, error) {
+
+	var nodeModifiers []nodeModifier
 	if node.Spec.ProviderID == "" {
```
andrewsykim, line 404: There also needs to be a check for when the passed in providerID is "" which can happen if the provider returns `cloudprovider.NotImplemented` like in line 423

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -513,6 +454,55 @@ func (cnc *CloudNodeController) getNodeModifiersFromCloudProvider(ctx context.Co
...
+			// do not, the taint will be removed, and this will not be retried
+			return "", err
+		}
+	}
+
+	return node.Spec.ProviderID, nil
+}
+
+// getInstanceMetadata get instance type and nodeAddresses, use Instances if InstancesV2 is off.
+func (cnc *CloudNodeController) getInstanceMetadata(ctx context.Context, providerID string, nodeName string, needType bool) (*cloudprovider.InstanceMetadata, error) {
```
andrewsykim, line 479: I'm not sure about having `needType` here because when v2 is enabled the type is always returned which can be confusing.

andrewsykim:
FYI https://github.com/kubernetes/kubernetes/pull/93569

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
gongguan: `if node.Spec.ProviderID == "" && providerID != ""`, right?

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
gongguan: How about always getting instance type?

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
gongguan: Done.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
gongguan: > How about always getting instance type?

Done.

andrewsykim: yeah or maybe a `getNodeAddress` method that only sets `NodeAddress` in `InstanceMetadata`.

andrewsykim: Not sure yet actually -- let's leave it as is for now (always return instance type in getInstanceMetadata)

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -515,6 +457,53 @@ func (cnc *CloudNodeController) getNodeModifiersFromCloudProvider(ctx context.Co
...
+	// If InstancesV2 not implement, use Instances.
+	instances, ok := cnc.cloud.Instances()
+	if !ok {
+		return nil, fmt.Errorf("failed to get instances from cloud provider")
+	}
+	nodeAddresses, err := getNodeAddressesByProviderIDOrName(ctx, instances, node.Spec.ProviderID, node.Name)
+	if err != nil {
+		return nil, err
+	}
+	instanceType, err := getInstanceTypeByProviderIDOrName(ctx, instances, node.Spec.ProviderID, node.Name)
```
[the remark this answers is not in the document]
gongguan: Agree with you as we can't drop legacy Instances in a short time. Compared with reducing a http request in a frequently called function, it's worth to duplicate it.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -515,6 +457,53 @@ func (cnc *CloudNodeController) getNodeModifiersFromCloudProvider(ctx context.Co
 	return nodeModifiers, nil
 }
 
+func (cnc *CloudNodeController) getProviderID(ctx context.Context, node *v1.Node) (string, error) {
+	if node.Spec.ProviderID == "" {
```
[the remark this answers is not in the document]
gongguan: Done, also simplify following if and else conditions to `return providerID, err`, not sure comments in else condition should be adjusted.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -515,6 +457,74 @@ func (cnc *CloudNodeController) getNodeModifiersFromCloudProvider(ctx context.Co
...
+	}
+
+	// if the cloud provider being used supports provider IDs, we want
+	// to propagate the error so that we re-try in the future; if we
+	// do not, the taint will be removed, and this will not be retried
+	return providerID, err
+}
+
+// getInstanceMetadata get instance type and nodeAddresses, use Instances if InstancesV2 is off.
+func (cnc *CloudNodeController) getInstanceMetadata(ctx context.Context, node *v1.Node) (*cloudprovider.InstanceMetadata, error) {
```
gongguan, line 481: @andrewsykim  one concern here: when providerID not passed here, `InstanceMetadata`, `getNodeAddressesByProviderIDOrName` and `getInstanceTypeByProviderIDOrName` could still work fine, right?

gongguan:
Ping @andrewsykim , could you please take another look?

andrewsykim:
Sorry for the delay @gongguan -- I'll review this soon

gongguan:
> Sorry for the delay @gongguan -- I'll review this soon

No sorry, there is something wrong I'll fix it now, you can review after ci passed.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
gongguan: > @andrewsykim one concern here

No more since I set ProviderID in line 474.

gongguan:
@andrewsykim ready for review now :)

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -155,7 +155,12 @@ func (cnc *CloudNodeController) UpdateNodeStatus(ctx context.Context) {
 	}
 
 	for i := range nodes.Items {
-		cnc.updateNodeAddress(ctx, &nodes.Items[i])
+		instanceMetadata, err := cnc.getInstanceNodeAddresses(ctx, &nodes.Items[i])
+		if err != nil {
+			klog.Errorf("Error get instance metadata: %v", err)
```
andrewsykim, line 160: nit: "Error getting instance metadata for node addresses: %v"

andrewsykim (commented):
LGTM, minor comment about error logging

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
gongguan: done

gongguan:
> LGTM, minor comment about error logging

@andrewsykim addressed and rebased, PTAL.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -407,86 +395,16 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
...
+) ([]nodeModifier, error) {
 
-		return instanceMetadata, nil
-	}
-
-	instanceMeta, err := instanceMetadataGetter(providerID, node.Name, node)
-	if err != nil {
-		return nil, err
+	var nodeModifiers []nodeModifier
+	if node.Spec.ProviderID != providerID {
```
andrewsykim, line 406: I think this check needs to be updated to
```
if node.Spec.ProviderID == "" && providerID != "" {
```

otherwise we might try to set an existing `node.Spec.ProviderID` to "". What do you think?

andrewsykim (commented):
Sorry, had one more comment

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
gongguan: Done, your suggestion is more legible.

> otherwise we might try to set an existing `node.Spec.ProviderID` to "". What do you think?

No, in line 454(getProviderID function), providerID will be set as node.Spec.ProviderID if node.Spec.ProviderID not nil. That means node.Spec.ProviderID not equal to providerID only occurs when node.Spec.ProviderID not set.

andrewsykim: > No, in line 454(getProviderID function), providerID will be set as node.Spec.ProviderID if node.Spec.ProviderID not nil. That means node.Spec.ProviderID not equal to providerID only occurs when node.Spec.ProviderID not set.

Good point

gongguan:
`should set TCP CLOSE_WAIT timeout [Privileged]` failed.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -155,7 +155,12 @@ func (cnc *CloudNodeController) UpdateNodeStatus(ctx context.Context) {
 	}
 
 	for i := range nodes.Items {
-		cnc.updateNodeAddress(ctx, &nodes.Items[i])
+		instanceMetadata, err := cnc.getInstanceNodeAddresses(ctx, &nodes.Items[i])
+		if err != nil {
+			klog.Errorf("Error getting instance metadata for node addresses: %v", err)
+			return
```
andrewsykim, line 161: this should probably be `continue`?

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -363,9 +341,19 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
 		return
 	}
 
-	// TODO: getNodeModifiersFromCloudProvider and updateNodeAddress both call cloud api to get instanceMetadata,
-	// get instanceMetadata and pass it to getNodeModifiersFromCloudProvider and updateNodeAddress which reduces api calls.
-	nodeModifiers, err := cnc.getNodeModifiersFromCloudProvider(ctx, curNode)
+	providerID, err := cnc.getProviderID(ctx, curNode)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("failed to initialize node %s at cloudprovider: %v", node.Name, err))
```
andrewsykim, line 346: should error message should probably be specific to getting the providerID:
```
utilruntime.HandleError(fmt.Errorf("failed to get provider ID for node %s: %v", node.Name, err))
```

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -363,9 +341,19 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
...
+		return
+	}
+
+	instanceMetadata, err := cnc.getInstanceMetadata(ctx, providerID, curNode)
+	if err != nil {
+		utilruntime.HandleError(err)
```
andrewsykim, line 352: Wrap this error similar to other calls to HandleError here
```
utilruntime.HandleError(fmt.Errorf("failed to get instance metadata for node %s: %v", node.Name, err))
```

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
gongguan: You're right, fixed.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
gongguan: done

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
gongguan: done

andrewsykim (commented):
/approve
/lgtm

Thanks @gongguan

PR #93286, opened 2020-07-21
kube-aggregator: sync startup with APIService handlers

sttts:
There is a race of the apiserver setting up the api service handlers and reporting readiness. This PR adds a synchronization between two via a post-start hook.

/kind bug

```release-note
NONE
```

### DIFF
--- a/staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
+++ b/staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
@@ -18,11 +18,18 @@ package apiserver
 
 import (
 	"fmt"
+	"k8s.io/klog/v2"
 	"net/http"
+	"strings"
+	"sync"
 	"time"
 
+
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/wait"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	"k8s.io/apiserver/pkg/server/egressselector"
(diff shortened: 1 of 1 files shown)


lavalamp:
Some sort of a test would be nice?

staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
```diff
@@ -263,6 +271,34 @@ func (c completedConfig) NewWithDelegate(delegationTarget genericapiserver.Deleg
...
+			expected := sets.NewString()
+			for _, apiservice := range apiservices {
+				if v1helper.IsAPIServiceConditionTrue(apiservice, v1.Available) {
+					expected.Insert(apiservice.Name)
+				}
+			}
+
+			s.proxyHandlersLock.Lock()
+			handled := sets.StringKeySet(s.proxyHandlers)
+			s.proxyHandlersLock.Unlock()
```
yue9944882, line 291: nit: defer?

staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
```diff
@@ -331,9 +370,16 @@ func (s *APIAggregator) AddAPIService(apiService *v1.APIService) error {
 	}
 	proxyHandler.updateAPIService(apiService)
 	if s.openAPIAggregationController != nil {
-		s.openAPIAggregationController.AddAPIService(proxyHandler, apiService)
+		// this is calling a controller.  It should already handle being async.
+		go func() {
```
yue9944882, line 374: this is suspicious. the `AddAPIService` is merely pushing tasks into queue, we don't need to spawn a new routine, is it?

staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
```diff
@@ -18,11 +18,18 @@ package apiserver
 
 import (
 	"fmt"
+	"k8s.io/klog/v2"
```
marun, line 21: Import ordering?

PR #93293, opened 2020-07-21
kubeconfig: add explicit path, if specified to loading precedence

soltysh:
**What type of PR is this?**
/kind bug

Currently if you invoke:
```
kubectl --kubeconfig=/home/user/kubeconfig config use-context new-context
```
The actual file locked during this process is not the one specified above. This is because the `ModifyConfig` relies on the file precedence from here: https://github.com/kubernetes/kubernetes/blob/b6174e605fd6e23b48131dec9e45043b38372c3c/staging/src/k8s.io/client-go/tools/clientcmd/config.go#L169 which does not include the one specified through `--kubeconfig`. But `writeCurrentContext` is aware of that and checks this accordingly, see https://github.com/kubernetes/kubernetes/blob/b6174e605fd6e23b48131dec9e45043b38372c3c/staging/src/k8s.io/client-go/tools/clientcmd/config.go#L390.
Similarly in [loader.go](https://github.com/kubernetes/kubernetes/blob/b6174e605fd6e23b48131dec9e45043b38372c3c/staging/src/k8s.io/client-go/tools/clientcmd/loader.go#L187) `Load` is aware of explicit path, but `GetLoadingRules` ignores it.

/assign @deads2k @liggitt 

/priority critical-urgent 
/sig cli

<!--
```release-note
NONE
```

### DIFF
--- a/staging/src/k8s.io/client-go/tools/clientcmd/config.go
+++ b/staging/src/k8s.io/client-go/tools/clientcmd/config.go
@@ -83,10 +83,13 @@ func (o *PathOptions) GetEnvVarFiles() []string {
 }
 
 func (o *PathOptions) GetLoadingPrecedence() []string {
+	if o.IsExplicitFile() {
+		return []string{o.GetExplicitFile()}
+	}
+
 	if envVarFiles := o.GetEnvVarFiles(); len(envVarFiles) > 0 {
 		return envVarFiles
 	}
-
 	return []string{o.GlobalFile}
 }
 
--- a/staging/src/k8s.io/client-go/tools/clientcmd/loader.go
+++ b/staging/src/k8s.io/client-go/tools/clientcmd/loader.go
@@ -304,6 +304,10 @@ func (rules *ClientConfigLoadingRules) Migrate() error {
 
 // GetLoadingPrecedence implements ConfigAccess
 func (rules *ClientConfigLoadingRules) GetLoadingPrecedence() []string {
+	if len(rules.ExplicitPath) > 0 {
+		return []string{rules.ExplicitPath}
+	}
+
 	return rules.Precedence
 }
 
--- a/staging/src/k8s.io/client-go/tools/clientcmd/client_config_test.go
+++ b/staging/src/k8s.io/client-go/tools/clientcmd/client_config_test.go
@@ -287,7 +287,7 @@ func TestModifyContext(t *testing.T) {
 
 	// there should now be two contexts
 	if len(startingConfig.Contexts) != len(expectedCtx) {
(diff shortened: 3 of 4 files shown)


staging/src/k8s.io/client-go/tools/clientcmd/config.go
```diff
@@ -83,6 +83,10 @@ func (o *PathOptions) GetEnvVarFiles() []string {
 }
 
 func (o *PathOptions) GetLoadingPrecedence() []string {
+	if o.IsExplicitFile() {
```
deads2k, line 86: this looks correct to me for the same reason

staging/src/k8s.io/client-go/tools/clientcmd/loader.go
```diff
@@ -304,6 +304,10 @@ func (rules *ClientConfigLoadingRules) Migrate() error {
 
 // GetLoadingPrecedence implements ConfigAccess
 func (rules *ClientConfigLoadingRules) GetLoadingPrecedence() []string {
+	if len(rules.ExplicitPath) > 0 {
+		return []string{rules.ExplicitPath}
```
deads2k, line 308: seems like when you set this field, perhaps you should reset rules.Precendence?  At least from genericclioptions?

soltysh:
@deads2k comments addressed, I've added also some basic test case covering loading precedence. I'm working on a follow-up to remove the code duplication that is between `config.go` and `loader.go`.

soltysh:
/hold 
for more debugging

staging/src/k8s.io/client-go/tools/clientcmd/loader.go
soltysh: I tried, but it looks like this is breaking all of our test-cmd, so I don't think it's the right fix. With proper LoadingPrecedence we should be good, though.

soltysh:
/hold cancel
I'll return to this topic after I'm back from PTO, so I'm clearing milestone since this can wait and will eventually be cherry-picked to 1.19.
/milestone clear

staging/src/k8s.io/client-go/tools/clientcmd/loader.go
deads2k: it is also special cased above in `Load` so I guess I can live with it.  It doesn't seem awesome though.

soltysh: I'm planning to squashing all `GetLoadingPrecedence` methods into single execution path, to avoid duplicated code and problems like the one this PR is fixing. But I wanted to make sure that approach isn't breaking anything.

PR #93299, opened 2020-07-21
[WIP]  Support Windows to run sandbox containers by Windows logic not linux logic

### LINKED ISSUE #92963: Windows pods need to support `RunAsUserName` like Linux pods, not just work containers, not pause containers
**What happened**:
If we specify `SecurityContext.WindowsOptions.RunAsUserName` when the pod is created, the windows pod will work properly, but check the running username of the pause container and find that it is still the default username of the pause image itself, not `SecurityContext.WindowsOptions.RunAsUserName`

Once the Linux pod is running, specify `RunAsUser`, and all the containers to which the pod belongs will be run by changing the identity, including the pause container

The behavior of the windows pod should be consistent with the linux pod, otherwise it will cause unexpected security problems (at present I have not thought of any specific security risks)

Feel free to run a windows pod, specify `SecurityContext.WindowsOptions.RunAsUserName`, and then check the running identity of all containers in the pod can be reproduced

According to the code, when creating sandboxpod, it does not cover the corresponding logic of windows. If you need to implement this function, you need to make the following changes

- When sandboxpod is created, it needs to handle `WindowsOptions`
- CRI needs to deal with this logic and currently only receives linux related parameters

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

wawa0210:
**What type of PR is this?**

/kind feature

Support users to pass in different usernames to run sandbox container and work container through `SecurityContext.WindowsOptions.RunAsUserName`. Currently working containers are supported, but sandbox does not

Fixes #92963

```release-note

```

```docs

```

/sig windows


wawa0210:
/hold

This pr is still in progress

pkg/kubelet/dockershim/docker_sandbox_windows.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	dockertypes "github.com/docker/docker/api/types"
+	dockercontainer "github.com/docker/docker/api/types/container"
+	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+)
+
+// applySandboxOptions applies LinuxPodSandboxConfig to dockercontainer.HostConfig and dockercontainer.ContainerCreateConfig.
+func (ds *dockerService) applySandboxOptions(hc *dockercontainer.HostConfig, c *runtimeapi.PodSandboxConfig, createConfig *dockertypes.ContainerCreateConfig, image string, separator rune) error {
+	wc := c.GetWindows()
```
ddebroy, line 29: Is there a corresponding PR over in https://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtime/v1alpha2/api.pb.go to introduce a GetWindows() for runtimeapi.PodSandboxConfig? Note that it only has GetLinux() today.

wawa0210: I added the `GetWindows()` method to the current pr's `staging/src/k8s.io/cri-api/pkg/apis/runtime/v1alpha2/api.proto` file.

jsturtevant: Should coordinate changes to Containerd to support this additional CRI field as well.

wawa0210:
/close

This pr is more complicated than expected, I will try to split it into several prs

PR #93305, opened 2020-07-21
Mount kubelet and container runtime rootdir on LSSD

alculquicondor:
**What type of PR is this?**

/kind feature

When environment variable NODE_LOCAL_SSD_EPHEMERAL=true,
create a RAID 0 array on all attached Local SSDs on NVMe interfaces to mount:

- kubelet root dir
- container runtime root dir
- pod logs dir

Those directories account for all ephemeral storage.

```release-note
NONE
```


cluster/gce/gci/configure-helper.sh
```diff
@@ -425,6 +429,56 @@ function ensure-local-ssds() {
...
+
+  # Create RAID array
+  local md_device=(/dev/md/*)
+  md_device=${md_device[0]}
+  if [ ! -e "$md_device" ]; then
+    md_device="/dev/md/0"
+    echo "y" | mdadm --create "${md_device}" --level=0 --raid-devices=${#devices[@]} ${devices[@]}
+  fi
+  local ephemeral_mountpoint="/mnt/disks/kube-ephemeral-ssd"
+  safe-format-and-mount "${md_device}" "${ephemeral_mountpoint}"
```
msau42, line 463: Are the mount options and directory permissions what we want, such as the "a+w" permissions?

cluster/gce/gci/configure-helper.sh
```diff
@@ -425,6 +429,56 @@ function ensure-local-ssds() {
...
+
+  # mount container runtime root dir on SSD
+  local container_runtime="${CONTAINER_RUNTIME:-docker}"
+  systemctl stop ${container_runtime}
+  # Some images mount the container runtime root dir.
+  umount "/var/lib/${container_runtime}" || true
+  mkdir -p "${ephemeral_mountpoint}/${container_runtime}"
+  safe-bind-mount "${ephemeral_mountpoint}/${container_runtime}" "/var/lib/${container_runtime}"
```
msau42, line 471: Same question here wrt permissions and mount options

cluster/gce/gci/configure-helper.sh
```diff
@@ -376,6 +376,10 @@ function mount-ext(){
 # Local ssds, if present, are mounted or symlinked to their appropriate
 # locations
 function ensure-local-ssds() {
+  if [ "${NODE_LOCAL_SSDS_EPHEMERAL:-false}" == "true" ]; then
```
msau42, line 379: Do we need more plumbing of this env? Like in configure-default.sh, configure-test.sh?

alculquicondor: What are the effects or those?

cluster/gce/gci/configure-helper.sh
```diff
@@ -425,6 +429,56 @@ function ensure-local-ssds() {
...
+  if [ ! -e ${devices[0]} ]; then
+    echo "No local SCSI SSD disks found."
+    devices=()
+  fi
+  # Get nvme devices
+  for ssd in "/dev/nvme*"; do
+    if [ -e "${ssd}" ]; then
+      # This workaround to find if the NVMe device is a disk is required because
+      # the existing Google images does not expose NVMe devices in /dev/disk/by-id
+      if [[ `udevadm info --query=property --name=${ssd} | grep DEVTYPE | sed "s/DEVTYPE=//"` == "disk" ]]; then
```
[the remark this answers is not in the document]
alculquicondor: It was copied over from `ensure-local-ssds`

cluster/gce/gci/configure-helper.sh
```diff
@@ -425,6 +429,56 @@ function ensure-local-ssds() {
   done
 }
 
+function ensure-local-ssds-ephemeral-storage() {
+  # Get scsi devices
+  local devices=(/dev/disk/by-id/google-local-ssd-*)
+  if [ ! -e ${devices[0]} ]; then
+    echo "No local SCSI SSD disks found."
+    devices=()
```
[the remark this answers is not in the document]
alculquicondor: It's kind of the intention, but should we enforce it here?

cluster/gce/gci/configure-helper.sh
alculquicondor: yes, verified.

cluster/gce/gci/configure-helper.sh
alculquicondor: Yes, verified.

cluster/gce/gci/configure-helper.sh
```diff
@@ -425,6 +429,56 @@ function ensure-local-ssds() {
...
+  if [ ! -e "$md_device" ]; then
+    md_device="/dev/md/0"
+    echo "y" | mdadm --create "${md_device}" --level=0 --raid-devices=${#devices[@]} ${devices[@]}
+  fi
+  local ephemeral_mountpoint="/mnt/disks/kube-ephemeral-ssd"
+  safe-format-and-mount "${md_device}" "${ephemeral_mountpoint}"
+
+  # mount container runtime root dir on SSD
+  local container_runtime="${CONTAINER_RUNTIME:-docker}"
+  systemctl stop ${container_runtime}
```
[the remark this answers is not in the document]
alculquicondor: Fixed.

cluster/gce/gci/configure-helper.sh
```diff
@@ -425,6 +429,56 @@ function ensure-local-ssds() {
   done
 }
 
+function ensure-local-ssds-ephemeral-storage() {
```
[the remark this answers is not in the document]
alculquicondor: Done

cluster/gce/gci/configure-helper.sh
```diff
@@ -425,6 +429,56 @@ function ensure-local-ssds() {
...
+  fi
+
+  # Create RAID array
+  local md_device=(/dev/md/*)
+  md_device=${md_device[0]}
+  if [ ! -e "$md_device" ]; then
+    md_device="/dev/md/0"
+    echo "y" | mdadm --create "${md_device}" --level=0 --raid-devices=${#devices[@]} ${devices[@]}
+  fi
+  local ephemeral_mountpoint="/mnt/disks/kube-ephemeral-ssd"
```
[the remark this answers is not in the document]
alculquicondor: The rest of /mnt is read-only. I see this for `/mnt/disks`:

```
tmpfs /mnt/disks tmpfs rw,relatime,size=256k,mode=755 0 0
```

One possibility would be `/mnt/stateful_partition/ephemeral_storage`, but then we are putting it inside the boot disk.

cluster/gce/gci/configure-helper.sh
```diff
@@ -425,6 +429,56 @@ function ensure-local-ssds() {
...
+      # the existing Google images does not expose NVMe devices in /dev/disk/by-id
+      if [[ `udevadm info --query=property --name=${ssd} | grep DEVTYPE | sed "s/DEVTYPE=//"` == "disk" ]]; then
+        devices+=("${ssd}")
+      fi
+    else
+      echo "No local NVMe SSD disks found."
+    fi
+  done
+  if [ "${#devices[@]}" -eq 0 ]; then
+    return
```
mattcary, line 455: Can we error or log or something here? I don't know how we can surface node startup problems, but if there is one it would be nice if it were possible to see it clearly in logs or whatever.

alculquicondor: We are already logging in the lines above. If there are no SSDs, I don't think it should be considered an error.

cluster/gce/gci/configure-helper.sh
msau42: Those are used for kube-up.sh and kubetest

cluster/gce/gci/configure-helper.sh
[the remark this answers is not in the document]
msau42: I think the existing logic doesn't handle PD as nvme and assumes all nvme devices are local SSDs. @mattcary do you want to restrict this new ephemeral option to only scsi PD boot disk for now, and update this logic to handle nvme PD boot disk later?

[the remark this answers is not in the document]
alculquicondor: We want nvme support. Is there any doc on how to detect nvme PDs?

cluster/gce/gci/configure-helper.sh
alculquicondor: Done

cluster/gce/gci/configure-helper.sh
mattcary: ok, I agree those logs should be enough.

cluster/gce/gci/configure-helper.sh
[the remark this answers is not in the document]
mattcary: It seems like the conservative choice would be to restrict to one or the other. Allowing both maybe has the possibility of the node getting wedged in a weird way. Being restrictive on how we do things might be more reliable.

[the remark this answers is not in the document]
alculquicondor: Any preference on which should take preference if both happen to be present? Or I could just remove the checks for scsi.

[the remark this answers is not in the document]
mattcary: Preferring NVMe seems right.

If this happens, does it mean that someone has done some weird manual configuration? So the fact that the ephemeral storage is less than the total attache LSSD won't be a surprise.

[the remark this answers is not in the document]
alculquicondor: Yeah, it would only be due to manual overrides.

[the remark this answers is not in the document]
alculquicondor: Removed scsi support

cluster/gce/gci/configure-helper.sh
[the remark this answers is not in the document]
alculquicondor: using ID_MODEL=nvme_card

cluster/gce/gci/configure-helper.sh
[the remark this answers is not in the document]
alculquicondor: Done

cluster/gce/gci/configure-helper.sh
```diff
@@ -406,25 +410,83 @@ function ensure-local-ssds() {
...
+      # This workaround to find if the NVMe device is a local disk is required
+      # because the existing Google images does not expose NVMe local SSDs in
+      # /dev/disk/by-id
+      if [[ `udevadm info --query=property --name=${ssd} | grep DEVTYPE | sed "s/DEVTYPE=//"` != "disk" ]]; then
+        continue
+      fi
+      if [[ `udevadm info --query=property --name=${ssd} | grep ID_MODEL | sed "s/ID_MODEL=//"` != "nvme_card" ]]; then
+        continue
       fi
+      local devicenum=`echo ${ssd} | sed -e 's/\/dev\/nvme0n\([0-9]*\)/\1/'`
```
alculquicondor, line 424: I think we can't rely on it being `nvme0n`, but we can fix in a followup

mattcary: Yeah let's not change existing behavior (which is what ensure-local-ssds is for, right?)

mattcary: (although nvme[0-9]* would probably do the trick?)

alculquicondor: assuming that doesn't produce a match group. I need to refresh my `sed` knowledge

cluster/gce/gci/configure-helper.sh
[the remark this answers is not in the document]
alculquicondor: Updated to lsblk

cluster/gce/gci/configure-helper.sh
alculquicondor: I reverted changes in the existing function. Let's fix that in a followup

mattcary:
/lgtm
as well

alculquicondor:
Friendly ping @cheftako 

Giving the freeze, we are not looking for merging it now, but just to get early feedback

cluster/gce/util.sh
```diff
@@ -1215,6 +1215,7 @@ CONTAINER_RUNTIME_TEST_HANDLER: $(yaml-quote ${CONTAINER_RUNTIME_TEST_HANDLER:-}
 UBUNTU_INSTALL_CONTAINERD_VERSION: $(yaml-quote ${UBUNTU_INSTALL_CONTAINERD_VERSION:-})
 UBUNTU_INSTALL_RUNC_VERSION: $(yaml-quote ${UBUNTU_INSTALL_RUNC_VERSION:-})
 NODE_LOCAL_SSDS_EXT: $(yaml-quote ${NODE_LOCAL_SSDS_EXT:-})
+NODE_LOCAL_SSDS_EPHEMERAL: $(yaml-quote ${NODE_LOCAL_SSDS_EPHEMERAL:-})
```
cheftako, line 1218: Thanks :D

alculquicondor:
Turns out the tool complains about the creation of a single device array.

I'm skipping the array creation for this case. PTAL

cluster/gce/gci/configure-helper.sh
```diff
@@ -449,15 +449,19 @@ function ensure-local-ssds-ephemeral-storage() {
...
 
-  # Create RAID array
-  local md_device=(/dev/md/*)
-  md_device=${md_device[0]}
-  if [ ! -e "$md_device" ]; then
-    md_device="/dev/md/0"
-    echo "y" | mdadm --create "${md_device}" --level=0 --raid-devices=${#devices[@]} ${devices[@]}
+  local device="${devices[0]}"
+  if [ "${#devices[@]}" -ne 1 ]; then
+    local device=(/dev/md/*)
```
msau42, line 454: Is this going to set a new variable locally, and not modify the one at L452?

alculquicondor: Oddly enough, it worked in manual testing. Fixed.

alculquicondor:
/hold

I'm checking some kube-proxy issues

alculquicondor:
/hold cancel

Fixed problems with preloaded images. PTAL

cluster/gce/gci/configure-helper.sh
```diff
@@ -479,7 +479,11 @@ function ensure-local-ssds-ephemeral-storage() {
   systemctl stop ${container_runtime}
   # Some images mount the container runtime root dir.
   umount "/var/lib/${container_runtime}" || true
-  mkdir -p "${ephemeral_mountpoint}/${container_runtime}"
+  # Move the container runtime's directory to the new location to preserve
+  # preloaded images.
+  if [ ! -d "${ephemeral_mountpoint}/${container_runtime}" ]; then
+    mv "/var/lib/${container_runtime}" "${ephemeral_mountpoint}/${container_runtime}"
```
msau42, line 485: We unmounted "/var/lib/${container_runtime}" above. Is there any data here to move?

alculquicondor: They are rw remounts. I have yet to verify in Ubuntu, though. I'll get back to this.

mattcary: Sorry, I don't follow. Do you mean that the cases where there are preloaded images, /var/lib/docker isn't mounted so the unmount is a no-op?

alculquicondor: I rechecked. The OS images that mount are the containerd ones. And it's just a RW remount in the same disk. We need to unmount prior to moving the folder.

This is independent of having preloaded images or not.

alculquicondor:
There are new shell check requirements. PTAL at last 2 commits

cluster/gce/gci/configure-helper.sh
mattcary: I'm still not following. Why do we need to unmount? If we're concerned about someone writing to it during the move, the umount isn't enough because it will silently fail if the dir is already in use (ie there's a race).

alculquicondor: We need to unmount to be able to move the data. Otherwise we would get a "resource busy" error. Nothing is writing to it, as the container runtime is stopped.

mattcary: finally I understand, bind-mounted dirs make that resource busy error. thx

cluster/gce/gci/configure-helper.sh
```diff
@@ -436,6 +440,62 @@ function ensure-local-ssds() {
...
+    echo "No local NVMe SSD disks found."
+    return
+  fi
+
+  local device="${devices[0]}"
+  if [ "${#devices[@]}" -ne 1 ]; then
+    seen_arrays=(/dev/md/*)
+    device=${seen_arrays[0]}
+    echo "Setting RAID array with local SSDs on device ${device}"
+    if [ ! -e "$device" ]; then
```
mattcary, line 468: Why can we assume an existing raid array has to be our local SSDs?

cluster/gce/gci/configure-helper.sh
```diff
@@ -436,6 +440,62 @@ function ensure-local-ssds() {
...
+    fi
+  fi
+
+  local ephemeral_mountpoint="/mnt/stateful_partition/kube-ephemeral-ssd"
+  safe-format-and-mount "${device}" "${ephemeral_mountpoint}"
+
+  # mount container runtime root dir on SSD
+  local container_runtime="${CONTAINER_RUNTIME:-docker}"
+  systemctl stop "$container_runtime"
+  # Some images mount the container runtime root dir.
```
mattcary, line 480: Would it be more precise to say "remount" here? That gives a better hint as to what's going on IMHO.

alculquicondor: Done

cluster/gce/gci/configure-helper.sh
mattcary: answer: we have to, because it's too complicated to figure out what an existing raid is from when the node is restarted.

alculquicondor: There's no other mechanism that creates RAID arrays in the startup script.

mattcary:
/lgtm

fwiw

alculquicondor:
Rebased and squashed

@cheftako

mattcary:
Still 
/lgtm

alculquicondor:
ping @cheftako 

/retest

PR #93307, opened 2020-07-21
Ensure runtimeCache contains all observed started containers on pod delete

### LINKED ISSUE #93210: MirrorPodWithGracePeriod test flaking in 5% of pull-kubernetes-node-e2e runs

liggitt:
**What type of PR is this?**
/kind bug
/kind flake

Ensures the runtimeCache used to send graceful deletion signals to running containers is at least as fresh as the most recent container start observed by the kubelet and reported to the API.

See discussion in https://github.com/kubernetes/kubernetes/pull/93261 and on https://github.com/liggitt/kubernetes/commit/da1f636514251fa84d89471d1f66c9fb20909d11#r40791980

Fixes #93210 

```release-note
NONE
```

/cc @sjenning @dashpole 
/sig node
/milestone v1.19

### DIFF
--- a/pkg/kubelet/BUILD
+++ b/pkg/kubelet/BUILD
@@ -29,6 +29,7 @@ go_library(
         "reason_cache.go",
         "runonce.go",
         "runtime.go",
+        "time_cache.go",
         "util.go",
         "volume_host.go",
     ],
@@ -179,6 +180,7 @@ go_test(
         "pod_workers_test.go",
         "reason_cache_test.go",
         "runonce_test.go",
+        "time_cache_test.go",
     ],
     embed = [":go_default_library"],
     deps = [
@@ -251,6 +253,7 @@ go_test(
         "//staging/src/k8s.io/client-go/util/testing:go_default_library",
         "//staging/src/k8s.io/component-base/featuregate/testing:go_default_library",
         "//staging/src/k8s.io/component-base/version:go_default_library",
+        "//vendor/github.com/golang/groupcache/lru:go_default_library",
         "//vendor/github.com/google/cadvisor/info/v1:go_default_library",
         "//vendor/github.com/google/cadvisor/info/v2:go_default_library",
         "//vendor/github.com/stretchr/testify/assert:go_default_library",
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -525,6 +525,7 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 		experimentalHostUserNamespaceDefaulting: utilfeature.DefaultFeatureGate.Enabled(features.ExperimentalHostUserNamespaceDefaultingGate),
 		keepTerminatedPodVolumes:                keepTerminatedPodVolumes,
 		nodeStatusMaxImages:                     nodeStatusMaxImages,
+		lastContainerStartedTime:                newTimeCache(),
 	}
 
 	if klet.cloud != nil {
@@ -974,6 +975,9 @@ type Kubelet struct {
 
 	// lastStatusReportTime is the time when node status was last reported.
 	lastStatusReportTime time.Time
+
+	// lastContainerStartedTime is the time of the last ContainerStarted event observed per pod
+	lastContainerStartedTime *timeCache
 
 	// syncNodeStatusMux is a lock on updating the node status, because this path is not thread-safe.
 	// This lock is used by Kubelet.syncNodeStatus function and shouldn't be used anywhere else.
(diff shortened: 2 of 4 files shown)


sjenning:
We'll probably want to do a few runs of node-e2e to look for the flake.  We can use the WIP hold to keep it out until we feel it is tested sufficiently.
/lgtm
/test pull-kubernetes-node-e2e

liggitt:
/retest
fun with CI pod resource requests

liggitt:
Green: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/93307/pull-kubernetes-node-e2e/1285701973324075008/
/test pull-kubernetes-node-e2e

liggitt:
Green 2: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/93307/pull-kubernetes-node-e2e/1285718078499328000/

/retest
/test pull-kubernetes-node-e2e

liggitt:
green node e2e run 3

/retest
/test pull-kubernetes-node-e2e

liggitt:
removing WIP, adding hold until we're happy
/hold

liggitt:
green node e2e run 4

/test pull-kubernetes-node-e2e

liggitt:
green run number 5

/retest
/test pull-kubernetes-node-e2e

liggitt:
I ran this test locally 60 times in parallel:

```sh
make WHAT=test/e2e_node/e2e.test
hack/local-up-cluster.sh &
for i in $( seq 1 60 ); do
  sudo `which e2e.test` --ginkgo.focus=MirrorPodWithGracePeriod --start-services=false --stop-services=false 2>&1 | egrep 'Failed|not found' &
done
```

I consistently got one or more of the `not found` failures running against master, and got 100% success running that scenario 5 times against this branch.

liggitt:
/hold cancel

I'm satisfied with the CI runs and my local runs

tedyu:
Thanks Jordan for the fix.

Since the introduction of lastContainerStartedTime is for graceful termination where only pod with mirror pod is concerned, I think we can narrow the call to lastContainerStartedTime.Add() for pods with mirror pod.

I was experimenting with the following change:
```
diff --git a/pkg/kubelet/kubelet.go b/pkg/kubelet/kubelet.go
index 051671a9fc8..943ef5c2b19 100644
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -1665,6 +1665,7 @@ func (kl *Kubelet) deletePod(pod *v1.Pod) error {
                if err := kl.runtimeCache.ForceUpdateIfOlder(lastContainerStarted); err != nil {
                        return fmt.Errorf("error updating containers: %v", err)
                }
+               kl.lastContainerStartedTime.Remove(pod.UID)
        }
        // Runtime cache may not have been updated to with the pod, but it's okay
        // because the periodic cleanup routine will attempt to delete again later.
@@ -1854,7 +1855,16 @@ func (kl *Kubelet) syncLoopIteration(configCh <-chan kubetypes.PodUpdate, handle
                        // record the most recent time we observed a container start for this pod.
                        // this lets us selectively invalidate the runtimeCache when processing a delete for this pod
                        // to make sure we don't miss handling graceful termination for containers we reported as having started.
-                       kl.lastContainerStartedTime.Add(e.ID, time.Now())
+                       if pod, ok := kl.podManager.GetPodByUID(e.ID); ok {
+                               if _, found := kl.podManager.GetMirrorPodByPod(pod); found {
+                                       klog.Infof("recording start time for %q / %q", pod.Namespace, pod.Name)
+                                       kl.lastContainerStartedTime.Add(e.ID, time.Now())
+                               } else {
+                                       klog.V(4).Infof("didn't find mirror pod for %q", pod.Name)
+                               }
+                       } else {
+                               klog.V(4).Infof("didn't find pod for %q", e.ID)
+                       }
                }
                if isSyncPodWorthy(e) {
                        // PLEG event for a pod; sync it.
```
In Kubelet#deletePod, we can remove the entry when runtimeCache is updated.

Please let me know what you think.

Thanks

liggitt:
> Since the introduction of lastContainerStartedTime is for graceful termination where only pod with mirror pod is concerned, I think we can narrow the call to lastContainerStartedTime.Add() for pods with mirror pod.

This seems like it would also apply to graceful termination of non-mirror pods.

> In Kubelet#deletePod, we can remove the entry when runtimeCache is updated.

Since the time cache is a small LRU, letting the recorded time just get evicted seems simpler to me

PR #93310, opened 2020-07-21
add instrumentation approvers as per SIG governance

logicalhan:
**What type of PR is this?**

/kind cleanup

This cleans up an emeritus approver and adds the approvers who should be here (from SIG governance and @RainbowMango who is an owner on all instrumentation directories anyway). 

```release-note
NONE
```

/sig instrumentation


RainbowMango:
Thanks! Even though it's not a must for 1.19 but it can't hurt. So let's give it the necessary labels to push forward.
/milestone v1.19
/priority backlog

PR #93311, opened 2020-07-21
Add bootstrap policy for monitoring endpoints

logicalhan:
We enable metrics by default, but that doesn't mean we should have (nor encourage) full cluster-admin access to use those endpoints. 

Also, having a bootstrapping role defined for /metrics and health check endpoints means that we have access to instrumentation earlier in the apiserver startup lifecycle, which is desirable when you have larger clusters (which have correspondingly longer startups) or in the case that you have failures in the startup sequence before (and possibly preventing) new policies from being applied. 

**What type of PR is this?**

/kind cleanup

```release-note
Adds a bootstrapping ClusterRole, ClusterRoleBinding and group for /metrics, /livez/*, /readyz/*, & /healthz/* endpoints. 
```

/sig instrumentation api-machinery auth

### DIFF
--- a/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go
+++ b/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go
@@ -196,7 +196,8 @@ func ClusterRoles() []rbacv1.ClusterRole {
 			},
 		},
 		{
-			// a role which provides just enough power to determine if the server is ready and discover API versions for negotiation
+			// a role which provides just enough power to determine if the server is
+			// ready and discover API versions for negotiation
 			ObjectMeta: metav1.ObjectMeta{Name: "system:discovery"},
 			Rules: []rbacv1.PolicyRule{
 				rbacv1helpers.NewRule("get").URLs(
@@ -205,6 +206,20 @@ func ClusterRoles() []rbacv1.ClusterRole {
 					"/openapi", "/openapi/*",
 					"/api", "/api/*",
 					"/apis", "/apis/*",
+				).RuleOrDie(),
+			},
+		},
+		{
+			// a role which provides minimal read access to the monitoring endpoints
+			// (i.e. /metrics, /livez/*, /readyz/*, /healthz/*, /livez, /readyz, /healthz)
+			// The splatted health check endpoints allow read access to individual health check
+			// endpoints which may contain more sensitive cluster information information
+			ObjectMeta: metav1.ObjectMeta{Name: "system:monitoring"},
+			Rules: []rbacv1.PolicyRule{
+				rbacv1helpers.NewRule("get").URLs(
+					"/metrics",
+					"/livez", "/readyz", "/healthz",
+					"/livez/*", "/readyz/*", "/healthz/*",
 				).RuleOrDie(),
 			},
 		},
@@ -563,6 +578,7 @@ const systemNodeRoleName = "system:node"
 func ClusterRoleBindings() []rbacv1.ClusterRoleBinding {
 	rolebindings := []rbacv1.ClusterRoleBinding{
 		rbacv1helpers.NewClusterBinding("cluster-admin").Groups(user.SystemPrivilegedGroup).BindingOrDie(),
+		rbacv1helpers.NewClusterBinding("system:monitoring").Groups(user.MonitoringGroup).BindingOrDie(),
 		rbacv1helpers.NewClusterBinding("system:discovery").Groups(user.AllAuthenticated).BindingOrDie(),
 		rbacv1helpers.NewClusterBinding("system:basic-user").Groups(user.AllAuthenticated).BindingOrDie(),
 		rbacv1helpers.NewClusterBinding("system:public-info-viewer").Groups(user.AllAuthenticated, user.AllUnauthenticated).BindingOrDie(),
--- a/staging/src/k8s.io/apiserver/pkg/authentication/user/user.go
+++ b/staging/src/k8s.io/apiserver/pkg/authentication/user/user.go
@@ -70,6 +70,7 @@ func (i *DefaultInfo) GetExtra() map[string][]string {
 const (
 	SystemPrivilegedGroup = "system:masters"
 	NodesGroup            = "system:nodes"
+	MonitoringGroup       = "system:monitoring"
 	AllUnauthenticated    = "system:unauthenticated"
 	AllAuthenticated      = "system:authenticated"
 
--- a/cluster/gce/gci/configure-helper.sh
+++ b/cluster/gce/gci/configure-helper.sh
@@ -687,6 +687,9 @@ function create-master-auth {
     append_or_replace_prefixed_line "${known_tokens_csv}" "${KONNECTIVITY_SERVER_TOKEN},"     "system:konnectivity-server,uid:system:konnectivity-server"
     create-kubeconfig "konnectivity-server" ${KONNECTIVITY_SERVER_TOKEN}
   fi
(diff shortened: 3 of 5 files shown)


brancz:
Makes sense to me.

/lgtm

plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go
```diff
@@ -208,6 +209,17 @@ func ClusterRoles() []rbacv1.ClusterRole {
...
 			},
 		},
+		{
+			// a role which provides minimal read access to the instrumentation endpoints
+			// (i.e. /metrics, /debug/pprof/*")
+			ObjectMeta: metav1.ObjectMeta{Name: "system:monitoring"},
+			Rules: []rbacv1.PolicyRule{
+				rbacv1helpers.NewRule("get").URLs(
+					"/metrics",
+					"/debug/pprof/*",
```
lavalamp, line 219: hrm, I'm a little surprised to see /debug here. That seems more powerful than /metrics. Also is /debug useful if you can only send GET requests?

What about adding /healthz*, /livez*, /readyz*, componentstatuses (ugh), /version, and /logs?

logicalhan: ahahaha you mentioned, by name, the endpoint that must remain unmentioned by name. 

Yeah I can add the other ones but for the other one I will defer to @liggitt (whom I tried unsuccessfully to convince of this earlier)

liggitt: yeah, I really don't like a built-in role for this... I think people should define a role for their monitoring component that includes just the endpoints they want

staging/src/k8s.io/apiserver/pkg/authentication/user/user.go
```diff
@@ -70,8 +70,10 @@ func (i *DefaultInfo) GetExtra() map[string][]string {
 const (
 	SystemPrivilegedGroup = "system:masters"
 	NodesGroup            = "system:nodes"
-	AllUnauthenticated    = "system:unauthenticated"
-	AllAuthenticated      = "system:authenticated"
+	// TODO(logicalhan): configure flowcontrol for requests from this group
+	MonitoringGroup    = "system:monitoring"
```
liggitt, line 74: I'm -1 on a built-in monitoring group ... that is not well-defined across clusters like nodes are

plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go
logicalhan: Had out-of-band conversation with @liggitt, we've agreed to leave out pprof and the endpoint that must not be referred to by name but added the healthcheck splats for the error reporting.

(Also, we actually don't need version since it's already accessible by public-info-viewer to unauthenticated.)

lavalamp:
I think this is pretty reasonable now, I'll let Jordan take a final look though

staging/src/k8s.io/apiserver/pkg/authentication/user/user.go
```diff
@@ -70,8 +70,10 @@ func (i *DefaultInfo) GetExtra() map[string][]string {
 const (
 	SystemPrivilegedGroup = "system:masters"
 	NodesGroup            = "system:nodes"
-	AllUnauthenticated    = "system:unauthenticated"
-	AllAuthenticated      = "system:authenticated"
+	// TODO(logicalhan): configure flowcontrol for requests from this group
```
liggitt, line 73: can you describe what you envision for this? not sure this TODO belongs in a list of constants

plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go
```diff
@@ -208,6 +209,21 @@ func ClusterRoles() []rbacv1.ClusterRole {
...
+			// (i.e. /metrics, /livez/*, /readyz/*, /healthz/*)
+			// The splatted health check endpoints allow read access to individual health check
+			// endpoints which may contain more sensitive cluster information information
+			// (differentiating this from the access provided by system:public-info-viewer"
+			// which only provides top level health check access and omits detailed
+			// failure information)
+			ObjectMeta: metav1.ObjectMeta{Name: "system:monitoring"},
+			Rules: []rbacv1.PolicyRule{
+				rbacv1helpers.NewRule("get").URLs(
+					"/metrics", "/livez/*", "/readyz/*", "/healthz/*",
```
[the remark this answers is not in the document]
logicalhan: `"/livez*", "/readyz*", "/healthz*"` should be okay?

staging/src/k8s.io/apiserver/pkg/authentication/user/user.go
logicalhan: :P 

Yeah, I was thinking we probably don't actually want to give this role the same sort of priority as more critical requests. Philosophically, I believe that instrumentation should degrade gracefully without imploding the control-plane (i.e. I'd rather have my control-plane working without metrics than have my metrics take down my control-plane).

plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go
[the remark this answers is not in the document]
liggitt: don't glob the paths without a trailing / or you'll match hypothetically /healthzxy as well

staging/src/k8s.io/apiserver/pkg/authentication/user/user.go
brancz: There are a couple of monitoring/alerting philosophies at play here, I'm sure you can learn a lot more than from me internally at Google, but the gist is there are vaguely two opinions: when unable to monitor, you don't know the state of the system, so you assume the worst and read it as unavailability, or not. This is of course largely more about measuring, than the applications behavior, but I would keep this in mind. I tend to prefer the first, meaning monitoring should actually have equal if not higher priority than the rest (much like schedling priority itself), as otherwise we cannot actually reason about this system.

logicalhan: Yeah, and I don't necessarily disagree with you but this is specifically about flow control and prioritization of traffic to the apiserver. Saying metrics endpoint traffic is lower priority doesn't mean that metrics are low priority. It just means if we get DDOS'd on the metrics endpoint then we'll start queuing some of those requests and letting stuff like internal apiserver self-requests go through (and hopefully succeed).

liggitt: thanks for the context, can you copy this thread into an issue rather than a todo here?

logicalhan: I'm going to remove the TODO since someone has filed an issue around this already (https://github.com/kubernetes/kubernetes/issues/93359) and we are basically converging on reusing this group to fix that existing issue.

staging/src/k8s.io/apiserver/pkg/authentication/user/user.go
```diff
@@ -70,8 +70,10 @@ func (i *DefaultInfo) GetExtra() map[string][]string {
 const (
 	SystemPrivilegedGroup = "system:masters"
 	NodesGroup            = "system:nodes"
-	AllUnauthenticated    = "system:unauthenticated"
-	AllAuthenticated      = "system:authenticated"
+
```
liggitt, line 73: nit, drop newline

PR #93327, opened 2020-07-22
e2e dns test autodetect the IP family

aojea:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

instead of using tag to discriminate the e2e test, we can use
the e2e framework IPFamily method.

```release-note
NONE
```

### DIFF
--- a/test/e2e/network/dns_configmap.go
+++ b/test/e2e/network/dns_configmap.go
@@ -26,7 +26,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2eservice "k8s.io/kubernetes/test/e2e/framework/service"
-	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
 
 	"github.com/onsi/ginkgo"
 )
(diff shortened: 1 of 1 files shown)


cmluciano (commented):
This does look a lot better but doesn't removing the tag mean that I can't choose whether or not I want to run the v4/v6 tests individually? Perhaps this is intentional since these will eventually be promoted to conformance tests.

aojea:
> This does look a lot better but doesn't removing the tag mean that I can't choose whether or not I want to run the v4/v6 tests individually? Perhaps this is intentional since these will eventually be promoted to conformance tests.

You can still run them individually, tags are used to specify that a test only works with an IPFamily, if they don't have a tag they should work with whatever IPFamily the cluster has configured

aojea:
can we put this in milestone 1.19?
, it will fix some reds here

aojea:
> can we put this in milestone 1.19?
> , it will fix some reds here
> https://k8s-testgrid.appspot.com/sig-network-kind#sig-network-kind,%20IPv6,%20master

@hasheddan ^^^

hasheddan (approved):
/lgtm
/milestone v1.19

PR #93330, opened 2020-07-22
make route creation concurrency configurable

databus23:
This allows cloud provider implementations to modify this when setting up the controller.

Signed-off-by: Fabian Ruff <fabian.ruff@sap.com>

**What type of PR is this?**
/kind feature

It allows cloud provider implementations to configure how many concurrent requests to perform against the IAAS network/route api.
It was marked a TODO in the code for a long time.

```release-note
NONE
```

### DIFF
--- a/staging/src/k8s.io/cloud-provider/controllers/route/route_controller.go
+++ b/staging/src/k8s.io/cloud-provider/controllers/route/route_controller.go
@@ -45,9 +45,8 @@ import (
 )
 
 const (
-	// Maximal number of concurrent CreateRoute API calls.
-	// TODO: This should be per-provider.
-	maxConcurrentRouteCreations int = 200
+	// Default maximal number of concurrent CreateRoute API calls.
+	defaultMaxConcurrentRouteCreations int = 200
 )
 
 var updateNetworkConditionBackoff = wait.Backoff{
@@ -57,6 +56,8 @@ var updateNetworkConditionBackoff = wait.Backoff{
 }
 
 type RouteController struct {
+	MaxConcurrentRouteCreations int
+
 	routes           cloudprovider.Routes
 	kubeClient       clientset.Interface
 	clusterName      string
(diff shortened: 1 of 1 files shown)


staging/src/k8s.io/cloud-provider/controllers/route/route_controller.go
```diff
@@ -57,6 +56,8 @@ var updateNetworkConditionBackoff = wait.Backoff{
 }
 
 type RouteController struct {
+	MaxConcurrentRouteCreations int
```
andrewsykim, line 59: Modifying the field for RouteController is probably not the right approach here. 

Not 100% sure yet, but it _might_ be better to add new config type for route controller with flags, similar to service controller https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/service/config/types.go#L19-L25

andrewsykim: Alternatively we may want to add methods to `Routes` interface in the cloud provider to tweak this but there hasn't been much precedent in the past for adding rate limiting knobs there.

databus23: > Modifying the field for RouteController is probably not the right approach here as this struct is not easily accessible when vendoring the cloud provider lib.

It is? The `New` function returns a `*RouteController` I don't see how that wouldn't be accessible by cloud providers vendoring this lib.
Obviously I took the current approach because it was the easiest without changing any apis and maintaining backwards compatibility for a rather unimportant (in general not to me of course :)) knob.

That being said I'm open to add a new config type for the RouteController if that is the preferred way.

I also don't like changing the `Routes` interface for this.

andrewsykim: >It is? The New function returns a *RouteController. I don't see how that wouldn't be accessible by cloud providers vendoring this lib.

The type is accessible, but typically you don't import this controller directly, you import `cmd/cloud-controller-manager` which will call New. There's no existing hooks to modify the instance of RouteController returned by New. 

> That being said I'm open to add a new config type for the RouteController if that is the preferred way.

I think this might be the best approach.

PR #93331, opened 2020-07-22
Disable CFS CPU quota with guaranteed QoS pod when the CPU manager is enabled

### LINKED ISSUE #70585: Disable cpu quota(use only cpuset) for pod Guaranteed

Disable cpu quota when we use cpuset for guaranteed pods.

**Why is this needed**:

For now, kubelet default adds cpu quota and share for all cases when `cpuCFSQuota` is open, even if *cpuset* will be set in guaranteed pod case. 

However, this kind of use(cpuset with quota) may sometimes bring us cpu throttle time which we don't expect.

My question is why not make a `if` statement when we add cpuQuota config, if it's *guaranteed* case, just set `cpuQuota=-1`?

```
	if m.cpuCFSQuota {
		// if cpuLimit.Amount is nil, then the appropriate default value is returned
		// to allow full usage of cpu resource.
		cpuQuota, cpuPeriod := milliCPUToQuota(cpuLimit.MilliValue())
		lc.Resources.CpuQuota = cpuQuota
		lc.Resources.CpuPeriod = cpuPeriod
	}
``` 

/kind feature

coderwangke:
/kind bug

Disable CFS CPU quota with guaranteed QoS pod at pod and container level when the CPU manager is enabled. In the case of  the CpusetCpus, the setting of cfs quota will trigger the cpu throttle problem.

Fixes #70585 

```release-note
Disable CFS CPU quota with guaranteed QoS pod at pod and container level when the CPU manager is enabled.
```


coderwangke:
I signed it

2rs2ts (approved):
LGTM. I tried looking through the codebase to see if any stones may have been left unturned but couldn't find any. I also looked into the -1 value since that's a magic number to me and from what I can tell the moby/libcontainer API allows the value to be nil, but I couldn't find enough evidence one way or another to tell if that's the right number to set. I do see that this code was already setting it to -1 if quotas were disabled so assuming that works as expected then this should as well.

Have you considered adding tests for this?

cswindle:
Is there any update on this issue, we have also hit scenarios where we are being significantly throttled on our guaranteed QoS pods, where there is no reason to.

odinuge:
Hi!

I have looked through this PR, and I think this feature will require a [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/README.md) since the changes would end up quite big. The current implementation in this PR would not work properly, since there are a lot of aspects it doesn't account for. This PR would eg. disable CFS for all guaranteed pods, but exclusive CPU pinning only applies to pods with integer cpu request/limit.

A real solution would have to (at least) work together with the CPU Manager, and only disable cfs for containers/pods where a cpu set is assigned to a container/pod. This therefore has to be discussed further, and implemented in a way that doesn't break the current behavior of cgroup enforcement.

I do however see that this feature can be beneficial (even tho it shouldn't be necessary, since linux shouldn't throttle the processes), but I think we can continue the discussion in #70585.

/close

2rs2ts:
>I do however see that this feature can be beneficial (even tho it shouldn't be necessary, since linux shouldn't throttle the processes)

Linux definitely _does_ throttle the processes, FWIW. It's a massive problem at my company, and we've spent thousands of engineer-hours trying to mitigate it.

A KEP seems only borderline appropriate for this. It's very easy to see how this is a bugfix, but you are right that it would change a large chunk of behavior. I am curious why we couldn't just discuss the bugs in this PR and fix them.

odinuge:
Hi @2rs2ts! Thanks for showing interest in this PR! :smile: 

> Linux definitely does throttle the processes, FWIW. It's a massive problem at my company, and we've spent thousands of engineer-hours trying to mitigate it.

Yeah, I do have experienced it myself, and know it might be a problem. I think that this should be discussed with the issue and not here, in order to keep all the information/pros/cons in the same place. As multiple people say in the related issues, this is more of a bug in the kernel and not in k8s, but I agree that such feature in k8s would be helpful to mitigate it when a container/pod get a exclusive cpuset.

> A KEP seems only borderline appropriate for this. It's very easy to see how this is a bugfix, but you are right that it would change a large chunk of behavior. 

The reason I suggest using a KEP is to make the whole process faster, since it is easier for the whole sig to read the reasoning behind the change, and it makes it easier for other to give feedback about the selected approach. It is also way easier to get reviewer/approver bandwith by creating a good KEP compared to a just a PR that require a lot of discussion. There isn't a clear requirement to use a KEP, but I think it will be beneficial for everyone due to the nature of this issue.

Imo. this change should also contain a set of tests and a proper testplan to ensure we don't get a regression later, and more important that it doesn't change the current behavior of cgroup enforcement. In the current state, this PR would disable CFS for all guaranteed workloads, no matter if they have exclusive cpusets, and that _will_ create possible issues for users relying on CFS enforcement. That is a situation we would like to avoid. Maybe we even should create a featuregate to make it possible to opt out in case the implementation doesn't work as expected.

It would also be interesting to see how this would interact with the pod level resource limits/requests, so that would in theory be another edge case.

I hope this helps as more of a background. I can help creating a KEP if someone are interested (as I want this myself)!

> I am curious why we couldn't just discuss the bugs in this PR and fix them.

We definitely can, and I/we can reopen the PR if the author is interested in continuing to work on this. :smile: I am however not sure if I agree that this is a bug in Kubernetes, I personally think this would be more of a mitigation or a new feature of CPU Manager, and that the overall change would be quite complex.

cc author @coderwangke.

coderwangke:
> Hi @2rs2ts! Thanks for showing interest in this PR! 😄
> 
> > Linux definitely does throttle the processes, FWIW. It's a massive problem at my company, and we've spent thousands of engineer-hours trying to mitigate it.
> 
> Yeah, I do have experienced it myself, and know it might be a problem. I think that this should be discussed with the issue and not here, in order to keep all the information/pros/cons in the same place. As multiple people say in the related issues, this is more of a bug in the kernel and not in k8s, but I agree that such feature in k8s would be helpful to mitigate it when a container/pod get a exclusive cpuset.
> 
> > A KEP seems only borderline appropriate for this. It's very easy to see how this is a bugfix, but you are right that it would change a large chunk of behavior.
> 
> The reason I suggest using a KEP is to make the whole process faster, since it is easier for the whole sig to read the reasoning behind the change, and it makes it easier for other to give feedback about the selected approach. It is also way easier to get reviewer/approver bandwith by creating a good KEP compared to a just a PR that require a lot of discussion. There isn't a clear requirement to use a KEP, but I think it will be beneficial for everyone due to the nature of this issue.
> 
> Imo. this change should also contain a set of tests and a proper testplan to ensure we don't get a regression later, and more important that it doesn't change the current behavior of cgroup enforcement. In the current state, this PR would disable CFS for all guaranteed workloads, no matter if they have exclusive cpusets, and that _will_ create possible issues for users relying on CFS enforcement. That is a situation we would like to avoid. Maybe we even should create a featuregate to make it possible to opt out in case the implementation doesn't work as expected.
> 
> It would also be interesting to see how this would interact with the pod level resource limits/requests, so that would in theory be another edge case.
> 
> I hope this helps as more of a background. I can help creating a KEP if someone are interested (as I want this myself)!
> 
> > I am curious why we couldn't just discuss the bugs in this PR and fix them.
> 
> We definitely can, and I/we can reopen the PR if the author is interested in continuing to work on this. 😄 I am however not sure if I agree that this is a bug in Kubernetes, I personally think this would be more of a mitigation or a new feature of CPU Manager, and that the overall change would be quite complex.
> 
> cc author @coderwangke.

Sorry for the late reply. 
@2rs2ts I will add tests for this PR.

2rs2ts:
Thanks for reopening. FWIW I do not think that this is a bug in the kernel, it's working as expected and designed and I don't even think it's really a design flaw. I strongly believe that it's just a simple matter of the implementation of CPU Manager being incomplete, and I thought that was the consensus in the linked issue, not that it was a Linux bug.

I agree that we need tests and also we shouldn't disable the quota for all Guaranteed pods, it's gotta be only the pods that are being given exclusive CPUs when CPU Manager is enabled. I will look over this PR again when these are addressed. I'm eager to help make this change happen!

coderwangke:
> Thanks for reopening. FWIW I do not think that this is a bug in the kernel, it's working as expected and designed and I don't even think it's really a design flaw. I strongly believe that it's just a simple matter of the implementation of CPU Manager being incomplete, and I thought that was the consensus in the linked issue, not that it was a Linux bug.
> 
> I agree that we need tests and also we shouldn't disable the quota for all Guaranteed pods, it's gotta be only the pods that are being given exclusive CPUs when CPU Manager is enabled. I will look over this PR again when these are addressed. I'm eager to help make this change happen!

coderwangke:
> Thanks for reopening. FWIW I do not think that this is a bug in the kernel, it's working as expected and designed and I don't even think it's really a design flaw. I strongly believe that it's just a simple matter of the implementation of CPU Manager being incomplete, and I thought that was the consensus in the linked issue, not that it was a Linux bug.
> 
> I agree that we need tests and also we shouldn't disable the quota for all Guaranteed pods, it's gotta be only the pods that are being given exclusive CPUs when CPU Manager is enabled. I will look over this PR again when these are addressed. I'm eager to help make this change happen!

I looked at the PR again, and it did disable the quota for all Guaranteed pods.
I will see how to fix this.

PR #93354, opened 2020-07-22
Increase the size of the buckets for PodSchedulingDuration scheduler metric.

ahg-g:
**What type of PR is this?**

/kind feature

PodSchedulingDuration includes the time the pod is sitting in the queue and all scheduling attempts, this duration is expected to
be from seconds to minutes rather than milliseconds to seconds.

-->
```release-note
NONE
```

/priority important-soon
/milestone v1.19

### DIFF
--- a/pkg/scheduler/metrics/metrics.go
+++ b/pkg/scheduler/metrics/metrics.go
@@ -159,8 +159,8 @@ var (
 			Subsystem: SchedulerSubsystem,
 			Name:      "pod_scheduling_duration_seconds",
 			Help:      "E2e latency for a pod being scheduled which may include multiple scheduling attempts.",
-			// Start with 1ms with the last bucket being [~16s, Inf)
-			Buckets:        metrics.ExponentialBuckets(0.001, 2, 15),
+			// Start with 10ms with the last bucket being [~88m, Inf).
+			Buckets:        metrics.ExponentialBuckets(0.01, 2, 20),
 			StabilityLevel: metrics.ALPHA,
 		},
 		[]string{"attempts"})


pkg/scheduler/metrics/metrics.go
```diff
@@ -159,8 +159,8 @@ var (
 			Subsystem: SchedulerSubsystem,
 			Name:      "pod_scheduling_duration_seconds",
 			Help:      "E2e latency for a pod being scheduled which may include multiple scheduling attempts.",
-			// Start with 1ms with the last bucket being [~16s, Inf)
-			Buckets:        metrics.ExponentialBuckets(0.001, 2, 15),
+			// Start with 1.875s with the last bucket being [64m, Inf)
+			Buckets:        metrics.ExponentialBuckets(1.875, 2, 12),
```
Huang-Wei, line 163: Do we have some DP proving `1.875s` is a good starting point?

ahg-g: I think the current one is certainly too small and not useful.

I expect that pods will take in the range of few seconds if scheduled on the first attempt, and that will depend on the queue length. If scheduling the pod will require multiple attempts, then I expect this to be in the tens of seconds to minutes.

ahg-g: we could start with something smaller, but I wanted to align with minutes at the end.

ahg-g: how about 0.1 second (100ms)?

Huang-Wei: SG.

Huang-Wei:
/lgtm
/hold
for @logicalhan to take a look as I'm not quite sure it's conventional to start with a non-round number (0.1171875).

ahg-g:
The other option is to start with .01s (100ms, nice start number) and have 20 buckets: 

0.01s
0.02s
0.04s
0.08s
0.16s
0.32s
0.64s
1.28s
2.56s
5.12s
10.24s
20.48s
40.96s
1.365333333m
2.730666667m
5.461333333m
10.92266667m
21.84533333m
43.69066667m
87.38133333m

logicalhan (approved):
/lgtm
/approve

liggitt:
relevant failure?

> [measurement call SchedulingThroughput - SchedulingThroughput error: scheduler throughput: actual throughput 75.800000 lower than threshold 90.000000]

ahg-g:
> relevant failure?
> 
> > [measurement call SchedulingThroughput - SchedulingThroughput error: scheduler throughput: actual throughput 75.800000 lower than threshold 90.000000]

no, this metric isn't used for any tests.

PR #93368, opened 2020-07-22
Avoid dereferencing endpoint twice on the deletion of a service

### LINKED ISSUE #93229: Windows Kubeproxy has Connectivity issues when one of the multiple services referencing a same endpoint is deleted or updated
**What happened**:
Two services referring the same endpoint. One of the services was deleted. This brought down the connectivity of the other service.

The connectivity to service should continue to work

1. Create two services referencing same endpoint
2. Delete one of the services
3. The connectivity is broken to the other service

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sbangari:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

When a service is deleted, the corresponding remote endpoints are dereferenced. But the map holding the list of remote endpoints created for this Service is not updated. This causes the remote endpoints to become stale and they are attempted to be cleaned up a second time later causing a second dereference.

In a scenario where the remote endpoints are shared across two services, the second dereference causes the remote endpoint to be deleted from HNS.

Fixes #93229

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/proxy/winkernel/proxier.go
+++ b/pkg/proxy/winkernel/proxier.go
@@ -319,19 +319,21 @@ func (ep *endpointsInfo) Cleanup() {
 	Log(ep, "Endpoint Cleanup", 3)
 	if ep.refCount != nil {
 		*ep.refCount--
-	}
 
-	// Remove the remote hns endpoint, if no service is referring it
-	// Never delete a Local Endpoint. Local Endpoints are already created by other entities.
-	// Remove only remote endpoints created by this service
-	if (ep.refCount == nil || *ep.refCount <= 0) && !ep.GetIsLocal() {
-		klog.V(4).Infof("Removing endpoints for %v, since no one is referencing it", ep)
-		err := ep.hns.deleteEndpoint(ep.hnsID)
-		if err == nil {
-			ep.hnsID = ""
-		} else {
-			klog.Errorf("Endpoint deletion failed for %v: %v", ep.IP(), err)
+		// Remove the remote hns endpoint, if no service is referring it
+		// Never delete a Local Endpoint. Local Endpoints are already created by other entities.
+		// Remove only remote endpoints created by this service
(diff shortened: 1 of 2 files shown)


elweb9858 (approved):
lgtm. Have you thought about writing a test for this scenario? Maybe creating two services referencing the same remote endpoints, deleting one of the services, and verifying the endpoints still exist.

elweb9858:
also are you aiming to get this bug fix into v1.19? fyi @daschott

pkg/proxy/winkernel/proxier.go
```diff
@@ -279,6 +279,9 @@ func (proxier *Proxier) onServiceMapChange(svcPortName *proxy.ServicePortName) {
 
 		klog.V(3).Infof("Updating existing service port %q at %s:%d/%s", svcPortName, svcInfo.ClusterIP(), svcInfo.Port(), svcInfo.Protocol())
 		svcInfo.cleanupAllPolicies(proxier.endpointsMap[*svcPortName])
+		// Delete the endpoint map entry for this service, since the
+		// endpoints are already cleanedup
+		delete(proxier.endpointsMap, *svcPortName)
```
kumarvin123, line 284: This is a  problem. The endpoints might be still there and the Svc might have been updated. We clean up all policies assuming they might be stale. This code is unconditionally deleting all endpoints from the map.

sbangari: I opened a new PR for this issue. #93368

sbangari:
> lgtm. Have you thought about writing a test for this scenario? Maybe creating two services referencing the same remote endpoints, deleting one of the services, and verifying the endpoints still exist.

@elweb9858   I opened a new PR for this issue and which also has new tests added for the failing scenario. #93368

Issue #93217, opened 2020-07-19
[flaky] SchedulingThroughput - SchedulingThroughput error: scheduler throughput: actual throughput 81.200000 lower than threshold 90.000000]


liggitt:
**Which jobs are flaking**:

* ci-kubernetes-e2e-gci-gce-scalability
* ci-kubernetes-e2e-gci-gce-scalability-networkpolicies
* pull-kubernetes-e2e-gce-100-performance

**Which test(s) are flaking**:

```
testing/density/config.yaml

SchedulingThroughput error: scheduler throughput: actual throughput 81.800000 lower than threshold 90.000000]
```

**Testgrid link**:

* https://testgrid.k8s.io/google-gce#gce-cos-master-scalability-100&width=5
* https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-e2e-gce-100-performance&width=5

**Anything else we need to know**:

Looking just at periodic jobs (not presubmit), the number of failures has been steadily rising since May:
* 3: https://storage.googleapis.com/k8s-gubernator/triage/index.html?date=2020-05-30&text=scheduler%20throughput
* 6: https://storage.googleapis.com/k8s-gubernator/triage/index.html?date=2020-06-06&text=scheduler%20throughput
* 11: https://storage.googleapis.com/k8s-gubernator/triage/index.html?date=2020-06-13&text=scheduler%20throughput
* 16: https://storage.googleapis.com/k8s-gubernator/triage/index.html?date=2020-06-20&text=scheduler%20throughput
* 19: https://storage.googleapis.com/k8s-gubernator/triage/index.html?date=2020-06-27&text=scheduler%20throughput
* 26: https://storage.googleapis.com/k8s-gubernator/triage/index.html?date=2020-07-04&text=scheduler%20throughput
* 34: https://storage.googleapis.com/k8s-gubernator/triage/index.html?date=2020-07-11&text=scheduler%20throughput
* 65: https://storage.googleapis.com/k8s-gubernator/triage/index.html?date=2020-07-18&text=scheduler%20throughput

/sig scalability
/sig scheduling

liggitt:
moving into the milestone for investigation

@kubernetes/sig-scalability-bugs @kubernetes/sig-scheduling-bugs

justaugustus:
FYI @kubernetes/ci-signal

ahg-g:
looking at few "failed" runs, the binding latency is the main culprit.

Example failed runs:
[...]

PR #93371, opened 2020-07-23
Optimize VolumeRestriction plugin

ahg-g:
**What type of PR is this?**

 /kind feature

Avoids redundant iterations in the VolumeRestrictions scheduler plugin. This may have positive impact for workloads that use secrets/configmaps etc.

Part of #93217

```release-note
NONE
```

/assign @wojtek-t

### DIFF
--- a/pkg/scheduler/framework/plugins/volumerestrictions/volume_restrictions.go
+++ b/pkg/scheduler/framework/plugins/volumerestrictions/volume_restrictions.go
@@ -42,12 +42,7 @@ func (pl *VolumeRestrictions) Name() string {
 	return Name
 }
 
-func isVolumeConflict(volume v1.Volume, pod *v1.Pod) bool {
-	// fast path if there is no conflict checking targets.
-	if volume.GCEPersistentDisk == nil && volume.AWSElasticBlockStore == nil && volume.RBD == nil && volume.ISCSI == nil {
-		return false
-	}
-
+func isVolumeConflict(volume *v1.Volume, pod *v1.Pod) bool {
 	for _, existingVolume := range pod.Spec.Volumes {
 		// Same GCE disk mounted by multiple pods conflicts unless all pods mount it read-only.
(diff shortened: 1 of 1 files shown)


ahg-g:
unlikely to cause too much difference, but worth doing it.

pkg/scheduler/framework/plugins/volumerestrictions/volume_restrictions.go
```diff
@@ -118,7 +113,13 @@ func haveOverlap(a1, a2 []string) bool {
 // - Ceph RBD forbids if any two pods share at least same monitor, and match pool and image, and the image is read-only
 // - ISCSI forbids if any two pods share at least same IQN and ISCSI volume is read-only
 func (pl *VolumeRestrictions) Filter(ctx context.Context, _ *framework.CycleState, pod *v1.Pod, nodeInfo *framework.NodeInfo) *framework.Status {
-	for _, v := range pod.Spec.Volumes {
+	for i := range pod.Spec.Volumes {
```
Huang-Wei, line 116: It's no harm to keep the original `for _, v := range pod.Spec.Volumes` logic, no?

Huang-Wei (commented):
One question. LGTM otherwise.

pkg/scheduler/framework/plugins/volumerestrictions/volume_restrictions.go
ahg-g: why should we copy the object twice?

Huang-Wei: True, but I'd argue the cost (the size of `Volume` is trivial) is trivial.

/lgtm
anyway

wojtek-t:
LGTM - thanks!

hasheddan (approved):
/lgtm
/kind feature

PR #93382, opened 2020-07-23
Fix annotation sorting in kuberuntime convert test

mfojtik:
**What type of PR is this?**

/kind flake

The source image annotation is based on map[string]string which is not sorted in Go. The resulting slice of annotations can be different between test runs and produce flakes, see: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_kubernetes/166/pull-ci-openshift-kubernetes-master-unit/1286171049615953920

```
=== RUN   TestConvertToKubeContainerImageSpec
--- FAIL: TestConvertToKubeContainerImageSpec (0.00s)
    convert_test.go:95: 
        	Error Trace:	convert_test.go:95
        	Error:      	Not equal: 
        	            	expected: container.ImageSpec{Image:"test", Annotations:[]container.Annotation{container.Annotation{Name:"kubernetes.io/os", Value:"linux"}, container.Annotation{Name:"kubernetes.io/runtimehandler", Value:"handler"}}}
        	            	actual  : container.ImageSpec{Image:"test", Annotations:[]container.Annotation{container.Annotation{Name:"kubernetes.io/runtimehandler", Value:"handler"}, container.Annotation{Name:"kubernetes.io/os", Value:"linux"}}}
```
```release-note
NONE
```

### DIFF
--- a/pkg/kubelet/kuberuntime/convert_test.go
+++ b/pkg/kubelet/kuberuntime/convert_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package kuberuntime
 
 import (
+	"sort"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


mattjmcnaughton (approved):
/retest
/lgtm

liggitt:
ha, I found the same thing independently. I fixed it in the source rather than in the test so the output was determinstic in https://github.com/kubernetes/kubernetes/pull/93610

derekwaynecarr:
i see no issue in having the fix in both test and producer.

/approve

mfojtik:
/close

@liggitt thanks !

PR #93384, opened 2020-07-23
Support kubectl delete orphan/foreground/background options

### LINKED ISSUE #93364: Allow foreground cascading deletion in kubectl
Currently, `kubectl delete` only allows orphaning (`--cascade=false`) or background cascading (`--cascade=true`, the default). There is no option I know to allow foreground cascading deletion. I propose that we maybe do the same thing we did for `--dry-run`, that is, move from a boolean to an enum by accepting `foreground|background|orphan` while keeping compatibility with existing boolean forms.

I know this is a duplicate of another bug, I'll look for it.

/sig cli
/cc @julianvmodesto

zhouya0:
**What type of PR is this?**
/kind feature

Support kubectl delete orphan/foreground/background options.

Fixes https://github.com/kubernetes/kubernetes/issues/93364

```release-note
Support kubectl delete orphan/foreground/background options
```

### DIFF
--- a/staging/src/k8s.io/kubectl/pkg/cmd/apply/patcher.go
+++ b/staging/src/k8s.io/kubectl/pkg/cmd/apply/patcher.go
@@ -25,6 +25,7 @@ import (
 	"github.com/jonboulle/clockwork"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -57,10 +58,10 @@ type Patcher struct {
 	Overwrite bool
 	BackOff   clockwork.Clock
 
-	Force       bool
-	Cascade     bool
-	Timeout     time.Duration
-	GracePeriod int
+	Force             bool
+	CascadingStrategy metav1.DeletionPropagation
+	Timeout           time.Duration
+	GracePeriod       int
 
 	// If set, forces the patch against a specific resourceVersion
 	ResourceVersion *string
@@ -78,21 +79,21 @@ func newPatcher(o *ApplyOptions, info *resource.Info, helper *resource.Helper) (
 	}
 
 	return &Patcher{
-		Mapping:       info.Mapping,
-		Helper:        helper,
-		Overwrite:     o.Overwrite,
-		BackOff:       clockwork.NewRealClock(),
-		Force:         o.DeleteOptions.ForceDeletion,
-		Cascade:       o.DeleteOptions.Cascade,
-		Timeout:       o.DeleteOptions.Timeout,
-		GracePeriod:   o.DeleteOptions.GracePeriod,
-		OpenapiSchema: openapiSchema,
-		Retries:       maxPatchRetry,
+		Mapping:           info.Mapping,
+		Helper:            helper,
+		Overwrite:         o.Overwrite,
+		BackOff:           clockwork.NewRealClock(),
+		Force:             o.DeleteOptions.ForceDeletion,
+		CascadingStrategy: o.DeleteOptions.CascadingStrategy,
+		Timeout:           o.DeleteOptions.Timeout,
+		GracePeriod:       o.DeleteOptions.GracePeriod,
+		OpenapiSchema:     openapiSchema,
+		Retries:           maxPatchRetry,
 	}, nil
 }
 
 func (p *Patcher) delete(namespace, name string) error {
-	options := asDeleteOptions(p.Cascade, p.GracePeriod)
+	options := asDeleteOptions(p.CascadingStrategy, p.GracePeriod)
 	_, err := p.Helper.DeleteWithOptions(namespace, name, &options)
 	return err
 }
--- a/staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
+++ b/staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
@@ -17,12 +17,16 @@ limitations under the License.
 package delete
 
 import (
+	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/spf13/cobra"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/klog/v2"
 )
 
 // DeleteFlags composes common printer flag structs
@@ -32,20 +36,20 @@ type DeleteFlags struct {
 	LabelSelector *string
 	FieldSelector *string
 
-	All            *bool
-	AllNamespaces  *bool
-	Cascade        *bool
-	Force          *bool
-	GracePeriod    *int
-	IgnoreNotFound *bool
-	Now            *bool
-	Timeout        *time.Duration
-	Wait           *bool
-	Output         *string
-	Raw            *string
+	All               *bool
+	AllNamespaces     *bool
+	CascadingStrategy *string
+	Force             *bool
+	GracePeriod       *int
+	IgnoreNotFound    *bool
+	Now               *bool
+	Timeout           *time.Duration
+	Wait              *bool
+	Output            *string
+	Raw               *string
 }
 
-func (f *DeleteFlags) ToOptions(dynamicClient dynamic.Interface, streams genericclioptions.IOStreams) *DeleteOptions {
+func (f *DeleteFlags) ToOptions(dynamicClient dynamic.Interface, streams genericclioptions.IOStreams) (*DeleteOptions, error) {
 	options := &DeleteOptions{
 		DynamicClient: dynamicClient,
 		IOStreams:     streams,
@@ -73,8 +77,12 @@ func (f *DeleteFlags) ToOptions(dynamicClient dynamic.Interface, streams generic
 	if f.AllNamespaces != nil {
 		options.DeleteAllNamespaces = *f.AllNamespaces
 	}
-	if f.Cascade != nil {
-		options.Cascade = *f.Cascade
+	if f.CascadingStrategy != nil {
+		var err error
+		options.CascadingStrategy, err = getCascadingStrategy(*f.CascadingStrategy)
+		if err != nil {
+			return nil, err
+		}
 	}
 	if f.Force != nil {
 		options.ForceDeletion = *f.Force
@@ -98,7 +106,7 @@ func (f *DeleteFlags) ToOptions(dynamicClient dynamic.Interface, streams generic
 		options.Raw = *f.Raw
 	}
 
-	return options
+	return options, nil
 }
 
 func (f *DeleteFlags) AddFlags(cmd *cobra.Command) {
@@ -118,8 +126,13 @@ func (f *DeleteFlags) AddFlags(cmd *cobra.Command) {
 	if f.Force != nil {
 		cmd.Flags().BoolVar(f.Force, "force", *f.Force, "If true, immediately remove resources from API and bypass graceful deletion. Note that immediate deletion of some resources may result in inconsistency or data loss and requires confirmation.")
 	}
-	if f.Cascade != nil {
-		cmd.Flags().BoolVar(f.Cascade, "cascade", *f.Cascade, "If true, run background cascade deletion of the resources managed by this resource (e.g. Pods created by a ReplicationController).  Default true.")
+	if f.CascadingStrategy != nil {
+		cmd.Flags().StringVar(
+			f.CascadingStrategy,
+			"cascade",
+			*f.CascadingStrategy,
+			`Must be "background", "orphan", or "foreground". Selects the deletion cascading strategy for the dependents (e.g. Pods created by a ReplicationController). Defaults to background.`)
+		cmd.Flags().Lookup("cascade").NoOptDefVal = "background"
 	}
 	if f.Now != nil {
 		cmd.Flags().BoolVar(f.Now, "now", *f.Now, "If true, resources are signaled for immediate shutdown (same as --grace-period=1).")
@@ -146,7 +159,7 @@ func (f *DeleteFlags) AddFlags(cmd *cobra.Command) {
 
 // NewDeleteCommandFlags provides default flags and values for use with the "delete" command
 func NewDeleteCommandFlags(usage string) *DeleteFlags {
-	cascade := true
+	cascadingStrategy := "background"
 	gracePeriod := -1
 
 	// setup command defaults
@@ -172,8 +185,8 @@ func NewDeleteCommandFlags(usage string) *DeleteFlags {
 		LabelSelector: &labelSelector,
 		FieldSelector: &fieldSelector,
 
-		Cascade:     &cascade,
-		GracePeriod: &gracePeriod,
+		CascadingStrategy: &cascadingStrategy,
+		GracePeriod:       &gracePeriod,
 
 		All:            &all,
 		AllNamespaces:  &allNamespaces,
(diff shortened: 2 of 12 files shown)


staging/src/k8s.io/kubectl/pkg/cmd/apply/patcher.go
```diff
@@ -57,10 +57,10 @@ type Patcher struct {
 	Overwrite bool
 	BackOff   clockwork.Clock
 
-	Force       bool
-	Cascade     bool
-	Timeout     time.Duration
-	GracePeriod int
+	Force          bool
+	DeleteStratedy cmdutil.DeleteStrategy
```
apelisse, line 61: What about `CascadingStrategy`? Also, you have a typo here that you "propagated" all over the place

staging/src/k8s.io/kubectl/pkg/cmd/util/helpers.go
```diff
@@ -587,6 +596,48 @@ func PrintFlagsWithDryRunStrategy(printFlags *genericclioptions.PrintFlags, dryR
 	return printFlags
 }
 
+type DeleteStrategy int
```
apelisse, line 599: I would just drop this type and use the existing `metav1.DeletePropagation` type since this is a one-to-one mapping.

staging/src/k8s.io/kubectl/pkg/cmd/util/helpers.go
```diff
@@ -587,6 +596,48 @@ func PrintFlagsWithDryRunStrategy(printFlags *genericclioptions.PrintFlags, dryR
...
+			return Foreground, nil
+		case "background":
+			return Background, nil
+		default:
+			return Background, fmt.Errorf(`Invalid cascade value (%v). Must be "background", "orphan", or "foreground".`, deleteFlag)
+		}
+	}
+	// The flag was a boolean
+	if b {
+		klog.Warningf(`--cascade=%v is deprecated (boolean value) and can be replaced with --cascade=%s.`, deleteFlag, "background")
```
apelisse, line 634: OK I think it makes sense to deprecate the former boolean usage of the flag.

apelisse (commented):
Beside the extra type that I think is not needed and the variable name, I think that's great. But I haven't looked in enough detail yet, I'd really like @julianvmodesto to confirm that this is right.

Good job @zhouya0!

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
```diff
@@ -34,7 +34,6 @@ type DeleteFlags struct {
 
 	All            *bool
 	AllNamespaces  *bool
-	Cascade        *bool
```
julianvmodesto, line 37: I think we want a metav1.DeletePropagation here?

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
```diff
@@ -73,9 +72,6 @@ func (f *DeleteFlags) ToOptions(dynamicClient dynamic.Interface, streams generic
 	if f.AllNamespaces != nil {
 		options.DeleteAllNamespaces = *f.AllNamespaces
 	}
-	if f.Cascade != nil {
-		options.Cascade = *f.Cascade
```
julianvmodesto, line 77: Since other commands use DeleteFlags.ToOptions, is it possible to parse the propagation option here?

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
```diff
@@ -118,9 +114,6 @@ func (f *DeleteFlags) AddFlags(cmd *cobra.Command) {
 	if f.Force != nil {
 		cmd.Flags().BoolVar(f.Force, "force", *f.Force, "If true, immediately remove resources from API and bypass graceful deletion. Note that immediate deletion of some resources may result in inconsistency or data loss and requires confirmation.")
 	}
-	if f.Cascade != nil {
-		cmd.Flags().BoolVar(f.Cascade, "cascade", *f.Cascade, "If true, cascade the deletion of the resources managed by this resource (e.g. Pods created by a ReplicationController).  Default true.")
-	}
```
julianvmodesto, line 123: Similar here, do we want to add the delete propagation option flag here?

julianvmodesto (commented):
Nice, yes the flag Add + Get look good to me for accepting a string instead of a bool now.

I agree that using the metav1.DeletePropagation type would be good, and it would simplify the asDeleteOptions methods.

I think we want to continue to keep behavior for any other commands that use DeleteFlags, so the Add + Get methods should be there? Added a couple of comments where we removed DeleteFlags.Cascade.

And if we're deprecating the bool, then we can update test/cmd/ now too.

staging/src/k8s.io/kubectl/pkg/cmd/apply/patcher.go
zhouya0: Done :) And thanks for pointing out my typo.

staging/src/k8s.io/kubectl/pkg/cmd/util/helpers.go
zhouya0: Agree, done.

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
zhouya0: I moved `metav1,DeletePropagation` to DeleteOptions. Here we use `string` to get value from Cobra command :)

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
zhouya0: Done .

zhouya0:
/retest
@apelisse @julianvmodesto, really thanks for your review.  Changes are done according your advices.
Please review it again. :)

apelisse (commented):
I think it's good. The main point I think is that we don't have to change the behavior of `--cascade`.

The difference with dry-run is that we are trying to CHANGE the default for dry-run from client-side to server-side and we want the empty use-case to fail for a while.

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
```diff
@@ -32,20 +36,21 @@ type DeleteFlags struct {
...
+	Now               *bool
+	Timeout           *time.Duration
+	Wait              *bool
+	Output            *string
+	Raw               *string
 }
 
-func (f *DeleteFlags) ToOptions(dynamicClient dynamic.Interface, streams genericclioptions.IOStreams) *DeleteOptions {
+func (f *DeleteFlags) ToOptions(dynamicClient dynamic.Interface, streams genericclioptions.IOStreams) (*DeleteOptions, error) {
+	var err error
```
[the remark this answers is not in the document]
zhouya0: Done :)

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
```diff
@@ -118,8 +126,13 @@ func (f *DeleteFlags) AddFlags(cmd *cobra.Command) {
...
 		cmd.Flags().BoolVar(f.Force, "force", *f.Force, "If true, immediately remove resources from API and bypass graceful deletion. Note that immediate deletion of some resources may result in inconsistency or data loss and requires confirmation.")
 	}
-	if f.Cascade != nil {
-		cmd.Flags().BoolVar(f.Cascade, "cascade", *f.Cascade, "If true, cascade the deletion of the resources managed by this resource (e.g. Pods created by a ReplicationController).  Default true.")
+	if f.CascadingStrategy != nil {
+		cmd.Flags().StringVar(
+			f.CascadingStrategy,
+			"cascade",
+			*f.CascadingStrategy,
+			`Must be "background", "orphan", or "foreground". Determines to delete dependents in the background, foreground or orphan the dependents (e.g. Pods created by a ReplicationController). Default background `)
```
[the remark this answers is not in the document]
zhouya0: I'll wait for @julianvmodesto's response :)

zhouya0:
@apelisse  I've corrected the error scope. And the problem of `--cascade`, I'm not sure if I got your point correctly,
Should I remove the warning message when user only specified `--cascade`?

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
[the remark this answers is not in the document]
julianvmodesto: Right, no we don't document the boolean because we want people to stop using it.

apelisse:
> Should I remove the warning message when user only specified `--cascade`?

Not only the warning, but yes, I would keep the default value for `--cascade` (without the = form) to default to background the same way it does today. In other word, the existing behavior should stay exactly the same.

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
```diff
@@ -203,12 +216,39 @@ func NewDeleteFlags(usage string) *DeleteFlags {
...
 		Wait:    &wait,
 	}
 }
+
+func getCascadingStrategy(cascadingFlag string) (metav1.DeletionPropagation, error) {
+	b, err := strconv.ParseBool(cascadingFlag)
+	// The flag is not a boolean
+	if err != nil {
+		switch cascadingFlag {
+		case "unchanged":
```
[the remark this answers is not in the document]
zhouya0: Done :)

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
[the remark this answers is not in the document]
zhouya0: Great, so I guess the message here is suitable right?

apelisse (commented):
We're almost there, maybe keep a test where cascade is used in its "simple form": `--cascade`, and also if you could think of a way to make a foreground test that'd be lovely!

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
```diff
@@ -118,8 +126,13 @@ func (f *DeleteFlags) AddFlags(cmd *cobra.Command) {
...
 	}
-	if f.Cascade != nil {
-		cmd.Flags().BoolVar(f.Cascade, "cascade", *f.Cascade, "If true, cascade the deletion of the resources managed by this resource (e.g. Pods created by a ReplicationController).  Default true.")
+	if f.CascadingStrategy != nil {
+		cmd.Flags().StringVar(
+			f.CascadingStrategy,
+			"cascade",
+			*f.CascadingStrategy,
+			`Must be "background", "orphan", or "foreground". Determines to delete dependents in the background, foreground or orphan the dependents (e.g. Pods created by a ReplicationController). Default background `)
+		cmd.Flags().Lookup("cascade").NoOptDefVal = "background"
```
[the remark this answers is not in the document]
zhouya0: Without this, it would complain like `Error: flag needs an argument`

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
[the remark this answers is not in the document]
zhouya0: Done :)

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_test.go
```diff
@@ -112,8 +112,8 @@ func hasExpectedPropagationPolicy(body io.ReadCloser, policy *metav1.DeletionPro
 	return *policy == *parsedBody.PropagationPolicy
 }
 
-// Tests that DeleteOptions.OrphanDependents is appropriately set while deleting objects.
-func TestOrphanDependentsInDeleteObject(t *testing.T) {
+// TestCascadingStrategy tests that DeleteOptions.DeletionPropagation is appropriately set while deleting objects.
```
zhouya0, line 115: For review: `foreground` cascading strategy is done here.

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
```diff
@@ -118,8 +126,13 @@ func (f *DeleteFlags) AddFlags(cmd *cobra.Command) {
...
 		cmd.Flags().BoolVar(f.Force, "force", *f.Force, "If true, immediately remove resources from API and bypass graceful deletion. Note that immediate deletion of some resources may result in inconsistency or data loss and requires confirmation.")
 	}
-	if f.Cascade != nil {
-		cmd.Flags().BoolVar(f.Cascade, "cascade", *f.Cascade, "If true, cascade the deletion of the resources managed by this resource (e.g. Pods created by a ReplicationController).  Default true.")
+	if f.CascadingStrategy != nil {
+		cmd.Flags().StringVar(
+			f.CascadingStrategy,
+			"cascade",
+			*f.CascadingStrategy,
+			`Must be "background", "orphan", or "foreground". Selects to deletion cascading strategy for the dependents (e.g. Pods created by a ReplicationController). Defaults to background `)
```
apelisse, line 134: "Select the deletion ..." and you still have an extra space at the end.

zhouya0: Sorry... Done :) Next time I'll be more careful about spelling!

apelisse:
We're close to being done and should be able to merge when the queue re-opens

zhouya0:
Rebase done :)

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
```diff
@@ -203,12 +216,36 @@ func NewDeleteFlags(usage string) *DeleteFlags {
...
+		GracePeriod:       &gracePeriod,
 
 		// add non-defaults
 		Force:   &force,
 		Timeout: &timeout,
 		Wait:    &wait,
 	}
 }
+
+func getCascadingStrategy(cascadingFlag string) (metav1.DeletionPropagation, error) {
```
apelisse, line 229: nit, you don't have to do it here, but `getCascadingStrategy` is not the exact intent, this is more like `parseCascadingFlag`.

apelisse (commented):
/approve
/lgtm

Thanks!

apelisse:
/hold

@kubernetes/sig-cli-maintainers Do we need a formal approval/kep/discussion to change that?

apelisse:
According to @zhouya0, this has been approved by @kubernetes/sig-scalability-leads during sig-cli meeting.

/unhold

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
```diff
@@ -203,12 +216,36 @@ func NewDeleteFlags(usage string) *DeleteFlags {
...
+	b, err := strconv.ParseBool(cascadingFlag)
```
soltysh, line 230: nit: use longer variable name

staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go
```diff
@@ -203,12 +216,36 @@ func NewDeleteFlags(usage string) *DeleteFlags {
...
+		default:
+			return metav1.DeletePropagationBackground, fmt.Errorf(`Invalid cascade value (%v). Must be "background", "foreground", or "orphan".`, cascadingFlag)
+		}
+	}
+	// The flag was a boolean
+	if b {
+		klog.Warningf(`--cascade=%v is deprecated (boolean value) and can be replaced with --cascade=%s.`, cascadingFlag, "background")
+		return metav1.DeletePropagationBackground, nil
+	}
+	klog.Warningf(`--cascade=%v is deprecated (boolean value) and can be replaced with --cascade=%s.`, cascadingFlag, "orphan")
```
soltysh, line 249: @zhouya0 followup nit: don't use `klog`, but rather print directly to stderr

PR #53185, opened 2017-09-28
populate object name for admission attributes when CREATE

dixudx:
**What this PR does / why we need it**:
The admission attributes given to admission plugins do not populate object name on create.

Since the object is available, the name in the admission attributes should be populated if available.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #46107

**Special notes for your reviewer**:

**Release note**:

```release-note
populate object name for admission attributes when CREATE
```

PR #93389, opened 2020-07-23
Return err directly when nodename in node object is not consistent with in cert

fisherxu:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

When nodename in node ob is not consistent with in cert, it's no need to do other calculation.

Fixes #

/cc @liggitt 

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/plugin/pkg/admission/noderestriction/admission.go
+++ b/plugin/pkg/admission/noderestriction/admission.go
@@ -390,6 +390,11 @@ func (p *Plugin) admitPVCStatus(nodeName string, a admission.Attributes) error {
 
 func (p *Plugin) admitNode(nodeName string, a admission.Attributes) error {
 	requestedName := a.GetName()
+
+	if requestedName != nodeName {
+		return admission.NewForbidden(a, fmt.Errorf("node %q is not allowed to modify node %q", nodeName, requestedName))
+	}
+
 	if a.GetOperation() == admission.Create {
 		node, ok := a.GetObject().(*api.Node)
 		if !ok {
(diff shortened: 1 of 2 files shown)


liggitt:
looks like https://github.com/kubernetes/kubernetes/pull/53185/commits/34cab8f80a9104be7fe5d8a5a22c3207dab46268 removed the need to delay the check until after the create block

also looks like unit tests need updating to:
1. populate admission attributes on create tests to include the name
2. adjust the expected error

fisherxu:
> looks like [34cab8f](https://github.com/kubernetes/kubernetes/commit/34cab8f80a9104be7fe5d8a5a22c3207dab46268) removed the need to delay the check until after the create block
> 
> also looks like unit tests need updating to:
> 
> 1. populate admission attributes on create tests to include the name
> 2. adjust the expected error

Done, has updated the unit tests, added the node name in request info, or it can pull the node name from node object.  @liggitt

plugin/pkg/admission/noderestriction/admission.go
```diff
@@ -390,6 +390,11 @@ func (p *Plugin) admitPVCStatus(nodeName string, a admission.Attributes) error {
 
 func (p *Plugin) admitNode(nodeName string, a admission.Attributes) error {
 	requestedName := a.GetName()
+
+	if requestedName != nodeName {
+		return admission.NewForbidden(a, fmt.Errorf("node %q is not allowed to create or modify node %q", nodeName, requestedName))
```
liggitt, line 395: keep the message text the same as it was previously and revert the unit test change... I didn't notice the message had changed

fisherxu: Done ;-)

PR #93394, opened 2020-07-23
Shorten watch restart test, run in parallel

liggitt:
**What type of PR is this?**
/kind cleanup

Reduces the test time for this package from ~6 minutes to ~50 seconds

Almost all the time is spent waiting for watches to timeout and fail or recover.

Best viewed ignoring whitespace: https://github.com/kubernetes/kubernetes/pull/93394/files?w=1

```release-note
NONE
```

/cc @caesarxuchao 
fyi @BenTheElder 
/sig api-machinery

### DIFF
--- a/test/integration/apimachinery/watch_restart_test.go
+++ b/test/integration/apimachinery/watch_restart_test.go
@@ -64,7 +64,7 @@ func noop() {}
 
 func TestWatchRestartsIfTimeoutNotReached(t *testing.T) {
 	// Has to be longer than 5 seconds
-	timeout := 2 * time.Minute
+	timeout := 30 * time.Second
 
 	// Set up a master
 	masterConfig := framework.NewIntegrationTestMasterConfig()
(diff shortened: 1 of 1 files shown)


test/integration/apimachinery/watch_restart_test.go
```diff
@@ -204,71 +204,74 @@ func TestWatchRestartsIfTimeoutNotReached(t *testing.T) {
...
-
-			// Record current time to be able to asses if the timeout has been reached
-			startTime := time.Now()
-			ctx, cancel := watchtools.ContextWithOptionalTimeout(context.Background(), timeout)
-			defer cancel()
-			_, err = watchtools.UntilWithoutRetry(ctx, watcher, func(event watch.Event) (bool, error) {
-				s, ok := event.Object.(*corev1.Secret)
-				if !ok {
-					t.Fatalf("Received an object that is not a Secret: %#v", event.Object)
+	t.Run("group", func(t *testing.T) {
```
caesarxuchao, line 207: What's the purpose of this outer `t.Run`? It doesn't change the control of parallelism.

liggitt: it ensures the outer setup/teardown code doesn't run until all the subtests are complete:

See https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks

> Run does not return until parallel subtests have completed, providing a way to clean up after a group of parallel tests:
> 
> ```
> func TestTeardownParallel(t *testing.T) {
>     // This Run will not return until the parallel tests finish.
>     t.Run("group", func(t *testing.T) {
>         t.Run("Test1", parallelTest1)
>         t.Run("Test2", parallelTest2)
>         t.Run("Test3", parallelTest3)
>     })
>     // <tear-down code>
> }
> ```

caesarxuchao: Got it. I found https://github.com/golang/go/issues/17791 explaining the ordering better than the doc.

PR #93398, opened 2020-07-23
Allow integration test servers extra time to start

liggitt:
**What type of PR is this?**
/kind flake

In CI environments without reserved resources for the test pod, the CPU dedicated to an individual integration test can be much lower than expected. The test apiserver has much more to do during startup before /healthz will return healthy (all post-startup hooks have to complete), and occasionally this does not complete within 30 seconds.

See https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&text=%2Fhealthz&job=pull-kubernetes-integration

Avoid failing tests due to unrelated CPU starvation.

```release-note
NONE
```

/sig api-machinery
/cc @deads2k


hasheddan (approved):
/lgtm
/retest

liggitt:
including in the milestone to improve CI signal

PR #93401, opened 2020-07-23
[DO NOT MERGE] Add health-for-strangers FlowSchema

### LINKED ISSUE #93359: kube-apiserver livenessProbe needs to be excempt from flowcontrol
on kubernetes 1.18 with flowcontrol enabled

when our api-server is under heavy load for a long time it dies because the livenessProbe fails
this is dangerous since it can lead to etcd compaction never running and then etcd falling over

I did not find a way to configure a serviceaccount for livenessProbe except for exposing the token in the pod config that could then be stolen by any user with pod read access [suggested here](https://github.com/kubernetes/kubernetes/issues/43784)

is it possible to either configure a serviceaccount for the probe or make the livenessProbe itself exempt from flowcontrol / given a higher priority ?

Another way we found was using an exec probe and making that read the service-account token from disk, then giving that token priority, but that's still pretty hacky.

MikeSpreitzer:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR adds the FlowSchema named health-for-strangers,
to prevent crowding out health checks from control-node kubelets,
which typically make secured requests with no credentials.
This FlowSchema exempts from limitation all unauthenticated requests to `/healthz`, `/livez`, and `/readyz`.

This is a quick-fix for issue #93359
but opens the server to overload from anybody.

Fixes #93359

```release-note
Updated the suggested API Priority and Fairness configuration to include a FlowSchema named "health-for-strangers" that exempts from limitation every unauthenticated request to /healthz, /livez, and /readyz , so that other requests do not starve control-node kubelets from successfully reading those paths.
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/apis/flowcontrol/bootstrap/default.go
+++ b/staging/src/k8s.io/apiserver/pkg/apis/flowcontrol/bootstrap/default.go
@@ -70,6 +70,7 @@ var (
 		SuggestedFlowSchemaKubeScheduler,             // references "workload-high" priority-level
 		SuggestedFlowSchemaKubeSystemServiceAccounts, // references "workload-high" priority-level
 		SuggestedFlowSchemaServiceAccounts,           // references "workload-low" priority-level
+		SuggestedFlowSchemaHealthForStrangers,        // references "exempt" priority-level
 		SuggestedFlowSchemaGlobalDefault,             // references "global-default" priority-level
 	}
 )
@@ -373,6 +374,18 @@ var (
 				nonResourceRule(
 					[]string{flowcontrol.VerbAll},
 					[]string{flowcontrol.NonResourceAll}),
+			},
+		},
+	)
+	SuggestedFlowSchemaHealthForStrangers = newFlowSchema(
+		"health-for-strangers", "exempt", 1000,
+		"",
+		flowcontrol.PolicyRulesWithSubjects{
+			Subjects: groups(user.AllUnauthenticated),
+			NonResourceRules: []flowcontrol.NonResourcePolicyRule{
+				nonResourceRule(
+					[]string{flowcontrol.VerbAll},
+					[]string{"/healthz", "/livez", "/readyz"}),
 			},
 		},
 	)


MikeSpreitzer:
@stp-ip 
@yue9944882 
@lavalamp 
@deads2k 
@rngy

MikeSpreitzer:
Here is a YAML expression of the added config object:
```
apiVersion: flowcontrol.apiserver.k8s.io/v1alpha1
kind: FlowSchema
metadata:
  name: health-for-all
spec:
  matchingPrecedence: 1000
  priorityLevelConfiguration:
    name: exempt
  rules:
  - nonResourceRules:
    - nonResourceURLs:
      - "/healthz"
      - "/livez"
      - "/readyz"
      verbs:
      - "*"
    subjects:
    - kind: Group
      group: 
        name: system:unauthenticated 
    - kind: Group
      group: 
        name: system:authenticated 
```

grosser (approved):

MikeSpreitzer:
@logicalhan suggested merging this into #93311 , but I am not sure we want it being part of the default, since it opens a big hole.

staging/src/k8s.io/apiserver/pkg/apis/flowcontrol/bootstrap/default.go
```diff
@@ -376,6 +377,18 @@ var (
...
 	)
+	SuggestedFlowSchemaHealthForStrangers = newFlowSchema(
+		"health-for-strangers", "exempt", 1000,
+		"",
+		flowcontrol.PolicyRulesWithSubjects{
+			Subjects: groups(user.AllUnauthenticated),
+			NonResourceRules: []flowcontrol.NonResourcePolicyRule{
+				nonResourceRule(
+					[]string{flowcontrol.VerbAll},
+					[]string{"/healthz", "/livez", "/readyz"}),
```
yue9944882, line 388: `/metrics` should be in the list, is it?

staging/src/k8s.io/apiserver/pkg/apis/flowcontrol/bootstrap/default.go
```diff
@@ -376,6 +377,18 @@ var (
 			},
 		},
 	)
+	SuggestedFlowSchemaHealthForStrangers = newFlowSchema(
+		"health-for-strangers", "exempt", 1000,
```
yue9944882, line 381: AFAIK, downloading the metrics too frequently can also impact the performance of apiserver, make this exempt looks suspicious to me..

staging/src/k8s.io/apiserver/pkg/apis/flowcontrol/bootstrap/default.go
yue9944882: how about the other health sub-checks? consider `/[healthz|livez|readyz]/*` in the list?

MikeSpreitzer: This was a draft of a response to #93359 , which specified exactly the paths listed here.

MikeSpreitzer:
Per  https://github.com/kubernetes/kubernetes/issues/93359#issuecomment-663286899 , creating a doc PR instead.
/close

staging/src/k8s.io/apiserver/pkg/apis/flowcontrol/bootstrap/default.go
logicalhan: We don't want the detailed health check endpoints to be exempt, since they aren't in any of the critical health-checking paths (i.e. control-plane kubelet liveness-readiness probing).

Issue #92776, opened 2020-07-03
[sig-network] EndpointSlice should create and delete Endpoints and EndpointSlices for a Service with a selector specified


RobertKielty:
**Which jobs are flaking**: 
gce-ubuntu-master-default

**Which test(s) are flaking**: 
[sig-network] EndpointSlice should create and delete Endpoints and EndpointSlices for a Service with a selector specified

**Testgrid link**:
https://testgrid.k8s.io/sig-release-master-informing#gce-ubuntu-master-default&sort-by-flakiness=&embed=

**Reason for failure**:
from [Spyglass for job 1278952987569426433](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-ubuntu-gce/1278952987569426433)

we have

[I0703 07:42:24.519 Jul  3 07:42:20.620: FAIL: EndpointSlice resource not deleted after Service endpointslice-3248/example-empty-selector was deleted: timed out waiting for the condition](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-ubuntu-gce/1278952987569426433#1:build-log.txt%3A40945)

following thru to the [corresponding gubernator logs/ci-kubernetes-e2e-ubuntu-gce/1278952987569426433/]( https://gubernator.k8s.io/build/kubernetes-jenkins/logs/ci-kubernetes-e2e-ubuntu-gce/1278952987569426433/nodelog?junit=junit_09.xml&wrap=on)

we see lot of occurrences of 

`W0703 07:45:10.958376       8 request.go:161] Auditing failed of  request: encoding failed: /v1, Kind=DeleteOptions is unstructured and is not suitable for converting to "cloud.google.com/v1beta1"`

**Anything else we need to know**:
- links to go.k8s.io/triage appreciated

athenabot:
/triage unresolved

Comment `/remove-triage unresolved` when the issue is assessed and confirmed.

🤖 I am a bot run by vllry. 👩‍🔬

robscott:
Thanks for filing this bug @RobertKielty ! The `encoding failed: /v1, Kind=DeleteOptions` is a very strange error message, not sure what's happening there. Am I correct that this test has only failed once? Are there other test failures elsewhere?

athenabot:
@robscott
If this issue has been triaged, please comment `/remove-triage unresolved`.

[...]

PR #93402, opened 2020-07-23
Updating EndpointSlice e2e tests to be less flaky and easier to debug

### LINKED ISSUE #93374: [Flaky test] EndpointSlice should create Endpoints and EndpointSlices for Pods matching a Service
**Which jobs are flaking**:

* ci-kubernetes-kind-e2e-parallel
* pr:pull-kubernetes-e2e-gce
* pr:pull-kubernetes-e2e-gce-ubuntu-containerd
* pr:pull-kubernetes-e2e-gce-ubuntu-containerd
* ci-kubernetes-kind-ipv6-e2e-parallel

Second flakiest test in the e2e job according to http://storage.googleapis.com/k8s-metrics/flakes-latest.json

**Testgrid link**:

https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-e2e-gce&include-filter-by-regex=EndpointSlice%20should%20create%20Endpoints%20and%20EndpointSlices%20for%20Pods%20matching%20a%20Service&width=5

**Reason for failure**:

```
test/e2e/network/endpointslice.go:70
Jul 22 22:13:01.180: EndpointSlice resource not deleted after Service endpointslice-4585/example-empty-selector was deleted: timed out waiting for the condition
test/e2e/network/endpointslice.go:158
```

https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&test=EndpointSlice%20should%20create%20Endpoints%20and%20EndpointSlices%20for%20Pods%20matching%20a%20Service

/sig network
/assign @robscott

robscott:
**What type of PR is this?**
/kind cleanup
/kind failing-test

This adjusts tests that were waiting for Pods to be ready to instead just wait for them to have IPs assigned to them. This relies on the associated publishNotReadyAddresses field on Services. Additionally this increases the the length of time we'll wait for EndpointSlices to be garbage collected from 12s to 18s. Finally, this adds additional logging to ExpectNoError calls so it's easier to understand where and why a test failed.

I'm hoping this PR will fix #93374 and #92776, but all I can say for sure is that it "works on my cluster". 

Although it seems backwards to even be testing for EndpointSlice garbage collection we need some kind of test coverage to make sure that the EndpointSlice controller does not conflict with the garbage collector. Maybe there's a better way to do that than have this long timeout here. It could potentially be more straightforward to just immediately delete both Endpoints and EndpointSlices when a Service is marked for deletion, but that could potentially be undesirable if there are long running finalizers in place.

```release-note
NONE
```

/cc @aojea @freehan @RobertKielty @hasheddan 
/assign @liggitt

### DIFF
--- a/test/e2e/network/endpointslice.go
+++ b/test/e2e/network/endpointslice.go
@@ -49,14 +49,14 @@ var _ = SIGDescribe("EndpointSlice", func() {
 		namespace := "default"
 		name := "kubernetes"
 		endpoints, err := cs.CoreV1().Endpoints(namespace).Get(context.TODO(), name, metav1.GetOptions{})
-		framework.ExpectNoError(err)
+		framework.ExpectNoError(err, "error creating Endpoints resource")
 		if len(endpoints.Subsets) != 1 {
 			framework.Failf("Expected 1 subset in endpoints, got %d: %#v", len(endpoints.Subsets), endpoints.Subsets)
 		}
 
 		endpointSubset := endpoints.Subsets[0]
 		endpointSlice, err := cs.DiscoveryV1beta1().EndpointSlices(namespace).Get(context.TODO(), name, metav1.GetOptions{})
-		framework.ExpectNoError(err)
+		framework.ExpectNoError(err, "error creating EndpointSlice resource")
 		if len(endpointSlice.Ports) != len(endpointSubset.Ports) {
 			framework.Failf("Expected EndpointSlice to have %d ports, got %d: %#v", len(endpointSubset.Ports), len(endpointSlice.Ports), endpointSlice.Ports)
 		}
@@ -85,7 +85,7 @@ var _ = SIGDescribe("EndpointSlice", func() {
 		})
 
 		// Expect Endpoints resource to be created.
-		if err := wait.PollImmediate(2*time.Second, 12*time.Second, func() (bool, error) {
+		if err := wait.PollImmediate(2*time.Second, wait.ForeverTestTimeout, func() (bool, error) {
 			_, err := cs.CoreV1().Endpoints(svc.Namespace).Get(context.TODO(), svc.Name, metav1.GetOptions{})
 			if err != nil {
 				return false, nil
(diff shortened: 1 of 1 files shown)


test/e2e/network/endpointslice.go
```diff
@@ -126,9 +126,10 @@ var _ = SIGDescribe("EndpointSlice", func() {
 		}
 
 		err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
-		framework.ExpectNoError(err)
+		framework.ExpectNoError(err, "error deleting Service")
 
-		// Expect Endpoints resource to be deleted when Service is.
+		// Expect Endpoints resource to not be deleted when Service is.
+		// TODO(robscott): Update this test when this bug is fixed.
```
hasheddan, line 132: If we expect that the `Endpoints` resource will not be deleted here wouldn't this test always fail?

aojea: yep, I don't understand the comment either, the code seems to check they are deleted

robscott: 🤦 Apparently I momentarily confused this part with another part that is waiting for an Endpoints related fix and thought I'd forgotten to add a TODO here. As it turns out the TODO already existed in the appropriate place: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/endpointslice.go#L296-L297. That bug will actually be fixed with an [open PR](https://github.com/kubernetes/kubernetes/pull/93030) so I need to remember to connect those dots.

test/e2e/network/endpointslice.go
```diff
@@ -143,7 +143,7 @@ var _ = SIGDescribe("EndpointSlice", func() {
 		}
 
 		// Expect EndpointSlice resource to be deleted when Service is.
-		if err := wait.PollImmediate(2*time.Second, 12*time.Second, func() (bool, error) {
+		if err := wait.PollImmediate(2*time.Second, 18*time.Second, func() (bool, error) {
```
liggitt, line 146: this is a surprisingly specific time... why not wait.ForeverTestTimeout (30 seconds)?

test/e2e/network/endpointslice.go
```diff
@@ -126,7 +126,7 @@ var _ = SIGDescribe("EndpointSlice", func() {
 		}
 
 		err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
-		framework.ExpectNoError(err)
+		framework.ExpectNoError(err, "error deleting Service")
 
 		// Expect Endpoints resource to be deleted when Service is.
 		if err := wait.PollImmediate(2*time.Second, 12*time.Second, func() (bool, error) {
```
liggitt, line 132: same question about why we're picking such a specific duration

test/e2e/network/endpointslice.go
robscott: Somehow I think I'd thought the only reason it would take longer than this would be a legitimate failure, clearly that's not always the case. Also, I didn't know about `wait.ForeverTestTimeout` - I've updated all 4 instances of 12 second times to use that instead, thanks!

test/e2e/network/endpointslice.go
robscott: Updated, thanks!

PR #93403, opened 2020-07-24
Remove unuseful error message about updating pod conditions not owned…

hex108:
… by kubelet

**What type of PR is this?**
/kind cleanup

The error message seems not useful and not needed, because kubelet will not update these conditions when [mergePodStatus]( https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/status/status_manager.go#L718). After using [pod readniess gates](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0007-pod-ready%2B%2B.md), many pods have their own conditions which are not owned by kubelet, and a lot of error messages are really noisy and misleading.

Fixes #

```release-note
NONE
```

```docs

```

/sig node

### DIFF
--- a/pkg/kubelet/status/status_manager.go
+++ b/pkg/kubelet/status/status_manager.go
@@ -189,12 +189,6 @@ func (m *manager) SetPodStatus(pod *v1.Pod, status v1.PodStatus) {
 	m.podStatusesLock.Lock()
 	defer m.podStatusesLock.Unlock()
 
-	for _, c := range pod.Status.Conditions {
-		if !kubetypes.PodConditionByKubelet(c.Type) {
-			klog.Errorf("Kubelet is trying to update pod condition %q for pod %q. "+
-				"But it is not owned by kubelet.", string(c.Type), format.Pod(pod))
-		}
-	}
 	// Make sure we're caching a deep copy.
 	status = *status.DeepCopy()
 


SergeyKanzhelev:
similar to https://github.com/kubernetes/kubernetes/pull/91049

SergeyKanzhelev:
let's discuss what's the right fix there.

/hold

SergeyKanzhelev:
/hold cancel
/retest

See discussion here: https://github.com/kubernetes/kubernetes/pull/91049#issuecomment-667418639 it looks like the right fix is to remove this error message.

SergeyKanzhelev:
@kubernetes/release-managers I wonder if this can be considered for 1.19 release? It's small and straightforward. It also noisy for whoever using custom conditions.

tpepper:
> @kubernetes/release-managers I wonder if this can be considered for 1.19 release? It's small and straightforward. It also noisy for whoever using custom conditions.

If this merges to master in the next couple days it will be in 1.19, we're still fast forwarding content from the frozen master branch to release-1.19 branch.  To merge during code freeze, this will need SIG Node's approval and adding it to the milestone.

derekwaynecarr:
From the history in this code block, its not clear to me why we needed the check here.  I agree merge pod status will ignore and append conditions not sourced from kubelet prior to doing the patch, so this error is confusing and not accurate.

/approve
/lgtm
/milestone v1.19

PR #93408, opened 2020-07-24
jsonpath: disallow multiple immediate recursive descent

### LINKED ISSUE #93400: Exponential entity expansion allowed by jsonpath function
**What happened**:

This issue was originally reported via the Kubernetes HackerOne bug bounty by HackerOne user [lazydog](https://hackerone.com/lazydog) but we have decided that due to the high privileges required, it can be addressed publicly.

A user able to create CRDs could create a malicious CRD such that listing CRs will cause enormous amounts of CPU usage on the API server.

The `jsonpath` function should not expand strings with more than two consecutive periods. Either it should collapse additional repeated periods or it should return an error.

```
$ kubectl create -f - <<EOF
---
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: foos.example.com
spec:
  group: example.com
  scope: Namespaced
  names:
    plural: foos
    singular: foo
    kind: Foo
  version: v1
  additionalPrinterColumns:
  - name: FOO
    type: string
    JSONPath: ........................................................................................................................................................................................................
EOF

$ kubectl create -f - <<EOF
---
apiVersion: example.com/v1
kind: Foo
metadata:
  name: foo-cr
spec:
  foo:
    bar:
      baz:
        qux: data
EOF

$ kubectl get foo
# ( and the API server CPU usage jumps way up )
```

/kind bug
/area apiserver
/area security
/committee product-security
/priority important-longterm

joelsmith:
**What type of PR is this?**
/kind bug

Disallows invalid jsonpath expressions such as `.....................................` which currently cause nested recursion and can use large amounts of CPU.

Fixes #93400

```release-note
kube-apiserver: jsonpath expressions with consecutive recursive descent operators are no longer evaluated for custom resource printer columns
```

### DIFF
--- a/staging/src/k8s.io/client-go/util/jsonpath/jsonpath_test.go
+++ b/staging/src/k8s.io/client-go/util/jsonpath/jsonpath_test.go
@@ -263,6 +263,8 @@ func TestStructInput(t *testing.T) {
 		{"allarray", "{.Book[*].Author}", storeData, "Nigel Rees Evelyn Waugh Herman Melville", false},
 		{"allfields", `{range .Bicycle[*]}{ "{" }{ @.* }{ "} " }{end}`, storeData, "{red 19.95 true} {green 20.01 false} ", false},
 		{"recurfields", "{..Price}", storeData, "8.95 12.99 8.99 19.95 20.01", false},
+		{"recurdotfields", "{...Price}", storeData, "8.95 12.99 8.99 19.95 20.01", false},
+		{"superrecurfields", "{............................................................Price}", storeData, "", true},
 		{"allstructsSlice", "{.Bicycle}", storeData,
 			`[{"Color":"red","Price":19.95,"IsNew":true},{"Color":"green","Price":20.01,"IsNew":false}]`, false},
 		{"allstructs", `{range .Bicycle[*]}{ @ }{ " " }{end}`, storeData,
--- a/staging/src/k8s.io/client-go/util/jsonpath/parser.go
+++ b/staging/src/k8s.io/client-go/util/jsonpath/parser.go
@@ -214,8 +214,11 @@ func (p *Parser) parseIdentifier(cur *ListNode) error {
 	return p.parseInsideAction(cur)
 }
 
-// parseRecursive scans the recursive desent operator ..
+// parseRecursive scans the recursive descent operator ..
 func (p *Parser) parseRecursive(cur *ListNode) error {
+	if lastIndex := len(cur.Nodes) - 1; lastIndex >= 0 && cur.Nodes[lastIndex].Type() == NodeRecursive {
+		return fmt.Errorf("invalid multiple recursive descent")
+	}
 	p.pos += len("..")
 	p.consumeText()
 	cur.append(newRecursive())
--- a/staging/src/k8s.io/client-go/util/jsonpath/parser_test.go
+++ b/staging/src/k8s.io/client-go/util/jsonpath/parser_test.go
@@ -141,6 +141,7 @@ func TestFailParser(t *testing.T) {
 		{"invalid number", "{+12.3.0}", "cannot parse number +12.3.0"},
 		{"unterminated array", "{[1}", "unterminated array"},
 		{"unterminated filter", "{[?(.price]}", "unterminated filter"},
+		{"invalid multiple recursive descent", "{........}", "invalid multiple recursive descent"},
 	}
 	for _, test := range failParserTests {
 		_, err := Parse(test.name, test.text)


staging/src/k8s.io/client-go/util/jsonpath/jsonpath_test.go
```diff
@@ -263,6 +263,7 @@ func TestStructInput(t *testing.T) {
 		{"allarray", "{.Book[*].Author}", storeData, "Nigel Rees Evelyn Waugh Herman Melville", false},
 		{"allfields", `{range .Bicycle[*]}{ "{" }{ @.* }{ "} " }{end}`, storeData, "{red 19.95 true} {green 20.01 false} ", false},
 		{"recurfields", "{..Price}", storeData, "8.95 12.99 8.99 19.95 20.01", false},
+		{"superrecurfields", "{............................................................Price}", storeData, "", true},
```
knight42, line 266: should we also add a test to `TestFailParser` in `parser_test.go` to check if the expected error is returned?

staging/src/k8s.io/client-go/util/jsonpath/parser.go
```diff
@@ -221,6 +221,9 @@ func (p *Parser) parseRecursive(cur *ListNode) error {
 	cur.append(newRecursive())
 	if r := p.peek(); isAlphaNumeric(r) {
 		return p.parseField(cur)
+	} else if r == '.' {
+		// one or more periods following the recursive decent operator .. is invalid
+		return fmt.Errorf("invalid recursive decent")
```
liggitt, line 226: it looks like returning an error here will make the CR handler fall back to default printing:

https://github.com/kubernetes/kubernetes/blob/607c5daabd4204fcebe90f983ba06edfcc6fc4e5/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/customresource_handler.go#L751-L754

Do we need to do something to detect this and make it more visible in the CRD status?

liggitt: cc @sttts

joelsmith: We could, but IMO that's a separate issue that already exists today for any invalid jsonpath expression in a CRD column.

Also, it's probably worth mentioning that we could swallow all the extra dots immediately following 2 dots (instead of returning an error) if we think that's better.

knight42: In addition, we could also validate the jsonpath in apiserver.

liggitt: tightening validation introduces skew problems with existing persisted data. let's not add that without further discussion.

staging/src/k8s.io/client-go/util/jsonpath/parser.go
```diff
@@ -221,6 +221,9 @@ func (p *Parser) parseRecursive(cur *ListNode) error {
 	cur.append(newRecursive())
 	if r := p.peek(); isAlphaNumeric(r) {
 		return p.parseField(cur)
+	} else if r == '.' {
+		// one or more periods following the recursive descent operator .. is invalid
```
liggitt, line 225: why is one period invalid?

both of these appear equivalent today:
```
kubectl get ns -o 'jsonpath={..spec}'
kubectl get ns -o 'jsonpath={...spec}'
```

that also matches the behavior of https://jsonpath.com/ and both implementations at https://jsonpath.curiousconcept.com/

preventing double-recursion (`....`) seems more reasonable to me

joelsmith: I suppose one period following `..` could be considered valid, even if it has no effect. I have an alternate version of the code which ignores any periods immediately following `..`
I think one could argue that is a valid interpretation of what a jsonpath author intended if their expression includes more than 2 periods.

```diff
 // parseRecursive scans the recursive desent operator ..

 func (p *Parser) parseRecursive(cur *ListNode) error {
        p.pos += len("..")
        p.consumeText()
        cur.append(newRecursive())
+       // swallow any periods immediately following the recursive descent operator
+       for r := p.peek(); r == '.'; r = p.peek() {
+               p.pos += len(".")
+               p.consumeText()
+       }
        if r := p.peek(); isAlphaNumeric(r) {
                return p.parseField(cur)
        }
        return p.parseInsideAction(cur)
 }

```

I'm happy to post that version instead if it seems like more appropriate behavior.

liggitt: it is specifically the repeated recursive descent that seems invalid to me

would something like this work:

```diff
diff --git a/staging/src/k8s.io/client-go/util/jsonpath/parser.go b/staging/src/k8s.io/client-go/util/jsonpath/parser.go
index e1aab6804f8..62a2c62abe4 100644
--- a/staging/src/k8s.io/client-go/util/jsonpath/parser.go
+++ b/staging/src/k8s.io/client-go/util/jsonpath/parser.go
@@ -216,6 +216,9 @@ func (p *Parser) parseIdentifier(cur *ListNode) error {
 
 // parseRecursive scans the recursive desent operator ..
 func (p *Parser) parseRecursive(cur *ListNode) error {
+	if lastIndex := len(cur.Nodes) - 1; lastIndex >= 0 && cur[lastIndex].Type() == NodeRecursive {
+		return fmt.Errorf("invalid multiple recursive descent")
+	}
 	p.pos += len("..")
 	p.consumeText()
 	cur.append(newRecursive())
```

joelsmith: Yes, that worked (with a slight modification). I have updated the unit tests accordingly.

liggitt:

/hold

PR #81549, opened 2019-08-17
Support GlobalAccess for gce Internal Loadbalancers

prameshj:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:
This PR adds Support for GlobalAccess for gce Internal Loadbalancers.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Added support to specify a global-access annotation for gce ILB.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #93423, opened 2020-07-24
Fix GCE ILB code to support protocol change

prameshj:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This bug was introduced when adding support for Global Access - https://github.com/kubernetes/kubernetes/pull/81549
The forwarding rule deletion logic was refactored, which will now error out if ILB service protocol is changed. This PR fixes the behavior and adds a unit test.

Fixes # 93422

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/legacy-cloud-providers/gce/gce_loadbalancer_internal.go
+++ b/staging/src/k8s.io/legacy-cloud-providers/gce/gce_loadbalancer_internal.go
@@ -173,12 +173,6 @@ func (g *Cloud) ensureInternalLoadBalancer(clusterName, clusterID string, svc *v
 		return nil, err
 	}
 
-	bsDescription := makeBackendServiceDescription(nm, sharedBackend)
-	err = g.ensureInternalBackendService(backendServiceName, bsDescription, svc.Spec.SessionAffinity, scheme, protocol, igLinks, hc.SelfLink)
-	if err != nil {
-		return nil, err
-	}
-
 	fwdRuleDescription := &forwardingRuleDescription{ServiceName: nm.String()}
 	fwdRuleDescriptionString, err := fwdRuleDescription.marshal()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


staging/src/k8s.io/legacy-cloud-providers/gce/gce_loadbalancer_internal.go
```diff
@@ -200,10 +194,31 @@ func (g *Cloud) ensureInternalLoadBalancer(clusterName, clusterID string, svc *v
 	if options.AllowGlobalAccess {
 		newFwdRule.AllowGlobalAccess = options.AllowGlobalAccess
 	}
-	if err := g.ensureInternalForwardingRule(existingFwdRule, newFwdRule); err != nil {
+
+	fwdRuleDeleted := false
+	if existingFwdRule != nil && !forwardingRulesEqual(existingFwdRule, newFwdRule) {
```
freehan, line 199: I have some concern regarding this line. If there is any bug in the logic generating newFwdRule, this is likely to cause a delete/create loop for the forwarding rules.

The bug can potentially due to 1. a new version of controller added a new field 2. GCE defaulting behavior change etc

Feels a bit risky to compare the whole forwarding rule. Does it make sense to do spot checks? For instance, check the protocol and port and etc....

freehan: Not sure how was this handled before?

prameshj: This is the same code from before - https://github.com/kubernetes/kubernetes/blob/a34f1e483104bd51c3e9a6aec3dbbcf6301789da/staging/src/k8s.io/legacy-cloud-providers/gce/gce_loadbalancer_internal.go#L143

prameshj: forwardingRulesEqual compares select fields - https://github.com/kubernetes/kubernetes/blob/a34f1e483104bd51c3e9a6aec3dbbcf6301789da/staging/src/k8s.io/legacy-cloud-providers/gce/gce_loadbalancer_internal.go#L750

freehan (approved):
/lgtm
/approve

prameshj:
Thanks @justaugustus !

PR #93425, opened 2020-07-24
Fix int->string casts

### LINKED ISSUE #93407: The build failed for some tests while running make test fo GO version go1.15beta1
Test Failures are
```
FAIL    k8s.io/kubernetes/pkg/kubelet [build failed]
FAIL    k8s.io/kubernetes/pkg/kubelet/container [build failed]
FAIL    k8s.io/kubernetes/pkg/kubelet/network/dns [build failed]
FAIL    k8s.io/kubernetes/pkg/scheduler [build failed]
FAIL    k8s.io/kubernetes/vendor/k8s.io/client-go/plugin/pkg/client/auth/azure [build failed]
FAIL    k8s.io/kubernetes/vendor/k8s.io/client-go/tools/record [build failed]
FAIL    k8s.io/kubernetes/vendor/k8s.io/kubectl/pkg/cmd/create [build failed]
FAIL    k8s.io/kubernetes/vendor/k8s.io/kubectl/pkg/drain [build failed]
FAIL    k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/gce [build failed]
FAIL
```

**Environment**:
Client Version: version.Info{Major:"1", Minor:"20+", GitVersion:"v1.20.0-alpha.0.300+ff33efc164e78d-dirty", GitCommit:"ff33efc164e78d12c2a250f901b6ec725184fd06", GitTreeState:"dirty", BuildDate:"2020-07-22T10:11:16Z", GoVersion:"go1.15beta1", Compiler:"gc", Platform:"linux/ppc64le"}
linux/ppc64le
```NAME="Red Hat Enterprise Linux"
VERSION="8.1 (Ootpa)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="8.1"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Red Hat Enterprise Linux 8.1 (Ootpa)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:8.1:GA"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
```
**Full logs of make file**
[make.log](https://github.com/kubernetes/kubernetes/files/4970250/make.log)

**Full Logs of make test**
[make_test.log](https://github.com/kubernetes/kubernetes/files/4970245/make_test.log)

Golang Location: https://dl.google.com/go/go1.15beta1.linux-ppc64le.tar.gz

liggitt:
**What type of PR is this?**
/kind bug

Fixes incorrect int->string casts, which result in a single-character rune string.

go1.15 flags these as build errors, but we should fix them even on earlier versions

Fixes https://github.com/kubernetes/kubernetes/issues/93407

```release-note
NONE
```

/cc @justaugustus

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"math/rand"
 	"os"
+	"strconv"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
@@ -613,7 +614,7 @@ func getAttemptsLabel(p *framework.QueuedPodInfo) string {
 	if p.Attempts >= 15 {
 		return "15+"
 	}
-	return string(p.Attempts)
+	return strconv.Itoa(p.Attempts)
 }
 
 func (sched *Scheduler) profileForPod(pod *v1.Pod) (*profile.Profile, error) {
--- a/pkg/kubelet/container/cache_test.go
+++ b/pkg/kubelet/container/cache_test.go
@@ -49,7 +49,7 @@ func getTestPodIDAndStatus(numContainers int) (types.UID, *PodStatus) {
 		status = &PodStatus{ID: id}
 	}
 	for i := 0; i < numContainers; i++ {
-		status.ContainerStatuses = append(status.ContainerStatuses, &Status{Name: string(i)})
+		status.ContainerStatuses = append(status.ContainerStatuses, &Status{Name: strconv.Itoa(i)})
 	}
 	return id, status
 }
--- a/pkg/kubelet/kubelet_test.go
+++ b/pkg/kubelet/kubelet_test.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"os"
 	"sort"
+	"strconv"
 	"testing"
 	"time"
 
@@ -355,7 +356,7 @@ func newTestPods(count int) []*v1.Pod {
 				HostNetwork: true,
 			},
 			ObjectMeta: metav1.ObjectMeta{
-				UID:  types.UID(10000 + i),
+				UID:  types.UID(strconv.Itoa(10000 + i)),
 				Name: fmt.Sprintf("pod%d", i),
 			},
 		}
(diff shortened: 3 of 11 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -607,13 +607,18 @@ func (sched *Scheduler) scheduleOne(ctx context.Context) {
...
 	// We breakdown the pod scheduling duration by attempts capped to a limit
 	// to avoid ending up with a high cardinality metric.
 	if p.Attempts >= 15 {
 		return "15+"
 	}
-	return string(p.Attempts)
+	if p.Attempts < 0 {
+		return "-"
+	}
+	return attempts[p.Attempts]
```
liggitt, line 621: cc @ahg-g 

avoided strconv here since it is expensive for frequently called paths

pkg/scheduler/scheduler.go
```diff
@@ -607,13 +607,18 @@ func (sched *Scheduler) scheduleOne(ctx context.Context) {
 	}()
 }
 
+var attempts = []string{"0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15"}
```
adtac, line 610: Don't need the last element, will never be accessed.

pkg/scheduler/scheduler.go
ahg-g: this is not called frequently, once per scheduled pod.

adtac: FWIW I don't think this codepath is executed that frequently; it's executed once per successfully scheduled pod, so IMO I prefer the simpler approach.

(Also, `Itoa` has a fast path for small ints: https://golang.org/src/strconv/itoa.go?s=784:824#L15)

liggitt: > `Itoa` has a fast path for small ints: https://golang.org/src/strconv/itoa.go?s=784:824#L15

ah, great... I cargo-culted this wisdom from our request metrics which log the response code which can be up in the 5xx range and exceeds the small number optimization

simplified back to strconv here

liggitt:
verify failure, no logs

/retest

liggitt:
> Pod pending timeout.

/retest

justaugustus:
/retest
(pods pending timeout)

justaugustus:
/retest
(pods pending timeout)

PR #93426, opened 2020-07-24
Add subjectAltName to test serving certificates

liggitt:
**What type of PR is this?**
/kind bug

Fixes the test certificates used to include the hostname as a subjectAltName

go1.15 starts enforcing this, but the RFC requires this to be the case already, so go ahead and fix the issue.

```release-note
NONE
```

/cc @justaugustus


liggitt:
> Pod pending timeout.

/retest

PR #93434, opened 2020-07-24
scheduler.go: convert int to string properly

adtac:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**: A `string` cannot be cast into an `int` just like that. This PR adds a `strconv.Itoa` to do this correctly.

```release-note
NONE
```

/sig scheduling
/cc @ahg-g @alculquicondor

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"math/rand"
 	"os"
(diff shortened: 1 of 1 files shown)


liggitt:
this is one of many instances included in https://github.com/kubernetes/kubernetes/pull/93425

adtac:
ack, closing in favour of that PR

/close

PR #93439, opened 2020-07-24
validate types in kubectl expose

### LINKED ISSUE #92901: kubectl expose does not check type validity when --dry-run=client flag is present
**What happened**:

`kubectl expose  RESOURCE RESOURCE_NAME --type=x --port=NUM --dry-run=client -o yaml` 
The above command populates the yaml with `spec.type=x`

The type gets validated before the yaml is populated

```
kubectl run temp --image=nginx 

kubectl expose pod temp --type=x --port=80 --dry-run=client -o yaml
```

**Environment**:
- Kubernetes version (use `kubectl version`): 1.18
- OS (e.g: `cat /etc/os-release`): Ubuntu 18.04
- Kernel (e.g. `uname -a`): Linux master 4.15.0-101-generic
- Install tools:
- Others:

SaiHarshaK:
**What type of PR is this?**
/kind bug

Currently on dry-run the command doesn't validate the service type and output with the incorrect service type. This PR validates the service types passed in kubectl expose, showing appropriate error.
Fixes #92901

```release-note
Will validate service type on dry-run
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/kubectl/pkg/cmd/expose/expose.go
+++ b/staging/src/k8s.io/kubectl/pkg/cmd/expose/expose.go
@@ -17,6 +17,7 @@ limitations under the License.
 package expose
 
 import (
+	"fmt"
 	"regexp"
 	"strings"
 
@@ -98,6 +99,7 @@ type ExposeServiceOptions struct {
 	MapBasedSelectorForObject func(runtime.Object) (string, error)
 	PortsForObject            polymorphichelpers.PortsForObjectFunc
 	ProtocolsForObject        func(runtime.Object) (map[string]string, error)
+	Type                      string
 
 	Namespace string
 	Mapper    meta.RESTMapper
@@ -136,6 +138,7 @@ func NewCmdExposeService(f cmdutil.Factory, streams genericclioptions.IOStreams)
 		Example:               exposeExample,
 		Run: func(cmd *cobra.Command, args []string) {
 			cmdutil.CheckErr(o.Complete(f, cmd))
+			cmdutil.CheckErr(o.Validate())
 			cmdutil.CheckErr(o.RunExpose(cmd, args))
 		},
 		ValidArgs: validArgs,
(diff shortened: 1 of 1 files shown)


staging/src/k8s.io/kubectl/pkg/cmd/expose/expose.go
```diff
@@ -218,6 +222,18 @@ func (o *ExposeServiceOptions) Complete(f cmdutil.Factory, cmd *cobra.Command) e
 	return err
 }
 
+// Validate checks to the ExposeServiceOptions to see if there is sufficient information run the command.
+func (o *ExposeServiceOptions) Validate() error {
+	if len(o.Type) > 0 {
+		// Valid options for type are ExternalName, ClusterIP, NodePort, and LoadBalancer.
+		if o.Type != "ExternalName" && o.Type != "ClusterIP" && o.Type != "NodePort" && o.Type != "LoadBalancer" {
```
eddiezane, line 229: Is there a list or constants of this somewhere instead of hard coding these? If not we might want to add a `ValidateServiceType` type of function.

staging/src/k8s.io/kubectl/pkg/cmd/expose/expose.go
```diff
@@ -218,6 +222,18 @@ func (o *ExposeServiceOptions) Complete(f cmdutil.Factory, cmd *cobra.Command) e
 	return err
 }
 
+// Validate checks to the ExposeServiceOptions to see if there is sufficient information run the command.
+func (o *ExposeServiceOptions) Validate() error {
```
eddiezane, line 226: Let's add a test for this one.

SaiHarshaK: Okay, will do.

staging/src/k8s.io/kubectl/pkg/cmd/expose/expose.go
eddiezane: Looks like [these](https://github.com/kubernetes/kubernetes/blob/b497fa1d34ec15c4219d591e541c2cf811271357/staging/src/k8s.io/api/core/v1/types.go#L3916-L3934) are used in a [few places](https://cs.k8s.io/?q=ServiceTypeNodePort&i=nope&files=&repos=kubernetes/kubernetes).

Yea let's stick that in there and test it unless someone thinks of a better place.

SaiHarshaK: So i use the constants instead right? Earlier what i meant was that i didn't find all them as a list.

eddiezane:
@SaiHarshaK I synced with @liggitt on this one and his recommendation is to not implement this to avoid tightly coupling the client to the server. It's better to keep the client simple and let the server validate the wrong field. Let's say a new valid service type is added -- users would be required to download a new version kubectl to create a service with the new type.

Appreciate your work on this one and hopefully you learned something about the code base for the future.

Are you ok with closing this?

SaiHarshaK:
Sure, It is fine.
/close

PR #93441, opened 2020-07-25
Fixing memory leak in EndpointSliceMirroring EndpointSlice tracker

robscott:
**What type of PR is this?**
/kind bug

This mirrors an [earlier fix to the EndpointSlice controller](https://github.com/kubernetes/kubernetes/pull/92838). I'll make a follow up PR to move this component to a shared package, but that seems beyond the scope of a bug fix PR.

I'm not a milestone maintainer, but it would be great to get this into 1.19 if anyone can add that to this PR.

```release-note
Fix memory leak in EndpointSliceTracker for EndpointSliceMirroring controller.
```

/sig network
/priority important-soon
/assign @freehan

### DIFF
--- a/pkg/controller/endpointslicemirroring/endpointslice_tracker.go
+++ b/pkg/controller/endpointslicemirroring/endpointslice_tracker.go
@@ -45,61 +45,76 @@ func newEndpointSliceTracker() *endpointSliceTracker {
 	}
 }
 
-// has returns true if the endpointSliceTracker has a resource version for the
+// Has returns true if the endpointSliceTracker has a resource version for the
 // provided EndpointSlice.
-func (est *endpointSliceTracker) has(endpointSlice *discovery.EndpointSlice) bool {
+func (est *endpointSliceTracker) Has(endpointSlice *discovery.EndpointSlice) bool {
 	est.lock.Lock()
 	defer est.lock.Unlock()
 
-	rrv := est.relatedResourceVersions(endpointSlice)
-	_, ok := rrv[endpointSlice.Name]
+	rrv, ok := est.relatedResourceVersions(endpointSlice)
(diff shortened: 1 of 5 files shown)


pkg/controller/endpointslicemirroring/endpointslice_tracker.go
```diff
@@ -45,61 +45,76 @@ func newEndpointSliceTracker() *endpointSliceTracker {
 	}
 }
 
-// has returns true if the endpointSliceTracker has a resource version for the
+// Has returns true if the endpointSliceTracker has a resource version for the
 // provided EndpointSlice.
-func (est *endpointSliceTracker) has(endpointSlice *discovery.EndpointSlice) bool {
+func (est *endpointSliceTracker) Has(endpointSlice *discovery.EndpointSlice) bool {
```
hasheddan, line 50: nitpick: do these need to be exported prior to be moved into a shared package in https://github.com/kubernetes/kubernetes/pull/93443?

robscott: Yeah you're completely right. I was updating this to match the EndpointSlice controller implementation precisely which already included exported functions. Since the end goal was to extract this to a shared package (https://github.com/kubernetes/kubernetes/pull/93443) it seemed like this would result in less long term conflicts. Especially if this commit ends up applying to earlier versions of Kubernetes than the shared package does. I'd thought this might be easier to backport than extracting the tracker into a shared package.

liggitt:
if this is fixing a memory leak introduced in 1.19, then it should be fixed in the 1.19 milestone

/milestone v1.19

freehan (approved):
/lgtm
/approve

PR #93442, opened 2020-07-25
Updating EndpointSliceMirroring controller to copy labels from Endpoints

robscott:
**What type of PR is this?**
/kind bug

The [KEP specifies](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190603-endpointslices/README.md#mirroring-details) that the controller will "mirror all labels from the Endpoints resource and all endpoints and ports from the corresponding subset". I'd missed that in my initial implementation, this should fix that.

I'm not a milestone maintainer, but it would be great to get this into 1.19 if anyone can add that to this PR.

```release-note
EndpointSliceMirroring controller now copies labels from Endpoints to EndpointSlices.
```

/sig network
/priority important-soon
/assign @freehan

### DIFF
--- a/pkg/controller/endpointslicemirroring/utils.go
+++ b/pkg/controller/endpointslicemirroring/utils.go
@@ -88,10 +88,7 @@ func newEndpointSlice(endpoints *corev1.Endpoints, ports []discovery.EndpointPor
 	ownerRef := metav1.NewControllerRef(endpoints, gvk)
 	epSlice := &discovery.EndpointSlice{
 		ObjectMeta: metav1.ObjectMeta{
-			Labels: map[string]string{
-				discovery.LabelServiceName: endpoints.Name,
-				discovery.LabelManagedBy:   controllerName,
-			},
+			Labels:          map[string]string{},
 			OwnerReferences: []metav1.OwnerReference{*ownerRef},
 			Namespace:       endpoints.Namespace,
 		},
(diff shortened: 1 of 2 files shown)


pkg/controller/endpointslicemirroring/utils.go
```diff
@@ -88,10 +88,7 @@ func newEndpointSlice(endpoints *corev1.Endpoints, ports []discovery.EndpointPor
 	ownerRef := metav1.NewControllerRef(endpoints, gvk)
 	epSlice := &discovery.EndpointSlice{
 		ObjectMeta: metav1.ObjectMeta{
-			Labels: map[string]string{
-				discovery.LabelServiceName: endpoints.Name,
-				discovery.LabelManagedBy:   controllerName,
-			},
+			Labels:          endpoints.Labels,
```
liggitt, line 91: this needs to copy the endpoints.Labels map to avoid mutating the informer-owned labels map when we assign LabelServiceName/LabelManagedBy below

liggitt: add a test that verifies the endpoints object passed into `newEndpointSlice` does not have its labels modified

robscott: Thanks for catching this! I've updated this to copy labels and added a test to verify that the labels on the Endpoints resource are not mutated.

Issue #88212, opened 2020-02-16
kube-proxy in ipvs mode has high cpu usage with thousands of services 


lbernail:
**What happened**:
Most of our workloads use headless services but on a specific clusters we have 2000 services with a ClusterIP and ~50 with Nodeports.
While looking a depth at node not performing as expected, we noticed that kube-proxy was using 65+% of core almost constantly.

We took a flamegraph of kube-proxy and noticed that GetNodeAddresses was a very user (accounting for 40+% of cpu usage). I remember a PR related to this (https://github.com/kubernetes/kubernetes/pull/79444) and back-ported it to the version we are using. Many thanks again @cezarsa 
This patch is only available since 1.16 I think it really makes sense to backport it as far as possible (so probably 1.15 now that 1.18-alpha is out?).

Things were much better, but still not enough so after looking at a new flamegraph, we noticed that EnsureAddressBind was now responsible for 25-30% of kube-proxy cpu time. It tries to bind every ClusterIP to kube-ipvs0 during each which is not efficient (it generates too many netlink calls). I addressed this by retrieving all bound addresses only once and only doing the netlink calls when required (PR coming soon).

A few more info on kube-proxy performances in ipvs mode for people looking for info:
- https://github.com/kubernetes/kubernetes/pull/66149 helps a lot with SyncIPSetEntries (available since 1.13)
- on large clusters, increasing `ipvs-min-sync-period` helps decrease CPU usage a lot (we set it to 5s now)

Some optimizations we should look into in the future
- SyncIPSetEntries is really not optimal because it uses the ipset host binary and parses the result using regex. Hopefully, the netlink library we use will soon support ipset (https://github.com/vishvananda/netlink/pull/408). I'll reach to see how we can help
- GetVirtualServer and GetVirtualServers should be optimized. I think we could probably run GetVirtualServers once, store results into a map and modify GetVirtualServer to use this map (instead of doing a netlink call every time)
- memory allocation is probably not very efficient and could probably be improved a lot  

Thanks @lallydd for your analysis

**What you expected to happen**:
kube-proxy uses less CPU

**How to reproduce it (as minimally and precisely as possible)**:
Create a cluster with 2000+ clusterIP services and profile the kube-proxy process

**Flamegraph**:
kube-proxy (1.18-alpha) in ipvs mode with sync-period and min-sync-period set to 2s (to force frequent sync) on a cluster with 2000 services.
[...]

PR #92609, opened 2020-06-29
proxier/ipvs: check already binded addresses in the IPVS dummy interface

andrewsykim:
Signed-off-by: Andrew Sy Kim <kim.andrewsy@gmail.com>
Co-authored-by: Laurent Bernaille <laurent.bernaille@gmail.com>

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Improves IPVS proxy performance by only running `EnsureDummyInterface` if the virtual server address is not binded already.

**Which issue(s) this PR fixes**:
Replaces https://github.com/kubernetes/kubernetes/pull/88213

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Improves IPVS proxy performance by only running `EnsureDummyInterface` if the virtual server address is not binded already.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #88213, opened 2020-02-16
Improve SyncService performances

lbernail:
Partially addresses #88212 by making sure we only call EnsureBindAddress when needed (retrieve all addresses bound to the dummy interface only once during a sync).

I'm not sure this is the best approach, and I'm more than happy to modify the PR. The results are very promising.

Flamegraphs before and after patch (showing that EnsureBindAddress is almost gone)
[flamegraphs.tar.gz](https://github.com/kubernetes/kubernetes/files/4209850/flamegraphs.tar.gz)

kube-proxy cpu metric on a large cluster (2000 services with ClusterIP), showing a 30% perf improvement (from 15% of a core to 10% after the patch):
image

/sig network
/area ipvs
/kind feature
/assign @andrewsykim 

```release-note
kube-proxy: improve performances in IPVS mode on large clusters
```

PR #93445, opened 2020-07-25
Optimize ipset invoking in ipvs proxy

Lion-Wei:
**What type of PR is this?**
/kind bug

- Add `RestoreSets` and `SaveAllSets` interface for ipset.
- Use `ipset save/restore` instead `ipset create/add/del`, to reduce command call in `syncProxyRules`.
- Reduce ipset call to only 1 `ipset save` and 1 `ipset restore` call in each sync.
- Reorganized ipset related logic, add `ipsetManager` struct to manager ipsets entries cache, make more efficient and easier to read.

Related issue: #88212

Currently we use ipset host binary to sync ipset sets and entries, each ipset entry need an `ipset add` command call. 
So in bunch services or create/update/delete case, the `syncIPSetEntries`  cost can be very expensive.

In my test case, when 2500 service creating, `syncProxyRules` takes more then 15 seconds, and from glame graph we can see that `syncIPSetEntries` takes the biggest part. 

bunch-services-create

And here are the typical syncProxyRules take time data I got from ipvs log.

| test case | 2 services, resync | 2500 services creating | 2500 services resync | 2500 services deleting |
| --------- | ------------------ | ---------------------- | -------------------- | ---------------------- |
| before    | 70.790783ms        | 14.231572761s          | 924.981262ms         | 20.482706001s          |
| after     | 30.051448ms        | 2.794356405s           | 916.360592ms         | 8.438355251s           |

Obviously, this optimization can highly improve ipvs proxy mode performance, espically in bunch service create or delete case.

```release-note
NONE
```

### DIFF
--- a/pkg/proxy/ipvs/ipset.go
+++ b/pkg/proxy/ipvs/ipset.go
@@ -17,14 +17,15 @@ limitations under the License.
 package ipvs
 
 import (
-	"k8s.io/apimachinery/pkg/util/sets"
-	utilversion "k8s.io/apimachinery/pkg/util/version"
-	utilipset "k8s.io/kubernetes/pkg/util/ipset"
-
+	"bytes"
 	"fmt"
 	"strings"
 
 	"k8s.io/klog/v2"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	utilversion "k8s.io/apimachinery/pkg/util/version"
+	utilipset "k8s.io/kubernetes/pkg/util/ipset"
 )
 
 const (
@@ -75,6 +76,10 @@ const (
 
 	kubeNodePortLocalSetSCTPComment = "Kubernetes nodeport SCTP port with externalTrafficPolicy=local with type 'hash ip:port'"
 	kubeNodePortLocalSetSCTP        = "KUBE-NODE-PORT-LOCAL-SCTP-HASH"
+
+	addCommand    = "add"
+	deleteCommand = "del"
+	createCommand = "create"
 )
 
 // IPSetVersioner can query the current ipset version.
@@ -83,17 +88,38 @@ type IPSetVersioner interface {
 	GetVersion() (string, error)
 }
 
-// IPSet wraps util/ipset which is used by IPVS proxier.
-type IPSet struct {
-	utilipset.IPSet
-	// activeEntries is the current active entries of the ipset.
-	activeEntries sets.String
-	// handle is the util ipset interface handle.
-	handle utilipset.Interface
+type ipsetInfo struct {
+	name    string
+	setType utilipset.Type
+	comment string
 }
 
-// NewIPSet initialize a new IPSet struct
-func NewIPSet(handle utilipset.Interface, name string, setType utilipset.Type, isIPv6 bool, comment string) *IPSet {
+// ipsetManager wraps util/ipset which is used by IPVS proxier.
+type ipsetManager struct {
+	// setList a map of current sets list, with key of set name and value of set content
+	setList map[string]*utilipset.IPSet
+	// activeEntries is a map of current active entries, with key of set name and value of set entries
+	activeEntries map[string]sets.String
+	// handler is the util ipset interface handler.
+	handler utilipset.Interface
+}
+
+func newIPSetManager(isIPv6 bool, setInfoList []ipsetInfo, ipset utilipset.Interface) *ipsetManager {
+	m := &ipsetManager{
+		setList:       make(map[string]*utilipset.IPSet, len(setInfoList)),
+		activeEntries: make(map[string]sets.String, len(setInfoList)),
+		handler:       ipset,
+	}
+	for _, info := range setInfoList {
+		set := newIPSet(info.name, info.setType, isIPv6, info.comment)
+		m.setList[info.name] = set
+		m.activeEntries[info.name] = sets.NewString()
+	}
+	return m
+}
+
+// newIPSet initialize a new IPSet
+func newIPSet(name string, setType utilipset.Type, isIPv6 bool, comment string) *utilipset.IPSet {
 	hashFamily := utilipset.ProtocolFamilyIPV4
 	if isIPv6 {
 		hashFamily = utilipset.ProtocolFamilyIPV6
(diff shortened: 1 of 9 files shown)


Lion-Wei:
pull-kubernetes-integration

Lion-Wei:
Ready for another review.
/cc @andrewsykim , PTAL, thanks. : )

pkg/proxy/ipvs/ipset.go
```diff
@@ -112,76 +137,84 @@ func NewIPSet(handle utilipset.Interface, name string, setType utilipset.Type, i
...
+	}
+	return entry.Validate(s.setList[setName])
 }
 
-func (set *IPSet) syncIPSetEntries() {
-	appliedEntries, err := set.handle.ListEntries(set.Name)
-	if err != nil {
-		klog.Errorf("Failed to list ip set entries, error: %v", err)
-		return
+func (s *ipsetManager) insetIPSetEntry(setName string, entry *utilipset.Entry) (invalid bool) {
```
andrewsykim, line 171: named return is not needed here?

pkg/proxy/ipvs/proxier.go
```diff
@@ -1173,11 +1157,7 @@ func (proxier *Proxier) syncProxyRules() {
 				IP2:      epIP,
 				SetType:  utilipset.HashIPPortIP,
 			}
-			if valid := proxier.ipsetList[kubeLoopBackIPSet].validateEntry(entry); !valid {
-				klog.Errorf("%s", fmt.Sprintf(EntryInvalidErr, entry, proxier.ipsetList[kubeLoopBackIPSet].Name))
-				continue
-			}
-			proxier.ipsetList[kubeLoopBackIPSet].activeEntries.Insert(entry.String())
+			proxier.ipsetManager.insetIPSetEntry(kubeLoopBackIPSet, entry)
```
andrewsykim, line 1160: s/inset/insert

pkg/proxy/ipvs/proxier.go
```diff
@@ -1190,11 +1170,7 @@ func (proxier *Proxier) syncProxyRules() {
 		}
 		// add service Cluster IP:Port to kubeServiceAccess ip set for the purpose of solving hairpin.
 		// proxier.kubeServiceAccessSet.activeEntries.Insert(entry.String())
-		if valid := proxier.ipsetList[kubeClusterIPSet].validateEntry(entry); !valid {
-			klog.Errorf("%s", fmt.Sprintf(EntryInvalidErr, entry, proxier.ipsetList[kubeClusterIPSet].Name))
-			continue
-		}
-		proxier.ipsetList[kubeClusterIPSet].activeEntries.Insert(entry.String())
+		proxier.ipsetManager.insetIPSetEntry(kubeClusterIPSet, entry)
```
andrewsykim, line 1173: s/inset/insert

pkg/proxy/ipvs/proxier.go
```diff
@@ -1264,18 +1240,10 @@ func (proxier *Proxier) syncProxyRules() {
 			}
 
 			if utilfeature.DefaultFeatureGate.Enabled(features.ExternalPolicyForExternalIP) && svcInfo.OnlyNodeLocalEndpoints() {
-				if valid := proxier.ipsetList[kubeExternalIPLocalSet].validateEntry(entry); !valid {
-					klog.Errorf("%s", fmt.Sprintf(EntryInvalidErr, entry, proxier.ipsetList[kubeExternalIPLocalSet].Name))
-					continue
-				}
-				proxier.ipsetList[kubeExternalIPLocalSet].activeEntries.Insert(entry.String())
+				proxier.ipsetManager.insetIPSetEntry(kubeExternalIPLocalSet, entry)
```
andrewsykim, line 1243: s/inset/insert

pkg/proxy/ipvs/proxier.go
```diff
@@ -1264,18 +1240,10 @@ func (proxier *Proxier) syncProxyRules() {
...
 			} else {
 				// We have to SNAT packets to external IPs.
-				if valid := proxier.ipsetList[kubeExternalIPSet].validateEntry(entry); !valid {
-					klog.Errorf("%s", fmt.Sprintf(EntryInvalidErr, entry, proxier.ipsetList[kubeExternalIPSet].Name))
-					continue
-				}
-				proxier.ipsetList[kubeExternalIPSet].activeEntries.Insert(entry.String())
+				proxier.ipsetManager.insetIPSetEntry(kubeExternalIPSet, entry)
```
andrewsykim, line 1246: s/inset/insert

pkg/proxy/ipvs/proxier.go
```diff
@@ -1319,28 +1287,16 @@ func (proxier *Proxier) syncProxyRules() {
 				// proxier.kubeServiceAccessSet.activeEntries.Insert(entry.String())
 				// If we are proxying globally, we need to masquerade in case we cross nodes.
 				// If we are proxying only locally, we can retain the source IP.
-				if valid := proxier.ipsetList[kubeLoadBalancerSet].validateEntry(entry); !valid {
-					klog.Errorf("%s", fmt.Sprintf(EntryInvalidErr, entry, proxier.ipsetList[kubeLoadBalancerSet].Name))
-					continue
-				}
-				proxier.ipsetList[kubeLoadBalancerSet].activeEntries.Insert(entry.String())
+				proxier.ipsetManager.insetIPSetEntry(kubeLoadBalancerSet, entry)
```
andrewsykim, line 1290: s/inset/insert 

same comment rest of the file :)

pkg/proxy/ipvs/ipset.go
Lion-Wei: Actually we don't need return here 😅

Lion-Wei:
@andrewsykim , Thanks for reviewing, all comments are addressed, please another look when you got time. : )
@lbernail I think you should be interested in this, would you please also take a look, thanks.

pkg/proxy/ipvs/ipset.go
```diff
@@ -83,17 +88,35 @@ type IPSetVersioner interface {
...
-	handle utilipset.Interface
+type ipsetInfo struct {
+	name    string
+	setType utilipset.Type
+	comment string
+}
+
+// ipsetManager wraps util/ipset which is used by IPVS proxier.
+type ipsetManager struct {
+	setList       map[string]*utilipset.IPSet
```
andrewsykim, line 99: can we add comments to these fields?

pkg/proxy/ipvs/ipset.go
```diff
@@ -112,76 +135,84 @@ func NewIPSet(handle utilipset.Interface, name string, setType utilipset.Type, i
...
+	return entry.Validate(s.setList[setName])
 }
 
-func (set *IPSet) syncIPSetEntries() {
-	appliedEntries, err := set.handle.ListEntries(set.Name)
-	if err != nil {
-		klog.Errorf("Failed to list ip set entries, error: %v", err)
+func (s *ipsetManager) insertIPSetEntry(setName string, entry *utilipset.Entry) {
+	if !s.entryValid(setName, entry) {
+		klog.Errorf("%s", fmt.Sprintf(EntryInvalidErr, entry, setName))
```
andrewsykim, line 171: Instead of logging here, would prefer `insertIPSetEntry` return an error and log at proxier level, what do you think?

Lion-Wei: Normally this `Validate` should not return err, and if this do happened, I think this should not block the whole syncRule work, so I think logging here or at proxier level doesn't metter. Is there anything I haven't thought about?

pkg/proxy/ipvs/ipset.go
Lion-Wei: Sure, absolutely necessary.

pkg/proxy/ipvs/ipset.go
andrewsykim: Logging at the proxier level is better for visibility since you'll know at what point of `syncProxyRules` the error is happening. If you log at the insert level, all insert failures log at the same place.

Lion-Wei: Good point, thanks a lot. : )

lbernail:
I remember seeing this when I looked at flamegraphs in the latest optimization we did
This is great news
Can you edit the PR to explain the logic change (reduce calls to ipset by grouping changes if I understand the patch correctly)

Also I checked again and the netlink library still doesn't support ipsets (this would be a lot nicer...)

Lion-Wei:
> Can you edit the PR to explain the logic change (reduce calls to ipset by grouping changes if I understand the patch correctly)

Sure thing, I added another description in this pr.
```
Reduce ipset call to only 1 ipset save and 1 ipset restore call in each sync.
Reorganized ipset related logic, add ipsetManager struct to manager ipsets entries cache, make more efficient and easier to read.
```

> Also I checked again and the netlink library still doesn't support ipsets (this would be a lot nicer...)

Yes, I also checked it before start this pr.

Lion-Wei:
BTW, I noticed bind and unbind addresses from `kube-ipvs0` also takes lot time in one sync, and bind address to `kube-ipvs0` will result in many other issues. 

Recently I was trying to solve this problem, luckly already got some progress, I'll start an issue to discuss this.

lbernail:
Thanks!
Re: bind addresses: Are you running a version that includes this PR? https://github.com/kubernetes/kubernetes/pull/92609
(more context in the original PR: https://github.com/kubernetes/kubernetes/pull/88213)

Lion-Wei:
Hi, @andrewsykim , sorry for the delay response, I think all comments are addressed, please take another look when you got time, thanks. : )

andrewsykim:
Sorry for the delay, I'll review thi ssoon

pkg/util/ipset/testing/fake.go
```diff
@@ -140,4 +142,73 @@ func (f *FakeIPSet) ListSets() ([]string, error) {
...
+		case "create":
+			if len(strArr) < 3 {
+				return fmt.Errorf("restore sets err, err line: %q", line)
+			}
+			setName, set := strArr[1], strings.Join(strArr[1:], " ")
+			if f.Sets[setName] == "" {
+				f.Sets[setName] = set
+				f.Entries[setName] = sets.NewString()
+			} else {
+				if !ignoreExist {
```
JornShen, line 187: why not use else if?

pkg/util/ipset/testing/fake.go
```diff
@@ -140,4 +142,73 @@ func (f *FakeIPSet) ListSets() ([]string, error) {
...
+					return fmt.Errorf("restore sets err, set %q exist", setName)
+				}
+			}
+		case "add":
+			if len(strArr) != 3 {
+				return fmt.Errorf("restore sets err, err line: %q", line)
+			}
+			setName, entry := strArr[1], strArr[2]
+			if f.Entries[setName].Has(entry) {
+				if !ignoreExist {
```
JornShen, line 197: can here merge to the upper level if?

pkg/proxy/ipvs/ipset.go
```diff
@@ -199,3 +232,47 @@ func checkMinVersion(vstring string) bool {
...
+		if strings.HasPrefix(line, addCommand) {
+			entrySlices := strings.Split(line, " ")
+			if len(entrySlices) < 3 {
+				klog.Errorf("Read set entries, got invalid entry: %s", line)
+				continue
+			}
+			set, entry := entrySlices[1], entrySlices[2]
+			if setEntries[set] == nil {
+				setEntries[set] = sets.NewString(entry)
+			} else {
```
JornShen, line 254: Can here change to:
```
if setEntries[set] == nil {
    setEntries[set] = sets.NewString()
}
setEntries[set].Insert(entry)
```

zhifei92:
this PR still not merged

zhifei92:
@Lion-Wei hi，This PR looks good. Why didn't it go on?

Issue #92937, opened 2020-07-09
Very few PRs are merging


spiffxp:
I opted to open a new issue instead of re-opening https://github.com/kubernetes/kubernetes/issues/90647

The tide pool for kubernetes/kubernetes master is much larger than usual.  Put another way, a lot of PRs are eligible for merge, but they're not getting merged.  This is due to batches of PRs failing tide's retest-prior-to-merge step.

![Screen Shot 2020-07-09 at 8 32 47 AM](https://user-images.githubusercontent.com/49258/87059962-ca49c280-c1be-11ea-8209-25e343a58ffc.png)
(taken from: https://monitoring.prow.k8s.io/d/d69a91f76d8110d3e72885ee5ce8038e/tide-dashboard?orgId=1&from=now-30d&to=now)

Some of this is likely due to a rush to land PRs before today's Code Freeze deadline (ref: https://github.com/kubernetes/sig-release/tree/master/releases/release-1.19#tldr)

Some useful links for troubleshooting
- https://prow.k8s.io/tide-history?repo=kubernetes%2Fkubernetes&branch=master - what tide is doing (trigger / merge events)
- https://prow.k8s.io/?repo=kubernetes%2Fkubernetes&type=batch - results of jobs for batches of PRs triggered by tide

Some links for context / prior discussion
- [#testing-ops from yesterday onward](https://kubernetes.slack.com/archives/C7J9RP96G/p1594227886310500)
- [#sig-testing thread](https://kubernetes.slack.com/archives/C09QZ4DQB/p1594304440039600)

/priority critical-urgent
/sig release
/sig testing
/milestone v1.19

FYI @kubernetes/ci-signal @kubernetes/release-team

spiffxp:
For folks who are concerned their PRs aren't going to land prior to Code Freeze: historically, as long as a PR is eligible for merge prior to Code Freeze, nobody's going to block it or kick it out.

In the past few months when we've encountered stuck merges, one thing that's been done is to apply a /hold to all PRs to see if we can force smaller batches / identify "bad" PRs.  In the event that we need to do this again, it would help if all PR's that are eligible for merge are in the v1.19 milestone.

[...]

Issue #87807, opened 2020-02-04
[Flaky Test] verify-master (ci-kubernetes-verify-master)


droslean:
**Which jobs are failing**:
`verify-master (ci-kubernetes-verify-master) `

**Since when has it been failing**:
`4th Feb 19:53 PST`

**Testgrid link**:
https://testgrid.k8s.io/sig-release-master-blocking#verify-master

**Reason for failure**:
```console
 I0204 09:12:04.373746    2848 parse.go:432] type checker: /usr/local/go/src/runtime/panic.go:292:34: cannot use (_defer literal) (value of type _defer) as unsafe.ArbitraryType value in argument to unsafe.Sizeof
I0204 09:12:04.373793    2848 parse.go:432] type checker: /usr/local/go/src/runtime/panic.go:292:20: unsafe.Sizeof((_defer literal)) (value of type uintptr) is not constant
I0204 09:12:04.381637    2848 parse.go:432] type checker: /usr/local/go/src/runtime/sema.go:51:45: cannot use (semaRoot literal) (value of type semaRoot) as unsafe.ArbitraryType value in argument to unsafe.Sizeof
I0204 09:12:04.381691    2848 parse.go:432] type checker: /usr/local/go/src/runtime/sema.go:51:8: array length cpu.CacheLinePadSize - unsafe.Sizeof((semaRoot literal)) (value of type uintptr) must be constant
I0204 09:12:04.391416    2848 parse.go:432] type checker: /usr/local/go/src/runtime/time.go:53:44: cannot use (timersBucket literal) (value of type timersBucket) as unsafe.ArbitraryType value in argument to unsafe.Sizeof
I0204 09:12:04.391556    2848 parse.go:432] type checker: /usr/local/go/src/runtime/time.go:53:7: array length cpu.CacheLinePadSize - unsafe.Sizeof((timersBucket literal)) % cpu.CacheLinePadSize (value of type uintptr) must be constant
I0204 09:12:04.391932    2848 parse.go:432] type checker: /usr/local/go/src/runtime/trace.go:159:30: cannot use (traceBufHeader literal) (value of type traceBufHeader) as unsafe.ArbitraryType value in argument to unsafe.Sizeof 
```

**Anything else we need to know**:
/cc @kubernetes/ci-signal 
/cc @kubernetes/release-managers 

/sig testing
/priority critical-urgent
/milestone v1.18

liggitt:
[...]

PR #93448, opened 2020-07-25
Add dependencycheck tool to address long running no-vendor-cycles test

hasheddan:
**What type of PR is this?**

/kind cleanup
> /kind flake

Adds a new tool, `dependencycheck`, to detect cyclical dependencies on main and staging repos in vendored packages. This is currently being handled by `hack/verify-no-vendor-cycles.sh`, which takes exceedingly long to run (3-4 minutes) for the task it is accomplishing. `dependencycheck` typically runs in a few seconds. In addition, when taking a look at the logs for `verify-master` and trying it out locally, I am not certain that this script is actually accomplishing its goal. For instance, if a package depended on `k8s.io/apimachinery/pkg/runtime`, it appears it is being printed as such, yet the script is checking for a match of `k8s.io/kubernetes/vendor/k8s.io` then `k8s.io/apimachinery`. I am guessing this may have to do with dependencies formerly being printed by their full path, but I am not sure. See more here: https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-verify-master/1287005337945116672/build-log.txt

Ref: https://github.com/kubernetes/kubernetes/issues/87807 https://github.com/kubernetes/kubernetes/issues/92937

I wasn't sure who to add to the `OWNERS` file here so I just picked some of the usual suspects.

/cc @BenTheElder @dims @liggitt 

```release-note
NONE
```

```docs

```

### DIFF
--- a/cmd/dependencycheck/dependencycheck.go
+++ b/cmd/dependencycheck/dependencycheck.go
@@ -1,0 +1,115 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Checks for restricted dependencies in go packages. Does not check transitive
+// dependencies implicitly, so they must be supplied in dependencies file if
+// they are to be evaluated.
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"regexp"
+)
+
+var (
+	exclude  = flag.String("exclude", "", "skip packages regex pattern (e.g. '^k8s.io/kubernetes/')")
+	restrict = flag.String("restrict", "", "restricted dependencies regex pattern (e.g. '^k8s.io/(apimachinery|client-go)/')")
+)
+
+type goPackage struct {
+	Name         string
+	ImportPath   string
+	Imports      []string
+	TestImports  []string
+	XTestImports []string
+}
+
+func main() {
+	flag.Parse()
+
+	args := flag.Args()
+
+	if len(args) != 1 {
+		log.Fatalf("usage: dependencycheck <json-dep-file> (e.g. 'go list -mod=vendor -test -deps -json ./vendor/...')")
+	}
+	if *restrict == "" {
+		log.Fatalf("Must specify restricted regex pattern")
+	}
+	depsPattern, err := regexp.Compile(*restrict)
+	if err != nil {
+		log.Fatalf("Error compiling restricted dependencies regex: %v", err)
+	}
+	var excludePattern *regexp.Regexp
+	if *exclude != "" {
+		excludePattern, err = regexp.Compile(*exclude)
+		if err != nil {
+			log.Fatalf("Error compiling excluded package regex: %v", err)
+		}
+	}
+	b, err := ioutil.ReadFile(args[0])
+	if err != nil {
+		log.Fatalf("Error reading dependencies file: %v", err)
+	}
+
+	packages := []goPackage{}
+	decoder := json.NewDecoder(bytes.NewBuffer(b))
+	for {
+		pkg := goPackage{}
+		if err := decoder.Decode(&pkg); err != nil {
+			if err == io.EOF {
+				break
+			}
+			log.Fatalf("Error unmarshaling dependencies file: %v", err)
+		}
+		packages = append(packages, pkg)
+	}
+
+	violations := map[string][]string{}
+	for _, p := range packages {
+		if excludePattern != nil && excludePattern.MatchString(p.ImportPath) {
+			continue
+		}
+		importViolations := []string{}
+		allImports := []string{}
+		allImports = append(allImports, p.Imports...)
+		allImports = append(allImports, p.TestImports...)
+		allImports = append(allImports, p.XTestImports...)
+		for _, i := range allImports {
+			if depsPattern.MatchString(i) {
+				importViolations = append(importViolations, i)
+			}
+		}
+		if len(importViolations) > 0 {
+			violations[p.ImportPath] = importViolations
+		}
+	}
+
+	if len(violations) > 0 {
+		for k, v := range violations {
+			fmt.Printf("Found dependency violations in package %s:\n", k)
+			for _, a := range v {
+				fmt.Println("--> " + a)
+			}
+		}
+		log.Fatal("Found restricted dependency violations in packages")
+	}
+}
--- a/hack/verify-no-vendor-cycles.sh
+++ b/hack/verify-no-vendor-cycles.sh
@@ -21,7 +21,6 @@
 set -o errexit
 set -o nounset
 set -o pipefail
-set -x;
 
 KUBE_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 source "${KUBE_ROOT}/hack/lib/init.sh"
@@ -34,21 +33,24 @@ staging_repos_pattern=$(IFS="|"; echo "${staging_repos[*]}")
 
 cd "${KUBE_ROOT}"
 
-failed=false
-while IFS= read -r dir; do
-  deps=$(go list -f '{{range .Deps}}{{.}}{{"\n"}}{{end}}' "${dir}" 2> /dev/null || echo "")
-  deps_on_main=$(echo "${deps}" | grep -v "k8s.io/kubernetes/vendor/" | grep "k8s.io/kubernetes" || echo "")
-  if [ -n "${deps_on_main}" ]; then
-    echo "Package ${dir} has a cyclic dependency on the main repository."
-    failed=true
-  fi
-  deps_on_staging=$(echo "${deps}" | grep "k8s.io/kubernetes/vendor/k8s.io" | grep -E "k8s.io\/${staging_repos_pattern}\>" || echo "")
-  if [ -n "${deps_on_staging}" ]; then
-    echo "Package ${dir} has a cyclic dependency on staging repository packages: ${deps_on_staging}"
-    failed=true
-  fi
-done < <(find ./vendor -type d)
-
-if [[ "${failed}" == "true" ]]; then
+# Check for any module that is not main or staging and depends on main or staging
+bad_deps=$(go mod graph | grep -vE "^k8s.io\/(kubernetes|${staging_repos_pattern})" | grep -E "\sk8s.io\/(kubernetes|${staging_repos_pattern})" || true)
+if [[ -n "${bad_deps}" ]]; then
+  echo "Found disallowed dependencies that transitively depend on k8s.io/kubernetes or staging modules:"
+  echo "${bad_deps}"
   exit 1
 fi
+
+kube::util::ensure-temp-dir
+
+# Get vendored packages dependencies
+# Use -deps flag to include transitive dependencies
+go list -mod=vendor -test -deps -json ./vendor/... > "${KUBE_TEMP}/deps.json"
+
+# Check for any vendored package that imports main repo
+# Staging repos are explicitly excluded even though go list does not currently consider symlinks
+go run cmd/dependencycheck/dependencycheck.go -restrict "^k8s\.io/kubernetes/" -exclude "^k8s\.io/(${staging_repos_pattern})(/|$)" "${KUBE_TEMP}/deps.json"
+
+# Check for any vendored package that imports a staging repo
+# Staging repos are explicitly excluded even though go list does not currently consider symlinks
+go run cmd/dependencycheck/dependencycheck.go -restrict "^k8s\.io/(${staging_repos_pattern})(/|$)" -exclude "^k8s\.io/(${staging_repos_pattern})(/|$)" "${KUBE_TEMP}/deps.json"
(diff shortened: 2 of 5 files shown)


hasheddan:
/retest
`verify.no-vendor-cycles` did not get run that time because of the spike in other tests

cmd/vendorcycle/vendorcycle.go
```diff
@@ -0,0 +1,95 @@
...
+		log.Fatalf("Error compiling excluded package regex: %v", err)
+	}
+
+	violations := map[string][]string{}
+	err = filepath.Walk(args[0], func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.IsDir() {
+			p, err := build.Default.Import(fmt.Sprintf("./%s", path), ".", 0)
```
BenTheElder, line 58: fun thought: build tags?
I _think_ we're fine, the default tags should get setup in the script below, and non-default build tags shouldn't be a problem, probably?

hasheddan: That is a valid point. I'm not sure how often we will hit it in practice, but I think regardless it might be better to refactor this to use https://godoc.org/golang.org/x/tools/go/packages

liggitt:
Thanks for jumping on this, much appreciated. I need to double check whether the current impl detects module level transitive dependencies on k8s.io/kubernetes or staging repos even if the particular packages we use don't depend back on these repos. I think the source scanning technique would miss those. If we currently catch/prevent those, we shouldn't regress that. I'll queue this up to look at early next week.

cmd/vendorcycle/vendorcycle.go
hasheddan: Interesting observation, using `packages`, even with only `NeedImports` mode set is _significantly_ slower, which makes me think that it is actually pulling the modules rather than inspecting the source (this may have already been known by others :sweat_smile:). If we can afford to only inspect the source rather than loading the packages then that would certainly be advantageous as demonstrated by the 2 second `verify.no-vendor-cycles` run on the successful `verify-master` here.

hasheddan:
@liggitt thanks! I _think_ that `go list` will only check package level transitive dependencies, but I'll test and make sure. Regardless of whether it does, I think an adjustment is needed here to make this check "correct" even if we are unable to improve its latency for the time-being.

dims:
/uncc

glad to see @liggitt on this!

liggitt:
> I need to double check whether the current impl detects module level transitive dependencies on k8s.io/kubernetes or staging repos even if the particular packages we use don't depend back on these repos. I think the source scanning technique would miss those. If we currently catch/prevent those, we shouldn't regress that. I'll queue this up to look at early next week.

looks like we only checked packages found in vendor previously

if possible, I'd like to lean on `go list` to do the dependency resolution, with something like this:
```
go list -mod=vendor -test -deps -json ./vendor/... > deps.json
```

A go program to process that deps.json file seems better than bash

I would envision something like:
1. Determine the list of staging repos
2. Run `go list` to capture vendor dependencies
3. Run `dependencycheck deps.json k8s.io/kubernetes` to fail if any package in deps.json depends on k8s.io/kubernetes/...
4. Run `dependencycheck deps.json k8s.io/client-go k8s.io/api k8s.io/apimachinery ...` to fail if any package in deps.json depends on staging modules

As a follow-up, for the module-level cycle checks, we can separately use `go mod graph` to detect/protect against this, forbidding things that match `go mod graph | grep " k8s.io/kubernetes"`, etc

hasheddan:
@liggitt do we need to explicitly run 3 & 4 separately? This check will execute quickly, so it is not a huge deal, but we are iterating twice through when it seems like one iteration checking for both would be faster.

liggitt:
it's a little easier to treat staging repos and k8s.io/kubernetes separately, but there's not a strong reason to. I expect the tool will take less than a second to run so I doubt calling it twice would matter

RobertKielty:
@hasheddan @liggitt  What's the next step here? 

I'm looking at verify-master failures and I just want to understand if there more changes needed on this PR?

hasheddan:
@RobertKielty I need to refactor as has been suggested. I should have this up today. However, this is not currently causing the `verify-master` failures. That is due to pinning resource requests and limits, then typecheck exceeding the memory limit. Opening an issue now with artifacts.

RobertKielty:
Thanks @hasheddan

hasheddan:
@liggitt I have updated and this is ready for another review :+1: I had one question about the module-level dependency checks. If we are checking at the module level (which I have not added yet), do we also need to be checking at the package level (i.e. aren't the module-level dependencies a superset of the packages)?

liggitt:
> If we are checking at the module level (which I have not added yet), do we also need to be checking at the package level (i.e. aren't the module-level dependencies a superset of the packages)?

No. Not all our dependencies declare go.mod files, which means their transitive dependencies are not listed in `go mod graph` and we have to drop down to package level traversals.

hasheddan:
@liggitt understood! Thanks for the clarification 👍

hack/verify-no-vendor-cycles.sh
```diff
@@ -34,21 +34,8 @@ staging_repos_pattern=$(IFS="|"; echo "${staging_repos[*]}")
 
 cd "${KUBE_ROOT}"
```
liggitt, line 36: this will create and clean up a temp dir for you:
```
kube::util::ensure-temp-dir()
go list -mod=vendor -test -deps -json ./vendor/... > "${KUBE_TEMP}/deps.json"
```

hasheddan:
@liggitt updated with module checks

cmd/dependencycheck/dependencycheck.go
```diff
@@ -0,0 +1,96 @@
...
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"regexp"
+)
+
+var (
+	exclude  = flag.String("exclude", `a^`, "skip packages pattern")
```
BenTheElder, line 31: we should probably specify that this is a regex. does this default make sense?

cmd/dependencycheck/dependencycheck.go
```diff
@@ -0,0 +1,96 @@
...
+	restrict = flag.String("restrict", `a^`, "restricted dependencies pattern")
+)
+
+type goPackage struct {
+	Name       string   `json:",omitempty"`
+	ImportPath string   `json:",omitempty"`
+	Deps       []string `json:",omitempty"`
+}
+
+func jsonObjsToArray(raw []byte) []byte {
```
BenTheElder, line 41: might deserve a comment

cmd/dependencycheck/dependencycheck.go
hasheddan: Agreed that we should specify. The default here should never match.

hasheddan: However, we could just set it to empty string and if exclude is empty string we can return immediately (no reason to check if we don't care what is matched)

liggitt: yeah, leave empty and ignore if unset

cmd/dependencycheck/dependencycheck.go
```diff
@@ -0,0 +1,96 @@
...
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"regexp"
+)
+
+var (
+	exclude  = flag.String("exclude", `a^`, "skip packages pattern")
+	restrict = flag.String("restrict", `a^`, "restricted dependencies pattern")
```
liggitt, line 32: leave empty and require to be set

cmd/dependencycheck/dependencycheck.go
```diff
@@ -0,0 +1,96 @@
...
+	b := append([]byte("["), raw...)
+	return append(b, ']')
+}
+
+func main() {
+	flag.Parse()
+
+	args := flag.Args()
+
+	if len(args) == 0 {
```
liggitt, line 52:
```suggestion
	if len(args) != 1 {
```

cmd/dependencycheck/dependencycheck.go
```diff
@@ -0,0 +1,96 @@
...
+		log.Fatalf("Error reading dependencies file: %v", err)
+	}
+	packages := []goPackage{}
+	if err := json.Unmarshal(jsonObjsToArray(b), &packages); err != nil {
+		log.Fatalf("Error unmarshaling dependencies file: %v", err)
+	}
+
+	violations := map[string][]string{}
+	for _, p := range packages {
+		if !excludePattern.MatchString(p.ImportPath) {
```
liggitt, line 74: nit: continue early and unnest 

```go
if excludePattern != nil && excludePattern.MatchString(p.ImportPath) {
  continue
}
importViolations := []string{}
for _, d := range p.Deps {
  ...
```

hasheddan:
@liggitt I believe all comments have been addressed and this is ready for another pass :+1:

liggitt:
> no-vendor-cycles  6s

down from 5m18s in a recent run 🎉

cmd/dependencycheck/dependencycheck.go
```diff
@@ -0,0 +1,108 @@
...
+	Deps       []string
+}
+
+func main() {
+	flag.Parse()
+
+	args := flag.Args()
+
+	if len(args) != 1 {
+		log.Fatalf("usage: dependencycheck <json-dep-file>")
```
liggitt, line 48: nit, give an example of how to generate this file (e.g. `go list -mod=vendor -test -deps -json ./vendor/...`)

cmd/dependencycheck/dependencycheck.go
```diff
@@ -0,0 +1,108 @@
...
+		log.Fatalf("Error compiling restricted dependencies regex: %v", err)
+	}
+	var excludePattern *regexp.Regexp
+	if *exclude != "" {
+		excludePattern, err = regexp.Compile(*exclude)
+		if err != nil {
+			log.Fatalf("Error compiling excluded package regex: %v", err)
+		}
+	}
+	fmt.Println(excludePattern)
```
liggitt, line 64: drop debug logging?

hack/verify-no-vendor-cycles.sh
```diff
@@ -34,21 +34,16 @@ staging_repos_pattern=$(IFS="|"; echo "${staging_repos[*]}")
...
-    echo "Package ${dir} has a cyclic dependency on staging repository packages: ${deps_on_staging}"
-    failed=true
-  fi
-done < <(find ./vendor -type d)
-
-if [[ "${failed}" == "true" ]]; then
-  exit 1
-fi
+# Check for any module that is not main or staging and depends on main or staging
+go mod graph | grep -vE "^k8s.io\/(kubernetes|${staging_repos_pattern})" | grep -E "\sk8s.io\/(kubernetes|${staging_repos_pattern})" && exit 1
```
liggitt, line 38: this will not output a helpful error if it fails

suggest something like 

```sh
bad_deps=$(go mod graph | grep -vE "^k8s.io\/(kubernetes|${staging_repos_pattern})" | grep -E "\sk8s.io\/(kubernetes|${staging_repos_pattern})" || true)
if [[ -n "${bad_deps}" ]]; then
  echo "Found disallowed dependencies that transitively depend on k8s.io/kubernetes or staging modules:"
  echo "${bad_deps}"
  exit 1
fi
```

hack/verify-no-vendor-cycles.sh
```diff
@@ -34,21 +34,16 @@ staging_repos_pattern=$(IFS="|"; echo "${staging_repos[*]}")
```
liggitt, line 34: also suggest dropping `set -x;` it's almost completely noise

hack/verify-no-vendor-cycles.sh
```diff
@@ -34,21 +34,16 @@ staging_repos_pattern=$(IFS="|"; echo "${staging_repos[*]}")
...
+kube::util::ensure-temp-dir
+
+# Get vendored packages dependencies
+go list -mod=vendor -test -deps -json ./vendor/... > "${KUBE_TEMP}/deps.json"
+
+# Check for any vendored package that depends on main
+go run cmd/dependencycheck/dependencycheck.go -restrict "^k8s\.io/kubernetes/" "${KUBE_TEMP}/deps.json"
+
+# Check for any vendored package that depends on staging
+go run cmd/dependencycheck/dependencycheck.go -restrict "^k8s\.io/(${staging_repos_pattern})(/|$)" "${KUBE_TEMP}/deps.json"
```
liggitt, line 49: I think both of these need to exclude staging repos from the checks:

`-exclude "^k8s\.io/(${staging_repos_pattern})(/|$)"`

cmd/dependencycheck/dependencycheck.go
```diff
@@ -0,0 +1,108 @@
...
+
+var (
+	exclude  = flag.String("exclude", "", "skip packages regex pattern (e.g. '^k8s.io/kubernetes/')")
+	restrict = flag.String("restrict", "", "restricted dependencies regex pattern (e.g. '^k8s.io/(apimachinery|client-go)/')")
+)
+
+type goPackage struct {
+	Name       string
+	ImportPath string
+	Deps       []string
```
liggitt, line 39: Deps includes all transitive dependencies of this package, which is likely to be enormous

Imports + TestImports + XTestImports is likely to be more useful in identifying the actual problematic import that needs to be dealt with

suggest:
```diff
--- a/cmd/dependencycheck/dependencycheck.go
+++ b/cmd/dependencycheck/dependencycheck.go
@@ -34,9 +34,11 @@ var (
 )
 
 type goPackage struct {
-       Name       string
-       ImportPath string
-       Deps       []string
+       Name         string
+       ImportPath   string
+       Imports      []string
+       TestImports  []string
+       XTestImports []string
 }
 
 func main() {
@@ -86,7 +88,12 @@ func main() {
                        continue
                }
                importViolations := []string{}
-               for _, d := range p.Deps {
+
+               allImports := []string{}
+               allImports = append(allImports, p.Imports...)
+               allImports = append(allImports, p.TestImports...)
+               allImports = append(allImports, p.XTestImports...)
+               for _, d := range allImports {
                        if depsPattern.MatchString(d) {
                                importViolations = append(importViolations, d)
                        }
```

justaugustus:
Very cool, Dan!

hasheddan:
@liggitt updates complete :+1: However, this is going to fail on `pull-kubernetes-verify` due to https://github.com/kubernetes/kubernetes/issues/93623

BenTheElder:
thanks @hasheddan @liggitt

PR #93455, opened 2020-07-26
Run etcd as non root.

vinayakankugoyal:
**What type of PR is this?**
/kind feature

This PR allows us to run etcd as non root.

N/A

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster/gce/manifests/etcd.manifest
+++ b/cluster/gce/manifests/etcd.manifest
@@ -17,6 +17,16 @@
 "containers":[
     {
     "name": "etcd-container",
+    "securityContext": {
+      "runAsUser": {{runAsUser}},
+      "runAsGroup": {{runAsGroup}},
+      "allowPrivilegeEscalation": false,
+      "capabilities": {
+        "drop": [
+	  "all"
+	]
+      }
+    },
     "image": "{{ pillar.get('etcd_docker_repository', 'k8s.gcr.io/etcd') }}:{{ pillar.get('etcd_docker_tag', '3.4.13-0') }}",
     "resources": {
       "requests": {
--- a/cluster/gce/gci/configure-helper.sh
+++ b/cluster/gce/gci/configure-helper.sh
@@ -1769,6 +1769,8 @@ function prepare-etcd-manifest {
   fi
   # Replace the volume host path.
   sed -i -e "s@/mnt/master-pd/var/etcd@/mnt/disks/master-pd/var/etcd@g" "${temp_file}"
+  sed -i -e "s@{{runAsUser}}@${ETCD_RUNASUSER:-2030}@g" "${temp_file}"
+  sed -i -e "s@{{runAsGroup}}@${ETCD_RUNASGROUP:-2030}@g" "${temp_file}"
   mv "${temp_file}" /etc/kubernetes/manifests
 }
 
@@ -1789,10 +1791,10 @@ function start-etcd-servers {
   if [[ -e /etc/init.d/etcd ]]; then
     rm -f /etc/init.d/etcd
   fi
-  prepare-log-file /var/log/etcd.log
+  prepare-log-file /var/log/etcd.log "${ETCD_RUNASUSER:-2030}" "${ETCD_RUNASGROUP:-2030}"
   prepare-etcd-manifest "" "2379" "2380" "200m" "etcd.manifest"
 
-  prepare-log-file /var/log/etcd-events.log
+  prepare-log-file /var/log/etcd-events.log "${ETCD_RUNASUSER:-2030}" "${ETCD_RUNASGROUP:-2030}"
   prepare-etcd-manifest "-events" "4002" "2381" "100m" "etcd-events.manifest"
 }
 
--- a/cluster/gce/gci/master.yaml
+++ b/cluster/gce/gci/master.yaml
@@ -2,6 +2,7 @@
 
 users:
 - name: etcd
+  uid: 2030
   homedir: /var/etcd
   lock_passwd: true
   ssh_redirect_user: true
--- a/test/OWNERS
+++ b/test/OWNERS
@@ -1,7 +1,6 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
 reviewers:
-  - aojea
   - neolit123
   - johnSchnake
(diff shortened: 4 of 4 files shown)


immutableT (approved):

vinayakankugoyal:
kubernetes/release-managers can this change be added to the v1.19 milestone?

dims:
@vinayakankugoyal it's way tooooooo late to get this into 1.19.

liggitt:
> @vinayakankugoyal it's way tooooooo late to get this into 1.19.

agree. I also have questions about what the change in uid does with file permissions for existing data on upgrade. this does not seem like a safe change to merge without working upgrade tests.

/hold

dims:
@vinayakankugoyal some shell check issues as well https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93455/pull-kubernetes-verify/1300944761179344899

vinayakankugoyal:
> > @vinayakankugoyal it's way tooooooo late to get this into 1.19.
> 
> agree. I also have questions about what the change in uid does with file permissions for existing data on upgrade. this does not seem like a safe change to merge without working upgrade tests.
> 
> /hold

@liggitt - All the prow tests are passing. Are there a specific set of tests that you want me to run? Happy to do the additional validation. Thanks!

cluster/gce/manifests/etcd.manifest
```diff
@@ -59,7 +54,62 @@
...
+      },
+      { "name": "etc",
+        "mountPath": "/etc/srv/kubernetes",
+        "readOnly": false
+      }
+    ]
+    },
+    {
+    "name": "etcd{{ suffix }}-init-container-2",
+    "image": "k8s.gcr.io/debian-base:v2.0.0",
```
liggitt, line 75: I didn't expect a hard-coded image repository here

cluster/gce/manifests/etcd.manifest
```diff
@@ -59,7 +54,62 @@
...
+    "command": [
+      "/bin/sh",
+      "-c",
+      "chown -R {{runAsUser}}:{{runAsGroup}} /var/etcd/"
```
liggitt, line 79: if this fails, what happens to the pod? does it retry? permanently fail?

cluster/gce/manifests/etcd.manifest
```diff
@@ -14,20 +14,15 @@
...
-      }
-    },
     "command": [
-              "/bin/sh",
-              "-c",
-              "set -o errexit; if [ -e /usr/local/bin/migrate-if-needed.sh ]; then /usr/local/bin/migrate-if-needed.sh 1>>/var/log/etcd{{ suffix }}.log 2>&1; fi; exec /usr/local/bin/etcd --name etcd-{{ hostname }} --listen-peer-urls {{ etcd_protocol }}://{{ host_ip }}:{{ server_port }} --initial-advertise-peer-urls {{ etcd_protocol }}://{{ hostname }}:{{ server_port }} --advertise-client-urls {{ etcd_apiserver_protocol }}://127.0.0.1:{{ port }} --listen-client-urls {{ etcd_apiserver_protocol }}://{{ listen_client_ip }}:{{ port }} {{ quota_bytes }} --data-dir /var/etcd/data{{ suffix }} --initial-cluster-state {{ cluster_state }} --initial-cluster {{ etcd_cluster }} {{ etcd_creds }} {{ etcd_apiserver_creds }} {{ etcd_extra_args }} 1>>/var/log/etcd{{ suffix }}.log 2>&1"
-            ],
+        "/bin/sh",
+	"-c",
+        "set -o errexit; if [ -e /usr/local/bin/migrate-if-needed.sh ]; then /usr/local/bin/migrate-if-needed.sh 1>>/var/log/etcd{{ suffix }}.log 2>&1; fi;"
```
liggitt, line 24: same question about what happens if this fails - retries? terminates?

liggitt:
> @liggitt - All the prow tests are passing. Are there a specific set of tests that you want me to run? Happy to do the additional validation. Thanks!

Do we have tests that exercise running this with the previous configuration on persistent data, then upgrading to this new configuration?

cluster/gce/manifests/etcd.manifest
vinayakankugoyal: It will retry. See [this](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#understanding-init-containers) for details.

cluster/gce/manifests/etcd.manifest
vinayakankugoyal: It will retry. See [this](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#understanding-init-containers) for details.

vinayakankugoyal:
> > @liggitt - All the prow tests are passing. Are there a specific set of tests that you want me to run? Happy to do the additional validation. Thanks!
> 
> Do we have tests that exercise running this with the previous configuration on persistent data, then upgrading to this new configuration?

[These](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/cloud/gcp/cluster_upgrade.go) tests should be doing that.

liggitt:
> [These](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/cloud/gcp/cluster_upgrade.go) tests should be doing that.

Those tests have not run since 1.18 because of https://github.com/kubernetes/release/pull/868#issuecomment-673727541

If they were running and passing, that would cover the scenario I was asking about

liggitt:
I see automated upgrade tests as essential here

vinayakankugoyal:
> I see automated upgrade tests as essential here

That sounds reasonable. Manual tests can only catch so much. So is all development blocked because of upgrade tests not running in gce? I can't imagine other changes being approved as they might break upgrade? How do we prioritize getting that issue fixed? Is there an issue for that?

liggitt:
> How do we prioritize getting that issue fixed?

I know this is on sig-testing's radar (cc @BenTheElder)

> Is there an issue for that?

opened https://github.com/kubernetes/kubernetes/issues/94487 to track/coordinate

BenTheElder:
I was out for a bit and wrapping up some other things but upgrade is indeed on my radar.

PR #83273, opened 2019-09-28
Do not create StatefulSet pods when PVC is being deleted

ymmt2005:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

Pod with PVC will not be scheduled if the PVC is being deleted.
This can happen when the PVC has finalizers of storage plugins.

Such a pod becomes pending.  Unfortunately, after the finalizer
finishes and PVC is deleted, the pod remains pending forever.
The StatefulSet controller does nothing for this pending pod.

This commit prevents the StatefulSet controller from creating
such pods when PVC is to be deleted.

Reprocedure:

1. Create a single node cluster with [kind](https://github.com/kubernetes-sigs/kind).
2. Create a StatefulSet with the following manifests:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: test
  namespace: default
spec:
  clusterIP: None
  selector:
    app: test
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: test
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test
  serviceName: test
  template:
    metadata:
      labels:
        app: test
    spec:
      containers:
      - command:
        - pause
        image: quay.io/cybozu/ubuntu:18.04
        name: ubuntu
        volumeMounts:
        - mountPath: /usr/share/nginx/html
          name: www
  volumeClaimTemplates:
  - metadata:
      name: www
    spec:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 1Gi
      volumeMode: Filesystem
```

3. Once the pod gets running, edit PVC to add a finalizer `aaa.bbb/ccc`.
4. Run `kubectl delete --wait=false pvc/www-test-0`.
5. Run `kubectl delete --wait=false pods/test-0`.
6. See a new pod is created and become pending.
7. Edit the PVC to remove the `aaa.bbb/ccc` finalizer.
8. See the PVC is get deleted.
9. See the Pod remains pending forever.

**Does this PR introduce a user-facing change?**:
```release-note
Adding fix to statefulset controller to wait for pvc deletion before creating pods.
```

PR #79164, opened 2019-06-19
Bugfix:statefulSet would not recreate the pvc when a pvc is deleted before a pod

cwdsuzhou:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

Currently, statefulSet would not recreate the pvc when a pvc is deleted before a pod, which would lead to pod in pending status.

**Which issue(s) this PR fixes**:

Fixes #74374

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Bugfix: statefulSet would not recreate the pvc when a pvc is deleted before a pod 
```

Issue #74374, opened 2019-02-21
Storage protection feature does not integrate well with StatefulSet pod recreation


msau42:
**What happened**:
The storage protection feature prevents PVC deletion until all Pods referencing it are deleted.  This doesn't work well with Statefulsets when you want to delete the PVC and have the StatefulSet controller recreate it.  The StatefulSet controller only [creates missing PVCs when creating a new Pod](https://sourcegraph.com/github.com/kubernetes/kubernetes/-/blob/pkg/controller/statefulset/stateful_pod_control.go#L75).  So you could have a sequence like this:

1. Try to delete PVC. Deletion is pending due to the storage protection finalizer.
2. Delete the Pod.
3. StatefulSet controller notices that the pod is deleted and creates the replacement pod.  But the PVC already exists, so a new one is not created.
4. Storage protection controller removes the finalizer and the PVC is deleted
5. New pod is pending. First because PVC deletion is pending, and then 2nd because PVC is deleted.

Nothing tries to recreate the PVC again and the Pod is stuck in pending forever.

**What you expected to happen**:
We could potentially add logic in the statefulset controller to check if the PVC is terminating [here](https://sourcegraph.com/github.com/kubernetes/kubernetes/-/blob/pkg/controller/statefulset/stateful_pod_control.go#L182).  However, it uses an informer so the race may still exist.

Another option is to make the StatefulSet controller actively reconcile PVC creation. This would probably require more substantial changes to the controller.

@kubernetes/sig-storage-bugs 
@kubernetes/sig-apps-bugs

msau42:
A workaround in the meantime is to delete the StatefulSet pod a second time if it's stuck in pending due to missing PVC.

cscetbon:
Hey @msau42 thank you for having opened the issue and the temporary workaround

any news on that bug ?

KevinTHU:
I think this need to be fixed, but not just workaround. a second delete of pod is not a graceful operation.

msau42:
I would be happy to help guide someone to come up with a possible design and implementation for a fix.

/help

zhangxiaoyu-zidif:
when recreate pod, check if the PVC was label with deletion. maybe we could delete the PVC immediately by force  and create a new PVC before Pod creation.

mucahitkurt:
Hey @msau42, 

[...]

PR #93457, opened 2020-07-27
Do not create StatefulSet pods when PVC is being deleted

ymmt2005:
This reopens #79164 and #83273 and similar PRs, as they have been ignored or closed w/o merging.
I have experienced the problem multiple times in our production clusters, and want to fix it.

**What type of PR is this?**

/kind bug

Pod with PVC will not be scheduled if the PVC is being deleted.
This can happen when the PVC has finalizers of storage plugins.

Such a pod becomes pending.  Unfortunately, after the finalizer
finishes and PVC is deleted, the pod remains pending forever.
The StatefulSet controller does nothing for this pending pod.

This commit prevents the StatefulSet controller from creating
such pods when PVC is to be deleted.

Reprocedure:

1. Create a single node cluster with [kind](https://github.com/kubernetes-sigs/kind).
2. Create a StatefulSet with the following manifests:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: test
  namespace: default
spec:
  clusterIP: None
  selector:
    app: test
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: test
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test
  serviceName: test
  template:
    metadata:
      labels:
        app: test
    spec:
      containers:
      - command:
        - pause
        image: quay.io/cybozu/ubuntu:18.04
        name: ubuntu
        volumeMounts:
        - mountPath: /usr/share/nginx/html
          name: www
  volumeClaimTemplates:
  - metadata:
      name: www
    spec:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 1Gi
      volumeMode: Filesystem
```

3. Once the pod gets running, edit PVC to add a finalizer `aaa.bbb/ccc`.
4. Run `kubectl delete --wait=false pvc/www-test-0`.
5. Run `kubectl delete --wait=false pods/test-0`.
6. See a new pod is created and become pending.
7. Edit the PVC to remove the `aaa.bbb/ccc` finalizer.
8. See the PVC is get deleted.
9. See the Pod remains pending forever.

```release-note
Adding fix to the statefulset controller to wait for pvc deletion before creating pods.
```

### DIFF
--- a/pkg/controller/statefulset/stateful_pod_control.go
+++ b/pkg/controller/statefulset/stateful_pod_control.go
@@ -181,7 +181,7 @@ func (spc *realStatefulPodControl) recordClaimEvent(verb string, set *apps.State
 func (spc *realStatefulPodControl) createPersistentVolumeClaims(set *apps.StatefulSet, pod *v1.Pod) error {
 	var errs []error
 	for _, claim := range getPersistentVolumeClaims(set, pod) {
-		_, err := spc.pvcLister.PersistentVolumeClaims(claim.Namespace).Get(claim.Name)
+		pvc, err := spc.pvcLister.PersistentVolumeClaims(claim.Namespace).Get(claim.Name)
 		switch {
 		case apierrors.IsNotFound(err):
 			_, err := spc.client.CoreV1().PersistentVolumeClaims(claim.Namespace).Create(context.TODO(), &claim, metav1.CreateOptions{})
@@ -194,6 +194,10 @@ func (spc *realStatefulPodControl) createPersistentVolumeClaims(set *apps.Statef
 		case err != nil:
 			errs = append(errs, fmt.Errorf("failed to retrieve PVC %s: %s", claim.Name, err))
 			spc.recordClaimEvent("create", set, pod, &claim, err)
+		case err == nil:
+			if pvc.DeletionTimestamp != nil {
+				errs = append(errs, fmt.Errorf("pvc %s is being deleted", claim.Name))
+			}
 		}
 		// TODO: Check resource requirements and accessmodes, update if necessary
 	}
--- a/pkg/controller/statefulset/stateful_pod_control_test.go
+++ b/pkg/controller/statefulset/stateful_pod_control_test.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"strings"
 	"testing"
+	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -29,6 +30,7 @@ import (
 	"k8s.io/client-go/tools/record"
 
 	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes/fake"
 	corelisters "k8s.io/client-go/listers/core/v1"
 	_ "k8s.io/kubernetes/pkg/apis/apps/install"
@@ -121,6 +123,42 @@ func TestStatefulPodControlCreatePodPvcCreateFailure(t *testing.T) {
 	events := collectEvents(recorder.Events)
 	if eventCount := len(events); eventCount != 2 {
 		t.Errorf("PVC create failure: got %d events, but want 2", eventCount)
+	}
+	for i := range events {
+		if !strings.Contains(events[i], v1.EventTypeWarning) {
+			t.Errorf("Found unexpected non-warning event %s", events[i])
+		}
+	}
+}
+func TestStatefulPodControlCreatePodPvcDeleting(t *testing.T) {
+	recorder := record.NewFakeRecorder(10)
+	set := newStatefulSet(3)
+	pod := newStatefulSetPod(set, 0)
+	fakeClient := &fake.Clientset{}
+	pvcs := getPersistentVolumeClaims(set, pod)
+	pvcIndexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+	deleteTime := time.Date(2019, time.January, 1, 0, 0, 0, 0, time.UTC)
+	for k := range pvcs {
+		pvc := pvcs[k]
+		pvc.DeletionTimestamp = &metav1.Time{Time: deleteTime}
+		pvcIndexer.Add(&pvc)
+	}
+	pvcLister := corelisters.NewPersistentVolumeClaimLister(pvcIndexer)
+	control := NewRealStatefulPodControl(fakeClient, nil, nil, pvcLister, recorder)
+	fakeClient.AddReactor("create", "persistentvolumeclaims", func(action core.Action) (bool, runtime.Object, error) {
+		create := action.(core.CreateAction)
+		return true, create.GetObject(), nil
+	})
+	fakeClient.AddReactor("create", "pods", func(action core.Action) (bool, runtime.Object, error) {
+		create := action.(core.CreateAction)
+		return true, create.GetObject(), nil
+	})
+	if err := control.CreateStatefulPod(set, pod); err == nil {
+		t.Error("Failed to produce error on deleting PVC")
+	}
+	events := collectEvents(recorder.Events)
+	if eventCount := len(events); eventCount != 1 {
+		t.Errorf("Deleting PVC: got %d events, but want 1", eventCount)
 	}
 	for i := range events {
 		if !strings.Contains(events[i], v1.EventTypeWarning) {


ymmt2005:
@tnozicka 
Hi.  I'm glad if you can take a look at this.  This is a re-sent of #83273.
Thank you!

msau42:
/assign @tnozicka @janetkuo @kow3ns 
@kubernetes/sig-apps-pr-reviews

janetkuo (approved):

janetkuo:
/lgtm

Would leave it to @kow3ns or @tnozicka to approve.

ymmt2005:
`publishing-bot` test in `pull-kubernetes-verify` failed twice.  I have no idea about that.
/retest

ymmt2005:
The failure in `pull-kubernetes-verify` came from some infrastructure not related to this PR.

```
Verifying verify-publishing-bot.py

+++ Running case: verify.publishing-bot
+++ working dir: /home/prow/go/src/k8s.io/kubernetes
+++ command: python "hack/make-rules/../../hack/verify-publishing-bot.py"
Please install missing pyyaml module and re-run hack/make-rules/../../hack/verify-publishing-bot.py
+++ exit code: 1
+++ error: 1
ESC[0;31mFAILEDESC[0m   verify-publishing-bot.py        1s
========================
ESC[0;31mFAILED TESTSESC[0m
========================
ESC[0;31mhack/make-rules/../../hack/verify-publishing-bot.pyESC[0m
make: *** [Makefile:129: verify] Error 1
```

ymmt2005:
This.
https://github.com/kubernetes/kubernetes/issues/93623

ymmt2005:
All tests passed.

ymmt2005:
@kow3ns @tnozicka 
Hello.  Could you take a look at this?

pkg/controller/statefulset/stateful_pod_control.go
```diff
@@ -194,6 +194,8 @@ func (spc *realStatefulPodControl) createPersistentVolumeClaims(set *apps.Statef
 		case err != nil:
 			errs = append(errs, fmt.Errorf("failed to retrieve PVC %s: %s", claim.Name, err))
 			spc.recordClaimEvent("create", set, pod, &claim, err)
+		case pvc.DeletionTimestamp != nil:
```
tnozicka, line 197: the switch is on the error states so I'd prefer explicit
```
case err=nil:
    if pvc.DeletionTimestamp != nil
```

pkg/controller/statefulset/stateful_pod_control.go
```diff
@@ -194,6 +194,8 @@ func (spc *realStatefulPodControl) createPersistentVolumeClaims(set *apps.Statef
...
+			errs = append(errs, fmt.Errorf("pvc %s is to be deleted", claim.Name))
```
tnozicka, line 198:
```suggestion
			errs = append(errs, fmt.Errorf("pvc %s is being deleted", claim.Name))
```

pkg/controller/statefulset/stateful_pod_control_test.go
```diff
@@ -128,6 +130,42 @@ func TestStatefulPodControlCreatePodPvcCreateFailure(t *testing.T) {
...
+		return true, create.GetObject(), nil
+	})
+	fakeClient.AddReactor("create", "pods", func(action core.Action) (bool, runtime.Object, error) {
+		create := action.(core.CreateAction)
+		return true, create.GetObject(), nil
+	})
+	if err := control.CreateStatefulPod(set, pod); err == nil {
+		t.Error("Failed to produce error on deleting PVC")
+	}
+	events := collectEvents(recorder.Events)
```
tnozicka, line 159: Would it be easier to compare the arrays with a nice diff?
```golang
if !apiequality.Sematics.DeepEqual(events, expectedEvents) {
    t.Errorf("Expected events differ: %s", cmd.Diff(expectedEvents, events))
}
```

tnozicka (commented):
few nits but overall lgtm

pkg/controller/statefulset/stateful_pod_control_test.go
ymmt2005: This is aligned with the other test cases.
I think it should be done separately with the other tests.

pkg/controller/statefulset/stateful_pod_control.go
ymmt2005: Fixed as suggested.

pkg/controller/statefulset/stateful_pod_control.go
ymmt2005: Fixed

ymmt2005:
@tnozicka @janetkuo 
Could you take another look?

tnozicka (approved):
/approve
/lgtm

ymmt2005:
@janetkuo @tnozicka 
Thank you!  It seems that this PR needs a milestone.  Could you set one?

/milestone v1.19

pkg/controller/statefulset/stateful_pod_control.go
```diff
@@ -194,6 +194,10 @@ func (spc *realStatefulPodControl) createPersistentVolumeClaims(set *apps.Statef
 		case err != nil:
 			errs = append(errs, fmt.Errorf("failed to retrieve PVC %s: %s", claim.Name, err))
 			spc.recordClaimEvent("create", set, pod, &claim, err)
+		case err == nil:
+			if pvc.DeletionTimestamp != nil {
+				errs = append(errs, fmt.Errorf("pvc %s is being deleted", claim.Name))
+			}
```
cofyc, line 200: cc @janetkuo @msau42 I think we should check the API server if the `pvc.DeletionTimestamp == nil`. Otherwise, we didn't solve the cache inconsistency issue.

pkg/controller/statefulset/stateful_pod_control.go
```diff
@@ -181,7 +181,7 @@ func (spc *realStatefulPodControl) recordClaimEvent(verb string, set *apps.State
 func (spc *realStatefulPodControl) createPersistentVolumeClaims(set *apps.StatefulSet, pod *v1.Pod) error {
 	var errs []error
 	for _, claim := range getPersistentVolumeClaims(set, pod) {
-		_, err := spc.pvcLister.PersistentVolumeClaims(claim.Namespace).Get(claim.Name)
+		pvc, err := spc.pvcLister.PersistentVolumeClaims(claim.Namespace).Get(claim.Name)
```
ymmt2005, line 184:
```suggestion
		pvc, err := spc.client.CoreV1().PersistentVolumeClaims(claim.Namespace).Get(context.TODO(), claim.Name, metav1.GetOptions{})
```

pkg/controller/statefulset/stateful_pod_control.go
ymmt2005: @cofyc 
Does this change make sense?
https://github.com/kubernetes/kubernetes/pull/93457/files#r478778760

cofyc: no, we can first check the informer cache, then check the API server if the `pvc.DeletionTimestamp == nil`.

- if the PVC in informer cache is in terminating state, we don't need to check API server (good performance)
  - the terminating state cann't be rollbacked, so this is reliabe 
- if the PVC in informer cache is not in terminating state (`pvc.DeletionTimestamp == nil`), we need to check the API server, because it's possible there is a delay between the API server and informer cache

ymmt2005, line 200:
```suggestion
		case err == nil:
			if pvc.DeletionTimestamp == nil {
				// consult apiserver to get the latest state
				pvc, err = spc.client.CoreV1().PersistentVolumeClaims(claim.Namespace).Get(context.TODO(), claim.Name, metav1.GetOptions{})
				if err != nil {
					errs = append(errs, fmt.Errorf("failed to retrieve PVC %s: %s", claim.Name, err))
					break
				}
			}
			if pvc.DeletionTimestamp != nil {
				errs = append(errs, fmt.Errorf("pvc %s is being deleted", claim.Name))
			}
```

ymmt2005: @cofyc 
OK, so how about this?  looks good?
https://github.com/kubernetes/kubernetes/pull/93457#discussion_r478793874

cofyc: This looks good to me! @tnozicka what do you think?

tnozicka: how does the live call help? sounds racy the same way

maybe the statefulset should reconcile PVC (not only on pod creation) but that would be a bigger change

cofyc: this can help if the PVC is always deleted before the pod

https://github.com/kubernetes/kubernetes/issues/74374#issuecomment-682024912

> maybe the statefulset should reconcile PVC (not only on pod creation) but that would be a bigger change

yes, after some investigations, we found that hitting the issue (pod has no PVC to use) is high if the PVC is deleted after the controller checks the PVC's state because the PVC can be deleted successfully if the newly created pod is not scheduled, see the discussion here: https://github.com/kubernetes/kubernetes/issues/74374 

my suggestions:

- checking the deletion timestamp (live call maybe is not needed as we will have PVC reconciliation) 
- add PVC reconciliation

bpineau:
Could this fix be considered for a 1.19 backport?

andyxning:
This should be cherypicked to 1.19.

pkg/controller/statefulset/stateful_pod_control.go
andyxning: Why this is not added finally in the PR？

ymmt2005: Because this is not perfect too.  See https://github.com/kubernetes/kubernetes/pull/93457#discussion_r479962362

PR #93466, opened 2020-07-27
Add events message when failed to create routes

### LINKED ISSUE #93277: [Azure] Node Event doesn't contain the detailed errors when controller manager failed to create routes
**What happened**:

Node Events doesn't contain the detailed errors when controller manager failed to create routes, only time out errors are shown in node events.

The detailed errors should be shown in the node events for easier troubleshooting.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

/kind bug
/sig cloud-provider
/area provider/azure

zhouya0:
**What type of PR is this?**
/kind bug

Add events message when failed to create routes.

Fixes https://github.com/kubernetes/kubernetes/issues/93277

/cc @feiskyer

```release-note
None
```

### DIFF
--- a/staging/src/k8s.io/cloud-provider/controllers/route/route_controller.go
+++ b/staging/src/k8s.io/cloud-provider/controllers/route/route_controller.go
@@ -202,7 +202,7 @@ func (rc *RouteController) reconcile(nodes []*v1.Node, routes []*cloudprovider.R
 									Name:      string(nodeName),
 									UID:       types.UID(nodeName),
 									Namespace: "",
-								}, v1.EventTypeWarning, "FailedToCreateRoute", msg)
+								}, v1.EventTypeWarning, "FailedToCreateRoute: %v", msg)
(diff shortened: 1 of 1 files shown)


staging/src/k8s.io/cloud-provider/controllers/route/route_controller.go
```diff
@@ -202,7 +202,7 @@ func (rc *RouteController) reconcile(nodes []*v1.Node, routes []*cloudprovider.R
 									Name:      string(nodeName),
 									UID:       types.UID(nodeName),
 									Namespace: "",
-								}, v1.EventTypeWarning, "FailedToCreateRoute", msg)
+								}, v1.EventTypeWarning, "FailedToCreateRoute: %v", msg)
```
wojtek-t, line 205: Reason shouldn't depend on the exact object etc.

wojtek-t (commented):
/close

PR #93471, opened 2020-07-27
nodelocaldns.yaml: remove force_tcp option which used for external dn…

lemonli:
**What type of PR is this?**
>

/kind cleanup

I think `force_tcp` is not necessary when query an external domain name according to this [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0030-nodelocal-dns-cache.md#proposal) 
and udp may perform better as default.

Fixes #

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml
+++ b/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml
@@ -97,9 +97,7 @@ data:
         reload
         loop
         bind __PILLAR__LOCAL__DNS__ __PILLAR__DNS__SERVER__
-        forward . __PILLAR__UPSTREAM__SERVERS__ {
-                force_tcp
-        }
+        forward . __PILLAR__UPSTREAM__SERVERS__
         prometheus :9253
         }
 ---


prameshj:
The force_tcp option was removed for stubDomains in node-cache-1.15.13. It could be removed for external dns as well.

prameshj:
Was a specific performance issue observed with the force_tcp option?

lemonli:
> Was a specific performance issue observed with the force_tcp option?
Some cloud providers deny  tcp connections to their dns server

lemonli:
@MrHohn @prameshj  Ready to merge

prameshj:
> /hold
> 
> Would like to understand this more. We did this on purpose to make the upstream more reliable.

This PR is changing the behavior only for external DNS queries. The lookups to clusterDNS will still be over TCP.

bowei:
ok

/remove-hold

lemonli:
@bowei @MrHohn ready to merge, thx.

MrHohn (approved):
/approve

PR #79489, opened 2019-06-27
Adds WindowsOptions.RunAsUserName field

claudiubelu:
**What type of PR is this?**

/kind api-change

/sig windows

**What this PR does / why we need it**:

This adds windows configuration and wires up username in the podspec to the runtime interface.

#64009 added run_as_username to the container runtime interface, but did not hook it up in the Kubernetes v1.Container.SecurityContext.WindowsOptions.runAsUser field.

This PR also validates the ``RunAsUserName`` field, making sure that it valid, having the format ``DOMAIN\USER`` (case insensitive), where ``DOMAIN\`` is optional and has to be a valid NetBios or DNS domain name. For more information about the restrictions on the ``DOMAIN`` and ``USER`` parts, look here: [1] [2]

Adds the ``WindowsRunAsUserName`` alpha feature gate. By default, it is disabled. If the feature gate is not enabled, the ``WindowsOptions.RunAsUserName`` field will be dropped from both the ``PodSecurityContext`` and container ``SecurityContext``.

[1] https://support.microsoft.com/en-us/help/909264/naming-conventions-in-active-directory-for-computers-domains-sites-and
[2] https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.localaccounts/new-localuser?view=powershell-5.1

**Which issue(s) this PR fixes**:

Fixes: #73387

**Special notes for your reviewer**:

The original PR and discussions on it is here: https://github.com/kubernetes/kubernetes/pull/73609

**Does this PR introduce a user-facing change?**:

```release-note
Add v1.Container.SecurityContext.WindowsOptions.RunAsUserName to the pod spec 
```

PR #93475, opened 2020-07-27
Strip unnecessary security contexts on Windows

ravisantoshgudimetla:
**What type of PR is this?**

 /kind bug
/kind cleanup

As of now, the kubelet is passing the security context to container runtime even
if the security context has invalid options for a particular OS. As a result,
the pod fails to come up on the node. This error is particularly pronounced on
the Windows nodes where kubelet is allowing Linux specific options like SELinux,
RunAsUser etc where as in [documentation](https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#v1-container),
we clearly state they are not supported. This PR ensures that the kubelet strips
the security contexts of the pod, if they don't make sense on the Windows OS.

Fixes #

```release-note
Strip unnecessary security contexts on Windows
```

```docs

```
/sig windows

### DIFF
--- a/pkg/kubelet/kuberuntime/kuberuntime_container_windows.go
+++ b/pkg/kubelet/kuberuntime/kuberuntime_container_windows.go
@@ -19,7 +19,6 @@ limitations under the License.
 package kuberuntime
 
 import (
-	"fmt"
 	"runtime"
 
 	"k8s.io/api/core/v1"
@@ -125,10 +124,7 @@ func (m *kubeGenericRuntimeManager) generateWindowsContainerConfig(container *v1
 
 	// setup security context
 	effectiveSc := securitycontext.DetermineEffectiveSecurityContext(pod, container)
-	// RunAsUser only supports int64 from Kubernetes API, but Windows containers only support username.
-	if effectiveSc.RunAsUser != nil {
-		return nil, fmt.Errorf("run as uid (%d) is not supported on Windows", *effectiveSc.RunAsUser)
-	}
+
 	if username != "" {
 		wc.SecurityContext.RunAsUsername = username
 	}
--- a/pkg/kubelet/kuberuntime/kuberuntime_sandbox.go
+++ b/pkg/kubelet/kuberuntime/kuberuntime_sandbox.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"runtime"
 	"sort"
 
 	v1 "k8s.io/api/core/v1"
@@ -143,6 +144,9 @@ func (m *kubeGenericRuntimeManager) generatePodSandboxConfig(pod *v1.Pod, attemp
 }
 
 // generatePodSandboxLinuxConfig generates LinuxPodSandboxConfig from v1.Pod.
+// We've to call PodSandboxLinuxConfig always irrespective of the underlying OS as securityContext is not part of
+// podSandboxConfig. It is currently part of LinuxPodSandboxConfig. In future, if we have securityContext pulled out
+// in podSandboxConfig we should be able to use it.
 func (m *kubeGenericRuntimeManager) generatePodSandboxLinuxConfig(pod *v1.Pod) (*runtimeapi.LinuxPodSandboxConfig, error) {
 	cgroupParent := m.runtimeHelper.GetPodCgroupParent(pod)
 	lc := &runtimeapi.LinuxPodSandboxConfig{
@@ -169,15 +173,15 @@ func (m *kubeGenericRuntimeManager) generatePodSandboxLinuxConfig(pod *v1.Pod) (
 
 	if pod.Spec.SecurityContext != nil {
 		sc := pod.Spec.SecurityContext
-		if sc.RunAsUser != nil {
+		if sc.RunAsUser != nil && runtime.GOOS != "windows" {
 			lc.SecurityContext.RunAsUser = &runtimeapi.Int64Value{Value: int64(*sc.RunAsUser)}
 		}
-		if sc.RunAsGroup != nil {
+		if sc.RunAsGroup != nil && runtime.GOOS != "windows" {
 			lc.SecurityContext.RunAsGroup = &runtimeapi.Int64Value{Value: int64(*sc.RunAsGroup)}
 		}
 		lc.SecurityContext.NamespaceOptions = namespacesForPod(pod)
 
-		if sc.FSGroup != nil {
+		if sc.FSGroup != nil && runtime.GOOS != "windows" {
 			lc.SecurityContext.SupplementalGroups = append(lc.SecurityContext.SupplementalGroups, int64(*sc.FSGroup))
 		}
 		if groups := m.runtimeHelper.GetExtraSupplementalGroupsForPod(pod); len(groups) > 0 {
@@ -188,7 +192,7 @@ func (m *kubeGenericRuntimeManager) generatePodSandboxLinuxConfig(pod *v1.Pod) (
 				lc.SecurityContext.SupplementalGroups = append(lc.SecurityContext.SupplementalGroups, int64(sg))
 			}
 		}
-		if sc.SELinuxOptions != nil {
+		if sc.SELinuxOptions != nil && runtime.GOOS != "windows" {
 			lc.SecurityContext.SelinuxOptions = &runtimeapi.SELinuxOption{
 				User:  sc.SELinuxOptions.User,
 				Role:  sc.SELinuxOptions.Role,
--- a/test/e2e/windows/security_context.go
+++ b/test/e2e/windows/security_context.go
@@ -31,7 +31,7 @@ import (
 
 const runAsUserNameContainerName = "run-as-username-container"
 
-var _ = SIGDescribe("[Feature:Windows] SecurityContext RunAsUserName", func() {
+var _ = SIGDescribe("[Feature:Windows] SecurityContext", func() {
 	f := framework.NewDefaultFramework("windows-run-as-username")
 
 	ginkgo.It("should be able create pods and run containers with a given username", func() {
@@ -71,6 +71,29 @@ var _ = SIGDescribe("[Feature:Windows] SecurityContext RunAsUserName", func() {
 		f.TestContainerOutput("check overridden username", pod, 0, []string{"ContainerUser"})
 		f.TestContainerOutput("check pod SecurityContext username", pod, 1, []string{"ContainerAdministrator"})
 	})
+	ginkgo.It("should ignore Linux Specific SecurityContext if set", func() {
+		ginkgo.By("Creating a pod with SELinux options")
+		// It is sufficient to show that the pod comes up here. Since we're stripping the SELinux and other linux
+		// security contexts in apiserver and not updating the pod object in the apiserver, we cannot validate the
+		// the pod object to not have those security contexts. However the pod coming to running state is a sufficient
+		// enough condition for us to validate since prior to https://github.com/kubernetes/kubernetes/pull/93475
+		// the pod would have failed to come up.
+		windowsPodWithSELinux := createTestPod(f, windowsBusyBoximage, windowsOS)
+		windowsPodWithSELinux.Spec.Containers[0].Args = []string{"test-webserver-with-selinux"}
+		windowsPodWithSELinux.Spec.SecurityContext = &v1.PodSecurityContext{}
+		containerUserName := "ContainerAdministrator"
+		windowsPodWithSELinux.Spec.SecurityContext.SELinuxOptions = &v1.SELinuxOptions{Level: "s0:c24,c9"}
+		windowsPodWithSELinux.Spec.Containers[0].SecurityContext = &v1.SecurityContext{
+			SELinuxOptions: &v1.SELinuxOptions{Level: "s0:c24,c9"},
+			WindowsOptions: &v1.WindowsSecurityContextOptions{RunAsUserName: &containerUserName}}
+		windowsPodWithSELinux.Spec.Tolerations = []v1.Toleration{{Key: "os", Value: "Windows"}}
+		windowsPodWithSELinux, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).Create(context.TODO(),
+			windowsPodWithSELinux, metav1.CreateOptions{})
+		framework.ExpectNoError(err)
+		framework.Logf("Created pod %v", windowsPodWithSELinux)
+		framework.ExpectNoError(e2epod.WaitForPodNameRunningInNamespace(f.ClientSet, windowsPodWithSELinux.Name,
+			f.Namespace.Name), "failed to wait for pod %s to be running", windowsPodWithSELinux.Name)
+	})
 })
 
 func runAsUserNamePod(username *string) *v1.Pod {
@@ -80,6 +103,7 @@ func runAsUserNamePod(username *string) *v1.Pod {
 			Name: podName,
 		},
 		Spec: v1.PodSpec{
+			NodeSelector: map[string]string{"kubernetes.io/os": "windows"},
 			Containers: []v1.Container{
 				{
 					Name:    runAsUserNameContainerName,


pkg/securitycontext/util.go
```diff
@@ -150,27 +152,27 @@ func securityContextFromPodSecurityContext(pod *v1.Pod) *v1.SecurityContext {
 
 	synthesized := &v1.SecurityContext{}
 
-	if pod.Spec.SecurityContext.SELinuxOptions != nil {
+	if  runtime.GOOS != "windows" && pod.Spec.SecurityContext.SELinuxOptions != nil {
```
sjenning, line 155: extra space causing gofmt failure

pkg/securitycontext/util.go
```diff
@@ -61,12 +62,13 @@ func DetermineEffectiveSecurityContext(pod *v1.Pod, container *v1.Container) *v1
 		return containerSc
 	}
```
ddebroy, line 64: `DetermineEffectiveSecurityContext` returns right away if the checks [above](https://github.com/kubernetes/kubernetes/blob/3365ed5625797cfaafbdab368e901e9f4739329a/pkg/securitycontext/util.go#L57-L62) are satisfied. In that case, if the `effectiveSc` or `containerSc` happens to have the members of a security context that do not apply to Windows, they will continue to be returned (and will not be stripped as below).

Since the desired action is specific to Windows, have you considered cleaning up the security contexts in the Windows [caller](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_container_windows.go#L128)? That would:
1. Allow the early/optimized exit paths above to continue to work.
2. Allow `DetermineEffectiveSecurityContext` to be invoked from a cluster-wide controller context (if needed in the future) rather than the node-specific kubelet context. The `runtime.GOOS != "windows"` checks assumes `DetermineEffectiveSecurityContext` will never be invoked from a cluster-wide controller context.

pkg/securitycontext/util.go
```diff
@@ -61,12 +62,13 @@ func DetermineEffectiveSecurityContext(pod *v1.Pod, container *v1.Container) *v1
...
 
-	if containerSc.SELinuxOptions != nil {
+	// To preserve backwards compatibility
+	if runtime.GOOS != "windows" && effectiveSc != nil && containerSc != nil && containerSc.SELinuxOptions != nil {
```
ddebroy, line 66: The `effectiveSc != nil && containerSc != nil` checks seem to be taken care of by the checks [above](https://github.com/kubernetes/kubernetes/blob/3365ed5625797cfaafbdab368e901e9f4739329a/pkg/securitycontext/util.go#L57-L62) but I think this is safer code.

ddebroy (commented):
One overall comment on best place to perform the desired goal of this PR: Strip unnecessary security contexts on Windows.

pjh:
Can we add a test to verify that the desired behavior works once this PR is merged?

yliaog:
@ravisantoshgudimetla what recent change caused this pod crash behavior?

pkg/securitycontext/util.go
jsturtevant: Doing this work in the windows specific code makes sense. We have run into a few other places where this has come up with security context: https://github.com/kubernetes/kubernetes/pull/92355. In the future there could be a place where some of the these attributes might be available if a form of privileged containers becomes available.

The move to doing the checks in windows specific code makes the resulting logic here much simpler over adding complex if statements which will become hard to maintain and easily introduce bugs.

pkg/securitycontext/util.go
```diff
@@ -81,42 +83,42 @@ func DetermineEffectiveSecurityContext(pod *v1.Pod, container *v1.Container) *v1
...
 	}
 
-	if containerSc.RunAsGroup != nil {
+	if runtime.GOOS != "windows" && effectiveSc != nil && containerSc != nil && containerSc.RunAsGroup != nil {
 		effectiveSc.RunAsGroup = new(int64)
 		*effectiveSc.RunAsGroup = *containerSc.RunAsGroup
 	}
 
-	if containerSc.RunAsNonRoot != nil {
+	if runtime.GOOS != "windows" && effectiveSc != nil && containerSc != nil && containerSc.RunAsNonRoot != nil {
```
jsturtevant, line 106: I don't think this one should be overridden due to the conversation on https://github.com/kubernetes/kubernetes/pull/92355#issuecomment-654996758 where we decided to check that the `ContainerAdministrator` is not set as for a RunAsNonRoot check.

pkg/securitycontext/util.go
```diff
@@ -150,27 +152,27 @@ func securityContextFromPodSecurityContext(pod *v1.Pod) *v1.SecurityContext {
...
 	}
 
-	if pod.Spec.SecurityContext.RunAsGroup != nil {
+	if runtime.GOOS != "windows" && pod.Spec.SecurityContext.RunAsGroup != nil {
 		synthesized.RunAsGroup = new(int64)
 		*synthesized.RunAsGroup = *pod.Spec.SecurityContext.RunAsGroup
 	}
 
-	if pod.Spec.SecurityContext.RunAsNonRoot != nil {
+	if runtime.GOOS != "windows" && pod.Spec.SecurityContext.RunAsNonRoot != nil {
```
jsturtevant, line 175: same as above via: https://github.com/kubernetes/kubernetes/pull/92355#issuecomment-654996758

jsturtevant:
> Can we add a test to verify that the desired behavior works once this PR is merged?

Is there a reason this PR should not add tests for this iteration instead of after the fact?  At the very least should have unit tests for this behavior. Having an e2e test for this would also be valuable to verify the the behavior does improve.

pjh:
> Is there a reason this PR should not add tests for this iteration instead of after the fact?

Yeah that's what I meant, I just phrased it confusingly :)

ravisantoshgudimetla:
> @ravisantoshgudimetla what recent change caused this pod crash behavior?

This is not something new. This behavior was there from a long time

ravisantoshgudimetla:
Thank you @ddebroy @jsturtevant @pjh for your reviews. I addressed your comments. I added a test but I am not sure, if it is in right place. PTAL.

ravisantoshgudimetla:
Output of running this test on OpenShift cluster with Windows nodes.

```
[sig-windows] [Feature:Windows] SecurityContext
/home/ravig/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/windows/framework.go:27
  should ignore Linux Specific SecurityContext if set
  /home/ravig/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/windows/security_context.go:73
------------------------------
{"msg":"PASSED [sig-windows] [Feature:Windows1] SecurityContext should ignore Linux Specific SecurityContext if set","total":1,"completed":1,"skipped":3045,"failed":0}
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSJul 29 22:50:57.044: INFO: Running AfterSuite actions on all nodes
Jul 29 22:50:57.044: INFO: Running AfterSuite actions on node 1
{"msg":"Test Suite completed","total":1,"completed":1,"skipped":5237,"failed":0}

Ran 1 of 5238 Specs in 7.360 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 5237 Skipped
PASS
```

pkg/kubelet/kuberuntime/kuberuntime_sandbox.go
```diff
@@ -143,6 +144,9 @@ func (m *kubeGenericRuntimeManager) generatePodSandboxConfig(pod *v1.Pod, attemp
 }
 
 // generatePodSandboxLinuxConfig generates LinuxPodSandboxConfig from v1.Pod.
+// We've to call PodSandboxLinuxConfig always irrespective of the underlying OS as securityContext is not part of
```
ddebroy, line 147: Good point - it seems there is no point in calling `generatePodSandboxLinuxConfig` for Windows and therefore the `runtime.GOOS != "windows"` checks below can be moved [up](https://github.com/kubernetes/kubernetes/blob/428b500c5a93daded135f347428724334163859a/pkg/kubelet/kuberuntime/kuberuntime_sandbox.go#L140) to the caller? Note that in case of Windows, the relevant security context details (covering GMSA credspec details and runAsUsername) is passed as part of the [container security context]( https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1alpha2/api.proto#L789 ) (rather than pod/sandbox level context)

ddebroy (commented):
Thanks for addressing the comments above. I wanted to clarify a couple of things.

test/e2e/windows/security_context.go
```diff
@@ -71,6 +71,24 @@ var _ = SIGDescribe("[Feature:Windows] SecurityContext RunAsUserName", func() {
 		f.TestContainerOutput("check overridden username", pod, 0, []string{"ContainerUser"})
 		f.TestContainerOutput("check pod SecurityContext username", pod, 1, []string{"ContainerAdministrator"})
 	})
+	ginkgo.It("should ignore Linux Specific SecurityContext if set", func() {
+		ginkgo.By("Creating a pod with SELinux options")
+
+		windowsPodWithSELinux := createTestPod(f, "mcr.microsoft.com/powershell:lts-nanoserver-1809", windowsOS)
```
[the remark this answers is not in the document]
ravisantoshgudimetla: I started with `agnhost` host image on Kubernetes cluster and it is working fine, however since I made this test distro specific(particularly SELinux), I tested with `OpenShift cluster` with Windows 1809 nodes, I noticed that the `agnhost` image pulling failed on Windows 1809 nodes. I can revert the change and test it downstream and get back to you later.

test/e2e/windows/security_context.go
```diff
@@ -71,6 +71,24 @@ var _ = SIGDescribe("[Feature:Windows] SecurityContext RunAsUserName", func() {
...
+		containerUserName := "ContainerAdministrator"
+		windowsPodWithSELinux.Spec.SecurityContext.SELinuxOptions = &v1.SELinuxOptions{Level: "s0:c24,c9"}
+		windowsPodWithSELinux.Spec.Containers[0].SecurityContext = &v1.SecurityContext{
+			WindowsOptions: &v1.WindowsSecurityContextOptions{RunAsUserName: &containerUserName}}
+		windowsPodWithSELinux.Spec.Tolerations = []v1.Toleration{{Key: "os", Value: "Windows"}}
+		windowsPodWithSELinux, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).Create(context.TODO(),
+			windowsPodWithSELinux, metav1.CreateOptions{})
+		framework.ExpectNoError(err)
+		framework.Logf("Created pod %v", windowsPodWithSELinux)
+		framework.ExpectNoError(e2epod.WaitForPodNameRunningInNamespace(f.ClientSet, windowsPodWithSELinux.Name,
```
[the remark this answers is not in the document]
ravisantoshgudimetla: I thought about it if you're talking about looking at the pod fields if SELinux options are stripped, it may not be possible as since we are stripping the SELinux info in the kubelet and we're not making any apiserver call to update the pod spec, that information will not be available in the apiserver, so the pod.Spec will still have information related to SELinux, just that we ignore them on the kubelet side.

Having said that, IIUC, without this change the pod was not coming to running state, now if it comes, it means we should be good and the output of the container logs doesn't matter?

ravisantoshgudimetla (commented):
Thanks for the review @ddebroy. Please look at my responses

test/e2e/windows/security_context.go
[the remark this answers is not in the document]
ddebroy: Makes sense that the apiserver will not be able to reflect the adjustments done in the kubelet. Thanks for pointing that out! The pod coming up would indeed be the only verification we can perform in the e2e. If you can add this fact as a comment in the test somewhere, that would be great.

pkg/kubelet/kuberuntime/kuberuntime_sandbox.go
ddebroy: Got it. I have filed https://github.com/kubernetes/kubernetes/issues/93580 to keep track of the followup around this.

ravisantoshgudimetla: Thank you,

test/e2e/windows/security_context.go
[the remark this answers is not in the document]
ravisantoshgudimetla: Added the comment.

ravisantoshgudimetla (commented):
Thank you @ddebroy for the review.  I've updated my PR with the changes requested. PTAL

test/e2e/windows/security_context.go
```diff
@@ -71,6 +71,28 @@ var _ = SIGDescribe("[Feature:Windows] SecurityContext RunAsUserName", func() {
 		f.TestContainerOutput("check overridden username", pod, 0, []string{"ContainerUser"})
 		f.TestContainerOutput("check pod SecurityContext username", pod, 1, []string{"ContainerAdministrator"})
 	})
+	ginkgo.It("should ignore Linux Specific SecurityContext if set", func() {
+		ginkgo.By("Creating a pod with SELinux options")
+		// It is sufficient to show that the pod comes up here. Since we're stripping the SELinux and other linux
+		// security contexts in apiserver and not updating the pod object in the apiserver, we cannnot validate the
```
ddebroy, line 77: typo in spelling `cannnot`. `pull-kubernetes-verify` does not like it :-)

test/e2e/windows/security_context.go
```diff
@@ -80,6 +102,7 @@ func runAsUserNamePod(username *string) *v1.Pod {
 			Name: podName,
 		},
 		Spec: v1.PodSpec{
+			NodeSelector: map[string]string{"kubernetes.io/os": "windows"},
```
jsturtevant, line 105: Just a note that it is a requirement to taint the nodes before running the windows tests. We didn't want to change the specs for all the tests to add the labels so this would be why it this wasn't there initially. I think because this tests is windows only and this pod has windows specific setting it ok to add here.

test/e2e/windows/security_context.go
```diff
@@ -71,6 +71,28 @@ var _ = SIGDescribe("[Feature:Windows] SecurityContext RunAsUserName", func() {
...
+		// security contexts in apiserver and not updating the pod object in the apiserver, we cannot validate the
+		// the pod object to not have those security contexts. However the pod coming to running state is a sufficient
+		// enough condition for us to validate since prior to https://github.com/kubernetes/kubernetes/pull/93475
+		// the pod would have failed to come up.
+		windowsPodWithSELinux := createTestPod(f, windowsBusyBoximage, windowsOS)
+		windowsPodWithSELinux.Spec.Containers[0].Args = []string{"test-webserver-with-selinux"}
+		windowsPodWithSELinux.Spec.SecurityContext = &v1.PodSecurityContext{}
+		containerUserName := "ContainerAdministrator"
+		windowsPodWithSELinux.Spec.SecurityContext.SELinuxOptions = &v1.SELinuxOptions{Level: "s0:c24,c9"}
+		windowsPodWithSELinux.Spec.Containers[0].SecurityContext = &v1.SecurityContext{
```
jsturtevant, line 86: I think it would make sense to add linux options to the container level as well to ensure it gets cleared as well when set there

pkg/kubelet/kuberuntime/kuberuntime_container_windows.go
```diff
@@ -125,9 +124,31 @@ func (m *kubeGenericRuntimeManager) generateWindowsContainerConfig(container *v1
...
+		effectiveSc.Capabilities = nil
+	}
+
+	if effectiveSc.Privileged != nil {
+		effectiveSc.Privileged = nil
+	}
+
+	if effectiveSc.ProcMount != nil {
+		effectiveSc.ProcMount = nil
+	}
```
jsturtevant, line 147: given the e2e test doesn't test every one of these scenarios, I would suggest add unit tests here as well

pkg/kubelet/kuberuntime/kuberuntime_container_windows.go
```diff
@@ -125,9 +124,31 @@ func (m *kubeGenericRuntimeManager) generateWindowsContainerConfig(container *v1
 
 	// setup security context
 	effectiveSc := securitycontext.DetermineEffectiveSecurityContext(pod, container)
+
+	// Strip down all the unnecessary options on the Windows
```
jsturtevant, line 128: Is it worth adding a comment around what is available vs what is not for future contributors.  I could see it easily coming to this and wondering why some fields are cleared and some are not

jsturtevant:
Thanks for adding the tests!

It might not be possible due to the fact that some fields are allow through and others not but is there any way to ensure we don't run into this issue if a field is added to the security context in the future?  I guess adding the e2e test is that way to catch it?

test/e2e/windows/security_context.go
ravisantoshgudimetla: We have mixed nodes(both linux+windows) in OpenShift cluster, so it became necessary to run e2es.

pkg/kubelet/kuberuntime/kuberuntime_container_windows.go
ddebroy: Setting the fields of `effectiveSc` to `nil` in this function is not necessary as [discussed](https://github.com/kubernetes/kubernetes/pull/93475/files#r463014435) - those fields aren't used/referred to anywhere right now. It also has adds a little bit of a maintenance burden as @jsturtevant mentions above. The core intent here is to avoid returning the error [below](https://github.com/kubernetes/kubernetes/pull/93475/files#diff-4d8e81d13b02ca9a4b8919ef8cf8b6f9L130). So how about just removing the error? That error was introduced in https://github.com/kubernetes/kubernetes/pull/64009 before `WindowsOptions.RunAsUserName` was introduced in https://github.com/kubernetes/kubernetes/pull/79489 and documented as the way to pass a username for containers. So removing just that error should be good now (rather than setting the various fields to nil). Thoughts, @jsturtevant ?

ravisantoshgudimetla (commented):
> Thanks for adding the tests!
> 
> It might not be possible due to the fact that some fields are allow through and others not but is there any way to ensure we don't run into this issue if a field is added to the security context in the future? I guess adding the e2e test is that way to catch it?

I modified the code to remove other checks as @ddebroy suggested. PTAL @ddebroy and @jsturtevant

ddebroy:
Thanks @ravisantoshgudimetla . `pull-kubernetes-integration` failures appear to be unrelated flakes.

/lgtm

test/e2e/windows/security_context.go
jsturtevant: make sense, for running tests on testgrid we taint the linux nodes (in this case it's only the control nodes): 

https://github.com/kubernetes-sigs/windows-testing/blob/ee418e4b7c5d6ce74cc90cf74e0b48754ce0463b/extensions/master_extension/v1/win-e2e-master-extension.sh#L52
https://github.com/kubernetes-sigs/windows-testing/blob/ee418e4b7c5d6ce74cc90cf74e0b48754ce0463b/gce/run-e2e.sh#L15

Adding the node selector here isn't a problem due to the fact this is in the `sig-windows` folder and only applicable to windows nodes.

pkg/kubelet/kuberuntime/kuberuntime_container_windows.go
jsturtevant: It seems that https://github.com/kubernetes/kubernetes/pull/93475/files#diff-0170c1534446e377b025a1c54ab478e2R176 also effectively doesn't copy the RunAsUserName into the Security context.  Removing the dependency with #93580 should make this even better.

wawa0210:
I am glad to see this pr.I am very excited to see this pr, I have some questions to ask:

- Currently pr is just Strip unnecessary security contexts on Windows, and does not apply some of the windows' own te'x
- There is currently a bug in windows. If we specify the runasuser of the pod, the windows sandbox is not used, but directly uses the `ContainerAdministrator` identity to run the container

It seems that the current pr does not really solve these problems

If we want to solve these problems, we need to run different logics according to the operating system when sanbox is created, and at the same time pass the identity information to cri

I am also trying to fix this logic recently. I mentioned an issue #92963 and a pr #93299 before(wip).  Can you help discuss whether this method is ok

ravisantoshgudimetla:
@wawa0210 - Thank you for reviewing the PR.

> Currently pr is just Strip unnecessary security contexts on Windows, and does not apply some of the windows' own te'x

Please look at the comment - https://github.com/kubernetes/kubernetes/pull/93475#discussion_r462818414. It is my understanding that for the Windows pods there is no need to have sandbox level security Constraints. If we want to push the securityContext outside from Container Security Context to sandbox security constraints it needs a broader discussion as it needs api changes and beyond the scope of current PR

> There is currently a bug in windows. If we specify the runasuser of the pod, the windows sandbox is not used, but directly uses the ContainerAdministrator identity to run the container

The first question that we need to ask is do we need container sandbox for Windows considering all the security options are at the container Security contexts. What does the new WindowsSandboxConfig offer - just the security options, for example in case of linux we need to identify parent cgroup etc. If it is just the security options there is no need to define new api around it.

wawa0210:
> Please look at the comment - #93475 (comment). It is my understanding that for the Windows pods there is no need to have sandbox level security Constraints. If we want to push the securityContext outside from Container Security Context to sandbox security constraints it needs a broader discussion as it needs api changes and beyond the scope of current PR

Thank you for your reply. It is true that Windows pods support sandbox level security Constraints and do not belong to the scope of the current pr. But I think this feature is really needed, from https://github.com/kubernetes/kubernetes/issues/92963#issuecomment-663504447. Because if we need to implement this feature, it will include the changes in the current pr

If possible, we can discuss this topic in another pr or issue  #92963 or #93299

liggitt:

jsturtevant:
I took a look at this from the perspective of where sig-windows would like to go longer term with the [KEP for Privileged Containers](https://docs.google.com/document/d/12EUtMdWFxhTCfFrqhlBGWV70MkZZPOgxw0X-LTR0VAo/edit#) and in relation to the issue for supporting runasuser in pod sandbox (https://github.com/kubernetes/kubernetes/issues/92963)

A few things standout:

1. As of right not there is no sandbox level security constraints, with Privileged containers that will change.  It will be necessary to ensure the user that is running in sandbox and containers have the right privileges to restrict file access and to turn the container into a privileged container.
2. There is a need to remove the dependency on linux sandbox config (https://github.com/kubernetes/kubernetes/issues/93580) 
3. The e2e tests that is added helps ensure this type of issue doesn't come up again

I think the approach in #93299 is closer to what we need longer term (solves 1 & 2) and would negate the need for the code here that strips the security constraints on the linux configuration but there is more discussion I think that needs to happen (CRI changes and containerd would need to support the sandbox configuration). I also think the test this PR introduces is valuable for sig-windows. 

Since the Privileged Container KEP is still in review and the CRI changes are bigger, this LGTM with the caveat that once we split the dependency on linux configuration these checks should be removed (thought they wouldn't hurt anything if it was moved to a linux only file as proposed by #93299)  

Thoughts?

jsturtevant:
one additional thought, @ravisantoshgudimetla  did you look into adding unit tests for this in https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_sandbox_test.go?

test/e2e/windows/security_context.go
ravisantoshgudimetla: Added

ravisantoshgudimetla:
> Since the Privileged Container KEP is still in review and the CRI changes are bigger, this LGTM with the caveat that once we split the dependency on linux configuration these checks should be removed (though they wouldn't hurt anything if it was moved to a linux only file as proposed by #93299)

@jsturtevant - I'd also like to remove the check on the kubelet side and have the api to have windowsSpecific fields in the long term.


> one additional thought, @ravisantoshgudimetla did you look into adding unit tests for this in https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_sandbox_test.go?

I initially wanted to add units but there were not many edge cases considering the changes we're making OS specific changes

jsturtevant:
@SergeyKanzhelev @ddebroy could we get a review/approval?

pkg/kubelet/kuberuntime/kuberuntime_sandbox.go
```diff
@@ -169,15 +173,15 @@ func (m *kubeGenericRuntimeManager) generatePodSandboxLinuxConfig(pod *v1.Pod) (
 
 	if pod.Spec.SecurityContext != nil {
 		sc := pod.Spec.SecurityContext
-		if sc.RunAsUser != nil {
+		if sc.RunAsUser != nil && runtime.GOOS != "windows" {
```
SergeyKanzhelev, line 176: it looks like misconfiguration if windows pod configured like this. Should this `if` has else to report this misconfiguration as a warning?

SergeyKanzhelev:
@ddebroy your approval is needed for test/e2e/windows/OWNERS

pkg/kubelet/kuberuntime/kuberuntime_sandbox.go
jsturtevant: we discussed this a bit as well at sig-windows, my understanding is the use case where users are running into this is when there was automation that assigns security contexts to all pods.  The expectation from the windows perspective since these fields are marked as not supported on windows they would be ignored. @ravisantoshgudimetla can confirm

sjenning:
Issue capturing the weirdness of calling `generatePodSandboxLinuxConfig` then checking for `runtime.GOOS != "windows"`, which should be implied by the function name, is captured in https://github.com/kubernetes/kubernetes/issues/93580.  We should remedy that.

In the meantime,
/approve

Issue #70137, opened 2018-10-23
Decide what to do with conformance tests that need privileged mode


dims:
Based on findings in:
- https://github.com/kubernetes/kubernetes/issues/59978
- https://github.com/kubernetes/kubernetes/pull/69935#issuecomment-432340200

We have SIG-Storage tests that have been coded to use `privileged` mode when they do not really need the `privileged` mode. These tests only fail under SELinux *if* `privileged` mode is set to false. So what can we do now?

- Option 1 : Nudge SIG-Storage to see if there is a way to get the feature working under SELinux without needing `privileged` mode
- Option 2 : Change the tests to detect if SELinux is on and set `privileged` mode is set to true if so
- Option 3 : Mark these tests with `[privileged]` and say that it is ok to skip these tests when vendors are testing conformance where/when they lock down kubelet to disallow privileged mode
- Option 4 : Drop these tests from conformance

/area conformance
/sig storage
/sig architecture

cc @kubernetes/cncf-conformance-wg

mgdevstack:
@dims No existing Conformance tests from the suite are running in Privileged mode. Though there are few e2e tests which verify privileged behavior viz -
> should support r/w [NodeConformance]
> PrivilegedPod [NodeConformance] should enable privileged commands
>etc....

Can we close this issue?

spiffxp:
I'm willing to take your word for it, but still feel compelled to do a sanity check (which is going to take me some time)

I'm wondering if there's a way we could lint this somehow

mgdevstack:
For a documented purpose maintaining list of existing e2e dealing with `Privileged: true`. (Please keep adding test in comment incase any e2e is missed to mention here)

1. [sig-auth] PodSecurityPolicy should allow pods under the privileged policy.PodSecurityPolicy
2. [k8s.io] [sig-node] AppArmor load AppArmor profiles should enforce an AppArmor profile
3. [k8s.io] [sig-node] AppArmor load AppArmor profiles can disable an AppArmor profile, using unconfined
4. [sig-storage] HostPath should support r/w [NodeConformance]
5. [k8s.io] PrivilegedPod [NodeConformance] should enable privileged commands [LinuxOnly]
[...]

PR #93482, opened 2020-07-27
Categorize existing conformance tests as Base or Privileged

johnbelamaric:
**What type of PR is this?**
/kind cleanup

Implements the first two conformance profiles by labeling existing tests as described in the [Profiles KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-architecture/1618-conformance-profiles).

This categorizes the tests based upon https://github.com/kubernetes/community/pull/4994 and there is a worksheet (open to k-dev) describing the specific reason for each test put into "Privileged":

https://docs.google.com/spreadsheets/d/1pUiZj_ZJ013wx9ZHvBw18oy7mpV7Xaq2d7Wv75Xk7iI/edit#gid=758339929

Note that this implementation is the simplest possible; it just adds a single string parameter to `ConformanceIt` that defines the profile. This is sufficient for now; if we want overlapping profiles in the future it would need to change.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/apimachinery/aggregator.go
+++ b/test/e2e/apimachinery/aggregator.go
@@ -94,7 +94,7 @@ var _ = SIGDescribe("Aggregator", func() {
 		    Description: Ensure that the sample-apiserver code from 1.17 and compiled against 1.17
 			will work on the current Aggregator/API-Server.
 	*/
-	framework.ConformanceIt("Should be able to support the 1.17 Sample API Server using the current Aggregator", func() {
+	framework.ConformanceIt("Privileged", "Should be able to support the 1.17 Sample API Server using the current Aggregator", func() {
 		// Testing a 1.17 version of the sample-apiserver
 		TestSampleAPIServer(f, aggrclient, imageutils.GetE2EImage(imageutils.APIServer))
 	})
--- a/test/e2e/apimachinery/crd_conversion_webhook.go
+++ b/test/e2e/apimachinery/crd_conversion_webhook.go
@@ -144,7 +144,7 @@ var _ = SIGDescribe("CustomResourceConversionWebhook [Privileged:ClusterAdmin]",
 		Description: Register a conversion webhook and a custom resource definition. Create a v1 custom
 		resource. Attempts to read it at v2 MUST succeed.
 	*/
-	framework.ConformanceIt("should be able to convert from CR v1 to CR v2", func() {
+	framework.ConformanceIt("Privileged", "should be able to convert from CR v1 to CR v2", func() {
 		testcrd, err := crd.CreateMultiVersionTestCRD(f, "stable.example.com", func(crd *apiextensionsv1.CustomResourceDefinition) {
 			crd.Spec.Versions = apiVersions
 			crd.Spec.Conversion = &apiextensionsv1.CustomResourceConversion{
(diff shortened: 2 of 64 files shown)


test/e2e/framework/framework.go
```diff
@@ -593,8 +593,8 @@ func KubeDescribe(text string, body func()) bool {
 }
 
 // ConformanceIt is wrapper function for ginkgo It.  Adds "[Conformance]" tag and makes static analysis easier.
-func ConformanceIt(text string, body interface{}, timeout ...float64) bool {
-	return ginkgo.It(text+" [Conformance]", body, timeout...)
+func ConformanceIt(profile, text string, body interface{}, timeout ...float64) bool {
+	return ginkgo.It(text+" [Conformance] [Profile:" + profile +"]", body, timeout...)
```
neolit123, line 597: while the format seems fine, the conformance testing KEP or the conformance profiles KEP do not set the format in stone (or at least i could not find it).

does it make sense to do that?

test/e2e/apimachinery/aggregator.go
```diff
@@ -94,7 +94,7 @@ var _ = SIGDescribe("Aggregator", func() {
 		    Description: Ensure that the sample-apiserver code from 1.17 and compiled against 1.17
 			will work on the current Aggregator/API-Server.
 	*/
-	framework.ConformanceIt("Should be able to support the 1.17 Sample API Server using the current Aggregator", func() {
+	framework.ConformanceIt("Privileged", "Should be able to support the 1.17 Sample API Server using the current Aggregator", func() {
```
neolit123, line 97: there is opportunity to have these as constants under /test/e2e/profiles:
```
import "k8s.io/kubernetes/test/e2e/profiles"

// usage:
// profiles.ConformanceBase
// profiles.ConformancePriviliged
```

ideally, the framework should be unaware of the profile declarations.

smarterclayton: Yeah, i'm not even sure framework.ConformanceIt is really appropriate.  I think the package being called conformance`conformance.BaseProfileIt(...)` in general would be clearer.

Conformance is a subset of our suite.

johnbelamaric: that sounds reasonable

test/e2e/apimachinery/crd_conversion_webhook.go
```diff
@@ -144,7 +144,7 @@ var _ = SIGDescribe("CustomResourceConversionWebhook [Privileged:ClusterAdmin]",
 		Description: Register a conversion webhook and a custom resource definition. Create a v1 custom
 		resource. Attempts to read it at v2 MUST succeed.
 	*/
-	framework.ConformanceIt("should be able to convert from CR v1 to CR v2", func() {
+	framework.ConformanceIt("Privileged", "should be able to convert from CR v1 to CR v2", func() {
```
smarterclayton, line 147: So a great example here of privileged - we are going to eventually have to add this somewhere else (like an integrator is going to want to have a profile that helps them focus on whether their operator / extension / api-extension works).  So then we need something later on that tags a test into multiple profiles?

Jefftree: +1, I think we should change this to an array if we want to support multiple profiles for a test.

spiffxp:
ref: https://github.com/kubernetes/kubernetes/issues/70137

test/e2e/apimachinery/aggregator.go
spiffxp: I'm fine with all of these as improvements but don't view them as necessary to land in order to land segmenting tests into profiles via this PR.

If we're going to shuffle packages around and rename, I would suggest waiting to do that as part of testing-commons' effort to detangle test/e2e/framework enough to move into staging.

memorygardener:
Thanks for taking things forward @johnbelamaric!

Looking at the referenced excel sheet, we have chosen to categorize those sets of tests that request for node level resources (HostPort, HostNetwork, HostIPC, HostPath, HostPID) or require Cluster Admin as "Privileged". I think this makes a lot of sense, and maybe we should document the criteria somewhere so tests that get added in the future can reference it? Maybe this exists, but if not, I'm happy to help create this :-) I suppose a README in the relevant directory so its closer to the code itself.

One additional comment I had was around where we see NodePort land while expressing these profiles. It seems to me that it should also be considered "Privileged" in that the K8s application here is requesting for a node level resource from the node's root namespace TCP stack (on all the nodes in the cluster). Most cloud providers need to add routes or poke holes in the firewall to allow for traffic to ingress into NodePort, and those operations require a higher privilege level than the devops persona. If we think HostPort is "Privileged", why would NodePort not be considered the same? In both cases, the pod is utilizing a resource from the node's TCP stack as opposed to from its own namespace.

PR #93483, opened 2020-07-27
Remove false positive warning in kubeadm cmd

### LINKED ISSUE #2230: remove "WARNING: kubeadm cannot validate component configs for API groups"
there were a number of cases in the #kubeadm slack channel where users are asking why they are getting this warning for commands that do not accept `--config` - e.g. `kubeadm token ...`

ideally the warning should be printed only if the user has provided a `--config`, but i'm pretty sure this logic is difficult to untangle, so we should instead just remove this warning or hide it behind a higher verbosity (e.g. `klog.V(1).Infof(....)`).

cc @rosti
/help
/kind cleanup
/priority backlog
/milestone v1.20

helenfeng737:
**What type of PR is this?**
> /kind cleanup

This PR is to remove false positive warning message when using several kubeadm commands. 
 
Fixes https://github.com/kubernetes/kubeadm/issues/2230

```release-note
None
```

```docs
None
```

### DIFF
--- a/cmd/kubeadm/app/componentconfigs/configset.go
+++ b/cmd/kubeadm/app/componentconfigs/configset.go
@@ -17,8 +17,6 @@ limitations under the License.
 package componentconfigs
 
 import (
-	"sort"
-
 	"github.com/pkg/errors"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -338,13 +336,7 @@ func GetVersionStates(clusterCfg *kubeadmapi.ClusterConfiguration, client client
 }
 
 // Validate is a placeholder for performing a validation on an already loaded component configs in a ClusterConfiguration
-// Currently it prints a warning that no validation was performed
+// TODO: investigate if the function can be repurposed for validating component config via CLI
 func Validate(clusterCfg *kubeadmapi.ClusterConfiguration) field.ErrorList {
-	groups := []string{}
-	for group := range clusterCfg.ComponentConfigs {
-		groups = append(groups, group)
-	}
-	sort.Strings(groups) // The sort is needed to make the output predictable
-	klog.Warningf("WARNING: kubeadm cannot validate component configs for API groups %v", groups)
 	return field.ErrorList{}
 }


SataQiu (commented):
Thanks @ZhiFeng1993!
/ok-to-test
/kind cleanup
/priority backlog

cmd/kubeadm/app/componentconfigs/configset.go
```diff
@@ -345,6 +345,5 @@ func Validate(clusterCfg *kubeadmapi.ClusterConfiguration) field.ErrorList {
 		groups = append(groups, group)
 	}
 	sort.Strings(groups) // The sort is needed to make the output predictable
-	klog.Warningf("WARNING: kubeadm cannot validate component configs for API groups %v", groups)
 	return field.ErrorList{}
```
rosti, line 348: @ZhiFeng1993 , can you remove everything else from this func but this line?
Also, the "Currently it prints a warining ..." comment above the function needs removal too.

rosti (changes requested):
Thanks for taking this on @ZhiFeng1993 !

cmd/kubeadm/app/componentconfigs/configset.go
ZhiFeng1993: Done. I'm a bit confused about this function though. What's the point of returning an empty list?

neolit123: we spoke briefly about this function during yesterday's kubeadm office hours meeting.

there are two options:
- remove the function until we need it.
- add a TODO on top of the function to identify this is reserved for component config validation in the future.

@rosti should be able to provide more information.

rosti: @ZhiFeng1993 
> Done. I'm a bit confused about this function though. What's the point of returning an empty list?

An older version of this function was performing actual validation of component configs in the previous implementation of the package. In the new implementation this becomes impossible due to the fact that we can no longer import the validation logic of each component.
As a drop-in replacement, I have started an effort to introduce a standard command line interface for each Kubernetes component to validate its config when invoked from the CLI.
Hence, when I re-implemented the `componentconfigs` package, this function's old body was deleted and replaced with a "warning" placeholder. However, validating the configs is not so critical, plus the effort for the standard CLI for validation has ground to a stop ATM.
So this function can either be deleted or left empty with a TODO. I don't have a personal preference actually.

ZhiFeng1993: Thank you both for the explanation.  Since you guys have future plans for this function, I'll do as suggested.

cmd/kubeadm/app/componentconfigs/configset.go
```diff
@@ -338,13 +336,7 @@ func GetVersionStates(clusterCfg *kubeadmapi.ClusterConfiguration, client client
 }
 
 // Validate is a placeholder for performing a validation on an already loaded component configs in a ClusterConfiguration
-// Currently it prints a warning that no validation was performed
+// TODO: This empty function is a place holder for future work on standard CLI interface for config validation.
```
neolit123, line 339: requesting a minor adjustment to the TODO:

> TODO: investigate if the function can be repurposed for validating component config via CLI

/approve

neolit123:
/lgtm
thanks @ZhiFeng1993 
this should merge once 1.19 code freeze is over.

Issue #93485, opened 2020-07-27
[Flaky Test] integration-master scheduler tests (ci-kubernetes-integration-master)


liggitt:
**Which jobs are flaking**:

* ci-kubernetes-integration-master
* pull-kubernetes-integration

**Which test(s) are flaking**:

Flaking in pre and post-submit:
* TestPostFilterPlugin
* TestReservePluginUnreserve
* TestFilterPlugin
* TestScorePlugin
* TestNodeAffinity

**Testgrid link**:

* https://testgrid.k8s.io/sig-release-master-blocking#integration-master&width=5&include-filter-by-regex=test%2Fintegration%2Fscheduler
* https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-integration&width=5&include-filter-by-regex=test%2Fintegration%2Fscheduler

**Reason for failure**:

```
    framework_test.go:671: Expected the filter plugin to be called 2 times, but got 4.
    framework_test.go:674: Expected the postfilter plugin to be called 1 times, but got 2.
```

Triage link: https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&test=TestPostFilterPlugin%7CTestReservePluginUnreserve%7CTestFilterPlugin%7CTestScorePlugin%7CTestNodeAffinity

/sig scheduling
@kubernetes/sig-scheduling-test-failures

liggitt:
/milestone v1.19

since these are flaking in a post-submit release-blocking job

liggitt:
@kubernetes/ci-signal

Huang-Wei:
I will take a look.

/assign

Huang-Wei:
https://testgrid.k8s.io/sig-release-master-blocking#integration-master&width=5&include-filter-by-regex=test%2Fintegration%2Fscheduler is all green since the last PR (#93550) gets merged around 4:30 AM PST Jul 30.

I will keep an eye on it for the next couple of days.

liggitt:
Solid green since that merged, thanks

PR #93490, opened 2020-07-28
Fix integration test flake on TestFilter and TestPostFilter

Huang-Wei:
**What type of PR is this?**

/kind flake
/sig scheduling
/priority critical-urgent

The flake of `TestFilter` and `TestPostFilter` integration test assumes the incoming Pod goes through ONE scheduling cycle, however, this is not unit test and we cannot control precisely how many cycles it goes through in an integration test. In the flake logs ([1], [2]), we can observe the same pattern that the Pod is scheduled twice (L1 and L13) below:

```
1 I0727 09:34:20.746786  112487 scheduler.go:455] Attempting to schedule pod: postfilter2-4cc9af20-f4bc-41a2-aeaf-d499d3afadcb/test-pod
2 I0727 09:34:20.746944  112487 factory.go:445] "Unable to schedule pod; no fit; waiting" pod="postfilter2-4cc9af20-f4bc-41a2-aeaf-d499d3afadcb/test-pod" err="0/1 nodes are available: 1 reject pod test-pod."
3 I0727 09:34:20.747001  112487 scheduler.go:347] Updating pod condition for postfilter2-4cc9af20-f4bc-41a2-aeaf-d499d3afadcb/test-pod to (PodScheduled==False, Reason=Unschedulable)
4 I0727 09:34:20.749183  112487 httplog.go:89] "HTTP" verb="POST" URI="/api/v1/namespaces/default/services" latency="17.073728ms" userAgent="scheduler.test/v0.0.0 (linux/amd64) kubernetes/$Format" srcIP="127.0.0.1:48432" resp=201
5 I0727 09:34:20.750992  112487 httplog.go:89] "HTTP" verb="GET" URI="/api/v1/namespaces/default/endpoints/kubernetes" latency="1.301658ms" userAgent="scheduler.test/v0.0.0 (linux/amd64) kubernetes/$Format" srcIP="127.0.0.1:48432" resp=404
6 I0727 09:34:20.751886  112487 httplog.go:89] "HTTP" verb="POST" URI="/apis/events.k8s.io/v1/namespaces/postfilter2-4cc9af20-f4bc-41a2-aeaf-d499d3afadcb/events" latency="2.062588ms" userAgent="scheduler.test/v0.0.0 (linux/amd64) kubernetes/$Format" srcIP="127.0.0.1:48442" resp=201
7 I0727 09:34:20.757972  112487 httplog.go:89] "HTTP" verb="POST" URI="/api/v1/namespaces/default/endpoints" latency="6.512474ms" userAgent="scheduler.test/v0.0.0 (linux/amd64) kubernetes/$Format" srcIP="127.0.0.1:48432" resp=201
8 I0727 09:34:20.766625  112487 httplog.go:89] "HTTP" verb="GET" URI="/apis/discovery.k8s.io/v1beta1/namespaces/default/endpointslices/kubernetes" latency="8.178741ms" userAgent="scheduler.test/v0.0.0 (linux/amd64) kubernetes/$Format" srcIP="127.0.0.1:48432" resp=404
9 I0727 09:34:20.766793  112487 httplog.go:89] "HTTP" verb="PATCH" URI="/api/v1/namespaces/postfilter2-4cc9af20-f4bc-41a2-aeaf-d499d3afadcb/pods/test-pod/status" latency="15.602242ms" userAgent="scheduler.test/v0.0.0 (linux/amd64) kubernetes/$Format" srcIP="127.0.0.1:48444" resp=200
10 W0727 09:34:20.766812  112487 warnings.go:67] discovery.k8s.io/v1beta1 EndpointSlice is deprecated in v1.22+, unavailable in v1.25+
11 I0727 09:34:20.766828  112487 httplog.go:89] "HTTP" verb="GET" URI="/api/v1/namespaces/postfilter2-4cc9af20-f4bc-41a2-aeaf-d499d3afadcb/pods/test-pod" latency="10.194614ms" userAgent="scheduler.test/v0.0.0 (linux/amd64) kubernetes/$Format" srcIP="127.0.0.1:48442" resp=200
12 I0727 09:34:20.767125  112487 scheduling_queue.go:812] About to try and schedule pod postfilter2-4cc9af20-f4bc-41a2-aeaf-d499d3afadcb/test-pod
13 I0727 09:34:20.767143  112487 scheduler.go:455] Attempting to schedule pod: postfilter2-4cc9af20-f4bc-41a2-aeaf-d499d3afadcb/test-pod
```

How can this happen? Although we've recently worked on a couple of PRs eliminating unneeded attempts, after the Pod is claimed as unschedulable (L3), there is still chance triggering moving the unschedulable Pod back to activeQ. In the above log, the trigger may be L4 - a service creation event.

In this PR, the filter and postFilter counter is only increased when the Pod is being tried for the first time.

Fixes TestFilter and TestPostFilter flakes in #93485.

The other flakes in #93485 need separate PR(s).

```release-note
NONE
```

[1]: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-integration-master/1287675755827826688
[2]: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-integration/1286356773338353665

### DIFF
--- a/test/integration/scheduler/framework_test.go
+++ b/test/integration/scheduler/framework_test.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
+	podutil "k8s.io/kubernetes/pkg/api/v1/pod"
 	"k8s.io/kubernetes/pkg/scheduler"
 	schedulerconfig "k8s.io/kubernetes/pkg/scheduler/apis/config"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/defaultbinder"
@@ -216,6 +217,14 @@ func (sp *ScoreWithNormalizePlugin) ScoreExtensions() framework.ScoreExtensions 
 	return sp
 }
 
+// schedAttempted returns true if the scheduler already attempted to schedule the pod.
+// Note that the logic used here works only for our integration tests because kubelet is not running,
+// and so the condition will continue to exist even if the pod eventually got scheduled.
+func schedAttempted(pod *v1.Pod) bool {
+	_, cond := podutil.GetPodCondition(&pod.Status, v1.PodScheduled)
+	return cond != nil && cond.Status == v1.ConditionFalse && cond.Reason == v1.PodReasonUnschedulable
+}
+
 // Name returns name of the plugin.
 func (fp *FilterPlugin) Name() string {
 	return filterPluginName
@@ -230,7 +239,9 @@ func (fp *FilterPlugin) reset() {
 // Filter is a test function that returns an error or nil, depending on the
 // value of "failFilter".
 func (fp *FilterPlugin) Filter(ctx context.Context, state *framework.CycleState, pod *v1.Pod, nodeInfo *framework.NodeInfo) *framework.Status {
-	fp.numFilterCalled++
+	if !schedAttempted(pod) {
+		fp.numFilterCalled++
+	}
 
 	if fp.failFilter {
 		return framework.NewStatus(framework.Error, fmt.Sprintf("injecting failure for pod %v", pod.Name))
@@ -402,7 +413,10 @@ func (pp *PostFilterPlugin) Name() string {
 }
 
 func (pp *PostFilterPlugin) PostFilter(ctx context.Context, state *framework.CycleState, pod *v1.Pod, _ framework.NodeToStatusMap) (*framework.PostFilterResult, *framework.Status) {
-	pp.numPostFilterCalled++
+	if !schedAttempted(pod) {
+		pp.numPostFilterCalled++
+	}
+
 	nodeInfos, err := pp.fh.SnapshotSharedLister().NodeInfos().List()
 	if err != nil {
 		return nil, framework.NewStatus(framework.Error, err.Error())


ahg-g:
This is honestly not clear. 

How about we check that num of calls is `x` when the pod is expected to schedule from the first time, and greater than or equal to `x` when not? here is an example for TestPostFilter.

```
if tt.rejectFilter {
  if err = wait.Poll(10*time.Millisecond, 10*time.Second, podUnschedulable(testCtx.ClientSet, pod.Namespace, pod.Name)); err != nil {
    t.Errorf("Didn't expect the pod to be scheduled.")
  }
  if filterPlugin.numFilterCalled != tt.expectFilterNumCalled {
    t.Errorf("Expected the filter plugin to be called %v times, but got %v.", tt.expectFilterNumCalled, filterPlugin.numFilterCalled)
  }
  if postFilterPlugin.numPostFilterCalled != tt.expectPostFilterNumCalled {
    t.Errorf("Expected the postfilter plugin to be called %v times, but got %v.", tt.expectPostFilterNumCalled, postFilterPlugin.numPostFilterCalled)
  }
} else {
  if err = testutils.WaitForPodToSchedule(testCtx.ClientSet, pod); err != nil {
    t.Errorf("Expected the pod to be scheduled. error: %v", err)
  }
  if filterPlugin.numFilterCalled < tt.expectFilterNumCalled {
    t.Errorf("Expected the filter plugin to be called at least %v times, but got %v.", tt.expectFilterNumCalled, filterPlugin.numFilterCalled)
  }
  if postFilterPlugin.numPostFilterCalled < tt.expectPostFilterNumCalled {
    t.Errorf("Expected the postfilter plugin to be called at least %v times, but got %v.", tt.expectPostFilterNumCalled, postFilterPlugin.numPostFilterCalled)
  }
}

```

The reason I prefer the above approach is because it communicates better the expectations for each test case.

Huang-Wei:
> How about we check that num of calls is x when the pod is expected to schedule from the first time, and greater than or equal to x when not?

I thought of that as well. That would be safe for sure. However, it may not catch some failure case. For example, we expect the `Filter()` plugin to be called at least twice (suppose there is only one node) - one time in Filter() plugin by itself, and the other time by PostFitler(). However, a failure case may fail to call Filter() during PostFilter(), and Filter() plugin happen to be called twice both by itself (the second call occurs at the 2nd scheduling attempt). So this case will also pass.

Huang-Wei:
@ahg-g what do you think of https://github.com/kubernetes/kubernetes/pull/93490#issuecomment-664764234.

Another option is to pass on a StateData from PostFilter plugin to FilterPlugin plugin, so that we can hold a separate counter in FilterPlugin to record how many times it's called by PostFilter, i.e., not mixed with the regular counter.

test/integration/scheduler/framework_test.go
```diff
@@ -215,6 +216,12 @@ func (sp *ScoreWithNormalizePlugin) ScoreExtensions() framework.ScoreExtensions
 	return sp
 }
 
+// schedAttempted returns true if the given Pod has been attempted scheduling; otherwise return true.
```
ahg-g, line 219:
```suggestion
// schedAttempted returns true if the scheduler already attempted to schedule the pod. Note that the logic used here works only for our integration tests because kubelet is not run, and so the condition will continue to exist even if the pod eventually got scheduled.
```

Can you confirm if the above is accurate?

Huang-Wei: It looks good to me.

PR #93491, opened 2020-07-28
Improving logging in EndpointSlice e2e tests

robscott:
**What type of PR is this?**
/kind cleanup
/kind failing-test

When these tests failed it was unclear that the reason for the failure could have been more EndpointSlices than expected. It was also unclear what EndpointSlices were actually found when that occurred. This fixes both of those issues.

Despite my best efforts, I have been unable to reproduce #93374. This PR will ensure that the next test failure provides more helpful information.

```release-note
NONE
```

/cc @hasheddan 
/assign @liggitt

### DIFF
--- a/test/e2e/network/endpointslice.go
+++ b/test/e2e/network/endpointslice.go
@@ -18,6 +18,7 @@ package network
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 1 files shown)


test/e2e/network/endpointslice.go
```diff
@@ -562,3 +565,17 @@ func createServiceReportErr(cs clientset.Interface, ns string, service *v1.Servi
 	framework.ExpectNoError(err, "error deleting Service")
 	return svc
 }
+
+func logEndpointSlice(epSlice discoveryv1beta1.EndpointSlice) {
+	ports := []string{}
+	for _, port := range epSlice.Ports {
+		ports = append(ports, fmt.Sprint(*port.Port))
```
liggitt, line 572: this can be null, don't dereference without checking?

test/e2e/network/endpointslice.go
```diff
@@ -562,3 +565,17 @@ func createServiceReportErr(cs clientset.Interface, ns string, service *v1.Servi
 	framework.ExpectNoError(err, "error deleting Service")
 	return svc
 }
+
+func logEndpointSlice(epSlice discoveryv1beta1.EndpointSlice) {
```
liggitt, line 569: would it be better to marshal to JSON and log the full content? if there's a diff, it might not be in the few fields you selected to log

robscott: I'd initially thought that would just end up polluting the logs but I think it's entirely worthwhile here, updated to use that approach, thanks!

test/e2e/network/endpointslice.go
robscott: Updated, thanks!

PR #93492, opened 2020-07-28
bind metadata proxy to 0.0.0.0

zshihang:
**What type of PR is this?**
/kind failing-test

fix test `pull-kubernetes-e2e-gci-gce-ipvs` failure in https://github.com/kubernetes/kubernetes/pull/92938

```release-note
None
```


zshihang:
add `! -i eth0` to block non-pod and non-local traffic

mikedanese:
> add `! -i eth0` to filter out non-pod and non-local traffic

I don't see how this could hurt, but I though that this was covered by putting the rule in the PREROUTING chain. Presumably this chain is only hit for forwarded packets.

PR #93495, opened 2020-07-28
Fix an integration test flake on NodeAfffinity ScorePlugin

Huang-Wei:
**What type of PR is this?**

/kind flake
/sig scheduling
/priority critical-urgent

The behavior of Score plugins are not deterministic as the Scores of a plugin can be easily diluted by others.

This PR constructs a scheduler profile by disabling all Score plugins except for `NodeAffiniy`, so the final score is only impacted by `NodeAffiniy` Score plugin.

Fixes NodeAffinity flake in #93485.

```release-note
NONE
```


test/integration/scheduler/priorities_test.go
```diff
@@ -36,7 +39,29 @@ import (
...
+	testCtx := testutils.InitTestSchedulerWithOptions(
+		t,
+		testutils.InitTestMaster(t, "node-affinity", nil),
+		false,
+		nil,
+		0,
+		scheduler.WithProfiles(prof),
+	)
+	testutils.SyncInformerFactory(testCtx)
+	go testCtx.Scheduler.Run(testCtx.Ctx)
```
ahg-g, line 64: I expect all other score tests have the same issue, can we wrap this around a function and use it everywhere?:

```
func initTestWithPlugins(name string, enabledPlugins []schedulerconfig.Plugin)  *TestContext {
        prof := schedulerconfig.KubeSchedulerProfile{
		SchedulerName: v1.DefaultSchedulerName,
		Plugins: &schedulerconfig.Plugins{
			Score: &schedulerconfig.PluginSet{
				Enabled:enabledPlugins,
				Disabled: []schedulerconfig.Plugin{
					{Name: "*"},
				},
			},
		},
	}
	testCtx := testutils.InitTestSchedulerWithOptions(
		t,
		testutils.InitTestMaster(t, name, nil),
		false,
		nil,
		0,
		scheduler.WithProfiles(prof),
	)
	testutils.SyncInformerFactory(testCtx)
	go testCtx.Scheduler.Run(testCtx.Ctx)
        return testCtx
}
```

ahg-g:
> The behavior of Score plugins are not deterministic as the Scores of a plugin can be easily diluted by others. In the flake test, an explicit weight: 20 can only mitigate the flake.

Where was the weight of 20 set? or are you mentioning that as an example?

test/integration/scheduler/priorities_test.go
Huang-Wei: sure, will update.

Huang-Wei:
Ah, I misread the NodeAffinity weight as the plugin weight... Updated the wordings.

ahg-g:
Thanks, please squash!

Huang-Wei:
Verify job timed out.

/retest

PR #92991, opened 2020-07-11
Make sure pod tested for eviction with PDB is not scheduled for deletion

hasheddan:
**What type of PR is this?**

/kind flake

**What this PR does / why we need it**:

In #91342 attempting to evict a Pod with a DeletionTimestamp caused
checking of PDBs to be ignored due to the fact that a Pod scheduled for
deletion should not be factored into a disruption budget. However, PDB
eviction tests currently will sometimes select a Pod already scheduled
for deletion, expecting that attempting to evict it will conflict with
the PDB. This updates those tests to make sure a Pod with deletion
timestamp is not selected for eviction when it is intended to violate a
PDB.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

**Which issue(s) this PR fixes**:

Fixes #92977 

**Special notes for your reviewer**:

This flaky test currently affects a number of jobs including:
https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-default
https://testgrid.k8s.io/sig-release-master-blocking#gce-ubuntu-master-containerd
https://testgrid.k8s.io/sig-release-master-blocking#kind-master-parallel
https://testgrid.k8s.io/sig-release-master-blocking#kind-ipv6-master-parallel
https://testgrid.k8s.io/sig-release-master-informing#gce-master-scale-correctness
https://testgrid.k8s.io/sig-release-master-informing#gce-ubuntu-master-default

See increase in failures since #91342 here: https://storage.googleapis.com/k8s-gubernator/triage/index.html?test=DisruptionController%20should%20block%20an%20eviction%20until%20the%20PDB%20is%20updated%20to%20allow%20it

/cc @liggitt @BenTheElder @hakman 
/sig apps
/cc @kubernetes/ci-signal 

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #93501, opened 2020-07-28
Fix nil panic in PDB tests

tnozicka:
**What type of PR is this?**
/kind failing-test

DeletionTimestamp is a pointer so accessing it when nil will panic

fixes https://github.com/kubernetes/kubernetes/pull/92991/files#r461425286

```release-note
NONE
```

/cc @hasheddan @BenTheElder @kubernetes/sig-apps-pr-reviews

### DIFF
--- a/test/e2e/apps/disruption.go
+++ b/test/e2e/apps/disruption.go
@@ -491,7 +491,7 @@ func waitForPodsOrDie(cs kubernetes.Interface, ns string, n int) {
 		ready := 0
 		for i := range pods.Items {
 			pod := pods.Items[i]
-			if podutil.IsPodReady(&pod) && pod.ObjectMeta.DeletionTimestamp.IsZero() {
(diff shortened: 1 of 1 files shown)


hasheddan (changes requested):
Calling a method on a nil pointer in Go is not a panic as long as the nil pointer is not dereferenced in the method. `IsZero()` checks for nil so I don't think this change is needed here.

liggitt:
agree, IsZero() covers both the nil and zero-value case:

https://github.com/kubernetes/kubernetes/blob/8d74486a6a1a8fb2246bd89faf1746393135a463/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/time.go#L60-L66

/close

tnozicka:
interesting, it did not know that is special cased there :) kind of used to not calling methods on nil pointers

PR #93503, opened 2020-07-28
bugfix(cloud-controller-manager): flush logs before exit.

hustclf:
**What type of PR is this?**
kind/bug

a small optimization:  flush logs before cloud-controller-manager exit.

```release-note
NONE
```

### DIFF
--- a/cmd/kube-apiserver/apiserver.go
+++ b/cmd/kube-apiserver/apiserver.go
@@ -34,6 +34,9 @@ func main() {
 
 	command := app.NewAPIServerCommand()
 
+	exitCode := 0
+	defer func() { os.Exit(exitCode) }()
+
 	// TODO: once we switch everything over to Cobra commands, we can go back to calling
 	// utilflag.InitFlags() (by removing its pflag.Parse() call). For now, we have to set the
 	// normalize func and add the go flag set by hand.
@@ -42,6 +45,7 @@ func main() {
 	defer logs.FlushLogs()
 
 	if err := command.Execute(); err != nil {
-		os.Exit(1)
+		exitCode = 1
+		return
 	}
 }


hustclf:
I signed it

neolit123:
hi, the change seems good to me, but we are in code freeze for 1.19:
https://github.com/kubernetes/sig-release/tree/master/releases/release-1.19

once code freeze is lifted, reviewer can LGTM / approve it.

/kind bug
/priority backlog

cmd/cloud-controller-manager/controller-manager.go
```diff
@@ -48,9 +48,9 @@ func main() {
 	// normalize func and add the go flag set by hand.
 	// utilflag.InitFlags()
 	logs.InitLogs()
-	defer logs.FlushLogs()
```
cheftako, line 51: I understand that os.Exit is causing the defer to be skipped as this is main and os.Exit in main means main never returns. However it is not guaranteed that we will go down the path leading to the os.Exit and for those cases we are not not flushing the logs. The best solution would seem to be leave this defer here and add a comment explaining why we have both.

hustclf: Yes,  when `command.Execute()` not return error,  `FlushLogs()` will not run.

but keep two `logs.FlushLogs()` may not be the best solution.  I take the `os.Exit()` in defer can handle it.

Please review it again. Thanks.

cheftako: defer is LIFO so that should work.

cmd/cloud-controller-manager/controller-manager.go
```diff
@@ -51,6 +54,7 @@ func main() {
 	defer logs.FlushLogs()
 
 	if err := command.Execute(); err != nil {
-		os.Exit(1)
+		exitCode = 1
```
cheftako, line 57: Sadly this doesn't work (the os.Exit part with this) because the deferred functions parameters are evaluated immediately. See https://play.golang.org/p/MuR_9TV-twr .  This also demonstrated that a referenced variable rather than passed parameter can work. (Beware using this in a goroutine as its very racy)

hustclf: Yes buddy, you are right.

hustclf: Call `os.Exit(exitCode)` in anonymous function should work.

I udpate the code , and here is an example:
```
package main

import (
	"os"
)

func main() {
	exitCode := 0
	defer func() { os.Exit(exitCode) }()
	exitCode = 1
}
```

cmd/cloud-controller-manager/controller-manager.go
```diff
@@ -51,6 +54,7 @@ func main() {
...
+		return
 	}
```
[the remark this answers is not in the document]
hustclf: Yes, It is just a small change.

flush log before exit seems reasonable logically, but I am not sure if there are useful logs to be flushed. I will check again.

Thanks.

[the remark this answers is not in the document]
cheftako: Personally when a process "crashes", I find the last things to be logged the most interesting. So I at least am a fan of trying to flush the logs on the way out.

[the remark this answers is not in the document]
hustclf: I think flush before exit is necessary because the default klog flush interval is 5s, according to
`vendor/k8s.io/component-base/logs/logs.go:57`, which may lose logs.

There are 7 components to change, should we change it all in current mr ?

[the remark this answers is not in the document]
cheftako: The more components in the PR, the harder it is get all the sign-offs before your PR is stale. 
I would suggest merging this and doing the other components as a follow up.

hustclf:
Hi @cheftako,  I found the pr keep in inactivity for long time.  Should we still continue the pr, or just close it?

cmd/kube-apiserver/apiserver.go
```diff
@@ -42,6 +45,7 @@ func main() {
 	defer logs.FlushLogs()
 
 	if err := command.Execute(); err != nil {
-		os.Exit(1)
+		exitCode = 1
+		return
```
cheftako, line 49: Not sure I understand why we need a return here since we're about to finish the execution of main().

cheftako:
> Hi Walter Fender, I found the pr keep in inactivity for long time. Should we still continue the pr, or just close it?

Making sure we flush logs prior to an unusual exit still seems valuable. I'm all for merging it. Not sure why it wasn't merged before.

sttts:
The change looks good. But I see this in many other binaries in the repo. Can we fix this everywhere?

Issue #81368, opened 2019-08-13
TestVolumeProvision is flaky 


liggitt:
**Which jobs are failing**:

pull-kubernetes-integration

**Which test(s) are failing**:

TestVolumeProvision

Example: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/81330/pull-kubernetes-integration/1161315691412852736/

```
--- FAIL: TestVolumeProvision (12.16s)
    volume_binding_test.go:1149: Provisoning annotaion on PVC volume-schedulingfc108048-dc1f-4999-8323-f3302c9eb3ca/pvc-w-canbind not bahaviors as expected: PVC volume-schedulingfc108048-dc1f-4999-8323-f3302c9eb3ca/pvc-w-canbind not expected to be provisioned, but found selected-node annotation
    volume_binding_test.go:1191: PV pv-w-canbind phase not Bound, got Available
```

**Since when has it been failing**:

Not sure, [triage board](https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&test=TestVolumeProvision) and [test grid](https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-integration&sort-by-flakiness=) aren't displaying any results

/sig storage

msau42:
/assign @msau42 @cofyc 
/remove-lifecycle rotten
Still seeing this, although not very frequently. Could indicate a race.

cofyc:
latest failures:

- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-integration-master/1287138465619644420
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-integration-master/1287061710439452674
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-integration-master/1286724218758631424

cofyc:
take `https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-integration-master/1287138465619644420` as an example

the test fails b/c `selected-node` is set to PVC `pvc-w-pv-prebound`, which is not expected.

```
[...]

PR #93511, opened 2020-07-28
Hold Pod in cache until all other cleanup work is completed

Huang-Wei:
**What type of PR is this?**

/kind flake
/sig scheduling
/priority critical-urgent

The PreBind runs in a goroutine, and unfortunately, the `assumedPod` gets removed from cache **prior to** clean-up work (e.g., Unreserve) gets finished:

https://github.com/kubernetes/kubernetes/blob/498b1e227981e1fe317d91143617436e81c27775/pkg/scheduler/scheduler.go#L577-L588

This causes tricky flakes. The flaky test was built on a buffered channel (len=3) to receive Unreserve events. However, it's possible that when the binding goroutine of the previous scheduling attempt is waiting for Unreserve to finish (L585), another scheduling attempt starts, and it proceeds to also run Unreserve. The 2 Unreserve running together will compete to fill in the buffered channel (len=3), which causes L586 (above) in either goroutines unreachable. And in the [flake log], there are 32 binding goroutines running at the same time...

This PR tries to fix the problem by:

- Move "ForgetPod" after "RunReservePluginsUnreserve", so that the cache would hold the pod to avoid it's being retried simutaneously until Unreserve is completed.
- Move "assume" ahead of "RunReservePluginsReserve". This is based on the fact that "ForgetPod" is the last step of failure path, so "assume" should be reversly treated as the first step. The current failure path is like this:
    ```
    assume -> reserve -> unreserve -> forgetPod -> recordingFailure
    ``` 
- Make subtests of `TestReservePluginUnreserve` stateless

Fixes TestReservePluginUnreserve flake in #93485.

```release-note
NONE
```

[flake log]: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-integration-master/1287660403844714499

### DIFF
--- a/test/integration/scheduler/framework_test.go
+++ b/test/integration/scheduler/framework_test.go
@@ -971,57 +971,6 @@ func TestPrebindPlugin(t *testing.T) {
 // tests that the order of invocation of Unreserve operation is executed in the
 // reverse order of invocation of the Reserve operation.
 func TestReservePluginUnreserve(t *testing.T) {
-	numReservePlugins := 3
-	pluginInvokeEventChan := make(chan pluginInvokeEvent, numReservePlugins)
-
-	preBindPlugin := &PreBindPlugin{
-		failPreBind: true,
-	}
-	var reservePlugins []*ReservePlugin
-	for i := 0; i < numReservePlugins; i++ {
-		reservePlugins = append(reservePlugins, &ReservePlugin{
-			name:                  fmt.Sprintf("%s-%d", reservePluginName, i),
-			pluginInvokeEventChan: pluginInvokeEventChan,
-		})
-	}
-
-	registry := frameworkruntime.Registry{
-		// TODO(#92229): test more failure points that would trigger Unreserve in
-		// reserve plugins than just one pre-bind plugin.
-		preBindPluginName: newPlugin(preBindPlugin),
-	}
-	for _, pl := range reservePlugins {
-		registry[pl.Name()] = newPlugin(pl)
-	}
-
-	// Setup initial reserve and prebind plugin for testing.
-	prof := schedulerconfig.KubeSchedulerProfile{
-		SchedulerName: v1.DefaultSchedulerName,
-		Plugins: &schedulerconfig.Plugins{
-			Reserve: &schedulerconfig.PluginSet{
-				// filled by looping over reservePlugins
-			},
-			PreBind: &schedulerconfig.PluginSet{
-				Enabled: []schedulerconfig.Plugin{
-					{
-						Name: preBindPluginName,
-					},
-				},
-			},
-		},
-	}
-	for _, pl := range reservePlugins {
-		prof.Plugins.Reserve.Enabled = append(prof.Plugins.Reserve.Enabled, schedulerconfig.Plugin{
-			Name: pl.Name(),
-		})
-	}
-
-	// Create the master and the scheduler with the test plugin set.
-	testCtx := initTestSchedulerForFrameworkTest(t, testutils.InitTestMaster(t, "reserve-plugin-unreserve", nil), 2,
-		scheduler.WithProfiles(prof),
-		scheduler.WithFrameworkOutOfTreeRegistry(registry))
-	defer testutils.CleanupTest(t, testCtx)
-
 	tests := []struct {
 		name             string
 		failReserve      bool
@@ -1044,6 +993,57 @@ func TestReservePluginUnreserve(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
+			numReservePlugins := 3
+			pluginInvokeEventChan := make(chan pluginInvokeEvent, numReservePlugins)
+
+			preBindPlugin := &PreBindPlugin{
+				failPreBind: true,
+			}
+			var reservePlugins []*ReservePlugin
+			for i := 0; i < numReservePlugins; i++ {
+				reservePlugins = append(reservePlugins, &ReservePlugin{
+					name:                  fmt.Sprintf("%s-%d", reservePluginName, i),
+					pluginInvokeEventChan: pluginInvokeEventChan,
+				})
+			}
+
+			registry := frameworkruntime.Registry{
+				// TODO(#92229): test more failure points that would trigger Unreserve in
+				// reserve plugins than just one pre-bind plugin.
+				preBindPluginName: newPlugin(preBindPlugin),
+			}
+			for _, pl := range reservePlugins {
+				registry[pl.Name()] = newPlugin(pl)
+			}
+
+			// Setup initial reserve and prebind plugin for testing.
+			prof := schedulerconfig.KubeSchedulerProfile{
+				SchedulerName: v1.DefaultSchedulerName,
+				Plugins: &schedulerconfig.Plugins{
+					Reserve: &schedulerconfig.PluginSet{
+						// filled by looping over reservePlugins
+					},
+					PreBind: &schedulerconfig.PluginSet{
+						Enabled: []schedulerconfig.Plugin{
+							{
+								Name: preBindPluginName,
+							},
+						},
+					},
+				},
+			}
+			for _, pl := range reservePlugins {
+				prof.Plugins.Reserve.Enabled = append(prof.Plugins.Reserve.Enabled, schedulerconfig.Plugin{
+					Name: pl.Name(),
+				})
+			}
+
+			// Create the master and the scheduler with the test plugin set.
+			testCtx := initTestSchedulerForFrameworkTest(t, testutils.InitTestMaster(t, "reserve-plugin-unreserve", nil), 2,
+				scheduler.WithProfiles(prof),
+				scheduler.WithFrameworkOutOfTreeRegistry(registry))
+			defer testutils.CleanupTest(t, testCtx)
+
 			preBindPlugin.failPreBind = test.failPreBind
 			if test.failReserve {
 				reservePlugins[test.failReserveIndex].failReserve = true
@@ -1079,11 +1079,6 @@ func TestReservePluginUnreserve(t *testing.T) {
 						t.Errorf("reservePlugins[%d].numUnreserveCalled = %d, want 0", i, pl.numUnreserveCalled)
 					}
 				}
-			}
-
-			preBindPlugin.reset()
-			for _, pl := range reservePlugins {
-				pl.reset()
 			}
 			testutils.CleanupPods(testCtx.ClientSet, t, []*v1.Pod{pod})
 		})
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -501,16 +501,6 @@ func (sched *Scheduler) scheduleOne(ctx context.Context) {
 	// This allows us to keep scheduling without waiting on binding to occur.
 	assumedPodInfo := podInfo.DeepCopy()
 	assumedPod := assumedPodInfo.Pod
-
-	// Run the Reserve method of reserve plugins.
-	if sts := prof.RunReservePluginsReserve(schedulingCycleCtx, state, assumedPod, scheduleResult.SuggestedHost); !sts.IsSuccess() {
-		sched.recordSchedulingFailure(prof, assumedPodInfo, sts.AsError(), SchedulerError, "")
-		metrics.PodScheduleError(prof.Name, metrics.SinceInSeconds(start))
-		// trigger un-reserve to clean up state associated with the reserved Pod
-		prof.RunReservePluginsUnreserve(schedulingCycleCtx, state, assumedPod, scheduleResult.SuggestedHost)
-		return
-	}
-
 	// assume modifies `assumedPod` by setting NodeName=scheduleResult.SuggestedHost
 	err = sched.assume(assumedPod, scheduleResult.SuggestedHost)
 	if err != nil {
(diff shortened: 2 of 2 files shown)


alculquicondor:
if we go with the logic that "assume" is a reserve operation, we should go: reserve->assume->forget->unreserve (like in master)
or: assume->reserve->unreserve->forget (another option that we could consider, but not sure if it works well with volume binding).

Recall that we run the Reserve plugins's Unreserve in the opposite order.

Isn't the bug just in the integration test implementation? We shouldn't be relying in a single channel.

alculquicondor:
cc @cofyc

Huang-Wei:
> or: assume->reserve->unreserve->forget (another option that we could consider, but not sure if it works well with volume binding).

This sounds good to me - if we think of "assume" as a special "reserve". @cofyc WDYT?

> Recall that we run the Reserve plugins's Unreserve calls in the opposite order.

Correct. I didn't touch that piece.

> Isn't the bug just in the integration test implementation? We shouldn't be relying in a single channel.

Yes or no. The users may also make similar assumptions I guess?

adtac:
What I don't fully understand is, why is another scheduler attempt being made when there's no indication that the existing schedule attempt is complete? Perhaps I've misunderstood something, but isn't `sched.recordSchedulingFailure` used to let the scheduler know that a pod is unschedulable so that it can be put back into the queue? If so, why would another attempt begin for this pod before `sched.recordSchedulingFailure` is called from within the goroutine?

alculquicondor:
Is it a single pod being scheduled?

adtac:
Yes.

alculquicondor:
Then different pods' scheduling attempts are competing for the same channel. Recall that binding is a separate go routine.

alculquicondor:
synced with @adtac offline. He's going to investigate the fix in the test.

I fear that changing the order in which reserve and assume happen might trigger other problems. Plugins should be more careful in their implementation.

Huang-Wei:
> Perhaps I've misunderstood something, but isn't sched.recordSchedulingFailure used to let the scheduler know that a pod is unschedulable so that it can be put back into the queue?

It's because in the current logic, the "assumedPod" gets removed (`ForgetPod`) from cache first, and then it proceeds to run Unreserve, so during that, if the Pod gets updated somehow (in the test, it's because of the forced recync. We can disable that, but I guess it's just mitigation), we don't have the cache entry to block the update-event trggered scheduling attempt. So the 2nd attempt would compete with the original ongoing Unserve on the shared buffer-channel, and both will hang due to the limited capacity.

Huang-Wei:
> I fear that changing the order in which reserve and assume happen might trigger other problems. Plugins should be more careful in their implementation.

@alculquicondor Do you want me to append a commit to move "assumePod" prior to "RunReservePluginsReserve"?

alculquicondor:
I'm more in favor of keeping the current behavior. At least until we get some input from Yecheng.

adtac:
Two points:

1. creating a separate event channel for each `t.Run` invocation is probably good practice even though it won't help with this flaky test.

2. @Huang-Wei can you explain why the order of recordSchedulingFailure and RunReservePluginsUnreserve is reversed compared to master? I understand why the pair of them must occur before the cache forget operation, but why is the order within the pair reversed?

alculquicondor:
To be more specific, @cofyc do you see any problems with the following order:

```assume->reserve->unreserve->forget```

**Comment by cofyc** [comment]
> > or: assume->reserve->unreserve->forget (another option that we could consider, but not sure if it works well with volume binding).
> 
> This sounds good to me - if we think of "assume" as a special "reserve". @cofyc WDYT?
> 

it should not have any problem with volume binding.

IIUC, in asynchronous binding goroutine, if the pod is forgotten by the scheduler cache, then it can be retried by the future scheduler cycles?

Then I think this is better compared to current behavior. The pod should be released (forgot in scheduler cache) after all cleanups are finished.

**Comment by cofyc** [comment]
> To be more specific, @cofyc do you see any problems with the following order:
> 
> `assume->reserve->unreserve->forget`

I don't see any problems for now.

**Comment by Huang-Wei** [comment]
> 1. creating a separate event channel for each t.Run invocation is probably good practice even though it won't help with this flaky test.

That needs to be updated for sure, but not key to fix the issue. We can come up with a general PR fixing all the non-stateless tests.

> 2. @Huang-Wei can you explain why the order of recordSchedulingFailure and RunReservePluginsUnreserve is reversed compared to master? I understand why the pair of them must occur before the cache forget operation, but why is the order within the pair reversed?

As I explained in the description section, it's optional. I just felt that given `recordSchedulingFailure` is already a fact, and running Unreserve may take time, so it makes sense to move it ahead of Unreserve, which can make other actors that rely on a unschedulable status act fast accordingly. And also, some existing paths such as reserve<->unreserve and assume<->unreserve is already putting `recordSchedulingFailure` ahead of unreserve.
But again, it's optional.

**Comment by Huang-Wei** [comment]
> IIUC, in asynchronous binding goroutine, if the pod is forgotten by the scheduler cache, then it can be retried by the future scheduler cycles?

Yes. In any failure path, the failed Pod will be pathed with an (updated) unschedualbe status. So that will trigger the failed pod to be re-enqueued.

**Comment by alculquicondor** [comment]
So `recordSchedulingFailure` should happen at the end.

Are you sure the updated scheduled status causes the pod to be put back into active?

**Comment by Huang-Wei** [comment]
@adtac Due to https://github.com/kubernetes/kubernetes/pull/93511#issuecomment-665316132, I think we still need to put `recordSchedulingFailure` after `ForgetPod`. So I'd propose the order to be:

- Happy path: assume ->reserve->unreserve->forget
- Negative path: assume->reserve->unreserve->forget->recordSchedulingFailure

@alculquicondor @cofyc WDYT?

**Comment by Huang-Wei** [comment]
> Are you sure the updated scheduled status causes the pod to be put back into active?

Not necessarily activeQ, but should be queued anyhow:

https://github.com/kubernetes/kubernetes/blob/22329fbe1338934d5482d91c5dec18851c7ddeaf/pkg/scheduler/eventhandlers.go#L264

**Comment by alculquicondor** [comment]
> Not necessarily activeQ, but should be queued anyhow:

The pod is put back into unschedulable queue at the end of the scheduling/binding cycles in case of error as well.

> Happy path: assume ->reserve->unreserve->forget

I thought unreserve and forget is only called in the unhappy path.

**Comment by Huang-Wei** [comment]
> The pod is put back into unschedulable queue at the end of the scheduling/binding cycles in case of error as well.

Yes, I appended a commit to keep recordingFailure at the last step of failure path.

> I thought unreserve and forget are only called in the unhappy path.

Yes, updated the wordings.

**Comment by Huang-Wei** [comment]
> Yes, I appended a commit to keep recordingFailure at the last step of failure path.

@alculquicondor @adtac @cofyc PTAL.

**Comment by cofyc** [comment]
> * r

SGTM

**Comment by cofyc** [comment]
> @adtac Due to [#93511 (comment)](https://github.com/kubernetes/kubernetes/pull/93511#issuecomment-665316132), I think we still need to put `recordSchedulingFailure` after `ForgetPod`. So I'd propose the order to be:
> 
> * assume->reserve->unreserve->forget->recordSchedulingFailure.
> 
> @alculquicondor @cofyc WDYT?

SGTM

**Comment by cofyc** [comment]
https://github.com/kubernetes/kubernetes/blob/22329fbe1338934d5482d91c5dec18851c7ddeaf/pkg/scheduler/scheduler.go#L590-L595

when the binding operation fails, the `unreserve` functions are called after the `ForgetPod` (called in the `sched.bind`). should we update this too?

**Comment by cofyc** [comment]
https://github.com/kubernetes/kubernetes/blob/22329fbe1338934d5482d91c5dec18851c7ddeaf/pkg/scheduler/scheduler.go#L505-L527

I guess `sched.assume` must be called before calling `reserve` functions. and no `unreserve` functions need to be called when it fails.

`assume/forget` works like a special `reserve` plugin which is always the first in the plugin list.

**Comment by Huang-Wei** [comment]
> I guess `sched.assume` must be called before calling `reserve` functions. and no `unreserve` functions need to be called when it fails.

@cofyc that makes sense to me. But maybe in a separate PR, and companies with a test.

**Comment by alculquicondor** [comment]
PR description is still outdated.

**Comment by alculquicondor** [comment]
/lgtm cancel

Let's do assume->reserve

**Comment by Huang-Wei** [comment]
@alculquicondor @adtac Please take a look at the latest 2 commits:

- one commit moves assume() ahead of RunReservePluginsReserve()
- the other commit makes subtests stateless. Although it increases the overall running time a bit, it still worth it IMO, as based on the updated logic, the **shared** buffered channel may have already been re-filled up (due to retried goroutines) when a later subtest starts

@ahg-g would you mind adding it to 1.19 release? I will squash later.

File: test/integration/scheduler/framework_test.go [hunk 1]
```diff
@@ -1080,12 +1080,11 @@ func TestReservePluginUnreserve(t *testing.T) {
 					}
 				}
 			}
-
-			preBindPlugin.reset()
-			for _, pl := range reservePlugins {
-				pl.reset()
-			}
-			testutils.CleanupPods(testCtx.ClientSet, t, []*v1.Pod{pod})
+			// preBindPlugin.reset()
```
alculquicondor, line 1083: Is this comment intentional?

alculquicondor (commented):
code lgtm. @adtac, PTAL at the updated test.

test/integration/scheduler/framework_test.go
adtac: Should probably be removed since each set of plugins is used only once.

test/integration/scheduler/framework_test.go
```diff
@@ -1080,12 +1080,11 @@ func TestReservePluginUnreserve(t *testing.T) {
...
+			// for _, pl := range reservePlugins {
+			// 	pl.reset()
+			// }
+			// testutils.CleanupPods(testCtx.ClientSet, t, []*v1.Pod{pod})
```
adtac, line 1087: This shouldn't be commented out, however.

adtac:
IIRC there were other integration tests that shared a registry, KubeSchedulerProfile, etc. too; I followed the pattern in those tests when writing TestReservePluginUnreserve. While the shared channel is the primary source of test failures, not having ANY shared state might be a good idea in other tests too.

test/integration/scheduler/framework_test.go
Huang-Wei: Yes, reset() are not necessary as each subtest is using its own plugin.

test/integration/scheduler/framework_test.go
Huang-Wei: Yes, it should be kept.

Huang-Wei:
@adtac It'd much appreciate if you could craft a followup PR to clean up those tests that hold shared state.

alculquicondor (commented):
/lgtm

adtac:
/lgtm

(but I don't think my /lgtm does anything)

Huang-Wei:
/hold

Need to take a look at the volumebinding failure: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93511/pull-kubernetes-integration/1288528886769913856

adtac:
>It'd much appreciate if you could craft a followup PR to clean up those tests that hold shared state.

Will try to do that :+1:

Huang-Wei:
Listing records of TestVolumeBinding:

- Pass: [P1] 
- Failure: [F1]

[P1]: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93511/pull-kubernetes-integration/1288542065629597697
[F1]: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93511/pull-kubernetes-integration/1288528886769913856

cofyc:
> https://github.com/kubernetes/kubernetes/blob/22329fbe1338934d5482d91c5dec18851c7ddeaf/pkg/scheduler/scheduler.go#L590-L595
> 
> when the binding operation fails, the `unreserve` functions are called after the `ForgetPod` (called in the `sched.bind`). should we update this too?

just a reminder that we should run `unreserve` functions before the `ForgetPod` when the binding operation fails for consistency. refactor in a follow-up PR?

Huang-Wei:
> just a reminder that we should run `unreserve` functions before the `ForgetPod` when the binding operation fails for consistency. refactor in a follow-up PR?

Yes, we should do that. Except for that, do you think the failures in https://github.com/kubernetes/kubernetes/pull/93511#issuecomment-665861157 is related to the new changes of this PR?

cofyc:
> > just a reminder that we should run `unreserve` functions before the `ForgetPod` when the binding operation fails for consistency. refactor in a follow-up PR?
> 
> Yes, we should do that. Apart from that, do you think the failures in [#93511 (comment)](https://github.com/kubernetes/kubernetes/pull/93511#issuecomment-665861157) is related to the new changes of this PR?

I don't think so. Similar failures happen in master and other PRs too. https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&test=TestVolumeProvision

I'll track this flaky test in https://github.com/kubernetes/kubernetes/issues/81368.

Huang-Wei:
Thanks @cofyc ! How about `TestVolumeBinding`, I observed 2 failures (F1 & F3) so far in https://github.com/kubernetes/kubernetes/pull/93511#issuecomment-665861157.

cofyc:
> Thanks @cofyc ! How about `TestVolumeBinding`, I observed 2 failures (F1 & F3) so far in [#93511 (comment)](https://github.com/kubernetes/kubernetes/pull/93511#issuecomment-665861157).

I think it's unrelated.

Huang-Wei:
> I think it's unrelated.

Agree. I spot it [flaked](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93550/pull-kubernetes-integration/1288687271234506755) in an irrelevant PR (#93550)

So
/hold cancel

PR #93516, opened 2020-07-28
Create new slice to avoid modify original fake data

soulxu:
When create fake data for the nodeTree unittests, The 'append' is invoked
on the common fake data set. That makes the unittests is running with unexpected
fake data after that.

**What type of PR is this?**
/kind cleanup

The common fake data set is modified through each test, which makes the unit test is running with unexpected fake data.

None

None

None

### DIFF
--- a/pkg/scheduler/internal/cache/node_tree_test.go
+++ b/pkg/scheduler/internal/cache/node_tree_test.go
@@ -428,14 +428,14 @@ func TestNodeTreeMultiOperations(t *testing.T) {
 		},
 		{
 			name:           "add more nodes to an exhausted zone",
-			nodesToAdd:     append(allNodes[4:9], allNodes[3]),
+			nodesToAdd:     append(allNodes[4:9:9], allNodes[3]),
 			nodesToRemove:  nil,
 			operations:     []string{"add", "add", "add", "add", "add", "next", "next", "next", "next", "add", "next", "next", "next"},
-			expectedOutput: []string{"node-4", "node-6", "node-7", "node-8", "node-3", "node-4", "node-6"},
+			expectedOutput: []string{"node-4", "node-5", "node-6", "node-7", "node-3", "node-8", "node-4"},
 		},
 		{
 			name:           "remove zone and add new to ensure exhausted is reset correctly",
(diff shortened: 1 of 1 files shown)


pkg/scheduler/internal/cache/node_tree_test.go
```diff
@@ -428,14 +428,14 @@ func TestNodeTreeMultiOperations(t *testing.T) {
 		},
 		{
 			name:           "add more nodes to an exhausted zone",
-			nodesToAdd:     append(allNodes[4:9], allNodes[3]),
+			nodesToAdd:     append(append(make([]*v1.Node, 0), allNodes[4:9]...), allNodes[3]),
```
liggitt, line 431: you can also triple-slice the data to get a sub slice of bounded capacity so the append always allocates a new slice

liggitt: e.g. `allNodes[4:9:9]`

draveness (commented):
/ok-to-test

pkg/scheduler/internal/cache/node_tree_test.go
soulxu: Do you mean 'append(allNodes[4:9:9], allNodes[3])'? it doesn't seems work.

I did a test like this:
```
func main() {
    var original = [4]int{1, 2, 3, 4}
    copied := append(original[0:2:4], original[2])
    fmt.Printf("copied: %v\n", copied)
    copied[0] = 666
    fmt.Printf("original %v, copied: %v\n", original, copied)
}
```

The output is:
```
copied: [1 2 3]
original [666 2 3 4], copied: [666 2 3]
```

The orignal array still changed.

soulxu: Thanks!, I got it, it is more tidy. Will update soon

chendave (commented):
/release-note-none
 
nice cleanup!

alculquicondor:
/approve
/lgtm

thanks

Huang-Wei:
Thanks @soulxu for the catch and @liggitt for the sweet tip.

liggitt:

/hold

Issue #56171, opened 2017-11-21
Kubelet flag precedence order vs files/ConfigMaps is dangerous


mtaufen:
See https://docs.google.com/document/d/18-MsChpTkrMGCSqAQN9QGgWuuFoK90SznBbwVkfZryo/ for details.

Doc reprinted here, with worse formatting:

[...]

PR #93518, opened 2020-07-29
[WIP] Convert Kubelet to use legacyflag

mtaufen:
**What type of PR is this?**
/kind cleanup

This converts Kubelet flag registrations that target Kubelet's ComponentConfig type (KubeletConfiguration) to use the legacyflag approach. Flag registrations become more verbose, but enforcing "flag precedence" (#56171) when parsing various sources of KubeletConfiguration becomes much more straightforward, because we no longer have to re-parse the command-line multiple times (with careful logic to avoid unwanted side-effects). Instead, legacyflag lets us parse the command-line once, store values for flags that target KubeletConfiguration in a scratch space, and apply values from explicitly set flags onto multiple KubeletConfigurations with no risk of distant side-effects.

legacyflag: https://github.com/kubernetes-sigs/legacyflag
legacyflag KEP: https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190129-legacyflag.md

```release-note
NONE
```

/assign @liggitt @sttts
/cc @stealthybox 

PTAL when you have a chance. I'm only working on this as I have time, but fortunately had some time today to get this started.

### DIFF
--- a/cmd/kubelet/app/options/BUILD
+++ b/cmd/kubelet/app/options/BUILD
@@ -41,6 +41,7 @@ go_library(
         "//staging/src/k8s.io/kubelet/config/v1beta1:go_default_library",
         "//vendor/github.com/spf13/pflag:go_default_library",
         "//vendor/k8s.io/klog/v2:go_default_library",
+        "//vendor/sigs.k8s.io/legacyflag/pkg/legacyflag:go_default_library",
     ] + select({
         "@io_bazel_rules_go//go/platform:android": [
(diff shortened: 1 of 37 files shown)


MonzElmasry:

mtaufen:
Whoops, thanks for the head's up! I untagged it from the milestone to avoid any confusion.

As this hasn't gotten traction I'm going to close it for now. It may still serve as a useful example if we come back to this in the future.

Issue #92928, opened 2020-07-09
FailedToUpdateEndpointSlices Error updating Endpoint Slices for Service


jijojv:
**What happened**:
Hi Folks, after every deployment we see this for about an hour. Seems to be harmless but wondering if this is a bug in v1.17.3
`kubectl  describe svc my-svc`

```
Events:
  Type     Reason                        Age   From                       Message
  ----     ------                        ----  ----                       -------
  Warning  FailedToUpdateEndpointSlices  35m   endpoint-slice-controller  Error updating Endpoint Slices for Service my-svc/my-app: Error updating my-app-h7q6v EndpointSlice for Service my-svc/my-app: Operation cannot be fulfilled on endpointslices.discovery.k8s.io "my-app-h7q6v": the object has been modified; please apply your changes to the latest version and try again

```
**What you expected to happen**:
`Events:            <none>`

**How to reproduce it (as minimally and precisely as possible)**:
 kubectl rollout restart deployment my_deploy

**Anything else we need to know?**:

**Environment**:
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"archive", BuildDate:"2020-03-20T16:41:14Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}

athenabot:
/triage unresolved

Comment `/remove-triage unresolved` when the issue is assessed and confirmed.

🤖 I am a bot run by vllry. 👩‍🔬

robscott:
Hey @jijojv, thanks for reporting this! This is not actually anything to worry about and I think the best solution will be for us to stop publishing that event if the error is related to an out of date cache like this. Due to the nature of the controller reacting to changes in Services and attempting to update related EndpointSlices, it can run into problems if the locally cached copy of EndpointSlices it has is out of date. It will naturally retry and resolve the issue when the cache updates. I'll work on a fix here to lower the logging and see if there are some ways to reduce the probability of this happening.

/remove-triage unresolved

robscott:
As a quick update, https://github.com/kubernetes/kubernetes/pull/93520 should fix this.

thockin:
@robscott is this a proper bug?  Or does it still need triage?

robscott:
[...]

PR #93443, opened 2020-07-25
Extracting EndpointSliceTracker into a new shared EndpointSlice package

robscott:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
The EndpointSlice and EndpointSliceMirroring controllers share a lot of logic, most notably the EndpointSliceTracker which was nearly identical. This extracts that out into a new util package. This package may also be useful in the future for extracting other code shared by the controllers.

**Special notes for your reviewer**:
This is a follow up to https://github.com/kubernetes/kubernetes/pull/93441. I'm not sure if this should or should not be part of the 1.19 release.

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

/sig network
/priority important-longterm
/assign @freehan

PR #93520, opened 2020-07-29
Delaying EndpointSlice controller syncs until cache is updated

### LINKED ISSUE #93374: [Flaky test] EndpointSlice should create Endpoints and EndpointSlices for Pods matching a Service

robscott:
**What type of PR is this?**
/kind bug
/kind failing-test

This updates the EndpointSlice controller to wait until the cache is updated (or a max delay of 3 seconds is exceeded) before running a sync for a Service. This should help avoid race conditions that would result in duplicate EndpointSlices or failed attempts to update stale EndpointSlices.

This will eventually fix #93374 and #92928.

This is very much still a WIP. I filed this early to get some initial feedback to at least see if I'm working in a reasonable direction. I do not love this approach but haven't thought of a better one yet, so very open to alternative ideas here.

I'd initially hoped that I could rely on the retry mechanism in the queue to simplify this implementation but that simply does not work since so many events can trigger a new sync. Currently I'm planning on updating the endpointSliceTracker to also track the last time the controller modified an EndpointSlice for each Service and then updating the logic to bypass the outdated cache check after a set number of seconds have passed since the last update. Other significant todos here include adding some test coverage and ensuring that a cache being outdated at one point in time does not forever doom a Service to slow sync times.

```release-note
EndpointSlice controller syncs now wait for cache updates to limit caching issues.
```

/sig network
/cc @liggitt @freehan

### DIFF
--- a/pkg/controller/endpointslice/endpointslice_controller.go
+++ b/pkg/controller/endpointslice/endpointslice_controller.go
@@ -55,6 +55,11 @@ const (
 	// 1s, 2s, 4s, 8s, 16s, 32s, 64s, 128s, 256s, 512s, 1000s (max)
 	maxRetries = 15
 
+	// maxCacheWaitDelay indicates the maximum amount of time the controller
+	// will wait for the EndpointSlice informer cache to be up to date for a
+	// Service before syncing it again.
+	maxCacheWaitDelay = 3 * time.Second
+
 	// endpointSliceChangeMinSyncDelay indicates the mininum delay before
 	// queuing a syncService call after an EndpointSlice changes. If
 	// endpointUpdatesBatchPeriod is greater than this value, it will be used
@@ -141,6 +146,7 @@ func NewController(podInformer coreinformers.PodInformer,
 	c.endpointSliceTracker = newEndpointSliceTracker()
 
 	c.maxEndpointsPerSlice = maxEndpointsPerSlice
+	c.maxCacheWaitDelay = maxCacheWaitDelay
 
 	c.reconciler = &reconciler{
 		client:               c.client,
@@ -216,6 +222,11 @@ type Controller struct {
 	// should be added to an EndpointSlice
 	maxEndpointsPerSlice int32
 
+	// maxCacheWaitDelay indicates the maximum amount of time the controller
+	// will wait for the EndpointSlice informer cache to be up to date for a
+	// Service before syncing it again.
+	maxCacheWaitDelay time.Duration
+
 	// workerLoopPeriod is the time between worker runs. The workers
 	// process the queue of service and pod changes
 	workerLoopPeriod time.Duration
@@ -268,7 +279,25 @@ func (c *Controller) processNextWorkItem() bool {
 	}
 	defer c.queue.Done(cKey)
 
-	err := c.syncService(cKey.(string))
+	namespace, name, err := cache.SplitMetaNamespaceKey(cKey.(string))
+	if err != nil {
+		c.handleErr(err, cKey)
+		return true
+	}
+
+	if c.endpointSliceTracker.ServiceCacheOutdated(namespace, name) {
+		if c.endpointSliceTracker.ServiceCacheUpdatedSince(namespace, name, time.Now().Add(-c.maxCacheWaitDelay)) {
+			klog.V(2).Infof("Informer cache for %s/%s EndpointSlices outdated, retrying", namespace, name)
+			c.queue.AddRateLimited(cKey)
+			return true
+		}
+		// Mark all EndpointSlices for a Service as updated if the wait delay is
+		// exceeded so this is not stuck in a delayed sync forever.
+		c.endpointSliceTracker.MarkServiceCacheUpdated(namespace, name)
+		klog.Warningf("Informer cache for %s/%s EndpointSlices outdated and maxCacheWaitDelay exceeded, attempting sync", namespace, name)
+	}
+
+	err = c.syncService(namespace, name)
 	c.handleErr(err, cKey)
 
 	return true
@@ -291,16 +320,11 @@ func (c *Controller) handleErr(err error, key interface{}) {
 	utilruntime.HandleError(err)
 }
 
-func (c *Controller) syncService(key string) error {
+func (c *Controller) syncService(namespace, name string) error {
 	startTime := time.Now()
 	defer func() {
-		klog.V(4).Infof("Finished syncing service %q endpoint slices. (%v)", key, time.Since(startTime))
+		klog.V(4).Infof("Finished syncing service %s/%s endpoint slices. (%v)", namespace, name, time.Since(startTime))
 	}()
-
-	namespace, name, err := cache.SplitMetaNamespaceKey(key)
-	if err != nil {
-		return err
-	}
 
 	service, err := c.serviceLister.Services(namespace).Get(name)
 	if err != nil {
@@ -320,7 +344,7 @@ func (c *Controller) syncService(key string) error {
 		return nil
 	}
 
-	klog.V(5).Infof("About to update endpoint slices for service %q", key)
+	klog.V(5).Infof("About to update endpoint slices for service %s/%s", namespace, name)
 
 	podLabelSelector := labels.Set(service.Spec.Selector).AsSelectorPreValidated()
 	pods, err := c.podLister.Pods(service.Namespace).List(podLabelSelector)
@@ -395,8 +419,11 @@ func (c *Controller) onEndpointSliceAdd(obj interface{}) {
 		utilruntime.HandleError(fmt.Errorf("Invalid EndpointSlice provided to onEndpointSliceAdd()"))
 		return
 	}
-	if managedByController(endpointSlice) && c.endpointSliceTracker.Stale(endpointSlice) {
-		c.queueServiceForEndpointSlice(endpointSlice)
+	if managedByController(endpointSlice) {
+		c.endpointSliceTracker.MarkCacheUpdated(endpointSlice)
+		if c.endpointSliceTracker.Stale(endpointSlice) {
+			c.queueServiceForEndpointSlice(endpointSlice)
+		}
 	}
 }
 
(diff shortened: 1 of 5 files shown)


robscott:
So I think this is a bit closer now, I've updated the logic to have a max delay instead of a max number of retries and I've added some admittedly less than ideal logic that will ensure that a single failure doesn't doom a Service to slow syncs forever. I've been testing this with kind and e2e clusters to try to understand the behavior of an informer and it's been frustratingly inconsistent. Sometimes adds or updates are taking significantly longer than my 3 second max delay to show up. At this point that is purely anecdotal and I don't have sufficient data/evidence to back that up. Will work on trying to figure out exactly what is going on here, but would certainly appreciate a second set of eyes on this in case I'm missing something obvious or even just if this direction doesn't actually make sense.

robscott:
I talked a bit with @wojtek-t this morning and he suggested trying a larger master. I've moved this up to a `n2-standard-4` and only scaling up a 100 pod deployment. Here's a copy of the kube-controller-manager logs from that (using the debug logging adding in this PR). The good news is that this approach generally seems to work. The bad news is that even with a larger master I'm still missing cache updates. 

There are some examples where this works exactly as expected:
 https://gist.github.com/robscott/df1832cd9a6d5a69adeccfa2830edf9a#file-kube-controller-manager-log-L427-L434

There are also other examples where the cache still takes >3s to be updated though:
~3.2s: https://gist.github.com/robscott/df1832cd9a6d5a69adeccfa2830edf9a#file-kube-controller-manager-log-L451-L489
~5s: https://gist.github.com/robscott/df1832cd9a6d5a69adeccfa2830edf9a#file-kube-controller-manager-log-L506-L526

Although this code is still quite messy, this is still the best solution I can think of. Unless @liggitt or @wojtek-t have better ideas I'll try to get this cleaned up and add some tests so this can actually be ready for review.

robscott:
Added some additional test coverage, going to move this out of draft mode so I can get full e2e tests on this. There are some EndpointSlice controller tests added by @mborsz around Pod batching that this breaks, so at least those will fail. Those are heavily dependent on timing and this is breaking that. I think there's a way to get that those to work with this, just haven't figured it out yet.

robscott:
I think this PR is ready for an initial round of review. I ended up updating the pod batch timing tests to essentially bypass this functionality by setting a `maxCacheDelay` of 0 seconds. Unfortunately given the nature of those tests being heavily dependent on timing and this introducing some inconsistency to the timing of when syncService() will be called, any non-zero value for `maxCacheDelay` could have introduced flakes here. 

As much as I would appreciate review on this PR now, I know that I still need to add test coverage for  `MarkServiceCacheUpdated` and `ServiceCacheUpdatedSince`. Additionally, if this new approach makes sense we should likely apply the same updates to the EndpointSliceMirroring controller. At that point my work towards [moving this out to a separate package](https://github.com/kubernetes/kubernetes/pull/93443) that can be referenced by both controllers becomes relevant. Adding a new package does seem like a bit of a stretch for a bug fix though.

With that said, my tests have shown that this does fix #93374 and the related bugs. I know this is not an ideal solution though, so if anyone has other ideas I'm very open to them.

/assign @liggitt @freehan @wojtek-t

robscott:
Can a milestone maintainer add this to the v1.19 milestone so it's easier to track? (The associated issue is already part of the v1.19 milestone).

andrewsykim:
/milestone v1.19

(as mentioned by Rob, adding to v1.19 since the issue is already in the v1.19 milestone)

robscott:
It sounds like @liggitt is really underwater at the moment, would really like to get some feedback on this PR and to get this fix into 1.19. This feels like an issue that should be fixed long term within the API Machinery - some easier way to wait for cache updates before running another sync loop. In the meantime this is the best possible fix I can come up with. I'm interested in feedback on this approach or ideas for other approaches that might work here. 

/cc @kubernetes/sig-api-machinery-api-reviews 
/assign @bowei @thockin

lavalamp:
Do you have a concise statement of the problem this is fixing, like not the effects, but what is wrong with the code?

lavalamp:
Is it this: "That would result in an annoying situation where an essentially duplicate EndpointSlice was created due to a stale cache"?

If so, the way to fix that is to make your creates deterministic, not to work on making the cache perfect (which is more or less impossible).

robscott:
Followed up with @lavalamp on chat and he had some great feedback. As a quick summary, there are 2 significant issues this is trying to fix:
1. Potential for duplicate EndpointSlices being created if cache is out of date
2. Controller attempts (and fails) to update stale copies of EndpointSlices

Here are the suggestions as I understood them (feel free to correct):
1. To fix 1 we can just compare the names of the EndpointSlices the controller expects with the names that the cache has. He'd suggested hashing here but as I've thought about it more I think we already have all the data we need in EndpointSliceTracker and we could achieve this with minimal changes to the original code.
2. Fixing 2 is more complicated. We need to be sure that we immediately return if a create, update, or delete fails (in that order). That is not currently the case but an update here would help limit the potential weird states we could get into.
3. We should avoid using `resourceVersion` to test for equality. Instead we should rely on a field we can increment on our own and compare safely. This would have to come in a later Kubernetes release.

1 and 2 are easy to implement and will result in a fix for #93374 and should help mitigate #92928. I'll work on a follow up PR tomorrow that includes those simpler fixes.

I'd still like to find a way to limit the number of unnecessary EndpointSlice syncs (before cache is updated) in 1.19. Every time we sync with an out of date cache we increase the probability that syncs will at least partially fail and leave us with a less than ideal state. I admit that this PR is a less than ideal solution, and @lavalamp has pointed out that it's nearly impossible to guarantee that our state is completely accurate, maybe there's still a way we can improve this somehow?

danwinship:
So I haven't spent a lot of time looking at EndpointSliceController, but it seems like maybe instead of doing

 - get an update
 - delay processing the update (`c.queue.AddAfter()`) to avoid too many endpoint reconciliations
 - eventually update caches and reconcile

it would be better to do

 - get an update
 - immediately update caches
 - delay reconciliation to avoid too many endpoint reconciliations

which is more or less how kube-proxy works (and in fact, the "delay reconciliation" step could use `async.BoundedFrequencyRunner` just like kube-proxy). Then you don't have to worry about the cache being out of date, because it's always up to date.

liggitt:
> it would be better to do
> 
> * get an update
> * immediately update caches
> * delay reconciliation to avoid too many endpoint reconciliations

that has a couple implications I'm not sure we want:
* a cache other than the informer-fed cache must be used
* it delays availability of initial endpoints for a pod

danwinship:
> * a cache other than the informer-fed cache must be used

By "immediately update caches", I was referring to EndpointSliceController's cache of what it expects the EndpointSlice state to be, which already exists. I was just saying that instead of having that cache sometimes be lagging behind the informer caches, we should instead have it sometimes be ahead of the actual state of EndpointSlices.

> * it delays availability of initial endpoints for a pod

EndpointSliceController is already doing that. It's just that currently it inserts the delay before processing the Pod event rather than after.

liggitt:
> instead of having that cache sometimes be lagging behind the informer caches, we should instead have it sometimes be ahead of the actual state of EndpointSlices.

the question is what to do when those don't converge, and how to distinguish between the following situations:
* they haven't converged because we're waiting for our informer to catch up
* they haven't converged because stuff happened that got collapsed by the informer (e.g. watch drops, our create happens, someone else's delete happens, then our informer relists)

robscott:
So on further thought I've realized that approach 1 is not quite as simple as I'd thought it would be:

> To fix 1 we can just compare the names of the EndpointSlices the controller expects with the names that the cache has. 

The problem with any approach like this is that we still need some kind of max delay per Service. This could also be triggered if someone deleted an EndpointSlice and a sync was called to recreate this. A check of only expected EndpointSlice names would delay that sync until a theoretical timeout/max delay was reached. Once we add that max delay it seems like we're awfully close to this PR just with a less effective comparison (name exists vs resource version has been received at some point).

lavalamp:
Fundamentally, controllers need to have some memory of what they did, and some expectation about what the effect on the environment will be. Otherwise, in addition to double actuation, you risk oscillation.

A cheap hack is to at least make actions deterministic enough that they conflict, that way you can't double actuate. Hence instead of using generate name, choose a new name that depends on the names you currently observe. That probably helps for most cases, but perhaps not if you need to e.g. make multiple objects at a time.

I really think you need to keep two caches, a cache of the observed state (which you already have) and a cache of pending actions. You may need to keep a logical clock in objects so you can tell which cache is more recent.

There's some chance a "write through cache" would do the thing you want, but we don't have one ready to go, and I'd have to think about it.

robscott:
Talked with @bowei this morning about this some more and he pointed out that this specific PR is more of an optimization than a correction. With that said, I think I'm OK with leaving this specific PR/optimization problem for later and making the following smaller fixes:

* Update the flaky test to consider duplicate EndpointSlices valid.
* Update the docs to make it very clear that consumers of the EndpointSlice API will have to do deduplication (as kube-proxy already does).
* Update the controller error handling to return immediately if a create, update, or delete errors (in that order).

I'm hoping to have PRs for all of the above ready later today.

liggitt:
explicitly holding until we decide if we want to go with this approach

thockin:
Rob - keep or close?

robscott:
Thanks for the reminder on this one @thockin! This specific approach is likely not ideal, but it does seem like we'll need some optimizations here. At present, we're doing a lot of unnecessary work due to the controller operating with an out of date cache. Although certainly not elegant, I believe even this PR would represent an improvement over the status quo. I think @lavalamp's suggestion above to try to use/implement something resembling a write-through cache could be helpful. 

Fundamentally I don't think we want to move away from the informer cache as the source of truth here, so we just need to find a way reduce the odds that we'll run a sync with before that cache has caught up. Those odds are currently rather high during a rolling update, and I don't think it would take a very complex solution to provide a significant improvement here.

/remove-lifecycle stale

aojea:
> A cheap hack is to at least make actions deterministic enough that they conflict

@robscott if the endpoint slice name is deterministic, will this be solved?

```
kubectl get endpointslices -o wide
NAME         ADDRESSTYPE   PORTS     ENDPOINTS    AGE
test-001   IPv4          <unset>   <unset>      19d
test-002   IPv4          <unset>   <unset>      19d
```

robscott:
Unfortunately I don't think the name itself is sufficient. If I'm remembering the problem correctly, we're also seeing a large number of these warnings from the controller attempting to update a stale copy of an EndpointSlice.

robscott:
Created a new PR that takes a bit of a simpler approach here: #99345. Closing this one in favor of that.

/close

Issue #12964, opened 2015-08-20
Endpoints are not persistented


wb14123:
I create a glusterfs endpoint, but after I restart kube-apiserver, the endpoint is missing. Does Kubernetes persistence endpoints?

thockin:
Do you mean and Endpoints object in the API or something else?

wb14123:
Here is what I do:

```
 > kubectl create -f glusterfs-endpoints.yaml
endpoints/glusterfs-cluster

> cat glusterfs-endpoints.yaml
kind: Endpoints
apiVersion: v1
metadata:
  name: glusterfs-cluster
subsets:
  - addresses:
      - IP: 192.168.1.1
    ports:
      - port: 1
  - addresses:
      - IP: 192.168.1.2
    ports:
      - port: 1
```

Then use `kubectl get endpoints` I can see this line:

```
glusterfs-cluster          192.168.1.1:1,192.168.1.2:1
```

But after I restart kube-apiserver, this endpoint is lost.

thockin:
Do you have a Service named `glusterfs-cluster` ?

Otherwise it should persist, just the same as any other API object

wb14123:
No. This endpoints is just used for persistent volumes, which is used by other pods. I followed [this doc](http://kubernetes.io/v1.0/examples/glusterfs/README.html). Is it necessary to define a service?

thockin:
@wojtek-t Can one of you guys try to repro?

markturansky:
> But after I restart kube-apiserver, this endpoint is lost.

What does that mean exactly?  What kind of cluster are you running?  Does *anything* persist and endure between restarts of the API server?  When you restart the API server (why?), are you also restarting etcd or other services?

wb14123:
The cluster just have one master. The pods, secrets, controllers and services are recovered after the restart, but the endpoints are not.

About why I restart the API server, well, it may fail (or something else). You cannot expect the api-server runs and never restart. After I restart it, I expect all the things are recovered. In my case, I just restart api-server through /etc/init.d/kube-apiserver (kill the process manually should do the same thing). Etcd is always running and the data in it is not lost.

wb14123:
Any update on this? I cannot use kubernetes in production if this problem is not solved. Or do you have a work around, like define a service with the glusterfs endpoints?

jonathan-kosgei:
[...]

PR #93535, opened 2020-07-29
Create service for gluster endpoint

gnufied:
Creating an endpoint without a service can cause endpoint to disappear. See - https://github.com/kubernetes/kubernetes/issues/12964

This fixes an e2e flake that can happen, as endpoints may disappear during tests.

/kind cleanup
/sig storage
/priority important-soon

### DIFF
--- a/test/e2e/framework/volume/fixtures.go
+++ b/test/e2e/framework/volume/fixtures.go
@@ -173,6 +173,23 @@ func NewGlusterfsServer(cs clientset.Interface, namespace string) (config TestCo
 	}
 	pod, ip = CreateStorageServer(cs, config)
 
+	service := &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


test/e2e/framework/volume/fixtures.go
```diff
@@ -202,6 +202,23 @@ func NewGlusterfsServer(cs clientset.Interface, namespace string) (config TestCo
 	_, err := cs.CoreV1().Endpoints(namespace).Create(context.TODO(), endpoints, metav1.CreateOptions{})
 	framework.ExpectNoError(err, "failed to create endpoints for Gluster server")
 
+	service := &v1.Service{
```
jsafrane, line 205: Should the test create Service first and then the Endpoint? Because the endpoint is without a service for a short time and could be garbage collected,

gnufied: done.

Issue #93532, opened 2020-07-29
typecheck uses exorbitant amounts of memory


hasheddan:
**Background**:

We recently introduced Guaranteed Pod QOS on jobs running in `k8s-infra-prow-build` GCP project (https://github.com/kubernetes/test-infra/pull/18471). Introducing limits on jobs naturally started causing some failures, one of which was `verify-master` ([testgrid](https://testgrid.k8s.io/sig-release-master-blocking#verify-master)). The original memory limit of `46Gi` caused the pod to be unscheduleable, so it was bumped down to `42Gi` (https://github.com/kubernetes/test-infra/pull/18502) at which point the pod became scheduleable and the job ran successfully ([job run](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-verify-master/1288169770910224385)). At that point, after observing metrics, I mistakenly concluded that this was much higher than needed and suggested bumping down again (https://github.com/kubernetes/test-infra/pull/18502#issuecomment-665191173). This caused the job to continue to run, but to fail on `verify.typecheck` and sometimes `verify.typecheck-dockerless`.

`test/typecheck` was recently refactored to not depend on GOPATH (https://github.com/kubernetes/kubernetes/pull/92212). This was a fairly significant change so I looked at the metrics before and after the rewrite was merged.

---

`verify-* pre-merge` (June 25th)
![pre-merge](https://user-images.githubusercontent.com/31777345/88804188-2c209b00-d173-11ea-9055-9fa33bf9cb3c.png)

---

`verify-* post-merge` (June 27th)
![post-merge](https://user-images.githubusercontent.com/31777345/88804361-6b4eec00-d173-11ea-9e4d-37170c511139.png)

Memory usage nearly double for `verify-master` post-merge, which appears to be directly related to the rewrite given that `verfiy-*` on other branches remained relatively constant. This can be affirmed further by the fact that `verify-master` and `verify-beta` are failing with the new resource limits, while older release branches are not.

I can see a few options for how we may be able to remedy this issue in the short term:
- Revert the rewrite to get back to lower (yet still very high) memory usage
- Separate `typecheck` into a separate job to not reduce signal on the rest of `verify-master`
- Bump memory limits back up to `~42Gi` to get this running successfully again for now

I plan to look into fixing up `typecheck` but think it would be a good idea to take one of the directions mentioned above if this cannot be remedied quickly.

/priority critical-urgent
/sig testing
[...]

PR #93544, opened 2020-07-29
Run verify-typecheck in serial to reduce memory usage

hasheddan:
**What type of PR is this?**

/kind failing-test

verify-typecheck is currently hitting memory limits in periodic CI jobs.
This changes it to run checks for each platform serially to reduce
memory usage at the expense of a slightly longer run time.

Ref: https://github.com/kubernetes/kubernetes/issues/93532

This was mostly based off local runs. It looks like when we have successful run this takes 2-3 minutes (ref: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-verify-master/1288169770910224385). Running serially should have a linear effect on duration while drastically reducing the required memory allocation for the `verify-*` jobs. It would be good to circle back here and see if running platform checks in parallel can be improved. Note also that `typecheck-dockerless` appears to experience a similar problem, but I wanted to test this out here before also running serially there.

/cc @BenTheElder @spiffxp @liggitt 

```release-note
NONE
```

```docs

```

### DIFF
--- a/hack/verify-typecheck.sh
+++ b/hack/verify-typecheck.sh
@@ -35,7 +35,7 @@ make --no-print-directory -C "${KUBE_ROOT}" generated_files
 # force this tooling to run in a fake GOPATH.
 ret=0
 hack/run-in-gopath.sh \
-    go run test/typecheck/main.go "$@" || ret=$?
+    go run test/typecheck/main.go --serial "$@" || ret=$?
 if [[ $ret -ne 0 ]]; then
   echo "!!! Type Check has failed. This may cause cross platform build failures." >&2
   echo "!!! Please see https://git.k8s.io/kubernetes/test/typecheck for more information." >&2


BenTheElder:
instead of serial, can we tune how concurrently?

hasheddan:
@BenTheElder do you mean run some platforms concurrently, just not all at once?

BenTheElder:
Right, along the lines of ginkgo num nodes

hasheddan:
@BenTheElder sounds good, update coming :+1:

liggitt:
oops, didn't see the parallel request got acked, opened https://github.com/kubernetes/kubernetes/pull/93572

hasheddan:
@liggitt No worries :+1: Do you think we should try this out at the default `4` at first?

PR #93546, opened 2020-07-29
Mitigate admission race condition for notfound ingressclass

### LINKED ISSUE #92198: Update ingressclass admission controller to handle IngressClass no found
**What happened**:

The [existing ingressclass admission controller](https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/defaultingressclass/admission.go) is susceptible to rejecting pods containing both an ingressclass and an ingress. 

NotFound lookups should fall back to a live lookup in admission

J Liggitt pointed out a similar fix in [RuntimeClass](https://github.com/kubernetes/kubernetes/pull/91315/files#diff-c764b2d17e2092fc99633557915a282d) that we can use.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.18
- OS (e.g: `cat /etc/os-release`): N/A
- Install tools: N/A
- Others:

cmluciano:
**What type of PR is this?**
/kind bug

This PR should mitigate problems with a NotFound lookup on an IngressClass from the informer cache within the admission controller. If this happens, the additional logic should now attempt a live lookup and manifests with both ingressclass/ingress should succeed.

Fixes #92198

```release-note
NONE
```


plugin/pkg/admission/defaultingressclass/admission.go
```diff
@@ -63,6 +69,11 @@ func newPlugin() *classDefaulterPlugin {
 	}
 }
 
+// SetExternalKubeClientSet sets the client for the plugin
+func (a *classDefaulterPlugin) SetExternalKubeClientSet(client kubernetes.Interface) {
+	a.client = client.NetworkingV1().IngressClasses()
```
liggitt, line 74: ensure this is non-nil in `classDefaulterPlugin#ValidateInitialization()`

plugin/pkg/admission/defaultingressclass/admission.go
```diff
@@ -125,8 +136,17 @@ func (a *classDefaulterPlugin) Admit(ctx context.Context, attr admission.Attribu
 }
 
 // getDefaultClass returns the default IngressClass from the store, or nil.
-func getDefaultClass(lister networkingv1listers.IngressClassLister) (*networkingv1.IngressClass, error) {
+func (a *classDefaulterPlugin) getDefaultClass(ctx context.Context, lister networkingv1listers.IngressClassLister, ingress *networking.Ingress) (*networkingv1.IngressClass, error) {
 	list, err := lister.List(labels.Everything())
+	if apierrors.IsNotFound(err) {
+		// if not found, informer cache could be stale, try a live lookup
+		ing, err := a.client.Get(ctx, *ingress.Spec.IngressClassName, metav1.GetOptions{})
```
liggitt, line 143: `getDefaultClass` is only called when `ingress.Spec.IngressClassName` is nil... this will panic

plugin/pkg/admission/defaultingressclass/admission.go
```diff
@@ -125,8 +136,17 @@ func (a *classDefaulterPlugin) Admit(ctx context.Context, attr admission.Attribu
 }
 
 // getDefaultClass returns the default IngressClass from the store, or nil.
-func getDefaultClass(lister networkingv1listers.IngressClassLister) (*networkingv1.IngressClass, error) {
+func (a *classDefaulterPlugin) getDefaultClass(ctx context.Context, lister networkingv1listers.IngressClassLister, ingress *networking.Ingress) (*networkingv1.IngressClass, error) {
 	list, err := lister.List(labels.Everything())
+	if apierrors.IsNotFound(err) {
```
liggitt, line 141: List won't return a NotFound error, it will return an empty list

PR #93548, opened 2020-07-29
kubectl: Allow --patch-file to be passed to `kubectl patch`

smarterclayton:
The current structure is limited in the size of the patch it can
accept depending on the shell of the environment (for instance, a
900kb patch cannot be applied in bash 4 on Fedora 32). Allow a user
to specify `--patch-file` to provide a file containing the patch
rather than directly on the command line.

/kind bug

```release-note
Very large patches may now be specified to `kubectl patch` with the `--patch-file` flag instead of including them directly on the command line. The `--patch` and `--patch-file` flags are mutually exclusive.
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/kubectl/pkg/cmd/patch/patch.go
+++ b/staging/src/k8s.io/kubectl/pkg/cmd/patch/patch.go
@@ -18,6 +18,7 @@ package patch
 
 import (
 	"fmt"
+	"io/ioutil"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


brianpursley:
/lgtm

It might be a little confusing to someone at first because there are both `-f` and `--patch-file` flags, but they serve different purposes. I think it is fine.

brianpursley (approved):

smarterclayton:
Yeah, unfortunately we can't really alter `-f` and it's in broad use across so many commands, whereas in this specific case it would be more natural to patch from a file than read options from a file.

liggitt:

/retest

PR #93549, opened 2020-07-30
Fix ResourceQuota comments

### LINKED ISSUE #93312: Comments on ResourceQuota are incomplete
The following comments are incomplete.

https://github.com/kubernetes/kubernetes/blob/98814409bab818d942e3c341d76eb594f12675a5/staging/src/k8s.io/api/core/v1/types.go#L5493-L5504

For example, comments for "ResourceQuotaScopeTerminating" should be:

```
// Match all pod objects where `spec.activeDeadlineSeconds >=0`.
```

Ref: https://kubernetes.io/docs/concepts/policy/resource-quotas/#quota-scopes

/good-first-issue

Dean-Coakley:
#### What type of PR is this?

/kind documentation

#### What this PR does / why we need it:

* Makes comments in line with the following docs: https://kubernetes.io/docs/concepts/policy/resource-quotas/#quota-scopes

#### Which issue(s) this PR fixes:a

Fixes: https://github.com/kubernetes/kubernetes/issues/93312

#### Special notes for your reviewer:
* Note that `pkg/apis/core/types.go` changes are not described in the issue but they seem to make sense.

cc: @Huang-Wei do these make sense?^

#### Does this PR introduce a user-facing change?:

```release-note
NONE
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:
```
```

### DIFF
--- a/pkg/apis/core/types.go
+++ b/pkg/apis/core/types.go
@@ -4771,9 +4771,9 @@ type ResourceQuotaScope string
 
 // These are valid values for resource quota spec
 const (
-	// Match all pod objects where spec.activeDeadlineSeconds
+	// Match all pod objects where spec.activeDeadlineSeconds >=0
 	ResourceQuotaScopeTerminating ResourceQuotaScope = "Terminating"
-	// Match all pod objects where !spec.activeDeadlineSeconds
+	// Match all pod objects where spec.activeDeadlineSeconds is nil
 	ResourceQuotaScopeNotTerminating ResourceQuotaScope = "NotTerminating"
 	// Match all pod objects that have best effort quality of service
 	ResourceQuotaScopeBestEffort ResourceQuotaScope = "BestEffort"
--- a/staging/src/k8s.io/api/core/v1/types.go
+++ b/staging/src/k8s.io/api/core/v1/types.go
@@ -5491,9 +5491,9 @@ const (
 type ResourceQuotaScope string
 
 const (
-	// Match all pod objects where spec.activeDeadlineSeconds
+	// Match all pod objects where spec.activeDeadlineSeconds >=0
 	ResourceQuotaScopeTerminating ResourceQuotaScope = "Terminating"
-	// Match all pod objects where !spec.activeDeadlineSeconds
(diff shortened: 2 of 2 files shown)


Huang-Wei:
@Dean-Coakley LGTM, we should update the comments in both places.

pkg/apis/core/types.go
```diff
@@ -4771,9 +4771,9 @@ type ResourceQuotaScope string
 
 // These are valid values for resource quota spec
 const (
-	// Match all pod objects where spec.activeDeadlineSeconds
+	// Match all pod objects where `spec.activeDeadlineSeconds >=0`
```
Huang-Wei, line 4774: I think we should only quote `spec.activeDeadlineSeconds` (applies elsewhere):

```suggestion
	// Match all pod objects where `spec.activeDeadlineSeconds` >=0
```

Huang-Wei:
LGTM. Please squash.

/retest

pkg/apis/core/types.go
```diff
@@ -4771,9 +4771,9 @@ type ResourceQuotaScope string
 
 // These are valid values for resource quota spec
 const (
-	// Match all pod objects where spec.activeDeadlineSeconds
+	// Match all pod objects where `spec.activeDeadlineSeconds` >=0
```
yliaog, line 4774: why adding backtick?

Dean-Coakley: Hmm I saw it in the example solution in the issue.

While grepping around I couldn't really see any uses like this though. - Happy to remove it.

Dean-Coakley: Removed them and squashed down to one commit.

Dean-Coakley:
I think this needs `/approve` and to be tagged for 1.19 to be merged?

Huang-Wei:
> I think this needs `/approve` and to be tagged for 1.19 to be merged?

I think we can leave it for 1.20 as it's a glitch on commenting.

Huang-Wei:
/milestone v1.21

@kubernetes/sig-api-machinery-misc for /approve.

Issue #93509, opened 2020-07-28
API-Server should validate ipFamily and configured service-cluster-ip-ranges


johscheuer:
**What happened**:
Based on this issue: https://github.com/kubernetes/kubernetes/issues/93446 I would expect that the API server returns an error if I try to create a Service for an `ipFamily` that has no according `service-cluster-ip-range` instead of silently swallow the mistake and use the default `ipFamily`.

**What you expected to happen**:
That the API server provides an error "Service x uses undefined ipFamily: IPv6".

**How to reproduce it (as minimally and precisely as possible)**:
Create a Kubernetes cluster with [Dualstack](https://kubernetes.io/docs/concepts/services-networking/dual-stack) enabled and only provide an IPv4  `service-cluster-ip-range` e.g. `--service-cluster-ip-range=10.96.0.0/12` now create a Service with `ipFamily: IPv6`:

```bash
cat << EOF | kubectl apply -f -
apiVersion: v1
kind: Service
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  ipFamily: IPv6
  type: LoadBalancer
  selector:
    app: nginx
  ports:
  - protocol: TCP
    port: 80
    targetPort: 80
EOF
```

and validate the service IP is IPv4: `kubectl get svc nginx`

**Anything else we need to know?**:
I think this behaviour is more user friendly and makes it more obvious what pieces might missing.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.18.5
- Cloud provider or hardware configuration: none
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

athenabot:
/triage unresolved

Comment `/remove-triage unresolved` when the issue is assessed and confirmed.

🤖 I am a bot run by vllry. 👩‍🔬

liggitt:
> Create a Kubernetes cluster with [Dualstack](https://kubernetes.io/docs/concepts/services-networking/dual-stack) enabled and only provide an IPv4 `service-cluster-ip-range`

Is that a valid configuration? I thought both ranges must be enabled to actually enable the feature.

johscheuer:
[...]

PR #91824, opened 2020-06-05
dual stack services

khenidak:
implements https://github.com/kubernetes/enhancements/pull/1679 in its eventual form.

status of work
- [x] api structure change
- api work
  - [x]  helpers, utils etc..
  - [x] defaulting 
  - [x] defaulting for existing services (headful + headless + headless without selectors)
  - [x] validation 
  - [x] conversion
- tie dualstack services to EndPointSlice feature - startup flag validation
   - [x] api server
   - [x] controller manager
- EndpointSlice
   - [x] ensure that EndpointSlice handles multi family services
- service ip alloc
   - [x] ~~default kuberentes service~~
   - [x] ~~force kuberenetes default service into dual stack~~
   - [x] user services
   - [x] strategy: clear IPFamilies, IPFamilyPolicy, ClusterIPs when converting ExternalName type.
   - [x] api-server repair loop
- pod/pod net
  - [x] ~~env vars handle dual stack services~~
  - [x] remove kubenet v4,v6 enforcement
- kube proxy
  - [x] flag validation, tie dual stack to EndPointSlice
  - [x] reserve NodePort for both ipv6/ipv4 by default
  - [x] ~~modify metaproxier to work only with EndPointSlice~~
  - [x] ~~modify metaproxier to expect EndPointSlice with both families endpoints~~
  - [x] force ipvs proxier to lookup ClusterIP by family (and correctly select ranges and related fields accordingly)
  - [x] force iptables to lookup ClusterIP by family (and correctly select ranges and related fields accordingly)
  - [x] `FIX` ipvs mishandling of families.
  - [x] validate that topology keys does not conflict with dualstack 
- kubectl
  - [x] modify `kubectl describe` to work with multi ClusterIPs
- e2e
  - [x] fix existing e2e tests that will break due to api change
  - [x] add new e2e tests for upgrade and downgrade scenarios

```release-note
Add dual-stack Services (alpha).  This is a BREAKING CHANGE to an alpha API.
It changes the dual-stack API wrt Service from a single ipFamily field to 3
fields: ipFamilyPolicy (SingleStack, PreferDualStack, RequireDualStack),
ipFamilies (a list of families assigned), and clusterIPs (inclusive of
clusterIP).  Most users do not need to set anything at all, defaulting will
handle it for them.  Services are single-stack unless the user asks for
dual-stack.  This is all gated by the "IPv6DualStack" feature gate.
```

PR #93554, opened 2020-07-30
Add additional test cases for dualstack ip family validation

johscheuer:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

 /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

I added two additional test case to the dualstack `ipFamily` validation to make it more clear that we also catch the case that users provide the wrong `ipFamily`. I stumbled over this here: https://github.com/kubernetes/kubernetes/issues/93509. 

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/apis/core/validation/conditional_validation_test.go
+++ b/pkg/apis/core/validation/conditional_validation_test.go
@@ -461,6 +461,28 @@ func TestValidateServiceIPFamily(t *testing.T) {
 			},
 			expectErr: nil,
 		},
+		{
+			name:             "provided IP Family not in allowed list (IPv6)",
+			dualStackEnabled: true,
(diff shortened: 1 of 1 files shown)


johscheuer:
JFYI @aojea @danwinship

danwinship:
Given that that code is going to be heavily rewritten with the dual-stack changes in 1.20, I'm not sure it makes sense to add these new tests now. You could review #91824 and make sure the new versions of the tests there end cover everything though?

(Because _really_ your problem isn't that checks or tests were missing in 1.19, it's that nothing worked right at all in 1.18.)

johscheuer:
That makes sense and is totally fine for me. I just thought these tests cases make it a little bitt more clear that these cases are also covered.

PR #93551, opened 2020-07-30
Remove special case for node e2e tests when expecting pod rejection

brianpursley:
**What type of PR is this?**
/kind failing-test

**What this PR does / why we need it**:
See periodic e2e job: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2enode-ubuntu1-k8sbeta-gkespec/1288587113490550786

The following two tests fail with error `pods \"test-runtimeclass-runtimeclass-9403-nonexistent-\" is forbidden: pod rejected: RuntimeClass \"runtimeclass-9403-nonexistent\" not found`:

* [sig-node] RuntimeClass should reject a Pod requesting a non-existent RuntimeClass
* [sig-node] RuntimeClass should reject a Pod requesting a deleted RuntimeClass

The problem is that this comment is no longer true:
```
	// The Node E2E doesn't run the RuntimeClass admission controller, so we expect the rejection to
	// happen by the Kubelet.
```

In a previous commit there was a [change made to include RuntimeClass in the enabled admission controllers for testing](https://github.com/kubernetes/kubernetes/commit/a4f8ee17ee5d5beb1efb2a28fa8466ed3ca70ad3).

Also, RuntimeClass is on by default (it looks like since 1.17)
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubeapiserver/options/plugins.go#L160
https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go#L713

This PR removes the special case handling for `framework.TestContext.NodeE2E` which is no longer needed.

**Which issue(s) this PR fixes**:
No issue created

**Special notes for your reviewer**:
/cc @liggitt 

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:
```docs

```

PR #93555, opened 2020-07-30
integration: Wait for PV to be available in TestVolumeProvision

### LINKED ISSUE #81368: TestVolumeProvision is flaky

cofyc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
/kind flake

Fixes #81368 

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/integration/volumescheduling/volume_binding_test.go
+++ b/test/integration/volumescheduling/volume_binding_test.go
@@ -762,7 +762,7 @@ func testVolumeProvision(t *testing.T, storageCapacity bool) {
 	}
 
 	run := func(t *testing.T, test testcaseType) {
-		// Create StorageClasses
+		t.Log("Creating StorageClass")
 		suffix := rand.String(4)
 		classes := map[string]*storagev1.StorageClass{}
 		classes[classImmediate] = makeDynamicProvisionerStorageClass(fmt.Sprintf("immediate-%v", suffix), &modeImmediate, nil)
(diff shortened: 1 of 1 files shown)


cofyc:
(run more times to check if it's stable)

cofyc:
(run more times to check if it's stable)

test/integration/volumescheduling/volume_binding_test.go
```diff
@@ -783,35 +783,43 @@ func testVolumeProvision(t *testing.T, storageCapacity bool) {
...
 		}
-		// Create PVs
+
+		t.Log("Creating PVs")
 		for _, pvConfig := range test.pvs {
 			pv := makePV(pvConfig.name, classes[pvConfig.scName].Name, pvConfig.preboundPVC, config.ns, pvConfig.node)
 			if _, err := config.client.CoreV1().PersistentVolumes().Create(context.TODO(), pv, metav1.CreateOptions{}); err != nil {
 				t.Fatalf("Failed to create PersistentVolume %q: %v", pv.Name, err)
 			}
+			if err := waitForPVPhase(config.client, pvConfig.name, v1.VolumeAvailable); err != nil {
```
msau42, line 793: Does this mean the race exists in a real cluster too?

cofyc: It's more like the uncertainty in testing. Our volume binder will try to dynamically provision if no static matching PV is found. But in some scenarios, we expect the volume binder to use the static PV, so we need to wait for the PV to be available. 

I think it's fine in real clusters. In theory, it's possible that some newly created PVs may not be synced to the scheduler or not be in the available state when a PVC is created to request a PV from the storage class. But, normally both the static provisioning and dynamic provisioning are acceptable for users if a storage class supports both of them.

cofyc:
@ahg-g can you add this fix into 1.19 milestone?

cofyc:
/retest

found another issue (https://github.com/kubernetes/kubernetes/issues/93830), which I'm going to fix in https://github.com/kubernetes/kubernetes/pull/93831

liggitt:
/hold

Will fixing https://github.com/kubernetes/kubernetes/issues/93830 remove the flake in this test? To @msau42's question, does this failure represent a real bug? If so, we probably want to leave the test that exposed the issue as-is

cofyc:
> /hold
> 
> Will fixing #93830 remove the flake in this test? To @msau42's question, does this failure represent a real bug? If so, we probably want to leave the test that exposed the issue as-is

yes, #93830 is a bug that was introduced in #93511. It's a real bug which we should [fix in the scheduler code](https://github.com/kubernetes/kubernetes/pull/93831).

This PR fixes an issue in the test that [causes failures](https://github.com/kubernetes/kubernetes/issues/81368#issuecomment-666043508) with or without #93551. 
It's not a real bug. When a storage class supports both static and dynamic provisioning, a PVC can be bound to a static PV or dynamic provisioned. However, in some test cases, we want the PVC to be statically bound, so we must wait for pre-provisioned static PVs to be available first. Fixing this issue will not hide the #93830 issue.

liggitt:
ack, thanks

/hold cancel

PR #93557, opened 2020-07-30
fix(e2e::network): refactor NoSNAT test 

### LINKED ISSUE #93525: Failing Test [sig-network] NoSNAT [Feature:NoSNAT] [Slow] Should be able to send traffic between Pods without SNAT

The test currently only works on Cloud Providers or environments that set the External IP on nodes, because it is using the framework directly to check the IPs of the pods, using a no-snat-test-proxy to check if the Pods are not using SNAT.

However, this can be simplified, removing the no-snat-test-proxy Pod and executing the queries directly in a Pod.

**Why is this needed**:

The test should work in all environments, and it's failing

https://k8s-testgrid.appspot.com/sig-network-kind#sig-network-kind,%20master&include-filter-by-regex=NoSNAT

knight42:

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation

/kind failing-test
> /kind feature
> /kind flake

> @aojea: removing the no-snat-test-proxy Pod and executing the queries directly in a Pod

Fixes #93525 

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/network/no_snat.go
+++ b/test/e2e/network/no_snat.go
@@ -19,115 +19,42 @@ package network
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
-	"net/http"
-	"strconv"
-	"strings"
+	"net"
 	"time"
 
+	"github.com/onsi/ginkgo"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+
 	"k8s.io/kubernetes/test/e2e/framework"
-
-	"github.com/onsi/ginkgo"
-
 	imageutils "k8s.io/kubernetes/test/utils/image"
 )
 
 const (
-	testPodPort = 8080
-
-	testProxyPort = 31235 // Firewall rule allows external traffic on ports 30000-32767. I just picked a random one.
+	testPodPort    = "8080"
+	noSNATTestName = "no-snat-test"
 )
 
 var (
 	testPod = v1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
-			GenerateName: "no-snat-test",
+			GenerateName: noSNATTestName,
 			Labels: map[string]string{
-				"no-snat-test": "",
+				noSNATTestName: "",
 			},
 		},
 		Spec: v1.PodSpec{
 			Containers: []v1.Container{
 				{
-					Name:  "no-snat-test",
+					Name:  noSNATTestName,
 					Image: imageutils.GetE2EImage(imageutils.Agnhost),
-					Args:  []string{"no-snat-test", "--port", strconv.Itoa(testPodPort)},
-					Env: []v1.EnvVar{
-						{
-							Name:      "POD_IP",
-							ValueFrom: &v1.EnvVarSource{FieldRef: &v1.ObjectFieldSelector{FieldPath: "status.podIP"}},
-						},
-					},
-				},
-			},
-		},
-	}
-
-	testProxyPod = v1.Pod{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "no-snat-test-proxy",
-		},
-		Spec: v1.PodSpec{
-			HostNetwork: true,
-			Containers: []v1.Container{
-				{
-					Name:  "no-snat-test-proxy",
-					Image: imageutils.GetE2EImage(imageutils.Agnhost),
-					Args:  []string{"no-snat-test-proxy", "--port", strconv.Itoa(testProxyPort)},
-					Ports: []v1.ContainerPort{
-						{
-							ContainerPort: testProxyPort,
-							HostPort:      testProxyPort,
-						},
-					},
+					Args:  []string{"netexec", "--http-port", testPodPort},
 				},
 			},
 		},
 	}
 )
-
-// Produces a pod spec that passes nip as NODE_IP env var using downward API
-func newTestPod(nodename string, nip string) *v1.Pod {
-	pod := testPod
-	nodeIP := v1.EnvVar{
-		Name:  "NODE_IP",
-		Value: nip,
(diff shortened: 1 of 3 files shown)


aojea:
hi @knight42 
the test is failing to me, also adding `e2enetwork.NewNetworkingTestConfig` to the test adds more overhead and complexity (it creates services and additional pods that are not being used)

I was wondering if we can simplify the test, also get rid of the agnhost no-snat-test-* commands.

Let me know what do you think if we rewrite the test to be something like this

https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/service.go#L944-L1026

replacing the service by a pod

knight42:
/cc @aojea

I have simplified the test a bit, and run the test locally and it passed.

test/e2e/network/no_snat.go
```diff
@@ -194,64 +128,22 @@ var _ = SIGDescribe("NoSNAT [Feature:NoSNAT] [Slow]", func() {
...
-			for _, ip := range podIPs {
-				if ip == pip {
+		for _, sourcePod := range pods.Items {
+			for _, targetPod := range pods.Items {
+				if targetPod.Name == sourcePod.Name {
 					continue
 				}
-				ips = append(ips, ip)
+				targetAddr := net.JoinHostPort(targetPod.Status.PodIP, testPodPort)
+				sourceIP, execPodIP := execSourceipTest(sourcePod, targetAddr)
```
aojea, line 143: awesome

aojea:
This is great @knight42 
/lgtm
/assign @hasheddan @mrhohn

Daniel this fixes a failing test, asking for adding it to mileston 1.19 :smile:

hasheddan (approved):
/lgtm
/milestone v1.19

test/e2e/network/no_snat.go
```diff
@@ -154,29 +108,9 @@ var _ = SIGDescribe("NoSNAT [Feature:NoSNAT] [Slow]", func() {
 			framework.ExpectNoError(err)
 		}
 
-		// In some (most?) scenarios, the test harness doesn't run in the same network as the Pods,
```
andrewsykim, line 157: Do we still need to feed the node IP in `newTestPod` or can that be removed now?

aojea: I think so, so we test the connectivity from all the nodes in the cluster that is being tested, and make the test predictable

aojea: we can get rid of all the environment variables though

andrewsykim: > we can get rid of all the environment variables though

yeah was specifically referring to the NODE_IP env var we're adding to all the tests above. Seems enough to just create a test pod per node without the env var now.

aojea: duh, scratch my comment, you are absolutely right, nodeName is what we need :upside_down_face:

test/e2e/network/no_snat.go
```diff
@@ -194,64 +97,22 @@ var _ = SIGDescribe("NoSNAT [Feature:NoSNAT] [Slow]", func() {
...
-			for _, ip := range podIPs {
-				if ip == pip {
+		for _, sourcePod := range pods.Items {
+			for _, targetPod := range pods.Items {
+				if targetPod.Name == sourcePod.Name {
 					continue
 				}
-				ips = append(ips, ip)
+				targetAddr := net.JoinHostPort(targetPod.Status.PodIP, testPodPort)
+				sourceIP, execPodIP := execSourceipTest(sourcePod, targetAddr)
```
andrewsykim, line 112: really minor comment: can we refactor the parameters in `execSourceipTest` to be:
```
func execSourceipTest(pod v1.Pod, targetAddress string) (string, string) 
```

Since it no longer for service addresses only.

andrewsykim: and also maybe move it to `test/e2e/network/utils.go`?

knight42: sure thing!

test/e2e/network/util.go
```diff
@@ -125,3 +126,37 @@ func CheckSCTPModuleLoadedOnNodes(f *framework.Framework, nodes *v1.NodeList) bo
...
+		if strings.TrimSpace(stdout) == "" {
+			framework.Logf("got empty stdout, retry until timeout")
+			continue
+		}
+		break
+	}
+
+	framework.ExpectNoError(err)
+
+	// The stdout return from RunHostCmd is in this format: x.x.x.x:port or [xx:xx:xx::x]:port
```
andrewsykim, line 155: I suspect this test may fail in some scenarios with v6 cause of the `[ ]` :thinking: 

Something to watch out for in the future.

test/e2e/network/util.go
```diff
@@ -125,3 +126,37 @@ func CheckSCTPModuleLoadedOnNodes(f *framework.Framework, nodes *v1.NodeList) bo
 	}
 	return false
 }
+
+// execSourceIPTest executes curl to access "/clientip" endpoint on target address
+// from pause Pod to check if source ip is preserved.
+func execSourceIPTest(pausePod v1.Pod, targetAddr string) (string, string) {
```
andrewsykim, line 132: Just `pod` instead of `pausePod`?

test/e2e/network/util.go
knight42: looks like `net.SplitHostPort` would strip the brackets: https://play.golang.org/p/YoTX6bnpXas

andrewsykim: good to know!

andrewsykim (commented):
/approve

Can you squash commits please?

andrewsykim (commented):
/lgtm

knight42:
looks like `pull-kubernetes-verify` is constantly failing now: #93623

PR #93569, opened 2020-07-30
cloud provider: add zone/region to InstanceMetadata

andrewsykim:

**What type of PR is this?**
/kind feature

The existing Zones interface was mainly created for the kubelet
to interface with a local metadata server to fetch a node's zone
and region. With external cloud providers, it makes more sense for
the zone/region logic to be coupled to the Instances interface.
This commit adds zone/region information to the recently added
InstanceMetadata type we're using as part of the new InstancesV2
interface in the cloud provider.

In most cases the zone/region information is contained in the "get instance" API call of a cloud provider. This change would reduce that extra call for zone/region in the cloud provider if the new instance v2 interface is enabled. 

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/cloud-provider/cloud.go
+++ b/staging/src/k8s.io/cloud-provider/cloud.go
@@ -294,4 +294,15 @@ type InstanceMetadata struct {
 	// NodeAddress contains information for the instance's address.
 	// The node addresses returned here will be set on the node's status.addresses field.
 	NodeAddresses []v1.NodeAddress
+
+	// Zone is the zone that the instance is in.
+	// The value set here is applied as the following labels on the node:
+	//   * topology.kubernetes.io/zone=<zone>
+	//   * failure-domain.beta.kubernetes.io/zone=<zone> (DEPRECATED)
+	Zone string
+	// Region is the region that the instance is in.
+	// The value set here is applied as the following labels on the node:
+	//   * topology.kubernetes.io/region=<region>
+	//   * failure-domain.beta.kubernetes.io/region=<region> (DEPRECATED)
+	Region string
 }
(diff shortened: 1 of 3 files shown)


andrewsykim:
Should help with https://github.com/kubernetes/kubernetes/issues/90652

andrewsykim (commented):
/close

staging/src/k8s.io/cloud-provider/cloud.go
```diff
@@ -294,4 +294,15 @@ type InstanceMetadata struct {
 	// NodeAddress contains information for the instance's address.
 	// The node addresses returned here will be set on the node's status.addresses field.
 	NodeAddresses []v1.NodeAddress
+
+	// Zone is the zone that the instance is in.
+	// The value set here is applied as the following labels on the node:
+	//   * topology.kubernetes.io/zone=<zone>
+	//   * failure-domain.beta.kubernetes.io/zone=<zone> (DEPRECATED)
+	Zone string
```
wojtek-t, line 302: I'm fine with adding those (that sounds like a good direction to me), but given that all in-tree providers already support "GetInstanceMetadata", this basically will break supportability of it.

Can you at least add appropriate TODOs for all cloud providers in the InstancesV2 method that those two will have to be propagated before enabling the inferface?

andrewsykim: > I'm fine with adding those (that sounds like a good direction to me), but given that all in-tree providers already support "GetInstanceMetadata", this basically will break supportability of it.

So in https://github.com/kubernetes/kubernetes/pull/93582 I actually removed all the implementations of `InstanceMetadataByProviderID` since that code path isn't actually run anywhere. 

> Can you at least add appropriate TODOs for all cloud providers in the InstancesV2 method that those two will have to be propagated before enabling the inferface?

Also added TODOs comments in https://github.com/kubernetes/kubernetes/pull/93582 for providerse to implement InstanceV2 when out-of-tree.

cheftako: As you correctly point out, this API is still marked as experimental. Do we have a graduation plan? It seems to suggest (but does not explicitly state) we should be graduating in 1.20

Issue #93342, opened 2020-07-22
componentstatuses doesn't work when Scheduler and Controller Manager are running on secure port


mitchellmaler:
**What happened**:

We recently upgraded to 1.17.9 using kubeadm and now the controller manager and scheduler are serving on their default secure port and the old insecure port is 0 which disabled it. Now when you run `kubectl get componentstatuses` it shows that these two components are unhealthy even though they are not. Looking at the output the health check is hitting the old port and using http instead of https.

```
scheduler            Unhealthy   Get http://127.0.0.1:10251/healthz: dial tcp 127.0.0.1:10251: connect: connection refused
controller-manager   Unhealthy   Get http://127.0.0.1:10252/healthz: dial tcp 127.0.0.1:10252: connect: connection refused
```

**What you expected to happen**:

I would expect the componentstatuses to use the secure ports that these components are running on and show they are healthy.

**How to reproduce it (as minimally and precisely as possible)**:

Built a control-plane node making sure that the scheduler and controller-manager are running only their default secure port. After it is started up run the `kubectl get componentstatuses` against that control-plan node.

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): 1.17.9
- Cloud provider or hardware configuration: on-prem vmware
- OS (e.g: `cat /etc/os-release`): Centos 7
- Kernel (e.g. `uname -a`): 3.10.0-1127.10.1.el7.x86_64
- Install tools: kubeadm

liggitt:
The componentstatus implementation does not work in many topologies... it is a hard-coded implementation in the API server that assumes a particular topology for the scheduler/controller manager.

See discussions in https://github.com/kubernetes/kubernetes/issues/19570 and https://github.com/kubernetes/enhancements/issues/553 about why this approach should be deprecated and removed.

Changing the hard-coded values in the kube-apiserver would be as likely to break existing users that currently configure their components in a way that matches the hard-coded expectations.

/cc @deads2k

deads2k:
> The componentstatus implementation does not work in many topologies... it is a hard-coded implementation in the API server that assumes a particular topology (localhost, unsecured) for the scheduler/controller manager.
> 
[...]

PR #82247, opened 2019-09-02
register cloud controller manager when necessary

m3ngyang:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Users can obtain cloud controller status via `kubectl get cs` when the cluster is deployed by a cloud provider.

**Which issue(s) this PR fixes**:

Fixes #74643

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Users can obtain cloud controller status via `kubectl get cs` when the cluster is deployed by a cloud provider, just like `controller-manager`
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

Issue #74643, opened 2019-02-27
Obtain cloud controller status via `kubectl get cs`


m3ngyang:
**What would you like to be added**:
Users can obtain cloud controller status via `kubectl get cs`, for example,
```
# kubectl get cs
NAME                 STATUS    MESSAGE              ERROR
controller-manager   Healthy   ok
scheduler            Healthy   ok
cloud-controller     Healthy   ok
etcd-0               Healthy   {"health": "true"}
etcd-1               Healthy   {"health": "true"}
etcd-2               Healthy   {"health": "true"}

# kubectl get cs cloud-controller -o json
{
    "apiVersion": "v1",
    "conditions": [
        {
            "message": "ok",
            "status": "True",
            "type": "Healthy"
        }
    ],
    "kind": "ComponentStatus",
    "metadata": {
        "creationTimestamp": null,
        "name": "cloud-controller",
        "selfLink": "/api/v1/componentstatuses/cloud-controller"
    }
}
```

**Why is this needed**:
For kubernetes deployed in cloud, `cloud controller` is a key master-like component, as important as `controller-manager`. It status may be included and printed via `kubectl get cs`. 
Currently, `component status` only detects health of `controller-manager` and `scheduler` besides `etcd`s, supposing they are deployed on the same node. Another problem is that how to design an elegant way to detect health of `cloud controller` because it can be deployed in different ways by cloud providers.

https://github.com/kubernetes/kubernetes/blob/38a325250fbefa8785740d00358978eefa160dde/pkg/registry/core/rest/storage_core.go#L255-L258

liggitt:
See https://github.com/kubernetes/kubernetes/pull/82247#discussion_r322381376 and https://github.com/kubernetes/enhancements/issues/553#issuecomment-529591505 for why this behavior does not belong in the kube-apiserver

Issue #19570, opened 2016-01-12
componentstatus fails when components are not running on the same host as apiserver


stepanstipl:
`kubectl get componentstatus` fails when components are not running on the same host (for example when you run multiple Api servers according to HA guide http://kubernetes.io/v1.1/docs/admin/high-availability.html).

I can see the problem being in `kubernetes/pkg/master/master.go` where code expects `127.0.0.1` for both controller-manager and scheduler components.

```
func (m *Master) getServersToValidate(c *Config) map[string]apiserver.Server {
    serversToValidate := map[string]apiserver.Server{
        "controller-manager": {Addr: "127.0.0.1", Port: ports.ControllerManagerPort, Path: "/healthz"},
        "scheduler":          {Addr: "127.0.0.1", Port: ports.SchedulerPort, Path: "/healthz"},
    }
```

```
$kubectl version                                                                                                                                   
Client Version: version.Info{Major:"1", Minor:"1", GitVersion:"v1.1.3+6a81b50", GitCommit:"6a81b50c7e97bbe0ade075de55ab4fa34f049dc2", GitTreeState:"not a git tree"}
Server Version: version.Info{Major:"1", Minor:"1", GitVersion:"v1.1.3", GitCommit:"6a81b50c7e97bbe0ade075de55ab4fa34f049dc2", GitTreeState:"clean"}
```

lavalamp:
@karlkfi is reworking componentstatus, it's not obvious how to fix this assumption in the meantime.

stepanstipl:
Thanks for info, not a big deal, just spent some time worried that my cluster is not working before I discovered that this was the problem... I would like to help, just not sure how with this...?

Maybe a bit of info about my setup: I'm running 3 api servers on AWS behind ELB and using podmaster service to control which of them runs scheduler and controller-manager, pretty much what's in HA guide.

On my setup the check will always succeed once and fail twice, depending on which api server will be queried behind the ELB.

karlkfi:
for reference:
- https://github.com/kubernetes/kubernetes/issues/18610
- https://github.com/kubernetes/kubernetes/pull/13216

gg7:
`componentstatus` will fail even if components are running on the same host as the apiserver, but not listening on the loopback interface.

Exact error so this is easier to google:

```
NAME                 STATUS      MESSAGE                                                                                        ERROR
[...]

PR #93570, opened 2020-07-30
Mark componentstatus as deprecated

liggitt:
**What type of PR is this?**
/kind api-change
/kind deprecation

xref https://github.com/kubernetes/enhancements/issues/553#issuecomment-529591505 and  https://github.com/kubernetes/kubernetes/issues/19570

The current state of this API is problematic, and requires reversing the actual data flow (it requires the API server to call to its clients), and is not functional across deployment topologies.

Leaving it in place attracts new attempts to make additions to it (https://github.com/kubernetes/kubernetes/issues/74643, kubernetes/kubernetes#82247) and leads to confusion or bug reports for deployments that do not match its topology assumptions (https://github.com/kubernetes/kubernetes/issues/93342, https://github.com/kubernetes/kubernetes/issues/93472). It should be clearly marked as deprecated.

```release-note
kube-apiserver: the componentstatus API is deprecated. This API provided status of etcd, kube-scheduler, and kube-controller-manager components, but only worked when those components were local to the API server, and when kube-scheduler and kube-controller-manager exposed unsecured health endpoints. Instead of this API, etcd health is included in the kube-apiserver health check and kube-scheduler/kube-controller-manager health checks can be made directly against those components' health endpoints.
```

/cc @deads2k @lavalamp @neolit123 
/sig api-machinery cluster-lifecycle

### DIFF
--- a/pkg/apis/core/types.go
+++ b/pkg/apis/core/types.go
@@ -5109,6 +5109,7 @@ type ComponentCondition struct {
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // ComponentStatus (and ComponentStatusList) holds the cluster validation info.
+// Deprecated: This API is deprecated in v1.19+
 type ComponentStatus struct {
 	metav1.TypeMeta
 	// +optional
@@ -5121,6 +5122,7 @@ type ComponentStatus struct {
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // ComponentStatusList represents the list of component statuses
+// Deprecated: This API is deprecated in v1.19+
 type ComponentStatusList struct {
 	metav1.TypeMeta
 	// +optional
--- a/staging/src/k8s.io/api/core/v1/BUILD
+++ b/staging/src/k8s.io/api/core/v1/BUILD
@@ -21,6 +21,7 @@ go_library(
         "annotation_key_constants.go",
         "doc.go",
         "generated.pb.go",
+        "lifecycle.go",
         "objectreference.go",
         "register.go",
         "resource.go",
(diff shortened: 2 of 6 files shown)


dims:
@liggitt when will we be able to remove it? (if at all?)

liggitt:
> @liggitt when will we be able to remove it? (if at all?)

I'm not sure, but communicating the lack of continued development and known inconsistent behavior of what is currently there is important.

neolit123:
thanks!
/lgtm

lavalamp:
/unhold

I'll get it eventually...
/joke

liggitt:
I'd actually pull this into 1.19 … we should have communicated this a year ago, imo

lavalamp:
It's not a functional change, so it sounds safe. It's vaguely safety-focused, so I guess I wouldn't block it on the grounds the tests need to be fixed first.

liggitt:
adjusted to 1.19

dims:
/lgtm

reapply previous `/lgtm`

YanzhaoLi:
Hi, any idea for alternative method to check status?

Napsty:
> Instead of this API, etcd health is included in the kube-apiserver health check and kube-scheduler/kube-controller-manager health checks can be made directly against those components' health endpoints.

Any example on this would be greatly appreciated. ;-)

eumel8:
Now it comes nearly to an end for the componentstatus. If you headed here from Google and you are a Rancher user you can use Rancher management API to ask the status:

```bash
$ kubectl get clusters.management.cattle.io <your-cluster-id> -o json | jq  '.status.componentStatuses[] | .name,.conditions[].message'
"controller-manager"
"ok"
"etcd-0"
"{\"health\":\"true\"}"
"etcd-1"
"{\"health\":\"true\"}"
"etcd-2"
"{\"health\":\"true\"}"
"scheduler"
"ok"
```

plain curl:

```bash
$ curl -s -H "Content-Type: application/json" -H "authorization: Bearer <token>"  https://<rancher-server>/k8s/clusters/local/apis/management.cattle.io/v3/clusters/<your-cluster-id> | jq '.status.componentStatuses[] | .name,.conditions[].message'
"controller-manager"
"ok"
"etcd-0"
"{\"health\":\"true\"}"
"etcd-1"
"{\"health\":\"true\"}"
"etcd-2"
"{\"health\":\"true\"}"
"scheduler"
"ok"'
```

PR #93571, opened 2020-07-30
Document blocking behaviour of RunOrDie and Run

ChrsMark:
**What type of PR is this?**
/kind documentation

`RunOrDie` being a blocking function without mentioning might be confusing. 

`RunOrDie` starts a `LeaderElector` at https://github.com/kubernetes/kubernetes/blob/72a62bcade268183e7a1a5d40b4d2663c30507ed/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go#L222 which will not return, blocking at https://github.com/kubernetes/kubernetes/blob/72a62bcade268183e7a1a5d40b4d2663c30507ed/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go#L203 or at https://github.com/kubernetes/kubernetes/blob/72a62bcade268183e7a1a5d40b4d2663c30507ed/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go#L209

If anyone need to avoid this blocking behaviour need to call `RunOrDie` in async mode with `go RunOrDie(...)`

```release-note
NONE
```

Related to https://github.com/kubernetes/client-go/issues/837, cc: @liggitt

### DIFF
--- a/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go
+++ b/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go
@@ -193,7 +193,9 @@ type LeaderElector struct {
 	name string
 }
 
-// Run starts the leader election loop
+// Run starts the leader election loop. Run will not return
+// before leader election loop is stopped by ctx or it has
+// stopped holding the leader lease
 func (le *LeaderElector) Run(ctx context.Context) {
 	defer runtime.HandleCrash()
 	defer func() {
@@ -210,7 +212,8 @@ func (le *LeaderElector) Run(ctx context.Context) {
 }
 
 // RunOrDie starts a client with the provided config or panics if the config
-// fails to validate.
+// fails to validate. RunOrDie blocks until leader election loop is
+// stopped by ctx or it has stopped holding the leader lease
 func RunOrDie(ctx context.Context, lec LeaderElectionConfig) {
 	le, err := NewLeaderElector(lec)
 	if err != nil {


staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go
```diff
@@ -210,7 +210,8 @@ func (le *LeaderElector) Run(ctx context.Context) {
 }
 
 // RunOrDie starts a client with the provided config or panics if the config
-// fails to validate.
+// fails to validate. RunOrDie blocks until leader election loop is 
+// is stopped by ctx.
```
liggitt, line 214: add similar documentation to `Run`?

ChrsMark: Updated @liggitt . Let me know what you think.

staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go
```diff
@@ -193,7 +193,8 @@ type LeaderElector struct {
 	name string
 }
 
-// Run starts the leader election loop
+// Run starts the leader election loop. Run will not return
+// before leader election loop is stopped by ctx
```
liggitt, line 197: or until it loses leader, right?

staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go
```diff
@@ -210,7 +211,8 @@ func (le *LeaderElector) Run(ctx context.Context) {
 }
 
 // RunOrDie starts a client with the provided config or panics if the config
-// fails to validate.
+// fails to validate. RunOrDie blocks until leader election loop is 
+// is stopped by ctx.
```
liggitt, line 215: typo: is is

also, if it loses leader, it will return, right?

staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go
ChrsMark: 👍  seems that yes.

staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go
ChrsMark: 👍

liggitt:
looks good, go ahead and squash to a single commit

ChrsMark:
> looks good, go ahead and squash to a single commit

Done

ChrsMark:
@liggitt there was a failing test and I had to run `./hack/update-gofmt.sh` and commit again, so whenever you get the chance please give a `lgtm`.

liggitt:
/lgtm

verify failure was fixed in master
/retest

ChrsMark:
@liggitt  thanks! Is there anything else we need to do for this PR or it will be merged when it's the proper time for its milestone?

liggitt:
this is all set, once master reopens for 1.20 it will merge

liggitt:

PR #93572, opened 2020-07-30
Limit typecheck parallelism by default

### LINKED ISSUE #93532: typecheck uses exorbitant amounts of memory

liggitt:
**What type of PR is this?**
/kind cleanup

typecheck currently uses ~5GB of memory per platform check. This limits the parallel checks by default so we don't require ~50GB of memory to run this job and adds a parallel option to tune that further.

Alternative to https://github.com/kubernetes/kubernetes/pull/93544

Fixes #93532

```release-note
NONE
```

/cc @hasheddan @BenTheElder

### DIFF
--- a/test/typecheck/main.go
+++ b/test/typecheck/main.go
@@ -38,7 +38,8 @@ var (
 	platforms  = flag.String("platform", "", "comma-separated list of platforms to typecheck")
 	timings    = flag.Bool("time", false, "output times taken for each phase")
 	defuses    = flag.Bool("defuse", false, "output defs/uses")
-	serial     = flag.Bool("serial", false, "don't type check platforms in parallel")
+	serial     = flag.Bool("serial", false, "don't type check platforms in parallel (equivalent to --parallel=1)")
+	parallel   = flag.Int("parallel", 4, "limits how many platforms can be checked in parallel. 0 means no limit.")
 	skipTest   = flag.Bool("skip-test", false, "don't type check test code")
 	tags       = flag.String("tags", "", "comma-separated list of build tags to apply in addition to go's defaults")
 	ignoreDirs = flag.String("ignore-dirs", "", "comma-separated list of directories to ignore in addition to the default hardcoded list including staging, vendor, and hidden dirs")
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
/lgtm

BenTheElder:
Thank you! Still a lovely ~10m presubmit, not bad!

PR #92188, opened 2020-06-16
revert InstanceMetadataByProviderID definition and related deprecations

gongguan:
Ref [#90652](https://github.com/kubernetes/kubernetes/issues/90652)

/kind cleanup

**Special notes for your reviewer**:
See https://github.com/kubernetes/kubernetes/pull/90894#issuecomment-633707002

```release-note
None
```

PR #90894, opened 2020-05-08
Implement InstanceMetadataByProviderID

gongguan:
Ref [#90652](https://github.com/kubernetes/kubernetes/issues/90652)

**Special notes for your reviewer**:
Implement `InstanceMetadataByProviderID`

`InstanceMetadataByProviderID` used to get instance providerID, type and nodeAddressed in one function to reduce the number of calls to cloud API.
**Does this PR introduce a user-facing change?**:

```release-note
None
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #93582, opened 2020-07-30
Mark cloud provider InstanceV2 as experimental and remove provider ID references

andrewsykim:
**What type of PR is this?**
/kind bug

In https://github.com/kubernetes/kubernetes/pull/90894, a new method InstanceMetadataByProviderID was added to the cloud provider Instances interface. This was later reverted in https://github.com/kubernetes/kubernetes/pull/92188 and re-added into a new v2 interface in https://github.com/kubernetes/kubernetes/pull/92367. I had not realized that the existing implementations of `InstanceMetadataByProviderID` were left here even though the `InstanceV2` gate was turned off. We should remove the implementation added in https://github.com/kubernetes/kubernetes/pull/90894 and let the individual providers implement it as they enable instancesV2. The existing implementations of `InstanceMetadataByProviderID` are also unused since it only applies to external cloud providers. 

This PR also makes breaking changes to a new public API we added for external cloud providers in v1.19. It moves us away from referencing "provider IDs" and instead just passes the node object for every method. This is the desired approach since we can defer to the cloud provider to use either a node name, provider ID, or annotations/labels to discover a node. Since InstancesV2 was added in v1.19, it would be ideal to merge this PR in v1.19 to ensure we don't have to make breaking changes or deprecations in a future release.
  
In addition I'd like to mark InstancesV2 as "experimental" for v1.19 so we are free to make some breaking changes in v1.20. Though I don't suspect there will be much more churn after v1.20. 

Removal of `InstanceMetadataByProviderID` is safe because the `InstancesV2` gate is disabled for all in-tree providers.

```release-note
NONE
```

```docs

```


staging/src/k8s.io/legacy-cloud-providers/aws/aws.go
```diff
@@ -1670,31 +1671,6 @@ func (c *Cloud) InstanceShutdownByProviderID(ctx context.Context, providerID str
 	return false, nil
 }
 
-// InstanceMetadataByProviderID returns metadata of the specified instance.
-func (c *Cloud) InstanceMetadataByProviderID(ctx context.Context, providerID string) (*cloudprovider.InstanceMetadata, error) {
```
andrewsykim, line 1674: Safe to remove since InstanceV2 is disabled https://github.com/kubernetes/kubernetes/blob/d3c1e81d5e6796db9389cd73073f48d6a37c4621/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L1397-L1400

staging/src/k8s.io/legacy-cloud-providers/azure/azure_instances.go
```diff
@@ -239,80 +239,6 @@ func (az *Cloud) InstanceShutdownByProviderID(ctx context.Context, providerID st
 	return status == vmPowerStateStopped || status == vmPowerStateDeallocated || status == vmPowerStateDeallocating, nil
 }
 
-// InstanceMetadataByProviderID returns metadata of the specified instance.
-// InstanceMetadataByProviderID is part of InstancesV2 interface and is only used in cloud node-controller.
-func (az *Cloud) InstanceMetadataByProviderID(ctx context.Context, providerID string) (*cloudprovider.InstanceMetadata, error) {
```
andrewsykim, line 244: Safe to remove since InstanceV2 is disabled: https://github.com/kubernetes/kubernetes/blob/d3c1e81d5e6796db9389cd73073f48d6a37c4621/staging/src/k8s.io/legacy-cloud-providers/azure/azure.go#L667-L670

staging/src/k8s.io/legacy-cloud-providers/gce/gce_instances.go
```diff
@@ -210,36 +210,6 @@ func (g *Cloud) InstanceShutdownByProviderID(ctx context.Context, providerID str
 	return false, cloudprovider.NotImplemented
 }
 
-// InstanceMetadataByProviderID returns metadata of the specified instance.
-func (g *Cloud) InstanceMetadataByProviderID(ctx context.Context, providerID string) (*cloudprovider.InstanceMetadata, error) {
```
andrewsykim, line 214: Safe to remove since instance v2 is disabled: 

https://github.com/kubernetes/kubernetes/blob/d3c1e81d5e6796db9389cd73073f48d6a37c4621/staging/src/k8s.io/legacy-cloud-providers/gce/gce.go#L662-L664

staging/src/k8s.io/legacy-cloud-providers/openstack/openstack_instances.go
```diff
@@ -157,37 +158,6 @@ func (i *Instances) InstanceShutdownByProviderID(ctx context.Context, providerID
 	return false, nil
 }
 
-// InstanceMetadataByProviderID returns metadata of the specified instance.
-func (i *Instances) InstanceMetadataByProviderID(ctx context.Context, providerID string) (*cloudprovider.InstanceMetadata, error) {
```
andrewsykim, line 161: Safe to remove since instance v2 is disabled:

https://github.com/kubernetes/kubernetes/blob/d3c1e81d5e6796db9389cd73073f48d6a37c4621/staging/src/k8s.io/legacy-cloud-providers/openstack/openstack_instances.go#L65-L68

staging/src/k8s.io/legacy-cloud-providers/vsphere/vsphere.go
```diff
@@ -797,17 +798,6 @@ func (vs *VSphere) InstanceShutdownByProviderID(ctx context.Context, providerID
 	return !isActive, nil
 }
 
-// InstanceMetadataByProviderID returns metadata of the specified instance.
-func (vs *VSphere) InstanceMetadataByProviderID(ctx context.Context, providerID string) (*cloudprovider.InstanceMetadata, error) {
```
andrewsykim, line 801: Safe to remove since InstancesV2 is disabled:

https://github.com/kubernetes/kubernetes/blob/d3c1e81d5e6796db9389cd73073f48d6a37c4621/staging/src/k8s.io/legacy-cloud-providers/vsphere/vsphere.go#L563-L566

andrewsykim:
I'd like to put this v1.19 actually, pending some discussion with @cheftako

dims:
LGTM, but let's wait for @liggitt tomorrow

/assign @liggitt

liggitt:
> We should remove the implementation added in #90894 and let the individual providers implement it as they enable instancesV2. The existing implementations of InstanceMetadataByProviderID are also unused since it only applies to external cloud providers.

I agree with removing unused implementations added in https://github.com/kubernetes/kubernetes/pull/90894 to complete the interface revert in https://github.com/kubernetes/kubernetes/pull/92188. If the methods aren't referenced today, that seems like a zero-risk change.

> This PR also makes breaking changes to a new public API we added for external cloud providers in v1.19. It moves us away from referencing "provider IDs" and instead just passes the node object for every method. This is the desired approach since we can defer to the cloud provider to use either a node name or the provider ID to discover a node. Since InstancesV2 was added in v1.19, it would be ideal to merge this PR in v1.19 to ensure we don't have to make breaking changes or deprecations in a future release.
> 
> In addition I'd like to mark InstancesV2 as "experimental" for v1.19 so we are free to make some breaking changes in v1.20.

I'm less clear on these. Making last-minute interface changes to avoid breaking go interface compatibility, while explicitly adding a note that the interface is experimental and can be broken in the future seems strange to me.

andrewsykim:
> I'm less clear on these. Making last-minute interface changes to avoid breaking go interface compatibility, while explicitly adding a note that the interface is experimental and can be broken in the future seems strange to me.

Good point -- my thought was we should make the breaking changes we want now to prevent breaking changes in v1.20, and since instancesv2 requires an explciti opt-in, the warning is still valid for cloud provider implementors. I expect most of them won't even opt-into this in v1.19.

https://github.com/kubernetes/cloud-provider/blob/c3ccadc5014b01f937700a5ecfb7d78b690d325e/cloud.go#L52-L54

I'm not sure what better way we could mark a Go interface as "experimental", happy to remove the warning if we find it causes more confusion.

liggitt:
the case for reverting the unused implementations seems really clear.

the interface change I'm less sure about (passing in the entire Node object is a significant expansion of the interface surface area and makes calling the methods without a node instance in hand a lot harder), but will defer to the cloud provider owners

PR #93583, opened 2020-07-30
support ipv6 in e2e policy tests

aojea:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind deprecation
> /kind design
> /kind documentation

/kind failing-test

> /kind feature
> /kind flake

Network policy tests are not testing the feature correctly for IPv6

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/network/network_policy.go
+++ b/test/e2e/network/network_policy.go
@@ -19,6 +19,8 @@ package network
 import (
 	"context"
 	"encoding/json"
+	"fmt"
+	"net"
 
 	v1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
@@ -31,8 +33,7 @@ import (
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
 	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
 	imageutils "k8s.io/kubernetes/test/utils/image"
-
-	"fmt"
+	utilnet "k8s.io/utils/net"
 
 	"github.com/onsi/ginkgo"
 )
@@ -1373,8 +1374,11 @@ var _ = SIGDescribe("NetworkPolicy [LinuxOnly]", func() {
 			if err != nil {
 				framework.ExpectNoError(err, "Error occurred while getting pod status.")
 			}
-
-			podServerCIDR := fmt.Sprintf("%s/32", podServerStatus.Status.PodIP)
+			hostMask := 32
+			if utilnet.IsIPv6String(podServerStatus.Status.PodIP) {
+				hostMask = 128
+			}
+			podServerCIDR := fmt.Sprintf("%s/%d", podServerStatus.Status.PodIP, hostMask)
 
 			// Creating pod-b and service-b
 			podServerB, serviceB = createServerPodAndService(f, f.Namespace, "pod-b", []int{80})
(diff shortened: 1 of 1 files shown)


test/e2e/network/network_policy.go
```diff
@@ -1450,9 +1453,15 @@ var _ = SIGDescribe("NetworkPolicy [LinuxOnly]", func() {
...
 			}
 
-			podServerAllowCIDR := fmt.Sprintf("%s/24", podServerStatus.Status.PodIP)
+			allowMask := 24
+			exceptMask := 32
+			if utilnet.IsIPv6String(podServerStatus.Status.PodIP) {
+				allowMask = 64
+				exceptMask = 128
+			}
+			podServerAllowCIDR := fmt.Sprintf("%s/%d", podServerStatus.Status.PodIP, allowMask)
```
danwinship, line 1462: oops, we're testing undefined behavior anyway; we're saying to allow to, eg, "192.168.1.5/24" when we should be saying to allow to "192.168.1.**0**/24".

You need to do:

    _, podServerAllowSubnet, err := net.ParseCIDR(fmt.Sprintf("%s/%d", podServerStatus.Status.PodIP, allowMask))
    framework.ExpectNoError(err, "could not parse allow subnet")
    podServerAllowCIDR := podServerAllowSubnet.String()

likewise in the test below

test/e2e/network/network_policy.go
```diff
@@ -1515,8 +1524,14 @@ var _ = SIGDescribe("NetworkPolicy [LinuxOnly]", func() {
...
+			podServerIP := fmt.Sprintf("%s/%d", podServerStatus.Status.PodIP, exceptMask)
 			// Exclude podServer's IP with an Except clause
 			podServerExceptList := []string{podServerIP}
```
danwinship, line 1536: It's not really an "IP" when it has a prefix length. Maybe just get rid of the `podServerIP` variable and do this like in the above test

aojea: yep

test/e2e/network/network_policy.go
aojea: yeah, strictly speaking that's correct, just curiosity, do you think that app should "normalize it" and use it or reject it?

danwinship: There's no way for us to know whether the user actually meant "192.168.1.5/**32**" or "192.168.1.**0**/24"; either mistake is possible. So we can't "normalize it". Ideally we would have rejected this at validation time.

test/e2e/network/network_policy.go
```diff
@@ -1373,8 +1374,13 @@ var _ = SIGDescribe("NetworkPolicy [LinuxOnly]", func() {
...
 			}
-
-			podServerCIDR := fmt.Sprintf("%s/32", podServerStatus.Status.PodIP)
+			mask := 32
+			if utilnet.IsIPv6String(podServerStatus.Status.PodIP) {
+				mask = 128
+			}
+			_, podServerSubnet, err := net.ParseCIDR(fmt.Sprintf("%s/%d", podServerStatus.Status.PodIP, mask))
+			framework.ExpectNoError(err, "could not parse pod subnet")
+			podServerCIDR := podServerSubnet.String()
```
danwinship, line 1383: you don't need to do this in the 32/128 case; in that case we're keeping all the bits of the pod IP, so just `podServerCIDR := fmt.Sprintf("%s/%d", podServerStatus.Status.PodIP, mask)` is fine

aojea:
@danwinship I totally forget about this one, addressed the comments
PTAL

PR #93597, opened 2020-07-31
Wait for both endpoints and endpointslices in e2e tests

wojtek-t:
Ref https://github.com/kubernetes/kubernetes/issues/91236

```release-note
NONE
```

/kind flake

### DIFF
--- a/test/e2e/framework/endpointslice/ports.go
+++ b/test/e2e/framework/endpointslice/ports.go
@@ -14,10 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package endpoints
+package endpointslice
 
 import (
-	v1 "k8s.io/api/core/v1"
+	discoveryv1beta1 "k8s.io/api/discovery/v1beta1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
@@ -25,16 +25,20 @@ import (
 type PortsByPodUID map[types.UID][]int
 
 // GetContainerPortsByPodUID returns a PortsByPodUID map on the given endpoints.
-func GetContainerPortsByPodUID(ep *v1.Endpoints) PortsByPodUID {
+func GetContainerPortsByPodUID(eps []discoveryv1beta1.EndpointSlice) PortsByPodUID {
 	m := PortsByPodUID{}
-	for _, ss := range ep.Subsets {
-		for _, port := range ss.Ports {
-			for _, addr := range ss.Addresses {
-				containerPort := port.Port
-				if _, ok := m[addr.TargetRef.UID]; !ok {
-					m[addr.TargetRef.UID] = make([]int, 0)
+
+	for _, es := range eps {
+		for _, port := range es.Ports {
+			if port.Port == nil {
+				continue
+			}
+			for _, ep := range es.Endpoints {
+				containerPort := *port.Port
+				if _, ok := m[ep.TargetRef.UID]; !ok {
+					m[ep.TargetRef.UID] = make([]int, 0)
 				}
-				m[addr.TargetRef.UID] = append(m[addr.TargetRef.UID], int(containerPort))
+				m[ep.TargetRef.UID] = append(m[ep.TargetRef.UID], int(containerPort))
 			}
 		}
 	}
--- a/test/e2e/network/service.go
+++ b/test/e2e/network/service.go
@@ -37,6 +37,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	discoveryv1beta1 "k8s.io/api/discovery/v1beta1"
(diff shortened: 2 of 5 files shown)


test/e2e/network/service.go
```diff
@@ -3779,12 +3783,29 @@ func validateEndpointsPorts(c clientset.Interface, namespace, serviceName string
 			// Retry the error
 			return false, nil
 		}
-		portsByPodUID := e2eendpoints.GetContainerPortsByPodUID(ep)
+		endpointsPortsByPodUID := portsByPodUID(e2eendpoints.GetContainerPortsByPodUID(ep))
+
+		opts := metav1.ListOptions{
+			LabelSelector: "kubernetes.io/service-name=" + serviceName,
```
aojea, line 3789: TIL from now  on we have to use List and labels always, we can not do get

aojea:
/hold
to check the test results, 
feel free to unhold when needed

test/e2e/network/service.go
```diff
@@ -3779,12 +3783,29 @@ func validateEndpointsPorts(c clientset.Interface, namespace, serviceName string
...
+		}
+		es, err := c.DiscoveryV1beta1().EndpointSlices(namespace).List(context.TODO(), opts)
+		if err != nil {
+			framework.Logf("Failed go list EndpointSlice objects: %v", err)
```
hasheddan, line 3793: Hmm looks like this is causing an issue in `pull-kubernetes-conformance-kind-ga-only-parallel`: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93597/pull-kubernetes-conformance-kind-ga-only-parallel/1289156168177946628#1:build-log.txt%3A12459

Used here: https://github.com/kubernetes/kubernetes/blob/ee72e029070128447c673b75dbaedba978e31f48/test/e2e/network/service.go#L823

wojtek-t: This is interesting - the test claims that "EndpointSlice" type is not registered there, which doesn't make much sense...

liggitt:
This must only check endpointslices if the API exists. The GA-only conformance job disables all beta APIs and features, so it is using endpoints only

wojtek-t:
> This must only check endpointslices if the API exists. The GA-only conformance job disables all beta APIs and features, so it is using endpoints only

Thanks Jordan - will fix in a couple minutes.

wojtek-t:
@aojea @liggitt - PTAL

aojea:
/lgtm
defer to Jordan to unhold it :smile:

hasheddan (approved):
/lgtm
/milestone v1.19

wojtek-t:
/retest
[unrelated node-e2e failure]

test/e2e/network/service.go
```diff
@@ -3773,21 +3778,42 @@ func validateEndpointsPorts(c clientset.Interface, namespace, serviceName string
...
 			if i%5 == 0 {
-				framework.Logf("Unexpected endpoints: found %v, expected %v, will retry", portsByPodUID, expectedEndpoints)
+				framework.Logf("Unexpected endpoints: found %v, expected %v, will retry", portsByUID, expectedEndpoints)
 			}
 			return false, nil
 		}
+
+		// If EndpointSlice API is enabled, then validate if appropriate EndpointSlice objects
+		// were also create/updated/deleted.
+		if _, err := c.Discovery().ServerResourcesForGroupVersion(discoveryv1beta1.SchemeGroupVersion.String()); err == nil {
```
aojea, line 3799: I think that there is a trailing whitespace in this line and gofmt is complaining

test/e2e/framework/endpointslice/ports.go
```diff
@@ -0,0 +1,46 @@
...
+	for _, es := range eps {
+		for _, port := range es.Ports {
+			if port.Port == nil {
+				continue
+			}
+			for _, ep := range es.Endpoints {
+				containerPort := *port.Port
+				if _, ok := m[ep.TargetRef.UID]; !ok {
+					m[ep.TargetRef.UID] = make([]int, 0)
+				}
```
liggitt, line 40: nit: these lines aren't needed, can append to a nil slice (https://play.golang.org/p/pKdx39pAfLs)

no need to change this unless you need to touch the commit for other reasons

wojtek-t:
Unrelated failure on up.
/retest

aojea:

[sig-storage] CSI mock volume CSI online volume expansion should expand volume without restarting pod if attach=on, nodeExpansion=on expand_less

wojtek-t:
I've never seen 3 different people retesting my PR within 3 hours so far :-)

hasheddan:
`pull-kubernetes-verify` has an issue and will continue to fail. Working on a fix 😕

liggitt:

/hold

PR #93602, opened 2020-07-31
Update k8s.io/utils

### LINKED ISSUE #93531: Trace library seem to incorrect print start_time
While debugging scalability tests, I discovered that there seem to be some problem with tracing library.

Taking this as an example run:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability/1287819459200815104

and grepping for all Trace entires, all seem to have start_time where seconds are always set to 0.
i.e. the start time is always"

`27-Jul-2020 XX:XX:00.XXX`

That makes debugging harder

/kind bug

@jpbetz @dims @liggitt

liggitt:
**What type of PR is this?**
/kind bug

Fixes #93531

Picks up https://github.com/kubernetes/utils/pull/182

Also picks up https://github.com/kubernetes/utils/pull/179, so cc @kubernetes/sig-storage-pr-reviews @kubernetes/sig-windows-misc for review

```release-note
NONE
```

/cc @wojtek-t


liggitt:
unit test flake is https://github.com/kubernetes/kubernetes/issues/93604
node e2e is #93210

/retest

PR #93264, opened 2020-07-20
Update Golang to v1.15.0-rc.1

justaugustus:
**What type of PR is this?**

/kind cleanup
/area dependency release-eng
/sig release

**What this PR does / why we need it**:

Update Golang to v1.15.0

Depends on fixes in earlier PRs (which must also be included if go1.15 support is backported):
- [x] https://github.com/kubernetes/kubernetes/pull/93425 (int->string casts)
- [x] https://github.com/kubernetes/kubernetes/pull/93426 (add subjectAltName to serving certificates)

Followups (which must also be done in earlier releases if go1.15 support is backported):
- [x] update CI images to go1.15rc1 - https://github.com/kubernetes/test-infra/pull/18475
- [ ] update CI jobs to use go1.15.rc1 images (bot PR)
- [x] update lint-dependencies to stop exempting golang.org/x/... deps from minimal version selection per https://github.com/kubernetes/kubernetes/pull/93264#issuecomment-663933730 - https://github.com/kubernetes/kubernetes/pull/93603
- [ ] once 1.15.0 is released, set `minimum_go_version=go1.15` in `hack/lib/golang.sh`

Signed-off-by: Stephen Augustus <saugustus@vmware.com>

cc: @kubernetes/release-engineering @kubernetes/release-team-leads @kubernetes/sig-scalability-feature-requests @liggitt @dims 
Tracking issue: https://github.com/kubernetes/release/issues/1421

**Special notes for your reviewer**:

This is testing against a pre-release version of go1.15 (`go1.15rc1`).
~Expect a slew of errors, as various parts of our code are not configured to accept Go pre-release versions.~

~Also, there will definitely be bazel toolchain + rules_go errors.~

**Does this PR introduce a user-facing change?**:

```release-note
- Kubernetes is now built with golang 1.15.0-rc.1.
- The deprecated, legacy behavior of treating the CommonName field on X.509 serving certificates as a host name when no Subject Alternative Names are present is now disabled by default. It can be temporarily re-enabled by adding the value x509ignoreCN=0 to the GODEBUG environment variable.
```

PR #93603, opened 2020-07-31
Remove golang.org exception for preferring latest minimal version

liggitt:
**What type of PR is this?**
/kind cleanup

Removes the golang.org exception for preferring the latest minimal version, per https://github.com/kubernetes/kubernetes/pull/93264#issuecomment-663933730

```release-note
NONE
```

/cc @justaugustus


liggitt:
/sig release
/priority important-longterm

/hold

PR #93605, opened 2020-07-31
Stop ignoring unit test flakes

liggitt:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:

* Stops ignoring unit test flakes. They mask real test issues and bugs.
* Increases the number of times a test must pass from 1/3 to 3

Flakes and bugs found via this PR:

Fixed + merged:
- [x] Make toKubeContainerImageSpec deterministic - #93610
- [x] Test_NodesDeleted data race - https://github.com/kubernetes/kubernetes/pull/93605#issuecomment-667407692, https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93605/pull-kubernetes-bazel-test/1289323742211608581 - https://github.com/kubernetes/kubernetes/pull/93619
- [x] CSR fuzzing did not match defaulting - https://github.com/kubernetes/kubernetes/pull/93614
- [x] TestControllerSync - https://github.com/kubernetes/kubernetes/pull/93605#issuecomment-667428441, https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93605/pull-kubernetes-bazel-test/1289323742211608581 - https://github.com/kubernetes/kubernetes/pull/93655
- [x] Daemonset controller failing consistently - https://github.com/kubernetes/kubernetes/issues/93604 / https://github.com/kubernetes/kubernetes/pull/93617
- [x] TestUpdateNodeStatusWithLease race - https://github.com/kubernetes/kubernetes/pull/93605#issuecomment-668141263 - https://github.com/kubernetes/kubernetes/issues/93715, https://github.com/kubernetes/kubernetes/pull/93717
- [x] TestEnsureBackendPoolDeletedConcurrently race - https://github.com/kubernetes/kubernetes/pull/93846
- [x] timeout on //pkg/scheduler - TestSchedulerMultipleProfilesScheduling - https://github.com/kubernetes/kubernetes/pull/93605#issuecomment-668127993, fixed in https://github.com/kubernetes/kubernetes/pull/93893
- [x] TestWatchHTTPTimeout - https://github.com/kubernetes/kubernetes/pull/93605#issuecomment-668332578 - https://github.com/kubernetes/kubernetes/issues/93891, fixed in https://github.com/kubernetes/kubernetes/pull/93895
- [x] (bug) TestPluginRegistration timeout - https://github.com/kubernetes/kubernetes/pull/93605#issuecomment-667316127 - #93622
[...]

PR #90061, opened 2020-04-10
Add annotations to CRI ImageSpec objects

marosset:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

**What this PR does / why we need it**:
This PR adds Annotations to the ImageSpec object in CRI api.

The expectation is that for Windows in order to use non-default runtime classes Pods would specify a `kubernetse.io/runtimehandler` annotation. This would get passed to the CRI and used during various image pull and sandbox creation operations.
This behavior would be optional and would need to be implemented for each CRI independently.

Usage and more information can be found here: https://github.com/kubernetes/enhancements/blob/master/keps/sig-windows/windows-runtimeclass-support.md#adding-annotations-to-imagespec 

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Add Annotations to CRI-API ImageSpec objects.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

KEP: 

```docs
[KEP]: https://github.com/kubernetes/enhancements/blob/master/keps/sig-windows/windows-runtimeclass-support.md
```

PR #93610, opened 2020-07-31
Make toKubeContainerImageSpec deterministic

liggitt:
The annotations added in https://github.com/kubernetes/kubernetes/pull/90061 in 1.19 are in non-deterministic order, because they are built by iterating over a map.

The unit test expects an exact order. Failure seen in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93605/pull-kubernetes-bazel-test/1289254188986929154

Bazel unit tests are currently running up to 3 times, masking this flake. Found this in https://github.com/kubernetes/kubernetes/pull/93605

In addition to stabilizing the test, this ensures anything downstream of this method has deterministic output for equality comparison

/kind bug
/sig node
/milestone v1.19

### DIFF
--- a/pkg/kubelet/kuberuntime/convert.go
+++ b/pkg/kubelet/kuberuntime/convert.go
@@ -17,6 +17,8 @@ limitations under the License.
 package kuberuntime
 
 import (
+	"sort"
+
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 	kubecontainer "k8s.io/kubernetes/pkg/kubelet/container"
 )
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev:
related to https://github.com/kubernetes/kubernetes/pull/93382/ ?

liggitt:
yes, looks like it is, though my change fixes the source rather than just the test

SergeyKanzhelev:
> yes, looks like it is, though my change fixes the source rather than just the test

What's the benefits of having those annotations sorted? Would runtime rely on them being sorted?

liggitt:
> What's the benefits of having those annotations sorted? Would runtime rely on them being sorted?

deterministic output is generally better than non-deterministic. there are code paths in the kubelet/cri interactions that compare running containers to expected containers, so it is not inconceivable a consumer of this method would expect stable output

liggitt:

/hold

Issue #93358, opened 2020-07-22
[Flaky test] "PersistentVolumeClaims not all in phase Bound within 5m0s" failing in ~2% of kind presubmits


liggitt:
**Which jobs are flaking**:

* pull-kubernetes-e2e-kind-ipv6
* pull-kubernetes-e2e-kind

**Which test(s) are flaking**:

* CSI mock volume ... tests

**Testgrid link**:

* https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-e2e-kind&include-filter-by-regex=CSI%20mock%20volume&width=5&sort-by-failures=
* https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-e2e-kind-ipv6&include-filter-by-regex=CSI%20mock%20volume&width=5&sort-by-failures=

**Reason for failure**:

```
test/e2e/storage/csi_mock_volume.go:351
Jul 22 00:33:25.985: Failed waiting for PVC to be bound PersistentVolumeClaims [pvc-7fnn6] not all in phase Bound within 5m0s
Unexpected error:
    <*errors.errorString | 0xc002072890>: {
        s: "PersistentVolumeClaims [pvc-7fnn6] not all in phase Bound within 5m0s",
    }
    PersistentVolumeClaims [pvc-7fnn6] not all in phase Bound within 5m0s
occurred
test/e2e/storage/csi_mock_volume.go:1200
```

**Anything else we need to know**:

* https://storage.googleapis.com/k8s-gubernator/triage/index.html?ci=0&pr=1&text=not%20all%20in%20phase%20Bound%20within%205m0s&test=CSI%20mock%20volume

Example failures:
* https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-kind/1286019551993008128
* https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-kind/1285857244977565699
* https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-kind/1285673283571683329

/sig storage

liggitt:
might be worth looking at the average bind time in successful run cases over time and seeing how close to the 5 minute timeout we are... what is the expected time to bind in these tests?

BenTheElder:
[...]

PR #93618, opened 2020-07-31
Fix flaky test: wait for mock CSI Driver bringup to perform e2e test

Jiawei0227:
**What type of PR is this?**
/kind failing-test

In our current mock CSI driver e2e test, we are not waiting for the CSI driver register successfully to perform test, e.g. provision PVC. This can lead to timeout error when the CSI driver pod takes longer to be up and running.

This change adds the waiting part so that the system will wait for up to 10 minutes for the driver to be ready. This
normally won't take this long. However, under a resource-constraint environment it can take longer time than expected.

#93358 

```release-note
NONE
```

```docs

```


Jiawei0227:
cc @msau42

Jiawei0227:
Another thought on this: the waitForCSIDriver actually makes sense before. Because it seems like we are using driver-registrar to generate the CSIDriver object in the past. So that the CSIDriver object will only be created after the CSIDriver pod is up and running. :)

test/e2e/storage/drivers/csi.go
```diff
@@ -608,17 +608,18 @@ func waitForCSIDriverRegistrationOnAllNodes(driverName string, cs clientset.Inte
...
 }
 
-func waitForCSIDriverRegistrationOnNode(nodeName string, driverName string, cs clientset.Interface) error {
-	const csiNodeRegisterTimeout = 5 * time.Minute
+// WaitForCSIDriverRegistrationOnNode waits for the CSINode object generated by the node-registrar on a certain node
+func WaitForCSIDriverRegistrationOnNode(nodeName string, driverName string, cs clientset.Interface, timeout time.Duration) error {
+	framework.Logf(fmt.Sprintf("waiting up to %v for CSIDriver %v to register on node %v", timeout, driverName, nodeName))
 
-	waitErr := wait.PollImmediate(10*time.Second, csiNodeRegisterTimeout, func() (bool, error) {
+	waitErr := wait.PollImmediate(10*time.Second, timeout, func() (bool, error) {
```
gnufied, line 622: not introduced this PR, but IMO we should use exponential backoff for checking CSINode registration. That might be less expensive on api-server and may cause these functions to return sooner in happy path. I am thinking something like - "initial: 2, factor: 1.4, count: 12"

Jiawei0227: Done

test/e2e/storage/drivers/csi.go
```diff
@@ -608,17 +608,18 @@ func waitForCSIDriverRegistrationOnAllNodes(driverName string, cs clientset.Inte
 		return err
 	}
 	for _, node := range nodes.Items {
-		if err := waitForCSIDriverRegistrationOnNode(node.Name, driverName, cs); err != nil {
+		if err := WaitForCSIDriverRegistrationOnNode(node.Name, driverName, cs, 5*time.Minute); err != nil {
 			return err
```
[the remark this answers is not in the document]
Jiawei0227: Because previously the wait time for waitForCSIDriverRegistrationOnNode is set to 5 minutes. And the reason we are using 10 minutes in other place is that our tests are flaky occasionally because the pod get scheduled onto the node longer than 5 minutes. I don't want to change the timeout threshold for other places. Or do you think we should keep them the same?

In the flaky tests I have triaged, it can took 6-8 minutes for the registration. That's why I set the timeout to 10min. This should be long enough to get the pod running to avoid such flakiness. 

I have addressed your comment to change the wait to be ExponentialBackoff. This should keep the timeout in sync with other places.

Jiawei0227 (commented):
Thanks for your review!

test/e2e/storage/csi_mock_volume.go
```diff
@@ -156,6 +156,11 @@ var _ = utils.SIGDescribe("CSI mock volume", func() {
 				destroyCSIDriver(cs, m.config.GetUniqueDriverName())
 			})
 		}
+
+		// Wait for the CSIDriver actually get deployed and CSINode object to be generated.
+		// This indicates the mock CSI driver pod is up and running healthy.
+		err = drivers.WaitForCSIDriverRegistrationOnNode(m.config.ClientNodeSelection.Name, m.config.GetUniqueDriverName(), cs)
+		framework.ExpectNoError(err, "Failed to register CSIDriver %v: %v", m.config.GetUniqueDriverName(), err)
```
gnufied, line 163: minor nit: I think `framework.ExpectNoError` will already log `err` and hence there is no need to supply `err` to string format again.

Jiawei0227: Done. I have seen multiple places where we log the err in the explanation part but I guess that should be fine.

BenTheElder:
i'm not sure about the timings but I will defer to storage there
this test fix should go in
/milestone v1.19

PR #93622, opened 2020-08-01
Fix race condition in pluginWatcher

knight42:

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature

/kind flake

Part of #93605
xref: https://github.com/kubernetes/kubernetes/pull/93605#issuecomment-667316127

I add some logging with the following patch:
<details>

```
diff --git pkg/kubelet/pluginmanager/pluginwatcher/example_plugin.go pkg/kubelet/pluginmanager/pluginwatcher/example_plugin.go
index d0859cc8967..a67a214556e 100644
--- pkg/kubelet/pluginmanager/pluginwatcher/example_plugin.go
+++ pkg/kubelet/pluginmanager/pluginwatcher/example_plugin.go
@@ -29,9 +29,10 @@ import (
 	"k8s.io/klog/v2"
 
 	registerapi "k8s.io/kubelet/pkg/apis/pluginregistration/v1"
+
 	"k8s.io/kubernetes/pkg/kubelet/pluginmanager/cache"
-	v1beta1 "k8s.io/kubernetes/pkg/kubelet/pluginmanager/pluginwatcher/example_plugin_apis/v1beta1"
-	v1beta2 "k8s.io/kubernetes/pkg/kubelet/pluginmanager/pluginwatcher/example_plugin_apis/v1beta2"
+	"k8s.io/kubernetes/pkg/kubelet/pluginmanager/pluginwatcher/example_plugin_apis/v1beta1"
+	"k8s.io/kubernetes/pkg/kubelet/pluginmanager/pluginwatcher/example_plugin_apis/v1beta2"
 )
 
 // examplePlugin is a sample plugin to work with plugin watcher
@@ -96,6 +97,7 @@ func GetPluginInfo(plugin *examplePlugin) cache.PluginInfo {
 
 // GetInfo is the RPC invoked by plugin watcher
 func (e *examplePlugin) GetInfo(ctx context.Context, req *registerapi.InfoRequest) (*registerapi.PluginInfo, error) {
+	klog.Errorf("GetInfo called")
 	return &registerapi.PluginInfo{
 		Type:              e.pluginType,
 		Name:              e.pluginName,
@@ -146,6 +148,7 @@ func (e *examplePlugin) Serve(services ...string) error {
 	go func() {
 		defer e.wg.Done()
 		// Blocking call to accept incoming connections.
+		klog.Infof("listening on socket")
 		if err := e.grpcServer.Serve(lis); err != nil {
 			klog.Errorf("example server stopped serving: %v", err)
 		}
diff --git pkg/kubelet/pluginmanager/pluginwatcher/plugin_watcher.go pkg/kubelet/pluginmanager/pluginwatcher/plugin_watcher.go
index 30177f67bdc..e432d399032 100644
--- pkg/kubelet/pluginmanager/pluginwatcher/plugin_watcher.go
+++ pkg/kubelet/pluginmanager/pluginwatcher/plugin_watcher.go
@@ -49,7 +49,7 @@ func NewWatcher(sockDir string, desiredStateOfWorld cache.DesiredStateOfWorld) *
 
 // Start watches for the creation and deletion of plugin sockets at the path
 func (w *Watcher) Start(stopCh <-chan struct{}) error {
-	klog.V(2).Infof("Plugin Watcher Start at %s", w.path)
+	klog.Infof("Plugin Watcher Start at %s", w.path)
 
 	// Creating the directory to be watched if it doesn't exist yet,
 	// and walks through the directory to discover the existing plugins.
@@ -67,11 +67,13 @@ func (w *Watcher) Start(stopCh <-chan struct{}) error {
 	if err := w.traversePluginDir(w.path); err != nil {
 		klog.Errorf("failed to traverse plugin socket path %q, err: %v", w.path, err)
 	}
+	klog.Infof("list pluginDir")
 
 	go func(fsWatcher *fsnotify.Watcher) {
 		for {
 			select {
 			case event := <-fsWatcher.Events:
+				klog.Infof("receive event: %s", event)
 				//TODO: Handle errors by taking corrective measures
 				if event.Op&fsnotify.Create == fsnotify.Create {
 					err := w.handleCreateEvent(event)
@@ -130,6 +132,7 @@ func (w *Watcher) traversePluginDir(dir string) error {
 				Name: path,
 				Op:   fsnotify.Create,
 			}
+			klog.Infof("add existing socket: %s", path)
 			//TODO: Handle errors by taking corrective measures
 			if err := w.handleCreateEvent(event); err != nil {
 				klog.Errorf("error %v when handling create event: %s", err, event)
```

</details>

and find out that if the test fails, the output would be something like
```
=== RUN   TestPluginRegistration
I0801 11:43:21.268993   74831 example_plugin.go:121] starting example server at: /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/plugin_manager_test025788546/plugin.sock
I0801 11:43:21.269004   74831 plugin_watcher.go:52] Plugin Watcher Start at /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/plugin_manager_test025788546
I0801 11:43:21.269916   74831 example_plugin.go:127] example server started at: /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/plugin_manager_test025788546/plugin.sock
I0801 11:43:21.270146   74831 example_plugin.go:151] listening on socket
I0801 11:43:21.270248   74831 plugin_watcher.go:70] list pluginDir
I0801 11:43:21.270288   74831 plugin_manager.go:114] Starting Kubelet Plugin Manager # <---- stuck here
    TestPluginRegistration: plugin_manager_test.go:107: Timed out waiting for plugin to be added to actual state of world cache.
I0801 11:44:21.783013   74831 plugin_manager.go:119] Shutting down Kubelet Plugin Manager
--- FAIL: TestPluginRegistration (60.51s)
FAIL
exit status 1
FAIL    k8s.io/kubernetes/pkg/kubelet/pluginmanager     60.851s
```

while if the test passes, the output would be sth like
```
=== RUN   TestPluginRegistration
I0801 11:43:18.278273   74796 example_plugin.go:121] starting example server at: /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/plugin_manager_test291123507/plugin.sock
I0801 11:43:18.278302   74796 plugin_watcher.go:52] Plugin Watcher Start at /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/plugin_manager_test291123507
I0801 11:43:18.279166   74796 example_plugin.go:127] example server started at: /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/plugin_manager_test291123507/plugin.sock
I0801 11:43:18.279389   74796 example_plugin.go:151] listening on socket
I0801 11:43:18.279658   74796 plugin_watcher.go:135] add existing socket: /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/plugin_manager_test291123507/plugin.sock
I0801 11:43:18.279708   74796 plugin_watcher.go:70] list pluginDir
I0801 11:43:18.279730   74796 plugin_manager.go:114] Starting Kubelet Plugin Manager
E0801 11:43:18.280703   74796 example_plugin.go:100] GetInfo called
E0801 11:43:18.280951   74796 example_plugin.go:110] Registration is: &RegistrationStatus{PluginRegistered:true,Error:,}
I0801 11:43:18.779723   74796 plugin_manager.go:119] Shutting down Kubelet Plugin Manager
--- PASS: TestPluginRegistration (0.50s)
```

Or

```
=== RUN   TestPluginRegistration
I0801 11:43:13.759376   74763 example_plugin.go:121] starting example server at: /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/plugin_manager_test526537534/plugin.sock
I0801 11:43:13.759538   74763 plugin_watcher.go:52] Plugin Watcher Start at /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/plugin_manager_test526537534
I0801 11:43:13.760092   74763 plugin_watcher.go:70] list pluginDir
I0801 11:43:13.760104   74763 plugin_manager.go:114] Starting Kubelet Plugin Manager
I0801 11:43:13.760137   74763 example_plugin.go:127] example server started at: /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/plugin_manager_test526537534/plugin.sock
I0801 11:43:13.760198   74763 example_plugin.go:151] listening on socket
I0801 11:43:13.760342   74763 plugin_watcher.go:76] receive event: "/var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/plugin_manager_test526537534/plugin.sock": CREATE
E0801 11:43:14.761989   74763 example_plugin.go:100] GetInfo called
E0801 11:43:14.762267   74763 example_plugin.go:110] Registration is: &RegistrationStatus{PluginRegistered:true,Error:,}
I0801 11:43:15.764041   74763 plugin_manager.go:119] Shutting down Kubelet Plugin Manager
--- PASS: TestPluginRegistration (2.01s)
```

EDIT:
See also https://github.com/kubernetes/kubernetes/pull/93622#issuecomment-669915496
I think the root cause is more likely to be concurrently creating the socket and adding the plugin dir to `w.fsWatcher` in the test. I guess the execution order might be:

1. The plugin socket has not been created, plugin watcher is traversing the plugin dir, just read dir names but not call `walkFn` yet https://github.com/golang/go/blob/6f08e89ec3280bf6577c2bdb01243cbeeb1a259d/src/path/filepath/path.go#L358-L364, so plugin watcher would not find the socket later
2. The plugin socket is created
3. `walkFn` is invoked and the plugin dir is added to the watcher, so the watcher would not receive any events later.

I decided to watch the socket dir before traversing it.
1. if the socket is created before we watching the dir, then we would find the socket while we are traversing the socket dir
2. if the socket is created after we watching the dir, then we would receive the CREATE event.

----
Please notice when the test fails, the plugin watcher could neither find any sockets in the socket dir nor receive "CREATE" event later. ~~I suspect `fsnotify` is not extremely reliable and might miss events in some cases.~~

~~As a result, I decide to make sure the example plugin server is started before plugin manager, so that the socket always exists when the plugin manager is started. I have run the test 100 times locally after this change, they all pass.~~

```release-note
kubelet: fix race condition in pluginWatcher 
```

```docs

```

### DIFF
--- a/pkg/kubelet/pluginmanager/plugin_manager_test.go
+++ b/pkg/kubelet/pluginmanager/plugin_manager_test.go
@@ -20,6 +20,9 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
+	"reflect"
+	"strconv"
 	"sync"
 	"testing"
 	"time"
@@ -29,6 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/record"
 	registerapi "k8s.io/kubelet/pkg/apis/pluginregistration/v1"
+
 	"k8s.io/kubernetes/pkg/kubelet/config"
 	"k8s.io/kubernetes/pkg/kubelet/pluginmanager/pluginwatcher"
 )
@@ -39,25 +43,19 @@ var (
 )
 
 type fakePluginHandler struct {
-	validatePluginCalled   bool
-	registerPluginCalled   bool
-	deregisterPluginCalled bool
+	events []string
 	sync.RWMutex
 }
 
 func newFakePluginHandler() *fakePluginHandler {
-	return &fakePluginHandler{
-		validatePluginCalled:   false,
-		registerPluginCalled:   false,
-		deregisterPluginCalled: false,
-	}
+	return &fakePluginHandler{}
 }
 
 // ValidatePlugin is a fake method
 func (f *fakePluginHandler) ValidatePlugin(pluginName string, endpoint string, versions []string) error {
 	f.Lock()
 	defer f.Unlock()
-	f.validatePluginCalled = true
+	f.events = append(f.events, "validate "+pluginName)
 	return nil
 }
 
@@ -65,7 +63,7 @@ func (f *fakePluginHandler) ValidatePlugin(pluginName string, endpoint string, v
 func (f *fakePluginHandler) RegisterPlugin(pluginName, endpoint string, versions []string) error {
 	f.Lock()
 	defer f.Unlock()
-	f.registerPluginCalled = true
+	f.events = append(f.events, "register "+pluginName)
 	return nil
 }
 
@@ -73,7 +71,13 @@ func (f *fakePluginHandler) RegisterPlugin(pluginName, endpoint string, versions
 func (f *fakePluginHandler) DeRegisterPlugin(pluginName string) {
 	f.Lock()
 	defer f.Unlock()
-	f.deregisterPluginCalled = true
+	f.events = append(f.events, "deregister "+pluginName)
+}
+
+func (f *fakePluginHandler) Reset() {
+	f.Lock()
+	defer f.Unlock()
+	f.events = nil
 }
 
 func init() {
@@ -90,15 +94,17 @@ func cleanup(t *testing.T) {
 	os.MkdirAll(socketDir, 0755)
 }
 
-func waitForRegistration(t *testing.T, fakePluginHandler *fakePluginHandler) {
+func waitForRegistration(t *testing.T, fakePluginHandler *fakePluginHandler, pluginName string) {
+	expected := []string{"validate " + pluginName, "register " + pluginName}
 	err := retryWithExponentialBackOff(
-		time.Duration(500*time.Millisecond),
+		100*time.Millisecond,
 		func() (bool, error) {
 			fakePluginHandler.Lock()
 			defer fakePluginHandler.Unlock()
-			if fakePluginHandler.validatePluginCalled && fakePluginHandler.registerPluginCalled {
+			if reflect.DeepEqual(fakePluginHandler.events, expected) {
 				return true, nil
 			}
+			t.Logf("expected %#v, got %#v, will retry", expected, fakePluginHandler.events)
 			return false, nil
 		},
 	)
@@ -134,19 +140,29 @@ func TestPluginRegistration(t *testing.T) {
 	fakeHandler := newFakePluginHandler()
 	pluginManager.AddHandler(registerapi.DevicePlugin, fakeHandler)
 
-	// Add a new plugin
-	socketPath := fmt.Sprintf("%s/plugin.sock", socketDir)
-	pluginName := "example-plugin"
-	p := pluginwatcher.NewTestExamplePlugin(pluginName, registerapi.DevicePlugin, socketPath, supportedVersions...)
-	require.NoError(t, p.Serve("v1beta1", "v1beta2"))
+	const maxDepth = 3
+	// Make sure the plugin manager is aware of the socket in subdirectories
+	for i := 0; i < maxDepth; i++ {
+		fakeHandler.Reset()
+		pluginDir := socketDir
 
-	// Verify that the plugin is registered
-	waitForRegistration(t, fakeHandler)
+		for j := 0; j < i; j++ {
+			pluginDir = filepath.Join(pluginDir, strconv.Itoa(j))
+		}
+		require.NoError(t, os.MkdirAll(pluginDir, os.ModePerm))
+		socketPath := filepath.Join(pluginDir, fmt.Sprintf("plugin-%d.sock", i))
+
+		// Add a new plugin
+		pluginName := fmt.Sprintf("example-plugin-%d", i)
+		p := pluginwatcher.NewTestExamplePlugin(pluginName, registerapi.DevicePlugin, socketPath, supportedVersions...)
+		require.NoError(t, p.Serve("v1beta1", "v1beta2"))
+
+		// Verify that the plugin is registered
+		waitForRegistration(t, fakeHandler, pluginName)
+	}
 }
 
-func newTestPluginManager(
-	sockDir string) PluginManager {
-
+func newTestPluginManager(sockDir string) PluginManager {
 	pm := NewPluginManager(
 		sockDir,
 		&record.FakeRecorder{},
--- a/pkg/kubelet/pluginmanager/pluginwatcher/plugin_watcher.go
+++ b/pkg/kubelet/pluginmanager/pluginwatcher/plugin_watcher.go
@@ -21,7 +21,6 @@ import (
 	"os"
 	"runtime"
 	"strings"
-	"time"
 
 	"github.com/fsnotify/fsnotify"
 	"k8s.io/klog/v2"
@@ -36,7 +35,6 @@ type Watcher struct {
 	path                string
 	fs                  utilfs.Filesystem
 	fsWatcher           *fsnotify.Watcher
-	stopped             chan struct{}
 	desiredStateOfWorld cache.DesiredStateOfWorld
 }
 
@@ -52,8 +50,6 @@ func NewWatcher(sockDir string, desiredStateOfWorld cache.DesiredStateOfWorld) *
 // Start watches for the creation and deletion of plugin sockets at the path
 func (w *Watcher) Start(stopCh <-chan struct{}) error {
 	klog.V(2).Infof("Plugin Watcher Start at %s", w.path)
-
-	w.stopped = make(chan struct{})
 
 	// Creating the directory to be watched if it doesn't exist yet,
 	// and walks through the directory to discover the existing plugins.
@@ -73,7 +69,6 @@ func (w *Watcher) Start(stopCh <-chan struct{}) error {
 	}
 
 	go func(fsWatcher *fsnotify.Watcher) {
-		defer close(w.stopped)
 		for {
 			select {
 			case event := <-fsWatcher.Events:
@@ -93,14 +88,6 @@ func (w *Watcher) Start(stopCh <-chan struct{}) error {
 				}
 				continue
 			case <-stopCh:
-				// In case of plugin watcher being stopped by plugin manager, stop
-				// probing the creation/deletion of plugin sockets.
-				// Also give all pending go routines a chance to complete
-				select {
-				case <-w.stopped:
-				case <-time.After(11 * time.Second):
-					klog.Errorf("timeout on stopping watcher")
-				}
 				w.fsWatcher.Close()
 				return
 			}
@@ -123,6 +110,12 @@ func (w *Watcher) init() error {
 // Walks through the plugin directory discover any existing plugin sockets.
 // Ignore all errors except root dir not being walkable
 func (w *Watcher) traversePluginDir(dir string) error {
+	// watch the new dir
+	err := w.fsWatcher.Add(dir)
+	if err != nil {
+		return fmt.Errorf("failed to watch %s, err: %v", w.path, err)
+	}
+	// traverse existing children in the dir
 	return w.fs.Walk(dir, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			if path == dir {
@@ -130,6 +123,11 @@ func (w *Watcher) traversePluginDir(dir string) error {
 			}
 
 			klog.Errorf("error accessing path: %s error: %v", path, err)
+			return nil
+		}
+
+		// do not call fsWatcher.Add twice on the root dir to avoid potential problems.
+		if path == dir {
 			return nil
 		}
 


pkg/kubelet/pluginmanager/pluginwatcher/plugin_watcher.go
```diff
@@ -93,14 +88,6 @@ func (w *Watcher) Start(stopCh <-chan struct{}) error {
 				}
 				continue
 			case <-stopCh:
-				// In case of plugin watcher being stopped by plugin manager, stop
-				// probing the creation/deletion of plugin sockets.
-				// Also give all pending go routines a chance to complete
-				select {
-				case <-w.stopped:
```
knight42, line 100: `w.stopped` would only be closed in line # 76, but it won't be closed as # 76 is defer-ed. 

I am not sure what the original purpose of `w.stopped` is, but it seems to be useless now.

liggitt: I agree the stopped channel use was weird, so the `<-w.stopped` case would never have fired. That means the 11 second timeout would always have been hit.

This PR is removing that timeout, meaning the fsWatcher is closing immediately with no timeout. Is that correct? Does that cut off FS events we wanted to see/handle?

knight42: I am not quite confident, but since the `stopCh` is closed, I think the plugin manager which starts the watcher expects the watcher to exit regardless of any further events.

pkg/kubelet/pluginmanager/plugin_manager_test.go
```diff
@@ -125,8 +126,12 @@ func TestPluginRegistration(t *testing.T) {
 	// Start the plugin manager
 	stopChan := make(chan struct{})
 	defer close(stopChan)
+
+	pluginServing := make(chan struct{})
+
 	go func() {
 		sourcesReady := config.NewSourcesReady(func(_ sets.String) bool { return true })
+		<-pluginServing
 		pluginManager.Run(sourcesReady, stopChan)
```
liggitt, line 135: how is this different than simply moving this goroutine after the p.Serve call?

pkg/kubelet/pluginmanager/pluginwatcher/plugin_watcher.go
liggitt: Yeah, I think you're right. Would like an ack from a sig-node reviewer

pkg/kubelet/pluginmanager/plugin_manager_test.go
knight42: That's a good point! Updated as you said.

liggitt:
/cc @msau42 
since you've reviewed changes in the plugin manager

liggitt:
```
//pkg/scheduler

TIMEOUT: //pkg/scheduler:go_default_test (Summary)
      /bazel-scratch/.cache/bazel/_bazel_root/7989b31489f31aee54f32688da2f0120/execroot/io_k8s_kubernetes/bazel-out/k8-fastbuild/testlogs/pkg/scheduler/go_default_test/run_8_of_10/test.log
```
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93605/pull-kubernetes-bazel-test/1290315521706692610

pkg/kubelet/pluginmanager/pluginwatcher/plugin_watcher.go
dashpole: yes, that is what we want.

dashpole:
/lgtm
/approve
Thanks for the fix!

pkg/kubelet/pluginmanager/plugin_manager_test.go
```diff
@@ -140,13 +137,18 @@ func TestPluginRegistration(t *testing.T) {
 	p := pluginwatcher.NewTestExamplePlugin(pluginName, registerapi.DevicePlugin, socketPath, supportedVersions...)
 	require.NoError(t, p.Serve("v1beta1", "v1beta2"))
 
+	// Ensure example plugin is started before plugin manager, so that the socket is present
```
msau42, line 140: /hold

If fsnotify is flaky and can miss events, then that seems like a product issue, where we need to design the plugin manager to be able to handle that?

Having the test start the plugin before the plugin manager seems like a workaround to get the test to pass, but masks potentially real problems. In a real system, the plugin is usually started after the plugin manager.

liggitt: I think what was racing here was not the filesystem and pluginmanager, but the observation of the plugin and the registration of the handler

in real life, we add the handlers that are sent the registration events before calling pluginManager.Run()

https://github.com/kubernetes/kubernetes/blob/ae9f76153f559c62bf1e7cf962d39168061dc6aa/pkg/kubelet/kubelet.go#L1301-L1307

would moving this above the goroutine where we call `pluginManager.Run(sourcesReady, stopChan)` be sufficient?
```
fakeHandler := newFakePluginHandler()
pluginManager.AddHandler(registerapi.DevicePlugin, fakeHandler)
```

liggitt: hmm, just saw in the PR description:

> Please notice when the test fails, the plugin watcher could neither find any sockets in the socket dir nor receive "CREATE" event later. I suspect fsnotify is not extremely reliable and might miss events in some cases.
> 
> As a result, I decide to make sure the example plugin server is started before plugin manager, so that the socket always exists when the plugin manager is started. I have run the test 100 times locally after this change, they all pass.

that's troubling, and I agree with @msau42 that if that is the issue we need a real fix, not to mask it in a test

knight42: I agree, let me try creating a minimal reproducible example first.

knight42:
@msau42 @liggitt I wrote a small program https://gist.github.com/knight42/ca294733bac7f11a75e9379fefcceeb1 to reproduce the problem.

On my macbook with 2.3 GHz 8 cores, I got
```
$ for i in {1..100}; do ./test -delay 50us||break; done
2020/08/06 19:57:52 Listen on: /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/foo-293534030/plugin.sock
2020/08/06 19:57:52 watcher: add dir: /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/foo-293534030
2020/08/06 19:57:52 receive event: "/var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/foo-293534030/plugin.sock": CREATE
2020/08/06 19:57:52 Listen on: /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/foo-313295467/plugin.sock
2020/08/06 19:57:52 watcher: add dir: /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/foo-313295467
2020/08/06 19:57:52 receive event: "/var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/foo-313295467/plugin.sock": CREATE
2020/08/06 19:57:52 Listen on: /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/foo-341828800/plugin.sock
2020/08/06 19:57:52 watcher: add dir: /var/folders/bp/20yxytt90055krc_0hct5qrm0000gn/T/foo-341828800
2020/08/06 19:58:02 no socket found
```

On a c5.xlarge AWS EC2 instance(3.00GHz 4 cores), I got
```
[ec2-user@ip-172-31-44-16 ~]$ for i in {1..100}; do ./test -delay 0.2us || break; done
2020/08/06 12:02:40 Listen on: /tmp/foo-893352217/plugin.sock
2020/08/06 12:02:40 watcher: add dir: /tmp/foo-893352217
2020/08/06 12:02:40 add existing socket: /tmp/foo-893352217/plugin.sock
2020/08/06 12:02:40 receive event: "/tmp/foo-893352217/plugin.sock": REMOVE
2020/08/06 12:02:40 Listen on: /tmp/foo-116042815/plugin.sock
2020/08/06 12:02:40 watcher: add dir: /tmp/foo-116042815
2020/08/06 12:02:40 add existing socket: /tmp/foo-116042815/plugin.sock
2020/08/06 12:02:40 Listen on: /tmp/foo-841397517/plugin.sock
2020/08/06 12:02:40 watcher: add dir: /tmp/foo-841397517
2020/08/06 12:02:40 add existing socket: /tmp/foo-841397517/plugin.sock
2020/08/06 12:02:40 Listen on: /tmp/foo-850143221/plugin.sock
2020/08/06 12:02:40 watcher: add dir: /tmp/foo-850143221
2020/08/06 12:02:50 no socket found
```

I think the root cause is more likely to be concurrently creating the socket and adding the socket dir to `w.fsWatcher` in the test. I guess the execution order might be:
1. The plugin socket has not been created, plugin watcher is traversing the socket dir, just read dir names but not call `walkFn` yet https://github.com/golang/go/blob/6f08e89ec3280bf6577c2bdb01243cbeeb1a259d/src/path/filepath/path.go#L358-L364, so plugin watcher would not find the socket later
2. The plugin socket is created
3. `walkFn` is invoked and the socket dir is added to the watcher, so the watcher would not receive any events later.

I think such extreme case rarely happens in real world and it might be okay to make the execution order deterministic in the test.

liggitt:
> I think such extreme case rarely happens in real world and it might be okay to make the execution order deterministic in the test.

I disagree. The plugin watcher should handle any potential order of events robustly during startup.

knight42:
@liggitt I really appreciate your strictness 😄 . 

It occurs to me that we could watch the socket dir before traversing it, if the socket is created before we watching the dir then we would find the socket while we are traversing the socket dir, if the socket is created after we watching the dir then we would receive the CREATE event.

pkg/kubelet/pluginmanager/pluginwatcher/plugin_watcher.go
```diff
@@ -134,10 +125,6 @@ func (w *Watcher) traversePluginDir(dir string) error {
 		}
 
 		switch mode := info.Mode(); {
-		case mode.IsDir():
```
liggitt, line 137: dropping this means we won't add subdirectories while walking the tree, right?

knight42: yes, do we need to support plugin sockets in the subdirectories?

liggitt: I'll defer to the storage/node folks, but I believe we do, yes

knight42: I would like to hear from storage/node folks, but at least according to https://github.com/kubernetes/community/blob/master/contributors/design-proposals/resource-management/device-plugin.md#unix-socket, I am under the impression that device plugin server should only create one socket at the socket dir.

liggitt: this watcher is the single watcher instance for all plugins, right? so the path being added in Start is the root plugins folder, which then contains subfolders for individual plugins?

knight42: That might be a use case and I am not sure whether there are uses depending on this feature, but at least regarding to the device plugins I have seen, such as nvidia device plugin and kubevirt, they all create the socket at the root plugins dir:
* nvidia: https://github.com/NVIDIA/k8s-device-plugin/blob/055e87d0959e1e905c7332c80bb049692e8c11ae/mig-strategy.go#L90
* kubevirt: https://github.com/kubevirt/device-plugin-manager/blob/c9ca4e84bfedb232bf8c9956f6b8594f42c5cee3/pkg/dpm/plugin.go#L54

msau42: All csi drivers I know of are creating the socket in the top directory, but I don't think this was was explicitly set in stone. If we want to change this behavior, we probably need to put out a deprecation notice and wait for the deprecation period.

cc @saad-ali

msau42: If we want to retain the current behavior, is it possible to have the reconciler also periodically traverse the registration directory, instead of only operating on fsnotify events?

saad-ali: > All csi drivers I know of are creating the socket in the top directory, but I don't think this was was explicitly set in stone. If we want to change this behavior, we probably need to put out a deprecation notice and wait for the deprecation period.
>
> cc @saad-ali

Correct, the assumption was you can create a socket anywhere under the plugin directory (could be under a sub-directory). We should try to retain that behavior.

knight42: I came up with a question if we really want to support plugin socket in subdirectories.

Let's say `fsWatcher` is already watching `/var/lib/kubelet/device-plugins/`, and then user creates a socket at `/var/lib/kubelet/device-plugins/foo/plugin.sock`. Now please note that `/var/lib/kubelet/device-plugins/foo/` is unknown to `fsWatcher`, so kubelet would not know the `/var/lib/kubelet/device-plugins/foo/plugin.sock` is created.

What should we do to deal with this senario?
1. periodically traverse the registration directory?
2. automatically watch subdirectories when they are created(which may race, i.e. if the socket is created before we watch the subdirectory we would miss the CREATE event)

liggitt: > Now please note that /var/lib/kubelet/device-plugins/foo/ is unknown to fsWatcher

why is it unknown? if `foo` is present at instantiation time, the code in master adds a watch on the `foo` dir in `traversePluginDir`, and if `foo` is created after instantiation time, the watcher traverses it and starts watching in `handleCreateEvent`

> What should we do to deal with this senario?
> ...
> * automatically watch subdirectories when they are created

today, we automatically watch subdirectories when they are created

knight42: Ahh indeed! I have overlooked the related code, sorry about that.

pkg/kubelet/pluginmanager/pluginwatcher/plugin_watcher.go
```diff
@@ -65,6 +61,10 @@ func (w *Watcher) Start(stopCh <-chan struct{}) error {
 	if err != nil {
 		return fmt.Errorf("failed to start plugin fsWatcher, err: %v", err)
 	}
+	err = fsWatcher.Add(w.path)
+	if err != nil {
+		return fmt.Errorf("failed to watch %s, err: %v", w.path, err)
+	}
```
liggitt, line 67: it seems like the fundamental issue is a race inside traversePluginDir in the following scenario:

1. traversePluginDir("/foo") is called
2. `w.fs.Walk` enumerates children of `/foo`
3. a new child `/foo/bar` is created
4. the walker starts watching `/foo`

It seems like traversePluginDir should hoist watching dir before doing `w.fs.Walk`:

```
func (w *Watcher) traversePluginDir(dir string) error {
	// watch the new dir
	if err := w.fsWatcher.Add(path); err != nil {
		return fmt.Errorf("failed to watch %s, err: %v", path, err)
	}
	// traverse existing children in the dir
	return w.fs.Walk(dir, func(path string, info os.FileInfo, err error) error {
		...
	})
```

liggitt: if we fix traversePluginDir, I think we can drop this special-case call to `fsWatcher.Add(w.path)`

knight42: Great! I think this is much better!

pkg/kubelet/pluginmanager/pluginwatcher/plugin_watcher.go
```diff
@@ -123,6 +110,12 @@ func (w *Watcher) init() error {
 // Walks through the plugin directory discover any existing plugin sockets.
 // Ignore all errors except root dir not being walkable
 func (w *Watcher) traversePluginDir(dir string) error {
+	// watch the new dir
+	err := w.fsWatcher.Add(w.path)
```
liggitt, line 114: this should be:
```
w.fsWatcher.Add(dir)
```

also, is calling `w.fsWatcher.Add` twice on the same path problematic? do we need to avoid calling `w.fsWatcher.Add(path)` when `path == dir` inside the Walk callback?

knight42: > w.fsWatcher.Add(dir)

Nice catch... I shoud be more careful.

> is calling w.fsWatcher.Add twice on the same path problematic?

I have tested locally and it would not return an error, but according to the source code it would re-register the fd https://github.com/fsnotify/fsnotify/blob/7f4cf4dd2b522a984eaca51d1ccee54101d3414a/kqueue.go#L227, so I think we'd better avoid calling `fsWatcher.Add` twice on the same path.

liggitt: >> w.fsWatcher.Add(dir)
>
> Nice catch... I shoud be more careful.

>> is calling w.fsWatcher.Add twice on the same path problematic?
>
> I have tested locally and it would not return an error, but according to the source code it would re-register the fd

can we make sure there are tests that exercise these paths

knight42: It is already midnight here, I would fix that tomorrow :)

pkg/kubelet/pluginmanager/pluginwatcher/plugin_watcher.go
```diff
@@ -134,7 +127,8 @@ func (w *Watcher) traversePluginDir(dir string) error {
 		}
 
 		switch mode := info.Mode(); {
-		case mode.IsDir():
+		// do not call fsWatcher.Add twice on the root dir to avoid potential problems.
+		case path != dir && mode.IsDir():
```
liggitt, line 131: keep the case as-is so we don't fall through to the default case:

```suggestion
		case mode.IsDir():
			if path != dir {
				if err := w.fsWatcher.Add(path); err != nil {
					return fmt.Errorf("failed to watch %s, err: %v", path, err)
				}
			}
```

pkg/kubelet/pluginmanager/plugin_manager_test.go
```diff
@@ -134,19 +137,28 @@ func TestPluginRegistration(t *testing.T) {
...
+	// Make sure the plugin manager is aware of the socket in subdirectories
+	for i := 0; i < maxDepth; i++ {
+		pluginDir := socketDir
+		for j := 0; j < i; j++ {
+			pluginDir = filepath.Join(pluginDir, strconv.Itoa(j))
+		}
+		_ = os.MkdirAll(pluginDir, os.ModePerm)
+		socketPath := filepath.Join(pluginDir, "plugin.sock")
+		// Add a new plugin
+		p := pluginwatcher.NewTestExamplePlugin(pluginName, registerapi.DevicePlugin, socketPath, supportedVersions...)
```
liggitt, line 153: I don't know how the plugin system works, but do we need a unique name for each depth (e.g. `fmt.Sprintf("%s%d" pluginName, i)`)?

pkg/kubelet/pluginmanager/plugin_manager_test.go
```diff
@@ -134,19 +137,28 @@ func TestPluginRegistration(t *testing.T) {
...
+		require.NoError(t, p.Serve("v1beta1", "v1beta2"))
+
+		// Verify that the plugin is registered
+		waitForRegistration(t, fakeHandler)
```
liggitt, line 157: waitForRegistration only verifies a single plugin was registered, so after the first one at depth 0, it is not checking anything further... we need to pass in the particular path we're expecting to be registered, right?

liggitt:
the change in `traversePluginDir` looks good to me, though I'd like @msau42 or @jsafrane to look at it closely as well

just a couple comments on the test, and I wanted to check if you ran the whole pluginmanager package with the `stress` utility as described in https://github.com/kubernetes/kubernetes/issues/93902 to make sure this resolves all races and doesn't introduce new ones

pkg/kubelet/pluginmanager/plugin_manager_test.go
knight42: In this test, just the socket path needs to be unique, but it might be better to have unique plugin names.

pkg/kubelet/pluginmanager/plugin_manager_test.go
knight42: we don't need to pass the particular path, we just need to reset the handler.

liggitt: > we don't need to pass the particular path, we just need to reset the handler.

it would be better to check the specific path, otherwise a late/duplicate event from an early plugin could make it look like a later one passed

knight42: Actually a late event would not affect the test, as the test would fail immediately if registration timeout, and I think duplicate event rarely occurs on a local network. But I think this check is good to have.

knight42:
```
$ stress ./pluginmanager.test -test.run TestPluginRegistration
0 runs so far, 0 failures
19 runs so far, 0 failures
34 runs so far, 0 failures
50 runs so far, 0 failures
63 runs so far, 0 failures
81 runs so far, 0 failures
95 runs so far, 0 failures
109 runs so far, 0 failures
125 runs so far, 0 failures
141 runs so far, 0 failures
154 runs so far, 0 failures
169 runs so far, 0 failures
186 runs so far, 0 failures
198 runs so far, 0 failures
211 runs so far, 0 failures
224 runs so far, 0 failures
243 runs so far, 0 failures
256 runs so far, 0 failures
270 runs so far, 0 failures
284 runs so far, 0 failures
300 runs so far, 0 failures
316 runs so far, 0 failures
330 runs so far, 0 failures
346 runs so far, 0 failures
362 runs so far, 0 failures
378 runs so far, 0 failures
390 runs so far, 0 failures
406 runs so far, 0 failures
421 runs so far, 0 failures
437 runs so far, 0 failures
451 runs so far, 0 failures
468 runs so far, 0 failures
481 runs so far, 0 failures
495 runs so far, 0 failures
511 runs so far, 0 failures
527 runs so far, 0 failures
540 runs so far, 0 failures
553 runs so far, 0 failures
572 runs so far, 0 failures
587 runs so far, 0 failures
602 runs so far, 0 failures
616 runs so far, 0 failures
633 runs so far, 0 failures
648 runs so far, 0 failures
662 runs so far, 0 failures
673 runs so far, 0 failures
```

liggitt:
now that our fake handler is handling multiple plugins, I think we should record the actual order and plugin names we encounter, like this:

```diff
diff --git a/pkg/kubelet/pluginmanager/plugin_manager_test.go b/pkg/kubelet/pluginmanager/plugin_manager_test.go
index d81f0d42e83..14dd5387e12 100644
--- a/pkg/kubelet/pluginmanager/plugin_manager_test.go
+++ b/pkg/kubelet/pluginmanager/plugin_manager_test.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strconv"
 	"sync"
 	"testing"
@@ -42,26 +43,19 @@ var (
 )
 
 type fakePluginHandler struct {
-	validatePluginCalled   bool
-	registerPluginCalled   bool
-	deregisterPluginCalled bool
-	registerEndpoint       string
+	events []string
 	sync.RWMutex
 }
 
 func newFakePluginHandler() *fakePluginHandler {
-	return &fakePluginHandler{
-		validatePluginCalled:   false,
-		registerPluginCalled:   false,
-		deregisterPluginCalled: false,
-	}
+	return &fakePluginHandler{}
 }
 
 // ValidatePlugin is a fake method
 func (f *fakePluginHandler) ValidatePlugin(pluginName string, endpoint string, versions []string) error {
 	f.Lock()
 	defer f.Unlock()
-	f.validatePluginCalled = true
+	f.events = append(f.events, fmt.Sprintf("validate %s", pluginName))
 	return nil
 }
 
@@ -69,8 +63,7 @@ func (f *fakePluginHandler) ValidatePlugin(pluginName string, endpoint string, v
 func (f *fakePluginHandler) RegisterPlugin(pluginName, endpoint string, versions []string) error {
 	f.Lock()
 	defer f.Unlock()
-	f.registerPluginCalled = true
-	f.registerEndpoint = endpoint
+	f.events = append(f.events, fmt.Sprintf("register %s", pluginName))
 	return nil
 }
 
@@ -78,11 +71,13 @@ func (f *fakePluginHandler) RegisterPlugin(pluginName, endpoint string, versions
 func (f *fakePluginHandler) DeRegisterPlugin(pluginName string) {
 	f.Lock()
 	defer f.Unlock()
-	f.deregisterPluginCalled = true
+	f.events = append(f.events, fmt.Sprintf("deregister %s", pluginName))
 }
 
 func (f *fakePluginHandler) Reset() {
-	*f = fakePluginHandler{}
+	f.Lock()
+	defer f.Unlock()
+	f.events = nil
 }
 
 func init() {
@@ -99,15 +94,18 @@ func cleanup(t *testing.T) {
 	os.MkdirAll(socketDir, 0755)
 }
 
-func waitForRegistration(t *testing.T, fakePluginHandler *fakePluginHandler, expectedEndpoint string) {
+func waitForRegistration(t *testing.T, fakePluginHandler *fakePluginHandler, pluginName string) {
 	err := retryWithExponentialBackOff(
-		500*time.Millisecond,
+		100*time.Millisecond,
 		func() (bool, error) {
 			fakePluginHandler.Lock()
 			defer fakePluginHandler.Unlock()
-			return fakePluginHandler.validatePluginCalled &&
-				fakePluginHandler.registerPluginCalled &&
-				fakePluginHandler.registerEndpoint == expectedEndpoint, nil
+			expected := []string{"validate " + pluginName, "register " + pluginName}
+			if reflect.DeepEqual(fakePluginHandler.events, expected) {
+				return true, nil
+			}
+			t.Logf("expected %#v, got %#v, will retry", expected, fakePluginHandler.events)
+			return false, nil
 		},
 	)
 	if err != nil {
@@ -160,7 +158,7 @@ func TestPluginRegistration(t *testing.T) {
 		require.NoError(t, p.Serve("v1beta1", "v1beta2"))
 
 		// Verify that the plugin is registered
-		waitForRegistration(t, fakeHandler, socketPath)
+		waitForRegistration(t, fakeHandler, pluginName)
 	}
 }
 ```

(I also shortened the initial backoff time to 100ms which shortened the test from ~8 seconds to ~4)

liggitt:
/hold
until updates are complete

knight42:
@liggitt Please take another look

liggitt:
/lgtm

would like @msau42 to take a look as well before removing the hold

msau42:
/assign @saad-ali 
cc @chrishenzie

saad-ali (approved):
/lgtm
/approve
/hold cancel

PR #93629, opened 2020-08-01
fix flaky TestVolumeBinding unit test

### LINKED ISSUE #93009: The unit test TestVolumeBinding is flaky
**Which jobs are flaking**:

The unit test `TestVolumeBinding`

```
--- FAIL: TestVolumeBinding (0.71s)
    --- FAIL: TestVolumeBinding/bound_and_unbound_unsatisfied (0.10s)
        volume_binding_test.go:282: call PreFilter and check status
        volume_binding_test.go:291: check state after prefilter phase
        volume_binding_test.go:299: call Filter and check status
        volume_binding_test.go:302: filter status does not match: &{1 [could not find v1.PersistentVolume "pv-a"]}, want: &{3 [node(s) had volume node affinity conflict node(s) didn't find available persistent volumes to bind]}
FAIL
```

**Testgrid link**:

**Reason for failure**:

In volume scheduling, we maintain internal cache which listens on PV/PVC events. However, we don't have a method to wait for them to be synced in UT. 

In production, this is not a problem. Even if the scheduler marks the pod uncacheable due to the stale cache in PVAssumeCache/PVCAssumeCache, the pod will be moved to the active queue later as the scheduler [will move unschedulable pods to the active queue if they are not retried for more than 1 minute](https://github.com/kubernetes/kubernetes/pull/72558).

related:

- https://github.com/kubernetes/kubernetes/pull/73305
- https://github.com/kubernetes/kubernetes/pull/72558

cofyc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
/kind flake

fixes #93009 

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/scheduler/framework/plugins/volumebinding/volume_binding_test.go
+++ b/pkg/scheduler/framework/plugins/volumebinding/volume_binding_test.go
@@ -252,10 +252,7 @@ func TestVolumeBinding(t *testing.T) {
 				t.Fatal(err)
 			}
 
-			// Start informer factory after initialization
-			informerFactory.Start(ctx.Done())
-
-			// Feed testing data and wait for them to be synced
+			t.Log("Feed testing data and wait for them to be synced")
 			client.StorageV1().StorageClasses().Create(ctx, immediateSC, metav1.CreateOptions{})
 			client.StorageV1().StorageClasses().Create(ctx, waitSC, metav1.CreateOptions{})
 			if item.node != nil {
@@ -267,19 +264,21 @@ func TestVolumeBinding(t *testing.T) {
 			for _, pv := range item.pvs {
 				client.CoreV1().PersistentVolumes().Create(ctx, pv, metav1.CreateOptions{})
 			}
-			caches := informerFactory.WaitForCacheSync(ctx.Done())
-			for _, synced := range caches {
-				if !synced {
-					t.Errorf("error waiting for informer cache sync")
-				}
-			}
 
-			// Verify
+			t.Log("Start informer factory after initialization")
+			informerFactory.Start(ctx.Done())
+
+			t.Log("Wait for all started informers' cache were synced")
+			informerFactory.WaitForCacheSync(ctx.Done())
+
+			t.Log("Verify")
+
 			p := pl.(*VolumeBinding)
 			nodeInfo := framework.NewNodeInfo()
 			nodeInfo.SetNode(item.node)
 			state := framework.NewCycleState()
-			t.Logf("call PreFilter and check status")
+
+			t.Logf("Verify: call PreFilter and check status")
 			gotPreFilterStatus := p.PreFilter(ctx, state, item.pod)
 			if !reflect.DeepEqual(gotPreFilterStatus, item.wantPreFilterStatus) {
 				t.Errorf("filter prefilter status does not match: %v, want: %v", gotPreFilterStatus, item.wantPreFilterStatus)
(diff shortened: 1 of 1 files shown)


pkg/scheduler/framework/plugins/volumebinding/volume_binding_test.go
```diff
@@ -267,19 +264,25 @@ func TestVolumeBinding(t *testing.T) {
 			for _, pv := range item.pvs {
 				client.CoreV1().PersistentVolumes().Create(ctx, pv, metav1.CreateOptions{})
 			}
+
+			t.Log("Start informer factory after initialization")
+			informerFactory.Start(ctx.Done())
```
cofyc, line 269: see explanation here: https://github.com/kubernetes/kubernetes/issues/93009#issuecomment-667527689

In my understanding, with a real client and the real API Server events should not be lost.

Huang-Wei: > However, with the fake client, events may be lost in testing.

If you need to act upon the API change events on a UT, I think you have to try prepend/add reactor for a fake client. You can search:

- client.PrependReactor()
- client.PrependWatchReactor()

However, if the test just relies on the a fixed state of all initialized API objects (i think it's the case, correct me if I'm wrong), the current change (start informerFactory and wait for its cache to be synced) should be good enough.

pkg/scheduler/framework/plugins/volumebinding/volume_binding_test.go
```diff
@@ -267,19 +264,25 @@ func TestVolumeBinding(t *testing.T) {
...
+
 			caches := informerFactory.WaitForCacheSync(ctx.Done())
 			for _, synced := range caches {
 				if !synced {
 					t.Errorf("error waiting for informer cache sync")
 				}
 			}
```
Huang-Wei, line 276: Just for curiosity, is this logic necessary? IMO a common pattern is just:

```go
informerFactory.Start(ctx.Done())
informerFactory.WaitForCacheSync(ctx.Done())
```

pkg/scheduler/framework/plugins/volumebinding/volume_binding_test.go
cofyc: > However, if the test just relies on the a fixed state of all initialized API objects
yes, it is

pkg/scheduler/framework/plugins/volumebinding/volume_binding_test.go
cofyc: practically, not necessary. In the current implementation, this function returns with the cache not synced only if the `ctx.Done()` is closed early, which will not happen in our testing.

However, I think it's good practice to check all returned results and fail if something is not expected. This may seem redundant sometimes, but it does not assume anything. For example, controllers in kube-controller-manager may exit if the informers they care can't be synced;

https://github.com/kubernetes/kubernetes/blob/2b9837fdcda0163712a439b71211e34a3f27bd34/pkg/controller/volume/persistentvolume/pv_controller_base.go#L306-L308

cofyc: I removed this logic since it's not necessary in our testing. The code is more concise now.

Huang-Wei:
/lgtm
@ahg-g do you want to tag it to 1.19?

PR #93631, opened 2020-08-01
modify proto for ipv6 route filter

yanghaichao12:
What type of PR is this?
/kind bug

because the protocol type in  local table for ipv6 is none but not kernal，which is present in local table for ipv4.:
```json
local ::1 dev lo  proto none  metric 0  pref medium
local 100::1 dev lo  proto none  metric 0  pref medium
local 300::3 dev lo  proto none  metric 0  pref medium

local 10.46.177.110 dev ens3  proto kernel  scope host  src 10.46.177.110
broadcast 10.46.177.255 dev ens3  proto kernel  scope link  src 10.46.177.110
local 10.254.0.1 dev kube-ipvs0  proto kernel  scope host  src 10.254.0.1

```
 the case in function 'RouteListFiltered'  in the file of  vendor/github.com/vishvananda/netlink/route_linux.go could not work well
```go
case filterMask&RT_FILTER_PROTOCOL != 0 && route.Protocol != filter.Protocol:
				continue
```
 because of the inconsistencies between IPv4 and IPv6kernel.
For example, the 'NodePort'   is affected and cannot open the local port

Which issue(s) this PR fixes:
Fixes #
area/ipvs
/release-note-none
/priority backlog

### DIFF
--- a/pkg/proxy/ipvs/netlink_linux.go
+++ b/pkg/proxy/ipvs/netlink_linux.go
@@ -177,6 +177,16 @@ func (h *netlinkHandle) GetLocalAddresses(dev, filterDev string) (sets.String, e
 	if err != nil {
 		return nil, fmt.Errorf("error list route table, err: %v", err)
 	}
+
+	if h.isIPv6 {
+		unsRouteFilter := &netlink.Route{
+			Table:    unix.RT_TABLE_LOCAL,
+			Type:     unix.RTN_LOCAL,
+			Protocol: unix.RTPROT_UNSPEC,
+		}
+		unsroutes, _ := h.RouteListFiltered(netlink.FAMILY_V6, unsRouteFilter, filterMask)
+		routes = append(routes, unsroutes...)
+	}
 	res := sets.NewString()
 	for _, route := range routes {
 		if route.LinkIndex == filterLinkIndex {


aojea:
I see proto kernel for ipv6 routes, is this not ok?
```
ip -6 route
::1 dev lo proto kernel metric 256 pref medium
fc00:f853:ccd:e793::/64 dev br-fea52c6e7c01 proto kernel metric 256 linkdown pref medium
fe80::/64 dev br-fea52c6e7c01 proto kernel metric 256 linkdown pref medium
fe80::/64 dev wlp82s0 proto kernel metric 600 pref medium
```

yanghaichao12:
@aojea hi, test this : ' ip -6 r show table local'

aojea:
> @aojea hi, test this : ' ip -6 r show table local'

$ ip -6 r show table local
local ::1 dev lo proto kernel metric 0 pref medium
anycast fc00:f853:ccd:e793:: dev br-68d675e8e360 proto kernel metric 0 pref medium
local fc00:f853:ccd:e793::1 dev br-68d675e8e360 proto kernel metric 0 pref medium
local fd00:aaaa:bbbb:cccc::1 dev lo proto kernel metric 0 pref medium
anycast fe80:: dev enp2s0 proto kernel metric 0 pref medium
anycast fe80:: dev br-68d675e8e360 proto kernel metric 0 pref medium
anycast fe80:: dev docker0 proto kernel metric 0 pref medium
anycast fe80:: dev veth1409fee proto kernel metric 0 pref medium
anycast fe80:: dev vethbcf67ef proto kernel metric 0 pref medium
local fe80::1 dev br-68d675e8e360 proto kernel metric 0 pref medium
local fe80::42:fff:fec7:b453 dev docker0 proto kernel metric 0 pref medium
local fe80::42:67ff:fe49:4682 dev br-68d675e8e360 proto kernel metric 0 pref medium
local fe80::4261:86ff:fef5:96dd dev enp2s0 proto kernel metric 0 pref medium
local fe80::60b1:e7ff:fe5b:da1a dev vethbcf67ef proto kernel metric 0 pref medium
local fe80::7c7b:32ff:fe18:5395 dev veth1409fee proto kernel metric 0 pref medium
ff00::/8 dev enp2s0 metric 256 pref medium
ff00::/8 dev br-68d675e8e360 metric 256 linkdown pref medium
ff00::/8 dev docker0 metric 256 pref medium
ff00::/8 dev veth1409fee metric 256 pref medium
ff00::/8 dev vethbcf67ef metric 256 pref medium

yanghaichao12:
@aojea it‘s wiered. i check other machine and get same results:
image

maybe original proto is absent when using  'ip a add xxxx' (see: https://github.com/torvalds/linux/blob/06a81c1c7db9bd5de0bd38cd5acc44bb22b99150/net/core/fib_rules.c#L560).
 i'm not sure  and i'll check the source code of iproute2 later.  thx

yanghaichao12:
@aojea hi， how did you add those route? by 'ip a add' or 'ip r add ...proto kernel' or just 'route add...'? i add those by  'ip a add' and the routes are generated automatically. my systems are :
```bash
[root@ZSJ_redhat7 ~]# uname -a
Linux ZSJ_redhat7.3_1018 3.10.0-514.el7.x86_64 #1 SMP Wed Oct 19 11:24:13 EDT 2016 x86_64 x86_64 x86_64 GNU/Linux
```
or
```bash
root@k8s:/etc/iproute2# uname -a
Linux k8s 4.4.0-21-generic #37-Ubuntu SMP Mon Apr 18 18:33:37 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```

by tracing 'ip a add' , i got this:

```bash
sendmsg(3, {msg_name(12)={sa_family=AF_NETLINK, pid=0, groups=00000000}, msg_iov(1)=[{"\x40\x00\x00\x00\x14\x00\x05\x06\x90\xb5\x30\x5f\x00\x00\x00\x00\x0a\x80\x00\x00\x02\x00\x00\x00\x14\x00\x02\x00\x03\x00\x00\x00"..., 64}], msg_controllen=0, msg_flags=0}, 0) = 64
```
so i checked the function ’inet6_rtm_newaddr‘ in https://github.com/torvalds/linux/blob/fc80c51fd4b23ec007e88d4c688f2cac1b8648e7/net/ipv6/addrconf.c#L4805
and found 'tb[FRA_PROTOCOL]'  not being  intitial, maybe that's the reason. i'm not sure. 
i don‘t know how to fix it, maybe the better way is using both 'RTPROT_UNSPEC' and 'RTPROT_KERNEL' as filter.....

aojea:
> how did you add those route? by 'ip a add' or 'ip r add ...proto kernel' or just 'route add...'? i add those by 'ip a add' and the routes are generated automatically. my systems are :

Those routes are automatically generated once you configure an IP on an interface.

yanghaichao12:
@aojea how about that，using two filter：
```go
routeFilter := &netlink.Route{
		Table:    unix.RT_TABLE_LOCAL,
		Type:     unix.RTN_LOCAL,
		Protocol: unix.RTPROT_KERNEL,
	}
	filterMask := netlink.RT_FILTER_TABLE | netlink.RT_FILTER_TYPE | netlink.RT_FILTER_PROTOCOL

	// find chosen device
	if chosenLinkIndex != -1 {
		routeFilter.LinkIndex = chosenLinkIndex
		filterMask |= netlink.RT_FILTER_OIF
	}

	routes, err := h.RouteListFiltered(netlink.FAMILY_ALL, routeFilter, filterMask)
	if err != nil {
		return nil, fmt.Errorf("error list route table, err: %v", err)
	}

	if h.isIPv6 {
		unsRouteFilter := &netlink.Route{
			Table:    unix.RT_TABLE_LOCAL,
			Type:     unix.RTN_LOCAL,
			Protocol: unix.RTPROT_UNSPEC,
		}
		unsroutes, _ := h.RouteListFiltered(netlink.FAMILY_ALL, unsRouteFilter, filterMask)
		routes = append(routes, unsroutes...)
	}
```

aojea:
@yanghaichao12 , I know that there are some people are already using for IPv6 and dual stack and we didn't receive any bug related to this, is this a bug that you hit or is just your conclusion after reading the code?

However, let me ping some people with much more knowledge than me in this area
/assign @uablrek  @andrewsykim

uablrek:
I have never seen anything like;

> local ::1 dev lo  proto none  metric 0  pref medium

in any system. It is always `proto kernel` for all routes automatically created when doing a `route -6 address add ...`;

```
local ::1 dev lo proto kernel metric 0 pref medium
```

I picked `lo` as example since K8s never has anything to do with that address setting.

The PR as it is would create failures in _any_ system _except_ the issuers. The proposed code in above is better.

I do not think it is a good idea to alter the K8s code for this, rather try to figure out why `proto none` in showing up only on the issuers systems. Some sysctl setting maybe?

And I also think that an issue must be written if it doesn't exist(?) that is triaged and analyzed in the normal way.

yanghaichao12:
@aojea @uablrek  not just after reading the code. in fact ，i've to resolve the problem by using ’--nodeport-addresses‘.  but my NICs are so many, i don't want to add every address here, so
i debug the codes and review some  literature. At first, I suspected ’sysctl' but cannot find anything. if it's existing, i think that's best way. if you know, pls help me. i've looked for a long time and found nothing....thx

aojea:
> And I also think that an issue must be written if it doesn't exist(?) that is triaged and analyzed in the normal way.

@yanghaichao12 I agree with Lars here, we need to understand first the problem, the normal workflow should be to open an Issue, explain the problem and how to reproduce it.
Then you can offer a possible solution and discuss it, once there is agreement you can file the PR.

yanghaichao12:
@uablrek if you do `route -6  add ...`, it would not use netlink but ioctl:
```shell
socket(PF_INET6, SOCK_DGRAM, IPPROTO_IP) = 3
ioctl(3, SIOCGIFINDEX, {ifr_name="ens3", }) = 0
ioctl(3, SIOCADDRT, 0x7ffef015d780)     = 0

```
ioctl has different behavior to netlink

yanghaichao12:
@aojea I'll close this PR first, and then open an issue to discuss this matter. Thank you anyway ：）

uablrek:
@yanghaichao12 Please make a reference to this PR in the issue.

uablrek:
Reproduced it (with a really old kernel)!
```
$ uname -a
Linux vm-001 3.16.82 #10 SMP Wed Apr 1 08:37:46 CEST 2020 x86_64 GNU/Linux
$ ip -6 route show table local
local ::1 dev lo proto unspec metric 0 
...
```
Seem to be kernel dependent. I have assumed "proto none" and "proto unspec" are the same thing just printed different depending on the version of `ip` (iproute2).

rikatz:
Assuming RHEL 7/CentOS7 uses Kernel 3.10, this might be a problem into it also?

Taking a look into my environment with CentOS7:

```
# ip -6 route show table local
local ::1 dev lo proto unspec metric 0 pref medium
```
Can move this to the opened issue, but feeding here as this might be a problem for the Enterprise Linux versions :)

uablrek:
BTW  I used exactly the same `ip` (iproute2) in 3.16.82 as in my tests with later kernels (with "proto kernel") so the difference is in the kernel not in the `ip` program.

yanghaichao12:
@uablrek hi ,you'r right it's not about ip program， i‘ve see the syscall：
```shell
sendmsg(3, {msg_name(12)={sa_family=AF_NETLINK, pid=0, groups=00000000}, msg_iov(1)=[{"\x40\x00\x00\x00\x14\x00\x05\x06\x90\xb5\x30\x5f\x00\x00\x00\x00\x0a\x80\x00\x00\x02\x00\x00\x00\x14\x00\x02\x00\x03\x00\x00\x00"..., 64}], msg_controllen=0, msg_flags=0}, 0) = 64
```
just  x14 but not x18 . 
and i've opened a issue #93858

pkg/proxy/ipvs/netlink_linux.go
```diff
@@ -173,10 +172,21 @@ func (h *netlinkHandle) GetLocalAddresses(dev, filterDev string) (sets.String, e
 		routeFilter.LinkIndex = chosenLinkIndex
 		filterMask |= netlink.RT_FILTER_OIF
 	}
+
 	routes, err := h.RouteListFiltered(netlink.FAMILY_ALL, routeFilter, filterMask)
 	if err != nil {
 		return nil, fmt.Errorf("error list route table, err: %v", err)
 	}
+
+	if h.isIPv6 {
```
Lion-Wei, line 181: Instead of list all ipv4 and ipv6 routes, then filter out, can we only list routes that we expected?

Lion-Wei:
/ok-to-test
Thanks for fixing this, I left one comment, please check. : )

pkg/proxy/ipvs/netlink_linux.go
yanghaichao12: okay,i'll check it and modify later,thx!

aojea:
Checking in, do we still want to merge this?

uablrek:
It has gone stale 2 times. I don't know. Perhaps documentation is enough as proposed https://github.com/kubernetes/kubernetes/issues/93858#issuecomment-747121595.

aojea:
let's close it then, if is a kernel bug we should expect distros to fix them.

/close

PR #93636, opened 2020-08-02
Added container name in the back-off event log

### LINKED ISSUE #93575: "Back-off restarting failed container": log which container failed

The event `Back-off restarting failed container` should become `Back-off restarting failed container <CONTAINER_NAME>`

**Why is this needed**:

Its hard to tell which container failed, you need to dig through the describe output and find something like

```
    State:          Waiting
      Reason:       CrashLoopBackOff
```

senthilnathan:
**What type of PR is this?**
/kind cleanup

Fixes an issue.

Fixes #93575 

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kubelet/kuberuntime/kuberuntime_manager.go
+++ b/pkg/kubelet/kuberuntime/kuberuntime_manager.go
@@ -854,7 +854,7 @@ func (m *kubeGenericRuntimeManager) doBackOff(pod *v1.Pod, container *v1.Contain
 	key := getStableKey(pod, container)
 	if backOff.IsInBackOffSince(key, ts) {
 		if ref, err := kubecontainer.GenerateContainerRef(pod, container); err == nil {
-			m.recorder.Eventf(ref, v1.EventTypeWarning, events.BackOffStartContainer, "Back-off restarting failed container")
+			m.recorder.Eventf(ref, v1.EventTypeWarning, events.BackOffStartContainer, "Back-off restarting failed container %s", container.Name)
 		}
 		err := fmt.Errorf("back-off %s restarting failed container=%s pod=%s", backOff.Get(key), container.Name, format.Pod(pod))
 		klog.V(3).Infof("%s", err.Error())


pkg/kubelet/kuberuntime/kuberuntime_manager.go
```diff
@@ -854,7 +854,7 @@ func (m *kubeGenericRuntimeManager) doBackOff(pod *v1.Pod, container *v1.Contain
 	key := getStableKey(pod, container)
 	if backOff.IsInBackOffSince(key, ts) {
 		if ref, err := kubecontainer.GenerateContainerRef(pod, container); err == nil {
-			m.recorder.Eventf(ref, v1.EventTypeWarning, events.BackOffStartContainer, "Back-off restarting failed container")
+			m.recorder.Eventf(ref, v1.EventTypeWarning, events.BackOffStartContainer, "Back-off restarting failed container %s", container.Name)
```
liggitt, line 857: /hold

This increases the back off event writes per pod … the Kubelet throttles event writes and simply omits events over a certain threshold, so increasing the number and variety of events this way could easily drop any events fro other pods

SergeyKanzhelev:
/unassign @dchen1107
/assign @dashpole 
/sig instrumentation

@dashpole what do you think the right fix for this would be? increase log level? Correlate event with the log? I would assume that the original issue reported by @howardjohn from some Istio-related scenario where (I assume) it's important to quickly identify whether it's a sidecar issue or main container's fault.

howardjohn:
> from some Istio-related scenario where (I assume) it's important to quickly identify whether it's a sidecar issue or main container's fault.

That was the use case, yes. Its not impossible to find out if you dig through some of the other output of `kubectl describe` but its certainly an improvement. I thought it would be a lot more trivial to implement though :slightly_smiling_face:

pkg/kubelet/kuberuntime/kuberuntime_manager.go
dashpole: Is this actually the case?  The ref provided looks like it is already at the container granularity, so i'm not certain this actually increases the number or variety of events.  But then again, i'm not familiar with the deduplication logic.

SergeyKanzhelev: Fast code search results. It looks like the max events count is calculated based on localKey:

https://github.com/kubernetes/kubernetes/blob/cea1d4e20b4a7886d8ff65f34c6d4f95efcb4742/staging/src/k8s.io/client-go/tools/record/events_cache.go#L266-L268

Local key is the message (second return value in here):

https://github.com/kubernetes/kubernetes/blob/cea1d4e20b4a7886d8ff65f34c6d4f95efcb4742/staging/src/k8s.io/client-go/tools/record/events_cache.go#L156-L173

So this change may in fact make everything worse - more events may be lost.

Given that you can get the name of failed container with `describe` command, closing this off for now.

/close

Issue #90866, opened 2020-05-07
Volume expansion tests are not taking into account supported driver sizes


msau42:
**Which test(s) are failing**:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/volume_expand.go

These tests are hardcoding PVC size to 1Mi and expanding the PVC to 1Gi. 

Some drivers don't support creating volumes that small, and compensate by rounding up to 1Gi. This means when the test actually tries to expand the volume to 1Gi, the underlying storage expansion may not get invoked because the volume is already 1Gi.

msau42:
@kubernetes/sig-storage-test-failures 
/assign @gnufied

PR #93654, opened 2020-08-03
Increase the default size to 2Gi

gnufied:
1Mi is too small of size to create initial volume and expect volume expansion. Increase default size to 2Gi.

xref - https://github.com/kubernetes/kubernetes/issues/90866


xing-yang:
@gnufied Can you rebase this PR?

gnufied:
@xing-yang jan pulled the change in https://github.com/kubernetes/kubernetes/pull/93654, so I am going to close this PR.

PR #93655, opened 2020-08-03
Deflake TestControllerSync

liggitt:
**What type of PR is this?**
/kind flake

Removes timing-sensitive unit test failure by ensuring we wait for an asynchronous queued task to complete before failing on absence of an operation in the PV timestamp cache

xref https://github.com/kubernetes/kubernetes/pull/93605#issuecomment-667428441

```release-note
NONE
```

/sig storage


pkg/controller/volume/persistentvolume/pv_controller_test.go
```diff
@@ -160,11 +161,15 @@ func TestControllerSync(t *testing.T) {
 				for len(ctrl.claims.ListKeys()) > 0 {
```
msau42, line 161: Opened https://github.com/kubernetes/kubernetes/issues/93662 for further cleanup

PR #88114, opened 2020-02-13
storage capacity tests

pohly:
**What type of PR is this?**
/kind failing-test

**What this PR does / why we need it**:

This adds tests for https://github.com/kubernetes/kubernetes/issues/72031.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #93658, opened 2020-08-03
test/e2e: fail test rather than flooding logs if PVC watch is closed prematurely

dcbw:
If the watch is closed prematurely the for/select loop will
spew "PVC event : <nil>" as fast as possible until the test times
out, flooding logs.

```
STEP: Checking PVC events
Aug  3 16:35:53.930: INFO: PVC event ADDED: &v1.PersistentVolumeClaim{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"pvc-zrchk", GenerateName:"pvc-", Namespace:"e2e-csi-mock-volumes-5292", SelfLink:"/api/v1/namespaces/e2e-csi-mock-volumes-5292/persistentvolumeclaims/pvc-zrchk", UID:"221369cd-50c3-48be-8c08-dc4d8f6f36e0", ResourceVersion:"42154", Generation:0, CreationTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:63732069329, loc:(*time.Location)(0x9e74040)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string(nil), OwnerReferences:[]v1.OwnerReference(nil), Finalizers:[]string{"kubernetes.io/pvc-protection"}, ClusterName:"", ManagedFields:[]v1.ManagedFieldsEntry{v1.ManagedFieldsEntry{Manager:"openshift-tests", Operation:"Update", APIVersion:"v1", Time:(*v1.Time)(0xc001639060), FieldsType:"FieldsV1", FieldsV1:(*v1.FieldsV1)(0xc001639080)}}}, Spec:v1.PersistentVolumeClaimSpec{AccessModes:[]v1.PersistentVolumeAccessMode{"ReadWriteOnce"}, Selector:(*v1.LabelSelector)(nil), Resources:v1.ResourceRequirements{Limits:v1.ResourceList(nil), Requests:v1.ResourceList{"storage":resource.Quantity{i:resource.int64Amount{value:1073741824, scale:0}, d:resource.infDecAmount{Dec:(*inf.Dec)(nil)}, s:"1Gi", Format:"BinarySI"}}}, VolumeName:"", StorageClassName:(*string)(0xc000876e90), VolumeMode:(*v1.PersistentVolumeMode)(0xc000876ec0), DataSource:(*v1.TypedLocalObjectReference)(nil)}, Status:v1.PersistentVolumeClaimStatus{Phase:"Pending", AccessModes:[]v1.PersistentVolumeAccessMode(nil), Capacity:v1.ResourceList(nil), Conditions:[]v1.PersistentVolumeClaimCondition(nil)}}
Aug  3 16:35:53.930: INFO: PVC event MODIFIED: &v1.PersistentVolumeClaim{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"pvc-zrchk", GenerateName:"pvc-", Namespace:"e2e-csi-mock-volumes-5292", SelfLink:"/api/v1/namespaces/e2e-csi-mock-volumes-5292/persistentvolumeclaims/pvc-zrchk", UID:"221369cd-50c3-48be-8c08-dc4d8f6f36e0", ResourceVersion:"42159", Generation:0, CreationTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:63732069329, loc:(*time.Location)(0x9e74040)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string{"volume.beta.kubernetes.io/storage-provisioner":"csi-mock-e2e-csi-mock-volumes-5292"}, OwnerReferences:[]v1.OwnerReference(nil), Finalizers:[]string{"kubernetes.io/pvc-protection"}, ClusterName:"", ManagedFields:[]v1.ManagedFieldsEntry{v1.ManagedFieldsEntry{Manager:"kube-controller-manager", Operation:"Update", APIVersion:"v1", Time:(*v1.Time)(0xc0016a6060), FieldsType:"FieldsV1", FieldsV1:(*v1.FieldsV1)(0xc0016a6080)}, v1.ManagedFieldsEntry{Manager:"openshift-tests", Operation:"Update", APIVersion:"v1", Time:(*v1.Time)(0xc0016a60a0), FieldsType:"FieldsV1", FieldsV1:(*v1.FieldsV1)(0xc0016a60c0)}}}, Spec:v1.PersistentVolumeClaimSpec{AccessModes:[]v1.PersistentVolumeAccessMode{"ReadWriteOnce"}, Selector:(*v1.LabelSelector)(nil), Resources:v1.ResourceRequirements{Limits:v1.ResourceList(nil), Requests:v1.ResourceList{"storage":resource.Quantity{i:resource.int64Amount{value:1073741824, scale:0}, d:resource.infDecAmount{Dec:(*inf.Dec)(nil)}, s:"1Gi", Format:"BinarySI"}}}, VolumeName:"", StorageClassName:(*string)(0xc0012aa020), VolumeMode:(*v1.PersistentVolumeMode)(0xc0012aa030), DataSource:(*v1.TypedLocalObjectReference)(nil)}, Status:v1.PersistentVolumeClaimStatus{Phase:"Pending", AccessModes:[]v1.PersistentVolumeAccessMode(nil), Capacity:v1.ResourceList(nil), Conditions:[]v1.PersistentVolumeClaimCondition(nil)}}
Aug  3 16:35:53.930: INFO: PVC event MODIFIED: &v1.PersistentVolumeClaim{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"pvc-zrchk", GenerateName:"pvc-", Namespace:"e2e-csi-mock-volumes-5292", SelfLink:"/api/v1/namespaces/e2e-csi-mock-volumes-5292/persistentvolumeclaims/pvc-zrchk", UID:"221369cd-50c3-48be-8c08-dc4d8f6f36e0", ResourceVersion:"43201", Generation:0, CreationTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:63732069329, loc:(*time.Location)(0x9e74040)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string{"pv.kubernetes.io/bind-completed":"yes", "pv.kubernetes.io/bound-by-controller":"yes", "volume.beta.kubernetes.io/storage-provisioner":"csi-mock-e2e-csi-mock-volumes-5292"}, OwnerReferences:[]v1.OwnerReference(nil), Finalizers:[]string{"kubernetes.io/pvc-protection"}, ClusterName:"", ManagedFields:[]v1.ManagedFieldsEntry{v1.ManagedFieldsEntry{Manager:"openshift-tests", Operation:"Update", APIVersion:"v1", Time:(*v1.Time)(0xc0016a6100), FieldsType:"FieldsV1", FieldsV1:(*v1.FieldsV1)(0xc0016a6120)}, v1.ManagedFieldsEntry{Manager:"kube-controller-manager", Operation:"Update", APIVersion:"v1", Time:(*v1.Time)(0xc0016a6140), FieldsType:"FieldsV1", FieldsV1:(*v1.FieldsV1)(0xc0016a6160)}}}, Spec:v1.PersistentVolumeClaimSpec{AccessModes:[]v1.PersistentVolumeAccessMode{"ReadWriteOnce"}, Selector:(*v1.LabelSelector)(nil), Resources:v1.ResourceRequirements{Limits:v1.ResourceList(nil), Requests:v1.ResourceList{"storage":resource.Quantity{i:resource.int64Amount{value:1073741824, scale:0}, d:resource.infDecAmount{Dec:(*inf.Dec)(nil)}, s:"1Gi", Format:"BinarySI"}}}, VolumeName:"pvc-221369cd-50c3-48be-8c08-dc4d8f6f36e0", StorageClassName:(*string)(0xc0012aa090), VolumeMode:(*v1.PersistentVolumeMode)(0xc0012aa0a0), DataSource:(*v1.TypedLocalObjectReference)(nil)}, Status:v1.PersistentVolumeClaimStatus{Phase:"Pending", AccessModes:[]v1.PersistentVolumeAccessMode(nil), Capacity:v1.ResourceList(nil), Conditions:[]v1.PersistentVolumeClaimCondition(nil)}}
Aug  3 16:35:53.930: INFO: PVC event MODIFIED: &v1.PersistentVolumeClaim{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"pvc-zrchk", GenerateName:"pvc-", Namespace:"e2e-csi-mock-volumes-5292", SelfLink:"/api/v1/namespaces/e2e-csi-mock-volumes-5292/persistentvolumeclaims/pvc-zrchk", UID:"221369cd-50c3-48be-8c08-dc4d8f6f36e0", ResourceVersion:"43203", Generation:0, CreationTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:63732069329, loc:(*time.Location)(0x9e74040)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string{"pv.kubernetes.io/bind-completed":"yes", "pv.kubernetes.io/bound-by-controller":"yes", "volume.beta.kubernetes.io/storage-provisioner":"csi-mock-e2e-csi-mock-volumes-5292"}, OwnerReferences:[]v1.OwnerReference(nil), Finalizers:[]string{"kubernetes.io/pvc-protection"}, ClusterName:"", ManagedFields:[]v1.ManagedFieldsEntry{v1.ManagedFieldsEntry{Manager:"openshift-tests", Operation:"Update", APIVersion:"v1", Time:(*v1.Time)(0xc0016a6600), FieldsType:"FieldsV1", FieldsV1:(*v1.FieldsV1)(0xc0016a6620)}, v1.ManagedFieldsEntry{Manager:"kube-controller-manager", Operation:"Update", APIVersion:"v1", Time:(*v1.Time)(0xc0016a6660), FieldsType:"FieldsV1", FieldsV1:(*v1.FieldsV1)(0xc0016a6680)}}}, Spec:v1.PersistentVolumeClaimSpec{AccessModes:[]v1.PersistentVolumeAccessMode{"ReadWriteOnce"}, Selector:(*v1.LabelSelector)(nil), Resources:v1.ResourceRequirements{Limits:v1.ResourceList(nil), Requests:v1.ResourceList{"storage":resource.Quantity{i:resource.int64Amount{value:1073741824, scale:0}, d:resource.infDecAmount{Dec:(*inf.Dec)(nil)}, s:"1Gi", Format:"BinarySI"}}}, VolumeName:"pvc-221369cd-50c3-48be-8c08-dc4d8f6f36e0", StorageClassName:(*string)(0xc0012aa450), VolumeMode:(*v1.PersistentVolumeMode)(0xc0012aa460), DataSource:(*v1.TypedLocalObjectReference)(nil)}, Status:v1.PersistentVolumeClaimStatus{Phase:"Bound", AccessModes:[]v1.PersistentVolumeAccessMode{"ReadWriteOnce"}, Capacity:v1.ResourceList{"storage":resource.Quantity{i:resource.int64Amount{value:1073741824, scale:0}, d:resource.infDecAmount{Dec:(*inf.Dec)(nil)}, s:"1Gi", Format:"BinarySI"}}, Conditions:[]v1.PersistentVolumeClaimCondition(nil)}}
Aug  3 16:35:53.930: INFO: PVC event : <nil>
Aug  3 16:35:53.930: INFO: PVC event : <nil>
Aug  3 16:35:53.930: INFO: PVC event : <nil>
Aug  3 16:35:53.930: INFO: PVC event : <nil>
Aug  3 16:35:53.930: INFO: PVC event : <nil>
Aug  3 16:35:53.930: INFO: PVC event : <nil>
Aug  3 16:35:53.930: INFO: PVC event : <nil>
Aug  3 16:35:53.930: INFO: PVC event : <nil>
Aug  3 16:35:53.930: INFO: PVC event : <nil>
<11 million lines later...>
Aug  3 16:36:26.571: INFO: PVC event : <nil>
Aug  3 16:36:26.571: INFO: PVC event : <nil>
Aug  3 16:36:26.571: INFO: PVC event : <nil>
Aug  3 16:36:26.571: INFO: PVC event : <nil>
<test times out>
```

/kind failing-test
/sig testing
/sig storage
/priority important-soon
/milestone v1.19

```release-note
NONE
```

### DIFF
--- a/test/e2e/storage/csi_mock_volume.go
+++ b/test/e2e/storage/csi_mock_volume.go
@@ -892,7 +892,11 @@ var _ = utils.SIGDescribe("CSI mock volume", func() {
 			loop:
 				for {
 					select {
-					case event := <-pvcWatch.ResultChan():
+					case event, ok := <-pvcWatch.ResultChan():
+						if !ok {
+							framework.Failf("PVC watch ended prematurely")
+						}
+
 						framework.Logf("PVC event %s: %#v", event.Type, event.Object)
 						switch event.Type {
 						case watch.Modified:


gnufied:
Looks like to be coming from - https://github.com/kubernetes/kubernetes/pull/88114 

@pohly also IIRC we agreed to not use events in e2e since they are not reliable. cc @msau42 @jsafrane 

/assign @pohly

pohly:
> also IIRC we agreed to not use events in e2e since they are not reliable.

It's not ideal, but in this case there's no other way to be sure that Kubernetes behaved as intended.

test/e2e/storage/csi_mock_volume.go
```diff
@@ -892,7 +892,11 @@ var _ = utils.SIGDescribe("CSI mock volume", func() {
 			loop:
 				for {
 					select {
-					case event := <-pvcWatch.ResultChan():
+					case event, ok := <-pvcWatch.ResultChan():
+						if !ok {
+							framework.Failf("PVC watch ended prematurely")
```
pohly, line 897: How often does this happen? I suppose it is caused by load on the apiserver?

I'm wondering whether it would be better to skip the test when this happens. As proposed here, we still have a flaky test failure, it just fails without flooding the logs.

dcbw: @pohly your call, I'm happy to skip or fail. I've seen it a couple of times on our OpenShift tests on GCP. It may be related to leader elections on the apiserver (which can happen at any time), which can cause short blips where the watchers close their existing watch and relist everything when the apiserver comes back.

pohly: I prefer skipping the test in this case.

smarterclayton: Watches can time out at ANY time, and both etcd and kube-apiserver are allowed to do this.

No one should ever read from a watch stream and assume how long that watch is open.  We reserve the right to close the watch at any point from kube-apiserver.

So please, fix the test to not assume things that aren't true :)

smarterclayton:
This code is clearly wrong, and the test makes an assumption (watch cannot be closed gracfeully) which is not true.

This breaks test infrastructure elsewhere which is mitigated by properly exposing the test as flaky (hot loops are bad), so i'm going to approve and tag.

We must follow up to have a resilient pattern for watches in this test, the framework is still buggy.

smarterclayton:
/lgtm
/approve

With the understanding a proper follow up must fix this to retry the watch (single shot watch is not safe)

smarterclayton:
generally you should use "watch.UntilWithRetry" OR the work we have added to conformance where you are examining the events as they come (to ensure they happen in the right order) and then retry the whole test if we can't verify.  That latter is effectively a for loop over the entire test which is "clean up, start scenario over again" up to a specific numebr of times.

test/e2e/storage/csi_mock_volume.go
pohly: > fix this to retry the watch (single shot watch is not safe)

I think that restarting the watch is not going to work: what the test tries to verify is that rescheduling happened. The only indication of that through the API is that the pod annotation gets set and later removed. If the watch is closed while the annotation is set and later gets restarted when it is set again, the test misses the "annotation was reset" part and fails incorrectly.

I see several options:
- stop using watches and instead rely on log parsing to detect whether rescheduling happened (not sure though whether there is anything in the logs about it)
- skip then when the watch is closed
- remove the test

I'm leaning towards skipping because log parsing may also be problematic and a test that usually (I know, that's an assumption) works is better than no test.

dcbw: Perhaps the test could use a SharedInformer on top of the watch which would handle restarts correctly via the Reflector, and the DeltaFifo would ensure that you have correct updates to the object itself. You don't care about actual watch events, you only care about changes to the PVC object itself, right?  SharedInformer would let you concentrate on just the object and not the watch.

pohly: But does SharedInformer guarantee that I see all updates? My understanding of it is that it is possible to miss updates, i.e. the only guarantee is that the callbacks eventually get the final state of each object.

dcbw: > But does SharedInformer guarantee that I see all updates? My understanding of it is that it is possible to miss updates, i.e. the only guarantee is that the callbacks eventually get the final state of each object.

@pohly doesn't that indicate a flaw in the scheme of annotation removal/addition? Any client using shared informer might miss those PVC annotation changes too, if they are collapsed into a single update by DeltaFIFO and thus the annotation removal is elided.

I tried the sharedinformer approach in https://github.com/kubernetes/kubernetes/pull/93720 but don't think I'd have the time to push it forward. But it still the flaw you allude to; I think clients and the testcase need to assume they will only see the *current* object state and perhaps not any intermediate states, and only act on that (eg reconciliation).

pohly: No, the actual usage of the annotation is not affected because it's value-triggered: if set, the external-provisioner does something. If unset, the scheduler does something. Nothing else changes the annotation, therefore it is fine to just know the current state.

The difference is that test wants to verify *how* we got to the current state. If the code and/or test is flawed, then we might reach the same observable end-state without actually having gone through the rescheduling cycle in the middle.

pohly: I ended up just skipping the additional checks, i.e. the test will still be considered a success: https://github.com/kubernetes/kubernetes/pull/93777

@dcbw: perhaps you can review that PR?

liggitt: > If the watch is closed while the annotation is set and later gets restarted when it is set again, the test misses the "annotation was reset" part and fails incorrectly.

The watchtools helpers provide a watch interface backed by an informer which will restart a dropped watch from the last observed event, so it won’t miss any events as long as it gets an updated starting resourceVersion before compaction, which it generally should because watch bookmarks are now enabled. I’d recommend using that.

PR #93660, opened 2020-08-03
Add LabelSelector validation in Pod Affinity/AntiAffinity Filter and Score plugins

damemi:
**What type of PR is this?**
/kind bug

The lack of this validation on incoming pods causes unpredictable cluster outcomes
when later calculating affinity results against existing pods (see #92714). This fix
quickly addresses the main source where these problems should be caught.

It is unfortunately difficult to add this validation directly to the API server due
to the fact that it may break migrations with existing pods that fail this check. This
is a compromise to address the current issue.

The alternative to this approach is adding error handling to `NewPodInfo` when checking the LabelSelectors there (see https://github.com/kubernetes/kubernetes/pull/92714#issuecomment-667227790)

Ref https://github.com/kubernetes/kubernetes/pull/92714

```release-note
Pods with invalid Affinity/AntiAffinity LabelSelectors will now fail scheduling when these plugins are enabled
```

```docs

```
/sig scheduling

### DIFF
--- a/pkg/scheduler/framework/plugins/interpodaffinity/filtering.go
+++ b/pkg/scheduler/framework/plugins/interpodaffinity/filtering.go
@@ -251,6 +251,9 @@ func (pl *InterPodAffinity) PreFilter(ctx context.Context, cycleState *framework
 	}
 
 	podInfo := framework.NewPodInfo(pod)
+	if podInfo.ParseError != nil {
+		return framework.NewStatus(framework.UnschedulableAndUnresolvable, fmt.Sprintf("parsing pod: %+v", podInfo.ParseError))
+	}
 
 	// existingPodAntiAffinityMap will be used later for efficient check on existing pods' anti-affinity
 	existingPodAntiAffinityMap := getTPMapMatchingExistingAntiAffinity(pod, havePodsWithAffinityNodes)
--- a/pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
+++ b/pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
@@ -160,9 +160,15 @@ func (pl *InterPodAffinity) PreScore(
 		}
 	}
 
+	podInfo := framework.NewPodInfo(pod)
+	if podInfo.ParseError != nil {
+		// Ideally we never reach here, because errors will be caught by PreFilter
+		return framework.NewStatus(framework.Error, fmt.Sprintf("parsing pod: %+v", podInfo.ParseError))
+	}
+
 	state := &preScoreState{
 		topologyScore: make(map[string]map[string]int64),
-		podInfo:       framework.NewPodInfo(pod),
+		podInfo:       podInfo,
 	}
 
 	topoScores := make([]scoreMap, len(allNodes))
--- a/pkg/scheduler/framework/v1alpha1/types.go
+++ b/pkg/scheduler/framework/v1alpha1/types.go
@@ -73,6 +73,7 @@ type PodInfo struct {
 	RequiredAntiAffinityTerms  []AffinityTerm
 	PreferredAffinityTerms     []WeightedAffinityTerm
 	PreferredAntiAffinityTerms []WeightedAffinityTerm
+	ParseError                 error
 }
 
 // AffinityTerm is a processed version of v1.PodAffinityTerm.
@@ -88,53 +89,50 @@ type WeightedAffinityTerm struct {
 	Weight int32
 }
 
-func newAffinityTerm(pod *v1.Pod, term *v1.PodAffinityTerm) *AffinityTerm {
+func newAffinityTerm(pod *v1.Pod, term *v1.PodAffinityTerm) (*AffinityTerm, error) {
 	namespaces := schedutil.GetNamespacesFromPodAffinityTerm(pod, term)
 	selector, err := metav1.LabelSelectorAsSelector(term.LabelSelector)
 	if err != nil {
-		klog.Errorf("Cannot process label selector: %v", err)
-		return nil
+		return nil, err
 	}
-	return &AffinityTerm{Namespaces: namespaces, Selector: selector, TopologyKey: term.TopologyKey}
+	return &AffinityTerm{Namespaces: namespaces, Selector: selector, TopologyKey: term.TopologyKey}, nil
 }
 
 // getAffinityTerms receives a Pod and affinity terms and returns the namespaces and
 // selectors of the terms.
-func getAffinityTerms(pod *v1.Pod, v1Terms []v1.PodAffinityTerm) []AffinityTerm {
+func getAffinityTerms(pod *v1.Pod, v1Terms []v1.PodAffinityTerm) ([]AffinityTerm, error) {
 	if v1Terms == nil {
-		return nil
+		return nil, nil
 	}
 
 	var terms []AffinityTerm
 	for _, term := range v1Terms {
-		t := newAffinityTerm(pod, &term)
-		if t == nil {
-			// We get here if the label selector failed to process, this is not supposed
-			// to happen because the pod should have been validated by the api server.
-			return nil
+		t, err := newAffinityTerm(pod, &term)
+		if err != nil {
+			// We get here if the label selector failed to process
+			return nil, err
 		}
 		terms = append(terms, *t)
 	}
-	return terms
+	return terms, nil
 }
 
 // getWeightedAffinityTerms returns the list of processed affinity terms.
-func getWeightedAffinityTerms(pod *v1.Pod, v1Terms []v1.WeightedPodAffinityTerm) []WeightedAffinityTerm {
+func getWeightedAffinityTerms(pod *v1.Pod, v1Terms []v1.WeightedPodAffinityTerm) ([]WeightedAffinityTerm, error) {
 	if v1Terms == nil {
-		return nil
+		return nil, nil
 	}
 
 	var terms []WeightedAffinityTerm
 	for _, term := range v1Terms {
-		t := newAffinityTerm(pod, &term.PodAffinityTerm)
-		if t == nil {
-			// We get here if the label selector failed to process, this is not supposed
-			// to happen because the pod should have been validated by the api server.
-			return nil
+		t, err := newAffinityTerm(pod, &term.PodAffinityTerm)
+		if err != nil {
+			// We get here if the label selector failed to process
+			return nil, err
 		}
 		terms = append(terms, WeightedAffinityTerm{AffinityTerm: *t, Weight: term.Weight})
 	}
-	return terms
+	return terms, nil
 }
 
 // NewPodInfo return a new PodInfo
@@ -150,12 +148,32 @@ func NewPodInfo(pod *v1.Pod) *PodInfo {
 		}
 	}
 
+	// Attempt to parse the affinity terms
+	var parseErr error
+	requiredAffinityTerms, err := getAffinityTerms(pod, schedutil.GetPodAffinityTerms(pod.Spec.Affinity))
+	if err != nil {
+		parseErr = fmt.Errorf("requiredAffinityTerms: %w", err)
+	}
+	requiredAntiAffinityTerms, err := getAffinityTerms(pod, schedutil.GetPodAntiAffinityTerms(pod.Spec.Affinity))
+	if err != nil {
+		parseErr = fmt.Errorf("requiredAntiAffinityTerms: %w", err)
+	}
+	weightedAffinityTerms, err := getWeightedAffinityTerms(pod, preferredAffinityTerms)
+	if err != nil {
+		parseErr = fmt.Errorf("preferredAffinityTerms: %w", err)
+	}
+	weightedAntiAffinityTerms, err := getWeightedAffinityTerms(pod, preferredAntiAffinityTerms)
+	if err != nil {
+		parseErr = fmt.Errorf("preferredAntiAffinityTerms: %w", err)
+	}
+
 	return &PodInfo{
 		Pod:                        pod,
-		RequiredAffinityTerms:      getAffinityTerms(pod, schedutil.GetPodAffinityTerms(pod.Spec.Affinity)),
-		RequiredAntiAffinityTerms:  getAffinityTerms(pod, schedutil.GetPodAntiAffinityTerms(pod.Spec.Affinity)),
-		PreferredAffinityTerms:     getWeightedAffinityTerms(pod, preferredAffinityTerms),
-		PreferredAntiAffinityTerms: getWeightedAffinityTerms(pod, preferredAntiAffinityTerms),
+		RequiredAffinityTerms:      requiredAffinityTerms,
+		RequiredAntiAffinityTerms:  requiredAntiAffinityTerms,
+		PreferredAffinityTerms:     weightedAffinityTerms,
+		PreferredAntiAffinityTerms: weightedAntiAffinityTerms,
+		ParseError:                 parseErr,
 	}
 }
 
--- a/pkg/scheduler/framework/plugins/interpodaffinity/filtering_test.go
+++ b/pkg/scheduler/framework/plugins/interpodaffinity/filtering_test.go
@@ -19,6 +19,7 @@ package interpodaffinity
 import (
 	"context"
 	"reflect"
+	"strings"
 	"testing"
 
 	"k8s.io/api/core/v1"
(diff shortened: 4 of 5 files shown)


pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
```diff
@@ -144,6 +168,10 @@ func (pl *InterPodAffinity) PreScore(
 	hasAffinityConstraints := affinity != nil && affinity.PodAffinity != nil
 	hasAntiAffinityConstraints := affinity != nil && affinity.PodAntiAffinity != nil
 
+	if err := validatePreferredAffinityTerms(pod); err != nil {
```
soulxu, line 171: Is it ok to check the podInfo.PreferredAffinityTerms and podInfo.PreferredAntiAffinityTerms is nil or not in line 195? Since the NewPodInfo already did the same check, then we can reuse that result.

damemi: @soulxu using the check that's done in `NewPodInfo` was actually the other option I considered in https://github.com/kubernetes/kubernetes/pull/92714#issuecomment-667227790 for both PreScore and PreFilter. The tradeoff with checking it there is that `NewPodInfo` would then need to add error handling everywhere it is called (which is quite a few places).

So while this approach is redundant, I think it is clearer what is being checked and does it as early as possible. By making the check distinct we can also ensure that it stays consistent between versions (as various refactors have already changed how this error is handled, see my main comment in #92714)

Of course if people would prefer that we use the check in `NewPodInfo` I am open to doing it that way as well and can make a PR with that too

damemi: Oh I see, or we can just get the AffinityTerms lists *from* the PodInfo returned by NewPodInfo. Yeah that could save us having to do the check again I agree

soulxu: > Oh I see, or we can just get the AffinityTerms lists _from_ the PodInfo returned by NewPodInfo. Yeah that could save us having to do the check again I agree

Yea

damemi: Actually, that won't work because if NewPodInfo hits an error parsing the label selector, it just throws out the error and returns nil so we wouldn't be able to actually check that selector. Meaning we would still have to bubble the error up like in https://github.com/kubernetes/kubernetes/pull/92714#issuecomment-667227790

pkg/scheduler/framework/plugins/interpodaffinity/filtering.go
```diff
@@ -238,6 +239,29 @@ func getTPMapMatchingIncomingAffinityAntiAffinity(podInfo *framework.PodInfo, al
...
+		}
+	}
+	if pod.Spec.Affinity.PodAntiAffinity != nil {
+		for _, term := range pod.Spec.Affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution {
+			terms = append(terms, term)
+		}
+	}
+	for _, term := range terms {
+		if _, err := metav1.LabelSelectorAsSelector(term.LabelSelector); err != nil {
+			return err
```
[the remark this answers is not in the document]
damemi: I considered that, see https://github.com/kubernetes/kubernetes/pull/92714#issuecomment-667227790 and https://github.com/kubernetes/kubernetes/pull/93660#discussion_r464772308. basically NewPodInfo would require adding error handling in a lot of different places which is why I went with this approach first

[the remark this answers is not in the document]
alculquicondor: I was thinking of just storing the error and surfacing it on PreFilter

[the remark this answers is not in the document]
damemi: Storing it how? New field in PodInfo?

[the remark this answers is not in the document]
alculquicondor: yes. Not sure what others think of that though.

pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
soulxu: how about check like this: "pod.Spec.Affinity.PodAffinity.PreferredDuringSchedulingIgnoredDuringExecution != nil and podInfo.PreferredAffinityTerms == nil", then we know the pod has preferred affinity but newPodInfo return nil, it means there is invalid affinity.

alculquicondor: fine with me. The point is avoiding parsing, which has proven to be costly. Although it's acceptable one per scheduling cycle.

damemi: > The point is avoiding parsing, which has proven to be costly

Are you referring to the `metav1.LabelSelectorAsSelector` call? It would still be good to call that in order to get the error, or in that case lean toward storing it in PodInfo like you said in https://github.com/kubernetes/kubernetes/pull/93660#discussion_r465066386

alculquicondor: your call

damemi:
@soulxu @alculquicondor thanks for the feedback, I updated it to go with the suggestion to store the error in the `PodInfo` and simply check that once it's created. This means we only have to parse the labelSelectors (not to mention check if they're nil) only once. Please take a look, and let me know what you think

pkg/scheduler/framework/plugins/interpodaffinity/filtering.go
```diff
@@ -251,6 +251,9 @@ func (pl *InterPodAffinity) PreFilter(ctx context.Context, cycleState *framework
 	}
 
 	podInfo := framework.NewPodInfo(pod)
+	if podInfo.PodError != nil {
+		return framework.NewStatus(framework.Error, fmt.Sprintf("error parsing pod: %+v", err))
```
alculquicondor, line 255:
```suggestion
		return framework.NewStatus(framework.Error, fmt.Sprintf("parsing pod: %w", err))
```

pkg/scheduler/framework/v1alpha1/types.go
```diff
@@ -73,6 +73,7 @@ type PodInfo struct {
 	RequiredAntiAffinityTerms  []AffinityTerm
 	PreferredAffinityTerms     []WeightedAffinityTerm
 	PreferredAntiAffinityTerms []WeightedAffinityTerm
+	PodError                   error
```
alculquicondor, line 76: nit: `ParsingErr`

pkg/scheduler/framework/v1alpha1/types.go
```diff
@@ -150,12 +148,18 @@ func NewPodInfo(pod *v1.Pod) *PodInfo {
 		}
 	}
 
+	requiredAffinityTerms, podErr := getAffinityTerms(pod, schedutil.GetPodAffinityTerms(pod.Spec.Affinity))
+	requiredAntiAffinityTerms, podErr := getAffinityTerms(pod, schedutil.GetPodAntiAffinityTerms(pod.Spec.Affinity))
+	weightAffinityTerms, podErr := getWeightedAffinityTerms(pod, preferredAffinityTerms)
```
alculquicondor, line 153: worth wrapping (`fmt.Errorf`) the errors here to indicate which one failed

damemi: Done, I wrapped them in this spot with just `requiredAffinityTerm,preferredAntiAffinity,etc`. The other option would be wrapping the error in `newAffinityTerm` with the labelSelector that failed but I don't have a preference

pkg/scheduler/framework/v1alpha1/types.go
```diff
@@ -150,12 +148,32 @@ func NewPodInfo(pod *v1.Pod) *PodInfo {
 		}
 	}
 
+	// Parse affinity terms and if any return an error, wrap the error indicating which was the last to fail
```
alculquicondor, line 151: nit: no need for the comment, or at least not the wrapping part :)

alculquicondor (commented):
/approve

alculquicondor:
/lgtm

/assign @ahg
for milestone

alculquicondor:
/assign @ahg-g 
made a mistake again :)

ahg-g:
/milestone v1.19

Thanks Mike, we can't patch this to 1.18 as is, we need to have a different PR.

damemi:
@ahg-g could you lgtm again? Had to re-push because of 
```
# k8s.io/kubernetes/pkg/scheduler/framework/plugins/interpodaffinity
pkg/scheduler/framework/plugins/interpodaffinity/filtering.go:255:47: Sprintf call has error-wrapping directive %w
pkg/scheduler/framework/plugins/interpodaffinity/scoring.go:165:47: Sprintf call has error-wrapping directive %w
make[1]: *** [Makefile:387: vet] Error 1
```

I don't think using `%w` is right there since we're just passing a string to `framework.NewStatus`.

alculquicondor:
Yeah, I missed that. `%` is for Errorf

damemi:
> Thanks Mike, we can't patch this to 1.18 as is, we need to have a different PR.

I'll work on a separate fix for 1.18

pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
```diff
@@ -160,9 +160,14 @@ func (pl *InterPodAffinity) PreScore(
 		}
 	}
 
+	podInfo := framework.NewPodInfo(pod)
+	if podInfo.ParseError != nil {
+		return framework.NewStatus(framework.Error, fmt.Sprintf("parsing pod: %+v", err))
```
alculquicondor, line 165: /hold

Shouldn't we return `UnschedulableAndUnresolvable`?

`Error` implies internal error, but this is a user error.

damemi: `UnschedulableAndUnresolvable` seems to only be parsed in Filter plugins, and the [error code definitions](https://github.com/kubernetes/kubernetes/blob/release-1.19/pkg/scheduler/framework/v1alpha1/interface.go#L59) imply that unexpected input could cause an error.

So `UnschedulableAndUnresolvable` could work for the Filter plugin, but score just returns anything that's not success as an error (https://github.com/kubernetes/kubernetes/blob/release-1.19/pkg/scheduler/core/generic_scheduler.go#L512-L516) (also, in 1.18 [this case is already returned as a `framework.Error`](https://github.com/kubernetes/kubernetes/blob/release-1.18/pkg/scheduler/framework/plugins/interpodaffinity/filtering.go#L350) from the Filter plugin)

I think in the case of a syntax error, `Error` makes sense. `UnschedulableAndUnresolvable` sounds more related to cluster state that prevents the pod from scheduling (ie, it fails the filters and nothing can be preempted to fit it)

alculquicondor: What about doing UnschedulableAndUnresolvable for PreFilter, but just proceed with scheduling (ignore the badly formatted selectors) in the case of Score?

That doesn't solve the issue for 1.18 though. Unless we do what we are doing in 1.19.

The problem with using Error is that it messes up metrics. The metric should help identify internal errors, as opposed to user errors.

ahg-g: We could do both checks (preferred and required constraints) in PreFilter and return `UnschedulableAndUnresolvable`. I think PreFilter is the right place to do any pod-level checks.

ahg-g: I don't think this is going to mess up metrics too much, so I think it is fine if we are not accurate in 1.18.

ahg-g: > What about doing UnschedulableAndUnresolvable for PreFilter, but just proceed with scheduling (ignore the badly formatted selectors) in the case of Score?


I think we should check the score so the user get a signal that there is something wrong with their pod spec rather than going undetected.

ahg-g: wait, we are not going to get here anyways if there is any error because of the PreFilter check (unless the PreFilter is disabled).

damemi: > just proceed with scheduling (ignore the badly formatted selectors) in the case of Score?

this was what caused one of the initial problems, a pod with bad formatting was ignored during scheduling and then broke scheduling for any subsequent pods that tried to check against the existing pod's affinity. that's actually how we found the bug, so we need to return some kind of error.

> The metric should help identify internal errors, as opposed to user errors.

Note this isn't a scheduling bug, it's a hack to patch an API validation bug. Ideally, just trying to create the pod would return a validation error from the api server but this is our workaround to accomplish the same effect of preventing the pod from running. 

I think that blurs the line between Error and Unschedulable but both are treated as !=Success so I'm fine with returning either. The only other thing is that it's already an Error from PreFilter in 1.18, so that would change too

damemi: @alculquicondor pushed an update to change to `UnschedulableAndUnresolvable` in filter and score, wdyt?

ahg-g: we should probably keep using Error in PreScore to emphasize that `UnschedulableAndUnresolvable` is not a PreScore expected status.

pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
```diff
@@ -160,9 +160,14 @@ func (pl *InterPodAffinity) PreScore(
 		}
 	}
 
+	podInfo := framework.NewPodInfo(pod)
+	if podInfo.ParseError != nil {
```
alculquicondor, line 164: nit: add comment:

`This should have already failed on PreScore, unless extension point is disabled`

damemi: Assuming you mean `PreFilter`? (this is prescore 🙂)

And I agree it's unlikely someone disables the filter plugin and leaves the score plugin, but definitely possible which is why it doesn't hurt to handle the error in both.

Based on https://github.com/kubernetes/kubernetes/pull/93660#discussion_r466634075, going to use `Error` from PreScore and `UnschedulableAndUnresolvable` from PreFilter. That sounds good to me

alculquicondor: sg!

damemi:
As always, thanks for the help and feedback @alculquicondor and @ahg-g!

liggitt:
Shouldn't this include tests exercising the error condition?

alculquicondor:
/hold

Sure thing. @damemi should we put it in `pkg/scheduler/scheduler_test.go` or can you think of a better location?

damemi:
> /hold
> 
> Sure thing. @damemi should we put it in `pkg/scheduler/scheduler_test.go` or can you think of a better location?

Maybe right in `scoring_test.go` and `filtering_test.go` for the plugins? That looks simplest to me

alculquicondor:
Not really. We need a more "e2e" test

liggitt:
> Not really. We need a more "e2e" test

is the concern how the system behaves overall if an error is returned here? do we already have a test covering the "scheduler plugin returns an error on a pod" scenario? if so, a unit test might be sufficient for this.

alculquicondor:
> is the concern how the system behaves overall if an error is returned here? do we already have an e2e-type test covering the "scheduler plugin returns an error on a pod" scenario? if so, a unit test might be sufficient for this.

The concern is that the error is transmitted from the point the pod enters the queue until it gets an scheduled attempt. But I'm not suggesting a real E2E test, just a unit test that goes across some packages.

damemi:
I started working on basic unit tests to verify the PreScore/PreFilter functions return an error before @alculquicondor commented back, so I pushed those. Working on an "e2e-like" test now, but I had the same thought as @liggitt that if we already test that an error from Pre* functions cancel scheduling then the cycle check could already be covered

ahg-g:
I think a unit test in the plugin is sufficient.

alculquicondor:
ok with me

alculquicondor (commented):
/hold cancel
/lgtm

PR #93661, opened 2020-08-03
Deflake scheduler PostBind integration test

### LINKED ISSUE #93659: [Flaky Test] scheduler integration test TestPostBindPlugin
**Which jobs are flaking**:

- ci-kubernetes-integration-master
- pull-kubernetes-integration

TestPostBindPlugin

**Testgrid link**:

- https://testgrid.k8s.io/sig-release-master-blocking#integration-master&width=5&include-filter-by-regex=test%2Fintegration%2Fscheduler
- https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-integration&width=5&include-filter-by-regex=test%2Fintegration%2Fscheduler

**Reason for failure**:

```
framework_test.go:1337: Expected the postbind plugin to be called, was called 0 times.
```

- links to go.k8s.io/triage appreciated: https://storage.googleapis.com/k8s-gubernator/triage/index.html?test=TestPostBindPlugin

/sig scheduling

Huang-Wei:
**What type of PR is this?**

/kind flake
/sig scheduling

For a successful scheduling, `PostBind()` is for sure called after `Bind()`. However, when the test detects the pod is assigned ".spec.nodeName" **_right_** after `Bind()` (before `PostBind()`), and process to check the postFilter counter, the test may flake.

This PR tries to fix this flake by waiting for `PostBind()` to complete. Additionally, it also wraps the scheduler initialization logic into the sub-test, to make each sub-test independent.

Fixes #93659

The flake doesn't seem to block 1.19 release, so I think it can be merged to 1.20.

```release-note
NONE
```

### DIFF
--- a/test/integration/scheduler/framework_test.go
+++ b/test/integration/scheduler/framework_test.go
@@ -1262,45 +1262,9 @@ func TestBindPlugin(t *testing.T) {
 
 // TestPostBindPlugin tests invocation of postbind plugins.
 func TestPostBindPlugin(t *testing.T) {
-	// Create a plugin registry for testing. Register a prebind and a postbind plugin.
-	preBindPlugin := &PreBindPlugin{}
-	postBindPlugin := &PostBindPlugin{name: postBindPluginName}
-	registry := frameworkruntime.Registry{
-		preBindPluginName:  newPlugin(preBindPlugin),
-		postBindPluginName: newPlugin(postBindPlugin),
-	}
-
-	// Setup initial prebind and postbind plugin for testing.
-	prof := schedulerconfig.KubeSchedulerProfile{
-		SchedulerName: v1.DefaultSchedulerName,
-		Plugins: &schedulerconfig.Plugins{
-			PreBind: &schedulerconfig.PluginSet{
-				Enabled: []schedulerconfig.Plugin{
-					{
-						Name: preBindPluginName,
-					},
-				},
-			},
-			PostBind: &schedulerconfig.PluginSet{
-				Enabled: []schedulerconfig.Plugin{
-					{
-						Name: postBindPluginName,
-					},
-				},
-			},
-		},
-	}
-
-	// Create the master and the scheduler with the test plugin set.
-	testCtx := initTestSchedulerForFrameworkTest(t, testutils.InitTestMaster(t, "postbind-plugin", nil), 2,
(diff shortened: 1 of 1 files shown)


Huang-Wei:
`diff -w` would get a better diff for review.

/assign @ahg-g 
/priority important-longterm

test/integration/scheduler/framework_test.go
```diff
@@ -1333,13 +1336,16 @@ func TestPostBindPlugin(t *testing.T) {
 				if err = testutils.WaitForPodToSchedule(testCtx.ClientSet, pod); err != nil {
 					t.Errorf("Expected the pod to be scheduled. error: %v", err)
 				}
+				select {
+				case <-postBindPlugin.pluginInvokeEventChan:
+				case <-time.After(time.Second * 15):
+					t.Errorf("pluginInvokeEventChan timed out")
+				}
 				if postBindPlugin.numPostBindCalled == 0 {
```
soulxu, line 1344: Emm...this check should be useless? since if postBindPlugin is called, the `pluginInvokeEventChan` will return, if the postBindPlugin isn't called, the timeout will return.

Huang-Wei: > if the postBindPlugin isn't called, the timeout will return.

`t.Errorf` won't return - the logic proceeds to L1344.

soulxu: > > if the postBindPlugin isn't called, the timeout will return.
> 
> `t.Errorf` won't return - the logic proceeds to L1344.

Yea, I mean only the t.Errorf at line 1342 is enough, line 1342 already ensure the plugin isn't called.

alculquicondor (commented):
/lgtm

ahg-g:
/milestone v1.19

I think we should still include it in 1.19.

test/integration/scheduler/framework_test.go
Huang-Wei: I see your point. Although the root cause of L1342 and L1344 ~ L1346 is the same, but their symptoms are different, so IMO keeping L1344 ~ L1346 as is can make the semantics more explicit.

soulxu: got it, thanks

PR #91645, opened 2020-06-01
Promote new Event API to v1

chelseychen:
**What type of PR is this?**

/kind api-change

**What this PR does / why we need it**:

Migrate new Event API to GA.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs
- [KEP]: <https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/383-new-event-api-ga-graduation>
```

PR #93663, opened 2020-08-03
Set DeprecatedLastTimestamp and DeprecatedFirstTimestamp so LastTimestamp and FirstTimestamp are still set

### LINKED ISSUE #89689: kubectl get events LAST SEEN is <unknown>
**What happened**:

When use `kubectl get events` show events, some events `LAST SEEN` is `<unknown>`,  i use `kubectl get events -o json` found these events `firstTimestamp` is null. I want to know why?

I compared to v1.11.6, events have no this problem.

I want to know why some events firstTimestamp is null.

1. create a pod.

**Environment**:

`v1.17.2`

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

ingvagabund:
**What type of PR is this?**
/kind bug

Set `DeprecatedLastTimestamp` and `DeprecatedFirstTimestamp` so `LastTimestamp` and `FirstTimestamp` are still set when listing events with Last Seen column, resp. describing events when running `kubectl describe pod ...`

Also, when "oc get events --sort-by='.lastTimestamp'" it's essential to have all lastTimestamp fields
set. Otherwise, the following error message is produced:

```
F0715 09:18:21.343259   22629 sorter.go:354] Field {.lastTimestamp} in [][][]reflect.Value is an unsortable type: interface, err: unsortable interface: interface
```

The current output (LAST SEEN column is shown as `<unknown> ` when `LastTimestamp` is not set.
```
$ oc get events
LAST SEEN   TYPE      REASON                                OBJECT                                             MESSAGE
<unknown>   Normal    Scheduled                             pod/test                                           Successfully assigned default/test to ip-10-0-181-188.us-west-2.compute.internal
<unknown>   Normal    Scheduled                             pod/test                                           Successfully assigned default/test to ip-10-0-181-188.us-west-2.compute.internal
<unknown>   Normal    Scheduled                             pod/test                                           Successfully assigned default/test to ip-10-0-181-188.us-west-2.compute.internal

```

When neither of `LastTimestamp` nor `FirstTimestamp` are set:
```
Name:         annotation-second-scheduler
Namespace:    default
Events:
  Type    Reason          Age        From                                                 Message
  ----    ------          ----       ----                                                 -------
  Normal  Scheduled       <unknown>                                                       Successfully assigned default/annotation-second-scheduler to ip-xxx

```

Info: it's still valid to set lastTimestamp field when converting v1betav1/v1 event type to the core one before 1.19 (where the validation forbids the field to be set: https://github.com/kubernetes/kubernetes/pull/91645)

Fixes #89689

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/client-go/tools/events/event_broadcaster.go
+++ b/staging/src/k8s.io/client-go/tools/events/event_broadcaster.go
@@ -113,6 +113,7 @@ func (e *eventBroadcasterImpl) refreshExistingEventSeries() {
 	defer e.mu.Unlock()
 	for isomorphicKey, event := range e.eventCache {
 		if event.Series != nil {
+			event.DeprecatedLastTimestamp = metav1.Time{Time: event.Series.LastObservedTime.Time}
 			if recordedEvent, retry := recordEvent(e.sink, event); !retry {
 				if recordedEvent != nil {
 					e.eventCache[isomorphicKey] = recordedEvent
@@ -175,6 +176,9 @@ func (e *eventBroadcasterImpl) recordToSink(event *v1beta1.Event, clock clock.Cl
 			return eventCopy
 		}()
 		if evToRecord != nil {
+			if evToRecord.Series != nil {
+				evToRecord.DeprecatedLastTimestamp = metav1.Time{Time: evToRecord.Series.LastObservedTime.Time}
+			}
 			recordedEvent := e.attemptRecording(evToRecord)
 			if recordedEvent != nil {
 				recordedEventKey := getKey(recordedEvent)
@@ -251,6 +255,7 @@ func recordEvent(sink EventSink, event *v1beta1.Event) (*v1beta1.Event, bool) {
 func createPatchBytesForSeries(event *v1beta1.Event) ([]byte, error) {
 	oldEvent := event.DeepCopy()
 	oldEvent.Series = nil
+	oldEvent.DeprecatedLastTimestamp = metav1.Time{}
 	oldData, err := json.Marshal(oldEvent)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


yastij:
queueing this

/assign @yastij @wojtek-t

staging/src/k8s.io/client-go/tools/events/event_broadcaster.go
```diff
@@ -113,6 +113,7 @@ func (e *eventBroadcasterImpl) refreshExistingEventSeries() {
 	defer e.mu.Unlock()
 	for isomorphicKey, event := range e.eventCache {
 		if event.Series != nil {
+			event.DeprecatedLastTimestamp = metav1.Time{Time: event.Series.LastObservedTime.Time}
```
wojtek-t, line 116: Hmm - those are supposed to be handled here:
https://github.com/kubernetes/kubernetes/blob/master/pkg/printers/internalversion/printers.go#L1719

I think we should rather fix the printer itself to fix this issue if that logic isn't correct.

ingvagabund: `event.Series` is nil the first time the event is created. After that it takes minutes (given by `refreshTime`, resp. `finishTime`) before the event is updated (with `event.Series` getting set). Until that, `event.Series` is still nil and there's no way to deduce `LastTimestamp`.

wojtek-t: I don't understand. Series field will be updated on each subsequent update of the event.

wojtek-t: One of the important decision behind the API was exactly to not send every single event. So yes, the LastUpdate often will not be accurate to the time when it really happened in practice, but those events won't be send to apiserver thus there is no way to detect this time.

yastij:
closing per https://github.com/kubernetes/kubernetes/issues/94019#issuecomment-677658234

/close

PR #93672, opened 2020-08-04
add unit test for helper.go

nayihz:
**What type of PR is this?**

> /kind cleanup

add unit test for helper.go

add unit test

### DIFF
--- a/test/images/agnhost/netexec/netexec.go
+++ b/test/images/agnhost/netexec/netexec.go
@@ -491,9 +491,9 @@ func hostNameHandler(w http.ResponseWriter, r *http.Request) {
 // udp server supports the hostName, echo and clientIP commands.
 func startUDPServer(udpPort int) {
 	serverAddress, err := net.ResolveUDPAddr("udp", fmt.Sprintf(":%d", udpPort))
-	assertNoError(err)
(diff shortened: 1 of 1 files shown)


latteczy:
hi @smarterclayton @caesarxuchao @yujuhong , can anybody add ok-to-test label for this pr? Thanks.

pkg/apis/core/v1/helper/helpers_test.go
```diff
@@ -1509,3 +1509,93 @@ func TestHugePageUnitSizeFromByteSize(t *testing.T) {
 		}
 	}
 }
+
+func TestLoadBalancerStatusEqual(t *testing.T) {
+
+	testCases := []struct {
```
oomichi, line 1515: It is better to put `name` for each test pattern and output the name at line 1597 when the pattern is failed.

oomichi (commented):
LoadBalancerStatusEqual() is not covered on unit tests yet, thanks for doing this.
Just one comment about error message.

/cc @oomichi

pkg/apis/core/v1/helper/helpers_test.go
latteczy: https://github.com/kubernetes/kubernetes/pull/96353

Issue #93637, opened 2020-08-02
Fix root OWNERS approvers


dims:
```
Preface: IMO, we should not fix root OWNERS approvers by adding more people to it. We should fix it by distributing the 
responsibility (per area) to SIG TLs, and sub-project TLs. I actually think we should consider removing folks from root OWNERS approvers.
```

Follow up from: https://github.com/kubernetes/kubernetes/pull/85638#issuecomment-563356854

@bgrant0607 @brendandburns @dchen1107 @lavalamp @liggitt @smarterclayton @thockin @wojtek-t , your thoughts please!

thockin:
Without a way to self-limit, I think the group should be kept very small.  As I think about it, I intentionally DON'T approve PRs because I don't want it to apply too broadly.

I suggested on slack for contribex: What if I could `/approve <dir>` and that would only count as approval for that specific
directory?  I often want to /approve an API change but not the whole PR.  Talking to other root-OWNERs (overly-empowered are we) it seems like a useful self-limit on authority.

dims:
@spiffxp @nikhita WDYT?

spiffxp:
I want to get @cjwagner's input, I think he's got the most prior experience with approvers and repoowners

I think it's a good idea.  My gut tells me it would be somewhat involved, N weeks ish

spiffxp:
I also think, separately, that we should consider hitting root OWNERS a failure mode and immediately address by adding OWNERS in a deeper/more relevant dir.  

It used to be that the approval message would suggest the fewest possible approvers who could cover N OWNERS files but I _think_ it now outputs leaf approvers. 
 
Something like: either the root owner should boostrap a deeper file with themselves and then seek out more appropriate people, or the author of the PR shouldn't be allowed to merge it until they've added an OWNERS file that would rectify the situation going forward.

bgrant0607:
Regarding removing people from root approvers, I volunteer to be moved to emeritus status, as I can't respond to most/any PR approval requests, anyway.

bgrant0607:
Do we have a way to audit uses of root OWNERS approval power? If not, that would be useful to implement.

I used to get regular pings to approve changes to CHANGELOG.md. CONTRIBUTING.md and .gitignore were recently updated.

go.mod already has a distinct reviewer set.

BenTheElder:
Unfortunately the go.mod reviewer set cannot actually approve (it doesn't
function) so mostly it just results in auto assigning the entire team to
[...]

PR #93683, opened 2020-08-04
Move brendandburns to to emeritus status.

dims:

**What type of PR is this?**
/kind cleanup

Per note in https://github.com/kubernetes/kubernetes/issues/93637#issuecomment-668383474 @brendandburns has requested to be marked as emeritus.

Fixes #

@brendanburns - Thank you for all your work over the years! Should you find the time again to dive into things, we'll be so happy to move you back.

```release-note
NONE
```

```docs

```

/sig contributor-experience
/priority important-soon

/assign @brendandburns

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -10,7 +10,6 @@ filters:
       - thockin
       - liggitt
     approvers:
-      - brendandburns
       - dchen1107
       - lavalamp
       - smarterclayton
@@ -18,6 +17,7 @@ filters:
       - wojtek-t
       - liggitt
     emeritus_approvers:
+      - brendandburns
       - bgrant0607
       - jbeda
 


spiffxp:
@dims 
I think the verify failure is legitimate?  It appears to be having trouble with the `translation/OWNERS` removal.

(bindata is picking up way too much, it shouldn't need OWNERS, but that's orthogonal to this)

dims:
@spiffxp thanks. fixed.

/assign @liggitt

translations/OWNERS
```diff
@@ -1,6 +0,0 @@
-# See the OWNERS docs at https://go.k8s.io/owners
-
-reviewers:
```
liggitt, line 3: needs rebasing on the "add sigs to root folders" commit, I think, and then bindata regenerated

liggitt:
I wasn't sure if https://github.com/kubernetes/kubernetes/issues/93637#issuecomment-668383474 was referring to moving root approver to emeritus or dropping from review lists everywhere. once that is clarified and this is rebased, lgtm

dims:
@liggitt @brendanburns updated this to the bare minimum which was the root OWNERS switch. PTAL.

OWNERS
```diff
@@ -10,14 +10,14 @@ filters:
       - thockin
       - liggitt
     approvers:
-      - brendandburns
```
BenTheElder, line 13: should we have updated reviewers above as well?

thockin:
Probably.  I don't think Brendan is active at this level any more.

PR #92874, opened 2020-07-07
Fix a few places where source field is empty when creating events via events/v1

chelseychen:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:

New events API conversion is missing AddFieldLabelConversionsForEvent (#92870). Also it lacks conversion between `Source.Component` and `ReportingController` fields.

Since `DeprecatedSource` field assignment was removed in #92662, which means `Source` field becomes empty when converting to internal, which would cause e2e tests fail when listing events with `source` fieldSelector.

So `source` fieldSelector should be set with `reportingController` when `source` is empty.

This PR also resolves the aggregation issue in `EventAggregatorMessageFunc` (#92882).

**Which issue(s) this PR fixes**:

Fixes #92870 
Fixes #92882 
Fixes #92795 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs
- [KEP]: <https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/383-new-event-api-ga-graduation>
```

PR #93687, opened 2020-08-04
kubectl describe pod: use ReportingController as an event source

ingvagabund:
**What type of PR is this?**
/kind bug

With https://github.com/kubernetes/kubernetes/pull/91645 merged (promotion of v1beta1 events to v1),
it's no longer permitted to set Source field.
Based on https://stupefied-goodall-e282f7.netlify.app/contributors/design-proposals/instrumentation/events-redesign/#backward-compatibility
Source.Component is supposed to be replaced by ReportingController.
Source.Host is supposed to be replaced by DeprecatedHost, though DeprecatedHost
was never implemented. Thus, ignoring the second field.

Fixes #

Following changes in https://github.com/kubernetes/kubernetes/pull/92874/files#diff-756d0996ce76e743d8812e2afecf746dR107-R110

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/kubectl/pkg/describe/describe.go
+++ b/staging/src/k8s.io/kubectl/pkg/describe/describe.go
@@ -3873,11 +3873,15 @@ func DescribeEvents(el *corev1.EventList, w PrefixWriter) {
 				interval = translateMicroTimestampSince(e.EventTime)
 			}
 		}
+		source := e.Source.Component
+		if source == "" {
+			source = e.ReportingController
+		}
 		w.Write(LEVEL_1, "%v\t%v\t%s\t%v\t%v\n",
 			e.Type,
 			e.Reason,
 			interval,
-			formatEventSource(e.Source),
+			source,
 			strings.TrimSpace(e.Message),
 		)
 	}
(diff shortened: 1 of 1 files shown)


staging/src/k8s.io/kubectl/pkg/describe/describe.go
```diff
@@ -5001,15 +5005,6 @@ func translateTimestampSince(timestamp metav1.Time) string {
 	return duration.HumanDuration(time.Since(timestamp.Time))
 }
 
-// formatEventSource formats EventSource as a comma separated string excluding Host when empty
-func formatEventSource(es corev1.EventSource) string {
-	EventSourceString := []string{es.Component}
-	if len(es.Host) > 0 {
```
zhouya0, line 5007: Omit `EventSource.Host` in such way, should we care backward compatibility?

ingvagabund: Host is deprecated in 1.18 or less, can't be set in 1.19. How far back in the past kubectl compatibility goes? 2 or 3 versions back?

ingvagabund:
@zhouya0 @brianpursley PTAL

staging/src/k8s.io/kubectl/pkg/describe/describe.go
soltysh: Given this is targeting 1.20, and you can't set that in 1.19 this ok change.

soltysh: We support only +/- 1 version

soltysh (approved):
/lgtm
/approve
/milestone v1.20-phase-bug

Issue #93691, opened 2020-08-04
fix for https://github.com/kubernetes/kubernetes/issues/68211 not working on 1.18.1 with containerd


cbf123:
**What happened**:
I patched back https://github.com/kubernetes/kubernetes/pull/89629/commits/6c2562a579b2356c54898d262ec578d5fded1e24 to 1.18.1 and while the code applies cleanly the meaning of the variables doesn't correspond directly.

I used dlv to look at the values of mntInfo.Root and subpath.Path at the point in the code corresponding to line 117 of  pkg/volume/util/subpath/subpath_linux.go (line number as seen in the above patch) and I found that mntInfo.Root was a substring of subpath.Path rather than being identical.
```
(dlv) print mntInfo.Root
"/pods/c08168ae-9579-4c9f-a89d-6217337d666b/volumes/kubernetes.io~configmap/setup-script/..2020_07_28_17_28_48.155435548/metricbeat.yml"
(dlv) print subpath.Path
"/var/lib/kubelet/pods/c08168ae-9579-4c9f-a89d-6217337d666b/volumes/kubernetes.io~configmap/setup-script/..2020_07_28_17_28_48.155435548/metricbeat.yml"
(dlv) print subpath
k8s.io/kubernetes/pkg/volume/util/subpath.Subpath {
        VolumeMountIndex: 2,
        Path: "/var/lib/kubelet/pods/c08168ae-9579-4c9f-a89d-6217337d666b/volumes/kubernetes.io~configmap/setup-script/..2020_07_28_17_28_48.155435548/metricbeat.yml",
        VolumeName: "setup-script",
        VolumePath: "/var/lib/kubelet/pods/c08168ae-9579-4c9f-a89d-6217337d666b/volumes/kubernetes.io~configmap/setup-script",
        PodDir: "/var/lib/kubelet/pods/c08168ae-9579-4c9f-a89d-6217337d666b",
        ContainerName: "c-beat-setup",}
(dlv) print mntInfo
k8s.io/kubernetes/vendor/k8s.io/utils/mount.MountInfo {
        ID: 802,
        ParentID: 50,
        Major: 253,
        Minor: 6,
        Root: "/pods/c08168ae-9579-4c9f-a89d-6217337d666b/volumes/kubernetes.io~configmap/setup-script/..2020_07_28_17_28_48.155435548/metricbeat.yml",
        Source: "/dev/mapper/cgts--vg-kubelet--lv",
[...]

PR #93273, opened 2020-07-21
Automated cherry pick of #93272: fix 68211: modified subpath configmap mount fails when

msau42:
Cherry pick of #93272 on release-1.17.

#93272: fix 68211: modified subpath configmap mount fails when

For details on the cherry pick process, see the [cherry pick requests](https://git.k8s.io/community/contributors/devel/sig-release/cherry-picks.md) page.

```release-note
NONE
```

PR #93696, opened 2020-08-04
Revert "Automated cherry pick of #93272: fix 68211: modified subpath configmap mount fails when"

msau42:
Reverts kubernetes/kubernetes#93273

Problem with containerd identified in https://github.com/kubernetes/kubernetes/issues/93691

```release-note
None
```


tpepper:
/kind bug
Since the originals had a release note, should the reverts have a release note to un-note the change, or do we delete the release note in the other merged PRs so they aren't picked up by tooling?

hasheddan (approved):
/lgtm

PR #93698, opened 2020-08-04
pkg/proxy: remove duplicate conntrack rule

dustinspecker:
**What type of PR is this?**
/kind cleanup

After the refactor in https://github.com/kubernetes/kubernetes/commit/8c6956e5bb4f9557e6ba459f1f6bbcb80e3b5b80, the ipvs and iptables proxier added two
identical rules other than the comments were different.

This commit consolidates the duplicate rules and updates the comment to
not specify source/destination.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/proxy/ipvs/proxier.go
+++ b/pkg/proxy/ipvs/proxier.go
@@ -1812,19 +1812,12 @@ func (proxier *Proxier) writeIptablesRules() {
 		"-j", "ACCEPT",
 	)
 
-	// The following two rules ensure the traffic after the initial packet
+	// The following rule ensures the traffic after the initial packet
 	// accepted by the "kubernetes forwarding rules" rule above will be
 	// accepted.
 	utilproxy.WriteLine(proxier.filterRules,
 		"-A", string(KubeForwardChain),
-		"-m", "comment", "--comment", `"kubernetes forwarding conntrack pod source rule"`,
-		"-m", "conntrack",
-		"--ctstate", "RELATED,ESTABLISHED",
-		"-j", "ACCEPT",
-	)
-	utilproxy.WriteLine(proxier.filterRules,
-		"-A", string(KubeForwardChain),
-		"-m", "comment", "--comment", `"kubernetes forwarding conntrack pod destination rule"`,
+		"-m", "comment", "--comment", `"kubernetes forwarding conntrack pod rule"`,
 		"-m", "conntrack",
 		"--ctstate", "RELATED,ESTABLISHED",
 		"-j", "ACCEPT",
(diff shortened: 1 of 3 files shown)


pkg/proxy/ipvs/proxier.go
```diff
@@ -1780,19 +1780,12 @@ func (proxier *Proxier) writeIptablesRules() {
 		"-j", "ACCEPT",
 	)
 
-	// The following two rules ensure the traffic after the initial packet
```
andrewsykim, line 1783: This comment implies to me that at some point it was intentional to have two rules heres but we mistakenly made them the same.

andrewsykim: Interestingly in https://github.com/kubernetes/kubernetes/commit/8c6956e5bb4f9557e6ba459f1f6bbcb80e3b5b80#diff-d51765b83fe795b469e8a86276b12dc9L1517 we removed the cluster cidr check but didn't add a local detector check. 

@satyasm any idea why this is the case here?

aojea:
/ok-to-test
despite we should understand the reason why we ended with duplicates rules, they are really duplicates :-)
nice catch

dustinspecker:
> /ok-to-test
> ~despite we should understand the reason why we ended with duplicates rules, they are really duplicates :-)~
> I see, before they were using `-s` and `-d` to differentiate by source or destination, however, the local detector removed those options making the rules equal
> nice catch

Sorry about that, I should have mentioned the removal of `-s` and `-d` in the commit message.

danwinship:
agree that this is a correct refactoring of the current code, but I have the same question Andrew did earlier:

> Interestingly in 8c6956e#diff-d51765b83fe795b469e8a86276b12dc9L1517 we removed the cluster cidr check but didn't add a local detector check.
> 
> @satyasm any idea why this is the case here?

aojea:
/assign @satyasm 
It seems that the part of information is key for the review

pkg/proxy/ipvs/proxier.go
satyasm: we didn't add the local detector check because the intent would not be the same. The original code seemed to be an optimization to accept previously accepted traffic if it is bound for the cluster. Since node CIDR does not represent the whole cluster, as discussed in the KEP (https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20191104-iptables-no-cluster-cidr.md#iptables---accepting-traffic-after-first-packet-after-being-accepted-by-kubernetes-rules), it was felt better to just remove it.

PR #93701, opened 2020-08-05
Use “comma ok” idiom to get value from map

nayihz:
**What type of PR is this?**

/kind cleanup

Use “comma ok” idiom is a common way to judge if one key exist in map.

Use “comma ok” idiom to get value from map

### DIFF
--- a/pkg/registry/core/service/portallocator/controller/repair.go
+++ b/pkg/registry/core/service/portallocator/controller/repair.go
@@ -215,8 +215,8 @@ func collectServiceNodePorts(service *corev1.Service) []int {
 			continue
 		}
 		proto := string(port.Protocol)
-		s := seen[nodePort]
-		if s == nil { // have not seen this nodePort before
+		s, ok := seen[nodePort]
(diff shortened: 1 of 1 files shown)


lavalamp:
I think the existing code is safer, in that everything works OK if somehow a nil value gets stored in the map. It looks like that wouldn't be intentional, but without analyzing every map access it's hard to be certain that's not happening.

latteczy:
> I think the existing code is safer, in that everything works OK if somehow a nil value gets stored in the map. It looks like that wouldn't be intentional, but without analyzing every map access it's hard to be certain that's not happening.

I check the logic and find that the value in map will never be set to nil. @lavalamp 

https://github.com/kubernetes/kubernetes/blob/5832c44dbf06b147739a5f7e9179e4a022de9951/pkg/registry/core/service/portallocator/controller/repair.go#L218-L227

lavalamp:
Even if it's fine now, the code isn't static, and people might add bugs.
It's better to be defensive.

latteczy:
OK. I will close this pr.

PR #71663, opened 2018-12-03
kubeadm: properly umount dirs in /var/lib/kubelet

bart0sh:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

'kubeadm reset' uses incorrect way of unmounting /var/lib/kubelet
directories. It queries /proc/mounts for /var/lib/kubelet mount point.
If /var/lib/kubelet directory is also mounted it makes 'kubelet reset'
to unmount it too, which is incorrect. It also makes it fail as it
can't unmount /var/lib/kubelet before unmounting mounts inside it.

Fixed by querying /var/lib/kubelet/ instead of /var/lib/kubelet.
This should exclude /var/lib/kubelet from the query results even if
it's mounted.

**Which issue(s) this PR fixes**:

Fixes: kubernetes/kubeadm#1294

**Does this PR introduce a user-facing change?**:
```release-note
kubeadm reset correcty unmounts mount points inside /var/lib/kubelet
```

PR #93702, opened 2020-08-05
kubeadm: reset don't unmount /var/lib/kubelet if it is mounted

### LINKED ISSUE #2247: Regression for kubeadm reset --force (unmount /var/lib/kubelet/ directory)
## Is this a BUG REPORT or FEATURE REQUEST?

BUG REPORT

## Versions

**kubeadm version** (use `kubeadm version`): `kubeadm version: &version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-15T16:56:34Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}`

**Environment**:
- **Kubernetes version** (use `kubectl version`): `Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-15T16:58:53Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}`
- **Cloud provider or hardware configuration**: Baremetal
- **OS** (e.g. from /etc/os-release): OL 7.6
- **Kernel** (e.g. `uname -a`): 4.14.35-1818.3.3.el7uek.x86_64
- **Others**:

## What happened?
Running `kubeadm reset --force` unmount `/var/lib/kubelet` directory.
This happens previously and has been fixed by the following PR: https://github.com/kubernetes/kubernetes/pull/71663
But seems like there is a change in the reset code path since 1.15

## What you expected to happen?
Mounted `/var/lib/kubelet` shouldn't be unmounted when running `kubeadm reset --force`

## How to reproduce it (as minimally and precisely as possible)?
1. Mount /var/lib/kubelet on a block device
2. kubeadm reset --force
3. The block device will be unmounted 

```
# kubeadm version
kubeadm version: &version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-15T16:56:34Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
# mount | grep kubelet
/dev/sdc1 on /var/lib/kubelet type ext4 (rw,relatime,seclabel,data=ordered)
# kubeadm reset --force
[preflight] Running pre-flight checks
W0804 17:04:20.446650   25859 removeetcdmember.go:79] [reset] No kubeadm config, using etcd pod spec to get data directory
[reset] No etcd config found. Assuming external etcd
[reset] Please, manually reset etcd to prevent further issues
[reset] Stopping the kubelet service
[reset] Unmounting mounted directories in "/var/lib/kubelet"
[reset] Deleting contents of config directories: [/etc/kubernetes/manifests /etc/kubernetes/pki]
[reset] Deleting files: [/etc/kubernetes/admin.conf /etc/kubernetes/kubelet.conf /etc/kubernetes/bootstrap-kubelet.conf /etc/kubernetes/controller-manager.conf /etc/kubernetes/scheduler.conf]
[reset] Deleting contents of stateful directories: [/var/lib/kubelet /var/lib/dockershim /var/run/kubernetes /var/lib/cni]

The reset process does not clean CNI configuration. To do so, you must remove /etc/cni/net.d

The reset process does not reset or clean up iptables rules or IPVS tables.
If you wish to reset iptables, you must do so manually by using the "iptables" command.

If your cluster was setup to utilize IPVS, run ipvsadm --clear (or similar)
to reset your system's IPVS tables.

The reset process does not clean your kubeconfig files and you must remove them manually.
Please, check the contents of the $HOME/.kube/config file.
# mount | grep kubelet
#
```

thtanaka:
Previous PR that fixes this
https://github.com/kubernetes/kubernetes/pull/71663

Signed-off-by: Thomas Tanaka <thomas.tanaka@gmail.com>

**What type of PR is this?**
/kind bug

To fix regression in kubeadm reset behavior since 1.15

Fixes https://github.com/kubernetes/kubeadm/issues/2247

```release-note
kubeadm: ensure "kubeadm reset" does not unmount the root "/var/lib/kubelet" directory if it is mounted by the user
```

```docs

```

### DIFF
--- a/cmd/kubeadm/app/cmd/phases/reset/unmount_linux.go
+++ b/cmd/kubeadm/app/cmd/phases/reset/unmount_linux.go
@@ -32,6 +32,12 @@ func unmountKubeletDirectory(absoluteKubeletRunDirectory string) error {
 	if err != nil {
 		return err
 	}
+
+	if !strings.HasSuffix(absoluteKubeletRunDirectory, "/") {
+		// trailing "/" is needed to ensure that possibly mounted /var/lib/kubelet is skipped
+		absoluteKubeletRunDirectory += "/"
+	}
+
 	mounts := strings.Split(string(raw), "\n")
 	for _, mount := range mounts {
 		m := strings.Split(mount, " ")


cmd/kubeadm/app/cmd/phases/reset/unmount_linux.go
```diff
@@ -32,6 +32,11 @@ func unmountKubeletDirectory(absoluteKubeletRunDirectory string) error {
 	if err != nil {
 		return err
 	}
+
+	if !strings.HasSuffix(absoluteKubeletRunDirectory, "/") {
+		absoluteKubeletRunDirectory += "/"
+	}
+
```
neolit123, line 39: i'm trying to understand what happens here.

this call to EvalSymlinks results in a path that does not end in /
https://github.com/kubernetes/kubernetes/blob/323f34858de18b862d43c40b2cced65ad8e24052/cmd/kubeadm/app/cmd/phases/reset/cleanupnode.go#L97

when iterating the lines in /proc/mounts we find an entry that is `device /var/lib/kubelet`, thus it ends up being unmounted as it is not skipped at this line:

```
if len(m) < 2 || !strings.HasPrefix(m[1], absoluteKubeletRunDirectory) {
```

your change make the line if condition above skip the block device mount of /var/lib/kubelet?

neolit123:
@thtanaka 

in the PR description please change this line:
> Fixes # kubernetes/kubeadm#2247

to 
> Fixes kubernetes/kubeadm#2247

and the release note to be:
> kubeadm: ensure "kubeadm reset" does not unmount the root "/var/lib/kubelet" directory if it is mounted by the user

cmd/kubeadm/app/cmd/phases/reset/unmount_linux.go
thtanaka: Thanks for review, please correct me if I am wrong here, this was an attempt to replicate the following fixes https://github.com/kubernetes/kubernetes/pull/71663/files

Here is the sample test case from the above code snippets:
```
# mount | grep kubelet
/dev/sdc1 on /var/lib/kubelet type ext4 (rw,relatime,seclabel,data=ordered)
# awk '$2 ~ path {print $2}' path=/var/lib/kubelet/ /proc/mounts <== PR #71663 fixes this by adding a "/"
# awk '$2 ~ path {print $2}' path=/var/lib/kubelet /proc/mounts
/var/lib/kubelet
```
So in short, yes it will skip the `/var/lib/kubelet` directory during the loop

neolit123: thanks for the confirmation.

rosti: > i'm trying to understand what happens here.
> 
> this call to EvalSymlinks results in a path that does not end in /

Let's say that it's not guaranteed (especially if `kubeadmconstants.KubeletRunDirectory` is not a symlink). We shouldn't rely on `EvalSymlinks` to return a result with a trailing slash.

> https://github.com/kubernetes/kubernetes/blob/323f34858de18b862d43c40b2cced65ad8e24052/cmd/kubeadm/app/cmd/phases/reset/cleanupnode.go#L97
> 
> when iterating the lines in /proc/mounts we find an entry that is `device /var/lib/kubelet`, thus it ends up being unmounted as it is not skipped here:
> 
> ```
> if len(m) < 2 || !strings.HasPrefix(m[1], absoluteKubeletRunDirectory) {
> ```
> 
> your change makes the if condition above skip the block device mount of /var/lib/kubelet?

Yes. The paths in `/proc/mounts` don't have trailing slashes so adding an extra slash to `absoluteKubeletRunDirectory` would save the day.

cmd/kubeadm/app/cmd/phases/reset/unmount_linux.go
```diff
@@ -32,6 +32,11 @@ func unmountKubeletDirectory(absoluteKubeletRunDirectory string) error {
 	if err != nil {
 		return err
 	}
+
```
rosti, line 35: @thtanaka can we add a comment above the `if` statement as to why this is necessary for posterity?

rosti (commented):
Thanks @thtanaka !
Looks great although we need to add a comment to document the need of a trailing slash.

neolit123 (commented):
/lgtm
/approve

the change seems valid.
the PR can merge once 1.19 code freeze ends.

given this is not a critical fix, older versions should rely on a workaround to potentially re-mount after kubeadm reset.

cmd/kubeadm/app/cmd/phases/reset/unmount_linux.go
bart0sh: @rosti @thtanaka @neolit123 

> We shouldn't rely on EvalSymlinks to return a result with a trailing slash.

According to the [official golang documentation](https://golang.org/pkg/path/filepath/#EvalSymlinks) EvalSymlinks calls filepath.Clean on the result.
[Clean docs state that](https://golang.org/pkg/path/filepath/#Clean) ```The returned path ends in a slash only if it represents a root directory```

So, to my understanding, we can rely on that.

bart0sh (changes requested):

neolit123:
> The returned path ends in a slash only if it represents a root directory

I think this means there will be a trailing slash only if the path is "/", but any sub directory will not have it.

The PR is good. It can merge once master opens for bug fixes.

bart0sh:
@neolit123 

> I think this means there will be a trailing slash only if the path is "/", but any sub directory will not have it.

exactly. That means that ```if !strings.HasSuffix(absoluteKubeletRunDirectory, "/")``` check is not required unless absoluteKubeletRunDirectory is  '/'

bart0sh:
btw, [previous solution](https://github.com/kubernetes/kubernetes/pull/71663/commits/48a961cfa2f604a3d83bf8bc1a2e04c43644bd64) added leading '/' unconditionally and worked just fine until it was removed.

neolit123:
> btw, previous solution
<https://github.com/kubernetes/kubernetes/pull/71663/commits/48a961cfa2f604a3d83bf8bc1a2e04c43644bd64>
added
leading '/' unconditionally and worked just fine until it was removed.

Yes, we lost it during refactor.

bart0sh:
Just to be clear here. I'd propose to do ```absoluteKubeletRunDirectory += "/"'``` unconditionally.

neolit123:
Conditional is better because if the path is configurable one day it can end up with double slash, and the unmount will be always skipped.

bart0sh:
unconditional is better because filepath.Clear would remove all double slashes and other things that prevent matching mount path.

neolit123:
Hm where is Clear being called after the append? I am on mobile and cannot find it.

bart0sh:
It's called from EvalSymlinks and guarantees that the result doesn't have double slashes and trailing slash.

neolit123:
Ok, understood. We should have a comment about this over the append.

neolit123:
(back on desktop)

> unconditional is better because filepath.Clear would remove all double slashes and other things that prevent matching mount path.

so, this is not ideal because it means the caller of  unmountKubeletDirectory() has to call filepath.Clear or trim the trailing `/` as a pre-requisite. conditionally adding the `/` makes unmountKubeletDirectory() handles this on the side of the utility function.

bart0sh:
current solution is even less ideal as it hides call of unmountKubeletDirectory inside absoluteKubeletRunDirectory, so the caller of absoluteKubeletRunDirectory is not aware that it actually does some unmounts.

bart0sh:
so, moving filepath.EvalSymlinks call inside unmount_linux.go:unmountKubeletDirectory would eliminate need in absoluteKubeletRunDirectory, make the code cleaner and would make filepath.EvalSymlinks called only on Linux, where it actually needs to be called.

bart0sh:
so, moving filepath.EvalSymlinks call inside unmount_linux.go:unmountKubeletDirectory would eliminate need in absoluteKubeletRunDirectory, make the code cleaner and would make filepath.EvalSymlinks called only on Linux, where it actually needs to be called.

justaugustus:

/retest

PR #93703, opened 2020-08-05
PV Controller: add test cases that PV is retained and released

cofyc:
**What type of PR is this?**
/kind cleanup

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/controller/volume/persistentvolume/binder_test.go
+++ b/pkg/controller/volume/persistentvolume/binder_test.go
@@ -362,6 +362,38 @@ func TestSync(t *testing.T) {
 			newClaimArray("claim2-10", "uid2-10", "1Gi", "volume2-10", v1.ClaimPending, nil),
 			[]string{"Warning VolumeMismatch"}, noerrors, testSyncClaim,
 		},
+		{
+			// syncClaim with claim pre-bound to a released PV that is pre-bound to the
+			// claim by UID. Check it gets bound and no pvutil.AnnBoundByController is
+			// set.
+			"2-11 - claim prebound to released and prebound volume by UID",
+			newVolumeArray("volume2-5", "1Gi", "uid2-5", "claim2-5", v1.VolumeReleased, v1.PersistentVolumeReclaimRetain, classEmpty),
+			newVolumeArray("volume2-5", "1Gi", "uid2-5", "claim2-5", v1.VolumeBound, v1.PersistentVolumeReclaimRetain, classEmpty),
+			newClaimArray("claim2-5", "uid2-5", "1Gi", "volume2-5", v1.ClaimPending, nil),
+			newClaimArray("claim2-5", "uid2-5", "1Gi", "volume2-5", v1.ClaimBound, nil, pvutil.AnnBindCompleted),
+			noevents, noerrors, testSyncClaim,
+		},
+		{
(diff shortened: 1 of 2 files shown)


pkg/controller/volume/persistentvolume/pv_controller.go
```diff
@@ -409,7 +409,7 @@ func (ctrl *PersistentVolumeController) syncUnboundClaim(claim *v1.PersistentVol
 			klog.V(4).Infof("synchronizing unbound PersistentVolumeClaim[%s]: volume %q requested and found: %s", claimToClaimKey(claim), claim.Spec.VolumeName, getVolumeStatusForLogging(volume))
 			if volume.Spec.ClaimRef == nil {
 				// User asked for a PV that is not claimed
-				// OBSERVATION: pvc is "Pending", pv is "Available"
+				// OBSERVATION: pvc is "Pending"
```
cofyc, line 412: PV can be in any state. Typically, PV is in released state after the PVC is deleted.

pkg/controller/volume/persistentvolume/binder_test.go
```diff
@@ -362,6 +362,38 @@ func TestSync(t *testing.T) {
...
+			newClaimArray("claim2-5", "uid2-5", "1Gi", "volume2-5", v1.ClaimPending, nil),
+			newClaimArray("claim2-5", "uid2-5", "1Gi", "volume2-5", v1.ClaimBound, nil, pvutil.AnnBindCompleted),
+			noevents, noerrors, testSyncClaim,
+		},
+		{
+			// syncClaim with claim pre-bound to a released PV that is pre-bound to the
+			// claim by name. Check it gets bound and no pvutil.AnnBoundByController is
+			// set.
+			"2-12 - claim prebound to released and prebound volume by name",
+			newVolumeArray("volume2-5", "1Gi", "", "claim2-5", v1.VolumeReleased, v1.PersistentVolumeReclaimRetain, classEmpty),
```
jsafrane, line 381: This enters a gray area: under normal circumstances, a volume can't enter Released phase if it has no PVC UID. I know that PV controller ignores the Phase, so the volume looks Pending/Available and it binds the volume to PVC, however, I am not sure we should codify this behavior in unit tests. Does it cover any code path not covered by previous unit tests?

cofyc: what about this scenario?

- PV enters into Released state when associated PVC is deleted
-  User pre-binds the PV to the PVC `claim2-5` by setting `{Name: "claim2-5"}` to `claimRef` field
  - no UID, but PV is in the Released state
- Users create a PVC named `claim2-5`

cofyc: > Does it cover any code path not covered by previous unit tests?

This scenario is meaningful IMO. But I agree it does not cover any new code path. Because the PV controller ignores the phase, it's already covered by `"2-4 - claim prebound to prebound volume by name"` except the PV state is Available.

jsafrane: > old claimRef is removed, so no UID, but PV is in the Released state

This is possible, but PV controller will fix the status to Available in the next sync.

Issue #69010, opened 2018-09-24
Deprecate Node.Spec.Unschedulable when TaintNodeByCondition is GA'ed


bsalamat:
TaintNodeByCondition has replaced the existing node condition. With this feature node conditions that affect feasibility of nodes are reflected as taints on nodes. This feature is promoted to Beta in 1.12. We should deprecate node conditions once the feature is GA'ed in future releases.

- [ ] Announce deprecation of Node.Spec.Unschedulable
- [ ] Mark Node.Spec.Unschedulable as deprecated in our API
- [ ] Update our documents to note that Node.Spec.Unschedulable is deprecated

/sig scheduling
/sig node

tianshapjq:
hi~ I am interested in this job, and commit a pr to give it a try. Would you please take a look to check if I do it right? :)

bsalamat:
@tianshapjq Thanks for the PR, but as it is mentioned in the issue, we should deprecate Node.Spec.Unschedulable only when TaintNodeByCondition is GA. TaintNodeByCondition is beta in 1.12 and will probably not be in GA in 1.13. So, there will be some time before we can deprecate Node.Spec.Unschedulable.

draveness:
We can't deprecate the Unschedulable field unless we have v2 node API, per discussion in https://github.com/kubernetes/kubernetes/pull/84266#issuecomment-545951881. We'll revisit this issue after that

/lifecycle frozen

draveness:
/close

feel free to reopen this

PR #68984, opened 2018-09-23
Keep backward compatibility for 'node.Spec.Unschedulable'.

k82cn:
Signed-off-by: Da K. Ma <klaus1982.cn@gmail.com>

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #68899 

**Release note**:
```release-note
Fixes issue [#68899](https://github.com/kubernetes/kubernetes/issues/68899) where pods might schedule on an unschedulable node.
```

PR #93706, opened 2020-08-05
Fix an error in NodeUnschedulable plugin comment

houminz:
**What type of PR is this?**
/kind bug

It corrected an error in the comment of NodeUnschedulable Plugin struct.
"a plugin that priorities nodes according to the node annotation scheduler.alpha.kubernetes.io/preferAvoidPods"
->" a plugin that checks if a pod can be scheduled on a node with Unschedulable spec."

Fixes #

None

No

No

### DIFF
--- a/pkg/scheduler/framework/plugins/nodeunschedulable/node_unschedulable.go
+++ b/pkg/scheduler/framework/plugins/nodeunschedulable/node_unschedulable.go
@@ -25,8 +25,8 @@ import (
 	framework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
 )
 
-// NodeUnschedulable is a plugin that priorities nodes according to the node annotation
-// "scheduler.alpha.kubernetes.io/preferAvoidPods".
+// NodeUnschedulable plugin filters nodes that set node.Spec.Unschedulable=true unless
+// the pod tolerates {key=node.kubernetes.io/unschedulable, effect:NoSchedule} taint.
 type NodeUnschedulable struct {
 }
 


pkg/scheduler/framework/plugins/nodeunschedulable/node_unschedulable.go
```diff
@@ -25,8 +25,7 @@ import (
 	framework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
 )
 
-// NodeUnschedulable is a plugin that priorities nodes according to the node annotation
-// "scheduler.alpha.kubernetes.io/preferAvoidPods".
+// NodeUnschedulable is a plugin that checks if a pod can be scheduled on a node with Unschedulable spec.
```
hex108, line 28:
```suggestion
// NodeUnschedulable is a plugin that checks if a pod can be scheduled on a unschedulable node according to its tolerations.
```

hex108 (commented):
Thanks!

pkg/scheduler/framework/plugins/nodeunschedulable/node_unschedulable.go
SimpCosm: Hi, thanks for your suggestion, I have updated the commit

hex108:
/ok-to-test

Please squash the commits into one commit.

pkg/scheduler/framework/plugins/nodeunschedulable/node_unschedulable.go
```diff
@@ -25,8 +25,8 @@ import (
 	framework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
 )
 
-// NodeUnschedulable is a plugin that priorities nodes according to the node annotation
-// "scheduler.alpha.kubernetes.io/preferAvoidPods".
+// NodeUnschedulable is a plugin that checks if a pod can be scheduled on a
```
adtac, line 28: nit: a -> an :)

pkg/scheduler/framework/plugins/nodeunschedulable/node_unschedulable.go
```diff
@@ -25,8 +25,8 @@ import (
...
+// unschedulable node according to its tolerations.
```
adtac, line 29: I think "NodeUnschedulable is a plugin that filters pods from being scheduled on nodes with the `node.kubernetes.io/unschedulable` taint." would be a better description as it describes which specific taint would affect this plugin and what the effect is. What do the others think?

SimpCosm: Yes, according to the code, if the node with the `node.kubernetes.io/unschedulable` taint set to `NoSchedule` and the pod does not tolerate the unschedulable taint, the pod would not be scheduled to the node.

By the way, according to the comment @k82cn 
> TODO (k82cn): deprecates `node.Spec.Unschedulable` in 1.13.
`node.Spec.Unschedulable` would have been deprecated in 1.13. I am wondering if we still need to check node.Spec.Unschedulable here.
```go
   if nodeInfo.Node().Spec.Unschedulable && !podToleratesUnschedulable {
```

pkg/scheduler/framework/plugins/nodeunschedulable/node_unschedulable.go
SimpCosm: Thanks for your reminding, I would change that after we agree on the comment

pkg/scheduler/framework/plugins/nodeunschedulable/node_unschedulable.go
damemi: `node.Spec.Unschedulable` is still used in places like `kubectl cordon`: https://github.com/kubernetes/kubernetes/blob/aa0632208ec0a2191e956d86f3406afbb497b881/staging/src/k8s.io/kubectl/pkg/drain/cordon.go#L83

SimpCosm: Hi @damemi , thanks for your reply. For this PR, I have updated the comment as suggested by @adtac . Do we need to deprecate `node.Spec.Unschedulable` in another PR ? Maybe I can help with that.

ahg-g: > "NodeUnschedulable is a plugin that filters pods from being scheduled on nodes with the node.kubernetes.io/unschedulable taint set to NoSchedule."

This is not accurate, I think it should be: 

"NodeUnschedulable plugin filters nodes that set node.Spec.Unschedulable=true unless the pod tolerates `{key=node.kubernetes.io/unschedulable, effect:NoSchedule}` taint."

chendave: > Hi @damemi , thanks for your reply. For this PR, I have updated the comment as suggested by @adtac . Do we need to deprecate `node.Spec.Unschedulable` in another PR ? Maybe I can help with that.

I don't know the history, but I think this is why the filter is here, allow the pod to schedule on the pod which is tagged with `node.kubernetes.io/unschedulable` doesn't make sense to me, unless the case @damemi  mentioned, if `node.Spec.Unschedulable` is deprecated we need to figure out alternatives.

SimpCosm: As mentioned in #69010 , `node.Spec.Unschedulable ` was supposed to be deprecated, but would not be removed until the API v2. To keep backward compatibility [#68984](https://github.com/kubernetes/kubernetes/pull/68984) , we would filter the nodes that set node.Spec.Unschedulable=true unless the pod tolerates {key=node.kubernetes.io/unschedulable, effect:NoSchedule} taint.

Maybe we can fix the comment error here and when `node.Spec.Unschedulable ` is deprecated, we could keep working on that.

PR #93707, opened 2020-08-05
Fix issue in evaluating symlink path for subpath

jingxu97:
After removing driver letter assignment during disk partition and formatting, we found an issue that golang filepath.EvalSymlnik no longer works. See bug https://github.com/golang/go/issues/39786

We had used a simple powershell command (get-item -path %p).Target to get the symlink, but it does not address all the cases. If there is a subpath under symlink, this command does not resolve upper path symlink. During evaluating symlink, we need to make sure the returned target path does not contain any symlink.

This PR tries to recursively resolve symlink in the path until it checking all intermediate dirs. 

**What type of PR is this?**

Fixes #

```release-note
NONE

```

```docs

```

### DIFF
--- a/pkg/volume/util/subpath/subpath_windows.go
+++ b/pkg/volume/util/subpath/subpath_windows.go
@@ -31,6 +31,10 @@ import (
 	"k8s.io/utils/nsenter"
 )
 
+// MaxPathLength is the maximum length of Windows path. Normally, it is 260, but if long path is enable,
+// the max number is 32,767
+const MaxPathLength = 32767
+
 type subpath struct{}
 
 // New returns a subpath.Interface for the current system
@@ -44,34 +48,91 @@ func NewNSEnter(mounter mount.Interface, ne *nsenter.Nsenter, rootDir string) In
 	return nil
 }
 
-// evalPath returns the path name after the evaluation of any symbolic links.
-// If the path after evaluation starts with Volume or \??\Volume, it means that it was a symlink from
-// volume (represented by volumeID) to the given path. In this case, the given path is returned.
-func evalPath(path string) (linkedPath string, err error) {
-	cmd := fmt.Sprintf("Get-Item -Path %s | Select-Object -ExpandProperty Target", path)
-	output, err := exec.Command("powershell", "/c", cmd).CombinedOutput()
-	klog.V(4).Infof("evaluate symlink from %s: %s %v", path, string(output), err)
-	if err != nil {
-		return "", err
-	}
-	linkedPath = strings.TrimSpace(string(output))
-	if linkedPath == "" {
-		klog.V(4).Infof("Path '%s' has no target. Consiering it as evaluated.", path)
-		return path, nil
-	}
-	if isVolumePrefix(linkedPath) {
-		return path, err
-	}
-	return linkedPath, err
-}
-
-// isVolumePrefix returns true if the given path name starts with "Volume" or volume prefix including
-// "\\.\" or "\\?\". Otherwise, it returns false.
-func isVolumePrefix(path string) bool {
-	if strings.HasPrefix(path, "Volume") || strings.HasPrefix(path, "\\\\?\\") || strings.HasPrefix(path, "\\\\.\\") {
+// isDriveLetterPath returns true if the given path is empty or it ends with ":" or ":\" or ":\\"
+func isDriveLetterorEmptyPath(path string) bool {
+	if path == "" || strings.HasSuffix(path, ":\\\\") || strings.HasSuffix(path, ":") || strings.HasSuffix(path, ":\\") {
 		return true
 	}
 	return false
+}
+
+// isVolumePrefix returns true if the given path name starts with "Volume" or volume prefix including
+// "\\.\", "\\?\" for device path or "UNC" or "\\" for UNC path. Otherwise, it returns false.
+func isDeviceOrUncPath(path string) bool {
+	if strings.HasPrefix(path, "Volume") || strings.HasPrefix(path, "\\\\?\\") || strings.HasPrefix(path, "\\\\.\\") || strings.HasPrefix(path, "UNC") {
+		return true
+	}
+	return false
+}
+
+// getUpperPath removes the last level of directory.
+func getUpperPath(path string) string {
+	sep := fmt.Sprintf("%c", filepath.Separator)
+	upperpath := strings.TrimSuffix(path, sep)
+	return filepath.Dir(upperpath)
+}
+
+// Check whether a directory/file is a link type or not
+// LinkType could be SymbolicLink, Junction, or HardLink
+func isLinkPath(path string) (bool, error) {
+	cmd := fmt.Sprintf("(Get-Item -Path %s).LinkType", path)
+	output, err := exec.Command("powershell", "/c", cmd).CombinedOutput()
+	if err != nil {
+		return false, err
+	}
+	if strings.TrimSpace(string(output)) != "" {
+		return true, nil
+	}
+	return false, nil
+}
+
+// evalSymlink returns the path name after the evaluation of any symbolic links.
+// If the path after evaluation is a device path or network connection, the original path is returned
+func evalSymlink(path string) (string, error) {
+	path = mount.NormalizeWindowsPath(path)
+	if isDeviceOrUncPath(path) || isDriveLetterorEmptyPath(path) {
+		klog.V(4).Infof("Path '%s' is not a symlink, return its original form.", path)
+		return path, nil
+	}
+	upperpath := path
+	base := ""
+	for i := 0; i < MaxPathLength; i++ {
+		isLink, err := isLinkPath(upperpath)
+		if err != nil {
+			return "", err
+		}
+		if isLink {
+			break
+		}
+		// continue to check next layer
+		base = filepath.Join(filepath.Base(upperpath), base)
+		upperpath = getUpperPath(upperpath)
+		if isDriveLetterorEmptyPath(upperpath) {
+			klog.V(4).Infof("Path '%s' is not a symlink, return its original form.", path)
+			return path, nil
+		}
+	}
+	// This command will give the target path of a given symlink
+	cmd := fmt.Sprintf("(Get-Item -Path %s).Target", upperpath)
+	output, err := exec.Command("powershell", "/c", cmd).CombinedOutput()
+	if err != nil {
+		return "", err
+	}
+	klog.V(4).Infof("evaluate path %s: symlink from %s to %s", path, upperpath, string(output))
+	linkedPath := strings.TrimSpace(string(output))
+	if linkedPath == "" || isDeviceOrUncPath(linkedPath) {
+		klog.V(4).Infof("Path '%s' has a target %s. Return its original form.", path, linkedPath)
+		return path, nil
+	}
+	// If the target is not an absoluate path, join iit with the current upperpath
+	if !filepath.IsAbs(linkedPath) {
+		linkedPath = filepath.Join(getUpperPath(upperpath), linkedPath)
+	}
+	nextLink, err := evalSymlink(linkedPath)
+	if err != nil {
+		return path, err
+	}
+	return filepath.Join(nextLink, base), nil
 }
 
 // check whether hostPath is within volume path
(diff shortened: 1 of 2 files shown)


jingxu97:
cc @msau42

pkg/volume/util/subpath/subpath_windows.go
```diff
@@ -62,6 +62,32 @@ func evalPath(path string) (linkedPath string, err error) {
 	if isVolumePrefix(linkedPath) {
 		return path, err
 	}
+	// The following is a workaround solution for issue https://github.com/kubernetes/kubernetes/issues/93759
+	if _, err = os.Lstat(linkedPath); err != nil {
```
msau42, line 66: Should we specifically look for not exists error?

msau42: What happens if the linkedPath really does exist because of name collision between the directory name of the subpath and a real directory in the system? For example, your subpath was "/var/lib/kubelet/..../volumes/x/programs", and "C://programs" actually does exist?

msau42 (commented):
IIUC, since Get-Item isn't returning the proper path, we check until we find a parent directory that resolves correctly, and then append the rest of the path back. But doesn't this basically bypass the symlink resolution/verification?

pkg/volume/util/subpath/subpath_windows.go
jingxu97: In that case, the following check will fail
if !mount.PathWithinBase(currentFullPath, volumePath) {
			errorResult = fmt.Errorf("SubPath %q not within volume path %q", currentFullPath, volumePath)
			break
		}
https://github.com/kubernetes/kubernetes/blob/feaaf468ed83e214dd12a7d3c4fbf5177165c405/pkg/volume/util/subpath/subpath_windows.go#L170

jingxu97:
> resolution/verification?

I think I didn't change the original flow. Basically instead of using a power shell command (which might take some time to execute), here use a few more commands to get the path. Even if user change symlink during or after this process, we only allow subpath which is under base path. The lock is implemented after we eval the subpath ?https://github.com/kubernetes/kubernetes/blob/98814409bab818d942e3c341d76eb594f12675a5/pkg/volume/util/subpath/subpath_windows.go#L99

msau42:
My understanding is that we skip evaluating the parts of the path that don't resolve correctly, which means that the path that gets returned from this function could still contain symlinks?

pkg/volume/util/subpath/subpath_windows.go
```diff
@@ -45,30 +45,59 @@ func NewNSEnter(mounter mount.Interface, ne *nsenter.Nsenter, rootDir string) In
...
+		if isDriveLetterorEmptyPath(upperpath) {
+			klog.V(4).Infof("Path '%s' is not a symlink, return its original form.", path)
+			return path, nil
+		}
+		cmd := fmt.Sprintf("(Get-Item -Path %s).Attributes", upperpath)
+		output, err := exec.Command("powershell", "/c", cmd).CombinedOutput()
+		if err != nil {
+			return "", err
+		}
+		if strings.Contains(string(output), "ReparsePoint") {
```
msau42, line 63: Can you add a comment giving an example of this string in some output? And explain what it means?

pkg/volume/util/subpath/subpath_windows.go
```diff
@@ -45,30 +45,59 @@ func NewNSEnter(mounter mount.Interface, ne *nsenter.Nsenter, rootDir string) In
...
+	if isDeviceOrUncPath(linkedPath) {
+		klog.V(4).Infof("Path '%s' is linked to device or network connection %s. Return the path.", path, linkedPath)
+		return path, nil
+	}
+	klog.V(4).Infof("Path %s is linked to %s", path, filepath.Join(linkedPath, base))
+	return filepath.Join(linkedPath, base), nil
+}
+
+// isDriveLetterPath returns true if the given path is empty or it ends with ":" or ":\" or ":\\"
+func isDriveLetterorEmptyPath(path string) bool {
```
msau42, line 90: Can we add unit tests for these?

pkg/volume/util/subpath/subpath_windows.go
jingxu97: added. PTAL

jingxu97:
@msau42 PTAL

pkg/volume/util/subpath/subpath_windows.go
jingxu97: added

pkg/volume/util/subpath/subpath_windows.go
msau42: Some docs here seem to indicate that ReparsePoint is only supported by NTFS and also there are multiple kinds of ReparsePoints that are clarified in a fs-specific tag.

https://en.wikipedia.org/wiki/NTFS_reparse_point
https://docs.microsoft.com/en-us/windows/win32/fileio/reparse-points

cc @andyzhangx to help review and see if this is the correct way to identify a symlink.

jingxu97:
@msau42 PTAL

pkg/volume/util/subpath/subpath_windows.go
```diff
@@ -45,30 +45,63 @@ func NewNSEnter(mounter mount.Interface, ne *nsenter.Nsenter, rootDir string) In
...
+		// It means that the given path is a directory and also a symlink.
+		if strings.Contains(string(output), "ReparsePoint") {
+			break
+		}
+		base = filepath.Join(filepath.Base(upperpath), base)
+		upperpath = filepath.Dir(upperpath)
+	}
+
+	// This command will give the target path of a given symlink
+	cmd := fmt.Sprintf("(Get-Item -Path %s).Target", upperpath)
```
msau42, line 74: What if you have a chain of symlinks?  For example, "a/b/c/d", and `b` and `d` are symlinks?

test/e2e/common/host_path.go
```diff
@@ -65,8 +65,12 @@ var _ = ginkgo.Describe("[sig-storage] HostPath", func() {
 	ginkgo.It("should support r/w [NodeConformance]", func() {
 		filePath := path.Join(volumePath, "test-file")
 		retryDuration := 180
+		path := "/tmp"
```
msau42, line 68: Why is this needed? Was this working before on Windows?

pkg/volume/util/subpath/subpath_windows.go
```diff
@@ -45,30 +45,63 @@ func NewNSEnter(mounter mount.Interface, ne *nsenter.Nsenter, rootDir string) In
 }
 
 // evalPath returns the path name after the evaluation of any symbolic links.
-// If the path after evaluation starts with Volume or \??\Volume, it means that it was a symlink from
-// volume (represented by volumeID) to the given path. In this case, the given path is returned.
+// If the path after evaluation is a device path or network connection, the original path is returned
 func evalPath(path string) (linkedPath string, err error) {
-	cmd := fmt.Sprintf("Get-Item -Path %s | Select-Object -ExpandProperty Target", path)
+	// The following is a workaround solution for issue https://github.com/kubernetes/kubernetes/issues/93759
```
msau42, line 50: Do we have e2e tests that test these scenarios?

jingxu97: yes, those subpath e2e tests will exercise this part of code.

test/e2e/common/host_path.go
jingxu97: good point. it was working before. I realized this is related to my change. My code cannot handle path format like this /tmp/foo.., I think I should use normalizeWindowsPath before evaluating.

pkg/volume/util/subpath/subpath_windows.go
jingxu97: https://docs.microsoft.com/en-us/windows/win32/fileio/determining-whether-a-directory-is-a-volume-mount-point
from my understanding, reparsepoint can be used to check symlink in windows.

msau42: From that page, it sounds like you also have to parse "IO_REPARSE_TAG_SYMLINK"

jingxu97: The list of attributes for a file/dir is follows. I tried all possible options with mklink to create different types of links (https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/mklink), the attribute all include ReparsePoint

ReadOnly
Hidden
System
Directory
Archive
Device
Normal
Temporary
SparseFile
ReparsePoint
Compressed
Offline
NotContentIndexed
Encrypted

pkg/volume/util/subpath/subpath_windows.go
jingxu97: I think the whole function here is first evaluate the lowest layer /d, then it loop through each layer later here https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/util/subpath/subpath_windows.go#L122.
Windows and Linux share the same logic, I think

test/e2e/common/host_path.go
jingxu97: I did a small test in Linux too, filepath.EvalSymlinks(base) returns the lowest layer symlink target

pkg/volume/util/subpath/subpath_windows.go
```diff
@@ -44,31 +44,74 @@ func NewNSEnter(mounter mount.Interface, ne *nsenter.Nsenter, rootDir string) In
...
+}
+
+func evalWindowsSymlink(path string) (string, error) {
+	if isDeviceOrUncPath(path) || isDriveLetterorEmptyPath(path) {
+		klog.V(4).Infof("Path '%s' is not a symlink, return its original form.", path)
+		return path, nil
+	}
+	upperpath := path
+	base := ""
+	for {
```
andyzhangx, line 60: since it's a for loop, it's safer to set a max depth size, suggested max depth size: `256`

pkg/volume/util/subpath/subpath_windows.go
```diff
@@ -44,31 +44,74 @@ func NewNSEnter(mounter mount.Interface, ne *nsenter.Nsenter, rootDir string) In
...
 
-// evalPath returns the path name after the evaluation of any symbolic links.
-// If the path after evaluation starts with Volume or \??\Volume, it means that it was a symlink from
-// volume (represented by volumeID) to the given path. In this case, the given path is returned.
-func evalPath(path string) (linkedPath string, err error) {
-	cmd := fmt.Sprintf("Get-Item -Path %s | Select-Object -ExpandProperty Target", path)
+// evalSymlink returns the path name after the evaluation of any symbolic links.
+// If the path after evaluation is a device path or network connection, the original path is returned
+func evalSymlink(path string) (string, error) {
+	return evalWindowsSymlink(mount.NormalizeWindowsPath(path))
```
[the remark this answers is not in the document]
jingxu97: I was thinking to make mount.NormalizeWindowsPath(path) out of the recursive function. Maybe I should put it inside actually.

pkg/volume/util/subpath/subpath_windows.go
jingxu97: Discussed offline, the function should make sure the returned path does not have any symlink in it. Use a recursive function to make sure of it.

pkg/volume/util/subpath/subpath_windows.go
jingxu97: thanks! I think both approach should work. LinkType seems a new attributed added PowerShell v5.0?

pkg/volume/util/subpath/subpath_windows.go
jingxu97: good point. found information about directory length
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation

So default maximum length is 260, but it could be extended with some requirements. The new limit is 32,767 characters. We could set it very conservatively to 32767, or some reasonable value in between?

test/e2e/common/host_path.go
```diff
@@ -97,7 +97,6 @@ var _ = ginkgo.Describe("[sig-storage] HostPath", func() {
 
 		filePathInWriter := path.Join(volumePath, fileName)
 		filePathInReader := path.Join(volumePath, subPath, fileName)
-
```
oomichi, line 100: This seems unrelated change.
If removing this change from this PR, you don't need to get approval from `area/test`.

pkg/volume/util/subpath/subpath_windows.go
```diff
@@ -31,6 +31,8 @@ import (
 	"k8s.io/utils/nsenter"
 )
 
+const MaxPathLength = 32767
```
oomichi, line 34: Could you add the comment to explain why this value 32767 is defined?

oomichi (commented):
/cc @oomichi

pkg/volume/util/subpath/subpath_windows.go
jingxu97: added it. Thanks!

jingxu97:
@msau42 @andyzhangx @oomichi I made changes based on comments. PTAL

Thanks!

andyzhangx (approved):
/lgtm

andyzhangx (approved):
/lgtm
/approve

jingxu97:
should consider to cherrypick to 1.18

PR #93709, opened 2020-08-05
Make RetryWatcher sleep interruptible

ash2k:

`time.Sleep()` is not interruptible. It should be possible to stop sleeping if the watch owner wants to abort the watch during the sleep.

```release-note
NONE
```
/sig api-machinery
/kind cleanup

### DIFF
--- a/staging/src/k8s.io/client-go/tools/watch/retrywatcher.go
+++ b/staging/src/k8s.io/client-go/tools/watch/retrywatcher.go
@@ -268,7 +268,13 @@ func (rw *RetryWatcher) receive() {
 			return
 		}
 
-		time.Sleep(retryAfter)
+		timer := time.NewTimer(retryAfter)
+		select {
+		case <-ctx.Done():
+			timer.Stop()
+			return
+		case <-timer.C:
+		}
 
 		klog.V(4).Infof("Restarting RetryWatcher at RV=%q", rw.lastResourceVersion)
 	}, rw.minRestartDelay)


ash2k:
ping @roycaihw

ash2k:
ping @deads2k

staging/src/k8s.io/client-go/tools/watch/retrywatcher.go
```diff
@@ -268,7 +268,13 @@ func (rw *RetryWatcher) receive() {
...
 		}
 
-		time.Sleep(retryAfter)
+		timer := time.NewTimer(retryAfter)
+		defer timer.Stop()
+		select {
+		case <-ctx.Done():
+			return
+		case <-timer.C:
+		}
```
sttts, line 277: select {
case <- time.After(retryAdfter):
case <-ctx.Done():
  return
}

ash2k: Yes. But the code in the PR is written like that on purpose because `time.After` creates a timer that is not freed until it expires, even if it's not needed it's not garbage collected. If we create the timer ourself, we can call `Stop` on it to free up the resources.

https://www.arangodb.com/2020/09/a-story-of-a-memory-leak-in-go-how-to-properly-use-time-after/

Hm, and this article suggests a different pattern. I'm going to look into it now...

ash2k: It's not clear whether it's necessary to drain the channel to avoid the memory leak or it's just needed to prevent receives from it after stop has been called https://github.com/golang/go/blob/master/src/time/sleep.go#L59-L65

ash2k: Ok, looks like what I'm suggesting in this PR is the right way to do it https://github.com/golang/go/issues/27169

ash2k: See also https://github.com/golang/go/issues/37196

ash2k:
Well, ideally I'd like to get https://github.com/kubernetes/utils/pull/197 merged and use `Clock` here instead. But nothing is happening with my PRs :)

ash2k:
One year and counting :)

staging/src/k8s.io/client-go/tools/watch/retrywatcher.go
```diff
@@ -268,7 +271,13 @@ func (rw *RetryWatcher) receive() {
 			return
 		}
 
-		time.Sleep(retryAfter)
+		timer := rw.clk.NewTimer(retryAfter)
```
sttts, line 274: why not `time.After`? I don't see clk used anymore (I guess it was for mocking?).

ash2k: @sttts Please see above https://github.com/kubernetes/kubernetes/pull/93709#discussion_r522432185 =)

ash2k: I don't mind removing `clk` if you think we don't need it. Just added it for "completeness" - if there will be a need to write tests for this or other things in the future. In practice it's probably not worth it trying to test this (seems like a lot of setup would be needed).

ash2k: Removed clock.

ash2k:
@sttts PTAL

ash2k:
@deads2k PTAL

PR #93710, opened 2020-08-05
Detect volume attach-ability in the middle of attaching

### LINKED ISSUE #89173: Pod startup failed when CSI driver changes from attachable to non-attachable
**What happened**:
When a CSI driver changes from attachable (`CSIDriver.spec.attachRequired`: true) to non-attachable (`attachRequired: false`) in the middle of attaching a volume, A/D controller still wants to attach the volume.

A/D controller skips attach.

1. Install CSI mock driver:
    - With `--disable-attach=true`, so the mock driver does not require `ControllerPublish`)
    - Without the external-attacher sidecar, so we can simulate slow attach.
    - Without any CSIDriver instance, so A/D controller requires attach.
2. Create a PVC and start a pod.
3. Wait for time outs when attaching the provisioned volume, because the external-attacher is not running.
4. Create CSIDriver with `attachRequired: false`.

The pod never starts. The volume is already in A/D controller desired state of world and nothing removes it from there.

Pod events:
```
Successfully assigned default/testpod to ip-10-0-160-231.us-east-2.compute.internal
AttachVolume.Attach failed for volume "pvc-1da1900a-7514-4a4a-9a65-bb54f5c42308" : attachdetachment timeout for volume 5
AttachVolume.FindAttachablePluginBySpec failed for volume "pvc-1da1900a-7514-4a4a-9a65-bb54f5c42308"
Unable to attach or mount volumes: unmounted volumes=[vol], unattached volumes=[vol default-token-2mwq6]: timed out waiting for the condition
```

**Environment**:
- Kubernetes version (use `kubectl version`): 1.17.1

/sig storage
cc @gnufied @jingxu97 @msau42

Jiawei0227:
**What type of PR is this?**
/kind bug

Previously if a CSI volume changes it's attachability when the attaching is on-going, especially from attachable to non-attachable. The pod will not be started. This can happen if a pod using csi volume get deployed before the CSIDriver object, which has `attachRequired: false` being deployed.  Because the A/D controller will assume the csi plugin is attachable. 

This PR fixs the problem by checking the volume attachability even during attaching. And it will remove any unattachable volume from the desired state. 

Right now, the behavior is that: if the CSIDriver get deployed in the middle of attachment, the pod will be able to turn into running status. The VolumeAttachment object which was not supposed to exist will be deleted.

Fixes #89173

```release-note
CSIDriver object can be deployed during volume attachment.
```

```docs
Fix issue where pod startup failed when CSI driver changes from attachable to non-attachable.
```

/cc @saad-ali 
/cc @msau42 
/cc @jsafrane

### DIFF
--- a/pkg/controller/volume/attachdetach/populator/desired_state_of_world_populator.go
+++ b/pkg/controller/volume/attachdetach/populator/desired_state_of_world_populator.go
@@ -154,6 +154,22 @@ func (dswp *desiredStateOfWorldPopulator) findAndRemoveDeletedPods() {
 		klog.V(1).Infof("Removing pod %q (UID %q) from dsw because it does not exist in pod informer.", dswPodKey, dswPodUID)
 		dswp.desiredStateOfWorld.DeletePod(dswPodUID, dswPodToAdd.VolumeName, dswPodToAdd.NodeName)
 	}
+
+	// check if the existing volumes changes its attachability
+	for _, volumeToAttach := range dswp.desiredStateOfWorld.GetVolumesToAttach() {
+		// IsAttachableVolume() will result in a fetch of CSIDriver object if the volume plugin type is CSI.
+		// The result is returned from CSIDriverLister which is from local cache. So this is not an expensive call.
+		volumeAttachable := volutil.IsAttachableVolume(volumeToAttach.VolumeSpec, dswp.volumePluginMgr)
+		if !volumeAttachable {
+			klog.Infof("Volume %v changes from attachable to non-attachable.", volumeToAttach.VolumeName)
+			for _, scheduledPod := range volumeToAttach.ScheduledPods {
+				podUID := volutil.GetUniquePodName(scheduledPod)
+				dswp.desiredStateOfWorld.DeletePod(podUID, volumeToAttach.VolumeName, volumeToAttach.NodeName)
+				klog.V(4).Infof("Removing podUID: %v, volume: %v on node: %v from desired state of world"+
+					" because of the change of volume attachability.", podUID, volumeToAttach.VolumeName, volumeToAttach.NodeName)
+			}
+		}
+	}
 }
 
 func (dswp *desiredStateOfWorldPopulator) findAndAddActivePods() {
--- a/pkg/kubelet/volumemanager/cache/desired_state_of_world.go
+++ b/pkg/kubelet/volumemanager/cache/desired_state_of_world.go
@@ -120,6 +120,9 @@ type DesiredStateOfWorld interface {
 
 	// GetPodsWithErrors returns names of pods that have stored errors.
 	GetPodsWithErrors() []types.UniquePodName
+
+	// MarkVolumeAttachability updates the volume's attachability for a given volume
+	MarkVolumeAttachability(volumeName v1.UniqueVolumeName, attachable bool)
 }
 
 // VolumeToMount represents a volume that is attached to this node and needs to
@@ -233,8 +236,8 @@ func (dsw *desiredStateOfWorld) AddPodToVolume(
 
 	// The unique volume name used depends on whether the volume is attachable/device-mountable
 	// or not.
-	attachable := dsw.isAttachableVolume(volumeSpec)
-	deviceMountable := dsw.isDeviceMountableVolume(volumeSpec)
+	attachable := util.IsAttachableVolume(volumeSpec, dsw.volumePluginMgr)
+	deviceMountable := util.IsDeviceMountableVolume(volumeSpec, dsw.volumePluginMgr)
 	if attachable || deviceMountable {
 		// For attachable/device-mountable volumes, use the unique volume name as reported by
 		// the plugin.
@@ -410,31 +413,6 @@ func (dsw *desiredStateOfWorld) GetVolumesToMount() []VolumeToMount {
 	return volumesToMount
 }
 
-func (dsw *desiredStateOfWorld) isAttachableVolume(volumeSpec *volume.Spec) bool {
-	attachableVolumePlugin, _ :=
-		dsw.volumePluginMgr.FindAttachablePluginBySpec(volumeSpec)
-	if attachableVolumePlugin != nil {
-		volumeAttacher, err := attachableVolumePlugin.NewAttacher()
-		if err == nil && volumeAttacher != nil {
-			return true
-		}
-	}
-
-	return false
-}
-
-func (dsw *desiredStateOfWorld) isDeviceMountableVolume(volumeSpec *volume.Spec) bool {
-	deviceMountableVolumePlugin, _ := dsw.volumePluginMgr.FindDeviceMountablePluginBySpec(volumeSpec)
-	if deviceMountableVolumePlugin != nil {
-		volumeDeviceMounter, err := deviceMountableVolumePlugin.NewDeviceMounter()
-		if err == nil && volumeDeviceMounter != nil {
-			return true
-		}
-	}
-
-	return false
-}
-
 func (dsw *desiredStateOfWorld) AddErrorToPod(podName types.UniquePodName, err string) {
 	dsw.Lock()
 	defer dsw.Unlock()
@@ -468,4 +446,15 @@ func (dsw *desiredStateOfWorld) GetPodsWithErrors() []types.UniquePodName {
 		pods = append(pods, podName)
 	}
 	return pods
+}
+
+func (dsw *desiredStateOfWorld) MarkVolumeAttachability(volumeName v1.UniqueVolumeName, attachable bool) {
+	dsw.Lock()
+	defer dsw.Unlock()
+	volumeObj, volumeExists := dsw.volumesToMount[volumeName]
+	if !volumeExists {
+		return
+	}
+	volumeObj.pluginIsAttachable = attachable
+	dsw.volumesToMount[volumeName] = volumeObj
 }
--- a/pkg/kubelet/volumemanager/populator/desired_state_of_world_populator.go
+++ b/pkg/kubelet/volumemanager/populator/desired_state_of_world_populator.go
@@ -91,7 +91,8 @@ func NewDesiredStateOfWorldPopulator(
 	kubeContainerRuntime kubecontainer.Runtime,
 	keepTerminatedPodVolumes bool,
 	csiMigratedPluginManager csimigration.PluginManager,
-	intreeToCSITranslator csimigration.InTreeToCSITranslator) DesiredStateOfWorldPopulator {
+	intreeToCSITranslator csimigration.InTreeToCSITranslator,
+	volumePluginMgr *volume.VolumePluginMgr) DesiredStateOfWorldPopulator {
 	return &desiredStateOfWorldPopulator{
 		kubeClient:                kubeClient,
 		loopSleepDuration:         loopSleepDuration,
@@ -108,6 +109,7 @@ func NewDesiredStateOfWorldPopulator(
 		hasAddedPodsLock:         sync.RWMutex{},
 		csiMigratedPluginManager: csiMigratedPluginManager,
 		intreeToCSITranslator:    intreeToCSITranslator,
+		volumePluginMgr:          volumePluginMgr,
 	}
 }
 
@@ -127,6 +129,7 @@ type desiredStateOfWorldPopulator struct {
 	hasAddedPodsLock          sync.RWMutex
 	csiMigratedPluginManager  csimigration.PluginManager
 	intreeToCSITranslator     csimigration.InTreeToCSITranslator
+	volumePluginMgr           *volume.VolumePluginMgr
 }
 
 type processedPods struct {
@@ -222,6 +225,20 @@ func (dswp *desiredStateOfWorldPopulator) findAndRemoveDeletedPods() {
 	for _, volumeToMount := range dswp.desiredStateOfWorld.GetVolumesToMount() {
 		pod, podExists := dswp.podManager.GetPodByUID(volumeToMount.Pod.UID)
 		if podExists {
+
+			// check if the attachability has changed for this volume
+			if volumeToMount.PluginIsAttachable {
+				attachableVolumePlugin, err := dswp.volumePluginMgr.FindAttachablePluginBySpec(volumeToMount.VolumeSpec)
+				// only this means the plugin is truly non-attachable
+				if err == nil && attachableVolumePlugin == nil {
+					// It is not possible right now for a CSI plugin to be both attachable and non-deviceMountable
+					// So the uniqueVolumeName should remain the same after the attachability change
+					dswp.desiredStateOfWorld.MarkVolumeAttachability(volumeToMount.VolumeName, false)
+					klog.Infof("Volume %v changes from attachable to non-attachable.", volumeToMount.VolumeName)
+					continue
+				}
+			}
+
 			// Skip running pods
 			if !dswp.isPodTerminated(pod) {
 				continue
--- a/pkg/volume/util/operationexecutor/operation_generator.go
+++ b/pkg/volume/util/operationexecutor/operation_generator.go
@@ -342,7 +342,7 @@ func (og *operationGenerator) GenerateAttachVolumeFunc(
 				uncertainNode = derr.CurrentNode
 			}
 			addErr := actualStateOfWorld.MarkVolumeAsUncertain(
-				v1.UniqueVolumeName(""),
+				volumeToAttach.VolumeName,
 				volumeToAttach.VolumeSpec,
 				uncertainNode)
 			if addErr != nil {
@@ -418,10 +418,9 @@ func (og *operationGenerator) GenerateDetachVolumeFunc(
 	var err error
 
 	if volumeToDetach.VolumeSpec != nil {
-		attachableVolumePlugin, err =
-			og.volumePluginMgr.FindAttachablePluginBySpec(volumeToDetach.VolumeSpec)
+		attachableVolumePlugin, err = findDetachablePluginBySpec(volumeToDetach.VolumeSpec, og.volumePluginMgr)
 		if err != nil || attachableVolumePlugin == nil {
-			return volumetypes.GeneratedOperations{}, volumeToDetach.GenerateErrorDetailed("DetachVolume.FindAttachablePluginBySpec failed", err)
+			return volumetypes.GeneratedOperations{}, volumeToDetach.GenerateErrorDetailed("DetachVolume.findDetachablePluginBySpec failed", err)
 		}
 
 		volumeName, err =
@@ -1730,4 +1729,26 @@ func isDeviceOpened(deviceToDetach AttachedVolume, hostUtil hostutil.HostUtils) 
 		}
 	}
 	return deviceOpened, nil
+}
+
+// findDetachablePluginBySpec is a variant of VolumePluginMgr.FindAttachablePluginByName() function.
+// The difference is that it bypass the CanAttach() check for CSI plugin, i.e. it assumes all CSI plugin supports detach.
+// The intention here is that a CSI plugin volume can end up in an Uncertain state,  so that a detach
+// operation will help it to detach no matter it actually has the ability to attach/detach.
+func findDetachablePluginBySpec(spec *volume.Spec, pm *volume.VolumePluginMgr) (volume.AttachableVolumePlugin, error) {
+	volumePlugin, err := pm.FindPluginBySpec(spec)
+	if err != nil {
+		return nil, err
+	}
+	if attachableVolumePlugin, ok := volumePlugin.(volume.AttachableVolumePlugin); ok {
+		if attachableVolumePlugin.GetPluginName() == "kubernetes.io/csi" {
+			return attachableVolumePlugin, nil
+		}
+		if canAttach, err := attachableVolumePlugin.CanAttach(spec); err != nil {
+			return nil, err
+		} else if canAttach {
+			return attachableVolumePlugin, nil
+		}
+	}
+	return nil, nil
 }
--- a/test/e2e/storage/csi_mock_volume.go
+++ b/test/e2e/storage/csi_mock_volume.go
@@ -40,7 +40,9 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/pkg/controller/volume/scheduling"
+	"k8s.io/kubernetes/pkg/kubelet/events"
 	"k8s.io/kubernetes/test/e2e/framework"
+	e2eevents "k8s.io/kubernetes/test/e2e/framework/events"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
 	e2epv "k8s.io/kubernetes/test/e2e/framework/pv"
 	"k8s.io/kubernetes/test/e2e/storage/drivers"
@@ -56,6 +58,7 @@ const (
 	csiNodeLimitUpdateTimeout  = 5 * time.Minute
 	csiPodUnschedulableTimeout = 5 * time.Minute
 	csiResizeWaitPeriod        = 5 * time.Minute
+	csiVolumeAttachmentTimeout = 7 * time.Minute
 	// how long to wait for Resizing Condition on PVC to appear
 	csiResizingConditionWait = 2 * time.Minute
 
@@ -296,6 +299,77 @@ var _ = utils.SIGDescribe("CSI mock volume", func() {
 			})
 
 		}
+	})
+
+	ginkgo.Context("CSI CSIDriver deployment after pod creation using non-attachable mock driver", func() {
+		ginkgo.It("should bringup pod after deploying CSIDriver attach=false [Slow]", func() {
+			var err error
+			init(testParameters{registerDriver: false, disableAttach: true})
+			defer cleanup()
+
+			_, claim, pod := createPod(false /* persistent volume, late binding as specified above */)
+			if pod == nil {
+				return
+			}
+
+			ginkgo.By("Checking if attaching failed and pod cannot start")
+			eventSelector := fields.Set{
+				"involvedObject.kind":      "Pod",
+				"involvedObject.name":      pod.Name,
+				"involvedObject.namespace": pod.Namespace,
+				"reason":                   events.FailedAttachVolume,
+			}.AsSelector().String()
+			msg := "AttachVolume.Attach failed for volume"
+
+			err = e2eevents.WaitTimeoutForEvent(m.cs, pod.Namespace, eventSelector, msg, framework.PodStartTimeout)
+			if err != nil {
+				podErr := e2epod.WaitTimeoutForPodRunningInNamespace(m.cs, pod.Name, pod.Namespace, 10*time.Second)
+				framework.ExpectError(podErr, "Pod should not be in running status because attaching should failed")
+				// Events are unreliable, don't depend on the event. It's used only to speed up the test.
+				framework.Logf("Attach should fail and the corresponding event should show up, error: %v", err)
+			}
+
+			// VolumeAttachment should be created because the default value for CSI attachable is true
+			ginkgo.By("Checking if VolumeAttachment was created for the pod")
+			handle := getVolumeHandle(m.cs, claim)
+			attachmentHash := sha256.Sum256([]byte(fmt.Sprintf("%s%s%s", handle, m.provisioner, m.config.ClientNodeSelection.Name)))
+			attachmentName := fmt.Sprintf("csi-%x", attachmentHash)
+			_, err = m.cs.StorageV1().VolumeAttachments().Get(context.TODO(), attachmentName, metav1.GetOptions{})
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					framework.ExpectNoError(err, "Expected VolumeAttachment but none was found")
+				} else {
+					framework.ExpectNoError(err, "Failed to find VolumeAttachment")
+				}
+			}
+
+			ginkgo.By("Deploy CSIDriver object with attachRequired=false")
+			driverNamespace := m.config.DriverNamespace
+
+			canAttach := false
+			o := utils.PatchCSIOptions{
+				OldDriverName: "csi-mock",
+				NewDriverName: "csi-mock-" + f.UniqueName,
+				CanAttach:     &canAttach,
+			}
+			cleanupCSIDriver, err := utils.CreateFromManifests(f, driverNamespace, func(item interface{}) error {
+				return utils.PatchCSIDeployment(f, o, item)
+			}, "test/e2e/testing-manifests/storage-csi/mock/csi-mock-driverinfo.yaml")
+			if err != nil {
+				framework.Failf("fail to deploy CSIDriver object: %v", err)
+			}
+			m.testCleanups = append(m.testCleanups, cleanupCSIDriver)
+
+			ginkgo.By("Wait for the pod in running status")
+			err = e2epod.WaitForPodNameRunningInNamespace(m.cs, pod.Name, pod.Namespace)
+			framework.ExpectNoError(err, "Failed to start pod: %v", err)
+
+			ginkgo.By(fmt.Sprintf("Wait for the volumeattachment to be deleted up to %v", csiVolumeAttachmentTimeout))
+			// This step can be slow because we have to wait either a NodeUpdate event happens or
+			// the detachment for this volume timeout so that we can do a force detach.
+			err = waitForVolumeAttachmentTerminated(attachmentName, m.cs)
+			framework.ExpectNoError(err, "Failed to delete VolumeAttachment: %v", err)
+		})
 	})
 
 	ginkgo.Context("CSI workload information using mock driver", func() {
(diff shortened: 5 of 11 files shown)


pkg/volume/util/operationexecutor/operation_generator.go
```diff
@@ -1731,3 +1730,25 @@ func isDeviceOpened(deviceToDetach AttachedVolume, hostUtil hostutil.HostUtils)
...
+// The difference is that it bypass the CanAttach() check for CSI plugin, i.e. it assumes all CSI plugin supports detach.
+// The intention here is that a CSI plugin volume can end up in an Uncertain state,  so that a detach
+// operation will help it to detach no matter it actually has the ability to attach/detach.
+func findDetachablePluginBySpec(spec *volume.Spec, pm *volume.VolumePluginMgr) (volume.AttachableVolumePlugin, error) {
+	volumePlugin, err := pm.FindPluginBySpec(spec)
+	if err != nil {
+		return nil, err
+	}
+	if attachableVolumePlugin, ok := volumePlugin.(volume.AttachableVolumePlugin); ok {
+		if attachableVolumePlugin.GetPluginName() == "kubernetes.io/csi" {
```
Jiawei0227, line 1744: @saad-ali I thought it through and I think we don't actually need any other check(e.g. inline check) in the CanAttach() function because we are expecting to get a attachableVolumePlugin from the return. The CanAttach() should never return false to the caller.

That said, in this case, as long as it is a CSI plugin, it should be detachable no matter it is attachable or not. Please correct me if I am wrong.

pkg/volume/util/operationexecutor/operation_generator.go
```diff
@@ -342,7 +342,7 @@ func (og *operationGenerator) GenerateAttachVolumeFunc(
 				uncertainNode = derr.CurrentNode
 			}
 			addErr := actualStateOfWorld.MarkVolumeAsUncertain(
-				v1.UniqueVolumeName(""),
+				volumeToAttach.VolumeName,
```
Jiawei0227, line 345: I dont understand why we are not passing the VolumeName instead create an empty name here. This prevents us from marking an unattachable volume as `Uncertain`. I did some manual tests locally and find change it this way causes no problem. Please let me know if there is some special reason behind the empty name.

pkg/controller/volume/attachdetach/populator/desired_state_of_world_populator.go
```diff
@@ -113,6 +113,21 @@ func (dswp *desiredStateOfWorldPopulator) populatorLoopFunc() func() {
 // Iterate through all pods in desired state of world, and remove if they no
 // longer exist in the informer
 func (dswp *desiredStateOfWorldPopulator) findAndRemoveDeletedPods() {
+
+	// check if the existing volumes changes its attachability
+	for _, volumeToAttach := range dswp.desiredStateOfWorld.GetVolumesToAttach() {
```
saad-ali, line 118: Would it make sense to move this to the end (first remove deleted pods, then check for this)?

pkg/controller/volume/attachdetach/populator/desired_state_of_world_populator.go
```diff
@@ -113,6 +113,21 @@ func (dswp *desiredStateOfWorldPopulator) populatorLoopFunc() func() {
...
+		volumeAttachable := volutil.IsAttachableVolume(volumeToAttach.VolumeSpec, dswp.volumePluginMgr)
+		if !volumeAttachable {
+			klog.Infof("Volume %v changes from attachable to non-attachable.", volumeToAttach.VolumeName)
+			for _, scheduledPod := range volumeToAttach.ScheduledPods {
+				podUID := volutil.GetUniquePodName(scheduledPod)
+				dswp.desiredStateOfWorld.DeletePod(podUID, volumeToAttach.VolumeName, volumeToAttach.NodeName)
```
saad-ali, line 124: Can you please add unit tests for the new behavior.

pkg/controller/volume/attachdetach/populator/desired_state_of_world_populator.go
```diff
@@ -113,6 +113,21 @@ func (dswp *desiredStateOfWorldPopulator) populatorLoopFunc() func() {
 // Iterate through all pods in desired state of world, and remove if they no
 // longer exist in the informer
 func (dswp *desiredStateOfWorldPopulator) findAndRemoveDeletedPods() {
+
+	// check if the existing volumes changes its attachability
+	for _, volumeToAttach := range dswp.desiredStateOfWorld.GetVolumesToAttach() {
+		volumeAttachable := volutil.IsAttachableVolume(volumeToAttach.VolumeSpec, dswp.volumePluginMgr)
```
saad-ali, line 119: This results in a `csiDriver, err := p.csiDriverLister.Get(driver)`. Can you double check that does not result in a network call (that it uses in-memory cache). I believe informers will cache, not sure about listers.

Once you verify, put a comment indicating that this will result in a fetch of the CSIDriver object if the volume plugin type is CSI, but because the result is cached, this should not be an issue.

pkg/volume/util/operationexecutor/operation_generator.go
saad-ali: For CSI volume that are Ephemeral volume should not be attachable and therefore should not be detachable (which is checked by `CanAttach()`.

That said, if the proper checks happen on the attach side, detach may be less strict. So this should be fine though it makes me a little nervous.

pkg/volume/util/operationexecutor/operation_generator.go
```diff
@@ -1731,3 +1730,25 @@ func isDeviceOpened(deviceToDetach AttachedVolume, hostUtil hostutil.HostUtils)
 	}
 	return deviceOpened, nil
 }
+
+// findDetachablePluginBySpec is a variant of VolumePluginMgr.FindAttachablePluginByName() function.
+// The difference is that it bypass the CanAttach() check for CSI plugin, i.e. it assumes all CSI plugin supports detach.
+// The intention here is that a CSI plugin volume can end up in an Uncertain state,  so that a detach
+// operation will help it to detach no matter it actually has the ability to attach/detach.
+func findDetachablePluginBySpec(spec *volume.Spec, pm *volume.VolumePluginMgr) (volume.AttachableVolumePlugin, error) {
```
saad-ali, line 1738: Consider putting this in the same file as the other similar functions (e.g. `FindAttachablePluginByName`).

pkg/volume/util/operationexecutor/operation_generator.go
saad-ali: @jingxu97 can you recall why this was empty?

saad-ali (changes requested):

pkg/volume/util/operationexecutor/operation_generator.go
```diff
@@ -342,7 +342,7 @@ func (og *operationGenerator) GenerateAttachVolumeFunc(
 				uncertainNode = derr.CurrentNode
 			}
 			addErr := actualStateOfWorld.MarkVolumeAsUncertain(
```
saad-ali, line 344: How did you address the race we discussed? CSIDriver object is created (with `skipattach`) during first attach attempt, so uncertain state fails to set.

saad-ali (changes requested):

pkg/volume/util/operationexecutor/operation_generator.go
Jiawei0227: This is a very interesting point. In fact, if we pass the volumeName to the MarkVolumeAsUncertain() function, the markasuncertain operation will not fail. So in that case, the race will not exist because MarkVolumeAsUncertain will always be able to happen no matter when do we deploy the CSIDriver. That's why I am curious about why we are giving an empty volumeName. 

If we have to keep the empty volume name for some reason. What will happen is that the MarkVolumeAsUncertain will fail. And the volumeattachment object will be long existing. But the pod is able to bringup because kubelet will detect its attachability change.

pkg/volume/util/operationexecutor/operation_generator.go
jingxu97: I dig some old code, and think it might be historical reason. Before, MarkVolumeAsAttached has no uniqueVolumeName parameter. It only passes volume spec and use spec to find out the unique name. Later we added this parameter to fix some issue. It seems like we just continue to use volume spec to find out the volume name instead of passing it directly. 
So far I don't see an issue directly setting it to volume name.

jsafrane:
/ok-to-test
@gnufied, can you please take a look?

pkg/volume/util/operationexecutor/operation_generator.go
Jiawei0227: If that's the case, It should be safe to use the volumeName directly. Let's see the results of all e2e tests and unit tests to see if that changes the statement.

Jiawei0227 (commented):
Hi @saad-ali 

Thanks for your review! I have applied all the changes and made two unit tests: one for attach/detach controller and one for volumemanager. And also run the unit tests in my local to verify the result. Please let me know if you have any other suggestions.

jingxu97:
@Jiawei0227 Could you list the steps to reproduce the issue and the test? Based on this, we could add e2e tests later?
Just to make it more complete, could you check what will happen if volume is changed from non-attachable to attachable?

Jiawei0227:
Hey @jingxu97 
The repro process is well documented in the issue #89173

> How to reproduce it (as minimally and precisely as possible):
> 1. Install CSI mock driver:
> 2. With --disable-attach=true, so the mock driver does not require ControllerPublish)
> 3. Without the external-attacher sidecar, so we can simulate slow attach.
> 4. Without any CSIDriver instance, so A/D controller requires attach.
> 5. Create a PVC and start a pod.
> 6. Wait for time outs when attaching the provisioned volume, because the external-attacher is not running.
> 7. Create CSIDriver with attachRequired: false.

I will try to test if it becomes from nonattachable to attachable and update here.

msau42:
Adding an e2e test for this scenario is a great idea.

Jiawei0227:
e2e test added. I am using mock CSI driver to reproduce the scenario.
And back to @jingxu97 question, I thought about it and it seems like that should not be our concern.

If a csi driver is non-attachable at the beginning. Then the attach/detach controller will just ignore that. And the mount operation will be taken care of by the volumemanager in kubelet. So we are good. But if somehow it becomes attachable in the middle, which is very unlikely to happen, then the attach/detach controller will find the volume needs to be attached and add the volume into the desired_state_world. Then we will have the VolumeAttachment object being created. And since this CSI plugin becomes attachable, there should be an external-attacher to take care of that. So we are also good in this situation.

But if the CSIDriver is attachable in nature and we state that as non-attachable. I believe the mount operation for the volume should fail in the NodePulishVolume or NodeStageVolume should fail? And that should also be as expected.

pkg/kubelet/volumemanager/populator/desired_state_of_world_populator.go
```diff
@@ -222,6 +225,17 @@ func (dswp *desiredStateOfWorldPopulator) findAndRemoveDeletedPods() {
 	for _, volumeToMount := range dswp.desiredStateOfWorld.GetVolumesToMount() {
 		pod, podExists := dswp.podManager.GetPodByUID(volumeToMount.Pod.UID)
 		if podExists {
+
+			// check if the attachability has changed for this volume
+			if volumeToMount.PluginIsAttachable &&
```
gnufied, line 230: Why do we need to delete the volume from volumemanager's DSOW as a special case? Is this merely to mark the fact that, volume is no longer attachable?

pkg/controller/volume/attachdetach/populator/desired_state_of_world_populator.go
```diff
@@ -154,6 +154,22 @@ func (dswp *desiredStateOfWorldPopulator) findAndRemoveDeletedPods() {
...
+		if !volumeAttachable {
+			klog.Infof("Volume %v changes from attachable to non-attachable.", volumeToAttach.VolumeName)
+			for _, scheduledPod := range volumeToAttach.ScheduledPods {
+				podUID := volutil.GetUniquePodName(scheduledPod)
+				dswp.desiredStateOfWorld.DeletePod(podUID, volumeToAttach.VolumeName, volumeToAttach.NodeName)
+				klog.V(4).Infof("Removing podUID: %v, volume: %v on node: %v from desired state of world"+
+					" because of the change of volume attachability.", podUID, volumeToAttach.VolumeName, volumeToAttach.NodeName)
+			}
+		}
+	}
```
gnufied, line 172: we already have node, pod and volume name from `GetPodToAdd` function. Shouldn't we able to get volumeSpec from that and integrate removal of pod from DSOW in the loop above rather than creating a new loop?

Jiawei0227: IMHO, the way `GetPodToAdd` returns the pods list but that does not guarantee to have all the volumes we are referring to. There might be cases where pod A mounted volume x, y. And in the `GetPodToAdd` function, the final return pod list will only have a map of pod A to volume y. Because volume y has been overridden. Secondly, there might be cases where volume x has two corresponding pods. If we check the volume from the `GetPodToAdd` loop. There will be an overhead of `IsAttachableVolume` check for a volume.

Does this makes sense?

pkg/kubelet/volumemanager/populator/desired_state_of_world_populator.go
Jiawei0227: Because our original `findAndRemoveDeletedPods` loop does not have the ability to detect if a volume plugin changes its attach-ability. Thus, if we dont add this check here, what happened is the reconciler will never happy and not able to bringup the pod because it always wait for that volume to be attached.

pkg/controller/volume/attachdetach/populator/desired_state_of_world_populator.go
gnufied: hmm, the first issue sounds like inefficiency in populator. if a pod has more than 1 volume - it may require more than one reconciliation loop for both volumes and pod to disappear from DSOW, which could be 1 minute apart.  IMO - it is worth fixing that, but we can file this as a separate issue.

test/e2e/storage/csi_mock_volume.go
```diff
@@ -298,6 +299,75 @@ var _ = utils.SIGDescribe("CSI mock volume", func() {
...
+				var err error
+				init(testParameters{registerDriver: test.deployClusterRegistrar, disableAttach: test.disableAttach})
+				defer cleanup()
+
+				_, claim, pod := createPod(false)
+				if pod == nil {
+					return
+				}
+				err = e2epod.WaitTimeoutForPodRunningInNamespace(m.cs, pod.Name, pod.Namespace, 10*time.Second)
+				framework.ExpectError(err, "Pod should not be in running status because attaching should failed")
```
saad-ali, line 326: if `test.disableAttach` is `false` is this still expected to fail here?

saad-ali (approved):
nits otherwise LGTM

/lgtm
/approve
/hold

Adding hold to let @gnufied finish his review.

pkg/kubelet/volumemanager/populator/desired_state_of_world_populator.go
gnufied: can this cause a pod's volume to be unmounted if volume originally was attachable but somehow a CSIDriver deployment changed it to become unattachable?

Jiawei0227:
Some discussion I had with @saad-ali that worth keep a record here:

A/D controller is only responsible for attachable CSI drivers. The A/D controller will monitor on the node status `VolumesInUse`, and update the `VolumeAttached` on the node. The volumemanager on the other hand, will monitor the node status about `VolumeAttached` and update `VolumesInUse`. So that these two components are aware of when should we perform attach/detach and mount/unmount operation.

For the `VolumeInUse` field, one thing to mention here is that it only cares about attachable volumes. Since if the volume is not attachable, there is also no need to perform a detachment operation from the A/D point of view. 

The clarification is useful when we talk about the case if there is a CSIDriver deployment changes its attachability from attachable to non-attachable even if there are running pods that have a reference to a volume of its type.
> A/D only cares about attachable volumes. It depends on volumesInUse list for safe detach. If a volume switches from attachable to non-attachable, the volume is removed from volumesInUse as soon as possible, meaning A/D controller could do unsafe detach. But that shouldn't matter because the only reason that would happen is if user explicitly takes a attachable type and creates a CSIDriver object for it saying it is not-attachable after the volume has been attached.

A quote from @saad-ali that also well explained the situation here.

pkg/kubelet/volumemanager/populator/desired_state_of_world_populator.go
gnufied: But `IsAttachableVolume` check depends on informers and cache being in sync (in other words network), so it is not just a local state. For example - a CSI plugin that was attachable previously can be considered non-attachable if informers are being resynced or CSIDriver object can't be found (because of network error).  In which case kubelet would proceed to unmount pod's volume and can cause data loss and data corruption.

Jiawei0227: Hmm, yeah. That is a solid concern. But I guess the question is then how do we decide if it is a network issue or is the attachability changed for a CSIDriver. Also, I am new to this so I might be wrong, but isn't informers and cache also a local state? I am wondering how other components handle situation like this. @saad-ali any concerns or thoughts here?

gnufied: At very minimum - `IsAttachableVolume` check is flawed as means of determining whether volume should be removed from desired state of the world. If volume was previously attachable  but anything fails in `FindAttachablePluginBySpec`, we assume volume is not attachable anymore. Whereas this should be other way around - if and only if - `FindAttachablePluginBySpec` returns nil plugin but no error, we can safely assume that plugin is indeed not attachable.  The burden of verifying whether plugin is attachable or not is much higher in this case.

But I also think that - we are making this fix in wrong place IMO. Deleting the pod from desired state as I understand will cause all previous assumptions to be invalidated and hence pod can be started.  But can we not fix this in `GenerateMountXXX` function, so as if a plugin changes its attchability then mount operation does not wait for volume to become attached? is something I missed?

msau42: I agree with Hemant that an unmount/mount while the container is still running is undesirable and could cause data corruption/loss.

It may be useful to think about the different scenarios we could run into when a CSIDriver changes from attachable to non-attachable:

1. Before AD controller adds it to DSW
2. After AD controller adds it to DSW and before it creates the VolumeAttachment object
3. After AD controller creates the VolumeAttachment object and kubelet is stuck waiting (because there's no sidecar running)
4. Volume is already mounted and pod is already running (because previously the sidecar was running before the driver got upgraded)

I think the behavior we want is:
- Be able to resolve scenarios 2 + 3
- Delete the VolumeAttachment object
- Clean up Node.Status.VolumesAttached
- Clean up any internal state
- All the while, not disturbing any mounts

Jiawei0227: Thanks Hemant and Michelle for your feedback!

Re Michelle:
I also agree that unmount/mount while container is running is undesirable. But this might only happen if there is already a pod running using the volume and we change the volume from attachable to non-attachable. In that case, it is the user's choice to do the flip.

The problem with the scenarios you mentioned is that we not only need to care about A/D controller, but we also need to care about volumemanager which resides in the kubelet side. In the PR that I already made, we have taken care of the A/D side, so let's break down the scenarios you mentioned for A/D (The result is based on the PR I have):
1. There is no problem on this, because if the attach-ability change happens before AD add it to DSW, then it will never add it to DSW.
2. In my fix, there is a race between the VolumeAttachment creation and the removal of the volume from DSW. If VolumeAttachment creation happens first, then what gonna happen is we will remove the volume from DSW, and a detach operation will be triggered and then the VolumeAttachment will be deleted. If the removal happens first, then VolumeAttachment will not be created which is also good.
3. As mentioned in 2, we will remove the volume from DSW, and a detach operation will kick in. The detach will either timeout or get an update from nodeStatus to inform it that the volume is no long in the attachableVolume list. Both cases the VolumeAttachment will then be deleted.
4. The only problematic scenario is this one: so what will happen is that a detach operation will be triggered and there are multiple different outcomes depending on the driver. There might be cases where even the CSIDriver said it does not support attach/detach, it can still perform a detach operation which will make the pod lost its access to the volume anymore. There might be cases where the driver does not support detach operation so the detach for it is a no-op. Besides, we will perform a unmount/mount for the pod.

So I guess we have satisfied the behavior you want except `All the while, not disturbing any mounts`. But then the question is, if the pod is already running with the CSIDriver, why do we change the driver's attachability? The only possible scenario is we upgraded the CSIDriver to a version where it does not support attach/detach anymore, which IMO it should be deployed with a different CSIDriver name. So it is really the user's call if they decided to meddling the driver's attachability if there is already running pods using that driver.


Re: Hemant
I agree we need to loose the isAttachableVolume check so that it only states a volume is nonattachable when no err and it return nil.

Now let's talk about why I am deleting the pod from volume, it is true that it will invalidate the assumptions we made, but that comes with a good reason: we changed the volume attachability. Thus, a lot of assumptions we made in the previous run is wrong. Here is a detailed explanation:
The DSW in volumemanager keeps the records of all Volumes it needs to mount. However, it keeps a record of the volume attach-ability when it adds the volume into its list here: https://github.com/kubernetes/kubernetes/blob/04362870ad9f9812f487ab8f89a40f6bf2daa588/pkg/kubelet/volumemanager/cache/desired_state_of_world.go#L169
So in the beginning I thought maybe we can just flip the boolean so that it can proceed. But then I found out I cannot do so because the `VolumeUniqueName` which is the primary key for the volumeToMount list in the DSW is generated based on its attach-ability:
https://github.com/kubernetes/kubernetes/blob/04362870ad9f9812f487ab8f89a40f6bf2daa588/pkg/kubelet/volumemanager/cache/desired_state_of_world.go#L234
So there might be cases if we deployed the CSIDriver, the whole volume unique name for that volume will be changed. That can be problematic because the original volumeToMount in DSW will never be satisfied and thus the pod can never be bringup.

I could be wrong because the logic here is quite complicated but it is great to discuss in details for this.

msau42: To clarify what "driver is not attachable" means, there's a couple of scenarios:
1. CSI driver does not implement ControllerPublish, but is running the csi-attacher in the simple mode where all it does is set the VolumeAttachment.Status.attached = true without calling to the driver. This was what drivers had to do before we introduced the CSIDriver.Spec,AttachedRequired field.
2. CSI driver does not implement ControllerPublish, is not running csi-attacher, and creates a CSIDriver object and sets AttachedRequired = false.

Upgrading from 1->2 will hit this issue and I expect any non-attachable driver that was deployed before CSIDriver.AttachedRequired existed can have existing volumes that are mounted. I don't think we can leave it to the users to handle this case. The admin managing the CSI driver is different than the user running Pods, and in managed distributions that bundle CSI drivers, the upgrade of the CSI driver is potentially out of the user's control.

2) will hit this issue if the CSIDriver object is deployed later. Potentially this could be mitigated by deployed the csi driver well before any workloads can run, but that may not always be controllable.

Jiawei0227: Hmm, yeah that makes sense. I am trying to find out if we can update the internal state of DSW directly instead delete and create it.

saad-ali: 2 concerns stand out to me: 1) potential for detach or unmount of in-use volume, and 2) assuming default value on failure to fetch CSIDriver object.

--1--

With @Jiawei0227's current proposal, in this case, the A/D controller would delete the VolumeAttachment object because it depends on Node `volumesInUse` list for safe detach. When a volume switches from attachable to non-attachable, the volume is removed from `volumesInUse` by the kubelet as soon as possible, meaning A/D controller could do unsafe detach. This is problematic because of the use case @msau42 pointed out:

>>> Volume is already mounted and pod is already running (because previously the sidecar was running before the driver got upgraded)
>> 
>> The only problematic scenario is this one: so what will happen is that a detach operation will be triggered and there are multiple different outcomes depending on the driver. There might be cases where even the CSIDriver said it does not support attach/detach, it can still perform a detach operation which will make the pod lost its access to the volume anymore. There might be cases where the driver does not support detach operation so the detach for it is a no-op. Besides, we will perform a unmount/mount for the pod.
>
> Upgrading from 1->2 will hit this issue and I expect any non-attachable driver that was deployed before CSIDriver.AttachedRequired existed can have existing volumes that are mounted.

This^^ is a valid use case, and we should handle it gracefully. 
To handle it correctly, I don't think any further changes are needed on the a/d controller side, since it already depends on the kubelet to tell it when it is safe to detach -- the problem is on the kubelet side, with the new logic. the volume is removed from `volumeInUse` as soon as it switches from attachable to non-attachable. So would it be sufficient to add a condition on the kubelet volume manager side that does not remove from DSW if the pod is already running (@Jiawei0227 was considering this when we chatted yesterday)?

--2--

>> At very minimum - IsAttachableVolume check is flawed as means of determining whether volume should be removed from desired state of the world. If volume was previously attachable but anything fails in FindAttachablePluginBySpec, we assume volume is not attachable anymore. Whereas this should be other way around - if and only if - FindAttachablePluginBySpec returns nil plugin but no error, we can safely assume that plugin is indeed not attachable. The burden of verifying whether plugin is attachable or not is much higher in this case.
> 
> Hmm, yeah. That is a solid concern. But I guess the question is then how do we decide if it is a network issue or is the attachability changed for a CSIDriver. Also, I am new to this so I might be wrong, but isn't informers and cache also a local state? I am wondering how other components handle situation like this. @saad-ali any concerns or thoughts here?

This is also a valid concern. Ideally a failure should mean both A/d controller and kubelet volume manager continue to assume prior state about attachability. Can we surface errors, and compare them to assumed attachability state? If that is really difficult to do, we can consider sane defaults, but we'll need to think through the implications of that.

saad-ali: I met with @Jiawei0227 to discuss this.

Regarding issue number 1 above in my comment, after discussing with @Jiawei0227 I don't think it is a problem worth fixing. Because, In the case Michelle pointed out going from attachable (CSI Attacher with no-op attach) to non-attachable (CSI Driver object saying skip attach), even if detach is inadvertently called, because CSI attacher was a no-op, nothing bad should happen (no risk of data corruption on a no-op detach). The only case this would be a concern would be if user creates a CSIDriver object saying skip attach for a driver that implements a real attach/detach -- that does not seem like a realistic scenario.

Regarding issue number 2 above in my comment, after discussing with @Jiawei0227 he agreed to have the attachability check surface any error to his attachability checks in both A/d controller and kubelet volume manager, and if there is an error, have the checks continue to assume prior state about attachability (don't assumes state change unless explicitly confirmed). He will update his code to `FindAttachablePluginBySpec`.

Finally @Jiawei0227 pointed out @gnufied's main concern, which was to prevent rapid unmount/mount of volumes in use that could happen with the existing proposal. I agree this is a major issue. To fix it, @Jiawei0227 proposed instead of creating/deleting volumes in volume manager DSW, he will update the attachability state for the volume instead.

Jiawei0227: After discussing with @saad-ali 

We we decide to go with an update mechanism instead of a delete/create machoism for volumes in DSW of volume manager. This way we will not trigger an unmount/mount for existing running pods and thus bring no harms. Also, I have adopted what @gnufied suggested which is to use the FindAttachablePluginBySpec and only state a plugin is non-attachable if no err and nil volume plugin return.

pkg/kubelet/volumemanager/cache/desired_state_of_world.go
```diff
@@ -469,3 +447,21 @@ func (dsw *desiredStateOfWorld) GetPodsWithErrors() []types.UniquePodName {
...
+		return
+	}
+	dsw.volumesToMount[volumeName] = volumeToMount{
+		volumeName:              volumeObj.volumeName,
+		podsToMount:             volumeObj.podsToMount,
+		pluginIsAttachable:      attachable,
+		pluginIsDeviceMountable: volumeObj.pluginIsDeviceMountable,
+		volumeGidValue:          volumeObj.volumeGidValue,
+		reportedInUse:           volumeObj.reportedInUse,
+		desiredSizeLimit:        volumeObj.desiredSizeLimit,
```
gnufied, line 465: why not write this as:

```
volumeObj.pluginIsAttachable = attachable
dsw.volumesToMount[volumeName] = volumeObj
```

test/e2e/storage/csi_mock_volume.go
```diff
@@ -298,6 +299,62 @@ var _ = utils.SIGDescribe("CSI mock volume", func() {
...
+		ginkgo.It("should bringup pod after deploying CSIDriver attach=false [Slow]", func() {
+			var err error
+			init(testParameters{registerDriver: false, disableAttach: true})
+			defer cleanup()
+
+			_, claim, pod := createPod(false)
+			if pod == nil {
+				return
+			}
+			err = e2epod.WaitTimeoutForPodRunningInNamespace(m.cs, pod.Name, pod.Namespace, 10*time.Second)
```
gnufied, line 312: I think 10 seconds might be too small value for pod to fail to start, I know this will add to the test suite time but how about something like 1 minute?

msau42: Most of our e2es are using the podstartuptimeout defined in the e2e framework, which is 5 minutes I believe

gnufied: yeah - I kind of hate to add definitive 5 minute delay to the test suite. In other places 5 minute is the worst case timeout but in this case it will for sure hang around for 5minutes before erroring out.

pkg/kubelet/volumemanager/populator/desired_state_of_world_populator.go
```diff
@@ -222,6 +225,20 @@ func (dswp *desiredStateOfWorldPopulator) findAndRemoveDeletedPods() {
...
 		if podExists {
+
+			// check if the attachability has changed for this volume
+			if volumeToMount.PluginIsAttachable {
+				attachableVolumePlugin, err := dswp.volumePluginMgr.FindAttachablePluginBySpec(volumeToMount.VolumeSpec)
+				// only this means the plugin is truly non-attachable
+				if err == nil && attachableVolumePlugin == nil {
+					// It is not possible right now for a CSI plugin to be both attachable and non-deviceMountable
+					// So the uniqueVolumeName should remain the same after the attachability change
+					dswp.desiredStateOfWorld.MarkVolumeAttachability(volumeToMount.VolumeName, false)
```
gnufied, line 236: I like this much better and we are no longer deleting volumes from DSOW. thank you.

test/e2e/storage/csi_mock_volume.go
msau42: We can't rely on events, but what we've done in other negative test cases is to exit the condition early if we find the event we're looking for, otherwise wait for the full 5 minutes

pkg/kubelet/volumemanager/cache/desired_state_of_world.go
Jiawei0227: Done

test/e2e/storage/csi_mock_volume.go
Jiawei0227: @msau42 are you suggesting we should check events while waiting here? If so, can you give me an example of the negative test cases you mentioned?

test/e2e/storage/csi_mock_volume.go
```diff
@@ -298,6 +299,62 @@ var _ = utils.SIGDescribe("CSI mock volume", func() {
 		}
 	})
 
+	ginkgo.Context("CSI CSIDriver deployment after pod creation using non-attachable mock driver", func() {
+		ginkgo.It("should bringup pod after deploying CSIDriver attach=false [Slow]", func() {
+			var err error
+			init(testParameters{registerDriver: false, disableAttach: true})
+			defer cleanup()
+
+			_, claim, pod := createPod(false)
```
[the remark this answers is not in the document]
Jiawei0227: Done

pkg/kubelet/volumemanager/populator/desired_state_of_world_populator.go
Jiawei0227: Yeah, after some discussions this seems like a much easier and safe approach.

test/e2e/storage/csi_mock_volume.go
msau42: https://github.com/kubernetes/kubernetes/blob/6d0f4749a59099171540d4fd7c9523b029e71ceb/test/e2e/storage/testsuites/volumemode.go#L239

msau42: or if there's a pod status/condition we can check to speed up the wait

Jiawei0227: Done. I end up checking the events, dont know if there is a status or condition we can check here. An event seems to be working fine.

saad-ali (approved):
/hold cancel

Jiawei0227:
Hmm did I broke something...
/retest

PR #92989, opened 2020-07-11
add kubectl fish shell completion

WLun001:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

**What this PR does / why we need it**:
add fish shell completion, `kubectl completion fish`

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Add fish shell completion to kubectl
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #93714, opened 2020-08-05
Implement custom completions for some flags using Go

knight42:
**What type of PR is this?**

/kind feature

Implement custom completions using Go for flag `--context`, `--cluster`, `--user` and `--namespace`.

Part of https://github.com/kubernetes/kubectl/issues/882

### Test the new completion code

1. build kubectl from source
```sh
make WHAT=cmd/kubectl
```

2. generate completion code
```sh
# zsh
source <(./_output/bin/kubectl completion zsh)

# bash
source <(./_output/bin/kubectl completion bash)
```

3. play with it

Please note that the new completion code requires the new kubectl.

```
./_output/bin/kubectl get --context [TAB]

# debug
./_output/bin/kubectl __complete get --context ''
```

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kubectl/cmd/cmd.go
+++ b/pkg/kubectl/cmd/cmd.go
@@ -27,7 +27,7 @@ import (
 	"syscall"
 
 	"github.com/spf13/cobra"
-
+	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	cliflag "k8s.io/component-base/cli/flag"
@@ -52,6 +52,7 @@ import (
 	"k8s.io/kubectl/pkg/cmd/explain"
 	"k8s.io/kubectl/pkg/cmd/expose"
 	"k8s.io/kubectl/pkg/cmd/get"
+	"k8s.io/kubectl/pkg/cmd/kustomize"
 	"k8s.io/kubectl/pkg/cmd/label"
 	"k8s.io/kubectl/pkg/cmd/logs"
 	"k8s.io/kubectl/pkg/cmd/options"
@@ -72,11 +73,9 @@ import (
 	"k8s.io/kubectl/pkg/util/i18n"
 	"k8s.io/kubectl/pkg/util/templates"
 	"k8s.io/kubectl/pkg/util/term"
+
 	"k8s.io/kubernetes/pkg/kubectl/cmd/auth"
 	"k8s.io/kubernetes/pkg/kubectl/cmd/convert"
-
-	"k8s.io/cli-runtime/pkg/genericclioptions"
-	"k8s.io/kubectl/pkg/cmd/kustomize"
 )
 
 const (
@@ -287,15 +286,6 @@ __kubectl_custom_func() {
     esac
 }
 `
-)
-
-var (
-	bashCompletionFlags = map[string]string{
-		"namespace": "__kubectl_get_resource_namespace",
-		"context":   "__kubectl_config_get_contexts",
-		"cluster":   "__kubectl_config_get_clusters",
-		"user":      "__kubectl_config_get_users",
-	}
 )
 
 // NewDefaultKubectlCommand creates the `kubectl` command with default arguments
@@ -586,17 +576,7 @@ func NewKubectlCommand(in io.Reader, out, err io.Writer) *cobra.Command {
 
 	templates.ActsAsRootCommand(cmds, filters, groups...)
 
-	for name, completion := range bashCompletionFlags {
-		if cmds.Flag(name) != nil {
-			if cmds.Flag(name).Annotations == nil {
-				cmds.Flag(name).Annotations = map[string][]string{}
-			}
-			cmds.Flag(name).Annotations[cobra.BashCompCustom] = append(
-				cmds.Flag(name).Annotations[cobra.BashCompCustom],
-				completion,
-			)
-		}
-	}
+	registerCompletionFuncForGlobalFlags(cmds, f)
 
 	cmds.AddCommand(alpha)
 	cmds.AddCommand(cmdconfig.NewCmdConfig(f, clientcmd.NewDefaultPathOptions(), ioStreams))
--- a/pkg/kubectl/cmd/completion.go
+++ b/pkg/kubectl/cmd/completion.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/completion"
+)
+
+func registerCompletionFuncForGlobalFlags(cmd *cobra.Command, f util.Factory) {
+	util.CheckErr(cmd.RegisterFlagCompletionFunc(
+		"namespace",
+		func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			return completion.ListNamespaces(f, toComplete)
+		}))
+	util.CheckErr(cmd.RegisterFlagCompletionFunc(
+		"context",
+		func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			return completion.ListContextsInKubeconfig(f, toComplete)
+		}))
+	util.CheckErr(cmd.RegisterFlagCompletionFunc(
+		"cluster",
+		func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			return completion.ListClustersInKubeconfig(f, toComplete)
+		}))
+	util.CheckErr(cmd.RegisterFlagCompletionFunc(
+		"user",
+		func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			return completion.ListUsersInKubeconfig(f, toComplete)
+		}))
+}
--- a/staging/src/k8s.io/kubectl/pkg/completion/utils.go
+++ b/staging/src/k8s.io/kubectl/pkg/completion/utils.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package completion
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubectl/pkg/cmd/util"
+)
+
+// ListNamespaces returns a list of namespaces which begins with `toComplete`.
+func ListNamespaces(f util.Factory, toComplete string) ([]string, cobra.ShellCompDirective) {
+	clientSet, err := f.KubernetesClientSet()
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveDefault
+	}
+	ctx, cancel := context.WithTimeout(context.TODO(), time.Second*3)
+	defer cancel()
+	namespaces, err := clientSet.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveDefault
+	}
+	var ret []string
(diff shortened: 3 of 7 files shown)


knight42:
/hold

This PR is not intended to be merged now.

marckhouzam:
> Implement custom completions using Go for flag `--context`, `--cluster`, `--user` and `--namespace`.

Great!

> The changes depend on [spf13/cobra#1070](https://github.com/spf13/cobra/pull/1070), but it was not included in cobra 1.0.0. 

Actually, [spf13/cobra#1070](https://github.com/spf13/cobra/pull/1070) is not required for Go custom completions to work for both bash and zsh.   Those new custom completions will work for both bash and zsh with Cobra 1.0.  And even though zsh completion for kubectl currently re-uses the bash completion script, it will continue to work just fine with Go custom completions.  In fact, in Helm, we are using Cobra 1.0 and still using the zsh workaround used by kubectl, until the next release of Cobra is available.

So, if we aim for your orignal idea of keeping the PR small, I recommend staying on Cobra 1.0 and keeping the zsh completion code as it is in kubectl for now.  This PR could just add the Go custom completions for the four flags and they should work fine for both bash and zsh.

That would also mean you wouldn't need the `alpha completion` command but could just use the current, unchanged `completion` command.
>```
> # bash
> source <(./_output/bin/kubectl alpha completion bash)
> ```

This didn't seem to work right.  It generated a completion script for a binary called `alpha` instead of `kubectl`
But getting rid of the alpha completion command should fix all that.

What do you think?

pkg/kubectl/cmd/cmd.go
```diff
@@ -598,6 +602,54 @@ func NewKubectlCommand(in io.Reader, out, err io.Writer) *cobra.Command {
 		}
 	}
 
+	// TODO(knight42): handler error properly
```
marckhouzam, line 605: The error should be easy to handle.  An error at this point implies an error in the code, probably because more than one completion was registered for the same flag.  You can actually do a `log.Fatal` or something like that.  Since this will happen as soon as kubectl starts (I think), the problem will be found right away in development.

That is how helm does it: it just does a `log.Fatal` if this call returns an error.

knight42:
@marckhouzam A huge thank to your explanation! BTW should we remove the current bash completion flags? https://github.com/kubernetes/kubernetes/blob/382107e6c84374b229e6188207ef026621286aa2/pkg/kubectl/cmd/cmd.go#L293-L298

brianpursley:
> BTW should we remove the current bash completion flags?

Yes, I think as part of this, you should remove any custom bash code we have. After all, that is the main benefit of moving to this approach.

pkg/kubectl/cmd/cmd.go
```diff
@@ -598,6 +601,57 @@ func NewKubectlCommand(in io.Reader, out, err io.Writer) *cobra.Command {
 		}
 	}
 
+	cmdutil.CheckErr(cmds.RegisterFlagCompletionFunc("namespace", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
brianpursley, line 604: Any reason why this `RegisterFlagCompletionFunc` is done differently than the other three below?

pkg/kubectl/cmd/cmd.go
```diff
@@ -598,6 +601,57 @@ func NewKubectlCommand(in io.Reader, out, err io.Writer) *cobra.Command {
...
+		for name := range config.AuthInfos {
+			if strings.HasPrefix(name, toComplete) {
+				ret = append(ret, name)
+			}
+		}
+		return ret
+	})
+	cmdutil.CheckErr(cmds.RegisterFlagCompletionFunc("context", getContextFromConfig))
+	cmdutil.CheckErr(cmds.RegisterFlagCompletionFunc("cluster", getClusterFromConfig))
+	cmdutil.CheckErr(cmds.RegisterFlagCompletionFunc("user", getUserFromConfig))
```
brianpursley, line 607: It seems like there is going to be a lot of code here as you add more completions.  What about putting all the completion code into its own separate package and then make a one-line callout from here to register completions?

pkg/kubectl/cmd/cmd.go
knight42: because we could not get namespaces from local kubeconfig.

pkg/kubectl/cmd/cmd.go
knight42: updated

pkg/kubectl/cmd/cmd.go
```diff
@@ -116,31 +115,6 @@ __kubectl_override_flags()
...
-    __kubectl_parse_config "clusters"
-}
-
-__kubectl_config_get_users()
-{
-    __kubectl_parse_config "users"
-}
-
-# $1 has to be "contexts", "clusters" or "users"
-__kubectl_parse_config()
```
marckhouzam, line 135: So we'll need to keep this one since the above ones we need to keep use it.

marckhouzam (changes requested):
I haven't look at everything, but the cleanup was a bit too agressive 😄

pkg/kubectl/cmd/cmd.go
```diff
@@ -116,31 +115,6 @@ __kubectl_override_flags()
     done
 }
 
-__kubectl_config_get_contexts()
-{
```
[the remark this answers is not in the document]
knight42: Thanks for pointing it out. Fixed.

pkg/kubectl/cmd/completion.go
```diff
@@ -0,0 +1,108 @@
...
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		config, err := f.ToRawKubeConfigLoader().RawConfig()
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
+		return filter(config, toComplete), cobra.ShellCompDirectiveNoFileComp
+	}
+}
+
+func registerCompletionFuncForGlobalFlags(cmds *cobra.Command, f util.Factory) error {
```
marckhouzam, line 42: Just a nit. Why `cmds` with an 's'?  It's just a single command, no? I'm not sure why the calling function uses `cmds` with an 's'.

pkg/kubectl/cmd/completion.go
```diff
@@ -0,0 +1,108 @@
...
+		for name := range config.AuthInfos {
+			if strings.HasPrefix(name, toComplete) {
+				ret = append(ret, name)
+			}
+		}
+		return ret
+	})
+
+	err = cmds.RegisterFlagCompletionFunc("context", getContextFromConfig)
+	if err != nil {
```
marckhouzam, line 95: I'm not familiar with `CheckErr` but would it be more instructive to call it in this function instead of returning the error?

marckhouzam (commented):
Looks nice!
Just a couple of coding suggestions, to take or leave.

pkg/kubectl/cmd/completion.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		return ret, cobra.ShellCompDirectiveNoFileComp
+	})
+	if err != nil {
+		return err
+	}
+
+	getContextFromConfig := getFromConfig(f, func(config clientcmdapi.Config, toComplete string) []string {
+		var ret []string
+		for name := range config.Contexts {
```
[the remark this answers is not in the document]
knight42: It is worth noting that `config.Contexts` is a map. If I updated the code as you said, we would have to iterate twice, one for collecting and one for filtering. I am not sure whether it is better.

pkg/kubectl/cmd/completion.go
knight42: just copy-paste from the original code, we could trim the `s`.

pkg/kubectl/cmd/completion.go
knight42: Yes, I think you are right.

pkg/kubectl/cmd/completion.go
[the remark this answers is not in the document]
marckhouzam: Oh, I didn't realize that `config.Contexts` was a map.  I guess it's not worth it then

marckhouzam:
This is really great.
One final note I forgot to mention. Another advantage of using custom Go completions, is that it is now possible to write Go tests for the completion code.

What I do in helm is call the `__complete` command from the test and check the result.  For example, if you can trigger a test call to `kubectl __complete -n ""` you could verify the right namespaces are returned.

marckhouzam (approved):
I'll let the maintainers decide if this requires tests or not.
But the rest seems right to me.  Nice work!

pkg/kubectl/cmd/completion.go
```diff
@@ -0,0 +1,96 @@
...
+		config, err := f.ToRawKubeConfigLoader().RawConfig()
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
+		return filter(config, toComplete), cobra.ShellCompDirectiveNoFileComp
+	}
+}
+
+func registerCompletionFuncForGlobalFlags(cmd *cobra.Command, f util.Factory) {
+	util.CheckErr(cmd.RegisterFlagCompletionFunc(
```
brianpursley, line 43: First, I can't remember if I already asked this or not, but

why not...
```
getNamespaces := func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
	clientSet, err := f.KubernetesClientSet()
	if err != nil {
		return nil, cobra.ShellCompDirectiveDefault
	}
	ctx, cancel := context.WithTimeout(context.TODO(), time.Second*3)
	defer cancel()
	namespaces, err := clientSet.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
	if err != nil {
		return nil, cobra.ShellCompDirectiveDefault
	}
	var ret []string
	for _, ns := range namespaces.Items {
		if strings.HasPrefix(ns.Name, toComplete) {
			ret = append(ret, ns.Name)
		}
	}
	return ret, cobra.ShellCompDirectiveNoFileComp
})
```

and then...

```
util.CheckErr(cmd.RegisterFlagCompletionFunc("namespace", getNamespaces))
```

... like the other three?

---

But secondly, I wonder if these can just be regular functions instead of declared as local variables.

I think having them as regular functions will make it easier to write tests and be a little nicer from a code organization perspective.

pkg/kubectl/cmd/completion.go
```diff
@@ -0,0 +1,96 @@
...
+			var ret []string
+			for _, ns := range namespaces.Items {
+				if strings.HasPrefix(ns.Name, toComplete) {
+					ret = append(ret, ns.Name)
+				}
+			}
+			return ret, cobra.ShellCompDirectiveNoFileComp
+		}))
+
+	getContextFromConfig := getFromConfig(f, func(config clientcmdapi.Config, toComplete string) []string {
```
brianpursley, line 65: I don't understand the purpose of the `getFromConfig` function. 

Could it not just do these 4 lines of code instead of using getFromConfig to generate your function?
```
		config, err := f.ToRawKubeConfigLoader().RawConfig()
		if err != nil {
			return nil, cobra.ShellCompDirectiveNoFileComp
		}
```

I'm thinking something like this:
```
        getContextFromConfig := func(config clientcmdapi.Config, toComplete string) ([]string, cobra.ShellCompDirective) {
		var ret []string
		config, err := f.ToRawKubeConfigLoader().RawConfig()
		if err != nil {
			return nil, cobra.ShellCompDirectiveNoFileComp
		}
		for name := range config.Contexts {
			if strings.HasPrefix(name, toComplete) {
				ret = append(ret, name)
			}
		}
		return ret, cobra.ShellCompDirectiveNoFileComp
	}
```

It just doesn't seem to add much value to do it this way when there is such a little amount of code involved.

Can you explain your thoughts for why this abstraction is needed?

marckhouzam:
@knight42 how are things going for these Go completions?

knight42:
@marckhouzam Oh I think I have missed the comments from @brianpursley and just waiting for review. I am still working on it.

staging/src/k8s.io/kubectl/pkg/completion/utils.go
```diff
@@ -0,0 +1,90 @@
+/*
```
knight42, line 1: For reviewer: I move helper functions here as I think they might be reused for the completion of `kubectl config get-{contexts,clusters}`

knight42:
@brianpursley The original target of this PR is to only implement custom completions using Go for flag --context, --cluster, --user and --namespace. I think it is completed now.

>  but what else would need to be implemented in order to get rid of the custom bash and zsh code in these two locations?

We need to modify each command as described in https://github.com/spf13/cobra/blob/master/shell_completions.md#dynamic-completion-of-nouns. I plan to open separate PR for each command, perhaps we could open a tracking issue and call for community participation.

marckhouzam:
I feel this would have value to merge as is.  

It will work for both bash and zsh and is a step in the right direction to get rid of the custom bash code.

And it will work automatically for fish completion if/when it is added (https://github.com/kubernetes/kubernetes/pull/92989)

brianpursley:
I'm not trying to be difficult, but I am hesitant to merge this because it does not fully address the issue.  I don't really like the idea of having some of the completions in go, but others still using custom scripts.  I was hoping to see a PR that fully addresses the issue.

However, if someone with the ability to approve PRs is comfortable with doing it though, then I will go along with their opinion on this, as I don't want to make this a stocking point if everyone else sees things differently.

Either way, we need to make 100% sure this doesn't break existing completions for zsh and bash.  This probably needs some manual confirmation as I don't think we have a way to test completions using integration tests.  Also, when go completions are created, we should be able to have go unit tests for those now, so that might help too.  

My number one concern is not to break existing completions which I know tool a long time and a lot of work to get working satisfactorily.

marckhouzam:
@brianpursley I agree with your concerns about making sure this doesn't break anything.  That is why I personally like the gradual approach so as to make sure each change is safe.

That being said, you are also right that temporarily having half the completions in go and the other half in bash will still cause some confusion.

@knight42  are you hoping to continue this work for the rest of the custom completions?  I would like to help but I worry that my total lack of knowledge of the kubectl code base will prevent me from suggesting the correct go code for the completions.

marckhouzam:
@brianpursley I've built upon @knight42 work and removed all bash scripting in #96087.

knight42:
/close
in favor of #96087

PR #93720, opened 2020-08-05
test/e2e/storage: switch from watch to shared informer to handle watch restarts

dcbw:


test/e2e/storage/csi_mock_volume.go
```diff
@@ -875,49 +920,14 @@ var _ = utils.SIGDescribe("CSI mock volume", func() {
...
-							break loop
-						case watch.Error:
-							// Can this occur when the apiserver is under heavy load?
-							// If yes, then we should bail out of the test here early and
-							// skip further checks instead of treating it as a test failure.
-							framework.Failf("PVC watch failed prematurely: %v", event.Object)
-						}
+					case <-watchDone:
+						framework.Logf("PVC event handler finished")
+						break loop
```
aojea, line 930: do we need the for and the break now?

PR #40355, opened 2017-01-24
TaintController

gmarek:
```release-note
This PR adds a manager to NodeController that is responsible for removing Pods from Nodes tainted with NoExecute Taints. This feature is beta (as the rest of taints) and enabled by default. It's gated by controller-manager enable-taint-manager flag.
```

PR #93722, opened 2020-08-05
Do not evict pods which tolerate all NoExecute taints

### LINKED ISSUE #90794: Pods in Node are deleted even if the NoExecute taint that triggered deletion is removed
**What happened**:

A Pod was scheduled for deletion because a taint w/ effect `NoExecute` was observed in the Node the Pod was assigned to, and after the taint is removed from the Node, the Pod scheduled deletion isn't canceled.

After scheduling pod deletion, the *`NoExecuteTaintManager` only cancels pod deletion if all `NoExecute` taints are removed from the node*, including any taints that the user relies on for their use-cases (example below).

After the taint that triggered scheduling pod deletion is removed, `NoExecuteTaintManager` cancels the scheduled pod deletion.

- A node `MyNode` is tainted w/ `OnlyForMyUsage:NoExecute`, so any workloads already assigned to it which don't tolerate this taint are evicted.
- A pod `MyPod` is created w/ tolerations: `OnlyForMyUsage:NoExecute` w/ unspecified `tolerationSeconds` and `NotReady:NoExecute` w/ `tolerationSeconds: 300`.
- `MyPod` is assigned to `MyNode`.
- At some point in time after the previous steps took place, `MyNode` is tainted as `NotReady:NoExecute`.
- `NoExecuteTaintManager` gets an update event for `MyNode` and observes two `NoExecute` taints. It proceeds to calculate the minimum time (in seconds) that `MyPod` tolerates for the two taints which at this point is 300 seconds, and marks `MyPod` for deletion in ~300 seconds - this happens in-memory, in a timed-queue.
- At some point in time after the previous steps took place, and before the 300 seconds trigger happen, `NotReady:NoExecute` taint is removed from `MyNode`.
-  `NoExecuteTaintManager` gets an update event for `MyNode` and observes only one `NoExecute` taint, `OnlyForMyUsage:NoExecute`. It proceeds to calculate the minimum time (in seconds) that `MyPod` tolerates for the this taint which is _infinity_, and returns, not canceling the previous deletion. It *completely ignores the fact that the taint that triggered `MyPod` deletion is no longer observed*.

- Worked w/ @bobveznat to figure this out after all pods in a node got evicted without apparent reason.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.17.3 (but looking at the code seems to affect all of 1.17.x and 1.18.x and master.
- OS (e.g: `cat /etc/os-release`): N/A
- Kernel (e.g. `uname -a`): 5.x
- Install tools: N/A
- Others: N/A

cc @gmarek @bowei @k82cn (owners of this code)

liggitt:
**What type of PR is this?**

/kind bug

Fixes #90794

```release-note
Fixes a bug evicting pods after a taint with a limited tolerationSeconds toleration is removed from a node
```

/sig node
/priority important-soon

### DIFF
--- a/pkg/controller/nodelifecycle/scheduler/taint_manager.go
+++ b/pkg/controller/nodelifecycle/scheduler/taint_manager.go
@@ -358,7 +358,8 @@ func (tc *NoExecuteTaintManager) processPodOnNode(
 	minTolerationTime := getMinTolerationTime(usedTolerations)
 	// getMinTolerationTime returns negative value to denote infinite toleration.
 	if minTolerationTime < 0 {
-		klog.V(4).Infof("New tolerations for %v tolerate forever. Scheduled deletion won't be cancelled if already scheduled.", podNamespacedName.String())
+		klog.V(4).Infof("Current tolerations for %v tolerate forever, cancelling any scheduled deletion.", podNamespacedName.String())
+		tc.cancelWorkWithEvent(podNamespacedName)
 		return
 	}
 
--- a/pkg/controller/nodelifecycle/scheduler/BUILD
+++ b/pkg/controller/nodelifecycle/scheduler/BUILD
@@ -42,6 +42,7 @@ go_test(
         "//staging/src/k8s.io/apimachinery/pkg/apis/meta/v1:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/fields:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/labels:go_default_library",
+        "//staging/src/k8s.io/apimachinery/pkg/types:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/util/sets:go_default_library",
         "//staging/src/k8s.io/client-go/kubernetes/fake:go_default_library",
         "//staging/src/k8s.io/client-go/testing:go_default_library",
--- a/pkg/controller/nodelifecycle/scheduler/taint_manager_test.go
+++ b/pkg/controller/nodelifecycle/scheduler/taint_manager_test.go
@@ -27,6 +27,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/fake"
 	"k8s.io/kubernetes/pkg/controller/testutil"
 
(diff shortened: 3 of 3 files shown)


pkg/controller/nodelifecycle/scheduler/taint_manager.go
```diff
@@ -358,7 +358,8 @@ func (tc *NoExecuteTaintManager) processPodOnNode(
 	minTolerationTime := getMinTolerationTime(usedTolerations)
 	// getMinTolerationTime returns negative value to denote infinite toleration.
 	if minTolerationTime < 0 {
-		klog.V(4).Infof("New tolerations for %v tolerate forever. Scheduled deletion won't be cancelled if already scheduled.", podNamespacedName.String())
+		klog.V(4).Infof("Current tolerations for %v tolerate forever, cancelling any scheduled deletion.", podNamespacedName.String())
+		tc.cancelWorkWithEvent(podNamespacedName)
```
liggitt, line 362: I'd like special eyes on this change... the logged message explicitly called out the problematic behavior, so it seems like it was known, but it still seems incorrect to me.

If a pod infinitely tolerates all remaining taints on the node, I don't see why that should have different behavior than a node with no taints (for which we call cancelWorkWithEvent on line 348)

liggitt: original commit in https://github.com/kubernetes/kubernetes/pull/40355/files#diff-12d91ddb84ebd95c1ccaa6b5980108f6R337-R341, comments by @davidopp at https://github.com/kubernetes/kubernetes/pull/40355/files#r100682445

> eviction that has not yet started gets moved later or canceled (e.g. eviction is added to the eviction queue, then pod is updated with larger tolerationSeconds for its soonest toleration, or all taints are removed from node) -- IIUC we do want to move the eviction later (or cancel it if removing all taints) in that case?

I think "removing all taints" behavior should be equivalent to "removing all taints except ones that are tolerated infinitely"

dashpole:
I've been trying to think of why the previous behavior would be desirable, but can't come up with anything.  +1 to being consistent.
/lgtm

liggitt:
My inclination would be to include this in 1.19 and pick it back to supported releases, because:
1. it is a correctness issue that can result in incorrect deletion of workloads
2. the fix is very contained

I would like feedback from https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/nodelifecycle/OWNERS (though I'm not sure that full list is up to date)

dashpole:
maybe @k82cn has thoughts?

sjenning:
@ravisantoshgudimetla you worked on the taint-based eviction controller back in the day right (like... a year ago)?  You see anything wrong with this?

lgtm though

liggitt:
Opened picks to 1.16-1.18 to start exercising CI.

k82cn:
Just went through the diff, LGTM overall, thanks very much 👍

Also agree to cherry-pick this PR :)

liggitt:
Sounds good, thanks for reviewing

/milestone v1.19

PR #93725, opened 2020-08-05
Fix races in DaemonSet TestExpectationsOnRecreate

### LINKED ISSUE #93604: Daemonset TestExpectationsOnRecreate fails consistently
**Which jobs are failing**:

TestExpectationsOnRecreate unit test, added in https://github.com/kubernetes/kubernetes/pull/91915 along with expectations bugfix during 1.19 and cherry-picked to 1.18/1.17

`go test -race -count=10 ./pkg/controller/daemon -run TestExpectationsOnRecreate`

**Reason for failure**:

```
--- FAIL: TestExpectationsOnRecreate (0.21s)
    daemon_controller_test.go:548: Unexpected item in the queue
```

```
--- FAIL: TestExpectationsOnRecreate (0.22s)
    daemon_controller_test.go:569: There should be no expectations for DaemonSet "default/test" after it was deleted
    daemon_controller_test.go:620: No expectations found for DaemonSet "default/test"
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x293b910]

goroutine 60 [running]:
testing.tRunner.func1.1(0x2c29a00, 0x3b02a50)
	/Users/liggitt/.gvm/gos/go1.15rc1/src/testing/testing.go:1057 +0x46a
testing.tRunner.func1(0xc000551c80)
	/Users/liggitt/.gvm/gos/go1.15rc1/src/testing/testing.go:1060 +0x636
panic(0x2c29a00, 0x3b02a50)
	/Users/liggitt/.gvm/gos/go1.15rc1/src/runtime/panic.go:975 +0x3e9
k8s.io/kubernetes/pkg/controller.(*ControlleeExpectations).Fulfilled(0x0, 0x2e22477)
	/Users/liggitt/go/src/k8s.io/kubernetes/pkg/controller/controller_utils.go:284 +0x30
k8s.io/kubernetes/pkg/controller/daemon.TestExpectationsOnRecreate(0xc000551c80)
	/Users/liggitt/go/src/k8s.io/kubernetes/pkg/controller/daemon/daemon_controller_test.go:622 +0x2379
testing.tRunner(0xc000551c80, 0x2ea96e0)
	/Users/liggitt/.gvm/gos/go1.15rc1/src/testing/testing.go:1108 +0x203
created by testing.(*T).Run
	/Users/liggitt/.gvm/gos/go1.15rc1/src/testing/testing.go:1159 +0x797
```

These failures are masked in CI because bazel test runs with flake retries (https://github.com/kubernetes/kubernetes/blob/master/build/root/.bazelrc#L32), so the test can fail 2/3 times and the job can still pass (cc @BenTheElder)

tnozicka:
**What type of PR is this?**
/kind flake

Fixes https://github.com/kubernetes/kubernetes/issues/93604

alternative to https://github.com/kubernetes/kubernetes/pull/93617 based on @liggitt's findings

```release-note
None
```

/sig apps
/cc @liggitt

### DIFF
--- a/pkg/controller/daemon/daemon_controller_test.go
+++ b/pkg/controller/daemon/daemon_controller_test.go
@@ -28,6 +28,7 @@ import (
 
 	apps "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


liggitt:
I'm not a fan of depending on the side effects of lister/informer ordering

tnozicka:
> I'm not a fan of depending on the side effects of lister/informer ordering
yeah, it's just smarter sleep, maybe dummy would be better.
Is there a better way to achieve the same?

liggitt:
> yeah, it's just smarter sleep, maybe dummy would be better.
> 
> Is there a better way to achieve the same?

not that I can think of without exposing a lot more internal state

tnozicka:
/close
in favour of https://github.com/kubernetes/kubernetes/pull/93617

PR #93730, opened 2020-08-05
scheduler_perf/MixedSchedulingBasePod: label nodes for pod affinity

### LINKED ISSUE #93729: scheduler_perf: MixedSchedulingBasePod does not finish
**What happened**: The `BenchmarkPerfScheduling/MixedSchedulingBasePod` performance benchmark does not finish even after 10 minutes of waiting. This is because the nodes do not have a `failure-domain.beta.kubernetes.io/zone` label set but the pod affinity rules in the benchmark's pod spec expects one:

https://github.com/kubernetes/kubernetes/blob/382107e6c84374b229e6188207ef026621286aa2/test/integration/scheduler_perf/config/pod-with-pod-affinity.yaml#L14

**What you expected to happen**: The test should finish.

```bash
$ make test-integration WHAT=./test/integration/scheduler_perf KUBE_TEST_VMODULE="''" KUBE_TEST_ARGS="-alsologtostderr=false -logtostderr=false -run=^$$ -benchtime=1ns -bench=BenchmarkPerfScheduling/MixedSchedulingBasePod
```

adtac:
**What type of PR is this?**

/kind bug
/kind failing-test

**What this PR does / why we need it**: the `MixedSchedulingBasePod` test was failing due to the lack of a node label. Also, the `failure-domain.beta.kubernetes.io/zone` label has been deprecated in favour of `topology.kubernetes.io/zone`. 

Fixes https://github.com/kubernetes/kubernetes/issues/93729

```release-note
NONE
```

```docs
```

/sig scheduling
/cc @alculquicondor

### DIFF
--- a/test/integration/scheduler_perf/config/performance-config.yaml
+++ b/test/integration/scheduler_perf/config/performance-config.yaml
@@ -212,6 +212,11 @@
     numPodsToSchedule: 2000
 - template:
     desc: MixedSchedulingBasePod
+    nodes:
+      nodeTemplatePath: config/node-default.yaml
+      labelNodePrepareStrategy:
+        labelKey: "topology.kubernetes.io/zone"
+        labelValues: ["zone1"]
     initPods:
     - podTemplatePath: config/pod-default.yaml
     - podTemplatePath: config/pod-with-pod-affinity.yaml


test/integration/scheduler_perf/config/performance-config.yaml
```diff
@@ -212,6 +212,11 @@
     numPodsToSchedule: 2000
 - template:
     desc: MixedSchedulingBasePod
+    nodes:
+      nodeTemplatePath: config/node-default.yaml
+      labelNodePrepareStrategy:
+        labelKey: "topology.kubernetes.io/zone"
+        labelValues: ["zone1"]
```
alculquicondor, line 219: could you add a few more zones? If I remember correctly, the strategy supports it.

adtac: good point, looks like the nodes are assigned a label value based on round robin.

changed all zone-based topology labelling to 3 zones

alculquicondor:
/approve

please squash. Also I'm assuming you did manual runs :)

adtac:
>Also I'm assuming you did manual runs

I did, but as expected due to the difference in the number of zones, some percentiles are a bit different.

alculquicondor:
not sure if it's worth including this in the milestone. But if anything, it depends on #93733 being included as well

adtac:
Other way around, https://github.com/kubernetes/kubernetes/pull/93733 indirectly depends on this being merged first. Otherwise the tests will never finish (due to the mixed pods case waiting forever).

alculquicondor:
I see
@liggitt wdyt?

test/integration/scheduler_perf/config/performance-config.yaml
```diff
@@ -212,6 +212,11 @@
     numPodsToSchedule: 2000
 - template:
     desc: MixedSchedulingBasePod
+    nodes:
+      nodeTemplatePath: config/node-default.yaml
+      labelNodePrepareStrategy:
```
liggitt, line 217: adding labelNodePrepareStrategy to MixedSchedulingBasePod so the test can complete makes sense. if we want to merge this in 1.19, should we leave the other bits alone for now (continue using `failure-domain.beta.kubernetes.io/zone`, just have a single `zone1` value in SchedulingNodeAffinity, etc)?

adtac: Done.

adtac:
/cc @liggitt 

The do-not-merge and deprecation labels were added by mistake FWIW.

PR #93732, opened 2020-08-05
Update smd to v4

apelisse:
Update SMD to use an actual tag. I've also created the branch and everything so that it's ready if we have to make a cherry-pick for that in the future.

Note that it's using 4.0.1 because I initially pushed 4.0.0 to the wrong commit, and then it looks like something is unhappy if you force a tag :-)

Beside that, it should be a no-op.

**What type of PR is this?**

/kind cleanup

```release-note
NONE
```


go.mod
```diff
@@ -498,6 +498,7 @@ replace (
 	sigs.k8s.io/apiserver-network-proxy/konnectivity-client => sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.9
 	sigs.k8s.io/kustomize => sigs.k8s.io/kustomize v2.0.3+incompatible
 	sigs.k8s.io/structured-merge-diff/v3 => sigs.k8s.io/structured-merge-diff/v3 v3.0.1-0.20200706213357-43c19bbb7fba
+	sigs.k8s.io/structured-merge-diff/v4 => sigs.k8s.io/structured-merge-diff/v4 v4.0.1
```
apelisse, line 501: There must be something wrong if I still have v3 above.

apelisse: Yeah, looks like I missed a few files

liggitt:
Things targeting 1.19 should still go to master, it hasn't reopened for 1.20 yet

liggitt:
Closing in favor of https://github.com/kubernetes/kubernetes/pull/93867

PR #93737, opened 2020-08-05
Restart logging agent if installed on windows node.

barney-s:
**What type of PR is this?**
/kind bug

Fixes logging agent failure to connect to metadata server on startup.

This was present in earlier releases and was removed via https://github.com/kubernetes/kubernetes/commit/0239a7d85779a9cbbf4ceb81ecfdee927aa05ecf
Reverting a part of an earlier commit that removed this line.

NONE

### DIFF
--- a/cluster/gce/windows/k8s-node-setup.psm1
+++ b/cluster/gce/windows/k8s-node-setup.psm1
@@ -1623,6 +1623,10 @@ function Install-LoggingAgent {
     # installer binary and searched for flags to do this but found nothing. Oh
     # well.
     Log-Output ("Skip: Stackdriver logging agent is already installed")
+
+    # Restarting logging agent forces it to reconnect with the metadata 
+    # server. This makes the startup more reliable.
+    Restart-LoggingAgent
     return
   }
   


cluster/gce/windows/k8s-node-setup.psm1
```diff
@@ -1623,6 +1623,10 @@ function Install-LoggingAgent {
     # installer binary and searched for flags to do this but found nothing. Oh
     # well.
     Log-Output ("Skip: Stackdriver logging agent is already installed")
+
+    # Restarting logging agent forces it to reconnect with the metadata 
+    # server. This makes the startup more reliable.
+    Restart-LoggingAgent
```
jeremyje, line 1629: I think it'd be better if stackdriverlogger and friends were set to automatically restart.

```bash
sc failure /?
DESCRIPTION:
        Changes the actions upon failure
USAGE:
        sc <server> failure [service name] <option1> <option2>...

OPTIONS:
        reset=   <Length of period of no failures (in seconds)
                  after which to reset the failure count to 0 (may be INFINITE)>
                  (Must be used in conjunction with actions= )
        reboot=  <Message broadcast before rebooting on failure>
        command= <Command line to be run on failure>
        actions= <Failure actions and their delay time (in milliseconds),
                  separated by / (forward slash) -- e.g., run/5000/reboot/800
                  Valid actions are <run|restart|reboot> >
                  (Must be used in conjunction with the reset= option)
```

pjh: Good point - I filed a bug to track setting the logging service to auto-restart.

pjh:
Based on offline discussion, I don't think we need this change.

cluster/gce/windows/k8s-node-setup.psm1
jeremyje: Further look at the code the caller of this method actually invokes `Restart-LoggingAgent` afterwards.

```powershell
# Even if Stackdriver is already installed, the function will still [re]start the service.
  if (IsLoggingEnabled $kube_env) {
    Install-LoggingAgent
    Configure-LoggingAgent
    Restart-LoggingAgent
  }
```

PR #93738, opened 2020-08-05
check plugin supports expand before handing over to CSI

### LINKED ISSUE #93727: Volume expansion request should not reach to CSI driver, if in-tree plugin is not capable of expanding volume
**What happened**:
When the CSI migration is enabled, Kubernetes is sending a request to expand the volume to CSI even when the in-tree plugin is not supporting the volume resize.

When the migration is enabled, Kubernetes should not send a request to expand the volume to CSI when in-tree plugin is not capable of extending volume.

1. Enable vSphere CSI Migration.
2. Create PVC with StorageClass having `AllowVolumeExpansion` set to true using provisioner - `kubernetes.io/vsphere-volume`.
3. Resize PVC.
4. expand volume request is reaching to CSI controller.

We do not have any interface to check this and fail operation at the translation layer.

We can update the logic in Kubernetes to help verify this and stop volume expansion there - https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/expand/expand_controller.go#L258-L293

**Why do we want this?**
If the in-tree plugin is not capable of expanding volume, but CSI driver is capable to expand volume.

We see a race when CSI driver expands the volume, but before the file system is resized, if user disables the migration and move back to the in-tree plugin, file-system of the volume can never be expanded.

also, migration should not be adding new capabilities on migrated volumes.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.19 rc
- Cloud provider or hardware configuration: vSphere Cloud Provider

cc: @msau42 @xing-yang @SandeepPissay @misterikkit @chethanv28

divyenpatel:
**What type of PR is this?**
/kind bug

 When the migration is enabled, Kubernetes should not send a request to expand the volume to CSI when the in-tree plugin is not capable of extending volume.

This PR helps check if in-tree volume plugin supports expand, if not then does not hand over the expand to CSI when migration is enabled.

This helps prevent the situation when CSI driver expands the volume, but before the file system is resized, if user disables the migration and move back to the in-tree plugin, the file-system of the volume can never be expanded by in-tree plugin as it does not support expand volume feature.

Fixes https://github.com/kubernetes/kubernetes/issues/93727

Testing Done.

Verified when an in-tree plugin does not support the `expand volume`, the volume can not be expanded.
and the following message is emitted as event on PVC

> ignoring the PVC: didn't find a plugin capable of expanding the volume.

```
# kubectl describe pvc
Name:          vcppvc
Namespace:     default
StorageClass:  vcpsc
Status:        Bound
Volume:        pvc-10073eac-ee6c-40ae-9f89-db5fe502e1a2
Labels:        <none>
Annotations:   pv.kubernetes.io/bind-completed: yes
               pv.kubernetes.io/bound-by-controller: yes
               volume.beta.kubernetes.io/storage-provisioner: csi.vsphere.vmware.com
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:      2Gi
Access Modes:  RWO
VolumeMode:    Filesystem
Mounted By:    <none>
Events:
  Type    Reason                 Age                    From                                                                                                 Message
  ----    ------                 ----                   ----                                                                                                 -------
  Normal  Provisioning           2m37s                  csi.vsphere.vmware.com_vsphere-csi-controller-596dd6f757-2tg2h_60a59faa-2e77-44a8-bda6-4c18af7b7ee4  External provisioner is provisioning volume for claim "default/vcppvc"
  Normal  ExternalProvisioning   2m35s (x2 over 2m37s)  persistentvolume-controller                                                                          waiting for a volume to be created, either by external provisioner "csi.vsphere.vmware.com" or manually created by system administrator
  Normal  ProvisioningSucceeded  2m32s                  csi.vsphere.vmware.com_vsphere-csi-controller-596dd6f757-2tg2h_60a59faa-2e77-44a8-bda6-4c18af7b7ee4  Successfully provisioned volume pvc-10073eac-ee6c-40ae-9f89-db5fe502e1a2
  Normal  ExternalExpanding      119s                   volume_expand                                                                                        Ignoring the PVC: didn't find a plugin capable of expanding the volume.
```

```release-note
check in-tree plugin supports expand before handing over to CSI
```

cc: @msau42 @xing-yang @SandeepPissay @misterikkit @chethanv28

### DIFF
--- a/pkg/controller/volume/expand/expand_controller.go
+++ b/pkg/controller/volume/expand/expand_controller.go
@@ -250,12 +250,25 @@ func (expc *expandController) syncHandler(key string) error {
 
 	volumeResizerName := class.Provisioner
 	volumeSpec := volume.NewSpecFromPersistentVolume(pv, false)
+
+	volumePlugin, err := expc.volumePluginMgr.FindExpandablePluginBySpec(volumeSpec)
+	if err != nil || volumePlugin == nil {
+		msg := fmt.Errorf("didn't find a plugin capable of expanding the volume")
+		eventType := v1.EventTypeNormal
+		if err != nil {
+			eventType = v1.EventTypeWarning
+		}
+		expc.recorder.Event(pvc, eventType, events.ExternalExpanding, fmt.Sprintf("Ignoring the PVC: %v.", msg))
+		klog.Infof("Ignoring the PVC %q (uid: %q) : %v.", util.GetPersistentVolumeClaimQualifiedName(pvc), pvc.UID, msg)
+		return nil
+	}
+
 	migratable, err := expc.csiMigratedPluginManager.IsMigratable(volumeSpec)
 	if err != nil {
 		klog.V(4).Infof("failed to check CSI migration status for PVC: %s with error: %v", util.ClaimToClaimKey(pvc), err)
 		return nil
 	}
-	// handle CSI migration scenarios before invoking FindExpandablePluginBySpec for in-tree
+	// handle CSI migration scenario
 	if migratable {
 		msg := fmt.Sprintf("CSI migration enabled for %s; waiting for external resizer to expand the pvc", volumeResizerName)
 		expc.recorder.Event(pvc, v1.EventTypeNormal, events.ExternalExpanding, msg)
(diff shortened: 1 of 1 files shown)


liggitt:
/hold

Would like to resolve https://github.com/kubernetes/kubernetes/issues/93727#issuecomment-669604193 before doing special case handling of some CSI volumes contrary to what their storage class claims is supported

divyenpatel:
closing as per https://github.com/kubernetes/kubernetes/issues/93727#issuecomment-670561039

PR #93745, opened 2020-08-06
clean up naming

### LINKED ISSUE #93744: scheduler PreFilter NodePorts plugins  clean up getContainerPorts Loop  naming
When I learned more about the entire kubernetes scheduler frame from the source code.

I noticed that this is the getContainerPort function of nodeport plugins in the PreFilter stage.

This is the original code as following as bellowing.It is seems to me  that the naming and use of for loops is not very good for reading 
```go
func getContainerPorts(pods ...*v1.Pod) []*v1.ContainerPort {
	ports := []*v1.ContainerPort{}
	for _, pod := range pods {
		for j := range pod.Spec.Containers {
			container := &pod.Spec.Containers[j]
			for k := range container.Ports {
				ports = append(ports, &container.Ports[k])
			}
		}
	}
	return ports
}
```
how about let's revise the naming as bellowing.
```go
func getContainerPorts(pods ...*v1.Pod) []*v1.ContainerPort {
	ports := []*v1.ContainerPort{}
	for _, pod := range pods {
		for _, container := range pod.Spec.Containers {
			for k := range container.Ports{
				ports = append(ports, &container.Ports[k])
			}
		}
	}
	return ports
}
```

jjmengze:
**What type of PR is this?**
/kind cleanup

more clear code for reading 

Fixes #93744 

```release-note
none
```

```release-note
none
```

### DIFF
--- a/pkg/scheduler/framework/plugins/nodeports/node_ports.go
+++ b/pkg/scheduler/framework/plugins/nodeports/node_ports.go
@@ -61,9 +61,8 @@ func (pl *NodePorts) Name() string {
 func getContainerPorts(pods ...*v1.Pod) []*v1.ContainerPort {
 	ports := []*v1.ContainerPort{}
 	for _, pod := range pods {
-		for j := range pod.Spec.Containers {
-			container := &pod.Spec.Containers[j]
-			for k := range container.Ports {
+		for _, container := range pod.Spec.Containers {
+			for k := range container.Ports{
 				ports = append(ports, &container.Ports[k])
 			}
 		}


pkg/scheduler/framework/plugins/nodeports/node_ports.go
```diff
@@ -61,9 +61,8 @@ func (pl *NodePorts) Name() string {
 func getContainerPorts(pods ...*v1.Pod) []*v1.ContainerPort {
 	ports := []*v1.ContainerPort{}
 	for _, pod := range pods {
-		for j := range pod.Spec.Containers {
-			container := &pod.Spec.Containers[j]
-			for k := range container.Ports {
+		for _, container := range pod.Spec.Containers {
+			for k := range container.Ports{
```
xq2005, line 65: Can use the same way to clean the port name.

adtac: Since this appends a reference to the port, doint `for _, port := range container.Ports` and then appending `&port` would create a new reference, not a reference to the `container.Port` array element. As a result, changes to elements in `ports` would not get reflected in the original `pod.Spec.Containers` (but the in-tree approach would). So we should continue using the array index here.

adtac: @jjmengze but run go fmt please.

Huang-Wei: @adtac is right, we should be very cautious when appending a pointer inside looping over a slice with concrete types (in this case, []ContainerPort); otherwise, you may end up with storing the same pointer multiple times in the result.

pkg/scheduler/framework/plugins/nodeports/node_ports.go
```diff
@@ -61,9 +61,8 @@ func (pl *NodePorts) Name() string {
 func getContainerPorts(pods ...*v1.Pod) []*v1.ContainerPort {
 	ports := []*v1.ContainerPort{}
 	for _, pod := range pods {
-		for j := range pod.Spec.Containers {
-			container := &pod.Spec.Containers[j]
-			for k := range container.Ports {
+		for _, container := range pod.Spec.Containers {
```
Huang-Wei, line 64: Although it's basically stylish difference, I still prefer the old style as the new style would generate a copy of `Container` (unlike pods, pods is a slice with Pod pointer) each time, and `Container` has a ton of fields inside.

pkg/scheduler/framework/plugins/nodeports/node_ports.go
jjmengze: Thank you very much for your generous response

But I have some piece of question about for each in this scenario.

1. we copy pod form the pods slices (so let's get the new pointer of pod that is difference of original), the rest of the operations seem to be performed on this new pointer.It is seems to me that in this case we are not really care about it original pointer .

2.At next stage (filter stage), nodeport plugins take out the cycleState meanwhile 
translate it into preFilterState.

Through the fitsPorts function to compare the conflict situation, at this stage this plugins used for each again, and it seems to get a new pointer again. I am not sure about that the original pointer is needed be care.

Huang-Wei: > It is seems to me that in this case we are not really care about it original pointer .

Yes, but the new pointer still holds the same API object.

> I am not sure about that the original pointer is needed be care.

As long as it's a pointer, then copying is safe and cheap (comparing of copying a large object), so we don't care too much.

jjmengze: @Huang-Wei 
From this point of view, it is always copying the pointer in this scenario, let's make the code cleaner.

jjmengze: > As long as it's a pointer, then copying is safe and cheap (comparing of copying a large object), so we don't care too much.

From the original code, the pod pointer was copied for the first time for range

For range operation pointer in the inner layer (the pointer copied in the last second)

The innermost for range performs the same operation as the original copied pointer

@Huang-Wei  Would you suggest that we don't  need to trade readability for performance?

Huang-Wei: I'm not sure I fully understand your concern here. The basic rule when performing `for ... range` is:

- If the slice is `[]T`, and `T` is a big object, `for i := range []T` is better than `for i, obj := range []T` as in the latter case, obj is a local copy.
- If the slice is `[]*T`, usually we don't care too much as copying a pointer is cheap. You may see a lot of code like `for _, pod := range []*Pod` in k8s codebase.

jjmengze: @Huang-Wei 
I really appreciate your warm response and I fully comprehend what you mean.

Sorry for that I don't notice that `pod.Spec.Containers` is not a pointer meanwhile it is a big object

PR #93758, opened 2020-08-06
[release-1.18] Add labelSelector validation to Affinity/AntiAffinity score plugin

damemi:
This is a 1.18-compatible backport of #93660.

The lack of this validation on incoming pods causes unpredictable cluster outcomes
when later calculating affinity results against existing pods (see #92714). This fix
quickly addresses the main source where these problems should be caught.

It is unfortunately difficult to add this validation directly to the API server due
to the fact that it may break migrations with existing pods that fail this check. This
is a compromise to address the current issue.

**What type of PR is this?**
/kind bug

Ref #92714

Fixes #

```release-note
Pod Affinity/AntiAffinity label selectors are now validated in the pod affinity score plugin
```

```docs

```

### DIFF
--- a/pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
+++ b/pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
@@ -203,14 +203,12 @@ func (pl *InterPodAffinity) PreScore(
 	var antiAffinityTerms []*weightedAffinityTerm
 	if hasAffinityConstraints {
 		if affinityTerms, err = getWeightedAffinityTerms(pod, affinity.PodAffinity.PreferredDuringSchedulingIgnoredDuringExecution); err != nil {
-			klog.Error(err)
-			return nil
+			return framework.NewStatus(framework.Error, fmt.Sprintf("processing affinity terms: %+v", err))
 		}
 	}
 	if hasAntiAffinityConstraints {
 		if antiAffinityTerms, err = getWeightedAffinityTerms(pod, affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution); err != nil {
-			klog.Error(err)
-			return nil
+			return framework.NewStatus(framework.Error, fmt.Sprintf("processing affinity terms: %+v", err))
 		}
 	}
 
--- a/pkg/scheduler/framework/plugins/interpodaffinity/filtering.go
+++ b/pkg/scheduler/framework/plugins/interpodaffinity/filtering.go
@@ -347,7 +347,7 @@ func (pl *InterPodAffinity) PreFilter(ctx context.Context, cycleState *framework
 	// incomingPodAntiAffinityMap will be used later for efficient check on incoming pod's anti-affinity
 	incomingPodAffinityMap, incomingPodAntiAffinityMap, err := getTPMapMatchingIncomingAffinityAntiAffinity(pod, allNodes)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


damemi:
@ahg-g @alculquicondor here is the 1.18 port. I don't think we need to add the validation to Filter in 1.18, because it looks like it already has it (when PreFilter [calculates the `incomingPodMap`s](https://github.com/kubernetes/kubernetes/blob/5aa896df6f911eb246079ec5d0a73d15afa5fe85/pkg/scheduler/framework/plugins/interpodaffinity/filtering.go#L348) it calls [`getAffinityTerms`](https://github.com/kubernetes/kubernetes/blob/5aa896df6f911eb246079ec5d0a73d15afa5fe85/pkg/scheduler/framework/plugins/interpodaffinity/filtering.go#L276-L284) which [do return the LabelSelector error](https://github.com/kubernetes/kubernetes/blob/5aa896df6f911eb246079ec5d0a73d15afa5fe85/pkg/scheduler/framework/plugins/interpodaffinity/filtering.go#L276-L284)). Scoring does need to check it

pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
```diff
@@ -203,14 +203,12 @@ func (pl *InterPodAffinity) PreScore(
 	var antiAffinityTerms []*weightedAffinityTerm
 	if hasAffinityConstraints {
 		if affinityTerms, err = getWeightedAffinityTerms(pod, affinity.PodAffinity.PreferredDuringSchedulingIgnoredDuringExecution); err != nil {
-			klog.Error(err)
-			return nil
+			framework.NewStatus(framework.Error, fmt.Sprintf("processing affinity terms: %+v", err))
```
alculquicondor, line 206: same question as the other PR. However, I'm not sure if the framework can handle UnschedulableAndUnresolvable properly in 1.18 (I think it just converts it to Error anyways). Could you check?

damemi: Good point, checked it out and responded in https://github.com/kubernetes/kubernetes/pull/93660#discussion_r466610419

>`UnschedulableAndUnresolvable` seems to only be parsed in Filter plugins, and the [error code definitions](https://github.com/kubernetes/kubernetes/blob/release-1.19/pkg/scheduler/framework/v1alpha1/interface.go#L59) imply that unexpected input could cause an error.
>
>So `UnschedulableAndUnresolvable` could work for the Filter plugin, but score just returns anything that's not success as an error (https://github.com/kubernetes/kubernetes/blob/release-1.19/pkg/scheduler/core/generic_scheduler.go#L512-L516) (also, in 1.18 [this case is already returned as a `framework.Error`](https://github.com/kubernetes/kubernetes/blob/release-1.18/pkg/scheduler/framework/plugins/interpodaffinity/filtering.go#L350) from the Filter plugin)
>
>I think in the case of a syntax error, `Error` makes sense. `UnschedulableAndUnresolvable` sounds more related to cluster state that prevents the pod from scheduling (ie, it fails the filters and nothing can be preempted to fit it)

damemi:
/hold
should add some tests

damemi:
/hold cancel
added unit tests

alculquicondor (commented):
/lgtm

ahg-g:
@kubernetes/release-managers

damemi:
@ahg-g could you re-lgtm this as well?

PR #93762, opened 2020-08-06
Fix kube-addon-manager overwriting resources with EnsureExists set

spencer-p:
The addon manager readme states

> - Addons with label `addonmanager.kubernetes.io/mode=EnsureExists` will be checked for
> existence only. Users can edit these addons as they want.

However, the start_addon function was using `kubectl apply` to create resources regardless of mode. Users who change EnsureExists resources have their configs wiped out and replaced when the addon manager restarts.

This change switches between `kubectl create` and `kubectl apply` depending on the addon's mode.

Noting this fix may have to be re-implemented after #92978.

/kind bug

```release-note
NONE
```

### DIFF
--- a/cluster/addons/addon-manager/kube-addons-main.sh
+++ b/cluster/addons/addon-manager/kube-addons-main.sh
@@ -1,0 +1,78 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Import required functions. The addon manager is installed to /opt in
+# production use (see the Dockerfile)
+# Disabling shellcheck following files as the full path would be required.
+if [ -f "kube-addons.sh" ]; then
+  # shellcheck disable=SC1091
+  source "kube-addons.sh"
+elif [ -f "/opt/kube-addons.sh" ]; then
+  # shellcheck disable=SC1091
+  source "/opt/kube-addons.sh"
+else
+  # If the required source is missing, we have to fail.
+  log ERR "== Could not find kube-addons.sh (not in working directory or /opt) at $(date -Is) =="
+  exit 1
+fi
+
+# The business logic for whether a given object should be created
+# was already enforced by salt, and /etc/kubernetes/addons is the
+# managed result of that. Start everything below that directory.
+log INFO "== Kubernetes addon manager started at $(date -Is) with ADDON_CHECK_INTERVAL_SEC=${ADDON_CHECK_INTERVAL_SEC} =="
+
+# Wait for the default service account to be created in the kube-system namespace.
+token_found=""
+while [ -z "${token_found}" ]; do
+  sleep .5
+  # shellcheck disable=SC2086
+  # Disabling because "${KUBECTL_OPTS}" needs to allow for expansion here
+  if ! token_found=$(${KUBECTL} ${KUBECTL_OPTS} get --namespace="${SYSTEM_NAMESPACE}" serviceaccount default -o go-template="{{with index .secrets 0}}{{.name}}{{end}}"); then
+    token_found="";
+    log WRN "== Error getting default service account, retry in 0.5 second =="
+  fi
+done
+
+log INFO "== Default service account in the ${SYSTEM_NAMESPACE} namespace has token ${token_found} =="
+
+# Create admission_control objects if defined before any other addon services. If the limits
+# are defined in a namespace other than default, we should still create the limits for the
+# default namespace.
+while IFS=$'\n' read -r obj; do
+  start_addon "${obj}" 100 10 default &
+  log INFO "++ obj ${obj} is created ++"
+done < <(find /etc/kubernetes/admission-controls \( -name \*.yaml -o -name \*.json \))
+
+# Start the apply loop.
+# Check if the configuration has changed recently - in case the user
+# created/updated/deleted the files on the master.
+log INFO "== Entering periodical apply loop at $(date -Is) =="
+while true; do
+  start_sec=$(date +"%s")
+  if is_leader; then
+    ensure_addons
+    reconcile_addons
+  else
+    log INFO "Not elected leader, going back to sleep."
+  fi
+  end_sec=$(date +"%s")
+  len_sec=$((end_sec-start_sec))
+  # subtract the time passed from the sleep time
+  if [[ ${len_sec} -lt ${ADDON_CHECK_INTERVAL_SEC} ]]; then
+    sleep_time=$((ADDON_CHECK_INTERVAL_SEC-len_sec))
+    sleep ${sleep_time}
+  fi
+done
--- a/cluster/addons/addon-manager/kube-addons-test.sh
+++ b/cluster/addons/addon-manager/kube-addons-test.sh
@@ -1,0 +1,286 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# These tests enforce behavior of kube-addon-manager functions against a real
+# cluster. A working Kubernetes cluster must be set up with kubectl configured.
+# To run with the released version of kubectl, use `make test`.
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+# Default kubectl to the test users installation if needed.
+KUBECTL_BIN="${KUBECTL_BIN:-kubectl}"
+
+# Disabling shellcheck following files as the full path would be required.
+# shellcheck disable=SC1091
+source "kube-addons.sh"
+
+TEST_NS="kube-addon-manager-test"
+
+function retry() {
+  local tries=10
+  while [ "${tries}" -gt 0 ]; do
+    "$@" && return 0;
+    (( tries-- ))
+    sleep 1
+  done
+}
+
+function setup(){
+  retry kubectl create namespace "${TEST_NS}"
+}
+
+function teardown() {
+  retry kubectl delete namespace "${TEST_NS}"
+}
+
+function error() {
+  echo -e "\e[31m$*\e[0m"
+}
+
+function echo_green() {
+  echo -e "\e[32m$*\e[0m"
+}
+
+function echo_blue() {
+  echo -e "\e[34m$*\e[0m"
+}
+
+function test_create_resource_reconcile() {
+  local limitrange
+  read -r -d '' limitrange << EOF
+apiVersion: "v1"
+kind: "LimitRange"
+metadata:
+  name: "limits"
+  namespace: "${TEST_NS}"
+  labels:
+    addonmanager.kubernetes.io/mode: Reconcile
+spec:
+  limits:
+    - type: "Container"
+      defaultRequest:
+        cpu: "100m"
+EOF
+
+  # arguments are yaml text, number of tries, delay, name of file, and namespace
+  echo_blue "Creating initial resource to test Reconcile mode"
+  create_resource_from_string "${limitrange}" "10" "1" "limitrange.yaml" "${TEST_NS}"
+  if ! (kubectl get limits/limits -n "${TEST_NS}"); then
+    error "failed to create limits w/ reconcile"
+    return 1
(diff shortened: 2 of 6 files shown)


MrHohn:
@spencer-p It looks like the `start_addon` func, in which `kubectl apply` is used regardless of mode, is only triggered for manifests stored in `/etc/kubernetes/admission-controls`. Is extending the EnsureExists/Reconcile logic to cover admission-control objects the intention here?

spencer-p:
@mrhohn Yes. This would affect `cluster/gce/addons/limit-range/limit-range.yaml` in particular.

spencer-p:
polite bump - review greatly appreciated as this issue has moderately high priority for users

MrHohn:
(Looks like the /assign command failed?)
/assign

spencer-p:
Looks like prow jumped the gun on that commit. Looks like a lot of container pull errors, likely because I hadn't pushed images to staging.
/retest

MrHohn:
> Looks like prow jumped the gun on that commit. Looks like a lot of container pull errors, likely because I hadn't pushed images to staging.

Yep, we generally do this in three steps:
1. PR to edit the script.
2. Publish a new image.
3. PR to bump the image version.

spencer-p:
/retest
Looks like the previous tests failed due to an SSL certificate and what seems to be a prow bug (test failed but 37/37 passed??)

MrHohn (approved):
Thanks for the fix @spencer-p! I don't have much comment - looks like reasonable changes to me. Have you built a test image and run it on a test cluster to be sure? (Very easy to make mistake when dealing with shell script..)

/approve

spencer-p:
Thanks @MrHohn. I have tested these changes in a cluster.

Here's the manual tests I've done. Assuming you're logged into the leader master node and the `LimitRange` config is in `/etc/kubernetes/admission-controls`:

### Creates missing resource
1. `kubectl delete limits/limits`
2. `crictl ps | grep addon | awk '{print $1}' | xargs crictl stop` (this command will kill and restart the addon-manager)
3. `kubectl get limits/limits -oyaml` should return the object

### Does not re-create existing resource
1. `EDITOR="sed -i 's/100m/600m/'" kubectl edit limits/limits`
2. Restart the addon manager
3. `kubectl get limits/limits -oyaml` should reflect the 600m.
4.  Checking the kube addon manager logs should show skipping the resource.

And then editing `/etc/kubernetes/admission-controls/limit-range/limit-range.yaml` to have `=Reconcile` instead of `=EnsureExists`. Then the results of the first test should be identical, but the second test should overwrite it.

These aren't complicated tests (though they are somewhat time based, esp. when the expected result is "no change").

So if anyone has thought or guidance on adding E2E tests, it would be much appreciated.

MrHohn:
Adding Jake as he is re-implementing this.
/cc @dekkagaijin

There is an e2e test that exercises addon manager functionalities: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/cloud/gcp/addon_update.go ([dashboard](https://k8s-testgrid.appspot.com/google-gce#gce-cos-master-slow&include-filter-by-regex=Addon))

However this new logic only applies to start up time so might be trickier to set up the test. I'm fine with the manual tests you have done.

Could you squash the follow-up commits? Thanks!

dekkagaijin (approved):
/lgtm

thanks for the heads up

MrHohn:
Thanks!
/lgtm

dekkagaijin:
I'll implement these changes in the [Golang port](https://github.com/kubernetes-sigs/cluster-addons/pull/76) when both PRs get approved

cluster/addons/addon-manager/kube-addons.sh
```diff
@@ -158,11 +158,25 @@ function create_resource_from_string() {
   local -r config_name=$4;
   local -r namespace=$5;
   while [ "${tries}" -gt 0 ]; do
+    # Resources that are set to the addon mode EnsureExists should not be overwritten if they already exist.
+    # EnsureExists means we should use `kubectl create`, other modes mean we should use `kubectl apply`.
+    local verb="apply"
+    if echo "${config_string}" | grep --silent "^\s*${ADDON_MANAGER_LABEL}:\s\+EnsureExists"; then
```
liggitt, line 164: /hold 

does this work if the string contains multiple objects, some with EnsureExists and some without?

why is a different approach being taken here than in ensure_addons/reconcile_addons (which use label selectors)

spencer-p: It wouldn't work, good point. There don't seem to be any multi-object files in testing.

I seemed to be under the impression that the label selectors wouldn't work as intended previously, and I don't have any notes on why I thought that was the case. I will note that this image is currently bundled with kubectl version 1.13.2.

I'll test a change with separate `kubectl apply`/`kubectl create` commands with label selectors.

spencer-p:
I've made some significant changes to account for @liggitt 's concerns about strings with multiple resources and label selectors.

In summary:
 - `kube-addons.sh` now only contains functions and definitions.
 - `kube-addons-main.sh` now contains the main entry point for the container and all the code removed from `kube-addons.sh`
 - `kube-addons-test.sh` is a script that runs integration tests for the affected portions of the codebase.
 - You can run the tests with `make test`, and they require a cluster to be set up and configured for `kubectl`. Using `make-test` will use the same kubectl version that will be packaged in the final image.
 - I factored out functions from `create_resource_from_string` to handle the subsets of EnsureExists resources and Reconcile resources. `create_resource_from_string` only succeeds when both inner functions succeed.

liggitt:
Thanks for the update, I don't have review bandwidth for this at the moment, but if it is consistent with the handling of labels on multiple resources, my objections are resolved.

I will note that a significant refactor/rewrite of bash scripts require exceptionally careful review to ensure equivalent behavior of pieces that were not intended to change. Once approvers/reviewers are satisfied this is correct, they can remove the hold at will.

cluster/addons/addon-manager/kube-addons-main.sh
```diff
@@ -0,0 +1,78 @@
...
+elif [ -f "/opt/kube-addons.sh" ]; then
+  # shellcheck disable=SC1091
+  source "/opt/kube-addons.sh"
+else
+  # If the required source is missing, we have to fail.
+  log ERR "== Could not find kube-addons.sh (not in working directory or /opt) at $(date -Is) =="
+  exit 1
+fi
+
+# The business logic for whether a given object should be created
```
MrHohn, line 32: Just double checking - I assumed nothing changed after this line?

cluster/addons/addon-manager/kube-addons-main.sh
```diff
@@ -0,0 +1,78 @@
...
+
+log INFO "== Default service account in the ${SYSTEM_NAMESPACE} namespace has token ${token_found} =="
+
+# Create admission_control objects if defined before any other addon services. If the limits
+# are defined in a namespace other than default, we should still create the limits for the
+# default namespace.
+while IFS=$'\n' read -r obj; do
+  start_addon "${obj}" 100 10 default &
+  log INFO "++ obj ${obj} is created ++"
+done < <(find /etc/kubernetes/admission-controls \( -name \*.yaml -o -name \*.json \))
```
MrHohn, line 57: Instead of figuring out every single yaml/json file and reconcile/ensure them one by one, I have been wondering whether a simple `kubectl apply/create -f` would work? My understanding is the command automatically ignores unrelated formats?

There are probably hidden complexity in the retry logic during master startup, please feel free to point out :)

cluster/addons/addon-manager/kube-addons.sh
```diff
@@ -170,6 +168,53 @@ function create_resource_from_string() {
...
+  local -r config_name=$2;
+  local -r namespace=$3;
+
+  # kubectl_output must be declared ahead of time to allow capturing kubectl's exit code and not local's exit code.
+  local kubectl_output;
+  # shellcheck disable=SC2086
+  # Disabling because "${KUBECTL_OPTS}" needs to allow for expansion here
+  kubectl_output=$(echo "${config_string}" | ${KUBECTL} ${KUBECTL_OPTS} apply -f - \
+    --namespace="${namespace}" -l ${ADDON_MANAGER_LABEL}=Reconcile 2>&1) && \
+      log INFO "== Successfully started ${config_name} in namespace ${namespace} at $(date -Is)" && \
```
MrHohn, line 182: Would be great to log this slightly differently so that we can distinguish ensure and reconcile? e.g. `Successfully reconciled ...`

cluster/addons/addon-manager/kube-addons-test.sh
```diff
@@ -0,0 +1,273 @@
...
+# limitations under the License.
+
+# These tests enforce behavior of kube-addon-manager functions against a real
+# cluster. A working Kubernetes cluster must be set up with kubectl configured.
+# To run with the released version of kubectl, use `make test`.
+
+set -o errexit
+set -o pipefail
+set -o nounset
+#set -o xtrace
```
MrHohn, line 24: Remove this line?

cluster/addons/addon-manager/kube-addons-test.sh
```diff
@@ -0,0 +1,273 @@
...
+    return 1
+  fi
+
+  # Changes to addons with mode reconcile should be reflected.
+  limitrange="${limitrange//100m/50m}"
+  create_resource_from_string "${limitrange}" "10" "1" "limitrange.yaml" "${TEST_NS}"
+  if kubectl get limits/limits -n ${TEST_NS} -oyaml | grep --silent "100m"; then
+    error "failed to update resource, still has 100m"
+    return 1
+  fi
```
MrHohn, line 96: Add a step that make sure it is 50m now?

cluster/addons/addon-manager/kube-addons-test.sh
```diff
@@ -0,0 +1,273 @@
...
+
+function error() {
+  echo -e "\e[31m$*\e[0m"
+}
+
+function echo_green() {
+  echo -e "\e[32m$*\e[0m"
+}
+
+function test_create_resource_reconcile() {
```
MrHohn, line 63: I find it easy to follow the exact test steps when reading the codes, but not as easy when executing it. Could we log necessary info about what the test is doing during the process?

cluster/addons/addon-manager/kube-addons-test.sh
```diff
@@ -0,0 +1,273 @@
...
+  fi
+  create_resource_from_string "${limitrange}" "10" "1" "limitrange.yaml" "${TEST_NS}"
+  if ! ( kubectl get limits/limits2 -n ${TEST_NS} -oyaml | grep --silent "50m"); then
+    error "failed to update resource, user config was respected when it should have been rewritten"
+    return 1
+  fi
+}
+
+function run() {
+  local -r name="${1}"
```
MrHohn, line 249: nit: name the variable something like `test_func` to be clear?

cluster/addons/addon-manager/kube-addons-test.sh
```diff
@@ -0,0 +1,273 @@
...
+
+  echo "TEST ${name}"
+  setup
+  if ! "${name}"; then
+    failures=$((failures+1))
+    teardown
```
MrHohn, line 255: nit: call `teardown` after the if block?

MrHohn (commented):
Left some comments as the first pass, thanks for adding the test!

cluster/addons/addon-manager/kube-addons.sh
```diff
@@ -170,6 +168,53 @@ function create_resource_from_string() {
   return 1;
 }
 
+function create_reconcile_resource_from_string() {
```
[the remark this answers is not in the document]
spencer-p: I'll rename these. I'll also note that the lack of `--prune` is not a regression - the original `create_resource_from_string` does not prune.

cluster/addons/addon-manager/kube-addons.sh
```diff
@@ -57,10 +57,10 @@ else
   read -ra KUBECTL_PRUNE_WHITELIST <<< "${KUBECTL_PRUNE_WHITELIST_OVERRIDE}"
 fi
 
-ADDON_CHECK_INTERVAL_SEC=${TEST_ADDON_CHECK_INTERVAL_SEC:-60}
+export ADDON_CHECK_INTERVAL_SEC=${TEST_ADDON_CHECK_INTERVAL_SEC:-60}
```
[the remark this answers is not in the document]
spencer-p: I actually added the export because shellcheck noticed that this variable wasn't used in this file, and so should either be deleted or exported. The export doesn't have any effect other than making the variable available to sub-processes, which shouldn't have any effect here.

It might be more readable to just `# shellcheck disable` for those variables.

cluster/addons/addon-manager/kube-addons-main.sh
spencer-p: Correct :)

cluster/addons/addon-manager/kube-addons-main.sh
spencer-p: I have no idea why this was written with a file walk. I think this one is out of scope for this PR.

MrHohn (approved):
Thanks! I don't spot any issue, and now it seems to handle the addon-manager labels in the same way as the rest of the script.

Could you squash the fix up commits?

spencer-p:
Rebase is done 👍

MrHohn:
Thanks!
/lgtm
/hold cancel
Un-holding this to move forward. After bumping the manifest, lets monitor the e2e tests (e.g. https://k8s-testgrid.appspot.com/google-gce#gce-cos-master-slow&include-filter-by-regex=Addon) closely to double check the behavior.

liggitt:
were the 9.1.2 images ever used? the manifest still shows 9.1.1 at HEAD:

https://github.com/kubernetes/kubernetes/blob/f48972e671e5f338d0d6adfe6055aba4534569ad/cluster/gce/manifests/kube-addon-manager.yaml#L26

dekkagaijin:
@liggitt doesn't seem so. I'm still planning on integrating this fix into the golang re-implementation and releasing that soonish

PR #93763, opened 2020-08-06
Fix flaky kubectl test

### LINKED ISSUE #92468: [Flaky Test] [sig-cli] Kubectl client Kubectl logs should be able to retrieve and filter logs [Conformance]
**Which jobs are flaking**:

`skew-cluster-latest-kubectl-stable1-gce (ci-kubernetes-e2e-gce-master-new-gci-kubectl-skew)`

`[sig-cli] Kubectl client Kubectl logs should be able to retrieve and filter logs [Conformance]`

**Testgrid link**:

https://testgrid.k8s.io/sig-release-master-blocking#skew-cluster-latest-kubectl-stable1-gce

**Reason for failure**:

```
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/framework/framework.go:721
Jun 24 13:50:35.009: Expected
    <int>: 2
to equal
    <int>: 1
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/kubectl/kubectl.go:1410
```

/sig cli
/cc @kubernetes/ci-signal
/priority important-soon
/milestone v1.19

eddiezane:
**What type of PR is this?**
/kind flake
/sig cli
/area kubectl

A race condition could occur where the two calls to `kubectl logs` in the `[sig-cli] Kubectl client Kubectl logs should be able to retrieve and filter logs [Conformance]` e2e test would not have enough logs generated between them.

This PR adds a delay between test calls to `kubectl logs` to ensure that enough logs have been generated.

Fixes #92468 

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/kubectl/kubectl.go
+++ b/test/e2e/kubectl/kubectl.go
@@ -1481,6 +1481,7 @@ metadata:
 			time.Sleep(2500 * time.Millisecond) // ensure that startup logs on the node are seen as older than 1s
 			recentOut := framework.RunKubectlOrDie(ns, "logs", podName, containerName, nsFlag, "--since=1s")
 			recent := len(strings.Split(recentOut, "\n"))
+			time.Sleep(2500 * time.Millisecond) // ensure that enough logs on the node are created past previous call
 			olderOut := framework.RunKubectlOrDie(ns, "logs", podName, containerName, nsFlag, "--since=24h")
 			older := len(strings.Split(olderOut, "\n"))
 			gomega.Expect(recent).To(gomega.BeNumerically("<", older), "expected recent(%v) to be less than older(%v)\nrecent lines:\n%v\nolder lines:\n%v\n", recent, older, recentOut, olderOut)


eddiezane:
/assign @apelisse 

cc @hasheddan

test/e2e/kubectl/kubectl.go
```diff
@@ -1481,6 +1481,7 @@ metadata:
 			time.Sleep(2500 * time.Millisecond) // ensure that startup logs on the node are seen as older than 1s
 			recentOut := framework.RunKubectlOrDie(ns, "logs", podName, containerName, nsFlag, "--since=1s")
 			recent := len(strings.Split(recentOut, "\n"))
+			time.Sleep(2500 * time.Millisecond) // ensure that enough logs on the node are created past previous call
```
hasheddan, line 1484: @eddiezane I don't think this is actually addressing the issue because it appears this is trying to test that there were more logs (i.e. the startup logs) written before the `recent`, so waiting longer after reading the `recent` logs and then checking the `older` is not testing that the startup logs are included in `older` but not `recent`. See above:
```
// ensure that startup logs on the node are seen as older than 1s
```

Do you read differently?

eddiezane: @hasheddan this whole test is confusing.

```
gomega.Expect(recent).To(gomega.BeNumerically("<", older), "expected recent(%v) to be less than older(%v)\nrecent lines:\n%v\nolder lines:\n%v\n", recent, older, recentOut, olderOut)
```

I believe it's asserting that there are less `recent` (logs created 1 second ago `--since=1s`) than there are `older` (logs created 24 hours ago `--since=24h`). Which should be the case. The flake was occurring when there isn't enough time between the `recent` and `older` (because they are both basically starting from 0 logs).

Thoughts?

hasheddan: Hmm shouldn't the first wait (https://github.com/kubernetes/kubernetes/pull/93763/files#diff-fde2ce0a73639d19347a5ea6fe001ba6R1481) take care of making it true that there are more logs in the last 24h than in the last 1s? This is why this flake is concerning to me because it _seems_ like the test logic is sound. If anything, it looks like we should bump the first wait to a longer period.

eddiezane: @hasheddan I _think_ that the test starts with 0 logs (the pod is spun up at the very start of this test) so that initial wait doesn't create more `older` logs. The test failures looked like they were because the two vars were equal at a very low number (like 5).

hasheddan: @eddiezane I believe the purpose of the first wait is to make there be older logs than 1s. So if the pod starts up and starts producing logs, then we wait 2.5 seconds, if any logs are produced in the first 1.5 seconds they should be in `older` but not in `recent`. Therefore, extending the first wait would increase the likelihood that logs are produced before the last `1s` that `recent` looks at.

eddiezane:
Closing in favor of #93920 and #93921

PR #93764, opened 2020-08-06
Adding flexibility to CCM

cici37:
**What type of PR is this?**
/kind feature

The Cloud Provider would be responsible for creating the outer CCM process which is https://github.com/kubernetes/kubernetes/blob/master/cmd/cloud-controller-manager/controller-manager.go and have the flexibility to specify the controllers and cloud provider config. 

Fixes #

```release-note
NONE
```

```docs

```


cmd/cloud-controller-manager/controller-manager.go
```diff
@@ -17,32 +17,66 @@ limitations under the License.
...
+const (
+	cloudProviderName = "SampleCloudProviderName"
+	cloudConfigFile   = "SampleCloudProviderConfigFile"
+)
+
 func main() {
 	rand.Seed(time.Now().UnixNano())
 
-	command := app.NewCloudControllerManagerCommand()
+	cloud, err := cloudprovider.InitCloudProvider(cloudProviderName, cloudConfigFile)
```
andrewsykim, line 49: I think we'd still want to call this after flag parsing since a provider will likely depend on flags for the cloud config file path.

andrewsykim: In fact, it would be good to update this reference to parse the cloud config file from the flag option.

cmd/cloud-controller-manager/app/controllermanager.go
```diff
@@ -54,7 +54,7 @@ const (
 )
 
 // NewCloudControllerManagerCommand creates a *cobra.Command object with default parameters
-func NewCloudControllerManagerCommand() *cobra.Command {
+func NewCloudControllerManagerCommand(newControllerInitializers map[string]InitFunc, cloud cloudprovider.Interface) *cobra.Command {
```
andrewsykim, line 57: s/newControllerInitializers/controllerInitializers

cmd/cloud-controller-manager/app/controllermanager.go
```diff
@@ -267,24 +259,24 @@ func startControllers(c *cloudcontrollerconfig.CompletedConfig, stopCh <-chan st
 	select {}
 }
 
-// initFunc is used to launch a particular controller.  It may run additional "should I activate checks".
+// InitFunc is used to launch a particular controller.  It may run additional "should I activate checks".
 // Any error returned will cause the controller process to `Fatal`
 // The bool indicates whether the controller was enabled.
-type initFunc func(ctx *cloudcontrollerconfig.CompletedConfig, cloud cloudprovider.Interface, stop <-chan struct{}) (debuggingHandler http.Handler, enabled bool, err error)
+type InitFunc func(ctx *cloudcontrollerconfig.CompletedConfig, cloud cloudprovider.Interface, stop <-chan struct{}) (debuggingHandler http.Handler, enabled bool, err error)
```
cheftako, line 265: I think this (ctx *cloudcontrollerconfig.CompletedConfig) is going to have linking problems when a cloud provider attempts to add a controller which is not in k8s.io/cloud-provider (?new home of cloudcontrollerconfig "k8s.io/kubernetes/cmd/cloud-controller-manager/app/config"?). It would be nice if we our sample added something like NodeIpamController to both see what it takes. It would also demonstrate how a cloud provider should handle config parameters which are not part of k8s.io/cloud-provider.

Also worth looking at https://github.com/kubernetes/kubernetes/blob/release-1.19/cmd/kube-controller-manager/app/controllermanager.go#L357. We seem to be diverging again and it would be nice if we could converge. Might be worth looking at if we can curry in the extra fields we need for the CCM.

cmd/cloud-controller-manager/controller-manager.go
cheftako: I think it might be nice to show a split example. Have 1 parameter (provider name) hard coded to demonstrate they can get rid of these flag options and the other parameter (cloud config file) which as Andrew suggests demonstrates how to consume flag options.

cici37:
The current commit is adding nodeipam controller in current mechanism. The next step would be:
1. Creating an cloud controller interface for future controller adding
2. Moving config generation process to `/cmd/cloud-controller-manager/controller-manager.go` so it would be easier to modify by cloud provider.
3. Copy ControllerContext(https://github.com/kubernetes/kubernetes/blob/70a682345e3bee5c7acff20b35ce8cdfbc9fdf11/cmd/kube-controller-manager/app/controllermanager.go#L303) under //cmd/controller-manager and use it in CCM(for carrying more field).

cmd/cloud-controller-manager/app/controllermanager.go
cici37: Thanks. Addressed in the latest changes.

cmd/controller-manager/app/config/types.go
```diff
@@ -0,0 +1,79 @@
...
+	// InformerFactory gives access to informers for the controller.
+	InformerFactory informers.SharedInformerFactory
+
+	// ObjectOrMetadataInformerFactory gives access to informers for typed resources
+	// and dynamic resources by their metadata. All generic controllers currently use
+	// object metadata - if a future controller needs access to the full object this
+	// would become GenericInformerFactory and take a dynamic client.
+	ObjectOrMetadataInformerFactory controller.InformerFactory
+
+	ComponentConfig ccmconfig.CloudControllerManagerConfiguration
```
cheftako, line 53: Can we get rid of this configuration? If we want to make this a generic controller manager ControllerContext then is shouldn't have anything CCM specific in it.

cici37: Thanks for pointing out. Updated

cmd/cloud-controller-manager/app/controllermanager.go
```diff
@@ -176,7 +173,35 @@ func Run(c *cloudcontrollerconfig.CompletedConfig, stopCh <-chan struct{}) error
 	}
 
 	run := func(ctx context.Context) {
-		if err := startControllers(c, ctx.Done(), cloud, newControllerInitializers()); err != nil {
+		rootClientBuilder := controller.SimpleControllerClientBuilder{
+			ClientConfig: c.Kubeconfig,
+		}
+		var clientBuilder controller.ControllerClientBuilder
+		if c.ComponentConfig.KubeCloudShared.UseServiceAccountCredentials {
```
liggitt, line 180: denormalized/copied/moved?

cmd/cloud-controller-manager/controller-manager.go
```diff
@@ -17,33 +17,282 @@ limitations under the License.
...
+	// add an controller which will be used by cloud provider
+	// generate nodeipamconfig. Here is an sample code. Please pass the right parameter in your code.
+	nodeipamconfig := nodeipamconfig.NodeIPAMControllerConfiguration{
+		ServiceCIDR:          "sample",
+		SecondaryServiceCIDR: "sample",
+		NodeCIDRMaskSize:     11,
+		NodeCIDRMaskSizeIPv4: 11,
+		NodeCIDRMaskSizeIPv6: 111,
+	}
+	controllerInitializers["nodeipam"] = startNodeIpamControllerWrapper(c.Complete(), nodeipamconfig, cloud)
```
liggitt, line 105: note to self: this is an example of what an override/extension could look like

cmd/cloud-controller-manager/controller-manager.go
```diff
@@ -17,33 +17,282 @@ limitations under the License.
...
+
+	cloud, err := cloudprovider.InitCloudProvider(cloudProviderName, *cloudProviderConfigFile)
+	if err != nil {
+		klog.Fatalf("Cloud provider could not be initialized: %v", err)
+	}
+	if cloud == nil {
+		klog.Fatalf("cloud provider is nil")
+	}
+
+	controllerList := []string{"cloud-node", "cloud-node-lifecycle", "service", "route"}
```
liggitt, line 81: note to self: this is an example of allow-listing specific controller loops

cmd/controller-manager/app/config/types.go
```diff
@@ -0,0 +1,76 @@
...
+	// requested.
+	RESTMapper *restmapper.DeferredDiscoveryRESTMapper
+
+	// AvailableResources is a map listing currently available resources
+	AvailableResources map[schema.GroupVersionResource]bool
+
+	// Control for which control loops to be run
+	// IncludeCloudLoops is for a kube-controller-manager running all loops
+	// ExternalLoops is for a kube-controller-manager running with a cloud-controller-manager
+	LoopMode ControllerLoopMode
```
liggitt, line 63: this probably doesn't work in generic

cici37:
Will update after https://github.com/kubernetes/kubernetes/pull/96016 is merged.

cici37:
@cheftako Would you mind to take a look when you have time? Thank you :)

staging/src/k8s.io/cloud-provider/app/testing/testserver.go
```diff
@@ -114,10 +115,17 @@ func StartTestServer(t Logger, customFlags []string) (result TestServer, err err
 	if err != nil {
 		return result, fmt.Errorf("failed to create config from options: %v", err)
 	}
+	cloud, err := cloudprovider.InitCloudProvider(config.Complete().ComponentConfig.KubeCloudShared.CloudProvider.Name, config.Complete().ComponentConfig.KubeCloudShared.CloudProvider.CloudConfigFile)
```
cheftako, line 118: This is a test of the controller manager/cloud provider code. I think we would be better of here using the fake cloud provider.

cici37: This test would also be called by `test/integration/serving/serving_test.go` so that we could not use a random fake value.

cmd/cloud-controller-manager/controller-manager.go
cheftako: I think a follow up PR which split out the base/vanilla version from the sample showing usage of the extensions would be a good idea. It makes it easier on cloud providers who can just copy the base version and reference the extension sample. 
If we put the samples in its own directory it would also cut down on the .import-restrictions allowing us to ensure we didn't allow any private dependencies in the base version.

cmd/cloud-controller-manager/.import-restrictions
```diff
@@ -1,4 +1,37 @@
 rules:
   - selectorRegexp: k8s[.]io/kubernetes
     allowedPrefixes:
-      - k8s.io/kubernetes/cmd/cloud-controller-manager
\ No newline at end of file
+      - k8s.io/kubernetes/pkg/api/legacyscheme
```
liggitt, line 4: was adding these back intentional or a bad rebase?

cmd/cloud-controller-manager/controller-manager.go
```diff
@@ -17,32 +17,128 @@ limitations under the License.
...
+	// Initialize the cloud provider with a reference to the clientBuilder
+	cloud.Initialize(c.ClientBuilder, make(chan struct{}))
+	// Set the informer on the user cloud object
+	if informerUserCloud, ok := cloud.(cloudprovider.InformerUser); ok {
+		informerUserCloud.SetInformers(c.SharedInformers)
+	}
+
+	// remove the controller which current cloud provider does not need. e.g. cloud-node-lifecycle
+	controllerInitializers := app.DefaultControllerInitializers(c.Complete(), cloud)
+	delete(controllerInitializers, "cloud-node-lifecycle")
```
andrewsykim, line 104: this should probably be commented out before merge, just to prevent people from accidentally copy/pasting this and disabling the cloud-node-lifecycle controller accidentally

cmd/cloud-controller-manager/.import-restrictions
cici37: intentional. This is indirectly imported by `k8s.io/kubernetes/pkg/controller/testutil` -> `node_ipam_controller_test` in `k8s.io/kubernetes/pkg/controller/nodeipam'.

cmd/cloud-controller-manager/controller-manager.go
cici37: Good point. Let me commented out and add more comment to make it clear

cmd/cloud-controller-manager/.import-restrictions
cici37: I could have a follow up PR to put the samples(like nodeIpamController) in its own directory and also split out the base/vanilla version from the sample showing usage of the extensions as mentioned above. In this way the .import-restrictions file might look better and make less confusion for cloud providers to follow.

cici37:
@liggitt Would you please have a look when you have time? Thank you :)

The changes added after last review:
Commented out the example of deleting a controller to prevent it from accidentally deletion. Add more comments.

staging/src/k8s.io/cloud-provider/app/testing/testserver.go
```diff
@@ -84,13 +85,13 @@ func StartTestServer(t Logger, customFlags []string) (result TestServer, err err
 	if err != nil {
 		return TestServer{}, err
 	}
-	all, disabled := app.KnownControllers(), app.ControllersDisabledByDefault.List()
+	all := []string{"cloud-node", "cloud-node-lifecycle", "service", "route"}
```
liggitt, line 88: will anything notice if this drifts? does it matter if it does?

cmd/cloud-controller-manager/controller-manager.go
```diff
@@ -17,32 +17,131 @@ limitations under the License.
 // The external controller manager is responsible for running controller loops that
 // are cloud provider dependent. It uses the API to listen to new events on resources.
 
+// This file should be written by each cloud provider.
+// The current file demonstrate how other cloud provider should leverage CCM and it uses fake parameters. Please modify for your own use.
```
liggitt, line 21: This seems to be changing the current cloud-controller-manager binary to use fake / example params... am I reading that correctly? Is this in use by anyone who would be affected by these changes?

cmd/cloud-controller-manager/.import-restrictions
liggitt: The changes under staging look fine to me.

I'm having trouble understanding the current status/consumers and goals for each of the following:

1. current `cloud-controller-manager` command / binary (cmd/cloud-controller-manager)
  * is this included in current releases or image manifests built/pushed? are there consumers of this today?
2. ideal minimal example `main()`
  * I expected this to live in staging `k8s.io/cloud-provider` with no k8s.io/kubernetes dependencies
3. an example that stitches in node_ipam to the minimal example (either by moving out node_ipam or by living in k8s.io/kubernetes)

Modifying the existing `main()` to add node_ipam without first extracting the minimal k8s.io/kubernetes-free one is making the path to the minimal on harder to see. Is that still the plan?

cmd/cloud-controller-manager/controller-manager.go
cici37: Current plan is to keep this file as a sample main.go in k/k. Other cloud providers will have their own main.go file in their repo(e.g.:`https://github.com/kubernetes/cloud-provider-aws/blob/master/cmd/aws-cloud-controller-manager/main.go`).
The current cloud-controller-manager binary (cmd/cloud-controller-manager) will only contain sample code and will not have consumers.

staging/src/k8s.io/cloud-provider/app/testing/testserver.go
cici37: It doesn't matter even if we pass an empty string array. Used for validate: `https://github.com/kubernetes/kubernetes/blob/bd95fb101dd83cfca9d95e55b2f0d29720016e51/staging/src/k8s.io/controller-manager/options/generic.go#L90`

cici37:
@liggitt Thank you so much for the review. 

1. The current plan to to have `cmd/cloud-controller-manager` only contains sample code and each cloud provider will vendor in k8s.io/cloud-provider instead. 

2. `cmd/cloud-controller-manager/controller-manager.go` would be living in k/k as a detailed example to show new coming/existing cloud providers the way to leverage CCM. I added another minimal example `main()` under `k8s.io/cloud-provider/sample/main.go`.

3. Since `cmd/cloud-controller-manager/controller-manager.go` still live in k/k, I guess it would be ok to have node_ipam example living inside k/k as well. Plan to have a following PR to move node_ipam sample out to a separate dir so that the .import-restrictions file might look better.

cmd/cloud-controller-manager/.import-restrictions
cheftako: 1.a) As a fast follow it might be nice to split the binary. Have a vanilla binary with none of the extensions in use which a cloud provider could copy and modify to get started. A sample which shows how to use each of the extension mechanisms. I think the vanilla should be here and the sample under a sample sub directory?

1.b) The CCM in K/K is not included in any releases but it is built to ensure it remains buildable. It is forked and modified into various other repos where the modified version is released. These changes should allow the cloud providers to just depend on libraries, no more forking needed. 

2.a) +1, see the fast follow suggestion above. 

2.b) +Alot, if we break the sample into its own directory, that should allow the base version to have no K8s.io/kubernetes dependencies. More importantly we can then enforce that the base version have no K8s.io/kubernetes dependencies.

3) +1

liggitt:
talked offline... cmd/cloud-controller-manager is currently built, but not included in release artifacts

this PR is an interim state, and will be followed by the split described in https://github.com/kubernetes/kubernetes/pull/93764#discussion_r518907192

/approve
for dep changes

will leave lgtm to @cheftako

cici37:
Have the following PR to add more samples: https://github.com/kubernetes/kubernetes/pull/96385

PR #93769, opened 2020-08-07
e2e delete namespaces after finish

aojea:
**What type of PR is this?**

/kind bug
/kind cleanup
/kind failing-test
/kind flake

The test `[sig-network] [Feature:PerformanceDNS][Serial] Should answer DNS query for maximum number of services per cluster` is not cleaning up the resources after it runs, and causes etcd to exhaust the quota, thus, causing the subsequent tests to fail because of

```
"error":"error initializing the store: etcdserver: mvcc: database space exceeded","level":"fatal","msg":"error executing sensu-backend","time":"2019-10-30T09:27:15Z"Unexpected Error: "500"
```

xref https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-kind-network/1291532747889709057

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/network/dns_scale_records.go
+++ b/test/e2e/network/dns_scale_records.go
@@ -60,6 +60,7 @@ var _ = SIGDescribe("[Feature:PerformanceDNS][Serial]", func() {
 		for i := 0; i < numNs; i++ {
 			ns, _ := f.CreateNamespace(f.BaseName, nil)
 			namespaces = append(namespaces, ns.Name)
+			f.AddNamespacesToDelete(ns)
 		}
 
 		services := generateServicesInNamespaces(namespaces, maxServicesPerCluster)


wojtek-t:
Seems like misconfigured etcd to me too (this single test shouldn't result in exceeding etcd quota).
But the PR is obviously a reasonable mitigation.

/lgtm
/approve

aojea:
> Seems like misconfigured etcd to me too (this single test shouldn't result in exceeding etcd quota).
> But the PR is obviously a reasonable mitigation.
> 

It's the default configuration for KIND, that uses kubeadm
where is this limit for services/etcd configured?

aojea:
/cc @hasheddan 
for milestone approvad, is a e2e test fix

aojea:
> > Seems like misconfigured etcd to me too (this single test shouldn't result in exceeding etcd quota).
> > But the PR is obviously a reasonable mitigation.
> 
> It's the default configuration for KIND, that uses kubeadm
> where is this limit for services/etcd configured?

for the record, @BenTheElder answered it
https://github.com/kubernetes/kubernetes/blob/92e51c19017a9e417df291e469ffebb13b6b51e2/cmd/kubeadm/app/phases/etcd/local.go#L281

BenTheElder:
e2e-gce seems to also have the same snapshot count though, and I don't see anything else relevant, though I'm not an expert.

PR #93770, opened 2020-08-07
Add testcase to validate the reason of `FitError`

chendave:
The error message should be rolled up into one line summary.

**What type of PR is this?**

/kind cleanup

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/scheduler/core/generic_scheduler_test.go
+++ b/pkg/scheduler/core/generic_scheduler_test.go
@@ -1530,3 +1530,81 @@ func TestPreferNominatedNodeFilterCallCounts(t *testing.T) {
 		})
 	}
 }
+
+func TestFitErrorNoSpammy(t *testing.T) {
+	tests := []struct {
+		name     string
+		nodeNum1 int
+		nodeNum2 int
+		wErr     string
+	}{
+		{
+			name:     "small number of nodes cannot pass the filters",
+			nodeNum1: 2,
+			wErr:     "0/2 nodes are available: 2 injecting failure for pod fake.",
+		},
+		{
+			name:     "large number of nodes cannot pass the filters",
+			nodeNum1: 500,
+			wErr:     "0/500 nodes are available: 500 injecting failure for pod fake.",
+		},
+		{
+			name:     "cannot pass the filters with two different reasons",
+			nodeNum1: 490,
+			nodeNum2: 10,
+			wErr:     "0/500 nodes are available: 10 Nodes failed the fake plugin, 490 injecting failure for pod fake.",
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			pod := &v1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "fake", UID: types.UID("fake")}}
+			cache := internalcache.New(time.Duration(0), wait.NeverStop)
+			var nodes []*v1.Node
+			failedNodeReturnCodeMap := make(map[string]framework.Code)
+			for i := 0; i < test.nodeNum1; i++ {
+				uid := fmt.Sprintf("node%v", i)
+				node := &v1.Node{ObjectMeta: metav1.ObjectMeta{Name: uid, UID: types.UID(uid)}}
+				nodes = append(nodes, node)
+				cache.AddNode(node)
+				failedNodeReturnCodeMap[uid] = framework.Unschedulable
+			}
+
+			registerPlugins := []st.RegisterPluginFunc{
+				st.RegisterQueueSortPlugin(queuesort.Name, queuesort.New),
+				st.RegisterFilterPlugin(
+					"FakeFilter",
+					st.NewFakeFilterPlugin(failedNodeReturnCodeMap),
+				),
+				st.RegisterFilterPlugin("FalseFilter", st.NewFalseFilterPlugin),
+				st.RegisterBindPlugin(defaultbinder.Name, defaultbinder.New),
+			}
+			for i := test.nodeNum1; i < test.nodeNum1+test.nodeNum2; i++ {
+				uid := fmt.Sprintf("node%v", i)
+				node := &v1.Node{ObjectMeta: metav1.ObjectMeta{Name: uid, UID: types.UID(uid)}}
+				nodes = append(nodes, node)
+				cache.AddNode(node)
+			}
+			snapshot := internalcache.NewSnapshot(nil, nodes)
+			fwk, err := st.NewFramework(
+				registerPlugins,
+				frameworkruntime.WithSnapshotSharedLister(snapshot),
+				frameworkruntime.WithPodNominator(internalqueue.NewPodNominator()),
+			)
+			if err != nil {
+				t.Fatal(err)
+			}
+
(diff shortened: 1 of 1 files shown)


pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -1188,3 +1188,90 @@ func TestFairEvaluationForNodes(t *testing.T) {
 		}
 	}
 }
+
+func TestFitErrorNoSpammy(t *testing.T) {
+	tests := []struct {
+		name           string
+		nodeNum1       int
```
alculquicondor, line 1195: perhaps `errorNodes` and `falseNodes`?

pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -1188,3 +1188,90 @@ func TestFairEvaluationForNodes(t *testing.T) {
...
+			pod := &v1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "fake", UID: types.UID("fake")}}
+			cache := internalcache.New(time.Duration(0), wait.NeverStop)
+			var nodes []*v1.Node
+			failedNodeReturnCodeMap := make(map[string]framework.Code)
+			for i := 0; i < test.nodeNum1; i++ {
+				uid := fmt.Sprintf("node%v", i)
+				node := &v1.Node{ObjectMeta: metav1.ObjectMeta{Name: uid, UID: types.UID(uid)}}
+				nodes = append(nodes, node)
+				cache.AddNode(node)
+				failedNodeReturnCodeMap[uid] = framework.Unschedulable
```
alculquicondor, line 1230: What was the map for? shouldn't this be `framework.Error`?

pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -1188,3 +1188,90 @@ func TestFairEvaluationForNodes(t *testing.T) {
 		}
 	}
 }
+
+func TestFitErrorNoSpammy(t *testing.T) {
+	tests := []struct {
+		name           string
+		nodeNum1       int
+		nodeNum2       int
+		addFalseFilter bool
```
alculquicondor, line 1197: just `nodeNum2` being 0 should be enough

pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -1188,3 +1188,90 @@ func TestFairEvaluationForNodes(t *testing.T) {
...
+			scheduler := NewGenericScheduler(
+				cache,
+				snapshot,
+				[]framework.Extender{},
+				nil,
+				false,
+				schedulerapi.DefaultPercentageOfNodesToScore)
+
+			_, err = scheduler.Schedule(context.Background(), prof, framework.NewCycleState(), pod)
+			if fmt.Sprint(err) != test.wErr {
```
alculquicondor, line 1272: nit: `err.Error()`

pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -1188,3 +1188,90 @@ func TestFairEvaluationForNodes(t *testing.T) {
...
+				t.Errorf("want: %v, got: %v", test.wErr, fmt.Sprint(err))
```
alculquicondor, line 1273: got before want

alculquicondor (commented):
Thanks

pkg/scheduler/core/generic_scheduler_test.go
chendave: it's mimic `Unschedulable` status and `feasibleNodes` is `0`, so that `FitError` could be returned.

pkg/scheduler/core/generic_scheduler_test.go
chendave: Done

pkg/scheduler/core/generic_scheduler_test.go
chendave: Done

pkg/scheduler/core/generic_scheduler_test.go
chendave: Done

pkg/scheduler/core/generic_scheduler_test.go
chendave: Both `FakeFilter` and `FalseFilter` used here will return `unschedulable`, `errorNodes` might not fit here.

chendave (commented):
@alculquicondor thanks for your great suggestion! either replied or addressed.

pkg/scheduler/core/generic_scheduler_test.go
chendave: if it's not `unschedulable` status, it will not return `FitError` ,

https://github.com/kubernetes/kubernetes/blob/4f9317596ccb9fcb37bf183babf3e8640afcca9b/pkg/scheduler/core/generic_scheduler.go#L111-L122

and then the error message won't roll up to one line by below method, 
```func (f *FitError) Error()```

**Comment by chendave** [comment]
rebased.

@alculquicondor @ahg-g Does this one still have the chance to go?

File: pkg/scheduler/core/generic_scheduler_test.go [hunk 6]
```diff
@@ -1530,3 +1530,81 @@ func TestPreferNominatedNodeFilterCallCounts(t *testing.T) {
...
+				failedNodeReturnCodeMap[uid] = framework.Unschedulable
+			}
+
+			registerPlugins := []st.RegisterPluginFunc{
+				st.RegisterQueueSortPlugin(queuesort.Name, queuesort.New),
+				st.RegisterFilterPlugin(
+					"FakeFilter",
+					st.NewFakeFilterPlugin(failedNodeReturnCodeMap),
+				),
+				st.RegisterFilterPlugin("FalseFilter", st.NewFalseFilterPlugin),
```
alculquicondor, line 1579: do we need this extra filter?

alculquicondor (commented):
Sorry, I forgot about this PR.

So we are trying to test that the string representation of FitError is summarized. Why don't we directly test `FitError.Error` in the framework package? Or you can test `recordSchedulingFailure` in `scheduler` package sends the correct event message (I'm not sure if this is easy to test at unit level). `TestGenericScheduler` already tests that we return the correct `FitError`.

If your intent is to have a more E2E test, then the appropriate thing to do is to do it in the integration tests, where you can query the events and validate that message.

pkg/scheduler/core/generic_scheduler_test.go
chendave: yes, we need this filter.

I want to test the case that the pod cannot pass the filter with different reasons,  pls see my this testcase, 
```cannot pass the filters with two different reasons```

chendave:
>  Why don't we directly test FitError.Error in the framework package?
Test in framework workable as well, but the whole testcase need to rewrite.

alculquicondor:
Just test the `Error` method

chendave:
> Just test the Error method

Got you, will come back for this later.

chendave:
this is indeed a trivial change,  no need to waste reviewer's time one this. so close.

/close

PR #93690, opened 2020-08-04
test/e2e: fail test rather than flooding logs if PVC watch is closed prematurely

dcbw:
If the watch is closed prematurely the for/select loop will
spew "PVC event : <nil>" as fast as possible until the test times
out, flooding logs.

```
STEP: Checking PVC events
[...]

PR #93777, opened 2020-08-07
e2e storage: avoid flaky test failure when watch dies

pohly:
**What type of PR is this?**

/kind bug
/kind flake

As discussed in https://github.com/kubernetes/kubernetes/pull/93658,
relying on a watch to deliver all events is not acceptable, not even
for a test, because it can and did (at least for OpenShift testing) lead to
test flakes.

The solution in #93658 was to replace log flooding with a clear test
failure, but that didn't solve the flakiness.

There's no other good way to observe behavior over time either. Even
logs may get deleted before they can be retrieved. Therefore the watch
is now still used, but its output is optional for the test and only
used if the watch delivered all required events.

```release-note
NONE
```

### DIFF
--- a/test/e2e/storage/csi_mock_volume.go
+++ b/test/e2e/storage/csi_mock_volume.go
@@ -39,6 +39,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 	clientset "k8s.io/client-go/kubernetes"
+	cachetools "k8s.io/client-go/tools/cache"
+	watchtools "k8s.io/client-go/tools/watch"
 	"k8s.io/kubernetes/pkg/controller/volume/scheduling"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
@@ -821,7 +823,20 @@ var _ = utils.SIGDescribe("CSI mock volume", func() {
 
 				ctx, cancel := context.WithTimeout(context.Background(), csiPodRunningTimeout)
 				defer cancel()
-				pvcWatch, err := f.ClientSet.CoreV1().PersistentVolumeClaims(f.Namespace.Name).Watch(ctx, metav1.ListOptions{})
+
+				// In contrast to the raw watch, RetryWatcher is expected to deliver all events even
+				// when the underlying raw watch gets closed prematurely
+				// (https://github.com/kubernetes/kubernetes/pull/93777#discussion_r467932080).
+				// This is important because below the test is going to make assertions about the
+				// PVC state changes.
+				initResource, err := f.ClientSet.CoreV1().PersistentVolumeClaims(f.Namespace.Name).List(ctx, metav1.ListOptions{})
+				framework.ExpectNoError(err, "Failed to fetch initial PVC resource")
+				listWatcher := &cachetools.ListWatch{
+					WatchFunc: func(listOptions metav1.ListOptions) (watch.Interface, error) {
+						return f.ClientSet.CoreV1().PersistentVolumeClaims(f.Namespace.Name).Watch(ctx, listOptions)
+					},
+				}
+				pvcWatch, err := watchtools.NewRetryWatcher(initResource.GetResourceVersion(), listWatcher)
 				framework.ExpectNoError(err, "create PVC watch")
 				defer pvcWatch.Stop()
 
@@ -889,12 +904,14 @@ var _ = utils.SIGDescribe("CSI mock volume", func() {
 				ginkgo.By("Checking PVC events")
 				nodeAnnotationSet := false
 				nodeAnnotationReset := false
+				watchFailed := false
 			loop:
 				for {
 					select {
 					case event, ok := <-pvcWatch.ResultChan():
 						if !ok {
-							framework.Failf("PVC watch ended prematurely")
+							watchFailed = true
+							break loop
 						}
 
 						framework.Logf("PVC event %s: %#v", event.Type, event.Object)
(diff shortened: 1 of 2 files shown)


pohly:
@dims : this supersedes the currently open https://github.com/kubernetes/kubernetes/pull/93690 which is a cherry-pick of the previous PR for this issue.

Do I need to do a cherry-pick once this is merged?

hasheddan (commented):
@pohly As mentioned [here](https://github.com/kubernetes/kubernetes/pull/93690#pullrequestreview-463799732) as long as `release-1.19` is still having daily fast forward then cherry picks to the release branch are not needed 👍 

/milestone v1.19

test/e2e/storage/csi_mock_volume.go
```diff
@@ -889,12 +889,14 @@ var _ = utils.SIGDescribe("CSI mock volume", func() {
 				ginkgo.By("Checking PVC events")
 				nodeAnnotationSet := false
 				nodeAnnotationReset := false
+				watchFailed := false
 			loop:
 				for {
 					select {
 					case event, ok := <-pvcWatch.ResultChan():
 						if !ok {
-							framework.Failf("PVC watch ended prematurely")
```
liggitt, line 897: the main issue seems to be opening a raw watch and expecting it to never drop during the test. shouldn't we be using the watch tools helper that cam reestsblish a dropped watch and continue it?

liggitt: cc @tnozicka

pohly: We've discussed that in more detail in https://github.com/kubernetes/kubernetes/pull/93658#pullrequestreview-460530486: even these helpers only guarantee that eventually the current state is reported.

What the test would like to have (but cannot get reliably) is a list of all intermediate states. If it isn't able to get that, it cannot make assertions about these states, but that's not a test failure - at least that's how I have implemented it in this PR.

pohly: IMHO it is better to have a test that most of the time is able to check something than to have no test for that at all.

liggitt: > even these helpers only guarantee that eventually the current state is reported.

with watch bookmarks, they are far more reliable than a raw watch…

pohly: That sounds worthwhile. Where can I find these watch tool helpers?

liggitt: NewIndexerInformerWatcher (uses an informer, falls back to list if watch cannot be reestablished) or NewRetryWatcher (retries watch only) in k8s.io/client-go/tools/watch

pohly: After reading the documentation, it is not clear whether NewIndexerInformerWatcher is really less likely to miss intermediate object updates. NewRetryWatcher sounds more promising, but also is "not resilient to etcd cache not having the resource version anymore". Not sure how it then behaves.

Both seem have in common that they hide from the user when changes have been lost, right?

From the perspective of this test, that's worse than loosing the raw watch. When the raw watch fails, the test knows that it doesn't have all data. But when the watch tools internally hide such a failure, the test might flake because it makes assumption about incomplete data. I'm leaning towards keeping the test as it is now.

If there are concerns that it will not test the feature because the watch fails too frequently, perhaps adding a retry loop around the entire test with a limit and "test passed if limit reached" would address that?

liggitt: > NewRetryWatcher sounds more promising, but also is "not resilient to etcd cache not having the resource version anymore".

With watch bookmarks enabled, we expect that not to happen under normal circumstances

liggitt: > perhaps adding a retry loop around the entire test with a limit and "test passed if limit reached" would address that?

I would prefer deflaking the test over simply retrying it, especially since the storage tests already consume a disproportionate amount of time in the e2e

pohly: > With watch bookmarks enabled, we expect that not to happen under normal circumstances

In that case I'll switch to that.

/hold

pohly: Done.

/cancel hold

pohly: /hold cancel

test/e2e/storage/csi_mock_volume.go
```diff
@@ -821,7 +823,20 @@ var _ = utils.SIGDescribe("CSI mock volume", func() {
...
-				pvcWatch, err := f.ClientSet.CoreV1().PersistentVolumeClaims(f.Namespace.Name).Watch(ctx, metav1.ListOptions{})
+
+				// In contrast to the raw watch, RetryWatcher is expected to deliver all events even
+				// when the underlying raw watch gets closed prematurely
+				// (https://github.com/kubernetes/kubernetes/pull/93777#discussion_r467932080).
+				// This is important because below the test is going to make assertions about the
+				// PVC state changes.
+				initResource, err := f.ClientSet.CoreV1().PersistentVolumeClaims(f.Namespace.Name).List(ctx, metav1.ListOptions{})
+				framework.ExpectNoError(err, "Failed to fetch initial PVC resource")
+				listWatcher := &cachetools.ListWatch{
```
liggitt, line 834: is it intentional that this is omitting a ListFunc? should it implement one that errors and fails the test rather than getting a nil panic?

pohly: I was following the approach from `WatchEventSequenceVerifier` which also only specifies a `WatchFunc`: https://github.com/kubernetes/kubernetes/blob/b1b93e30131986319f1d46a62771feb9c0f097ae/test/e2e/framework/util.go#L1335-L1348

To be honest, I hadn't considered whether that is correct. Now that you asked, I had a closer look and came to the conclusion that just the `WatchFunc` is needed because `NewRetryWatcher` only needs an implementation of the `cache.Watcher` interface.

liggitt: that's true at the moment, but if that implementation ever changes we'll get a panic, which might be tough to track down. for purposes of this test, this seems ok for now.

liggitt:
the watch bits lgtm, I'll let a storage reviewer give the official lgtm+approve

PR #93778, opened 2020-08-07
Fix sample controller scalling

### LINKED ISSUE #92758: [sample controller] scalling does not work
**What happened**: Playing with [staging sample-controller](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/sample-controller) I noticed that scaling does not work as expected.

**What you expected to happen**: After running the sample controller as per instructions I create a deployment using one of the example deployments. The deployments and the rest of the resources are created (I see one pod). Then I try to scale it up to more pods which results in the pods being created and terminated soon afterwards.
image

**How to reproduce it (as minimally and precisely as possible)**: Run the sample controller as per instructions. Then create a deployment using one of the example deployments. Then I try to scale it up to more pods. (See img above)

**Anything else we need to know?**: I'm looking to get started with kube contributions. With a bit of guidance I could try to fix this one.

**Environment**:
- Kubernetes version (use `kubectl version`): ```Client Version: version.Info{Major:"", Minor:"", GitVersion:"v0.0.0-master+$Format:%h$", GitCommit:"$Format:%H$", GitTreeState:"", BuildDate:"1970-01-01T00:00:00Z", GoVersion:"go1.14.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.6-beta.0", GitCommit:"e7f962ba86f4ce7033828210ca3556393c377bcc", GitTreeState:"clean", BuildDate:"2020-01-15T08:18:29Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}```
- OS (e.g: `cat /etc/os-release`): OSx Catalina 10.15.5
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

zerogvt:
**What type of PR is this?**

/kind bug

Resolves scaling bug by setting cached replicas count equal to persistent deployment count.
 
Fixes #92758

```release-note
Sample controller persists scaling replicas up and down.
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/sample-controller/controller.go
+++ b/staging/src/k8s.io/sample-controller/controller.go
@@ -297,6 +297,7 @@ func (c *Controller) syncHandler(key string) error {
 	// should update the Deployment resource.
 	if foo.Spec.Replicas != nil && *foo.Spec.Replicas != *deployment.Spec.Replicas {
 		klog.V(4).Infof("Foo %s replicas: %d, deployment replicas: %d", name, *foo.Spec.Replicas, *deployment.Spec.Replicas)
+		*foo.Spec.Replicas = *deployment.Spec.Replicas
 		deployment, err = c.kubeclientset.AppsV1().Deployments(foo.Namespace).Update(context.TODO(), newDeployment(foo), metav1.UpdateOptions{})
 	}
 


zerogvt:
Signed and emailed CLA (username: zerogvt)

zerogvt:
I don't think this is ready for review as it is missing:
1. A unit test that catches the bug in the first place
2. A unit test that tests the fix
But first I need help to understand the test mechanism used in order to implement these. So, any pointers to the logic of the [test code](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/sample-controller/controller_test.go) would be appreciated.

PR #93781, opened 2020-08-07
Disable log rotation of kubernetes and pod logs

kisieland:
**What type of PR is this?**

/kind feature

Adds flags to kubernetes setup on GCE which allows to switch on/off log rotation.
This mechanism may not be needed, because as of 1.11 kublet feature `CRIContainerLogRotation` is enabled by default.

(Not 100% sure about this)
```release-note
NONE
```

NOTE: I know my CLA is not yet signed.


neolit123:
hi, ideally changes in kube-up and its "provider" code should not be accepted unless it benefits k8s e2e tests.
if you are using kube-up for creating your cluster, please try to move away from it because it's unsupported and deprecated.

/priority awaiting-more-evidence

kisieland:
cla should be signed.

kisieland:
With those changes e2e tests have only one mechanism of log rotation enabled.

Tested with:
```
$ ./kube-up.sh
...
$ ssh master/node
[ssh] $ ls /etc/logrotate.d
// empty
```

```
$ ENABLE_LOGROTATE_FILES=true ENABLE_POD_LOG=true ./kube-up.sh
...
$ ssh master/node
[ssh] $ ls /etc/logrotate.d
allpodlogs  allvarlogs
```

kisieland:
```
/assign @wojtek-t
```

wojtek-t:
/ok-to-test

> hi, ideally changes in kube-up and its "provider" code should not be accepted unless it benefits k8s e2e tests.
if you are using kube-up for creating your cluster, please try to move away from it because it's unsupported and deprecated.

@neolit123 - after recent changes it already benefits tests, as by default we stop logrotating files from two different places at the same time.

kisieland:
`
\assign @neolit123  @wojtek-t

kisieland:
`
\assign @neolit123 @wojtek-t 
`

cluster/gce/config-default.sh
```diff
@@ -425,13 +425,20 @@ METADATA_CLOBBERS_CONFIG="${METADATA_CLOBBERS_CONFIG:-false}"
 
 ENABLE_BIG_CLUSTER_SUBNETS="${ENABLE_BIG_CLUSTER_SUBNETS:-false}"
 
+# Optional: Enable log rotation for k8s services
+ENABLE_LOGROTATE_FILES="${ENABLE_LOGROTATE_FILES:-false}"
+PROVIDER_VARS="${PROVIDER_VARS:-} ENABLE_LOGROTATE_FILES"
```
wojtek-t, line 430: Let's add the same lines also to cluster/gce/config-test.sh

wojtek-t (commented):
One minor comment - other than that LGTM.

wojtek-t:
@kisieland - please squash the commits

jingxu97:
I noticed that on master, some logs do not rotate any more, so it ends up having very big log files. e.g.,
Is this desired behavior? @wojtek-t 

8.4G	/var/log/kube-apiserver-audit.log
3.1G	/var/log/kube-apiserver.log

wojtek-t:
@jingxu97 - that's a good question.
I was under impression that the log rotation should still be happening as described in the PR:
```
This mechanism may not be needed, because as of 1.11 kublet feature CRIContainerLogRotation is enabled by default.
```

If it's not the case and they are not rotated at all, then this should be fixed.
Looking into example logs from 5k run, clearly none of them were rotated:
https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1316424001190367232/artifacts/gce-scale-cluster-master/

So yeah - I think - this should be fixed.
@kisieland - can you please address this?

kisieland:
I might have made too big assumption about `CRIContainerLogRotation`.
I seems that it rotates logs in `/var/log/pods/*/*.log` but not in `/var/log/*.log`
@jingxu97 can you confirm that those logs, are from `/var/log`? (They look like that for me)

In the mean time I will try to confirm those findings and create patch for it.

jingxu97:
@kisieland yes, I can confirm those logs are from /var/log

kisieland:
@jingxu97 I've created PR with fix #95753

PR #93789, opened 2020-08-07
Add test cluster check when starting csi proxy

jingxu97:
remove the check for test cluster when downloading csi proxy

**What type of PR is this?**

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster/gce/windows/k8s-node-setup.psm1
+++ b/cluster/gce/windows/k8s-node-setup.psm1
@@ -419,8 +419,10 @@ function DownloadAndInstall-CSIProxyBinaries {
 
 # TODO(jingxu97): Make csi-proxy.exe as a service similar to kubelet.exe
 function Start-CSIProxy {
-  Log-Output 'Starting CSI Proxy'
-  Start-Process "${env:NODE_DIR}\csi-proxy.exe"
+  if (Test-IsTestCluster $kube_env) {
+    Log-Output 'Starting CSI Proxy'
+    Start-Process "${env:NODE_DIR}\csi-proxy.exe"
+  }
 }
 
 # TODO(pjh): this is copied from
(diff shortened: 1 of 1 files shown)


cluster/gce/windows/k8s-node-setup.psm1
```diff
@@ -930,14 +928,13 @@ function Configure-GcePdTools {
...
-    if (ShouldWrite-File ${env:K8S_DIR}\diskutil.exe) {
-      # The source code of this executable file is https://github.com/kubernetes-sigs/sig-windows-tools/blob/master/cmd/diskutil/diskutil.c
-      MustDownload-File -OutFile ${env:K8S_DIR}\diskutil.exe `
-        -URLs "https://ddebroywin1.s3-us-west-2.amazonaws.com/diskutil.exe"
-    }
-    Copy-Item ${env:K8S_DIR}\diskutil.exe -Destination "C:\Windows\system32"
+  if (ShouldWrite-File ${env:K8S_DIR}\diskutil.exe) {
+    # The source code of this executable file is https://github.com/kubernetes-sigs/sig-windows-tools/blob/master/cmd/diskutil/diskutil.c
+    MustDownload-File -OutFile ${env:K8S_DIR}\diskutil.exe `
+      -URLs "https://ddebroywin1.s3-us-west-2.amazonaws.com/diskutil.exe"
```
yliaog, line 934: it's kind of ok to download from the above URL if the cluster is for test, it is not ok if it is not for test.

jingxu97: we recently added an API in csi-proxy which will remove the need of this diskutil. I will update driver and remove this part after testing.

jingxu97:
@yliaog @pjh I updated the PR. PTAL

pjh:
Thanks Jing. So this is now just adding a check that we missed the first time around?

Was there anything else you wanted this PR to do? If we can remove the dependency on diskutil.exe like you said then that would be good, but if we can't do it now then we can just submit this.

jingxu97:
> Thanks Jing. So this is now just adding a check that we missed the first time around?
> 
> Was there anything else you wanted this PR to do? If we can remove the dependency on diskutil.exe like you said then that would be good, but if we can't do it now then we can just submit this.

Now csi-proxy can handle getDiskName already. But I am waiting for a new release cut so that our build and release pipeline can get the latest version. It should be done this week.

jingxu97:
@pjh @yliaog PTAL. Thanks!

pjh:
/lgtm

Thanks!

PR #93790, opened 2020-08-07
Fix namespace controller cleanup orphaning

liggitt:
**What type of PR is this?**

/kind bug

Fixes an issue where a transient NotFound or MethodNotSupported error encountered by the namespace controller can prevent cleanup of a particular resource in the future.

When the namespace controller is cleaning up a particular resource, it handles MethodNotSupported and NotFound errors when making deletecollection and list requests:

https://github.com/kubernetes/kubernetes/blob/f5334fad483856c41ceb1031036a96e96c32fcee/pkg/controller/namespace/deletion/namespaced_resources_deleter.go#L312-L315
https://github.com/kubernetes/kubernetes/blob/f5334fad483856c41ceb1031036a96e96c32fcee/pkg/controller/namespace/deletion/namespaced_resources_deleter.go#L334-L338
https://github.com/kubernetes/kubernetes/blob/f5334fad483856c41ceb1031036a96e96c32fcee/pkg/controller/namespace/deletion/namespaced_resources_deleter.go#L353-L356
https://github.com/kubernetes/kubernetes/blob/f5334fad483856c41ceb1031036a96e96c32fcee/pkg/controller/namespace/deletion/namespaced_resources_deleter.go#L369-L373

If both deletecollection and list requests indicate they are not supported, or return 404 errors for a given namespace, it is reasonable to take that to mean that no instances of that resource exist in that namespace any longer. 

However, we should *not* cache lack of support for deletecollection and list and skip even attempting those for future namespace cleanups (which is what the code currently does). That causes orphaning in the following scenario:

Imagine an uninstaller/installer that deletes and recreates a CRD and namespace:

1. delete CRD for resource `foo`; `foo` instances start getting cleaned up in the background
2. delete namespace `n`; the namespace controller discovers extant resource types (which still includes the resource `foo`) and starts iterating through them
3. `foo` instance cleanup completes, the CRD is removed, and the `foo` endpoints are no longer served
4. the namespace controller reaches resource `foo`, calls and gets 404s for deletecollection and list operations
5. the installer recreates the CRD for `foo`

If the namespace controller caches the lack of support for deletecollection/list for `foo` and does not even attempt to perform those operations in the future, all future namespace deletions will orphan `foo` instances until the controller restarts

```release-note
Fixes an issue that can result in namespaced custom resources being orphaned when their namespace is deleted, if the CRD defining the custom resource is removed concurrently with namespaces being deleted, then recreated.
```

/sig api-machinery
/kind bug
/priority critical-urgent
/cc @deads2k @adrianludwin


liggitt:
modules.txt fixup

liggitt:
since this is a correctness issue we consider critical, including in 1.19 and opening cherry-picks

/milestone v1.19

pkg/controller/namespace/deletion/namespaced_resources_deleter.go
```diff
@@ -265,11 +265,9 @@ func (d *namespacedResourcesDeleter) updateNamespaceStatusFunc(namespace *v1.Nam
 	if namespace.DeletionTimestamp.IsZero() || namespace.Status.Phase == v1.NamespaceTerminating {
 		return namespace, nil
 	}
-	newNamespace := v1.Namespace{}
-	newNamespace.ObjectMeta = namespace.ObjectMeta
-	newNamespace.Status = *namespace.Status.DeepCopy()
+	newNamespace := namespace.DeepCopy()
```
adrianludwin, line 268: What got lost when we only updated the status? Do you know why the prior authors thought it could be omitted? I noticed that `finalizeNamespace`, below, follows the same pattern but this time only for the spec, not the status.

I may come back and add some comments in my own PR.

liggitt: spec fields which are ignored by the server in real requests but which got stomped when submitting via the fake client in the unit test I added

PR #93797, opened 2020-08-07
[go1.15] Use errors.As to unwrap net errors

liggitt:
**What type of PR is this?**
/kind cleanup

Use errors.As to correctly unwrap permanent net errors returned by tls.Conn

See https://github.com/golang/go/issues/40554

```release-note
NONE
```

/cc @justaugustus 
/sig api-machinery


liggitt:
/milestone v1.19

required for go1.15

PR #76328, opened 2019-04-09
 removed flaky watch code and added NewIndexerInformerWatcher for LimitRange test case

pontiyaraja:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
>
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:
- fixed the time flaky
- removed flaky watch code and added NewIndexerInformerWatcher

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:
- removed flaky watch code and added NewIndexerInformerWatcher

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```
/area conformance
/area testing
/sig-testing
@kubernetes/cncf-conformance-wg 
@spiffxp 
@bgrant0607 
@brahmaroutu

PR #93803, opened 2020-08-08
Deflake LimitRange e2e test

### LINKED ISSUE #93782: [Flaky Test] [sig-scheduling] LimitRange should create a LimitRange with defaults and ensure pod has those defaults applied. [Conformance]
**Which jobs are flaking**:

`kind-ipv6-master-parallel (ci-kubernetes-kind-ipv6-e2e-parallel)`

`[sig-scheduling] LimitRange should create a LimitRange with defaults and ensure pod has those defaults applied. [Conformance]`

**Testgrid link**:

https://testgrid.k8s.io/sig-release-master-blocking#kind-ipv6-master-parallel

**Reason for failure**:

```
Pod "pfpod2" is invalid: [spec.containers[0].resources.requests: Invalid value: "600m": must be less than or equal to cpu limit, spec.containers[0].resources.requests: Invalid value: "600Gi": must be less than or equal to ephemeral-storage limit, spec.containers[0].resources.requests: Invalid value: "600Mi": must be less than or equal to memory limit]
```

/sig scheduling
/priority critical-urgent
/cc @kubernetes/ci-signal @ahg-g @Huang-Wei 
/milestone v1.19

This test waits for the limit range to be deleted before attempting to create the `Pod` so this error is surprising. It appears there are no limit ranges present when they are listed here: https://github.com/kubernetes/kubernetes/blob/97c5f1f7632f2d349303515830be76f6c1084b19/test/e2e/scheduling/limit_range.go#L220

Huang-Wei:
**What type of PR is this?**

/kind bug
/kind flake
/sig scheduling
/priority critical-urgent

The old LimitRange e2e test has 2 major problems on verifying LimitRange gets deleted:

- it returns true when LimitRange is still present (L229 below)
    https://github.com/kubernetes/kubernetes/blob/a6405451674428172956116b4c93710ac03ebddd/test/e2e/scheduling/limit_range.go#L224-L229
- error of polling logic is totally bypassed due to a regression introduced in #76328 (the plain `gomega.Expect()` is an no-op)

This PR tries to fix the above two problems.

Fixes #93782

```release-note
NONE
```

### DIFF
--- a/test/e2e/scheduling/limit_range.go
+++ b/test/e2e/scheduling/limit_range.go
@@ -38,7 +38,6 @@ import (
 	imageutils "k8s.io/kubernetes/test/utils/image"
 
 	"github.com/onsi/ginkgo"
-	"github.com/onsi/gomega"
 )
 
 const (
@@ -206,10 +205,8 @@ var _ = SIGDescribe("LimitRange", func() {
 		framework.ExpectNoError(err)
 
 		ginkgo.By("Verifying the LimitRange was deleted")
-		gomega.Expect(wait.Poll(time.Second*5, e2eservice.RespondingTimeout, func() (bool, error) {
-			selector := labels.SelectorFromSet(labels.Set(map[string]string{"name": limitRange.Name}))
-			options := metav1.ListOptions{LabelSelector: selector.String()}
-			limitRanges, err := f.ClientSet.CoreV1().LimitRanges(f.Namespace.Name).List(context.TODO(), options)
+		err = wait.Poll(time.Second*5, e2eservice.RespondingTimeout, func() (bool, error) {
+			limitRanges, err := f.ClientSet.CoreV1().LimitRanges(f.Namespace.Name).List(context.TODO(), metav1.ListOptions{})
 
 			if err != nil {
 				framework.Logf("Unable to retrieve LimitRanges: %v", err)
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
Thanks for addressing this so quickly @Huang-Wei!
/lgtm
/milestone v1.19

test/e2e/scheduling/limit_range.go
```diff
@@ -206,10 +205,8 @@ var _ = SIGDescribe("LimitRange", func() {
 		framework.ExpectNoError(err)
 
 		ginkgo.By("Verifying the LimitRange was deleted")
-		gomega.Expect(wait.Poll(time.Second*5, e2eservice.RespondingTimeout, func() (bool, error) {
-			selector := labels.SelectorFromSet(labels.Set(map[string]string{"name": limitRange.Name}))
-			options := metav1.ListOptions{LabelSelector: selector.String()}
-			limitRanges, err := f.ClientSet.CoreV1().LimitRanges(f.Namespace.Name).List(context.TODO(), options)
+		err = wait.Poll(time.Second*5, e2eservice.RespondingTimeout, func() (bool, error) {
+			limitRanges, err := f.ClientSet.CoreV1().LimitRanges(f.Namespace.Name).List(context.TODO(), metav1.ListOptions{})
```
ahg-g, line 209: why the list options was removed, should we be listing the specific limit range in question?

Huang-Wei: In theory, it's fine to just check the specific LimitRange is deleted (by checking if the error is of type APINotFound). The current style (to ensure no LimitRange in the namespace) don't need to check the returned error, and also can avoid the case that any other LimitRange gets created during the test.

PR #93804, opened 2020-08-08
Remove unevaluated nodes in filter phase from preemption candidates

mrkm4ntr:
**What type of PR is this?**
/kind bug

Not all nodes are evaluated in the filter phase. This means, in the current implementation, unevaluated nodes might be targets of preemption. This is not a preferred situation.
https://github.com/kubernetes/kubernetes/blob/1fdd8fb213e0361e8f18b1dd152dddb4c88ad183/pkg/scheduler/core/generic_scheduler.go#L232-L254

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
+++ b/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
@@ -222,7 +222,7 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
 		name := node.Node().Name
 		// We reply on the status by each plugin - 'Unschedulable' or 'UnschedulableAndUnresolvable'
 		// to determine whether preemption may help or not on the node.
-		if m[name].Code() == framework.UnschedulableAndUnresolvable {
+		if s, ok := m[name]; !ok || s.Code() == framework.UnschedulableAndUnresolvable {
 			continue
 		}
 		potentialNodes = append(potentialNodes, node)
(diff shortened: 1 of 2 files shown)


alculquicondor:
numNodesToFind is not the number of nodes that are evaluated. It's the number of nodes that proceed to the scoring phase.

This means that, when a pod doesn't fit anywhere in the cluster, all nodes were evaluated.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -222,7 +222,7 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
 		name := node.Node().Name
 		// We reply on the status by each plugin - 'Unschedulable' or 'UnschedulableAndUnresolvable'
 		// to determine whether preemption may help or not on the node.
-		if m[name].Code() == framework.UnschedulableAndUnresolvable {
+		if s, ok := m[name]; !ok || s.Code() == framework.UnschedulableAndUnresolvable {
```
adtac, line 225: `m[code].Code()` will return `Success` when `m[code]` is nil, so this is functionally the same as your change.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -1000,38 +1000,38 @@ func TestNodesWherePreemptionMightHelp(t *testing.T) {
 				"node2": framework.NewStatus(framework.UnschedulableAndUnresolvable, nodename.ErrReason),
 				"node3": framework.NewStatus(framework.UnschedulableAndUnresolvable, nodeunschedulable.ErrReasonUnschedulable),
 			},
-			expected: map[string]bool{"node1": true, "node4": true},
+			expected: map[string]bool{"node1": true},
```
adtac, line 1003: FWIW `node4` wasn't being returned without the change to `default_preemption.go` either. (Actually I think this test could be refactored to use a array `expected` instead of a map of bools; we should also check that the returned node list exactly matches the expected list.)

mrkm4ntr:
> numNodesToFind is not the number of nodes that are evaluated. It's the number of nodes that proceed to the scoring phase.
This means that, when a pod doesn't fit anywhere in the cluster, all nodes were evaluated.

I misunderstood, thanks!

PR #93808, opened 2020-08-08
Scheduler ListAlgorithmProviders function return providers directly

### LINKED ISSUE #93788: Scheduler registering info shows in --version flag
Using scheduler like `scheduler --help` or `scheduler --version`, it always get unwanted output like:
```
[root@masterdemo amd64]# ./kube-scheduler --version
I0808 00:08:38.914277   30987 registry.go:173] Registering SelectorSpread plugin
I0808 00:08:38.914519   30987 registry.go:173] Registering SelectorSpread plugin
Kubernetes v1.20.0-alpha.0.2+dd649bb7ef4788-dirty
```
It simply because when adding the flags in scheduler, it will try to look up the `ListAlgorithmProviders()`
```
fs.StringVar(&o.AlgorithmProvider, "algorithm-provider", o.AlgorithmProvider, "DEPRECATED: the scheduling algorithm provider to use, this sets the default plugins for component config profiles. Choose one of: "+algorithmprovider.ListAlgorithmProviders())
``` 
And this will invoke:
```
func applyFeatureGates(config *schedulerapi.Plugins) {
	if !utilfeature.DefaultFeatureGate.Enabled(features.DefaultPodTopologySpread) {
		// When feature is enabled, the default spreading is done by
		// PodTopologySpread plugin, which is enabled by default.
		klog.Infof("Registering SelectorSpread plugin")
		s := schedulerapi.Plugin{Name: selectorspread.Name}
		config.PreScore.Enabled = append(config.PreScore.Enabled, s)
		s.Weight = 1
		config.Score.Enabled = append(config.Score.Enabled, s)
	}
}
```
Since `version` flag already has `verflag.PrintAndExitIfRequested()` to prevent running other code when having `--version`:
We should not print the klog info when using `--help` or `--version` right?

zhouya0:
**What type of PR is this?**
/kind cleanup

This pr let function `ListAlgorithmProviders` return the providers directly to avoid unwanted `klog` shows in `./kube-scheduler --version`:
```
[root@masterdemo amd64]# ./kube-scheduler --version
I0808 00:08:38.914277   30987 registry.go:173] Registering SelectorSpread plugin
I0808 00:08:38.914519   30987 registry.go:173] Registering SelectorSpread plugin
Kubernetes v1.20.0-alpha.0.2+dd649bb7ef4788-dirty
```

Fixes #93788

/cc @Huang-Wei 

```release-note
None
```

### DIFF
--- a/pkg/scheduler/algorithmprovider/registry.go
+++ b/pkg/scheduler/algorithmprovider/registry.go
@@ -17,8 +17,7 @@ limitations under the License.
 package algorithmprovider
 
 import (
-	"sort"
-	"strings"
+	"fmt"
 
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/klog/v2"
@@ -66,13 +65,7 @@ func NewRegistry() Registry {
 
 // ListAlgorithmProviders lists registered algorithm providers.
 func ListAlgorithmProviders() string {
-	r := NewRegistry()
-	var providers []string
-	for k := range r {
-		providers = append(providers, k)
-	}
-	sort.Strings(providers)
-	return strings.Join(providers, " | ")
+	return fmt.Sprintf("%s | %s", ClusterAutoscalerProvider, schedulerapi.SchedulerDefaultProviderName)
 }
 
 func getDefaultConfig() *schedulerapi.Plugins {


pkg/scheduler/algorithmprovider/registry.go
```diff
@@ -50,17 +50,19 @@ const ClusterAutoscalerProvider = "ClusterAutoscalerProvider"
 // Registry is a collection of all available algorithm providers.
 type Registry map[string]*schedulerapi.Plugins
 
+// ApplyFeatureGates applies feature gates to registry
+func (r Registry) ApplyFeatureGates() Registry {
+	for _, config := range r {
```
draveness, line 55: for loop would range the map randomly, would this affect the result of the feature gates?

zhouya0: Thx,  agree, change done.

pkg/scheduler/algorithmprovider/registry.go
```diff
@@ -50,17 +50,23 @@ const ClusterAutoscalerProvider = "ClusterAutoscalerProvider"
...
+	if defaultConfig, ok := r[schedulerapi.SchedulerDefaultProviderName]; ok {
+		applyFeatureGates(defaultConfig)
+	}
+	if caConfig, ok := r[ClusterAutoscalerProvider]; ok {
+		applyFeatureGates(caConfig)
+	}
 
-	caConfig := getClusterAutoscalerConfig()
-	applyFeatureGates(caConfig)
+	return r
```
adtac, line 62: I don't think this should return `r`. Wouldn't it be simpler to make `ApplyFeatureGates` just be a function without a return value?

also, I think this is treating the symptom rather than the root cause; shouldn't we just add a verbosity filter to the `klog.Infof` that is polluting stdout?

Huang-Wei (commented):
We should ensure `--version` is respected and exit without running any real scheduler bootstrap logic.

```diff
diff --git a/cmd/kube-scheduler/app/server.go b/cmd/kube-scheduler/app/server.go
index b7c22674874..95532b387bf 100644
--- a/cmd/kube-scheduler/app/server.go
+++ b/cmd/kube-scheduler/app/server.go
@@ -78,6 +78,8 @@ kube-scheduler is the reference implementation.
 See [scheduling](https://kubernetes.io/docs/concepts/scheduling-eviction/)
 for more information about scheduling and the kube-scheduler component.`,
 		Run: func(cmd *cobra.Command, args []string) {
+			verflag.PrintAndExitIfRequested()
+
 			if err := runCommand(cmd, opts, registryOptions...); err != nil {
 				fmt.Fprintf(os.Stderr, "%v\n", err)
 				os.Exit(1)

```

damemi (commented):
Thanks for working on this! I've noticed this output even in things totally unrelated to the scheduler like `make update`:
```
$ make update
Running in silent mode, run with SILENT=false if you want to see script logs.
Running in short-circuit mode; run with FORCE_ALL=true to force all scripts to run.
Running update-generated-protobuf
Running update-codegen
Running update-generated-runtime
Running update-generated-device-plugin
Running update-generated-api-compatibility-data
Running update-generated-docs
I0810 15:19:57.085423    7075 registry.go:173] Registering SelectorSpread plugin
I0810 15:19:57.085511    7075 registry.go:173] Registering SelectorSpread plugin
```

I checked out your branch here and can confirm this message no longer shows with `make update` either

zhouya0:
@Huang-Wei Actually it already have:
https://github.com/kubernetes/kubernetes/blob/da5ec168e71d199721bab32caee565386ff47306/cmd/kube-scheduler/app/server.go#L120-L123
and actually can't resolve this issue since `klog` happens in adding flag phase. 
@adtac Add a verbosity filter can solve this but will break log info in newer versions.

pkg/scheduler/algorithmprovider/registry.go
zhouya0: I think, don't return `r` just make it more complex like:
```
r := NewRegistry()
r.ApplyFeatureGates()
```
Why not just `r.NewRegistry().ApplyFeatureGates()`?

Huang-Wei:
@zhouya0 sry for misreading that.

I think the root cause is that when configuring the flags, in particular when initializing the deprecated flags, it called:

https://github.com/kubernetes/kubernetes/blob/04362870ad9f9812f487ab8f89a40f6bf2daa588/pkg/scheduler/algorithmprovider/registry.go#L67-L69

which in turns called `NewRegistry()`. However, as `ListAlgorithmProviders()` is aimed to simply return the available in-tree algorithm provider strings, and also the providers are hard-coded in the end of `NewRegistry()`, so a simple fix is just to change the implementation of `ListAlgorithmProviders()`:

```go
return fmt.Sprintf("%s | %s", ClusterAutoscalerProvider, schedulerapi.SchedulerDefaultProviderName)
```

zhouya0:
@Huang-wei Agree with the simple fix. Done.

Huang-Wei:
Thanks!

/lgtm
/approve
/retest

Huang-Wei:
@zhouya0 would you mind updating the PR title as well as the description?

PR #93817, opened 2020-08-08
fix test/e2e_node staticcheck

habibrosyad:
**What type of PR is this?**

/kind cleanup

Fixes `test/e2e_node` staticcheck errors.

Part of #92402

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e_node/gke_environment_test.go
+++ b/test/e2e_node/gke_environment_test.go
@@ -360,7 +360,6 @@ var _ = framework.KubeDescribe("GKE system requirements [NodeConformance][Featur
 		framework.ExpectNoError(checkDockerSeccomp())
 	})
 	ginkgo.It("The docker storage driver should work", func() {
-		e2eskipper.Skipf("GKE does not currently require overlay")
 		framework.ExpectNoError(checkDockerStorageDriver())
 	})
 })
--- a/hack/.staticcheck_failures
+++ b/hack/.staticcheck_failures
@@ -7,7 +7,6 @@ pkg/volume/azure_dd
 pkg/volume/gcepd
 pkg/volume/testing
 test/e2e/autoscaling
-test/e2e_node
 test/integration/examples
 test/integration/framework
 test/integration/garbagecollector


test/e2e_node/gke_environment_test.go
```diff
@@ -303,6 +303,7 @@ func checkDockerSeccomp() error {
 
 // checkDockerStorageDriver checks whether the current storage driver used by
 // Docker is overlay.
+//lint:ignore U1000 referenced in other functions in this file.
```
BenTheElder, line 306: if it is, why isn't the test passing?

habibrosyad: It's being called here, https://github.com/kubernetes/kubernetes/blob/4a25f6ffcdbf97859c11f38cc2f78d11661a1e67/test/e2e_node/gke_environment_test.go#L365
However, somehow staticcheck complain that it's unused. Maybe its related with several issues around `U1000` checks, e.g. [this](https://github.com/dominikh/go-tools/issues/726) and [this](https://github.com/dominikh/go-tools/issues/768) which might be solved by this [commit](https://github.com/dominikh/go-tools/commit/5cfc85b70e7b778eb76fd7338e538d7c9af21e4e) which I think is part of release 2020.1.5.

SergeyKanzhelev: So the root cause if the skipped testcase. And the question is whether we need to keep it skipped or remove. Test is created back in 2017 https://github.com/kubernetes/kubernetes/pull/49443. 

Let me check whether it's an overlay driver now...

If not, perhaps better fix would be to not skip the test, but instead check that this is NOT an overlay driver.

SergeyKanzhelev: ok. it seems that the better fix will be to remove this line: 

`e2eskipper.Skipf("GKE does not currently require overlay")`

GKE seems to support overlay2 now.

test/e2e_node/util_xfs_unsupported.go
```diff
@@ -18,6 +18,7 @@ limitations under the License.
 
 package e2enode
 
+//lint:ignore U1000 applied to non-linux build constraints.
```
SergeyKanzhelev, line 21: I'm not sure why this is required. For instance, the neighbor file has a definition for the `IsCgroup2UnifiedMode` that doesn't need the same ignore from linter.

habibrosyad: Yes, I found that this weird too. I'm not that familiar with the linter, but somehow without that line, the linter complains about `isXfs` not being used. The only difference with `IsCgroup2UnifiedMode` in the neighbor file is `isXfs` is not exported.

test/e2e_node/gke_environment_test.go
habibrosyad: ok, will do that...

test/e2e_node/util_xfs_unsupported.go
habibrosyad: It turns out to be an issue in my local.

SergeyKanzhelev:
/assign @dims 

for hack/OWNERS

PR #93823, opened 2020-08-09
make function more readable

### LINKED ISSUE #93744: scheduler PreFilter NodePorts plugins  clean up getContainerPorts Loop  naming

alam0rt:
**What type of PR is this?**
/kind cleanup

This PR makes the  _getContainerPorts()_ function more readable

Fixes #93744

/cc @jjmengze

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/framework/plugins/nodeports/node_ports.go
+++ b/pkg/scheduler/framework/plugins/nodeports/node_ports.go
@@ -61,10 +61,10 @@ func (pl *NodePorts) Name() string {
 func getContainerPorts(pods ...*v1.Pod) []*v1.ContainerPort {
 	ports := []*v1.ContainerPort{}
 	for _, pod := range pods {
-		for j := range pod.Spec.Containers {
-			container := &pod.Spec.Containers[j]
-			for k := range container.Ports {
-				ports = append(ports, &container.Ports[k])
+		for _, container := range pod.Spec.Containers {
+			for i := range container.Ports {
+				port := &container.Ports[i]
+				ports = append(ports, port)
 			}
 		}
 	}
--- a/pkg/scheduler/framework/plugins/nodeports/node_ports.go
+++ b/pkg/scheduler/framework/plugins/nodeports/node_ports.go
@@ -60,8 +60,7 @@ func (pl *NodePorts) Name() string {
 // will be in the result; but it does not resolve port conflict.
 func getContainerPorts(pods ...*v1.Pod) []*v1.ContainerPort {
 	ports := []*v1.ContainerPort{}
-	for i := range pods {
-		pod := pods[i]
+	for _, pod := range pods {
 		for j := range pod.Spec.Containers {
 			container := &pod.Spec.Containers[j]
 			for k := range container.Ports {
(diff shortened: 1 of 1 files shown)


pkg/scheduler/framework/plugins/nodeports/node_ports.go
```diff
@@ -61,10 +61,10 @@ func (pl *NodePorts) Name() string {
 func getContainerPorts(pods ...*v1.Pod) []*v1.ContainerPort {
 	ports := []*v1.ContainerPort{}
 	for _, pod := range pods {
-		for j := range pod.Spec.Containers {
-			container := &pod.Spec.Containers[j]
-			for k := range container.Ports {
-				ports = append(ports, &container.Ports[k])
+		for _, container := range pod.Spec.Containers {
```
habibrosyad, line 64: I think this is a duplicate of https://github.com/kubernetes/kubernetes/pull/93745 which has already received several feedbacks.

alam0rt: Thanks for your feedback! I have tried to take the point about not wanting to create _new_ references into consideration. Not sure if this is any more readable.

pkg/scheduler/framework/plugins/nodeports/node_ports.go
```diff
@@ -64,7 +64,8 @@ func getContainerPorts(pods ...*v1.Pod) []*v1.ContainerPort {
 		for j := range pod.Spec.Containers {
 			container := &pod.Spec.Containers[j]
 			for k := range container.Ports {
-				ports = append(ports, &container.Ports[k])
+				port := &container.Ports[k]
+				ports = append(ports, port)
```
Huang-Wei, line 68: I don't quite think it's more readable - `&container.Ports[k]` is already simple enough.

pkg/scheduler/framework/plugins/nodeports/node_ports.go
Huang-Wei: Not worth introducing this change IMO: https://github.com/kubernetes/kubernetes/pull/93823#discussion_r468085964

pkg/scheduler/framework/plugins/nodeports/node_ports.go
alam0rt: Yeah I feel you're right. Closing

PR #93829, opened 2020-08-09
e2e services wait for endpoint and endpoint slices

aojea:
Since 1.19 endpoint slices is enabled by default, so all the e2e
tests should consider them.

Similar to https://github.com/kubernetes/kubernetes/pull/93597

I also found another test that need to be adapted
k8s.io/kubernetes/test/e2e/network/service_latency.go

https://github.com/kubernetes/kubernetes/blob/62b8faed4042e9f1a9c1e895ad1c36bb473eb6ee/test/e2e/network/service_latency.go#L295

```release-note
NONE
```

/kind flake

### DIFF
--- a/test/e2e/framework/service/jig.go
+++ b/test/e2e/framework/service/jig.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/onsi/ginkgo"
 	v1 "k8s.io/api/core/v1"
+	discoveryv1beta1 "k8s.io/api/discovery/v1beta1"
 	policyv1beta1 "k8s.io/api/policy/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -352,6 +353,8 @@ func (j *TestJig) waitForAvailableEndpoint(timeout time.Duration) error {
 	endpointSelector := fields.OneTermEqualSelector("metadata.name", j.Name)
 	stopCh := make(chan struct{})
 	endpointAvailable := false
+	endpointSliceAvailable := false
+
 	var controller cache.Controller
 	_, controller = cache.NewInformer(
 		&cache.ListWatch{
@@ -390,8 +393,54 @@ func (j *TestJig) waitForAvailableEndpoint(timeout time.Duration) error {
 
 	go controller.Run(stopCh)
 
+	// If EndpointSlice API is enabled, then validate if appropriate EndpointSlice objects were also create/updated/deleted.
+	if _, err := j.Client.Discovery().ServerResourcesForGroupVersion(discoveryv1beta1.SchemeGroupVersion.String()); err == nil {
+		var esController cache.Controller
+		_, esController = cache.NewInformer(
+			&cache.ListWatch{
+				ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+					options.LabelSelector = "kubernetes.io/service-name=" + j.Name
+					obj, err := j.Client.DiscoveryV1beta1().EndpointSlices(j.Namespace).List(context.TODO(), options)
+					return runtime.Object(obj), err
+				},
+				WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+					options.LabelSelector = "kubernetes.io/service-name=" + j.Name
+					return j.Client.DiscoveryV1beta1().EndpointSlices(j.Namespace).Watch(context.TODO(), options)
+				},
+			},
+			&discoveryv1beta1.EndpointSlice{},
+			0,
+			cache.ResourceEventHandlerFuncs{
+				AddFunc: func(obj interface{}) {
+					if es, ok := obj.(*discoveryv1beta1.EndpointSlice); ok {
+						// TODO: currently we only consider addreses in 1 slice, but services with
+						// a large number of endpoints (>1000) may have multiple slices. Some slices
+						// with only a few addresses. We should check the addresses in all slices.
+						if len(es.Endpoints) > 0 && len(es.Endpoints[0].Addresses) > 0 {
+							endpointSliceAvailable = true
+						}
+					}
+				},
+				UpdateFunc: func(old, cur interface{}) {
+					if es, ok := cur.(*discoveryv1beta1.EndpointSlice); ok {
+						// TODO: currently we only consider addreses in 1 slice, but services with
+						// a large number of endpoints (>1000) may have multiple slices. Some slices
+						// with only a few addresses. We should check the addresses in all slices.
+						if len(es.Endpoints) > 0 && len(es.Endpoints[0].Addresses) > 0 {
+							endpointSliceAvailable = true
+						}
+					}
+				},
+			},
+		)
+
+		go esController.Run(stopCh)
+
+	} else {
+		endpointSliceAvailable = true
+	}
 	err := wait.Poll(1*time.Second, timeout, func() (bool, error) {
-		return endpointAvailable, nil
+		return endpointAvailable && endpointSliceAvailable, nil
 	})
 	if err != nil {
 		return fmt.Errorf("no subset of available IP address found for the endpoint %s within timeout %v", j.Name, timeout)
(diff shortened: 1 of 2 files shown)


aojea:
[sig-cli] Kubectl client Kubectl client-side validation should create/apply a CR with unknown fields for CRD with no validation schema expand_less | 1m16s
```
test/e2e/kubectl/kubectl.go:1004 Aug  9 10:57:43.854: failed to create CR {"kind":"E2e-test-kubectl-2128-crd","apiVersion":"kubectl.example.com/v1","metadata":{"name":"test-cr"},"a":{"b":[{"c":"d"}]}} in namespace --namespace=kubectl-8200: error running /home/prow/go/src/k8s.io/kubernetes/bazel-bin/cmd/kubectl/linux_amd64_pure_stripped/kubectl --server=https://127.0.0.1:35115 --kubeconfig=/root/.kube/kind-test-config --namespace=kubectl-8200 create --validate=true -f -: Command stdout:  stderr: Unable to connect to the server: context deadline exceeded (Client.Timeout exceeded while awaiting headers)  error: exit status 1
```

test/e2e/framework/service/jig.go
```diff
@@ -390,10 +392,57 @@ func (j *TestJig) waitForAvailableEndpoint(timeout time.Duration) error {
 
 	go controller.Run(stopCh)
 
-	err := wait.Poll(1*time.Second, timeout, func() (bool, error) {
-		return endpointAvailable, nil
-	})
-	if err != nil {
+	// If EndpointSlice API is enabled, then validate if appropriate EndpointSlice objects were also create/updated/deleted.
+	if _, err := j.Client.Discovery().ServerResourcesForGroupVersion(discoveryv1beta1.SchemeGroupVersion.String()); err == nil {
```
wojtek-t, line 396: maybe:

```
if _, err := j.Client.Discovery().ServerResourcesForGroupVersion(discoveryv1beta1.SchemeGroupVersion.String()); err != nil {
  // EndpointSlice API is disabled.
  return nil
}

// everything else unintended thus easier to parse.

**Review by hasheddan — COMMENTED** [review]
/milestone v1.19

File: test/e2e/framework/service/jig.go (hunk 1 above)
Comment (robscott) [reply test/e2e/framework/service/jig.go:396]: So this is unfortunately messy to check... The API being enabled does not actually mean that the controller is running, just that it can be running. Checking to see if an EndpointSlice with the name `kubernetes` existed would tell you that the `EndpointSlice` feature gate was enabled on API Server, but that's about all we've got. So with that said, the check you already have is probably as much as we can do here, but it does mean we may end up waiting indefinitely for EndpointSlices in some cases.

Comment (aojea) [reply test/e2e/framework/service/jig.go:396]: @wojtek-t I'm creating a new variable inside the `if` block ... and the check is different using slices than not

@robscott we had to add this check https://github.com/kubernetes/kubernetes/pull/93597 because it was failing in a job that only enable GA features (thanks to Jordan for adding it). I think that for testing should be fair assume API enable = ESController Running,... when can that not happen?

Comment (robscott) [reply test/e2e/framework/service/jig.go:396]: @aojea I think that is pretty unlikely in the future, but it has happened in the past. In 1.17 the API was enabled by default but the controller and feature gate were not. It wasn't until 1.18 that both were enabled by default. I don't _think_ there are any test environments that would have this kind of weird configuration at this point.

Comment (wojtek-t) [reply test/e2e/framework/service/jig.go:396]: Yeah - I agree that enablement of the API is not the same as controller, but I think that for the purposes of e2e tests I think it's actually a good enough assumption.


> @wojtek-t I'm creating a new variable inside the if block ... and the check is different using slices than not

@aojea  I don't get it. I'm not asking to check any of the implemention. All I'm saying is:
- instead o having
```
if <condition> {
  // contents
}
```
have
```
if !<condition> {
  return nil
}
<contents>
```

Comment (wojtek-t) [reply test/e2e/framework/service/jig.go:396]: Actually - I see what you're saying now. Sorry.

So let me suggest slightly different approach
```
endpointAvailable := false
endpointSlicesAvailable := false

startEndpointInformer(stopCh)
if <api enabled> {
  startEndpointSliceInformer(stopCh)
} else {
  endpointSliceAvailable = true
}

err := wait.Poll(1*time.Second, timeout, func() (bool, error) {
  return endpointAvailable && endpointSliceAvailable, nil
})
...
```

Comment (aojea) [reply test/e2e/framework/service/jig.go:396]: done

File: test/e2e/framework/service/jig.go [hunk 2]
```diff
@@ -390,8 +393,52 @@ func (j *TestJig) waitForAvailableEndpoint(timeout time.Duration) error {
 
 	go controller.Run(stopCh)
 
+	// If EndpointSlice API is enabled, then validate if appropriate EndpointSlice objects were also create/updated/deleted.
+	if _, err := j.Client.Discovery().ServerResourcesForGroupVersion(discoveryv1beta1.SchemeGroupVersion.String()); err == nil {
+		var esController cache.Controller
+		_, esController = cache.NewInformer(
+			&cache.ListWatch{
+				ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+					opts := metav1.ListOptions{
```
wojtek-t, line 402: instead of creating new opts, please overwrite the LabelSelector field in the `options` parameter and use this one

Same in watch.

test/e2e/framework/service/jig.go
```diff
@@ -390,8 +393,52 @@ func (j *TestJig) waitForAvailableEndpoint(timeout time.Duration) error {
...
+					}
+					return j.Client.DiscoveryV1beta1().EndpointSlices(j.Namespace).Watch(context.TODO(), opts)
+				},
+			},
+			&discoveryv1beta1.EndpointSlice{},
+			0,
+			cache.ResourceEventHandlerFuncs{
+				AddFunc: func(obj interface{}) {
+					if es, ok := obj.(*discoveryv1beta1.EndpointSlice); ok {
+						if len(es.Endpoints) > 0 && len(es.Endpoints[0].Addresses) > 0 {
```
wojtek-t, line 420: Is this really the logic we would like to have? There can be more than a single EndpointSlice object.

wojtek-t: Actually - looking into Endpoints logic above it seems we also don't wait until Endpoints is fully propagated, but for it to exist.

That's not ideal imho, but it's at least consistent.

aojea: nah, we should fix it.
is enough to check that the endpoints are ready?

wojtek-t: Technically, i can imagine a service with 1000 pods, but endpoint/endpointslice at this point containing a single one.
I don't think we have enough information at this point to add a comprehensive check.

I would start with something (e.g. what you have) and drop a TODO to introduce a more comprehensive check and wouldn't block this PR on that.

aojea: done

test/e2e/framework/service/jig.go
aojea: done

test/e2e/framework/service/jig.go
```diff
@@ -390,8 +393,56 @@ func (j *TestJig) waitForAvailableEndpoint(timeout time.Duration) error {
 
 	go controller.Run(stopCh)
 
+	// If EndpointSlice API is enabled, then validate if appropriate EndpointSlice objects were also create/updated/deleted.
+	if _, err := j.Client.Discovery().ServerResourcesForGroupVersion(discoveryv1beta1.SchemeGroupVersion.String()); err == nil {
+		var esController cache.Controller
+		_, esController = cache.NewInformer(
+			&cache.ListWatch{
+				ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+					obj, err := j.Client.DiscoveryV1beta1().EndpointSlices(j.Namespace).List(context.TODO(), metav1.ListOptions{
```
wojtek-t, line 402: It seems I wasn't explicit enough - it's not what I meant.

I meant:
```
options.LabelsSelector = "kubernetes.io/service-name=" + j.Name
obj, err := c.Client.DiscoveryV1beta1().EndpointSlices(j.Names0ace).List(context.TODO(), options)
...


Same below

aojea: ups

aojea: I need more coffee :upside_down_face:

wojtek-t:
We can improve that further, but that's already net improvement. So let's make it.

/lgtm
/approve

PR #93831, opened 2020-08-09
scheduler: forget the pod when the reserve plugins fail

### LINKED ISSUE #93830: Scheduler should forget the pod when the reverse plugins fail
This is a bug found in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93555/pull-kubernetes-integration/1292393344655167493.

The scheduler fails to schedule the pod `pv-w-prebound` on the first try due to PV is out of sync.

```
I0809 10:06:11.676940  121717 scheduler_binder.go:349] AssumePodVolumes for pod "volume-scheduling-67333a8a-4c60-4a09-9da6-9a8d731fdeb6/pod-w-pv-prebound", node "node-1"
E0809 10:06:11.676993  121717 framework.go:800] error while running Reserve in "VolumeBinding" reserve plugin for pod "pod-w-pv-prebound": v1.PersistentVolume "pv-w-prebound" is out of sync (stored: 41381, assume: 41237)
E0809 10:06:11.677104  121717 factory.go:461] "Error scheduling pod; retrying" err="error while running Reserve in \"VolumeBinding\" reserve plugin for pod \"pod-w-pv-prebound\": v1.PersistentVolume \"pv-w-prebound\" is out of sync (stored: 41381, assume: 41237)" pod="volume-scheduling-67333a8a-4c60-4a09-9da6-9a8d731fdeb6/pod-w-pv-prebound"
I0809 10:06:11.677146  121717 scheduler.go:347] Updating pod condition for volume-scheduling-67333a8a-4c60-4a09-9da6-9a8d731fdeb6/pod-w-pv-prebound to (PodScheduled==False, Reason=SchedulerError)
...
I0809 10:06:11.685464  121717 scheduler_assume_cache.go:341] Restored v1.PersistentVolume "pv-w-prebound"
```

This may happen because the PV/PVC may be modified before the scheduler binder assumes them. 

However, the scheduler skips the pod update event and never try to scheduler the pod again:

```
I0809 10:06:11.685487  121717 eventhandlers.go:356] Skipping pod volume-scheduling-67333a8a-4c60-4a09-9da6-9a8d731fdeb6/pod-w-pv-prebound update
```

This is because when the reserve plugins fails, we didn't try to forget the pod. After the `sched.recordSchedulingFailure` is called, the pod is in the `assumed` state.

cofyc:
**What type of PR is this?**
/kind bug

Fixes #93830

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -422,9 +422,6 @@ func (sched *Scheduler) finishBinding(prof *profile.Profile, assumed *v1.Pod, ta
 	}
 	if err != nil {
 		klog.V(1).Infof("Failed to bind pod: %v/%v", assumed.Namespace, assumed.Name)
-		if err := sched.SchedulerCache.ForgetPod(assumed); err != nil {
-			klog.Errorf("scheduler cache ForgetPod failed: %v", err)
-		}
 		return
 	}
 
@@ -504,22 +501,25 @@ func (sched *Scheduler) scheduleOne(ctx context.Context) {
 	// assume modifies `assumedPod` by setting NodeName=scheduleResult.SuggestedHost
 	err = sched.assume(assumedPod, scheduleResult.SuggestedHost)
 	if err != nil {
+		metrics.PodScheduleError(prof.Name, metrics.SinceInSeconds(start))
 		// This is most probably result of a BUG in retrying logic.
 		// We report an error here so that pod scheduling can be retried.
 		// This relies on the fact that Error will check if the pod has been bound
 		// to a node and if so will not add it back to the unscheduled pods queue
 		// (otherwise this would cause an infinite loop).
 		sched.recordSchedulingFailure(prof, assumedPodInfo, err, SchedulerError, "")
-		metrics.PodScheduleError(prof.Name, metrics.SinceInSeconds(start))
 		return
 	}
 
 	// Run the Reserve method of reserve plugins.
 	if sts := prof.RunReservePluginsReserve(schedulingCycleCtx, state, assumedPod, scheduleResult.SuggestedHost); !sts.IsSuccess() {
-		sched.recordSchedulingFailure(prof, assumedPodInfo, sts.AsError(), SchedulerError, "")
 		metrics.PodScheduleError(prof.Name, metrics.SinceInSeconds(start))
 		// trigger un-reserve to clean up state associated with the reserved Pod
 		prof.RunReservePluginsUnreserve(schedulingCycleCtx, state, assumedPod, scheduleResult.SuggestedHost)
+		if forgetErr := sched.Cache().ForgetPod(assumedPod); forgetErr != nil {
+			klog.Errorf("scheduler cache ForgetPod failed: %v", forgetErr)
+		}
+		sched.recordSchedulingFailure(prof, assumedPodInfo, sts.AsError(), SchedulerError, "")
 		return
 	}
 
@@ -587,6 +587,9 @@ func (sched *Scheduler) scheduleOne(ctx context.Context) {
 			metrics.PodScheduleError(prof.Name, metrics.SinceInSeconds(start))
 			// trigger un-reserve plugins to clean up state associated with the reserved Pod
 			prof.RunReservePluginsUnreserve(bindingCycleCtx, state, assumedPod, scheduleResult.SuggestedHost)
+			if err := sched.SchedulerCache.ForgetPod(assumedPod); err != nil {
+				klog.Errorf("scheduler cache ForgetPod failed: %v", err)
+			}
 			sched.recordSchedulingFailure(prof, assumedPodInfo, fmt.Errorf("Binding rejected: %v", err), SchedulerError, "")
 		} else {
 			// Calculating nodeResourceString can be heavy. Avoid it if klog verbosity is below 2.
--- a/pkg/scheduler/scheduler_test.go
+++ b/pkg/scheduler/scheduler_test.go
@@ -214,17 +214,57 @@ func TestSchedulerScheduleOne(t *testing.T) {
 	errB := errors.New("binder")
 
 	table := []struct {
-		name             string
-		injectBindError  error
-		sendPod          *v1.Pod
-		algo             core.ScheduleAlgorithm
-		expectErrorPod   *v1.Pod
-		expectForgetPod  *v1.Pod
-		expectAssumedPod *v1.Pod
-		expectError      error
-		expectBind       *v1.Binding
-		eventReason      string
+		name                string
+		injectBindError     error
+		sendPod             *v1.Pod
+		algo                core.ScheduleAlgorithm
+		registerPluginFuncs []st.RegisterPluginFunc
+		expectErrorPod      *v1.Pod
+		expectForgetPod     *v1.Pod
+		expectAssumedPod    *v1.Pod
+		expectError         error
+		expectBind          *v1.Binding
+		eventReason         string
 	}{
+		{
+			name:    "error reserve pod",
+			sendPod: podWithID("foo", ""),
+			algo:    mockScheduler{core.ScheduleResult{SuggestedHost: testNode.Name, EvaluatedNodes: 1, FeasibleNodes: 1}, nil},
+			registerPluginFuncs: []st.RegisterPluginFunc{
+				st.RegisterReservePlugin("FakeReserve", st.NewFakeReservePlugin(framework.NewStatus(framework.Error, "reserve error"))),
+			},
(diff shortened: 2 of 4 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -516,10 +516,13 @@ func (sched *Scheduler) scheduleOne(ctx context.Context) {
...
 	// Run the Reserve method of reserve plugins.
 	if sts := prof.RunReservePluginsReserve(schedulingCycleCtx, state, assumedPod, scheduleResult.SuggestedHost); !sts.IsSuccess() {
-		sched.recordSchedulingFailure(prof, assumedPodInfo, sts.AsError(), SchedulerError, "")
 		metrics.PodScheduleError(prof.Name, metrics.SinceInSeconds(start))
 		// trigger un-reserve to clean up state associated with the reserved Pod
 		prof.RunReservePluginsUnreserve(schedulingCycleCtx, state, assumedPod, scheduleResult.SuggestedHost)
+		if forgetErr := sched.Cache().ForgetPod(assumedPod); forgetErr != nil {
+			klog.Errorf("scheduler cache ForgetPod failed: %v", forgetErr)
+		}
+		sched.recordSchedulingFailure(prof, assumedPodInfo, sts.AsError(), SchedulerError, "")
```
cofyc, line 525: - put `recordSchedulingFailure` after the `Forget` because if the re-enqueuing & pop-out of that failed Pod happens super fast, the pod update event may still be skipped (see https://github.com/kubernetes/kubernetes/pull/93511#issuecomment-665316132)
- put `ForgetPod` after `RunReservePluginsUnreserve` for consistency (assume -> reserve -> unreserve -> unassume)

ahg-g (commented):
@Huang-Wei should we do that at line 592 as well when binding fails?

@cofyc thanks for the fix, can you please add a unit test?

Huang-Wei:
> @Huang-Wei should we do that at line 592 as well when binding fails?

Yes, we should. It's discussed in https://github.com/kubernetes/kubernetes/pull/93511#issuecomment-666019546. Do you think we should put in in 1.19?

@cofyc thanks for the fix! It was [observed](https://github.com/kubernetes/kubernetes/pull/93511#issuecomment-665833750) actually when #93511 was introduced.

cofyc:
> @cofyc thanks for the fix, can you please add a unit test?

I've added a case in `TestSchedulerScheduleOne` to verify that the pod will be unassumed when reserve plugins fail.

However, the race may happen if we didn't follow the order (`assume -> reserve -> unreserve -> unassume -> recordSchedulingFailure`) is hard to verify in the UT, but our integration tests (e.g. TestVolumeBinding) can help.

cofyc:
> @cofyc thanks for the fix! It was [observed](https://github.com/kubernetes/kubernetes/pull/93511#issuecomment-665833750) actually when #93511 was introduced.

sorry, I didn't look closely at these failures. The `TestVolumeBining` integration test had been flaky for a while (https://github.com/kubernetes/kubernetes/issues/81368). I thought It was for the same reason.

ahg-g:
Yes, I think we should do it in this PR.

pkg/scheduler/scheduler.go
```diff
@@ -516,10 +516,13 @@ func (sched *Scheduler) scheduleOne(ctx context.Context) {
```
adtac, line 516: L513: we should probably make `metrics.PodScheduleError` calls be consistently ordered too while we're at it? It's metrics -> recordSchedulingFailure everywhere else except L513.

cofyc: SGTM

cofyc:
> > Yes, we should. It's discussed in #93511 (comment). Do you think we should put in in 1.19?
> 
> Yes, I think we should do it in this PR.

I will implement it in this PR.

pkg/scheduler/scheduler_test.go
```diff
@@ -296,10 +311,14 @@ func TestSchedulerScheduleOne(t *testing.T) {
...
 			})
-			fwk, err := st.NewFramework([]st.RegisterPluginFunc{
+			registerPluginFuncs := []st.RegisterPluginFunc{
 				st.RegisterQueueSortPlugin(queuesort.Name, queuesort.New),
 				st.RegisterBindPlugin(defaultbinder.Name, defaultbinder.New),
-			}, frameworkruntime.WithClientSet(client))
+			}
+			if len(item.registerPluginFuncs) > 0 {
+				registerPluginFuncs = append(registerPluginFuncs, item.registerPluginFuncs...)
+			}
```
Huang-Wei, line 320: nit:

```go
registerPluginFuncs := append(
	item.registerPluginFuncs,
	st.RegisterQueueSortPlugin(queuesort.Name, queuesort.New),
	st.RegisterBindPlugin(defaultbinder.Name, defaultbinder.New),
)
```

Huang-Wei (commented):
A nit in the test. LGTM otherwise.

Huang-Wei:
> I will implement it in this PR.

Thanks!

pkg/scheduler/scheduler.go
cofyc: done

pkg/scheduler/scheduler.go
```diff
@@ -587,6 +587,9 @@ func (sched *Scheduler) scheduleOne(ctx context.Context) {
 			metrics.PodScheduleError(prof.Name, metrics.SinceInSeconds(start))
 			// trigger un-reserve plugins to clean up state associated with the reserved Pod
 			prof.RunReservePluginsUnreserve(bindingCycleCtx, state, assumedPod, scheduleResult.SuggestedHost)
+			if err := sched.SchedulerCache.ForgetPod(assumedPod); err != nil {
+				klog.Errorf("scheduler cache ForgetPod failed: %v", err)
+			}
```
cofyc, line 592: move `ForgetPod` from `finishBinding` here to follow the function call order on failure (unreserve -> unassume -> recordSchedulingFailure)

pkg/scheduler/scheduler.go
```diff
@@ -504,22 +501,25 @@ func (sched *Scheduler) scheduleOne(ctx context.Context) {
 	// assume modifies `assumedPod` by setting NodeName=scheduleResult.SuggestedHost
 	err = sched.assume(assumedPod, scheduleResult.SuggestedHost)
 	if err != nil {
+		metrics.PodScheduleError(prof.Name, metrics.SinceInSeconds(start))
```
cofyc, line 504: suggested by @adtac, call metrics function before all other functions for consistency. this aligns with other places.

cofyc:
@Huang-Wei @ahg-g @adtac comments addressed

btw, two more test cases for permit and prebind phase

cofyc:
(run more times)

pkg/scheduler/testing/fake_plugins.go
```diff
@@ -152,3 +153,80 @@ func NewFakePreFilterPlugin(status *framework.Status) frameworkruntime.PluginFac
...
+// Permit invoked at the Permit extension point.
+func (pl *FakePermitPlugin) Permit(_ context.Context, _ *framework.CycleState, _ *v1.Pod, _ string) (*framework.Status, time.Duration) {
+	return pl.Status, pl.Timeout
+}
+
+// NewFakePermitPlugin initializes a fakePermitPlugin and returns it.
+func NewFakePermitPlugin(status *framework.Status, timeout time.Duration) frameworkruntime.PluginFactory {
+	return func(_ runtime.Object, _ framework.FrameworkHandle) (framework.Plugin, error) {
+		return &FakePermitPlugin{
+			Status: status,
```
adtac, line 229: Timeout is not being saved here.

cofyc: fixed, thanks!

adtac:
LGTM :+1:

hasheddan (commented):
/milestone v1.19

pkg/scheduler/scheduler.go
```diff
@@ -422,9 +422,6 @@ func (sched *Scheduler) finishBinding(prof *profile.Profile, assumed *v1.Pod, ta
 	}
 	if err != nil {
 		klog.V(1).Infof("Failed to bind pod: %v/%v", assumed.Namespace, assumed.Name)
-		if err := sched.SchedulerCache.ForgetPod(assumed); err != nil {
-			klog.Errorf("scheduler cache ForgetPod failed: %v", err)
-		}
```
Huang-Wei, line 427: Good removal. This ensures we run `ForgetPod()` once, and ensure it's called at the end the failure path instead of `defer finishBinding()` in `bind()`.

Issue #93792, opened 2020-08-07
new scheduler_perf framework: allow users to specify default pod and node specs


adtac:
Note: this issue should be viewed in the context of https://github.com/kubernetes/kubernetes/pull/93252.

**What would you like to be added**: Each workload currently requires the user to specify the node/pod specs explicitly. While this is sometimes useful in certain workloads, for around half the workloads, all benchmarks use the same pod and node specs across all `CreateNodes` and `CreatePods` ops. It'd be nice to specify these exactly once at the head of the testCase.

**Why is this needed**: Reduce config toil.

sanposhiho:
Hello, I'd like to take this issue
/assign

kapiljain1989:
I want to understand about this task, what we want to achieve?

kapiljain1989:
@adtac Could you please help me in this, Just little brief what we want to do in this?

sanposhiho:
@kapiljain1989 

This issue will be finished in https://github.com/kubernetes/kubernetes/pull/101799.

kapiljain1989:
@sanposhiho Thank you for the update and also it help me to understood what we wanna achieve in this task :)

PR #93452, opened 2020-07-26
[WIP] Run UDP-traffic-preservation-nodeport test in a loop to try to flush out test flakiness

mattfenwick:
**What type of PR is this?**

/kind flake

**What this PR does / why we need it**:

The goal of this PR is to help diagnose the flakiness reported in https://github.com/kubernetes/kubernetes/issues/91236 by running the test additional times to collect more data.

**Which issue(s) this PR fixes**:

This partially addresses https://github.com/kubernetes/kubernetes/issues/91236.

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

**Test plan**

```
make WHAT=test/e2e/e2e.test

export KUBERNETES_SERVICE_HOST=127.0.0.1
export KUBERNETES_SERVICE_PORT=32768

cat << EOF > kind-config.yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
- role: worker
- role: worker
EOF

kind create cluster --name=three-node --config kind-config.yaml

./_output/local/bin/darwin/amd64/e2e.test \
 --provider=local \
 --ginkgo.focus="should be able to preserve UDP traffic when server pod cycles for a NodePort" \
 --kubeconfig=/Users/mfenwick/.kube/config \
 --dump-logs-on-failure=false
```

PR #93837, opened 2020-08-09
intra-pod test improvements for bug-triage and node-level reporting

### LINKED ISSUE #93833: Breadth first polling in intra-pod networking Conformance verification

Right now we poll one node at a time for intra-pod tests.  We ultimately fail a test if any one node is unreachable.

If we instead polled all nodes, multiple times, and collected the number of times a node was unreachable, this would 
- tell us what % of nodes on the cluster were broken, rather then just telling us the first node that was broken.
- not break the testing contract (were still verifying intra-pod connectivity, just doing so in a way that gives us holistic cluster info)

**Why is this needed**:
I described why this would make triage alot easier, and in general make it easier to see why tests are failing, here ...  https://github.com/kubernetes/kubernetes/issues/92618

i'll collect a little more data on this to justify the decision .... as this is a conformance test so the bar for modification is higher (even if the test itself isnt changing) then regular tests.

jayunit100:
**What type of PR is this?**

/kind cleanup

- breadth first polling of pods
- fine grained debugging of pods that fail the first poll attempt
- summary of how many pods failed at the end

1) A breadth first poll is implemented like so, showing each host in the logs.  This way if one or two host are down you can clearly see which ones.

```
Aug 11 12:06:46.829: INFO: Breadth first check of 192.168.150.148 on host 172.18.0.2...
... 
Aug 11 11:57:27.078: INFO: reached 192.168.150.147 after 22/79 tries
```

This PR also adds a logging statement to clarify the way that "79" is calculated (as a function of the number of nodes ^2 + 30 :))... no clue where that all comes from. 
2) So, the logs now summarize:
```
Aug 11 12:06:47.169: INFO: reached 192.168.217.84 after 0/1 tries
Aug 11 12:06:47.169: INFO: Going to retry 0 out of 7 pods....
```
Fixes #93833 

```release-note
     Added fine grained debugging to the intra-pod conformance test for helping easily resolve networking issues for nodes that might be unhealthy when running conformance or sonobuoy tests.
```

### DIFF
--- a/test/e2e/common/networking.go
+++ b/test/e2e/common/networking.go
@@ -50,12 +50,12 @@ var _ = ginkgo.Describe("[sig-network] Networking", func() {
 					framework.Logf("Warning: Test failure (%v) will occur due to %v", len(errors)+1, err) // convenient error message for diagnosis... how many pods failed, and on what hosts?
 				}
 			}
-			if len(errors) == 0 {
+			if len(errors) > 0 {
 				framework.Logf("Pod polling failure summary:")
 				for _, e := range errors {
 					framework.Logf("%v", e)
 				}
-				framework.Failf("Failed due to %v errors polling %v pods", errors, len(config.EndpointPods))
+				framework.Failf("Failed due to %v errors polling %v pods", len(errors), len(config.EndpointPods))
 			}
 		})
 
--- a/test/e2e/common/networking.go
+++ b/test/e2e/common/networking.go
@@ -39,8 +39,23 @@ var _ = ginkgo.Describe("[sig-network] Networking", func() {
 		*/
 		framework.ConformanceIt("should function for intra-pod communication: http [NodeConformance]", func() {
 			config := e2enetwork.NewCoreNetworkingTestConfig(f, false)
+
+			// Extra debugging info since this is the most common diagnostic for failing clusters, and is a Conformance test.
+			errors := []error{}
 			for _, endpointPod := range config.EndpointPods {
-				config.DialFromTestContainer("http", endpointPod.Status.PodIP, e2enetwork.EndpointHTTPPort, config.MaxTries, 0, sets.NewString(endpointPod.Name))
+				if err := config.DialFromTestContainer("http", endpointPod.Status.PodIP, e2enetwork.EndpointHTTPPort, config.MaxTries, 0, sets.NewString(endpointPod.Name)); err != nil {
+					errors = append(errors, err)
+					framework.Logf("Was able to reach %v on %v ", endpointPod.Status.PodIP, endpointPod.Status.HostIP)
+				} else {
+					framework.Logf("Warning: Test failure (%v) will occur due to %v", len(errors)+1, err) // convenient error message for diagnosis... how many pods failed, and on what hosts?
+				}
+			}
+			if len(errors) == 0 {
+				framework.Logf("Pod polling failure summary:")
+				for _, e := range errors {
+					framework.Logf("%v", e)
+				}
+				framework.Failf("Failed due to %v errors polling %v pods", errors, len(config.EndpointPods))
 			}
 		})
 
--- a/test/e2e/common/networking.go
+++ b/test/e2e/common/networking.go
@@ -28,6 +28,25 @@ var _ = ginkgo.Describe("[sig-network] Networking", func() {
 
 	ginkgo.Describe("Granular Checks: Pods", func() {
 
+		checkNodeConnectivity := func(config *e2enetwork.NetworkingTestConfig, protocol string, port int) {
+			errors := []error{}
+			for _, endpointPod := range config.EndpointPods {
+				if err := config.DialFromTestContainer(protocol, endpointPod.Status.PodIP, port, config.MaxTries, 0, sets.NewString(endpointPod.Name)); err != nil {
+					errors = append(errors, err)
+					framework.Logf("Warning: Test failure (%v) will occur due to %v", len(errors)+1, err) // convenient error message for diagnosis... how many pods failed, and on what hosts?
+				} else {
+					framework.Logf("Was able to reach %v on %v ", endpointPod.Status.PodIP, endpointPod.Status.HostIP)
+				}
+			}
+			if len(errors) > 0 {
+				framework.Logf("Pod polling failure summary:")
+				for _, e := range errors {
+					framework.Logf("%v", e)
+				}
+				framework.Failf("Failed due to %v errors polling %v pods", len(errors), len(config.EndpointPods))
+			}
+		}
+
 		// Try to hit all endpoints through a test container, retry 5 times,
 		// expect exactly one unique hostname. Each of these endpoints reports
 		// its own hostname.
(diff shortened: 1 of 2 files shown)


mattfenwick:
Test plan (note to self, [see here](https://github.com/kubernetes/kubernetes/pull/93452)):

```
make WHAT=test/e2e/e2e.test

export KUBERNETES_SERVICE_HOST=127.0.0.1
export KUBERNETES_SERVICE_PORT=32768

cat << EOF > three-node-kind.yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
- role: worker
- role: worker
EOF

kind create cluster --name=three-node --config three-node-kind.yaml

./_output/local/bin/darwin/amd64/e2e.test \
 --provider=local \
 --ginkgo.focus="should function for intra-pod communication: http [NodeConformance]" \
 --kubeconfig=/Users/mfenwick/.kube/config \
 --dump-logs-on-failure=false
```

jayunit100:
cc @bowei not quite 'breadth first', but for lack of a better word, it does the trick - scopes out all the pods and gives a ratio of how many connections failed.

jayunit100:
#93792 also will be easier to address once we have this since it tells exactly how many intrapod tests are failing per node .

test/e2e/common/networking.go
```diff
@@ -80,9 +115,7 @@ var _ = ginkgo.Describe("[sig-network] Networking", func() {
 		*/
 		framework.ConformanceIt("should function for node-pod communication: udp [LinuxOnly] [NodeConformance]", func() {
 			config := e2enetwork.NewCoreNetworkingTestConfig(f, true)
-			for _, endpointPod := range config.EndpointPods {
-				config.DialFromNode("udp", endpointPod.Status.PodIP, e2enetwork.EndpointUDPPort, config.MaxTries, 0, sets.NewString(endpointPod.Name))
-			}
+			checkNodeConnectivity(config, "udp", e2enetwork.EndpointUDPPort)
```
aojea, line 118: this called `DialFromNode` and now calls `DialFromTestContainer` from `checkNodeConnectivity` :smiley:

jayunit100: whoops :) fixing

test/e2e/common/networking.go
```diff
@@ -28,6 +29,46 @@ var _ = ginkgo.Describe("[sig-network] Networking", func() {
 
 	ginkgo.Describe("Granular Checks: Pods", func() {
 
+		checkNodeConnectivity := func(config *e2enetwork.NetworkingTestConfig, protocol string, port int) {
+			// breadth first poll to quickly estimate failure.
+			failedPodsByHost := map[string][]*v1.Pod{}
+			// First time, we'll quickly try all pods, breadth first.
+			for _, endpointPod := range config.EndpointPods {
```
jayunit100, line 36: ... note, this first 'shot' doesn't change the test semantics, but it allows us to separate the test output between successful nodes , and potentially problematic ones, meaning its very natural to browse the diagnostic results.

test/e2e/common/networking.go
```diff
@@ -28,6 +29,46 @@ var _ = ginkgo.Describe("[sig-network] Networking", func() {
...
+				framework.Logf("Breadth first check of %v on host %v...", endpointPod.Status.PodIP, endpointPod.Status.HostIP)
+				if err := config.DialFromTestContainer(protocol, endpointPod.Status.PodIP, port, 1, 0, sets.NewString(endpointPod.Name)); err != nil {
+					if _, ok := failedPodsByHost[endpointPod.Status.HostIP]; !ok {
+						failedPodsByHost[endpointPod.Status.HostIP] = []*v1.Pod{}
+					}
+					failedPodsByHost[endpointPod.Status.HostIP] = append(failedPodsByHost[endpointPod.Status.HostIP], endpointPod)
+					framework.Logf("...failed...will try again in next pass")
+				}
+			}
+			errors := []error{}
```
jayunit100, line 46: this struct is aggregated later, so that the end resulting test , all errors are printed out together, and ratio  of polls that failed is available , so that you can clearly know wether a cluster is in catastrophic state or wether there's just a few problematic nodes

test/e2e/common/networking.go
```diff
@@ -28,6 +29,46 @@ var _ = ginkgo.Describe("[sig-network] Networking", func() {
...
+			// Second time, we pass through pods more carefully...
+			framework.Logf("Going to retry %v out of %v pods....", len(failedPodsByHost), len(config.EndpointPods))
+			for host, failedPods := range failedPodsByHost {
```
jayunit100, line 49: if *one* node is down, all the pods from that node will be shown together.

jayunit100: (imo) this isn't fundamentally changing the tests because, there is no order guarantee in the conformance definition.  that said, if we want it randomly ordered by pod, i can still do that explicitly.  

Note that the default is going to be polling like 80 times for a 6 node cluster (nodes^2 + 30) , so this test is *highly stable* in terms of retries, so the ordering isn't a sensitive thing at all.

fejta:
/priority backlog
/approve
for test, please have someone from networking do the review

test/e2e/common/networking.go
```diff
@@ -28,6 +29,46 @@ var _ = ginkgo.Describe("[sig-network] Networking", func() {
...
+						framework.Logf("Was able to reach %v on %v ", endpointPod.Status.PodIP, endpointPod.Status.HostIP)
+					}
+					framework.Logf("... Done probing pod [[[ %v ]]]", endpointPod.Status.PodIP)
+				}
+				framework.Logf("succeeded at polling %v out of %v connections", len(config.EndpointPods)-len(errors), len(config.EndpointPods))
+			}
+			if len(errors) > 0 {
+				framework.Logf("pod polling failure summary:")
+				for _, e := range errors {
+					framework.Logf("Collected error: %v", e)
```
oomichi, line 66: This log message seems duplicated with line 55 to explain what error happens on each DialFromTestContainer() call.

test/e2e/framework/network/utils.go
```diff
@@ -248,7 +248,9 @@ func makeCURLDialCommand(ipPort, dialCmd, protocol, targetIP string, targetPort
 // maxTries == minTries will confirm that we see the expected endpoints and no
 // more for maxTries. Use this if you want to eg: fail a readiness check on a
 // pod and confirm it doesn't show up as an endpoint.
-func (config *NetworkingTestConfig) DialFromContainer(protocol, dialCommand, containerIP, targetIP string, containerHTTPPort, targetPort, maxTries, minTries int, expectedResponses sets.String) {
+//
```
oomichi, line 251: nit: unnecessary blank line.

oomichi (commented):
/cc @oomichi

test/e2e/common/networking.go
mattfenwick: 👍

test/e2e/framework/network/utils.go
mattfenwick: 👍

oomichi:
Thanks for updating.

/lgtm

test/e2e/framework/network/utils.go
```diff
@@ -167,17 +167,17 @@ type NetexecDialResponse struct {
 }
 
 // DialFromEndpointContainer executes a curl via kubectl exec in an endpoint container.
-func (config *NetworkingTestConfig) DialFromEndpointContainer(protocol, targetIP string, targetPort, maxTries, minTries int, expectedEps sets.String) {
-	config.DialFromContainer(protocol, echoHostname, config.EndpointPods[0].Status.PodIP, targetIP, EndpointHTTPPort, targetPort, maxTries, minTries, expectedEps)
+func (config *NetworkingTestConfig) DialFromEndpointContainer(protocol, targetIP string, targetPort, maxTries, minTries int, expectedEps sets.String) error {
```
aojea, line 170: I 'm afraid we have to revert this commit, this method is used by other tests, and now instead of failing it returns an error, but the other tests are not asserting the error, so a failure will go unnoticed

https://github.com/kubernetes/kubernetes/blob/0b8c2bf1a10817a0e2b1f9a30737c85f6c65af63/test/e2e/network/networking.go#L211-L258

aojea:
it was my fault, I wanted to follow up and hold it before merge, because this is modifying something that is being used by other tests, and I' m not sure that the gain is worth the effort

see https://github.com/kubernetes/kubernetes/pull/93837/files#r492991061

Issue #93609, opened 2020-07-31
test/integration/master package tests are slow, package times out


liggitt:
**Which jobs are flaking**:

pull-kubernetes-integration

**Which test(s) are flaking**:

test/integration/master

**Testgrid link**:

**Reason for failure**:

There are ~5 tests that average more than 30 seconds each.

https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-integration&graph-metrics=test-duration-minutes&include-filter-by-regex=k8s.io/kubernetes/test/integration/master&exclude-filter-by-regex=%5C%5B1%5C%5D

This is an umbrella issue tracking speeding up these five tests:

- [x] k8s.io/kubernetes/test/integration/master.TestReconcilerMasterLease{MultiMoreMasters,MultiCombined,Combined} - @gitlawr - https://github.com/kubernetes/kubernetes/pull/93721
- [x] k8s.io/kubernetes/test/integration/master.TestAudit - @vareti - https://github.com/kubernetes/kubernetes/pull/94054 
    * suggestion/starting point: rather than setting up a new server and webhook for each iteration, we can set up a single server with a policy that logs differently named configmaps at different verbosity levels (using the resourceNames portion of the audit policy), and a webhook that only intercepts specifically labeled configmaps (using the objectSelector portion of the webhook).
    * Then all the iterations in the table test can run against the same server and webhook, selecting a different configmap name depending on what verbosity we want to exercise, and setting different labels on the configmap depending on whether we want the mutating webhook to intercept it or not.
- [ ] k8s.io/kubernetes/test/integration/master.TestKMSHealthz - needs an owner

**Anything else we need to know**:

https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&text=FAIL%09k8s.io%2Fkubernetes%2Ftest%2Fintegration%2Fmaster

/sig api-machinery
/priority important-soon

liggitt:
Possible approaches:
* when timeouts are configurable, speed them up (example in https://github.com/kubernetes/kubernetes/pull/93394)
* consider refactoring table tests to run in parallel (example in https://github.com/kubernetes/kubernetes/pull/93394)

liggitt:
if you want to help improving the runtime of one of these tests, leave a comment, then link to a WIP PR once it is open

liggitt:
cc @kubernetes/sig-api-machinery-bugs

liggitt:
[...]

PR #86368, opened 2019-12-17
Expose configuration options to control the frequency of kms-plugin's health checks.

immutableT:
**What type of PR is this?**
 /kind api-change
 /kind feature

/priority important-longterm
/milestone v1.18

**What this PR does / why we need it**:
The frequency of at which to perform health checks of kms-plugin may be very greatly based on the the environment.
Concretely, the cost (both monetary and performance) of performing crypto operations against an external KMS may greatly influence how often such checks should be performed.
Thus, this PR exposes the option of configuring the frequency of checks at cluster deployment time.
**Which issue(s) this PR fixes**:

**Does this PR introduce a user-facing change?**:

```release-note
 Yes, users will be able to control the cache TTL of kms-plugin's health check results.

Need to update:
- https://kubernetes.io/docs/tasks/administer-cluster/kms-provider/
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #93839, opened 2020-08-10
Speed up KMSHealthz test

helenfeng737:
**What type of PR is this?**

/kind flake

Speed up KMSHealthz test. 

Part of #93609 

```release-note
None
```

```docs
None
```

### DIFF
--- a/test/integration/master/kms_transformation_test.go
+++ b/test/integration/master/kms_transformation_test.go
@@ -206,9 +206,7 @@ resources:
 		t.Fatalf("failed to create mock of KMS Plugin #1: %v", err)
 	}
 
-	if err := pluginMock1.Start(); err != nil {
-		t.Fatalf("Failed to start kms-plugin, err: %v", err)
-	}
+	go pluginMock1.Start()
 	defer pluginMock1.CleanUp()
 	if err := mock.WaitForBase64PluginToBeUp(pluginMock1); err != nil {
 		t.Fatalf("Failed to start plugin #1, err: %v", err)
@@ -218,9 +216,7 @@ resources:
 	if err != nil {
 		t.Fatalf("Failed to create mock of KMS Plugin #2: err: %v", err)
 	}
-	if err := pluginMock2.Start(); err != nil {
-		t.Fatalf("Failed to start kms-plugin, err: %v", err)
-	}
+	go pluginMock2.Start()
 	defer pluginMock2.CleanUp()
 	if err := mock.WaitForBase64PluginToBeUp(pluginMock2); err != nil {
 		t.Fatalf("Failed to start KMS Plugin #2: err: %v", err)


test/integration/master/kms_transformation_test.go
```diff
@@ -206,9 +206,7 @@ resources:
 		t.Fatalf("failed to create mock of KMS Plugin #1: %v", err)
 	}
 
-	if err := pluginMock1.Start(); err != nil {
-		t.Fatalf("Failed to start kms-plugin, err: %v", err)
-	}
+	go pluginMock1.Start()
```
immutableT, line 209: How much time does this save?

I think this will come down to the trade-off between shaving-off some time vs. lose in the error reporting clarity.

ZhiFeng1993: Very little, like 1-2s. It couldn't bring the time down under 30s(originally 40-45s). I can close this PR if we need to modify strategy.

immutableT: I think the fundamental problem is that we hard-code the TTL for healthchecks:
```
kmsPluginHealthzNegativeTTL  = 3 * time.Second
kmsPluginHealthzPositiveTTL  = 20 * time.Second
```
I had a PR that proposed exposing these values in the EncryptionConfig, but the general opinion was that this was not something we wanted to expose to end-users.
Having these values available in the test setup would probably address this issue of slowness.

immutableT: This test will take at least 20s to run due to the hard-coded value for `kmsPluginHealthzPositiveTTL = 20 * times.Second`.
In other words, to simulate a failure of the kms-plugin (go from healthy to unhealthy), the test would need to wait 20s to confirm the transition.

@liggitt and @enj Should we take another look at #86368 which would address this issue.

liggitt: I'd start by making a test-only way to shorten that rather than jump right to config

immutableT: Sure, we could try the following approach - going from unhealthy to health - this could save 17s since the negative TTL is just 3 seconds. @ZhiFeng1993 does this make sense?

ZhiFeng1993: hmm @immutableT  In this test I don't think either `kmsPluginHealthzNegativeTTL` or `kmsPluginHealthzNegativeTTL` is being used. It's using `mustBeHealthy` and `mustBeUnhealthy` that are defined in this file. See here:
https://github.com/kubernetes/kubernetes/blob/master/test/integration/master/kms_transformation_test.go#L261

And in both cases the waiting time is `wait.ForeverTestTimeout`, which is 30s, We can change that value to shorten the waiting time. Do I miss anything?

ZhiFeng1993: I lowered `wait.ForeverTestTimeout` to 20s but haven't seen obvious effect. That's only for waiting time.

ZhiFeng1993: @immutableT  I think you are looking at the wrong test- `TestKMSPluginHealthzTTL`. Or am I wrong?

immutableT: I am looking at `TestKMSPluginHealthzTTL` - this is the slow test, right?

You are correct that the test itself does not use the values of `kmsPluginHealthzPositiveTTL` and `kmsPluginHealthzNegativeTTL`, rather the logic under test is using these hard-coded values to determine when to re-check the health of the plugin. 

Essentially, this tests goes through the following phases:
1. Setup (fast)
2. Assert that OK is returned for a healthy plugin (fast)
3. Place a plugin into failed state (fast)
4. Validate that !OK is returned for a failed plugin (slow, because we kube-apiserver will not re-check the status until the PositiveTTL expires - this is what the wait.ForeverTestTimeout does).

As I mentioned above, ideally we should expose these TTL values in the config so that users and tests could adjust this.
However, there the current opinion is that we should not complicate the API  of the plugin with this level of details.
So, we could try the following approach:

1. Setup with kms-plugin in the failed state (fast)
2. Assert that !OK is returned for a failed plugin (faster than before 2s vs. 20s)
3. Place a plugin into the healthy state (fast)
4. Validate that OK is returned for a failed plugin (faster than before 2s vs. 20s).

ZhiFeng1993: Thanks @immutableT  I think we are looking at: `k8s.io/kubernetes/test/integration/master.TestKMSHealthz`, not `TestKMSPluginHealthzTTL`

ZhiFeng1993: `TestKMSPluginHealthzTTL` is the one in `config_test.go`

PR #93843, opened 2020-08-10
Fast return when no any matched anti-affinity terms

soulxu:
When check the incoming pod's anti-affinity rules, there is change to
return early when there is no any matched anti-affinity terms in the
whole cluster.

**What type of PR is this?**
/kind cleanup

If there is no any matched anti-affinity terms in the whole cluster, there is no need to check each pod affinity term again.

None


adtac:
LGTM

damemi (commented):
lgtm, but would it work to add a similar check against `len(state.topologyToMatchedAffinityTerms)` in `satisfyPodAffinity`?

Huang-Wei:
/lgtm
leaving /approve to @ahg-g. 

> lgtm, but would it work to add a similar check against `len(state.topologyToMatchedAffinityTerms)` in `satisfyPodAffinity`?

That's a bit different - in PodAffinity, you would still need to check if each "topologyPair" exists.

PR #47579, opened 2017-06-15
Clean up Deployment overlap annotation code

wanghaoran1988:
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #43322

**Special notes for your reviewer**:

**Release note**:

```
None
```

PR #34952, opened 2016-10-17
Make overlapping deployments deletable

0xmichalis:
@kubernetes/deployment ptal

Fixes https://github.com/kubernetes/kubernetes/issues/34466 by 1) not adding the overlapping annotation in the working deployment, 2) updates observedGeneration for overlapping deployments, and 3) updates the kubectl deployment reaper to do non-cascading deletion for deployments with the overlapping annotation.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kubernetes/34952)

PR #93844, opened 2020-08-10
Remove unnecessary updates in deployment controller

woohhan:
**What type of PR is this?**
/kind cleanup

Code for overlapping deployment was added in #34952, but removed from #47579 later and This part should also have been cleaned up. The empty selector always return immediately, so there is no need to update the Generation. 

```release-note
NONE
```

### DIFF
--- a/pkg/controller/deployment/deployment_controller.go
+++ b/pkg/controller/deployment/deployment_controller.go
@@ -592,10 +592,6 @@ func (dc *DeploymentController) syncDeployment(key string) error {
 	everything := metav1.LabelSelector{}
 	if reflect.DeepEqual(d.Spec.Selector, &everything) {
 		dc.eventRecorder.Eventf(d, v1.EventTypeWarning, "SelectingAll", "This deployment is selecting all pods. A non-empty selector is required.")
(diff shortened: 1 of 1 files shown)


woohhan:
Hi, if you hava time could you please review the commit? @neolit123 @mortent @kow3ns

pkg/controller/deployment/deployment_controller.go
```diff
@@ -592,10 +592,6 @@ func (dc *DeploymentController) syncDeployment(key string) error {
 	everything := metav1.LabelSelector{}
 	if reflect.DeepEqual(d.Spec.Selector, &everything) {
 		dc.eventRecorder.Eventf(d, v1.EventTypeWarning, "SelectingAll", "This deployment is selecting all pods. A non-empty selector is required.")
-		if d.Status.ObservedGeneration < d.Generation {
```
soltysh, line 595: I wouldn't call this unnecessary updates, this gives a clear information that the controller noticed the deployment and reacted on it. It will only update when a deployment will change, so that's not that excessive.

soltysh (changes requested):
/lgtm cancel

PR #93856, opened 2020-08-10
Add request info to "authentication failed" logs

stlaz:
```
Add request information when logging authentication failure. Don't
log the failure as an error but merely as a V(2).Info since it's
usually not a failure of the authenticator but of the authenticating
side.
```

**What type of PR is this?**

/kind bug

The PR adds more information about the request that failed authentication. This is useful when identifying a third party or parties that attempt to authenticate with the same credentials in a loop even though the credentials are invalid.

There does not appear to be any specific issue.

```release-note
Request information was added to logs about failed authentication attempts.
```

```docs

```

/milestone 1.20

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/endpoints/filters/authentication.go
+++ b/staging/src/k8s.io/apiserver/pkg/endpoints/filters/authentication.go
@@ -50,15 +50,14 @@ func WithAuthentication(handler http.Handler, auth authenticator.Request, failed
 		defer recordAuthMetrics(resp, ok, err, apiAuds, authenticationStart)
 		if err != nil || !ok {
 			if err != nil {
-				klog.Errorf("Unable to authenticate the request due to an error: %v", err)
+				klog.Infof("request (%s): unable to authenticate due to an error: %v", shortRequestInfo(req), err)
 			}
 			failed.ServeHTTP(w, req)
 			return
 		}
 
 		if !audiencesAreAcceptable(apiAuds, resp.Audiences) {
-			err = fmt.Errorf("unable to match the audience: %v , accepted: %v", resp.Audiences, apiAuds)
-			klog.Error(err)
+			klog.Infof("request (%s): unable to match the audience: %v , accepted: %v", shortRequestInfo(req), resp.Audiences, apiAuds)
 			failed.ServeHTTP(w, req)
 			return
 		}
@@ -91,4 +90,16 @@ func audiencesAreAcceptable(apiAuds, responseAudiences authenticator.Audiences) 
 	}
 
 	return len(apiAuds.Intersect(responseAudiences)) > 0
+}
+
+func shortRequestInfo(req *http.Request) string {
+	src := req.RemoteAddr
+	if userAgent := req.UserAgent(); len(userAgent) > 0 {
+		src += " " + userAgent
+	}
+	dst := req.Method
+	if req.URL != nil && len(req.URL.Path) > 0 {
+		dst = req.URL.Path + " " + dst
+	}
+	return fmt.Sprintf("src %s: %s", src, dst)
 }


stlaz:
pushed fixes to the unit tests

staging/src/k8s.io/apiserver/pkg/endpoints/filters/authentication.go
```diff
@@ -50,15 +49,14 @@ func WithAuthentication(handler http.Handler, auth authenticator.Request, failed
 		defer recordAuthMetrics(resp, ok, err, apiAuds, authenticationStart)
 		if err != nil || !ok {
 			if err != nil {
-				klog.Errorf("Unable to authenticate the request due to an error: %v", err)
+				klog.V(2).Infof("request (src %s: %s %s): unable to authenticate due to an error: %v", req.RemoteAddr, req.Method, req.URL.Path, err)
```
sttts, line 52: user-agent?

stlaz: could be useful for source identification as well, added

staging/src/k8s.io/apiserver/pkg/endpoints/filters/authentication.go
```diff
@@ -50,15 +50,14 @@ func WithAuthentication(handler http.Handler, auth authenticator.Request, failed
 		defer recordAuthMetrics(resp, ok, err, apiAuds, authenticationStart)
 		if err != nil || !ok {
 			if err != nil {
-				klog.Errorf("Unable to authenticate the request due to an error: %v", err)
+				klog.V(2).Infof("request (%s): unable to authenticate due to an error: %v", shortRequestInfo(req), err)
```
liggitt, line 53: the PR description didn't mention lowering an error to info-level... why is this being done?

staging/src/k8s.io/apiserver/pkg/endpoints/filters/authentication.go
```diff
@@ -50,15 +50,14 @@ func WithAuthentication(handler http.Handler, auth authenticator.Request, failed
...
 			}
 			failed.ServeHTTP(w, req)
 			return
 		}
 
 		if !audiencesAreAcceptable(apiAuds, resp.Audiences) {
-			err = fmt.Errorf("unable to match the audience: %v , accepted: %v", resp.Audiences, apiAuds)
-			klog.Error(err)
+			klog.V(2).Infof("request (%s): unable to match the audience: %v , accepted: %v", shortRequestInfo(req), resp.Audiences, apiAuds)
```
liggitt, line 60: same question about lowering the severity of what seems like an important error condition from the configured authenticator

staging/src/k8s.io/apiserver/pkg/endpoints/filters/authentication.go
```diff
@@ -92,3 +91,15 @@ func audiencesAreAcceptable(apiAuds, responseAudiences authenticator.Audiences)
 
 	return len(apiAuds.Intersect(responseAudiences)) > 0
 }
+
+func shortRequestInfo(req *http.Request) string {
```
liggitt, line 95: this info (remote address, url path, and user agent) is already included in the httplog... doing `curl -k https://localhost:6443 -H "Authorization: Bearer foo"` logs this:

```
E0811 10:50:25.993001   59995 authentication.go:53] Unable to authenticate the request due to an error: invalid bearer token
I0811 10:50:25.993299   59995 httplog.go:89] "HTTP" verb="GET" URI="/" latency="501.505µs" userAgent="curl/7.64.1" srcIP="127.0.0.1:62602" resp=401
```

the logged 401 response seems sufficient to identify the problematic request

staging/src/k8s.io/apiserver/pkg/endpoints/filters/authentication.go
stlaz: The log lever lowering was a mistake, I too think it should be kept on the same level.

staging/src/k8s.io/apiserver/pkg/endpoints/filters/authentication.go
stlaz: This only seems to be the case for loglevels >= 3 - based on
https://github.com/kubernetes/kubernetes/blob/112dbd55860e600af525cedc255f2664e3f286aa/staging/src/k8s.io/apiserver/pkg/server/httplog/httplog.go#L88-L90

That means that for log levels up until 2 you get get your logs filled with messages like:
```
E0526 11:10:32.896176       1 authentication.go:53] Unable to authenticate the request due to an error: [invalid bearer token, context canceled]
E0526 11:11:12.842792       1 authentication.go:53] Unable to authenticate the request due to an error: [invalid bearer token, context canceled]
```
that aren't really actionable.

Asking a 3rd party for logs that are more verbose is an unnecessary roundtrip when you could include this basic information within this message already.

stlaz:
@liggitt I almost forgot about this PR since the not-so-verbose logs is not the worst thing that happened to the world, but I'd appreciate if you took another look

staging/src/k8s.io/apiserver/pkg/endpoints/filters/authentication.go
liggitt: constructing ad-hoc alternate request attribute serialization in specific filters (in ways that will duplicate existing logs when running with v>=3) doesn't seem great. If we want to log more request details specifically in 401 scenarios, would it make more sense to adjust the http logging conditions to incorporate that?

PR #84568, opened 2019-10-30
test/e2e/apps: Skip or scale LB test per node count

Miciah:
Skip the "Deployment should not disrupt a cloud load-balancer's connectivity during rollout" test if the number of nodes is less than 2; otherwise, set the deployment's replicas equal to the lesser of 5 and the number of nodes.

The test would fail if there were fewer nodes than replicas, but the test needs at least 2 nodes, and the likelihood of failure absent the feature under test increases with the number of replicas, so it is desirable to set replicas to a higher value, within reason.

* `test/e2e/apps/deployment.go`: Skip the load-balancer connectivity test unless there are at least 2 nodes.
(`testRollingUpdateDeploymentWithLocalTrafficLoadBalancer`): Set `replicas` to the min of 5 and the number of nodes.

**What type of PR is this?**
/kind failing-test

**What this PR does / why we need it**:

Ensure the "Deployment should not disrupt a cloud load-balancer's connectivity during rollout" test does not fail when the number of nodes is fewer than 3.

**Which issue(s) this PR fixes**:

None.

**Special notes for your reviewer**:

Follow-up to #80004.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #93857, opened 2020-08-10
Skip LB test per schedulable node count

soltysh:
**What type of PR is this?**
/kind failing-test

This is alternative approach to the problem described in #84568 in that it requires the 3 nodes to be schedulable before continuing with LB deployment test. 

/assign @tnozicka @Miciah 

```release-note
NONE
```


hasheddan (approved):
/lgtm

@soltysh is this affecting any release blocking jobs? If so, do you think this is a good candidate for v1.19 milestone?

soltysh:
> @soltysh is this affecting any release blocking jobs? If so, do you think this is a good candidate for v1.19 milestone?

Every test that improves stability of tests definitely deserves landing. The risk is low with this one and the gain can be significant. Having said that, I'm not aware of release blocking job atm.

hasheddan (commented):
/milestone v1.19

PR #88893, opened 2020-03-06
kube-apiserver: use SO_REUSEPORT when creating listener

invidian:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

**What this PR does / why we need it**:

So multiple instances of kube-apiserver can bind on the same address and
port, to provide seamless upgrades.

**Which issue(s) this PR fixes**:

Fixes #88785

**Special notes for your reviewer**:

Perhaps some e2e tests could be added for that too, but I don't know how to do that.

**Does this PR introduce a user-facing change?**:

```release-note
kube-apiserver, kube-scheduler and kube-controller manager now use SO_REUSEPORT socket option when listening on address defined by --bind-address and --secure-port flags, when running on Unix systems (Windows is NOT supported). This allows to run multiple instances of those processes on a single host with the same configuration, which allows to update/restart them in a graceful way, without causing downtime.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #93861, opened 2020-08-10
apiserver: add --permit-address-sharing flag to listen with SO_REUSEADDR

sttts:
If true, SO_REUSEADDR will be used when binding the port. This allows binding to wildcard IPs like 0.0.0.0 and specific IPs in parallel, and it avoids waiting for the kernel to release sockets in TIME_WAIT state.

This is useful when restarting kube-apiserver: sockets in TIME_WAIT can take up to some minutes depending on kernel settings to be released, stopping the new kube-apiserver instance to launch, even though the old one has long been terminated.

/kind feature
```release-note
Add --permit-address-sharing flag to kube-apiserver to listen with SO_REUSEADDR. While allowing to listen on wildcard IPs like 0.0.0.0 and specific IPs in parallel, it avoid waiting for the kernel to release socket in TIME_WAIT state, and hence, considably reducing kube-apiserver restart times under certain conditions.
```

For reference, https://stackoverflow.com/questions/14388706/how-do-so-reuseaddr-and-so-reuseport-differ/14388707#14388707 is a great discussion about the background of SO_REUSEADDR.

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/server/options/serving.go
+++ b/staging/src/k8s.io/apiserver/pkg/server/options/serving.go
@@ -23,6 +23,7 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/spf13/pflag"
 	"k8s.io/klog/v2"
@@ -71,6 +72,9 @@ type SecureServingOptions struct {
 	// PermitPortSharing controls if SO_REUSEPORT is used when binding the port, which allows
 	// more than one instance to bind on the same address and port.
 	PermitPortSharing bool
+
+	// PermitAddressSharing controls if SO_REUSEADDR is used when binding the port.
+	PermitAddressSharing bool
 }
 
 type CertKey struct {
@@ -203,6 +207,11 @@ func (s *SecureServingOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.PermitPortSharing, "permit-port-sharing", s.PermitPortSharing,
 		"If true, SO_REUSEPORT will be used when binding the port, which allows "+
 			"more than one instance to bind on the same address and port. [default=false]")
+
+	fs.BoolVar(&s.PermitAddressSharing, "permit-address-sharing", s.PermitAddressSharing,
+		"If true, SO_REUSEADDR will be used when binding the port. This allows binding "+
+			"to wildcard IPs like 0.0.0.0 and specific IPs in parallel, and it avoids waiting "+
+			"for the kernel to release sockets in TIME_WAIT state. [default=false]")
 }
 
 // ApplyTo fills up serving information in the server configuration.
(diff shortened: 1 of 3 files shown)


bparees:
it looks fine to me but there seems to be some debate within the apiserver team about the merits of this flag so i'm going to abstain from lgtm'ing it while that is negotiated.

liggitt:
how does this differ from https://github.com/kubernetes/kubernetes/pull/88893?

bparees:
> how does this differ from #88893?

REUSEADDR has different implications than REUSEPORT does, this is adding the former, not the latter.

https://stackoverflow.com/questions/14388706/how-do-so-reuseaddr-and-so-reuseport-differ

liggitt:
ah, missed that

sttts:
@liggitt any objections to add this?

liggitt:
It seems coherent with the `SO_REUSEPORT` option. Are there any risks we need to be aware of or call out for people using this option?

I am a little confused by the help text ("...This allows binding to ... specific IPs in parallel..."). Does that mean that the existing reuse-port option is not sufficient to let two servers both bind to `1.2.3.4:8080` at the same time?

sttts:
@liggitt take a look at https://stackoverflow.com/a/14388707. The behaviour of this flag is anything but trivial. Compare my downstream (https://github.com/openshift/kubernetes/pull/309) comment:

> This allows to bind before a closed socket leave TIME_WAIT state. 
> Note: this does not allow another process to bind to the same addr/port in non TIME_WAIT state.

In other words, this cuts down the time only until the socket is released by the existing process **after** it has closed it. This avoids that the port is blocked up to minutes sometimes (depending on kernel setting afaik) although the process is long terminated.

sttts:
If you have a good way to express that in one sentence for release notes, am open for suggestions. I fear that you have to read the actual kernel docs to really understand what's going on. Nevertheless the feature is without workaround if you really need it.

liggitt:
some verify errors, lgtm otherwise

sttts:
Fixed the gofmt error.

staging/src/k8s.io/apiserver/pkg/server/options/serving.go
```diff
@@ -203,6 +207,11 @@ func (s *SecureServingOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.PermitPortSharing, "permit-port-sharing", s.PermitPortSharing,
 		"If true, SO_REUSEPORT will be used when binding the port, which allows "+
 			"more than one instance to bind on the same address and port. [default=false]")
+
+	fs.BoolVar(&s.PermitAddressSharing, "permit-address-sharing", s.PermitPortSharing,
```
aojea, line 211: should be the `s.PermitPortSharing` at the end of the line `PermitAddressSharing`?

p0lyn0mial: yes, good catch.

/hold

p0lyn0mial: fixed

/hold cancel
/lgtm

pires:
cc @jdef

PR #88568, opened 2020-02-26
Clean up duplicate code and remove import cycle.

RainbowMango:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
In #86282, we deprecated all metrics under endpoint `/metrics/resource/v1alpha1` which locate at [pkg/kubelet/apis/resourcemetrics/v1alpha1/config.go](https://github.com/kubernetes/kubernetes/blob/b6b494b4484b51df8dc6b692fab234573da30ab4/pkg/kubelet/apis/resourcemetrics/v1alpha1/config.go).

It's hard to do a unit test for these deprecated metrics because of the import cycle.
Let's say :
- the package `pkg/kubelet/server` is `A`.
- the package `pkg/kubelet/server/stats` is `B`.
- the package `pkg/kubelet/apis/resourcemetrics/v1alpha1` is `C`.

`A` has both dependencies of `B` and `C` at [here](https://github.com/kubernetes/kubernetes/blob/6e35a13815adf221bc2b65eadcdb326de2283521/pkg/kubelet/server/server.go#L353). 
`C` has dependencies of `B` because of [stats.ResourceMetricsConfig](https://github.com/kubernetes/kubernetes/blob/6e35a13815adf221bc2b65eadcdb326de2283521/pkg/kubelet/apis/resourcemetrics/v1alpha1/config.go).

Given we want to test the collector in package `B`, by a call like  [NewPrometheusResourceMetricCollector(s.resourceAnalyzer, v1alpha1.Config())](https://github.com/kubernetes/kubernetes/blob/6e35a13815adf221bc2b65eadcdb326de2283521/pkg/kubelet/server/server.go#L353), that will involve `C`. The import cycle happens...

This PR intends to eliminate the import cycle and revise the ugly unit tests. 

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

/priority backlog

PR #93868, opened 2020-08-10
update ResourceMetricsAPI tests

MHBauer:
/metrics/resource/v1alpha1 was deprecated and moved to
/metrics/resource

Renames to remove v1alpha1 from function names and matcher variables.

Pod deletion was taking multiple minutes, so set GracePeriodSeconds to 0.

Removed restart loop during test pod startup.

Move ResourceMetricsAPI out of Orphans by giving it a NodeFeature tag.

API removed in 7b7c73bf8754f2d9c7c3bfa7f3a196fdb19e74f1 #88568
Test created 6051664c0fb0eb78a84cea8bf5a1e1e4584de50e #73946

**What type of PR is this?**

/kind cleanup
/kind failing-test

API moved without updating test.

Failing tests in https://testgrid.k8s.io/sig-node-kubelet#node-kubelet-orphans

/cc @vpickard 
/assign @dashpole 
Is there an explanation for the restart loop? It looks like it may have been a copy and paste from the garbage collection test.

```release-note
NONE
```

### DIFF
--- a/test/e2e_node/resource_metrics_test.go
+++ b/test/e2e_node/resource_metrics_test.go
@@ -21,7 +21,6 @@ import (
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	kubeletresourcemetricsv1alpha1 "k8s.io/kubernetes/pkg/kubelet/apis/resourcemetrics/v1alpha1"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2ekubectl "k8s.io/kubernetes/test/e2e/framework/kubectl"
 	e2emetrics "k8s.io/kubernetes/test/e2e/framework/metrics"
@@ -41,16 +40,16 @@ const (
 	maxStatsAge = time.Minute
 )
 
-var _ = framework.KubeDescribe("ResourceMetricsAPI", func() {
+var _ = framework.KubeDescribe("ResourceMetricsAPI [NodeFeature:ResourceMetrics]", func() {
 	f := framework.NewDefaultFramework("resource-metrics")
 	ginkgo.Context("when querying /resource/metrics", func() {
 		ginkgo.BeforeEach(func() {
-			ginkgo.By("Creating test pods")
+			ginkgo.By("Creating test pods to measure their resource usage")
 			numRestarts := int32(1)
 			pods := getSummaryTestPods(f, numRestarts, pod0, pod1)
 			f.PodClient().CreateBatch(pods)
 
-			ginkgo.By("Waiting for test pods to restart the desired number of times")
+			ginkgo.By("restarting the containers to ensure container metrics are still being gathered after a container is restarted")
 			gomega.Eventually(func() error {
 				for _, pod := range pods {
 					err := verifyPodRestartCount(f, pod.Name, len(pod.Spec.Containers), numRestarts)
@@ -64,13 +63,13 @@ var _ = framework.KubeDescribe("ResourceMetricsAPI", func() {
 			ginkgo.By("Waiting 15 seconds for cAdvisor to collect 2 stats points")
 			time.Sleep(15 * time.Second)
 		})
-		ginkgo.It("should report resource usage through the v1alpha1 resouce metrics api", func() {
-			ginkgo.By("Fetching node so we can know proper node memory bounds for unconstrained cgroups")
+		ginkgo.It("should report resource usage through the resouce metrics api", func() {
+			ginkgo.By("Fetching node so we can match against an appropriate memory limit")
 			node := getLocalNode(f)
 			memoryCapacity := node.Status.Capacity["memory"]
 			memoryLimit := memoryCapacity.Value()
 
-			matchV1alpha1Expectations := gstruct.MatchAllKeys(gstruct.Keys{
+			matchResourceMetrics := gstruct.MatchAllKeys(gstruct.Keys{
 				"scrape_error": gstruct.Ignore(),
 				"node_cpu_usage_seconds_total": gstruct.MatchAllElements(nodeID, gstruct.Elements{
 					"": boundedSample(1, 1e6),
@@ -90,14 +89,15 @@ var _ = framework.KubeDescribe("ResourceMetricsAPI", func() {
 				}),
 			})
 			ginkgo.By("Giving pods a minute to start up and produce metrics")
-			gomega.Eventually(getV1alpha1ResourceMetrics, 1*time.Minute, 15*time.Second).Should(matchV1alpha1Expectations)
+			gomega.Eventually(getResourceMetrics, 1*time.Minute, 15*time.Second).Should(matchResourceMetrics)
 			ginkgo.By("Ensuring the metrics match the expectations a few more times")
-			gomega.Consistently(getV1alpha1ResourceMetrics, 1*time.Minute, 15*time.Second).Should(matchV1alpha1Expectations)
+			gomega.Consistently(getResourceMetrics, 1*time.Minute, 15*time.Second).Should(matchResourceMetrics)
 		})
 		ginkgo.AfterEach(func() {
 			ginkgo.By("Deleting test pods")
-			f.PodClient().DeleteSync(pod0, metav1.DeleteOptions{}, 10*time.Minute)
-			f.PodClient().DeleteSync(pod1, metav1.DeleteOptions{}, 10*time.Minute)
+			var zero int64 = 0
+			f.PodClient().DeleteSync(pod0, metav1.DeleteOptions{GracePeriodSeconds: &zero}, 10*time.Minute)
+			f.PodClient().DeleteSync(pod1, metav1.DeleteOptions{GracePeriodSeconds: &zero}, 10*time.Minute)
 			if !ginkgo.CurrentGinkgoTestDescription().Failed {
 				return
 			}
@@ -110,8 +110,9 @@ var _ = framework.KubeDescribe("ResourceMetricsAPI", func() {
 	})
 })
 
-func getV1alpha1ResourceMetrics() (e2emetrics.KubeletMetrics, error) {
-	return e2emetrics.GrabKubeletMetricsWithoutProxy(framework.TestContext.NodeName+":10255", "/metrics/resource/"+kubeletresourcemetricsv1alpha1.Version)
+func getResourceMetrics() (e2emetrics.KubeletMetrics, error) {
+	ginkgo.By("getting stable resource metrics API")
+	return e2emetrics.GrabKubeletMetricsWithoutProxy(framework.TestContext.NodeName+":10255", "/metrics/resource")
 }
 
 func nodeID(element interface{}) string {
--- a/test/e2e_node/BUILD
+++ b/test/e2e_node/BUILD
@@ -156,7 +156,6 @@ go_test(
         "//pkg/kubelet:go_default_library",
         "//pkg/kubelet/apis/config:go_default_library",
         "//pkg/kubelet/apis/podresources/v1alpha1:go_default_library",
-        "//pkg/kubelet/apis/resourcemetrics/v1alpha1:go_default_library",
         "//pkg/kubelet/apis/stats/v1alpha1:go_default_library",
         "//pkg/kubelet/cm:go_default_library",
         "//pkg/kubelet/cm/cpumanager:go_default_library",


test/e2e_node/resource_metrics_test.go
```diff
@@ -41,36 +40,24 @@ const (
...
-					}
-				}
-				return nil
-			}, time.Minute, 5*time.Second).Should(gomega.Succeed())
-
 			ginkgo.By("Waiting 15 seconds for cAdvisor to collect 2 stats points")
 			time.Sleep(15 * time.Second)
 		})
-		ginkgo.It("should report resource usage through the v1alpha1 resouce metrics api", func() {
+		ginkgo.It("should report resource usage through the resouce metrics api", func() {
```
hasheddan, line 54: nitpick:
```suggestion
		ginkgo.It("should report resource usage through the resource metrics api", func() {
```

SergeyKanzhelev:
Just for my understanding, I see that the alpha endpoint is still being registered:

https://github.com/kubernetes/kubernetes/blob/da5ec168e71d199721bab32caee565386ff47306/pkg/kubelet/server/server.go#L381-L387

Were tests failing because of a timeout?

As for the change, even if alpha endpoint is still registered, I don't think it's worth validating it and the change looks good:

/lgtm
/retest

SergeyKanzhelev:
looks like bazel update is needed.

test/e2e_node/resource_metrics_test.go
```diff
@@ -41,36 +40,24 @@ const (
...
-			ginkgo.By("Creating test pods")
+			ginkgo.By("Creating test pods to measure their resource usage")
 			numRestarts := int32(1)
 			pods := getSummaryTestPods(f, numRestarts, pod0, pod1)
 			f.PodClient().CreateBatch(pods)
-
-			ginkgo.By("Waiting for test pods to restart the desired number of times")
-			gomega.Eventually(func() error {
-				for _, pod := range pods {
-					err := verifyPodRestartCount(f, pod.Name, len(pod.Spec.Containers), numRestarts)
```
dashpole, line 56: This was initially added to catch a bug in which container metrics would disappear after container restarts.  It was copied from the summary API test IIRC.  I'd like to keep this if possible.

test/e2e_node/resource_metrics_test.go
```diff
@@ -64,13 +63,13 @@ var _ = framework.KubeDescribe("ResourceMetricsAPI", func() {
 			ginkgo.By("Waiting 15 seconds for cAdvisor to collect 2 stats points")
 			time.Sleep(15 * time.Second)
 		})
-		ginkgo.It("should report resource usage through the v1alpha1 resouce metrics api", func() {
+		ginkgo.It("should report resource usage through the resource metrics api", func() {
 			ginkgo.By("Fetching node so we can know proper node memory bounds for unconstrained cgroups")
 			node := getLocalNode(f)
 			memoryCapacity := node.Status.Capacity["memory"]
 			memoryLimit := memoryCapacity.Value()
```
MHBauer, line 70: @dashpole can you help me understand this comment? It looks like it refers to the following couple of lines, and then goes into the matcher. I'm guessing that this is sort of a default return value of the node having access to all of it's own memory?

MHBauer (commented):
rebased + hack/update-bazel.sh

> Just for my understanding, I see that the alpha endpoint is still being registered:
> 
> https://github.com/kubernetes/kubernetes/blob/da5ec168e71d199721bab32caee565386ff47306/pkg/kubelet/server/server.go#L381-L387
> 
> Were tests failing because of a timeout?
> 
> As for the change, even if alpha endpoint is still registered, I don't think it's worth validating it and the change looks good:

Tests were failing because there was no content in the response. I think #88568 shows that the endpoint still exists, but has nothing in it, do I also need to remove the registration in `pkg/kubelet/server/server.go` ? 

```
_output/local/go/src/k8s.io/kubernetes/test/e2e_node/resource_metrics_test.go:67
Timed out after 60.000s.
Expected
    <string>: KubeletMetrics
to match keys: {
missing expected key scrape_error
missing expected key node_cpu_usage_seconds_total
missing expected key node_memory_working_set_bytes
missing expected key container_cpu_usage_seconds_total
missing expected key container_memory_working_set_bytes
}
```

Maybe I've gone about this the wrong way and we should put back the removed api? Or remove even more code?

test/e2e_node/resource_metrics_test.go
dashpole: The matcher checks that the returned value is within a range.  Memory usage should not exceed the capacity of the node, so it is an upper bound on some ranges IIRC.

dashpole:
Oh!  I didn't realize the resource metrics test was failing

We should definitely fix it in 1.19.  The existing endpoint should still work this release.  cc @RainbowMango

test/e2e_node/resource_metrics_test.go
MHBauer: thank you.

dashpole:
Would it be possible to test both endpoints?  I'd like to try and fix the v1alpha1 endpoint

test/e2e_node/resource_metrics_test.go
```diff
@@ -64,13 +63,13 @@ var _ = framework.KubeDescribe("ResourceMetricsAPI", func() {
 			ginkgo.By("Waiting 15 seconds for cAdvisor to collect 2 stats points")
 			time.Sleep(15 * time.Second)
 		})
-		ginkgo.It("should report resource usage through the v1alpha1 resouce metrics api", func() {
```
MHBauer, line 67: Add another clause like this for v1alpha1? Same test, different endpoints?

MHBauer (commented):
@dashpole Does that mean add back the code of https://github.com/kubernetes/kubernetes/pull/88568 ? 

 Same test, different endpoints?

test/e2e_node/resource_metrics_test.go
dashpole: Yes.  #88568 was supposed to keep the contents of the endpoint the same IIRC, but just refactor and add a new endpoint.

test/e2e_node/resource_metrics_test.go
MHBauer: I put back the loop with your explanation as comment.
I'm not sure this is going to have the desired effect since we wait later for the metrics to populate. WDYT?

pkg/kubelet/server/server.go
```diff
@@ -378,20 +378,17 @@ func (s *Server) InstallDefaultHandlers(enableCAdvisorJSONEndpoints bool) {
...
 	resourceRegistry := compbasemetrics.NewKubeRegistry()
 	resourceRegistry.CustomMustRegister(collectors.NewResourceMetricsCollector(s.resourceAnalyzer))
 	s.restfulCont.Handle(resourceMetricsPath,
 		compbasemetrics.HandlerFor(resourceRegistry, compbasemetrics.HandlerOpts{ErrorHandling: compbasemetrics.ContinueOnError}),
 	)
+	// deprecated endpoint which will be removed in release 1.20.0+. Returns same as above metrics/resource API.
+	s.addMetricsBucketMatcher("metrics/resource/v1alpha1")
+	s.restfulCont.Handle(path.Join(resourceMetricsPath, v1alpha1.Version),
+		compbasemetrics.HandlerFor(resourceRegistry, compbasemetrics.HandlerOpts{ErrorHandling: compbasemetrics.ContinueOnError}),
+	)
```
MHBauer, line 391: I think this is what we want to do? Hook up the alpha endpoint to the stable endpoint.

Alpha endpoint had to move below the existing one to delete the registry and pick up the stable registry endpoint.

test/e2e_node/resource_metrics_test.go
```diff
@@ -89,15 +89,19 @@ var _ = framework.KubeDescribe("ResourceMetricsAPI", func() {
...
 			})
+			ginkgo.By("ensuring both resource metrics APIs return the same values")
 			ginkgo.By("Giving pods a minute to start up and produce metrics")
-			gomega.Eventually(getV1alpha1ResourceMetrics, 1*time.Minute, 15*time.Second).Should(matchV1alpha1Expectations)
+			gomega.Eventually(getResourceMetrics, 1*time.Minute, 15*time.Second).Should(matchResourceMetrics)
+			gomega.Eventually(getV1alpha1ResourceMetrics, 1*time.Minute, 15*time.Second).Should(matchResourceMetrics)
 			ginkgo.By("Ensuring the metrics match the expectations a few more times")
-			gomega.Consistently(getV1alpha1ResourceMetrics, 1*time.Minute, 15*time.Second).Should(matchV1alpha1Expectations)
+			gomega.Consistently(getResourceMetrics, 1*time.Minute, 15*time.Second).Should(matchResourceMetrics)
+			gomega.Consistently(getV1alpha1ResourceMetrics, 1*time.Minute, 15*time.Second).Should(matchResourceMetrics)
```
MHBauer, line 98: I don't know ginkgo well enough to know if we can do two comparisons in one call here. 

This serializes everything, so it takes 2x as long since we wait a minute for each.

MHBauer (commented):
Checking to make sure I have the right idea here.
 /cc  @RainbowMango
as the original mover of some of this code.

test/e2e_node/resource_metrics_test.go
dashpole: Right.  IIRC, the bug that we fixed, and added this test for was where we only collected metrics for the first instance of a container, and not for ones created after the container restarts.  So we want to first have a pod with containers that have already restarted, and then make sure metrics look correct after that.  Make sense?

test/e2e_node/resource_metrics_test.go
dashpole: I'm fine with this, or i'm also fine having separate `If(` blocks, and making the entire test a function that takes the getResourceMetrics() function.  The advantage of the latter is that it is easy to tell which endpoint isn't working.

test/e2e_node/resource_metrics_test.go
MHBauer: I understand now. I'll think on this and if it needs to be rewritten again. Thank you.

test/e2e_node/resource_metrics_test.go
MHBauer: I'll mess around with it some more and see if I can split it up without introducing a bunch of duplication.

MHBauer: I need to modify the matcher to for each case to ignore the other's pods. Not sure if there is a namespaced call to resource API that this should use instead.

test/e2e_node/resource_metrics_test.go
MHBauer: I think this should be 
BY "restarting containers to ensure that future instances of a container besides the first instance collect metrics"
or something like that. 

Is there an issue/PR we can link to for demonstative purposes? Or does this need a separate Test case itself?

MHBauer:
That's an interesting failure, 
```
I0813 21:07:20.026] STEP: exposing timestamps
I0813 21:07:20.027] Aug 13 21:07:00.471: INFO: Running '/go/src/k8s.io/kubernetes/bazel-bin/cmd/kubectl/linux_amd64_pure_stripped/kubectl --server=https://[::1]:46649 --kubeconfig=/root/.kube/kind-test-config logs logs-generator logs-generator --namespace=kubectl-9623 --tail=1 --timestamps'
I0813 21:07:20.027] Aug 13 21:07:01.947: INFO: stderr: ""
I0813 21:07:20.027] Aug 13 21:07:01.947: INFO: stdout: "2020-08-13T21:07:01.709817451Z I0813 21:07:01.699522       1 logs_generator.go:76] 36 GET /api/v1/namespaces/kube-system/pods/csmw 221\n2020-08-13T21:07:01.897899022Z I0813 21:07:01.890402       1 logs_generator.go:76] 37 GET /api/v1/namespaces/default/pods/2t6 559\n"
I0813 21:07:20.028] Aug 13 21:07:01.947: INFO: got output "2020-08-13T21:07:01.709817451Z I0813 21:07:01.699522       1 logs_generator.go:76] 36 GET /api/v1/namespaces/kube-system/pods/csmw 221\n2020-08-13T21:07:01.897899022Z I0813 21:07:01.890402       1 logs_generator.go:76] 37 GET /api/v1/namespaces/default/pods/2t6 559\n"
I0813 21:07:20.028] Aug 13 21:07:01.948: FAIL: Expected
I0813 21:07:20.028]     <int>: 2
I0813 21:07:20.028] to equal
I0813 21:07:20.028]     <int>: 1 
```
Don't think it's related to this change, but maybe a bug in `kubectl get logs` to investigate.

MHBauer:
looks like flake #92468

RainbowMango:
> We should definitely fix it in 1.19. The existing endpoint should still work this release. 

Yes. In 1.19 both `metrics/resource/v1alpha1` and `metrics/resource` are available, but as per metrics deprecation policy, no metrics will be emitted from `metrics/resource/v1alpha1` by default, because all metrics in this endpoint will become to `hidden`.

MHBauer:
> > We should definitely fix it in 1.19. The existing endpoint should still work this release.
> 
> Yes. In 1.19 both `metrics/resource/v1alpha1` and `metrics/resource` are available, but as per metrics deprecation policy, no metrics will be emitted from `metrics/resource/v1alpha1` by default, because all metrics in this endpoint will become to `hidden`.

What does that mean here? Is there a way to unhide them?

test/e2e_node/resource_metrics_test.go
```diff
@@ -84,20 +84,30 @@ var _ = framework.KubeDescribe("ResourceMetricsAPI", func() {
...
+			gomega.Consistently(fetcher, 1*time.Minute, 15*time.Second).Should(matchResourceMetrics)
+		}
+
+		ginkgo.It("should report resource usage through the alpha resource metrics api", func() {
+			resourceFetcher := getV1alpha1ResourceMetrics
+			fetchAndTest(f, resourceFetcher)
+		})
+		ginkgo.It("should report resource usage through the stable resource metrics api", func() {
+			resourceFetcher := getResourceMetrics
+			fetchAndTest(f, resourceFetcher)
```
RainbowMango, line 104: We can't say the `/metrics/resource/v1alpha1` is the alpha resource metrics  API and `/metrics/resource` is not the stable resource metrics API, this will make confusion. They are both the resource metrics API in different endpoints.

test/e2e_node/resource_metrics_test.go
```diff
@@ -111,9 +121,15 @@ var _ = framework.KubeDescribe("ResourceMetricsAPI", func() {
 })
 
 func getV1alpha1ResourceMetrics() (e2emetrics.KubeletMetrics, error) {
```
RainbowMango, line 123: Just rename function name to `getResourceMetrics` will be fine.

RainbowMango (commented):
Given all metrics under `/metrics/resources/v1alpha1` now become to `hidden` state by default and will be removed soon. So I think we can migrate the E2E test to `/metrics/resources`.

RainbowMango:
> What this PR does / why we need it:
API moved without updating test.

@MHBauer The API didn't move, what we did is added a new endpoint(/metrics/resources) and deprecated all metrics under the old endpoint(/metrcis/resources/v1alpha1). 
I didn't realize we still have E2E test which cares about the resources metrics, sorry for that.

To sum up, what we need to do is update E2E test to validate the metrics from the new endpoint.

pkg/kubelet/server/server.go
RainbowMango: > I think this is what we want to do?

No, what we want to do is DO NOT touch the original endpoint `metrics/resource/v1alpha1` and add a new endpoint to replace the original one.

test/e2e_node/resource_metrics_test.go
RainbowMango: > Yes. #88568 was supposed to keep the contents of the endpoint the same IIRC, but just refactor and add a new endpoint.

Yes.

RainbowMango:
> Would it be possible to test both endpoints? I'd like to try and fix the v1alpha1 endpoint

I doubt it worth doing so. We are going to remove the v1alpha1 endpoint soon. So, I suggest stopping test the v1alpha1 endpoint.

SergeyKanzhelev:
> > Would it be possible to test both endpoints? I'd like to try and fix the v1alpha1 endpoint
> 
> I doubt it worth doing so. We are going to remove the v1alpha1 endpoint soon. So, I suggest stopping test the v1alpha1 endpoint.

Should it be release noted in 1.19 then that endpoint was deprecated?

RainbowMango:
> Should it be release noted in 1.19 then that endpoint was deprecated?

@SergeyKanzhelev 
We have noted the deprecation in v1.18 release note by #86282, I guess we don't need to note again.
```
kubelet: resource metrics endpoint `/metrics/resource/v1alpha1` as well as all metrics under this endpoint have been deprecated.
Please convert to the following metrics emitted by endpoint `/metrics/resource`:
```

dashpole:
Ah, thanks @RainbowMango.  I forgot metrics auto-hide a release after being deprecated...

I agree we don't need to test the v1alpha1 endpoint anymore.  We can also limit this PR to just updating the endpoint for the test.

test/e2e_node/resource_metrics_test.go
MHBauer: yea, that's an earlier version

MHBauer:
Okay, summarizing
1. don't enable the alpha api
2. fix the test for the regular api only.

SergeyKanzhelev:
Should then v1alpha1  endpoint registration be removed as well? Maybe a separate PR?

MHBauer:
Should I also remove the v1alpha1 api registration?

dashpole:
> Should then v1alpha1 endpoint registration be removed as well? Maybe a separate PR?

> Should I also remove the v1alpha1 api registration?

After 1.19 is cut.

dashpole:
We still need the endpoint, as metrics can be unhidden by flag.

MHBauer:
What's the flag to unhide? I am curious, more than anything

MHBauer:
alright, this is hopefully the final version 🤞

MHBauer:
I don't know what to do with this:
```
failed to try resolving symlinks in path "/var/log/pods/test-pods_09debb56-de5c-11ea-bf0d-0a7878750357_0b856363-de5c-11ea-9893-42010a800098/test/0.log": lstat /var/log/pods/test-pods_09debb56-de5c-11ea-bf0d-0a7878750357_0b856363-de5c-11ea-9893-42010a800098/test/0.log: no such file or directory 
```
/test pull-kubernetes-e2e-gce-100-performance

dashpole:
the flag is (e.g.) `--show-hidden-metrics-for-version=1.19` to re-enable these disabled metrics.

/lgtm
/approve

MHBauer:
looks like maybe a rare flake? 18 in the past week https://storage.googleapis.com/k8s-gubernator/triage/index.html?test=Services%20should%20be%20able%20to%20change%20the%20type%20from%20NodePort%20to%20ExternalName#2fa91b3a0e6e1a4514d4

MHBauer:
alrighty, looking forward to thaw.

test/e2e_node/resource_metrics_test.go
```diff
@@ -110,8 +110,9 @@ var _ = framework.KubeDescribe("ResourceMetricsAPI", func() {
 	})
 })
 
-func getV1alpha1ResourceMetrics() (e2emetrics.KubeletMetrics, error) {
-	return e2emetrics.GrabKubeletMetricsWithoutProxy(framework.TestContext.NodeName+":10255", "/metrics/resource/"+kubeletresourcemetricsv1alpha1.Version)
+func getResourceMetrics() (e2emetrics.KubeletMetrics, error) {
+	ginkgo.By("getting stable resource metrics API")
```
RainbowMango, line 114: nit: remove `stable`, or refactor the log with something like `grabbing kubelet resource metrics`.

RainbowMango:
@MHBauer 
v1.19 has been released and #94272 will clean up the deprecated `/metrics/resource/v1alpha1` endpoint.

MHBauer:
Looks like the matcher needs to be tweaked and/or the test marked as disruptived as it can be failing due to the local environment.

PR #93873, opened 2020-08-11
Apiserver updates storageversions API and filters certain write requests during bootstrap

roycaihw:
[KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190802-dynamic-coordinated-storage-version.md). Split off from #84016.

```release-note
Add a StorageVersionAPI feature gate that makes API server update storageversions before serving certain write requests. 
This feature allows the storage migrator to manage storage migration for built-in resources. 
Enabling internal.apiserver.k8s.io/v1alpha1 API and APIServerIdentity feature gate are required to use this feature. 
```

/sig api-machinery
/priority important-soon
/cc @caesarxuchao


staging/src/k8s.io/apiserver/pkg/endpoints/installer.go
```diff
@@ -506,6 +511,37 @@ func (a *APIInstaller) registerResourceHandlers(path string, storage rest.Storag
 		}
 	}
 
+	var resourceInfo *storageversion.ResourceInfo
+	if utilfeature.DefaultFeatureGate.Enabled(features.StorageVersionAPI) &&
+		isStorageVersionProvider &&
+		storageVersionProvider.StorageVersion() != nil &&
+		// we don't need to migrate storageversions objects, because everytime
+		// the server is upgraded, these object will be rewritten by the apiserver,
```
caesarxuchao, line 519: nit: s/object/objects

staging/src/k8s.io/apiserver/pkg/endpoints/installer.go
```diff
@@ -506,6 +511,37 @@ func (a *APIInstaller) registerResourceHandlers(path string, storage rest.Storag
...
+				Resource: apiResource.Name,
+			},
+			EncodingVersion: encodingGVK.GroupVersion().String(),
+			// We record EquivalentResourceMapper first instead of calculate
+			// DecodableVersions immediately because API installation must
+			// be completed first for us to know equivalent APIs
+			EquivalentResourceMapper: a.group.EquivalentResourceRegistry,
+		}
+		if len(a.group.GroupVersion.Group) == 0 {
+			// Group is a valid subdomain in DNS (RFC 1123).
```
caesarxuchao, line 540: Group must be ... ?

staging/src/k8s.io/apiserver/pkg/server/genericapiserver.go
```diff
@@ -420,14 +429,40 @@ func (s *GenericAPIServer) installAPIResources(apiPrefix string, apiGroupInfo *A
...
+		if err != nil {
 			return fmt.Errorf("unable to setup API %v: %v", apiGroupInfo, err)
 		}
+		resourceInfos = append(resourceInfos, r...)
+	}
+
+	if utilfeature.DefaultFeatureGate.Enabled(genericfeatures.StorageVersionAPI) {
+		// API installation happens before we start listening on the handlers,
+		// therefore it is safe to register ResourceInfos here. The handler will block
+		// write requests until the targeting resources' storage version is up-to-date.
```
caesarxuchao, line 442: nit: the storage versions of the targeting resources are updated.

staging/src/k8s.io/apiserver/pkg/server/genericapiserver.go
```diff
@@ -420,14 +429,40 @@ func (s *GenericAPIServer) installAPIResources(apiPrefix string, apiGroupInfo *A
...
 	}
 
 	return nil
 }
 
+// dedupResourceInfos dedups ResourceInfos with the same underlying storage.
+// ResourceInfos from different Groups may share the same underlying storage, e.g.
+// networking.k8s.io ingresses and extensions ingresses. The StorageVersion manager
+// only needs to update one StorageVersion for the equivalent Groups.
+func dedupResourceInfos(infos []*storageversion.ResourceInfo) []*storageversion.ResourceInfo {
```
caesarxuchao, line 453: Do we still need this function? I remember we used the EquivalentResourceMapper to dedup.

 This function is problematic. If two resources are stored at different path in the storage, but share the same encoding version, they should be deduplicated.

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
```diff
@@ -35,6 +35,11 @@ import (
 type ResourceInfo struct {
 	GroupResource schema.GroupResource
 
+	// Optional. Overwrites the group name when creating the StorageVersion object.
+	// Used by the core group to make the group name more readable and a valid DNS
+	// subdomain.
+	OverwriteGroupName string
```
caesarxuchao, line 41: Can we just write the ResourceInfo.GroupResource.Group when constructing the resourceInfo?

test/integration/storageversion/storage_version_filter_test.go
```diff
@@ -0,0 +1,235 @@
...
+	_, err := client.CoreV1().Namespaces().List(context.TODO(), metav1.ListOptions{})
+	assertBlocking("reads of built-in resources", t, err, shouldBlock)
+}
+
+// TestStorageVersionBootstrap ensures that before the StorageVersions are
+// updated, only the following the request are accepted by the apiserver:
+// 1. read requests
+// 2. non-persisting write requests
+// 3. write requests to the storageversion API
+// 4. requests to CR or aggregated API
```
caesarxuchao, line 147: Do we have a test for the aggregated API? If not, let's make it a todo.

test/integration/storageversion/storage_version_filter_test.go
```diff
@@ -0,0 +1,235 @@
...
+	// Write to non-persisted API should pass.
+	testNonPersistedWrite(t, cfg, false)
+	// Read of k8s native API should pass.
+	testBuiltinResourceRead(t, cfg, false)
+
+	// After the storage versions are updated, even though the
+	// StorageVersionManager.Complete() still returns false, the filter
+	// should not block any request.
+	close(startUpdateSV)
+	<-delegateUpdateFinished
```
caesarxuchao, line 204: Can we rename this variable to `updateFinished`? I don't get what "delegate" means here, though probably I came up with this name :)

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
roycaihw: If we change `ResourceInfo.GroupResource.Group`, the `EquivalentResourceMapper` won't recognize the new group name; also the name in `PendingUpdate` will be different, meaning we need to add special handling in both places. 

Essentially we just want to change the name of the StorageVersion object in etcd, the rest places remain unchanged:
- etcd: core.pods -- this makes the name validation happy, and the name more human readable (compared to `.pods`) for debugging purpose
- what `EquivalentResourceMapper` understands: GroupResource{Group: "", Resource: "pods"}
- `managedStatus` in `PendingUpdate`: GroupResource{Group: "", Resource: "pods"}
- `RequestInfo` we see in the filter: GroupResource{Group: "", Resource: "pods"}, which matches `PendingUpdate`

Alternatively we can remove the name validation in apiserver if we think the current implementation is not clean enough.

staging/src/k8s.io/apiserver/pkg/server/genericapiserver.go
roycaihw: Switched to use the EquivalentResourceMapper to dedup.

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
roycaihw: > Alternatively we can remove the name validation in apiserver

Talked offline, in the [API reference doc](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#endpoints-v1-core) the group name is "core".

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
```diff
@@ -146,6 +149,28 @@ func (s *defaultManager) UpdateStorageVersions(kubeAPIServerClientConfig *rest.C
...
 
+// dedupResourceInfos dedups ResourceInfos with the same underlying storage.
+// ResourceInfos from the same Group with different Versions share the same underlying storage.
+// ResourceInfos from different Groups may share the same underlying storage, e.g.
+// networking.k8s.io ingresses and extensions ingresses. The StorageVersion manager
+// only needs to update one StorageVersion for the equivalent Groups.
+func dedupResourceInfos(infos []*ResourceInfo) []*ResourceInfo {
+	var ret []*ResourceInfo
+	seen := make(map[schema.GroupResource]struct{})
+	for _, info := range infos {
```
caesarxuchao, line 160: We need to sort the infos to make sure the dedup always the pick the same GR to represent its equvalent resources, e.g., always pick network.k8s.io/ingress over extensions/ingress.

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
```diff
@@ -136,6 +149,28 @@ func (s *defaultManager) UpdateStorageVersions(kubeAPIServerClientConfig *rest.C
...
```
caesarxuchao, line 160: We need to sort the infos to make sure the dedup always the pick the same GR to represent its equvalent resources, e.g., always pick network.k8s.io/ingress over extensions/ingress.

caesarxuchao (commented):
Otherwise LGTM. @deads2k can you also review this PR? Thank you.

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
roycaihw: Good catch. I squashed the previous commits and added the sort.

cmd/kube-apiserver/app/testing/testserver.go
```diff
@@ -197,6 +210,16 @@ func StartTestServer(t Logger, instanceOptions *TestServerInstanceOptions, custo
 		}
 	}(stopCh)
 
+	if instanceOptions.SkipPostStartChecks {
```
deads2k, line 213: The calls below look to be related to reads, not writes.  What causes us to need to skip the poststart checks?

staging/src/k8s.io/apiserver/pkg/endpoints/filters/storageversion.go
```diff
@@ -0,0 +1,79 @@
...
+
+// WithStorageVersionPrecondition checks if the storage version barrier has
+// completed, if not, it only passes the following API requests:
+// 1. non-resource requests,
+// 2. read requests,
+// 3. write requests to the storageversion API,
+// 4. resources whose StorageVersion is not pending update, including non-persisted resources.
+func WithStorageVersionPrecondition(handler http.Handler, svm storageversion.Manager) http.Handler {
+	if svm == nil {
+		klog.Warningf("Storage Version barrier is disabled")
```
deads2k, line 40: This isn't a warning yet.  We're starting in alpha, right?

staging/src/k8s.io/apiserver/pkg/endpoints/filters/storageversion.go
```diff
@@ -0,0 +1,79 @@
...
+			handler.ServeHTTP(w, req)
+			return
+		}
+		// Allow read requests
+		if requestInfo.Verb == "get" || requestInfo.Verb == "list" || requestInfo.Verb == "watch" {
+			handler.ServeHTTP(w, req)
+			return
+		}
+		// Allow writes to the storage version API
+		if requestInfo.Resource == "storageversions" {
```
deads2k, line 65: needs group qualification as well.

staging/src/k8s.io/apiserver/pkg/endpoints/groupversion.go
```diff
@@ -94,19 +95,32 @@ type APIGroupVersion struct {
...
-	apiResources, ws, registrationErrors := installer.Install()
+	apiResources, resourceInfos, ws, registrationErrors := installer.Install()
 	versionDiscoveryHandler := discovery.NewAPIVersionHandler(g.Serializer, g.GroupVersion, staticLister{apiResources})
 	versionDiscoveryHandler.AddToWebService(ws)
 	container.Add(ws)
-	return utilerrors.NewAggregate(registrationErrors)
+	return filterResourceInfos(resourceInfos), utilerrors.NewAggregate(registrationErrors)
+}
+
+func filterResourceInfos(infos []*storageversion.ResourceInfo) []*storageversion.ResourceInfo {
```
deads2k, line 113: nit: I prefer a name that indicates what the filter does.  maybe `removeNonPersistedResources`?

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
```diff
@@ -35,6 +36,11 @@ import (
 type ResourceInfo struct {
 	GroupResource schema.GroupResource
 
+	// Optional. Overwrites the group name when creating the StorageVersion object.
+	// Used by the core group to make the group name more readable and a valid DNS
+	// subdomain.
+	OverwriteGroupName string
```
deads2k, line 42: if it's only used by the core group, can we hardcode the logic instead?

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
```diff
@@ -112,22 +124,28 @@ func (s *defaultManager) UpdateStorageVersions(kubeAPIServerClientConfig *rest.C
...
-	resources := []*ResourceInfo{}
+	resources := []ResourceInfo{}
 	for resource := range s.managedResourceInfos {
-		resources = append(resources, resource)
+		resources = append(resources, *resource)
 	}
 	s.mu.RUnlock()
 	hasFailure := false
-	for _, r := range resources {
+	sortResourceInfosByGroupResource(resources)
```
deads2k, line 133: why do we sort?

staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
```diff
@@ -264,6 +267,21 @@ func (c completedConfig) NewWithDelegate(delegationTarget genericapiserver.Deleg
...
 
+	if utilfeature.DefaultFeatureGate.Enabled(genericfeatures.StorageVersionAPI) {
+		// Spawn a goroutine in aggregator apiserver to update storage version for
+		// all built-in resources
+		s.GenericAPIServer.AddPostStartHookOrDie("built-in-resources-storage-version-updater", func(context genericapiserver.PostStartHookContext) error {
+			// All apiservers (aggregator-apiserver, kube-apiserver, apiextensions-apiserver)
+			// share the same generic apiserver config. The same StorageVersion manager is used
+			// to register all built-in resources when the generic apiservers install APIs.
+			s.GenericAPIServer.StorageVersion.UpdateStorageVersions(context.LoopbackClientConfig, s.GenericAPIServer.APIServerID)
+			wait.PollImmediateInfinite(10*time.Minute, func() (bool, error) {
```
deads2k, line 278: I like this to help prevent future restarts from producing "weird" effects.

cmd/kube-apiserver/app/testing/testserver.go
roycaihw: Since we wrapped the storage version poststart hook for testing, the /healthz won't become ok. This allows us to skip waiting for /healthz to return ok when bootstrapping the test server. Also the default namespace creation check gets skipped since it happens after the healthz check.

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
roycaihw: Do you mean having a boolean instead of the string?

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
roycaihw: To make sure we have a consistent dedup result, and therefore avoid creating unnecessarily duplicated StorageVersion objects. 

For example, `extensions.ingresses` and `networking.k8s.io.ingresses` share the same underlying storage. Without sorting, in an HA cluster, one apiserver may dedup and update StorageVersion for `extensions.ingresses`, while another apiserver may dedup and update StorageVersion for `networking.k8s.io.ingresses`. The storage migrator (which migrates objects per GroupResource) will migrate these resources twice, since both StorageVersion objects have `CommonEncodingVersion` (each with one server registered).

staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
roycaihw: Could you elaborate more?

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
roycaihw: We can also hardcode the logic before we create the StorageVersion object. A downside of that is if an aggregated server accidentally register a resource info for an aggregated API, but omitted the group. The manager will create a StorageVersion with `group=core` for it

cmd/kube-apiserver/app/testing/testserver.go
roycaihw: Made the if block capture the entire healthz check & namespace check section.

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
roycaihw: If an aggregated server tries to register a `group=""` resource. The storage migrator won't be able to migrate the resource anyway. 

I updated the code to hardcode the logic.

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
roycaihw: Added the explanation to comments.

staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
roycaihw: Discussed offline. Updated the poststart hook to reconcile StorageVersion objects every 10 minutes.

cmd/kube-apiserver/app/testing/testserver.go
```diff
@@ -49,9 +50,12 @@ type TearDownFunc func()
 type TestServerInstanceOptions struct {
 	// DisableStorageCleanup Disable the automatic storage cleanup
 	DisableStorageCleanup bool
-
 	// Enable cert-auth for the kube-apiserver
 	EnableCertAuth bool
+	// Skip healthz check and default namespace creation check
+	SkipPostStartChecks bool
```
deads2k, line 56: Instead of setting this, can we instead infer this from a non-nil storageversionwrapfunc?  I'd really like to avoid anyone thinking this is a generally safe, non-racy thing to set and no one will want to set a storageversionwrapfunc.

cmd/kube-apiserver/app/testing/testserver.go
```diff
@@ -96,7 +101,12 @@ func StartTestServer(t Logger, instanceOptions *TestServerInstanceOptions, custo
 	}
 
 	stopCh := make(chan struct{})
+	// This gives the post start hooks some time to run even if the actual
+	// tests run too fast. Otherwise post start hooks like
+	// crd-informer-synced may panic due to timeout.
+	minRunTime := time.After(10 * time.Second)
```
deads2k, line 107: I think we should only do this if the `StorageVersionWrapFunc` is set.  Otherwise this shouldn't be needed, right?

staging/src/k8s.io/apiserver/pkg/endpoints/filters/storageversion.go
```diff
@@ -0,0 +1,80 @@
...
+			handler.ServeHTTP(w, req)
+			return
+		}
+		// Allow read requests
+		if requestInfo.Verb == "get" || requestInfo.Verb == "list" || requestInfo.Verb == "watch" {
+			handler.ServeHTTP(w, req)
+			return
+		}
+		// Allow writes to the storage version API
+		if requestInfo.APIGroup == "apiserverinternal" && requestInfo.Resource == "storageversions" {
```
deads2k, line 66: should be `internal.apiserver.k8s.io`, right?  Also, this should have failed an integration test, right?

staging/src/k8s.io/apiserver/pkg/endpoints/installer.go
```diff
@@ -506,6 +511,33 @@ func (a *APIInstaller) registerResourceHandlers(path string, storage rest.Storag
...
 	}
 
+	var resourceInfo *storageversion.ResourceInfo
+	if utilfeature.DefaultFeatureGate.Enabled(features.StorageVersionAPI) &&
+		isStorageVersionProvider &&
+		storageVersionProvider.StorageVersion() != nil &&
+		// we don't need to migrate storageversions objects, because everytime
+		// the server is upgraded, these objects will be rewritten by the apiserver,
+		// they won't be stale
+		resource != "storageversions" {
```
deads2k, line 521: I didn't expect an override here.  With the right skip in the REST handler, we should be able to return this resourceinfo without an issue.

staging/src/k8s.io/apiserver/pkg/endpoints/filters/storageversion.go
roycaihw: good point. Looking

staging/src/k8s.io/apiserver/pkg/server/genericapiserver.go
```diff
@@ -200,6 +201,11 @@ type GenericAPIServer struct {
 	// The limit on the request body size that would be accepted and decoded in a write request.
 	// 0 means no limit.
 	maxRequestBodyBytes int64
+
+	// APIServerID is the ID of this API server
+	APIServerID string
+	// StorageVersion holds the storage versions of the API resources installed by this server.
+	StorageVersion storageversion.Manager
```
deads2k, line 208: nit: StorageVersionManager if you please.

staging/src/k8s.io/apiserver/pkg/server/config.go
```diff
@@ -223,6 +227,11 @@ type Config struct {
 	// EquivalentResourceRegistry provides information about resources equivalent to a given resource,
 	// and the kind associated with a given resource. As resources are installed, they are registered here.
 	EquivalentResourceRegistry runtime.EquivalentResourceRegistry
+
+	// APIServerID is the ID of this API server
+	APIServerID string
+	// StorageVersion holds the storage versions of the API resources installed by this server.
+	StorageVersion storageversion.Manager
```
deads2k, line 234: nit: StorageVersionManager if you would

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
```diff
@@ -90,6 +91,9 @@ func (s *defaultManager) AddResourceInfo(resources ...*ResourceInfo) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	for _, r := range resources {
+		if r == nil {
```
deads2k, line 94: this seems like a bug.  how did we get a nil?

staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
```diff
@@ -264,6 +267,33 @@ func (c completedConfig) NewWithDelegate(delegationTarget genericapiserver.Deleg
...
+				// All apiservers (aggregator-apiserver, kube-apiserver, apiextensions-apiserver)
+				// share the same generic apiserver config. The same StorageVersion manager is used
+				// to register all built-in resources when the generic apiservers install APIs.
+				s.GenericAPIServer.StorageVersion.UpdateStorageVersions(context.LoopbackClientConfig, s.GenericAPIServer.APIServerID)
+				return false, nil
+			}, context.StopCh)
+			// Once the storage version updater finishes the first round of update,
+			// the PostStartHook will return to unblock /healthz. The handler chain
+			// won't block write requests anymore.
+			wait.PollImmediateInfinite(30*time.Second, func() (bool, error) {
```
deads2k, line 290: use `PollImmediateUntil` and pass the stopCh so we close cleanly.

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
roycaihw: We didn't. Had the check just to be safe. I can remove it

staging/src/k8s.io/apiserver/pkg/endpoints/installer.go
roycaihw: Returning this resourceinfo won't affect the filter. But it will trigger the storage migrator to migrate storageversion objects, which is unnecessary.

staging/src/k8s.io/apiserver/pkg/endpoints/filters/storageversion.go
roycaihw: yeah, the test passed because we don't collect resource info for storageversions, so it is not in the to-be-updated list below anyway.

staging/src/k8s.io/apiserver/pkg/endpoints/installer.go
roycaihw: We need special handling for storageversions in one place or the other: 
1. skip collecting resourceinfo in API installer
2. skip blocking in the REST handler

1) saves the storage migrator from doing unnecessary migration (assuming kube-apiserver always writes the latest version of storageversion object after upgrade), but that may not hold if aggregated servers write storageversions. 

Let's go with 2). That makes the resource info collection more unified, and the REST handler control more centralized.

cmd/kube-apiserver/app/testing/testserver.go
roycaihw: actually I'm not sure why we need this. The description isn't related to the poststart hook being tested here. If a test calling `teardown` right after it sees `/healthz` returning `ok` would cause some other poststart hook to panic/flake, it's better to fix those hooks. Or we may just make the specific tests with such behavior to wait, instead of adding latency to all test. 

Removed this to see what happens.

staging/src/k8s.io/apiserver/pkg/endpoints/filters/storageversion.go
roycaihw: fixed

roycaihw:
@deads2k I addressed the comments. Thanks for the review!

cmd/kube-apiserver/app/testing/testserver.go
roycaihw: Talked with @caesarxuchao offline. Likely this is not needed anymore (fixed somewhere else)

deads2k:
squash to three or four-ish commits and I'll have a final pass tomorrow.

staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
```diff
@@ -264,6 +267,33 @@ func (c completedConfig) NewWithDelegate(delegationTarget genericapiserver.Deleg
...
+				// All apiservers (aggregator-apiserver, kube-apiserver, apiextensions-apiserver)
+				// share the same generic apiserver config. The same StorageVersion manager is used
+				// to register all built-in resources when the generic apiservers install APIs.
+				s.GenericAPIServer.StorageVersionManager.UpdateStorageVersions(context.LoopbackClientConfig, s.GenericAPIServer.APIServerID)
+				return false, nil
+			}, context.StopCh)
+			// Once the storage version updater finishes the first round of update,
+			// the PostStartHook will return to unblock /healthz. The handler chain
+			// won't block write requests anymore.
+			wait.PollImmediateUntil(30*time.Second, func() (bool, error) {
```
deads2k, line 290: nit: I think you can tighten this up. the check isn't that expensive.  I think a followup to 1-5s ish will help us in 1.21

deads2k:
I recommend a followup with the nit, but this is a solid start

/lgtm
/approve

staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
roycaihw: will do it in a followup

roycaihw:
/assign @lavalamp 
for vendor/ approval

pkg/registry/rbac/rest/storage_rbac.go
```diff
@@ -206,7 +215,8 @@ func (p *PolicyData) EnsureRBACPolicy() genericapiserver.PostStartHookFunc {
 					Client:  reconciliation.ClusterRoleModifier{Client: clientset.ClusterRoles()},
 					Confirm: true,
 				}
-				err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+				// ServiceUnavailble error is returned when the API server is blocked by storage version updates
+				err := retryOnConflictOrServiceUnavailable(retry.DefaultBackoff, func() error {
```
lavalamp, line 219: An alternative to this is to permit self-requests even when blocking other requests, right? Did we think about that?

lavalamp: Another option is to do a GET first. This might be useful anyway, as 99.99% of the time this stuff will already be correct.

staging/src/k8s.io/apiserver/pkg/endpoints/filters/storageversion.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		// If the resource's StorageVersion is not in the to-be-updated list, let it pass.
+		// Non-persisted resources are not in the to-be-updated list, so they will pass.
+		gr := schema.GroupResource{requestInfo.APIGroup, requestInfo.Resource}
+		if !svm.PendingUpdate(gr) {
+			handler.ServeHTTP(w, req)
+			return
+		}
+
+		responsewriters.ServiceUnavailabeError(w, req, errors.New(fmt.Sprintf("wait for storage version registration to complete for resource: %v, last seen error: %v", gr, svm.LastUpdateError(gr))))
```
lavalamp, line 78: If this is a function you added in this PR, s/Unavailabe/Unavailable/g :)

staging/src/k8s.io/apiserver/pkg/endpoints/filters/storageversion.go
```diff
@@ -0,0 +1,80 @@
...
+	"k8s.io/klog/v2"
+)
+
+// WithStorageVersionPrecondition checks if the storage version barrier has
+// completed, if not, it only passes the following API requests:
+// 1. non-resource requests,
+// 2. read requests,
+// 3. write requests to the storageversion API,
+// 4. resources whose StorageVersion is not pending update, including non-persisted resources.
+func WithStorageVersionPrecondition(handler http.Handler, svm storageversion.Manager) http.Handler {
```
lavalamp, line 38: Does this approach work for extension and aggregated apis?

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/errors.go
```diff
@@ -76,3 +76,10 @@ func InternalError(w http.ResponseWriter, req *http.Request, err error) {
 		http.StatusInternalServerError)
 	utilruntime.HandleError(err)
 }
+
+// ServiceUnavailabeError renders a simple internal error
+func ServiceUnavailabeError(w http.ResponseWriter, req *http.Request, err error) {
+	http.Error(w, sanitizer.Replace(fmt.Sprintf("Service Unavailable: %q: %v", req.RequestURI, err)),
+		http.StatusServiceUnavailable)
+	utilruntime.HandleError(err)
```
lavalamp, line 84: Why would we log this?

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/errors.go
```diff
@@ -76,3 +76,10 @@ func InternalError(w http.ResponseWriter, req *http.Request, err error) {
 		http.StatusInternalServerError)
 	utilruntime.HandleError(err)
 }
+
+// ServiceUnavailabeError renders a simple internal error
+func ServiceUnavailabeError(w http.ResponseWriter, req *http.Request, err error) {
+	http.Error(w, sanitizer.Replace(fmt.Sprintf("Service Unavailable: %q: %v", req.RequestURI, err)),
```
lavalamp, line 82: Why an error page instead of a Status formatted JSON response?

Also did we debate the status code? e.g., what about 429 try again later? Can we at least add text that explains the error beyond "Service Unavailable"?

lavalamp: I thought we had some error translation code somewhere already, maybe that's only for etcd sourced errors though?

staging/src/k8s.io/apiserver/pkg/endpoints/installer.go
```diff
@@ -182,7 +187,7 @@ func GetResourceKind(groupVersion schema.GroupVersion, storage rest.Storage, typ
 	return fqKindToRegister, nil
 }
 
-func (a *APIInstaller) registerResourceHandlers(path string, storage rest.Storage, ws *restful.WebService) (*metav1.APIResource, error) {
+func (a *APIInstaller) registerResourceHandlers(path string, storage rest.Storage, ws *restful.WebService) (*metav1.APIResource, *storageversion.ResourceInfo, error) {
```
lavalamp, line 190: I'm not really a fan of adding a return parameter, but I'm not sure what would be better. I guess it'd be weird to store things in the APIInstaller for reading later?

staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
```diff
@@ -264,6 +267,33 @@ func (c completedConfig) NewWithDelegate(delegationTarget genericapiserver.Deleg
 		return nil
 	})
 
+	if utilfeature.DefaultFeatureGate.Enabled(genericfeatures.StorageVersionAPI) {
```
lavalamp, line 270: What if the identity gate is not enabled?

test/integration/storageversion/storage_version_filter_test.go
```diff
@@ -0,0 +1,235 @@
...
+			// TODO: drop these once we stop allowing them to be served.
+			"--runtime-config=api/all=true,extensions/v1beta1/deployments=true,extensions/v1beta1/daemonsets=true,extensions/v1beta1/replicasets=true,extensions/v1beta1/podsecuritypolicies=true,extensions/v1beta1/networkpolicies=true,internal.apiserver.k8s.io/v1alpha1=true",
+		},
+		etcdConfig)
+	defer server.TearDownFn()
+
+	cfg := rest.CopyConfig(server.ClientConfig)
+
+	// Write to k8s native API object should be blocked.
+	testBuiltinResourceWrite(t, cfg, true)
```
lavalamp, line 185: Consider using t.Run() for these sub-tests?

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
```diff
@@ -112,22 +116,37 @@ func (s *defaultManager) UpdateStorageVersions(kubeAPIServerClientConfig *rest.C
...
+		resources = append(resources, *resource)
 	}
 	s.mu.RUnlock()
 	hasFailure := false
-	for _, r := range resources {
+	// Sorting the list to make sure we have a consistent dedup result, and
+	// therefore avoid creating unnecessarily duplicated StorageVersion objects.
+	// For example, extensions.ingresses and networking.k8s.io.ingresses share
+	// the same underlying storage. Without sorting, in an HA cluster, one
+	// apiserver may dedup and update StorageVersion for extensions.ingresses,
```
lavalamp, line 129: What happens if apiservers disagree about which groups are in what storage? e.g., I restart an apiserver with new flags that moves events?

(I actually think lots of things probably break but I want to know how it affects this system...)

pkg/registry/rbac/rest/storage_rbac.go
roycaihw: > An alternative to this is to permit self-requests even when blocking other requests

The purpose of the filter is to make sure no write request gets through before the apiserver updates the corresponding storage version record. I don't see how self-requests are different from other requests.  

> Another option is to do a GET first

Do you mean get the rule or get the storage version?

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/errors.go
roycaihw: > Why an error page instead of a Status formatted JSON response?

This was probably copied from the other helper. Other filters we have (authorization, requestinfo, audit) write an error page when internal error occurs. I changed it to return a Status since we claim "[Kubernetes will always return the Status kind from any API endpoint when an error occurs](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#response-status-kind)". 

> the status code

The conversation happened in https://github.com/kubernetes/enhancements/pull/1361#discussion_r345463628. [429 Too Many Requests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#error-codes) sounds less suited IMO:
>> 429 StatusTooManyRequests 
Indicates that the either the client rate limit has been exceeded or the server has received more requests then it can process.
Suggested client recovery behavior:
Read the Retry-After HTTP header from the response, and wait at least that long before retrying.


> error translation code

perhaps this? https://github.com/kubernetes/kubernetes/blob/8bdd10b7d73cb0d705b895e0c9a5646dfbb19a24/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/writers.go#L259

staging/src/k8s.io/apiserver/pkg/endpoints/filters/storageversion.go
roycaihw: removed this function. See https://github.com/kubernetes/kubernetes/pull/93873#discussion_r517830101

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/errors.go
roycaihw: removed this function. See https://github.com/kubernetes/kubernetes/pull/93873#discussion_r517830101

staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
roycaihw: API servers will still update their storage versions and serve traffic. Stale storage version records won't get garbage collected. The storage migrator won't trigger migration unless those stale records get manually removed.

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
roycaihw: The storage migrator won't be able to tell from the storage versions, if you move events to a different storage. Say server A and B keep events in etcd 1, and now server C keep events in etcd 2. Their records look the same: 

```
apiVersion: internal.apiserver.k8s.io/v1alpha1
kind: StorageVersion
metadata:
  name: core.events
status:
  storageVersions:
  - apiServerID: A
    decodableVersions:
    - v1
    - events.k8s.io/v1beta1
    - events.k8s.io/v1
    encodingVersion: v1
  - apiServerID: B
    decodableVersions:
    - v1
    - events.k8s.io/v1beta1
    - events.k8s.io/v1
    encodingVersion: v1
  - apiServerID: C
    decodableVersions:
    - v1
    - events.k8s.io/v1beta1
    - events.k8s.io/v1
    encodingVersion: v1
```

so migration will be triggered even though it shouldn't. Now you have events stored in two places (assume the writes get load balanced) in the cluster. And when you read events from apisevers, different apiservers have different vision.

staging/src/k8s.io/apiserver/pkg/endpoints/filters/storageversion.go
roycaihw: aggregated servers need to add the same mechanism using the building blocks we provide (lease controller, storage version manager, filter), then the storage migrator will be able to manage migration for them. 

for extension apis we (apiextensions-apiserver) will collect and update the storage version objects. The rest works the same. I will do that in a followup PR.

test/integration/storageversion/storage_version_filter_test.go
roycaihw: ack

staging/src/k8s.io/apiserver/pkg/endpoints/installer.go
roycaihw: another option is to group together `*metav1.APIResource` and `*storageversion.ResourceInfo` (what the [original PR](https://github.com/kubernetes/kubernetes/pull/84016/commits/c27760c5c6f6a8a5c572c6806ad349d1d73bb4f0#diff-a8da7b9bda33b6055a1f5deb7d0c33058c5d706ec5b44c665191a359a43160c6R195) did) in a bigger "API resource info". Later we need to [extract them separately](https://github.com/kubernetes/kubernetes/pull/84016/commits/c27760c5c6f6a8a5c572c6806ad349d1d73bb4f0#diff-43ce7444baa0a48d4a4ec37170c8def820afbeab7b16b5df37894d73b96d9071R111) because they are used in different places.

staging/src/k8s.io/apiserver/pkg/endpoints/filters/storageversion.go
roycaihw: > The rest works the same

I mean extension apis don't need the filter. The extensions apiserver will first update the storage version objects, then construct the custom resource handler. The storage migrator works the same way for extension apis as it works for built-in apis.

pkg/registry/rbac/rest/storage_rbac.go
lavalamp: Getting the rule.

staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
lavalamp: Hm, sounds like we at least need to document that you shouldn't run like that.

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
lavalamp: Yeah, that's a disaster, but I think it already is. We can document that for now, you have to shut everything down and manually move things in etcd if you want to do this.

staging/src/k8s.io/apiserver/pkg/endpoints/installer.go
lavalamp: I guess I won't ask for a change.

lavalamp:
OK I think this looks good, shall we merge the identity one first though?

staging/src/k8s.io/apiserver/pkg/storageversion/manager.go
roycaihw: ack. Will mention it in storage migrator doc.

pkg/registry/rbac/rest/storage_rbac.go
roycaihw: The ReconcileRole logic does do a [get](https://github.com/kubernetes/kubernetes/blob/bd95fb101dd83cfca9d95e55b2f0d29720016e51/staging/src/k8s.io/component-helpers/auth/rbac/reconciliation/reconcile_role.go#L110) first and skips if no Reconciliation is needed

roycaihw:
> shall we merge the identity one first though

there will be some minor conflicts. Let's wait for the identity one to get in since it's in the queue already. 

I will rework the last commit (GroupVersion validation) a little bit and adopt `t.Run` in the meantime

test/integration/storageversion/storage_version_filter_test.go
roycaihw: added
```
--- PASS: TestStorageVersionBootstrap (7.97s)
    --- PASS: TestStorageVersionBootstrap/before_storage_version_update (1.72s)
        --- PASS: TestStorageVersionBootstrap/before_storage_version_update/write_to_k8s_native_API_object_should_be_blocked (1.68s)
        --- PASS: TestStorageVersionBootstrap/before_storage_version_update/write_to_CRD_should_be_blocked (0.00s)
        --- PASS: TestStorageVersionBootstrap/before_storage_version_update/write_to_APIService_should_be_blocked (0.00s)
        --- PASS: TestStorageVersionBootstrap/before_storage_version_update/write_to_CR_should_pass (0.03s)
        --- PASS: TestStorageVersionBootstrap/before_storage_version_update/write_to_the_storage_version_API_should_pass (0.00s)
        --- PASS: TestStorageVersionBootstrap/before_storage_version_update/write_to_non-persisted_API_should_pass (0.00s)
        --- PASS: TestStorageVersionBootstrap/before_storage_version_update/read_of_k8s_native_API_should_pass (0.00s)
    --- PASS: TestStorageVersionBootstrap/after_storage_version_update (0.02s)
        --- PASS: TestStorageVersionBootstrap/after_storage_version_update/write_to_k8s_native_API_object_should_pass (0.00s)
        --- PASS: TestStorageVersionBootstrap/after_storage_version_update/write_to_CRD_should_pass (0.00s)
        --- PASS: TestStorageVersionBootstrap/after_storage_version_update/write_to_APIService_should_pass (0.00s)
        --- PASS: TestStorageVersionBootstrap/after_storage_version_update/write_to_the_storage_version_API_should_pass (0.01s)
        --- PASS: TestStorageVersionBootstrap/after_storage_version_update/write_to_non-persisted_API_should_pass (0.00s)
        --- PASS: TestStorageVersionBootstrap/after_storage_version_update/read_of_k8s_native_API_should_pass (0.00s)
    --- PASS: TestStorageVersionBootstrap/after_storage_version_manager_complete (0.60s)
PASS
ok  	k8s.io/kubernetes/test/integration/storageversion	8.324s
```

staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
roycaihw: I added a commit to make enabling apiserver-identity a hard requirement to run this feature.

pkg/apis/apiserverinternal/validation/validation_test.go
```diff
@@ -61,6 +61,62 @@ func TestValidateServerStorageVersion(t *testing.T) {
...
+			},
+			expectedErr: `[].encodingVersion: Invalid value: "/v3": group part must be non-empty`,
+		},
+		{
+			ssv: apiserverinternal.ServerStorageVersion{
+				APIServerID:       "fea",
+				EncodingVersion:   "v1",
+				DecodableVersions: []string{"v1", "mygroup_com/v2"},
+			},
+			expectedErr: `[].decodableVersions[1]: Invalid value: "mygroup_com/v2": group part a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')`,
```
lavalamp, line 102: missing a "must be"

pkg/apis/apiserverinternal/validation/validation_test.go
```diff
@@ -61,6 +61,62 @@ func TestValidateServerStorageVersion(t *testing.T) {
...
+		},
+		{
+			ssv: apiserverinternal.ServerStorageVersion{
+				APIServerID:       "fea",
+				EncodingVersion:   "v1",
+				DecodableVersions: []string{"v1", "mygroup.com/v2_"},
+			},
+			expectedErr: `[].decodableVersions[1]: Invalid value: "mygroup.com/v2_": version part a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')`,
```
lavalamp, line 110:
```suggestion
			expectedErr: `[].decodableVersions[1]: Invalid value: "mygroup.com/v2_": version part must be a DNS-1035 label, meaning it must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')`,
```

pkg/apis/apiserverinternal/validation/validation_test.go
```diff
@@ -61,6 +61,62 @@ func TestValidateServerStorageVersion(t *testing.T) {
...
+		},
+		{
+			ssv: apiserverinternal.ServerStorageVersion{
+				APIServerID:       "fea",
+				EncodingVersion:   "v1",
+				DecodableVersions: []string{"v1", "mygroup.com/v2/myresource"},
+			},
+			expectedErr: `[].decodableVersions[1]: Invalid value: "mygroup.com/v2/myresource": an apiVersion a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?') with an optional DNS subdomain prefix and '/' (e.g. 'example.com/MyVersion')`,
```
lavalamp, line 118:
```suggestion
			expectedErr: `[].decodableVersions[1]: Invalid value: "mygroup.com/v2/myresource": an apiVersion is a DNS-1035 label, which must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?') with an optional DNS subdomain prefix and '/' (e.g. 'example.com/MyVersion')`,
```

(there's probably lots of ways to fix the grammar, I don't have a preference about how)

pkg/registry/rbac/rest/storage_rbac.go
lavalamp: Great.

lavalamp: I guess in the future we might want to think about just retrying, period, rather than build up a long list of reasons.

staging/src/k8s.io/apiserver/pkg/storageversion/updater.go
```diff
@@ -91,10 +91,16 @@ func singleUpdate(c Client, apiserverID string, gr schema.GroupResource, encodin
...
 	updatedSV := localUpdateStorageVersion(sv, apiserverID, encodingVersion, decodableVersions)
 	if shouldCreate {
-		_, err := c.Create(context.TODO(), updatedSV, metav1.CreateOptions{})
+		createdSV, err := c.Create(context.TODO(), updatedSV, metav1.CreateOptions{})
+		if err != nil {
+			return err
+		}
+		// assign the calculated status to the object just created, then update status
+		createdSV.Status = updatedSV.Status
+		_, err = c.UpdateStatus(context.TODO(), createdSV, metav1.UpdateOptions{})
```
lavalamp, line 100: What happens if another apiserver modifies this object after the create but before this update?

staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
```diff
@@ -264,6 +267,35 @@ func (c completedConfig) NewWithDelegate(delegationTarget genericapiserver.Deleg
...
+			// Technically an apiserver only needs to update storage version once during bootstrap.
+			// Reconcile StorageVersion objects every 10 minutes will help in the case that the
+			// StorageVersion objects get accidentally modified/deleted by a different agent. In that
+			// case, the reconciliation ensures future storage migration still works. If nothing gets
+			// changed, the reconciliation update is a noop and gets short-circuited by the apiserver,
+			// therefore won't change the resource version and trigger storage migration.
+			go wait.PollImmediateUntil(10*time.Minute, func() (bool, error) {
+				// All apiservers (aggregator-apiserver, kube-apiserver, apiextensions-apiserver)
+				// share the same generic apiserver config. The same StorageVersion manager is used
+				// to register all built-in resources when the generic apiservers install APIs.
```
lavalamp, line 284: It'd be great to somewhere capture what actions authors of aggregated apiservers need to take. Maybe in the release notes?

lavalamp:
OK I made it all the way through, comments are pretty minor, I think.

staging/src/k8s.io/apiserver/pkg/storageversion/updater.go
roycaihw: The updater handles conflict and retry

staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go
roycaihw: will add it to storage migrator doc.

liggitt:
The alpha e2e job started failing many tests as soon as this was merged - see https://testgrid.k8s.io/google-gce#gci-gce-alpha-enabled-default&width=20

liggitt:
controller manager is crash-looping with this error for ~20 minutes before the test fails:

```
F1111 00:00:09.182571       9 controllermanager.go:243] error building controller context: unable to get any supported resources from server
goroutine 137 [running]:
k8s.io/kubernetes/vendor/k8s.io/klog/v2.stacks(0xc00000e001, 0xc0000499a0, 0x8d, 0x9b)
```

roycaihw:
> The alpha e2e job started failing many tests as soon as this was merged

looking

roycaihw:
The rbac hook failed to wait for storage version registration to complete. `last seen error: <nil>` means the storage version manager hasn't started registering clusterroles yet. 
```
E1110 05:18:57.295440       9 storage_rbac.go:236] unable to reconcile clusterrole.rbac.authorization.k8s.io/cluster-admin: wait for storage version registration to complete for resource: clusterroles.rbac.authorization.k8s.io, last seen error: <nil>
E1110 05:18:57.670415       9 storage_rbac.go:236] unable to reconcile clusterrole.rbac.authorization.k8s.io/system:discovery: wait for storage version registration to complete for resource: clusterroles.rbac.authorization.k8s.io, last seen error: <nil>
```
Happened 0.6s late:
```
$ grep "PUT.*storageversion.*clusterroles/status" ~/Desktop/alpha/kube-apiserver.log
I1110 05:18:57.827333       9 httplog.go:89] "HTTP" verb="PUT" URI="/apis/internal.apiserver.k8s.io/v1alpha1/storageversions/rbac.authorization.k8s.io.clusterroles/status" latency="6.996106ms" userAgent="kube-apiserver/v1.20.0 (linux/amd64) kubernetes/6e95025" srcIP="[::1]:51840" resp=200
```
The first storage version registration:
```
$ grep "PUT.*storageversion.*/status" ~/Desktop/alpha/kube-apiserver.log | grep "I1110 05:17\|I1110 05:18\|I1110 05:19" | head -1
I1110 05:18:56.401513       9 httplog.go:89] "HTTP" verb="PUT" URI="/apis/internal.apiserver.k8s.io/v1alpha1/storageversions/core.configmaps/status" latency="30.795772ms" userAgent="kube-apiserver/v1.20.0 (linux/amd64) kubernetes/6e95025" srcIP="[::1]:51840" resp=200
```
The last: 
```
$ grep "PUT.*storageversion.*/status" ~/Desktop/alpha/kube-apiserver.log | grep "I1110 05:17\|I1110 05:18\|I1110 05:19" | tail -1
I1110 05:18:58.150728       9 httplog.go:89] "HTTP" verb="PUT" URI="/apis/internal.apiserver.k8s.io/v1alpha1/storageversions/storage.k8s.io.volumeattachments/status" latency="11.690226ms" userAgent="kube-apiserver/v1.20.0 (linux/amd64) kubernetes/6e95025" srcIP="[::1]:51840" resp=200
```
<2s in total.


In this PR we've already made the rbac hook retry on 503. Let me verify the backoff behavior and decide if we want to extend the retry.

roycaihw:
```
$ grep -o ".*POST.*clusterrole.*503" ~/Desktop/alpha/kube-apiserver.log
# cluster-admin - failed
I1110 05:18:56.909640       9 httplog.go:89] "HTTP" verb="POST" URI="/apis/rbac.authorization.k8s.io/v1/clusterroles" latency="439.596µs" userAgent="kube-apiserver/v1.20.0 (linux/amd64) kubernetes/6e95025" srcIP="[::1]:51840" resp=503
I1110 05:18:56.926166       9 httplog.go:89] "HTTP" verb="POST" URI="/apis/rbac.authorization.k8s.io/v1/clusterroles" latency="457.404µs" userAgent="kube-apiserver/v1.20.0 (linux/amd64) kubernetes/6e95025" srcIP="[::1]:51840" resp=503
I1110 05:18:57.015692       9 httplog.go:89] "HTTP" verb="POST" URI="/apis/rbac.authorization.k8s.io/v1/clusterroles" latency="300.912µs" userAgent="kube-apiserver/v1.20.0 (linux/amd64) kubernetes/6e95025" srcIP="[::1]:51840" resp=503
I1110 05:18:57.295088       9 httplog.go:89] "HTTP" verb="POST" URI="/apis/rbac.authorization.k8s.io/v1/clusterroles" latency="360.648µs" userAgent="kube-apiserver/v1.20.0 (linux/amd64) kubernetes/6e95025" srcIP="[::1]:51840" resp=503
# system:discovery - failed
I1110 05:18:57.307569       9 httplog.go:89] "HTTP" verb="POST" URI="/apis/rbac.authorization.k8s.io/v1/clusterroles" latency="310.292µs" userAgent="kube-apiserver/v1.20.0 (linux/amd64) kubernetes/6e95025" srcIP="[::1]:51840" resp=503
I1110 05:18:57.331200       9 httplog.go:89] "HTTP" verb="POST" URI="/apis/rbac.authorization.k8s.io/v1/clusterroles" latency="438.666µs" userAgent="kube-apiserver/v1.20.0 (linux/amd64) kubernetes/6e95025" srcIP="[::1]:51840" resp=503
I1110 05:18:57.405698       9 httplog.go:89] "HTTP" verb="POST" URI="/apis/rbac.authorization.k8s.io/v1/clusterroles" latency="1.506347ms" userAgent="kube-apiserver/v1.20.0 (linux/amd64) kubernetes/6e95025" srcIP="[::1]:51840" resp=503
I1110 05:18:57.668579       9 httplog.go:89] "HTTP" verb="POST" URI="/apis/rbac.authorization.k8s.io/v1/clusterroles" latency="291.535µs" userAgent="kube-apiserver/v1.20.0 (linux/amd64) kubernetes/6e95025" srcIP="[::1]:51840" resp=503
# system:monitoring - created
I1110 05:18:57.706892       9 httplog.go:89] "HTTP" verb="POST" URI="/apis/rbac.authorization.k8s.io/v1/clusterroles" latency="287.548µs" userAgent="kube-apiserver/v1.20.0 (linux/amd64) kubernetes/6e95025" srcIP="[::1]:51840" resp=503
I1110 05:18:57.760045       9 httplog.go:89] "HTTP" verb="POST" URI="/apis/rbac.authorization.k8s.io/v1/clusterroles" latency="1.062415ms" userAgent="kube-apiserver/v1.20.0 (linux/amd64) kubernetes/6e95025" srcIP="[::1]:51840" resp=503
# system:basic-user - created
# system:public-info-viewer - created
# ...
```

roycaihw:
time differences:
```
I1110 05:18:56.909640
17ms
I1110 05:18:56.926166
89ms
I1110 05:18:57.015692
280ms
I1110 05:18:57.295088
```

steps in theory (5x factor, 0.1 Jitter, 4 steps)
1. 10-11ms
2. 50-55ms
3. 250-275ms

Note: before this PR, retry on happens on conflict. I doubt if the 5x backoff makes sense. 

If we add more:
4. 1250-1375ms
5. 6250-6875ms

One step might be sufficient. Adding two steps so we are extra safe. 

Also, when the hook talks to the same cluster, the wait on 503 should only happen on the first clusterrole created by the hook, so we only need to extend retries for the first reconciliation. I assume it's true (the hook talks to the local cluster) for HA clusters as well.

roycaihw:
> for the first reconciliation

I don't have fast way to track the first non-noop reconciliation. Sent a small workaround and keeping looking.

PR #93876, opened 2020-08-11
Fix waiting for PVC to get bound

jsafrane:
**What type of PR is this?**
/kind bug
/kind flake

When API server returns error on Get(), the loop that waits for PVCs to get bound should not finish and should not return nil. It should keep waiting, as the log message suggests.

Fixed `Poll` parameter name to `poll` and some error messages when at it (as separate commits). Especially this error message was quite confusing, because it missed `:`: `Failed waiting for PVC to be bound PV Get API error: <API server error>`

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/framework/pv/pv.go
+++ b/test/e2e/framework/pv/pv.go
@@ -714,12 +714,12 @@ func WaitForPVClaimBoundPhase(client clientset.Interface, pvclaims []*v1.Persist
 }
 
 // WaitForPersistentVolumePhase waits for a PersistentVolume to be in a specific phase or until timeout occurs, whichever comes first.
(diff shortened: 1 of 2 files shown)


jsafrane:
/assign @oomichi @andrewsykim
for approval
Note the real fix is only the first commit, the rest is code quality improvement.

andrewsykim (commented):
/approve
/lgtm

Should we put this into the v1.19 milestone since it's fixing a flaky test?

jsafrane:
It does not seem to flake in kubernetes/kubernetes, so I don't mind waiting for 1.20. master is going to open soon, right?

PR #93772, opened 2020-08-07
Add missing termination metrics to priority-and-fairness handler

bjrara:
**What type of PR is this?**

/kind regression

**What this PR does / why we need it**:
In maxinflight handler, when the request is terminated by tooManyRequest, it will be added to `apiserver_request_terminations_total` metrics. After migrate to priority-and-fairness handler, this metrics is missing.

**Which issue(s) this PR fixes**:

Fixes #
Missing metrics in the priority-and-fairness handler

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #93877, opened 2020-08-11
Bugfix missing metrics and test cases to ensure terminated request metrics is not missing

bjrara:
**What type of PR is this?**
/kind cleanup

Add missing metric `apiserver_dropped_requests_total` in apfhandler.
Add test cases for this bugfix.
This is a follow up PR of https://github.com/kubernetes/kubernetes/pull/93772.

Fixes #

```release-note
NONE
```


yue9944882:
i believe what @lavalamp expected is a unit test (instead of integration) and these [testing utilities](https://github.com/kubernetes/component-base/tree/master/metrics/testutil) around metrics will probably help

bjrara:
> i believe what @lavalamp expected is a unit test (instead of integration) and these [testing utilities](https://github.com/kubernetes/component-base/tree/master/metrics/testutil) around metrics will probably help

@yue9944882  Thanks for the reminder. Marked WIP to add UT.

lavalamp:
Yeah I'd prefer a unit test that exercises all the request entrance/exit paths.

bjrara:
> Yeah I'd prefer a unit test that exercises all the request entrance/exit paths.

@lavalamp Could you help suggest which unit tests would be better? Unit tests with a complete handler chain installed or unit tests to cover the path of a pure priority-and-fairness handler?

lavalamp:
I don't have a specific answer.
* It's not important to test non-APF parts of the handler chain
* It is important to test that every way a request can enter the APF system causes metric updates
* It is important to test that every way a request can exit the APF system causes metric updates

It seems easier to me to prove that entrance and exit updates are in 1:1 correspondence in unit tests that force every code path to be exercised. Maybe the integration tests get us some certainty? E.g., did this new test fail before your earlier fix?

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,194 @@
...
+	apifilters "k8s.io/apiserver/pkg/endpoints/filters"
+	epmetrics "k8s.io/apiserver/pkg/endpoints/metrics"
+	apirequest "k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/apiserver/pkg/server/mux"
+	utilflowcontrol "k8s.io/apiserver/pkg/util/flowcontrol"
+	fq "k8s.io/apiserver/pkg/util/flowcontrol/fairqueuing"
+	fcmetrics "k8s.io/apiserver/pkg/util/flowcontrol/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+	"k8s.io/component-base/metrics/testutil"
+	"k8s.io/kubernetes/staging/src/k8s.io/apiserver/pkg/apis/flowcontrol/bootstrap"
```
yue9944882, line 20: remove the staging imports and separate them into groups https://github.com/golang/go/wiki/CodeReviewComments#imports

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,194 @@
...
+		r = r.WithContext(apirequest.WithUser(r.Context(), &user.DefaultInfo{
+			Groups: []string{"system:unauthenticated"},
+		}))
+		apfHandler.ServeHTTP(w, r)
+	})
+}
+
+const apfServerAddress = "127.0.0.1:6443"
+
+func newApfServer(apfHandler http.Handler) (*http.Server, error) {
```
yue9944882, line 73: i presume this should be replace by `httptest.NewServer` which doesnt necessarily bind a constant address/port in the test. otherwise the test will fail on running in parallel, isnt it?

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+	checkForExpectedMetrics(t, metrics, []string{
+		"apiserver_current_inflight_requests",
+		"apiserver_current_inqueue_requests",
+		"apiserver_flowcontrol_read_vs_write_request_count_watermarks_bucket",
+
+	})
+}
+
+func scrapeMetrics() (testutil.Metrics, error) {
```
yue9944882, line 170: metrics gethering doesnt have to go thru the handler. in favor of `metrics.DefaultGatherer` to avoid unnecessary overhead..

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,194 @@
...
+func newApfServer(apfHandler http.Handler) (*http.Server, error) {
+	listener, err := net.Listen("tcp", apfServerAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	requestInfoFactory := &apirequest.RequestInfoFactory{APIPrefixes: sets.NewString("apis", "api"), GrouplessAPIPrefixes: sets.NewString("api")}
+	handler := apifilters.WithRequestInfo(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		r = r.WithContext(apirequest.WithUser(r.Context(), &user.DefaultInfo{
+			Groups: []string{"system:unauthenticated"},
```
yue9944882, line 82: nit: `user.AllUnauthenticated` instead of hand-coded

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
bjrara: Fixed

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
bjrara: Fixed, and thanks for the explanation!

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
bjrara: Done

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
bjrara: Done

bjrara:
@lavalamp Test cases updated

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,170 @@
...
+	}()
+	assert.NoError(t, err)
+
+	//wait metrics to be collected
+	time.Sleep(time.Second)
+
+	checkForExpectedMetrics(t, []string{
+		"apiserver_current_inflight_requests",
+		"apiserver_current_inqueue_requests",
+		"apiserver_flowcontrol_read_vs_write_request_count_watermarks",
```
MikeSpreitzer, line 146: apiserver_flowcontrol_read_vs_write_request_count_watermarks appears if and only if apiserver_flowcontrol_read_vs_write_request_count_samples appears.  They are not guaranteed to appear in such a quick test.  OTOH, if the test waits at least a millisecond (see https://github.com/kubernetes/apiserver/blob/09eab1cf2cce55a46e5d9042c90f136d91383e19/pkg/util/flowcontrol/metrics/metrics.go#L125) between `fcmetrics.Register()` and invoking the APF filter, then those two should appear.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	go func() {
+		if err := expectHTTPGet(fmt.Sprintf("%s/api/v1/namespaces/default", server.URL), http.StatusTooManyRequests); err != nil {
+			t.Error(err)
+		}
+	}()
+	assert.NoError(t, err)
+
+	//wait metrics to be collected
+	time.Sleep(time.Second)
```
MikeSpreitzer, line 115: Waiting exactly one second produces a race.  I would wait longer.

MikeSpreitzer (changes requested):

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func (t fakeApfFilter) Run(stopCh <-chan struct{}) error {
+	return nil
+}
+
+func (t fakeApfFilter) Install(c *mux.PathRecorderMux) {
+}
+
+func newApfDecisionHandler(decision int) http.Handler {
```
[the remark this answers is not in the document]
bjrara: Merged two function into one.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	server, err := newApfServer(apfHandler)
+	assert.NoError(t, err)
+	defer server.Close()
+
+	if err := expectHTTPGet(fmt.Sprintf("%s/api/v1/namespaces?watch=true", server.URL), http.StatusOK); err != nil {
+		// request should not be rejected
+		t.Error(err)
+	}
+	assert.NoError(t, err)
```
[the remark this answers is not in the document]
bjrara: Fixed

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,170 @@
...
+	}), longRunningRequestCheck, fakeApfFilter{mockDecision: decision})
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		r = r.WithContext(apirequest.WithUser(r.Context(), &user.DefaultInfo{
+			Groups: []string{user.AllUnauthenticated},
+		}))
+		apfHandler.ServeHTTP(w, r)
+	})
+}
+
+func newApfServer(apfHandler http.Handler) (*httptest.Server, error) {
```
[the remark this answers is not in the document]
bjrara: Fixed

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,170 @@
...
+		metricsIndexer[metric] = nil
+	}
+	var found int
+	for _, mf := range metricsFamily {
+		if _, ok := metricsIndexer[*mf.Name]; !ok {
+			continue
+		}
+		found++
+	}
+	if found != len(expectedMetrics) {
```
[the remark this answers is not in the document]
bjrara: Done

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
bjrara: Thanks for the explanation. Fixed the possible race here.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
bjrara: Set the duration to 1.5s.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,170 @@
...
+func (t fakeApfFilter) Handle(ctx context.Context,
+	requestDigest utilflowcontrol.RequestDigest,
+	noteFn func(fs *fctypesv1a1.FlowSchema, pl *fctypesv1a1.PriorityLevelConfiguration),
+	queueNoteFn fq.QueueNoteFn,
+	execFn func(),
+) {
+	noteFn(bootstrap.SuggestedFlowSchemaGlobalDefault, bootstrap.SuggestedPriorityLevelConfigurationGlobalDefault)
+	switch t.mockDecision {
+	case decisionExecute:
+		queueNoteFn(true)
```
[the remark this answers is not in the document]
bjrara: Thanks for the details in the new PR. I hope I've faked right this time...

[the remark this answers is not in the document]
MikeSpreitzer: Looks good now

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,211 @@
...
+		"apiserver_dropped_requests_total",
+	})
+}
+
+func checkForExpectedMetrics(t *testing.T, expectedMetrics []string) {
+	metricsFamily, err := legacyregistry.DefaultGatherer.Gather()
+	if err != nil {
+		t.Fatalf("Failed to gather metrics %v", err)
+	}
+
```
MikeSpreitzer, line 195: The logic here would be simpler if it first makes a map containing all the metric names from `metricsFamily` and then iterates through `expectedMetrics` to check for presence.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,211 @@
...
+	server := newApfServer(decisionReject)
+	defer server.Close()
+
+	if err := expectHTTPGet(fmt.Sprintf("%s/api/v1/namespaces/default", server.URL), http.StatusTooManyRequests); err != nil {
+		t.Error(err)
+	}
+
+	// wait metrics to be collected
+	time.Sleep(time.Millisecond * 1500)
```
MikeSpreitzer, line 113: The metrics tested in this func do not require any waiting, they are not set by an asynchronous periodic process.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,211 @@
...
+	noteFn func(fs *fctypesv1a1.FlowSchema, pl *fctypesv1a1.PriorityLevelConfiguration),
+	queueNoteFn fq.QueueNoteFn,
+	execFn func(),
+) {
+	if t.mockDecision == decisionError {
+		panic("Handle should not be invoked")
+	}
+	noteFn(bootstrap.SuggestedFlowSchemaGlobalDefault, bootstrap.SuggestedPriorityLevelConfigurationGlobalDefault)
+	switch t.mockDecision {
+	case decisionExempt:
```
MikeSpreitzer, line 48: BTW, exempt requests are not the only ones executed without queuing.  A priority level can be configured to impose a concurrency limit and do no queuing.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,211 @@
...
+		"apiserver_request_terminations_total",
+		"apiserver_dropped_requests_total",
+	})
+}
+
+func TestApfExemptRequest(t *testing.T) {
+	epmetrics.Register()
+	fcmetrics.Register()
+
+	// wait ReadWriteConcurrencyObserverPairGenerator to be registered
```
MikeSpreitzer, line 125: The wait here is not for an asynchronous process.  It is for time to pass so that the loop at https://github.com/kubernetes/apiserver/blob/09eab1cf2cce55a46e5d9042c90f136d91383e19/pkg/util/flowcontrol/metrics/sample_and_watermark.go#L185 will go through a non-zero number of iterations.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+	server := newApfServer(decisionExempt)
+	defer server.Close()
+
+	if err := expectHTTPGet(fmt.Sprintf("%s/api/v1/namespaces/default", server.URL), http.StatusOK); err != nil {
+		t.Error(err)
+	}
+
+	// wait metrics to be collected
+	time.Sleep(time.Millisecond * 1500)
```
MikeSpreitzer, line 136: On second throught, the problem here is relying on the CPU scheduling to get asynchronous work done by any particular deadline.  It would be more robust here to have a loop that sleeps for a little while (one second would be fine) and then looks for all the expected metrics, terminating the loop successfully if they are all found.  Put a limit on the number of iterations, and complain about the metrics not found in the last iteration.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,211 @@
...
+	server := newApfServer(decisionExecute)
+	defer server.Close()
+
+	if err := expectHTTPGet(fmt.Sprintf("%s/api/v1/namespaces/default", server.URL), http.StatusOK); err != nil {
+		t.Error(err)
+	}
+
+	// wait metrics to be collected
+	time.Sleep(time.Millisecond * 1500)
+
```
MikeSpreitzer, line 161: This one also fails when I run it in commit 3121fe16a20e046d85419a4aff06edbe220b39b2:
```
--- FAIL: TestApfExecuteRequest (1.57s)
    priority-and-fairness_test.go:208: Scraped metrics did not include expected metric apiserver_current_inqueue_requests
```

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,211 @@
...
+	if err := expectHTTPGet(fmt.Sprintf("%s/api/v1/namespaces/default", server.URL), http.StatusTooManyRequests); err != nil {
+		t.Error(err)
+	}
+
+	// wait metrics to be collected
+	time.Sleep(time.Millisecond * 1500)
+
+	checkForExpectedMetrics(t, []string{
+		"apiserver_current_inqueue_requests",
```
MikeSpreitzer, line 184: This one also fails to show up in commit 3121fe16a20e046d85419a4aff06edbe220b39b2:
```
--- FAIL: TestApfCancelWaitRequest (1.51s)
    priority-and-fairness_test.go:208: Scraped metrics did not include expected metric apiserver_current_inqueue_requests
```

bjrara: Could you help check if it fail again? If so, could you tell me how did you reproduce it?

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,246 @@
...
+)
+
+type fakeApfFilter struct {
+	mockDecision int
+	postEnqueue  func()
+	postDequeue  func()
+}
+
+func (t fakeApfFilter) MaintainObservations(stopCh <-chan struct{}) {
+	panic("implement me")
```
MikeSpreitzer, line 41: "do nothing" would be a valid implementation here, right?

MikeSpreitzer (commented):
(more later)

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
bjrara: yes, fixed.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
MikeSpreitzer: I just tested with commit 4981cd1ae0812bdb783e70b03402b512fa73ba4c , and got no failures.

I test on my Mac by simply `cd` to the `filters` directory and then `go test` or, to be more specific, `go test -run TestApfSkipLongRunningRequest`.

MikeSpreitzer:
/priority important-soon
This includes a bug fix.  The title could be clearer about this.

MikeSpreitzer (approved):

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,261 @@
...
+	for i := 0; i < maxRetries; i++ {
+		t.Logf("Check for expected metrics with retry %d", i)
+		metricsFamily, err := legacyregistry.DefaultGatherer.Gather()
+		if err != nil {
+			t.Fatalf("Failed to gather metrics %v", err)
+		}
+
+		metrics := map[string]interface{}{}
+		for _, mf := range metricsFamily {
+			mf := mf
```
MikeSpreitzer, line 235: I think this line does not accomplish anything.

MikeSpreitzer (approved):
/LGTM
This looks correct, although it has one unnecessary line that could be deleted.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+
+		time.Sleep(1 * time.Second)
+	}
+	for _, checkError := range checkErrors {
+		t.Error(checkError)
+	}
+}
+
+func checkForExpectedMetrics(expectedMetrics []string, metrics map[string]interface{}) []error {
```
ahg-g, line 251: is the expectation that the metrics will not exist if they didn't get set at least once? can check for explicit values? we can reset the metrics at the beginning of each test I think (or assert on some delta).

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,261 @@
...
+		mockDecision: decision,
+		postEnqueue: func() {
+			if atomicReadOnlyWaiting != 1 {
+				t.Errorf("Wanted %d requests in queue, got %d", 1, atomicReadOnlyWaiting)
+			}
+		},
+		postDequeue: func() {
+			if atomicReadOnlyWaiting != 0 {
+				t.Errorf("Wanted %d requests in queue, got %d", 0, atomicReadOnlyWaiting)
+			}
```
ahg-g, line 116: can you add a comment explaining that atomicReadOnlyWaiting can be either 1 or 0 because we are testing one request at a time.

ahg-g: also, can we test the case where multiple requests waiting or executing?

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+	server := newApfServer(decisionNoQueuingExecute, t)
+	defer server.Close()
+
+	if err := expectHTTPGet(fmt.Sprintf("%s/api/v1/namespaces/default", server.URL), http.StatusOK); err != nil {
+		t.Error(err)
+	}
+
+	checkForExpectedMetricsWithRetry(t, []string{
+		"apiserver_current_inflight_requests",
```
ahg-g, line 177: where is this metric defined?

MikeSpreitzer: https://github.com/kubernetes/apiserver/blob/8dc7971548234eb6700b52291a7dd817ddc8d323/pkg/endpoints/metrics/metrics.go#L166

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,261 @@
...
+		"apiserver_request_terminations_total",
+		"apiserver_dropped_requests_total",
+	})
+}
+
+func TestApfExemptRequest(t *testing.T) {
+	epmetrics.Register()
+	fcmetrics.Register()
+
+	// wait the first sampleAndWaterMark metrics to be collected
```
MikeSpreitzer, line 166: This comment is misleading, as @ahg-g 's remark below indicates.
Perhaps something along the following lines would be better.
> Wait for at least one sampling window to pass since creation of `k8s.io/apiserver/pkg/util/flowcontrol/metrics.ReadWriteConcurrencyObserverPairGenerator`,
> so that an observation will cause some data to go into the Prometheus metrics.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,261 @@
...
+	time.Sleep(time.Millisecond * 50)
```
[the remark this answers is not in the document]
MikeSpreitzer: I think you misunderstood what is going on here.  I have suggested an update to the comment that should point the reader in the right direction.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,261 @@
...
+		"apiserver_flowcontrol_read_vs_write_request_count_samples",
+	})
+}
+
+func TestApfExecuteRequest(t *testing.T) {
+	epmetrics.Register()
+	fcmetrics.Register()
+
+	// wait the first sampleAndWaterMark metrics to be collected
+	time.Sleep(time.Millisecond * 50)
```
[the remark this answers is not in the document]
MikeSpreitzer: See above.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
MikeSpreitzer: Yes, Prometheus `HistogramVector` objects are being used, which means the actual metrics do not exist until data is put into at least one Histogram in the Vector.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
ahg-g: Is there a deterministic way to check that a sampling window has passed? is there an API we can call to tell us that?

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
ahg-g: can we check for specific values instead of just checking that the metric exist?

MikeSpreitzer: yes, if the test func waits for at least one sampling period, after any activity, before reading the metrics.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
MikeSpreitzer: We can deterministically know that at least one sampling window has passed by waiting for at least the duration of a sampling window.  The existing test func does this.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
bjrara: Added

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
bjrara: Updated comment. Much appreciated your suggestions. @MikeSpreitzer

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
bjrara: Removed

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,354 @@
...
+	for i := 0; i < 4; i++ {
+		preWait[i] = sync.WaitGroup{}
+		preWait[i].Add(concurrentRequests)
+	}
+
+	postWait := make([]sync.WaitGroup, 3)
+	for i := 0; i < 3; i++ {
+		postWait[i] = sync.WaitGroup{}
+		postWait[i].Add(concurrentRequests)
+	}
```
MikeSpreitzer, line 240: The members of these arrays are all used in individual ways.  We would normally declare individuals in this case.
```suggestion
	var preStartExecute, postStartExecute, preEnqueue, postEnqueue, preDequeue, postDequeue, finishExecute sync.WaitGroup
	for _, wg := range ([]*sync.WaitGroup{&preStartExecute, &postStartExecute, &preEnqueue, &postEnqueue, &preDequeue, &postDequeue, &finishExecute}) {
	  wg.Add(concurrentRequests)
	}
```

ahg-g (commented):
few nits

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,354 @@
...
+const (
+	decisionNoQueuingExecute = iota
+	decisionQueuingExecute
+	decisionCancelWait
+	decisionReject
+	decisionSkipFilter
+)
+
+type fakeApfFilter struct {
+	mockDecision int
```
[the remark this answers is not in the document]
bjrara: Thanks for the suggestion! Updated the declaration.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,354 @@
...
+	apirequest "k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/apiserver/pkg/server/mux"
+	utilflowcontrol "k8s.io/apiserver/pkg/util/flowcontrol"
+	fq "k8s.io/apiserver/pkg/util/flowcontrol/fairqueuing"
+	fcmetrics "k8s.io/apiserver/pkg/util/flowcontrol/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+const (
+	decisionNoQueuingExecute = iota
```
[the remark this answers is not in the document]
bjrara: Thanks for the suggestion! Updated the declaration.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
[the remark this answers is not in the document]
bjrara: Done

AndreMouche (approved):

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness.go
```diff
@@ -78,6 +82,8 @@ func WithPriorityAndFairness(
 	}
 	startOnce.Do(func() {
 		startRecordingUsage(watermark)
+	})
+	apfStartOnce.Do(func() {
```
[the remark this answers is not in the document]
bjrara: Right. Separating `startOnce` of the two filters is used to avoid race condition when running `maxinflight_test.go` and `priority-and-fairness_test.go` together, in which case both `WithPriorityAndFairness` and `WithMaxInFlightLimit` are invoked.

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
bjrara: @MikeSpreitzer Thank you for the suggestions and the detailed explanation!

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness_test.go
```diff
@@ -0,0 +1,354 @@
...
+		preWait[2].Wait()
+		if atomicReadOnlyWaiting != 0 {
+			t.Errorf("Wanted %d requests in queue, got %d", 0, atomicReadOnlyWaiting)
+		}
+		postWait[2].Done()
+		postWait[2].Wait()
+	}
+
+	postExecute := func() {
+		preWait[3].Done()
```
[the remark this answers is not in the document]
bjrara: Replaced with `finishExecute` :D :D :D

staging/src/k8s.io/apiserver/pkg/server/filters/priority-and-fairness.go
```diff
@@ -62,6 +63,9 @@ var waitingMark = &requestWatermark{
 	mutatingObserver: fcmetrics.ReadWriteConcurrencyObserverPairGenerator.Generate(1, 1, []string{epmetrics.MutatingKind}).RequestsWaiting,
 }
 
+// apfStartOnce is used to avoid sharing one-time mutex with maxinflight handler
+var apfStartOnce sync.Once
```
lavalamp, line 67: Just a note that all these globals make it impossible to e.g. install a second APF filter with different settings for long-running requests, which might be one thing we eventually want to do.

Not asking for a change now.

MikeSpreitzer:
Hooray!  Thanks @bjrara and all who helped land this.

eddiezane:
This is potentially flaking or uncovering something else. Tracking at https://github.com/kubernetes/kubernetes/issues/96168.

liggitt:
There is specifically a data race in the test introduced here - https://github.com/kubernetes/kubernetes/issues/96243

ahg-g:
https://github.com/kubernetes/kubernetes/pull/96244 fixes the flaky test

PR #90749, opened 2020-05-05
fix: azure disk dangling attach issue on VMSS which would cause API throttling

andyzhangx:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
fix: azure disk dangling attach issue, and this issue only happens on **VMSS**.
PR(https://github.com/kubernetes/kubernetes/pull/81266) does not convert the VMSS node name which causes error like this:
```
failed to get azure instance id for node \"k8s-agentpool1-32474172-vmss_1216\" (not a vmss instance)
```

This is rarely happen issue since dangling error only happens when the previous detach disk failed, the disk was still attached to the node while no pod on that node is using that disk.
The disk detach failed could be due to 1) VM busy 2) kube-controller-manager restart, etc. 

#### VMSS nodeName value
 - `disk.ManagedBy` value: `/subscriptions/0e46bd28-a80f-4d3a-8200-d9eb8d80cb2e/resourceGroups/kubetest-b28d2392-498e-11ea-af8c-fab712ce9f74/providers/Microsoft.Compute/virtualMachineScaleSets/k8s-agentpool-36841236-vmss/virtualMachines/k8s-agentpool-36841236-vmss_1`

 - correct `nodeName` value: `k8s-agentpool-36841236-vmss000001`

That will make dangling attach return error, and k8s volume attach/detach controller will getVmssInstance, and since the nodeName is in an incorrect format, it will always clean vmss cache if node not found, thus incur a get vmss API call storm.
https://github.com/kubernetes/kubernetes/blob/3365ed5625797cfaafbdab368e901e9f4739329a/staging/src/k8s.io/legacy-cloud-providers/azure/azure_vmss.go#L172-L174

**Which issue(s) this PR fixes**:

Fixes #90762

**Special notes for your reviewer**:
```
Events:
  Type     Reason                  Age                    From                                        Message
  ----     ------                  ----                   ----                                        -------
  Normal   Scheduled               6m29s                  default-scheduler                           Successfully assigned default/nginx-azuredisk to k8s-agentpool-17772720-vmss000000
[...]

PR #81266, opened 2019-08-11
fix: detach azure disk issue using dangling error

andyzhangx:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
fix: detach azure disk issue using dangling error

I think this PR could fix most of the disk attach issue due to detach disk not succeed(could be due to many reasons), this PR could do the recover automatically, I will cherry pick the fix to all old releases.

Actually this PR fixed two issues:
 - `disk not found` issue when detaching an azure disk
This is fixed by compare disk URI with case insensitive (`strings.EqualFold`), error logs are like following(without this PR):
```
azure_controller_standard.go:134] detach azure disk: disk  not found, diskURI: /subscriptions/xxx/resourceGroups/andy-mg1160alpha3/providers/Microsoft.Compute/disks/xxx-dynamic-pvc-41a31580-f5b9-4f08-b0ea-0adcba15b6db
```

 - azure disk could not be attached to another node since disk is still attached to the original node
This is fixed by return dangling error, success events logs are like following:
```
Events:
  Type     Reason                  Age                    From                               Message
  ----     ------                  ----                   ----                               -------
  Normal   Scheduled               <unknown>              default-scheduler                  Successfully assigned default/nginx-azuredisk to k8s-agentpool-32686255-1
  Warning  FailedAttachVolume      2m46s (x7 over 3m18s)  attachdetach-controller            AttachVolume.Attach failed for volume "pvc-41a31580-f5b9-4f08-b0ea-0adcba15b6db" : disk(/subscriptions/xxx/resourceGroups/andy-mg1160alpha3/providers/Microsoft.Compute/disks/xxx-dynamic-pvc-41a31580-f5b9-4f08-b0ea-0adcba15b6db) already attached to node(k8s-agentpool-32686255-0), could not be attached to node(k8s-agentpool-32686255-1)
  Normal   SuccessfulAttachVolume  2m4s                   attachdetach-controller            AttachVolume.Attach succeeded for volume "pvc-41a31580-f5b9-4f08-b0ea-0adcba15b6db"
  Warning  FailedMount             75s                    kubelet, k8s-agentpool-32686255-1  Unable to attach or mount volumes: unmounted volumes=[disk01], unattached volumes=[disk01 default-token-xw626]: timed out waiting for the condition
  Normal   Pulling                 59s                    kubelet, k8s-agentpool-32686255-1  Pulling image "nginx"
[...]

PR #93880, opened 2020-08-11
Wrap InstanceNotFound into ErrorNotVmssInstance

sylr:
**What type of PR is this?**

/kind bug

This is a long shot but I've troubles on Azure with a 1.13.12 cluster (yeah I know ...) which randomly fails to detach and re-attach disks from a node to another one when draining.

The current patch aims to fix the following scenario:

- 1 "full" "old" node (a VMSS instance) with several (2) pods with Azure disks PVC  / 1 "empty" "new" node (also a VMSS instance)
- "old" node is drained -> workloads are migrated to the "new" node
- 1 azure disks fails to be detached from the old node so it can't be re-attached to the new node
- the cluster autoscaler kicks in and remove the old node so the azure disk is "physically" detached from the old node
- kubernetes does not find the old node anymore and refuses to go further in the process of re-attaching the disk to the new node

In the logs I see the following:

```
E0811 09:02:57.562954 1 attacher.go:119] azureDisk - Error checking if volumes ([k8s12-EUW-test-dynamic-pvc-f79095d2-5ad4-11e9-829e-000d3a43e398]) are attached to current node ("k8s12-euw-test-agentmntr1-vmss_3"). err=not a vmss instance
E0811 09:02:57.563011 1 operation_generator.go:190] VolumesAreAttached failed for checking on node "k8s12-euw-test-agentmntr1-vmss_3" with: not a vmss instance
```

 (I've backported 05d7570 in 1.13.12 and I am running it on the cluster, see https://github.com/sylr/kubernetes/commits/v1.13.12%2Bsylr).

[I have tried a less subtle version of this patch](https://github.com/sylr/kubernetes/commit/588fa9744f17b56833cd08fa574b02264109af14) and it seems to unblock my problem as long as the drained node is actually deleted by the cluster autoscaler.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/volume/azure_dd/attacher.go
+++ b/pkg/volume/azure_dd/attacher.go
@@ -19,6 +19,7 @@ limitations under the License.
 package azure_dd
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -70,7 +71,7 @@ func (a *azureDiskAttacher) Attach(spec *volume.Spec, nodeName types.NodeName) (
 	}
 
 	lun, err := diskController.GetDiskLun(volumeSource.DiskName, volumeSource.DataDiskURI, nodeName)
-	if err == cloudprovider.InstanceNotFound {
+	if errors.Is(err, cloudprovider.InstanceNotFound) {
 		// Log error and continue with attach
 		klog.Warningf(
 			"Error checking if volume is already attached to current node (%q). Will continue and try attach anyway. err=%v",
--- a/pkg/volume/gcepd/attacher_test.go
+++ b/pkg/volume/gcepd/attacher_test.go
@@ -300,7 +300,7 @@ func TestVerifyVolumesAttached(t *testing.T) {
 			test: func(testcase *testcase) error {
 				attacher := newAttacher(testcase)
 				_, err := attacher.VolumesAreAttached([]*volume.Spec{diskASpec}, nodeName1)
-				if err != cloudprovider.InstanceNotFound {
+				if !errors.Is(err, cloudprovider.InstanceNotFound) {
 					return fmt.Errorf("expected InstanceNotFound error, but got %v", err)
 				}
 				return err
--- a/staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
+++ b/staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
@@ -543,7 +543,7 @@ func ensureNodeExistsByProviderID(ctx context.Context, instances cloudprovider.I
 		var err error
 		providerID, err = instances.InstanceID(ctx, types.NodeName(node.Name))
 		if err != nil {
-			if err == cloudprovider.InstanceNotFound {
+			if errors.Is(err, cloudprovider.InstanceNotFound) {
 				return false, nil
 			}
 			return false, err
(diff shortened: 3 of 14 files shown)


andyzhangx:
hi @sylr, thanks for the contribution.
I think use `errors.Is` could fix a few wrap error issue. While I doubt whether it could fix azure disk detach issue: `azure disks fails to be detached from the old node so it can't be re-attached to the new node`, since azure disk detach failure is usually caused by 1) azure platform(e.g. CRP, disk RP busy) issue 2) kube-controller-manager restart
So I don't think use `errors.Is` could fix the disk detach issue.

The original dangling error fix(https://github.com/kubernetes/kubernetes/commit/05d757071367c160e1c589ab80673dbfbf28770d) does not work on VMSS, I worked out a new Fix on VMSS
   - [fix: azure disk dangling attach issue on VMSS which would cause API throttling](https://github.com/kubernetes/kubernetes/pull/90749)

| k8s version | fixed version |
| ---- | ---- |
| v1.15 | no fix |
| v1.16 | 1.16.9 |
| v1.17 | 1.17.6 |
| v1.18 | 1.18.3 |
| v1.19 | 1.19.0 |

sylr:
> since azure disk detach failure is usually caused by 1) azure platform(e.g. CRP, disk RP busy) issue 2) kube-controller-manager restart

Yeah, but if you end up to delete the "old" not while kubernetes is struggling detaching the disk from it (for whatever reason) it ends up with a "not a vmss instance" that the code does not currently handle as far as I can see.

andyzhangx:
> > since azure disk detach failure is usually caused by 1) azure platform(e.g. CRP, disk RP busy) issue 2) kube-controller-manager restart
> 
> Yeah, but if you end up to delete the "old" not while kubernetes is struggling detaching the disk from it (for whatever reason) it ends up with a "not a vmss instance" that the code does not currently handle as far as I can see.

`not a vmss instance` issue is fixed by https://github.com/kubernetes/kubernetes/pull/90749, introduced by https://github.com/kubernetes/kubernetes/pull/81266

andyzhangx:
@sylr could you cherry-pick #90749 to your branch, and check whether that PR could fix your issue? thanks.

sylr:
> The original dangling error fix(05d7570) does not work on VMSS, I worked out a new Fix on VMSS

The only commit in #90749 is 05d7570. Could you point me to the patch that is supposed to work on VMSS ?

andyzhangx:
> > The original dangling error fix([05d7570](https://github.com/kubernetes/kubernetes/commit/05d757071367c160e1c589ab80673dbfbf28770d)) does not work on VMSS, I worked out a new Fix on VMSS
> 
> The only commit in #90749 is [05d7570](https://github.com/kubernetes/kubernetes/commit/05d757071367c160e1c589ab80673dbfbf28770d). Could you point me to the patch that is supposed to work on VMSS ?

@sylr these are the 2 PRs that brings in leveraging azure disk dangling error to fix the issue that disk could not be detached forever:
https://github.com/kubernetes/kubernetes/pull/81266
https://github.com/kubernetes/kubernetes/pull/90749

The azure disk dangling error feature is already available on 1.16.10, 1.17.6, 1.18.3, 1.19.0

PR #93888, opened 2020-08-11
Add sigs for root folders

dims:
As @smarterclayton mentioned in last sig-arch call, we have holes where none of the sigs claim jurisdiction (if i understood correctly). Adding a few aliases and assigning the root folders

The current thing that came up was discussion around addon-manager here : https://github.com/kubernetes/kubernetes/pull/92978

**What type of PR is this?**

/kind cleanup

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/.github/OWNERS
+++ b/.github/OWNERS
@@ -8,9 +8,7 @@ reviewers:
   - nikhita
   - parispittman
 approvers:
-  - castrojo
-  - cblecker
-  - nikhita
+  - sig-contribex-approvers
   - parispittman
 emeritus_approvers:
   - Phillels
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,4 +1,9 @@
 aliases:
+  sig-architecture-approvers:
+    - dims
+    - derekwaynecarr
+    - johnbelamaric
+
   # sig-auth subproject aliases
   sig-auth-audit-approvers:
     - sttts
@@ -453,6 +458,20 @@ aliases:
   # sig-cluster-lifecycle-api-reviewers:
   #   -
   #   -
+
+  sig-contribex-approvers:
+    - castrojo
+    - mrbobbytables
+    - cblecker
+    - nikhita
+
+  sig-docs-approvers:
+    - jimangel
+    - kbarnard10
+    - zacharysarah
+    - kbhawkey
+    - onlydole
+    - sftim
 
   sig-node-api-reviewers:
     - dchen1107
--- a/docs/OWNERS
+++ b/docs/OWNERS
@@ -5,6 +5,7 @@ reviewers:
   - smarterclayton
   - thockin
 approvers:
+  - sig-docs-approvers
   - brendandburns
   - smarterclayton
   - thockin
--- a/third_party/OWNERS
+++ b/third_party/OWNERS
@@ -5,6 +5,7 @@ reviewers:
   - smarterclayton
   - thockin
 approvers:
+  - dep-approvers
   - lavalamp
   - smarterclayton
   - thockin
--- a/translations/OWNERS
+++ b/translations/OWNERS
@@ -3,4 +3,5 @@
 reviewers:
   - brendandburns
 approvers:
+  - sig-cli-maintainers
   - brendandburns
--- a/logo/OWNERS
+++ b/logo/OWNERS
@@ -3,4 +3,5 @@
 reviewers:
   - thockin
 approvers:
+  - sig-architecture-approvers
   - thockin
--- a/staging/src/k8s.io/kubectl/pkg/generated/bindata.go
+++ b/staging/src/k8s.io/kubectl/pkg/generated/bindata.go
@@ -84,6 +84,7 @@ var _translationsOwners = []byte(`# See the OWNERS docs at https://go.k8s.io/own
 reviewers:
   - brendandburns
 approvers:
+  - sig-cli-maintainers
   - brendandburns
 `)
 


docs/OWNERS
```diff
@@ -5,7 +5,8 @@ reviewers:
   - smarterclayton
   - thockin
 approvers:
+  - sig-architecture-approvers
```
liggitt, line 8: this one seems more like docs approvers?

cluster/addons/OWNERS
```diff
@@ -1,5 +1,8 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
+approvers:
+- sig-architecture-approvers
```
liggitt, line 4: I would have expected this to fall up to clusters/OWNERS

third_party/OWNERS
```diff
@@ -5,6 +5,7 @@ reviewers:
   - smarterclayton
   - thockin
 approvers:
+  - sig-architecture-approvers
```
liggitt, line 8: this is probably closer to dep-approvers

translations/OWNERS
```diff
@@ -3,4 +3,5 @@
 reviewers:
   - brendandburns
 approvers:
+  - sig-architecture-approvers
```
liggitt, line 6: I think this is only used by kubectl, so maybe sig-cli instead?

.github/OWNERS
```diff
@@ -8,6 +8,7 @@ reviewers:
   - nikhita
   - parispittman
 approvers:
+  - sig-architecture-approvers
```
liggitt, line 11: do we have a contribx alias? I think they mostly run this folder

nikhita: Yes, contribex owns this directory. I don't think we really need a contribex alias since it's pretty much just this folder, but if it helps identify the owning SIG, we could use one. :woman_shrugging:

docs/OWNERS
dims: yep.

cluster/addons/OWNERS
dims: ack moving this one level up

third_party/OWNERS
dims: good point

translations/OWNERS
dims: Done

.github/OWNERS
dims: Done!

.github/OWNERS
```diff
@@ -8,6 +8,7 @@ reviewers:
   - nikhita
   - parispittman
 approvers:
+  - sig-contribex-approvers
   - castrojo
```
nikhita, line 12: IMO if we are adding an alias, we can remove explicit approvers from here.

cc @castrojo @mrbobbytables @cblecker

OWNERS_ALIASES
```diff
@@ -454,6 +459,15 @@ aliases:
   #   -
   #   -
 
+  sig-contribex-approvers:
+    - castrojo
+    - mrbobbytables
```
nikhita, line 464: Can you also add me and @cblecker here?

.github/OWNERS
dims: Done

dims:
cc @castrojo @cblecker @mrbobbytables @jimangel @kbarnard10 @zacharysarah please see the new alias(es) and owner updates.

OWNERS_ALIASES
nikhita: Missed adding me :stuck_out_tongue: 

Want to ensure I am notified in case contribex gets pulled into something :)

OWNERS_ALIASES
```diff
@@ -454,6 +459,16 @@ aliases:
...
+    - cblecker
+
+  sig-docs-approvers:
```
liggitt, line 467: probably include TLs as well? https://github.com/kubernetes/community/tree/master/sig-docs#technical-leads

cluster/OWNERS
```diff
@@ -8,6 +8,7 @@ reviewers:
   - mikedanese
   - spiffxp
 approvers:
+  - sig-architecture-approvers
```
liggitt, line 11: this is the only one I'm unsure of... is arch the right sig for this folder?

OWNERS_ALIASES
zacharysarah: Came here to say what @liggitt said. 👍

OWNERS_ALIASES
dims: whooops!

OWNERS_ALIASES
dims: Done!

nikhita:
LGTM for SIG Contribex.

sftim:
LGTM for SIG Docs

OWNERS_ALIASES
onlydole: Thank you, @zacharysarah and @dims!

onlydole (approved):
LGTM - thank you! 

/lgtm

cluster/OWNERS
BenTheElder: cloud provider or testing fits for the actual cluster bits (formerly cluster lifecycle...).

cluster/kubemark is scalability, cluster/images/ should probably be moved and is a mix of testing / release / ...

derekwaynecarr: I do not have a strong bias one way or the other.  I am happy to have approve here, I have history on kube-up, most of this content at this level (not in a subtree) is under maintenance.  this change basically gives me and john new privileges.  dims appear to already have the privilege.

derekwaynecarr:
I am fine with what is presented here.  It slightly elevates @johnbelamaric and my own privilege at /cluster, but I think its not a huge deal, and would help.

/lgtm

cluster/OWNERS
dims: @liggitt used to be cluster-lifecycle a while ago. we need to split things up like Ben said and move bits and pieces off to other sigs. having additional hands to review/approve would be useful

liggitt: > we need to split things up like Ben said and move bits and pieces off to other sigs

that's fine, but sig-arch seems strange for this one. do we expect all future members of sig-architecture-approvers to have context on this dir?

dims: agree it's strange.

liggitt: if you want to drop this line, the rest of the PR looks helpful, we can iterate on this bit and hopefully target it where it makes more sense

BenTheElder:
https://github.com/kubernetes/kubernetes/pull/93888#issuecomment-672240883

currently yes, but it also in effect states that sig-arch is the owner for this directory, which doesn't seem like an accurate statement.

cluster/OWNERS
BenTheElder: this still appears out of place to me.

dims: Done!

dims:
@BenTheElder @liggitt Done! remove the line from cluster/OWNERS

liggitt:
/lgtm
/approve
/retest

It seems like I've seen a lot of bazel/debian fetch errors recently... @justaugustus / @BenTheElder any known cause for that?

BenTheElder:
it's failing to download the debian iptables image under bazel. I don't know why.

I know @spiffxp recently restricted the CPU of this job more ...

BenTheElder:
https://github.com/kubernetes/kubernetes/pull/93981 see also upgrading the rules responsible for the image pull to current (~2  years out of date)

Issue #93276, opened 2020-07-21
Presubmit tests for Windows in 1.20 and release blocking tests


jsturtevant:
**What would you like to be added**:
Presubmit tests for windows  in 1.20 and to see windows tests in release blocking [dashboard like 1.19](https://k8s-testgrid.appspot.com/sig-release-1.19-blocking) for 1.20

**Why is this needed**:
The current process for testing changes against Windows and identifying issues is reactive which causes lots of work after the fact.  For 1.19 we had at least 2 PR's that introduced breaking changes after code freeze. I believe this has happened in prior releases as well. 

Two bugs that caused windows nodes to not work recently:

- https://github.com/kubernetes/kubernetes/pull/80917
- https://github.com/kubernetes/kubernetes/pull/90948

The result was 1.19rc.1 did not work on windows and many tests built of main branch for [sig-windows](https://testgrid.k8s.io/sig-windows) tests failed for over a week.  Both issues needed to be fixed via a last minute pr for 1.19rc2: https://github.com/kubernetes/kubernetes/pull/93263

## Next Steps
What are the requirements for adding a presubmitts from the tests?  Is this documented somewhere (I did some searching and wasn't able to find anything)?

A few thoughts to start:

- [ ] Need to show stability in tests (can this be a subset of tests to start?)
- [ ] Speed of tests should be consistently under 2hrs (based on https://k8s-testgrid.appspot.com/sig-release-master-blocking#Conformance%20-%20GCE%20-%20master)
- [ ] need sig-windows support to drive out flakes that might be introduced
- [ ] can there be a subset of tests that can be run to atleast validate windows nodes come up and simple pods are scheduled to start
- [ ] do the blocking PR's need to be run on GCE or can we use tests from https://testgrid.k8s.io/sig-windows#aks-engine-azure-windows-master-staging?
- [ ] Image builder prs for windows need to be merged: https://github.com/kubernetes/kubernetes/pull/77398 and https://github.com/kubernetes/kubernetes/pull/77269 this will resolve issues when the images are changed and not require manual builds and pushes are required now.

My goal here is to learn what is required and start the conversation.  I am happy to help drive the conversation in the correct places.

/sig windows
/sig testing 
/sig release

liggitt:
cc @BenTheElder @kubernetes/sig-testing 

> Speed of tests should be consistently under 2hrs

[...]

PR #77398, opened 2019-05-03
test images: Adds Windows Container images support (part 2)

claudiubelu:
**What type of PR is this?**

/kind feature

/sig testing
/sig windows

**What this PR does / why we need it**:

Some images require a different ``Dockerfile`` for Windows images, since they have different ways of  installing dependencies. Because of this, if a image needs to be built for Windows, it will first check for a ``Dockerfile_windows`` file instead of the default one. If there isn't one, it means that the same ``Dockerfile``can be used for both Windows and Linux.

All Windows images will be based on the image "mcr.microsoft.com/windows/servercore:ltsc2019". There are a couple of features that are needed from this image, especially powershell.

Adds a README explaining the image building process, including the Windows Container image building process.

**Which issue(s) this PR fixes**:

Fixes #77268
Depends On: #76838

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Kubernetes E2E test image manifest lists now contain Windows images.
```

PR #77269, opened 2019-04-30
test images: Adds Windows Container images support (part 3)

claudiubelu:
**What type of PR is this?**

/kind feature

/sig testing
/sig windows

**What this PR does / why we need it**:

Adds the ``httpd``, ``nginx`` images that are used in tests.

Two different versions of nginx have to build, and thus, the have different folders. An ``ALIAS`` file was added to ``nginx-new`` in order to keep the same image name.

**Which issue(s) this PR fixes**:

Partially Fixes #77268 

**Special notes for your reviewer**:

This PR is supposed to be on top of https://github.com/kubernetes/kubernetes/pull/76838 (part 1 & 2) but it was separated due to size.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #89425, opened 2020-03-24
test images: uses nanoserver

claudiubelu:
**What type of PR is this?**

/kind feature

/sig testing
/sig windows

**What this PR does / why we need it**:

Using Windows nanoserver container images as a base instead of the current Windows servercore image will reduce the image size by about ~10x.

However, the nanoserver image lacks several things we need:
- netapi32.dll
- powershell
- certain powershell commands
- chocolatey cannot be used

When building the nanoserver images, we are going to use a Windows servercore helper, in which we are going to install the necessary dependencies, and then copy them over to our nanoserver image, including necessary DLLs.

Other notable changes include:
- switch from wget to curl (wget was a powershell alias).
- implement in code getting the DNS suffix list and DNS server list.
- reimplement getting file permissions for mounttest.

**Which issue(s) this PR fixes**:

Partially Fixes #77268

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #93889, opened 2020-08-11
Test images: switch to buildx

claudiubelu:
**What type of PR is this?**

/kind feature

/sig windows
/sig testing

Currently, some of the E2E test images have Windows support, and one of the goals is for most of them to have Windows support (#77398, #77269). For that, the Image Promoter is currently building those Windows container images using a few Windows Server nodes (for 1809, 1903, 1909) with Remote Docker enabled which are hosted on an azure subscription dedicated for CNCF.

With this PR, that dependency is reduced to a single Windows Server node, which will only need to build a helper image. One additional benefit to this is that adding new supported Windows OS versions to the E2E test images manifest lists becomes a lot easier (we wouldn't have to create a new Windows Server node that matches that new OS version, assign DNS name, update certificates, etc.), and it also becomes easier for other people to build their own E2E windows test images.

Alternatively, we can just host that helper image in a registry like ``e2eteam`` and just use that. That will remove the Windows Server nodes dependency entirely, and almost anyone could build their Windows E2E test images without requiring any Windows nodes (except in air-gapped environments). The most significant bits that the image's ``Dockerfile`` does is that it prepares busybox and prepares powershell for a nanoserver image. Technically, those two bits can easily be saved as separate, versioned images, that won't have any reason to change, for example: ``e2eteam/nanoserver-powershell:1.0``, ``e2eteam/busybox:1.0``.

However, there is a small concern when it comes to the build time: Windows servercore images are very large (for example, ``mcr.microsoft.com/windows/servercore:ltsc2019`` is 4.99GB uncompressed, and about ~2 GB compressed - those images are already cached on the Windows Server builder nodes, so this isn't an issue there), and we currently support 1809, 1903, and 1909, and ``docker buildx`` seems to prefer to repull them everytime. This can lead to build times that are too big.

We have changed the base image to nanoserver (uncompressed size: 250MB), but some images still require some DLLs or some other dependencies that can be fetched from a servercore image. So, we can resolve this issue in 2 ways:

-  we can define a separate job that would build a scratch ``windows-servercore-cache`` image weekly / monthly, and then we can just get those dependencies from this cache, which will be very small. This would be preferred, as the Windows images update periodically, and those dependencies could be updated as well.
- host the dependencies somewhere else, like ``github.com/kubernetes-sigs/windows-testing``, and just get them from there when building the images.

Related: #93276

Check only the last commit, since that's the only one that contains the ``docker buildx`` implementation. Other commits that adds Windows support to other test images are included in this PR for a general view on how their Dockerfiles change for ``docker buildx``,

```release-note
NONE
```

```docs

```


BenTheElder:
there's a lot to read here and this PR doesn't appear ready yet but, FWIW I am in favor of switching to buildx.

CC @kubernetes/release-engineering @

claudiubelu:
> 
> 
> there's a lot to read here and this PR doesn't appear ready yet but, FWIW I am in favor of switching to buildx.
> 
> CC @kubernetes/release-engineering

Indeed, there are few things left, and there is quite a bit to read, but we would really like to know with which proposal to go forth with. Those last implementation details depends on this, after all.

BenTheElder:
totally -- CCed release-engineering to attempt to get some addition input regarding building this.
will drop in some SIG channels as well.

justaugustus:
/assign
not sure that I'll have time for this this week though.

claudiubelu:
I've reduced this PR to just the nanoserver commit (#89425) and the buildx implementation itself. For the full picture view, including the other test images that will have Windows support, see: https://github.com/claudiubelu/kubernetes/commit/25c880934e31f5e9722a06d3c90ca37dd49d8212

The current proposal doesn't need *any* Windows build nodes. However, it uses the helper images ``e2eteam/busybox-helper:1.29.0`` and ``e2eteam/powershell-helper:6.2.0`` which have been built on a Windows node (see their Dockerfiles in the last commit). Those images won't change. If they do, the version will be updated too. Is that acceptable to you?

The last pending question is the ``windows-servercore-cache`` image. The proposal was to only build this weekly / monthly, as it is the most time consuming out of all, and it only has to cache some necessary DLLs for nanoserver.

chewong:
Thanks for providing valuable feedback in today's sig-testing meeting @BenTheElder @spiffxp. Yt looks like we have reached a consensus on using docker buildx to build Windows test images since it allows the community to build them locally without the need of Windows nodes.

To summarize, we addressed the following concerns in the meeting and proposed a solution for each of them as shown below:

| Concern                                                                                                                                                | Proposed Solution                                                                                                                              |
|--------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------|
|  We could face the rate-limiting issue if we host our helper / cache images (`busybox-helper`, `powershell-helper`, and `windows-servercore-cache`) in Dockerhub. | Push them to `gcr.io/k8s-staging-e2e-test-images` without promoting them to `k8s.gcr.io/e2e-test-images`. The image size should be negligible (~5MB).                                 |
| We will still require a Windows node to build `busybox-helper` and `powershell-helper` helper images since GCB does not support building Windows images.      | Given that we would rarely update `busybox-helper` and `powershell-helper` AND no one would use those images unless they are building Windows test images, we agreed that it's a fair compromise to build them in a remote Windows node, push them to a registry and never touch it again in exchange for the ability to leverage docker buildx. |
| The frequency at which `windows-servercore-cache` is being built.                                                                                      | Create a monthly prow job that builds & pushes an updated version of the cache image to match the release cadence of Microsoft's base images.                                    |


Please let us know if you have any additional comments/concerns and we'd love to land this PR (and potentially along with 
https://github.com/kubernetes/kubernetes/pull/77398 and https://github.com/kubernetes/kubernetes/pull/77269 after this is merged) by 1.20 so we can start automating the test image building process for Windows.

/priority important-soon

BenTheElder (commented):
with time this build has gotten a bit convoluted but I'm pretty confident in the linux / buildx changes now
I will defer to someone else for the powershell bits, I'm not fluent there :+)
/approve

test/images/windows/powershell-helper/VERSION
```diff
@@ -0,0 +1 @@
+6.2.0
```
marosset, line 1: Should we target 6.2.7
https://github.com/PowerShell/PowerShell/releases/tag/v6.2.7

6.2.0 is from March 2019.

marosset: > I will defer to someone else for the powershell bits, I'm not fluent there :+)

Other than this the powershell parts look good to me.

claudiubelu:
/retest

Image building logs with these changes + https://github.com/kubernetes/kubernetes/pull/95032 : https://paste.ubuntu.com/p/mB5GZp4K2y/

Additionally, we've already added the windows-servercore-cache job: https://github.com/kubernetes/test-infra/pull/19494

test/images/windows/powershell-helper/VERSION
claudiubelu: Done

BenTheElder:
looks like the builds are failing:
https://kubernetes.slack.com/archives/C09QZ4DQB/p1604365991474200
https://paste.centos.org/view/raw/7d1f237c
I think https://github.com/kubernetes/kubernetes/blob/43edb4063262751bbb6ba18daf3fe9582ae3b2f9/test/images/cloudbuild.yaml#L25 is failing, perhaps `DOCKER_CLI_EXPERIMENTAL=enabled` is not plumbing through.

claudiubelu:
> 
> 
> looks like the builds are failing:
> https://kubernetes.slack.com/archives/C09QZ4DQB/p1604365991474200
> https://paste.centos.org/view/raw/7d1f237c
> I think
> 
> https://github.com/kubernetes/kubernetes/blob/43edb4063262751bbb6ba18daf3fe9582ae3b2f9/test/images/cloudbuild.yaml#L25
> is failing, perhaps `DOCKER_CLI_EXPERIMENTAL=enabled` is not plumbing through.

Correct. I've been asking for some ideas on slack about this. I've been looking into this, I've been comparing the pause image build job and the e2e test image build job since they both use ``docker buildx``. The build pause image works fine. There are no significant differences between their ``cloudbuild.yaml`` files or prow jobs.

PR #91775, opened 2020-06-04
VolumeBinding: Skip/fail fast in PreFilter phase and improve error reporting 

cofyc:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:

- pod volumes information is checked only once in PreFilter phase
- return UnscheduableAndUnresolvable error if some immediate PVCs are not bound 

**Which issue(s) this PR fixes**:

Fixes #91755 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #93892, opened 2020-08-11
"unbound immediate PersistentVolumeClaims" should be UnschedulableAnd…

ahg-g:
**What type of PR is this?**

/kind bug

"unbound immediate PersistentVolumeClaims" should be UnschedulableAndUnresolvable error
    
This was fixed in 1.19 by refactoring that part into PreFilter in https://github.com/kubernetes/kubernetes/pull/91775

This is a bug because it is currently causing incorrect reporting in scheduling_attempts metric (internal error vs unschedulable).

Fixes ##91755

```release-note
"unbound immediate PersistentVolumeClaims" causes UnschedulableAndUnresolvable status rather than an Error in the scheduler. 
```

### DIFF
--- a/pkg/controller/volume/scheduling/scheduler_binder_test.go
+++ b/pkg/controller/volume/scheduling/scheduler_binder_test.go
@@ -809,9 +809,9 @@ func TestFindPodVolumesWithoutProvisioning(t *testing.T) {
 			shouldFail: true,
 		},
 		"prebound-pvc": {
-			podPVCs:    []*v1.PersistentVolumeClaim{preboundPVC},
-			pvs:        []*v1.PersistentVolume{pvNode1aBound},
-			shouldFail: true,
+			podPVCs: []*v1.PersistentVolumeClaim{preboundPVC},
+			pvs:     []*v1.PersistentVolume{pvNode1aBound},
+			reasons: ConflictReasons{ErrUnboundImmediatePVC},
 		},
 		"unbound-pvc,pv-same-node": {
 			podPVCs:          []*v1.PersistentVolumeClaim{unboundPVC},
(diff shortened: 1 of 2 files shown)


pkg/controller/volume/scheduling/scheduler_binder_test.go
```diff
@@ -809,9 +809,9 @@ func TestFindPodVolumesWithoutProvisioning(t *testing.T) {
 			shouldFail: true,
 		},
 		"prebound-pvc": {
-			podPVCs:    []*v1.PersistentVolumeClaim{preboundPVC},
```
ahg-g, line 812: @cofyc can you double check if the change to this test case is expected?

pkg/controller/volume/scheduling/scheduler_binder_test.go
```diff
@@ -850,9 +850,9 @@ func TestFindPodVolumesWithoutProvisioning(t *testing.T) {
 			reasons: ConflictReasons{ErrReasonBindConflict},
 		},
 		"one-prebound,one-unbound": {
```
ahg-g, line 852: ditto,@cofyc can you double check if the change to this test case is expected?

alculquicondor:
lgtm from scheduling PoV

pkg/controller/volume/scheduling/scheduler_binder_test.go
cofyc: yes

pkg/controller/volume/scheduling/scheduler_binder_test.go
cofyc: yes

msau42:
/approve

I think all cherry-picks need a release note.

ahg-g:
/sig-storage
/priority important-soon
@kubernetes/release-managers

ahg-g:
friendly ping @kubernetes/release-managers

hasheddan (approved):
/lgtm

PR #93896, opened 2020-08-11
Improve running time of TestSchedulerWithVolumeBinding

alculquicondor:
Only wait for finished binding or error, but not both

**What type of PR is this?**

/kind failing-test

The scheduler would produce either, but not both. No need for busy waiting.

The running time of the test goes from 17s down to 1s.

Somewhat related to investigations in #93890

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/scheduler_test.go
+++ b/pkg/scheduler/scheduler_test.go
@@ -953,28 +953,26 @@ func TestSchedulerWithVolumeBinding(t *testing.T) {
 				t.Fatalf("scheduling timeout after %v", wait.ForeverTestTimeout)
 			}
 			stopFunc()
-			// Wait for scheduling to return an error
+			// Wait for scheduling to return an error or succeed binding.
+			var (
+				gotErr  error
+				gotBind *v1.Binding
+			)
 			select {
-			case err := <-errChan:
-				if item.expectError == nil || !reflect.DeepEqual(item.expectError.Error(), err.Error()) {
-					t.Errorf("err \nWANT=%+v,\nGOT=%+v", item.expectError, err)
-				}
+			case gotErr = <-errChan:
+			case gotBind = <-bindingChan:
 			case <-time.After(chanTimeout):
-				if item.expectError != nil {
-					t.Errorf("did not receive error after %v", chanTimeout)
-				}
+				t.Fatalf("did not receive pod binding or error after %v", chanTimeout)
 			}
-
-			// Wait for pod to succeed binding
-			select {
-			case b := <-bindingChan:
-				if !reflect.DeepEqual(item.expectPodBind, b) {
-					t.Errorf("err \nWANT=%+v,\nGOT=%+v", item.expectPodBind, b)
+			if item.expectError != nil {
+				if gotErr == nil || item.expectError.Error() != gotErr.Error() {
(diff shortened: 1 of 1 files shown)


alculquicondor:
cc @cofyc 

/assign @Huang-Wei 

Thanks @liggitt for the suggestion

pkg/scheduler/scheduler_test.go
```diff
@@ -953,27 +953,19 @@ func TestSchedulerWithVolumeBinding(t *testing.T) {
...
+			// Wait for scheduling to return an error or succeed binding.
 			select {
 			case err := <-errChan:
 				if item.expectError == nil || !reflect.DeepEqual(item.expectError.Error(), err.Error()) {
 					t.Errorf("err \nWANT=%+v,\nGOT=%+v", item.expectError, err)
 				}
-			case <-time.After(chanTimeout):
-				if item.expectError != nil {
-					t.Errorf("did not receive error after %v", chanTimeout)
-				}
```
liggitt, line 965: move this to the `<-time.After(chanTimeout)` case

pkg/scheduler/scheduler_test.go
```diff
@@ -953,27 +953,19 @@ func TestSchedulerWithVolumeBinding(t *testing.T) {
...
-			// Wait for pod to succeed binding
-			select {
 			case b := <-bindingChan:
 				if !reflect.DeepEqual(item.expectPodBind, b) {
 					t.Errorf("err \nWANT=%+v,\nGOT=%+v", item.expectPodBind, b)
 				}
 			case <-time.After(chanTimeout):
 				if item.expectPodBind != nil {
-					t.Errorf("did not receive pod binding after %v", chanTimeout)
+					t.Errorf("did not receive pod binding or error after %v", chanTimeout)
```
Huang-Wei, line 968: We should also error out below `if item.expectError != nil`.

pkg/scheduler/scheduler_test.go
alculquicondor: Better done after the select.

pkg/scheduler/scheduler_test.go
alculquicondor: we don't actually need to check anything in the case of timeout. It's an error regardless.

alculquicondor:
@liggit do you want to include it in the milestone?

liggitt:
at this point, I'm only including flake fixes in the milestone, and am leaving test runtime improvements for when master reopens

pkg/scheduler/scheduler_test.go
```diff
@@ -953,28 +953,26 @@ func TestSchedulerWithVolumeBinding(t *testing.T) {
...
+				gotBind *v1.Binding
+			)
 			select {
-			case err := <-errChan:
-				if item.expectError == nil || !reflect.DeepEqual(item.expectError.Error(), err.Error()) {
-					t.Errorf("err \nWANT=%+v,\nGOT=%+v", item.expectError, err)
-				}
+			case gotErr = <-errChan:
+			case gotBind = <-bindingChan:
 			case <-time.After(chanTimeout):
```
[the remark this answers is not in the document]
krmayankk: sounds good, probably its just me then.

PR #93898, opened 2020-08-11
Always update pod IP in case of hostNetworking

### LINKED ISSUE #93897: Pod IP in hostNetwork mode does not update when node IP changes
**What happened**: 

When PODs are running in hostNetworking mode they use Kubernetes node IP address. I hit issue when actual node IP changes, but some of the pods have never reflected the change. See screenshot bellow:

image

The workaround is to restart kubelet. It seems like there is condition which does not update pod IP in hostNetworking mode during regular generateAPIPodStatus, if Pod IP is already present https://github.com/volterraedge/kubernetes/blob/master/pkg/kubelet/kubelet_pods.go#L1531

I tried to remove `&& s.PodIP == ""` and it helped. I'm going to submit PR with fix.

I expect IP change on on pod when node ip itself is updated.

Go and change IP address on the node.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.17.9
- Cloud provider or hardware configuration: onprem
- OS (e.g: `cat /etc/os-release`): centos 7.2004.1
- Kernel (e.g. `uname -a`): 
- Install tools:
- Network plugin and version (if this is a network-related bug): It uses hostNetworking

pupapaik:
/kind bug

**What type of PR is this?** 

It is bug fix.

It is need to fix bug described bellow.

Fixes https://github.com/kubernetes/kubernetes/issues/93897

I tested this on my setup where I was changing IP via DHCP server leases. It works fine now without any issues.

NONE

```release-note
NONE
```

### DIFF
--- a/pkg/kubelet/kubelet_pods.go
+++ b/pkg/kubelet/kubelet_pods.go
@@ -1514,7 +1514,7 @@ func (kl *Kubelet) generateAPIPodStatus(pod *v1.Pod, podStatus *kubecontainer.Po
 			klog.V(4).InfoS("Cannot get host IPs", "err", err)
 		} else {
 			s.HostIP = hostIPs[0].String()
-			if kubecontainer.IsHostNetworkPod(pod) && s.PodIP == "" {
+			if kubecontainer.IsHostNetworkPod(pod) && s.PodIP != s.HostIP {
 				s.PodIP = hostIPs[0].String()
 				s.PodIPs = []v1.PodIP{{IP: s.PodIP}}
 				if len(hostIPs) == 2 {


pupapaik:
I signed it

pupapaik:
@mtaufen and @Random-Liu can you please review?

pkg/kubelet/kubelet_pods.go
```diff
@@ -1528,7 +1528,7 @@ func (kl *Kubelet) generateAPIPodStatus(pod *v1.Pod, podStatus *kubecontainer.Po
 			klog.V(4).Infof("Cannot get host IP: %v", err)
 		} else {
 			s.HostIP = hostIP.String()
-			if kubecontainer.IsHostNetworkPod(pod) && s.PodIP == "" {
+			if kubecontainer.IsHostNetworkPod(pod) {
```
knight42, line 1531: I guess the following is better as we don't want to update the Pod's IP even if it is not changed.

```suggestion
			if kubecontainer.IsHostNetworkPod(pod) && s.HostIP != s.PodIP {
```

pupapaik: Make sense. Thanks a lot.

aojea:
> /lgtm
> /assign @vishh

hmm my apologies, I thought that this was simpler and wrongly reviewed this.
It seems that you can not change the node IP without restarting the kubelet ... the code history show a lot of changes and this has to be reviewed from someone with more experience
/lgtm cancel

thockin:
I don't know how SIG-Node thinks about node IP changing.  

Tell me if I am getting this right - if the node IP changes, you restart Kubelet and it seems to work, except this.  Is that right?

So a) Is that REALLY all that needs fixing?  What about `node.addresses` and `pod.status.hostIP`? b) Does sig-node want to "really" support this (e.g. at runtime without external intervention)? c) There should be an e2e test?

I don't know what else would need to be touched to make this totally safe, but I am skeptical that this is the only thing.

@dchen1107 @derekwaynecarr @sjenning

pupapaik:
@thockin thanks a lot for review. I spent few days on testing and observing IP address changes on hw boxes. It is valid in cases like LTE interface, where public IP from ISP is changing every day.

A) Yes, after I applied this change my node IP changed several times in last week and no issues observed. Both `node.addresses` and `pod.status.hostIP` got changed.
B) I think it is valid use case in the Edge environments.
C) Do you want me to do e2e test?

Do you think it can break some other place? Because today it seems like you have to restart kubelet to make it work, so if there is not any regression, then it cannot be worse?

pupapaik:
Any update please?

sjenning:
For me, the main question here is does the platform expect the pod IP to be immutable?

@pupapaik am I understanding correctly that the pod IP changes on kubelet restart if the host IP has changed?

pupapaik:
@sjenning correct, if you do kubelet restart, then pod IP changes to current node ip. It is not immutable field.

pupapaik:
I feel this change does not bring any regression issue. Can you guys please consider to merge it?

sjenning:
@pupapaik Sorry for the delay.  After thinking about this, I think it is the right thing.  Having the pod status having the wrong/unusable IP is not desirable.  If you could rebase, I'll approve.

aojea:
We should add at least an unit test to check that we don't regress,

pupapaik:
> @pupapaik Sorry for the delay. After thinking about this, I think it is the right thing. Having the pod status having the wrong/unusable IP is not desirable. If you could rebase, I'll approve.

I rebased. Can you please review again? Thanks

ehashman:
Per https://github.com/kubernetes/kubernetes/pull/93898#issuecomment-713317459 it would be great to add a unit test to verify this behaviour and protect against regressions.

pupapaik:
ok, thanks. I will try to find time this week and push it.

ehashman:
/hold

while waiting for tests

pupapaik:
@danwinship I checked your commit https://github.com/kubernetes/kubernetes/commit/971477d9b5cc4bf5ae62abe3bbc46e534f481e1b and what is not clear to me is why CRI  PodIPs should override Node Addresses? Can you help me to understand how can I identify if it is CRI PoIPs and I should not change PodIP to nodeIP? Because without CRI I actually wants to update pod ip from node, otherwise it is stucked.

danwinship:
> @danwinship I checked your commit [971477d](https://github.com/kubernetes/kubernetes/commit/971477d9b5cc4bf5ae62abe3bbc46e534f481e1b) and what is not clear to me is why CRI PodIPs should override Node Addresses? Can you help me to understand how can I identify if it is CRI PoIPs and I should not change PodIP to nodeIP? Because without CRI I actually want to update pod ip from node, otherwise it is stucked.

sorry, belatedly noticing this... That unit test is just testing the behavior that was already there for single-stack; if CRI asserts that a hostNetwork pod has a particular IP, then kubelet uses that IP. If it does not, then kubelet assigns the node IP as the pod IP.

After pod creation time kubelet doesn't keep track of whether it got the pod IP from the runtime or not because it never previously needed to care. I suppose if you care now, you'd have to query CRI explicitly.

ehashman:
/sig network

still waiting on a unit test

khenidak:
We don't really say if `PodIP/s` are immutable from an api stand point. However, we have never tested DNS/endpoint controller/endpointSlice controller against endpoint changes. AFIAK there is not a single test that validate the correctness of these controller against changing PodIP. That leaves us with two path forward:

1. treat node IP changes as described here by commenters. e.g., restart kubelet and pods.
2. Add enough unit test in these areas to ensure that we don't generate regression.

danwinship:
> We don't really say if `PodIP/s` are immutable from an api stand point.

Are you sure? I mean, definitely it's allowed to change from `""` to not-`""` some time after the Pod is created, but is it really allowed to change again after that? (I suppose kubelet doesn't care much, and kube-proxy only looks at Endpoints, not Pods, so maybe nothing in core cares, but do any network plugins deal with spontaneous PodIP changes?)

Though I suspect most network plugins / kube-proxy backends / etc don't deal well with node IPs changing either...

khenidak:
@danwinship PodIPs are mutable. Here is why: Assume PodX assigned to nodeN. PodX starts on NodeN we get PodIPs and all well. Node restarts. Node needs to start every assigned pod from scratch (i.e., the entire sandbox creation dance). The entire representation of PodX on NodeN (i.e, sandbox Id, PodIPs, etc) changed but the pod object didn't change (recreated). Hence the mutability. Same scenario if pod was forcibly delete on the node via let us say containerd or `kill` command.

We should be able to merge this once it passes tests and gets reviewed.

smarterclayton:
We never fully specified the behavior of CNI plugins - whether they had to guarantee the pod IP remains the same when various bits of state were lost.

@knobunc tested this and we confirmed a number of plugins in the ecosystem do not preserve pod IP after reset of the node and loss of the container cache.

At this point, while preservation of IP is very desirable for consumers, without a strong statement in the early days (which we didn’t make or test) the current state is “podIP can change”.

At a minimum, other clients should be aware that any podIP can change at any time and we should test this.  We *may* wish to offer guidance and tests for CNI plugins to encourage certain behaviors.  But we can’t assume they are immutable in practice nor enforce that without breaking existing plugins.

smarterclayton:
(note i’m making a general statement about podIP changing, pod IP on host pods is one of the most likely to change, but any pods IP can change over its lifetime)

aojea:
Main concerns is because it changes the current behavior, and since is not easy to predict all the consequences of this change, but we can hide it behind a FeatureGate, so we can have an easy roll back if the result is not the expected.

PR #93906, opened 2020-08-11
Deflake TestServiceRegistryUpdateDryRun test

### LINKED ISSUE #93899: [sig-network] TestServiceRegistryUpdateDryRun flakes with "already allocated" error

TestServiceRegistryUpdateDryRun

**Reason for failure**:

```
--- FAIL: TestServiceRegistryUpdateDryRun (0.15s)
    rest_test.go:1305: Expected no error: Service "foo3" is invalid: spec.clusterIP: Invalid value: "1.2.3.4": provided IP is already allocated
```

With that toleration removed in https://github.com/kubernetes/kubernetes/pull/93605, this flake has been seen (https://github.com/kubernetes/kubernetes/pull/93605#issuecomment-668142138).

1. Fix the bug in the test that casts the return value even when the create fails:

```diff
--- a/pkg/registry/core/service/storage/rest_test.go
+++ b/pkg/registry/core/service/storage/rest_test.go
@@ -1301,10 +1301,10 @@ func TestServiceRegistryUpdateDryRun(t *testing.T) {
                        }},
                },
        }, rest.ValidateAllObjectFunc, &metav1.CreateOptions{})
-       svc = obj.(*api.Service)
        if err != nil {
                t.Fatalf("Expected no error: %v", err)
        }
+       svc = obj.(*api.Service)
        _, _, err = storage.Update(ctx, svc.Name, rest.DefaultUpdatedObjectInfo(&api.Service{
                ObjectMeta: metav1.ObjectMeta{
                        Name:            svc.Name,
```

2. Omit logging on slow etcd writes because of etcd-io/etcd#12197

```
--- a/vendor/go.etcd.io/etcd/etcdserver/apply_v2.go
+++ b/vendor/go.etcd.io/etcd/etcdserver/apply_v2.go
@@ -114,7 +114,7 @@ func (a *applierV2store) Sync(r *RequestV2) Response {
 // applyV2Request interprets r as a call to v2store.X
 // and returns a Response interpreted from v2store.Event
 func (s *EtcdServer) applyV2Request(r *RequestV2) Response {
-	defer warnOfExpensiveRequest(s.getLogger(), time.Now(), r, nil, nil)
+	// defer warnOfExpensiveRequest(s.getLogger(), time.Now(), r, nil, nil)
 
 	switch r.Method {
 	case "POST":
```

3. Build a test binary for the affected package:

```
go test -race -c ./pkg/registry/core/service/storage
```

4. Stress the affected test using the [stress](https://godoc.org/golang.org/x/tools/cmd/stress) tool:

```
stress ./storage.test -test.run TestServiceRegistryUpdateDryRun
```

```sh
ls -lat $TMPDIR/go-stress*
```

/sig network

aojea:
The test suite was using a /24 cluster network for the allocator.
    The ip allocator, if no ip is specified when creating the cluster,
    picks one randomly, that means that we had 1/256 chances of
    collision.
    
    The TestServiceRegistryUpdateDryRun was creating a service without
    a ClusterIP, the ip allocator assigned one random, and it was
    never deleting it. The same test was checking later if one
    specific IP was not allocated, not taking into consideration
    that the same ip may have allocated to the first Service.
    
    To avoid any randomness, we create the first Service with a specific
    IP address.

```release-note
NONE
```

/kind flake

Fixes: #93899

### DIFF
--- a/pkg/registry/core/service/storage/rest_test.go
+++ b/pkg/registry/core/service/storage/rest_test.go
@@ -1182,10 +1182,10 @@ func TestServiceRegistryUpdateDryRun(t *testing.T) {
 			}},
 		},
 	}, rest.ValidateAllObjectFunc, &metav1.CreateOptions{})
-	svc := obj.(*api.Service)
 	if err != nil {
 		t.Fatalf("Expected no error: %v", err)
 	}
+	svc := obj.(*api.Service)
 
 	// Test dry run update request external name to node port
 	updatedSvc, created, err := storage.Update(ctx, svc.Name, rest.DefaultUpdatedObjectInfo(&api.Service{
@@ -1251,6 +1251,7 @@ func TestServiceRegistryUpdateDryRun(t *testing.T) {
 			Selector:        map[string]string{"bar": "baz"},
 			SessionAffinity: api.ServiceAffinityNone,
 			Type:            api.ServiceTypeNodePort,
+			ClusterIP:       "1.2.3.5",
 			Ports: []api.ServicePort{{
 				NodePort:   30020,
 				Port:       6502,
@@ -1259,10 +1260,11 @@ func TestServiceRegistryUpdateDryRun(t *testing.T) {
 			}},
 		},
 	}, rest.ValidateAllObjectFunc, &metav1.CreateOptions{})
-	svc = obj.(*api.Service)
 	if err != nil {
 		t.Fatalf("Expected no error: %v", err)
 	}
+	svc = obj.(*api.Service)
+
 	_, _, err = storage.Update(ctx, svc.Name, rest.DefaultUpdatedObjectInfo(&api.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            svc.Name,
@@ -1301,10 +1303,10 @@ func TestServiceRegistryUpdateDryRun(t *testing.T) {
 			}},
 		},
 	}, rest.ValidateAllObjectFunc, &metav1.CreateOptions{})
-	svc = obj.(*api.Service)
 	if err != nil {
 		t.Fatalf("Expected no error: %v", err)
 	}
+	svc = obj.(*api.Service)
 	_, _, err = storage.Update(ctx, svc.Name, rest.DefaultUpdatedObjectInfo(&api.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            svc.Name,


aojea:
```
$ stress ./storage.test -test.run TestServiceRegistryUpdateDryRun
108 runs so far, 0 failures
223 runs so far, 0 failures
333 runs so far, 0 failures
445 runs so far, 0 failures
552 runs so far, 0 failures
663 runs so far, 0 failures
770 runs so far, 0 failures
866 runs so far, 0 failures
965 runs so far, 0 failures
1067 runs so far, 0 failures
1167 runs so far, 0 failures
1267 runs so far, 0 failures
1364 runs so far, 0 failures
1466 runs so far, 0 failures
1565 runs so far, 0 failures
1663 runs so far, 0 failures
1759 runs so far, 0 failures
1861 runs so far, 0 failures
1960 runs so far, 0 failures
2058 runs so far, 0 failures
^C
````

File: pkg/registry/core/service/storage/rest_test.go [hunk 1]
```diff
@@ -262,7 +262,7 @@ func NewTestRESTWithPods(t *testing.T, endpoints *api.EndpointsList, pods *api.P
 }
 
 func makeIPNet(t *testing.T) *net.IPNet {
-	_, net, err := net.ParseCIDR("1.2.3.0/24")
```
liggitt, line 265: is makeIPNet6 good?

liggitt:
nice catch!

liggitt:
> That means that we had 1/256 chances of hitting a duplicate, using a /16 for the allocator makes more difficult to fail to duplicate IPs

is it still possible, though?

pkg/registry/core/service/storage/rest_test.go
aojea: it has a /108 that are 2^24 , we have it limited to 2^20 IIRC, should not be a problem if we don't hit it now with 2^16

aojea:
> > That means that we had 1/256 chances of hitting a duplicate, using a /16 for the allocator makes more difficult to fail to duplicate IPs
> 
> is it still possible, though?

yep
https://github.com/kubernetes/kubernetes/blob/343817ef93d6fe71ac74b62a9db85c4a8e59ba78/pkg/registry/core/service/allocator/bitmap.go#L202

:thinking:  but we can create the service in previous step with an specific IP

aojea:
yep, we can avoid any random behavior specifying the service cluster IP in one of the previous steps, where a service is created without it, so it picks one random

```
-       _, net, err := net.ParseCIDR("1.2.3.0/16")
+       _, net, err := net.ParseCIDR("1.2.3.0/28")
        if err != nil {
                t.Error(err)
        }
@@ -1251,6 +1251,7 @@ func TestServiceRegistryUpdateDryRun(t *testing.T) {
                        Selector:        map[string]string{"bar": "baz"},
                        SessionAffinity: api.ServiceAffinityNone,
                        Type:            api.ServiceTypeNodePort,
+                       ClusterIP:       "1.2.3.5",
                        Ports: []api.ServicePort{{
```
I've reduced the subnet size and didn't hit it
```
2057 runs so far, 0 failures
2154 runs so far, 0 failures
2248 runs so far, 0 failures
2347 runs so far, 0 failures
```

pkg/registry/core/service/storage/rest_test.go
```diff
@@ -1251,6 +1251,7 @@ func TestServiceRegistryUpdateDryRun(t *testing.T) {
 			Selector:        map[string]string{"bar": "baz"},
 			SessionAffinity: api.ServiceAffinityNone,
 			Type:            api.ServiceTypeNodePort,
+			ClusterIP:       "1.2.3.5",
```
aojea, line 1254: This service was created at the beginning of the test but never deleted, so it could take the IP `1.2.3.4` and never release it, causing the test to fail

aojea:
```
I0811 23:32:17.507395  117252 etcd.go:81] etcd already running at http://127.0.0.1:2379
FAIL	k8s.io/kubernetes/test/integration/master	626.176s
```

pkg/registry/core/service/storage/rest_test.go
liggitt: I'm not sure I understand this... wasn't the service created at the beginning an externalname service? I didn't think those got an assigned clusterIP

aojea: This one (L1248) is creating a new NodePort service, that's why I'm hardcoding the ClusterIP, so it does not get one assigned randomly that may collision

liggitt: oh, I see... *this* is the problematic service

aojea: ic the problem, my comment is misleading, 

> This service was created at the beginning of the test but never deleted,

replace "beginning of the test" by "before the service with the 1.2.3.4 ClusterIP" :upside_down_face:

liggitt: does setting an explicit IP here remove the need to broaden the range in `makeIPNet`?

aojea: yep, I also tested, 
https://github.com/kubernetes/kubernetes/pull/93906#issuecomment-672318048

do you prefer to keep the subnet size?

liggitt: yeah, let's keep the subnet size small (as-is) so any flakes due to non-deterministic clashing in the tests actually is observable/reproducible

aojea: done

PR #93908, opened 2020-08-12
Updating EndpointSlice controllers to return if error encountered

robscott:
**What type of PR is this?**
/kind bug

Previously the controllers would proceed with additional creates, updates, or deletes if 1 failed. That could potentially result in scenarios where an EndpointSlice create or update failing while a delete worked. This updates the logic so that removals will not happen if additions fail.

If anyone can, I'd appreciate this being added to the 1.19 milestone. More context in https://github.com/kubernetes/kubernetes/pull/93520#issuecomment-672210925

```release-note
EndpointSlice controllers now return immediately if they encounter an error creating, updating, or deleting resources.
```

/cc @lavalamp
/assign @bowei
/sig network
/priority critical-urgent

### DIFF
--- a/pkg/controller/endpointslice/reconciler.go
+++ b/pkg/controller/endpointslice/reconciler.go
@@ -27,7 +27,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	clientset "k8s.io/client-go/kubernetes"
 	corelisters "k8s.io/client-go/listers/core/v1"
@@ -180,8 +179,6 @@ func (r *reconciler) finalize(
 	slicesToDelete []*discovery.EndpointSlice,
 	triggerTime time.Time,
 ) error {
-	errs := []error{}
-
 	// If there are slices to create and delete, change the creates to updates
 	// of the slices that would otherwise be deleted.
 	for i := 0; i < len(slicesToDelete); {
@@ -218,11 +215,10 @@ func (r *reconciler) finalize(
 				if errors.HasStatusCause(err, corev1.NamespaceTerminatingCause) {
 					return nil
 				}
-				errs = append(errs, fmt.Errorf("Error creating EndpointSlice for Service %s/%s: %v", service.Namespace, service.Name, err))
-			} else {
-				r.endpointSliceTracker.Update(createdSlice)
-				metrics.EndpointSliceChanges.WithLabelValues("create").Inc()
+				return fmt.Errorf("failed to create EndpointSlice for Service %s/%s: %v", service.Namespace, service.Name, err)
 			}
+			r.endpointSliceTracker.Update(createdSlice)
+			metrics.EndpointSliceChanges.WithLabelValues("create").Inc()
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/controller/endpointslice/reconciler.go
```diff
@@ -218,36 +215,33 @@ func (r *reconciler) finalize(
 				if errors.HasStatusCause(err, corev1.NamespaceTerminatingCause) {
 					return nil
 				}
-				errs = append(errs, fmt.Errorf("Error creating EndpointSlice for Service %s/%s: %v", service.Namespace, service.Name, err))
-			} else {
-				r.endpointSliceTracker.Update(createdSlice)
-				metrics.EndpointSliceChanges.WithLabelValues("create").Inc()
+				return fmt.Errorf("Error creating EndpointSlice for Service %s/%s: %v", service.Namespace, service.Name, err)
```
wojtek-t, line 218: nit: s/Error/error/  [errors should start with small letter]

alternatively: "failed to create EndpointSlice ..."

pkg/controller/endpointslice/reconciler.go
```diff
@@ -218,36 +215,33 @@ func (r *reconciler) finalize(
...
 
 	for _, endpointSlice := range slicesToUpdate {
 		addTriggerTimeAnnotation(endpointSlice, triggerTime)
 		updatedSlice, err := r.client.DiscoveryV1beta1().EndpointSlices(service.Namespace).Update(context.TODO(), endpointSlice, metav1.UpdateOptions{})
 		if err != nil {
-			errs = append(errs, fmt.Errorf("Error updating %s EndpointSlice for Service %s/%s: %v", endpointSlice.Name, service.Namespace, service.Name, err))
-		} else {
-			r.endpointSliceTracker.Update(updatedSlice)
-			metrics.EndpointSliceChanges.WithLabelValues("update").Inc()
+			return fmt.Errorf("Error updating %s EndpointSlice for Service %s/%s: %v", endpointSlice.Name, service.Namespace, service.Name, err)
```
wojtek-t, line 229: same here and everywhere below

pkg/controller/endpointslice/reconciler.go
robscott: Updated, thanks!

pkg/controller/endpointslice/reconciler.go
robscott: Updated, thanks!

aojea:
ping @liggit for milestone

pkg/controller/endpointslice/reconciler.go
```diff
@@ -218,36 +215,33 @@ func (r *reconciler) finalize(
...
 
 	for _, endpointSlice := range slicesToUpdate {
 		addTriggerTimeAnnotation(endpointSlice, triggerTime)
 		updatedSlice, err := r.client.DiscoveryV1beta1().EndpointSlices(service.Namespace).Update(context.TODO(), endpointSlice, metav1.UpdateOptions{})
 		if err != nil {
-			errs = append(errs, fmt.Errorf("Error updating %s EndpointSlice for Service %s/%s: %v", endpointSlice.Name, service.Namespace, service.Name, err))
-		} else {
-			r.endpointSliceTracker.Update(updatedSlice)
-			metrics.EndpointSliceChanges.WithLabelValues("update").Inc()
+			return fmt.Errorf("failed to update %s EndpointSlice for Service %s/%s: %v", endpointSlice.Name, service.Namespace, service.Name, err)
```
bowei, line 229: do we track errors as a metric? (can be in follow PR)

liggitt:
I'll defer to sig-network leads for inclusion in the milestone

aojea:
> I'll defer to sig-network leads for inclusion in the milestone

ping @thockin

PR #93521, opened 2020-07-29
Update e2e csi images to k8s.gcr.io

msau42:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
Switch csi e2e test images from using k8s-staging to k8s.gcr.io

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #93909, opened 2020-08-12
Update snapshot controller to use k8s.gcr.io

msau42:
**What type of PR is this?**
/kind cleanup

Move to a stable image repo instead of staging which auto-deletes images after 60 days

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster/addons/volumesnapshots/volume-snapshot-controller/volume-snapshot-controller-deployment.yaml
+++ b/cluster/addons/volumesnapshots/volume-snapshot-controller/volume-snapshot-controller-deployment.yaml
@@ -22,8 +22,6 @@ spec:
       serviceAccount: volume-snapshot-controller
       containers:
         - name: volume-snapshot-controller
-          # TODO(xyang): Replace with an official image when it is released
-          image: gcr.io/k8s-staging-csi/snapshot-controller:v2.0.0-rc2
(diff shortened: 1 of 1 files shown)


xing-yang:
Have we also updated the csi-snapshotter sidecar to use v2.1.1 image?

msau42:
Snapshotter is still using v2.1.0: https://github.com/kubernetes/kubernetes/pull/93521

There shouldn't be a dependency between the two right?

xing-yang:
> Snapshotter is still using v2.1.0: #93521
> There shouldn't be a dependency between the two right? 

We have never tested them separately.  They may not work properly if versions are different.

xing-yang:
Is k8s.gcr.io/sig-storage/csi-snapshotter:v2.1.1 not available yet?  I can do docker pull of v2.1.0, but not v2.1.1.

msau42:
In principle though they are supposed to be compatible. The upgrade of the snapshot controller can happen independently of the sidecar since they are deployed separately. If there is a problem with the version skew, we need to fix it

msau42:
> Is k8s.gcr.io/sig-storage/csi-snapshotter:v2.1.1 not available yet? I can do docker pull of v2.1.0, but not v2.1.1.

We'll need to pull from quay and push to staging, and then promote.

PR #93918, opened 2020-08-12
test(userspace): deflake TestTCPProxyUpdateDeleteUpdate

### LINKED ISSUE #93904: [sig-network] TestTCPProxyUpdateDeleteUpdate flakes with "error connecting to server" error

TestTCPProxyUpdateDeleteUpdate

**Reason for failure**:

```
E0811 20:46:12.851363 2432471 proxysocket.go:98] Couldn't find an endpoint for testnamespace/echo:p: missing endpoints
E0811 20:46:12.851726 2432471 proxysocket.go:144] Failed to connect to balancer: missing endpoints
--- FAIL: TestTCPProxyUpdateDeleteUpdate (0.06s)
    proxier_test.go:181: error connecting to server: Get "http://127.0.0.1:42275/aaaaa": http: server closed idle connection
FAIL
```

With that toleration removed in https://github.com/kubernetes/kubernetes/pull/93605, this flake has been seen (https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93605/pull-kubernetes-bazel-test/1293275153643343872).

```
go test -race -c ./pkg/proxy/userspace
```

```
stress ./userspace.test -test.run TestTCPProxyUpdateDeleteUpdate
```

```sh
ls -lat $TMPDIR/go-stress*
```

/sig network
cc @kubernetes/sig-network-test-failures

knight42:

**What type of PR is this?**

/kind flake

I added the following logging:
```diff
diff --git pkg/proxy/userspace/roundrobin.go pkg/proxy/userspace/roundrobin.go
index 1a223bdd145..5c5de1d2dba 100644
--- pkg/proxy/userspace/roundrobin.go
+++ pkg/proxy/userspace/roundrobin.go
@@ -109,6 +109,7 @@ func (lb *LoadBalancerRR) DeleteService(svcPort proxy.ServicePortName) {
        klog.V(4).Infof("LoadBalancerRR DeleteService %q", svcPort)
        lb.lock.Lock()
        defer lb.lock.Unlock()
+       fmt.Println("DeleteService")
        delete(lb.services, svcPort)
 }

@@ -138,6 +139,7 @@ func (lb *LoadBalancerRR) NextEndpoint(svcPort proxy.ServicePortName, srcAddr ne
        // can prove it matters.
        lb.lock.Lock()
        defer lb.lock.Unlock()
+       fmt.Println("NextEndpoint")

        state, exists := lb.services[svcPort]
        if !exists || state == nil {
@@ -225,10 +227,10 @@ func (lb *LoadBalancerRR) OnEndpointsAdd(endpoints *v1.Endpoints) {

        lb.lock.Lock()
        defer lb.lock.Unlock()
+       fmt.Printf("OnEndpointsAdd: len(services)=%v\n", len(lb.services))

        for portname := range portsToEndpoints {
                svcPort := proxy.ServicePortName{NamespacedName: types.NamespacedName{Namespace: endpoints.Namespace, Name: endpoints.Name}, Port: portname}
                newEndpoints := portsToEndpoints[portname]
```
and got
```
...
OnEndpointsAdd: len(services)=0
NextEndpoint
DeleteService
OnEndpointsAdd: len(services)=0
NextEndpoint
PASS
OnEndpointsAdd: len(services)=0
NextEndpoint
NextEndpoint
OnEndpointsAdd: len(services)=1
DeleteService
NextEndpoint
E0812 13:06:39.707190   30332 proxysocket.go:98] Couldn't find an endpoint for testnamespace/echo:p: missing endpoints
E0812 13:06:39.707567   30332 proxysocket.go:144] Failed to connect to balancer: missing endpoints
--- FAIL: TestTCPProxyUpdateDeleteUpdate (0.06s)
    proxier_test.go:181: error connecting to server: Get "http://127.0.0.1:55175/aaaaa": EOF
FAIL
```

The log indicates that the service has not been deleted before we add the endpoint back at https://github.com/kubernetes/kubernetes/blob/fa13dc11d12ba9cab8c462db4a3fee5fd09d9b07/pkg/proxy/userspace/proxier_test.go#L654-L655, and the service is deleted after that which also removes the endpoint, hence the test fails.

To fix the problem, I decided to call `p.syncProxyRules()` after we delete the service to ensure the service is deleted before we add back the endpoint.

Fixes #93904

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/proxy/userspace/proxier_test.go
+++ b/pkg/proxy/userspace/proxier_test.go
@@ -30,7 +30,7 @@ import (
 	"testing"
 	"time"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	discovery "k8s.io/api/discovery/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -650,6 +650,8 @@ func TestTCPProxyUpdateDeleteUpdate(t *testing.T) {
 		t.Fatalf(err.Error())
 	}
 	waitForNumProxyLoops(t, p, 0)
+
+	p.syncProxyRules() // ensure the endpoint has been deleted
 
 	// need to add endpoint here because it got clean up during service delete
 	lb.OnEndpointsAdd(endpoint)


pkg/proxy/userspace/roundrobin.go
```diff
@@ -226,9 +226,8 @@ func (lb *LoadBalancerRR) OnEndpointsAdd(endpoints *v1.Endpoints) {
 	lb.lock.Lock()
 	defer lb.lock.Unlock()
 
-	for portname := range portsToEndpoints {
-		svcPort := proxy.ServicePortName{NamespacedName: types.NamespacedName{Namespace: endpoints.Namespace, Name: endpoints.Name}, Port: portname}
-		newEndpoints := portsToEndpoints[portname]
+	for portName, newEndpoints := range portsToEndpoints {
```
liggitt, line 229: what was the reason for this change? I'm pretty sure the previous assignment was to avoid reuse of the iterator value because of https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable

knight42: `newEndpoints` is a list and `util.ShuffleStrings` would make a copy of it.

liggitt: I'd still prefer to limit the changes in this PR to ones fixing the flake, especially if we intend to merge it in 1.19

knight42: That's okay.

aojea:
:/

pkg/proxy/userspace/proxier_test.go
```diff
@@ -651,6 +651,8 @@ func TestTCPProxyUpdateDeleteUpdate(t *testing.T) {
 	}
 	waitForNumProxyLoops(t, p, 0)
 
+	p.syncProxyRules() // ensure the endpoint has been deleted
```
dcbw, line 654: I don't think this correct. Waiting for syncProxyRules() to run is exactly what waitForNumProxyLoops() is supposed to do. This should be effectively running syncProxyRules() twice when we expect to only need it run once.

So the question is either:

1. is waitForNumProxyLoops() doing the right thing and actually waiting for syncProxyRules() to run once,
2. is syncProxyRules() getting correctly triggered by the OnServiceDelete

dcbw: @knight42 see https://github.com/kubernetes/kubernetes/pull/93979 for a larger cleanup that hopefully addresses the flake too

knight42: @dcbw Great! Thanks for your fix!

knight42:
Close in favor of #93979

PR #93919, opened 2020-08-12
Modify the warning message format from "%d" to "%v" in shared_informer.go.

a2ush:
**What type of PR is this?**
/kind feature

Modify the warning message format from "%d" to "%v".
Currently, we see the following warning log when resyncPeriod is `time.Second*1/2`.
```Go
informerFactory := informers.NewSharedInformerFactory(clientset, time.Second*1/2)
```
```
W0805 09:35:40.766019   26122 shared_informer.go:409] resyncPeriod 500000000 is too small. Changing it to the minimum allowed value of 1000000000
```
And, we also see the following warning log when resyncPeriod is `time.Second*10` and resyncCheckPeriod is `time.Second*30`.
```
W0805 13:11:40.864639   28885 shared_informer.go:415] resyncPeriod 1000000000 is smaller than resyncCheckPeriod 30000000000 and the informer has already started. Changing it to 30000000000
```
※ The unit looks like nanosecond(ns).

I think this warning logs are difficult to understand and they are difficult to find the fix.

When we change the format from "%d" to "%v", we can see the following logs.
```
W0805 17:41:50.543011   17910 shared_informer.go:409] resyncPeriod 500ms is too small. Changing it to the minimum allowed value of 1s
```
```
W0805 17:46:47.342462   18154 shared_informer.go:415] resyncPeriod 10s is smaller than resyncCheckPeriod 30s and the informer has already started. Changing it to 30s
```

I think this behavior is based on Fprintf function in Go language. I confirm Warningf function uses Fprintf.[1][2]
If we continue to use Warningf, I think it would be better to change the format (to use "%v").

[1] https://github.com/kubernetes/klog/blob/master/klog.go#L1413-L1417
[2] https://github.com/kubernetes/klog/blob/master/klog.go#L726-L739

### DIFF
--- a/staging/src/k8s.io/client-go/tools/cache/shared_informer.go
+++ b/staging/src/k8s.io/client-go/tools/cache/shared_informer.go
@@ -485,13 +485,13 @@ func (s *sharedIndexInformer) AddEventHandlerWithResyncPeriod(handler ResourceEv
 
 	if resyncPeriod > 0 {
 		if resyncPeriod < minimumResyncPeriod {
-			klog.Warningf("resyncPeriod %d is too small. Changing it to the minimum allowed value of %d", resyncPeriod, minimumResyncPeriod)
+			klog.Warningf("resyncPeriod %v is too small. Changing it to the minimum allowed value of %v", resyncPeriod, minimumResyncPeriod)
 			resyncPeriod = minimumResyncPeriod
 		}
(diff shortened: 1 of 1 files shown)


neolit123 (commented):
/lgtm
/priority backlog
/ok-to-test

Issue #88996, opened 2020-03-10
[Flaky Test] [sig-cli] Kubectl client Kubectl logs should be able to retrieve and filter logs [Conformance]


aojea:
**Which jobs are flaking**:

https://storage.googleapis.com/k8s-gubernator/triage/index.html?ci=0&pr=1&test=Kubectl%20logs%20should%20be%20able%20to%20retrieve%20and%20filter%20logs

**Which test(s) are flaking**:

https://storage.googleapis.com/k8s-gubernator/triage/index.html?ci=0&pr=1&test=Kubectl%20logs%20should%20be%20able%20to%20retrieve%20and%20filter%20logs

**Testgrid link**:

**Reason for failure**:
This occurrence fails because it has `tail=1` however it outputs 2 lines instead of 1
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/88989/pull-kubernetes-conformance-kind-ipv6-parallel/1237265279893901312/

```
Mar 10 06:51:34.637: INFO: Running '/go/src/k8s.io/kubernetes/bazel-bin/cmd/kubectl/linux_amd64_pure_stripped/kubectl --server=https://[::1]:46239 --kubeconfig=/root/.kube/kind-test-config logs logs-generator logs-generator --namespace=kubectl-1201 --tail=1 --timestamps'
Mar 10 06:51:35.034: INFO: stderr: ""
Mar 10 06:51:35.040: INFO: stdout: "2020-03-10T06:51:34.798382054Z I0310 06:51:34.796220       1 logs_generator.go:76] 31 PUT /api/v1/namespaces/default/pods/8rzv 347\n2020-03-10T06:51:35.004235712Z I0310 06:51:35.004040       1 logs_generator.go:76] 32 PUT /api/v1/namespaces/kube-system/pods/h89n 206\n"
Mar 10 06:51:35.040: INFO: got output "2020-03-10T06:51:34.798382054Z I0310 06:51:34.796220       1 logs_generator.go:76] 31 PUT /api/v1/namespaces/default/pods/8rzv 347\n2020-03-10T06:51:35.004235712Z I0310 06:51:35.004040       1 logs_generator.go:76] 32 PUT /api/v1/namespaces/kube-system/pods/h89n 206\n"
Mar 10 06:51:35.040: FAIL: Expected
    <int>: 2
to equal
    <int>: 1
```

liggitt:
[...]

PR #93920, opened 2020-08-12
[Flaky Test] Add limited lines to log when having tail option

zhouya0:
**What type of PR is this?**
/kind bug
/kind flake

This PR fixes part of the flaky test https://github.com/kubernetes/kubernetes/issues/92468, https://github.com/kubernetes/kubernetes/issues/88996
The e2e test show following error:
```
Mar 10 06:51:34.637: INFO: Running '/go/src/k8s.io/kubernetes/bazel-bin/cmd/kubectl/linux_amd64_pure_stripped/kubectl --server=https://[::1]:46239 --kubeconfig=/root/.kube/kind-test-config logs logs-generator logs-generator --namespace=kubectl-1201 --tail=1 --timestamps'
Mar 10 06:51:35.034: INFO: stderr: ""
Mar 10 06:51:35.040: INFO: stdout: "2020-03-10T06:51:34.798382054Z I0310 06:51:34.796220       1 logs_generator.go:76] 31 PUT /api/v1/namespaces/default/pods/8rzv 347\n2020-03-10T06:51:35.004235712Z I0310 06:51:35.004040       1 logs_generator.go:76] 32 PUT /api/v1/namespaces/kube-system/pods/h89n 206\n"
Mar 10 06:51:35.040: INFO: got output "2020-03-10T06:51:34.798382054Z I0310 06:51:34.796220       1 logs_generator.go:76] 31 PUT /api/v1/namespaces/default/pods/8rzv 347\n2020-03-10T06:51:35.004235712Z I0310 06:51:35.004040       1 logs_generator.go:76] 32 PUT /api/v1/namespaces/kube-system/pods/h89n 206\n"
Mar 10 06:51:35.040: FAIL: Expected
    <int>: 2
to equal
    <int>: 1
```
It's because `--tail` flag result in the starting index of the log file:
https://github.com/kubernetes/kubernetes/blob/fa13dc11d12ba9cab8c462db4a3fee5fd09d9b07/pkg/kubelet/kuberuntime/logs/logs.go#L297-L300
So when reading the log while writing continuously. The start index will not be the actually `--tail` number.

And this bug can be easy reproduced: having a pod with command like:
```
while true; do echo testlog; done;
```
Do the `kubectl logs pod/testpod --tail 1` and will find the output is unpredictable.

This PR will limit the writing action when reading the log file. The limit is the `--tail` number.
When the number is less than 0 , it will do nothing.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/kubelet/kuberuntime/logs/logs.go
+++ b/pkg/kubelet/kuberuntime/logs/logs.go
@@ -303,6 +303,8 @@ func ReadLogs(ctx context.Context, path, containerID string, opts *LogOptions, r
 		return fmt.Errorf("failed to seek %d in log file %q: %v", start, path, err)
 	}
 
+	limitedMode := (opts.tail >= 0) && (!opts.follow)
+	limitedNum := opts.tail
 	// Start parsing the logs.
 	r := bufio.NewReader(f)
 	// Do not create watcher here because it is not needed if `Follow` is false.
@@ -313,7 +315,7 @@ func ReadLogs(ctx context.Context, path, containerID string, opts *LogOptions, r
 	writer := newLogWriter(stdout, stderr, opts)
 	msg := &logMessage{}
 	for {
-		if stop {
+		if stop || (limitedMode && limitedNum == 0) {
 			klog.V(2).Infof("Finish parsing log file %q", path)
 			return nil
 		}
@@ -401,6 +403,10 @@ func ReadLogs(ctx context.Context, path, containerID string, opts *LogOptions, r
 			klog.Errorf("Failed with err %v when writing log for log file %q: %+v", err, path, msg)
 			return err
 		}
+		if limitedMode {
+			limitedNum--
+		}
+
 	}
 }
 


eddiezane (approved):
Maybe add a comment so this is clear.

/lgtm

mattjmcnaughton (commented):
Thanks for helping make tests more stable @zhouya0 !

One thing I'm noticing is that we updating non-test code in-order to deflake a test. I just want to double check that the issue is with the non-test code and not just that the test should be verifying a different, less-flaky behavior?

In other words, if we didn't have the flakey conformance test that you're trying to fix, would we still want to make this change?

Thanks again for your efforts to help get a constantly green test suite :)

zhouya0:
@mattjmcnaughton
Actually you can see this as a bug. Just it was found in a flaky test. And the bug is:
- When having pod<xxx> with log writing continuously, `kubectl logs pods/<xxx> --tail n` can give the log lines which is more than `n`.
But the possibility are relatively low and seems no one complains until this flaky test.

Since it is a bug , I think it might be reasonable to be fixed. 
/retest

mattjmcnaughton:
> @mattjmcnaughton
> Actually you can see this as a bug. Just it was found in a flaky test. And the bug is:
> 
>     * When having pod with log writing continuously, `kubectl logs pods/<xxx> --tail n` can give the log lines which is more than `n`.
> 
> 
> But the possibility are relatively low and seems no one complains until this flaky test.
> Since it is a bug , I think it might be reasonable to be fixed. Just the way I'm having here might not be the best solution :0
> 
> Or, the simple way to do this is just delete the flaky test if we don't want to change anything in actual code and get merged in 1.19.
> What action should I take? @hasheddan
> /retest

Gotcha, thanks @zhouya0 ! I agree that both are options, and I don't feel super strong about which one we choose. If we do decide to treat this as a bug, it might be nice to add a unit test verifying the new behavior that you're enforcing, in order to prevent regressions in the future. Thanks :)

RobertKielty:
Hi @mattjmcnaughton, hi @zhouya0 

I am walking the [CI Signal board](https://github.com/orgs/kubernetes/projects/11?fullscreen=true) and I found myself here via #92468

I agree that it would be a good idea to cover this change with a unit test.

I do not think we should disable the test in CI.

We can add a help-wanted label to get that unit test written if needed.

zhouya0:
@RobertKielty I have to say it's a little bit difficult to mock `ReadLogs` unit test.
/help wanted

brianpursley:
I added some unit tests for ReadLogs in PR https://github.com/kubernetes/kubernetes/pull/95099.

Testing the flake condition will be hard because it is a data race, but also because it will be hard to cause the condition at exactly the right time.

But hopefully the tests I added in the linked PR above will be good enough to let this PR proceed.

pkg/kubelet/kuberuntime/logs/logs.go
```diff
@@ -303,6 +303,8 @@ func ReadLogs(ctx context.Context, path, containerID string, opts *LogOptions, r
 		return fmt.Errorf("failed to seek %d in log file %q: %v", start, path, err)
 	}
 
+	limitedMode := (opts.tail >= 0) && (!opts.follow)
```
derekwaynecarr, line 306: just noting:

the default is -1 for tail (read all log) and default for follow is false.

pkg/kubelet/kuberuntime/logs/logs.go
```diff
@@ -313,7 +315,7 @@ func ReadLogs(ctx context.Context, path, containerID string, opts *LogOptions, r
 	writer := newLogWriter(stdout, stderr, opts)
 	msg := &logMessage{}
 	for {
-		if stop {
+		if stop || (limitedMode && limitedNum == 0) {
```
derekwaynecarr, line 318: I am confused why this is needed.  wont `FindTailLineStartIndex` relative to the tail option handle this?

derekwaynecarr: I read this closer, and understand if writes are still happening relative to this tail option, users may get additional log lines.

derekwaynecarr (approved):
I worried this was potentially backward incompatible, but I think its fine.

From PodLogOptions:

```
	// If set, the number of lines from the end of the logs to show. If not specified,
	// logs are shown from the creation of the container or sinceSeconds or sinceTime
	// +optional
	TailLines *int64 `json:"tailLines,omitempty" protobuf:"varint,7,opt,name=tailLines"`
```

This ensures that tail lines is actually representative of number of lines sent relative to the request, and additional lines are not returned if writes are occuring when the start index was calculated.

/approve
/lgtm

PR #93922, opened 2020-08-12
Fix auto-allocated ports conflicting with explicitly specified ports on service create

knight42:

**What type of PR is this?**

/kind flake
/kind bug

I added the following debugging:
```diff
diff --git pkg/registry/core/service/storage/rest.go pkg/registry/core/service/storage/rest.go
index 0d5379f8c11..8095f386431 100644
--- pkg/registry/core/service/storage/rest.go
+++ pkg/registry/core/service/storage/rest.go
@@ -208,6 +208,8 @@ func (rs *REST) Create(ctx context.Context, obj runtime.Object, createValidation
        // Handle ExternalTraffic related fields during service creation.
        if apiservice.NeedsHealthCheck(service) {
                if err := allocateHealthCheckNodePort(service, nodePortOp); err != nil {
+                       spec := service.Spec
+                       fmt.Printf("NodePort: %d, HealthCheckPort: %d\n", spec.Ports[0].NodePort, spec.HealthCheckNodePort)
                        return nil, errors.NewInternalError(err)
                }
        }
```

and got:
```
WARNING: Package "github.com/golang/protobuf/protoc-gen-go/generator" is deprecated.
        A future release of golang/protobuf will delete this package,
        which has long been excluded from the compatibility promise.

I0812 17:52:19.904068   78356 client.go:360] parsed scheme: "endpoint"
I0812 17:52:19.904194   78356 endpoint.go:68] ccResolverWrapper: sending new addresses to cc: [{unix://localhost:83760927833799529510  <nil> 0 <nil>}]
I0812 17:52:19.908507   78356 client.go:360] parsed scheme: "endpoint"
I0812 17:52:19.908588   78356 endpoint.go:68] ccResolverWrapper: sending new addresses to cc: [{unix://localhost:83760927833799529510  <nil> 0 <nil>}]
I0812 17:52:19.909254   78356 once.go:66] CPU time info is unavailable on non-linux or appengine environment.
I0812 17:52:19.914674   78356 client.go:360] parsed scheme: "endpoint"
I0812 17:52:19.914761   78356 endpoint.go:68] ccResolverWrapper: sending new addresses to cc: [{unix://localhost:83760927833799529510  <nil> 0 <nil>}]
I0812 17:52:19.918666   78356 client.go:360] parsed scheme: "endpoint"
I0812 17:52:19.918736   78356 endpoint.go:68] ccResolverWrapper: sending new addresses to cc: [{unix://localhost:83760927833799529510  <nil> 0 <nil>}]

NodePort: 30773, HealthCheckPort: 30773  <-- Please notice this line

W0812 17:52:19.961603   78356 clientconn.go:1223] grpc: addrConn.createTransport failed to connect to {unix://localhost:83760927833799529510  <nil> 0 <nil>}. Err :connection error: desc = "transport: Error while dialing dial unix localhost:83760927833799529510: connect: no such file or directory". Reconnecting...
W0812 17:52:19.961638   78356 clientconn.go:1223] grpc: addrConn.createTransport failed to connect to {unix://localhost:83760927833799529510  <nil> 0 <nil>}. Err :connection error: desc = "transport: Error while dialing dial unix localhost:83760927833799529510: connect: no such file or directory". Reconnecting...
W0812 17:52:19.961971   78356 clientconn.go:1223] grpc: addrConn.createTransport failed to connect to {unix://localhost:83760927833799529510  <nil> 0 <nil>}. Err :connection error: desc = "transport: Error while dialing dial unix localhost:83760927833799529510: connect: no such file or directory". Reconnecting...
--- FAIL: TestServiceRegistryExternalTrafficHealthCheckNodePortUserAllocation (1.98s)
    rest_test.go:2214: Unexpected failure creating service :Internal error occurred: failed to allocate requested HealthCheck NodePort 30773: provided port is already allocated
FAIL

ERROR: exit status 1
```

The log shows the root cause is the randomly generated health check nodePort happens to be the same as the service nodePort allocated later.

The flake also uncovers a bug that apiserver could potenially allocate a same port as the user-specified HealthCheck NodePort to be the Service NodePort when creating `NodePort` or `LoadBalancer` Service.

xref: https://github.com/kubernetes/kubernetes/pull/93605#issuecomment-668141670

```release-note
Fix a bug that apiserver could potenially allocate a same port as the user-specified HealthCheck NodePort to be the Service NodePort when creating `NodePort` or `LoadBalancer` Service.
```

```docs

```

### DIFF
--- a/pkg/registry/core/service/storage/rest.go
+++ b/pkg/registry/core/service/storage/rest.go
@@ -205,12 +205,6 @@ func (rs *REST) Create(ctx context.Context, obj runtime.Object, createValidation
 		}
 	}
 
-	// Handle ExternalTraffic related fields during service creation.
-	if apiservice.NeedsHealthCheck(service) {
-		if err := allocateHealthCheckNodePort(service, nodePortOp); err != nil {
-			return nil, errors.NewInternalError(err)
-		}
-	}
 	if errs := validation.ValidateServiceExternalTrafficFieldsCombination(service); len(errs) > 0 {
 		return nil, errors.NewInvalid(api.Kind("Service"), service.Name, errs)
 	}
@@ -618,18 +612,6 @@ func containsNodePort(serviceNodePorts []ServiceNodePort, serviceNodePort Servic
 	return false
 }
 
-// Loop through the service ports list, find one with the same port number and
-// NodePort specified, return this NodePort otherwise return 0.
-func findRequestedNodePort(port int, servicePorts []api.ServicePort) int {
-	for i := range servicePorts {
-		servicePort := servicePorts[i]
-		if port == int(servicePort.Port) && servicePort.NodePort != 0 {
-			return int(servicePort.NodePort)
-		}
-	}
-	return 0
-}
-
 // allocateHealthCheckNodePort allocates health check node port to service.
 func allocateHealthCheckNodePort(service *api.Service, nodePortOp *portallocator.PortAllocationOperation) error {
 	healthCheckNodePort := service.Spec.HealthCheckNodePort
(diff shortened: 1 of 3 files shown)


pkg/registry/core/service/storage/rest_test.go
```diff
@@ -2190,10 +2198,16 @@ func TestServiceRegistryExternalTrafficHealthCheckNodePortAllocation(t *testing.
...
-	randomNodePort := generateRandomNodePort()
 	ctx := genericapirequest.NewDefaultContext()
 	storage, registry, server := NewTestREST(t, nil, singleStackIPv4)
 	defer server.Terminate(t)
+
+	svcNodePort := generateRandomNodePort(storage.serviceNodePorts)
+	healthCheckPort := generateRandomNodePort(storage.serviceNodePorts)
+	for healthCheckPort == svcNodePort {
+		healthCheckPort = generateRandomNodePort(storage.serviceNodePorts)
+	}
```
liggitt, line 2209: why is it important for these ports to be random?

liggitt:
> The log shows the root cause is the randomly generated health check nodePort happens to be the same as the service nodePort allocated later.

nice catch

pkg/registry/core/service/storage/rest_test.go
knight42: because we don't know which port is available.

knight42: oh do you mean we could simply specify static node port? It makes sense to me.

liggitt: yeah, this is a unit test, I'd expect all ports to be available

liggitt: actually... does this flake mean that the API server is allocating a conflicting port? if so, isn't that a bug we should fix?

knight42: yeah, the apiserver doesn't take health port into account when allocating service node port. I would try to fix it.

knight42:
@liggitt Now that this PR actually fixes a bug, is it OK to add a release note?

pkg/registry/core/service/storage/rest.go
```diff
@@ -199,14 +199,25 @@ func (rs *REST) Create(ctx context.Context, obj runtime.Object, createValidation
...
 
+	needHealthCheck := apiservice.NeedsHealthCheck(service)
+	healthPort := int(service.Spec.HealthCheckNodePort)
 	if service.Spec.Type == api.ServiceTypeNodePort || service.Spec.Type == api.ServiceTypeLoadBalancer {
+		// If user has specified HealthCheck NodePort, we need to check if it conflicts with
+		// allocated ports.
+		// If there is no conflicts, then we reserve the given HealthCheck NodePort
+		// to prevent port allocator allocating the same port for Service NodePort later.
+		if needHealthCheck && healthPort > 0 {
+			if err := allocateHealthCheckNodePort(service, nodePortOp); err != nil {
```
liggitt, line 210: doesn't this create the same issue in reverse, where an auto-allocated health port could try to grab an explicit user-specified node port?

also, putting this inside a `if service.Spec.Type == api.ServiceTypeNodePort || service.Spec.Type == api.ServiceTypeLoadBalancer` block leaves a logical hole where the health port is not allocated if all of the following are true:

* needsHealthCheck == true
* healthPort > 0
* service.Spec.Type != api.ServiceTypeNodePort
* service.Spec.Type != api.ServiceTypeLoadBalancer

knight42: > doesn't this create the same issue in reverse

when allocating health port automatically, apiserver would call `nodePortOp.AllocateNext()` which skips the allocated ports https://github.com/kubernetes/kubernetes/blob/c780554a64fcab1c23d0e2d44d601e02db9a4036/pkg/registry/core/service/storage/rest.go#L644-L649

> leaves a logical hole where the health port is not allocated if all of the following are true:

HealthCheckNodePort can only be set on LoadBalancer service with ExternalTrafficPolicy=Local

liggitt: > HealthCheckNodePort can only be set on LoadBalancer service with ExternalTrafficPolicy=Local

it's not obvious from this code that all cases are covered... should we move the health port allocation above the `if service.Spec.Type == api.ServiceTypeNodePort || service.Spec.Type == api.ServiceTypeLoadBalancer` block?

knight42: hmmm 
1. healthCheckNodePort should make no effect if the type of Service is not `LoadBalancer` and `ExternalTrafficPolicy` is not `Local`
2. the validation will fail if users specify healthCheckNodePort on a Service other than `LoadBalancer` Service
3.  `apiservice.NeedsHealthCheck` always returns false if the type of Service is neither `LoadBalancer` nor `NodePort` https://github.com/kubernetes/kubernetes/blob/c780554a64fcab1c23d0e2d44d601e02db9a4036/pkg/api/service/util.go#L70-L86

liggitt: previously, we unconditionally did this:
```
	// Handle ExternalTraffic related fields during service creation.
	if apiservice.NeedsHealthCheck(service) {
		if err := allocateHealthCheckNodePort(service, nodePortOp); err != nil {
			return nil, errors.NewInternalError(err)
		}
	}
```

it seems good to keep the new code equally unconditional, e.g.:

```go
	// reserve the explicitly specified health check port before auto-allocating any node ports
	if apiservice.NeedsHealthCheck(service) && service.Spec.HealthCheckNodePort != 0 {
		if err := allocateHealthCheckNodePort(service, nodePortOp); err != nil {
			return nil, errors.NewInternalError(err)
		}
	}

	if service.Spec.Type == api.ServiceTypeNodePort || service.Spec.Type == api.ServiceTypeLoadBalancer {
		if err := initNodePorts(service, nodePortOp); err != nil {
			return nil, err
		}
	}

	// Handle ExternalTraffic related fields during service creation.
	// auto-allocate the health check port
	if apiservice.NeedsHealthCheck(service) && service.Spec.HealthCheckNodePort == 0 {
		if err := allocateHealthCheckNodePort(service, nodePortOp); err != nil {
			return nil, errors.NewInternalError(err)
		}
	}
```

liggitt: also, do we have a similar issue inside initNodePorts with auto-allocated node ports for items early in the `ports` list taking explicitly specified ports later in the list in a case like this:

```
spec:
  ports:
  - port: 1111
  - port: 2222
    nodePort: 30000
```

it really seems like we need a two-pass approach that allocates explicitly specified ports, then auto-allocates remaining ports

knight42: It is already midnight here, I would fix that tomorrow :)

knight42: The nodePort allocation logic is a bit complicated, I may need some time to sort it out :/

knight42:
@liggitt I think adding tests that describes our expected behavior might be a good start, so I add some tests in https://github.com/kubernetes/kubernetes/pull/93922/commits/cd1dbe121471ca88b3b4d67bac6f96ee3e7faa7c(which are intended to fail for now), and it turns out the problem you pointed out at https://github.com/kubernetes/kubernetes/pull/93922#discussion_r469350362 actually exists.

liggitt:
routing to sig-network for review, since this is now a bugfix and behavior change

BenTheElder:
```
Errors from staticcheck:
pkg/registry/core/service/storage/rest.go:617:6: func findRequestedNodePort is unused (U1000)

Please review the above warnings. You can test via:
  hack/verify-staticcheck.sh <failing package>
If the above warnings do not make sense, you can exempt the line or file. See:
  https://staticcheck.io/docs/#ignoring-problems
```

liggitt:
@bowei / @thockin - since this is a service REST storage change, it needs careful review, so I don't want to rush it, but it is blocking https://github.com/kubernetes/kubernetes/pull/93605 and full reopen of the 1.20 branch.

pkg/registry/core/service/storage/rest.go
```diff
@@ -696,51 +678,102 @@ func initClusterIP(service *api.Service, allocator ipallocator.Interface) (bool,
...
+			// TODO: what error should be returned here?  It's not a
+			// field-level validation failure (the field is valid), and it's
+			// not really an internal error.
+			return errors.NewInternalError(err)
+		}
+	}
+	for i := range service.Spec.Ports {
+		servicePort := &service.Spec.Ports[i]
+		if servicePort.NodePort == 0 {
+			nodePort, err := nodePortOp.AllocateNext()
```
aojea, line 767: I think that the only error possible here is that there are no ports free in the allocator, so it NewInternalError may make sense

liggitt:
To unblock https://github.com/kubernetes/kubernetes/pull/93605, I opened https://github.com/kubernetes/kubernetes/pull/94488 to temporarily deflake the affected unit test until this is reviewed/merged

knight42:
@bowei / @thockin I was wondering if there is anything I could do to help you review this PR?

pkg/registry/core/service/storage/rest.go
knight42: Good to know that! I've removed that comment.

pkg/registry/core/service/storage/rest_test.go
```diff
@@ -2145,6 +2186,200 @@ func TestUpdateServiceWithConflictingNamespace(t *testing.T) {
...
 }
 
+func TestServiceRegistryCreateServiceMultipleNodePortsWithSameProtocol(t *testing.T) {
+	ctx := genericapirequest.NewDefaultContext()
+	storage, registry, server := NewTestREST(t, nil, singleStackIPv4)
+	defer server.Terminate(t)
+
+	const count = 3
+	// The `PortAllocator` seems to allocate the first port randomly,
+	// So we run the test multiple times to ensure the NodePort is correctly allocated.
```
thockin, line 2196: The code change seems good, but I am not sure I understand this random + retry.  Can you help me understand?

knight42: The port allocator would allocate port randomly from the available ports, if we did not retry, the port allocator might happen to allocate the right port and the test got passed.

PR #93929, opened 2020-08-12
Validate ingress TLS secretName in v1

### LINKED ISSUE #93928: missing validation for ingress.spec.tls.secretName
In looking at the Ingress API, I checked validation and I don't see any on the .spec.tls.secretName.  See https://github.com/kubernetes/kubernetes/blob/2b091f60ca09d3d03923a26b7cd02fb3dc6c47e2/pkg/apis/networking/validation/validation.go#L235-L254

I would expect at least some validation.  The secretName field does not indicate that it is a kubernetes secret, so if the intent is to allow any string, the godoc should probably be made explicit that clients need to defend themselves against path traversal if necessary.  And we probably want a max length at least.

@kubernetes/sig-network-bugs 
@liggitt @robscott

liggitt:
**What type of PR is this?**
/kind bug
/kind api-change

Validates the Ingress TLS secretName field in `networking.k8s.io/v1`

Fixes #93928

Tightens validation in v1 for new objects, and for existing objects that already pass the tightened validation. Preserves backwards compatibility for existing objects and previously released APIs.

```release-note
When creating a networking.k8s.io/v1 Ingress API object, `spec.tls[*].secretName` values are required to pass validation rules for Secret API object names.
```

/cc @robscott @deads2k 
/sig network
/milestone v1.19

### DIFF
--- a/pkg/apis/networking/validation/validation.go
+++ b/pkg/apis/networking/validation/validation.go
@@ -199,6 +199,8 @@ var ValidateIngressName = apimachineryvalidation.NameIsDNSSubdomain
 
 // IngressValidationOptions cover beta to GA transitions for HTTP PathType
 type IngressValidationOptions struct {
+	// AllowInvalidSecretName indicates whether spec.tls[*].secretName values that are not valid Secret names should be allowed
+	AllowInvalidSecretName bool
 }
 
 // ValidateIngress validates Ingresses on create and update.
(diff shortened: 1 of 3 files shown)


pkg/apis/networking/validation/validation.go
```diff
@@ -523,3 +535,26 @@ func validateIngressTypedLocalObjectReference(params *api.TypedLocalObjectRefere
...
+		}
+	}
+	return false
+}
+
+func validateTLSSecretName(name string) []string {
+	if len(name) == 0 {
+		return nil
+	}
+	return apivalidation.ValidateSecretName(name, false)
```
deads2k, line 559: This makes sense to me, but the doc on the .secretName field is vague. it's not clear whether it must be a kube secret ref or if it can be a reference to nearly anything and we should only enforce some large length limit.

If the latter, the field doc needs to be improved to make it clear that consumers need to defend against traversal attacks.

liggitt: It's definitely a kubernetes secret. A separate field to indicate a named credential managed by the ingress controller was discussed as an alternative to the requests for cluster-scoped secrets, but was not accepted.

liggitt:
bazel debian error
/retest

liggitt:
/priority critical-urgent

to resolve API issue prior to initial release

deads2k:
lgtm, but I'd like someone from sig-network to sign off.

robscott:
Thanks!

/lgtm

Issue #75776, opened 2019-03-27
Failing tests: Multivolume in multi-zone/regional environments


msau42:
**Which jobs are failing**:
ci-kubernetes-e2e-gce-multizone
ci-kubernetes-e2e-gke-regional

**Which test(s) are failing**:
All the new multivolume tests

**Since when has it been failing**:
Since introduction on 3/26

**Testgrid link**:
https://k8s-testgrid.appspot.com/sig-storage-kubernetes#gce-multizone

**Reason for failure**:
Tests probably need to use delayed binding.

**Anything else we need to know**:

msau42:
/assign
@kubernetes/sig-storage-test-failures

msau42:
Tests introduced in https://github.com/kubernetes/kubernetes/pull/74693

PR #93930, opened 2020-08-12
e2e storage: skip multi-volume generic ephemeral volume test

pohly:
**What type of PR is this?**

/kind cleanup
/kind flake

This test might flake when run on a multi-zone cluster (similar to
persistent volumes, see
https://github.com/kubernetes/kubernetes/issues/75776). We don't do
that at the moment, but it's better to fix this anyway.

```release-note
NONE
```


pohly:
/assign @msau42 

We discussed this potential issue on Slack a while ago. I now ran into this problem for real when testing PMEM-CSI, which is "multi-zone".

Issue #78995, opened 2019-06-13
Remove the /cluster directory


timothysc:
For years we have publicly stated that the /cluster directory is deprecated and not maintained.  However every cycle it is updated and there are bugs found+fixed by sig-cluster-lifecycle.  

I'd like to enumerate in this issue what needs to get done in order for us to wholesale remove the /cluster directory. 

/assign @dims @spiffxp @justinsb @timothysc  
/cc @liggitt @neolit123

neolit123:
or potentially break it down and move certain sub-folders out of tree.
currently it contains a collection of items:
- image files (conformance image, etcd)
- kube-up
- random bash scripts used by test-infra/kubetest deployers as helpers
- addons and the addon manager.

liggitt:
looking at the references to it...

/sig testing
for e2e bringup

/sig scalability
for kubemark bringup

sftim:
Is this also relevant to SIG Docs?

timothysc:
xref -https://github.com/kubernetes/kubernetes/pull/78543 
^ Is an example of continued technical debt that we see and have to pay for in different ways across SCL.

timothysc:
xref -https://github.com/kubernetes/kubernetes/pull/78543 
^ Is an example of continued technical debt that we see and have to pay for in different ways across SCL.

jaypipes:
> Is this also relevant to SIG Docs?

Yes, I think this is a good example of where `/cluster` scripts are referenced from the docs: https://github.com/kubernetes/website/pull/14929

andrewsykim:
For v1.16: investigate if cluster API is a potential replacement and meets the same level of coverage as `/cluster` and enumerate what is missing cluster API

/assign @alejandrox1

dims:
Step 1 we talked about was ... look at all the CI jobs that use the kube-up in the cluster directory and inventory the knobs/settings/configurations they setup or use and cross check if cluster-api/kubeadm allows us to do the same.
Step 2 mirror the GCE e2e CI Job (`pull-kubernetes-e2e-gce` to be specific) using cluster-api for AWS.

Both steps can be done in parallel and will need help/effort/coordination on the wg-k8s-infra and sig-testing folks

alejandrox1:
I'll start with step 1 right away.
So step 2 is free if anyone wants to work on this as well 😃

dims:
@alejandrox1 sounds good. please start a google doc or something that we can use to compile notes on the various flags/options

[...]

PR #93937, opened 2020-08-12
WIP: move /cluster/images/conformance to /test/images/conformance

neolit123:

Move the source of the conformance image from /cluster/images/conformance to /test/images/conformance.

xref https://github.com/kubernetes/kubernetes/issues/78995
related test-infra change: https://github.com/kubernetes/test-infra/pull/18799

NONE

```release-note
NONE
```

```docs

```


build/images/conformance/README.md
```diff
@@ -37,5 +37,3 @@ If you don't want to push the images, run `make` or `make build` instead
 ```
 kubectl create -f conformance-e2e.yaml
 ```
-
-[![Analytics](https://kubernetes-site.appspot.com/UA-36037335-10/GitHub/cluster/images/conformance/README.md?pixel)]()
```
neolit123, line 41: this link is no longer valid. removed it.

BenTheElder:
I'm not sure how we want to organize build/, currently we have e.g. build/pause and build/go-runner top-level

I think maybe build/images/pause would be better for example, but it isn't how things are currently organized

BenTheElder:
This also maybe better belongs in test/... ?
The other images in build/ are critical to a functioning cluster or are base images.

neolit123:
> I think maybe build/images/pause would be better for example, but it isn't how things are currently organized

seems good. all images under /build should be under /build/images.
this can be done in separate PRs.

> This also maybe better belongs in test/..., actually ?

this actually makes more sense. i can update.

neolit123:
> This also maybe better belongs in test/..., actually ?

updated.

/assign @BenTheElder @dims

neolit123:
> pull-kubernetes-conformance-image-test

this will likely fail until https://github.com/kubernetes/test-infra/pull/18799 merges.

spiffxp:
I'm not interested in seeing this land for v1.19 but agree `cluster` is probably the wrong place for it

I'm less certain about `test/images` being the right place, if only because everything else in there, I think? is an image that is scheduled _by_ e2e tests.  The conformance image doesn't fit into that definition.

neolit123:
> I'm less certain about test/images being the right place, if only because everything else in there, I think? is an image that is scheduled by e2e tests. The conformance image doesn't fit into that definition.

i think this is mostly correct.

one partial exception that i found is:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/sample-apiserver/artifacts/example/deployment.yaml#L23

but if /build/images or /test/images are not the right locations, i'm out of ideas.

mikedanese:
Looks like some of this needs cleanup:

https://github.com/kubernetes/test-infra/blob/bec22c9286285b8f177b3436508c4dade18354ac/experiment/kind-conformance-image-e2e.sh#L121-L136

BenTheElder:
> I'm less certain about test/images being the right place, if only because everything else in there, I think? is an image that is scheduled by e2e tests. The conformance image doesn't fit into that definition.

Fair. test/e2e/image then? Little about it is conformance specific, it's a docker-wrapped e2e.test.

I don't think it belongs to build.

It's also a little weird because the only major consumer of this image has remained a downstream vendor owned tool, outside of the CI rig we have, I'm not aware of evidence that this image is used in any other way.

Also agree that this can wait for 1.20.

neolit123:
> Looks like some of this needs cleanup:
> https://github.com/kubernetes/test-infra/blob/bec22c9286285b8f177b3436508c4dade18354ac/experiment/kind-conformance-image-e2e.sh#L121-L136

yes, i've sent this yesterday:
https://github.com/kubernetes/test-infra/pull/18799

neolit123:
> test/e2e/image then?

or /test/conformance/image

/assign @spiffxp

desponda:
👋  Bug triage here, this MR has not had activity for a while, I am removing the milestone for this PR for now, please update if this needs to go into a future release

spiffxp:
Thanks for removing from v1.21

/remove-lifecycle stale
/milestone v1.22

neolit123:
doing cleanup on some of my old pending PRs. let's revisit this in the future.

PR #89908, opened 2020-04-06
Skip updating scheduler cache on pod update if the node was deleted

alculquicondor:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

/sig scheduling

Skip updating scheduler cache on pod update if the node was deleted. This can happen when events arrive out of order.
The fix doesn't skip updates on pods that might arrive before its node is created. There will always be a pod creation event first that would cause the cache to add a placeholder node.

Add unit test for out-of-order pod/node creation/removal.
This should help spot bugs when:
- Pod creation event arrives before node creation
- Node removal event arrives before corresponding pod removal.

**Which issue(s) this PR fixes**:

Fixes #89552

This only reproduced with at least 2 pods in the node being removed.

**Does this PR introduce a user-facing change?**:

```release-note
Fix scheduler crash when removing node before its pods
```

PR #86964, opened 2020-01-08
Remove nodes from cache immediately on delete events

alculquicondor:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

Remove nodes from cache's structs when getting a Remove event. Previously, we keep them in some caches until all remove pod events arrive.
This is unnecessary and it causes structs to be become out-of-sync.

We still keep pods in other intermediate caches until their delete events arrive. This is to avoid the complexity of having to handle removed elements when we receive the delete, forget or expires events (the last 2 are internal events to the scheduler, not from k8s). This is fine for the purposes of scheduling, as plugins and the generic scheduler implementation is only expected to use the node and pod listers, which use the list of nodes as source of truth.

/sig scheduling

**Which issue(s) this PR fixes**:

Fixes #86920

**Special notes for your reviewer**:

This builds on top of #86930, only review from 2nd commit.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #93938, opened 2020-08-12
Keep track of remaining pods when a node is deleted

alculquicondor:
**What type of PR is this?**

/kind bug

The apiserver is expected to send pod deletion events that might arrive at a different time. However, sometimes a node could be recreated without its pods being deleted.

This is a partial revert of #86964 and #89908

Since then, we have been more careful about direct usage of the map of nodes. In particular:

- there were no uses left of GetNodeInfo
- we skip nil node objects during snapshotting

Fixed Dump that still uses the node map (still useful to know, so we don't want to skip phantom nodes). And switched the only remaining list method that uses the map to return a count, and marked to only be used for tests.

The PR consists of 2 commits:

1. Removing direct usages of cache's map.
2. Actual fix.

```release-note
Scheduler bugfix: Scheduler doesn't lose pod information when nodes are quickly recreated. This could happen when nodes are restarted or quickly recreated reusing a nodename.
```

### DIFF
--- a/pkg/scheduler/framework/v1alpha1/types.go
+++ b/pkg/scheduler/framework/v1alpha1/types.go
@@ -625,6 +625,12 @@ func (n *NodeInfo) SetNode(node *v1.Node) error {
 	return nil
 }
 
+// RemoveNode removes the node object, leaving all other tracking information.
+func (n *NodeInfo) RemoveNode() {
+	n.node = nil
+	n.Generation = nextGeneration()
+}
+
 // FilterOutPods receives a list of pods and filters out those whose node names
 // are equal to the node of this NodeInfo, but are not found in the pods of this NodeInfo.
 //
--- a/pkg/scheduler/internal/cache/cache.go
+++ b/pkg/scheduler/internal/cache/cache.go
@@ -22,7 +22,6 @@ import (
 	"time"
 
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
@@ -315,7 +314,9 @@ func (cache *schedulerCache) removeDeletedNodesFromSnapshot(snapshot *Snapshot) 
 	}
 }
 
-func (cache *schedulerCache) ListPods(selector labels.Selector) ([]*v1.Pod, error) {
+// PodCount returns the number of pods in the cache (including those from deleted nodes).
+// DO NOT use outside of tests.
+func (cache *schedulerCache) PodCount() (int, error) {
 	cache.mu.RLock()
 	defer cache.mu.RUnlock()
 	// podFilter is expected to return true for most or all of the pods. We
@@ -325,15 +326,11 @@ func (cache *schedulerCache) ListPods(selector labels.Selector) ([]*v1.Pod, erro
 	for _, n := range cache.nodes {
 		maxSize += len(n.info.Pods)
 	}
-	pods := make([]*v1.Pod, 0, maxSize)
+	count := 0
 	for _, n := range cache.nodes {
-		for _, p := range n.info.Pods {
-			if selector.Matches(labels.Set(p.Pod.Labels)) {
-				pods = append(pods, p.Pod)
-			}
-		}
+		count += len(n.info.Pods)
 	}
-	return pods, nil
+	return count, nil
 }
 
 func (cache *schedulerCache) AssumePod(pod *v1.Pod) error {
@@ -734,19 +731,6 @@ func (cache *schedulerCache) expirePod(key string, ps *podState) error {
 	delete(cache.assumedPods, key)
 	delete(cache.podStates, key)
 	return nil
-}
-
-// GetNodeInfo returns cached data for the node name.
-func (cache *schedulerCache) GetNodeInfo(nodeName string) (*v1.Node, error) {
-	cache.mu.RLock()
-	defer cache.mu.RUnlock()
-
-	n, ok := cache.nodes[nodeName]
-	if !ok {
-		return nil, fmt.Errorf("node %q not found in cache", nodeName)
-	}
-
-	return n.info.Node(), nil
 }
 
 // updateMetrics updates cache size metric values for pods, assumed pods, and nodes
--- a/pkg/scheduler/internal/cache/cache.go
+++ b/pkg/scheduler/internal/cache/cache.go
@@ -420,13 +420,6 @@ func (cache *schedulerCache) addPod(pod *v1.Pod) {
 
 // Assumes that lock is already acquired.
 func (cache *schedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
-	if _, ok := cache.nodes[newPod.Spec.NodeName]; !ok {
-		// The node might have been deleted already.
-		// This is not a problem in the case where a pod update arrives before the
-		// node creation, because we will always have a create pod event before
-		// that, which will create the placeholder node item.
-		return nil
-	}
 	if err := cache.removePod(oldPod); err != nil {
 		return err
 	}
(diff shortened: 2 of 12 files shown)


pkg/scheduler/framework/v1alpha1/types.go
```diff
@@ -625,6 +625,12 @@ func (n *NodeInfo) SetNode(node *v1.Node) error {
 	return nil
 }
 
+// RemoveNode removes the node object, leaving any other tracking information.
```
ahg-g, line 628:
```suggestion
// RemoveNode removes the node object, leaving all other tracking information.
```

pkg/scheduler/internal/cache/cache.go
```diff
@@ -438,18 +438,22 @@ func (cache *schedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
 }
 
 // Assumes that lock is already acquired.
-// Removes a pod from the cached node info. When a node is removed, some pod
-// deletion events might arrive later. This is not a problem, as the pods in
-// the node are assumed to be removed already.
+// Removes a pod from the cached node info. If the node information was already
+// removed and there are no more pods left in the node, cleans up the pod from
```
ahg-g, line 442:
```suggestion
// removed and there are no more pods left in the node, cleans up the node from
```

pkg/scheduler/internal/cache/cache.go
```diff
@@ -438,18 +438,22 @@ func (cache *schedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
...
+// the cache.
 func (cache *schedulerCache) removePod(pod *v1.Pod) error {
 	n, ok := cache.nodes[pod.Spec.NodeName]
 	if !ok {
-		return nil
+		return fmt.Errorf("node %v not found", pod.Spec.NodeName)
```
ahg-g, line 447: why? was this how we did it in the original logic?

alculquicondor: Yes

pkg/scheduler/framework/v1alpha1/types.go
alculquicondor: Done

pkg/scheduler/internal/cache/cache.go
alculquicondor: Done

ahg-g:
/milestone v1.19
/lgtm
/hold

One thing we could do to reduce the chance of accessing nodeinfo.Node() without checking for nil is to return an error when the node object is nil (```n, err := nodeinfo.Node()```). It is less likely we miss checking the error than missing to check for nil. Just an idea that we could do in 1.20.

pkg/scheduler/internal/cache/interface.go
```diff
@@ -57,7 +57,8 @@ import (
 // - Both "Expired" and "Deleted" are valid end states. In case of some problems, e.g. network issue,
 //   a pod might have changed its state (e.g. added and deleted) without delivering notification to the cache.
 type Cache interface {
-	// ListPods lists all pods in the cache.
+	// ListPods returns all the pods in the cache (including those from deleted nodes).
+	// DO NOT use outside of tests.
```
Huang-Wei, line 61: It seems `TestSchedulerNoPhantomPodAfterExpire` in pkg/scheduler/scheduler_test.go is the only consumer of `ListPods()`. And that test only verifies if len(pods) == 0, so ideally, we can remove the signature of ListPods() by either (1) refactoring the test case, or (2) providing a function like `PodsLen()`. WDYT?

pkg/scheduler/internal/cache/cache.go
```diff
@@ -736,19 +751,6 @@ func (cache *schedulerCache) expirePod(key string, ps *podState) error {
 	return nil
 }
 
-// GetNodeInfo returns cached data for the node name.
```
Huang-Wei, line 739: I think 2 functions in `pkg/scheduler/internal/cache/fake/fake_cache.go` can also be removed:

- func (c *Cache) GetNodeInfo(nodeName string) (*v1.Node, error)
- func (c *Cache) ListNodes() []*v1.Node

Huang-Wei (commented):
some nits.

Huang-Wei:
> However, sometimes a node could be recreated without its pods being deleted.

@alculquicondor Is there any user reporting this issue? I'm curious about the real use case.

jkaniuk:
> @alculquicondor Is there any user reporting this issue? I'm curious about the real use case.

It could be triggered during GKE scalability tests and have some nasty effects - like high churn of deployments pods creation which are constantly created, assigned to nodes, kicked out out nodes with OutOfcpu, never reaching desired number of replicas.

/priority critical-urgent

pkg/scheduler/internal/cache/interface.go
alculquicondor: Went with 2.

pkg/scheduler/internal/cache/cache.go
alculquicondor: Done

alculquicondor:
I'm also reverting #89908, which only makes sense if #86964 is there.

pkg/scheduler/internal/cache/interface.go
```diff
@@ -57,8 +57,9 @@ import (
 // - Both "Expired" and "Deleted" are valid end states. In case of some problems, e.g. network issue,
 //   a pod might have changed its state (e.g. added and deleted) without delivering notification to the cache.
 type Cache interface {
-	// ListPods lists all pods in the cache.
-	ListPods(selector labels.Selector) ([]*v1.Pod, error)
+	// ListPods returns the number of pods in the cache (including those from deleted nodes).
```
Huang-Wei, line 60:
```suggestion
	// PodCount returns the number of pods in the cache (including those from deleted nodes).
```

pkg/scheduler/scheduler_test.go
```diff
@@ -527,12 +527,12 @@ func TestSchedulerNoPhantomPodAfterExpire(t *testing.T) {
 				return
 			default:
 			}
-			pods, err := scache.ListPods(labels.Everything())
+			pods, err := scache.PodCount(labels.Everything())
```
Huang-Wei, line 530: pods -> podsNum

ahg-g:
it would be nice if we can keep the actual fix in a separate commit separate from all the other updates to the tests.

pkg/scheduler/internal/cache/interface.go
alculquicondor: Done

alculquicondor:
> it would be nice if we can keep the actual fix in a separate commit separate from all the other updates to the tests.

Done

pkg/scheduler/internal/cache/cache.go
```diff
@@ -435,18 +428,23 @@ func (cache *schedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
...
-// Removes a pod from the cached node info. When a node is removed, some pod
-// deletion events might arrive later. This is not a problem, as the pods in
-// the node are assumed to be removed already.
+// Removes a pod from the cached node info. If the node information was already
+// removed and there are no more pods left in the node, cleans up the node from
+// the cache.
 func (cache *schedulerCache) removePod(pod *v1.Pod) error {
 	n, ok := cache.nodes[pod.Spec.NodeName]
 	if !ok {
+		klog.Errorf("node %v not found when trying to remove pod %v", pod.Spec.NodeName, pod.Name)
```
Huang-Wei, line 437: I recalled the original logic returned an error?

Huang-Wei:
Just a nit ^^. LGTM otherwise.

pkg/scheduler/internal/cache/cache.go
alculquicondor: It did. But returning nil is actually safer in the case of extraneous update events that might arrive before a node is created, and after the original node was completely removed.

ahg-g:
> > it would be nice if we can keep the actual fix in a separate commit separate from all the other updates to the tests.
> 
> Done

Thanks.

LGTM, will leave it to Wei to officially lgtm.

pkg/scheduler/internal/cache/cache.go
ahg-g: +1 to returning nil and just logging an error.

Huang-Wei: I checked the usage of `removePod()`, there are still a number of callers rely on the returned value. So I'd suggest to revert to the original state.

alculquicondor: Detail for each caller:

- ForgetPod: We actually want to proceed and clear the assumedPods and podStates.
- expirePod: Same as above.
- AddPod: it just logs the error returned, so same effect.
- RemovePod: We want to clear podStates.
- updatePod: This is the case where we want to prevent losing information.

alculquicondor: That said, for expirePod and ForgetPod, the node shouldn't have been removed because pods were already assigned.

Huang-Wei: Thanks. That's fair.

alculquicondor:
This is actually already merged, but the github UI is outdated.

/shrug

PR #93940, opened 2020-08-12
get: -o yaml, json set ServerPrint false

sallyom:
**What type of PR is this?**
/kind cleanup

Currently, with `kubectl get resource -o yaml|json|jsonpath=..|jsonpath-file=..|custom-columns=.. -v=2`  this warning is shown:
```
I0812 14:41:47.938840   81833 table_printer.go:45] Unable to decode server response into a Table. Falling back to hardcoded types: attempt to decode non-Table object
```
Instead, should set ServerPrint=false for those outputOptions (this PR) or, in the least, move that log level up - it's a 2 now.
Also, the current check to set ServerPrint=false if -o == "custom-columns" is flawed, need to check if contains, since users pass `custom-columns=...` 

```release-note
NONE
```


soltysh (approved):
/lgtm
/approve
/priority important-longterm

PR #89314, opened 2020-03-20
drain: ignore forbidden errors on drain

rphillips:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Drain is erroring when it is running and a namespace gets deleted. Drain should not be erroring in this case and we can safely ignore the error.

Once the namespace is marked to be deleted, the drain code can create a new eviction. This new eviction will be denied by the admissions controller with a forbidden message, because creating a new object within the namespace is forbidden. 

This fix will allow Drain() to run to completion without errors.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:
```
NONE
```

/cc @sjenning @smarterclayton

PR #93947, opened 2020-08-12
Fix drain forbidden

michaelgugino:
**What type of PR is this?**

/kind bug

Drain is returning an error when it is running and a namespace gets deleted. Drain should not be return error in this case and we can safely ignore the error.

Once the namespace is marked to be deleted, the drain code can create a new eviction. This new eviction will be denied by the admissions controller with a forbidden message, because creating a new object within the namespace is forbidden.

This fix will allow Drain() to run to completion without errors.

Fixes #

Replaces: https://github.com/kubernetes/kubernetes/pull/89314

```release-note
NONE
```

### DIFF
--- a/staging/src/k8s.io/kubectl/pkg/drain/drain.go
+++ b/staging/src/k8s.io/kubectl/pkg/drain/drain.go
@@ -254,6 +254,7 @@ func (d *Helper) evictPods(pods []corev1.Pod, policyGroupVersion string, getPodF
 	defer cancel()
 	for _, pod := range pods {
 		go func(pod corev1.Pod, returnCh chan error) {
+			refreshPod := false
 			for {
 				switch d.DryRunStrategy {
 				case cmdutil.DryRunServer:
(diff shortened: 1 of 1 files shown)


knight42 (commented):
/ok-to-test

staging/src/k8s.io/kubectl/pkg/drain/drain.go
```diff
@@ -277,8 +277,9 @@ func (d *Helper) evictPods(pods []corev1.Pod, policyGroupVersion string, getPodF
 				} else if apierrors.IsTooManyRequests(err) {
 					fmt.Fprintf(d.ErrOut, "error when evicting pod %q (will retry after 5s): %v\n", pod.Name, err)
 					time.Sleep(5 * time.Second)
-				} else if apierrors.IsForbidden(err) {
-					// an eviction request in a deleting namespace will throw a forbidden error
+				} else if !pod.ObjectMeta.DeletionTimestamp.IsZero() && apierrors.IsForbidden(err) {
```
soltysh, line 280: `DeletionTimestamp` is pointier which means can be nil so I'd consider having check for nil too.

soltysh (changes requested):

staging/src/k8s.io/kubectl/pkg/drain/drain.go
michaelgugino: @soltysh 
I know what you're saying, but golang is weird here: https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/time.go#L61

DeletionTimestamp.IsZero() is used to test nil here.  I think the method is a little redundant, but seems to be common practice in k8s.

soltysh: :+1: thx for pointing this out.

soltysh (approved):
/lgtm
/approve

staging/src/k8s.io/kubectl/pkg/drain/drain.go
alejandrox1: For this bit, there was a comment in the first PR about identifying the cause of the forbidden: should we maybe add a
```go
if apierrors.HasStatusCause(err, v1.NamespaceTerminatingCause)
```
?

alejandrox1: ref: https://github.com/kubernetes/kubernetes/pull/89314#discussion_r395869753

michaelgugino: Does 'NamespaceTerminatingCause' actually exist anywhere?

Here's all the metav1 causes: https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L883

alejandrox1: ha yeah, went through that search as well.
`NamespaceTerminatingCause` is defined in https://godoc.org/k8s.io/api/core/v1

alejandrox1: /hold
to follow up on https://github.com/kubernetes/kubernetes/pull/89314#discussion_r395869753
(at some point the master branch is going to be opened to PRs without milestone 1.20 in them)
/cc @smarterclayton 
but if yall think this is good as is please remove the hold

michaelgugino: Okay, let me add that...

staging/src/k8s.io/kubectl/pkg/drain/drain.go
```diff
@@ -277,6 +277,16 @@ func (d *Helper) evictPods(pods []corev1.Pod, policyGroupVersion string, getPodF
...
+				} else if !pod.ObjectMeta.DeletionTimestamp.IsZero() && apierrors.IsForbidden(err) && apierrors.HasStatusCause(err, corev1.NamespaceTerminatingCause) {
+					// an eviction request in a deleting namespace will throw a forbidden error,
+					// if the pod is already marked deleted, we can ignore this error, an eviction
+					// request will never succeed, but we will waitForDelete for this pod.
+					break
+				} else if apierrors.IsForbidden(err) && apierrors.HasStatusCause(err, corev1.NamespaceTerminatingCause) {
+					// an eviction request in a deleting namespace will throw a forbidden error,
+					// if the pod is not marked deleted, we retry until it is.
+					fmt.Fprintf(d.ErrOut, "error when evicting pod %q (will retry after 5s): %v\n", pod.Name, err)
+					time.Sleep(5 * time.Second)
```
alejandrox1, line 289: Should the pod also be refreshed here?
i.e., we start iterating over the pods here, we try to evict one but it fails because its namepace is being terminated, wouldn't this pod not have a non-nill (non-zero) deletionTimestaamp yet (and thus we would need to get the pod again after waiting for some 5 seconds)?

michaelgugino: Yes, great point.  I'll add that!

soltysh (approved):
I think points from @alejandrox1 are addressed
/lgtm
/hold cancel
/milestone v1.20

PR #93044, opened 2020-07-13
Initialize scheduler's podInformer in sharedInformerFactory

Huang-Wei:
**What type of PR is this?**

/kind cleanup
/sig scheduling

**What this PR does / why we need it**:

Scheduler used to initialize its podInfomer with specific filtering conditions: phase != Succeeded && phase != Failed. However, this caused a fragmentation with other informers - which are managed and initialized from the sharedInformerFactory.

Per https://github.com/kubernetes/kubernetes/pull/92801#discussion_r453553536, this PR uses `InformerFor` to initialize the podInfomer in-place.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

Some integration tests just initialize the podInfomer using `informerFactory.Core().V1().Pods()`, so they are not impacted. If an integration test wants to use a totally different podInformer, usually that needs a separate informerFactory.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #93959, opened 2020-08-13
Turn off unnecessary informers startup

Mr-Linus:
**What type of PR is this?**
/kind cleanup

If we directly use `cc.InformerFactory.Start(ctx.Done())` to start, it will cause many informers that are not related to the scheduler to start, such as:
- v1.StatefulSet
- v1.ReplicationController
...

But these informers have nothing to do with the scheduler logic and do not need to start these informers.

Moreover, `cc.InformerFactory` starts a redundant pod informer at the same time. So, when there are too many pods and workloads in the cluster, this causes a waste of memory.

We only need to start some infomer necessary for the scheduler.

NONE

```release-note
NONE
```

### DIFF
--- a/cmd/kube-scheduler/app/server.go
+++ b/cmd/kube-scheduler/app/server.go
@@ -186,7 +186,7 @@ func Run(ctx context.Context, cc *schedulerserverconfig.CompletedConfig, sched *
 
 	// Start all informers.
 	go cc.PodInformer.Informer().Run(ctx.Done())
-	cc.InformerFactory.Start(ctx.Done())
+	startInformers(cc.InformerFactory, ctx.Done())
 
 	// Wait for all caches to sync before scheduling.
 	cc.InformerFactory.WaitForCacheSync(ctx.Done())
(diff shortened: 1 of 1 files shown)


cmd/kube-scheduler/app/server.go
```diff
@@ -282,6 +282,16 @@ func getRecorderFactory(cc *schedulerserverconfig.CompletedConfig) profile.Recor
 		return cc.EventBroadcaster.NewRecorder(name)
 	}
 }
+// startInformers starts the informers that the scheduler needed.
+func startInformers(factory informers.SharedInformerFactory,stopChan <- chan struct{}){
+	// Start Informers
+	go factory.Core().V1().Pods().Informer().Run(stopChan)
```
adtac, line 288: I don't know whether this reduced set of informers is sufficient, but the `sharedInformerFactory.Start` function does more than just start goroutines:

https://github.com/kubernetes/kubernetes/blob/b1b93e30131986319f1d46a62771feb9c0f097ae/staging/src/k8s.io/client-go/informers/factory.go#L127

It first obtains a lock. Then it checks and sets `startedInformers` too.

Besides, doesn't the list of informers to start depend on the options provided?

adtac: Also what was the full set of informers and why is this reduced set sufficient? i.e. what informers are being removed?

Huang-Wei:
> it will cause many informers that are not related to the scheduler to start, such as:
> 
> * v1.StatefulSet
> * v1.ReplicationController
>   ...

If xyz.Informer() or xyz.Lister() is not called, my understanding is it won't result in extra usage on memory, isn't it?

In scheduler's case, only particular API objects get instantiated, there is no wasting on starting unneeded informers.

> Moreover, `cc.InformerFactory` starts a redundant pod informer at the same time.

Not really, it can only happen when the (default) podInformer's Informer() or Lister() function gets called. However, we have a PR standardizing the usage of podInformer: #93044.

cmd/kube-scheduler/app/server.go
Huang-Wei: +1 to @adtac. I don't think this is the right direction. No need to mention that the out-of-tree plugins can instantiate whatever XYZInformer they want.

Also my understanding is that there is wasting on starting unneeded informers: https://github.com/kubernetes/kubernetes/pull/93959#issuecomment-673569456

Mr-Linus:
Understand, thx

PR #93960, opened 2020-08-13
fix vendor/k8s.io/apimachinery/pkg/util/framer staticcheck

habibrosyad:
**What type of PR is this?**

/kind cleanup

Fixes `vendor/k8s.io/apimachinery/pkg/util/framer` staticcheck errors.

Part of #92402

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/util/framer/framer.go
+++ b/staging/src/k8s.io/apimachinery/pkg/util/framer/framer.go
@@ -132,12 +132,14 @@ func (r *jsonFrameReader) Read(data []byte) (int, error) {
 	// Return whatever remaining data exists from an in progress frame
 	if n := len(r.remaining); n > 0 {
 		if n <= len(data) {
+			//lint:ignore SA4006,SA4010 underlying array of data is modified here.
 			data = append(data[0:0], r.remaining...)
 			r.remaining = nil
 			return n, nil
 		}
 
 		n = len(data)
+		//lint:ignore SA4006,SA4010 underlying array of data is modified here.
 		data = append(data[0:0], r.remaining[:n]...)
 		r.remaining = r.remaining[n:]
 		return n, io.ErrShortBuffer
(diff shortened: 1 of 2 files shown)


staging/src/k8s.io/apimachinery/pkg/util/framer/framer.go
```diff
@@ -132,13 +132,11 @@ func (r *jsonFrameReader) Read(data []byte) (int, error) {
 	// Return whatever remaining data exists from an in progress frame
 	if n := len(r.remaining); n > 0 {
 		if n <= len(data) {
-			data = append(data[0:0], r.remaining...)
```
neolit123, line 135: curious what the staticcheck error is here?

these functions are supposed to append to `data` it seems, so i don't think this change is valid.

habibrosyad: The error is `this value of data is never used` ([SA4006](https://staticcheck.io/docs/checks#SA4006)). Yeah, it's appending to `data` but it doesn't change anything basically. https://github.com/kubernetes/kubernetes/blob/b1b93e30131986319f1d46a62771feb9c0f097ae/staging/src/k8s.io/apimachinery/pkg/util/framer/framer.go#L135-L137. In this block, it's returning immediately.

roycaihw: slices are passed by reference header, so updating the parameter does affect the actual `data` feed to the function https://play.golang.org/p/NB7h_e02y9s

habibrosyad: My mistake, you're right. It's false positive for staticcheck in this case. That line is basically just replacing the underlying array values with `r.remaining`. The condition `n <= len(data)` make sure that append doesn't create a new slice if the cap is insufficient. It looks like statickcheck failed to detect this and assume it does create a new slice (which never used).

Should this PR be closed or ignoring the linter is preferable?

liggitt: @roycaihw excellent catch, thanks

let's add explicit `//lint:ignore...` to the relevant lines with accompanying comments so this isn't attempted again in the future

habibrosyad: @liggitt updated

PR #93961, opened 2020-08-13
Refactor kubectl stdin test

### LINKED ISSUE #73099: The test "Simple pod should support inline execution and attach" is inherently racy.
The test case "executing a command with run and attach without stdin" is inherently racy.

The test runs`kubectl run run-test --image=docker.io/library/busybox:1.29 --restart=OnFailure --attach=true --stdin -- sh -c cat && echo 'stdin closed'`, and expected to see `stdin closed` in the output.

The case when the test passes:
* case 1:
  * Kubernetes creates and starts the container without stdin;
  * The container exits before attach runs;
  * kubectl falls back to get container log, which contains `stdin closed`. https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/run/run.go#L527
* case 2:
  * Kubernetes creates and starts the container without stdin;
  * kubectl attach before "stdin closed" is printed;
  * kubectl attach contains `stdin closed`.

The case when the test fails:
  * Kubernetes creates and starts the container without stdin;
  * The container prints "stdin closed", but hasn't exited yet;
  * kubectl attach contains nothing.

One example failure: https://gubernator.k8s.io/build/kubernetes-jenkins/logs/ci-containerd-e2e-ubuntu-gce/7057

Related issue: https://github.com/kubernetes/kubernetes/issues/24070

Can we modify this test not to check the output? Or use `kubectl log` to check output instead?

/cc @kubernetes/sig-node-bugs @kubernetes/sig-cli-bugs

aledbf:
**What type of PR is this?**

/kind failing-test
/kind flake

Use `kubectl log` to check output instead of the output from `kubectl run`

Fixes #73099

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/kubectl/kubectl.go
+++ b/test/e2e/kubectl/kubectl.go
@@ -558,11 +558,24 @@ var _ = SIGDescribe("Kubectl client", func() {
 		})
 
 		ginkgo.It("should support inline execution and attach", func() {
+			waitForStdinContent := func(pod, content string) string {
+				var logOutput string
+				err := wait.Poll(10*time.Second, 5*time.Minute, func() (bool, error) {
+					logOutput = framework.RunKubectlOrDie(ns, "logs", pod)
+					return strings.Contains(logOutput, content), nil
+				})
+
+				gomega.Expect(err).To(gomega.BeNil(), fmt.Sprintf("unexpected error waiting for '%v' output", content))
+				return logOutput
+			}
+
 			ginkgo.By("executing a command with run and attach with stdin")
 			// We wait for a non-empty line so we know kubectl has attached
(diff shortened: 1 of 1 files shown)


hasheddan (commented):
/milestone v1.19

aledbf:
/retest

(the kind failures are not related to the change in the PR)

BenTheElder:
I guess that means we no longer test attach in this sub-case? (IIRC log is a different call) cc @liggitt 

Super happy to see an approach to fixing this, was just trying to come up with a reasonable option...

hasheddan:
@janetkuo @soltysh could I get one of you to take a look here? :)

liggitt:
this is not release blocking, if we get a fix we want to cherry-pick, we can do that

soltysh:
> I guess that means we no longer test attach in this sub-case? (IIRC log is a different call) cc @liggitt

No, we're still attaching b/c that's how we inject data into stdin, the only change here is how we verify the output from that container.

soltysh (approved):
/lgtm
/approve
Thanks @aledbf

PR #93963, opened 2020-08-13
Fixes variable expansion false positive

### LINKED ISSUE #94006: Variable expansion test false positive
**Which jobs are failing**:

N/A

    should fail substituting values in a volume subpath with absolute path

N/A

**Testgrid link**:

N/A

**Reason for failure**:

Some of the tests are negative test cases which are supposed to ensure that those invalid usecases are handled properly.

However, some of the tests are false positives, they can pass due to various reasons. One such example is: should fail substituting values in a volume subpath with absolute path.
This test can pass if:

- the Pod cannot start due to various reasons (e.g.: the container image cannot be pulled or does not exist).
- the Pod ran to completion, even though the container was not supposed to start in the first place.

/sig testing

Related: #94007

claudiubelu:
**What type of PR is this?**

/kind bug

/sig testing
/sig windows

Some of the tests are negative test cases which are supposed to ensure that those invalid usecases are handled properly.

However, some of the tests are false positives, they can pass due to various reasons. One such example is: ``should fail substituting values in a volume subpath with absolute path``.
This test can pass if:

- the Pod cannot start due to various reasons (e.g.: the container image cannot be pulled or does
  not exist).
- the Pod ran to completion, even though the container was not supposed to start in the first place.

``filepath.IsAbs`` does not consider ``/`` or ``\`` prefixed paths on Windows as absolute, even though in the end they are treated as such. Because of this, absolute paths can be mounted as volume subpaths in Windows.

The test ``should fail substituting values in a volume subpath with absolute path`` previously passed on Windows because of a false positive in the test: the test did not specifically check that the container failed to be created, instead it checked if it was not Running (in most cases, the container was running to Completion instead, which is indeed not Running, but not correct).

Fixes #94006

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/framework/pod/resource.go
+++ b/test/e2e/framework/pod/resource.go
@@ -298,6 +298,29 @@ func podsRunning(c clientset.Interface, pods *v1.PodList) []error {
 	return e
 }
 
+func podContainerFailed(c clientset.Interface, namespace, podName string, containerIndex int, reason string) wait.ConditionFunc {
+	return func() (bool, error) {
+		pod, err := c.CoreV1().Pods(namespace).Get(context.TODO(), podName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		switch pod.Status.Phase {
+		case v1.PodPending:
+			if len(pod.Status.ContainerStatuses) == 0 {
+				return false, nil
+			}
+			containerStatus := pod.Status.ContainerStatuses[containerIndex]
+			if containerStatus.State.Waiting != nil && containerStatus.State.Waiting.Reason == reason {
+				return true, nil
+			}
+			return false, nil
+		case v1.PodFailed, v1.PodRunning, v1.PodSucceeded:
+			return false, fmt.Errorf("pod was expected to be pending, but it is in the state: %s", pod.Status.Phase)
+		}
+		return false, nil
+	}
+}
+
 // LogPodStates logs basic info of provided pods for debugging.
 func LogPodStates(pods []v1.Pod) {
 	// Find maximum widths for pod, node, and phase strings for column printing.
--- a/test/e2e/common/expansion.go
+++ b/test/e2e/common/expansion.go
@@ -181,11 +181,16 @@ var _ = framework.KubeDescribe("Variable Expansion", func() {
 		Description: Make sure a container's subpath can not be set using an expansion of environment variables when absolute path is supplied.
 	*/
 	framework.ConformanceIt("should fail substituting values in a volume subpath with absolute path [sig-storage][Slow]", func() {
+		absolutePath := "/tmp"
+		if framework.NodeOSDistroIs("windows") {
+			// Windows does not typically have a C:\tmp folder.
+			absolutePath = "C:\\Users"
+		}
 
 		envVars := []v1.EnvVar{
 			{
 				Name:  "POD_NAME",
-				Value: "/tmp",
+				Value: absolutePath,
 			},
 		}
 		mounts := []v1.VolumeMount{
--- a/test/e2e/common/expansion.go
+++ b/test/e2e/common/expansion.go
@@ -367,8 +367,8 @@ func testPodFailSubpath(f *framework.Framework, pod *v1.Pod) {
 		e2epod.DeletePodWithWait(f.ClientSet, pod)
 	}()
 
-	err := e2epod.WaitTimeoutForPodRunningInNamespace(f.ClientSet, pod.Name, pod.Namespace, framework.PodStartShortTimeout)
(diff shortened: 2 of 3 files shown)


test/e2e/framework/pod/resource.go
```diff
@@ -298,6 +298,26 @@ func podsRunning(c clientset.Interface, pods *v1.PodList) []error {
 	return e
 }
 
+func podContainerFailed(c clientset.Interface, namespace, podName string, containerIndex int, reason string) wait.ConditionFunc {
```
ZhiFeng1993, line 301: What is this containerIndex user for?

test/e2e/framework/pod/resource.go
```diff
@@ -298,6 +298,26 @@ func podsRunning(c clientset.Interface, pods *v1.PodList) []error {
...
+	return func() (bool, error) {
+		pod, err := c.CoreV1().Pods(namespace).Get(context.TODO(), podName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		switch pod.Status.Phase {
+		case v1.PodPending:
+			containerStatus := pod.Status.ContainerStatuses[containerIndex]
```
oomichi, line 309: This part makes an error like
```
test/e2e/framework/framework.go:597
Test Panicked
staging/src/k8s.io/apimachinery/pkg/util/runtime/runtime.go:55

Panic: runtime error: index out of range [0] with length 0

Full stack:
k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	staging/src/k8s.io/apimachinery/pkg/util/runtime/runtime.go:55 +0x10c
panic(0x486a820, 0xc003066c00)
	GOROOT/src/runtime/panic.go:969 +0x175
k8s.io/kubernetes/test/e2e/framework/pod.podContainerFailed.func1(0xc00299b360, 0x7f6fc02cb7f0, 0xc001c75440)
	test/e2e/framework/pod/resource.go:309 +0x3c6
```

oomichi: `pod.Status.ContainerStatueses` can be length 0.
So it is better to add the following check before current line 309
```
if len(status.ContainerStatuses) == 0 {
    return false, nil
}
```

oomichi (commented):
/cc @oomichi

test/e2e/framework/pod/resource.go
claudiubelu: good point, I've included this.

test/e2e/framework/pod/resource.go
claudiubelu: I was considering the possibility that a test could spawn a pod with multiple containers, so this would basically specify which one to check for. Niche usecase though.

oomichi:
Thanks for updating.
The test part looks good for me.

/lgtm
/approve

pkg/kubelet/kubelet_pods.go
```diff
@@ -185,7 +185,10 @@ func makeMounts(pod *v1.Pod, podDir string, container *v1.Container, hostName, h
 				return nil, cleanupAction, fmt.Errorf("volume subpaths are disabled")
 			}
 
-			if filepath.IsAbs(subPath) {
+			// On Windows, filepath.IsAbs does not consider \ or / prefixed paths as absolute, even though
```
marosset, line 188: I think this check is also performed at spec validation which should currently only ever happen on code compiled for linux.

https://github.com/kubernetes/kubernetes/blob/544b74c2cb94fce6d89019b85c7c891050df2af7/pkg/apis/core/validation/validation.go#L2427-L2430

calls

https://github.com/kubernetes/kubernetes/blob/544b74c2cb94fce6d89019b85c7c891050df2af7/pkg/apis/core/validation/validation.go#L1158-L1170

pkg/kubelet/kubelet_pods.go
```diff
@@ -185,7 +185,10 @@ func makeMounts(pod *v1.Pod, podDir string, container *v1.Container, hostName, h
...
+			// they are treated as such.
+			isWindowsAbs := runtime.GOOS == "windows" && (strings.HasPrefix(subPath, `/`) || strings.HasPrefix(subPath, `\`))
```
SergeyKanzhelev, line 190: should this logic has a separate method for validation? In a PR description I see something about other place having (or needing to have) the same logic. Would be great to cross-reference those or use the same method in both cases.

> even though in the end they are treated as such. Because of this, absolute paths can be mounted as volume subpaths in Windows.

claudiubelu:
A milestone should be added to this.

justaugustus:
ref: https://github.com/kubernetes/kubernetes/issues/94263

PR #93967, opened 2020-08-13
Update validation.go

jayunit100:
**What type of PR is this?**
/kind cleanup 

fix the env var so that it has the right values that need to get plumbed into ginkgo and so on.  also I think KUBENERETES_MASTER is going away anyways per new lang. guidelines 
```release-note
     replacing of KUBERNETES_MASTER error message from client-go with correct variables
```

### DIFF
--- a/staging/src/k8s.io/client-go/tools/clientcmd/validation.go
+++ b/staging/src/k8s.io/client-go/tools/clientcmd/validation.go
@@ -30,7 +30,7 @@ import (
 
 var (
 	ErrNoContext   = errors.New("no context chosen")
-	ErrEmptyConfig = NewEmptyConfigError("no configuration has been provided, try setting KUBERNETES_SERVICE_HOST or KUBECONFIG environment variable")
+	ErrEmptyConfig = NewEmptyConfigError("no configuration has been provided, try setting KUBERNETES_SERVICE_HOST, KUBECONFIG, or KUBERENTES_MASTER environment variable")
 	// message is for consistency with old behavior
 	ErrEmptyCluster = errors.New("cluster has no server defined")
 )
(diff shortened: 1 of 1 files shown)


justaugustus (commented):
@jayunit100 -- Great to see this from a WG Naming perspective!

Are we sure we can actually do this today though?
There are a few places that reference this variable name: https://cs.k8s.io/?q=KUBERNETES_MASTER&i=nope&files=&repos=

/hold
/wg naming

liggitt:
yeah, thanks for taking a look, but we can't change the message before the underlying code is actually resolved

jayunit100:
hm i thought  the only actual ref in client-go is  for the deprecated `func getDefaultServer() string ` ?

jayunit100:
To be clear WG naming was a positive but,   its a very misleading as a message also, bc KUBERNETES_MASTER is deprecated and KUBECONFIG is a sufficient replacement.... im not sure theres a use case where someone actually will be able to benefit from following those directions, unless they are working on some old build tooling --- and those users obviously know what theyre doing so dont need the message.



in any case --- added KUBERNETES_MASTER at the end, since it generally i think works w/ KUBECONFIG at least in, for example, ginkgo invocations... so the correct message is actually to indicate the other possible valid env var inputs, at the very least.

but happy to close if im missing something

liggitt:
> hm i thought the only actual ref in client-go is for the deprecated `func getDefaultServer() string ` ?

it is, but it is still the active envvar that allows controlling the server name.

KUBERNETES_SERVICE_HOST will not override the default server name, it is only effective in in-cluster configurations (where /var/run/kubernetes.io/serviceaccount/token is also present, and KUBERNETES_SERVICE_PORT is also set)

I think @soltysh is looking into removing the default server name entirely

jayunit100:
ahhh gotcha i guess i should close this

PR #93971, opened 2020-08-13
Refactor locks logic on registeredNodesLock to be non-blocking

### LINKED ISSUE #93973: vSphere create volume hangs forever on resource lock
<!-- Please use this template while reporting a bug and provide as much info as possible. Not doing so may result in your bug not being addressed in a timely manner. Thanks!

If the matter is security related, please disclose it privately via https://kubernetes.io/security/
-->


**What happened**:
Quote @gnufied:

I took goroutine dump of controller-managers that were stuck in volume provisioning and I think we have a deadlock in code:
```
goroutine 594 [semacquire, 79 minutes]:                                                                                                       
sync.runtime_SemacquireMutex(0xc000e7826c, 0x40d400, 0x0)                                                                                     
        /opt/rh/go-toolset-1.14/root/usr/lib/go-toolset-1.14-golang/src/runtime/sema.go:71 +0x47                                                                                                                                                                                            
sync.(*RWMutex).RLock(...)                                                                                                                                                                                                                                                                  
        /opt/rh/go-toolset-1.14/root/usr/lib/go-toolset-1.14-golang/src/sync/rwmutex.go:50                                    
k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere.(*NodeManager).GetNode(0xc000e78240, 0xc000d4a320, 0x1a, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere/nodemanager.go:253 +0x158      
k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere.(*NodeManager).RediscoverNode(0xc000e78240, 0xc000d4a320, 0x1a, 0x0, 0xc001e825c0)
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere/nodemanager.go:244 +0x9c                                                                                                                                              
k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere.(*NodeManager).getRefreshedNodeInfo(0xc000e78240, 0xc000d4a320, 0x1a, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere/nodemanager.go:327 +0x458
k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere.(*NodeManager).refreshNodes(0xc000e78240, 0x0, 0x0, 0x0)        
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere/nodemanager.go:311 +0x166                                                                          
k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere.(*VSphere).refreshNodesForSecretChange(0xc000b6cb00)                         
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere/vsphere.go:1566 +0x33
k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere.(*VSphere).SecretAdded(0xc000b6cb00, 0x457fcc0, 0xc00252a080)                
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere/vsphere.go:1538 +0xe4
k8s.io/kubernetes/vendor/k8s.io/client-go/tools/cache.ResourceEventHandlerFuncs.OnAdd(...)                              
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/client-go/tools/cache/controller.go:227
k8s.io/kubernetes/vendor/k8s.io/client-go/tools/cache.(*processorListener).run.func1()                                   
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/client-go/tools/cache/shared_informer.go:777 +0x129
k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1(0xc001398f60)
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:155 +0x5f
k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait.BackoffUntil(0xc001eabf60, 0x4dc5bc0, 0xc001000750, 0x3c22501, 0xc0000b6d80)
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:156 +0xa3
k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil(0xc001398f60, 0x3b9aca00, 0x0, 0xc00049c501, 0xc0000b6d80)
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133 +0x98
--                                  
                                                                                                                                              
goroutine 14153 [semacquire, 76 minutes]:
sync.runtime_SemacquireMutex(0xc000e78264, 0x0, 0x1)
        /opt/rh/go-toolset-1.14/root/usr/lib/go-toolset-1.14-golang/src/runtime/sema.go:71 +0x47
sync.(*Mutex).lockSlow(0xc000e78260)
        /opt/rh/go-toolset-1.14/root/usr/lib/go-toolset-1.14-golang/src/sync/mutex.go:138 +0xfc
sync.(*Mutex).Lock(...)
        /opt/rh/go-toolset-1.14/root/usr/lib/go-toolset-1.14-golang/src/sync/mutex.go:81
sync.(*RWMutex).Lock(0xc000e78260)
        /opt/rh/go-toolset-1.14/root/usr/lib/go-toolset-1.14-golang/src/sync/rwmutex.go:98 +0x97
k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere.(*NodeManager).GetNodeDetails(0xc000e78240, 0x0, 0x0, 0x0, 0x0, 0x0)
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere/nodemanager.go:291 +0x80
k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere.getSharedDatastoresInK8SCluster(0x4e3eae0, 0xc00b2cacc0, 0xc000e78240, 0x0, 0x45ec3f8, 0x14, 0xc00b4d2e70, 0x1)
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere/vsphere_util.go:130 +0x43
k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere.(*VSphere).CreateVolume.func1(0xc00b2dec80, 0x0, 0x0, 0x0, 0x0)
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere/vsphere.go:1377 +0x17a1
k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere.(*VSphere).CreateVolume(0xc000b6cb00, 0xc00b2dec80, 0x74db420, 0x0, 0x1, 0xc00b2a9f50)
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/legacy-cloud-providers/vsphere/vsphere.go:1453 +0xe4
k8s.io/kubernetes/pkg/volume/vsphere_volume.(*VsphereDiskUtil).CreateVolume(0x74d9b38, 0xc00b28be30, 0x0, 0x74d9b38, 0x0, 0x0, 0xc00b417740, 0x2, 0x2)
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/pkg/volume/vsphere_volume/vsphere_volume_util.go:142 +0xbf2
k8s.io/kubernetes/pkg/volume/vsphere_volume.(*vsphereVolumeProvisioner).Provision(0xc00b28be30, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc00b2a9f50, 0x28)
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/pkg/volume/vsphere_volume/vsphere_volume.go:382 +0x295
k8s.io/kubernetes/pkg/controller/volume/persistentvolume.(*PersistentVolumeController).provisionClaimOperation(0xc0008e21c0, 0xc00b3476c0, 0x7fdc57537c20, 0xc004b99f40, 0xc007ede600, 0x45c78b5, 0x9, 0x0, 0x0)
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/pkg/controller/volume/persistentvolume/pv_controller.go:1555 +0xfec
k8s.io/kubernetes/pkg/controller/volume/persistentvolume.(*PersistentVolumeController).provisionClaim.func1(0x99e752, 0xc008590180)
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/pkg/controller/volume/persistentvolume/pv_controller.go:1447 +0x204
k8s.io/kubernetes/pkg/util/goroutinemap.(*goRoutineMap).Run.func1(0xc000e694d0, 0xc00b372740, 0x40, 0xc00b37ca50, 0x0, 0x0)
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/pkg/util/goroutinemap/goroutinemap.go:115 +0xe5
created by k8s.io/kubernetes/pkg/util/goroutinemap.(*goRoutineMap).Run
        /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/pkg/util/goroutinemap/goroutinemap.go:108 +0x2e7
```

I think we have a problem introduced when we merged - https://github.com/kubernetes/kubernetes/pull/90836 and what is happening is, when `GetNodeDetails` in `nodemanager.go` is called it acquires a lock on `registeredNodesLock`, but same function ends up calling `GetNode` which acquires another lock on `registeredNodesLock` object . Now the problem is, `GetNode` function gets called in a loop and it will try to acquire lock for each node it processes and hence two operations could result in deadlock.

This results into inability to provision a `PVC` with a default `thin` StorageClass. The volume does not get to call the API endpoint, just hangs after `CreateVolume` call, which is clear in the goroutine dump.

1) Provision a `PVC` from `thin` `SC` successfully.
2) Don't be blocked by credentials secret change.

1) Provision a vSphere cluster with credentials secret set
2) Create volume
3) Wait...

**Environment**:
- Kubernetes version (use `kubectl version`): v1.19.0-rc.2+5241b27-dirty
- Cloud provider or hardware configuration: vSphere
- OS (e.g: `cat /etc/os-release`): Linux
- Kernel (e.g. `uname -a`): GNU/Linux
- Install tools: openshift-install (kubeadm)
- Others:

Danil-Grigorev:
**What type of PR is this?**
/kind bug
/kind regression

Refactor locks on `GetNodeDetails` and `refreshNodes` calls.

This lock was blocking volume creation and destroying the ability to provision
volume with `thin` SC.

Fixes https://github.com/kubernetes/kubernetes/issues/93973

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/legacy-cloud-providers/vsphere/nodemanager.go
+++ b/staging/src/k8s.io/legacy-cloud-providers/vsphere/nodemanager.go
@@ -259,6 +259,16 @@ func (nm *NodeManager) GetNode(nodeName k8stypes.NodeName) (v1.Node, error) {
 	return *node, nil
 }
 
+func (nm *NodeManager) getNodes() map[string]*v1.Node {
+	nm.registeredNodesLock.RLock()
+	defer nm.registeredNodesLock.RUnlock()
+	registeredNodes := make(map[string]*v1.Node, len(nm.registeredNodes))
+	for nodeName, node := range nm.registeredNodes {
+		registeredNodes[nodeName] = node
+	}
+	return registeredNodes
+}
+
 func (nm *NodeManager) addNode(node *v1.Node) {
 	nm.registeredNodesLock.Lock()
 	nm.registeredNodes[node.ObjectMeta.Name] = node
@@ -288,11 +298,9 @@ func (nm *NodeManager) GetNodeInfo(nodeName k8stypes.NodeName) (NodeInfo, error)
 //
 // This method is a getter but it can cause side-effect of updating NodeInfo objects.
 func (nm *NodeManager) GetNodeDetails() ([]NodeDetails, error) {
-	nm.registeredNodesLock.Lock()
-	defer nm.registeredNodesLock.Unlock()
 	var nodeDetails []NodeDetails
 
-	for nodeName, nodeObj := range nm.registeredNodes {
+	for nodeName, nodeObj := range nm.getNodes() {
 		nodeInfo, err := nm.GetNodeInfoWithNodeObject(nodeObj)
 		if err != nil {
 			return nil, err
@@ -304,10 +312,7 @@ func (nm *NodeManager) GetNodeDetails() ([]NodeDetails, error) {
 }
 
 func (nm *NodeManager) refreshNodes() (errList []error) {
-	nm.registeredNodesLock.Lock()
-	defer nm.registeredNodesLock.Unlock()
-
-	for nodeName := range nm.registeredNodes {
+	for nodeName := range nm.getNodes() {
 		nodeInfo, err := nm.getRefreshedNodeInfo(convertToK8sType(nodeName))
 		if err != nil {
 			errList = append(errList, err)


Danil-Grigorev:
@gnufied PTAL

divyenpatel:
@Danil-Grigorev I looked into code and observed following.

- GetNodeDetails holds the lock on `nm.registeredNodesLock.Lock()`
- GetNodeDetails calls GetNodeInfoWithNodeObject()
- GetNodeInfoWithNodeObject calls getRefreshedNodeInfo()
- getRefreshedNodeInfo() calls RediscoverNode() when retrieved nodeInfo is nil
- RediscoverNode() calls GetNode()
- GetNode() needs `nm.registeredNodesLock.RLock()`

Question is why getRefreshedNodeInfo() finds  nodeInfo nil, I see only one writer for `nodeInfoMap`, which does not write nil `nodeInfo`.

Screen Shot 2020-08-13 at 4 54 03 PM

staging/src/k8s.io/legacy-cloud-providers/vsphere/nodemanager.go
```diff
@@ -288,8 +288,6 @@ func (nm *NodeManager) GetNodeInfo(nodeName k8stypes.NodeName) (NodeInfo, error)
 //
 // This method is a getter but it can cause side-effect of updating NodeInfo objects.
 func (nm *NodeManager) GetNodeDetails() ([]NodeDetails, error) {
-	nm.registeredNodesLock.Lock()
-	defer nm.registeredNodesLock.Unlock()
 	var nodeDetails []NodeDetails
 
 	for nodeName, nodeObj := range nm.registeredNodes {
```
liggitt, line 293: ranging over registeredNodes in an unprotected way is not safe in combination with calls to addNode or removeNode

staging/src/k8s.io/legacy-cloud-providers/vsphere/nodemanager.go
```diff
@@ -304,9 +302,6 @@ func (nm *NodeManager) GetNodeDetails() ([]NodeDetails, error) {
 }
 
 func (nm *NodeManager) refreshNodes() (errList []error) {
-	nm.registeredNodesLock.Lock()
-	defer nm.registeredNodesLock.Unlock()
-
 	for nodeName := range nm.registeredNodes {
```
liggitt, line 305: same comment here, this is not threadsafe

staging/src/k8s.io/legacy-cloud-providers/vsphere/nodemanager.go
Danil-Grigorev: You are right, completely missed this part, thanks! :100:  Now all operations will be performed on a copy of the `nm.registeredNodes` map.

liggitt: What are the performance implications of copying this map? What are the correctness implications of iterating over potentially stale data outside the lock?

Danil-Grigorev: This lock is only set once there is an update requested for the `Node(s)`. If anything else needs to refresh the node info, it would do it the same way. Otherwise, under normal circumstances, the `nodeInfoMap` or `registeredNodes` map would be populated only once - it only accounts for changes outside of the `kubernetes` cluster management. So it is not critical to iterating over potentially stale data, but I would refer to `vSphere` contributors to confirm.

Danil-Grigorev: The performance implications are minimal - it would actually speed up the process, if something else is waiting on the lock to do the full cycle - now it won't. It is just a `map[string]*v1.Node` copy, after all, executed mostly once there is a `kcm` pod restart or other resource change.

Danil-Grigorev:
> @Danil-Grigorev while volume is being provisioned, new node was being added to the k8s cluster. Asking this to understand the case when this is happening. I am trying to figure out why we received nil `nodeInfo` from the nodeInfoMap in `getRefreshedNodeInfo()`

@divyenpatel It happens after the `kube-controller-manager` pod is restarted. Particulary in this case, there is a deadlock between initial `Node` object processing, and the `vsphere-credentials` `Secret`. Both of them had to be reconciled for newly established informers, and locked on the same resource.

At the same time `PV` creation is handled from `kcm`, and is using the same methods as the rest of the plugin. It is blocked on previous deadlock already.

divyenpatel (approved):

staging/src/k8s.io/legacy-cloud-providers/vsphere/nodemanager.go
```diff
@@ -259,6 +259,16 @@ func (nm *NodeManager) GetNode(nodeName k8stypes.NodeName) (v1.Node, error) {
 	return *node, nil
 }
 
+func (nm *NodeManager) getNodes() map[string]*v1.Node {
+	nm.registeredNodesLock.RLock()
+	defer nm.registeredNodesLock.RUnlock()
+	registeredNodes := make(map[string]*v1.Node, len(nm.registeredNodes))
+	for nodeName, node := range nm.registeredNodes {
+		registeredNodes[nodeName] = node
```
JoelSpeed, line 267: Should this be copying the nodes so that we don't mutate them later? Where is registered nodes updated, is this also done under lock?

divyenpatel: `registeredNodes` returned from here are not updated in the workflow. 

I see in the `(nm *NodeManager) refreshNodes()` we are just iterating over the key `nodeName` and do not update the node object.

we can return copy of nodes instead of a pointer. This will require further refactoring as other methods in the workflow are accepting node pointer as argument.

Danil-Grigorev: @JoelSpeed Every operation with registeredNodes is now done under lock supervision, and none of them requires special isolation except the one provided by the defined operations like add or delete, which already locking the resource themselves for the further processing. Each resource operating on the node pointer will update the Node object under the pointer, which is the intent of the method - but will do this while locking the needed resource.

JoelSpeed: Just to clarify, what I meant was should we do `registerdNodes[nodeName] = node.DeepCopy()` which would return a copy that is still a pointer, it would be a very minor change that wouldn't need refactoring anywhere else.

If would seem like the point of this function is to create a copy of `nm.registeredNodes` so that outside actors aren't mutating the NodeManager's internal state, if we aren't doing a DeepCopy before returning the nodes, then outside actors can still modify the internal state.

Having looked at where this is used, the only thing we are actually using from the result of this function is the names of the objects anyway, so it is safe as it currently is, my concern would be if someone started using it to modify stuff later down the line.

It also looks like this could be simplified, we don't actually need the node object where it is being used, but that doesn't need to be addressed now.

JoelSpeed (commented):
/hold cancel

Since this is functionally equivalent in terms of the copy/no-copy question, I'm happy for that to be addressed at a later date. I left a comment explaining my thoughts in place though. Looks like this code could be simplified at a later date.

michaelstoica:
Had the same issue in 1.18 version.

PR #93979, opened 2020-08-13
proxy/userspace: use waitgroups instead of sketchy atomic ops in testcases

### LINKED ISSUE #93904: [sig-network] TestTCPProxyUpdateDeleteUpdate flakes with "error connecting to server" error

dcbw:
Instead of relying on atomic ops to increment/decrement at the right time
just use waitgroups to provide hard synchronization points.

Supercedes https://github.com/kubernetes/kubernetes/pull/93918

@liggitt @danwinship @kubernetes/sig-network-misc 

/kind cleanup
/kind failing-test
/kind flake

```release-note
NONE
```

Fixes https://github.com/kubernetes/kubernetes/issues/93904

### DIFF
--- a/pkg/proxy/userspace/proxier.go
+++ b/pkg/proxy/userspace/proxier.go
@@ -68,6 +68,29 @@ type ServiceInfo struct {
 	stickyMaxAgeSeconds int
 	// Deprecated, but required for back-compat (including e2e)
 	externalIPs []string
+
+	// isStartedAtomic is set to non-zero when the service's socket begins
+	// accepting requests. Used in testcases. Only access this with atomic ops.
+	isStartedAtomic int32
+	// isFinishedAtomic is set to non-zero when the service's socket shuts
+	// down. Used in testcases. Only access this with atomic ops.
+	isFinishedAtomic int32
+}
+
+func (info *ServiceInfo) setStarted() {
+	atomic.StoreInt32(&info.isStartedAtomic, 1)
+}
+
+func (info *ServiceInfo) IsStarted() bool {
+	return atomic.LoadInt32(&info.isStartedAtomic) != 0
+}
+
+func (info *ServiceInfo) setFinished() {
+	atomic.StoreInt32(&info.isFinishedAtomic, 1)
+}
+
+func (info *ServiceInfo) IsFinished() bool {
+	return atomic.LoadInt32(&info.isFinishedAtomic) != 0
 }
 
 func (info *ServiceInfo) setAlive(b bool) {
@@ -124,7 +147,6 @@ type Proxier struct {
 	udpIdleTimeout  time.Duration
 	portMapMutex    sync.Mutex
 	portMap         map[portMapKey]*portMapValue
-	numProxyLoops   int32 // use atomic ops to access this; mostly for testing
 	listenIP        net.IP
 	iptables        iptables.Interface
 	hostIP          net.IP
@@ -427,14 +449,6 @@ func (proxier *Proxier) getServiceInfo(service proxy.ServicePortName) (*ServiceI
 	return info, ok
 }
 
-// addServiceOnPort lockes the proxy before calling addServiceOnPortInternal.
-// Used from testcases.
-func (proxier *Proxier) addServiceOnPort(service proxy.ServicePortName, protocol v1.Protocol, proxyPort int, timeout time.Duration) (*ServiceInfo, error) {
-	proxier.mu.Lock()
-	defer proxier.mu.Unlock()
-	return proxier.addServiceOnPortInternal(service, protocol, proxyPort, timeout)
-}
-
 // addServiceOnPortInternal starts listening for a new service, returning the ServiceInfo.
 // Pass proxyPort=0 to allocate a random port. The timeout only applies to UDP
 // connections, for now.
@@ -465,12 +479,10 @@ func (proxier *Proxier) addServiceOnPortInternal(service proxy.ServicePortName, 
 	proxier.serviceMap[service] = si
 
 	klog.V(2).Infof("Proxying for service %q on %s port %d", service, protocol, portNum)
-	go func(service proxy.ServicePortName, proxier *Proxier) {
+	go func() {
 		defer runtime.HandleCrash()
-		atomic.AddInt32(&proxier.numProxyLoops, 1)
 		sock.ProxyLoop(service, si, proxier.loadBalancer)
-		atomic.AddInt32(&proxier.numProxyLoops, -1)
-	}(service, proxier)
+	}()
 
 	return si, nil
 }
@@ -509,6 +521,7 @@ func (proxier *Proxier) mergeService(service *v1.Service) sets.String {
 			if err := proxier.cleanupPortalAndProxy(serviceName, info); err != nil {
 				klog.Error(err)
 			}
+			info.setFinished()
 		}
 		proxyPort, err := proxier.proxyPorts.AllocateNext()
 		if err != nil {
@@ -541,6 +554,8 @@ func (proxier *Proxier) mergeService(service *v1.Service) sets.String {
 			klog.Errorf("Failed to open portal for %q: %v", serviceName, err)
 		}
 		proxier.loadBalancer.NewService(serviceName, info.sessionAffinityType, info.stickyMaxAgeSeconds)
+
+		info.setStarted()
 	}
 
 	return existingPorts
@@ -578,6 +593,7 @@ func (proxier *Proxier) unmergeService(service *v1.Service, existingPorts sets.S
 			klog.Error(err)
 		}
 		proxier.loadBalancer.DeleteService(serviceName)
+		info.setFinished()
 	}
 	for _, svcIP := range staleUDPServices.UnsortedList() {
 		if err := conntrack.ClearEntriesForIP(proxier.exec, svcIP, v1.ProtocolUDP); err != nil {
--- a/pkg/proxy/userspace/proxier_test.go
+++ b/pkg/proxy/userspace/proxier_test.go
@@ -26,7 +26,6 @@ import (
 	"os"
 	"reflect"
 	"strconv"
-	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


pkg/proxy/userspace/proxier.go
```diff
@@ -461,16 +467,20 @@ func (proxier *Proxier) addServiceOnPortInternal(service proxy.ServicePortName,
...
+	si.finished.Add(1)
 	proxier.serviceMap[service] = si
 
 	klog.V(2).Infof("Proxying for service %q on %s port %d", service, protocol, portNum)
-	go func(service proxy.ServicePortName, proxier *Proxier) {
+	go func() {
 		defer runtime.HandleCrash()
-		atomic.AddInt32(&proxier.numProxyLoops, 1)
+		defer si.finished.Done()
+		si.started.Done()
```
liggitt, line 481: what is blocking on `si.started.Wait()` intended to indicate? it doesn't mean sock.ProxyLoop() has been called yet... it means it's *about* to be called

liggitt:
still seeing flakes with the stress test in the original issue...

```
--- FAIL: TestTCPProxyUpdateDeleteUpdate (0.01s)
    proxier_test.go:180: error connecting to server: Get "http://127.0.0.1:44739/aaaaa": EOF
```

pkg/proxy/userspace/proxier.go
dcbw: @liggitt nor did waitForProxyLoop(,.., 1) mean it was started yet either... I think we just care that the goroutine is started, since sometimes that takes a while.

pkg/proxy/userspace/proxier.go
```diff
@@ -68,6 +68,13 @@ type ServiceInfo struct {
 	stickyMaxAgeSeconds int
 	// Deprecated, but required for back-compat (including e2e)
 	externalIPs []string
+
+	// started is incremented when the service is created, and decremented
+	// when the service's socket beings accepting requests
```
danwinship, line 73: "begins"

pkg/proxy/userspace/proxier.go
```diff
@@ -68,6 +68,13 @@ type ServiceInfo struct {
...
+	started *sync.WaitGroup
+	// finished is incremented when the service is created, and decremented
+	// when the service's socket shuts down
+	finished *sync.WaitGroup
```
danwinship, line 77: How heavyweight are WaitGroups? Given that this is only used for testing, not for ordinary functioning of the proxy.

danwinship (commented):
Given that it's always `TestTCPProxyUpdateDeleteUpdate` that flakes, it seems likely that the flake must be happening in the _second_ connection attempt, since that's the part that's unique to this test.

Looking at one of the recent flakes:

```
E0811 20:10:25.430854      24 proxysocket.go:144] Failed to connect to balancer: missing endpoints
    proxier_test.go:181: error connecting to server: Get "http://127.0.0.1:39143/aaaaa": http: server closed idle connection
```

So it looks like the problem is that we need to wait for the `lb.OnEndpointsAdd(...)` to be processed, not just the `p.OnServiceAdd(...)`. (Well, also, the userspace proxy ought to not accept connections for Services with no Endpoints, but whatever.)

("server closed idle connection" is presumably a bad heuristic in `http.Client` that is just referring to the fact that the server (ie the userspace proxy) closed the connection on it as it was starting to write the request; `http.Client` is failing to consider that this happened on a brand new connection as opposed to a reused persistent connection, and so giving an incorrect error message.)

pkg/proxy/userspace/proxier.go
liggitt: I had the same thought

pkg/proxy/userspace/proxier.go
liggitt: > I think we just care that the goroutine is started, since sometimes that takes a while

I can't think of anything that would wait on si.started that wouldn't be flaky... the listener is already established before addServiceOnPortInternal returns, so there's no need to wait on si.started before sending requests to it... I think we should drop si.started and simplify `finished` to something like a channel we close

knight42:
@danwinship I have investigated the flake a bit at https://github.com/kubernetes/kubernetes/pull/93918#issue-466563115, and I think the root cause is that the service was not deleted from the `LoadBalancerRR` before we added back the endpoint.

pkg/proxy/userspace/proxier.go
dcbw: WaitGroup just uses atomic ops. Not significantly heavier weight IMO than the previous atomic ops stuff.

dcbw:
> @danwinship I have investigated the flake a bit at [#93918 (comment)](https://github.com/kubernetes/kubernetes/pull/93918#issue-466563115), and I think the root cause is that the service was not deleted from the `LoadBalancerRR` before we added back the endpoint.

@knight42 is that because something is not waiting for the previous stuff to be cleaned up, before trying to re-add the new endpoint?

pkg/proxy/userspace/proxier.go
dcbw: fixed

dcbw:
> So it looks like the problem is that we need to wait for the `lb.OnEndpointsAdd(...)` to be processed, not just the `p.OnServiceAdd(...)`. (Well, also, the userspace proxy ought to not accept connections for Services with no Endpoints, but whatever.)

@danwinship Looking at it though, OnEndpointsAdd() doesn't do anything interesting. Just takes a lock and stuff some things into a map. With the waitgroups, addition and deletion of the endpoints should be a condition of the service info started/finished waitgroups too.

dcbw:
@liggitt @danwinship how about now?

pkg/proxy/userspace/proxier.go
liggitt: https://golang.org/src/sync/waitgroup.go?s=2022:2057#L43 looks a lot heavier, especially for something we are only using in a unit test

pkg/proxy/userspace/proxier.go
```diff
@@ -509,6 +514,7 @@ func (proxier *Proxier) mergeService(service *v1.Service) sets.String {
 			if err := proxier.cleanupPortalAndProxy(serviceName, info); err != nil {
 				klog.Error(err)
 			}
+			info.finished.Done()
```
liggitt, line 517: can this be called more than once (sending `finished` negative)? if we `continue` above on line 510, do we miss calling this when we should?

pkg/proxy/userspace/proxier.go
```diff
@@ -578,6 +586,7 @@ func (proxier *Proxier) unmergeService(service *v1.Service, existingPorts sets.S
 			klog.Error(err)
 		}
 		proxier.loadBalancer.DeleteService(serviceName)
+		info.finished.Done()
```
liggitt, line 589: similar question about being called more than once (sending the counter negative)

liggitt:
> @liggitt @danwinship how about now?

I'm not familiar enough with the proxy sync flow to know whether these changes make sense, but I commented on a couple places that seemed weird

pkg/proxy/userspace/proxier.go
dcbw: @liggitt in this case, cleanupPortalAndProxy() -> stopProxy() removes the ServiceInfo from the map and the waitgroup being called is a member of that ServiceInfo, so it cannot be called again. I thought about putting the finished closer to the map deletion, but we want to cover the loadbalancer cleanup too.

pkg/proxy/userspace/proxier.go
dcbw: @liggitt I don't think 510 is an issue, because mergeService() is trying to update the proxy if the service changed. Line 510 means nothing changed so leave it as-is. If the service did change, and it's already in the map, it'll be removed (cleanupPortalAndProxy()) and then re-added with a new ServiceInfo/waitgroup by addServiceOnPortInternal()

gianarb:
Hello!
Tomorrow is release day! Is this PR gonna make it on time? Should we move it to 1.20?
Thanks

liggitt:
since this is addressing a long-lived test-only issue, it should be moved to 1.20

/milestone v1.20

dcbw:
@liggitt would you mind running this latest iteration through your stress test to see if the issue is gone? Thanks!

liggitt:
I'd like people to start building this toolset out themselves, so you should be able to run it. I put instructions for the stress test at https://github.com/kubernetes/kubernetes/issues/93904 (it's really simple to do). Let me know if you have any questions.

pkg/proxy/userspace/proxier.go
liggitt: This comment is still outstanding

dcbw:
> I'd like people to start building this toolset out themselves, so you should be able to run it. I put instructions for the stress test at #93904 (it's really simple to do). Let me know if you have any questions.

@liggitt thanks, didn't realize it was so easy :) It's gone through ~7000 iterations so far without failure, how many should I shoot for?

pkg/proxy/userspace/proxier.go
dcbw: @liggitt We don't actually need waitgroup values > 1, so if you'd like though, I'm happy to move it to custom-rolled atomics and a busyloop for wait.

But I would disagree that watigroup add/done is heavy especially since there's no contention whatsoever when testcases are not running.

liggitt: for something used only by tests, I'd like the cost to be as close to zero for the production code as possible

liggitt:
> @liggitt thanks, didn't realize it was so easy :) It's gone through ~30,000 iterations so far without failure, how many should I shoot for?

on master, it flakes immediately, so that sounds good:
```
219 runs so far, 4 failures
```

dcbw:
@liggitt how about now? pushed a fixup that shows the change from WaitGroup to atomic ops. If that looks better I'll squash and update the commit message.

pkg/proxy/userspace/proxier.go
dcbw: changed to use atomics for started/finished rather than waitgroups.

pkg/proxy/userspace/proxier.go
```diff
@@ -465,12 +479,15 @@ func (proxier *Proxier) addServiceOnPortInternal(service proxy.ServicePortName,
...
+	go func() {
 		defer runtime.HandleCrash()
-		atomic.AddInt32(&proxier.numProxyLoops, 1)
+		wg.Done()
 		sock.ProxyLoop(service, si, proxier.loadBalancer)
-		atomic.AddInt32(&proxier.numProxyLoops, -1)
-	}(service, proxier)
+	}()
+	// Wait until the ProxyLoop goroutine has actually started
+	wg.Wait()
```
liggitt, line 490: I'm not sure why this is being added. Does this actually do anything other than slow down the return a little? This still can return before `sock.ProxyLoop` is called, which doesn't seem fundamentally different in nature from what it was doing before.

liggitt:
> @liggitt how about now? pushed a fixup that shows the change from WaitGroup to atomic ops. If that looks better I'll squash and update the commit message. ~6000 stress iterations so far.

The atomic int operations look good. One comment on the remaining waitgroup in `proxier.go`.

I did not look at the test changes, I'll defer to a sig-network reviewer on whether they are correct / cover the desired functionality, but if stress is happy after O(1000) iterations, I'm happy from a CI signal perspective.

pkg/proxy/userspace/proxier.go
dcbw: @liggitt yeah, I guess that doesn't need to be there; all we really care about is that the socket is open and the iptables rules have been added by openPortal(), so I removed that bit and squashed the fixup.

liggitt:
thanks, if this passes the stress test now, it seems ready for a network reviewer and lgtm

dcbw:
> thanks, if this passes the stress test now, it seems ready for a network reviewer and lgtm

@liggitt `41522 runs so far, 0 failures`

liggitt:
@danwinship you want to do the review pass on the test updates and lgtm?

danwinship (commented):
Commit message / PR title are now incorrect

pkg/proxy/userspace/proxier_test.go
```diff
@@ -89,14 +88,60 @@ func waitForClosedPortUDP(p *Proxier, proxyPort int) error {
...
+
+	return svcInfo
+}
+
+func addServiceAndWaitForInfoIndex(t *testing.T, p *Proxier, servicePortName proxy.ServicePortName, service *v1.Service, idx int) *ServiceInfo {
+	p.OnServiceAdd(service)
+	return waitForServiceInfo(t, p, servicePortName, service, idx)
+}
+
+func addServiceAndWaitForInfo(t *testing.T, p *Proxier, servicePortName proxy.ServicePortName, service *v1.Service) *ServiceInfo {
```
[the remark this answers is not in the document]
dcbw: Fixed to just find the right ServicePort rather than indexes or whatever.

pkg/proxy/userspace/proxier_test.go
```diff
@@ -267,12 +299,15 @@ func TestTCPProxy(t *testing.T) {
 	startProxier(p, t)
 	defer p.shutdown()
 
-	svcInfo, err := p.addServiceOnPort(service, "TCP", 0, time.Second)
-	if err != nil {
-		t.Fatalf("error adding new service: %#v", err)
-	}
+	svcInfo := addServiceAndWaitForInfo(t, p, service, &v1.Service{
```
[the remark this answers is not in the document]
dcbw: Done, in a second commit.

dcbw:
@danwinship @liggitt PTAL thanks!

PR #93982, opened 2020-08-13
wait function and util comments and cleanup

jayunit100:
**What type of PR is this?**
/kind cleanup

- The current e2e wait function doesnt output clear information when running tests on 20 or 30 node clusters, around how much longer it will be waiting for the network to come up, instead it prints "=================", and then there's a pause.
- There needs to be some comments on why we use `unsschedulable` (its because of the way Taints work) as the way that we filter out nodes we're watching, bc the equation for how many total nodes are needed actually subtracts cordoned nodes, which can be confusing also. 

Didn't file an issue since just cleanup, but now the logs show the following, which they didnt show before.

```
Aug 13 17:05:20.730: INFO: Unschedulable nodes= 1, maximum value for starting tests= 0
...
Aug 13 17:05:20.730: INFO: ==== node wait: 3 out of 4 nodes are ready, max notReady allowed 0.  Need 1 more before starting.
```

```release-note
NONE
```

### DIFF
--- a/test/e2e/framework/node/wait.go
+++ b/test/e2e/framework/node/wait.go
@@ -207,19 +207,18 @@ func checkWaitListSchedulableNodes(c clientset.Interface) (*v1.NodeList, error) 
 	return nodes, nil
 }
 
-// CheckReadyForTests returns a method usable in polling methods which will check that the nodes are
-// in a testable state based on schedulability.
+// CheckReadyForTests returns a function which will return 'true' once the number of ready nodes is above the allowedNotReadyNodes threshold (i.e. to be used as a global gate for starting the tests).
 func CheckReadyForTests(c clientset.Interface, nonblockingTaints string, allowedNotReadyNodes, largeClusterThreshold int) func() (bool, error) {
 	attempt := 0
-	var notSchedulable []*v1.Node
 	return func() (bool, error) {
 		attempt++
-		notSchedulable = nil
+		var nodesNotReadyYet []v1.Node
 		opts := metav1.ListOptions{
 			ResourceVersion: "0",
-			FieldSelector:   fields.Set{"spec.unschedulable": "false"}.AsSelector().String(),
+			// remove uncordoned nodes from our calculation, TODO refactor if node v2 API removes that semantic.
+			FieldSelector: fields.Set{"spec.unschedulable": "false"}.AsSelector().String(),
 		}
-		nodes, err := c.CoreV1().Nodes().List(context.TODO(), opts)
+		allNodes, err := c.CoreV1().Nodes().List(context.TODO(), opts)
 		if err != nil {
 			e2elog.Logf("Unexpected error listing nodes: %v", err)
 			if testutils.IsRetryableAPIError(err) {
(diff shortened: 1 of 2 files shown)


oomichi (commented):
/lgtm

test/e2e/framework/node/wait.go
```diff
@@ -207,30 +207,28 @@ func checkWaitListSchedulableNodes(c clientset.Interface) (*v1.NodeList, error)
...
 	attempt := 0
-	var notSchedulable []*v1.Node
 	return func() (bool, error) {
 		attempt++
-		notSchedulable = nil
+		var nodesNotReadyYet []*v1.Node
 		opts := metav1.ListOptions{
 			ResourceVersion: "0",
-			FieldSelector:   fields.Set{"spec.unschedulable": "false"}.AsSelector().String(),
+			// remove uncordoned nodes from our calculation, TODO refactor after node v2 API removes that semantic.
```
neolit123, line 218: i'm not aware of the Node v2 plans. is this real ATM?
if not possibly we should omit the TODO.

test/e2e/framework/node/wait.go
```diff
@@ -207,30 +207,28 @@ func checkWaitListSchedulableNodes(c clientset.Interface) (*v1.NodeList, error)
...
 			}
 			return false, err
 		}
-		for i := range nodes.Items {
-			node := &nodes.Items[i]
-			if !readyForTests(node, nonblockingTaints) {
-				notSchedulable = append(notSchedulable, node)
+		for _, node := range allNodes.Items {
+			if !readyForTests(&node, nonblockingTaints) {
+				nodesNotReadyYet = append(nodesNotReadyYet, &node)
```
neolit123, line 231: i think we should still create a copy of the node object before appending it to the notReadyYet slice?

neolit123 (commented):
thanks for the PR @jayunit100 
/cc @oomichi 
PTAL if you have the time to double check the logic

neolit123:
we should remove the release note here in the OP, we usually don't do that for e2e test changes:
> Added status message about remaining not ready nodes while waiting for end to end tests to start.

neolit123:
> /cc @oomichi
> PTAL if you have the time to double check the logic

oh, never mind you LGTMed already.

test/e2e/framework/node/wait.go
jayunit100: spec.Unschedulable i think is fundamentally going to change at some point, right?

jayunit100: ( i believe this related to taintnodebycondition and so on, but in general, this is known to be a sore point of the API that is supposed to be, eventually, revisited, so i think we should keep the v2 node api comes up which will have a cleaner way of doing this )

test/e2e/framework/node/wait.go
jayunit100: ok, what is the purpose of the copy ?  i wasnt sure there was a place where this made a difference but maybe im missing something.  iirc the nodes arent going to get mutated anytime during the scope of this function.

test/e2e/framework/node/wait.go
neolit123: SGTM, hope the comment is not misleading to readers, as in "Node v2 is in the works as we speak".

test/e2e/framework/node/wait.go
neolit123: i'm pretty sure the purpose of the original copy was to avoid [this first pitfall listed here](https://medium.com/@pedram.esmaeeli/golang-pitfalls-f2ebae9c8208)

test/e2e/framework/node/wait.go
neolit123: actually...
> TODO refactor after node v2 API removes that semantic.

we don't know that for sure unless SIG Node confirms it 100%, so i'd avoid putting words in their mouth here.

i'd reword to:
> TODO refactor if Node v2 API removes that semantic.

test/e2e/framework/node/wait.go
jayunit100: fixed

jayunit100:
cc @oomichi needs a new lgtm :)

PR #53194, opened 2017-09-28
avoid kubelet converts and validates pods multiple times

dixudx:
**What this PR does / why we need it**:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #53171

**Special notes for your reviewer**:
/assign @yujuhong @caesarxuchao 

**Release note**:

```release-note
None
```

PR #93985, opened 2020-08-14
update test to match validation filter of pods

### LINKED ISSUE #93905: [sig-node] TestInvalidPodFiltered flakes with "Expected no update in channel" error

TestInvalidPodFiltered

**Reason for failure**:

```
--- FAIL: TestInvalidPodFiltered (0.00s)
    config_test.go:126: Expected no update in channel, Got types.PodUpdate{Pods:[]*v1.Pod{(*v1.Pod)(0xc0000b4000)}, Op:1, Source:"test"}
FAIL
```

With that toleration removed in https://github.com/kubernetes/kubernetes/pull/93605, this flake has been seen (https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93605/pull-kubernetes-bazel-test/1293289839906525184).

```
go test -race -c ./pkg/kubelet/config
```

```
stress ./config.test -test.run TestInvalidPodFiltered
```

```sh
ls -lat $TMPDIR/go-stress*
```

/sig node
cc @kubernetes/sig-node-test-failures

MHBauer:
invalid pod filter changed to name only

changed for k8s v1.9 in 811447ea0a9928c326965f3bbd8195826b20daf8 #53194

**What type of PR is this?**
/kind cleanup
/kind flake

Full validation was removed for v1.9 and changed to a duplicate name only check.

I'm not sure if we want to fix it this way or by going back to a full validation with a `validation.ValidatePod(`.

Fixes #93905

/assign @liggitt 
I ran the suggested stress for 20000.

/cc @knight42 
as was looking at the originating issue.

```release-note
NONE
```

### DIFF
--- a/pkg/kubelet/config/config_test.go
+++ b/pkg/kubelet/config/config_test.go
@@ -189,8 +189,8 @@ func TestInvalidPodFiltered(t *testing.T) {
 	channel <- podUpdate
 	expectPodUpdate(t, ch, CreatePodUpdate(kubetypes.ADD, TestSource, CreateValidPod("foo", "new")))
 
-	// add an invalid update
-	podUpdate = CreatePodUpdate(kubetypes.UPDATE, TestSource, &v1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "foo"}})
+	// add an invalid update, pod with the same name
+	podUpdate = CreatePodUpdate(kubetypes.ADD, TestSource, CreateValidPod("foo", "new"))
 	channel <- podUpdate
 	expectNoPodUpdate(t, ch)
 }


pkg/kubelet/config/config_test.go
```diff
@@ -189,8 +189,8 @@ func TestInvalidPodFiltered(t *testing.T) {
 	channel <- podUpdate
 	expectPodUpdate(t, ch, CreatePodUpdate(kubetypes.ADD, TestSource, CreateValidPod("foo", "new")))
 
-	// add an invalid update
-	podUpdate = CreatePodUpdate(kubetypes.UPDATE, TestSource, &v1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "foo"}})
+	// add an invalid update, pod with same name
+	podUpdate = CreatePodUpdate(kubetypes.ADD, TestSource, CreateValidPod("foo", "new"))
 	channel <- podUpdate
 	expectNoPodUpdate(t, ch)
```
knight42, line 195: just random thought, feel free to take it or not. Should we ensure `podStorage.Merge` has completed before we checking channel `ch` or at least wait for a while in `expectNoPodUpdate`? Otherwise we might receive event from `ch` after a while but `expectNoPodUpdate` fails to notice that.

pkg/kubelet/config/config_test.go
```diff
@@ -189,8 +189,8 @@ func TestInvalidPodFiltered(t *testing.T) {
 	channel <- podUpdate
 	expectPodUpdate(t, ch, CreatePodUpdate(kubetypes.ADD, TestSource, CreateValidPod("foo", "new")))
 
-	// add an invalid update
-	podUpdate = CreatePodUpdate(kubetypes.UPDATE, TestSource, &v1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "foo"}})
+	// add an invalid update, pod with same name
```
SergeyKanzhelev, line 192: nit
```suggestion
	// add an invalid update, pod with the same name
```

pkg/kubelet/config/config_test.go
```diff
@@ -189,8 +189,8 @@ func TestInvalidPodFiltered(t *testing.T) {
 	channel <- podUpdate
 	expectPodUpdate(t, ch, CreatePodUpdate(kubetypes.ADD, TestSource, CreateValidPod("foo", "new")))
 
-	// add an invalid update
```
SergeyKanzhelev, line 192: since there is no more invalid update, method name may be changed to `TestSameNamePodFiltered`?

liggitt:
/priority important-soon
/unassign

I'd like a node reviewer to look at this. Note that this is blocking https://github.com/kubernetes/kubernetes/pull/93605 and full reopen of 1.20, so getting this reviewed in the next couple days would be helpful.

cc @kubernetes/sig-node-pr-reviews

pkg/kubelet/config/config_test.go
MHBauer: expectNoPodUpdate is used inside of expectPodUpdate, so it would delay all the tests using that function, and I did not want to slow everything down for a probabilistic check. Do you have a suggestion for how I can force it?

pkg/kubelet/config/config_test.go
MHBauer: ok

pkg/kubelet/config/config_test.go
MHBauer: The internal function being tested is still filterInvalidPod, so I appreciate that matching name.

MHBauer (commented):
/assign @dashpole @derekwaynecarr

MHBauer:
I think the drawback of this is it is still possibly flaky in the same way depending on changes in the future.
Not sure how much it's worth testing or investigating in a deeper way for a more robust test.
A change doesn't immediately spring to mind for me, so I would appreciate a thought on how to change, or a push on if I should spend more time to update it.

MHBauer:
looks close to https://storage.googleapis.com/k8s-gubernator/triage/index.html?test=%20Aggregator%20Should%20be%20able%20to%20support#c100df879f7288e213ba but I see several similar failures
/retest

pkg/kubelet/config/config_test.go
SergeyKanzhelev: One thing comes to mind is adding and removing a pod expecting corresponding updates in `ch`. If something unexpected was delayed on the channel from the previous update - expected updates on this additional add/remove combination wouldn't match.

But this may be a separate work item I think.

SergeyKanzhelev: /lgtm

liggitt:
/approve

exercised with stress tool, no longer flakes

/milestone v1.20

PR #94002, opened 2020-08-14
Fix bug for inconsistent lists served from etcd

wojtek-t:
```release-note
kube-apiserver: fixed a bug returning inconsistent results from list requests which set a field or label selector and set a paging limit
```

/kind bug
/priority critical-urgent

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
+++ b/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
@@ -572,7 +572,7 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions, 
 		fromRV = &parsedRV
 	}
 
-	var returnedRV, continueRV int64
+	var returnedRV, continueRV, withRev int64
 	var continueKey string
 	switch {
 	case s.pagingEnabled && len(pred.Continue) > 0:
@@ -593,7 +593,7 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions, 
 		// continueRV==0 is invalid.
 		// If continueRV < 0, the request is for the latest resource version.
 		if continueRV > 0 {
-			options = append(options, clientv3.WithRev(continueRV))
+			withRev = continueRV
 			returnedRV = continueRV
 		}
 	case s.pagingEnabled && pred.Limit > 0:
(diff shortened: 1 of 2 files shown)


wojtek-t:
@deads2k @smarterclayton - in GKE we have watchcache enabled for everything but events, but I'm wondering why you've never hit this problem given you have watchcache disabled ...

staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
```diff
@@ -695,6 +695,10 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions,
 			break
 		}
 		key = string(lastKey) + "\x00"
+		if fromRV == nil {
+			*fromRV = uint64(returnedRV)
+			options = append(options, clientv3.WithRev(returnedRV))
```
liggitt, line 700: appending to options inside this for loop looks wrong... that means we'll have multiple WithRev options specified with different values

liggitt: edit: nevermind, I see this only happens when fromRV is nil

staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
```diff
@@ -695,6 +695,10 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions,
 			break
 		}
 		key = string(lastKey) + "\x00"
+		if fromRV == nil {
```
liggitt, line 698: if we're using ResourceVersionMatchNotOlderThan, a non-nil fromRV doesn't mean we added a WithRev option.

I think we should track whether we added WithRev explicitly, and make that the condition here.

liggitt: That also means we need a test that demonstrates paging+filtering+ResourceVersionMatchNotOlderThan that would exercise this

wojtek-t: Good catch, test added and test is already catching it.

Now I need to fix the code.

wojtek-t: Fixed

wojtek-t:
@liggitt - thanks for review - PTAL

liggitt:
can you open cherry-picks to 1.16-1.18 for this?

wojtek-t:
> can you open cherry-picks to 1.16-1.18 for this?

Will do on Monday, since these aren't automatic ones unfortunately....

wojtek-t:
Cherrypicks are opened.

PR #94015, opened 2020-08-14
run node bandwidth tests using iperf2 and collecting more detailed information

### LINKED ISSUE #94005: iperf tests outputs need to be cleaned up and improved
**What happened**:
The iperf tests were written a while ago but i think havent been updated.

- They measure bandwidth over multiple pods, but they probably should also have a test which reports individual pod's  bandwidth between one pod over a SEP.
- They should pretty print a table so its easy to see exactly what is going on with the bandwidth
- They should use iperf3 (need to check that)
- Maybe we should have permutations for udp and tcp?

The current output is something like:

```
[begin] Node,Bandwidth CSV
	 "192.168.88.1"	 1.696234E+10
```

That running the `ginkgo.focus="should transfer ~ 1GB onto the service endpoint"` tests would result in a table such as those that you can get from iperf3, i.e. 

```
Client on 192.168.5.227:  [  4]   0.00-1.00   sec   405 MBytes  3.40 Gbits/sec  
Client on 192.168.5.227:  [  4]   1.00-2.00   sec   442 MBytes  3.71 Gbits/sec   
Client on 192.168.5.227:  [  4]   2.00-3.00   sec   461 MBytes  3.87 Gbits/sec   
 ```

**How to reproduce it**:

run the e2es on any cluster, 
```KUBECONFIG=/Users/jayunit100/.kube/config KUBERNETES_SERVICE_HOST=127.0.0.1 KUBERNETES_SERVICE_PORT=50100 .//_output/local/go/bin/e2e.test --ginkgo.focus="should transfer ~" --provider=local```
**Environment**:
- Kubernetes version (use `kubectl version`): master, any version

### PR DESCRIPTION
**Pull request by mattfenwick**
**What type of PR is this?**

/kind cleanup

This improves the level of detail of networking performance information using iperf2's more detailed reports.  Here's an example of iperf2 results:

```
------------------------------------------------------------
Client connecting to iperf2-server, TCP port 6789 with pid 15
TCP window size: 85.0 KByte (default)
------------------------------------------------------------
[  3] local 10.244.1.45 port 57172 connected with 10.96.114.79 port 6789
[ ID] Interval        Transfer    Bandwidth       Write/Err
[  3] 0.00-1.00 sec  2.20 GBytes  18.9 Gbits/sec  18044/0
[  3] 1.00-2.00 sec  2.38 GBytes  20.5 Gbits/sec  19519/0
[  3] 2.00-3.00 sec  2.59 GBytes  22.2 Gbits/sec  21188/0
[  3] 3.00-4.00 sec  2.32 GBytes  19.9 Gbits/sec  18995/0
[  3] 4.00-5.00 sec  2.42 GBytes  20.8 Gbits/sec  19822/0
[  3] 5.00-6.00 sec  2.35 GBytes  20.2 Gbits/sec  19247/0
[  3] 6.00-7.00 sec  2.43 GBytes  20.8 Gbits/sec  19867/0
[  3] 7.00-8.00 sec  2.25 GBytes  19.4 Gbits/sec  18470/0
[  3] 8.00-9.00 sec  2.52 GBytes  21.7 Gbits/sec  20683/0
[  3] 0.00-9.97 sec  23.5 GBytes  20.2 Gbits/sec  192361/0
```

Or in CSV format:

```
20201210141800.884,10.244.2.24,47880,10.96.114.79,6789,3,0.0-1.0,1677852672,13422821376
20201210141801.881,10.244.2.24,47880,10.96.114.79,6789,3,1.0-2.0,1980760064,15846080512
20201210141802.883,10.244.2.24,47880,10.96.114.79,6789,3,2.0-3.0,1886650368,15093202944
20201210141803.882,10.244.2.24,47880,10.96.114.79,6789,3,3.0-4.0,2035417088,16283336704
20201210141804.879,10.244.2.24,47880,10.96.114.79,6789,3,4.0-5.0,1922957312,15383658496
20201210141805.881,10.244.2.24,47880,10.96.114.79,6789,3,5.0-6.0,2095316992,16762535936
20201210141806.882,10.244.2.24,47880,10.96.114.79,6789,3,6.0-7.0,1741291520,13930332160
20201210141807.879,10.244.2.24,47880,10.96.114.79,6789,3,7.0-8.0,1862926336,14903410688
20201210141808.878,10.244.2.24,47880,10.96.114.79,6789,3,8.0-9.0,1821245440,14569963520
20201210141809.849,10.244.2.24,47880,10.96.114.79,6789,3,0.0-10.0,18752208896,15052492511
```

Compare this to the existing iperf output -- in this comma-separated output, the final field is the bandwidth:

```
20201030110026,10.244.2.2,37112,10.96.37.89,8002,3,0.0-10.0,24580849664,19730437332
```

Combined results from iperf2:

```
Dec 10 09:13:05.658: INFO:                      From                       To         Bandwidth (MB/s)
Dec 10 09:13:05.658: INFO:         three-node-worker        three-node-worker                    20013
Dec 10 09:13:05.658: INFO:        three-node-worker2        three-node-worker                    13178
Dec 10 09:13:05.658: INFO:  three-node-control-plane        three-node-worker                    12676
```

### Why not iperf3?

Check out [this project](https://github.com/Pharb/kubernetes-iperf3) for quickly getting up to speed running iperf3 on kube.

However, iperf2 is already in the agnhost image whereas iperf3 is not.

Fixes https://github.com/kubernetes/kubernetes/issues/94005

```release-note
None
```

### DIFF
--- a/test/e2e/network/networking_perf.go
+++ b/test/e2e/network/networking_perf.go
@@ -18,145 +18,311 @@ package network
 
 // Tests network performance using iperf or other containers.
 import (
+	"context"
 	"fmt"
+	discoveryv1beta1 "k8s.io/api/discovery/v1beta1"
 	"time"
 
 	"github.com/onsi/ginkgo"
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
+	e2edeployment "k8s.io/kubernetes/test/e2e/framework/deployment"
 	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
-	e2eservice "k8s.io/kubernetes/test/e2e/framework/service"
 	imageutils "k8s.io/kubernetes/test/utils/image"
 )
 
 const (
-	// empirically derived as a baseline for expectations from running this test using kube-up.sh.
-	gceBandwidthBitsEstimate = int64(30000000000)
-	// on 4 node clusters, we found this test passes very quickly, generally in less then 100 seconds.
-	smallClusterTimeout = 200 * time.Second
+	// use this timeout for larger clusters
+	largeClusterTimeout = 400 * time.Second
+	// iperf2BaselineBandwidthMegabytesPerSecond sets a baseline for iperf2 bandwidth of 90 MB/s
+	iperf2BaselineBandwidthMegabytesPerSecond = 90
+	// iperf2Port selects an arbitrary, unique port to run iperf2's client and server on
+	iperf2Port = 6789
+	// labelKey is used as a key for selectors
+	labelKey = "app"
+	// clientLabelValue is used as a value for iperf2 client selectors
+	clientLabelValue = "iperf2-client"
+	// serverLabelValue is used as a value for iperf2 server selectors
+	serverLabelValue = "iperf2-server"
+	// serverServiceName defines the service name used for the iperf2 server
+	serverServiceName = "iperf2-server"
 )
 
-// Declared as Flakey since it has not been proven to run in parallel on small nodes or slow networks in CI
-var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-Performance]", func() {
-
-	f := framework.NewDefaultFramework("network-perf")
-
-	// A few simple bandwidth tests which are capped by nodes.
-	// TODO replace the 1 with the scale option implementation
-	// TODO: Make this a function parameter, once we distribute iperf endpoints, possibly via session affinity.
-	numClient := 1
-	numServer := 1
-	maxBandwidthBits := gceBandwidthBitsEstimate
-
-	familyStr := ""
-	if framework.TestContext.ClusterIsIPv6() {
-		familyStr = "-V "
+func iperf2ServerDeployment(client clientset.Interface, namespace string, isIPV6 bool) (*appsv1.Deployment, error) {
+	framework.Logf("deploying iperf2 server")
+	one := int64(1)
+	replicas := int32(1)
+	labels := map[string]string{labelKey: serverLabelValue}
+	args := []string{
+		"-s",
+		"-p",
+		fmt.Sprintf("%d", iperf2Port),
 	}
-
-	ginkgo.It(fmt.Sprintf("should transfer ~ 1GB onto the service endpoint %v servers (maximum of %v clients)", numServer, numClient), func() {
-		nodes, err := e2enode.GetReadySchedulableNodes(f.ClientSet)
-		framework.ExpectNoError(err)
-		totalPods := len(nodes.Items)
-		// for a single service, we expect to divide bandwidth between the network.  Very crude estimate.
-		expectedBandwidth := int(float64(maxBandwidthBits) / float64(totalPods))
-		appName := "iperf-e2e"
-		_, err = e2eservice.CreateServiceForSimpleAppWithPods(
-			f.ClientSet,
-			8001,
-			8002,
-			f.Namespace.Name,
-			appName,
-			func(n v1.Node) v1.PodSpec {
-				return v1.PodSpec{
-					Containers: []v1.Container{{
-						Name:    "iperf-server",
-						Image:   imageutils.GetE2EImage(imageutils.Agnhost),
-						Command: []string{"/bin/sh"},
-						Args: []string{
-							"-c",
-							"/usr/local/bin/iperf " + familyStr + "-s -p 8001 ",
-						},
-						Ports: []v1.ContainerPort{{ContainerPort: 8001}},
-					}},
-					NodeName:      n.Name,
-					RestartPolicy: v1.RestartPolicyOnFailure,
-				}
+	if isIPV6 {
+		args = append(args, "-V")
+	}
+	deploymentSpec := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   "iperf2-server-deployment",
+			Labels: labels,
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: &replicas,
+			Selector: &metav1.LabelSelector{
+				MatchLabels: labels,
 			},
-			// this will be used to generate the -service name which all iperf clients point at.
-			numServer, // Generally should be 1 server unless we do affinity or use a version of iperf that supports LB
-			true,      // Make sure we wait, otherwise all the clients will die and need to restart.
-		)
-
-		if err != nil {
-			framework.Failf("Fatal error waiting for iperf server endpoint : %v", err)
-		}
-
-		iperfClientPodLabels := e2enode.CreatePodsPerNodeForSimpleApp(
-			f.ClientSet,
-			f.Namespace.Name,
-			"iperf-e2e-cli",
-			func(n v1.Node) v1.PodSpec {
-				return v1.PodSpec{
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: labels,
+				},
+				Spec: v1.PodSpec{
+					TerminationGracePeriodSeconds: &one,
 					Containers: []v1.Container{
 						{
-							Name:    "iperf-client",
+							Name:    "iperf2-server",
 							Image:   imageutils.GetE2EImage(imageutils.Agnhost),
-							Command: []string{"/bin/sh"},
-							Args: []string{
-								"-c",
-								"/usr/local/bin/iperf " + familyStr + "-c service-for-" + appName + " -p 8002 --reportstyle C && sleep 5",
+							Command: []string{"iperf"},
+							Args:    args,
+							Ports: []v1.ContainerPort{
+								{
+									ContainerPort: iperf2Port,
+									Protocol:      v1.ProtocolTCP,
+								},
 							},
 						},
 					},
-					RestartPolicy: v1.RestartPolicyOnFailure, // let them successfully die.
+				},
+			},
+		},
+	}
+
+	deployment, err := client.AppsV1().Deployments(namespace).Create(context.TODO(), deploymentSpec, metav1.CreateOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("deployment %q Create API error: %v", deploymentSpec.Name, err)
+	}
+	framework.Logf("Waiting for deployment %q to complete", deploymentSpec.Name)
+	err = e2edeployment.WaitForDeploymentComplete(client, deployment)
+	if err != nil {
+		return nil, fmt.Errorf("deployment %q failed to complete: %v", deploymentSpec.Name, err)
+	}
+
+	return deployment, nil
+}
+
+func iperf2ServerService(client clientset.Interface, namespace string) (*v1.Service, error) {
+	service := &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{Name: serverServiceName},
+		Spec: v1.ServiceSpec{
+			Selector: map[string]string{
+				labelKey: serverLabelValue,
+			},
+			Ports: []v1.ServicePort{
+				{Protocol: v1.ProtocolTCP, Port: iperf2Port},
+			},
+		},
+	}
+	return client.CoreV1().Services(namespace).Create(context.TODO(), service, metav1.CreateOptions{})
+}
+
+func iperf2ClientDaemonSet(client clientset.Interface, namespace string) (*appsv1.DaemonSet, error) {
+	one := int64(1)
+	labels := map[string]string{labelKey: clientLabelValue}
+	spec := &appsv1.DaemonSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   "iperf2-clients",
+			Labels: labels,
+		},
+		Spec: appsv1.DaemonSetSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: labels,
+			},
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: labels,
+				},
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						{
+							Name:    "iperf2-client",
+							Image:   imageutils.GetE2EImage(imageutils.Agnhost),
+							Command: []string{"/agnhost"},
+							Args:    []string{"pause"},
+						},
+					},
+					TerminationGracePeriodSeconds: &one,
+				},
+			},
+		},
+		Status: appsv1.DaemonSetStatus{},
+	}
+
+	ds, err := client.AppsV1().DaemonSets(namespace).Create(context.TODO(), spec, metav1.CreateOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("daemonset %s Create API error: %v", spec.Name, err)
+	}
+	return ds, nil
+}
+
+// Test summary:
+//   This test uses iperf2 to obtain bandwidth data between nodes in the cluster, providing a coarse measure
+//   of the health of the cluster network.  The test runs two sets of pods:
+//     1. an iperf2 server on a single node
+//     2. a daemonset of iperf2 clients
+//   The test then iterates through the clients, one by one, running iperf2 from each of them to transfer
+//   data to the server and back for ten seconds, after which the results are collected and parsed.
+//   Thus, if your cluster has 10 nodes, then 10 test runs are performed.
+//     Note: a more complete test could run this scenario with a daemonset of servers as well; however, this
+//     would require n^2 tests, n^2 time, and n^2 network resources which quickly become prohibitively large
+//     as the cluster size increases.
+//   Finally, after collecting all data, the results are analyzed and tabulated.
+var _ = SIGDescribe("Networking IPerf2 [Feature:Networking-Performance]", func() {
+	// this test runs iperf2: one pod as a server, and a daemonset of clients
+	f := framework.NewDefaultFramework("network-perf")
+
+	ginkgo.It(fmt.Sprintf("should run iperf2"), func() {
+		readySchedulableNodes, err := e2enode.GetReadySchedulableNodes(f.ClientSet)
+		framework.ExpectNoError(err)
+
+		familyStr := ""
+		if framework.TestContext.ClusterIsIPv6() {
+			familyStr = "-V "
+		}
+
+		serverPodsListOptions := metav1.ListOptions{
+			LabelSelector: fmt.Sprintf("%s=%s", labelKey, serverLabelValue),
+		}
+
+		// Step 1: set up iperf2 server -- a single pod on any node
+		_, err = iperf2ServerDeployment(f.ClientSet, f.Namespace.Name, framework.TestContext.ClusterIsIPv6())
+		framework.ExpectNoError(err, "deploy iperf2 server deployment")
+
+		_, err = iperf2ServerService(f.ClientSet, f.Namespace.Name)
+		framework.ExpectNoError(err, "deploy iperf2 server service")
+
+		// Step 2: set up iperf2 client daemonset
+		//   initially, the clients don't do anything -- they simply pause until they're called
+		_, err = iperf2ClientDaemonSet(f.ClientSet, f.Namespace.Name)
+		framework.ExpectNoError(err, "deploy iperf2 client daemonset")
+
+		// Make sure the server is ready to go
+		framework.Logf("waiting for iperf2 server endpoints")
+		err = wait.Poll(2*time.Second, largeClusterTimeout, func() (done bool, err error) {
+			listOptions := metav1.ListOptions{LabelSelector: fmt.Sprintf("%s=%s", discoveryv1beta1.LabelServiceName, serverServiceName)}
+			esList, err := f.ClientSet.DiscoveryV1beta1().EndpointSlices(f.Namespace.Name).List(context.TODO(), listOptions)
+			framework.ExpectNoError(err, "Error fetching EndpointSlice for Service %s/%s", f.Namespace.Name, serverServiceName)
+
+			if len(esList.Items) == 0 {
+				framework.Logf("EndpointSlice for Service %s/%s not found", f.Namespace.Name, serverServiceName)
+				return false, nil
+			}
+			return true, nil
+		})
+		framework.ExpectNoError(err, "unable to wait for endpoints for the iperf service")
+		framework.Logf("found iperf2 server endpoints")
+
+		clientPodsListOptions := metav1.ListOptions{
+			LabelSelector: fmt.Sprintf("%s=%s", labelKey, clientLabelValue),
+		}
+
+		framework.Logf("waiting for client pods to be running")
+		var clientPodList *v1.PodList
+		err = wait.Poll(2*time.Second, largeClusterTimeout, func() (done bool, err error) {
+			clientPodList, err = f.ClientSet.CoreV1().Pods(f.Namespace.Name).List(context.TODO(), clientPodsListOptions)
+			if err != nil {
+				return false, err
+			}
+			if len(clientPodList.Items) < len(readySchedulableNodes.Items) {
+				return false, nil
+			}
+			for _, pod := range clientPodList.Items {
+				if pod.Status.Phase != v1.PodRunning {
+					return false, nil
 				}
-			},
-			numClient,
-		)
-		expectedCli := numClient
-		if len(nodes.Items) < expectedCli {
-			expectedCli = len(nodes.Items)
+			}
+			return true, nil
+		})
+		framework.ExpectNoError(err, "unable to wait for client pods to come up")
+		framework.Logf("all client pods are ready: %d pods", len(clientPodList.Items))
+
+		// Get a reference to the server pod for later
+		serverPodList, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).List(context.TODO(), serverPodsListOptions)
+		framework.ExpectNoError(err)
+		if len(serverPodList.Items) != 1 {
+			framework.Failf("expected 1 server pod, found %d", len(serverPodList.Items))
+		}
+		serverPod := serverPodList.Items[0]
+		framework.Logf("server pod phase %s", serverPod.Status.Phase)
+		for i, condition := range serverPod.Status.Conditions {
+			framework.Logf("server pod condition %d: %+v", i, condition)
+		}
+		for i, cont := range serverPod.Status.ContainerStatuses {
+			framework.Logf("server pod container status %d: %+v", i, cont)
 		}
 
-		framework.Logf("Reading all perf results to stdout.")
-		framework.Logf("date,cli,cliPort,server,serverPort,id,interval,transferBits,bandwidthBits")
+		framework.Logf("found %d matching client pods", len(clientPodList.Items))
 
-		// Extra 1/10 second per client.
-		iperfTimeout := smallClusterTimeout + (time.Duration(expectedCli/10) * time.Second)
-		iperfResults := &IPerfResults{}
-
-		iperfClusterVerification := f.NewClusterVerification(
-			f.Namespace,
-			framework.PodStateVerification{
-				Selectors:   iperfClientPodLabels,
-				ValidPhases: []v1.PodPhase{v1.PodSucceeded},
-			},
-		)
-
-		pods, err2 := iperfClusterVerification.WaitFor(expectedCli, iperfTimeout)
-		if err2 != nil {
-			framework.Failf("Error in wait...")
-		} else if len(pods) < expectedCli {
-			framework.Failf("IPerf restuls : Only got %v out of %v, after waiting %v", len(pods), expectedCli, iperfTimeout)
-		} else {
-			// For each builds up a collection of IPerfRecords
-			iperfClusterVerification.ForEach(
-				func(p v1.Pod) {
-					resultS, err := framework.LookForStringInLog(f.Namespace.Name, p.Name, "iperf-client", "0-", 1*time.Second)
-					if err == nil {
-						framework.Logf(resultS)
-						iperfResults.Add(NewIPerf(resultS))
-					} else {
-						framework.Failf("Unexpected error, %v when running forEach on the pods.", err)
-					}
-				})
+		nodeResults := &IPerf2NodeToNodeCSVResults{
+			ServerNode: serverPod.Spec.NodeName,
+			Results:    map[string]*IPerf2EnhancedCSVResults{},
 		}
-		fmt.Println("[begin] Node,Bandwidth CSV")
-		fmt.Println(iperfResults.ToTSV())
-		fmt.Println("[end] Node,Bandwidth CSV")
 
-		for ipClient, bandwidth := range iperfResults.BandwidthMap {
-			framework.Logf("%v had bandwidth %v.  Ratio to expected (%v) was %f", ipClient, bandwidth, expectedBandwidth, float64(bandwidth)/float64(expectedBandwidth))
+		// Step 3: iterate through the client pods one by one, running iperf2 in client mode to transfer
+		//   data to the server and back and measure bandwidth
+		for _, pod := range clientPodList.Items {
+			podName := pod.Name
+			nodeName := pod.Spec.NodeName
+
+			iperfVersion := f.ExecShellInPod(podName, "iperf -v || true")
+			framework.Logf("iperf version: %s", iperfVersion)
+
+			for try := 0; ; try++ {
+				/* iperf2 command parameters:
+				 *  -e: use enhanced reporting giving more tcp/udp and traffic information
+				 *  -p %d: server port to connect to
+				 *  --reportstyle C: report as Comma-Separated Values
+				 *  -i 1: seconds between periodic bandwidth reports
+				 *  -c %s: run in client mode, connecting to <host>
+				 */
+				command := fmt.Sprintf(`iperf %s -e -p %d --reportstyle C -i 1 -c %s && sleep 5`, familyStr, iperf2Port, serverServiceName)
+				framework.Logf("attempting to run command '%s' in client pod %s (node %s)", command, podName, nodeName)
+				output := f.ExecShellInPod(podName, command)
+				framework.Logf("output from exec on client pod %s (node %s): \n%s\n", podName, nodeName, output)
+
+				results, err := ParseIPerf2EnhancedResultsFromCSV(output)
+				if err == nil {
+					nodeResults.Results[nodeName] = results
+					break
+				} else if try == 2 {
+					framework.ExpectNoError(err, "unable to parse iperf2 output from client pod %s (node %s)", pod.Name, nodeName)
+				} else {
+					framework.Logf("Retrying: IPerf run failed: %+v", err)
+				}
+			}
+		}
+
+		// Step 4: after collecting all the client<->server data, compile and present the results
+		/*
+						Example output:
+
+			Dec 22 07:52:41.102: INFO:                                From                                 To    Bandwidth (MB/s)
+			Dec 22 07:52:41.102: INFO:              three-node-ipv6-worker            three-node-ipv6-worker2                2381
+			Dec 22 07:52:41.102: INFO:             three-node-ipv6-worker2            three-node-ipv6-worker2                2214
+			Dec 22 07:52:41.102: INFO:             three-node-ipv6-worker3            three-node-ipv6-worker2                3123
+
+		*/
+		framework.Logf("%35s%35s%20s", "From", "To", "Bandwidth (MB/s)")
+		for nodeFrom, results := range nodeResults.Results {
+			framework.Logf("%35s%35s%20d", nodeFrom, nodeResults.ServerNode, results.Total.bandwidthMB())
+		}
+		for clientNode, results := range nodeResults.Results {
+			megabytesPerSecond := results.Total.bandwidthMB()
+			if megabytesPerSecond < iperf2BaselineBandwidthMegabytesPerSecond {
+				framework.Failf("iperf2 MB/s received below baseline of %d for client %s to server %s: %d", iperf2BaselineBandwidthMegabytesPerSecond, clientNode, nodeResults.ServerNode, megabytesPerSecond)
+			}
 		}
 	})
 })
--- a/test/e2e/network/util_iperf.go
+++ b/test/e2e/network/util_iperf.go
@@ -21,20 +21,26 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"github.com/pkg/errors"
+	"math"
 	"strconv"
 	"strings"
 
 	"k8s.io/kubernetes/test/e2e/framework"
 )
 
-// IPerfResults is a struct that stores some IPerfResult
+const (
+	megabyte = 1024 * 1024
+)
+
+// IPerfResults is a struct that stores some IPerfCSVResult
 type IPerfResults struct {
 	BandwidthMap map[string]int64
 }
 
-// IPerfResult struct modelling an iperf record....
+// IPerfCSVResult struct modelling an iperf record....
 // 20160314154239,172.17.0.3,34152,172.17.0.2,5001,3,0.0-10.0,33843707904,27074774092
-type IPerfResult struct {
+type IPerfCSVResult struct {
 	date          string // field 1 in the csv
 	cli           string // field 2 in the csv
 	cliPort       int64  // ...
@@ -46,8 +52,12 @@ type IPerfResult struct {
 	bandwidthBits int64
 }
 
+func (i *IPerfCSVResult) bandwidthMB() int64 {
+	return int64(math.Round(float64(i.bandwidthBits) / float64(megabyte) / 8))
+}
+
 // Add adds a new result to the Results struct.
-func (i *IPerfResults) Add(ipr *IPerfResult) {
+func (i *IPerfResults) Add(ipr *IPerfCSVResult) {
 	if i.BandwidthMap == nil {
 		i.BandwidthMap = map[string]int64{}
 	}
@@ -68,14 +78,17 @@ func (i *IPerfResults) ToTSV() string {
 	return buffer.String()
 }
 
-// NewIPerf parses an IPerf CSV output line into an IPerfResult.
-func NewIPerf(csvLine string) *IPerfResult {
+// NewIPerf parses an IPerf CSV output line into an IPerfCSVResult.
+func NewIPerf(csvLine string) (*IPerfCSVResult, error) {
+	if len(csvLine) == 0 {
+		return nil, errors.New("No iperf output received in csv line")
+	}
 	csvLine = strings.Trim(csvLine, "\n")
 	slice := StrSlice(strings.Split(csvLine, ","))
 	if len(slice) != 9 {
-		framework.Failf("Incorrect fields in the output: %v (%v out of 9)", slice, len(slice))
+		return nil, errors.Errorf("Incorrect fields in the output: %v (%v out of 9)", slice, len(slice))
 	}
-	i := IPerfResult{}
+	i := IPerfCSVResult{}
 	i.date = slice.get(0)
 	i.cli = slice.get(1)
 	i.cliPort = intOrFail("client port", slice.get(2))
@@ -85,7 +98,7 @@ func NewIPerf(csvLine string) *IPerfResult {
 	i.interval = slice.get(6)
 	i.transferBits = intOrFail("transfer port", slice.get(7))
 	i.bandwidthBits = intOrFail("bandwidth port", slice.get(8))
-	return &i
+	return &i, nil
 }
 
 // StrSlice represents a string slice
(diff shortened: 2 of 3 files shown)


mattfenwick:
Test plan:

```
make WHAT=test/e2e/e2e.test

export KUBERNETES_SERVICE_HOST=127.0.0.1
export KUBERNETES_SERVICE_PORT=32768

cat << EOF > kind-config.yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
- role: worker
- role: worker
EOF

kind create cluster --name=three-node --config kind-config.yaml

./_output/local/bin/darwin/amd64/e2e.test \  
 --provider=local \
 --ginkgo.focus="should run iperf3" \
 --kubeconfig=/Users/mfenwick/.kube/config \
 --dump-logs-on-failure=false
```

test/e2e/network/networking_perf.go
```diff
@@ -37,6 +42,150 @@ const (
...
+		_, err = iperf3ServerDeployment(f.ClientSet, f.Namespace.Name)
+		framework.ExpectNoError(err, "deploy iperf3 server deployment")
+
+		_, err = iperf3ServerService(f.ClientSet, f.Namespace.Name)
+		framework.ExpectNoError(err, "deploy iperf3 server service")
+
+		_, err = iperf3ClientDaemonSet(f.ClientSet, f.Namespace.Name)
+		framework.ExpectNoError(err, "deploy iperf3 client daemonset")
+
+		// TODO find method to wait for client pods to come up
```
mattfenwick, line 166: Use a waitForPod method or waitForDaemonset method.

test/e2e/network/networking_perf.go
```diff
@@ -162,6 +318,10 @@ func networkingIPerfTest(isIPv6 bool) {
 	})
 }
 
+var _ = SIGDescribe("Networking IPerf3 IPv4 [Experimental] [Feature:Networking-IPv4] [Slow] [Feature:Networking-Performance]", func() {
```
jayunit100, line 321: can we delete or replace one of the other tests here or do you think 
ite better to add this as new?  

Maybe add a //TODO above about how if this proves useful to delete the older tests as well with it.

test/e2e/network/networking_perf.go
```diff
@@ -37,6 +42,157 @@ const (
 	smallClusterTimeout = 200 * time.Second
 )
 
+func iperf3ServerDeployment(client clientset.Interface, namespace string) (*appsv1.Deployment, error) {
```
jayunit100, line 45: thanks for doing this, this test looks much better then the ones we had
previously from what you posted in the other issue.

test/utils/image/manifest.go
```diff
@@ -218,6 +223,7 @@ func initImageConfigs() map[int]Config {
 	configs[HttpdNew] = Config{dockerLibraryRegistry, "httpd", "2.4.39-alpine"}
 	configs[InvalidRegistryImage] = Config{invalidRegistry, "alpine", "3.1"}
 	configs[IpcUtils] = Config{e2eRegistry, "ipc-utils", "1.0"}
+	configs[Iperf3] = Config{dockerIORegistry, "networkstatic/iperf3", "latest"}
```
jayunit100, line 226: i thought we had an iperf3 image but maybe we didnt.

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
```diff
@@ -27,15 +32,21 @@ import (
...
+	clientset "k8s.io/client-go/kubernetes"
 )
 
 const (
 	// empirically derived as a baseline for expectations from running this test using kube-up.sh.
 	gceBandwidthBitsEstimate = int64(30000000000)
 	// on 4 node clusters, we found this test passes very quickly, generally in less then 100 seconds.
 	smallClusterTimeout = 200 * time.Second
+	// set a baseline for iperf3 bandwidth
+	eightBillion = float64(8_000_000_000)
```
mattfenwick, line 46: TODO set this to a more reasonable number

test/utils/image/manifest.go
```diff
@@ -68,6 +69,7 @@ func initReg() RegistryList {
 	registry := RegistryList{
 		GcAuthenticatedRegistry: "gcr.io/authenticated-image-pulling",
 		DockerLibraryRegistry:   "docker.io/library",
+		DockerIORegistry:        "docker.io",
```
aojea, line 72: with new dockerhub limitations we are trying to avoid it if possible

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +171,198 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+			Start         float64 `json:"start"`
+			End           float64 `json:"end"`
+			Seconds       float64 `json:"seconds"`
+			Bytes         float64 `json:"bytes"`
+			BitsPerSecond float64 `json:"bits_per_second"`
+		} `json:"sum_received"`
+	} `json:"end"`
+}
+
+var _ = SIGDescribe("Networking IPerf3 IPv4 [Experimental] [Feature:Networking-IPv4] [Slow] [Feature:Networking-Performance]", func() {
```
aojea, line 306: this is already running in the CI for ipv4 and for ipv6
https://testgrid.k8s.io/sig-network-kind#sig-network-kind,%20IPv6,%20master , it must work in both

test/utils/image/manifest.go
mattfenwick: Good point ... is there another registry we can keep an iperf3 in?

test/e2e/network/networking_perf.go
mattfenwick: Ah good call, I'll make the same change as in the existing test to support both!

mattfenwick: 👍

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/utils/image/manifest.go
aojea: `"gcr.io/kubernetes-e2e-test-images"`

alejandrox1: Pinging @claudiubelu for help with this one. Claudiu configured the image promotion for test images used for conformance in https://github.com/kubernetes/kubernetes/tree/master/test/images .
Think we should be able to piggy back off of the automation built for that.

claudiubelu: In the past, we used to have an ``iperf`` image, but we've centralized it into agnhost since then. Is there a specific reason you require iperf3? If iperf2 is sufficient, you can simply use the agnhost image instead. If not, you can add it to the agnhost image in ``kubernetes/test/images/agnhost/Dockerfile``, and also the ``Dockerfile_windows`` file too, IMO.

jayunit100:
matt sounds mostly unblocked on this now, thanks to @alejandrox1  for helping out

test/utils/image/manifest.go
mattfenwick: iperf2 doesn't seem to be in the agnhost image:

```
✗ docker run --rm -it --entrypoint sh k8s.gcr.io/e2e-test-images/agnhost:2.21
/ # iperf2
sh: iperf2: not found
```

so I'll try adding iperf3 instead.

aojea:
hmm, we are using iperf to test multicast, that seems is not a thing with iperf3 
https://community.mellanox.com/s/article/iperf--iperf2--iperf3

is there any possibility to keep going with iperf2, that seems the default in alpine?

mattfenwick:
@aojea thanks for the feedback 👍 

> is there any possibility to keep going with iperf2

this is a great callout, I know the whole point of this PR is specifically to take advantage of the much more detailed and informative data present in the iperf3 output, but if we could do this with iperf2 -- totally agree that'd be ideal.

I'll hack on this some more with iperf2 and see how much we can get out of that.

Since you mentioned multicast -- is that important for this test?  I don't see it mentioned explicitly anywhere in this test, are we using it implicitly?

aojea:
> Since you mentioned multicast -- is that important for this test? I don't see it mentioned explicitly anywhere in this test, are we using it implicitly?

It´s an openshift e2e test, well, in this case for OVN-Kubernetes, but I´m happy to add the e2e multicast tests upstream if other CNI want to test multicast , I was going to but honestly I forget about it until I see this
https://github.com/ovn-org/ovn-kubernetes/pull/1741

aojea:
> Since you mentioned multicast -- is that important for this test?

sorry, didn´t read it through, no , multicast has nothing to do with this PR, is just that it needs iperf2

but, you can use the `-e` flag for enhanced reports with iperf2

```
iperf -c localhost -e  -i 1
------------------------------------------------------------
Client connecting to localhost, TCP port 5001 with pid 53231
Write buffer size:  128 KByte
TCP window size: 2.76 MByte (default)
------------------------------------------------------------
[  6] local 127.0.0.1 port 41282 connected with 127.0.0.1 port 5001 (ct=0.06 ms)
[ ID] Interval        Transfer    Bandwidth       Write/Err  Rtry     Cwnd/RTT        NetPwr
[  6] 0.00-1.00 sec  2.97 GBytes  25.5 Gbits/sec  24349/0          0     1215K/31 us  102950713.81
[  6] 1.00-2.00 sec  4.00 GBytes  34.4 Gbits/sec  32786/0          0     2302K/75 us  57297687.89
[  6] 2.00-3.00 sec  4.32 GBytes  37.1 Gbits/sec  35413/0          0     2621K/33 us  140656143.52
[  6] 3.00-4.00 sec  2.58 GBytes  22.1 Gbits/sec  21117/0          0     3197K/38 us  72838090.11
[  6] 4.00-5.00 sec  2.17 GBytes  18.7 Gbits/sec  17787/0          0     3197K/22 us  105971712.00
[  6] 5.00-6.00 sec  2.50 GBytes  21.5 Gbits/sec  20514/0          0     3197K/25 us  107552440.32
[  6] 6.00-7.00 sec  3.76 GBytes  32.3 Gbits/sec  30803/0          1     3197K/33 us  122345782.30
[  6] 7.00-8.00 sec  4.03 GBytes  34.6 Gbits/sec  33027/0          1     3197K/54 us  80165091.56
[  6] 8.00-9.00 sec  2.30 GBytes  19.8 Gbits/sec  18870/0          0     3197K/181 us  13664799.12
[  6] 9.00-10.00 sec  2.18 GBytes  18.8 Gbits/sec  17887/0          3     3197K/30 us  78149495.47
```

test/utils/image/manifest.go
mattfenwick: nvm -- misunderstood iperf, `iperf -v` shows that iperf2 is indeed in agnhost 👍 , and covers our use case so agnhost as-is should be fine!

mattfenwick:
@aojea thanks for the suggestion, that seems to work well!

@jayunit100 we discovered that:

 - the agnhost image already includes iperf2
 - iperf2 has CLI switches which make it produce more detailed reports 

How do you feel about moving forward with iperf2 and enhanced output formats, instead of iperf3?

test/e2e/network/networking_perf.go
```diff
@@ -37,6 +42,157 @@ const (
...
+		framework.ExpectNoError(err)
+		framework.Logf("found %d matching client pods", len(clientPodList.Items))
+		if len(clientPodList.Items) == 0 {
+			framework.Failf("found 0 matching client pods")
+		}
+
+		for _, pod := range clientPodList.Items {
+			podName := pod.Name
+			hostIp := pod.Status.HostIP
+			command := fmt.Sprintf("iperf3 -c iperf3-server -T \"Client on %s\"", hostIp)
```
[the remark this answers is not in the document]
mattfenwick: I don't understand -- I don't see where the original test is adding a certain amount of transfer data. ??

jayunit100:
i think this is a strict improvement on what we have.  If it doesnt require any image changes, it seems like it should be ready?

mattfenwick:
TODO for @mattfenwick add comments (especially a header comment)

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,227 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
 		}
 	})
 })
+
```
jayunit100, line 175: @mattfenwick can you add one of those 'summary' comments like you did in the other PR of how this test works, and also write "TODO remove other tests", so that once this is stable we can throw away the old tests which are mostly unreadable ?

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,227 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+							Operator: v1.TolerationOpEqual,
+							Effect:   v1.TaintEffectNoSchedule,
+						},
+					},
+					Containers: []v1.Container{
+						{
+							Name:    "iperf2-client",
+							Image:   imageutils.GetE2EImage(imageutils.Agnhost),
+							Command: []string{"/bin/sh"},
+							Args:    []string{"-c", "sleep 99999999"},
```
jayunit100, line 275: someone will probably nit this :) so , pre-emptively make it a constant

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,227 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+			return true, nil
+		})
+		framework.ExpectNoError(err, "unable to wait for client pods to come up")
+
+		serverPodList, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).List(context.TODO(), serverPodsListOptions)
+		framework.ExpectNoError(err)
+		if len(serverPodList.Items) != 1 {
+			framework.Failf("expected 1 server pod, found %d", len(serverPodList.Items))
+		}
+		serverPod := serverPodList.Items[0]
```
jayunit100, line 351: explain this one^

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,227 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+			if parsedResults == nil || len(parsedResults) == 0 {
+				framework.Failf("no results parsed from iperf output from client %s", pod.Name)
+			}
+			results := &iperf2NodeResults{
+				Intervals: parsedResults[:len(parsedResults)-1],
+				Total:     parsedResults[len(parsedResults)-1],
+			}
+			nodeResults.Results[pod.Spec.NodeName] = results
+		}
+		framework.Logf("%25s%25s%25s", "From", "To", "Bandwidth (MB/s)")
```
jayunit100, line 386: paste the string output above this

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,227 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+
+		nodeResults := &iperf2Results{
+			ServerNode: serverPod.Spec.NodeName,
+			Results:    map[string]*iperf2NodeResults{},
+		}
+		for _, pod := range clientPodList.Items {
+			podName := pod.Name
+			hostIP := pod.Status.HostIP
+
+			command := fmt.Sprintf(`iperf -e -p 6789 --reportstyle C -i 1 -c iperf2-server -T "Client on %s"`, hostIP)
```
jayunit100, line 368: assuming 6789 is a random port, why this one?  worth a note

test/e2e/network/networking_perf.go
```diff
@@ -27,15 +33,21 @@ import (
...
+	clientset "k8s.io/client-go/kubernetes"
 )
 
 const (
 	// empirically derived as a baseline for expectations from running this test using kube-up.sh.
 	gceBandwidthBitsEstimate = int64(30000000000)
 	// on 4 node clusters, we found this test passes very quickly, generally in less then 100 seconds.
 	smallClusterTimeout = 200 * time.Second
+	// set a baseline for iperf2 bandwidth
+	eightBillion = 8_000_000_000
```
jayunit100, line 47: @eth0XFEED what should we set here?

rikatz: Maybe once defining the baseline we can call this const with the unit (like oneGigaperSecond = 1*1024*etc

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,227 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+		Status: appsv1.DaemonSetStatus{},
+	}
+
+	ds, err := client.AppsV1().DaemonSets(namespace).Create(context.TODO(), spec, metav1.CreateOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("daemonset %s Create API error: %v", spec.Name, err)
+	}
+	return ds, nil
+}
+
```
jayunit100, line 291: comments on these scripts and maybe move the `IPerfResult` to colocate w / them ?

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,227 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+			nodeResults.Results[pod.Spec.NodeName] = results
+		}
+		framework.Logf("%25s%25s%25s", "From", "To", "Bandwidth (MB/s)")
+		for nodeFrom, results := range nodeResults.Results {
+			mb := 1024 * 1024
+			bandwidthMB := math.Round(float64(results.Total.bandwidthBits) / float64(mb))
+			framework.Logf("%25s%25s%25.0f", nodeFrom, nodeResults.ServerNode, bandwidthMB)
+		}
+		for nodeFrom, results := range nodeResults.Results {
+			if results.Total.bandwidthBits < eightBillion {
```
rikatz, line 393: After changing the const name on the bottom, using the same const (with the unit in the name) here will make things more clear

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,227 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+			results := &iperf2NodeResults{
+				Intervals: parsedResults[:len(parsedResults)-1],
+				Total:     parsedResults[len(parsedResults)-1],
+			}
+			nodeResults.Results[pod.Spec.NodeName] = results
+		}
+		framework.Logf("%25s%25s%25s", "From", "To", "Bandwidth (MB/s)")
+		for nodeFrom, results := range nodeResults.Results {
+			mb := 1024 * 1024
+			bandwidthMB := math.Round(float64(results.Total.bandwidthBits) / float64(mb))
```
jayunit100, line 389: maybe just for readability define the bandwidth equation units here since its `bandwidthBits/mb`

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,227 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+					TerminationGracePeriodSeconds: &zero,
+					Containers: []v1.Container{
+						{
+							Name:    "iperf2-server",
+							Image:   imageutils.GetE2EImage(imageutils.Agnhost),
+							Command: []string{"iperf"},
+							Args: []string{
+								"-s",
+								"-p",
+								"6789",
```
jayunit100, line 204: 6789 magic number clean up elsewhere as well

test/e2e/network/networking_perf.go
aojea: there is a `pause` command in agnhost that does exactly that,  just pass the arg `pause` only , don't need to use the command
https://github.com/kubernetes/kubernetes/blob/ee8983705a30e1bcaa17e655a64533a582e250ae/test/images/agnhost/README.md#pause

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,227 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
 		}
 	})
 })
+
+func iperf2ServerDeployment(client clientset.Interface, namespace string) (*appsv1.Deployment, error) {
+	zero := int64(1)
```
aojea, line 177: lol, 0 = 1 ?? :smile:

test/e2e/network/networking_perf.go
```diff
@@ -37,6 +42,157 @@ const (
...
+		clientPodList, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).List(context.TODO(), metav1.ListOptions{
+			LabelSelector: "app=iperf3-client",
+		})
+		framework.ExpectNoError(err)
+		framework.Logf("found %d matching client pods", len(clientPodList.Items))
+		if len(clientPodList.Items) == 0 {
+			framework.Failf("found 0 matching client pods")
+		}
+
+		for _, pod := range clientPodList.Items {
```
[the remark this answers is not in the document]
mattfenwick: 👍

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
mattfenwick: yeah I don't know what happened .. omg :) 

fixed !

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,275 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+20201210141803.882,10.244.2.24,47880,10.96.114.79,6789,3,3.0-4.0,2035417088,16283336704
+20201210141804.879,10.244.2.24,47880,10.96.114.79,6789,3,4.0-5.0,1922957312,15383658496
+20201210141805.881,10.244.2.24,47880,10.96.114.79,6789,3,5.0-6.0,2095316992,16762535936
+20201210141806.882,10.244.2.24,47880,10.96.114.79,6789,3,6.0-7.0,1741291520,13930332160
+20201210141807.879,10.244.2.24,47880,10.96.114.79,6789,3,7.0-8.0,1862926336,14903410688
+20201210141808.878,10.244.2.24,47880,10.96.114.79,6789,3,8.0-9.0,1821245440,14569963520
+20201210141809.849,10.244.2.24,47880,10.96.114.79,6789,3,0.0-10.0,18752208896,15052492511
+*/
+// TODO remove other tests in this file if this test proves to be more stable, informative, and useful
+var _ = SIGDescribe("Networking IPerf2", func() {
```
mattfenwick, line 338: TODO readd tags here:  `[Experimental] [Slow] [Feature:Networking-Performance]`

test/e2e/network/networking_perf.go
```diff
@@ -19,6 +19,10 @@ package network
 // Tests network performance using iperf or other containers.
 import (
 	"fmt"
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
```
aojea, line 24: rearrange the imports please, usually standards first and together

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,275 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+var _ = SIGDescribe("Networking IPerf2", func() {
+	// this test runs iperf2: one pod as a server, and a daemonset of clients
+	f := framework.NewDefaultFramework("network-perf")
+
+	ginkgo.It(fmt.Sprintf("should run iperf2"), func() {
+		nodeList, err := e2enode.GetReadySchedulableNodes(f.ClientSet)
+		framework.ExpectNoError(err)
+
+		serverPodsListOptions := metav1.ListOptions{
+			LabelSelector: "app=iperf2-server",
```
aojea, line 347: you should pass the labels as a parameter to the functions that creates the deployments and daemonsets, ...., otherwise you are defining them in 2 places and it can be easy to go out of sync

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,275 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+		// Make sure the server is ready to go
+		framework.ExpectNoError(e2edeployment.WaitForDeploymentComplete(f.ClientSet, serverDeployment), "unable to wait for server deployment")
+
+		// Make sure the clients are ready to go
+		err = wait.PollImmediate(2*time.Second, 100*time.Second, func() (done bool, err error) {
+			clientPodList, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).List(context.TODO(), clientPodsListOptions)
+			if err != nil {
+				return false, err
+			}
+			if len(clientPodList.Items) < len(nodeList.Items) {
```
aojea, line 375: I think that there are clusters that doesn´t allow to schedule pods in all nodes, but I don´t know exactly how does it work, just something to take into account

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,275 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+		// Get a reference to the server pod for later
+		serverPodList, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).List(context.TODO(), serverPodsListOptions)
+		framework.ExpectNoError(err)
+		if len(serverPodList.Items) != 1 {
+			framework.Failf("expected 1 server pod, found %d", len(serverPodList.Items))
+		}
+		serverPod := serverPodList.Items[0]
+
+		// Get a reference to the client pods for later
+		clientPodList, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).List(context.TODO(), clientPodsListOptions)
```
aojea, line 396: You´ve already get the clientPodList previously in line 369
```
		// Make sure the clients are ready to go
		err = wait.PollImmediate(2*time.Second, 100*time.Second, func() (done bool, err error) {
			clientPodList, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).List(context.TODO(), clientPodsListOptions)
```

you can define the clientPodList out of the loop if you want to reuse it

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,275 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+			Results:    map[string]*IPerf2EnhancedResults{},
+		}
+		// Step 3: iterate through the client pods one by one, running iperf2 in client mode to transfer
+		//   data to the server and back and measure bandwidth
+		for _, pod := range clientPodList.Items {
+			podName := pod.Name
+			nodeName := pod.Spec.NodeName
+			hostIP := pod.Status.HostIP
+
+			command := fmt.Sprintf(`iperf -e -p %d --reportstyle C -i 1 -c iperf2-server -T "Client on %s"`, iperf2Port, hostIP)
```
aojea, line 415: add a comment explaining the iperf flags please
// iperf
// -e enhanced report
// -p iperf2Port destination port
// -reportstyle C ...
// -i 1 report every 1 second interval
// -c iperf2-server connect to the iperf2-server service
..

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +172,275 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+			Dec 10 09:13:05.658: INFO:         abc-control-plane               abc-worker                     1267
+
+		*/
+		framework.Logf("%25s%25s%25s", "From", "To", "Bandwidth (MB/s)")
+		for nodeFrom, results := range nodeResults.Results {
+			framework.Logf("%25s%25s%25d", nodeFrom, nodeResults.ServerNode, results.Total.bandwidthMB())
+		}
+		for clientNode, results := range nodeResults.Results {
+			megabytesPerSecond := results.Total.bandwidthMB()
+			if megabytesPerSecond < iperf2BaselineBandwidthMegabytesPerSecond {
```
aojea, line 441: there are a lot of people these days running kubernetes in raspberry and targeting IOT scenarios that I assume will have very little bandwidth, just food for thoughts

test/e2e/network/util_iperf.go
```diff
@@ -46,6 +52,10 @@ type IPerfResult struct {
 	bandwidthBits int64
 }
 
+func (i *IPerfResult) bandwidthMB() int {
+	return int(math.Round(float64(i.bandwidthBits) / float64(megabyte) / 8))
+}
```
aojea, line 57: the IPerfResult type is using int64, this return an `int` and you are converting it to a float64 :smile:

test/e2e/network/networking_perf.go
mattfenwick: +1 this is a tricky one, the goal is to set the bandwidth cutoff high enough that you know that the cluster's network isn't broken, but low enough that it won't report a working cluster as broken.  

Seems like a tricky balance, any suggestions how we can find that balance?

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
mattfenwick: This node list is from a call to `e2enode.GetReadySchedulableNodes`, that should take care of that right?

test/e2e/network/networking_perf.go
```diff
@@ -34,10 +40,15 @@ const (
...
 	smallClusterTimeout = 200 * time.Second
+	// iperf2BaselineBandwidthMegabytesPerSecond sets a baseline for iperf2 bandwidth of 100 MB/s
+	iperf2BaselineBandwidthMegabytesPerSecond = 100
+	// iperf2Port selects an arbitrary, unique port to run iperf2's client and server on
+	iperf2Port = 6789
 )
 
+// TODO delete iperf-based tests if iperf2 tests prove useful
 // Declared as Flakey since it has not been proven to run in parallel on small nodes or slow networks in CI
-var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-Performance]", func() {
```
mattfenwick, line 40: TODO readd these tags

test/e2e/network/util_iperf.go
mattfenwick: I wish there were a way to just round to an int in golang!!

test/e2e/network/networking_perf.go
mattfenwick: updated the variable name to `readySchedulableNodes`

aojea: Good point, totally missed it, sorry

test/e2e/network/networking_perf.go
mattfenwick: I'm fine with this one either way ... lmk if you want me to change it!

test/e2e/network/util_iperf.go
aojea: why not using int64? int depends on the platform, it can be int32 or int64

test/e2e/network/networking_perf.go
aojea: we save an API call and a check, I think is worth it

test/e2e/network/util_iperf.go
mattfenwick: oh got it!  💯

park-cheif5:
@aka_chiefy @akachiefy

mattfenwick:
The new tests all passed with the test tags removed, so readding them.

Test results:

 - https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/94015/pull-kubernetes-e2e-gce-ubuntu-containerd/1339713838517325824/
 - https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/94015/pull-kubernetes-e2e-kind/1339713838617989120
 - https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/94015/pull-kubernetes-e2e-kind-ipv6/1339713838651543552

test/e2e/network/networking_perf.go
aojea: I think we can remove Experimental and Slow, it takes less than a minute to run, we discussed it somewhere and don´t remember the details but I think that the bar was in 1 minute.
Also, I was running this more some months and is not experimental, 
we have the Feature tag to discriminate

test/e2e/network/networking_perf.go
aojea: I think that we should not target throughput (throughput is part of performance not performance) but correctness , based on that my criteria to pass the test will be :
- keep at least 90Mbps (consider a 100 mbps nic plus a 10% margin)
- less than 3 % of errors

test/e2e/network/networking_perf.go
```diff
@@ -160,3 +173,285 @@ var _ = SIGDescribe("Networking IPerf [Experimental] [Slow] [Feature:Networking-
...
+20201210141803.882,10.244.2.24,47880,10.96.114.79,6789,3,3.0-4.0,2035417088,16283336704
+20201210141804.879,10.244.2.24,47880,10.96.114.79,6789,3,4.0-5.0,1922957312,15383658496
+20201210141805.881,10.244.2.24,47880,10.96.114.79,6789,3,5.0-6.0,2095316992,16762535936
+20201210141806.882,10.244.2.24,47880,10.96.114.79,6789,3,6.0-7.0,1741291520,13930332160
+20201210141807.879,10.244.2.24,47880,10.96.114.79,6789,3,7.0-8.0,1862926336,14903410688
+20201210141808.878,10.244.2.24,47880,10.96.114.79,6789,3,8.0-9.0,1821245440,14569963520
+20201210141809.849,10.244.2.24,47880,10.96.114.79,6789,3,0.0-10.0,18752208896,15052492511
+*/
+// TODO remove other tests in this file if this test proves to be more stable, informative, and useful
+var _ = SIGDescribe("Networking IPerf2 [Experimental] [Slow] [Feature:Networking-Performance]", func() {
```
aojea, line 336: ditto https://github.com/kubernetes/kubernetes/pull/94015/files#r545765363

aojea:
In this case, I think that we can replace the old test  for the newer  ... AFAIK I was the only person running it, and it was broken for a long time before I was able to make it work

mattfenwick:
Awesome sounds good @aojea -- I removed the old ones!

test/e2e/network/networking_perf.go
mattfenwick: This is odd, I don't see error counts with the iperf `--reportstyle C` option, gonna hack around some more on this and see why that's not showing up.

aojea: 🤔 the snippet in the description show errors

mattfenwick: Unfortunately the CSV output doesn't include the errors, let me whip together something to parse the non-CSV output!

aojea: indeed, surprisingly it doesn' have the same output
```
$ iperf -c localhost -e -y C
20201221235135.693,127.0.0.1,55932,127.0.0.1,5001,3,0.0-10.0,53096480768,42477116651
```

```
$ iperf -c localhost -e 
------------------------------------------------------------
Client connecting to localhost, TCP port 5001 with pid 3551429
Write buffer size:  128 KByte
TCP window size: 4.00 MByte (default)
------------------------------------------------------------
[  3] local 127.0.0.1 port 56228 connected with 127.0.0.1 port 5001 (ct=0.05 ms)
[ ID] Interval        Transfer    Bandwidth       Write/Err  Rtry     Cwnd/RTT        NetPwr
[  3] 0.00-10.00 sec  48.3 GBytes  41.5 Gbits/sec  395726/0          0       -1K/23 us  225515464.25

```

mattfenwick: okay put together https://github.com/kubernetes/kubernetes/pull/94015/commits/bf4874a921c258bbb701a6a76680a37d00c2435c to scrape out the error rate -- what do you think @aojea which approach is better?  The parsing code is kinda wonky and fragile but it gets the job done

test/e2e/network/networking_perf.go
aojea: > Also, I was running this more some months and is not experimental,
> we have the Feature tag to discriminate

it seems you´ve removed the Feature tag, we need it for filtering

test/e2e/network/networking_perf.go
aojea: >  The parsing code is kinda wonky and fragile but it gets the job done
indeed, I think that since you implemented both, first we use the CSV one and we let it soak, and we revisit after some time ... my apologies, thought it was simpler

mattfenwick: np, glad we tried it out both ways!  👍

test/e2e/network/networking_perf.go
mattfenwick: 👍

test/e2e/network/networking_perf.go
```diff
@@ -18,145 +18,275 @@ package network
...
+	err = e2edeployment.WaitForDeploymentComplete(client, deployment)
+	if err != nil {
+		return nil, fmt.Errorf("deployment %q failed to complete: %v", deploymentSpec.Name, err)
+	}
+	return deployment, nil
+}
+
+func iperf2ServerService(client clientset.Interface, namespace string, labelKey string, labelValue string) (*v1.Service, error) {
+	service := &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{Name: labelValue},
```
aojea, line 107: this is misleading, it took me a while to realise how iperf clients connects to the service, and is because the service name has the labelValue as name, hence, labelValue has a DNS record pointing to the service IP.
Also, this is no consistent with the other objects created (deployments, pods, containers, ... ) where you are hardcoding the names.

https://github.com/kubernetes/kubernetes/pull/94015/files#diff-c70b0afa39c9d7108ad9cd29eee4a49894807e5044eb82b348919109b6ae8b6eR261

mattfenwick: How would you like this to be organized?  I'm fine with anything 👍

aojea: maybe we can declare the `Names` as constants?

mattfenwick: 👍 yep sounds good !!

aojea: sorry, it seems I didn't explain myself, I didn't mean to replace label key and values, just the `Name` field.
So, in this case, the `Service` `Name` is the same that the `labelValue`, that is the part that is misleading.
I was suggesting to use a new const `iperfServiceName` or something like that and use it to configure the `Service` and to use it as the `iperf` client destination https://github.com/kubernetes/kubernetes/pull/94015/files#diff-c70b0afa39c9d7108ad9cd29eee4a49894807e5044eb82b348919109b6ae8b6eR261

mattfenwick: ahh 👍 got it, yeah that makes sense!  🦖 🦖

aojea:
/retest
this LGTM, @mattfenwick do you plan to do more changes or can we remove the WIP from the title?

mattfenwick:
Removed the WIP from the title , thanks for the review @aojea !

aojea:
/lgtm
Thanks

mattfenwick:
pinging @alejandrox1 to complete the pull review process!

jayunit100:
looks like hit a build flake last time, 
/test pull-kubernetes-e2e-ubuntu-gce-network-policies

aojea:
hmm it flaked on this test
```

Kubernetes e2e suite: [sig-network] Networking IPerf2 [Feature:Networking-Performance] should run iperf2 expand_less | 39s
-- | --
test/e2e/network/networking_perf.go:183 Jan 11 20:59:20.059: Incorrect fields in the output: [] (1 out of 9) test/e2e/network/util_iperf.go:142
```

aojea:
it failed again :thinking:

test/e2e/network/util_iperf.go
```diff
@@ -68,14 +78,14 @@ func (i *IPerfResults) ToTSV() string {
 	return buffer.String()
 }
 
-// NewIPerf parses an IPerf CSV output line into an IPerfResult.
-func NewIPerf(csvLine string) *IPerfResult {
+// NewIPerf parses an IPerf CSV output line into an IPerfCSVResult.
+func NewIPerf(csvLine string) *IPerfCSVResult {
```
aojea, line 82:
```
if len(csvLine) == 0 {
  framework.Failf("No ooutput recieved")
}
```

aojea:
/hold

@mattfenwick we have to wait that the pod for the iperf server is running

The client tries to connect at
> Jan 11 22:06:11.253: INFO: attempting to run command 'iperf  -e -p 6789 --reportstyle C -i 1 -c iperf2-server && sleep 5' in client pod iperf2-clients-9f2n2 (node bootstrap-e2e-minion-group-3849)

and the iperf server is running at
> Jan 11 22:06:17.735: INFO: At 2021-01-11 22:06:05 +0000 UTC - event for iperf2-server-deployment-5c7898669c-jvr5p: {kubelet bootstrap-e2e-minion-group-8lfm} Started: Started container iperf2-server

test/e2e/network/networking_perf.go
```diff
@@ -18,145 +18,279 @@ package network
...
 		}
-		fmt.Println("[begin] Node,Bandwidth CSV")
-		fmt.Println(iperfResults.ToTSV())
-		fmt.Println("[end] Node,Bandwidth CSV")
 
-		for ipClient, bandwidth := range iperfResults.BandwidthMap {
-			framework.Logf("%v had bandwidth %v.  Ratio to expected (%v) was %f", ipClient, bandwidth, expectedBandwidth, float64(bandwidth)/float64(expectedBandwidth))
+		// Make sure the server is ready to go
+		framework.ExpectNoError(e2edeployment.WaitForDeploymentComplete(f.ClientSet, serverDeployment), "unable to wait for server deployment")
+
```
aojea, line 214: we have to wait until the endpoints are ready:

```go
err = wait.Poll(2*time.Second, largeClusterTimeout, func() (done bool, err error) {
	listOptions := metav1.ListOptions{LabelSelector: fmt.Sprintf("%s=%s", discoveryv1beta1.LabelServiceName, serverServiceName)}
	esList, err := f.ClientSet.DiscoveryV1beta1().EndpointSlices(f.Namespace.Name).List(context.TODO(), listOptions)
	framework.ExpectNoError(err, "Error fetching EndpointSlice for Service %s/%s", f.Namespace.Name, serverServiceName)

	if len(esList.Items) == 0 {
		framework.Logf("EndpointSlice for Service %s/%s not found", f.Namespace.Name, serverServiceName)
		return false,nil
	}
        return true, nil
})
framework.ExpectNoError(err, "unable to wait for endpoints for the iperf service")

Comment (mattfenwick) [reply test/e2e/network/networking_perf.go:214]: 👍 thanks for the pointer @aojea, I'll take a look!

**Comment by jayunit100** [comment]
Oh ok Looks like an easy fix then ....

File: test/e2e/network/networking_perf.go (hunk 33 above)
Comment (mattfenwick) [reply test/e2e/network/networking_perf.go:215]: okay just pushed this up, let's see how it does ... :excited: 😄

**Comment by aojea** [comment]
/lgtm
/hold cancel
this time it went through , great :muscle:

**Comment by jayunit100** [comment]
```
Jan 12 20:55:48.583: INFO: ExecWithOptions {Command:[/bin/sh -c iperf  -e -p 6789 --reportstyle C -i 1 -c iperf2-server && sleep 5] Namespace:network-perf-2606 PodName:iperf2-clients-7czbp ContainerName:iperf2-client Stdin:<nil> CaptureStdout:true CaptureStderr:true PreserveWhitespace:false Quiet:false}
Jan 12 20:55:48.583: INFO: >>> kubeConfig: /workspace/.kube/config
Jan 12 20:58:03.369: INFO: Exec stderr: "connect failed: Operation timed out"
Jan 12 20:58:03.369: INFO: output from exec on client pod iperf2-clients-7czbp (node bootstrap-e2e-minion-group-m8bl): 


Jan 12 20:58:03.370: FAIL: No iperf output received in csv line

Full Stack Trace
k8s.io/kubernetes/test/e2e/network.ParseIPerf2EnhancedResultsFromCSV(0x0, 0x0, 0x4ef2341, 0x31, 0xc001bcd4a8)
	test/e2e/network/util_iperf.go:145 +0xe5
k8s.io/kubernetes/test/e2e/network.glob..func21.1()
	test/e2e/network/networking_perf.go:284 +0x1007
k8s.io/kubernetes/test/e2e.RunE2ETests(0xc00177a180)
	test/e2e/e2e.go:130 +0x36c
k8s.io/kubernetes/test/e2e.TestE2E(0xc00177a180)
	test/e2e/e2e_test.go:144 +0x2b
testing.tRunner(0xc00177a180, 0x4ffc6b8)
	GOROOT/src/testing/testing.go:1123 +0xef
created by testing.(*T).Run
```

@mattfenwick , probably another quick fix for you ?

mattfenwick:
/hold

added some more logging to help debug this

aojea:
new agnhost image has merged, let's see if it makes a difference
/test pull-kubernetes-e2e-ubuntu-gce-network-policies

jayunit100:
```=--ginkgo.focus=\[sig-network\]|\[Conformance\]|\[Feature:NetworkPolicy\] --ginkgo.skip=\[Feature:(Networking-IPv6|Example|Federation|PerformanceDNS)\]|DualStack|GCE|SCTP|Disruptive|Serial|SNAT|LB.health.check|LoadBalancer|load.balancer|ESIPP``` <- matt is looking into reproducing this locally i think, fyi those are the test tags

mattfenwick:
I've noticed some test runs show they're on iperf 2.0.13, others are on iperf 2.0.9, due to agnhost image versions.  Does this mean the clusters aren't all running the same agnhost image version?

Look for `iperf version 2.0.13 (21 Jan 2019) pthreads` in the logs.

jayunit100:
if it passes again, id say this is now stable, and we should merge.

mattfenwick:
> if it passes again, id say this is now stable, and we should merge.

It's passed 4 straight times now, and has hit and successfully passed the retry case as well -- we're good to go!

aojea:
Great ... What was the problem?

mattfenwick:
> Great ... What was the problem?

The iperf client command just flakes sometimes, so we added a retry mechanism.

Would love to get a better idea of why the command flakes, but the shell/command/exec function `func (f *Framework) ExecShellInPod(podName string, cmd string) string` doesn't split out an error, I think we're losing some context there.

aojea:
/lgtm
Thanks for following up and solving it, 

> Would love to get a better idea of why the command flakes, but the shell/command/exec function `func (f *Framework) ExecShellInPod(podName string, cmd string) string` doesn't split out an error, I think we're losing some context there.

yeah, this can affect more tests ... this is challenging right? flake hunting is tricky ;)

jayunit100:
@mattfenwick can you file a follow on issue for the `ExecShellInPod` if you think it needs to be looked at by someone from sig-testing?

aojea:
/assign @MrHohn 
This is flaking, is not a required job but scare users that see a job failing in their PRs :smile:

MrHohn (approved):
/approve

PR #94021, opened 2020-08-15
Support specifying custom LB retry period from cloud provider

### LINKED ISSUE #88902: Allow for more flexible retry logic with Cloud Controller Manager
Currently, CCM uses exponential backoff when failing to process a service change. It would be great if this behavior—not only the minimum and maximum delay values—could be adjusted if needed.

https://github.com/kubernetes/kubernetes/blob/8508875e4d400c1454561aa6d879860984bcdf59/pkg/controller/service/controller.go#L54-L58
https://github.com/kubernetes/kubernetes/blob/8508875e4d400c1454561aa6d879860984bcdf59/pkg/controller/service/controller.go#L150

**Why is this needed**:
It is possible to change the minimum and maximum delay values but in some cases that's not enough. Changing those values affects all kinds of service changes, and it would be great if it were possible to adjust that behavior in specific circumstances.

Take for example a new Load Balancer being created. We might want to reduce the max retry delay to just a few seconds long so it's ready within the cluster with minimal delay, but for other types of sync failures, keep the max retry delay set to 5 minutes as to not overload the API endpoint that is being polled.

/sig cloud-provider

/cc @timoreimann — Timo, could you please take a look and make sure this is in line with what we discussed?

timoreimann:
**What type of PR is this?**
/kind feature

This change allows cloud providers to specify a custom retry period by returning a `RetryError`. The purpose is to bypass the work queue-driven exponential backoff algorithm when there is no need to back off.

Specifically, this can be the case when a cloud load balancer operation such as a create or delete is still pending and the cloud API should be polled for completion at a constant interval. A backoff algorithm would not always be reasonable to apply here since there is no API or performance degradation warranting an increasing wait time between API requests.

Fixes #88902

~For now, the PR is meant to provide a starting point for discussing. Hence, I have not invested into adding tests. Once/if we have consent on the general direction of this PR, I am going to complete the missing pieces.~ (Tests have been added.)

```release-note
Support specifying a custom retry period for cloud load-balancer operations
```

Not sure if needed. None so far.

/sig cloud-provider
/cc @andrewsykim


staging/src/k8s.io/cloud-provider/controllers/service/controller.go
```diff
@@ -272,8 +274,15 @@ func (s *Controller) processNextWorkItem() bool {
 		return true
 	}
 
-	runtime.HandleError(fmt.Errorf("error processing service %v (will retry): %v", key, err))
-	s.queue.AddRateLimited(key)
+	var re *api.RetryError
```
andrewsykim, line 277: I might have missed something, I don't see `RetryError` added to `k8s.io/cloud-provider/api`, maybe you forgot to push the commit for this change?

timoreimann: Ah yes indeed 🤦 thanks for pointing out, I just pushed the missing file.

staging/src/k8s.io/cloud-provider/controllers/service/controller.go
```diff
@@ -272,8 +274,15 @@ func (s *Controller) processNextWorkItem() bool {
...
+	if stderrors.As(err, &re) {
```
andrewsykim, line 278: What method of LoadBalancer are we expecting to return RetryError? Asking because many of the errors are wrapped and not sure this check will work even when RetryError is returned.

staging/src/k8s.io/cloud-provider/controllers/service/controller.go
```diff
@@ -18,6 +18,7 @@ package service
 
 import (
 	"context"
+	stderrors "errors"
```
andrewsykim, line 21: would prefer this to be just "errors" and alias the one from apimachinery to be "apierrors"

andrewsykim (commented):
cc @MrHohn @bowei @cheftako

staging/src/k8s.io/cloud-provider/controllers/service/controller.go
andrewsykim: > For now, the PR is meant to provide a starting point for discussing. Hence, I have not invested into adding tests. Once/if we have consent on the general direction of this PR, I am going to complete the missing pieces.

Ah sorry, I missed this from the PR description. This seems pretty reasonable to me. @cheftako @bowei @MrHohn thoughts?

staging/src/k8s.io/cloud-provider/controllers/service/controller.go
timoreimann: 👍 done

staging/src/k8s.io/cloud-provider/controllers/service/controller.go
timoreimann: >What method of LoadBalancer are we expecting to return `RetryError`?

To add to that: for now, `RetryError` responses are only honored from `EnsureLoadBalancer` calls. It would be trivial to also do so for `GetLoadBalancer` and `EnsureLoadBalancerDeleted` (by replacing `%v` with `%w` in the right places as well), though I am not sure how useful that might be. It definitely isn't for DigitalOcean, so I tried to keep things minimalistic for now.

MrHohn: +1, enabling cloud providers to have control over how the retry should happen sounds reasonable to me as well.

andrewsykim (commented):
/ok-to-test

timoreimann:
@andrewsykim @MrHohn does this proposal warrant a KEP, or would it suffice to move forward with the PR? In the latter case, I'd like to invest additional time to update the tests and bring the code into a mergeable state.

andrewsykim:
> @andrewsykim @MrHohn does this proposal warrant a KEP, or would it suffice to move forward with the PR? In the latter case, I'd like to invest additional time to update the tests and bring the code into a mergeable state.

I don't think we need a KEP for this, let's get the tests added and try to merge this for v1.20

staging/src/k8s.io/cloud-provider/controllers/service/controller_test.go
```diff
@@ -1077,22 +1081,24 @@ func TestSyncService(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		ctx, cancel := context.WithCancel(context.Background())
-		defer cancel()
+		t.Run(tc.testName, func(t *testing.T) {
```
timoreimann, line 1084: Note to reviewers: drive-by refactoring wrapping a `t.Run()` around the test body so that test cases can be executed and managed individually.

staging/src/k8s.io/cloud-provider/cloud.go
```diff
@@ -131,19 +131,25 @@ func GetInstanceProviderID(ctx context.Context, cloud Interface, nodeName types.
 // irrespective of the ImplementedElsewhere error. Additional finalizers for
 // LB services must be managed in the alternate implementation.
 type LoadBalancer interface {
-	// TODO: Break this up into different interfaces (LB, etc) when we have more than one type of service
 	// GetLoadBalancer returns whether the specified load balancer exists, and
 	// if so, what its status is.
 	// Implementations must treat the *v1.Service parameter as read-only and not modify it.
-	// Parameter 'clusterName' is the name of the cluster as presented to kube-controller-manager
+	// Parameter 'clusterName' is the name of the cluster as presented to kube-controller-manager.
+	// TODO: Break this up into different interfaces (LB, etc) when we have more than one type of service
```
timoreimann, line 138: Note to reviewers: Moved the `TODO` down to please GoDoc linters (function/method name should always come first).

timoreimann:
After a two year hiatus on this PR, I finally managed to get back to it and do the remaining work (primarily adding tests).

To facilitate testing, I needed to create a custom test double queue that allowed me to spy on added keys/items. I decided to put that into a central-ish place for potential reuse going forward. If a more local placement is rather preferred though, let me know and I move it closer to the system-under-test.
For what it's worth, I draw inspiration for `SpyWorkQueue` from [this fakeQueue implementation](https://github.com/kubernetes/kubernetes/blob/76fe4039e5a50ceccce94f5c7bd94b6152ee9086/pkg/controller/cronjob/cronjob_controllerv2_test.go#L1350-L1359). If we do want to stick with my reusable test queue, then `fakeQueue` could be refactored away accordingly. Will certainly wait to do anything in that direction though until we have consensus on this PR.

@andrewsykim @MrHohn are you able to take a look at this again? Will _promise_ to get to any feedback / comments in a very timely fashion this time around.

timoreimann:
@andrewsykim I moved the test queue bit into cloud-provider for now to reduce the scope of the review. If the approach finds general consent, I can follow up with another PR to move `SpyWorkQueue` to a more central location (and update the other caller I discovered and draw inspiration from at the same time).

timoreimann:
@MrHohn would very much appreciate a review from you as well in case you have capacity for this rather small-ish PR.

andrewsykim:
@timoreimann I don't think we should include the SpyWorkQueue implementation in the cloud provider package. I would prefer we have it in a central location first and import it here aftewards.

andrewsykim:
Sorry -- I didn't realize the queue implementation was just for testing, I think it's fine to leave it there. I would remove the `testing` package and name the queue something like `fakeWorkQueue` to keep the implementation close to the test using it

andrewsykim (commented):
/approve
/assign @aojea 
/assign @swetharepakula 

LGTM looking for a second set of eyes from @aojea or @swetharepakula

dims:
cc @nckturner @josh-ferrell

timoreimann:
hey @aojea @swetharepakula 👋 appreciate if you could get some 👀 s on this one. Would love to see my PR getting merged eventually. :)

staging/src/k8s.io/cloud-provider/api/retry_error.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"time"
+)
+
+// RetryError indicates that a service reconciliation should be retried after a
+// fixed duration (as opposed to backing off exponentially).
+type RetryError struct {
+	msg        string
+	retryAfter time.Duration
```
aojea, line 27: what should be the interpretation of 0?

staging/src/k8s.io/cloud-provider/controllers/service/controller_test.go
```diff
@@ -2281,3 +2362,66 @@ func (l *fakeNodeLister) Get(name string) (*v1.Node, error) {
 	}
 	return nil, nil
 }
+
+type fakeServiceLister struct {
```
aojea, line 2366: you don't need to create these mocks, you can use a cache
```
		serviceCache := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
		serviceLister := v1listers.NewServiceLister(serviceCache)
		for i := range test.services {
			if err := serviceCache.Add(test.services[i]); err != nil {
				t.Fatalf("%s unexpected service add error: %v", test.name, err)
			}
		}
```

staging/src/k8s.io/cloud-provider/api/retry_error.go
timoreimann: There'd be no special interpretation. Instead, the retry would be immediate (see also [where the value is used](https://github.com/kubernetes/kubernetes/pull/94021/files#diff-f167d56c014f674b14a82e27dd7cf48484e362e08909dc016d63c5270da6b7a0R296)).

The need to retry right away may be uncommon or even rare, but I personally wouldn't want to disallow it. Maybe a user's network is very slow, or there are already some natural / drive-by delays that don't warrant another extra wait at the client side?

I think a zero delay can be legitimate, but let me know if you think differently.

staging/src/k8s.io/cloud-provider/controllers/service/controller_test.go
timoreimann: That's much better -- adjusted, thanks!

timoreimann:
@aojea PTAL.

timoreimann:
(Updated the copyright year to 2023 real quick)

staging/src/k8s.io/cloud-provider/api/retry_error.go
aojea: > I think a zero delay can be legitimate, but let me know if you think differently.

I think is ok, I just wanted to double check we all have the same interpretation

aojea:
/lgtm
/approve

Thanks

PR #94023, opened 2020-08-15
Fix building with GOFLAGS=-v

odinuge:
**What type of PR is this?**

/kind bug

With go1.15 running go version with "-v" fails, and we should support
it. Commands like `make all WHAT=cmd/kubeadm GOFLAGS=-v` (that we have as examples in the makefile) require this change to keep working.

Relevant change in go: https://go-review.googlesource.com/c/go/+/221397

Current behavior in master:

```
> make all WHAT=cmd/kubeadm GOFLAGS=-v
go version: flags can only be used with arguments
/home/odin/go/src/k8s.io/kubernetes/hack/lib/golang.sh: line 471: go_version[2]: unbound variable
/home/odin/go/src/k8s.io/kubernetes/hack/lib/golang.sh: line 471: go_version[2]: unbound variable
go version: flags can only be used with arguments
/home/odin/go/src/k8s.io/kubernetes/hack/lib/golang.sh: line 471: go_version[2]: unbound variable
/home/odin/go/src/k8s.io/kubernetes/hack/lib/golang.sh: line 471: go_version[2]: unbound variable
!!! [0815 17:55:28] Call tree:
!!! [0815 17:55:28]  1: hack/make-rules/build.sh:27 kube::golang::build_binaries(...)
make[1]: *** [Makefile.generated_files:195: _output/bin/prerelease-lifecycle-gen] Error 1
make: *** [Makefile:546: generated_files] Error 2
```

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/hack/lib/golang.sh
+++ b/hack/lib/golang.sh
@@ -465,7 +465,7 @@ EOF
   fi
 
   local go_version
-  IFS=" " read -ra go_version <<< "$(go version)"
+  IFS=" " read -ra go_version <<< "$(GOFLAGS='' go version)"
   local minimum_go_version
   minimum_go_version=go1.15.0
   if [[ "${minimum_go_version}" != $(echo -e "${minimum_go_version}\n${go_version[2]}" | sort -s -t. -k 1,1 -k 2,2n -k 3,3n | head -n1) && "${go_version[2]}" != "devel" ]]; then
@@ -777,7 +777,7 @@ kube::golang::build_binaries() {
   (
     # Check for `go` binary and set ${GOPATH}.
     kube::golang::setup_env
-    V=2 kube::log::info "Go version: $(go version)"
+    V=2 kube::log::info "Go version: $(GOFLAGS='' go version)"
 
     local host_platform
     host_platform=$(kube::golang::host_platform)


liggitt:
I don't mind this as a workaround, but this seems like a regression in go1.15 if it was tolerated in previous versions

``` 
GOFLAGS
	A space-separated list of -flag=value settings to apply
	to go commands by default, when the given flag is known by
	the current command. Each entry must be a standalone flag.
	Because the entries are space-separated, flag values must
	not contain spaces. Flags listed on the command line
	are applied after this list and therefore override it.
``` 

Mind opening an issue at golang/go if this worked as documented with <go1.15?

liggitt:
Actually, it looks intentional:

> Flag parsing
>
> Various flag parsing issues in go test and go vet have been fixed. Notably, flags specified in GOFLAGS are handled more consistently, and the -outputdir flag now interprets relative paths relative to the working directory of the go command (rather than the working directory of each individual test).

liggitt:
I think that means we should update any places we use GOFLAGS with argumentless flags?

justaugustus:
I guess testing/release are the closest SIGs for this
/sig testing release
/lgtm

PR #94025, opened 2020-08-15
Add logging for making it easier to debug e2e flakes

gnufied:
This PR introduces logging for making it easier to debug pods that did not start running within expected timeout.

The purpose of logging pod's namespace/name, UID and PVC name (if available) is to build tools that can extract this information  automatically and more intelligently point to pods that failed to start and correlate that information with other bits of information in controller-manager or kubelet logs.

/sig storage
/kind cleanup
/release-note-none

### DIFF
--- a/test/e2e/framework/log.go
+++ b/test/e2e/framework/log.go
@@ -26,6 +26,8 @@ import (
 	"github.com/onsi/ginkgo"
 
 	// TODO: Remove the following imports (ref: https://github.com/kubernetes/kubernetes/issues/81245)
+	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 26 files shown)


gnufied:
cc @msau42 @jsafrane

jsafrane:
I'm not sure that adding extra logs to all test is going to work in the long run - there will always be some calls where we forget to add the logs.

Can you add more context to errors returned from `e2epod.WaitForPodNameRunningInNamespace` / `WaitForPodRunningInNamespaceSlow` / `CreateSecPod` / `CreatePod` so `ExpectNoError(err)` is more useful?

gnufied:
@jsafrane that is a good idea. I will try and convert this PR to do that.

PR #94028, opened 2020-08-15
Mixed protocol support for Services with type=LoadBalancer

### LINKED ISSUE #23880: Support mixed protocols in service.type=loadbalancer
It should be possible to use a single IP to direct traffic to multiple protocols with a single Service of Type=Loadbalancer. We just need to:
1. Promote ephemeral to static ip
2. Create another forwarding rule with the same static ip, but a different protocol/port

Unfortunately we need this dance because a single forwarding rule only supports 1 protocol.

When we do this we should make sure the firewall rules are opened up for the right protocol: https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/gce/gce.go#L927 (that just takes the first protocol for the firewall rule). 

Is there a reason we didn't do this initially?

janosi:
**What type of PR is this?**
/kind feature
/kind api-change

This PR implements the following KEP: https://github.com/kubernetes/enhancements/commit/4ccf8f53d9ffa4ffdb98afecfc5eb0ed4f5550f3

It enables the usage of different protocol values in the same LoadBalancer Service.
It extends the Service's Status definition so cloud provider implementations can include more details about the status of the cloud load balancer into that.

KEP issue: https://github.com/kubernetes/enhancements/issues/1435

Fixes https://github.com/kubernetes/kubernetes/issues/23880

```release-note
The usage of mixed protocol values in the same LoadBalancer Service is possible if the new feature gate MixedProtocolLBService is enabled.
"action required"
The feature gate is disabled by default. The user has to enable it for the API Server.
```

```docs
- [KEP]: https://github.com/kubernetes/enhancements/commit/4ccf8f53d9ffa4ffdb98afecfc5eb0ed4f5550f3
```


pkg/apis/core/types.go
```diff
@@ -3486,6 +3486,10 @@ type ServiceStatus struct {
 	// if one is present.
 	// +optional
 	LoadBalancer LoadBalancerStatus
+
+	//Current service state
+	// +optional
+	Conditions []ServiceCondition
```
thockin, line 3492: The KEP says: "Our feature does not introduce new values or new fields"

 am not saying that adding a field is wrong, but it is not part of the committed KEP.  I re-read the discussion on the KEP and we agreed to this in theory, but it was never updated

pkg/apis/core/types.go
```diff
@@ -3508,6 +3512,14 @@ type LoadBalancerIngress struct {
 	// (typically AWS load-balancers)
 	// +optional
 	Hostname string
+
+	//The list has one entry per service port in the manifest
```
thockin, line 3516: nit: comments want a " " between `//` and text

thockin: Godoc style starts with the field name

thockin: We need to tighten the spec.  This field is optional, but if it is provided, there must be one entry per service port, mapped by port number and protocol (avoid the word "manifest, tooooo vague).

pkg/apis/core/types.go
```diff
@@ -3508,6 +3512,14 @@ type LoadBalancerIngress struct {
...
+	// +optional
+	PortStatuses []PortStatus
```
thockin, line 3518: name "ports"

pkg/apis/core/types.go
```diff
@@ -3508,6 +3512,14 @@ type LoadBalancerIngress struct {
...
+
+	//A human readable message indicating details about the condition of the load-balancer.
```
thockin, line 3520: Is human-readable what we want here?  We have conditions and ports which have messages - what is THIS for that isn't better in one of those?

pkg/apis/core/types.go
```diff
@@ -5351,3 +5363,30 @@ type TopologySpreadConstraint struct {
 	// +optional
 	LabelSelector *metav1.LabelSelector
 }
+
+// PortStatus represents the condition of a service port condition
+type PortStatus struct {
+	Port     int32
```
thockin, line 5369: comments on all these

pkg/apis/core/types.go
```diff
@@ -5351,3 +5363,30 @@ type TopologySpreadConstraint struct {
...
+	Protocol Protocol
+	Ready    bool
+	Message  string
```
thockin, line 5372: Especially comments on this.  What is it for, and how should clouds use it?

pkg/apis/core/types.go
```diff
@@ -5351,3 +5363,30 @@ type TopologySpreadConstraint struct {
 	// +optional
 	LabelSelector *metav1.LabelSelector
 }
+
+// PortStatus represents the condition of a service port condition
+type PortStatus struct {
+	Port     int32
+	Protocol Protocol
+	Ready    bool
```
thockin, line 5371: What does this mean?  Does it mean traffic is flowing or that it has been configured already or just that it is accepted and being configured?  Is "ready" the best name?

pkg/apis/core/types.go
thockin: We should use `metav1.Condition` from staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go

pkg/apis/core/validation/conditional_validation.go
```diff
@@ -226,3 +228,43 @@ func podSCTPFields(podSpec *api.PodSpec, fldPath *field.Path) []*field.Path {
...
+}
+
+func serviceHasMixedProtocols(service *api.Service) bool {
+	if service == nil {
+		return false
+	}
+	var protocol api.Protocol = "init"
+	for _, port := range service.Spec.Ports {
+		if protocol == "init" {
+			if port.Protocol == "" {
```
thockin, line 254: You should not need to handle "" - that will be defaulted

thockin: This whole thing should be as easy as:

```
protos := map[string]bool{}
for _, port := range service.Spec.Ports {
    protos[string(port.Protocol)] = true
}
return len(protos) > 1
```

pkg/apis/core/validation/validation.go
```diff
@@ -4232,18 +4232,12 @@ func ValidateService(service *core.Service, allowAppProtocol bool) field.ErrorLi
 
 	if service.Spec.Type == core.ServiceTypeLoadBalancer {
```
thockin, line 4233: I think this whole block can go away - the same check (`supportedPortProtocols.Has`) exists in validateServicePort() already.  Please double-check that I am not missing something

staging/src/k8s.io/api/core/v1/types.go
```diff
@@ -3971,6 +3977,15 @@ type LoadBalancerIngress struct {
 	// (typically AWS load-balancers)
 	// +optional
 	Hostname string `json:"hostname,omitempty" protobuf:"bytes,2,opt,name=hostname"`
+
+	//The list has one entry per service port in the manifest
+	// +listType=map
```
thockin, line 3982: Why not atomic?  Do we really need to patch individual records?  The key for the map is multi-field, which is unpleasant

staging/src/k8s.io/api/core/v1/types.go
```diff
@@ -6104,3 +6119,42 @@ const (
...
 	PortForwardRequestIDHeader = "requestID"
 )
+
+// PortStatus represents the condition of a service port condition
+type PortStatus struct {
+	// The port that will be exposed by this service.
+	Port int32 `json:"port" protobuf:"varint,1,opt,name=port"`
+	// The IP protocol for this port. Supports "TCP", "UDP", and "SCTP".
+	// Default is TCP.
+	// +optional
```
thockin, line 6129: If this is optional with a default, you need to write the rule in defaults.go to set it.

pkg/apis/core/types.go
janosi: > but it is not part of the committed KEP. I re-read the discussion on the KEP and we agreed to this in theory, but it was never updated

Hmmm, we have in the KEP the following: 
```
In order to provide a way for the CPI to indicate port statuses to the user we would add the following new portStatus list of port statuses to Service.status.loadBalancer.ingress, so the CPI can indicate the status of the LB. Also we add a conditions list of conditions to the Service.status.loadBalancer, with the first official condition LoadBalancerMixedProtocolNotSupported defined.
```

And then there is the specification of these fields, too.

Am I missing something?

pkg/apis/core/types.go
janosi: > Godoc style starts with the field name

[Here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#edit-typesgo) I read this

> Comments for types should not contain the type name; API documentation is generated from these comments and end-users should not be exposed to golang type names.

pkg/apis/core/types.go
janosi: Yeah, so from the KEP.

> "ready": {
      "description": "Specifies whether the port was configured for the load-balancer.",
      "type": "boolean"

pkg/apis/core/types.go
janosi: So, should I remove this API change (ServiceStatus) from this PR? I put it in here, as it is part of the KEP.

thockin: You are right, I guess I skimmed too fast.  Sorry.

pkg/apis/core/types.go
thockin: What about "configured" or maybe flip it and make it `error string` which references a condition name ("" means no error)

pkg/apis/core/validation/conditional_validation.go
janosi: Yeah. I guess, next time I should just leave the implementation of stuff to others better in this.

pkg/apis/core/types.go
janosi: (I just wonder how I should interpret all caps here).
Well, this is the part of the agreed and accepted KEP. Which has been reviewed and accepted by people much better in K8s API design than I.
Of course things change and I am OK to remove it from here. Just like the whole API implementation as said above.

thockin: /KEPs are not written in stone, and API reviews can tweak the exact structure.  I am asking what this particular field is for that is not already covered by the other messages?

pkg/apis/core/types.go
liggitt: > So, I started using metav1.Condition there instead of ServiceConditions, but as I can see, bazel-test fails due to that https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/94028/pull-kubernetes-bazel-test/1296919306033958912

chatted in slack, but for posterity, looks like metav1.Condition types should be added to https://github.com/kubernetes/kubernetes/blob/c3b888f647f99eba246468ac3e13afc49542c06d/pkg/controlplane/import_known_versions_test.go#L53-L73

janosi: @liggitt Thank you for your help. Indeed, it is OK now.

pkg/apis/core/types.go
```diff
@@ -5351,3 +5360,23 @@ type TopologySpreadConstraint struct {
...
 }
+
+// PortStatus represents the condition of a service port
+type PortStatus struct {
+	// Port is the port number of the service port of which status is recorded here
+	Port int32
+	// Protocol is the protocol of the service port of which status is recorded here
+	Protocol Protocol
+	// Error is to record any problem with the service port
+	Error string
```
thockin, line 5371: Is the intention to populate this status block ONLY if there is an error?  Or to poplulate always and sometimes error is empty?  If the latter (which is what I expect) we should make error a pointer (optional)

pkg/apis/core/validation/conditional_validation.go
```diff
@@ -226,3 +228,29 @@ func podSCTPFields(podSpec *api.PodSpec, fldPath *field.Path) []*field.Path {
...
+// validateMixedProtocolLBSVC checks if the old Service has type=LoadBalancer and whether the Service has different Protocols
+// on its ports. If the MixedProtocolLBSVC feature flag is disabled the usage of different Protocols in the new Service is
+// valid only if the old Service has different Protocols, too.
+func validateMixedProtocolLBSVC(service, oldService *api.Service) (errs field.ErrorList) {
+	if service.Spec.Type != api.ServiceTypeLoadBalancer {
+		return
+	}
+	if !utilfeature.DefaultFeatureGate.Enabled(features.MixedProtocolLBSVC) {
+		if serviceHasMixedProtocols(service) && !serviceHasMixedProtocols(oldService) {
+			errs = append(errs, field.Invalid(field.NewPath("spec.ports"), service.Spec.Ports, "cannot create an external load balancer with mix protocols"))
```
thockin, line 241: nit: "may not contain more than 1 protocol when `type` is 'LoadBalancer'"

thockin: note the quoting

pkg/features/kube_features.go
```diff
@@ -642,6 +642,12 @@ const (
 	//
 	// Disables Accelerator Metrics Collected by Kubelet
 	DisableAcceleratorUsageMetrics featuregate.Feature = "DisableAcceleratorUsageMetrics"
+
+	// owner: @janosi
+	// alpha: v1.20
+	//
+	// Enables the usage of different protocols in the same Service with type=LoadBalancer
+	MixedProtocolLBSVC featuregate.Feature = "MixedProtocolLBSVC"
```
thockin, line 650: minor: s/SVC/Service/g

thockin (commented):
My comments are minor - this looks reasonable to me.

pkg/apis/core/types.go
janosi: The intention is to use it in all cases. Modified the Error to string pointer.

pkg/apis/core/validation/conditional_validation.go
janosi: Fixed.

pkg/features/kube_features.go
janosi: Fixed.

pkg/apis/core/validation/conditional_validation.go
```diff
@@ -226,3 +228,29 @@ func podSCTPFields(podSpec *api.PodSpec, fldPath *field.Path) []*field.Path {
...
+// validateMixedProtocolLBService checks if the old Service has type=LoadBalancer and whether the Service has different Protocols
+// on its ports. If the MixedProtocolLBService feature flag is disabled the usage of different Protocols in the new Service is
+// valid only if the old Service has different Protocols, too.
+func validateMixedProtocolLBService(service, oldService *api.Service) (errs field.ErrorList) {
+	if service.Spec.Type != api.ServiceTypeLoadBalancer {
+		return
+	}
+	if !utilfeature.DefaultFeatureGate.Enabled(features.MixedProtocolLBService) {
+		if serviceHasMixedProtocols(service) && !serviceHasMixedProtocols(oldService) {
+			errs = append(errs, field.Invalid(field.NewPath("spec.ports"), service.Spec.Ports, "may not contain more than 1 protocol when type is 'LoadBalancer'"))
```
thockin, line 241: this should be `field.NewPath("spec", "ports")`

thockin (commented):
Almost there!

pkg/apis/core/types.go
```diff
@@ -5351,3 +5360,23 @@ type TopologySpreadConstraint struct {
...
 	LabelSelector *metav1.LabelSelector
 }
+
+// PortStatus represents the condition of a service port
+type PortStatus struct {
+	// Port is the port number of the service port of which status is recorded here
+	Port int32
+	// Protocol is the protocol of the service port of which status is recorded here
+	Protocol Protocol
+	// Error is to record any problem with the service port
```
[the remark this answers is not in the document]
janosi: Actually that made me think more about it. 
I wonder if a "Status" field would be better than that "Error".  A "Status" could be set to "Enabled" or "Active" explicitly when the port is OK, so it would not be the missing "Error" that would show that state.
Another value I could imagine as an LB provider agnostic value is the "Protocol not supported".
Then there could be the "Port value is outside of the allowed port range".
And then maybe the "Not applicable with the configured set of ports"

And of course there is the option to put here any provider specific status message about the given port.

I.e. we would provide some pre-defined constant values for the most straightforward and common cases. But we would not restrict the possible set of values to those. 

Also then I think we would need a new condition? I mean, for the case when mixed protocols are supported, but there is a problem with an actual port definitions in the LB.

What do you think?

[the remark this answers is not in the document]
thockin: So far we have gotten away without detailed error in the API, and I think we should try to stay there, lest we end up enumerating all sorts of hypotheticals.

What about a single condition that says "one or more ports are not successfully configured", and a per-port error (as you specced).  The error can be a condition.type style name (FooBar or company.com/specific-error).  If we need to define a SMALL set of common values, we can (e.g. MixedProtocolNotSupported).

Does that suffice?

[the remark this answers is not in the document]
janosi: Yeah. So I made the changes. Let's see if I interpreted right.
- the condition is about port status in general
- as I understand PortStatus would be a place for error reports, so I added Error back as mandatory field
- I have not defined any error value for the time being. Providers can use the Error field for their own values.

pkg/apis/core/validation/conditional_validation.go
janosi: Fixed.

pkg/apis/core/types.go
```diff
@@ -3508,6 +3512,11 @@ type LoadBalancerIngress struct {
 	// (typically AWS load-balancers)
 	// +optional
 	Hostname string
+
+	// Ports is a list of error records of service ports
+	// If used every port defined in the service must have an entry in it
```
thockin, line 3517: really minor and if it is the only fixup I won't block:

"If used, every port defined in the service must have an entry" - the comma make it easier to read and "in it" is not needed

pkg/apis/core/types.go
thockin: This question is back.  Let's be clear - if there's not an error on this port the struct is still populated in status...ports[] but the Error field is blank, right?

If so, the field itself should be optional and omitempty.

pkg/apis/core/types.go
```diff
@@ -5394,3 +5403,23 @@ type TopologySpreadConstraint struct {
 	// +optional
 	LabelSelector *metav1.LabelSelector
 }
+
+// PortStatus represents the error condition of a service port
+type PortStatus struct {
+	// Port is the port number of the service port of which status is recorded here
+	Port int32
+	// Protocol is the protocol of the service port of which status is recorded here
+	Protocol Protocol
```
thockin, line 5412: Will this default to TCP like other instances of Protocol or is it just required?  Given that this is not a user-oriented API it's fine to not default (I just wanted to ask)

pkg/apis/core/types.go
```diff
@@ -5394,3 +5403,23 @@ type TopologySpreadConstraint struct {
...
+	// Error is to record the problem with the service port.
```
thockin, line 5413: We need to spec the format.  I suggest a condition.type style name (FooBar or company.com/specific-error) where we reserve FooBar style names for a TBD list of conditions.. Maybe we seed that list with "MixedProtocolNotSupported" ?

pkg/apis/core/types.go
```diff
@@ -3508,6 +3512,11 @@ type LoadBalancerIngress struct {
 	// (typically AWS load-balancers)
 	// +optional
 	Hostname string
+
+	// Ports is a list of error records of service ports
+	// If used every port defined in the service must have an entry in it
+	// +optional
+	Ports []PortStatus
```
thockin, line 3519: Will we validate that this list is identical to service ports? That seems problematic for edits to spec.Ports, where thi would not be in sync.  Maybe change "must" to "should" in the above?

pkg/apis/core/types.go
janosi: Thank you, fixed.

pkg/apis/core/types.go
janosi: It is just required. Actually, I have not thought about defining a default value for it.

pkg/apis/core/types.go
janosi: I added a more verbose description and stole the validation of condition.types from metav1. Please check.

pkg/apis/core/types.go
janosi: No, I have not planned validation this far. I changed it to "should".

pkg/apis/core/types.go
```diff
@@ -3480,12 +3480,23 @@ const (
 	ServiceExternalTrafficPolicyTypeCluster ServiceExternalTrafficPolicyType = "Cluster"
 )
 
+// These are valid conditions of a service.
+const (
+	// LoadBalancerPortsNotConfiguredSuccessfully represents the status of the requested ports
+	// on the same cloud Load Balancer instance.
+	LoadBalancerPortsNotConfiguredSuccessfully = "LoadBalancerPortsNotConfiguredSuccessfully"
```
thockin, line 3487: sorry to pick: "LoadBalancerPortsNotConfiguredSuccessfully" vs "LoadBalancerPortsError"  ?

thockin (commented):
naming point then LGTM

pkg/apis/core/types.go
janosi: Yep, no problem. It supposed to be a nice API :) Fixed.

pkg/apis/core/validation/conditional_validation.go
```diff
@@ -0,0 +1,270 @@
...
+	netutils "k8s.io/utils/net"
+)
+
+// ValidateConditionalService validates conditionally valid fields. allowedIPFamilies is an ordered
+// list of the valid IP families (IPv4 or IPv6) that are supported. The first family in the slice
+// is the cluster default, although the clusterIP here dictates the family defaulting.
+func ValidateConditionalService(service, oldService *api.Service, allowedIPFamilies []api.IPFamily) field.ErrorList {
+	var errs field.ErrorList
+	// If the SCTPSupport feature is disabled, and the old object isn't using the SCTP feature, prevent the new object from using it
+	if !utilfeature.DefaultFeatureGate.Enabled(features.SCTPSupport) && len(serviceSCTPFields(oldService)) == 0 {
```
aojea: this was removed 2 weeks ago 
https://github.com/kubernetes/kubernetes/pull/95566/
SCTP is GA we shouldn't check for the featuregate, right?

thockin (commented):
Thanks!

/lgtm
/approve

andrewsykim:
From the KEP about kube-proxy:

> The kube-proxy should use the port status information from `Service.status.loadBalancer.ingress` in order not to allow traffic to those ports that could not be opened by the load balancer either. 

Will this be addressed in a follow-up PR?

andrewsykim:
And for the cloud providers that don't support mixed protocol, should they handle setting the condition in v1.20 or in a future release (ideally this should just be done out-of-tree)?

thockin:
Ohh, good catch

thockin:
We should open a bug on each provider, probably?

thockin:
Removing LGTM until we answer these

janosi:
Yep, I did not want to hide these :)
OK, so my thinking so far was, that the new service condition and port status fields are optional, so it is up to the cloud provider to set those. 
And because those are optional, kubelet should also be able to live without those. I.e. executing its current functionality. 
So, following this mindpath I came to the conclusion that this PR is the core functionality that can be implemented independently from the kubelet and cloud provider implementations. 
If a cloud provider does not support mixed protocols the feature gate can be kept in a turned off state, so LB services with mixed protocols are blocked by the API server, like without this feature.

It was my way of thinking. Please correct me if you see problems with it, I am sure you have a wider set of considerations than me.

Thank you!

thockin:
> the new service condition and port status fields are optional, so it is up to the cloud provider to set those

We could seed the discussion by changing all of the built-in providers to accept the first protocol and automatically set the error on any additional protocol-ports.  Then it becomes up to providers to implement mixed support and NOT set the error.  Is that too far?  Otherwise the UX is "silence" which implies success.

> If a cloud provider does not support mixed protocols the feature gate can be kept in a turned off state

Only until it GA's :)

As for kube-proxy, it's not clear to me, without auditing, that it will do the right thing for mixed-protocol LB VIPs (we know NodePort works).  Beyond that, making it NOT capture VIP:port pairs for ports that report errors seems reasonable to me.  We could maybe do that as a followup PR if you prefer, but we should do it (or decide not to).

andrewsykim:
> If a cloud provider does not support mixed protocols the feature gate can be kept in a turned off state

I would be fine to merge this in v1.20 as long as we don't enable it by default in a future release until we know there are enough cloud providers handling mixed protocols -- either by allowing it or setting the right conditions. 

I'm not sure about the kube-proxy changes yet either.

janosi:
Unfortunately, I am not that familiar with kube-proxy either. I would need someones expertise on this.

So, if I understand @andrewsykim right we should keep this feature in alpha (beta?) as long as there is an agreement from the cloud providers that it can become GA? I do not see any problems with that, but maybe there are some disciplines that I do not know. I mean, for example, how fast a feature should become beta/GA.

thockin:
Regarding alpha:  I would want to stay Alpha until we know that SOMEONE is using it, and preferably more than just 1 someones.  That is a different discussion than going into our in-tree provider code and fixing all of those to set errors when mixed-protocol is encountered.  That seems like a relatively easy change to make. - look at `func (s *Controller) ensureLoadBalancer(service *v1.Service)` in `staging/src/k8s.io/cloud-provider/controllers/service/controller.go`.  Something like "if the impl does not set Ports[] itself, we set them to indicate mixed-protocol is not supported".  That is maybe a step too far?  It couples adding Ports[] status support to supporting mixed-protocol.  Are we willing to say that you MUST set ports[] in status if you support multi-protocol?

Otherwise just hitting the 5-ish implementations and just adding a snippet of code to each doesn't seem bad.

Kube-proxy isn't so scary.  Once https://github.com/kubernetes/kubernetes/pull/92312 lands, your change is pretty trivial

andrewsykim:
> Regarding alpha: I would want to stay Alpha until we know that SOMEONE is using it, and preferably more than just 1 someones. That is a different discussion than going into our in-tree provider code and fixing all of those to set errors when mixed-protocol is encountered. That seems like a relatively easy change to make.

The only caveat to this is that we are disallowing features in the in-tree providers starting v1.21, more on this discussion here https://groups.google.com/g/kubernetes-sig-cloud-provider/c/UkG46pNc6Cw. But regardless I think adoption from some external cloud providers should be a requirement anyways, myself and @cheftako can poke some people to make this happen in 1-2 releases. Similar to what Tim said, we can also add some hooks in service controller to the cloud provider (e.g. `ProtocolsSupported()`) that could make implementing this easier.  So overall, LGTM to merge alpha now and revisit beta requirements in the KEP. 

> Kube-proxy isn't so scary. Once #92312 lands, your change is pretty trivial

Having a hard time seeing how this PR and 92312 overlap. Can you expand on that a bit @thockin?

thockin:
92312 changes from returning a list of LBIPs to a list of status Ingress structs, so you can actually look to see if this port has an error.  That can probably stand to be refactored, but I wont hold you to that task  .

I'll approve and hold now, and let @janosi decide whether to wait and do kube-proxy in this PR or to do a follwup PR.

/approve
/lgtm
/hold

janosi:
Before proceeding, I would like to ask if you are OK with having the kube-proxy change only in 1.21, while this one would be in 1.20? The reason is, that I can contribute as a freetime activity only, and my current days are pretty busy with my regular tasks. So, I cannot promise that the kube-proxy side would be ready by the 1.20 code freeze. Thank you!

thockin:
That will possibly extend the alpha - we'll need a cycle of Alpha with the kube-proxy changes.  That seems OK to me.

PTAL: https://github.com/kubernetes/enhancements/pull/2135

lukasmrtvy:
Is there any chance for cherrypick to 1.17 1.18 1.19 ?

janosi:
@thockin I had to rebase. If it still stands, could you please give lgtm again? Thank you!

@lukasmrtvy I do not know the K8s project / backporting policies, I am sorry.

thockin (commented):
I must have missed this before, sorry:

When adding a new field (2 in this PR) that field MUST be feature-gated for one release, to make it safe to roll-back API versions.  To do that, you need to add these fields to `dropServiceDisabledFields` in pkg/registry/core/service/strategy.go (and the commensurate test).

janosi:
@thockin 

> When adding a new field (2 in this PR) that field MUST be feature-gated for one release, to make it safe to roll-back API versions. To do that, you need to add these fields to dropServiceDisabledFields in pkg/registry/core/service/strategy.go (and the commensurate test).

Implemented this.

thockin:
needs rebase, so sorry.

Maybe we should break kube_features into sig-specific files :)

thockin (commented):
Needs rebase because of a revert :(

/lgtm
/approve

kikisdeliveryservice:
Hi @janosi 

Please rebase and get final reapproval ASAP.

Best,
Kirsten
(ref: https://github.com/kubernetes/enhancements/issues/1435#issuecomment-726540685)

/milestone 1.20

janosi:
Hello @kikisdeliveryyservice I can do that this afternoon/evening Central
European time. Thank you!

Laszlo

janosi:
@thockin I think, I finished the rebase.

andrewsykim:
@janosi please squash the commits

pkg/apis/core/validation/conditional_validation.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// validateMixedProtocolLBService checks if the old Service has type=LoadBalancer and whether the Service has different Protocols
+// on its ports. If the MixedProtocolLBService feature flag is disabled the usage of different Protocols in the new Service is
+// valid only if the old Service has different Protocols, too.
+func validateMixedProtocolLBService(service, oldService *api.Service) (errs field.ErrorList) {
+	if service.Spec.Type != api.ServiceTypeLoadBalancer {
+		return
+	}
+	if !utilfeature.DefaultFeatureGate.Enabled(features.MixedProtocolLBService) {
```
andrewsykim, line 42: non-blocking comment:

This would read slightly better as:
```
if utilfeature.DefaultFeatureGate.Enabled(features.MixedProtocolLBService) {
   return
}

if serviceHasMixedProtocols(service) && !serviceHasMixedProtocols(oldService) {
     errs = append(errs, field.Invalid(field.NewPath("spec", "ports"), service.Spec.Ports, "may not contain more than 1 protocol when type is 'LoadBalancer'"))
}
```

pkg/features/kube_features.go
```diff
@@ -708,6 +708,11 @@ const (
 	// alpha: v1.20
 	// Adds support for kubelet to detect node shutdown and gracefully terminate pods prior to the node being shutdown.
 	GracefulNodeShutdown featuregate.Feature = "GracefulNodeShutdown"
+	// owner: @janosi
```
andrewsykim, line 711: non-blocking nit: there should be a newline above this line

pkg/apis/core/validation/validation_test.go
```diff
@@ -10288,12 +10288,12 @@ func TestValidateServiceCreate(t *testing.T) {
 			numErrs: 0,
 		},
 		{
-			name: "invalid load balancer with mix protocol",
+			name: "load balancer with mix protocol",
```
andrewsykim, line 10291: non-blocking comment, there should be a test here that accounts for the feature gate

pkg/apis/core/validation/validation.go
```diff
@@ -4249,22 +4249,6 @@ func ValidateService(service *core.Service) field.ErrorList {
 		allErrs = append(allErrs, field.NotSupported(specPath.Child("type"), service.Spec.Type, supportedServiceType.List()))
 	}
 
-	if service.Spec.Type == core.ServiceTypeLoadBalancer {
-		portsPath := specPath.Child("ports")
-		includeProtocols := sets.NewString()
-		for i := range service.Spec.Ports {
-			portPath := portsPath.Index(i)
-			if !supportedPortProtocols.Has(string(service.Spec.Ports[i].Protocol)) {
```
andrewsykim, line 4257: It looks like we dropped this check which ensures ports are using valid protocol names, I think we should keep it and only drop the validation below for using more than 1 protocol.

andrewsykim (commented):
Sorry for the late review, most of my comments are non-blocking but there's one comment about dropping validation for protocol names that I think we want to address, unless I missed some context on that discussion.

pkg/apis/core/validation/validation.go
janosi: @thockin had this comment: https://github.com/kubernetes/kubernetes/pull/94028#discussion_r471735567
So I removed it from here as part of that block.

andrewsykim: ack, thanks

pkg/apis/core/validation/validation_test.go
janosi: The feature gate is checked in a different function than `ValidateServiceCreate`. The feature gate is checked in `ValidateConditionalService` and for that we have the unit tests.

pkg/apis/core/validation/conditional_validation.go
janosi: Fixed

pkg/features/kube_features.go
janosi: Fixed

janosi:
@andrewsykim squashed and fixed.

thockin (commented):
Thanks!

/lgtm
/approve

thockin:
FWIW regarding squash, you can set the`tide/merge-method-squash` and let the bots handle it for you, if the goal is a complete squash.  Just make sure your first commit message is good :)

janosi:
> FWIW regarding squash, you can set the`tide/merge-method-squash` and let the bots handle it for you, if the goal is a complete squash. Just make sure your first commit message is good :)

Good to know. Thank you! :)

janosi:
@thockin  new rebase was required

janosi:
@andrewsykim @thockin @kikisdeliveryservice @liggitt Thank you all for your help!

zimmertr:
Thanks everyone for this!

Issue #64429, opened 2018-05-29
CSI attach/detach should use shared informer


jsafrane:
/kind bug
/sig storage

CSI volume unnecessarily abuses API server with Create/Watch/Delete VolumeAttachment calls, while A/D controller already has a shared informer for `VolumeAttachment` objects. CSI's `Attach()` and `Detach()` should use the shared informer from A/D controller:

- to check if `VolumeAttachment` exist before creating it at https://github.com/kubernetes/kubernetes/blob/4a7a18653cf1790fa946b8e87f68d86ce1612d8f/pkg/volume/csi/csi_attacher.go#L89

- to watch for `VolumeAttachment` changes at https://github.com/kubernetes/kubernetes/blob/4a7a18653cf1790fa946b8e87f68d86ce1612d8f/pkg/volume/csi/csi_attacher.go#L156

- to check if `VolumeAttach` is deleted or marked for deletion at https://github.com/kubernetes/kubernetes/blob/4a7a18653cf1790fa946b8e87f68d86ce1612d8f/pkg/volume/csi/csi_attacher.go#L380 and associated watch at https://github.com/kubernetes/kubernetes/blob/4a7a18653cf1790fa946b8e87f68d86ce1612d8f/pkg/volume/csi/csi_attacher.go#L423

Perhaps we could smuggle the shared informer into the plugin as a new parameter to `ProbeVolumePlugins()` at https://github.com/kubernetes/kubernetes/blob/4a7a18653cf1790fa946b8e87f68d86ce1612d8f/pkg/volume/csi/csi_plugin.go#L50 (but more investigation is needed).

misterikkit:
+1 to this. The current implementation creates a Watch per Attach operation, and that could result in dozens of concurrent watch requests during a spike.

chrishenzie:
Looking into this. The CSI plugin has access to a volume attachment lister which we can use to check for the existence of volume attachments:
https://github.com/kubernetes/kubernetes/blob/58bd3e5230df253cb23c22641883b87da92e5f43/pkg/volume/csi/csi_plugin.go#L197

chrishenzie:
I've been working on this for quite a while now and keep getting tangled in the process. Below are some notes and open questions.

- As an intermediate step, I tried replacing the API `get()` calls with `volumeAttachmentLister.get()`, however this is not possible because
  - We clear the `volumeAttachmentLister` if the volume host is a `KubeletVolumeHost`
[...]

PR #94031, opened 2020-08-16
test(csi): deflake TestAttacherWithCSIDriver

### LINKED ISSUE #93970: [sig-storage] TestAttacherWithCSIDriver flakes with timeout error

TestAttacherWithCSIDriver

**Reason for failure**:

```
--- FAIL: TestAttacherWithCSIDriver (120.07s)
    --- PASS: TestAttacherWithCSIDriver/CSIDriver_not_attachable (0.01s)
    --- PASS: TestAttacherWithCSIDriver/CSIDriver_is_attachable (0.02s)
    --- PASS: TestAttacherWithCSIDriver/CSIDriver.AttachRequired_not_set__->_failure (0.02s)
    --- FAIL: TestAttacherWithCSIDriver/CSIDriver_does_not_exist_not_set__->_failure (120.02s)
```

With that toleration removed in https://github.com/kubernetes/kubernetes/pull/93605, this flake has been seen (https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93605/pull-kubernetes-bazel-test/1293963409628336128).

```
go test -race -c ./pkg/volume/csi
```

2. Stress the affected test using the [stress](https://godoc.org/golang.org/x/tools/cmd/stress) tool (setting the timeout to a generous timeout relative to the normal test runtime of .02 seconds):

```
stress ./csi.test -test.run TestAttacherWithCSIDriver -test.timeout=10s
```

```sh
ls -lat $TMPDIR/go-stress*
```

/sig storage
cc @kubernetes/sig-storage-test-failures

knight42:

**What type of PR is this?**

/kind flake
/kind cleanup

I added the following debugging:
```diff
diff --git pkg/volume/csi/csi_attacher.go pkg/volume/csi/csi_attacher.go
index e429dc9a645..e2b62c6af47 100644
--- pkg/volume/csi/csi_attacher.go
+++ pkg/volume/csi/csi_attacher.go
@@ -461,6 +461,7 @@ func (c *csiAttacher) waitForVolumeAttachDetachStatus(attach *storage.VolumeAtta
        if err != nil {
                return fmt.Errorf("watch error:%v for volume %v", err, volumeHandle)
        }
+       klog.Infof("waitForVolumeAttachDetachStatus: watching")

        ch := watcher.ResultChan()
        defer watcher.Stop()
```

and noticed that:

```
=== RUN   TestAttacherWithCSIDriver/CSIDriver_is_attachable
    csi_attacher_test.go:96: attachment found on try 0, stopping wait...
    csi_attacher_test.go:100: stopped waiting for attachment
    csi_attacher_test.go:106: updating attachment csi-5988a16c07b08340b5bb360f76ea9b78d2a5f743091347e7d758253051dc6cad with attach status {true map[] nil nil}
I0816 12:56:56.233589   80627 csi_attacher.go:464] waitForVolumeAttachDetachStatus: watching <--- watcher was created after updating the status of attachment
panic: test timed out after 10s


=== RUN   TestAttacherWithCSIDriver/CSIDriver_does_not_exist_not_set__->_failure
    csi_attacher_test.go:96: attachment found on try 0, stopping wait...
    csi_attacher_test.go:100: stopped waiting for attachment
    csi_attacher_test.go:106: updating attachment csi-fc1b3cde1442bf6cfe265647fbbc1ae0ade3490a32e8bbc35e864ee0624cb4c1 with attach status {true map[] nil nil}
I0816 12:56:58.301347   80664 csi_attacher.go:464] waitForVolumeAttachDetachStatus: watching <--- watcher was created after updating the status of attachment
panic: test timed out after 10s
```

From the above logs, we could tell the root cause is the watcher in attacher https://github.com/kubernetes/kubernetes/blob/8f9f2fb2d64f81f1bd8aeb4056c4d01a21d2e910/pkg/volume/csi/csi_attacher.go#L460-L463 was created after we update the status of attachment in the test, so the watcher could not receive the update event and just stuck there.

FWIW, it is worth noting that `TestAttacherAttach` and `TestAttacherAttachWithInline` may also have the same problem, but we do not wait for the goroutine to exit in the test so the problem is masked.

Fixes #93970

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/volume/csi/csi_attacher_test.go
+++ b/pkg/volume/csi/csi_attacher_test.go
@@ -211,6 +211,7 @@ func TestAttacherAttach(t *testing.T) {
 
 			csiAttacher := attacher.(*csiAttacher)
 
+			// FIXME: We need to ensure this goroutine exits in the test.
 			go func(spec *volume.Spec, nodename string, fail bool) {
 				attachID, err := csiAttacher.Attach(spec, types.NodeName(nodename))
 				if !fail && err != nil {
@@ -294,6 +295,7 @@ func TestAttacherAttachWithInline(t *testing.T) {
 			}
 			csiAttacher := attacher.(*csiAttacher)
 
+			// FIXME: We need to ensure this goroutine exits in the test.
 			go func(spec *volume.Spec, nodename string, fail bool) {
 				attachID, err := csiAttacher.Attach(spec, types.NodeName(nodename))
 				if fail != (err != nil) {
@@ -356,6 +358,22 @@ func TestAttacherWithCSIDriver(t *testing.T) {
 			plug, _, tmpDir, _ := newTestWatchPlugin(t, fakeClient, true)
 			defer os.RemoveAll(tmpDir)
 
+			attachmentWatchCreated := make(chan core.Action)
+			// Make sure this is the first reactor
+			fakeClient.Fake.PrependWatchReactor("volumeattachments", func(action core.Action) (bool, watch.Interface, error) {
+				select {
+				case <-attachmentWatchCreated:
+					// already closed
+				default:
+					// The attacher is already watching the attachment, notify the test goroutine to
+					// update the status of attachment.
+					// TODO: In theory this still has a race condition, because the actual watch is created by
+					// the next reactor in the chain and we unblock the test goroutine before returning here.
+					close(attachmentWatchCreated)
+				}
+				return false, nil, nil
+			})
+
 			attacher, err := plug.NewAttacher()
 			if err != nil {
 				t.Fatalf("failed to create new attacher: %v", err)
@@ -377,8 +395,8 @@ func TestAttacherWithCSIDriver(t *testing.T) {
 			}
 			var wg sync.WaitGroup
 			wg.Add(1)
-			go func(volSpec *volume.Spec, expectAttach bool) {
-				attachID, err := csiAttacher.Attach(volSpec, types.NodeName("fakeNode"))
+			go func(volSpec *volume.Spec) {
+				attachID, err := csiAttacher.Attach(volSpec, "fakeNode")
 				defer wg.Done()
 
 				if err != nil {
@@ -387,13 +405,16 @@ func TestAttacherWithCSIDriver(t *testing.T) {
 				if attachID != "" {
 					t.Errorf("Expected empty attachID, got %q", attachID)
 				}
-			}(spec, test.expectVolumeAttachment)
+			}(spec)
 
 			if test.expectVolumeAttachment {
 				expectedAttachID := getAttachmentName("test-vol", test.driver, "fakeNode")
 				status := storage.VolumeAttachmentStatus{
 					Attached: true,
 				}
+				// We want to ensure the watcher, which is created in csiAttacher,
+				// has been started before updating the status of attachment.
+				<-attachmentWatchCreated
 				markVolumeAttached(t, csiAttacher.k8s, nil, expectedAttachID, status)
 			}
 			wg.Wait()
@@ -1568,6 +1589,13 @@ func newTestWatchPlugin(t *testing.T, fakeClient *fakeclient.Clientset, setupInf
 	}
 
 	factory.Start(wait.NeverStop)
+	ctx, cancel := context.WithTimeout(context.Background(), TestInformerSyncTimeout)
+	defer cancel()
+	for ty, ok := range factory.WaitForCacheSync(ctx.Done()) {
+		if !ok {
+			t.Fatalf("failed to sync: %#v", ty)
+		}
+	}
 
 	host := volumetest.NewFakeVolumeHostWithCSINodeName(t,
 		tmpDir,
@@ -1589,15 +1617,5 @@ func newTestWatchPlugin(t *testing.T, fakeClient *fakeclient.Clientset, setupInf
 		t.Fatalf("cannot assert plugin to be type csiPlugin")
 	}
 
-	// Wait until the informer in CSI volume plugin has all CSIDrivers.
-	wait.PollImmediate(TestInformerSyncPeriod, TestInformerSyncTimeout, func() (bool, error) {
-		return csiDriverInformer.Informer().HasSynced(), nil
-	})
-
-	if volumeAttachmentInformer != nil {
-		wait.PollImmediate(TestInformerSyncPeriod, TestInformerSyncTimeout, func() (bool, error) {
-			return volumeAttachmentInformer.Informer().HasSynced(), nil
-		})
-	}
 	return csiPlug, fakeWatcher, tmpDir, fakeClient
 }


msau42:
/assign @misterikkit 
cc @Jiawei0227 who has been making some changes in this area

pkg/volume/csi/csi_attacher_test.go
```diff
@@ -1556,6 +1562,21 @@ func newTestWatchPlugin(t *testing.T, fakeClient *fakeclient.Clientset, setupInf
...
+	// Use buffered channel here to prevent blocking the reactor
+	attachmentWatchActions := make(chan core.Action, 5)
+	fakeClient.Fake.PrependWatchReactor("volumeattachments", func(action core.Action) (bool, watch.Interface, error) {
+		if action.GetVerb() != "watch" {
+			return false, nil, nil
+		}
+		select {
+		case attachmentWatchActions <- action:
+		default:
+			// TODO(knight42): Is it okay to panic here?
```
Jiawei0227, line 1574: I am wondering if it would make sense if we do 
```
if action.GetVerb() != "watch" || len(attachmentWatchActions) >=5 {
	return false, nil, nil
}
```
So that dont need to worry about the channel get full.
Or is there any other solution that we can do to locate this specific volumeattachment?

pkg/volume/csi/csi_attacher_test.go
```diff
@@ -387,13 +389,17 @@ func TestAttacherWithCSIDriver(t *testing.T) {
...
+			}(spec)
 
 			if test.expectVolumeAttachment {
 				expectedAttachID := getAttachmentName("test-vol", test.driver, "fakeNode")
 				status := storage.VolumeAttachmentStatus{
 					Attached: true,
 				}
+				// We want to ensure the second watcher, which is the watcher created in csiAttacher,
+				// has been started before updating the status of attachment.
+				<-attachmentWatchActions
```
Jiawei0227, line 401: Can you also add a comment on what is the first watcher?

Jiawei0227 (commented):
Thanks for addressing this! I am relatively new to this area so not sure if there is a better solution to locate the second watcher. Otherwise this looks good. Maybe @misterikkit can give some more thoughts.

pkg/volume/csi/csi_attacher_test.go
```diff
@@ -1589,15 +1617,5 @@ func newTestWatchPlugin(t *testing.T, fakeClient *fakeclient.Clientset, setupInf
 		t.Fatalf("cannot assert plugin to be type csiPlugin")
 	}
 
-	// Wait until the informer in CSI volume plugin has all CSIDrivers.
```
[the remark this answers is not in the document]
knight42: They are equivalent.

https://github.com/kubernetes/client-go/blob/e7a1d9ada0d5ef8eee6e55f0a84ad6e3d08441f6/informers/factory.go#L155-L159

`factory.WaitForCacheSync` would invoke `cache.WaitForCacheSync` on each informer saved in `sharedInformerFactory`.

pkg/volume/csi/csi_attacher_test.go
```diff
@@ -1556,6 +1562,21 @@ func newTestWatchPlugin(t *testing.T, fakeClient *fakeclient.Clientset, setupInf
 		fakeClient.Fake.PrependWatchReactor("volumeattachments", core.DefaultWatchReactor(fakeWatcher, nil))
 	}
 
+	// Use buffered channel here to prevent blocking the reactor
```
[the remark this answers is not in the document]
knight42: Thanks for pointing it out, I think this is better.

pkg/volume/csi/csi_attacher_test.go
```diff
@@ -211,6 +211,7 @@ func TestAttacherAttach(t *testing.T) {
 
 			csiAttacher := attacher.(*csiAttacher)
 
+			// FIXME: We need to ensure this goroutine exits in the test.
```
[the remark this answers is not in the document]
knight42: Some test cases which should fail actually expects timeout, but the default timeout is 2 min https://github.com/kubernetes/kubernetes/blob/3f579d8971fcce96d6b01b968a46c720f10940b8/pkg/volume/csi/csi_plugin.go#L51, maybe we should figure out how to adjust the time `csiAttacher.Attach` should wait for before fixing it.

pkg/volume/csi/csi_attacher_test.go
knight42: Now we move the `fakeClient.Fake.PrependWatchReactor` out of `newTestWatchPlugin`, there is only one watcher now.

knight42: The first watcher is the one created in `newTestWatchPlugin`.

pkg/volume/csi/csi_attacher_test.go
knight42: Now that we move the `fakeClient.Fake.PrependWatchReactor` out of `newTestWatchPlugin`, it is likely this line is unreachable.

knight42:
pull-kubernetes-e2e-gce-csi-serial, pull-kubernetes-e2e-gce-storage-slow, pull-kubernetes-e2e-gce-storage-snapshot, pull-kubernetes-e2e-gce-ubuntu-containerd all fail because:
```
W0819 03:49:50.568] 2020/08/19 03:49:50 main.go:727: --gcp-project is missing, trying to fetch a project from boskos.
W0819 03:49:50.568] (for local runs please set --gcp-project to your dev project)
W0819 03:49:50.568] 2020/08/19 03:49:50 main.go:739: provider gce, will acquire project type gce-project from boskos
W0819 03:54:50.568] 2020/08/19 03:54:50 process.go:96: Saved XML output to /workspace/_artifacts/junit_runner.xml.
W0819 03:54:50.568] 2020/08/19 03:54:50 process.go:153: Running: bash -c . hack/lib/version.sh && KUBE_ROOT=. kube::version::get_version_vars && echo "${KUBE_GIT_VERSION-}"
W0819 03:54:50.865] 2020/08/19 03:54:50 process.go:155: Step 'bash -c . hack/lib/version.sh && KUBE_ROOT=. kube::version::get_version_vars && echo "${KUBE_GIT_VERSION-}"' finished in 297.201637ms
W0819 03:54:50.866] 2020/08/19 03:54:50 main.go:312: Something went wrong: failed to prepare test environment: --provider=gce boskos failed to acquire project: resources not found
W0819 03:54:50.874] Traceback (most recent call last):
W0819 03:54:50.874]   File "/workspace/./test-infra/jenkins/../scenarios/kubernetes_e2e.py", line 720, in <module>
W0819 03:54:50.874]     main(parse_args())
W0819 03:54:50.874]   File "/workspace/./test-infra/jenkins/../scenarios/kubernetes_e2e.py", line 570, in main
W0819 03:54:50.874]     mode.start(runner_args)
W0819 03:54:50.875]   File "/workspace/./test-infra/jenkins/../scenarios/kubernetes_e2e.py", line 228, in start
W0819 03:54:50.875]     check_env(env, self.command, *args)
W0819 03:54:50.875]   File "/workspace/./test-infra/jenkins/../scenarios/kubernetes_e2e.py", line 111, in check_env
W0819 03:54:50.875]     subprocess.check_call(cmd, env=env)
W0819 03:54:50.875]   File "/usr/lib/python2.7/subprocess.py", line 190, in check_call
W0819 03:54:50.875]     raise CalledProcessError(retcode, cmd)
```

pkg/volume/csi/csi_attacher_test.go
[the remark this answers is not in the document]
knight42: For instance, `TestAttacherAttach/mismatch_vol` and `TestAttacherAttach/mismatch_node` would finish after 2m:

```
=== RUN   TestAttacherAttach/mismatch_node
    csi_attacher_test.go:203: test case: mismatch node
E0819 14:40:20.168036   58170 csi_plugin.go:199] kubernetes.io/csi: VolumeAttachmentLister not found on AttachDetachVolumeHost
    csi_attacher_test.go:100: stopped waiting for attachment
    csi_attacher_test.go:103: attachment not found for id:csi-70fa8b122628ff5436f098759b79c93487b3894e2bff5637c618405a059b0b47
E0819 14:42:20.185348   58170 csi_attacher.go:500] kubernetes.io/csi: attachdetacher.WaitForDetach timeout after 2m0s [volume=vol02; attachment.ID=csi-cad5d366e9d018c22553557296636318d4a3d2026611b6ebaa5ab7789ebc7
dbb]
```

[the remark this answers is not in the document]
knight42: It would be great if the `Attach` method could accept `context.Context` as its argument, but I think that needs to be discussed first.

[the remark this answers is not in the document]
Jiawei0227: Yeah, that's a good point. But it seems the current test suite does not even function correctly. As it does not wait for the go routine to finish. If you are not intend to fix this in this PR, maybe worth creating a new issue for it?

pkg/volume/csi/csi_attacher_test.go
```diff
@@ -356,6 +358,21 @@ func TestAttacherWithCSIDriver(t *testing.T) {
 			plug, _, tmpDir, _ := newTestWatchPlugin(t, fakeClient, true)
 			defer os.RemoveAll(tmpDir)
 
+			// Use buffered channel here to prevent blocking the reactor
+			attachmentWatchActions := make(chan core.Action, 5)
```
Jiawei0227, line 362: nit:  dont know if it worth it to use buffered channel since now we just have one attachmentvolume watcher after prepending the reactor.

Jiawei0227 (commented):
/lgtm

pkg/volume/csi/csi_attacher_test.go
knight42: IMHO I don't want to block the following reactors in any case and I think buffered channel does not cost much more than unbuffered ones, I think this is okay.

misterikkit:
I love deflaking tests!

@knight42 Is this a flake that causes the test to fail sometimes, or are you saying that it will not cover the intended functionality depending on the race condition?

I did not get any test failures in 1500 runs of `go test -race -v ./pkg/volume/csi` 

My head is currently pointing at d20d894d72d7289b6e6f02d2fa0b947c25bab7b0

liggitt:
> I did not get any test failures in 1500 runs of `go test -race -v ./pkg/volume/csi`

check with the reproducer steps in https://github.com/kubernetes/kubernetes/issues/93970#issue-678645093 (uses -race + the `stress` tool to run lots of tests in parallel, which does a better job of exercising CPU starvation/timing issues)

misterikkit:
> uses -race + the `stress` tool

Thanks Jordan, I was already using those, but I think I was holding it wrong.

`stress go test -race ./foo` will returned cached test results unless you add `-count 1`. But precompiled test binary does not cache results.

I was not using a timeout of 10s.

knight42:
@misterikkit It is recommended to limit the timeout to 10s, or you would have to wait for at least 2min https://github.com/kubernetes/kubernetes/blob/3f579d8971fcce96d6b01b968a46c720f10940b8/pkg/volume/csi/csi_plugin.go#L51 to see the failed test.

misterikkit:
Now that I've read the csi attacher code, I think the actual bug is in csi_attacher.go and not csi_attacher_test.go. Specifically, we do a Get followed by a Watch with the goal of "wait for desired state". If the state changes between Get and Watch, then we are not guaranteed to see any events from the watcher.

This is the race @knight42 found in test. I would guess it is **extremely** difficult to hit this in prod, but I would prefer the attacher fixed instead of the test.

It may be best to go with https://github.com/kubernetes/kubernetes/issues/64429 to actually fix this attacher, in which case I would be fine merging this CL to fix tests.

liggitt:
> Specifically, we do a Get followed by a Watch with the goal of "wait for desired state". If the state changes between Get and Watch, then we are not guaranteed to see any events from the watcher.
>
> This is the race @knight42 found in test. I would guess it is extremely difficult to hit this in prod, but I would prefer the attacher fixed instead of the test.

+1 to fixing the bug instead of changing the test to mask the issue

misterikkit:
One option for a short term fix would be to create the watch and then do a Get. That requires a bit of re-plumbing since currently the Get is called before we enter the func that creates the watch.

This creates a different race where we start a watch, events can start filling the watch channel, then we do a get, then we start looping on the watch channel. This means we might see states out of chronological order, but I have a hunch it will be okay.

liggitt:
do you have a pointer to the problematic line in the attacher you tracked down?

misterikkit:
The call stack is

- Attach()
  - waitForVolumeAttachment()
    - waitForVolumeAttachmentInternal()
      - waitForVolumeAttachDetachStatus()

Get is in `waitForVolumeAttachmentInternal`:
https://github.com/kubernetes/kubernetes/blob/1c548c328a32c5b6819195760d1abba115ab4969/pkg/volume/csi/csi_attacher.go#L157

Watch is in `waitForVolumeAttachDetachStatus`:
https://github.com/kubernetes/kubernetes/blob/1c548c328a32c5b6819195760d1abba115ab4969/pkg/volume/csi/csi_attacher.go#L460

chrishenzie:
Hi @knight42, myself and @misterikkit were going to planning to look into the shared informer issue https://github.com/kubernetes/kubernetes/issues/64429 which may fix this test flake. I wanted to check if you were already looking at this, or if we should continue.

knight42:
> One option for a short term fix would be to create the watch and then do a Get. That requires a bit of re-plumbing since currently the Get is called before we enter the func that creates the watch.

@misterikkit I am not sure if this would actually fix the flake. Now we call `csiAttacher.Attach` in a separate goroutine, and update the status of attachment in another. It is not guaranteed that attacher is already watching the attachment before we update it in the test.

> Hi @knight42, myself and @misterikkit were going to planning to look into the shared informer issue #64429 which may fix this test flake. I wanted to check if you were already looking at this, or if we should continue.

It would be great if you could look into the issue, please let me know if there is anything I could do, I am willing to help.

In addition, while you are fixing the problem, perhaps you could check if there is any chance to update the signature of `volume.Attacher` to
```go
	Attach(ctx context.Context, spec *Spec, nodeName types.NodeName) (string, error)
```
so that we could specify a shorter timeout in the test, which is helpful for the problem I pointed out here https://github.com/kubernetes/kubernetes/pull/94031#discussion_r472767229.

liggitt:
> @liggitt How would you prioritize de-flaking this test vs refactoring CSI Attacher?

short-term, deflaking the test is a priority so we can resolve https://github.com/kubernetes/kubernetes/pull/93605 and protect the rest of the code base from newly introduced flakes

* if the only way to deflake the test without completely rewriting it means changing the test so it is not actually testing what we care about, then we should treat the relevant controller code as untested and have a high priority issue tracking rewriting the test to actually exercise the controller code
* it also sounds like we've identified at least two bugs in the controller that should be fixed with high priority:
  * starting a get from an individual item's resourceVersion (should start from a list's resourceVersion)
  * not tolerating/retrying a dropped watch

liggitt:
> ... vs refactoring CSI Attacher? The latter will take longer.

looking at the controller, I think the waitForVolumeAttachmentInternal/waitForVolumeDetachmentInternal/waitForVolumeAttachDetachStatus bits could be minimally tweaked to do the right thing:

1. in waitForVolumeAttachmentInternal/waitForVolumeDetachmentInternal:
  * do a list with a metadata.name field selector
  * switch NotFound error checks for len(attachments.Items) == 0 checks
  * pass in the list's resourceVersion to waitForVolumeAttachDetachStatus
2. in waitForVolumeAttachDetachStatus, use `watchtools.NewRetryWatcher` with a metadata.name field selector instead of doing a raw watch

The biggest change will be fixing unit tests that do not respond correctly to list requests

knight42:
@liggitt looks like this patch could be applied to this PR?

liggitt:
> @liggitt looks like this patch could be applied to this PR?

Doing so requires rewriting the unit tests to correctly respond to list requests with a resourceVersion. That's a larger change that would require careful review to make sure we didn't regress something, so I'd like to separate that from a fix that stabilizes the unit test flake

misterikkit:
Thanks for the detailed analysis. @liggitt could you explain why list is preferred over get here? Is it because an empty list still returns a resourceVersion?

liggitt:
> could you explain why list is preferred over get here? Is it because an empty list still returns a resourceVersion?

the resourceVersion of the list is the version of etcd storage and represents a valid version to start a watch from.

the resourceVersion of the individual item represents the last time that item was modified, which could have been a long time ago, and etcd could have been compacted since then, and it might not even be valid to start a watch from

misterikkit:
Oh, thank you! I re-read [this](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes) and now I see that we shouldn't have the race condition I described if we correctly handle `resourceVersion` correctly (as per your `List` suggestion)

liggitt:
/priority important-soon

Note that this is blocking #93605 and full reopen of 1.20, so resolving the test flake in the next few days would be helpful.

knight42:
hmmm I am a bit confused. Is the woraround provided in this PR acceptable? If not, perhaps we should file an separate PR to fix the bug?

liggitt:
> hmmm I am a bit confused. Is the woraround provided in this PR acceptable? If not, perhaps we should file an separate PR to fix the bug?

I'm not familiar enough with the attacher flow to know whether this is a reasonable short-term change, so I'll defer to the storage reviewers

Jiawei0227:
From my perspective, I think this fix makes sense because it simulate the real-world situation. What this fix does is waiting for the watcher to start up and then perform the markVolumeAttached() call. In reallife, attaching through VA is not likely to perform faster than the watcher to be established. 

Even if it does happen during this gap, worst case is attach timeout. And the next attach will succeed. So in terms of priority, I think fixing this flaky test sounds good to me. @misterikkit what's your thoughts?

pkg/volume/csi/csi_attacher_test.go
```diff
@@ -356,6 +358,21 @@ func TestAttacherWithCSIDriver(t *testing.T) {
...
+			fakeClient.Fake.PrependWatchReactor("volumeattachments", func(action core.Action) (bool, watch.Interface, error) {
+				if action.GetVerb() != "watch" {
+					return false, nil, nil
+				}
+				select {
+				case attachmentWatchActions <- action:
+				default:
+					panic("Unreachable: channel full")
+				}
+				return false, nil, nil
```
misterikkit, line 373: In *theory* this still has a race condition, because you've unblocked line 416 before returning here, and the actual watch is created by the next reactor in the chain.

However, I ran these tests 2000 times with your change and got no failures, so I think it is good enough.

pkg/volume/csi/csi_attacher_test.go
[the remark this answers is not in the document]
misterikkit: I believe we have not observed any flakes on these tests. Possibly they are failing to fail. At least it's not blocking CI tests...

misterikkit (commented):
Thanks for your patience and effort on this bug!

I think we should merge this PR, I just have some minor feedback about closing the channel.

Please also squash commits.

pkg/volume/csi/csi_attacher_test.go
```diff
@@ -356,6 +358,21 @@ func TestAttacherWithCSIDriver(t *testing.T) {
 			plug, _, tmpDir, _ := newTestWatchPlugin(t, fakeClient, true)
 			defer os.RemoveAll(tmpDir)
 
+			// Use buffered channel here to prevent blocking the reactor
+			attachmentWatchActions := make(chan core.Action, 5)
+			// Make sure this is the first reactor
+			fakeClient.Fake.PrependWatchReactor("volumeattachments", func(action core.Action) (bool, watch.Interface, error) {
+				if action.GetVerb() != "watch" {
```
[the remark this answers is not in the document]
knight42: yeah this check is redundant in this case

pkg/volume/csi/csi_attacher_test.go
knight42: > In theory this still has a race condition, because you've unblocked line 416 before returning here, and the actual watch is created by the next reactor in the chain.

I think you are right, I added a comment in the test to note it is still possible to race.

misterikkit (approved):
/lgtm

Thanks again! :smile:

liggitt:
/approve cancel

would like an approver from https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/OWNERS to ack

liggitt:
/assign jsafrane saad-ali

this is a test-only fix for a unit test flake to unblock https://github.com/kubernetes/kubernetes/pull/93605, PTAL

Issue #93988, opened 2020-08-14
The CPU.Time is always set to Container StartTime in pod metrics


AbelHu:
**What happened**:
https://github.com/kubernetes/kubernetes/blob/55412b6af1764f8bc3d9dfa8f3bd29f1a65f4590/pkg/kubelet/stats/cri_stats_provider.go#L467-L479
CPU.Time is set to cs.StartTime but ps.Memory.Time is set to timestamp when collecting the memoryusage.

**What you expected to happen**:
CPU.Time is set to cs.CPU.Time

**How to reproduce it (as minimally and precisely as possible)**:
Check the code

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
master branch when creating this issue.
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

qiutongs:
I am happy to take over this. 

/assign

PR #94034, opened 2020-08-16
change cs.StartTime to cs.CPU.Time in addPodCPUMemoryStats

Zheaoli:
**What type of PR is this?**
/kind bug

Fixes # https://github.com/kubernetes/kubernetes/issues/93988

```release-note

```

```docs

```

### DIFF
--- a/pkg/kubelet/stats/cri_stats_provider.go
+++ b/pkg/kubelet/stats/cri_stats_provider.go
@@ -464,7 +464,7 @@ func (p *criStatsProvider) addPodCPUMemoryStats(
 			ps.CPU = &statsapi.CPUStats{}
 		}
 
-		ps.CPU.Time = cs.StartTime
+		ps.CPU.Time = cs.CPU.Time
 		usageCoreNanoSeconds := getUint64Value(cs.CPU.UsageCoreNanoSeconds) + getUint64Value(ps.CPU.UsageCoreNanoSeconds)
 		usageNanoCores := getUint64Value(cs.CPU.UsageNanoCores) + getUint64Value(ps.CPU.UsageNanoCores)
 		ps.CPU.UsageCoreNanoSeconds = &usageCoreNanoSeconds
--- a/pkg/kubelet/stats/cri_stats_provider_test.go
+++ b/pkg/kubelet/stats/cri_stats_provider_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package stats
 
 import (
+	"fmt"
 	"math/rand"
 	"os"
 	"path/filepath"
@@ -28,13 +29,13 @@ import (
 	cadvisorfs "github.com/google/cadvisor/fs"
 	cadvisorapiv2 "github.com/google/cadvisor/info/v2"
 	"github.com/stretchr/testify/assert"
-	assertModule "github.com/stretchr/testify/assert"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 	critest "k8s.io/cri-api/pkg/apis/testing"
+	"k8s.io/klog/v2"
 	statsapi "k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
 	cadvisortest "k8s.io/kubernetes/pkg/kubelet/cadvisor/testing"
 	"k8s.io/kubernetes/pkg/kubelet/cm"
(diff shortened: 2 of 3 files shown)


dashpole:
/ok-to-test

I think you will have to update the unit test for this

/approve

Zheaoli:
@dashpole thanks to remind me for that. I'm working on it

Zheaoli:
It seems all test pass, don't need extra unit tests. PTAL  @dashpole

dashpole:
I think that means it actually just isn't unit tested at all...

Looking at https://github.com/kubernetes/kubernetes/blob/fd74333a971e2048b5fb2b692a9e043483d63fba/pkg/kubelet/stats/stats_provider_test.go#L503, we don't even set timestamps on the CPU metrics returned by the fake.  Can you update the tests so that we are actually testing this?

Zheaoli:
@dashpole Sure, I'll update the test ASAP

pkg/kubelet/stats/cri_stats_provider.go
```diff
@@ -464,7 +464,7 @@ func (p *criStatsProvider) addPodCPUMemoryStats(
 			ps.CPU = &statsapi.CPUStats{}
 		}
 
-		ps.CPU.Time = cs.StartTime
+		ps.CPU.Time = cs.CPU.Time
```
krmayankk, line 467: @Zheaoli  does this require a doc bug fix ? Where do these stats show up  for users ?

Zheaoli: I'm not sure about the doc. I will check it again

dims:
Need to remove WIP from title

/retest

Zheaoli:
@dims I work on fix the unit test. but maybe I need some help

SergeyKanzhelev:
@Zheaoli do you still plan to update unit tests to validate this field?

Zheaoli:
@SergeyKanzhelev If it's OK I will use another PR to update the related unit test.

pkg/kubelet/stats/cri_stats_provider.go
dashpole: I don't think we document this that specifically.

dashpole: This is for creating the Summary API.

dashpole:
@Zheaoli we will likely want to cherrypick this to previous releases, and it would be easier to do so with the unit test included.  If you can push what you have that is failing, we may be able to help debug failures.

pkg/kubelet/stats/cri_stats_provider_test.go
```diff
@@ -421,87 +421,89 @@ func TestCRIListPodCPUAndMemoryStats(t *testing.T) {
 	)
 
 	stats, err := provider.ListPodCPUAndMemoryStats()
-	assert := assert.New(t)
-	assert.NoError(err)
-	assert.Equal(4, len(stats))
+	assertObject := assert.New(t)
```
SergeyKanzhelev, line 424: curious, why to rename the variable?

SergeyKanzhelev: this definitely makes review harder. Like this line wasn't changed, but still highlighted:

image

SergeyKanzhelev:
> /retest

those seemed to be a valid issues. For package name:

from `verify-pkg-names.sh`

>   echo "!!! Some package aliases break go conventions."
>   echo "To fix these errors, do not use capitalized or underlined characters"
>   echo "in pkg aliases. Refer to https://blog.golang.org/package-names for more info."

As for test failure, as @dashpole mentioned perhaps the timestamp needs to be returned by fake.

Also what I cannot understand, there is already check for the pod's stats timestamp here:

https://github.com/kubernetes/kubernetes/blob/ac97b2d65e6aad6c472b6699b1474a0c0fd276b7/pkg/kubelet/stats/cri_stats_provider_test.go#L746

I wonder if it needs to be updated once fake would start behaving.

Zheaoli:
@SergeyKanzhelev @dashpole Actually I think we don't need mock on the https://github.com/kubernetes/kubernetes/blob/fd74333a971e2048b5fb2b692a9e043483d63fba/pkg/kubelet/stats/stats_provider_test.go#L503

the issue code is here 

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/stats/cri_stats_provider.go#L467

I want to change it to `cs.CPU.Time` and I think we have already had the mock data, take a look at this https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/stats/cri_stats_provider_test.go#L619

But the assertion in my new test will fail on the online test, I think it's weird

pkg/kubelet/stats/cri_stats_provider_test.go
```diff
@@ -28,6 +28,7 @@ import (
 	cadvisorfs "github.com/google/cadvisor/fs"
 	cadvisorapiv2 "github.com/google/cadvisor/info/v2"
 	"github.com/stretchr/testify/assert"
+	assertModule "github.com/stretchr/testify/assert"
```
dashpole, line 31: looks like we already import this.  Also, `assertModule` is camelcased, which imports shouldn't be.

pkg/kubelet/stats/cri_stats_provider_test.go
```diff
@@ -457,6 +458,8 @@ func TestCRIListPodCPUAndMemoryStats(t *testing.T) {
 	assert.Nil(c1.Logs)
 	assert.Nil(c1.Accelerators)
 	assert.Nil(c1.UserDefinedMetrics)
+	assert.True(assertModule.ObjectsAreEqual(p0.CPU.Time, metav1.NewTime(time.Unix(0, containerStats1.Cpu.Timestamp))) || assertModule.ObjectsAreEqual(p0.CPU.Time, metav1.NewTime(time.Unix(0, containerStats0.Cpu.Timestamp))))
```
dashpole, line 461: While we are debugging, can you make the error messages more detailed?  Perhaps stick to the form:
`Got %v, want %v`, and print out p0.CPU.Time, cs.CPU.Time and cs.StartTime if it isn't included in the Got, want print statement.

pkg/kubelet/stats/cri_stats_provider_test.go
SergeyKanzhelev: +1 on name. The reason it is renamed is that we declare variable with the same name `assert` and cannot use `assert.ObjectsAreEqual` method after this - variable overrides the package name

Zheaoli:
I will close this PR and reopen it when I solve this problems

SergeyKanzhelev:
I'm sorry with your experience implementing it. It feels like an important issue to fix. Are you actively working on it? If not - please let me know so I can track resolution for it.

PR #94036, opened 2020-08-16
Use /usr/bin/env bash in Makefiles and scripts

saschagrunert:
**What type of PR is this?**

/kind cleanup

This allows the execution of the Makefiles from distributions which do
have `bash` in a different path.

None

None

```release-note
None
```

### DIFF
--- a/build/root/Makefile
+++ b/build/root/Makefile
@@ -30,7 +30,7 @@ endif
 #   clean: Clean up.
 
 # It's necessary to set this because some environments don't link sh -> bash.
-SHELL := /bin/bash
+SHELL := /usr/bin/env bash
 
 # We don't need make's built-in rules.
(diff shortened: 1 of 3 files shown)


build/root/Makefile
```diff
@@ -30,7 +30,7 @@ endif
 #   clean: Clean up.
 
 # It's necessary to set this because some environments don't link sh -> bash.
-SHELL := /bin/bash
+SHELL := /usr/bin/env bash
```
spiffxp, line 33: I'd be curious to know what lines in the Makefile(s) need this, my understanding is that [make doesn't export this](https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html)

spiffxp (approved):
/approve
/lgtm

liggitt:

PR #94054, opened 2020-08-17
start kube-apiserver and webhook server only once to shorten the webhook audit test time

vareti:
**What type of PR is this?**
/kind cleanup

This PR reduces the webhook audit integration test run time from ~45s to ~20s. It starts kube-apiserver and webhook server only once instead of starting them for every audit level that is tested. 

It fixes the first part of the umbrella issue https://github.com/kubernetes/kubernetes/issues/93609
Fixes #

```
NONE
```

/assign @liggitt

### DIFF
--- a/test/integration/master/audit_test.go
+++ b/test/integration/master/audit_test.go
@@ -58,18 +58,34 @@ var (
 apiVersion: {version}
 kind: Policy
 rules:
-  - level: {level}
+  - level: RequestResponse
+    namespaces: ["no-webhook-namespace"]
+    resources:
+      - group: "" # core
+        resources: ["configmaps"]
+  - level: Metadata
+    namespaces: ["webhook-audit-metadata"]
+    resources:
+      - group: "" # core
+        resources: ["configmaps"]
+  - level: Request
+    namespaces: ["webhook-audit-request"]
+    resources:
+      - group: "" # core
+        resources: ["configmaps"]
+  - level: RequestResponse
+    namespaces: ["webhook-audit-response"]
     resources:
       - group: "" # core
         resources: ["configmaps"]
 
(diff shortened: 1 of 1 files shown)


vareti:
test failures are unrelated to change. trying a retest

test/integration/master/audit_test.go
```diff
@@ -258,24 +243,60 @@ func testAudit(t *testing.T, version string, level auditinternal.Level, enableMu
...
+		t.Run(fmt.Sprintf("%s.%s.%t", version, tc.auditLevel, tc.enableMutatingWebhook), func(t *testing.T) {
+			testAudit(t, version, tc.auditLevel, tc.enableMutatingWebhook, tc.namespace, kubeclient, logFile)
+		})
+	}
+}
+
+func testAudit(t *testing.T, version string, level auditinternal.Level, enableMutatingWebhook bool, namespace string, kubeclient kubernetes.Interface, logFile *os.File) {
 	var lastMissingReport string
+	createNamespace(t, kubeclient, namespace)
+	defer deleteNamespace(t, kubeclient, namespace)
```
liggitt, line 287: I don't think you need to delete the namespace if we're only testing each one one time. Issuing a delete for a namespace will only mark it terminating, it will not actually remove the namespace if the namespace cleanup controller is not running.

liggitt:
one comment on the namespace cleanup, this looks reasonable otherwise

test/integration/master/audit_test.go
vareti: That is good to know. :+1:

vareti:
@liggitt thanks for the review. I addressed the review comment. Please take a look.

PR #94062, opened 2020-08-17
Updated kubectl e2e test to handle version skew with >= 1.19 server version

brianpursley:
*** ATTENTION ***
This PR is to merge against the `release-1.18` branch, so that it can fix a skew test failure

**What type of PR is this?**
/kind failing-test

One of the periodic e2e skew tests is failing running a 1.18 client against a 1.20 server.
```
[sig-cli] Kubectl client Simple pod should return command exit codes
```
https://testgrid.k8s.io/sig-cli-master#skew-cluster-stable1-kubectl-latest-gce

In 1.19, the e2e tests were changed from expecting no error to expecting a timeout.
https://github.com/kubernetes/kubernetes/blob/1afc53514032a44d091ae4a9f6e092171db9fe10/test/e2e/kubectl/kubectl.go#L534-L553

This PR adds a check to see if it is running against a pre-1.19 server and if so, performing the old check, and it a 1.19 server or later, performing the new check

This PR is to merge against the `release-1.18` branch, so that it can fix a skew test failure

```release-note
NONE
```

```docs

```

/cc @soltysh

### DIFF
--- a/test/e2e/kubectl/kubectl.go
+++ b/test/e2e/kubectl/kubectl.go
@@ -486,17 +486,37 @@ var _ = SIGDescribe("Kubectl client", func() {
 			framework.ExpectEqual(ok, true)
 			framework.ExpectEqual(ee.ExitStatus(), 42)
 
+			serverVersion, _ := f.ClientSet.Discovery().ServerVersion()
+			serverMajorVersion, _ := strconv.Atoi(serverVersion.Major)
+			serverMinorVersion, _ := strconv.Atoi(regexp.MustCompile(`^\d+`).FindString(serverVersion.Minor))
+
 			ginkgo.By("running a failing command without --restart=Never")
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
/lgtm

@soltysh can you take a look here?

test/e2e/kubectl/kubectl.go
```diff
@@ -486,17 +486,37 @@ var _ = SIGDescribe("Kubectl client", func() {
 			framework.ExpectEqual(ok, true)
 			framework.ExpectEqual(ee.ExitStatus(), 42)
 
+			serverVersion, _ := f.ClientSet.Discovery().ServerVersion()
+			serverMajorVersion, _ := strconv.Atoi(serverVersion.Major)
```
soltysh, line 490: How about using something like this: https://github.com/kubernetes/kubernetes/blob/release-1.18/test/e2e/framework/skipper/skipper.go#L246?

soltysh (changes requested):
One nit and you're good to go.

PR #94066, opened 2020-08-18
add audit-log-compress to apiserver

lojies:
**What type of PR is this?**
/kind feature

Add --audit-log-compress to apiserver.This can determine if the rotated log files should be compressed using gzip.

Fixes #

```release-note
kube-apiserver: added support for compressing rotated audit log files with `--audit-log-compress`
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/server/options/audit.go
+++ b/staging/src/k8s.io/apiserver/pkg/server/options/audit.go
@@ -120,6 +120,7 @@ type AuditLogOptions struct {
 	MaxBackups int
 	MaxSize    int
 	Format     string
+	Compress   bool
 
 	BatchOptions    AuditBatchOptions
 	TruncateOptions AuditTruncateOptions
@@ -449,6 +450,7 @@ func (o *AuditLogOptions) AddFlags(fs *pflag.FlagSet) {
 			strings.Join(pluginlog.AllowedFormats, ",")+".")
 	fs.StringVar(&o.GroupVersionString, "audit-log-version", o.GroupVersionString,
 		"API group and version used for serializing audit events written to log.")
+	fs.BoolVar(&o.Compress, "audit-log-compress", o.Compress, "If set, the rotated log files will be compressed using gzip.")
 }
 
 func (o *AuditLogOptions) Validate() []error {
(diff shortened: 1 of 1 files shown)


fedebongio:
Hi @lojies we assume you intend this one for 1.20, are we right?
/cc @tallclair

lojies:
@fedebongio yes.

liggitt:
This seems sensible to me. @deads2k, any concerns with this?

lojies:
@deads2k

sftim:
Hi.
I'd expect this command line argument to work like: `--audit-log-compress=gzip` or `--audit-log-compress=zstd` (which allows future extension). If I were implementing this I'd force setting the compressed encoding you plan to use rather than inferring gzip as a default, because then that allows a future change that drops gzip support.

Is that a good model for how it can work?

staging/src/k8s.io/apiserver/pkg/server/options/audit.go
```diff
@@ -449,6 +450,7 @@ func (o *AuditLogOptions) AddFlags(fs *pflag.FlagSet) {
 			strings.Join(pluginlog.AllowedFormats, ",")+".")
 	fs.StringVar(&o.GroupVersionString, "audit-log-version", o.GroupVersionString,
 		"API group and version used for serializing audit events written to log.")
+	fs.BoolVar(&o.Compress, "audit-log-compress", o.Compress, "Compress determines if the rotated log files should be compressed using gzip.")
```
deads2k, line 453: nit: "If set, the rotated log files will be compressed using gzip".

deads2k:
> Hi.
> I'd expect this command line argument to work like: `--audit-log-compress=gzip` or `--audit-log-compress=zstd` (which allows future extension). If I were implementing this I'd force setting the compressed encoding you plan to use rather than inferring gzip as a default, because then that allows a future change that drops gzip support.
> 
> Is that a good model for how it can work?

The base library doesn't provide that option.  If that library changed someday, with some effort this could be extended, but I wouldn't make it a prerequisite.

Looks straight-forward, but I'd like the nit on the help.

/approve

sftim:
(to be clear, I'm _not_ asking for the `zstd` encoding, just to leave room for adding it)
Also don't feel that this feedback should block approval if reviewers are happy.

PR #94082, opened 2020-08-18
Fix golint issues in pkg/controller/statefulset

ScrapCodes:
**What type of PR is this?**

/kind cleanup
/kind api-change

Cleans up all the golint issues in `pkg/controller/statefulset`

#68026

Fixes #

Currently, I am not sure, what goes in here. This patch does change some exposed functions.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/controller/statefulset/stateful_set.go
+++ b/pkg/controller/statefulset/stateful_set.go
@@ -48,13 +48,13 @@ import (
 // controllerKind contains the schema.GroupVersionKind for this controller type.
 var controllerKind = apps.SchemeGroupVersion.WithKind("StatefulSet")
 
-// StatefulSetController controls statefulsets.
-type StatefulSetController struct {
+// Controller controls statefulsets.
+type Controller struct {
 	// client interface
 	kubeClient clientset.Interface
 	// control returns an interface capable of syncing a stateful set.
 	// Abstracted out for testing.
-	control StatefulSetControlInterface
+	control ControlInterface
 	// podControl is used for patching pods.
 	podControl controller.PodControlInterface
 	// podLister is able to list/get pods from a shared informer's store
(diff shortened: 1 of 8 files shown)


neolit123:
thanks for the PR.

/ok-to-test
/priority backlog

ScrapCodes:
Hi @oomichi Thank you !, are there any next steps for me?

oomichi:
@ScrapCodes 
As https://github.com/kubernetes/kubernetes/pull/94082#issuecomment-675481724 you need to get approvals from approvers of
```
    hack/OWNERS
    pkg/controller/statefulset/OWNERS
    test/OWNERS
```

ScrapCodes:
@dims would you like to take a look ?

ScrapCodes:
/retest all

Trying to rerun tests, as some time has elapsed, since all the tests above passed.

ScrapCodes:
Hi @liggitt, you are one of the owners listed. 

Can you please take a look at this pr? In case, this finds you busy then can you suggest/ defer it to someone who might want to take a look.

Thanks !!

pkg/controller/statefulset/stateful_set.go
```diff
@@ -48,13 +48,13 @@ import (
 // controllerKind contains the schema.GroupVersionKind for this controller type.
 var controllerKind = apps.SchemeGroupVersion.WithKind("StatefulSet")
 
-// StatefulSetController controls statefulsets.
-type StatefulSetController struct {
+// Controller controls statefulsets.
+type Controller struct {
```
soltysh, line 52: Please don't change those, it's not desirable to blindly follow golint failures.

soltysh (changes requested):
/hold

pkg/controller/statefulset/stateful_set.go
ScrapCodes: Thank you so much for taking a look, I was myself in a similar impression. And asked the same question (here)[https://github.com/kubernetes/kubernetes/issues/68026#issuecomment-675431654].

 Is there is a way to ignore only a specific failure of Go lint, I am aware that we can have a comment to ignore an entire file. 

Shall I just close this PR? or is there a reason to hold - just trying to be sure.

soltysh: If this is the only change in your PR, then I guess so.

Issue #93925, opened 2020-08-12
Pod stuck in Terminating state


fasaxc:
**What happened**:

The Calico team is doing a churn test to test Calico but we seem to have run into a kubernetes issue.  Some pods are getting stuck in the Terminating state.  That's bad enough but it seems that

* kubelet is getting far enough that it calls the CNI plugin
* CNI plugin deletes the interface, thus freeing the IP for reuse
* kubelet doesn't remove the IP from the pod spec
* CNI plugin gets called for another pod and allocates the same IP to a new pod
* now we have two pods with the same IP allocated and they're both in states where Calico thinks that they should be networked (we keep networking up when a pod is in Terminating state so that apps that are gracefully shutting down can shut down network connections).

The describe output for the pod says
```
Events:
  Type     Reason     Age                    From                                                   Message
  ----     ------     ----                   ----                                                   -------
  Normal   Killing    21m (x205 over 7h3m)   kubelet, ip-172-16-102-106.us-west-2.compute.internal  Stopping container nginx
  Warning  Unhealthy  76s (x2531 over 7h2m)  kubelet, ip-172-16-102-106.us-west-2.compute.internal  Readiness probe failed: cannot exec in a stopped state: unknown
```

So looks like it's trying to kill the pod.

Examining `docker ps`, I see that the `k8s_POD` container is gone but the workload container is still running:
```
b8e1c8d7a96f        a61ec0c44569                                "nginx -g 'daemon of…"   8 hours ago          Up 8 hours                              k8s_nginx_server2-cali-ns-2-55b5df445f-mc9qm_cali-ns-2_f3dfc21f-605b-4270-9cd7-4cec7f3712a7_0
```

**What you expected to happen**:

kubelet cleans up the pod, including the workload container

**How to reproduce it (as minimally and precisely as possible)**:

This happened after running pod and namespace churn for a few hours.  We scale up and down nginx deployments behind services and clients that connect to them.  Namespaces with clients/servers come and go.

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.10-dispatcher", GitCommit:"f5757a1dee5a89cc5e29cd7159076648bf21a02b", GitTreeState:"clean", BuildDate:"2020-02-06T03:29:33Z", GoVersion:"go1.12.12b4", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #39606, opened 2017-01-09
optimize killPod() and syncPod() functions

NickrenREN:
make sure that one of the two arguments must be non-nil: runningPod, status ,just like the function note says
and judge the return value in syncPod() function before setting podKilled

PR #94084, opened 2020-08-18
Add logging when fail to kill container or pod

brianpursley:
**What type of PR is this?**
/kind feature

While investigating #93925, I noticed there are a couple places where killing a container or pod might fail without a message being logged. This PR adds logging when a container or pod fails to be killed.

```release-note
NONE
```

### DIFF
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -1524,6 +1524,8 @@ func (kl *Kubelet) syncPod(o syncPodOptions) error {
 		if !pcm.Exists(pod) && !firstSync {
 			if err := kl.killPod(pod, nil, podStatus, nil); err == nil {
 				podKilled = true
+			} else {
+				klog.Errorf("killPod for pod %q (podStatus=%v) failed: %v", format.Pod(pod), podStatus, err)
 			}
 		}
 		// Create and Update pod's Cgroups
--- a/pkg/kubelet/kuberuntime/kuberuntime_container.go
+++ b/pkg/kubelet/kuberuntime/kuberuntime_container.go
@@ -658,6 +658,7 @@ func (m *kubeGenericRuntimeManager) killContainersWithSyncResult(pod *v1.Pod, ru
 			killContainerResult := kubecontainer.NewSyncResult(kubecontainer.KillContainer, container.Name)
 			if err := m.killContainer(pod, container.ID, container.Name, "", gracePeriodOverride); err != nil {
 				killContainerResult.Fail(kubecontainer.ErrKillContainer, err.Error())
+				klog.Errorf("killContainer %q(id=%q) for pod %q failed: %v", container.Name, container.ID, format.Pod(pod), err)
 			}
 			containerResults <- killContainerResult
 		}(container)


pkg/kubelet/kubelet.go
```diff
@@ -1524,6 +1524,8 @@ func (kl *Kubelet) syncPod(o syncPodOptions) error {
 		if !pcm.Exists(pod) && !firstSync {
 			if err := kl.killPod(pod, nil, podStatus, nil); err == nil {
 				podKilled = true
+			} else {
+				klog.Errorf("killPod for pod %q failed: %v", format.Pod(pod), err)
```
SergeyKanzhelev, line 1528: I wonder if in this case it should return as well. Alternative is that pod is terminating, runtime failed to kill the pod, and  will not be restarted.

Also, interesting thing here is that condition was added https://github.com/kubernetes/kubernetes/pull/39606 seemingly for the case when `podStatus` is `nil`. So it will not even attempt to call into the runtime

brianpursley:
I went ahead and updated this PR with the changes I think are correct. 

If they are not, I can update this PR, but I think it seems right to not kill the sandbox until the containers are killed.  I updated the description accordingly.

brianpursley:
@SergeyKanzhelev I certainly think there is some opportunity to clean up the pod/container kill code.

I'm not confident enough that changing the behavior will be OK at this point, which is in part why I would like to add some logging... to see if there are failures to kill pods/containers or not.

I can follow-on with another PR to change the behavior to return, but I would like to get some extra eyes on that change.

pkg/kubelet/kubelet.go
SergeyKanzhelev, line 1528: as you picking the route of Can you also log the `podStatus`. I'm still concerned that this `if` was introduced in the PR #39606 that is checking for `podStatus` to be `nil`.

SergeyKanzhelev:
> @SergeyKanzhelev I certainly think there is some opportunity to clean up the pod/container kill code.
> 
> I'm not confident enough that changing the behavior will be OK at this point, which is in part why I would like to add some logging... to see if there are failures to kill pods/containers or not.
> 
> I can follow-on with another PR to change the behavior to return, but I would like to get some extra eyes on that change.

Do you have an environment where you can try to repro the bug that initiated this PR? If so, this gradual approach totally makes sense.

brianpursley:
> > @SergeyKanzhelev I certainly think there is some opportunity to clean up the pod/container kill code.
> > I'm not confident enough that changing the behavior will be OK at this point, which is in part why I would like to add some logging... to see if there are failures to kill pods/containers or not.
> > I can follow-on with another PR to change the behavior to return, but I would like to get some extra eyes on that change.
> 
> Do you have an environment where you can try to repro the bug that initiated this PR? If so, this gradual approach totally makes sense.

Not sure if I can reproduce it, but I can try in a local cluster.

pkg/kubelet/kubelet.go
brianpursley: Sure, I updated the log message to include podStatus

Issue #93496, opened 2020-07-28
[Failing test] TestEndpointSliceMirroring/Service_without_selector


liggitt:
**Which jobs are flaking**:

ci-kubernetes-integration-master

**Which test(s) are flaking**:

ci-kubernetes-integration-master

**Testgrid link**:

https://testgrid.k8s.io/sig-release-master-blocking#integration-master&width=5&include-filter-by-regex=TestEndpointSliceMirroring

**Reason for failure**:

```
    endpointslicemirroring_test.go:202: Waiting for EndpointSlice to be created
    endpointslicemirroring_test.go:222: Timed out waiting for conditions: timed out waiting for the condition
    --- FAIL: TestEndpointSliceMirroring/Service_without_selector (5.05s)
```

**Anything else we need to know**:

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-integration-master/1287935223459221504

5 seconds seems aggressive for a controller test run in parallel with other tests ... unless this is testing performance-oriented changes, wait.ForeverTestTimeout seems more appropriate

athenabot:
/triage unresolved

Comment `/remove-triage unresolved` when the issue is assessed and confirmed.

🤖 I am a bot run by vllry. 👩‍🔬

robscott:
Thanks for filing this! I've opened https://github.com/kubernetes/kubernetes/pull/93497 to fix this.

/remove-triage unresolved

liggitt:
still saw one occurrence at https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-integration-master/1293374774533165056

liggitt:
in presubmits: https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-integration&include-filter-by-regex=TestEndpointSliceMirroring&width=5

`TestEndpointSliceMirroring/Service_with_selector` races the endpoints controller by creating the service object first, then the endpoints object, and not tolerating an AlreadyExists error because the controller created the endpoints object first:

https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-integration/1295516772354494464

liggitt:
For the `TestEndpointSliceMirroring/Service_without_selector` failures, I don't see any evidence the endpointslicemirroring controller is running

Failures:
[...]

PR #94086, opened 2020-08-18
Updating EndpointSlice controller to wait for all caches to be synced

robscott:
**What type of PR is this?**
/kind bug

Previously the EndpointSlice controller was not waiting for EndpointSlices or Nodes to be synced.

As identified by @liggitt in https://github.com/kubernetes/kubernetes/issues/93496#issuecomment-675640236

```release-note
The EndpointSlice controller now waits for EndpointSlice and Node caches to be synced before starting.
```

/sig network
/priority critical-urgent
/cc @freehan @liggitt

### DIFF
--- a/pkg/controller/endpointslice/endpointslice_controller.go
+++ b/pkg/controller/endpointslice/endpointslice_controller.go
@@ -237,7 +237,7 @@ func (c *Controller) Run(workers int, stopCh <-chan struct{}) {
 	klog.Infof("Starting endpoint slice controller")
 	defer klog.Infof("Shutting down endpoint slice controller")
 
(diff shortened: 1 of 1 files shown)


liggitt:
thanks, did you sweep all the controllers involved in endpointslices to make sure they waited correctly?

liggitt:
/lgtm
/approve
/milestone v1.19

required to avoid endpointslice churn on startup. this should be picked to all releases where the controller was enabled

robscott:
@liggitt thanks for checking! I did confirm that both Endpoints and EndpointSliceMirroring controllers are waiting for the appropriate caches. For the record, EndpointSliceMirroring waits for endpoints, endpointslices, and services. Endpoints waits for pods, services, and endpoints.

Issue #93338, opened 2020-07-22
Kubelet rejects pod scheduled based on newly added node labels which have not been observed by the kubelet yet


Joseph-Goergen:
**What happened**:

Whenever a node is added or updated, there is a small window where pods are scheduled to that node, before any beta labels are applied to it. This can cause issues with pods that are queued up to be scheduled and that have a NodeAffinity (in our case) to the now deprecated `beta.kubernetes.io/os` label.

**What you expected to happen**:

The proper labels to be applied to workers before the scheduling of pods on that node.

**How to reproduce it (as minimally and precisely as possible)**:

(Not 100 percent success rate)
- deploy 1.19 cluster with no workers
- apply a deployment with a node affinity for the beta.kubernetes.io/os label
- add worker

**Anything else we need to know?**:

I have been told this step used to be done on the worker side, but is now done on the master side. Which could explain why this is happening. https://github.com/kubernetes/kubernetes/blob/v1.19.0-rc.2/pkg/controller/nodelifecycle/node_lifecycle_controller.go#L1534-L1578

/sig scheduling

liggitt:
> This can cause issues with pods that are queued up to be scheduled and that have a NodeAffinity (in our case) to the now deprecated beta.kubernetes.io/os label.

Can you describe the issue?

liggitt:
The controller currently ensures the beta OS and OS labels match. What issue are you seeing occur with affinity?

rtheis:
@liggitt We believe the flow is as follows:
1. Deploy control plane components, API server etc.
1. Deploy addon components (https://github.com/operator-framework/operator-lifecycle-manager in our case).  These components will stay in Pending status until worker nodes are ready.
1. Deploy worker nodes.
1. Worker node X is registered with the API server (no `beta.kubernetes.io/os` label yet).
1. Scheduler schedules the OLM addon catalog pod which uses the `beta.kubernetes.io/os` label. This pod ends up in NodeAffinity status.
1. Control plan adds the `beta.kubernetes.io/os` label to worker node X.
1. Deleting the OLM addon catalog pod triggers a new pod to be created which is scheduled properly.

There is a timing window here since we don't always see the problem.

liggitt:
> 2\. Deploy addon components (https://github.com/operator-framework/operator-lifecycle-manager in our case).  These components will stay in Pending status until worker nodes are ready.

Do you have the manifests that are deployed?

[...]

PR #91046, opened 2020-05-13
kubelet no longer registers "beta.kubernetes.io/os" and "beta.kubernetes.io/arch" node labels to apiserver

wawa0210:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
/kind cleanup

**What this PR does / why we need it**:

- [x] Starting from 1.14, this type of label is marked as deprecated and will be removed after 1.18

- [x] Kubernetes / kubernetes-sigs projects and documentation has been replaced

**Some repo have exceptions**:

1.  https://github.com/kubernetes/minikube/pull/7353
2. https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/77

**This pr will cover the following**:

*  Update the node lifecycle controller to treat the GA os/arch labels as the authoritative source and copy them to the beta labels (flip the source of truth)

*  Update the kubelet to only set the GA labels

**The remaining to-do items are as follows**:

- [ ] no earlier than 1.22: update the node lifecycle controller to drop the beta labels if they differ from the GA labels

**Which issue(s) this PR fixes**:

Fixes #89477

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
`beta.kubernetes.io/os` and `beta.kubernetes.io/arch` node labels are deprecated. Update node selectors to use `kubernetes.io/os` and `kubernetes.io/arch`.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```
/assign @dchen1107

PR #94087, opened 2020-08-18
kubelet waits for node lister to sync at least once

### LINKED ISSUE #92067: Kubernetes scheduler spams cluster with pods in NodeAffinity status
**What happened**:

A deployment is created that uses a `nodeSelector` after the cluster master is created but before the cluster worker nodes are fully initialized.  During the initialization of the worker nodes, the nodes are temporarily available for scheduling without the necessary label to match the deployment's node selector.  Depending on how long the nodes are available for scheduling without the necessary node labels for the deployment, the scheduler will start to spam with cluster with pods in `NodeAffinity` status.  This spamming stops once the worker nodes are fully initialized and the pods are scheduled successful.  The exact timing of all of this is TBD.

Pods in `NodeAffinity` status are annoying and require manual cleanup.  It would be preferred that the scheduler avoid this an simply issue warning events.

See issue description.

**Anything else we need to know?**: No.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.18.3
- Cloud provider or hardware configuration: IBM Cloud Kubernetes Service
- OS (e.g: `cat /etc/os-release`): Ubuntu 18.04.4 LTS
- Kernel (e.g. `uname -a`): 4.15.0-96-generic
- Install tools: Managed service install
- Others: N/A

derekwaynecarr:
**What type of PR is this?**
/kind bug

If node has a kube client, it will wait to ensure the node lister has synced at least once before trusting its data.

Fixes https://github.com/kubernetes/kubernetes/issues/92067

still under discussion where to best wait (in getter, or in main kubelet startup)

```release-note
NONE
```

### DIFF
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -125,6 +125,9 @@ const (
 	// Max amount of time to wait for the container runtime to come up.
 	maxWaitForContainerRuntime = 30 * time.Second
 
+	// Max amount of time to wait for node list/watch to initially sync
+	maxWaitForAPIServerSync = 10 * time.Second
+
 	// nodeStatusUpdateRetry specifies how many times kubelet retries when posting node status failed.
 	nodeStatusUpdateRetry = 5
 
@@ -431,14 +434,31 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 		serviceHasSynced = func() bool { return true }
 	}
 
-	nodeIndexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
+	var nodeHasSynced cache.InformerSynced
+	var nodeLister corelisters.NodeLister
+
 	if kubeDeps.KubeClient != nil {
-		fieldSelector := fields.Set{api.ObjectNameField: string(nodeName)}.AsSelector()
-		nodeLW := cache.NewListWatchFromClient(kubeDeps.KubeClient.CoreV1().RESTClient(), "nodes", metav1.NamespaceAll, fieldSelector)
-		r := cache.NewReflector(nodeLW, &v1.Node{}, nodeIndexer, 0)
-		go r.Run(wait.NeverStop)
+		kubeInformers := informers.NewSharedInformerFactoryWithOptions(kubeDeps.KubeClient, 0, informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+			options.FieldSelector = fields.Set{api.ObjectNameField: string(nodeName)}.String()
+		}))
+		nodeLister = kubeInformers.Core().V1().Nodes().Lister()
+		nodeHasSynced = func() bool {
+			if kubeInformers.Core().V1().Nodes().Informer().HasSynced() {
+				klog.Infof("kubelet nodes sync")
+				return true
+			}
+			klog.Infof("kubelet nodes not sync")
+			return false
+		}
+		kubeInformers.Start(wait.NeverStop)
+		klog.Info("Kubelet client is not nil")
+	} else {
+		// we dont have a client to sync!
+		nodeIndexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
+		nodeLister = corelisters.NewNodeLister(nodeIndexer)
+		nodeHasSynced = func() bool { return true }
+		klog.Info("Kubelet client is nil")
 	}
-	nodeLister := corelisters.NewNodeLister(nodeIndexer)
 
 	// TODO: get the real node object of ourself,
 	// and use the real node name and UID.
@@ -489,6 +509,7 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 		serviceLister:                           serviceLister,
 		serviceHasSynced:                        serviceHasSynced,
 		nodeLister:                              nodeLister,
+		nodeHasSynced:                           nodeHasSynced,
 		masterServiceNamespace:                  masterServiceNamespace,
 		streamingConnectionIdleTimeout:          kubeCfg.StreamingConnectionIdleTimeout.Duration,
 		recorder:                                kubeDeps.Recorder,
@@ -873,7 +894,9 @@ type Kubelet struct {
 	serviceHasSynced cache.InformerSynced
 	// nodeLister knows how to list nodes
 	nodeLister corelisters.NodeLister
-
+	// nodeHasSynced indicates whether nodes have been sync'd at least once.
+	// Check this before trusting a response from the node lister.
+	nodeHasSynced cache.InformerSynced
 	// a list of node labels to register
 	nodeLabels map[string]string
 
--- a/pkg/kubelet/kubelet_getters.go
+++ b/pkg/kubelet/kubelet_getters.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"net"
 	"path/filepath"
+	"time"
 
 	cadvisorapiv1 "github.com/google/cadvisor/info/v1"
 	cadvisorv2 "github.com/google/cadvisor/info/v2"
@@ -32,6 +33,7 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/kubernetes/pkg/kubelet/cm"
 	"k8s.io/kubernetes/pkg/kubelet/config"
 	kubecontainer "k8s.io/kubernetes/pkg/kubelet/container"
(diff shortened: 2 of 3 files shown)


derekwaynecarr:
@sjenning idea is to do similar as seen here, just need to work out pro/con for where we wait for node lister to sync at least once.

derekwaynecarr:
after following up with @deads2k seems like it would be a good time to move away from list/watch and to a node informer with filter, then we can just use regular HasSynced function.  will iterate, just trying to determine the best location to wait for the node informer to sync at least once and not cause issues (do not want to do it in the kubelet_getters if possible)

sjenning:
this might also fix https://github.com/kubernetes/kubernetes/issues/93338

derekwaynecarr:
@sjenning take a look, the idea is to move the predicate function to call GetNode, which if we have a kubeclient ensures we synced the node lister at least once (otherwise, it errors).  I am still trying to think through an edge case for the scheduler scenario that we may miss, but this would ensure that we do not fall back to the kubelet "initialNode" and stop tripping up scheduler.

pkg/kubelet/kubelet_getters.go
```diff
@@ -245,7 +256,7 @@ func (kl *Kubelet) GetNode() (*v1.Node, error) {
 // zero capacity, and the default labels.
 func (kl *Kubelet) getNodeAnyWay() (*v1.Node, error) {
 	if kl.kubeClient != nil {
-		if n, err := kl.nodeLister.Get(string(kl.nodeName)); err == nil {
+		if n, err := kl.GetNode(); err == nil {
```
sjenning, line 259: pretty sure this whole function collapses to just `return kl.GetNode()` since `GetNode()` contains the standalone logic

pkg/kubelet/kubelet_getters.go
```diff
@@ -235,6 +237,15 @@ func (kl *Kubelet) GetNode() (*v1.Node, error) {
 	if kl.kubeClient == nil {
 		return kl.initialNode(context.TODO())
 	}
+	// if we have a valid kube client, we wait up to 5s for initial lister to sync
+	if !kl.nodeHasSynced() {
+		err := wait.PollImmediate(time.Second, 5*time.Second, func() (bool, error) {
```
deads2k, line 242: choose a weird number so we can find this later if it comes up. How about 8 seconds

pkg/kubelet/kubelet_getters.go
derekwaynecarr: GetNode does not guarantee a response in case where kubeclient is not nil, whereas getNodeAnyway does (it falls back to initial node).

cwdsuzhou:
Hi we have run into the same issue and open this pr https://github.com/kubernetes/kubernetes/pull/95824

seems similar with this one.


I have a question that why not use cache.WaitForCacheSynced.

alculquicondor:
/priority critical-urgent

What is this missing to not be a WIP?

liggitt:
Kubelets dropping the beta OS/Arch labels on registration in https://github.com/kubernetes/kubernetes/pull/91046 made this issue much more common. I think we should re-enable kubelets self-adding the beta labels in 1.19/1.20 to preserve the bring-up characteristics of nodes and deployments/daemonsets targeting well-known labels, then try to land this in 1.21. Opened https://github.com/kubernetes/kubernetes/pull/96810 to do that

alculquicondor:
Can this be prioritized?

Cherry-pick deadline for the 1.19, 1.18, and 1.17 branches is Friday, December 04, 2020 EOD PT.

ehashman:
@liggitt I see #96810 was merged, is there anything left blocking this PR?

derekwaynecarr:
i want to look at this again, i am still confused how a test that was red earlier went green without any clear change on my part.

alculquicondor:
Could it have been #96810?
Or is the test using other labels?

ehashman:
/retitle kubelet waits for node lister to sync at least once
/hold

talked to Derek about this, it's not really a WIP but the tests were previously failing. I'm going to look into why they turned green and see if we can get this merged

ehashman:
Me and @troy0820 have reviewed this in depth, I am comfortable with this change assuming tests pass. There's no test history available so we can't confirm why this was previously failing or if @liggitt's PR is the reason things turned green.

@derekwaynecarr I'm going to remove the hold but I'll watch for failures on the retest.

/lgtm
/hold cancel

liggitt:
> There's no test history available so we can't confirm why this was previously failing or if @liggitt's PR is the reason things turned green.

You could run the tests against a combination of this PR and a revert of https://github.com/kubernetes/kubernetes/pull/96810 if you wanted to verify this addressed the issue

ehashman:
/hold

I'll throw up a PR to try that

ehashman:
@liggitt thus far I've been unable to reproduce any failures with #96810 reverted on #97931, I'm retesting a third time

derekwaynecarr:
i am fine removing my own hold on this as its possible what i saw was related to another unreliable aspect of CI at that time.

/hold cancel

pkg/kubelet/kubelet.go
```diff
@@ -431,14 +434,31 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
...
+		kubeInformers := informers.NewSharedInformerFactoryWithOptions(kubeDeps.KubeClient, 0, informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+			options.FieldSelector = fields.Set{api.ObjectNameField: string(nodeName)}.String()
+		}))
+		nodeLister = kubeInformers.Core().V1().Nodes().Lister()
+		nodeHasSynced = func() bool {
+			if kubeInformers.Core().V1().Nodes().Informer().HasSynced() {
+				klog.Infof("kubelet nodes sync")
+				return true
+			}
+			klog.Infof("kubelet nodes not sync")
```
neolit123, line 450: this message is posted *a lot* in the logs, i think it can be omitted.

pkg/kubelet/kubelet_getters.go
```diff
@@ -235,6 +237,15 @@ func (kl *Kubelet) GetNode() (*v1.Node, error) {
...
 	}
+	// if we have a valid kube client, we wait for initial lister to sync
+	if !kl.nodeHasSynced() {
+		err := wait.PollImmediate(time.Second, maxWaitForAPIServerSync, func() (bool, error) {
+			return kl.nodeHasSynced(), nil
+		})
+		if err != nil {
+			return nil, fmt.Errorf("nodes have not yet been read at least once, cannot construct node object")
+		}
+	}
```
neolit123, line 248: this poll is problematic.
GetNode is called in a loop, which means that the poll is present on each call.

the sequence looks like:
- GetNode()
  - Poll for nodeHasSynced()
  - Poll ....
- GetNode()
  - Poll for nodeHasSynced()
  - Poll ....

if there is a valid client, arguably the informer sync check should be outside of this loop.

neolit123:
the Minikube maintainers (cc @medyagh) [found](https://github.com/kubernetes/kubeadm/issues/2395) that this change introduced a performance regression.
when client credentials are fed to a primary kubelet in a cluster (`--kubeconfig`), but there are no other nodes or an API server running yet (because this primary kubelet will boot it from a static pod), this blocks for extra ~40 seconds.

i can see the informer wait as something desired but i think the logic here can be improved.
cc @derekwaynecarr @deads2k @liggitt 

we also backported this change in the support skew and since we don't have direct performance tests, nobody saw it until now.

pkg/kubelet/kubelet.go
pacoxu: Line 447 `klog.Infof("kubelet nodes sync")` is removed in #98137

pkg/kubelet/kubelet_getters.go
alculquicondor: Wouldn't it be enough to check `HasSynced` once before creating the Kubelet object?

neolit123: it would only sync after the api server is up.

alculquicondor: if the API server is down, there is nothing to sync. Am I missing something?

neolit123: BTW we are discussion how to improve this in:
https://github.com/kubernetes/kubernetes/pull/99336

if the kubelet is managing the first server instance in the cluster as a static pod and if the HasSynced check is before the kubelet object creation this means for such a kubelet instance the single HasSynced check will always be false.

maybe that's what we want, given this is the first node in the cluster and there is no need to sync it.

for subsequent kubelets the check should pass.

pkg/kubelet/kubelet.go
neolit123: it's still present at HEAD:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet.go#L452

snowplayfire:
If these pods are re-scheduled automatically,  can the problem be solved? @derekwaynecarr

PR #94089, opened 2020-08-18
Fixing a flaky EndpointSliceMirroring integration test

robscott:
**What type of PR is this?**
/kind failing-test

This test was trying to create an Endpoints resource that the Endpoints controller would also attempt to create. This could result in a failure if the Endpoints controller created the resource before the test did.

This should fix at least part of https://github.com/kubernetes/kubernetes/issues/93496. 

/sig network
/priority critical-urgent
/cc @freehan @liggitt

### DIFF
--- a/test/integration/endpointslice/endpointslicemirroring_test.go
+++ b/test/integration/endpointslice/endpointslicemirroring_test.go
@@ -84,7 +84,7 @@ func TestEndpointSliceMirroring(t *testing.T) {
 	testCases := []struct {
 		testName                     string
 		service                      *corev1.Service
-		endpoints                    *corev1.Endpoints
+		customEndpoints              *corev1.Endpoints
 		expectEndpointSlice          bool
 		expectEndpointSliceManagedBy string
 	}{{
@@ -102,11 +102,6 @@ func TestEndpointSliceMirroring(t *testing.T) {
 				},
 			},
 		},
-		endpoints: &corev1.Endpoints{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: "test-123",
-			},
-		},
 		expectEndpointSlice:          true,
 		expectEndpointSliceManagedBy: "endpointslice-controller.k8s.io",
 	}, {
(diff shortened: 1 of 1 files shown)


liggitt:
/lgtm
/approve
/milestone v1.19

including test-only change to improve release CI signal

robscott:
Flake in `TestUnschedulablePodBecomesSchedulable/other_pod_gets_deleted`
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/94089/pull-kubernetes-integration/1295857916779171842
/retest

robscott:
Flake on pull-kubernetes-e2e-gce-100-performance
```
testing/density/config.yaml expand_less | 7m10s
-- | --
:0 [measurement call APIResponsivenessPrometheus - APIResponsivenessPrometheusSimple error: top latency metric: there should be no high-latency requests
```
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/94089/pull-kubernetes-e2e-gce-100-performance/1295857916787560448/

/retest

liggitt:
Hmm. That's two of the same failure. Clearly not this PR since this is changing an integration test. @kubernetes/sig-scalability do we have a regression in master?

liggitt:
maybe not new. Looks sporadic for a while in https://storage.googleapis.com/k8s-gubernator/triage-go/index.html?ci=0&pr=1&text=APIResponsivenessPrometheus&job=performance

wojtek-t:
> maybe not new. Looks sporadic for a while in https://storage.googleapis.com/k8s-gubernator/triage-go/index.html?ci=0&pr=1&text=APIResponsivenessPrometheus&job=performance

Yeah - it's not super new, happens from time to time.
We need to look deeper, but based on when I was looking into scheduler throughput issues 2 weeks ago or so, I found that master seem to be throttled from time to time. I think it may be the reason for those flakes, but we need to look a bit deeper into it. But that's on our radar.

wojtek-t:
@mm4tt ^^

mm4tt:
Opened https://github.com/kubernetes/kubernetes/issues/94098 to track this

PR #94090, opened 2020-08-18
remove pod presets

### LINKED ISSUE #93900: Pod Presets still in alpha since v1.6, time to remove it?
- https://kubernetes.io/docs/concepts/workloads/pods/podpreset/
- https://github.com/kubernetes/enhancements/issues/162

There was an effort at one point to promote it to Beta https://github.com/kubernetes/enhancements/issues/348

This may have moved to Service Catalog repo https://github.com/kubernetes-sigs/service-catalog/search?q=podpreset&type=Commits

So question, do we remove it from kubernetes/kubernetes ?

deads2k:
fixes https://github.com/kubernetes/kubernetes/issues/93900

This has lingered since 1.single-digit.  In keeping with not allowing permabeta and following the example of sig-auth removing alpha features that aren't graduating, this PR removes PodPresets.

/priority important-soon
/kind api

```release-note
The v1alpha1 PodPreset API and admission plugin has been removed with no built-in replacement. Admission webhooks can be used to modify pods on creation.
```

@kubernetes/sig-service-catalog 
@kubernetes/api-approvers


vendor/modules.txt
```diff
@@ -1582,7 +1582,6 @@ k8s.io/api/rbac/v1beta1
 k8s.io/api/scheduling/v1
 k8s.io/api/scheduling/v1alpha1
 k8s.io/api/scheduling/v1beta1
-k8s.io/api/settings/v1alpha1
```
liggitt, line 1585: looks like a few other packages need cleaning up from this file, hack/update-vendor.sh should resolve it

liggitt (commented):
also drop `staging/src/k8s.io/api/testdata/HEAD/settings.k8s.io.v1alpha1.PodPreset.*`

liggitt:
looks like the admission plugin was enabled in test (the API wasn't):

https://github.com/kubernetes/kubernetes/blob/3352c44949684edc2a72fc7891fbaa2273a2aa35/cluster/gce/config-test.sh#L408

failing the e2e job, https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/94090/pull-kubernetes-e2e-gce-ubuntu-containerd/1296120899644166147/artifacts/e2e-158b911473-a7d53-master/kube-apiserver.log:
```
Error: enable-admission-plugins plugin "PodPreset" is unknown
```

liggitt:

liggitt:
Yeah, https://github.com/kubernetes/kubernetes/pull/94090#issuecomment-676562711

liggitt:
needs rebase

dims:
woohoo!

dims:
needs a rebase @deads2k

PR #94102, opened 2020-08-19
kubeadm: adjust the logic around etcd data directory creation

neolit123:

- Ensure the directory is created with 0700 via a new function
called CreateDataDirectory().
- Call this function in the init phases instead of the manual call
to MkdirAll.
- Call this function when joining control-plane nodes with local etcd.

If the directory creation is left to the kubelet via the
static Pod hostPath mounts, it will end up with 0755
which is not desired.

xref https://github.com/kubernetes/kubeadm/issues/2256

should be backported in case older versions of kubeadm decide to use newer etcd, which is a supported scenario by kubeadm.

```release-note
kubeadm: ensure the etcd data directory is created with 0700 permissions during control-plane init and join
```

```docs

```

### DIFF
--- a/cmd/kubeadm/app/cmd/phases/join/controlplanejoin.go
+++ b/cmd/kubeadm/app/cmd/phases/join/controlplanejoin.go
@@ -29,6 +29,7 @@ import (
 	etcdphase "k8s.io/kubernetes/cmd/kubeadm/app/phases/etcd"
 	markcontrolplanephase "k8s.io/kubernetes/cmd/kubeadm/app/phases/markcontrolplane"
 	uploadconfigphase "k8s.io/kubernetes/cmd/kubeadm/app/phases/uploadconfig"
+	etcdutil "k8s.io/kubernetes/cmd/kubeadm/app/util/etcd"
 )
 
 var controlPlaneJoinExample = cmdutil.Examples(`
@@ -129,6 +130,11 @@ func runEtcdPhase(c workflow.RunData) error {
 	if cfg.Etcd.External != nil {
 		fmt.Println("[control-plane-join] using external etcd - no local stacked instance added")
 		return nil
+	}
+
+	// Create the etcd data directory
+	if err := etcdutil.CreateDataDirectory(cfg.Etcd.Local.DataDir); err != nil {
+		return err
 	}
 
 	// Adds a new etcd instance; in order to do this the new etcd instance should be "announced" to
--- a/cmd/kubeadm/app/cmd/phases/init/BUILD
+++ b/cmd/kubeadm/app/cmd/phases/init/BUILD
@@ -44,6 +44,7 @@ go_library(
         "//cmd/kubeadm/app/preflight:go_default_library",
         "//cmd/kubeadm/app/util/apiclient:go_default_library",
         "//cmd/kubeadm/app/util/dryrun:go_default_library",
+        "//cmd/kubeadm/app/util/etcd:go_default_library",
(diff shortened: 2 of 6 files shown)


neolit123:
/approve cancel

there is some discussion here whether this should be in 1.19 if a fix in etcd is made:
https://github.com/kubernetes/kubeadm/issues/2256#issuecomment-676583030

fabriziopandini:
+1 to wait for an etcd fix (same as Kubernetes)
forcing to 0700 might break all sort of thing around this folder (e.g. backups), and this step should be taken only if there is no alternatives

rosti (approved):
Thanks @neolit123 !
/lgtm

rosti:
@fabriziopandini @neolit123 the chmod call is not recursive, plus stomping on the permissions there and setting up secure sane permissions is a good thing to do. Imagine someone with 0777 there. Practically everyone can edit or delete the data store.

cmd/kubeadm/app/cmd/phases/join/controlplanejoin.go
```diff
@@ -131,6 +132,12 @@ func runEtcdPhase(c workflow.RunData) error {
 		return nil
 	}
 
+	// Create/chmod the etcd data directory
+	if err := etcdutil.CreateDataDirectory(cfg.Etcd.Local.DataDir); err != nil {
+		return err
+
```
neolit123, line 138: todo: fix whitespace

neolit123:
@rosti 

> @fabriziopandini @neolit123 the chmod call is not recursive, plus stomping on the permissions there and setting up secure sane permissions is a good thing to do. Imagine someone with 0777 there. Practically everyone can edit or delete the data store.

i guess if we enforce the permission it's more good than harm.

@fabriziopandini 
> +1 to wait for an etcd fix (same as Kubernetes)
forcing to 0700 might break all sort of thing around this folder (e.g. backups), and this step should be taken only if there is no alternatives

we've heard about a potential fix yesterday, but not 100% sure it will be applied to etcd.

backups would required root interaction probably and given kubeadm requires root they can chmod the directory to change permissions before backup if they wish to.

SataQiu (commented):
/lgtm

fabriziopandini:
@neolit123 if I got this right etcd is going to be fixed (it will throw a warning in case of permission different than 0700, but not block).

neolit123:
this is what i saw too.

we need to make sure that in 1.20 kubeadm defaults to an etcd version that does not break.
in the meantime the chmod part of this patch can be removed, as we don't want to override user chmod settings.

/milestone v1.20

neolit123:
> milestone v1.20

actually, we the new milestone rules critical fixes should merge last.

/milestone clear

rosti:
If etcd folks decide to change the error to a warning, I'd suggest that we do a pre-flight check to verify the permissions if the etcd data dir exists and dump a warning on our side. In that case we should remove the chmod, but I would advocate to keep the mkdirs with 0700 so that kubeadm creates this directory with sane default permissions if it doesn't exist.

neolit123:
updated the PR to only create the directory if it does not exist on init/join-control-plane, but not chmod it.

/remove-priority important-soon
/priority backlog
lowering priority since the fix in etcd was applied and 1.20 will include 3.4.13+.

neolit123:
/milestone v1.20
/approve

/cc @rosti @SataQiu 
looking for LGTM.

fabriziopandini (commented):
/lgtm

PR #94107, opened 2020-08-19
Updating kube-proxy to trim space from loadBalancerSourceRanges

robscott:
**What type of PR is this?**
/kind bug

This updates kube-proxy to trim any space from loadBalancerSourceRanges to match validation. A better long term fix will require updating Service validation for future Kubernetes versions to not trim spaces before validation.

This feels like the kind of fix that should be part of 1.19 and potentially backported to older Kubernetes versions.

```release-note
kube-proxy now trims extra spaces found in loadBalancerSourceRanges to match Service validation.
```

/sig network
/priority critical-urgent
/assign @bowei

### DIFF
--- a/pkg/proxy/service.go
+++ b/pkg/proxy/service.go
@@ -146,10 +146,14 @@ func (sct *ServiceChangeTracker) newBaseServiceInfo(port *v1.ServicePort, servic
 		topologyKeys:           service.Spec.TopologyKeys,
 	}
 
+	loadBalancerSourceRanges := make([]string, len(service.Spec.LoadBalancerSourceRanges))
+	for i, sourceRange := range service.Spec.LoadBalancerSourceRanges {
+		loadBalancerSourceRanges[i] = strings.TrimSpace(sourceRange)
+	}
+
 	if sct.isIPv6Mode == nil {
 		info.externalIPs = make([]string, len(service.Spec.ExternalIPs))
-		info.loadBalancerSourceRanges = make([]string, len(service.Spec.LoadBalancerSourceRanges))
-		copy(info.loadBalancerSourceRanges, service.Spec.LoadBalancerSourceRanges)
+		info.loadBalancerSourceRanges = loadBalancerSourceRanges
 		copy(info.externalIPs, service.Spec.ExternalIPs)
 		// Deep-copy in case the service instance changes
 		info.loadBalancerStatus = *service.Status.LoadBalancer.DeepCopy()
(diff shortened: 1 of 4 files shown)


pkg/proxy/service.go
```diff
@@ -146,10 +146,14 @@ func (sct *ServiceChangeTracker) newBaseServiceInfo(port *v1.ServicePort, servic
 		topologyKeys:           service.Spec.TopologyKeys,
 	}
 
+	loadBalancerSourceRanges := make([]string, len(service.Spec.LoadBalancerSourceRanges))
+	for i, sourceRange := range service.Spec.LoadBalancerSourceRanges {
+		loadBalancerSourceRanges[i] = strings.TrimSpace(sourceRange)
+	}
```
bowei, line 152: Does this solve the actual crashloop though? If the input is invalid

robscott:
@bowei I've added another check and test now for better coverage here. Now if syncProxyRules ever runs into an invalid CIDR it simply drops it - this happens after TrimSpace is called so extra spaces do not invalidate a CIDR. I don't think this should ever get hit though. The validation for this field actually trims spaces before running validation. Whether or not that's a good idea I'm not sure, but we were not doing that in kube-proxy as a consumer of the API. Changing the core validation would require a much longer release cycle and would likely be relatively difficult since the actual code that should change is in the utils repo.

If you're curious, here's the validation flow:
1. [Core Service validation calls GetLoadBalancerSourceRanges](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L4335-L4338)
2. [GetLoadBalancerSourceRanges calls utilnet.ParseIPNets](https://github.com/kubernetes/kubernetes/blob/master/pkg/api/service/util.go#L50) 
3. [utilnet.ParseIPNets call strings.TrimSpace before validating](https://github.com/kubernetes/utils/blob/master/net/ipnet.go#L32).

pkg/proxy/service.go
robscott: Thanks for double checking. I did some more thorough testing to ensure this would fix the problem and have given a more detailed explanation below.

aojea:
shouldn't we caught this in validation?
https://github.com/kubernetes/kubernetes/blob/14a1106/pkg/apis/core/validation/validation.go#L4143

robscott:
@aojea you would think so, but unfortunately not. Outlined the flow in https://github.com/kubernetes/kubernetes/pull/94107#issuecomment-677988557. Unfortunately `strings.TrimSpace` is called before any meaningful validation is run but not before kube-proxy used the same field.

bowei:
The validation function is technically wrong -- validation should not be mutating the input before validating. An IP address with extra whitespace is not a valid IP address.
@robscott will file an issue to track potentially fixing that behavior. It's likely complicated to fix due to how old it is in the codebase.

aojea:
yeah, didn't mean to ask for changes. I was curious about what will be the "ideal" solution
> An IP address with extra whitespace is not a valid IP address.
... , and I was thinking if validation should reject the IP as invalid ...
> It's likely complicated to fix due to how old it is in the codebase.
absolutely agree on this

liggitt:

/retest

cpanato:
@robscott @bowei do we need to backport this to 1.18 and 1.17 as well?

robscott:
@cpanato @bowei Created PRs to backport this to 1.18 and 1.17 as well.

PR #94109, opened 2020-08-19
Cleanup kubelet TODOs that are no longer pertinent.

derekwaynecarr:
**What type of PR is this?**
/kind cleanup

The `kubelet` has grown a number of TODO blocks organically where many of the enumerated tasks are no longer pertinent.  In an attempt to better focus where we should make updates, I am slowly pushing my way through each TODO to see if still relevant.

TODO hygiene to better focus energy moving forward on cleanups.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -138,7 +138,7 @@ const (
 	housekeepingPeriod = time.Second * 2
 
 	// Period for performing eviction monitoring.
-	// TODO ensure this is in sync with internal cadvisor housekeeping.
+	// ensure this is kept in sync with internal cadvisor housekeeping.
 	evictionMonitoringPeriod = time.Second * 10
 
 	// The path in containers' filesystems where the hosts file is mounted.
@@ -440,9 +440,7 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 	}
 	nodeLister := corelisters.NewNodeLister(nodeIndexer)
 
-	// TODO: get the real node object of ourself,
-	// and use the real node name and UID.
-	// TODO: what is namespace for node?
+	// construct a node reference used for events
 	nodeRef := &v1.ObjectReference{
 		Kind:      "Node",
 		Name:      string(nodeName),
@@ -948,8 +946,6 @@ type Kubelet struct {
 	streamingRuntime kubecontainer.StreamingRuntime
 
 	// Container runtime service (needed by container runtime Start()).
-	// TODO(CD): try to make this available without holding a reference in this
-	//           struct. For example, by adding a getter to generic runtime.
 	runtimeService internalapi.RuntimeService
 
 	// reasonCache caches the failure reason of the last creation of all containers, which is
@@ -1069,7 +1065,6 @@ type Kubelet struct {
 	// maintains Node.Spec.Unschedulable value from previous run of tryUpdateNodeStatus()
 	lastNodeUnschedulable bool
 
-	// TODO: think about moving this to be centralized in PodWorkers in follow-on.
 	// the list of handlers to call during pod admission.
 	admitHandlers lifecycle.PodAdmitHandlers
 
@@ -1275,7 +1270,6 @@ func (kl *Kubelet) initializeModules() error {
 func (kl *Kubelet) initializeRuntimeDependentModules() {
 	if err := kl.cadvisor.Start(); err != nil {
 		// Fail kubelet and rely on the babysitter to retry starting kubelet.
-		// TODO(random-liu): Add backoff logic in the babysitter
 		klog.Fatalf("Failed to start cAdvisor %v", err)
 	}
 
@@ -1680,7 +1674,6 @@ func (kl *Kubelet) deletePod(pod *v1.Pod) error {
 	podPair := kubecontainer.PodPair{APIPod: pod, RunningPod: &runningPod}
 
 	kl.podKiller.KillPod(&podPair)
-	// TODO: delete the mirror pod here?
 
 	// We leave the volume/directory cleanup to the periodic cleanup routine.
 	return nil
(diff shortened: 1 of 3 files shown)


pkg/kubelet/kubelet.go
```diff
@@ -138,7 +138,7 @@ const (
 	housekeepingPeriod = time.Second * 2
 
 	// Period for performing eviction monitoring.
-	// TODO ensure this is in sync with internal cadvisor housekeeping.
+	// ensure this is kept in sync with internal cadvisor housekeeping.
```
derekwaynecarr, line 141: note: this is in sync with cadvisor housekeeping, so no action is required.

pkg/kubelet/kubelet.go
```diff
@@ -440,9 +440,7 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 	}
 	nodeLister := corelisters.NewNodeLister(nodeIndexer)
 
-	// TODO: get the real node object of ourself,
-	// and use the real node name and UID.
-	// TODO: what is namespace for node?
+	// construct a node reference used for events
```
derekwaynecarr, line 443: this code is so old, it speculated that a node would have a namespace ;-)

pkg/kubelet/kubelet.go
```diff
@@ -1275,7 +1270,6 @@ func (kl *Kubelet) initializeModules() error {
 func (kl *Kubelet) initializeRuntimeDependentModules() {
 	if err := kl.cadvisor.Start(); err != nil {
 		// Fail kubelet and rely on the babysitter to retry starting kubelet.
-		// TODO(random-liu): Add backoff logic in the babysitter
```
derekwaynecarr, line 1278: note: no action needed in kubelet, process monitor is not a concept in k/k repositories.

pkg/kubelet/kubelet.go
```diff
@@ -1680,7 +1674,6 @@ func (kl *Kubelet) deletePod(pod *v1.Pod) error {
 	podPair := kubecontainer.PodPair{APIPod: pod, RunningPod: &runningPod}
 
 	kl.podKiller.KillPod(&podPair)
-	// TODO: delete the mirror pod here?
```
derekwaynecarr, line 1683: we cleaned up mirror pod deletion flow (including grace period) via: https://github.com/kubernetes/kubernetes/pull/92442
https://github.com/kubernetes/kubernetes/pull/93307

pkg/kubelet/kubelet.go
```diff
@@ -2003,8 +1996,6 @@ func (kl *Kubelet) HandlePodUpdates(pods []*v1.Pod) {
 			kl.handleMirrorPod(pod, start)
 			continue
 		}
-		// TODO: Evaluate if we need to validate and reject updates.
```
derekwaynecarr, line 2006: pods are validated already.

pkg/kubelet/kubelet.go
```diff
@@ -2093,7 +2084,6 @@ func (kl *Kubelet) updateRuntimeUp() {
 		return
 	}
 	// Periodically log the whole runtime status for debugging.
-	// TODO(random-liu): Consider to send node event when optional
```
derekwaynecarr, line 2096: we should not promote more event spam.

pkg/kubelet/kubelet_node_status.go
```diff
@@ -524,8 +524,6 @@ func (kl *Kubelet) tryUpdateNodeStatus(tryNumber int) error {
 // message for the node.
 func (kl *Kubelet) recordNodeStatusEvent(eventType, event string) {
 	klog.V(2).Infof("Recording %s event message for node %s", event, kl.nodeName)
-	// TODO: This requires a transaction, either both node status is updated
-	// and event is recorded or neither should happen, see issue #6055.
```
derekwaynecarr, line 528: there is no desire to ever fix the referenced issue to align a status update with another transaction in apiserver.

pkg/kubelet/kubelet_pods.go
```diff
@@ -836,7 +836,6 @@ func containerResourceRuntimeValue(fs *v1.ResourceFieldSelector, pod *v1.Pod, co
 }
 
 // One of the following arguments must be non-nil: runningPod, status.
-// TODO: Modify containerRuntime.KillPod() to accept the right arguments.
```
derekwaynecarr, line 839: determining the right pod data to pass in kill pod is right here, and kill pod takes the right arguments to handle issues like graceperiod already.

pkg/kubelet/kubelet_pods.go
```diff
@@ -1087,7 +1086,6 @@ func (kl *Kubelet) HandlePodCleanups() error {
 		desiredPods[pod.UID] = sets.Empty{}
 	}
 	// Stop the workers for no-longer existing pods.
-	// TODO: is here the best place to forget pod workers?
```
derekwaynecarr, line 1090: this is where we centralize cleanup for all non-desired pods, so there isnt really a better place to do this now.

derekwaynecarr:
@dashpole @dchen1107 @sjenning going through kubelet code to find legacy TODOs that confuse so we can focus on the right TODOs moving fwd.  I put a few comments in the PR why each was no longer relevant.

mattjmcnaughton (approved):
Thanks for doing this @derekwaynecarr ! Really helpful :)

pkg/kubelet/kubelet.go
```diff
@@ -2093,7 +2084,6 @@ func (kl *Kubelet) updateRuntimeUp() {
...
 	// condition is unmet.
```
SergeyKanzhelev, line 2087: remove this line as well?

pkg/kubelet/kubelet.go
```diff
@@ -2093,7 +2084,6 @@ func (kl *Kubelet) updateRuntimeUp() {
...
 	klog.V(4).Infof("Container runtime status: %v", s)
```
SergeyKanzhelev, line 2088:
```suggestion
	klog.V(4).Infof("Container runtime status: %v", s)
```

pkg/kubelet/kubelet.go
dashpole: actually, we now do eviction with on-demand metrics, so it doesn't matter how it relates to the housekeeping interval.

SergeyKanzhelev:
I will lgtm this and will remove leftover line from the comment as a separate PR. Just to move it forward

/lgtm

PR #94110, opened 2020-08-19
fix e2e test "should resolve connection reset"

aojea:
the test is not working because is trying to execute bash in a
busybox image, that is not present. Using sh works.

/kind failing-test

```release-note
NONE
```


aojea:
the test right now is broken, the client never works, but it's always passing because of that, and is confusing. 
 I'd love to have it in 1.19 if possible.

MrHohn (approved):
/lgtm
/approve

PR #5939, opened 2015-03-25
Implement multi-port endpoints

thockin:
Another piece of multi-port services.

Instead of endpoints being a flat list, it is now a list of "subsets"
where each is a struct of {Addresses, Ports}.  To generate the list of
endpoints you need to take union of the Cartesian products of the
subsets.  This is compact in the vast majority of cases, yest still
represents named ports and corner cases (e.g. each pod has a different
port number).

This also stores subsets in a deterministic order (sorted by hash) to
avoid spurious updates and comparison problems.

Also closes #5895

PR #94112, opened 2020-08-19
Remove canonicalization of endpoints by endpoints controller for better comparison

damemi:
**What type of PR is this?**
/kind bug

The EndpointsController compares labels and subsets to determine if an update is needed. It sorts the calculated subsets on a service but does not sort the subsets it compares to. 

In addition, we compare the *Endpoint's* labels to the *Service's* labels, and while the Endpoint may have a `headless` label, the service does not and also leads to false no-op update calls.

This leads to *severe* performance issues in large clusters, with unimportant updates being called on as many as 2/3 (or more) of the services in the cluster at every informer resync. On clusters with thousands of services, this causes significant degradation.

From what we can tell, this issue exists at least as far back as 1.11 (likely earlier) and should be backported.

Before change:
```
$ oc get services --all-namespaces | wc -l
      70
$ oc logs pod/kube-controller-manager-ci-ln-j9ksx82-f76d1-x65w5-master-1 -n openshift-kube-controller-manager -c kube-controller-manager | grep endpoints_controller | grep -c "About to update endpoints"
64

$ oc logs pod/kube-controller-manager-ci-ln-j9ksx82-f76d1-x65w5-master-1 -n openshift-kube-controller-manager -c kube-controller-manager | grep endpoints_controller | grep -c "Update endpoints for"
22
```

These numbers correlate to how many total services are being checked vs how many are being updated. As you can see, about 1/3 of the services in the cluster are being updated without any changes happening.

After change:

```
$ oc logs pod/kube-controller-manager-ci-ln-5kyrnwk-f76d1-5qbkz-master-2 -n openshift-kube-controller-manager -c kube-controller-manager | grep endpoints_controller | grep -c "About to update"

64
$ oc logs pod/kube-controller-manager-ci-ln-ihf6wv2-f76d1-ch2s6-master-1 -n openshift-kube-controller-manager -c kube-controller-manager | grep endpoints_controller | grep -c "Update endpoints for"
1
```

Fixes #

```release-note
Fixed a bug where improper storage and comparison of endpoints led to excessive API traffic from the endpoints controller
```

```docs

```

### DIFF
--- a/pkg/controller/endpoint/endpoints_controller.go
+++ b/pkg/controller/endpoint/endpoints_controller.go
@@ -472,9 +472,18 @@ func (e *EndpointController) syncService(key string) error {
 
 	createEndpoints := len(currentEndpoints.ResourceVersion) == 0
 
+	// Compare the sorted subsets and labels
+	// Remove the HeadlessService label from the endpoints if it exists,
+	// as this won't be set on the service itself
+	// and will cause a false negative in this diff check.
+	// But first check if it has that label to avoid expensive copies.
+	compareLabels := currentEndpoints.Labels
+	if _, ok := currentEndpoints.Labels[v1.IsHeadlessService]; ok {
+		compareLabels = utillabels.CloneAndRemoveLabel(currentEndpoints.Labels, v1.IsHeadlessService)
+	}
 	if !createEndpoints &&
 		apiequality.Semantic.DeepEqual(currentEndpoints.Subsets, subsets) &&
-		apiequality.Semantic.DeepEqual(currentEndpoints.Labels, service.Labels) {
+		apiequality.Semantic.DeepEqual(compareLabels, service.Labels) {
 		klog.V(5).Infof("endpoints are equal for %s/%s, skipping update", service.Namespace, service.Name)
 		return nil
 	}
--- a/pkg/registry/core/endpoint/strategy.go
+++ b/pkg/registry/core/endpoint/strategy.go
@@ -22,7 +22,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/apiserver/pkg/storage/names"
-	endptspkg "k8s.io/kubernetes/pkg/api/endpoints"
 	"k8s.io/kubernetes/pkg/api/legacyscheme"
 	api "k8s.io/kubernetes/pkg/apis/core"
 	"k8s.io/kubernetes/pkg/apis/core/validation"
@@ -60,8 +59,6 @@ func (endpointsStrategy) Validate(ctx context.Context, obj runtime.Object) field
 
 // Canonicalize normalizes the object after validation.
 func (endpointsStrategy) Canonicalize(obj runtime.Object) {
-	endpoints := obj.(*api.Endpoints)
-	endpoints.Subsets = endptspkg.RepackSubsets(endpoints.Subsets)
 }
 
 // AllowCreateOnUpdate is true for endpoints.
--- a/test/integration/endpoints/endpoints_test.go
+++ b/test/integration/endpoints/endpoints_test.go
@@ -1,0 +1,180 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package endpoints
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
+	clientset "k8s.io/client-go/kubernetes"
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/kubernetes/pkg/controller/endpoint"
+	"k8s.io/kubernetes/test/integration/framework"
+)
+
+func TestEndpointUpdates(t *testing.T) {
+	masterConfig := framework.NewIntegrationTestMasterConfig()
+	_, server, closeFn := framework.RunAMaster(masterConfig)
+	defer closeFn()
+
+	config := restclient.Config{Host: server.URL}
+	client, err := clientset.NewForConfig(&config)
+	if err != nil {
+		t.Fatalf("Error creating clientset: %v", err)
+	}
+
+	informers := informers.NewSharedInformerFactory(client, 0)
+
+	epController := endpoint.NewEndpointController(
+		informers.Core().V1().Pods(),
+		informers.Core().V1().Services(),
+		informers.Core().V1().Endpoints(),
+		client,
+		0)
+
+	// Start informer and controllers
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+	informers.Start(stopCh)
+	go epController.Run(1, stopCh)
+
+	// Create namespace
+	ns := framework.CreateTestingNamespace("test-endpoints-updates", server, t)
+	defer framework.DeleteTestingNamespace(ns, server, t)
+
+	// Create a pod with labels
+	pod := &v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-pod",
+			Namespace: ns.Name,
+			Labels:    labelMap(),
+		},
+		Spec: v1.PodSpec{
+			NodeName: "fakenode",
+			Containers: []v1.Container{
+				{
+					Name:  "fake-name",
+					Image: "fakeimage",
+				},
+			},
+		},
+	}
+
+	createdPod, err := client.CoreV1().Pods(ns.Name).Create(context.TODO(), pod, metav1.CreateOptions{})
+	if err != nil {
+		t.Fatalf("Failed to create pod %s: %v", pod.Name, err)
+	}
+
+	// Set pod IPs
+	createdPod.Status = v1.PodStatus{
+		Phase:  v1.PodRunning,
+		PodIPs: []v1.PodIP{{IP: "1.1.1.1"}, {IP: "2001:db8::"}},
+	}
+	_, err = client.CoreV1().Pods(ns.Name).UpdateStatus(context.TODO(), createdPod, metav1.UpdateOptions{})
+	if err != nil {
+		t.Fatalf("Failed to update status of pod %s: %v", pod.Name, err)
+	}
+
+	// Create a service associated to the pod
+	svc := newService(ns.Name, "foo1")
+	svc1, err := client.CoreV1().Services(ns.Name).Create(context.TODO(), svc, metav1.CreateOptions{})
+	if err != nil {
+		t.Fatalf("Failed to create service %s: %v", svc.Name, err)
+	}
+
+	// Obtain ResourceVersion of the new endpoint created
+	var resVersion string
+	if err := wait.PollImmediate(1*time.Second, wait.ForeverTestTimeout, func() (bool, error) {
+		endpoints, err := client.CoreV1().Endpoints(ns.Name).Get(context.TODO(), svc.Name, metav1.GetOptions{})
+		if err != nil {
+			t.Logf("error fetching endpoints: %v", err)
+			return false, nil
+		}
+		resVersion = endpoints.ObjectMeta.ResourceVersion
+		return true, nil
+	}); err != nil {
+		t.Fatalf("endpoints not found: %v", err)
+	}
+
+	// Force recomputation on the endpoint controller
+	svc1.SetAnnotations(map[string]string{"foo": "bar"})
+	_, err = client.CoreV1().Services(ns.Name).Update(context.TODO(), svc1, metav1.UpdateOptions{})
+	if err != nil {
+		t.Fatalf("Failed to update service %s: %v", svc1.Name, err)
+	}
+
+	// Create a new service and wait until it has been processed,
+	// this way we can be sure that the endpoint for the original service
+	// was recomputed before asserting, since we only have 1 worker
+	// in the endpoint controller
+	svc2 := newService(ns.Name, "foo2")
+	_, err = client.CoreV1().Services(ns.Name).Create(context.TODO(), svc2, metav1.CreateOptions{})
+	if err != nil {
+		t.Fatalf("Failed to create service %s: %v", svc.Name, err)
+	}
+
+	if err := wait.PollImmediate(1*time.Second, wait.ForeverTestTimeout, func() (bool, error) {
+		_, err := client.CoreV1().Endpoints(ns.Name).Get(context.TODO(), svc2.Name, metav1.GetOptions{})
+		if err != nil {
+			t.Logf("error fetching endpoints: %v", err)
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		t.Fatalf("endpoints not found: %v", err)
+	}
+
+	// the endpoint controller should not update the endpoint created for the original
+	// service since nothing has changed, the resource version has to be the same
+	endpoints, err := client.CoreV1().Endpoints(ns.Name).Get(context.TODO(), svc.Name, metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("error fetching endpoints: %v", err)
+	}
+	if resVersion != endpoints.ObjectMeta.ResourceVersion {
+		t.Fatalf("endpoints resource version does not match, expected %s received %s", resVersion, endpoints.ObjectMeta.ResourceVersion)
+	}
+
+}
+
+func labelMap() map[string]string {
+	return map[string]string{"foo": "bar"}
+}
+
+// newService returns a service with selector and exposing ports
+func newService(namespace, name string) *v1.Service {
+	return &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    labelMap(),
+		},
+		Spec: v1.ServiceSpec{
+			Selector: labelMap(),
+			Ports: []v1.ServicePort{
+				{Name: "port-1338", Port: 1338, Protocol: "TCP", TargetPort: intstr.FromInt(1338)},
+				{Name: "port-1337", Port: 1337, Protocol: "TCP", TargetPort: intstr.FromInt(1337)},
+			},
+		},
+	}
+
+}
--- a/pkg/controller/endpoint/endpoints_controller_test.go
+++ b/pkg/controller/endpoint/endpoints_controller_test.go
@@ -383,6 +383,33 @@ func TestSyncEndpointsProtocolTCP(t *testing.T) {
 	endpointsHandler.ValidateRequest(t, "/api/v1/namespaces/"+ns+"/endpoints/foo", "PUT", &data)
 }
 
+func TestSyncEndpointsHeadlessServiceLabel(t *testing.T) {
+	ns := metav1.NamespaceDefault
+	testServer, endpointsHandler := makeTestServer(t, ns)
+	defer testServer.Close()
+	endpoints := newController(testServer.URL, 0*time.Second)
+	endpoints.endpointsStore.Add(&v1.Endpoints{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            "foo",
+			Namespace:       ns,
+			ResourceVersion: "1",
+			Labels: map[string]string{
+				v1.IsHeadlessService: "",
+			},
+		},
+		Subsets: []v1.EndpointSubset{},
+	})
+	endpoints.serviceStore.Add(&v1.Service{
+		ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: ns},
+		Spec: v1.ServiceSpec{
+			Selector: map[string]string{"foo": "bar"},
+			Ports:    []v1.ServicePort{{Port: 80}},
+		},
+	})
+	endpoints.syncService(ns + "/foo")
+	endpointsHandler.ValidateRequestCount(t, 0)
+}
+
 func TestSyncEndpointsProtocolUDP(t *testing.T) {
 	ns := "other"
 	testServer, endpointsHandler := makeTestServer(t, ns)
--- a/pkg/registry/core/endpoint/BUILD
+++ b/pkg/registry/core/endpoint/BUILD
@@ -13,7 +13,6 @@ go_library(
     ],
     importpath = "k8s.io/kubernetes/pkg/registry/core/endpoint",
     deps = [
-        "//pkg/api/endpoints:go_default_library",
         "//pkg/api/legacyscheme:go_default_library",
         "//pkg/apis/core:go_default_library",
         "//pkg/apis/core/validation:go_default_library",
--- a/test/integration/BUILD
+++ b/test/integration/BUILD
@@ -50,6 +50,7 @@ filegroup(
         "//test/integration/deployment:all-srcs",
         "//test/integration/disruption:all-srcs",
         "//test/integration/dryrun:all-srcs",
+        "//test/integration/endpoints:all-srcs",
         "//test/integration/endpointslice:all-srcs",
         "//test/integration/etcd:all-srcs",
         "//test/integration/events:all-srcs",
--- a/test/integration/endpoints/BUILD
+++ b/test/integration/endpoints/BUILD
@@ -1,0 +1,35 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_test")
+
+go_test(
+    name = "go_default_test",
+    srcs = [
+        "endpoints_test.go",
+        "main_test.go",
+    ],
+    tags = ["integration"],
+    deps = [
+        "//pkg/controller/endpoint:go_default_library",
+        "//staging/src/k8s.io/api/core/v1:go_default_library",
+        "//staging/src/k8s.io/apimachinery/pkg/apis/meta/v1:go_default_library",
+        "//staging/src/k8s.io/apimachinery/pkg/util/intstr:go_default_library",
+        "//staging/src/k8s.io/apimachinery/pkg/util/wait:go_default_library",
+        "//staging/src/k8s.io/client-go/informers:go_default_library",
+        "//staging/src/k8s.io/client-go/kubernetes:go_default_library",
+        "//staging/src/k8s.io/client-go/rest:go_default_library",
+        "//test/integration/framework:go_default_library",
+    ],
+)
+
+filegroup(
+    name = "package-srcs",
+    srcs = glob(["**"]),
+    tags = ["automanaged"],
+    visibility = ["//visibility:private"],
+)
+
+filegroup(
+    name = "all-srcs",
+    srcs = [":package-srcs"],
+    tags = ["automanaged"],
+    visibility = ["//visibility:public"],
+)
--- a/test/integration/endpoints/main_test.go
+++ b/test/integration/endpoints/main_test.go
@@ -1,0 +1,27 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package endpoints
+
+import (
+	"testing"
+
+	"k8s.io/kubernetes/test/integration/framework"
+)
+
+func TestMain(m *testing.M) {
+	framework.EtcdMain(m.Run)
+}


marun:
@damemi Nice work in diagnosing this! 

Would it make sense to add test coverage to both validate the fix and ensure against regression? That this behavior was latent for so long (since at least 1.11?) suggests that there is a gap in the existing test coverage.

pkg/controller/endpoint/endpoints_controller.go
```diff
@@ -472,9 +472,12 @@ func (e *EndpointController) syncService(key string) error {
 
 	createEndpoints := len(currentEndpoints.ResourceVersion) == 0
 
+	// Compare the sorted subsets and labels
+	// Remove the HeadlessService label from the endpoints if it exists, as this won't be set on the service itself
+	// and will cause a false negative in this diff check
 	if !createEndpoints &&
-		apiequality.Semantic.DeepEqual(currentEndpoints.Subsets, subsets) &&
-		apiequality.Semantic.DeepEqual(currentEndpoints.Labels, service.Labels) {
+		apiequality.Semantic.DeepEqual(endpoints.RepackSubsets(currentEndpoints.Subsets), subsets) &&
```
wojtek-t, line 479: This shouldn't be needed - endpoints controller is creating endpoints with already sorted subsets.

So as long, as the endpoint has been created by endpoints controller, this isn't needed. If this wasn't then that sounds like two things fighting with each other, so it's also a problem.

So I think this isn't needed.

pkg/controller/endpoint/endpoints_controller.go
```diff
@@ -472,9 +472,12 @@ func (e *EndpointController) syncService(key string) error {
...
+		apiequality.Semantic.DeepEqual(utillabels.CloneAndRemoveLabel(currentEndpoints.Labels, v1.IsHeadlessService), service.Labels) {
```
wojtek-t, line 480: This is nice catch - please add a test that will ensure that we won't regress this again in the future.

pkg/controller/endpoint/endpoints_controller.go
soltysh: Theoretically I'd agree, but the test clusters were we experienced it proof otherwise. We don't have any external actors modifying endpoints, yet the comparison here causes excessive updates to all endpoints. Not sure why @damemi went with `RepackSubsets` since we initially discussed only the necessity to sort them. I'll catch up with Mike when he's back online. This is especially visible on bigger clusters with +10k, but also reproducible on smaller ones.

wojtek-t: > We don't have any external actors modifying endpoints, yet the comparison here causes excessive updates to all endpoints.

Hmm - so I think that means we have some other bug somewhere else. So instead of this mitigation, I would say that we should try understand why is this happening...

damemi: Are we sure it's sorting endpoints before creating/updating them? And are they sorted when they return from the `endpointsLister` call above? (https://github.com/kubernetes/kubernetes/pull/94112/files#diff-a1a9c0efe93384ed9a010e65e8ad4604L459)

wojtek-t: @robscott @freehan to confirm, but it seems we're sorting them before creating/updated.

[and obviously endpointsLister is returning whatever we committed to etcd earlier via creations/updates]

pkg/controller/endpoint/endpoints_controller.go
damemi: pushed to add a test for this

pkg/controller/endpoint/endpoints_controller.go
wojtek-t: @soltysh @damemi - thanks for the data

Hmm - this looks super strange...
I looked into the code a bit more, and we are sorting ports by hash when repacking in both places:
https://github.com/kubernetes/kubernetes/blob/14a11060a0775ed609f0810898ebdbe737c59441/pkg/api/v1/endpoints/util.go#L197
https://github.com/kubernetes/kubernetes/blob/14a11060a0775ed609f0810898ebdbe737c59441/pkg/api//endpoints/util.go#L196
[ideally, we should try to combine these two functions, as there is high-probability of divergence here - @robscott ]

So that looks fine.
Neither of our client machinery (informers, listers, etc.) should NOT modify the objects, these are supposed to only cache them and not touch anything.
Deserialization code looks fine, so the deserialized object should actually preserve the order of port.

> Having said that, I think it's most reliable (and efficient too) if our controllers do not rely on a specific order, as long as data is correct. @wojtek-t wdyt?

I'm slightly afraid that this would mask some real issue somewhere in the system. I would much rather prefer to track it down first. @liggitt - WDYT?

liggitt: > Hmm - so I think that means we have some other bug somewhere else. So instead of this mitigation, I would say that we should try understand why is this happening...

I agree with this. We need to root cause the issue rather than compensating for it.

pkg/controller/endpoint/endpoints_controller.go
liggitt: should we only CloneAndRemoveLabel if the headless label is present to avoid expensive map copies?

pkg/controller/endpoint/endpoints_controller.go
robscott: Thanks for identifying this issue @damemi! It seems like we really need to understand what is causing the change in Endpoints order. Have you been able to tell if any Endpoints updates are happening that cause the change in port ordering? This does feel like a lower level issue than the controller itself, so would be good to dig a bit further to understand what's happening. You've mentioned that this is most noticeable in larger clusters, is there a way to consistently reproduce this behavior at smaller scales?

pkg/controller/endpoint/endpoints_controller.go
soltysh: Good points :+1:

pkg/controller/endpoint/endpoints_controller.go
soltysh: I spoke shortly with Jordan about it on slack, I'll try have another look into it, focusing on what might be causing this mutation.

liggitt: I'm looking at things that could explain a difference between the server canonicalized order and the controller-computed order:

* Are there any non-nil empty lists in the object when the controller packs subsets? If so, that could change the computed hash in https://github.com/kubernetes/kubernetes/blob/release-1.18/pkg/api/endpoints/util.go#L198, then those non-nil empty lists would be nil when the server observed them (because protobuf serialization does not distinguish between nil and non-nil empty lists), and the server could repack in a different order
* I see a lot of uses of `sort.Sort` by RepackSubsets, which does not guarantee stable sort order for items whose sort keys are identical (e.g. endpoint addresses which share an IP and targetRef.UID) ... do identical items like that exist in the cases you are seeing?
* Is the controller sending an object with empty fields that will have defaulting applied to them (e.g. https://github.com/kubernetes/kubernetes/blob/release-1.18/pkg/apis/core/v1/defaults.go#L321-L331)? That seems like it could influence the hash and order

aojea: I'm able to reproduce it using the integration test framework https://github.com/kubernetes/kubernetes/pull/94277
The problem seems to be with the endpoint strategy
```
TEST endpoint strategy [{[] [{1.1.1.1  0xc00566ba50 0xc000556380}] [{port-1337 1337 TCP <nil>} {port-1338 1338 TCP <nil>}]}]
TEST endpoint strategy after [{[] [{1.1.1.1  0xc00566ba50 0xc000556380}] [{port-1338 1338 TCP <nil>} {port-1337 1337 TCP <nil>}]}]
```

seems @wojtek-t was right https://github.com/kubernetes/kubernetes/pull/94112/files#r475439845

aojea: if I skip `Canonicalize()` then it works, 
```
I0827 13:36:28.154446  192183 endpoints_controller.go:475] TEST endpoints are equal?
 current subsets: [{[] [{1.1.1.1  0xc004ec7150 &ObjectReference{Kind:Pod,Namespace:test-endpoints-updates,Name:test-pod,UID:3082f47b-4a7b-4b69-bae8-56cee3fdf0ba,APIVersion:,ResourceVersion:124,FieldPath:,}}] [{port-1337 1337 TCP <nil>} {port-1338 1338 TCP <nil>}]}]
 subsets [{[] [{1.1.1.1  0xc004bfa1d8 &ObjectReference{Kind:Pod,Namespace:test-endpoints-updates,Name:test-pod,UID:3082f47b-4a7b-4b69-bae8-56cee3fdf0ba,APIVersion:,ResourceVersion:124,FieldPath:,}}] [{port-1337 1337 TCP <nil>} {port-1338 1338 TCP <nil>}]}]
I0827 13:36:28.154559  192183 endpoints_controller.go:480] endpoints are equal for test-endpoints-updates/foo, skipping update
```

liggitt: > I'm able to reproduce it using the integration test framework #94277

excellent... that will make tracking down the issue much easier

> if I skip `Canonicalize()` then it works

can you dump the JSON serialization of the endpoints object in the following places:
* before the controller calls repack
* after the controller calls repack
* before Canonicalize calls repack
* after Canonicalize calls repack

liggitt: I also noticed few additional things:
* we've had a v1 and internal version of RepackSubsets since bff3cb4df68d699285285c09c1c46be69ef1465a, which is problematic if they drift ... we should verify they still match and put fuzzing tests in place to make sure the equivalent v1/internal versions pack identically
* anything using `hashObject` is vulnerable to skew between the controller and API server when a new field (like AppProtocol) is added to one of the types... the newer component will hash based on the new field with a nil value, and the older component will omit that from the hash
  * any changes to RepackSubsets/SortSubsets need to be done *extremely* carefully, and in a way that does not trigger this issue for all endpoints when in a skew state. Our lack of skew testing makes us especially vulnerable here.

liggitt: looks like there's a diff between the v1 and internal implementations:

```
import (
	"reflect"
	"testing"

	"github.com/davecgh/go-spew/spew"
	"github.com/google/go-cmp/cmp"
	v1 "k8s.io/api/core/v1"
	v1util "k8s.io/kubernetes/pkg/api/v1/endpoints"
	apiv1 "k8s.io/kubernetes/pkg/apis/core/v1"
	"k8s.io/utils/pointer"
)

func TestSortSubsetsCompatibility(t *testing.T) {
	// Start with v1 endpoints
	v1Endpoints := &v1.Endpoints{
		Subsets: []v1.EndpointSubset{{
			NotReadyAddresses: []v1.EndpointAddress{{
				IP:       "1.1.1.1",
				NodeName: pointer.StringPtr("fakenode"),
				TargetRef: &v1.ObjectReference{
					Kind:            "Pod",
					Namespace:       "test-endpoints-updates",
					Name:            "test-pod",
					UID:             "277ecc82-f0a9-4d57-92da-5966466d1afc",
					ResourceVersion: "126828",
				},
			}},
			Ports: []v1.EndpointPort{
				{Name: "port-1337", Port: 1337, Protocol: "TCP"},
				{Name: "port-1338", Port: 1338, Protocol: "TCP"},
			},
		}},
	}

	// Pack with v1 util, matching what the controller will do
	v1Endpoints.Subsets = v1util.RepackSubsets(v1Endpoints.Subsets)

	// Convert to internal
	internalEndpoints := &api.Endpoints{}
	if err := apiv1.Convert_v1_Endpoints_To_core_Endpoints(v1Endpoints.DeepCopy(), internalEndpoints, nil); err != nil {
		t.Fatal(err)
	}

	// Save a copy to compare
	beforeInternal := internalEndpoints.DeepCopy()

	// Pack with internal util, matching with the API server will do
	internalEndpoints.Subsets = RepackSubsets(internalEndpoints.Subsets)
	// Should be no diff
	if diff := cmp.Diff(beforeInternal.Subsets, internalEndpoints.Subsets); len(diff) > 0 {
		t.Error(diff)
	}
}
```

```sh
go test ./pkg/api/endpoints -run TestSortSubsetsCompatibility
--- FAIL: TestSortSubsetsCompatibility (0.00s)
    util_test.go:71:   []core.EndpointSubset{
          	{
          		Addresses:         nil,
          		NotReadyAddresses: []core.EndpointAddress{{IP: "1.1.1.1", NodeName: &"fakenode", TargetRef: &core.ObjectReference{Kind: "Pod", Namespace: "test-endpoints-updates", Name: "test-pod", UID: "277ecc82-f0a9-4d57-92da-5966466d1afc", ResourceVersion: "126828"}}},
          		Ports: []core.EndpointPort{
          			{
        - 				Name:        "port-1337",
        + 				Name:        "port-1338",
        - 				Port:        1337,
        + 				Port:        1338,
          				Protocol:    "TCP",
          				AppProtocol: nil,
          			},
          			{
        - 				Name:        "port-1338",
        + 				Name:        "port-1337",
        - 				Port:        1338,
        + 				Port:        1337,
          				Protocol:    "TCP",
          				AppProtocol: nil,
          			},
          		},
          	},
          }
        
FAIL
FAIL	k8s.io/kubernetes/pkg/api/endpoints	0.284s
```

liggitt: One issue is with object hash:

The v1 implementation is hashing this string:
> `([]endpoints.addrReady)[{addr:(*v1.EndpointAddress){IP:(string)1.1.1.1 Hostname:(string) NodeName:(*string)fakenode TargetRef:(*v1.ObjectReference){Kind:(string)Pod Namespace:(string)test-endpoints-updates Name:(string)test-pod UID:(types.UID)277ecc82-f0a9-4d57-92da-5966
466d1afc APIVersion:(string) ResourceVersion:(string)126828 FieldPath:(string)}} ready:(bool)false}]`

The internal implementation is hashing this string:
> `([]endpoints.addrReady)[{addr:(*core.EndpointAddress){IP:(string)1.1.1.1 Hostname:(string) NodeName:(*string)fakenode TargetRef:(*core.ObjectReference){Kind:(string)Pod Namespace:(string)test-endpoints-updates Name:(string)test-pod UID:(types.UID)277ecc82-f0a9-4d57-92da-
5966466d1afc APIVersion:(string) ResourceVersion:(string)126828 FieldPath:(string)}} ready:(bool)false}]`

Note the different type names:
* `v1.EndpointAddress` vs `core.EndpointAddress`
* `v1.ObjectReference` vs `core.ObjectReference`

For that reason, and because of the version skew issue noted in https://github.com/kubernetes/kubernetes/pull/94112/files#r478397569, using hashes of the go struct seems really problematic here.

liggitt: So, to summarize, I see the following issues that need to be fixed if we continue to normalize order both in the server and controller and expect them to match perfectly in order to settle:

* hashing go structs differs from internal/external struct type names
* hashing go structs differs when there's skew between the API server version and controller version with new fields
* hashing go structs differs from nil empty slices and non-nil empty slices with protobuf collapsing the difference between those
* not using sort.Stable doesn't guarantee the same sort order for values whose sort keys are identical
* server-side defaulting applied before server normalization can affect hash values and make them differ from the controller-computed order
* independent implementations of normalization for v1 (used by controller) and internal (used by API server) can drift

wojtek-t: > we've had a v1 (used by the controller) and internal (used by the API server) version of RepackSubsets since bff3cb4, which is problematic if they drift .

Yeah - I also mentioned that above in:  https://github.com/kubernetes/kubernetes/pull/94112#discussion_r475439845


BTW - nice list above, thanks Jordan!

soltysh: Sine the biggest issue is around hasing objects in `RepackSubsets` I took two different approaches and here are the results:
1. Remove types in the string representation coming from spew library in [DeepHashObject](https://github.com/kubernetes/kubernetes/blob/72350a3cf7b1b021eb31d72ab98013ebe8455fc4/pkg/util/hash/hash.go#L28).
2. Convert `v1.EndpointSubset` to `core.EndpointSubset`, invoke internal version of `RepackSubsets` and convert back to v1.

Benchmark results of those are:
1. 17737 ns/op
2. 13305 ns/op

as a comparison, the current implementation is running roughly 13222 ns/op. (Those are results from my env.)

The first is  ~30% slower, rather naive approach and fragile to changes. Second one on the other hand gives us similar performance to what we currently have, and most importantly removes code duplication. We could probably provide the both versions and ensure the conversions are hidden from users. 
For second solution I was also considering having a 3rd `RepackSubset` internal representation, but I'm not sure that's something we'd like to pursuit. 

@liggitt @wojtek-t @thockin @robscott opinions?

liggitt: Three of the issues could be addressed by converting/using a single Repack method (and using `sort.Stable` everywhere):

* [x] internal/external struct type names differ, causing hash differences
* [x] independent implementations of normalization for v1 (used by controller) and internal (used by API server) can drift
* [x] not using sort.Stable doesn't guarantee the same sort order for values whose sort keys are identical

These issues would still remain:
* [ ] version skew between the API server and controller with new fields in the structs causes differences
* [ ] nil empty slices and non-nil empty slices (caused by protobuf collapsing the difference between those) causes differences
* [ ] server-side defaulting applied before server normalization can affect hash values and make them differ from the controller-computed order

If we were going to convert/use a single repack method, I would lean toward having the server convert to v1.Endpoints and use the method intended for use with the versioned object.

I'm actually wondering if it makes more sense to drop the server-side canonicalization if re-ordering is the only change being made (I could see keeping it in deduplication cases, which would make it obvious the client was not sending what it thought was a repacked subset).

wojtek-t: > I'm actually wondering if it makes more sense to drop the server-side canonicalization if re-ordering is the only change being made (I could see keeping it in deduplication cases, which would make it obvious the client was not sending what it thought was a repacked subset).

I was suggesting dropping it completely, but what you suggest makes more sense to me.
So basically, if there aren't duplicates, leave as is, if there are duplicates, do repacking.

soltysh: I can definitely support the idea of dropping the server-side canonicalization, but I'm not sure about the history behind it. Maybe @thockin or @robscott might provide them? 

In general endpoints controller is quite frequently going through all of them, so even if somebody posts unordered Endpoints, it will be corrected in a timely fashion.

robscott: I've talked about this with @freehan and neither of us can think of any reason endpoints would need to be consistently sorted. I can not find anything inside kube-proxy that requires this kind of specific ordering. It feels surprising and inconsistent that resources would be modified in this way before they were saved. I can potentially understand the desire to leave deduplication in as part of the strategy, but full repacking at this point seems both unnecessary and unexpected.

seh: I thought the sorting had to do with making it easier to find the relative complement of the old and new sets, in order to be able to update _iptables_ and _ipvs_ rules.

akutz: > Theoretically I'd agree, but the test clusters were we experienced it proof otherwise. We don't have any external actors modifying endpoints, yet the comparison here causes excessive updates to all endpoints. Not sure why @damemi went with `RepackSubsets` since we initially discussed only the necessity to sort them. I'll catch up with Mike when he's back online. This is especially visible on bigger clusters with +10k, but also reproducible on smaller ones. 

We also saw the same thing. Please see the _Additional Info_ section in https://github.com/kubernetes/kubernetes/issues/94339.

I spent just about three weeks debugging this issue (the aforementioned section describes why it took so long), and it turns out we were working on close to the same problem the entire time :)

aojea: @thockin it seems that you've added the repacking of the endpoints in the apiserver in #5939  ?to deal with the migration to multi-port endpoints? 
based on the comments in this thread it seems that there is no reason now to keep doing it. Can you confirm if is ok to _drop the server-side canonicalization_?
If affirmative, @damemi @soltysh  I went ahead and have a commit https://github.com/kubernetes/kubernetes/pull/94277/commits/fba1e3f0b65d9395e06a1bce171c7602a114d264  dropping the canonicalization in the apiserver and  a test case covering, feel free to take and add it to this PR

akutz: I see no harm in the endpoint canonicalization in-and-of-itself. It may still serve a purpose. The issue, IMO, is the conflict with the ServerSideApply described in #94339. Without SSA enabled, this is a non-issue -- the canonicalization works as intended and additional updates are no-ops. It is only once SSA was enabled by default that this became a problem.

Which raises a much larger issue -- should clients be expected to know and reproduce any server-side storage strategies and their data mutation logic to avoid superfluous `MODIFIED` events? My vote would be no, I think this is an unreasonable expectation to place upon users. Since SSA is likely not going anywhere, and storage strategies can and do make sense, shouldn't we try to make the two work together?

liggitt: > I see no harm in the endpoint canonicalization in-and-of-itself. It may still serve a purpose.

Github is eliding the comment, but there are six issues identified with this specific server-side normalization - https://github.com/kubernetes/kubernetes/pull/94112#discussion_r478422983

> The issue, IMO, is the conflict with the ServerSideApply described in #94339.

That is an additional issue/interaction which needs to be resolved, but is orthogonal.

akutz: > there are six issues identified with this specific server-side normalization

ack

> That is an additional issue/interaction which needs to be resolved, but is orthogonal.

ack

damemi:
Sorry for the lack of responses from my side, I was out last week and trying to catch up.

There's been a lot of discussion in this bug, but I don't quite understand if we've identified the underlying issue? There seem to be numerous proposed solutions regarding SSA, canonicalization, hashing, and differing sort/repack functions. That sounds like a lot of different overlapping areas (though I might be misinterpreting the conversation above).

Given that this is a critical bug, which is currently causing severe performance impacts in large-scale production clusters (going back several releases), I would just like to summarize if we have a consensus on how to fix it.

From the discussion above, it sounds like this specific PR *shouldn't* be necessary if the root causes are fixed. Is that correct? If so, what are the specific changes that need to be implemented?

wojtek-t:
> From the discussion above, it sounds like this specific PR shouldn't be necessary if the root causes are fixed. Is that correct? If so, what are the specific changes that need to be implemented?

https://github.com/kubernetes/kubernetes/commit/fba1e3f0b65d9395e06a1bce171c7602a114d264

damemi:
> > From the discussion above, it sounds like this specific PR shouldn't be necessary if the root causes are fixed. Is that correct? If so, what are the specific changes that need to be implemented?
> 
> [fba1e3f](https://github.com/kubernetes/kubernetes/commit/fba1e3f0b65d9395e06a1bce171c7602a114d264)

Thanks @wojtek-t and I see @aojea's comment above now:

>  I went ahead and have a commit fba1e3f in #94277 dropping the canonicalization in the apiserver and a test case covering, feel free to take and add it to this PR

So I'll update this PR with that commit

damemi:
Updated with the remove-canonicalization commit, I also added the check @liggitt suggested in https://github.com/kubernetes/kubernetes/pull/94112#discussion_r475672136 for the headless label.

pkg/controller/endpoint/endpoints_controller.go
```diff
@@ -472,9 +472,16 @@ func (e *EndpointController) syncService(key string) error {
...
+	// Remove the HeadlessService label from the endpoints if it exists, as this won't be set on the service itself
+	// and will cause a false negative in this diff check. But first check if it has that label to avoid expensive copies.
+	compareLabels := currentEndpoints.Labels
+	if _, ok := currentEndpoints.Labels[v1.IsHeadlessService]; ok {
+		compareLabels = utillabels.CloneAndRemoveLabel(currentEndpoints.Labels, v1.IsHeadlessService)
+	}
 	if !createEndpoints &&
-		apiequality.Semantic.DeepEqual(currentEndpoints.Subsets, subsets) &&
-		apiequality.Semantic.DeepEqual(currentEndpoints.Labels, service.Labels) {
+		apiequality.Semantic.DeepEqual(endpoints.RepackSubsets(currentEndpoints.Subsets), subsets) &&
```
wojtek-t, line 483: We no longer want to repack here.

damemi: Updated, and changed the commit message to reflect that we're just removing the Headless label if necessary now

pkg/controller/endpoint/endpoints_controller.go
```diff
@@ -472,9 +472,16 @@ func (e *EndpointController) syncService(key string) error {
 
 	createEndpoints := len(currentEndpoints.ResourceVersion) == 0
 
+	// Compare the sorted subsets and labels
+	// Remove the HeadlessService label from the endpoints if it exists, as this won't be set on the service itself
```
wojtek-t, line 476: nit: can you please split the lines to make the shorter?

pkg/registry/core/endpoint/strategy.go
```diff
@@ -60,8 +59,6 @@ func (endpointsStrategy) Validate(ctx context.Context, obj runtime.Object) field
 
 // Canonicalize normalizes the object after validation.
 func (endpointsStrategy) Canonicalize(obj runtime.Object) {
-	endpoints := obj.(*api.Endpoints)
-	endpoints.Subsets = endptspkg.RepackSubsets(endpoints.Subsets)
```
wojtek-t, line 64: I'm fine with that I described it already before, but it would be good if @thockin could confirm it.

pkg/controller/endpoint/endpoints_controller.go
damemi: done

pkg/registry/core/endpoint/strategy.go
robscott: We talked about this today as part of sig-network and the general consensus from @thockin and others seemed to be that we should remove this.

wojtek-t: Thanks - resolving then.

test/integration/endpoints/endpoints_test.go
```diff
@@ -0,0 +1,159 @@
...
+	_, server, closeFn := framework.RunAMaster(masterConfig)
+	defer closeFn()
+
+	config := restclient.Config{Host: server.URL}
+	client, err := clientset.NewForConfig(&config)
+	if err != nil {
+		t.Fatalf("Error creating clientset: %v", err)
+	}
+
+	resyncPeriod := 30 * time.Second
```
wojtek-t, line 46: Why do we set resync period? Let's change it to 0 - we generally set it to 0 in ~all controllers that don't depend on external state.

test/integration/endpoints/endpoints_test.go
```diff
@@ -0,0 +1,159 @@
...
+	informers := informers.NewSharedInformerFactory(client, resyncPeriod)
+
+	epController := endpoint.NewEndpointController(
+		informers.Core().V1().Pods(),
+		informers.Core().V1().Services(),
+		informers.Core().V1().Endpoints(),
+		client,
+		1*time.Second)
```
wojtek-t, line 54: Let's set to 0, given 0 is a default - setting to 1 may lead to pushing the updates in the future and mask real issue this is going to test.

test/integration/endpoints/endpoints_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	// Start informer and controllers
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+	informers.Start(stopCh)
+	go epController.Run(5, stopCh)
```
wojtek-t, line 60: Do we really need 5 workers?

test/integration/endpoints/endpoints_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	// Create namespace
+	ns := framework.CreateTestingNamespace("test-endpoints-updates", server, t)
+	defer framework.DeleteTestingNamespace(ns, server, t)
+
+	// Create pod and a service with selector to generate endpoints associated to them
+	pod := &v1.Pod{
+		TypeMeta: metav1.TypeMeta{
```
wojtek-t, line 68: You don't need to set TypeMeta.

test/integration/endpoints/endpoints_test.go
```diff
@@ -0,0 +1,159 @@
...
+		},
+	}
+
+	_, err = client.CoreV1().Pods(ns.Name).Create(context.TODO(), pod, metav1.CreateOptions{})
+	if err != nil {
+		t.Fatalf("Failed to create pod %s: %v", pod.Name, err)
+	}
+
+	// Set pod IPs
+	if err := wait.PollImmediate(3*time.Second, 2*time.Minute, func() (bool, error) {
```
wojtek-t, line 94: In general retries are good practice, but I'm wondering why do we need them here (especially that you e.g. don't retry the creation above).
[There aren't any components with which you can race, also.]

test/integration/endpoints/endpoints_test.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	_, err = client.CoreV1().Services(ns.Name).Create(context.TODO(), svc, metav1.CreateOptions{})
+	if err != nil {
+		t.Fatalf("Failed to create service %s: %v", svc.Name, err)
+	}
+
+	// Obtain new endpoints created and its ResourceVersion
+	var resVersion string
+	if err := wait.PollImmediate(3*time.Second, 2*time.Minute, func() (bool, error) {
```
wojtek-t, line 137: Let's reduce the 3 seconds to something smaller.

Also the 2 minutes to wait.ForeverTestTimeout.

test/integration/endpoints/endpoints_test.go
```diff
@@ -0,0 +1,159 @@
...
+			t.Logf("error fetching endpoints: %v", err)
+			return false, nil
+		}
+		resVersion = endpoints.ObjectMeta.ResourceVersion
+		return true, nil
+	}); err != nil {
+		t.Fatalf("endpoints not found: %v", err)
+	}
+	// wait until the recently created endpoints are processed again during the resync period
+	time.Sleep(2 * resyncPeriod)
```
wojtek-t, line 149: I'm actually not a fan of sleeps in tests. Can we instead force recomputation of that service that shouldn't cause changing of the endpoints?
e.g. by updating the service (say by setting a new annotation or sth like that):
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/endpoint/endpoints_controller.go#L278


And to avoid sleeping and make the test deterministic, you can then create a second service and wait for the endpoint for that to be created. If you have a single worked in the controller (see my comment above), you will have them processed in the right order, so when the second is processed, the first also had been processed. That will make test deterministic and not depend on sleeps.

wojtek-t (commented):
Some comments to make the test deterministic.

test/integration/endpoints/endpoints_test.go
wojtek-t: @aojea - FYI

test/integration/endpoints/endpoints_test.go
aojea: it waits until the pod gets created or the update is sucessful
If the creation fails the test already fails

wojtek-t: Creation of pod is synchronous - so Get() can't return 404 (it may return other errors like 500 or 429, but the same can happen for creation which isn't retried above).

So I don't understand why you retry here and not retry elsewhere.

aojea:
@damemi I've implemented @wojtek-t suggestions here cfc865cc320787d23ad06fb55cf39610faddec48

it now runs in only 6 seconds :)

test/integration/endpoints/endpoints_test.go
aojea: I see, then it was mostly ignorance :sweat_smile: 
what will be the best way? retrying everywhere or only waiting for the endpoints to be generated?

wojtek-t: For integration tests, I think not retrying is good enough (we don't retry pretty much anywhere).

wojtek-t:
> @damemi I've implemented @wojtek-t suggestions here 662c2d0

cool - I added one nit, but other than that LGTM.
And it's great to hear that it's so much faster now - that was my other implicit motivation :)

aojea:
> > @damemi I've implemented @wojtek-t suggestions here [662c2d0](https://github.com/kubernetes/kubernetes/commit/662c2d012010b825f867fbc70280643ac076b32c)
> 
> cool - I added one nit, but other than that LGTM.
> And it's great to hear that it's so much faster now - that was my other implicit motivation :)

https://github.com/kubernetes/kubernetes/pull/94277/commits/e5cbe71f8a40a90b3e11c64c59c56a20b9e5d90c

damemi:
Thanks @aojea and @wojtek-t, I've replaced the previous commit https://github.com/kubernetes/kubernetes/pull/94112/commits/55ef9a8b8e57a3d1f407f5b06ca6d0d3f71d4869 with the updated changes in https://github.com/kubernetes/kubernetes/commit/e5cbe71f8a40a90b3e11c64c59c56a20b9e5d90c

wojtek-t:
/lgtm
/approve

Thanks!

[Once merged we need to cherrypick it back]

liggitt:
> does the failure in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/94112/pull-kubernetes-bazel-test/1301894280050118658 look related or a flake?

exposed by https://github.com/kubernetes/kubernetes/pull/93605, fixed in https://github.com/kubernetes/kubernetes/pull/94529

liggitt:
https://github.com/kubernetes/kubernetes/issues/94545 flake
/retest

ialidzhikov:
> /lgtm
> /approve
> 
> Thanks!
> 
> [Once merged we need to cherrypick it back]

The PR is merged. I would be great if someone can file the cherry-picks.

wojtek-t:
Agree.
@damemi - can you please open cherrypick?

damemi:
> Agree.
> @damemi - can you please open cherrypick?

@wojtek-t absolutely, backports created here:
1.18 - https://github.com/kubernetes/kubernetes/pull/94934
1.17 - https://github.com/kubernetes/kubernetes/pull/94935

Should I go back any further?

liggitt:
1.19 as well, right?

no more 1.16 patch releases are being released

damemi:
@liggitt ah whoops, thought this PR was in 1.19. Opened here: https://github.com/kubernetes/kubernetes/pull/94937

wojtek-t:
@damemi - this PR should have release note (so that the cherrypicks will also pick it up).

damemi:
> @damemi - this PR should have release note (so that the cherrypicks will also pick it up).

@wojtek-t not sure what you mean, I created the cherry-picks manually so just copied release-note/none there. Is there some way I should link them or a specific format that they should reference this PR in their release notes?

damemi:
@wojtek-t I updated this PR and the backports with the release-note `Fixed a bug where improper storage and comparison of endpoints led to excessive API traffic from the endpoints controller`

wojtek-t:
> @wojtek-t not sure what you mean, I created the cherry-picks manually so just copied release-note/none there. Is there some way I should link them or a specific format that they should reference this PR in their release notes?

We generally use `hack/cherrypick_pull.sh` script for cherrypicking and then you have that done automatically I think.

PR #90552, opened 2020-04-28
Separate out cri related code into self contained package

dims:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
Previous attempts to extract cri related code met with stiff resistance. We had PRs earlier to move streaming to staging/ etc. So in this PR we just consider moving code into a package that be used standalone w/o dragging in all of k/k. 

Essentially the `pkg/kubelet/cri` uses only stuff in staging/ and does not use anything else inside k8s.io/kubernetes. We have a new .import-restrictions in that directory to ensure that things stay that way.

two things to note:
- had to move logreduction into component-base (anyone have a better suggestion?)
- copied some of the utils code into remote/ so we will have some duplication.

Also note that i've prototyped this on the containerd/cri side as well:
https://github.com/containerd/cri/pull/1463

If we do this, `containerd/cri` repo vendor/ directory will pull in 15k LOC less than they do today

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #87281, opened 2020-01-16
Respect timeout in ExecInContainer

tedyu:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
As #26895 stated, timeoutSeconds is ignored for probe.

This PR adds code to respect the timeout parameter passed to ExecInContainer.

**Which issue(s) this PR fixes**:
Fixes #26895

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #58925, opened 2018-01-28
Add timeout option for docker's exec operation

louyihua:
**What this PR does / why we need it**:
Currently, docker's exec operation never timeout. This patch introduces a new `ExecOptions` that contains a timeout value into the parameters of `libdocker.StartExec`. When a positive timeout is passed and the exec operation timeouts, the `StartExec` function returns the `libdocker.OperationTimeout` error.

**Release note**:

```release-note
Add timeout option into libdocker.StartExec
```

PR #94115, opened 2020-08-20
kubelet: respect exec probe timeouts

### LINKED ISSUE #94080: exec-type liveness or readiness probes ignore timeout
When defining a liveness or readiness probe of type exec with a timeout, like this

    readinessProbe:
      exec:
        command:
        - "curl"
        - "-f"
        - "localhost:8080/slow?timeInSecs=10"
      initialDelaySeconds: 5
      periodSeconds: 3
      timeoutSeconds: 1

where the response is taking longer than the defined timeoutSeconds, the pod is still considered running (in case of a liveness probe) resp. resp. reaches ready state (in case of a readiness probe).
What I would expect is that the command is aborted after the timeout is reached and that the probe is considered failed.
Instead, what I am observing when I exec into the container is that the readinessProbe command "hangs" for ten seconds, then completes 

    Every 0.2s: ps aux     2020-08-18 11:35:03
    
    PID   USER     TIME  COMMAND
        1 root      0:07 /opt/openjdk-16/bin/java -jar /slow-1.0.0-exec.jar
       50 root      0:00 bash
       62 root      0:00 watch -d -n 0.2 ps aux
     5368 root      0:00 curl -f localhost:8080/slow?timeInSecs=10
     5386 root      0:00 ps aux

and the pod reaches ready state:

    slow-deployment-5bbc7f667d-w2lgr   1/1     Running   0          14m

This was first observed on an EKS cluster (v1.15.11-eks), but I also tried various other Kubernetes versions via minikube, e.g.,  1.16.14, 1.17.11, and 1.18.8, same behavior.

Attached you can find a YAML file (slow-deployment.zip) with a sample deployment, a ZIP with the very simple SpringBoot application (app-src.zip), and a ZIP with a Dockerfile (slow-docker.zip, plus a shell script which runs the same healthcheck).
For convenience, the image is also available on Dockerhub.

Note that the Dockerfile also defines a Docker healthcheck. Running this container on plain Docker gives the expected behavior:

    $ docker ps
    CONTAINER ID        IMAGE        COMMAND                  CREATED         STATUS                   
    e58e15f84fca        slow:1.0.0   "/opt/openjdk-16/bin…"   51 seconds ago  Up 22 seconds (unhealthy)

[app-src.zip](https://github.com/kubernetes/kubernetes/files/5090090/app-src.zip)
[slow-docker.zip](https://github.com/kubernetes/kubernetes/files/5090062/slow-docker.zip)
[slow-deployment.zip](https://github.com/kubernetes/kubernetes/files/5090063/slow-deployment.zip)

andrewsykim:

**What type of PR is this?**

/kind bug

This PR fixes exec timeout issues for both dockershim and containerd. 

When using kubelet dockershim, the exec timeout is not respected which results in timeouts for exec readiness/liveness probes to not be respected as well. This PR ensures the timeout passed into RunInContainer for dockershim is respected. The timeout value passed into RunInContainer is derived from the probe timeout in  the case of readiness/liveness probes. 

For containerd, the prober ignores timeout errors from remote runtime's ExecSync since it expects `utilexec.ExitCodeError` for any failed probes. Any other error ends up being ignored by the prober. This PR updates ExecSync to return `utilexec.ExitCodeError` when the grpc error from CRI is `DeadlineContextExceeded`. 

Fixes https://github.com/kubernetes/kubernetes/issues/94080

```release-note
ACTION REQUIRED: a bug was fixed in kubelet where exec probe timeouts were not respected. Ensure that pods relying on this behavior are updated to correctly handle probe timeouts.

This change in behavior may be unexpected for some clusters and can be disabled by turning off the ExecProbeTimeout feature gate. This gate will be locked and removed in future releases so that exec probe timeouts are always respected.
```

```docs

```


andrewsykim:
I was able to reproduce this on a GKE cluster. 

Deployed a pod where the exec probe takes 10s to pass and the timeout is set to 1. On nodes with existing kubelet, the probe passes since the timeout is not respected. 

I manually updated one of the kubelets with this change and now the timeout is respected and the readiness probe fails. 
```
$ kubectl get po -o wide
NAME                               READY   STATUS    RESTARTS   AGE     IP         NODE                                       NOMINATED NODE   READINESS GATES
slow-deployment-6989cc5777-bc92t   1/1     Running   0          4h26m   10.4.2.5   gke-cluster-1-default-pool-9c83cf52-q55l   <none>           <none>
slow-deployment-6989cc5777-cvhhz   1/1     Running   0          4h26m   10.4.0.7   gke-cluster-1-default-pool-9c83cf52-nchv   <none>           <none>
slow-deployment-6989cc5777-f4vhj   0/1     Running   0          4h26m   10.4.1.6   gke-cluster-1-default-pool-9c83cf52-n6x4   <none>           <none>
```

pkg/kubelet/dockershim/exec.go
```diff
@@ -110,27 +110,31 @@ func (*NativeExecHandler) ExecInContainer(client libdocker.Interface, container
...
 			}
-			break
-		}
 
-		count++
-		if count == 5 {
-			klog.Errorf("Exec session %s in container %s terminated but process still running!", execObj.ID, container.ID)
-			break
+			count++
+			if count == 5 {
```
andrewsykim, line 133: I think this count check should be removed now since we're checking the timeout now, but I'll defer to kubelet maintainers since I don't have context on why this is here originally.

dims: looks like earlier we just had an infinite loop until `InspectExec` can tell us that the process we started exited... the count was added to break out of it in https://github.com/kubernetes/kubernetes/commit/7748a02e37cb2233cfb59ef6d608457d526d7bf2

dims:
@andrewsykim also additional context, see https://github.com/kubernetes/kubernetes/pull/50176

pkg/kubelet/dockershim/exec.go
```diff
@@ -110,27 +110,31 @@ func (*NativeExecHandler) ExecInContainer(client libdocker.Interface, container
...
+				klog.Errorf("Exec session %s in container %s terminated but process still running!", execObj.ID, container.ID)
+				break
```
andrewsykim, line 135: oops, forgot `break` applies for `select` as well, need to fix this.

pkg/kubelet/dockershim/exec.go
andrewsykim: As I think about this more, I think we should remove it since this means the max timeout you can set for an exec probe with dockershim is 10s (5 x 2s tick).

SergeyKanzhelev:
is it an alternative to this PR: https://github.com/kubernetes/kubernetes/pull/58925/?

andrewsykim:
e2e failures look legit, will dig into it

/hold

pkg/kubelet/dockershim/exec.go
```diff
@@ -110,28 +110,32 @@ func (*NativeExecHandler) ExecInContainer(client libdocker.Interface, container
 	}
 
 	ticker := time.NewTicker(2 * time.Second)
+	execTimeout := time.After(timeout)
```
andrewsykim, line 113: need to check for 0 timeout, since exec outside of probes will have 0 timeout

pkg/kubelet/dockershim/exec.go
andrewsykim: Nvm, I think we need to keep this since it's expected behavior for kubelet to exit at some point with a nil error and allow the exec'd command to  continue running.

andrewsykim: That or there should be some default timeout here, but don't want to introduce new changes here.

dchen1107:
@andrewsykim Can you add a regression test for this case? I believe we saw a similar issue with Containerd in the past, but was addressed. Can you double confirm it?

andrewsykim:
> @andrewsykim Can you add a regression test for this case? 

Sure thing! Will try to add one today.

> I believe we saw a similar issue with Containerd in the past, but was addressed. Can you double confirm it?

I originally tried to reproduce this on containerd and I couldn't.

pkg/kubelet/dockershim/exec.go
```diff
@@ -109,29 +109,42 @@ func (*NativeExecHandler) ExecInContainer(client libdocker.Interface, container
...
-		if err2 != nil {
-			return err2
-		}
-		if !inspect.Running {
-			if inspect.ExitCode != 0 {
-				err = &dockerExitError{inspect}
+		select {
+		case <-execTimeout:
+			return fmt.Errorf("timed out waiting for exec commands %v to finish for container %s", cmd, container.ID)
+		default:
```
andrewsykim, line 127: This needs to be `default` instead of `<-ticker.C`, otherwise we delay the initial InspectExec by 2 seconds.

andrewsykim:
@dchen1107 looks like there was a test for this that was being skipped 

https://github.com/kubernetes/kubernetes/blob/ace97738e2824c3852146bba1097607d1101842a/test/e2e/common/container_probe.go#L215-L228

Even though the Docker's remote API doesn't support timeouts natively, seems like we should still respect the probe timeout by inspecting the state of the exec command? I've updated the PR to stop skipping to see if the test passes now.

andrewsykim:
I re-enabled the existing exec liveness probe test and the test now passes with dockershim, but fails with containerd (w/ Kind). 

Makes me think the timeout is not respected in some versions of containerd possibly?

pkg/kubelet/cri/remote/remote_runtime.go
```diff
@@ -387,6 +389,17 @@ func (r *RemoteRuntimeService) ExecSync(containerID string, cmd []string, timeou
 	resp, err := r.runtimeClient.ExecSync(ctx, req)
 	if err != nil {
 		klog.Errorf("ExecSync %s '%s' from runtime service failed: %v", containerID, strings.Join(cmd, " "), err)
+
+		// If exec timed out, return utilexec.CodeExitError with an exit status as expected
+		// from prober for failed probes.
+		// TODO: utilexec should have a TimedoutError type and we should return it here once available.
+		if status.Code(err) == codes.DeadlineExceeded {
```
andrewsykim, line 396: This is a little hacky but I think it's reasonable until we add a dedicated `TimedOutError` error for utilexec. For now we have to return `utilexec.CodeExitError` since the exec prober is expecting this for failed probes

https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/pkg/probe/exec/exec.go#L60-L69

Alternative, we can update the prober worker to stop ignoring errors from the prober

https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/pkg/kubelet/prober/worker.go#L245-L249

But that change seems riskier since a probe error does not always lead to a probe failure.

andrewsykim (commented):
> @andrewsykim Can you add a regression test for this case? I believe we saw a similar issue with Containerd in the past, but was addressed. Can you double confirm it?

@dchen1107 can confirm that the exec liveness probe timeout was not working for containerd as well. Last commit also has the fix for containerd. I also re-enabeld the exec liveness timeout e2e test which was previously being skipped.

dashpole (commented):
Just to make sure I understand:
1. Docker still doesn't support timeouts for exec commands
2. We should consider a probe "timed out" after the timeout period, even though the exec probe is still running in the docker container.

SergeyKanzhelev:
> > not killing the process in docker may cause to some unwanted effects when probes overwhelm the pod and starve it's resources. This situation is something that was discussed on Container Notifications design review. I believe @liggitt brought it up.
> 
> We don't kill the exec process in docker shim today anyways though. This is still the case with the changes proposed here, the only difference is that we're respecting the probe timeout already passed into dockershim and returning an error to the prober so that liveness probes registers timeouts as failures.

Understood. My worry is whether the timeout return without killing the process will increase a number of probe calls thus overwhelming the pod.

> I agree though that the proper solution here probably involves having a timeout in dockershim's StartExec call. I did try to update it at first but was worried of breaking some existing behavior since today we don't kill any exec process for dockershim. This was my attempt at a more iterative approach to respect probe timeouts and not change any other behavior for dockershim exec. Happy to have a follow-up PR to address this or follow-up in #58925
> 
> > I'm worried that this PR doesn't have any tests, even unit tests
> 
> The commit [4a137a8](https://github.com/kubernetes/kubernetes/commit/4a137a8230c01cbe8881c817d756371ae7fecbe0) unskips an e2e test that tests exec timeouts for liveness probes.

Yep, sorry didn't notice this. With the new test it's more apparent.

andrewsykim:
> Understood. My worry is whether the timeout return without killing the process will increase a number of probe calls thus overwhelming the pod.

Ah, I see what you mean now. I think it shouldn't because the number of probes should be based on `periodSeconds` and not `timeoutSeconds`.

SergeyKanzhelev:
> > Understood. My worry is whether the timeout return without killing the process will increase a number of probe calls thus overwhelming the pod.
> 
> Ah, I see what you mean now. I think it shouldn't because the number of probes should be based on `periodSeconds` and not `timeoutSeconds`.

You are right. The only scenario I can imagine when this may become an issue is readiness probes. All other probes when exceed timeout would indicate a failure and will restart the pod. With the readiness probes - let's say DB check is implemented in readiness probe and network to DB just got super slow. Pod now will be quickly identified as not ready (which is definitely an improvement), but new readiness probes will keep coming. If actual executable slow down is less than periodSeconds, then there will be no stuck processes. Thus it would be a rare case when this change overwhelmes the pod.

Thinking about it, there is another thing we may want to raise awareness of. Not trying to block this pr or something, just share some ideas on what can also go wrong... The expected behavior of StartupProbe may change. I'd think that startup probes might be most affected by cold start initialization issues. With the [default timeout of 1 second](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes) some pods using startup probe may start experiencing failures which would lead to pod restart. CC: @matthyx Same concern I think may be about the very first liveness probe. CRI may be less affected as it adds 2 minutes to timeout https://github.com/kubernetes/kubernetes/blob/683e9b8cc848e13e71458e16e2e834e1e02ac26f/pkg/kubelet/cri/remote/remote_runtime.go#L261-L263. I didn't find the same adjustment in dockershim

matthyx:
> With the [default timeout of 1 second](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes) some pods using startup probe may start experiencing failures which would lead to pod restart.

In that particular case we can always set `initialDelaySeconds` to mimic the CRI timeout. But that's a good point and maybe worth mentioning somewhere in the doc.

andrewsykim:
> Thinking about it, there is another thing we may want to raise awareness of. Not trying to block this pr or something, just share some ideas on what can also go wrong... The expected behavior of StartupProbe may change. I'd think that startup probes might be most affected by cold start initialization issues. With the default timeout of 1 second some pods using startup probe may start experiencing failures which would lead to pod restart. CC: @matthyx Same concern I think may be about the very first liveness probe. CRI may be less affected as it adds 2 minutes to timeout 

Good catch -- I agree this could potentially cause some issues if users were previously relying on the fact that dockershim never respected the exec timeout. So enforcing the default 1s timeout could end up failing probes that may have previously succeeded due to the timeout being ignored. Consequently though, since the exec timeout is ignored today, a startup probe that takes longer than 10s will be falsely reported as succeeding when the pod should have never started. 

> In that particular case we can always set initialDelaySeconds to mimic the CRI timeout. But that's a good point and maybe worth mentioning somewhere in the doc.

Agree that at the very least we should document this change in behavior in the release notes, even though the expected behavior was never respected.

> CRI may be less affected as it adds 2 minutes to timeout 

I thought the added 2 min is only for the gRPC request timeout and doesn't actually apply for the exec timeout. So in the case of exec probes, the 2 min wouldn't apply, unless there was delays in the request unrelated to the actual exec call.

SergeyKanzhelev:
btw, found another active PR attempting the same: https://github.com/kubernetes/kubernetes/pull/87281/ =)

Adding this as a topic for weekly meeting

andrewsykim:
Thanks @SergeyKanzhelev! Will try to make it :)

dashpole:
/approve cancel
while discussion is ongoing

dchen1107:
As a followup of today's SIG Node and @andrewsykim's comment at: https://github.com/kubernetes/kubernetes/pull/94115#issuecomment-678681860

I believe Containerd does respect ExecSync timeout value, and the related code is here: https://github.com/containerd/cri/blob/master/pkg/server/container_execsync.go#L152-L171 I believe this is what Michael Cosby was trying to say earlier at the meeting. 

We need to figure out why the timeout value wasn't set on the kubelet side for ExecSync properly. I shared some reasons I roughly remembered at today's meeting. Also after timeout, what are the semantic at Kubernetes API level. From today's conversation, looks like there is no one-size-fit-all solution for all probes yet, at least we didn't reach an agreement @andrewsykim will document the existing usecases for all probes and share it with SIG Node community next week. Thanks!

andrewsykim:
Thanks @dchen1107 -- I'll try to summarize everything for next week's SIG meeting.

> I believe Containerd does respect ExecSync timeout value, and the related code is here: https://github.com/containerd/cri/blob/master/pkg/server/container_execsync.go#L152-L171 I believe this is what Michael Cosby was trying to say earlier at the meeting.

One clarifying point here, though the timeout is respected in ExecSync -- the grpc error we receive from the timeout (DeadlineExceeded) is ignored from the probe worker

https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/pkg/kubelet/prober/worker.go#L245-L249 

This commit https://github.com/kubernetes/kubernetes/pull/94115/commits/018bd2417830110876fab6e6f6f17d269b5c6e08 fixes it by returning the `utilexec.CodeExitError` error which is then properly interpreted by the prober as a failed probe, otherwise any other error is ignored:

https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/pkg/probe/exec/exec.go#L60-L69

SergeyKanzhelev:
This PR and its side effects is actively being discussed and worked on. I am closing previous PRs around this topic. @tnqn, @tedyu, @louyihua, thank you for attempting to fix this important issue before. Please join our discussion here. PRs being closed:

- https://github.com/kubernetes/kubernetes/pull/92465/
- https://github.com/kubernetes/kubernetes/pull/87281/
- https://github.com/kubernetes/kubernetes/pull/58925/

test/e2e/common/container_probe.go
```diff
@@ -213,8 +212,6 @@ var _ = framework.KubeDescribe("Probing container", func() {
 		Description: A Pod is created with liveness probe with a Exec action on the Pod. If the liveness probe call  does not return within the timeout specified, liveness probe MUST restart the Pod.
 	*/
 	ginkgo.It("should be restarted with a docker exec liveness probe with timeout ", func() {
-		// TODO: enable this test once the default exec handler supports timeout.
-		e2eskipper.Skipf("The default exec handler, dockertools.NativeExecHandler, does not support timeouts due to a limitation in the Docker Remote API")
```
oomichi, line 217: Nice :-)

test/e2e/common/container_probe.go
```diff
@@ -625,3 +639,35 @@ func RunLivenessTest(f *framework.Framework, pod *v1.Pod, expectNumRestarts int,
 			ns, pod.Name, expectNumRestarts, observedRestarts)
 	}
 }
+
+func RunReadinessFailTest(f *framework.Framework, pod *v1.Pod, notReadyUntil time.Duration) {
```
oomichi, line 643: nite: We can keep this function as a private one with `runReadinessFailTest`

oomichi (commented):
/cc @oomichi

SergeyKanzhelev:
cross linking to the KEP: https://github.com/kubernetes/enhancements/issues/1972

pkg/probe/exec/errors.go
```diff
@@ -0,0 +1,47 @@
...
+*/
+
+package exec
+
+import (
+	"time"
+)
+
+// NewTimedoutError returns a new TimedoutError.
+func NewTimedoutError(err error, timeout time.Duration) *TimeoutError {
```
SergeyKanzhelev, line 24: Perhaps keep the name consistent:

```suggestion
func NewTimeoutError(err error, timeout time.Duration) *TimeoutError {
```

pkg/probe/exec/errors.go
```diff
@@ -0,0 +1,47 @@
...
+	return &TimeoutError{
+		err:     err,
+		timeout: timeout,
+	}
+}
+
+// TimedoutError is an error returned on exec probe timeouts. It should be returned by CRI implementations
```
SergeyKanzhelev, line 31:
```suggestion
// TimeoutError is an error returned on exec probe timeouts. It should be returned by CRI implementations
```

pkg/probe/exec/exec.go
```diff
@@ -66,6 +68,16 @@ func (pr execProber) Probe(e exec.Cmd) (probe.Result, string, error) {
 			}
 			return probe.Failure, string(data), nil
 		}
+
+		timedoutErr, ok = err.(*TimeoutError)
```
SergeyKanzhelev, line 72: names consistency...

```suggestion
		timeoutErr, ok = err.(*TimeoutError)
```

pkg/probe/exec/exec.go
```diff
@@ -66,6 +68,16 @@ func (pr execProber) Probe(e exec.Cmd) (probe.Result, string, error) {
...
+		if ok {
+			if utilfeature.DefaultFeatureGate.Enabled(features.ExecProbeTimeout) {
+				return probe.Failure, string(data), nil
+			}
+
+			return probe.Unknown, "", timedoutErr
```
SergeyKanzhelev, line 78: this changes old behavior by returning a different error type. Should we keep the original `err` inside the `timedoutErr` to return here?

Just thinking out loud.

pkg/probe/exec/errors.go
andrewsykim: Yeah good call, fixed.

pkg/probe/exec/errors.go
andrewsykim: Fixed

pkg/probe/exec/exec.go
andrewsykim: Fixed

pkg/probe/exec/exec.go
andrewsykim: Updated to return original error, though I think it doesn't matter as long as we return `probe.Unknown` but a non-nil err.

pkg/kubelet/cri/.import-restrictions
```diff
@@ -3,3 +3,4 @@ rules:
   - selectorRegexp: k8s[.]io/kubernetes
     allowedPrefixes:
       - k8s.io/kubernetes/pkg/kubelet/cri
+      - k8s.io/kubernetes/pkg/probe/exec
```
andrewsykim, line 6: Don't have context on why we're restricting internal imports here, can't think of a better location to put TimeoutError though

andrewsykim: @dims in https://github.com/kubernetes/kubernetes/pull/90552 it seems like this import restrictions file was moved here from another kubelet package. The comment above "prevent exposing internal api in streaming packages" makes me think that this was initially intended for `pkg/kubelet/cri/streaming` but I'm not sure anymore. Could use your guidance on what's best here -- what's the reason for not allowing internal imports to `pkg/kubelet/cri`, were there plans to move that to staging at some point?

dims: @andrewsykim yes, we needed `streaming` hermetically sealed :)

andrewsykim: Okay, so can I move this import restriction file to `pkg/kubelet/cri/streaming` instead of in `pkg/kubelet/cri` where it currently lives? That would allow for some imports I'm trying to add here (`pkg/probe/exec`)

dims: 👍

pkg/kubelet/dockershim/exec.go
```diff
@@ -110,29 +112,47 @@ func (*NativeExecHandler) ExecInContainer(client libdocker.Interface, container
 		return err
 	}
 
+	var execTimeout <-chan time.Time
```
SergeyKanzhelev, line 115: this needs to be only done when feature flag is set. Otherwise this will change the today's behavior.

andrewsykim: Good catch -- updated so that execTimeout is nil if the gate is not enabled.

pkg/kubelet/dockershim/exec.go
andrewsykim: Done, thanks

SergeyKanzhelev:
/lgtm

not sure what's wrong with tests, I might have time later to look into it

andrewsykim:
They weirdly seem related to the changes, especially the ones failing due to unready pods. Thanks @SergeyKanzhelev

pkg/probe/exec/exec.go
```diff
@@ -66,6 +68,16 @@ func (pr execProber) Probe(e exec.Cmd) (probe.Result, string, error) {
...
 			return probe.Failure, string(data), nil
 		}
+
+		timeoutErr, ok := err.(*TimeoutError)
+		if ok {
+			if utilfeature.DefaultFeatureGate.Enabled(features.ExecProbeTimeout) {
+				return probe.Failure, string(data), nil
+			}
+
+			klog.Warningf("Exec probe timed out after %s but ExecProbeTimeout feature gate was disabled", timeoutErr.Timeout())
```
ehashman, line 78: Why are we surfacing this as a log warning and not in a Prometheus metric?

andrewsykim: Couple reasons:
* this log warning is only going to be around for 1 release until we always enforce probe timeouts in v1.21., it wouldn't make sense to add a prometheus metric now and then start a deprecation period for it one release later.
* there is an action tied to this warning and hints that the user should not be relying on this behavior going forward -- this would be hard to convey in a prometheus metric.

SergeyKanzhelev:
/lgtm

may be a good ida to squash commits as per contributing guide

andrewsykim:
> may be a good ida to squash commits as per contributing guide

I tried to keep each commit distinct units of work as per the contributing guidelines:

> All commits left on your branch after a review should represent meaningful milestones or units of work. Use commits to add clarity to the development and review process.

dims:
@derekwaynecarr @sjenning this is ready, please take a look, i'd like to see this soak well in our CI before we hit code freeze. let's please get this in.

MonzElmasry:
Hello @derekwaynecarr @sjenning 
Just wanted to post a friendly reminder that the code freeze is starting 12.11.2020 (about 1.5 weeks from now) and while there is still plenty of time, we want to ensure that each PR has a chance to be merged on time.

pkg/probe/exec/exec.go
ehashman: > it wouldn't make sense to add a prometheus metric now and then start a deprecation period for it one release later.

FWIW this is the point of alpha metrics :) see: https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/20190404-kubernetes-control-plane-metrics-stability.md#stability-classes

test/e2e/common/container_probe.go
oomichi: I've confirmed this test problem can be reproduced if not applying this PR change with https://github.com/kubernetes/kubernetes/pull/96178 like
```
Kubernetes e2e suite: [k8s.io] Probing container should be restarted with a docker exec liveness probe with timeout 

test/e2e/common/container_probe.go:214 Nov  3 20:43:35.606: pod container-probe-5360/busybox-40384554-8b2e-41ed-8bd2-8c432418dc52 - expected number of restarts: 1, found restarts: 0 test/e2e/common/container_probe.go:224
```
This looks good for me to fix the bug and looking forward to seeing this in v1.20.

oomichi (commented):
/lgtm

MonzElmasry:
@andrewsykim can you resolve the conflict for this pr to make it to 1.20, we have 3 days left for code freeze

andrewsykim:
> can you resolve the conflict for this pr to make it to 1.20, we have 3 days left for code freeze

This is ready to merge, just needs an approver. @derekwaynecarr @dchen1107 can you take a look please?

SergeyKanzhelev:
/hold

@andrewsykim let's update release note to mention the new flag to disable this functionality? Please feel free to remove the hold if disagree.

Sorry for commenting this late.

andrewsykim:
Done thanks for calling that out @SergeyKanzhelev 

/hold cancel

pkg/kubelet/dockershim/exec.go
```diff
@@ -110,29 +114,48 @@ func (*NativeExecHandler) ExecInContainer(client libdocker.Interface, container
...
+
+			// Only limit the amount of InspectExec calls if the exec timeout was not set.
+			// When a timeout is not set, we stop polling the exec session after 5 attempts and allow the process to continue running.
+			if execTimeout == nil {
+				count++
+				if count == 5 {
+					klog.Errorf("Exec session %s in container %s terminated but process still running!", execObj.ID, container.ID)
+					return nil
+				}
+			}
```
dionysius, line 156: Doesn't this mean that any probe, **without timeout**, taking longer than 10 seconds will automatically be **no** error, independent of what the actual exit of the command is?

ehashman: There have been a number of follow-ups to this PR since it landed so I suggest you take a look at what's committed on the master branch.

dionysius: Ignore my question - interpretation error from my side, on master (and probably on this commit as well) `client.StartExec()` is blocking and thus this code block happens **after** the command has finished in some form. It is 10 seconds maximum for gathering the exit info, although the code on master is slightly different with the same effect.

PrabhuMathi:
Facing same issue in 1.19.7 (AKS) with containerd(1.5.0-beta) as cri. Could you please advise in which version of containerd or kubernetes patch this was fixed?

Pod stuck in running 0/1 status

`    readinessProbe:
      exec:
        command:
        - sleep
        - "15"
      failureThreshold: 3
      initialDelaySeconds: 10
      periodSeconds: 10
      successThreshold: 1
      timeoutSeconds: 2
`

`Readiness probe errored: rpc error: code = DeadlineExceeded desc = failed to exec in container: timeout 2s exceeded: context deadline exceeded`

matthyx:
@PrabhuMathi I don't think this commit has been cherry-picked in 1.19 - you'll have to use 1.20 or above

PrabhuMathi:
> @PrabhuMathi I don't think this commit has been cherry-picked in 1.19 - you'll have to use 1.20 or above

@matthyx thanks for the update, Let me give timeout in the command itself as mentioned and proceed. Also started facing readiness probe failure after migrating to Containerd any ref on this please?

SergeyKanzhelev:
> > @PrabhuMathi I don't think this commit has been cherry-picked in 1.19 - you'll have to use 1.20 or above
> 
> @matthyx thanks for the update, Let me give timeout in the command itself as mentioned and proceed. Also started facing readiness probe failure after migrating to Containerd any ref on this please?

As far as I remember, with Containerd before 1.19, probe will timeout, but it will not be treated as an error. With Docker there will be no timeout. Is it what you experience?

PrabhuMathi:
> > > @PrabhuMathi I don't think this commit has been cherry-picked in 1.19 - you'll have to use 1.20 or above
> > 
> > 
> > @matthyx thanks for the update, Let me give timeout in the command itself as mentioned and proceed. Also started facing readiness probe failure after migrating to Containerd any ref on this please?
> 
> As far as I remember, with Containerd before 1.19, probe will timeout, but it will not be treated as an error. With Docker there will be no timeout. Is it what you experience?

Yes true in docker same pod runs fine, but when i move it to contained node facing this issue, but actual problem i'm facing is not about timeout but readiness probe is failing with command (sleep 15)

matthyx:
@PrabhuMathi can you share your podspec?

PR #94125, opened 2020-08-20
Only process all nodes when incoming pod has no preferred affinity

soulxu:
To calculate the score for inter-pod affinity, the PreScore plugin needs to process the
two cases below:

1. The incoming pod's preferred affinity/anti-affinity matches with the existing pod.
2. The existing pod's preferred affinity/anti-affinity matches with the incoming pod.

For #1, if the incoming pod with any preferred affinity/anti-affinity terms, all the existing
pod on all nodes should be checked with the incoming pod's preferred affinity/anti-affinity
terms. 

For #2, if the existing pod with any preferred affinity/anti-affinity terms, then it should
check with the incoming pod.

If the incoming pod without any preferred affinity/anti-affinity terms, then only needs
to check the existing pod with preferred affinity/anti-affinity terms.

The current implementation only checks whether the incoming pod with affinity/anti-affinity or not,
to determine whether needs to check all the nodes. That will include the case of the
incoming pod with required affinity/anti-affinity terms, which isn't the case PreScore plugin cares about.

So this PR only checks the preferred affinity/anti-affinity to determine whether should
process all the nodes.

**What type of PR is this?**
/kind cleanup

It is more efficient for reducing the number of nodes that need to be processed.

### DIFF
--- a/pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
+++ b/pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
@@ -141,14 +141,14 @@ func (pl *InterPodAffinity) PreScore(
 	}
 
 	affinity := pod.Spec.Affinity
-	hasAffinityConstraints := affinity != nil && affinity.PodAffinity != nil
-	hasAntiAffinityConstraints := affinity != nil && affinity.PodAntiAffinity != nil
+	hasPreferredAffinityConstraints := affinity != nil && affinity.PodAffinity != nil && len(affinity.PodAffinity.PreferredDuringSchedulingIgnoredDuringExecution) > 0
+	hasPreferredAntiAffinityConstraints := affinity != nil && affinity.PodAntiAffinity != nil && len(affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution) > 0
 
-	// Unless the pod being scheduled has affinity terms, we only
+	// Unless the pod being scheduled has preferred affinity terms, we only
 	// need to process nodes hosting pods with affinity.
 	var allNodes []*framework.NodeInfo
 	var err error
-	if hasAffinityConstraints || hasAntiAffinityConstraints {
+	if hasPreferredAffinityConstraints || hasPreferredAntiAffinityConstraints {
 		allNodes, err = pl.sharedLister.NodeInfos().List()
 		if err != nil {
 			framework.NewStatus(framework.Error, fmt.Sprintf("get all nodes from shared lister error, err: %v", err))
@@ -178,10 +178,10 @@ func (pl *InterPodAffinity) PreScore(
 		if nodeInfo.Node() == nil {
 			return
 		}
-		// Unless the pod being scheduled has affinity terms, we only
+		// Unless the pod being scheduled has preferred affinity terms, we only
 		// need to process pods with affinity in the node.
 		podsToProcess := nodeInfo.PodsWithAffinity
-		if hasAffinityConstraints || hasAntiAffinityConstraints {
+		if hasPreferredAffinityConstraints || hasPreferredAntiAffinityConstraints {
 			// We need to process all the pods.
 			podsToProcess = nodeInfo.Pods
 		}


pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
```diff
@@ -141,14 +141,14 @@ func (pl *InterPodAffinity) PreScore(
 	}
 
 	affinity := pod.Spec.Affinity
-	hasAffinityConstraints := affinity != nil && affinity.PodAffinity != nil
-	hasAntiAffinityConstraints := affinity != nil && affinity.PodAntiAffinity != nil
+	hasPreferredAffinityConstraints := affinity != nil && affinity.PodAffinity != nil && len(affinity.PodAffinity.PreferredDuringSchedulingIgnoredDuringExecution) > 0
+	hasPreferredAntiAffinityConstraints := affinity != nil && affinity.PodAntiAffinity != nil && len(affinity.PodAffinity.PreferredDuringSchedulingIgnoredDuringExecution) > 0
```
chendave, line 145: should be

```go
len(affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution)
```

chendave (commented):
looks to me a sound improvement, just need to fix the nits.

chendave:
testcase of "Anti Affinity: pod that doesnot match existing pods in node will get high score" seems like is impacted by this change.

pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
soulxu: oops! thanks, dave!

pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
```diff
@@ -181,7 +181,7 @@ func (pl *InterPodAffinity) PreScore(
 		// Unless the pod being scheduled has affinity terms, we only
 		// need to process pods with affinity in the node.
 		podsToProcess := nodeInfo.PodsWithAffinity
-		if hasAffinityConstraints || hasAntiAffinityConstraints {
+		if hasPreferredAffinityConstraints || hasPreferredAntiAffinityConstraints {
```
adtac, line 184: nit: update the comments to s/affinity/prefferred affinity/g?

pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
```diff
@@ -141,14 +141,14 @@ func (pl *InterPodAffinity) PreScore(
 	}
 
 	affinity := pod.Spec.Affinity
-	hasAffinityConstraints := affinity != nil && affinity.PodAffinity != nil
-	hasAntiAffinityConstraints := affinity != nil && affinity.PodAntiAffinity != nil
+	hasPreferredAffinityConstraints := affinity != nil && affinity.PodAffinity != nil && len(affinity.PodAffinity.PreferredDuringSchedulingIgnoredDuringExecution) > 0
+	hasPreferredAntiAffinityConstraints := affinity != nil && affinity.PodAntiAffinity != nil && len(affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution) > 0
 
 	// Unless the pod being scheduled has affinity terms, we only
```
adtac, line 147: nit: same here

pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
soulxu: thanks, nice catch!

pkg/scheduler/framework/plugins/interpodaffinity/scoring.go
soulxu: will update

Huang-Wei (commented):
Thanks @soulxu ! Based on current logic, this PR looks good.

However, it reminds me of a potential "issue": which is, should we honor `InterPodAffinityArgs.HardPodAffinityWeight` in a **symmetric** manner? Right now, seems we just honor it when looking at **existing** Pods which has required affinity constraints:

https://github.com/kubernetes/kubernetes/blob/1c548c328a32c5b6819195760d1abba115ab4969/pkg/scheduler/framework/plugins/interpodaffinity/scoring.go#L106-L114

So what if the **incoming** Pod has required affinity constraints? Shouldn't we check all existing Pods as `HardPodAffinityWeight` actually acts as an implicit preferred constraint?

---

Also I noticed that the comments change the semantics a little bit.

Original version:

https://github.com/kubernetes/kubernetes/blob/1c548c328a32c5b6819195760d1abba115ab4969/staging/src/k8s.io/kube-scheduler/config/v1/types.go#L37-L40

Current version:

https://github.com/kubernetes/kubernetes/blob/1c548c328a32c5b6819195760d1abba115ab4969/pkg/scheduler/apis/config/types_pluginargs.go#L30-L32

cc/ @ahg-g @alculquicondor

ahg-g:
I think we should look at the history behind `HardPodAffinityWeight` and see why it was only limited to existing pods. Note that this will make sense only for affinity (not anti-affinity), but I am not too sure about mixing the semantics of required with preferred.

Huang-Wei:
> but I am not too sure about mixing the semantics of required with preferred.

I felt the same. It looks a bit weird to expose a "preferred" semantics via plugin-level `HardPodAffinityWeight` argument onto all "required" constraints, plus it's not symmetric. Probably we can revisit to drop this field, as an alternative, users can impose "preferred" semantics via preferred constraints explicitly.

soulxu:
> Thanks @soulxu ! Based on current logic, this PR looks good.

Thanks for the review! 

> So what if the **incoming** Pod has required affinity constraints? Shouldn't we check all existing Pods as `HardPodAffinityWeight` actually acts as an implicit preferred constraint?

Hope I understand the issue. For the incoming Pod which has required affinity, then the filters will ensure every node match that affinity, so it should be meaningless for scoring.

soulxu:
> I think we should look at the history behind `HardPodAffinityWeight` and see why it was only limited to existing pods. Note that this will make sense only for affinity (not anti-affinity), but I am not too sure about mixing the semantics of required with preferred.

Yea, both the incoming pod's required affinity and the existing required anti-affinity were checked by the filter, so it doesn't make sense for scoring.

soulxu:
> I felt the same. It looks a bit weird to expose a "preferred" semantics via plugin-level `HardPodAffinityWeight` argument onto all "required" constraints, plus it's not symmetric. Probably we can revisit to drop this field, as an alternative, users can impose "preferred" semantics via preferred constraints explicitly.

Should I file an issue for the discussion?

Huang-Wei:
> Hope I understand the issue. For the incoming Pod which has required affinity, then the filters will ensure every node match that affinity, so it should be meaningless for scoring.

Once the incoming pod (with required affinity, excludes anti-affinity per @ahg-g mentioned) gets scheduled and hence becomes an existing Pod, it does impact the incoming Pod's score:

https://github.com/kubernetes/kubernetes/blob/1c548c328a32c5b6819195760d1abba115ab4969/pkg/scheduler/framework/plugins/interpodaffinity/scoring.go#L106-L114

So my point is, in a reversed order, when it's being scheduled, why its Score is not adjusted via `HardPodAffinityWeight` by checking existing pods? The prerequisite is that we read `HardPodAffinityWeight` as a symmetric value.

soulxu:
> So my point is, in a reversed order, when it's being scheduled, why its Score is not adjusted via `HardPodAffinityWeight` by checking existing pods? The prerequisite is that we read `HardPodAffinityWeight` as a symmetric value.

Ah..I probably get the point. So in the end, from the user point of view, the topology which is with more existing pods match the incoming pod's required affinity will have higher score?

Huang-Wei:
> Ah..I probably get the point. So in the end, from the user point of view, the topology which is with more existing pods match the incoming pod's required affinity will have higher score?

Correct - if we interpret  `HardPodAffinityWeight` (a.k.a. `HardPodAffinitySymmetricWeight`) in a symmetric manner.

soulxu:
After search history, I found this:

https://github.com/kubernetes/kubernetes/blob/d88304507d3b3ff8a15f0948c21be5892693a12f/pkg/scheduler/factory.go#L82

From that comment, the original idea is `the symmetric of the incoming pod's required affinity rule is adding the scoring of the exiting pod's required affinity rule.` which is different from the one @Huang-Wei  pointed out.

so guessing we have some mess for the definition of `symmetric`  :) 

the `symmetric` can be:
	1. The symmetric for the incoming pod's required affinity rule is the scoring of the existing pod's required affinity rule. (This is the current implementation)
	2. The symmetric for the incoming pod's required affinity rule is filtering the existing pod's required affinity rule with the incoming pod. (this isn't implemented)
	3. The symmetric for the scoring of the existing pod's required affinity rule is scoring the incoming pod's required affinity rule. (this is pointed by Huang Wei, not implement)
	
For #1, the `HardPodAffinityWeight` adds an implicitly preferred affinity rule. But the user also can explicitly specify a preferred affinity rule if there is any specific term he is interesting in. `HardPodAffinityWeight` is more a default value for any term match the incoming pod. (Will there be a conflict between the user-specified one and `HardPodAffinityWeight`?)

For #2, there is no way to this now, we only implement the anti-affinity check for the existing pod.
https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/interpodaffinity/filtering.go#L178

For #3, just like #1, this can be done by specifying a preferred affinity which is the same as the existing pod's required affinity also. Also, this is for the case the user knowing the specific affinity term he wants to score. `HardPodAffinityWeight` is the default value for any term.

For @Huang-Wei 's point is related to #1 and #3. #2 can be discussed separately.

So…maybe I can nail down as two questions:
	A. For #1, since the user can specify explicitly for a specific term. So the only left question is do we need per pod parameter rather than plugin arg?
	
	generically, the explicit parameter better than the implicit parameter. The user won't know cluster-admin configure what weight in the cluster. So the user only can guess the final behavior. But just as said, the user can specify explicitly the preferred term, so maybe we can just deprecated this plugin arg?
	
For #3, do we want to implement that symmetric? If yes, that should be separated parameter with #1. And whether we want it to be per pod parameter rather than plugin arg also?

@Huang-Wei @ahg-g

chendave:
> Then how to apply some preference to the incoming Pod? Here `HardPodAffinityWeight` comes onto stage. It applies additional weight to each existing Pod that has PodAffinity which matches the incoming Pod, to favor particular nodes instead of randomly place the incoming Pod.

I agree with most of the comments here, but something that is still not much clear enough is,
- why the incoming pod without the affinity **cannot** randomly place the incoming pod, if there is any preference, or need to attract the incoming pod, why not define the affinity rule for the incoming instead?
- why it's `RequiredAffinityTerms` comes into place to take the role for such a preference? `RequiredAffinityTerms` mean a hard requirement this is semantically against the word **preference** here.

chendave:
maybe we just need an usecase to show the reason of the existing pod with affinity rule needs to influence the incoming pod without the affinity rule and how it matters in that particular case.

alculquicondor:
/hold

this needs a release note, as it is a behavioral change. Also I find the PR description a bit confusing. Could you try to enhance it? Perhaps including some conclusions from the rest of the discussion.

soulxu:
> this needs a release note, as it is a behavioral change. Also I find the PR description a bit confusing. Could you try to enhance it? Perhaps including some conclusions from the rest of the discussion.

@alculquicondor, this PR doesn't change any behavior from the user's point of view. It only reduces the number of pods to be processed, remove the process of those pods are useless for the scoring. Should I add a release note for this case?

I will improve the PR description.

alculquicondor:
You are right. Then this shouldn't be kind-feature @Huang-Wei ?

ahg-g:
lets keep it feature (as in performance enhancement) so that it gets merged later.

soulxu:
@alculquicondor, the PR description updated, let me whether it is ok. thanks for your review in advance!

alculquicondor:
I would remove the last paragraph, as it not relevant to the change.

Huang-Wei:
In the description section:

> When InterPodAffinityArgs.HardPodAffinityWeight is set as true

Should be: "... is set to a non-zero value".

> Then this shouldn't be kind-feature @Huang-Wei ?

@ahg-g explained in https://github.com/kubernetes/kubernetes/pull/94125#issuecomment-680963997.

soulxu:
Thanks for the comment! updated

PR #94132, opened 2020-08-20
Fix lint failures for pkg/apis/authentication/v1

pablitoergosum:
**What type of PR is this?**
/kind cleanup

It fixes golint failures in pkg/apis/authentication/v1

Fixes part of [#68026 ](https://github.com/kubernetes/kubernetes/issues/68026)

NONE


liggitt:
Thanks for the PR, but generated files cannot be manually modified like this to satisfy golint, and golint does not provide ways to skip specific checks or instances. From the golint readme:

> We will not be adding pragmas or other knobs to suppress specific warnings, so do not expect or require code to be completely "lint-free". In short, this tool is not, and will never be, trustworthy enough for its suggestions to be enforced automatically, for example as part of a build process.

/close

PR #94134, opened 2020-08-20
Make similar buckets for api and etcd request duration histogram

tkashem:
**What type of PR is this?**
/kind bug

`etcd_request_duration_seconds` uses the default buckets provided by prometheus client library. 
`DefBuckets = []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}`
The maximum bucket size is `10s`. On the other hand, `apiserver_request_duration_seconds` uses more fine grained bucket sizes and the maximum bucket size is `60s`

image

The left panel shows latency for `Deployment-DELETE` api (metric=`apiserver_request_duration_seconds`), this is taking about  `40s` to complete. On the other hand, etcd latency (metric=`etcd_request_duration_seconds`) for the same object  `apps.Deployment-delete` is capped at `10s`. Now the difference in latency is hard to account for. It cloud be latency from ectd but we can't answer this question by looking at the metrics. 

If the etcd metric has similar bucket sizes, we could account for the difference in latency. 

This PR makes the bucket sizes for both metrics similar. Also,  no existing bucket for `etcd_request_duration_seconds` was dropped.

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/storage/etcd3/metrics/metrics.go
+++ b/staging/src/k8s.io/apiserver/pkg/storage/etcd3/metrics/metrics.go
@@ -35,8 +35,12 @@ import (
 var (
 	etcdRequestLatency = compbasemetrics.NewHistogramVec(
 		&compbasemetrics.HistogramOpts{
-			Name:           "etcd_request_duration_seconds",
-			Help:           "Etcd request latency in seconds for each operation and object type.",
+			Name: "etcd_request_duration_seconds",
+			Help: "Etcd request latency in seconds for each operation and object type.",
+			// Keeping it similar to the buckets used by the apiserver_request_duration_seconds metric so that
+			// api latency and etcd latency can be more comparable side by side.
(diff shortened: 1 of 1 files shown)


hexfusion:
+1 this makes a lot of sense to me, thanks for doing this @tkashem

logicalhan (commented):
/lgtm

/cc @brancz 
(since we're changing buckets)

wojtek-t:
I'm fine with it, but I will wait with approving for @brancz comment.

/assign @brancz

s-urbaniak:
hey :wave: afaik @brancz is currently still vacationing :-)

metalmatze:
I like this _a lot_. We have latencies on our dashboards for the API server and etcd too but didn't try to correlate those just yet. If we can make this change, this will become a lot easier indeed. :100:

squat:
Looks great! This would be very handy (:

s-urbaniak:
/approve

also from my side :+1: (not sure if that approval works or not)

brancz:
Looks good from instrumentation side.

/lgtm

PR #94135, opened 2020-08-20
agnhost: add functionality to inject out of order tcp packets

### LINKED ISSUE #93281: The test "should resolve connection reset issue #74839 [Slow]" does not support IPv6
**Which jobs are failing**:

n/a

"should resolve connection reset issue #74839 [Slow]"

it never worked

**Testgrid link**:

n/a

**Reason for failure**:

The image used for the test is using only IPv4

https://github.com/kubernetes/kubernetes/tree/master/test/images/regression-issue-74839

We should:

- [x] refactor the code to support IPv6 
- [ ] ~consolidate the image in agnhost~
- [x] configure an IPv6 job to test network

aojea:
The test for regression issue 74839, was using its own image
named regression-issue-74839. It only works for IPv4

This adds the same functionaliy to the agnhost image, and
adds support for IPv6

This is the first step to fix
#93281

/kind failing-test
```release-note
NONE
```


test/images/agnhost/tcp-seq-rst/tcp.go
```diff
@@ -0,0 +1,189 @@
...
+
+	return pkt
+}
+
+func checksumTCP(src, dest net.IP, tcpHeader, data []byte) uint16 {
+	log.Printf("calling checksumTCP: %v %v %v %v", src, dest, tcpHeader, data)
+	chk := &tcpChecksumer{}
+
+	// Encode pseudoheader.
+	chk.add(src.To4())
```
aojea, line 139: I know this is incorrect for IPv6, but the test works, 
I think that the tcp checksum offloading fixes it :smiley:

aojea:
/assign @claudiubelu
/assign @BenTheElder 
:thinking:

test/images/agnhost/tcp-seq-rst/main.go
```diff
@@ -0,0 +1,188 @@
...
+			if seq, ok := pending[addr.String()]; ok {
+				log.Println("connection established")
+				delete(pending, addr.String())
+
+				badPkt := &tcpPacket{
+					SrcPort:    pkt.DestPort,
+					DestPort:   pkt.SrcPort,
+					Ack:        seq,
+					Seq:        pkt.Ack - 100000,      // Bad: seq out-of-window
+					Flags:      (5 << 12) | PSH | ACK, // Offset and Flags  oooo000F FFFFFFFF (o:offset, F:flags)
```
danwinship, line 140: const for the `12`? Also, it's not clear what the `5` refers to...

test/images/agnhost/tcp-seq-rst/main.go
```diff
@@ -0,0 +1,188 @@
...
+					log.Printf("conn.WriteTo() error: %v", err)
+				}
+				log.Printf("boom packet: %+v, flag: %v, data: %v, addr: %v", badPkt, badPkt.FlagString(), data, addr)
+
+			}
+		}
+	}
+}
+
+func getIP() net.IP {
```
danwinship, line 158: returning a single IP means it will work for single-stack IPv4 or IPv6 but only has a 50-50 chance of doing what you expect in dual-stack. (ie, you might connect via IPv6 but it's snooping IPv4).

Which leads to another problem; if the snooping thread fails catastrophically (eg, snooping the wrong IP), the behavior is that the e2e test will succeed. It would be good if we could verify somehow that it was actually working.

test/images/agnhost/tcp-seq-rst/tcp.go
danwinship: "The test works" how? If the packet gets dropped for having a bad TCP checksum then the e2e test will pass but you aren't actually testing that the iptables filter rule works.

test/images/agnhost/tcp-seq-rst/tcp.go
```diff
@@ -0,0 +1,189 @@
...
+	chk.add(dest.To4())
+
+	var pseudoHeader [4]byte
+	pseudoHeader[1] = tcpProtoNum
+	binary.BigEndian.PutUint16(pseudoHeader[2:], uint16(len(data)+len(tcpHeader)))
+	chk.add(pseudoHeader[:])
+	chk.add(tcpHeader)
+	chk.add(data)
+
+	log.Printf("checksumer: %+v", chk)
```
danwinship, line 149: "checksummer" with two "m"s (everywhere)

test/images/agnhost/agnhost.go
```diff
@@ -44,12 +44,13 @@ import (
...
 	resconsumerctrl "k8s.io/kubernetes/test/images/agnhost/resource-consumer-controller"
 	servehostname "k8s.io/kubernetes/test/images/agnhost/serve-hostname"
+	tcpseqrst "k8s.io/kubernetes/test/images/agnhost/tcp-seq-rst"
 	testwebserver "k8s.io/kubernetes/test/images/agnhost/test-webserver"
 	"k8s.io/kubernetes/test/images/agnhost/webhook"
 )
 
 func main() {
-	rootCmd := &cobra.Command{Use: "app", Version: "2.20"}
+	rootCmd := &cobra.Command{Use: "app", Version: "2.22"}
```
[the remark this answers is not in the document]
BenTheElder: https://github.com/kubernetes/kubernetes/pull/94152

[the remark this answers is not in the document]
BenTheElder: this will need to rebase and go to 2.23 after that PR, which will catch this in the future

test/images/agnhost/tcp-seq-rst/main.go
```diff
@@ -0,0 +1,188 @@
...
+
+func init() {
+	CmdTCPSeqRst.Flags().IntVar(&port, "port", 9000, "Port number.")
+}
+
+func main(cmd *cobra.Command, args []string) {
+	// get local IP address
+	ip := getIP()
+	if ip == nil {
+		panic(ip)
```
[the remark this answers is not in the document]
aojea: yeah, I was a bit lazy here since a pod not having an IP will have another problems. Let me fix it

test/images/agnhost/README.md
```diff
@@ -627,6 +627,24 @@ Usage:
...
 ```
 
+### tcp-seq-rsq
+
+Network services with heavy load will cause "connection reset" from time to
+time. Especially those with big payloads. When packets with sequence number
+out-of-window arrived k8s node, conntrack marked them as INVALID. kube-proxy
+will ignore them, without rewriting DNAT. The packet goes back the original
+pod, who doesn't recognize the packet because of the wrong source ip, end up
+RSTing the connection.
```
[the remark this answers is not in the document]
aojea: :+1:

test/images/agnhost/README.md
```diff
@@ -627,6 +627,24 @@ Usage:
     kubectl exec test-agnhost -- /agnhost test-webserver [--port <port>]
 ```
 
+### tcp-seq-rsq
```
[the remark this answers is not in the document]
aojea: `bad-tcp-seq-server`

test/images/agnhost/README.md
```diff
@@ -627,6 +627,24 @@ Usage:
...
+pod, who doesn't recognize the packet because of the wrong source ip, end up
+RSTing the connection.
+
+xref: https://github.com/kubernetes/kubernetes/issues/74839
+
+
+```console
+    kubectl run test-agnhost \
+      --generator=run-pod/v1 \
+      --image=k8s.gcr.io/e2e-test-images/agnhost:2.22 \
```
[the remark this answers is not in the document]
aojea: I was afraid someone was trying to use it with a lower version

test/images/agnhost/tcp-seq-rst/tcp.go
aojea: this is copied verbatim :)

test/images/agnhost/tcp-seq-rst/main.go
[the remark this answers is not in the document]
aojea: however, I think panic is ok for this, I don't want to add external dependencies and keep it as simple as possible.

danwinship:
yes, the other image can go away once the references to it have been removed

aojea:
> So, this basically adds the functionality of the `regression-issue-74839` image into agnhost, right? Then, can we remove that image as well? I don't think we need the same functionality in 2 separate images.

I'm doing that in a follow up PR https://github.com/kubernetes/kubernetes/pull/94136.
I'm not yet fully familiar with the image promotion processes

claudiubelu:
> 
> 
> > So, this basically adds the functionality of the `regression-issue-74839` image into agnhost, right? Then, can we remove that image as well? I don't think we need the same functionality in 2 separate images.
> 
> I'm doing that in a follow up PR #94136.
> I'm not yet fully familiar with the image promotion processes

Gotcha. As for image promotion process, I've detailed the steps here: https://github.com/kubernetes/kubernetes/tree/master/test/images#making-changes-to-images (more specifically, in the 2nd half). But for a tl;dr version:

- this PR will have to merge first.
- wait for the postsubmit job to finish ( https://testgrid.k8s.io/sig-testing-images#post-kubernetes-push-e2e-test-images )
- send update PR here with the newly built image's digest (you can get it by running: ``manifest-tool inspect --raw gcr.io/k8s-staging-e2e-test-images/${IMAGE_NAME}:${VERSION} | jq '.[0].Digest'``): https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/images/k8s-staging-e2e-test-images/images.yaml
- update your other PR you mentioned to also bump the agnhost version in ``test/utils/image/manifest.go``.

claudiubelu:
FWIW, the test passes on Windows with these agnhost changes on the updated test you've sent.

test/images/agnhost/bad-tcp-seq-server/tcp.go
```diff
@@ -0,0 +1,202 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Partially copied from https://github.com/bowei/lighthouse/blob/master/pkg/probe/tcp.go
+
+package badtcpseqserver
```
bowei, line 19: do we have unit test for this?

test/images/agnhost/bad-tcp-seq-server/tcp.go
```diff
@@ -0,0 +1,202 @@
...
+	oddByte byte
+	length  int
+}
+
+func (c *tcpChecksummer) finalize() uint16 {
+	ret := c.sum
+	if c.length%2 > 0 {
+		ret += uint32(c.oddByte)
+	}
+	log.Println("ret: ", ret)
```
bowei, line 178: delete calls to log.Println (rest of file as well)

test/images/agnhost/bad-tcp-seq-server/tcp.go
danwinship: It would be good to have some sort of test to ensure that the test program as a whole actually disrupts the TCP stream in the way it claims to (see also https://github.com/kubernetes/kubernetes/pull/94135#discussion_r474175888), because if it _fails_ to do that, then the associated e2e test will _pass_ even though it didn't actually test the thing it meant to be testing. If we had that then I don't think unit tests here would be especially necessary.

But this wouldn't be a kube e2e test, it would be a test of agnhost itself, to be run when building a new agnhost release image, which is something I think doesn't exist yet.

(There are unit tests in some of the other subdirs, so we could at least test that this parsing code here works correctly (by testing on previously-tcpdumped data), but the actual server requires `CAP_NET_RAW`, and I don't think it would be possible to test it without `CAP_NET_ADMIN` as well, so I don't think we can do that from unit tests.)

claudiubelu:
You might want to take a another look at what you commited. It seems like you've dropped some code after rebase.

aojea:
> You might want to take a another look at what you commited. It seems like you've dropped some code after rebase.

hey, yep, that was on purpose, I just wanted to comment this with you @claudiubelu to ask for your opinion, let me explain:

I realised that for some each new networking scenario we need to create a custom golang pkg that need to go through all the promotion process, and this is slowing a lot the development of new test cases.

I've used before scapy, that is a very rich python framework to craft packages and do a lot of "nice" things with the network, and I was wondering if it would be it better to include it in the agnhost image so adding a new network feature is just creating a configmap with the python script, mount it as a volume and run it in the pod.

I'll add an example here so you can check how it will look like

/hold

test/e2e/network/conntrack.go
```diff
@@ -275,6 +275,159 @@ var _ = SIGDescribe("Conntrack", func() {
...
+		ginkgo.By("creating a TCP service " + serviceName + " with type=NodePort in " + ns)
+		tcpService, err := tcpJig.CreateTCPService(func(svc *v1.Service) {
+			svc.Spec.Type = v1.ServiceTypeClusterIP
+			svc.Spec.Ports = []v1.ServicePort{
+				{Port: 80, Name: "tcp", Protocol: v1.ProtocolTCP, TargetPort: intstr.FromInt(80)},
+			}
+		})
+		framework.ExpectNoError(err)
+
+		// Store the script in a configmap so we can use it from the pods
```
aojea, line 298: @claudiubelu PTAL, the script is a copy paste example I took from the internet, just focus on the approach, 
You just create a configmap and execute it from agnhost, it is more flexible than keep adding subcommands to agnhost, at least for this oneshot things that are not going to be reused.

aojea:
this is too much complexity for low return, will try another approach

Issue #74839, opened 2019-03-02
"Connection reset by peer" due to invalid conntrack packets


anfernee:
**What happened**: Network services with heavy load will cause "connection reset" from time to time. Especially those with big payloads. When packets with sequence number out-of-window arrived k8s node, conntrack marked them as INVALID. kube-proxy will ignore them, without rewriting DNAT. The packet goes back the the original pod, who doesn't recognize the packet because of the wrong source ip, end up RSTing the connection. 

**What you expected to happen**: connection not reset

**How to reproduce it (as minimally and precisely as possible)**: 
@tcarmet wrote this just to reproduce the issue: https://github.com/tcarmet/k8s-connection-reset
Basically run it in a cluster, and after a while you will be able to see "Connection reset by peer" on both server and client pods. 

**Anything else we need to know?**:
Related issue: 
https://github.com/docker/libnetwork/issues/1090

**Environment**:
- Kubernetes version (use `kubectl version`): 1.11.7
- Cloud provider or hardware configuration: gke (should be general to all provider)
- OS (e.g: `cat /etc/os-release`): 
PRETTY_NAME="Debian GNU/Linux 9 (stretch)"
NAME="Debian GNU/Linux"
VERSION_ID="9"
VERSION="9 (stretch)"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
- Kernel (e.g. `uname -a`): Linux gke-cluster-1-11-default-pool-161c67fc-6nzv 4.14.89+ #1 SMP Wed Jan 9 13:35:00 PST 2019 x86_64 GNU/Linux

anfernee:
I created a small app to reproduce this issue: https://github.com/anfernee/k8s-issue-74839

zuzzas:
I've been hitting this exact problem for packets between a Pod and the outside world.
So the DROP rule has to go into the INPUT chain. Is it the job of a CNI driver or the kube-proxy?

kaitoy:
[The fix for this issue](https://github.com/kubernetes/kubernetes/pull/74840) was put to iptables mode but not to [ipvs mode](https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/ipvs/proxier.go#L1817), right?

Isn't ipvs mode affected by this issue, too?

wenlxie:
[...]

PR #94136, opened 2020-08-20
Rename test "resolve connection reset issue #74839" and add IPv6 support

### LINKED ISSUE #93281: The test "should resolve connection reset issue #74839 [Slow]" does not support IPv6

aojea:
/kind cleanup
/kind failing-test

- Rename the test with a meaningful description, it was:
"should resolve connection reset issue #74839"

- Use the agnhost image with the new functionality so it works for IPv6 and IPv4

- Remove the old image used by the test

```release-note
NONE
```

Fixes: #93281
Depends on: #94135

I ran it locally and we can observe how is creating the invalid packets are dropped

```
root@kind-worker:/# ip6tables -L  -v | grep DROP
    0     0 DROP       all      any    any     anywhere             anywhere             /* kubernetes firewall for dropping marked packets */ mark match 0x8000/0x8000
   24  1608 DROP       all      any    any     anywhere             anywhere             ctstate INVALID
```


aojea:
/hold
until the agnhost image is bumped with the new functionality
/assign @danwinship

test/e2e/network/kube_proxy.go
```diff
@@ -236,7 +236,7 @@ var _ = SIGDescribe("Network", func() {
 	// a problem where spurious retransmits in a long-running TCP connection to a service
 	// IP could result in the connection being closed with the error "Connection reset by
 	// peer"
-	ginkgo.It("should resolve connection reset issue #74839 [Slow]", func() {
+	ginkgo.It("should drop invalid conntrack entries", func() {
```
danwinship, line 239: `It("should not be confused by out-of-order packets")`

also, if the test program is going to keep its current behavior of holding the connection open for 10 seconds on success, then you should keep the `[Slow]` (though, it would be better to rewrite the test program to not do that probably)

aojea:
superseded by https://github.com/kubernetes/kubernetes/pull/95351

PR #94137, opened 2020-08-20
Clarify error message when non-existing API object update fails with a conflict error because of a precondition

### LINKED ISSUE #89985: Update of a non-existent object returns 409 instead of 404
**What happened**:
```
mspreitz@mjs-dev-1:~$ curl -X PUT -H "Content-type: application/json" -d @z http://localhost:8080/api/v1/namespaces/default/pods/xyzzy/status
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "Operation cannot be fulfilled on pods \"xyzzy\": StorageError: invalid object, Code: 4, Key: /registry/pods/default/xyzzy, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: 1234, UID in object meta: ",
  "reason": "Conflict",
  "details": {
    "name": "xyzzy",
    "kind": "pods"
  },
  "code": 409
}

mspreitz@mjs-dev-1:~$ curl -X PUT -H "Content-type: application/json" -d @z http://localhost:800/api/v1/namespaces/default/pods/xyzzy
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "Operation cannot be fulfilled on pods \"xyzzy\": StorageError: invalid object, Code: 4, Key: /registry/pods/default/xyzzy, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: 1234, UID in object meta: ",
  "reason": "Conflict",
  "details": {
    "name": "xyzzy",
    "kind": "pods"
  },
  "code": 409
}
```
The contents of `z` are:
```
{
    "apiVersion": "v1",
    "kind": "Pod",
    "metadata": {
        "name": "xyzzy",
        "uid": "1234"
    }
}
```

I get similar results with configmaps.

I expected to get a 404 Not Found

As above.

**Environment**:
- Kubernetes version (use `kubectl version`): current `master` branch, commit 0393b5994a0e02f0a0f3fb50311b3aa191f926ce

- Cloud provider or hardware configuration: Intel Xen VM with 16 virtual CPUs and 64 GB of memory

- OS (e.g: `cat /etc/os-release`): Ubuntu 18.04.2 LTS
- Kernel (e.g. `uname -a`): Linux mjs-dev-1 4.15.0-91-generic #92-Ubuntu SMP Fri Feb 28 11:09:48 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
- Install tools: hack/local-up-cluster.sh
- Others:

matte21:
Clarify the error message that is returned when an update precondition on UID or ResourceVersion fails because the updated object does not exist.
More precisely, clearly say that the update failed because the object does not exist rather than mentioning a generic precondition failure.

**What type of PR is this?**
/kind documentation

Clarify an error message that was confusing to users.

Fixes #89985

The error message that is returned when an API object update fails because of a UID/ResourceVersion precondition failure caused by the fact that the API object does not exist now explicitly mentions that the object does not exist rather than talking about a generic precondition failure.

```release-note

```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/storage/preconditions_test.go
+++ b/staging/src/k8s.io/apiserver/pkg/storage/preconditions_test.go
@@ -1,0 +1,192 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package storage
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apiserver/pkg/apis/example"
+)
+
+// Constant UID and ResourceVersion values used in tests.
+const (
+	uid1             = "u1111"
+	uid2             = "u2222"
+	resourceVersion1 = "rv1111"
+	resourceVersion2 = "rv2222"
+)
+
+func TestSafeCheck(t *testing.T) {
+	testCases := map[string]struct {
+		preconditions *Preconditions
+		objName       string
+		obj           runtime.Object
+		objExists     bool
+		// expectedErrorMessageWords contains the words that we expect in the
+		// error message returned by the preconditions check. A test passes if
+		// and only if there's at least one inner slice for which all words are
+		// present in the error message.
+		expectedErrorMessageWords [][]string
+	}{
+		"nil preconditions": {
+			preconditions: nil,
+			objName:       "foo",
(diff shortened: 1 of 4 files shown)


matte21:
@MikeSpreitzer
@liggitt
@lavalamp

liggitt:
can you add unit tests that exercise the uid and resourceVersion preconditions in the "not found" and "not matching" cases that demonstrate the error message we expect

matte21:
> I'd introduce a sibling function there, label the existing one deprecated, and then make it call the new function you add.

@lavalamp 04c2660 does what you're suggesting, besides factoring out the preconditions code to a dedicated file and adding unit tests for the new function. I wonder whether we should also add tests to GuaranteedUpdate, the caller of the preconditions check, to have a test that fails before the fix in this PR and succeeds after it.

staging/src/k8s.io/apiserver/pkg/storage/preconditions_test.go
```diff
@@ -0,0 +1,192 @@
...
+			},
+		},
+	}
+
+	for description, tc := range testCases {
+		t.Run(description, func(t *testing.T) {
+			gotErr := tc.preconditions.SafeCheck(tc.objName, tc.obj, tc.objExists)
+			if err := pass(gotErr, tc.expectedErrorMessageWords); err != nil {
+				t.Fatal(err)
+			}
```
lavalamp, line 149: Can you also run the regular check method and verify that it returns the same pass / fail value (but no need to make assertions about the error it returns)? That way it will be harder for people to not update both functions.

PR #94144, opened 2020-08-20
Reduce offline volume expansion flake

gnufied:
This PR makes two fixes:

- A volume can be expanded again if we still have stale PV in informer. In worst case this causes a flake because volume expansion will be retried.
- It also increases time we wait for 2nd pod to startup because that time includes time for volume to detach from previous node and attach to new one.

/sig storage

### DIFF
--- a/pkg/controller/volume/expand/expand_controller.go
+++ b/pkg/controller/volume/expand/expand_controller.go
@@ -17,6 +17,7 @@ limitations under the License.
 package expand
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"time"
@@ -28,6 +29,7 @@ import (
 	authenticationv1 "k8s.io/api/authentication/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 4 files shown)


pkg/controller/volume/expand/expand_controller.go
```diff
@@ -335,9 +337,9 @@ func (expc *expandController) runWorker() {
...
 }
 
-func getPersistentVolume(pvc *v1.PersistentVolumeClaim, pvLister corelisters.PersistentVolumeLister) (*v1.PersistentVolume, error) {
+func (expc *expandController) getPersistentVolume(pvc *v1.PersistentVolumeClaim) (*v1.PersistentVolume, error) {
 	volumeName := pvc.Spec.VolumeName
-	pv, err := pvLister.Get(volumeName)
+	pv, err := expc.kubeClient.CoreV1().PersistentVolumes().Get(context.TODO(), volumeName, metav1.GetOptions{})
 
 	if err != nil {
 		return nil, fmt.Errorf("failed to find PV %q in PV informer cache with error : %v", volumeName, err)
```
jsafrane, line 345: The message is not correct now.

pkg/controller/volume/expand/expand_controller.go
```diff
@@ -224,7 +226,7 @@ func (expc *expandController) syncHandler(key string) error {
 		return err
 	}
 
-	pv, err := getPersistentVolume(pvc, expc.pvLister)
+	pv, err := expc.getPersistentVolume(pvc)
```
jsafrane, line 229: Did you consider adding a PV to `expc.pvLister` after the PV is saved with the new size? Would that save us from an API call?

gnufied: I did. it is kind of complicated to do that because expansion call is handled by operation_generator(and we don't have informer there) and it will require some larger re-factoring to do that correctly. I think overall since this call is only made after it has been determined expansion is needed and we don't expect people to be resizing volumes that often, it is I think acceptable tradeoff.

jsafrane: Ack

pkg/controller/volume/expand/expand_controller.go
gnufied: fixed the message

PR #94146, opened 2020-08-20
Make sampleAndWaterMarkHistograms not fall very far behind

### LINKED ISSUE #94122: Inadequate comment on k/apiserver/pkg/util/flowcontrol.Interface.Handle
**What happened**:
In reviewing #93877 I noticed an incorrect fake implementation of Interface.  I looked at the comment on the Handle method, and noticed that it does not say when/how the `queueNoteFn` is called.

I expected the `queueNoteFn` parameter to be as well documented as the others.

Look at https://github.com/kubernetes/apiserver/blob/09eab1cf2cce55a46e5d9042c90f136d91383e19/pkg/util/flowcontrol/apf_filter.go#L39-L44 and see how it does not say when/whether/with-what-parameter-values `queueNoteFn` is called.

**Environment**:
- Kubernetes version (use `kubectl version`): current
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

### LINKED ISSUE #94145: sampleAndWaterMarkHistograms.innerSet can take a long time
**What happened**:
Not observed yet, but deduced from analyzing the code.  If a k/apiserver/pkg/util/flowcontrol/fairqueuing.QueueSet, whose metrics.TimedObserverPair is a pair of k/apiserver/pkg/util/flowcontrol/metrics.sampleAndWaterMarkHistograms, gets zero activity for a long while and then gets some activity, then the loop at https://github.com/kubernetes/apiserver/blob/09eab1cf2cce55a46e5d9042c90f136d91383e19/pkg/util/flowcontrol/metrics/sample_and_watermark.go#L185 will go through a _lot_ of iterations, and do so with a couple of locks held.

No big hiccups in any scenario.

**Environment**:
- Kubernetes version (use `kubectl version`): `master` branch
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

MikeSpreitzer:
**What type of PR is this?**
/kind bug

This PR fixes two bugs.  The more important one is #94145.  The fix is to add a goroutine that periodically (every 10 seconds) prods each QueueSet to make sure its TimedObserver pair is up-to-date.

This PR also fixes the comment called out in #94122 

Fixes #94145
Fixes #94122

This bug fix should get into release 1.19.

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/server/config.go
+++ b/staging/src/k8s.io/apiserver/pkg/server/config.go
@@ -626,6 +626,7 @@ func (c completedConfig) New(name string, delegationTarget DelegationTarget) (*G
 	if s.isPostStartHookRegistered(priorityAndFairnessConfigConsumerHookName) {
 	} else if c.FlowControl != nil {
 		err := s.AddPostStartHook(priorityAndFairnessConfigConsumerHookName, func(context PostStartHookContext) error {
+			go c.FlowControl.MaintainObservations(context.StopCh)
 			go c.FlowControl.Run(context.StopCh)
 			return nil
 		})
--- a/staging/src/k8s.io/apiserver/pkg/util/flowcontrol/apf_controller.go
+++ b/staging/src/k8s.io/apiserver/pkg/util/flowcontrol/apf_controller.go
@@ -227,6 +227,23 @@ func (cfgCtlr *configController) initializeConfigController(informerFactory kube
 		}})
 }
 
+// MaintainObservations keeps the observers from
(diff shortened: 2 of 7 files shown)


MikeSpreitzer:

MikeSpreitzer:
/cc @deads2k 
@rngy 
@bjrara 
@lavalamp

MikeSpreitzer:
@MikeSpreitzer

bjrara:
I'm thinking if the same risk exists in the apf filter as well. If apiservers are deployed as a cluster, and one of the instances doesn't serve any request for sometime, the iteration will block the requests from processing, will it?

PR #94147, opened 2020-08-20
Update json-patch to v4.9.0 tagged release

liggitt:
**What type of PR is this?**

/kind cleanup

Updates to a tagged version of json-patch. This is a no-op for k8s.io/kubernetes (no code changes), but ensures that consumers of the k8s.io libraries pick up the desired version (the v0.0.0... version would allow a downstream consumer to use v4.6.0 without realizing it and miss the desired fixes).

```release-note
NONE
```

/sig api-machinery
/cc @dims


liggitt:
lint caught an unused dep that needed removing from modules.txt

dims (approved):

PR #94150, opened 2020-08-20
[1.16] Update to json-patch 4.9.0

liggitt:
**What type of PR is this?**
/kind cleanup

Includes the same version bump as #94147 on master.

Does not change code, but ensures go module consumers pick up the correct version of json-patch containing fixes.

```release-note
NONE
```

/cc @dims


liggitt:
lint caught an unused dep I needed to drop

fedebongio:
@liggitt curious why does not show cherrypick in the title?

liggitt:
it was manually done, dependency updates don't cherry-pick cleanly

PR #88591, opened 2020-02-26
kubelet: Avoid sending no-op patches

smarterclayton:
In an e2e run, out of 1857 pod status updates executed by the Kubelet 453 (25%) were no-ops - they only contained the UID of the pod and no status changes. If the patch is a no-op we can
avoid invoking the server and continue.

Should reduce load on the API Server in large clusters.

/kind bug

```release-note
Kubelets perform fewer unnecessary pod status update operations on the API server.
```

PR #42662, opened 2017-03-07
Print dereferenced pod status fields when logging status update

wongma7:
Before: "Terminated:0xc421932af0"
After:"Terminated:&ContainerStateTerminated{ExitCode:0,Signal:0,Reason:Completed,Message:,StartedAt:0001-01-01 00:00:00 +0000 UTC,FinishedAt:2017-03-07 14:50:48 -0500 EST,ContainerID:docker://bd453bb969264b3ace2b3934a568af7679a0d51fee543a5f8a82429ff654970e,}"

"Ignoring same status for pod" messages already print status fully, these "Status for pod updated" messages should too IMO

```release-note
NONE
```

PR #94151, opened 2020-08-20
kubelet/status: don't produce confusing logs

kolyshkin:
/kind bug

The debug log statement in syncPod was initially added by commit 1dabce981591
(PR #42662), and most recently modified by commit b2528654797e2
(PR #88591).

The problem is, some of the status.status sub-structures may contain
multi-line fields that are printed on a separate line (see example
at [1]) which is very confusing.

For now, let's just print status.Phase. If better debug info is needed,
I suggest it should be implemented as a custom DebugString() method
of PodStatus, which should make sure that those fields that could
contain newlines are either not printed, or the newlines are escaped.

Fix the debug log statement in updateStatusInternal the same way.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1821730

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1821730

```release-note
NONE
```

### DIFF
--- a/pkg/kubelet/status/status_manager.go
+++ b/pkg/kubelet/status/status_manager.go
@@ -426,7 +426,7 @@ func (m *manager) updateStatusInternal(pod *v1.Pod, status v1.PodStatus, forceUp
 	// The intent here is to prevent concurrent updates to a pod's status from
 	// clobbering each other so the phase of a pod progresses monotonically.
 	if isCached && isPodStatusByKubeletEqual(&cachedStatus.status, &status) && !forceUpdate {
-		klog.V(3).Infof("Ignoring same status for pod %q, status: %+v", format.Pod(pod), status)
+		klog.V(3).Infof("Ignoring same status for pod %q, status.Phase: %s", format.Pod(pod), status.Phase)
 		return false // No new status.
 	}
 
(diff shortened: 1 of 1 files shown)


stefins (approved):

pkg/kubelet/status/status_manager.go
```diff
@@ -426,7 +426,7 @@ func (m *manager) updateStatusInternal(pod *v1.Pod, status v1.PodStatus, forceUp
 	// The intent here is to prevent concurrent updates to a pod's status from
 	// clobbering each other so the phase of a pod progresses monotonically.
 	if isCached && isPodStatusByKubeletEqual(&cachedStatus.status, &status) && !forceUpdate {
-		klog.V(3).Infof("Ignoring same status for pod %q, status: %+v", format.Pod(pod), status)
+		klog.V(3).Infof("Ignoring same status for pod %q, status.Phase: %s", format.Pod(pod), status.Phase)
```
SergeyKanzhelev, line 429: one more place in this file: https://github.com/kubernetes/kubernetes/blob/9a497c8be6edf04d3b406fd4015f375be066c2ae/pkg/kubelet/status/status_manager.go#L166-L167

SergeyKanzhelev: And here:

https://github.com/kubernetes/kubernetes/blob/9a497c8be6edf04d3b406fd4015f375be066c2ae/pkg/kubelet/status/status_manager.go#L443-L451

pkg/kubelet/status/status_manager.go
```diff
@@ -567,9 +567,9 @@ func (m *manager) syncPod(uid types.UID, status versionedPodStatus) {
 		return
 	}
 	if unchanged {
-		klog.V(3).Infof("Status for pod %q is up-to-date: (%d)", format.Pod(pod), status.version)
+		klog.V(3).Infof("Status for pod %q is up-to-date: (version: %d)", format.Pod(pod), status.version)
 	} else {
-		klog.V(3).Infof("Status for pod %q updated successfully: (%d, %+v)", format.Pod(pod), status.version, status.status)
+		klog.V(3).Infof("Status for pod %q updated successfully: (version: %d, phase: %s)", format.Pod(pod), status.version, status.status.Phase)
```
SergeyKanzhelev, line 572: this may not be ideal to troubleshoot the pod readiness ++ feature where pod condition is set from the API server and than used to evaluate pod readiness. https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-gate

As you mentioned in description, adding a formatter may be a good way to introduce some logic on what is important in podPhase and what is less important

kolyshkin:
I retract my PR. The more proper way would be to fix formatting of the messages (escape newlines at least).

PR #89582, opened 2020-03-27
Add SCTP Network Policy tests

fedepaol:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

**What this PR does / why we need it**:

This PR completes the checklist from [https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0015-20180614-SCTP-support.md#sctp-connectivity-tests](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0015-20180614-SCTP-support.md#sctp-connectivity-tests), testing network policies with SCTP.

This is achieved by adding SCTP support to both the `agnhost` `connect` command and `porter` command, and to use then in the network policy tests to check if network policies are working.
This is complementary to https://github.com/kubernetes/kubernetes/pull/87926

In order to maintain the ` [Feature:SCTPConnectivity]` and have the sctp tests optional, I started converting the current policy tests in a tabular form. To do that (and to maintain sctp optional), I removed the `BeforeEach` function and converted it to a custom one invoked with the different protocols in each table body.
Not sure if this is the right approach but I could not find a better one that prevents duplication while still allowing us to skip the SCTP tests.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

NONE

```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

[KEP] : https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0015-20180614-SCTP-support.md

```

PR #94152, opened 2020-08-21
Agnhost version sync

BenTheElder:
**What type of PR is this?**

/kind cleanup

Fixes #

I also:
- made the cobra.Command in agnhost.go multiline for easier matching on the right field
- updated the out of sync agnhost source image versions

```release-note
NONE
```

```docs

```


BenTheElder:
these have de-synced multiple times. most recently caught in https://github.com/kubernetes/kubernetes/pull/94135#discussion_r474162724

build/dependencies.yaml
```diff
@@ -1,4 +1,20 @@
...
+  - name: "agnhost"
+    version: "2.22"
+    refPaths:
+    - path: test/images/agnhost/VERSION
+      match: \d.\d
+    - path: test/images/agnhost/agnhost.go
+      match: "Version:"
+
+  # then after merge and successful postsubmit image push / promotion, bump this
+  - name: "agnhost: dependents"
```
BenTheElder, line 12: I suspect there are more, but I didn't turn up any that are obviously intended to be in sync here yet.
If nothing else though, this entry should prevent an incorrect addition of this file to the entry above.

There is also the rarely updated private image from this that we literally only use to test pulling a private image (w/ auth).

spiffxp:
This missed updating test/utils/image/manifest.go, suggested update in https://github.com/kubernetes/kubernetes/pull/89582/files#r480392573

Issue #93693, opened 2020-08-04
[Failing Test]  sig-network-ingress-gce


spencerhance:
**Which test(s) are failing**:

* ~~[sig-network] Loadbalancing: L7 GCE [Slow] [Feature:Ingress] should support multiple TLS certs~~

* ~~[sig-network] Loadbalancing: L7 GCE [Slow] [Feature:Ingress] multicluster ingress should get instance group annotation~~
* ~~[sig-network] Loadbalancing: L7 GCE [Slow] [Feature:Ingress] should conform to Ingress spec~~
* ~~[sig-network] Loadbalancing: L7 GCE [Slow] [Feature:Ingress] should create ingress with pre-shared certificate~~
* ~~[sig-network] Loadbalancing: L7 GCE [Slow] [Feature:NEG] rolling update backend pods should not cause service disruption~~
* ~~[sig-network] Loadbalancing: L7 GCE [Slow] [Feature:NEG] should be able to create a ClusterIP service~~
* ~~[sig-network] Loadbalancing: L7 GCE [Slow] [Feature:NEG] should be able to switch between IG and NEG modes~~
* ~~[sig-network] Loadbalancing: L7 GCE [Slow] [Feature:NEG] should conform to Ingress spec~~
* ~~[sig-network] Loadbalancing: L7 GCE [Slow] [Feature:NEG] should create NEGs for all ports with the Ingress annotation, and NEGs for the standalone annotation otherwise~~
* ~~[sig-network] Loadbalancing: L7 GCE [Slow] [Feature:NEG] should sync endpoints for both Ingress-referenced NEG and standalone NEG~~
* ~~[sig-network] Loadbalancing: L7 GCE [Slow] [Feature:NEG] should sync endpoints to NEG~~

**Since when has it been failing**:
The full testgrid history is red, but it has been failing at least since 5/29

**Testgrid link**:

https://testgrid.k8s.io/sig-network-ingress-gce-e2e#ingress-gce-e2e&width=5

**Reason for failure**:

```
FAIL: error in waiting for ingress to get an address: Ingress failed to acquire an IP address within 10m0s
```

**Anything else we need to know**:

Originally these failures were due to an rbac permission issue (see: #93031), but now that that has been resolved, the tests are still not passing.

liggitt:
adding to the milestone for investigation, since ingress was touched this release

/milestone v1.19

athenabot:
/triage unresolved

[...]

PR #94153, opened 2020-08-21
[Failing Test] Add rbac patch permissions for system:controller:glbc

spencerhance:
/kind failing-test

Issue: #93693

**What this PR does / why we need it:**

Adds Patch permissions for system:controller:glbc.  This should fix the test.

```release-note
NONE
```

### DIFF
--- a/cluster/addons/rbac/cluster-loadbalancing/glbc/roles.yaml
+++ b/cluster/addons/rbac/cluster-loadbalancing/glbc/roles.yaml
@@ -43,7 +43,7 @@ rules:
   verbs: ["update"]
 - apiGroups: ["extensions", "networking.k8s.io"]
   resources: ["ingresses/status"]
-  verbs: ["update"]
+  verbs: ["update", "patch"]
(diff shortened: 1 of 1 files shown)


cluster/addons/rbac/cluster-loadbalancing/glbc/roles.yaml
```diff
@@ -43,7 +43,7 @@ rules:
   verbs: ["update"]
 - apiGroups: ["extensions", "networking.k8s.io"]
   resources: ["ingresses/status"]
-  verbs: ["update"]
+  verbs: ["patch"]
```
liggitt, line 46: leave update in place as well. if you have one, you effectively have the other, and leaving both in place allows a smooth transition from a previous version which used update

spencerhance: Good call!  Done

liggitt:
/lgtm
/approve
/ok-to-test

/milestone v1.19
fixes test failure in the milestone

PR #94156, opened 2020-08-21
Add simple heuristic to select AccessMode for mounting

nixpanic:
**What type of PR is this?**

/sig storage
/kind cleanup

Include some simple heuristic to decide which AccessMode to use for the
CSI NodeStage request.

1. default to ReadWriteOnce
2. if the PVC Spec has AccessModes, use the 1st as default
3. for ReadOnly volumes, use ReadOnlyMany if it is in the PCV Spec
4. for non-ReadOnly volumes, use the first non-ReadOnlyMany mode from
   the PVC Spec

This should solve issues when an initial Pod populates a PVC in
ReadWriteOnce, and exists. Follow-up Pods with the volume in ReadOnly
mode will use ReadOnlyMany to allow running the Pods on different nodes.

This was listed as a `TODO` with pointer to @vladimirvivien

```release-note
NONE
```

### DIFF
--- a/pkg/volume/csi/csi_attacher.go
+++ b/pkg/volume/csi/csi_attacher.go
@@ -354,10 +354,27 @@ func (c *csiAttacher) MountDevice(spec *volume.Spec, devicePath string, deviceMo
 		return nil
 	}
 
-	//TODO (vladimirvivien) implement better AccessModes mapping between k8s and CSI
 	accessMode := v1.ReadWriteOnce
 	if spec.PersistentVolume.Spec.AccessModes != nil {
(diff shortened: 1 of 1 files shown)


nixpanic:
/remove-lifecycle rotten

Going to rebase and address CI issues later today.

msau42:
/assign @chrishenzie 
It'll be interesting to see if this works with gce pd, which also requires attach to be readonly.

chrishenzie:
/remove-lifecycle rotten

@humblec Just checking if this feature is implemented or something else. I remember you did some work on dynamic use of ReadOnly / access modes

PR #94165, opened 2020-08-21
Use separate pathSpec for local and remote to properly handle cleaning paths

soltysh:
**What type of PR is this?**
/kind bug

~~We should be consistently be using https://golang.org/pkg/path/filepath/ instead of https://golang.org/pkg/path/ when working with paths.~~

This PR defines `remotePath` and `localPath` type which properly handles os-related operations such as cleaning, joining paths, etc. 

/assign @sallyom @seans3 

```release-note
NONE
```

/sig cli
/priority important-longterm
/milestone v1.19

### DIFF
--- a/staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
+++ b/staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
@@ -24,11 +24,8 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"path"
-	"path/filepath"
 	"strings"
 
-	"github.com/lithammer/dedent"
 	"github.com/spf13/cobra"
 
 	"k8s.io/cli-runtime/pkg/genericclioptions"
@@ -67,12 +64,6 @@ var (
 
 		# Copy /tmp/foo from a remote pod to /tmp/bar locally
 		kubectl cp <some-namespace>/<some-pod>:/tmp/foo /tmp/bar`))
-
-	cpUsageStr = dedent.Dedent(`
-		expected 'cp <file-spec-src> <file-spec-dest> [-c container]'.
-		<file-spec> is:
-		[namespace/]pod-name:/file/path for a remote file
-		/file/path for a local file`)
 )
 
 // CopyOptions have the data required to perform the copy operation
@@ -158,6 +149,7 @@ func NewCmdCp(f cmdutil.Factory, ioStreams genericclioptions.IOStreams) *cobra.C
 		},
 		Run: func(cmd *cobra.Command, args []string) {
 			cmdutil.CheckErr(o.Complete(f, cmd))
+			cmdutil.CheckErr(o.Validate(cmd, args))
 			cmdutil.CheckErr(o.Run(args))
 		},
 	}
@@ -167,26 +159,21 @@ func NewCmdCp(f cmdutil.Factory, ioStreams genericclioptions.IOStreams) *cobra.C
 	return cmd
 }
 
-type fileSpec struct {
-	PodNamespace string
-	PodName      string
-	File         string
-}
-
 var (
 	errFileSpecDoesntMatchFormat = errors.New("filespec must match the canonical format: [[namespace/]pod:]file/path")
-	errFileCannotBeEmpty         = errors.New("filepath can not be empty")
 )
 
 func extractFileSpec(arg string) (fileSpec, error) {
 	i := strings.Index(arg, ":")
 
-	if i == -1 {
-		return fileSpec{File: arg}, nil
-	}
 	// filespec starting with a semicolon is invalid
 	if i == 0 {
 		return fileSpec{}, errFileSpecDoesntMatchFormat
+	}
+	if i == -1 {
+		return fileSpec{
+			File: newLocalPath(arg),
+		}, nil
 	}
 
 	pod, file := arg[:i], arg[i+1:]
@@ -195,13 +182,13 @@ func extractFileSpec(arg string) (fileSpec, error) {
 	case 1:
 		return fileSpec{
 			PodName: pieces[0],
-			File:    file,
+			File:    newRemotePath(file),
 		}, nil
 	case 2:
 		return fileSpec{
 			PodNamespace: pieces[0],
 			PodName:      pieces[1],
-			File:         file,
+			File:         newRemotePath(file),
 		}, nil
 	default:
 		return fileSpec{}, errFileSpecDoesntMatchFormat
@@ -235,16 +222,13 @@ func (o *CopyOptions) Complete(f cmdutil.Factory, cmd *cobra.Command) error {
 // Validate makes sure provided values for CopyOptions are valid
 func (o *CopyOptions) Validate(cmd *cobra.Command, args []string) error {
 	if len(args) != 2 {
-		return cmdutil.UsageErrorf(cmd, cpUsageStr)
+		return fmt.Errorf("source and destination are required")
 	}
 	return nil
 }
 
 // Run performs the execution
 func (o *CopyOptions) Run(args []string) error {
-	if len(args) < 2 {
-		return fmt.Errorf("source and destination are required")
-	}
 	srcSpec, err := extractFileSpec(args[0])
 	if err != nil {
 		return err
@@ -256,6 +240,9 @@ func (o *CopyOptions) Run(args []string) error {
 
 	if len(srcSpec.PodName) != 0 && len(destSpec.PodName) != 0 {
 		return fmt.Errorf("one of src or dest must be a local file specification")
+	}
+	if len(srcSpec.File.String()) == 0 || len(destSpec.File.String()) == 0 {
+		return errors.New("filepath can not be empty")
 	}
 
 	if len(srcSpec.PodName) != 0 {
@@ -283,7 +270,7 @@ func (o *CopyOptions) checkDestinationIsDir(dest fileSpec) error {
 			PodName:   dest.PodName,
 		},
 
-		Command:  []string{"test", "-d", dest.File},
+		Command:  []string{"test", "-d", dest.File.String()},
 		Executor: &exec.DefaultRemoteExecutor{},
 	}
 
@@ -291,29 +278,24 @@ func (o *CopyOptions) checkDestinationIsDir(dest fileSpec) error {
 }
 
 func (o *CopyOptions) copyToPod(src, dest fileSpec, options *exec.ExecOptions) error {
-	if len(src.File) == 0 || len(dest.File) == 0 {
-		return errFileCannotBeEmpty
-	}
-	if _, err := os.Stat(src.File); err != nil {
+	if _, err := os.Stat(src.File.String()); err != nil {
 		return fmt.Errorf("%s doesn't exist in local filesystem", src.File)
 	}
 	reader, writer := io.Pipe()
 
-	// strip trailing slash (if any)
-	if dest.File != "/" && strings.HasSuffix(string(dest.File[len(dest.File)-1]), "/") {
-		dest.File = dest.File[:len(dest.File)-1]
-	}
+	srcFile := src.File.(localPath)
+	destFile := dest.File.(remotePath)
 
 	if err := o.checkDestinationIsDir(dest); err == nil {
 		// If no error, dest.File was found to be a directory.
 		// Copy specified src into it
-		dest.File = dest.File + "/" + path.Base(src.File)
+		destFile = destFile.Join(srcFile.Base())
 	}
 
-	go func() {
+	go func(src localPath, dest remotePath, writer io.WriteCloser) {
 		defer writer.Close()
-		cmdutil.CheckErr(makeTar(src.File, dest.File, writer))
-	}()
+		cmdutil.CheckErr(makeTar(src, dest, writer))
+	}(srcFile, destFile, writer)
 	var cmdArr []string
 
 	// TODO: Improve error messages by first testing if 'tar' is present in the container?
@@ -322,9 +304,9 @@ func (o *CopyOptions) copyToPod(src, dest fileSpec, options *exec.ExecOptions) e
 	} else {
 		cmdArr = []string{"tar", "-xmf", "-"}
 	}
-	destDir := path.Dir(dest.File)
-	if len(destDir) > 0 {
-		cmdArr = append(cmdArr, "-C", destDir)
+	destFileDir := destFile.Dir().String()
+	if len(destFileDir) > 0 {
+		cmdArr = append(cmdArr, "-C", destFileDir)
 	}
 
 	options.StreamOptions = exec.StreamOptions{
@@ -345,10 +327,6 @@ func (o *CopyOptions) copyToPod(src, dest fileSpec, options *exec.ExecOptions) e
 }
 
 func (o *CopyOptions) copyFromPod(src, dest fileSpec) error {
-	if len(src.File) == 0 || len(dest.File) == 0 {
-		return errFileCannotBeEmpty
-	}
-
 	reader, outStream := io.Pipe()
 	options := &exec.ExecOptions{
 		StreamOptions: exec.StreamOptions{
@@ -363,7 +341,7 @@ func (o *CopyOptions) copyFromPod(src, dest fileSpec) error {
 		},
 
 		// TODO: Improve error messages by first testing if 'tar' is present in the container?
-		Command:  []string{"tar", "cf", "-", src.File},
+		Command:  []string{"tar", "cf", "-", src.File.String()},
 		Executor: &exec.DefaultRemoteExecutor{},
 	}
 
@@ -371,49 +349,28 @@ func (o *CopyOptions) copyFromPod(src, dest fileSpec) error {
 		defer outStream.Close()
 		cmdutil.CheckErr(o.execute(options))
 	}()
-	prefix := getPrefix(src.File)
-	prefix = path.Clean(prefix)
+
+	srcFile := src.File.(remotePath)
+	destFile := dest.File.(localPath)
+
 	// remove extraneous path shortcuts - these could occur if a path contained extra "../"
 	// and attempted to navigate beyond "/" in a remote filesystem
-	prefix = stripPathShortcuts(prefix)
-	return o.untarAll(src, reader, dest.File, prefix)
+	prefix := stripPathShortcuts(srcFile.StripSlashes().Clean().String())
+	return o.untarAll(src.PodNamespace, src.PodName, prefix, srcFile, destFile, reader)
 }
 
-// stripPathShortcuts removes any leading or trailing "../" from a given path
-func stripPathShortcuts(p string) string {
-	newPath := path.Clean(p)
-	trimmed := strings.TrimPrefix(newPath, "../")
-
-	for trimmed != newPath {
-		newPath = trimmed
-		trimmed = strings.TrimPrefix(newPath, "../")
-	}
-
-	// trim leftover {".", ".."}
-	if newPath == "." || newPath == ".." {
-		newPath = ""
-	}
-
-	if len(newPath) > 0 && string(newPath[0]) == "/" {
-		return newPath[1:]
-	}
-
-	return newPath
-}
-
-func makeTar(srcPath, destPath string, writer io.Writer) error {
+func makeTar(src localPath, dest remotePath, writer io.Writer) error {
 	// TODO: use compression here?
 	tarWriter := tar.NewWriter(writer)
 	defer tarWriter.Close()
 
-	srcPath = path.Clean(srcPath)
-	destPath = path.Clean(destPath)
-	return recursiveTar(path.Dir(srcPath), path.Base(srcPath), path.Dir(destPath), path.Base(destPath), tarWriter)
+	srcPath := src.Clean()
+	destPath := dest.Clean()
+	return recursiveTar(srcPath.Dir(), srcPath.Base(), destPath.Dir(), destPath.Base(), tarWriter)
 }
 
-func recursiveTar(srcBase, srcFile, destBase, destFile string, tw *tar.Writer) error {
-	srcPath := path.Join(srcBase, srcFile)
-	matchedPaths, err := filepath.Glob(srcPath)
+func recursiveTar(srcDir, srcFile localPath, destDir, destFile remotePath, tw *tar.Writer) error {
+	matchedPaths, err := srcDir.Join(srcFile).Glob()
 	if err != nil {
 		return err
 	}
@@ -430,13 +387,14 @@ func recursiveTar(srcBase, srcFile, destBase, destFile string, tw *tar.Writer) e
 			if len(files) == 0 {
 				//case empty directory
 				hdr, _ := tar.FileInfoHeader(stat, fpath)
-				hdr.Name = destFile
+				hdr.Name = destFile.String()
 				if err := tw.WriteHeader(hdr); err != nil {
 					return err
 				}
 			}
 			for _, f := range files {
-				if err := recursiveTar(srcBase, path.Join(srcFile, f.Name()), destBase, path.Join(destFile, f.Name()), tw); err != nil {
+				if err := recursiveTar(srcDir, srcFile.Join(newLocalPath(f.Name())),
+					destDir, destFile.Join(newRemotePath(f.Name())), tw); err != nil {
 					return err
 				}
 			}
@@ -450,7 +408,7 @@ func recursiveTar(srcBase, srcFile, destBase, destFile string, tw *tar.Writer) e
 			}
 
 			hdr.Linkname = target
-			hdr.Name = destFile
+			hdr.Name = destFile.String()
 			if err := tw.WriteHeader(hdr); err != nil {
 				return err
 			}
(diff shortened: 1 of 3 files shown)


hasheddan:
@soltysh does this need to be in v1.19? If so, who do we need for lgtm?

liggitt:
/milestone v1.20

This is late for 1.19, can resolve in 1.19.x if needed

lukehinds:
where are we with this ? as per #94165  `filepath` is problematic for windows, so it sounds like we need to approach this fix with a different approach?

soltysh:
> where are we with this ? as per #94165 `filepath` is problematic for windows, so it sounds like we need to approach this fix with a different approach?

I'm verifying all the edge cases and should have the answer today/tomorrow at latest.

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
```diff
@@ -138,7 +137,7 @@ func extractFileSpec(arg string) (fileSpec, error) {
 	}
 
 	pod, file := arg[:i], arg[i+1:]
-	pieces := strings.Split(pod, "/")
+	pieces := strings.Split(pod, string(filepath.Separator))
```
liggitt, line 140: this is not right, the pod name is either `<pod>` or `<namespace>/<pod>`, so this should continue to split on "/"

See the documented example:
```
kubectl cp <some-namespace>/<some-pod>:/tmp/foo /tmp/bar
```

If a test did not catch this, then we need to add a test explicitly for this case.

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
```diff
@@ -248,14 +247,14 @@ func (o *CopyOptions) copyToPod(src, dest fileSpec, options *exec.ExecOptions) e
 	reader, writer := io.Pipe()
 
 	// strip trailing slash (if any)
-	if dest.File != "/" && strings.HasSuffix(string(dest.File[len(dest.File)-1]), "/") {
+	if dest.File != string(filepath.Separator) && strings.HasSuffix(string(dest.File[len(dest.File)-1]), string(filepath.Separator)) {
```
liggitt, line 250: can we strip both `/` and `\` characters from the end?

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
```diff
@@ -331,20 +330,20 @@ func (o *CopyOptions) copyFromPod(src, dest fileSpec) error {
 
 // stripPathShortcuts removes any leading or trailing "../" from a given path
 func stripPathShortcuts(p string) string {
-	newPath := path.Clean(p)
-	trimmed := strings.TrimPrefix(newPath, "../")
+	newPath := filepath.Clean(p)
+	trimmed := strings.TrimPrefix(newPath, fmt.Sprintf("..%c", filepath.Separator))
```
liggitt, line 334: can we drop both `../` and `..\` prefixes?

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
```diff
@@ -331,20 +330,20 @@ func (o *CopyOptions) copyFromPod(src, dest fileSpec) error {
...
 
 	for trimmed != newPath {
 		newPath = trimmed
-		trimmed = strings.TrimPrefix(newPath, "../")
+		trimmed = strings.TrimPrefix(newPath, fmt.Sprintf("..%c", filepath.Separator))
```
liggitt, line 338: can we drop both `../` and `..\` prefixes?

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
```diff
@@ -331,20 +330,20 @@ func (o *CopyOptions) copyFromPod(src, dest fileSpec) error {
...
 	}
 
 	// trim leftover {".", ".."}
 	if newPath == "." || newPath == ".." {
 		newPath = ""
 	}
 
-	if len(newPath) > 0 && string(newPath[0]) == "/" {
+	if len(newPath) > 0 && string(newPath[0]) == string(filepath.Separator) {
```
liggitt, line 346: can we drop both `/` and `\` prefixes?

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
```diff
@@ -510,7 +509,7 @@ func isDestRelative(base, dest string) bool {
 
 func getPrefix(file string) string {
 	// tar strips the leading '/' if it's there, so we will too
-	return strings.TrimLeft(file, "/")
+	return strings.TrimLeft(file, string(filepath.Separator))
```
liggitt, line 512: can we drop both "/" and "\"?

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp_test.go
```diff
@@ -307,13 +306,13 @@ func TestTarUntar(t *testing.T) {
 		},
 		{
 			name:     "relative_to_dest",
-			data:     path.Join(dir2, "foo"),
```
liggitt, line 310: are these mimicking the data that would come from a tar file? if so, shouldn't we duplicate the testcases to continue exercising "/" (which is the standard tar format) and add testcases with "\" (which is the windows-targeted case this PR is fixing)?

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
soltysh: We have a test here: https://github.com/kubernetes/kubernetes/blob/28b46be97bb453be2e00eaf9e35ea1d6ecd41861/staging/src/k8s.io/kubectl/pkg/cmd/cp/cp_test.go#L55-L112 but since we're running that on a linux machine it works just fine.

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp_test.go
soltysh: I've dug into it, and there's a problem that each OS treats the other separators differently so we can't have reliable test. Iow.:
- For linux `\` is valid character that can be part of the name, so `my\dir` creates a single directory named just that. For `/` you'll get `dir` inside `my`, which is expected.
- For windows either approach `my/dir` or `my\dir` will create `dir` inside `my`. 
This means I can't reliable write windows test and execute this on linux and in the other direction when using `filepath` module which handles that for us. We'd need to manually handle linux and windows separately. So for this particular one we'll have to rely on manual testing unfortunately :disappointed:

soltysh:
@liggitt I've addressed comments, where possible and left comments in other cases.

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp_test.go
liggitt: that's sort of why I was wondering if we should drop both `/` and `\` characters

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
```diff
@@ -509,8 +511,9 @@ func isDestRelative(base, dest string) bool {
 }
 
 func getPrefix(file string) string {
-	// tar strips the leading '/' if it's there, so we will too
-	return strings.TrimLeft(file, "/")
+	// tar strips the leading '/' and '\\' if it's there, so we will too
+	prefix := strings.TrimLeft(file, "/")
```
liggitt, line 515: should be `strings.TrimLeft(file, "/\\")`, the current code wouldn't modify `\/foo`

MonzElmasry:
Hello! Bug Triage here :wave: This has not been updated for a long time, so I'd like to check what's the status. The code freeze is starting 12.11.2020 (about a weeks and half from now) and while there is still plenty of time, we want to ensure that each PR has a chance to be merged on time.

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
```diff
@@ -247,15 +246,16 @@ func (o *CopyOptions) copyToPod(src, dest fileSpec, options *exec.ExecOptions) e
 	}
 	reader, writer := io.Pipe()
 
-	// strip trailing slash (if any)
-	if dest.File != "/" && strings.HasSuffix(string(dest.File[len(dest.File)-1]), "/") {
+	// strip trailing slash (if any) both unix and windows style
+	if (dest.File != "/" && strings.HasSuffix(string(dest.File[len(dest.File)-1]), "/")) ||
+		(dest.File != "\\" && strings.HasSuffix(string(dest.File[len(dest.File)-1]), "\\")) {
```
liggitt, line 251: nit: use ``` `\` ``` throughout to avoid needing double-slashes

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
```diff
@@ -247,15 +246,16 @@ func (o *CopyOptions) copyToPod(src, dest fileSpec, options *exec.ExecOptions) e
...
 		dest.File = dest.File[:len(dest.File)-1]
 	}
 
 	if err := o.checkDestinationIsDir(dest); err == nil {
 		// If no error, dest.File was found to be a directory.
 		// Copy specified src into it
-		dest.File = dest.File + "/" + path.Base(src.File)
+		dest.File = filepath.Join(dest.File, filepath.Base(src.File))
```
liggitt, line 258: in `copyToPod`, isn't our destination assumed to be a linux container? if so, won't this set dest.File to windows path when the `cp` command is run on windows, and break the command constructed below?

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
```diff
@@ -271,7 +271,7 @@ func (o *CopyOptions) copyToPod(src, dest fileSpec, options *exec.ExecOptions) e
 	} else {
 		cmdArr = []string{"tar", "-xmf", "-"}
 	}
-	destDir := path.Dir(dest.File)
+	destDir := filepath.Dir(dest.File)
```
liggitt, line 274: this means we'll send a command to (presumably) a linux container containing (potentially) a windows destDir... that doesn't seem like it would work

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
```diff
@@ -322,7 +322,7 @@ func (o *CopyOptions) copyFromPod(src, dest fileSpec) error {
 		cmdutil.CheckErr(err)
 	}()
 	prefix := getPrefix(src.File)
-	prefix = path.Clean(prefix)
+	prefix = filepath.Clean(prefix)
```
liggitt, line 325: if we're copying from a pod, isn't src.File a linux path? if so, will filepath.Clean do the right thing when `cp` is run on windows?

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
```diff
@@ -509,8 +511,8 @@ func isDestRelative(base, dest string) bool {
 }
 
 func getPrefix(file string) string {
-	// tar strips the leading '/' if it's there, so we will too
-	return strings.TrimLeft(file, "/")
+	// tar strips the leading '/' and '\\' if it's there, so we will too
+	return strings.TrimLeft(file, "/\\")
```
liggitt, line 515: ``` `/\` ```

liggitt (commented):
It is not clear to me which OS separators src.File and dest.File will contain when running `cp` on a windows client against a linux container. How can we exercise the various cases in tests to ensure this works properly?

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
soltysh: We want to ensure that the client-side cleaning is relevant for their particular system. Actually this should do both unix and windows cleaning, I'll probably replace all windows and proceed with processing only linux-style separator.

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
soltysh: Theoretically - yes, but we need to ensure that if either is somehow injected we can fix that. I'll actually revert my changes to work only with linux style separator and will add replacing at the beginning of this function.

soltysh:
> It is not clear to me which OS separators src.File and dest.File will contain when running `cp` on a windows client against a linux container. How can we exercise the various cases in tests to ensure this works properly?

It's tricky, but I think that with latest code changes I've manged to add some cases for windows-specific paths. ptal

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp_test.go
```diff
@@ -290,7 +332,7 @@ func TestTarUntar(t *testing.T) {
 			fileType: RegularFile,
 		},
 		{
-			name:     "some/other/directory/",
+			name:     "l295some/other/directory",
```
liggitt, line 335: was the l295 prefix intentional? should we add a test case without a trailing slash rather than modify this testcase?

liggitt (commented):
I think this looks good, had one question on one of the tests. Would like a second set of eyes since it is pretty twisty to follow.

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp_test.go
soltysh: yikes, not intentional, fixing...

liggitt:
/cc @tallclair 

would like extra eyes on this

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
```diff
@@ -356,13 +368,13 @@ func makeTar(srcPath, destPath string, writer io.Writer) error {
...
+	destPath = filepath.Clean(destPath)
+	return recursiveTar(filepath.Dir(srcPath), filepath.Base(srcPath), filepath.Dir(destPath), filepath.Base(destPath), tarWriter)
 }
 
 func recursiveTar(srcBase, srcFile, destBase, destFile string, tw *tar.Writer) error {
-	srcPath := path.Join(srcBase, srcFile)
+	srcPath := filepath.Join(srcBase, srcFile)
 	matchedPaths, err := filepath.Glob(srcPath)
 	if err != nil {
 		return err
```
liggitt, line 380: does this change mean `matchedPaths` could be `\`-separated on Windows? do we need to ensure the path passed to `tar.FileInfoHeader(stat, fpath)` is a unix path?

tallclair:
I don't see any issues, but here's a suggestion to make this code easier to reason about:
- declare a separate type for `localPath` and `remotePath`
- those types should have methods for file operations on the path (clean, join, etc.). Local should use filepath, remote should use path where relevant. Probably want a `func parse{Local,Remote}(path string) {local,remote}Path` too.
- if there are methods that need to handle both path types, you could also add a `universalPath` interface.

soltysh:
@tallclair @liggitt this is updated per proposed remote vs local path spec

staging/src/k8s.io/kubectl/pkg/cmd/cp/filespec.go
```diff
@@ -0,0 +1,192 @@
...
+
+import (
+	"path"
+	"path/filepath"
+	"strings"
+)
+
+type fileSpec struct {
+	PodName      string
+	PodNamespace string
```
tallclair, line 27: (optional) If it's not too difficult, it probably makes sense to split this into a remote & local fileSpec, since local files won't have a pod name & namespace. This would make the method signatures for methods like `copyToPod` a bit easier to understand too.

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
```diff
@@ -231,31 +225,23 @@ func (o *CopyOptions) checkDestinationIsDir(dest fileSpec) error {
...
-	// strip trailing slash (if any)
-	if dest.File != "/" && strings.HasSuffix(string(dest.File[len(dest.File)-1]), "/") {
-		dest.File = dest.File[:len(dest.File)-1]
-	}
-
 	if err := o.checkDestinationIsDir(dest); err == nil {
 		// If no error, dest.File was found to be a directory.
 		// Copy specified src into it
-		dest.File = dest.File + "/" + path.Base(src.File)
+		dest.File = newRemotePath(dest.File.String() + "/" + src.File.Base().String())
```
tallclair, line 244:
```suggestion
		dest.File = dest.File.JoinElem(src.File.Base()))
```

staging/src/k8s.io/kubectl/pkg/cmd/cp/filespec.go
```diff
@@ -0,0 +1,192 @@
...
+	return newLocalPath(stripLeadingSlash(p.file))
+}
+
+func (p localPath) StripShortcuts() pathSpec {
+	p = (p.Clean()).(localPath)
+	return newLocalPath(stripPathShortcuts(p.file))
+}
+
+func (p localPath) IsRelative(dest string) bool {
+	return false
```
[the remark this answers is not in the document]
tallclair: Looks like this is unused? Maybe it would be better to just panic in that case? `panic("unsupported operation: localPath.IsRelative")`

staging/src/k8s.io/kubectl/pkg/cmd/cp/filespec.go
```diff
@@ -0,0 +1,192 @@
...
+func (p remotePath) Prefix() pathSpec {
+	return newLocalPath(stripLeadingSlash(p.file))
+}
+
+func (p remotePath) StripShortcuts() pathSpec {
+	p = (p.Clean()).(remotePath)
+	return newLocalPath(stripPathShortcuts(p.file))
+}
+
+func (p remotePath) IsRelative(dest string) bool {
```
tallclair, line 148: nit: I think this is only used on `pathSpec`s outside of testing?
```suggestion
func (p remotePath) IsRelative(dest pathSpec) bool {
```

tallclair (commented):
This is awesome. So much easier to follow! Thanks for making these changes.

staging/src/k8s.io/kubectl/pkg/cmd/cp/filespec.go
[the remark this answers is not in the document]
soltysh: Yeah, I considered that during implementation and eventually didn't fix it, changing.

staging/src/k8s.io/kubectl/pkg/cmd/cp/filespec.go
```diff
@@ -0,0 +1,192 @@
...
+func (p remotePath) Clean() pathSpec {
+	return newRemotePath(path.Clean(p.file))
+}
+
+func (p remotePath) JoinElem(elem pathSpec) pathSpec {
+	return newRemotePath(path.Join(p.file, elem.String()))
+}
+
+func (p remotePath) JoinFile(elem string) pathSpec {
+	return newRemotePath(filepath.Join(p.file, elem))
```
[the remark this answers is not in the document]
soltysh: Good catch!

staging/src/k8s.io/kubectl/pkg/cmd/cp/filespec.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func (p remotePath) Glob() (matches []string, err error) {
+	return nil, nil
```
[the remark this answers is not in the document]
soltysh: Updated.

staging/src/k8s.io/kubectl/pkg/cmd/cp/filespec.go
```diff
@@ -0,0 +1,192 @@
...
+	return newLocalPath(stripLeadingSlash(p.file))
+}
+
+func (p remotePath) StripShortcuts() pathSpec {
+	p = (p.Clean()).(remotePath)
+	return newLocalPath(stripPathShortcuts(p.file))
+}
+
+func (p remotePath) IsRelative(dest string) bool {
+	relative, err := filepath.Rel(p.file, dest)
```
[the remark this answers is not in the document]
soltysh: Yes, fixed.

staging/src/k8s.io/kubectl/pkg/cmd/cp/filespec.go
soltysh: That was my initial take into it, but that turned out to unnecessarily complicate the interface and implementation in several places. That's why I went with just the `filePath` bits.

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
soltysh: Fixed.

staging/src/k8s.io/kubectl/pkg/cmd/cp/filespec.go
soltysh: Updated.

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
soltysh: I actually squashed `JoinElem` and `JoinFile` into a single `Join` which requires you to pass `filePath` to make sure we clean them properly during processing instead of blindly taking a string value as a parameter.

staging/src/k8s.io/kubectl/pkg/cmd/cp/filespec.go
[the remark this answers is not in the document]
soltysh: Actually this is wrong, `filepath.Rel` should be client-side call, since that's invoked to ensure we can create a path relative file. So I left it as is, but moved to `localPath`, and also fixed all the places where during copying from pod I used `remotePath` instead of `localPath`.

soltysh:
@tallclair all fixed, should be good for final review now

staging/src/k8s.io/kubectl/pkg/cmd/cp/filespec.go
```diff
@@ -0,0 +1,183 @@
...
+func (p localPath) Prefix() pathSpec {
+	return newLocalPath(stripLeadingSlash(p.file))
+}
+
+func (p localPath) StripShortcuts() pathSpec {
+	p = (p.Clean()).(localPath)
+	return newLocalPath(stripPathShortcuts(p.file))
+}
+
+func (p localPath) IsRelative(dest pathSpec) bool {
```
tallclair, line 89: nit: consider asserting that pathSpec is local. It's probably an error to ask if a remote path is relative to a local path.

staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go
```diff
@@ -378,13 +329,13 @@ func recursiveTar(srcBase, srcFile, destBase, destFile string, tw *tar.Writer) e
...
 				hdr, _ := tar.FileInfoHeader(stat, fpath)
-				hdr.Name = destFile
+				hdr.Name = destFile.String()
 				if err := tw.WriteHeader(hdr); err != nil {
 					return err
 				}
 			}
 			for _, f := range files {
-				if err := recursiveTar(srcBase, path.Join(srcFile, f.Name()), destBase, path.Join(destFile, f.Name()), tw); err != nil {
+				if err := recursiveTar(srcBase, srcFile.Join(newLocalPath(f.Name())), destBase, destFile.Join(newRemotePath(f.Name())), tw); err != nil {
```
tallclair, line 338: How do you know that `src` is local and `dest` is remote here?

tallclair: oh, I think because this method is only ever called locally? Remote tar always execs the binary?

In that case, maybe the parameters should use the locol/remote types rather than the generic interface, to make this explicit?

Same goes for `untar` below.

staging/src/k8s.io/kubectl/pkg/cmd/cp/filespec.go
tallclair: on second thought, maybe this (and `Join`) should take the typed (local/remote) path as the argument? That doesn't work with the interface, but it might make sense to just remove these from the interface, and put the onus on the caller to ensure that the right method is being called.

staging/src/k8s.io/kubectl/pkg/cmd/cp/filespec.go
```diff
@@ -0,0 +1,183 @@
...
+	return newLocalPath(stripLeadingSlash(p.file))
+}
+
+func (p remotePath) StripShortcuts() pathSpec {
+	p = (p.Clean()).(remotePath)
+	return newLocalPath(stripPathShortcuts(p.file))
+}
+
+func (p remotePath) IsRelative(dest pathSpec) bool {
+	panic("unsupported operation: remotePath.IsRelative")
```
tallclair, line 144: Thinking about this a bit more, I think it would be better to remove these methods that are unsupported by the `remotePath` from the interface. That way the caller is forced to ensure the `pathSpec` is a `localPath` before calling these.

tallclair (commented):
I'm really sorry I've been so delinquent on this review.

I have a few suggestions around making the path types explicit where we're expecting to be dealing with a specific path type.

tallclair (commented):
/lgtm
/approve
🎉

Issue #7452, opened 2015-04-28
SPDY is deprecated. Switch to HTTP/2.


lavalamp:
Filing for tracking purposes.

See #7392 where the spdy package we use doesn't even exist any more. (And btw it has bugs anyway, possibly security bugs-- see the trophies here: https://github.com/dvyukov/go-fuzz)

Blocked on github.com/docker/spdystream switching to HTTP/2.

(Note: HTTP/2 is based on SPDY, with just a few differences-- hopefully this transition will be easy.)

@thockin @ncdc

ncdc:
Waiting on HTTP/2 support for Go that includes low level client/server ability to create and handle streams (https://github.com/bradfitz/http2).

https://github.com/docker/spdystream/pull/53 pulls the upstream spdy code into spdystream.

lavalamp:
Ah, after docker/spdystream#53 is merged, we can at least make godep happy again. :)

ncdc:
@lavalamp spdystream PR is merged, FYI

ncdc:
@thockin @lavalamp it looks like the Go HTTP/2 issue was just closed (https://github.com/golang/go/issues/6891#issuecomment-161070230). I glanced quickly through the code, but I don't see a way to hijack the connection and create our own streams, like we do with spdystream. Hopefully we can get this functionality in there at some point...

ncdc:
cc @smarterclayton

ncdc:
I created https://github.com/golang/go/issues/13444 as an RFE for end-user stream support

ddysher:
What's the plan for the deprecation?  We are using nginx in front of apiserver which drops support for SPDY since version 1.9.5.  Our workaround is to use an older nginx, but would like to see http/2 support go upstream :)

ncdc:
I prototyped this with a prerelease version of go1.6. It worked somewhat, but there were some rough spots (probably stuff I was doing incorrectly). We can't move to http/2 until go1.6 at the earliest, and even then we'll have to do R&D to see if it's technically feasible. I had been hoping that the http/2 support in go1.6 would allow me to create streams just like we can do in spdystream, but that isn't exposed via the API. For my prototype, I ended up writing a simple multiplexer just like @smarterclayton wrote for the websocket support we have in Kubernetes now.

ddysher:
Thanks for the heads up. Looking at the linked comments, there is no intention to add the support in go1.6, so I guess it's probably still months away.

[...]

PR #94170, opened 2020-08-21
spdy: add optional periodic Pings on the connection

awly:
**What type of PR is this?**

/kind cleanup

When an SPDY connection goes over an intermediate box (proxy or load
balancer, e.g. AWS ELB), it can get interrupted due to idleness (default
in ELB is 60s).  For example, this happens with `kubectl exec` sessions
are left open without any activity for a while.

TCP-level keep-alives are not sufficient for all intermediate boxes,
they may pay attention to application-layer traffic only. SPDY pings
make the connection appear active, letting it survive a period of
idleness.

Note: this commit adds support for pings in
`k8s.io/apimachinery/pkg/util/httpstream/spdy`, but doesn't enable it
anywhere in the calling code. There is no behavior change for existing
callers.

N/A

This will be used in https://github.com/gravitational/teleport/, which acts as a man-in-the-middle auth proxy in for k8s API. When teleport is fronted by AWS ELB (same probably applies to other load balancers), `kubectl exec` sessions get disconnected after 60s of idleness. This PR addresses that.

```release-note
NONE
```

N/A

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/connection.go
+++ b/staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/connection.go
@@ -34,38 +34,62 @@ type connection struct {
 	streams          []httpstream.Stream
 	streamLock       sync.Mutex
 	newStreamHandler httpstream.NewStreamHandler
+	ping             func() (time.Duration, error)
 }
 
 // NewClientConnection creates a new SPDY client connection.
 func NewClientConnection(conn net.Conn) (httpstream.Connection, error) {
+	return NewClientConnectionWithPings(conn, 0)
+}
+
+// NewClientConnectionWithPings creates a new SPDY client connection.
+//
+// If pingPeriod is non-zero, a background goroutine will send periodic Ping
+// frames to the server. Use this to keep idle connections through certain load
+// balancers alive longer.
+func NewClientConnectionWithPings(conn net.Conn, pingPeriod time.Duration) (httpstream.Connection, error) {
 	spdyConn, err := spdystream.NewConnection(conn, false)
 	if err != nil {
 		defer conn.Close()
 		return nil, err
 	}
 
-	return newConnection(spdyConn, httpstream.NoOpNewStreamHandler), nil
+	return newConnection(spdyConn, httpstream.NoOpNewStreamHandler, pingPeriod, spdyConn.Ping), nil
 }
 
 // NewServerConnection creates a new SPDY server connection. newStreamHandler
 // will be invoked when the server receives a newly created stream from the
 // client.
 func NewServerConnection(conn net.Conn, newStreamHandler httpstream.NewStreamHandler) (httpstream.Connection, error) {
+	return NewServerConnectionWithPings(conn, newStreamHandler, 0)
+}
+
+// NewServerConnectionWithPings creates a new SPDY server connection.
+// newStreamHandler will be invoked when the server receives a newly created
+// stream from the client.
+//
+// If pingPeriod is non-zero, a background goroutine will send periodic Ping
+// frames to the server. Use this to keep idle connections through certain load
+// balancers alive longer.
+func NewServerConnectionWithPings(conn net.Conn, newStreamHandler httpstream.NewStreamHandler, pingPeriod time.Duration) (httpstream.Connection, error) {
 	spdyConn, err := spdystream.NewConnection(conn, true)
 	if err != nil {
 		defer conn.Close()
 		return nil, err
 	}
 
-	return newConnection(spdyConn, newStreamHandler), nil
+	return newConnection(spdyConn, newStreamHandler, pingPeriod, spdyConn.Ping), nil
 }
 
 // newConnection returns a new connection wrapping conn. newStreamHandler
 // will be invoked when the server receives a newly created stream from the
 // client.
-func newConnection(conn *spdystream.Connection, newStreamHandler httpstream.NewStreamHandler) httpstream.Connection {
-	c := &connection{conn: conn, newStreamHandler: newStreamHandler}
+func newConnection(conn *spdystream.Connection, newStreamHandler httpstream.NewStreamHandler, pingPeriod time.Duration, pingFn func() (time.Duration, error)) httpstream.Connection {
+	c := &connection{conn: conn, newStreamHandler: newStreamHandler, ping: pingFn}
 	go conn.Serve(c.newSpdyStream)
+	if pingPeriod > 0 && pingFn != nil {
+		go c.sendPings(pingPeriod)
+	}
 	return c
 }
 
@@ -142,4 +166,22 @@ func (c *connection) newSpdyStream(stream *spdystream.Stream) {
 // it is automatically closed.
 func (c *connection) SetIdleTimeout(timeout time.Duration) {
 	c.conn.SetIdleTimeout(timeout)
+}
+
+func (c *connection) sendPings(period time.Duration) {
+	t := time.NewTicker(period)
+	defer t.Stop()
+	for {
+		select {
+		case <-c.conn.CloseChan():
+			return
+		case <-t.C:
+		}
+		if _, err := c.ping(); err != nil {
+			klog.V(3).Infof("SPDY Ping failed: %v", err)
+			// Continue, in case this is a transient failure.
+			// c.conn.CloseChan above will tell us when the connection is
+			// actually closed.
+		}
+	}
 }
--- a/staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/roundtripper.go
+++ b/staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/roundtripper.go
@@ -30,6 +30,7 @@ import (
 	"net/http/httputil"
 	"net/url"
 	"strings"
+	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -70,6 +71,9 @@ type SpdyRoundTripper struct {
 	// requireSameHostRedirects restricts redirect following to only follow redirects to the same host
 	// as the original request.
 	requireSameHostRedirects bool
+	// pingPeriod is a period for sending Ping frames over established
+	// connections.
+	pingPeriod time.Duration
 }
 
 var _ utilnet.TLSClientConfigHolder = &SpdyRoundTripper{}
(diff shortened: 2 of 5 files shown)


fedebongio:
/assign @caesarxuchao 
(@Jefftree was safe because on vacay)

staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/connection.go
```diff
@@ -143,3 +156,24 @@ func (c *connection) newSpdyStream(stream *spdystream.Stream) {
...
+		case <-t.C:
+		}
+		if _, err := c.conn.Ping(); err != nil {
+			klog.V(3).Infof("SPDY Ping failed: %v", err)
+			// Continue, in case this is a transient failure.
+			// c.conn.CloseChan above will tell us when the connection is
+			// actually closed.
+		} else {
+			// Count successful pings, for testing.
+			atomic.AddInt64(&c.pingsSent, 1)
```
caesarxuchao, line 176: Can we avoid this testing-only code? Dose the SPDY library provide any testing support that exposes if a ping frame is received?

caesarxuchao (commented):
The transition from SPDY to HTTP/2 is blocked on https://github.com/kubernetes/kubernetes/issues/7452#issuecomment-491896658, so in the mean time, I think we can accept PRs maintaining SPDY. @deads2k @lavalamp what do you think?

@awly How's the user going to enable the ping?

This is similar to https://github.com/kubernetes/kubernetes/pull/81179. cc @liggitt who reviewed that PR.

staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/connection.go
awly: Unfortunately, the library [handles incoming pings internally](https://github.com/docker/spdystream/blob/master/connection.go#L598-L607) without any external hooks I can find.
This unexported field is the least intrusive way I could verify the behavior. Using `sync/atomic`, performance overhead should be negligible.

awly (commented):
> How's the user going to enable the ping?

@caesarxuchao this PR doesn't expose it to k8s users.
My motivation is for https://github.com/gravitational/teleport/ which uses this library to intercept `kubectl exec` connections for audit purposes (recording session traffic). For `kube-apiserver`, pings are usually unnecessary since it's generally not fronted by a strict load balancer killing idle connections.

If you'd like, I can plumb a flag up to `kube-apiserver`. But it adds unnecessary complexity, if nobody is asking to use it there.

staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/connection.go
```diff
@@ -34,38 +35,50 @@ type connection struct {
...
+	pingsSent        int64
 }
 
 // NewClientConnection creates a new SPDY client connection.
-func NewClientConnection(conn net.Conn) (httpstream.Connection, error) {
+//
+// If pingPeriod is non-zero, a background goroutine will send periodic Ping
+// frames to the server. Use this to keep idle connections through certain load
+// balancers alive longer.
+func NewClientConnection(conn net.Conn, pingPeriod time.Duration) (httpstream.Connection, error) {
```
lavalamp, line 46: Please instead add sibling methods, then you don't have to modify call sites (or break downstream users, if there are any). (E.g., NewClientConnectionWithPings())

staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/connection.go
```diff
@@ -34,38 +35,50 @@ type connection struct {
...
 
 // NewServerConnection creates a new SPDY server connection. newStreamHandler
 // will be invoked when the server receives a newly created stream from the
 // client.
-func NewServerConnection(conn net.Conn, newStreamHandler httpstream.NewStreamHandler) (httpstream.Connection, error) {
+//
+// If pingPeriod is non-zero, a background goroutine will send periodic Ping
+// frames to the server. Use this to keep idle connections through certain load
+// balancers alive longer.
+func NewServerConnection(conn net.Conn, newStreamHandler httpstream.NewStreamHandler, pingPeriod time.Duration) (httpstream.Connection, error) {
```
lavalamp, line 63: Same comment-- don't change public interfaces.

staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/connection.go
lavalamp: +1, if you want to test this, add a member variable of type function for calling Ping and let the test swap that out. It's not performance, it's just a code smell.

staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/upgrade.go
```diff
@@ -63,8 +65,12 @@ func (w *connWrapper) Close() error {
 // NewResponseUpgrader returns a new httpstream.ResponseUpgrader that is
 // capable of upgrading HTTP responses using SPDY/3.1 via the
 // spdystream package.
-func NewResponseUpgrader() httpstream.ResponseUpgrader {
-	return responseUpgrader{}
+//
+// If connPingPeriod is non-zero, for each incoming connection a background
+// goroutine will send periodic Ping frames to the server. Use this to keep
+// idle connections through certain load balancers alive longer.
+func NewResponseUpgrader(connPingPeriod time.Duration) httpstream.ResponseUpgrader {
```
lavalamp, line 72: Here too, don't change the public signature.

lavalamp:
I think I'm OK with adding this and I might even be OK with turning it on, as long as we do that early in the release cycle. But please don't change the function signatures. :)

staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/connection.go
awly: Great idea, done

staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/connection.go
awly: done

staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/upgrade.go
awly: done

staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/connection.go
awly: fair enough, added a swap-able function field in `connection`

awly (commented):
PTAL

lavalamp:
I think I'm OK with this now, @caesarxuchao can you do a final review?

awly:
@caesarxuchao could you take another look please?

staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/connection.go
```diff
@@ -34,38 +34,62 @@ type connection struct {
...
 
 // newConnection returns a new connection wrapping conn. newStreamHandler
 // will be invoked when the server receives a newly created stream from the
 // client.
-func newConnection(conn *spdystream.Connection, newStreamHandler httpstream.NewStreamHandler) httpstream.Connection {
-	c := &connection{conn: conn, newStreamHandler: newStreamHandler}
+func newConnection(conn *spdystream.Connection, newStreamHandler httpstream.NewStreamHandler, pingPeriod time.Duration, pingFn func() (time.Duration, error)) httpstream.Connection {
+	c := &connection{conn: conn, newStreamHandler: newStreamHandler, ping: pingFn}
 	go conn.Serve(c.newSpdyStream)
+	if pingPeriod > 0 {
```
caesarxuchao, line 90: Also check if `c.ping` is nil?

caesarxuchao (commented):
Two nits. Otherwise lgtm.

staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/connection.go
awly: done

staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/roundtripper.go
```diff
@@ -79,20 +83,53 @@ var _ utilnet.Dialer = &SpdyRoundTripper{}
...
 
+// RoundTripperConfig is a set of options for an SpdyRoundTripper.
+type RoundTripperConfig struct {
+	// TLS configuration used by the round tripper.
+	TLS *tls.Config
+	// Proxier is a proxy function invoked on each request. Optional.
+	Proxier func(*http.Request) (*url.URL, error)
+	// ConnPingPeriod is a period for sending SPDY Pings on the connection.
+	// Optional.
+	ConnPingPeriod time.Duration
```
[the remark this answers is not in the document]
awly: renamed, here an in `upgrade.go`

awly (commented):
Thanks @caesarxuchao, nits fixed

staging/src/k8s.io/apimachinery/pkg/util/httpstream/spdy/connection.go
```diff
@@ -34,38 +34,62 @@ type connection struct {
...
-	return newConnection(spdyConn, newStreamHandler), nil
+	return newConnection(spdyConn, newStreamHandler, pingPeriod, spdyConn.Ping), nil
 }
 
 // newConnection returns a new connection wrapping conn. newStreamHandler
 // will be invoked when the server receives a newly created stream from the
 // client.
-func newConnection(conn *spdystream.Connection, newStreamHandler httpstream.NewStreamHandler) httpstream.Connection {
-	c := &connection{conn: conn, newStreamHandler: newStreamHandler}
+func newConnection(conn *spdystream.Connection, newStreamHandler httpstream.NewStreamHandler, pingPeriod time.Duration, pingFn func() (time.Duration, error)) httpstream.Connection {
```
knight42, line 87: nit: given that this is a private function, and it looks like `pingFn` is always `spdyConn.Ping`. Would it be better to omit the `pingFn` argument and invoke `c.conn.Ping` directly in `c.sendPings`?

knight42: Oh I overlook the test..

awly:
Ping @caesarxuchao

awly:
@caesarxuchao this PR is still missing your approval, i think

awly:
ping @caesarxuchao

PR #94171, opened 2020-08-21
Updating EndpointSliceMirroring controller to listen for Service changes

### LINKED ISSUE #93496: [Failing test] TestEndpointSliceMirroring/Service_without_selector

robscott:
**What type of PR is this?**
/kind bug
/kind flake

This fixes a bug that could occur if a custom Endpoints resource was created before a Service was created.

Fixes #93496

A better longterm solution here would be to add a Service event handler here to handle transitions between a Service with/without a selector. That transition is not handled well by any controller right now and is generally just not a well supported use case. This seems to be the smallest possible fix for the bug identified by this flaky test.

```release-note
EndpointSliceMirroring controller now mirrors Endpoints that do not have a Service associated with them.
```

/cc @freehan
/assign @liggitt 
/sig network
/priority critical-urgent

### DIFF
--- a/pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go
+++ b/pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go
@@ -25,7 +25,6 @@ import (
 	v1 "k8s.io/api/core/v1"
 	discovery "k8s.io/api/discovery/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -121,6 +120,11 @@ func NewController(endpointsInformer coreinformers.EndpointsInformer,
 
 	c.serviceLister = serviceInformer.Lister()
 	c.servicesSynced = serviceInformer.Informer().HasSynced
+	serviceInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    c.onServiceAdd,
+		UpdateFunc: c.onServiceUpdate,
+		DeleteFunc: c.onServiceDelete,
+	})
 
 	c.maxEndpointsPerSubset = maxEndpointsPerSubset
 
@@ -273,28 +277,47 @@ func (c *Controller) syncEndpoints(key string) error {
 		return err
 	}
 
-	endpointSlices, err := endpointSlicesMirroredForService(c.endpointSliceLister, namespace, name)
-
+	endpoints, err := c.endpointsLister.Endpoints(namespace).Get(name)
 	if err != nil {
-		ep := &v1.Endpoints{ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace}}
-		c.eventRecorder.Eventf(ep, FailedToListEndpointSlices,
-			"Error listing EndpointSlices for Endpoints %s/%s: %v", ep.Namespace, ep.Name, err)
+		if apierrors.IsNotFound(err) {
+			klog.V(4).Infof("%s/%s Endpoints not found, cleaning up any mirrored EndpointSlices", namespace, name)
+			c.endpointSliceTracker.DeleteService(namespace, name)
+			return c.deleteMirroredSlices(namespace, name)
+		}
 		return err
 	}
 
-	endpoints, err := c.endpointsLister.Endpoints(namespace).Get(name)
-	if err != nil || !c.shouldMirror(endpoints) {
-		if apierrors.IsNotFound(err) || !c.shouldMirror(endpoints) {
+	if !c.shouldMirror(endpoints) {
+		klog.V(4).Infof("%s/%s Endpoints should not be mirrored, cleaning up any mirrored EndpointSlices", namespace, name)
+		c.endpointSliceTracker.DeleteService(namespace, name)
+		return c.deleteMirroredSlices(namespace, name)
+	}
+
+	svc, err := c.serviceLister.Services(namespace).Get(name)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(4).Infof("%s/%s Service not found, cleaning up any mirrored EndpointSlices", namespace, name)
 			c.endpointSliceTracker.DeleteService(namespace, name)
-			return c.reconciler.deleteEndpoints(namespace, name, endpointSlices)
+			return c.deleteMirroredSlices(namespace, name)
 		}
+		return err
+	}
+
+	// This means that if a Service transitions away from a nil selector, any
+	// mirrored EndpointSlices will not be cleaned up. #91072 tracks this issue
+	// for this controller along with the Endpoints and EndpointSlice
+	// controllers.
+	if svc.Spec.Selector != nil {
+		return nil
+	}
+
+	endpointSlices, err := endpointSlicesMirroredForService(c.endpointSliceLister, namespace, name)
+	if err != nil {
 		return err
 	}
 
 	err = c.reconciler.reconcile(endpoints, endpointSlices)
 	if err != nil {
-		c.eventRecorder.Eventf(endpoints, v1.EventTypeWarning, FailedToUpdateEndpointSlices,
-			"Error updating EndpointSlices for Endpoints %s/%s: %v", endpoints.Namespace, endpoints.Name, err)
 		return err
 	}
 
@@ -314,28 +337,54 @@ func (c *Controller) queueEndpoints(obj interface{}) {
 
 // shouldMirror returns true if an Endpoints resource should be mirrored by this
 // controller. This will be false if:
+// - the Endpoints resource is nil.
 // - the Endpoints resource has a skip-mirror label.
 // - the Endpoints resource has a leader election annotation.
-// - the corresponding Service resource does not exist.
-// - the corresponding Service resource has a non-nil selector.
+// This does not ensure that a corresponding Service exists with a nil selector.
+// That check should be performed separately.
 func (c *Controller) shouldMirror(endpoints *v1.Endpoints) bool {
 	if endpoints == nil || skipMirror(endpoints.Labels) || hasLeaderElection(endpoints.Annotations) {
 		return false
 	}
 
-	svc, err := c.serviceLister.Services(endpoints.Namespace).Get(endpoints.Name)
-	if err != nil {
-		if !apierrors.IsNotFound(err) {
-			klog.Errorf("Error fetching %s/%s Service: %v", endpoints.Namespace, endpoints.Name, err)
-		}
-		return false
-	}
-
-	if svc.Spec.Selector != nil {
-		return false
-	}
-
 	return true
+}
+
+// onServiceAdd queues a sync for the relevant Endpoints resource.
+func (c *Controller) onServiceAdd(obj interface{}) {
+	service := obj.(*v1.Service)
+	if service == nil {
+		utilruntime.HandleError(fmt.Errorf("onServiceAdd() expected type v1.Service, got %T", obj))
+		return
+	}
+	if service.Spec.Selector == nil {
+		c.queueEndpoints(obj)
+	}
+}
+
+// onServiceUpdate queues a sync for the relevant Endpoints resource.
+func (c *Controller) onServiceUpdate(prevObj, obj interface{}) {
+	service := obj.(*v1.Service)
+	prevService := prevObj.(*v1.Service)
+	if service == nil || prevService == nil {
+		utilruntime.HandleError(fmt.Errorf("onServiceUpdate() expected type v1.Endpoints, got %T, %T", prevObj, obj))
+		return
+	}
+	if (service.Spec.Selector == nil) != (prevService.Spec.Selector == nil) {
+		c.queueEndpoints(obj)
+	}
+}
+
+// onServiceDelete queues a sync for the relevant Endpoints resource.
+func (c *Controller) onServiceDelete(obj interface{}) {
+	service := getServiceFromDeleteAction(obj)
+	if service == nil {
+		utilruntime.HandleError(fmt.Errorf("onServiceDelete() expected type v1.Service, got %T", obj))
+		return
+	}
+	if service.Spec.Selector == nil {
+		c.queueEndpoints(obj)
+	}
 }
 
 // onEndpointsAdd queues a sync for the relevant Endpoints resource.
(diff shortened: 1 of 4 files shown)


pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go
```diff
@@ -327,11 +327,17 @@ func (c *Controller) shouldMirror(endpoints *v1.Endpoints) bool {
 	if err != nil {
 		if !apierrors.IsNotFound(err) {
 			klog.Errorf("Error fetching %s/%s Service: %v", endpoints.Namespace, endpoints.Name, err)
+			svc, err = c.client.CoreV1().Services(endpoints.Namespace).Get(context.TODO(), endpoints.Name, metav1.GetOptions{})
```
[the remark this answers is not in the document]
liggitt: the current code means in a race condition where the lister is stale but a service with a selector really does exist, we assume the service does not exist and mirror the endpoints, which doesn't seem right

robscott:
@liggitt I've updated the code to do the following on each Endpoints change:

1. If there is an error fetching an Endpoints resource from cache, try a client lookup
2. If the Service can be found and it has a selector, don't mirror it.
3. If not mirroring, do a cache lookup for EndpointSlices mirrored for the Endpoints resource.
4. If an EndpointSlices have been mirrored for the Endpoints resource, delete them.

This ensures that we will not avoid mirroring an Endpoints resource if there is a lookup error. Duplicate EndpointSlices are not an issue, only stale ones. To ensure we don't end up with stale resources, for any Endpoints resources we shouldn't be mirroring, we check the cache for mirrored slices, and delete any that exist.

liggitt:
discussed offline:

1. we really need to requeue syncs for endpoints we can't determine a service for, rather than defaulting either to mirror (which can leave stale endpointslices if a service flips from no selector to selector) or not mirror (which can leave unmirrored endpoints if a service shows up later)
2. doing comparison work in the event handlers instead of the sync handler is more expensive that queuing every endpoint/service add/update/delete and doing the check in the sync handler (because no deduplicating occurs and we still check on every endpoints update)

Proposed plan (@robscott, correct me if I misstate something):
1. add service handlers:
    * onAdd: queue endpoints
    * onUpdate: queue endpoints if selector changed from nil->non-nil or vice versa
    * onDelete: queue endpoints
2. in endpoints handlers, just queue the endpoints (don't do shouldMirror checks)
3. change `shouldMirror() bool` to `shouldMirror() (bool, error)`, return `false, nil` for NotFound errors and `false, err` for other errors
4. in the sync handler, react to an error from `shouldMirror` by requeueing the endpoint with a backoff

robscott:
@liggitt Thanks for the great summary. I ended up making a few tweaks to the approach we discussed above:

1. `shouldMirror()` was still helpful for Endpoints event handlers, so I kept it in a limited capacity. There are some Endpoints that we know we'll never want to mirror just by looking at labels and annotations.
2. I queued Services whenever a selector was non-nil. I did not check the transition between nil and non-nil selectors because it would only be relevant if `syncEndpoints()` deleted EndpointSlices when a Service had a non-nil selector - it doesn't. That would be rather expensive (EndpointSlice selector lookup) and would happen on most Endpoints updates since we're queueing almost everything now. Leaving these behind matches the behavior of the EndpointSlice and Endpoints controllers when a selector transitions between nil and non-nil states.

And finally, as the failing tests indicate, I still need to update some tests. I just wanted to get this out ASAP for an initial review.

pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go
```diff
@@ -273,6 +278,36 @@ func (c *Controller) syncEndpoints(key string) error {
...
+		}
+		klog.Errorf("Error fetching %s/%s Service: %v", namespace, name, err)
+		return err
+	}
+
+	// This means that if a Service transitions away from a nil selector, any
+	// mirrored EndpointSlices will not be cleaned up. This matches the behavior
+	// of the Endpoints and EndpointSlice controllers.
+	if svc.Spec.Selector != nil {
+		klog.V(4).Infof("Not mirroring EndpointSlice for %s/%s because selector is not nil", namespace, name)
```
liggitt, line 307: won't this get logged a *ton* (basically on every endpoints update for almost all services)? I'm not sure I'd log here at all

pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go
```diff
@@ -273,6 +278,36 @@ func (c *Controller) syncEndpoints(key string) error {
...
+		return nil
```
liggitt, line 308: > This matches the behavior of the Endpoints and EndpointSlice controllers.

do you mean it is the inverse of the behavior of those controllers? (e.g. those controllers wouldn't clean up an endpoints/endpointsslice object if a service switched from a selector to a nil selector)?

pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go
```diff
@@ -314,28 +340,54 @@ func (c *Controller) queueEndpoints(obj interface{}) {
...
-		return false
+// onServiceUpdate queues a sync for the relevant Endpoints resource.
+func (c *Controller) onServiceUpdate(prevObj, obj interface{}) {
+	service := obj.(*v1.Service)
+	prevService := prevObj.(*v1.Service)
+	if service == nil || prevService == nil {
+		utilruntime.HandleError(fmt.Errorf("onServiceUpdate() expected type v1.Endpoints, got %T, %T", prevObj, obj))
+		return
+	}
+	if service.Spec.Selector == nil {
```
liggitt, line 376: I expected:
```
if (service.Spec.Selector == nil) != (prevService.Spec.Selector == nil) {
```

do we care about a service update in any other case?

pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go
liggitt: this also seems like a comment that should have an associated issue link to clean up this behavior for all three controllers

liggitt:
looks like the should_mirror unit test needs updating since it no longer checks service:

```
    endpointslicemirroring_controller_test.go:336: Expected false to be returned, got true
--- FAIL: TestShouldMirror (0.01s)
    --- PASS: TestShouldMirror/Service_without_selector_with_matching_endpoints (0.00s)
    --- PASS: TestShouldMirror/Service_without_selector,_matching_Endpoints_with_skip-mirror=true (0.00s)
    --- PASS: TestShouldMirror/Service_without_selector,_matching_Endpoints_with_skip-mirror=invalid (0.00s)
    --- PASS: TestShouldMirror/Service_without_selector,_matching_Endpoints_with_leader_election_annotation (0.00s)
    --- PASS: TestShouldMirror/Service_without_selector,_matching_Endpoints_without_skip_label_in_different_namespace (0.00s)
    --- PASS: TestShouldMirror/Service_without_selector_or_matching_endpoints (0.00s)
    --- FAIL: TestShouldMirror/Endpoints_without_matching_Service (0.00s)
``

File: pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go (hunk 3 above)
Comment (robscott) [reply pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go:308]: Yeah that was very poorly worded. I meant to say that they all do not clean up resources during a transition to/from a nil selector. Instead I've added a link to https://github.com/kubernetes/kubernetes/issues/91072.

File: pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go (hunk 2 above)
Comment (robscott) [reply pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go]: Good call, removed.

File: pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go (hunk 4 above)
Comment (robscott) [reply pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go]: Updated, thanks!

**Comment by robscott** [comment]
Just pushed another update that should resolve the initial feedback from @liggitt and fix any existing tests. I think I may need to add some additional test coverage here to make up for some of the test cases that no longer made sense for shouldMirror, but at least the existing tests should pass now.

File: pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go [hunk 5]
```diff
@@ -273,6 +278,36 @@ func (c *Controller) syncEndpoints(key string) error {
 		return err
 	}
 
+	endpoints, err := c.endpointsLister.Endpoints(namespace).Get(name)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return c.deleteMirroredSlices(namespace, name)
```
liggitt, line 284: just to double-check, deleteMirroredSlices returns success (nil error) if there are no mirrored endpointslices for the given namespace/name, right?

robscott: Correct. There are 2 kinds of errors that can be returned here:

1. An error resulting from `endpointSliceLister.EndpointSlices(namespace).List(esLabelSelector)`
2. An error resulting from `r.client.DiscoveryV1beta1().EndpointSlices(namespace).Delete(context.TODO(), endpointSlice.Name, metav1.DeleteOptions{})`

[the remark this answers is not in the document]
robscott: Updated, thanks!

pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go
```diff
@@ -273,6 +278,36 @@ func (c *Controller) syncEndpoints(key string) error {
...
+		}
+		return err
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go
```diff
@@ -273,6 +278,36 @@ func (c *Controller) syncEndpoints(key string) error {
...
+	}
+
+	if !c.shouldMirror(endpoints) {
+		return c.deleteMirroredSlices(namespace, name)
+	}
+
+	svc, err := c.serviceLister.Services(namespace).Get(name)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(4).Infof("Skipping mirroring for %s/%s due to missing Service", namespace, name)
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go
```diff
@@ -273,6 +278,36 @@ func (c *Controller) syncEndpoints(key string) error {
...
+	endpoints, err := c.endpointsLister.Endpoints(namespace).Get(name)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return c.deleteMirroredSlices(namespace, name)
+		}
+		return err
+	}
+
+	if !c.shouldMirror(endpoints) {
+		return c.deleteMirroredSlices(namespace, name)
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

robscott:
Thanks for the reviews @bowei and @liggitt! I think I've resolved everything now. I've also added some new tests covering `syncEndpoints()` functionality that has moved from `shouldMirror()`.

pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go
```diff
@@ -273,21 +278,48 @@ func (c *Controller) syncEndpoints(key string) error {
...
 	if err != nil {
-		ep := &v1.Endpoints{ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace}}
-		c.eventRecorder.Eventf(ep, FailedToListEndpointSlices,
-			"Error listing EndpointSlices for Endpoints %s/%s: %v", ep.Namespace, ep.Name, err)
+		if apierrors.IsNotFound(err) {
+			klog.V(4).Infof("%s/%s Endpoints not found, cleaning up any mirrored EndpointSlices", namespace, name)
+			c.endpointSliceTracker.DeleteService(namespace, name)
+			return c.deleteMirroredSlices(namespace, name)
+		}
+		klog.Errorf("Error fetching %s/%s Endpoints: %v", namespace, name, err)
```
liggitt, line 288: anywhere we return an error from syncEndpoints, we're already logging in handleErr with a warning... do we need to log again here?

pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go
```diff
@@ -273,21 +278,48 @@ func (c *Controller) syncEndpoints(key string) error {
...
+
+	svc, err := c.serviceLister.Services(namespace).Get(name)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(4).Infof("%s/%s Service not found, cleaning up any mirrored EndpointSlices", namespace, name)
 			c.endpointSliceTracker.DeleteService(namespace, name)
-			return c.reconciler.deleteEndpoints(namespace, name, endpointSlices)
+			return c.deleteMirroredSlices(namespace, name)
 		}
+		klog.Errorf("Error fetching %s/%s Service: %v", namespace, name, err)
```
liggitt, line 305: same comment about double logging

pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go
```diff
@@ -273,21 +278,48 @@ func (c *Controller) syncEndpoints(key string) error {
...
+	if svc.Spec.Selector != nil {
+		return nil
+	}
+
+	endpointSlices, err := endpointSlicesMirroredForService(c.endpointSliceLister, namespace, name)
+
+	if err != nil {
+		ep := &v1.Endpoints{ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace}}
+		c.eventRecorder.Eventf(ep, FailedToListEndpointSlices,
+			"Error listing EndpointSlices for Endpoints %s/%s: %v", ep.Namespace, ep.Name, err)
```
liggitt, line 322: double logging

pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go
robscott: Updated, thanks!

pkg/controller/endpointslicemirroring/endpointslicemirroring_controller.go
robscott: Updated, thanks!

PR #94172, opened 2020-08-22
fail early if not on linux in local-up-cluster.sh

BenTheElder:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**: Instead of failing during bringup on non-linux platforms, check `uname` and automatically fail fast if not on `Linux`. This prevents issues like https://github.com/kubernetes/kubernetes/issues/94164

Further, automating the response from https://github.com/kubernetes/kubernetes/issues/94164 by including a pointer to [kind](https://github.com/kubernetes/kubernetes/issues/94164) which, while not quite the same, does work on most platforms now to create a local cluster from source.

I can drop the latter if we don't want to do this, but it seems more helpful than just failing, and both are upstream.

Fixes #

```release-note
NONE
```

```docs

```

/sig testing

### DIFF
--- a/hack/local-up-cluster.sh
+++ b/hack/local-up-cluster.sh
@@ -20,6 +20,14 @@ KUBE_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 # this as root to allow kubelet to open docker's socket, and to write the test
 # CA in /var/run/kubernetes.
 # Usage: `hack/local-up-cluster.sh`.
+# NOTE: kubelet only works on Linux! It will more or less turn the host
+# into a node ...
+if [[ "$(uname)" != "Linux" && "${START_MODE:-}" != "nokubelet" ]]; then
+  echo "ERROR: local-up-cluster.sh requires disabling kubelet unless run on Linux!" 1>&2
+  echo "You must set START_MODE=nokubelet to disable kubelet." 1>&2
+  echo "On other platforms you can use http://sigs.k8s.io/kind to create a cluster from source including kubelet." 1>&2
+  exit 1
+fi
 
 DOCKER_OPTS=${DOCKER_OPTS:-""}
 export DOCKER=(docker "${DOCKER_OPTS[@]}")


BenTheElder:
Updated to recommend START_MODE=nokubelet first and only fail if START_MODE is not set to nokubelet

spiffxp (approved):
/approve
/lgtm

spiffxp:
/milestone v1.20
cleanup

BenTheElder:
/skip
it just flaked, not related to this PR, and not blocking.

hack/local-up-cluster.sh
```diff
@@ -20,6 +20,14 @@ KUBE_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 # this as root to allow kubelet to open docker's socket, and to write the test
 # CA in /var/run/kubernetes.
 # Usage: `hack/local-up-cluster.sh`.
+# NOTE: kubelet only works on Linux! It will more or less turn the host
+# into a node ...
+if [[ "$(uname)" != "Linux" && "${START_MODE:-}" != "nokubelet" ]]; then
```
liggitt, line 25: /hold

you can't set this and `nokubeproxy` at the same time, and the kubelet startup currently graciously degrades with a message about kubelet not being supported

liggitt:
I don't object to bailing early, as long as we have a way to disable kubelet *and* kube-proxy

hack/local-up-cluster.sh
BenTheElder: should we gracefully degrade to not running kube-proxy the same as kubelet and print a warning then ..?

BenTheElder: that seems better than forcing the user to set the START_MODE, as long as we think they're going to see the warning and understand that this won't have pods or services ...

liggitt: auto-detecting and skipping kubelet and kube-proxy makes sense to me

BenTheElder:
will rewrite to https://github.com/kubernetes/kubernetes/pull/94172#discussion_r479498994 later
/hold

MonzElmasry:

hack/local-up-cluster.sh
joakimr-axis: No need for double quotes for the strings that are not to expand anything. Better use single quotes (or none).
```shell
if [[ "$(uname)" != 'Linux' && "${START_MODE:-}" != 'nokubelet' ]]; then
```

hack/local-up-cluster.sh
```diff
@@ -20,6 +20,14 @@ KUBE_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
...
+  echo "ERROR: local-up-cluster.sh requires disabling kubelet unless run on Linux!" 1>&2
```
joakimr-axis, line 26: More efficient to use single `echo` call (and also no need for double quotes since nothing is to be expanded).
```shell
  echo 'ERROR: local-up-cluster.sh requires disabling kubelet unless run on Linux!
You must set START_MODE=nokubelet to disable kubelet.
On other platforms you can use http://sigs.k8s.io/kind to create a cluster from source including kubelet.' 1>&2
```

BenTheElder:
alternative idea makes sense to me, but I don't use this and haven't really had time to come back to this.

PR #94174, opened 2020-08-22
set persistenet to false

like-inspur:
this config can reduce disk io usage
/kind bug

### DIFF
--- a/cluster/addons/fluentd-elasticsearch/fluentd-es-configmap.yaml
+++ b/cluster/addons/fluentd-elasticsearch/fluentd-es-configmap.yaml
@@ -343,7 +343,7 @@ data:
       matches [{ "_SYSTEMD_UNIT": "docker.service" }]
       <storage>
         @type local
-        persistent true
+        persistent false
         path /var/log/journald-docker.pos
       </storage>
       read_from_head true
@@ -356,7 +356,7 @@ data:
       matches [{ "_SYSTEMD_UNIT": "{{ fluentd_container_runtime_service }}.service" }]
       <storage>
         @type local
-        persistent true
+        persistent false
         path /var/log/journald-container-runtime.pos
       </storage>
       read_from_head true
@@ -369,7 +369,7 @@ data:
       matches [{ "_SYSTEMD_UNIT": "kubelet.service" }]
       <storage>
         @type local
-        persistent true
+        persistent false
         path /var/log/journald-kubelet.pos
       </storage>
       read_from_head true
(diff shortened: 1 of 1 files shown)


coffeepac:
@like-inspur this will lower disk I/O but only in the 'happy state' where the fluentd container is very stable.  if the fluentd container is not stable then fluentd will be re-pulling lots and lot of logs.

do you have any number on how much this decreases I/O and what the number of pods and log files being read are?

like-inspur:
@coffeepac  no, I just found if config persistent to false, fluentd will not always occupy pos file and change filed , so the cpu and disk usage of fluentd will not be always high

like-inspur:
@coffeepac I add suppress_repeated_stacktrace to prevent form fluentd print lot of logs when it is not stable

coffeepac:
@like-inspur I think the suppression flag is a full positive, but I am hesitant to okay the removal of the pos file.  fluentd is commonly run in a daemonset on long-running nodes.  we should be able to release a new version of the fluentd image without repeating a whole bunch of log messages.  

this feels like something that would better as a config option in a helm chart and not this fairly basic but conservative add-on.

like-inspur:
@coffeepac this config not represent that ths removal of this pos, but represent change pos every 10 seconds not each fluentd read journal file, this code and default config of fluentd is under this: https://github.com/fluent/fluentd/blob/b593e6f4adb08dd9f2d2fd69f83fe7a6a1514d43/lib/fluent/plugin/storage.rb#L31

coffeepac:
@like-inspur I will admit I am not terribly familiar with this part of fluentd so I'm reading code as I go.

In the interest of being both conservative and obvious, can we also set the autosave_interval for all of this plugins to be 10 seconds?  While this will have no immediate practical effect, it will provide a clue to users of this plugin who don't know fluentd very well that there is an important knob here.

also, this requires bumping the fluentd image version values starting here: https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/fluentd-elasticsearch/fluentd-es-ds.yaml#L48 (a few locations) and here:  https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/fluentd-elasticsearch/fluentd-es-image/Makefile#L19

and please note there are already out of sync so please do a patch bump to v3.0.5 in all locations.  

and thank you!

like-inspur:
@coffeepac ok, I wil wait 3.0.5 version issue and recommit a new pull request that contain autosave_interval config, thak you

pacoxu:
This PR still needs LGTM from @sig-instrumentation-reviewers ? 
Should this have a release note?

coffeepac:
@pacoxu we can close this.

@like-inspur this plugin is moving to https://github.com/kubernetes-sigs/instrumentation-addons/pull/1, when that PR is merged, feel free to re-open this if you still want it.

/close

PR #81774, opened 2019-08-22
kubelet: sync node allocatable cgroups upon status update

rojkov:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

Eliminate discrepancy between resources enforced with cgroups and the node's capacity which can lead to scheduling a workload to the node that can't actually provide the required resources.

**Which issue(s) this PR fixes**:
Fixes #71233

**Special notes for your reviewer**:
An alternative solution could be disabling node status updates for a pre-defined set of resources, i.e. HugePages.

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

PR #94176, opened 2020-08-22
Shallow copy machineInfo to avoid race condition

odinuge:
**What type of PR is this?**

/kind bug

This is a backport of this fix in cadvisor, that is present in v1.19: https://github.com/google/cadvisor/pull/2490

Cadvisor is updating his each 5min, and since we are using a separate
lock, this could lead to a race condition.

More info can be found here: https://github.com/kubernetes/kubernetes/pull/81774#issuecomment-674394507

This will result in a functional change since huge page count will no longer be updated without a restart of kubelet, but the current behavior (1.18 and prev) doesn't deal with cgroups properly, so it leads to a lot of strange bugs (see discussion here: #81774). With this backport the behavior in v1.18 will be similar to v1.19.

Fixes #

```release-note
Fix possible race condition with cadvisor machineInfo
```

```docs

```

### DIFF
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -649,7 +649,9 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 	if err != nil {
 		return nil, err
 	}
-	klet.machineInfo = machineInfo
+	// Copy machineInfo to avoid race condition when the value is updated inside cadvisor
+	machineInfoCopy := *machineInfo
+	klet.machineInfo = &machineInfoCopy
 
 	imageBackOff := flowcontrol.NewBackOff(backOffPeriod, MaxContainerBackOff)
 


dashpole:
The "correct" way to do this is to cherry-pick the cAdvisor PR to the corresponding branch (I think the release-v0.36 branch), and then update the cAdvisor dependency on the 1.18 branch.

odinuge:
> The "correct" way to do this is to cherry-pick the cAdvisor PR to the corresponding branch (I think the release-v0.36 branch), and then update the cAdvisor dependency on the 1.18 branch.

Yeah, if the cadvisor folks (and I guess that is you @dashpole) are willing to do that, that would definitely be the best solution. We should do the same for all previous branches, so that might be some extra work tho. We should probably also first decide if we would like to actually to this before starting the work in cadvisor.

dashpole:
> ... all previous branches ...

Are there more cherrypicks i'm not aware of?

I'm happy to review cherrypicks to cAdvisor, and I can cut a patch release after the cherrypick is complete.  After that, we just need to bump the cAdvisor dependency in this branch of k/k

SergeyKanzhelev:
@odinuge thank you for this PR! Let's go cAdvisor path for this bug fix.

/close

odinuge:
> Are there more cherrypicks i'm not aware of?

Ahh, turns out all the supported versions of k8s that require a backport use the same version of cadvisor, so then it would be quite quick to get done!

> I'm happy to review cherrypicks to cAdvisor, and I can cut a patch release after the cherrypick is complete. After that, we just need to bump the cAdvisor dependency in this branch of k/k

Nice. Made a PR here: google/cadvisor#2672

PR #94178, opened 2020-08-22
Adjust pod start up based on OS

jingxu97:
For windows, pod start up time is longer than Linux due to formatting and
image downloading

**What type of PR is this?**

Fixes #

```release-note
NONE
```

```docs

```


andrewsykim:
> For windows, pod start up time is longer than Linux due to formatting and
image downloading

Is this causing some tests to fail or flake?

test/e2e/framework/pod/wait.go
```diff
@@ -53,6 +54,9 @@ const (
 	// podStartTimeout is how long to wait for the pod to be started.
 	podStartTimeout = 5 * time.Minute
 
+	// windowsPodStartTimeout is how long to wait for the pod to be started on Windows.
```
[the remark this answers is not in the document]
jingxu97: @andrewsykim thanks for review. I updated the comments.

jingxu97:
> > For windows, pod start up time is longer than Linux due to formatting and
> > image downloading
> 
> Is this causing some tests to fail or flake?

Yes, I am testing gce pd driver presubmit, which most of the tests are failing now due to timeout issue.

andrewsykim:
There are some other storage tests that are setting podStartTimeout, should those be based on OS as well?

https://github.com/kubernetes/kubernetes/blob/3352c44949684edc2a72fc7891fbaa2273a2aa35/test/e2e/storage/csi_mock_volume.go#L1112-L1115

jingxu97:
> There are some other storage tests that are setting podStartTimeout, should those be based on OS as well?
> 
> https://github.com/kubernetes/kubernetes/blob/3352c44949684edc2a72fc7891fbaa2273a2aa35/test/e2e/storage/csi_mock_volume.go#L1112-L1115

we don't run windows test with csi mock driver. So it is not needed.

andrewsykim (commented):
/approve

Will defer lgtm to someone from SIG windows

jingxu97:
cc @msau42

test/e2e/framework/pod/wait.go
```diff
@@ -226,13 +231,20 @@ func WaitForPodCondition(c clientset.Interface, ns, podName, desc string, timeou
 	return fmt.Errorf("Gave up after waiting %v for pod %q to be %q", timeout, podName, desc)
 }
 
+func podTimeout() time.Duration {
```
marosset, line 234: For most of the e2e tests we run in prow the e2etest binary is compiled and runs in linux containers so I'm worried still return the linux timeout most of the time.

This should be checking if the pod is scheduled to a Windows nodes but I am having trouble recommending a generic way to do this in the e2e test framwork at the moment :(

marosset: The best way that I can think of to accomplish this is to add a command-line argument to the e2etest binary to override the default pod timeout and then set that in the prow configs for the Windows test passes.

@adelina-t @claudiubelu can either of you think of a better way to set a global pod start-up timeout based on OS?

hasheddan:
@jingxu97 does this need to be in v1.19 milestone? If so, who do we need for lgtm?

test/e2e/framework/pod/wait.go
jingxu97: My understanding is runtime.GOOS will check when the code is executed? I did similar change before and it is working?

jingxu97: @marosset thanks for pointing it out. 
I changed to use framework.NodeOSDistroIs("windows"), I think this should work. When running the test nodeOSDistrols will set to "windows" for windows tests.

marosset: Didn't know that existed but this should work nicely.

test/e2e/framework/pod/wait.go
```diff
@@ -226,13 +231,20 @@ func WaitForPodCondition(c clientset.Interface, ns, podName, desc string, timeou
...
+	if framework.NodeOSDistroIs("windows") {
```
msau42, line 235: Does this mean we don't run e2e tests on mixed linux + windows node clusters?

marosset: In general we do not.

michmike (approved):
/lgtm
/approve

test/e2e/framework/pod/wait.go
```diff
@@ -34,6 +34,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubectl/pkg/util/podutils"
+	"k8s.io/kubernetes/test/e2e/framework"
```
hasheddan, line 37: @jingxu97 looks like you will need a `./hack/update-bazel.sh` to include new dependency

justaugustus:
Moving this to the 1.20 milestone.
/milestone v1.20

liggitt:
> Yes, I am testing gce pd driver presubmit, which most of the tests are failing now due to timeout issue.

Is there a specific image that test is using that we could add a longer timeout for? Five minutes is already a while, and doubling that for all pods on windows seems likely to mask startup times slowing down unexpectedly

jingxu97:
> > Yes, I am testing gce pd driver presubmit, which most of the tests are failing now due to timeout issue.
> 
> Is there a specific image that test is using that we could add a longer timeout for? Five minutes is already a while, and doubling that for all pods on windows seems likely to mask startup times slowing down unexpectedly

Many of the tests are using aghost. I noticed that in most recent testing, disk formatting already took 2-3 mins, and more than 2 mins image downloading.

test/e2e/framework/pod/wait.go
jingxu97: it is a mixed cluster, but in test config for Windows testing, we pass Windows as nodeOSDistro.

andrewsykim:
/approve cancel

(to stop automatic retests

jingxu97:
This PR hit the import cycle dependency issue. #81245.
Currently open another PR to workaround it #95018

PR #94187, opened 2020-08-24
Avoid the API call to update pod if nothing is changed

chendave:

**What type of PR is this?**

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/scheduler/util/utils.go
+++ b/pkg/scheduler/util/utils.go
@@ -133,6 +133,11 @@ func PatchPod(cs kubernetes.Interface, old *v1.Pod, new *v1.Pod) error {
 	if err != nil {
 		return fmt.Errorf("failed to create merge patch for pod %q/%q: %v", old.Namespace, old.Name, err)
 	}
+
+	if "{}" == string(patchBytes) {
+		return nil
+	}
+
 	_, err = cs.CoreV1().Pods(old.Namespace).Patch(context.TODO(), old.Name, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{}, "status")
 	return err
 }


pkg/scheduler/util/utils.go
```diff
@@ -120,6 +121,9 @@ func GetPodAntiAffinityTerms(affinity *v1.Affinity) (terms []v1.PodAffinityTerm)
 // PatchPod calculates the delta bytes change from <old> to <new>,
 // and then submit a request to API server to patch the pod changes.
 func PatchPod(cs kubernetes.Interface, old *v1.Pod, new *v1.Pod) error {
+	if reflect.DeepEqual(old, new) {
```
hex108, line 124: Is there any place calling this function when old and new are equal?

BTW: This function seems only change pod status, so `reflect.DeepEqual(old.Status, new.Status)` might be enough.

ahg-g: A couple of things:
1) DeepEqual is expensive
2) We don't call this function unless there is something to update (either updating the pod condition or clearing a set NominatedNodeName).

ahg-g (commented):
/hold

This is not necessary

pkg/scheduler/util/utils.go
ahg-g: one thing we can do is check `if len(patchBytes) == 0 {return nil}` below

chendave: I'd hope quality of the public method like this doesn't depends on how it's got called even though it might not happen today. 
Have done a check that `len(patchBytes)` won't be zero even old and new are identical, maybe something is padding there, I am not sure.

alculquicondor: Could you check what's the output of `CreateTwoWayMergePatch` for the same object? It might be something like `{}`. There's no point on adding an extra traversal over the Pod objects.

chendave: Here it is when they are the same object, 
`(dlv) p patchBytes
[]uint8 len: 2, cap: 8, [123,125]
`
not sure what's those bytes stand for, will check it later.

chendave: seems like the ASCII is "S",  "U".

alculquicondor: It is indeed `{}`

chendave: hmm... sorry, I cannot get your point here, why it's `{}` when it's hold two bytes there? for this case, `len(patchBytes) == 2` instead of `0`, am I wrong?

chendave: Get what your meant,  PR is updated.

alculquicondor: why unmarshal instead of doing a simple string comparison?

pkg/scheduler/util/utils.go
```diff
@@ -133,6 +133,16 @@ func PatchPod(cs kubernetes.Interface, old *v1.Pod, new *v1.Pod) error {
...
+
+	modifiedMap := map[string]interface{}{}
+	err = json.Unmarshal(patchBytes, &modifiedMap)
+	if err != nil {
+		fmt.Println("unable to unmarshal patchBytes:", err)
+	}
+	if n := len(modifiedMap); n == 0 {
+		return nil
+	}
+
```
Huang-Wei, line 145: This logic feels like overkill to me. I'd expect we add this (or similar) logic only when necessary, or at least this behavior can be verified by a test case.

chendave: yes, follow @alculquicondor 's suggestion, the change should be light now.

alculquicondor:
/lgtm

leaving approval to @Huang-Wei.

chendave:
kindly ping @ahg-g , are you okay to lift the hold tag? 

@Huang-Wei  I saw there are multiple PRs are ready to be merged but without the milestone, do you know who are eligible to set the milestone?

alculquicondor:
we can wait for the milestone requirement to be lifted, as this is not critical

PR #94190, opened 2020-08-24
Maintain sampleAndWaterMarkHistograms in server handlers

bjrara:
**What type of PR is this?**

/kind bug

According to the problem statement in https://github.com/kubernetes/kubernetes/issues/94145, the same would happen to maxinflight handler and apf handler. 

This PR adds goroutines in both handlers to avoid iterations in `sampleAndWaterMarkHistograms.innerSet` from blocking requests.

https://github.com/kubernetes/kubernetes/pull/94146#issuecomment-678949641

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/server/filters/maxinflight.go
+++ b/staging/src/k8s.io/apiserver/pkg/server/filters/maxinflight.go
@@ -41,6 +41,11 @@ const (
 	// the metrics tracks maximal value over period making this
 	// longer will increase the metric value.
 	inflightUsageMetricUpdatePeriod = time.Second
+
+	// How often the observers should be updated. The periodical
+	// updates can ensure sampleAndWaterMarkHistograms do
+	// not fall far behind.
+	observationUpdatePeriod = 10 * time.Second
 )
 
 var nonMutatingRequestVerbs = sets.NewString("get", "list", "watch")
@@ -103,6 +108,13 @@ func startRecordingUsage(watermark *requestWatermark) {
 	}()
 }
 
+func startMaintainingObservation(watermark *requestWatermark) {
+	go wait.Forever(func() {
+		watermark.readOnlyObserver.Add(0)
+		watermark.mutatingObserver.Add(0)
+	}, observationUpdatePeriod)
+}
+
 var startOnce sync.Once
 
 // WithMaxInFlightLimit limits the number of in-flight requests to buffer size of the passed in channel.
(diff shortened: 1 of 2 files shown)


MikeSpreitzer (commented):
I think this is a non-issue.  There is never going to be a long period of time with zero mutating or zero read-only requests.

MikeSpreitzer:
On second thought, this might be needed in some cases.  Let's not bet on a fragile property.
/reopen

MikeSpreitzer (approved):
/LGTM
Thanks!

staging/src/k8s.io/apiserver/pkg/server/filters/maxinflight.go
```diff
@@ -103,6 +105,15 @@ func startRecordingUsage(watermark *requestWatermark) {
 	}()
 }
 
+func startMaintainingObservation(watermark *requestWatermark) {
+	go func() {
+		wait.Forever(func() {
```
smarterclayton, line 110: style: Generally this is `go wait.Forever(...)`

staging/src/k8s.io/apiserver/pkg/server/filters/maxinflight.go
```diff
@@ -41,6 +41,8 @@ const (
 	// the metrics tracks maximal value over period making this
 	// longer will increase the metric value.
 	inflightUsageMetricUpdatePeriod = time.Second
+
+	observationUpdatePeriod = 10 * time.Second
```
smarterclayton, line 45: Comment justifying this period please

MikeSpreitzer:
FYI, https://github.com/prometheus/client_golang/issues/796#issuecomment-689888087 .  I am pursuing an approach that will remove the need for the periodic background process.  See #94672 .

bjrara:
> FYI, and in response to [#95303 (comment)](https://github.com/kubernetes/kubernetes/pull/95303#issuecomment-704644448) : [prometheus/client_golang#796 (comment)](https://github.com/prometheus/client_golang/issues/796#issuecomment-689888087) . I am pursuing an approach that will remove the need for the periodic background process. See #94672 .

Thanks for letting me know. #94672 is definitely a more elegant approach to handle the fall behind issue. I guess I could hold this PR and close it after #94672 is done.

staging/src/k8s.io/apiserver/pkg/server/filters/maxinflight.go
bjrara: done

MikeSpreitzer:
#95371 covers this, with an improvement (stopping the goroutines when the server stops).

PR #94200, opened 2020-08-24
log metrics: drop error if runtime is not setup

haircommander:
**What type of PR is this?**

/kind cleanup

kubelet initializes the LogMetricsCollector in its initializeModules() function. This function does not require the runtime to be up.
However, if the runtime is not up, running cadvisor.ListPodStats() before the runtime is up can cause an error: `failed to get pod stats: failed to get imageFs info: non-existent label "crio-images"`

Instead of outputting this error (which unnerves users), we should just silently ignore the error. This is in line with what the VolumeStatsCollector does

Signed-off-by: Peter Hunt <pehunt@redhat.com>

Fixes #

```release-note
none
```
/area kubelet

```docs
none
```

### DIFF
--- a/pkg/kubelet/metrics/collectors/log_metrics.go
+++ b/pkg/kubelet/metrics/collectors/log_metrics.go
@@ -18,7 +18,6 @@ package collectors
 
 import (
 	"k8s.io/component-base/metrics"
-	"k8s.io/klog/v2"
 	statsapi "k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
 )
 
@@ -63,7 +62,6 @@ func (c *logMetricsCollector) DescribeWithStability(ch chan<- *metrics.Desc) {
 func (c *logMetricsCollector) CollectWithStability(ch chan<- metrics.Metric) {
 	podStats, err := c.podStats()
 	if err != nil {
-		klog.Errorf("failed to get pod stats: %v", err)
 		return
 	}
 


pkg/kubelet/metrics/collectors/log_metrics.go
```diff
@@ -63,7 +62,6 @@ func (c *logMetricsCollector) DescribeWithStability(ch chan<- *metrics.Desc) {
 func (c *logMetricsCollector) CollectWithStability(ch chan<- metrics.Metric) {
 	podStats, err := c.podStats()
 	if err != nil {
-		klog.Errorf("failed to get pod stats: %v", err)
```
jodavis-suse, line 66: Ah, interesting.  From the description I thought there would need to be a check to determine if the runtime was up, but from the change I can see it is much simpler that the error message is not needed.

jodavis-suse (commented):
lgtm

dashpole:
I don't think we want to completely drop errors from metric collection... 

Instead we might be able to:

Qualify the error message.  E.g. "... If the kubelet just started, dependencies may not be ready yet."
or
Plumb through the isInitialized() function and check it before querying stats.  Have a warning instead of an error for initialization.

haircommander:
ah, forgot about this. I actually found out the excessive logs were because cri-o was lying to cadvisor about where the image store was. The logging was only excessive because of that lie. I believe we are fine to close this, as it was a symptom of a more problematic cause.

/close

PR #94204, opened 2020-08-24
Impersonated user with a specified group should not fail flow schema match in Priority & Fairness

tkashem:
**What type of PR is this?**
/kind bug

Currently we add the `system:authenticated` group for any authenticated user (except when the user is `system:anonymous` or the user is in `system:unauthenticated` group). 
On the other hand, if a group is specified (via `--as-group`) for an impersonated user, we don't add `system:authenticated`. The impersonated user's information is stored in the request context and is used to serve the request.

This causes priority and fairness matching to fail if there are no `FlowSchema` rules explicitly permitting the specified group. The `catch-all` flow schema needs the user to be in either the 'system:authenticated' or the 'system:unauthenticated' group. An impersonated user with a specified group is in neither.
If there is no match, the request fails with a panic. Here is an example:

Potential solutions:
Solution A: change the impersonation filter to add `system:authenticated` group, with some constraints:
- the user is not `system:anonymous`
- the user is not in `system:unauthenticated` group

I would like to know if this approach have any other ripple effects. But today I can simulate this by doing the following:
`kubectl -n kube-system get services --as=test --as-group=mygroup --as-group=system:authenticated`

Solution B: change the `catch-all` flow schema rules to allow `all` group
```
kind: FlowSchema
spec:
    subjects:
    - group:
        name: '*'
      kind: Group
```

In this work in progress PR I have outlined solution A. I am happy to change it depending on the preferred solution.

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/endpoints/filters/impersonation.go
+++ b/staging/src/k8s.io/apiserver/pkg/endpoints/filters/impersonation.go
@@ -117,10 +117,37 @@ func WithImpersonation(handler http.Handler, a authorizer.Authorizer, s runtime.
 			}
 		}
 
-		if !groupsSpecified && username != user.Anonymous {
-			// When impersonating a non-anonymous user, if no groups were specified
-			// include the system:authenticated group in the impersonated user info
-			groups = append(groups, user.AllAuthenticated)
+		if username != user.Anonymous {
+			// When impersonating a non-anonymous user, include the 'system:authenticated' group
+			// in the impersonated user info:
+			// - if no groups were specified
+			// - if a group has been specified other than 'system:authenticated'
+			//
+			// If 'system:unauthenticated' group has been specified we should not include
+			// the 'system:authenticated' group.
+			addAuthenticated := true
+			for _, group := range groups {
+				if group == user.AllAuthenticated || group == user.AllUnauthenticated {
+					addAuthenticated = false
+					break
+				}
+			}
+
+			if addAuthenticated {
+				groups = append(groups, user.AllAuthenticated)
+			}
+		} else {
+			addUnauthenticated := true
+			for _, group := range groups {
+				if group == user.AllUnauthenticated {
+					addUnauthenticated = false
+					break
+				}
+			}
+
+			if addUnauthenticated {
+				groups = append(groups, user.AllUnauthenticated)
+			}
 		}
 
 		newUser := &user.DefaultInfo{
--- a/test/cmd/authorization.sh
+++ b/test/cmd/authorization.sh
@@ -71,8 +71,8 @@ run_impersonation_tests() {
 
     # --as-group
     kubectl create -f hack/testdata/csr.yml "${kube_flags_with_token[@]:?}" --as=user1 --as-group=group2 --as-group=group1 --as-group=,,,chameleon
-    kube::test::get_object_assert 'csr/foo' '{{len .spec.groups}}' '3'
-    kube::test::get_object_assert 'csr/foo' '{{range .spec.groups}}{{.}} {{end}}' 'group2 group1 ,,,chameleon '
+    kube::test::get_object_assert 'csr/foo' '{{len .spec.groups}}' '4'
+    kube::test::get_object_assert 'csr/foo' '{{range .spec.groups}}{{.}} {{end}}' 'group2 group1 ,,,chameleon system:authenticated '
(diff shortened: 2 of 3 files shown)


MikeSpreitzer (approved):
I was clearly told (by @lavalamp and/or @deads2k ) that every request will be either in group `system:authenticated` or group `system:unauthenticated`.  This is also documented in the last example in https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-binding-examples .  So this leads me to favor solution A.

yue9944882:
/ok-to-test

it sounds more of a bug in the impersonation so am +1 on A. and this pull needs a unit-test.

tkashem:
Thanks for your feedback @MikeSpreitzer  @yue9944882. I will go ahead and take care of the unit tests.

lavalamp:
/sig auth

It sounds like a bug (thanks for catching it!), let's let SIG Auth take a look.

test/cmd/authorization.sh
```diff
@@ -71,8 +71,8 @@ run_impersonation_tests() {
 
     # --as-group
     kubectl create -f hack/testdata/csr.yml "${kube_flags_with_token[@]:?}" --as=user1 --as-group=group2 --as-group=group1 --as-group=,,,chameleon
-    kube::test::get_object_assert 'csr/foo' '{{len .spec.groups}}' '3'
-    kube::test::get_object_assert 'csr/foo' '{{range .spec.groups}}{{.}} {{end}}' 'group2 group1 ,,,chameleon '
+    kube::test::get_object_assert 'csr/foo' '{{len .spec.groups}}' '4'
+    kube::test::get_object_assert 'csr/foo' '{{range .spec.groups}}{{.}} {{end}}' 'group2 group1 ,,,chameleon system:authenticated '
```
liggitt, line 75: cc @deads2k for changes to group impersonation

staging/src/k8s.io/apiserver/pkg/endpoints/filters/impersonation.go
```diff
@@ -117,10 +117,25 @@ func WithImpersonation(handler http.Handler, a authorizer.Authorizer, s runtime.
 			}
 		}
 
-		if !groupsSpecified && username != user.Anonymous {
-			// When impersonating a non-anonymous user, if no groups were specified
-			// include the system:authenticated group in the impersonated user info
-			groups = append(groups, user.AllAuthenticated)
+		if username != user.Anonymous {
```
deads2k, line 120: I'd like to an `else` on this that ensure that system:anonymous always has `system:unauthenticated` as a group too.

tkashem: @deads2k does not look like we are adding `system:unauthenticated` group for a `system:anonymous` user in the authenticated group adder -  https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/authentication/group/authenticated_group_adder.go#L46. 

I am not sure if we are doing it elsewhere. Do you want me to fix this?

tkashem: I am wrong, it gets added here - https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/authentication/request/anonymous/anonymous.go#L38

tkashem: > I'd like to an `else` on this that ensure that system:anonymous always has `system:unauthenticated` as a group too.

@deads2k  done, please check.

deads2k:
/lgtm

@tkashem This should get cherrypicked back to the supported releases.

MikeSpreitzer:
@tkashem : is this suffering from a known flake?  Is there a flake to report here?

tkashem:
/retest

it seems to be suffering from a flake. I see `pull-kubernetes-e2e-kind` failing on a different PR as well. The following tests are failing:
[k8s.io] [sig-node] Events should be sent by kubelets and the scheduler about pods scheduling and running [Conformance] 
[sig-node] RuntimeClass should reject a Pod requesting a RuntimeClass with an unconfigured handler

liggitt:
flake was due to https://github.com/kubernetes/test-infra/issues/19080#issuecomment-685141853

/retest

PR #94205, opened 2020-08-24
add configuration for controller migration.

jiahuif:
**What type of PR is this?**

/kind feature

This PR is the first step towards HA controller migration, as purposed in [this KEP ](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cloud-provider/20190422-cloud-controller-manager-migration.md). The PR adds commonly used configuration types to `staging/controller-manager`.

```release-note
NONE
```

```docs

```


staging/src/k8s.io/controller-manager/config/types.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// LeaderMigrationConfiguration provides versioned configuration for all migrating leader locks.
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type LeaderMigrationConfiguration struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// LeaderName is the name of the resource under which the controllers should be run.
```
cheftako, line 28: I think we need more detail to this explanation. I believe this will be the name of the leader election resource/instance which protects the migration. Eg. "1-20-KCM-to-1-21-CCM"

cheftako:
Are we planning a follow up PR which will then load this configuration from idisk and use it?

jiahuif:
Let's make sure this configuration look good, then I will follow up with configuration parsing, etc.

staging/src/k8s.io/controller-manager/config/types.go
jiahuif: Updated a bit...hopefully it is clearer.

staging/src/k8s.io/controller-manager/config/v1alpha1/types.go
```diff
@@ -0,0 +1,45 @@
...
+	LeaderName string `json:"leaderName"`
+
+	// ControllerLeaders contains a list of migrating leader lock configurations
+	ControllerLeaders []ControllerLeaderConfiguration `json:"controllerLeaders"`
+}
+
+// ControllerLeaderConfiguration provides the configuration for a migrating leader lock.
+type ControllerLeaderConfiguration struct {
+	// Name is the name of the controller being migrated
+	// E.g. service-controller, route-controller, cloud-node-controller, etc
```
lavalamp, line 39: I'd kinda like to see the validation code, I assume this will be checked on startup and controller manager will fail if it refers to an unknown controller?

staging/src/k8s.io/controller-manager/config/v1alpha1/types.go
```diff
@@ -0,0 +1,45 @@
...
+
+package config
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// LeaderMigrationConfiguration provides versioned configuration for all migrating leader locks.
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type LeaderMigrationConfiguration struct {
```
lavalamp, line 25: I'm not sure I understand the rationale for plumbing this as a config file, which I guess is the plan? It seems like it will be easy to accidentally get it different between controller managers.

Did you think about maybe using a configmap to make sure that all controller binaries can agree on this?

Sorry for not asking this in the KEP :)

@cheftako @liggitt

liggitt: Consuming as a file lets you feed in config from practically any source, including a configmap.

Tying directly to an API-observed configmap either assumes that you'll consume the config from the same cluster you're managing (which isn't always desired) or requires adding a second API client config path.

cc @deads2k for opinions on config via file vs API-fetched configmap.

staging/src/k8s.io/controller-manager/config/v1alpha1/types.go
cheftako: +1. However that seems like it comes in more with the config load than the config itself. I think the controller manager should only fail if its the referenced controller manager and its an unknown controller.

staging/src/k8s.io/controller-manager/config/v1alpha1/types.go
cheftako: So the plan is a config file. I prefer the config file as its local to the control plane and safe from something like being accidentally deleted be a customer etc. Also we can ensure that the config files exist as part of all relevant version deployments as they are in the image, in either direction (upgrade or downgrade). Ensuring that the resource exists seems more error prone, especially in the downgrade case. Also the clean up of the file seems more natural. The 1.21 to 1.22 migration file exists in all 1.21 and 1.22 control plane releases. For the resource I have to worry about when its safe to delete (and re-add) the resource.

cheftako: Also as a resource this becomes and extension point for customers to disable built in controllers. Not sure that is something I want to "accidentally" implement.

staging/src/k8s.io/controller-manager/config/types.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// LeaderMigrationConfiguration provides versioned configuration for all migrating leader locks.
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type LeaderMigrationConfiguration struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// LeaderName is the name of the leader election resource that protects the migration
+	// E.g. 1-20-KCM-to-1-21-CCM
+	LeaderName string `json:"leaderName"`
```
cheftako, line 30: I wonder if we should specify the resource-lock. We need the two controller managers to use the same resource-lock across the migration for this to work. Seems like we should either specify a resource lock (defaulting to 'leases') or specify that it will use 'leases'.

jiahuif: I think we should have specified resourceLock considering the migration to lease API may not be completed everywhere. I've added a filed to indicate whether to use "endpoints" or "leases"

staging/src/k8s.io/controller-manager/config/types.go
```diff
@@ -0,0 +1,49 @@
...
+type LeaderMigrationConfiguration struct {
+	metav1.TypeMeta
+
+	// LeaderName is the name of the leader election resource that protects the migration
+	// E.g. 1-20-KCM-to-1-21-CCM
+	LeaderName string
+
+	// ResourceLock indicates the resource object type that will be used to lock
+	// Should be "leases" or "endpoints"
+	ResourceLock string
```
cheftako, line 34: Wish we had made this an enum, oh well.

jiahuif: Let's make room for future resource lock types, considering https://github.com/kubernetes/kubernetes/blob/8deb70c2ad6df1c02f88c0eec521c8ca1a4829f7/staging/src/k8s.io/component-base/config/v1alpha1/types.go#L50 They did not foresee `leases` either.

lavalamp: Why not go ahead and make this an enum? (i.e., give it its own type?)

I guess that is backwards compatible and can be added in the future.

cheftako: I'd like to be consistent with other instances of resource type. I think we should but that seems like clean up better done in a separate PR.

PR #94210, opened 2020-08-25
Systemd watchdog support

beekhof:
/kind feature

Type=notify supports detection and recovery of hung processes as well as terminated ones.
Currently only the terminated case is handled.  
    
Only activated if the distribution or admin configures a value for `WatchdogSec` in kubelet.serevice
See man(systemd.service) for details.

A dedicated timer is created based on the configured `WatchdogSec` value.

None

None

```release-note
Automated recovery of hung kubelet processes if WatchdogSec is defined in kubelet.serevice
```

### DIFF
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -18,6 +18,7 @@ package kubelet
 
 import (
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"math"
 	"net"
@@ -1868,24 +1869,24 @@ func (kl *Kubelet) syncLoop(updates <-chan kubetypes.PodUpdate, handler SyncHand
 	if watchdogTimeout, err := daemon.SdWatchdogEnabled(false); watchdogTimeout != 0 || err != nil {
 		switch {
 		case err != nil:
-			klog.Errorf("Reading watchdog configuration failed: %v", err)
+			klog.ErrorS(err, "Reading watchdog configuration failed")
 			break
 		case watchdogTimeout != 0:
 			notifyPeriod := watchdogTimeout / 3
 			loopTimeout := 2 * kl.resyncInterval
 			healthy := true
-			klog.Infof("Arming watchdog: configured watchdog timeout is %v, will report every %v, and stop reporting when sync loop is stalled for more than %v", watchdogTimeout, notifyPeriod, loopTimeout)
+			klog.InfoS("Arming watchdog", "watchdogTimeout", watchdogTimeout, "notifyPeriod", notifyPeriod, "loopTimeout", loopTimeout)
 			go wait.Forever(func() {
 				// check syncLoopMonitor: if it reports an old time, something is going wrong
 				if kl.clock.Now().After(kl.LatestLoopEntryTime().Add(loopTimeout)) {
 					if healthy {
 						healthy = false
-						klog.Warningf("SyncLoopMonitor not updated for at least %v, stopping to notify watchdog", loopTimeout)
+						klog.ErrorS(errors.New("SyncLoopMonitor not updated in time"), "Stopping to notify watchdog", "latestLoopEntryTime", kl.LatestLoopEntryTime(), "loopTimeout", loopTimeout)
 					}
 				} else {
 					if !healthy {
 						healthy = true
-						klog.Infof("SyncLoopMonitor was updated, restarting to notify watchdog")
+						klog.InfoS("SyncLoopMonitor was updated, restarting to notify watchdog")
 					}
 					_, _ = daemon.SdNotify(false, daemon.SdNotifyWatchdog)
 				}
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -33,6 +33,8 @@ import (
 
 	"k8s.io/client-go/informers"
 
+	"github.com/coreos/go-systemd/daemon"
+
 	cadvisorapi "github.com/google/cadvisor/info/v1"
 	"k8s.io/mount-utils"
 	"k8s.io/utils/integer"
@@ -1861,6 +1863,37 @@ func (kl *Kubelet) syncLoop(updates <-chan kubetypes.PodUpdate, handler SyncHand
 	housekeepingTicker := time.NewTicker(housekeepingPeriod)
 	defer housekeepingTicker.Stop()
 	plegCh := kl.pleg.Watch()
+
+	// If the starting systemd unit asks for a watchdog, notify periodically as long as the sync loop is called
+	if watchdogTimeout, err := daemon.SdWatchdogEnabled(false); watchdogTimeout != 0 || err != nil {
+		switch {
+		case err != nil:
+			klog.Errorf("Reading watchdog configuration failed: %v", err)
+			break
+		case watchdogTimeout != 0:
+			notifyPeriod := watchdogTimeout / 3
+			loopTimeout := 2 * kl.resyncInterval
+			healthy := true
+			klog.Infof("Arming watchdog: configured watchdog timeout is %v, will report every %v, and stop reporting when sync loop is stalled for more than %v", watchdogTimeout, notifyPeriod, loopTimeout)
+			go wait.Forever(func() {
+				// check syncLoopMonitor: if it reports an old time, something is going wrong
+				if kl.clock.Now().After(kl.LatestLoopEntryTime().Add(loopTimeout)) {
+					if healthy {
+						healthy = false
+						klog.Warningf("SyncLoopMonitor not updated for at least %v, stopping to notify watchdog", loopTimeout)
+					}
+				} else {
+					if !healthy {
+						healthy = true
+						klog.Infof("SyncLoopMonitor was updated, restarting to notify watchdog")
+					}
+					_, _ = daemon.SdNotify(false, daemon.SdNotifyWatchdog)
+				}
+			}, notifyPeriod)
+			break
+		}
+	}
+
 	const (
 		base   = 100 * time.Millisecond
 		max    = 5 * time.Second
--- a/pkg/kubelet/kubelet_test.go
+++ b/pkg/kubelet/kubelet_test.go
@@ -426,7 +426,9 @@ func TestWatchdog(t *testing.T) {
 	// configure watchdog
 	wusec := 3000000 // 3 secs
 	os.Setenv("NOTIFY_SOCKET", notifySocket)
+	defer os.Unsetenv("NOTIFY_SOCKET")
 	os.Setenv("WATCHDOG_USEC", fmt.Sprintf("%v", wusec))
+	defer os.Unsetenv("WATCHDOG_USEC")
 
 	// read notify messages in a loop
 	var msgCount int
--- a/pkg/kubelet/kubelet_test.go
+++ b/pkg/kubelet/kubelet_test.go
@@ -17,14 +17,18 @@ limitations under the License.
 package kubelet
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"os"
 	"sort"
 	"strconv"
+	"sync"
 	"testing"
 	"time"
 
+	"github.com/coreos/go-systemd/daemon"
 	cadvisorapi "github.com/google/cadvisor/info/v1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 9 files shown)


odinuge (commented):
Hey @beekhof! This is outside my expertise, so i'll defer the review for someone else. Otherwise the change looks good. It do however deserve an entry into the changelog imo., and maybe a test as well.

Maybe Derek has some knowledge about this.

/cc @derekwaynecarr
/ok-to-test

pkg/kubelet/kubelet.go
```diff
@@ -1899,6 +1901,9 @@ func (kl *Kubelet) syncLoopIteration(configCh <-chan kubetypes.PodUpdate, handle
 			handler.HandlePodSyncs([]*v1.Pod{pod})
 		}
 	case <-housekeepingCh:
+		if len(os.Getenv("WATCHDOG_USEC")) > 0 {
+			daemon.SdNotify(false, daemon.SdNotifyWatchdog)
+		}
```
sftim, line 1906: I'm puzzled; I'd expect to see the value from `os.Getenv("WATCHDOG_USEC")` used to calculate a period for a timer - usually you divide by 2.

Typical implementations run some code before the watchdog expires to check that the main event loop is not stuck, and then (satisfied that the app is not obviously broken) notify the supervisor to reset its watchdog timer.

sftim: I worry that this could result in a lot more notifies than systemd was expecting / designed for.

sftim:
> Value configured should be a multiple of `housekeepingPeriod`

To make documentation easier and users less likely to be confused, could the kubelet run another interval timer that's guaranteed to be less than half the watchdog interval? I can imagine someone running a kubelet with a watchdog interval of 3s and then _most of the time_ seeing it work, then wondering what's happening to their Pods and Nodes when systemd occasionally intervenes.

pkg/kubelet/kubelet.go
beekhof: > I worry that this could result in a lot more notifies than systemd was expecting / designed for.

Given that the threshold is in microseconds, I don't anticipate any issues notifying systemd every 2s.

> I'm puzzled; I'd expect to see the value from os.Getenv("WATCHDOG_USEC") used to calculate a period for a timer - usually you divide by 2.

I was going for minimally invasive, and `housekeepingPeriod` seemed sufficiently short, but I can create a dedicated channel derived from `WATCHDOG_USEC` instead (although at least x/3 would be preferable - we want to be *really* sure there is actually a problem at the point we initiate recovery).

> Typical implementations run some code before the watchdog expires to check that the main event loop is not stuck

Perhaps I am fundamentally misunderstanding golang or `syncLoopIteration()`, but IIUC `SdNotify()` would not be called if the main event loop was stuck.  Making the proposed code a sufficient indicator that kubelet is not hung. 

Deciding whether to expand "health" to non-hang scenarios (such as if `kl.sourcesReady.AllReady()` remains untrue for "too long") is beyond the scope of what I wish to achieve at this time.

sftim: Ok, so the main challenge is if the sysadmin configures the watchdog with an interval of say 1s and that's too short for the housekeeping interval to keep the watchdog alive.
(and for me it's also more about the behavior being confusing and hard to document, than that a 6s watchdog interval is unacceptable once you know that's a plausible minimum).

beekhof:
@sftim what do you suggest as a next step to getting this merged?

sftim:
This needs a review from a SIG Node reviewer to move forward. See https://github.com/kubernetes/kubernetes/pull/94210#issuecomment-679485092 for some tips.

pkg/kubelet/kubelet.go
```diff
@@ -1787,6 +1808,7 @@ func (kl *Kubelet) syncLoop(updates <-chan kubetypes.PodUpdate, handler SyncHand
 // 2.  handler:        the SyncHandler to dispatch pods to
 // 3.  syncCh:         a channel to read periodic sync events from
 // 4.  housekeepingCh: a channel to read housekeeping events from
+// 4.  watchdogCh:     a channel to read watchdog events from
```
odinuge, line 1811: nit: These numbers should probably be fixed.

pkg/kubelet/kubelet.go
```diff
@@ -1747,6 +1750,24 @@ func (kl *Kubelet) syncLoop(updates <-chan kubetypes.PodUpdate, handler SyncHand
...
 	defer housekeepingTicker.Stop()
 	plegCh := kl.pleg.Watch()
+
+	var watchdogCh <-chan time.Time
+	var watchdogTicker *time.Ticker
+
+	// If a value for WatchdogSec is defined in our systemd unit file,
+	// enable detection of hung kubelet processes.  Require *at least* 2
+	// genuinely missed heartbeats before recovery action is taken
+	watchdogEnv, err := strconv.ParseInt(os.Getenv("WATCHDOG_USEC"), 10, 64)
```
odinuge, line 1760: Should we add cases to check where the env var has an unexpected value? This implementation will just ignore it and don't log anything.

odinuge: After reading here: https://www.freedesktop.org/software/systemd/man/sd_watchdog_enabled.html, it looks like `WATCHDOG_PID` is related as well. Should we handle that?

pkg/kubelet/kubelet.go
```diff
@@ -1909,6 +1932,8 @@ func (kl *Kubelet) syncLoopIteration(configCh <-chan kubetypes.PodUpdate, handle
 				klog.Errorf("Failed cleaning pods: %v", err)
 			}
 		}
+	case <-watchdogCh:
```
odinuge, line 1935: Are we 100% sure that this is the correct place to do this? I think i'd prefer if you could lay out some test cases and create e2e tests to make sure it works as expected. Are we sure that we don't end up killing a healthy kubelet before it properly starts, or when the apiserver is down eg.?

Would also be nice with a writeup about when we intend to stop reporting `SdNotify` (and when we do want to report it), to make it easier to debug in case it behaves in a non-understandable manner. Since the env-var is in micro secs I am a bit concerned that small values may results in weird behavior, but that is just a thought.

pkg/kubelet/kubelet.go
```diff
@@ -1747,6 +1750,24 @@ func (kl *Kubelet) syncLoop(updates <-chan kubetypes.PodUpdate, handler SyncHand
...
 	plegCh := kl.pleg.Watch()
+
+	var watchdogCh <-chan time.Time
+	var watchdogTicker *time.Ticker
+
+	// If a value for WatchdogSec is defined in our systemd unit file,
+	// enable detection of hung kubelet processes.  Require *at least* 2
+	// genuinely missed heartbeats before recovery action is taken
+	watchdogEnv, err := strconv.ParseInt(os.Getenv("WATCHDOG_USEC"), 10, 64)
+	watchdogPeriod := time.Duration(watchdogEnv / 3)
```
odinuge, line 1761: I believe `time.Duration` tracks nano-secs and not micro-secs.

odinuge (commented):
Hey. This looks nice, but just a few questions. Would also make sense to ask about this in the sig-node weekly meeting to get some more feedback. :smile:

pkg/kubelet/kubelet.go
slintes: Hi, this is Marc, @beekhof asked me to continue with this. PR. Thanks for your feedback so far! I will update the PR soonish, just need to get a bit more familiar with the kubelet code.
You are right with the comments here, I will address them by using `func SdWatchdogEnabled(unsetEnvironment bool) (time.Duration, error)` of the watchdog lib Andrew already introduced, see https://github.com/beekhof/kubernetes/blob/d462e1e8d7e45894bf2abb7dc5d939f1bd3cae68/vendor/github.com/coreos/go-systemd/daemon/watchdog.go#L35

pkg/kubelet/kubelet.go
slintes: - I will check if I find a better place for this code, and add tests as soon as I'm more familiar with the kubelet / test code
- We should stop notifying when kubelet is not healthy anymore. Will see if I find something useful to determine this. If you have ideas, please let me know
- the env var is in micro seconds, but the value which actually is set in the service config is in seconds, so I'm less concerned about this (even though the smallest value of 1 sec still is not that much...). See https://www.freedesktop.org/software/systemd/man/systemd.service.html#WatchdogSec=

slintes:
> Are we 100% sure that this is the correct place to do this?

I'm not so sure anymore. I don't have a good idea how long one run of `syncLoopIteration()` takes, but there might be some risk that it prevents a notify in time for low watchdog timeouts?

@sftim @odinuge What do you think about something like this instead, in `syncLoop()`? This will run the watchdog notify in its own loop, and it stops notifying in case the `syncLoopMonitor` wasn't updated for some time.

```
	// If the starting systemd unit asks for a watchdog, notify periodically
	if watchdogDuration, err := daemon.SdWatchdogEnabled(false); watchdogDuration != 0 || err != nil {
		switch {
		case err != nil:
			klog.Errorf("Reading watchdog configuration failed: %v", err)
			break
		case watchdogDuration != 0:
			go wait.Forever(func(){
				// check syncLoopMonitor: if it reports an old time, something is going wrong
				healthyPeriod := 30 * time.Second
				if kl.LatestLoopEntryTime().Before(time.Now().Add(-healthyPeriod)) {
					klog.Errorf("syncLoopMonitor not updated for at least %v, not going to notify watchdog", healthyPeriod)
				} else {
					_, _ = daemon.SdNotify(false, daemon.SdNotifyWatchdog)
				}
			}, watchdogDuration / 3)
			break
		}
	}

**Comment by slintes** [comment]
I went with the approach suggested in above comment, and added a unit test.
Ready for another round of review :slightly_smiling_face:

File: pkg/kubelet/kubelet.go [hunk 6]
```diff
@@ -1750,24 +1750,26 @@ func (kl *Kubelet) syncLoop(updates <-chan kubetypes.PodUpdate, handler SyncHand
...
+			klog.Errorf("Reading watchdog configuration failed: %v", err)
+			break
+		case watchdogTimeout != 0:
+			notifyPeriod := watchdogTimeout / 3
+			loopTimeout := 2 * kl.resyncInterval
+			klog.Infof("Arming watchdog: configured watchdog timeout is %v, will report every %v, and stop reporting when sync loop is stalled for more than %v", watchdogTimeout, notifyPeriod, loopTimeout)
+			go wait.Forever(func() {
+				// check syncLoopMonitor: if it reports an old time, something is going wrong
+				if kl.clock.Now().After(kl.LatestLoopEntryTime().Add(loopTimeout)) {
+					klog.Warningf("SyncLoopMonitor not updated for at least %v, not going to notify watchdog", loopTimeout)
```
sftim, line 1766: Sorry to quibble, but I want to know: could this end up spamming the logs with many reports of a stall?

slintes: Yes, fair point, fixed.
And rebased...

slintes:
Hi, can we another review please? Or are we missing anything?
Thanks!  :)

slintes:
/remove-lifecycle stale

any news on this?

@ehashman I see a lot of PRs which needs reviewer. I guess we just have to wait until it's our turn?

ehashman:
@slintes yes, that's right... we have a lot of PRs that need review but limited reviewer time and there's no priority set on this PR (and it doesn't have a linked issue it fixes)

slintes:
@ehashman we created an issue (thanks Andrew!)
And we weren't aware that we add a priority ourself.

/priority important-longterm
Fixes #98981

SergeyKanzhelev:
likely needs rebase baecase of bazel

slintes:
rebased and squashed commits

jsturtevant:
it looks like this is a noop for Windows but would like to verify by running the Windows tests. It is a optional test and doesn't block merging.  

/test pull-kubernetes-e2e-aks-engine-windows-containerd

beekhof:
Real life example where this PR would have helped:
   https://bugzilla.redhat.com/show_bug.cgi?id=1948052

pkg/kubelet/kubelet.go
```diff
@@ -1853,6 +1855,37 @@ func (kl *Kubelet) syncLoop(updates <-chan kubetypes.PodUpdate, handler SyncHand
...
+	if watchdogTimeout, err := daemon.SdWatchdogEnabled(false); watchdogTimeout != 0 || err != nil {
+		switch {
+		case err != nil:
+			klog.Errorf("Reading watchdog configuration failed: %v", err)
+			break
+		case watchdogTimeout != 0:
+			notifyPeriod := watchdogTimeout / 3
+			loopTimeout := 2 * kl.resyncInterval
+			healthy := true
+			klog.Infof("Arming watchdog: configured watchdog timeout is %v, will report every %v, and stop reporting when sync loop is stalled for more than %v", watchdogTimeout, notifyPeriod, loopTimeout)
```
endocrimes, line 1869: `klog.InfoS` here too?

pkg/kubelet/kubelet_test.go
```diff
@@ -381,6 +385,102 @@ func newTestPods(count int) []*v1.Pod {
...
+	laddr := net.UnixAddr{
+		Name: notifySocket,
+		Net:  "unixgram",
+	}
+	listener, err := net.ListenUnixgram("unixgram", &laddr)
+	require.NoError(t, err, "Failed to setup fake systemd server")
+
+	// configure watchdog
+	wusec := 3000000 // 3 secs
+	os.Setenv("NOTIFY_SOCKET", notifySocket)
```
endocrimes, line 425: can you also add `defer os.Unsetenv`'s to these too please - otherwise adding other tests that interact with this code will potentially behave unexpectedly.

endocrimes (commented):
This approach seems reasonable to me. The tests have the potential to be a little racy which isn't ideal, and rely on sleeping a lot, but I can't think of a better way to test them that wouldn't complicate the code you're testing (decoupling the forever loop and sleeping from the other logic)

pkg/kubelet/kubelet.go
```diff
@@ -1853,6 +1855,37 @@ func (kl *Kubelet) syncLoop(updates <-chan kubetypes.PodUpdate, handler SyncHand
 	housekeepingTicker := time.NewTicker(housekeepingPeriod)
 	defer housekeepingTicker.Stop()
 	plegCh := kl.pleg.Watch()
+
+	// If the starting systemd unit asks for a watchdog, notify periodically as long as the sync loop is called
+	if watchdogTimeout, err := daemon.SdWatchdogEnabled(false); watchdogTimeout != 0 || err != nil {
+		switch {
+		case err != nil:
+			klog.Errorf("Reading watchdog configuration failed: %v", err)
```
pacoxu, line 1863: For structured log check, it should be 
```
			klog.ErrorS(err, "Reading watchdog configuration failed")
```
After `git rebase upstream/master`, run `hack/verify-structured-logging.sh` to check.

> /Users/xupaco/git/gopath/src/k8s.io/kubernetes/pkg/kubelet/kubelet.go:1868:4: unstructured logging function "Errorf" should not be used
> /Users/xupaco/git/gopath/src/k8s.io/kubernetes/pkg/kubelet/kubelet.go:1874:4: unstructured logging function "Infof" should not be used
> /Users/xupaco/git/gopath/src/k8s.io/kubernetes/pkg/kubelet/kubelet.go:1880:7: unstructured logging function "Warningf" should not be used
> /Users/xupaco/git/gopath/src/k8s.io/kubernetes/pkg/kubelet/kubelet.go:1885:7: unstructured logging function "Infof" should not be used
>

slintes:
@endocrimes @pacoxu thanks for the reviews, I fixed the logging and unset the test env vars.

The `hack/verify-structured-logging.sh` doesn't work great for me though. As it is, it complains about absolute import paths. When I change line 48 to `  migrated_packages+=("./$line")` (`.` instead of `$KUBE_ROOT`), it reports tons of errors not related to logging, like undeclared names and failed imports...? 🤔

pacoxu:
> The `hack/verify-structured-logging.sh` doesn't work great for me though. As it is, it complains about absolute import paths. When I change line 48 to ` migrated_packages+=("./$line")` (`.` instead of `$KUBE_ROOT`), it reports tons of errors not related to logging, like undeclared names and failed imports...? 🤔

We may discuss it in a new issue for the verify script if it would be a problem.

slintes:
that didn't went well... rebased

PR #94214, opened 2020-08-25
test(endpointslice): deflake TestSyncEndpoints

knight42:

**What type of PR is this?**

/kind flake

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93605/pull-kubernetes-bazel-test/1297925434742149121
```
=== RUN   TestSyncEndpoints/Endpoints_with_1001_addresses_-_1_should_not_be_mirrored
W0824 16:06:59.118771      23 reconciler.go:122] 1 addresses in default/testing-sync-endpoints Endpoints were skipped due to exceeding MaxEndpointsPerSubset
    endpointslicemirroring_controller_test.go:207: Expected 2 additional client actions, got 1: []testing.Action{testing.CreateActionImpl{ActionImpl:testing.ActionImpl{Namespace:"default", Verb:"create", Resource:schema.GroupVersionResource{Group:"discovery.k8s.io", Version:"v1beta1", Resource:"endpointslices"}, Subresource:""}, Name:"", Object:(*v1beta1.EndpointSlice)(0xc00037f760)}}
--- FAIL: TestSyncEndpoints (0.41s)
```

The test `Endpoints with 1001 addresses - 1 should not be mirrored` expects 2 additional creation actions, one for creating `EndpointSlices`, the other for creating warning event https://github.com/kubernetes/kubernetes/blob/4db3a096ce8ac730b2280494422e1c4cf5fe875e/pkg/controller/endpointslicemirroring/reconciler.go#L119-L125

This test is flaky because we expects something has been done in a separate goroutine without any synchronization(a super common cause of flakes). The warning event is created in a separate goroutine, so it is not guaranteed the expected warning event is created after reconcilation. The error could be reproduced by the following patch:
```patch
diff --git staging/src/k8s.io/client-go/tools/record/event.go staging/src/k8s.io/client-go/tools/record/event.go
index fce4410c0dd..f39a24e0318 100644
--- staging/src/k8s.io/client-go/tools/record/event.go
+++ staging/src/k8s.io/client-go/tools/record/event.go
@@ -341,6 +341,7 @@ func (recorder *recorderImpl) generateEvent(object runtime.Object, annotations m
        go func() {
                // NOTE: events should be a non-blocking operation
+               time.Sleep(time.Second)
                recorder.Action(watch.Added, event)
        }()
 }
```

IMHO this flake does not indicate an internal bug, waiting until the expected event shows up seems to be sufficient.

Part of #93605

### Alternative

We could probably refactor the `endpointSliceMirroringController` to ultilize [`FakeRecorder`](https://github.com/kubernetes/kubernetes/blob/4db3a096ce8ac730b2280494422e1c4cf5fe875e/pkg/controller/testutil/test_utils.go#L352-L358) in testing, which creates events synchronously and saves all events https://github.com/kubernetes/kubernetes/blob/4db3a096ce8ac730b2280494422e1c4cf5fe875e/pkg/controller/testutil/test_utils.go#L375-L386, so that we could check the events directly.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/controller/endpointslicemirroring/BUILD
+++ b/pkg/controller/endpointslicemirroring/BUILD
@@ -65,6 +65,7 @@ go_test(
         "//staging/src/k8s.io/apimachinery/pkg/runtime/schema:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/types:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/util/rand:go_default_library",
+        "//staging/src/k8s.io/apimachinery/pkg/util/wait:go_default_library",
         "//staging/src/k8s.io/client-go/informers:go_default_library",
         "//staging/src/k8s.io/client-go/kubernetes/fake:go_default_library",
         "//staging/src/k8s.io/client-go/kubernetes/scheme:go_default_library",
--- a/pkg/controller/endpointslicemirroring/endpointslicemirroring_controller_test.go
+++ b/pkg/controller/endpointslicemirroring/endpointslicemirroring_controller_test.go
@@ -25,10 +25,12 @@ import (
 	v1 "k8s.io/api/core/v1"
 	discovery "k8s.io/api/discovery/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes/fake"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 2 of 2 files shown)


liggitt:
cc @robscott

liggitt:
I agree this is a test-only issue and waiting for the asynchronous operation is reasonable here

Targeting for beginning of 1.20, we're soaking release-1.19 aiming for release tomorrow

/lgtm
/approve
/milestone v1.20

justaugustus:
ref: https://github.com/kubernetes/kubernetes/issues/94263

PR #94219, opened 2020-08-25
`exec` completion should behave the same as `logs`

### LINKED ISSUE #890: Autocomplete for container name in kubectl exec
Auto-completion for container name in `kubectl exec`. 
`kubectl exec -it mypod -c [TAB]`

Currently it proposes pod names after `-c` (in fact after anything ~unknown it seems).

More generally, it should stop proposing pod names after the first pod name in cmdline, especially after `--`.

(tested with kubectl v1.18.3)

**Why is this needed**:
Autocompletion allow smooth workload exploration and usage: no need to explicitly discover with `kubectl describe pod mypod` then copy/paste the container name in `kubectl exec` cmdline.

(container name autocompletion already works great for `kubectl logs mypod`)

rayw000:
### What changed
1. `kubectl exec` command completion should have a similar behavior like
`kubectl logs`. That is, when you hit `TAB` after `kubectl exec <flags> <pod
name> <with/without -c>`, container completion should always be triggered.
2. Previously the `exec` command doesn't take inline [Container] argument, this
commit fixed it.

Fix https://github.com/kubernetes/kubectl/issues/890

### What's not fixed yet:
Still cannot stop completing even if a `-c <container_name>` argument has already existed. However completion doesn't work after `-c` if there has been an inline [Container] argument. Seems the completion always expects an inline [Container] argument. This is the same to `kubectl logs`.

### Release Note
```release-note
kubectl exec command performs completion for container names.
```

### DIFF
--- a/pkg/kubectl/cmd/cmd.go
+++ b/pkg/kubectl/cmd/cmd.go
@@ -258,11 +258,11 @@ __kubectl_custom_func() {
             __kubectl_get_resource
             return
             ;;
-        kubectl_logs)
+        kubectl_logs | kubectl_exec)
             __kubectl_require_pod_and_container
             return
             ;;
-        kubectl_exec | kubectl_port-forward | kubectl_top_pod | kubectl_attach)
+        kubectl_port-forward | kubectl_top_pod | kubectl_attach)
             __kubectl_get_resource_pod
             return
             ;;
--- a/staging/src/k8s.io/kubectl/pkg/cmd/exec/exec.go
+++ b/staging/src/k8s.io/kubectl/pkg/cmd/exec/exec.go
@@ -174,7 +174,7 @@ func (p *ExecOptions) Complete(f cmdutil.Factory, cmd *cobra.Command, argsIn []s
 
 	// Also respect inline [Container] argument.
 	// This behavior is very similar to `kubectl logs`.
-	if len(argsIn) > 1 && argsLenAtDash > 1 {
+	if argsLenAtDash > 1 {
 		p.ContainerName = argsIn[1]
 	}
 	if argsLenAtDash > -1 {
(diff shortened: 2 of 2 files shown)


thomas-riccardi:
Until `--` is mandatory to separate the commandline to be executed in the pod, I'm not sure it's OK to accept container name without `-c`.

rayw000:
@thomas-riccardi 
Thanks for your reply. Actually in this commit, `exec` command will accept inline CONTAINER argument in a very strict situation, which must satisfy both the following two conditions simultaneously (refer: https://github.com/kubernetes/kubernetes/pull/94219/files#diff-5c46adf16c80aadcd66a4fd20ee6ad59R177).
- `len(argsIn) > 1`: there are at least two inline arguments (include inline arguments both before and after `--`)
- `argsLenAtDash > 1`: `--` detected and there are at least two inline arguments before `--`

> Perhaps `len(argsIn) > 1` could be removed.


If there is no `--` detected, `argsLenAtDash` will always be `-1`.
`argsLenAtDash > COUNT` means there are `COUNT + 1` inline argument(s) after `exec` before `--`.

So, only commands in the following format would accept inline CONTAINER argument as container name.
```
kubectl exec POD [flags] CONTAINER -- [COMMAND] [...]
```

Hope it is useful. 🙂

rayw000:
Hi @eddiesimeon
Would you please give me some advice about why this PR is failed? 
I'm confused that it is an IPv6 network failure, but diff shows nothing related to network. What could I do to pass this test?
Thanks a lot!

rayw000:
I've removed `len(argsIn) > 1` because it is always true when `argsLenAtDash > 1`.

soltysh (changes requested):
/hold
I agree with @thomas-riccardi on this one, I'd prefer consistency to go towards deprecating that functionality in logs rather than adding it in exec.

rayw000:
@soltysh No problem. Currently I'm working on the issue mentioned in top of my description _What's not fixed yet_.

Issue #94223, opened 2020-08-25
[Flaky Test] [k8s.io] MirrorPodWithGracePeriod when create a mirror pod mirror pod termination should satisfy grace period when static pod is deleted [NodeConformance]


hasheddan:
**Which jobs are flaking**:

`node-kubelet-master (ci-kubernetes-node-kubelet)`

**Which test(s) are flaking**:

`[k8s.io] MirrorPodWithGracePeriod when create a mirror pod mirror pod termination should satisfy grace period when static pod is deleted [NodeConformance]`

**Testgrid link**:

https://testgrid.k8s.io/sig-release-master-blocking#node-kubelet-master

**Reason for failure**:

```
_output/local/go/src/k8s.io/kubernetes/test/e2e_node/mirror_pod_grace_period_test.go:54
Unexpected error:
    <*errors.StatusError | 0xc000f2ab40>: {
        ErrStatus: {
            TypeMeta: {Kind: "", APIVersion: ""},
            ListMeta: {
                SelfLink: "",
                ResourceVersion: "",
                Continue: "",
                RemainingItemCount: nil,
            },
            Status: "Failure",
            Message: "pods \"graceful-pod-318b04ef-32f9-4be7-be28-72dbb4dcc6e5-tmp-node-e2e-0ead0b6b-cos-81-12871-1190-0\" not found",
            Reason: "NotFound",
            Details: {
                Name: "graceful-pod-318b04ef-32f9-4be7-be28-72dbb4dcc6e5-tmp-node-e2e-0ead0b6b-cos-81-12871-1190-0",
                Group: "",
                Kind: "pods",
                UID: "",
                Causes: nil,
                RetryAfterSeconds: 0,
            },
            Code: 404,
        },
    }
    pods "graceful-pod-318b04ef-32f9-4be7-be28-72dbb4dcc6e5-tmp-node-e2e-0ead0b6b-cos-81-12871-1190-0" not found
occurred
_output/local/go/src/k8s.io/kubernetes/test/e2e_node/mirror_pod_grace_period_test.go:72
```

**Anything else we need to know**:

/cc @kubernetes/ci-signal @kubernetes/sig-node-bugs 
/sig node
/priority critical-urgent
/milestone v1.19

sjenning:
History of this failure

https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&test=MirrorPodWithGracePeriod

[...]

PR #94229, opened 2020-08-25
Verify running mirror pod has running containers

derekwaynecarr:
**What type of PR is this?**
/kind bug
/kind flake

Investigating logs for https://github.com/kubernetes/kubernetes/issues/94223 makes me suspicious that looking at pod phase alone is not sufficient.  Adding more verification in the test to ensure pod is running and all containers are running.

None

```release-note
NONE
```

### DIFF
--- a/test/e2e_node/mirror_pod_test.go
+++ b/test/e2e_node/mirror_pod_test.go
@@ -192,6 +192,11 @@ func checkMirrorPodRunning(cl clientset.Interface, name, namespace string) error
 	if pod.Status.Phase != v1.PodRunning {
 		return fmt.Errorf("expected the mirror pod %q to be running, got %q", name, pod.Status.Phase)
 	}
+	for i := range pod.Status.ContainerStatuses {
+		if pod.Status.ContainerStatuses[i].State.Running == nil {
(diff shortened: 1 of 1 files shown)


derekwaynecarr:
/assign @sjenning 

still testing this locally, but seeing the pleg event report unknown->running , the api server getting that state, and the test responding in such short time sequence is making me question the reported pod phase the test is operating against.

hasheddan (approved):
/lgtm
/hold for @sjenning

sjenning:
this makes sense to me.  we can do a few runs of e2e in this PR and see if we can't recreate.

derekwaynecarr:
ironically, this is breaking for me.  need more coffee.

/hold

derekwaynecarr:
this is an attempt to see if we are ever hitting this state in pleg calculation where pod phase is wrong.

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_pods.go#L1459

in that scenario, no containers are running, but pod is still running.

this seems to align with situations where pleg showed unknown -> running just prior which implies the container may have restarted....

sjenning:
re
/lgtm
/priority low

Should eliminate a potential cause of this test flaking. If it is flaking.

sjenning:
gah, wrong bot
/priority important-soon

PR #94231, opened 2020-08-25
Changing label to NodeAlphaFeature to include tests in Node Testgrid

javidiaz:
**What type of PR is this?**
/kind cleanup

This PR changes the label of the e2e tests of the [alpha feature setHostnameAsFQDN](https://github.com/kubernetes/enhancements/issues/1797) to include them in the Node Alpha Feature test grid

### DIFF
--- a/test/e2e/node/pod_hostnamefqdn.go
+++ b/test/e2e/node/pod_hostnamefqdn.go
@@ -63,7 +63,7 @@ func testPod() *v1.Pod {
 	return pod
 }
 
-var _ = SIGDescribe("Hostname of Pod [Feature:SetHostnameAsFQDN]", func() {
+var _ = SIGDescribe("Hostname of Pod [Feature:SetHostnameAsFQDN][NodeAlphaFeature:SetHostnameAsFQDN]", func() {
 	f := framework.NewDefaultFramework("hostfqdn")
 
 	/*
@@ -71,7 +71,7 @@ var _ = SIGDescribe("Hostname of Pod [Feature:SetHostnameAsFQDN]", func() {
 	   Testname: Create Pod without fully qualified domain name (FQDN)
 	   Description: A Pod that does not define the subdomain field in it spec, does not have FQDN.
 	*/
-	ginkgo.It("a pod without subdomain field does not have FQDN [Feature:SetHostnameAsFQDN]", func() {
+	ginkgo.It("a pod without subdomain field does not have FQDN", func() {
 		pod := testPod()
 		pod.Spec.Containers[0].Command = []string{"sh", "-c", "echo $(hostname)';'$(hostname -f)';'"}
 		output := []string{fmt.Sprintf("%s;%s;", pod.ObjectMeta.Name, pod.ObjectMeta.Name)}
@@ -85,7 +85,7 @@ var _ = SIGDescribe("Hostname of Pod [Feature:SetHostnameAsFQDN]", func() {
 	   Description: A Pod that does not define the subdomain field in it spec, does not have FQDN.
 	                Hence, SetHostnameAsFQDN feature has no effect.
 	*/
-	ginkgo.It("a pod without FQDN is not affected by SetHostnameAsFQDN field [Feature:SetHostnameAsFQDN]", func() {
+	ginkgo.It("a pod without FQDN is not affected by SetHostnameAsFQDN field", func() {
 		pod := testPod()
 		// Setting setHostnameAsFQDN field to true should have no effect.
 		setHostnameAsFQDN := true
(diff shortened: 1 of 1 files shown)


test/e2e/node/pod_hostnamefqdn.go
```diff
@@ -63,15 +63,15 @@ func testPod() *v1.Pod {
...
+var _ = SIGDescribe("Hostname of Pod [NodeAlphaFeature:SetHostnameAsFQDN]", func() {
 	f := framework.NewDefaultFramework("hostfqdn")
 
 	/*
 	   Release: v1.19
 	   Testname: Create Pod without fully qualified domain name (FQDN)
 	   Description: A Pod that does not define the subdomain field in it spec, does not have FQDN.
 	*/
-	ginkgo.It("a pod without subdomain field does not have FQDN [Feature:SetHostnameAsFQDN]", func() {
+	ginkgo.It("a pod without subdomain field does not have FQDN [NodeAlphaFeature:SetHostnameAsFQDN]", func() {
```
SergeyKanzhelev, line 74: actually I don't think you need to duplicate the NodeAlphaFeature:SetHostnameAsFQDN] on every test. It would be inherited from the describe call.

javidiaz: great, I updated it. Thanks.

SergeyKanzhelev:
/lgtm
/assign @dashpole 
/priority important-soon

setting priority to important soon as we want these tests to run

SergeyKanzhelev:
tests failures are legit. Seems like feature flag needs to be added to test-infra definitions

javidiaz:
> tests failures are legit. Seems like feature flag needs to be added to test-infra definitions

Agree, I think the feature flag is one issue and it also seems hostname -f returns error in the ipv6 tests. I will add the flag and then see what I can do about the ipv6 issue.

javidiaz:
@SergeyKanzhelev I think it is good now. I added Feature so those CIs that failed skipped the test. I guess we have to wait until it merges for node-kubelet-alpha to pick it up?

test/e2e/node/pod_hostnamefqdn.go
```diff
@@ -63,15 +63,15 @@ func testPod() *v1.Pod {
 	return pod
 }
 
-var _ = SIGDescribe("Hostname of Pod [Feature:SetHostnameAsFQDN]", func() {
+var _ = SIGDescribe("Hostname of Pod [Feature:SetHostnameAsFQDN][NodeAlphaFeature:SetHostnameAsFQDN]", func() {
```
spiffxp, line 66: Do you need the `[Feature:SetHostnameAsFQDN]` tag? I can't find any jobs in kubernetes/test-infra that specifically mention it

SergeyKanzhelev: it was an original change:

image


Based on @javidiaz reply, adding Feature removing it from one of the queries that runs as part of a CI.

@javidiaz did you figure out which test is running these ginko tests? I'm curious if adding Feature is the right fix here.


I think this is lgtm even with Feature enabled. At least we will start running these tests as part of node-kubelet-alpha test.

/lgtm

javidiaz: @spiffxp Without having Feature tag, some CIs that do not have Alpha features enabled run them (e.g. pull-kubernetes-e2e-kind). I found other NodeAlphaFeature e2e tests that have both Feature and NodeAlphaFeature (e.g., TTLAfterFinished, ShareProcessNamespace, CPUManager). I am happy to change things to whatever you think it is right.

@SergeyKanzhelev I do not think any of the tests checking this PR are running these ginko tests right now.

spiffxp: 🤦  right, thanks, that makes sense in context.  To dedupe, we'd have to go update a bunch of skip regexes to now also include `\[NodeAlphaFeature:.+\]` and hey why not `\[NodeSpecialFeature:.+\]` while we're at it.

Instead of having to change a bunch of jobs, would it make more sense to instead use `[NodeAlphaFeature][Feature:Foo]` and update the few node jobs that test these feature?

spiffxp: (I'm fine with this change being done as a followup since it requires a few cross-repo PR's to get it right, just want to get consensus and a followup issue open before putting this into the merge pool)

SergeyKanzhelev: I really like `[NodeAlphaFeature][Feature:Foo]` approach going forward.

SergeyKanzhelev: i can help with the follow up PR

javidiaz: Thanks guys. Just FYI there are also tests with [NodeFeature:.+][Feature:.+] tags. 

Now that 1.19 is out I think we can merge this, right? :)

SergeyKanzhelev:
@dashpole please approve, this is ready to merge and is helping improve test coverage by actually running these tests

test/e2e/node/pod_hostnamefqdn.go
spiffxp: opened https://github.com/kubernetes/kubernetes/issues/94289

PR #93638, opened 2020-08-03
Avoid dereferencing same endpoint twice on the deletion or update of a service

sbangari:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:
When a service is deleted or updated, the service becomes stale and is cleaned up. In the current logic this service cleanup can be triggered multiple times. Except the first cleanup the subsequent cleanup calls are a no-op. But as part of service cleanup the endpoints in the endpoint map are attempted to be cleaned up a second time and dereferenced twice.

In a scenario where the remote endpoints are shared across two services, the second dereference causes the remote endpoint to be deleted from HNS. This breaks connectivity to the service.
**Which issue(s) this PR fixes**:

Fixes #93229

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #94232, opened 2020-08-25
Add e2e multiple endpoint services test

### LINKED ISSUE #93397: E2e Test for multiple services with endpoints
**What happened**:
https://github.com/kubernetes/kubernetes/issues/93229 indicates that we're missing an e2e test for.

This specific issue is just a tracker to create a new E2E test that demonstrated 93229 .

swetharepakula:
- ensures that when multiple services point to the same pod, each
 service can be deleted without affecting others

**What type of PR is this?**

This PR adds e2e test coverage for the bug found in #93229 in which multiple services that point to the same backend IPs should continue to have connectivity when one of the services is deleted.

Fixes #93397

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/framework/network/utils.go
+++ b/test/e2e/framework/network/utils.go
@@ -611,11 +611,11 @@ func (config *NetworkingTestConfig) createNodePortServiceSpec(svcName string, se
 }
 
 func (config *NetworkingTestConfig) createNodePortService(selector map[string]string) {
-	config.NodePortService = config.createService(config.createNodePortServiceSpec(nodePortServiceName, selector, false))
+	config.NodePortService = config.CreateService(config.createNodePortServiceSpec(nodePortServiceName, selector, false))
 }
 
 func (config *NetworkingTestConfig) createSessionAffinityService(selector map[string]string) {
-	config.SessionAffinityService = config.createService(config.createNodePortServiceSpec(sessionAffinityServiceName, selector, true))
+	config.SessionAffinityService = config.CreateService(config.createNodePortServiceSpec(sessionAffinityServiceName, selector, true))
 }
 
 // DeleteNodePortService deletes NodePort service.
@@ -651,7 +651,8 @@ func (config *NetworkingTestConfig) createTestPods() {
 	}
 }
 
-func (config *NetworkingTestConfig) createService(serviceSpec *v1.Service) *v1.Service {
+// CreateService creates the provided service in config.Namespace and returns created service
+func (config *NetworkingTestConfig) CreateService(serviceSpec *v1.Service) *v1.Service {
 	_, err := config.getServiceClient().Create(context.TODO(), serviceSpec, metav1.CreateOptions{})
 	framework.ExpectNoError(err, fmt.Sprintf("Failed to create %s service: %v", serviceSpec.Name, err))
 
--- a/test/e2e/network/networking.go
+++ b/test/e2e/network/networking.go
@@ -226,6 +226,38 @@ var _ = SIGDescribe("Networking", func() {
 			config.DialFromEndpointContainer("udp", config.NodeIP, config.NodeUDPPort, config.MaxTries, 0, config.EndpointHostnames())
 		})
 
+		// This test ensures that in a situation where multiple services exist with the same selector,
+		// deleting one of the services does not affect the connectivity of the remaining service
+		ginkgo.It("should function for multiple endpoint-Services with same selector", func() {
+			config := e2enetwork.NewNetworkingTestConfig(f, false, false)
+			ginkgo.By("creating a second service with same selector")
+			svc2, httpPort := createSecondNodePortService(f, config)
+
+			// original service should work
+			ginkgo.By(fmt.Sprintf("dialing(http) %v (endpoint) --> %v:%v (config.clusterIP)", config.EndpointPods[0].Name, config.ClusterIP, e2enetwork.ClusterHTTPPort))
+			config.DialFromEndpointContainer("http", config.ClusterIP, e2enetwork.ClusterHTTPPort, config.MaxTries, 0, config.EndpointHostnames())
+
+			ginkgo.By(fmt.Sprintf("dialing(http) %v (endpoint) --> %v:%v (nodeIP)", config.EndpointPods[0].Name, config.NodeIP, config.NodeHTTPPort))
+			config.DialFromEndpointContainer("http", config.NodeIP, config.NodeHTTPPort, config.MaxTries, 0, config.EndpointHostnames())
+
+			// Dial second service
+			ginkgo.By(fmt.Sprintf("dialing(http) %v (endpoint) --> %v:%v (svc2.clusterIP)", config.EndpointPods[0].Name, svc2.Spec.ClusterIP, e2enetwork.ClusterHTTPPort))
+			config.DialFromEndpointContainer("http", svc2.Spec.ClusterIP, e2enetwork.ClusterHTTPPort, config.MaxTries, 0, config.EndpointHostnames())
+
+			ginkgo.By(fmt.Sprintf("dialing(http) %v (endpoint) --> %v:%v (nodeIP)", config.EndpointPods[0].Name, config.NodeIP, httpPort))
+			config.DialFromEndpointContainer("http", config.NodeIP, httpPort, config.MaxTries, 0, config.EndpointHostnames())
+
+			ginkgo.By("deleting the original node port service")
+			config.DeleteNodePortService()
+
+			// Second service should continue to function unaffected
+			ginkgo.By(fmt.Sprintf("dialing(http) %v (endpoint) --> %v:%v (svc2.clusterIP)", config.EndpointPods[0].Name, svc2.Spec.ClusterIP, e2enetwork.ClusterHTTPPort))
+			config.DialFromEndpointContainer("http", svc2.Spec.ClusterIP, e2enetwork.ClusterHTTPPort, config.MaxTries, 0, config.EndpointHostnames())
+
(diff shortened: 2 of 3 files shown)


aojea:
this should merge with the fix and its own test, not modifying a current one

swetharepakula:
To my understanding, the windows bug (#93229) and e2e test (#93397) issue were separate. This aims resolve the e2e issue.

I didn't create a new test per @jayunit100 suggestions in #93397, but if a new test would be better I can do that.

aojea:
sorry, maybe I did not understand it well, but if this test merge that means that some CI jobs using this test will fail?
If that's the case, we should try to avoid that, and submit the fix with a test that covers it.

In the other hand, if this is a new test to increase the coverage, I personally think that this should be a new test, not a modification of a previous one

swetharepakula:
I was wondering about CI as well. I guess my question is how soon does this test coverage need to be merged in? From the issue, it felt like a test was wanted sooner than later. If we can wait, that might be the best approach. Otherwise I can modify this to have the `Linux Only` tag as I believe only Windows has the bug.

In regards to breaking this out into its own test, I will make that change.

Thanks @aojea !

aojea:
> I was wondering about CI as well. I guess my question is how soon does this test coverage need to be merged in? From the issue, it felt like a test was wanted sooner than later. If we can wait, that might be the best approach. Otherwise I can modify this to have the `Linux Only` tag as I believe only Windows has the bug.

I don't think there is any rule written, just wanted to mean "please check with the windows folks before merging" :), 
By theway, seems they already have a fix for the bug https://github.com/kubernetes/kubernetes/pull/93638

test/e2e/framework/network/utils.go
```diff
@@ -618,6 +619,27 @@ func (config *NetworkingTestConfig) createSessionAffinityService(selector map[st
 	config.SessionAffinityService = config.createService(config.createNodePortServiceSpec(sessionAffinityServiceName, selector, true))
 }
 
+func (config *NetworkingTestConfig) CreateSecondNodePortService() (*v1.Service, int, int) {
```
fejta, line 622: This is only used by test/e2e/networking, and only one file in that package. Please declare the function in that package.

fejta (commented):
/assign aojea m1093782566 rramkumar
/ok-to-test

swetharepakula:
I have created a new test specifically for this scenario and moved the service creation function into the networking test package.

Thanks for the feedback @aojea & @fejta !

test/e2e/network/networking.go
```diff
@@ -226,6 +226,36 @@ var _ = SIGDescribe("Networking", func() {
 			config.DialFromEndpointContainer("udp", config.NodeIP, config.NodeUDPPort, config.MaxTries, 0, config.EndpointHostnames())
 		})
 
+		ginkgo.It("should function for multiple endpoint-Services with same selector", func() {
```
robscott, line 229: I think it could be important here to note the original motivation for this test - ensuring that connectivity is retained after 1/2 Services targeting a Pod is deleted.

test/e2e/network/networking.go
```diff
@@ -226,6 +226,36 @@ var _ = SIGDescribe("Networking", func() {
...
+			config := e2enetwork.NewNetworkingTestConfig(f, false, false)
+			ginkgo.By("creating a second service with same selector")
+			svc2, udpPort := CreateSecondNodePortService(f, config)
+
+			// original service should work
+			ginkgo.By(fmt.Sprintf("dialing(udp) %v (endpoint) --> %v:%v (config.clusterIP)", config.EndpointPods[0].Name, config.ClusterIP, e2enetwork.ClusterUDPPort))
```
robscott, line 235: Any reason to use udp instead of tcp here?

test/e2e/network/util.go
```diff
@@ -162,3 +166,36 @@ func execSourceIPTest(sourcePod v1.Pod, targetAddr string) (string, string) {
 	}
 	return sourcePod.Status.PodIP, host
 }
+
+// CreateSecondNodePortService creates a service with the same selector as config.NodePortService and same UDP Port
+func CreateSecondNodePortService(f *framework.Framework, config *e2enetwork.NetworkingTestConfig) (*v1.Service, int) {
```
robscott, line 171: Does this function need to be exported?

test/e2e/network/util.go
```diff
@@ -26,13 +26,17 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2enetwork "k8s.io/kubernetes/test/e2e/framework/network"
 	"k8s.io/kubernetes/test/e2e/storage/utils"
 	imageutils "k8s.io/kubernetes/test/utils/image"
 )
 
+//secondNodePortSvcName is the name of the secondary node port service
```
robscott, line 37: Nit:
```suggestion
// secondNodePortSvcName is the name of the secondary node port service
```

test/e2e/network/util.go
```diff
@@ -162,3 +166,36 @@ func execSourceIPTest(sourcePod v1.Pod, targetAddr string) (string, string) {
...
+// CreateSecondNodePortService creates a service with the same selector as config.NodePortService and same UDP Port
+func CreateSecondNodePortService(f *framework.Framework, config *e2enetwork.NetworkingTestConfig) (*v1.Service, int) {
+	svc := &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: secondNodePortSvcName,
+		},
+		Spec: v1.ServiceSpec{
+			Type: v1.ServiceTypeNodePort,
+			Ports: []v1.ServicePort{
+				{Port: e2enetwork.ClusterUDPPort, Name: "udp", Protocol: v1.ProtocolUDP, TargetPort: intstr.FromInt(e2enetwork.EndpointUDPPort)},
```
robscott, line 179: Nit: This is a long line, might be worth splitting it up.

test/e2e/framework/network/utils.go
```diff
@@ -65,6 +65,7 @@ const (
 	testPodName                = "test-container-pod"
 	hostTestPodName            = "host-test-container-pod"
 	nodePortServiceName        = "node-port-service"
+	secondNodePortServiceName  = "second-node-port-service"
```
robscott, line 68: I don't think this is needed anymore.

robscott (commented):
Thanks for your work on this!

fejta:
Generally approve, delegating to networking as subject-matter experts.

test/e2e/network/networking.go
swetharepakula: That is a good point! I will add a clarifying comment.

test/e2e/network/networking.go
swetharepakula: No reason. I think I started with the UDP test which was right above. I will switch it to tcp.

test/e2e/network/networking.go
```diff
@@ -226,6 +226,38 @@ var _ = SIGDescribe("Networking", func() {
...
+			config.DialFromEndpointContainer("http", config.ClusterIP, e2enetwork.ClusterHTTPPort, config.MaxTries, 0, config.EndpointHostnames())
+
+			ginkgo.By(fmt.Sprintf("dialing(http) %v (endpoint) --> %v:%v (nodeIP)", config.EndpointPods[0].Name, config.NodeIP, config.NodeHTTPPort))
+			config.DialFromEndpointContainer("http", config.NodeIP, config.NodeHTTPPort, config.MaxTries, 0, config.EndpointHostnames())
+
+			// Dial second service
+			ginkgo.By(fmt.Sprintf("dialing(http) %v (endpoint) --> %v:%v (svc2.clusterIP)", config.EndpointPods[0].Name, svc2.Spec.ClusterIP, e2enetwork.ClusterHTTPPort))
+			config.DialFromEndpointContainer("http", svc2.Spec.ClusterIP, e2enetwork.ClusterHTTPPort, config.MaxTries, 0, config.EndpointHostnames())
+
+			ginkgo.By(fmt.Sprintf("dialing(http) %v (endpoint) --> %v:%v (nodeIP)", config.EndpointPods[0].Name, config.NodeIP, httpPort))
```
aojea, line 247: nit, instead of nodeIP it can be svc2.NodeIP or something that helps to differentiate from the first service.
This will help debugging in case of failures

test/e2e/network/networking.go
```diff
@@ -226,6 +226,38 @@ var _ = SIGDescribe("Networking", func() {
...
+			config.DialFromEndpointContainer("http", config.NodeIP, httpPort, config.MaxTries, 0, config.EndpointHostnames())
+
+			ginkgo.By("deleting the original node port service")
+			config.DeleteNodePortService()
+
+			// Second service should continue to function unaffected
+			ginkgo.By(fmt.Sprintf("dialing(http) %v (endpoint) --> %v:%v (svc2.clusterIP)", config.EndpointPods[0].Name, svc2.Spec.ClusterIP, e2enetwork.ClusterHTTPPort))
+			config.DialFromEndpointContainer("http", svc2.Spec.ClusterIP, e2enetwork.ClusterHTTPPort, config.MaxTries, 0, config.EndpointHostnames())
+
+			ginkgo.By(fmt.Sprintf("dialing(http) %v (endpoint) --> %v:%v (nodeIP)", config.EndpointPods[0].Name, config.NodeIP, httpPort))
```
aojea, line 257: same comment as above, something to distinguish in the comment this is the NodePort of the second svc

aojea:
@swetharepakula some nits for improving the test output, feel free to disregard them
this is really nice, thanks
/lgtm

PR #94244, opened 2020-08-26
Resource policy rules: matching names

yue9944882:


MikeSpreitzer:
@MikeSpreitzer

pkg/apis/flowcontrol/types.go
```diff
@@ -265,6 +266,15 @@ type ResourcePolicyRule struct {
 	// +optional
 	// +listType=set
 	Namespaces []string
+
+	// `resourceNames` is a list of target resource names that restricts
+	// matches. A (get, update, patch, delete) request that specifies
+	// a target name matches if either (a) this list contains that target
+	// names or (b) this list contains "*".
+	// This list may be empty.
```
lavalamp, line 274: I would not make empty permitted. I would make it default to `["*"]` instead.

lavalamp: It looks like the rest of the PR does it the way I suggest, so it's actually just the comment that should be changed :)

lavalamp:
I think we had a bit of a debate about whether this would be used or not, which I don't think we came to an agreement on. @deads2k @MikeSpreitzer continue debate here?

(This change looks correct other than the comment, if we want to do this. Thanks, Min!)

MikeSpreitzer:
The pull-kubernetes-bazel-test job is failing because round-trip tests are failing.  The error messages show differences due to defaulting and canonicalization.  Those things are not really round-trip violations, right?  Is there something wrong with the original objects being chosen?  What is specifying them?

pkg/apis/flowcontrol/fuzzer/fuzzer.go
```diff
@@ -0,0 +1,67 @@
...
+
+package fuzzer
+
+import (
+	fuzz "github.com/google/gofuzz"
+	runtimeserializer "k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/kubernetes/pkg/apis/flowcontrol"
+)
+
+// Funcs returns the fuzzer functions for the authentication api group.
```
MikeSpreitzer, line 25: This comment needs to be adjusted.

pkg/apis/flowcontrol/fuzzer/fuzzer.go
```diff
@@ -0,0 +1,67 @@
...
+var Funcs = func(codecs runtimeserializer.CodecFactory) []interface{} {
+	return []interface{}{
+		func(obj *flowcontrol.ResourcePolicyRule, c fuzz.Continue) {
+			c.FuzzNoCustom(obj) // fuzz self without calling this function again
+			// fuzzing api groups
+			if c.RandBool() {
+				obj.APIGroups = []string{"*"}
+			} else {
+				obj.APIGroups = []string{c.RandString()}
```
MikeSpreitzer, line 34: Should we test a list longer than 1?

pkg/apis/flowcontrol/validation/validation.go
```diff
@@ -303,6 +303,20 @@ func ValidateFlowSchemaResourcePolicyRule(rule *flowcontrol.ResourcePolicyRule,
 		}
 	}
 
+	if len(rule.ResourceNames) == 0 {
+		allErrs = append(allErrs, field.Required(fldPath.Child("resourceNames"), `resource rules must supply at least one resource names`))
```
MikeSpreitzer, line 307: the English text should be "at least one resource name".

pkg/apis/flowcontrol/validation/validation_test.go
```diff
@@ -860,6 +876,152 @@ func TestFlowSchemaValidation(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			expectedErrors: field.ErrorList{
+				field.Invalid(field.NewPath("spec").Child("rules").Index(0).Child("resourceRules").Index(0).Child("resourceNames").Index(0), "", `each member of this list must not be an empty string`),
+				field.Invalid(field.NewPath("spec").Child("rules").Index(0).Child("resourceRules").Index(0).Child("resourceNames").Index(1), "", `each member of this list must not be an empty string`),
+			},
+		},
+		{
```
MikeSpreitzer, line 988: This looks like a copy of the previous test.

staging/src/k8s.io/apiserver/pkg/util/flowcontrol/gen_test.go
```diff
@@ -562,6 +562,7 @@ func genResourceRule(rng *rand.Rand, pfx string, mayMatchClusterScope, matchAllR
 		rr.ClusterScope = true
 		rr.Namespaces = []string{fcv1a1.NamespaceEvery}
 	}
+	rr.ResourceNames = []string{fcv1a1.ResourceNameEvery}
```
MikeSpreitzer, line 565: TODO, OK if it is in a follow-up PR: exercise the other possibilities here.

pkg/apis/flowcontrol/validation/validation_test.go
MikeSpreitzer: Seems to have been resolved

staging/src/k8s.io/apiserver/pkg/util/flowcontrol/match_test.go
```diff
@@ -120,162 +120,236 @@ func TestLiterals(t *testing.T) {
...
+			Namespaces:    []string{"*"},
+			ResourceNames: []string{"eman"},
+		}}}, {
+		Subjects: []fcv1a1.Subject{{Kind: fcv1a1.SubjectKindUser,
+			User: &fcv1a1.UserSubject{"goodu"}}},
+		ResourceRules: []fcv1a1.ResourcePolicyRule{{
+			Verbs:         []string{"goodverb"},
+			APIGroups:     []string{"goodapig"},
+			Resources:     []string{"goodrscs"},
+			Namespaces:    []string{"*"},
```
MikeSpreitzer, line 198: I would make this `"goodns"` instead of `"*"` because this series of test cases is testing individual wildcards.

staging/src/k8s.io/apiserver/pkg/util/flowcontrol/match_test.go
```diff
@@ -120,162 +120,236 @@ func TestLiterals(t *testing.T) {
...
 		ResourceRules: []fcv1a1.ResourcePolicyRule{{
-			Verbs:      []string{"goodverb"},
-			APIGroups:  []string{"goodapig"},
-			Resources:  []string{"goodrscs"},
-			Namespaces: []string{"goodns"}}}}, {
+			Verbs:         []string{"goodverb"},
+			APIGroups:     []string{"goodapig"},
+			Resources:     []string{"goodrscs"},
+			Namespaces:    []string{"goodns"},
+			ResourceNames: []string{"eman"},
```
MikeSpreitzer, line 127: Given that we are now distinguishing between matching and mismatching values for ResourceName, I would not object if the value here were `"goodn"` rather than `"eman"` (with corresponding changes in the test cases).

yue9944882: fixed

staging/src/k8s.io/apiserver/pkg/util/flowcontrol/match_test.go
yue9944882: yes, updated

pkg/apis/flowcontrol/fuzzer/fuzzer.go
yue9944882: fixed

pkg/apis/flowcontrol/fuzzer/fuzzer.go
yue9944882: fixed

pkg/apis/flowcontrol/validation/validation.go
yue9944882: updated

pkg/apis/flowcontrol/validation/validation_test.go
yue9944882: done

staging/src/k8s.io/apiserver/pkg/util/flowcontrol/gen_test.go
yue9944882: fixed

yue9944882:
@MikeSpreitzer i suppose all review comments is addressed now. plz let me know if i miss anything, thanks!

pkg/apis/flowcontrol/fuzzer/fuzzer.go
```diff
@@ -0,0 +1,71 @@
...
+// Funcs returns the fuzzer functions for the flowcontrol api group.
+var Funcs = func(codecs runtimeserializer.CodecFactory) []interface{} {
+	return []interface{}{
+		func(obj *flowcontrol.ResourcePolicyRule, c fuzz.Continue) {
+			c.FuzzNoCustom(obj) // fuzz self without calling this function again
+			// fuzzing api groups
+			if c.RandBool() {
+				obj.APIGroups = []string{"*"}
+			} else {
+				obj.APIGroups = []string{c.RandString()}
```
MikeSpreitzer, line 34: I still see only lists of length 1 here.

staging/src/k8s.io/apiserver/pkg/util/flowcontrol/match_test.go
```diff
@@ -120,162 +120,236 @@ func TestLiterals(t *testing.T) {
...
+		}}},
+		{
+			Subjects: []fcv1a1.Subject{{Kind: fcv1a1.SubjectKindUser,
+				User: &fcv1a1.UserSubject{"goodu"}}},
+			ResourceRules: []fcv1a1.ResourcePolicyRule{{
+				Verbs:         []string{"goodverb"},
+				APIGroups:     []string{"goodapig"},
+				Resources:     []string{"goodrscs"},
+				ClusterScope:  false,
+				ResourceNames: []string{"goodn"},
```
MikeSpreitzer, line 351: (a) the indentation in this test case is different than in the others.

(b) this test case has the same values as the preceding one.

(c) we are missing a test case where there is a mismatch on the resource name.

pkg/apis/flowcontrol/fuzzer/fuzzer.go
yue9944882: ah oversight, PTAL the latest diff

staging/src/k8s.io/apiserver/pkg/util/flowcontrol/match_test.go
yue9944882: updated

MikeSpreitzer (approved):
/LGTM

pkg/apis/flowcontrol/types.go
```diff
@@ -265,6 +266,15 @@ type ResourcePolicyRule struct {
 	// +optional
 	// +listType=set
 	Namespaces []string
+
+	// `resourceNames` is a list of target resource names that restricts
+	// matches. A (get, update, patch, delete) request that specifies
+	// a target name matches if either (a) this list contains that target
+	// names or (b) this list contains "*".
```
MikeSpreitzer, line 273: Look above at the comment on `ResourcePolicyRule`.  That needs to be extended, in the same style that it already has, and this field's comment needs to play its part.

staging/src/k8s.io/apiserver/pkg/apis/flowcontrol/bootstrap/default.go
```diff
@@ -460,13 +482,14 @@ func kubeSystemServiceAccount(names ...string) []flowcontrol.Subject {
 	return subjects
 }
 
-func resourceRule(verbs []string, groups []string, resources []string, namespaces []string, clusterScoped bool) flowcontrol.ResourcePolicyRule {
+func resourceRule(verbs []string, groups []string, resources []string, namespaces []string, clusterScoped bool, resourceNames []string) flowcontrol.ResourcePolicyRule {
```
MikeSpreitzer, line 485: It looks to me like we do not need to add a parameter here, every call site passes the same value through this parameter.

staging/src/k8s.io/apiserver/pkg/util/flowcontrol/gen_test.go
```diff
@@ -570,23 +574,29 @@ func genResourceRule(rng *rand.Rand, pfx string, mayMatchClusterScope, matchAllR
...
 	ans := make([]*request.RequestInfo, 0, m)
 	for _, coord := range coords {
 		ans = append(ans, &request.RequestInfo{
 			IsResourceRequest: true,
 			Verb:              verbs[coord%nv],
 			APIGroup:          apiGroups[coord/nv%ng],
 			Resource:          resources[coord/nv/ng%nr],
-			Namespace:         namespaces[coord/nv/ng/nr]})
+			Namespace:         namespaces[coord/nv/ng/nr%nns],
+			Name:              names[coord/nv/ng/nr/nns%nn],
```
MikeSpreitzer, line 598: The remainder operation is not needed here, just as it was not needed at the end of the original series.

PR #94246, opened 2020-08-26
Fix issue on skipTest in storage suits

### LINKED ISSUE #94202: [e2e] DynamicSnapshot option should not be in default TestPattern
**What happened**:

In 1.19 we added a "dynamic snapshot" option to the default TestPattern.
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testpatterns/testpattern.go#L113

But it is causing tests to skip for the drivers that don't support snapshots. Even though the test is not actually testing snapshot functionality. Like 
`External Storage [Driver: standard-rwo] [Testpattern: Dynamic PV (block volmode)] stress multiple pods should access different volumes repeatedly [Slow] [Serial] [BeforeEach]`

So it seems we should not add snapshots to default options. And instead we should have an explicit snapshots testpattern for snapshot tests.

```
ginkgo -p -focus='External.*Storage.*multiple.pods.should.access.different.volumes.*' \
       ./_output/bin/e2e.test \
       -- \
       -storage.testdriver=/path/to/config/test-driver.yaml
```

/sig storage
/kind bug
/cc @msau42

jingxu97:
After PR https://github.com/kubernetes/kubernetes/pull/92555, there are a number of gce pd default fs tests skipped. Here the testpatten has SnapshotType set because some provisioning tests use snapshots. But for drivers such as In-tree gce pd driver, the tests will be skipped because of the logic in skipUnsupportedTesthttps://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/base.go#L154
Since multiple drivers might test with the same pattern, so I think we need keep SnapshotType here.

This PR removes the part of the logic in skipUnsupportedTest. This should be ok because all snapshot tests will check whether a driver has snapshot capability or not.

**What type of PR is this?**

Fixes https://github.com/kubernetes/kubernetes/issues/94202

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/storage/testsuites/base.go
+++ b/test/e2e/storage/testsuites/base.go
@@ -149,50 +149,36 @@ func skipUnsupportedTest(driver TestDriver, pattern testpatterns.TestPattern) {
 	dInfo := driver.GetDriverInfo()
 	var isSupported bool
 
-	// 1. Check if Whether SnapshotType is supported by driver from its interface
-	// if isSupported, we still execute the driver and suite tests
-	if len(pattern.SnapshotType) > 0 {
-		switch pattern.SnapshotType {
-		case testpatterns.DynamicCreatedSnapshot:
-			_, isSupported = driver.(SnapshottableTestDriver)
-		default:
-			isSupported = false
-		}
-		if !isSupported {
-			e2eskipper.Skipf("Driver %s doesn't support snapshot type %v -- skipping", dInfo.Name, pattern.SnapshotType)
-		}
-	} else {
-		// 2. Check if Whether volType is supported by driver from its interface
-		switch pattern.VolType {
-		case testpatterns.InlineVolume:
-			_, isSupported = driver.(InlineVolumeTestDriver)
-		case testpatterns.PreprovisionedPV:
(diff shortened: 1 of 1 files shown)


msau42:
/lgtm
/approve
/kind failing-test

We should also cherry-pick this to 1.19 so that we get back our test coverage.

test/e2e/storage/testsuites/base.go
```diff
@@ -149,50 +149,36 @@ func skipUnsupportedTest(driver TestDriver, pattern testpatterns.TestPattern) {
...
-		case testpatterns.PreprovisionedPV:
-			_, isSupported = driver.(PreprovisionedPVTestDriver)
-		case testpatterns.DynamicPV, testpatterns.GenericEphemeralVolume:
-			_, isSupported = driver.(DynamicPVTestDriver)
-		case testpatterns.CSIInlineVolume:
-			_, isSupported = driver.(EphemeralTestDriver)
-		default:
-			isSupported = false
-		}
+	// 1. Check if Whether volType is supported by driver from its interface
```
Jiawei0227, line 152: The comments for skipUnsupportedTest may also need to be updated.

Issue #93749, opened 2020-08-06
[Flaky Test] [sig-storage] CSI mock volume CSI attach test using mock driver should not require VolumeAttach for drivers without attachment 


hasheddan:
**Which jobs are flaking**:

`gce-ubuntu-master-containerd (ci-kubernetes-e2e-ubuntu-gce-containerd)`

**Which test(s) are flaking**:

`[sig-storage] CSI mock volume CSI attach test using mock driver should not require VolumeAttach for drivers without attachment `

**Testgrid link**:

https://testgrid.k8s.io/sig-release-master-blocking#gce-ubuntu-master-containerd

**Reason for failure**:

This appears to be an issue with the test cleaning up a namespace too early.

```
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/storage/csi_mock_volume.go:267
Aug  5 12:10:36.725: deploying csi mock driver: create StatefulSet: statefulsets.apps "csi-mockplugin" is forbidden: unable to create new content in namespace csi-mock-volumes-7035 because it is being terminated
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/storage/drivers/csi.go:410
```

**Anything else we need to know**:

/cc @kubernetes/ci-signal @kubernetes/sig-storage-test-failures 
/sig storage
/milestone v1.19
/priority important-soon

msau42:
I wonder if the generated namespace suffix is not unique enough.

cc @Jiawei0227 to take a look

Jiawei0227:
We do check if the namespace already exists here:
https://github.com/kubernetes/kubernetes/blob/a6405451674428172956116b4c93710ac03ebddd/test/e2e/framework/util.go#L346

msau42:
For CSI tests, we generate a 2nd namespace for the driver. Does that use the same method? cc @gnufied

Jiawei0227:
Another look at the raw-log. It seems like we do deleted the namespace in another test...
This namespace was in the testcase namespace in that test. And it is also the csi-driver namespace for this test. And after the first test terminated, that namespace get deleted. That's why the second test failed.

Jiawei0227:
Okay, I found the issue.
This happens because of the cleanup mechanism in CSI mock volume.
[...]

PR #94247, opened 2020-08-26
test(e2e::storage): avoid deleting primary and driver namespace twice

### LINKED ISSUE #94221: [Flaky Test] [sig-storage] In-tree Volumes [Driver: gluster]  subPath should support file as subpath [LinuxOnly]
**Which jobs are flaking**:
ci-kubernetes-e2e-ubuntu-gce-containerd

[sig-storage] In-tree Volumes [Driver: gluster] [Testpattern: Pre-provisioned PV (default fs)] subPath should support file as subpath [LinuxOnly] 

**Testgrid link**:

**Reason for failure**:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-ubuntu-gce-containerd/1298011624451346432

```
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/storage/testsuites/subpath.go:227
Aug 24 21:52:33.298: failed to create service for Gluster server
Unexpected error:
    <*errors.StatusError | 0xc001718820>: {
        ErrStatus: {
            ListMeta: {
                SelfLink: "",
                ResourceVersion: "",
                Continue: "",
                RemainingItemCount: nil,
            },
            Status: "Failure",
            Message: "services \"gluster-server\" is forbidden: unable to create new content in namespace provisioning-9612 because it is being terminated",
            Reason: "Forbidden",
            Details: {
                Name: "gluster-server",
                Group: "",
                Kind: "services",
                UID: "",
                Causes: [
                    {
                        Type: "NamespaceTerminating",
                        Message: "namespace provisioning-9612 is being terminated",
                        Field: "metadata.namespace",
                    },
                ],
                RetryAfterSeconds: 0,
            },
            Code: 403,
        },
    }
    services "gluster-server" is forbidden: unable to create new content in namespace provisioning-9612 because it is being terminated
occurred
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/framework/volume/fixtures.go:191
```

Will look at it shortly, why it is flaking.

knight42:

**What type of PR is this?**

/kind flake

As described in https://github.com/kubernetes/kubernetes/issues/94221#issuecomment-680147677, the root cause is the namespace would be deleted twice in the test, it is more obvious if we take a closer look at the access log of apiserver:
```
# created namespace "provisioning-9612" in test "subPath should support existing single file"
45875:I0824 21:48:27.739634       9 httplog.go:89] "HTTP" verb="POST" URI="/api/v1/namespaces" latency="10.521185ms" userAgent="e2e.test/v1.20.0 (linux/amd64) kubernetes/3229073 -- [sig-storage] In-tree Volumes [Driver: local][LocalVolumeType: dir] [Testpattern: Pre-provisioned PV (default fs)] subPath should support existing single file [LinuxOnly]" srcIP="34.121.82.231:58558" resp=201

# deleted namespace "provisioning-9612" in test "subPath should support existing single file"
122918:I0824 21:52:03.923611       9 httplog.go:89] "HTTP" verb="DELETE" URI="/api/v1/namespaces/provisioning-9612" latency="5.37683ms" userAgent="e2e.test/v1.20.0 (linux/amd64) kubernetes/3229073 -- [sig-storage] CSI Volumes [Driver: csi-hostpath] [Testpattern: Dynamic PV (default fs)] subPath should support existing single file [LinuxOnly]" srcIP="34.121.82.231:58578" resp=200

# created namespace "provisioning-9612" in test "subPath should support file as subpath"
127903:I0824 21:52:18.766531       9 httplog.go:89] "HTTP" verb="POST" URI="/api/v1/namespaces" latency="3.974101ms" userAgent="e2e.test/v1.20.0 (linux/amd64) kubernetes/3229073 -- [sig-storage] In-tree Volumes [Driver: gluster] [Testpattern: Pre-provisioned PV (default fs)] subPath should support file as subpath [LinuxOnly]" srcIP="34.121.82.231:33176" resp=201

# deleted namespace "provisioning-9612" in test "subPath should support existing single file"
130263:I0824 21:52:28.584696       9 httplog.go:89] "HTTP" verb="DELETE" URI="/api/v1/namespaces/provisioning-9612" latency="13.93603ms" userAgent="e2e.test/v1.20.0 (linux/amd64) kubernetes/3229073 -- [sig-storage] CSI Volumes [Driver: csi-hostpath] [Testpattern: Dynamic PV (default fs)] subPath should support existing single file [LinuxOnly]" srcIP="34.121.82.231:58578" resp=200

# deleted namespace "provisioning-9612" in test "subPath should support file as subpath"
132603:I0824 21:52:37.288535       9 httplog.go:89] "HTTP" verb="DELETE" URI="/api/v1/namespaces/provisioning-9612" latency="2.292099ms" userAgent="e2e.test/v1.20.0 (linux/amd64) kubernetes/3229073 -- [sig-storage] In-tree Volumes [Driver: gluster] [Testpattern: Pre-provisioned PV (default fs)] subPath should support file as subpath [LinuxOnly]" srcIP="34.121.82.231:33176" resp=200
```

Perhaps I am missing something, but I think we don't need to delete the namespace during the test, as the namespace would be deleted after the test: https://github.com/kubernetes/kubernetes/blob/4db3a096ce8ac730b2280494422e1c4cf5fe875e/test/e2e/framework/framework.go#L375-L395

Fixes #94221

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/framework/util.go
+++ b/test/e2e/framework/util.go
@@ -341,7 +341,7 @@ func CreateTestingNS(baseName string, c clientset.Interface, labels map[string]s
 	}
 	// Be robust about making the namespace creation call.
 	var got *v1.Namespace
-	if err := wait.PollImmediate(Poll, 30*time.Second, func() (bool, error) {
+	if err := wait.PollImmediate(Poll, wait.ForeverTestTimeout, func() (bool, error) {
 		var err error
 		got, err = c.CoreV1().Namespaces().Create(context.TODO(), namespaceObj, metav1.CreateOptions{})
 		if err != nil {
--- a/test/e2e/storage/drivers/csi.go
+++ b/test/e2e/storage/drivers/csi.go
@@ -173,7 +173,6 @@ func (h *hostpathCSIDriver) PrepareTest(f *framework.Framework) (*testsuites.Per
 	// Create secondary namespace which will be used for creating driver
 	driverNamespace := utils.CreateDriverNamespace(f)
 	ns2 := driverNamespace.Name
-	ns1 := f.Namespace.Name
 
 	ginkgo.By(fmt.Sprintf("deploying %s driver", h.driverInfo.Name))
 	cancelLogging := testsuites.StartPodLogs(f, driverNamespace)
@@ -210,11 +209,6 @@ func (h *hostpathCSIDriver) PrepareTest(f *framework.Framework) (*testsuites.Per
 	// Cleanup CSI driver and namespaces. This function needs to be idempotent and can be
 	// concurrently called from defer (or AfterEach) and AfterSuite action hooks.
 	cleanupFunc := func() {
-		ginkgo.By(fmt.Sprintf("deleting the test namespace: %s", ns1))
-		// Delete the primary namespace but its okay to fail here because this namespace will
-		// also be deleted by framework.Aftereach hook
-		tryFunc(deleteNamespaceFunc(f.ClientSet, ns1, framework.DefaultNamespaceDeletionTimeout))
-
 		ginkgo.By("uninstalling csi mock driver")
 		tryFunc(cleanup)
 		tryFunc(cancelLogging)
@@ -332,7 +326,6 @@ func (m *mockCSIDriver) PrepareTest(f *framework.Framework) (*testsuites.PerTest
 	// Create secondary namespace which will be used for creating driver
 	driverNamespace := utils.CreateDriverNamespace(f)
 	ns2 := driverNamespace.Name
-	ns1 := f.Namespace.Name
 
 	ginkgo.By("deploying csi mock driver")
 	cancelLogging := testsuites.StartPodLogs(f, driverNamespace)
@@ -413,11 +406,6 @@ func (m *mockCSIDriver) PrepareTest(f *framework.Framework) (*testsuites.PerTest
 	// Cleanup CSI driver and namespaces. This function needs to be idempotent and can be
 	// concurrently called from defer (or AfterEach) and AfterSuite action hooks.
 	cleanupFunc := func() {
-		ginkgo.By(fmt.Sprintf("deleting the test namespace: %s", ns1))
-		// Delete the primary namespace but its okay to fail here because this namespace will
-		// also be deleted by framework.Aftereach hook
-		tryFunc(deleteNamespaceFunc(f.ClientSet, ns1, framework.DefaultNamespaceDeletionTimeout))
-
 		ginkgo.By("uninstalling csi mock driver")
 		tryFunc(func() {
 			err := f.ClientSet.CoreV1().ConfigMaps(ns2).Delete(context.TODO(), hooksConfigMapName, metav1.DeleteOptions{})
@@ -537,7 +525,6 @@ func (g *gcePDCSIDriver) PrepareTest(f *framework.Framework) (*testsuites.PerTes
 	// Create secondary namespace which will be used for creating driver
 	driverNamespace := utils.CreateDriverNamespace(f)
 	ns2 := driverNamespace.Name
-	ns1 := f.Namespace.Name
 
 	cancelLogging := testsuites.StartPodLogs(f, driverNamespace)
 	// It would be safer to rename the gcePD driver, but that
@@ -574,11 +561,6 @@ func (g *gcePDCSIDriver) PrepareTest(f *framework.Framework) (*testsuites.PerTes
 	// Cleanup CSI driver and namespaces. This function needs to be idempotent and can be
 	// concurrently called from defer (or AfterEach) and AfterSuite action hooks.
 	cleanupFunc := func() {
-		ginkgo.By(fmt.Sprintf("deleting the test namespace: %s", ns1))
-		// Delete the primary namespace but its okay to fail here because this namespace will
-		// also be deleted by framework.Aftereach hook
-		tryFunc(deleteNamespaceFunc(f.ClientSet, ns1, framework.DefaultNamespaceDeletionTimeout))
-
 		ginkgo.By("uninstalling csi mock driver")
 		tryFunc(cleanup)
 		tryFunc(cancelLogging)
--- a/test/e2e/storage/drivers/csi.go
+++ b/test/e2e/storage/drivers/csi.go
@@ -172,7 +172,6 @@ func (h *hostpathCSIDriver) GetSnapshotClass(config *testsuites.PerTestConfig) *
 func (h *hostpathCSIDriver) PrepareTest(f *framework.Framework) (*testsuites.PerTestConfig, func()) {
 	// Create secondary namespace which will be used for creating driver
 	driverNamespace := utils.CreateDriverNamespace(f)
-	ns2 := driverNamespace.Name
 
 	ginkgo.By(fmt.Sprintf("deploying %s driver", h.driverInfo.Name))
 	cancelLogging := testsuites.StartPodLogs(f, driverNamespace)
@@ -213,8 +212,6 @@ func (h *hostpathCSIDriver) PrepareTest(f *framework.Framework) (*testsuites.Per
 		tryFunc(cleanup)
 		tryFunc(cancelLogging)
 
-		ginkgo.By(fmt.Sprintf("deleting the driver namespace: %s", ns2))
-		tryFunc(deleteNamespaceFunc(f.ClientSet, ns2, framework.DefaultNamespaceDeletionTimeout))
 		// cleanup function has already ran and hence we don't need to run it again.
 		// We do this as very last action because in-case defer(or AfterEach) races
 		// with AfterSuite and test routine gets killed then this block still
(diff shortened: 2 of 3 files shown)


test/e2e/framework/util.go
```diff
@@ -501,7 +501,9 @@ func LoadClientset() (*clientset.Clientset, error) {
 
 // RandomSuffix provides a random sequence to append to pods,services,rcs.
 func RandomSuffix() string {
-	return strconv.Itoa(rand.Intn(10000))
+	nonce := time.Now().Nanosecond()
+	// Reduce the chance of collision
+	return strconv.Itoa(nonce) + strconv.Itoa(rand.Intn(10000))
```
Jiawei0227, line 506: wondering if this might be too long? but it should be okay I guess...

test/e2e/storage/drivers/csi.go
```diff
@@ -210,11 +209,6 @@ func (h *hostpathCSIDriver) PrepareTest(f *framework.Framework) (*testsuites.Per
 	// Cleanup CSI driver and namespaces. This function needs to be idempotent and can be
 	// concurrently called from defer (or AfterEach) and AfterSuite action hooks.
 	cleanupFunc := func() {
-		ginkgo.By(fmt.Sprintf("deleting the test namespace: %s", ns1))
```
Jiawei0227, line 213: We have to delete the namespace here as per comments from @gnufied in https://github.com/kubernetes/kubernetes/issues/93749#issuecomment-670653716

knight42: > We have to delete the first namespace (which includes test pods), otherwise if you delete them after deleting the driver pods - we will have pods stuck forever in terminating state.

I am a bit curious why the pods would stuck forever?

test/e2e/framework/util.go
knight42: According to https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names, the max length of the name is 253, I think it is okay.

knight42: hmmm it is actually too long in other scenarios.. https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/94247/pull-kubernetes-e2e-kind-ipv6/1298521049540857857

knight42: reverted

test/e2e/storage/drivers/csi.go
knight42: Alternaltively could we cleanup the pods in the namespace rather than deleting the namespace?

knight42: Looks like the job `pull-kubernetes-e2e-gce-ubuntu-containerd` passed even if we do not delete the namespace twice: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/94247/pull-kubernetes-e2e-gce-ubuntu-containerd/1298528224002707456/

knight42:
```
I0826 08:19:21.475] [1mSTEP[0m: Destroying namespace "provisioning-2585" for this suite.
I0826 08:19:21.476] [1mSTEP[0m: Destroying namespace "provisioning-2585-3784" for this suite.
I0826 08:19:21.476] Aug 26 08:19:21.448: INFO: Namespace provisioning-2585-3784 was already deleted
```

@Jiawei0227 It looks like we don't need to delete the driver namespace during testing as well, because the driver namespace is created by `Framework.CreateNamespace()`, so the namespace would be deleted after test: https://github.com/kubernetes/kubernetes/blob/4db3a096ce8ac730b2280494422e1c4cf5fe875e/test/e2e/framework/framework.go#L482-L485

gnufied:
@knight42  we need to delete the test namespace and then driver namespace in order because:

1. If driver namespace is deleted before deleting the test namespace then pods in test namespace will be stuck in "terminating" state because CSI driver that is responsible for unmounting volumes will not be available.
2. We need to delete the test namespace (rather than say delete the pods which we do anyways), because deleting the test namespace first causes the test to start aborting. It works like a signal because it will cause executing test to exit and start running clean up routines.

knight42:
> @gnufied If driver namespace is deleted before deleting the test namespace then pods in test namespace will be stuck in "terminating" state because CSI driver that is responsible for unmounting volumes will not be available.

then I think we are good, as we create the primary namespace first then the driver namespace, they would be deleted in the same order: https://github.com/kubernetes/kubernetes/blob/4db3a096ce8ac730b2280494422e1c4cf5fe875e/test/e2e/framework/framework.go#L383-L386

gnufied:
> then I think we are good, as we create the primary namespace first then the driver namespace, they would be deleted in the same order: 

No the code you linked deletes the namespaces without waiting for them, so it is as good as deleting the driver namespace and test namespace almost together.

hasheddan (commented):
/priority important-soon

knight42:
> No the code you linked deletes the namespaces without waiting for them, so it is as good as deleting the driver namespace and test namespace almost together.

yeah but I think we could wait for the deletion and still avoid deleting the namespace twice.

gnufied:
I am putting this on hold so as concerns about csi driver cleanup are addressed properly. 

/hold

gnufied:
Another point is - there are other tests that delete the namespace in their own test suite. I think it is not realistic that nobody other than framework.AfterEach should delete the nameapce.  If we had namespace collision bug then, we should address that directly rather than doing what this PR proposes.

knight42:
> If we had namespace collision bug then, we should address that directly rather than doing what this PR proposes.

IMHO this is not a namespace collision bug, from the access log of apiserver I gave above:
```
# created namespace "provisioning-9612" in test "subPath should support existing single file"
45875:I0824 21:48:27.739634       9 httplog.go:89] "HTTP" verb="POST" URI="/api/v1/namespaces" latency="10.521185ms" userAgent="e2e.test/v1.20.0 (linux/amd64) kubernetes/3229073 -- [sig-storage] In-tree Volumes [Driver: local][LocalVolumeType: dir] [Testpattern: Pre-provisioned PV (default fs)] subPath should support existing single file [LinuxOnly]" srcIP="34.121.82.231:58558" resp=201

# deleted namespace "provisioning-9612" in test "subPath should support existing single file"
122918:I0824 21:52:03.923611       9 httplog.go:89] "HTTP" verb="DELETE" URI="/api/v1/namespaces/provisioning-9612" latency="5.37683ms" userAgent="e2e.test/v1.20.0 (linux/amd64) kubernetes/3229073 -- [sig-storage] CSI Volumes [Driver: csi-hostpath] [Testpattern: Dynamic PV (default fs)] subPath should support existing single file [LinuxOnly]" srcIP="34.121.82.231:58578" resp=200

# created namespace "provisioning-9612" in test "subPath should support file as subpath"
127903:I0824 21:52:18.766531       9 httplog.go:89] "HTTP" verb="POST" URI="/api/v1/namespaces" latency="3.974101ms" userAgent="e2e.test/v1.20.0 (linux/amd64) kubernetes/3229073 -- [sig-storage] In-tree Volumes [Driver: gluster] [Testpattern: Pre-provisioned PV (default fs)] subPath should support file as subpath [LinuxOnly]" srcIP="34.121.82.231:33176" resp=201

# deleted namespace "provisioning-9612" in test "subPath should support existing single file"
130263:I0824 21:52:28.584696       9 httplog.go:89] "HTTP" verb="DELETE" URI="/api/v1/namespaces/provisioning-9612" latency="13.93603ms" userAgent="e2e.test/v1.20.0 (linux/amd64) kubernetes/3229073 -- [sig-storage] CSI Volumes [Driver: csi-hostpath] [Testpattern: Dynamic PV (default fs)] subPath should support existing single file [LinuxOnly]" srcIP="34.121.82.231:58578" resp=200
```
It is literally valid for another test to use namespace `provisioning-9612` as this namespace does not exist before from its point of view, it is the second deletion that cause another test to fail.

> I think it is not realistic that nobody other than framework.AfterEach should delete the nameapce

I generally agree this point, but now that we create the namespace by calling `Framework.CreateNamespace()`, and `Framework.CreateNamespace()` would make this namespace to be deleted after the test, I think we don't need to bother deleteing it manually.

test/e2e/storage/drivers/csi.go
gnufied: I added https://github.com/kubernetes/kubernetes/pull/94247#issuecomment-680847436 why we can't skip deleting the namespace.

Another way of fixing this is to ensure that if primary namespace is deleted elsewhere then it is removed from `namespacesToDelete` array. But if we do that, one thing to keep in mind is - in case of failure etc, namespace information must still be dumped.

Jiawei0227:
I kind of agree with @knight42 idea of only deleting the namespace by `framework.AfterEach`. In my opinion, whoever creates the namespace should be responsible for cleanup the resources. So since framework added the namespace, we should let the framework do its work.  

My question for @gnufied is that is there any special reason we have to delete it twice? Because the root cause for this issue is the twice deletion of the namespace. Unless there is concern about the resource not getting cleanup correctly, I think it is reasonable to handle the namespace deletion to the framework.

knight42:

@gnufied Are you satisfied with this change https://github.com/kubernetes/kubernetes/pull/94247/commits/8779bd00bcc40d0be905402f28cd371fb742cb92?

gnufied:
@knight42 No it does not fixes that. The interaction we are trying to fix is kind of subtle. framework's AfterEach and AfterSuite can race with each other in certain cases(when interrupted for example) and in that case the defer/AfterEach within the test will start executing first and will remove the driver without waiting for the test pods to be deleted and that in turn would prevent deletion of test pods and hence causes resource leakage in case e2e timesout. 

@Jiawei0227 I have tried fixing the namespace being deleted twice bug with https://github.com/kubernetes/kubernetes/pull/94318 . The main issue to consider is how AfterEach and AfterSuite can race with each other. I have tried to document this https://github.com/kubernetes/kubernetes/pull/91689#issue-426705225 and in inline comments.

knight42:
close in favor of #94318

PR #69159, opened 2018-09-27
network/cni: more informative log messages

dcbw:
Log details about the pod name, namespace, sandbox ID, and result.
Remove some useless messages that are now redundant.

@freehan @thockin @squeed @bowei 

```release-note
NONE
```

PR #36020, opened 2016-11-01
Separate Direct and Indirect streaming paths, implement indirect path for CRI

timstclair:
This PR refactors the `pkg/kubelet/container.Runtime` interface to remove the `ExecInContainer`, `PortForward` and `AttachContainer` methods. Instead, those methods are part of the `DirectStreamingRuntime` interface which all "legacy" runtimes implement. I also added an `IndirectStreamingRuntime` which handles the redirect path and is implemented by CRI runtimes. To control the size of this PR, I did not fully setup the indirect streaming path for the dockershim, so I left legacy path behind.

Most of this PR is moving & renaming associated with the refactoring. To understand the functional changes, I suggest tracing the code from `getExec` in `pkg/kubelet/server/server.go`, which calls `GetExec` in `pkg/kubelet/kubelet_pods.go` to determine whether to follow the direct or indirect path.

For https://github.com/kubernetes/kubernetes/issues/29579

/cc @kubernetes/sig-node

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kubernetes/36020)

PR #94251, opened 2020-08-26
Use podDesc from kubelet/util/format

AlexeyPerevalov:
This commit just removes code duplication.

/kind cleanup


pkg/kubelet/dockershim/network/cni/cni.go
```diff
@@ -30,12 +30,14 @@ import (
 
 	"github.com/containernetworking/cni/libcni"
 	cnitypes "github.com/containernetworking/cni/pkg/types"
+	machinerytypes "k8s.io/apimachinery/pkg/types"
```
SergeyKanzhelev, line 33: this is a new alias for this import =). Consider using one of existing ones:

```
kubetypes "k8s.io/apimachinery/pkg/types"
k8stypes "k8s.io/apimachinery/pkg/types"
types "k8s.io/apimachinery/pkg/types"
commontypes "k8s.io/apimachinery/pkg/types"
```

also it is used so often so maybe worth placing into `hack/.import-aliases`

pkg/kubelet/dockershim/network/cni/cni.go
```diff
@@ -347,18 +349,14 @@ func (plugin *cniNetworkPlugin) TearDownPod(namespace string, name string, id ku
 	return plugin.deleteFromNetwork(cniTimeoutCtx, plugin.getDefaultNetwork(), name, namespace, id, netnsPath, nil)
 }
 
-func podDesc(namespace, name string, id kubecontainer.ContainerID) string {
-	return fmt.Sprintf("%s_%s/%s", namespace, name, id.ID)
```
SergeyKanzhelev, line 351: `format.PodDesc` is using `"%s_%s(%s)"` as a format. Here the `/` is used. It may be intentional.

SergeyKanzhelev: Comment to when this method was introduced (for history): https://github.com/kubernetes/kubernetes/pull/69159#discussion_r221402772

SergeyKanzhelev: and method existed long before this: https://github.com/kubernetes/kubernetes/pull/36020/files

SergeyKanzhelev:
I'm not sure what we gain out of this refactoring. Also it changes output (insignificantly). I'd suggest to hold this PR for better description on why it's needed.

/hold

AlexeyPerevalov:
> I'm not sure what we gain out of this refactoring. Also it changes output (insignificantly). I'd suggest to hold this PR for better description on why it's needed.
> 
> /hold

Are you afraid it could break log analyzers?

bboreham:
@SergeyKanzhelev has a valid point.  I don't think it's worth the type-cast to save 3 lines.

However I am now uneasy that people could be confused by the similar-looking rendering of pod UID or container ID.

SergeyKanzhelev:
> > I'm not sure what we gain out of this refactoring. Also it changes output (insignificantly). I'd suggest to hold this PR for better description on why it's needed.
> > /hold
> 
> Are you afraid it could break log analyzers?

This too. But mostly being cautious. With the different format there is at least some difference between pod ID and container ID. Maybe @kubernetes/sig-instrumentation-pr-reviews may suggest something or give some historical perspective.

SergeyKanzhelev:
/close

please re-open if there is a desire to pursue it further

PR #94258, opened 2020-08-26
update drain message to include namespace for pod 

deads2k:
in kubectl get compatible format.

when draining fails and pods are listed, you need to know the namespace to find the pod. Making it copy/pastable into `kubectl get pods/foo -n bar` is convenient.

/kind bug
Bug because the information isn't usable enough to fix a cluster
/priority important-soon
@kubernetes/sig-cli-maintainers 

```release-note
NONE
```


hasheddan (approved):
/lgtm

cgwalters:
Thanks, this would have helped us debug a drain issue.  Looks like it's just missing a milestone label?

JAORMX (approved):

Issue #94141, opened 2020-08-20
Update etcd version to 3.4.13


wenjiaswe:
Bump etcd version to 3.4.13 to pick up a few important fixes:

* etcd [changelog](https://github.com/etcd-io/etcd/blob/master/CHANGELOG-3.4.md)
* Fix [CVE-2020-15106](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-15106) and [CVE-2020-15112](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-15112). 

/sig api-machinery
/kind feature
/assign @jingyih 
/cc @liggitt @gyuho @xiang90 @wojtek-t @ptabor

wojtek-t:
Actually - we may consider waiting for the next release that will contain: https://github.com/etcd-io/etcd/pull/12242

[That will allow us to avoid adjusting the permissions issue we would hit.]

neolit123:
looks like https://github.com/etcd-io/etcd/pull/12242 merged.

wenjiaswe:
https://github.com/etcd-io/etcd/pull/12250 is the one that back ports the fix to release 3.4, once that's merged and a new 3.4.13 is released, we could use that for k8s. cc @gyuho

gyuho:
@wenjiaswe https://github.com/etcd-io/etcd/releases/tag/v3.4.13 has been released.

saschagrunert:
How do we want to handle this? Kubernetes v1.19.0 will still contain etc client/server v3.4.9, but I think we should update ASAP, right?

BenTheElder:
3.4.12 client* already https://github.com/kubernetes/kubernetes/pull/94105/

PR #94259, opened 2020-08-26
Update etcd client to v3.4.13

### LINKED ISSUE #12144: Upgrading from v3.4.9 to v3.4.10 requires manual action
After upgrading existing cluster from version `v3.4.9` to `v.3.4.10`, etcd fails to start with following error:
```
{"level":"fatal","ts":"2020-07-17T07:11:27.596Z","caller":"etcdmain/etcd.go:271","msg":"discovery failed","error":"cannot access data directory: directory \"/theory-controllers-1.etcd\",\"drwxr-xr-x\" exist without desired file permission \
"-rwx------\".","stacktrace":"go.etcd.io/etcd/etcdmain.startEtcdOrProxyV2\n\t/tmp/etcd-release-3.4.10/etcd/release/etcd/etcdmain/etcd.go:271\ngo.etcd.io/etcd/etcdmain.Main\n\t/tmp/etcd-release-3.4.10/etcd/release/etcd/etcdmain/main.go:46\nm
ain.main\n\t/tmp/etcd-release-3.4.10/etcd/release/etcd/main.go:28\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:200"}
```

To get it to work, manual `sudo chmod 0700` execution is required on all members user data directories to get etcd running again. This seems to be a new requirement (or old one, which was not enforced) and it is not mentioned in the changelog nor in the upgrade guide, which was unexpected.

Perhaps this should be documented somewhere?

jingyih:
**What type of PR is this?**
/kind bug

We are bumping etcd client version to v3.4.13 in this PR. Then we can bump etcd server (used in testing and manifests) to the same version to pick up https://github.com/etcd-io/etcd/pull/12242 which fixes https://github.com/etcd-io/etcd/issues/12144 and https://github.com/etcd-io/etcd/issues/12148.

Related:
https://github.com/kubernetes/kubernetes/issues/94141
https://github.com/kubernetes/kubernetes/pull/94102
https://github.com/kubernetes/kubeadm/issues/2256

```release-note
Update etcd client side to v3.4.13
```

/sig api-machinery
/area etcd

cc @kubernetes/release-engineering @neolit123 @justaugustus @saschagrunert @BenTheElder @wojtek-t @wenjiaswe


neolit123:
LGTM, thanks @jingyih

jingyih:
Can we still get this in Kubernetes v1.19?

neolit123:
> Can we still get this in Kubernetes v1.19?

i don't think we can because 1.19 is releasing later today and usually new etcd needs to be tested in CI for a while.

hasheddan:
@jingyih @neolit123 v1.19 is now live

jingyih:
Got it. Thanks:)

/milestone v1.20

saschagrunert (approved):
/lgtm

wojtek-t:
Milestone will be applied by release team later according to staged merge proposal.

/approve

liggitt:
does this actually matter on the client side? the permission check being fixed is run by the server, right?

jingyih:
> does this actually matter on the client side? the permission check being fixed is run by the server, right?

You are correct. The fix is on server side. But I think it is a good idea to keep the client and server versions match.

liggitt:
sure, just wanting to make sure this is characterized correctly

jingyih:
> sure, just wanting to make sure this is characterized correctly

Oh I see. Thanks! Let me correct the PR description.

PR #94267, opened 2020-08-27
Override hostname with instance name

lizhuqi:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**
To join Windows nodes to AD, we need to shorten their names, as NetBIOS name must be <= 15 characters, and generated names can be longer than that. 

To perform the join in an automated way, it's preferable to apply the rename and domain join when an instance is being prepared for the image creation process. However, preparation is complete and the machine restarts, kubelet bootstrapping should not use the shortened computer name, and instead use the instance's name by using --hostname-override, otherwise kubelet and kube-proxy will not be able to run properly, as explained in doc below:
https://cloud.google.com/kubernetes-engine/docs/how-to/creating-a-cluster-windows#using_gmsa

Today kubelet and kube-proxy parameters are defined in the kube-env instance metadata field. Adding the --hostname-override parameter with the name of the instance to the metadata field (in the KUBELET_ARGS and KUBEPROXY_ARGS) will enable the bootstrapping to use the correct hostname.
I have ran the e2e test below to verify the script update:
https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/windows/README-GCE-Windows-kube-up.md**What this PR does / why we need it**: 

```release-note
None
```

### DIFF
--- a/cluster/gce/windows/k8s-node-setup.psm1
+++ b/cluster/gce/windows/k8s-node-setup.psm1
@@ -1143,9 +1143,20 @@ function Start-WorkerServices {
   $kubelet_args_str = ${kube_env}['KUBELET_ARGS']
   $kubelet_args = $kubelet_args_str.Split(" ")
   Log-Output "kubelet_args from metadata: ${kubelet_args}"
+
+  # To join GCE instances to AD, we need to shorten their names, as NetBIOS name
+  # must be <= 15 characters, and GKE generated names are longer than that.
+  # To perform the join in an automated way, it's preferable to apply the rename
+  # and domain join in the GCESysprep step. However, after sysprep is complete
+  # and the machine restarts, kubelet bootstrapping should not use the shortened
+  # computer name, and instead use the instance's name by using --hostname-override,
+  # otherwise kubelet and kube-proxy will not be able to run properly.
+  $instance_name = "$(Get-InstanceMetadata 'name' | Out-String)"
   $default_kubelet_args = @(`
-      "--pod-infra-container-image=${env:INFRA_CONTAINER}"
+      "--pod-infra-container-image=${env:INFRA_CONTAINER}",
+      "--hostname-override=${instance_name}"
   )
+
   $kubelet_args = ${default_kubelet_args} + ${kubelet_args}
   if (-not (Test-NodeUsesAuthPlugin ${kube_env})) {
     Log-Output 'Using bootstrap kubeconfig for authentication'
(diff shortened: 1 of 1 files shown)


lizhuqi:
I've already signed a CLA

cluster/gce/windows/k8s-node-setup.psm1
```diff
@@ -1143,9 +1143,16 @@ function Start-WorkerServices {
   $kubelet_args_str = ${kube_env}['KUBELET_ARGS']
   $kubelet_args = $kubelet_args_str.Split(" ")
   Log-Output "kubelet_args from metadata: ${kubelet_args}"
+
+  # Windows nodes are renamed to a name <= 15 characters to join a AD domain.
```
pjh, line 1147: "are renamed" is vague. Can you be a little more specific in this comment: who/what does the renaming, and when, and under what conditions? Does the rename happen to the hostname at the Windows level, or to the instance name in GCE, or both?

cluster/gce/windows/k8s-node-setup.psm1
```diff
@@ -1143,9 +1143,16 @@ function Start-WorkerServices {
...
+  # Both kubelet and kube-proxy should use instance_name.
+  # No impact if a node is not renamed as hostname and instance name are the same.
+  $instance_name = "$(Get-InstanceMetadata 'name' | Out-String)"
   $default_kubelet_args = @(`
-      "--pod-infra-container-image=${env:INFRA_CONTAINER}"
+      "--pod-infra-container-image=${env:INFRA_CONTAINER}",
+      "--hostname-override=${instance_name}"
```
pjh, line 1153: Would there ever be a reason not to use --hostname-override?

We could make the use of this flag conditional based on whether or not some key is set in the kube-env, but not sure if that level of control is needed.

yliaog:
please test it for both GCE and GKE to ensure no surprises, i.e. create a cluster (GCE, and GKE), run related e2e tests.

cluster/gce/windows/k8s-node-setup.psm1
lizhuqi: We can check if the hostname and instance name are different and apply the flag when they are different only. Do you feel this will be better?

pjh: I assume that it wouldn't make a difference if we apply the flag when the hostname and instance name are the same.

I still have the question, would there ever be a reason not to use --hostname-override? If the answer is no, that's fine and we can leave this as is.

lizhuqi:
> please test it for both GCE and GKE to ensure no surprises, i.e. create a cluster (GCE, and GKE), run related e2e tests.

e2e tests pass

cluster/gce/windows/k8s-node-setup.psm1
lizhuqi: If we can make all GKE generated name to be <= 15 characters then we won't need to override the hostname. This change will not be trivial and have user impact as users are already familiar with today's naming convention. So it is not something we can complete very soon.  I can't think out a reason to not use --hostname-override for now.

lizhuqi:
/unassign yujuhong
/unassign phj
/unassign yliaog

Sorry guys, I was trying to request for review but wrongly used assign so remove all the assigns now.

pjh:
@liggitt perhaps you could help add this to the 1.20 milestone? It's low-risk, essentially a no-op for all k8s-on-Windows-on-GCE deployments right now.

liggitt:
I'll defer to @yliaog who has more expertise here

/cc @yliaog

PR #94274, opened 2020-08-27
[WIP] ignore node ready condition as they're handled by taint manager

yastij:
Signed-off-by: Yassine TIJANI <ytijani@vmware.com>

**What type of PR is this?**

/kind bug

**What this PR does / why we need it**: This PR ignores the node Ready condition from being processed by taintByNodeCondition feature as this condition should be handled by the taint manager. Before this PR,  nodes were tainted as `NoSchedule` where they should be tainted as `NoExecute` (using `NoExecute` should be fine wrt pod safety)

**Which issue(s) this PR fixes**: Fixes #

/assign @wojtek-t 
cc @andrewsykim 

```release-note

```

```docs

```

### DIFF
--- a/pkg/controller/nodelifecycle/node_lifecycle_controller.go
+++ b/pkg/controller/nodelifecycle/node_lifecycle_controller.go
@@ -624,12 +624,15 @@ func (nc *Controller) doNoScheduleTaintingPass(nodeName string) error {
 	// Map node's condition to Taints.
 	var taints []v1.Taint
 	for _, condition := range node.Status.Conditions {
-		if taintMap, found := nodeConditionToTaintKeyStatusMap[condition.Type]; found {
-			if taintKey, found := taintMap[condition.Status]; found {
-				taints = append(taints, v1.Taint{
-					Key:    taintKey,
-					Effect: v1.TaintEffectNoSchedule,
-				})
+		// ingnore nodeReady conditions as they're handled by the taint manager
+		if nc.runTaintManager && condition.Type != v1.NodeReady {
+			if taintMap, found := nodeConditionToTaintKeyStatusMap[condition.Type]; found {
+				if taintKey, found := taintMap[condition.Status]; found {
+					taints = append(taints, v1.Taint{
+						Key:    taintKey,
+						Effect: v1.TaintEffectNoSchedule,
+					})
+				}
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


yastij:
failures are expected, we'll need to update taints related to `taintByNodeCondition`

pkg/controller/nodelifecycle/node_lifecycle_controller.go
```diff
@@ -624,12 +624,15 @@ func (nc *Controller) doNoScheduleTaintingPass(nodeName string) error {
...
 	var taints []v1.Taint
 	for _, condition := range node.Status.Conditions {
-		if taintMap, found := nodeConditionToTaintKeyStatusMap[condition.Type]; found {
-			if taintKey, found := taintMap[condition.Status]; found {
-				taints = append(taints, v1.Taint{
-					Key:    taintKey,
-					Effect: v1.TaintEffectNoSchedule,
-				})
+		// ingnore nodeReady conditions as they're handled by the taint manager
+		if condition.Type != v1.NodeReady {
```
andrewsykim, line 628: This check should probably be conditional based on whether taint manager is running or not (see nc.runTaintManager).

yastij:
@liggitt - added a commit to showcase how the test should be wired, as we were testing the wrong behaviour IMO. I'll also add another commit to test `TaintByNodeCondition` but without taint-manager enabled

yastij:
ping @wojtek-t can you PTAL ?

pkg/controller/nodelifecycle/node_lifecycle_controller.go
```diff
@@ -624,12 +624,15 @@ func (nc *Controller) doNoScheduleTaintingPass(nodeName string) error {
...
 	var taints []v1.Taint
 	for _, condition := range node.Status.Conditions {
-		if taintMap, found := nodeConditionToTaintKeyStatusMap[condition.Type]; found {
-			if taintKey, found := taintMap[condition.Status]; found {
-				taints = append(taints, v1.Taint{
-					Key:    taintKey,
-					Effect: v1.TaintEffectNoSchedule,
-				})
+		// ingnore nodeReady conditions as they're handled by the taint manager
+		if nc.runTaintManager && condition.Type != v1.NodeReady {
```
michaelgugino, line 628: This is confusing.  Is this code not part of the taintManager itself?  If this code is part of taintManager, then saying 'nodeReady conditions handled by taint manager' doesn't make sense.

If this code is not the taint manager, then what should happen if runTaintManager is false?  Who will be setting these taints?

mrajashree: I have the same question ^. Also the bug says that Unreachable taint with NoExecute effect is not set for NodeReady=Unknown. With this change even the Unreachable=NoSchedule taint won't get set?

PR #94288, opened 2020-08-27
API server self request metric

### LINKED ISSUE #90908: Add a metric for apiserver self-requests
Currently we don't have an easy way of determining what number of requests are apiserver self-requests. This has historically bitten us, since we used to have a 50(?) QPS limit on apiserver self-requests (which thus has been removed). Even so, it would be a good idea to know this generally, since this may be helpful in pinpointing performance regressions for apiserver.

/sig api-machinery instrumentation

LogicalShark:
/kind feature
/sig api-machinery instrumentation

Adds self-request metric for the API server

Fixes https://github.com/kubernetes/kubernetes/issues/90908

```release-note
Added counter metric "apiserver_request_self" to count API server self-requests with labels for verb, resource, and subresource.
```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/BUILD
+++ b/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/BUILD
@@ -16,6 +16,7 @@ go_library(
         "//staging/src/k8s.io/apimachinery/pkg/types:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/util/sets:go_default_library",
         "//staging/src/k8s.io/apiserver/pkg/audit:go_default_library",
+        "//staging/src/k8s.io/apiserver/pkg/authentication/user:go_default_library",
         "//staging/src/k8s.io/apiserver/pkg/endpoints/request:go_default_library",
         "//staging/src/k8s.io/apiserver/pkg/features:go_default_library",
         "//staging/src/k8s.io/apiserver/pkg/util/feature:go_default_library",
--- a/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go
+++ b/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	utilsets "k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apiserver/pkg/audit"
+	"k8s.io/apiserver/pkg/authentication/user"
 	"k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/apiserver/pkg/features"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
@@ -185,6 +186,16 @@ var (
 		},
 		[]string{"verb", "group", "version", "resource", "subresource", "scope", "component", "code"},
 	)
+
+	apiSelfRequestCounter = compbasemetrics.NewCounterVec(
+		&compbasemetrics.CounterOpts{
+			Name:           "apiserver_selfrequest_total",
+			Help:           "Counter of apiserver self-requests broken out for each verb, API resource and subresource.",
+			StabilityLevel: compbasemetrics.ALPHA,
+		},
+		[]string{"verb", "resource", "subresource"},
+	)
+
 	kubectlExeRegexp = regexp.MustCompile(`^.*((?i:kubectl\.exe))`)
 
 	metrics = []resettableCollector{
@@ -201,6 +212,7 @@ var (
 		currentInflightRequests,
 		currentInqueueRequests,
 		requestTerminationsTotal,
+		apiSelfRequestCounter,
 	}
 
 	// these are the known (e.g. whitelisted/known) content types which we will report for
@@ -351,6 +363,11 @@ func MonitorRequest(req *http.Request, verb, group, version, resource, subresour
 	elapsedSeconds := elapsed.Seconds()
 	cleanContentType := cleanContentType(contentType)
 	requestCounter.WithLabelValues(reportedVerb, dryRun, group, version, resource, subresource, scope, component, cleanContentType, codeToString(httpCode)).Inc()
+	// MonitorRequest happens after authentication, so we can trust the username given by the request
+	info, ok := request.UserFrom(req.Context())
+	if ok && info.GetName() == user.APIServerUser {
+		apiSelfRequestCounter.WithLabelValues(reportedVerb, resource, subresource).Inc()
+	}
 	if deprecated {
 		deprecatedRequestGauge.WithLabelValues(group, version, resource, subresource, removedRelease).Set(1)
 		audit.AddAuditAnnotation(req.Context(), deprecatedAnnotationKey, "true")


LogicalShark:
@logicalhan

staging/src/k8s.io/apiserver/pkg/authentication/request/bearertoken/bearertoken.go
```diff
@@ -65,3 +53,21 @@ func (a *Authenticator) AuthenticateRequest(req *http.Request) (*authenticator.R
 
 	return resp, ok, err
 }
+
+// ExtractBearerToken gets the bearer token of a request
+func ExtractBearerToken(req *http.Request) (token string, ok bool) {
```
logicalhan, line 58: This looks unit-testable.

staging/src/k8s.io/apiserver/pkg/server/servermetrics.go
```diff
@@ -0,0 +1,48 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package server
+
+import (
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus"
```
logicalhan, line 22: We don't actually use prometheus directly. This would have to be registered to our wrapper registry (https://github.com/kubernetes/kubernetes/blob/eda07adf6e3ef8d066ba95a9b1039a13d2e3f8ea/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go#L266, for instance).

logicalhan (commented):
What is the rationale for not using the existing filter/entrypoint? Specifically I'm thinking `k8s.io/apiserver/pkg/endpoints/metrics/metrics.go`.

staging/src/k8s.io/apiserver/pkg/server/servermetrics.go
LogicalShark: Should this documentation be changed? https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/instrumentation.md

logicalhan: Yes, it should actually. Would you mind filing an issue so that we don't lose track of it?

LogicalShark: Done https://github.com/kubernetes/community/issues/5080

staging/src/k8s.io/apiserver/pkg/authentication/request/bearertoken/bearertoken.go
LogicalShark: Added

staging/src/k8s.io/apiserver/pkg/endpoints/filters/metrics.go
```diff
@@ -113,3 +133,18 @@ func compressUsername(username string) string {
 		return "other"
 	}
 }
+
+// splitSubresource checks if the given storage path is the path of a subresource and returns
+// the resource and subresource components.
+func splitSubresource(path string) (string, string, error) {
```
logicalhan, line 139: this also looks unit-testable

logicalhan (commented):
This is looking a lot better! I think we're getting close! Thanks for the patience during the iterations!

staging/src/k8s.io/apiserver/go.mod
```diff
@@ -27,6 +27,7 @@ require (
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822
 	github.com/pkg/errors v0.9.1
 	github.com/pquerna/cachecontrol v0.0.0-20171018203845-0dec1b30a021 // indirect
+	github.com/prometheus/client_golang v1.7.1
```
logicalhan, line 30: This probably shouldn't be here.

cmd/kube-scheduler/app/server.go
```diff
@@ -220,7 +220,7 @@ func buildHandlerChain(handler http.Handler, authn authenticator.Request, authz
 	failedHandler := genericapifilters.Unauthorized(scheme.Codecs)
 
 	handler = genericapifilters.WithAuthorization(handler, authz, scheme.Codecs)
-	handler = genericapifilters.WithAuthentication(handler, authn, failedHandler, nil)
+	handler = genericapifilters.WithAuthentication(handler, authn, failedHandler, nil, genericapifilters.NoopCheck)
```
logicalhan, line 223: this is inconsistent with `cmd/controller-manager/app/serve.go` which passes nil. 

Since you're handling the NoOp thing in `WithAuthentication`, it probably makes more sense to pass nil here.

staging/src/k8s.io/apiserver/pkg/endpoints/filters/authentication_test.go
```diff
@@ -91,6 +92,7 @@ func TestAuthenticateRequestWithAud(t *testing.T) {
 					}
 				}),
 				tc.apiAuds,
+				NoopCheck,
```
logicalhan, line 95: nit: nil for consistency

staging/src/k8s.io/apiserver/pkg/endpoints/filters/authentication_test.go
```diff
@@ -128,6 +130,7 @@ func TestAuthenticateRequest(t *testing.T) {
 			t.Errorf("unexpected call to failed")
 		}),
 		nil,
+		NoopCheck,
```
logicalhan, line 133: nit: nil for consistency

staging/src/k8s.io/apiserver/pkg/endpoints/filters/metrics.go
```diff
@@ -90,6 +100,16 @@ func recordAuthMetrics(resp *authenticator.Response, ok bool, err error, apiAudi
 
 	authenticatedAttemptsCounter.WithLabelValues(resultLabel).Inc()
 	authenticationLatency.WithLabelValues(resultLabel).Observe(time.Since(authStart).Seconds())
+
+	if isAPISelfRequest {
+		// When to use canonicalVerb from apiserver/pkg/endpoints/metrics.go?
```
logicalhan, line 105: I would use `canonicalVerb` here.

staging/src/k8s.io/apiserver/pkg/endpoints/filters/metrics_test.go
```diff
@@ -146,6 +146,7 @@ func TestMetrics(t *testing.T) {
 					close(done)
 				}),
 				tt.apiAudience,
+				NoopCheck,
```
logicalhan, line 149: nit: nil for consistency

logicalhan (commented):
It's almost there! Thanks for iterating over this!

logicalhan:
FYI you are still failing the CLA check. 

```If you've already signed a CLA, it's possible we don't have your GitHub username or you're using a different email address. Check your existing CLA data and verify that your email is set on your git commits.```

**Review by logicalhan — COMMENTED** [review]
ping? it's looking good, just a couple of things plus a squash and it should be ready for review.

**Comment by logicalhan** [comment]
> ```
> Added a metric to count API server self-requests.
> ```

Would you mind adding the literal metric name and the labels here?

Also, can you make this PR no longer WIP/draft state?

File: staging/src/k8s.io/apiserver/pkg/authentication/request/bearertoken/bearertoken.go [hunk 10]
```diff
@@ -65,3 +53,21 @@ func (a *Authenticator) AuthenticateRequest(req *http.Request) (*authenticator.R
 
 	return resp, ok, err
 }
+
+// ExtractBearerToken gets the bearer token of a request, "ok" will be false if the token is missing
+func ExtractBearerToken(req *http.Request) (token string, ok bool) {
```
lavalamp, line 58: Can you send this move as a separate PR? It's not good to mix refactoring and functionality change.

staging/src/k8s.io/apiserver/pkg/endpoints/filters/authentication.go
```diff
@@ -27,27 +27,46 @@ import (
...
+func NoopCheck(req *http.Request) bool {
+	return false
+}
+
 // WithAuthentication creates an http handler that tries to authenticate the given request as a user, and then
 // stores any such user found onto the provided context for the request. If authentication fails or returns an error
 // the failed handler is used. On success, "Authorization" header is removed from the request and handler
 // is invoked to serve the request.
-func WithAuthentication(handler http.Handler, auth authenticator.Request, failed http.Handler, apiAuds authenticator.Audiences) http.Handler {
+func WithAuthentication(handler http.Handler, auth authenticator.Request, failed http.Handler, apiAuds authenticator.Audiences, apiSelfRequestCheck func(*http.Request) bool, requestInfoFactory *apirequest.RequestInfoFactory) http.Handler {
```
lavalamp, line 45: When I said the authentication stuff knows this, I was not expecting you to modify this method, I actually don't recommend this, sorry for not being clear!

Instead, I was expecting a wrapper around the specific authenticator that handles the self requests, that passes through requests and increments the metric if that authenticator returns true.

staging/src/k8s.io/apiserver/pkg/server/config.go
```diff
@@ -786,3 +787,12 @@ func AuthorizeClientBearerToken(loopback *restclient.Config, authn *Authenticati
 	tokenAuthorizer := authorizerfactory.NewPrivilegedGroups(user.SystemPrivilegedGroup)
 	authz.Authorizer = authorizerunion.New(tokenAuthorizer, authz.Authorizer)
 }
+
+// IsSelfRequest returns whether req is a self-request on the api server
+func (c *Config) IsSelfRequest(req *http.Request) bool {
+	tok, ok := bearertoken.ExtractBearerToken(req)
+	if ok && (tok == c.LoopbackClientConfig.BearerToken) {
```
lavalamp, line 794: Yeah, don't do this here. Look for where an authenticator is constructed to match this token instead.

PR #94290, opened 2020-08-27
Fix prefixed global flags for kubectl plugins

### LINKED ISSUE #93432: Kubectl parsing of --all-namespaces flag is broken or at best, confusing
**What happened**:

Running a `kubectl get` with --all-namespaces in the wrong place consumes one extra argument, resulting in a confusing error.
 
A sane error.

```
$ k get pods
NAME                         READY   STATUS    RESTARTS   AGE
hostnames-6956d6795f-lgv5m   1/1     Running   0          14d

$ k -n default get pods
NAME                         READY   STATUS    RESTARTS   AGE
hostnames-6956d6795f-lgv5m   1/1     Running   0          14d

$ k --all-namespaces get pods
Error: unknown command "pods" for "kubectl"

Did you mean this?
	logs

Run 'kubectl --help' for usage.

$ k get pods --all-namespaces 
NAMESPACE     NAME                                                       READY   STATUS             RESTARTS   AGE
# lots of result
```

**Environment**:

This is a kubectl build at HEAD

/sig cli

cmurphy:
**What type of PR is this?**

/kind bug

For normal kubectl commands, global flags can be used anywhere within
the command line:

```
  $ kubectl --kubeconfig admin.conf get pod
  No resources found in default namespace.
  $ kubectl get --kubeconfig admin.conf pod
  No resources found in default namespace.
  $ kubectl get pod --kubeconfig admin.conf
  No resources found in default namespace.
```

However, without this patch, when calling a plugin as a subcommand, the
flag can only work if it comes after the plugin subcommand:

```
  $ kubectl --kubeconfig admin.conf hns tree default
  Error: unknown command "hns" for "kubectl"
  Run 'kubectl --help' for usage.
  $ kubectl hns --kubeconfig admin.conf tree default
  default
  $ kubectl hns tree --kubeconfig admin.conf default
  default
  $ kubectl hns tree default --kubeconfig admin.conf
  default
```

This is an inconsistency that can lead to the user confusion regarding
whether they have installed the plugin correctly and so is an unfriendly
user experience. This patch corrects the issue for plugin handling by
relocating persistent flags to the end of the argument list before
attempting to search for the plugin binary.

Fixes #93432

```release-note
Fixed parsing of prefixed global flags for kubectl plugins.
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/kubectl/pkg/cmd/cmd_test.go
+++ b/staging/src/k8s.io/kubectl/pkg/cmd/cmd_test.go
@@ -65,6 +65,12 @@ func TestKubectlCommandHandlesPlugins(t *testing.T) {
 			name:             "test that normal commands are able to be executed, when no plugin overshadows them",
 			args:             []string{"kubectl", "get", "foo"},
 			expectPlugin:     "",
+			expectPluginArgs: nil,
+		},
+		{
+			name:             "test that a plugin executable is found with no flags",
+			args:             []string{"kubectl", "foo"},
+			expectPlugin:     "plugin/testdata/kubectl-foo",
 			expectPluginArgs: []string{},
 		},
 		{
@@ -103,8 +109,8 @@ func TestKubectlCommandHandlesPlugins(t *testing.T) {
 				t.Fatalf("unexpected plugin execution: expected %q, got %q", test.expectPlugin, pluginsHandler.executedPlugin)
 			}
 
-			if len(pluginsHandler.withArgs) != len(test.expectPluginArgs) {
-				t.Fatalf("unexpected plugin execution args: expected %q, got %q", test.expectPluginArgs, pluginsHandler.withArgs)
+			if !reflect.DeepEqual(pluginsHandler.withArgs, test.expectPluginArgs) {
+				t.Fatalf("unexpected plugin execution args: expected %#v, got %#v", test.expectPluginArgs, pluginsHandler.withArgs)
 			}
 		})
 	}


pkg/kubectl/cmd/cmd.go
```diff
@@ -303,6 +304,49 @@ func NewDefaultKubectlCommand() *cobra.Command {
...
+	return flag.NoOptDefVal != ""
+}
+
+func appendPersistentFlagsAfterCommand(args []string, c *cobra.Command) []string {
+	flags := c.PersistentFlags()
+	for i := 0; i < len(args); i++ {
+		s := args[i]
+		switch {
+		case strings.HasPrefix(s, "--") && !strings.Contains(s, "=") && !hasNoOptDefVal(s[2:], flags):
+			// If '--flag arg' then jump past flag and arg
```
brianpursley, line 333: arg is optional for boolean flag, so I don't think you can make this assumption.  For example `--all-namespaces`

pkg/kubectl/cmd/cmd.go
```diff
@@ -303,6 +304,49 @@ func NewDefaultKubectlCommand() *cobra.Command {
...
+	}
+	return flag.NoOptDefVal != ""
+}
+
+func appendPersistentFlagsAfterCommand(args []string, c *cobra.Command) []string {
+	flags := c.PersistentFlags()
+	for i := 0; i < len(args); i++ {
+		s := args[i]
+		switch {
+		case strings.HasPrefix(s, "--") && !strings.Contains(s, "=") && !hasNoOptDefVal(s[2:], flags):
```
brianpursley, line 332: Trying to wrap my head around what this case is saying. I find the double-negative confusing `!hasNoOptDefVal`.

I think this case is matching an arg starting with `--`, without an `=` and without a default value?

Maybe rename `hasNoOptDefVal` so you don't have to negate it.

brianpursley:
https://github.com/kubernetes/kubectl/issues/884

brianpursley:
@cmurphy I think you need to run `hack/update-bazel.sh` to fix the build error.

brianpursley:
This is a tricky issue to solve completely and has been discussed several times before.
https://github.com/kubernetes/kubectl/issues/884
https://github.com/kubernetes/kubernetes/issues/77666
https://github.com/kubernetes/kubernetes/issues/93432

I think this PR would help though, by relocating as much as possible after the plugin name.  This would relocate global flags, and even non-global flags with no arg or that do not have a space before the arg.

There is still an edge case where if you provide a non-global flag with a space before the arg, the plugin it will fail.... ie. `kubectl --foo bar myplugin`. kubectl will think `bar` is the plugin name. Without knowing the plugin's flag set it is more difficult to try to solve this one.

I think this PR is an improvement though and fixes *most* cases.

/lgtm

cmurphy:
[**@brianpursley**](https://github.com/brianpursley) commented on [Aug 31, 2020, 7:28 AM PDT](https://github.com/kubernetes/kubernetes/pull/94290#issuecomment-683813933 "2020-08-31T14:28:03Z - Replied by Github Reply Comments"):
> This is a tricky issue to solve completely and has been discussed several times before.  
> [kubernetes/kubectl#884](https://github.com/kubernetes/kubectl/issues/884)  
> [#77666](https://github.com/kubernetes/kubernetes/issues/77666)  
> [#93432](https://github.com/kubernetes/kubernetes/issues/93432)

Thanks for the review and the background information!

> 
> I think this PR would help though, by relocating as much as possible after the plugin name. This would relocate global flags, and even non-global flags with no arg or that do not have a space before the arg.
> 
> There is still an edge case where if you provide a non-global flag with a space before the arg, the plugin it will fail.... ie. `kubectl --foo bar myplugin`. kubectl will think `bar` is the plugin name. Without knowing the plugin's flag set it is more difficult to try to solve this one.

I think this is an oversight on my part, I don't think it should work this way. I think whether the flag-before-the-command has an argument or not, it should be treated the same way, and in this case I think if it is not a recognized global flag it should just be rejected because the "unknown command myplugin" is more confusing than an unrecognized flag. I'll try to fix this.

> 
> I think this PR is an improvement though and fixes _most_ cases.
> 
> /lgtm

I should mention (and will add in a comment) that the logic and function names are largely inspired by similar functionality in [cobra](https://github.com/spf13/cobra/blob/02a0d2fbc9e61d26f8e5979749f6030964a55a3e/command.go#L558-L594) - I would have reused it more directly if any of it was exported.

pkg/kubectl/cmd/cmd.go
```diff
@@ -303,6 +304,44 @@ func NewDefaultKubectlCommand() *cobra.Command {
 	return NewDefaultKubectlCommandWithArgs(NewDefaultPluginHandler(plugin.ValidPluginFilenamePrefixes), os.Args, os.Stdin, os.Stdout, os.Stderr)
 }
 
+func lookupFlag(flag string, fs *pflag.FlagSet) *pflag.Flag {
+	flag = strings.SplitN(flag, "=", 2)[0]
+	switch {
+	case strings.HasPrefix(flag, "--"):
+		return fs.Lookup(flag[2:])
+	case strings.HasPrefix(flag, "-"):
+		return fs.ShorthandLookup(flag[1:2])
```
brianpursley, line 313: Is this OK?

My concern is if someone accidentally does `-` when they meant `--` and if this only takes the first character after the `-` then it could cause the wrong flag to be used instead of failing.

For example, in `kubectl run` there is an `--image` flag and a `-i` flag which is shorthand for `--stdin`.

In this case it probably would not cause a bad problem because the command would fail, but if I used `-image` instead of `--image`, I think the code above would interpret `-image` as `-i`, right?

What if it just does this?
```
return fs.ShorthandLookup(flag[1:])
```

Or return an error if the length of the single-dash flag is not 1.

Just an idea

brianpursley:
Looks pretty good I think. Just one minor comment.

Flag parsing is a little hard to imagine given all the scenarios, but I like that you have tests to cover them.

pkg/kubectl/cmd/cmd.go
cmurphy: Thanks for catching that, I handled it in the original version and mistakenly removed it

brianpursley:
Looks good. Can you squash your commits into a single commit?

cmurphy:
[**@brianpursley**](https://github.com/brianpursley) commented on [Sep 4, 2020, 10:30 AM PDT](https://github.com/kubernetes/kubernetes/pull/94290#issuecomment-687284953 "2020-09-04T17:30:50Z - Replied by Github Reply Comments"):
> Looks good. Can you squash your commits into a single commit?

done

brianpursley (approved):

brianpursley:
note to other reviewers. I know we have discussed this topic before on several occasions (I liked above in a previous comment).

The reason I think this PR is fine because it improves the handling of global flags before the plugin name, which helps in the majority of the problem cases where people have aliased kubectl with a flag included.  

It still cannot handle non-global flags before the plugin name, but I think that is just going to have to remain a known limitation.

soltysh (changes requested):
@cmurphy I greatly appreciate your effort and time you spend putting this PR together but as mentioned in https://github.com/kubernetes/kubernetes/pull/92343 and several other places, which @brianpursley also mentioned sig-cli decided (see https://docs.google.com/document/d/1r0YElcXt6G5mOWxwZiXgGu_X6he3F--wKwg-9UBc29I/edit#heading=h.ji0nwpg7dyuz) that we're not going to re-schuffle flags but rather we'll put a warning in place. I've spent decent amount of time carefully examining your PR today and I was still able to find issues with global flags. Namely passing `--v=2`, `-v=2` or `-v 2` which works just fine with `kubectl` does not b/c `-v` is not registered as global flags. We could theoretically have a list of exceptions but I'm worried we'd be playing never-ending catch up game with that. 

My proposal for this PR and the overall problem are as follows. For this PR I'd like to see the current test case fixed (see my comment below about comparing counts and not actual arguments passed to plugins) and a new test which will run plugin commands with no flags (1), make sure they get routed to, then run with flags after the commands and make sure they get routed to (2), then run with flags before the commands and make sure it errors (3). 

Secondly, in the long run I think we should look into cobra flag parsing mechanism and check what we could do to make the flag parsing more predictable. 

/hold

cmurphy:
Thank you for the thorough review, @soltysh !

> as mentioned in #92343 and several other places, which @brianpursley also mentioned sig-cli decided (see https://docs.google.com/document/d/1r0YElcXt6G5mOWxwZiXgGu_X6he3F--wKwg-9UBc29I/edit#heading=h.ji0nwpg7dyuz) that we're not going to re-schuffle flags but rather we'll put a warning in place. I've spent decent amount of time carefully examining your PR today and I was still able to find issues with global flags. Namely passing --v=2, -v=2 or -v 2 which works just fine with kubectl does not b/c -v is not registered as global flags. We could theoretically have a list of exceptions but I'm worried we'd be playing never-ending catch up game with that.

Could you clarify here - why is `-v` not registered as a global flag? From my experimentation, it seems like it is:

```
$ git diff
diff --git a/pkg/kubectl/cmd/cmd.go b/pkg/kubectl/cmd/cmd.go
index 9b01843796a..1cdede694bc 100644
--- a/pkg/kubectl/cmd/cmd.go
+++ b/pkg/kubectl/cmd/cmd.go
@@ -650,6 +650,8 @@ func NewKubectlCommand(in io.Reader, out, err io.Writer) *cobra.Command {
        cmds.AddCommand(apiresources.NewCmdAPIResources(f, ioStreams))
        cmds.AddCommand(options.NewCmdOptions(ioStreams.Out))
 
+       fmt.Printf("%#v\n", cmds.PersistentFlags().Lookup("v"))
+
        return cmds
 }
$ _output/bin/kubectl get pod
&pflag.Flag{Name:"v", Shorthand:"v", Usage:"number for the log level verbosity", Value:(*pflag.flagValueWrapper)(0xc0004278e0), DefValue:"0", Changed:false, NoOptDefVal:"", Deprecated:"", Hidden:false, ShorthandDeprecated:"", Annotations:map[string][]string(nil)}
No resources found in default namespace.
```

A plugin may still not respect -v even though it is a global flag, but it now gives an error message about the flag, the same as it would if the flag came at the end:

```
$ /usr/bin/kubectl version --client
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-15T00:00:00Z", GoVersion:"go1.15.3", Compiler:"gc", Platform:"linux/amd64"}
$ /usr/bin/kubectl hns tree default --v=2
Error: unknown flag: --v
Usage:
  kubectl tree [flags]
...
$ _output/bin/kubectl version --client
Client Version: version.Info{Major:"1", Minor:"20+", GitVersion:"v1.20.0-alpha.0.1203+ee15e434cb2686", GitCommit:"ee15e434cb2686043c7d1c6b6eaf957a5ce6e02d", GitTreeState:"clean", BuildDate:"2020-10-28T20:10:17Z", GoVersion:"go1.15.3", Compiler:"gc", Platform:"linux/amd64"}
$ _output/bin/kubectl --v=2 hns tree default 
Error: unknown flag: --v
```

There was an issue with the unit tests where the registered name of the flag became `"test.v"` for reasons I don't understand so I didn't use it in the tests, but I could dig into that.

Regarding the concern about being able to tell the difference between a flag that has an argument and a flag without an argument, I think I have covered that as well:

```
$ _output/bin/kubectl --insecure-skip-tls-verify hns tree default # boolean flag
default
$ _output/bin/kubectl --insecure-skip-tls-verify=true hns tree default # boolean with =
default
$ _output/bin/kubectl --namespace=foo hns tree default # flag with arguments with =
default
$ _output/bin/kubectl --namespace foo hns tree default # flag with arguments without =
default
$ _output/bin/kubectl --namespace hns tree default # expected error since the plugin name is interpreted as belonging to the flag
Error: unknown command "tree" for "kubectl"
Run 'kubectl --help' for usage.
```

What case of flags did I miss where this is doing the wrong thing? Or is the last case in this example what's not acceptable?

I support just emitting an error as in the the other PR, that is certainly simpler and fixes the essence of the problem. I just want to make sure I understand where I went wrong here so I can learn :)

> My proposal for this PR and the overall problem are as follows. For this PR I'd like to see the current test case fixed (see my comment below about comparing counts and not actual arguments passed to plugins) and a new test which will run plugin commands with no flags (1), make sure they get routed to, then run with flags after the commands and make sure they get routed to (2), 

Happy to turn this PR into fixing the tests

> then run with flags before the commands and make sure it errors (3).

This part I think should either wait for https://github.com/kubernetes/kubernetes/pull/92343 to merge, or perhaps just be part of that PR?

Thanks again for your time!

soltysh:
> Could you clarify here - why is `-v` not registered as a global flag? From my experimentation, it seems like it is:

IIRC the main issues is that `-v` (with single hyphen is coming from glog, not sure if this still applies to klog, but looking through https://github.com/kubernetes/kubernetes/blob/6dddea5abfa20a22e93114d66735441f04f4a39c/cmd/kubectl/kubectl.go#L46 https://github.com/kubernetes/kubernetes/blob/6dddea5abfa20a22e93114d66735441f04f4a39c/staging/src/k8s.io/kubectl/pkg/util/logs/logs.go#L33 you'll get to https://github.com/kubernetes/kubernetes/blob/6dddea5abfa20a22e93114d66735441f04f4a39c/vendor/k8s.io/klog/v2/klog.go#L433 which is behind this. With kubectl in staging we can probably try separating these two somehow to not to rely on this, but we need to ensure that still the verbosity setting works as expected. 

> What case of flags did I miss where this is doing the wrong thing? Or is the last case in this example what's not acceptable?

I can't remember exactly, but when I checked this last time not all of the ones you mentioned worked as I expect them. I'd need to give it another try. 

> Happy to turn this PR into fixing the tests

yeah - fixing tests is always nice seen, the more tests the better for us.

> This part I think should either wait for #92343 to merge, or perhaps just be part of that PR?

That merged by now.

cmurphy:
Removed the changes to the plugin flag handling, fixed the unit tests to do a real comparison of the argument list and added a test for when no flag is provided to the plugin.

I couldn't test the error case, because the plugin simply does an immediate exit 1 instead of returning, so the unit tests aren't able to handle it safely:

https://github.com/kubernetes/kubernetes/blob/e64ebe01314b41fa68cff827a31ab133ec6963a4/staging/src/k8s.io/kubectl/pkg/cmd/cmd.go#L321

Moreover, the error handling introduced in https://github.com/kubernetes/kubernetes/pull/92343 doesn't work if there is only one prefixed flag and one command in the plugin name:

```
$ cat ~/bin/kubectl-foo
echo "My first command-line argument was $1"
$ _output/bin/kubectl version --client
Client Version: version.Info{Major:"1", Minor:"20+", GitVersion:"v1.20.0-beta.2.29+844561654d6f0c", GitCommit:"844561654d6f0ca2d6956af37c295d351e42a64c", GitTreeState:"clean", BuildDate:"2020-11-19T01:00:10Z", GoVersion:"go1.15.4", Compiler:"gc", Platform:"linux/amd64"}
$ _output/bin/kubectl --bar foo
Error: unknown flag: --bar
See 'kubectl --help' for usage.
$ _output/bin/kubectl --bar foo bar
Error: flags cannot be placed before plugin name: --bar
```

PR #94294, opened 2020-08-28
Remove duplicate nodeSelector

millermatt:
/kind bug

Makes the dns-horizontal-autoscaler.yaml file valid yaml.


MrHohn (approved):
Thanks!
/lgtm
/approve

PR #94299, opened 2020-08-28
fix kubeadm update coredns with skip pending pod

### LINKED ISSUE #2267: isCoreDNSVersionSupported: avoid panics and wait for pending containers
Following up from #2078:

```console
[daniel@reef kured ]$ minikube start --vm-driver kvm2 --kubernetes-version 1.19.0
😄  minikube v1.12.3 on Ubuntu 20.04
✨  Using the kvm2 driver based on existing profile
👍  Starting control plane node minikube in cluster minikube
🏃  Updating the running kvm2 "minikube" VM ...
🐳  Preparing Kubernetes v1.19.0 on Docker 19.03.12 ...
🤦  Unable to restart cluster, will reset it: addons: /bin/bash -c "sudo env PATH=/var/lib/minikube/binaries/v1.19.0:$PATH kubeadm init phase addon all --config /var/tmp/minikube/kubeadm.yaml": Process exited with status 2
stdout:

stderr:
W0828 09:07:48.306773    5151 configset.go:348] WARNING: kubeadm cannot validate component configs for API groups [kubelet.config.k8s.io kubeproxy.config.k8s.io]
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns.isCoreDNSVersionSupported(0x1d5b6a0, 0xc0001491e0, 0xc00060c000, 0xc0005b4000, 0x153)
	/workspace/anago-v1.19.0-rc.4.197+594f888e19d8da/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns/dns.go:413 +0x38f
k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns.createCoreDNSAddon(0xc0005e4000, 0x7fb, 0xb5f, 0xc0008b1680, 0x309, 0x418, 0xc00080f800, 0x20a, 0x3dc, 0x1d5b6a0, ...)
	/workspace/anago-v1.19.0-rc.4.197+594f888e19d8da/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns/dns.go:271 +0x187
k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns.coreDNSAddon(0xc0004f8da0, 0x1d5b6a0, 0xc0001491e0, 0xc0005ffafc, 0x0, 0x0)
	/workspace/anago-v1.19.0-rc.4.197+594f888e19d8da/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns/dns.go:252 +0x505
k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns.EnsureDNSAddon(0xc0004f8da0, 0x1d5b6a0, 0xc0001491e0, 0x1d5b6a0, 0xc0001491e0)
	/workspace/anago-v1.19.0-rc.4.197+594f888e19d8da/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns/dns.go:109 +0x13e
k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/init.runCoreDNSAddon(0x1a58420, 0xc00003c9a0, 0xc0004cdca0, 0xd)
	/workspace/anago-v1.19.0-rc.4.197+594f888e19d8da/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/init/addons.go:92 +0x85
k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow.(*Runner).Run.func1(0xc00061f600, 0x0, 0x0)
	/workspace/anago-v1.19.0-rc.4.197+594f888e19d8da/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow/runner.go:234 +0x1e8
k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow.(*Runner).visitAll(0xc0006bcb40, 0xc0007f3ce0, 0x0, 0x2)
	/workspace/anago-v1.19.0-rc.4.197+594f888e19d8da/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow/runner.go:422 +0x6e
k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow.(*Runner).Run(0xc0006bcb40, 0xc0003c6fc0, 0x0, 0x2, 0x726420, 0xc000521d48)
	/workspace/anago-v1.19.0-rc.4.197+594f888e19d8da/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow/runner.go:207 +0x153
k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow.(*Runner).BindToCommand.func1.1(0xc0002b1080, 0xc0003c6fc0, 0x0, 0x2, 0x0, 0x0)
	/workspace/anago-v1.19.0-rc.4.197+594f888e19d8da/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow/runner.go:348 +0xda
k8s.io/kubernetes/vendor/github.com/spf13/cobra.(*Command).execute(0xc0002b1080, 0xc0003c6f60, 0x2, 0x2, 0xc0002b1080, 0xc0003c6f60)
	/workspace/anago-v1.19.0-rc.4.197+594f888e19d8da/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/github.com/spf13/cobra/command.go:842 +0x47c
k8s.io/kubernetes/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc00038a840, 0xc00000e010, 0x1cecec0, 0xc00000e018)
	/workspace/anago-v1.19.0-rc.4.197+594f888e19d8da/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/github.com/spf13/cobra/command.go:950 +0x375
k8s.io/kubernetes/vendor/github.com/spf13/cobra.(*Command).Execute(...)
	/workspace/anago-v1.19.0-rc.4.197+594f888e19d8da/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/github.com/spf13/cobra/command.go:887
k8s.io/kubernetes/cmd/kubeadm/app.Run(0x0, 0x18249a0)
	/workspace/anago-v1.19.0-rc.4.197+594f888e19d8da/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/app/kubeadm.go:50 +0x225
main.main()
	_output/dockerized/go/src/k8s.io/kubernetes/cmd/kubeadm/kubeadm.go:25 +0x25

🔎  Verifying Kubernetes components...
🌟  Enabled addons: default-storageclass, storage-provisioner
🏄  Done! kubectl is now configured to use "minikube"
[daniel@reef kured ]$ minikube version
minikube version: v1.12.3
commit: 2243b4b97c131e3244c5f014faedca0d846599f5-dirty
[daniel@reef kured ]$ minikube kubectl -- version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0", GitCommit:"e19964183377d0ec2052d1f1fa930c4d7575bd50", GitTreeState:"clean", BuildDate:"2020-08-26T14:30:33Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0", GitCommit:"e19964183377d0ec2052d1f1fa930c4d7575bd50", GitTreeState:"clean", BuildDate:"2020-08-26T14:23:04Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
[daniel@reef kured ]$ 

```

This is with minikube 1.12.3, k8s 1.19.0 and a manual restart of minikube (because of https://github.com/kubernetes/minikube/issues/2874) as I wanted to test the new kured release.

_Originally posted by @dholbach in https://github.com/kubernetes/kubeadm/issues/2078#issuecomment-682420921_

---------

related report report by @kvaps here:
https://github.com/kubernetes/kubernetes/issues/94286

> **What happened**:
> 
> during update to v1.19 can not upgrade coredns deployment:
> 
> ```
> # kubeadm init phase addon coredns --config /config/kubeadmcfg.yaml
> W0827 17:45:09.167031     251 kubelet.go:200] cannot automatically set CgroupDriver when starting the Kubelet: cannot execute 'docker info -f {{.CgroupDriver}}': executable file not found in $PATH
> W0827 17:45:09.167066     251 kubelet.go:213] cannot determine if systemd-resolved is active: no supported init system detected, skipping checking for services
> W0827 17:45:09.475045     251 configset.go:348] WARNING: kubeadm cannot validate component configs for API groups [kubelet.config.k8s.io kubeproxy.config.k8s.io]
> error execution phase addon/coredns: unable to match SHA256 digest ID in ""
> To see the stack trace of this error execute with --v=5 or higher
> ```
> 
> **What you expected to happen**:
> 
> cordns succefully upgraded
> 
> **How to reproduce it (as minimally and precisely as possible)**
> 
> try to apply my manifests: [coredns.txt](https://github.com/kubernetes/kubernetes/files/5137973/coredns.txt), and do:
> 
> ```
> kubeadm init phase addon coredns
> ```
> 
> **Anything else we need to know?**:
> 
> **Environment**:
> 
> * Kubernetes version (use `kubectl version`): v1.18.8 --> v1.19.0
> * Cloud provider or hardware configuration: kubefarm
> * OS (e.g: `cat /etc/os-release`): ubuntu 20.04.1
> * Kernel (e.g. `uname -a`): 5.4.0-42-generic
> * Install tools: kubeadm
> * Network plugin and version (if this is a network-related bug):
> * Others:
>   
>   * [detailed_log.txt](https://github.com/kubernetes/kubernetes/files/5137998/detailed_log.txt)
>   * [coredns-pods.txt](https://github.com/kubernetes/kubernetes/files/5138023/coredns-pods.txt)

### LINKED ISSUE #94286: error execution phase addon/coredns: unable to match SHA256 digest ID in ""
**What happened**:

during update to v1.19 can not upgrade coredns deployment:

```
# kubeadm init phase addon coredns --config /config/kubeadmcfg.yaml
W0827 17:45:09.167031     251 kubelet.go:200] cannot automatically set CgroupDriver when starting the Kubelet: cannot execute 'docker info -f {{.CgroupDriver}}': executable file not found in $PATH
W0827 17:45:09.167066     251 kubelet.go:213] cannot determine if systemd-resolved is active: no supported init system detected, skipping checking for services
W0827 17:45:09.475045     251 configset.go:348] WARNING: kubeadm cannot validate component configs for API groups [kubelet.config.k8s.io kubeproxy.config.k8s.io]
error execution phase addon/coredns: unable to match SHA256 digest ID in ""
To see the stack trace of this error execute with --v=5 or higher
```

cordns succefully upgraded

**How to reproduce it (as minimally and precisely as possible)**

try to apply my manifests: [coredns.txt](https://github.com/kubernetes/kubernetes/files/5137973/coredns.txt), and do:
```
kubeadm init phase addon coredns
```

**Environment**:
- Kubernetes version (use `kubectl version`): v1.18.8 --> v1.19.0
- Cloud provider or hardware configuration: kubefarm
- OS (e.g: `cat /etc/os-release`): ubuntu 20.04.1
- Kernel (e.g. `uname -a`): 5.4.0-42-generic
- Install tools: kubeadm
- Others:
  - [detailed_log.txt](https://github.com/kubernetes/kubernetes/files/5137998/detailed_log.txt)
  - [coredns-pods.txt](https://github.com/kubernetes/kubernetes/files/5138023/coredns-pods.txt)

zouyee:
**What type of PR is this?**
/kind bug

 Due to the imageID field of the pod can only be set when it is scheduled and running, it should be skipped when pod is pending.

Fixes kubernetes/kubeadm#2267

```release-note
kubeadm: avoid a panic when determining if the running version of CoreDNS is supported during upgrades
```

```docs
NONE
```

### DIFF
--- a/cmd/kubeadm/app/phases/addons/dns/dns.go
+++ b/cmd/kubeadm/app/phases/addons/dns/dns.go
@@ -23,6 +23,7 @@ import (
 	"net"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/caddyserver/caddy/caddyfile"
 	"github.com/coredns/corefile-migration/migration"
@@ -34,6 +35,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kuberuntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 	clientsetscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/klog/v2"
@@ -398,29 +400,48 @@ var (
 )
 
 func isCoreDNSVersionSupported(client clientset.Interface) (bool, error) {
-	isValidVersion := true
-	coreDNSPodList, err := client.CoreV1().Pods(metav1.NamespaceSystem).List(
-		context.TODO(),
-		metav1.ListOptions{
-			LabelSelector: "k8s-app=kube-dns",
-		},
-	)
+	var lastError error
+	var pods []v1.Pod
+
+	pollTimeout := 10 * time.Second
+	err := wait.PollImmediate(kubeadmconstants.APICallRetryInterval, pollTimeout, func() (bool, error) {
+		coreDNSPodList, err := client.CoreV1().Pods(metav1.NamespaceSystem).List(
+			context.TODO(),
+			metav1.ListOptions{
+				LabelSelector: "k8s-app=kube-dns",
+			},
+		)
+		if err != nil {
+			lastError = err
+			return false, nil
+		}
+
+		for _, pod := range coreDNSPodList.Items {
+			if pod.Status.Phase != v1.PodRunning {
+				lastError = errors.New("found non-running CoreDNS pods")
+				return false, nil
+			}
+		}
+		pods = coreDNSPodList.Items
+		return true, nil
+	})
+
 	if err != nil {
-		return false, errors.Wrap(err, "unable to list CoreDNS pods")
+		return false, errors.Wrapf(lastError, "could not list the running CoreDNS pods after %v", pollTimeout)
 	}
 
-	for _, pod := range coreDNSPodList.Items {
+	for _, pod := range pods {
 		imageID := imageDigestMatcher.FindStringSubmatch(pod.Status.ContainerStatuses[0].ImageID)
 		if len(imageID) != 2 {
-			return false, errors.Errorf("unable to match SHA256 digest ID in %q", pod.Status.ContainerStatuses[0].ImageID)
+			return false, errors.Errorf("pod %s unable to match SHA256 digest ID in %q", pod.GetName(), pod.Status.ContainerStatuses[0].ImageID)
 		}
 		// The actual digest should be at imageID[1]
 		if !migration.Released(imageID[1]) {
-			isValidVersion = false
+			return false, errors.Errorf("unknown digest %q for pod %s", imageID[1], pod.GetName())
 		}
 	}
 
-	return isValidVersion, nil
+	return true, nil
 }
 
 func migrateCoreDNSCorefile(client clientset.Interface, cm *v1.ConfigMap, corefile, currentInstalledCoreDNSVersion string) error {
--- a/cmd/kubeadm/app/phases/addons/dns/BUILD
+++ b/cmd/kubeadm/app/phases/addons/dns/BUILD
@@ -46,6 +46,7 @@ go_library(
         "//staging/src/k8s.io/apimachinery/pkg/apis/meta/v1:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/runtime:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/types:go_default_library",
+        "//staging/src/k8s.io/apimachinery/pkg/util/wait:go_default_library",
         "//staging/src/k8s.io/client-go/kubernetes:go_default_library",
         "//staging/src/k8s.io/client-go/kubernetes/scheme:go_default_library",
         "//vendor/github.com/caddyserver/caddy/caddyfile:go_default_library",


cmd/kubeadm/app/phases/addons/dns/dns.go
```diff
@@ -410,6 +410,9 @@ func isCoreDNSVersionSupported(client clientset.Interface) (bool, error) {
 	}
 
 	for _, pod := range coreDNSPodList.Items {
+		if pod.Status.Phase == v1.PodPending {
+			continue
+		}
```
neolit123, line 415: this would return a valid version for an invalid digest in a Pod that is pending.

/cc @rajansandeep 

should this block until the Pods are running?

neolit123: we also have a panic related to this area of the code:
https://github.com/kubernetes/kubeadm/issues/2267

rajansandeep: > should this block until the Pods are running?

Yes, I am leaning towards that. 
It seems this issue is caused due to either the coredns pod being in a `Pending` state(kubernetes/kubeadm#2267) or it being in a `ContainerCreating ` state (#94286).

Skipping the check if the coredns pod is not running might result in a false validation. (especially if all coredns pods are in not running state)

zouyee: If there is no running pod, then select the image name as the validated object and output a warning log?

neolit123: for the time being we should apply a waiting function to get all the Pods as "Running".

the alternative is to relax the version validation and only use the image version/tag.

rosti: It probably does make sense to have some wait if the pods are stuck in a `ContainerCreating` state. However, I am certain that this may not be the case with `PodPending`. If a pod is stuck in that state for more than a few seconds, then there's something elsewhere that needs tackling. In that case it's best to just return a proper error as we are clearly trying to perform this operation on a cluster that's in a bad state.

neolit123: we can have a 10seconds timeout with checks every 500ms?

@zouyee please adjust the PR description to "Fixes" this issue instead:
https://github.com/kubernetes/kubeadm/issues/2267

we should include a PollImmediate call (see example here):
https://github.com/kubernetes/kubernetes/blob/3a2e96efff7860ffb06eeba6d8603be53cb5125e/cmd/kubeadm/app/util/apiclient/idempotency.go#L64

that goes to the list of pods:
- checks if they running
- have a valid container
- store the image ID of the container in slice.

iterate over the retrieved imageIDs and set isValidVersion = false if one of them does not matches.

@zouyee do you want to work on this fix in this PR?

neolit123: @rajansandeep @rosti just to double check, our preflight has downloaded the new coredns image at this point, so we don't have to wait for the download?

zouyee: > @zouyee do you want to work on this fix in this PR?

yes

cmd/kubeadm/app/phases/addons/dns/dns.go
```diff
@@ -398,29 +401,46 @@ var (
...
+		if err != nil {
+			lastError = err
+			return false, nil
+		}
 
-	for _, pod := range coreDNSPodList.Items {
-		imageID := imageDigestMatcher.FindStringSubmatch(pod.Status.ContainerStatuses[0].ImageID)
-		if len(imageID) != 2 {
-			return false, errors.Errorf("unable to match SHA256 digest ID in %q", pod.Status.ContainerStatuses[0].ImageID)
+		if len(coreDNSPodList.Items) <= 0 {
```
chrisohaver, line 419: This would only ensure that at least one pod is running, there could be other coredns pods in other states.
So, we would only be checking versions on the running pods.

I dont know how important this distinction is.  Usually all of the pods would be running the same version... the exception being if there is a rollout in progress.

cmd/kubeadm/app/phases/addons/dns/dns.go
```diff
@@ -398,29 +401,46 @@ var (
...
-			LabelSelector: "k8s-app=kube-dns",
-		},
-	)
-	if err != nil {
-		return false, errors.Wrap(err, "unable to list CoreDNS pods")
-	}
+	passed := true
+	var lastError error
+	err := wait.PollImmediate(constants.APICallRetryInterval, 10 * time.Second, func() (bool, error) {
+		coreDNSPodList, err := client.CoreV1().Pods(metav1.NamespaceSystem).List(
```
neolit123, line 407: i think we can have the call to get the list outside of the poll.
the purpose of the poll would be to get the pods in a Running state and have containers where we can get the ImageIDs.

cmd/kubeadm/app/phases/addons/dns/dns.go
neolit123: i think we should just remove this check.

cmd/kubeadm/app/phases/addons/dns/dns.go
```diff
@@ -398,29 +401,46 @@ var (
...
-	coreDNSPodList, err := client.CoreV1().Pods(metav1.NamespaceSystem).List(
-		context.TODO(),
-		metav1.ListOptions{
-			LabelSelector: "k8s-app=kube-dns",
-		},
-	)
-	if err != nil {
-		return false, errors.Wrap(err, "unable to list CoreDNS pods")
-	}
+	passed := true
```
neolit123, line 404: this can be an integer.

on each poll loop it is reset to zero.
on each valid image ID it is incremented with 1.

after the poll, if passed == "total Pods", the validation has passed.

neolit123:
the deadline for cherry picks for 1.19.1 is this Friday.

cmd/kubeadm/app/phases/addons/dns/dns.go
zouyee: IMO, the poll function should be used to obtain a list of pods, and whether the list is empty as a loop condition.

cmd/kubeadm/app/phases/addons/dns/dns.go
zouyee: In fact, we don’t need to perform + 1, we only need to operate the flag once the conditions are not met, and finally return the flag

cmd/kubeadm/app/phases/addons/dns/dns.go
zouyee: > I dont know how important this distinction is. Usually all of the pods would be running the same version... the exception being if there is a deployment rollout in progress (with image change).

IMO, I agree with this point of view, but the original implementation is to check that there is, whether it needs to be modified?

neolit123:
./hack/verify-gofmt.sh seems to be failing.
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/94299/pull-kubernetes-verify/1301094623908728832
you can try ./hack/update-gofmt.sh

cmd/kubeadm/app/phases/addons/dns/dns.go
neolit123: i have never seem a LIST request return < 0 items. is that possible?

cmd/kubeadm/app/phases/addons/dns/dns.go
```diff
@@ -398,29 +401,47 @@ var (
...
+				FieldSelector: "status.phase=Running",
+			},
+		)
+		if err != nil {
+			lastError = err
+			return false, nil
+		}
+
+		if len(coreDNSPodList.Items) <= 0 {
+			lastError = errors.New("unable to list CoreDNS Running pods")
```
neolit123, line 420: this should be `errors.Errorf("could not find any Running CoreDNS pods")`

cmd/kubeadm/app/phases/addons/dns/dns.go
```diff
@@ -398,29 +401,47 @@ var (
...
+			return false, nil
+		}
+		pods = coreDNSPodList.Items
+		return true, nil
+	})
+
+	if err != nil || len(pods) <= 0 {
+		return false, fmt.Errorf("%v: %v", err, lastError)
```
neolit123, line 428: `errors.Wrapf(lastError, "could not find any Running CoreDNS pods after %v", pollTimeout)`

and `pollTimeout` can be set to `10 * time.Second` above the poll.

cmd/kubeadm/app/phases/addons/dns/dns.go
```diff
@@ -398,29 +401,47 @@ var (
...
 	}
 
-	for _, pod := range coreDNSPodList.Items {
+	isValidVersion := true
+	for _, pod := range pods {
 		imageID := imageDigestMatcher.FindStringSubmatch(pod.Status.ContainerStatuses[0].ImageID)
 		if len(imageID) != 2 {
-			return false, errors.Errorf("unable to match SHA256 digest ID in %q", pod.Status.ContainerStatuses[0].ImageID)
+			lastError = errors.Errorf("pod %s unable to match SHA256 digest ID in %q", pod.GetName() ,pod.Status.ContainerStatuses[0].ImageID)
+			return false, lastError
```
neolit123, line 436: just return without setting lastError?
```
return false, errors.Errorf("pod %s unable to match SHA256 digest ID in %q", pod.GetName() ,pod.Status.ContainerStatuses[0].ImageID)
```

cmd/kubeadm/app/phases/addons/dns/dns.go
```diff
@@ -398,29 +401,47 @@ var (
...
 		}
 		// The actual digest should be at imageID[1]
 		if !migration.Released(imageID[1]) {
 			isValidVersion = false
```
neolit123, line 440: i don't think isValidVersion and lastError are needed.

here the function can return:
`fasle, errors.Errorf("unknown digest %q for pod %s", imageID[1], pod.GetName())`

cmd/kubeadm/app/phases/addons/dns/dns.go
```diff
@@ -398,29 +401,47 @@ var (
...
 		}
 	}
 
-	return isValidVersion, nil
+	return isValidVersion, lastError
```
neolit123, line 444: if this point is reached it can return `true, nil`?

cmd/kubeadm/app/phases/addons/dns/dns.go
neolit123: ok, agreed.

cmd/kubeadm/app/phases/addons/dns/dns.go
chrisohaver: > i think we should just remove this check and assume the number of pods we go from the above LIST call is what we want to ImageID-check against.

The list call has a field selector for `status.phase=Running`. So, if a Pod is in a state other that "Running", it wont be in the list returned, i.e. we skip the check for all pending Pods.

neolit123: yes, i think the latest version of the PR where the list with the selector being inside the poll, should hopefully get all Pods as running and successfully get the digests from them.

chrisohaver: Oh OK, so the idea is to _only_ check versions of Running Pods, and not to check versions of any CoreDNS Pods in any other states. I was thinking we wanted to wait for _all_ pods to be non-pending before checking.

neolit123: actually, that makes more sense. IIRC @rajansandeep mentioned that with auto-scaling there could be ~100 of coredns pods in larger clusters, that should not be a problem, but not sure if the 10 second check will be sufficient for all cases.

@zouyee we could loop over all coredns pods and wait until they are Running (max=10 seconds). only then proceed with the checks.

cmd/kubeadm/app/phases/addons/dns/dns.go
```diff
@@ -398,29 +401,47 @@ var (
...
-		context.TODO(),
-		metav1.ListOptions{
-			LabelSelector: "k8s-app=kube-dns",
-		},
-	)
-	if err != nil {
-		return false, errors.Wrap(err, "unable to list CoreDNS pods")
+	var lastError error
+	var pods []v1.Pod
+	err := wait.PollImmediate(constants.APICallRetryInterval, 10 * time.Second, func() (bool, error) {
```
rajansandeep, line 406:
```suggestion
	err := wait.PollImmediate(kubeadmconstants.APICallRetryInterval, 10 * time.Second, func() (bool, error) {
```

rajansandeep (commented):
Few nits

cmd/kubeadm/app/phases/addons/dns/dns.go
```diff
@@ -34,10 +35,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kuberuntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 	clientsetscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/klog/v2"
 	kubeadmapi "k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm"
+	"k8s.io/kubernetes/cmd/kubeadm/app/constants"
 	kubeadmconstants "k8s.io/kubernetes/cmd/kubeadm/app/constants"
```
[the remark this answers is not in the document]
zouyee: Thank you for your reminder，it has been fixed.

cmd/kubeadm/app/phases/addons/dns/dns.go
zouyee: Done

cmd/kubeadm/app/phases/addons/dns/dns.go
```diff
@@ -398,29 +400,56 @@ var (
...
+				FieldSelector: "status.phase!=Running",
+			},
+		)
+		if err != nil {
+			lastError = err
+			return false, nil
+		}
+
+		if len(coreDNSPodList.Items) > 0 {
+			lastError = errors.New("find non-running CoreDNS pods")
```
neolit123, line 419: should be:
> found non-running CoreDNS pods

cmd/kubeadm/app/phases/addons/dns/dns.go
```diff
@@ -398,29 +400,56 @@ var (
...
+			return false, nil
+		}
+
+		return true, nil
+	})
+
+	if err != nil {
+		return false, errors.Wrapf(lastError, "after %v", pollTimeout)
```
neolit123, line 427:
```suggestion
		return false, errors.Wrapf(lastError, "could not list the running CoreDNS pods after %v", pollTimeout)
```

cmd/kubeadm/app/phases/addons/dns/dns.go
```diff
@@ -398,29 +400,56 @@ var (
...
+	}
+
 	coreDNSPodList, err := client.CoreV1().Pods(metav1.NamespaceSystem).List(
 		context.TODO(),
 		metav1.ListOptions{
 			LabelSelector: "k8s-app=kube-dns",
+			FieldSelector: "status.phase=Running",
```
neolit123, line 434: list operations are expensive if there are a lot of instances...
why not instead of using a FieldSelector LIST all pods and return an error (continue the poll) if one of them is not Running?

neolit123 (commented):
/lgtm
/approve
/milestone v1.20-phase-bug

@zouyee thanks. this seems good to me.
we also have to cherry pick this to `release-1.19`, please let me know if you'd like to send that:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-release/cherry-picks.md#initiate-a-cherry-pick

zouyee:
@neolit123 PTAL

rajansandeep (approved):
This looks good, thank you @zouyee!
/lgtm

neolit123:
@zouyee please update the PR description to have the release note:
> kubeadm: avoid a panic when determining if the running version of CoreDNS is supported during upgrades

instead of `NONE`

neolit123:
@zouyee 

for some reason this broken kubeadm init...

```
I0904 03:44:26.329986   18746 request.go:581] Throttling request took 181.523909ms, request: PUT:https://192.168.0.102:6443/apis/rbac.authorization.k8s.io/v1/namespaces/kube-public/rolebindings/kubeadm:bootstrap-signer-clusterinfo?timeout=10s
I0904 03:44:26.340869   18746 kubeletfinalize.go:88] [kubelet-finalize] Assuming that kubelet client certificate rotation is enabled: found "/var/lib/kubelet/pki/kubelet-client-current.pem"
[kubelet-finalize] Updating "/etc/kubernetes/kubelet.conf" to point to a rotatable kubelet client certificate and key
I0904 03:44:26.343756   18746 kubeletfinalize.go:132] [kubelet-finalize] Restarting the kubelet to enable client certificate rotation
found non-running CoreDNS pods
k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns.isCoreDNSVersionSupported.func1
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns/dns.go:421
k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait.runConditionWithCrashProtection
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:211
k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait.WaitFor
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:541
k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait.pollInternal
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:427
k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait.pollImmediateInternal
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:452
k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait.PollImmediate
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:441
k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns.isCoreDNSVersionSupported
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns/dns.go:407
k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns.createCoreDNSAddon
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns/dns.go:273
k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns.coreDNSAddon
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns/dns.go:254
k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns.EnsureDNSAddon
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns/dns.go:111
k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/init.runCoreDNSAddon
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/init/addons.go:92
k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow.(*Runner).Run.func1
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow/runner.go:234
k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow.(*Runner).visitAll
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow/runner.go:422
k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow.(*Runner).Run
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow/runner.go:207
k8s.io/kubernetes/cmd/kubeadm/app/cmd.NewCmdInit.func1
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/cmd/kubeadm/app/cmd/init.go:153
k8s.io/kubernetes/vendor/github.com/spf13/cobra.(*Command).execute
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/github.com/spf13/cobra/command.go:842
k8s.io/kubernetes/vendor/github.com/spf13/cobra.(*Command).ExecuteC
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/github.com/spf13/cobra/command.go:950
k8s.io/kubernetes/vendor/github.com/spf13/cobra.(*Command).Execute
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/github.com/spf13/cobra/command.go:887
k8s.io/kubernetes/cmd/kubeadm/app.Run
	/home/lubo-it/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/cmd/kubeadm/app/kubeadm.go:50
main.main
	_output/local/go/src/k8s.io/kubernetes/cmd/kubeadm/kubeadm.go:25
```

it applies the coredns addon, but we should definitely be waiting for the coredns pods to be running on "init".

neolit123:
the problem here is the following:
- start "kubeadm init" and wait for it to finish
- start "kubeadm init" again
- kubeadm should print some preflight messages that certificates / paths already exists
- instead of finishing, if fails at the coredns step because it finds the coredns pods, but cni is not yet applied

calling init after init is obviously not supported, but this change to wait for running pods breaks idempotency principles around the addon phase in case the CNI is not installed yet...its not clear to me how the original code worked.

kubeadm shares the same addon logic for "init" and "upgrade" and while this should work fine during "upgrade" given a running cluster, during "init" this is problematic.

i'm going to send a PR to remove the isCoreDNSVersionSupported function and rely on the coredns migration to throw errors if needed. the requirement for validating the digests of images is way to strict and has caused these problems.

rajansandeep:
> (during init this is the culprit: https://github.com/kubernetes/kubernetes/compare/048e32c9b6be1fc10e1315f12d6fc297b32335b2..a04bf278a2468fa07a0a3a793ce5f8678096ceb1)

@neolit123  I'm not sure this is part of the final code

rajansandeep:
> calling init after init is obviously not supported, but this change to wait for running pods breaks idempotency principles around the addon phase in case the CNI is not installed yet...its not clear to me how the original code worked.

Previously,  a node not having any CNI installed would result in the node being `NotReady`, and hence running `upgrade` on the node was not possible.
I'm not sure when this behavior was changed, but since now the node is `Ready` even without any CNI plugin, the `upgrade` proceeds (This means the CoreDNS pods would remain in `ContainerCreating` in this case)

neolit123:
i guess given the the message "found non-running CoreDNS pods" the problem is here:
https://github.com/kubernetes/kubernetes/pull/94299/files#diff-46182be33a6e1280b49fb67c2d22cc58R421

neolit123:
@rajansandeep 
please have a look at this PR:
https://github.com/kubernetes/kubernetes/pull/94506

> I'm not sure when this behavior was changed, but since now the node is Ready even without any CNI plugin, the upgrade proceeds (This means the CoreDNS pods would remain in ContainerCreating in this case)

odd, i noticed something similar but don't recall when this changed.

neolit123:
>> I'm not sure when this behavior was changed, but since now the node is Ready even without any CNI plugin, the upgrade proceeds (This means the CoreDNS pods would remain in ContainerCreating in this case)

> odd, i noticed something similar but don't recall when this changed.

actually i just checked this and the Node remains NotReady for me without CNI and the CoreDNS are Pending.
the kubelet i'm testing with is at 1.18.x though.

PR #94302, opened 2020-08-28
Adding system cluster-role for default admissions in aggregated-apiservers

yue9944882:
```release-note
NONE
```

/sig auth

when adding a new extension-apiserver to a cluster, it will keep crashing unless i disable some default admission plugins by:

>  --disable-admission-plugins=MutatingAdmissionWebhook,ValidatingAdmissionWebhook,NamespaceLifecycle

adding corresponding cluster-roles so that we can grant the permission by simply adding a rolebinding..

### DIFF
--- a/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go
+++ b/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go
@@ -42,24 +42,25 @@ var (
 )
 
 const (
-	legacyGroup            = ""
-	appsGroup              = "apps"
-	authenticationGroup    = "authentication.k8s.io"
-	authorizationGroup     = "authorization.k8s.io"
-	autoscalingGroup       = "autoscaling"
-	batchGroup             = "batch"
-	certificatesGroup      = "certificates.k8s.io"
-	coordinationGroup      = "coordination.k8s.io"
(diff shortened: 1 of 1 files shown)


plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go
```diff
@@ -391,6 +392,20 @@ func ClusterRoles() []rbacv1.ClusterRole {
...
+			ObjectMeta: metav1.ObjectMeta{Name: "extension-apiserver-generic-admission-namespacelifecycle"},
+			Rules: []rbacv1.PolicyRule{
+				rbacv1helpers.NewRule("list", "watch").Groups(legacyGroup).Resources("namespaces").RuleOrDie(),
+			},
+		},
+		{
+			// a role to use for allowing extension-apiserver's admission-webhook plugin to list-watch webhook-configurations
+			ObjectMeta: metav1.ObjectMeta{Name: "extension-apiserver-generic-admission-webhooks"},
+			Rules: []rbacv1.PolicyRule{
+				rbacv1helpers.NewRule("list", "watch").Groups(admissionRegistrationGroup).Resources("*").RuleOrDie(),
```
enj, line 406: Do not use `*`.

PR #94303, opened 2020-08-28
[WIP] Introduce max limit of objects fetched from etcd in a single list call.

mborsz:
**What type of PR is this?**
/kind feature

It tries to address the following problem:
1. We have a cluster with huge number of objects of given type (e.g. 500k events).
2. A client does LIST call like `GET /api/v1/events?limit=500&fieldSelector=field=AAA
3. If less than 500 (limit) events matches a field selector, we will internally try to fetch all 500k events in batches of 500 elements. In our experiment setup, each such batch takes ~200-300ms, so we can fetch approx. 120k objects before we reach 1m request timeout.
4. If we have more than 120k objects and only few of them matches a selector, the client is not able to fetch those objects as the request will timeout before their get response.

**Proposal:** Splitting scanning phase into hunks of X objects (let's say 50k). We can say that single paginated list request will be interrupted after fetching X objects from etcd (even if we haven't found requested number of elements) and partial result with continuation token will be returned. This way, the client can use this token to scan the next X elements in the next request. Each single list request will take at most (50k/limit) requests which for limit = 500 will take approx. 25-30s.

The drawback is that we change behavior that we always return exactly "limit" elements if we return continue token, but per my understanding of official documentation we never guaranteed anything like that.

This is only a partial solution, if client specifies e.g. limit = 1, we will still do a huge number of range calls to etcd and time out anyway.
 
```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
+++ b/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
@@ -47,6 +47,11 @@ import (
 	utiltrace "k8s.io/utils/trace"
 )
 
+const (
+	// maxKeysFetchedInSingleList is a maximum number of elements fetched from etcd in a single list call.
+	maxKeysFetchedInSingleList = 50000
+)
+
 // authenticatedDataString satisfies the value.Context interface. It uses the key to
 // authenticate the stored data. This does not defend against reuse of previously
 // encrypted values under the same key, but will prevent an attacker from using an
@@ -642,6 +647,7 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions, 
 	var lastKey []byte
 	var hasMore bool
 	var getResp *clientv3.GetResponse
+	var allKeysCount int
 	for {
 		startTime := time.Now()
 		getResp, err = s.client.KV.Get(ctx, key, options...)
@@ -683,6 +689,7 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions, 
 				return err
 			}
 		}
+		allKeysCount += len(getResp.Kvs)
 
 		// indicate to the client which resource version was returned
 		if returnedRV == 0 {
@@ -695,6 +702,12 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions, 
 		}
 		// we're paging but we have filled our bucket
 		if int64(v.Len()) >= pred.Limit {
+			break
+		}
+		// There is still place in our bucket, but we have iterated too many elements not matching pred.
+		// To avoid returning timeout before we fill out the bucket, interrupt after maxKeysFetchedInSingleList elements
+		// and return continuation to resume scanning from the current position in the next request.
+		if allKeysCount >= maxKeysFetchedInSingleList {
 			break
 		}
 		key = string(lastKey) + "\x00"


mborsz:
/cc @wojtek-t 
/assign @liggitt 

Jordan, I would like to hear your opinion on the proposal described in the PR's description.

The PR contains some simple implementation of this proposal, but we can figure out if it's correct once we agree that the high level idea makes sense.

staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
```diff
@@ -697,6 +704,12 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions,
 		if int64(v.Len()) >= pred.Limit {
 			break
 		}
+		// There is still place in our bucket, but we have iterated too many elements not matching pred.
+		// To avoid returning timeout before we fill out the bucket, interrupt after maxKeysFetchedInSingleList elements
+		// and return continuation to resume scanning from the current position in the next request.
+		if allKeysCount >= maxKeysFetchedInSingleList {
+			break
```
wojtek-t, line 711: @liggitt - I've discussed this with Maciek and I'm not sure what to do with it.
Basically, with this change we can return a list with less than <limit> items together with continuation token. On one hand it's not a documented contract (so that's fine), otoh I'm wondering if someone doesn't rely on it (client-go doesn't seem to, but that's only a single example).

[BTW - if we agree it's fine, it might make sense to break it based on time (e.g. break after 50s or sth like that to no hit 1m default timeout), instead of breaking based on key count. But let's wait for Jordan`s (and others?) opinions whether we actually want to go this way before changing anything.]

@liggitt - WDYT?

liggitt:
/assign @smarterclayton 
who originated the internal-paging-to-fill-filtered-paged-response loop

mborsz:
@smarterclayton, what is your opinion on this proposal?

mborsz:
@smarterclayton Friendly ping

staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
smarterclayton: I think this changes the contract in a way that needs sufficient review and discussion that it should be a PR to the KEP after we have talked it out here, no matter what.

Forming some thoughts

smarterclayton:
> The drawback is that we change behavior that we always return exactly "limit" elements if we return continue token, but per my understanding of official documentation we never guaranteed anything like that.

It is probably a behavior change that would break some clients, so we have to be cautious about such a change regardless of whether we specified it.  We would want to review a sufficiently broad set of clients.

However, it's not clear to me that we can't solve the smaller part of the problem - better internal paging strategies (below).

> Splitting scanning phase into hunks of X objects (let's say 50k). We can say that single paginated list request will be interrupted after fetching X objects from etcd (even if we haven't found requested number of elements) and partial result with continuation token will be returned. This way, the client can use this token to scan the next X elements in the next request. Each single list request will take at most (50k/limit) requests which for limit = 500 will take approx. 25-30s.

This is a more complicated way of saying "the paging strategy should take predicates into account" :)  I think in general the presence of predicates would justify larger internal page sizes.  Probably even a crude selectivity estimation would allow us to pick a better page size, and if truly necessary we could provide that selectivity hint to subsequent pages.  Nothing about the internal pager requires a constant stride (we could start at limit, then for instance exponentially double to approximate selectivity, then provide the selectivity approximation in the continue token).

I would probably suggest we explore being slightly more clever on the backend (in a way that works for a broad range of scenarios) before we change the limit semantic.

Is that exploration (thinking through paging strategies that might be more effective) something you are comfortable trying here?

jpbetz:
I've been trying to think of any practical problems this might cause for clients and haven't come up with anything. This seems like a reasonable approach. It should only be triggered in high scale use cases, where the current alternative would be a timeout. The API has never guaranteed to clients that the limit is respected, so any misbehaved client that is checking the limit and not the continue token to determine if there are additional pages to fetch would already be in trouble.

There are been some previous discussions in the etcd community, and in api-machinery, about enabling etcd to index secondary fields somehow, so it can do the filtering directly. I had written up [Design proposal: etcd Symlinks](https://docs.google.com/document/d/15YszUyr0StaIB6UO63tXhpaKw3UkyM_NlTvrSzMth9A/edit#heading=h.tvw5pkbaoe47) but we never finalized on a design or resourced it. (cc @ptabor, @jingyih).

It needs to be documented carefully on https://kubernetes.io/docs/reference/using-api/api-concepts/#retrieving-large-results-sets-in-chunks.

wojtek-t:
> I've been trying to think of any practical problems this might cause for clients and haven't come up with anything.

Yeah - I've spent some time thinking about it and didn't come up with any too.
@smarterclayton - is your fear about users checking if <limit> items were returned instead of looking at the existence of the continue token? If so, that sounds like asking for trouble anyway.

That said:

> (we could start at limit, then for instance exponentially double to approximate selectivity, then provide the selectivity approximation in the continue token).

I actually really like this path. This may not solve all the potential cases, but it may solve majority of those. That also won't change the contract so it's for sure less controversial, and shouldn't be more than couple lines of code actually (starting at 500, doubling if we didn't fill the list, capping at e.g. 32*500 = 16k). 

We may consider more sophisticated strategies, but I'm actually not convinced they will buy us much more, so I'm a bit sceptical about spending too much time on it.
@smarterclayton - WDYT?

mborsz:
Thanks for all comments. I agree that experimenting with better internal etcd paging strategies seems to be a good starting point (it's a less controversial than changing a pagination contract I suggested here). I will try to play with it a little bit and once I have some results I will create some PR.

ptabor:
With the unexpectedly huge gain from splitting big etcd transactions (leases) into smaller (1000 objects) in https://github.com/kubernetes/kubernetes/pull/98257, I think it's important to explore splitting the big range scans.

From etcd perspective I expect: 
  - lower RAM consumption / possibility of OOM. 
  - lower tail latency.  
  - (so) possibility to define stricter SLOs. E.g. reads>10000 entries might get excluded from SLO. 
  - lower fragmentation of the DB (shorter bolt transactions)

KnVerey:
@mborsz are you still working on this?

mborsz:
> Maciej Borsz are you still working on this?

I'm not actively working on this. I think it's still useful to do, but I didn't have enough time to work on this.

ptabor:
Long-term: If etcd exposed  streaming-style API for lists, would that be a good match for API-server ? 
The benefit of streaming API are following: 
    - piggy back to grpc-layer flow-control 
    - Limit of 1000 for 1 MB object is still 1GB of payload... with streaming API we would have 'fine grained&' control.

dims:
please reopen if you still need this WIP PR. Trying to declutter the open PR(s)

PR #88119, opened 2020-02-13
Prefer apps version over deprecated extensions

soltysh:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
Due to backwards compatibility issues (https://github.com/kubernetes/kubernetes/issues/42392)
introduced in 1.6 when adding new API group we've set extensions to be preferred group over apps
(https://github.com/kubernetes/kubernetes/pull/43553/).

Now that extensions is entirely deprecated (https://github.com/kubernetes/kubernetes/issues/43214) it's time to revert this change and prefer correct API groups.

**Special notes for your reviewer**:
/assign @deads2k @liggitt 

**Does this PR introduce a user-facing change?**:
```release-note
Prefer apps over extensions in discovery
```

PR #94309, opened 2020-08-28
Deprioritize extensions/v1beta1 in discovery

liggitt:
**What type of PR is this?**

/kind cleanup

Ingress is the only resource remaining in extensions/v1beta1, has been deprecated since 1.14, has a target removal of 1.22, and has a GA replacement in networking.k8s.io/v1 as of 1.19.

Deprioritize the extensions/v1beta1 API group so that an unqualified `kubectl get ingresses` prefers the GA version.

```release-note
`kubectl get ingress` now prefers the `networking.k8s.io/v1` over `extensions/v1beta1` (deprecated since v1.14). To explicitly request the deprecated version, use `kubectl get ingress.v1beta1.extensions`.
```

/sig cli
/sig networking

/cc @soltysh 
similar to https://github.com/kubernetes/kubernetes/pull/88119; I'm more convinced now that we have a GA alternative and that `kubectl get ingress` would return a deprecation warning.

/cc @robscott 
for networking / ingress visibility

### DIFF
--- a/cmd/kube-apiserver/app/aggregator.go
+++ b/cmd/kube-apiserver/app/aggregator.go
@@ -234,9 +234,6 @@ type priority struct {
 // That ripples out every bit as far as you'd expect, so for 1.7 we'll include the list here instead of being built up during storage.
 var apiVersionPriorities = map[schema.GroupVersion]priority{
 	{Group: "", Version: "v1"}: {group: 18000, version: 1},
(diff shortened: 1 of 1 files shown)


soltysh (approved):
/lgtm
/milestone v1.20
/hold
for visibility

robscott:
Thanks!

/lgtm

PR #94310, opened 2020-08-28
Moving SetHostnameAsFQDn e2e tests to e2e_node directory to include tests in Node Testgrid

javidiaz:
I still do not see these tests in the node alpha feature CI. Looking more into it I just realized that we are not supposed to add tests to the test/e2e directory. Moving the tests to e2e_node. 
/cleanup
/area test

### DIFF
--- a/test/e2e/node/BUILD
+++ b/test/e2e/node/BUILD
@@ -12,7 +12,6 @@ go_library(
         "mount_propagation.go",
         "node_problem_detector.go",
         "pod_gc.go",
-        "pod_hostnamefqdn.go",
         "pods.go",
         "pre_stop.go",
         "recreate_node.go",
--- a/test/e2e_node/BUILD
+++ b/test/e2e_node/BUILD
@@ -132,6 +132,7 @@ go_test(
         "node_container_manager_test.go",
         "node_perf_test.go",
         "pids_test.go",
+        "pod_hostnamefqdn_test.go",
         "pods_container_manager_test.go",
         "quota_lsci_test.go",
         "resource_metrics_test.go",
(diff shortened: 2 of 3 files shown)


spiffxp (approved):
/ok-to-test
/approve

alejandrox1:
@javidiaz where in testgrid do you want these tests to show up (what tab exactly https://testgrid.k8s.io/sig-node-kubelet#node-kubelet-alpha ? or another one)?

Also, how do you want to run these tests
1. on a full kubernetes cluster
2. on a kubelet (with an api server and a namespace controller)
?

If you put tests in test/e2e_node then you will be running them in scenario 2.

javidiaz:
> @javidiaz where in testgrid do you want these tests to show up (what tab exactly https://testgrid.k8s.io/sig-node-kubelet#node-kubelet-alpha ? or another one)?
> 
> Also, how do you want to run these tests
> 
>     1. on a full kubernetes cluster
> 
>     2. on a kubelet (with an api server and a namespace controller)
>        ?
> 
> 
> If you put tests in test/e2e_node then you will be running them in scenario 2.

Thanks @alejandrox1. Does scenario 2 mean all k8s control plane components run together, versus scenario 1 you run your tests on a proper/large? cluster? If that is the case I think scenario 2 is enough. The functionality these tests are checking is in the kubelet.

alejandrox1:
Scneario 2 (all tests in test/e2e_node) run on top of a kubelet with an API server, a namespace controller, and an etcd (no other components).
Just making sure that that is the environment you are looking for.

javidiaz:
> Scneario 2 (all tests in test/e2e_node) run on top of a kubelet with an API server, a namespace controller, and an etcd (no other components).
> Just making sure that that is the environment you are looking for.

Thanks for clarifying, this is good to know. I think these tests should work well in scenario 2.

PR #94235, opened 2020-08-25
Add exponential backoff for connection refused errors

kevindelgado:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
This PR adds an exponential backoff when ListAndWatch() receives a connection refused error.

Currently when ListAndWatch() receives a connection refused error, it is
assumed to be due to the apiserver being transiently unresponsive. 

In situations where a controller is running outside the k8s cluster it's
controlling, it is more common for the controller to lose connection
permanently to the apiserver and needs to exponentially backoff its
retry rather than continuously spamming logs with Watch attempts that
will never succeed.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #94317, opened 2020-08-28
Fix FakeClock::Reset to always succeed

kevindelgado:
**What type of PR is this?**
/kind bug

Currently, FakeClock::Reset only successfully resets the timer and returns true when the timer has neither fired nor been stopped.

This is incorrect behavior as real clocks allow for resetting in both of those situations (as long has the channel has been drained).

This is useful in tests that use a fake clock to test wait.BackofManager, as the timer resets after each iteration of Backoff() after the timer has already fired.

```release-note
NONE
```

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/util/clock/clock.go
+++ b/staging/src/k8s.io/apimachinery/pkg/util/clock/clock.go
@@ -348,7 +348,13 @@ func (f *fakeTimer) Stop() bool {
 // Reset conditionally updates the firing time of the timer.  If the
 // timer has neither fired nor been stopped then this call resets the
 // timer to the fake clock's "now" + d and returns true, otherwise
-// this call returns false.  This is like time.Timer::Reset.
+// it creates a new waiter, adds it to the clock, and returns true.
+//
+// It is not possible to return false, because a fake timer can be reset
+// from any state (waiting to fire, already fired, and stopped).
+//
+// See the GoDoc for time.Timer::Reset for more context on why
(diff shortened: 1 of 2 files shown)


kevindelgado:
This PR came about while trying to write the unit test for https://github.com/kubernetes/kubernetes/pull/94235 using a fake clock instead of a real clock. The way the backoff manager resets its timer after successive iterations exposed the fact that the current fake clock implementation does not allow for resetting fake timers that have already fired.

kevindelgado:
One outstanding question I have is whether this conflicts with the behavior previously fixed by https://github.com/kubernetes/kubernetes/commit/3e898d69210918433fd0cd0aa89392faac61b83a#diff-2feab94c4d7d3f50c86be9853504ace6L313-L319. It seems to me that previously Reset() would reset the timer if already fired (while returning false). 

I’m unclear on the context of that change and if this is reverting something that was intentionally modified (albeit, it was modified to a behavior misaligned with the behavior of real clock).

lavalamp:
/ok-to-test

Let's see if this breaks anything :)

kevindelgado:
looks like it passed, PTAL @lavalamp

lavalamp:
/lgtm
/approve
/milestone 1.20

thanks a ton for fixing this! I don't think this should break e2e tests, only unit tests. So should be good to go.

lavalamp:
/milestone v1.20

Missing the v, I guess that's v for "very sure I want 1.20".

justaugustus:

/retest

PR #94333, opened 2020-08-29
fix(validation): ensure the MilliValue of storage resource could be divided by 1000

### LINKED ISSUE #94313: PVCs specified with quantity containing suffix `m` are not counted in ResourceQuota and block ResourceQuota updates
**What happened**:
When a pvc was specified with a quantity including the suffix `m`:
```yaml
...
spec:
    accessModes:
    - ReadWriteOnce
    resources:
      requests:
        storage: 100m
...
```

attempts to include that quantity in the resourcequota's storage  resulted in the following error, which blocked all other updates to the resourcequota:

```
E0827 23:17:08.653750 1744486 resource_quota_controller.go:251] ResourceQuota "namespaced-quota" is invalid: status.used[scaleio.storageclass.storage.k8s.io/requests.storage]: Invalid value: resource.Quantity{i:resource.int64Amount{value:518617300992100, scale:-3}, d:resource.infDecAmount{Dec:(*inf.Dec)(nil)}, s:"518617300992100m", Format:"DecimalSI"}: must be an integer
```

- Updates to all other quantities in the resourcequota should not have been blocked by the error above
- In general, `m` as a suffix (milli) doesn't really make sense when used in the context of memory, and causes issues like the above. Would it be possible to deny the usage of this suffix at admission time for things like pvcs? (Our mitigation right now is to add a rule that does this in Open Policy Agent).

Create a resourcequota object with limits on pvcs and the a storage type, e.g.
```
...
  spec:
    hard:
      count/persistentvolumeclaims: "100"
      scaleio.storageclass.storage.k8s.io/requests.storage: 3Ti
...
```

Then create a PVC with the following spec:
```
spec:
    accessModes:
    - ReadWriteOnce
    resources:
      requests:
        storage: 100m
```

The resource quota controller should start producing the error log above.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.17.3
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

knight42:

**What type of PR is this?**

/kind bug

Fixes #94313

```release-note
PVC: make sure the value of `requests.storage` is multiple of bytes
```

```docs

```

### DIFF
--- a/pkg/apis/core/validation/validation.go
+++ b/pkg/apis/core/validation/validation.go
@@ -1978,7 +1978,7 @@ func ValidatePersistentVolumeClaimSpec(spec *core.PersistentVolumeClaimSpec, fld
 	} else if errs := ValidatePositiveQuantityValue(storageValue, fldPath.Child("resources").Key(string(core.ResourceStorage))); len(errs) > 0 {
 		allErrs = append(allErrs, errs...)
 	} else {
-		allErrs = append(allErrs, ValidateResourceQuantityValue(string(core.ResourceStorage), storageValue, fldPath.Child("resources").Key(string(core.ResourceStorage)))...)
+		allErrs = append(allErrs, ValidateStorageQuantityValue(string(core.ResourceStorage), storageValue, fldPath.Child("resources").Key(string(core.ResourceStorage)))...)
 	}
 
 	if spec.StorageClassName != nil && len(*spec.StorageClassName) > 0 {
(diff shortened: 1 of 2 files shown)


pkg/apis/core/validation/validation.go
```diff
@@ -1978,7 +1978,7 @@ func ValidatePersistentVolumeClaimSpec(spec *core.PersistentVolumeClaimSpec, fld
 	} else if errs := ValidatePositiveQuantityValue(storageValue, fldPath.Child("resources").Key(string(core.ResourceStorage))); len(errs) > 0 {
 		allErrs = append(allErrs, errs...)
 	} else {
-		allErrs = append(allErrs, ValidateResourceQuantityValue(string(core.ResourceStorage), storageValue, fldPath.Child("resources").Key(string(core.ResourceStorage)))...)
+		allErrs = append(allErrs, ValidateStorageQuantityValue(string(core.ResourceStorage), storageValue, fldPath.Child("resources").Key(string(core.ResourceStorage)))...)
```
liggitt, line 1981: /hold

standard questions when we want to tighten validation:
* does it prevent a no-op update of existing persisted objects that would fail this new validation? (would block storage migration tasks)
* does it prevent deletion of existing persisted objects with finalizers that would fail this new validation? (update is required to remove finalizers to allow deletion)
* does it prevent controllers from updating status of existing persisted objects that would fail this new validation?

we could consider ratcheting validation:
* apply this tightened validation on create
* apply this tightened validation on update where the old object passed this validation (don't let a valid object become invalid)
* skip this tightened validation on update where the old object fails this validation

c.f. https://github.com/kubernetes/kubernetes/issues/59984

knight42: I think the only case that the `MilliValute` of `requests.storage` could not by divided by 1000 is user set the value to something like `1234m` which means `1.234` byte which doesn't make any sense in real world. At least the suffix is valid to describe the size, such as 'Ki', 'M', 'Mi' etc., then we are good.

But one can never be too cautious, I agree it is possible to break the API if we tighten the validation. If I try ratcheting validation as you said, is this good to have? If not, I think we could leave the current logic intact.

liggitt: opened https://github.com/kubernetes/kubernetes/pull/94620 to fix the issue with the quota evaluator to only record whole byte values

will separately track warning about this issue in PVC creation

liggitt:
/close
in favor of https://github.com/kubernetes/kubernetes/pull/94620

PR #94334, opened 2020-08-29
Add e2e test to validate performance metrics of volume lifecycle operations

RaunakShah:
**What type of PR is this?**
/kind feature
(Not sure if a test counts as a feature)

This test validates latency and throughput of volume
provisioning against a high baseline.

Test Steps:
1. Create a Storage Class
2. Set up an informer on PVCs
3. Create 500 PVCs
- Store create time for each PVC
4. Wait for all PVCs to enter Bound state
- Use informer to store time when every PVC enters Bound state. 
5. End wait if:
- All PVCs enter Bound state (or)
- Test times out
5. Calculate performance metrics based on above times
6. Validate performance metrics against baseline.

Future work: Add more metrics and calculate performance metrics for other types of operations. 

Follow up to comments from @msau42 based on some performance improvements we've made to external sidecars.

Baselines are set to quite high values. Not sure what we should keep it as right now..

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/storage/drivers/csi.go
+++ b/test/e2e/storage/drivers/csi.go
@@ -109,6 +109,18 @@ func initHostPathCSIDriver(name string, capabilities map[storageframework.Capabi
 				NumPods:      10,
 				NumSnapshots: 10,
 			},
+			PerformanceTestOptions: &storageframework.PerformanceTestOptions{
+				ProvisioningOptions: &storageframework.PerformanceTestProvisioningOptions{
+					VolumeSize: "1Mi",
+					Count:      300,
+					// Volume provisioning metrics are compared to a high baseline.
+					// Failure to pass would suggest a performance regression.
+					ExpectedMetrics: &storageframework.Metrics{
+						AvgLatency: 2 * time.Minute,
+						Throughput: 0.5,
+					},
+				},
+			},
 		},
 		manifests:        manifests,
 		volumeAttributes: volumeAttributes,
--- a/test/e2e/storage/framework/testdriver.go
+++ b/test/e2e/storage/framework/testdriver.go
@@ -17,12 +17,13 @@ limitations under the License.
 package framework
 
 import (
-	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2evolume "k8s.io/kubernetes/test/e2e/framework/volume"
+	"time"
 )
 
 // TestDriver represents an interface for a driver to be tested in TestSuite.
@@ -205,6 +206,8 @@ type DriverInfo struct {
 	StressTestOptions *StressTestOptions
 	// [Optional] Scale parameters for volume snapshot stress tests.
 	VolumeSnapshotStressTestOptions *VolumeSnapshotStressTestOptions
+	// [Optional] Parameters for performance tests
+	PerformanceTestOptions *PerformanceTestOptions
 }
 
 // StressTestOptions contains parameters used for stress tests.
@@ -223,4 +226,25 @@ type VolumeSnapshotStressTestOptions struct {
 	NumPods int
 	// Number of snapshots to create for each volume.
 	NumSnapshots int
+}
+
+// Metrics to evaluate performance of an operation
+// TODO: Add metrics like median, mode, standard deviation, percentile
+type Metrics struct {
+	AvgLatency time.Duration
+	Throughput float64
+}
+
+// PerformanceTestProvisioningOptions contains parameters for
+// testing provisioning operation performance.
+type PerformanceTestProvisioningOptions struct {
+	VolumeSize string
+	Count      int
+	// Expected metrics from PVC creation till PVC being Bound.
+	ExpectedMetrics *Metrics
+}
+
+// PerformanceTestOptions contains parameters used for performance tests
+type PerformanceTestOptions struct {
+	ProvisioningOptions *PerformanceTestProvisioningOptions
 }
--- a/test/e2e/storage/testsuites/volumeperf.go
+++ b/test/e2e/storage/testsuites/volumeperf.go
@@ -1,0 +1,312 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package testsuites
+
+import (
+	"context"
+	"fmt"
+	"github.com/davecgh/go-spew/spew"
+	"sync"
+	"time"
+
+	"github.com/onsi/ginkgo"
+
+	"k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/watch"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/kubernetes/test/e2e/framework"
+	e2epv "k8s.io/kubernetes/test/e2e/framework/pv"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+	storageframework "k8s.io/kubernetes/test/e2e/storage/framework"
+)
+
+type volumePerformanceTestSuite struct {
+	tsInfo storageframework.TestSuiteInfo
+}
+
+var _ storageframework.TestSuite = &volumePerformanceTestSuite{}
+
+const testTimeout = 15 * time.Minute
+
+// Time intervals when a resource was created, it enters
+// the desired state and the elapsed time between these
+// two states
+type interval struct {
+	create            time.Time
+	enterDesiredState time.Time
+	elapsed           time.Duration
+}
+
+// Consolidates performance stats for any operation
+type performanceStats struct {
+	mutex             *sync.Mutex
+	perObjectInterval map[string]*interval
+	operationMetrics  *storageframework.Metrics
+}
+
+// waitForProvisionCh receives a signal from controller
+// when all PVCs are Bound
+// The signal received on this channel is the list of
+// PVC objects that are created in the test
+// The test blocks until waitForProvisionCh receives a signal
+// or the test times out
+var waitForProvisionCh chan []*v1.PersistentVolumeClaim
+
+// InitVolumePerformanceTestSuite returns volumePerformanceTestSuite that implements TestSuite interface
+func InitVolumePerformanceTestSuite() storageframework.TestSuite {
+	return &volumePerformanceTestSuite{
+		tsInfo: storageframework.TestSuiteInfo{
+			Name: "volume-lifecycle-performance",
+			TestPatterns: []storageframework.TestPattern{
+				storageframework.FsVolModeDynamicPV,
+			},
+		},
+	}
+}
+
+func (t *volumePerformanceTestSuite) GetTestSuiteInfo() storageframework.TestSuiteInfo {
+	return t.tsInfo
+}
+
+func (t *volumePerformanceTestSuite) SkipUnsupportedTests(driver storageframework.TestDriver, pattern storageframework.TestPattern) {
+}
+
+func (t *volumePerformanceTestSuite) SkipRedundantSuite(driver storageframework.TestDriver, pattern storageframework.TestPattern) {
+}
+
+func (t *volumePerformanceTestSuite) DefineTests(driver storageframework.TestDriver, pattern storageframework.TestPattern) {
+	type local struct {
+		config      *storageframework.PerTestConfig
+		testCleanup func()
+		cs          clientset.Interface
+		ns          *v1.Namespace
+		scName      string
+		pvcs        []*v1.PersistentVolumeClaim
+		options     *storageframework.PerformanceTestOptions
+		stopCh      chan struct{}
+	}
+	var (
+		dInfo *storageframework.DriverInfo
+		l     *local
+	)
+	ginkgo.BeforeEach(func() {
+		// Check preconditions
+		dDriver := driver.(storageframework.DynamicPVTestDriver)
+		if dDriver == nil {
+			e2eskipper.Skipf("Test driver does not support dynamically created volumes")
+
+		}
+		dInfo = dDriver.GetDriverInfo()
+		if dInfo == nil {
+			e2eskipper.Skipf("Failed to get Driver info -- skipping")
+		}
+		if dInfo.PerformanceTestOptions == nil || dInfo.PerformanceTestOptions.ProvisioningOptions == nil {
+			e2eskipper.Skipf("Driver %s doesn't specify performance test options -- skipping", dInfo.Name)
+		}
+	})
+
+	frameworkOptions := framework.Options{
+		ClientQPS:   200,
+		ClientBurst: 400,
+	}
+	f := framework.NewFramework("volume-lifecycle-performance", frameworkOptions, nil)
+	f.AddAfterEach("cleanup", func(f *framework.Framework, failed bool) {
+		ginkgo.By("Closing informer channel")
+		close(l.stopCh)
+		ginkgo.By("Deleting all PVCs")
+		for _, pvc := range l.pvcs {
+			err := e2epv.DeletePersistentVolumeClaim(l.cs, pvc.Name, pvc.Namespace)
+			framework.ExpectNoError(err)
+			err = e2epv.WaitForPersistentVolumeDeleted(l.cs, pvc.Spec.VolumeName, 1*time.Second, 5*time.Minute)
+			framework.ExpectNoError(err)
+		}
+		ginkgo.By(fmt.Sprintf("Deleting Storage Class %s", l.scName))
+		err := l.cs.StorageV1().StorageClasses().Delete(context.TODO(), l.scName, metav1.DeleteOptions{})
+		framework.ExpectNoError(err)
+		l.testCleanup()
+	})
+
+	ginkgo.It("should provision volumes at scale within performance constraints [Slow] [Serial]", func() {
+		l = &local{
+			cs:      f.ClientSet,
+			ns:      f.Namespace,
+			options: dInfo.PerformanceTestOptions,
+		}
+		l.config, l.testCleanup = driver.PrepareTest(f)
+
+		// Stats for volume provisioning operation
+		// TODO: Add stats for attach, resize and snapshot
+		provisioningStats := &performanceStats{
+			mutex:             &sync.Mutex{},
+			perObjectInterval: make(map[string]*interval),
+			operationMetrics:  &storageframework.Metrics{},
+		}
+		sc := driver.(storageframework.DynamicPVTestDriver).GetDynamicProvisionStorageClass(l.config, pattern.FsType)
+		ginkgo.By(fmt.Sprintf("Creating Storage Class %v", sc))
+		// TODO: Add support for WaitForFirstConsumer volume binding mode
+		if sc.VolumeBindingMode != nil && *sc.VolumeBindingMode == storagev1.VolumeBindingWaitForFirstConsumer {
+			e2eskipper.Skipf("WaitForFirstConsumer binding mode currently not supported for performance tests")
+		}
+		ginkgo.By(fmt.Sprintf("Creating Storage Class %s", sc.Name))
+		sc, err := l.cs.StorageV1().StorageClasses().Create(context.TODO(), sc, metav1.CreateOptions{})
+		framework.ExpectNoError(err)
+		l.scName = sc.Name
+
+		// Create a controller to watch on PVCs
+		// When all PVCs provisioned by this test are in the Bound state, the controller
+		// sends a signal to the channel
+		controller := newPVCWatch(f, l.options.ProvisioningOptions.Count, provisioningStats)
+		l.stopCh = make(chan struct{})
+		go controller.Run(l.stopCh)
+		waitForProvisionCh = make(chan []*v1.PersistentVolumeClaim)
+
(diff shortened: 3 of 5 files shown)


RaunakShah:
cc: @msau42 @xing-yang

test/e2e/storage/testsuites/volumeperf.go
```diff
@@ -0,0 +1,287 @@
...
+	ginkgo.It("volume lifecycle performance [Slow]", func() {
+		l := &local{
+			cs: f.ClientSet,
+			ns: f.Namespace,
+		}
+		l.config, l.testCleanup = driver.PrepareTest(f)
+		defer l.testCleanup()
+
+		// Stats for volume provisioning operation
+		// TODO: Add stats for attach, resize and snapshot?
```
xing-yang, line 126: +1

RaunakShah: @xing-yang thinking of adding this in a follow up change to keep the PR size manageable..

xing-yang: Yeah, that's reasonable.

xing-yang:
@msau42, you mentioned several different types of stress tests.  Can you take a look of this PR?

test/e2e/storage/testsuites/volumeperf.go
```diff
@@ -0,0 +1,287 @@
...
+	tsInfo TestSuiteInfo
+}
+
+var _ TestSuite = &volumePerformanceTestSuite{}
+
+const (
+	provisionCount     = 500
+	testTimeout        = 15 * time.Minute
+	expectedLatency    = 5 * time.Minute
+	expectedThroughput = float64(1)
```
pohly, line 50: I think these numbers should be configurable per storage driver because performance and supported number of volumes will vary considerably. We used arguments before, but that turned out to be not easily discoverable. Better extend https://github.com/kubernetes/kubernetes/blob/613ee7f0dbc148e088e67ee7b4f13d97f9655ee0/test/e2e/storage/testsuites/testdriver.go#L195-L202 and skip the test if those options are not set.

The volume size also should be configurable.

test/e2e/storage/testsuites/volumeperf.go
```diff
@@ -0,0 +1,287 @@
...
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/watch"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/kubernetes/test/e2e/framework"
+	e2epv "k8s.io/kubernetes/test/e2e/framework/pv"
+	"k8s.io/kubernetes/test/e2e/storage/testpatterns"
+)
+
+type volumePerformanceTestSuite struct {
```
msau42, line 40: Should we merge this with the stress testsuite?

test/e2e/storage/testsuites/volumeperf.go
```diff
@@ -0,0 +1,287 @@
...
+		defer close(stopCh)
+		waitForProvisionCh = make(chan struct{})
+
+		ginkgo.By(fmt.Sprintf("Creating %d PVCs", provisionCount))
+		for i := 0; i < provisionCount; i++ {
+			pvc := e2epv.MakePersistentVolumeClaim(e2epv.PersistentVolumeClaimConfig{
+				ClaimSize:        "5Mi",
+				StorageClassName: &sc.Name,
+			}, l.ns.Name)
+			pvc, err = l.cs.CoreV1().PersistentVolumeClaims(l.ns.Name).Create(context.TODO(), pvc, metav1.CreateOptions{})
```
msau42, line 158: StorageClasses with WaitForFirstConsumer require a pod to be scheduled in order for provisioning to happen. For now, maybe we need to force all StorageClasses to use Immediate binding.

But longer term, we should have a test that can also use WaitForFirstConsumer binding so we can test the topology path.

test/e2e/storage/testsuites/volumeperf.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+	ginkgo.BeforeEach(func() {
+		// Check preconditions
+		if dDriver = driver.(DynamicPVTestDriver); dDriver == nil {
+			framework.Failf("Test driver does not support dynamically created volumes")
+		}
+	})
+
+	f := framework.NewDefaultFramework("volume-lifecycle-performance")
+	ginkgo.It("volume lifecycle performance [Slow]", func() {
```
msau42, line 117: This should probably also be [Serial] so that other tests don't impact the results.

test/e2e/storage/testsuites/volumeperf.go
```diff
@@ -0,0 +1,287 @@
...
+			provisioningStats.perObjectInterval[pvc.Name] = &interval{
+				create: pvc.CreationTimestamp.Time,
+			}
+			provisioningStats.mutex.Unlock()
+			l.pvcs = append(l.pvcs, pvc)
+		}
+		defer func() {
+			ginkgo.By("Deleting all PVCs")
+			for _, pvc := range l.pvcs {
+				err := l.cs.CoreV1().PersistentVolumeClaims(l.ns.Name).Delete(context.TODO(), pvc.Name, metav1.DeleteOptions{})
```
msau42, line 171: We also need to wait for the PV objects to be deleted. Otherwise, we may uninstall the csi driver before the pv deletion is complete

test/e2e/storage/testsuites/volumeperf.go
```diff
@@ -0,0 +1,287 @@
...
+		now := time.Now()
+		pvcMetrics.mutex.Lock()
+		defer pvcMetrics.mutex.Unlock()
+		countLock.Lock()
+		defer countLock.Unlock()
+
+		// Check if PVC entered the bound state
+		if oldPVC.Status.Phase != v1.ClaimBound && newPVC.Status.Phase == v1.ClaimBound {
+			newPVCInterval, ok := pvcMetrics.perObjectInterval[newPVC.Name]
+			framework.ExpectEqual(ok, true, "PVC %s should exist in interval map already", newPVC.Name)
```
msau42, line 252: If we have an assertion in a go routine, we need to also call GinkgoRecover() so that errors are handled appropriately. See [example](https://github.com/kubernetes/kubernetes/blob/05a46dbb6026dfd9ae11600f8dc08a99994acdec/test/e2e/storage/testsuites/stress.go#L170)

test/e2e/storage/testsuites/volumeperf.go
msau42: Agree let's make this configurable and we could enable the test for hostpath and mock csi drivers

RaunakShah: Done

test/e2e/storage/testsuites/volumeperf.go
RaunakShah: I like the idea of them being separate. There's a lot more we could add here

test/e2e/storage/testsuites/volumeperf.go
RaunakShah: Added a TODO

test/e2e/storage/testsuites/volumeperf.go
RaunakShah: Done

jingxu97:
@RaunakShah could you pls check the test failure?

RaunakShah:
@jingxu97 looks like all tests are passing

test/e2e/storage/testsuites/volumeperf.go
```diff
@@ -0,0 +1,295 @@
...
+		// When all PVCs provisioned by this test are in the Bound state, the controller
+		// sends a signal to the channel
+		controller := newPVCWatch(f, l.options.ProvisioningOptions.Count, provisioningStats)
+		stopCh := make(chan struct{})
+		go controller.Run(stopCh)
+		defer close(stopCh)
+		waitForProvisionCh = make(chan []*v1.PersistentVolumeClaim)
+
+		ginkgo.By(fmt.Sprintf("Creating %d PVCs of size %s", l.options.ProvisioningOptions.Count, l.options.ProvisioningOptions.VolumeSize))
+		for i := 0; i < l.options.ProvisioningOptions.Count; i++ {
```
msau42, line 156: Creating a bunch of resources all at once will probably hit api throttling, which is ok for stress tests, but potentially undesirable for performance benchmarking. In clusterloader2 scale tests, typically groups of resources are created in batches. We may want to do something similar here and make the interval configurable.

test/e2e/storage/testsuites/testdriver.go
```diff
@@ -201,6 +204,26 @@ type StressTestOptions struct {
...
+// Metrics to evaluate performance of an operation
+// TODO: Add metrics like median, mode, standard deviation, percentile
+type Metrics struct {
+	Latency    time.Duration
+	Throughput float64
+}
+
+// PerformanceTestProvisioningOptions contains parameters for
+// testing provisioning operation performance
+type PerformanceTestProvisioningOptions struct {
```
msau42, line 216: Can we enable these options for hostpath and mock csi drivers?

RaunakShah: What values do we default to? Looking at `StressTestOptions`, its enabled for `gcePDCSIDriver` with some default values and also configurable?

test/e2e/storage/testsuites/volumeperf.go
RaunakShah: That's true, but we also want to test throttling of sidecars by dumping requests. What about setting an absurdly high value for `QPS` and `Burst` for the k8sclient that the test uses?

test/e2e/storage/testsuites/volumeperf.go
```diff
@@ -0,0 +1,295 @@
...
+			framework.ExpectNoError(err)
+		}()
+
+		// Create a controller to watch on PVCs
+		// When all PVCs provisioned by this test are in the Bound state, the controller
+		// sends a signal to the channel
+		controller := newPVCWatch(f, l.options.ProvisioningOptions.Count, provisioningStats)
+		stopCh := make(chan struct{})
+		go controller.Run(stopCh)
+		defer close(stopCh)
```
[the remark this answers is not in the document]
RaunakShah: Yes, AFAIK closing the channel kills the goroutine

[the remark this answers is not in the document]
chrishenzie: This would be worth testing to verify. I encountered issues where `defer`ed cleanup functions did not run when I ctrl+c'd a running test. You can work around this by adding cleanup steps via `f.AddAfterEach()`.

test/e2e/storage/testsuites/volumeperf.go
msau42: I think qps and burst should be configurable, but we want to also be able to measure default performance behavior as well.

test/e2e/storage/testsuites/volumeperf.go
msau42: Now that I think about it some more, I think we should support WaitForFirstConsumer in the first iteration because almost all the drivers we test in CI use it. Can we add a Pod.Create call here so that we can handle the WaitForFirstConsumer case?

test/e2e/storage/testsuites/volumeperf.go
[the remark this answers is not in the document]
RaunakShah: @chrishenzie do you have an example usage of `f.AddAfterEach()`? I only see it being used in `NewFramework` in the entire e2e test suite. Alternatively does `f.AfterEach()` solve the problem?

test/e2e/storage/testsuites/volumeperf.go
RaunakShah: What i mean is increasing qps and burst of the k8sclient used by the test to create pvc's on the API server. Line 124 of the latest commit. Doesn't that solve the issue of not throttling while also not modifying the defaults for the components we're aiming to test?

test/e2e/storage/testsuites/volumeperf.go
RaunakShah: Done

test/e2e/storage/testsuites/volumeperf.go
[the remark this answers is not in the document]
chrishenzie: See here for an example usage of `AddAfterEach()`:
https://github.com/kubernetes/kubernetes/blob/fb6bc4f8b0eb6805ca9b766722a176a281bd54d0/test/e2e/storage/testsuites/volume_stress.go#L171-L174

`f.AfterEach()` is ran once either a test finishes or is interrupted, and performs cleanup work (like deleting the test namespace). The way to set additional cleanup steps is by using `f.AddAfterEach()`, which will execute these functions when `f.AfterEach()` gets called by ginkgo.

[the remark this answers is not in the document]
RaunakShah: done!

RaunakShah:
@chrishenzie can you take another look?

chrishenzie:
@RaunakShah This looks good to me. Were you able to see resources cleaned up when CTRL-C'ing the running test?

RaunakShah:
> Were you able to see resources cleaned up when CTRL-C'ing the running test?

Actually now that i try it again, i see that it depends on the timing:

```
^CNov 16 14:15:13.787: INFO: Running AfterSuite actions on all nodes

---------------------------------------------------------
Received interrupt.  Running AfterSuite...
^C again to terminate immediately
STEP: deleting the test namespace: volume-lifecycle-performance-5815
STEP: Waiting for namespaces [volume-lifecycle-performance-5815] to vanish
...
```
and
```
$ kubectl get ns
NAME                                    STATUS        AGE
default                                 Active        6d15h
...
volume-lifecycle-performance-5815       Terminating   39s
volume-lifecycle-performance-5815-622   Active        39s
```

At this point if i CTRL+C again, i see that the namespace that's active doesnt get cleaned up:
```
$ kubectl get ns
NAME                                    STATUS   AGE
...
volume-lifecycle-performance-5815-622   Active   2m18s
```

So when we say CTRL+C, do we mean pressing it once or twice?

chrishenzie:
I just mean once. Pressing more than once it should cancel the cleanup.

RaunakShah:
Thanks for clarifying! In that case, the resources get cleaned up as expected.

msau42:
@RaunakShah do yos still have time to continue on this?

RaunakShah:
@msau42 yes! Do we want to merge this PR and then add some follow up?

pohly:
Here's a quick comparison between running tests via e2e.test (this PR) and [Cluster Loader 2 = CL2](https://github.com/kubernetes/perf-tests/tree/master/clusterloader2). For an example of using the latter, see https://github.com/kubernetes/perf-tests/blob/master/adhoc/log/distributed-provisioning.md

Advantages of CL2:
- tests are divided into steps, with timing and success recorded in a JUnit.xml file for each step
- flexible configuration mechanism via overrides
- templates for object creation
- request rate configurable for each step
- can gather metrics data

Advantages of e2e.test:
- integration with existing tests
- more familiar to developer (?) than CL2

My two cents regarding steps foward: I am worried that if we start adding more stress tests to e2e.test, then eventually we'll want more of the features that CL2 already has and will end up re-implementing CL2 here. I'd prefer to focus on e2e.tests for functional tests that need little to no additional parameters besides the driver configuration and on CL2 for performance tests.

RaunakShah:
Discussed the two options (clusterloader2 vs e2e.test) with @xing-yang and @pohly. We concluded on the following:

1. Since the original intention of this PR was to run this performance test as part of the external sidecar prow jobs, it makes sense to keep this in e2e.test. @pohly since the other reviewers have already approved this PR, can you take a look as well and provide any comments? 
2. We'll dig deeper into clusterloader2 to leverage it for long term performance bench marking.

test/e2e/storage/drivers/csi.go
```diff
@@ -96,6 +96,16 @@ func initHostPathCSIDriver(name string, capabilities map[storageframework.Capabi
 				NumPods:      10,
 				NumSnapshots: 10,
 			},
+			PerformanceTestOptions: &storageframework.PerformanceTestOptions{
+				ProvisioningOptions: &storageframework.PerformanceTestProvisioningOptions{
+					VolumeSize: "1Mi",
+					Count:      300,
+					ExpectedMetrics: &storageframework.Metrics{
+						AvgLatency: 2 * time.Minute,
+						Throughput: 0.5,
```
pohly, line 105: Where did these numbers come from? If they reflect what was seen in some specific system (for example, testing on Prow) then it would be worthwhile to document this here in a comment.

I'm a bit worried about random test flakes in Prow jobs. Anything timing based is problematic because performance is not very deterministic.

test/e2e/storage/framework/testdriver.go
```diff
@@ -224,3 +227,23 @@ type VolumeSnapshotStressTestOptions struct {
...
+	AvgLatency time.Duration
+	Throughput float64
+}
+
+// PerformanceTestProvisioningOptions contains parameters for
+// testing provisioning operation performance
+type PerformanceTestProvisioningOptions struct {
+	VolumeSize      string
+	Count           int
+	ExpectedMetrics *Metrics
```
pohly, line 243: What is the operation that gets measured, i.e. "average latency" of what?

I suppose it is "PVC created" till "PVC is bound", but that's worth mentioning here.

test/e2e/storage/framework/testdriver.go
```diff
@@ -205,6 +206,8 @@ type DriverInfo struct {
 	StressTestOptions *StressTestOptions
 	// [Optional] Scale parameters for volume snapshot stress tests.
 	VolumeSnapshotStressTestOptions *VolumeSnapshotStressTestOptions
+	// [Optional] Parameters for performance tests
+	PerformanceTestOptions *PerformanceTestOptions
```
pohly, line 210: If someone wants to run stress tests with different parameters (for example, different storage classes or different sizes), then this can only be done by defining different test drivers.

I think that's okay, I just wanted to call this out. The alternative (multiple `PerformanceTestOptions` and then instantiating the test multiple times) would be complex, too.

pohly (commented):
Overall this looks fine. I've mostly looked at this from a user perspective, so my comments are mostly around documentation.

test/e2e/storage/drivers/csi.go
RaunakShah: Opted for these numbers based on testing on my local setup (kind + hostpath CSI) and then gave it a loooot of leeway. The intention (for now) is to set the expected metrics so high that if this test fails there's definitely a performance regression.

test/e2e/storage/framework/testdriver.go
RaunakShah: done!

test/e2e/storage/drivers/csi.go
pohly: Can you put that into a comment in the source code?

RaunakShah: done

test/e2e/storage/testsuites/volumeperf.go
```diff
@@ -0,0 +1,327 @@
...
+	storageframework "k8s.io/kubernetes/test/e2e/storage/framework"
+)
+
+type volumePerformanceTestSuite struct {
+	tsInfo storageframework.TestSuiteInfo
+}
+
+var _ storageframework.TestSuite = &volumePerformanceTestSuite{}
+
+const testTimeout = 15 * time.Minute
```
msau42, line 49: How long does this test normally take? There is a push to move any long running tests to opt-in jobs.

test/e2e/storage/testsuites/volumeperf.go
```diff
@@ -0,0 +1,327 @@
...
+			// Store create time for each PVC
+			provisioningStats.mutex.Lock()
+			provisioningStats.perObjectInterval[pvc.Name] = &interval{
+				create: pvc.CreationTimestamp.Time,
+			}
+			provisioningStats.mutex.Unlock()
+			if *sc.VolumeBindingMode == storagev1.VolumeBindingWaitForFirstConsumer {
+				ginkgo.By("creating a pod referring to the claim")
+				var pod *v1.Pod
+				pod, err = e2epod.CreatePod(l.cs, pvc.Namespace, nil /* nodeSelector */, []*v1.PersistentVolumeClaim{pvc}, true /* isPrivileged */, "" /* command */)
```
msau42, line 206: The default max pods per node is 100, so running this with 300 pods is going to fail.

test/e2e/storage/testsuites/volumeperf.go
```diff
@@ -0,0 +1,327 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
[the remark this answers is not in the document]
RaunakShah: done

test/e2e/storage/testsuites/volumeperf.go
RaunakShah: It takes about 10 minutes on my local env and most of that time is for cleanup (we wait for all resources to be deleted).

test/e2e/storage/testsuites/volumeperf.go
RaunakShah: Do the prow jobs only run with a single node?

RaunakShah:
@msau42 i see that the test passed in ~8m

```
Kubernetes e2e suite: [sig-storage] CSI Volumes [Driver: csi-hostpath] [Testpattern: Dynamic PV (filesystem volmode)] volume-lifecycle-performance should provision volumes at scale within performance constraints [Serial] | 8m15s
-- | --
```

Do you want me to add the tag back or does that not count as slow enough for `[Slow]`?

RaunakShah:
Also these were the metrics from the test:

```
I0309 18:29:48.623] Mar  9 18:29:48.620: INFO: Metrics to evaluate: (*framework.Metrics)(0xc003d47190)({
I0309 18:29:48.623]  AvgLatency: (time.Duration) 1m15.751037471s,
I0309 18:29:48.623]  Throughput: (float64) 2.4025551621793855
I0309 18:29:48.623] })
```

test/e2e/storage/testsuites/volumeperf.go
msau42: Ok I guess the reason why this test is passing is because it's not running with "WaitForFirstConsumer". But I suspect that if you change the hostpath driver test to use it, then setting 300 pods is going to fail because you can only have a max of 100 pods per node.

msau42:
> Do you want me to add the tag back or does that not count as slow enough for [Slow]?

Looking at the other tests running here, I see most are taking 1-2 minutes, so I think it makes sense to add the [Slow] tag back.

test/e2e/storage/testsuites/volumeperf.go
RaunakShah: Yes i see what you mean? Do you want me to remove support for WFFC for now and revisit it once i've tested it locally?

RaunakShah: Actually this test is currently skipping WFFC - https://github.com/kubernetes/kubernetes/pull/94334/files#diff-7f88229dff9847a80f3ff9c3fe105a83c19b0b0ef596186a84cfc29337fbaf88R174

msau42: I guess since the test is already skipping it then it's fine. It's just confusing because this code will never be executed, so removing it would avoid that confusion.

RaunakShah: Removed!

msau42:
/lgtm
/approve

Thanks!

PR #94338, opened 2020-08-30
fixed golint error in pkg/apis/admissionregistration/validation

joremysh:
**What type of PR is this?**
/kind cleanup

Fixed golint eror in pkg/apis/admissionregistration

Part of #68026

run hack/update-bazel.sh and hack/update-gofmt.sh ok

```release-note
NONE
```

### DIFF
--- a/hack/.golint_failures
+++ b/hack/.golint_failures
@@ -8,7 +8,6 @@ pkg/apis/abac/v1beta1
 pkg/apis/admission
 pkg/apis/admissionregistration/v1
 pkg/apis/admissionregistration/v1beta1
-pkg/apis/admissionregistration/validation
 pkg/apis/apps/v1
 pkg/apis/apps/v1beta1
 pkg/apis/apps/v1beta2
--- a/pkg/apis/admissionregistration/v1/defaults.go
+++ b/pkg/apis/admissionregistration/v1/defaults.go
@@ -27,6 +27,7 @@ func addDefaultingFuncs(scheme *runtime.Scheme) error {
 	return RegisterDefaults(scheme)
 }
 
+// SetDefaults_ValidatingWebhook sets defaults for webhook validating
 func SetDefaults_ValidatingWebhook(obj *admissionregistrationv1.ValidatingWebhook) {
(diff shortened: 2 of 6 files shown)


neolit123:
> Fixes #68026
please change this to something like:

> Part of #68026

or it will close the referenced ticket.

neolit123 (commented):
/ok-to-test
/priority backlog
/sig auth

pkg/apis/admissionregistration/v1/defaults.go
```diff
@@ -27,7 +27,8 @@ func addDefaultingFuncs(scheme *runtime.Scheme) error {
 	return RegisterDefaults(scheme)
 }
 
-func SetDefaults_ValidatingWebhook(obj *admissionregistrationv1.ValidatingWebhook) {
```
liggitt, line 30: The name of defaulting methods is fixed and cannot be changed without also changing the generated defaulters. That is not currently planned, due to the wide ripples in all codebases that depend on generated defaults.

hack/.golint_failures
```diff
@@ -6,9 +6,6 @@ cmd/kubeadm/app/apis/kubeadm/v1beta2
 pkg/apis/abac/v0
 pkg/apis/abac/v1beta1
 pkg/apis/admission
-pkg/apis/admissionregistration/v1
-pkg/apis/admissionregistration/v1beta1
-pkg/apis/admissionregistration/validation
```
liggitt, line 11: the validation package can be resolved, but the defaulting functions in the v1/v1beta1 packages should be left as-is

joremysh (commented):
Only update in comments now.

hack/.golint_failures
JeremyShih: updated.

pkg/apis/admissionregistration/v1/defaults.go
JeremyShih: The changes on names of method are all removed now.

JeremyShih:
Hi @liggitt could you give the PR a milestone?

PR #94347, opened 2020-08-31
Make kubeconfig command-line help message of kube-proxy more precise

### LINKED ISSUE #94346: Ambiguous help message in kube-proxy
**What happened**:
Run `kube-proxy -h`, we could see two related command-line options.
The explanation for `--kubeconfig` is ambiguous, it says the master location is ought to be set by the master flag.
This statement is inconsistent with the actual behavior.
```
--kubeconfig string    Path to kubeconfig file with authorization information (the master location is set by the master flag).
--master string           The address of the Kubernetes API server (overrides any value in kubeconfig)
```

In fact, the `--master` flag is optional, it only overrides the control plane address in `--kubeconfig`.

That means, if we are using a single-context kubeconfig(or never changes the context), it's OK to just set `--kubeconfig <path_to_kubeconfig>` and leave  `--master` unset. Kube-proxy will parse the kubeconfig and use the credentials and master address defined in current context to contact cluster.

Truth from code shows that kubeconfig parser do honors context setting:
https://github.com/kubernetes/kubernetes/blob/02b5b60ecdf663acc00bafe534ce7483aaf2d782/cmd/kube-proxy/app/server.go#L548-L563

https://github.com/kubernetes/kubernetes/blob/02b5b60ecdf663acc00bafe534ce7483aaf2d782/staging/src/k8s.io/client-go/tools/clientcmd/client_config.go#L418-L429

`kube-proxy -h | grep master`

**Environment**:
- Kubernetes version (use `kubectl version`): v1.19.0

/sig network
/area kube-proxy

Einsfier:
**What type of PR is this?**
/kind documentation

The original statement is ambiguous. It indicates that the Kubernetes master API server should be explicitly set by `--master` flag. This is redundant in certain cases, such as an exclusive kubeconfig for `kube-proxy` which is already recommended in `kubelet` configuration.

Fixes #94346 

docs update is here https://github.com/kubernetes/website/pull/23580

```release-note
NONE
```

```docs

```

### DIFF
--- a/cmd/kube-proxy/app/server.go
+++ b/cmd/kube-proxy/app/server.go
@@ -144,7 +144,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 
 	fs.StringVar(&o.ConfigFile, "config", o.ConfigFile, "The path to the configuration file.")
 	fs.StringVar(&o.WriteConfigTo, "write-config-to", o.WriteConfigTo, "If set, write the default configuration values to this file and exit.")
-	fs.StringVar(&o.config.ClientConnection.Kubeconfig, "kubeconfig", o.config.ClientConnection.Kubeconfig, "Path to kubeconfig file with authorization information (the master location is set by the master flag).")
+	fs.StringVar(&o.config.ClientConnection.Kubeconfig, "kubeconfig", o.config.ClientConnection.Kubeconfig, "Path to kubeconfig file with authorization information (the master location can be overridden by the master flag).")
 	fs.StringVar(&o.config.ClusterCIDR, "cluster-cidr", o.config.ClusterCIDR, "The CIDR range of pods in the cluster. When configured, traffic sent to a Service cluster IP from outside this range will be masqueraded and traffic sent from pods to an external LoadBalancer IP will be directed to the respective cluster IP instead")
 	fs.StringVar(&o.config.ClientConnection.ContentType, "kube-api-content-type", o.config.ClientConnection.ContentType, "Content type of requests sent to apiserver.")
 	fs.StringVar(&o.master, "master", o.master, "The address of the Kubernetes API server (overrides any value in kubeconfig)")


dcbw:
@povsister Two quick questions... can you explain "master location in context"?  Also, are there other components that have similar options and what do those say?

povsister:
@dcbw 
> can you explain "master location in context"

OK. "master location in context" means the kube apiserver location in current-context.

The context is designed for multi-cluster and multi-user environment, provides a various cluster endpoints and authentication mechanisms in a single kubeconfig file. The only thing that controls which cluster to connect and which credential to use is the "current-context" setting in the file, of course client-go honors this setting, as well as all kube components who uses kubeconfig file to access kube apiserver.

Since this part is located in “documents -> reference -> CLI reference”
I assuming that people has already acknowledged the [context concept in kubeconfig](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/)


> are there other components that have similar options and what do those say?

I just checked the documentation, the answer is Yes.
Scheduler and Controller-manager also have such options pair described as below.
```
--kubeconfig string        Path to kubeconfig file with authorization and master location information.
--master string      The address of the Kubernetes API server (overrides any value in kubeconfig).
```
Comparing to the kube-proxy, there is no confusing with the description. Only lack the mention of the context setting in kubeconfig.

danwinship:
Yeah the "in context" confused me at first too, because it looks like you're just using the phrase "in context" rather than referring to a section of the kubeconfig. You can just remove that from the message though and I think it still makes sense.

povsister:
@danwinship 
Nice advice.
The statement is a bit hard to get the key-point for users who are not familiar with "context" concept in kubeconfig.
I used to express things precisely and technically, especially for Devs. But it feels not good without hyperlinks.. haha.

How about make the statement same as kube-scheduler/kube-controller says ?
I am now in favor of their descriptions on this flag.

Ps: I am out and on vacation. Unable to push any fix till tomorrow.

aojea:
/lgtm
Thanks

Issue #93505, opened 2020-07-28
race condition detected during the scheduling with preemption 


chendave:
**What happened**:
race condition detected during the scheduling with preemption 

**What you expected to happen**:
the preemptor with high priority should be scheduled successfully.

**How to reproduce it (as minimally and precisely as possible)**:
In the case of preemption,  with the below assumption, 

- node1 has the replication controller with couple of pods running, and consider to be the candidate node for the preemption, 

   ![image](https://user-images.githubusercontent.com/6007417/88650641-7ebe7080-d0fb-11ea-884b-a93a982b2e56.png)

- node2 also has some pods running on it but will not be considered to be the candidate node due to either priority or topology constraints, 
![image](https://user-images.githubusercontent.com/6007417/88651160-f2607d80-d0fb-11ea-8c4f-a009e6fd88ad.png)

the scheduling sequence very likely will hit the race condition like below, 

step1: preemptor pod (pod6) is coming and found no fit node to schedule
step2: error is recorded and the pod6 is added into the unschedulableQ
step3: preemption is happen on node1 and pod1 is deleted
step4: pod1 is added to activeQ
step5: field of `NominatedNodeName` is patched to pod6
step6: next scheduling cycle is started and the pod1 is only pod in the `activeQ`
step7: going to schedule pod in the `activeQ` (pod1) and the candidate node is node1
step8: event of `OnUpdate` of the preemptor pod (pod6) is captured and pod6 is moved from `unschedulableQ` to `activeQ`
step9: pod1 is scheduled on the node1
step10: next scheduling cycle is started and the pod6 is the only pod in the `activeQ`
step11: preemption is happened again and pod2 is removed this time.
step12: error is recorded again and the pod is moved to `podBackoffQ`, and the `nominatedNode` is still node1
step13: pod `condition` is not changed, `nominatedNode` is not changed, so the pod is not updated, pod6 is still in the `podBackoffQ`.
step14:  pod6 needs to wait for backoff time expired before getting moved from  `podBackoffQ` to `activeQ`
step15:  pod2 is scheduled on node1.

looping forever from here !!!  along with pod deletion and preemption on node1.
...
...

The root cause is step8 `OnUpdate`  event is not guaranteed to be finished before step7, they are handled asynchronously.

[...]

Issue #94009, opened 2020-08-14
Avoid moving pods out of unschedulable status unconditionally


ahg-g:
Currently we unconditionally move pods out of unschedulable status when an event happens. However, we can optimize this by setting some conditions for each event to avoid unnecessarily doing that:

1) PV, PVC, StorageClass or CSINode add/update: move only pods with a PVC.
2) Service add/update: only if ServiceAffinity plugin is enabled in one of the profiles
3) Node add/update: test the pods against the [kubelet admission logic](https://github.com/kubernetes/kubernetes/blob/3b5aedcef43fa01538518ecbc7de369b27ecd191/pkg/kubelet/lifecycle/predicate.go#L224), this include: NodeAffinity, NodeName NodePorts, NodeResources and TaintsTolerations filters.
4) Pod delete: keep it as is (spreading or affinity constraints must be tested, which is expensive to do)

This is necessary to avoid wasting scheduling cycles.

Related proposal: https://docs.google.com/document/d/1Dw1qPi4eryllSv0F419sKVbGXiPvSv6N_mJd6AVSg74/edit

/sig scheduling
/assign @adtac

Huang-Wei:
/cc @denkensk 

I'm thinking of if we can amplify the scope: which is to say, not only enable/disable moving **(all)** pods, but also provides mechanics to let plugins notify the scheduler framework which **particular** pods should be moved.

This would benefits in @denkensk 's coscheduling implementation to move the preceding (N-1) pods back to activeQ when the Nth pod comes - which satisfies the "minAvaialbe" requirement. Moreover, this would help in getting rid of flushing 
 pods back to activeQ #87850.

ahg-g:
The suggestion is not to enable/disable moving all pods in all events, this is only true for the Service events. For the other two, node and volumes, we will test each pod and move the ones that pass the condition.

I agree that our proposal of integrating this with filters/permit plugins is more generic, but it adds a lot of complexity and potential for bugs, so we need to be careful with that I think.

ahg-g:
One approach could be to start simple, validate and establish a reference point in terms of perf improvements, and then migrate to something more generic. The more generic approach will not be more performant than what is proposed in this issue.

Huang-Wei:
> I agree that our proposal of integrating this with filters/permit plugins is more generic, but it adds a lot of complexity and potential for bugs, so we need to be careful with that I think.

[...]

PR #94358, opened 2020-08-31
Exposure the race condition on the pod preemption

chendave:
- Fix a potential test flake.
- Add new testcase to exposure the race condition on pod preemption.
- Take the chance to fix a typo.

**What type of PR is this?**

/kind bug

Ref https://github.com/kubernetes/kubernetes/issues/93505

```release-note

```

```docs

```

### DIFF
--- a/pkg/scheduler/internal/queue/scheduling_queue_test.go
+++ b/pkg/scheduler/internal/queue/scheduling_queue_test.go
@@ -359,16 +359,76 @@ func TestPriorityQueue_Delete(t *testing.T) {
 }
 
 func TestPriorityQueue_MoveAllToActiveOrBackoffQueue(t *testing.T) {
-	q := NewPriorityQueue(newDefaultQueueSort())
+	c := clock.NewFakeClock(time.Now())
+	q := NewPriorityQueue(newDefaultQueueSort(), WithClock(c))
 	q.Add(&medPriorityPod)
 	q.AddUnschedulableIfNotPresent(q.newQueuedPodInfo(&unschedulablePod), q.SchedulingCycle())
 	q.AddUnschedulableIfNotPresent(q.newQueuedPodInfo(&highPriorityPod), q.SchedulingCycle())
+	// Move clock by 1 second so that pods is guaranteed to be added into unschedulableQ first and
+	// then the move request is happened later, otherwise there is potential a test flake here, as
+	// the move request has a chance to be finished before the pods get into the unschedulableQ.
+	// In that case, the pods will be moved to backoffQ even `QueuedPodInfo.Attempts` is zero.
+	c.Step(time.Second)
+	// move the pods from unschedulableQ to activeQ.
 	q.MoveAllToActiveOrBackoffQueue("test")
+	if q.activeQ.Len() != 3 {
+		t.Error("Expected 3 item to be in activeQ")
+	}
+	if q.podBackoffQ.Len() != 0 {
+		t.Error("Expected 0 items to be in podBackoffQ")
+	}
+}
+
+// TestPriorityQueue_RaceConditionOnScheduling mimic a race condition could be found with preemption.
+// In this particular case, the high priority pod is stuck in the backoffQ and if new pod is coming,
+// regardless it's priority, the new pod will be scheduled first.
+// The preemption could happen again but there is a chance the pod will never got scheduled as long as
+// there is new pod coming for scheduling since the high priority pod could be able to moved from backoffQ
+// to activeQ only when the time for backoff expiration is up.
+// see https://github.com/kubernetes/kubernetes/issues/93505 for the details.
+func TestPriorityQueue_RaceConditionOnScheduling(t *testing.T) {
+	c := clock.NewFakeClock(time.Now())
+	q := NewPriorityQueue(newDefaultQueueSort(), WithClock(c))
+	// comes a lowPriority pod and is added to activeQ, this mimic a pod that is managed by
+	// replication controller is preempted by the `highPriorityPod` and a new pod is going to
+	// be re-scheduled.
+	q.Add(&unschedulablePod)
+	// events that make the move request is happen, pod deletion, service add etc.
+	// which will trigger the move request.
+	q.MoveAllToActiveOrBackoffQueue("test")
+	// Move the clock by 1 second in order to guarantee the move request is happened before the
+	// highPriorityPod wants to be added to unschedulableQ.
+	c.Step(time.Second)
+	// even the unschedulableQ is empty and `QueuedPodInfo.Attempts` is empty, the highPriorityPod
+	// is pushed to backoffQ instead of unschedulableQ.
+	q.AddUnschedulableIfNotPresent(q.newQueuedPodInfo(&highPriorityPod), q.SchedulingCycle())
 	if q.activeQ.Len() != 1 {
 		t.Error("Expected 1 item to be in activeQ")
 	}
-	if q.podBackoffQ.Len() != 2 {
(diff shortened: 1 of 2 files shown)


chendave:
/cc @Huang-Wei  @ahg-g  @alculquicondor  @soulxu 

I will propose a fix later, the preemption could eventually got convergent if we skip the move rquest when the `unscheduleQ` is empty.

pkg/scheduler/internal/queue/scheduling_queue_test.go
```diff
@@ -363,12 +363,69 @@ func TestPriorityQueue_MoveAllToActiveOrBackoffQueue(t *testing.T) {
 	q.Add(&medPriorityPod)
 	q.AddUnschedulableIfNotPresent(q.newQueuedPodInfo(&unschedulablePod), q.SchedulingCycle())
 	q.AddUnschedulableIfNotPresent(q.newQueuedPodInfo(&highPriorityPod), q.SchedulingCycle())
+	// sleep 1 second so that pods is guaranteed to be added into unschedulableQ first and then
+	// the move request is happened later, otherwise there is potential a test flake here, as the
+	// move request has a chance to be finished before the pods get into the unschedulableQ.
+	// In that case, the pods will be moved to backoffQ even `QueuedPodInfo.Attempts` is zero.
+	time.Sleep(1000 * time.Millisecond)
```
alculquicondor, line 370: Do not do this. Instead, create the queue with a fake clock: `q := NewPriorityQueue(newDefaultQueueSort(), WithClock(clock.NewFakeClock(time.Now())))`

pkg/scheduler/internal/queue/scheduling_queue_test.go
```diff
@@ -363,12 +363,69 @@ func TestPriorityQueue_MoveAllToActiveOrBackoffQueue(t *testing.T) {
...
+	}
+}
+
+// TestPriorityQueue_RaceConditionOnScheduling mimic a race condition could be found with preemption.
+// In this particular case, the high priority pod is stuck in the backoffQ and if new pod is coming,
+// regardless it's priority, the new pod will be scheduled first.
+// The preemption could happen again but there is a chance the pod will never got scheduled as long as
+// there is new pod coming for scheduling since the high priority pod could be moved from backoffQ
+// to activeQ only when the time for backoff expiration is up.
+// see https://github.com/kubernetes/kubernetes/issues/93505 for the details.
```
alculquicondor, line 386: I'm not sure how is this related. Could you document your findings in the issue?

pkg/scheduler/internal/queue/scheduling_queue_test.go
```diff
@@ -363,12 +363,69 @@ func TestPriorityQueue_MoveAllToActiveOrBackoffQueue(t *testing.T) {
...
+	q.AddUnschedulableIfNotPresent(q.newQueuedPodInfo(&highPriorityPod), q.SchedulingCycle())
 	if q.activeQ.Len() != 1 {
 		t.Error("Expected 1 item to be in activeQ")
 	}
-	if q.podBackoffQ.Len() != 2 {
-		t.Error("Expected 2 items to be in podBackoffQ")
+	if q.podBackoffQ.Len() != 1 {
+		t.Error("Expected 1 items to be in podBackoffQ")
+	}
+	// the lowPriority pod is popped and got scheduled while the highPriorityPod is stuck in the
```
alculquicondor, line 408: This is by design to avoid starvation from high priority pods. Let's discuss in the original issue. However, we intend to reduce the problems of this with #94009

pkg/scheduler/internal/queue/scheduling_queue_test.go
chendave: Done

pkg/scheduler/internal/queue/scheduling_queue_test.go
chendave: problem here is that the `highPriorityPod` should not be in `backoffQ` in the first place, it is moved to be `backoffQ` just because there was a move request happened first.
`lowPriority` pod popped here should be fine, but there should a way for the preemption to get convergent,  pls see the code following, 
```golang
	// another pod is added to activeQ.
	q.Add(&unschedulablePod)
	if q.activeQ.Len() != 1 {
		t.Error("Expected 1 item to be in activeQ")
	}
```
it's possible a new pod is added to Queue while the `highPriorityPod` is still backoff-ing, after the backoff is up, the `highPriorityPod` needs to preempt the `lowPriority` pod again.

pkg/scheduler/internal/queue/scheduling_queue_test.go
chendave: updated in that issue, pls kindly check it at your convenient time.

pkg/scheduler/internal/queue/scheduling_queue_test.go
```diff
@@ -359,16 +359,76 @@ func TestPriorityQueue_Delete(t *testing.T) {
...
+	c := clock.NewFakeClock(time.Now())
+	q := NewPriorityQueue(newDefaultQueueSort(), WithClock(c))
 	q.Add(&medPriorityPod)
 	q.AddUnschedulableIfNotPresent(q.newQueuedPodInfo(&unschedulablePod), q.SchedulingCycle())
 	q.AddUnschedulableIfNotPresent(q.newQueuedPodInfo(&highPriorityPod), q.SchedulingCycle())
+	// Move clock by 1 second so that pods is guaranteed to be added into unschedulableQ first and
+	// then the move request is happened later, otherwise there is potential a test flake here, as
+	// the move request has a chance to be finished before the pods get into the unschedulableQ.
+	// In that case, the pods will be moved to backoffQ even `QueuedPodInfo.Attempts` is zero.
+	c.Step(time.Second)
```
alculquicondor, line 371: Could you rather leave the original test case without the Step?

Or we could have 2 tests.

pkg/scheduler/internal/queue/scheduling_queue_test.go
alculquicondor: I prefer this to be split in a different PR so we can already merge the tests using the deterministic clock.

pkg/scheduler/internal/queue/scheduling_queue_test.go
```diff
@@ -359,16 +359,76 @@ func TestPriorityQueue_Delete(t *testing.T) {
...
+	q.AddUnschedulableIfNotPresent(q.newQueuedPodInfo(&highPriorityPod), q.SchedulingCycle())
+	if q.activeQ.Len() != 1 {
+		t.Error("Expected 1 item to be in activeQ")
+	}
+	if q.podBackoffQ.Len() != 1 {
+		t.Error("Expected 1 items to be in podBackoffQ")
+	}
+	// the lowPriority pod is popped and got scheduled while the highPriorityPod is stuck in the
+	// backoffQ.
+	q.Pop()
```
soulxu, line 413: I still stuck at the problem of the low priority pod shouldn't be gotten a chance to schedule successful, since when the low priority pod getting scheduling, the resources already token by high priority pod.

chendave: please thinking about the case that high priority pod is still in the `unscheduleQ` or `backoffQ`, low priority pod is the only one in the `activeQ`.

pkg/scheduler/internal/queue/scheduling_queue_test.go
chendave: Done, it's here: https://github.com/kubernetes/kubernetes/pull/94413

PR #94363, opened 2020-08-31
Add WatchListPageSize to cache.Config

mborsz:
**What type of PR is this?**
/kind feature

Currently there is no way to specify WatchListPageSize used by Controller. This PR adds a field that can be used to specify this.

Context: https://github.com/kubernetes/perf-tests/pull/1441#issuecomment-683658227

```release-note
NONE
```

```docs

```

/assign @wojtek-t

### DIFF
--- a/staging/src/k8s.io/client-go/tools/cache/controller.go
+++ b/staging/src/k8s.io/client-go/tools/cache/controller.go
@@ -72,6 +72,9 @@ type Config struct {
 
 	// Called whenever the ListAndWatch drops the connection with an error.
 	WatchErrorHandler WatchErrorHandler
+
+	// WatchListPageSize is the requested chunk size of initial and relist watch lists.
+	WatchListPageSize int64
 }
 
 // ShouldResyncFunc is a type of function that indicates if a reflector should perform a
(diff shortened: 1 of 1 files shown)


mborsz:
While I'm not sure if we will be able to use this change to fix issue in https://github.com/kubernetes/perf-tests/pull/1441#issuecomment-683658227 due to potential k/k release timelines, but I think it's a good change anyway (and we should be able to switch slo-monitor to it eventually).

staging/src/k8s.io/client-go/tools/cache/controller.go
```diff
@@ -72,6 +72,9 @@ type Config struct {
 
 	// Called whenever the ListAndWatch drops the connection with an error.
 	WatchErrorHandler WatchErrorHandler
+	
+	// WatchListPageSize is the requested chunk size of initial and resync watch lists.
```
wojtek-t, line 76: We use "resync" terminology in the context of resyncing from in-memory state. What you means is relist.

mborsz: I copied this line from reflector.go :D Fixed here. I can fix in reflector.go as well if you want.

PR #94364, opened 2020-08-31
Efficient watch resumption

wojtek-t:
Ref https://github.com/kubernetes/enhancements/issues/1904

/kind feature

```release-note
NONE
```

### DIFF
--- a/pkg/controlplane/instance.go
+++ b/pkg/controlplane/instance.go
@@ -258,7 +258,7 @@ func (c *Config) createLeaseReconciler() reconcilers.EndpointReconciler {
 	if err != nil {
 		klog.Fatalf("Error determining service IP ranges: %v", err)
 	}
-	leaseStorage, _, err := storagefactory.Create(*config)
+	leaseStorage, _, err := storagefactory.Create(*config, nil)
 	if err != nil {
 		klog.Fatalf("Error creating storage factory: %v", err)
 	}
--- a/staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
+++ b/staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
@@ -29,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/apiserver/pkg/storage"
+	"k8s.io/apiserver/pkg/storage/etcd3/metrics"
 	"k8s.io/apiserver/pkg/storage/value"
 
 	"go.etcd.io/etcd/clientv3"
@@ -68,6 +70,8 @@ func TestOnlySetFatalOnDecodeError(b bool) {
 type watcher struct {
 	client      *clientv3.Client
 	codec       runtime.Codec
+	newFunc     func() runtime.Object
+	objectType  string
 	versioner   storage.Versioner
 	transformer value.Transformer
 }
@@ -78,6 +82,7 @@ type watchChan struct {
 	key               string
 	initialRev        int64
 	recursive         bool
+	progressNotify    bool
 	internalPred      storage.SelectionPredicate
 	ctx               context.Context
 	cancel            context.CancelFunc
@@ -86,13 +91,20 @@ type watchChan struct {
 	errChan           chan error
 }
 
-func newWatcher(client *clientv3.Client, codec runtime.Codec, versioner storage.Versioner, transformer value.Transformer) *watcher {
-	return &watcher{
+func newWatcher(client *clientv3.Client, codec runtime.Codec, newFunc func() runtime.Object, versioner storage.Versioner, transformer value.Transformer) *watcher {
+	res := &watcher{
 		client:      client,
 		codec:       codec,
+		newFunc:     newFunc,
 		versioner:   versioner,
 		transformer: transformer,
 	}
+	if newFunc == nil {
+		res.objectType = "<unknown>"
+	} else {
+		res.objectType = reflect.TypeOf(newFunc()).String()
+	}
+	return res
 }
 
 // Watch watches on a key and returns a watch.Interface that transfers relevant notifications.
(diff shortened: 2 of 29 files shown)


staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
```diff
@@ -86,13 +91,20 @@ type watchChan struct {
 	errChan           chan error
 }
 
-func newWatcher(client *clientv3.Client, codec runtime.Codec, versioner storage.Versioner, transformer value.Transformer) *watcher {
-	return &watcher{
+func newWatcher(client *clientv3.Client, codec runtime.Codec, newFunc func() runtime.Object, versioner storage.Versioner, transformer value.Transformer) *watcher {
```
wojtek-t, line 94: Unit tests are still missing for changes in this file, but they are blocked on our ability to tun etcd for testing:
https://github.com/etcd-io/etcd/issues/12265

Will work on making that happen.

wojtek-t:
@lavalamp @deads2k - this is ready for the first pass of review - PTAL when you will have chance (there are some missing unit tests blocked on etcd changes to their integration framework use, but that doesn't block reviewing the main code changes)

staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
```diff
@@ -299,6 +321,19 @@ func (wc *watchChan) transform(e *event) (res *watch.Event) {
...
+			return nil
+		}
+		object := wc.watcher.newFunc()
+		if err := wc.watcher.versioner.UpdateObject(object, uint64(e.rev)); err != nil {
+			klog.Errorf("failed to propagate object version: %v", err)
+			return nil
+		}
+		res = &watch.Event{
+			Type:   watch.Bookmark,
+			Object: object,
```
lavalamp, line 335: I guess I would have expected a new bookmark type of object rather than a object that has the same type as the other objects. Is there a reason you did it this way?

wojtek-t: Two reasons:
Compatibility: this is exactly what we did for watch bookmarks https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190206-watch-bookmark.md#proposal:
```
As a result, we will represent bookmark event by setting Bookmark type and Object of appropriate type with just ObjectMeta.ResourceVersion field set.
```

[The other reason is also that our client libraries (reflector, etc.) assume that incoming objects are all of the exact same type.]

lavalamp: I see. I guess I didn't see/understand that originally, I wish we hadn't done it that way. I'm a little concerned unsophisticated clients might confuse this for a real object. I probably assumed "of appropriate type" meant bookmark-typed, not object-typed.

I guess, clients have to opt-in for bookmarks anyway, which kind of destroys my argument here. But it also means we could have made a bookmark type and fixed our libraries that made bad assumptions.

I guess it's too late to fix now without making clients opt-in again :(

staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
wojtek-t: The tests are now added.

staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
wojtek-t: Yeah - I would change that in this PR at least (we may consider doing that independently from that change for watch bookmarks, but even for that I'm not sure it's really worth it).

lavalamp:
I think I read all of this (?) and didn't have other thoughts. @deads2k do you want to take a look?

jingyih:
> @lavalamp @deads2k - this is ready for the first pass of review - PTAL when you will have chance (there are some missing unit tests blocked on etcd changes to their integration framework use, but that doesn't block reviewing the main code changes)

I just sent #94711 to see if we can unblock this PR sooner.

wojtek-t:
@lavalamp @deads2k - this is ready for the next pass - all dependencies are also merged and the PR is complete. PTAL

pkg/controlplane/instance.go
```diff
@@ -258,7 +258,7 @@ func (c *Config) createLeaseReconciler() reconcilers.EndpointReconciler {
 	if err != nil {
 		klog.Fatalf("Error determining service IP ranges: %v", err)
 	}
-	leaseStorage, _, err := storagefactory.Create(*config)
+	leaseStorage, _, err := storagefactory.Create(*config, nil)
```
lavalamp, line 261: Instead of changing every call site, can the new parameter be a member of the config struct?

staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
```diff
@@ -86,13 +91,20 @@ type watchChan struct {
 	errChan           chan error
 }
 
-func newWatcher(client *clientv3.Client, codec runtime.Codec, versioner storage.Versioner, transformer value.Transformer) *watcher {
-	return &watcher{
+func newWatcher(client *clientv3.Client, codec runtime.Codec, newFunc func() runtime.Object, versioner storage.Versioner, transformer value.Transformer) *watcher {
+	res := &watcher{
 		client:      client,
 		codec:       codec,
+		newFunc:     newFunc,
```
lavalamp, line 98: I'm a little surprised you can't get this out of the codec or versioner already.

staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
lavalamp: It'll probably never be worth it :'(

lavalamp: I guess code-wise it seems like this change is a lot bigger than it needs to be because of the need to plumb the newFunc everywhere. Would it make sense to keep a bookmark-type object internally and only convert to the "real" type object right before encoding to the wire? (which I believe is here?)

lavalamp:
I'd like one more set of eyes on this before I approve, @jingyih can you review? I see you're already in the comments.

/assign @jingyih

staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
wojtek-t: I think it could make sense in the past. But given we already did it like that for watch bookmarks, I think it's a bit too late.

> right before encoding to the wire? (which I believe is here?)

It actually isn't here. There are a number of layers that this event goes through (watchache, the fixup part (which sets selflink) etc.) so that actually wouldn't be easier codewise.

I agree that that passing newFunc is a bit unfortunate, but it's also boilerplate easy to review (and super low-risk change).

staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
wojtek-t: Unfortunately you can't.
Codec only supports encoding and decoding:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/runtime/interfaces.go#L94
Versioner is only about conversions:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/runtime/interfaces.go#L220

There is ObjectCreater interface:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/runtime/interfaces.go#L254
that our runtime.Scheme implements, but that doesn't really change much in terms of passing arguments.

pkg/controlplane/instance.go
wojtek-t: This is a good point - will fix tomorrow.

wojtek-t: Actually - this is part of `createLeaseResonciler()` method - this is basically callled once in the kube-apiserver lifetime.
So I don't think it makes sense to change anything.

lavalamp: I meant to include all the "generic.NewRawStorage(config, nil)" calls, too. If you think it's not worth it, that's fine.

It still seems weird to me to have a config thingy, and an option that's not part of the config thingy. What is the rule for adding to the config thingy vs adding another parameter?

wojtek-t: Technically it may be possible to make it a part of config.

That said I looked deeper into the code (this time again), and piping the necessary information down to storageConfig (if we do that, we shouldn't pipe newFunc, but rather groupKind and use ObjectCreator from runtime.Scheme to create object) and that would be at least couple hundreds line of code (and e.g. path for CRDs is different than for built-in resources there).

I'm happy to look at this, but given this commit is actually fairly simple and relatively small (and almost all changes are in tests), I would prefer to not block this PR on it, so that we can get more soaking time of the feature itself.

If you're fine with this, I'm happy to open an issue and assign to myself to fix that once this one is merged.

lavalamp: What you have sounds better than plumbing it everywhere. I guess I was expecting something like, this code makes a shallow copy of the config and sets the additional field, so it'd be about the same change as what you have here, just call sites that are happy with `nil` wouldn't need to change at all.

But I wouldn't insist on this or anything, it might not be an improvement.

jingyih (commented):
/lgtm

PR #94372, opened 2020-08-31
autoscaling: Omit optional field `StabilizationWindowSeconds` when empty

mikkeloscar:
**What type of PR is this?**
/kind bug

Annotates the optional field `StabilizationWindowSeconds` with `omitempty` such that it will be omitted when creating resources where the field is not set.

The motivation for this change is that we're importing the `autoscaling/v2beta2` structs and using them in CRDs and without this fix the structs will not be correctly marshaled when using the Go clientset to create the resources e.g. you get an error like:

```
StackSet.zalando.org "hpa-behavior" is invalid: spec.stackTemplate.spec.horizontalPodAutoscaler.behavior.scaleDown.stabilizationWindowSeconds: Invalid value: "null": spec.stackTemplate.spec.horizontalPodAutoscaler.behavior.scaleDown.stabilizationWindowSeconds in body must be of type integer: "null"
```

Because the field is not set and thus defaults to `nil`.

It seems to me that it's a mistake that the field wasn't annotated with `omitempty` to being with. Thus it shouldn't harm to do it.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/api/autoscaling/v2beta2/types.go
+++ b/staging/src/k8s.io/api/autoscaling/v2beta2/types.go
@@ -169,7 +169,7 @@ type HPAScalingRules struct {
 	// - For scale up: 0 (i.e. no stabilization is done).
 	// - For scale down: 300 (i.e. the stabilization window is 300 seconds long).
 	// +optional
-	StabilizationWindowSeconds *int32 `json:"stabilizationWindowSeconds" protobuf:"varint,3,opt,name=stabilizationWindowSeconds"`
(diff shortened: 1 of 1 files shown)


mikkeloscar:
It's not 100% clear to me if this requires a review or not. I.e. is it considered: "Any change to the meaning, validation, or behavior of a field"?

/label api-review

lavalamp:
oh, I'm a little surprised no generated files needed to be touched.

/lgtm
/approve
/milestone v1.20

(and yes, this needed an api review)

PR #77653, opened 2019-05-09
apiextensions: disallow metadata specs other than name and generateName

sttts:
Based on https://github.com/kubernetes/kubernetes/pull/77207.

Fixes https://github.com/kubernetes/kubernetes/issues/74620

CRDs should not restrict ObjectMeta fields other than name and generateName. Restricting other metadata values is considered breaking the API conventions. We better don't allow schemas which do that.

With structural schema introduction we have the chance to enforce this.

We do not care about `x-kubernetes-embedded-resource:true` and `metadata` there as those objects are not yet Kubernetes objects and the API server is not aware of them as resources. 

```release-note
CRD validation schemas should not specify `metadata` fields other than `name` and `generateName`. A schema will not be considered structural (and therefore ready for future features) if `metadata` is specified in any other way.
```

Issue #74620, opened 2019-02-26
Metadata validation for CustomResources


roycaihw:
We allow CRD authors to write [validation for CustomResources](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/customresources-validation.md). Currently we don't restrict writing validation for `metadata`, which means people can have [additive validation constraints to `metadata` after the built-in validation](https://github.com/kubernetes/kubernetes/blob/74036c082aeec92c196a887460d2aec0c060cd8e/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/validator.go#L55-L64).  

For example, people can have additive name constraints (require CR `metadata.name` to be `default`) with the following schema:

```
  validation:
   openAPIV3Schema:
     description: Foo CRD for Testing
     properties:
       metadata:
         description: Metadata of Foo
         properties:
           name:
             description: Name of Foo
             pattern: default
             type: string
```

This brings us two questions: 

1. how do we [publish CR validation schema](https://github.com/kubernetes/kubernetes/pull/71192) if "metadata" is included in the schema: 
  - simplest option is to publish standard metav1.ObjectMeta schema info, and use any CRD-specified "metadata" properties as additive validation (just like in-tree types do, publishing standard ObjectMeta schema info, and implementing type-specific name validation internally)
  - alternatively, we can try merging user-specified metadata schema with the standard one. Technically people should only add extra validation to existing metadata properties (adding new fields to metadata won't work since they will be pruned by built-in validation first), but it's unclear to me how arbitrary people may want the metadata validation to be

2. longer term question is whether to allow people specifying metadata validation at all, or just allow a subset (like name validation) in CRD v1

/sig api-machinery
/area custom-resources
cc @apelisse @lavalamp @liggitt @mbohlool @smarterclayton @sttts

mbohlool:
>* simplest option is to publish standard metav1.ObjectMeta schema info, and use any CRD-specified "metadata" properties as additive validation (just like in-tree types do, publishing standard ObjectMeta schema info, and implementing type-specific name validation internally)
>* alternatively, we can try merging user-specified metadata schema with the standard one. Technically people should only add extra validation to existing metadata properties (adding new fields to metadata won't work since they will be pruned by built-in validation first), but it's unclear to me how arbitrary people may want the metadata validation to be

What is the difference between these two? I mean the end result is the same right? The only thing that can be added is validation. Nothing else works unless you are considering extensions?

[...]

PR #94379, opened 2020-08-31
apiextensions: allow metadata description

abursavich:
**What type of PR is this?**

/kind bug

When restricting any specification of metadata fields, the existing validation also restricts any description of the metadata itself, for which there doesn't seem to be any reason. While controller-tools has been updated to drop its metadata field specifications, it leaves any metadata description, which then causes validation to fail. The current workaround is to remove any documentation on Go struct metadata fields for custom resources, which is not ideal for users (e.g. [Prometheus](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#prometheus)).

N/A

Relates to https://github.com/kubernetes/kubernetes/issues/74620, https://github.com/kubernetes/kubernetes/pull/77653, https://github.com/kubernetes-sigs/controller-tools/issues/216, and https://github.com/kubernetes-sigs/controller-tools/pull/266

```release-note
NONE
```

```docs
- [KEP]: https://github.com/kubernetes/enhancements/blob/bb44066fbeaa571fb824c164742c37a289cbd608/keps/sig-api-machinery/20190425-structural-openapi.md
- [Other]: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema
```

### DIFF
--- a/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/schema/validation.go
+++ b/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/schema/validation.go
@@ -160,6 +160,7 @@ func validateStructuralInvariants(s *Structural, lvl level, fldPath *field.Path)
 			metadata.Properties = nil
 		}
 		metadata.Type = ""
+		metadata.Description = ""
 		metadata.Default.Object = nil // this is checked in API validation (and also tested)
 		if metadata.ValueValidation == nil {
 			metadata.ValueValidation = &ValueValidation{}
--- a/staging/src/k8s.io/apiextensions-apiserver/test/integration/validation_test.go
+++ b/staging/src/k8s.io/apiextensions-apiserver/test/integration/validation_test.go
@@ -1517,6 +1517,17 @@ properties:
 			expectedViolations: []string{},
 		},
 		{
+			desc: "metadata with description",
+			globalSchema: `
+type: object
+properties:
+ metadata:
+   description: 'Standard object metadata. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata'
+   type: object
+`,
+			expectedViolations: []string{},
+		},
+		{
 			desc: "metadata under junctors",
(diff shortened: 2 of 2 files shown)


abursavich:
cc: @sttts @DirectXMan12

apelisse:
That seems innocuous, but I'll let Stefan decide if it is.
/assign @sttts

abursavich:
ping

abursavich:
ping, @sttts

abursavich:
ping, @sttts

apelisse:
It looks like a validation change, which is typically a problem with HA clusters. Would we be OK with this if it has some sort of ratcheting?

liggitt:
Metadata field descriptions are not effective (the in-tree descriptions override when publishing the schema). I don't see a reason to allow descriptions specially when they will not be used. 

Additionally, relaxing validation in this way breaks compatibility with n-1 API servers, so if we were going to do this, it would have to be tolerated only on update and only if the existing object already had data specified this way for at least one release.

liggitt:
> The current workaround is to remove any documentation on Go struct metadata fields for custom resources, which is not ideal for users

Given the metadata descriptions published for users come from the in-tree metadata definition, can you give more details about why omitting them here is problematic?

abursavich:
> > The current workaround is to remove any documentation on Go struct metadata fields for custom resources, which is not ideal for users
> 
> Given the metadata descriptions published for users come from the in-tree metadata definition, can you give more details about why omitting them here is problematic?

Go structs field comments are a form of documentation and are used in other forms of generated type documentation (e.g [godoc](https://godoc.org/github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1#Prometheus) and [markdown](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#prometheus)). It's unfortunate that they lose information to appease the machinery. However, based on your feedback it seems like the issue is with controller-tools for including the comment/description in the CRD.

Issue #94275, opened 2020-08-27
kubectl server side diff errors due to field type changes


jpmsilva:
**What happened**:
_Spawned from #94121_

Running kubectl diff server side errors out due to the internal data types being different.

```
# kubectl diff -f test.yaml --server-side=true
Error from server: failed to create typed patch object: errors:
  .spec.limits[0].default.cpu: expected string, got &value.valueUnstructured{Value:1}
  .spec.limits[0].defaultRequest.cpu: expected string, got &value.valueUnstructured{Value:0.5}
```

**What you expected to happen**:
No error should be shown, and the diff (if any) should be shown instead.

**How to reproduce it (as minimally and precisely as possible)**:
Given the following test manifest:

```YAML
apiVersion: v1
kind: LimitRange
metadata:
  name: default-limit-range
spec:
  limits:
  - default:
      cpu: 1
      memory: 2Gi
    defaultRequest:
      cpu: 0.5
      memory: 1Gi
    max:
      memory: 8Gi
    type: Container
```

Apply it to the cluster:
```
# kubectl apply -f test.yaml
```

Run a diff server side:
```
# kubectl diff -f test.yaml --server-side=true
Error from server: failed to create typed patch object: errors:
  .spec.limits[0].default.cpu: expected string, got &value.valueUnstructured{Value:1}
  .spec.limits[0].defaultRequest.cpu: expected string, got &value.valueUnstructured{Value:0.5}
```

**Anything else we need to know?**:
Running `kubectl apply --dry-run=server -f ...` always detects incoming changes, which might be a consequence of this issue (I can't tell for sure):

```
# kubectl apply -f test.yaml --dry-run=server
limitrange/default-limit-range configured (server dry run)
```

This issue is related to #94121, where this behavior was first observed.

**Environment**:
- Kubernetes version (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.8", GitCommit:"9f2892aab98fe339f3bd70e3c470144299398ace", GitTreeState:"clean", BuildDate:"2020-08-13T16:12:48Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #94380, opened 2020-08-31
Ensure `kubectl diff --server-side` does not conflict with client-side apply

julianvmodesto:
**What type of PR is this?**

/kind bug

In 1.19, we made it so that upgrading from client-side `kubectl apply`
to server-side `kubectl apply --server-side` does not conflict.

This means that `kubectl diff --server-side` should work the same:
server-side apply diffing a resource managed originally with client-side apply
should not result in conflicts.

Related to #94275

```release-note
NONE
```


julianvmodesto:
It seems like this test works if it's reproducing the reported issue correctly.

Marking as cleanup for now since we should have this test updated.

/kind cleanup

PR #88337, opened 2020-02-19
kubectl/drain add support for custom pod filters

michaelgugino:
Currently, there is no way to supply custom pod filters to
exclude pods meeting arbitrary criteria when using drain
as a library.

This commit adds the ability for developers to add custom
filters to the drain operation when utilizing drain
as a library.

This commit exports certain types that were previously
private to allow for better code reuse.

This commit also adds appropriate unit tests.

**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes # https://github.com/kubernetes/enhancements/issues/1557

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #85571, opened 2019-11-23
kubectl/drain: add disable-eviction option

michaelgugino:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

**What this PR does / why we need it**:
Currently, if eviction is supported during a drain operation,
eviction is always used.

This commit allows the user to specify disabling eviction.
This is particularly useful when you wish to ignore
PodDisruptionBudgets after a normal drain has failed for
some time.

**Which issue(s) this PR fixes**:

Fixes https://github.com/kubernetes/kubernetes/issues/83307

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
kubectl/drain: add disable-eviction option.
Force drain to use delete, even if eviction is supported. This will bypass checking PodDisruptionBudgets, and should be used with caution.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #83906, opened 2019-10-14
Allow deletion of pending pods when using PDBS

michaelgugino:
Currently, if you have a PDB set, it is possible for
a pod stuck in pending state to be prevented from
deletion even though there are in fact enough healthy
replicas.

This commit allows pods in Pending state to be removed.

This commit also adds associated unit tests.

Fixes: #80389

**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:
Currently, if you have a PDB set, it is possible for
a pod stuck in pending state to be prevented from
deletion even though there are in fact enough healthy
replicas.
**Which issue(s) this PR fixes**:

Fixes: https://github.com/kubernetes/kubernetes/issues/80389

**Special notes for your reviewer**:
Here's where 'healthy' pods are counted: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/disruption/disruption.go#L675

**Does this PR introduce a user-facing change?**:

```release-note
When evicting, Pods in Pending state are removed without checking PDBs.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #94381, opened 2020-08-31
Allow deletion of unhealthy pods if enough healthy

### LINKED ISSUE #80389: PDBs using minUnavailable prevent draining unready pods
**What happened**:

We were unable to drain pods on an unhealthy node that were stuck in pod initializing. The disruption budget has a `minAvailable` of 2. We had two healthy pods and a pod stuck in initializing thus it's unhealthy. A drain on the node with the unhealthy pod stuck in initializing was prevented. 

The pods should have been evicted because they are in a not Ready state. 

Using a PDB with `minAvailable` attempt to drain a node with a pod in a not Ready state. 

The reason this is prevented is because the disruption controller determines the number of allowed disruptions by looking at [`currentHealthy - desiredHealthy`](https://github.com/kubernetes/kubernetes/blob/a07b027261383127494fc6d4e3ca2c544fb4a464/pkg/controller/disruption/disruption.go#L749-L752). When `minAvailable` is used your `desiredHealthy` count becomes that value, which in our case is 2. The `currentHealthy` count is 2 also because the 3rd pod is stuck in an initializing state on the bad node. This leaves the PDB with an allowed disruption of 0 and thus blocks the drain. 

The eviction logic in the storage only allows pods in a terminal state to bypass the PDB. 

https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/core/pod/storage/eviction.go#L124-L133

If they're not in a terminal state the eviction will only be allowed to occur if there are allowable disruptions available on the PDB.

It seems that the storage logic should allow draining pods that are not ready instead of just allowing terminal pods. 

/sig apps

michaelgugino:
**What type of PR is this?**

/kind bug

Currently, if you have a PDB with 0 disruptions
available and you attempt to evict a non-healthy
pod, the eviction request will always fail.  This
is because the eviction API does not currently
take in to account that the pod you are removing
is the unhealthy one.

This commit accounts for trying to evict an
unhealthy pod as long as there are enough healthy
pods to satisfy the PDB's requirements.  To
protect against race conditions, a ResourceVersion
constraint is enforced.  This will ensure that
the target pod does not go healthy and allow
any race condition to occur which might disrupt
too many pods at once.

This commit also exports a public function
from the disruption controller to ensure we
always calculate pod health in the same manner
as PDBs.  This will help prevent regressions.

Fixes #80389

yes

```release-note
Unhealthy pods covered by PDBs can be successfully evicted if enough healthy pods are available.
```

```docs

```

### DIFF
--- a/pkg/registry/core/pod/storage/eviction.go
+++ b/pkg/registry/core/pod/storage/eviction.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/apiserver/pkg/util/dryrun"
 	policyclient "k8s.io/client-go/kubernetes/typed/policy/v1beta1"
 	"k8s.io/client-go/util/retry"
+	podutil "k8s.io/kubernetes/pkg/api/pod"
 	api "k8s.io/kubernetes/pkg/apis/core"
 	"k8s.io/kubernetes/pkg/apis/policy"
 )
@@ -145,19 +146,18 @@ func (r *EvictionREST) Create(ctx context.Context, name string, obj runtime.Obje
 		}
 
 		// the PDB can be ignored, so delete the pod
-		deletionOptions := originalDeleteOptions.DeepCopy()
+		deleteOptions := originalDeleteOptions
+
 		// We should check if resourceVersion is already set by the requestor
 		// as it might be older than the pod we just fetched and should be
 		// honored.
 		if shouldEnforceResourceVersion(pod) && resourceVersionIsUnset(originalDeleteOptions) {
-			// Set deletionOptions.Preconditions.ResourceVersion to ensure we're not
+			// Set deleteOptions.Preconditions.ResourceVersion to ensure we're not
 			// racing with another PDB-impacting process elsewhere.
-			if deletionOptions.Preconditions == nil {
-				deletionOptions.Preconditions = &metav1.Preconditions{}
-			}
-			deletionOptions.Preconditions.ResourceVersion = &pod.ResourceVersion
+			deleteOptions = deleteOptions.DeepCopy()
+			setPreconditionsResourceVersion(deleteOptions, &pod.ResourceVersion)
 		}
-		_, _, err = r.store.Delete(ctx, eviction.Name, rest.ValidateAllObjectFunc, deletionOptions)
+		_, _, err = r.store.Delete(ctx, eviction.Name, rest.ValidateAllObjectFunc, deleteOptions)
 		if err != nil {
 			return err
 		}
@@ -181,6 +181,8 @@ func (r *EvictionREST) Create(ctx context.Context, name string, obj runtime.Obje
 
 	var rtStatus *metav1.Status
 	var pdbName string
+	updateDeletionOptions := false
+
 	err = func() error {
 		pdbs, err := r.getPodDisruptionBudgets(ctx, pod)
 		if err != nil {
@@ -201,6 +203,13 @@ func (r *EvictionREST) Create(ctx context.Context, name string, obj runtime.Obje
 
 		pdb := &pdbs[0]
 		pdbName = pdb.Name
+
+		// If the pod is not ready, it doesn't count towards healthy and we should not decrement
+		if !podutil.IsPodReady(pod) && pdb.Status.CurrentHealthy >= pdb.Status.DesiredHealthy && pdb.Status.DesiredHealthy > 0 {
+			updateDeletionOptions = true
+			return nil
+		}
+
 		refresh := false
 		err = retry.RetryOnConflict(EvictionsRetry, func() error {
 			if refresh {
@@ -232,16 +241,41 @@ func (r *EvictionREST) Create(ctx context.Context, name string, obj runtime.Obje
 		return rtStatus, nil
 	}
 
-	// At this point there was either no PDB or we succeeded in decrementing
+	// At this point there was either no PDB or we succeeded in decrementing or
+	// the pod was unready and we have enough healthy replicas
+
+	deleteOptions := originalDeleteOptions
+
+	// Set deleteOptions.Preconditions.ResourceVersion to ensure
+	// the pod hasn't been considered ready since we calculated
+	if updateDeletionOptions {
+		// Take a copy so we can compare to client-provied Options later.
+		deleteOptions = deleteOptions.DeepCopy()
+		setPreconditionsResourceVersion(deleteOptions, &pod.ResourceVersion)
+	}
 
 	// Try the delete
-	_, _, err = r.store.Delete(ctx, eviction.Name, rest.ValidateAllObjectFunc, originalDeleteOptions.DeepCopy())
+	_, _, err = r.store.Delete(ctx, eviction.Name, rest.ValidateAllObjectFunc, deleteOptions)
 	if err != nil {
+		if errors.IsConflict(err) && updateDeletionOptions &&
+			(originalDeleteOptions.Preconditions == nil || originalDeleteOptions.Preconditions.ResourceVersion == nil) {
+			// If we encounter a resource conflict error, we updated the deletion options to include them,
+			// and the original deletion options did not specify ResourceVersion, we send back
+			// TooManyRequests so clients will retry.
+			return nil, createTooManyRequestsError(pdbName)
+		}
 		return nil, err
 	}
 
 	// Success!
 	return &metav1.Status{Status: metav1.StatusSuccess}, nil
+}
+
+func setPreconditionsResourceVersion(deleteOptions *metav1.DeleteOptions, resourceVersion *string) {
+	if deleteOptions.Preconditions == nil {
+		deleteOptions.Preconditions = &metav1.Preconditions{}
+	}
+	deleteOptions.Preconditions.ResourceVersion = resourceVersion
 }
 
 // canIgnorePDB returns true for pod conditions that allow the pod to be deleted
@@ -268,16 +302,21 @@ func resourceVersionIsUnset(options *metav1.DeleteOptions) bool {
 	return options.Preconditions == nil || options.Preconditions.ResourceVersion == nil
 }
 
+func createTooManyRequestsError(name string) error {
+	// TODO(mml): Add a Retry-After header.  Once there are time-based
+	// budgets, we can sometimes compute a sensible suggested value.  But
+	// even without that, we can give a suggestion (10 minutes?) that
+	// prevents well-behaved clients from hammering us.
+	err := errors.NewTooManyRequests("Cannot evict pod as it would violate the pod's disruption budget.", 0)
+	err.ErrStatus.Details.Causes = append(err.ErrStatus.Details.Causes, metav1.StatusCause{Type: "DisruptionBudget", Message: fmt.Sprintf("The disruption budget %s is still being processed by the server.", name)})
+	return err
+}
+
 // checkAndDecrement checks if the provided PodDisruptionBudget allows any disruption.
 func (r *EvictionREST) checkAndDecrement(namespace string, podName string, pdb policyv1beta1.PodDisruptionBudget, dryRun bool) error {
 	if pdb.Status.ObservedGeneration < pdb.Generation {
-		// TODO(mml): Add a Retry-After header.  Once there are time-based
-		// budgets, we can sometimes compute a sensible suggested value.  But
-		// even without that, we can give a suggestion (10 minutes?) that
-		// prevents well-behaved clients from hammering us.
-		err := errors.NewTooManyRequests("Cannot evict pod as it would violate the pod's disruption budget.", 0)
-		err.ErrStatus.Details.Causes = append(err.ErrStatus.Details.Causes, metav1.StatusCause{Type: "DisruptionBudget", Message: fmt.Sprintf("The disruption budget %s is still being processed by the server.", pdb.Name)})
-		return err
+
+		return createTooManyRequestsError(pdb.Name)
 	}
 	if pdb.Status.DisruptionsAllowed < 0 {
 		return errors.NewForbidden(policy.Resource("poddisruptionbudget"), pdb.Name, fmt.Errorf("pdb disruptions allowed is negative"))
--- a/pkg/registry/core/pod/storage/BUILD
+++ b/pkg/registry/core/pod/storage/BUILD
@@ -14,6 +14,7 @@ go_test(
     ],
     embed = [":go_default_library"],
     deps = [
+        "//pkg/api/pod:go_default_library",
         "//pkg/apis/core:go_default_library",
         "//pkg/apis/policy:go_default_library",
         "//pkg/registry/registrytest:go_default_library",
--- a/pkg/registry/core/pod/storage/eviction_test.go
+++ b/pkg/registry/core/pod/storage/eviction_test.go
@@ -31,6 +31,7 @@ import (
 	genericapirequest "k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/apiserver/pkg/registry/rest"
 	"k8s.io/client-go/kubernetes/fake"
+	podapi "k8s.io/kubernetes/pkg/api/pod"
 	api "k8s.io/kubernetes/pkg/apis/core"
 	"k8s.io/kubernetes/pkg/apis/policy"
 )
@@ -219,6 +220,7 @@ func TestEvictionIngorePDB(t *testing.T) {
 		podName             string
 		expectedDeleteCount int
 		podTerminating      bool
+		prc                 *api.PodCondition
 	}{
 		{
 			name: "pdbs No disruptions allowed, pod pending, first delete conflict, pod still pending, pod deleted successfully",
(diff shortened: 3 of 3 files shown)


michaelgugino:
WIP: Need to invest in some unit tests here.

pkg/controller/disruption/disruption.go
```diff
@@ -688,6 +688,12 @@ func (dc *DisruptionController) getExpectedScale(pdb *policy.PodDisruptionBudget
 	return
 }
 
+// IsPodHealthy determines if an individual pod will be considered in
+// the health count of a PDB.
+func IsPodHealthy(pod *v1.Pod) bool {
```
michaelgugino, line 693: Could do this in a separate patch, if necessary.

michaelgugino:
@deads2k You helped review the previous eviction patch, would you mind taking a look at this one?

JoelSpeed (commented):
Looks good, just one question for clarification of the intention of the change here, otherwise I have nothing to add

pkg/controller/disruption/disruption.go
michaelgugino: No logic change, but wrapping the underlying implementation in a public function so it can be consumed in the eviction API.  This way if the healthy calculation for PDBs were to change in the future to something other than just 'podutil.IsPodReady(pod)' it will be picked up in those places.

pkg/registry/core/pod/storage/eviction.go
```diff
@@ -201,6 +206,15 @@ func (r *EvictionREST) Create(ctx context.Context, name string, obj runtime.Obje
 
 		pdb := &pdbs[0]
 		pdbName = pdb.Name
+
+		// If the pod is not ready, it doesn't count towards healthy and we should not
+		// decrement.  IsPodHealthy requires a v1Pod, so if we get some other type
+		// of pod, we cannot accurately determine pod health.
+		if validV1Pod && !disruption.IsPodHealthy(v1Pod) && pdb.Status.CurrentHealthy >= pdb.Status.DesiredHealthy && pdb.Status.DesiredHealthy > 0 {
+			updateDeletionOptions = true
```
elmiko, line 214: i'm not following this update, if we are about to return do we need to set this variable? (i didn't see a defer'd function)

elmiko (commented):
this makes sense and looks good to me, i just have a question about one of the clauses

pkg/registry/core/pod/storage/eviction.go
michaelgugino: This code is inside an anonymous function: https://github.com/kubernetes/kubernetes/pull/94381/files#diff-9644c646d1076728c605baed12a5813aR189

So, that's the code we're returning from.

elmiko: ahh, thanks! i knew i was missing something

elmiko (commented):
/lgtm

mortent:
I think this is an improvement over the existing behavior where all running but non-ready pods would be blocked from eviction no matter the value in the PDB. I'm still a bit concerned that this doesn't fully resolve the discrepancy between what is considered a healthy/ready pod in the disruption controller and the eviction service. A pod that is running but not ready will still not be counted as available by the disruption controller, but is still blocked from eviction.

This makes me wonder if the outcome here is the same as if we changed the criteria in both the disruption controller and in the eviction service so that any pod that is in the running phase is considered `available` and therefore eviction of the pod must check for PDBs?

michaelgugino:
@mortent 
> I'm still a bit concerned that this doesn't fully resolve the discrepancy between what is considered a healthy/ready pod in the disruption controller and the eviction service. 

That's what this line is intended to address: https://github.com/kubernetes/kubernetes/pull/94381/files#diff-af923020fb5800d949a2b1acd9117903R693

Exporting that function and then calling it from the eviction API ensures we're using the same logic to determine pod as healthy/disrupted.

> A pod that is running but not ready will still not be counted as available by the disruption controller, but is still blocked from eviction.

This is exactly what this patch set is intended to change.  Now, if a pod is not ready, it will be unblocked from eviction as long as there are enough *other* healthy pods.

> This makes me wonder if the outcome here is the same as if we changed the criteria in both the disruption controller and in the eviction service so that any pod that is in the running phase is considered `available` and therefore eviction of the pod must check for PDBs?

I'm not sure what you mean by this.  The current criteria is 'healthy' == 'Ready'.  If we made 'healthy' == 'Running' then that would undermine the effectiveness of PDBs.

For example, imagine you have data sharded across 3 pods.  pod 'A' goes down, the replacement pod 'A' comes up.  That pod is running, but not 'ready.'  It's not 'ready' because it's replicating the data from the other 2 hosts that are primary on itself.  If we treated pod 'A' as healthy and then evicted pod 'B', then we lose data.

mortent:
> That's what this line is intended to address: https://github.com/kubernetes/kubernetes/pull/94381/files#diff-af923020fb5800d949a2b1acd9117903R693
> 
> Exporting that function and then calling it from the eviction API ensures we're using the same logic to determine pod as healthy/disrupted.

I think that is fine. What I was thinking about is that a pod that is `running` but not `ready` will not be counted as available by the disruption controller, which would suggest it is not ready/healthy. But the eviction API will not evict a pod in the same state without checking against the PDB. So the pod is in some ways not healthy enough to be counted as available by the disruption controller, but at the same time we check against the very same PDB in order to allow eviction.

Even with this fix, it seems like situations as the one described in https://github.com/kubernetes/kubernetes/issues/72320 can still happen. In the end, this comes down to the question of when it is safe to evict a pod that is in CrashLoopBackoff, which is discussed at length in https://github.com/kubernetes/kubernetes/issues/80389 and https://github.com/kubernetes/kubernetes/pull/83906.

> 
> I'm not sure what you mean by this. The current criteria is 'healthy' == 'Ready'. If we made 'healthy' == 'Running' then that would undermine the effectiveness of PDBs.

Yeah, you are right. I didn't think this through.

> 
> For example, imagine you have data sharded across 3 pods. pod 'A' goes down, the replacement pod 'A' comes up. That pod is running, but not 'ready.' It's not 'ready' because it's replicating the data from the other 2 hosts that are primary on itself. If we treated pod 'A' as healthy and then evicted pod 'B', then we lose data.

Not sure I follow this. There is certainly an availability issue, but seems unavoidable as long as containers are crashlooping. Presumably any data that needs to be persisted across pod failures is on a persistent volume and can be mounted into a replacement pod when it comes up.

Anyway, this PR already has lgtm, so we need an approver to take a look at this.
/assign @janetkuo @soltysh

michaelgugino:
@mortent 
> Even with this fix, it seems like situations as the one described in #72320 can still happen. In the end, this comes down to the question of when it is safe to evict a pod that is in CrashLoopBackoff, which is discussed at length in #80389 and #83906.

This patch would allow you to evict a single pod that is in crash loop backoff if there are enough other healthy pods covered by the PDB.  Blocking eviction when there are not enough healthy pods is pretty much the entire point of the eviction API, so if all pods are in crash loop backoff, I would expect eviction to always fail.  We can't automatically know if the crashloop backoff will happen indefinitely or if it's just a temporary condition (eg, on start, check some web service, if down, exit 1, repeat).

Generally speaking, PDBs are a single tool in a large tool set.  Administrators and/or application owners have some responsibility to ensure their nodes aren't crashloop-backoff wastelands if they want to use the eviction API.  This might include monitoring and alerting, or some other controller that is smart enough to do the right thing.  I most cases eviction API refers to someone draining a node, so it's trivial to look at the output of the drain command and just delete the broken pods.

> Presumably any data that needs to be persisted across pod failures is on a persistent volume and can be mounted into a replacement pod when it comes up.

I agree with this, unless the pod is something like a ceph OSD, and 'evicting' pod A is part of the process of deleting the node in which it's running on and it's underlying storage.  Anyway, in the general case, we can't assume what a pod is doing, and I think the current guidelines to count healthy==ready make sense.

mortent:
> This patch would allow you to evict a single pod that is in crash loop backoff if there are enough other healthy pods covered by the PDB. Blocking eviction when there are not enough healthy pods is pretty much the entire point of the eviction API, so if all pods are in crash loop backoff, I would expect eviction to always fail. We can't automatically know if the crashloop backoff will happen indefinitely or if it's just a temporary condition (eg, on start, check some web service, if down, exit 1, repeat).

That is fair. But we can use ResourceVersion to make sure we don't evict pods that have become Ready, and controllers will usually replace any evicted pods. If a pod isn't owned by a controller, eviction will be problematic in all situations.

> Generally speaking, PDBs are a single tool in a large tool set. Administrators and/or application owners have some responsibility to ensure their nodes aren't crashloop-backoff wastelands if they want to use the eviction API. This might include monitoring and alerting, or some other controller that is smart enough to do the right thing. I most cases eviction API refers to someone draining a node, so it's trivial to look at the output of the drain command and just delete the broken pods.

This is true for running a small number of clusters. At scale, looking at the output of drain commands and manually delete pods quickly becomes impractical. I think the eviction API is used by cluster autoscalers where there aren't anyone who can manually look at this. It is also an important tool for doing node upgrades, which again will often be performed by automation and usually by a different team/company than the owners of the workloads.
Whenever possible, we should try to avoid situations that require manual intervention.

michaelgugino:
> That is fair. But we can use ResourceVersion to make sure we don't evict pods that have become Ready, and controllers will usually replace any evicted pods. If a pod isn't owned by a controller, eviction will be problematic in all situations.

My concern isn't necessarily draining the individual pod, but if there is an automated process that's going to drain several nodes at once, or as I mentioned, serially drain nodes resulting in multiple pods from a set being evicted.  If a pod isn't owned by a controller, the user can still elect to specify a PDB.  They might have some off-cluster component that is watching their pods and ensuring proper amount and deployment, or it might just be a development pod, etc.

> Whenever possible, we should try to avoid situations that require manual intervention.

I completely agree.  My primary use case for this is automated systems, especially cluster-api.  Projects/administrators have more tools at their disposal than just running drain with default options.  After a timeout, they could consider using https://github.com/kubernetes/kubernetes/pull/88337 to specify custom pod filters and https://github.com/kubernetes/kubernetes/pull/85571 to force delete instead of eviction.  For example, after running normal drain for some length of time, they write a filter to exclude all pods that aren't in a state they wish to ignore (eg, crashloop backoff), and specify the disable-eviction option.  This will result in just crash-loop backoff pods being removed.

In any case, a node being blocked from drain by an automated system is no worse off in this scenario as currently, if there's a healthy pod you want to drain and disrupted pods elsewhere, you won't be able to.  Inevitably, there are going to be situations where administrators have to intervene or use brute-force and potentially disrupt applications.  We can't solve all those scenarios in the eviction API, IMO.

janetkuo (approved):

michaelgugino:
@smarterclayton the merge bot requires your approval here, PTAL.

pkg/registry/core/pod/storage/eviction.go
```diff
@@ -34,7 +35,9 @@ import (
 	policyclient "k8s.io/client-go/kubernetes/typed/policy/v1beta1"
 	"k8s.io/client-go/util/retry"
 	api "k8s.io/kubernetes/pkg/apis/core"
+	corev1 "k8s.io/kubernetes/pkg/apis/core/v1"
 	"k8s.io/kubernetes/pkg/apis/policy"
+	"k8s.io/kubernetes/pkg/controller/disruption"
```
smarterclayton, line 40: API should not take dependencies on controllers, it should be to a third party package.  The method you carved off should be in podutil, not in controller, and both places should reference it.

pkg/registry/core/pod/storage/eviction.go
```diff
@@ -232,18 +246,48 @@ func (r *EvictionREST) Create(ctx context.Context, name string, obj runtime.Obje
...
 	// Try the delete
-	_, _, err = r.store.Delete(ctx, eviction.Name, rest.ValidateAllObjectFunc, originalDeleteOptions.DeepCopy())
+	_, _, err = r.store.Delete(ctx, eviction.Name, rest.ValidateAllObjectFunc, deletionOptions.DeepCopy())
 	if err != nil {
+		if errors.IsConflict(err) && updateDeletionOptions &&
+			(originalDeleteOptions.Preconditions == nil || originalDeleteOptions.Preconditions.ResourceVersion == nil) {
+			// If we encounter a resource conflict error, we updated the deletion options to include them,
+			// and the original deletion options did not specify ResourceVersion, we send back
+			// TooManyRequests so clients will retry.
+			return nil, createTooManyRequestsError(pdbName)
```
smarterclayton, line 268: Can you explain from history *why* we were returning TooManyRequests in this code base?  Was it a quick workaround to trigger retry?  Or was there a longer discussion about the appropriate response code.  I don't have confidence from the comments here that it's the right error code, and I'd want to be confident before we spread it to a second spot (and update the comments such that it's explicitly clear that no other status code is appropriate, if so)

pkg/registry/core/pod/storage/eviction.go
smarterclayton: Also, I'm not really a fan of sharing a wrapper when the underlying condition is more clear.  We don't have a "Healthy" concept in the API.  We have "Ready".  Both controller and API should be defined in terms of "Ready", or have local conditions.  But creating a synthetic API "healthy" in shared code is something I generally oppose because it is not modelled as part of our API.

pkg/registry/core/pod/storage/eviction.go
```diff
@@ -181,6 +181,11 @@ func (r *EvictionREST) Create(ctx context.Context, name string, obj runtime.Obje
 
 	var rtStatus *metav1.Status
 	var pdbName string
+	updateDeletionOptions := false
+
+	v1Pod := &v1.Pod{}
+	validV1Pod := getV1Pod(pod, v1Pod)
```
smarterclayton, line 187: Not really a fan of this.  Our API internal endpoints should work with internal APIs.  We have `pkg/api/pod.IsPodReady` for internal pods and it is more appropriate to use than doing the conversion.

pkg/registry/core/pod/storage/eviction.go
michaelgugino: I'll just remove that change, podutil.IsPodReady(pod) is already being called, so I'll just use that in the eviction api.

pkg/registry/core/pod/storage/eviction.go
michaelgugino: This is a trade off.  If we want to use the identical code that PDBs use to evaluate 'healthy' then we do this.  If we don't want to use the identical code, we can use the internal pod util as you highlight.

Either option is fine for me.

pkg/registry/core/pod/storage/eviction.go
michaelgugino: I agree with not using the terminology 'healthy' but that's what PDBs expose: https://godoc.org/k8s.io/api/policy/v1beta1#PodDisruptionBudgetStatus

I can't find any specific wording to indicate that Healthy == Ready.

pkg/registry/core/pod/storage/eviction.go
michaelgugino: Here's the original implementing PR, this comment makes me believe that we chose an error number out of a hat: https://github.com/kubernetes/kubernetes/pull/31033#discussion_r75573684

Current state of the world as far as kubectl/drain is that while trying to evict a pod, if I get a 429, that indicates the PDB did not allow eviction at this time, retry.

We are sending 429 here to emulate a blocking PDB (eg, PDB with 0 disruptions available) as we do above.  This is because if we fail the ResourceVersion constraint here, we are no longer in position to evaluate the PDBs.

I think signaling the client to optionally retry is the most sensible option here, and fits the current behavior: If I had 0 disruptions left (because the pod I'm trying to evict is already disrupted) we send back a 429.

smarterclayton: Thinking about the status codes we've assigned in the past (as the rough originator of most of the error stanzas I probably should create on separately for this if we will use it in the future), there's a couple that are close but not equivalent:

If you specify `generateName` and we keep getting collisions - the server generates a random name and then attempts to create the object in etcd, which could fail because someone else could choose that name - after 10 we return ServerTimeout

```
// NewServerTimeout returns an error indicating the requested action could not be completed due to a
// transient error, and the client should try again.
```

Patch retries forever, as does a guaranteed update.

smarterclayton: I guess 429 is fine for now, I can't really think of a better one and I'd rather not couple this change to to a more dramatic client behavior change.

pkg/registry/core/pod/storage/eviction.go
```diff
@@ -232,18 +239,41 @@ func (r *EvictionREST) Create(ctx context.Context, name string, obj runtime.Obje
 		return rtStatus, nil
 	}
 
-	// At this point there was either no PDB or we succeeded in decrementing
+	// At this point there was either no PDB or we succeeded in decrementing or
+	// the pod was unready and we have enough healthy replicas
+
+	deletionOptions := originalDeleteOptions.DeepCopy()
```
smarterclayton, line 245: I would move this copy operation into the if block and rename it to `deleteOptions`.

```
deleteOptions := originalDeleteOptions

if updateDeleteOptions {
  deleteOptions = deleteOptions.DeepCopy()
  setPreconditionsResourceVersion(...)
}
```

No need to do a deep copy except when you are explicitly changing the object.

pkg/registry/core/pod/storage/eviction.go
```diff
@@ -232,18 +239,41 @@ func (r *EvictionREST) Create(ctx context.Context, name string, obj runtime.Obje
...
+	// the pod hasn't been considered ready since we calculated
+	if updateDeletionOptions {
+		setPreconditionsResourceVersion(deletionOptions, &pod.ResourceVersion)
+	}
 
 	// Try the delete
-	_, _, err = r.store.Delete(ctx, eviction.Name, rest.ValidateAllObjectFunc, originalDeleteOptions.DeepCopy())
+	_, _, err = r.store.Delete(ctx, eviction.Name, rest.ValidateAllObjectFunc, deletionOptions.DeepCopy())
 	if err != nil {
+		if errors.IsConflict(err) && updateDeletionOptions &&
```
smarterclayton, line 256: Is this if block tested via a unit test already?  If not it would be good to add one now.

michaelgugino: All new logic in this PR is covered by unit testing.  Here's the coverage HTML output if you don't want to run coverage locally: https://gist.github.com/michaelgugino/9d20be02b1f0c4810893c6e81963eaf8

smarterclayton:
/lgtm
/approve

Thanks, this is the right change (PDB should not block in scenarios where the app is already effectively unhealthy).

vpnachev:
Is this PR cherry-picked? I think no, but it would be nice see it in v1.17+.

PR #93581, opened 2020-07-30
Clean up in port mapping functionality

SergeyKanzhelev:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
This PR clears up the confusion from the https://github.com/kubernetes/kubernetes/pull/80470/files#r375015375 Name of the port mapping is not being used in CNI interfaces and only needed for deduplication of entries.

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

- PR where the hostport functionality was removed from DockerShim: https://github.com/kubernetes/kubernetes/pull/62212
- `GenerateRunContainerOptions` do not call into `MakePortMappings ` because ports are configured sandbox-wide. So port mapping is not used for container set up.

De-duplication logic is not strictly required and it's faulty as it allows duplicated entries with unique names, but same ports. But removing it is not a clean up type PR - it may break some back compat. So leaving it for a separate PR.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs
NONE
```

PR #62212, opened 2018-04-06
Clean up the network plugin code

yujuhong:
Moves the `pkg/kubelet/network` package down to `dockershim` since network
setup is done by the runtime. Also remove various interfaces/code added only for
backward compatibility with non-CRI runtimes, which are now removed from the
codebase.

**Release note**:

```release-note
NONE
```

Issue #35457, opened 2016-10-24
[CRI] Add HostPort Support for Dockershim


freehan:
#34780 migrates the network logic into dockershim without adding hostport support.  `GetPodHostportMapping` method implemented by dockershim returns empty. This interface is a hook for network plugins to retrieve local pod hostport information, so that network plugins can implement the hostports. Need a way to retrieve hostport mapping from running pods in dockershim. This can be achieved by parsing container labels or caching.

yujuhong:
I spoke to @bprashanth offline. In the long run, we want CNI to support host ports directly. Right now, we can either 1) adapt the the existing batch sync implementation and checkpoint the ports used by containers (via labels?) as suggested by @freehan initially, or 2) choose not to support host ports and wait for a better fix (favored by @bprashanth).

Just wanted to point out that host port support is less important for an experimental feature (CRI) in the 1.5 release. However, if we are going to wait for CNI to support this, it could potentially block kubelet to switch to the CRI implementation as the default in the near future.

@bprashanth @thockin, can the network team remove the milestone if we've decided to drop host port support in 1.5? Thanks!

freehan:
ref: #31307

freehan:
Merged https://github.com/kubernetes/kubernetes/pull/41223

PR #94382, opened 2020-08-31
dockershim hostport manager should use HostIP

aojea:
/kind bug

the hostport manager was not taking into consideration the hostIP
when binding the socket of the hostPort, causing that the same
HostPort can not be used with different IP addresses.

```release-note
fix a bug where the hostport manager used by dockershim didn't allow to bind the same HostPort on different HostIPs
```

### DIFF
--- a/pkg/kubelet/dockershim/network/hostport/hostport_manager.go
+++ b/pkg/kubelet/dockershim/network/hostport/hostport_manager.go
@@ -152,10 +152,17 @@ func (hm *hostportManager) Add(id string, podPortMapping *PodPortMapping, natInt
 
 		// DNAT to the podIP:containerPort
 		hostPortBinding := net.JoinHostPort(podIP, strconv.Itoa(int(pm.ContainerPort)))
-		writeLine(natRules, "-A", string(chain),
-			"-m", "comment", "--comment", fmt.Sprintf(`"%s hostport %d"`, podFullName, pm.HostPort),
-			"-m", protocol, "-p", protocol,
-			"-j", "DNAT", fmt.Sprintf("--to-destination=%s", hostPortBinding))
+		if pm.HostIP == "" || pm.HostIP == "0.0.0.0" || pm.HostIP == "::" {
+			writeLine(natRules, "-A", string(chain),
+				"-m", "comment", "--comment", fmt.Sprintf(`"%s hostport %d"`, podFullName, pm.HostPort),
+				"-m", protocol, "-p", protocol,
+				"-j", "DNAT", fmt.Sprintf("--to-destination=%s", hostPortBinding))
+		} else {
+			writeLine(natRules, "-A", string(chain),
+				"-m", "comment", "--comment", fmt.Sprintf(`"%s hostport %d"`, podFullName, pm.HostPort),
+				"-m", protocol, "-p", protocol, "-d", pm.HostIP,
+				"-j", "DNAT", fmt.Sprintf("--to-destination=%s", hostPortBinding))
+		}
 	}
 
 	// getHostportChain should be able to provide unique hostport chain name using hash
@@ -312,6 +319,8 @@ func (hm *hostportManager) closeHostports(hostportMappings []*PortMapping) error
 				continue
 			}
 			delete(hm.hostPortMap, hp)
+		} else {
+			klog.V(5).Infof("host port %s does not have an open socket", hp.String())
 		}
 	}
 	return utilerrors.NewAggregate(errList)
@@ -324,7 +333,7 @@ func (hm *hostportManager) closeHostports(hostportMappings []*PortMapping) error
 // WARNING: Please do not change this function. Otherwise, HostportManager may not be able to
 // identify existing iptables chains.
 func getHostportChain(id string, pm *PortMapping) utiliptables.Chain {
-	hash := sha256.Sum256([]byte(id + strconv.Itoa(int(pm.HostPort)) + string(pm.Protocol)))
+	hash := sha256.Sum256([]byte(id + strconv.Itoa(int(pm.HostPort)) + string(pm.Protocol) + pm.HostIP))
 	encoded := base32.StdEncoding.EncodeToString(hash[:])
 	return utiliptables.Chain(kubeHostportChainPrefix + encoded[:16])
 }
--- a/pkg/kubelet/dockershim/network/hostport/hostport_manager_test.go
+++ b/pkg/kubelet/dockershim/network/hostport/hostport_manager_test.go
@@ -35,6 +35,7 @@ func TestOpenCloseHostports(t *testing.T) {
 		podPortMapping *PodPortMapping
 		expectError    bool
 	}{
+		// no portmaps
 		{
 			&PodPortMapping{
 				Namespace: "ns1",
@@ -42,6 +43,7 @@ func TestOpenCloseHostports(t *testing.T) {
 			},
 			false,
 		},
+		// allocate port 80/TCP, 8080/TCP and 443/TCP
 		{
 			&PodPortMapping{
 				Namespace: "ns1",
@@ -54,6 +56,7 @@ func TestOpenCloseHostports(t *testing.T) {
 			},
 			false,
 		},
+		// fail to allocate port previously allocated 80/TCP
 		{
 			&PodPortMapping{
 				Namespace: "ns1",
@@ -64,6 +67,7 @@ func TestOpenCloseHostports(t *testing.T) {
 			},
 			true,
 		},
+		// fail to allocate port previously allocated 8080/TCP
 		{
 			&PodPortMapping{
 				Namespace: "ns1",
@@ -75,6 +79,7 @@ func TestOpenCloseHostports(t *testing.T) {
 			},
 			true,
 		},
+		// allocate port 8081/TCP
 		{
 			&PodPortMapping{
 				Namespace: "ns1",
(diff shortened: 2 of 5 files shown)


aojea:
/assign @thockin @danwinship @dcbw 

I need to check if there is some job exercising this code with e2e tests, lot of jobs were move to containerd and (I think) using the CNI portmap plugin (at least the KIND ones)

thockin:
I am frankly surprised this works at all in the scheduler - does it?  Do we really want it to?  What's the use case?  This seems verrrrry subtle.

aojea:
> I am frankly surprised this works at all in the scheduler - does it? Do we really want it to? What's the use case? This seems verrrrry subtle.

There is a conformance test covering exactly this use case
https://github.com/kubernetes/kubernetes/blob/62b8faed4042e9f1a9c1e895ad1c36bb473eb6ee/test/e2e/scheduling/predicates.go#L654-L660

It seems that works (at least in CRIO https://github.com/cri-o/cri-o/pull/4130#issue-475492178) because, after the first attempt to allocate the HostPort, it removes ALL sockets listening in that port, so the second attempt works

aojea:
I can't find any job in the CI running this code :/

I was reading the issue tracking the hostport implementation #35457, and it seems that this was temporary until the CNI plugins implements the portmapping,

@thockin  do you think that does not make sense to pursue this and let it as is, fact is nobody noticed this based on a quick github search.

If we don fix it here., @danwinship I can make it a package in crio instead of vendoring it

danwinship:
> do you think that does not make sense to pursue this and let it as is, fact is nobody noticed this based on a quick github search.

Well, besides the fact that we know some people are vendoring this code, there's also the fact that it means that people using kubernetes with dockershim get non-conforming behavior. The fact that we don't currently run any CI jobs in that configuration doesn't change the fact that it's broken.

The fact that the conformance test fails to accurately test the behavior it's supposed to be testing is... interesting. We should probably fix the conformance test in 1.20 (by making it check that the first hostport still connects to the first pod after the second pod has been created) but not backport the fix, so as to not retroactively de-conform systems that previously passed conformance in earlier releases. At least, I think that's the right approach?

pkg/kubelet/dockershim/network/hostport/hostport_manager_test.go
```diff
@@ -106,13 +136,18 @@ func TestOpenCloseHostports(t *testing.T) {
 		execer:      exec.New(),
 	}
 
+	// open all hostports defined in the test cases
 	for _, tc := range openPortCases {
 		mapping, err := manager.openHostports(tc.podPortMapping)
+		for hostport, socket := range mapping {
+			manager.hostPortMap[hostport] = socket
```
danwinship, line 143: Why is this needed here? This seems like something that should either happen in the actual manager code, or else not be needed.

If it's just for the counting below, then that's really only testing the behavior of the test program, not the behavior of the manager...

pkg/kubelet/dockershim/network/hostport/hostport_manager_test.go
```diff
@@ -372,6 +510,8 @@ func TestHostportManager(t *testing.T) {
 	for _, port := range portOpener.mem {
 		assert.EqualValues(t, true, port.closed)
 	}
+	// Clear all elements in hostPortMap
+	assert.Zero(t, len(manager.hostPortMap))
```
danwinship, line 514: that's not clearing the elements, it's asserting that someone else already did

aojea: yeah, I wonder what I was thinking :)

pkg/kubelet/dockershim/network/hostport/hostport_manager_test.go
aojea: this was actually a bug in the test.
The same loop to add the open sockets to the hostPortMap map does happen in the manager.Add() method, https://github.com/kubernetes/kubernetes/blob/12fc5e50b9394cc3fa498b24a96c894d96b3bff2/pkg/kubelet/dockershim/network/hostport/hostport_manager.go#L107-L114

We need this because later we test the `closeHostports()` method that iterates over that map https://github.com/kubernetes/kubernetes/blob/12fc5e50b9394cc3fa498b24a96c894d96b3bff2/pkg/kubelet/dockershim/network/hostport/hostport_manager.go#L310-L327
in 
```
	for _, tc := range closePortCases {
		err := manager.closeHostports(tc.portMappings)
```

if we don't populate the map the closeHostports will always succeed because it's map. 
I've also added the counting below as a safety check and to make it easier to debug

pkg/kubelet/dockershim/network/hostport/hostport_manager_test.go
```diff
@@ -35,13 +35,15 @@ func TestOpenCloseHostports(t *testing.T) {
 		podPortMapping *PodPortMapping
 		expectError    bool
 	}{
+		// no portmaps
```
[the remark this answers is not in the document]
aojea: Naming is not useful here because the test cases are not independent. Personally, I don't like using this test structure but didn't want to change the whole test,so I carried over with current status. At least the comments explain better what is supposed to happen

thockin:
I did some quick tests on my own cluster and, shock, it does seem to work.
I dislike this very much.  It seems like a great opportunity to make a
mistake (honestly, hostIP at all seems poorly thought out but that dates to
2014, sigh).

So the choice is really a) fix it in dockershim to match containerd or b)
don't allow same-port-different-IP between pods.

I want to say B, but I suspect the answer is A.  Can we document this
behavior on the API types.go, please?

pkg/kubelet/dockershim/network/hostport/hostport_manager.go
```diff
@@ -152,10 +152,17 @@ func (hm *hostportManager) Add(id string, podPortMapping *PodPortMapping, natInt
 
 		// DNAT to the podIP:containerPort
 		hostPortBinding := net.JoinHostPort(podIP, strconv.Itoa(int(pm.ContainerPort)))
-		writeLine(natRules, "-A", string(chain),
-			"-m", "comment", "--comment", fmt.Sprintf(`"%s hostport %d"`, podFullName, pm.HostPort),
-			"-m", protocol, "-p", protocol,
-			"-j", "DNAT", fmt.Sprintf("--to-destination=%s", hostPortBinding))
+		if pm.HostIP == "" || pm.HostIP == "0.0.0.0" || pm.HostIP == "::" {
```
andrewsykim, line 155: Does kubelet know to use iptables for IPv6 if HostIP is `::` here?

aojea: kubelet is ALWAYS single stack as today, and its IP family depends on the nodeIP, 

https://github.com/kubernetes/kubernetes/blob/c236285708d43ad7535a37cbff41fa380afc199d/pkg/kubelet/kubelet.go#L468-L473

The hostport_manager has its own iptables handler, but depends on the kubelet for some rules, however, due to this https://github.com/kubernetes/kubernetes/issues/82125 , those rules are added by kube-proxy

Now that you mention it, without kube-proxy this won't work for dual-stack @danwinship

aojea:
> I want to say B, but I suspect the answer is A. Can we document this behavior on the API types.go, please?

I guess it is  https://github.com/kubernetes/kubernetes/blob/adac1bfb8b52def074022e0a3b07740fdc1a7adf/pkg/apis/core/types.go#L1762-L1764

pkg/kubelet/dockershim/network/hostport/hostport_manager_test.go
```diff
@@ -319,24 +439,41 @@ func TestHostportManager(t *testing.T) {
...
+		`:KUBE-HP-TUKTZ736U5JD5UTK - [0:0]`: true,
+		`:KUBE-HP-CAAJ45HDITK7ARGM - [0:0]`: true,
+		`:KUBE-HP-WFUNFVXVDLD5ZVXN - [0:0]`: true,
+		`:KUBE-HP-4MFWH2F2NAOMYD6A - [0:0]`: true,
+		"-A KUBE-HOSTPORTS -m comment --comment \"pod3_ns1 hostport 8443\" -m tcp -p tcp --dport 8443 -j KUBE-HP-WFBOALXEP42XEMJK":                        true,
+		"-A KUBE-HOSTPORTS -m comment --comment \"pod1_ns1 hostport 8081\" -m udp -p udp --dport 8081 -j KUBE-HP-63UPIDJXVRSZGSUZ":                        true,
+		"-A KUBE-HOSTPORTS -m comment --comment \"pod1_ns1 hostport 8080\" -m tcp -p tcp --dport 8080 -j KUBE-HP-IJHALPHTORMHHPPK":                        true,
+		"-A KUBE-HOSTPORTS -m comment --comment \"pod1_ns1 hostport 8083\" -m sctp -p sctp --dport 8083 -j KUBE-HP-XU6AWMMJYOZOFTFZ":                      true,
+		"-A KUBE-HOSTPORTS -m comment --comment \"pod5_ns5 hostport 8888\" -m tcp -p tcp --dport 8888 -j KUBE-HP-TUKTZ736U5JD5UTK":                        true,
+		"-A KUBE-HOSTPORTS -m comment --comment \"pod5_ns5 hostport 8888\" -m tcp -p tcp --dport 8888 -j KUBE-HP-CAAJ45HDITK7ARGM":                        true,
```
aojea, line 451: actually this does not seem to work as expected, it should load balance between the different backend HostIPs

aojea:
I just realized that the hostport_ syncer code was removed recently 
https://github.com/kubernetes/kubernetes/pull/93581

In the kubernetes repo only the kubenet seems to use this code :thinking:

SergeyKanzhelev:
> I just realized that the hostport_ syncer code was removed recently
> #93581
> 
> In the kubernetes repo only the kubenet seems to use this code 🤔

My PR was just a follow up clean up after this one: https://github.com/kubernetes/kubernetes/pull/62212 I wasn't sure about hostport functionality and whether it is actually used.

One more discussion point here - there is a leftover of duplicates suppression logic for [nodeport mappings](https://github.com/kubernetes/kubernetes/blob/19064f47389429f8360e1d6684bef3d770db9368/pkg/kubelet/container/helpers.go#L324-L345). Is it needed in kubelet if CNI should be responsible for setting it all up?

aojea:
> One more discussion point here - there is a leftover of duplicates suppression logic for [nodeport mappings](https://github.com/kubernetes/kubernetes/blob/19064f47389429f8360e1d6684bef3d770db9368/pkg/kubelet/container/helpers.go#L324-L345). Is it needed in kubelet if CNI should be responsible for setting it all up?

@thockin I think that this is only being used by kubenet in k8s, the rest of the imports are just for the type `[]*hostport.PortMapping`

is kubenet going to move out of the kubelet?

PR #41234, opened 2017-02-10
Enforce Node Allocatable via cgroups

vishh:
This PR enforces node allocatable across all pods using a top level cgroup as described in https://github.com/kubernetes/community/pull/348

This PR also provides an option to enforce `kubeReserved` and `systemReserved` on user specified cgroups. 

This PR will by default make kubelet create top level cgroups even if `kubeReserved` and `systemReserved` is not specified and hence `Allocatable = Capacity`.

```release-note
New Kubelet flag `--enforce-node-allocatable` with a default value of `pods` is added which will make kubelet create a top level cgroup for all pods to enforce Node Allocatable. Optionally, `system-reserved` & `kube-reserved` values can also be specified separated by comma to enforce node allocatable on cgroups specified via `--system-reserved-cgroup` & `--kube-reserved-cgroup` respectively. Note the default value of the latter flags are "".
This feature requires a **Node Drain** prior to upgrade failing which pods will be restarted if possible or terminated if they have a `RestartNever` policy.
```

cc @kubernetes/sig-node-pr-reviews @kubernetes/sig-node-feature-requests 

TODO:

- [x] Adjust effective Node Allocatable to subtract hard eviction thresholds
- [x] Add unit tests
- [x] Complete pending e2e tests
- [x] Manual testing
- [x] Get the proposal merged

@dashpole is working on adding support for evictions for enforcing Node allocatable more gracefully. That work will show up in a subsequent PR for v1.6

PR #94390, opened 2020-09-01
More testcases for GOAWAY

### LINKED ISSUE #91131: GOAWAY may cause client error when client request server in high concurrency
**What happened**:

We enabled [GOAWAY](https://github.com/kubernetes/kubernetes/pull/88567) feature in our large-scale cluster, and we found client(Kubelet) may got error(500 http code response) like `request declared a Content-Length of N but only wrote 0 bytes` after received GOAWAY frame. 

I had already re-produce this bug and open an [issue](https://github.com/golang/go/issues/39086) in golang repo. So far I think this bug only occur with a small probability when client request server in high concurrency use a same httpclient instance, and it's a golang sdk bug in http client.

**Before I solved this issue, I suggest do NOT enable GOAWAY(`--goaway-chance=0`) feature. As it's a golang sdk bug(currently I known and doubt) in http client, revert GOAWAY feature(codes) is not necessary.**

answer1991:
**What type of PR is this?**

/kind cleanup

1. Add a testcase(`TestGOAWAYConcurrency`) to check if GOAWAY affect concurrency requests
2. Refactor GOAWAY testcases

Fixes #91131

1. Golang version `go1.14.7 linux/amd64`(has not fixed GOAWAY issue), testcase `TestGOAWAYConcurrency` failed with error like:
```
--- FAIL: TestGOAWAYConcurrency (5.19s)
    goaway_test.go:487: failed to request "/post", err: expect response status code: 200, but got: 500. response body: request declared a Content-Length of 5 but only wrote 0 bytes
    goaway_test.go:487: failed to request "/post", err: expect response status code: 200, but got: 500. response body: request declared a Content-Length of 5 but only wrote 0 bytes
    goaway_test.go:487: failed to request "/post", err: expect response status code: 200, but got: 500. response body: request declared a Content-Length of 5 but only wrote 0 bytes
    goaway_test.go:487: failed to request "/post", err: expect response status code: 200, but got: 500. response body: request declared a Content-Length of 5 but only wrote 0 bytes
```

2. Golang version  `go1.15 linux/amd64`, testcase `TestGOAWAYConcurrency` succeed.

```release-note
None
```

```docs
None
```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/server/filters/goaway_test.go
+++ b/staging/src/k8s.io/apiserver/pkg/server/filters/goaway_test.go
@@ -17,12 +17,18 @@ limitations under the License.
 package filters
 
 import (
+	"bytes"
+	"context"
 	"crypto/tls"
+	"fmt"
 	"io"
+	"io/ioutil"
 	"math/rand"
 	"net"
 	"net/http"
 	"net/http/httptest"
+	"reflect"
+	"sync"
 	"testing"
 	"time"
 
@@ -85,25 +91,34 @@ func TestProbabilisticGoawayDecider(t *testing.T) {
 	}
 }
 
-// TestClientReceivedGOAWAY tests the in-flight watch requests will not be affected and new requests use a
-// connection after client received GOAWAY, and server response watch request with GOAWAY will not break client
-// watching body read.
-func TestClientReceivedGOAWAY(t *testing.T) {
-	const (
-		urlNormal          = "/normal"
-		urlWatch           = "/watch"
-		urlGoaway          = "/goaway"
-		urlWatchWithGoaway = "/watch-with-goaway"
-	)
+const (
+	urlGet             = "/get"
+	urlPost            = "/post"
+	urlWatch           = "/watch"
+	urlGetWithGoaway   = "/get-with-goaway"
+	urlPostWithGoaway  = "/post-with-goaway"
+	urlWatchWithGoaway = "/watch-with-goaway"
+)
 
-	const (
-		// indicate the bytes watch request will be sent
-		// used to check if watch request was broke by GOAWAY
-		watchExpectSendBytes = 5
-	)
+var (
+	// responseBody is the response body which test GOAWAY server sent for each request,
+	// for watch request, test GOAWAY server push 1 byte in every second.
+	responseBody = []byte("hello")
 
+	// responseBodySize is the size of response body which test GOAWAY server sent for watch request,
+	// used to check if watch request was broken by GOAWAY frame.
+	responseBodySize = len(responseBody)
+
+	// requestPostBody is the request body which client must send to test GOAWAY server for POST method,
+	// otherwise, test GOAWAY server will respond 400 HTTP status code.
+	requestPostBody = responseBody
+)
+
+// newTestGOAWAYServer return a test GOAWAY server instance.
+func newTestGOAWAYServer() (*httptest.Server, error) {
 	watchHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		timer := time.NewTicker(time.Second)
+		defer timer.Stop()
 
 		w.Header().Set("Transfer-Encoding", "chunked")
 		w.WriteHeader(200)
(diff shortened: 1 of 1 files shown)


answer1991:
As we expected, after upgrading to golang 1.15, GOAWAY feature will not affect concurrency requests. We now are ready to enable GOAWAY feature.

PTAL @lavalamp @deads2k @wojtek-t

staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go
```diff
@@ -176,6 +176,14 @@ var (
 		},
 		[]string{"request_kind"},
 	)
+	// GOAWAYResponse is a number of requests which server response with GOAWAY frame
+	GOAWAYResponse = compbasemetrics.NewCounter(
+		&compbasemetrics.CounterOpts{
+			Name:           "apiserver_goaway_response",
+			Help:           "Number of requests which server response with GOAWAY frame",
```
dims, line 183: response -> responded

dims: Also what time period? if you see the previous 2, they say `in last second`

dims: Looks like we are adding something new `apiserver_goaway_response` so this is not just adding test cases. am i reading this right?

answer1991: Sorry, I used a wrong metrics type. I will fixed it later.

so this is not just adding test cases. am i reading this right? - yes, 2 commits.

staging/src/k8s.io/apiserver/pkg/server/filters/goaway_test.go
```diff
@@ -85,25 +91,34 @@ func TestProbabilisticGoawayDecider(t *testing.T) {
...
+	// responseBody is the response body which test GOAWAY server sent for each request,
+	// for watch request, test GOAWAY server push 1 byte in every second.
+	responseBody = []byte("hello")
 
+	// responseBodySize is the size of response body which test GOAWAY server sent for watch request,
+	// used to check if watch request was broken by GOAWAY frame.
+	responseBodySize = len(responseBody)
+
+	// requestPostBody is the request body which client must send to test GOAWAY server for POST method,
+	// otherwise, test GOAWAY server will response 400 HTTP status code.
```
lavalamp, line 113: (nit: "response" is a noun, "respond" is the verb)

staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go
lavalamp: > Also what time period? if you see the previous 2, they say in last second

The last two metrics are gauges, this is a counter, counters don't have time frames.

However I do think it'd be good to add this metric in a separate PR. I'd like to take the tests first.

The metric is about *connections* (right?), but the other metrics here are about requests. So it's confusing and I think we should talk about that more. Looks like the tests don't depend on this metric, so hopefully it won't be much trouble to send it in a separate PR.

answer1991:
Fixed the nit and removed the metrics commit.

staging/src/k8s.io/apiserver/pkg/server/filters/goaway_test.go
```diff
@@ -305,3 +431,94 @@ func TestHTTP1Requests(t *testing.T) {
 		t.Errorf("expect response HTTP header Connection to be empty, but got: %s", v)
 	}
 }
+
+// TestGOAWAYConcurrency tests GOAWAY frame will not affect concurrency requests in a single http client instance.
+// Known issues in history: https://github.com/kubernetes/kubernetes/issues/91131.
+func TestGOAWAYConcurrency(t *testing.T) {
```
liggitt, line 437: this test fails 100% of the time when run with `go test -race k8s.io/apiserver/pkg/server/filters -run TestGOAWAYConcurrency`, disabling in https://github.com/kubernetes/kubernetes/pull/94529 and opened https://github.com/kubernetes/kubernetes/issues/94532 to track fixing the test or the component

staging/src/k8s.io/apiserver/pkg/server/filters/goaway_test.go
```diff
@@ -195,55 +345,26 @@ func TestClientReceivedGOAWAY(t *testing.T) {
...
-						localAddr = append(localAddr, conn.LocalAddr().String())
-						return
-					},
-				},
+			client, err := newGOAWAYClient(func(network, addr string, cfg *tls.Config) (conn net.Conn, err error) {
+				conn, err = tls.Dial(network, addr, cfg)
+				if err != nil {
+					t.Fatalf("unexpect connection err: %v", err)
+				}
+				localAddr = append(localAddr, conn.LocalAddr().String())
```
liggitt, line 353: this added a data race on unguarded localAddr, fixed in https://github.com/kubernetes/kubernetes/pull/94530

staging/src/k8s.io/apiserver/pkg/server/filters/goaway_test.go
answer1991: caused by golang http/http2 pkg, I have already opened an issue [#41234](https://github.com/golang/go/issues/41234) .

staging/src/k8s.io/apiserver/pkg/server/filters/goaway_test.go
answer1991: As far as I know, it's not the root cause of TestClientReceivedGOAWAY flake, because dial a new connection only when client send a new request, and in this case, requests are sent one by one.

I'm still debug the root cause of TestClientReceivedGOAWAY flake, for now, I doubt that I used an incorrect http2 server/client configuration(`http.Transport.MaxConnsPerHost=1` or SETTINGS_MAX_CONCURRENT_STREAMS limited), or the dial function ignored context.Context which will cause more than one goroutine dial the connection when the first canceled.

Another interesting thing is I got a data race result that write at `if tc.expectConnections != len(localAddr)`(main testing goroutine)  and read at `localAddr = append(localAddr, conn.LocalAddr().String())`, how does client try to set-up a new connection when we finished all requests?

PR #94395, opened 2020-09-01
kubeadm: Ensure etcd manifests are regenerated upon upgrade

### LINKED ISSUE #2273: kubeadm upgrade ignores a configmap change in etcd's image repository if image has the same version
I checked if there was anything similar by searching etcd or image, but didn't find something like this.

## Is this a BUG REPORT or FEATURE REQUEST?

Bug report.

## Versions

**kubeadm version** (use `kubeadm version`): master

**Environment**:
- **Kubernetes version** (use `kubectl version`): 1.18 (not that it matters)
- **Cloud provider or hardware configuration**: openstack (not that it matters)
- **OS** (e.g. from /etc/os-release): SLES (not that it matters)
- **Kernel** (e.g. `uname -a`): 5.3.18-24.9-default  (not that it matters :p )
- **Others**: 

## What happened?

If I have changed my image repository, and I execute an upgrade, and my version of etcd is the same in the new repository, then
the static pod manifest will not be updated according to the configmap. No change will be written to disk.

See also https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/upgrade/staticpods.go#L325 for that logic (permalink: https://github.com/kubernetes/kubernetes/blob/e23d83eead3b5ae57731afb0209f4a2aaa4009dd/cmd/kubeadm/app/phases/upgrade/staticpods.go#L320-L326 ).

## What you expected to happen?

For me, the static pod manifest should be rewritten if there was a change.
We should not only compare the version to prevent a downgrade, but also rewrite the manifest if its content is different than the expected content (we could hash the file's content and compare?)

## How to reproduce it (as minimally and precisely as possible)?

Push an etcd image with the same version in two different repos.
Change the config map to point to the new image repository
`kubeadm upgrade apply` see the manifest not in sync with configmap.

## Anything else we need to know?
This kind of problem might not be limited to etcd static pods.

rosti:
**What type of PR is this?**
/kind bug

A check that verifies that kubeadm does not "upgrade" to an older release was
overly optimized by skipping upgrade if the new version is the same as the old
one. This somewhat makes sense, but that way changes in any of the etcd fields
in the ClusterConfiguration won't be applied if the etcd version is not
changed.

Hence, this simple change ensures that the upgrade is done even when no version
change takes place.

Fixes kubernetes/kubeadm#2273

/cc @kubernetes/sig-cluster-lifecycle-pr-reviews
/area kubeadm
/priority important-longterm
/assign @fabriziopandini @neolit123 @ereslibre

```release-note
kubeadm now makes sure the etcd manifest is regenerated upon upgrade even when no etcd version change takes place
```

```docs

```

### DIFF
--- a/cmd/kubeadm/app/phases/upgrade/staticpods.go
+++ b/cmd/kubeadm/app/phases/upgrade/staticpods.go
@@ -321,8 +321,8 @@ func performEtcdStaticPodUpgrade(certsRenewMgr *renewal.Manager, client clientse
 	if err != nil {
 		return true, errors.Wrapf(err, "failed comparing the current etcd version %q to the desired one %q", currentEtcdVersionStr, desiredEtcdVersion)
 	}
-	if cmpResult < 1 {
-		return false, errors.Errorf("the desired etcd version %q is not newer than the currently installed %q. Skipping etcd upgrade", desiredEtcdVersion, currentEtcdVersionStr)
+	if cmpResult < 0 {
+		return false, errors.Errorf("the desired etcd version %q is older than the currently installed %q. Skipping etcd upgrade", desiredEtcdVersion, currentEtcdVersionStr)
 	}
 
 	beforeEtcdPodHash, err := waiter.WaitForStaticPodSingleHash(cfg.NodeRegistration.Name, constants.Etcd)


evrardjp (approved):
LGTM, but someone else should approve.

neolit123 (commented):
/lgtm
/approve

the change conforms:
https://github.com/kubernetes/kubeadm/issues/2273#issuecomment-684814116
> this guarantees that a new version of kubeadm that uses the old etcd can still apply certain manifest fixes that we introduced.

so SGTM, thanks.

PR #94397, opened 2020-09-01
Stop setting SelfLink in kube-apiserver.

wojtek-t:
Ref https://github.com/kubernetes/enhancements/issues/1164

```release-note
Stop propagating SelfLink (deprecated in 1.16) in kube-apiserver
```

/kind cleanup
/priority important-soon


staging/src/k8s.io/apiserver/pkg/endpoints/apiserver_test.go
```diff
@@ -1144,8 +1144,7 @@ func TestList(t *testing.T) {
 			t.Logf("%d: body: %s", i, string(body))
 			continue
 		}
-		// TODO: future, restore get links
-		if !selfLinker.called {
+		if utilfeature.DefaultFeatureGate.Enabled(features.RemoveSelfLink) == selfLinker.called {
 			t.Errorf("%d: never set self link", i)
```
lavalamp, line 1148: this error message is no longer correct for all the ways you could get into this if statement.

staging/src/k8s.io/apiserver/pkg/endpoints/apiserver_test.go
```diff
@@ -1279,8 +1278,7 @@ func TestListCompression(t *testing.T) {
...
```
lavalamp, line 1282: same comment (not that it matters that much)

staging/src/k8s.io/apiserver/pkg/endpoints/apiserver_test.go
```diff
@@ -1596,7 +1600,7 @@ func TestExport(t *testing.T) {
 		t.Errorf("Expected: exported, saw: %s", itemOut.Other)
 	}
 
-	if !selfLinker.called {
+	if utilfeature.DefaultFeatureGate.Enabled(features.RemoveSelfLink) == selfLinker.called {
 		t.Errorf("Never set self link")
```
lavalamp, line 1604: I'll stop commenting on these :)

staging/src/k8s.io/apiserver/pkg/features/kube_features.go
```diff
@@ -159,7 +159,7 @@ var defaultKubernetesFeatureGates = map[featuregate.Feature]featuregate.FeatureS
 	StorageVersionHash:      {Default: true, PreRelease: featuregate.Beta},
 	WatchBookmark:           {Default: true, PreRelease: featuregate.GA, LockToDefault: true},
 	APIPriorityAndFairness:  {Default: false, PreRelease: featuregate.Alpha},
-	RemoveSelfLink:          {Default: false, PreRelease: featuregate.Alpha},
+	RemoveSelfLink:          {Default: true, PreRelease: featuregate.Beta},
```
lavalamp, line 162: Wait, this was already here? Do we not test feature gates in both positions in tests?

Did it work before? If it didn't work before, then no one could possibly have tested it and it's not clear to me that it's fair or safe to go to beta.

lavalamp (commented):
only minor comments.

staging/src/k8s.io/apiserver/pkg/features/kube_features.go
wojtek-t: Apparently not in unit tests...

The feature gate was there and the self-linking logic itself was gated.

wojtek-t: Actually - it seems that only a few e2e tests are running with all Alpha features enabled:
https://github.com/kubernetes/test-infra/blob/03cbb16ef550658d61fb14912d3dd19b1cd7fa21/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L428

This is why all of these weren't caught...

wojtek-t:
/hold

"TestEmptyList" integration test seem to be a real failure I don't really understand yet. Will investigate in the next few days.

staging/src/k8s.io/apiserver/pkg/endpoints/apiserver_test.go
wojtek-t: done

staging/src/k8s.io/apiserver/pkg/endpoints/apiserver_test.go
wojtek-t: done

staging/src/k8s.io/apiserver/pkg/endpoints/apiserver_test.go
wojtek-t: done

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/rest.go
```diff
@@ -328,6 +328,12 @@ func checkName(obj runtime.Object, name, namespace string, namer ScopeNamer) err
 //   interfaces
 func setObjectSelfLink(ctx context.Context, obj runtime.Object, req *http.Request, namer ScopeNamer) error {
 	if utilfeature.DefaultFeatureGate.Enabled(features.RemoveSelfLink) {
+		// Ensure that for empty lists we don't return <nil> items.
```
wojtek-t, line 331: FYI - this was clearly a bug previously. Fortunately it was caught by an existing integration test.

wojtek-t (commented):
Comments applied, failing test fixed (it was a real issue).

/hold cancel

wojtek-t:
Ref https://github.com/kubernetes/kubernetes/issues/94414

/retest

wojtek-t:
@lavalamp @deads2k - it's ready for review now

wojtek-t:
@lavalamp - are you ok with applying the milestone now?

lavalamp:
Yes, sorry, I keep forgetting

/milestone v1.20

park-cheif5:
Hope y'all having a good day

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/rest.go
sttts: can you comment why the fix is here in `setObjectSelfLink`? Doesn't sound like this func's task is to set an items slice to non-nil.

wojtek-t: For consistency - before this change (i.e. before adding this branch with return) - setting that list to non-nil was happening before in this function too. I mean - it still does - see line 370.

I agree it would be better to move it elsewhere, but that's orthogonal to this PR.

PR #94398, opened 2020-09-01
kubeadm: make the scheduler and KCM connect to the local API endpoint

neolit123:

Pinning the kube-controller-manager and kube-scheduler kubeconfig files
to point to the control-plane-endpoint can be problematic during
immutable upgrades if one of these components ends up contacting an N-1
kube-apiserver:
https://kubernetes.io/docs/setup/release/version-skew-policy/#kube-controller-manager-kube-scheduler-and-cloud-controller-manager
For example, the components can send a request for a non-existing API
version.

Instead of using the CPE for these components, use the LocalAPIEndpoint.
This guarantees that the components would talk to the local
kube-apiserver, which should commonly be the same version, unless the user
explicitly patched manifests.

xref kubernetes/kubeadm#2271

```release-note
kubeadm: make the kubeconfig files for the kube-controller-manager and kube-scheduler use the LocalAPIEndpoint instead of the ControlPlaneEndpoint. This makes kubeadm clusters more reseliant to version skew problems during immutable upgrades: https://kubernetes.io/docs/setup/release/version-skew-policy/#kube-controller-manager-kube-scheduler-and-cloud-controller-manager
```

```docs

```

### DIFF
--- a/cmd/kubeadm/app/phases/kubeconfig/kubeconfig.go
+++ b/cmd/kubeadm/app/phases/kubeconfig/kubeconfig.go
@@ -362,32 +362,37 @@ func ValidateKubeconfigsForExternalCA(outDir string, cfg *kubeadmapi.InitConfigu
 }
 
 func getKubeConfigSpecsBase(cfg *kubeadmapi.InitConfiguration) (map[string]*kubeConfigSpec, error) {
-	apiServer, err := kubeadmutil.GetControlPlaneEndpoint(cfg.ControlPlaneEndpoint, &cfg.LocalAPIEndpoint)
+	controlPlaneEndpoint, err := kubeadmutil.GetControlPlaneEndpoint(cfg.ControlPlaneEndpoint, &cfg.LocalAPIEndpoint)
 	if err != nil {
 		return nil, err
 	}
+	localAPIEndpoint, err := kubeadmutil.GetLocalAPIEndpoint(&cfg.LocalAPIEndpoint)
+	if err != nil {
+		return nil, err
+	}
+
 	return map[string]*kubeConfigSpec{
 		kubeadmconstants.AdminKubeConfigFileName: {
-			APIServer:  apiServer,
+			APIServer:  controlPlaneEndpoint,
 			ClientName: "kubernetes-admin",
 			ClientCertAuth: &clientCertAuth{
 				Organizations: []string{kubeadmconstants.SystemPrivilegedGroup},
 			},
 		},
 		kubeadmconstants.KubeletKubeConfigFileName: {
-			APIServer:  apiServer,
+			APIServer:  controlPlaneEndpoint,
 			ClientName: fmt.Sprintf("%s%s", kubeadmconstants.NodesUserPrefix, cfg.NodeRegistration.Name),
 			ClientCertAuth: &clientCertAuth{
 				Organizations: []string{kubeadmconstants.NodesGroup},
 			},
 		},
 		kubeadmconstants.ControllerManagerKubeConfigFileName: {
-			APIServer:      apiServer,
+			APIServer:      localAPIEndpoint,
 			ClientName:     kubeadmconstants.ControllerManagerUser,
 			ClientCertAuth: &clientCertAuth{},
 		},
 		kubeadmconstants.SchedulerKubeConfigFileName: {
-			APIServer:      apiServer,
+			APIServer:      localAPIEndpoint,
 			ClientName:     kubeadmconstants.SchedulerUser,
 			ClientCertAuth: &clientCertAuth{},
 		},
--- a/cmd/kubeadm/app/util/endpoint.go
+++ b/cmd/kubeadm/app/util/endpoint.go
@@ -34,22 +34,10 @@ import (
 // - if the controlPlaneEndpoint is defined but without a port number, use the controlPlaneEndpoint + localEndpoint.BindPort is used.
 // - Otherwise, in case the controlPlaneEndpoint is not defined, use the localEndpoint.AdvertiseAddress + the localEndpoint.BindPort.
 func GetControlPlaneEndpoint(controlPlaneEndpoint string, localEndpoint *kubeadmapi.APIEndpoint) (string, error) {
-	// parse the bind port
-	bindPortString := strconv.Itoa(int(localEndpoint.BindPort))
-	if _, err := ParsePort(bindPortString); err != nil {
-		return "", errors.Wrapf(err, "invalid value %q given for api.bindPort", localEndpoint.BindPort)
-	}
-
-	// parse the AdvertiseAddress
-	var ip = net.ParseIP(localEndpoint.AdvertiseAddress)
-	if ip == nil {
-		return "", errors.Errorf("invalid value `%s` given for api.advertiseAddress", localEndpoint.AdvertiseAddress)
-	}
-
-	// set the control-plane url using localEndpoint.AdvertiseAddress + the localEndpoint.BindPort
-	controlPlaneURL := &url.URL{
-		Scheme: "https",
-		Host:   net.JoinHostPort(ip.String(), bindPortString),
+	// get the URL of the local endpoint
+	localAPIEndpoint, err := GetLocalAPIEndpoint(localEndpoint)
+	if err != nil {
+		return "", err
 	}
 
 	// if the controlplane endpoint is defined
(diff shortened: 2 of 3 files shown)


cmd/kubeadm/app/phases/kubeconfig/kubeconfig.go
```diff
@@ -362,32 +362,39 @@ func ValidateKubeconfigsForExternalCA(outDir string, cfg *kubeadmapi.InitConfigu
 }
 
 func getKubeConfigSpecsBase(cfg *kubeadmapi.InitConfiguration) (map[string]*kubeConfigSpec, error) {
-	apiServer, err := kubeadmutil.GetControlPlaneEndpoint(cfg.ControlPlaneEndpoint, &cfg.LocalAPIEndpoint)
+	controlPlaneEndpoint, err := kubeadmutil.GetControlPlaneEndpoint(cfg.ControlPlaneEndpoint, &cfg.LocalAPIEndpoint)
```
neolit123, line 365: @fabriziopandini alternatively i can make GetControlPlaneEndpoint() return `controlPlaneEndpoint, localAPIEndpoint, error`

cmd/kubeadm/app/phases/kubeconfig/kubeconfig.go
```diff
@@ -362,32 +362,39 @@ func ValidateKubeconfigsForExternalCA(outDir string, cfg *kubeadmapi.InitConfigu
...
+	if err != nil {
+		return nil, err
+	}
+
+	localAPIEndpointIP, localAPIEndpointPort, err := kubeadmutil.ParseAPIEndpoint(&cfg.LocalAPIEndpoint)
```
fabriziopandini, line 370: What about implementing `kubeadmutil.GetLocalAPIEndpoint` so we are consistent with the same path used above?

cmd/kubeadm/app/phases/kubeconfig/kubeconfig.go
fabriziopandini: I'm ok with the solution you are proposing.
Having `GetControlPlaneEndpoint` returning two endpoints seems a little bit weird to me...

fabriziopandini (commented):
@neolit123 many thanks for jumping on this issue
Overall LGTM, only one minor nit but not blocking
/cc @ereslibre @rosti

cmd/kubeadm/app/phases/kubeconfig/kubeconfig.go
neolit123: SGTM, i will add this extra utility function.

fabriziopandini:
Test failures seem unrelated to the PR...

cmd/kubeadm/app/phases/kubeconfig/kubeconfig.go
```diff
@@ -362,32 +362,39 @@ func ValidateKubeconfigsForExternalCA(outDir string, cfg *kubeadmapi.InitConfigu
...
-			APIServer:  apiServer,
+			APIServer:  controlPlaneEndpoint,
 			ClientName: "kubernetes-admin",
 			ClientCertAuth: &clientCertAuth{
 				Organizations: []string{kubeadmconstants.SystemPrivilegedGroup},
 			},
 		},
 		kubeadmconstants.KubeletKubeConfigFileName: {
-			APIServer:  apiServer,
+			APIServer:  controlPlaneEndpoint,
```
yastij, line 385: for control plane nodes, it might make sense for the kubelet to speak to the local API server (e.g. dynamic environments, machines that are using DHCP where the lease expire)

yastij (commented):
lgtm

cmd/kubeadm/app/phases/kubeconfig/kubeconfig.go
fabriziopandini: From the issue I understood this is going to be tackled in a separated PR, is this right @neolit123

neolit123: that is true and we spoke about this yesterday.

the problem is that kubeadm join has a kubelet-start phase that blocks until it connects to an api-server at the CPE.
https://github.com/kubernetes/kubernetes/blob/f2e3154a140dc192760a5cf8c01f8044e1aa867b/cmd/kubeadm/app/cmd/phases/join/kubelet.go#L99

https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-join/
the CP manifests + etcd for this node are written after the TLS bootstrap.

this implies rework of the kubeadm join phases and needs some discussion (maybe tomorrow @ office hours).

ereslibre (approved):
I'm +1 to this -- thank you @neolit123, it's a safe approach reducing the moving parts specially when upgrading.

/lgtm
/hold
^ for other reviews, feel free to unhold.

cmd/kubeadm/app/phases/kubeconfig/kubeconfig.go
yastij: sounds good, I'll join

neolit123: i will make this PR not close the issue and we can have multiple checkboxes in it to leave time for investigating the kubelet.conf optional change.

neolit123: done.

detiber: +1 to both using the local address for the kubelet and also treating it as a separate PR, this will help to greatly simplify the bootstrapping process when using a load balancer, since the load balancer will not need to be pre-created for bootstrapping.

timothysc:
/approve 

@neolit123 What layers of testing do we need to ensure we find these issues earlier?

cmd/kubeadm/app/phases/kubeconfig/kubeconfig.go
neolit123: updated.

neolit123:
@timothysc 

> @neolit123 What layers of testing do we need to ensure we find these issues earlier?

node join skew testing:
https://github.com/kubernetes/kubeadm/issues/1661

justaugustus:
Also, this needs re-/lgtm.

justaugustus:
FYI ^^ @kubernetes/ci-signal @hasheddan

BenTheElder:
/retest
https://github.com/kubernetes/test-infra/issues/19080

BenTheElder:
https://github.com/kubernetes/test-infra/issues/19080#issuecomment-685191653
/retest

rosti (approved):
Thanks @neolit123 !
/lgtm

ncdc:
Are we ok to remove the hold?

neolit123:
yes, i will send a cherry pick later today.

neolit123:
cherry pick is here: https://github.com/kubernetes/kubernetes/pull/94442

zhangguanzhang:
This breaks the rules, the `controlPlaneEndpoint` maybe a domain, if this is a domain, so it will not run ok after your code

oldthreefeng:
when use method `CreateJoinControlPlaneKubeConfigFiles`  with  controlPlaneEndpoint like ` apiserver.cluster.local` to generate config files.  and use  `kubeadm init --config=/root/kubeadm-config.yaml --upload-certs -v 5`
the error occurs like  
```
I0910 15:15:54.436430   52511 kubeconfig.go:84] creating kubeconfig file for controller-manager.conf
currentConfig.Clusters[currentCluster].Server:  https://apiserver.cluster.local:6443 
config.Clusters[expectedCluster].Server:  https://192.168.160.243:6443
a kubeconfig file "/etc/kubernetes/controller-manager.conf" exists already but has got the wrong API Server URL
```

cmd/kubeadm/app/util/endpoint.go
```diff
@@ -34,22 +34,10 @@ import (
 // - if the controlPlaneEndpoint is defined but without a port number, use the controlPlaneEndpoint + localEndpoint.BindPort is used.
```
jdef, line 34: these docs are now out of date w/ respect to how this method works

neolit123: this can be fixed in the master branch.

neolit123: @jdef actually...why do you think the docs of the function are no longer valid?

all comments seem guaranteed to me by the unit tests that we have:
```
// GetControlPlaneEndpoint returns a properly formatted endpoint for the control plane built according following rules:
// - If the controlPlaneEndpoint is defined, use it.
// - if the controlPlaneEndpoint is defined but without a port number, use the controlPlaneEndpoint + localEndpoint.BindPort is used.
// - Otherwise, in case the controlPlaneEndpoint is not defined, use the localEndpoint.AdvertiseAddress + the localEndpoint.BindPort.
```

https://github.com/neolit123/kubernetes/blob/d9441906c4155173ce1a75421d8fcd1d2f79c471/cmd/kubeadm/app/util/endpoint_test.go#L31-L196

PR #94403, opened 2020-09-01
Always set relevant variables for cross compiling

### LINKED ISSUE #94291: Support customized gcc and cross compilation of kubernetes on platforms other than amd64
There should be a way to override the default gcc compliers mentioned in the [golang.sh](https://github.com/kubernetes/kubernetes/blob/35942f65709d73a8bb8c42904eb6ca3204f0e29a/hack/lib/golang.sh#L397) file with custom gcc compilers for different platforms. Also, cross compilation should be supported on platforms other than amd64.

**Why is this needed**:
Right now, there seems to be no way to override the platform specific cc flags that kubernetes sets. We should be able to override those values.
For example, bottlerocket sets customized CC flags for different platforms using this [patch](https://github.com/bottlerocket-os/bottlerocket/blob/41a72fe21400fa227543fcbed2e59eb0a8b633c1/packages/kubernetes-1.17/0001-always-set-relevant-variables-for-cross-compiling.patch). We should be able to override the default values in kubernetes using environment variables or build time flags and build kubernetes with customized gcc variants that bottlerocket uses. Also, we need an option to set gcc compilers for amd64 platform, so that amd64 targets can be built on different platforms.

bnrjee:
**What type of PR is this?**
/kind feature

1. Right now, there seems to be no way to override the platform specific cc flags that kubernetes sets. We should be able to override those values.
For example, bottlerocket sets customized CC flags for different platforms using this [patch](https://github.com/bottlerocket-os/bottlerocket/blob/41a72fe21400fa227543fcbed2e59eb0a8b633c1/packages/kubernetes-1.17/0001-always-set-relevant-variables-for-cross-compiling.patch). We should be able to override the default values in kubernetes using environment variables or build time flags and build kubernetes with customized gcc variants that bottlerocket uses. 
2. We want to be able to build kubernetes on host platforms other than amd64. For example, we may want to build amd64 targets on arm platforms. For that to happen, we need to remove the condition that checks if the build platform is amd64. If no platform is specified, we can use default values.

Fixes #94291

```
```
**Release note**:
```release-note
Allow cross compilation of kubernetes on different platforms.
```

### DIFF
--- a/hack/lib/golang.sh
+++ b/hack/lib/golang.sh
@@ -392,30 +392,36 @@ kube::golang::set_platform_envs() {
 
   export GOOS=${platform%/*}
   export GOARCH=${platform##*/}
-
-  # Do not set CC when building natively on a platform, only if cross-compiling from linux/amd64
-  if [[ $(kube::golang::host_platform) == "linux/amd64" ]]; then
-    # Dynamic CGO linking for other server architectures than linux/amd64 goes here
-    # If you want to include support for more server platforms than these, add arch-specific gcc names here
-    case "${platform}" in
-      "linux/arm")
-        export CGO_ENABLED=1
-        export CC=arm-linux-gnueabihf-gcc
-        ;;
-      "linux/arm64")
-        export CGO_ENABLED=1
-        export CC=aarch64-linux-gnu-gcc
-        ;;
-      "linux/ppc64le")
-        export CGO_ENABLED=1
-        export CC=powerpc64le-linux-gnu-gcc
-        ;;
-      "linux/s390x")
-        export CGO_ENABLED=1
-        export CC=s390x-linux-gnu-gcc
-        ;;
-    esac
+  # Apply standard values for CGO_ENABLED and CC unless KUBE_BUILD_PLATFORMS is set.
+  if [ -z "${KUBE_BUILD_PLATFORMS:-}" ] ; then
+      export CGO_ENABLED=0
+      export CC=gcc
+      return
   fi
+  # Dynamic CGO linking for other server architectures goes here
+  # If you want to include support for more server platforms than these, add arch-specific gcc names here
+  case "${platform}" in
+    "linux/amd64")
+      export CGO_ENABLED=1
+      export CC=${LINUX_AMD64_CC:-gcc}
+      ;;
+    "linux/arm")
+      export CGO_ENABLED=1
+      export CC=${LINUX_ARM_CC:-arm-linux-gnueabihf-gcc}
+      ;;
+    "linux/arm64")
+      export CGO_ENABLED=1
+      export CC=${LINUX_ARM64_CC:-aarch64-linux-gnu-gcc}
+      ;;
+    "linux/ppc64le")
+      export CGO_ENABLED=1
+      export CC=${LINUX_PPC64LE_CC:-powerpc64le-linux-gnu-gcc}
+      ;;
+    "linux/s390x")
+      export CGO_ENABLED=1
+      export CC=${LINUX_S390X_CC:-s390x-linux-gnu-gcc}
+      ;;
+  esac
 }
 
 kube::golang::unset_platform_envs() {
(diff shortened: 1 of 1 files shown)


bnrjee:
@juanvallejo, @SataQiu , would it be possible to add this to milestone v1.20?

spiffxp:
/assign @BenTheElder 
This LGTM but I'd like a second pair of eyes for the /approve
/milestone v1.20
/lgtm

bnrjee:
Hi @BenTheElder, 
Would it be possible for you to take a look at this?

hack/lib/golang.sh
```diff
@@ -392,30 +392,36 @@ kube::golang::set_platform_envs() {
...
-      "linux/ppc64le")
-        export CGO_ENABLED=1
-        export CC=powerpc64le-linux-gnu-gcc
-        ;;
-      "linux/s390x")
-        export CGO_ENABLED=1
-        export CC=s390x-linux-gnu-gcc
-        ;;
-    esac
+  # Apply standard values for CGO_ENABLED and CC unless KUBE_BUILD_PLATFORMS is set.
```
BenTheElder, line 395: why?

hack/lib/golang.sh
```diff
@@ -392,30 +392,36 @@ kube::golang::set_platform_envs() {
...
+  if [ -z "${KUBE_BUILD_PLATFORMS:-}" ] ; then
+      export CGO_ENABLED=0
+      export CC=gcc
+      return
   fi
+  # Dynamic CGO linking for other server architectures goes here
+  # If you want to include support for more server platforms than these, add arch-specific gcc names here
```
BenTheElder, line 402: this is no longer guarded by being on a supported build host.

BenTheElder (commented):
This change does not seem to gel with the motivation comment in the PR body.

If the motivation is being able to override the CC flags, we should just support _overriding_ them instead of altering how they're set to begin with.
I don't think the switch from checking the host platform to checking if KUBE_BUILD_PLATFORMS is set makes sense.

bnrjee:
> This change does not seem to gel with the motivation comment in the PR body.
> 
> If the motivation is being able to override the CC flags, we should just support _overriding_ them instead of altering how they're set to begin with.
> I don't think the switch from checking the host platform to checking if KUBE_BUILD_PLATFORMS is set makes sense.
Hi @BenTheElder
Sorry about the confusion. I have edited the motivation statement in PR body. We want to override the cc values. We also want to be able to build kubernetes on platforms other than amd64.

hack/lib/golang.sh
bnrjee: We want to build kubernetes on platforms other than amd64. Updated the motivation statement to reflect that.

hack/lib/golang.sh
bnrjee: So, the idea here is to use default values, if we do not want cross compilation.

bnrjee:
Hi  @BenTheElder,
I have updated the motivation statement. Would it be possible for you to take another look at these code changes?

hack/lib/golang.sh
BenTheElder: ok, but that's not possible with this PR alone, so we're no longer correctly checking ...?

hack/lib/golang.sh
BenTheElder: we should just set them as default without the branch then and override them when cross compiling.
this is not an accurate check for cross-compilation.

BenTheElder (commented):
please remove the check for support from the upstream code once it's actually supported to cross build from another platform
the upstream kube-cross / kube build image is only amd64 currently, when that is fixed we can possibly support building from another host.

BenTheElder:
cc @kubernetes/release-engineering @kubernetes/sig-release 
cc @dims <-- recently led some discussions around architecture support

hack/lib/golang.sh
```diff
@@ -392,30 +392,36 @@ kube::golang::set_platform_envs() {
...
+  case "${platform}" in
+    "linux/amd64")
+      export CGO_ENABLED=1
+      export CC=${LINUX_AMD64_CC:-gcc}
```
BenTheElder, line 406: this change is fine. though usually we guard non-standard env with a `KUBE_` prefix

dims:
i'll wait for @BenTheElder to be happy :)

PR #94413, opened 2020-09-02
Enhancement on the testcase to cover more possibilities

chendave:
This change enriches the testcase of `TestPriorityQueue_MoveAllToActiveOrBackoffQueue`
to cover the case that pods in the unschedulableQ could be moved to activeQ after the
backing off.

**What type of PR is this?**

/kind cleanup

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/scheduler/internal/queue/scheduling_queue_test.go
+++ b/pkg/scheduler/internal/queue/scheduling_queue_test.go
@@ -359,17 +359,41 @@ func TestPriorityQueue_Delete(t *testing.T) {
 }
 
 func TestPriorityQueue_MoveAllToActiveOrBackoffQueue(t *testing.T) {
-	q := NewPriorityQueue(newDefaultQueueSort())
+	c := clock.NewFakeClock(time.Now())
+	q := NewPriorityQueue(newDefaultQueueSort(), WithClock(c))
 	q.Add(&medPriorityPod)
 	q.AddUnschedulableIfNotPresent(q.newQueuedPodInfo(&unschedulablePod), q.SchedulingCycle())
 	q.AddUnschedulableIfNotPresent(q.newQueuedPodInfo(&highPriorityPod), q.SchedulingCycle())
+	// Pods is still backing off, move the pod into backoffQ.
 	q.MoveAllToActiveOrBackoffQueue("test")
 	if q.activeQ.Len() != 1 {
-		t.Error("Expected 1 item to be in activeQ")
+		t.Errorf("Expected 1 item to be in activeQ, but got: %v", q.activeQ.Len())
 	}
 	if q.podBackoffQ.Len() != 2 {
-		t.Error("Expected 2 items to be in podBackoffQ")
+		t.Errorf("Expected 2 items to be in podBackoffQ, but got: %v", q.podBackoffQ.Len())
 	}
+
+	// pop out the pods in the backoffQ.
+	q.podBackoffQ.Pop()
+	q.podBackoffQ.Pop()
+
+	q.schedulingCycle++
+	q.AddUnschedulableIfNotPresent(q.newQueuedPodInfo(&unschedulablePod), q.SchedulingCycle())
+	q.AddUnschedulableIfNotPresent(q.newQueuedPodInfo(&highPriorityPod), q.SchedulingCycle())
+	if q.unschedulableQ.get(&unschedulablePod) == nil || q.unschedulableQ.get(&highPriorityPod) == nil {
+		t.Errorf("Expected %v and %v in the unschedulableQ", unschedulablePod.Name, highPriorityPod.Name)
+	}
+	// Move clock by podInitialBackoffDuration, so that pods in the unschedulableQ would pass the backing off,
+	// and the pods will be moved into activeQ.
+	c.Step(q.podInitialBackoffDuration)
+	q.MoveAllToActiveOrBackoffQueue("test")
+	if q.activeQ.Len() != 3 {
+		t.Errorf("Expected 3 items to be in activeQ, but got: %v", q.activeQ.Len())
+	}
+	if q.podBackoffQ.Len() != 0 {
+		t.Errorf("Expected 0 item to be in podBackoffQ, but got: %v", q.podBackoffQ.Len())
+	}
+
 }
 
 // TestPriorityQueue_AssignedPodAdded tests AssignedPodAdded. It checks that


pkg/scheduler/internal/queue/scheduling_queue_test.go
```diff
@@ -359,17 +359,38 @@ func TestPriorityQueue_Delete(t *testing.T) {
 }
 
 func TestPriorityQueue_MoveAllToActiveOrBackoffQueue(t *testing.T) {
-	q := NewPriorityQueue(newDefaultQueueSort())
+	c := clock.NewFakeClock(time.Now())
+	q := NewPriorityQueue(newDefaultQueueSort(), WithClock(c))
 	q.Add(&medPriorityPod)
+	q.MoveAllToActiveOrBackoffQueue("test")
```
alculquicondor, line 365: why is this line necessary?

alculquicondor (commented):
Please run the test several times to verify there are no other race conditions:

```
go test -run <testname> -count 100
```

pkg/scheduler/internal/queue/scheduling_queue_test.go
soulxu: I guess that is for updating the `PriorityQueue.moveRequestCycle`, maybe we can assert  the `PriorityQueue.moveRequestCycle` and `PriorityQueue.schedulingCycle` after this line, that help the people understand the purpose for this line.

pkg/scheduler/internal/queue/scheduling_queue_test.go
```diff
@@ -359,17 +359,38 @@ func TestPriorityQueue_Delete(t *testing.T) {
...
+	// Make sure the move request is happened first.
+	c.Step(time.Second)
```
soulxu, line 367: The next two lines for `AddUnschedulableIfNotPresent`, they depend on the `PriorityQueue.moveRequestCycle` and `PriorityQueue.schedulingCycle` to decide going to the backoff queue or unschedule queue. So this should be useless, it is only useful for `MoveAllToActiveOrBackoffQueue`, it needs the clock to determine the pod is backoff or not.

pkg/scheduler/internal/queue/scheduling_queue_test.go
```diff
@@ -359,17 +359,38 @@ func TestPriorityQueue_Delete(t *testing.T) {
...
+		t.Errorf("Expected 2 items to be in podBackoffQ, but got: %v", q.podBackoffQ.Len())
 	}
+
+	// pop out the pods in the backoffQ.
+	q.podBackoffQ.Pop()
+	q.podBackoffQ.Pop()
+
+	q.schedulingCycle++
+	q.AddUnschedulableIfNotPresent(q.newQueuedPodInfo(&unschedulablePod), q.SchedulingCycle())
+	q.AddUnschedulableIfNotPresent(q.newQueuedPodInfo(&highPriorityPod), q.SchedulingCycle())
```
soulxu, line 383: so they are going to unschedulable queue here? this is different with line 368 an 369, so it is better to have assert about it.

pkg/scheduler/internal/queue/scheduling_queue_test.go
chendave: I pulled this line ahead compared with original testcase, combined with the line below `c.Step(time.Second)` just to make sure the move request is happened first, and `p.moveRequestCycle` got updated.

pkg/scheduler/internal/queue/scheduling_queue_test.go
chendave: > So this should be useless, it is only useful for MoveAllToActiveOrBackoffQueue, it needs the clock to determine the pod is backoff or not.

sorry, I am quite follow you here.  In this testcase, all that `MoveAllToActiveOrBackoffQueue` did is set `p.moveRequestCycle`,  I cannot see `MoveAllToActiveOrBackoffQueue` needs the clock anywhere.

pkg/scheduler/internal/queue/scheduling_queue_test.go
chendave: yes.

pkg/scheduler/internal/queue/scheduling_queue_test.go
soulxu: > > So this should be useless, it is only useful for MoveAllToActiveOrBackoffQueue, it needs the clock to determine the pod is backoff or not.
> 
> sorry, I am quite follow you here. In this testcase, all that `MoveAllToActiveOrBackoffQueue` did is set `p.moveRequestCycle`, I cannot see `MoveAllToActiveOrBackoffQueue` needs the clock anywhere.

Sorry, I didn't describe clearly.  :)

I mean `AddUnschedulableIfNotPresent` needn't the clock, so you needn't step the clock here

We only need to step the clock when you test the `MoveAllToActiveOrBackoffQueue `, like line 385. It needs the clock to determine pod going to which queue:

`https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/internal/queue/scheduling_queue.go#L514`

`https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/internal/queue/scheduling_queue.go#L281`

pkg/scheduler/internal/queue/scheduling_queue_test.go
chendave: Done

pkg/scheduler/internal/queue/scheduling_queue_test.go
chendave: > why is this line necessary?

not necessary actually, have reverted it back.

pkg/scheduler/internal/queue/scheduling_queue_test.go
chendave: Done

chendave:
/remove-kind flake

remove the tag since there is low chance to reproduce the flake unless we deliberately add some time consuming operation before the `MoveAllToActiveOrBackoffQueue`, this not true for now.

Big thanks to @alculquicondor  @soulxu  for review!

pkg/scheduler/internal/queue/scheduling_queue_test.go
```diff
@@ -359,17 +359,41 @@ func TestPriorityQueue_Delete(t *testing.T) {
...
+		t.Errorf("Expected %v and %v in the unschedulableQ", unschedulablePod.Name, highPriorityPod.Name)
+	}
+	// Move clock by podInitialBackoffDuration, so that pods in the unschedulableQ would pass the backing off,
+	// and the pods will be moved into activeQ.
+	c.Step(q.podInitialBackoffDuration)
+	q.MoveAllToActiveOrBackoffQueue("test")
+	if q.activeQ.Len() != 3 {
+		t.Errorf("Expected 3 items to be in activeQ, but got: %v", q.activeQ.Len())
+	}
+	if q.podBackoffQ.Len() != 0 {
```
soulxu, line 393: This assertion seems meanless if without assert the unschedulable queue, since line 378 the pod backoff queue becomes empty, and `AddUnschedulableIfNotPresent` adds the pod to unschedulable queue.  we should assert pods move out unschedulable queue.

alculquicondor:
/lgtm

Did you run the test 100 times?

chendave:
> /lgtm
> 
> Did you run the test 100 times?

yes, run multiple times :)

alculquicondor:
Thanks!

/approve

pkg/scheduler/internal/queue/scheduling_queue_test.go
chendave: Line @390 should already done that, right? this check is used to compare with Line @372 to tell us the pods are not moved to backoffQ, I think it's not useless.

soulxu: Emm...you are right, sorry about that.

PR #94433, opened 2020-09-02
remove iperf e2e test ip family tags

aojea:
The framework has support to detect the IP family of the cluster, so we don't need to use the tags

The test has been running successfully for a while in KIND https://testgrid.k8s.io/sig-network-kind

```release-note
NONE
```


aojea:
/assign @BenTheElder 
thanks @jayunit100 for this test, it was a gem I've found very useful in the framework

MrHohn (approved):
/lgtm
/approve

PR #94444, opened 2020-09-02
Add support to size memory backed volumes

derekwaynecarr:
**What type of PR is this?**
/kind feature

Size memory backed emptyDir volumes as the min of memory available to pod and local emptyDir sizeLimit.  This is important for applications that want more reliable emptyDir sizing than 50% of node memory (which is Linux default) when or if their applications run on variety of node sources.

```release-note
Add feature to size memory backed volumes
```

```docs
- [KEP]: https://github.com/kubernetes/enhancements/pull/1968
```

### DIFF
--- a/pkg/volume/emptydir/empty_dir.go
+++ b/pkg/volume/emptydir/empty_dir.go
@@ -29,7 +29,10 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
+	"k8s.io/kubernetes/pkg/features"
+	"k8s.io/kubernetes/pkg/kubelet/cm"
 	"k8s.io/kubernetes/pkg/volume"
 	volumeutil "k8s.io/kubernetes/pkg/volume/util"
 	"k8s.io/kubernetes/pkg/volume/util/fsquota"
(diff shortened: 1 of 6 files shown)


pkg/volume/emptydir/empty_dir.go
```diff
@@ -107,17 +110,58 @@ func (plugin *emptyDirPlugin) NewMounter(spec *volume.Spec, pod *v1.Pod, opts vo
...
+	return sizeLimit, nil
+}
+
 func (plugin *emptyDirPlugin) newMounterInternal(spec *volume.Spec, pod *v1.Pod, mounter mount.Interface, mountDetector mountDetector, opts volume.VolumeOptions) (volume.Mounter, error) {
 	medium := v1.StorageMediumDefault
-
+	sizeLimit := &resource.Quantity{}
 	if spec.Volume.EmptyDir != nil { // Support a non-specified source as EmptyDir.
 		medium = spec.Volume.EmptyDir.Medium
+		size, err := plugin.calculateEmptyDirMemorySize(spec, pod)
```
dashpole, line 154: We should only call this if the medium is memory, right?

dashpole (commented):
also needs unit testing.

pkg/volume/emptydir/empty_dir.go
derekwaynecarr: Ha, yeah.  Good catch :-)

derekwaynecarr:
will add e2e and unit testing next week to this pr.

derekwaynecarr:
/hold cancel

@dashpole added e2e and unit test.

dashpole:
/lgtm
/priority important-longterm

/assign @dchen1107 
for approval

pkg/volume/emptydir/empty_dir.go
```diff
@@ -107,17 +110,58 @@ func (plugin *emptyDirPlugin) NewMounter(spec *volume.Spec, pod *v1.Pod, opts vo
...
+		if podMemoryLimit.Cmp(zero) > 0 && podMemoryLimit.Cmp(*sizeLimit) < 1 {
+			sizeLimit = podMemoryLimit
+		}
+	}
+
+	// volume local size is  used if and only if less than what pod could consume
+	if spec.Volume.EmptyDir.SizeLimit != nil {
+		volumeSizeLimit := spec.Volume.EmptyDir.SizeLimit
+		// ensure 0 < value < size
+		if volumeSizeLimit.Cmp(zero) > 0 && volumeSizeLimit.Cmp(*sizeLimit) < 1 {
```
2rs2ts, line 140: I know this is just alpha so it's fine if you plan on tackling this later, but I think that there should be some sort of indicator to the user if they try to use a size limit in their config that's bigger than what the size limit actually is. At the very least something should be logged, though I think either the pod should be not allowed to be admitted or an event should be created.

2rs2ts (commented):
Thank you for this. Very exciting to see.

pkg/volume/emptydir/empty_dir.go
```diff
@@ -271,8 +316,13 @@ func (ed *emptyDir) setupTmpfs(dir string) error {
 		return nil
 	}
 
-	klog.V(3).Infof("pod %v: mounting tmpfs for volume %v", ed.pod.UID, ed.volName)
-	return ed.mounter.Mount("tmpfs", dir, "tmpfs", nil /* options */)
+	var options []string
+	// Linux system default is 50% of capacity.
```
jingxu97, line 320: what do you mean in this comment?

dims:
```
size:      The limit of allocated bytes for this tmpfs instance. The 
           default is half of your physical RAM without swap. If you
           oversize your tmpfs instances the machine will deadlock
           since the OOM handler will not be able to free that memory.
```
from: https://www.kernel.org/doc/Documentation/filesystems/tmpfs.txt

derekwaynecarr:
pushed rebase, hoping to still land this in 1.20 since the enhancement paperwork was done.

pkg/volume/emptydir/empty_dir.go
derekwaynecarr: @2rs2ts maybe we can throw an event.  the pod author is unlikely to have access to node logs in many contexts.  we can explore that after this lands.

dashpole:
did another pass
/lgtm

pkg/volume/emptydir/empty_dir.go
```diff
@@ -271,8 +316,14 @@ func (ed *emptyDir) setupTmpfs(dir string) error {
 		return nil
 	}
 
+	var options []string
+	// Linux system default is 50% of capacity.
+	if ed.sizeLimit != nil && ed.sizeLimit.Value() > 0 {
+		options = []string{fmt.Sprintf("size=%d", ed.sizeLimit.Value())}
+	}
```
gnufied, line 323: is it a problem if `EmptyDir.SizeLimit` is greater than calculated size limit but pod does not get evicted until `EmptyDir.SizeLimit` in API? should eviction thresholds be updated as well?

derekwaynecarr: the limit is distinct from usage.

the limit will not be able to exceed allocatable memory for a pod.

derekwaynecarr: so to answer, i do not think there is a problem here.

gnufied:
The changes look good from storage perspective.

/lgtm

pkg/volume/emptydir/empty_dir.go
2rs2ts: >we can explore that after this lands

Fine by me.

>the pod author is unlikely to have access to node logs in many contexts

Yes, that'll be the case for most of my coworkers 😅 as a cluster admin I would want logs at a minimum because I could write a scheduled search in our log storage to detect if it happened to anybody, and reach out to them. From a user perspective, events are visible, but they don't get shoved in their face, so I think most of my colleagues would miss the event. So blocking admission would feel like the best way to make sure my users know this is happening, but as long as there's some sort of indicator, rather than this silent change, I'd be able to adopt the feature

derekwaynecarr:
rebased to pick up change in kube-features, will need to retag.

gnufied:
putting hold for a bit, so @xing-yang  can take a look.
/hold

webmutation:
What version of Kubernetes is this PR merged to? 

Setting the size as follows on kubernetes 1.21 but the shm still reports 50% of available ram. instead of the value declared in the yaml file.

Also, we are getting evicted pods when the limit is exceeded instead of getting a firendly message like we have with the 64Mb default limit... 

dd: writing to '/dev/shm/test': No space left on device if we reach the 64Mb limit, the pod does not get evicted
if we try to do the same for the pod we get a pod evicted event with and whitout the limit. So it feels like we are missing something to have a more reliable behavior.

```
    spec:
      containers:
      - args:
        - infinity
        command:
        - sleep
        image: centos:7
        name: centos
        resources:
          limits:
            memory: 256Mi
        volumeMounts:
        - mountPath: /dev/shm
          name: cache-volume
      volumes:
      - emptyDir:
          medium: Memory
          sizeLimit: 128Mi
        name: cache-volume
```

JKrehling:
https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

@webmutation 

Looks like the feature gate is false in 1.21.  

SizeMemoryBackedVolumes


<br class="Apple-interchange-newline">

webmutation:
Indeed, thank you. I will try again in 1.23 cluster.

PR #94448, opened 2020-09-02
Build non-static Kubernetes binaries as position-independent executables (PIE)

### LINKED ISSUE #90311: Build position independent executables
## What would you like to be added:

Currently `kubelet` is dynamically linked but is not built as a position independent executable. It would be great if it could be built with Go’s `-buildmode=pie` flag to enable ASLR.

## Why is this needed:

I’m not a security expert, and I’m not familiar with those concepts, but it seems that ASLR is a pretty common (and recommended) security mechanism.

If I understand correctly, statically linked binaries (e.g. all the Kubernetes binaries except `kubelet`) do not require/support PIE so this would only apply to `kubelet`.

## Related issues/PR:

There was an attempt at using Go’s `-buildmode=pie` in #57225 but it seems that it was abandoned due to an incompatibility issue with Windows.

abhay-krishna:
**What type of PR is this?**
/kind feature

Currently, `kubelet` which is dynamically-linked is not built as position-independent executable. This PR applies Go's buildmode=pie flag to the non-static binaries. Building them as PIE enables the [ASLR](https://en.wikipedia.org/wiki/Address_space_layout_randomization) security mechanism, which facilitates defense in depth when building the binaries. This fix does not apply to statically-linked binaries. 

Bottlerocket applies this [patch](https://github.com/bottlerocket-os/bottlerocket/blob/41a72fe21400fa227543fcbed2e59eb0a8b633c1/packages/kubernetes-1.17/0003-enable-PIE-for-platform-binaries.patch) for Kubernetes 1.17 to build dynamically linked binaries as PIE, but bearing security in mind, it is recommended to set PIE as default in upstream Kubernetes. This fix also allows removal of similar patches from downstream Kubernetes.

Fixes [#90311](https://github.com/kubernetes/kubernetes/issues/90311)

```release-note
NONE
```

**Release Notes**
```
Apply Go's buildmode=pie flag to enable ASLR for non-static binary builds
```

```docs

```

### DIFF
--- a/cmd/kubelet/BUILD
+++ b/cmd/kubelet/BUILD
@@ -10,6 +10,7 @@ load("//staging/src/k8s.io/component-base/version:def.bzl", "version_x_defs")
 go_binary(
     name = "kubelet",
     embed = [":go_default_library"],
+    linkmode = "pie",
     x_defs = version_x_defs(),
 )
 


abhay-krishna:
I signed it

BenTheElder:
this should also be enabled in bazel to match (which is how we build in presubmit)
see: https://github.com/bazelbuild/rules_go/

abhay-krishna:
Ben Cressey has signed it

abhay-krishna:
I signed it

abhay-krishna:
> this should also be enabled in bazel to match (which is how we build in presubmit)
> see: https://github.com/bazelbuild/rules_go/

Correct me if I'm wrong, but I believe bazel already supports building with pie using the -linkmode=pie attribute. This works in the same way as the go -buildmode flag and accepts "pie" as a valid argument value.

abhay-krishna:
@BenTheElder Could you provide some input on this? Thanks!

BenTheElder:
> Correct me if I'm wrong, but I believe bazel already supports building with pie using the -linkmode=pie attribute. This works in the same way as the go -buildmode flag and accepts "pie" as a valid argument value.

having support is great! can we please enable it in this PR so as to avoid skew between the build outputs?

BenTheElder:
To clarify again: please configure the bazel build to match with PIE. If rules_go already supports it, that's excellent, but please actually change the build settings to enable it to match.

If you need help figuring this out, #bazel is often useful. I'm a bit out of bandwidth right now, but you probably just need to apply some setting in https://github.com/kubernetes/kubernetes/blob/fd5d61060a20c1fdf56fdb292228dc06b466e7f3/cmd/kubelet/BUILD#L10

> Currently, kubelet and a few other Kubernetes are dynamically-linked but are not built as position-independent executables.

I don't remember any dynamically linked binaries other than kubelet, what are you referring to here?

/sig security

mikedanese:
Why doesn't go build PIE executables by default? Presumably, this would be on by default of there were no trade offs.

neolit123:
for Windows `go build` 1.15x defaulted to pie some time ago, so this change is not needed there:
https://github.com/golang/go/blob/release-branch.go1.15/src/cmd/go/internal/work/init.go#L158-L159

but i cannot find what architectures PIE mode is well supported on Linux.

a search about `-buildmode=pie` in the go repo reveals tickets such as:
https://github.com/golang/go/issues/28531
https://github.com/golang/go/search?q=%22-buildmode%3Dpie%22&type=issues

on the other hand, apparently Fedora did the switch around 26:
https://fedoraproject.org/wiki/Changes/golang-buildmode-pie

@abhay-krishna have you done research about this and whether our list architectures will be supported:
e.g. https://github.com/kubernetes/kubernetes/blob/106ee38796be33073425a5ac731ea0adbde3d151/build/pause/Makefile#L36

abhay-krishna:
@neolit123 Thank you for your comments regarding the architectures, and apologies for the delay in response.

I had referred to [this](https://github.com/golang/go/blob/b53df56001a6bd375b1dc2026908f331729cd731/src/cmd/internal/sys/supported.go#L87-L98) to find out whether buildmode=pie is supported in all the Linux architectures we build for.

neolit123:
@abhay-krishna 

> I had referred to this to find out whether buildmode=pie is supported in all the Linux architectures we build for.

thanks for the link.
this change now only includes the bazel change, is PIE default on all Linux architectures when using gobuild?
unless i misread https://github.com/golang/go/blob/release-branch.go1.15/src/cmd/go/internal/work/init.go it wasn't.

hasheddan (commented):
@abhay-krishna now that [bazel has been removed](https://github.com/kubernetes/kubernetes/pull/99561) this PR needs to be updated. Do you have bandwidth to do this?

Issue #72956, opened 2019-01-16
Eliminate shellcheck failures


BenTheElder:
**What happened**: After #24614 / #68438  we had 283 known failing `*.sh` files being excluded from shellcheck lints.

**What you expected to happen**: We should be linting all files. When files are listed in `hack/.shellcheck_failures` we may miss actual bugs in them. Initially to get the linter in place we had to list many files there to keep the PR reasonably sized, but ideally we can fix these over time and eliminate this file.

**How to reproduce it (as minimally and precisely as possible)**: `hack/verify-shellcheck.sh`

**Anything else we need to know?**: I intend to continue working on this, EG #72861, #72955, ...

/sig testing
/assign

### Known failing files TODO:

 - [x] ./build/build-image/rsyncd.sh (Done in: https://github.com/kubernetes/kubernetes/pull/72955)
 - [x] ./build/common.sh (Done in: https://github.com/kubernetes/kubernetes/pull/73117)
 - [x] ./build/copy-output.sh (Done in: https://github.com/kubernetes/kubernetes/pull/72955)
 - [x] ./build/lib/release.sh
 - [x] ./build/make-build-image.sh (Done in: https://github.com/kubernetes/kubernetes/pull/72955)
 - [x] ./build/make-clean.sh (Done in: https://github.com/kubernetes/kubernetes/pull/72955)
 - [x] ./build/package-tarballs.sh (Done in: https://github.com/kubernetes/kubernetes/pull/72955)
 - [x] ./build/release-images.sh (Done in: https://github.com/kubernetes/kubernetes/pull/72955)
 - [x] ./build/release-in-a-container.sh (Done in: https://github.com/kubernetes/kubernetes/pull/72955)
 - [x] ./build/release.sh (Done in: https://github.com/kubernetes/kubernetes/pull/72955)
 - [x] ./build/run.sh (Done in: https://github.com/kubernetes/kubernetes/pull/72955)
 - [x] ./build/shell.sh (Done in: https://github.com/kubernetes/kubernetes/pull/72955)
 - [x] ./build/util.sh (Done in: https://github.com/kubernetes/kubernetes/pull/72955)
 - [x] ./cluster/addons/addon-manager/kube-addons.sh
 - [x] ./cluster/addons/fluentd-elasticsearch/es-image/run.sh
 - [x] ./cluster/addons/fluentd-elasticsearch/fluentd-es-image/run.sh
[...]

PR #94456, opened 2020-09-03

joakimr-axis:
**What type of PR is this?**
/kind cleanup

ref: #72956 (part of #90405)
This patch adds shellcheck disables where we deliberately want to have the current code that shellcheck triggers on.

```release-note
NONE
```

### DIFF
--- a/cluster/gce/util.sh
+++ b/cluster/gce/util.sh
@@ -119,6 +119,8 @@ fi
 NODE_INSTANCE_PREFIX=${NODE_INSTANCE_PREFIX:-"${INSTANCE_PREFIX}-minion"}
 WINDOWS_NODE_INSTANCE_PREFIX=${WINDOWS_NODE_INSTANCE_PREFIX:-"${INSTANCE_PREFIX}-windows-node"}
 
+# NODE_TAG (expected to be) defined by caller
+# shellcheck disable=SC2153
 NODE_TAGS="${NODE_TAG}"
 
 ALLOCATE_NODE_CIDRS=true
@@ -149,6 +151,8 @@ function verify-prereqs() {
   if [ "$(openssl version | cut -d\  -f1)" == "LibreSSL" ]; then
     echo "LibreSSL is not supported. Please ensure openssl points to an OpenSSL binary"
     if [ "$(uname -s)" == "Darwin" ]; then
+      # We want this print just the way it is
+      # shellcheck disable=SC2016
       echo 'On macOS we recommend using homebrew and adding "$(brew --prefix openssl)/bin" to your PATH'
     fi
     exit 1
@@ -1216,7 +1220,7 @@ CONTAINER_RUNTIME_TEST_HANDLER: $(yaml-quote "${CONTAINER_RUNTIME_TEST_HANDLER:-
 UBUNTU_INSTALL_CONTAINERD_VERSION: $(yaml-quote "${UBUNTU_INSTALL_CONTAINERD_VERSION:-}")
 UBUNTU_INSTALL_RUNC_VERSION: $(yaml-quote "${UBUNTU_INSTALL_RUNC_VERSION:-}")
 NODE_LOCAL_SSDS_EXT: $(yaml-quote "${NODE_LOCAL_SSDS_EXT:-}")
-NODE_LOCAL_SSDS_EPHEMERAL: "$(yaml-quote ${NODE_LOCAL_SSDS_EPHEMERAL:-})"
+NODE_LOCAL_SSDS_EPHEMERAL: $(yaml-quote "${NODE_LOCAL_SSDS_EPHEMERAL:-}")
 LOAD_IMAGE_COMMAND: $(yaml-quote "${LOAD_IMAGE_COMMAND:-}")
 ZONE: $(yaml-quote "${ZONE}")
 REGION: $(yaml-quote "${REGION}")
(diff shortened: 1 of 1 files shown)


cluster/gce/util.sh
```diff
@@ -119,7 +119,9 @@ fi
 NODE_INSTANCE_PREFIX=${NODE_INSTANCE_PREFIX:-"${INSTANCE_PREFIX}-minion"}
 WINDOWS_NODE_INSTANCE_PREFIX=${WINDOWS_NODE_INSTANCE_PREFIX:-"${INSTANCE_PREFIX}-windows-node"}
 
-NODE_TAGS="${NODE_TAG}"
+# NODE_TAG (expected to be) defined by caller
+# shellcheck disable=SC2153
+NODE_TAGS=$NODE_TAG
```
spiffxp, line 124: agree with the disable, but why drop the quotes/braces?

cluster/gce/util.sh
```diff
@@ -1680,7 +1684,9 @@ function generate-certs {
     cd "${CERT_DIR}"
     ./easyrsa init-pki
     # this puts the cert into pki/ca.crt and the key into pki/private/ca.key
+    # shellcheck disable=SC2153
```
spiffxp, line 1687: I know you allude to shellcheck directives in the comment above, but it would be more helpful for future maintainers to call out the specific variable(s) the disable directive is being used each time you use it

cluster/gce/util.sh
```diff
@@ -1680,7 +1684,9 @@ function generate-certs {
...
     ./easyrsa --batch "--req-cn=${PRIMARY_CN}@$(date +%s)" build-ca nopass
+    # shellcheck disable=SC2153
```
spiffxp, line 1689: same here

cluster/gce/util.sh
```diff
@@ -1837,6 +1843,7 @@ function update-or-verify-gcloud() {
     sudo_prefix="sudo"
   fi
   # update and install components as needed
+  # shellcheck disable=SC2086
```
spiffxp, line 1846: why?

spiffxp (changes requested):
Mostly looks good, a few nits

cluster/gce/util.sh
joakimr-axis: Quotes are superfluous at variable assignments.

```
echo 'foo bar' > foobar.txt
a=$(cat foobar.txt)
b="$(cat foobar.txt)"
```

will result in `a` and `b` having the same content, `foo bar`.

cluster/gce/util.sh
joakimr-axis: Agree, I will add such clarifying comments.

cluster/gce/util.sh
phenixblue: The quotes and braces are stylistic to the standards used by the project. We have style guidelines to prevent personal preferences from littering the codebase with inconsistencies given the number of contributors involved.

Please reference the style guidelines for the project [here](https://github.com/kubernetes/community/blob/master/contributors/guide/coding-conventions.md#code-conventions)

And specifically these sections:

- https://google.github.io/styleguide/shellguide.html#quoting
- https://google.github.io/styleguide/shellguide.html#variable-expansion

```
Stay consistent with what you find; quote your variables; prefer "${var}" over "$var"
```

joakimr-axis: Thank you for your attention, PR has been updated accordingly.

cluster/gce/util.sh
```diff
@@ -1216,7 +1220,7 @@ CONTAINER_RUNTIME_TEST_HANDLER: $(yaml-quote "${CONTAINER_RUNTIME_TEST_HANDLER:-
 UBUNTU_INSTALL_CONTAINERD_VERSION: $(yaml-quote "${UBUNTU_INSTALL_CONTAINERD_VERSION:-}")
 UBUNTU_INSTALL_RUNC_VERSION: $(yaml-quote "${UBUNTU_INSTALL_RUNC_VERSION:-}")
 NODE_LOCAL_SSDS_EXT: $(yaml-quote "${NODE_LOCAL_SSDS_EXT:-}")
-NODE_LOCAL_SSDS_EPHEMERAL: "$(yaml-quote ${NODE_LOCAL_SSDS_EPHEMERAL:-})"
+NODE_LOCAL_SSDS_EPHEMERAL: $(yaml-quote "${NODE_LOCAL_SSDS_EPHEMERAL:-}")
```
phenixblue, line 1223: Is this change related to a spellcheck failure?

phenixblue (commented):
Looks good to me pending clarification on inline comment

cluster/gce/util.sh
joakimr-axis: No, it is not. Well spotted. It is a fix for quotes that were bogusly applied to this line in the quote-fix for the `yaml-quote` statements.

cluster/gce/util.sh
joakimr-axis: If we were to quote `gcloud_prompt`, say it contained _foo bar_, we would have the single argument `foo bar` instead of the two arguments `foo` `bar` passed to gcloud. And if gcloud_prompt was empty, we would get an empty argument instead of no argument.

spiffxp (approved):
/approve
/lgtm

PR #94457, opened 2020-09-03

joakimr-axis:
**What type of PR is this?**
/kind cleanup

ref: #72956 (part of #90405)
Remove unnecessary use of cat.

```release-note
NONE
```

### DIFF
--- a/cluster/gce/util.sh
+++ b/cluster/gce/util.sh
@@ -1611,24 +1611,24 @@ function create-certs {
 
   # By default, linux wraps base64 output every 76 cols, so we use 'tr -d' to remove whitespaces.
   # Note 'base64 -w0' doesn't work on Mac OS X, which has different flags.
-  CA_KEY_BASE64=$(cat "${CERT_DIR}/pki/private/ca.key" | base64 | tr -d '\r\n')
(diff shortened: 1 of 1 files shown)


oomichi:
Why don't you remove cluster/gce/util.sh line from hack/.shellcheck_failures?

/cc @oomichi

joakimr-axis:
> Why don't you remove cluster/gce/util.sh line from hack/.shellcheck_failures?

This PR only addresses some of the shellcheck errors/warnings in this file. As stated in the initial comment, it is part of #90405, which takes care of all of them but has been regarded as too large and thus split into this and several other sub-PRs.
So a removal of that line would cause shellcheck to still fail this cluster/gce/util.sh (although with fewer issues after this PR than without it).

oomichi:
> This PR only addresses some of the shellcheck errors/warnings in this file. As stated in the initial comment, it is part of #90405, which takes care of all of them but has been regarded as too large and thus split into this and several other sub-PRs.

If so, it is better to write the scope of this PR in the PR message clearly like `Replace the combination of cat and base64 with just base64` or something.

joakimr-axis:
> If so, it is better to write the scope of this PR in the PR message clearly like `Replace the combination of cat and base64 with just base64` or something.

Thank you for your input. I have now updated the PR message accordingly.

oomichi:
Thanks for updating.

/lgtm

PR #94459, opened 2020-09-03

joakimr-axis:
**What type of PR is this?**
/kind cleanup

ref: #72956 (part of #90405)
Split declaration and assignment for local variables where shellcheck triggers risk of shadowing warnings.
Some missing exports also added, and an unused variable was removed.

```release-note
NONE
```

### DIFF
--- a/cluster/gce/util.sh
+++ b/cluster/gce/util.sh
@@ -754,9 +754,9 @@ function construct-common-kubelet-flags {
 # Sets KUBELET_ARGS with the kubelet flags for Linux nodes.
 # $1: if 'true', we're rendering flags for a master, else a node
 function construct-linux-kubelet-flags {
-  local node_type=$1
+  local node_type="$1"
   local flags
-  flags=$(construct-common-kubelet-flags)
+  flags="$(construct-common-kubelet-flags)"
   # Keep in sync with CONTAINERIZED_MOUNTER_HOME in configure-helper.sh
   flags+=" --experimental-mounter-path=/home/kubernetes/containerized_mounter/mounter"
   flags+=" --experimental-check-node-capabilities-before-mount=true"
@@ -816,7 +816,7 @@ function construct-linux-kubelet-flags {
   fi
   flags+=" --volume-plugin-dir=${VOLUME_PLUGIN_DIR}"
   local node_labels
-  node_labels=$(build-linux-node-labels "${node_type}")
+  node_labels="$(build-linux-node-labels "${node_type}")"
   if [[ -n "${node_labels:-}" ]]; then
     flags+=" --node-labels=${node_labels}"
   fi
@@ -845,13 +845,13 @@ function construct-linux-kubelet-flags {
 # string delimiters.
 function construct-windows-kubelet-flags {
   local flags
-  flags=$(construct-common-kubelet-flags)
+  flags="$(construct-common-kubelet-flags)"
 
   # Note: NODE_KUBELET_TEST_ARGS is empty in typical kube-up runs.
   flags+=" ${NODE_KUBELET_TEST_ARGS:-}"
 
   local node_labels
-  node_labels=$(build-windows-node-labels)
+  node_labels="$(build-windows-node-labels)"
   if [[ -n "${node_labels:-}" ]]; then
     flags+=" --node-labels=${node_labels}"
   fi
@@ -1843,7 +1843,8 @@ function get-env-val() {
 # Load the master env by calling get-master-env, and extract important values
 function parse-master-env() {
   # Get required master env vars
-  local master_env=$(get-master-env)
+  local master_env
+  master_env=$(get-master-env)
   KUBE_PROXY_TOKEN=$(get-env-val "${master_env}" "KUBE_PROXY_TOKEN")
   NODE_PROBLEM_DETECTOR_TOKEN=$(get-env-val "${master_env}" "NODE_PROBLEM_DETECTOR_TOKEN")
   CA_CERT_BASE64=$(get-env-val "${master_env}" "CA_CERT")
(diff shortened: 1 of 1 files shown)


cluster/gce/util.sh
```diff
@@ -743,8 +745,9 @@ function construct-common-kubelet-flags {
 # Sets KUBELET_ARGS with the kubelet flags for Linux nodes.
 # $1: if 'true', we're rendering flags for a master, else a node
 function construct-linux-kubelet-flags {
-  local node_type="$1"
-  local flags="$(construct-common-kubelet-flags)"
+  local node_type=$1
+  local flags
+  flags=$(construct-common-kubelet-flags)
```
spiffxp, line 750: why drop the quotes?

cluster/gce/util.sh
```diff
@@ -803,7 +806,8 @@ function construct-linux-kubelet-flags {
     flags+=" --non-masquerade-cidr=${NON_MASQUERADE_CIDR}"
   fi
   flags+=" --volume-plugin-dir=${VOLUME_PLUGIN_DIR}"
-  local node_labels="$(build-linux-node-labels ${node_type})"
+  local node_labels
+  node_labels=$(build-linux-node-labels "${node_type}")
```
spiffxp, line 810: same here, I'm not convinced this is the same as before?
```suggestion
  node_labels="$(build-linux-node-labels ${node_type})"
```

cluster/gce/util.sh
```diff
@@ -3833,7 +3862,8 @@ function test-teardown() {
     "${NODE_TAG}-http-alt" \
     "${NODE_TAG}-nodeports"
   if [[ ${MULTIZONE:-} == "true" && -n ${E2E_ZONES:-} ]]; then
-    local zones=( ${E2E_ZONES} )
+    local zones
+    while IFS= read -r line; do zones+=("$line"); done <<< "${E2E_ZONES}"
```
spiffxp, line 3866: are you sure this is the same as before? I feel like this new logic expects E2E_ZONES to be multi-line, and I don't think that's correct

spiffxp (changes requested):
There are a number of changes that end up dropping quoting, and it's not clear to me why they should

e.g. I would expect
```sh
local foo="$(my-awesome-cmd)"
```
to become
```sh
local foo
foo="$(my-awesome-cmd)"
```

cluster/gce/util.sh
joakimr-axis: Quotes are superfluous on assignments.

```
echo 'foo bar' > foobar.txt
a=$(cat foobar.txt)
b="$(cat foobar.txt)"
```

will result in `a` and `b` having the same content, `foo bar`.

On the other hand, there are indeed added quotes around `${node_type}` to silence shellcheck complaining about the not being quoted. In cases where `node_type` would contain a whitespace, this would bring a different behavior, but there is no such cases, is there? (One option is of course a `#shellcheck disable` statement and a comment explaining the disable.)

cluster/gce/util.sh
joakimr-axis: Yes, you are right. I will update this at once.

cluster/gce/util.sh
joakimr-axis: Quotes are superfluous on assignments.

cluster/gce/util.sh
joakimr-axis: Re-added quotes according to previous discussion (in other PR).

cluster/gce/util.sh
joakimr-axis: Re-added quotes according to previous discussion (in other PR).

joakimr-axis:
> There are a number of changes that end up dropping quoting, and it's not clear to me why they should
> 
> e.g. I would expect
> 
> ```shell
> local foo="$(my-awesome-cmd)"
> ```
> 
> to become
> 
> ```shell
> local foo
> foo="$(my-awesome-cmd)"
> ```

Fixed

spiffxp (approved):
/approve
/lgtm

Cheef5 (approved):
Lgtm

joakimr-axis:
> Lgtm

I had to surrender on the `local subnet` assignment that the CI pipeline just wouldn't accept. But The PR in its current state passes all the tests in the pipeline.

cluster/gce/util.sh
```diff
@@ -2342,6 +2350,8 @@ function create-subnetworks() {
 
   # Look for the alias subnet, it must exist and have a secondary
   # range configured.
+  # Deliberately declare and assign at the same time or the CI pipeline fails
```
BenTheElder, line 2353: huh? are we depending on masking this error?

BenTheElder:
thank you for continuing to work on this with immense patience ...

I think if we depend on masking this error we should go to seperate declare and assign but add `|| true`.

cluster/gce/util.sh
joakimr-axis: I just couldn't wrap my head around it. I fail to see why it does not work having the declaration and assignment separated, but repeated experiments with the CI pipeline suggests reality is something else. :-)

joakimr-axis:
> thank you for continuing to work on this with immense patience ...

:)

> I think if we depend on masking this error we should go to seperate declare and assign but add `|| true`.

You mean something like

```shell
local subnet
subnet=$(gcloud compute networks subnets describe \
    --project "${NETWORK_PROJECT}" \
    --region ${REGION} \
    ${IP_ALIAS_SUBNETWORK} 2>/dev/null)
    ${IP_ALIAS_SUBNETWORK} 2>/dev/null || true)
```

right?

Let's try!

joakimr-axis:
> Let's try!

It works! Thanks @BenTheElder!

Issue #87006, opened 2020-01-09
Cleanup converter to not support arbitrary conversions - just use the ones explicitly registered


wojtek-t:
cc @smarterclayton @liggitt

wojtek-t:
/remove-lifecycle stale

This is happening as we speak.

wojtek-t:
The code path to delete default-conversions has been disabled.

We will keep that in that state at least until the next release and will remove that code path then (it will allow us to simplify stuff a little bit too).

wojtek-t:
This is now fixed.

PR #94460, opened 2020-09-03
Remove default conversions

wojtek-t:
Ref https://github.com/kubernetes/kubernetes/issues/87006

```release-note
NONE
```

/kind cleanup
/priority important-longterm

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/conversion/converter.go
+++ b/staging/src/k8s.io/apimachinery/pkg/conversion/converter.go
@@ -26,16 +26,6 @@ type typePair struct {
 	dest   reflect.Type
 }
 
-type typeNamePair struct {
-	fieldType reflect.Type
-	fieldName string
-}
-
-// DebugLogger allows you to get debugging messages if necessary.
-type DebugLogger interface {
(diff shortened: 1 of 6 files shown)


staging/src/k8s.io/apimachinery/pkg/conversion/converter.go
```diff
@@ -192,117 +160,22 @@ func (c ConversionFuncs) Merge(other ConversionFuncs) ConversionFuncs {
 	return merged
 }
 
-// Meta is supplied by Scheme, when it calls Convert.
-type Meta struct {
-	// KeyNameMapping is an optional function which may map the listed key (field name)
-	// into a source and destination value.
-	KeyNameMapping FieldMappingFunc
-	// Context is an optional field that callers may use to pass info to conversion functions.
-	Context interface{}
```
liggitt, line 201: don't drop meta and change the signature in this PR, just drop the reflective conversion bits that were already dead code

wojtek-t: done

wojtek-t:
Thanks for the review - PTAL

PR #94463, opened 2020-09-03

joakimr-axis:
**What type of PR is this?**
/kind cleanup

ref: #72956 (part of #90405)
This patch uses shell internals instead of sed and awk where suitable and also uses lower/upper classes for tr instead of a-z and A-Z.

```release-note
NONE
```

### DIFF
--- a/cluster/gce/util.sh
+++ b/cluster/gce/util.sh
@@ -307,7 +307,8 @@ function upload-tars() {
   if which md5 > /dev/null 2>&1; then
     project_hash=$(md5 -q -s "$PROJECT")
   else
-    project_hash=$(echo -n "$PROJECT" | md5sum | awk '{ print $1 }')
+    project_hash=$(echo -n "$PROJECT" | md5sum)
+    project_hash=${project_hash%%[[:blank:]]*}
   fi
 
   # This requires 1 million projects before the probability of collision is 50%
@@ -526,12 +527,14 @@ function tars_from_version() {
     SERVER_BINARY_TAR_URL="https://storage.googleapis.com/kubernetes-release/release/${KUBE_VERSION}/kubernetes-server-linux-amd64.tar.gz"
     # TODO: Clean this up.
     KUBE_MANIFESTS_TAR_URL="${SERVER_BINARY_TAR_URL/server-linux-amd64/manifests}"
-    KUBE_MANIFESTS_TAR_HASH=$(curl ${KUBE_MANIFESTS_TAR_URL} --silent --show-error | ${sha512sum} | awk '{print $1}')
+    KUBE_MANIFESTS_TAR_HASH=$(curl "${KUBE_MANIFESTS_TAR_URL}" --silent --show-error | ${sha512sum})
+    KUBE_MANIFESTS_TAR_HASH=${KUBE_MANIFESTS_TAR_HASH%%[[:blank:]]*}
   elif [[ ${KUBE_VERSION} =~ ${KUBE_CI_VERSION_REGEX} ]]; then
     SERVER_BINARY_TAR_URL="https://storage.googleapis.com/kubernetes-release-dev/ci/${KUBE_VERSION}/kubernetes-server-linux-amd64.tar.gz"
     # TODO: Clean this up.
     KUBE_MANIFESTS_TAR_URL="${SERVER_BINARY_TAR_URL/server-linux-amd64/manifests}"
-    KUBE_MANIFESTS_TAR_HASH=$(curl ${KUBE_MANIFESTS_TAR_URL} --silent --show-error | ${sha512sum} | awk '{print $1}')
+    KUBE_MANIFESTS_TAR_HASH=$(curl "${KUBE_MANIFESTS_TAR_URL}" --silent --show-error | ${sha512sum})
+    KUBE_MANIFESTS_TAR_HASH=${KUBE_MANIFESTS_TAR_HASH%%[[:blank:]]*}
   else
     echo "Version doesn't match regexp" >&2
     exit 1
@@ -568,7 +571,7 @@ function get-master-env() {
 # "strip out quotes", and we really should be using a YAML library for
 # this, but PyYAML isn't shipped by default, and *rant rant rant ... SIGH*
 function yaml-quote {
-  echo "'$(echo "${@:-}" | sed -e "s/'/''/g")'"
+  echo "${@:-}" | sed -e "s/'/''/g;s/^/'/i;s/$/'/i"
 }
 
 # Writes the cluster location into a temporary file.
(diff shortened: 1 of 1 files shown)


cluster/gce/util.sh
```diff
@@ -520,12 +520,14 @@ function tars_from_version() {
     SERVER_BINARY_TAR_URL="https://storage.googleapis.com/kubernetes-release/release/${KUBE_VERSION}/kubernetes-server-linux-amd64.tar.gz"
     # TODO: Clean this up.
     KUBE_MANIFESTS_TAR_URL="${SERVER_BINARY_TAR_URL/server-linux-amd64/manifests}"
-    KUBE_MANIFESTS_TAR_HASH=$(curl ${KUBE_MANIFESTS_TAR_URL} --silent --show-error | ${sha1sum} | awk '{print $1}')
+    KUBE_MANIFESTS_TAR_HASH=$(curl "${KUBE_MANIFESTS_TAR_URL}" --silent --show-error | ${sha1sum})
+    KUBE_MANIFESTS_TAR_HASH=${KUBE_MANIFESTS_TAR_HASH%%[[:blank:]]*}
```
spiffxp, line 524: I would like to hear from other cluster/ approvers on this.  I personally find this harder to comprehend than the awk version

cluster/gce/util.sh
```diff
@@ -1202,7 +1204,7 @@ ADVANCED_AUDIT_WEBHOOK_INITIAL_BACKOFF: $(yaml-quote "${ADVANCED_AUDIT_WEBHOOK_I
 GCE_API_ENDPOINT: $(yaml-quote "${GCE_API_ENDPOINT:-}")
 GCE_GLBC_IMAGE: $(yaml-quote "${GCE_GLBC_IMAGE:-}")
 CUSTOM_INGRESS_YAML: |
-$(echo "${CUSTOM_INGRESS_YAML:-}" | sed -e "s/'/''/g")
+${CUSTOM_INGRESS_YAML//\'/\'\'}
```
spiffxp, line 1207: Same question here

cluster/gce/util.sh
```diff
@@ -1595,7 +1597,7 @@ function create-certs {
   # Determine extra certificate names for master
   local octets=($(echo "${SERVICE_CLUSTER_IP_RANGE}" | sed -e 's|/.*||' -e 's/\./ /g'))
   ((octets[3]+=1))
-  local -r service_ip=$(echo "${octets[*]}" | sed 's/ /./g')
+  local -r service_ip=${octets[*]// /}
```
spiffxp, line 1600: I don't think this is doing the same thing, how does this replace spaces with periods?

cluster/gce/util.sh
```diff
@@ -2835,7 +2837,7 @@ function detach-internal-master-ip() {
   # Detach ${MASTER_NAME}-internal-ip from ${name}
   gcloud compute instances network-interfaces update "${name}" --project "${PROJECT}" --zone "${zone}" --aliases="${aliases}"
   gcloud compute instances remove-metadata "${name}" --zone "${zone}" --keys=kube-master-internal-ip
-  run-gcloud-command "${name}" "${zone}" 'sudo ip route del to local '${ip}'/32 dev $(ip route | grep default | awk '\''{print $5}'\'')' || true
+  run-gcloud-command "${name}" "${zone}" "'sudo ip route del to local ${ip}/32 dev \$(ip route | grep default | while read -r d d d d d; do echo \$d; done)'" || true
```
spiffxp, line 2840: This also feels weirder than the awk version to me, what do others think?

spiffxp (changes requested):
I'm pretty sure one of these changes behavior.  The rest I would like to hear from others on whether the proposed new style is more maintainable

cluster/gce/util.sh
joakimr-axis: I guess that depends on how accustomed one is to `awk`. :-)
The `%%` replacement is a basic POSIX standard parameter expansion modifier, also clearly documented in the man pages.
And using shell internals makes the script more robust and portable (and efficient, even if that is not considered as needed).

cluster/gce/util.sh
joakimr-axis: To get rid of the shellcheck issues, it is either this or a `#shellcheck disable` statement since shellcheck will complain here.
Although not a big fan of Bash myself, I think if one has Bash at hand anyway, using its functionality comes natural. Using shell internals instead of external calls brings robustness/portability and speed too.

cluster/gce/util.sh
joakimr-axis: You are right, this is wrong. I looked into this creatuon of `service_ip` and found a much simpler way to make shellcheck happy about this with 2 lines of POSIX shell-internals. Coming up.

cluster/gce/util.sh
joakimr-axis: It is in fact pretty much a schoolbook use of `read`, IMO:
Read line-by-line, divide the 5 first entries by IFS to the variables `a b c d e` where `a-d` are of no interest to us since we want the 5th column only, make shellcheck happy by not having unused variables and use one variable (here: `d`) for all of them and it will contain the last value (5th column) for each line.

joakimr-axis: I have now updated this to use the perhaps less confusing

`while read -r _ _ _ _ dev; do echo \$dev; done`

instead.

spiffxp:
/cc @BenTheElder @cheftako @dims 
for input on https://github.com/kubernetes/kubernetes/pull/94463#pullrequestreview-502229746

dims:
going to give this a whirl to see if it sticks

/ok-to-test

cluster/gce/util.sh
BenTheElder: I think the underscore version might click more readily 👍

joakimr-axis: Now updated to

`while read -r _ _ _ _ dev _`

:-)

cluster/gce/util.sh
BenTheElder: the former version is not more readable imho, and this is definitely going to be faster and lower dependency.
imho: sed is useful when you're not storing in a variable, or doing many expressions, neither of which we need here.

I support this change.

BenTheElder:
/lgtm
thanks!

BenTheElder:
> W0129 21:57:02.578] /go/src/k8s.io/kubernetes/kubernetes/cluster/../cluster/../cluster/gce/util.sh: line 1122: CUSTOM_NETD_YAML: //\'/\'\': syntax error: operand expected (error token is "//\'/\'\'")

🙃

joakimr-axis:
> > W0129 21:57:02.578] /go/src/k8s.io/kubernetes/kubernetes/cluster/../cluster/../cluster/gce/util.sh: line 1122: CUSTOM_NETD_YAML: //'/'': syntax error: operand expected (error token is "//'/''")

Thanks, @BenTheElder! It was one of those classical cases with remaining unwanted characters (here `:`) that are so hard for the human brain to spot since we're evolved not to. Now it is fixed and the CI pipeline likes me. That LGTM badge got wiped away by chasing down that `:`, but now I feel content with that is in place here!

BenTheElder:
/lgtm
Thanks!

BenTheElder:
/retest
flaking on clone?

joakimr-axis:
> /retest
> flaking on clone?

Only 2 lines in the (failing) test log:

```shell
Cloning into 'test-infra'...
fatal: unable to access 'https://github.com/kubernetes/test-infra/': Could not resolve host: github.com
```

PR #94465, opened 2020-09-03

joakimr-axis:
**What type of PR is this?**
/kind cleanup

ref: #72956 (part of #90405)
This patch fixes the use of arrays.

```release-note
NONE
```

### DIFF
--- a/cluster/gce/util.sh
+++ b/cluster/gce/util.sh
@@ -1594,7 +1594,8 @@ function create-certs {
   local -r primary_cn="${1}"
 
   # Determine extra certificate names for master
-  local octets=($(echo "${SERVICE_CLUSTER_IP_RANGE}" | sed -e 's|/.*||' -e 's/\./ /g'))
+  local octets
+  read -r -a octets <<< "$(echo "${SERVICE_CLUSTER_IP_RANGE}" | sed -e 's|/.*||' -e 's/\./ /g')"
   ((octets[3]+=1))
   local -r service_ip=$(echo "${octets[*]}" | sed 's/ /./g')
   local sans=""
--- a/cluster/gce/util.sh
+++ b/cluster/gce/util.sh
@@ -1594,8 +1594,7 @@ function create-certs {
   local -r primary_cn="${1}"
 
   # Determine extra certificate names for master
-  local octets
-  kube::util::read-array octets < <(echo "${SERVICE_CLUSTER_IP_RANGE}" | sed -e 's|/.*||' -e 's/\./ /g')
+  local octets=($(echo "${SERVICE_CLUSTER_IP_RANGE}" | sed -e 's|/.*||' -e 's/\./ /g'))
   ((octets[3]+=1))
   local -r service_ip=$(echo "${octets[*]}" | sed 's/ /./g')
   local sans=""
@@ -2063,7 +2062,7 @@ function create-node-template() {
 
   local local_ssds=()
   local_ssd_ext_count=0
-  if [[ -n ${NODE_LOCAL_SSDS_EXT:-} ]]; then
+  if [[ -n "${NODE_LOCAL_SSDS_EXT:-}" ]]; then
     IFS=";" read -r -a ssdgroups <<< "${NODE_LOCAL_SSDS_EXT:-}"
     for ssdgroup in "${ssdgroups[@]}"
     do
--- a/cluster/gce/util.sh
+++ b/cluster/gce/util.sh
@@ -1594,7 +1594,8 @@ function create-certs {
   local -r primary_cn="${1}"
 
   # Determine extra certificate names for master
-  local octets=($(echo "${SERVICE_CLUSTER_IP_RANGE}" | sed -e 's|/.*||' -e 's/\./ /g'))
+  local octets
+  kube::util::read-array octets < <(echo "${SERVICE_CLUSTER_IP_RANGE}" | sed -e 's|/.*||' -e 's/\./ /g')
   ((octets[3]+=1))
   local -r service_ip=$(echo "${octets[*]}" | sed 's/ /./g')
   local sans=""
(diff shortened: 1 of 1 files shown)


phenixblue:
Has it been discussed how we want to handle things with testing for these scripts that have shellcheck fixes broken up into multiple PR's?

joakimr-axis:
> Has it been discussed how we want to handle things with testing for these scripts that have shellcheck fixes broken up into multiple PR's?

The script is run in the CI step, so do you mean checking if the amount of remaining shellcheck issues has indeed been reduced?

phenixblue:
Assuming the plan is to wait until all PR's for this file are merged, removing the script from `hack/.shellcheck_failures` on the final PR. What the order is if one PR needs to be merged before the other. Trying to see if there's an order of priority for getting these reviewed.

phenixblue:
I see most of the PR's from your comment [here](https://github.com/kubernetes/kubernetes/pull/90405#issuecomment-686323844)  have been merged, but from a quick glance I didn't see a PR that actually removed the script from `hack/.shellcheck_failures`.

joakimr-axis:
> I see most of the PR's from your comment [here](https://github.com/kubernetes/kubernetes/pull/90405#issuecomment-686323844) have been merged, but from a quick glance I didn't see a PR that actually removed the script from `hack/.shellcheck_failures`.

Please see the referenced PR #90405, in which the line `./cluster/gce/util.sh` is removed from `hack/.shellcheck_failures`:
 https://github.com/kubernetes/kubernetes/pull/90405/files

joakimr-axis:
> What the order is if one PR needs to be merged before the other.

The sub-PRs have been made to be independent of each other, and the merge order of them should not matter.

phenixblue:
> > I see most of the PR's from your comment [here](https://github.com/kubernetes/kubernetes/pull/90405#issuecomment-686323844) have been merged, but from a quick glance I didn't see a PR that actually removed the script from `hack/.shellcheck_failures`.
> 
> Please see the referenced PR #90405, in which the line `./cluster/gce/util.sh` is removed from `hack/.shellcheck_failures`:
> https://github.com/kubernetes/kubernetes/pull/90405/files

Ok, got it. It wasn't obvious that the original PR was still in play here. That clears things up.

joakimr-axis:
> Ok, got it. It wasn't obvious that the original PR was still in play here. That clears things up.

As some of the sub-PRs have now been merged, the size of the original PR has been reduced and is currently L. With all the sub PRs merged, it should then contain only a small amount of misc leftovers that were not to be categorized into a certain sub PR group.

cluster/gce/util.sh
```diff
@@ -2046,29 +2047,29 @@ function create-node-template() {
...
   local gcloud="gcloud"
 
-  local accelerator_args=""
+  local accelerator_args=()
   # VMs with Accelerators cannot be live migrated.
   # More details here - https://cloud.google.com/compute/docs/gpus/add-gpus#create-new-gpu-instance
-  if [[ ! -z "${NODE_ACCELERATORS}" ]]; then
-    accelerator_args="--maintenance-policy TERMINATE --restart-on-failure --accelerator ${NODE_ACCELERATORS}"
+  if [[ -n "${NODE_ACCELERATORS}" ]]; then
+    accelerator_args+=(--maintenance-policy TERMINATE --restart-on-failure --accelerator "${NODE_ACCELERATORS}")
```
phenixblue, line 2054: Should `kube::util::read-array` be used here for uniformity, or is it purposefully not used here?

cluster/gce/util.sh
```diff
@@ -2046,29 +2047,29 @@ function create-node-template() {
...
     gcloud="gcloud beta"
   fi
 
-  local preemptible_minions=""
+  local preemptible_minions=()
   if [[ "${PREEMPTIBLE_NODE}" == "true" ]]; then
-    preemptible_minions="--preemptible --maintenance-policy TERMINATE"
+    preemptible_minions+=(--preemptible --maintenance-policy TERMINATE)
```
phenixblue, line 2060: Same here

cluster/gce/util.sh
```diff
@@ -2046,29 +2047,29 @@ function create-node-template() {
...
+  if [[ -n ${NODE_LOCAL_SSDS_EXT:-} ]]; then
     IFS=";" read -r -a ssdgroups <<< "${NODE_LOCAL_SSDS_EXT:-}"
     for ssdgroup in "${ssdgroups[@]}"
     do
       IFS="," read -r -a ssdopts <<< "${ssdgroup}"
-      validate-node-local-ssds-ext "${ssdopts}"
+      validate-node-local-ssds-ext "${ssdopts[@]}"
       for ((i=1; i<=ssdopts[0]; i++)); do
-        local_ssds="$local_ssds--local-ssd=interface=${ssdopts[1]} "
+        local_ssds+=("--local-ssd=interface=${ssdopts[1]}")
```
phenixblue, line 2072: Ditto

cluster/gce/util.sh
```diff
@@ -2094,11 +2095,11 @@ function create-node-template() {
...
     "${IP_ALIAS_SIZE:-}")
 
-  local node_image_flags=""
+  local node_image_flags=()
   if [[ "${os}" == 'linux' ]]; then
-      node_image_flags="--image-project ${NODE_IMAGE_PROJECT} --image ${NODE_IMAGE}"
+      node_image_flags+=(--image-project "${NODE_IMAGE_PROJECT}" --image "${NODE_IMAGE}")
   elif [[ "${os}" == 'windows' ]]; then
-      node_image_flags="--image-project ${WINDOWS_NODE_IMAGE_PROJECT} --image ${WINDOWS_NODE_IMAGE}"
+      node_image_flags+=(--image-project "${WINDOWS_NODE_IMAGE_PROJECT}" --image "${WINDOWS_NODE_IMAGE}")
```
phenixblue, line 2102: Ditto

phenixblue (commented):
Just the one main question on the `kube::util::read-array` usage consistency.

cluster/gce/util.sh
joakimr-axis: `kube::util::read-array` reads line by line and adds it to the array, while here it is a single line that is added. I consider these two separate use cases where `kube::util::read-array` is the best solution for the case for line-by-line, being Bash 3 backwards compatible and all, while this way of writing is the easiest to use/follow/maintain for the single line case.

cluster/gce/util.sh
```diff
@@ -1593,7 +1593,8 @@ function create-certs {
   local -r primary_cn="${1}"
 
   # Determine extra certificate names for master
-  local octets=($(echo "${SERVICE_CLUSTER_IP_RANGE}" | sed -e 's|/.*||' -e 's/\./ /g'))
+  local octets
+  kube::util::read-array octets < <(echo "${SERVICE_CLUSTER_IP_RANGE}" | sed -e 's|/.*||' -e 's/\./ /g')
```
spiffxp, line 1597: I'm not sure this is the same behavior? My read is that `read-array` assumes SERVICE_CLUSTER_IP_RANGE is multi-line, where the previous method allows for space-separated terms

cluster/gce/util.sh
```diff
@@ -2046,29 +2047,29 @@ function create-node-template() {
...
   if [[ "${PREEMPTIBLE_NODE}" == "true" ]]; then
-    preemptible_minions="--preemptible --maintenance-policy TERMINATE"
+    preemptible_minions+=(--preemptible --maintenance-policy TERMINATE)
   fi
 
-  local local_ssds=""
+  local local_ssds=()
   local_ssd_ext_count=0
-  if [[ ! -z ${NODE_LOCAL_SSDS_EXT:-} ]]; then
+  if [[ -n ${NODE_LOCAL_SSDS_EXT:-} ]]; then
```
spiffxp, line 2065: https://google.github.io/styleguide/shellguide.html#quoting
```suggestion
  if [[ -n "${NODE_LOCAL_SSDS_EXT:-}" ]]; then
```

cluster/gce/util.sh
```diff
@@ -3691,12 +3693,12 @@ function check-resources() {
   echo "Looking for already existing resources"
   KUBE_RESOURCE_FOUND=""
 
-  if [[ -n "${INSTANCE_GROUPS[@]:-}" ]]; then
-    KUBE_RESOURCE_FOUND="Managed instance groups ${INSTANCE_GROUPS[@]}"
+  if [[ -n "${INSTANCE_GROUPS[*]:-}" ]]; then
```
spiffxp, line 3696: why?

spiffxp (changes requested):
Most of these look good, but there are a few I'm uncertain about

cluster/gce/util.sh
joakimr-axis: Thanks, as you probably noticed I am trying to smoke out what is the problematic change in the equation.
It sure looks like you are right. I will investigate.

cluster/gce/util.sh
joakimr-axis: > why?

https://github.com/koalaman/shellcheck/wiki/SC2199

I believe what we want to achieve here is the

`Alternatively, if you do want to concatenate all the values in the array into a single string for your test, use "$*" or "${array[*]}" to make this explicit.
`

usecase they describe.

cluster/gce/util.sh
joakimr-axis: Bullseye, @spiffxp! I did an experiment and left this change out and now all tests pass nicely.
I will look into an updated (=working) solution tomorrow; it is getting late in this neck of the woods.

joakimr-axis:
> Most of these look good, but there are a few I'm uncertain about

With the latest updates, things would be ok now, right?

spiffxp (approved):
/approve
/lgtm

PR #94474, opened 2020-09-03
DualStack: kubelet: do dual-stack iptables rules

danwinship:
**What type of PR is this?**
/kind bug

kubelet currently only creates a single set of iptables rules, meaning that in dual-stack clusters some necessary iptables rules may be missing. ("May" because kube-proxy redundantly creates some of the same rules as kubelet does, sometimes.)

Kubelet does not currently know whether the cluster is supposed to be single or dual-stack, so this patch just makes it always create dual-stack rules if the dual-stack feature gate is enabled. In some cases this means it will create IPv6 rules that aren't needed, but this isn't a big deal because it only creates a handful of rules. If https://github.com/kubernetes/enhancements/pull/1665 or something similar happens we may be able to be more precise here. (And if some "move networking stuff out of the kubelet" plan happens this might all move somewhere else anyway.)

Not sure if there are any issues specifically about this... It was most recently discussed in https://github.com/kubernetes/kubernetes/pull/94382#discussion_r481848900

```release-note
In dual-stack clusters, kubelet will now set up both IPv4 and IPv6 iptables rules, which may
fix some problems, eg with HostPorts.
```

/priority important-soon
/sig network
/cc aojea

### DIFF
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -35,9 +35,7 @@ import (
 
 	cadvisorapi "github.com/google/cadvisor/info/v1"
 	"k8s.io/mount-utils"
-	utilexec "k8s.io/utils/exec"
 	"k8s.io/utils/integer"
-	utilnet "k8s.io/utils/net"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -111,7 +109,6 @@ import (
 	"k8s.io/kubernetes/pkg/kubelet/volumemanager"
 	"k8s.io/kubernetes/pkg/security/apparmor"
 	sysctlwhitelist "k8s.io/kubernetes/pkg/security/podsecuritypolicy/sysctl"
-	utilipt "k8s.io/kubernetes/pkg/util/iptables"
 	"k8s.io/kubernetes/pkg/util/oom"
 	"k8s.io/kubernetes/pkg/util/selinux"
 	"k8s.io/kubernetes/pkg/volume"
@@ -466,11 +463,6 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 	}
 	httpClient := &http.Client{}
 	parsedNodeIP := net.ParseIP(nodeIP)
-	protocol := utilipt.ProtocolIPv4
-	if utilnet.IsIPv6(parsedNodeIP) {
-		klog.V(0).Infof("IPv6 node IP (%s), assume IPv6 operation", nodeIP)
-		protocol = utilipt.ProtocolIPv6
-	}
 
 	klet := &Kubelet{
 		hostname:                                hostname,
@@ -518,7 +510,6 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 		nodeIPValidator:                         validateNodeIP,
 		clock:                                   clock.RealClock{},
 		enableControllerAttachDetach:            kubeCfg.EnableControllerAttachDetach,
-		iptClient:                               utilipt.New(utilexec.New(), protocol),
 		makeIPTablesUtilChains:                  kubeCfg.MakeIPTablesUtilChains,
 		iptablesMasqueradeBit:                   int(kubeCfg.IPTablesMasqueradeBit),
 		iptablesDropBit:                         int(kubeCfg.IPTablesDropBit),
@@ -819,7 +810,6 @@ type Kubelet struct {
 	runtimeCache    kubecontainer.RuntimeCache
 	kubeClient      clientset.Interface
 	heartbeatClient clientset.Interface
-	iptClient       utilipt.Interface
 	rootDirectory   string
 
 	lastObservedNodeAddressesMux sync.RWMutex
--- a/pkg/kubelet/kubelet_network_linux.go
+++ b/pkg/kubelet/kubelet_network_linux.go
@@ -23,15 +23,39 @@ import (
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/wait"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/klog/v2"
+	"k8s.io/kubernetes/pkg/features"
 	utiliptables "k8s.io/kubernetes/pkg/util/iptables"
+	utilexec "k8s.io/utils/exec"
+	utilnet "k8s.io/utils/net"
 )
 
 func (kl *Kubelet) initNetworkUtil() {
-	kl.syncNetworkUtil()
-	go kl.iptClient.Monitor(utiliptables.Chain("KUBE-KUBELET-CANARY"),
-		[]utiliptables.Table{utiliptables.TableMangle, utiliptables.TableNAT, utiliptables.TableFilter},
-		kl.syncNetworkUtil, 1*time.Minute, wait.NeverStop)
+	exec := utilexec.New()
+
+	// At this point in startup we don't know the actual node IPs, so we configure dual stack iptables
+	// rules if the node _might_ be dual-stack, and single-stack based on requested nodeIP otherwise.
+	maybeDualStack := utilfeature.DefaultFeatureGate.Enabled(features.IPv6DualStack)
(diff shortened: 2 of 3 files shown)


pkg/kubelet/kubelet_network_linux.go
```diff
@@ -41,21 +41,6 @@ func (kl *Kubelet) initNetworkUtil() {
...
-		klog.Errorf("invalid iptables-masquerade-bit %v not in [0, 31]", kl.iptablesMasqueradeBit)
-		return
-	}
-
-	if kl.iptablesDropBit < 0 || kl.iptablesDropBit > 31 {
-		klog.Errorf("invalid iptables-drop-bit %v not in [0, 31]", kl.iptablesDropBit)
-		return
-	}
-
-	if kl.iptablesDropBit == kl.iptablesMasqueradeBit {
```
aojea, line 54: where is this `kl.iptablesDropBit == kl.iptablesMasqueradeBit` validated, 
I can see the 2 first checks here https://github.com/kubernetes/kubernetes/blob/fe8aa6eba8026f9562cfbd9d6ceee51a0939f4c4/pkg/kubelet/apis/config/validation/validation.go#L75-L80
Should we add this to validation too if we remove it here?

aojea: scratch my first comment, I found it,  it's here https://github.com/kubernetes/kubernetes/blob/fe8aa6eba8026f9562cfbd9d6ceee51a0939f4c4/pkg/kubelet/kubelet.go#L368

pkg/util/iptables/iptables.go
```diff
@@ -203,11 +203,10 @@ type runner struct {
 
 // newInternal returns a new Interface which will exec iptables, and allows the
 // caller to change the iptables-restore lockfile path
-func newInternal(exec utilexec.Interface, protocol Protocol, lockfilePath string) Interface {
+func newInternal(exec utilexec.Interface, protocol Protocol, lockfilePath string) (Interface, error) {
 	version, err := getIPTablesVersion(exec, protocol)
 	if err != nil {
-		klog.Warningf("Error checking iptables version, assuming version at least %s: %v", MinCheckVersion, err)
-		version = MinCheckVersion
+		return nil, err
```
aojea, line 209: should we differentiate between an error parsing the output from another error?
ie. return only if ExitStatus() != 0 and warn if we are not able to parse the version?

danwinship: We shouldn't ever have an error parsing the output though. If we can't parse the output, the most likely reason is that we aren't actually talking to an iptables binary...

I guess maybe they could change the format of "iptables --version" but that seems very unlikely

danwinship:
/auto-cc
didn't mean to _only_ have Antonio review this...

aojea:
/assign @thockin @khenidak 
We need this for dual stack, otherwise k8s without kube-proxy won't work in one of the IP families (unless they workaround creating the iptables rules)
this LGTM

cmd/kube-proxy/app/server_others.go
```diff
@@ -182,27 +185,34 @@ func newProxyServer(
 
 	klog.V(2).Info("DetectLocalMode: '", string(detectLocalMode), "'")
 
+	var ipt [2]utiliptables.Interface
+	dualStack := utilfeature.DefaultFeatureGate.Enabled(features.IPv6DualStack) && proxyMode != proxyModeUserspace
```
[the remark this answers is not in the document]
danwinship: `proxyModeKernelspace` is Windows-only and this file is non-Windows-only.

I debated "&& (iptables || ipvs)" vs "&& !userspace", but I figure probably we're going to want any other future new additions to be dual-stack-ready, so "!userspace" seemed right

pkg/kubelet/dockershim/network/kubenet/kubenet_linux.go
```diff
@@ -190,8 +196,13 @@ func (plugin *kubenetNetworkPlugin) ensureMasqRule() error {
 	if plugin.nonMasqueradeCIDR != zeroCIDRv4 && plugin.nonMasqueradeCIDR != zeroCIDRv6 {
 		// switch according to target nonMasqueradeCidr ip family
 		ipt := plugin.iptables
+		family := "IPv4"
```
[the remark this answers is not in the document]
danwinship: It's used as a display string not an API value... I wasn't sure about using someone else's constant in that way... though I guess there's no point in `utiliptables` declaring them that way if they don't want people printing them as strings... I'll see if we use them that way in other places

danwinship:
Updated to now only fail `utiliptables.New` if running `iptables --version` actually fails, _not_ if it succeeds but outputs a bogus version

pkg/kubelet/kubelet.go
```diff
@@ -466,10 +464,8 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 	}
 	httpClient := &http.Client{}
 	parsedNodeIP := net.ParseIP(nodeIP)
-	protocol := utilipt.ProtocolIPv4
 	if utilnet.IsIPv6(parsedNodeIP) {
 		klog.V(0).Infof("IPv6 node IP (%s), assume IPv6 operation", nodeIP)
```
aojea, line 468: nit, doubt, ...
this is not really true , right? it will depend on the dual stack flag the operation mode

aojea:
/retest
flake?
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/94474/pull-kubernetes-bazel-test/1308215316295192576

```
--- FAIL: TestDispatchEventWillNotBeBlockedByTimedOutWatcher (0.12s)
    cacher_whitebox_test.go:1012: watcher is blocked by slower one (count: 34)
W0922 01:53:18.603645  178038 reflector.go:427] k8s.io/apiserver/pkg/storage/cacher/watch_cache_test.go:503: watch of *v1.Pod ended with: very short watch: k8s.io/apiserver/pkg/storage/cacher/watch_cache_test.go:503: Unexpected watch close - watch lasted less than a second and no items received
```

pkg/kubelet/kubelet.go
danwinship: single-stack IPv6 is not feature-gated

aojea: sorry, I added the comment without context, it maybe dualstack or it may be ipv6 only at this point

danwinship: ah. Yeah, and the comment isn't really very useful... "You explicitly passed an IPv6 `--node-ip`, so just FYI, I'm going to assume you want to use IPv6".

dcbw:
@danwinship verify failed:
```
diff -Naupr -x _output /home/prow/go/src/k8s.io/kubernetes/pkg/kubelet/BUILD /tmp/verify-bazel.ZFzOhN/go/src/k8s.io/kubernetes/pkg/kubelet/BUILD
--- /home/prow/go/src/k8s.io/kubernetes/pkg/kubelet/BUILD	2020-09-30 13:58:17.173626513 +0000
+++ /tmp/verify-bazel.ZFzOhN/go/src/k8s.io/kubernetes/pkg/kubelet/BUILD	2020-09-30 13:59:31.322834411 +0000
@@ -153,10 +153,15 @@ go_library(
         "//vendor/k8s.io/klog/v2:go_default_library",
         "//vendor/k8s.io/utils/exec:go_default_library",
         "//vendor/k8s.io/utils/integer:go_default_library",
-        "//vendor/k8s.io/utils/net:go_default_library",
         "//vendor/k8s.io/utils/path:go_default_library",
         "//vendor/k8s.io/utils/strings:go_default_library",
     ] + select({
+        "@io_bazel_rules_go//go/platform:android": [
+            "//vendor/k8s.io/utils/net:go_default_library",
+        ],
+        "@io_bazel_rules_go//go/platform:linux": [
+            "//vendor/k8s.io/utils/net:go_default_library",
+        ],
         "@io_bazel_rules_go//go/platform:windows": [
             "//pkg/kubelet/winstats:go_default_library",
         ],

Run ./hack/update-bazel.sh
+++ exit code: 1
+++ error: 1
[0;31mFAILED[0m   verify-bazel.sh	20s
```

thockin:
Thanks!

/lgtm
/approve

PR #94479, opened 2020-09-03
kubeadm: apply resource requests in the etcd pod spec

### LINKED ISSUE #2195: etcd pod spec should include requests for memory, cpu, etc
## FEATURE REQUEST

The pod spec for etcd does not include any `requests`:

https://github.com/kubernetes/kubernetes/blob/db8a88721e587cbc54e7081cb88374e253cde4ed/cmd/kubeadm/app/phases/etcd/local.go#L222-L232

This has the following consequences:
 * etcd is not included in kubelet's bookkeeping of how much should run on this node
 * when machine CPU is at 100%, etcd will be throttled to maximum 0.2%
 * ~when machine is low on memory, etcd will be a candidate to terminate~ (Not in 1.19+)

kubeadm should add `requests` for `cpu`, `memory` and `ephemeral_storage`.

It is impossible to pick "correct" values for a component that needs more resources as your cluster gets bigger, but I submit that we can do better by default than zero.

knight42:

**What type of PR is this?**

/kind feature

Fixes https://github.com/kubernetes/kubeadm/issues/2195

```release-note
kubeadm: make etcd pod request 100m CPU, 100Mi memory and 100Mi ephemeral_storage by default
```

```docs

```

### DIFF
--- a/cmd/kubeadm/app/phases/etcd/local.go
+++ b/cmd/kubeadm/app/phases/etcd/local.go
@@ -26,17 +26,18 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"k8s.io/klog/v2"
-
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
+	utilsnet "k8s.io/utils/net"
+
 	kubeadmapi "k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm"
 	kubeadmconstants "k8s.io/kubernetes/cmd/kubeadm/app/constants"
 	"k8s.io/kubernetes/cmd/kubeadm/app/images"
 	kubeadmutil "k8s.io/kubernetes/cmd/kubeadm/app/util"
 	etcdutil "k8s.io/kubernetes/cmd/kubeadm/app/util/etcd"
 	staticpodutil "k8s.io/kubernetes/cmd/kubeadm/app/util/staticpod"
-	utilsnet "k8s.io/utils/net"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


cmd/kubeadm/app/phases/etcd/local.go
```diff
@@ -247,6 +248,14 @@ func GetEtcdPodSpec(cfg *kubeadmapi.ClusterConfiguration, endpoint *kubeadmapi.A
 				staticpodutil.NewVolumeMount(etcdVolumeName, cfg.Etcd.Local.DataDir, false),
 				staticpodutil.NewVolumeMount(certsVolumeName, cfg.CertificatesDir+"/etcd", false),
 			},
+			Resources: v1.ResourceRequirements{
```
knight42, line 251: @neolit123 looks like we don't need to use `staticpodutil.ComponentResources` here, WDYT?

neolit123: your are correct.
we don't need the utility function and the method i outlined in the issue.

cmd/kubeadm/app/phases/etcd/local.go
```diff
@@ -247,6 +248,14 @@ func GetEtcdPodSpec(cfg *kubeadmapi.ClusterConfiguration, endpoint *kubeadmapi.A
...
+				Requests: v1.ResourceList{
+					// TODO: make these values configurable
```
neolit123, line 253: please remove this TODO for now.
users can use the --experimental-patches functionality to apply patches during init/join/upgrade.
the feature can one day be in the kubeadm config too.

cmd/kubeadm/app/phases/etcd/local.go
```diff
@@ -26,17 +26,18 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"k8s.io/klog/v2"
-
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
```
neolit123, line 32: this import should remain on top under pkg/errors.

neolit123:
/retitle kubeadm: apply resource requests in the etcd pod spec
/approve

neolit123:
/cc @rosti @ereslibre 
PTAL

neolit123:
we didn't get immediate objections on https://github.com/kubernetes/kubeadm/issues/2195 so let's see how this goes.

/hold cancel
/milestone v1.20

neolit123:
/lgtm
thanks @knight42

PR #94480, opened 2020-09-03
Remove FieldMatchingFlags

wojtek-t:
Ref https://github.com/kubernetes/kubernetes/issues/87006

```release-note
NONE
```

/kind cleanup
/priority important-longterm

### DIFF
--- a/staging/src/k8s.io/code-generator/cmd/conversion-gen/generators/conversion.go
+++ b/staging/src/k8s.io/code-generator/cmd/conversion-gen/generators/conversion.go
@@ -812,17 +812,23 @@ func (g *genConversion) doMap(inType, outType *types.Type, sw *generator.Snippet
 				sw.Do("$.|raw$(val)\n", outType.Elem)
 			}
 		} else {
-			sw.Do("newVal := new($.|raw$)\n", outType.Elem)
+			conversionExists := true
 			if function, ok := g.preexists(inType.Elem, outType.Elem); ok {
+				sw.Do("newVal := new($.|raw$)\n", outType.Elem)
 				sw.Do("if err := $.|raw$(&val, newVal, s); err != nil {\n", function)
 			} else if g.convertibleOnlyWithinPackage(inType.Elem, outType.Elem) {
+				sw.Do("newVal := new($.|raw$)\n", outType.Elem)
 				sw.Do("if err := "+nameTmpl+"(&val, newVal, s); err != nil {\n", argsFromType(inType.Elem, outType.Elem))
 			} else {
-				sw.Do("// TODO: Inefficient conversion - can we improve it?\n", nil)
-				sw.Do("if err := s.Convert(&val, newVal, 0); err != nil {\n", nil)
+				args := argsFromType(inType.Elem, outType.Elem)
+				sw.Do("// FIXME: Provide conversion function to convert $.inType|raw$ to $.outType|raw$\n", args)
+				sw.Do("compileErrorOnMissingConversion()\n", nil)
+				conversionExists = false
 			}
-			sw.Do("return err\n", nil)
-			sw.Do("}\n", nil)
+			if conversionExists {
+				sw.Do("return err\n", nil)
+				sw.Do("}\n", nil)
+			}
 			if inType.Key == outType.Key {
 				sw.Do("(*out)[key] = *newVal\n", nil)
 			} else {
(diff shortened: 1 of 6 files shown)


staging/src/k8s.io/code-generator/cmd/conversion-gen/generators/conversion.go
```diff
@@ -812,21 +813,27 @@ func (g *genConversion) doMap(inType, outType *types.Type, sw *generator.Snippet
...
+				sw.Do("newVal := new($.|raw$)\n", outType.Elem)
 				sw.Do("if err := $.|raw$(&val, newVal, s); err != nil {\n", function)
 			} else if g.convertibleOnlyWithinPackage(inType.Elem, outType.Elem) {
+				sw.Do("newVal := new($.|raw$)\n", outType.Elem)
 				sw.Do("if err := "+nameTmpl+"(&val, newVal, s); err != nil {\n", argsFromType(inType.Elem, outType.Elem))
 			} else {
-				sw.Do("// TODO: Inefficient conversion - can we improve it?\n", nil)
-				sw.Do("if err := s.Convert(&val, newVal, 0); err != nil {\n", nil)
+				args := argsFromType(inType.Elem, outType.Elem)
+				sw.Do("// FIXME: Provide conversion function to convert $.inType|raw$ to $.outType|raw$", args)
```
roycaihw, line 825: (I guess you know it already :) ) A trailing `\n` is missing in these FIXME comments. Fixing these comments would mask the actual problem in the test process though. I tested locally and `make verify` passed.

wojtek-t:
/assign @liggitt 
Jordan - please take a look when you will have a chance.

staging/src/k8s.io/code-generator/cmd/conversion-gen/generators/conversion.go
```diff
@@ -971,14 +975,18 @@ func (g *genConversion) doStruct(inType, outType *types.Type, sw *generator.Snip
...
 				continue
 			}
+			conversionExists := true
 			if g.convertibleOnlyWithinPackage(inMemberType, outMemberType) {
 				sw.Do("if err := "+nameTmpl+"(&in.$.name$, &out.$.name$, s); err != nil {\n", args)
 			} else {
-				sw.Do("// TODO: Inefficient conversion - can we improve it?\n", nil)
-				sw.Do("if err := s.Convert(&in.$.name$, &out.$.name$, 0); err != nil {\n", args)
+				args := argsFromType(inMemberType, outMemberType)
+				sw.Do("// FIXME: Provide conversion function to convert $.inType|raw$ to $.outType|raw$\n", args)
```
deads2k, line 983: I don't think this is a strong enough marker.  I would like compilation of the generated code to fail so that these problems are caught quickly.  I think we're going to hit at least two in openshift and I doubt we'd find it easily without a compilation error.

deads2k:
I'm in favor of the change, but I think the generator needs to fail more forcefully.  I think it will actually be easier to find and fix if we generate code that fails compilation so that the author knows exactly where to look.

staging/src/k8s.io/code-generator/cmd/conversion-gen/generators/conversion.go
wojtek-t: I added:
sw.Do("compileErrorOnMissingConversion()\n")

in all places which will throw you a compile error.
Does that address your concern?

wojtek-t:
It seems we still have non-determinism in our verify scripts - I'm following up on those errors in: https://github.com/kubernetes/kubernetes/issues/94691

deads2k:
Looks like verify is failing related to this PR.  Also, that compile error may have saved our bacon.

```
pkg/scheduler/apis/config/v1beta1/zz_generated.conversion.go:293:2: undefined: compileErrorOnMissingConversion
pkg/scheduler/apis/config/v1beta1/zz_generated.conversion.go:295:2: undefined: compileErrorOnMissingConversion
pkg/scheduler/apis/config/v1beta1/zz_generated.conversion.go:303:2: undefined: compileErrorOnMissingConversion
pkg/scheduler/apis/config/v1beta1/zz_generated.conversion.go:331:2: undefined: compileErrorOnMissingConversion
pkg/scheduler/apis/config/v1beta1/zz_generated.conversion.go:333:2: undefined: compileErrorOnMissingConversion
pkg/scheduler/apis/config/v1beta1/zz_generated.conversion.go:341:2: undefined: compileErrorOnMissingConversion
```

from https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/94480/pull-kubernetes-verify/1321753498307530752

wojtek-t:
This is blocked on https://github.com/kubernetes/kubernetes/issues/96015

wojtek-t:
The third commit seems to address the problem - just put that here to confirm.
But we should submit it first independently from this PR, so holding this one until this happens.

/hold

PR #94481, opened 2020-09-03
Fix custom metrics conversion functions

wojtek-t:
Ref #87006

```release-note
Fix conversions for custom metrics.
```

/kind cleanup
/priority important-longterm

### DIFF
--- a/staging/src/k8s.io/metrics/pkg/apis/custom_metrics/v1beta2/zz_generated.conversion.go
+++ b/staging/src/k8s.io/metrics/pkg/apis/custom_metrics/v1beta2/zz_generated.conversion.go
@@ -124,8 +124,7 @@ func Convert_custom_metrics_MetricListOptions_To_v1beta2_MetricListOptions(in *c
 }
 
 func autoConvert_v1beta2_MetricValue_To_custom_metrics_MetricValue(in *MetricValue, out *custommetrics.MetricValue, s conversion.Scope) error {
-	// TODO: Inefficient conversion - can we improve it?
-	if err := s.Convert(&in.DescribedObject, &out.DescribedObject, 0); err != nil {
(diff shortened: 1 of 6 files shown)


wojtek-t:
@liggitt - PTAL when you will have a chance

staging/src/k8s.io/metrics/pkg/apis/custom_metrics/v1beta2/zz_generated.conversion.go
```diff
@@ -124,8 +124,7 @@ func Convert_custom_metrics_MetricListOptions_To_v1beta2_MetricListOptions(in *c
 }
 
 func autoConvert_v1beta2_MetricValue_To_custom_metrics_MetricValue(in *MetricValue, out *custommetrics.MetricValue, s conversion.Scope) error {
-	// TODO: Inefficient conversion - can we improve it?
-	if err := s.Convert(&in.DescribedObject, &out.DescribedObject, 0); err != nil {
```
liggitt, line 128: was this a bug? would this call actually work in 1.19?

wojtek-t: I thought it would when I was changing that, but you're right - it was a bug.
scope.Convert calls converter:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/conversion/converter.go#L183
which in turn were only looking into registered conversions:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/conversion/converter.go#L289

which doesn't exist (until this PR).

So yes - it seemed to be a 1.19 bug (it also means we have 0 tests for custom metrics...)

PR #94483, opened 2020-09-03
Improve ability to debug Events e2e failure

derekwaynecarr:
**What type of PR is this?**
/kind cleanup

Improve ability to debug flakes we see in our downstream CI setup for Event testing on some platforms.
see: https://bugzilla.redhat.com/show_bug.cgi?id=1846529

This lets us see why kubelets are not able to send an event on earlier attempts.
Related test: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/node/events.go

```release-note
NONE
```


smarterclayton:
Since the error is already redundant, displaying the extra info doesn't seem like a problem.

/lgtm
/approve

PR #94484, opened 2020-09-03
fix documentation and maybe validation of NetworkPolicy CIDRs

danwinship:
**What type of PR is this?**
/kind bug
/kind api-change

Currently it is possible to create a NetworkPolicy with an invalid `ipBlock` CIDR such as

    ipBlock:
      cidr: 192.168.1.1/24

For purposes of matching/filtering traffic, it only makes sense to allow CIDR strings where all bits outside of the mask are 0; ie, this should be either 192.168.1.**0**/24 (ie, "192.168.1.1 and everything else in the same subnet") or 192.168.1.1/**32** ("just 192.168.1.1"). Either of those CIDR strings could plausibly be used in a NetworkPolicy, depending on what the user wanted, so there is no way to just guess which version the user meant.

Network plugins encountering this ambiguous syntax might:
- Intentionally or accidentally interpret it as "192.168.1.0/24"
- Intentionally or accidentally interpret it as "192.168.1.1/32"
- Pass the string unchanged to some lower-level software (eg iptables), which will then do one of the above
- Notice the problem and ignore the `ipBlock` rather than possibly allowing incorrect traffic

This PR makes the apiserver reject NetworkPolicies like the above as invalid. I realize this might be considered an API break and not possible. I guess in that case plan B would probably be to define a specific required interpretation of such CIDR values, and update the e2e tests to enforce that. The safest, most-restrictive intepretation would be:

- If `cidr` is ambiguous:
  - if there are `except` values, then interpret the `cidr` value as a subnet, not a host (since it wouldn't make sense to have exceptions if `cidr` only identified a single IP)
  - otherwise interpret the `cidr` as a host
- If any `except` item is ambiguous, interpret it as a subnet 

That ensures that the policy might block traffic the user meant to allow, but it will not allow traffic the user meant to block.

None

```release-note
NetworkPolicy validation is now more restrictive, and policies that were previously
allowed may now be rejected for containing invalid CIDR match strings such as
"192.168.1.1/24" in ipBlocks. These need to be replaced with the correct values (eg,
"192.168.1.0/24" or "192.168.1.1/32" depending on which was meant.)
```

/sig network
/priority important-soon

### DIFF
--- a/pkg/apis/networking/types.go
+++ b/pkg/apis/networking/types.go
@@ -144,15 +144,15 @@ type NetworkPolicyPort struct {
 	Port *intstr.IntOrString
 }
 
-// IPBlock describes a particular CIDR (Ex. "192.168.1.1/24","2001:db9::/64") that is allowed
+// IPBlock describes a particular CIDR (Ex. "192.168.1.0/24","2001:db9::/64") that is allowed
 // to the pods matched by a NetworkPolicySpec's podSelector. The except entry describes CIDRs
 // that should not be included within this rule.
 type IPBlock struct {
 	// CIDR is a string representing the IP Block
-	// Valid examples are "192.168.1.1/24" or "2001:db9::/64"
+	// Valid examples are "192.168.1.0/24" or "2001:db9::/64"
 	CIDR string
 	// Except is a slice of CIDRs that should not be included within an IP Block
-	// Valid examples are "192.168.1.1/24" or "2001:db9::/64"
+	// Valid examples are "192.168.1.1/32" or "2001:db9::/64"
 	// Except values will be rejected if they are outside the CIDR range
 	// +optional
 	Except []string
--- a/pkg/apis/networking/validation/validation.go
+++ b/pkg/apis/networking/validation/validation.go
@@ -164,14 +164,28 @@ func ValidateNetworkPolicyUpdate(update, old *networking.NetworkPolicy) field.Er
 	return allErrs
 }
 
-// ValidateIPBlock validates a cidr and the except fields of an IpBlock NetworkPolicyPeer
+// validateCIDRMask validates that a string is a valid CIDR string with no bits set in the masked-out part
+// (eg, "192.168.1.0/24" is allowed, but "192.168.1.1/24" is not)
+func validateCIDRMask(cidr string) (*net.IPNet, error) {
+	ip, net, err := net.ParseCIDR(cidr)
+	if err != nil {
+		return nil, err
+	}
+	if !ip.Equal(net.IP) {
+		maskLen, addrLen := net.Mask.Size()
+		return nil, fmt.Errorf("%q is not in canonical form (should be %s/%d or %s/%d?)", cidr, ip.Mask(net.Mask).String(), maskLen, ip.String(), addrLen)
+	}
+	return net, nil
+}
+
+// ValidateIPBlock validates a cidr and the except fields of an IPBlock NetworkPolicyPeer
 func ValidateIPBlock(ipb *networking.IPBlock, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 	if len(ipb.CIDR) == 0 || ipb.CIDR == "" {
 		allErrs = append(allErrs, field.Required(fldPath.Child("cidr"), ""))
 		return allErrs
 	}
-	cidrIPNet, err := apivalidation.ValidateCIDR(ipb.CIDR)
+	cidrIPNet, err := validateCIDRMask(ipb.CIDR)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


caseydavenport:
Hm, yeah this is a tricky one. I'm not sure how to make the call on API breaking or not. In the strictest sense I think this would be an API break. I believe we've made similar changes elsewhere that broke some users under the guise of a bugfix.

If we decide this is an API breaking change and that we can't tighten the validation, then my preference would be to pick a single interpretation of the mixed format (rather than trying to assume the user's intent). e.g., always interpret it as a CIDR rather than a host.

danwinship:
> my preference would be to pick a single interpretation of the mixed format (rather than trying to assume the user's intent) e.g., always interpret it as a CIDR rather than a host.

FTR that's consistent with the way that the validation code currently parses things. eg, this passes validation:

    - cidr: 192.168.1.1/24
      except:
        - 192.168.1.2/32

because the validation code treats `cidr` as though it was 192.168.1.0/24, and 192.168.1.2/32 fits inside that. But this does not validate:

    - cidr: 192.168.1.0/24
      except:
        - 192.168.1.2/24

because the `except` clause would be parsed as `192.168.1.0/24`, which is equal to the `cidr` rather than being a proper subset of it, which is not allowed.

pkg/apis/networking/types.go
```diff
@@ -144,15 +144,15 @@ type NetworkPolicyPort struct {
 	Port *intstr.IntOrString
 }
 
-// IPBlock describes a particular CIDR (Ex. "192.168.1.1/24","2001:db9::/64") that is allowed
+// IPBlock describes a particular CIDR (Ex. "192.168.1.0/24","2001:db9::/64") that is allowed
```
jayunit100, line 147: can we also add the fact that `IPBlock` generally applies to endpoints?
some folks have tried to use this to target non-pod services, which can lead
to alot of confusion

thockin:
> Currently it is possible to create a NetworkPolicy with an invalid ipBlock CIDR such as

I'm going to go ahead and disagree that this is "invalid".  It's unambiguous what it meant, even if it is a bit jarring to a human.  Go parses it, though RFC 4632 suggests that trailing bits should be zero.

> Pass the string unchanged to some lower-level software (eg iptables), which will then do one of the above

Any software that mis-interprets the meaning here is broken, IMO.  

>  I realize this might be considered an API break and not possible.

Correct.  Worse, we really have no protocol to ratchet validation to be more restrictive.  So if we wanted to do this, it would have to be opt-in basically forever.

>  interpret the cidr value as a subnet, not a host 

There's no ambiguity.  We do not allow non-CIDR-notation values (see `ValidateCIDR()`) so SOMEONE either said "1.2.3.4/24" or "1.2.3.4/32" - the former is CLEARLY a subnet.  The latter is CLEARLY a "host" (a subnet of 1 IP).  I will die on this hill.

> But this does not validate:
> 
> ```
> - cidr: 192.168.1.0/24
>   except:
>     - 192.168.1.2/24
> ```

And it shouldn't!  those are the same subnet and that isn't allowed.

thockin:
FWIW, I wouldn't object to changing all docs to recommend strict notation, and even to throw an even when creating/updating non-strict notation, and MAYBE even adding status and conditions to indicate non-strict notation.

danwinship:
> > Currently it is possible to create a NetworkPolicy with an invalid ipBlock CIDR such as
> 
> I'm going to go ahead and disagree that this is "invalid". It's unambiguous what it meant, even if it is a bit jarring to a human. Go parses it, though RFC 4632 suggests that trailing bits should be zero.

It's _not_ unambiguous, and to say that "Go parses it" is misleading; `net.ParseCIDR("192.168.1.1/24")` returns *two* values; a `net.IP` corresponding to `192.168.1.1` and a `net.IPNet` corresponding to `191.268.1.0/24`. 

> Any software that mis-interprets the meaning here is broken, IMO.

Unix userspace tools were using CIDR notation with non-0 trailing bits for a long time before Kubernetes came around:

```
$ ip addr show dev wlp4s0
3: wlp4s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether d4:25:8b:b8:b0:28 brd ff:ff:ff:ff:ff:ff
    inet 192.168.0.91/16 brd 192.168.255.255 scope global dynamic noprefixroute wlp4s0
       valid_lft 11439sec preferred_lft 11439sec
```

thockin:
Regardless of how we feel about it, the API exists and can't really be
changed.  We can change all of the examples, we can update the docs, but we
can't break users.

One option might be to return a warning if there are any bits set on the
RHS of the mask.  Is that worth the effort?

PR #80470, opened 2019-07-23
fix autogenerated container port name conflicts with multiple hostports

zouyee:
**What type of PR is this?**
/kind bug
/area kubelet

**What this PR does / why we need it**:
fix autogenerated container port name conflicts with multiple hostports

**Which issue(s) this PR fixes**:

Fixes #63906

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #94494, opened 2020-09-03
Allow to map the same container port to different host ports

### LINKED ISSUE #63906: Autogenerated container port name conflicts with multiple hostports

/kind bug

**What happened**:

A container ports spec such as:
```
        ports:
        - containerPort: 4443
          hostPort: 443
        - containerPort: 4443
          hostPort: 4443
```
Results in this error:
```
kubelet[11270]: W0515 17:44:41.623947   11270 helpers.go:352] Port name conflicted, "cntr-TCP:4443" is defined more than once
```
From https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/container/helpers.go#L321 . Because of this, the port mappings are not setup correctly on the host when using a CNI-compliant portmap solution.

Auto-generated port names should never result in duplicates

[d2.yml.txt](https://github.com/kubernetes/kubernetes/files/1897012/d2.yml.txt)

/sig node

**Environment**:
Client Version: version.Info{Major:"1", Minor:"8", GitVersion:"v1.8.7", GitCommit:"b30876a5539f09684ff9fde266fda10b37738c9c", GitTreeState:"clean", BuildDate:"2018-01-16T21:59:57Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"8", GitVersion:"v1.8.11", GitCommit:"1df6a8381669a6c753f79cb31ca2e3d57ee7c8a3", GitTreeState:"clean", BuildDate:"2018-04-05T17:16:46Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}
- Cloud provider or hardware configuration: VirtualBox
- OS (e.g. from /etc/os-release): Ubuntu
- Kernel (e.g. `uname -a`): Linux node2 4.4.0-104-generic #127-Ubuntu SMP Mon Dec 11 12:16:42 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
- Install tools: kubeadm
- Others:

SergeyKanzhelev:
**What type of PR is this?**

/kind bug

This PR is a replacement for https://github.com/kubernetes/kubernetes/pull/80470 (by @zouyee). It relaxes the validation in kubelet before passing node port mappings to CNI. Stated use case for this is to allow port transition for daemons. So both ports (old and new) are configured for the transition period and after some time old port being removed.

Fixes  #63906

There is a question of whether this deduplication of port mappings is needed at all. For example, once the mapping is named, no duplication is happening and needs to be handled by CNI anyway. But removing this deduplication is a bigger change that may require a larger announcement and can potentially affect payload.

This PR also includes the relaxing of deduplication by hostIP. Similar to https://github.com/kubernetes/kubernetes/pull/94382 Same port is allows to be configured on different IPs.

```release-note
port mapping allows to map the same `containerPort` to multiple `hostPort` without naming the mapping explicitly.
```

```docs

```

### DIFF
--- a/pkg/kubelet/container/helpers.go
+++ b/pkg/kubelet/container/helpers.go
@@ -335,7 +335,7 @@ func MakePortMappings(container *v1.Container) (ports []PortMapping) {
 
 		var name string = p.Name
 		if name == "" {
-			name = fmt.Sprintf("%s-%s:%d", family, p.Protocol, p.ContainerPort)
+			name = fmt.Sprintf("%s-%s-%s:%d:%d", family, p.Protocol, p.HostIP, p.ContainerPort, p.HostPort)
 		}
 
 		// Protect against a port name being used more than once in a container.
--- a/pkg/kubelet/container/helpers_test.go
+++ b/pkg/kubelet/container/helpers_test.go
@@ -580,8 +580,6 @@ func TestMakePortMappings(t *testing.T) {
 					port("foo", v1.ProtocolUDP, 555, 5555, ""),
 					// Duplicated, should be ignored.
 					port("foo", v1.ProtocolUDP, 888, 8888, ""),
-					// Duplicated, should be ignored.
-					port("", v1.ProtocolTCP, 80, 8888, "127.0.0.1"),
 					// Duplicated with different address family, shouldn't be ignored
 					port("", v1.ProtocolTCP, 80, 8080, "::"),
 					// No address family specified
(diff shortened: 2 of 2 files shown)


pkg/kubelet/container/helpers.go
```diff
@@ -335,7 +335,7 @@ func MakePortMappings(container *v1.Container) (ports []PortMapping) {
 
 		var name string = p.Name
 		if name == "" {
-			name = fmt.Sprintf("%s-%s:%d", family, p.Protocol, p.ContainerPort)
+			name = fmt.Sprintf("%s-%s-%s:%d:%d", family, p.Protocol, p.HostIP, p.ContainerPort, p.HostPort)
```
SergeyKanzhelev, line 338: @thockin this name now does not leave the function. Only used inside this func to deduplicate entries

thockin:
This doesn't fully replace #94382 though, right?  That has a lot more deltas.

thockin:
Thanks!

/lgtm
/approve

SergeyKanzhelev:
> This doesn't fully replace #94382 though, right? That has a lot more deltas.

You are correct. That one is separate - it's about actual logic applying it. This PR is about mapping pre-validation by kubelet.

Issue #76714, opened 2019-04-17
PKI creation using now.UTC breaks after system clock drift corrections


mikeparker:
We're tracking a rare bug in Docker Desktop where users are getting the error message:
`the certificate is not valid yet` (see: https://github.com/docker/for-win/issues/2913)

This can be reproduced by manually changing your system clock forwards and backwards, but we believe it can also occur naturally when a system clock drifts forwards and a reboot or other sync corrects the system clock backwards again. 

We're using the code here to generate our kubernetes certificates:
https://github.com/kubernetes/client-go/blob/master/util/cert/cert.go#L65
which uses `time.Now` and `NotBefore : now.UTC()`

Some options:
a) Switch away from this utility function and write our own code to provide similar functionality
b) Open a PR here to provide flexibility in the NotBefore time, either setting the NotBefore time to a long time ago (e.g. 1970), providing an optional grace period
c) Open a PR here to change the default NotBefore time to use a hard coded grace period or set start time in the past

WDYT?

mikeparker:
From previous thread, @liggitt said:
> it looks like https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/pkiutil/pki_helpers.go#L249-L256 does client-side validation, which is problematic for the clock drift reasons mentioned. it should not.
> 
> it is also reasonable to allow _some_ buffer in the NotBefore time (c.f. https://github.com/kubernetes/client-go/blob/master/util/cert/cert.go#L95)

neolit123:
@kubernetes/sig-cluster-lifecycle
/sig api-machinery

neolit123:
/lifecycle frozen

> but we believe it can also occur naturally when a system clock drifts forwards and a reboot or other sync corrects the system clock backwards again.

did you confirm that this can happen naturally?

> which uses time.Now and NotBefore : now.UTC()

i think the logic there is pretty sane and i doubt changes will be accepted to it, by the owners of client-go.

if you want to get feedback on this the forums for that are the SIG API machinery meeting or slack channel (#sig-api-machinery).

mikeparker:
> did you confirm that this can happen naturally?

Yes, users are getting the bug simply with day to day usage, they aren't manually changing their computer clock.

> i think the logic there is pretty sane and i doubt changes will be accepted to it, by the owners of client-go.

[...]

PR #94504, opened 2020-09-04
kubeadm: print warnings on invalid cert period instead of erroring out

neolit123:
Client side period validation of certificates should not be
fatal, as local clock skews are not so uncommon. The validation
should be left to the running servers.

- Remove this validation from TryLoadCertFromDisk().
- Add a new function ValidateCertPeriod(), that can be used for this
purpose on demand.
- In phases/certs add a new function CheckCertificatePeriodValidity()
that will print warnings if a certificate does not pass period
validation, and caches certificates that were already checked.
- Use the function in a number of places where certificates
are loaded from disk.

xref https://github.com/kubernetes/kubernetes/issues/76714

```release-note
kubeadm: do not throw errors if the current system time is outside of the NotBefore and NotAfter bounds of a loaded certificate. Print warnings instead.
```

```docs

```

### DIFF
--- a/cmd/kubeadm/app/phases/certs/certs.go
+++ b/cmd/kubeadm/app/phases/certs/certs.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sync"
 
 	"github.com/pkg/errors"
 	"k8s.io/client-go/util/keyutil"
@@ -30,6 +31,12 @@ import (
 	pkiutil "k8s.io/kubernetes/cmd/kubeadm/app/util/pkiutil"
 
 	kubeadmconstants "k8s.io/kubernetes/cmd/kubeadm/app/constants"
+)
+
+var (
+	// certPeriodValidation is used to store if period validation was done for a certificate
+	certPeriodValidationMutex sync.Mutex
+	certPeriodValidation      = map[string]struct{}{}
 )
 
 // CreatePKIAssets will create and write to disk all PKI assets necessary to establish the control plane.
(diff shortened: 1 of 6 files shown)


cmd/kubeadm/app/phases/kubeconfig/kubeconfig.go
```diff
@@ -124,6 +125,9 @@ func getKubeConfigSpecs(cfg *kubeadmapi.InitConfiguration) (map[string]*kubeConf
 	if err != nil {
 		return nil, errors.Wrap(err, "couldn't create a kubeconfig; the CA files couldn't be loaded")
 	}
+	// Validate period
```
yagonobre, line 128: Should we move this comment to line 129 and keep the line 128 empty?

yagonobre (commented):
Thanks!
Added some nits and doubts, but LGTM
/lgtm

cmd/kubeadm/app/phases/certs/certs.go
```diff
@@ -438,3 +457,21 @@ func validateCertificateWithConfig(cert *x509.Certificate, baseName string, cfg
 	}
 	return nil
 }
+
+// CheckCertificatePeriodValidity takes a certificate and prints a warning if its period
+// is not valid related to the current time. It does so only if the certificate was not validated already
+// by keeping track with a cache.
+func CheckCertificatePeriodValidity(baseName string, cert *x509.Certificate) {
```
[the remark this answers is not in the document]
neolit123: i'm still debating whether the cache is wanted here. the only benefit is that during a "kubeadm" process execution, it can omit printing the same validation warning multiple times for the same certificate.

cmd/kubeadm/app/phases/kubeconfig/kubeconfig.go
neolit123: i think it's consistent with the other function `WriteKubeConfigWithClientCert` where `// Validate period` is printed right after a line with `}`

neolit123:
let's unblock this because original problem has existed for a while.
https://github.com/kubernetes/kubernetes/issues/76714

PR #94508, opened 2020-09-04
kubelet: remove alpha warnings for CNI flags

andrewsykim:

**What type of PR is this?**
/kind cleanup

CNI should not be considered "alpha" by kubelet as it is widely used and adopted. We should remove the alpha warnings for kubelet flags that were originally added from the early days of CNI. This came up in today's SIG Network meeting.

Fixes #

```release-note
kubelet: remove alpha warnings for CNI flags.
```

```docs

```

### DIFF
--- a/pkg/kubelet/config/flags.go
+++ b/pkg/kubelet/config/flags.go
@@ -94,9 +94,9 @@ func (s *ContainerRuntimeOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.DurationVar(&s.ImagePullProgressDeadline.Duration, "image-pull-progress-deadline", s.ImagePullProgressDeadline.Duration, fmt.Sprintf("If no pulling progress is made before this deadline, the image pulling will be cancelled. %s", dockerOnlyWarning))
 
 	// Network plugin settings for Docker.
-	fs.StringVar(&s.NetworkPluginName, "network-plugin", s.NetworkPluginName, fmt.Sprintf("<Warning: Alpha feature> The name of the network plugin to be invoked for various events in kubelet/pod lifecycle. %s", dockerOnlyWarning))
-	fs.StringVar(&s.CNIConfDir, "cni-conf-dir", s.CNIConfDir, fmt.Sprintf("<Warning: Alpha feature> The full path of the directory in which to search for CNI config files. %s", dockerOnlyWarning))
-	fs.StringVar(&s.CNIBinDir, "cni-bin-dir", s.CNIBinDir, fmt.Sprintf("<Warning: Alpha feature> A comma-separated list of full paths of directories in which to search for CNI plugin binaries. %s", dockerOnlyWarning))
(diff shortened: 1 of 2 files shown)


pkg/kubelet/config/flags.go
```diff
@@ -95,8 +95,8 @@ func (s *ContainerRuntimeOptions) AddFlags(fs *pflag.FlagSet) {
 
 	// Network plugin settings for Docker.
 	fs.StringVar(&s.NetworkPluginName, "network-plugin", s.NetworkPluginName, fmt.Sprintf("<Warning: Alpha feature> The name of the network plugin to be invoked for various events in kubelet/pod lifecycle. %s", dockerOnlyWarning))
```
SergeyKanzhelev, line 97: curious if this is still alpha? this flag is here for a long time, but docs saying 1.19: https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/

andrewsykim: Good catch, going to remove the alpha warning here too.

I've always wondered why the CNI flags are still marked "alpha". My assumption is that we added them initially and we just forgot to remove them.

andrewsykim: We should also update the docs if we remove it here

danwinship:
/sig network

> CNI has not reached 1.0](https://github.com/containernetworking/cni/tags) yet.

Most of those tags indicate versions of the golang implementation, not of the specification itself. The spec is currently version 0.4.0. However it is not clear what the exact semantics of that leading "0" are. While the spec may not be _final_, it is definitely _stable_.

SergeyKanzhelev:
> > CNI has [not reached 1.0](https://github.com/containernetworking/cni/tags) yet.
> 
> Most of those tags indicate versions of the golang implementation, not of the specification itself. The spec is currently version 0.4.0. However it is not clear what the exact semantics of that leading "0" are. While the spec may not be _final_, it is definitely _stable_.

Quick googling [shown that back in 2017](https://www.cncf.io/wp-content/uploads/2020/08/Introduction-to-CNI-2.pdf) the desire was to call stable version 1.0:

image

Just for my education, is it officially stable now? I am confused by `0.` version like others might be.

danwinship:
I mean, I suppose it's possible that the CNI maintainers will suddenly go insane and declare that they only support Windows on ARM in CNI 0.5.0 and later. And in that case, Kubernetes will just clarify that it only supports CNI 0.2.0, 0.3.0, and 0.4.0.

The use of Kubernetes with plugins that support CNI 0.2.0 - 0.4.0 is completely stable and supported, and no future changes to CNI can change that.

thockin (commented):
Thanks!

/lgtm
/approve

andrewsykim (commented):
/milestone v1.20

sftim:
What would it mean for CNI to reach general availability; is that different from reaching 1.0?

I suspect this is a similar situation to Kubernetes' Ingress resource: it's been in alpha so long that the existing, formally not yet stabilized, API is the one we all need to agree on and use.
It feels odd to document the API as stable but not also bump it to 1.0 though, which is why I'm asking.

PR #94509, opened 2020-09-04
Release lock on cancel for KCM and CCM

tnozicka:
**What type of PR is this?**
/kind bug

When KCM and CCM are exiting they should release the lock they are holding so another instance can takeover faster and doesn't have to wait for the lock to expire.

```release-note
NONE
```

### DIFF
--- a/cmd/kube-controller-manager/app/controllermanager.go
+++ b/cmd/kube-controller-manager/app/controllermanager.go
@@ -293,8 +293,9 @@ func Run(c *config.CompletedConfig, stopCh <-chan struct{}) error {
 				klog.Fatalf("leaderelection lost")
 			},
 		},
-		WatchDog: electionChecker,
-		Name:     "kube-controller-manager",
+		WatchDog:        electionChecker,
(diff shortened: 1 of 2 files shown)


soltysh (approved):
/lgtm

cmd/kube-controller-manager/app/controllermanager.go
```diff
@@ -293,8 +293,9 @@ func Run(c *config.CompletedConfig, stopCh <-chan struct{}) error {
 				klog.Fatalf("leaderelection lost")
 			},
 		},
-		WatchDog: electionChecker,
-		Name:     "kube-controller-manager",
+		WatchDog:        electionChecker,
+		Name:            "kube-controller-manager",
+		ReleaseOnCancel: true,
```
deads2k, line 298: this doesn't work.  Fix the contexts and rewire the controllers

deads2k:
/hold
/lgtm cancel

To do this properly will require first plumbing the correct contexts.  Let's do that.

PR #94514, opened 2020-09-04
remove feature gate SupportIPVSProxyMode.

### LINKED ISSUE #94497: Feature gate SupportIPVSProxyMode is GA - when does it get removed?
When should it be removed?

@rikatz 

/sig network
/kind cleanup

tangwz:
**What type of PR is this?**
/kind cleanup
/sig network

According to #94497, feature gate SupportIPVSProxyMode need to be removed since 1.11.

Fixes #94497

@rikatz 

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -194,12 +194,6 @@ const (
 	// Postpone deletion of a PV or a PVC when they are being used
 	StorageObjectInUseProtection featuregate.Feature = "StorageObjectInUseProtection"
 
-	// owner: @m1093782566
-	// GA: v1.11
-	//
-	// Implement IPVS-based in-cluster service load balancing
-	SupportIPVSProxyMode featuregate.Feature = "SupportIPVSProxyMode"
-
 	// owner: @dims, @derekwaynecarr
 	// alpha: v1.10
 	// beta: v1.14
@@ -682,7 +676,6 @@ var defaultKubernetesFeatureGates = map[featuregate.Feature]featuregate.FeatureS
 	CSINodeInfo:                    {Default: true, PreRelease: featuregate.GA, LockToDefault: true}, // remove in 1.19
 	BlockVolume:                    {Default: true, PreRelease: featuregate.GA, LockToDefault: true}, // remove in 1.20
 	StorageObjectInUseProtection:   {Default: true, PreRelease: featuregate.GA},
-	SupportIPVSProxyMode:           {Default: true, PreRelease: featuregate.GA},
 	SupportPodPidsLimit:            {Default: true, PreRelease: featuregate.GA, LockToDefault: true}, // remove in 1.21
 	SupportNodePidsLimit:           {Default: true, PreRelease: featuregate.GA, LockToDefault: true}, // remove in 1.21
 	HyperVContainer:                {Default: false, PreRelease: featuregate.Alpha},
--- a/pkg/proxy/ipvs/README.md
+++ b/pkg/proxy/ipvs/README.md
@@ -334,28 +334,19 @@ export KUBE_PROXY_MODE=ipvs
 
 ### Cluster Created by Kubeadm
 
-If you are using kubeadm with a [configuration file](https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-init/#config-file), you have to add `mode: ipvs` and also add `SupportIPVSProxyMode: true` below the `kubeProxy` field as part of the kubeadm configuration.
+If you are using kubeadm with a [configuration file](https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-init/#config-file), you have to add `mode: ipvs` below the `kubeProxy` field as part of the kubeadm configuration.
 
 ```yaml
 ...
 kubeProxy:
   config:
-    featureGates:
-      SupportIPVSProxyMode: true
     mode: ipvs
 ...
 ```
-Note that in Kubernetes 1.11 and later, `SupportIPVSProxyMode` is set to `true` by default.
 
 before running
 
 `kube init --config <path_to_configuration_file>`
-
-If you are using Kubernetes v1.8, you can also add the flag `--feature-gates=SupportIPVSProxyMode=true` (deprecated since v1.9) in `kubeadm init` command
-
-```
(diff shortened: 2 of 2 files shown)


rikatz:
/priority backlog

@tangwz left a review about the README inside the proxy package, and also there's a need to adjust the PR from the docs (opened in kubernetes/website).

Thank you!

pkg/proxy/ipvs/README.md
```diff
@@ -334,24 +334,9 @@ export KUBE_PROXY_MODE=ipvs
...
-...
-```
 Note that in Kubernetes 1.11 and later, `SupportIPVSProxyMode` is set to `true` by default.
 
-before running
-
-`kube init --config <path_to_configuration_file>`
-
-If you are using Kubernetes v1.8, you can also add the flag `--feature-gates=SupportIPVSProxyMode=true` (deprecated since v1.9) in `kubeadm init` command
+If you are using Kubernetes v1.8, you can add the flag `--feature-gates=SupportIPVSProxyMode=true` (deprecated since v1.9) in `kubeadm init` command
```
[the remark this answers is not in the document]
tangwz: PTAL.

tangwz:
@rikatz PTAL

pkg/proxy/ipvs/README.md
```diff
@@ -334,31 +334,8 @@ export KUBE_PROXY_MODE=ipvs
 
 ### Cluster Created by Kubeadm
 
-If you are using kubeadm with a [configuration file](https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-init/#config-file), you have to add `mode: ipvs` and also add `SupportIPVSProxyMode: true` below the `kubeProxy` field as part of the kubeadm configuration.
```
rikatz, line 337: This part you can remove the reference to the feature flag, but keep the 'mode: ipvs' as this is still true

pkg/proxy/ipvs/README.md
```diff
@@ -334,31 +334,8 @@ export KUBE_PROXY_MODE=ipvs
...
-```yaml
-...
-kubeProxy:
-  config:
-    featureGates:
-      SupportIPVSProxyMode: true
-    mode: ipvs
-...
-```
 Note that in Kubernetes 1.11 and later, `SupportIPVSProxyMode` is set to `true` by default.
```
rikatz, line 337: This can be removed, as the SupportIPVSProxyMode is being completely removed from the code

rikatz (changes requested):
Left some last changes in README.md file. 

Thanks

tangwz:
@rikatz PTAL.

Thanks.

rikatz:
@tangwz overall looks good to me, thank you!

Please squash your commits into one, and I'll ask for the sig approval of the PR :)

Thank you!

tangwz:
@rikatz I have rebase my commits into one. PTAL.

Thanks.

rikatz:
/lgtm

Thank you!

rikatz:
/assign @thockin 
/cc @thockin 

Tim, this Feature Gate can be removed :)

andrewsykim (commented):
I think there are still references to the feature gate under the "Why kube-proxy can't start IPVS mode" section.

tangwz:
Thanks Andrew.

@rikatz PTAL.

rikatz:
/lgtm

@andrewsykim I think now it's fine :)

pkg/features/kube_features.go
```diff
@@ -680,7 +674,6 @@ var defaultKubernetesFeatureGates = map[featuregate.Feature]featuregate.FeatureS
 	CSINodeInfo:                    {Default: true, PreRelease: featuregate.GA, LockToDefault: true}, // remove in 1.19
 	BlockVolume:                    {Default: true, PreRelease: featuregate.GA, LockToDefault: true}, // remove in 1.20
 	StorageObjectInUseProtection:   {Default: true, PreRelease: featuregate.GA},
-	SupportIPVSProxyMode:           {Default: true, PreRelease: featuregate.GA},
```
andrewsykim, line 683: Based on https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecation I think we need to set `LockToDefault` for 2 releases before we can remove it.

thockin: Deprecation talks about announcing the deprecation (which we did in CHANGELOG/CHANGELOG-1.10.md) and about a warning when used (which happens by nature of being marked GA).

Deprecation also says "must be documented in both in the release notes and the corresponding CLI help" but CLI help filters out GA and deprecated gates.

Given the error is in the gates lib, I think we should be OK to delete this, and then we can fix the doc and the lib.

Disaagree?

rikatz: I think in this case I agree with that, as there are no further references in other parts of the code about the Feature gate.

But in case of https://github.com/kubernetes/kubernetes/pull/94581 I actually don't know if we should remove (as the Feature Gate went to GA in v1.18) or only mark it as LockToDefault: true, as that PR also removes some logic from the code. This also wasn't announced in the CHANGELOG/CHANGELOG-1.18.md when it turned into GA, should it be?

andrewsykim: > a warning when used (which happens by nature of being marked GA).

TIL :) 

In rule #7 of the Deprecation Policy:

> GA: the feature gate is deprecated (see "Deprecation") and becomes non-operational.

Does "non-operational" mean we don't check the gate anywhere or does it also imply the gate is locked? If the former, then sgtm to remove it now.

thockin:
I interpret Non-Operational as "has no effect".

We need to formalize the lifecycle better, but I can't see how removing
this gate would be worse than locking it.  We announced the deprecation and
anyone who is setting the gate is already getting a warning (and no effect).

andrewsykim:
Makes sense to me, thanks for clarifying Tim :)

andrewsykim (commented):
/approve
/lgtm

Thanks @tangwz!

andrewsykim (commented):
/milestone v1.20

andrewsykim:
@tangwz looks like there's a merge conflict, can you rebase please?

tangwz:
@andrewsykim PTAL.

Thanks.

rikatz:
/lgtm

Thanks, let's see if this finally merges :)

PR #94526, opened 2020-09-04
Add vCenter info metric

### LINKED ISSUE #94651: Add vSphere metrics to report vCenter version

We need reporting of `vSphere` `vCenter` version to be shown in the metrics.

**Why is this needed**:

The goal behind this is to add some info about the underlying infrastructure for the kubernetes cluster. Some customers were experiencing issues during the cluster upgrades, but nothing in the upgrades themselves or kubernetes code for `vSphere` provider seemingly could cause those. The issue is storage related, and happening on a variety of releases with a significant delay from any other underlying infrastructure change, to which the kubernetes cluster has no relation and a way to know about. So knowing used `vCenter` hostnames, and `vSphere` versions associated with it over time would be extremely beneficial for debugging in similar circumstances.

Some of the use cases we face:

1) A `vSpehre` `DRS` is enabled in a customer environment, and so there occurs a periodic automated vMotion for VMs backing kuberentes `Nodes`. We'd like to know on which host the actual kubernetes `Node` was located at any moment, and if any vCenter version transitions happened over time, new hosts added/removed/renamed.

Danil-Grigorev:
Metric provides information about used vCenter version
for every connected vCenter host. Metric displays the info about `vCenter` version, build and hostname

image

**What type of PR is this?**
/kind feature

Currently, the vCenter version is unknown to the Kubernetes cluster. Adding this metric will report the existing vCenter version for every connected host defined in `vsphere.conf` file.

Fixes https://github.com/kubernetes/kubernetes/issues/94651

```release-note
Add a new `vSphere` metric: `cloudprovider_vsphere_vcenter_versions`. It's content show `vCenter` hostnames with the associated server version.
```

```docs
NONE
```

### DIFF
--- a/staging/src/k8s.io/legacy-cloud-providers/vsphere/vclib/connection.go
+++ b/staging/src/k8s.io/legacy-cloud-providers/vsphere/vclib/connection.go
@@ -65,6 +65,7 @@ func (connection *VSphereConnection) Connect(ctx context.Context) error {
 			klog.Errorf("Failed to create govmomi client. err: %+v", err)
 			return err
 		}
+		setVCenterInfoMetric(connection)
 		return nil
 	}
 	m := session.NewManager(connection.Client)
@@ -83,6 +84,7 @@ func (connection *VSphereConnection) Connect(ctx context.Context) error {
 		klog.Errorf("Failed to create govmomi client. err: %+v", err)
 		return err
 	}
+	setVCenterInfoMetric(connection)
 	return nil
 }
 
--- a/staging/src/k8s.io/legacy-cloud-providers/vsphere/vclib/vsphere_metrics.go
+++ b/staging/src/k8s.io/legacy-cloud-providers/vsphere/vclib/vsphere_metrics.go
@@ -17,8 +17,10 @@ limitations under the License.
 package vclib
 
 import (
+	"sync"
 	"time"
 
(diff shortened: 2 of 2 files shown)


brancz:
This is a bit of an odd metric to be in Kubernetes, I would expect this to be in a vsphere exporter running on each node instead.

Danil-Grigorev:
@brancz This is the simplest way to track, if the `Nodes` were vMotioned to a different vCenter, in a multi-center cluster setup. This is poorly supported on the kuberentes side in the in-tree implementation, while the out-of-tree does not suffer from that problem. The feature work on in-tree is not encouraged, as the code is marked deprecated. The PR aims to keep minimal information about automatic transitions between vCenters, and if the issue occurs, which version of the cluster was affected, and does not add a requirement to run custom tooling. This also allowes to reuse of existing infrastructure around the default metrics, in order to collect this information.

brancz:
Fair enough. Looks good from instrumentation side.

/lgtm

andrewsykim:
In general we are trying to move away form adding more features/enhancements to the in-tree vSphere provider since this will be removed soon. 

I would really like to see this feature added to the external provider instead https://github.com/kubernetes/cloud-provider-vsphere. Happy to help review/approve a PR there as well.

divyenpatel (approved):
/approve

Issue #94528, opened 2020-09-04
[Failing Test] bazel-test-master (periodic-kubernetes-bazel-test-master)


hasheddan:
**Which jobs are failing**:

`bazel-test-master (periodic-kubernetes-bazel-test-master)`

**Which test(s) are failing**:

```
//cmd/kube-scheduler/app/go_default_test:run_1_of_3
//cmd/kube-scheduler/app/go_default_test:run_3_of_3
//pkg/util/iptables/go_default_test:run_2_of_3
//cmd/kube-scheduler/app/go_default_test:run_2_of_3
...
```

**Since when has it been failing**:

```
09-04-20 01:03 CDT
```

**Testgrid link**:

https://testgrid.k8s.io/sig-release-master-blocking#bazel-test-master

**Reason for failure**:

Errors differ between each of the failing tests. Looks like failure started between `26d0f8460` and `fe8aa6eba`.

**Anything else we need to know**:

/cc @kubernetes/ci-signal 
/priority critical-urgent
/milestone v1.20

liggitt:
Almost certainly https://github.com/kubernetes/kubernetes/pull/93605, which means we have remaining unit test flakes

liggitt:
Looks like three primary packages are flaking:
- cmd/kube-scheduler/app
- pkg/probe/http
- k8s.io/apiserver/pkg/server/filters

knight42:
I would like to help. I am investigating `TestClientReceivedGOAWAY` and `TestGOAWAYConcurrency` now.

liggitt:
thanks, I've got TestHTTPProbeProxy

knight42:
@liggitt `TestGOAWAYConcurrency` seems to be a bit different. 

> TestGOAWAYConcurrency tests GOAWAY frame will not affect concurrency requests in a single http client instance

This test is meant to use a single http client with http/2 transport, but as per the log:
```
==================
WARNING: DATA RACE
Write at 0x00c0007ad650 by goroutine 166:
  net/http.(*readTrackingBody).Read()
      GOROOT/src/net/http/transport.go:620 +0x44
  k8s.io/kubernetes/vendor/golang.org/x/net/http2.(*clientStream).writeRequestBody()
      vendor/golang.org/x/net/http2/transport.go:1308 +0x761
  k8s.io/kubernetes/vendor/golang.org/x/net/http2.(*Transport).getBodyWriterState.func1()
[...]

PR #94534, opened 2020-09-04
test(iptables): deflake TestRestoreAllGrabOldLock

knight42:

**What type of PR is this?**
/kind flake

From the log:
```
=== RUN   TestRestoreAllGrabOldLock
    iptables_test.go:1115: expected to lock @xtables, got listen unix @xtables: bind: address already in use
--- FAIL: TestRestoreAllGrabOldLock (0.00s)
```
I think the cause is now we run the unit test 3 times and the unix socket is not deleted in time.

Part of #94528

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/util/iptables/iptables_test.go
+++ b/pkg/util/iptables/iptables_test.go
@@ -26,9 +26,11 @@ import (
 	"reflect"
 	"strings"
 	"testing"
+	"time"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 	utilversion "k8s.io/apimachinery/pkg/util/version"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/utils/exec"
 	fakeexec "k8s.io/utils/exec/testing"
 )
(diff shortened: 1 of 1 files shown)


hasheddan (commented):
/priority critical-urgent

pkg/util/iptables/iptables_test.go
```diff
@@ -1109,8 +1109,15 @@ func TestRestoreAllGrabOldLock(t *testing.T) {
 	runner := newInternal(&fexec, ProtocolIPv4, TestLockfilePath)
 	defer os.Remove(TestLockfilePath)
 
+	socketName := "@xtables"
+	_, err := os.Stat(socketName)
+	if err == nil { // socket exists, need to delete it
```
liggitt, line 1114: I'm not sure about this... what does this do if multiple instances of the test are running in parallel?

pkg/util/iptables/iptables_test.go
```diff
@@ -1109,8 +1109,15 @@ func TestRestoreAllGrabOldLock(t *testing.T) {
...
+		if err := os.Remove(socketName); err != nil && !os.IsNotExist(err) {
+			t.Fatalf("Failed to remove socket `%s`: %v", socketName, err)
+		}
+	}
 	// Grab the abstract @xtables socket
-	runLock, err := net.ListenUnix("unix", &net.UnixAddr{Name: "@xtables", Net: "unix"})
```
liggitt, line 1113: if we have to lock a global shared resource in this unit test, I think we should tolerate/retry on "already in use" errors some number of times

pkg/util/iptables/iptables_test.go
knight42: I thought the test were running serially.. Did you know if we could specify a different socket name each time?

liggitt: > I thought the test were running serially..

I think they each run in their own process, but could run in parallel

> Did you know if we could specify a different socket name each time?

It looks hard-coded in the iptables.go file, so I don't think so

pkg/util/iptables/iptables_test.go
liggitt: the point of the test is to lock @xtables, then make sure iptables fails because it is locked. We can poll/retry acquiring the xtables lock here until wait.ForeverTestTimeout

liggitt: looks good, did you try building the test and running in parallel with `stress`?

knight42: sadly this test could only be run on linux and I am using macOS, so I haven't test it locally..

liggitt: ran this several times on a linux box and verified it passed 

```
bazelisk test --runs_per_test=10 --config=unit //pkg/util/iptables/...
//pkg/util/iptables:go_default_test                                      PASSED in 9.3s
  Stats over 3 runs: max = 9.3s, min = 5.3s, avg = 7.3s, dev = 1.6s
```

without this fix, the package failed almost every time

PR #94535, opened 2020-09-04
Speed up slow Kubeadm unit tests, avoid mutating global variables

liggitt:
**What type of PR is this?**

/kind bug
/kind flake

1. Ensures all kubeadm unit tests have unique names
2. Runs slow upgrade tests in parallel (xref https://github.com/kubernetes/kubernetes/issues/94528#issuecomment-687208028 and https://testgrid.k8s.io/sig-release-master-blocking#bazel-test-master&include-filter-by-regex=%2Fupgrade&width=20)

    ```
    TestStaticPodControlPlane takes 73 seconds
    	parallelizing -> 22 seconds
    TestRenewCertsByComponent takes 18 seconds
    	parallelize -> 5 seconds
    ```

3. Avoids mutating global state in package variables (discovered when running tests in parallel)

    [KubeadmCert#GetConfig](https://github.com/kubernetes/kubernetes/blob/release-1.19/cmd/kubeadm/app/phases/certs/certlist.go#L50-L60) mutates the receiver, and pointers to package KubeadmCert variables were being passed around and referenced. This PR changes those to generator functions so we can be sure all uses get a pristine copy of the object.

```release-note
NONE
```

/cc @neolit123

### DIFF
--- a/cmd/kubeadm/app/cmd/alpha/certs_test.go
+++ b/cmd/kubeadm/app/cmd/alpha/certs_test.go
@@ -98,9 +98,9 @@ func TestRunRenewCommands(t *testing.T) {
 	CACerts := map[string]*x509.Certificate{}
 	CAKeys := map[string]crypto.Signer{}
 	for _, ca := range []*certsphase.KubeadmCert{
-		&certsphase.KubeadmCertRootCA,
-		&certsphase.KubeadmCertFrontProxyCA,
-		&certsphase.KubeadmCertEtcdCA,
+		certsphase.KubeadmCertRootCA(),
+		certsphase.KubeadmCertFrontProxyCA(),
+		certsphase.KubeadmCertEtcdCA(),
 	} {
 		caCert, caKey, err := ca.CreateAsCA(cfg)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


liggitt:
flake on `k8s.io/legacy-cloud-providers/azure/cache` (xref https://github.com/kubernetes/kubernetes/issues/94528#issuecomment-687208028)

liggitt:
unrelated e2e flake
/retest

neolit123:
> KubeadmCert#GetConfig mutates the receiver, and pointers to package KubeadmCert variables were being passed around and referenced. This PR changes those to generator functions so we can be sure all uses get a pristine copy of the object.

this seems to have passed review in 1.12.
the changes are fine, thanks for sending this!

/lgtm
/hold cancel

liggitt:
bazel debian fetch error in https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/94535/pull-kubernetes-e2e-gce-100-performance/1301993935178043392/
/retest

liggitt:
`Services should have session affinity timeout work for NodePort service` flake in e2e

/retest

liggitt:
Unrelated work queue metrics flake, will open a separate issue for that
``` 
=== RUN   TestMetrics
    metrics_test.go:263: expected 0.001, got 0
--- FAIL: TestMetrics (0.00s)
```

PR #94539, opened 2020-09-04
Use unique socket name per cm test

liggitt:
**What type of PR is this?**

/kind failing-test
/kind flake

Test-only change to help parallel runs of cm/devicemanager tests not conflict. xref https://github.com/kubernetes/kubernetes/issues/94528#issuecomment-687208028 and https://testgrid.k8s.io/sig-release-master-blocking#bazel-test-master&width=20&sort-by-name=&include-filter-by-regex=pkg/kubelet/cm/devicemanager

```release-note
NONE
```


hasheddan (approved):
/lgtm

PR #94543, opened 2020-09-04
Deflake port forward tests to avoid hard-coded local ports

liggitt:
**What type of PR is this?**

/kind flake

Avoids assuming specific local ports are available

xref https://github.com/kubernetes/kubernetes/issues/94528#issuecomment-687208028 and https://testgrid.k8s.io/sig-release-master-blocking#bazel-test-master&width=20&sort-by-name=&include-filter-by-regex=pkg/client/tests

```release-note
NONE
```

### DIFF
--- a/pkg/client/tests/portfoward_test.go
+++ b/pkg/client/tests/portfoward_test.go
@@ -113,10 +113,10 @@ func TestForwardPorts(t *testing.T) {
 		serverSends map[int32]string
 	}{
 		"forward 1 port with no data either direction": {
-			ports: []string{"5000"},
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
/lgtm

liggitt:
unrelated `Kubectl client Simple pod should support inline execution and attach` flake (https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/94543/pull-kubernetes-e2e-gce-ubuntu-containerd/1301967133663563776/)
/retest

PR #94544, opened 2020-09-04
Deflake serving options, avoid hard-coding ports

liggitt:
**What type of PR is this?**

/kind flake

Avoids hard-coding ports in the serving tests, to avoid flakes if the ports are in use or the test is run multiple times in parallel

xref https://github.com/kubernetes/kubernetes/issues/94528#issuecomment-687208028, https://testgrid.k8s.io/sig-release-master-blocking#bazel-test-master&width=20&sort-by-name=&include-filter-by-regex=k8s.io/apiserver/pkg/server/options/go

```release-note
NONE
```

/assign @sttts

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/server/options/serving_unix_test.go
+++ b/staging/src/k8s.io/apiserver/pkg/server/options/serving_unix_test.go
@@ -19,31 +19,37 @@ limitations under the License.
 package options
 
 import (
+	"fmt"
 	"net"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
/lgtm

liggitt:
`HostPath should support r/w [NodeConformance]` flake (https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/94544/pull-kubernetes-node-e2e/1301972922163793920/)
/retest

Issue #94547, opened 2020-09-04
[Flaky unit test] TestDevicePluginReRegistrationProbeMode: Devices of previous registered should be removed


liggitt:
**Which jobs are flaking**:

pull-kubernetes-bazel-test

**Which test(s) are flaking**:

TestDevicePluginReRegistrationProbeMode

**Testgrid link**:

https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-bazel-test&include-filter-by-regex=pkg%2Fkubelet%2Fcm%2Fdevicemanager

**Reason for failure**:

```
=== RUN   TestDevicePluginReRegistrationProbeMode
I0904 23:07:54.382417 2813190 fake_topology_manager.go:29] [fake topologymanager] NewFakeManager
W0904 23:07:54.382652 2813190 manager.go:596] Failed to retrieve checkpoint for "kubelet_internal_checkpoint": checkpoint is not found
I0904 23:07:54.386357 2813190 plugin_manager.go:114] Starting Kubelet Plugin Manager
I0904 23:07:54.390968 2813190 device_plugin_stub.go:131] Starting to serve on /tmp/device_plugin394710885/device-plugin.sock
I0904 23:07:54.393796 2813190 device_plugin_stub.go:153] GetInfo
E0904 23:07:54.394527 2813190 goroutinemap.go:150] Operation for "/tmp/device_plugin394710885/server.sock" failed. No retries permitted until 2020-09-04 23:07:54.894365704 +0000 UTC m=+0.587686396 (durationBeforeRetry 500ms). Error: "RegisterPlugin error -- failed to get
 plugin info using RPC GetInfo at socket /tmp/device_plugin394710885/server.sock, err: rpc error: code = Unimplemented desc = unknown service pluginregistration.Registration"
I0904 23:07:54.401946 2813190 device_plugin_stub.go:227] ListAndWatch
I0904 23:07:54.408602 2813190 device_plugin_stub.go:131] Starting to serve on /tmp/device_plugin394710885/device-plugin.sock.new
E0904 23:07:55.390620 2813190 endpoint.go:107] listAndWatch ended unexpectedly for device plugin fake-domain/resource with error rpc error: code = Canceled desc = grpc: the client connection is closing
I0904 23:07:55.394203 2813190 device_plugin_stub.go:153] GetInfo
[...]

Issue #93902, opened 2020-08-11
[Flaky unit test] Test_Run_OneVolumeAttachAndDetachMultipleNodesWithReadWriteMany flakes with "Wrong total GetAttachCallCount" error


liggitt:
**Which test(s) are flaking**:

Test_Run_OneVolumeAttachAndDetachMultipleNodesWithReadWriteMany

**Reason for failure**:

```
--- FAIL: Test_Run_OneVolumeAttachAndDetachMultipleNodesWithReadWriteMany (0.63s)
    reconciler_test.go:811: Warning: Wrong NewAttacherCallCount. Expected: <2> Actual: <0>. Will retry.
    reconciler_test.go:811: Warning: Wrong NewAttacherCallCount. Expected: <2> Actual: <1>. Will retry.
    reconciler_test.go:909: Warning: Wrong total GetAttachCallCount(). Expected: <2> Actual: <3>. Will retry.
    reconciler_test.go:909: Warning: Wrong total GetAttachCallCount(). Expected: <2> Actual: <3>. Will retry.
    reconciler_test.go:909: Warning: Wrong total GetAttachCallCount(). Expected: <2> Actual: <3>. Will retry.
    reconciler_test.go:909: Warning: Wrong total GetAttachCallCount(). Expected: <2> Actual: <3>. Will retry.
    reconciler_test.go:909: Warning: Wrong total GetAttachCallCount(). Expected: <2> Actual: <3>. Will retry.
    reconciler_test.go:909: Warning: Wrong total GetAttachCallCount(). Expected: <2> Actual: <3>. Will retry.
    reconciler_test.go:919: Total AttachCallCount does not match expected value. Expected: <2>
FAIL
```

This is flaking rarely enough that it is not caught by our CI jobs, which currently tolerate up to 2 unit test failures per run (!).

With that toleration removed in https://github.com/kubernetes/kubernetes/pull/93605, this flake has been seen (https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93605/pull-kubernetes-bazel-test/1293257915502694408).

Reproducible with the following steps:

1. Build a test binary for the affected package:

```
go test -race -c ./pkg/controller/volume/attachdetach/reconciler
```

2. Stress the affected test using the [stress](https://godoc.org/golang.org/x/tools/cmd/stress) tool:

```
stress ./reconciler.test -test.run Test_Run_OneVolumeAttachAndDetachMultipleNodesWithReadWriteMany
[...]

PR #94553, opened 2020-09-05
Deflake TestRestoreAllWaitOldIptablesRestore

knight42:

**What type of PR is this?**

/kind flake

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/periodic-kubernetes-bazel-test-master/1302138009876959232

```
=== RUN   TestRestoreAllWaitOldIptablesRestore
I0905 07:22:09.261781  188833 trace.go:205] Trace[1298498081]: "iptables restore" (05-Sep-2020 07:22:07.260) (total time: 2000ms):
Trace[1298498081]: [2.000885887s] [2.000885887s] END
    iptables_test.go:1028: expected success, got failed to acquire old iptables lock: timed out waiting for the condition
--- FAIL: TestRestoreAllWaitOldIptablesRestore (2.00s)
```

Part of #94528

Since this test could only be run on linux, I have not verified the fix locally.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/util/iptables/iptables_linux.go
+++ b/pkg/util/iptables/iptables_linux.go
@@ -49,7 +49,7 @@ func (l *locker) Close() error {
 	return utilerrors.NewAggregate(errList)
 }
 
-func grabIptablesLocks(lockfilePath string) (iptablesLocker, error) {
+func grabIptablesLocks(lockfilePath14x, lockfilePath16x string) (iptablesLocker, error) {
 	var err error
 	var success bool
 
@@ -66,9 +66,9 @@ func grabIptablesLocks(lockfilePath string) (iptablesLocker, error) {
 	// can't assume which lock method it'll use.
 
 	// Roughly duplicate iptables 1.6.x xtables_lock() function.
-	l.lock16, err = os.OpenFile(lockfilePath, os.O_CREATE, 0600)
+	l.lock16, err = os.OpenFile(lockfilePath16x, os.O_CREATE, 0600)
 	if err != nil {
-		return nil, fmt.Errorf("failed to open iptables lock %s: %v", lockfilePath, err)
+		return nil, fmt.Errorf("failed to open iptables lock %s: %v", lockfilePath16x, err)
 	}
 
 	if err := wait.PollImmediate(200*time.Millisecond, 2*time.Second, func() (bool, error) {
@@ -82,7 +82,7 @@ func grabIptablesLocks(lockfilePath string) (iptablesLocker, error) {
 
 	// Roughly duplicate iptables 1.4.x xtables_lock() function.
 	if err := wait.PollImmediate(200*time.Millisecond, 2*time.Second, func() (bool, error) {
-		l.lock14, err = net.ListenUnix("unix", &net.UnixAddr{Name: "@xtables", Net: "unix"})
+		l.lock14, err = net.ListenUnix("unix", &net.UnixAddr{Name: lockfilePath14x, Net: "unix"})
 		if err != nil {
 			return false, nil
 		}
--- a/pkg/util/iptables/iptables_test.go
+++ b/pkg/util/iptables/iptables_test.go
@@ -35,7 +35,11 @@ import (
 	fakeexec "k8s.io/utils/exec/testing"
 )
 
-const TestLockfilePath = "xtables.lock"
+func getLockPaths() (string, string) {
+	lock14x := fmt.Sprintf("@xtables-%d", time.Now().Nanosecond())
+	lock16x := fmt.Sprintf("xtables-%d.lock", time.Now().Nanosecond())
+	return lock14x, lock16x
+}
 
 func testIPTablesVersionCmds(t *testing.T, protocol Protocol) {
 	version := " v1.4.22"
@@ -934,8 +938,8 @@ func TestRestoreAll(t *testing.T) {
 			func(cmd string, args ...string) exec.Cmd { return fakeexec.InitFakeCmd(&fcmd, cmd, args...) },
 		},
 	}
-	runner := newInternal(&fexec, ProtocolIPv4, TestLockfilePath)
-	defer os.Remove(TestLockfilePath)
+	lockPath14x, lockPath16x := getLockPaths()
+	runner := newInternal(&fexec, ProtocolIPv4, lockPath14x, lockPath16x)
 
 	err := runner.RestoreAll([]byte{}, NoFlushTables, RestoreCounters)
 	if err != nil {
@@ -975,8 +979,8 @@ func TestRestoreAllWait(t *testing.T) {
 			func(cmd string, args ...string) exec.Cmd { return fakeexec.InitFakeCmd(&fcmd, cmd, args...) },
 		},
 	}
-	runner := newInternal(&fexec, ProtocolIPv4, TestLockfilePath)
-	defer os.Remove(TestLockfilePath)
+	lockPath14x, lockPath16x := getLockPaths()
+	runner := newInternal(&fexec, ProtocolIPv4, lockPath14x, lockPath16x)
 
 	err := runner.RestoreAll([]byte{}, NoFlushTables, RestoreCounters)
 	if err != nil {
@@ -1020,8 +1024,8 @@ func TestRestoreAllWaitOldIptablesRestore(t *testing.T) {
 			func(cmd string, args ...string) exec.Cmd { return fakeexec.InitFakeCmd(&fcmd, cmd, args...) },
 		},
 	}
-	runner := newInternal(&fexec, ProtocolIPv4, TestLockfilePath)
-	defer os.Remove(TestLockfilePath)
+	lockPath14x, lockPath16x := getLockPaths()
+	runner := newInternal(&fexec, ProtocolIPv4, lockPath14x, lockPath16x)
 
 	err := runner.RestoreAll([]byte{}, NoFlushTables, RestoreCounters)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


pkg/util/iptables/iptables_linux.go
```diff
@@ -81,7 +81,7 @@ func grabIptablesLocks(lockfilePath string) (iptablesLocker, error) {
 	}
 
 	// Roughly duplicate iptables 1.4.x xtables_lock() function.
-	if err := wait.PollImmediate(200*time.Millisecond, 2*time.Second, func() (bool, error) {
```
liggitt, line 84: I don't think extending this for testing purposes is correct. Doing something to avoid concurrent test runs might be preferable

knight42: do we have a better solution? listening on the same unix socket in the test is already able to prevent concurrent test running.

liggitt: perhaps something test-specific, like locking on a test lock file in iptables#TestMain?

danwinship: The locking tests already override the iptables-1.6-compatible lock path. They should probably override the iptables-1.4-compatible lock path as well. (In the iptables 1.4 case the "path" is actually an abstract socket name (`"@xtables"`), but it can still just be overridden with an arbitrary alternate name.)

pkg/util/iptables/iptables_test.go
```diff
@@ -934,7 +938,7 @@ func TestRestoreAll(t *testing.T) {
 			func(cmd string, args ...string) exec.Cmd { return fakeexec.InitFakeCmd(&fcmd, cmd, args...) },
 		},
 	}
-	runner := newInternal(&fexec, ProtocolIPv4, TestLockfilePath)
+	runner := newInternal(&fexec, ProtocolIPv4, getLockPath14x(), TestLockfilePath)
```
danwinship, line 941: it seems random to have a hardcoded testing lock name for 1.6 but a per-process generated one for 1.4...

danwinship (commented):
and does this resolve the flaking?

pkg/util/iptables/iptables_test.go
knight42: According to the report of previous tests https://testgrid.k8s.io/sig-release-master-blocking#bazel-test-master&width=20&sort-by-name=&include-filter-by-regex=pkg/util/iptables, looks like only acquring lock for 1.4 would time out, I am a bit curious as well.

liggitt: > it seems random to have a hardcoded testing lock name for 1.6 but a per-process generated one for 1.4...

I agree

> looks like only acquring lock for 1.4 would time out, I am a bit curious as well.

TestLockfilePath is not fully-qualified, perhaps the working dir is different per test-run? I agree that we should be consistent in how we assign random filenames for the test

knight42: > TestLockfilePath is not fully-qualified

The original 1.4 lock file `@xtables` is neither fully-qualified.

> I agree that we should be consistent in how we assign random filenames for the test

I am okay to update the tests.

liggitt: > The original 1.4 lock file `@xtables` is neither fully-qualified.

abstract sockets are handled differently than lock files and are not treated as pathnames

knight42: > abstract sockets are handled differently than lock files and are not treated as pathnames

just out of curiosity, I am sorry if I ask a dumb question. Perhaps I am missing something, currently `net.ListenUnix("unix", &net.UnixAddr{Name: "@xtables", Net: "unix"})` is called to grab the lock file https://github.com/kubernetes/kubernetes/blob/0e58014d71ba92a055e41541b3b7b7373cc0343e/pkg/util/iptables/iptables_linux.go#L84-L89
If there is no error, an unix domain socket named `@xtables` will be created under the curret working dir. I am confused why `@xtables` is not treated as pathname?

knight42:
`Test_Run_OneVolumeAttachAndDetachMultipleNodesWithReadWriteMany` is known to be flaky #93902 

pkg/util/iptables/iptables_test.go
liggitt: > If there is no error, an unix domain socket named `@xtables` will be created under the curret working dir.

Only on OSX. On linux, `@xtables` is an abstract socket in a global namespace unrelated to filesystem paths.

knight42: Ahhh got it! Thanks for your explanation!

knight42:
`TestDevicePluginReRegistrationProbeMode` is known to be flaky #94547

knight42:
@danwinship @liggitt  I could confirm the flaking is resolved by running `stress ./iptables.test` locally.

On master branch(commit 686ffd397f679472b62d1345f2e8ad507c80b342):
```
[root@56b79af149d7 iptables]# stress ./iptables.test
0 runs so far, 0 failures

/tmp/go-stress-20200910T061259-159706869
W0910 06:12:59.534090   80134 iptables.go:209] Error checking iptables version, assuming version at least 1.4.11: no iptables version found in string: total junk
I0910 06:13:01.545643   80134 trace.go:205] Trace[1298498081]: "iptables restore" (10-Sep-2020 06:12:59.544) (total time: 2000ms):
Trace[1298498081]: [2.000914s] [2.000914s] END
I0910 06:13:03.549184   80134 trace.go:205] Trace[2019727887]: "iptables restore" (10-Sep-2020 06:13:01.546) (total time: 2002ms):
Trace[2019727887]: [2.0021906s] [2.0021906s] END
--- FAIL: TestRestoreAllGrabOldLock (2.00s)
    iptables_test.go:1138: expected timeout error, got failed to acquire new iptables lock: timed out waiting for the condition
W0910 06:13:04.467634   80134 iptables.go:556] Could not set up iptables canary mangle/MONITOR-TEST-CANARY: error creating chain "MONITOR-TEST-CANARY": could not get xtables.lock, etc:
FAIL


ERROR: exit status 1


/tmp/go-stress-20200910T061259-923741136
W0910 06:12:59.532660   80138 iptables.go:209] Error checking iptables version, assuming version at least 1.4.11: no iptables version found in string: total junk
I0910 06:13:01.545517   80138 trace.go:205] Trace[1298498081]: "iptables restore" (10-Sep-2020 06:12:59.541) (total time: 2003ms):
Trace[1298498081]: [2.003542s] [2.003542s] END
--- FAIL: TestRestoreAllWaitOldIptablesRestore (2.00s)
    iptables_test.go:1028: expected success, got failed to acquire new iptables lock: timed out waiting for the condition
I0910 06:13:03.549082   80138 trace.go:205] Trace[2019727887]: "iptables restore" (10-Sep-2020 06:13:01.546) (total time: 2002ms):
Trace[2019727887]: [2.0020951s] [2.0020951s] END
I0910 06:13:05.555066   80138 trace.go:205] Trace[1427131847]: "iptables restore" (10-Sep-2020 06:13:03.550) (total time: 2004ms):
Trace[1427131847]: [2.0046564s] [2.0046564s] END
W0910 06:13:06.476182   80138 iptables.go:556] Could not set up iptables canary mangle/MONITOR-TEST-CANARY: error creating chain "MONITOR-TEST-CANARY": could not get xtables.lock, etc:
FAIL


ERROR: exit status 1

^C
```

After this patch:
```
[root@56b79af149d7 iptables]# stress ./iptables.test
0 runs so far, 0 failures
10 runs so far, 0 failures
20 runs so far, 0 failures
30 runs so far, 0 failures
40 runs so far, 0 failures
50 runs so far, 0 failures
60 runs so far, 0 failures
70 runs so far, 0 failures
80 runs so far, 0 failures
90 runs so far, 0 failures
100 runs so far, 0 failures
^C
```

liggitt:
this lgtm, will let @dcbw ack

pkg/util/iptables/iptables_test.go
```diff
@@ -35,7 +35,12 @@ import (
 	fakeexec "k8s.io/utils/exec/testing"
 )
 
-const TestLockfilePath = "xtables.lock"
+func getLockPaths() (string, string) {
+	lock14x := fmt.Sprintf("@xtables-%d", time.Now().Nanosecond())
+	lock16x := fmt.Sprintf("xtables-%d.lock", time.Now().Nanosecond())
+	// The abstract socket and lock file will be removed by calling `runner.Close()`
```
danwinship, line 41: There is no `runner.Close()` method. They get removed by `restoreInternal` whenever they're successfully created. I don't think there needs to be any comment about this here though. So just remove this, then lgtm.

knight42: Oh it is a typo.. I mean `locker.Close()`. Will remove it.

knight42:
@liggitt @danwinship Hi, would you mind adding the milestone?

pkg/util/iptables/iptables_test.go
```diff
@@ -934,8 +938,8 @@ func TestRestoreAll(t *testing.T) {
 			func(cmd string, args ...string) exec.Cmd { return fakeexec.InitFakeCmd(&fcmd, cmd, args...) },
 		},
 	}
-	runner := newInternal(&fexec, ProtocolIPv4, TestLockfilePath)
-	defer os.Remove(TestLockfilePath)
```
BenTheElder, line 938: are you sure this was a safe change?
I'm finding files like:
```
pkg/util/iptables/xtables-497740575.lock
pkg/util/iptables/xtables-499450913.lock
pkg/util/iptables/xtables-499708492.lock
```

and currently this is the most likely looking cause. #98947

PR #94555, opened 2020-09-05
Fix the bug that kubeadm tries to call 'docker info' even if the CRI socket was for another CR

### LINKED ISSUE #2270: kubeadm tries to call "docker info" even if the CRI socket was for another CR
this warning can be seen even if the user specified the containerd socket or if they let kubeadm detect the socket:

> W0831 17:19:34.759405     352 kubelet.go:200] cannot automatically set CgroupDriver when starting the Kubelet: cannot execute 'docker info -f {{.CgroupDriver}}': executable file not found in $PATH

we should try not showing the warning, for CRI != docker.

SataQiu:
**What type of PR is this?**
/kind bug

Fix the bug that kubeadm tries to call 'docker info' even if the CRI socket was for another CR

Fixes kubernetes/kubeadm#2270

```release-note
kubeadm: fix the bug that kubeadm tries to call 'docker info' even if the CRI socket was for another CR
```

```docs

```

### DIFF
--- a/cmd/kubeadm/app/util/config/cluster.go
+++ b/cmd/kubeadm/app/util/config/cluster.go
@@ -125,7 +125,7 @@ func getInitConfigurationFromCluster(kubeconfigDir string, client clientset.Inte
 		// However, if newControlPlane == true, initcfg.NodeRegistration is not used at all and it's overwritten later on.
 		// Thus it's necessary to supply some default value, that will avoid the call to DetectCRISocket() and as
 		// initcfg.NodeRegistration is discarded, setting whatever value here is harmless.
-		initcfg.NodeRegistration.CRISocket = constants.DefaultDockerCRISocket
+		initcfg.NodeRegistration.CRISocket = "/var/run/unknown.sock"
 	}
 	return initcfg, nil
 }


cmd/kubeadm/app/util/config/cluster.go
```diff
@@ -125,7 +125,11 @@ func getInitConfigurationFromCluster(kubeconfigDir string, client clientset.Inte
 		// However, if newControlPlane == true, initcfg.NodeRegistration is not used at all and it's overwritten later on.
 		// Thus it's necessary to supply some default value, that will avoid the call to DetectCRISocket() and as
 		// initcfg.NodeRegistration is discarded, setting whatever value here is harmless.
-		initcfg.NodeRegistration.CRISocket = constants.DefaultDockerCRISocket
+		if criSocket != "" {
+			initcfg.NodeRegistration.CRISocket = criSocket
+		} else {
+			initcfg.NodeRegistration.CRISocket = constants.DefaultDockerCRISocket
+		}
```
neolit123, line 132: I think we can omit the else case here because the initconfiguration defaulting will already srt it to docker on empty.

neolit123 (commented):
Thanks @SataQiu
The change makes sense and explains the problem.

/cc @rosti

cmd/kubeadm/app/util/config/cluster.go
SataQiu: Emm... But I think we'd better keep `else` here because of the above comment.
Anyway, we should set `initcfg.NodeRegistration.CRISocket` as some value here to avoid running `DetectCRISocket()` before it should be run. But for `kubeadm join` case, we should set it to the correct value to avoid executing `docker info`.

cmd/kubeadm/app/util/config/cluster.go
```diff
@@ -61,12 +61,12 @@ func (ue *unretriableError) Error() string {
 }
 
 // FetchInitConfigurationFromCluster fetches configuration from a ConfigMap in the cluster
-func FetchInitConfigurationFromCluster(client clientset.Interface, w io.Writer, logPrefix string, newControlPlane, skipComponentConfigs bool) (*kubeadmapi.InitConfiguration, error) {
+func FetchInitConfigurationFromCluster(client clientset.Interface, w io.Writer, logPrefix string, criSocket string, newControlPlane, skipComponentConfigs bool) (*kubeadmapi.InitConfiguration, error) {
```
rosti, line 64: The problem is that we are trying to do a couple of things with a single function here.
1. We pass `newControlPlane` parameter here to distinguish if we need to load just the ClusterConfiguration (+ optionally the component configs) or we need the NodeRegistrationOptions and local API address too. The fact is that the latter two are not needed in the case of `kubeadm join` and if we do `kubeadm upgrade node` on worker node and they need to be separated in another func.
2. We call `SetInitDynamicDefaults` in the func unconditionally where it should be called only if `newControlPlane == false`.

My proposal is the following:
1. Rename `getInitConfigurationFromCluster` to `FetchClusterConfigurationFromCluster`. Then change the return type to `*ClusterConfiguration`, remove the `newControlPlane` parameter from the func and move the code connected to it to the `FetchInitConfigurationFromCluster` func itself.
2. `FetchInitConfigurationFromCluster` should call `FetchClusterConfigurationFromCluster`. Then, in accordance with the `newControlPlane` parameter, it should call `getNodeRegistration`, `getAPIEndpoint`, and `SetInitDynamicDefaults`.
3. `fetchInitConfiguration` should be merged into `fetchInitConfigurationFromJoinConfiguration`. It should call `FetchClusterConfigurationFromCluster`. `fetchInitConfigurationFromJoinConfiguration` should also call `SetInitDynamicDefaults` at its end.
4. Sort of the same thing (as in 3) needs to be done for `kubeadm upgrade node` when upgrading worker node (possibly move `fetchInitConfigurationFromJoinConfiguration` right next to `FetchInitConfigurationFromCluster`?).
5. Adjust tests accordingly.

And, of course, this is caused by misusing InitConfiguration...

rosti (changes requested):
Thanks for taking on this @SataQiu !

cmd/kubeadm/app/util/config/cluster.go
SataQiu: Thank you for your detailed explanation @rosti . This makes sense! I'll have a try:)

SataQiu: @neolit123 @rosti 
I think I've found a simple way to avoid this problem. 
Just set an irrelevant value, and then we can skip the unnecessary checks.
Maybe we can fix the problem as easily as this :)

cmd/kubeadm/app/util/config/cluster.go
```diff
@@ -125,7 +125,7 @@ func getInitConfigurationFromCluster(kubeconfigDir string, client clientset.Inte
 		// However, if newControlPlane == true, initcfg.NodeRegistration is not used at all and it's overwritten later on.
 		// Thus it's necessary to supply some default value, that will avoid the call to DetectCRISocket() and as
 		// initcfg.NodeRegistration is discarded, setting whatever value here is harmless.
-		initcfg.NodeRegistration.CRISocket = constants.DefaultDockerCRISocket
+		initcfg.NodeRegistration.CRISocket = "/var/run/unknown.sock"
```
neolit123, line 128: @SataQiu based on the comments above the assignment this seems fine.

looks like @rosti is on vacation this week, so i'm going to LGTM / approve this, but let's iterate again in the 1.20 cycle if more changes are needed.

thanks.

PR #94572, opened 2020-09-06
fix: golint failures in `pkg/proxy/apis/{config, v1alpha1}` and `pkg/proxy/userspace`

mayankshah1607:
**What type of PR is this?**
/kind cleanup

fix golint failures in `pkg/proxy/apis/{config, v1alpha1}` and `pkg/proxy/userspace`

Fixes #

```release-note

```

```docs

```


pkg/proxy/userspace/roundrobin.go
```diff
@@ -34,8 +34,10 @@ import (
 )
 
 var (
+	// ErrMissingServiceEntry holds the error message for when the service entry is missing
 	ErrMissingServiceEntry = errors.New("missing service entry")
-	ErrMissingEndpoints    = errors.New("missing endpoints")
+	// ErrMissingEndpoints holds the error message for when the endpoints are missing
+	ErrMissingEndpoints = errors.New("missing endpoints")
```
aojea, line 40: should go.fmt warning about this? it seems it has not the same indentation

aojea:
/ok-to-test
please add the release-note NONE to the description

aojea:
you have to squash the commits where you have added and remove the file my mistake, so we only have 1 commit instead of two

pkg/proxy/apis/config/types.go
```diff
@@ -180,11 +180,13 @@ type KubeProxyConfiguration struct {
 // insufficient, this always falls back to the userspace proxy. IPVS mode will be enabled when proxy mode is set to 'ipvs',
 // and the fall back path is firstly iptables and then userspace.
 
+// ProxyMode variable is used to hold the mode in which the proxy will run.
```
thockin, line 183: If we're going to fix this comment, shouldn't we fold in the comment just above this?  E.g.

```
ProxyMode defines options for how the proxier implements proxying.  On the Linux platform the available values are ... On the Windows platform the available options are ...
```

The rest of these comments (fallbacks) should go to where this is used.

pkg/proxy/userspace/roundrobin.go
thockin: The comment makes it not have to

PR #94581, opened 2020-09-07
LockToDefault the ExternalPolicyForExternalIP feature gate

### LINKED ISSUE #94499: Feature gate ExternalPolicyForExternalIP is GA - when does it get removed?
When should it be removed?

@rikatz 

/sig network
/kind cleanup

knabben:
**What type of PR is this?**

/kind cleanup

Lock ExternalPolicyForExternalIP to default and remove the related code, it will be removed in 1.22.

Fixes https://github.com/kubernetes/kubernetes/issues/94499

```release-note
Lock ExternalPolicyForExternalIP to default, this feature gate will be removed in 1.22.
```

### DIFF
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -734,7 +734,7 @@ var defaultKubernetesFeatureGates = map[featuregate.Feature]featuregate.FeatureS
 	ImmutableEphemeralVolumes:                      {Default: true, PreRelease: featuregate.Beta},
 	DefaultIngressClass:                            {Default: true, PreRelease: featuregate.GA, LockToDefault: true}, // remove in 1.20
 	HugePageStorageMediumSize:                      {Default: true, PreRelease: featuregate.Beta},
-	ExternalPolicyForExternalIP:                    {Default: true, PreRelease: featuregate.GA}, // remove in 1.20
+	ExternalPolicyForExternalIP:                    {Default: true, PreRelease: featuregate.GA, LockToDefault: true}, // remove in 1.22
 	AnyVolumeDataSource:                            {Default: false, PreRelease: featuregate.Alpha},
 	DefaultPodTopologySpread:                       {Default: false, PreRelease: featuregate.Alpha},
 	SetHostnameAsFQDN:                              {Default: false, PreRelease: featuregate.Alpha},
--- a/pkg/proxy/iptables/BUILD
+++ b/pkg/proxy/iptables/BUILD
@@ -39,7 +39,6 @@ go_test(
     srcs = ["proxier_test.go"],
     embed = [":go_default_library"],
     deps = [
-        "//pkg/features:go_default_library",
         "//pkg/proxy:go_default_library",
         "//pkg/proxy/healthcheck:go_default_library",
         "//pkg/proxy/util:go_default_library",
@@ -54,8 +53,6 @@ go_test(
         "//staging/src/k8s.io/apimachinery/pkg/apis/meta/v1:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/types:go_default_library",
         "//staging/src/k8s.io/apimachinery/pkg/util/intstr:go_default_library",
-        "//staging/src/k8s.io/apiserver/pkg/util/feature:go_default_library",
-        "//staging/src/k8s.io/component-base/featuregate/testing:go_default_library",
         "//vendor/github.com/stretchr/testify/assert:go_default_library",
         "//vendor/k8s.io/klog/v2:go_default_library",
         "//vendor/k8s.io/utils/exec:go_default_library",
--- a/pkg/proxy/iptables/proxier.go
+++ b/pkg/proxy/iptables/proxier.go
@@ -1105,7 +1105,7 @@ func (proxier *Proxier) syncProxyRules() {
 
 				destChain := svcXlbChain
 				// We have to SNAT packets to external IPs if externalTrafficPolicy is cluster.
-				if !(utilfeature.DefaultFeatureGate.Enabled(features.ExternalPolicyForExternalIP) && svcInfo.OnlyNodeLocalEndpoints()) {
+				if !svcInfo.OnlyNodeLocalEndpoints() {
 					destChain = svcChain
 					writeLine(proxier.natRules, append(args, "-j", string(KubeMarkMasqChain))...)
 				}
--- a/pkg/proxy/iptables/proxier_test.go
+++ b/pkg/proxy/iptables/proxier_test.go
@@ -34,9 +34,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
-	utilfeature "k8s.io/apiserver/pkg/util/feature"
-	featuregatetesting "k8s.io/component-base/featuregate/testing"
-	"k8s.io/kubernetes/pkg/features"
 	"k8s.io/kubernetes/pkg/proxy"
 	"k8s.io/kubernetes/pkg/proxy/healthcheck"
 	utilproxy "k8s.io/kubernetes/pkg/proxy/util"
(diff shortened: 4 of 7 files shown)


rikatz:
@knabben Thank you!

As you may have noticed, one test is failing because of inconsistencies with bazel. 

This happens because some libraries has also been removed, so Bazel files needs to be re-created. 

Can you please run ./hack/update-bazel.sh to update the files and push the changes into another commit?

Also later I'll ask you to squash commits and rebase but right now, just update bazel files also and let's see if the tests passes. 

Thank you!

rikatz:
/lgtm
Thanks @knabben 


/assign @andrewsykim @freehan 
Folks, can you please take a look?

pkg/features/kube_features.go
```diff
@@ -738,7 +731,6 @@ var defaultKubernetesFeatureGates = map[featuregate.Feature]featuregate.FeatureS
 	ImmutableEphemeralVolumes:                      {Default: true, PreRelease: featuregate.Beta},
 	DefaultIngressClass:                            {Default: true, PreRelease: featuregate.GA, LockToDefault: true}, // remove in 1.20
 	HugePageStorageMediumSize:                      {Default: true, PreRelease: featuregate.Beta},
-	ExternalPolicyForExternalIP:                    {Default: true, PreRelease: featuregate.GA}, // remove in 1.20
```
andrewsykim, line 741: Based on https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecation I think we need to set `LockToDefault` for 2 releases before we can remove it.

knabben: @andrewsykim thanks, rolling it back and setting the LockToDefault

rikatz:
@knabben as per Andrew last comment, and per feature lifecycle (https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecation) please keep all the existing code into kube_features.go but change it to ``LockToDefault: true``.

The rest of the code (bazel, proxier_test.go, etc) can be safelly removed so the only needed changed here right now is the one in kube_features.go, maintaining the FeatureGate but removing any other code.

@thockin FYI, @liggitt  provided us a feedback in Slack about locking a feature gate to 'true' before finally removing it from the code, so the complete removal of the FeatureGate in kube_feature.go should be scheduled to milestone v1.22, sounds good?

We can keep the issue open so in v1.22 the feature can be completely removed following the deprecation lifecycle.

thockin:
Yeah, we need to get some clarity around lifecycle.  The code and docs are inconsistent.   :(

I suspect what we'll end up with is something like:

Feature goes GA:
- Gate set to GA, not locked
- Setting to non-default values is allowed but ignored (warning)

+ 1 release:
- Gate becomes locked
- Setting to non-default values is an error

+ 1 release:
- Gate is removed
- Setting to any value is an error
 

I will talk to Jordan again tomorrow.

knabben:
@thockin @liggitt any feedback on this? At this point (it wasn't set a LockToDefault but yet passed 2 releases) should we remove OR set LockToDefault and wait for more 2 releases?

liggitt:
> Setting to non-default values is allowed but ignored (warning)

I don't think we want that... that lets people upgrade an n-1 server where they overrode/disabled the feature to version n thinking they still have the feature off when in reality it is on.

That was the intent of [this language](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecation): 
> When an invocation tries to disable a non-operational feature gate, the call fails in order to avoid unsupported scenarios that might otherwise run silently.

rikatz:
Do we have any mechanism, when a FeatureGate is in GA but passed to the program to warn about its future deprecation/removal?

Shouldn't any GA FeatureGate still being used need to issue a Warning, and if it's already locked providing the additional message that it's configuration is going to be ignored?

liggitt:
> Do we have any mechanism, when a FeatureGate is in GA but passed to the program to warn about its future deprecation/removal?

Yes, setting a GA feature gate already prints a warning:

https://github.com/kubernetes/kubernetes/blob/323f34858de18b862d43c40b2cced65ad8e24052/staging/src/k8s.io/component-base/featuregate/feature_gate.go#L234-L235

rikatz:
Oh great, thank you @liggitt 

So maybe we can have a consensus here and for future cases like:

Moved to GA - will issue a warning. Feature needs to be scheduled to lock feature in +x releases

+x releases - GA + lock feature to default: all the code logic BUT the feature_gate.go code can be removed, will issue a warning or error (which one is best here? Probably a warning of deprecation and that the FeatureGate doesn't exists anymore)

+x+y releases - remove the feature_gate.go entry and this will issue an error, completing the fg deprecation lifecycle

Sounds good?

Particularly I think that x=2 and y=2 a too wide window, like we may take one year to fully remove a FG. But this is my opinion :)

liggitt:
> Moved to GA - will issue a warning. Feature needs to be scheduled to lock feature in +x releases

I think gates should lock to the default when they go GA. If we still need an escape hatch for people to disable it, that's probably a sign it either isn't actually GA-quality or needs a non-feature-gate mechanism to enable/disable ("Feature gates are intended to cover the development life cycle of a feature - they are not intended to be long-term APIs")

liggitt:
> I think gates should lock to the default and code checking the gate should be removed when they go GA.

This is on @thockin and my queue to hammer out… if this is where we end up, I'd envision spawning the following work items:
* audit of existing GA gates to verify conformance
* dev-facing docs describing this approach
* removal of the "LockToDefault" bit of GA implies locking to the default
* static analysis in presubmit to make sure nothing in-tree is still checking enablement of GA gates

rikatz:
> > Moved to GA - will issue a warning. Feature needs to be scheduled to lock feature in +x releases
> 
> I think gates should lock to the default and code checking the gate should be removed when they go GA. If we still need an escape hatch for people to disable it, that's probably a sign it either isn't actually GA-quality or needs a non-feature-gate mechanism to enable/disable ("Feature gates are intended to cover the development life cycle of a feature - they are not intended to be long-term APIs")

About this, my only concern is 'when a FeatureGate have it's default value changed, but not locked?'. My opinion in this is that once a FeatureGate is defaulted, every user get it working and this can raise some bugs/corner case that might need a fast action/workaround like 'disable this' while a patch isn't released.

OTOH I agree that a FeatureGate that turns into GA and have issues is a bug and needs a bugfix, and not a way to turn it off.

> audit of existing GA gates to verify conformance
About this, as suggested by @andrewsykim I've taken the question of 'how can we keep track of GA Feature Gates' and @johnbelamaric proposed that once a KEP is marked as Implemented, a new issue with the future milestone is going to be created in k/k repo for the owning SIG to mark/warn about the need of the FG removal from the code in the next milestone.

About the other items, don't know if I have competency to discuss about them :) But maybe this is a discussion we can move away of this PR :)

About the PR, my suggestion is that @knabben marks it as locked, remove all the logics from the code and we open a future issue to completely wipe it from the feature_gates.go code, wdyt?

liggitt:
> About this, my only concern is 'when a FeatureGate have it's default value changed, but not locked?'. My opinion in this is that once a FeatureGate is defaulted, every user get it working and this can raise some bugs/corner case that might need a fast action/workaround like 'disable this' while a patch isn't released.

That typically happens in the beta phase and spends at least a release enabled by default but still able to be disabled before graduating to GA and locking on.

> About the PR, my suggestion is that @knabben marks it as locked, remove all the logics from the code and we open a future issue to completely wipe it from the feature_gates.go code, wdyt?

That SGTM (include a comment beside the locked gate to indicate the issue and release to remove - probably 1.22)

andrewsykim (commented):
/approve
/lgtm

Thanks @knabben

andrewsykim:
/re-title LockToDefault the ExternalPolicyForExternalIP feature gate

andrewsykim:
/retitle LockToDefault the ExternalPolicyForExternalIP feature gate

andrewsykim:
/lgtm cancel

Can you update the commit message accordingly please?

andrewsykim:
/lgtm
/milestone v1.20 

Thanks!

rikatz:
/retest 
Maybe a flake? Doesn't seems related

PR #94589, opened 2020-09-07
fixes a data race in SerializeObject function

p0lyn0mial:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: It changes `deferredResponseWriter` to return after calling the `Close()` method.
Previously all sorts of errors including a data race were possible because `deferredResponseWriter` **resets** the writer and **returns** it to the pool. An attempt to write to a nil writer will lead to `invalid memory address or nil pointer dereference`. Sharing the same instance of `deferredResponseWriter` might lead to `index out of range [43] with length 30` and `recovered from err index > windowEnd` errors

Fixes #

```release-note
NONE

```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/writers.go
+++ b/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/writers.go
@@ -96,9 +96,11 @@ func SerializeObject(mediaType string, encoder runtime.Encoder, hw http.Response
 	err := encoder.Encode(object, w)
 	if err == nil {
 		err = w.Close()
-		if err == nil {
-			return
+		if err != nil {
+			// we cannot write an error to the writer anymore as the Encode call was successful.
+			utilruntime.HandleError(fmt.Errorf("apiserver was unable to close cleanly the response writer: %v", err))
 		}
+		return
 	}
 
 	// make a best effort to write the object if a failure is detected
--- a/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/writers_test.go
+++ b/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/writers_test.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -28,7 +29,7 @@ import (
 	"reflect"
 	"testing"
 
-	"k8s.io/apimachinery/pkg/api/errors"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/diff"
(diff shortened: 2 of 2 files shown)


staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/writers.go
```diff
@@ -96,9 +96,10 @@ func SerializeObject(mediaType string, encoder runtime.Encoder, hw http.Response
 	err := encoder.Encode(object, w)
 	if err == nil {
 		err = w.Close()
-		if err == nil {
-			return
+		if err != nil {
+			utilruntime.HandleError(fmt.Errorf("apiserver was unable to close cleanly the response writer: %v", err))
```
sttts, line 100: // we cannot write an error to the writer anymore as the Encode call was succesful.

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/writers.go
```diff
@@ -96,9 +96,10 @@ func SerializeObject(mediaType string, encoder runtime.Encoder, hw http.Response
...
 		}
+		return
```
sttts, line 102: change looks good otherwise.

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/writers.go
p0lyn0mial: done.

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/writers.go
liggitt: The whole point of &deferredResponseWriter was to be able to write useful errors to the response if there was an encoding error. We should be able to determine if the error came from writing to the underlying http writer, and if not, we should continue trying to write an error to the http response

/hold

liggitt: edit: never mind, misread the error flow… was thrown off by the non-standard ` if err == nil` early return pattern

/hold cancel

yogeek:
Hi @p0lyn0mial and @sttts ! Can this fix be backported to 1.19 or is the upgrade to 1.20 mandatory please ?

p0lyn0mial:
> Hi @p0lyn0mial and @sttts ! Can this fix be backported to 1.19 or is the upgrade to 1.20 mandatory please ?

it was already backported to 1.19 - https://github.com/kubernetes/kubernetes/pull/98393
this PR was targeting 1.20

liggitt:
this was already backported to 1.18 and 1.19

yogeek:
@p0lyn0mial @liggitt Oh ok, thank you for the quick reply.
So in my current 1.19.8 cluster, normally, the fix is already there ?

I ask this because I arrived here from this one : https://github.com/kubernetes-sigs/prometheus-adapter/issues/292 and we are in 1.19.8 and still observe these errors in the api-server logs : 
```
E0330 13:25:31.344690       1 writers.go:105] apiserver was unable to write a JSON response: http: Handler timeout
```

Additionnal note : we looked at apiserver logs to try to understand a recent increase in apiserver memory consumption that cause some of our masters to be "NotReady" lately.

liggitt:
> still observe these errors in the api-server logs :
> 
> ```
> E0330 13:25:31.344690       1 writers.go:105] apiserver was unable to write a JSON response: http: Handler timeout
> ```
> 
> Additionnal note : we looked at apiserver logs to try to understand a recent increase in apiserver memory consumption that cause some of our masters to be "NotReady" lately.

Those are unrelated to this issue. The symptoms of this issue are listed in the description:

> An attempt to write to a nil writer will lead to invalid memory address or nil pointer dereference. Sharing the same instance of deferredResponseWriter might lead to index out of range [43] with length 30 and recovered from err index > windowEnd errors

yogeek:
ok @liggitt thank you for your help

PR #85527, opened 2019-11-22
Revert "kube-proxy: check KUBE-MARK-DROP"

aojea:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:

This reverts commit 1ca0ffeaf2c4401549b82f549f7481313308d4b9.

kube-proxy is not recreating the rules associated to the
KUBE-MARK-DROP chain, that is created by the kubelet.

Is preferrable to avoid the dependency between the kubelet and
kube-proxy, so each of them handles their own rules.

**Which issue(s) this PR fixes**:

Fixes #85414

**Special notes for your reviewer**:

This is only needed for kube-proxy operating in dual-stack with iptables, and that PR wasn't merged.
Also, there is a discussion about this topic with a better solution than this
https://github.com/kubernetes/kubernetes/issues/82125#issuecomment-553255113

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #82214, opened 2019-09-01
Ensure the KUBE-MARK-DROP chain in kube-proxy mode=ipvs

uablrek:
**What type of PR is this?**
/kind bug
/sig network
/area ipvs

**What this PR does / why we need it**:

The `kube-proxy` expects the `KUBE-MARK-DROP` to exist. This chain is created by `kubelet`, it works but creates an unnecessary race. Ensuring the chain in both `kube-proxy` and `kubelet` removes the race without affecting the function.

Also in dual-stack the  `KUBE-MARK-DROP` is only created for the "cluster default" family by `kubelet` so dual-stack operation fails for `kube-proxy` unless the chain is created by other means, e.g with;
```
ip6tables -t nat -N KUBE-MARK-DROP
```

**Which issue(s) this PR fixes**:

Fixes #80462

**Special notes for your reviewer**:

Nobody seem to remember why the `KUBE-MARK-DROP` is created by `kubelet`.

**Does this PR introduce a user-facing change?**:

```release-note
Ensure the KUBE-MARK-DROP chain in kube-proxy mode=ipvs. The chain is ensured for both ipv4 and ipv6 in dual-stack operation.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

PR #94591, opened 2020-09-07
Fix failing test "Services should only allow access from service loadbalancer source ranges"

### LINKED ISSUE #94356: [failing test][ipvs] Services should only allow access from service loadbalancer source ranges
**Which jobs are failing**:

ci-kubernetes-e2e-gci-gce-ipvs

[sig-network] Services should only allow access from service loadbalancer source ranges [Slow]

**Testgrid link**:

https://k8s-testgrid.appspot.com/sig-network-gce#gci-gce-ipvs&width=5

**Reason for failure**:

```
namespace=services-9031 execpod-dropl5dxf -- /bin/sh -x -c wget -T 5 -qO- "35.203.137.95"'
Aug 31 06:41:19.868: INFO: stderr: "+ wget -T 5 -qO- 35.203.137.95\n"
Aug 31 06:41:19.868: INFO: stdout: "NOW: 2020-08-31 06:41:19.830826081 +0000 UTC m=+169.852164876"
Aug 31 06:41:19.868: INFO: Expect target NOT to be reachable. But it is reachable. Retry until timeout
Aug 31 06:41:19.868: INFO: Running '/workspace/kubernetes/platforms/linux/amd64/kubectl --server=https://34.82.86.196 --kubeconfig=/workspace/.kube/config exec --namespace=services-9031 execpod-dropl5dxf -- /bin/sh -x -c wget -T 5 -qO- "35.203.137.95"'
Aug 31 06:41:20.424: INFO: stderr: "+ wget -T 5 -qO- 35.203.137.95\n"
Aug 31 06:41:20.424: INFO: stdout: "NOW: 2020-08-31 06:41:20.387332326 +0000 UTC m=+170.408671114"
Aug 31 06:41:20.424: INFO: Expect target NOT to be reachable. But it is reachable. Retry until timeout
Aug 31 06:41:20.425: FAIL: Unexpected error:
    <*errors.errorString | 0xc0001fa1f0>: {
        s: "timed out waiting for the condition",
    }
    timed out waiting for the condition
occurred
```

The test is failing consistently, however, there is also test that is failing a lot 

>  [sig-network] ESIPP [Slow] should work from pods 

but not always, may it be related?

```

[sig-network] ESIPP [Slow] should work from pods | 11m37s
-- | --
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/network/service.go:3187 Aug 30 20:48:49.749: Source IP not preserved from pause-pod-deployment-55bc87cf68-bxf4f, expected '10.64.1.37' got '' /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/network/service.go:3242
```

Lion-Wei:
**What type of PR is this?**

/kind bug
/kind failing-test

kube-proxy IPVS mode will check `KUBE-MARK-DROP` chain in each sync, which will result in kube-proxy cleanup rules associated to the `KUBE-MARK-DROP` chain, and this will break firewall logic.

And this result in "Services should only allow access from service loadbalancer source ranges" case keep failing.

This pr removed check  `KUBE-MARK-DROP` in ipvs mode, to make sure kubelet and kube-proxy handle their own rules.

Fixes #94356

```release-note
NONE
```

### DIFF
--- a/pkg/proxy/iptables/proxier.go
+++ b/pkg/proxy/iptables/proxier.go
@@ -399,6 +399,13 @@ var iptablesJumpChains = []iptablesJumpChain{
 	{utiliptables.TableNAT, kubePostroutingChain, utiliptables.ChainPostrouting, "kubernetes postrouting rules", nil},
 }
 
+var iptablesEnsureChains = []struct {
+	table utiliptables.Table
+	chain utiliptables.Chain
+}{
+	{utiliptables.TableNAT, KubeMarkDropChain},
+}
+
 var iptablesCleanupOnlyChains = []iptablesJumpChain{}
 
 // CleanupLeftovers removes all iptables rules and chains created by the Proxier
@@ -864,6 +871,14 @@ func (proxier *Proxier) syncProxyRules() {
 		)
 		if _, err := proxier.iptables.EnsureRule(utiliptables.Prepend, jump.table, jump.srcChain, args...); err != nil {
 			klog.Errorf("Failed to ensure that %s chain %s jumps to %s: %v", jump.table, jump.srcChain, jump.dstChain, err)
+			return
+		}
+	}
+
+	// ensure KUBE-MARK-DROP chain exist but do not change any rules
+	for _, ch := range iptablesEnsureChains {
+		if _, err := proxier.iptables.EnsureChain(ch.table, ch.chain); err != nil {
+			klog.Errorf("Failed to ensure that %s chain %s exists: %v", ch.table, ch.chain, err)
 			return
 		}
 	}
--- a/pkg/proxy/ipvs/proxier.go
+++ b/pkg/proxy/ipvs/proxier.go
@@ -117,8 +117,14 @@ var iptablesChains = []struct {
 	{utiliptables.TableNAT, KubeNodePortChain},
 	{utiliptables.TableNAT, KubeLoadBalancerChain},
 	{utiliptables.TableNAT, KubeMarkMasqChain},
-	{utiliptables.TableNAT, KubeMarkDropChain},
 	{utiliptables.TableFilter, KubeForwardChain},
+}
+
+var iptablesEnsureChains = []struct {
+	table utiliptables.Table
+	chain utiliptables.Chain
+}{
+	{utiliptables.TableNAT, KubeMarkDropChain},
 }
 
 var iptablesCleanupChains = []struct {
@@ -1857,6 +1863,14 @@ func (proxier *Proxier) acceptIPVSTraffic() {
 func (proxier *Proxier) createAndLinkeKubeChain() {
 	existingFilterChains := proxier.getExistingChains(proxier.filterChainsData, utiliptables.TableFilter)
 	existingNATChains := proxier.getExistingChains(proxier.iptablesData, utiliptables.TableNAT)
+
+	// ensure KUBE-MARK-DROP chain exist but do not change any rules
+	for _, ch := range iptablesEnsureChains {
+		if _, err := proxier.iptables.EnsureChain(ch.table, ch.chain); err != nil {
+			klog.Errorf("Failed to ensure that %s chain %s exists: %v", ch.table, ch.chain, err)
+			return
+		}
+	}
 
 	// Make sure we keep stats for the top-level chains
 	for _, ch := range iptablesChains {


pkg/proxy/ipvs/proxier.go
```diff
@@ -117,7 +117,6 @@ var iptablesChains = []struct {
 	{utiliptables.TableNAT, KubeNodePortChain},
 	{utiliptables.TableNAT, KubeLoadBalancerChain},
 	{utiliptables.TableNAT, KubeMarkMasqChain},
-	{utiliptables.TableNAT, KubeMarkDropChain},
```
andrewsykim, line 120: If I recall correctly, this is only ensuring the drop chain doesn't exist but I don't think we're writing any rules here that would conflict with kubelet. Can you point me to where we're doing this?

aojea: lol, I've introduced this bug in kube-proxy iptables :) https://github.com/kubernetes/kubernetes/pull/85527

https://github.com/kubernetes/kubernetes/commit/9e6a6873527f2654be87fd153a8bcac969ac2d85

aojea:
/assign @andrewsykim @uablrek 
xref https://github.com/kubernetes/kubernetes/pull/85527

pkg/proxy/ipvs/proxier.go
andrewsykim: Ah I see -- so as part of the save/restore it is also clearing out rules that kubelet created. 

The issue that @uablrek mentioned in https://github.com/kubernetes/kubernetes/pull/82214 still seems valid though? Can we ensure the chain exists but not try any rules to it?

uablrek: How is the dual-stack situation handled for proxy-mode=iptables? I use a work-around by creating the chain in a script before K8s is started;
```
/etc/init.d/30kube-prep.rc:   ip6tables -t nat -N KUBE-MARK-DROP
```

uablrek: Since I start `kubelet` and `kube-proxy` from scripts I also get the race. `kube-proxy` crashes and re-starts a couple of times untilt the chain is created by `kubelet`. Not a big deal, but not the most elegant solotion either.

aojea: I think that the right fix is here https://github.com/kubernetes/kubernetes/pull/94474

uablrek: I still think it's appropriate that `kube-proxy` ensures the chain before is tries to use it but if adding it to the list also enables an undesired side-effect (sync) it must of course be corrected.

And IMHO ipvs and iptables mode should do this the same way, so the best is probably to revert the patch and wait for the fix above. Until then the chains should really be created beforehand to ensure dual-stack and, in my case, prevent the race.

aojea: xref https://github.com/kubernetes/kubernetes/issues/82125#issuecomment-624790496
:smile:

Lion-Wei: Thanks for reviewing 😄 

As what we did in #81517 , we chosed to belived nobody change or delete iptables chains created by kubernetes, except  system firewall be reloaded.

But we do have race issues like kube-proxy up before kubelet, maybe  ensures the chain before is tries to use it is a better way, but not try to change any rule.

After all, this is because `KUBE-MARK-DROP` chain created by kubelet and used by kube-proxy, as @aojea metioned, kube-proxy is the only consumer, maybe moving KUBE-MARK-DROP to the kube-proxies is another solution.

Overall, I think we have two potential solutions to solve this problem:
1. Ensure the chain exists but not try any rules to it (Maybe the same with `KUBE-MARK-MASQ`)
2.  Moving KUBE-MARK-DROP to the kube-proxies.

I think the first one is a better way to solve current failing test, the second way would need further discussion.

danwinship: > The issue that @uablrek mentioned in #82214 still seems valid though? Can we ensure the chain exists but not try any rules to it?

As Antonio said, the problem there is mostly that kube-proxy supports dual-stack and kubelet does not currently, and that should be fixed by #94474. Given #94474, the fix here is probably safe, although yes, I agree it would be better if ipvs could ensure that the chain exists but not delete it.

And then longer-term, #82125

andrewsykim: > I agree it would be better if ipvs could ensure that the chain exists but not delete it.

Can we do this now for both ipvs and iptables? Seems like the fix is to run `EnsureChain` for `KUBE-MARK-DROP` but not add anything to the nat chain buffer.

andrewsykim: https://github.com/kubernetes/kubernetes/blob/3fc1bc76a95d8a6a6cd33394526f00c00a98ce07/pkg/proxy/ipvs/proxier.go#L1866-L1882

pkg/proxy/ipvs/proxier.go
```diff
@@ -116,9 +116,9 @@ var iptablesChains = []struct {
 	{utiliptables.TableNAT, KubeFireWallChain},
 	{utiliptables.TableNAT, KubeNodePortChain},
 	{utiliptables.TableNAT, KubeLoadBalancerChain},
+	{utiliptables.TableFilter, KubeForwardChain},
 	{utiliptables.TableNAT, KubeMarkMasqChain},
 	{utiliptables.TableNAT, KubeMarkDropChain},
-	{utiliptables.TableFilter, KubeForwardChain},
```
andrewsykim, line 121: What's the reason for changing the order here?

pkg/proxy/ipvs/proxier.go
```diff
@@ -1867,6 +1859,11 @@ func (proxier *Proxier) createAndLinkeKubeChain() {
 			klog.Errorf("Failed to ensure that %s chain %s exists: %v", ch.table, ch.chain, err)
 			return
 		}
+		// KUBE-MARK-MASQ and KUBE-MARK-DROP is created in kubelet
+		// kube-proxy only ensure it exist but not do anything to those chains
+		if ch.chain == KubeMarkDropChain || ch.chain == KubeMarkMasqChain {
```
andrewsykim, line 1864: Instead of checking specifically for KubeMarkDropChain and KubeMarkMasqChain here and skipping, I would prefer a separate list of chains containing `KubeMarkDropChain` and `KubeMarkMasqChain` only (similar to iptablesChains) where we only run `proxier.iptables.EnsureChain`.

pkg/proxy/ipvs/proxier.go
```diff
@@ -1862,6 +1852,12 @@ func (proxier *Proxier) createAndLinkeKubeChain() {
 	existingNATChains := proxier.getExistingChains(proxier.iptablesData, utiliptables.TableNAT)
 
 	// Make sure we keep stats for the top-level chains
```
andrewsykim, line 1854: This comment still belongs below I think.

pkg/proxy/ipvs/proxier.go
```diff
@@ -1823,14 +1821,6 @@ func (proxier *Proxier) writeIptablesRules() {
 	}
 	writeLine(proxier.natRules, masqRule...)
 
-	// Install the kubernetes-specific masquerade mark rule. We use a whole chain for
-	// this so that it is easier to flush and change, for example if the mark
-	// value should ever change.
-	writeLine(proxier.natRules, []string{
-		"-A", string(KubeMarkMasqChain),
-		"-j", "MARK", "--or-mark", proxier.masqueradeMark,
```
andrewsykim, line 1831: I don't think it's safe to delete this yet?

pkg/proxy/ipvs/proxier.go
Lion-Wei: No reason, actually. 😂

pkg/proxy/ipvs/proxier.go
Lion-Wei: Sure, that might be better. Addressed. : )

pkg/proxy/ipvs/proxier.go
Lion-Wei: I think KubeMarkDropChain and KubeMarkMasqChain also chains that we should keep stat?

pkg/proxy/ipvs/proxier.go
Lion-Wei: Is that so? IMO, KUBE-MARK-MASQ and KUBE-MARK-DROP are similar.

andrewsykim: They are similar, but kube-proxy relies on kubelet to apply the drop rules in KUBE-MARK-DROP, where-as kube-proxy adds rules to KUBE-MARK-MASQ like here, which may overlap with kubelet.

pkg/proxy/iptables/proxier.go
```diff
@@ -920,6 +920,11 @@ func (proxier *Proxier) syncProxyRules() {
 			writeLine(proxier.natChains, utiliptables.MakeChainLine(chainName))
 		}
 	}
+	// ensure KUBE-MARK-DROP chain exist but not do any change
+	if _, err := proxier.iptables.EnsureChain(utiliptables.TableNAT, KubeMarkDropChain); err != nil {
+		klog.Errorf("Failed to ensure that %s chain %s exists: %v", utiliptables.TableNAT, KubeMarkDropChain, err)
+		return
```
aojea, line 926: in a dual stack cluster,  that we know this rule does not exist in one of the ip families, 
kube-proxy will stop to work in a one family or will stop to work at all?

Lion-Wei: Sorry I didn't notice this before. Then both ipvs and iptbales mode have this issue, I think one of the ip familier don't have this rule is a bug, maybe we can wait https://github.com/kubernetes/kubernetes/pull/94474 to be merged.

andrewsykim: Are we still waiting on https://github.com/kubernetes/kubernetes/pull/94474 for this? 

It seems like there are two separate issues here. For IPVS specifically, we are doing what iptables used to do before https://github.com/kubernetes/kubernetes/pull/85527. Sounds like we should fix that first independent of the dual-stack PR? Am I missing something there?

aojea:
@Lion-Wei I think that this is not the problem, the test is failing with the same error in other jobs with ipvs

https://github.com/kubernetes/kubernetes/issues/95225

pkg/proxy/iptables/proxier.go
```diff
@@ -920,6 +920,11 @@ func (proxier *Proxier) syncProxyRules() {
 			writeLine(proxier.natChains, utiliptables.MakeChainLine(chainName))
 		}
 	}
+	// ensure KUBE-MARK-DROP chain exist but not do any change
+	if _, err := proxier.iptables.EnsureChain(utiliptables.TableNAT, KubeMarkDropChain); err != nil {
```
andrewsykim, line 924: In a way this is reverting https://github.com/kubernetes/kubernetes/pull/85527, without having kube-proxy overwrite the iptable rules from kubelet. @cmluciano mentioned this fixed the issue on this environment, @aojea @uablrek can you check if this seems okay based on past discussions in https://github.com/kubernetes/kubernetes/pull/85527 and https://github.com/kubernetes/kubernetes/pull/82214

uablrek: Lgtm. The `EnsureChain` just creates the chain if it doesn't exist, if it already exists it does not return an error.

IMO it should be some rule that any program that uses a chain **must** call `EnsureChain` before use.

The (way too long) debate whenever that is right or not can go on independently, but at least things works.

aojea: I think that this should go before Line 872, 
https://github.com/kubernetes/kubernetes/pull/94591/files#diff-d51765b83fe795b469e8a86276b12dc9R872

the comment in that line is very explicit about it
```
// Below this point we will not return until we try to write the iptables rules.
```

and there is also a loop checking that the other chains exist

cmluciano:
@Lion-Wei Let me know if you need any more help working on this PR.

pkg/proxy/iptables/proxier.go
Lion-Wei: @aojea , Fixed, thanks.

pkg/proxy/iptables/proxier.go
```diff
@@ -868,6 +868,12 @@ func (proxier *Proxier) syncProxyRules() {
 		}
 	}
 
+	// ensure KUBE-MARK-DROP chain exist but not do any change
+	if _, err := proxier.iptables.EnsureChain(utiliptables.TableNAT, KubeMarkDropChain); err != nil {
```
andrewsykim, line 872: Similar to iptablesJumpChains can we have a variable like `iptablesEnsureChain` containing only KubeMarkDropChain?

pkg/proxy/ipvs/proxier.go
```diff
@@ -1858,6 +1857,12 @@ func (proxier *Proxier) createAndLinkeKubeChain() {
 	existingFilterChains := proxier.getExistingChains(proxier.filterChainsData, utiliptables.TableFilter)
 	existingNATChains := proxier.getExistingChains(proxier.iptablesData, utiliptables.TableNAT)
 
+	// ensure KUBE-MARK-DROP chain exist but not do any change
```
andrewsykim, line 1860:
```
// ensure KUBE-MARK-DROP chain exist but not do any change any rules
```

andrewsykim (commented):
Left some minor comments but overall I think this looks okay -- someone else with more context on kubelet / kube-proxy iptables ownership (https://github.com/kubernetes/kubernetes/issues/82125) should probably be the approver

aojea:
>  kubelet / kube-proxy iptables ownership (#82125) should probably be the approver
:)

/assign @danwinship @thockin

pkg/proxy/ipvs/proxier.go
Lion-Wei: Done

pkg/proxy/iptables/proxier.go
Lion-Wei: That will be more readable. Addressed : )

pkg/proxy/iptables/proxier.go
```diff
@@ -399,6 +399,13 @@ var iptablesJumpChains = []iptablesJumpChain{
 	{utiliptables.TableNAT, kubePostroutingChain, utiliptables.ChainPostrouting, "kubernetes postrouting rules", nil},
 }
 
+var iptablesEnsureChains = []struct{
+	table utiliptables.Table
+	chain utiliptables.Chain
+}{
+	{utiliptables.TableNAT, KubeMarkMasqChain},
```
andrewsykim, line 406: This should be KubeMarkDropChain?

pkg/proxy/ipvs/proxier.go
```diff
@@ -117,10 +117,16 @@ var iptablesChains = []struct {
...
 	{utiliptables.TableNAT, KubeMarkMasqChain},
-	{utiliptables.TableNAT, KubeMarkDropChain},
 	{utiliptables.TableFilter, KubeForwardChain},
 }
 
+var iptablesEnsureChains = []struct{
+	table utiliptables.Table
+	chain utiliptables.Chain
+}{
+	{utiliptables.TableNAT, KubeMarkMasqChain},
```
andrewsykim, line 127: This should be KubeMarkDropChain?

Lion-Wei: Oops, sorry

PR #91592, opened 2020-05-29
new NetworkPolicy Validation suite

jayunit100:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:

Implementation of https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20200204-NetworkPolicy-verification-rearchitecture.md 

**Which issue(s) this PR fixes**:

See https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20200204-NetworkPolicy-verification-rearchitecture.md 

**Special notes for your reviewer**:
none this is a WIP, just to share and hack on.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

<!--
This section can be blank if this pull request does not require a release note.

When adding links which point to resources within git repositories, like
KEPs or supporting documentation, please reference a specific commit and avoid
linking directly to the master branch. This ensures that links reference a
specific point in time, rather than a document that may change over time.

See here for guidance on getting permanent links to files: https://help.github.com/en/articles/getting-permanent-links-to-files

Please use the following format for linking documentation:
- [KEP]: https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20200204-NetworkPolicy-verification-rearchitecture.md
- [Usage]: `e2e.test --provider=local --ginkgo.focus="NetworkPolicy" --kubeconfig=/home/ubuntu/.kube/config`

PR #94597, opened 2020-09-07
Add NC, SCTP support and stages to Jig verifier

### LINKED ISSUE #94598: Update agnhost nc version, add support for SCTP to TestJig

- functionality for TestJig fixture might be able to be reused in var (example #91592)  (example #91592) ious emerging fixture based tests and the jig.go file should unify the functions under the Jig struct (so as to be able to add defensive checks in tests and so on that use the client of the jig).
- functionality to the agnhost to support sctp via newer nc versions since its a commonly used polling tool in k8s e2e.

**Why is this needed**:

- SCTP support to the TestJig as well  (example #91592 which creates pod fixtures and then polls them needs to fully support SCTP).
- as @aojea  points out, some containers (Such as current `agnhost`) dont have an `nc` version which supports `--sctp`.  This can i think easily be fixed by replacing the `apk` install from `netcat-openbsd` (which is old, 2017) to `nmap-netcat` (2019).  Since these all alias to the same `nc` command, its not going to change the way other tests need to run.

jayunit100:
**What type of PR is this?**

Fixes #94598

- Test jig method vs functions arent really needed as they're al internal to the Jig stuff
- The addition of test methods to to Jig mean that they can reuse the client and other fixtures we might eventually add
- We need SCTP verification as well

This might allow us to also use the JIG as the driver in the netpol test that is going on, and is part of an overall effort to make the E2Es more fixture driven for better reasoning over time.

/kind cleanup

Adding SCTP support to Test-jig and having it pre-verify execution target

Might be adopted in #91592

```release-note
NONE
```

### DIFF
--- a/test/e2e/framework/service/jig.go
+++ b/test/e2e/framework/service/jig.go
@@ -813,99 +813,6 @@ func (j *TestJig) waitForPodsReady(pods []string) error {
 	return nil
 }
 
-func testReachabilityOverServiceName(serviceName string, sp v1.ServicePort, execPod *v1.Pod) error {
-	return testEndpointReachability(serviceName, sp.Port, sp.Protocol, execPod)
-}
-
-func testReachabilityOverClusterIP(clusterIP string, sp v1.ServicePort, execPod *v1.Pod) error {
-	// If .spec.clusterIP is set to "" or "None" for service, ClusterIP is not created, so reachability can not be tested over clusterIP:servicePort
-	isClusterIPV46, err := regexp.MatchString(e2enetwork.RegexIPv4+"||"+e2enetwork.RegexIPv6, clusterIP)
-	if err != nil {
-		return fmt.Errorf("unable to parse ClusterIP: %s", clusterIP)
-	}
-	if isClusterIPV46 {
-		return testEndpointReachability(clusterIP, sp.Port, sp.Protocol, execPod)
-	}
-	return nil
-}
-
-func testReachabilityOverNodePorts(nodes *v1.NodeList, sp v1.ServicePort, pod *v1.Pod, clusterIP string) error {
-	internalAddrs := e2enode.CollectAddresses(nodes, v1.NodeInternalIP)
-	externalAddrs := e2enode.CollectAddresses(nodes, v1.NodeExternalIP)
-	isClusterIPV4 := net.ParseIP(clusterIP).To4() != nil
-
-	for _, internalAddr := range internalAddrs {
-		// If the node's internal address points to localhost, then we are not
-		// able to test the service reachability via that address
-		if isInvalidOrLocalhostAddress(internalAddr) {
-			framework.Logf("skipping testEndpointReachability() for internal adddress %s", internalAddr)
-			continue
-		}
-		isNodeInternalIPV4 := net.ParseIP(internalAddr).To4() != nil
-		// Check service reachability on the node internalIP which is same family
-		// as clusterIP
-		if isClusterIPV4 != isNodeInternalIPV4 {
-			framework.Logf("skipping testEndpointReachability() for internal adddress %s as it does not match clusterIP (%s) family", internalAddr, clusterIP)
-			continue
-		}
-
-		err := testEndpointReachability(internalAddr, sp.NodePort, sp.Protocol, pod)
-		if err != nil {
-			return err
-		}
-	}
-	for _, externalAddr := range externalAddrs {
-		isNodeExternalIPV4 := net.ParseIP(externalAddr).To4() != nil
-		if isClusterIPV4 != isNodeExternalIPV4 {
-			framework.Logf("skipping testEndpointReachability() for external adddress %s as it does not match clusterIP (%s) family", externalAddr, clusterIP)
-			continue
-		}
-		err := testEndpointReachability(externalAddr, sp.NodePort, sp.Protocol, pod)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
-}
-
-// isInvalidOrLocalhostAddress returns `true` if the provided `ip` is either not
-// parsable or the loopback address. Otherwise it will return `false`.
-func isInvalidOrLocalhostAddress(ip string) bool {
-	parsedIP := net.ParseIP(ip)
-	if parsedIP == nil || parsedIP.IsLoopback() {
-		return true
-	}
-	return false
-}
-
-// testEndpointReachability tests reachability to endpoints (i.e. IP, ServiceName) and ports. Test request is initiated from specified execPod.
-// TCP and UDP protocol based service are supported at this moment
-// TODO: add support to test SCTP Protocol based services.
-func testEndpointReachability(endpoint string, port int32, protocol v1.Protocol, execPod *v1.Pod) error {
-	ep := net.JoinHostPort(endpoint, strconv.Itoa(int(port)))
-	cmd := ""
-	switch protocol {
-	case v1.ProtocolTCP:
-		cmd = fmt.Sprintf("nc -zv -t -w 2 %s %v", endpoint, port)
-	case v1.ProtocolUDP:
-		cmd = fmt.Sprintf("nc -zv -u -w 2 %s %v", endpoint, port)
-	default:
-		return fmt.Errorf("service reachablity check is not supported for %v", protocol)
-	}
-
-	err := wait.PollImmediate(1*time.Second, ServiceReachabilityShortPollTimeout, func() (bool, error) {
-		if _, err := framework.RunHostCmd(execPod.Namespace, execPod.Name, cmd); err != nil {
-			framework.Logf("Service reachability failing with error: %v\nRetrying...", err)
-			return false, nil
-		}
-		return true, nil
-	})
-	if err != nil {
-		return fmt.Errorf("service is not reachable within %v timeout on endpoint %s over %s protocol", ServiceReachabilityShortPollTimeout, ep, protocol)
-	}
-	return nil
-}
-
 // checkClusterIPServiceReachability ensures that service of type ClusterIP is reachable over
 // - ServiceName:ServicePort, ClusterIP:ServicePort
 func (j *TestJig) checkClusterIPServiceReachability(svc *v1.Service, pod *v1.Pod) error {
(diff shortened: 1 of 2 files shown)


test/e2e/framework/service/jig.go
```diff
@@ -1051,3 +958,121 @@ func (j *TestJig) CreateSCTPServiceWithPort(tweak func(svc *v1.Service), port in
...
+	// 3) Finally take care of the polling
+	ep := net.JoinHostPort(endpoint, strconv.Itoa(int(port)))
+	cmd := ""
+	switch protocol {
+	case v1.ProtocolTCP:
+		cmd = fmt.Sprintf("nc -zv -t -w 2 %s %v", endpoint, port)
+	case v1.ProtocolUDP:
+		cmd = fmt.Sprintf("nc -zv -u -w 2 %s %v", endpoint, port)
+	case v1.ProtocolSCTP:
+		cmd = fmt.Sprintf("nc -zv --sctp -w 2 %s %v", endpoint, port)
```
aojea, line 1062: what netcat version is this? the one provided in the agnhost  image used by most of the test does not have this option
```
 nc -h
OpenBSD netcat (Debian patchlevel 1.130)
```

test/e2e/framework/service/jig.go
```diff
@@ -1051,3 +958,121 @@ func (j *TestJig) CreateSCTPServiceWithPort(tweak func(svc *v1.Service), port in
...
+// The following test functions build on top of the 'jig', but are still methods of it because so that
+// they can reuse some of the fixtures (i.e. the client).
+
+func (j *TestJig) testReachabilityOverServiceName(serviceName string, sp v1.ServicePort, execPod *v1.Pod) error {
+	return j.testEndpointReachability(serviceName, sp.Port, sp.Protocol, execPod)
+}
+
+func (j *TestJig) testReachabilityOverClusterIP(clusterIP string, sp v1.ServicePort, execPod *v1.Pod) error {
+	// If .spec.clusterIP is set to "" or "None" for service, ClusterIP is not created, so reachability can not be tested over clusterIP:servicePort
+	isClusterIPV46, err := regexp.MatchString(e2enetwork.RegexIPv4+"||"+e2enetwork.RegexIPv6, clusterIP)
```
aojea, line 971: you can use `isInvalidOrLocalhostAddress()` or 	
```
parsedIP := net.ParseIP(ip)
	if parsedIP == nil 
```

aojea:
> 
> Test jig method vs functions arent really needed as they're al internal to the Jig stuff
> The addition of test methods to to Jig mean that they can reuse the client and other fixtures we might eventually add

I defer to framework maintainers the decision, but shouldn't be better to move functions to methods when we have a reason to do that?  
I mean, it may happen that we want to export the functions in the future :shrug:

test/e2e/framework/service/jig.go
jayunit100: (move of the earlier code) but valid point i think

jayunit100: Right now this file has 2 sets of functions:
- those that are mehods of TestJig
- those that are not methods of TestJig

Now, it turns out that  one of the reusable functions here, `TestServiceReachability`, *needs* a client (so that it can preverify pods and so on before failing an exec...as that way the use of the function has much more clear test faillure resutls).

So, it seems like, since these funcs are all in the same file (and not heavily relied on outside this file) we, can couple all the test functionality here *to* the testJig, so that all functions in this file can reuse the other TestJig functionality.

*alternative*

now maybe we could also consider splitting the "TestJig" functionality out if we want to, and we could manually plumb the client into the function .

jayunit100:
> > Test jig method vs functions arent really needed as they're al internal to the Jig stuff
> > The addition of test methods to to Jig mean that they can reuse the client and other fixtures we might eventually add
> 
> I mean, it may happen that we want to export the functions in the future shrug

The reason it *seems* like these should all be TestJig methods, rather then standalone, is so that they can reuse the TestJig functionality (including client, so that preverification of pod states can happen before running `exec` statements and so on).

There is a workaround - we can plumb the client manually into testServiceReachability... and then have a wrapper func in TestJig? But that seems silly since the *only* reference to these functions is *from* the TestJig, and likely we'll be moving *more* towards a higher level fixture type tests over time as opposed to 'roll your own' style tests, hence the idea that you need to depend on TestJig for this functionality seems idiomatic to me... and also because, after all, the entire file is named *jig.go* :)

test/e2e/framework/service/jig.go
jayunit100: This function takes the pod as *input* so .  That said you are right, many pods might not support the `--sctp` option.  However *some pods* will (specifically, we need to support this for sure in #91592, so we'll make sure of that)

jayunit100: Looking more

- it looks like `netcat-openbsd` doesnt support sctp
- `netcat` the regular package *does* support sctp

But again, this function's ability to use sctp is going to be container dependent, regardless, and so it seems like its reasonable to say ' if the pod has an the current version of  nc, this will work, if it has an older version, it might not work'.

jayunit100:
Just as an idea - i pushed another commit here to update  `agnhost` which swaps out the old freebsd nc from 2015 with the newer nmap-ncat (2019).  Will see how that blows over :) @aojea  :)

jayunit100:
btw, not sure, but does `agnhost` get rebuild per-pull ?

aojea:
> btw, not sure, but does `agnhost` get rebuild per-pull ?

it has a process explained here
https://github.com/kubernetes/kubernetes/pull/94135#issuecomment-680037641

jayunit100:
Gotcha. Ok so we can merge this first before worrying about promoting the image which is fine .... because the new image will
Be backward compatible w the nc commands used for other tests .

test/images/agnhost/Dockerfile
```diff
@@ -27,7 +27,7 @@ CROSS_BUILD_COPY qemu-QEMUARCH-static /usr/bin/
 # - iproute2: includes ss used in NodePort tests
 # from iperf image
 # install necessary packages: iperf, bash
-RUN apk --update add bind-tools curl netcat-openbsd iproute2 iperf bash && rm -rf /var/cache/apk/* \
+RUN apk --update add bind-tools curl nmap-ncat iproute2 iperf bash && rm -rf /var/cache/apk/* \
```
aojea, line 30: in order to use this, you have to follow a process, it is described here
https://github.com/kubernetes/kubernetes/pull/94135#issuecomment-680037641

basically, you have to promote a new image first, and then submit a new PR with the test using it

jayunit100: ok, shall I move forward with that?  wanted to make sure the PR overall is in good standing b4

test/e2e/framework/service/jig.go
cmluciano: Trying this out in a related issue #96482 . Once you install the nmap-netcat, does it symlink `nc` to `ncat` ?

test/e2e/framework/service/jig.go
```diff
@@ -1051,3 +958,121 @@ func (j *TestJig) CreateSCTPServiceWithPort(tweak func(svc *v1.Service), port in
...
+	}
+
+	// 3) Finally take care of the polling
+	ep := net.JoinHostPort(endpoint, strconv.Itoa(int(port)))
+	cmd := ""
+	switch protocol {
+	case v1.ProtocolTCP:
+		cmd = fmt.Sprintf("nc -zv -t -w 2 %s %v", endpoint, port)
+	case v1.ProtocolUDP:
+		cmd = fmt.Sprintf("nc -zv -u -w 2 %s %v", endpoint, port)
```
cmluciano, line 1060: Why are we using the -z option here? Is there a noted reason to not send data? Also I'm wondering if we want to move away from netcat in general in favor of agnhost. Agnhost seems to better align with the way many connections in NetPol are tested [test/e2e/network/network_policy.go](https://github.com/kubernetes/kubernetes/blob/1c403c311e80dcdb075219d117ac803f8891de39/test/e2e/network/network_policy.go#L2049-L2051)

cmluciano (commented):
couple of comments about nc usage

test/images/agnhost/Dockerfile
cmluciano: Why are we adding ncat if we use nc above?

cmluciano (commented):
Comment on netcat addition

test/images/agnhost/Dockerfile
jayunit100: Ncat is the new one that has all the good stuff in it, like sctp.... I think this is what you want in your related Pr to add sctp functionality to the DialFromNode verifiers

test/e2e/framework/service/jig.go
jayunit100: In favor of agn host but I thought we still used nc in various places.  As long as we do use it.  The ability to use NCat (as opposed to the old netcat) gives us parity between sctp udp tcp ....

test/e2e/framework/service/jig.go
jayunit100: Yup it does (correct me if I'm wrong here been a few months since I put this patch up)

cmluciano: I don't think that it did, at least when I tried the test with your change to the Docker image, it just seemed to pickup the nc that is installed in BusyBox. 

I added the changes and tried to use the new ncat explicitly in the following branch that you can try out https://github.com/cmluciano/kubernetes/tree/cml/testnchelp

`export KUBE_TEST_REPO_LIS=$GOPATH/src/k8s.io/kubernetes/repo_list.yaml`
```bazel build //test/e2e:e2e.test && \
 cd src/k8s.io/kubernetes/bazel-out/k8-fastbuild-ST-77d62fd15d15fc7ab69058da1afba983b16c05cd7d7cd724cce6b37445e7a9d1/bin/test/e2e/e2e.test_ \
&& ./e2e.test -context kind-kind -ginkgo.focus="RunUdpTest" -ginkgo.trace```

When I try that out, my output shows nc pointing to the busybox version and it is symlinked to the normal `/usr/bin/nc`.

test/e2e/framework/service/jig.go
cmluciano: A quck scan for `nc -` matches in the codebase point toward main uses for udp port checking. This likely is due to agnhost connect cannot be used for UDP.

jayunit100:
Should I finish this Pr? Or close ? Have a few danglers so trying to clear my My queue :).

cmluciano:
> Should I finish this Pr? Or close ? Have a few danglers so trying to clear my My queue :).

I'm in favor of finishing it out if we can clear up issues that  mentioned regarding installation of ncat but e2es are now using the busybox nc. Ncat does not appear to symlink nc in my testing https://github.com/kubernetes/kubernetes/pull/94597#discussion_r529617882 .

This probably will need to be a 2 step change because the Dockerfile changes are promoted using the process @aojea mentioned here https://github.com/kubernetes/kubernetes/pull/94597#issuecomment-688456420.

Issue #80289, opened 2019-07-18
Migrate all uses of leader-election to use Lease API


wojtek-t:
Currently, Kubernetes components (scheduler, kcm, ...) are using leader election that is based on either Endpoints or ConfigMap objects.
Given that both of these are watched by different components, this is generating a lot of unnecessary load.

We should migrate all leader-election to use Lease API (that was designed exactly for this case).

The tricky part is that in order to do that safely, I think the only reasonable way of doing that would be to:
- in the first phase switch components to:
  1. acquire lock on the current object (endpoints or configmap)
  2. acquire lock on the new lease object
  3. only the proceed with its regular functionality
  [ loosing any of those two, should result in panicing and restarting the component]
- in the second phase (release after) remove point 1

@kubernetes/sig-scalability-bugs

wojtek-t:
@tedyu - if you would like to help with that, it would be great

wojtek-t:
This is basically a spin-off from https://github.com/kubernetes/kubernetes/issues/14733 that we never did.

shyamjvs:
@wojtek-t Do you think this is something we can add to sig-scalability "good first issues"?

wojtek-t:
Added help-wanted - not sure it's really a good "first contribution" issue. Will tag couple more.

ricky1993:
I can help to do with it. @wojtek-t

wojtek-t:
@ricky1993 - great! Please ping my on slack (wojtekt) if you have any questions about this.

tedyu:
I am also interested in contributing.
Since there are multiple components, it would be good to keep discussion on this issue so that more people can get to know how the migration is done.

tedyu:
@ricky1993 
Let me know your progress.
I assume I can contribute to the component(s) which you haven't covered.

justinsb:
A big user of leader-election is kubebuilder operators.  They use leader election to ensure that only a single instance is active.  Do we have advice on whether it's better to use leader-election or a StatefulSet of size 1 to achieve this?

Using leader election requires more RBAC permissions, and it sounds like it's suboptimal from a scalability point of view also; I'm wondering if there's a reason to prefer leader-election.

wojtek-t:
Actually, let me clarify couple things:
- the reason why leader-election use to be expensive is that it was using Endpoints object, which in turn is watched by all kube-proxies, DNS instances etc.
[...]

PR #94603, opened 2020-09-08
Migrate scheduler, controller-manager and cloud-controller-manager to use LeaseLock

wojtek-t:
Ref https://github.com/kubernetes/kubernetes/issues/80289

```release-note
Migrate scheduler, controller-manager and cloud-controller-manager to use LeaseLock
```

/kind cleanup
/priority important-longterm


staging/src/k8s.io/component-base/config/v1alpha1/defaults.go
```diff
@@ -44,7 +44,7 @@ func RecommendedDefaultLeaderElectionConfiguration(obj *LeaderElectionConfigurat
 		obj.RetryPeriod = metav1.Duration{Duration: 2 * time.Second}
 	}
 	if obj.ResourceLock == "" {
-		// TODO: Migrate to LeaseLock.
+		// TODO(#80289): Migrate to LeaseLock when graduating to v1beta1.
```
wojtek-t, line 47: @liggitt - who owns that part now?
I would be happy to promote it to v1beta1 in 1.20 to clean this up, but wanted to check with the owner if there aren't any known blockers for doing that (or other things we want to fix during this migration).

liggitt: I'm not sure... cc @mtaufen?

liggitt, line 47: Going straight from Endpoints -> Lease isn't safe, right? We have to go Endpoints -> EndpointsLeases -> Lease over three releases, right?

liggitt: Separately, I'm not sure we can know when it's safe to migrate to LeaseLock. With other components, we know their rollout cadence and skew support, so we can go endpoints -> endpointsleases -> leases over three versions. With this helper method, we have no idea what the consuming component's release schedule and skew support is.

wojtek-t: Yes - going Endpoints->Leases isn't safe.

wojtek-t: done

cmd/cloud-controller-manager/app/options/options.go
```diff
@@ -115,6 +116,10 @@ func NewCloudControllerManagerOptions() (*CloudControllerManagerOptions, error)
 // NewDefaultComponentConfig returns cloud-controller manager configuration object.
 func NewDefaultComponentConfig(insecurePort int32) (*ccmconfig.CloudControllerManagerConfiguration, error) {
 	versioned := &ccmconfigv1alpha1.CloudControllerManagerConfiguration{}
+	// Use lease-based leader election to reduce cost.
+	// The default endpoints-leases one has already been used for couple releases.
+	versioned.LeaderElection.ResourceLock = resourcelock.LeasesResourceLock
```
liggitt, line 121: be specific about which release we switched the default to endpointslease. also, why not change the default in SetDefaults_CloudControllerManagerConfiguration instead of here?

cmd/kube-controller-manager/app/options/options.go
```diff
@@ -210,6 +211,10 @@ func NewKubeControllerManagerOptions() (*KubeControllerManagerOptions, error) {
 // NewDefaultComponentConfig returns kube-controller manager configuration object.
 func NewDefaultComponentConfig(insecurePort int32) (kubectrlmgrconfig.KubeControllerManagerConfiguration, error) {
 	versioned := kubectrlmgrconfigv1alpha1.KubeControllerManagerConfiguration{}
+	// Use lease-based leader election to reduce cost.
+	// The default endpoints-leases one has already been used for couple releases.
+	versioned.LeaderElection.ResourceLock = resourcelock.LeasesResourceLock
```
liggitt, line 216: be specific about which release we switched the default to endpointslease. also, why not change the default in SetDefaults_KubeControllerManagerConfiguration instead of here?

cmd/kube-scheduler/app/options/options.go
```diff
@@ -138,6 +138,9 @@ func splitHostIntPort(s string) (string, int, error) {
 func newDefaultComponentConfig() (*kubeschedulerconfig.KubeSchedulerConfiguration, error) {
 	versionedCfg := kubeschedulerconfigv1beta1.KubeSchedulerConfiguration{}
 	versionedCfg.DebuggingConfiguration = *configv1alpha1.NewRecommendedDebuggingConfiguration()
+	// Use lease-based leader election to reduce cost.
+	// The default endpoints-leases one has already been used for couple releases.
+	versionedCfg.LeaderElection.ResourceLock = resourcelock.LeasesResourceLock
```
liggitt, line 143: be specific about which release we switched the default to endpointslease. also, why not change the default in SetDefaults_KubeSchedulerConfiguration instead of here?

cmd/cloud-controller-manager/app/options/options.go
wojtek-t: Added a comment about introducing in 1.17 (same below).

Modifying defaulting seems like backward-incompatible (though technically it's probably very unlikely someone will reuse it for something else...).

Should I really change to modifying the defaulting?

wojtek-t:
@liggitt - PTAL

liggitt:
/lgtm

would like an ack from someone on those component's teams:
/cc @ahg-g @cheftako

ahg-g:
are we making an exception here in the sense that a CC default value is being changed without increasing the version (to v1beta2 in the scheduler case)? Note that we are planning to make some changes to scheduler's CC and introduce v1beta2, so we could make this change in v1beta2 and keep the old behavior for v1beta1.

liggitt:
> are we making an exception here in the sense that a CC default value is being changed without increasing the version (to v1beta2 in the scheduler case)? Note that we are planning to make some changes to scheduler's CC and introduce v1beta2, so we could make this change in v1beta2 and keep the old behavior for v1beta1.

that would be fine as well. the controller manager configs aren't exposed as config files yet, so the default change is only affecting the CLI flag defaults

pkg/controller/apis/config/v1alpha1/defaults.go
```diff
@@ -130,7 +130,10 @@ func RecommendedDefaultGenericControllerManagerConfiguration(obj *kubectrlmgrcon
 	}
 
 	if len(obj.LeaderElection.ResourceLock) == 0 {
-		obj.LeaderElection.ResourceLock = "endpointsleases"
+		// Use lease-based leader election to reduce cost.
+		// We migrated for EndpointsLease lock in 1.17 and starting in 1.20 we
```
cheftako, line 134: We migrated *to* EndpointLease lock in 1.17

staging/src/k8s.io/component-base/config/v1alpha1/defaults.go
mtaufen: I don't think it's a particularly active file, but it's a core dependency so it would be good to stabilize the version. I'm not sure what the blockers would be off the top of my head. I'm happy to help if you want to work on moving it forward.

wojtek-t:
Cancelling hold based on Abdullah and Walter lgtms above.

/hold cancel

staging/src/k8s.io/component-base/config/v1alpha1/defaults.go
wojtek-t: OK - I will try to get back to in in the next couple weeks.

mtaufen: Ok, no rush as I'll be OOO for the next couple weeks anyway.

PR #94501, opened 2020-09-03
Fix staging/src/k8s.io/apimachinery/pkg/labels golint findings

ialidzhikov:
**What type of PR is this?**
/kind cleanup

**Which issue(s) this PR fixes**:

Part of #68026

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #94606, opened 2020-09-08
cleanup NamespaceScoped() comments of validating and mutating WebhookConfigurationStrategy

borgerli:
The comments for NamespaceScoped() method of validating and mutating WebhookConfigurationStrategy are misleading, saying that the method returns false. However, the method returns true insted.

This PR is just to fix this.


ialidzhikov:
This is actually done in #94501.

borgerli:
> This is actually done in #94501.

Cool. Closing this PR then.

PR #94610, opened 2020-09-08
Remove pkg/api/endpoints

soltysh:
**What type of PR is this?**
/kind cleanup
/sig network

This is a followup to #94112 where we removed `RepackSubsets` from `Canonicalize` in the storage layer. The fact that we have two copies of exactly same functions was the root cause of the problem. This ensures that we don't have that second copy, and interested parties should always use the one from https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/endpoints/util.go

/assign @liggitt @damemi 

```release-note
NONE
```


damemi (commented):
/lgtm
wonder if these functions would be good candidates for the incoming `component-helpers` repo in https://github.com/kubernetes/kubernetes/pull/92507?

PR #94611, opened 2020-09-08
[WIP] fix inotify memory leak

### LINKED ISSUE #94583: Inotify:memory leak due to blocked operation
**What happened**:
i wrote a shell：
```bash
#!/bin/bash

while [ true ]
do
        echo "reduce pod..."
        kubectl scale --replicas=0 rc ubuntu
        kubectl get pod -A | grep Running | wc -l
        sleep 60s
        kubectl get pod -A | grep Running | wc -l

        echo "add pod ..."
        kubectl scale --replicas=110 rc ubuntu
        kubectl get pod -A | grep Running | wc -l
        sleep 60s
        kubectl get pod -A | grep Running | wc -l
done

```
then the kubelet's RSS increases continuously and does not release，
i make some trace and found some watched objects have never been released  even these objects have not existed in real word
https://github.com/kubernetes/kubernetes/blob/16ea9dc6bc260c37589e6b104a47009680c26772/vendor/k8s.io/utils/inotify/inotify_linux.go#L103-L104

i notice the fucntion readEvent  is producer for these abnormal objects 
https://github.com/kubernetes/kubernetes/blob/16ea9dc6bc260c37589e6b104a47009680c26772/vendor/k8s.io/utils/inotify/inotify_linux.go#L135

and print the length of every event，some of which is beyond the length of buf：
https://github.com/kubernetes/kubernetes/blob/16ea9dc6bc260c37589e6b104a47009680c26772/vendor/k8s.io/utils/inotify/inotify_linux.go#L136

even when i increase the array by 10 times，the problem is still there and  the limits are broken through again and again.
so i realized the process of reading cannot be blocked.

now i added a queue for  the  buf of every inotify event and added another goroutine to consume the queue. everythings 
seems good for now.

fixed the bug
running the shell :
```bash
#!/bin/bash

while [ true ]
do
        echo "reduce pod..."
        kubectl scale --replicas=0 rc ubuntu
        kubectl get pod -A | grep Running | wc -l
        sleep 60s
        kubectl get pod -A | grep Running | wc -l

        echo "add pod ..."
        kubectl scale --replicas=50 rc ubuntu
        kubectl get pod -A | grep Running | wc -l
        sleep 60s
        kubectl get pod -A | grep Running | wc -l
done

```

**Environment**:
- Kubernetes version (use `kubectl version`):1.17.9 
```bash
NAME="Ubuntu"
VERSION="16.04 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04 LTS"
VERSION_ID="16.04"
UBUNTU_CODENAME=xenial

```
- Kernel (e.g. `uname -a`):
```bash
uname -a
Linux k8s 4.4.0-21-generic #37-Ubuntu SMP Mon Apr 18 18:33:37 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

```
- Install tools:
- Others:

yanghaichao12:
**What type of PR is this?**

/kind bug

fix inotify memory leak

Fixes #94583

### DIFF
--- a/vendor/k8s.io/utils/inotify/inotify_linux.go
+++ b/vendor/k8s.io/utils/inotify/inotify_linux.go
@@ -32,6 +32,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"sync"
 	"strings"
 	"syscall"
 	"unsafe"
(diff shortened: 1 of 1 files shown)


dims:
@yanghaichao12 please fix in https://github.com/kubernetes/utils before we can use it here.

yanghaichao12:
@dims okay

PR #94552, opened 2020-09-05
build: Fix go-runner arch in server images

justaugustus:
**What type of PR is this?**

/kind bug regression
/area release-eng dependency
/priority critical-urgent
/sig release

**What this PR dos / why we need it**:

**This will need to be cherry picked for the v1.19.1 release**

- build/lib/release: Explicitly use '--platform' in building server images

  When we switched to go-runner for building the apiserver,
  controller-manager, and scheduler server components, we no longer
  reference the individual architectures in the image names, specifically
  in the 'FROM' directive of the server image Dockerfiles.
  
  As a result, server images for non-amd64 images copy in the go-runner
  amd64 binary instead of the go-runner that matches that architecture.
  
  This commit explicitly sets the '--platform=linux/${arch}' to ensure
  we're pulling the correct go-runner arch from the manifest list.
  
  Before:
  ```dockerfile
  FROM ${base_image}
  ```

  After:
  ```dockerfile
  FROM --platform=linux/${arch} ${base_image}
  ```

- build/common.sh: Remove extraneous reference to debian-base image

  debian-base is no longer used for building core Kubernetes server
  components, so we remove the unnecessary local variable referencing it
  from `kube::build::get_docker_wrapped_binaries()`.

**Which issue(s) this PR fixes**:

Reported in https://github.com/kubernetes/kubernetes/issues/94540
cc: @smira

/assign @liggitt 
cc: @kubernetes/release-engineering 

**Special notes for your reviewer**:

I'll drop some debugging details in here later; just getting this up so CI can run.

**Does this PR introduce a user-facing change?**:

```release-note
Fixed a regression in non-amd64 server images in 1.19 that would not use the correct architecture base image
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

Issue #94540, opened 2020-09-04
kube-apiserver (and other official images) for arm64 contain /go-runner built for amd64


smira:
**What happened**:

Image: `k8s.gcr.io/kube-apiserver-arm64:v1.19.0`

kube-apiserver fails to start on arm64 with command line like: `/go-runner /usr/local/bin/kube-apiserver ....` with an error:

```
2020-09-04T18:18:52.840446862Z stderr F standard_init_linux.go:211: exec user process caused "exec format error"
```

`kube-apiserver` binary is built for arm64, while `go-runner` is built for amd64.

**What you expected to happen**:

kube-apiserver running normally, `go-runner` built for arm64

**How to reproduce it (as minimally and precisely as possible)**:

Extract contents of `k8s.gcr.io/kube-apiserver-arm64:v1.19.0` and check the executable architecture:

```
 $ file ./usr/local/bin/kube-apiserver
usr/local/bin/kube-apiserver: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, stripped
 $  file ./go-runner
go-runner: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, stripped
```

**Anything else we need to know?**:

**Environment**: n/a
- Kubernetes version (use `kubectl version`): 1.19.0
- Cloud provider or hardware configuration: n/a
- OS (e.g: `cat /etc/os-release`): Talos
- Kernel (e.g. `uname -a`):
- Install tools: Talos/bootkube
- Network plugin and version (if this is a network-related bug):
- Others:

liggitt:
cc @justaugustus

justaugustus:
@smira -- thanks for the report!
I'll take a look.

/assign

justaugustus:
Proposed fix in https://github.com/kubernetes/kubernetes/pull/94552.

justaugustus:
FYI @kubernetes/release-engineering

justaugustus:
Closing this now that [v1.19.1](https://groups.google.com/g/kubernetes-announce/c/1WdgO7uvXjM) has been released, which includes the [backport](https://github.com/kubernetes/kubernetes/pull/94613).

/close

PR #94613, opened 2020-09-08
[1.19] build: Fix go-runner arch in server images

justaugustus:
**What type of PR is this?**

/kind bug regression
/area release-eng dependency
/priority critical-urgent
/sig release

**What this PR dos / why we need it**:

**Cherry pick of https://github.com/kubernetes/kubernetes/pull/94552.**

- build/lib/release: Explicitly use '--platform' in building server images

  When we switched to go-runner for building the apiserver,
  controller-manager, and scheduler server components, we no longer
  reference the individual architectures in the image names, specifically
  in the 'FROM' directive of the server image Dockerfiles.
  
  As a result, server images for non-amd64 images copy in the go-runner
  amd64 binary instead of the go-runner that matches that architecture.
  
  This commit explicitly sets the '--platform=linux/${arch}' to ensure
  we're pulling the correct go-runner arch from the manifest list.
  
  Before:
  ```dockerfile
  FROM ${base_image}
  ```

  After:
  ```dockerfile
  FROM --platform=linux/${arch} ${base_image}
  ```

- build/common.sh: Remove extraneous reference to debian-base image

  debian-base is no longer used for building core Kubernetes server
  components, so we remove the unnecessary local variable referencing it
  from `kube::build::get_docker_wrapped_binaries()`.

Reported in https://github.com/kubernetes/kubernetes/issues/94540.

/assign @liggitt @dims 

```release-note
Fixed a regression in non-amd64 server images in 1.19 that would not use the correct architecture base image
```

```docs

```

### DIFF
--- a/build/common.sh
+++ b/build/common.sh
@@ -94,7 +94,6 @@ readonly KUBE_CONTAINER_RSYNC_PORT=8730
 # $1 - server architecture
 kube::build::get_docker_wrapped_binaries() {
   local arch=$1
-  local debian_base_version=v2.1.3
   local debian_iptables_version=v12.1.2
   local go_runner_version=buster-v2.0.0
   ### If you change any of these lists, please also update DOCKERIZED_BINARIES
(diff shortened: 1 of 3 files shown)


justaugustus:
/hold cancel
(parent PR has merged)

hasheddan (approved):
/lgtm

xmudrii (approved):
/lgtm

cpanato (approved):
/lgtm

PR #90839, opened 2020-05-07
Fix bandwidth burst too large

liuxu623:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Fix bandwidth burst too large

**Which issue(s) this PR fixes**:

Fixes #90838

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #94624, opened 2020-09-08
Deprecate Dockershim

dims:
See KEP in progress - https://github.com/kubernetes/enhancements/pull/1985

dockershim has been part of kubelet for far too long. we vendor in a whole bunch of stuff from moby/moby to make things work. The correct split/architecture is to follow the same design as with containerd and cri-o namely a separate process that kubelet talks to via grpc that in turn talks to a runtime. If folks are interested to do this, please do. An example of such a split is https://github.com/dims/cri-dockerd (NO! i am not maintaining that, nice try).

In this PR, we start the clock on removing dockershim in a subsequent release. We already have `dockerless` tag and CI jobs to ensure that we don't break anything when we eventually drop the built-in dockershim.

**What type of PR is this?**
/kind cleanup
/kind deprecation

Fixes #

```release-note
Action Required: Docker support in the kubelet is now deprecated and will be removed in a future release. The kubelet uses a module called "dockershim" which implements CRI support for Docker and it has seen maintenance issues in the Kubernetes community. We encourage you to evaluate moving to a container runtime that is a full-fledged implementation of CRI (v1alpha1 or v1 compliant) as they become available.
```

```docs

```

### DIFF
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -297,6 +297,7 @@ func PreInitRuntimeService(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 
 	switch containerRuntime {
 	case kubetypes.DockerContainerRuntime:
+		klog.Warningf("Using dockershim is deprecated, please consider using a full-fledged CRI implementation")
 		if err := runDockershim(
 			kubeCfg,
 			kubeDeps,


MHBauer:
As far as I can tell the only CI job is for typecheck, it's not tested with a built configuration, but I can't see how building without the code would break anything.
/lgtm

marosset:
The first of containerd release with Windows support was 1.4.0 and just released in Aug 2020.
The Windows community has been actively investing in containerd support since 1.18 and will continue to do so.

I'm concerned that marking dockershim as deprecated before ContainerD on Windows has had time to get any real adoption might cause some users to question Windows support in Kubernetes.

I'm in favor of deprecating dockershim and also removing that code from a Windows standpoint, I just think it might be better to do this in 1.21.

/cc @michmike

adammw:
> We now have robust implementations of the CRI interface

Is there a CRI-based implementation for docker? I haven't been able to find any guides on how to migrate to that - could someone link me to docs around that?

michmike:
> The first containerd release with Windows support was 1.4.0 and just released in Aug 2020.
> The Windows community has been actively investing in containerd support since 1.18 and will continue to do so but this is still new.
> 
> I'm concerned that marking dockershim as deprecated before ContainerD on Windows has had time to get any real adoption might cause some users to question Windows support in Kubernetes.
> 
> I'm in favor of deprecating dockershim and also removing that code from a Windows standpoint, I just think it might be better to do this in 1.21.
> 
> /cc @michmike

100% agree with Mark. we need to wait for containerd support in windows to stabilize and iron out any kinks. 1.21 is the right timeframe

neolit123:
> The kubelet has a built-in dockershim which was used to support Docker. We now have robust implementations of the CRI interface, so we will be removing dockershim in a subsequent release

possibly requires some language clarification - are removing it entirely or just from the kubelet?

moving dockershim outside of the kubelet to a community maintained repository seems the right action here. this survey from KubeCon EU (this year) performed by SIG Cluster Lifecycle:
https://docs.google.com/forms/d/1AA75xUUdsboenrGPFN-Dtnx8N8vfR_b3kmmQNrdXX4g/edit#responses

search for "What is your preferred container runtime (CRI)?". Docker still has majority but containerd is gaining traction slowly.

dims:
@marosset @michmike Starting the countdown does not imply we *have* to remove things in a specific time period. The more heads up people get the better. So i'd rather throw this out earlier (in 1.20) than later (1.21).

dims:
@MHBauer we had added a kind CI job that used the `dockerless` tag. Did that get removed?

liggitt:
> @marosset @michmike Starting the countdown does not imply we _have_ to remove things in a specific time period. The more heads up people get the better. So i'd rather throw this out earlier (in 1.20) than later (1.21).

I agree

neolit123:
as mentioned, i don't entirely agree with the plan here...this PR effectively deprecates Docker support in Kubernetes, but the release note does not communicate this clearly.

what would be  guide to explain the users how to move their cluster from docker to e.g. containerd?
immutable node upgrades are maybe fine, but what about in-place on limited # bare metal machines? matching cgroup drivers? state cleanup?

my proposal:
> Action Required: Docker support in the kubelet is now deprecated and will be removed in a future release. The kubelet uses a module called "dockershim" which implements CRI support for Docker and has seen maintenance issues in the Kubernetes community. We encourage you to move to a container runtime that is a full-fledged implementation of CRI instead. To transition your existing clusters follow [this guide](link)

and the "this guide" document has to be created, before we apply this deprecation.

pkg/kubelet/kubelet.go
```diff
@@ -297,6 +297,7 @@ func PreInitRuntimeService(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 
 	switch containerRuntime {
 	case kubetypes.DockerContainerRuntime:
+		klog.Warningf("Using Docker is deprecated, please consider using a full-fledged CRI implementation")
```
jayunit100, line 300: thanks, should we allude to the fact that it's not being actively maintained anymore (re would help make issues like  https://github.com/kubernetes/kubernetes/pull/90839 go by faster :)

liggitt: indicating the built-in dockershim implementation is deprecated would be clearer... using docker via a CRI impl could still be reasonable, right?

dims:
@neolit123 i'd prefer the community to write the "guide". I am just interested in starting the clock in this PR.

pkg/kubelet/kubelet.go
dims: Ack @liggitt

alculquicondor:
Playing devil's advocate here, but shouldn't CRI itself move out of alpha before dockershim is deprecated?

https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1alpha2

liggitt:
> shouldn't CRI itself move out of alpha before dockershim is deprecated?

As far as I know, it is alpha in name only. I agree that if it is being treated as a GA API surface it should be named accordingly.

dims:
@alculquicondor @liggitt +1 to file a subsequent PR to promote the CRI API. I just need to start the clock here!

guineveresaenger:
hi everyone! from an end user's perspective -

1. what is the KEP associated with this work?

2. I want to +1 what @neolit123 said - I would like to see a migration guide. I'm not sure what you mean by having "the community" write it - that feels a bit, well, not user friendly?

neolit123:
> what is the KEP associated with this work?

a related KEP that i know of is:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20200205-build-kubelet-without-docker.md

we should like to the KEP in the OP under:

>> Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

also there is no tracking issue under kubernetes/enhancements for this KEP / work. cc @mattjmcnaughton
we should create one and link it under the same block, so that users can comment somewhere.

dims:
@guineveresaenger This is a one line intention of something that we need to march towards. Whoever is interested needs to band together under sig-node to work on this. It's totally unfair to say that a one line deprecation warning message needs a full guide that covers all aspects of all the permutations and combinations of how folks use docker today and a way to migrate everything.

In this PR, the intent is to warn anyone who will be installing 1.20 that they should be considering an alternate CRI implementation for a new cluster. There are quite a few choices other than docker and they need to explore those. This does *not* set a date for when we will remove dockershim. So asking for a guide to exist is not fair.

marosset:
From my experience enterprise users tend to prefer Docker because they can pay for a support subscription.
These same users tend to notice and care about messages that they are using deprecating features like the one being added.

I don't think this needs a full KEP but do agree with others that some form of user-facing messaging/doc around motivations/alternatives/timelines would be a good idea here.

liggitt:
> From my experience enterprise users tend to prefer Docker because they can pay for a support subscription.
> These same users tend to notice and care about messages that they are using deprecating features like the one being added.

Note that docker can still be used, but via the CRI interface rather than the built-in docker-specific kubelet<->docker integration

marosset:
> Note that docker can still be used, but via the CRI interface rather than the built-in docker-specific kubelet<->docker integration

Not for Windows unfortunately.

derekwaynecarr:
I am +1 on announcing the deprecation while being sensitive to our user community.
I think we should author a KEP that documents how that deprecation is coordinated BEFORE merging this PR.
I am happy to work with @dims on that document.

dims:
@marosset very eager to hear why not? ( hope you caught the https://github.com/dims/cri-dockerd for any experiments )

marosset:
I had a chat with @dims on slack
tl;dr; - For Windows, dockershim contains some workarounds needed to have the kubelet play nicely with docker. 
For containerd on Windows we have more flexibility to iterate/address issues in the Windows OS itself and/or containerd and have not been adding workarounds for CRI codepaths.
The cri-dockerd example shared still uses a dockershim hosted outside of kubelet and may work but at that point we might as well just use dockershim hosted by kubelet.

As mentioned above we've been investing in containerd support for Windows and are hoping to finish stabilizing everything in 1.20. The proposal to add doc user-facing messaging around timelines/expectations for dockershim deprecation satisfies all of my concerns here.

dims:
Folks, please review the KEP https://github.com/kubernetes/enhancements/pull/1985

neolit123:
i tested hot-swapping the CR (docker -> containerd), in-place, on a small cluster yesterday and it worked fine.
my concerns were around whether the kubelet (sandbox) would tolerate the socket change after a restart, but the worked.

happy to discuss more.

danielsand:
> i tested hot-swapping the CR (docker -> containerd), in-place, on a small cluster yesterday and it worked fine (to my understanding, at least)
> my concerns were around whether the kubelet (sandbox) would tolerate the socket change after a restart, but that worked.
> 
> happy to discuss more.

some hands on expierience

the problem i have seen from switching afterwards is as example the use of **docker build** from a CI like concourse as example.
crio-ctl has no build option. crio seperated it to crio-build

switching from docker - container-d / cri-o on a bare-metal cluster running concourse broke the container build pipeline. 
So the switch is not 

there is **crio-build** but i dropped further investigations and joined a docker k8s node to monkey-patch the problem.

liggitt:
> as example the use of docker build from a CI like concourse as example.

Build capabilities are not exposed via the Kubernetes API or used by Kubernetes components like the kubelet. Nothing precludes docker from remaining available on nodes for use by build systems that make use of it.

neolit123:
> the problem i have seen from switching afterwards is as example the use of docker build from a CI like concourse as example.
crio-ctl has no build option.

`docker build` uses buildkit:
https://github.com/moby/buildkit
https://docs.docker.com/develop/develop-images/build_enhancements/

and it can run as a Service in k8s too:
https://github.com/estesp/buildkit-cluster-example

danielsand:
> > the problem i have seen from switching afterwards is as example the use of docker build from a CI like concourse as example.
> > crio-ctl has no build option.
> 
> `docker build` uses buildkit:
> https://github.com/moby/buildkit
> https://docs.docker.com/develop/develop-images/build_enhancements/
> 
> and it can run as a Service in k8s too:
> https://github.com/estesp/buildkit-cluster-example

sure - im not talking about that its impossible. 
Im talking on the fact that without big Warning signs this will create a lot of downtime and frustration.
thats all

kfox1111:
Since docker depends on containerd these days, can't you just run both to handle the concourse issue? Install docker which should get you containerd, plumb k8s up to containerd directly and docker build would still work?

derekwaynecarr:
@dims given that we had to push cri proto updates to span two releases, i think this needs to move to 1.21.

derekwaynecarr:
i think this is for the best, deprecated is still supported, and this gives the right signal to prepare for 1.21.  updated the release note text to refer to either v1alpha1 or v1 (when available) as it captures the reality of world today.

/approve
/lgtm

dhiltgen:
A number of folks have commented about still relying on `docker build` in their environments.  There are various projects and approaches to build on a kubernetes cluster, but if you happen to be looking for a "drop in" replacement for the `docker build` UX, you might consider https://github.com/vmware-tanzu/buildkit-cli-for-kubectl/ as a possible option to consider (which works great on k8s+containerd)

IAXES:
Good day,

In simple/layperson terms, does this mean that Docker containers will not be supported at all in the future? For example, Docker's landing page notes that it **uses** `containerd` (i.e. "Containerd is used by Docker, Kubernetes CRI, and a few other projects" [1]). Is `containerd` able to effectively work as a drop-in replacement? If not, my understanding is that, if we want to use newer builds of K8S going forward (i.e. once the deprecated features are removed outright), individuals/teams would have to (please feel free to correct any of these points if they're off the mark):

- Change how they build images, i.e.:
  - Re-write build scripts to be something other than `Dockerfile`.
  - Use tooling other than `docker build`.
  - Use different image registry tools for publishing images.

If the above assumptions are correct, are there de-facto (i.e. most commonly used in the K8S community, or recommended by K8S developers and maintainers) replacements for Docker (i.e. in terms of tooling, publishing, etc.) so individuals/teams can migrate effectively? 

Thank you.

1. "What is containerd?", <https://www.docker.com/blog/what-is-containerd-runtime/>, last accessed 2020-12-02.

BenTheElder:
In simple terms: docker containers / images will absolutely be supported. docker may not be the tool running them, but the same Kubernetes YAML, docker images, etc. will continue to work with other container runtimes.

Images use an open format common to all runtimes, you don't need to change image builds.

It should be a drop in replacement. Our (Kubernetes) CI uses them interchangably. If you've ever used microk8s, k3s, or KIND then you've already used containerd to run Kubernetes workloads.

This should only affect cluster operators / implementors,

IAXES:
> 
> 
> In simple terms (I hope, can further clarify if needed): docker containers / images will absolutely be supported. docker may not be the tool running them, but the same Kubernetes YAML, docker images, etc. will continue to work with other container runtimes.
> 
> Images use an open format common to all runtimes, you don't need to change image builds.
> 
> It should be a drop in replacement. Our (Kubernetes) CI uses them interchangably. If you've ever used microk8s, k3s, or KIND then you've already used containerd to run Kubernetes workloads.
> 
> This should only affect cluster operators / implementors,

Thank you very much. This is indeed helpful.

neolit123:
@dims i'd add a message in the release note that this is not deprecating support for Docker containers...
the next pre-release will pick the updated release note.

adamparco:
A happy update: [Mirantis](https://github.com/mirantis/cri-dockerd) and Docker have partnered on supporting the dockershim as a standalone component outside of Kube.  You can read more about it here: https://www.mirantis.com/blog/mirantis-to-take-over-support-of-kubernetes-dockershim-2/.  Hopefully this will put any remaining lingering concerns to rest.

dims:
@adamparco thanks for the good news, let's chat about a migration plan and testing strategies to make sure we are able to support folks who will want to stick to using docker in the future. 

For those who will check on this KEP, the plan is still in progress :) i.e, investigate and switch over to a CRI implementation that's available in the community (containerd, cri-o etc).

If you want to continue to use docker, You will need to switch to new stuff that will show up in https://github.com/Mirantis/cri-dockerd as we will drop dockershim per the timetable in the KEP.

Thanks!

zhucan:
@dims  Is it needed  to install docker-ce when install kubernetes， if docker-shim been deprecated .

dims:
@zhucan please see the 2 blog posts
- https://kubernetes.io/blog/2020/12/02/dont-panic-kubernetes-and-docker/
- https://kubernetes.io/blog/2020/12/02/dockershim-faq/

dims:
Please see https://github.com/Mirantis/cri-dockerd

PR #94629, opened 2020-09-08
Sort list of formats for --logging-format description to make it deterministic

### LINKED ISSUE #94628: 1.19: 1 bit non-determinism in kube-scheduler.1
**What happened**:
50% of builds have a different man page content

It should be possible to get bit-identical build results - might need a patch similar to #68983 - @dims knows these things.

build kubernetes1.19 openSUSE package a few times and compare results. e.g. with [my rbk script](https://github.com/bmwiedemann/reproducibleopensuse/blob/master/rbk) 

**Environment**:
- OS: openSUSE-Tumbleweed 20200906

While working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds), I found that
when building the kubernetes1.19 package, there was sometimes such a difference between builds:

```diff
--- old//usr/share/man/man1/kube-scheduler.1    2020-09-08 18:04:00.987086806 +0000
+++ new//usr/share/man/man1/kube-scheduler.1    2020-09-08 18:04:01.015086912 +0000
@@ -257,7 +257,7 @@

 .PP
 \fB\-\-logging\-format string                                   Sets the log format. Permitted formats\fP
-        "json", "text".
+        "text", "json".

 .PP
 .RS
```

dims:

**What type of PR is this?**

/kind cleanup
/kind flake

Fixes #94628

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/component-base/logs/registry.go
+++ b/staging/src/k8s.io/component-base/logs/registry.go
@@ -18,6 +18,7 @@ package logs
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/go-logr/logr"
 	json "k8s.io/component-base/logs/json"
@@ -84,12 +85,13 @@ func (lfr *LogFormatRegistry) Delete(name string) error {
 	return nil
 }
 
-// List names of registered log formats
+// List names of registered log formats (sorted)
 func (lfr *LogFormatRegistry) List() []string {
 	formats := make([]string, 0, len(lfr.registry))
 	for f := range lfr.registry {
 		formats = append(formats, f)
 	}
+	sort.Strings(formats)
 	return formats
 }
 


monotek (approved):
/lgtm

ehashman (approved):
/approve
/lgtm

bmwiedemann:
Tested the patch. Works fine.

The non-determinism probably comes from https://golang.org/src/hash/maphash/maphash.go?#L170 which uses randomness to avoid attacks where user-supplied hash keys are chosen to force the hash-table into a long list that slows processing down as a DoS.

I wonder if it would make it easier to be reproducible-by-default if go had a hash that has a constant hash seed for cases without user-input. Or if there was something like `$PERL_HASH_SEED` that can be set globally in the build environment.

PR #94636, opened 2020-09-09
Added config parameter for CPU threads

### LINKED ISSUE #94011: Scheduler: Make number of threads configurable

A field in kube-scheduler configuration API to set the number of threads. Currently, it's a constant with value 16.

https://github.com/kubernetes/kubernetes/blob/3b5aedcef43fa01538518ecbc7de369b27ecd191/pkg/scheduler/internal/parallelize/parallelism.go#L26

**Why is this needed**:

Some clusters might have master nodes with more CPUs available.

/sig scheduling
/good-first-issue

hprateek43:
**What type of PR is this?**
/kind feature

Adds configurable CPU threads for scheduler.

Fixes #94011

```release-note
NONE
```

```docs

```

### DIFF
--- a/cmd/kube-scheduler/app/options/options_test.go
+++ b/cmd/kube-scheduler/app/options/options_test.go
@@ -298,6 +298,7 @@ profiles:
 			},
 			expectedUsername: "config",
 			expectedConfig: kubeschedulerconfig.KubeSchedulerConfiguration{
+				Parallelism:        16,
 				AlgorithmSource:    kubeschedulerconfig.SchedulerAlgorithmSource{Provider: &defaultSource},
 				HealthzBindAddress: "0.0.0.0:10251",
 				MetricsBindAddress: "0.0.0.0:10251",
@@ -395,6 +396,7 @@ profiles:
 			},
 			expectedUsername: "flag",
 			expectedConfig: kubeschedulerconfig.KubeSchedulerConfiguration{
+				Parallelism:        16,
 				AlgorithmSource:    kubeschedulerconfig.SchedulerAlgorithmSource{Provider: &defaultSource},
 				HealthzBindAddress: "", // defaults empty when not running from config file
 				MetricsBindAddress: "", // defaults empty when not running from config file
@@ -459,6 +461,7 @@ profiles:
 				Logs: logs.NewOptions(),
 			},
 			expectedConfig: kubeschedulerconfig.KubeSchedulerConfiguration{
+				Parallelism:        16,
 				AlgorithmSource:    kubeschedulerconfig.SchedulerAlgorithmSource{Provider: &defaultSource},
 				HealthzBindAddress: "", // defaults empty when not running from config file
 				MetricsBindAddress: "", // defaults empty when not running from config file
@@ -498,6 +501,7 @@ profiles:
 			},
 			expectedUsername: "config",
 			expectedConfig: kubeschedulerconfig.KubeSchedulerConfiguration{
+				Parallelism:        16,
 				AlgorithmSource:    kubeschedulerconfig.SchedulerAlgorithmSource{Provider: &defaultSource},
 				HealthzBindAddress: "0.0.0.0:10251",
 				MetricsBindAddress: "0.0.0.0:10251",
@@ -572,6 +576,7 @@ profiles:
 			},
 			expectedUsername: "config",
 			expectedConfig: kubeschedulerconfig.KubeSchedulerConfiguration{
+				Parallelism:        16,
 				AlgorithmSource:    kubeschedulerconfig.SchedulerAlgorithmSource{Provider: &defaultSource},
 				HealthzBindAddress: "0.0.0.0:10251",
 				MetricsBindAddress: "0.0.0.0:10251",
@@ -648,6 +653,7 @@ profiles:
 			},
 			expectedUsername: "flag",
 			expectedConfig: kubeschedulerconfig.KubeSchedulerConfiguration{
+				Parallelism:     16,
 				AlgorithmSource: kubeschedulerconfig.SchedulerAlgorithmSource{Provider: &defaultSource},
 				DebuggingConfiguration: componentbaseconfig.DebuggingConfiguration{
 					EnableProfiling:           true,
@@ -700,6 +706,7 @@ profiles:
 			},
 			expectedUsername: "flag",
 			expectedConfig: kubeschedulerconfig.KubeSchedulerConfiguration{
+				Parallelism:     16,
 				AlgorithmSource: kubeschedulerconfig.SchedulerAlgorithmSource{Provider: &defaultSource},
 				DebuggingConfiguration: componentbaseconfig.DebuggingConfiguration{
 					EnableProfiling:           true,
--- a/cmd/kube-scheduler/app/server.go
+++ b/cmd/kube-scheduler/app/server.go
@@ -322,6 +322,7 @@ func Setup(ctx context.Context, opts *options.Options, outOfTreeRegistryOptions 
 		scheduler.WithPodMaxBackoffSeconds(cc.ComponentConfig.PodMaxBackoffSeconds),
 		scheduler.WithPodInitialBackoffSeconds(cc.ComponentConfig.PodInitialBackoffSeconds),
 		scheduler.WithExtenders(cc.ComponentConfig.Extenders...),
+		scheduler.WithParallelism(cc.ComponentConfig.Parallelism),
 	)
 	if err != nil {
 		return nil, nil, err
--- a/pkg/scheduler/apis/config/scheme/scheme_test.go
+++ b/pkg/scheduler/apis/config/scheme/scheme_test.go
@@ -466,6 +466,7 @@ profiles:
 			name:    "v1beta1 in-tree and out-of-tree plugins from internal",
 			version: v1beta1.SchemeGroupVersion,
 			obj: &config.KubeSchedulerConfiguration{
+				Parallelism: 8,
 				Profiles: []config.KubeSchedulerProfile{
 					{
 						PluginConfig: []config.PluginConfig{
@@ -521,6 +522,7 @@ leaderElection:
   resourceNamespace: ""
   retryPeriod: 0s
 metricsBindAddress: ""
+parallelism: 8
 percentageOfNodesToScore: 0
 podInitialBackoffSeconds: 0
 podMaxBackoffSeconds: 0
--- a/pkg/scheduler/apis/config/types.go
+++ b/pkg/scheduler/apis/config/types.go
@@ -55,6 +55,9 @@ const (
 type KubeSchedulerConfiguration struct {
 	metav1.TypeMeta
 
+	// Parallelism defines the amount of parallelism in algorithms for scheduling a Pods. Must be greater than 0. Defaults to 16
+	Parallelism int32
+
 	// AlgorithmSource specifies the scheduler algorithm source.
 	// TODO(#87526): Remove AlgorithmSource from this package
 	// DEPRECATED: AlgorithmSource is removed in the v1beta1 ComponentConfig
--- a/pkg/scheduler/apis/config/v1beta1/defaults.go
+++ b/pkg/scheduler/apis/config/v1beta1/defaults.go
@@ -44,6 +44,11 @@ func addDefaultingFuncs(scheme *runtime.Scheme) error {
 
 // SetDefaults_KubeSchedulerConfiguration sets additional defaults
 func SetDefaults_KubeSchedulerConfiguration(obj *v1beta1.KubeSchedulerConfiguration) {
+
+	if obj.Parallelism == nil {
+		obj.Parallelism = pointer.Int32Ptr(16)
+	}
+
 	if len(obj.Profiles) == 0 {
 		obj.Profiles = append(obj.Profiles, v1beta1.KubeSchedulerProfile{})
 	}
--- a/pkg/scheduler/apis/config/v1beta1/defaults_test.go
+++ b/pkg/scheduler/apis/config/v1beta1/defaults_test.go
@@ -45,6 +45,7 @@ func TestSchedulerDefaults(t *testing.T) {
 			name:   "empty config",
 			config: &v1beta1.KubeSchedulerConfiguration{},
 			expected: &v1beta1.KubeSchedulerConfiguration{
+				Parallelism:        pointer.Int32Ptr(16),
 				HealthzBindAddress: pointer.StringPtr("0.0.0.0:10251"),
 				MetricsBindAddress: pointer.StringPtr("0.0.0.0:10251"),
 				DebuggingConfiguration: componentbaseconfig.DebuggingConfiguration{
@@ -85,6 +86,7 @@ func TestSchedulerDefaults(t *testing.T) {
 				},
 			},
 			expected: &v1beta1.KubeSchedulerConfiguration{
+				Parallelism:        pointer.Int32Ptr(16),
 				HealthzBindAddress: pointer.StringPtr("0.0.0.0:10251"),
 				MetricsBindAddress: pointer.StringPtr("0.0.0.0:10251"),
 				DebuggingConfiguration: componentbaseconfig.DebuggingConfiguration{
@@ -121,6 +123,7 @@ func TestSchedulerDefaults(t *testing.T) {
 		{
 			name: "two profiles",
 			config: &v1beta1.KubeSchedulerConfiguration{
+				Parallelism: pointer.Int32Ptr(16),
 				Profiles: []v1beta1.KubeSchedulerProfile{
 					{
 						PluginConfig: []v1beta1.PluginConfig{
@@ -140,6 +143,7 @@ func TestSchedulerDefaults(t *testing.T) {
 				},
 			},
 			expected: &v1beta1.KubeSchedulerConfiguration{
+				Parallelism:        pointer.Int32Ptr(16),
 				HealthzBindAddress: pointer.StringPtr("0.0.0.0:10251"),
 				MetricsBindAddress: pointer.StringPtr("0.0.0.0:10251"),
 				DebuggingConfiguration: componentbaseconfig.DebuggingConfiguration{
@@ -185,10 +189,12 @@ func TestSchedulerDefaults(t *testing.T) {
 		{
 			name: "metrics and healthz address with no port",
 			config: &v1beta1.KubeSchedulerConfiguration{
+				Parallelism:        pointer.Int32Ptr(16),
 				MetricsBindAddress: pointer.StringPtr("1.2.3.4"),
 				HealthzBindAddress: pointer.StringPtr("1.2.3.4"),
 			},
 			expected: &v1beta1.KubeSchedulerConfiguration{
+				Parallelism:        pointer.Int32Ptr(16),
 				HealthzBindAddress: pointer.StringPtr("1.2.3.4:10251"),
 				MetricsBindAddress: pointer.StringPtr("1.2.3.4:10251"),
 				DebuggingConfiguration: componentbaseconfig.DebuggingConfiguration{
@@ -224,8 +230,44 @@ func TestSchedulerDefaults(t *testing.T) {
 				HealthzBindAddress: pointer.StringPtr(":12345"),
 			},
 			expected: &v1beta1.KubeSchedulerConfiguration{
+				Parallelism:        pointer.Int32Ptr(16),
 				HealthzBindAddress: pointer.StringPtr("0.0.0.0:12345"),
 				MetricsBindAddress: pointer.StringPtr("0.0.0.0:12345"),
+				DebuggingConfiguration: componentbaseconfig.DebuggingConfiguration{
+					EnableProfiling:           &enable,
+					EnableContentionProfiling: &enable,
+				},
+				LeaderElection: componentbaseconfig.LeaderElectionConfiguration{
+					LeaderElect:       pointer.BoolPtr(true),
+					LeaseDuration:     metav1.Duration{Duration: 15 * time.Second},
+					RenewDeadline:     metav1.Duration{Duration: 10 * time.Second},
+					RetryPeriod:       metav1.Duration{Duration: 2 * time.Second},
+					ResourceLock:      "leases",
+					ResourceNamespace: "kube-system",
+					ResourceName:      "kube-scheduler",
+				},
+				ClientConnection: componentbaseconfig.ClientConnectionConfiguration{
+					QPS:         50,
+					Burst:       100,
+					ContentType: "application/vnd.kubernetes.protobuf",
+				},
+				PercentageOfNodesToScore: pointer.Int32Ptr(0),
+				PodInitialBackoffSeconds: pointer.Int64Ptr(1),
+				PodMaxBackoffSeconds:     pointer.Int64Ptr(10),
+				Profiles: []v1beta1.KubeSchedulerProfile{
+					{SchedulerName: pointer.StringPtr("default-scheduler")},
+				},
+			},
+		},
+		{
+			name: "set non default parallelism",
+			config: &v1beta1.KubeSchedulerConfiguration{
+				Parallelism: pointer.Int32Ptr(8),
+			},
+			expected: &v1beta1.KubeSchedulerConfiguration{
+				Parallelism:        pointer.Int32Ptr(8),
+				HealthzBindAddress: pointer.StringPtr("0.0.0.0:10251"),
+				MetricsBindAddress: pointer.StringPtr("0.0.0.0:10251"),
 				DebuggingConfiguration: componentbaseconfig.DebuggingConfiguration{
 					EnableProfiling:           &enable,
 					EnableContentionProfiling: &enable,
--- a/pkg/scheduler/apis/config/v1beta1/zz_generated.conversion.go
+++ b/pkg/scheduler/apis/config/v1beta1/zz_generated.conversion.go
@@ -290,6 +290,9 @@ func Convert_config_InterPodAffinityArgs_To_v1beta1_InterPodAffinityArgs(in *con
 }
 
 func autoConvert_v1beta1_KubeSchedulerConfiguration_To_config_KubeSchedulerConfiguration(in *v1beta1.KubeSchedulerConfiguration, out *config.KubeSchedulerConfiguration, s conversion.Scope) error {
+	if err := metav1.Convert_Pointer_int32_To_int32(&in.Parallelism, &out.Parallelism, s); err != nil {
+		return err
+	}
 	if err := v1alpha1.Convert_v1alpha1_LeaderElectionConfiguration_To_config_LeaderElectionConfiguration(&in.LeaderElection, &out.LeaderElection, s); err != nil {
 		return err
 	}
@@ -330,6 +333,9 @@ func autoConvert_v1beta1_KubeSchedulerConfiguration_To_config_KubeSchedulerConfi
 }
 
 func autoConvert_config_KubeSchedulerConfiguration_To_v1beta1_KubeSchedulerConfiguration(in *config.KubeSchedulerConfiguration, out *v1beta1.KubeSchedulerConfiguration, s conversion.Scope) error {
+	if err := metav1.Convert_int32_To_Pointer_int32(&in.Parallelism, &out.Parallelism, s); err != nil {
+		return err
+	}
 	// WARNING: in.AlgorithmSource requires manual conversion: does not exist in peer-type
 	if err := v1alpha1.Convert_config_LeaderElectionConfiguration_To_v1alpha1_LeaderElectionConfiguration(&in.LeaderElection, &out.LeaderElection, s); err != nil {
 		return err
--- a/pkg/scheduler/apis/config/validation/validation.go
+++ b/pkg/scheduler/apis/config/validation/validation.go
@@ -38,6 +38,10 @@ func ValidateKubeSchedulerConfiguration(cc *config.KubeSchedulerConfiguration) f
 	allErrs = append(allErrs, componentbasevalidation.ValidateLeaderElectionConfiguration(&cc.LeaderElection, field.NewPath("leaderElection"))...)
 
 	profilesPath := field.NewPath("profiles")
+	if cc.Parallelism <= 0 {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("parallelism"), cc.Parallelism, "should be an integer value greater than zero"))
+	}
+
 	if len(cc.Profiles) == 0 {
 		allErrs = append(allErrs, field.Required(profilesPath, ""))
 	} else {
--- a/pkg/scheduler/apis/config/validation/validation_test.go
+++ b/pkg/scheduler/apis/config/validation/validation_test.go
@@ -31,6 +31,7 @@ func TestValidateKubeSchedulerConfiguration(t *testing.T) {
 	podInitialBackoffSeconds := int64(1)
 	podMaxBackoffSeconds := int64(1)
 	validConfig := &config.KubeSchedulerConfiguration{
+		Parallelism:        8,
 		HealthzBindAddress: "0.0.0.0:10254",
 		MetricsBindAddress: "0.0.0.0:10254",
 		ClientConnection: componentbaseconfig.ClientConnectionConfiguration{
@@ -90,6 +91,9 @@ func TestValidateKubeSchedulerConfiguration(t *testing.T) {
 			},
 		},
 	}
+
+	invalidParallelismValue := validConfig.DeepCopy()
+	invalidParallelismValue.Parallelism = 0
 
 	resourceNameNotSet := validConfig.DeepCopy()
 	resourceNameNotSet.LeaderElection.ResourceName = ""
@@ -151,6 +155,10 @@ func TestValidateKubeSchedulerConfiguration(t *testing.T) {
 		"good": {
 			expectedToFail: false,
 			config:         validConfig,
+		},
+		"bad-parallelism-invalid-value": {
+			expectedToFail: true,
+			config:         invalidParallelismValue,
 		},
 		"bad-resource-name-not-set": {
 			expectedToFail: true,
--- a/pkg/scheduler/internal/parallelize/parallelism.go
+++ b/pkg/scheduler/internal/parallelize/parallelism.go
@@ -23,21 +23,31 @@ import (
 	"k8s.io/client-go/util/workqueue"
 )
 
-const parallelism = 16
+var (
+	parallelism = 16
+)
+
+// SetParallelism sets the parallelism for all scheduler algorithms.
+// TODO(#95952): Remove global setter in favor of a struct that holds the configuration.
+func SetParallelism(p int) {
+	parallelism = p
+}
 
 // chunkSizeFor returns a chunk size for the given number of items to use for
 // parallel work. The size aims to produce good CPU utilization.
-func chunkSizeFor(n int) workqueue.Options {
+// returns max(1, min(sqrt(n), n/Parallelism))
+func chunkSizeFor(n int) int {
 	s := int(math.Sqrt(float64(n)))
+
 	if r := n/parallelism + 1; s > r {
 		s = r
 	} else if s < 1 {
 		s = 1
 	}
-	return workqueue.WithChunkSize(s)
+	return s
 }
 
 // Until is a wrapper around workqueue.ParallelizeUntil to use in scheduling algorithms.
 func Until(ctx context.Context, pieces int, doWorkPiece workqueue.DoWorkPieceFunc) {
-	workqueue.ParallelizeUntil(ctx, parallelism, pieces, doWorkPiece, chunkSizeFor(pieces))
+	workqueue.ParallelizeUntil(ctx, parallelism, pieces, doWorkPiece, workqueue.WithChunkSize(chunkSizeFor(pieces)))
 }
--- a/pkg/scheduler/internal/parallelize/parallelism_test.go
+++ b/pkg/scheduler/internal/parallelize/parallelism_test.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package parallelize
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestChunkSize(t *testing.T) {
+	tests := []struct {
+		input      int
+		wantOutput int
+	}{
+		{
+			input:      32,
+			wantOutput: 3,
+		},
+		{
+			input:      16,
+			wantOutput: 2,
+		},
+		{
+			input:      1,
+			wantOutput: 1,
+		},
+		{
+			input:      0,
+			wantOutput: 1,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(fmt.Sprintf("%d", test.input), func(t *testing.T) {
+			if chunkSizeFor(test.input) != test.wantOutput {
+				t.Errorf("Expected: %d, got: %d", test.wantOutput, chunkSizeFor(test.input))
+			}
+		})
+	}
+}
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -42,6 +42,7 @@ import (
 	frameworkplugins "k8s.io/kubernetes/pkg/scheduler/framework/plugins"
 	frameworkruntime "k8s.io/kubernetes/pkg/scheduler/framework/runtime"
 	internalcache "k8s.io/kubernetes/pkg/scheduler/internal/cache"
+	"k8s.io/kubernetes/pkg/scheduler/internal/parallelize"
 	internalqueue "k8s.io/kubernetes/pkg/scheduler/internal/queue"
 	"k8s.io/kubernetes/pkg/scheduler/metrics"
 	"k8s.io/kubernetes/pkg/scheduler/profile"
@@ -106,6 +107,14 @@ type Option func(*schedulerOptions)
 func WithProfiles(p ...schedulerapi.KubeSchedulerProfile) Option {
 	return func(o *schedulerOptions) {
 		o.profiles = p
+	}
+}
+
+// WithParallelism sets the parallelism for all scheduler algorithms. Default is 16.
+// TODO(#95952): Remove global setter in favor of a struct that holds the configuration.
+func WithParallelism(threads int32) Option {
+	return func(o *schedulerOptions) {
+		parallelize.SetParallelism(int(threads))
 	}
 }
 
--- a/staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
+++ b/staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
@@ -44,6 +44,9 @@ const (
 type KubeSchedulerConfiguration struct {
 	metav1.TypeMeta `json:",inline"`
 
+	// Parallelism defines the amount of parallelism in algorithms for scheduling a Pods. Must be greater than 0. Defaults to 16
+	Parallelism *int32 `json:"parallelism,omitempty"`
+
 	// LeaderElection defines the configuration of leader election client.
 	LeaderElection componentbaseconfigv1alpha1.LeaderElectionConfiguration `json:"leaderElection"`
 
--- a/pkg/scheduler/BUILD
+++ b/pkg/scheduler/BUILD
@@ -25,6 +25,7 @@ go_library(
         "//pkg/scheduler/framework/runtime:go_default_library",
         "//pkg/scheduler/internal/cache:go_default_library",
         "//pkg/scheduler/internal/cache/debugger:go_default_library",
+        "//pkg/scheduler/internal/parallelize:go_default_library",
         "//pkg/scheduler/internal/queue:go_default_library",
         "//pkg/scheduler/metrics:go_default_library",
         "//pkg/scheduler/profile:go_default_library",
--- a/pkg/scheduler/internal/parallelize/BUILD
+++ b/pkg/scheduler/internal/parallelize/BUILD
@@ -27,6 +27,9 @@ filegroup(
 
 go_test(
     name = "go_default_test",
-    srcs = ["error_channel_test.go"],
+    srcs = [
+        "error_channel_test.go",
+        "parallelism_test.go",
+    ],
     embed = [":go_default_library"],
 )
--- a/staging/src/k8s.io/kube-scheduler/config/v1beta1/zz_generated.deepcopy.go
+++ b/staging/src/k8s.io/kube-scheduler/config/v1beta1/zz_generated.deepcopy.go
@@ -87,6 +87,11 @@ func (in *InterPodAffinityArgs) DeepCopyObject() runtime.Object {
 func (in *KubeSchedulerConfiguration) DeepCopyInto(out *KubeSchedulerConfiguration) {
 	*out = *in
 	out.TypeMeta = in.TypeMeta
+	if in.Parallelism != nil {
+		in, out := &in.Parallelism, &out.Parallelism
+		*out = new(int32)
+		**out = **in
+	}
 	in.LeaderElection.DeepCopyInto(&out.LeaderElection)
 	out.ClientConnection = in.ClientConnection
 	if in.HealthzBindAddress != nil {


pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -21,15 +21,14 @@ import (
 	"math"
 
 	"k8s.io/client-go/util/workqueue"
+	kubeschedulerconfig "k8s.io/kubernetes/pkg/scheduler/apis/config"
```
chendave, line 24: The alias "kubeschedulerconfig" is not needed.

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -21,15 +21,14 @@ import (
...
 )
 
-const parallelism = 16
-
 // chunkSizeFor returns a chunk size for the given number of items to use for
 // parallel work. The size aims to produce good CPU utilization.
-func chunkSizeFor(n int) workqueue.Options {
-	s := int(math.Sqrt(float64(n)))
-	if r := n/parallelism + 1; s > r {
+func chunkSizeFor(n int, in *KubeSchedulerConfiguration) workqueue.Options {
```
chendave, line 29: I guess you mean `config.KubeSchedulerConfiguration`?

hprateek43: I did, the problem is that the `config.KubeSchedulerConfiguration` is a struct, and I am unable to get the currently loaded configuration of the scheduler via a function call. I see that its propagated via function variables

pkg/scheduler/apis/config/types.go
```diff
@@ -55,6 +55,9 @@ const (
 type KubeSchedulerConfiguration struct {
 	metav1.TypeMeta
 
+	// CPU threads
+	CPUThreads int
```
adtac, line 59: There isn't a one-to-one correspondence between real CPU threads and the parallelism we're talking about here. Goroutines are multiplexed over OS threads/processes that is configurable with `GOMAXPROCS` (which defaults to the number of cores available).

We should probably define this `Parallelism int` to not mislead users. As an example to show the difference between the different things involved here, the user can set `GOMAXPROCS=4` and `Parallelism = 16` in a machine with 8 physical CPUs -- the actual number of CPU threads in this example is 4.

staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
```diff
@@ -44,6 +44,9 @@ const (
 type KubeSchedulerConfiguration struct {
 	metav1.TypeMeta `json:",inline"`
 
+	// CPU threads
+	CPUThreads int `json:"cpuThreads"`
```
adtac, line 48: This should be a `*int32` to make the parameter optional.

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -38,6 +37,6 @@ func chunkSizeFor(n int) workqueue.Options {
 }
 
 // Until is a wrapper around workqueue.ParallelizeUntil to use in scheduling algorithms.
-func Until(ctx context.Context, pieces int, doWorkPiece workqueue.DoWorkPieceFunc) {
-	workqueue.ParallelizeUntil(ctx, parallelism, pieces, doWorkPiece, chunkSizeFor(pieces))
+func Until(ctx context.Context, pieces int, doWorkPiece workqueue.DoWorkPieceFunc,in *KubeSchedulerConfiguration) {
+	workqueue.ParallelizeUntil(ctx, in.CPUThreads, pieces, doWorkPiece, chunkSizeFor(pieces))
```
adtac, line 41: did you forget to pass `in` as an argument to chunkSizeFor here?

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -21,15 +21,14 @@ import (
...
 
-const parallelism = 16
-
 // chunkSizeFor returns a chunk size for the given number of items to use for
 // parallel work. The size aims to produce good CPU utilization.
-func chunkSizeFor(n int) workqueue.Options {
-	s := int(math.Sqrt(float64(n)))
-	if r := n/parallelism + 1; s > r {
+func chunkSizeFor(n int, in *KubeSchedulerConfiguration) workqueue.Options {
+	s := int(math.Sqrt(float64(
```
adtac, line 30: ... looks like the rest of this line got clipped?

pkg/scheduler/apis/config/v1beta1/zz_generated.conversion.go
```diff
@@ -290,6 +290,7 @@ func Convert_config_InterPodAffinityArgs_To_v1beta1_InterPodAffinityArgs(in *con
 }
 
 func autoConvert_v1beta1_KubeSchedulerConfiguration_To_config_KubeSchedulerConfiguration(in *v1beta1.KubeSchedulerConfiguration, out *config.KubeSchedulerConfiguration, s conversion.Scope) error {
+	out.CPUThreads = in.CPUThreads
```
adtac, line 293: once you make the int -> `*int32` change, please re-codegen this file

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -21,15 +21,14 @@ import (
 	"math"
 
 	"k8s.io/client-go/util/workqueue"
+	kubeschedulerconfig "k8s.io/kubernetes/pkg/scheduler/apis/config"
 )
 
-const parallelism = 16
```
adtac, line 26: We still need a default of 16. See `pkg/scheduler/apis/config/v1beta1/defaults.go` to do that.

staging/src/k8s.io/code-generator/_examples/apiserver/openapi/zz_generated.openapi.go
```diff
@@ -30,67 +30,58 @@ import (
...
-		"k8s.io/code-generator/_examples/apiserver/apis/example/v1.TestTypeList":       schema_apiserver_apis_example_v1_TestTypeList(ref),
-		"k8s.io/code-generator/_examples/apiserver/apis/example/v1.TestTypeStatus":     schema_apiserver_apis_example_v1_TestTypeStatus(ref),
-		"k8s.io/code-generator/_examples/apiserver/apis/example2/v1.TestType":          schema_apiserver_apis_example2_v1_TestType(ref),
-		"k8s.io/code-generator/_examples/apiserver/apis/example2/v1.TestTypeList":      schema_apiserver_apis_example2_v1_TestTypeList(ref),
-		"k8s.io/code-generator/_examples/apiserver/apis/example2/v1.TestTypeStatus":    schema_apiserver_apis_example2_v1_TestTypeStatus(ref),
-		"k8s.io/code-generator/_examples/apiserver/apis/example3.io/v1.TestType":       schema_apiserver_apis_example3io_v1_TestType(ref),
-		"k8s.io/code-generator/_examples/apiserver/apis/example3.io/v1.TestTypeList":   schema_apiserver_apis_example3io_v1_TestTypeList(ref),
-		"k8s.io/code-generator/_examples/apiserver/apis/example3.io/v1.TestTypeStatus": schema_apiserver_apis_example3io_v1_TestTypeStatus(ref),
+		"k8s.io/apimachinery/pkg/apis/meta/v1.APIGroup":                  schema_pkg_apis_meta_v1_APIGroup(ref),
```
adtac, line 33: I'm not really sure what the changes to this file are about...

pkg/scheduler/apis/config/types.go
hprateek43: I agree, but a universal solution may require more simulations. For now, I have updated the PR, there were some hunks missing. I was testing some changes and this got converted from draft. Right now the pressing challenge is to get the currently loaded configuration of the scheduler on parallelism as I see that the config is propagated via arguments.

pkg/scheduler/apis/config/v1beta1/defaults.go
```diff
@@ -44,6 +44,9 @@ func addDefaultingFuncs(scheme *runtime.Scheme) error {
 
 // SetDefaults_KubeSchedulerConfiguration sets additional defaults
 func SetDefaults_KubeSchedulerConfiguration(obj *v1beta1.KubeSchedulerConfiguration) {
+
+	obj.CPUThreads = 16
```
adtac, line 48: do this only if it the pointer is nil

pkg/scheduler/apis/config/types.go
adtac: No, I meant parallelism within the scheduler, not parallelism elsewhere i.e., change the parameter's name from `CPUThreads` to `Parallelism` so that we don't mislead users of `KubeSchedulerConfiguration`.

alculquicondor: +1 to `Parallelism` or `Routines`, but I prefer the first one. Note that this name is part of an API, and as such, it can be changed easily. Also a good comment is worth. We don't want people to think this is about several Pods being scheduled at the same time.

Maybe even `AlgorithmsParallelism` is more appropriate, but I'm not convinced.

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -21,15 +21,15 @@ import (
...
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/kubernetes/pkg/scheduler/apis/config"
 )
 
-const parallelism = 16
-
 // chunkSizeFor returns a chunk size for the given number of items to use for
 // parallel work. The size aims to produce good CPU utilization.
-func chunkSizeFor(n int) workqueue.Options {
+func chunkSizeFor(n int, in *config.KubeSchedulerConfiguration) workqueue.Options {
```
alculquicondor, line 29: receive the int instead. There's no really a reason for this package to depend on configuration for one value.

alculquicondor (commented):
I assume this is a WIP, given that the callers haven't of the function haven't been updated.

pkg/scheduler/internal/parallelize/parallelism.go
alculquicondor: Alternatively, you could expose a function `SetParallelism` that is called once once the configuration has been loaded.

pkg/scheduler/apis/config/v1beta1/defaults.go
soulxu: should we check ensure CPUThreads > 0 at pkg/scheduler/apis/config/validation.ValidateKubeSchedulerConfiguration?

pkg/scheduler/apis/config/types.go
adtac: Actually, upon further thought, I think `Concurrency` (or `AlgorithmsConcurrency`) might be better since goroutines are only concurrent, not necessarily parallel.

alculquicondor: That's a bit nit-picky on implementation details.

In their most common interpretation, Concurrency is about dealing with lots of things at once (like multiple unrelated requests, or multiple Pods in k8s world), whereas parallelism is about doing lots of things at once, like subtasks (so more suited to talk about algorithms).

alculquicondor:
@hprateek43 any progress on this?

hprateek43:
Im working on this. The scheduler does not have a function that can provide the currently active config, and the config is passed down via a variable. Im working on propogating this to the parallism function calls.

alculquicondor:
Don't propagate the config, just the values that are needed.

adtac:
@hprateek43 https://github.com/kubernetes/kubernetes/pull/95157 might be helpful in case you want some documentation on what extra changes you need make. It's geared towards scheduler plugins, but most points there should apply to root-level KubeSchedulerConfiguration parameters.

cmd/kube-scheduler/app/options/options_test.go
```diff
@@ -220,6 +220,7 @@ profiles:
 	if err := ioutil.WriteFile(multiProfilesConfig, []byte(fmt.Sprintf(`
 apiVersion: kubescheduler.config.k8s.io/v1beta1
 kind: KubeSchedulerConfiguration
+cpuThreads: 16
```
alculquicondor, line 223: you can leave this out to the defaulting.

cmd/kube-scheduler/app/server.go
```diff
@@ -319,6 +318,7 @@ func Setup(ctx context.Context, opts *options.Options, outOfTreeRegistryOptions
 		cc.PodInformer,
 		recorderFactory,
 		ctx.Done(),
+		scheduler.SetParallelism(cc.ComponentConfig.CPUThreads),
```
alculquicondor, line 321: WithParallelism

pkg/scheduler/apis/config/types.go
alculquicondor: Let's make it `Parallelism`

pkg/scheduler/apis/config/v1beta1/defaults.go
alculquicondor: Yes to only if pointer is nil and yes to validation.

pkg/scheduler/apis/config/validation/validation.go
```diff
@@ -38,6 +38,9 @@ func ValidateKubeSchedulerConfiguration(cc *config.KubeSchedulerConfiguration) f
 	allErrs = append(allErrs, componentbasevalidation.ValidateLeaderElectionConfiguration(&cc.LeaderElection, field.NewPath("leaderElection"))...)
 
 	profilesPath := field.NewPath("profiles")
+	if !(cc.CPUThreads > 0) {
```
alculquicondor, line 41: use `cc.Parallelism <= 0`

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -39,5 +41,5 @@ func chunkSizeFor(n int) workqueue.Options {
 
 // Until is a wrapper around workqueue.ParallelizeUntil to use in scheduling algorithms.
 func Until(ctx context.Context, pieces int, doWorkPiece workqueue.DoWorkPieceFunc) {
```
alculquicondor, line 43: better pass it as a parameter

hprateek43: Do you mean passing `Parallelism` as an argument?

alculquicondor: yes

api/api-rules/violation_exceptions.list
```diff
@@ -597,6 +597,7 @@ API rule violation: names_match,k8s.io/kube-controller-manager/config/v1alpha1,V
 API rule violation: names_match,k8s.io/kube-controller-manager/config/v1alpha1,VolumeConfiguration,PersistentVolumeRecyclerConfiguration
 API rule violation: names_match,k8s.io/kube-proxy/config/v1alpha1,KubeProxyConfiguration,IPTables
 API rule violation: names_match,k8s.io/kube-scheduler/config/v1,LegacyExtender,EnableHTTPS
+API rule violation: names_match,k8s.io/kube-scheduler/config/v1beta1,KubeSchedulerConfiguration,Parallelism
```
adtac, line 600: why are we doing this?

pkg/scheduler/apis/config/scheme/scheme_test.go
```diff
@@ -390,6 +391,7 @@ clientConnection:
   contentType: ""
   kubeconfig: ""
   qps: 0
+cpuThreads: 16
```
adtac, line 394: test something other than 16 (to make sure we can override the default)

pkg/scheduler/apis/config/types.go
```diff
@@ -55,6 +55,9 @@ const (
 type KubeSchedulerConfiguration struct {
 	metav1.TypeMeta
 
+	// CPU threads
```
adtac, line 58: "Parallelism defines the amount of parallelism that should be used in scheduler algorithms and some plugins."

pkg/scheduler/apis/config/v1beta1/conversion_test.go
```diff
@@ -34,8 +34,8 @@ func TestV1beta1ToConfigKubeSchedulerConfigurationConversion(t *testing.T) {
 	}{
 		{
 			name:   "default conversion v1beta1 to config",
-			config: v1beta1.KubeSchedulerConfiguration{},
-			want:   config.KubeSchedulerConfiguration{AlgorithmSource: config.SchedulerAlgorithmSource{Provider: pointer.StringPtr(v1beta1.SchedulerDefaultProviderName)}},
+			config: v1beta1.KubeSchedulerConfiguration{Parallelism: 16},
+			want:   config.KubeSchedulerConfiguration{Parallelism: 16, AlgorithmSource: config.SchedulerAlgorithmSource{Provider: pointer.StringPtr(v1beta1.SchedulerDefaultProviderName)}},
```
adtac, line 38: this is getting unreadable :) can you break up the struct into multiple lines?

staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
```diff
@@ -44,6 +44,9 @@ const (
 type KubeSchedulerConfiguration struct {
 	metav1.TypeMeta `json:",inline"`
 
+	// CPU threads
```
adtac, line 47: Same as the other field's description.

pkg/scheduler/apis/config/validation/validation.go
```diff
@@ -38,6 +38,9 @@ func ValidateKubeSchedulerConfiguration(cc *config.KubeSchedulerConfiguration) f
 	allErrs = append(allErrs, componentbasevalidation.ValidateLeaderElectionConfiguration(&cc.LeaderElection, field.NewPath("leaderElection"))...)
 
 	profilesPath := field.NewPath("profiles")
+	if cc.Parallelism <= 0 {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("Parallelism"), cc.Parallelism, ""))
```
adtac, line 42: add a more descriptive error message please

pkg/scheduler/apis/config/v1beta1/defaults_test.go
```diff
@@ -224,6 +230,7 @@ func TestSchedulerDefaults(t *testing.T) {
 				HealthzBindAddress: pointer.StringPtr(":12345"),
 			},
 			expected: &v1beta1.KubeSchedulerConfiguration{
+				Parallelism:        16,
```
adtac, line 233: add a test that sets a non-default parallelism?

pkg/scheduler/apis/config/types.go
```diff
@@ -55,6 +55,9 @@ const (
...
+	Parallelism int
```
adtac, line 59: int32, btw

api/api-rules/violation_exceptions.list
hprateek43: Parallelism was reported as an API violation, so had to run an update on exceptions

staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
```diff
@@ -44,6 +44,9 @@ const (
 type KubeSchedulerConfiguration struct {
 	metav1.TypeMeta `json:",inline"`
 
+	// Parallelism defines the amount of parallelism that should be used in scheduler algorithms and some plugins.
+	Parallelism int32 `json:"cpuThreads"`
```
alculquicondor, line 48: The violation is because this should be `json:"parallelism"`. Please fix.

pkg/scheduler/scheduler.go
```diff
@@ -96,6 +97,8 @@ type schedulerOptions struct {
 	profiles                   []schedulerapi.KubeSchedulerProfile
 	extenders                  []schedulerapi.Extender
 	frameworkCapturer          FrameworkCapturer
+	// Set the parallel thread count for parallism package
+	cputhreadcount int32
```
alculquicondor, line 101: let's use the same variable name `parallelism` for consistency

alculquicondor:
/retitle [WIP] Added config parameter for CPU threads

Remove WIP when tests are passing

pkg/scheduler/scheduler.go
hprateek43: I think we can leave the `CPUThreadCount` in the `parallelize` package for better understanding and readability, if we want this parallelism to have a  correlation with CPU Threads.

hprateek43:
@adtac @alculquicondor Can you review once, the tests seem to go through, but some refactoring or restructuring may be required. I have tried to capture all major comments.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -52,10 +52,17 @@ import (
...
 	internalqueue "k8s.io/kubernetes/pkg/scheduler/internal/queue"
 	st "k8s.io/kubernetes/pkg/scheduler/testing"
 )
 
+func init() {
+	// set the default parallelize.Parallelism to 16 as scheduler defaults are not
+	// loaded in with generic scheduler
+	parallelize.Parallelism = 16
+}
+
```
hprateek43, line 65: I decided to add this in the `init` as there was no simpler way to configure this for all tests, some rely on `genericScheduler` and other directly invoke the preemption and other business logics, this way have a common solution for all tests in a package, or we can do this at a per test level

cmd/kube-scheduler/app/options/options_test.go
```diff
@@ -220,6 +220,7 @@ profiles:
 	if err := ioutil.WriteFile(multiProfilesConfig, []byte(fmt.Sprintf(`
 apiVersion: kubescheduler.config.k8s.io/v1beta1
 kind: KubeSchedulerConfiguration
+parallelism: 16
```
alculquicondor, line 223: remove this line. This test is only for things exposed through command line flags.

pkg/scheduler/apis/config/scheme/scheme_test.go
```diff
@@ -483,6 +484,7 @@ leaderElection:
   resourceName: ""
   resourceNamespace: ""
   retryPeriod: 0s
+parallelism: 16
```
alculquicondor, line 487: put a different number

pkg/scheduler/apis/config/v1beta1/conversion_test.go
```diff
@@ -34,8 +34,10 @@ func TestV1beta1ToConfigKubeSchedulerConfigurationConversion(t *testing.T) {
 	}{
 		{
 			name:   "default conversion v1beta1 to config",
-			config: v1beta1.KubeSchedulerConfiguration{},
-			want:   config.KubeSchedulerConfiguration{AlgorithmSource: config.SchedulerAlgorithmSource{Provider: pointer.StringPtr(v1beta1.SchedulerDefaultProviderName)}},
+			config: v1beta1.KubeSchedulerConfiguration{Parallelism: 16},
```
alculquicondor, line 37: leave this empty

staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
```diff
@@ -44,6 +44,9 @@ const (
 type KubeSchedulerConfiguration struct {
 	metav1.TypeMeta `json:",inline"`
 
+	// Parallelism defines the amount of parallelism that should be used in scheduler algorithms and some plugins.
+	Parallelism int32 `json:"parallelism"`
```
alculquicondor, line 48: Make this a pointer and add `omitempty` to the serialization

pkg/scheduler/apis/config/v1beta1/defaults.go
```diff
@@ -44,6 +44,11 @@ func addDefaultingFuncs(scheme *runtime.Scheme) error {
 
 // SetDefaults_KubeSchedulerConfiguration sets additional defaults
 func SetDefaults_KubeSchedulerConfiguration(obj *v1beta1.KubeSchedulerConfiguration) {
+
+	if obj.Parallelism == 0 {
```
alculquicondor, line 48: `== nil`

pkg/scheduler/apis/config/v1beta1/defaults_test.go
alculquicondor: +1

pkg/scheduler/apis/config/validation/validation.go
```diff
@@ -38,6 +38,9 @@ func ValidateKubeSchedulerConfiguration(cc *config.KubeSchedulerConfiguration) f
 	allErrs = append(allErrs, componentbasevalidation.ValidateLeaderElectionConfiguration(&cc.LeaderElection, field.NewPath("leaderElection"))...)
 
 	profilesPath := field.NewPath("profiles")
+	if cc.Parallelism <= 0 {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("Parallelism"), cc.Parallelism, "Parallelism should be a greater than zero integer value"))
```
alculquicondor, line 42:
```suggestion
		allErrs = append(allErrs, field.Invalid(field.NewPath("parallelism"), cc.Parallelism, "should be an integer value greater than zero"))
```

pkg/scheduler/apis/config/validation/validation_test.go
```diff
@@ -31,6 +31,7 @@ func TestValidateKubeSchedulerConfiguration(t *testing.T) {
 	podInitialBackoffSeconds := int64(1)
 	podMaxBackoffSeconds := int64(1)
 	validConfig := &config.KubeSchedulerConfiguration{
+		Parallelism:        16,
```
alculquicondor, line 34: Add a test for an invalid parallelism

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -570,6 +570,10 @@ func NewGenericScheduler(
 	nodeInfoSnapshot *internalcache.Snapshot,
 	extenders []framework.Extender,
 	percentageOfNodesToScore int32) ScheduleAlgorithm {
+
+	// set the default parallelize.Parallelism to 16 as scheduler defaults are not
+	// loaded in with generic scheduler
+	parallelize.Parallelism = 16
```
alculquicondor, line 576: There is nothing changing this value when parallelism is actually set.

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -23,13 +23,15 @@ import (
 	"k8s.io/client-go/util/workqueue"
 )
 
-const parallelism = 16
+// Parallelism is used the store value from scheduler config
+var Parallelism int
```
alculquicondor, line 27: var Parallelism = 16

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
alculquicondor: Please remove this init. See comment below.

pkg/scheduler/framework/plugins/podtopologyspread/filtering_test.go
```diff
@@ -37,6 +37,12 @@ import (
 	"k8s.io/utils/pointer"
 )
 
+func init() {
```
alculquicondor, line 40: Remove this one too

pkg/scheduler/framework/runtime/framework_test.go
```diff
@@ -34,9 +34,16 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/kubernetes/pkg/scheduler/apis/config"
 	"k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
+	"k8s.io/kubernetes/pkg/scheduler/internal/parallelize"
 	"k8s.io/kubernetes/pkg/scheduler/metrics"
 )
 
+func init() {
```
alculquicondor, line 41: Remove this one

pkg/scheduler/scheduler.go
```diff
@@ -109,6 +112,14 @@ func WithProfiles(p ...schedulerapi.KubeSchedulerProfile) Option {
 	}
 }
 
+// WithParallelism sets CPU Threads for Scheduler parallelize package. By default it is 16
+func WithParallelism(threads int32) Option {
+	return func(o *schedulerOptions) {
+		o.parallelism = threads
+		parallelize.Parallelism = int(threads)
```
alculquicondor, line 119: If you are already going to set `parallelize.Parallelism`, there is no need to store `o.parallelism`.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -86,6 +93,8 @@ var (
 )
 
 func TestPostFilter(t *testing.T) {
+	// set the default parallelize.Parallelism to 16 as scheduler defaults are not
```
alculquicondor, line 96: Remove this comment.

staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
hprateek43: So in `KubeSchedulerConfig` can i assign thevalue of parallelism like
```
&config.KubeSchedulerConfiguration{
		Parallelism:        func(val int32) *int32 { return &val }(10),
...
```

as it a pointer, or some other way is suggested ?

hprateek43:
@alculquicondor Can you review now ?

pkg/scheduler/apis/config/types.go
```diff
@@ -55,8 +55,8 @@ const (
 type KubeSchedulerConfiguration struct {
 	metav1.TypeMeta
 
-	// CPU threads
-	CPUThreads int
+	// Parallelism defines the amount of parallelism that should be used in scheduler algorithms and some plugins.
```
alculquicondor, line 58: So this one shouldn't be a pointer (only the one in staging)

pkg/scheduler/apis/config/v1beta1/conversion_test.go
```diff
@@ -35,7 +35,8 @@ func TestV1beta1ToConfigKubeSchedulerConfigurationConversion(t *testing.T) {
 		{
 			name:   "default conversion v1beta1 to config",
 			config: v1beta1.KubeSchedulerConfiguration{},
-			want:   config.KubeSchedulerConfiguration{AlgorithmSource: config.SchedulerAlgorithmSource{Provider: pointer.StringPtr(v1beta1.SchedulerDefaultProviderName)}},
+			want: config.KubeSchedulerConfiguration{
```
alculquicondor, line 38: revert this one to avoid making unrelated changes in this PR

pkg/scheduler/apis/config/validation/validation.go
```diff
@@ -38,6 +38,9 @@ func ValidateKubeSchedulerConfiguration(cc *config.KubeSchedulerConfiguration) f
 	allErrs = append(allErrs, componentbasevalidation.ValidateLeaderElectionConfiguration(&cc.LeaderElection, field.NewPath("leaderElection"))...)
 
 	profilesPath := field.NewPath("profiles")
+	if *cc.Parallelism <= 0 {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("Parallelism"), cc.Parallelism, "should be an integer value greater than zero"))
```
alculquicondor, line 42: `NewPath("parallelism")`. The errors should use the serialized (lowercase) form

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
alculquicondor: Still remove

pkg/scheduler/framework/plugins/podtopologyspread/filtering_test.go
alculquicondor: Please do this

pkg/scheduler/framework/plugins/interpodaffinity/filtering_test.go
```diff
@@ -22,7 +22,7 @@ import (
 	"strings"
 	"testing"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
alculquicondor, line 25: Revert unnecessary change

pkg/scheduler/framework/runtime/framework_test.go
```diff
@@ -1809,6 +1809,7 @@ func TestRunBindPlugins(t *testing.T) {
 			wantStatus: v1alpha1.Error,
 		},
 	}
+
```
alculquicondor, line 1812: revert

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -23,13 +23,15 @@ import (
...
+// Parallelism is used the store value from scheduler config
+var Parallelism = 16
 
 // chunkSizeFor returns a chunk size for the given number of items to use for
 // parallel work. The size aims to produce good CPU utilization.
 func chunkSizeFor(n int) workqueue.Options {
 	s := int(math.Sqrt(float64(n)))
-	if r := n/parallelism + 1; s > r {
+
+	if r := Parallelism + 1; s > r {
```
alculquicondor, line 34: `n/Parallelism`.

It looks like we need a unit test for this function. I don't normally ask for unit tests for non-exported functions, but there is no other way to test the intended chunk size in this case.

pkg/scheduler/scheduler.go
```diff
@@ -96,6 +97,8 @@ type schedulerOptions struct {
 	profiles                   []schedulerapi.KubeSchedulerProfile
 	extenders                  []schedulerapi.Extender
 	frameworkCapturer          FrameworkCapturer
+	// Set the parallel thread count for parallism package
```
alculquicondor, line 100: remove this field

pkg/scheduler/scheduler.go
```diff
@@ -109,6 +112,13 @@ func WithProfiles(p ...schedulerapi.KubeSchedulerProfile) Option {
 	}
 }
 
+// WithParallelism sets CPU Threads for Scheduler parallelize package. By default it is 16
```
alculquicondor, line 115: `WithParallelism sets the parallelism for all scheduler algorithms. Default is 16.`

staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
```diff
@@ -44,6 +44,9 @@ const (
 type KubeSchedulerConfiguration struct {
 	metav1.TypeMeta `json:",inline"`
 
+	// Parallelism defines the amount of parallelism that should be used in scheduler algorithms and some plugins.
```
alculquicondor, line 47: I would say:

`Parallelism defines the amount of parallelism in algorithms for scheduling a Pod`

This makes it clear that we are NOT scheduling multiple Pods at a time and plugins are internal details that users of this field don't need to know.

pkg/scheduler/apis/config/validation/validation_test.go
alculquicondor: Please do this

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -23,13 +23,15 @@ import (
 	"k8s.io/client-go/util/workqueue"
 )
 
-const parallelism = 16
+// Parallelism is used the store value from scheduler config
+var Parallelism = 16
 
 // chunkSizeFor returns a chunk size for the given number of items to use for
 // parallel work. The size aims to produce good CPU utilization.
```
alculquicondor, line 30: Also say:

returns max(1, min(sqrt(n), n/Parallelism))

pkg/scheduler/framework/plugins/interpodaffinity/filtering_test.go
hprateek43: This seems to be done by the VS Code go plugin, let me revert this

pkg/scheduler/framework/runtime/framework_test.go
hprateek43: This seems to be done by the VS Code go plugin, let me revert this

pkg/scheduler/internal/parallelize/parallelism.go
hprateek43: Do you want me to convert it to a named return or add a new function ?

alculquicondor: no, just add a comment to the function header.

pkg/scheduler/scheduler.go
hprateek43: You want me to remove the `parallelism` field ? Would that disrupt the rest of the logic ?

pkg/scheduler/internal/parallelize/parallelism.go
hprateek43: fun can be compared to `nil` only in go, so how should we plan the UT for this ? The chunk size directly creates the workflow, rather than returning the int. Should I refactor ?

alculquicondor: refactoring is fine

pkg/scheduler/apis/config/scheme/scheme_test.go
```diff
@@ -397,6 +397,7 @@ profiles:
 }
 
 func TestCodecsEncodePluginConfig(t *testing.T) {
+	parallelism := int32(8)
```
adtac, line 400:
```go
func i32ptr(n int32) *int32 {
  return &n
}
```

is a more common pattern to create a pointer to a int literal

pkg/scheduler/apis/config/types.go
```diff
@@ -55,6 +55,9 @@ const (
 type KubeSchedulerConfiguration struct {
 	metav1.TypeMeta
 
+	// Parallelism defines the amount of parallelism in algorithms for scheduling a Pod
```
adtac, line 58:
```suggestion
	// Parallelism defines the amount of parallelism in algorithms for scheduling a Pod.
```

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -23,21 +23,24 @@ import (
 	"k8s.io/client-go/util/workqueue"
 )
 
-const parallelism = 16
+// Parallelism is used the store value from scheduler config
```
adtac, line 26:
```suggestion
// Parallelism is amount of parallelism that should be used in scheduler algorithms when scheduling a pod.
```

adtac: Also, worth adding a quick comment about how this should not be modified after the scheduler has booted up?

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -23,21 +23,24 @@ import (
...
+var Parallelism = 16
```
adtac, line 27: add int32 to make it explicitly an int32. are we setting `= 16` here for tests and other places that use parallelism without going through the whole scheduler bootstrap? defaults.go already sets a default, so I'm slightly hesitant to have the same number in two different places.

hprateek43: We have added it here as the generic scheduler does not go through the bootstrapping

hprateek43: Also why do we want to set `int32` specifically when the `n/Parallelism + 1; s > r ` computation requires an `int` ?

pkg/scheduler/apis/config/scheme/scheme_test.go
alculquicondor: Actually, line 411 should just use `pointer.Int32ptr`

pkg/scheduler/internal/parallelize/parallelism.go
alculquicondor: I actually prefer just `int`. `int32` is only relevant in API configuration.

Also, note that other binaries (cluster autoscaler for example) don't bootstrap a scheduler, only a framework. I think this is fine for now.

hprateek43:
@alculquicondor Can we go for a merge now  ?

pkg/scheduler/apis/config/scheme/scheme_test.go
alculquicondor: Please do this

alculquicondor (commented):
You can squash as you do the latest fix.

staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
alculquicondor: Please update this

alculquicondor (commented):
/approve

/assign @liggitt 
for api-review

staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
alculquicondor: Still not updated

alculquicondor:
What happened here?

/remove-area apiserver cloudprovider code-generation comformance dependency e2e-test-framework kubectl kubelet provider/gcp release-eng test

alculquicondor:
/remove-sig api-machinery apps architecture auth cli cloud-provider cluster-lifecycle instrumentation network node release scalability storage testing windows

hprateek43:
> What happened here?
> 
> /remove-area apiserver cloudprovider code-generation comformance dependency e2e-test-framework kubectl kubelet provider/gcp release-eng test

During rebase, I faced some issue. Partly due to WSL crash and partly due to the rebase identifying the entire directory as changed

alculquicondor:
/hold

What happened to the changes in the file `pkg/scheduler/config/apis/v1beta1/types.go` and `pkg/scheduler/scheduler.go`. The configuration parameter is being lost

pkg/scheduler/apis/config/scheme/scheme_test.go
```diff
@@ -424,6 +425,7 @@ leaderElection:
   resourceName: ""
   resourceNamespace: ""
   retryPeriod: 0s
+parallelism: 8
```
alculquicondor, line 428: Sorry for the back and forth, but sometimes my attention is caught by other files.

This test is named "TestCodecsEncodePluginConfig" so we shouldn't be testing Parallelism here.
But rather than adding another test case, just remove the change in this file. We already have a test in defaults_test.go

The tests in this file are really about plugins, because they have a complex setup in the serializer infrastructure. The rest of defaults should be tested where they are (defaults.go)

hprateek43: The tests fail if I don't make this change. The default value propagated was 16. I swapped the value to 8 to test with a non default value

alculquicondor: true. Just leave it in 16 then (that is, remove the other line 351)

hprateek43: I have updated this as per recommendation. I thought we were planning to test the non default parallelism here as well

alculquicondor: No, for the reasons above

/hold cancel

alculquicondor:
Now ready for api-review @liggitt

alculquicondor:
/hold
sorry, still not ready. It looks like other files that were already reviewed were lost during rebasing

alculquicondor (commented):
/lgtm
/hold cancel

pkg/scheduler/internal/parallelize/parallelism.go
liggitt: still looks outstanding

pkg/scheduler/internal/parallelize/parallelism.go
liggitt, line 27: I expected this to go away now that there was a config option

pkg/scheduler/scheduler.go
```diff
@@ -109,6 +110,13 @@ func WithProfiles(p ...schedulerapi.KubeSchedulerProfile) Option {
 	}
 }
 
+// WithParallelism sets the parallelism for all scheduler algorithms. Default is 16.
+func WithParallelism(threads int32) Option {
+	return func(o *schedulerOptions) {
+		parallelize.Parallelism = int(threads)
```
liggitt, line 116: setting global vars based on config is not what I expected

staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
```diff
@@ -44,6 +44,9 @@ const (
 type KubeSchedulerConfiguration struct {
 	metav1.TypeMeta `json:",inline"`
 
+	// Parallelism defines the amount of parallelism in algorithms for scheduling a Pods.
```
liggitt, line 47: indicate the default if unspecified, and the allowed min (and max, if we set a max)

pkg/scheduler/apis/config/validation/validation.go
```diff
@@ -38,6 +38,10 @@ func ValidateKubeSchedulerConfiguration(cc *config.KubeSchedulerConfiguration) f
 	allErrs = append(allErrs, componentbasevalidation.ValidateLeaderElectionConfiguration(&cc.LeaderElection, field.NewPath("leaderElection"))...)
 
 	profilesPath := field.NewPath("profiles")
+	if cc.Parallelism <= 0 {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("parallelism"), cc.Parallelism, "should be an integer value greater than zero"))
+	}
```
liggitt, line 43: do we need a maximum? any other considerations we should requirement/document or require (power-of-two, etc)?

hprateek43: Do we want to enforce them , or just add them to documentation as a reference ?

staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
hprateek43:
```
Parallelism defines the amount of parallelism in algorithms for scheduling a Pods. Default is 16
```
Does something like this look fine ?

pkg/scheduler/internal/parallelize/parallelism.go
hprateek43: I don't this there is a simpler way to do this as we have different invocations of the scheduler (some load defaults and some don't), so we need a way to set parallelism at the package level. @alculquicondor What do you recommend ?

alculquicondor: Right, some plugins' tests run without a scheduler object, so we would have to configure it (or pass as parameter) for every test and every external user (autoscaler, descheduler, etc).

This seems sufficient to me for now, but we can revisit in the future.

pkg/scheduler/internal/parallelize/parallelism.go
alculquicondor: this is answered in the other thread

pkg/scheduler/internal/parallelize/parallelism_test.go
```diff
@@ -0,0 +1,54 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
alculquicondor, line 2: 2020

staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go
alculquicondor: `Must be greater than 0. Defaults to 16`

hprateek43:
@alculquicondor @liggitt I have made the changes, we should be good for a merge now

pkg/scheduler/internal/parallelize/parallelism.go
liggitt: Global mutable package vars are still fragile, so I'm not thrilled about plumbing this this way. At the very least we should document the expectations (should only be set once, should be set from the main goroutine before starting any background goroutines that will read it). Ideally this would be protected via a setter that only lets the parallelism be set once on normal scheduler invocations and ensures the value is > 0

liggitt (commented):
the API bits look fine, left a couple comments on the global mutable package var

pkg/scheduler/apis/config/validation/validation.go
alculquicondor: I've never heard of non-power-of-two being a problem for parallelism. If anything, this should be up to the operator to decide if they want to use all the cores they have or not.

We could recommend, in the comment in types.go, to use up to the number of cores available in the system. But a global maximum doesn't necessarily make sense to me. We know current computers go up to about 256 cores, but I don't think we should hardcode that. On the other hand, a hard limit of 256 routines for a 2 core CPU wouldn't prevent an operator from achieving bad performance.

pkg/scheduler/internal/parallelize/parallelism.go
alculquicondor: A setter `MustSetParallelism` sgtm. This should panic if called more than once or value is <= 0

hprateek43:
> the API bits look fine, left a couple comments on the global mutable package var

I think that we can add a check that the parallelism is compared with `runtime.NumCPU()` and a factor to ensure that it is not out of place. Also I don't feel that it should be a two to power number, odd numbers can do just fine.

Regarding global mutable packages, we would require a cleanup of generic scheduler altogether to achieve a more consistent flow, once we have all the subsystems using the scheduler with bootstrapping, we can add logics around prevent the value from being altered after initialisation.

alculquicondor:
> once we have all the subsystems using the scheduler with bootstrapping, we can add logics around prevent the value from being altered after initialisation.

Nothing wrong with doing it in this PR

> I think that we can add a check that the parallelism is compared with runtime.NumCPU() and a factor to ensure that it is not out of place.

API validation shouldn't vary depending on the architecture. I prefer we just keep in the documentation (go doc)

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -23,21 +23,38 @@ import (
 	"k8s.io/client-go/util/workqueue"
 )
 
-const parallelism = 16
+// Parallelism is used the store value from scheduler config
+var parallelism int
```
alculquicondor, line 27: this is going to be 0 initially, which would cause problems:

```golang
var (
parallelism = 16
parallelismOverridden = false
)
```

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -23,21 +23,38 @@ import (
...
+// isParallelsimOveridden checks if the parallelsim value was already initialised.
+var isParallelsimOveridden = false
+
+// SetParallelism sets the parallelism for all scheduler algorithms.
+func SetParallelism(p int) {
+	if !(isParallelsimOveridden) {
+		isParallelsimOveridden = false
+		parallelism = p
+	} else {
+		panic("Parallelism can be initialised once.")
```
alculquicondor, line 38:
```golang
if isParallelismOverridden {
  panic("Cannot set parallelism more than once.")
}
parallelism = p
```

hprateek43:
@liggitt there are test cases that initialise the scheduler in a loop, leading to failure, eg,`cmd/kube-scheduler/app/server_test.go`, so how should we handle the setter in such scenarios ? or should we refactor the test cases ?

alculquicondor:
I'd say let's remove the panic for this PR.

For a followup, the solution should go like so:
scheduler.New creates a `Parallelizer` object. This object holds the parallelism and it's also shared with `framework.Framework`. Then, plugins that need to do parallelism call a `Parallelize` method from the `FrameworkHandle`, which uses the `Parallelizer` object

alculquicondor:
/lgtm

@liggitt do you agree with https://github.com/kubernetes/kubernetes/pull/94636#issuecomment-716574253 ?

pkg/scheduler/scheduler.go
```diff
@@ -109,6 +110,13 @@ func WithProfiles(p ...schedulerapi.KubeSchedulerProfile) Option {
 	}
 }
 
+// WithParallelism sets the parallelism for all scheduler algorithms. Default is 16.
+func WithParallelism(threads int32) Option {
+	return func(o *schedulerOptions) {
+		parallelize.SetParallelism(int(threads))
```
liggitt, line 116: add a `TODO(name or issue)` here to move to schedulerOptions and stop calling the global setter here

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -23,21 +23,30 @@ import (
 	"k8s.io/client-go/util/workqueue"
 )
 
-const parallelism = 16
+var (
+	parallelism = 16
+)
+
+// SetParallelism sets the parallelism for all scheduler algorithms.
```
liggitt, line 30: add a `TODO(name or issue)` here to move to framework and drop the package var, and indicate in a godoc comment this should only be called once per scheduler process

liggitt:
> @liggitt do you agree with [#94636 (comment)](https://github.com/kubernetes/kubernetes/pull/94636#issuecomment-716574253) ?

as long as there is a tracking issue and comments on the global setter and config option setter, I'm ok with this as a temporary state

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -23,21 +23,31 @@ import (
...
+// TODO: #95952
```
alculquicondor, line 31:
```
TODO(#95952): Remove global setter in favor of a struct that holds the configuration.
```

pkg/scheduler/scheduler.go
alculquicondor:
```
TODO(#95952): Remove global setter in favor of a struct that holds the configuration.
```

alculquicondor: Still not solved

alculquicondor:
/lgtm

@liggitt anything to add?

liggitt:
/approve

API bits lgtm. Would like to see progress on the TODOs to plumb the option more correctly.

Issue #88282, opened 2020-02-18
Document metadata.namespace behavior for webhooks


riking:
**What would you like to be added**:
Webhook authors should be aware that the metadata.namespace of submitted objects is not trustworthy in Mutating Webhooks, and may be blank in Validation Webhooks. The namespace needs to always be read out of the AdmissionReview object.

**Why is this needed**:
If a webhook author does not test their code with a modified kubectl default namespace, a validating webhook may incorrectly think a resource is being created in the `default` namespace.
Namespaces that mismatch between the URL and resource are not blocked by the apiserver until after the object has been sent to Mutating webhooks. A mutating webhook that reads the wrong namespace field may be vulnerable to information disclosure in error messages, or in extreme cases acting as a Confused Deputy if the webhook has side effects (please do not make a mutating webhook with side effects).

The correct fix for these problems is to fix the webhook code via developer education. Improve the docs so this is obvious.

/sig api-machinery
/sig docs
/area ecosystem

sftim:
I looked at the page https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook and saw that it [currently](https://github.com/kubernetes/website/blob/dff2f7b9601470dc8383d883b23188ed450d89c2/content/en/docs/reference/access-authn-authz/admission-controllers.md) mentions side effects of a webhook without any cautionary note about this being a bad idea.

riking:
> I looked at the page https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook and saw that it [currently](https://github.com/kubernetes/website/blob/dff2f7b9601470dc8383d883b23188ed450d89c2/content/en/docs/reference/access-authn-authz/admission-controllers.md) mentions side effects of a webhook without any cautionary note about this being a bad idea.

Yep, I think I had that paragraph in mind while writing the issue text.

riking:
This section https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#best-practices-and-warnings would be an excellent place to put this warning.

riking:
This section https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#best-practices-and-warnings would be an excellent place to put this warning.

To be clear, this is the issue:

> Namespaces that mismatch between the URL and resource are not blocked until after the object has been sent to Mutating webhooks.

/remove-lifecycle rotten

[...]

PR #94637, opened 2020-09-09
set/validate object namespace before admission

liggitt:
**What type of PR is this?**

/kind bug
/kind cleanup

Ensures the namespace population/check that occurs before an object is persisted is done before it is sent to admission. This ensures the namespace in the object matches the request namespace attribute.

Currently, namespaced objects can be sent to admission with empty namespaces during creation (and are defaulted to match the request namespace just before persisting) or a mismatched namespace (and are rejected just before persisting).

Added tests around the following scenarios:
* cluster-scoped, no namespace in object
* cluster-scoped, mismatched namespace in object
* namespaced, no namespace in object
* namespaced, matching namespace in object
* namespaced, mismatched namespace in object

xref https://github.com/kubernetes/kubernetes/issues/88282

```release-note
kube-apiserver: ensures the namespace of objects sent to admission webhooks matches the request namespace. Previously, objects without a namespace set would have the request namespace populated after mutating admission, and objects with a namespace that did not match the request namespace would be rejected after admission.
```

/cc @deads2k

### DIFF
--- a/pkg/registry/core/namespace/storage/storage_test.go
+++ b/pkg/registry/core/namespace/storage/storage_test.go
@@ -74,7 +74,7 @@ func TestCreateSetsFields(t *testing.T) {
 	defer server.Terminate(t)
 	defer storage.store.DestroyFunc()
 	namespace := validNewNamespace()
-	ctx := genericapirequest.NewContext()
+	ctx := genericapirequest.WithNamespace(genericapirequest.NewContext(), metav1.NamespaceNone)
 	_, err := storage.Create(ctx, namespace, rest.ValidateAllObjectFunc, &metav1.CreateOptions{})
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
@@ -152,7 +152,7 @@ func TestDeleteNamespaceWithIncompleteFinalizers(t *testing.T) {
 	defer server.Terminate(t)
 	defer storage.store.DestroyFunc()
 	key := "namespaces/foo"
-	ctx := genericapirequest.NewContext()
+	ctx := genericapirequest.WithNamespace(genericapirequest.NewContext(), metav1.NamespaceNone)
 	now := metav1.Now()
 	namespace := &api.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
@@ -167,6 +167,7 @@ func TestDeleteNamespaceWithIncompleteFinalizers(t *testing.T) {
 	if err := storage.store.Storage.Create(ctx, key, namespace, nil, 0, false); err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
+	ctx = genericapirequest.WithNamespace(genericapirequest.NewContext(), namespace.Name)
 	obj, immediate, err := storage.Delete(ctx, "foo", rest.ValidateAllObjectFunc, nil)
 	if err != nil {
 		t.Fatalf("unexpected error")
@@ -188,7 +189,7 @@ func TestUpdateDeletingNamespaceWithIncompleteMetadataFinalizers(t *testing.T) {
 	defer server.Terminate(t)
 	defer storage.store.DestroyFunc()
 	key := "namespaces/foo"
-	ctx := genericapirequest.NewContext()
+	ctx := genericapirequest.WithNamespace(genericapirequest.NewContext(), metav1.NamespaceNone)
 	now := metav1.Now()
 	namespace := &api.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
@@ -204,6 +205,7 @@ func TestUpdateDeletingNamespaceWithIncompleteMetadataFinalizers(t *testing.T) {
 	if err := storage.store.Storage.Create(ctx, key, namespace, nil, 0, false); err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
+	ctx = genericapirequest.WithNamespace(genericapirequest.NewContext(), namespace.Name)
 	ns, err := storage.Get(ctx, "foo", &metav1.GetOptions{})
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
@@ -223,7 +225,7 @@ func TestUpdateDeletingNamespaceWithIncompleteSpecFinalizers(t *testing.T) {
 	defer server.Terminate(t)
 	defer storage.store.DestroyFunc()
 	key := "namespaces/foo"
-	ctx := genericapirequest.NewContext()
+	ctx := genericapirequest.WithNamespace(genericapirequest.NewContext(), metav1.NamespaceNone)
 	now := metav1.Now()
 	namespace := &api.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
@@ -242,6 +244,7 @@ func TestUpdateDeletingNamespaceWithIncompleteSpecFinalizers(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
+	ctx = genericapirequest.WithNamespace(genericapirequest.NewContext(), namespace.Name)
 	if _, _, err = storage.Update(ctx, "foo", rest.DefaultUpdatedObjectInfo(ns), rest.ValidateAllObjectFunc, rest.ValidateAllObjectUpdateFunc, false, &metav1.UpdateOptions{}); err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
@@ -257,7 +260,7 @@ func TestUpdateDeletingNamespaceWithCompleteFinalizers(t *testing.T) {
 	defer server.Terminate(t)
 	defer storage.store.DestroyFunc()
 	key := "namespaces/foo"
-	ctx := genericapirequest.NewContext()
+	ctx := genericapirequest.WithNamespace(genericapirequest.NewContext(), metav1.NamespaceNone)
 	now := metav1.Now()
 	namespace := &api.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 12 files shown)


lavalamp:
/assign @deads2k 
cc @MikeSpreitzer @yue9944882

yue9944882 (approved):

MikeSpreitzer:
Do not forget that CRDs are still not fully up to the job of obviating the original way of defining new resources using aggregated custom apiservers.  Some of us are doing that, with a distinct etcd cluster for the custom resources too.  Now think about backup/restore of the etcd clusters.  It would be best if there is a rigorous independence of their content.  It might be best to not require one etcd cluster to contain a namespace object in order to persist an object in a different etcd cluster.

liggitt:
> Do not forget that CRDs are still not fully up to the job of obviating the original way of defining new resources using aggregated custom apiservers. Some of us are doing that, with a distinct etcd cluster for the custom resources too. Now think about backup/restore of the etcd clusters. It would be best if there is a rigorous independence of their content. It might be best to not require one etcd cluster to contain a namespace object in order to persist an object in a different etcd cluster.

That is unrelated to this PR.

The API server currently ensures the namespace in the request path and the namespace in the object match prior to persisting in etcd, but that check is done post-admission. This PR ensures that check is done prior to admission to avoid sending namespaced objects to admission with their namespace field unset (to be filled in later prior to persistence) or mismatching the request (to be rejected later prior to persistence).

staging/src/k8s.io/apiserver/pkg/registry/rest/meta_test.go
```diff
@@ -54,31 +54,3 @@ func TestHasObjectMetaSystemFieldValues(t *testing.T) {
...
-	if ValidNamespace(ctx, &resource.ObjectMeta) {
-		t.Fatalf("Expected error that resource and context errors do not match since context has no namespace")
-	}
-
-	ctx = genericapirequest.NewContext()
-	ns := genericapirequest.NamespaceValue(ctx)
-	if ns != "" {
-		t.Fatalf("Expected the empty string")
-	}
-}
```
sftim, line 84: Do there need to be any new tests to stand in place of these ones?

liggitt: yup, hence https://github.com/kubernetes/kubernetes/pull/94637#issuecomment-974701175 :)

sftim: OK, wasn't sure if you meant “This isn't complete until I write new tests” or “I want to see if the existing tests fail”.

liggitt:
updated with tests

liggitt:
the unit test failures are legitimate... there's a lot of unit tests not populating the context with the same namespace info a real API request would have gotten :-/

dims:
Ack @liggitt

MadhavJivrajani:
@liggitt 👋
I should be able to help out with updating the unit tests next week :)

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/create.go
```diff
@@ -152,7 +152,7 @@ func createHandler(r rest.NamedCreater, scope *RequestScope, admit admission.Int
 		if len(name) == 0 {
 			_, name, _ = scope.Namer.ObjectName(obj)
 		}
-		if len(namespace) == 0 && *gvk == namespaceGVK {
+		if len(namespace) == 0 && scope.Resource == namespaceGVR {
```
liggitt, line 155: this is existing, but comparing to the request scope resource rather than the decoded object Kind is more correct

liggitt:
pulled in unit test fixes from https://github.com/kubernetes/kubernetes/pull/108068

pkg/registry/core/persistentvolume/storage/storage_test.go
```diff
@@ -169,7 +169,7 @@ func TestUpdateStatus(t *testing.T) {
 	storage, statusStorage, server := newStorage(t)
 	defer server.Terminate(t)
 	defer storage.Store.DestroyFunc()
-	ctx := genericapirequest.NewContext()
+	ctx := genericapirequest.NewDefaultContext()
```
MadhavJivrajani, line 172:
```suggestion
	ctx := genericapirequest.WithNamespace(genericapirequest.NewContext(), metav1.NamespaceNone)
```
From https://github.com/kubernetes/kubernetes/pull/108068#discussion_r804682973
(I hadn't pushed suggested changes on account of https://github.com/kubernetes/kubernetes/pull/108068#discussion_r808130936

liggitt: ah, right... I squashed the comments into your commit, and split out the test bit that *has* to be part of this PR because it relies on the create/update handler changes

I think the current `Update tests to have namespace set in req context` commit in this PR could actually pass tests and merge to master ahead of this PR if you want to pull it back out... sorry for the churn :)

MadhavJivrajani: Nope, this looks good! Thanks!
The only thing I can think of is my commit message which contains details pre-your-review, you might have to one last rebase to change that 😬

fedebongio:
/assign @jpbetz 
Joe would you be able to take a look? 
/remove-triage accepted

pkg/registry/core/namespace/storage/storage_test.go
```diff
@@ -167,6 +167,7 @@ func TestDeleteNamespaceWithIncompleteFinalizers(t *testing.T) {
 	if err := storage.store.Storage.Create(ctx, key, namespace, nil, 0, false); err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
+	ctx = genericapirequest.WithNamespace(genericapirequest.NewContext(), namespace.Name)
```
deads2k, line 170: note to self to look up the current state of namespace value on the context of request for namespaces.

liggitt: see `request.WithNamespace` calls inside endpoints/handlers/* for current state

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/create.go
```diff
@@ -163,6 +163,17 @@ func createHandler(r rest.NamedCreater, scope *RequestScope, admit admission.Int
 
 		userInfo, _ := request.UserFrom(ctx)
 
+		// ensure namespace on the object is correct, or error if a conflicting namespace was set in the object
+		objectMeta, err := meta.Accessor(obj)
```
deads2k, line 167: breadcrumbs for me to chase.  I don't recall us previously enforcing objects to have objectmeta, so https://github.com/openshift/api/blob/master/authorization/v1/types.go#L238-L243 doesn't have it for instance. Perhaps it overrides this entire flow?  i wouldn't expect it to because that would bypass audit, but maybe.

liggitt: fair... if we can't get an accessor, I think it's reasonable to skip this here.

We still have the backstop in staging/src/k8s.io/apiserver/pkg/registry/rest/create.go and staging/src/k8s.io/apiserver/pkg/registry/rest/update.go that require Accessor and repeat the `EnsureObjectNamespaceMatchesRequestNamespace` check before persisting regular objects (types without objectmeta that implement their own Create impl skip *that* flow)

staging/src/k8s.io/apiserver/pkg/registry/rest/meta.go
```diff
@@ -31,13 +30,44 @@ func FillObjectMetaSystemFields(meta metav1.Object) {
...
+		return nil
+
+	case objNamespace == metav1.NamespaceNone:
+		// unset, default to request namespace
+		obj.SetNamespace(requestNamespace)
+		return nil
+
+	case requestNamespace == metav1.NamespaceNone:
+		// cluster-scoped, clear namespace
+		obj.SetNamespace(metav1.NamespaceNone)
```
deads2k, line 50: interesting.  We used to clear this?

staging/src/k8s.io/apiserver/pkg/registry/rest/update.go
```diff
@@ -110,12 +111,14 @@ func BeforeUpdate(strategy RESTUpdateStrategy, ctx context.Context, obj, old run
 	if kerr != nil {
 		return kerr
 	}
-	if strategy.NamespaceScoped() {
-		if !ValidNamespace(ctx, objectMeta) {
-			return errors.NewBadRequest("the namespace of the provided object does not match the namespace sent on the request")
-		}
-	} else if len(objectMeta.GetNamespace()) > 0 {
-		objectMeta.SetNamespace(metav1.NamespaceNone)
```
deads2k, line 118: so we did.  weird.

staging/src/k8s.io/apiserver/pkg/registry/rest/meta.go
liggitt: yeah, see BeforeCreate/BeforeUpdate

staging/src/k8s.io/apiserver/pkg/registry/rest/meta.go
```diff
@@ -31,13 +30,44 @@ func FillObjectMetaSystemFields(meta metav1.Object) {
 	meta.SetSelfLink("")
 }
 
-// ValidNamespace returns false if the namespace on the context differs from
-// the resource.  If the resource has no namespace, it is set to the value in
-// the context.
-func ValidNamespace(ctx context.Context, resource metav1.Object) bool {
-	ns, ok := genericapirequest.NamespaceFrom(ctx)
-	if len(resource.GetNamespace()) == 0 {
-		resource.SetNamespace(ns)
```
deads2k, line 40: how on earth did we ever write a ValidFoo method that mutates.  I'm too afraid to check blame.

PR #94639, opened 2020-09-09
Add roundtrip tests for metrics repo

wojtek-t:
Ref https://github.com/kubernetes/kubernetes/pull/94481

This is adding tests that would catch bug fixed by https://github.com/kubernetes/kubernetes/pull/94481

```release-note
NONE
```

/kind bug
/priority important-soon


wojtek-t:
@liggitt - once this get merge I will cherrypick both together to 1.19

PR #94643, opened 2020-09-09
Avoid unnecessary calls if other error occurred

lonng:
Signed-off-by: Lonng <heng@lonng.org>

**What type of PR is this?**

/kind cleanup

Tiny refactor of Kubelet startup function `run` and avoid unnecessary calls `cpuset.Parse` if an error occurred already (`kubeDeps.CAdvisorInterface.MachineInfo()`).

There is no logical changed and keep the consistent semantic with the previous structure but with some tiny optimization.

NONE

```release-note
NONE
```

### DIFF
--- a/cmd/kubelet/app/server.go
+++ b/cmd/kubelet/app/server.go
@@ -646,14 +646,7 @@ func run(ctx context.Context, s *options.KubeletServer, kubeDeps *kubelet.Depend
 		}
 
 		var reservedSystemCPUs cpuset.CPUSet
-		var errParse error
 		if s.ReservedSystemCPUs != "" {
-			reservedSystemCPUs, errParse = cpuset.Parse(s.ReservedSystemCPUs)
-			if errParse != nil {
-				// invalid cpu list is provided, set reservedSystemCPUs to empty, so it won't overwrite kubeReserved/systemReserved
-				klog.Infof("Invalid ReservedSystemCPUs \"%s\"", s.ReservedSystemCPUs)
-				return errParse
-			}
 			// is it safe do use CAdvisor here ??
 			machineInfo, err := kubeDeps.CAdvisorInterface.MachineInfo()
 			if err != nil {
@@ -661,6 +654,13 @@ func run(ctx context.Context, s *options.KubeletServer, kubeDeps *kubelet.Depend
 				klog.Warning("Failed to get MachineInfo, set reservedSystemCPUs to empty")
 				reservedSystemCPUs = cpuset.NewCPUSet()
 			} else {
+				var errParse error
+				reservedSystemCPUs, errParse = cpuset.Parse(s.ReservedSystemCPUs)
+				if errParse != nil {
+					// invalid cpu list is provided, set reservedSystemCPUs to empty, so it won't overwrite kubeReserved/systemReserved
+					klog.Infof("Invalid ReservedSystemCPUs \"%s\"", s.ReservedSystemCPUs)
+					return errParse
+				}
 				reservedList := reservedSystemCPUs.ToSlice()
 				first := reservedList[0]
 				last := reservedList[len(reservedList)-1]


cmd/kubelet/app/server.go
```diff
@@ -646,21 +646,21 @@ func run(ctx context.Context, s *options.KubeletServer, kubeDeps *kubelet.Depend
...
-				return errParse
-			}
 			// is it safe do use CAdvisor here ??
 			machineInfo, err := kubeDeps.CAdvisorInterface.MachineInfo()
 			if err != nil {
 				// if can't use CAdvisor here, fall back to non-explicit cpu list behavor
 				klog.Warning("Failed to get MachineInfo, set reservedSystemCPUs to empty")
 				reservedSystemCPUs = cpuset.NewCPUSet()
 			} else {
+				var errParse error
```
JornShen, line 657: no need to define errParse, next line you can write like this 
> reservedSystemCPUs, errParse := cpuset.Parse(s.ReservedSystemCPUs)

JornShen:
I have have a more nice code 
```go
func run(ctx context.Context, s *options.KubeletServer, kubeDeps *kubelet.Dependencies, featureGate featuregate.FeatureGate) (err error) {
	// Set global feature gates based on the value on the initial KubeletServer
		...
	if kubeDeps.ContainerManager == nil {
	    
	    ...
		var reservedSystemCPUs cpuset.CPUSet
		// refector func
		if s.ReservedSystemCPUs == "" {
			reservedSystemCPUs = cpuset.NewCPUSet()
		} else if machineInfo, err := kubeDeps.CAdvisorInterface.MachineInfo(); err != nil {
			// if can't use CAdvisor here, fall back to non-explicit cpu list behavor
			klog.Warning("Failed to get MachineInfo, set reservedSystemCPUs to empty")
			reservedSystemCPUs = cpuset.NewCPUSet()
		} else if reservedSystemCPUs, errParse := cpuset.Parse(s.ReservedSystemCPUs); errParse != nil {
			// invalid cpu list is provided, set reservedSystemCPUs to empty, so it won't overwrite kubeReserved/systemReserved
			klog.Infof("Invalid ReservedSystemCPUs \"%s\"", s.ReservedSystemCPUs)
			return errParse
		} else {
			reservedList := reservedSystemCPUs.ToSlice()
			first := reservedList[0]
			last := reservedList[len(reservedList)-1]
			if first < 0 || last >= machineInfo.NumCores {
				// the specified cpuset is outside of the range of what the machine has
				klog.Infof("Invalid cpuset specified by --reserved-cpus")
				return fmt.Errorf("Invalid cpuset %q specified by --reserved-cpus", s.ReservedSystemCPUs)
			}
		}
 		...
 	}
 	...
 }
```
That seems to be more clear.

you can consider to change to it, maybe. :)
@lonng

cmd/kubelet/app/server.go
lonng: No, the result will not as expect as you, see: https://play.golang.org/p/G4CwknE4G1X

lonng:
@JornShen IMO, I think the `if {...} else if {..} else if {..}` doesn't make it more clear.

cmd/kubelet/app/server.go
JornShen: sorry, you are right.

JornShen: :(, sorry, you are right.

JornShen:
> @JornShen IMO, I think the `if {...} else if {..} else if {..}` doesn't make it more clear.

all right. Just a little advise. I think this code's cycle complexity is a bit high. In the most inside code, as you added, increase the cycle complexity.  Just my opinion..... :)

lonng:
@JornShen Thanks for your advice, extract the outer-most `if` statement as an individual function, which may reduce the cycle complexity. But current implementation if more straight-forward.

lonng:
@JornShen Can this PR be merged?

lonng:
ping @JornShen @oomichi @derekwaynecarr

oomichi:
@lonng I don't have power to merge/approve this PR which is under `cmd/kubelet`. @derekwaynecarr has.

lonng:
How can I merge this PR?  @resouer @mattjmcnaughton

JornShen:
> How can I merge this PR? @resouer @mattjmcnaughton

I think you can send this PR to sig-node channel on slack,  or connect them to review and approve it directly.

lonng:
@derekwaynecarr I had sent this PR into `sig-node`, can you approve this PR and help merge it? Or give some suggestions about how to merge PR?

cmd/kubelet/app/server.go
```diff
@@ -646,21 +646,21 @@ func run(ctx context.Context, s *options.KubeletServer, kubeDeps *kubelet.Depend
...
+				reservedSystemCPUs, errParse = cpuset.Parse(s.ReservedSystemCPUs)
```
rphillips, line 658: we should reuse the err from line 651.

```
reservedSystemCPUs, err = cpuset.Parse(s.ReservedSystemCPUs)
```

rphillips (changes requested):

PR #94647, opened 2020-09-09
storage E2E: explicitly wait for PV deletion after ephemeral test

pohly:
**What type of PR is this?**
/kind bug
/kind failing-test
/kind flake

Even with foreground deletion, removal of the PVs that may have been
created for a pod with generic ephemeral volumes happens
asynchronously, in the worst case after the test has completed and the
driver for the volume got removed.

Perhaps this can be fixed in Kubernetes itself, but for now we need to
deal with it as part of the test.

Not observed to be a problem in Kubernetes itself, but was encountered when running the test against a third-party CSI driver (PMEM-CSI).

```release-note
NONE
```

### DIFF
--- a/test/e2e/storage/testsuites/provisioning.go
+++ b/test/e2e/storage/testsuites/provisioning.go
@@ -725,6 +725,28 @@ func StopPodAndDependents(c clientset.Interface, pod *v1.Pod) {
 	} else {
 		framework.Logf("Pod %s has the following logs: %s", pod.Name, body)
 	}
+
+	// We must wait explicitly for removal of the generic ephemeral volume PVs.
+	// For that we must find them first...
+	pvs, err := c.CoreV1().PersistentVolumes().List(context.TODO(), metav1.ListOptions{})
+	framework.ExpectNoError(err, "list PVs")
+	var podPVs []v1.PersistentVolume
+	for _, pv := range pvs.Items {
+		if pv.Spec.ClaimRef == nil ||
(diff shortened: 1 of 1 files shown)


test/e2e/storage/testsuites/provisioning.go
```diff
@@ -725,6 +725,28 @@ func StopPodAndDependents(c clientset.Interface, pod *v1.Pod) {
...
 	}
+
+	// We must wait explicitly for removal of the generic ephemeral volume PVs.
+	// For that we must find them first...
+	pvs, err := c.CoreV1().PersistentVolumes().List(context.TODO(), metav1.ListOptions{})
+	framework.ExpectNoError(err, "list PVs")
+	var podPVs []v1.PersistentVolume
+	for _, pv := range pvs.Items {
+		if pv.Spec.ClaimRef == nil ||
+			pv.Spec.ClaimRef.Namespace != pod.Namespace {
```
msau42, line 736: How do we know these PVs are generic ephemeral volumes and not just normal PVs?

msau42: I guess either way, we want all PVs to be cleaned up regardless.

pohly: > How do we know these PVs are generic ephemeral volumes and not just normal PVs?

Because they are bound to the PVCs which are owned by the pod that is being deleted.

PR #94650, opened 2020-09-09
[do not merge] verify that local-up-cluster fails in CI

jayunit100:
This is a test to probe wether or not local-e2e.yaml test jobs in test-infra are capable of passing.

I think the fix for this test, is https://github.com/kubernetes/test-infra/pull/19142.


jayunit100:
we can close this now, I just wanted to verify that the local-e2e test fails , which it does:... https://github.com/kubernetes/test-infra/pull/19142 should fix this issue.

PR #94652, opened 2020-09-09
[WIP] Switch kubectl to prefer communicating with kube-apiserver using protobufs

wojtek-t:
```release-note
Switch kubectl to prefer communicating with kube-apiserver using protobufs
```

/kind enhancement
/priority important-soon
/sig scalability
/sig cli

### DIFF
--- a/staging/src/k8s.io/kubectl/pkg/cmd/get/get.go
+++ b/staging/src/k8s.io/kubectl/pkg/cmd/get/get.go
@@ -31,7 +31,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	metav1beta1 "k8s.io/apimachinery/pkg/apis/meta/v1beta1"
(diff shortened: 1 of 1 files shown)


staging/src/k8s.io/kubectl/pkg/cmd/util/factory_client_access.go
```diff
@@ -60,7 +75,12 @@ func NewFactory(clientGetter genericclioptions.RESTClientGetter) Factory {
 }
 
 func (f *factoryImpl) ToRESTConfig() (*restclient.Config, error) {
-	return f.clientGetter.ToRESTConfig()
+	cfg, err := f.clientGetter.ToRESTConfig()
+	if err != nil {
+		return nil, err
+	}
+	cfg.AcceptContentTypes = kubectlAcceptHeader()
```
liggitt, line 82: we don't want to switch to requesting proto table access globally ... only in the get command (https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/cmd/get/get.go#L434-L438)

wojtek-t: Switched - that said it may not help given we're using unstructured...

PR #94656, opened 2020-09-09
Do not add namespace to a request when creating it using SSA

### LINKED ISSUE #94168: patch-based (server-side-apply) and post-based Namespace creation requests to different endpoints populate different namespace attribute in AdmissionReview
**What happened**:

It appears that validating admission requests for server-side apply are populating the `namespace` for requests to create Namespace objects.

Non server-side apply requests don't do this, and a missing namespace field is the only way to distinguish cluster-scoped resources from namespace-scoped resources. This appears to be a breaking behavioral change.

See https://github.com/open-policy-agent/gatekeeper/issues/792#issuecomment-678437342

Requests to create namespaces would have a consistent signature between server-side apply and non-server-side apply contexts.

`kubectl apply -f ns.yml -v 8 --server-side`

See https://github.com/open-policy-agent/gatekeeper/issues/792

**Environment**:
- Kubernetes version (use `kubectl version`): `1.18.2`
- Cloud provider or hardware configuration: kind
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

nodo:
**What type of PR is this?**

/kind bug

When creating a new namespace using server-side apply a`PATCH` request is sent at `/namespaces/new-namespace`.  In this case, we don't want to add the namespace to the request.

The namespace is fetched [here](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/patch.go#L92) and then used all subsequent requests in the webhooks, which caused the issue below.

Fixes #94168

```release-note
NONE
```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go
+++ b/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go
@@ -178,7 +178,12 @@ func (r *RequestInfoFactory) NewRequestInfo(req *http.Request) (*RequestInfo, er
 
 	// URL forms: /namespaces/{namespace}/{kind}/*, where parts are adjusted to be relative to kind
 	if currentParts[0] == "namespaces" {
-		if len(currentParts) > 1 {
+		// When creating a new namespace using server-side apply a PATCH request is
+		// sent at /namespaces/new-namespace.  In this case, we don't want to add
+		// the namespace to the request.
+		if len(currentParts) == 2 && requestInfo.Verb == "patch" {
+			requestInfo.Namespace = metav1.NamespaceNone
+		} else if len(currentParts) > 1 {
 			requestInfo.Namespace = currentParts[1]
 
 			// if there is another step after the namespace name and it is not a known namespace subresource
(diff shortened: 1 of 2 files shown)


staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go
```diff
@@ -178,7 +178,12 @@ func (r *RequestInfoFactory) NewRequestInfo(req *http.Request) (*RequestInfo, er
 
 	// URL forms: /namespaces/{namespace}/{kind}/*, where parts are adjusted to be relative to kind
 	if currentParts[0] == "namespaces" {
-		if len(currentParts) > 1 {
+		// When creating a new namespace using server-side apply a PATCH request is
+		// sent at /namespaces/new-namespace.  In this case, we don't want to add
+		// the namespace to the request.
+		if len(currentParts) == 2 && requestInfo.Verb == "patch" {
+			requestInfo.Namespace = metav1.NamespaceNone
```
liggitt, line 185: this is not the right layer to make this change at... we should understand where the data is being set into the object and why it differs for patch requests

liggitt: for request-info purposes, updates to a namespace *are* namespaced... setting this to "" here means that a user that was previously authorized to patch a specifically named namespace will get rejected

liggitt:
/hold

need to understand what attribute is being set differently for server-side apply and address that issue, not modify requestinfo parsing

liggitt:
resolved the root cause and described why this is currently working as expected in https://github.com/kubernetes/kubernetes/issues/94168#issuecomment-689696269

staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go
nodo: Oh, makes sense! thanks for spotting it!

PR #94661, opened 2020-09-09
[KEP-1933] Provide hack/ testing target for static analysis.

PurelyApplied:
**What type of PR is this?**

/kind feature

Provides a test target at `kubernetes/hack/verify-govet-levee.sh` for running [go-flow-levee](https://github.com/google/go-flow-levee) static analysis, as proposed in KEP-1933.  Prow integration in `test-infra` will follow.

Introduces analyzer dependencies and execution script.

In an ideal situation, analysis would be wired into a `sh_test` Bazel target.
However, because of the separation of build and testing, and because of Bazel's sandboxing of tests, `sh_test` (and `go_test`) targets have access to neither build artifacts nor a buildable environment (e.g., a configured `GOCACHE`, etc).
As a result, any analysis depending on `buildssa` or other callgraph-capable package will suffer panic at runtime.

Such analysis has more in common with `go vet`, and indeed `go vet -vettool=...` is the standard entry-point for analyzers based on the `analysis` package.  Within Kubernetes, we apply vet via `make vet`, with a vestigial `hack/verify-govet.sh` providing redirection.

Given intent of this analysis as an explicit test, it seemed to me that a developer would want a wrapping script to run locally if errors are found during PR testing, rather than requiring wider knowledge of how to properly construct `make vet WHAT="-vettool=... -config=..."`.

```release-note
NONE
```

KEP-1933: [at kubernetes/enhancements](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1933-secret-logging-static-analysis).

### DIFF
--- a/hack/testdata/levee/OWNERS
+++ b/hack/testdata/levee/OWNERS
@@ -1,0 +1,6 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+- spiffxp
+reviewers:
+- spiffxp
--- a/hack/tools/go.mod
+++ b/hack/tools/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/bazelbuild/buildtools v0.0.0-20200228172928-c9d9e342afdb
 	github.com/cespare/prettybench v0.0.0-20150116022406-03b8cfe5406c
 	github.com/client9/misspell v0.3.4
+	github.com/google/go-flow-levee v0.1.4-0.20201102181719-72c65d71b1d3
 	golang.org/x/lint v0.0.0-20200302205851-738671d3881b
 	gotest.tools v2.2.0+incompatible
 	gotest.tools/gotestsum v0.3.5
--- a/hack/testdata/levee/levee-config.yaml
+++ b/hack/testdata/levee/levee-config.yaml
@@ -1,0 +1,60 @@
+# This file holds configuration for taint propagation analysis of Kubernetes source via go-flow-levee.
+# It defines sources which may contain credentials and sinks where these should not be logged.
+# Sources may be identified by the FieldTags element, or by matching package, type, and field explicitly in the Sources element.
+# Sanitizers permit sources to safely reach a sink.
+# False positives may be suppressed in the Exclude block.
+# Note that `*RE` keys have regexp values.
+---
+# These field tags were introduced by KEP-1753 to fields which may contain credentials
+FieldTags:
+  - Key: "datapolicy"
+    Val: "security-key"
+  - Key: "datapolicy"
+    Val: "token"
+  - Key: "datapolicy"
+    Val: "password"
+
+# This preliminary collection of source types should be removed once
+# KEP-1753 adds tags to the relevant fields.
+Sources:
+  - PackageRE: ""
+    TypeRE: "^(?:admin)?Secret$|Token"
+    FieldRE: ""
+  - PackageRE: "k8s.io/client-go/tools/clientcmd/api(?:/v1)?"
+    TypeRE: "^(?:Named)?AuthInfo$"
+    FieldRE: ""
+  - PackageRE: "k8s.io/kubernetes/pkg/credentialprovider"
+    TypeRE: "DockerConfigEntry"
+    FieldRE: "Password"
+  - PackageRE: "k8s.io/client-go/transport"
+    TypeRE: "requestInfo"
+    FieldRE: "RequestHeaders"
+  - PackageRE: "k8s.io/kubernetes/pkg/volume/rbd"
+    TypeRE: "rbdMounter"
+    FieldRE: "adminSecret"
+  - PackageRE: "^k8s.io/client-go/rest$"
+    TypeRE: "^TLSClientConfig$"
+    FieldRE: "Password|BearerToken$|"
+  - PackageRE: "^k8s.io/client-go/rest$"
+    TypeRE: "^Config$"
+    FieldRE: "Password|BearerToken$|"
+
+# Sinks are functions that should not be called with source or source-tainted arguments.
+# This configuration should capture all log unfiltered log calls.
+Sinks:
+  - PackageRE: "\bk?log\b"
+    ReceiverRE: ""
+    MethodRE: "Info|Warning|Error|Fatal|Exit"
+  - PackageRE: "\bk?log\b"
+    ReceiverRE: "Verbose"
+    MethodRE: "Info|Error"
+
+# Sanitizers permit a source to reach a sink by explicitly removing the source data.
+Sanitizers:
+  - PackageRE: "k8s.io/client-go/transport"
+    MethodRE: "maskValue"
+
+# False positives may be suppressed here.
+# Exclude reporting within a given function by specifying it similar to Sinks, i.e.,
+# PackageRE | ReceiverRE | MethodRE regexp
+Exclude: []
--- a/hack/tools/LICENSES/vendor/github.com/google/go-flow-levee/LICENSE
+++ b/hack/tools/LICENSES/vendor/github.com/google/go-flow-levee/LICENSE
@@ -1,0 +1,206 @@
+= vendor/github.com/google/go-flow-levee licensed under: =
+
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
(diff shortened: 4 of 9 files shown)


test/instrumentation/levee/OWNERS
```diff
@@ -0,0 +1,11 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+- sig-instrumentation-approvers
+- logicalhan
```
ehashman, line 5: @logicalhan is a sig-instrumentation-approver, can we update this file? It could match the one below https://github.com/kubernetes/kubernetes/pull/94661/files#diff-7cb13a5003b89536f79c038d68c9e843

ehashman (commented):
/ok-to-test
/assign
/wg security

ehashman:
Let's try again....


/area security
/wg security-audit

go.sum
```diff
@@ -231,6 +231,8 @@ github.com/google/cadvisor v0.37.0 h1:t3txV4zNZZGTuwuA/Onm3HToPhg16GjigAHZHEVIz+
 github.com/google/cadvisor v0.37.0/go.mod h1:OhDE+goNVel0eGY8mR7Ifq1QUI1in5vJBIgIpcajK/I=
 github.com/google/go-cmp v0.4.0 h1:xsAVV57WRhGj6kEIi8ReJzQlHHqcBYCElAvkovg3B/4=
 github.com/google/go-cmp v0.4.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
+github.com/google/go-flow-levee v0.1.3 h1:bg7F/zMjwFzWrV/pF4TIfR2cO+K+zdpG1k9eWgFp3WM=
+github.com/google/go-flow-levee v0.1.3/go.mod h1:C435ZDtvbgiwHqcjW2tAPJseY4bpiLJrarS3il+uNek=
```
BenTheElder, line 235: this should not be in the primary vendor / go.sum
this is a development-time dependency only?

we have a pattern for avoiding this with the hack/verify* targets. see e.g. staticcheck

BenTheElder: /hold

test/instrumentation/levee/verify-govet-levee.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
BenTheElder, line 1: this should be under `hack/` like the other verify targets.

test/instrumentation/levee/main.go
```diff
@@ -0,0 +1,26 @@
...
+
+package main
+
+import (
+	"github.com/google/go-flow-levee/pkg/levee"
+	"golang.org/x/tools/go/analysis/singlechecker"
+)
+
+func main() {
+	singlechecker.Main(levee.Analyzer)
```
BenTheElder, line 25: does go-flow-levee not provide a binary for this upstream?

test/instrumentation/levee/verify-govet-levee.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# This script vets all *.go files by running `go vet -vettool=levee`.
+# It is equivalent to `make vet` with appropraite flags.
+# Usage: `test/instrumentation/levee/verify-govet-levee.sh`.
+
+set -o errexit
+set -o pipefail
+
+KUBE_ROOT=$(dirname $(realpath "${BASH_SOURCE[0]}"))/../../../
+BAZEL_OUT_DIR="$KUBE_ROOT/bazel-bin"
```
BenTheElder, line 25: what's the reason for using bazel here?
none of the other verification tools use it. we use it for some large build / unit test for caching purposes, but even that is not necessarily something we're investing in further / may remove.

importantly I don't think we can use bazel here while not polluting vendor / go.mod

test/instrumentation/levee/main.go
PurelyApplied: It does, but I was struggling to get it to play nicely with the `go.mod`.  But with your other comments, it looks like I missed a useful pattern in that regard.  I'll investigate shortly.

hack/testdata/levee-config.json
```diff
@@ -0,0 +1,32 @@
+{
```
PurelyApplied, line 1: This config is a cursory first-pass.  It will be iterated upon, and [KEP-1753](https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/1753-logs-sanitization/README.md) will introduce useful field tags to drive identification of `Source` fields that should not be logged.

PurelyApplied:
@BenTheElder I want to make sure I have the general outline correct.

* Analysis testing should be exposed as a point in `k/kubernetes`, as in the script in this PR.  The configuration here is the source of truth for analysis and the third_party version consumed.
* That test point should be targeted by a new job in `k/test-infra`.

Given the above, if I were to write some Kubernetes-specific tests to verify analysis and the correctness of our configuration, should those supplementary tests live in `k/kubernetes/hack` as well?

PurelyApplied:
/assign @fejta

Hello, fejta!  I pinged Ben above regarding best practices for where configuration tests should live.  If you have any input on those fronts, I would appreciate it.

hack/tools/go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/bazelbuild/buildtools v0.0.0-20200228172928-c9d9e342afdb
 	github.com/cespare/prettybench v0.0.0-20150116022406-03b8cfe5406c
 	github.com/client9/misspell v0.3.4
+	github.com/google/go-flow-levee v0.1.3
```
PurelyApplied, line 10: TODO: Target stationary version.

PurelyApplied:
@ehashman Thank you for your review on KEP-1933 itself.  Any input here on the implementation?  As I'm still finding my feet in the community, would you recommend anyone else to ping on this PR?

hack/testdata/levee-config.json
```diff
@@ -0,0 +1,57 @@
...
```
sttts, line 1: what are these? Can we use a better name which makes it clear? Are those exceptions from govet?

PurelyApplied: Json `"Comment":` blocks added for clarity.

It's a configuration for the taint-propagation analysis being introduced, specifying where we keep credentials and where those creds should not go.  The source specification will be almost entirely replaced by field tags when they are introduced by KEP-1753.  And in fact, while I adding these comments, a collection of ~20 such PRs were opened to "Add datapolicy tags to ...".

BenTheElder:
> verify: govet-levee | 5m50s

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/94661/pull-kubernetes-verify/1321882815230382080

Noting that this takes ~6 minutes but that overally verify is at 35m which is not bad.
This might be long enough to split out like `pull-kubernetes-typecheck` in the future, but shouldn't be necessary to start.

spiffxp:
flake failure for //pkg/registry/autoscaling/horizontalpodautoscaler/storage

flake failure for Kubernetes e2e suite: [sig-apps] ReplicationController should test the lifecycle of a ReplicationController [Conformance]

Agree with @BenTheElder that leaving in verify job for now sounds ok.  If it becomes too painful, may merit its own presubmit similar to typecheck. Also, creating an optional/manually triggered presubmit for this may allow for faster iteration.

hack/testdata/levee-config.json
```diff
@@ -0,0 +1,84 @@
...
```
spiffxp, line 1: I would recommend a less generic name for this subdir, and giving it an OWNERS file with focused approvers so you don't have to keep hitting /hack approvers

spiffxp (changes requested):
I'm ok with this, and like the additional hack script.  I have one specific concern to address for iterating on `levee-config.json`

It is a shame the tool in question doesn't support yaml, so we could comment on why the specific rules.  Adding comment fields for those might make sense for future maintainers, but I won't block, can be done as followup.

ehashman (commented):
Looks okay at a glance to me, deferring to SIG Testing.

PurelyApplied:
> It is a shame the tool in question doesn't support yaml, so we could comment on why the specific rules. Adding comment fields for those might make sense for future maintainers, but I won't block, can be done as followup.

Incidentally, I think I was working on the migration from the `json` marshaller to the `sigs.k8s.io/yaml` marhsaller when you were making your comments on Friday.  Tool reference updated and configuration converted to YAML for improved comments and readability.

hack/testdata/levee-config.json
PurelyApplied: I've added a `testdata` subdirectory to hold the configuration and related OWNERS file.  Is that how you meant your comment for a less generic subdir?  `testdata` itself is already home to 100+ files, so I didn't figure you meant that when you said "this subdir."  Alternatively, I could just update the `testdata` OWNERS with a filter on this config, if that seems better organized.

It's on the SIG Security agenda this afternoon to discuss taking ownership of this tool, but I've gone ahead and put `sig-security` as the tentative OWNERS for the config.

hack/testdata/levee/OWNERS
```diff
@@ -0,0 +1,6 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+- sig-security
+reviewers:
+- sig-security
```
spiffxp, line 6: This group doesn't exist in `/OWNERS_ALIASES`.  I would recommend putting github usernames in here, and deciding later whether you want to go the route of adding and using a group alias.  You would need root approvers to update `/OWNERS_ALIASES`

hack/testdata/levee-config.json
spiffxp: Filters don't work well for approvers.

I meant instead of `/hack/testdata` which sounds like it should be owned by all kinds of different people, use something named after the tool you're owning.  `/hack/testdata/levee` is fine.  I would also have been OK with something like `/hack/levee`

spiffxp (changes requested):
Very cool that you migrated this to YAML!

My only remaining concern is that you address the bot's comment: https://github.com/kubernetes/kubernetes/pull/94661/files#issuecomment-720693580

I suggested how to here: https://github.com/kubernetes/kubernetes/pull/94661/files#r516229248

hack/testdata/levee/OWNERS
PurelyApplied: Gotcha.

Is there anyone in particular you'd like to see in the OWNERS then?  You, me, and dupree?

go.sum
BenTheElder: this hold was addressed @PurelyApplied -- btw ~anyone can `/hold cancel` :sweat_smile:

BenTheElder:
Can you squash the commit history? We usually only merge PRs with logical commits, rather than the full iterative history (to aid bisecting etc., we have a *ton* of commits..)

PurelyApplied:
Yeah, I'll do that.  I try to avoid force-pushing too much, since it strips context from the comment chains.

hack/testdata/levee/OWNERS
```diff
@@ -0,0 +1,6 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+- spiffxp
+reviewers:
+- spiffxp
```
spiffxp, line 6: I'll accept this for now, when we update with other members, I would like to be removed

spiffxp (approved):
/hold cancel
/approve
/lgtm

hack/testdata/levee/OWNERS
PurelyApplied: Thanks!

PurelyApplied:
Big money no whammy stop!

/retest

PurelyApplied:
Once more with feeling    
/retest

PurelyApplied:
> @PurelyApplied: The following tests failed, say ...to rerun all failed tests:

You lied to me, ci-bot.  You reran passing tests.

Guess I'll call them out one at a time.

PurelyApplied:
--

BenTheElder:
Passing tests must pass again against (HEAD of target branch + latest PR changes).
If something merges since the last pass, then prow will invalidate results before merge.

BenTheElder:
https://github.com/kubernetes/test-infra/blob/master/prow/cmd/tide/README.md (merge robot)

PurelyApplied:
Thanks for confirming my suspicions, Ben!

PR #94449, opened 2020-09-02
[go1.15] Update to go1.15.2

justaugustus:
**What type of PR is this?**

/kind feature
/area dependency

**What this PR does / why we need it**:

- Update to go1.15.2
- build: Update to k/repo-infra@v0.1.1 (supports go1.15.2)
- build: Use go-runner:buster-v2.0.1 (built using go1.15.1)
- bazel: Replace --features with Starlark build settings flag
- hack/lib/util.sh: some bash cleanups
  
  - switched one spot to use kube::logging
  - make kube::util::find-binary return an error when it doesn't find
    anything so that hack scripts fail fast instead of with '' binary not
    found errors.
  - this required deleting some genfeddoc stuff. the binary no longer
    exists in k/k repo since we removed federation/, and I don't see it
    in https://github.com/kubernetes-sigs/kubefed/ either. I'm assuming
    that it's gone for good now.

- bazel: output go_binary rule directly from go_binary_conditional_pure
  
  From: @mikedanese:
  Instead of aliasing. Aliases are annoying in a number of ways. This is
  specifically bugging me now because they make the action graph harder to
  analyze programmatically. By using aliases here, we would need to handle
  potentially aliased go_binary targets and dereference to the effective
  target.

  The comment references an issue with `pure = select(...)` which appears
  to be resolved considering this now builds.

- make kube::util::find-binary not dependent on bazel-out/ structure

  Implement an aspect that outputs go_build_mode metadata for go binaries,
  and use that during binary selection.

Tracking issue: https://github.com/kubernetes/release/issues/1517
ref: https://github.com/kubernetes/release/pull/1514, https://github.com/kubernetes/k8s.io/pull/1205, https://github.com/kubernetes/release/pull/1516
cc: @kubernetes/release-engineering 

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

~/hold this will fail some tests while I'm working on the k/repo-infra update~

**Does this PR introduce a user-facing change?**:

```release-note
- Kubernetes is now built using go1.15.2
- build: Update to k/repo-infra@v0.1.1 (supports go1.15.2)
- build: Use go-runner:buster-v2.0.1 (built using go1.15.1)
- bazel: Replace --features with Starlark build settings flag
- hack/lib/util.sh: some bash cleanups
  
  - switched one spot to use kube::logging
  - make kube::util::find-binary return an error when it doesn't find
    anything so that hack scripts fail fast instead of with '' binary not
    found errors.
[...]

PR #94662, opened 2020-09-09
output go_binary rule directly from go_binary_conditional_pure

mikedanese:
Instead of aliasing. Aliases are annoying in a number of ways. This is
specifically bugging me now because they make the action graph harder to
analyze programmatically. By using aliases here, we would need to handle
potentially aliased go_binary targets and dereference to the effective
target.

The comment references an issue with `pure = select(...)` which appears
to be resolved considering this now builds.

Part of unblocking: #94449

/kind cleanup
/sig release

```release-note
NONE
```

cc @fejta @BenTheElder @justaugustus

### DIFF
--- a/build/go.bzl
+++ b/build/go.bzl
@@ -14,41 +14,22 @@
 
 load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_test")
 
-# Defines several go_binary rules to work around a Bazel issue which makes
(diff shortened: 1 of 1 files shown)


mikedanese:
Heh, so this only works after https://github.com/kubernetes/kubernetes/pull/94449 goes in... Let's work around it for now. I might reopen later.

PR #94634, opened 2020-09-08
Switch cos version to M85

vteratipally:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
Fixes #
**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #94669, opened 2020-09-09
Revert "Switch cos version to M85"

vteratipally:
Reverts kubernetes/kubernetes#94634

Containerd tests are failing which have dependent tests  that are failing. Reverting the changes till the root cause is figured.


vteratipally:
@dims could you please set the milestone so that it is merged

PR #94670, opened 2020-09-10
WIP: kube::util::find-binary should look at action graph for bazel artifacts

mikedanese:
Opening to test in CI and see what breaks.


hack/lib/util.sh
```diff
@@ -207,19 +207,31 @@ kube::util::find-binary-for-platform() {
...
+
+    # Query the action graph. The output of the command is of the form:
+    #
+    # https://cs.opensource.google/bazel/bazel/+/master:src/main/protobuf/analysis.proto
+    local toolchain="@io_bazel_rules_go//go/toolchain:${platform/\//_}"
+    local bzl_location=$(2>/dev/null \
+      bazel \
+        aquery --platforms "${toolchain}" --output jsonproto \
+        "//cmd/${lookfor}:${lookfor}" \
+        | jq -r \
```
mikedanese, line 223: We can drop the use of jq here after we update bazel to 3.4.1.

mikedanese: Removed jq dep in favor of sed and grep.

PR #94672, opened 2020-09-10
[WIP] Add sampling histogram

### LINKED ISSUE #796: Feature Request: Permit histogram's Observe() method to take an observation count
We had a case where we in theory update a histogram at some frequency. In actuality, we only update it when things are happening, and when nothing is happening, we just count the amount of time that has passed and then update it with that many ~zeros~ [edit: observations] when the next thing happens.

Right now adding a value to a historgram N times takes N calls of Observe(). It would be very nice if we could pass N as a parameter instead, and if I follow the code, it could be done in constant time.

MikeSpreitzer:
**What type of PR is this?**

/kind cleanup

In https://github.com/kubernetes/apiserver/blob/master/pkg/util/flowcontrol/metrics/sample_and_watermark.go today there is the abstraction of sampling a variable to populate a histogram.  It has an awkward bit of code: a loop to call the `Observe` method of a Prometheus Histogram a variable number of times.  This led to the suggestion https://github.com/prometheus/client_golang/issues/796 .  The discussion there suggested I create a new Prometheus abstraction, possibly out of (prometheus) tree.  This PR is my start on such a creation, indeed out of the prometheus tree.

There are three bits of tension here:
1. The support for "vectors" of metrics in the Prometheus golang library is not public --- and it is non-trivial.  This makes doing this work out of the prometheus tree problematic.
1. `sampling-histogram_test.go` uses fake clocks.  That's the only reasonable way to get sufficient control in a test that runs quickly.  But the Prometheus tree does not include such clocks, so moving this code into the Prometheus tree would be problematic.
1. `sample_and_watermark.go` traffics in _passive_ clocks (and with good reason: to support event-based testing of API Priority and Fairness code), but k8s.io/utils/clock does not define passive clocks.

Fixes https://github.com/prometheus/client_golang/issues/796

For the moment I put a personal copyright on the code that could go in either Kubernetes or Prometheus, depending on how the relevant tensions are relaxed.

```release-note
NONE
```

```docs

```

### DIFF
--- a/build/tools.go
+++ b/build/tools.go
@@ -29,4 +29,5 @@ import (
 	_ "k8s.io/gengo/examples/import-boss/generators"
 	_ "k8s.io/gengo/examples/set-gen/generators"
 	_ "k8s.io/kube-openapi/cmd/openapi-gen"
+	_ "k8s.io/utils/clock/testing"
 )
--- a/staging/src/k8s.io/component-base/metrics/BUILD
+++ b/staging/src/k8s.io/component-base/metrics/BUILD
@@ -27,7 +27,7 @@ go_library(
     visibility = ["//visibility:public"],
     deps = [
         "//staging/src/k8s.io/apimachinery/pkg/version:go_default_library",
-        "//staging/src/k8s.io/component-base/metrics/prometheus_extension:go_default_library",
+        "//staging/src/k8s.io/component-base/metrics/prometheusextension:go_default_library",
         "//staging/src/k8s.io/component-base/version:go_default_library",
         "//vendor/github.com/blang/semver:go_default_library",
         "//vendor/github.com/prometheus/client_golang/prometheus:go_default_library",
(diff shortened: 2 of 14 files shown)


MikeSpreitzer:
This is waiting on https://github.com/prometheus/client_golang/pull/803 to merge.

MikeSpreitzer:
https://github.com/kubernetes/utils/pull/190 is going to be important

beorn7:
https://github.com/prometheus/client_golang/pull/803 is merged now. Sorry for the delay. I'll cut a release v1.8.0 of prometheus/client_golang today so you don't need to lock to a particular commit in your `go.mod` file anymore.

If you need help with what I [said about implementing a more efficient sampling histogram](https://github.com/prometheus/client_golang/issues/796#issuecomment-692218541), let me know. For example, I could create a PR on top of this, once it is merged (or even before, if you prefer that).

beorn7:
As this has gone stale now, let me repeat my offer above to help implementing [my suggestion](https://github.com/prometheus/client_golang/issues/796#issuecomment-692218541). Let me know if that's something you would appreciate.

MikeSpreitzer:
/remove-lifecycle rotten
Yes, @beorn7 , thanks for your offer, I am interested.  Just a little swamped now and slow on getting things done.

beorn7:
OK, I'll try to propose something in the next few days.

beorn7:
@MikeSpreitzer  I created a [PR against your fork](https://github.com/MikeSpreitzer/kubernetes/pull/2) to demonstrate my proposal.

PR #63303, opened 2018-04-30
Return attach error to A/D controller.

jsafrane:
The plugin should not wait with errors for WaitForAttach, WaitForAttach runs on nodes, not in A/D controller.

**Which issue(s) this PR fixes**
Fixes #63300

~~Work in progress, missing testing.~~

**Release note**:

```release-note
Fixed error reporting of CSI volumes attachment.
```

/sig storage
@saad-ali @vladimirvivien

PR #94674, opened 2020-09-10
[Flaky unit test]TestCSI_VolumeAll detach must wait for attach done

### LINKED ISSUE #94619: [flaky unit test] TestCSI_VolumeAll/PersistentVolume_with_driver_info
**Which jobs are flaking**:

pull-kubernetes-bazel-test

TestCSI_VolumeAll/PersistentVolume_with_driver_info

**Testgrid link**:

https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-bazel-test&include-filter-by-regex=csi%2Fgo_default_test

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93305/pull-kubernetes-bazel-test/1303342948955459585

**Reason for failure**:

Can't really tell

/sig storage

zhouya0:
**What type of PR is this?**
/kind flake

As error log show in https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-bazel-test/1303342948955459585:
```
    csi_test.go:558: csiTest.VolumeAll preparing detacher.Detach...
E0908 15:00:03.081837      23 csi_attacher.go:632] kubernetes.io/csi: VolumeAttachment [vol2] has been deleted, will not continue to wait for attachment
    csi_test.go:310: csiTest.VolumeAll attacher.Attach failed: volume attachment has been deleted
    csi_test.go:568: csiTest.VolumeAll detacher.Detach succeeded for volume simple-driver^vol2
```

It comes from the goroutine in `csi_test.go`https://github.com/kubernetes/kubernetes/blob/88e3c95d7c46431e0842e0fdc2cb99279c038515/pkg/volume/csi/csi_test.go#L307-L316 
Which will call `waitForVolumeAttachment` in `Attach` function:
https://github.com/kubernetes/kubernetes/blob/88e3c95d7c46431e0842e0fdc2cb99279c038515/pkg/volume/csi/csi_attacher.go#L124-L125

It obvious that `Detach` happens before  in this case: https://github.com/kubernetes/kubernetes/blob/bd39d3933be27d70bf27067fc494194be21f8ff4/pkg/volume/csi/csi_test.go#L565

So the reason for flaky is:
If `Attach` goroutine runs slow and before it runs to `waitForVolumeAttachment`, `Detach` has already done, then `Attach` function will reture error.

So my change here is to wait `Attach` is done before `Detach`.

By the way, before `Attach` will not report error in its wait logic until this [PR](https://github.com/kubernetes/kubernetes/pull/63303).

Fixes https://github.com/kubernetes/kubernetes/issues/94619

```release-note
None
```

### DIFF
--- a/pkg/volume/csi/csi_test.go
+++ b/pkg/volume/csi/csi_test.go
@@ -293,6 +293,7 @@ func TestCSI_VolumeAll(t *testing.T) {
 			}
 
 			var devicePath string
+			waitForAttachDone := make(chan interface{})
 			if attachPlug != nil {
 				t.Log("csiTest.VolumeAll attacher.Attach starting")
 
@@ -306,6 +307,7 @@ func TestCSI_VolumeAll(t *testing.T) {
 				// creates VolumeAttachment and blocks until it is marked attached (done by external attacher)
 				go func(spec *volume.Spec, nodeName types.NodeName) {
 					attachID, err := volAttacher.Attach(spec, nodeName)
+					close(waitForAttachDone)
 					if err != nil {
 						t.Errorf("csiTest.VolumeAll attacher.Attach failed: %s", err)
 						return
@@ -316,6 +318,7 @@ func TestCSI_VolumeAll(t *testing.T) {
 
 				// Simulates external-attacher and marks VolumeAttachment.Status.Attached = true
 				markVolumeAttached(t, host.GetKubeClient(), fakeWatcher, attachName, storage.VolumeAttachmentStatus{Attached: true})
+				<-waitForAttachDone
 
 				devicePath, err = volAttacher.WaitForAttach(volSpec, "", pod, 500*time.Millisecond)
 				if err != nil {


zhouya0:
/hold 
Hold for now.

pkg/volume/csi/csi_test.go
```diff
@@ -306,6 +307,7 @@ func TestCSI_VolumeAll(t *testing.T) {
 				// creates VolumeAttachment and blocks until it is marked attached (done by external attacher)
 				go func(spec *volume.Spec, nodeName types.NodeName) {
 					attachID, err := volAttacher.Attach(spec, nodeName)
+					waitForAttachDone <- attachID
```
JornShen, line 310: close(waitForAttachDone)  may be more good， my opinion. :)

zhouya0: Thanks! Sorry I forget to do that :)

pkg/volume/csi/csi_test.go
```diff
@@ -293,6 +293,8 @@ func TestCSI_VolumeAll(t *testing.T) {
 			}
 
 			var devicePath string
+			waitForAttachDone := make(chan string)
+			defer close(waitForAttachDone)
```
JornShen, line 297: Hey, It 's not suitable here to call defer. Usually we put defer on outside, in main process.  But here the variable waitForAttachDone is in the for loop(line 222). I understand, waitForAttachDone will be assign in every loop. So this code mean it will set every time. It 's very easy to case Ambiguity.

pkg/volume/csi/csi_test.go
JornShen: It seems that you don't understand me. I still advise to use the close here. Because it's no blocking. This routine doesn't need to wait the others to receive it. On the other side, can you promise that the code in 568 line can execute?  it has a if judge before in line 555.if the code doesn't execute, this routine will block forever. So when we close it here, you don't worry about block scene and also you can  promise that Detach execute after the Attach. :)

pkg/volume/csi/csi_test.go
```diff
@@ -562,6 +565,7 @@ func TestCSI_VolumeAll(t *testing.T) {
 				}
 				csiDetacher := volDetacher.(*csiAttacher)
 				csiDetacher.csiClient = csiClient
+				<-waitForAttachDone
```
JornShen, line 568: if waitForAttachDone has been closed, the code will skip. And also means that Attach has been executed. :)

JornShen (commented):
My review opinion,  welcome to correct to me. :)

pkg/volume/csi/csi_test.go
JornShen: https://github.com/golang/go/blob/9214677e7df1e6130249bc83d721130b00d829c4/src/runtime/chan.go#L509-L518

pkg/volume/csi/csi_test.go
zhouya0: Yes, you are right. The routine blocked indeed.
I hope I can be more careful next time:(  .Thanks, again.

pkg/volume/csi/csi_test.go
zhouya0: Done :0

pkg/volume/csi/csi_test.go
JornShen: it can be removed. I don't explain it clearly. sorry!  For the most start,  I advise that this line can be changed to `close(waitForAttachDone)`.

pkg/volume/csi/csi_test.go
JornShen: Well done! 👍

zhouya0:
/priority backlog
/retest
@hasheddan could you please review this flaky?

pkg/volume/csi/csi_test.go
```diff
@@ -306,6 +307,7 @@ func TestCSI_VolumeAll(t *testing.T) {
 				// creates VolumeAttachment and blocks until it is marked attached (done by external attacher)
 				go func(spec *volume.Spec, nodeName types.NodeName) {
 					attachID, err := volAttacher.Attach(spec, nodeName)
+					close(waitForAttachDone)
```
hasheddan, line 310: Do we want to close before checking error?

liggitt: yes, we don't want to hang the test in case of an error

pkg/volume/csi/csi_test.go
```diff
@@ -562,6 +564,7 @@ func TestCSI_VolumeAll(t *testing.T) {
 				}
 				csiDetacher := volDetacher.(*csiAttacher)
 				csiDetacher.csiClient = csiClient
+				<-waitForAttachDone
```
liggitt, line 567: does the WaitForAttach call on line 322 not guarantee the attach was done by the time we got here?

liggitt:
cc @mikedanese for similar PR (though waiting in a different place) - #94735

pkg/volume/csi/csi_test.go
hasheddan: Ah yes my confusion was around thinking this was in the main thread of execution here, which would mean that if the encapsulating function returned first that the routine would be terminated. Not sure what the odds of that even happening would be, but this goroutine is not encapsulated by the main thread of execution. However, let's say the last test case was being run and we hit an error here, could we feasibly exit the encapsulating function _and_ the main thread of execution prior to calling `t.Errorf`, thus causing the test to pass (assuming no other errors)?

pkg/volume/csi/csi_test.go
zhouya0: Yes, that's better. Just first I'am trying to solve `Detach` problem so I move it to line 567. Now I moved it to line 322.

zhouya0:
Thanks for you guys reviewing. But since https://github.com/kubernetes/kubernetes/pull/94735 is merged. I'll close this PR.

PR #94685, opened 2020-09-10
Convert to EphemeralContainers for validation in EphemeralContainersREST

### LINKED ISSUE #92557: The presence of admission validating webhooks breaks Ephemeral Containers
**What happened**:
Adding a validating webhook with resources section of `"pods/*"` or `"*/*"` breaks ability to add Ephemeral Containers as described in https://kubernetes.io/docs/tasks/debug-application-cluster/debug-running-pod/#ephemeral-container

```
kubectl alpha debug -it rabbit --image=busybox --target=rabbitmq
```
Debug command should attach to the running pod and allow user to interact with the ephemeral container image. This works when no validating webhook is configured.

Using a K8 cluster with Ephemeral Containers feature gate enabled and v1.18.3
1. Create a validating webhook (contents don't matter except that resources is "pods/*")
<details>
  <summary>whconfig.yaml</summary>

  ```
apiVersion: admissionregistration.k8s.io/v1beta1
kind: ValidatingWebhookConfiguration
metadata:
  name: test-repro
webhooks:
- name: test-repro.test.k8s.io
  clientConfig:
    url: https://localhost:8675/hook
  rules:
  - apiGroups: ["*"]
    apiVersions: ["*"]
    operations: ["*"]
    resources: ["pods/*"]
```
</details>

```
kubectl apply -f whconfig.yaml
```

2. Create a test pod
<details>
  <summary>test_pod.yaml</summary>

  ```
apiVersion: v1
kind: Pod
metadata:
  name: rabbit
  labels:
    role: myrole
spec:
  shareProcessNamespace: true
  containers:
    - name: rabbitmq
      image: rabbitmq
      ports:
        - name: rabbit
          containerPort: 5672
          protocol: TCP
```
</details>

```
kubectl apply -f test_pod.yaml
```

3. Try to add an Ephemeral Container using alpha debug command
```
kubectl alpha debug -it rabbit --image=busybox --target=rabbitmq
```
Error:
```
$ kubectl alpha debug -it rabbit --image=busybox --target=rabbitmq
Defaulting debug container name to debugger-jvc8h.
error: error updating ephemeral containers: Internal error occurred: converting (core.Pod) to (v1.EphemeralContainers): EphemeralContainers not present in src
```

When using `v=10`, can see that the api server appears to be throwing the error when trying to convert the object for the webhook invocation.
API returning the 500 error: `http://localhost:8080/api/v1/namespaces/default/pods/rabbit/ephemeralcontainers`

**Environment**:
- Cloud provider or hardware configuration: Minikube
- OS (e.g: `cat /etc/os-release`):  MacOS
- Kernel (e.g. `uname -a`): 
- Install tools:
- Others:

verb:
**What type of PR is this?**
/kind bug
/sig node

When updating ephemeral containers, convert Pod to EphemeralContainers in storage validation. This resolves a bug where admission webhook validation fails for ephemeral container updates because the webhook client cannot perform the conversion.

Also enable the EphemeralContainers feature gate for the admission control integration test, which would have caught this bug.

Fixes #92557

```release-note
Fixed a bug that prevents the use of ephemeral containers in the presence of a validating admission webhook.
```

```docs

```

### DIFF
--- a/pkg/registry/core/pod/storage/storage.go
+++ b/pkg/registry/core/pod/storage/storage.go
@@ -346,11 +346,24 @@ func (r *EphemeralContainersREST) Update(ctx context.Context, name string, objIn
 		return newPod, nil
 	})
 
-	obj, _, err = r.store.Update(ctx, name, updatedPodInfo, createValidation, updateValidation, false, options)
+	// Validation should be passed the API kind (EphemeralContainers) rather than the storage kind.
+	obj, _, err = r.store.Update(ctx, name, updatedPodInfo, toEphemeralContainersCreateValidation(createValidation), toEphemeralContainersUpdateValidation(updateValidation), false, options)
 	if err != nil {
 		return nil, false, err
 	}
 	return ephemeralContainersInPod(obj.(*api.Pod)), false, err
+}
+
+func toEphemeralContainersCreateValidation(f rest.ValidateObjectFunc) rest.ValidateObjectFunc {
+	return func(ctx context.Context, obj runtime.Object) error {
+		return f(ctx, ephemeralContainersInPod(obj.(*api.Pod)))
+	}
+}
+
+func toEphemeralContainersUpdateValidation(f rest.ValidateObjectUpdateFunc) rest.ValidateObjectUpdateFunc {
+	return func(ctx context.Context, obj, old runtime.Object) error {
+		return f(ctx, ephemeralContainersInPod(obj.(*api.Pod)), ephemeralContainersInPod(old.(*api.Pod)))
+	}
 }
 
 // Extract the list of Ephemeral Containers from a Pod
--- a/test/integration/apiserver/admissionwebhook/admission_test.go
+++ b/test/integration/apiserver/admissionwebhook/admission_test.go
@@ -480,6 +480,8 @@ func testWebhookAdmission(t *testing.T, watchCache bool) {
 		"--disable-admission-plugins=ServiceAccount,StorageObjectInUseProtection",
 		// force enable all resources so we can check storage.
 		"--runtime-config=api/all=true",
+		// enable feature-gates that protect resources to check their storage, too.
+		"--feature-gates=EphemeralContainers=true",
 	}, etcdConfig)
 	defer server.TearDownFn()
 


pkg/registry/core/pod/storage/storage.go
```diff
@@ -346,13 +346,26 @@ func (r *EphemeralContainersREST) Update(ctx context.Context, name string, objIn
...
+
+func toEphemeralContainersUpdateValidation(f rest.ValidateObjectUpdateFunc) rest.ValidateObjectUpdateFunc {
+	return func(ctx context.Context, obj, old runtime.Object) error {
+		return f(ctx, ephemeralContainersInPod(obj.(*api.Pod)), ephemeralContainersInPod(old.(*api.Pod)))
+	}
+}
+
 // Extract the list of Ephemeral Containers from a Pod
 func ephemeralContainersInPod(pod *api.Pod) *api.EphemeralContainers {
 	ephemeralContainers := pod.Spec.EphemeralContainers
```
liggitt, line 371: is there a reason the entire pod metadata isn't being copied, like this?
```
ObjectMeta: pod.ObjectMeta,
```

verb: I copied this behavior from scale:
 https://github.com/kubernetes/kubernetes/blob/f99ed92ed149e8cbb09dfcdcff78e6fb4024d3a0/pkg/registry/apps/deployment/storage/storage.go#L334 where there's a comment `TODO: Create a variant of ObjectMeta type that only contains the fields below.`

I'm not sure of the original reason, but I'm happy to experiment here.

liggitt: hmm... ok. the labels and annotations seem most likely to be of interest. managedFields would not be coherent if copied from the pod metadata.

test/integration/apiserver/admissionwebhook/admission_test.go
```diff
@@ -480,6 +480,8 @@ func testWebhookAdmission(t *testing.T, watchCache bool) {
 		"--disable-admission-plugins=ServiceAccount,StorageObjectInUseProtection",
 		// force enable all resources so we can check storage.
 		"--runtime-config=api/all=true",
+		// enable feature-gates that protect resources to check their storage, too.
+		"--feature-gates=EphemeralContainers=true",
```
liggitt, line 484: just to verify, adding this made the test fail until the conversion was done, right?

verb: That's correct, enabling the feature gate made the test fail without the conversion, which I thought was impressive.

pkg/registry/core/pod/storage/storage.go
verb: We could copy the entire ObjectMeta and prune out fields that don't make sense. I guess it's not uncommon to derive `ObjectMeta`, so maybe there should be a helper for it.

verb: @liggitt shall we submit this as-is to fix the bug and open a follow-up issue?

liggitt: yes, follow-up for considering what we want to do with this bit is fine.

PR #94687, opened 2020-09-10
Non-zero cfs quota period duration requires feature flag

### LINKED ISSUE #94686: CustomCPUCFSQuotaPeriod feature is required for using cpuCFSQuotaPeriod in kubelet config
See https://github.com/kubernetes/kubernetes/issues/79833 for more context.

**What happened**:

When you set `cpuCFSQuota: true` and `cpuCFSQuotaPeriod: 10ms`, you'd expect container cfs quota period to be 10000. However, unless you also set `CustomCPUCFSQuotaPeriod: true` in the feature flags, the quota value will be silently ignored.

At least a warning be logged if the feature is not enabled but `cpuCFSQuotaPeriod` is set.

https://github.com/kubernetes/kubernetes/issues/79833#issue-464575246

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

karan:
**What type of PR is this?**

/kind bug

Validates that the feature flag is set if `CPUCFSQuotaPeriod` is specified in kubelet config.

Fixes #94686

`CPUCFSQuotaPeriod` is defaulted to 100ms. So if that's set, then we do not enforce this validation.

```release-note
Require feature flag CustomCPUCFSQuotaPeriod if setting a non-default cpuCFSQuotaPeriod in kubelet config.
```

### DIFF
--- a/pkg/kubelet/apis/config/validation/validation.go
+++ b/pkg/kubelet/apis/config/validation/validation.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilvalidation "k8s.io/apimachinery/pkg/util/validation"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
@@ -29,6 +30,10 @@ import (
 	kubeletconfig "k8s.io/kubernetes/pkg/kubelet/apis/config"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
 	kubetypes "k8s.io/kubernetes/pkg/kubelet/types"
+)
+
+var (
+	defaultCFSQuota = metav1.Duration{Duration: 100 * time.Millisecond}
 )
 
 // ValidateKubeletConfiguration validates `kc` and returns an error if it is invalid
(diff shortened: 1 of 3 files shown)


pkg/kubelet/apis/config/validation/validation.go
```diff
@@ -60,6 +65,9 @@ func ValidateKubeletConfiguration(kc *kubeletconfig.KubeletConfiguration) error
 	if kc.HealthzPort != 0 && utilvalidation.IsValidPortNum(int(kc.HealthzPort)) != nil {
 		allErrors = append(allErrors, fmt.Errorf("invalid configuration: HealthzPort (--healthz-port) %v must be between 1 and 65535, inclusive", kc.HealthzPort))
 	}
+	if !localFeatureGate.Enabled(features.CPUCFSQuotaPeriod) && (kc.CPUCFSQuotaPeriod != zeroDuration && kc.CPUCFSQuotaPeriod != metav1.Duration{Duration: 100 * time.Millisecond}) {
```
mtaufen, line 68: nit: don't think we need the extra parens around `kc.CPUCFSQuotaPeriod != zeroDuration && kc.CPUCFSQuotaPeriod != metav1.Duration{Duration: 100 * time.Millisecond}`

pkg/kubelet/apis/config/validation/validation.go
```diff
@@ -31,6 +32,10 @@ import (
 	kubetypes "k8s.io/kubernetes/pkg/kubelet/types"
 )
 
+var (
+	zeroDuration = metav1.Duration{}
```
mtaufen, line 36: Is this already defined elsewhere?

pkg/kubelet/apis/config/validation/validation_test.go
```diff
@@ -115,9 +121,9 @@ func TestValidateKubeletConfiguration(t *testing.T) {
 		RegistryPullQPS:             -10,
 		HairpinMode:                 "foo",
 		NodeLeaseDurationSeconds:    -1,
-		CPUCFSQuotaPeriod:           metav1.Duration{Duration: 0},
+		CPUCFSQuotaPeriod:           metav1.Duration{Duration: 100},
```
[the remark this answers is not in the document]
karan: I made it explicitly 100 ms. This is to test the case where the default value is specified and so the validation should succeed.

pkg/kubelet/apis/config/validation/validation.go
karan: Yeah I put it there for my clarity. Removed it.

pkg/kubelet/apis/config/validation/validation.go
karan: A couple of places but it's local to those packages. It's the default value so I didn't want to introduce another dependency.

mtaufen: k sg

pkg/kubelet/apis/config/validation/validation.go
```diff
@@ -60,6 +66,9 @@ func ValidateKubeletConfiguration(kc *kubeletconfig.KubeletConfiguration) error
 	if kc.HealthzPort != 0 && utilvalidation.IsValidPortNum(int(kc.HealthzPort)) != nil {
 		allErrors = append(allErrors, fmt.Errorf("invalid configuration: HealthzPort (--healthz-port) %v must be between 1 and 65535, inclusive", kc.HealthzPort))
 	}
+	if !localFeatureGate.Enabled(features.CPUCFSQuotaPeriod) && kc.CPUCFSQuotaPeriod != zeroDuration && kc.CPUCFSQuotaPeriod != defaultCFSQuota {
```
mtaufen, line 69: Thinking about this a little more, if 100ms is the default, then why is zeroDuration an exception to the validation?

karan: True - that's also covered by the next validation anyways. Will remove that check.

PR #94689, opened 2020-09-10
Default converters should not be dropped in 1.19

### LINKED ISSUE #94688: Can no longer call DecodeParameters with url.Values in 1.19.0 client-go
**What happened**:
The following code works in 1.18 but is broken in 1.19.0
```go
import (
	"k8s.io/client-go/kubernetes/scheme"
	corev1 "k8s.io/api/core/v1"
)
...
       	target := &corev1.PodLogOptions{}
	return scheme.ParameterCodec.DecodeParameters(req.URL.Query(), corev1.SchemeGroupVersion, target)
```

This results in the error: `converting (url.Values) to (v1.PodLogOptions): unknown conversion`

This seems to be caused by this line which seems like a mistake because it's an early exit and the comment right below indicates this should be removed in 1.20.  https://github.com/kubernetes/kubernetes/blob/v1.19.0/staging/src/k8s.io/apimachinery/pkg/conversion/converter.go#L454

Is there a different way to decode PodLogOptions?  If you remove the the default converter I don't understand how I'm supposed to do this anymore because there is no converter registered for url.Values that I'm aware of. 

**Environment**:
- Kubernetes version (use `kubectl version`): 1.19.0 client-go/api

ibuildthecloud:
Signed-off-by: Darren Shepherd <darren@rancher.com>

**What type of PR is this?**
/kind bug
/kind regression

It seems the default converters were unintentionally dropped in 1.19.

Fixes #94688

Sorry if this is the wrong way to submit a PR for 1.19. This change is not applicable for master so I'm not putting in a PR to master that can be cherry picked.

```release-note

```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/conversion/converter.go
+++ b/staging/src/k8s.io/apimachinery/pkg/conversion/converter.go
@@ -451,7 +451,6 @@ func (c *Converter) doConversion(src, dest interface{}, flags FieldMatchingFlags
 	if err != nil {
 		return err
 	}
-	return fmt.Errorf("converting (%s) to (%s): unknown conversion", sv.Type(), dv.Type())
 
 	// TODO: Everything past this point is deprecated.
 	//  Remove in 1.20 once we're sure it didn't break anything.


lavalamp:
/assign @caesarxuchao @sttts 

Why isn't this change applicable to 1.20/head?

dims:
@wojtek-t can you please peek? looks like this regression was introduced in https://github.com/kubernetes/kubernetes/commit/8c1f9616cf2875cac9489df23eed3f7f849d2e08

@lavalamp a subsequent fix in master seems to have cleaned up after ^^^
https://github.com/kubernetes/kubernetes/commit/410d575d4219ee46c3726d31acf86e555f0a5f1b

wojtek-t:
/hold

https://github.com/kubernetes/kubernetes/issues/94688#issuecomment-690898910

caesarxuchao:
Seems like we should close this one. @ibuildthecloud, I'll leave it to you.

/unassign

wojtek-t:
Closing based on a discussion on the #94688

Issue #94437, opened 2020-09-02
Missing a single update message will cause the scheduler never able to schedule pods to the right nodes


srteam2020:
**What happened**:
If the scheduler misses a single update message "node delete", then it won't be able to know the node is deleted and still tries to schedule some pods to the deleted node, which will make those pods never able to be scheduled to the right nodes. kubernetes/kubernetes#56261 should be related but somehow the issue still happens.

In a k8s cluster with one apiserver, one scheduler and two workers (worker1 and worker2), after deleting a worker node from the cluster (by `kubectl delete node worker1`), the scheduler should hear from apirserver that “worker1 is deleted” by the node informer. However, if the “worker1 is deleted” update is missed by the scheduler due to some network issue (which means `deleteNodeFromCache` won’t be called), then the scheduler won’t know that worker1 is deleted. When we create pods in the cluster, the scheduler will keep trying to schedule some of the pods to the non-existing worker1, and these pods will never be running successfully.

**What you expected to happen**:
The scheduler should be able to tell that "worker1" doesn't exist any more after several retrying to schedule pods to "worker1" but fails.

**How to reproduce it (as minimally and precisely as possible)**:
The simplest way to reproduce this issue is to use kind on a single machine.
1. Start a cluster with one cp node (`kind-control-plane`) and two worker nodes (`kind-worker` and `kind-worker2`)
2. Delete one worker node (`kubectl delete node kind-worker`), and now we only have `kind-control-plane` and `kind-worker2` as show here:
    ```
    NAME                 STATUS   ROLES    AGE   VERSION
    kind-control-plane   Ready    master   32m   v1.18.0
    kind-worker2         Ready    <none>   31m   v1.18.0
    ```
3. Drop the “deleted” message from apiserver to scheduler by setting the iptable rules or instrumenting the k8s src code to make the scheduler "unaware" of the "delete" message. For example, we can do this by simply skipping `r.store.Delete(event.Object)` for "node delete" event in `reflector.go` or  removing the `deleteNodeFromCache` like below (both are fine):
    ```
    diff --git a/pkg/scheduler/eventhandlers.go b/pkg/scheduler/eventhandlers.go
    index f6d8bb1c..d5ef9e5d 100644
    --- a/pkg/scheduler/eventhandlers.go
    +++ b/pkg/scheduler/eventhandlers.go
    @@ -407,7 +407,6 @@ func addAllEventHandlers(
                    cache.ResourceEventHandlerFuncs{
[...]

PR #94692, opened 2020-09-10
Wrap errors from VolumeBinding and DefaultBinder plugins

alculquicondor:
**What type of PR is this?**

/kind cleanup

/kind regression

When using framework's Status, we recreate errors from messages. This loses error information necessary for error handling. In particular #94437 raised the concern that we are missing deleted nodes errors which were used to for a node deletion.

This is an outcome of the migration to scheduler framework.

In this patch, we fix the PreBind and Bind framework runtime to wrap the errors in a Status. We fix the DefaultBinder and VolumeBinding plugins to wrap API errors.

Ref #94437

This is a minimal patch that can be potentially patched to 1.19 and 1.18. In particular, there are several unwrapped errors (for extension points that don't access the API) that are left because they require significantly more refactoring.

They will be fixed in follow ups.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/volume/scheduling/scheduler_binder.go
+++ b/pkg/controller/volume/scheduling/scheduler_binder.go
@@ -443,7 +443,7 @@ func (b *volumeBinder) BindPodVolumes(assumedPod *v1.Pod, podVolumes *PodVolumes
 		return b, err
 	})
 	if err != nil {
-		return fmt.Errorf("Failed to bind volumes: %v", err)
+		return fmt.Errorf("binding volumes: %w", err)
 	}
 	return nil
 }
@@ -543,7 +543,7 @@ func (b *volumeBinder) checkBindings(pod *v1.Pod, bindings []*BindingInfo, claim
 
 	node, err := b.nodeLister.Get(pod.Spec.NodeName)
 	if err != nil {
-		return false, fmt.Errorf("failed to get node %q: %v", pod.Spec.NodeName, err)
+		return false, fmt.Errorf("failed to get node %q: %w", pod.Spec.NodeName, err)
 	}
 
 	csiNode, err := b.csiNodeLister.Get(node.Name)
@@ -559,7 +559,7 @@ func (b *volumeBinder) checkBindings(pod *v1.Pod, bindings []*BindingInfo, claim
 	_, err = b.podLister.Pods(pod.Namespace).Get(pod.Name)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			return false, fmt.Errorf("pod %q does not exist any more", podName)
+			return false, fmt.Errorf("pod does not exist any more: %w", err)
 		}
 		klog.Errorf("failed to get pod %s/%s from the lister: %v", pod.Namespace, pod.Name, err)
 	}
@@ -567,12 +567,12 @@ func (b *volumeBinder) checkBindings(pod *v1.Pod, bindings []*BindingInfo, claim
 	for _, binding := range bindings {
 		pv, err := b.pvCache.GetAPIPV(binding.pv.Name)
 		if err != nil {
-			return false, fmt.Errorf("failed to check binding: %v", err)
+			return false, fmt.Errorf("failed to check binding: %w", err)
 		}
 
 		pvc, err := b.pvcCache.GetAPIPVC(getPVCName(binding.pvc))
 		if err != nil {
-			return false, fmt.Errorf("failed to check binding: %v", err)
+			return false, fmt.Errorf("failed to check binding: %w", err)
 		}
 
 		// Because we updated PV in apiserver, skip if API object is older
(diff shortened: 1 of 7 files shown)


pkg/scheduler/scheduler_test.go
```diff
@@ -211,6 +211,7 @@ func TestSchedulerScheduleOne(t *testing.T) {
 	eventBroadcaster := events.NewBroadcaster(&events.EventSinkImpl{Interface: client.EventsV1()})
 	errS := errors.New("scheduler")
 	errB := errors.New("binder")
+	preBindErr := errors.New("pre bind error")
```
ahg-g, line 214:
```suggestion
	preBindErr := errors.New("PreBind error")
```

ahg-g (commented):
/lgtm

pkg/controller/volume/scheduling/scheduler_binder.go
```diff
@@ -559,20 +559,20 @@ func (b *volumeBinder) checkBindings(pod *v1.Pod, bindings []*BindingInfo, claim
 	_, err = b.podLister.Pods(pod.Namespace).Get(pod.Name)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			return false, fmt.Errorf("pod %q does not exist any more", podName)
+			return false, fmt.Errorf("pod does not exist any more: %w", err)
```
adtac, line 562: keep podName as a part of the message? or does having that as a part of the string cause unwrapping issues?

alculquicondor: the error from the API already includes the pod name. Other than that, we should add the metadata to the error when we are logging it. See #94697

pkg/scheduler/scheduler_test.go
alculquicondor: `errors.New("on PreBind")`? errors shouldn't start with capital letters.

pkg/controller/volume/scheduling/scheduler_binder.go
```diff
@@ -646,7 +646,7 @@ func (b *volumeBinder) checkBindings(pod *v1.Pod, bindings []*BindingInfo, claim
 			}
 
 			if err := volumeutil.CheckNodeAffinity(pv, node.Labels); err != nil {
-				return false, fmt.Errorf("pv %q node affinity doesn't match node %q: %v", pv.Name, node.Name, err)
+				return false, fmt.Errorf("pv %q node affinity doesn't match node %q: %w", pv.Name, node.Name, err)
 			}
 		}
```
Huang-Wei, line 652: (comment not related with this line)

Do you want to cleanup all `%v.*err` occurrences?

```
⇒  ag "%v.*err" pkg/controller/volume/scheduling/scheduler_binder.go
367:		klog.V(5).Infof("AssumePodVolumes: GetBindVolumeToClaim for pod %q, PV %q, PVC %q.  newPV %p, dirty %v, err: %v",
496:			klog.V(4).Infof("updating PersistentVolume[%s]: binding to %q failed: %v", binding.pv.Name, claimKey, err)
552:		klog.V(4).Infof("Could not get a CSINode object for the node %q: %v", node.Name, err)
564:		klog.Errorf("failed to get pod %s/%s from the lister: %v", pod.Namespace, pod.Name, err)
698:		return false, nil, fmt.Errorf("error getting PVC %q: %v", pvcKey, err)
769:		klog.V(4).Infof("Could not get a CSINode object for the node %q: %v", node.Name, err)
786:			klog.V(4).Infof("PersistentVolume %q, Node %q mismatch for Pod %q: %v", pvName, node.Name, podName, err)
964:		klog.Errorf("unexpected error converting %+v to a label selector: %v", capacity.NodeTopology, err)
1037:		return nil, fmt.Errorf("could not get plugin name from pv: %v", err)
1050:		return nil, fmt.Errorf("could not translate pv: %v", err)
```

pkg/scheduler/scheduler_test.go
ahg-g: sg

pkg/controller/volume/scheduling/scheduler_binder.go
alculquicondor: I want to keep this PR as small as possible (although I couldn't contain myself from doing some cleanups) just in case we need to cherry pick. I opened #94695 for the rest.

pkg/scheduler/scheduler_test.go
alculquicondor: Done

cofyc:
/approve
for pkg/controller/volume/scheduling

alculquicondor:
@ahg-g for milestone if deemed appropriate

ahg-g:
/milestone v1.20

yup, cleanup PRs like this are eligible.

alculquicondor:
I still need lgtm :)

PR #59283, opened 2018-02-02
bzl: make integration tests actually work

mikedanese:
#59283
```release-note
NONE
```

PR #45855, opened 2017-05-15
move build related files out of the root directory

mikedanese:
Turn the files in root to symlinks into build/root/

alternative to OWNERS changes. what do you guys think of this?

PR #41280, opened 2017-02-11
Add verify-gofmt as a Bazel test.

spxtr:
Also, moved `--flaky_test_attempts=3` out of `.bazelrc` since it was annoying. I think this verify-all pattern is working well in test-infra, we should try and make it happen here.

**Release note**:
```release-note
NONE
```

PR #94699, opened 2020-09-10
Run unit tests 2 instead of 3 times via bazel

spiffxp:
**What type of PR is this?**

/kind cleanup
/kind flake

Followup to https://github.com/kubernetes/kubernetes/pull/93605 which changed from retryng 3 times to avoid flakes, to requiring a pass 3 times.

In order to migrate the pull-kubernetes-bazel-test job to a communty owned cluster with dedicated resources, we need to stop using RBE.  When we did so, we started seeing many more flakes. (ref: https://github.com/kubernetes/test-infra/issues/19070#issuecomment-690348770)

RBE fans out to a large pool of compute, whereas now we're running in a single pod on a single node.

It may be that using 2 runs instead of 3 will relieve some resource contention, giving us an appropriate compromise between not ignoring flakes but not leaning all the way into causing them. (ref: https://github.com/kubernetes/test-infra/issues/19070#issuecomment-690354428)

n/a

/hold
I'd like to see if the behavior of pull-kubernetes-bazel-test improves now that we've given it 7 CPUs instead of 4. I'm teeing this up as a further mitigation if we decide that's not enough, or if 7 CPUs is excessive resource consumption.

```release-note
NONE
```

/cc @liggitt @BenTheElder

### DIFF
--- a/build/root/.bazelrc
+++ b/build/root/.bazelrc
@@ -28,7 +28,7 @@ test:unit --features=race
 
 test:unit --build_tests_only
 test:unit --test_tag_filters=-e2e,-integration
-test:unit --runs_per_test=3
+test:unit --runs_per_test=2
 
 test:integration --local_test_jobs 4
 test:integration --test_tag_filters=integration


liggitt:
> 7 CPUs is excessive resource consumption

7 CPUs to run unit tests feels excessive. If we are not actually caching unit test results from unaffected packages and are running all tests 3 times on every PR push, then reducing to 2 runs is reasonable.

liggitt:
/lgtm

unhold at will

BenTheElder:
>7 CPUs to run unit tests feels excessive. If we are not actually caching unit test results from unaffected packages and are running all tests 3 times on every PR push, then reducing to 2 runs is reasonable.

cached-ness is easily confirmed in the log output.

a line like:
> //pkg/kubelet/cm/cpumanager:go_default_test                              PASSED in 1.1s

means we ran it. it will say something like `(cached)` at the end otherwise.

spiffxp:
/milestone v1.20
/hold cancel
Yeah I'm going to go ahead and drop us to 2.  These are graphs for the CI job, which uses 7 cpu already.  They're from a local grafana instance that queries against k8s-gubernator:builds.

It's pretty easy to see when https://github.com/kubernetes/kubernetes/pull/93605 landed.  Makes me wonder if the retry_flake_attempts setting was added before the move to RBE
Screen Shot 2020-09-10 at 3 55 01 PM

spiffxp:
(Pulling this out of slack) and yes it turns out if we run multiple times, we can't cache, unless we want to accept potentially caching failures https://docs.bazel.build/versions/master/user-manual.html#flag--cache_test_results

>If this option is set to 'auto' (the default) then Bazel will only rerun a test if any of the following conditions applies:
>- Bazel detects changes in the test or its dependencies
>- the test is marked as external
>- **multiple test runs were requested with --runs_per_test**
>- the test failed.
>
>If 'no', all tests will be executed unconditionally.
>If 'yes', the caching behavior will be the same as auto except that it may cache test failures and test runs with --runs_per_test.

spiffxp:
>Makes me wonder if the retry_flake_attempts setting was added before the move to RBE

https://github.com/kubernetes/kubernetes/pull/59283 added `test:unit --flaky_test_attempts=3` on 2018-02-06

... but it was in build/root/Makefile as bazel flags before that
... before https://github.com/kubernetes/kubernetes/pull/45855 moved Makefile to build/root
... before https://github.com/kubernetes/kubernetes/pull/41280 moved it from .bazelrc to Makefile

So, yeah, it was in the first PR that added .bazelrc on 2017-01-20 (https://github.com/kubernetes/kubernetes/pull/40231).  Definitely before we tried running bazel via RBE

PR #94711, opened 2020-09-11
Bump etcd to dd1b699fc489

jingyih:
**What type of PR is this?**
/kind feature

We recently bumped etcd version to `v3.4.13` in #94259. This PR picks up an additional change (https://github.com/etcd-io/etcd/pull/12280) which is merged right after `v3.4.13`.

We normally only bump etcd to a released tag. However I think it is OK to pick up [dd1b699fc489](https://github.com/etcd-io/etcd/commit/dd1b699fc4895de8cc23c3cac5a428c37eee384a) instead of waiting for the next etcd patch release, because: 

1) we are just picking up 1 change which is in etcd integration test.
2) we unblock #94364 sooner. 

```release-note
NONE
```

/sig api-machinery
/area etcd

cc @kubernetes/release-engineering @wojtek-t @jpbetz @wenjiaswe


wojtek-t:
Thanks @jingyih - that makes sense to me, but I would like @liggitt to share his thoughts.
/assign @liggitt

liggitt:
it's not that big of a deal to pick this up since this is a test-only change... just so I understand, this is in service of unblocking integration tests by a week or so until the next etcd patch release is cut?

jingyih:
> just so I understand, this is in service of unblocking integration tests by a week or so until the next etcd patch release is cut?

Actually, given that in etcd 3.4 branch there is only a test-only change since the last patch release, I am not sure if it makes sense to cut a new patch release now.

liggitt:
/priority important-longterm
/lgtm
/approve

ok, a test-only change that we'll pick up on the next release to unblock dev work early in this cycle isn't troubling to me

PR #94712, opened 2020-09-11
[credentialprovider] avoid potential secret leaking while reading .dockercfg

droslean:
There are a lot of scenarios where an invalid .dockercfg file
will still contain secrets. This commit removes the logging of the
contents to avoid any potential leaking and manages the actual error
by printing to the user the actual location of the invalid file.

/kind design

Signed-off-by: Nikolaos Moraitis <nmoraiti@redhat.com>

### DIFF
--- a/pkg/credentialprovider/config.go
+++ b/pkg/credentialprovider/config.go
@@ -117,10 +117,14 @@ func ReadDockercfgFile(searchPaths []string) (cfg DockerConfig, err error) {
 			continue
 		}
 		cfg, err := readDockerConfigFileFromBytes(contents)
-		if err == nil {
-			klog.V(4).Infof("found .dockercfg at %s", absDockerConfigFileLocation)
-			return cfg, nil
+		if err != nil {
+			klog.V(4).Infof("couldn't get the config from %q contents: %v", absDockerConfigFileLocation, err)
+			continue
 		}
+
+		klog.V(4).Infof("found .dockercfg at %s", absDockerConfigFileLocation)
+		return cfg, nil
+
 	}
 	return nil, fmt.Errorf("couldn't find valid .dockercfg after checking in %v", searchPaths)
 }
@@ -230,8 +234,7 @@ func ReadDockerConfigFileFromURL(url string, client *http.Client, header *http.H
 
 func readDockerConfigFileFromBytes(contents []byte) (cfg DockerConfig, err error) {
 	if err = json.Unmarshal(contents, &cfg); err != nil {
-		klog.Errorf("while trying to parse blob %q: %v", contents, err)
-		return nil, err
+		return nil, errors.New("error occurred while trying to unmarshal json")
 	}
 	return
 }
@@ -239,8 +242,7 @@ func readDockerConfigFileFromBytes(contents []byte) (cfg DockerConfig, err error
 func readDockerConfigJSONFileFromBytes(contents []byte) (cfg DockerConfig, err error) {
 	var cfgJSON DockerConfigJSON
 	if err = json.Unmarshal(contents, &cfgJSON); err != nil {
-		klog.Errorf("while trying to parse blob %q: %v", contents, err)
-		return nil, err
+		return nil, errors.New("error occurred while trying to unmarshal json")
 	}
 	cfg = cfgJSON.Auths
 	return
--- a/pkg/credentialprovider/config_test.go
+++ b/pkg/credentialprovider/config_test.go
@@ -309,3 +309,96 @@ func TestDockerConfigEntryJSONCompatibleEncode(t *testing.T) {
 		}
 	}
 }
+
+func TestReadDockerConfigFileFromBytes(t *testing.T) {
+	testCases := []struct {
+		id               string
+		input            []byte
+		expectedCfg      DockerConfig
+		errorExpected    bool
+		expectedErrorMsg string
+	}{
+		{
+			id:    "valid input, no error expected",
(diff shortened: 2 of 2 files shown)


soltysh (approved):
/lgtm

pkg/credentialprovider/config.go
```diff
@@ -230,17 +233,15 @@ func ReadDockerConfigFileFromURL(url string, client *http.Client, header *http.H
 
 func readDockerConfigFileFromBytes(contents []byte) (cfg DockerConfig, err error) {
 	if err = json.Unmarshal(contents, &cfg); err != nil {
-		klog.Errorf("while trying to parse blob %q: %v", contents, err)
-		return nil, err
+		return nil, fmt.Errorf("while trying to parse blob: %v", err)
```
deads2k, line 236: add a test to make sure this does what you want.

pkg/credentialprovider/config.go
```diff
@@ -117,10 +117,13 @@ func ReadDockercfgFile(searchPaths []string) (cfg DockerConfig, err error) {
 			continue
 		}
 		cfg, err := readDockerConfigFileFromBytes(contents)
-		if err == nil {
-			klog.V(4).Infof("found .dockercfg at %s", absDockerConfigFileLocation)
-			return cfg, nil
+		if err != nil {
+			return nil, fmt.Errorf("couldn't get the config from %q contents: %v", absDockerConfigFileLocation, err)
```
deads2k, line 121: %w for errors

pkg/credentialprovider/config.go
deads2k, line 236: %w for errors

deads2k:
I think a test is in order.  it's not obvious that an  unmarshaller does what you want here.

/lgtm cancel

pkg/credentialprovider/config.go
droslean: Addressed

pkg/credentialprovider/config.go
```diff
@@ -117,10 +117,13 @@ func ReadDockercfgFile(searchPaths []string) (cfg DockerConfig, err error) {
 			continue
 		}
 		cfg, err := readDockerConfigFileFromBytes(contents)
-		if err == nil {
-			klog.V(4).Infof("found .dockercfg at %s", absDockerConfigFileLocation)
-			return cfg, nil
+		if err != nil {
+			return nil, fmt.Errorf("couldn't get the config from %q contents: %w", absDockerConfigFileLocation, err)
```
liggitt, line 121: this changes behavior (a parsing error does not continue to the next potential file) and is unrelated to preventing contents from being logged

pkg/credentialprovider/config.go
```diff
@@ -230,17 +233,15 @@ func ReadDockerConfigFileFromURL(url string, client *http.Client, header *http.H
 
 func readDockerConfigFileFromBytes(contents []byte) (cfg DockerConfig, err error) {
 	if err = json.Unmarshal(contents, &cfg); err != nil {
-		klog.Errorf("while trying to parse blob %q: %v", contents, err)
-		return nil, err
+		return nil, fmt.Errorf("while trying to parse blob: %w", err)
```
liggitt, line 236: the error returned from the json parser could potentially include snippets of the content… we can't include it at all if we don't want any possibility of revealing anything about the content

pkg/credentialprovider/config.go
```diff
@@ -230,17 +233,15 @@ func ReadDockerConfigFileFromURL(url string, client *http.Client, header *http.H
...
 	}
 	return
 }
 
 func readDockerConfigJSONFileFromBytes(contents []byte) (cfg DockerConfig, err error) {
 	var cfgJSON DockerConfigJSON
 	if err = json.Unmarshal(contents, &cfgJSON); err != nil {
-		klog.Errorf("while trying to parse blob %q: %v", contents, err)
-		return nil, err
+		return nil, fmt.Errorf("while trying to parse blob: %w", err)
```
liggitt, line 244: same comment about snippets of content potentially being in the error returned from the parser

pkg/credentialprovider/config.go
droslean: Thanks, addressed

pkg/credentialprovider/config.go
droslean: I wasn't able to find a scenario where the parser will leak the secret. It seems that the error output of the parser points directly to the invalid char, and that makes impossible to leak anything.

liggitt: That could change in future releases, or in different parsers. We should not propagate an error that could contain file content

droslean: Yea, that makes sense. However the current code is already returning the error. Shall we just return an error message that we weren't able to unmarshal the JSON file without even printing the actual error, or we should find a way how to cencor any potential secrets from the json parser errors?

liggitt: > Shall we just return an error message that we weren't able to unmarshal the JSON file without even printing the actual error, or we should find a way how to cencor any potential secrets from the json parser errors?

just return an error we construct saying we got a json unmarshal error... don't try to censor the warning returned from the unmarshaler

droslean: Addressed

liggitt:
#94547 flake

sfowl:
I opened cherry-picks given the potential sec impact.

PR #94723, opened 2020-09-11
refactor: migrate node e2e tests off insecure port

### LINKED ISSUE #94658: Migrate node E2E off insecure port
The insecure port is being removed, but the node E2E test suite still depends no it. We need to migrate the test suite off the insecure port.

/milestone 1.20
/sig node
/priority important-soon
/help

knight42:
**What type of PR is this?**

/kind cleanup

Disable apiserver insecure serving in node e2e tests.

Fixes #94658

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/framework/kubelet/config.go
+++ b/test/e2e/framework/kubelet/config.go
@@ -75,7 +75,7 @@ func pollConfigz(timeout time.Duration, pollInterval time.Duration, nodeName, na
 		framework.Logf("http requesting node kubelet /configz")
 		endpoint = fmt.Sprintf("http://127.0.0.1:%d/api/v1/nodes/%s/proxy/configz", port, nodeName)
 	} else {
-		endpoint = fmt.Sprintf("http://127.0.0.1:8080/api/v1/nodes/%s/proxy/configz", framework.TestContext.NodeName)
+		endpoint = fmt.Sprintf("%s/api/v1/nodes/%s/proxy/configz", framework.TestContext.Host, framework.TestContext.NodeName)
 	}
 	tr := &http.Transport{
 		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
@@ -83,10 +83,13 @@ func pollConfigz(timeout time.Duration, pollInterval time.Duration, nodeName, na
 	client := &http.Client{Transport: tr}
 	req, err := http.NewRequest("GET", endpoint, nil)
 	framework.ExpectNoError(err)
+	if !useProxy {
+		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", framework.TestContext.BearerToken))
+	}
 	req.Header.Add("Accept", "application/json")
 
 	var resp *http.Response
-	wait.PollImmediate(pollInterval, timeout, func() (bool, error) {
+	err = wait.PollImmediate(pollInterval, timeout, func() (bool, error) {
 		resp, err = client.Do(req)
 		if err != nil {
 			framework.Logf("Failed to get /configz, retrying. Error: %v", err)
@@ -99,6 +102,7 @@ func pollConfigz(timeout time.Duration, pollInterval time.Duration, nodeName, na
 
 		return true, nil
 	})
+	framework.ExpectNoError(err, "Failed to get successful response from /configz")
 	return resp
 }
 
--- a/test/e2e/framework/test_context.go
+++ b/test/e2e/framework/test_context.go
@@ -17,9 +17,12 @@ limitations under the License.
 package framework
 
 import (
+	"crypto/rand"
+	"encoding/base64"
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"os"
 	"sort"
 	"strings"
@@ -32,11 +35,12 @@ import (
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	cliflag "k8s.io/component-base/cli/flag"
 	"k8s.io/klog/v2"
+
 	kubeletconfig "k8s.io/kubernetes/pkg/kubelet/apis/config"
 )
 
 const (
-	defaultHost = "http://127.0.0.1:8080"
+	defaultHost = "https://127.0.0.1:6443"
 
 	// DefaultNumNodes is the number of nodes. If not specified, then number of nodes is auto-detected
 	DefaultNumNodes = -1
@@ -77,6 +81,7 @@ type TestContextType struct {
 	KubeVolumeDir      string
 	CertDir            string
 	Host               string
+	BearerToken        string
 	// TODO: Deprecating this over time... instead just use gobindata_util.go , see #23987.
 	RepoRoot                string
 	DockershimCheckpointDir string
@@ -286,7 +291,7 @@ func RegisterCommonFlags(flags *flag.FlagSet) {
 	flags.BoolVar(&TestContext.DeleteNamespaceOnFailure, "delete-namespace-on-failure", true, "If true, framework will delete test namespace on failure. Used only during test debugging.")
 	flags.IntVar(&TestContext.AllowedNotReadyNodes, "allowed-not-ready-nodes", 0, "If non-zero, framework will allow for that many non-ready nodes when checking for all ready nodes.")
 
-	flags.StringVar(&TestContext.Host, "host", "", fmt.Sprintf("The host, or apiserver, to connect to. Will default to %s if this argument and --kubeconfig are not set", defaultHost))
+	flags.StringVar(&TestContext.Host, "host", "", fmt.Sprintf("The host, or apiserver, to connect to. Will default to %s if this argument and --kubeconfig are not set.", defaultHost))
 	flags.StringVar(&TestContext.ReportPrefix, "report-prefix", "", "Optional prefix for JUnit XML reports. Default is empty, which doesn't prepend anything to the default name.")
 	flags.StringVar(&TestContext.ReportDir, "report-dir", "", "Path to the directory where the JUnit XML reports should be saved. Default is empty, which doesn't generate these reports.")
 	flags.Var(cliflag.NewMapStringBool(&TestContext.FeatureGates), "feature-gates", "A set of key=value pairs that describe feature gates for alpha/experimental features.")
@@ -402,6 +407,18 @@ func createKubeConfig(clientCfg *restclient.Config) *clientcmdapi.Config {
 	return configCmd
 }
 
+func generateSecureToken(tokenLen int) (string, error) {
+	// Number of bytes to be tokenLen when base64 encoded.
+	tokenSize := math.Ceil(float64(tokenLen) * 6 / 8)
+	rawToken := make([]byte, int(tokenSize))
+	if _, err := rand.Read(rawToken); err != nil {
+		return "", err
+	}
+	encoded := base64.RawURLEncoding.EncodeToString(rawToken)
+	token := encoded[:tokenLen]
+	return token, nil
+}
+
 // AfterReadingAllFlags makes changes to the context after all flags
 // have been read.
 func AfterReadingAllFlags(t *TestContextType) {
@@ -419,6 +436,13 @@ func AfterReadingAllFlags(t *TestContextType) {
 		if len(t.KubeConfig) == 0 {
 			klog.Warningf("Unable to find in-cluster config, using default host : %s", defaultHost)
 			t.Host = defaultHost
+		}
+	}
+	if len(t.BearerToken) == 0 {
+		var err error
+		t.BearerToken, err = generateSecureToken(16)
+		if err != nil {
+			klog.Fatalf("Failed to generate bearer token: %v", err)
 		}
 	}
 	// Allow 1% of nodes to be unready (statistically) - relevant for large clusters.
--- a/test/e2e_node/e2e_node_suite_test.go
+++ b/test/e2e_node/e2e_node_suite_test.go
@@ -56,18 +56,21 @@ import (
 	"k8s.io/klog/v2"
 )
 
-var e2es *services.E2EServices
+var (
+	e2es *services.E2EServices
 
-// TODO(random-liu): Change the following modes to sub-command.
-var runServicesMode = flag.Bool("run-services-mode", false, "If true, only run services (etcd, apiserver) in current process, and not run test.")
-var runKubeletMode = flag.Bool("run-kubelet-mode", false, "If true, only start kubelet, and not run test.")
-var systemValidateMode = flag.Bool("system-validate-mode", false, "If true, only run system validation in current process, and not run test.")
-var systemSpecFile = flag.String("system-spec-file", "", "The name of the system spec file that will be used for node conformance test. If it's unspecified or empty, the default system spec (system.DefaultSysSpec) will be used.")
+	// TODO(random-liu): Change the following modes to sub-command.
+	runServicesMode    = flag.Bool("run-services-mode", false, "If true, only run services (etcd, apiserver) in current process, and not run test.")
+	runKubeletMode     = flag.Bool("run-kubelet-mode", false, "If true, only start kubelet, and not run test.")
+	systemValidateMode = flag.Bool("system-validate-mode", false, "If true, only run system validation in current process, and not run test.")
+	systemSpecFile     = flag.String("system-spec-file", "", "The name of the system spec file that will be used for node conformance test. If it's unspecified or empty, the default system spec (system.DefaultSysSpec) will be used.")
+)
 
 // registerNodeFlags registers flags specific to the node e2e test suite.
 func registerNodeFlags(flags *flag.FlagSet) {
 	// Mark the test as node e2e when node flags are api.Registry.
 	framework.TestContext.NodeE2E = true
+	flags.StringVar(&framework.TestContext.BearerToken, "bearer-token", "", "The bearer token to authenticate with. If not specified, it would be a random token. Currently this token is only used in node e2e tests.")
 	flags.StringVar(&framework.TestContext.NodeName, "node-name", "", "Name of the node to run tests on.")
 	// TODO(random-liu): Move kubelet start logic out of the test.
 	// TODO(random-liu): Move log fetch logic out of the test.
@@ -205,8 +208,12 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 	// Reference common test to make the import valid.
 	commontest.CurrentSuite = commontest.NodeE2E
 
-	return nil
-}, func([]byte) {
+	// ginkgo would spawn multiple processes to run tests.
+	// Since the bearer token is generated randomly at run time,
+	// we need to distribute the bearer token to other processes to make them use the same token.
+	return []byte(framework.TestContext.BearerToken)
+}, func(token []byte) {
+	framework.TestContext.BearerToken = string(token)
 	// update test context with node configuration.
 	gomega.Expect(updateTestContext()).To(gomega.Succeed(), "update test context with node config.")
 })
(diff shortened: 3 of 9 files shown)


cluster/gce/gci/configure-kubeapiserver.sh
```diff
@@ -76,8 +76,9 @@ function start-kube-apiserver {
   configure-etcd-params params
 
   params+=" --secure-port=443"
-  if [[ "${ENABLE_APISERVER_INSECURE_PORT:-false}" != "true" ]]; then
-    # Default is :8080
+  if [[ "${ENABLE_APISERVER_INSECURE_PORT:-false}" == "true" ]]; then
+    params+=" --insecure-port=8080"
+  else
```
tallclair, line 81: Do we need to keep this? The plan is to remove the flag behavior this release.

test/e2e_node/services/kubelet.go
```diff
@@ -355,22 +355,24 @@ func createPodDirectory() (string, error) {
 }
 
 // createKubeconfig creates a kubeconfig file at the fully qualified `path`. The parent dirs must exist.
-func createKubeconfig(path string) error {
-	kubeconfig := []byte(`apiVersion: v1
+func createKubeconfig(path, token string) error {
+	kubeconfig := []byte(fmt.Sprintf(`apiVersion: v1
 kind: Config
 users:
 - name: kubelet
```
tallclair, line 362: Does the username here need to match the user name in the server-side token file?

cluster/gce/gci/configure-kubeapiserver.sh
tallclair: Actually, I think this is unrelated to this PR, and you can remove this file. I'll pick it up in the follow up PR to remove the insecure port.

test/e2e/framework/test_context.go
```diff
@@ -36,7 +36,8 @@ import (
 )
 
 const (
-	defaultHost = "http://127.0.0.1:8080"
+	defaultHost        = "https://127.0.0.1:6443"
+	defaultBearerToken = "insecure-token"
```
tallclair, line 40: What happens if we randomize this, instead of using a const? Do the tests pass, or is there a problem with the way the process is forked for running the different services?

tallclair (commented):
Thanks so much for picking this up! Looks good, just a few nits & questions.

test/e2e_node/services/apiserver.go
```diff
@@ -96,9 +102,14 @@ func (a *APIServer) Name() string {
 }
 
 func getAPIServerClientURL() string {
-	return apiserverClientURL
+	return framework.TestContext.Host
```
tallclair, line 105: What does this get set to? Because you're hardcoding the bind address to `127.0.0.1`, so I think anything else (except maybe localhost) won't work. Note that the bind address MUST be 127.0.0.1 as long as you're using the `insecure-token`.

test/e2e_node/services/kubelet.go
knight42: they don't need to match, but I am okay to make them the same.

test/e2e_node/services/apiserver.go
```diff
@@ -96,9 +104,15 @@ func (a *APIServer) Name() string {
...
 }
 
 func getAPIServerHealthCheckURL() string {
-	return apiserverHealthCheckURL
+	return framework.TestContext.Host + "/healthz"
+}
+
+func generateTokenFile(tokenFilePath string) error {
+	klog.Info("DEBUG: Generating token")
+	tokenFile := fmt.Sprintf("%s,kubelet,uid,system:masters\n", framework.TestContext.BearerToken)
```
knight42, line 116: @tallclair Is it okay to use the built-in group `system:masters` here? Do we need to use a new group with limited privileges?

test/e2e_node/services/apiserver.go
knight42: Oh I forgot the listen address is hard-coded.. then do we need to enforce `framework.TestContext.Host` to be `https://127.0.0.1:6443` as well?

cluster/gce/gci/configure-kubeapiserver.sh
knight42: I removed this file from this PR. Would you mind me changing the default value of `--insecure-port` in a separate PR as finishing the remaining work in https://github.com/tallclair/kubernetes/pull/2?

test/e2e/framework/test_context.go
knight42: I think using random token is OK, since the token is only generated once.

knight42: hmm looks like we could not save random token in `framework.TestContext`, it would be difference across the tests.

knight42: perhaps we could specify a random token via `--bearer-token` flag.

test/e2e_node/services/apiserver.go
tallclair: The privileged group is fine I think.

cluster/gce/gci/configure-kubeapiserver.sh
tallclair: If you want to pick up the remaining work, please do!

test/e2e/framework/test_context.go
tallclair: ok. In that case, can we make the flag required, instead of defaulting to the insecure setup?

test/e2e_node/services/apiserver.go
tallclair: Maybe we should just default the TestContext.Host to localhost? (only for the node suite though)

test/e2e/framework/test_context.go
knight42: If we made the flag required, we need to also update the job config https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml to pass this flag, otherwise the test won't pass.

I was considering could we use a default token first, then update the job config to pass `BUILD_ID`(https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables) as token to the job, and finally mark the token flag as required.

cluster/gce/gci/configure-kubeapiserver.sh
knight42: Thanks, I would love to!

test/e2e_node/services/apiserver.go
knight42: `TestContext.Host` is actually the `defaultHost` (i.e. `https://127.0.0.1:6443`)https://github.com/kubernetes/kubernetes/blob/7a2812ca6087e62d0c9b6dde050065fe2f665a0e/test/e2e/framework/test_context.go#L419-L421 in node e2e test.

test/e2e/framework/kubelet/config.go
```diff
@@ -99,6 +102,7 @@ func pollConfigz(timeout time.Duration, pollInterval time.Duration, nodeName, na
 
 		return true, nil
 	})
+	framework.ExpectNoError(err)
```
neolit123, line 105: We should annotate the error here:
`err, "failed to..."`

test/e2e/framework/test_context.go
```diff
@@ -286,7 +289,8 @@ func RegisterCommonFlags(flags *flag.FlagSet) {
 	flags.BoolVar(&TestContext.DeleteNamespaceOnFailure, "delete-namespace-on-failure", true, "If true, framework will delete test namespace on failure. Used only during test debugging.")
 	flags.IntVar(&TestContext.AllowedNotReadyNodes, "allowed-not-ready-nodes", 0, "If non-zero, framework will allow for that many non-ready nodes when checking for all ready nodes.")
 
-	flags.StringVar(&TestContext.Host, "host", "", fmt.Sprintf("The host, or apiserver, to connect to. Will default to %s if this argument and --kubeconfig are not set", defaultHost))
+	flags.StringVar(&TestContext.Host, "host", "", fmt.Sprintf("The host, or apiserver, to connect to. Will default to %s if this argument and --kubeconfig are not set.", defaultHost))
+	flags.StringVar(&TestContext.BearerToken, "bearer-token", defaultBearerToken, "The bearer token to authenticate with.")
```
neolit123, line 293: Could we clarify in the description "where" this authn / flag is needed - i.e. node e2e.

test/e2e_node/services/util.go
```diff
@@ -17,19 +17,33 @@ limitations under the License.
...
 	"time"
+
+	"k8s.io/klog/v2"
+
+	"k8s.io/kubernetes/test/e2e/framework"
 )
 
 // terminationSignals are signals that cause the program to exit in the
 // supported platforms (linux, darwin, windows).
 var terminationSignals = []os.Signal{syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT}
```
neolit123, line 36: We can omit this empty line / place all vars in the same `vars (` block.

test/e2e_node/services/util.go
```diff
@@ -17,19 +17,33 @@ limitations under the License.
...
 
+var insecureHTTPClient = &http.Client{}
+
+const tokenFilePath = "known_tokens.csv"
+
+func init() {
```
neolit123, line 41: Can we avoid this init function - e.g. make the client local to functions that need it?

neolit123 (commented):
Thanks for the PR @knight42 !
I have added some minor suggestions.

test/e2e/framework/test_context.go
knight42: should we only add this flag when running node e2e tests?

neolit123: i'd say yes.

test/e2e_node/services/util.go
```diff
@@ -17,18 +17,25 @@ limitations under the License.
...
 )
 
-// terminationSignals are signals that cause the program to exit in the
-// supported platforms (linux, darwin, windows).
-var terminationSignals = []os.Signal{syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT}
+var (
+	// terminationSignals are signals that cause the program to exit in the
+	// supported platforms (linux, darwin, windows).
+	terminationSignals = []os.Signal{syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT}
+	tokenFilePath      = "known_tokens.csv"
```
neolit123, line 37: i think this can be a constant in Start()

test/e2e_node/services/util.go
```diff
@@ -80,3 +93,20 @@ func readinessCheck(name string, urls []string, errCh <-chan error) error {
...
+// Only returns non-recoverable errors.
+func healthCheck(client *http.Client, url string) bool {
+	req, err := http.NewRequest("HEAD", url, nil)
+	if err != nil {
+		return false
+	}
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", framework.TestContext.BearerToken))
+	resp, err := client.Do(req)
+	if err != nil {
+		klog.Warningf("Health check on '%s' failed, error=%v", url, err)
```
neolit123, line 107: `'%s'` -> `%q`

test/e2e_node/services/util.go
```diff
@@ -80,3 +93,20 @@ func readinessCheck(name string, urls []string, errCh <-chan error) error {
...
+	} else if resp.StatusCode != http.StatusOK {
+		klog.Warningf("Health check on '%s' failed, status=%d", url, resp.StatusCode)
```
neolit123, line 109: `'%s'` -> `%q`

neolit123 (commented):
/approve
please squash the commits.

neolit123:
/cc @tallclair 
PTAL

knight42:
@neolit123 BTW what do you think of my propose at https://github.com/kubernetes/kubernetes/pull/94723#discussion_r493746912?

test/e2e/framework/test_context.go
neolit123: > hmm looks like we could not save random token in framework.TestContext, it would be difference across the tests.

but this should not be a problem if we randomize the value of `framework.TestContext` if it equals `defaultBearerToken` once, at each run of the node suite [here](https://github.com/kubernetes/kubernetes/blob/3f585835d0a38462bb928454245ec01e834e53a0/test/e2e_node/e2e_node_suite_test.go#L120)?

neolit123:
> @neolit123 BTW what do you think of my propose at #94723 (comment)?

commented, i think it's safe to randomize per run.

test/e2e/framework/test_context.go
knight42: Actually I have tried this before, and it turns out that wouldn't help.

The root cause is that the test binary would invoke itself again with flag `--run-services-mode` to start dependant services, such as apiserver and etcd, like
```
I0926 21:26:13.050] I0926 21:24:05.791988     980 server.go:102] Starting server "services" with command "/tmp/node-e2e-20200926T212229/e2e_node.test --run-services-mode --test.timeout=24h0m0s --ginkgo.seed=1601155359 --ginkgo.focus=\\[NodeConformance\\] --ginkgo.skip=\\[Flaky\\]|\\[Slow\\]|\\[Serial\\] --ginkgo.parallel.node=1 --ginkgo.parallel.total=8 ... # remaining flags omitted
```
and the token that apiserver used is saved as file in this process.

Now that there are actually 2 process running, if we randomize the token inside the process, their value would be different.

neolit123:
Ok but this means we can start the child processes by passing the token
flag explicitly, with the random taken that the parent process generated.
I can double check again later.

knight42:
> Ok but this means we can start the child processes by passing the token flag explicitly, with the random taken that the parent process generated.

@neolit123 Ah I think that make sense, should I update the code as you said now?

neolit123:
> Ah I think that make sense, should I update the code as you said now?

yes, please.

knight42:
@neolit123 It seems that ginkgo would spawn multiple processes instead of goroutines to run tests, so we have to pass the generated token to other processes in `ginkgo.SynchronizedBeforeSuite`.

Now a random berarer token is used in node e2e test.

test/e2e/framework/test_context.go
```diff
@@ -421,6 +424,9 @@ func AfterReadingAllFlags(t *TestContextType) {
 			t.Host = defaultHost
 		}
 	}
+	if len(t.BearerToken) == 0 {
+		t.BearerToken = rand.String(12)
```
neolit123, line 428: given the tests allow remote hosts, we should use crypto/rand and a token with more bits:
```
tokenBytes := make([]byte, 16)
if _, err := rand.Read(tokenBytes); err != nil {
	panic("could not generate random bearer token")
}	
t.BearerToken = hex.EncodeToString(tokenBytes)
```

test/e2e_node/e2e_node_suite_test.go
```diff
@@ -205,8 +208,13 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 	// Reference common test to make the import valid.
 	commontest.CurrentSuite = commontest.NodeE2E
 
-	return nil
-}, func([]byte) {
+	// ginkgo would spawn multiple processes to run tests.
+	// Since the bearer token is generated randomly at run time,
+	// we need to distribute the bearer token to other processes to make them use the same token.
+	klog.Infof("Distribute bearer token: %s", framework.TestContext.BearerToken)
```
neolit123, line 214: i think we can omit this klog.Infof() line.

test/e2e_node/e2e_node_suite_test.go
```diff
@@ -205,8 +208,13 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+	return []byte(framework.TestContext.BearerToken)
+}, func(token []byte) {
+	framework.TestContext.BearerToken = string(token)
```
neolit123, line 217: this seems fine as per the docs for SynchronizedBeforeSuite
https://github.com/onsi/ginkgo/blob/master/ginkgo_dsl.go#L504

just to double check my understanding here, does this mean that the rest of framework.TestContext is empty when the child processes are run and only the token field is set?

test/e2e/framework/test_context.go
knight42: I agree we could use a token with more bits, but currently apiserver only listen on localhost:
```go
	o.SecureServing.BindAddress = net.ParseIP("127.0.0.1")
	// Disable insecure serving
	o.InsecureServing.BindPort = 0
```

test/e2e_node/e2e_node_suite_test.go
knight42: > does this mean that the rest of framework.TestContext is empty when the child processes are run and only the token field is set?

nope, `framework.TestContext` is set like the parent process and the token field is another random token.

test/e2e_node/e2e_node_suite_test.go
knight42: fixed

test/e2e/framework/test_context.go
neolit123: ok, i was confusing this with `TestContext.Host`.

maybe we should add this comment just in case:

```
// if support for listening on external addresses is added, the BearerToken should be randomized with
// a cryptographically secure algorithm - e.g. crypto/rand.
o.SecureServing.BindAddress = net.ParseIP("127.0.0.1")
```

test/e2e_node/e2e_node_suite_test.go
neolit123: ok, i see. so this persists the original token that we generated in the parent process.

test/e2e/framework/test_context.go
knight42: comment added

test/e2e_node/e2e_node_suite_test.go
knight42: yeah you are right.

neolit123 (commented):
thanks @knight42 

please squash the commits again (usually i just amend the local commit and use `git push -f ...`)
ideally, this change should get another review from @tallclair .

knight42:
@neolit123 Thanks for your advice! I thought it might be clearer to see the changes after review in a separate commit. If you prefer a single commit, I could also do that.

knight42:
kindly ping @tallclair

test/e2e/framework/test_context.go
```diff
@@ -421,6 +424,9 @@ func AfterReadingAllFlags(t *TestContextType) {
 			t.Host = defaultHost
 		}
 	}
+	if len(t.BearerToken) == 0 {
+		t.BearerToken = rand.String(16)
```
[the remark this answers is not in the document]
knight42: fixed

bai:
Hey 👋 Bug Triage here. Just wanted to post a friendly reminder that the code freeze is starting 12.11.2020 (about 3-4 weeks from now) and while there is still plenty of time, we want to ensure that each PR has a chance to be merged on time.

tallclair:
/lgtm

Thanks!

test/e2e_node/services/services.go
```diff
@@ -65,14 +65,16 @@ func NewE2EServices(monitorParent bool) *E2EServices {
...
+		}
 		// Start kubelet
 		e.kubelet, err = e.startKubelet()
 		if err != nil {
 			return fmt.Errorf("failed to start kubelet: %v", err)
 		}
 	}
-	e.services, err = e.startInternalServices()
-	return err
+	return nil
```
MHBauer, line 77: I think our problem for #95877, we don't even start the services in conformance mode now.

mmerkes: Perhaps this was moved because we start the kubelet [here in conformance tests?](https://github.com/kubernetes/kubernetes/blob/master/test/e2e_node/remote/node_conformance.go#L176) So I guess we need to separate starting the internal services.

MHBauer: Ya, that's why the kubelet start is inside.

But there isn't any change to node_conformance to start the services.

PR #94730, opened 2020-09-11
Ensuring EndpointSlices are recreated after Service recreation

### LINKED ISSUE #94720: endpointslices: When Service is recreate too fast endpoint slices are not re-created.
<!-- Please use this template while reporting a bug and provide as much info as possible. Not doing so may result in your bug not being addressed in a timely manner. Thanks!

If the matter is security related, please disclose it privately via https://kubernetes.io/security/
-->


**What happened**:
At https//github.com/kubevirt/kubevirt we have a operator that recreate services if they get deleted, this is done programatically quite fast, after upgrading our cluster to k8s-1.19 doing so prevent service connectivity with `connection resused`, doing some investigations we found that the `endpointslices` (they have being activated at kube-proxy by default at k8s-1.19) does not get re-created by the endpointslices controller at k8s and that prevents connectivity since iptables rule does not appear at kube-proxy.

At fast Service recreate endpointslices has to be recreated too.

```bash
#!/bin/bash                                                                     
                                                                                
set -xe                                                                         
                                                                                
                                                                                
                                                                                
# Dump files instead of using stdin with kubectl it runs kubectl create         
# faster and race condition appear                                              
                                                                                
cat << EOF > pod.yaml                                                           
---                                                                             
apiVersion: v1                                                                  
kind: Pod                                                                       
metadata:                                                                       
  labels:                                                                       
    run: http-server                                                            
  name: http-server                                                             
spec:                                                                           
  containers:                                                                   
  - image: danjellz/http-server                                                 
    imagePullPolicy: Always                                                     
    name: http-server                                                           
    ports:                                                                      
    - containerPort: 8080                                                       
      name: http                                                                
      protocol: TCP                                                             
EOF                                                                             
                                                                                
cat << EOF > service.yaml                                                       
---                                                                             
apiVersion: v1                                                                  
kind: Service                                                                   
metadata:                                                                       
  name: http-server                                                             
spec:                                                                           
  ports:                                                                        
  - name: http                                                                  
    port: 80                                                                    
    protocol: TCP                                                               
    targetPort: http                                                            
  selector:                                                                     
    run: http-server                                                            
  sessionAffinity: None                                                         
  type: ClusterIP                                                               
status:                                                                         
  loadBalancer: {}                                                              
EOF                                                                             
                                                                                
kubectl delete --ignore-not-found -f pod.yaml                                   
kubectl delete --ignore-not-found -f service.yaml                               
                                                                                
kubectl create -f service.yaml                                                  
kubectl create -f pod.yaml                                                      
                                                                                
sleep 10

#Will fail if the endpointslice is not there                                   
kubectl get endpointslices |grep http-server                                    
kubectl delete -f service.yaml && kubectl create -f service.yaml                
                                                                                
sleep 10                                                                        
                                                                                
kubectl get endpointslices |grep http-server                                    
```

**Environment**:
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0", GitCommit:"e19964183377d0ec2052d1f1fa930c4d7575bd50", GitTreeState:"clean", BuildDate:"2020-08-26T14:30:33Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0", GitCommit:"e19964183377d0ec2052d1f1fa930c4d7575bd50", GitTreeState:"clean", BuildDate:"2020-08-26T14:23:04Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
```
```
AME="CentOS Linux"
VERSION="8 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="8"
PLATFORM_ID="platform:el8"
PRETTY_NAME="CentOS Linux 8 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:8"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-8"
CENTOS_MANTISBT_PROJECT_VERSION="8"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="8"
```
- Kernel (e.g. `uname -a`):
`Linux node01 4.18.0-227.el8.x86_64 #1 SMP Tue Jul 28 22:53:29 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`
- Install tools:
`calico 3.15.3`
- Others:

robscott:
**What type of PR is this?**
/kind bug

This fixes a bug that occurred when a Service was rapidly recreated. This relied on an unfortunate series of events:

1. When the Service is deleted, the EndpointSlice controller removes it from the EndpointSliceTracker along with any associated EndpointSlices.

2. When the Service is recreated, the EndpointSlice controller sees that there are still appropriate EndpointSlices for the Service and does nothing. (They have not yet been garbage collected).

3. When the EndpointSlice is deleted, the EndpointSlice controller checks with the EndpointSliceTracker to see if it thinks we should have this EndpointSlice (it does not because of step 1). This check was intended to ensure we wouldn't requeue a Service every time we delete an EndpointSlice for it.

This adds a check in reconciler to ensure that EndpointSlices it is working with are owned by a Service with a matching UID. If not, it will mark those EndpointSlices for deletion (assuming they're about to be garbage collected anyway) and create new EndpointSlices.

Fixes #94720

```release-note
Fixes a regression in 1.19 where EndpointSlices would not be recreated after rapid Service recreation.
```

/sig network
/priority critical-urgent
/cc @freehan @liggitt @qinqon

### DIFF
--- a/pkg/controller/endpointslice/reconciler.go
+++ b/pkg/controller/endpointslice/reconciler.go
@@ -70,7 +70,7 @@ func (r *reconciler) reconcile(service *corev1.Service, pods []*corev1.Pod, exis
 	existingSlicesByPortMap := map[endpointutil.PortMapKey][]*discovery.EndpointSlice{}
 	numExistingEndpoints := 0
 	for _, existingSlice := range existingSlices {
-		if existingSlice.AddressType == addressType {
+		if existingSlice.AddressType == addressType && ownedBy(existingSlice, service) {
 			epHash := endpointutil.NewPortMapKey(existingSlice.Ports)
 			existingSlicesByPortMap[epHash] = append(existingSlicesByPortMap[epHash], existingSlice)
 			numExistingEndpoints += len(existingSlice.Endpoints)
@@ -187,13 +187,15 @@ func (r *reconciler) finalize(
 		}
 		sliceToDelete := slicesToDelete[i]
 		slice := slicesToCreate[len(slicesToCreate)-1]
-		// Only update EndpointSlices that have the same AddressType as this
-		// field is considered immutable. Since Services also consider IPFamily
-		// immutable, the only case where this should matter will be the
-		// migration from IP to IPv4 and IPv6 AddressTypes, where there's a
+		// Only update EndpointSlices that are owned by this Service and have
+		// the same AddressType. We need to avoid updating EndpointSlices that
+		// are being garbage collected for an old Service with the same name.
+		// The AddressType field is immutable. Since Services also consider
+		// IPFamily immutable, the only case where this should matter will be
+		// the migration from IP to IPv4 and IPv6 AddressTypes, where there's a
 		// chance EndpointSlices with an IP AddressType would otherwise be
 		// updated to IPv4 or IPv6 without this check.
-		if sliceToDelete.AddressType == slice.AddressType {
+		if sliceToDelete.AddressType == slice.AddressType && ownedBy(sliceToDelete, service) {
 			slice.Name = sliceToDelete.Name
 			slicesToCreate = slicesToCreate[:len(slicesToCreate)-1]
 			slicesToUpdate = append(slicesToUpdate, slice)
--- a/pkg/controller/endpointslice/utils.go
+++ b/pkg/controller/endpointslice/utils.go
@@ -201,6 +201,17 @@ func objectRefPtrChanged(ref1, ref2 *corev1.ObjectReference) bool {
 	return false
 }
 
+// ownedBy returns true if the provided EndpointSlice is owned by the provided
+// Service.
+func ownedBy(endpointSlice *discovery.EndpointSlice, svc *corev1.Service) bool {
+	for _, o := range endpointSlice.OwnerReferences {
+		if o.UID == svc.UID && o.Kind == "Service" && o.APIVersion == "v1" {
+			return true
+		}
+	}
+	return false
+}
+
 // getSliceToFill will return the EndpointSlice that will be closest to full
 // when numEndpoints are added. If no EndpointSlice can be found, a nil pointer
 // will be returned.
--- a/pkg/controller/endpointslice/endpointslice_controller_test.go
+++ b/pkg/controller/endpointslice/endpointslice_controller_test.go
@@ -29,6 +29,8 @@ import (
 	discovery "k8s.io/api/discovery/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
@@ -126,7 +128,7 @@ func TestSyncServiceWithSelector(t *testing.T) {
 	ns := metav1.NamespaceDefault
 	serviceName := "testing-1"
 	client, esController := newController([]string{"node-1"}, time.Duration(0))
-	standardSyncService(t, esController, ns, serviceName, "true")
+	standardSyncService(t, esController, ns, serviceName)
 	expectActions(t, client.Actions(), 1, "create", "endpointslices")
 
 	sliceList, err := client.DiscoveryV1beta1().EndpointSlices(ns).List(context.TODO(), metav1.ListOptions{})
@@ -192,7 +194,7 @@ func TestSyncServicePodSelection(t *testing.T) {
 	pod2.Labels["foo"] = "boo"
 	esController.podStore.Add(pod2)
 
-	standardSyncService(t, esController, ns, "testing-1", "true")
+	standardSyncService(t, esController, ns, "testing-1")
 	expectActions(t, client.Actions(), 1, "create", "endpointslices")
 
 	// an endpoint slice should be created, it should only reference pod1 (not pod2)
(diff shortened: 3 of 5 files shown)


robscott:
One additional note. My testing showed 1 alternative approach that works well here - delaying queuing from Service delete events for 1 second. This would result in very minimal additional overhead for the controller but it seemed difficult to guarantee that it would fix 100% of the issues here.

robscott:
I discussed this PR with @bowei and he had a much better solution - ensuring the UIDs in the EndpointSlice ownerRef match the current Service that is being processed. If not we'll mark those for deletion (assuming they're about to be garbage collected anyway) and create new EndpointSlices for the new Service. I've left the e2e test in place since I think it's at least somewhat valuable, but I believe this can now be adequately covered by the unit tests I've added as well.

pkg/controller/endpointslice/utils.go
```diff
@@ -201,6 +201,17 @@ func objectRefPtrChanged(ref1, ref2 *corev1.ObjectReference) bool {
 	return false
 }
 
+// ownedBy returns true if the provided EndpointSlice is owned by the provided
+// Service.
+func ownedBy(endpointSlice *discovery.EndpointSlice, service *corev1.Service) bool {
+	for _, ownerRef := range endpointSlice.OwnerReferences {
+		if ownerRef.UID == service.UID {
```
liggitt, line 208: check the kind and apiVersion as well

bowei:
/lgtm 

apart from Jordan's comment :-)

pkg/controller/endpointslice/reconciler.go
```diff
@@ -187,13 +187,15 @@ func (r *reconciler) finalize(
...
+		// Only update EndpointSlices that are owned by this Service and have
+		// the same AddressType. We need to avoid updating EndpointSlices that
+		// are being garbage collected for an old Service with the same name.
+		// The AddressType field is immutable. Since Services also consider
+		// IPFamily immutable, the only case where this should matter will be
+		// the migration from IP to IPv4 and IPv6 AddressTypes, where there's a
 		// chance EndpointSlices with an IP AddressType would otherwise be
 		// updated to IPv4 or IPv6 without this check.
-		if sliceToDelete.AddressType == slice.AddressType {
+		if sliceToDelete.AddressType == slice.AddressType && ownedBy(sliceToDelete, service) {
```
aojea, line 198: why not filtering all "not owned" slices in `reconcile`?

aojea: Do you think that `syncService` should be a better place? so `reconcile()` and `finalize()` does not have to deal with "undesired" slices?

robscott: That's a good idea, I'd thought about it as well but everything I tried ended up being messier/more complex than this change. Part of it is that syncService is intentionally small in scope and adding a check like this there ended up increasing the scope/complexity more than expected. Related to that, the majority of existing test coverage is focused on reconciler functions, so adding key functionality outside of that requires a larger investment in new testing.

pkg/controller/endpointslice/utils.go
robscott: Updated, thanks! Somehow I'd thought that UID was actually a UUID.

robscott:
Thanks for the feedback, I think I've responded to everything now. PTAL when you have time.

pkg/controller/endpointslice/reconciler.go
aojea: fair enough, I was thinking more in how endpoint slices consumers should implement this, and trying to come up with a "recommended" way of listing slices that belong to a service, based on this issue it seems that Listing based on label only is not enough.


```go
	esLabelSelector := labels.Set(map[string]string{
		discovery.LabelServiceName: service.Name,
		discovery.LabelManagedBy:   controllerName,
	}).AsSelectorPreValidated()
	endpointSlices, err := c.endpointSliceLister.EndpointSlices(service.Namespace).List(esLabelSelector)
```

Unrelated to this PR based on your comments but ... maybe we should offer some helper function that returns the slices that belong to a current service?

robscott: That's a great point. I think listing slices based on label has to be sufficient simply because we don't and can't control all the ways EndpointSlices can be created. An owner ref is technically optional for producers here. I think this is more of a problem for producers to ensure they clean up old resources than for consumers. If/when this is approved, I'll have to make a very similar PR for the EndpointSliceMirroring controller.

test/e2e/network/endpointslice.go
```diff
@@ -161,6 +152,62 @@ var _ = SIGDescribe("EndpointSlice", func() {
...
+
+		ensureValidEmptyEndpointSlice(endpointSlice)
+
+		err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
+		framework.ExpectNoError(err, "error deleting Service")
+
+		svc = createServiceReportErr(cs, f.Namespace.Name, svcConfig)
+
+		// Expect EndpointSlice resource to exist after 5 second delay.
+		time.Sleep(5 * time.Second)
```
aojea, line 199: this is a bit scary, the good thing is that it will only have false positives, I means the slice is created later than that

robscott: That's a good point, do you think there's a safer/more reliable way to cover this with an e2e test? I think this is already covered fairly well with unit tests so I could get rid of this entirely.

aojea: Yeah, or integration or unit, but this e2e does not seems worthy. Imho this is good without the e2e

robscott: Good call, updated to remove the e2e test entirely.

robscott:
@liggitt do you mind adding this to v1.20 milestone?

robscott:
Actually should be asking for milestone from someone in sig-network, @dcbw or @thockin are you able to add this to the 1.20 milestone?

ialidzhikov:
@robscott, can you also file a cherry-pick PR fo 1.19?

robscott:
Thanks for the reminder @ialidzhikov! Cherry pick PR created: #94948

PR #94734, opened 2020-09-11
It's an 'Instance' of apiserver

lavalamp:
**What type of PR is this?**

/kind cleanup

```release-note
NONE
```

### DIFF
--- a/pkg/controlplane/instance.go
+++ b/pkg/controlplane/instance.go
@@ -225,8 +225,8 @@ type EndpointReconcilerConfig struct {
 	Interval   time.Duration
 }
 
-// Master contains state for a Kubernetes cluster master/api server.
-type Master struct {
+// Instance contains state for a Kubernetes cluster api server instance.
+type Instance struct {
 	GenericAPIServer *genericapiserver.GenericAPIServer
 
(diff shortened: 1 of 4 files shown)


pkg/controlplane/instance.go
```diff
@@ -225,8 +225,8 @@ type EndpointReconcilerConfig struct {
 	Interval   time.Duration
 }
 
-// Master contains state for a Kubernetes cluster master/api server.
-type Master struct {
+// Instance contains state for a Kubernetes cluster api server instance.
```
hasheddan, line 228: Looks like this needs to be reflected in https://github.com/kubernetes/kubernetes/blob/66334f02e8c520df7973c397246da82cd4db2769/pkg/controlplane/instance_test.go#L202 also. Might be worth also updating var names there as well?

wojtek-t:
```
/b/f/w/pkg/controlplane/instance_test.go:202:32: undefined: Master
```

Once fixed lgtm

deads2k:
/lgtm
/hold

you have a compile error and a nit.

lavalamp:
I got too busy to fix the compile error, but @spiffxp has got it in #95044

PR #94736, opened 2020-09-11
cloud node controller: implement with workqueues and node lister

nicolehanjing:
As title

**What type of PR is this?**
/kind cleanup

Refactor the node controller to use Workqueue and Node Lister so that Informer watches for changes on the current state of Kubernetes objects and sends events to Workqueue where events are then popped up by worker(s) to process.
If we can use Workqueue in the implementation, it will be more readable and gain better maintenance.

Fixes #

```release-note
NONE
```

```docs

```


staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -132,26 +150,135 @@ func NewCloudNodeController(
...
+	// Wait for the caches to be synced before starting workers
+	klog.Info("Waiting for informer caches to sync")
+	if ok := cache.WaitForCacheSync(stopCh, cnc.nodesSynced); !ok {
+		klog.Errorf("failed to wait for caches to sync")
+		return
+	}
+
 	// Start a loop to periodically update the node addresses obtained from the cloud
-	wait.Until(func() { cnc.UpdateNodeStatus(context.TODO()) }, cnc.nodeStatusUpdateFrequency, stopCh)
+	go wait.Until(cnc.runWorker, cnc.nodeStatusUpdateFrequency, stopCh)
```
andrewsykim, line 173: I think we still want the periodic run of `UpdateNodeStatus` based on `cnc.nodeStatusUpdateFrequency` and `syncHandler` should actually only call `UpdateCloudNode` and not `UpdateNodeStatus`. So this should look like this instead:
```
go wait.Until(func() { cnc.UpdateNodeStatus(context.TODO()) }, cnc.nodeStatusUpdateFrequency, stopCh)
go wait.Until(c.runWorker, time.Second, stopCh)
```

andrewsykim:
Thanks for the PR @HaibaraAi96! 

I think overall we need this PR to update cloud node controller to use workqueues so that failed syncs aren't dropped. I left some initial feedback.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -132,26 +150,135 @@ func NewCloudNodeController(
...
+		// The Node resource may no longer exist, in which case we stop
+		// processing.
+		if apierrors.IsNotFound(err) {
+			utilruntime.HandleError(fmt.Errorf("Node '%s' in work queue no longer exists", key))
+			return nil
+		}
+
+		return err
+	}
+
```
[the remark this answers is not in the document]
HaibaraAi96: gotcha, will update soon!

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -120,6 +135,9 @@ func NewCloudNodeController(
 		recorder:                  recorder,
 		cloud:                     cloud,
 		nodeStatusUpdateFrequency: nodeStatusUpdateFrequency,
+		nodesLister:               nodeInformer.Lister(),
+		nodesSynced:               nodeInformer.Informer().HasSynced,
+		workqueue:                 workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "Nodes"),
```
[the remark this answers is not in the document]
JornShen: same

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -132,26 +150,135 @@ func NewCloudNodeController(
...
+	}
+
+	cnc.recorder.Event(node, v1.EventTypeNormal, NodeSuccessSynced, NodeMessageResourceSynced)
+	return nil
 }
 
 // UpdateNodeStatus updates the node status, such as node addresses
-func (cnc *CloudNodeController) UpdateNodeStatus(ctx context.Context) {
+func (cnc *CloudNodeController) UpdateNodeStatus(ctx context.Context) error {
 	nodes, err := cnc.kubeClient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{ResourceVersion: "0"})
```
JornShen, line 278: Is it necessary for us to still call kubeclient api to list nodes?  we can just get all nodes info from the cache(nodelisterner).

andrewsykim: > Is it necessary for us to still call kubeclient api to list nodes? we can just get all nodes info from the cache(nodelisterner).

Don't think so, we can probably update this to list nodes from the informer cache as well.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -120,38 +135,146 @@ func NewCloudNodeController(
...
+		workqueue:                 workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "Nodes"),
 	}
 
 	// Use shared informer to listen to add/update of nodes. Note that any nodes
 	// that exist before node controller starts will show up in the update method
 	cnc.nodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    func(obj interface{}) { cnc.AddCloudNode(context.TODO(), obj) },
-		UpdateFunc: func(oldObj, newObj interface{}) { cnc.UpdateCloudNode(context.TODO(), oldObj, newObj) },
+		AddFunc:    cnc.enqueueNode,
+		UpdateFunc: func(oldObj, newObj interface{}) { cnc.enqueueNode(newObj) },
```
HaibaraAi96, line 147: @andrewsykim updated! 
I think probably we don't need `AddCloudNode` function anymore since it is basically equivalent to `UpdateCloudNode ` but with different logs 🤔

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -295,7 +432,8 @@ func (cnc *CloudNodeController) updateNodeAddress(ctx context.Context, node *v1.
 // in a retry-if-conflict loop.
 type nodeModifier func(*v1.Node)
 
-func (cnc *CloudNodeController) UpdateCloudNode(ctx context.Context, _, newObj interface{}) {
+// UpdateCloudNode handles updating existing nodes registered with the cloud taint.
+func (cnc *CloudNodeController) UpdateCloudNode(ctx context.Context, newObj interface{}) {
```
andrewsykim, line 436: maybe just name this  `syncNode` now?

andrewsykim: Also thinking that  syncNode should receive the name of the node instead of `newObj interface{}` and in `initializeNode` we fetch the latest node from the informer cache line 478, instead of from apiserver.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -120,38 +135,146 @@ func NewCloudNodeController(
...
+		return nil
+	}
+
+	// Get the Node resource with this namespace/name
+	node, err := cnc.nodesLister.Get(name)
+	if err != nil {
+		// The Node resource may no longer exist, in which case we stop
+		// processing.
+		if apierrors.IsNotFound(err) {
+			utilruntime.HandleError(fmt.Errorf("Node '%s' in work queue no longer exists", key))
```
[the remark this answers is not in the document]
HaibaraAi96: sure, will remove that

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -295,45 +416,28 @@ func (cnc *CloudNodeController) updateNodeAddress(ctx context.Context, node *v1.
...
-func (cnc *CloudNodeController) UpdateCloudNode(ctx context.Context, _, newObj interface{}) {
-	node, ok := newObj.(*v1.Node)
-	if !ok {
-		utilruntime.HandleError(fmt.Errorf("unexpected object type: %v", newObj))
-		return
-	}
-
+// SyncNode handles updating existing nodes registered with the cloud taint.
+func (cnc *CloudNodeController) SyncNode(ctx context.Context, node *v1.Node) {
+	klog.Infof("Syncing node %s", node.Name)
```
HaibaraAi96, line 421: @andrewsykim I think maybe we should keep passing Node object for `SyncNode()` but only passing nodeName for `initializeNode()` cuz we still need to get **cloudTaint** of the node and see if there is any error

staging/src/k8s.io/cloud-provider/controllers/node/node_controller_test.go
```diff
@@ -162,7 +166,7 @@ func TestEnsureNodeExistsByProviderID(t *testing.T) {
 	}
 }
 
-func Test_AddCloudNode(t *testing.T) {
+func Test_SyncNode(t *testing.T) {
```
HaibaraAi96, line 169: @andrewsykim Also changed the unit test signature since we don't have `AddCloudNode()` function anymore

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
andrewsykim: I think we can remove the cloud taint check here since we have one already on line 457

andrewsykim: With that said it seems like we can probably collapse `SyncNode` and `initializeNode` into a single method `syncNode`.

andrewsykim: If we remove checking the cloud taint here, we should check the cloud taint in `initializeNode` _before_ adding `NodeNetworkUnavailable` condition for GCE on line 435

andrewsykim: If we remove checking the cloud taint here, we should check the cloud taint in `initializeNode` _before_ adding `NodeNetworkUnavailable` condition for GCE on line 435

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -367,7 +471,7 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
 	// get instanceMetadata and pass it to getNodeModifiersFromCloudProvider and updateNodeAddress which reduces api calls.
 	nodeModifiers, err := cnc.getNodeModifiersFromCloudProvider(ctx, curNode)
```
andrewsykim, line 472: We should deep copy the node object here before passing it into the node modifiers, since you shouldn't mutate cached objects
```
copyNode := curNode.DeepCopy()
nodeModifiers, err := cnc.getNodeModifiersFromCloudProvider(ctx, copyNode)
if err != nil {
  utilruntime.HandleError(fmt.Errorf("failed to initialize node %s at cloudprovider: %v", nodeName, err))
  return
}
```

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -376,7 +480,7 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
 	})
 
 	err = clientretry.RetryOnConflict(UpdateNodeSpecBackoff, func() error {
-		curNode, err := cnc.kubeClient.CoreV1().Nodes().Get(context.TODO(), node.Name, metav1.GetOptions{})
+		curNode, err := cnc.kubeClient.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
```
andrewsykim, line 483: I think this can also be updated to read from cache, but we should pass in a copied object to the modifer functions below.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
HaibaraAi96: gotcha, will update!

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
HaibaraAi96: agree, will collapse these two functions and also make corresponding changes

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -295,39 +416,18 @@ func (cnc *CloudNodeController) updateNodeAddress(ctx context.Context, node *v1.
...
-
-	cnc.initializeNode(ctx, node)
-}
-
-// AddCloudNode handles initializing new nodes registered with the cloud taint.
-func (cnc *CloudNodeController) AddCloudNode(ctx context.Context, obj interface{}) {
-	node := obj.(*v1.Node)
+// SyncNode handles updating existing nodes registered with the cloud taint
+// and processes nodes that were added into the cluster, and cloud initialize them if appropriate.
+func (cnc *CloudNodeController) SyncNode(ctx context.Context, node *v1.Node) {
```
HaibaraAi96, line 421: @andrewsykim Updated!
I collapse 2 functions and made corresponding changes, unit tests passed, let me know if there is any improvement 😃

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -84,6 +86,15 @@ var UpdateNodeSpecBackoff = wait.Backoff{
 	Jitter:   1.0,
 }
 
+const (
+	// NodeSuccessSynced is used as part of the Event 'reason' when a Node is synced
+	NodeSuccessSynced = "Synced"
```
andrewsykim, line 91: This constant and `NodeMessageResourceSynced` is only used once so we probably don't need constants and can just use the string literal directly in the event message.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -120,38 +135,130 @@ func NewCloudNodeController(
...
 
+// Run will sync informer caches and starting workers.
 // This controller updates newly registered nodes with information
 // from the cloud provider. This call is blocking so should be called
 // via a goroutine
 func (cnc *CloudNodeController) Run(stopCh <-chan struct{}) {
 	defer utilruntime.HandleCrash()
+	defer cnc.workqueue.ShutDown()
 
 	// The following loops run communicate with the APIServer with a worst case complexity
```
andrewsykim, line 161: This comment only applies for calls to `UpdateNodeStatus` but I think it's a bit confusing with the current wording. This comment should be moved right above line 173 and should be updated to something like:
```
// The periodic  loop for updateNodeStatus communicates with the APIServer with a worst case complexity
// of O(num_nodes) per cycle. These functions are justified here because these events fire
// very infrequently. DO NOT MODIFY this to perform frequent operations.
```

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -120,38 +135,130 @@ func NewCloudNodeController(
...
+		// Node resource to be synced.
+		if err := cnc.syncHandler(key); err != nil {
+			// Put the item back on the workqueue to handle any transient errors.
+			cnc.workqueue.AddRateLimited(key)
+			return fmt.Errorf("error syncing '%s': %s, requeuing", key, err.Error())
+		}
+		// Finally, if no error occurs we Forget this item so it does not
+		// get queued again until another change happens.
+		cnc.workqueue.Forget(obj)
+		klog.V(4).Infof("Successfully synced '%s'", key)
```
andrewsykim, line 216: nit: If you use %q it will quote strings automatically:
```
klog.V(4).Infof("Successfully synced %q", key)
```

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -295,39 +416,19 @@ func (cnc *CloudNodeController) updateNodeAddress(ctx context.Context, node *v1.
...
-	cnc.initializeNode(ctx, node)
-}
-
-// AddCloudNode handles initializing new nodes registered with the cloud taint.
-func (cnc *CloudNodeController) AddCloudNode(ctx context.Context, obj interface{}) {
-	node := obj.(*v1.Node)
+// SyncNode handles updating existing nodes registered with the cloud taint
+// and processes nodes that were added into the cluster, and cloud initialize them if appropriate.
+func (cnc *CloudNodeController) SyncNode(ctx context.Context, node *v1.Node) {
+	klog.Infof("Initializing node %s with cloud provider", node.Name)
```
andrewsykim, line 422: This log message should only happen after the cloud taint since we're technically not initializing the node if it doesn't have the cloud taint.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -295,39 +416,19 @@ func (cnc *CloudNodeController) updateNodeAddress(ctx context.Context, node *v1.
...
-
-	cnc.initializeNode(ctx, node)
-}
-
-// AddCloudNode handles initializing new nodes registered with the cloud taint.
-func (cnc *CloudNodeController) AddCloudNode(ctx context.Context, obj interface{}) {
-	node := obj.(*v1.Node)
+// SyncNode handles updating existing nodes registered with the cloud taint
+// and processes nodes that were added into the cluster, and cloud initialize them if appropriate.
+func (cnc *CloudNodeController) SyncNode(ctx context.Context, node *v1.Node) {
```
andrewsykim, line 421: Since on line 454 we're fetching the latest node from cache, I think syncNode should be updated to receive a node name and not the entire node object, and likewise we can remove getting a node from cache on line 239

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -120,38 +135,130 @@ func NewCloudNodeController(
...
+		// The Node resource may no longer exist, in which case we stop
+		// processing.
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+
+		return err
+	}
+
+	cnc.SyncNode(context.TODO(), node)
```
andrewsykim, line 250: Since we're using workqueues now and syncHandler returns an error, thinking `syncNode` should be refactored to reteurn an error, what do you think?

staging/src/k8s.io/cloud-provider/controllers/node/node_controller_test.go
```diff
@@ -1294,10 +1298,14 @@ func Test_AddCloudNode(t *testing.T) {
 				recorder:                  eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: "cloud-node-controller"}),
 				nodeStatusUpdateFrequency: 1 * time.Second,
 			}
+
+			cloudNodeController.nodeInformer.Informer().GetIndexer().Add(test.existingNode)
+			cloudNodeController.nodesSynced = alwaysReady
```
andrewsykim, line 1303: Do we need to mock `nodesSynced` here? I think since we're not calling `Run` and only `SyncNode`, we don't need to set this.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
andrewsykim, line 421: Since SyncNode is not called from outside this package it can be private (i.e. first letter should be lower case `syncNode`)

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
HaibaraAi96: gotcha, good to know!

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
HaibaraAi96: I agree, was thinking about this before but forget to create a comment, will update!

staging/src/k8s.io/cloud-provider/controllers/node/node_controller_test.go
HaibaraAi96: gotcha!

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -238,7 +238,12 @@ func (cnc *CloudNodeController) syncHandler(key string) error {
 		return err
 	}
 
-	err = cnc.syncNode(context.TODO(), node)
+	cloudTaint := getCloudTaint(node.Spec.Taints)
+	if cloudTaint == nil {
```
HaibaraAi96, line 242: @andrewsykim can we instead check cloudTaint here?
I think we still need to keep getting the node resource on Line 230 and use that in Line 251

andrewsykim: Would prefer to remove it here actually since there's an additional cache read here we don't need. For the event recorder on line 251, can we call that near the end of `syncNode` instead?

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -295,45 +415,17 @@ func (cnc *CloudNodeController) updateNodeAddress(ctx context.Context, node *v1.
...
-
-// This processes nodes that were added into the cluster, and cloud initialize them if appropriate
-func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Node) {
-	klog.Infof("Initializing node %s with cloud provider", node.Name)
+// syncNode handles updating existing nodes registered with the cloud taint
+// and processes nodes that were added into the cluster, and cloud initialize them if appropriate.
+func (cnc *CloudNodeController) syncNode(ctx context.Context, nodeName string) error {
+	klog.Infof("Syncing node %s", nodeName)
 
 	err := clientretry.RetryOnConflict(UpdateNodeSpecBackoff, func() error {
```
andrewsykim, line 423: This call to update node conditions for GCE should be moved after the cloud taint check

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -347,60 +439,64 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
...
 	}
+	klog.Infof("Initializing node %s with cloud provider", nodeName)
 
-	curNode, err := cnc.kubeClient.CoreV1().Nodes().Get(context.TODO(), node.Name, metav1.GetOptions{})
+	curNode, err := cnc.nodeInformer.Lister().Get(nodeName)
 	if err != nil {
-		utilruntime.HandleError(fmt.Errorf("failed to get node %s: %v", node.Name, err))
-		return
+		utilruntime.HandleError(fmt.Errorf("failed to get node %s: %v", nodeName, err))
+		return err
```
andrewsykim, line 449: I think we want to return an nil error if the node is not found anymore since that means we no longer need to initialize it.
```
if err != nil {
     return nil
  }

  return err
}
```

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -347,60 +439,64 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
...
 	}
 
 	cloudTaint := getCloudTaint(curNode.Spec.Taints)
 	if cloudTaint == nil {
 		// Node object received from event had the cloud taint but was outdated,
 		// the node has actually already been initialized.
-		return
+		return fmt.Errorf("this node %s is registered without the cloud taint, will not process", nodeName)
```
andrewsykim, line 456: This should return a nil error as well, since a node without the cloud taint means it has been initialized already, so this sync event can be ignored.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -347,60 +439,64 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
...
 
 	// TODO: getNodeModifiersFromCloudProvider and updateNodeAddress both call cloud api to get instanceMetadata,
 	// get instanceMetadata and pass it to getNodeModifiersFromCloudProvider and updateNodeAddress which reduces api calls.
-	nodeModifiers, err := cnc.getNodeModifiersFromCloudProvider(ctx, curNode)
+	copyNode := curNode.DeepCopy()
+	nodeModifiers, err := cnc.getNodeModifiersFromCloudProvider(ctx, copyNode)
 	if err != nil {
-		utilruntime.HandleError(fmt.Errorf("failed to initialize node %s at cloudprovider: %v", node.Name, err))
-		return
+		utilruntime.HandleError(fmt.Errorf("failed to initialize node %s at cloudprovider: %v", nodeName, err))
```
andrewsykim, line 464: I think we can remove the utilruntime.HandleError  and just return a wrapped error
```
return fmt.Errorf("failed to get node modifiers from cloud provider: %v", err)
```

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -347,60 +439,64 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
...
 	})
 
 	err = clientretry.RetryOnConflict(UpdateNodeSpecBackoff, func() error {
-		curNode, err := cnc.kubeClient.CoreV1().Nodes().Get(context.TODO(), node.Name, metav1.GetOptions{})
+		curNode, err := cnc.nodeInformer.Lister().Get(nodeName)
 		if err != nil {
 			return err
 		}
 
+		copyNode := curNode.DeepCopy()
```
andrewsykim, line 478: nit: I think `newNode` might be a better name here.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -347,60 +439,64 @@ func (cnc *CloudNodeController) initializeNode(ctx context.Context, node *v1.Nod
...
 		// So that users do not see any significant delay in IP addresses being filled into the node
-		cnc.updateNodeAddress(ctx, curNode)
+		cnc.updateNodeAddress(ctx, copyNode)
 
-		klog.Infof("Successfully initialized node %s with cloud provider", node.Name)
+		klog.Infof("Successfully initialized node %s with cloud provider", nodeName)
 		return nil
 	})
 	if err != nil {
 		utilruntime.HandleError(err)
```
andrewsykim, line 496: This call to `utilruntime.HandleError` can be removed as well since it will be called from the caller of syncNode

staging/src/k8s.io/cloud-provider/controllers/node/node_controller_test.go
```diff
@@ -40,6 +40,10 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+var (
+	alwaysReady = func() bool { return true }
```
andrewsykim, line 44: I think we can remove this since we don't need to set `nodesSynced` to test`syncNode`?

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
HaibaraAi96: gotcha, let me update that

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
HaibaraAi96: 👌🏻

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
HaibaraAi96: Can I know what's the main difference here? I feel adding **utilruntime.HandleError** is a safer way 🤔

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
HaibaraAi96: I thought this function just log something and the errors may be sent for analysis, isn't that a safer way?

staging/src/k8s.io/cloud-provider/controllers/node/node_controller_test.go
HaibaraAi96: ooops, I deleted it once but forget to commit haha, let me update!

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
HaibaraAi96: @andrewsykim If I just move the check like this, the unit test `TestGCECondition` is always failing, the node's status won't update correspondingly, am I missing something? 🤔 
```
klog.Infof("Syncing node %s", nodeName)

	curNode, err := cnc.nodeInformer.Lister().Get(nodeName)
	if err != nil {
			return nil
		}

		return err
	}

	cloudTaint := getCloudTaint(curNode.Spec.Taints)
	if cloudTaint == nil {
		// Node object received from event had the cloud taint but was outdated,
		// the node has actually already been initialized, so this sync event can be ignored.
		return nil
	}

	err = clientretry.RetryOnConflict(UpdateNodeSpecBackoff, func() error {
...
}
```

andrewsykim: I think the nodeLister is not being set in the controller on line 1369 of `node_controller_test.go `

HaibaraAi96: @andrewsykim I think you are right, I wasn't fully&correctly set up the lister in `Test_syncNode()` so when I copy and paste it into these two unit tests, it's always failing

staging/src/k8s.io/cloud-provider/controllers/node/node_controller_test.go
```diff
@@ -1294,10 +1294,18 @@ func Test_AddCloudNode(t *testing.T) {
...
 				nodeStatusUpdateFrequency: 1 * time.Second,
 			}
+
+			// activate node informer
+			factory.Core().V1().Nodes().Informer()
+			factory.Start(nil)
+			factory.WaitForCacheSync(nil)
+
+			cloudNodeController.nodesLister = factory.Core().V1().Nodes().Lister()
+
```
HaibaraAi96, line 1304: @andrewsykim I use this way to set up the lister, not sure if this is correct but I got unit tests passing, let me know if I need to improve anything :)

andrewsykim: Looks correct but I think it can be a bit cleaner by setting node lister when `CloudNodeController` is initialized. I believe calling `factory.Core().V1().Nodes().Lister()` will also activate the informer.
```go
			cloudNodeController := &CloudNodeController{
				kubeClient:                clientset,
				nodeInformer:              factory.Core().V1().Nodes(),
                                nodeLister:               factory.Core().V1().Nodes().Lister(),
				 cloud:                     test.fakeCloud,
				recorder:                  eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: "cloud-node-controller"}),
				nodeStatusUpdateFrequency: 1 * time.Second,
			}

			factory.Start(nil)
			factory.WaitForCacheSync(nil)
```

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -120,38 +127,118 @@ func NewCloudNodeController(
...
+		if err := cnc.syncHandler(key); err != nil {
+			// Put the item back on the workqueue to handle any transient errors.
+			cnc.workqueue.AddRateLimited(key)
+			return fmt.Errorf("error syncing '%s': %s, requeuing", key, err.Error())
+		}
+
+		// Finally, if no error occurs we Forget this item so it does not
+		// get queued again until another change happens.
+		cnc.workqueue.Forget(obj)
+		klog.V(4).Infof("Successfully synced %q", key)
```
andrewsykim, line 208: can we actually remove this log line since we're already logging this in `initializeNode`?

HaibaraAi96: sure, let me update!

staging/src/k8s.io/cloud-provider/controllers/node/node_controller_test.go
```diff
@@ -1452,14 +1462,19 @@ func TestGCECondition(t *testing.T) {
...
 		nodeInformer:              factory.Core().V1().Nodes(),
+		nodesLister:               factory.Core().V1().Nodes().Lister(),
 		cloud:                     fakeCloud,
 		recorder:                  eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: "cloud-node-controller"}),
 		nodeStatusUpdateFrequency: 1 * time.Second,
 	}
+
+	factory.Start(nil)
+	factory.WaitForCacheSync(nil)
+
```
HaibaraAi96, line 1473: @andrewsykim I got a wired thing, even if I set up the lister correctly, I still get 
```
--- FAIL: TestGCEConditionV2 (0.10s)
    node_controller_test.go:1403:
        	Error Trace:	node_controller_test.go:1403
        	Error:      	Should be true
        	Test:       	TestGCEConditionV2
        	Messages:   	Network Route Condition for GCE not added by external cloud initializer
I0915 13:56:13.190000   55392 node_controller.go:438] Initializing node node0 with cloud provider
I0915 13:56:13.190204   55392 node_controller.go:472] Successfully initialized node node0 with cloud provider
I0915 13:56:13.190262   55392 event.go:282] Event(v1.ObjectReference{Kind:"Node", Namespace:"", Name:"node0", UID:"", APIVersion:"v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'Synced' Node synced successfully
--- FAIL: TestGCECondition (0.10s)
    node_controller_test.go:1491:
        	Error Trace:	node_controller_test.go:1491
        	Error:      	Should be true
        	Test:       	TestGCECondition
        	Messages:   	Network Route Condition for GCE not added by external cloud initializer
FAIL
FAIL	k8s.io/cloud-provider/controllers/node	1.959s
FAIL
```
But the logs look fine
And when I trigger the test directly from my IDE, I got the tests running correctly, but running "go test ./..." commands will show the failing tests with above message

HaibaraAi96: and https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/94736/pull-kubernetes-bazel-test/1305973363159601155 logs the error 🤔 But not sure where goes wrong

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -295,112 +395,83 @@ func (cnc *CloudNodeController) updateNodeAddress(ctx context.Context, node *v1.
...
+		_, err = cnc.kubeClient.CoreV1().Nodes().Update(context.TODO(), newNode, metav1.UpdateOptions{})
 		if err != nil {
 			return err
 		}
 
+		if cnc.cloud.ProviderName() == "gce" {
+			// TODO(wlan0): Move this logic to the route controller using the node taint instead of condition
+			// Since there are node taints, do we still need this?
+			// This condition marks the node as unusable until routes are initialized in the cloud provider
+			if err := cloudnodeutil.SetNodeCondition(cnc.kubeClient, types.NodeName(nodeName), v1.NodeCondition{
```
andrewsykim, line 451: I think we actually want the GCE network unavailable condition to happen before the Update on line 442, since we want the condition in place before the cloud provider init taint is removed. Otherwise there will be a brief period where the node doesn't have the cloud taint and the condition.

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -120,38 +127,117 @@ func NewCloudNodeController(
...
+
+	// Wait for the caches to be synced before starting workers
+	klog.Info("Waiting for informer caches to sync")
+	if ok := cache.WaitForCacheSync(stopCh, cnc.nodesSynced); !ok {
+		klog.Errorf("failed to wait for caches to sync")
+		return
+	}
 
-	// The following loops run communicate with the APIServer with a worst case complexity
+	// The periodic  loop for updateNodeStatus communicates with the APIServer with a worst case complexity
```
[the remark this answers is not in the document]
HaibaraAi96: wow good catch!

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
HaibaraAi96: @andrewsykim The unit tests would fail if I move GCE network unavailable condition to happen before the **Update**
Not sure why it causes the error, still investigating 🤔

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -273,122 +368,99 @@ func (cnc *CloudNodeController) updateNodeAddress(ctx context.Context, node *v1.
...
 	nodeModifiers = append(nodeModifiers, func(n *v1.Node) {
 		n.Spec.Taints = excludeCloudTaint(n.Spec.Taints)
 	})
 
 	err = clientretry.RetryOnConflict(UpdateNodeSpecBackoff, func() error {
-		curNode, err := cnc.kubeClient.CoreV1().Nodes().Get(context.TODO(), node.Name, metav1.GetOptions{})
-		if err != nil {
-			return err
+		var curNode *v1.Node
+
```
andrewsykim, line 414: nit: remove line here

andrewsykim (commented):
/approve
/lgtm

Thanks @HaibaraAi96!

staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go
```diff
@@ -273,122 +368,98 @@ func (cnc *CloudNodeController) updateNodeAddress(ctx context.Context, node *v1.
...
 	nodeModifiers = append(nodeModifiers, func(n *v1.Node) {
 		n.Spec.Taints = excludeCloudTaint(n.Spec.Taints)
 	})
 
 	err = clientretry.RetryOnConflict(UpdateNodeSpecBackoff, func() error {
-		curNode, err := cnc.kubeClient.CoreV1().Nodes().Get(context.TODO(), node.Name, metav1.GetOptions{})
-		if err != nil {
-			return err
+		var curNode *v1.Node
+		if cnc.cloud.ProviderName() == "gce" {
```
andrewsykim, line 414: FYI @cheftako we should probably discuss getting rid of this custom check for GCE, I can't recall the initial reason behind the GCE specific check. Would probably be nice to have something generic here.

cheftako: +1. Most of these checks are because the test needs something which wasn't generally available. Not sure about this particular test but I've written a few of these. Its frequently something like the webhook tests. You need to create a special test container and deploy it to your registry. The intention was always that other cloud providers would follow adding the container to their registry.

Issue #84226, opened 2019-10-23
Check if volume is attached before deleting it


jsafrane:
**What would you like to be added**:
PV controller should not try to delete volumes that are attached. Right now, it tries to delete a volume as soon as PVC is deleted, which is too early. When user deletes a namespace, a pod and PVC is deleted at the same time, causing unnecessary delete attempts when the volume is still attached to a node. VolumeInUseProtection won't help us, as a Pod object is deleted before its volumes are detached.

We've seen this in CI both on AWS and on GCE, I assume other clouds behave similarly.

**Why is this needed**:
* It will reduce pressure on cloud API, as we try to delete the volume quite often (at least few initial retries are fast, until exp. backoff is high enough). On AWS, it would reduce our DescribeVolumes usage.
* It will reduce event noise / PVC being in Failed state.
* It will reduce logging noise.

A simple implementation could pass Node informer into PV controller and PV controller could check `node.status.volumesAttached` before deleting a volume. The largest change would be actually in the unit test framework.

/sig storage

@gnufied @msau42 @jingxu97 @wongma7, what do you think? Any volunteer for implementation? :-)

msau42:
What if you put it in protection controller instead so we avoid having to make significant changes to PV controller?

wongma7:
Agree. Another idea I was thinking is now that we have volumeAttachment objects we could use those for synchronization between attach/detach controller and pv controller, but actually using `node.status.volumesAttached` is much simpler and works for all plugins, not just csi.

As for putting it in the protection controller, I think it is achievable but it will be in a roundabout way.

If we use protection controller, the logic in PV controller for when to delete a backing volume will be the same: if PV.state==Released && ReclaimPolicy==Delete. I guess protection controller would somehow prevent PV.state from getting to Released until the volume is detached?

If we use PV controller, the logic will change to be: if PV.state==Released && ReclaimPolicy==Delete && PV is not present in any node.status.volumesAttached. I don't think this is too scary/big of a change compared to changing protection controller.

To the point "It will reduce logging noise:" we had a customer who saw these many "deletion failed because volume is still attached" and thought they were a "real" error. Tried to give a technical explanation and they (and some other folks) were confused why pv controller and a/d controller don't communicate with each other...

msau42:
[...]

Issue #77258, opened 2019-04-30
Can't remove terminating persistent volumes (PV)


XzenTorXz:
**What happened**:

The pod and namespaces got removed, but sometimes the pv still remains in "Terminating" status even with `ReclaimPolicy` `Delete`.

![grafik](https://user-images.githubusercontent.com/3819224/56957908-f395aa00-6b48-11e9-930a-90606adeebd9.png)

```
Name:            pvc-3252ac79-674e-11e9-9651-5404a6f17789
Labels:          <none>
Annotations:     pv.kubernetes.io/provisioned-by: rancher.io/longhorn
Finalizers:      [external-attacher/io-rancher-longhorn]
StorageClass:    longhorn1rep
Status:          Terminating (lasts <invalid>)
Claim:           sentry/sentry-data
Reclaim Policy:  Delete
Access Modes:    RWO
VolumeMode:      Filesystem
Capacity:        10Gi
Node Affinity:   <none>
Message:
Source:
    Type:              CSI (a Container Storage Interface (CSI) volume source)
    Driver:            io.rancher.longhorn
    VolumeHandle:      pvc-3252ac79-674e-11e9-9651-5404a6f17789
    ReadOnly:          false
    VolumeAttributes:      numberOfReplicas=1
                           staleReplicaTimeout=60
                           storage.kubernetes.io/csiProvisionerIdentity=1556113772288-8081-rancher.io/longhorn
Events:                <none>
```
I tried to delete it with `kubectl delete pv pvc-3252ac79-674e-11e9-9651-5404a6f17789 --force --grace-period=0` but it gets stuck and remains terminating.

**What you expected to happen**:

PV's should get removed.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.7", GitCommit:"6f482974b76db3f1e0f5d24605a9d1d38fad9a2b", GitTreeState:"clean", BuildDate:"2019-03-25T02:52:13Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}
[...]

Issue #66387, opened 2018-07-19
Pod with NFS volume attached cannot be terminated


AlexMuresan:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:
When trying to delete a nfs-volume pod it gets stuck in "Terminating" state.

**What you expected to happen**:
The pod should be deleted

**How to reproduce it (as minimally and precisely as possible)**:
1. Create one namespace called:  e2e-tests-kubelet-bugtest1
2. Create a second namespace called: e2e-tests-kubelet-bugtest2
3. Create a new nfs-server pod using the nfs-server.yaml
4. Add the ip of the nfs-server pod to the pod-nfs-vol.yaml
5. Create a second pod using the pod-nfs-volume.yaml
6. Delete the nfs-server pod without stopping the server
7. Try to delete pod-nfs-volume pod.

**Anything else we need to know?**:
Recreating this using YAML files is easier (and possibly more accurate):

- nfs-server: https://paste.ubuntu.com/p/ZC2dj4SQbh/
- pod-nfs-volume: https://paste.ubuntu.com/p/YVd6Txc299/

**Environment**:
- Kubernetes version (use `kubectl version`):

    - Client Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.3", GitCommit:"d2835416544f298c919e2ead3be3d0864b52323b", GitTreeState:"clean", BuildDate:"2018-02-07T12:22:21Z", GoVersion:"go1.9.2", Compiler:"gc", Platform:"linux/amd64"}

    - Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.1", GitCommit:"b1b29978270dc22fecc592ac55d903350454310a", GitTreeState:"clean", BuildDate:"2018-07-17T18:43:26Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"} 

- Cloud provider or hardware configuration:
    - Windows server 1803 nodes

- OS (e.g. from /etc/os-release):
    - NAME="Ubuntu"
    - VERSION="16.04.3 LTS (Xenial Xerus)"
    - ID=ubuntu
    - ID_LIKE=debian
    - PRETTY_NAME="Ubuntu 16.04.4 LTS"
    - VERSION_ID="16.04"

- Kernel (e.g. `uname -a`): 
[...]

PR #94737, opened 2020-09-12
Fix race condition between CSI NodeUnstage and DeleteVolume calls

### LINKED ISSUE #94108: Controller delete volume is called before node unstage volume returns success
**What happened**:
We are seeing an issue where Controller issued Controller Delete volume while node unstage volume was still in flight but unpublish volume had returned success. This is a block volume on K8s 1.18.2. Is this the desired behavior ? Or is there a bug open on this ? According to the spec, my understanding was controller delete volume would be issued after node unstage volume returned success
The problem is unstage volume makes a call to the backend storage provider but since delete volume was issued before that. The volume gets deleted in the backend and the unstage volume (or umount) never finishes. We are left with stale devices on the K8s nodes

Controller delete volume to be issued only after node unstage volume returns success

I think its a race condition. Not sure how to reproduce it.

NodeUnpublish volume does - $umount on TargetPath and then $rm -rf on TargetPath. Possible that K8s is looking at this mount point and decided unstage happened and issued delete volume. 

**Environment**:
- Kubernetes version (use `kubectl version`):1.18.2

**Log snippets**:
- Unpublish success:  
Aug 18 23:08:41 c3-k8s-n5 kubelet[998]: I0818 23:08:41.303322 998 operation_generator.go:1145] UnmapVolume succeeded for volume "kubernetes.io/csi/csi-hxcsi^2b6fdb11-ebbd-4847-a089-e3d846cc808a" (OuterVolumeSpecName: "test-volume-block") pod "e4bbf342-a446-4fb3-bd7f-ab5c17fe269f"

- Unstage invoked:
Aug 18 23:08:41 c3-k8s-n5 kubelet[998]: I0818 23:08:41.389774 998 reconciler.go:312] operationExecutor.UnmountDevice started for volume "pvc-98407c6c-9e3c-4d27-8728-1859f9cffde8" (UniqueName: "kubernetes.io/csi/csi-hxcsi^2b6fdb11-ebbd-4847-a089-e3d846cc808a") on node "c3-k8s-n5"

- Unstage failed: 
Aug 18 23:08:43 c3-k8s-n5 kubelet[998]: E0818 23:08:43.238461 998 nestedpendingoperations.go:301] Operation for "{volumeName:kubernetes.io/csi/csi-hxcsi^2b6fdb11-ebbd-4847-a089-e3d846cc808a podName: nodeName:}" failed.

- Delete Volume invoked:
{"log":"I0819 06:08:41.821406 1 controller.go:1311] delete \"pvc-98407c6c-9e3c-4d27-8728-1859f9cffde8\": started\n","stream":"stderr","time":"2020-08-19T06:08:41.821566667Z"}

saikat-royc:
**What type of PR is this?**

This PR aims to fix the race condition between NodeUnstage and DeleteVolume calls.

Fixes #94108 

For volumes which are unpublished but still staged on the node, we can prevent the delete volume race by ensuring the node.status.volumesInUse captures volumes in such state. This acts as a barrier for PVC protection controller to not remove the finalizer on the PVC until its completely unstaged from the node.

```release-note
Fix race condition between CSI NodeUnstage and DeleteVolume calls
```

```docs

```

### DIFF
--- a/pkg/controller/volume/pvcprotection/pvc_protection_controller.go
+++ b/pkg/controller/volume/pvcprotection/pvc_protection_controller.go
@@ -19,11 +19,13 @@ package pvcprotection
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	coreinformers "k8s.io/client-go/informers/core/v1"
@@ -51,6 +53,12 @@ type Controller struct {
 	podListerSynced cache.InformerSynced
 	podIndexer      cache.Indexer
 
+	nodeLister       corelisters.NodeLister
+	nodeListerSynced cache.InformerSynced
+
+	pvLister       corelisters.PersistentVolumeLister
+	pvListerSynced cache.InformerSynced
+
 	queue workqueue.RateLimitingInterface
 
 	// allows overriding of StorageObjectInUseProtection feature Enabled/Disabled for testing
@@ -60,8 +68,13 @@ type Controller struct {
 	genericEphemeralVolumeFeatureEnabled bool
 }
 
+const (
+	volumeHandleCSIPrefix = "kubernetes.io/csi/"
+	volumeHandleSep       = "^"
+)
+
 // NewPVCProtectionController returns a new instance of PVCProtectionController.
-func NewPVCProtectionController(pvcInformer coreinformers.PersistentVolumeClaimInformer, podInformer coreinformers.PodInformer, cl clientset.Interface, storageObjectInUseProtectionFeatureEnabled, genericEphemeralVolumeFeatureEnabled bool) (*Controller, error) {
+func NewPVCProtectionController(pvcInformer coreinformers.PersistentVolumeClaimInformer, podInformer coreinformers.PodInformer, nodeInformer coreinformers.NodeInformer, pvInformer coreinformers.PersistentVolumeInformer, cl clientset.Interface, storageObjectInUseProtectionFeatureEnabled, genericEphemeralVolumeFeatureEnabled bool) (*Controller, error) {
 	e := &Controller{
 		client:                               cl,
 		queue:                                workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "pvcprotection"),
@@ -99,6 +112,11 @@ func NewPVCProtectionController(pvcInformer coreinformers.PersistentVolumeClaimI
 		},
 	})
 
+	e.nodeLister = nodeInformer.Lister()
+	e.nodeListerSynced = nodeInformer.Informer().HasSynced
+
+	e.pvLister = pvInformer.Lister()
+	e.pvListerSynced = pvInformer.Informer().HasSynced
 	return e, nil
 }
 
@@ -110,7 +128,7 @@ func (c *Controller) Run(workers int, stopCh <-chan struct{}) {
 	klog.Infof("Starting PVC protection controller")
 	defer klog.Infof("Shutting down PVC protection controller")
 
-	if !cache.WaitForNamedCacheSync("PVC protection", stopCh, c.pvcListerSynced, c.podListerSynced) {
+	if !cache.WaitForNamedCacheSync("PVC protection", stopCh, c.pvcListerSynced, c.podListerSynced, c.nodeListerSynced, c.pvListerSynced) {
 		return
 	}
 
@@ -140,19 +158,32 @@ func (c *Controller) processNextWorkItem() bool {
 		return true
 	}
 
-	err = c.processPVC(pvcNamespace, pvcName)
-	if err == nil {
+	// 'needsReque' is set to true in two cases:
+	// 1. Errors (other than NotFound error) while retrieving PVC from cache.
+	// 2. if PV is detected to be in use by a node, and there is no Pod referencing the
+	// PVC(which is marked for delete). We expect the condition to be short lived as the
+	// volume should be detached from the node soon.
+	// For the above cases, we retry without any exponential backup. Even if we do not use
+	// exponential backoff, default ratelimiter still uses the BucketRateLimiter to control
+	// the overall rate at which keys are added to the queue.
+	needsReque, err := c.processPVC(pvcNamespace, pvcName)
+	if err == nil && !needsReque {
 		c.queue.Forget(pvcKey)
 		return true
+	} else if needsReque {
+		c.queue.Forget(pvcKey)
 	}
 
-	utilruntime.HandleError(fmt.Errorf("PVC %v failed with : %v", pvcKey, err))
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("PVC %v failed with : %v", pvcKey, err))
+	}
+	klog.V(4).Infof("Reque PVC key %v", pvcKey)
 	c.queue.AddRateLimited(pvcKey)
-
 	return true
 }
 
-func (c *Controller) processPVC(pvcNamespace, pvcName string) error {
+// This function processes a PVC, and returns error encountered (if any), and a bool indicating if the PVC needs to be requeued.
+func (c *Controller) processPVC(pvcNamespace, pvcName string) (bool, error) {
 	klog.V(4).Infof("Processing PVC %s/%s", pvcNamespace, pvcName)
 	startTime := time.Now()
 	defer func() {
@@ -162,23 +193,25 @@ func (c *Controller) processPVC(pvcNamespace, pvcName string) error {
 	pvc, err := c.pvcLister.PersistentVolumeClaims(pvcNamespace).Get(pvcName)
 	if apierrors.IsNotFound(err) {
 		klog.V(4).Infof("PVC %s/%s not found, ignoring", pvcNamespace, pvcName)
-		return nil
+		return false, nil
 	}
 	if err != nil {
-		return err
+		// We expect this error to be resolved soon, hence requeue the PVC key.
+		return true, err
 	}
 
 	if protectionutil.IsDeletionCandidate(pvc, volumeutil.PVCProtectionFinalizer) {
 		// PVC should be deleted. Check if it's used and remove finalizer if
 		// it's not.
-		isUsed, err := c.isBeingUsed(pvc)
+		isUsed, needsRequeue, err := c.isBeingUsed(pvc)
 		if err != nil {
-			return err
+			return needsRequeue, err
 		}
 		if !isUsed {
-			return c.removeFinalizer(pvc)
+			return needsRequeue, c.removeFinalizer(pvc)
 		}
 		klog.V(2).Infof("Keeping PVC %s/%s because it is still being used", pvc.Namespace, pvc.Name)
+		return needsRequeue, nil
 	}
 
 	if protectionutil.NeedToAddFinalizer(pvc, volumeutil.PVCProtectionFinalizer) {
@@ -186,9 +219,9 @@ func (c *Controller) processPVC(pvcNamespace, pvcName string) error {
 		// finalizer should be added by admission plugin, this is just to add
 		// the finalizer to old PVCs that were created before the admission
 		// plugin was enabled.
-		return c.addFinalizer(pvc)
+		return false, c.addFinalizer(pvc)
 	}
-	return nil
+	return false, nil
 }
 
 func (c *Controller) addFinalizer(pvc *v1.PersistentVolumeClaim) error {
@@ -219,25 +252,92 @@ func (c *Controller) removeFinalizer(pvc *v1.PersistentVolumeClaim) error {
 	return nil
 }
 
-func (c *Controller) isBeingUsed(pvc *v1.PersistentVolumeClaim) (bool, error) {
+// This function checks if a PVC is in use and returns values inUse,
+// needsRequeue and error.
+// isUse: This value indicates if a PVC is in use by any pod or node
+// needsRequeue: This value indicates that the controller loop should
+// requeue the PVC in the controller, instead of waiting for a new event
+// handler call on the PVC. This is an optimization to react faster to a
+// change in state where no pod references the PVC.
+func (c *Controller) isBeingUsed(pvc *v1.PersistentVolumeClaim) (bool, bool, error) {
+	// If PVC is marked for delete, before the PVC.Spec.Volume is populated,
+	// we should not block PVC deletion. By the time PVC is deleted, if PV
+	// creation was already triggered and the volume plugin is in the process
+	// of creation of the volume, then after the PV is created, the PV controller
+	// will trigger a delete for the PV (if it has a reclaim policy of delete),
+	// since the claim will not be found.
+	var pv *v1.PersistentVolume
+	if pvc.Spec.VolumeName != "" {
+		var err error
+		pv, err = c.getPV(pvc)
+		// If PV is not found, then node check for volumes in use will be skipped.
+		if err != nil && !apierrors.IsNotFound(err) {
+			return false, false, err
+		}
+	}
+
 	// Look for a Pod using pvc in the Informer's cache. If one is found the
 	// correct decision to keep pvc is taken without doing an expensive live
-	// list.
-	if inUse, err := c.askInformer(pvc); err != nil {
+	// list. No requeue is needed as we will wait for a Pod event.
+	if inUse, err := c.askPodInformer(pvc); err != nil {
 		// No need to return because a live list will follow.
 		klog.Error(err)
 	} else if inUse {
-		return true, nil
+		// If we reach here, this means, the the PVC is still in use by a Pod, hence no
+		// reque is needed. We will wait for an event to process the PVC.
+		return true, false, nil
+	}
+
+	// Check for volumes in use by nodes for CSI volume plugins.
+	// In-tree volume plugins need investigation on how to map a pvc.Spec.VolumeName
+	// to node.Status.VolumesInUse.
+	if pv != nil && pv.Spec.CSI != nil {
+		uniqueVolHandle, err := c.generateUniqueCSIVolumeHandle(pv.Spec.CSI)
+		if err != nil {
+			return false, false, err
+		}
+		if inUse, err := c.askNodeInformer(uniqueVolHandle, pvc); err != nil {
+			// No need to return because a live list will follow.
+			klog.Error(err)
+		} else if inUse {
+			// If we reach here, this means we have not found any Pod , referencing this PVC,
+			// in the informer cache, and the volume is still in use by a node. So we reque
+			// the PVC again to process it faster.
+			return true, true, nil
+		}
 	}
 
 	// Even if no Pod using pvc was found in the Informer's cache it doesn't
 	// mean such a Pod doesn't exist: it might just not be in the cache yet. To
 	// be 100% confident that it is safe to delete pvc make sure no Pod is using
 	// it among those returned by a live list.
-	return c.askAPIServer(pvc)
+	if inUse, err := c.askAPIServerForPod(pvc); err != nil {
+		return false, false, err
+	} else if inUse {
+		// If we reach here, this means, the the PVC is still in use by a Pod, hence no
+		// reque is needed. We will wait for an event to process the PVC.
+		return true, false, nil
+	}
+
+	if pv != nil && pv.Spec.CSI != nil {
+		uniqueVolHandle, err := c.generateUniqueCSIVolumeHandle(pv.Spec.CSI)
+		if err != nil {
+			return false, false, err
+		}
+		if inUse, err := c.askAPIServerForNode(uniqueVolHandle, pvc); err != nil {
+			return false, false, err
+		} else if inUse {
+			// If we reach here, this means we have not found any Pod , referencing this PVC,
+			// and the volume is still in use by a node. So we reque the PVC again to process
+			// it faster.
+			return true, true, nil
+		}
+	}
+
+	return false, false, nil
 }
 
-func (c *Controller) askInformer(pvc *v1.PersistentVolumeClaim) (bool, error) {
+func (c *Controller) askPodInformer(pvc *v1.PersistentVolumeClaim) (bool, error) {
 	klog.V(4).Infof("Looking for Pods using PVC %s/%s in the Informer's cache", pvc.Namespace, pvc.Name)
 
 	// The indexer is used to find pods which might use the PVC.
@@ -274,7 +374,7 @@ func (c *Controller) askInformer(pvc *v1.PersistentVolumeClaim) (bool, error) {
 	return false, nil
 }
 
-func (c *Controller) askAPIServer(pvc *v1.PersistentVolumeClaim) (bool, error) {
+func (c *Controller) askAPIServerForPod(pvc *v1.PersistentVolumeClaim) (bool, error) {
 	klog.V(4).Infof("Looking for Pods using PVC %s/%s with a live list", pvc.Namespace, pvc.Name)
 
 	podsList, err := c.client.CoreV1().Pods(pvc.Namespace).List(context.TODO(), metav1.ListOptions{})
@@ -417,4 +517,61 @@ func (c *Controller) enqueuePVCs(pod *v1.Pod, deleted bool) {
 			c.queue.Add(pod.Namespace + "/" + pod.Name + "-" + volume.Name)
 		}
 	}
+}
+
+func (c *Controller) getPV(pvc *v1.PersistentVolumeClaim) (*v1.PersistentVolume, error) {
+	pv, err := c.pvLister.Get(pvc.Spec.VolumeName)
+	if err != nil {
+		klog.V(3).Infof("unexpected error getting persistent volume %q from cache for claim %s/%s: %v", pvc.Spec.VolumeName, pvc.Namespace, pvc.Name, err)
+		pv, err = c.client.CoreV1().PersistentVolumes().Get(context.TODO(), pvc.Spec.VolumeName, metav1.GetOptions{})
+	}
+	return pv, err
+}
+
+func (c *Controller) askNodeInformer(volumeHandle string, pvc *v1.PersistentVolumeClaim) (bool, error) {
+	klog.V(3).Infof("For claim %s/%s checking for volume handle %s in node informer cache", pvc.Namespace, pvc.Name, volumeHandle)
+	if volumeHandle == "" {
+		return false, fmt.Errorf("Empty volume handle for claim %s/%s", pvc.Namespace, pvc.Name)
+	}
+
+	nodes, err := c.nodeLister.List(labels.Everything())
+	if err != nil {
+		return false, fmt.Errorf("unexpected error listing nodes from node informer: %v", err)
+	}
+	return c.isVolumeInUse(volumeHandle, nodes), nil
+}
+
+func (c *Controller) askAPIServerForNode(volumeHandle string, pvc *v1.PersistentVolumeClaim) (bool, error) {
+	klog.V(3).Infof("For claim %s/%s, volume %s, list nodes from API server", pvc.Namespace, pvc.Name, volumeHandle)
+	nodes, err := c.client.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return false, fmt.Errorf("List of nodes from API server failed: %s", err.Error())
+	}
+	nodesPtr := []*v1.Node{}
+	for i := range nodes.Items {
+		nodesPtr = append(nodesPtr, &nodes.Items[i])
+	}
+
+	return c.isVolumeInUse(volumeHandle, nodesPtr), nil
+}
+
+// Helper function to determine if a given volume is present in the node's volumesInUse list.
+func (c *Controller) isVolumeInUse(volHandle string, nodes []*v1.Node) bool {
+	for _, node := range nodes {
+		for _, volInUse := range node.Status.VolumesInUse {
+			volInUseStr := strings.TrimPrefix(string(volInUse), volumeHandleCSIPrefix)
+			if volInUseStr == volHandle {
+				klog.Infof("Found volume %s in use at node %s", volInUseStr, node.Name)
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func (c *Controller) generateUniqueCSIVolumeHandle(source *v1.CSIPersistentVolumeSource) (string, error) {
+	if source.Driver == "" || source.VolumeHandle == "" {
+		return "", fmt.Errorf("Failed to fetch CSI driver plugin or volume for CSI source %+v", source)
+	}
+	return fmt.Sprintf("%s%s%s", source.Driver, volumeHandleSep, source.VolumeHandle), nil
 }
--- a/pkg/controller/volume/pvcprotection/pvc_protection_controller_test.go
+++ b/pkg/controller/volume/pvcprotection/pvc_protection_controller_test.go
@@ -18,6 +18,7 @@ package pvcprotection
 
 import (
 	"errors"
+	"fmt"
 	"reflect"
 	"testing"
 	"time"
@@ -46,12 +47,32 @@ type reaction struct {
 }
 
 const (
-	defaultNS       = "default"
-	defaultPVCName  = "pvc1"
-	defaultPodName  = "pod1"
-	defaultNodeName = "node1"
-	defaultUID      = "uid1"
+	defaultNS           = "default"
+	defaultPVCName      = "pvc1"
+	defaultPodName      = "pod1"
+	defaultNodeName     = "node1"
+	defaultUID          = "uid1"
+	defaultVolumeName   = "test-vol"
+	otherVolumeName     = "other-vol"
+	defaultPVName       = "test-vol"
+	defaultCSIDriver    = "test-csi-driver"
+	testCSIVolumePrefix = volumeHandleCSIPrefix + defaultCSIDriver + volumeHandleSep
 )
+
+func withVolInUse(n *v1.Node, volumeInUse string) *v1.Node {
+	n.Status = v1.NodeStatus{
+		VolumesInUse: []v1.UniqueVolumeName{v1.UniqueVolumeName(volumeInUse)},
+	}
+	return n
+}
+
+func node() *v1.Node {
+	return &v1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: defaultNodeName,
+		},
+	}
+}
 
 func pod() *v1.Pod {
 	return &v1.Pod{
@@ -108,13 +129,35 @@ func withUID(uid types.UID, pod *v1.Pod) *v1.Pod {
 	return pod
 }
 
+func pv(withCsiSource bool) *v1.PersistentVolume {
+	pv := &v1.PersistentVolume{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: defaultPVName,
+		},
+		Spec: v1.PersistentVolumeSpec{},
+	}
+	if withCsiSource {
+		pv.Spec.CSI = &v1.CSIPersistentVolumeSource{
+			VolumeHandle: defaultVolumeName,
+			Driver:       defaultCSIDriver,
+		}
+	}
+	return pv
+}
+
 func pvc() *v1.PersistentVolumeClaim {
 	return &v1.PersistentVolumeClaim{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      defaultPVCName,
 			Namespace: defaultNS,
 		},
+		Spec: v1.PersistentVolumeClaimSpec{},
 	}
+}
+
+func pvcWithVolumeName(pvc *v1.PersistentVolumeClaim) *v1.PersistentVolumeClaim {
+	pvc.Spec.VolumeName = "test-vol"
+	return pvc
 }
 
 func withProtectionFinalizer(pvc *v1.PersistentVolumeClaim) *v1.PersistentVolumeClaim {
@@ -146,6 +189,16 @@ func generateUpdateErrorFunc(t *testing.T, failures int) clienttesting.ReactionF
 	}
 }
 
+func generateNodeListErrorFunc(t *testing.T) clienttesting.ReactionFunc {
+	return func(action clienttesting.Action) (bool, runtime.Object, error) {
+		_, ok := action.(clienttesting.ListAction)
+		if !ok {
+			t.Fatalf("Reactor got non-list action: %+v", action)
+		}
+		return true, nil, fmt.Errorf("mock error")
+	}
+}
+
 func testPVCProtectionController(t *testing.T, genericEphemeralVolumeFeatureEnabled bool) {
 	pvcGVR := schema.GroupVersionResource{
 		Group:    v1.GroupName,
@@ -161,6 +214,21 @@ func testPVCProtectionController(t *testing.T, genericEphemeralVolumeFeatureEnab
 		Group:   v1.GroupName,
 		Version: "v1",
 		Kind:    "Pod",
+	}
+	nodeGVR := schema.GroupVersionResource{
+		Group:    v1.GroupName,
+		Version:  "v1",
+		Resource: "nodes",
+	}
+	nodeGVK := schema.GroupVersionKind{
+		Group:   v1.GroupName,
+		Version: "v1",
+		Kind:    "Node",
+	}
+	pvGVR := schema.GroupVersionResource{
+		Group:    v1.GroupName,
+		Version:  "v1",
+		Resource: "persistentvolumes",
 	}
 
 	tests := []struct {
@@ -179,17 +247,21 @@ func testPVCProtectionController(t *testing.T, genericEphemeralVolumeFeatureEnab
 		// Pod event to simulate. This Pod will be automatically added to
 		// initialObjects.
 		updatedPod *v1.Pod
+		// Node event to simulate. This Node will be automatically added to
+		// initialObjects.
+		updatedNode *v1.Node
 		// Pod event to simulate. This Pod is *not* added to
 		// initialObjects.
 		deletedPod *v1.Pod
+		// PV event to simulate. This PV will be automatically added to
+		// initialObjects.
+		updatedPV *v1.PersistentVolume
 		// List of expected kubeclient actions that should happen during the
 		// test.
 		expectedActions                     []clienttesting.Action
 		storageObjectInUseProtectionEnabled bool
 	}{
-		//
 		// PVC events
-		//
 		{
 			name:       "StorageObjectInUseProtection Enabled, PVC without finalizer -> finalizer is added",
 			updatedPVC: pvc(),
@@ -312,9 +384,7 @@ func testPVCProtectionController(t *testing.T, genericEphemeralVolumeFeatureEnab
 			},
 			storageObjectInUseProtectionEnabled: true,
 		},
-		//
 		// Pod events
-		//
 		{
 			name: "updated running Pod -> no action",
 			initialObjects: []runtime.Object{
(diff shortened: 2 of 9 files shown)


pkg/controller/volume/pvcprotection/pvc_protection_controller_test.go
```diff
@@ -400,6 +443,142 @@ func testPVCProtectionController(t *testing.T, genericEphemeralVolumeFeatureEnab
...
+			expectedActions: []clienttesting.Action{
+				// Pods listed, no pod found referencing the pvc
+				clienttesting.NewListAction(podGVR, podGVK, defaultNS, metav1.ListOptions{}),
+				// node informer is did not report a node with the given volume, hence nodes listed from API server
+				clienttesting.NewListAction(nodeGVR, nodeGVK, "", metav1.ListOptions{}),
+				//PVC finalizer removed.
+				clienttesting.NewUpdateAction(pvcGVR, defaultNS, deleted(pvcWithVolumeName(pvc()))),
+			},
+			storageObjectInUseProtectionEnabled: true,
+		},
```
mattcary, line 581: Do you have a test case where a node has the volume marked in use but the PVC isn't stamped with the volume name?

It shouldn't matter given the current code but might be useful to check in order to prevent regressions.

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -234,7 +246,17 @@ func (c *Controller) isBeingUsed(pvc *v1.PersistentVolumeClaim) (bool, error) {
 	// mean such a Pod doesn't exist: it might just not be in the cache yet. To
 	// be 100% confident that it is safe to delete pvc make sure no Pod is using
 	// it among those returned by a live list.
-	return c.askAPIServer(pvc)
+	inUse, err := c.askAPIServer(pvc)
+	if err != nil {
+		return false, err
```
jingxu97, line 251: I see the possible error is pods are failed to list. To be safer, I think we should not return false.

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -234,7 +246,17 @@ func (c *Controller) isBeingUsed(pvc *v1.PersistentVolumeClaim) (bool, error) {
...
+	}
+	if inUse {
+		return inUse, nil
+	}
+
+	// If we reach here, PVC is not referenced by any pod. Proceed with checking nodes
+	// to determine if the PV backed by the PVC is in use by any node.
+	return c.isPVCAttachedToNodes(pvc)
```
jingxu97, line 259: maybe call it isVolumeInUseByNode?

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -418,3 +440,62 @@ func (c *Controller) enqueuePVCs(pod *v1.Pod, deleted bool) {
...
+		if err != nil {
+			klog.V(3).Infof("unexpected error listing nodes from node informer: %v", err)
+		} else if c.isVolumeInUse(volName, nodes) {
+			return true, nil
+		}
+	}
+
+	// If we reach here, we could not determine if the volume is attached to any node.
+	// Fetch node list from API server to check node.Status.VolumesInUse.
+	nodes, err := c.client.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
```
jingxu97, line 478: I think normally if nodeLister.List failed, we return the error, and controller will try again. We want to avoid calling API server directly to avoid constant API calls in case of failure.

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
saikat-royc: Didn't quite understand the comment. If we return an error != nil, the bool value is ignored and function returns 
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/pvcprotection/pvc_protection_controller.go#L175

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
saikat-royc: Ack good point.

pkg/controller/volume/pvcprotection/pvc_protection_controller_test.go
saikat-royc: to understand correctly, PVC spec.VolumeName not set, and node.status.VolumesInUse = ['vol1'], not sure how this situation can arise. However I have added a test case, we skip the check for volumes in use since pvc spec volume name is missing.

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
saikat-royc: Done

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
jingxu97: Here I mean you can continue to check volumeInUse instead of return false? False means pvc is not being used, and finalizer will be removed, right?

saikat-royc: No since err != nil, finalizer will not be removed
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/pvcprotection/pvc_protection_controller.go#L176

jingxu97: got it. I am also thinking it is better to check volumeInUse first. The number of nodes should be much smaller than number of pods. Also askAPIServer is calling API server directly instead of using informer cache?

pkg/controller/volume/pvcprotection/pvc_protection_controller_test.go
mattcary: Yeah the point is to check the code path in case something changes in the future.

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
saikat-royc: Ack. Changed the order of the check : nodes checked first before pods.

saikat-royc:
Looking at one of the test failures, 
pull-kubernetes-e2e-gce-storage-slow
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/94737/pull-kubernetes-e2e-gce-storage-slow/1305608296467533825/

The test failed because the wait for pv deletion time is too less 30 secs. We should increase the timeout
test deleted the pvc at 21:28:29, but pvc protection controller found the volume to be still in use on a node.
```
I0914 21:28:28.9859129 pvc_protection_controller.go:379] Got event on PVC volume-provisioning-9528/pvc-6jc8g 
I0914 21:28:28.9859249 pvc_protection_controller.go:168] Processing PVC volume-provisioning-9528/pvc-6jc8g
I0914 21:28:28.9859339 pvc_protection_controller.go:464] For claim volume-provisioning-9528/pvc-6jc8g, checking for volume pvc-f36b6642-21d7-4f45-8a63-14f7f6747e0a in node informer cache
I0914 21:28:28.9859479 pvc_protection_controller.go:494] Found volume kubernetes.io/gce-pd/e2e-b91b9ffbe0-390e9-d-pvc-f36b6642-21d7-4f45-8a63-14f7f6747e0a in use at node e2e-b91b9ffbe0-390e9-minion-group-f4zc
I0914 21:28:28.9859539 pvc_protection_controller.go:193] Keeping PVC volume-provisioning-9528/pvc-6jc8g because it is still being used
I0914 21:28:28.9859599 pvc_protection_controller.go:171] Finished processing PVC volume-provisioning-9528/pvc-6jc8g (29.35µs)
```
Then at 21:30 pvc event was received and pvc finalizer was removed as expectecd.
```
I0914 21:30:38.3537499 pvc_protection_controller.go:230] Removed protection finalizer from PVC volume-provisioning-9528/pvc-6jc8g
I0914 21:30:38.3537729 pvc_protection_controller.go:171] Finished processing PVC volume-provisioning-9528/pvc-6jc8g (130.580021ms)
```

We need to increase the wait for pv delete timeout in some of the tests
e.g.
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/volume_provisioning.go#L418
similar to here
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/volume_provisioning.go#L590

jingxu97:
> Looking at one of the test failures,
> pull-kubernetes-e2e-gce-storage-slow
> https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/94737/pull-kubernetes-e2e-gce-storage-slow/1305608296467533825/
> 
> The test failed because the wait for pv deletion time is too less 30 secs. We should increase the timeout
> test deleted the pvc at 21:28:29, but pvc protection controller found the volume to be still in use on a node.
> 
> ```
> I0914 21:28:28.9859129 pvc_protection_controller.go:379] Got event on PVC volume-provisioning-9528/pvc-6jc8g 
> I0914 21:28:28.9859249 pvc_protection_controller.go:168] Processing PVC volume-provisioning-9528/pvc-6jc8g
> I0914 21:28:28.9859339 pvc_protection_controller.go:464] For claim volume-provisioning-9528/pvc-6jc8g, checking for volume pvc-f36b6642-21d7-4f45-8a63-14f7f6747e0a in node informer cache
> I0914 21:28:28.9859479 pvc_protection_controller.go:494] Found volume kubernetes.io/gce-pd/e2e-b91b9ffbe0-390e9-d-pvc-f36b6642-21d7-4f45-8a63-14f7f6747e0a in use at node e2e-b91b9ffbe0-390e9-minion-group-f4zc
> I0914 21:28:28.9859539 pvc_protection_controller.go:193] Keeping PVC volume-provisioning-9528/pvc-6jc8g because it is still being used
> I0914 21:28:28.9859599 pvc_protection_controller.go:171] Finished processing PVC volume-provisioning-9528/pvc-6jc8g (29.35µs)
> ```
> 
> Then at 21:30 pvc event was received and pvc finalizer was removed as expectecd.
> 
> ```
> I0914 21:30:38.3537499 pvc_protection_controller.go:230] Removed protection finalizer from PVC volume-provisioning-9528/pvc-6jc8g
> I0914 21:30:38.3537729 pvc_protection_controller.go:171] Finished processing PVC volume-provisioning-9528/pvc-6jc8g (130.580021ms)
> ```
> 
> We need to increase the wait for pv delete timeout in some of the tests
> e.g.
> https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/volume_provisioning.go#L418
> similar to here
> https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/volume_provisioning.go#L590

roughly how long it takes to unpublish and unstage finish?

saikat-royc:
> > Looking at one of the test failures,
> > pull-kubernetes-e2e-gce-storage-slow
> > https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/94737/pull-kubernetes-e2e-gce-storage-slow/1305608296467533825/
> > The test failed because the wait for pv deletion time is too less 30 secs. We should increase the timeout
> > test deleted the pvc at 21:28:29, but pvc protection controller found the volume to be still in use on a node.
> > ```
> > I0914 21:28:28.9859129 pvc_protection_controller.go:379] Got event on PVC volume-provisioning-9528/pvc-6jc8g 
> > I0914 21:28:28.9859249 pvc_protection_controller.go:168] Processing PVC volume-provisioning-9528/pvc-6jc8g
> > I0914 21:28:28.9859339 pvc_protection_controller.go:464] For claim volume-provisioning-9528/pvc-6jc8g, checking for volume pvc-f36b6642-21d7-4f45-8a63-14f7f6747e0a in node informer cache
> > I0914 21:28:28.9859479 pvc_protection_controller.go:494] Found volume kubernetes.io/gce-pd/e2e-b91b9ffbe0-390e9-d-pvc-f36b6642-21d7-4f45-8a63-14f7f6747e0a in use at node e2e-b91b9ffbe0-390e9-minion-group-f4zc
> > I0914 21:28:28.9859539 pvc_protection_controller.go:193] Keeping PVC volume-provisioning-9528/pvc-6jc8g because it is still being used
> > I0914 21:28:28.9859599 pvc_protection_controller.go:171] Finished processing PVC volume-provisioning-9528/pvc-6jc8g (29.35µs)
> > ```
> > 
> > 
> > Then at 21:30 pvc event was received and pvc finalizer was removed as expectecd.
> > ```
> > I0914 21:30:38.3537499 pvc_protection_controller.go:230] Removed protection finalizer from PVC volume-provisioning-9528/pvc-6jc8g
> > I0914 21:30:38.3537729 pvc_protection_controller.go:171] Finished processing PVC volume-provisioning-9528/pvc-6jc8g (130.580021ms)
> > ```
> > 
> > 
> > We need to increase the wait for pv delete timeout in some of the tests
> > e.g.
> > https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/volume_provisioning.go#L418
> > similar to here
> > https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/volume_provisioning.go#L590
> 
> roughly how long it takes to unpublish and unstage finish?

The behavior that I see is sometimes it takes time for the PVC AddUpdate event to be triggered. worst case ~6 mins.

i.e say the pvc protection controller determines volume still attached to node, and returns false. In this case, the queue.forget is [called](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/pvcprotection/pvc_protection_controller.go#L145). This means the next time the PVC will be processed is when either there is an event or we reach the resync period (which would trigger the pvcAddedUpdated) and then the key is added back to the queue for processing.

jingxu97:
> > > Looking at one of the test failures,
> > > pull-kubernetes-e2e-gce-storage-slow
> > > https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/94737/pull-kubernetes-e2e-gce-storage-slow/1305608296467533825/
> > > The test failed because the wait for pv deletion time is too less 30 secs. We should increase the timeout
> > > test deleted the pvc at 21:28:29, but pvc protection controller found the volume to be still in use on a node.
> > > ```
> > > I0914 21:28:28.9859129 pvc_protection_controller.go:379] Got event on PVC volume-provisioning-9528/pvc-6jc8g 
> > > I0914 21:28:28.9859249 pvc_protection_controller.go:168] Processing PVC volume-provisioning-9528/pvc-6jc8g
> > > I0914 21:28:28.9859339 pvc_protection_controller.go:464] For claim volume-provisioning-9528/pvc-6jc8g, checking for volume pvc-f36b6642-21d7-4f45-8a63-14f7f6747e0a in node informer cache
> > > I0914 21:28:28.9859479 pvc_protection_controller.go:494] Found volume kubernetes.io/gce-pd/e2e-b91b9ffbe0-390e9-d-pvc-f36b6642-21d7-4f45-8a63-14f7f6747e0a in use at node e2e-b91b9ffbe0-390e9-minion-group-f4zc
> > > I0914 21:28:28.9859539 pvc_protection_controller.go:193] Keeping PVC volume-provisioning-9528/pvc-6jc8g because it is still being used
> > > I0914 21:28:28.9859599 pvc_protection_controller.go:171] Finished processing PVC volume-provisioning-9528/pvc-6jc8g (29.35µs)
> > > ```
> > > 
> > > 
> > > Then at 21:30 pvc event was received and pvc finalizer was removed as expectecd.
> > > ```
> > > I0914 21:30:38.3537499 pvc_protection_controller.go:230] Removed protection finalizer from PVC volume-provisioning-9528/pvc-6jc8g
> > > I0914 21:30:38.3537729 pvc_protection_controller.go:171] Finished processing PVC volume-provisioning-9528/pvc-6jc8g (130.580021ms)
> > > ```
> > > 
> > > 
> > > We need to increase the wait for pv delete timeout in some of the tests
> > > e.g.
> > > https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/volume_provisioning.go#L418
> > > similar to here
> > > https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/volume_provisioning.go#L590
> > 
> > 
> > roughly how long it takes to unpublish and unstage finish?
> 
> The behavior that I see is sometimes it takes time for the PVC AddUpdate event to be triggered. worst case ~6 mins.
> 
> i.e say the pvc protection controller determines volume still attached to node, and returns false. In this case, the queue.forget is [called](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/pvcprotection/pvc_protection_controller.go#L145). This means the next time the PVC will be processed is when either there is an event or we reach the resync period (which would trigger the pvcAddedUpdated) and then the key is added back to the queue for processing.

what is the resync period?

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -418,3 +437,64 @@ func (c *Controller) enqueuePVCs(pod *v1.Pod, deleted bool) {
...
+	// isVolumeInUseByNode is called only when PVC is a delete candidate. In the
+	// scenario where a PVC is already marked for deletion and the volume name is
+	// not stamped on the PVC, skip the nodes attachment check.
+	if volName == "" {
+		klog.Infof("Failed to fetch PV name for the PVC %s/%s which is marked for deletion, skip node attached check for PV", pvc.Namespace, pvc.Name)
+		return false, nil
+	}
+
+	// Check node status in the informer cache, if error encountered or volume not found, do not return immediately, try the API server for consistency.
+	if nodeName, ok := pvc.Annotations[AnnSelectedNode]; ok {
```
msau42, line 454: This is not correct. This was the node at provisioning time, but a pod can be rescheduled to different nodes after that.

I think the only way to do this is to scan all node objects. We could potentially create a reverse index to speed up.

msau42 (commented):
> This means the next time the PVC will be processed is when either there is an event or we reach the resync period (which would trigger the pvcAddedUpdated) and then the key is added back to the queue for processing

Can this be resolved by having an informer event handler on Node status updates?

msau42:
Once we finalize on a design, can we update https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/postpone-pvc-deletion-if-used-in-a-pod.md

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
saikat-royc: Ack

saikat-royc:
> > This means the next time the PVC will be processed is when either there is an event or we reach the resync period (which would trigger the pvcAddedUpdated) and then the key is added back to the queue for processing
> 
> Can this be resolved by having an informer event handler on Node status updates?

Good point. This should speed up events when node.status is updated. Let me try out the change.

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
msau42: Another possibility is since we are still processing pods, we can get the nodes from Pod.NodeName.

Longer term though, if we want to replace the Pod check altogether with the Node check, then we would need to do some sort of node listing method. cc @gnufied

mattcary: Do we want to replace the Pod check? I would have thought the common case is the correct pod state is found in the cache.

Although I suppose the expensive part is just querying all the nodes, and so if we're going to do that anyway it would be simpler to have a single check.

jingxu97: from the current implementation, it query API server to get the list of pods, I think. And then loop through the each pod spec to match volume name. So I think this is more expensive than checking the nodes (which get it from informer cache) since # of nodes typically smaller # pods.

I am not sure why not use cache to get list of pods in previous implementation, though.

mattcary: I think it does use the cache: https://github.com/kubernetes/kubernetes/blob/c799b926fc3576849089dd85b3deb26867583690/pkg/controller/volume/pvcprotection/pvc_protection_controller.go#L245

Anyway you're right, there will be fewer nodes. And we have to check all pods any time it's not in use which is going to happen at all deletions.

So +1 to replacing the pod check I guess is my short version.

Thanks for all the help

msau42: In the original design, there were race conditions if we used the cache instead of live lookup. But I believe we solved it by doing a live lookup on kubelet, not the pvc protection controller: https://github.com/kubernetes/community/blame/master/contributors/design-proposals/storage/postpone-pvc-deletion-if-used-in-a-pod.md#L39

If we do eventually want to replace the pod check with a node check, we'll need to go through the exercise again of making sure there's no race conditions due to node + pod objects being out of sync.

jingxu97: ha, yes, I was looking at the live lookup here https://github.com/kubernetes/kubernetes/blob/7ceac2baf0820fd354259aa8b7f0e37b0bc6b814/pkg/controller/volume/pvcprotection/pvc_protection_controller.go#L277:22

I am wondering whether it is absolutely necessary. Not matter how hard we try, there might be a time that pod not yet updated when it checks, but after that it is updated.

msau42: The design relies on the scheduler to stop scheduling pods once the pvc has a deletion timestamp, and for kubelet to reject pods that use pvcs with a deletion timestamp

msau42: The scheduler only uses cache though, so it can slip through. That is why kubelet does a live lookup of the pvc.

msau42: Here's more background information on one of the possible races that we addressed in the original design: https://github.com/kubernetes/community/pull/1174#discussion_r144097411

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -418,3 +437,64 @@ func (c *Controller) enqueuePVCs(pod *v1.Pod, deleted bool) {
...
+	}
+
+	return c.isVolumeInUse(volName, nodesPtr), nil
+}
+
+// Helper function to determine if a given volume is present in the node's volumesInUse list.
+func (c *Controller) isVolumeInUse(volName string, nodes []*v1.Node) bool {
+	for _, node := range nodes {
+		for _, volInUse := range node.Status.VolumesInUse {
+			if strings.Contains(string(volInUse), volName) {
```
jingxu97, line 493: need to double check whether strings.contains always work..
do you know the volume format of VolumeInUse? I remember it is a unique key which contains some volume name.

saikat-royc: example entry in volumesInUse:
```
"kubernetes.io/csi/pd.csi.storage.gke.io^projects/mattcary-saikatroyc-test/zones/us-central1-b/disks/pvc-8198854a-e42e-4cb2-8d30-a911c3e7261e"
```
the part pvc-8198854a-e42e-4cb2-8d30-a911c3e7261e matches the pvc.Spec.VolumeName

jingxu97: To be safer, better to use filter the last part, and then use exact match.

saikat-royc: Can we assume uniform volume handle naming for all PVs? "/" separated? then we can split by '/' and compare against the last part?

jingxu97:
> > This means the next time the PVC will be processed is when either there is an event or we reach the resync period (which would trigger the pvcAddedUpdated) and then the key is added back to the queue for processing
> 
> Can this be resolved by having an informer event handler on Node status updates?

@saikat-royc, the volumesInUse string contains the pv name. e.g., kubernetes.io/csi/pd.csi.storage.gke.io^projects/jing-k8s-dev/zones/us-central1-b/disks/pvc-467d25bc-4143-4118-be52-476dc99c4c52.  By checking PV spec's ClaimName, you can get pvc namespace and name.

saikat-royc:
After some more discussions about the various scenarios, here is a summary of the proposed change:
Two issues to solve:
1. How can we prevent deleteVolume operation before nodeUnstageVolume operation finishes.
2. To achieve 1, we have a signal to determine if nodeUnstageVolume has not finished (node.status.VolumesInUse). How can we map the PV used by a PVC to an entry in the node.status.volumesInUse?

Current logic in Code:
In the PVC protection controller, the following logic currently determines PVC deletion.
PVC controller reacts to PVC add/update/periodic informer resync events and POD add/update/delete/periodic informer resync events.  Only when a PVC has a deletion timestamp, it is considered for deletion (ie, removing the finalizer). A PVC is considered safe to delete if it is not referenced by any Pod. The logic resides in the IsBeingUsed() function. When such a PVC is detected, the PVC protection controller removes the “kubernetes.io/pvc-protection" finalizer from the PVC.

Solution to issue 1
High level idea of the new change is “In addition to checking usage by a Pod, enhance the logic to determine if a PVC is being used, by considering whether the PV backed by the PVC is being used by any node”. Use node.Status.VolumesInUse as a signal to act as a barrier for the PVC protection controller to remove the finalizer on a PVC. Following values will be returned from the IsBeingUsed() function.

1. inUse: This value indicates if a PVC is in use by any pod or node, or if it is safe to delete the PVC.
2. needsRequeue: This value indicates that the controller loop should requeue the PVC in the controller, instead of waiting for a new event handler call on the PVC. This is an optimization to react faster to a change in state where no node references the PVC.

After checking the cached pod list, we check volumeInUse from cached nodes. Any reference to the PVC here means it is not safe to delete and we return without requeuing. Otherwise, first pods and then nodes are queried from the API server in case the caches are stale. In an optimization, in the case where a PVC is not referenced by a pod, but is marked in-use on a node, we will requeue the PVC. The reasoning is that such a volume is soon expected to be unstaged, as no pod is referencing the volume, and we shouldn’t wait for a new event on the PVC. The alternative to requeing would be to watch node events, and then queue the PVC. However we do not have any elegant/cheap way to determine the pvc from node.status.volumesInUse.

New IsBeingUsed function
```
IsBeingUsed(pvc) {
    podList = List pods from cache
    If (pvc in use by any pod in podList)
        Return inUse = true // it is expected a later pod event should put the pvc back in the queue for processing

    Bool volumeInUse = Check in node list from cache if pvc is being referenced by any node.status.VolumesInUse

    If volumeInUse == true:
        Return inUse = true, needsRequeue = true

    // volume not found in the node cache and pod cache
    podList = Lookup pod list from api server
    If pvc.spec.volumeName in use in podList:
        Return inUse = true, needsRequeue=false

    nodeList = lookup live node list from api server (we do this to cover scenarios where node cache does not capture volumesInUse but the live list does) // the same volume may be staged on multiple nodes, hence we list all nodes.
    If volume found in node nodeList:
        Return inUse = true, needsRequeue=true // we expect the unstage to happen soon here, which would remove the volume from the node.status.VolumesInUse. And since the pvc is marked for delete, we do not expect any new pod to reference it.

    Return InUse =  false
}
```
The logic above fixes the problem in the current problem when a pod is deleted before node unstage has completed. It still exposes the following race condition.
Pod created referring a volume V
A very slow kubelet starts working on the volume stage and mount, node.status.VolumesInUse not yet updated.
Pod deleted immediately
PVC controller sees that no pod referencing the PVC, and no node yet referencing the volume V, proceeds to remove the finalizer on the PVC.

This particular race hasn’t been observed in practice. Since it only happens on rapid pod churn (create followed by delete) we think it will be rare. It is not clear how to solve this race without some heavyweight change like adding a finalizer on pods with volumes that is removed after nodes have been appropriately updated.

Solution to Issue 2
How do we identify that a PVC.Spec.VolumeName is the same as reported in VolumesInUse?

For CSI driver node.Status.volumesInUse reports something like this
kubernetes.io/csi/<plugin>^<volumeHandle>
E.g. for PD
kubernetes.io/csi/pd.csi.storage.gke.io^projects/<project>/zones/<zone>/disks/pvc-c1e87907-1877-4b5d-9f32-fce3d059c5dc

For CSI plugins, 
Fetch PV spec for the pvc.Spec.VolumeName, and check if PV is provisioned by CSI plugin i.e pv.Spec.Csi != nil
If yes, check if pv.Spec.Csi.VolumeHandle is a substr in volumesInUse entry for a given node object.

Open item
For In-tree plugins, we need a way to uniformly map the pvc.Spec.VolumeName to the volumeHandle.

Plan of action:
1. In this PR, implement the fix which addresses most of the common scenarios for CSI plugins.
Next steps:
1. Investigate in-tree drivers volume nomenclature.
2. Attempt to fix the remaining race condition mentioned above with a robust design (if it’s worth it).

mattcary:
sgtm, thanks for the detailed description.

To reiterate, plan steps 2 & 3 are being done in future PRs.

saikat-royc:
The failing test are CSI mock volume tests, where the volume handles are like PV.Spec.CSI.VolumeHandle = "4", and this does not map to the node.status.VolumesInUse, which is causing issues. Needs investigation

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -219,25 +250,81 @@ func (c *Controller) removeFinalizer(pvc *v1.PersistentVolumeClaim) error {
...
+	}
+
+	// If PVC is marked for delete, before the PVC.Spec.Volume is populated,
+	// we should not block PVC deletion. By the time PVC is deleted, if PV
+	// creation was already triggered and the volume plugin is in the process
+	// of creation of the volume, then after the PV is created, the PV controller
+	// will trigger a delete for the PV (if it has a reclaim policy of delete),
+	// since the claim will not be found.
+	var pv *v1.PersistentVolume
+	if pvc.Spec.VolumeName != "" {
```
mattcary, line 278: Why do we not check this case before asking the pod informer? It might save a requeue.

pkg/controller/volume/pvcprotection/pvc_protection_controller_test.go
```diff
@@ -400,6 +470,160 @@ func testPVCProtectionController(t *testing.T, genericEphemeralVolumeFeatureEnab
...
+		{
+			name: "StorageObjectInUseProtection Enabled, initial node object with volume in use, initial pv object with csi source, deleted PVC with a volume name and finalizer, node informer empty, node queried from API server, volume found in the node.status.volumesInUse -> finalizer is not removed",
+			initialObjects: []runtime.Object{
+				withVolInUse(node(), testCSIVolumePrefix+defaultVolumeName),
+				pv(true /* with csi source */),
+			},
+			// node informer not yet updated with the node object.
+			informersAreLate: true,
+			updatedPVC:       deleted(withProtectionFinalizer(pvcWithVolumeName(pvc()))),
+			//updatedPV:        pv(true /* with csi source */),
```
mattcary, line 574: remove commented code

pkg/controller/volume/pvcprotection/pvc_protection_controller_test.go
```diff
@@ -400,6 +470,160 @@ func testPVCProtectionController(t *testing.T, genericEphemeralVolumeFeatureEnab
...
+		{
+			name: "StorageObjectInUseProtection Enabled, deleted PVC with a volume name, finalizer, node informer contains node, different volume found in the node.status.volumesInUse, API server node list invoked -> finalizer is removed",
+			initialObjects: []runtime.Object{
+				withVolInUse(node(), otherVolumeName),
+			},
+			// PVC does not contain selected-node annotation.
+			updatedPVC: deleted(withProtectionFinalizer(pvcWithVolumeName(pvc()))),
+			updatedPV:  pv(true /* with csi source */),
+			expectedActions: []clienttesting.Action{
+				//clienttesting.NewGetAction(pvGVR, "", defaultPVName),
```
mattcary, line 594: remove commented code

pkg/controller/volume/pvcprotection/pvc_protection_controller_test.go
saikat-royc: Done

pkg/controller/volume/pvcprotection/pvc_protection_controller_test.go
saikat-royc: Done

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
saikat-royc: Done

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -162,33 +192,34 @@ func (c *Controller) processPVC(pvcNamespace, pvcName string) error {
 	pvc, err := c.pvcLister.PersistentVolumeClaims(pvcNamespace).Get(pvcName)
 	if apierrors.IsNotFound(err) {
 		klog.V(4).Infof("PVC %s/%s not found, ignoring", pvcNamespace, pvcName)
-		return nil
+		return false, nil
 	}
 	if err != nil {
-		return err
+		return false, err
```
jingxu97, line 198: maybe return true, err since in this case pvc will be requeued?

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -219,25 +250,81 @@ func (c *Controller) removeFinalizer(pvc *v1.PersistentVolumeClaim) error {
...
+	// If PVC is marked for delete, before the PVC.Spec.Volume is populated,
+	// we should not block PVC deletion. By the time PVC is deleted, if PV
+	// creation was already triggered and the volume plugin is in the process
+	// of creation of the volume, then after the PV is created, the PV controller
+	// will trigger a delete for the PV (if it has a reclaim policy of delete),
+	// since the claim will not be found.
+	var pv *v1.PersistentVolume
+	if pvc.Spec.VolumeName != "" {
+		var err error
+		pv, err = c.getPV(pvc)
```
jingxu97, line 270: what if pv is no longer exist?

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -219,25 +250,81 @@ func (c *Controller) removeFinalizer(pvc *v1.PersistentVolumeClaim) error {
...
 	// correct decision to keep pvc is taken without doing an expensive live
-	// list.
-	if inUse, err := c.askInformer(pvc); err != nil {
+	// list. No requeue is needed as we will wait for a Pod event.
+	if inUse, err := c.askPodInformer(pvc); err != nil {
 		// No need to return because a live list will follow.
 		klog.Error(err)
 	} else if inUse {
-		return true, nil
+		return true, false, nil
```
jingxu97, line 283: maybe add a comment why requeue is false?

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
saikat-royc: if PV no longer exists, PV object would be nil, and volume in use by node check will be skipped

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -418,3 +505,63 @@ func (c *Controller) enqueuePVCs(pod *v1.Pod, deleted bool) {
...
+				klog.Infof("Found volume %s in use at node %s", volInUse, node.Name)
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func (c *Controller) generateUniqueCSIVolumeHandle(source *v1.CSIPersistentVolumeSource) (string, error) {
+	if source.Driver == "" || source.VolumeHandle == "" {
```
jingxu97, line 563: in what case driver name is empty or source volumeHandle is empty?

saikat-royc: [CSIPersistentVolumeSource](https://github.com/kubernetes/kubernetes/blob/75ff1f71381c961499989547e8438ddae6c0340f/staging/src/k8s.io/api/core/v1/types.go#L1687:6) mentions Driver and VolumeHandle field as required.
From this definition I handled this missing fields as an error case. @msau42  can these fields be empty?

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -418,3 +505,63 @@ func (c *Controller) enqueuePVCs(pod *v1.Pod, deleted bool) {
...
 	}
 }
+
+func (c *Controller) getPV(pvc *v1.PersistentVolumeClaim) (*v1.PersistentVolume, error) {
+	pv, err := c.pvLister.Get(pvc.Spec.VolumeName)
+	if err != nil {
+		klog.V(3).Infof("unexpected error getting persistent volume %q from cache for claim %s/%s: %v", pvc.Spec.VolumeName, pvc.Namespace, pvc.Name, err)
+		pv, err = c.client.CoreV1().PersistentVolumes().Get(context.TODO(), pvc.Spec.VolumeName, metav1.GetOptions{})
+	}
+	if pv == nil {
```
jingxu97, line 515: do we really need to handle pv==nil here? Directly return pv, err would have the same result?

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
jingxu97: if pv is nil, currently logic is err != nil, when err != nil, you will return false, false, err which means pvc finalizer will not be removed, right?

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -418,3 +505,63 @@ func (c *Controller) enqueuePVCs(pod *v1.Pod, deleted bool) {
...
+	}
+
+	return c.isVolumeInUse(volumeHandle, nodesPtr), nil
+}
+
+// Helper function to determine if a given volume is present in the node's volumesInUse list.
+func (c *Controller) isVolumeInUse(volHandle string, nodes []*v1.Node) bool {
+	for _, node := range nodes {
+		for _, volInUse := range node.Status.VolumesInUse {
+			volInUse = v1.UniqueVolumeName(strings.TrimPrefix(string(volInUse), volumeHandleCSIPrefix))
```
jingxu97, line 552: UniqueVolumeName is needed?

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -219,25 +250,81 @@ func (c *Controller) removeFinalizer(pvc *v1.PersistentVolumeClaim) error {
...
+	if pv != nil && pv.Spec.CSI != nil {
+		uniqueVolHandle, err := c.generateUniqueCSIVolumeHandle(pv.Spec.CSI)
+		if err != nil {
+			return false, false, err
+		}
+		if inUse, err := c.askNodeInformer(uniqueVolHandle, pvc); err != nil {
+			// No need to return because a live list will follow.
+			klog.Error(err)
+		} else if inUse {
+			return true, true, nil
```
jingxu97, line 298: add command why requeue is true

jingxu97:
Here we are adding one more check when removing finalizer for pvc. One concern I have is we have seen in some cases PVC takes long time/forever to delete due to finalizer. So we need to make sure this check will not cause some corner cases that makes PVC cannot be deleted.

Some possible scenario I see now 
- kubelet somehow cannot update node status. 
- driver cannot unstage volume forever (e.g., nfs server is down..)

Without this extra check, the current behavior for above cases is able to delete pvc, but with this fix, it will not. I am thinking we should handle these if possible. The rough idea is if pod is deleted, but volumeInUse by node, we should have a deadline, say after 5 minutes, remove finalizer of pvc.

mattcary:
> Without this extra check, the current behavior for above cases is able to delete pvc, but with this fix, it will not. I am thinking we should handle these if possible. The rough idea is if pod is deleted, but volumeInUse by node, we should have a deadline, say after 5 minutes, remove finalizer of pvc.

Note that the motivating issue for this bug has a deletevolume happening before the node unstage, which IIUC causes unmount on the node to be impossible and so stale devices on the node. This seems even harder to clean up than stale PVCs (at least those are visible in the KRM).

Maybe we could surface the error conditions better, but I'm not sure that if your nfs server is down and there are a bunch of stuck mounts the best thing to do is to let all associated PVCs go away as if everything is normal.

If the node is bad it will eventually die and the volume won't be in use so the finalizer will go away. 

For NFS at least, AFAICT if the pod is killed and no processes area accessing the mount, then it can be unmounted even if the nfs server is down.

So I'm not sure if we should have a deadline on the finalizer removal. I think (but am not sure) that I'd prefer just continuing to try to unstage.

jingxu97:
> > Without this extra check, the current behavior for above cases is able to delete pvc, but with this fix, it will not. I am thinking we should handle these if possible. The rough idea is if pod is deleted, but volumeInUse by node, we should have a deadline, say after 5 minutes, remove finalizer of pvc.
> 
> Note that the motivating issue for this bug has a deletevolume happening before the node unstage, which IIUC causes unmount on the node to be impossible and so stale devices on the node. This seems even harder to clean up than stale PVCs (at least those are visible in the KRM).
> 
> Maybe we could surface the error conditions better, but I'm not sure that if your nfs server is down and there are a bunch of stuck mounts the best thing to do is to let all associated PVCs go away as if everything is normal.
> 
> If the node is bad it will eventually die and the volume won't be in use so the finalizer will go away.
Right, but a little concern about some user might depend on certain existing behavior, e.g, expect pvc can be deleted even node has issues.
> 
> For NFS at least, AFAICT if the pod is killed and no processes area accessing the mount, then it can be unmounted even if the nfs server is down.

I have seen this mentioned before https://github.com/kubernetes/kubernetes/issues/66387#issuecomment-417689637

> 
> So I'm not sure if we should have a deadline on the finalizer removal. I think (but am not sure) that I'd prefer just continuing to try to unstage.

yeah, there is a tradeoff. Similar to detach behavior, we implemented a timeout, so after 6 mins, even if node status is still showing volumeInUse, attach_detach_controller will still go ahead to detach. This avoids certain cases that volume could be detached sooner instead of waiting forever. It is definitely not perfect, just in this split brain situation, we don't have much choices. 
By quickly checking the current issues, there is one related to pv deletion https://github.com/kubernetes/kubernetes/issues/77258, there are lots of them related to pod stuck terminating due to finalize. So my thought is trying to see how we could improve this situation, and avoid making it happen more.

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
msau42: I cannot think of a case where they would be empty.

saikat-royc:
> > Without this extra check, the current behavior for above cases is able to delete pvc, but with this fix, it will not. I am thinking we should handle these if possible. The rough idea is if pod is deleted, but volumeInUse by node, we should have a deadline, say after 5 minutes, remove finalizer of pvc.
> 
> Note that the motivating issue for this bug has a deletevolume happening before the node unstage, which IIUC causes unmount on the node to be impossible and so stale devices on the node. This seems even harder to clean up than stale PVCs (at least those are visible in the KRM).
> 
> Maybe we could surface the error conditions better, but I'm not sure that if your nfs server is down and there are a bunch of stuck mounts the best thing to do is to let all associated PVCs go away as if everything is normal.
> 
> If the node is bad it will eventually die and the volume won't be in use so the finalizer will go away.
> 
> For NFS at least, AFAICT if the pod is killed and no processes area accessing the mount, then it can be unmounted even if the nfs server is down.
> 
> So I'm not sure if we should have a deadline on the finalizer removal. I think (but am not sure) that I'd prefer just continuing to try to unstage.

I am leaning more towards @mattcary's suggestion. This is slightly different from a detach, since we are deleting pvc and possibly underlying storage and data (retain policy delete), we should be extra cautious with having a timeout based auto-delete. If user wants such a behavior this should be something ideally captured in the PVC object as an input. Since technically the PV is not completely detached from a node, keeping the PVC around looks fine to me. We can take up this improvement in a separate patch. Let's discuss more on this offline in a chat to reach a consensus.

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
saikat-royc: so handling this as an error case looks fine to me, as we never expect to have such a source.

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
saikat-royc: My bad that is true. Fixing this and will handle a not found case, so that we skip the node check if pv is nil.

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
saikat-royc: Done

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
saikat-royc: not strictly needed. I was re-using the volInUse variable. Storing as a string now in a new variable.

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
saikat-royc: Done

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
saikat-royc: removed this. And added a check for notfound in the caller to skip node in use checks. Added a test case also for it.

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
saikat-royc: Yes we can retry here since its not a not found error.

msau42:
FYI when we initially were brainstorming synchornizing detach and delete, there were concerns brought up about the node unhealthy case:
https://github.com/kubernetes/kubernetes/issues/84226

cc @cofyc

msau42:
cc @bswartz

gnufied (commented):
This PR will prevent workflows that might have been in use from working if node is unresponsive or turned-off.   I thought - if we can use some hacks and allow PVC deletion to skip node check if graceperiod is set to 0 , but it will still not be strictly backward compatible.
say for example - `kubectl delete pvc --force=true --grace-peridod=0 <pvc_name>`

Alternately - we could treat nodes with shutdown taint as unique and permit deletion of PVCs from those nodes. but then - not every deployment uses those taints. so there is no clear way to detect - a unresponsive node afaict.

Have we considered doing reference counting in code that checks for mounted volumes before permitting pod deletion? If a volume is in use by just one pod and staged, we could prevent pod deletion until volume is unstaged. If multiple pods are using same volume, prevent pod deletion only for last pod.

msau42:
The workaround we did for the detach case was we allowed PVC deletion to proceed, but we added logic in the external provisioner to prevent DeleteVolume from being called. That was simple to do because there's already a finalizer on the volumeattachment object. Adding a list Node to pv controller may work. It's another special case where we bypass pvc protection though. cc @jsafrane

msau42:
> Have we considered doing reference counting in code that checks for mounted volumes before permitting pod deletion?

This is also a possibility but is there a way to make this not racy?

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -219,25 +252,92 @@ func (c *Controller) removeFinalizer(pvc *v1.PersistentVolumeClaim) error {
 	return nil
 }
 
-func (c *Controller) isBeingUsed(pvc *v1.PersistentVolumeClaim) (bool, error) {
+// This function checks if a PVC is in use and returns values inUse,
+// needsRequeue and error.
+// isUse: This value indicates if a PVC is in use by any pod or node
```
xing-yang, line 257: isUse -> inUse

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -219,25 +252,92 @@ func (c *Controller) removeFinalizer(pvc *v1.PersistentVolumeClaim) error {
...
 	// correct decision to keep pvc is taken without doing an expensive live
-	// list.
-	if inUse, err := c.askInformer(pvc); err != nil {
+	// list. No requeue is needed as we will wait for a Pod event.
+	if inUse, err := c.askPodInformer(pvc); err != nil {
 		// No need to return because a live list will follow.
 		klog.Error(err)
 	} else if inUse {
-		return true, nil
+		// If we reach here, this means, the the PVC is still in use by a Pod, hence no
```
xing-yang, line 286: nit: the the PVC -> the PVC

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -219,25 +252,92 @@ func (c *Controller) removeFinalizer(pvc *v1.PersistentVolumeClaim) error {
...
 
 	// Even if no Pod using pvc was found in the Informer's cache it doesn't
 	// mean such a Pod doesn't exist: it might just not be in the cache yet. To
 	// be 100% confident that it is safe to delete pvc make sure no Pod is using
 	// it among those returned by a live list.
-	return c.askAPIServer(pvc)
+	if inUse, err := c.askAPIServerForPod(pvc); err != nil {
+		return false, false, err
+	} else if inUse {
+		// If we reach here, this means, the the PVC is still in use by a Pod, hence no
```
xing-yang, line 317: nit: the the PVC -> the PVC

pkg/controller/volume/pvcprotection/pvc_protection_controller.go
```diff
@@ -219,25 +252,92 @@ func (c *Controller) removeFinalizer(pvc *v1.PersistentVolumeClaim) error {
...
+		// reque is needed. We will wait for an event to process the PVC.
+		return true, false, nil
+	}
+
+	if pv != nil && pv.Spec.CSI != nil {
+		uniqueVolHandle, err := c.generateUniqueCSIVolumeHandle(pv.Spec.CSI)
```
xing-yang, line 323: Can we save uniqueVolHandle from line 295 and re-use here?

saikat-royc:
In this PR, PVC protection controller will not remove the finalizer on a PVC (which is marked for delete) until the volume is detached from the node.
cons: A delay in pvc deletion, will prevent users from using the pvc with the same name to claim a new volume or slow down this process. E.g. when using statefulset, pods have stable storage identities. If associated volume has failures or a new volume is required, new pvc can only be created after the old volume is detached.

Alternate solution :  Move the logic to check node.status.volumesInUse to the external-provisioner sidecar.
High level changes:
1. kublet/volumemanager changes to report volumesInuse correctly for drivers which do not implement a controllerpublish. (https://github.com/kubernetes/kubernetes/pull/94737/files#diff-367a5b4b2c1d046a94b4eeb471202ef0c3b2a2af3b15fd4d061a17277dd364cf)
2. get the volume handle from pv.spec.CSI (https://github.com/kubernetes/kubernetes/pull/94737/files#diff-d07f416aac5b044888ff88865a44f2b5a5669336535ac3023c539b503be43a94R572)
3. reuse the nodeLister in csi provisioner sidecar (https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go#L205)
4. In CanDelete() function [https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller.go#L1065]
    1. check node.Status.VolumesInUse first from node informer cache, else from api server node list. If found return false
    2. Else, proceed to existing VA checks. (this is only for plugins which support controller publish)
pro:
1. we do not block pvc deletion.
cons:
  1.  nodes + volumes in use  checks on every delete call per pv
open question:
1. If the node is unavailable, do we need a force delete? A force delete may not work anyway, if the volume is really in use at a the node.

Known race which would still exist:
1. Pod created referring a volume V
2.A very slow kubelet starts working on the volume stage and mount, node.status.VolumesInUse not yet updated.
3. Pod deleted immediately
4. PVC controller sees that no pod referencing the PVC and proceeds to remove the finalizer on the PVC, sidecar finds no node yet reporting the volume V in use, and proceeds with deletion.

other suggestions by @gnufied, https://github.com/kubernetes/kubernetes/pull/94737#pullrequestreview-512288237 (e.g refcounting and prevent deletion of last pod until volume is unstaged, this would pvc protection controller would not proceed with removal of the finalizer)

saikat-royc:
Please check new PRs
1. https://github.com/kubernetes/kubernetes/pull/98890
2. https://github.com/kubernetes-csi/external-provisioner/pull/562

PR #94739, opened 2020-09-12
test: ensure WaitForCacheSync is called after starting sharedInformerFacotry

knight42:

**What type of PR is this?**

/kind flake

https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-bazel-test/1304635954161520642
```
=== RUN   TestExpectationsOnRecreate
    replica_set_test.go:1172: initial RS didn't result in new item in the queue: timed out waiting for the condition
--- FAIL: TestExpectationsOnRecreate (30.01s)
```

We need to call `WaitForCacheSync` after starting shared informer, otherwise we might miss events.

Part of #94528

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/controller/replicaset/replica_set_test.go
+++ b/pkg/controller/replicaset/replica_set_test.go
@@ -31,7 +31,7 @@ import (
 	"time"
 
 	apps "k8s.io/api/apps/v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -599,6 +599,7 @@ func TestWatchControllers(t *testing.T) {
 		BurstReplicas,
 	)
 	informers.Start(stopCh)
+	informers.WaitForCacheSync(stopCh)
 
 	var testRSSpec apps.ReplicaSet
 	received := make(chan string)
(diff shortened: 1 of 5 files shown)


pkg/controller/replicaset/replica_set_test.go
```diff
@@ -1151,6 +1151,7 @@ func TestExpectationsOnRecreate(t *testing.T) {
 		100,
 	)
 	f.Start(stopCh)
+	f.WaitForCacheSync(stopCh)
 	fakePodControl := controller.FakePodControl{}
```
liggitt, line 1155: thanks… this is the second time an issue has come up like this in a unit test. do you mind sweeping for other places we're setting up NewSharedInformerFactory in unit tests to see if there are other places we need to do this?

knight42: @liggitt I would love to and I wrote a small program https://gist.github.com/knight42/73c13a0166a5d0884545f78bafc64acc to find out all the cases, and we seem to be lucky:

```
$ ./detect pkg
pkg/controller/replicaset/replica_set_test.go TestWatchControllers
pkg/controller/replicaset/replica_set_test.go TestExpectationsOnRecreate
$ ./detect staging
```

Note that there might be false negatives, i.e. if sharedInformerFactory is not created or started in test function directly, the program would fail to detect.

liggitt: great, thanks! do you mind adding the waits to those tests as well in this PR?

PR #94751, opened 2020-09-13
Fix misusage of RLock in timeCache lru.Cache.Get()

auxten:
**What type of PR is this?**

/kind bug

Lru's Get() method also updates the least recent usage information. So, we should not use RLock() and even RWLock here.
An alternative fix is just using `github.com/hashicorp/golang-lru`, but that will introduce more diffs. If that is the better fix, I can also do that. 

Fixes a potential race condition

```release-note
Fixes a regression in 1.19 with a race condition in kubelet pod handling
```

```docs

```

### DIFF
--- a/pkg/kubelet/time_cache.go
+++ b/pkg/kubelet/time_cache.go
@@ -27,7 +27,7 @@ import (
 
 // timeCache stores a time keyed by uid
 type timeCache struct {
-	lock  sync.RWMutex
+	lock  sync.Mutex
 	cache *lru.Cache
 }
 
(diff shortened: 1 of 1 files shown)


dims:
/assign @liggitt 

Jordan, since this was added by you. please review

liggitt:
good catch, thanks! can you also open a cherry-pick to release-1.19?

/lgtm
/approve
/milestone v1.20
/priority important-soon

auxten:
> good catch, thanks! can you also open a cherry-pick to release-1.19?
> 
> /lgtm
> /approve
> /milestone v1.20
> /priority important-soon

Sure, see #94774  @liggitt

PR #94752, opened 2020-09-13
Refine MatchingCache by using fnv64a and replacing RWMutex with Mutex

auxten:
**What type of PR is this?**

/kind cleanup

MatchingCache uses fnv32a for key hash and convert the result to uint64. I think just use fnv64a is better.
Also, the RWMutex here is not necessary, cause it just uses Lock and Unlock.

Fixes #

```release-note
NONE
```

```docs

```


auxten:
/retest
it's a docker pull issue

Issue #94719, opened 2020-09-11
Field selector podIPs not supported but enabled


zhouya0:
**What happened**:
When doing something like:
```
kubectl get pods --field-selector status.podIPs=10.100.6.153,fd00::6
```
It will report error:
```
Error from server (BadRequest): Unable to find "/v1, Resource=pods" that match label selector "", field selector "status.podIPs==10.100.6.153,fd00::6": invalid selector: 'status.podIPs==10.100.6.153,fd00::6'; can't understand 'fd00::6'
```
Because the logic to handle the field selector will separate `,`:
```
func splitTerms(fieldSelector string) []string {
	if len(fieldSelector) == 0 {
		return nil
	}

	terms := make([]string, 0, 1)
	startIndex := 0
	inSlash := false
	for i, c := range fieldSelector {
		switch {
		case inSlash:
			inSlash = false
		case c == '\\':
			inSlash = true
		case c == ',':
			terms = append(terms, fieldSelector[startIndex:i])
			startIndex = i + 1
		}
	}

	terms = append(terms, fieldSelector[startIndex:])

	return terms
}
```
Which I think `--fied-selector` should not support arrays. But `podIPs` was enable by https://github.com/kubernetes/kubernetes/pull/83123.

**What you expected to happen**:
1. Tell user not supported like:
```
[root@master ~]# kubectl get pods --field-selector spec.priority=0
Error from server (BadRequest): Unable to find "/v1, Resource=pods" that match label selector "", field selector "spec.priority=0": field label not supported: spec.priority
```
2. Provide a way to use array in `field selector`

**How to reproduce it (as minimally and precisely as possible)**:
```
kubectl get pods --field-selector status.podIPs=10.100.6.153,fd00::6
```

**Anything else we need to know?**:

**Environment**: 
- Kubernetes version (use `kubectl version`): 1.19
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`): 
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

liggitt:
The value for this field contains a comma. Commas are significant in field selectors to separate terms.

That means to match this value you need to escape the comma:

```
kubectl get pods --field-selector 'status.podIPs=10.100.6.153\,fd00::6'
```

PR #94756, opened 2020-09-14
Support status.podIPs field selector

zhouya0:
**What type of PR is this?**
/kind feature
/kind bug

This actually comes from another [issue](https://github.com/kubernetes/kubernetes/issues/94719).
But when I'm tesing `kubectl get pods --field-selector 'status.podIPs'` in my cluster with pod:
```
  podIP: 172.29.174.178
  podIPs:
  - ip: 172.29.174.178
  - ip: fdff:ffff:ffff:ffff::24d0:aea9
``` 
I'll get :
```
[root@10-6-150-83 ~]# kubectl get pods --field-selector 'status.podIPs=172.29.174.178\,fdff:ffff:ffff:ffff::24d0:aea9' -n default
No resources found in default namespace.
```
Then I realise there's a `TODO` which block this field selector, so after this PR:
```
[root@10-6-150-83 ~]# kubectl get pods --field-selector 'status.podIPs=172.29.174.178\,fdff:ffff:ffff:ffff::24d0:aea9' -n default
NAME                        READY   STATUS    RESTARTS   AGE
nginx-6d6dfbb6ff-hs6zh   1/1     Running   1          28d
```

Fixes #

/assign @liggitt 

```release-note
Support status.podIPs field selector
```

### DIFF
--- a/pkg/registry/core/pod/strategy.go
+++ b/pkg/registry/core/pod/strategy.go
@@ -240,18 +240,22 @@ func ToSelectableFields(pod *api.Pod) fields.Set {
 	// amount of allocations needed to create the fields.Set. If you add any
 	// field here or the number of object-meta related fields changes, this should
 	// be adjusted.
-	podSpecificFieldsSet := make(fields.Set, 9)
+	podSpecificFieldsSet := make(fields.Set, 10)
 	podSpecificFieldsSet["spec.nodeName"] = pod.Spec.NodeName
 	podSpecificFieldsSet["spec.restartPolicy"] = string(pod.Spec.RestartPolicy)
 	podSpecificFieldsSet["spec.schedulerName"] = string(pod.Spec.SchedulerName)
 	podSpecificFieldsSet["spec.serviceAccountName"] = string(pod.Spec.ServiceAccountName)
 	podSpecificFieldsSet["status.phase"] = string(pod.Status.Phase)
-	// TODO: add podIPs as a downward API value(s) with proper format
 	podIP := ""
+	var podIPs []string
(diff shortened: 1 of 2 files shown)


aojea:
this is a dual-stack only feature,
should this be "feature gated" @liggitt ?

pkg/registry/core/pod/strategy.go
```diff
@@ -240,18 +240,22 @@ func ToSelectableFields(pod *api.Pod) fields.Set {
...
 	podIP := ""
+	var podIPs []string
 	if len(pod.Status.PodIPs) > 0 {
 		podIP = string(pod.Status.PodIPs[0].IP)
+		for _, v := range pod.Status.PodIPs {
+			podIPs = append(podIPs, v.IP)
+		}
 	}
 	podSpecificFieldsSet["status.podIP"] = podIP
+	podSpecificFieldsSet["status.podIPs"] = strings.Join(podIPs, ",")
```
liggitt, line 258: It's not clear to me that a field selector on this field is desired... the use case I'm aware of for a field selector on the singular field was to be able to find the pod with a given IP address. To use this new selector you would need to know all the IPs for a pod (in the right order)... is that information typically available to API clients that don't already know the specific pod instance?

zhouya0: Yes, that's why the first time in https://github.com/kubernetes/kubernetes/issues/94719 I suggest to disable `podIPs` field-selector.
Maybe we just can't support array in field-selector.

pacoxu:
@kubernetes/sig-network-pr-reviews

PR #94049, opened 2020-08-17
Read ssh username from env variable

harche:
Signed-off-by: Harshal Patil <harpatil@redhat.com>

**What type of PR is this?**

/kind failing-test

**What this PR does / why we need it**:
Right now `pull-kubernetes-node-crio1-18-e2e` fails because it uses Fedora CoreOS image to run node tests. Fedora CoreOS only allows `core` user to ssh into the system. 

Once this PR is merged, https://github.com/kubernetes/test-infra/pull/18536 can be merged so that we can pass the user `core` to the node tests. 

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

No
```release-note
None
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #94760, opened 2020-09-14
Add a check for crio service before starting node e2e tests

### LINKED ISSUE #94759: crio node e2e test fails
**Which jobs are failing**:
/test pull-kubernetes-node-crio1-18-e2e

pull-kubernetes-node-crio1-18-e2e

**Reason for failure**:
Just before node e2e tests are started, we verify that the runtime (containerd or docker) is properly running or not on the target system. This check doesn't include crio, so the targeted node is not reported ready if it's using crio as the runtime. 

```bash
I0909 10:32:31.445] 
I0909 10:32:31.446] >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
I0909 10:32:31.447] >                              START TEST                                >
I0909 10:32:31.447] >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
I0909 10:32:31.447] Start Test Suite on Host 
I0909 10:32:31.447] 
I0909 10:32:31.447] Failure Finished Test Suite on Host 
I0909 10:32:31.448] unable to create gce instance with running docker daemon for image fedora-coreos-32-20200824-3-0-gcp-x86-64.  instance tmp-node-e2e-7eb5b1c5-fedora-coreos-32-20200824-3-0-gcp-x86-64 not running docker/containerd daemon - Command failed: prow@104.199.120.168: Permission denied (publickey,gssapi-keyex,gssapi-with-mic).
I0909 10:32:31.448] 
I0909 10:32:31.448] <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
I0909 10:32:31.448] <                              FINISH TEST                               <
I0909 10:32:31.448] <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
I0909 10:32:31.448] 
```

harche:
**What type of PR is this?**
/kind bug

Fixes #94759 

`None`

```release-note
None
```

```docs
None
```

### DIFF
--- a/test/e2e_node/runner/remote/run_remote.go
+++ b/test/e2e_node/runner/remote/run_remote.go
@@ -668,14 +668,15 @@ func createInstance(imageConfig *internalGCEImage) (string, error) {
 
 		var output string
 		output, err = remote.SSH(name, "sh", "-c",
-			"'systemctl list-units  --type=service  --state=running | grep -e docker -e containerd'")
+			"'systemctl list-units  --type=service  --state=running | grep -e docker -e containerd -e crio'")
 		if err != nil {
-			err = fmt.Errorf("instance %s not running docker/containerd daemon - Command failed: %s", name, output)
+			err = fmt.Errorf("instance %s not running docker/containerd/crio daemon - Command failed: %s", name, output)
 			continue
 		}
 		if !strings.Contains(output, "docker.service") &&
-			!strings.Contains(output, "containerd.service") {
-			err = fmt.Errorf("instance %s not running docker/containerd daemon: %s", name, output)
+			!strings.Contains(output, "containerd.service") &&
+			!strings.Contains(output, "crio.service") {
+			err = fmt.Errorf("instance %s not running docker/containerd/crio daemon: %s", name, output)
 			continue
 		}
 		instanceRunning = true


test/e2e_node/remote/ssh.go
```diff
@@ -70,7 +70,8 @@ func GetHostnameOrIP(hostname string) string {
 		host = ip
 	}
 
-	if *sshUser == "" {
+	// Do no ignore the KUBE_SSH_USER if it's set explicitly
+	if os.Getenv("KUBE_SSH_USER") != "" {
```
oomichi, line 74: I think the specified sshUser should be used instead of the env variable if specifying `--ssh-user` option.

oomichi (commented):
/cc @oomichi

test/e2e_node/remote/ssh.go
harche: @oomichi Thanks for your reply. For the Fedora CoreOS host, the only user available on host is `core`.  We have seen failures like, 

```
prow@104.199.120.168: Permission denied (publickey,gssapi-keyex,gssapi-with-mic)
```
The `prow` user doesn't exist on the Fedora CoreOS. In fact, `KUBE_SSH_USER` was [introduced](https://github.com/kubernetes/kubernetes/pull/94049/files) to handle Fedora CoreOS hosts (which will be used extensively for CRI-O testing as well as in future RuntimeClass tests)

This PR doesn't prevent the existing flow, it's just that for Fedora CoreOS we want to make sure we use the username that we know works. 

cc @dims

oomichi: @harche Thanks for your response. I see your point.
I guess that comes from https://github.com/kubernetes/test-infra/blob/master/kubetest/e2e.go#L575 which gets the specified username on `--ssh-user` option from the env value `USER`.
I feel it is better to get the username from the env value `KUBE_SSH-USER` if existing. If not existing, get the username from the env value `USER` as current implementation.

I'd like to get help from @BenTheElder for the above idea.

dims:
@oomichi let's unblock the CRI-O folks as we know this works. We can log a follow up issue to tweak if needed. How does that sound?

oomichi:
> @oomichi let's unblock the CRI-O folks as we know this works. We can log a follow up issue to tweak if needed. How does that sound?

As you know, for example kubectl tries reading the specified kubeconfig at first. If not specified, the env value KUBECONFIG is read. Such behavior seems pretty straightforward and that is the same as the existing one.
And current PR changes the above behavior like makes the env value stronger than the specified option value.
I'd like to see opinions from other people.

test/e2e_node/remote/ssh.go
BenTheElder: I agree that if --ssh-user is set we should respect that.
It also seems like https://github.com/kubernetes/test-infra/blob/feccbf60454db394d955e00977a9e2829712f050/kubetest/e2e.go#L575 could handle KUBE_SSH_USER overridding USER

test/e2e_node/remote/ssh.go
```diff
@@ -70,7 +70,8 @@ func GetHostnameOrIP(hostname string) string {
...
 		*sshUser = os.Getenv("KUBE_SSH_USER")
```
neolit123, line 75: as discussed above the flag should take precedence over the environment variable.

please change the logic to something like:

```
var user string
if *sshUser != "" {
	user = *sshUser
	goto add_user
}
if os.Getenv("KUBE_SSH_USER") != "" {
	user = os.Getenv("KUBE_SSH_USER")
	goto add_user
}
return host
add_user:
	return fmt.Sprintf("%s@%s", user, host)
```

harche:
@oomichi @BenTheElder @dims based on the feedback, I have created https://github.com/kubernetes/test-infra/pull/19273 to set the ssh username. 

The only change in this PR now is just to add the check to see if the remote node is running `crio.service` or not. 

Thanks everyone for taking time out for reviewing this PR.

test/e2e_node/remote/ssh.go
harche: https://github.com/kubernetes/test-infra/pull/19273  making `kubetest/e2e.go` handle KUBE_SSH_USER overridding USER. 

cc @oomichi @BenTheElder

neolit123 (commented):
/lgtm

neolit123:
i've commented on the test-infra PR.
/priority important-longterm

oomichi:
Thanks for your work @harche 

/lgtm

harche:
> /approve
> /lgtm
> /milestone v1.20

Thanks @dims, it seem for some reason the `approve` label is not getting assigned.

dims:
@harche i don't have approval rights for this directory :)

harche:
> @harche i don't have approval rights for this directory :)

oh ok. @sjenning do you have the approval rights?

neolit123:
Ideally someone from #sig-node (k8s slack)should approve.

PR #94766, opened 2020-09-14
apiserver: fix healthz vs. livez vs. readyz log output

sttts:
In logging and health endpoint results we statically wrote healthz although the request went to `/livez` or `/readyz`.

/kind bug

```release-note
NONE
```


logicalhan (approved):
/lgtm

logicalhan:
The test failures look legit though, probably due to a string match somewhere on the old output.

roycaihw:
/lgtm
/hold

the [expectedResponses](https://github.com/kubernetes/kubernetes/blob/f5a54e3f5862994f8cf422a60f4973583e4f98ba/staging/src/k8s.io/apiserver/pkg/server/healthz/healthz_test.go#L104-L114) need to be fixed

sttts:
@roycaihw fixed

roycaihw:
The test still fails. It always expects "healthz" in responses even when hitting the /ready endpoint.

PR #94773, opened 2020-09-14
count of etcd object should be limited to the specified resource

tkashem:
**What type of PR is this?**

/kind bug

Currently `Count` includes keys from different resource(s): any key (belonging to a different resource) that has the specified key as a prefix will be included in the count.

Consider the following keys:
```
A: <storage-prefix>//foo.bar.io/machines
B: <storage-prefix>//foo.bar.io/machinesets
```
If we ask for the count of key A, the result will also include the keys from key B since key B shares the same prefix as key A.

Append a separator to mark the end of the key, this will exclude all other keys from a different resource that is a prefix of the specified key.

* I am wondering if this change has any ripple effects. Is there any code site that relies on the current behavior (prefix matching) of the `Count` function.

```release-note
Fix etcd_object_counts metric reported by kube-apiserver
```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
+++ b/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
@@ -452,6 +452,14 @@ func getNewItemFunc(listObj runtime.Object, v reflect.Value) func() runtime.Obje
 
 func (s *store) Count(key string) (int64, error) {
 	key = path.Join(s.pathPrefix, key)
+
+	// We need to make sure the key ended with "/" so that we only get children "directories".
+	// e.g. if we have key "/a", "/a/b", "/ab", getting keys with prefix "/a" will return all three,
+	// while with prefix "/a/" will return only "/a/b" which is the correct answer.
+	if !strings.HasSuffix(key, "/") {
+		key += "/"
+	}
+
 	startTime := time.Now()
 	getResp, err := s.client.KV.Get(context.Background(), key, clientv3.WithRange(clientv3.GetPrefixRangeEnd(key)), clientv3.WithCountOnly())
 	metrics.RecordEtcdRequestLatency("listWithCount", key, startTime)
--- a/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store_test.go
+++ b/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store_test.go
@@ -1818,6 +1818,12 @@ func testSetup(t *testing.T) (context.Context, *store, *integration.ClusterV3) {
 func testPropogateStore(ctx context.Context, t *testing.T, store *store, obj *example.Pod) (string, *example.Pod) {
 	// Setup store with a key and grab the output for returning.
 	key := "/testkey"
+	return key, testPropogateStoreWithKey(ctx, t, store, key, obj)
+}
+
+// testPropogateStoreWithKey helps propagate store with objects, the given object will be stored at the specified key.
+func testPropogateStoreWithKey(ctx context.Context, t *testing.T, store *store, key string, obj *example.Pod) *example.Pod {
+	// Setup store with the specified key and grab the output for returning.
 	v, err := conversion.EnforcePtr(obj)
 	if err != nil {
 		panic("unable to convert output object to pointer")
@@ -1830,7 +1836,7 @@ func testPropogateStore(ctx context.Context, t *testing.T, store *store, obj *ex
 	if err := store.Create(ctx, key, obj, setOutput, 0); err != nil {
 		t.Fatalf("Set failed: %v", err)
 	}
-	return key, setOutput
+	return setOutput
 }
 
 func TestPrefix(t *testing.T) {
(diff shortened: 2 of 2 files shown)


staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
```diff
@@ -452,6 +452,19 @@ func getNewItemFunc(listObj runtime.Object, v reflect.Value) func() runtime.Obje
...
+
+	// Consider the following keys:
+	// A: <storage-prefix>//foo.bar.io/machines
+	// B: <storage-prefix>//foo.bar.io/machinesets
+	// If we ask for the count of key A, the result will also include the count for key B
+	// since key B shares the same prefix as key A.
+	//
+	// Add a separator to mark the end of the key here, this will exclude all other
+	// keys from a different resource that is a prefix of the specified key.
+	if len(key) > 0 && key[len(key) - 1] != '/' {
```
wojtek-t, line 464: Great finding - those this deserves a test to ensure we won't regress in the future.

tkashem: Thanks @wojtek-t. I am working on a test. I wanted to make sure this is a good fix.

tkashem: @wojtek-t, just added a test, can you please take a look?

staging/src/k8s.io/apiserver/pkg/storage/etcd3/store_test.go
```diff
@@ -1818,6 +1818,13 @@ func testSetup(t *testing.T) (context.Context, *store, *integration.ClusterV3) {
 func testPropogateStore(ctx context.Context, t *testing.T, store *store, obj *example.Pod) (string, *example.Pod) {
 	// Setup store with a key and grab the output for returning.
 	key := "/testkey"
+	setOutput := testPropogateStoreWithKey(ctx, t, store, key, obj)
+	return key, setOutput
```
wojtek-t, line 1822: nit:
```
return key, testPropagateStoreWithKey(ctx, t, store, key, obj)
```

staging/src/k8s.io/apiserver/pkg/storage/etcd3/store_test.go
```diff
@@ -2075,3 +2082,41 @@ func TestConsistentList(t *testing.T) {
...
+
+	resourceA := "/foo.bar.io/abc"
+
+	// resourceA is intentionally a prefix of resourceB to ensure that the count
+	// for resourceA does not include any objects from resourceB.
+	resourceB := fmt.Sprintf("%sdef", resourceA)
+
+	resourceACountExpected := 5
+	for i := 1; i <= resourceACountExpected; i++ {
+		obj := &example.Pod{ObjectMeta: metav1.ObjectMeta{Name: "foo"}}
```
wojtek-t, line 2098: nit: can we ensure thse have unique names?

Name: fmt.Sprintf("foo-%d", i)

staging/src/k8s.io/apiserver/pkg/storage/etcd3/store_test.go
```diff
@@ -2075,3 +2082,41 @@ func TestConsistentList(t *testing.T) {
...
+
+		key := fmt.Sprintf("%s/%d", resourceA, i)
+		testPropogateStoreWithKey(ctx, t, store, key, obj)
+	}
+
+	resourceBCount := 4
+	for i := 1; i <= resourceBCount; i++ {
+		obj := &example.Pod{ObjectMeta: metav1.ObjectMeta{Name: "foo"}}
+
```
wojtek-t, line 2107: same here

wojtek-t (commented):
Just some nits - thanks for fixing it!

staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
```diff
@@ -452,6 +452,19 @@ func getNewItemFunc(listObj runtime.Object, v reflect.Value) func() runtime.Obje
...
+	// A: <storage-prefix>//foo.bar.io/machines
+	// B: <storage-prefix>//foo.bar.io/machinesets
+	// If we ask for the count of key A, the result will also include the count for key B
+	// since key B shares the same prefix as key A.
+	//
+	// Add a separator to mark the end of the key here, this will exclude all other
+	// keys from a different resource that is a prefix of the specified key.
+	if len(key) > 0 && key[len(key)-1] != '/' {
+		key = fmt.Sprintf("%s/", key)
+	}
```
liggitt, line 466: use the same code as List():

https://github.com/kubernetes/kubernetes/blob/f64f687fcfb22e5099fbcb990f3e6faa44446fff/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go#L548-L553

wojtek-t:
/lgtm
/approve

I added release-note - this should be cherrypicked back once it is merged at head.

tkashem:
@liggitt @wojtek-t feedback addressed, please take a look.

tkashem:
> I added release-note - this should be cherrypicked back once it is merged at head.

Hi @wojtek-t, I will cherry pick for 1.19, 1.8 and 1.17. That should be enough, right?

PR #94774, opened 2020-09-14
Fix misusage of RLock in timeCache lru.Cache.Get() for release-1.19

auxten:
**What type of PR is this?**

/kind bug

Lru's Get() method also updates the least recent usage information. So, we should not use RLock() and even RWLock here.
An alternative fix is just using `github.com/hashicorp/golang-lru`, but that will introduce more diffs. If that is the better fix, I can also do that. 

Fixes a potential race condition

```release-note
Fixes a regression in 1.19 with a race condition in kubelet pod handling
```

```docs
Cherry-pick #94751 to release-1.19  @liggitt 
```

### DIFF
--- a/pkg/kubelet/time_cache.go
+++ b/pkg/kubelet/time_cache.go
@@ -27,7 +27,7 @@ import (
 
 // timeCache stores a time keyed by uid
 type timeCache struct {
-	lock  sync.RWMutex
+	lock  sync.Mutex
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev:
ping @kubernetes/release-managers

cpanato (approved):
/lgtm

PR #89703, opened 2020-03-31
Rename scheduler/nodeinfo pkg to scheduler/types

ahg-g:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:
renames pkg/scheduler/nodeinfo to pkg/scheduler/types. This is in preparation to introduce a new type to the scheduler named PodInfo which will be placed in this new package.

**Which issue(s) this PR fixes**:
Part of  #89528

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

/assign @alculquicondor

PR #94775, opened 2020-09-14
Move scheduler interface to pkg/scheduler/framework

### LINKED ISSUE #91029: Split scheduler framework implementation into a separate pkg
Split the scheduler framework into an interface pkg and an implementation pkg. 

We can keep the interface files (interface.go, types.go, listers.go and cyclestate.go) under ```scheduler/framework/v1alpha1``` and move the implementation to ```scheduler/framework/runtime```. 

Ideally plugins should only depend on the interface pkg. 

/cc @Huang-Wei @alculquicondor 
/sig scheduling

farah:
**What type of PR is this?**
/kind cleanup
/sig scheduling

Towards #91029 

**Notes for reviewer**

Moved files `pkg/scheduler/framework/v1alpha1` to `pkg/scheduler/framework`. 
Deleted `v1alpha1` as there's only one version of the framework interface.


pkg/scheduler/framework/plugins/noderesources/balanced_allocation.go
```diff
@@ -24,8 +24,9 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
+
 	"k8s.io/kubernetes/pkg/features"
-	framework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
+	framework "k8s.io/kubernetes/pkg/scheduler/framework"
```
farah, line 29: I think I should have removed the framework variable.

farah:
There is a golint error failing:

`pkg/scheduler/framework/interface.go:495:6: type name will be used as framework.FrameworkHandle by other packages, and that stutters; consider calling this Handle`

with the message:

> Please review the above warnings. You can test via "golint" and commit the result.
> If the above warnings do not make sense, you can exempt this package from golint

https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/v1alpha1/interface.go#L494

should I change it or exempt it?

farah:
How can I fix this error? (verify-import-boss)

```
E0915 00:09:47.669989  185511 main.go:45] Error: Failed executing generator: some packages had errors:
errors in package "k8s.io/kubernetes/test/e2e/framework/providers/kubemark":
the following imports did not match any allowed prefix:
  k8s.io/kubernetes/pkg/scheduler/framework
```

ahg-g:
I don't think we can do this in one PR. We should move everything to `pkg/scheduler/framework`, but keep `pkg/scheduler/framework/v1alpha1/types.go` with aliases to the types in `pkg/scheduler/framework`, this will allow us to do the internal cleanup first and then migrate the external dependencies (kubelet and e2e tests) in separate PRs.

ahg-g:
see https://github.com/kubernetes/kubernetes/pull/89703/files as an example, specifically the file ` pkg/scheduler/nodeinfo/node_info.go`

farah:
@ahg-g thanks for the tip. Quick question, how can I achieve something similar to type aliasing with functions?

For example in `pkg/scheduler/internal/cache/cache.go` there are references to the function `[GetPodKey](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/v1alpha1/types.go#L701)`, which exists in `types.go` under `v1alpha1`.

I need this function inside the new [types.go](https://github.com/kubernetes/kubernetes/blob/c336279ad641af36de0464a5c09e94a696ed67bf/pkg/scheduler/framework/types.go) in `pkg/scheduler/framework`, not sure how to proceed with this. Do I just copy and paste the function?

Btw I will try follow up on feedback asap.

pkg/scheduler/framework/types.go
```diff
@@ -0,0 +1,52 @@
...
+
+type ImageStateSummary = schedulertypes.ImageStateSummary
+
+type TransientSchedulerInfo = schedulertypes.TransientSchedulerInfo
+
+type HostPortInfo = schedulertypes.HostPortInfo
+
+type ProtocolPort = schedulertypes.ProtocolPort
+
+type NewResource = schedulertypes.NewResource
```
farah, line 52: @ahg-g This is an error as `NewResource` is a function. How can I handle migration for functions?

pkg/scheduler/framework/interface.go
```diff
@@ -229,15 +231,15 @@ type Plugin interface {
 }
 
 // LessFunc is the function to sort pod info
-type LessFunc func(podInfo1, podInfo2 *QueuedPodInfo) bool
+type LessFunc func(podInfo1, podInfo2 *v1alpha1.QueuedPodInfo) bool
```
farah, line 235: this should be *framework

pkg/scheduler/framework/types.go
ahg-g: define a new function that invokes the new one

pkg/scheduler/framework/interface.go
ahg-g: anything in the scheduler pkg should point to the framework pkg

farah: fixed

pkg/scheduler/framework/types.go
farah: yes that was obvious, thx & fixed

ahg-g:
before uploading a new commit, run `go test ./pkg/kubelet/lifecycle/...` and `go test ./pkg/scheduler/...` to get an idea if the code properly builds.

farah:
@ahg-g I just saw your comment, thanks.

pkg/scheduler/framework/types.go
```diff
@@ -0,0 +1,54 @@
+/*
```
ahg-g, line 1: the suggestion was actually to do the opposite: to have this file in `scheduler/framework/v1alpha1` and the original one moved here (to `scheduler/framework`), here we only need the types that external components like kubelet needs.

pkg/scheduler/core/extender.go
```diff
@@ -29,8 +29,9 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	restclient "k8s.io/client-go/rest"
 	extenderv1 "k8s.io/kube-scheduler/extender/v1"
+
```
ahg-g, line 32: we usually only leave an empty line  between builtin imports and the k8s one, so this line is not needed.

ahg-g: all other files as well

pkg/scheduler/framework/v1alpha1/types.go
```diff
@@ -697,15 +696,6 @@ func (n *NodeInfo) FilterOutPods(pods []*v1.Pod) []*v1.Pod {
 	return filtered
 }
 
-// GetPodKey returns the string key of a pod.
```
ahg-g, line 700: lets revert this and avoid any other cleanup because this PR is massive as is.

pkg/scheduler/framework/interface.go
```diff
@@ -277,7 +278,7 @@ type PreFilterPlugin interface {
 // running the pod.
 type FilterPlugin interface {
 	Plugin
-	// Filter is called by the scheduling framework.
+	// Filter is called by the scheduling
```
adtac, line 281: recheck this please

pkg/scheduler/framework/v1alpha1/types.go
```diff
@@ -76,7 +75,7 @@ type PodInfo struct {
 	ParseError                 error
 }
 
-// AffinityTerm is a processed version of v1.PodAffinityTerm.
+// AffinityTerm is a processed version of v1.PodAffinityTerm.pkg/scheduler/internal/cache/cache_test.go
```
adtac, line 78: please recheck this

farah:
@ahg-g @adtac  apologies but I have to create a new PR  #95069 as I screwed this one up by accidentally deleting the remote branch and it seems I'm unable to reopen pr even after restoring the branch. I get the message 'branch was forrced-pushed or recreated'.

PR #94783, opened 2020-09-14
add e2e test for services with pods hostNetwork: true as backends

aojea:
we were missing tests with backend pods with hostNetwork: true
for services.
This just adds another dimension to current "services regular checks", using the service backend pods with hostNetwork: true

```release-note
NONE
```

### DIFF
--- a/test/e2e/framework/network/utils.go
+++ b/test/e2e/framework/network/utils.go
@@ -98,6 +98,14 @@ func NewNetworkingTestConfig(f *framework.Framework, hostNetwork, SCTPEnabled bo
 	return config
 }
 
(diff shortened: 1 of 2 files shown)


neolit123:
the change looks good to me, ideally someone from sig-network should be assigned for review.

aojea:
I messed this up somehow with a wrong rebase and added all the sig and labels, sorry
/hold
using hostNetwork true makes that the tests can not run in parallel because the allocated ports conflict and can not be scheduled

aojea:
let me close this, is spanning all sigs. 
I' ll reopen later
sorry for the noise
/close

PR #94785, opened 2020-09-14
Increasing acceptable timeout for EndpointSlice garbage collection

robscott:
**What type of PR is this?**
/kind cleanup
/kind flake

This should help minimize flakiness for these tests. A more reliable solution here would likely involve updating the EndpointSlice controller to also clean up EndpointSlices. That seems like it should not be necessary though.

This should help with https://github.com/kubernetes/kubernetes/issues/92776

I'd missed following up on this earlier and noticed this test was still flakier than expected. https://storage.googleapis.com/k8s-gubernator/triage/index.html?text=EndpointSlice%20resource%20not%20deleted%20after%20Service

```release-note
NONE
```

/sig network
/priority important-soon
/cc @hasheddan 
/assign @liggitt

### DIFF
--- a/test/e2e/network/endpointslice.go
+++ b/test/e2e/network/endpointslice.go
@@ -144,8 +144,11 @@ var _ = SIGDescribe("EndpointSlice", func() {
 			framework.Failf("Endpoints resource not deleted after Service %s/%s was deleted: %s", svc.Namespace, svc.Name, err)
 		}
 
-		// Expect EndpointSlice resource to be deleted when Service is.
-		if err := wait.PollImmediate(2*time.Second, wait.ForeverTestTimeout, func() (bool, error) {
+		// Expect EndpointSlice resource to be deleted when Service is. Wait for
+		// up to 90 seconds since garbage collector only polls every 30 seconds
+		// and may need to retry informer resync at some point during an e2e
+		// run.
+		if err := wait.PollImmediate(2*time.Second, 90*time.Second, func() (bool, error) {
 			endpointSliceList, err := cs.DiscoveryV1beta1().EndpointSlices(svc.Namespace).List(context.TODO(), metav1.ListOptions{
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
/lgtm

test/e2e/network/endpointslice.go
```diff
@@ -145,7 +145,7 @@ var _ = SIGDescribe("EndpointSlice", func() {
 		}
 
 		// Expect EndpointSlice resource to be deleted when Service is.
-		if err := wait.PollImmediate(2*time.Second, wait.ForeverTestTimeout, func() (bool, error) {
```
liggitt, line 148: a comment here would be good as to why this was selected

robscott: I don't have a great reason other than to deflake tests. I'm not sure what a reasonable amount of time to wait for garbage collection to complete is, does this feel like a reasonable timeout?

liggitt: GC discovers served resource types every 30 seconds, and if they have changed (which they do frequently in e2e suites creating/deleting CRDs) it pauses GC while it sets up informers for the new types. If one of the discovered resource types becomes unavailable, the informer will not sync, and GC will timeout waiting for it after 30 seconds and rediscover available resources. I would add 1 minute for allowed GC controller-pausing overhead on top of whatever timeout you would wait for the resource itself to be cleaned up.

liggitt: https://github.com/kubernetes/kubernetes/blob/f64f687fcfb22e5099fbcb990f3e6faa44446fff/test/e2e/apimachinery/garbage_collector.go#L102-L110

robscott: Thanks! That's very helpful - just updated the comment and increased the timeout to 90 seconds.

PR #94800, opened 2020-09-15
Revert "conntrack e2e test debug connections"

mborsz:
Reverts kubernetes/kubernetes#92643

After this PR has been merged, gce-master-scale-correctness started OOMing: https://screenshot.googleplex.com/773vvp4MTsfMxv6

I'm not 100% it's caused by this change, but it seems to be likely given that this PR adds dumping iptables-save for each node pair + submission time correlates with the time when OOMs has started.

```release-note
NONE
```

### DIFF
--- a/test/e2e/network/conntrack.go
+++ b/test/e2e/network/conntrack.go
@@ -120,7 +120,6 @@ var _ = SIGDescribe("Conntrack", func() {
 		// TODO(#91236): Remove once the test is debugged and fixed.
 		// dump conntrack table for debugging
 		defer dumpConntrack(cs)
-		defer dumpIptables(cs)
 
 		// Create a NodePort service
 		udpJig := e2eservice.NewTestJig(cs, ns, serviceName)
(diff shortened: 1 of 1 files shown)


test/e2e/network/conntrack.go
```diff
@@ -328,25 +297,3 @@ func dumpConntrack(cs clientset.Interface) {
 		}
 	}
 }
-
-func dumpIptables(cs clientset.Interface) {
```
wojtek-t, line 332: Can we revert only the iptables part?

This is problematic because we log it for all nodes (5k), but logging conntrack should be fine as we log that for two nodes...

@aojea - FYI

aojea: yeah sure, , sorry,

mborsz: @wojtek-t, I'm not sure if I understand your comment: I think both dumpIptables and dumpConntrack iterates over all nodes.

wojtek-t: Sorry - I think this has changed in the meantime, I think we were dumping only particular nodes in the past.

wojtek-t:
@aojea - I'm going to approve it to fix scalability tests; we should re-revert by dumping iptables/conntrack only for nodes that are interested for us.

/lgtm
/approve
/retest

aojea:
@danehans this is breaking scsalability tests, we need milestone here please

wojtek-t:
Added milestone.

PR #94803, opened 2020-09-15
Limit the max number of splitting.

a2ush:
**What type of PR is this?**
/kind feature

Currently, kubernetes api-server authenticates a request that has header that includes "bearer", "token", and " strings"(space + strings).
For example, we can get the information form the api-server when we do the bellow command in a pod with serviceAccount.
```
root@test:/# for i in {0..10000};do echo -n $i' ' >> 10k; done
root@test:/# KK=`cat 10k`
root@test:/# TOKEN=`cat /var/run/secrets/kubernetes.io/serviceaccount/token`
root@test:/# curl -H "Authorization: Bearer ${TOKEN} ${KK}" --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt https://kubernetes/api/v1/namespaces/default/pods
{
  "kind": "PodList",
  "apiVersion": "v1",
  "metadata": {
...
```
If there is the header including many spaces and strings, unnecessary slicing will occur, and processing time will be late.

I confirm that strings.Split() and strings.SplitN() are doing the same[1][2].
So I think it is better to limit the max number of splitting with strings.SplitN().

And also, as far as I read the RFC[3], I think the format "Authorization: Bearer token" is the expected to be specify to the header.
Even if there is a value(space + strings) after the token, this value is not used in any place.

In addition, if we change the condition from `len(parts) > 2` to `len(parts)!= 2`, we can prevent the api-server from doing a authentication process when a invalid value inputs after token.
In a typical web server (apache, nginx and so on), the maximum size of HTTP header values is about 8K.(That is, they have the limit of the header length.)

So, I think it is better to set the limit to header, and if there is the format except "Authorization: Bearer token" in the header, kubernetes api-server should not perform the authentication process.

I think this PR is close to the specification change.
If you have opinions about this change (especially changing the condition from `len(parts) > 2` to `len(parts)!= 2`), could you tell me that?

```release-note
None
```

[1] https://golang.org/src/strings/strings.go?s=7164:7198#L289
[2] https://golang.org/src/strings/strings.go?h=func+genSplit#236
[3] https://tools.ietf.org/html/rfc6750#section-2.1

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/authentication/request/bearertoken/bearertoken.go
+++ b/staging/src/k8s.io/apiserver/pkg/authentication/request/bearertoken/bearertoken.go
@@ -39,7 +39,7 @@ func (a *Authenticator) AuthenticateRequest(req *http.Request) (*authenticator.R
 	if auth == "" {
 		return nil, false, nil
 	}
-	parts := strings.Split(auth, " ")
+	parts := strings.SplitN(auth, " ", 3)
 	if len(parts) < 2 || strings.ToLower(parts[0]) != "bearer" {
 		return nil, false, nil
 	}
--- a/staging/src/k8s.io/apiserver/pkg/authentication/request/bearertoken/bearertoken_test.go
+++ b/staging/src/k8s.io/apiserver/pkg/authentication/request/bearertoken/bearertoken_test.go
@@ -42,6 +42,28 @@ func TestAuthenticateRequest(t *testing.T) {
 	}
 }
 
+func TestAuthenticateRequestIncludingValueAfterToken(t *testing.T) {
+	testCases := []struct {
+		Req *http.Request
+	}{
+		{Req: &http.Request{Header: http.Header{"Authorization": []string{"Bearer token a"}}}},
+		{Req: &http.Request{Header: http.Header{"Authorization": []string{"Bearer token a b c"}}}},
+		{Req: &http.Request{Header: http.Header{"Authorization": []string{"Bearer token   a"}}}},
+	}
+	for i, testCase := range testCases {
+		auth := New(authenticator.TokenFunc(func(ctx context.Context, token string) (*authenticator.Response, bool, error) {
+			if token != "token" {
+				t.Errorf("unexpected token: %s", token)
+			}
+			return &authenticator.Response{User: &user.DefaultInfo{Name: "user"}}, true, nil
+		}))
+		resp, ok, err := auth.AuthenticateRequest(testCase.Req)
+		if !ok || resp == nil || err != nil {
+			t.Errorf("%d: expected valid user", i)
+		}
+	}
+}
+
 func TestAuthenticateRequestTokenInvalid(t *testing.T) {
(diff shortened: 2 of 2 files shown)


staging/src/k8s.io/apiserver/pkg/authentication/request/bearertoken/bearertoken.go
```diff
@@ -39,8 +39,8 @@ func (a *Authenticator) AuthenticateRequest(req *http.Request) (*authenticator.R
 	if auth == "" {
 		return nil, false, nil
 	}
-	parts := strings.Split(auth, " ")
-	if len(parts) < 2 || strings.ToLower(parts[0]) != "bearer" {
+	parts := strings.SplitN(auth, " ", 3)
+	if len(parts) != 2 || strings.ToLower(parts[0]) != "bearer" {
```
liggitt, line 43: wouldn't this break a request that included a trailing space after the token?

liggitt:
/hold

Since this is a breaking change, the scope of the impact and the benefit needs to be made clear and justify the break.

staging/src/k8s.io/apiserver/pkg/authentication/request/bearertoken/bearertoken.go
azush26: Yes, that's right.
But is there a use case to specify any values after token?

liggitt: I expect the main cause would be accidental header construction. If this were new code, being strict would be fine. Since there are existing clients that would be impacted, this change introduces unnecessary risk.

azush26: Agreed.
So, would this be fine?
```diff
-     	parts := strings.Split(auth, " ")
+ 	parts := strings.SplitN(auth, " ", 3) 
```

I mean that I don't edit the bellow code and merely make Split() bit faster.
```
if len(parts) < 2 || strings.ToLower(parts[0]) != "bearer" {
```

liggitt: that would be fine

azush26: Thank you!
I'll change the code.

azush26: I modified the code and PR title.

Thank you for confirming!

liggitt:
/hold cancel
/ok-to-test

go ahead and squash to a single commit and add a unit test showing this behaves the same if there are trailing spaces

azush26:
> go ahead and squash to a single commit and add a unit test showing this behaves the same if there are trailing spaces

I squashed the previous commit to a single one, and added a unit test.
Could you confirm that?

PR #94812, opened 2020-09-15
e2e test for services using pods with hostNetwork as backend

aojea:
we are missing tests that check the connectivity against services
that have backend pods with hostNetwork: true.

Because the tests run in parallel, it is possible that the pods used as
backends try to bind to the same port, and since all of them use the
host network, the scheduler will fail to create them due to port conflicts,
so we run them serially.

We have to skip networking tests with udp and endpoints using
hostNetwork, because they have a known issue.

```release-note
NONE
```

### DIFF
--- a/test/e2e/framework/network/utils.go
+++ b/test/e2e/framework/network/utils.go
@@ -53,8 +53,9 @@ const (
 	// EndpointUDPPort is an endpoint UDP port for testing.
 	EndpointUDPPort = 8081
 	// EndpointSCTPPort is an endpoint SCTP port for testing.
-	EndpointSCTPPort      = 8082
-	testContainerHTTPPort = 8080
+	EndpointSCTPPort = 8082
+	// testContainerHTTPPort is the test container http port.
+	testContainerHTTPPort = 9080
 	// ClusterHTTPPort is a cluster HTTP port for testing.
 	ClusterHTTPPort = 80
 	// ClusterUDPPort is a cluster UDP port for testing.
@@ -573,8 +574,7 @@ func (config *NetworkingTestConfig) createTestPodSpec() *v1.Pod {
 					ImagePullPolicy: v1.PullIfNotPresent,
 					Args: []string{
 						"netexec",
-						fmt.Sprintf("--http-port=%d", EndpointHTTPPort),
-						fmt.Sprintf("--udp-port=%d", EndpointUDPPort),
+						fmt.Sprintf("--http-port=%d", testContainerHTTPPort),
 					},
 					Ports: []v1.ContainerPort{
 						{
@@ -585,10 +585,6 @@ func (config *NetworkingTestConfig) createTestPodSpec() *v1.Pod {
 				},
 			},
 		},
-	}
-	// we want sctp to be optional as it will load the sctp kernel module
-	if config.SCTPEnabled {
-		pod.Spec.Containers[0].Args = append(pod.Spec.Containers[0].Args, fmt.Sprintf("--sctp-port=%d", EndpointSCTPPort))
 	}
 	return pod
 }
--- a/test/e2e/framework/network/utils.go
+++ b/test/e2e/framework/network/utils.go
@@ -90,9 +90,34 @@ const (
 // NetexecImageName is the image name for agnhost.
 var NetexecImageName = imageutils.GetE2EImage(imageutils.Agnhost)
 
+// Option is used to configure the NetworkingTest object
+type Option func(*NetworkingTestConfig)
+
+// EnableSCTP listen on SCTP ports on the endpoints
+func EnableSCTP(config *NetworkingTestConfig) {
+	config.SCTPEnabled = true
+}
+
+// UseHostNetwork run the test container with HostNetwork=true.
+func UseHostNetwork(config *NetworkingTestConfig) {
+	config.HostNetwork = true
+}
+
+// EndpointsUseHostNetwork run the endpoints pods with HostNetwork=true.
+func EndpointsUseHostNetwork(config *NetworkingTestConfig) {
+	config.EndpointsHostNetwork = true
+}
+
 // NewNetworkingTestConfig creates and sets up a new test config helper.
-func NewNetworkingTestConfig(f *framework.Framework, hostNetwork, SCTPEnabled bool) *NetworkingTestConfig {
-	config := &NetworkingTestConfig{f: f, Namespace: f.Namespace.Name, HostNetwork: hostNetwork, SCTPEnabled: SCTPEnabled}
+func NewNetworkingTestConfig(f *framework.Framework, setters ...Option) *NetworkingTestConfig {
+	// default options
+	config := &NetworkingTestConfig{
+		f:         f,
+		Namespace: f.Namespace.Name,
+	}
+	for _, setter := range setters {
+		setter(config)
+	}
 	ginkgo.By(fmt.Sprintf("Performing setup for networking test in namespace %v", config.Namespace))
 	config.setup(getServiceSelector())
 	return config
@@ -100,7 +125,12 @@ func NewNetworkingTestConfig(f *framework.Framework, hostNetwork, SCTPEnabled bo
 
 // NewCoreNetworkingTestConfig creates and sets up a new test config helper for Node E2E.
 func NewCoreNetworkingTestConfig(f *framework.Framework, hostNetwork bool) *NetworkingTestConfig {
-	config := &NetworkingTestConfig{f: f, Namespace: f.Namespace.Name, HostNetwork: hostNetwork}
+	// default options
+	config := &NetworkingTestConfig{
+		f:           f,
+		Namespace:   f.Namespace.Name,
+		HostNetwork: hostNetwork,
+	}
 	ginkgo.By(fmt.Sprintf("Performing setup for networking test in namespace %v", config.Namespace))
 	config.setupCore(getServiceSelector())
 	return config
@@ -125,6 +155,8 @@ type NetworkingTestConfig struct {
 	HostTestContainerPod *v1.Pod
 	// if the HostTestContainerPod is running with HostNetwork=true.
 	HostNetwork bool
+	// if the endpoints Pods are running with HostNetwork=true.
+	EndpointsHostNetwork bool
 	// if the test pods are listening on sctp port. We need this as sctp tests
 	// are marked as disruptive as they may load the sctp module.
 	SCTPEnabled bool
@@ -213,7 +245,11 @@ func (config *NetworkingTestConfig) diagnoseMissingEndpoints(foundEndpoints sets
 func (config *NetworkingTestConfig) EndpointHostnames() sets.String {
 	expectedEps := sets.NewString()
 	for _, p := range config.EndpointPods {
-		expectedEps.Insert(p.Name)
+		if config.EndpointsHostNetwork {
+			expectedEps.Insert(p.Spec.NodeSelector["kubernetes.io/hostname"])
+		} else {
+			expectedEps.Insert(p.Name)
+		}
 	}
 	return expectedEps
 }
@@ -745,6 +781,7 @@ func (config *NetworkingTestConfig) createNetProxyPods(podName string, selector 
 		hostname, _ := n.Labels["kubernetes.io/hostname"]
 		pod := config.createNetShellPodSpec(podName, hostname)
 		pod.ObjectMeta.Labels = selector
+		pod.Spec.HostNetwork = config.EndpointsHostNetwork
 		createdPod := config.createPod(pod)
 		createdPods = append(createdPods, createdPod)
 	}
--- a/test/e2e/network/networking.go
+++ b/test/e2e/network/networking.go
@@ -434,6 +434,78 @@ var _ = SIGDescribe("Networking", func() {
 				framework.Failf("failed dialing endpoint, %v", err)
 			}
 		})
+
+		// skip because pods can not reach the endpoint in the same host if using UDP and hostNetwork
+		// xref: #95565
+		ginkgo.It("should function for pod-Service(hostNetwork): udp", func() {
+			e2eskipper.Skipf("skip because pods can not reach the endpoint in the same host if using UDP and hostNetwork #95565")
+			config := e2enetwork.NewNetworkingTestConfig(f, e2enetwork.EndpointsUseHostNetwork)
+			ginkgo.By(fmt.Sprintf("dialing(udp) %v --> %v:%v (config.clusterIP)", config.TestContainerPod.Name, config.ClusterIP, e2enetwork.ClusterUDPPort))
+			err := config.DialFromTestContainer("udp", config.ClusterIP, e2enetwork.ClusterUDPPort, config.MaxTries, 0, config.EndpointHostnames())
+			if err != nil {
+				framework.Failf("failed dialing endpoint, %v", err)
+			}
+
+			ginkgo.By(fmt.Sprintf("dialing(udp) %v --> %v:%v (nodeIP)", config.TestContainerPod.Name, config.NodeIP, config.NodeUDPPort))
+			err = config.DialFromTestContainer("udp", config.NodeIP, config.NodeUDPPort, config.MaxTries, 0, config.EndpointHostnames())
+			if err != nil {
+				framework.Failf("failed dialing endpoint, %v", err)
+			}
+		})
+
+		// if the endpoints pods use hostNetwork, several tests can't run in parallel
+		// because the pods will try to acquire the same port in the host.
+		// We run the test in serial, to avoid port conflicts.
+		ginkgo.It("should function for service endpoints using hostNetwork", func() {
+			config := e2enetwork.NewNetworkingTestConfig(f, e2enetwork.UseHostNetwork, e2enetwork.EndpointsUseHostNetwork)
+
+			ginkgo.By("pod-Service(hostNetwork): http")
+
+			ginkgo.By(fmt.Sprintf("dialing(http) %v --> %v:%v (config.clusterIP)", config.TestContainerPod.Name, config.ClusterIP, e2enetwork.ClusterHTTPPort))
+			err := config.DialFromTestContainer("http", config.ClusterIP, e2enetwork.ClusterHTTPPort, config.MaxTries, 0, config.EndpointHostnames())
+			if err != nil {
+				framework.Failf("failed dialing endpoint, %v", err)
+			}
+			ginkgo.By(fmt.Sprintf("dialing(http) %v --> %v:%v (nodeIP)", config.TestContainerPod.Name, config.NodeIP, config.NodeHTTPPort))
+			err = config.DialFromTestContainer("http", config.NodeIP, config.NodeHTTPPort, config.MaxTries, 0, config.EndpointHostnames())
+			if err != nil {
+				framework.Failf("failed dialing endpoint, %v", err)
+			}
+
+			ginkgo.By("node-Service(hostNetwork): http")
+
+			ginkgo.By(fmt.Sprintf("dialing(http) %v (node) --> %v:%v (config.clusterIP)", config.NodeIP, config.ClusterIP, e2enetwork.ClusterHTTPPort))
+			config.DialFromNode("http", config.ClusterIP, e2enetwork.ClusterHTTPPort, config.MaxTries, 0, config.EndpointHostnames())
+			ginkgo.By(fmt.Sprintf("dialing(http) %v (node) --> %v:%v (nodeIP)", config.NodeIP, config.NodeIP, config.NodeHTTPPort))
+			config.DialFromNode("http", config.NodeIP, config.NodeHTTPPort, config.MaxTries, 0, config.EndpointHostnames())
+
+			ginkgo.By("node-Service(hostNetwork): udp")
+
+			ginkgo.By(fmt.Sprintf("dialing(udp) %v (node) --> %v:%v (config.clusterIP)", config.NodeIP, config.ClusterIP, e2enetwork.ClusterUDPPort))
+			config.DialFromNode("udp", config.ClusterIP, e2enetwork.ClusterUDPPort, config.MaxTries, 0, config.EndpointHostnames())
+			ginkgo.By(fmt.Sprintf("dialing(udp) %v (node) --> %v:%v (nodeIP)", config.NodeIP, config.NodeIP, config.NodeUDPPort))
+			config.DialFromNode("udp", config.NodeIP, config.NodeUDPPort, config.MaxTries, 0, config.EndpointHostnames())
+
+			ginkgo.By("handle large requests: http(hostNetwork)")
+
+			ginkgo.By(fmt.Sprintf("dialing(http) %v --> %v:%v (config.clusterIP)", config.TestContainerPod.Name, config.ClusterIP, e2enetwork.ClusterHTTPPort))
+			message := strings.Repeat("42", 1000)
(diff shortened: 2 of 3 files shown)


aojea:
/assign @danwinship 
let me know if you think this is the expected behavior, I 'm still not able to understand why the UDP services with hostNetwork pods are not able to reach endpoints in the same node, but HTTP services does

aojea:
> I 'm still not able to understand why the UDP services with hostNetwork pods are not able to reach endpoints in the same node, but HTTP services does

I think that udp always hit the same endpoint because of conntrack, until the conntrack entry times out the request always hit the same backend.

With udp also the hairpin traffic does not work

danwinship:
> * UDP ports show a different behavior, not allowing hairpin traffic, however, http seems to work
> 
> * a pod that is an endpoint of a service can not reach its service endpoints

These both seem like bugs (in the test case, the network plugin, or kube-proxy). I would expect both of these to work.

aojea:
I had a bug in the test case, progressing slowly

situation now, in kind jobs only 2 test cases fail with the same symptom, endpoint in the same node that the test container does not respond

```
Kubernetes e2e suite: [sig-network] Networking Granular Checks: Services should function for pod-Service(hostNetwork): udp
Kubernetes e2e suite: [sig-network] Networking Granular Checks: Services should update endpoints: udp(hostNetwork) 

```

in gce 4 test cases fail, the 2 mentioned previously with the same symptoms, plus:

```

[sig-network] Networking Granular Checks: Services should function for endpoint-Service(hostNetwork): http
[sig-network] Networking Granular Checks: Services should function for endpoint-Service(hostNetwork): udp 
```

these can not reach any endpoint

andrewsykim:
> let me know if you think this is the expected behavior, I 'm still not able to understand why the UDP services with hostNetwork pods are not able to reach endpoints in the same node, but HTTP services does

@aojea I've recently actually ran into a similar problem recently. HTTP worked but it was slow and had long delays. Did you see similar symptoms?

aojea:
summarizing, the traffic that does not work in the test is udp.
The scenario is:
- one service with pods in each node with host network as endpoints.
- another pod that tries to access the service (not part of the service)
- the udp traffic to the endpoint in the same node does not work

I should verify manually if for tcp it happens the same for the traffic in the same node

aojea:
/retest
unrelated failures

danwinship:
> the e2e test container used for the "Networking Granular Checks: Services"
> tests only needs to listen in one port to perform do network checks.
> This port is unrelated to the other ports used in the test, so we may
> use a different number to avoid possible conflicts.

I don't get this commit; the function you're modifying is used by lots of different tests, not just the "Networking Granular Checks: Services" ones. And some of the "Granular Checks" ones need UDP or SCTP...

aojea:
> > the e2e test container used for the "Networking Granular Checks: Services"
> > tests only needs to listen in one port to perform do network checks.
> > This port is unrelated to the other ports used in the test, so we may
> > use a different number to avoid possible conflicts.
> 
> I don't get this commit; the function you're modifying is used by lots of different tests, not just the "Networking Granular Checks: Services" ones. And some of the "Granular Checks" ones need UDP or SCTP...

Because they are using normal pods, hence they can allocate the same port because they have different ip.

I've added 8 new test cases that spawn pods with host network, all of them share the same IP and can run in parallel. If they try to allocate the same port they can not be scheduled because they try to allocate the same port

danwinship:
But I mean, you changed it so the function only listens on http, not udp or sctp. But some of the test cases need udp or sctp

aojea:
> But I mean, you changed it so the function only listens on http, not udp or sctp. But some of the test cases need udp or sctp

sorry, I was reading the comment in the mobile and I completely took it wrong, I though you was referring to other commit.

This test container is only used to perform the checks, you do a curl to the port exposed and it does the corresponding check with the parameters passed in the url, the other ports are never used.

danwinship:
ah, right, can you clarify that better in the commit and/or comments?

aojea:
@jayunit100 this is the scenario I've commented we were not testing

aojea:
/retest
unrelated 

Kubernetes e2e suite: [sig-storage] Subpath Atomic writer volumes should support subpaths with downward pod [LinuxOnly] [Conformance] expand_less

aojea:
The tests from pod (not hostnetwork) -> UDP service -> backends (with hostnetwork) fail to reach the local node because, the packet returns directly from the host to the pod, so NAT is not reversed:

```
    [NEW] udp      17 30 src=10.244.1.3 dst=10.98.132.194 sport=52350 dport=80 [UNREPLIED] src=172.19.0.4 dst=10.244.1.3 sport=8080 dport=52350
    [NEW] udp      17 30 src=10.244.1.1 dst=10.244.1.3 sport=8080 dport=52350 [UNREPLIED] src=10.244.1.3 dst=10.244.1.1 sport=52350 dport=8080
```

looking at tcpdump
```
08:53:39.925602 IP 10.244.1.3.51737 > 10.98.132.194.80: UDP, length 9
08:53:39.925705 IP 10.244.1.1.8080 > 10.244.1.3.51737: UDP, length 12
```

if it goes to a backend in a different node it works:

```
   [NEW] udp      17 30 src=10.244.1.3 dst=10.98.132.194 sport=41562 dport=80 [UNREPLIED] src=172.19.0.3 dst=172.19.0.4 sport=8080 dport=41562
 [UPDATE] udp      17 30 src=10.244.1.3 dst=10.98.132.194 sport=41562 dport=80 src=172.19.0.3 dst=172.19.0.4 sport=8080 dport=41562
```

tcpdump for the same scenario:
```
08:54:16.917668 IP 10.244.1.3.43708 > 10.98.132.194.80: UDP, length 9
08:54:16.917699 IP 172.19.0.4.43708 > 172.19.0.3.8080: UDP, length 9
08:54:16.917816 IP 172.19.0.3.8080 > 172.19.0.4.43708: UDP, length 11
08:54:16.917828 IP 10.98.132.194.80 > 10.244.1.3.43708: UDP, length 11
```

with TCP in the same node it works

```
    [NEW] tcp      6 120 SYN_SENT src=10.244.1.3 dst=10.98.177.30 sport=37566 dport=8080 [UNREPLIED] src=172.19.0.4 dst=10.244.1.3 sport=8080 dport=37566
 [UPDATE] tcp      6 60 SYN_RECV src=10.244.1.3 dst=10.98.177.30 sport=37566 dport=8080 src=172.19.0.4 dst=10.244.1.3 sport=8080 dport=37566
 [UPDATE] tcp      6 86400 ESTABLISHED src=10.244.1.3 dst=10.98.177.30 sport=37566 dport=8080 src=172.19.0.4 dst=10.244.1.3 sport=8080 dport=37566 [ASSURED]
 [UPDATE] tcp      6 120 FIN_WAIT src=10.244.1.3 dst=10.98.177.30 sport=37566 dport=8080 src=172.19.0.4 dst=10.244.1.3 sport=8080 dport=37566 [ASSURED]
 [UPDATE] tcp      6 30 LAST_ACK src=10.244.1.3 dst=10.98.177.30 sport=37566 dport=8080 src=172.19.0.4 dst=10.244.1.3 sport=8080 dport=37566 [ASSURED]
 [UPDATE] tcp      6 120 TIME_WAIT src=10.244.1.3 dst=10.98.177.30 sport=37566 dport=8080 src=172.19.0.4 dst=10.244.1.3 sport=8080 dport=37566 [ASSURED]
```


it seems that the UDP packet to the same node is not being SNATed, @danwinship @dcbw is this a kube-proxy problem or some sort of conntrack limitation?

danwinship:
> it seems that the UDP packet to the same node is not being SNATed, @danwinship @dcbw is this a kube-proxy problem or some sort of conntrack limitation?

It should not need to be SNAT'ed, because the node-to-pod response is in the same network namespace as the DNAT-to-node-IP was, so the return packet should be recognized and un-DNAT'ed. (Service traffic only needs to be SNAT'ed if the response would otherwise bypass the namespace where the DNAT happened. eg, pod to Service IP then hairpin back to same pod needs an SNAT because otherwise the pod would receive a packet with its own IP as both source and destination, and so the reply would stay within the pod namespace rather than bouncing back up to get un-DNAT'ed.) So this seems like a conntrack problem, but maybe there's some complication in the UDP case?

I don't understand your tcpdumps because I don't know which IP ranges are pods/services/nodes.

aojea:
> I don't understand your tcpdumps because I don't know which IP ranges are pods/services/nodes.

upps :sweat_smile: 
podIP probing the service 10.244.1.3
UDP service 10.98.132.194:80 
endpoints 172.19.0.4:8080 and 172.19.0.3:8080
hostip where the pod is running 172.19.0.4 and 10.244.1.1

aojea:
@danwinship I think that has to be something related to conntrack, or just that is not possible for conntrack to know if the reply belongs to the original request :shrug:  

more details in this gist
https://gist.github.com/aojea/72de8754bc20e7f8f930d0d6cf31b82b

Anyway, I'd like to merge the PR so we cover these hostNetwork scenarios. 
I will comment the udp failing tests and open a bug, this seems a very odd scenario

aojea:
/retest
:thinking:

test/e2e/framework/network/utils.go
```diff
@@ -93,15 +93,46 @@ var NetexecImageName = imageutils.GetE2EImage(imageutils.Agnhost)
...
 }
 
+// NewHostNetworkingTestConfig creates and sets up a new test config helper using backend pods using host network.
+// we need to set an offset because the tests are run in parallel and, since the pods are using the hostNetwork,
+// they can not expose the same port or they will have a conflict and will not be scheduled
+func NewHostNetworkingTestConfig(f *framework.Framework, hostNetwork bool, portOffset int) *NetworkingTestConfig {
+	config := &NetworkingTestConfig{f: f,
+		Namespace:           f.Namespace.Name,
+		HostNetwork:         hostNetwork,
+		NetProxyHostNetwork: true,
```
danwinship, line 116: The old `HostNetwork`/`hostNetwork` is now confusing. And "proxy" to me sounds like it means "in the middle", but `NetProxyHostNetwork` isn't whether the "middle" pod is hostnetwork but rather whether the final pod is hostnetwork.

I'd call the old field `HostNetworkProxy` or `HostNetworkIntermediary` and the new one `HostNetworkDestination`.

And maybe it would be better to have just

    func NewNetworkingTestConfig(f *framework.Framework, options *NetworkingTestConfig) *NetworkingTestConfig

and the caller can pass in a partially-constructed test config with `HostNetworkProxy`, `HostNetworkDestination`, and `SCTPEnabled` set if they need those options?

And get rid of the `portOffset` arg; have `NewNetworkingTestConfig` set it itself when using `HostNetworkDestination`. Just have a global (and a mutex protecting it) and take the next available offset each time it's called.

test/e2e/network/networking.go
```diff
@@ -206,20 +206,21 @@ var _ = SIGDescribe("Networking", func() {
...
-			}
-
-			ginkgo.By(fmt.Sprintf("dialing(udp) %v --> %v:%v (nodeIP)", config.TestContainerPod.Name, config.NodeIP, config.NodeUDPPort))
-			err = config.DialFromTestContainer("udp", config.NodeIP, config.NodeUDPPort, config.MaxTries, 0, config.EndpointHostnames())
-			if err != nil {
-				framework.Failf("failed dialing endpoint, %v", err)
-			}
-		})
+		// Skip because of #95565
+		// ginkgo.It("should function for pod-Service(hostNetwork): udp", func() {
```
danwinship, line 210: rather than commenting it out you could just `e2eskipper.Skipf()` it...

aojea: done

test/e2e/framework/network/utils.go
aojea: > and the caller can pass in a partially-constructed test config with HostNetworkProxy, HostNetworkDestination, and SCTPEnabled set if they need those options?

Better using functional options since it seems the idiomatic way and is allows to add new options easily

> The old HostNetwork/hostNetwork is now confusing. And "proxy" to me sounds like it means "in the middle", but NetProxyHostNetwork isn't whether the "middle" pod is hostnetwork but rather whether the final pod is hostnetwork.

absolutely agree, very bad naming, 
`HostNetwork` remains the same, for endpoints with `HostNetwork` the new variable is `EndpointsHostNetwork` now, and they have independent setters

> And get rid of the portOffset arg; have NewNetworkingTestConfig set it itself when using HostNetworkDestination. Just have a global (and a mutex protecting it) and take the next available offset each time it's called.

yeah, the portOffset is wrong but I prefer to avoid to use a global because it will make the test port unpredictable and the randomness cause lot of troubles to debug flakes and failures, also, I'm not sure how ginkgo parallel will work with that approach (see the `Parallel Specs` section in https://onsi.github.io/ginkgo/)

I prefer to keep it simple,  I've added setters so we can configure the endpoints ports in each test, and is not that we expect more changes on this part in the future.

aojea:
@danwinship can you PTAL

test/e2e/framework/network/utils.go
```diff
@@ -94,24 +94,50 @@ var NetexecImageName = imageutils.GetE2EImage(imageutils.Agnhost)
...
 
+// Run the test container with HostNetwork=true.
 func UseHostNetwork() Option {
 	return func(config *NetworkingTestConfig) {
 		config.HostNetwork = true
 	}
 }
 
+// Set the endpoint HTTP port for the test
+func PodHTTPPort(port int) Option {
```
danwinship, line 119: you introduce these here but you don't actually use them yet...

test/e2e/network/networking.go
```diff
@@ -191,6 +191,36 @@ var _ = SIGDescribe("Networking", func() {
 			}
 		})
 
+		ginkgo.It("should function for pod-Service(hostNetwork): http", func() {
+			config := e2enetwork.NewNetworkingTestConfig(f, e2enetwork.EndpointsUseHostNetwork(), e2enetwork.PodHTTPPort(8090), e2enetwork.PodUDPPort(8091))
```
danwinship, line 195: I still feel like this is more likely to lead to conflicts/bugs in the future than just having `NewNetworkingTestConfig` pick ports itself. Like, if someone copies+pastes the test, they have to remember to change the ports, and also scan through all the other tests to figure out what the next available port is.

test/e2e/network/networking.go
```diff
@@ -206,7 +206,9 @@ var _ = SIGDescribe("Networking", func() {
 			}
 		})
 
+		// skip because of #95565
```
danwinship, line 209: maybe clarify a _little_ more?

    // This ought to work but for some reason it doesn't. http://issues.k8s.io/95565

test/e2e/framework/network/utils.go
```diff
@@ -91,45 +91,45 @@ const (
 // NetexecImageName is the image name for agnhost.
 var NetexecImageName = imageutils.GetE2EImage(imageutils.Agnhost)
 
-// Use functional options to configure the NetworkingTest object
+// Option is used to configure the NetworkingTest object
```
danwinship, line 94: squash the golint fixes into the correct commits?

test/e2e/framework/network/utils.go
```diff
@@ -91,9 +91,31 @@ const (
 // NetexecImageName is the image name for agnhost.
 var NetexecImageName = imageutils.GetE2EImage(imageutils.Agnhost)
 
+// Use functional options to configure the NetworkingTest object
+type Option func(*NetworkingTestConfig)
+
+func EnableSCTP() Option {
+	return func(config *NetworkingTestConfig) {
```
danwinship, line 99: So I know you have to use a function-returning-a-function for the port ones, but I don't think it necessarily makes sense here... (especially if we end up getting rid of the port-setting options)

aojea: We have to add another one soon `EnableDualStack()`

https://github.com/kubernetes/kubernetes/pull/95063/commits/a733598569da1974a43d4010cb92022ffc0193c3#diff-c26eb8ec2a51d898f9bbd94080b25dbef525d3ff1578bbcd0a6c5d10bc521c00R101

aojea: was meant to simplify the review process only, each commit is not independent

test/e2e/framework/network/utils.go
aojea: was meant to simplify the review process only, each commit is not independent

test/e2e/network/networking.go
aojea: Maybe I'm not understanding well your proposal, do you propose to have a pool of ports and then the tests allocate ports from it?
you need to size it correctly so you never have to change it, because adding a new test will imply to count all the number of tests using the pool, you can not oversubscribe. What size should we reserve, 100 ports? 8900 - 9000, 
Another problem I see is, if someone adds a new test that uses a port belonging to the pool range, i.e 8915, you can not grep in the code, and if you don't know that there are some tests in the code that allocate from a pool, you'll be having port conflict errors with different tests.

I still think constants are a better option, :smile:

test/e2e/network/networking.go
aojea: yep

test/e2e/framework/network/utils.go
danwinship: right, but I mean, for the ones that don't take arguments, you could just do

```
func EnableSCTP(config *NetworkingTestConfig) {
	config.SCTPEnabled = true
}
```

and then pass `e2enetwork.EnableSCTP` rather than `e2enetwork.EnableSCTP()`. I guess there are arguments either way.

test/e2e/network/networking.go
danwinship: > Another problem I see is, if someone adds a new test that uses a port belonging to the pool range, i.e 8915, you can not grep in the code

True. OK.

(Though really, the way the code is written, we could just let the apiserver allocate free ports itself and just read them back and set `config.NodeHTTPPort` etc from that...)

aojea: actually, I'm not happy with my solution too, let' s see if we can find a middle ground ...

test/e2e/framework/network/utils.go
aojea: yep

aojea:
/hold
we need to find a good way to do the port allocation for hostnetwork pods

aojea:
/hold cancel
@danwinship this will give us coverage without too much changes, we can revisit later if needed.

MonzElmasry:
Hello! Bug Triage here :wave: This is a friendly reminder that code freeze is starting 12.11.2020 (about a weeks and half from now) and while there is still plenty of time, we want to ensure that each PR has a chance to be merged on time. so i'd like to check the status of this one, will it make it to 1.20 release ?

aojea:
> Hello! Bug Triage here  This is a friendly reminder that code freeze is starting 12.11.2020 (about a weeks and half from now) and while there is still plenty of time, we want to ensure that each PR has a chance to be merged on time. so i'd like to check the status of this one, will it make it to 1.20 release ?

yeah, all comments are addressed, just need a final review and approval

aojea:
/assign @MrHohn 
for approval once @danwinship confirm his reviews were addressed

test/e2e/framework/network/utils.go
```diff
@@ -212,7 +245,11 @@ func (config *NetworkingTestConfig) diagnoseMissingEndpoints(foundEndpoints sets
 func (config *NetworkingTestConfig) EndpointHostnames() sets.String {
 	expectedEps := sets.NewString()
 	for _, p := range config.EndpointPods {
-		expectedEps.Insert(p.Name)
+		if config.EndpointsHostNetwork {
+			expectedEps.Insert(p.Spec.NodeSelector["kubernetes.io/hostname"])
```
andrewsykim, line 249: Should this be checking `p.Spec.NodeName` or `p.Labels` instead? Checking node selector to extract the hostname seems weird.

test/e2e/network/networking.go
```diff
@@ -426,20 +426,92 @@ var _ = SIGDescribe("Networking", func() {
...
 			err := config.DialEchoFromTestContainer("udp", config.ClusterIP, e2enetwork.ClusterUDPPort, config.MaxTries, 0, message)
 			if err != nil {
 				framework.Failf("failed dialing endpoint, %v", err)
 			}
 		})
+
+		// skip because pods can not reach the endpoint in the same host if using UDP and hostNetwork
+		// xref: #95565
+		ginkgo.It("should function for pod-Service(hostNetwork): udp", func() {
+			e2eskipper.Skipf("skip because pods can not reach the endpoint in the same host if using UDP and hostNetwork #95565")
```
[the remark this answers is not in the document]
aojea: I had a problem with the test, I' ve opened an issue with more details about the specific problemand it only happens with UDP

test/e2e/framework/network/utils.go
aojea: it is the weird it is configured in the manifest, you can see the code here 
https://github.com/kubernetes/kubernetes/blob/b61faa388c0d03fe2ae99b5b36e1d6f4d71eba29/test/e2e/framework/network/utils.go#L578-L580

so I took it directly, totally open to suggestions

aojea:
kindly reminder @andrewsykim @MrHohn , I need this merged to add another set of e2e tests for dual stack that depends on this refactor of the `NetworkingTest`

test/e2e/framework/network/utils.go
aojea: this is taking it from the configuration object, it does not read it from the API ,so it seems the only place to take it

test/e2e/framework/network/utils.go
```diff
@@ -89,17 +90,47 @@ const (
...
+
+// EndpointsUseHostNetwork run the endpoints pods with HostNetwork=true.
+func EndpointsUseHostNetwork(config *NetworkingTestConfig) {
+	config.EndpointsHostNetwork = true
+}
+
 // NewNetworkingTestConfig creates and sets up a new test config helper.
-func NewNetworkingTestConfig(f *framework.Framework, hostNetwork, SCTPEnabled bool) *NetworkingTestConfig {
-	config := &NetworkingTestConfig{f: f, Namespace: f.Namespace.Name, HostNetwork: hostNetwork, SCTPEnabled: SCTPEnabled}
+func NewNetworkingTestConfig(f *framework.Framework, setters ...Option) *NetworkingTestConfig {
```
MrHohn, line 112: Thanks, I like this generic setter much more than piling up bool vars.

MrHohn (approved):
Approving this to unblock the up-coming e2e tests. The open comments could be addresses in a follow up if needed :) 
/approve

test/e2e/framework/network/utils.go
```diff
@@ -53,8 +53,9 @@ const (
 	// EndpointUDPPort is an endpoint UDP port for testing.
 	EndpointUDPPort = 8081
 	// EndpointSCTPPort is an endpoint SCTP port for testing.
-	EndpointSCTPPort      = 8082
-	testContainerHTTPPort = 8080
+	EndpointSCTPPort = 8082
+	// testContainerHTTPPort is the test container http port.
+	testContainerHTTPPort = 9080
```
andrewsykim, line 58: this changed from 8080 to 9080, expected?

aojea: this container is wrongly using the same ports that the other containers that are endpoints from the service under test.
It is a different container that expose a handler to execute commands, not to be polled

test/e2e/framework/network/utils.go
aojea: yeah, and I have to add some dual stack tests that need to add more variables, that will make it even worse :upside_down_face:

test/e2e/framework/network/utils.go
aojea: IIRC I added it in a separate commit and the description should add a bit more of detail

PR #94813, opened 2020-09-15
Don't start goroutine for noMetrics

### LINKED ISSUE #94811: updateUnfinishedWorkLoop uses a large amount of CPU at idle
**What happened**:
When we scale Rancher to 1000's of clusters we end up spending a large percentage of
our time waking up and doing nothing across 10s of thousands of goroutines. This was seen as constantly consuming 2-3 cores while the system was in an 'idle' state with approx. 78k queues being active.

I would expect these goroutines to not be running if there is no metrics for them to write to.

Calling [New](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/util/workqueue/queue.go#L36) in a loop 

**Environment**:
- Kubernetes version (use `kubectl version`): v1.19.1
- Cloud provider or hardware configuration: DO droplets
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

dramich:
**What type of PR is this?**
/kind bug

Problem:
When calling newQueue metrics can be of type noMetrics when just calling
New. When doing this a new goroutine is created to update the metrics
but in this case there are no metrics so it's just creating goroutines
that don't do anything but consume resources.

Solution:
If the incoming metrics is of type noMetrics, don't start the goroutine

Fixes #94811

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/client-go/util/workqueue/queue.go
+++ b/staging/src/k8s.io/client-go/util/workqueue/queue.go
@@ -55,7 +55,13 @@ func newQueue(c clock.Clock, metrics queueMetrics, updatePeriod time.Duration) *
 		metrics:                    metrics,
 		unfinishedWorkUpdatePeriod: updatePeriod,
 	}
-	go t.updateUnfinishedWorkLoop()
+
+	// Don't start the goroutine for a type of noMetrics so we don't consume
+	// resources unnecessarily
+	if _, ok := metrics.(noMetrics); !ok {
+		go t.updateUnfinishedWorkLoop()
+	}
+
 	return t
 }
 


dramich:
Check my CLA please

roycaihw:
nit: perhaps add a comment to the code about why we need it as well

/lgtm
/retest

dramich:
@roycaihw Thanks for reviewing, I added a comment per your suggestion.

dramich:
@roycaihw Are you able to take another look at this? Thanks.

staging/src/k8s.io/client-go/util/workqueue/queue.go
```diff
@@ -55,7 +55,13 @@ func newQueue(c clock.Clock, metrics queueMetrics, updatePeriod time.Duration) *
 		metrics:                    metrics,
 		unfinishedWorkUpdatePeriod: updatePeriod,
 	}
-	go t.updateUnfinishedWorkLoop()
+
+	// Don't start the goroutine for a type of noMetrics so we don't consume
```
roycaihw, line 59: I think the gofmt test is complaining about a trailing space in this line: 
```
-	// Don't start the goroutine for a type of noMetrics so we don't consume 
+	// Don't start the goroutine for a type of noMetrics so we don't consume
```

dramich: Removed it and repushed.

roycaihw:
/lgtm
Thanks!

roycaihw:
/assign @lavalamp 
for approval

dramich:
@lavalamp Have you had a chance to look at this?

lavalamp:
Sorry no I didn't see it

/approve

lavalamp:
(I would add a test if you want this to not break in the future!)

PR #94814, opened 2020-09-15
optimise defaultpreemption: enumerate fewer candidates

### LINKED ISSUE #89036: Optimization for scheduler preemption

adtac:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**: Instead of considering all nodes as preemption candidates, look for a smaller pool when dry running preemption and stop when a threshold number of candidates is found. In the average case (a good number of nodes are preemptible), this optimisation produces very good results (~25.3% throughput improvement). In the worst case (preemption is impossible), this approach will do no better or worse than the in-tree approach. The improvement is especially pronounced in large clusters (> 1k nodes). Clusters smaller than 100 nodes will not see any improvement from this optimisation.

The following benchmarks were performed with a 5,000 node cluster where 20,000 low priority pods are scheduled and then 5,000 high priority pods are scheduled (test run time is ~10 minutes). The high priority pods do not fit in any node without preempting one or more low priority pods. See the in-tree `scheduler_perf` integration benchmark for more details. Throughput percentile numbers are left out because median is 0 (kinda meaningless as a result).

| Quantity | Before | After | Change |
| - | - | - | - |
| Throughput average | 13.12 pods/sec | 16.45 pods/sec | +25.4% |
| `preemption_evaluation_seconds` average | 37.96ms | 30.20ms | -20.4% |
| `preemption_evaluation_seconds` p50 | 7.12ms | 6.22ms | -12.6% |
| `preemption_evaluation_seconds` p90 | 30.42ms | 24.01ms | -21.1% |
| `preemption_evaluation_seconds` p99 | 57.65ms | 41.75ms | -27.6% |
| Time spent in main thread (fraction of test time) | 6.18% | 2.51% | -59.3% |
| Time spent in parallel code (fraction of test time) | 16.26% | 6.25% | -61.6% |

Full diff (minus is after optimisation, plus is before): http://ix.io/2xHm

| Before | After |
| - | - |
| Screenshot_2020-09-15 scheduler_perf test cpu(2) | Screenshot_2020-09-15 scheduler_perf test cpu(3) |

No performance regression in 100 node tests (in fact, throughput is up slightly by 9%, not sure how).

Ref #89036 

```release-note
Fewer candidates are enumerated for preemption to improve performance in large clusters
```

```docs

```

/sig scheduling
/cc @alculquicondor @ahg-g


adtac:
Will add/modify tests tomorrow.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,33 +243,81 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
...
+func (co *candidateOffset) calculateOffset(numNodes int) (int, int) {
+	if numNodes < minCandidates {
+		return 0, numNodes
+	}
+	offset := co.offset
+	numCandidates := calculateNumCandidates(numNodes)
+	if co.offset >= numNodes {
+		// This will never happen unless numNodes is lower than what we saw last
+		// time (cluster resize?) and the previous offset is past numNodes. Reset
+		// the counter and restart from the beginning.
```
soulxu, line 280: I guess this will happen normally since the potential nodes are based on each pod's filter status:

`https://github.com/kubernetes/kubernetes/pull/94814/commits/80c36642e2d55f9707b210a76bd7120b30b098e5#diff-7ee2d8038590ef2f75c2bfb4e5a4e944R174`

I'm also thinking of the `offset` should be meaningful for the same set of potential nodes, but actually, we will get different set of potential nodes for each pod.

alculquicondor:
Did you add a new performance test case? Or are the above numbers from the existing ones?

Also, avoid `fixes #123` from the description, as that will cause github to close the issue once this merges.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,33 +243,81 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
...
+// every time. If <offset> + <numCandidates> runs past the number of nodes, the
+// caller is expected to wrap around the node list.
+func (co *candidateOffset) calculateOffset(numNodes int) (int, int) {
+	if numNodes < minCandidates {
+		return 0, numNodes
+	}
+	offset := co.offset
+	numCandidates := calculateNumCandidates(numNodes)
+	if co.offset >= numNodes {
+		// This will never happen unless numNodes is lower than what we saw last
```
alculquicondor, line 278: It does happen :)

```
// This happens when cluster is down scaled.
```

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,33 +243,81 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
 	return potentialNodes
 }
 
+// calculateNumCandidates returns the number of candidates the findCandidates
+// method must produce from dry running based on the constraints given by
+// maxCandidatesAsPercentageOfNodes and minCandidates.
+func calculateNumCandidates(numNodes int) int {
+	if numNodes < minCandidates || maxCandidatesAsPercentageOfNodes >= 100 {
```
alculquicondor, line 250: This conditional is not adding anything that the lines below can't handle. We can just save the branching.

ahg-g:
Just noticed a problem with the windowing approach, it breaks the promise of picking "A node with minimum number of PDB violations.": https://github.com/kubernetes/kubernetes/blob/5d095c8a4427954049341ef8c0464545c03f6acf/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go#L356

I am not sure we want to break that promise.

alculquicondor:
> I am not sure we want to break that promise.

I think it's ok to break it as long as the default is kept at 100%

ahg-g:
> > I am not sure we want to break that promise.
> 
> I think it's ok to break it as long as the default is kept at 100%

I think we can modify the promise in the following way:
1) the scheduler is free to select any node where the evicted pods don't violate pod disruption budget
2) if we can't find a node where the evicted pods don't violate pdb, then we have to select one where the evicted pods have the lowest impact on pdb

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: Once we standardize the args, I'd expect the validation logic (maxCandidatesAsPercentageOfNodes >= 100) to be put elsewhere.

BTW: `if numNodes < minCandidates` is already handled in `calculateOffset()`, right?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -48,13 +48,20 @@ import (
...
+	// maxCandidatesAsPercentageOfNodes.
+	minCandidates = 100
 )
 
 // DefaultPreemption is a PostFilter plugin implements the preemption logic.
 type DefaultPreemption struct {
 	fh        framework.FrameworkHandle
 	podLister corelisters.PodLister
 	pdbLister policylisters.PodDisruptionBudgetLister
+	co        *candidateOffset
```
Huang-Wei, line 64: Can we just make it `co *int`?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,33 +243,81 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
...
+	parallelCtx, cancel := context.WithCancel(ctx)
 
 	checkNode := func(i int) {
-		nodeInfoCopy := potentialNodes[i].Clone()
+		nodeInfoCopy := potentialNodes[(offset + i) % len(potentialNodes)].Clone()
 		stateCopy := state.Clone()
 		pods, numPDBViolations, fits := selectVictimsOnNode(ctx, fh, stateCopy, pod, nodeInfoCopy, pdbs)
 		if fits {
-			resultLock.Lock()
+			idx := atomic.AddInt32(&candidatesLen, 1)
```
Huang-Wei, line 303: Move it down to L308.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -150,11 +158,11 @@ func (pl *DefaultPreemption) preempt(ctx context.Context, state *framework.Cycle
 	return bestCandidate.Name(), nil
 }
 
-// FindCandidates calculates a slice of preemption candidates.
```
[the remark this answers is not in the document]
ahg-g: lets move those functions out of the plugin and place them in a separate file so that it is clear which logic is default plugin specific and which can be used by others.

Huang-Wei:
> I think we can modify the promise in the following way:
> 
> 1. the scheduler is free to select any node where the evicted pods don't violate pod disruption budget (the minimum within the window)
> 2. if we can't find a node where the evicted pods don't violate pdb, then we have to select one where the evicted pods have the lowest impact on pdb across all nodes

The semantics looks good to me, however, in practice, we should evaluate the difficulty of getting them implemented. If it's too much work (I'd assume we need to pre-know the nodes that don't have pods violating PDBs, so as to prioritize them in the searching scope), maybe it's ok to break the promise - as we're just increasing the disruption budget by 1, rather than totally breaking the PDB violation (semantics of PDB is guarded by API server).

alculquicondor:
> The semantics looks good to me, however, in practice, we should evaluate the difficulty of getting them implemented. 

It shouldn't be hard at all. It's a matter of continuing iterating while:
- a node with zero violations is not found, OR
- percentage has not been reached.

adtac:
Yes, it's just a matter of changing the context cancel condition slightly; we already have the PDB violation data for each node to make that decision. I'm still working on the component config changes haha, they're pretty confusing :)

Huang-Wei:
> It shouldn't be hard at all. It's a matter of continuing iterating while:
> 
> * a node with zero violations is not found, OR
> * percentage has not been reached.

What if the percentage has been reached, and all so-far-calculated candidates have non-zero violations? Would you continue searching?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -48,13 +48,20 @@ import (
 const (
 	// Name of the plugin used in the plugin registry and configurations.
 	Name = "DefaultPreemption"
+	// Maximum number of candidates to shortlist when dry running preemption as a
+	// percentage of number of nodes.
+	maxCandidatesAsPercentageOfNodes = 10
```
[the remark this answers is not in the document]
ahg-g: How about we use two parameters, percentage (`maxCandidatesAsPercentage`) and explicit value (`maxCandidates`). If set to higher than 0, `maxCandidates` takes precedence.  `minCandidates` should simply be hardcoded at 2 all the time.

I think this is a good case for the power of two choices: http://www.eecs.harvard.edu/~michaelm/NEWWORK/postscripts/twosurvey.pdf

We need to think about how to make the selection random, perhaps we can randomize the starting offset of node search, but that might not be enough.

adtac:
>What if the percentage has been reached, and all so-far-calculated candidates have non-zero violations? Would you continue searching?

Yes. The exit condition is basically `len(candidates) > threshold && nonViolatingNodeFound`. In the worst case (all nodes need to be searched), we don't do any better/worse than the in-tree approach. In the average case, we check fewer nodes.

Huang-Wei:
> Yes. The exit condition is basically `len(candidates) > threshold && nonViolatingNodeFound`. In the worst case (all nodes need to be searched), we don't do any better/worse than the in-tree approach. In the average case, we check fewer nodes.

SG.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: Done (int instead of *int).

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: Done.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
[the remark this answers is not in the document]
adtac: One possible issue with that is in small-medium sized clusters with the default configuration. Consider a cluster with 50 nodes. Let's say a good default for `maxCandidatesAsPercentageOfNodes` is 10% (doesn't matter what it is as long as it's non-100%). If the cluster does not override `maxCandidatesAsPercentageOfNodes`, only 5 nodes will be enlisted. Is this sufficient? In such small clusters, the performance benefits between checking just 5 nodes will be outweighed by the better preemption we might get from checking all 50 nodes (which wouldn't be *that* slow).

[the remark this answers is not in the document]
ahg-g: I am trying to allow using an absolute value without introducing a third parameter. I guess we can do that using the two parameters you have: if `maxCandidatesAsPercentage` is set to 0 `MinCandidates` becomes the maximum in absolute terms. Lets explicitly clarify that the number of nodes we look at is `max(nodeCount* MaxCandidatesAsPercentage, MinCandidates)`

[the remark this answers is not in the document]
ahg-g: sorry, should have mentioned this earlier, it would have probably been easier if we split the PR into API change and core scheduler change (the api change requires api review), but it is fine since you did that already.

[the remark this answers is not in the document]
Huang-Wei: @adtac Do you want to exercise using type `IntOrString` to represent an "either percentage or absolute number" semantic?

https://github.com/kubernetes/kubernetes/blob/f1012030f78ce73919ef426673b064809abc58dc/staging/src/k8s.io/api/core/v1/types.go#L4180

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
soulxu: > Perhaps I misunderstood your comment, could you please elaborate what you meant by "offset should be meaningful for the same set of potential nodes"?

I mean, for example, we have 4 nodes: A, B ,C, D. then we try A and B first time, we have offset to record that next start from C. Then next around will try C and D. But due to the potential nodes are based on filter status. So maybe for PodA, we got A and B as potential  nodes only. After try the A and B, the offset back to 0. But for PodB, we got A, B, C and D as potential nodes.  Since offset is 0, then we try A and B again. Sorry for not describe clearly in the beginning :)

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,32 +250,89 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
...
+					cancel()
+				}
+			}
+			mu.Unlock()
+		}
+	}
+	parallelize.Until(parallelCtx, len(potentialNodes), checkNode)
+
+	var candidates []Candidate
+	for _, c := range append(nonViolatingCandidates, violatingCandidates...) {
```
alculquicondor, line 330: actually, if we have nonViolating Candidates... we won't be using the violating ones, so you can discard the later ones.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,32 +250,89 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
...
-	var candidates []Candidate
+// and returns preemption candidates. The number of candidates depends on the
+// constraints defined in the plugin's args. In the returned list of
+// candidates, ones that do not violate PDB are preferred over ones that do.
+func (pl *DefaultPreemption) dryRunPreemption(ctx context.Context, fh framework.PreemptHandle,
+	state *framework.CycleState, pod *v1.Pod, potentialNodes []*framework.NodeInfo,
+	pdbs []*policy.PodDisruptionBudget) []Candidate {
+	offset, numCandidates := pl.calculateOffset(int32(len(potentialNodes)))
+	var mu sync.Mutex
+	var nonViolatingCandidates []Candidate
```
alculquicondor, line 296: worth initializing with capacity? Do only if `append` shows significant overhead in the profiling.

pkg/scheduler/apis/config/v1beta1/defaults.go
```diff
@@ -158,6 +158,15 @@ func SetDefaults_KubeSchedulerConfiguration(obj *v1beta1.KubeSchedulerConfigurat
 	}
 }
 
+func SetDefaults_DefaultPreemptionArgs(obj *v1beta1.DefaultPreemptionArgs) {
+	if obj.MaxCandidatesAsPercentageOfNodes == nil {
+		obj.MaxCandidatesAsPercentageOfNodes = pointer.Int32Ptr(10)
```
alculquicondor, line 163: @Huang-Wei What do you think about 10%?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: We cannot assert, in this case, violating ones would be useless, as later we will call extenders, and also this function should serve as a common library for both in-tree and out-of-tree plugins - out-of-tree plugins may rely on the candidates and choose a different preemption strategy.

However, the logic can be simplified a bit:

```
candidates := append(nonViolatingCandidates, violatingCandidates...)
if len(candidates) >= numCandidates {
    return candidates[:numCandidates]
}
return candidates
```

pkg/scheduler/apis/config/v1beta1/defaults.go
Huang-Wei: LGTM.

pkg/scheduler/apis/config/types_pluginargs.go
```diff
@@ -23,6 +23,21 @@ import (
...
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
+// DefaultPreemptionArgs holds arguments used to configure the
+// DefaultPreemption plugin.
+type DefaultPreemptionArgs struct {
+	metav1.TypeMeta
+
+	// MaxCandidatesAsPercentageOfNodes is the maximum number of candidates to
+	// shortlist when dry running preemption as a percentage of number of nodes.
+	MaxCandidatesAsPercentageOfNodes int32
```
Huang-Wei, line 33: I feel the name can be simplified, like `PercentageOfCandidateNodes`?

alculquicondor: or `MaxCandidateNodesPercentage`. I think it's a good practice to have the "units" at the end of the name.

adtac:
Unfortunately, github won't let me reply to comments until I post my pending comments, so I'll copy paste the important ones until I'm ready to post the whole set of comments (which I'll do once I push my changes -- currently writing tests).

>I feel the name can be simplified, like PercentageOfCandidateNodes?

technically, it's the percentage of nodes that are considered to be potential candidates, so "percentage of candidate nodes" would be slightly misleading (but not really since it's just yet another filtering phase within the plugin).

I'm open to the suggestion as it's simpler though :+1:

Huang-Wei:
> I'm open to the suggestion as it's simpler though

@alculquicondor 's suggestion (`MaxCandidateNodesPercentage`) looks good to me.

(btw: there are no much points we can refer across k8s codebase, but I do find a name `CurrentCPUUtilizationPercentage` in autoscaler, so I think keeps the "units" at the end is more conventional, which is also aligned with prometheus metric names)

pkg/scheduler/apis/config/types_pluginargs.go
adtac: Done.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: re @Huang-Wei's suggestion - this wouldn't let the garbage collector reclaim the backing array. For example, let's say `numCandidates` is 100, but we had to dry run preemption on 500 nodes to finally find a non-violating node. This would mean len(candidates) is 500. Even though the returned slice will have a len/cap of 100, the backing array will never be garbage collected and all 500 elements will always be in memory.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: I understand what you mean now. I agree that could happen, but I also think that scenario is less common than the case that round-robin solves. Round-robin guarantees that each node will have been tried for preemption in the average case after a certain number of pods.

(average case = nearly all nodes are available for preemption)

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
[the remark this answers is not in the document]
adtac: re absolute maximum: I don't see the value of such an option immediately. While the 0% workaround would work, why would a cluster want to set an absolute value?

re the IntOrString type field -- I'm not a big fan of untyped fields, especially given YAML's reputation for being confusing :) I think it might be very confusing

>Lets explicitly clarify that the number of nodes we look at is max(nodeCount* MaxCandidatesAsPercentage, MinCandidates)

Done.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
ahg-g: why do we need to iterate here at all? just return the merged slice, it doesn't really matter if we are returning more than numCandidates.

pkg/scheduler/apis/config/types_pluginargs.go
```diff
@@ -23,6 +23,21 @@ import (
...
+	// MinCandidates is the minimum number of candidates to shortlist. Has
+	// precendence over maxCandidatesAsPercentageOfNodes.
```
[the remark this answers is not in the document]
ahg-g: the comment is probably outdated

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
alculquicondor: Also, there is the `sort` after this, isn't there? You could have only one slice and just a counter for the non-violating Nodes. And then you can remove the lock.

ahg-g: right, you could initialize the slice with numCandidates capacity, and use atomic increments for the index and the non-violating counter, see example here: https://github.com/kubernetes/kubernetes/blob/f99ed92ed149e8cbb09dfcdcff78e6fb4024d3a0/pkg/scheduler/framework/plugins/interpodaffinity/scoring.go#L194

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -125,7 +140,7 @@ func (pl *DefaultPreemption) preempt(ctx context.Context, state *framework.Cycle
 	}
 
 	// 2) Find all preemption candidates.
-	candidates, err := FindCandidates(ctx, cs, state, pod, m, ph, nodeLister, pl.pdbLister)
+	candidates, err := FindCandidates(ctx, cs, state, pod, m, ph, pl, nodeLister, pl.pdbLister)
```
Huang-Wei, line 143: Can we hold passing the whole `pl` to `FindCandidates`? `pl` here represents the whole DefaultPreemption struct.

An alternative is to make `candidateOffset` a pointer, and `args` needs to be passed as well.

alculquicondor: +1. Passing the whole Plugin would make the function unusable for custom plugins.

-1 to Passing args. This is the configuration struct for this particular plugin. Custom plugins shouldn't have to import the internal config package. It think we should use the options pattern. `WithCandidateOffset`, `WithMaxPercentage`, etc.

Huang-Wei: > -1 to Passing args. This is the configuration struct for this particular plugin. Custom plugins shouldn't have to import the internal config package.

Good point.

pkg/scheduler/apis/config/types_pluginargs.go
```diff
@@ -23,6 +23,24 @@ import (
...
+// DefaultPreemption plugin.
+type DefaultPreemptionArgs struct {
+	metav1.TypeMeta
+
+	// MaxCandidateNodesPercentage is the maximum number of candidates to
+	// shortlist when dry running preemption as a percentage of number of nodes.
+	MaxCandidateNodesPercentage int32
+	// MinCandidates is the minimum number of candidates to shortlist. The
+	// likely number of candidates enumerated for dry running preemption is given
+	// by the formula max(NumNodes * MaxCandidateNodesPercentage, MinCandidates).
```
hex108, line 36: Since the goal is to reduce the number of candidates, why not use the formula `min(NumNodes * MaxCandidateNodesPercentage, MinCandidates)`?

If we use `max(...)`,  will `MinCandidateNodesPercentage` be a better name?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,32 +250,87 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
...
+
+// calculateOffset returns an <offset> and <numCandidates> that the caller must
+// use to find candidates. A variable offset is used as the starting index when
+// dry running preemption so that the same set of nodes don't get preempted
+// every time. If <offset> + <numCandidates> runs past the number of nodes, the
+// caller is expected to wrap around the node list.
+func (pl *DefaultPreemption) calculateOffset(numNodes int32) (int32, int32) {
+	offset := pl.candidateOffset
+	numCandidates := pl.calculateNumCandidates(numNodes)
+	if pl.candidateOffset >= numNodes {
```
hex108, line 276: We aslo need update `offset` if `offset >= numNodes`?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,32 +250,87 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
...
+	}
+	return n
+}
+
+// calculateOffset returns an <offset> and <numCandidates> that the caller must
+// use to find candidates. A variable offset is used as the starting index when
+// dry running preemption so that the same set of nodes don't get preempted
+// every time. If <offset> + <numCandidates> runs past the number of nodes, the
+// caller is expected to wrap around the node list.
+func (pl *DefaultPreemption) calculateOffset(numNodes int32) (int32, int32) {
```
hex108, line 273: According to the function, it just returns `numCandidates`  and previous `pl.candidateOffset`. Would it be better to  get offset from `pl.candidateOffset` and rename the function name to `updateOffset()`?

pkg/scheduler/apis/config/types_pluginargs.go
alculquicondor: The goal is to reduce the number of candidates in a controlled way. `max` seems safer.

`MinCandidateNodesPercentage` is actually more accurate, given that we won't stop until we find at least one Node where PDB is not violated.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
alculquicondor: not a big deal, as we use mod later, but yeah, we can

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: Done.

pkg/scheduler/apis/config/types_pluginargs.go
adtac: Good catch, changed the two arg names to `minCandidateNodesPercentage` and `minCandidateNodesAbsolute`. Should be much more self-explanatory now.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: Done.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: Done (renamed to updateAndReturnOffset), but it's simpler to have this function return the correct offset than use candidateOffset directly because of some corner cases around the zero value.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,33 +318,73 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
 	return potentialNodes
 }
 
+type candidateList struct {
+	idx int32
+	max int32
```
ahg-g, line 323: cap(arr) is the max

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,33 +318,73 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
...
+	return &candidateList{max: max, idx: -1, arr: make([]Candidate, max)}
+}
+
+// add adds a new candidate to the internal array atomically.
+func (cl *candidateList) add(c *candidate) {
+	if idx := atomic.AddInt32(&cl.idx, 1); idx < cl.max {
+		cl.arr[idx] = c
+	} else {
+		atomic.AddInt32(&cl.idx, -1)
+	}
```
ahg-g, line 337: not necessary, right?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,33 +318,73 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
...
+
+// size returns the number of candidate stored atomically.
+func (cl *candidateList) size() int32 {
+	return atomic.LoadInt32(&cl.idx) + 1
+}
+
+// get returns the internal candidate array. This function is NOT atomic and
+// assumes that the caller has acquired mutual exclusion.
+func (cl *candidateList) get() []Candidate {
+	return cl.arr[:cl.idx+1]
```
ahg-g, line 348: why do we want a copy? this is getting appended to create a slice anyways at line 387

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -66,6 +66,59 @@ type DefaultPreemption struct {
 
 var _ framework.PostFilterPlugin = &DefaultPreemption{}
 
+type findCandidatesOpts struct {
+	candidateOffset       *int32
+	defaultPreemptionArgs *config.DefaultPreemptionArgs
```
alculquicondor, line 71: do not use pointers for this

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -66,6 +66,59 @@ type DefaultPreemption struct {
...
+		// beginning.
+		*fco.candidateOffset = 0
+		offset = 0
+	} else {
+		*fco.candidateOffset = (*fco.candidateOffset + numCandidates) % numNodes
+	}
+	return offset, numCandidates
+}
+
+func withDefaultPreemptionArgs(args *config.DefaultPreemptionArgs) func(*findCandidatesOpts) {
```
alculquicondor, line 110: This function needs to be exported. Because of that, it shouldn't depend on DefaultPreemptionArgs. Just expose each parameter individually.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -169,7 +222,7 @@ func (pl *DefaultPreemption) preempt(ctx context.Context, state *framework.Cycle
 // Each candidate is executable to make the given <pod> schedulable.
 func FindCandidates(ctx context.Context, cs kubernetes.Interface,
 	state *framework.CycleState, pod *v1.Pod, m framework.NodeToStatusMap, ph framework.PreemptHandle,
-	pl *DefaultPreemption, nodeLister framework.NodeInfoLister, pdbLister policylisters.PodDisruptionBudgetLister) ([]Candidate, error) {
+	nodeLister framework.NodeInfoLister, pdbLister policylisters.PodDisruptionBudgetLister, opts ...interface{}) ([]Candidate, error) {
```
alculquicondor, line 225: create a type for the options. See https://github.com/kubernetes/kubernetes/blob/90be932b600dd277601614840fbe510161fc813d/pkg/scheduler/scheduler.go#L183

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
alculquicondor: rather `len`

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: Done.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: Ah I see, done.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: Done.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: Done.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: Done.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: This is just a slice, not a copy. If there are fewer than `cl.idx+1`, then `cl.arr` will have nil padding, which we don't want to append (`arr` has len = cap).

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -53,21 +56,95 @@ const (
 // DefaultPreemption is a PostFilter plugin implements the preemption logic.
 type DefaultPreemption struct {
 	fh        framework.FrameworkHandle
+	args      config.DefaultPreemptionArgs
 	podLister corelisters.PodLister
 	pdbLister policylisters.PodDisruptionBudgetLister
+	// candidateOffset maintains state about the last offset used for dry running
```
ahg-g, line 62:
```suggestion
	// candidateOffset is the last offset used for dry running
```

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -53,21 +56,95 @@ const (
...
+	MinCandidateNodesAbsolute   int32
+}
+
+type FindCandidatesOption func(*FindCandidatesOptions)
+
+// calculateNumCandidates returns the number of candidates the FindCandidates
+// method must produce from dry running based on the constraints given by
+// <minCandidateNodesPercentage> and <minCandidateNodesAbsolute>. The number of candidates
+// returned will never be greater than <numNodes>.
+func (fco *FindCandidatesOptions) calculateNumCandidates(numNodes int32) int32 {
```
ahg-g, line 81: this needs to be calculated once, lets just do that at New and store it in a plugin variable.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -53,21 +56,95 @@ const (
...
 
 var _ framework.PostFilterPlugin = &DefaultPreemption{}
 
+type FindCandidatesOptions struct {
+	CandidateOffset             *int32
+	MinCandidateNodesPercentage int32
+	MinCandidateNodesAbsolute   int32
+}
+
+type FindCandidatesOption func(*FindCandidatesOptions)
```
ahg-g, line 75: I honestly feel like we are using a big hammer to hit a small nail :) I don't think the "With option" pattern applicable to what we need here, too much boiler plate for just calculating the number of nodes and offset.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,33 +327,74 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
...
+	arr []Candidate
+}
+
+func newCandidateList(size int32) *candidateList {
+	return &candidateList{idx: -1, arr: make([]Candidate, size)}
+}
+
+// add adds a new candidate to the internal array atomically.
+func (cl *candidateList) add(c *candidate) {
+	if idx := atomic.AddInt32(&cl.idx, 1); idx < int32(len(cl.arr)) {
```
ahg-g, line 341: semantically speaking `cap` not `len` is the max capacity, which is what we want here.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: So that we only need 2 options: WithCandidateOffset() and WithNumCandidates().

Huang-Wei: In that case, out-of-tree that has its own pluginArgs (or empty pluginArgs) won't be able to get the benefits.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: For in-tree defaultpreemption plugin, it does. However, it's more for out-of-tree plugins to use FindCandidates() while also enjoying the feature of looping over fewer candidates.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,33 +327,74 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
...
+		cl.arr[idx] = c
+	}
+}
+
+// size returns the number of candidate stored atomically.
+func (cl *candidateList) size() int32 {
```
Huang-Wei, line 347: Is this equivalent to `len(arr)`?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,33 +327,74 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
 	return potentialNodes
 }
 
+type candidateList struct {
+	idx int32
+	arr []Candidate
```
Huang-Wei, line 332: arr = array? how about `eles` or `elements`.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -66,19 +66,22 @@ type DefaultPreemption struct {
 
 var _ framework.PostFilterPlugin = &DefaultPreemption{}
 
-type findCandidatesOpts struct {
-	candidateOffset       *int32
-	defaultPreemptionArgs *config.DefaultPreemptionArgs
+type FindCandidatesOptions struct {
```
alculquicondor, line 69: this struct doesn't need to be exported.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
alculquicondor: yeah, the idea is to make `FindCandidates` stateless

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
alculquicondor: `items`, shorter but still a full word :)

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
alculquicondor: not really (at least in golang), because the slice is already initialized with `len = cap` and if you try to access an index that is `> len` but `< cap`  in go, that's produces a panic.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -254,13 +267,9 @@ func FindCandidates(ctx context.Context, cs kubernetes.Interface,
 		return nil, err
 	}
 
-	fco := &findCandidatesOpts{}
+	fco := &FindCandidatesOptions{}
```
alculquicondor, line 270: Note that if a caller doesn't put any `With` options, the values will be all zero, thus 0 candidates will be returned. That wouldn't be a good default. I think the default can be just 100%. @Huang-Wei ?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: Since it's `make([]Candidate, N)`, it should be `len` right? because we're directly setting the array elements, not appending to an allocated array. Even if `cap` were greater than `len` hypothetically, we'd still want to check if `idx < len`, not `idx < cap` because `idx < cap` wouldn't guarantee that `arr[idx]` would work.

ahg-g: Ah, I thought cap defines the upper range of the index.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
ahg-g: I think we are lacking proper separation between what we want to export for out-of-tree preemption plugins and what is specific to the default preemption plugin. Having everything here is honestly bad and confusing.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
ahg-g: Sg

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: True, however, a neat default configuration for default-scheduler outweighs the out-of-tree callers IMO.

How about adding a comment in FindCandidates() to remind callers to set the percentage to 100 if they want to return all candidates?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
Huang-Wei: Agree. We should polish the logic, maybe in a follow-up. Also I'm not quite satisfied with the current function signature - to make it stateless, it even has 8 parameters! Which is way too much, needs to shrink the args number a bit.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -180,11 +261,22 @@ func FindCandidates(ctx context.Context, cs kubernetes.Interface, state *framewo
...
 	}
-	return dryRunPreemption(ctx, ph, state, pod, potentialNodes, pdbs), nil
+
+	fco := &FindCandidatesOptions{}
+	for _, opt := range opts {
+		opt(fco)
+	}
+	offset, numCandidates := fco.updateAndReturnOffset(int32(len(potentialNodes)))
+	if klog.V(5).Enabled() {
+		klog.Infof("dry running preemption to find ~%d candidates (starting offset: %d)", numCandidates, offset)
```
soulxu, line 276: How about move the log line 258~262 to here, then we can get the node name also.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: I think it's cleaner to add a sanity check in `calculateNumCandidates` so that if the returned value is <= 0, set it to `numNodes`. Done.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: I completely misunderstood why we want FindCandidates exported haha. The member fields do not need to be exported either; only the `WithOption` functions need to I guess.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: https://github.com/kubernetes/kubernetes/issues/95131

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: Done.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
ahg-g: I still don't think we need all of that, I expect the refactoring to result in a new type (e.g. `PreemptionEvaluator`) and a group of member methods. The refactoring will basically move the state to this new type along with the functions.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -150,11 +165,47 @@ func (pl *DefaultPreemption) preempt(ctx context.Context, state *framework.Cycle
...
+	}
+	return offset, numCandidates
+}
+
 // FindCandidates calculates a slice of preemption candidates.
 // Each candidate is executable to make the given <pod> schedulable.
-func FindCandidates(ctx context.Context, cs kubernetes.Interface, state *framework.CycleState, pod *v1.Pod,
-	m framework.NodeToStatusMap, ph framework.PreemptHandle, nodeLister framework.NodeInfoLister,
-	pdbLister policylisters.PodDisruptionBudgetLister) ([]Candidate, error) {
+func FindCandidates(ctx context.Context, cs kubernetes.Interface,
```
ahg-g, line 206: `func (pl *DefaultPreemption)`  instead of passing it as an argument.

ahg-g: ph, nodeLister and pdbLister can all be accessed from pl as well.

ahg-g: sorry, `cs` as well.

adtac: Ah, I missed that! Done.

ahg-g:
looks good to me, I think you can squash, @Huang-Wei @alculquicondor ?

pkg/scheduler/internal/parallelize/parallelism.go
```diff
@@ -23,13 +23,15 @@ import (
 	"k8s.io/client-go/util/workqueue"
 )
 
-const parallelism = 16
+// Parallelism is exported for tests where non-determinism from parallelism is
+// not desired. Do NOT modify outside tests.
+var Parallelism = 16
```
alculquicondor, line 28: Could we avoid this by:

1. Having a test for 100% of nodes.
2. Having another test for different percentages, validate the size of the list and makes sure all the elements belong to the known list of all preemptible Nodes.

adtac: I don't think it's worth it IMHO. The size of the list isn't deterministic (in spite of cancelling the context, some dry runs could already be in-flight) and the elements of the list aren't deterministic. Due to the small number of nodes in tests (5) and the large parallelism (16), we would need to add all nodes to the acceptable list of candidates, which defeats the purpose of the test entirely.

alculquicondor: If a dry run is already in-flight, it's result would be discarded because the limited was exceeded, or not?

adtac: It wouldn't be discarded. Once a dry run begins, if it can fit the preemptor, it will be added to one of the lists. And that is fine, `numCandidates` is only an approximation; it doesn't really matter if we dry run 100 nodes or 102 nodes.

alculquicondor: I see. FYI @hprateek43 for #94636

alculquicondor: I'm ok with the change, it's just that in general we should think of all our options before exposing something just for tests, specially if it may change the behavior. The test can be designed to allow non-deterministic results.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -246,185 +259,215 @@ func TestPostFilter(t *testing.T) {
...
-		pod                     *v1.Pod
-		pods                    []*v1.Pod
+		testPods                []*v1.Pod
+		initPods                []*v1.Pod
 		registerPlugins         []st.RegisterPluginFunc
 		pdbs                    []*policy.PodDisruptionBudget
 		fakeFilterRC            framework.Code // return code for fake filter plugin
-		expected                []Candidate
-		expectedNumFilterCalled int32
+		DisableParallelism      bool
```
Huang-Wei, line 269:
```suggestion
		disableParallelism      bool
```

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -613,20 +680,201 @@ func TestDryRunPreemption(t *testing.T) {
...
+		{
+			name: "preemption offset across multiple scheduling cycles",
+			args: &config.DefaultPreemptionArgs{MinCandidateNodesPercentage: 40, MinCandidateNodesAbsolute: 1},
+			registerPlugins: []st.RegisterPluginFunc{
+				st.RegisterPluginAsExtensions(noderesources.FitName, noderesources.NewFit, "Filter", "PreFilter"),
+			},
+			nodeNames: []string{"node1", "node2", "node3", "node4", "node5"},
+			testPods: []*v1.Pod{
+				st.MakePod().Name("tp1").UID("tp1").Priority(highPriority).Req(largeRes).Obj(),
+				st.MakePod().Name("tp2").UID("tp2").Priority(highPriority).Req(largeRes).Obj(),
```
Huang-Wei, line 781: Can you add a testPod "tp3" to verify the with slice boundaries? tp3 is expected to come up with candidates [node5, node1], right?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
```diff
@@ -673,36 +921,59 @@ func TestDryRunPreemption(t *testing.T) {
...
-				t.Fatal(err)
-			}
-			// Sort the values (inner victims) and the candidate itself (by its NominatedNodeName).
-			for i := range got {
-				victims := got[i].Victims().Pods
-				sort.Slice(victims, func(i, j int) bool {
-					return victims[i].Name < victims[j].Name
+			sort.Slice(nodeInfos, func(i, j int) bool {
+				return nodeInfos[i].Node().Name < nodeInfos[j].Node().Name
+			})
```
Huang-Wei, line 930: Is this necessary? I'd assume the test cases to have specified the nodes ordered by name.

adtac: Yes, otherwise the returned node list is in arbitrary order for some reason and there's no way to deterministically test for correct offsets and wrap arounds.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption_test.go
adtac: Great point! Done.

pkg/scheduler/internal/parallelize/parallelism.go
adtac: Yep, I agree that we can test non-deterministic results with some significant test refactor, but the benefit vs cost is just too high to justify it IMO. Besides, I think there will be a cleaner API to set parallelism once https://github.com/kubernetes/kubernetes/pull/94636 is merged.

alculquicondor (commented):
/approve
for scheduler

I would say it's ready for squash

Huang-Wei:
@liggitt could you help review/approve the API changes? Thanks!

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -167,24 +216,28 @@ func FindCandidates(ctx context.Context, cs kubernetes.Interface, state *framewo
...
 		}
 		return nil, nil
 	}
+
+	pdbs, err := getPodDisruptionBudgets(pl.pdbLister)
+	if err != nil {
+		return nil, err
+	}
+
+	offset, numCandidates := pl.updateAndReturnOffset(int32(len(potentialNodes)))
```
liggitt, line 231: would it be simpler to pick a random start point rather than making *DefaultPreemption stateful?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,33 +288,74 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
...
+	if n >= int32(len(cl.items)) {
+		n = int32(len(cl.items))
+	}
+	return n
+}
+
+// get returns the internal candidate array. This function is NOT atomic and
+// assumes that the caller has acquired mutual exclusion on the list.
+func (cl *candidateList) get() []Candidate {
+	return cl.items[:cl.size()]
```
liggitt, line 319: returning a subslice here means that callers can modify members of `cl.items` beyond `cl.size()` which will be overwritten if cl.add is called again (see https://play.golang.org/p/Uw_cq4NK4Tz)

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
```diff
@@ -23,6 +23,25 @@ import (
...
+
+	// MinCandidateNodesPercentage is the minimum number of candidates to
+	// shortlist when dry running preemption as a percentage of number of nodes.
+	MinCandidateNodesPercentage *int32 `json:"minCandidateNodesPercentage,omitempty"`
+	// MinCandidateNodesAbsolute is the absolute minimum number of candidates to
+	// shortlist. The likely number of candidates enumerated for dry running
+	// preemption is given by the formula:
+	// numCandidates = max(numNodes * minCandidateNodesPercentage, minCandidateNodesAbsolute)
+	// We say "likely" because there are other factors such as PDB violations
+	// that play a role in the number of candidates shorlisted.
```
liggitt, line 39:
```suggestion
	// that play a role in the number of candidates shortlisted.
```

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
```diff
@@ -23,6 +23,25 @@ import (
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
+// DefaultPreemptionArgs holds arguments used to configure the
+// DefaultPreemption plugin.
+type DefaultPreemptionArgs struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// MinCandidateNodesPercentage is the minimum number of candidates to
+	// shortlist when dry running preemption as a percentage of number of nodes.
```
liggitt, line 32: document default, min, and max

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
liggitt, line 39: document default and min

pkg/scheduler/apis/config/v1beta1/defaults.go
```diff
@@ -158,6 +158,15 @@ func SetDefaults_KubeSchedulerConfiguration(obj *v1beta1.KubeSchedulerConfigurat
...
 
+func SetDefaults_DefaultPreemptionArgs(obj *v1beta1.DefaultPreemptionArgs) {
+	if obj.MinCandidateNodesPercentage == nil {
+		obj.MinCandidateNodesPercentage = pointer.Int32Ptr(10)
+	}
+	if obj.MinCandidateNodesAbsolute == nil {
+		obj.MinCandidateNodesAbsolute = pointer.Int32Ptr(100)
+	}
+}
+
```
liggitt, line 169: do these match the current behavior or do they represent a change in scheduler behavior?

ahg-g: This is a change in the scheduler behavior, but it is a change to a scoring behavior which is by definition best effort.

ahg-g: Let me expand on this, the invariants that are documented with regard to preemption are two (https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/): 1) preempt lower priority pods, 2) respect PDB whenever possible. Those two are preserved. Beyond that, selecting which exact pods to preempt isn't part of the contract, it is based on a heuristic.

liggitt: I see, so this is not saying we will only search 10% / 100 nodes to find candidate nodes, it is saying once we have 10% / 100 of nodes as candidates, we'll stop looking

ahg-g: yes, exactly.

pkg/scheduler/apis/config/validation/validation_pluginargs.go
```diff
@@ -26,6 +26,38 @@ import (
...
+	if args.MinCandidateNodesPercentage == 0 && args.MinCandidateNodesAbsolute == 0 {
+		return fmt.Errorf("both minCandidateNodesPercentage and minCandidateNodesAbsolute cannot be zero")
+	}
+	return nil
+}
+
+// validateMinCandidateNodesPercentage validates that
+// minCandidateNodesPercentage is within the allowed range.
+func validateMinCandidateNodesPercentage(minCandidateNodesPercentage int32) error {
+	if minCandidateNodesPercentage < 0 || minCandidateNodesPercentage > 100 {
```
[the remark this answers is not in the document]
adtac: the absolute config parameter has precedence over the percentage one, so one can use `minCandidateNodesPercentage = 0%` to denote that they only want to use an absolute minimum

there's a problem only if both the absolute and the percentage parameter are both zero as that would break things; and we check for that scenario in `ValidateDefaultPreemptionArgs`

pkg/scheduler/apis/config/validation/validation_pluginargs.go
```diff
@@ -26,6 +26,38 @@ import (
...
+		return fmt.Errorf("minCandidateNodesPercentage is not in the range [0, 100]")
+	}
+	return nil
+}
+
+// validateMinCandidateNodesAbsolute validates that minCandidateNodesAbsolute
+// is within the allowed range.
+func validateMinCandidateNodesAbsolute(minCandidateNodesAbsolute int32) error {
+	if minCandidateNodesAbsolute < 0 {
```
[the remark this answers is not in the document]
adtac: setting `minCandidateNodesAbsolute = 0` could be used to turn off the absolute knob entirely. For example, if an operator wants only 10% of the cluster size to be evaluated without an absolute minimum, they can do this

see previous comment re both absolute and percentage both being zero

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: we did consider that, but the round robin-like approach guarantees that it will evenly distributes preemption across all nodes, whereas a random approach could result in confusing results to the end-user (e.g. the same node getting picked for preemption twice in a row)

but I totally see why a random starting index might be nice to have too -- this is something that's certainly worth exploring in the future once we have more feedback from users

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: IMO this isn't a concern. `get()` is guaranteed to be called after all `add()` operations, but perhaps it's worth adding this as a note to the function documentation -- done.

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
adtac: Done.

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
adtac: Done.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
ahg-g: picking the same node twice isn't an issue (not doing so is not part of the contract), it could happen with the existing logic as well.

pkg/scheduler/apis/config/types_pluginargs.go
```diff
@@ -23,6 +23,25 @@ import (
...
+
+	// MinCandidateNodesPercentage is the minimum number of candidates to
+	// shortlist when dry running preemption as a percentage of number of nodes.
+	MinCandidateNodesPercentage int32
+	// MinCandidateNodesAbsolute is the absolute minimum number of candidates to
+	// shortlist. The likely number of candidates enumerated for dry running
+	// preemption is given by the formula:
+	// numCandidates = max(numNodes * minCandidateNodesPercentage, minCandidateNodesAbsolute)
+	// We say "likely" because there are other factors such as PDB violations
+	// that play a role in the number of candidates shorlisted.
```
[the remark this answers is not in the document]
adtac: Done.

pkg/scheduler/apis/config/v1beta1/defaults.go
adtac: It is also worth noting that we'll potentially keep going even after 10% / 100 nodes if we don't have a PDB-respecting candidate. So the 10% / 100 is only a lower limit (the upper limit being the entire cluster, which is the preemption behaviour in-tree).

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
adtac: Done.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: >it could happen with the existing logic as well

picking the same node twice *in a row* could happen, but if that happens, it also guarantees that all other nodes have been tried

liggitt: making this stateful and self-modifying makes it a lot harder to reason about (especially in terms of thread safety and data races)

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
liggitt: making the other methods atomically safe implies multiple goroutines are interacting with this object... are all parallel operations guaranteed to be stopped before this is called or can a timeout return control to the get() caller while some background operations are still running?

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
```diff
@@ -235,33 +288,74 @@ func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.Node
...
+
+// add adds a new candidate to the internal array atomically.
+func (cl *candidateList) add(c *candidate) {
+	if idx := atomic.AddInt32(&cl.idx, 1); idx < int32(len(cl.items)) {
+		cl.items[idx] = c
+	}
+}
+
+// size returns the number of candidate stored atomically.
+func (cl *candidateList) size() int32 {
```
liggitt, line 308: size() is also not atomic and assumes all add() operations have been completed

pkg/scheduler/internal/parallelize/parallelism.go
liggitt: can you add a todo to make this private again once #94636 merges (or as part of that PR?)

alculquicondor:
can we get this to a mergeable state? code freeze is approaching

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: >are all parallel operations guaranteed to be stopped before this is called

Yes, parallelize.Until will guarantee that all goroutines interacting with this candidateList will have completed before this called.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: True, but we also don't access the list elements until all add() operations complete. I've updated the function doc to reflect this more accurately.

pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
adtac: Done. Switched to a random offset. There is no difference in throughput between random and moving offset in integration benchmarks (as expected).

adtac:
ping @liggitt :)

liggitt:
/approve
for API bits

scheduler reviewers have LGTM

PR #94820, opened 2020-09-16
verify-generated-swagger-docs: remove unnecessary build

### LINKED ISSUE #94691: Conversion generation isn't fully deterministic
When working on https://github.com/kubernetes/kubernetes/pull/94480 I faced a strange issue.

It seems that while auto-generating conversions it works fine,
but during things like `command: bash "verify-generated-swagger-docs.sh` or `make vet` in the presubmits what is generated is different.
Looking into test results:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/94480/pull-kubernetes-verify/1304065207634497536/build-log.txt
it seems that the order in which the files are processed is different then.

/kind bug
/sig api-machinery

roycaihw:
**What type of PR is this?**
/kind bug

As part of `make verify` (prow job `pull-kubernetes-verify`), `verify-generated-swagger-docs` gets invoked with a dirty directory. The make rule triggers `gen_conversion`, which doesn't work well with dirty dir and causes https://github.com/kubernetes/kubernetes/issues/94691. 

Removing the make, since the binary is unused (the [update script](https://github.com/kubernetes/kubernetes/blob/965137a992ed755474f17fa6377ee3555745addd/hack/update-generated-swagger-docs.sh#L39) installs the binary separatelly), and this way doesn't affect users who run this script stand-alone. 

Fixes #94691

```release-note
NONE
```

/cc @wojtek-t 
/sig api-machinery

### DIFF
--- a/hack/verify-generated-swagger-docs.sh
+++ b/hack/verify-generated-swagger-docs.sh
@@ -28,21 +28,6 @@ source "${KUBE_ROOT}/hack/lib/init.sh"
 
 kube::golang::setup_env
 
-make -C "${KUBE_ROOT}" WHAT=cmd/genswaggertypedocs
-
-# Find binary
-genswaggertypedocs=$(kube::util::find-binary "genswaggertypedocs")
-
-if [[ ! -x "$genswaggertypedocs" ]]; then
-  {
-    echo "It looks as if you don't have a compiled genswaggertypedocs binary"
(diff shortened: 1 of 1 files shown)


wojtek-t:
/lgtm
/approve
/milestone v1.20

Thanks!

PR #94821, opened 2020-09-16
Fix potential nil dereference in nodeinfo

yuzhiquan:
**What type of PR is this?**

/kind bug

`if err!=nil`, `nodeinfo` could be nil, `nodeInfo.Node()` may be panic.

Fixes #

```release-note
NONE
```

```docs

```


ahg-g:
`nodeinfo.Node()` will return `nil` if `nodeinfo` is `nil`:

ahg-g:
see here: https://github.com/kubernetes/kubernetes/blob/965137a992ed755474f17fa6377ee3555745addd/pkg/scheduler/framework/v1alpha1/types.go#L429

so this PR is not needed.

yuzhiquan:
> see here:
> 
> https://github.com/kubernetes/kubernetes/blob/965137a992ed755474f17fa6377ee3555745addd/pkg/scheduler/framework/v1alpha1/types.go#L429
> 
> so this PR is not needed.

Oops, only check `Get` function, not check the `Node` function, sorry for this.

PR #94822, opened 2020-09-16
Replace e2essh on e2e service tests

### LINKED ISSUE #94703: Replace e2essh on e2e service tests
There are a few e2e sig-network tests  that are using e2essh, limiting the number of environments where they can run.

https://github.com/kubernetes/kubernetes/blob/0fd10997dfee0f9ec2c413004bf93d310d76c84b/test/e2e/network/service.go#L2592-L2594

some of them can be easily replaced executing the commands on a pod, per example the ones that execute commands on kube-proxy can be done with:

```
func execKubeProxyCommand(cs clientset.Interface, cmd string) error {
	namespace := "kube-system"
	pods, err := cs.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{})
	if err != nil || len(pods.Items) == 0 {
		framework.Logf("failed to list kube-proxy pods in namespace: %s", namespace)
		return
	}
		if strings.Contains(pod.Name, "kube-proxy") {
			stdout, err := framework.RunHostCmd(namespace, pod.Name, cmd)
			if err != nil {
				framework.Logf("Failed to exec command on node %s: %v", pod.Spec.NodeName, err)
				return err
			}
			framework.Logf("command outoput on node %s: %s", pod.Spec.NodeName, stdout)
		}
	}
        return nil
}
```

JornShen:
**What type of PR is this?**

/kind bug
/kind cleanup
 
- Backgroud

There are a few e2e sig-network tests that are using e2essh, limiting the number of environments where they can run.
some of them can be easily replaced executing the commands on a pod. This PR can reach that e2e network service test can execute whever hosts have ssh or not.

ref: https://github.com/kubernetes/kubernetes/issues/94703#issue-698508813

- What's the changes 

There are three tests  in network service suits of e2e relevant to it.

> ginkgo.It("should be able to up and down services", ...)
> ginkgo.It("should implement service.kubernetes.io/service-proxy-name",  ...)
> ginkgo.It("should implement service.kubernetes.io/headless", ...)

change:  replace ssh host to exec host network pod to verify the service up

- Running Result
```
{"msg":"PASSED [sig-network] Services should implement service.kubernetes.io/service-proxy-name","total":1,"completed":1,"skipped":3969,"failed":0}

{"msg":"PASSED [sig-network] Services should be able to up and down services","total":1,"completed":1,"skipped":818,"failed":0}

{"msg":"PASSED [sig-network] Services should implement service.kubernetes.io/headless","total":1,"completed":1,"skipped":3288,"failed":0}

```

Fixes https://github.com/kubernetes/kubernetes/issues/94703

**release-note**
```release-note
NONE
```

### DIFF
--- a/test/e2e/network/service.go
+++ b/test/e2e/network/service.go
@@ -303,56 +303,44 @@ func StopServeHostnameService(clientset clientset.Interface, ns, name string) er
 }
 
 // verifyServeHostnameServiceUp wgets the given serviceIP:servicePort from the
-// given host and from within a pod. The host is expected to be an SSH-able node
-// in the cluster. Each pod in the service is expected to echo its name. These
-// names are compared with the given expectedPods list after a sort | uniq.
-func verifyServeHostnameServiceUp(c clientset.Interface, ns, host string, expectedPods []string, serviceIP string, servicePort int) error {
-	execPod := e2epod.CreateExecPodOrFail(c, ns, "execpod-", nil)
+// the host exec pod of host network type and from the exec pod of container network type.
+// Each pod in the service is expected to echo its name. These names are compared with the
+// given expectedPods list after a sort | uniq.
+func verifyServeHostnameServiceUp(c clientset.Interface, ns string, expectedPods []string, serviceIP string, servicePort int) error {
+	// to verify from host network
+	hostExecPod := launchHostExecPod(c, ns, "verify-service-up-host-exec-pod")
+
+	// to verify from container's network
+	execPod := e2epod.CreateExecPodOrFail(c, ns, "verify-service-up-exec-pod-", nil)
 	defer func() {
+		e2epod.DeletePodOrFail(c, ns, hostExecPod.Name)
 		e2epod.DeletePodOrFail(c, ns, execPod.Name)
 	}()
 
-	// Loop a bunch of times - the proxy is randomized, so we want a good
-	// chance of hitting each backend at least once.
-	buildCommand := func(wget string) string {
+	// verify service from pod
+	cmdFunc := func(podName string) string {
+		wgetCmd := "wget -q -T 1 -O -"
 		serviceIPPort := net.JoinHostPort(serviceIP, strconv.Itoa(servicePort))
-		return fmt.Sprintf("for i in $(seq 1 %d); do %s http://%s 2>&1 || true; echo; done",
-			50*len(expectedPods), wget, serviceIPPort)
-	}
-	commands := []func() string{
-		// verify service from node
-		func() string {
-			cmd := "set -e; " + buildCommand("wget -q --timeout=0.2 --tries=1 -O -")
-			framework.Logf("Executing cmd %q on host %v", cmd, host)
-			result, err := e2essh.SSH(cmd, host, framework.TestContext.Provider)
-			if err != nil || result.Code != 0 {
-				e2essh.LogResult(result)
-				framework.Logf("error while SSH-ing to node: %v", err)
-			}
-			return result.Stdout
-		},
-		// verify service from pod
-		func() string {
-			cmd := buildCommand("wget -q -T 1 -O -")
-			framework.Logf("Executing cmd %q in pod %v/%v", cmd, ns, execPod.Name)
-			// TODO: Use exec-over-http via the netexec pod instead of kubectl exec.
-			output, err := framework.RunHostCmd(ns, execPod.Name, cmd)
-			if err != nil {
-				framework.Logf("error while kubectl execing %q in pod %v/%v: %v\nOutput: %v", cmd, ns, execPod.Name, err, output)
-			}
-			return output
-		},
+		cmd := fmt.Sprintf("for i in $(seq 1 %d); do %s http://%s 2>&1 || true; echo; done",
+			50*len(expectedPods), wgetCmd, serviceIPPort)
+		framework.Logf("Executing cmd %q in pod %v/%v", cmd, ns, podName)
+		// TODO: Use exec-over-http via the netexec pod instead of kubectl exec.
+		output, err := framework.RunHostCmd(ns, podName, cmd)
+		if err != nil {
+			framework.Logf("error while kubectl execing %q in pod %v/%v: %v\nOutput: %v", cmd, ns, podName, err, output)
+		}
+		return output
 	}
 
 	expectedEndpoints := sets.NewString(expectedPods...)
 	ginkgo.By(fmt.Sprintf("verifying service has %d reachable backends", len(expectedPods)))
-	for _, cmdFunc := range commands {
+	for _, podName := range []string{hostExecPod.Name, execPod.Name} {
 		passed := false
 		gotEndpoints := sets.NewString()
 
 		// Retry cmdFunc for a while
 		for start := time.Now(); time.Since(start) < e2eservice.KubeProxyLagTimeout; time.Sleep(5 * time.Second) {
-			for _, endpoint := range strings.Split(cmdFunc(), "\n") {
+			for _, endpoint := range strings.Split(cmdFunc(podName), "\n") {
 				trimmedEp := strings.TrimSpace(endpoint)
 				if trimmedEp != "" {
 					gotEndpoints.Insert(trimmedEp)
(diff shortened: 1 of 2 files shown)


aojea:
let's have buy in from sig-testing too
/assign @BenTheElder @spiffxp 
the improvement is those test run in KIND too now  ;)

test/e2e/network/service.go
```diff
@@ -1166,10 +1205,9 @@ var _ = SIGDescribe("Services", func() {
 	ginkgo.It("should work after restarting apiserver [Disruptive]", func() {
 		// TODO: use the ServiceTestJig here
 		e2eskipper.SkipUnlessProviderIs("gce", "gke")
```
aojea, line 1207: is this a gce test only?

JornShen: yes, same as test case ("should work after restarting kube-proxy [Disruptive]").  I try to remove this skip, but no success. Beacause I find that  next the test restart api-server or kube-proxy  will still judge  the env is gce or gke. 

https://github.com/kubernetes/kubernetes/blob/fcba5fe4e4eb1b1a2e2b14d2a68037fecd75e3ab/test/e2e/network/service.go#L712-L716

and

https://github.com/kubernetes/kubernetes/blob/fcba5fe4e4eb1b1a2e2b14d2a68037fecd75e3ab/test/e2e/network/service.go#L3802-L3807

aojea: and what happens if you remove the check for `ProviderIs` in those functions? does it work? can you add a temp commit to this PR removing it to see the result?
I mean, the test is tagged as disruptive, people can filter based on the tags, and the idea is to remove all the provider specific code ...

JornShen: All right. I think these two test case can be discussed about. Before I say that that no success, it was running in my local robot using local-cluster-up.sh to simulate a k8s cluster. But when I see the code,  I find that the code treat kube-apiserver and kube-proxy as container pod which is different from the process in my local env. So the test will failed in my env definitely.

test/e2e/network/service.go
```diff
@@ -1080,12 +1130,7 @@ var _ = SIGDescribe("Services", func() {
 		podNames2, svc2IP, err := StartServeHostnameService(cs, getServeHostnameService(svc2), ns, numPods)
 		framework.ExpectNoError(err, "failed to create replication controller with service: %s in the namespace: %s", svc2, ns)
 
-		hosts, err := e2essh.NodeSSHHosts(cs)
-		framework.ExpectNoError(err, "failed to find external/internal IPs for every node")
-		if len(hosts) == 0 {
-			framework.Failf("No ssh-able nodes")
-		}
-		host := hosts[0]
+		host := getSSHHost(cs)
```
Lion-Wei, line 1133: Move `checkHostSupportSSH` here would change the current situation that this case can only run in providers gce/gke/aws/local.

test/e2e/network/service.go
```diff
@@ -303,11 +303,11 @@ func StopServeHostnameService(clientset clientset.Interface, ns, name string) er
 }
 
 // verifyServeHostnameServiceUp wgets the given serviceIP:servicePort from the
-// given host and from within a pod. The host is expected to be an SSH-able node
-// in the cluster. Each pod in the service is expected to echo its name. These
-// names are compared with the given expectedPods list after a sort | uniq.
+// given host or from within a pod(if there are no host satisfied, we will only verify from pod).
+// The host is expected to be an SSH-able node in the cluster. Each pod in the service is expected to echo its name.
+// These names are compared with the given expectedPods list after a sort | uniq.
```
Lion-Wei, line 308: IMO, we still need check hoth host network and contianer network, our target is to use `framework.RunHostCmd` replace `e2essh.SSH`.

test/e2e/network/service.go
```diff
@@ -319,9 +319,11 @@ func verifyServeHostnameServiceUp(c clientset.Interface, ns, host string, expect
 		return fmt.Sprintf("for i in $(seq 1 %d); do %s http://%s 2>&1 || true; echo; done",
 			50*len(expectedPods), wget, serviceIPPort)
 	}
-	commands := []func() string{
+
+	var commands []func() string
+	if host != "" {
```
Lion-Wei, line 324: We still need verify host network connectivity, but this parameter is not necessary anymore, you can launch a pod with hostNetwork, and use `framework.RunHostCmd` to run command from this pod, reference:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/service.go#L2007

test/e2e/network/service.go
```diff
@@ -330,20 +332,21 @@ func verifyServeHostnameServiceUp(c clientset.Interface, ns, host string, expect
...
-			if err != nil {
-				framework.Logf("error while kubectl execing %q in pod %v/%v: %v\nOutput: %v", cmd, ns, execPod.Name, err, output)
-			}
-			return output
-		},
+		})
 	}
 
+	// verify service from pod
+	commands = append(commands, func() string {
```
Lion-Wei, line 339: IMO, the original build commands code would be better. 🤔

test/e2e/network/service.go
```diff
@@ -384,25 +387,49 @@ func verifyServeHostnameServiceUp(c clientset.Interface, ns, host string, expect
 }
 
 // verifyServeHostnameServiceDown verifies that the given service isn't served.
-func verifyServeHostnameServiceDown(c clientset.Interface, host string, serviceIP string, servicePort int) error {
+// Firstly it will verify from the host. If there are no host satisfied or verify from host no success,
+//it will verify from the exec pod then.
+func verifyServeHostnameServiceDown(c clientset.Interface, ns, host string, serviceIP string, servicePort int) error {
```
Lion-Wei, line 392: Same with `verifyServeHostnameServiceUp`

test/e2e/network/service.go
JornShen: I will remove it later. This case will not exist any host var.

test/e2e/network/service.go
JornShen: You are right. I should think it more carefully. 👍

test/e2e/network/service.go
JornShen: done.

test/e2e/network/service.go
JornShen: done.

test/e2e/network/service.go
```diff
@@ -3798,84 +3703,64 @@ func validateEndpointsPorts(c clientset.Interface, namespace, serviceName string
...
-			return fmt.Errorf("failed to restart apiserver: %v", err)
-		}
-		return waitForApiserverRestarted(cs, initialRestartCount)
-	}
-	// GKE doesn't allow ssh access, so use a same-version master
-	// upgrade to teardown/recreate master.
-	v, err := cs.Discovery().ServerVersion()
+// restartComponent restarts component static pod
+func restartComponent(cs clientset.Interface, cName, ns string, labelSet labels.Set) error {
+	pods, err := getComponentPod(cs, ns, labelSet)
```
aojea, line 3708: this is assuming that the apiserver or kubeproxy are pods, and that is not always true

aojea: however, seems the test was already considering that, right?

aojea: the `sshRestartMaster()` is not clear about that :/

test/e2e/network/service.go
```diff
@@ -83,6 +82,11 @@ const (
 	// AffinityConfirmCount is the number of needed continuous requests to confirm that
 	// affinity is enabled.
 	AffinityConfirmCount = 15
+
+	kubeProxyLabelName  = "kube-proxy"
+	kubeAPIServerLabelName = "kube-apiserver"
+	clusterComponentKey = "component"
+	clusterAddonLabelKey  = "k8s-app"
```
aojea, line 89: are these labels standard?
can you add a comment explaining the origin?

by the way, the go-fmt on this file, it seems need to be indented

test/e2e/network/service.go
aojea: also the `restartKubeProxy` kills a process

aojea:
@JornShen I have to apologize for this comment https://github.com/kubernetes/kubernetes/pull/94822#discussion_r489408728, I only wanted to check if it was feasible and simple to change those tests, but it seems that it made a lot of assumptions on how the kube-porxy and apiserver and deployed. Based on that I think we should forget about those `[Disruptive]` tests and the new logic about restarting service, the rest LGTM overall

test/e2e/network/service.go
JornShen: It seems that the case before thinking the api-server or kubelet as static pod. So they kill the pods to let kubelet restart the component (kube-proxy or apiserver). As you say, there is no doubt that apiserver or kube-proxy can be running as process. But if we want to restart them, who will restart them when we kill them or other way to stop them? I used to running the local-up-cluster.sh as they are process, no will start them when i kill them.  :）

test/e2e/network/service.go
JornShen: the const str "kube-apiserver"  and "component" is appear before. And I just set them as const var in this PR. For go-fmt I will fix it right now. And also the comment explaining.

test/e2e/network/service.go
aojea: yeah, I'm sorry, seems this is not a big win and overcomplicates everything too much

BenTheElder: An init system will restart them?

aojea: let's tackle the Disruptive tests in another PR, and have tese tests now working on KIND jobs
```
ginkgo.It("should be able to up and down services", ...)
ginkgo.It("should implement service.kubernetes.io/service-proxy-name", ...)
ginkgo.It("should implement service.kubernetes.io/headless", ...)
```

JornShen: All right. I will take annother PR to see more details about the change:
```
{"msg":"PASSED [sig-network] Services should work after restarting apiserver [Disruptive]","total":1,"completed":1,"skipped":2981,"failed":0}

{"msg":"PASSED [sig-network] Services should work after restarting kube-proxy [Disruptive]","total":1,"completed":1,"skipped":2061,"failed":0}
```

aojea:
/lgtm
thanks

Lion-Wei:
/lgtm
thanks

spiffxp (commented):
/hold
I just want to confirm this isn't adversely impacting conformance tests, need to read through

spiffxp:
/hold cancel
/lgtm
thanks!

PR #94828, opened 2020-09-16
Fix sctp job flakes

### LINKED ISSUE #92041: [Flaky test] gci-gce-basic-sctp should create a Pod with SCTP HostPort
**Which jobs are failing**:

https://k8s-testgrid.appspot.com/sig-network-gce#gci-gce-basic-sctp

sig-network] SCTP [Feature:SCTP] [LinuxOnly] should create a Pod with SCTP HostPort

Test [sig-network] SCTP [Feature:SCTP] [LinuxOnly] should create a Pod with SCTP HostPort has failed 3 times since it last passed at commit/d49d045a3. It started failing between commit/d49d045a3 (Tue Jun 09 13:25 PDT) and commit/d62762f09 (Wed Jun 10 01:26 PDT): https://github.com/kubernetes/kubernetes/compare/d49d045a3...d62762f09?.

**Testgrid link**:

https://k8s-testgrid.appspot.com/sig-network-gce#gci-gce-basic-sctp&show-stale-tests=

**Reason for failure**:

I'm not sure if we already did it but there was a proposal to change from dockershim to containerd for the GCE jobs

aojea:
We were only checking that the services or pods are created, but we should wait until the iptables rules are programmed by the corresponding components before checking.

Adding sleeps are not great but is the simpler solution and 5 seconds should be more than enough for kube-proxy or the host port manager to write the iptables rules.

It also adds support for ipv6 to the tests

/kind flake

Fixes: #92041 

```release-note
NONE
```

### DIFF
--- a/test/e2e/network/service.go
+++ b/test/e2e/network/service.go
@@ -4001,6 +4001,9 @@ var _ = SIGDescribe("SCTP [Feature:SCTP] [LinuxOnly]", func() {
 
 		ginkgo.By("dumping iptables rules on the node")
 		cmd = "sudo iptables-save"
+		if framework.TestContext.ClusterIsIPv6() {
+			cmd = "sudo ip6tables-save"
+		}
 		framework.Logf("Executing cmd %q on node %v", cmd, node.Name)
 		result, err := hostExec.IssueCommandWithResult(cmd, node)
 		if err != nil {
@@ -4061,6 +4064,9 @@ var _ = SIGDescribe("SCTP [Feature:SCTP] [LinuxOnly]", func() {
 		node, err := e2enode.GetRandomReadySchedulableNode(cs)
 		framework.ExpectNoError(err)
 		cmd := "sudo iptables-save"
+		if framework.TestContext.ClusterIsIPv6() {
+			cmd = "sudo ip6tables-save"
+		}
 		framework.Logf("Executing cmd %q on node %v", cmd, node.Name)
 		result, err := hostExec.IssueCommandWithResult(cmd, node)
 		if err != nil {
--- a/test/e2e/network/service.go
+++ b/test/e2e/network/service.go
@@ -3998,27 +3998,28 @@ var _ = SIGDescribe("SCTP [Feature:SCTP] [LinuxOnly]", func() {
 			err := cs.CoreV1().Pods(f.Namespace.Name).Delete(context.TODO(), podName, metav1.DeleteOptions{})
 			framework.ExpectNoError(err, "failed to delete pod: %s in namespace: %s", podName, f.Namespace.Name)
 		}()
-
-		ginkgo.By("dumping iptables rules on the node")
+		// wait until host port manager syncs rules
 		cmd = "sudo iptables-save"
 		if framework.TestContext.ClusterIsIPv6() {
 			cmd = "sudo ip6tables-save"
 		}
-		framework.Logf("Executing cmd %q on node %v", cmd, node.Name)
-		result, err := hostExec.IssueCommandWithResult(cmd, node)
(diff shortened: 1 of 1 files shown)


test/e2e/network/service.go
```diff
@@ -3998,9 +3998,15 @@ var _ = SIGDescribe("SCTP [Feature:SCTP] [LinuxOnly]", func() {
 			err := cs.CoreV1().Pods(f.Namespace.Name).Delete(context.TODO(), podName, metav1.DeleteOptions{})
 			framework.ExpectNoError(err, "failed to delete pod: %s in namespace: %s", podName, f.Namespace.Name)
 		}()
+		// wait until host port manager syncs rules
+		// TODO: make this more reliable
+		time.Sleep(5 * time.Second)
```
danwinship, line 4003: We just need to wait for the pod to become Ready; that implies that the CNI plugin has done its work, which implies that the hostports should be set up. There's some framework method that can be used for this.

test/e2e/network/service.go
```diff
@@ -4054,13 +4060,18 @@ var _ = SIGDescribe("SCTP [Feature:SCTP] [LinuxOnly]", func() {
 
 		err = e2enetwork.WaitForService(f.ClientSet, ns, serviceName, true, 5*time.Second, e2eservice.TestTimeout)
 		framework.ExpectNoError(err, fmt.Sprintf("error while waiting for service:%s err: %v", serviceName, err))
-
+		// wait until kube-proxy syncs rules
+		// TODO: make this more reliable
+		time.Sleep(5 * time.Second)
```
danwinship, line 4065: `e2enetwork.WaitForService` waits for the Service to be created, but we should wait for the _Endpoints_ to be created (which in turn implies that both the Service and the Pod have been created). There's a method for that. After that, kube-proxy is allowed to take up to `e2eservice.KubeProxyEndpointLagTimeout` to update iptables, but you should poll rather than just sleeping that whole time. There should be examples around somewhere.

aojea: I thought the same, but IIUIC this test does not create endpoints, it just checks for the service reject rule.
I've checked locally with udp service without pods associated
```
-A KUBE-SERVICES -d 10.96.57.5/32 -p udp -m comment --comment "default/my-service: has no endpoints" -m udp --dport 80 -j REJECT --reject-with icmp-port-unreachable
```

test/e2e/network/service.go
aojea: CreateSync() already waits it is running 

https://github.com/kubernetes/kubernetes/blob/3f585835d0a38462bb928454245ec01e834e53a0/test/e2e/framework/pods.go#L99-L108

danwinship: If the pod has an IP then shouldn't it be guaranteed that the hostport manager already ran?

test/e2e/network/service.go
danwinship: ok, in that case it should be `e2eservice.KubeProxyLagTimeout` (which is weirdly high and should probably be fixed, but anyway, definitely poll rather than sleep)

test/e2e/network/service.go
aojea: maybe lagging programming the iptables rules?
let me add an active polling for the iptables check

test/e2e/network/service.go
aojea: polling is the right way, I was a bit lazy here, sorry

aojea:
/assign @MrHohn 
PTAL

MrHohn (approved):
The changes look good to me. Let's try this out.
/lgtm
/approve

PR #92249, opened 2020-06-18
Save container image when creating container, keep images in spec and…

loveRhythm1990:
… status consistent

**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Store image tag in container annotation and fetch it when call CRI interface ContainerStatus, to keep image in ContainerStatus and image in pod spec consistent when a image has multiple tags on a node.

**Which issue(s) this PR fixes**:

Fixes #51017

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
None
```

PR #94833, opened 2020-09-16
Store high level pod container image to runtime container annotations.

### LINKED ISSUE #51017: container statuses contain wrong/misleading image tag

/kind bug

**What happened**:

When multiple tags of an image share the same SHA256 digest the value status.containerStatuses[].image might return the wrong image tag. Unfortunately `docker inspect` doesn't include the tag that was used to create the container either so this might actually be a bug in docker?

status.containerStatuses[].image contains the same tag as specified in spec.containers[].image

 - Create an image named `test:staging` and re-tag it as `test:production`.
 - You now have 2 images with the same name and digest but different tags.
 - Start 2 pods: one using `test:staging` and the other `test:production` as image.
 - The container status for both pods will state that the same image tag was used to create the container which is at least misleading. Although it doesn't really make a difference as both image tags share the same digest it is technically wrong.

This should imho be considered for #46255.

**Environment**:
`Client Version: version.Info{Major:"1", Minor:"7", GitVersion:"v1.7.3", GitCommit:"2c2fe6e8278a5db2d15a013987b53968c743f2a1", GitTreeState:"clean", BuildDate:"2017-08-03T07:00:21Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"7", GitVersion:"v1.7.3", GitCommit:"2c2fe6e8278a5db2d15a013987b53968c743f2a1", GitTreeState:"clean", BuildDate:"2017-08-03T06:43:48Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}`
- Cloud provider or hardware configuration**: self-hosted cluster
- OS (e.g. from /etc/os-release): `Ubuntu 16.04.3 LTS (Xenial Xerus)`
- Kernel (e.g. `uname -a`): `Linux atgrzsl1751 4.10.0-32-generic #36~16.04.1-Ubuntu SMP Wed Aug 9 09:19:02 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`
- Install tools: custom ansible playbook
- Others: Docker version: `docker.io (1.12.6-0ubuntu1~16.04.1)`

whypro:
**What type of PR is this?**

/kind bug

Current container runtime implementations set the image in container status to the **first** image from inspect result. This can result in image mismatch when multiple image tags with same hash. (See also: https://github.com/kubernetes/kubernetes/issues/51017)

This PR sets/gets high level container image of a pod to/from runtime container annotations to solve this issue.

Fixes #51017

Thanks for @loveRhythm1990 's previous work (https://github.com/kubernetes/kubernetes/pull/92249) and also thanks for @hex108 's advise.

```release-note
Ensure image tag correctly in container statuses when two images have same hash.
```

```docs

```

### DIFF
--- a/pkg/kubelet/kuberuntime/kuberuntime_container.go
+++ b/pkg/kubelet/kuberuntime/kuberuntime_container.go
@@ -568,6 +568,9 @@ func toKubeContainerStatus(status *runtimeapi.ContainerStatus, runtimeName strin
 		State:        toKubeContainerState(status.State),
 		CreatedAt:    time.Unix(0, status.CreatedAt),
 	}
+	if annotatedInfo.Image != "" {
+		cStatus.Image = annotatedInfo.Image
+	}
 
 	if status.State != runtimeapi.ContainerState_CONTAINER_CREATED {
 		// If container is not in the created state, we have tried and
--- a/test/e2e/node/pods.go
+++ b/test/e2e/node/pods.go
@@ -40,6 +40,7 @@ import (
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/kubernetes/pkg/kubelet/events"
 	"k8s.io/kubernetes/test/e2e/framework"
+	e2ejob "k8s.io/kubernetes/test/e2e/framework/job"
 	e2ekubelet "k8s.io/kubernetes/test/e2e/framework/kubelet"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
 	imageutils "k8s.io/kubernetes/test/utils/image"
@@ -220,6 +221,62 @@ var _ = SIGDescribe("Pods Extended", func() {
 				podFastDeleteScenario{client: podClient.PodInterface, delayMs: 2000, initContainer: true},
 				podClient.PodInterface,
 			)
+		})
+
+		ginkgo.It("should be consistent between pod spec and status for container image", func(ctx context.Context) {
+			ginkgo.By("creating the pod")
+			name := "pod-inplace-update-" + string(uuid.NewUUID())
+			pod := &v1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: name,
+					Labels: map[string]string{
+						"name": name,
+					},
+				},
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						{
+							Name:  "busybox",
+							Image: imageutils.GetE2EImage(imageutils.Nginx),
+							Resources: v1.ResourceRequirements{
+								Requests: v1.ResourceList{
+									v1.ResourceCPU:    resource.MustParse("5m"),
+									v1.ResourceMemory: resource.MustParse("10Mi"),
+								},
+							},
+						},
+					},
+				},
+			}
+
+			ginkgo.By("submitting the pod to kubernetes")
+			podClient.Create(ctx, pod)
+
+			ginkgo.By("waiting for pod running")
+			err := e2epod.WaitForPodNameRunningInNamespace(ctx, f.ClientSet, pod.Name, f.Namespace.Name)
+			framework.ExpectNoError(err)
+
+			ginkgo.By("updating container image")
+			podClient.Update(ctx, name, func(pod *v1.Pod) {
+				pod.Spec.Containers[0].Image = imageutils.GetE2EImage(imageutils.NginxCopy)
+			})
+
+			ginkgo.By("waiting for pod restart and running")
+			err = e2epod.WaitForPodCondition(ctx, f.ClientSet, f.Namespace.Name, pod.Name, "inplace-updated", e2ejob.JobTimeout,
+				func(pod *v1.Pod) (bool, error) {
+					if pod.Status.ContainerStatuses[0].RestartCount < 1 {
+						return false, nil
+					}
+					if pod.Status.ContainerStatuses[0].Image != pod.Spec.Containers[0].Image {
+						return false, nil
+					}
+					if pod.Status.Phase != v1.PodRunning {
+						return false, nil
+					}
+					return true, nil
+				},
+			)
+			framework.ExpectNoError(err)
 		})
 	})
 
--- a/test/utils/image/manifest.go
+++ b/test/utils/image/manifest.go
@@ -188,6 +188,8 @@ const (
 	NFSProvisioner
 	// Nginx image
 	Nginx
+	// NginxCopy image
+	NginxCopy
 	// NginxNew image
 	NginxNew
 	// NodePerfNpbEp image
@@ -252,6 +254,7 @@ func initImageConfigs(list RegistryList) (map[ImageID]Config, map[ImageID]Config
 	configs[Nautilus] = Config{list.PromoterE2eRegistry, "nautilus", "1.7"}
 	configs[NFSProvisioner] = Config{list.SigStorageRegistry, "nfs-provisioner", "v3.0.1"}
 	configs[Nginx] = Config{list.PromoterE2eRegistry, "nginx", "1.14-4"}
+	configs[NginxCopy] = Config{list.PromoterE2eRegistry, "nginx", "1.14-4-copy"}
 	configs[NginxNew] = Config{list.PromoterE2eRegistry, "nginx", "1.15-4"}
 	configs[NodePerfNpbEp] = Config{list.PromoterE2eRegistry, "node-perf/npb-ep", "1.2"}
 	configs[NodePerfNpbIs] = Config{list.PromoterE2eRegistry, "node-perf/npb-is", "1.2"}
--- a/pkg/kubelet/kuberuntime/kuberuntime_container_test.go
+++ b/pkg/kubelet/kuberuntime/kuberuntime_container_test.go
@@ -224,6 +224,26 @@ func TestToKubeContainerStatus(t *testing.T) {
 				StartedAt: time.Unix(0, startedAt),
 			},
 		},
+		"right image from annotations": {
+			input: &runtimeapi.ContainerStatus{
+				Id:        cid.ID,
+				Metadata:  meta,
+				Image:     imageSpec,
+				State:     runtimeapi.ContainerState_CONTAINER_RUNNING,
+				CreatedAt: createdAt,
+				StartedAt: startedAt,
+				Annotations: map[string]string{
+					containerImageLabel: "fimage-alias",
+				},
+			},
+			expected: &kubecontainer.Status{
+				ID:        *cid,
+				Image:     "fimage-alias",
+				State:     kubecontainer.ContainerStateRunning,
+				CreatedAt: time.Unix(0, createdAt),
+				StartedAt: time.Unix(0, startedAt),
+			},
+		},
 	} {
 		actual := toKubeContainerStatus(test.input, cid.Type)
 		assert.Equal(t, test.expected, actual, desc)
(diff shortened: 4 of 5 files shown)


whypro:
cc @hex108  @zhouya0 
PTAL, thanks!

zhouya0 (approved):

zhouya0:
/lgtm
Thanks for the fix!

loveRhythm1990:
/lgtm
Thanks for the fix!, since I am busy with other things,  I will close my origin PR, hope this problem been fixed soon.

hex108:
/lgtm

Thanks!

whypro:
ping @Random-Liu @yujuhong . 
PTAL, thanks!

whypro:
cc @derekwaynecarr @sjenning

sjenning:
This seems reasonable since there is no way to disambiguate without storing this information from the spec (as far as I can tell).

@mrunalp what do you think?

sftim:
For the issue in https://github.com/kubernetes/kubernetes/issues/51017 I'd expect that the kubelet could disambiguate which container name and tag are used, based on the `.spec` of the Pod that the container belongs to.

https://github.com/kubernetes/kubernetes/issues/51017 implies that I'm wrong though. When can't the kubelet use the `.spec` to work out the right answer?

mrunalp:
Will review this tomorrow.

derekwaynecarr:
can we add a node e2e?

derekwaynecarr (changes requested):
I have no objection to the new label.

Can you add a node e2e test so we have something that verifies this does not regress?

whypro:
@derekwaynecarr OK. I'll add e2e tests later.

hex108:
Hi @whypro, is there any progess here?

ehashman:
/hold

please remove the hold once you've added tests to flag a reviewer's attention, thanks!

whypro:
@derekwaynecarr @hex108 @ehashman 

Sorry for the delay because I am busy with other things.

I have added the e2e test, but the test relies on an image of `nginx:1.14-1-copy`. I know from the [documentation](https://github.com/kubernetes/k8s.io/tree/main/k8s.gcr.io) that I need to upload it to `k8s.gcr.io/k8s-staging-e2e-test-images`, then promote to `k8s.gcr.io/e2e-test-images` repo.

But it seems that I didn't have push permission, would you like to tell me how to do it or help me push it?

claudiubelu:
I think the best approach would to to separate the bugfix from the test, so we can at least proceed with the bugfix and worry about the test later. It seems that the Image Promoter can't really handle 2 image tags having the same SHA.

claudiubelu:
> 
> 
> I think the best approach would to to separate the bugfix from the test, so we can at least proceed with the bugfix and worry about the test later. It seems that the Image Promoter can't really handle 2 image tags having the same SHA.

The ``nginx:1.14-1-copy`` image has been pushed and it has the same SHA as the ``nginx:1.14-1`` image (https://github.com/kubernetes/k8s.io/pull/2351). This only needs to be rebased.

whypro:
@claudiubelu Thanks very much for help! I've rebased the branch and it now is ready for review.
cc @derekwaynecarr @Random-Liu @yujuhong

ritpanjw:
Hi @dchen1107 👋 I'm checking in from the bug triage team for release 1.24. Is this PR targeted for release 1.24?

ritpanjw:
Hi @dchen1107 👋 I'm checking in from the bug triage team for release 1.24. we are very near to code freeze, any updated?

dims:
@derekwaynecarr @mrunalp 

Do we want this in 1.24?

pkg/kubelet/kuberuntime/kuberuntime_container.go
```diff
@@ -561,6 +561,9 @@ func toKubeContainerStatus(status *runtimeapi.ContainerStatus, runtimeName strin
 		State:        toKubeContainerState(status.State),
 		CreatedAt:    time.Unix(0, status.CreatedAt),
 	}
+	if annotatedInfo.Image != "" {
```
ehashman, line 564: This will always override the `status.Image.Image` above if the annotated info is present, is that what we want?

pacoxu: Before the new label is set, we keep using `status.Image.Image`.
If we can get from the label(info annotation), we will use it instead.
/lgtm

test/e2e/node/pods.go
```diff
@@ -473,6 +474,62 @@ var _ = SIGDescribe("Pods Extended", func() {
...
+			ginkgo.By("submitting the pod to kubernetes")
+			podClient.Create(pod)
+
+			ginkgo.By("waiting for pod running")
+			err := e2epod.WaitForPodNameRunningInNamespace(f.ClientSet, pod.Name, f.Namespace.Name)
+			framework.ExpectNoError(err)
+
+			ginkgo.By("updating container image")
+			podClient.Update(name, func(pod *v1.Pod) {
+				pod.Spec.Containers[0].Image = imageutils.GetE2EImage(imageutils.NginxCopy)
```
oomichi, line 513: Does this `NginxCopy` image have the same hash as the original `Nginx` image to reproduce the issue?

oomichi (commented):
/cc @oomichi

test/e2e/node/pods.go
```diff
@@ -473,6 +474,62 @@ var _ = SIGDescribe("Pods Extended", func() {
...
+
+			ginkgo.By("waiting for pod restart and running")
+			err = e2epod.WaitForPodCondition(f.ClientSet, f.Namespace.Name, pod.Name, "inplace-updated", e2ejob.JobTimeout,
+				func(pod *v1.Pod) (bool, error) {
+					if pod.Status.ContainerStatuses[0].RestartCount < 1 {
+						return false, nil
+					}
+					if pod.Status.ContainerStatuses[0].Image != pod.Spec.Containers[0].Image {
+						return false, nil
+					}
```
ehashman, line 524: I think it's possible for this to race. Won't there be a lag between the update of the container image and the propagation of the status with new image?

ehashman (commented):
This was last tested so long ago that there are no results in prow.

ehashman:
/milestone v1.25

Would like to see this land but I think we have other priorities right now for code freeze. Let's try to land this early in 1.25.

test/e2e/node/pods.go
claudiubelu: ``NginxCopy`` (``nginx-1.14.2-copy``) has the same SHA as the ``Nginx`` image (``nginx-1.14.2``), according to the SHAs promoted: https://github.com/kubernetes/k8s.io/blob/c011745da40d143dfbec89737901143cfed154e7/k8s.gcr.io/images/k8s-staging-e2e-test-images/images.yaml#L78

test/e2e/node/pods.go
whypro: Yes, there will be a lag. `WaitForPodCondition` will poll until two image are consistent.

whypro:
Forgive me for taking a long time to response.
@claudiubelu @helayoty @ehashman PTAL and review this PR again, thanks.

dims:
If you still need this PR then please rebase, if not, please close the PR

vinaykul:
> pull-kubernetes-e2e-inplace-pod-resize-containerd-main-v2

Please ignore this test. The failure is because InPlacePodVerticalScaling feature gate is not defined.
This test has now been limited to executing only on PR https://github.com/kubernetes/kubernetes/pull/102884 for the time being.

bart0sh:
@vinaykul 
> > pull-kubernetes-e2e-inplace-pod-resize-containerd-main-v2
> 
> Please ignore this test. The failure is because InPlacePodVerticalScaling feature gate is not defined. This test has now been limited to executing only on PR #102884 for the time being.

I saw this test failing in many unrelated PRs. It confuses contributors. Can we trigger it only for the code related to inplace pod vertical scaling?

sftim:
> **Does this PR introduce a user-facing change?**:
> 
> ```
> NONE
> ```

Is that right? I'd expect that end users would be able to notice that the change had been made.

whypro:
Thanks for notice. I've added it to release-note.

vinaykul:
> @vinaykul
> 
> > > pull-kubernetes-e2e-inplace-pod-resize-containerd-main-v2
> > 
> > 
> > Please ignore this test. The failure is because InPlacePodVerticalScaling feature gate is not defined. This test has now been limited to executing only on PR #102884 for the time being.
> 
> I saw this test failing in many unrelated PRs. It confuses contributors. Can we trigger it only for the code related to inplace pod vertical scaling?

@bart0sh I believe the fix is already in. @bobbypage submitted PR https://github.com/kubernetes/test-infra/pull/28229 that limits the trigger.

Job history shows the change is effective, https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-inplace-pod-resize-containerd-main-v2

A retest command though would trigger it even in unrelated PRs because a failed test history has been captured. Please use retest-required for now or ignore this particular test.

Once we merge the API change (hopefully soon) and the FG definition, this should start succeeding as a no-op.

vinaykul:
/skip pull-kubernetes-e2e-inplace-pod-resize-containerd-main-v2

whypro:
ping @derekwaynecarr . Could you review changes again?

SergeyKanzhelev:
/assign @ruiwen-zhao 

@ruiwen-zhao mostly FYI

test/e2e/node/pods.go
```diff
@@ -221,6 +222,62 @@ var _ = SIGDescribe("Pods Extended", func() {
...
+				pod.Spec.Containers[0].Image = imageutils.GetE2EImage(imageutils.NginxCopy)
+			})
+
+			ginkgo.By("waiting for pod restart and running")
+			err = e2epod.WaitForPodCondition(ctx, f.ClientSet, f.Namespace.Name, pod.Name, "inplace-updated", e2ejob.JobTimeout,
+				func(pod *v1.Pod) (bool, error) {
+					if pod.Status.ContainerStatuses[0].RestartCount < 1 {
+						return false, nil
+					}
+					if pod.Status.ContainerStatuses[0].Image != pod.Spec.Containers[0].Image {
```
SergeyKanzhelev, line 270: is the test always failing without this fix? does runtime uses already existing image in status?

test/utils/image/manifest.go
```diff
@@ -188,6 +188,8 @@ const (
 	NFSProvisioner
 	// Nginx image
 	Nginx
+	// NginxCopy image
```
SergeyKanzhelev, line 191: can you please extend the comment to make sure there it's clear we want the same SHA here, but different tag

SergeyKanzhelev:
Depending on what will be merged faster, the ContainerStatus docs will needs to be updated. Like this: https://github.com/kubernetes/kubernetes/pull/115463

SergeyKanzhelev:
/lgtm cancel

Cancelling to make sure the comment on nginx copy image is updated before the merge.

ruiwen-zhao:
One thing to call out here is that containerd prefixes some image names with `docker.io/library`.

Currently, without this change, if you deploy a container that runs `nginx:1.14.2`, then you will see  `docker.io/library/nginx:1.14.2` in the container status. But after this change, container status will have `nginx:1.14.2`. 

I don't think this is something we should be concerned about, but just calling out the change of behavior.

sftim:
> One thing to call out here is that containerd prefixes some image names with `docker.io/library`.
> 
> Currently, without this change, if you deploy a container that runs `nginx:1.14.2`, then you will see `docker.io/library/nginx:1.14.2` in the container status. But after this change, container status will have `nginx:1.14.2`.

Good spot. We shouldn't hard code or imply a presumption about how to canonicalize an image name.

PR #94835, opened 2020-09-16
upgrade test for BoundServiceAccountTokenVolume

zshihang:
**What type of PR is this?**

/kind feature

this PR adds an upgrade test for BoundServiceAccountTokenVolume feature to make sure pods will continue working after the disruptive upgrade.

```release-note
NONE
```

```
- [KEP]: https://github.com/kubernetes/enhancements/pull/1912
```


liggitt (commented):
first, thanks for working on this!

since we can't exercise this test in presubmit, and we don't actually have job configs that exercise an upgrade to master (e.g. 1.20.0-alpha...), I think doing this in 3 PRs would make sense:

1. a k/test-infra PR to create a job config that exercises upgrade of latest stable to master (e.g. 1.19 to 1.20-dev... not sure what the version markers are for that... ci/stable1 and ci/latest maybe?)
2. a kubernetes/kubernetes PR with the portions of this PR that fixup the upgrade test, so the existing tests/jobs pass to the point of exercising individual tests (once we have the job added in step 1 running, I'd like to see the fixes in step 2 cherry-picked back to 1.17-1.19 so we can start getting signal on those releases again)
3. a kubernetes/kubernetes PR with the portions of this PR that add the service account token test

cluster/gce/upgrade.sh
```diff
@@ -464,13 +464,8 @@ function update-coredns-config() {
...
     rm -rf "${download_dir}"
   }
-  trap cleanup EXIT
+  trap cleanup RETURN
 
-  # Get the new installed CoreDNS version
-  echo "Waiting for CoreDNS to update"
-  until [[ $("${KUBE_ROOT}"/cluster/kubectl.sh -n kube-system get deployment coredns -o=jsonpath='{$.metadata.resourceVersion}') -ne ${COREDNS_DEPLOY_RESOURCE_VERSION} ]]; do
-     sleep 1
-  done
```
liggitt, line 473: out of curiosity, how did you determine this was the issue? was there a log in the existing test job runs that pointed here? how did you run the upgrade test to verify dropping this let the upgrades start working again?

cluster/gce/upgrade.sh
```diff
@@ -464,13 +464,8 @@ function update-coredns-config() {
   cleanup() {
     rm -rf "${download_dir}"
   }
-  trap cleanup EXIT
+  trap cleanup RETURN
```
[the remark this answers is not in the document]
zshihang: the cleanup here is referencing a local variable `download_dir` which doesn't exist when EXIT.

cluster/gce/upgrade.sh
zshihang: > how did you determine this was the issue? was there a log in the existing test job runs that pointed here?

when i ran this test locally, the test stuck here forever and the fact is the `resourceVersion` doesn't change, i.e. the version `COREDNS_DEPLOY_RESOURCE_VERSION ` before upgrade == the resource version after upgrade. removing this let this upgrade test working again. i didn't verify other upgrades.

> how did you run the upgrade test to verify dropping this let the upgrades start working again?

i modified https://github.com/kubernetes/kubernetes/blob/73b4f09dce7dfa0f66910483f7ce06a87dc31108/test/e2e/cloud/gcp/cluster_upgrade.go#L51-L51 with default to ` "v1.20.0-alpha.0.1462+b14fcc544bd5a1-bazel` to get around the hash issue and then ran the command `kubetest --build --up --test --test_args="--ginkgo.focus=\[Feature:BoundServiceAccountTokenVolume\]" --check-version-skew=false`

liggitt: > i modified kubernetes/test/e2e/cloud/gcp/cluster_upgrade.go with default to "v1.20.0-alpha.0.1462+b14fcc544bd5a1-bazel to get around the hash issue

we need to resolve that in a way that permits the automated job runs to succeed... a fix for the hash issue needs to be part of the kubernetes/kubernetes PR with the upgrade test fix described in https://github.com/kubernetes/kubernetes/pull/94835#pullrequestreview-489995883

zshihang: yes, i planned to submit a PR to fix the hash issue.

test/e2e/upgrades/serviceaccount_admission_controller_migration.go
```diff
@@ -0,0 +1,105 @@
...
+
+	if pod.GetUID() != t.pod.GetUID() {
+		framework.Failf("Pod %q GetUID() = %q, want %q.", pod.Name, pod.GetUID(), t.pod.GetUID())
+	}
+	if pod.Status.ContainerStatuses[0].RestartCount != 0 {
+		framework.Failf("Pod %q RestartCount = %d, want 0.", pod.Name, pod.Status.ContainerStatuses[0].RestartCount)
+	}
+	if len(pod.Spec.Volumes) != 1 || pod.Spec.Volumes[0].Secret == nil {
+		framework.Failf("Pod %q Volumes[0].Secret = nil, want non-nil.", pod.Name)
+	}
```
liggitt, line 69: rather than checking for specific characteristics of the pod's service account token, I think we want to check instead that the pod's token (however it was injected) is still working properly... that means executing something in the context of the pod (like the agnhost in-cluster-client subcommand) and making sure it's not getting 401/403 errors

zshihang: done

test/e2e/upgrades/serviceaccount_admission_controller_migration.go
```diff
@@ -0,0 +1,135 @@
...
+func inClusterClientMustWork(f *framework.Framework, pod *v1.Pod) {
+	var logs string
+	if err := wait.Poll(1*time.Minute, 20*time.Minute, func() (done bool, err error) {
+		framework.Logf("Polling logs")
+		logs, err = e2epod.GetPodLogs(f.ClientSet, pod.Namespace, pod.Name, "inclusterclient")
+		if err != nil {
+			framework.Logf("Error pulling logs: %v", err)
+			return false, nil
+		}
+		_, err = e2eauth.ParseInClusterClientLogs(logs)
```
liggitt, line 99: if we have a single-server cluster, network request errors during the upgrade are expected. we specifically want to fail if this encounters an unauthorized error, right?

test/e2e/upgrades/serviceaccount_admission_controller_migration.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// Teardown cleans up any remaining resources.
+func (t *ServiceAccountAdmissionControllerMigrationTest) Teardown(f *framework.Framework) {
+	// rely on the namespace deletion to clean up everything
+}
+
+func inClusterClientMustWork(f *framework.Framework, pod *v1.Pod) {
+	var logs string
+	if err := wait.Poll(1*time.Minute, 20*time.Minute, func() (done bool, err error) {
```
liggitt, line 92: this will always wait 1 minute before checking the first time... should this use PollImmediate and/or a shorter interval?

test/e2e/upgrades/serviceaccount_admission_controller_migration.go
liggitt: I'm not sure the existing ParseInClusterClientLogs method is what we want here... we want to verify:
1. no unauthorized (401) errors are seen (ParseInClusterClientLogs fails on any error)
2. at least one successful authenticated request was made after a specific time, possibly using the SinceTime pod log option (ParseInClusterClientLogs scans all logs, which means we can't distinguish a successful request before and after the upgrade):
  * in setup, after the pod creation
  * after upgrade, after the API server restarts

test/e2e/upgrades/serviceaccount_admission_controller_migration.go
```diff
@@ -0,0 +1,135 @@
...
+	framework.ExpectNoError(err)
+	if podBeforeMigration.GetUID() != t.pod.GetUID() {
+		framework.Failf("Pod %q GetUID() = %q, want %q.", podBeforeMigration.Name, podBeforeMigration.GetUID(), t.pod.GetUID())
+	}
+	if podBeforeMigration.Status.ContainerStatuses[0].RestartCount != 0 {
+		framework.Failf("Pod %q RestartCount = %d, want 0.", podBeforeMigration.Name, podBeforeMigration.Status.ContainerStatuses[0].RestartCount)
+	}
+	if len(podBeforeMigration.Spec.Volumes) != 1 || podBeforeMigration.Spec.Volumes[0].Secret == nil {
+		framework.Failf("Pod %q Volumes[0].Secret = nil, want non-nil.", podBeforeMigration.Name)
+	}
```
liggitt, line 73: this check will fail if the "from" version this test suite is run with already had the bound tokens enabled... we want this test to be robust into the future

liggitt: I also expected a call to `inClusterClientMustWork(f, podBeforeMigration)` here, passing in the time of the upgrade so we could verify correctness of the pod after upgrade

test/e2e/upgrades/serviceaccount_admission_controller_migration.go
zshihang: yes, there are bunch of network errors when testing `pod-before-migration` after upgrade as expected. to only test on 401/403 we probably need to change the inclusterclient to report the HTTP code of err in the log and rebuild the image. the stronger assertion that there is no failure since upgrade is sufficient right? the only issue is that it might fail for other reasons as well. WDYT?

test/e2e/upgrades/serviceaccount_admission_controller_migration.go
zshihang: since i was not checking whether we have seen tokens in log, PollImmediate might only execute once if there is no matched logs.

i changed to PollImmediate and checking that we at least see one token in the log.

test/e2e/upgrades/serviceaccount_admission_controller_migration.go
zshihang: done

liggitt: inClusterClientMustWork looks good. drop the `if len(podBeforeMigration.Spec.Volumes) != 1 || podBeforeMigration.Spec.Volumes[0].Secret == nil {` check... the only thing we really care about is that the pod created before the upgrade still works after the upgrade

test/e2e/upgrades/serviceaccount_admission_controller_migration.go
```diff
@@ -0,0 +1,141 @@
...
+func createPod(f *framework.Framework, podName string) *v1.Pod {
+	pod := &v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podName,
+			Namespace: f.Namespace.Name,
+		},
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{{
+				Name:  "inclusterclient",
+				Image: imageutils.GetE2EImage(imageutils.Agnhost),
```
liggitt, line 125: add a shorter poll period with `"--poll-interval=..."` (like 5 or 10... default is 30) so we don't have to wait so long to pick up logs

test/e2e/upgrades/serviceaccount_admission_controller_migration.go
```diff
@@ -0,0 +1,141 @@
...
+		if err != nil {
+			framework.Logf("Error pulling logs: %v", err)
+			return false, nil
+		}
+		numTokens, err := e2eauth.ParseInClusterClientLogs(logs)
+		if err != nil {
+			return false, fmt.Errorf("inclusterclient reported an error: %v", err)
+		}
+		if numTokens == 0 {
+			return false, nil
```
liggitt, line 107: might want to log something about "no authenticated API calls found" so the test isn't silent here

test/e2e/upgrades/serviceaccount_admission_controller_migration.go
```diff
@@ -0,0 +1,141 @@
...
+	if err := wait.PollImmediate(15*time.Second, 5*time.Minute, func() (done bool, err error) {
+		framework.Logf("Polling logs")
+		logs, err = e2epod.GetPodLogsSince(f.ClientSet, pod.Namespace, pod.Name, "inclusterclient", since)
+		if err != nil {
+			framework.Logf("Error pulling logs: %v", err)
+			return false, nil
+		}
+		numTokens, err := e2eauth.ParseInClusterClientLogs(logs)
+		if err != nil {
+			return false, fmt.Errorf("inclusterclient reported an error: %v", err)
```
liggitt, line 104: might want to write `logs` with framework.Logf before returning an error to help with debugging a test failure

liggitt (commented):
a few comments, but looks close to me

test/e2e/upgrades/serviceaccount_admission_controller_migration.go
zshihang: good point, done.

test/e2e/upgrades/serviceaccount_admission_controller_migration.go
zshihang: done

test/e2e/upgrades/serviceaccount_admission_controller_migration.go
zshihang: done

test/e2e/upgrades/serviceaccount_admission_controller_migration.go
zshihang: done

zshihang:
we need to cherrypick this test to 1.19 so that we can run upgrade test for 1.19-1.20. currently we can only do 1.20-1.20.

liggitt (commented):
/lgtm
/approve

> we need to cherrypick this test to 1.19 so that we can run upgrade test for 1.19-1.20. currently we can only do 1.20-1.20.

Is there a test job configured that exercises 1.19 -> 1.20?

PR #94838, opened 2020-09-16
[1.19][go1.15] Update to go1.15.2

justaugustus:
**What type of PR is this?**

/kind feature
/area dependency

(Cherry pick of https://github.com/kubernetes/kubernetes/pull/94449)

- Update to go1.15.2
- build: Update to k/repo-infra@v0.1.1 (supports go1.15.2)
- build: Use go-runner:buster-v2.0.1 (built using go1.15.1)
- bazel: Replace --features with Starlark build settings flag
- hack/lib/util.sh: some bash cleanups
  
  - switched one spot to use kube::logging
  - make kube::util::find-binary return an error when it doesn't find
    anything so that hack scripts fail fast instead of with '' binary not
    found errors.
  - this required deleting some genfeddoc stuff. the binary no longer
    exists in k/k repo since we removed federation/, and I don't see it
    in https://github.com/kubernetes-sigs/kubefed/ either. I'm assuming
    that it's gone for good now.

- bazel: output go_binary rule directly from go_binary_conditional_pure
  
  From: @mikedanese:
  Instead of aliasing. Aliases are annoying in a number of ways. This is
  specifically bugging me now because they make the action graph harder to
  analyze programmatically. By using aliases here, we would need to handle
  potentially aliased go_binary targets and dereference to the effective
  target.

  The comment references an issue with `pure = select(...)` which appears
  to be resolved considering this now builds.

- make kube::util::find-binary not dependent on bazel-out/ structure

  Implement an aspect that outputs go_build_mode metadata for go binaries,
  and use that during binary selection.

Tracking issue: https://github.com/kubernetes/release/issues/1517
ref: https://github.com/kubernetes/release/pull/1514, https://github.com/kubernetes/k8s.io/pull/1205, https://github.com/kubernetes/release/pull/1516

Fixes #

/hold I want to push an updated go-runner image with this as well

```release-note
- Kubernetes is now built using go1.15.2
- build: Update to k/repo-infra@v0.1.1 (supports go1.15.2)
- build: Use go-runner:buster-v2.0.1 (built using go1.15.1)
- bazel: Replace --features with Starlark build settings flag
- hack/lib/util.sh: some bash cleanups
  
  - switched one spot to use kube::logging
  - make kube::util::find-binary return an error when it doesn't find
    anything so that hack scripts fail fast instead of with '' binary not
    found errors.
  - this required deleting some genfeddoc stuff. the binary no longer
    exists in k/k repo since we removed federation/, and I don't see it
    in https://github.com/kubernetes-sigs/kubefed/ either. I'm assuming
    that it's gone for good now.

- bazel: output go_binary rule directly from go_binary_conditional_pure
  
  From: @mikedanese:
  Instead of aliasing. Aliases are annoying in a number of ways. This is
  specifically bugging me now because they make the action graph harder to
  analyze programmatically. By using aliases here, we would need to handle
  potentially aliased go_binary targets and dereference to the effective
  target.

  The comment references an issue with `pure = select(...)` which appears
  to be resolved considering this now builds.

- make kube::util::find-binary not dependent on bazel-out/ structure

  Implement an aspect that outputs go_build_mode metadata for go binaries,
  and use that during binary selection.
```

```docs

```


justaugustus:
/hold cancel
(I'll look at the go-runner update in a follow-up.)

hasheddan (approved):
/lgtm

Issue #94695, opened 2020-09-10
Improve Error handling in framework


alculquicondor:
Starting from golang 1.13, it's possible to wrap errors to be able to detect and compare root errors more easily.

In the scheduler framework, we only handled messages and used a status code to identify errors. This loses error information. #94437 is fixing this for a specific case, but we need to do a sweep through all the framework runtime and the plugins to **always** wrap errors.

Stretch:
It might also be worth introducing an error type for the framework that also holds the error code and implements `Is(target error) bool` and `Unwrap() error`

/sig scheduling

krishchow:
@alculquicondor unless you were planning on working on this, I would be happy to take a look at this issue and start working on a fix :)

alculquicondor:
Sounds good. Just wait for #94692 to get merged first :)

alculquicondor:
It is merged

alculquicondor:
Whenever possible, send small self-contained PRs (affecting a particular extension point and it's corresponding test).

arghya88:
@SataQiu I have submitted PRs for volume-zone, taint-toleration, service-affinity, podtoplogyspread, selectorspread plugins

Huang-Wei:
@alculquicondor could you help to negotiate to assign the tasks?

alculquicondor:
/remove-help

My bad for making this so open ended.

But as general guidance, please leave some time for the contributor already assigned to start the work. Specially given that we are so far from code freeze and this work is low priority.

In this particular case, I would like to give @krishchow and @SataQiu to work on this, as they were the first ones to offer help.

alculquicondor:
Also please try to coordinate in slack before sending PRs.

alculquicondor:
I closed some PRs that haven't been reviewed yet. I left the others open to avoid losing the work done.

alculquicondor:
@SataQiu would you mind putting a list of all the packages or plugins that already have open/merged PRs?

alculquicondor:
@SataQiu is there more pending work?

gavinfish:
@SataQiu @krishchow I am working on #94696 which has some dependencies on this issue, I will pick this up if you are working on this anymore.

krishchow:
Go ahead @gavinfish :)

PR #94850, opened 2020-09-17
Using structured logging in scheduler framework runtime

### LINKED ISSUE #94697: Make use of structured logging in framework runtime
/sig scheduling
/kind cleanup

In `pkg/scheduler/framework/runtime/framework.go`, we use `klog.Error` with a message that wraps an error:

https://github.com/kubernetes/kubernetes/blob/0fd10997dfee0f9ec2c413004bf93d310d76c84b/pkg/scheduler/framework/runtime/framework.go#L400-L401

Instead, we should be using structured logging, like so:

```golang
err := satus.AsError()
klog.ErrorS(err, "Failed running PreFilter plugin", "plugin", pl.Name(), "pod", klog.KObj(pod))
return framework.AsStatus(fmt.Errorf("on prefilter plugin %s: %w", pl.Name(), err)
```

SataQiu:
**What type of PR is this?**
/kind cleanup

Using structured logging in scheduler framework runtime

Fixes #94697

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/scheduler/framework/runtime/framework.go
+++ b/pkg/scheduler/framework/runtime/framework.go
@@ -273,7 +273,7 @@ func NewFramework(r Registry, plugins *config.Plugins, args []config.PluginConfi
 		}
 		p, err := factory(args, f)
 		if err != nil {
-			return nil, fmt.Errorf("error initializing plugin %q: %v", name, err)
+			return nil, fmt.Errorf("initializing plugin %q: %w", name, err)
 		}
 		pluginsMap[name] = p
 
@@ -397,9 +397,9 @@ func (f *frameworkImpl) RunPreFilterPlugins(ctx context.Context, state *framewor
 			if status.IsUnschedulable() {
 				return status
 			}
-			msg := fmt.Sprintf("prefilter plugin %q failed for pod %q: %v", pl.Name(), pod.Name, status.Message())
-			klog.Error(msg)
-			return framework.NewStatus(framework.Error, msg)
+			err := status.AsError()
+			klog.ErrorS(err, "Failed running PreFilter plugin", "plugin", pl.Name(), "pod", klog.KObj(pod))
+			return framework.AsStatus(fmt.Errorf("running PreFilter plugin %q: %w", pl.Name(), err))
 		}
 	}
 
@@ -432,10 +432,9 @@ func (f *frameworkImpl) RunPreFilterExtensionAddPod(
 		}
 		status = f.runPreFilterExtensionAddPod(ctx, pl, state, podToSchedule, podToAdd, nodeInfo)
 		if !status.IsSuccess() {
-			msg := fmt.Sprintf("error while running AddPod for plugin %q while scheduling pod %q: %v",
-				pl.Name(), podToSchedule.Name, status.Message())
-			klog.Error(msg)
-			return framework.NewStatus(framework.Error, msg)
+			err := status.AsError()
+			klog.ErrorS(err, "Failed running AddPod on PreFilter plugin", "plugin", pl.Name(), "pod", klog.KObj(podToSchedule))
+			return framework.AsStatus(fmt.Errorf("running AddPod on PreFilter plugin %q: %w", pl.Name(), err))
 		}
 	}
 
@@ -468,10 +467,9 @@ func (f *frameworkImpl) RunPreFilterExtensionRemovePod(
 		}
 		status = f.runPreFilterExtensionRemovePod(ctx, pl, state, podToSchedule, podToRemove, nodeInfo)
 		if !status.IsSuccess() {
-			msg := fmt.Sprintf("error while running RemovePod for plugin %q while scheduling pod %q: %v",
-				pl.Name(), podToSchedule.Name, status.Message())
-			klog.Error(msg)
-			return framework.NewStatus(framework.Error, msg)
+			err := status.AsError()
+			klog.ErrorS(err, "Failed running RemovePod on PreFilter plugin", "plugin", pl.Name(), "pod", klog.KObj(podToSchedule))
+			return framework.AsStatus(fmt.Errorf("running RemovePod on PreFilter plugin %q: %w", pl.Name(), err))
 		}
 	}
 
@@ -577,9 +575,9 @@ func (f *frameworkImpl) RunPreScorePlugins(
 	for _, pl := range f.preScorePlugins {
 		status = f.runPreScorePlugin(ctx, pl, state, pod, nodes)
 		if !status.IsSuccess() {
-			msg := fmt.Sprintf("error while running %q prescore plugin for pod %q: %v", pl.Name(), pod.Name, status.Message())
-			klog.Error(msg)
-			return framework.NewStatus(framework.Error, msg)
+			err := status.AsError()
+			klog.ErrorS(err, "Failed running PreScore plugin", "plugin", pl.Name(), "pod", klog.KObj(pod))
+			return framework.AsStatus(fmt.Errorf("running PreScore plugin %q: %w", pl.Name(), err))
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/scheduler/framework/runtime/framework.go
```diff
@@ -432,10 +432,9 @@ func (f *frameworkImpl) RunPreFilterExtensionAddPod(
 		}
 		status = f.runPreFilterExtensionAddPod(ctx, pl, state, podToSchedule, podToAdd, nodeInfo)
 		if !status.IsSuccess() {
-			msg := fmt.Sprintf("error while running AddPod for plugin %q while scheduling pod %q: %v",
-				pl.Name(), podToSchedule.Name, status.Message())
-			klog.Error(msg)
-			return framework.NewStatus(framework.Error, msg)
+			err := status.AsError()
+			klog.ErrorS(err, "Failed running AddPod extension for PreFilter plugin", "plugin", pl.Name(), "pod", klog.KObj(podToSchedule))
```
chendave, line 436: Is that accurate to say `AddPod` is an extension?

pkg/scheduler/framework/runtime/framework.go
```diff
@@ -397,9 +397,9 @@ func (f *frameworkImpl) RunPreFilterPlugins(ctx context.Context, state *framewor
 			if status.IsUnschedulable() {
 				return status
 			}
-			msg := fmt.Sprintf("prefilter plugin %q failed for pod %q: %v", pl.Name(), pod.Name, status.Message())
-			klog.Error(msg)
-			return framework.NewStatus(framework.Error, msg)
+			err := status.AsError()
+			klog.ErrorS(err, "Failed running PreFilter plugin", "plugin", pl.Name(), "pod", klog.KObj(pod))
+			return framework.AsStatus(fmt.Errorf("prefilter plugin %q failed for pod \"%v/%v\": %w", pl.Name(), pod.Namespace, pod.Name, err))
```
chendave, line 402: what's the point of coming up different error message for `klog` and `status`?

alculquicondor: The klog error is to be displayed immediately, thus, it should have all the information necessary for debugging.

The status is to be returned, and will be eventually displayed in a k8s event (like when you do `kubectl describe pod`). And this error might be concatenated with more context as it moves up.

Fix now: Do not include the pod name in the status, as it will be redundant up the stack.

For debate: Is this the correct place to produce the klog? My thinking is that, the closer to the actual error is, the easier it is to debug. But then we have some repeated code. The alternative is to klog one level above (in scheduler.go). But then the errors wouldn't show in other binaries, such as cluster autoscaler or the descheduler. Thoughts @damemi @MaciekPytel ?

pkg/scheduler/framework/runtime/framework.go
alculquicondor: It's ok: we are not saying "extension point". OTH, it doesn't add any useful information over `Failed running AddPod for PreFilter plugin`, so I prefer that.

pkg/scheduler/framework/runtime/framework.go
```diff
@@ -432,10 +432,9 @@ func (f *frameworkImpl) RunPreFilterExtensionAddPod(
 		}
 		status = f.runPreFilterExtensionAddPod(ctx, pl, state, podToSchedule, podToAdd, nodeInfo)
 		if !status.IsSuccess() {
-			msg := fmt.Sprintf("error while running AddPod for plugin %q while scheduling pod %q: %v",
-				pl.Name(), podToSchedule.Name, status.Message())
-			klog.Error(msg)
-			return framework.NewStatus(framework.Error, msg)
+			err := status.AsError()
+			klog.ErrorS(err, "Failed running AddPod extension for PreFilter plugin", "plugin", pl.Name(), "pod", klog.KObj(podToSchedule))
+			return framework.AsStatus(fmt.Errorf("error while running AddPod for plugin %q while scheduling pod \"%v/%v\": %w", pl.Name(), podToSchedule.Namespace, podToSchedule.Name, err))
```
alculquicondor, line 437: Do not use the word `error`, `failed`, etc, as we are already crafting an error, so it's redundant. And also no need to say the pod name, as said in the other thread.

Something like: `running AddPod on plugin %q: %w` is enough. Other information is to be added up or down in the call hierarchy, but please check that. As said earlier, we don't need the `podToSchedule` information, but we need the `podToAddd`.

Ultimately, this can be checked in the unit test.

alculquicondor (commented):
Fix unit tests :)
This will also help verify what's the full output.

pkg/scheduler/framework/runtime/framework.go
```diff
@@ -628,9 +626,8 @@ func (f *frameworkImpl) RunScorePlugins(ctx context.Context, state *framework.Cy
 		}
```
alculquicondor, line 626: above: `errCh.SendErrorWithCancel(status.AsError(), cancel)`.

However, we should also preserve the plugin name so we can add it to `ErrorS`. But we can leave this for a followup, as part of #94695.

pkg/scheduler/framework/runtime/framework.go
```diff
@@ -648,9 +645,8 @@ func (f *frameworkImpl) RunScorePlugins(ctx context.Context, state *framework.Cy
...
```
alculquicondor, line 645: Let's make the Errorf construction above consistent with `runScorePlugin`. Use `%w`, like in #94695.

pkg/scheduler/framework/runtime/framework.go
```diff
@@ -628,9 +626,8 @@ func (f *frameworkImpl) RunScorePlugins(ctx context.Context, state *framework.Cy
...
 	})
 	if err := errCh.ReceiveError(); err != nil {
-		msg := fmt.Sprintf("error while running score plugin for pod %q: %v", pod.Name, err)
```
alculquicondor, line 631: `plugins`, until we can pinpoint the plugin name.

damemi:
@alculquicondor I agree with your point in https://github.com/kubernetes/kubernetes/pull/94850#discussion_r490288038, having the error as close to where it occurs is best for consumers and I think it's up to those consumers to decide how they want to handle it to reduce redundancy (including scheduler.go)

pkg/scheduler/framework/runtime/framework.go
```diff
@@ -648,9 +645,8 @@ func (f *frameworkImpl) RunScorePlugins(ctx context.Context, state *framework.Cy
 		}
 	})
 	if err := errCh.ReceiveError(); err != nil {
-		msg := fmt.Sprintf("error while running normalize score plugin for pod %q: %v", pod.Name, err)
-		klog.Error(msg)
-		return nil, framework.NewStatus(framework.Error, msg)
+		klog.ErrorS(err, "Failed running Normalize Score plugin", "pod", klog.KObj(pod))
```
alculquicondor, line 648: `Normalize on Score plugins`

pkg/scheduler/framework/runtime/framework.go
```diff
@@ -743,9 +738,9 @@ func (f *frameworkImpl) RunBindPlugins(ctx context.Context, state *framework.Cyc
 			continue
 		}
 		if !status.IsSuccess() {
-			err := fmt.Errorf("plugin %q failed to bind pod \"%v/%v\": %w", bp.Name(), pod.Namespace, pod.Name, status.AsError())
-			klog.Error(err)
-			return framework.AsStatus(err)
+			err := status.AsError()
+			klog.ErrorS(err, "Failed running Bind plugin", "plugin", bp.Name(), "pod", klog.KObj(pod))
+			return framework.AsStatus(fmt.Errorf("running bind plugin %q: %w", bp.Name(), err))
```
alculquicondor, line 743: `running Bind plugins`. Capitalization applies for others too.

SataQiu:
@alculquicondor It looks like the unit tests have passed :)

pkg/scheduler/framework/runtime/framework.go
```diff
@@ -642,15 +639,14 @@ func (f *frameworkImpl) RunScorePlugins(ctx context.Context, state *framework.Cy
 		}
 		status := f.runScoreExtension(ctx, pl, state, pod, nodeScoreList)
 		if !status.IsSuccess() {
-			err := fmt.Errorf("normalize score plugin %q failed with error %v", pl.Name(), status.Message())
+			err := fmt.Errorf("normalize score plugin %q failed with error %w", pl.Name(), status.AsError())
```
alculquicondor, line 642: let's make this error construction consistent with the previous parallelize. I'll let you decide which pattern is best for now. But avoid saying `error` in an error message.

alculquicondor (commented):
/approve

/lgtm

For the future: do not amend commits so that it's easier to identify what the changes are since the last review. Once you PR is approved, the approver will ask you to amend.

Also I don't see any progress in #94695. Feel free to follow up with the contributor that volunteered to work on it and see if you can take it. If you want, of course :)

PR #94854, opened 2020-09-17
test(apply): deflake run_kubectl_apply_tests(round 3)

### LINKED ISSUE #93651: [Flaky Test] run_kubectl_apply_tests (ci-kubernetes-integration-master)
**Which jobs are flaking**:

`integration master (ci-kubernetes-integration-master)`

`test-cmd.run_kubectl_apply_tests `

**Testgrid link**:

https://testgrid.k8s.io/sig-release-master-blocking#integration-master

**Reason for failure**:

```
Error from server (NotFound): resources.mygroup.example.com "myobj" not found
```

/sig cli
/cc @kubernetes/ci-signal @soltysh 
/milestone v1.20
/prority important-soon

knight42:

**What type of PR is this?**

/kind flake

The message
```
Error from server (NotFound): resources.mygroup.example.com "myobj" not found
```
is misleading, it is actually expected and caused by https://github.com/kubernetes/kubernetes/blob/3eb90c19d0cf90b756c3e08e32c6495b91e0aeed/test/cmd/apply.sh#L159-L162 and https://github.com/kubernetes/kubernetes/blob/3eb90c19d0cf90b756c3e08e32c6495b91e0aeed/test/cmd/apply.sh#L471-L474

From the log https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-integration-master/1305183184433975296/build-log.txt, we could see that
```
[32mapply.sh:177: Successful get pods b -n nsb {{.metadata.name}}: b
(B[m!!! [0913 16:45:10] Call tree:
!!! [0913 16:45:10]  1: /home/prow/go/src/k8s.io/kubernetes/test/cmd/../../third_party/forked/shell2junit/sh2ju.sh:47 run_kubectl_apply_tests(...)
!!! [0913 16:45:10]  2: /home/prow/go/src/k8s.io/kubernetes/test/cmd/../../third_party/forked/shell2junit/sh2ju.sh:112 eVal(...)
!!! [0913 16:45:10]  3: /home/prow/go/src/k8s.io/kubernetes/test/cmd/legacy-script.sh:131 juLog(...)
!!! [0913 16:45:10]  4: /home/prow/go/src/k8s.io/kubernetes/test/cmd/legacy-script.sh:521 record_command(...)
!!! [0913 16:45:10]  5: hack/make-rules/test-cmd.sh:152 runTests(...)
+++ exit code: 1
+++ error: 1
Error when running run_kubectl_apply_tests
Recording: run_kubectl_server_side_apply_tests
Running command: run_kubectl_server_side_apply_tests
```

The last line run in the test is # 177, and we could take a look at the script https://github.com/kubernetes/kubernetes/blob/3eb90c19d0cf90b756c3e08e32c6495b91e0aeed/test/cmd/apply.sh#L177-L180
If line # 180 was run, it would print message to the log, but we didn't see it, so it is reasonable to suspect it is line # 179 
```sh
output_message=$(! kubectl get pods a -n nsb 2>&1 "${kube_flags[@]:?}")
```
that cause the test to fail.

Note that three options are set in the script:
```sh
set -o errexit
set -o nounset
set -o pipefail
```
and it is possible that pod `a` still exists(maybe in `Terminating` state) when line # 179 was executed in the test, so the test script exit at once.

A minimal reproducer:
```sh
$ cat test.sh
#!/bin/bash

set -o errexit
set -o nounset
set -o pipefail

succeed() {
    echo success
    return 0
}

fail() {
    echo failure
    return 1
}

out=$(! succeed)
echo $out
$ ./test.sh; echo $?
1
```

Fixes #93651

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/cmd/apply.sh
+++ b/test/cmd/apply.sh
@@ -159,7 +159,7 @@ __EOF__
   # Dry-run create the CR
   kubectl "${kube_flags[@]:?}" apply --dry-run=server -f hack/testdata/CRD/resource.yaml "${kube_flags[@]:?}"
   # Make sure that the CR doesn't exist
-  ! kubectl "${kube_flags[@]:?}" get resource/myobj || exit 1
+  ! kubectl "${kube_flags[@]:?}" get resource/myobj 2>/dev/null || exit 1
 
   # clean-up
   kubectl "${kube_flags[@]:?}" delete customresourcedefinition resources.mygroup.example.com
@@ -173,11 +173,8 @@ __EOF__
   kube::test::get_object_assert 'pods a -n nsb' "{{${id_field:?}}}" 'a'
   # apply b with namespace
   kubectl apply --namespace nsb --prune -l prune-group=true -f hack/testdata/prune/b.yaml "${kube_flags[@]:?}"
-  # check right pod exists
-  kube::test::get_object_assert 'pods b -n nsb' "{{${id_field:?}}}" 'b'
-  # check wrong pod doesn't exist
-  output_message=$(! kubectl get pods a -n nsb 2>&1 "${kube_flags[@]:?}")
-  kube::test::if_has_string "${output_message}" 'pods "a" not found'
+  # check right pod exists and wrong pod doesn't exist
+  kube::test::wait_object_assert 'pods -n nsb' "{{range.items}}{{${id_field:?}}}:{{end}}" 'b:'
 
   # cleanup
   kubectl delete pods b -n nsb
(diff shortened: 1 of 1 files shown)


test/cmd/apply.sh
```diff
@@ -176,8 +176,7 @@ __EOF__
   # check right pod exists
   kube::test::get_object_assert 'pods b -n nsb' "{{${id_field:?}}}" 'b'
```
liggitt, line 177: can drop these two lines if we're expecting only b to exist on the next line

test/cmd/apply.sh
```diff
@@ -191,8 +190,7 @@ __EOF__
   # check right pod exists
   kube::test::get_object_assert 'pods a' "{{${id_field:?}}}" 'a'
   # check wrong pod doesn't exist
-  output_message=$(! kubectl get pods b -n nsb 2>&1 "${kube_flags[@]:?}")
-  kube::test::if_has_string "${output_message}" 'pods "b" not found'
+  kube::test::wait_object_assert 'pods -n nsb' "{{range.items}}{{${id_field:?}}}:{{end}}" ''
```
liggitt, line 193: I'm confused why this was looking in namespace nsb... the apply on line 188 is not for that namespace

knight42: I think that's because https://github.com/kubernetes/kubernetes/blob/master/hack/testdata/prune/b.yaml specify the namespace, while https://github.com/kubernetes/kubernetes/blob/master/hack/testdata/prune/a.yaml not.

liggitt: but we're not applying prune/b.yaml between line 182 and here

knight42: Hmm I am not the author of this test, perhaps he wants to make sure pod b does not exist before applying b.yaml?

PR #94858, opened 2020-09-17
fix bugs when copying deployment annotations to replicaSet if value is empty

waynepeking348:
**What type of PR is this?**

/kind bug

In our production environment, we relay on annotations to identify deployment with certain types; in some cases, we may set the annotation value with empty string, which may lead to unexpected problems if a rollout action is trigged. 

1. deployment created with annotation `deployment.test.kubernetes.io: ""`; replicaSet1 will not copy this key from deployment since map (in Golang) will return empty string if the given key doesn't exist (just as this belong logic)
```
    if skipCopyAnnotation(k) || rs.Annotations[k] == v {
        continue
    }
```
2. upgrade deployment with a new image and then rollout the the previous version, then deployment is recreated from replicaSet1 (which doesn't have annotation `deployment.test.kubernetes.io`), then this deployment is unexpected because it lost the identification in its annotations.

Fixes #

```release-note
Copy annotations with empty value when deployment rolls back
```

### DIFF
--- a/pkg/controller/deployment/util/deployment_util.go
+++ b/pkg/controller/deployment/util/deployment_util.go
@@ -319,7 +319,7 @@ func copyDeploymentAnnotationsToReplicaSet(deployment *apps.Deployment, rs *apps
 		// newRS revision is updated automatically in getNewReplicaSet, and the deployment's revision number is then updated
 		// by copying its newRS revision number. We should not copy deployment's revision to its newRS, since the update of
 		// deployment revision number may fail (revision becomes stale) and the revision number in newRS is more reliable.
-		if skipCopyAnnotation(k) || rs.Annotations[k] == v {
+		if _, exist := rs.Annotations[k]; skipCopyAnnotation(k) || (exist && rs.Annotations[k] == v) {
 			continue
 		}
 		rs.Annotations[k] = v


cheftako:
The fix looks correct to me. However I'm a little worried about backward compatibility. Just because no one should have been relying on this behavior doesn't mean they aren't. At a minimum we should probably have a release note for this change. Interested to hear @deads2k thoughts.

waynepeking348:
> The fix looks correct to me. However I'm a little worried about backward compatibility. Just because no one should have been relying on this behavior doesn't mean they aren't. At a minimum we should probably have a release note for this change. Interested to hear @deads2k thoughts.

according to current logic, deployment will copy those annotations to the latest replicaSet, it won't affect others; I also think a release note is reasonable. would you please have a look at this fix @deads2k ?

waynepeking348:
@kow3ns  @deads2k @mortent could you please have a look up at this fix?

waynepeking348:
any thought /assign @cheftako

> > The fix looks correct to me. However I'm a little worried about backward compatibility. Just because no one should have been relying on this behavior doesn't mean they aren't. At a minimum we should probably have a release note for this change. Interested to hear @deads2k thoughts.
> 
> according to current logic, deployment will copy those annotations to the latest replicaSet, it won't affect others; I also think a release note is reasonable. would you please have a look at this fix @deads2k ?

any thought? /assign @cheftako

MikeSpreitzer:
@waynepeking348 , @cheftako , @deads2k : BTW, where is it documented that the `metadata.annotations` of a Deployment object are copied to the `metadata.annotations` of its owned ReplicaSet objects?  And, why is this a good idea?

Issue #94863, opened 2020-09-17
Graduate DefaultPodTopologySpread to Beta


alculquicondor:
kubernetes/enhancements#1258

/sig scheduling

**Why is this needed**:

It allows to replace SelectorSpread plugin with a configurable default for the PodTopologySpread plugin.

This is important not just to reduce maintenance burden, but also because the new plugin provides better spreading, both at node level and bigger topologies. Since the score considers all the pods in the cluster, it is able to maintain zone balance even at 5k-node clusters.

Performance benchmarks will be used to ensure scalability is not hurt for vanilla pods.

alculquicondor:
cc @ahg-g @Huang-Wei @wojtek-t @adtac

PR #94864, opened 2020-09-17
Tests for empty constraints array when DefaultPodTopologySpread is enabled

alculquicondor:
**What type of PR is this?**

/kind cleanup

An empty array in the configuration signifies that the user operator intends to disable the k8s provided default for TopologyPodSpread.

This is to preserve in TopologyPodSpread plugin the existing ability of disabling default spreading (disabling the SelectorSpread plugin).

Ref #94863

```release-note
NONE
```

```docs
- [KEP]: https://git.k8s.io/enhancements/keps/sig-scheduling/1258-default-even-pod-spreading
- [Usage]: https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/#internal-default-constraints
```


pkg/scheduler/apis/config/scheme/scheme_test.go
```diff
@@ -291,6 +297,73 @@ profiles:
...
+									{
+										MaxSkew:           3,
+										TopologyKey:       corev1.LabelHostname,
+										WhenUnsatisfiable: corev1.ScheduleAnyway,
+									},
+									{
+										MaxSkew:           5,
+										TopologyKey:       corev1.LabelZoneFailureDomainStable,
+										WhenUnsatisfiable: corev1.ScheduleAnyway,
+									},
```
Huang-Wei, line 337: nit: reuse the default value defined in `pkg/scheduler/apis/config/v1beta1/defaults.go`?

pkg/scheduler/apis/config/scheme/scheme_test.go
```diff
@@ -291,6 +297,73 @@ profiles:
...
+		{
+			name: "empty PodTopologySpread, feature DefaultPodTopologySpread enabled",
+			data: []byte(`
+apiVersion: kubescheduler.config.k8s.io/v1beta1
+kind: KubeSchedulerConfiguration
+profiles:
+- pluginConfig:
+  - name: PodTopologySpread
+    args:
+      defaultConstraints:
```
Huang-Wei, line 317: This works the same with:

```yaml
- name: PodTopologySpread
```

correct?

alculquicondor:
/assign @liggitt 

I remember we discussed before about distinguishing between nil and empty array. However, I think in this case the distinction is necessary and provides clarity.

pkg/scheduler/apis/config/scheme/scheme_test.go
alculquicondor: It does, but I wanted to be more explicit

pkg/scheduler/apis/config/scheme/scheme_test.go
alculquicondor: for the whole array? It's not exported, and I think it's best to keep it like that.

pkg/scheduler/apis/config/v1beta1/defaults_test.go
```diff
@@ -409,13 +409,23 @@ func TestPluginArgsDefaults(t *testing.T) {
...
+		{
+			name:    "PodTopologySpreadArgs empty array, DefaultPodTopologySpread feature enabled",
+			feature: features.DefaultPodTopologySpread,
+			in: &v1beta1.PodTopologySpreadArgs{
+				DefaultConstraints: []v1.TopologySpreadConstraint{},
+			},
+			want: &v1beta1.PodTopologySpreadArgs{
+				DefaultConstraints: []v1.TopologySpreadConstraint{},
+			},
+		},
```
soulxu, line 421: There can add a test for the DefaultPodTopologySpread feature disabled.

A question, why we add those default test to encode/decode tests at scheme_test.go, seems the tests in defaults_test.go should be enough.

alculquicondor: > There can add a test for the DefaultPodTopologySpread feature disabled.

That's the default, so it's already tested above.

> A question, why we add those default test to encode/decode tests at scheme_test.go, seems the tests in defaults_test.go should be enough.

Because tests in scheme_test.go are more e2e-like, ensuring that we didn't mistype function names and that our assumptions of what the api-machinery does align with what we expect (nil vs empty array).

alculquicondor:
ping @liggitt

alculquicondor:
:/ I should have realized that this was getting merged even without @liggitt approval

PTAL and I can revert if necessary.

liggitt:
if I'm reading this correctly, this PR doesn't actually change behavior, just enshrines the current behavior in tests?

> An empty array in the configuration signifies that the user operator intends to disable the k8s provided default for TopologyPodSpread.

making `nil` and `[]` behave differently is not recommended... the distinction isn't preserved in some serialization methods, and an empty list meaning "disable" is not obvious from inspection.

If the config specifies values for `defaultConstraints`, that overrides the k8s defaults, correct?

If the user wants to remove all defaultConstraints, wouldn't it be clearer to disable the plugin entirely?

alculquicondor:
> if I'm reading this correctly, this PR doesn't actually change behavior, just enshrines the current behavior in tests?

Correct. But the default value is alpha in 1.19.

> If the config specifies values for `defaultConstraints`, that overrides the k8s defaults, correct?

Correct

> If the user wants to remove all defaultConstraints, wouldn't it be clearer to disable the plugin entirely?

It would, but note that the plugin also serves [Pod Topology Spread Constratins](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints), which a Pod can set (and that overrides `defaultConstraints`)

liggitt:
> So we need a mechanism to just disable the default, and this looks, to me, like the best option.

can you describe the behavior of the plugin when all constraints are removed? what does it actually do?

in the past, we've recommended explicitly disabling the list (e.g. a single "none" value for a `[]string` type, or a peer boolean field for a complex type). that still seems preferable over `absent / null` and `[]` behaving differently

alculquicondor:
> can you describe the behavior of the plugin when all constraints are removed? what does it actually do?

It runs scores based on the constraints that a Pod specifies (in PodSpec). These can also be empty, in which case the plugin will do nothing.

> in the past, we've recommended explicitly disabling the list.

That actually sounds like a reasonable idea. We could have a value to signify disabled constraints (that Pods could also use).

Something like (but open to better suggestions, as I'm not convinced).

```yaml
defaultConstraints:
  - MaxSkew: -1
```

Note that a constraint looks like this:

```yaml
  - maxSkew: 3
    topologyKey: "kubernetes.io/hostname"
    whenUnsatisfiable: ScheduleAnyway
```

@Huang-Wei thoughts?

Huang-Wei:
I'm not a fan of `MaxSkew: -1` - which makes the semantic implicit and hard to maintain.

Two possible ways I can think of:

- Define `defaultConstraints` as type `*slice` - if it's nil, then defaulting to predefined constraints; if it's explicitly set (by user), no matter it's an empty slice or not, we do nothing. But I'm not that sure it's a good API pattern.
- Distinguish the top-level API field (of `defaultConstraints`) by telling it's **nil** or **zero value** - in our case, it's `PodTopologySpreadArgs`. If users don't customize  `PodTopologySpreadArgs`, I'd assume [getPluginArgsOrDefault()](https://github.com/kubernetes/kubernetes/blob/7614e331861e68c1b7578cdb45011d264a69067e/pkg/scheduler/framework/runtime/framework.go#L324) would return nil, and then we just need to tweak the logic in [getArgs()](https://github.com/kubernetes/kubernetes/blob/7614e331861e68c1b7578cdb45011d264a69067e/pkg/scheduler/framework/plugins/podtopologyspread/plugin.go#L88) to return a pointer instead of a zero value of Args.

alculquicondor:
I think suggestion 1 has the same problem.

I don't quite understand suggestion 2. But `getPluginArgsOrDefault` is just supposed to call the appropriate function in defaults.go, so we have to fix the implementation there to make it consistent with the rest of the plugins.

The only other option seems to be adding a boolean alongside `defaultConstraints`, perhaps called `disableDefaultConstraints`. But honestly... If an operator says `defaultConstraints: []` I find it confusing that we would override that.

But I would like a solution that would for `Pod.spec.topologySpreadConstraints` as well. Currently we use `len(constraints)` to see if we apply the default constraints.

Huang-Wei:
> I think suggestion 1 has the same problem.

Not sure I understand it, but after walking through some issues, this doesn't seem to be welcomed. So never mind.

> I don't quite understand suggestion 2. But getPluginArgsOrDefault is just supposed to call the appropriate function in defaults.go, so we have to fix the implementation there to make it consistent with the rest of the plugins.

Yes, with this option, existing logic needs to be tweaked to differentiate nil and zero-valued XYZArgs.

> The only other option seems to be adding a boolean alongside `defaultConstraints`, perhaps called `disableDefaultConstraints`

That's definitely an option that has the clearest semantics, but a bit verbose IMHO.

alculquicondor:
How do you feel about having

```yaml
defaultConstraints:
  - disabled: true
```

And then pod specs can use the same. In api validation, we would make sure that if disabled is true, it's the only value in the array.

liggitt:
adding a field to the item schema in order to represent disabling the entire list seems strange

```
disableDefaultConstraints: true
```
mutually exclusive with any items in defaultConstraints seems clearer

alculquicondor:
in the PodSpec, if we wanted to have the same knob (not sure if we do) we would have to add:

```yaml
disableDefaultTopologySpreadConstraints: true
```

in mutual exclusion with TopologySpreadConstraints, but we can leave that discussion for the KEP.

alculquicondor:
I won't be reverting this PR, but I'll update the test to use the suggested boolean.

Unless there are strong opinions against it @ahg-g?

liggitt:
> in the PodSpec

does this field and semantic (default nil differently than []) also exist in podspec?

alculquicondor:
No, it's treated the same. So we apply the default when `len()==0`

Huang-Wei:
> does this field and semantic (default nil differently than []) also exist in podspec?

Nope - nil or empty constraints from podspec both mean the user don't have a preference on spreading the pods.

liggitt:
> Nope - nil or empty constraints from podspec both mean the user don't have a preference on spreading the pods.

great, then I don't think any changes are needed there

alculquicondor:
The change is not needed, but it might be worth giving the Pods the knob to disable default spreading.

liggitt:
that seems pretty niche... I'd probably wait until the need for it was obvious

ahg-g:
sorry, late to the party, but we discussed this offline in the past and using an empty slice to disable default spreading was the preferred approach. We should update the comment to explain those semantics: https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/apis/config/types_pluginargs.go#L78

ahg-g:
I misunderstood the conclusion, using a boolean is also a good alternative that we already discussed, I was confused when you mentioned that the test will not be reverted and Jordan saying no changes is required (I guess he was referring to the pod api)

Huang-Wei:
> I guess he was referring to the pod api

Yes, pod.spec.topologySpreadConstraints keep its semantics as is.

> the test will not be reverted

I believe @alculquicondor meant to amend the tests by leveraging the new bool field.

alculquicondor:
Correct. Sorry for the confusion, I didn't want to merge this PR until we had agreement. But now that it is merged I'll just add the boolean field and make the test use them.

soulxu:
apologize for merging this PR unintend. I misunderstand this doc about only approver/reviewer can trigger the merge https://kubernetes.io/docs/contribute/review/for-approvers/#prow-commands-for-reviewing. Finally, I understand that the approved PR can be triggered by lgtm from here. I did saw alculquicondor 'at' liggit for a review, definitely no mean to merge the PR except just comment a normal lgtm. I will be careful next time.

Issue #90402, opened 2020-04-23
Make the "GetResourceRequest" logic in scheduler pkg more clean


fengzixu:
**> This issue is for tracking some problems I found and I plan to fix.**

**Why is this needed**:
During I fixed the bug [**bugfix: initcontainer is also taken into account when calculating resource requests**](https://github.com/kubernetes/kubernetes/pull/89222), I found some issues of calculating resource request logic existed in scheduler package.  

1. **Duplication**: Actually, in Kubernetes code base, only kube-scheduler gives the default resource request to every container for calculating score purpose. So we have our own logic. But there is a [unified function](https://github.com/kubernetes/kubernetes/blob/5713c22eecff4610026643fbd3d37c33a43c168d/pkg/api/v1/resource/helpers.go#L133) which also can be used to get the resource request of pod  in `/pkg/api/v1/resource` package. 

2. **Default resource request value**: In the [current implementation](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/noderesources/resource_allocation.go#L118), we check every container of one pod if it specifies the explicit resource request. If there is no explicit value, we will give a default resource request value to the **container**. Actually, we can change the default value of per container to per pod. It likes 
```go
resourceRequest := max(sum(pod.Spec.Containers), max(pod.Spec.InitContainers))
if  resourceRequest== 0 {
              resourceRequest = defaultResourceRequestValue
}
```
3. **Explicit zero resource request**:  In this [unit test](https://github.com/kubernetes/kubernetes/blob/3538320d74f98fc0972206d20cd3d76186d88aa9/pkg/scheduler/framework/plugins/noderesources/least_allocated_test.go#L100), it specifies the explicit zero resource request. It looks like this test doesn't want to get the default resource request. It should be considered carefully ****If we should give the default resource request value** for the pod which specifies the explicit zero value**. [Context](https://github.com/kubernetes/kubernetes/pull/89222#discussion_r411617331)

[...]

PR #94866, opened 2020-09-17
scheduler: Implement resource metrics at /metrics/resources

smarterclayton:
Implement one part of https://github.com/kubernetes/enhancements/issues/1748.

Implement pod resource metrics as described in [KEP 1916](https://github.com/kubernetes/enhancements/pull/1916). The new `/metrics/resources` endpoint is exposed on the active scheduler and reports `kube_pod_resources_...` metrics that present the effective requests and limits for all resources on the pods as calculated by the scheduler and kubelet. This allows administrators using the system to quickly perform resource consumption, reservation, and pending utilization calculations when those metrics are read. Because metrics calculation is on-demand, there is no additional resource consumption incurred by the scheduler unless the endpoint is scraped.

/kind feature

```release-note
A new set of alpha metrics are reported by the Kubernetes scheduler under the `/metrics/resources` endpoint that allow administrators to easily see the resource consumption (requests and limits for all resources on the pods) and compare it to actual pod usage or node capacity.
```

```docs
[KEP]: https://github.com/kubernetes/enhancements/pull/1916
```


pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,296 @@
...
+	for _, p := range pods {
+		_, reqs, _, limits, _, terminal := podRequestsAndLimitsByLifecycle(p, reuseReqs, reuseLimits)
+		if terminal {
+			// terminal pods are excluded from resource usage calculations
+			continue
+		}
+		for _, t := range []struct {
+			desc    resourceLifecycleDescriptors
+			total   v1.ResourceList
+			running v1.ResourceList
```
alculquicondor, line 121: it looks like you forgot to remove this :)

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,296 @@
...
+// containers of the pod. If PodOverhead feature is enabled, pod overhead is added to the
+// total container resource requests and to the total container limits which have a
+// non-zero quantity. If the pod is assigned to a node and non-terminal, the current requests and
+// limits are returned for the pod. The caller may avoid allocations of resource lists by passing
+// a requests and limits list to the function, which will be cleared before use.
+func podRequestsAndLimitsByLifecycle(pod *v1.Pod, reuseReqs, reuseLimits v1.ResourceList) (lifecycle string, reqs, currentReqs, limits, currentLimits v1.ResourceList, terminal bool) {
+	var initializing, running bool
+	switch {
+	case len(pod.Spec.NodeName) == 0:
+		lifecycle = "Pending"
```
alculquicondor, line 215: this is not necessary anymore, right?

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,296 @@
...
+	case pod.Status.Phase == v1.PodSucceeded, pod.Status.Phase == v1.PodFailed:
+		lifecycle = "Completed"
```
alculquicondor, line 217: metrics newbie question: don't we have to clear up anything in prometheus so that there is no extrapolation from the previous value of the metric?
Hopefully my question makes sense.

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,296 @@
...
+		return
+	}
+	ch <- metrics.NewLazyConstMetric(desc, metrics.GaugeValue,
+		val.AsApproximateFloat64(),
+		namespace, name, nodeName, schedulerName, priority, string(resourceName), unit,
+	)
+}
+
+// addResourceList adds the resources in newList to list
+func addResourceList(list, newList v1.ResourceList) {
```
alculquicondor, line 179: So this is reimplementing the logic again, which is against the purpose of putting this in the scheduler code.

Here is one implementation: https://github.com/kubernetes/kubernetes/blob/d9d626f1e67f2a6eb1d68ed06b265fa877ec5764/pkg/scheduler/framework/plugins/noderesources/fit.go#L148:6

And here is another:
https://github.com/kubernetes/kubernetes/blob/d9d626f1e67f2a6eb1d68ed06b265fa877ec5764/pkg/scheduler/framework/types.go#L601:6

I'm not sure if you will be able to use either, but at least at a TODO for the unification of the code.
Here is the cleanup bug that was not worked on 😞  #90402

pkg/scheduler/metrics/resources/resources.go
Huang-Wei: The informerFactory scheduler uses has excluded `Completed` Pods, so logic won't reach here. (under the hood, completed Pods, no matter it's Succeeded or Failed, would trigger a Delete event, then scheduler catches that and removes the pod from its internal cache, so thst completed pods don't occupy resource in scheduler's view)

So I think the logic related with "terminal" can be removed?

Another question: what's the point of returning "currentReq/currentLimits", I don't see they're used anywhere.

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+// podRequestsAndLimitsByLifecycle returns a dictionary of all defined resources summed up for all
+// containers of the pod. If PodOverhead feature is enabled, pod overhead is added to the
+// total container resource requests and to the total container limits which have a
+// non-zero quantity. If the pod is assigned to a node and non-terminal, the current requests and
+// limits are returned for the pod. The caller may avoid allocations of resource lists by passing
+// a requests and limits list to the function, which will be cleared before use.
+func podRequestsAndLimitsByLifecycle(pod *v1.Pod, reuseReqs, reuseLimits v1.ResourceList) (lifecycle string, reqs, currentReqs, limits, currentLimits v1.ResourceList, terminal bool) {
+	var initializing, running bool
```
Huang-Wei, line 212: In addition to these states, do you want to add a "terminating" state when `pod.DeletionTimestamp != ""`?

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,296 @@
...
+		reqs = make(v1.ResourceList, 4)
+	}
+	if reuseLimits != nil {
+		limits = reuseLimits
+		for k := range limits {
+			delete(limits, k)
+		}
+	} else {
+		limits = make(v1.ResourceList, 4)
+	}
```
Huang-Wei, line 248: nit: code can be de-duplicagted:

```go
reqs, limits := reuse(reuseReqs), reuse(reuseLimits)
```

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,296 @@
...
+			list[name] = quantity.DeepCopy()
+		} else {
+			value.Add(quantity)
+			list[name] = value
+		}
+	}
+}
+
+// maxResourceList sets list to the greater of list/newList for every resource
+// either list
```
Huang-Wei, line 191: Uncompleted comment..

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,296 @@
...
+func maxResourceList(list, new v1.ResourceList) {
+	for name, quantity := range new {
+		if value, ok := list[name]; !ok {
+			list[name] = quantity.DeepCopy()
+			continue
+		} else {
+			if quantity.Cmp(value) > 0 {
+				list[name] = quantity.DeepCopy()
+			}
+		}
```
Huang-Wei, line 201: nit: prefer the style:

```go
value, ok := list[name]
if !ok || quantity.Cmp(value) > 0 {
	list[name] = quantity.DeepCopy()
}
```

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+// addResourceList adds the resources in newList to list
+func addResourceList(list, newList v1.ResourceList) {
+	for name, quantity := range newList {
+		if value, ok := list[name]; !ok {
+			list[name] = quantity.DeepCopy()
+		} else {
+			value.Add(quantity)
+			list[name] = value
```
Huang-Wei, line 185: Just `list[name].Add(quantity)`?

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+}
+
+// podRequestsAndLimitsByLifecycle returns a dictionary of all defined resources summed up for all
+// containers of the pod. If PodOverhead feature is enabled, pod overhead is added to the
+// total container resource requests and to the total container limits which have a
+// non-zero quantity. If the pod is assigned to a node and non-terminal, the current requests and
+// limits are returned for the pod. The caller may avoid allocations of resource lists by passing
+// a requests and limits list to the function, which will be cleared before use.
+func podRequestsAndLimitsByLifecycle(pod *v1.Pod, reuseReqs, reuseLimits v1.ResourceList) (lifecycle string, reqs, currentReqs, limits, currentLimits v1.ResourceList, terminal bool) {
```
Huang-Wei, line 211: is "inUse" more descriptive than "current"?

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	// if PodOverhead feature is supported, add overhead for running a pod
+	// to the sum of reqeuests and to non-zero limits:
+	if pod.Spec.Overhead != nil {
+		addResourceList(reqs, pod.Spec.Overhead)
+		for name, quantity := range pod.Spec.Overhead {
+			if value, ok := limits[name]; ok && !value.IsZero() {
+				value.Add(quantity)
+				limits[name] = value
```
Huang-Wei, line 267: Again, directly use `limits[name].Add(quantity)`?

cmd/kube-scheduler/app/server.go
```diff
@@ -226,25 +243,23 @@ func buildHandlerChain(handler http.Handler, authn authenticator.Request, authz
...
-	//lint:ignore SA1019 See the Metrics Stability Migration KEP
-	defaultMetricsHandler := legacyregistry.Handler().ServeHTTP
-	pathRecorderMux.HandleFunc("/metrics", func(w http.ResponseWriter, req *http.Request) {
-		if req.Method == "DELETE" {
-			w.Header().Set("Content-Type", "text/plain; charset=utf-8")
-			w.Header().Set("X-Content-Type-Options", "nosniff")
-			io.WriteString(w, "metrics reset\n")
+	pathRecorderMux.Handle("/metrics", legacyregistry.HandlerWithReset())
+
+	resourceMetricsHandler := resources.Handler(informers.Core().V1().Pods().Lister())
```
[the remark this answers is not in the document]
alculquicondor: Right, but the scheduler informer adds some filters.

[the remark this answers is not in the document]
alculquicondor: Oh, got it. The informer factory itself is overwritten.

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,296 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package resources provides a metrics collector that reports the
+// resource consumption (requests and limits) of the pods in the cluster
+// as the scheduler and kubelet would interpret it.
```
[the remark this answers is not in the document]
smarterclayton: True but the kubelet applies the same rules to resources to static and non static pods.  I think in this case the comment is about the resource model, and the kubelet and scheduler as high level components.  There are definitely special case behaviors all over both of those components but they are special cases.

staging/src/k8s.io/apimachinery/pkg/api/resource/quantity_test.go
```diff
@@ -1177,6 +1179,183 @@ func TestNegateRoundTrip(t *testing.T) {
...
+		// very large numbers
+		{Quantity{d: maxAllowed, Format: DecimalSI}, math.MaxInt64},
+		{Quantity{d: maxAllowed, Format: BinarySI}, math.MaxInt64},
+		{decQuantity(12, 18, DecimalSI), 1.2e19},
+
+		// infinities caused due to float64 overflow
+		{decQuantity(12, 500, DecimalSI), math.Inf(0)},
+		{decQuantity(-12, 500, DecimalSI), math.Inf(-1)},
+
+		// // We'll accept fractional binary stuff, too.
```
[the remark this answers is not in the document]
smarterclayton: Copy and paste.  I'll remove them, they don't actually stress the float part.

cmd/kube-scheduler/app/server.go
```diff
@@ -226,25 +243,23 @@ func buildHandlerChain(handler http.Handler, authn authenticator.Request, authz
...
+	pathRecorderMux.HandleFunc("/metrics/resources", func(w http.ResponseWriter, req *http.Request) {
+		if !isLeader() {
```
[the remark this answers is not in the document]
smarterclayton: It means the leader will return no metrics, but the default behavior of something like kube-prometheus is just to scrape all three and it will only see metrics from the current leader.  In practice this is actually pretty common since some metrics aren't registered until controllers start, and controllers don't start until they become leader (in kube-controller-maanger).

cmd/kube-scheduler/app/server.go
```diff
@@ -226,25 +243,23 @@ func buildHandlerChain(handler http.Handler, authn authenticator.Request, authz
...
+func installMetricHandler(pathRecorderMux *mux.PathRecorderMux, informers informers.SharedInformerFactory, isLeader func() bool) {
 	configz.InstallHandler(pathRecorderMux)
-	//lint:ignore SA1019 See the Metrics Stability Migration KEP
-	defaultMetricsHandler := legacyregistry.Handler().ServeHTTP
-	pathRecorderMux.HandleFunc("/metrics", func(w http.ResponseWriter, req *http.Request) {
-		if req.Method == "DELETE" {
-			w.Header().Set("Content-Type", "text/plain; charset=utf-8")
-			w.Header().Set("X-Content-Type-Options", "nosniff")
-			io.WriteString(w, "metrics reset\n")
+	pathRecorderMux.Handle("/metrics", legacyregistry.HandlerWithReset())
```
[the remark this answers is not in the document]
smarterclayton: Yes, HandlerWithReset implements the DELETE logic previously exposed here.

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,296 @@
...
+var _ metrics.StableCollector = &podResourceCollector{}
+
+// NewPodResourcesMetricsCollector registers a O(pods) cardinality metric that
+// reports the current resources requested by all pods on the cluster within
+// the Kubernetes resource model. Metrics are broken down by pod, node, resource,
+// and phase of lifecycle. Each pod returns two series per resource - one for
+// their aggregate usage (required to schedule) and one for their phase specific
+// usage. This allows admins to assess the cost per resource at different phases
+// of startup and compare to actual resource usage.
+func NewPodResourcesMetricsCollector(podLister corelisters.PodLister) metrics.StableCollector {
```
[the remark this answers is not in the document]
smarterclayton: Doesn't need to but in general we don't aggressively enforce "minimal public API" as a style rule in Kube, and I prefer the clarity of composition here.

pkg/scheduler/metrics/resources/resources.go
smarterclayton: Trivial helper reuse is not something I consider reimplementation.  Agree on some of the other types below.

pkg/scheduler/metrics/resources/resources.go
smarterclayton: I believe that would be illegal because it's not a pointer receiver.

pkg/scheduler/metrics/resources/resources.go
smarterclayton: > metrics newbie question: don't we have to clear up anything in prometheus so that there is no extrapolation from the previous value of the metric?

In the collector pattern no as used in this PR no - prometheus will scrape and then mark series it doesn't see any more as stale, and handle closing them out later.  In the static registered metrics pattern (whenever you see metrics.NewGauge / metrics.MustRegister) yes - that would be registering a go object in a map inside of the process and you'd have to periodically clear those out.

smarterclayton: > Another question: what's the point of returning "currentReq/currentLimits", I don't see they're used anywhere.

They are from the previous track and I will remove them

pkg/scheduler/metrics/resources/resources.go
smarterclayton: I'm not opposed to moving these to helpers though, although if we do helpers i would like them in staging rather than in k/k

alculquicondor: It might be trivial, but the duplication has already lead to bugs. Please add the TODO and link to cleanup issue.

pkg/scheduler/metrics/resources/resources.go
smarterclayton: Removed code

pkg/scheduler/metrics/resources/resources.go
smarterclayton: I need to verify that the scheduler logic is actually correct first and then would share the logic.

pkg/scheduler/metrics/resources/resources.go
smarterclayton: Within the context of the scheduler I agree on better reuse (your examples above).  I'm less supportive of reuse across broad domains (scheduler / kubelet) unless it's a standard API, and even then there has to be some caution because I've seen this mask other problems and changing that behavior then has client impacts.

smarterclayton:
> The informerFactory scheduler uses has excluded Completed Pods, so logic won't reach here. (under the hood, completed Pods, no matter it's Succeeded or Failed, would trigger a Delete event, then scheduler catches that and removes the pod from its internal cache, so thst completed pods don't occupy resource in scheduler's view). So I think the logic related with "terminal" can be removed?

This method should be agnostic of the details of the informer cache.  Note that the scheduler cache is currently holding on to pods that have terminated (according to kubelet) longer than it has to during deletion, and that behavior would not be implemented by the apiserver filters and thus not be in the cache, so future clarifications would be inconsistent.  Better to model that here and test it here, regardless of how cache behavior changes (and someone changing the cache may not be able to find all the impacted places).

The subsequent follow up issue I'll create would potentially unify that into a helper for v1 which is "PodTerminalForResourceConsiderations" (etc) and thus be usable in multiple places.

smarterclayton:
Updated with all feedback

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,237 @@
...
+// containers of the pod. If PodOverhead feature is enabled, pod overhead is added to the
+// total container resource requests and to the total container limits which have a
+// non-zero quantity. The caller may avoid allocations of resource lists by passing
+// a requests and limits list to the function, which will be cleared before use.
+// This method is the same as v1resource.PodRequestsAndLimits but avoids allocating in several
+// scenarios for efficiency.
+func podRequestsAndLimitsByLifecycle(pod *v1.Pod, reuseReqs, reuseLimits v1.ResourceList) (reqs, limits v1.ResourceList, terminal bool) {
+	switch {
+	case len(pod.Spec.NodeName) == 0:
+		// unscheduled pods cannot be terminal
```
alculquicondor, line 189: why have the check then?

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,237 @@
...
+		//   has the same logic
+	}
+	if terminal {
+		return
+	}
+
+	// attempt to reuse the maps if passed, or allocate otherwise
+	reqs, limits = reuseOrClearResourceList(reuseReqs), reuseOrClearResourceList(reuseLimits)
+
+	for _, container := range pod.Spec.Containers {
```
alculquicondor, line 202: Why not use `v1resource.PodRequestsAndLimits`?

sigh for yet another duplication I had forgotten about.

pkg/scheduler/metrics/resources/resources.go
smarterclayton: so that we fall through clearly

pkg/scheduler/metrics/resources/resources.go
smarterclayton: Because that allocates every time, and changing the public signature is expensive, which is why we generally don't over consolidate things like that.  I am not averse to adding a new model that allows reuse though in this case.

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,237 @@
...
+		v1resource.AddResourceList(limits, container.Resources.Limits)
+	}
+	// init containers define the minimum of any resource
+	for _, container := range pod.Spec.InitContainers {
+		v1resource.MaxResourceList(reqs, container.Resources.Requests)
+		v1resource.MaxResourceList(limits, container.Resources.Limits)
+	}
+
+	// if PodOverhead feature is supported, add overhead for running a pod
+	// to the sum of reqeuests and to non-zero limits:
```
Huang-Wei, line 213:
```suggestion
	// to the sum of requests and to non-zero limits:
```

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,237 @@
...
+	if pod.Spec.Overhead != nil {
```
Huang-Wei, line 214: Would be good to guard this additionally by feature gate: (also be consistent with other places)

https://github.com/kubernetes/kubernetes/blob/d41f79154251aaa54120015d8ad5b69319b44af8/pkg/scheduler/framework/types.go#L623-L624

pkg/scheduler/metrics/resources/resources.go
alculquicondor: I see.
What if you add a new function in v1resource that accepts already allocated maps and then you have this function and `v1resource.PodRequestsAndLimits` call that?

pkg/scheduler/metrics/resources/resources.go
alculquicondor: is it possible that an unscheduled Pod gets a Succeeded or Failed phase though?

pkg/scheduler/metrics/resources/resources.go
smarterclayton: Weird, I thought I copied that out of core library.  Will verify.

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,237 @@
...
+// a requests and limits list to the function, which will be cleared before use.
+// This method is the same as v1resource.PodRequestsAndLimits but avoids allocating in several
+// scenarios for efficiency.
+func podRequestsAndLimitsByLifecycle(pod *v1.Pod, reuseReqs, reuseLimits v1.ResourceList) (reqs, limits v1.ResourceList, terminal bool) {
+	switch {
+	case len(pod.Spec.NodeName) == 0:
+		// unscheduled pods cannot be terminal
+	case pod.Status.Phase == v1.PodSucceeded, pod.Status.Phase == v1.PodFailed:
+		terminal = true
+		// TODO: add a condition for deleted pods where all containers are terminated once scheduler
```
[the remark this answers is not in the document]
smarterclayton: I plan on creating one before merge once all review comments are addressed and updating the TODO

pkg/scheduler/metrics/resources/resources.go
alculquicondor: That's why we should have a single implementation :)

pkg/scheduler/metrics/resources/resources.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func (d resourceMetricsDescriptors) Describe(ch chan<- *metrics.Desc) {
+	d.requests.Describe(ch)
+	d.limits.Describe(ch)
+}
+
+var podResourceDesc = resourceMetricsDescriptors{
+	requests: resourceLifecycleDescriptors{
+		total: metrics.NewDesc("kube_pod_resource_request",
```
ehashman, line 56: I would prefer we prefix this with `scheduler_*` rather than `kube_*` since ultimately this metric is being generated by the scheduler.

`kube_*` is the standard prefix for KSM-generated/derived metrics, so as an end user, when trying to understand or find documentation for this metric I'd check those docs, and then would be unable to find it: https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md

ehashman (commented):
Apologies for not reviewing sooner, didn't get a ping on this one until yesterday.

pkg/scheduler/metrics/resources/resources.go
smarterclayton: The KEP created the name under the global namespace because these are intended to be global metrics - the location within the project was chosen as a secondary characteristic, so I don't think location should dominate where the metric is.  Part of the KEP is about standardizing discovery of metrics that reflect the project - if we change names to reflect location, I think that worsens the end administrator experience because if we decide to move these in the future we would break name compatibility. I agree in general that we lack a single unifying principle of metric naming that would have a space for what I would consider "API metrics" (as seen by #95839) - the metrics that the Kube project owns that represent a key connection between end user and project without exposing underlying component implementation.

I don't think KSM "owns" `kube_pod_` any more than Kubernetes the project does so it's well within our scope to see if we can identify a better precedent here.  I don't think `scheduler_` is correct because this is a metric that is broader than either component - the scheduler is just the best / simplest place to expose them today.  It's not `kubelet_`.  I worry that `pod_` lacks any identifying info, and we should preserve some prefix, hence `kube_pod_*`

pkg/scheduler/metrics/resources/resources.go
smarterclayton: Yeah that's what I was considering.  Worried about misuse, but i'd be ok sharing the impl now that we no longer need to track lifecycle here.

pkg/scheduler/metrics/resources/resources.go
smarterclayton: I don't think it was in the core library when i copied it :)

pkg/scheduler/metrics/resources/resources.go
smarterclayton: Another note too - these metrics are intended to replace `kube_pod_*` KSM metrics that are difficult and inefficient to use (which was the original feedback), so this will be replacing a bigger set of `kube_pod_*` KSM metrics with a more correct and easier to use `kube_pod_resources_*` metric.

Huang-Wei:
@smarterclayton In terms of the CI failures, somehow Github messed up the ref to a PR. I've observed it several times recently: https://kubernetes.slack.com/archives/C7J9RP96G/p1605128896118500

pkg/scheduler/metrics/resources/resources.go
alculquicondor: Please do

pkg/scheduler/metrics/resources/resources.go
smarterclayton: yes, if someone violates the API rules.  An extension can easily write pod status entries, and we shouldn't behave differently in that case (defensive).

pkg/scheduler/metrics/resources/resources.go
lilic: I don't see a problem with the kube_pod prefix, especially since these metrics will have different target labels, so both could be scrapped if a user wants to and you can always identify them by their target labels (e.g. job label being one of them). So this naming sounds good from my side, at least as kube-state-metrics maintainer. :)

As for metrics replacing the kube-state-metrics ones, agreed we reverted the changes we made recently to resource metrics so they can sort of serve as a drop-in replacement for these. Note opening issue in the future would be good so we don't forget to remove them or at least discuss their removal.

pkg/scheduler/metrics/resources/resources.go
smarterclayton: https://github.com/kubernetes/kubernetes/issues/96515 covers this

pkg/scheduler/metrics/resources/resources.go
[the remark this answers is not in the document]
smarterclayton: https://github.com/kubernetes/kubernetes/issues/96515 and in comments

smarterclayton:
> I don't see a problem with the kube_pod prefix, especially since these metrics will have different target labels, so both could be scrapped if a user wants to and you can always identify them by their target labels (e.g. job label being one of them). So this naming sounds good from my side, at least as kube-state-metrics maintainer. 

That's an even better point (I forgot about target labels)!

smarterclayton:
I think all comments are addressed - PTAL

pkg/api/v1/resource/helpers.go
```diff
@@ -21,14 +21,14 @@ import (
...
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/kubernetes/pkg/features"
 )
 
-// addResourceList adds the resources in newList to list
-func addResourceList(list, newList v1.ResourceList) {
+// AddResourceList adds the resources in newList to list.
```
alculquicondor, line 30: Better unexport this now that there's the reuse version

alculquicondor (commented):
/approve
from sig-scheduling

smarterclayton:
Any other comments or can someone tag this LGTM?

Huang-Wei:
/lgtm

Thanks!

ehashman:
Also

/lgtm

PR #94869, opened 2020-09-17
kubeadm: remove the --kubelet-version flag for "upgrade node"

### LINKED ISSUE #2021: "upgrade node": deprecate and delete --kubelet-version command line option
The `--kubelet-version` parameter exists in `kubeadm upgrade node` for primarily historic reasons. However, supporting it becomes more and more cumbersome. In addition to that, it bypasses the kubeadm component configs machinery, thus it may produce unexpected results (such as patching with kubeadm defaults, validation, etc.)

This option is part of a GA command line interface. Hence, the [GA command line interface deprecation policies](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-a-flag-or-cli) should be used.

Marked as deprecated: kubernetes/kubernetes#87942 (Kubernetes 1.18)
Deleted: TBD (1.18 release date + 12 months or 1.20 release, whichever takes longer)

/cc @timothysc @neolit123 @fabriziopandini @ereslibre

/kind cleanup
/area UX

neolit123:
The flag was deprecated as it is problematic since it allows
overrides of the kubelet configuration that is downloaded
from the cluster during upgrade.

Kubeadm node upgrades already download the KubeletConfiguration
and store it in the internal ClusterConfiguration type. It is then
only a matter of writing that KubeletConfiguration to disk.

Fixes https://github.com/kubernetes/kubeadm/issues/2021

```release-note
kubeadm: remove the deprecated --kubelet-config flag for the command "kubeadm upgrade node"
```

```docs

```

### DIFF
--- a/cmd/kubeadm/app/cmd/phases/upgrade/node/data.go
+++ b/cmd/kubeadm/app/cmd/phases/upgrade/node/data.go
@@ -28,7 +28,6 @@ type Data interface {
 	EtcdUpgrade() bool
 	RenewCerts() bool
 	DryRun() bool
-	KubeletVersion() string
 	Cfg() *kubeadmapi.InitConfiguration
 	IsControlPlaneNode() bool
 	Client() clientset.Interface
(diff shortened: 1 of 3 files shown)


yagonobre (commented):
/lgtm

PR #94873, opened 2020-09-18
Fix the `cloudprovider_azure_api_request_duration_seconds` metric buckets to correctly capture the latency metrics

marwanad:
**What type of PR is this?**
/kind bug

The existing metric uses the default prometheus buckets:
```
[]float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}
```
leading to the majority of mutating calls falling in the "+Inf" bucket so the metric becomes only useful for GET calls.

Fixes #

```release-note
Fix the `cloudprovider_azure_api_request_duration_seconds` metric buckets to correctly capture the latency metrics. Previously, the majority of the calls would fall in the "+Inf" bucket.
```

```docs

```


feiskyer (commented):
LGTM
/priority important-soon

PR #94879, opened 2020-09-18
kubeadm: make "alpha kubeconfig user" accept --config

knight42:
**What type of PR is this?**

/kind feature

xref https://github.com/kubernetes/kubeadm/issues/2292

```release-note
kubeadm: make the command "kubeadm alpha kubeconfig user" accept a "--config" flag and remove the following flags:
- apiserver-advertise-address / apiserver-bind-port: use either localAPIEndpoint from InitConfiguration or controlPlaneEndpoint from ClusterConfiguration.
- cluster-name: use clusterName from ClusterConfiguration
- cert-dir: use certificatesDir from ClusterConfiguration
```

```docs

```

### DIFF
--- a/cmd/kubeadm/app/cmd/alpha/kubeconfig.go
+++ b/cmd/kubeadm/app/cmd/alpha/kubeconfig.go
@@ -19,9 +19,8 @@ package alpha
 import (
 	"io"
 
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
-	kubeadmscheme "k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/scheme"
+
 	kubeadmapiv1beta2 "k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta2"
 	"k8s.io/kubernetes/cmd/kubeadm/app/cmd/options"
 	cmdutil "k8s.io/kubernetes/cmd/kubeadm/app/cmd/util"
@@ -39,8 +38,8 @@ var (
 	` + cmdutil.AlphaDisclaimer)
 
 	userKubeconfigExample = cmdutil.Examples(`
-	# Output a kubeconfig file for an additional user named foo
-	kubeadm alpha kubeconfig user --client-name=foo
+	# Output a kubeconfig file for an additional user named foo using a kubeadm config file bar
+	kubeadm alpha kubeconfig user --client-name=foo --config=bar
 	`)
 )
 
@@ -62,12 +61,10 @@ func newCmdUserKubeConfig(out io.Writer) *cobra.Command {
 	initCfg := &kubeadmapiv1beta2.InitConfiguration{}
 	clusterCfg := &kubeadmapiv1beta2.ClusterConfiguration{}
 
-	// Default values for the cobra help text
-	kubeadmscheme.Scheme.Default(initCfg)
-	kubeadmscheme.Scheme.Default(clusterCfg)
-
-	var token, clientName string
-	var organizations []string
+	var (
+		token, clientName, cfgPath string
+		organizations              []string
+	)
 
 	// Creates the UX Command
 	cmd := &cobra.Command{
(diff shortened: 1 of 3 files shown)


cmd/kubeadm/app/cmd/alpha/kubeconfig.go
```diff
@@ -81,29 +77,23 @@ func newCmdUserKubeConfig(out io.Writer) *cobra.Command {
...
 	}
 
-	// Add ClusterConfiguration backed flags to the command
-	cmd.Flags().StringVar(&clusterCfg.CertificatesDir, options.CertificatesDir, clusterCfg.CertificatesDir, "The path where certificates are stored")
-	cmd.Flags().StringVar(&clusterCfg.ClusterName, "cluster-name", clusterCfg.ClusterName, "Cluster name to be used in kubeconfig")
-
-	// Add InitConfiguration backed flags to the command
-	cmd.Flags().StringVar(&initCfg.LocalAPIEndpoint.AdvertiseAddress, options.APIServerAdvertiseAddress, initCfg.LocalAPIEndpoint.AdvertiseAddress, "The IP address the API server is accessible on")
-	cmd.Flags().Int32Var(&initCfg.LocalAPIEndpoint.BindPort, options.APIServerBindPort, initCfg.LocalAPIEndpoint.BindPort, "The port the API server is accessible on")
+	options.AddConfigFlag(cmd.Flags(), &cfgPath)
```
knight42, line 96: @neolit123 Currently this command only accept local kubeadm config, should we also consider the config saved in cluster?

neolit123: thanks for the PR, i can have a deeper look next week.

that is a good question. we had previous discussions whether certain kubeadm commands should use the kubeadm-config CM. my take is that if a command has potential to *not operate* on the running cluster it should not fetch the CM.

for example, if the command is currently being run on a Node part of a cluster and if a kubeconfig to access the cluster is in the user home dir, the user might still want to create a kubeadm for a completely different cluster, but the command will use the configuration from the cluster, which might not be desired.

i had the same argument for the command "kubeadm config images...". "kubeadm alpha certs renew" on the other hand has CM fetch, but the idea of the command is to renew certs for this Node / cluster.

so my preference here would be the following:
- the user should feed --config and the command should fail without it.
- if the user wants to feed the --config for the current cluster they can fetch the ClusterConfiguration from the CM.

knight42: I got it, thanks for you detailed clarification!

neolit123:
doing a minor retitle:

/retitle kubeadm: make "alpha kubeconfig user" accept --config

neolit123:
please change the release note to:

```
kubeadm: make the command "kubeadm alpha kubeconfig user" accept a "--config" flag and remove the following flags:
- apiserver-advertise-address / apiserver-bind-port: use either localAPIEndpoint from InitConfiguration or controlPlaneEndpoint from ClusterConfiguration.
- cluster-name: use clusterName from ClusterConfiguration
- cert-dir: use certificatesDir from ClusterConfiguration
```

cmd/kubeadm/app/cmd/alpha/kubeconfig.go
```diff
@@ -79,36 +75,35 @@ func newCmdUserKubeConfig(out io.Writer) *cobra.Command {
...
 
 	// Add command specific flags
 	cmd.Flags().StringVar(&token, options.TokenStr, token, "The token that should be used as the authentication mechanism for this kubeconfig, instead of client certificates")
 	cmd.Flags().StringVar(&clientName, "client-name", clientName, "The name of user. It will be used as the CN if client certificates are created")
 	cmd.Flags().StringSliceVar(&organizations, "org", organizations, "The orgnizations of the client certificate. It will be used as the O if client certificates are created")
 
+	_ = cmd.MarkFlagRequired(options.CfgPath)
+	_ = cmd.MarkFlagRequired("client-name")
```
neolit123, line 107: these can be without the leading `_ =`, unless linters complain?

cmd/kubeadm/app/cmd/alpha/kubeconfig.go
```diff
@@ -79,36 +75,35 @@ func newCmdUserKubeConfig(out io.Writer) *cobra.Command {
 			if clientName == "" {
 				return errors.New("missing required argument --client-name")
 			}
+			if cfgPath == "" {
+				return errors.New("missing required argument --config")
+			}
```
neolit123, line 80: we can remove the above `if` checks if cmd.MarkFlagRequired is used.

cmd/kubeadm/app/cmd/alpha/kubeconfig_test.go
```diff
@@ -19,16 +19,66 @@ package alpha
...
+			APIVersion: "kubeadm.k8s.io/v1beta2",
+			Kind:       "InitConfiguration",
+		},
+		LocalAPIEndpoint: kubeadmapiv1beta2.APIEndpoint{
+			AdvertiseAddress: "1.2.3.4",
+			BindPort:         1234,
+		},
+		NodeRegistration: kubeadmapiv1beta2.NodeRegistrationOptions{
+			CRISocket: "/path/to/dockershim.sock",
+		},
```
neolit123, line 53: i think the NodeRegistration above can be removed

cmd/kubeadm/app/cmd/alpha/kubeconfig_test.go
```diff
@@ -19,16 +19,66 @@ package alpha
...
+	"sigs.k8s.io/yaml"
+
+	kubeadmapiv1beta2 "k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta2"
 	kubeadmconstants "k8s.io/kubernetes/cmd/kubeadm/app/constants"
 	"k8s.io/kubernetes/cmd/kubeadm/app/util/pkiutil"
 	testutil "k8s.io/kubernetes/cmd/kubeadm/test"
 	kubeconfigtestutil "k8s.io/kubernetes/cmd/kubeadm/test/kubeconfig"
 )
 
+func generateKubeadmConfig(dir, id, certDir, clusterName string) (string, error) {
```
neolit123, line 39: possibly best to indicate this is used for tests package scope wise - e.g. name the function generateTestKubeadmConfig

cmd/kubeadm/app/cmd/alpha/kubeconfig_test.go
```diff
@@ -19,16 +19,66 @@ package alpha
...
+	}
+	buf.Write(data)
+	buf.WriteString("---\n")
+	data, err = yaml.Marshal(&clusterCfg)
+	if err != nil {
+		return "", err
+	}
+	buf.Write(data)
+
+	err = ioutil.WriteFile(cfgPath, buf.Bytes(), 0o644)
```
neolit123, line 78: should be `0644` for consistency with the rest of kubeadm code.

cmd/kubeadm/app/cmd/alpha/kubeconfig_test.go
```diff
@@ -19,16 +19,66 @@ package alpha
...
+
+	kubeadmapiv1beta2 "k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta2"
 	kubeadmconstants "k8s.io/kubernetes/cmd/kubeadm/app/constants"
 	"k8s.io/kubernetes/cmd/kubeadm/app/util/pkiutil"
 	testutil "k8s.io/kubernetes/cmd/kubeadm/test"
 	kubeconfigtestutil "k8s.io/kubernetes/cmd/kubeadm/test/kubeconfig"
 )
 
+func generateKubeadmConfig(dir, id, certDir, clusterName string) (string, error) {
+	id = strings.ReplaceAll(id, " ", "_") + ".conf"
```
neolit123, line 40: i think we can leave "id" here as is, without replacing " " and adding the extension.

cmd/kubeadm/app/cmd/alpha/kubeconfig.go
```diff
@@ -62,11 +62,7 @@ func newCmdUserKubeConfig(out io.Writer) *cobra.Command {
 	initCfg := &kubeadmapiv1beta2.InitConfiguration{}
```
neolit123, line 62: we should update the above example too:
```
userKubeconfigExample = cmdutil.Examples(`
	# Output a kubeconfig file for an additional user named foo using a kubeadm config file bar
	kubeadm alpha kubeconfig user --client-name=foo --config=bar
	`)
```

neolit123 (commented):
@knight42 thanks for the PR, i've added some minor comments.
i think we can keep the commits squashed to 1 here.

LMK if you have any questions.

cmd/kubeadm/app/cmd/alpha/kubeconfig.go
knight42: yeah just want to make goland happy.

cmd/kubeadm/app/cmd/alpha/kubeconfig.go
```diff
@@ -40,7 +39,7 @@ var (
 
 	userKubeconfigExample = cmdutil.Examples(`
 	# Output a kubeconfig file for an additional user named foo
```
neolit123, line 41: probably best to indicate that --config is the kubeadm configuration, as mentioned here:
https://github.com/kubernetes/kubernetes/pull/94879/files#r492257006

neolit123 (commented):
added one final comment. i think this is good to go after that.
/approve

neolit123:
thank you @knight42 
if someone objects to the change made here we can amend before 1.20 is released.
/lgtm
/milestone v1.20

PR #94881, opened 2020-09-18
e2e: skip write/read cache when running Block-mode RWX PVC tests

### LINKED ISSUE #94882: Multivolume test failure: RBD with access mode RWX
**What happened**:

The following test fails consistently when testing RBD block volmode with RWX access mode (not running in Kubernetes CI):

```
External Storage [Driver: openshift-storage.rbd.csi.ceph.com] [Testpattern: Dynamic PV (block volmode)] multiVolume [Slow] should concurrently access the single volume from pods on different node
```

Example:
```
fail [k8s.io/kubernetes@v1.19.0-rc.2/test/e2e/storage/utils/utils.go:85]: "dd if=/mnt/volume1 bs=64 count=1 | sha256sum | grep -Fq 53b29bcc6c22fffff3af5346dd3e471764c588530f7b883655ccf3d7b2ceb6ec" should succeed, but failed with error message "command terminated with exit code 1"
Unexpected error:
	<exec.CodeExitError>: {
		Err: {
			s: "command terminated with exit code 1",
		},
		Code: 1,
	}
	command terminated with exit code 1
occurred

failed: (53.7s) 2020-09-15T08:19:55 "External Storage [Driver: openshift-storage.rbd.csi.ceph.com] [Testpattern: Dynamic PV (block volmode)] multiVolume [Slow] should concurrently access the single volume from pods on different node"
```

The reason this fails is because a RBD block (used in different nodes) expects that the application (in this case `dd`) uses direct I/O to access the device. In the failure above, `dd` is reading the bytes from the OS **page cache** rather than the block itself.

1. Created 2 pods sharing a RBD block volume with access mode RWX
2. Make sure they are *not* running on the same node
3. From pod A, write some random data to the block device,  then calculate the `sha256sum`
4. From pod B, calculate the `sha256sum` of the shared block volume. It will return a different sum

/sig storage
/assign @nixpanic

nixpanic:
**What type of PR is this?**

/kind failing-test
/sig storage

For testing certain features, the BusyBox image does not provide all the
tools that are needed. Notably 'dd' from BusyBox does not support
direct-io that is required for skipping caches while doing writes and
reads on a Block-mode PVC attached to different nodes.

By passing "oflag=nocache" and "iflag=direct", caching should be
disabled while writing/reading with "dd" to a block device. The
TestConcurrentAccessToSingleVolume() test is known to fail with certain
storage backends (like Ceph RBD) when caching is enabled.

The default BusyBox image used for testing does not support the required
options for "dd". So instead of running with BusyBox, run the test with
a Debian image.

Fixes #94882

Marked **[WIP]** as the current changes force a single sector write when direct-io is requested. This is probably not the nicest way of doing things. Recommendations welcome.

```release-note
NONE
```

### DIFF
--- a/test/e2e/storage/testsuites/multivolume.go
+++ b/test/e2e/storage/testsuites/multivolume.go
@@ -35,6 +35,7 @@ import (
 	e2evolume "k8s.io/kubernetes/test/e2e/framework/volume"
 	"k8s.io/kubernetes/test/e2e/storage/testpatterns"
 	"k8s.io/kubernetes/test/e2e/storage/utils"
+	imageutils "k8s.io/kubernetes/test/utils/image"
 )
 
 type multiVolumeTestSuite struct {
@@ -423,14 +424,14 @@ func testAccessMultipleVolumes(f *framework.Framework, cs clientset.Interface, n
 
 		if readSeedBase > 0 {
 			ginkgo.By(fmt.Sprintf("Checking if read from the volume%d works properly", index))
-			utils.CheckReadFromPath(f, pod, *pvc.Spec.VolumeMode, path, byteLen, readSeedBase+int64(i))
+			utils.CheckReadFromPath(f, pod, *pvc.Spec.VolumeMode, false, path, byteLen, readSeedBase+int64(i))
 		}
 
 		ginkgo.By(fmt.Sprintf("Checking if write to the volume%d works properly", index))
-		utils.CheckWriteToPath(f, pod, *pvc.Spec.VolumeMode, path, byteLen, writeSeedBase+int64(i))
+		utils.CheckWriteToPath(f, pod, *pvc.Spec.VolumeMode, false, path, byteLen, writeSeedBase+int64(i))
 
 		ginkgo.By(fmt.Sprintf("Checking if read from the volume%d works properly", index))
-		utils.CheckReadFromPath(f, pod, *pvc.Spec.VolumeMode, path, byteLen, writeSeedBase+int64(i))
+		utils.CheckReadFromPath(f, pod, *pvc.Spec.VolumeMode, false, path, byteLen, writeSeedBase+int64(i))
 	}
 
 	pod, err = cs.CoreV1().Pods(pod.Namespace).Get(context.TODO(), pod.Name, metav1.GetOptions{})
@@ -481,6 +482,7 @@ func TestConcurrentAccessToSingleVolume(f *framework.Framework, cs clientset.Int
 		ginkgo.By(fmt.Sprintf("Creating pod%d with a volume on %+v", index, node))
 		podConfig := e2epod.Config{
 			NS:            ns,
+			ImageID:       imageutils.DebianIptables,
 			PVCs:          []*v1.PersistentVolumeClaim{pvc},
 			SeLinuxLabel:  e2epv.SELinuxLabel,
 			NodeSelection: node,
@@ -506,6 +508,14 @@ func TestConcurrentAccessToSingleVolume(f *framework.Framework, cs clientset.Int
 
 	var seed int64
 	byteLen := 64
+	directIO := false
+	// direct IO is needed for Block-mode PVs
+	if *pvc.Spec.VolumeMode == v1.PersistentVolumeBlock {
+		// byteLen should be the size of a sector to enable direct I/O
+		byteLen = 512
+		directIO = true
+	}
+
 	path := "/mnt/volume1"
 	// Check if volume can be accessed from each pod
 	for i, pod := range pods {
@@ -521,17 +531,17 @@ func TestConcurrentAccessToSingleVolume(f *framework.Framework, cs clientset.Int
 		if i != 0 {
 			ginkgo.By(fmt.Sprintf("From pod%d, checking if reading the data that pod%d write works properly", index, index-1))
 			// For 1st pod, no one has written data yet, so pass the read check
-			utils.CheckReadFromPath(f, pod, *pvc.Spec.VolumeMode, path, byteLen, seed)
+			utils.CheckReadFromPath(f, pod, *pvc.Spec.VolumeMode, directIO, path, byteLen, seed)
 		}
 
 		// Update the seed and check if write/read works properly
 		seed = time.Now().UTC().UnixNano()
 
 		ginkgo.By(fmt.Sprintf("Checking if write to the volume in pod%d works properly", index))
-		utils.CheckWriteToPath(f, pod, *pvc.Spec.VolumeMode, path, byteLen, seed)
+		utils.CheckWriteToPath(f, pod, *pvc.Spec.VolumeMode, directIO, path, byteLen, seed)
 
 		ginkgo.By(fmt.Sprintf("Checking if read from the volume in pod%d works properly", index))
-		utils.CheckReadFromPath(f, pod, *pvc.Spec.VolumeMode, path, byteLen, seed)
+		utils.CheckReadFromPath(f, pod, *pvc.Spec.VolumeMode, directIO, path, byteLen, seed)
 	}
 
 	// Delete the last pod and remove from slice of pods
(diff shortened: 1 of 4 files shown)


test/e2e/storage/testsuites/multivolume.go
```diff
@@ -521,17 +523,17 @@ func TestConcurrentAccessToSingleVolume(f *framework.Framework, cs clientset.Int
...
-			utils.CheckReadFromPath(f, pod, *pvc.Spec.VolumeMode, path, byteLen, seed)
+			utils.CheckReadFromPath(f, pod, *pvc.Spec.VolumeMode, true, path, byteLen, seed)
 		}
 
 		// Update the seed and check if write/read works properly
 		seed = time.Now().UTC().UnixNano()
 
 		ginkgo.By(fmt.Sprintf("Checking if write to the volume in pod%d works properly", index))
-		utils.CheckWriteToPath(f, pod, *pvc.Spec.VolumeMode, path, byteLen, seed)
+		utils.CheckWriteToPath(f, pod, *pvc.Spec.VolumeMode, true, path, byteLen, seed)
```
mkimuram, line 534: I might not completely understand the issue, but can it be resolved by just adding "sync command" here and the similar code in recheck?

test/e2e/storage/utils/utils.go
```diff
@@ -656,33 +656,55 @@ func genBinDataFromSeed(len int, seed int64) []byte {
...
+func CheckReadFromPath(f *framework.Framework, pod *v1.Pod, volMode v1.PersistentVolumeMode, disableCache bool, path string, len int, seed int64) {
 	var pathForVolMode string
+	nocache := ""
 	if volMode == v1.PersistentVolumeBlock {
 		pathForVolMode = path
+		if disableCache {
+			// TODO: "len" (dd: bs+count) need to be adjusted to
+			// sector(s). Make sure CheckWriteToPath() wrote enough
+			// bytes!
+			len = 512
```
mkimuram, line 671: Is this length change specific to RBD or will it be also required for other backend?

mkimuram (commented):
@nixpanic

Thank you for reporting an issue and providing the fix for it.

test/e2e/storage/testsuites/multivolume.go
nixpanic: Maybe a real `sync` is sufficient, but the command that is part of BusyBox is not (no error, not synced). I tried it with a CentOS container, and there permissions prevented it from working.

test/e2e/storage/utils/utils.go
nixpanic: This is the right size for most block devices, possibly not all. It can be detected with `blockdev --getss <dev>` though. Making sure to detect it, and write the right size in the `dd` command becomes really ugly, unfortunately.

test/e2e/storage/testsuites/multivolume.go
bertinatto: `sync` flushes pending writes to the disc, AFAICT it won't update the the page cache (at least on Linux).

Ideally we'd just drop the page cache in the "reading" node by doing `echo 3 > /proc/sys/vm/drop_caches `) before calling `dd`, but I don't think that's possible inside a container.

mkimuram: Thank you for your explanation and testing.

How about `conv=fsync` option of dd command?

The dd command in busybox seems to support it and https://unix.stackexchange.com/questions/473854/block-device-cache-v-s-a-filesystem and https://stackoverflow.com/questions/49489798/is-running-sync-necessary-after-writing-a-disk-image seem to say that it will work well.

nixpanic: I tried that as well. Unfortunately it does not help with the reading part where some readahead seems to happen (detecting partition table?). Because reading needs the whole sector for direct-io, the same amount of bytes need to be written (to make the checksum match).

mkimuram: Now I understand that we need a way to avoid reader from reading from the cache, not just to guarantee the writer to write to the actual device before reader starts to read.
And it seems that the only way to achieve it in container is to use direct I/O, so I agree to use direct I/O for this purpose.

nixpanic: > Reader side's requirement for avoid reading from cache needs to be resolved by using `iflag=nocache` in reading, however does writer side really needs direct I/O?

Unfortunately, there is no `iflag=nocache` option, the `nocache` is only for the `oflag` used by the writer.

You are correct, the writer does not require direct I/O. That side only needs to flush/sync the data after writing. The `oflag=nocache` can be replaced by `oflag=fdatasync` or similar. However, I do expect that applications using RWX block-mode PVCs use direct I/O on both sides when they need the data available on all participating nodes (like QEMU live-migration).

> 512 byte restriction seems to come from using `iflag=direct`, so I'm thinking about avoiding it by using other options in writing, like `conv=fsync` and/or `oflag=nocache` if we can really sync by these options.
> (Using debian image would be needed to use `iflag=nocache`, but avoiding direct I/O might still be possible. And it will potentially remove the TODOs in your codes.)
> 
> Does it make sense?

It really does, and you understood the issue perfectly! However, the missing support for `iflag=nocache` is preventing this from being an option. So the `iflag=direct` is the only alternative that is left. Direct I/O is the only option for readers to open a file/device without caching, see `man 2 open` and the `O_DIRECT` description, all the sync-flags for `open()` are related to writing only.

bertinatto: > Direct I/O is the only option for readers to open a file/device without caching, see man 2 open and the O_DIRECT description, all the sync-flags for open() are related to writing only.

This can be confirmed by running this little program in pods allocated in different nodes (and with a shared RWX volume): https://gist.github.com/bertinatto/6783734c636907e699d999b7a99bd622

With the O_DIRECT flag it prints the same output regardless the node where the container runs. Without the flag, the results are different for each node.

mkimuram: @nixpanic @bertinatto

Thank you for your explanation.

I understand that even though `iflag=nocache` can be passed to dd command, but won't be sufficient to skip cache.
So, I agree to apply this combination of options.

test/e2e/storage/utils/utils.go
mkimuram: Should we consider changing `byteLen` in `TestKubeletRestartsAndRestoresMount` to 512?
I think that we don't need to keep comment as TODO, but we need a comment on block size consideration for block volume test case,
so that developer for any backend devices which have 512+ block size can notice it and change the size bigger.

test/e2e/storage/utils/utils.go
```diff
@@ -656,33 +656,55 @@ func genBinDataFromSeed(len int, seed int64) []byte {
...
+			nocache = "iflag=direct"
```
mkimuram, line 672: Should we consider renaming to `iflag`?  
(Then the one in `CheckWriteToPath` should be `oflag`.)

mkimuram (commented):
Entire codes look almost good to me. There are some comments for minor changes.

test/e2e/storage/utils/utils.go
nixpanic: Modified the test to just write 512 bytes and explained the sector size in comments. I like this much better :+1:

test/e2e/storage/utils/utils.go
nixpanic: I have adapted the names of the variables, `nocache bool` and `directIO bool` as arguments, `oflag` and `iflag` as parameters to `dd`.

nixpanic:
Removed some generated files, so the following labels can be removed, I think:

- area/kubelet
- sig/network
- sig/node
- sig/scheduling

nixpanic:
Not sure where this fails, but the modified `dd` command passes:

    I0925 14:10:55.928] STEP: Checking if read from the volume in pod1 works properly
    I0925 14:10:55.929] Sep 25 14:09:38.188: INFO: ExecWithOptions {Command:[/bin/sh -c dd if=/mnt/volume1 iflag=direct bs=512 count=1 | sha256sum] Namespace:multivolume-3815 PodName:pod-aba9c0a1-dbf1-4a76-94ff-2d51afd0f723 ContainerName:write-pod Stdin:<nil> CaptureStdout:true CaptureStderr:true PreserveWhitespace:false}

mkimuram:
@nixpanic

Thank you for your fix.
I think the time out error isn't due to your change.

/lgtm

@msau42

Could you review and approve this PR?

test/e2e/storage/testsuites/multivolume.go
```diff
@@ -506,6 +508,14 @@ func TestConcurrentAccessToSingleVolume(f *framework.Framework, cs clientset.Int
 
 	var seed int64
 	byteLen := 64
+	directIO := false
+	// direct IO is needed for Block-mode PVs
+	if *pvc.Spec.VolumeMode == v1.PersistentVolumeBlock {
+		// byteLen should be the size of a sector to enable direct I/O
+		byteLen = 512
```
msau42, line 515: What happens if you make byteLen 512 in all cases?

test/e2e/storage/testsuites/multivolume.go
```diff
@@ -506,6 +508,14 @@ func TestConcurrentAccessToSingleVolume(f *framework.Framework, cs clientset.Int
 
 	var seed int64
 	byteLen := 64
+	directIO := false
+	// direct IO is needed for Block-mode PVs
```
msau42, line 512: Can you explain why in the comment?

test/e2e/storage/testsuites/multivolume.go
```diff
@@ -506,6 +508,14 @@ func TestConcurrentAccessToSingleVolume(f *framework.Framework, cs clientset.Int
...
+	if *pvc.Spec.VolumeMode == v1.PersistentVolumeBlock {
```
msau42, line 513: Should we also check AccessMode = rwx?

msau42: nm we only get here if it's rwx already

msau42:
/approve

We'll need to revisit this whenever Windows supports raw block. cc @jingxu97 @ddebroy

test/e2e/storage/testsuites/multivolume.go
nixpanic: When attaching a block PV to a node, the Linux kernel tries to detect a partition table (reading the first sector(s) of the device). Because of read caching, writes to the device from an other node will not show up until the caches are dropped. Dropping caches is a privileged operation, and can not be done from within a (standard) pod. The alternative is to read with caches disabled, which is what direct I/O does.

PR #94888, opened 2020-09-18
apiextensions: prune array type without items in published OpenAPI

sttts:
kubectl falls over arrays without item schema. Hence, we have to publish a less precise OpenAPI spec (similar to other pruning we already do for the same reason).

Without this kubectl will show e.g.

```
error: SchemaError(io.openshift.tuned.v1.Tuned.spec.recommend.match.match): array should have exactly one sub-item
```

for the following schema in a CRD:

```yaml
type: array
x-kubernetes-preserve-unknown-fields: true
items:
  type: string
```

/kind bug

```release-note
Fix regression in kubectl validation in 1.18+ on CRDs with schema using x-kubernetes-preserve-unknown-fields on array types.
```

### DIFF
--- a/staging/src/k8s.io/apiextensions-apiserver/pkg/controller/openapi/v2/BUILD
+++ b/staging/src/k8s.io/apiextensions-apiserver/pkg/controller/openapi/v2/BUILD
@@ -24,6 +24,7 @@ go_test(
         "//vendor/github.com/googleapis/gnostic/openapiv2:go_default_library",
         "//vendor/gopkg.in/yaml.v2:go_default_library",
         "//vendor/k8s.io/kube-openapi/pkg/util/proto:go_default_library",
+        "//vendor/k8s.io/utils/pointer:go_default_library",
     ],
 )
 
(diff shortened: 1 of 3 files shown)


liggitt:
Stopping publishing openapi for existing cases concerns me more the longer we go. What would it look like to start fixing these edge cases in kubectl as we encounter them instead?

soltysh:
> Stopping publishing openapi for existing cases concerns me more the longer we go. What would it look like to start fixing these edge cases in kubectl as we encounter them instead?

Honestly, nothing stops us, although I must admit I don't know what is the current state of openapi in kubectl. I need to check that.

sttts:
> Stopping publishing openapi for existing cases concerns me more the longer we go. What would it look like to start fixing these edge cases in kubectl as we encounter them instead?

@liggitt We would have to fork go-openapi/validate and make use of it in kubectl. All doable, just needs men power. If you have an engineer to do this, I am happy to coach.

soltysh (approved):
/lgtm

liggitt:
If someone was successfully using a CRD with this type of schema and making use of the published openapi, wouldn't this PR break them?

PR #94895, opened 2020-09-18
volume: Change owner of symlinks too

mauriciovasquezbernal:
**What type of PR is this?**
/kind cleanup

This commit uses Lchown instead of Chown to change the owner of symlinks too.
It doesn't change any behaviour. However, it could avoid some confusions as the
symlinks are updated to the correct owner too.

Please consider the following Pod with a configMap:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: myconfigmap
  namespace: default
data:
  content: '"This is a ConfigMap"'
---
apiVersion: v1
kind: Pod
metadata:
  name: mypod
spec:
  securityContext:
    fsGroup: 2000
  containers:
  - name: container1
    image: busybox
    command: ["sh"]
    args: ["-c", "sleep infinity"]
    volumeMounts:
    - name: config-volume
      mountPath: /etc/configmap/content
  volumes:
  - name: config-volume
    configMap:
      name: myconfigmap
      defaultMode: 0400
```

The current behaviour shows that the group owner is `0` for the symlink point to the real content file:
```
$ kubectl exec -it mypod -- /bin/sh -c 'ls -la /etc/configmap/content/content'
lrwxrwxrwx    1 0     0            14 Sep 18 15:15 /etc/configmap/content/content -> ..data/content
```

The behaviour after this PR shows as group owner `2000` as specified in `fsGroup`:
```
kubectl exec -it mypod -- /bin/sh -c 'ls -la /etc/configmap/content/content'
lrwxrwxrwx    1 0     2000            14 Sep 18 15:15 /etc/configmap/content/content -> ..data/content
```

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/volume/volume_linux.go
+++ b/pkg/volume/volume_linux.go
@@ -89,30 +89,20 @@ func legacyOwnershipChange(mounter Mounter, fsGroup *int64) error {
 }
 
 func changeFilePermission(filename string, fsGroup *int64, readonly bool, info os.FileInfo) error {
-	// chown and chmod pass through to the underlying file for symlinks.
+	err := os.Lchown(filename, -1, int(*fsGroup))
+	if err != nil {
+		klog.Errorf("Lchown failed on %v: %v", filename, err)
+	}
+
+	// chmod passes through to the underlying file for symlinks.
 	// Symlinks have a mode of 777 but this really doesn't mean anything.
 	// The permissions of the underlying file are what matter.
 	// However, if one reads the mode of a symlink then chmods the symlink
 	// with that mode, it changes the mode of the underlying file, overridden
 	// the defaultMode and permissions initialized by the volume plugin, which
-	// is not what we want; thus, we skip chown/chmod for symlinks.
+	// is not what we want; thus, we skip chmod for symlinks.
 	if info.Mode()&os.ModeSymlink != 0 {
 		return nil
-	}
-
-	stat, ok := info.Sys().(*syscall.Stat_t)
-	if !ok {
-		return nil
-	}
-
-	if stat == nil {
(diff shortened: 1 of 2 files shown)


pkg/volume/volume_linux.go
```diff
@@ -82,32 +82,22 @@ func legacyOwnershipChange(mounter Mounter, fsGroup *int64) error {
...
-	if stat == nil {
-		klog.Errorf("Got nil stat_t for path %v while setting ownership of volume", filename)
-		return nil
-	}
-
-	err := os.Chown(filename, int(stat.Uid), int(*fsGroup))
-	if err != nil {
-		klog.Errorf("Chown failed on %v: %v", filename, err)
-	}
-
```
gnufied, line 110: Isn't this changing the existing behaviour?

mauriciovasquezbernal: Strictly speaking, yes. Now the group owner of the symlinks is updated too. Am I missing anything else?

jingxu97:
could you add unit test? Thanks!

cc @msau42

mauriciovasquezbernal:
@jingxu97 I added the unit test.

mauriciovasquezbernal:
@gnufied @sjenning any chance you can give it a look?

gnufied:
It looks good to me as well. But it would be nice to have a confirmation from @sjenning because he originally made the change that prevented permissions of symlinks from being changed. But at that time - we were calling plain` chown` which evaluates the symlink and now you are using `lchown`.

joelsmith:
Nice that you were able to get rid of an extra stat syscall by using -1 for the UID!

/lgtm

PR #94899, opened 2020-09-18
Ensure secret pulled images

mikebrow:
/kind feature

```release-note
Implement support in kubelet to ensure images pulled with pod imagePullSecrets are authenticated
by other pods that do not have the same credentials.

```

- [x] https://github.com/kubernetes/enhancements/pull/1608
  code complete  
 [x] [feature gate added for issue comment](https://github.com/kubernetes/enhancements/pull/1608#issuecomment-838805092)
- [ ] usage (not needed for alpha)
- [ ] docs (not needed for alpha)
- [x] test bucket(s)
  unit buckets complete.. for fg on and off 

Signed-off-by: Mike Brown <brownwm@us.ibm.com>


mikebrow:
@Random-Liu found some time to work this one... needs testing, docs, etc.. just showing where I was going with the code changes to get some feedback if there are interested parties.

pkg/kubelet/images/image_manager.go
```diff
@@ -177,3 +231,57 @@ func applyDefaultImageTag(image string) (string, error) {
...
+		auth := &runtimeapi.AuthConfig{
+			Username:      currentCreds.Username,
+			Password:      currentCreds.Password,
+			Auth:          currentCreds.Auth,
+			ServerAddress: currentCreds.ServerAddress,
+			IdentityToken: currentCreds.IdentityToken,
+			RegistryToken: currentCreds.RegistryToken,
+		}
+
+		hash := kubecontainer.HashAuth(auth)
```
andrewsykim, line 275: Would this need to check whether the provided credentials were extracted from the imagePullSecrets and not from another docker auth provder in the keyring?

mikebrow: > Wondering if this need to check whether the provided credentials were extracted from the imagePullSecrets and not from another docker auth provder in the keyring?

see: https://github.com/kubernetes/kubernetes/pull/94899/files#diff-7a6e7cc9b15ce528a5503356cbe743bcR31-L59

andrewsykim: Right, so was wondering if we account for when the auth config came from the default keyring `m.keyring` and not from image pull secrets.

https://github.com/kubernetes/kubernetes/blob/f4112710f5bbfbd9cc7a68ede0df519af30d5e40/pkg/kubelet/kuberuntime/kuberuntime_image.go#L38-L41

Here it seems like we hash all the provided auth configs, not just image pull secrets. Is that expected?

mikebrow: ah.. Yes the idea is to match the hash of the auth cred passed to the container runtime that was used when the pull succeeded... to what is about to become a subsequent attempt. If a cred in the creds I get from the keyring already passed authentication, then it should pass this time as well was my thought. I was thinking we could time stamp it and force authentication with the same creds after some duration.  Or I could just use a single boolean for the imageRef indicating it was pulled with auth and thus should be pulled each time to re-verify, but that would be less performant for the PullIfNotPresent (with auth) scenario.

mikebrow: Pinged you on slack maybe we can have a talk about this. Near as I can tell it was coded this way since the first code drop July of 2016. @Random-Liu /  @feiskyer do you remember or know of any expectation here wrt. @andrewsykim 's question?

pkg/kubelet/images/image_manager.go
```diff
@@ -57,21 +76,35 @@ func NewImageManager(recorder record.EventRecorder, imageService kubecontainer.I
...
+		keyring:      keyring,
 	}
 }
 
 // shouldPullImage returns whether we should pull an image according to
 // the presence and pull policy of the image.
-func shouldPullImage(container *v1.Container, imagePresent bool) bool {
+func shouldPullImage(container *v1.Container, imagePresent, pulledBySecret, ensuredBySecret bool) bool {
 	if container.ImagePullPolicy == v1.PullNever {
 		return false
```
adisky, line 87: consider the case 
```
PodA
  image: imageA
  imagePullPolicy: pullIfNotPresent
  pullImageSecret: ###
```

```
PodB
 image: imageA
 imagePullPolicy: Never
```
with image pull policy `never`, Pod B will be able to access  imageA, shouldn't we handle this case also?

mikebrow: definitely should handle that case to.. can't pull though.. so result is still false here and below on line ..

ensuredBySecret will be false for this case and if pulledBySecret we have a new error case...

so below in this snippet on line 161:
```
		if present {
 			msg := fmt.Sprintf("Container image %q already present on machine", container.Image)
 			m.logIt(ref, v1.EventTypeNormal, events.PulledImage, logPrefix, msg, klog.Info)
			return imageRef, "", nil
		}
``` 
we'll have to add an error:
```
if present {
    if pulledBySecret && !ensuredBySecret {
	msg := fmt.Sprintf("Container image %q is present with pull policy of Never but does not have the proper permission (image secret) that was used to pull the image", container.Image)
	m.logIt(ref, v1.EventTypeWarning, events.ErrImageNeverPullPolicy, logPrefix, msg, klog.Warning)
	return "", msg, ErrImageNeverPull
   }
   msg := fmt.Sprintf("Container image %q already present on machine", container.Image)
   m.logIt(ref, v1.EventTypeNormal, events.PulledImage, logPrefix, msg, klog.Info)
   return imageRef, "", nil
}

```

ehashman:
/remove-priority important-soon

Seems the KEP missed the 1.22 deadline. @mikebrow are you filing an exception request?

mikebrow:
> /remove-priority important-soon
> 
> Seems the KEP missed the 1.22 deadline. @mikebrow are you filing an exception request?

need to find some time to figure out all the kep document changes I need to do for the new process, and what changes I need to do subsequent to adding a feature gate what, if any, additional tests do I need above and beyond the tests I added to run with the feature gate both on and off... what where do I document the feature gate... yada yada

ehashman:
Exception denied :disappointed: https://groups.google.com/g/kubernetes-sig-node/c/6RbfKP9aEL8/m/SeiY3VmbBAAJ

/hold
/priority important-longterm

pkg/kubelet/images/image_manager.go
```diff
@@ -177,3 +244,57 @@ func applyDefaultImageTag(image string) (string, error) {
...
+
+// isEnsuredBySecret - returns true if the secret for an auth used to pull an
+// image has already been authenticated through a successful pull request
+// and the same auth exists for this podSandbox/image/
+func (m *imageManager) isEnsuredBySecret(imageRef string, image kubecontainer.ImageSpec, pullSecrets []v1.Secret) (pulledBySecret, ensuredBySecret bool) {
+	if imageRef == "" {
+		return
+	}
+	if ensureSecretPulledImage[imageRef] != nil {
+		pulledBySecret = true
```
ruiwen-zhao, line 256: So this will ensure that preloaded images (or images pulled in some other hacky ways) will not be broken by this change, right? We will only block images that are pulled through kubelet.

mikebrow: An image pulled without a secret by kubelet (through kubelet) means that image was available on that node without a pod secret. We have no way to detect today, if the container runtime pulled an image with an auth or without auth., no way other than this new flag.  If an image is preloaded and the policy is pull always .. we have to pull anyways. If an image is preloaded and the policy is pull never then auth is not involved. If an image is present because it was pulled by kubelet with auth and the policy for a pod is pull never then we can't use the otherwise present image for that pod.  If the policy is pull if not present then we have to check if present and pull if not.. this new check makes sure an image pulled with auth is not used as the present check without being able to use the same auth.

mikebrow: right.. but just in case let me qualify

> we will only block

Block is an active word..  We will force an auth check for the manifests if an auth was required for the image pull during this instance of kubelet. So it's not really a block its more validate access when we are sure access was used. Thus for this active instance of kubelet an image pulled with access will only be present for pods having that access.  Images pulled outside this instance of kubelet are treated as unknown to this ensure check. 

Admins wanting more secure access could use garbage collection at the top of each kubelet instance and turn the feature gate on. 

Admins wanting to provide/configure preloaded images with image pull policies of pull never and pull if not present, where preloaded images are considered accessible by a plurality of pods, have the ability to continue to do so with or without using the feature gate.

see cases .. https://github.com/kubernetes/kubernetes/pull/94899/files#diff-7297f08c72da9bf6479e80c03b45e24ea92ccb11c0031549e51b51f88a91f813R382-R406

Broken.. if a use pull never, but the image was pulled by auth by another pod..

mikebrow:
@dims @vikramcse for this PR I needed to mock an additional return value... https://github.com/kubernetes/kubernetes/pull/94899/files#diff-585fa33ffeb80eeeef9b1444036d99263f8a5f7ec21ee41fe44a8a2486ddc895R115-R117

However, after https://github.com/kubernetes/kubernetes/pull/104933  I now have a merge conflict.. I'm wondering if you have a suggestion for how I should merge the reply field into the new go mock generated code..

vikramcse:
Hello @mikebrow 
Earlier all the mocks are generated manually using `mockery (go/testify)`. Now we have converted the mockery mock files to mockgen (gomock). Earlier this PR we have created a automated script which used `go:generate` to auto generate mock files for the interfaces that we have (https://github.com/kubernetes/kubernetes/pull/103631).

In this case we have a add an additional return parameter to the `PullImage` mock. For that what we can do is
1. The file `pkg/kubelet/container/testing/runtime_mock.go` is generated from `pkg/kubelet/container/runtime.go`
https://github.com/kubernetes/kubernetes/blob/0ac956ff2bef9c84b17d4b9cae2cffe71d5c6386/pkg/kubelet/container/testing/runtime_mock.go#L14-L22

which is generated from this `go:generate command`
https://github.com/kubernetes/kubernetes/blob/0ac956ff2bef9c84b17d4b9cae2cffe71d5c6386/pkg/kubelet/container/runtime.go#L14-L22

2. In this case one return parameter is added for that we have to modify the Interface in `runtime.go`
image

3. Now run `sudo hack/update-mocks.sh`, which will re generate the mock file, here is generated file which above change
image

Please let me know if there is anything needed from my side.

Thank You 
:)

mikebrow:
> .... run `sudo hack/update-mocks.sh`...

Thx @vikramcse Now to find a clean rebase commit :-)

rphillips:
Cancelling the hold which was set for 1.22 feature freeze.

/hold cancel

This looks ready to go.

/lgtm

mikebrow:
rebased..

mikebrow:
easycla auth request sent in to the IBM authorizer :-)

mikebrow:
rebased again ;-)

pkg/kubelet/cri/remote/remote_runtime_test.go
```diff
@@ -68,6 +68,8 @@ func TestVersion(t *testing.T) {
 	r := createRemoteRuntimeService(endpoint, t)
 	version, err := r.Version(apitest.FakeVersion)
 	assert.NoError(t, err)
-	assert.Equal(t, apitest.FakeVersion, version.Version)
-	assert.Equal(t, apitest.FakeRuntimeName, version.RuntimeName)
+	if version != nil {
+		assert.Equal(t, apitest.FakeVersion, version.Version)
+		assert.Equal(t, apitest.FakeRuntimeName, version.RuntimeName)
+	}
```
endocrimes, line 74:
```suggestion
	require.NoError(t, err)
	require.NotNil(t, version)
	require.Equal(t, apitest.FakeVersion, version.Version)
	require.Equal(t, apitest.FakeRuntimeName, version.RuntimeName)
```

require will end the test early rather than stacking failures, which I think acomplishes the same thing you were going for here? but without the potential for accidental false positives (if err && version == nil)

pkg/kubelet/images/image_manager.go
endocrimes: I'm not sure what we'd change based on accounting for that tbh? (and if anything started changing based on that it'd likely be a pretty big conceptual change?)

smarterclayton:
This probably needs a really good node_e2e test that deals with kubelet restart and down registries, because we have recently realized our testing in restart and recovery scenarios (and outage scenarios!) is particularly bad.  @ehashman do we think we can make some recommendations about what new kubelet features need to do w.r.t. kubelet restart testing and propose them for this PR in sig-node?

mikebrow:
> This probably needs a really good node_e2e test that deals with kubelet restart and down registries, because we have recently realized our testing in restart and recovery scenarios (and outage scenarios!) is particularly bad. @ehashman do we think we can make some recommendations about what new kubelet features need to do w.r.t. kubelet restart testing and propose them for this PR in sig-node?

nod.. Had/have intention to add such tests in the next version (post this alpha default off version) of this feature when it is storing and loading the pull metadata for each restart.. In this version the metadata is in memory only and thus reset on kubelet restart.

pkg/kubelet/cri/remote/remote_runtime_test.go
mikebrow: > accidental false positives

Good question.. I'm not against this test failing on nil in certain circumstances.. but this is tricky and I don't think the bucket is ready/well written for all the possible expected cases. Let me explain.

In the "field" the timing of (1) start runtime, (2)- connect to grpc socket remote, and (3) start making calls to the remote.. requires some delay before r.Version() and other apis are expected to actually work correctly. This is because of grpc timeouts and the runtime posting up the available grpc services in it's init cycle before it is finished with all of it's init work and actually ready for grpc server requests. There's a timing window here to be resolved. Typically we add a until "remoteService.Version() == nil; sleep(1sec)" type loop on clients to make sure we are actually talking to the server.

Clearly the intent of this test is to check if the fake runtime returns the expected meta.. and depending on system timing issues you might get a grpc timeout or bad connection instead and that gets swallowed.  We could add a sleep loop and wait on every cri api to return non nil. We could address this by considering if we need to add a client side error when grpc errors occur. We could also modify NewFakeRemoteRuntime or createRemoteRuntimeService to wait some time until the runtime is actually ready. 

IOW a slow startup or other grpc connection issue is not necessarily an error, unless it has a sleep loop in the test and the grpc induced timeouts go beyond some reasonable amount of time. The ability of the client to be considerate of a not-actually ready runtime or temporary grpc disconnect is an issue that needs some number of buckets for testing understanding and resolution. 

Another issue, should not avail (nil) be supported at all? The API does not specify...

I could add a "TODO: (Mike Brown) refactor for startup timing issues" if you like

pkg/kubelet/images/image_manager.go
mikebrow: > 

nod..

mikebrow: > 

nod

mikebrow: > I'm not sure what we'd change based on accounting for that tbh? (and if anything started changing based on that it'd likely be a pretty big conceptual change?)

nod..

pkg/kubelet/cri/remote/remote_runtime_test.go
dims: TODO sounds good @mikebrow

dims:
@mikebrow +1 to open an issue for a new test like you mentioned

pkg/kubelet/cri/remote/remote_runtime_test.go
endocrimes: SGTM

pkg/features/kube_features.go
```diff
@@ -855,6 +855,14 @@ const (
 	//
 	// Enables GRPC probe method for {Liveness,Readiness,Startup}Probe.
 	GRPCContainerProbe featuregate.Feature = "GRPCContainerProbe"
+
+	// owner: @mikebrow
+	// kep: http://kep.k8s.io/2535
+	// alpha: v1.23
```
pacoxu, line 861: nit: v1.24 😄

pacoxu: Overall this is LGTM and the KEP was merged in the v1.23 release cycle.

pkg/kubelet/images/image_manager_test.go
```diff
@@ -322,97 +325,114 @@ func TestShouldPullImage(t *testing.T) {
...
+			expectedWithFGOn:  false,
 		},
 		{
-			description:     "PullNever should never pull even even if pulled by secret and not ensured by secret",
-			container:       pullNever,
-			imagePresent:    true,
-			pulledBySecret:  true,
-			ensuredBySecret: false,
-			expected:        false,
+			description:       "PullNever should never pull even even if pulled by secret and not ensured by secret",
```
mrunalp, line 418: nit: `even even` -> `even`

endocrimes:
flake has in issue here: https://github.com/kubernetes/kubernetes/issues/106904

endocrimes (commented):
/lgtm

pacoxu (approved):
> /assign @mrunalp @klueska

/lgtm

mrunalp:
@mikebrow Could you clean up the commits? I am making a pass and it will be good if it is ready for merge :)

mikebrow:
> @mikebrow Could you clean up the commits? I am making a pass and it will be good if it is ready for merge :)

squashed .. :-)

dims:
/approve

(for pkg/features)

/hold

(for @mrunalp to wrap up review, please wait for his review before removing hold)

pkg/kubelet/images/image_manager.go
```diff
@@ -155,6 +205,24 @@ func (m *imageManager) EnsureImageExists(pod *v1.Pod, container *v1.Container, p
...
+
+	if imagePullResult.hash == "" {
+		// successful pull no auth hash returned, auth was not required so we should reset the hashmap for this
+		// imageref since auth is no longer required for the local image cache, allowing use of the ImageRef
+		// by other pods if it remains cached and pull policy is PullIfNotPresent
+		ensureSecretPulledImage[imageRef] = nil
+	}
+	// store/create hashMatch map entry for auth config hash key used to pull the image
+	// for this imageref (digest)
+	if imagePullResult.hash != "" {
```
mrunalp, line 217: nit: could be moved up to else branch of previous block.

pkg/kubelet/images/image_manager.go
```diff
@@ -177,3 +245,57 @@ func applyDefaultImageTag(image string) (string, error) {
...
+			Username:      currentCreds.Username,
+			Password:      currentCreds.Password,
+			Auth:          currentCreds.Auth,
+			ServerAddress: currentCreds.ServerAddress,
+			IdentityToken: currentCreds.IdentityToken,
+			RegistryToken: currentCreds.RegistryToken,
+		}
+
+		hash := kubecontainer.HashAuth(auth)
+		if hash != "" {
```
mrunalp, line 290: When can this be empty?

pkg/kubelet/images/image_manager.go
```diff
@@ -22,28 +22,50 @@ import (
...
 
+// ensureSecretPulledImage - map of imageref (image digest) to successful secret pulled image details
+var ensureSecretPulledImage map[string]*ensureSecretPulledImageDigest
+
+type ensureSecretPulledImageDigest struct {
+	// TODO: (mikebrow) time of last pull for this imageRef
+	// TODO: (mikebrow) time of pull for each particular auth hash
+
+	// map of auth hash (keys) used to successfully pull this imageref
+	HashMatch map[string]bool
```
mrunalp, line 47: nit: Could use https://github.com/kubernetes/apimachinery/blob/master/pkg/util/sets/string.go

pkg/kubelet/images/image_manager.go
mikebrow: should never happen

mikebrow: would require a failure to marshal auth.. fallback would be to fall back to prior (slower) 
behavior

pkg/kubelet/images/image_manager.go
mikebrow: will do

pkg/kubelet/images/image_manager.go
liggitt: > Admins wanting more secure access could use garbage collection at the top of each kubelet instance and turn the feature gate on.

It should be made super clear in the docs that this feature doesn't reliably protect access to images pulled via other credentials unless you do this, and doing this requires killing all existing containers and deleting images to restart the kubelet safely. That's not really a process I can see many admins signing up for.

Looking at the KEP, it looks like this "during current instance of Kubelet" aspect is only during the alpha phase? Is there anything we can do to indicate use of an image was allowed because it was pre-existing when the kubelet started? I'm trying to think of ways to head off bug reports that the feature is broken because people are confused by a kubelet restart.

I also think the on-disk persistence to make the protection survive kubelet restarts should be implemented and tested while in alpha (even if it's not in the first release)

pkg/kubelet/images/image_manager.go
```diff
@@ -155,6 +205,24 @@ func (m *imageManager) EnsureImageExists(pod *v1.Pod, container *v1.Container, p
 	}
 	m.logIt(ref, v1.EventTypeNormal, events.PulledImage, logPrefix, fmt.Sprintf("Successfully pulled image %q in %v", container.Image, time.Since(startTime)), klog.Info)
 	m.backOff.GC()
+
+	if imagePullResult.hash == "" {
+		// successful pull no auth hash returned, auth was not required so we should reset the hashmap for this
+		// imageref since auth is no longer required for the local image cache, allowing use of the ImageRef
+		// by other pods if it remains cached and pull policy is PullIfNotPresent
+		ensureSecretPulledImage[imageRef] = nil
```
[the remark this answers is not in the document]
liggitt: also, `delete(ensureSecretPulledImage, imageRef)` rather than storing an explicit nil?

pkg/kubelet/container/helpers.go
```diff
@@ -107,6 +108,14 @@ func HashContainer(container *v1.Container) uint64 {
 	return uint64(hash.Sum32())
 }
 
+// HashAuth - returns a hash code for a CRI pull image auth
+func HashAuth(auth *runtimeapi.AuthConfig) string {
+	hash := fnv.New64a()
```
liggitt, line 113: fnv is a non-cryptographic hash algorithm... squinting at whether that matters here, but non-cryptographic + credentials + security make me squint

pkg/kubelet/images/image_manager.go
mikebrow: per our conversation adding TODO to line 44 :-)

pkg/kubelet/images/image_manager.go
```diff
@@ -22,28 +22,50 @@ import (
...
+	"k8s.io/kubernetes/pkg/credentialprovider"
+	credentialprovidersecrets "k8s.io/kubernetes/pkg/credentialprovider/secrets"
+	"k8s.io/kubernetes/pkg/features"
 	kubecontainer "k8s.io/kubernetes/pkg/kubelet/container"
 	"k8s.io/kubernetes/pkg/kubelet/events"
+	"k8s.io/kubernetes/pkg/util/parsers"
 )
 
+// ensureSecretPulledImage - map of imageref (image digest) to successful secret pulled image details
+var ensureSecretPulledImage map[string]*ensureSecretPulledImageDigest
```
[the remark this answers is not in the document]
mikebrow: retains all for now.. see TODOs for discussion regarding lifecycle of authentication

pkg/kubelet/container/helpers.go
```diff
@@ -107,6 +108,14 @@ func HashContainer(container *v1.Container) uint64 {
 	return uint64(hash.Sum32())
 }
 
+// HashAuth - returns a hash code for a CRI pull image auth
+func HashAuth(auth *runtimeapi.AuthConfig) string {
```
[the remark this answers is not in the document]
mikebrow: not persisted in memory only.. TODOs/discussions in kep and earlier discussions that we'd start persisting and have a lifecycle post alpha / the next KEP for this

pkg/kubelet/images/image_manager.go
```diff
@@ -57,21 +79,39 @@ func NewImageManager(recorder record.EventRecorder, imageService kubecontainer.I
...
+func shouldPullImage(container *v1.Container, imagePresent, pulledBySecret, ensuredBySecret bool) bool {
 	if container.ImagePullPolicy == v1.PullNever {
 		return false
 	}
 
-	if container.ImagePullPolicy == v1.PullAlways ||
-		(container.ImagePullPolicy == v1.PullIfNotPresent && (!imagePresent)) {
+	if container.ImagePullPolicy == v1.PullAlways {
 		return true
 	}
```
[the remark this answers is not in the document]
mikebrow: suggest looking over the old code.. compare with new.. this PR has some refactoring in it as well and test cases to verify old and new behavior.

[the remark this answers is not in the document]
liggitt: it should be trivial to verify feature gated code is inert and doesn't change existing behavior… mixing a refactor here and inverting the feature gate block makes that really hard to see here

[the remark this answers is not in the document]
mikebrow: ok.. np

pkg/kubelet/container/helpers.go
mrunalp: Maybe we switch to sha256 so when we add persistence it is less of a worry.

pkg/kubelet/images/image_manager.go
mikebrow: agreed.. we should keep it in alpha while testing the persistence mechanism, and also while coming up with a way to validate a policy for image cache preloads (either prior to said persistence or through a load operation) 

I don't think we'll have a great answer to this question until we have image cache policies and apply those through the CRI to the image service's cache.

pkg/kubelet/images/image_manager.go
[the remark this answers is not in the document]
mikebrow: only removes entries if.. there is a successful pull without auth..

pkg/kubelet/images/image_manager.go
```diff
@@ -22,28 +22,50 @@ import (
...
+	// TODO: (mikebrow) time of last pull for this imageRef
+	// TODO: (mikebrow) time of pull for each particular auth hash
+
+	// map of auth hash (keys) used to successfully pull this imageref
+	HashMatch map[string]bool
+}
+
+func init() {
+	ensureSecretPulledImage = make(map[string]*ensureSecretPulledImageDigest)
+}
```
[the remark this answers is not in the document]
mikebrow: me too :-) been in draft for years don't remember why I started it this way.. fixing

pkg/kubelet/images/image_manager.go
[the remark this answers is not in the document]
mikebrow: per discussion moving into manager instance...

pkg/kubelet/images/image_manager.go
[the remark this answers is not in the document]
mikebrow: good point, will secure reads and writes to this map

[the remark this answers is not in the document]
mikebrow: it's synchronous for each pod.. but each pod being run is on it's own worker thread...
``` go func(pod *v1.Pod) {
			err := kl.runPod(pod, retryDelay)
			ch <- RunPodResult{pod, err}
		}(pod)
```

pkg/kubelet/container/helpers.go
mikebrow: I could add a TODO if you like

pkg/kubelet/images/image_manager.go
mikebrow: going to sleep on this..

pkg/kubelet/images/image_manager.go
pacoxu: Two special cases:

1. pinned image:
https://github.com/kubernetes/kubernetes/blob/6f1612dcae9eddd3e12223d417662971c63952fa/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1/api.pb.go#L6067-L6070

If the image is pinned, we should skip the check. (For daemonset or something, we may use pinned.)

2. kubelet restart:
After kubelet restarts, can we check pod's status.containerStatuses[x].imageID to compare the sha256 of the image with already pulled image. If they are the same, we can assume the image is pulled by kubelet with secret.

pkg/kubelet/container/helpers.go
JAORMX: I think changing to sha256 would be ideal (or even making this configurable to future-proof this?)

deads2k:
> not persisted across kublet restart, is in memory only.. TODOs/discussions in kep and earlier discussions that we'd start persisting and have a lifecycle post alpha / the next KEP for this

This looks like a significant problem, even in alpha.  Not being persisted means that the mechanism defies consumer expectations, because images that are protected five minutes ago will not be protected after the kubelet restarts.  This means that a cluster-admin enabling this feature cannot count on it to protect his images.  Being able to distinguish between images pulled by the kubelet and images not pulled by the kubelet across kubelet restarts is needed for predictability.

Additionally, the feature seems more likely to function as desired if the hash of credentials that are allowed to pull an image is stored on disk so that knowledge of cached access lasts across kubelet restarts.  I see this as less critical, since identification of "kubelet at some point pulled this image" is sufficient to require a re-check of authorization and allows failing in the safe direction.

mikebrow:
> > not persisted across kublet restart, is in memory only.. TODOs/discussions in kep and earlier discussions that we'd start persisting and have a lifecycle post alpha / the next KEP for this
> 
> This looks like a significant problem, even in alpha. Not being persisted means that the mechanism defies consumer expectations, because images that are protected five minutes ago will not be protected after the kubelet restarts. This means that a cluster-admin enabling this feature cannot count on it to protect his images. Being able to distinguish between images pulled by the kubelet and images not pulled by the kubelet across kubelet restarts is needed for predictability.
> 
> Additionally, the feature seems more likely to function as desired if the hash of credentials that are allowed to pull an image is stored on disk so that knowledge of cached access lasts across kubelet restarts. I see this as less critical, since identification of "kubelet at some point pulled this image" is sufficient to require a re-check of authorization and allows failing in the safe direction.

Yes, even if we had bullet proof persistence for the history of all k8s pulled images, we'd still need to resolve how to handle Pre-loaded images (used for pull never policy and disconnected scenarios) and images loaded into container runtime caches for the k8s.io namespace at runtime. We need an auth policy for the images to be negotiated on with the container runtime. Otherwise someone could simply copy an image, stop kubelet, load the image, start kubelet. Thus breaking our persistence cache.  Additionally there needs to be cache policies for a resolution between container runtime cache garbage collection and the orchestrator garbage collection, and for requiring re-auth, and for sharing access or not (in and out of a particular namespace, by user/owner, ...), and for encryption, and ...  

As you point out we also need a better credentials/auth design over the current image pull secrets design.

If we had these things.. this Ensure Image code would be working with the container runtime to determine if a full pull, auth or re-auth is required based on cache policies, lease contracts, or what not for the desired images (possibly to the detail of manifests, layers, and current lazy downloads of same.. ).  Either way we need to begin to build and improve a framework for Ensuring Images.  The use pull always answer.. needs improvement. 

At least with this start the admin has some control over pull if not present outside forcing all pull if not present policies to pull always, if he manages (garbage collects) the pre-loaded images on a reboot of kubelet. Agreed, it is not the end-game.

pkg/kubelet/images/image_manager.go
mikebrow: Should discuss. There may be a way we can overload / exploit the new pinned feature for further cases, yes.. 

The sha of the image isn't a link to authentication that was used pull it.. or permission for another user to access it by sha.

pkg/kubelet/images/image_manager.go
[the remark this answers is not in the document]
mikebrow: @liggit added a lock that protects the current reads and writes .. will redo it with a more formal api when we figure out the persistence solution going forward, either through cri or a common k8s persistence mechanism...

pkg/kubelet/container/helpers.go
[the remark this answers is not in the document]
mikebrow: agree with the later points here that we will want to keep it in another alpha cycle to cover this and the other issues around integration with container runtime cache, expirations for auth persistence, additional pull/cache policies, ...

liggitt:
I tend to agree with @deads2k that an implementation that doesn't actually protect previously-pulled-with-credentials images if the kubelet restarts defies consumer expectations. I'm skeptical we should defer design of a solution to that part of the problem and merge this without it. In many ways, "best effort" security features are worse than no security, because people don't realize they're not actually protected.

mikebrow:
> I tend to agree with @deads2k that an implementation that doesn't actually protect previously-pulled-with-credentials images if the kubelet restarts defies consumer expectations. I'm skeptical we should defer design of a solution to that part of the problem and merge this without it. In many ways, "best effort" security features are worse than no security, because people don't realize they're not actually protected.

nod.. 

summarizing 
1) Current behavior/behavior with feature gate off (default): force (via admission controller) pull always (the current workaround solution for many years) note: I'm not convinced everyone is forcing always pull, just as likely some don't know they have the problem caused by pull if not present when auth is needed for pull or just for private images.
2) suggest admins with interest in getting away from 1, turn the feature gate on for testing but only if they have a process enabled for garbage collecting (on boot of kubelet and/or some daily cycle) all but the never pull images when they allow the pull if not present image pull policy (are not using the admission controller).
3) change how kubelet and the container runtimes manage (ensure) images... lots of ideas, long discussion

sftim:
/retitle Ensure secret pulled images

I think there might be another title that explains this better, without being as long as the release note.

pkg/features/kube_features.go
```diff
@@ -826,6 +826,14 @@ const (
 	//
 	// Stop auto-generation of secret-based service account tokens.
 	LegacyServiceAccountTokenNoAutoGeneration featuregate.Feature = "LegacyServiceAccountTokenNoAutoGeneration"
+
+	// owner: @mikebrow
+	// kep: http://kep.k8s.io/2535
+	// alpha: v1.24
+	//
+	// Enables kubelet to ensure images pulled with pod imagePullSecrets are authenticated
+	// by other pods that do not have the same credentials.
```
sftim, line 835: I don't know what this means and that I can't explain it to other people. I could read the code; I haven't, because I'm checking whether the docs part makes sense.

What effect does this feature gate have?

mikebrow: Described in more detail in the KEP.  The sentence presumes the reader knows what kubelet is, container images, pods, image pull secrets (credentials), and authentication to registries using existing image pull policies and said credentials... and yes, that is a lot of assumptions. There are existing descriptions for the pull if not present, pull always policies, and warnings for using mitigations to force the pull always policy, because kubelet does not currently "ensure images pulled with pod imagePullSecrets are authenticated by other pods that do not have the same credentials" when pull always is not requested. We could adds links. This is a somewhat complex topic regarding currently known potential security issues and subsequent performance issues with current mitigations.  

Suggestions, to simplify the description are more than welcome. 

Looks like we will not be merging in this release... Instead we will add in additional phase II suggested items, such as persistence of the ensure metadata, a gc sync process with container runtime image cache, (possibly) a solution for identifying which container images are loaded to be used as never pull images for the pod pull never container image policy, and a few other TODO items that were proposed for the next phase.

mikebrow: overly simple explanation:
pod spec A has a private image that requires credentials (aka image pull secrets)
pod spec B uses the private image (This pod doesn't need credentials because the image is already cached)

sftim: Could we switch to a revised comment that explains this clearly enough? 
(I'm afraid I'm convinced that the current text falls some way short).

mikebrow: will do

pacoxu:
@mikebrow do you have time to follow this up?

mikebrow:
Missed your ping.. Yes of course I do.

pacoxu:
Some small changes based on [rphillips](https://github.com/rphillips)'s comments
https://github.com/kubernetes/kubernetes/pull/112179/commits/88811d5d366b93bfd7a3b7c4f072e831d784dc65

mikebrow:
@pacoxu here's the persistence change in the kep update.. if still interested in helping :-)

PR #94903, opened 2020-09-18
Take into account the latency incurred in server filters

tkashem:
**What type of PR is this?**

/kind bug

`apiserver_request_duration_seconds` does not take into account the time a request spends in the server filters. If a filter takes longer then the latency incurred will not be reflected in the apiserver latency metrics.
For example, the amount of time a request spends in priority and fairness machinery or in shuffle queues will not be accounted for. 

Take the following approach to resolve this issue:
- Add a server filter that attaches request received timestamp to the request context very early in in the handler chain (as soon as net/http hands over control to apiserver).
- Use the above received timestamp in the apiserver latency metrics `apiserver_request_duration_seconds`.
- Use the above received timestamp in the audit layer to set `RequestReceivedTimestamp`.

I can noticeably reproduce this issue by adding a server filter that adds artificial delay of let's say `20s` to every request originating from a certain user. The two snapshots of `apiserver_request_duration_seconds` show before and after this change. The before snapshot does not show the latency incurred in the above filter. The after snapshot clearly shows the latency incurred.

Before:
image

After:
image

```release-note
Both apiserver_request_duration_seconds metrics and RequestReceivedTimestamp field of an audit event take 
into account the time a request spends in the apiserver request filters. 
```


staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go
```diff
@@ -368,8 +368,12 @@ func MonitorRequest(req *http.Request, verb, group, version, resource, subresour
 // InstrumentRouteFunc works like Prometheus' InstrumentHandlerFunc but wraps
 // the go-restful RouteFunction instead of a HandlerFunc plus some Kubernetes endpoint specific information.
 func InstrumentRouteFunc(verb, group, version, resource, subresource, scope, component string, deprecated bool, removedRelease string, routeFunc restful.RouteFunction) restful.RouteFunction {
-	return restful.RouteFunction(func(request *restful.Request, response *restful.Response) {
+	return restful.RouteFunction(func(req *restful.Request, response *restful.Response) {
 		now := time.Now()
+		requestReceivedTimestamp, ok := request.ReceivedTimestampFrom(req.Request.Context())
+		if ok {
+			now = requestReceivedTimestamp
```
wojtek-t, line 375: nit: let's rename now to startTime - it will be less misleading.

staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go
```diff
@@ -384,16 +388,20 @@ func InstrumentRouteFunc(verb, group, version, resource, subresource, scope, com
...
 	})
 }
 
 // InstrumentHandlerFunc works like Prometheus' InstrumentHandlerFunc but adds some Kubernetes endpoint specific information.
 func InstrumentHandlerFunc(verb, group, version, resource, subresource, scope, component string, deprecated bool, removedRelease string, handler http.HandlerFunc) http.HandlerFunc {
 	return func(w http.ResponseWriter, req *http.Request) {
 		now := time.Now()
+		requestReceivedTimestamp, ok := request.ReceivedTimestampFrom(req.Context())
+		if ok {
+			now = requestReceivedTimestamp
```
wojtek-t, line 403: same here

wojtek-t (commented):
This looks reasonable to me, but I would like someone else to also take a look.

wojtek-t:
@kubernetes/sig-scalability - this may potentially affect scalability tests

wojtek-t:
LGTM - thanks!
@sttts @deads2k - for another pair of eyes

[I don't expect it to go up visibly up, but we will see for sure]

staging/src/k8s.io/apiserver/pkg/endpoints/request/received_time.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"context"
+	"time"
+)
+
+type requestReceivedTimestampKeyType int
+
+// requestReceivedTimestampKey is the ReceivedTimestamp (the time the request reached the apiserver)
+// key for the context.
+const requestReceivedTimestampKey requestReceivedTimestampKeyType = iota
```
sttts, line 28: I don't think iotas are different in different packages. A good trick is to define a custom empty struct type and use that as a key.

staging/src/k8s.io/apiserver/pkg/server/handler.go
```diff
@@ -186,5 +187,9 @@ func serviceErrorHandler(s runtime.NegotiatedSerializer, serviceErr restful.Serv
 
 // ServeHTTP makes it an http.Handler
 func (a *APIServerHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	a.FullHandlerChain.ServeHTTP(w, r)
+	// This is the entry point where net/http delegates to our handler.
+	// Wire the "request received timestamp" handler here.
+	handler := filters.WithRequestReceivedTimestamp(a.FullHandlerChain)
```
sttts, line 192: this belongs in to the default handler chain constructor.

staging/src/k8s.io/apiserver/pkg/server/config.go
```diff
@@ -691,6 +691,10 @@ func DefaultBuildHandlerChain(apiHandler http.Handler, c *Config) http.Handler {
 	handler = genericapifilters.WithWarningRecorder(handler)
 	handler = genericapifilters.WithCacheControl(handler)
 	handler = genericfilters.WithPanicRecovery(handler)
+
+	// Wire the "request received timestamp" handler here so that we can attach the ReceivedTimestamp
+	// in the request context very early in the handler chain.
+	handler = genericapifilters.WithRequestReceivedTimestamp(handler)
```
sttts, line 697: no need for the comment IMO.

staging/src/k8s.io/apiserver/pkg/endpoints/request/received_time.go
sttts: nvmd. The type cast to requestReceivedTimestampKeyType makes it unique.

PR #94905, opened 2020-09-18
[WIP] Raise error if the user has multiple contexts but no current set

### LINKED ISSUE #936: Unexpected behavior when using multiple contexts in kubeconfig
When using a config with multiple contexts, one might see a surprising message if there is no context set.
Let me demostrate:
```
$ k config get-contexts 
CURRENT   NAME               CLUSTER            AUTHINFO         NAMESPACE
          coldsweet          microk8s-cluster   admin-microk8s   coldsweet
          devops             devops-calico      admin            
          devops-cilium      devops-cilium      admin-cilium     
*         microk8s-context   microk8s-cluster   admin-microk8s   
```

Here, I have a context set. So `kubectl get pods` or any other command works fine.
If there is no current-context, kubectl says it can't connect to localhost (that's reasonable if you know k8s, more on that below).
```
sed -i '/current-context/d' ~/.kube/config
 $ k config get-contexts 
CURRENT   NAME               CLUSTER            AUTHINFO         NAMESPACE
          coldsweet          microk8s-cluster   admin-microk8s   coldsweet
          devops             devops-calico      admin            
          devops-cilium      devops-cilium      admin-cilium     
          microk8s-context   microk8s-cluster   admin-microk8s   
$ k get pods
The connection to the server localhost:8080 was refused - did you specify the right host or port?
```
The above message is baffling if you encounter this for the first time. A better message would be something like this:
```
Mutliple contexts found but no context is set. Which cluster shall I use? Please set an active context
```
**Why is this needed**:
I give Kubernetes workshops, where students use multiple clusters. As we go through setting `~/.kube/confing` for usage with multiple clusters, some students will rush to try stuff, and forget to set a context. This results in frustration and funny facial  expressions staring at the screen. We explain that `kubectl`, uses localhost per default - and that older versions of kubernetes where not secured on localhost. A much better UI will be simply telling the user he didn't set up a context.

I'm willing to send a PR for improving this UI. This should not be hard to fix (I have already submitted code to k8s/k8s).

/sig usability

oz123:
The current behavior is to ignore this and try to connect to `localhost:8080`.
[This is quite confusing, as reported in this issue][1].

This commit improves the UX by telling the users they need to explicitly choose a context, if there are multiple contexts:

```
$ kubectl get pods
error: Multiple contexts found. However, no current-context is set in the configuration
```

[1]: https://github.com/kubernetes/kubectl/issues/936

**What type of PR is this?**
/kind bug

This is an attept to improve a confusing error message. Please see the issue linked.

Fixes https://github.com/kubernetes/kubectl/issues/936

I am aware that this PR is still missing test cases. However, I'm publishing this to solicit feedback, and I do not expect this to be merged at this state for the reasons explained below.
With the current patch, if you have not set a context, you see a clear error message. However, you can now only use an editor to fix the issue. That is because this patch disables the ability to use `kubeconfig config set-context`.
```
$ ./kubectl config use-context foo
error: Multiple contexts found. However, no current-context is set in the configuration
```
Users who are a bit familiar with the shell can circumvent the problem:
```
$ echo "current-context: foo" >> ~/.kube/config
```

IMHO, this would be a better interface to exclude the `config` sub-command from producing this error. There are two possible solutions I could think of:
1. Pass the sub-command to the config validator. 
2. Check for the number of contexts in each individual command, e.g. modify all entry points in `kubernetes/vendor/k8s.io/kubectl/pkg/cmd`.

I haven't implemented any yet. It's quite obvious that both approaches make the PR significantly larger. The latter approach makes even larger than the former approach.

Maybe there is another way I have not thought of or not aware of?
I would be happy to get some feedback.

In any case, once I know which approach is preferred I will add the required changes and include tests.

```release-note
kubectl will now raise an error if the user has multiple contexts but no current set.
Previously, it simply tried to connect to localhost:8080.
```

### DIFF
--- a/staging/src/k8s.io/client-go/tools/clientcmd/config.go
+++ b/staging/src/k8s.io/client-go/tools/clientcmd/config.go
@@ -101,6 +101,9 @@ func (o *PathOptions) GetStartingConfig() (*clientcmdapi.Config, error) {
 		return clientcmdapi.NewConfig(), nil
 	}
 	if err != nil {
+		if errors.Is(err, ErrNoContextSet) {
+			return &rawConfig, err
+		}
 		return nil, err
 	}
 
@@ -164,7 +167,7 @@ func NewDefaultPathOptions() *PathOptions {
 // (no nil strings), we're forced have separate handling for them.  In the kubeconfig cases, newConfig should have at most one difference,
 // that means that this code will only write into a single file.  If you want to relativizePaths, you must provide a fully qualified path in any
 // modified element.
-func ModifyConfig(configAccess ConfigAccess, newConfig clientcmdapi.Config, relativizePaths bool) error {
+func ModifyConfig(configAccess ConfigAccess, newConfig clientcmdapi.Config, relativizePaths bool, isSetContext bool) error {
 	if UseModifyConfigLock {
 		possibleSources := configAccess.GetLoadingPrecedence()
 		// sort the possible kubeconfig files so we always "lock" in the same order
@@ -179,8 +182,13 @@ func ModifyConfig(configAccess ConfigAccess, newConfig clientcmdapi.Config, rela
 	}
 
 	startingConfig, err := configAccess.GetStartingConfig()
+
 	if err != nil {
-		return err
+		if isSetContext && errors.Is(err, ErrNoContextSet) {
+
+		} else {
+			return err
+		}
 	}
 
 	// We need to find all differences, locate their original files, read a partial config to modify only that stanza and write out the file.
@@ -191,8 +199,14 @@ func ModifyConfig(configAccess ConfigAccess, newConfig clientcmdapi.Config, rela
 	}
 
 	if startingConfig.CurrentContext != newConfig.CurrentContext {
-		if err := writeCurrentContext(configAccess, newConfig.CurrentContext); err != nil {
-			return err
+		if err := writeCurrentContext(configAccess, newConfig.CurrentContext, isSetContext);
+
+		err != nil {
+			if isSetContext && errors.Is(err, ErrNoContextSet) {
+
+			} else {
+				return err
+			}
 		}
 	}
 
@@ -371,7 +385,7 @@ func (p *persister) Persist(config map[string]string) error {
 	authInfo, ok := newConfig.AuthInfos[p.user]
 	if ok && authInfo.AuthProvider != nil {
 		authInfo.AuthProvider.Config = config
-		ModifyConfig(p.configAccess, *newConfig, false)
+		ModifyConfig(p.configAccess, *newConfig, false, false)
 	}
 	return nil
 }
@@ -380,9 +394,12 @@ func (p *persister) Persist(config map[string]string) error {
 // If newCurrentContext is the same as the startingConfig's current context, then we exit.
 // If newCurrentContext has a value, then that value is written into the default destination file.
 // If newCurrentContext is empty, then we find the config file that is setting the CurrentContext and clear the value from that file
-func writeCurrentContext(configAccess ConfigAccess, newCurrentContext string) error {
-	if startingConfig, err := configAccess.GetStartingConfig(); err != nil {
-		return err
+func writeCurrentContext(configAccess ConfigAccess, newCurrentContext string, isSetContext bool) error {
+	startingConfig, err := configAccess.GetStartingConfig()
+	if err != nil {
+		if !(isSetContext == true && errors.Is(err, ErrNoContextSet)){
+			return err
+		}
 	} else if startingConfig.CurrentContext == newCurrentContext {
 		return nil
 	}
--- a/staging/src/k8s.io/client-go/tools/clientcmd/loader.go
+++ b/staging/src/k8s.io/client-go/tools/clientcmd/loader.go
@@ -244,6 +244,9 @@ func (rules *ClientConfigLoadingRules) Load() (*clientcmdapi.Config, error) {
 	mergo.MergeWithOverwrite(config, mapConfig)
 	mergo.MergeWithOverwrite(config, nonMapConfig)
 
+	if (len(config.Contexts) > 1 && config.CurrentContext == "") {
+		errlist = append(errlist, ErrNoContextSet)
+	}
 	if rules.ResolvePaths() {
 		if err := ResolveLocalPaths(config); err != nil {
 			errlist = append(errlist, err)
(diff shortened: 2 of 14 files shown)


dougsland:
looks sane to me. Just wondering why the title is WIP? Still more progress to made related to this RFE? @oz123

oz123:
@dougsland well, it's WIP, because I think there is still some work to do.
For the following reasons:
 * I didn't add any tests
 * I was not sure about the implementation, hence I was thinking out loud in my comment.

I am OK if it's merged like this, but I'm not sure it will be accepted.

dougsland:
> @dougsland well, it's WIP, because I think there is still some work to do.
> For the following reasons:
> 
> * I didn't add any tests
> * I was not sure about the implementation, hence I was thinking out loud in my comment.

I agree we might useful to add a test here.
 
> I am OK if it's merged like this, but I'm not sure it will be accepted.

I can't merge patches but let me try to dig here and help moving this one forward.

dougsland:
> > @dougsland well, it's WIP, because I think there is still some work to do.
> > For the following reasons:
> > 
> > * I didn't add any tests
> > * I was not sure about the implementation, hence I was thinking out loud in my comment.
> 
> I agree we might useful to add a test here.
> 
> > I am OK if it's merged like this, but I'm not sure it will be accepted.
> 
> I can't merge patches but let me try to dig here and help moving this one forward.

Hi @oz123, 

Okay, I think you didn't get attention deserved here because this is the staging tree under kubernetes git repo (See README from https://github.com/kubernetes/kubernetes/tree/master/staging). 

Please submit your PR to client-go tree: https://github.com/kubernetes/client-go
The tests related to clientcmd are available to look via: validation*.go in the same dir you made the changes.

Next step for me is give a try in your patch. Please let me know if you need any additional help and the new PR url (client-go tree).

staging/src/k8s.io/client-go/tools/clientcmd/loader.go
```diff
@@ -244,11 +244,15 @@ func (rules *ClientConfigLoadingRules) Load() (*clientcmdapi.Config, error) {
...
 
+	if (len(config.Contexts) > 1 && config.CurrentContext == "") {
+		errlist = append(errlist, ErrNoContextSet)
+	}
 	if rules.ResolvePaths() {
 		if err := ResolveLocalPaths(config); err != nil {
 			errlist = append(errlist, err)
 		}
 	}
+
```
dougsland, line 255: just nit: no need to add extra space.

dougsland (commented):
just double checked, this is the correct repo to send the PR, my bad. The client-go is a mirror.

dougsland:
@ahmetb might help here.

oz123:
@douglasland thanks 😊 you beat me to respond. I was trying to comment on this.

corneliusweig:
It's not an option to break `kubectl config set-context`. Maybe you can inspect the error list that is returned from the loader in the config subcommand and filter out the error you added here?

staging/src/k8s.io/client-go/tools/clientcmd/loader.go
oz123: Thanks. This is fixed now.

staging/src/k8s.io/client-go/tools/clientcmd/config.go
```diff
@@ -191,8 +199,14 @@ func ModifyConfig(configAccess ConfigAccess, newConfig clientcmdapi.Config, rela
 	}
 
 	if startingConfig.CurrentContext != newConfig.CurrentContext {
-		if err := writeCurrentContext(configAccess, newConfig.CurrentContext); err != nil {
-			return err
+		if err := writeCurrentContext(configAccess, newConfig.CurrentContext, isSetContext);
+
+		err != nil {
+			if isSetContext && errors.Is(err, ErrNoContextSet) {
+
```
corneliusweig, line 206: empty if blocks are unidiomatic in Go. Try to refactor.

staging/src/k8s.io/client-go/tools/clientcmd/config.go
```diff
@@ -179,8 +182,13 @@ func ModifyConfig(configAccess ConfigAccess, newConfig clientcmdapi.Config, rela
 	}
 
 	startingConfig, err := configAccess.GetStartingConfig()
+
 	if err != nil {
-		return err
+		if isSetContext && errors.Is(err, ErrNoContextSet) {
```
corneliusweig, line 187: You need to be careful with `errors.Is()`. It returns true if at least one of the reported errors is the respective instance. However, please think about what should happen, if multiple errors are returned in the aggregate.

corneliusweig (commented):
Your change has led to some fundamental changes. I think it's better to iterate on it via an issue than a PR to avoid wasting time. The go-to person for these questions are @.soltysh or @.seans3.

staging/src/k8s.io/client-go/tools/clientcmd/config.go
oz123: My intent here was to keep the old behavior. Other errors should be reported as before. I added `isSetContext`, so one can still use `set-context`. I tried explaining this in the commit message.
Without this, the check I added caused an error on all the commands.
I will need some time to think of an alternative without `isSetContext`.

corneliusweig: Maybe I missed some detail why this is necessary. However, my main point is to discuss this change in an issue before investing more time into the implementation. It's possible that there is a reason for the current behavior that I don't know.

oz123: I understand that the change should be discussed in an issue first. This is why I created one and initially sent a very minimal patch. 
Can the discussion for this issue happen on the original issue I opened in the kubectl repository or should I open a new issue in this repository?

staging/src/k8s.io/client-go/tools/clientcmd/config.go
```diff
@@ -164,7 +167,7 @@ func NewDefaultPathOptions() *PathOptions {
 // (no nil strings), we're forced have separate handling for them.  In the kubeconfig cases, newConfig should have at most one difference,
 // that means that this code will only write into a single file.  If you want to relativizePaths, you must provide a fully qualified path in any
 // modified element.
-func ModifyConfig(configAccess ConfigAccess, newConfig clientcmdapi.Config, relativizePaths bool) error {
+func ModifyConfig(configAccess ConfigAccess, newConfig clientcmdapi.Config, relativizePaths bool, isSetContext bool) error {
```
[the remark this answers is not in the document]
soltysh: This is changing external API in a library we provide, you can't do that b/c you're breaking all the existing users.

soltysh (changes requested):
The rest of the comments from Cornelius still hold.

PR #91513, opened 2020-05-28
Upgraded aws-sdk-go dependency to v1.35.5 to support new regions like Milan

jqmichael:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Upgraded aws-sdk-go to v1.35.5 to include new regions, i.e. Millan

**Which issue(s) this PR fixes**:

Fixes #
NONE

**Special notes for your reviewer**:
Commands to upgrade the version.
```
./hack/pin-dependency.sh github.com/aws/aws-sdk-go v1.35.5
./hack/update-vendor.sh
./hack/lint-dependencies.sh

// Additional changes required by lint-dependencies
hack/pin-dependency.sh github.com/jmespath/go-jmespath v0.4.0
./hack/update-vendor.sh
./hack/lint-dependencies.sh
```

**Does this PR introduce a user-facing change?**:
NONE

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:
NONE

```release-note
NONE
```

PR #94909, opened 2020-09-19
Update aws-sdk-go to v1.34.27 to support new regions

hakman:
**What type of PR is this?**
/kind feature

Update aws/aws-sdk-go to v1.34.27 to support new regions.
Supersedes https://github.com/kubernetes/kubernetes/pull/91513, that was never finished.

n/a

```bash
hack/pin-dependency.sh github.com/aws/aws-sdk-go v1.34.27
hack/update-vendor.sh
hack/lint-dependencies.sh

#required by lint-dependencies
hack/pin-dependency.sh github.com/jmespath/go-jmespath v0.3.0
./hack/update-vendor.sh
./hack/lint-dependencies.sh
hack/pin-dependency.sh github.com/stretchr/testify v1.5.1
./hack/update-vendor.sh
./hack/lint-dependencies.sh
```

```release-note
NONE
```

```docs

```


go.mod
```diff
@@ -264,6 +264,7 @@ replace (
 	github.com/go-openapi/swag => github.com/go-openapi/swag v0.19.5
 	github.com/go-openapi/validate => github.com/go-openapi/validate v0.19.5
 	github.com/go-ozzo/ozzo-validation => github.com/go-ozzo/ozzo-validation v3.5.0+incompatible
+	github.com/go-sql-driver/mysql => github.com/go-sql-driver/mysql v1.5.0
```
liggitt, line 267: why are we adding a mysql driver to our dependency tree?

liggitt: looks like this was added in https://github.com/aws/aws-sdk-go/pull/3118/files#diff-37aff102a57d3d7b797f152915a6dc16R4 in v1.29.0+

this is a good example of the types of dependencies we don't want to include in the kubernetes/kubernetes dependency tree

liggitt: It looks like this is only used in an example file inside the aws repo here:

https://github.com/aws/aws-sdk-go/blob/master/example/service/rds/rdsutils/authentication/iam_authentication.go#L11

The AWS library should isolate that example to prevent all downstream consumers of the AWS sdk from inheriting a mysql driver dependency. Possible ways to do that:

* rename the authentication examples folder to `_authentication`, then run `go mod tidy` on the root (prevents inheriting mysql dependency, unblocking this PR)
* rename the example folder to to `_example`, then run `go mod tidy` on the root (prevents inheriting any example dependencies, which seems ideal)
* create a go.mod file inside the authentication or examples folder to isolate that into a separate module

hakman: Is there any way around this, or the best that could be done is to use the latest v1.28.x?

liggitt: The AWS library maintainers have to fix this, reported in https://github.com/aws/aws-sdk-go/issues/3548

liggitt: > Is there any way around this, or the best that could be done is to use the latest v1.28.x?

it looks like v1.28.14 is the latest that can be used that does not have this issue

hakman: Thanks. Will wait and see what the aws-sdk-go maintainers say.

hakman:
https://github.com/kubernetes/kubernetes/pull/91513 was updated and approved. This is no longer needed.
/close

PR #94911, opened 2020-09-19
Enhance log information of verifyRunAsNonRoot, add pod, container information

### LINKED ISSUE #94898: container has runAsNonRoot and image will run as root warning should include container name
when deploying a container that has runAsNonRoot but runs as root this error appears, but for pods with lots of containers it can be confusing which container failed, especially if there are 3-rd party webhooks that inject containers
please add the container name to the message, making it `container "foo" has runAsNonRoot and image will run as root`

wawa0210:
**What type of PR is this?**

/kind cleanup

If the pod contains multiple containers, the prompt message here cannot quickly determine which container `verifyRunAsNonRoot` has the problem. 

Therefore, add information about pod and container to the log for quick positioning

Fixes #94898

```release-note
Enhance log information of verifyRunAsNonRoot, add pod, container information
```

```docs

```

### DIFF
--- a/pkg/kubelet/kuberuntime/security_context_others.go
+++ b/pkg/kubelet/kuberuntime/security_context_others.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 
 	"k8s.io/api/core/v1"
+	"k8s.io/kubernetes/pkg/kubelet/util/format"
 	"k8s.io/kubernetes/pkg/securitycontext"
 )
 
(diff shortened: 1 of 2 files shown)


grosser:
adding the container sounds right, I'm not sure if we need the pod since that might be already in the message/be the object of the event, but I bet there are cases where it's not, so should be a good addition (and in doubt better have more info to debug) 👍

pkg/kubelet/kuberuntime/security_context_others.go
```diff
@@ -35,16 +35,16 @@ func verifyRunAsNonRoot(pod *v1.Pod, container *v1.Container, uid *int64, userna
 
 	if effectiveSc.RunAsUser != nil {
 		if *effectiveSc.RunAsUser == 0 {
-			return fmt.Errorf("container's runAsUser breaks non-root policy")
+			return fmt.Errorf("container's runAsUser breaks non-root policy (pod: %s, container: %s)", pod.Name, container.Name)
```
SergeyKanzhelev, line 38: ass pod namespace. `format.Pod` would be the best

SergeyKanzhelev: same below

tallclair: We have a standard utility for this: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/util/format/pod.go#L32

SergeyKanzhelev (approved):
/lgtm

pkg/kubelet/kuberuntime/security_context_others.go
wawa0210: done~

krmayankk: nice , didnt know about this . Curious why is this utility in pkg/kubelet, it can be useful outside of kubelet as well

SergeyKanzhelev: > nice , didnt know about this . Curious why is this utility in pkg/kubelet, it can be useful outside of kubelet as well

I think the idea is to eventually switch to structured logging where this would be solved out of the box

wawa0210:
@sjenning @ yujuhong

PR #94915, opened 2020-09-19
fix kube-proxy cleanup

### LINKED ISSUE #94914: kube-proxy cleanup not work
**What happened**:

Currently kube-proxy cleanup not work, which will left iptables (and ipvs) rules in host and.

```
./_output/bin/kube-proxy --cleanup
W0919 16:21:16.970463   11908 server.go:226] WARNING: all flags other than --config, --write-config-to, and --cleanup are deprecated. Please begin using a config file ASAP.
I0919 16:21:16.970684   11908 server.go:555] Neither kubeconfig file nor master URL was specified. Falling back to in-cluster config.
F0919 16:21:16.970703   11908 server.go:495] unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
```

Kube-proxy cleanup can work and not relay on hostName or ClientConnection config.

`kube-proxy --cleanup`

Lion-Wei:
**What type of PR is this?**

/kind bug

Fix kube-proxy cleanup.

Fixes #94914

kube-proxy replay ip protocol to init iptables interface, and need hostName and ClientConnectionConfiguration to decide ip protocol, so user need specify those config in cleanup.

Cleanup action should clean all rules that kube-proxy created, include IPv4 and IPv6 rules, and user don't have to specify any other config.

```release-note
NONE
```

### DIFF
--- a/cmd/kube-proxy/app/server.go
+++ b/cmd/kube-proxy/app/server.go
@@ -801,11 +801,20 @@ func getConntrackMax(config kubeproxyconfig.KubeProxyConntrackConfiguration) (in
 	return 0, nil
 }
 
-// CleanupAndExit remove iptables rules and exit if success return nil
+// CleanupAndExit remove iptables rules and ipset/ipvs rules in ipvs proxy mode
+// and exit if success return nil
 func (s *ProxyServer) CleanupAndExit() error {
-	encounteredError := userspace.CleanupLeftovers(s.IptInterface)
-	encounteredError = iptables.CleanupLeftovers(s.IptInterface) || encounteredError
-	encounteredError = ipvs.CleanupLeftovers(s.IpvsInterface, s.IptInterface, s.IpsetInterface, s.CleanupIPVS) || encounteredError
+	// cleanup IPv6 and IPv4 iptables rules
+	ipts := []utiliptables.Interface{
+		utiliptables.New(s.execer, utiliptables.ProtocolIPv4),
+		utiliptables.New(s.execer, utiliptables.ProtocolIPv6),
+	}
+	var encounteredError bool
+	for _, ipt := range ipts {
+		encounteredError = userspace.CleanupLeftovers(ipt) || encounteredError
+		encounteredError = iptables.CleanupLeftovers(ipt) || encounteredError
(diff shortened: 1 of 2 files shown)


aojea:
I always found confusing this cleanup flag :smile: , let me assign people with much more knowledge on this
/assign @danwinship @thockin

cmd/kube-proxy/app/server.go
```diff
@@ -801,11 +801,20 @@ func getConntrackMax(config kubeproxyconfig.KubeProxyConntrackConfiguration) (in
...
+		utiliptables.New(s.execer, utiliptables.ProtocolIPv6),
+	}
+	var encounteredError bool
+	for _, ipt := range ipts {
+		encounteredError = userspace.CleanupLeftovers(ipt) || encounteredError
+		encounteredError = iptables.CleanupLeftovers(ipt) || encounteredError
+		encounteredError = ipvs.CleanupLeftovers(s.IpvsInterface, ipt, s.IpsetInterface, s.CleanupIPVS) || encounteredError
+	}
 	if encounteredError {
 		return errors.New("encountered an error while tearing down rules")
```
danwinship, line 819: So this will end up logging an error if `ip6tables` is uninstalled (even though there presumably weren't IPv6 iptables rules to clean up in that case). But that will need to be fixed to reconcile this with #94474 anyway.

If this is only going into 1.20 I'd rather hold this until #94474 and then rebase on top of that, but if this was going to be backported to 1.19 then it might be better to just merge this as-is and ignore the somewhat spurious error message in that case...

Lion-Wei: I think cherry-pick to 1.19 is necessary, obviously 1.19 have this problem. 
But currently I have no idea about how to avoid spurious error message in this case, so maybe hold this until #94474 and then rebase on top of that still an option.

danwinship:
I've filed #94971 which I think can replace this.

danwinship:
Actually, it seems like in basically all distros, `iptables` and `ip6tables` are part of the same package, so I think my previous objection was irrelevant.
/lgtm
/approve

ialidzhikov:
@Lion-Wei. @aojea does it makes sense to file cherry-pick of this PR for `release-1.19`?

aojea:
I think it makes sense

PR #94916, opened 2020-09-19
split systemd mount from mount

### LINKED ISSUE #94913: systemd-run will hang and timeout when numbers of pods and volumes schedule on the node
/sig storage
/kind bug
**What happened**:
Our pods will stay creating or slow to running when lots of pods and volume scheduled on one node.
At this time visited systemd will return timeout or full queue:
```-bash-4.2# systemd-run --scope -- mount -t nfs -o ro xxx:yyy /var/mnt/kubelet/pods/11b30ddf-9e38-4509-8306-4fbf9b3d5878/volumes/kubernetes.io~nfs/aihadoop
Failed to start transient scope unit: Rejected: destination has a full message queue, 0 matched rules; type="method_call", sender=":1.240948" (uid=0 pid=1609 comm="systemd-run --scope -- mount -t nfs -o ro slcmfile") interface="org.freedesktop.systemd1.Manager" member="StartTransientUnit" error name="(unset)" requested_reply="0" destination="org.freedesktop.systemd1" (uid=0 pid=1 comm="/usr/lib/systemd/systemd --switched-root --system ")
``` 

```
mount failed: exit status 1\nMounting command: systemd-run\nMounting arguments: --description=Kubernetes transient mount for /var/mnt/kubelet/pods/85670d15-7507-4b7f-b172-6147b4640450/volumes/kubernetes.io~nfs/aivol --scope -- mount -t nfs xxx:yyy /var/mnt/kubelet/pods/85670d15-7507-4b7f-b172-6147b4640450/volumes/kubernetes.io~nfs/aivol\nOutput: Failed to start transient scope unit: Connection timed out\n\n"
```
Systemd run well and pod could be running stablely
Create the deployment scheduled on one node with 5-8 volumes. and scale replicas to 90 wait 60 seconds and scale to 0 wait 60 seconds to 90 repetive.

We tried different kernel of 4.15 and 5.4 all has this issue. And we enlarge the systemd queue and replace the systemd binary also not work.
**Environment**:
- Kubernetes version (use `kubectl version`): 1.15
- Cloud provider or hardware configuration: NA
- OS (e.g: `cat /etc/os-release`): centos 7 and centos 8
- Kernel (e.g. `uname -a`): Linux 4.15.0-43.generic.x86_64 #1 SMP Wed Jun 17 19:05:57 -07 2020 x86_64 x86_64 x86_64 GNU/Linux or Linux 5.4.0-34.generic.x86_64 #1 SMP Wed Jul 1 21:44:10 -07 2020 x86_64 x86_64 x86_64 GNU/Linux
- Install tools: NA
- Network plugin and version (if this is a network-related bug): NA
- Others: NA

smileusd:
**What type of PR is this?**

/kind bug

Fix [issue94913](https://github.com/kubernetes/kubernetes/issues/94913) 
To fix the systemd hang problems

Fixes #94913 

None

```release-note
Change the mount way from systemd to normal mount except ceph and glusterfs intree-volume.
```

```docs
None
```

### DIFF
--- a/pkg/volume/flexvolume/util.go
+++ b/pkg/volume/flexvolume/util.go
@@ -140,7 +140,7 @@ func prepareForMount(mounter mount.Interface, deviceMountPath string) (bool, err
 // Mounts the device at the given path.
 // It is expected that prepareForMount has been called before.
 func doMount(mounter mount.Interface, devicePath, deviceMountPath, fsType string, options []string) error {
-	err := mounter.Mount(devicePath, deviceMountPath, fsType, options)
+	err := mounter.MountSensitiveWithoutSystemd(devicePath, deviceMountPath, fsType, options, nil)
 	if err != nil {
 		klog.Errorf("Failed to mount the volume at %s, device: %s, error: %s", deviceMountPath, devicePath, err.Error())
 		return err
--- a/staging/src/k8s.io/mount-utils/mount_linux.go
+++ b/staging/src/k8s.io/mount-utils/mount_linux.go
@@ -83,11 +83,11 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 	mounterPath := ""
 	bind, bindOpts, bindRemountOpts, bindRemountOptsSensitive := MakeBindOptsSensitive(options, sensitiveOptions)
 	if bind {
-		err := mounter.doMount(mounterPath, defaultMountCommand, source, target, fstype, bindOpts, bindRemountOptsSensitive)
+		err := mounter.doMount(mounterPath, defaultMountCommand, source, target, fstype, bindOpts, bindRemountOptsSensitive, true)
 		if err != nil {
 			return err
 		}
-		return mounter.doMount(mounterPath, defaultMountCommand, source, target, fstype, bindRemountOpts, bindRemountOptsSensitive)
+		return mounter.doMount(mounterPath, defaultMountCommand, source, target, fstype, bindRemountOpts, bindRemountOptsSensitive, true)
 	}
 	// The list of filesystems that require containerized mounter on GCI image cluster
 	fsTypesNeedMounter := map[string]struct{}{
@@ -99,12 +99,37 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 	if _, ok := fsTypesNeedMounter[fstype]; ok {
 		mounterPath = mounter.mounterPath
 	}
-	return mounter.doMount(mounterPath, defaultMountCommand, source, target, fstype, options, sensitiveOptions)
+	return mounter.doMount(mounterPath, defaultMountCommand, source, target, fstype, options, sensitiveOptions, true)
+}
+
+// MountSensitiveWithoutSystemd is the same as MountSensitive() but disable using systemd mount.
+func (mounter *Mounter) MountSensitiveWithoutSystemd(source string, target string, fstype string, options []string, sensitiveOptions []string) error {
+	mounterPath := ""
+	bind, bindOpts, bindRemountOpts, bindRemountOptsSensitive := MakeBindOptsSensitive(options, sensitiveOptions)
+	if bind {
+		err := mounter.doMount(mounterPath, defaultMountCommand, source, target, fstype, bindOpts, bindRemountOptsSensitive, false)
+		if err != nil {
+			return err
+		}
+		return mounter.doMount(mounterPath, defaultMountCommand, source, target, fstype, bindRemountOpts, bindRemountOptsSensitive, false)
+	}
+	// The list of filesystems that require containerized mounter on GCI image cluster
+	fsTypesNeedMounter := map[string]struct{}{
+		"nfs":       {},
+		"glusterfs": {},
+		"ceph":      {},
+		"cifs":      {},
+	}
+	if _, ok := fsTypesNeedMounter[fstype]; ok {
+		mounterPath = mounter.mounterPath
+	}
+	return mounter.doMount(mounterPath, defaultMountCommand, source, target, fstype, options, sensitiveOptions, false)
 }
 
 // doMount runs the mount command. mounterPath is the path to mounter binary if containerized mounter is used.
 // sensitiveOptions is an extension of options except they will not be logged (because they may contain sensitive material)
-func (mounter *Mounter) doMount(mounterPath string, mountCmd string, source string, target string, fstype string, options []string, sensitiveOptions []string) error {
+// systemdMountRequired is an extension of option to decide whether uses systemd mount.
+func (mounter *Mounter) doMount(mounterPath string, mountCmd string, source string, target string, fstype string, options []string, sensitiveOptions []string, systemdMountRequired bool) error {
 	mountArgs, mountArgsLogStr := MakeMountArgsSensitive(source, target, fstype, options, sensitiveOptions)
 	if len(mounterPath) > 0 {
 		mountArgs = append([]string{mountCmd}, mountArgs...)
@@ -112,7 +137,7 @@ func (mounter *Mounter) doMount(mounterPath string, mountCmd string, source stri
 		mountCmd = mounterPath
 	}
 
-	if mounter.withSystemd {
+	if mounter.withSystemd && systemdMountRequired {
 		// Try to run mount via systemd-run --scope. This will escape the
 		// service where kubelet runs and any fuse daemons will be started in a
 		// specific scope. kubelet service than can be restarted without killing
--- a/pkg/volume/awsebs/aws_ebs.go
+++ b/pkg/volume/awsebs/aws_ebs.go
@@ -399,7 +399,7 @@ func (b *awsElasticBlockStoreMounter) SetUpAt(dir string, mounterArgs volume.Mou
 		options = append(options, "ro")
 	}
 	mountOptions := util.JoinMountOptions(options, b.mountOptions)
-	err = b.mounter.Mount(globalPDPath, dir, "", mountOptions)
+	err = b.mounter.MountSensitiveWithoutSystemd(globalPDPath, dir, "", mountOptions, nil)
 	if err != nil {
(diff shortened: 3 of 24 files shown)


staging/src/k8s.io/mount-utils/mount_linux.go
```diff
@@ -81,8 +81,19 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 	// Path to mounter binary if containerized mounter is needed. Otherwise, it is set to empty.
 	// All Linux distros are expected to be shipped with a mount utility that a support bind mounts.
 	mounterPath := ""
+	needSystemdMount := map[string]struct{}{
+		"glusterfs": {},
+		"ceph":      {},
```
jsafrane, line 86: `k8s.io/mount-utils` is used by CSI drivers outside of Kubernetes. This change could break them.

IMO, `Mounter` should have a flag whether the caller/user needs systemd unit. And the default should be to use system-unit to make sure it does not break existing drivers.

@msau42, or can we break the compatibility in 1.20? I haven't found any compatibility promise about staging repos.

msau42: Can we add a new API and pass this in as an option? Prefer not to break existing interfaces.

msau42 (commented):
FWIW I have seen this systemd scaling issue when there are too many mounts on a node (You don't even need PVs, just hundreds of secrets/configmaps will do). Our kernel team triaged and patched systemd to fix it. I will check if the fix has been upstreamed yet.

smileusd:
> FWIW I have seen this systemd scaling issue when there are too many mounts on a node (You don't even need PVs, just hundreds of secrets/configmaps will do). Our kernel team triaged and patched systemd to fix it. I will check if the fix has been upstreamed yet.

Could you provide which version or commit you cherry-pick to fix this problem in systemd?

staging/src/k8s.io/mount-utils/mount_linux.go
smileusd: @jsafrane  @msau42  I have create a pr with new interface name as "SystemdMount" in my company inner pr. But this change default use normal mount if not change any plugin code. It will change user behavior if csi user design a plugin to use "Mount" inter face to implement systemd-run mount. So the way should reversed: add an interface name as "NoSystemdMount" and change all of the intree volume plugins except glusterfs and cephfs(actually the intree ceph-fuse also not use systemd-run mount now).

smileusd:
> Can we add a new API and pass this in as an option? Prefer not to break existing interfaces.
@msau42  Yes, i know it is a hardcode for mount libarary. You mean add options for the `Mount` interface? Or add new interface with the options.

jsafrane:
Sorry about the delay... This PR is heading the right direction, I just don't like the naming. `MountSensitiveWithSystemd(... false)` does not suggest it's actually about skipping systemd.

Can there be `Mounter.SetSystemdUnit(false)` and store the flag in mounter struct, without any changes in `Mount` or `MountSensitive`?

Or even use option pattern:

```go
func NewWithFlags(mounterPath string,  options ...MounterOption) Interface {
    mnt := &Mounter{
		mounterPath: mounterPath,
		withSystemd: detectSystemd(),
	}
    for _, opt := range options {
        opt(mnt)
    }
    return mnt
}

type MounterOption func(*mounter)

func WithoutSystemd(mnt *mounter) {
    mnt.withSystemd=false
}
```

(Be careful with naming, "mount option" has a distinct meaning.)

smileusd:
@jsafrane Thanks, yes, i will change the name.
Your options are good directions, but the `Mounter` is global from kubelet, that means all the plugin will share the options if add mnt.withSystemd=false. If a driver plugin developer doesn't know the context, he/she could be confused for the output. As for other developers know this context, every time call mount() need to set this option also looks complex.

smileusd:
@jsafrane @msau42 PTAl

jsafrane:
I am sorry, you're right, I missed kubelet has one global `Interface`.

`mount.WithoutSystemd(b.mounter).Mount()` does look even worse than `mount.MountSensitiveWithoutSystemd()` (or even `MountSensitiveWithoutSystemService()`. No flags, no configs, just a new method with a clear name.
@msau42 WDYT?

smileusd:
@jsafrane done 
cc  @msau42

staging/src/k8s.io/mount-utils/mount_linux.go
```diff
@@ -99,20 +99,45 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
...
+		mounterPath = mounter.mounterPath
+	}
+	return mounter.doMount(mounterPath, defaultMountCommand, source, target, fstype, options, sensitiveOptions, false)
 }
 
 // doMount runs the mount command. mounterPath is the path to mounter binary if containerized mounter is used.
 // sensitiveOptions is an extension of options except they will not be logged (because they may contain sensitive material)
-func (mounter *Mounter) doMount(mounterPath string, mountCmd string, source string, target string, fstype string, options []string, sensitiveOptions []string) error {
+// isSystemdMount is an extension of option to decide whether uses systemd mount.
+func (mounter *Mounter) doMount(mounterPath string, mountCmd string, source string, target string, fstype string, options []string, sensitiveOptions []string, isSystemdMount bool) error {
```
jsafrane, line 132: minor nit: isSystemdMount -> systemdMountRequired

jsafrane (commented):
lgtm-ish. @msau42 what do you think?

staging/src/k8s.io/mount-utils/mount_linux.go
smileusd: done

smileusd:
@jsafrane @msau42 Does mounter interface changes must block azure e2e test? 
```/home/prow/go/src/sigs.k8s.io/azuredisk-csi-driver/test/e2e/dynamic_provisioning_test.go:264 ```

**Comment by jsafrane** [comment]
Azure flake looks unrelated, probably caused by https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/568

**Comment by smileusd** [comment]
@jsafrane looks not work?

File: staging/src/k8s.io/mount-utils/mount_linux.go [hunk 3]
```diff
@@ -99,20 +99,45 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
...
+	if bind {
+		err := mounter.doMount(mounterPath, defaultMountCommand, source, target, fstype, bindOpts, bindRemountOptsSensitive, false)
+		if err != nil {
+			return err
+		}
+		return mounter.doMount(mounterPath, defaultMountCommand, source, target, fstype, bindRemountOpts, bindRemountOptsSensitive, false)
+	}
+	// The list of filesystems that require containerized mounter on GCI image cluster
+	fsTypesNeedMounter := map[string]struct{}{
+		"nfs":       {},
```
msau42, line 118: Not your change, but we should find a way to make this configurable instead of hardcoded. Let's consider this in a followup. cc @mattcary

msau42 (commented):
lgtm. Slightly worried that we're changing almost all the plugins but let's merge this early to make sure we get enough soak time.

msau42 (commented):
Oh gcepd wasn't changed to not use systemd?

staging/src/k8s.io/mount-utils/mount_linux.go
smileusd: Yes, it is also an issue when i read this code. 
Maybe we need a struct as a argument for mount to enclosure these options like:
```
type struct MountUtilsOptions {
    MountOptions      []string
    SensitiveOptions []string
    MounterOptions  []string
    OtherOptions       []string
}
```

smileusd:
> Oh gcepd wasn't changed to not use systemd?

done and add flexvolume fix

jsafrane:
/lgtm
/approve
@smileusd thanks a lot for this fix and your patience!

jsafrane:
@andyzhangx, e2e part of the last azure test succeeded, but the test is marked as failed: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/94916/pull-kubernetes-e2e-azure-disk-windows/1315893803965485056

pkg/volume/flexvolume/util.go
```diff
@@ -140,7 +140,7 @@ func prepareForMount(mounter mount.Interface, deviceMountPath string) (bool, err
 // Mounts the device at the given path.
 // It is expected that prepareForMount has been called before.
 func doMount(mounter mount.Interface, devicePath, deviceMountPath, fsType string, options []string) error {
-	err := mounter.Mount(devicePath, deviceMountPath, fsType, options)
+	err := mounter.MountSensitiveWithoutSystemd(devicePath, deviceMountPath, fsType, options, nil)
```
msau42, line 143: It's possible flex volumes can use fuse right?

andyzhangx: @msau42 you mean Linux fuse file system? why is this PR related?

msau42: systemd is needed to mount storage backends that spawn fuse daemons. If we think flex volumes could use fuse, then we need to keep it using systemd.

andyzhangx: there is still flexvolume driver that use fuse: https://github.com/Azure/kubernetes-volume-drivers/tree/master/flexvolume/blobfuse,  this will break blobfuse flexvol driver?

jsafrane: `doMount` in flex volume plugin is used only to bind-mount volume into pod-local directory in case a flex driver does not implement its own `mount` call. I assume such driver implements `mountdevice` and it can use systemd as it wants there.

See `-o bind` forcefully added here:
https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/pkg/volume/flexvolume/mounter-defaults.go#L37

I think it is safe to bind mount without systemd.

smileusd: THX @jsafrane yes bind mount not use systemd

Issue #94918, opened 2020-09-19
PreStop e2e node marked as flaky


knabben:
**Which jobs are flaking**:

ci-cri-containerd-e2e-gci-gce-flaky 

**Which test(s) are flaking**:

```
Kubernetes e2e suite.[k8s.io] [sig-node] PreStop graceful pod terminated should wait until preStop hook completes the process [Flaky]
```

**Testgrid link**:

https://testgrid.k8s.io/sig-node-containerd#e2e-gci-flaky

**Reason for failure**:

The intent is to test, the following documented behavior: 

```
If the pod has defined a preStop hook, it is invoked inside of the pod.
If the preStop hook is still running after the grace period expires, 
step 2 is then invoked with a small (2 second) extended grace period
```

A *time.Sleep* is used to hold the graceful period, after that a wait.Poll is used to check if the pod still exists, in some situations the pod is killed before the polling check.

PR #83792, opened 2019-10-11
Mark 'wait until preStop hook completes the process' flaky

liggitt:
**What type of PR is this?**
/kind flake

**What this PR does / why we need it**:
The error in the prestop test was not getting checked. When a staticcheck fix PR started checking it, the test started flaking badly (xref https://github.com/kubernetes/kubernetes/pull/83681#issuecomment-540826068). This marks it flaky so sig-node can investigate/correct the test.

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

/sig node
/cc @BenTheElder @Random-Liu

PR #94922, opened 2020-09-20
Verify pod termination with E2E PreStop hook

knabben:
**What type of PR is this?**
/kind cleanup
/kind flake
/sig node
/sig testing

This PR de-flaky the PreStop E2E node test by reducing the wait time period for checking the preStop hook test is being executed.

Related #94918

```release-note
NONE
```

### DIFF
--- a/test/e2e/node/pre_stop.go
+++ b/test/e2e/node/pre_stop.go
@@ -187,10 +187,11 @@ var _ = SIGDescribe("PreStop", func() {
 		err = podClient.Delete(context.TODO(), pod.Name, *metav1.NewDeleteOptions(gracefulTerminationPeriodSeconds))
 		framework.ExpectNoError(err, "failed to delete pod")
 
-		//wait up to graceful termination period seconds
-		time.Sleep(30 * time.Second)
+		// wait for less than the gracePeriod termination ensuring the
+		// preStop hook is still executing.
+		time.Sleep(15 * time.Second)
 
-		ginkgo.By("verifying the pod running state after graceful termination")
+		ginkgo.By("verifying the pod is running while in the graceful period termination")
 		result := &v1.PodList{}
 		err = wait.Poll(time.Second*5, time.Second*60, func() (bool, error) {
 			client, err := e2ekubelet.ProxyRequest(f.ClientSet, pod.Spec.NodeName, "pods", ports.KubeletPort)


test/e2e/node/pre_stop.go
```diff
@@ -181,8 +179,8 @@ var _ = SIGDescribe("PreStop", func() {
 		testPreStop(f.ClientSet, f.Namespace.Name)
 	})
 
-	ginkgo.It("graceful pod terminated should wait until preStop hook completes the process [Flaky]", func() {
```
SergeyKanzhelev, line 184: this PR effectively change what the test is validating. Original one was testing that if gracePeriod allows it and the pod preStop hook is configured to be long enough, pod will be running that preStop hook for the whole duration. New test checks that pod will not exist after grace period. Which doesn't confirm that preStop hook was executed. Is it correct understanding or a proposed change?

test/e2e/node/pre_stop.go
```diff
@@ -201,28 +199,8 @@ var _ = SIGDescribe("PreStop", func() {
 		err = podClient.Delete(context.TODO(), pod.Name, *metav1.NewDeleteOptions(gracefulTerminationPeriodSeconds))
 		framework.ExpectNoError(err, "failed to delete pod")
 
-		//wait up to graceful termination period seconds
-		time.Sleep(30 * time.Second)
```
SergeyKanzhelev, line 205: should the original test be fixed by changing this timeout to 15 second? 30 seconds is on the edge of graceful termination timeout and pod MAY or MAY NOT run at this ponit. After 15 seconds is is still running for sure if preStop hook is being executed.

test/e2e/node/pre_stop.go
knabben: @SergeyKanzhelev yes, the test was trying to validate that after the gracePeriod, the pod still have 2 seconds of [extended grace period](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination) (which cause the flakyness). The proposed change is that even with a still running preStop hook, the pod eventually will be terminated.

test/e2e/node/pre_stop.go
```diff
@@ -202,7 +202,7 @@ var _ = SIGDescribe("PreStop", func() {
 		framework.ExpectNoError(err, "failed to delete pod")
 
 		//wait up to graceful termination period seconds
```
SergeyKanzhelev, line 204: just a small suggestion - perhaps we can explain why it is 15 seconds in a comment here.

SergeyKanzhelev:
/ok-to-test
/lgtm
(will re-lgtm with the update of comment)

spiffxp:
What's the proof the flakiness is solved?

I'll run the node e2e test a few times for grins

/test pull-kubernetes-node-e2e

SergeyKanzhelev:
Is this the link to check job result statuses? https://prow.k8s.io/pr-history/?org=kubernetes&repo=kubernetes&pr=94922 I might have re-started job too early. Not sure

knabben:
Failed with another flaky -  *[sig-storage] HostPath should support r/w [NodeConformance]*

knabben:
/test pull-kubernetes-node-e2e

@spiffxp good to go here?

oomichi:
This mark `flake` has been added since https://github.com/kubernetes/kubernetes/pull/83792 and the test was failed flaky at https://github.com/kubernetes/kubernetes/blob/14d380c925a9df7bf1e6d045608298895f0def3d/test/e2e/node/pre_stop.go#L212

I am not sure what ratio of this test failure, can we know that at this time to make the test un-flake safely?
/test pull-kubernetes-node-e2e

knabben:
@oomichi the test was sleeping for the entire graceful period (30 seconds), and checking if the pod is running in the extended grace period +2 seconds, reducing the sleep time in half will ensure the test is still running and is inside the grace period.

 /test pull-kubernetes-node-e2e

test/e2e/node/pre_stop.go
```diff
@@ -201,8 +201,9 @@ var _ = SIGDescribe("PreStop", func() {
 		err = podClient.Delete(context.TODO(), pod.Name, *metav1.NewDeleteOptions(gracefulTerminationPeriodSeconds))
 		framework.ExpectNoError(err, "failed to delete pod")
 
-		//wait up to graceful termination period seconds
-		time.Sleep(30 * time.Second)
+		// wait for less than the gracePeriod termination ensuring the
+		// preStop hook is still executing.
+		time.Sleep(15 * time.Second)
 
 		ginkgo.By("verifying the pod running state after graceful termination")
```
oomichi, line 208: > @oomichi the test was sleeping for the entire graceful period (30 seconds), and checking if the pod is running in the extended grace period +2 seconds, reducing the sleep time in half will ensure the test is still running and is inside the grace period.

Thanks for your explanation. I guess I got the point.
I'd like to summarize my understanding to know that is correct:

The original e2e test expects the pod is still `Running` after graceful termination (30 seconds in this test case) because of the extended 2 seconds as you said since
```
If one of the Pod's containers has defined a preStop hook, the kubelet runs that hook inside of the container.
If the preStop hook is still running after the grace period expires, the kubelet requests a small, one-off grace
period extension of 2 seconds.
```
of https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination

However that 2 seconds can make the test flake due to high workload at upstream CI systems sometimes.
Then this PR changes to verify the pod is still running before expiring the graceful termination during preStop hook is running.

If the above is correct, there are 2 points.
- The above message needs to be updated because the graceful termination doesn't happen at the time.
- Technically this test doesn't check preStop hook is running. the preStop outputs `preStop` to its stdout every 1 second but that is not checked on the test side. I don't think we need to add this check in this PR for the scope, it is better to add it with another PR.

knabben: Yes, this is the exact rationality here.

1. Yes, maybe *verifying the pod is in running state while in the graceful period termination*
2. Sure, the conformance test for the preHook follows this idea. I can propose another PR  or even a new test with the stdout check logic.

oomichi: Thanks, the above sounds good for me.

SergeyKanzhelev:
/lgtm

thank you for updating the comment

oomichi:
Thanks for updating.

/lgtm

spiffxp:
I think I would prefer to merge the change in this PR, but remove the Flaky tag in a followup PR, based on data from https://testgrid.k8s.io/google-gce#gce-cos-master-flaky-repro

knabben:
I agree, @spiffxp PTAL, I open another one to remove the flaky tag after we ensure it's been running flat.

oomichi:
Nice point to keep Flaky label to get data.

/lgtm

spiffxp (approved):
/approve
Thank you! And thanks for your patience in getting this through

knabben:
Flaky from:

go_test: //staging/src/k8s.io/legacy-cloud-providers/vsphere/go_default_test:run_2_of_2

PR #94936, opened 2020-09-21
Return the Kubernetes version which stopped serving deprecated APIs by default

### LINKED ISSUE #94933: Warning message for deprecated APIs for some resources show incorrect Kubernetes version
**What happened**:

When calling API https://github.com/kubernetes/apiserver/blob/master/pkg/endpoints/deprecation/deprecation.go#L101, it returns incorrect message for some resources. For example:

```console
extensions/v1beta1 Deployment is deprecated in v1.8+, unavailable in v1.18+; use apps/v1 Deployment
```

The above API is defined in https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/extensions/v1beta1/types.go#L80. It would seem that other deprecated and removed versions maybe incorrectly defined in the annotations as well.

For example above, I would have expected:

```console
extensions/v1beta1 Deployment is deprecated in v1.9+, unavailable in v1.16+; use apps/v1 Deployment
```

This is because the API is removed in 1.16. This applies to other resources and API versions as well.

- Try to deploy a Deployment object with `apiVersion: extensions/v1beta1` in Kubernetes 1.16. It should fail to deploy.
- https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/ provide a good description of API versions

**Environment**:
- Kubernetes version (use `kubectl version`): 1.19
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

liggitt:
**What type of PR is this?**

/kind cleanup

These resources stopped being served by default in 1.16, and could no longer be served in 1.18. We didn't think that hard about which version to put here when the lifecycle feature was added in 1.19 because it didn't affect 1.19 server behavior at all.

Even though particular cluster administrators could re-enable serving these resources in 1.16 and 1.17, clients that introspect the API types directly to get early warning on deprecated APIs and target removal versions should be told the default configurations most clusters will follow.

Fixes #94933

```release-note
NONE
```

/cc @deads2k


liggitt:
cc @hickeyma @joejulian

liggitt:
fixed verify error (stale generated file)

hickeyma (changes requested):
@liggitt Thanks for pushing this PR. The removed versions look good. Do you mind checking the deprecated versions? I think that some of the deprecated versions are 1.9? Here is a list of some of the objects: https://github.com/hickeyma/helm-mapkubeapis/blob/master/config/Map.yaml.

joejulian (approved):
Awesome, thank you.

liggitt:
> Do you mind checking the deprecated versions? I think that some of the deprecated versions are 1.9? Here is a list of some of the objects: https://github.com/hickeyma/helm-mapkubeapis/blob/master/config/Map.yaml.

the first couple I spot checked looked correct as is:

* https://github.com/kubernetes/api/blob/release-1.8/extensions/v1beta1/types.go#L165-L168
* https://github.com/kubernetes/api/blob/release-1.8/apps/v1beta1/types.go#L265-L268

vs

* https://github.com/hickeyma/helm-mapkubeapis/blob/master/config/Map.yaml#L2-L4

joejulian:
Once merged, please cherry-pick to 1.19

liggitt:
picked to https://github.com/kubernetes/kubernetes/pull/94973

PR #84061, opened 2019-10-17
apiserver: disable info level gRPC logging

gyuho:
gRPC balancer wrapper has (non-leveled) info logging on "Notify" call.
Which maens we are logging every single endpoint resolution.
When the cluster scales up, the logging for apiserver can dramatically
increase.

Since etcd v3 client already provides all the relevant information,
let's disable gRPC info logging to reduce the log size.

Signed-off-by: Gyuho Lee <leegyuho@amazon.com>

**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Disable gRPC info-level logging. We need it to reduce the logging size. We don't need these logs:

```
I1017 18:22:59.790791       6 asm_amd64.s:1337] balancerWrapper: got update addr from Notify: []
I1017 18:22:59.790877       6 controlbuf.go:382] transport: loopyWriter.run returning. connection error: desc = "transport is closing"
I1017 18:22:59.790986       6 asm_amd64.s:1337] balancerWrapper: got update addr from Notify: [{...:2379 <nil>}]
I1017 18:22:59.791038       6 asm_amd64.s:1337] balancerWrapper: got update addr from Notify: [{...:2379 <nil>}]
I1017 18:22:59.793949       6 asm_amd64.s:1337] balancerWrapper: got update addr from Notify: [{...:2379 <nil>}]
```

**Which issue(s) this PR fixes**:

Fixes https://github.com/kubernetes/kubernetes/issues/80741

**Does this PR introduce a user-facing change?**:

```release-note
Disable gRPC info level logging
```

Issue #80741, opened 2019-07-30
kube-apiserver print  lots of grpc log: got update addr from Notify


RayHuangCN:
I got lots of logs like: 
I0730 15:38:27.137684       1 asm_amd64.s:1337] balancerWrapper: got update addr from Notify: [{etcd.xxx.com:2379 <nil>}]
I0730 15:38:27.137709       1 asm_amd64.s:1337] balancerWrapper: got update addr from Notify: [{etcd.xxx.com:2379 <nil>}]
I0730 15:38:27.137733       1 asm_amd64.s:1337] balancerWrapper: got update addr from Notify: [{etcd.xxx.com:2379 <nil>}]
I0730 15:38:27.142843       1 asm_amd64.s:1337] balancerWrapper: got update addr from Notify: [{etcd.xxx.com:2379 <nil>}]

(10+ lines / 20s)

etcd.xxx.com is LB of etcd cluster
kube-apiserver version is 1.14.3 
i found that cluster can work normally
so ,how can i close those logs??

tedyu:
This is from vendor/google.golang.org/grpc/balancer_v1_wrapper.go

Maybe surface to grpc project ?

xieydd:
I meet this problem too, k8s version is 1.15.1

gyuho:
We do set the gRPC logger by default, in etcd

https://github.com/etcd-io/etcd/blob/a65e4d3357c260504d47f08a433b355e36dbcb51/clientv3/logger.go#L36-L40

to ignore all such logs.

@jpbetz @liggitt Do we override gRPC logger somewhere in kube-apiserver?

I tried

```
env:
- name: GRPC_GO_LOG_SEVERITY_LEVEL
  value: warning
```

but no luck...

```
I1017 18:22:59.790791       6 asm_amd64.s:1337] balancerWrapper: got update addr from Notify: []
I1017 18:22:59.790877       6 controlbuf.go:382] transport: loopyWriter.run returning. connection error: desc = "transport is closing"
I1017 18:22:59.790986       6 asm_amd64.s:1337] balancerWrapper: got update addr from Notify: [{...:2379 <nil>}]
I1017 18:22:59.791038       6 asm_amd64.s:1337] balancerWrapper: got update addr from Notify: [{...:2379 <nil>}]
I1017 18:22:59.793949       6 asm_amd64.s:1337] balancerWrapper: got update addr from Notify: [{...:2379 <nil>}]
```

[...]

PR #94941, opened 2020-09-21
Silence grpc connection log noise.

retroflexer:
A hack to eliminate the superfluous grpc connection log spam from etcd client. Warnings and errors are still emitted.
kubernetes/kubernetes#80741
kubernetes/kubernetes#84061
https://github.com/openshift/cluster-etcd-operator/pull/368
**What type of PR is this?**
/kind cleanup

To eliminate the superfluous grpc connection log spam from etcd client. 

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/storage/etcd3/logger.go
+++ b/staging/src/k8s.io/apiserver/pkg/storage/etcd3/logger.go
@@ -32,15 +32,24 @@ type klogWrapper struct{}
 const klogWrapperDepth = 4
 
 func (klogWrapper) Info(args ...interface{}) {
-	klog.InfoDepth(klogWrapperDepth, args...)
+	//Discard all Info logging, if verbose level is less than 6
+	if klog.V(6).Enabled() {
+		klog.InfoDepth(klogWrapperDepth, args...)
+	}
 }
 
 func (klogWrapper) Infoln(args ...interface{}) {
-	klog.InfoDepth(klogWrapperDepth, fmt.Sprintln(args...))
+	//Discard all Info logging, if verbose level is less than 6
+	if klog.V(6).Enabled() {
+		klog.InfoDepth(klogWrapperDepth, fmt.Sprintln(args...))
+	}
 }
 
 func (klogWrapper) Infof(format string, args ...interface{}) {
-	klog.InfoDepth(klogWrapperDepth, fmt.Sprintf(format, args...))
+	//Discard all Info logging, if verbose level is less than 6
+	if klog.V(6).Enabled() {
+		klog.InfoDepth(klogWrapperDepth, fmt.Sprintf(format, args...))
+	}
 }
 
 func (klogWrapper) Warning(args ...interface{}) {


retroflexer:
I am looking the old thread:
https://github.com/kubernetes/kubernetes/pull/84061/files#r336218221
I will push another change with klog writers. 
Also need to update the vendor.

retroflexer:
I signed it.

staging/src/k8s.io/apiserver/pkg/storage/etcd3/logger.go
```diff
@@ -32,15 +32,15 @@ type klogWrapper struct{}
...
 }
 
 func (klogWrapper) Infoln(args ...interface{}) {
-	klog.InfoDepth(klogWrapperDepth, fmt.Sprintln(args...))
+	//Discard all Info logging. Do nothing.
 }
 
 func (klogWrapper) Infof(format string, args ...interface{}) {
-	klog.InfoDepth(klogWrapperDepth, fmt.Sprintf(format, args...))
+	//Discard all Info logging. Do nothing.
```
sttts, line 43: Depending on verbosity level we still want these. Maybe at 9 ? Don't know how we map that elsewhere.

sttts: Or maybe 4 or 6.

sttts:
cc @hexfusion

staging/src/k8s.io/apiserver/pkg/storage/etcd3/logger.go
ehashman: Agree, seems needless to totally throw this away... can we not bump it to a more verbose log level?

aojea:
this is super hacky :upside_down_face:

liggitt:
I'll defer to some of the etcd folks on whether this is a good approach - cc @ptabor / @mm4tt

retroflexer:
> this is super hacky upside_down_face

Why do you think it is hacky? You may notice that the logwrapper was already there before this PR, and I just used the existing log wrapper to silence the grpc log noise.

ptabor:
I think that etcd client 3.5 have a lot of improvement in that area: 
  - Since https://github.com/etcd-io/etcd/pull/12671  etcd client does not use custom balancer, just the standard one, so the logging should be the same as all the other services. 
  - Control using `ETCD_CLIENT_DEBUG` env variable (https://github.com/etcd-io/etcd/pull/12786)

aojea:
> Why do you think it is hacky? You may notice that the logwrapper was already there before this PR, and I just used the existing log wrapper to silence the grpc log noise.

well, it is in the description :)

> A hack to eliminate the superfluous grpc connection log spam from etcd client. Warnings and errors are still emitted.

I have some questions, because I know and agree that the grpc logs are to verbose, but:
- Is there consensus that ALL the info grpc level is noise? the grpc-go people doesn't seem to have the same opinion

https://github.com/grpc/grpc-go/pull/2943#issuecomment-516881207
https://github.com/grpc/grpc-go/blob/master/Documentation/log_levels.md

- what if there is something that "is not noise" later? 
debug 6 is too verbose, there is no way to enable that in production just to see a grpc message

retroflexer:
closing as the replicated PR 102896 got merged.

/close

Issue #93320, opened 2020-07-22
google.golang.org/grpc cannot be upgraded >= v1.30.0


entropitor:
**What happened**: Kubernetes is using GRPC for a lot of stuff (CSI, ...) to allow for extensibility, however, the grpc-go driver contained a bug which basically means that writing GRPC servers for kubernetes was not allowed due to a broken implementation in the go client (https://github.com/grpc/grpc-go/issues/2628). So ideally, this grpc-go dependency should be upgraded across the board and maybe even backported to deliver on the promise of interoperability with other languages

**What you expected to happen**: Being able to e.g. write a CSI driver in typescript / rust

**How to reproduce it (as minimally and precisely as possible)**: See https://github.com/grpc/grpc-go/issues/2628 for more information, basically the grpc-go client sets the authority field to an HTTP2 spec-uncompliant value and so it's not working in rust / nodejs

**Anything else we need to know?**: 

**Environment**:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

nikhita:
/sig architecture
/area code-organization
/area dependency

related - https://github.com/kubernetes/kubernetes/issues/91881

cc @dims

dims:
@entropitor same comment when we wanted to do this last https://github.com/kubernetes/kubernetes/pull/91898#issuecomment-640544757

Can you please update etcd to newer grpc and then we can pull that in here? grpc ends up bringing in a bunch of things that has broken us in the past and our major usage is etcd oriented, so it would make sense to update etcd master first to give us some confidence.

nikhita:
etcd poc - https://github.com/etcd-io/etcd/pull/12155

TheEnbyperor:
Any updates on this? Currently hitting this issue myself writing some things in Rust.

dims:
we need to update etcd, containerd and others that we depend on to the latest google.golang.org/grpc *before* we can update kubernetes. If someone wants to propose PRs in those repositories (At least etcd and containerd!)

https://github.com/grpc/grpc-go/compare/v1.27.0...v1.31.1 (looks like a bunch of changes in there)

nikhita:
update from etcd maintainers - https://github.com/etcd-io/etcd/issues/12124#issuecomment-674368288

[...]

PR #94942, opened 2020-09-21
Apiserver tracing

dashpole:
**What type of PR is this?**

/kind feature

Enhancement issue: https://github.com/kubernetes/enhancements/issues/647
Implementation of https://github.com/kubernetes/enhancements/pull/1458.

Add tracing for API Server requests using opentelemetry-go, and the OTLP exporter.  This has two primary components:

1. Use otelhttp "wrapper" library in api server's http server
2. Add file-based configuration for enabling and configuring tracing using the OTLP exporter.

The following will be done as follow-ups:

* Use otelhttp in client-go to instrument outgoing http requests
* Add e2e test for tracing
* Use otelgrpc with the etcd client to instrument requests to etcd, and integrate with https://github.com/etcd-io/etcd/pull/12919

**Dependencies added**

* Http instrumentation for opentelemetry: [go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/net/http/otelhttp)

```release-note
Add distributed tracing to the kube-apiserver.  It is can be enabled with the feature gate: APIServerTracing=true
```

```docs
KEP: https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/0034-distributed-tracing-kep.md
```


dashpole:
/test all

I manually tested the following configurations, and both work:

```yaml
- name: opentelemetry
  connection:
    proxyProtocol: Direct
```
and 
```yaml
- name: opentelemetry
  connection:
    proxyProtocol: GRPC
    transport:
      uds:
        udsName: /etc/srv/kubernetes/konnectivity-server/konnectivity-server.socket
```

I think this validates that the opentelemetry exporter can work with the egress selector feature.

dashpole:
/test all
I manually tested, and this now correctly traces reentrant API calls and webhook calls.

staging/src/k8s.io/client-go/transport/round_trippers.go
```diff
@@ -61,9 +65,27 @@ func HTTPWrappersForConfig(config *Config, rt http.RoundTripper) (http.RoundTrip
...
 }
 
+type otelWrapper struct {
+	otelRT http.RoundTripper
+	baseRT http.RoundTripper
+}
+
+// NewOtelRoundTripper returns an otelhttp-wrapped round tripper that implements
+// WrappedRoundTripper
+func NewOtelRoundTripper(rt http.RoundTripper) http.RoundTripper {
```
dashpole, line 79: note to reviewers: This is required because unit tests "unwrap" the RoundTripper to inspect the underlying http Transport

staging/src/k8s.io/apiserver/pkg/apis/apiserver/v1alpha1/types.go
```diff
@@ -147,3 +147,36 @@ type TLSConfig struct {
 	// +optional
 	ClientCert string `json:"clientCert,omitempty"`
 }
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
dashpole, line 151: note to reviewers: This is copied from the KEP

staging/src/k8s.io/apiserver/pkg/endpoints/filters/traces.go
```diff
@@ -0,0 +1,27 @@
...
+	"net/http"
+
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/propagators"
+)
+
+// WithTracing adds tracing to requests if the incoming request is sampled
+func WithTracing(handler http.Handler) http.Handler {
+	return otelhttp.NewHandler(handler, "KubernetesAPI", otelhttp.WithPropagators(otel.NewCompositeTextMapPropagator(propagators.TraceContext{}, propagators.Baggage{})))
```
dashpole, line 29: Note to reviewers: This is a No-op unless an "sdk" is registered.  See the added component-base utils.

staging/src/k8s.io/client-go/transport/round_trippers.go
```diff
@@ -61,9 +65,27 @@ func HTTPWrappersForConfig(config *Config, rt http.RoundTripper) (http.RoundTrip
...
+	return &otelWrapper{otelRT: otelhttp.NewTransport(rt, otelhttp.WithPropagators(otel.NewCompositeTextMapPropagator(propagators.TraceContext{}, propagators.Baggage{}))), baseRT: rt}
```
dashpole, line 80: note to reviewers: This is a no-op unless a trace "sdk" is registered.  See the component base utils for how that can be done.

staging/src/k8s.io/apiserver/pkg/server/options/opentelemetry.go
```diff
@@ -0,0 +1,140 @@
...
+			opts = append(opts, otlp.WithGRPCDialOption(grpc.WithContextDialer(otelDialer)))
+		}
+	}
+
+	traces.InitTraces("kube-apiserver", opts...)
+	return nil
+}
+
+// Validate verifies flags passed to OpenTelemetryOptions.
+func (o *OpenTelemetryOptions) Validate() []error {
```
logicalhan, line 128: `func (o *OpenTelemetryOptions) Validate() (errs []error) {`

I think allows you to drop line 133 (you'll have to remove the nil on 130 too)

staging/src/k8s.io/apiserver/pkg/opentelemetry/config.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+var cfgScheme = runtime.NewScheme()
+
+func init() {
+	install.Install(cfgScheme)
+}
+
+// ReadOpenTelemetryConfiguration reads the opentelemetry configuration from a file
+func ReadOpenTelemetryConfiguration(configFilePath string) (*apiserver.OpenTelemetryClientConfiguration, error) {
```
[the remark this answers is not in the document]
dashpole: note to self: defaulting happens before validation by convention: https://github.com/kubernetes/apiserver/blob/02be95754551452cc1aab7ebc00ec73d67728bca/pkg/registry/rest/create_update.go#L38

staging/src/k8s.io/apiserver/pkg/server/options/opentelemetry.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	errs := opentelemetry.ValidateOpenTelemetryConfiguration(npConfig)
+	if len(errs) > 0 {
+		return fmt.Errorf("failed to validate opentelemetry configuration: %v", errs.ToAggregate())
+	}
+
+	if npConfig == nil {
+		// No config file was specified, so don't enable exporting
+		return nil
+	}
```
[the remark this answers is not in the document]
dashpole: npConfig is returned by `ReadOpenTelemetryConfiguration`.  I could change it so that we check to see if `o.ConfigFile == ""` instead, and move it up.  Then I would make ReadOpenTelemetryConfiguration return an error if the config file is empty.  Does that sound better?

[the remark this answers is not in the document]
logicalhan: yeah i think that'd be a bit cleaner.

staging/src/k8s.io/apiserver/pkg/server/options/opentelemetry.go
```diff
@@ -0,0 +1,140 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net"
+
+	"github.com/spf13/pflag"
+	"go.opentelemetry.io/otel/exporters/otlp"
+	"google.golang.org/grpc"
+	"k8s.io/utils/path"
```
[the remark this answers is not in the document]
dashpole: done

staging/src/k8s.io/apiserver/pkg/server/options/opentelemetry.go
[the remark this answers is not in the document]
dashpole: done

staging/src/k8s.io/apiserver/pkg/opentelemetry/config.go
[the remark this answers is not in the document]
dashpole: done

staging/src/k8s.io/apiserver/pkg/server/options/opentelemetry.go
dashpole: done

dashpole:
/hold
(this is blocked by #93320)

dashpole:
tests are expected to fail until I can bump the grpc version

cmd/kube-apiserver/app/options/options.go
```diff
@@ -58,6 +58,7 @@ type ServerRunOptions struct {
 	EgressSelector          *genericoptions.EgressSelectorOptions
 	Metrics                 *metrics.Options
 	Logs                    *logs.Options
+	OpenTelemetry           *genericoptions.OpenTelemetryOptions
```
logicalhan, line 61: we should probably eventually refactor this into flags defined from `component-base/traces` (but not necessarily right now)

staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go
```diff
@@ -134,6 +136,8 @@ func newETCD3Client(c storagebackend.TransportConfig) (*clientv3.Client, error)
 		grpc.WithBlock(), // block until the underlying connection is up
 		grpc.WithUnaryInterceptor(grpcprom.UnaryClientInterceptor),
 		grpc.WithStreamInterceptor(grpcprom.StreamClientInterceptor),
+		grpc.WithUnaryInterceptor(otelgrpc.UnaryClientInterceptor(global.Tracer(""))),
```
logicalhan, line 139: just curious, why the empty string?

logicalhan (commented):
/lgtm 

(minus the hack failures)

staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go
dashpole: empty string implies the default tracer.  In future versions of the otelgrpc plugin, the argument is removed entirely, and it only uses the default.

dims:
/approve
(/vendor looks good, 3 new Apache Licensed packages)

dims:
/assign @smarterclayton @liggitt 

(for root approval)

staging/src/k8s.io/apiserver/pkg/opentelemetry/config.go
```diff
@@ -0,0 +1,128 @@
...
+	if configFilePath == "" {
+		return nil, fmt.Errorf("opentelemetry config file was empty")
+	}
+	data, err := ioutil.ReadFile(configFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read opentelemetry configuration from %q [%v]", configFilePath, err)
+	}
+	var decodedConfig v1alpha1.OpenTelemetryClientConfiguration
+	err = yaml.Unmarshal(data, &decodedConfig)
```
liggitt, line 53: this won't handle versioning correctly or detect if the config file is in a different or unrecognized version... normally we would do something like this:

```
internalConfig := &apiserver.OpenTelemetryClientConfiguration{}
codecs := serializer.NewCodecFactory(cfgScheme)
// this handles json/yaml/whatever, and decodes all registered version to the internal version
if err := runtime.DecodeInto(codecs.UniversalDecoder(), data, internalConfig); err != nil {
  ...
}
```

cmd/kube-apiserver/app/options/options.go
liggitt: is this kube-apiserver specific or would it be applicable to any apiserver?

staging/src/k8s.io/apiserver/pkg/opentelemetry/config.go
```diff
@@ -0,0 +1,128 @@
...
+	internalConfig := &apiserver.OpenTelemetryClientConfiguration{}
+	if err := cfgScheme.Convert(&decodedConfig, internalConfig, nil); err != nil {
+		// we got an error where the decode wasn't related to a missing type
+		return nil, err
+	}
+	return internalConfig, nil
+}
+
+// DefaultOpenTelemetryConfiguration defaults unset fields in the OpenTelemetryClientConfiguration
+func DefaultOpenTelemetryConfiguration(config *apiserver.OpenTelemetryClientConfiguration) {
```
liggitt, line 70: if we use the regular decoding path these can be set in a normal API object defaulting function and will get set automatically at decode time

staging/src/k8s.io/apiserver/pkg/server/config.go
```diff
@@ -749,6 +749,7 @@ func DefaultBuildHandlerChain(apiHandler http.Handler, c *Config) http.Handler {
 	handler = genericfilters.WithTimeoutForNonLongRunningRequests(handler, c.LongRunningFunc, c.RequestTimeout)
 	handler = genericfilters.WithWaitGroup(handler, c.LongRunningFunc, c.HandlerChainWaitGroup)
 	handler = genericapifilters.WithRequestInfo(handler, c.RequestInfoResolver)
+	handler = genericapifilters.WithTracing(handler)
```
liggitt, line 752: I expected to see a feature or configuration gate here so this filter was not active if feature was off or unconfigured

staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go
liggitt: same comment about being inert if the feature is off

cluster/gce/gci/configure-helper.sh
```diff
@@ -995,6 +995,13 @@ users:
 EOF
   fi
 
+  if [[ "${ENABLE_APISERVER_TRACING:-false}" == "true" ]]; then
+    cat <<EOF >/etc/srv/kubernetes/opentelemetry_config.yaml
+apiVersion: apiserver.k8s.io/v1alpha1
```
liggitt, line 1000: config types use `*.config.k8s.io`, so `apiserver.config.k8s.io/...`

using the normal decoding path will catch this

liggitt (commented):
has the e2e test been enabled in the alpha e2e job and passed? I don't see it in the latest green runs

dashpole:
I'm getting a bunch of these errors in e2e tests: `request connection cannot be hijacked: *metrics.ResponseWriterDelegator`.  I must have broken something with the downgrade to an earlier version of opentelemetry to work around https://github.com/kubernetes/kubernetes/issues/93320...  I don't think i'm going to get this working tonight

kikisdeliveryservice:
Hi @dashpole 

[Code Freeze](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#code-freeze) is now in effect. Your PR is still unmerged, with no approvals and has a hold. If you believe this should be in the 1.20 Release, please file an [Exception](https://github.com/kubernetes/sig-release/blob/master/releases/EXCEPTIONS.md).

Tracking Issue: https://github.com/kubernetes/enhancements/issues/647

Best,
Kirsten

dashpole:
I confirmed that v0.6.0 (the last OTel release using v1.27.x for grpc) of opentelemetry won't work in kubernetes.  To fix the errors I [found above](https://github.com/kubernetes/kubernetes/pull/94942#issuecomment-726488969), I need this patch: https://github.com/open-telemetry/opentelemetry-go/pull/979, which isn't available until v0.11.0.  However, v0.11.0 requires grpc v1.31.0, which is blocked by https://github.com/kubernetes/kubernetes/issues/93320.  So this won't be possible to do in 1.20.

staging/src/k8s.io/apiserver/pkg/server/config.go
dashpole: Unless an "SDK" is registered (which is done in the InitTraces function in staging/src/k8s.io/component-base/traces/utils.go using `otel.SetTracerProvider(...)`), OpenTelemetry uses a NoOp tracer and doesn't propagate context.  This means the Tracing filter doesn't do anything unless we register an SDK, which is controlled by the feature gate.

staging/src/k8s.io/apiserver/pkg/opentelemetry/config.go
dashpole: done

staging/src/k8s.io/apiserver/pkg/opentelemetry/config.go
dashpole: done

cluster/gce/gci/configure-helper.sh
dashpole: Interesting... I assumed this belonged in the same place as the EgressSelectorConfiguration.  Does that mean I should extract the type out of staging/src/k8s.io/apiserver/pkg/apis/apiserver, and move it somewhere else?

dashpole: Figured it out.  The apiserver.k8s.io group was only used for v1alpha1 and v1beta1 apis, and the v1 group had moved to apiserver.config.k8s.io.  I had to add apiserver.config.k8s.io to the v1alpha1 and v1beta1 groups.

Done.

dashpole:
To get most of the feature implemented, i'm switching to using the stdout exporter instead of the OTLP exporter.  This is only temporary, but will allow us to stabilize tests and such early on in the release cycle.

The only thing left is to modify and re-add the e2e test so it looks in the kube-apiserver logs instead of deploying an OpenTelemetry collector and looking in the collector's logs.

/hold cancel

dashpole:
This is all ready to go.  I'll have to wait for https://github.com/kubernetes/test-infra/pull/20369 to merge before the test can run here.

justaugustus:
> This is all ready to go. I'll have to wait for [kubernetes/test-infra#20369](https://github.com/kubernetes/test-infra/pull/20369) to merge before the test can run here.

Approved. It should merge shortly! :)

dashpole:
The tracing e2e test passed, so this is ready for review.

liggitt:
/assign @lavalamp 
/cc @deads2k 

Will defer to Daniel for review of apiserver changes and quality of the added dependencies (of which there are many :-/). Once changes have lgtm, I can ack the dependency updates.

staging/src/k8s.io/apiserver/pkg/server/config.go
lavalamp: I'd still prefer to feature gate this.

staging/src/k8s.io/component-base/traces/utils.go
```diff
@@ -0,0 +1,64 @@
...
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/semconv"
+	"google.golang.org/grpc"
+
+	"k8s.io/klog/v2"
+)
+
+// InitTraces initializes tracing in the component.
+func InitTraces(service, address string, opts ...grpc.DialOption) {
```
lavalamp, line 34: Does this have to be global? I regret every global I've ever held my nose for.

staging/src/k8s.io/apiserver/pkg/server/config.go
lavalamp: Also I don't think this should be relying on globals to be set up right elsewhere, I think a handle or context should be passed here.

lavalamp:
> Will defer to Daniel for review of apiserver changes and quality of the added dependencies (of which there are many :-/). Once changes have lgtm, I can ack the dependency updates.

This is enormous, I glanced and pointed out a couple things. I don't know how I will find a block of time to actually review in detail though.

cmd/kube-apiserver/app/options/options.go
sttts: I would expect this in every apiserver. Though this does't make it misplaced. But it should be part of `RecommendedOptions` too if it is't yet.

cmd/kube-apiserver/app/options/options_test.go
```diff
@@ -303,12 +304,15 @@ func TestAddFlags(t *testing.T) {
...
-		Metrics:                           &metrics.Options{},
-		Logs:                              logs.NewOptions(),
+		EnableLogsHandler:       false,
+		EnableAggregatorRouting: true,
+		ProxyClientKeyFile:      "/var/run/kubernetes/proxy.key",
+		ProxyClientCertFile:     "/var/run/kubernetes/proxy.crt",
+		Metrics:                 &metrics.Options{},
+		Logs:                    logs.NewOptions(),
+		OpenTelemetry: &apiserveroptions.OpenTelemetryOptions{
+			ConfigFile: "/var/run/kubernetes/opentelemetry_config.yaml",
```
sttts, line 314: what's the value of a default path? I would prefer to see this being disabled by default.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/v1alpha1/defaults.go
```diff
@@ -0,0 +1,46 @@
...
+	if config == nil {
+		return
+	}
+	// Default the service port to the default OTLP port
+	if config.Service != nil && config.Service.Port == nil {
+		config.Service.Port = &defaultPort
+	}
+	// If niether URL or service is set, use the default URL
+	if config.Service == nil && config.URL == nil {
+		config.URL = &defaultURL
```
sttts, line 44: a default like this is odd. Is `localhost:55680` something to be expected on every node in an OpenTelemetry enabled cluster? I.e., is this a convention?

staging/src/k8s.io/apiserver/pkg/apis/apiserver/v1alpha1/defaults_test.go
```diff
@@ -0,0 +1,91 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
sttts, line 2: 2021 and probably in many other places

staging/src/k8s.io/apiserver/pkg/apis/apiserver/v1beta1/register.go
```diff
@@ -44,6 +48,9 @@ func init() {
 
 // Adds the list of known types to the given scheme.
 func addKnownTypes(scheme *runtime.Scheme) error {
+	scheme.AddKnownTypes(LegacySchemeGroupVersion,
+		&EgressSelectorConfiguration{},
```
sttts, line 52: all these legacy/new group name changes should not be mixed into the tracing commit.

staging/src/k8s.io/apiserver/pkg/opentelemetry/config.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// ReadOpenTelemetryConfiguration reads the opentelemetry configuration from a file
+func ReadOpenTelemetryConfiguration(configFilePath string) (*apiserver.OpenTelemetryClientConfiguration, error) {
+	if configFilePath == "" {
+		return nil, fmt.Errorf("opentelemetry config file was empty")
+	}
+	data, err := ioutil.ReadFile(configFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read opentelemetry configuration from %q [%v]", configFilePath, err)
```
sttts, line 47: `... : %v` for the error

staging/src/k8s.io/apiserver/pkg/opentelemetry/config.go
```diff
@@ -0,0 +1,101 @@
...
+	allErrs := field.ErrorList{}
+	if config == nil {
+		// OpenTelemetry is disabled
+		return allErrs
+	}
+	if config.Service != nil && config.URL != nil {
+		allErrs = append(allErrs, field.Invalid(
+			field.NewPath("service"),
+			config.Service,
+			"Service and URL cannot both be set"))
```
sttts, line 68: lower-case

staging/src/k8s.io/apiserver/pkg/opentelemetry/config.go
```diff
@@ -0,0 +1,101 @@
...
+	return allErrors
+}
+
+func validateURL(u string, fldPath *field.Path) field.ErrorList {
+	errs := field.ErrorList{}
+	_, err := url.Parse(u)
+	if err != nil {
+		return append(errs, field.Invalid(
+			fldPath, u,
+			fmt.Sprintf("Unable to parse URL: %v", err)))
```
sttts, line 98: lower-case

staging/src/k8s.io/apiserver/pkg/server/config.go
sttts: Commented above already: even when the feature gate is on by default in the future, the functionality should be disabled if there is no config provided. The current code has a default config file path and I assume will fail when the config does not exist. 

In other words: +1 for opt-in, plus even after feature gate goes away some day.

staging/src/k8s.io/component-base/traces/utils.go
```diff
@@ -0,0 +1,64 @@
...
+	// TODO(dashpole): replace with otlp exporter
+	exporter, err := stdout.NewExporter([]stdout.Option{
+		stdout.WithPrettyPrint(),
+	}...)
+	if err != nil {
+		klog.Fatalf("failed to initialize stdout export pipeline: %v", err)
+	}
+
+	res, err := resource.New(context.Background(),
```
sttts, line 43: background contexts are suspicious. Why isn't this a parameter?

test/e2e/instrumentation/tracing/apiserver.go
```diff
@@ -0,0 +1,96 @@
...
+				return err
+			}
+			if result.Stderr != "" {
+				return fmt.Errorf("Non-empty stderr when querying for logs on the master: %v", result.Stderr)
+			}
+			// Check the opentelemetry collector logs to see if they contain our trace ID
+			if strings.Contains(logs, traceID.String()) {
+				return nil
+			}
+			return fmt.Errorf("Failed to find trace ID %v in log: \n%v", traceID.String(), logs)
```
sttts, line 92: lower-case errors everywhere

test/e2e/instrumentation/tracing/apiserver.go
```diff
@@ -0,0 +1,96 @@
...
+			_, err := c.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+			if err != nil {
+				return err
+			}
+
+			// Get logs from the opentelemetry collector pod on the master.
+			// We must use SSH because we can't fetch logs from master pods
+			// using the pod logs subresource.
+			result, err := e2essh.SSH(
+				"sudo cat /var/log/kube-apiserver.log",
```
sttts, line 77: am not sure about similar dependencies nowadays in e2e tests. But this is a very strong assumption about the e2e environments. k/k is not the only one that runs these tests.

Isn't there a much simpler way to integation test this inside of k8s.io/kube-apiserver?

staging/src/k8s.io/component-base/traces/utils.go
dashpole: It does not have to be global.  I'll rework it so that it isn't.

lavalamp:
> It does not have to be global. I'll rework it so that it isn't.

I'm halfway (ok, well, partway) through a review so don't rush on that. :)

cmd/kube-apiserver/app/options/options_test.go
dashpole: It is disabled by default.  This test is checking that the `--opentelemetry-config-file=/var/run/kubernetes/opentelemetry_config.yaml` flag above is translated into the correct OpenTelemetryOptions struct here.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/v1alpha1/defaults.go
dashpole: Yes, 55680 was the default port (now we have an [IANA reserved port](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt) of 4317; thanks for reminding me to update it).  Localhost is a default because it is the most commonly used, as the collector is often run as a sidecar container.  Since master components are generally on host network, the expected way to use this is to run the collector as a sidecar.

test/e2e/instrumentation/tracing/apiserver.go
dashpole: done

staging/src/k8s.io/component-base/traces/utils.go
dashpole: added as a parameter

staging/src/k8s.io/apiserver/pkg/opentelemetry/config.go
dashpole: done

staging/src/k8s.io/apiserver/pkg/opentelemetry/config.go
dashpole: done

staging/src/k8s.io/apiserver/pkg/opentelemetry/config.go
dashpole: done.  here and below.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/v1alpha1/defaults_test.go
dashpole: fixed here and elsewhere.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/v1alpha1/defaults.go
dashpole: udpated.

dashpole:
rebased

cmd/kube-apiserver/app/options/options.go
```diff
@@ -180,6 +182,7 @@ func (s *ServerRunOptions) Flags() (fss cliflag.NamedFlagSets) {
 	s.Admission.AddFlags(fss.FlagSet("admission"))
 	s.Metrics.AddFlags(fss.FlagSet("metrics"))
 	s.Logs.AddFlags(fss.FlagSet("logs"))
+	s.OpenTelemetry.AddFlags(fss.FlagSet("opentelemetry"))
```
lilic, line 185: > we should probably eventually refactor this into flags defined from component-base/traces (but not necessarily right now)

As per @logicalhan  comment above, should we not call this `traces` instead of `opentelemetry`. We can always wrap later but this is user facing no?

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
```diff
@@ -147,3 +147,36 @@ type TLSConfig struct {
 	// +optional
 	ClientCert string
 }
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// OpenTelemetryClientConfiguration provides versioned configuration for opentelemetry clients.
```
lilic, line 153: Again is this something we can change later, or should we just use the TracingClientConfiguratio name now?

lilic (commented):
Few comments regarding using the tracing/traces wording instead of OpenTelemetry. To me OpenTelemetry includes other signals and it might be confusing to our users. I would prefer any user facing configuration or flags to be called tracing or traces.

cmd/kube-apiserver/app/options/options.go
brancz: Agreed. We should be consistent here and call this `traces` as opposed to `opentelemetry`.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/v1beta1/register.go
dashpole: Moved to https://github.com/kubernetes/kubernetes/pull/99906

cmd/kube-apiserver/app/options/options.go
dashpole: I added it to RecommendedOptions.

staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go
dashpole: I removed the etcd integration for this PR, but will keep that in-mind when we add this in later.

staging/src/k8s.io/apiserver/pkg/server/config.go
dashpole: This line is now feature gated.  The feature is disabled if the flag: `--tracing-config-file` is not provided.

staging/src/k8s.io/component-base/traces/utils.go
dashpole: This no longer modifies global configuration, and instead plumbs the `TracerProvider` (tracing equivalent of a Logger) to where it is needed.

pkg/features/kube_features.go
```diff
@@ -677,6 +677,12 @@ const (
 	//
 	// Enables the usage of different protocols in the same Service with type=LoadBalancer
 	MixedProtocolLBService featuregate.Feature = "MixedProtocolLBService"
+
+	// owner: @dashpole
+	// alpha: v1.20
```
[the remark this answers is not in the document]
dashpole: done

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
```diff
@@ -147,3 +147,36 @@ type TLSConfig struct {
...
+type OpenTelemetryClientConfiguration struct {
+	metav1.TypeMeta
+
+	// +optional
+	// URL of the collector that's running on the master.
+	// if URL is specified, APIServer uses the egressType Master when sending data to the collector.
+	URL *string
+
+	// +optional
+	// Service that's the frontend of the collector deployment running in the cluster.
```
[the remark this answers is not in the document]
dashpole: renamed.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
```diff
@@ -147,3 +147,36 @@ type TLSConfig struct {
...
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// OpenTelemetryClientConfiguration provides versioned configuration for opentelemetry clients.
+type OpenTelemetryClientConfiguration struct {
+	metav1.TypeMeta
+
+	// +optional
+	// URL of the collector that's running on the master.
+	// if URL is specified, APIServer uses the egressType Master when sending data to the collector.
```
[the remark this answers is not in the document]
dashpole: renamed to ControlPlane

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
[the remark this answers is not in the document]
dashpole: As discussed, we removed service from the configuration.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/v1alpha1/register.go
```diff
@@ -22,7 +22,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
-const GroupName = "apiserver.k8s.io"
+const LegacyGroupName = "apiserver.k8s.io"
```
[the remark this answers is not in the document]
dashpole: moved to #99906.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
```diff
@@ -147,3 +147,36 @@ type TLSConfig struct {
...
+	URL *string
+
+	// +optional
+	// Service that's the frontend of the collector deployment running in the cluster.
+	// If Service is specified, APIServer uses the egressType Cluster when sending data to the collector.
+	Service *ServiceReference
```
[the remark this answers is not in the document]
dashpole: Removed.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/v1alpha1/types.go
```diff
@@ -147,3 +147,36 @@ type TLSConfig struct {
...
+
+	// +optional
+	// URL of the collector that's running on the master.
+	// if URL is specified, APIServer uses the egressType Master when sending data to the collector.
+	URL *string `json:"url,omitempty"`
+
+	// +optional
+	// Service that's the frontend of the collector deployment running in the cluster.
+	// If Service is specified, APIServer uses the egressType Cluster when sending data to the collector.
+	Service *ServiceReference `json:"service,omitempty"`
```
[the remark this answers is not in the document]
dashpole: service is removed.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/v1alpha1/types.go
```diff
@@ -147,3 +147,36 @@ type TLSConfig struct {
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// OpenTelemetryClientConfiguration provides versioned configuration for opentelemetry clients.
+type OpenTelemetryClientConfiguration struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// +optional
+	// URL of the collector that's running on the master.
+	// if URL is specified, APIServer uses the egressType Master when sending data to the collector.
+	URL *string `json:"url,omitempty"`
```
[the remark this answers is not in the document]
dashpole: I updated the KEP to cover this in https://github.com/kubernetes/enhancements/pull/2526

cmd/kube-apiserver/app/options/options.go
dashpole: Changed all references in the PR to use `tracing`/`traces` instead of `opentelemetry`.  I also updated to KEP to clarify this in https://github.com/kubernetes/enhancements/pull/2526.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
dashpole: Agreed.  Changed to `TracingConfiguration`.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/v1alpha1/defaults.go
```diff
@@ -0,0 +1,46 @@
...
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apiserver/pkg/apis/apiserver"
+)
+
+var (
+	defaultPort = int32(55680)
+	defaultURL  = "localhost:55680"
```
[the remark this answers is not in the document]
dashpole: It is currently HTTP, as the collector is expected to be run as a sidecar to the API Server.  Is this still a concern in that case?

test/e2e/instrumentation/tracing/apiserver.go
dashpole: We aren't doing anything fancy with opentelemetry like recording custom spans.  All we are doing is using the otelhttp.Handler for incoming requests.  We could check that the Handler is emitting spans as expected, but that would just be duplicating existing integration tests in the opentelemetry-go-contrib repo.

Since 99% of this feature is configuration for tracing, and integration with the collector, i'd like to ensure those integration points are working properly.  As far as i'm aware, the only way to accomplish that is with an e2e test.

We skip the test if SSH isn't possible, so this won't run if the e2e environment doesn't support it.  I agree SSHing to get at logs on the control plane isn't the best solution.  As a follow-up, i'm planning to change from using stdout and parsing logs to making the collector exit success once a particular trace is found, which would eliminate the need for SSH.  But I think that will require substantial work.

dashpole:
pull-kubernetes-dependencies is expected to fail until  #93320 is resolved.
pull-kubernetes-e2e-gce-alpha-features and pull-kubernetes-unit are expected to fail until this is rebased on #99906

@lavalamp I've resolved all comments except https://github.com/kubernetes/kubernetes/pull/94942#discussion_r569853723, which is about using HTTP.
@sttts I've resolved all comments except https://github.com/kubernetes/kubernetes/pull/94942#discussion_r568607497, which is about e2e testing.

Thanks everyone for the reviews!

staging/src/k8s.io/apiserver/pkg/server/options/tracing.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+
+		otelDialer := func(ctx context.Context, addr string) (net.Conn, error) {
+			return egressDialer(ctx, "tcp", addr)
+		}
+		opts = append(opts, otlp.WithGRPCDialOption(grpc.WithContextDialer(otelDialer)))
+	}
+
+	sampler := sdktrace.NeverSample()
+	if npConfig.SamplingRatePerMillion != nil && *npConfig.SamplingRatePerMillion > 0 {
```
SergeyKanzhelev, line 98: nit: check that `SamplingRatePerMillion` is less than million

staging/src/k8s.io/apiserver/pkg/server/options/tracing.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// Validate verifies flags passed to TracingOptions.
+func (o *TracingOptions) Validate() (errs []error) {
+	if o == nil || o.ConfigFile == "" {
+		return
+	}
+
+	if exists, err := path.Exists(path.CheckFollowSymlink, o.ConfigFile); !exists || err != nil {
+		errs = append(errs, fmt.Errorf("tracing-config-file %s does not exist", o.ConfigFile))
```
SergeyKanzhelev, line 114: nit: include `err` in the error message if not nil

staging/src/k8s.io/apiserver/pkg/server/options/tracing.go
SergeyKanzhelev: nvm

staging/src/k8s.io/apiserver/pkg/tracing/config_test.go
```diff
@@ -0,0 +1,201 @@
...
+    metadata:
+      labels:
+        k8s-app: agent
+    spec:
+      containers:
+        - image: k8s.gcr.io/busybox
+          name: agent
+`,
+			expectedResult: nil,
+			expectedError:  strptr("unable to decode tracing configuration data: no kind \"DaemonSet\" is registered for version \"apps/v1\" in scheme \"k8s.io/apiserver/pkg/tracing/config.go:36\""),
```
SergeyKanzhelev, line 101: is this comparison too strict `k8s.io/apiserver/pkg/tracing/config.go:36`?

annajung:
@dashpole  Hi there, a friendly reminder that the code freeze for 1.21 is today.
Please make sure to rebase and get the necessary reviews for this to merge by EOD PST.

staging/src/k8s.io/apiserver/pkg/server/options/tracing.go
dashpole: done.

staging/src/k8s.io/apiserver/pkg/tracing/config_test.go
dashpole: yeah, I made it check the prefix instead of the full string.

staging/src/k8s.io/apiserver/pkg/server/options/tracing.go
```diff
@@ -113,7 +113,7 @@ func (o *TracingOptions) Validate() (errs []error) {
 	if exists, err := path.Exists(path.CheckFollowSymlink, o.ConfigFile); !exists {
 		errs = append(errs, fmt.Errorf("tracing-config-file %s does not exist", o.ConfigFile))
 	} else if err != nil {
-		errs = append(errs, fmt.Errorf("error checking if tracing-config-file %s esists", o.ConfigFile, err))
+		errs = append(errs, fmt.Errorf("error checking if tracing-config-file %s esists: %v", o.ConfigFile, err))
```
logicalhan, line 116: s/esists/exists/

dashpole: done

dashpole:
I'm also going to cut the e2e test from this PR, as it is much easier to implement if client-go is instrumented (and I previously decided to omit client-go changes)

cmd/kube-apiserver/app/options/options.go
lilic: Thank you!

dims:
/approve cancel

(for now, let's see how this shapes up)

dashpole:
Now that #100488 is merged, this PR is all ready to go!

staging/src/k8s.io/apiserver/pkg/server/options/tracing.go
```diff
@@ -0,0 +1,116 @@
...
+func (o *TracingOptions) AddFlags(fs *pflag.FlagSet) {
+	if o == nil {
+		return
+	}
+
+	fs.StringVar(&o.ConfigFile, "tracing-config-file", o.ConfigFile,
+		"File with apiserver tracing configuration.")
+}
+
+// ApplyTo adds the tracing settings to the global configuration.
```
lilic, line 60: By global configuration what do you mean here?

lilic (commented):
Looking good, just two questions. 👍  🎉 

More things will be done in a follow I see, so we might need to adjust things later on anyways:
- Use otelhttp in client-go to instrument outgoing http requests

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
```diff
@@ -147,3 +147,21 @@ type TLSConfig struct {
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// TracingConfiguration provides versioned configuration for tracing clients.
+type TracingConfiguration struct {
+	metav1.TypeMeta
+
+	// +optional
+	// URL of the collector that's running on the control-plane node.
+	// The APIServer uses the egressType ControlPlane when sending data to the collector.
+	// Defaults to localhost:4317
```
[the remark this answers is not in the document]
dims: if the url is set we pass it through here from what i can tell:
https://github.com/kubernetes/kubernetes/pull/94942/files#diff-0e1453d49880a4f88f0353b9902c7e914513dd1274e3115100ea42c2d116136bR76-R79

And if it is not set, then it defaults to whatever the library uses which seems to be here:
https://github.com/open-telemetry/opentelemetry-go/blob/377f7ce402a2350d5d1a2f09e1d7da884b3bc3d9/exporters/otlp/otlptrace/otlptracehttp/options.go#L67-L71

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
```diff
@@ -147,3 +147,21 @@ type TLSConfig struct {
...
+	URL *string
```
dims, line 161: @dashpole the etcd config options seems to have a instance-id and a service-name:
https://github.com/etcd-io/etcd/pull/12919/files

would that be useful to us too?

staging/src/k8s.io/apiserver/pkg/server/options/tracing.go
dashpole: Ah, the initial implementation set the global TracerProvider and propagators.  Since they are no longer global, I updated the language here.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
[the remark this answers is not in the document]
dashpole: Updated to specify that localhost:4317 is the otlp grpc default

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
dashpole: The service name should be a constant.  An instance id is a great idea.  It turns out there is already an APIServerID field we generate for other purposes, so i'll just use that.

staging/src/k8s.io/apiserver/pkg/server/options/tracing.go
lilic: Nice!! Thanks thought I was missing something obvious :D

lilic (commented):
/lgtm
/approve

Thank you! 🎉 Leaving hold for @dims to have a look again. 

/hold

staging/src/k8s.io/apiserver/pkg/tracing/config.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// ValidateTracingConfiguration validates the tracing configuration
+func ValidateTracingConfiguration(config *apiserver.TracingConfiguration) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if config == nil {
+		// Tracing is disabled
+		return allErrs
+	}
+	if config.SamplingRatePerMillion != nil {
```
dims, line 67: Should we validate `config.URL` is a valid url? 

If we don't, then when will it fail? does it fail when otel stuff is initialized or will it fail when we try to add a trace for something?

dims:
@dashpole one question inline, LGTM otherwise

/assign @liggitt 
/hold cancel

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
```diff
@@ -147,3 +147,21 @@ type TLSConfig struct {
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// TracingConfiguration provides versioned configuration for tracing clients.
+type TracingConfiguration struct {
+	metav1.TypeMeta
+
+	// +optional
+	// URL of the collector that's running on the control-plane node.
+	// The APIServer uses the egressType ControlPlane when sending data to the collector.
+	// Defaults to the otlp grpc default, localhost:4317
```
liggitt, line 160: is a scheme allowed in this URL? if unspecified, does it default to http or https? if https, what TLS configuration is used?

liggitt: also, update doc for external config type with this info

staging/src/k8s.io/apiserver/pkg/endpoints/filters/traces.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"k8s.io/component-base/traces"
+)
+
+// WithTracing adds tracing to requests if the incoming request is sampled
+func WithTracing(handler http.Handler, tp *trace.TracerProvider) http.Handler {
+	opts := []otelhttp.Option{
+		otelhttp.WithPropagators(traces.Propagators()),
+	}
+	if tp != nil {
```
liggitt, line 33: if tp is nil, should we just do `return handler`?

staging/src/k8s.io/apiserver/pkg/tracing/config.go
dashpole: I renamed it to Endpoint, since it is technically a GRPC target, and it follows https://github.com/grpc/grpc/blob/master/doc/naming.md.  It must be a URI, so I added validation as well.

If the endpoint isn't valid, it will log an error every few minutes when it tries to connect.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
dashpole: It must adhere to https://github.com/grpc/grpc/blob/master/doc/naming.md, which allows unix, dns, and unix-abstract schemes.  It defaults to dns as described in the grpc naming doc.  Currently, we do not support TLS, but we could add that in the future.

I updated the description to include those details.

Can you point me to where the docs for external config types is?  I didn't see any other docs in k/k for other apiserver config types

staging/src/k8s.io/apiserver/pkg/endpoints/filters/traces.go
dashpole: A tracer provider is required to export traces, but even without a tracer provider we can still propagate context.  That would allow, for example, linking etcd traces with API requests even if trace collection is not configured for the api server.  I added a comment linking to otel documentation.

staging/src/k8s.io/apiserver/pkg/tracing/config.go
dims: excellent. thanks!

dims:
/retest
/approve
/lgtm

looks good @dashpole let's wrap up addressing the comments from @liggitt and we can ship this!

staging/src/k8s.io/apiserver/pkg/server/options/tracing.go
```diff
@@ -0,0 +1,124 @@
...
+		return fmt.Errorf("failed to validate tracing configuration: %v", errs.ToAggregate())
+	}
+
+	opts := []otlpgrpc.Option{}
+	if npConfig.Endpoint != nil {
+		opts = append(opts, otlpgrpc.WithEndpoint(*npConfig.Endpoint))
+	}
+	if es != nil {
+		// Only use the egressselector dialer if egressselector is enabled.
+		// URL is on the "ControlPlane" network
```
cbandy, line 84: Is "URL" referring to `npConfig.Endpoint` here? (From before it was called Endpoint?)

staging/src/k8s.io/apiserver/pkg/tracing/config_test.go
```diff
@@ -0,0 +1,240 @@
...
+		})
+	}
+}
+
+func TestValidateTracingConfiguration(t *testing.T) {
+	samplingRate := int32(12378)
+	negativeRate := int32(-1)
+	tooHighRate := int32(1000001)
+	validEndpoint := "localhost:4317"
+	dnsEndpoint := "dns:https://google.com:4317"
```
cbandy, line 143: Is `dns:https://` correct here? These [gRPC docs](https://github.com/grpc/grpc/blob/master/doc/naming.md) show `dns://`, I think?

Its test case below has `expectError: false`.

staging/src/k8s.io/apiserver/pkg/server/options/tracing.go
dashpole: Updated to Endpoint.  Yes, it is referring to Endpoint

staging/src/k8s.io/apiserver/pkg/tracing/config_test.go
dashpole: Changed to `dns://`, which is correct.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
liggitt: I just meant the serialized file - staging/src/k8s.io/apiserver/pkg/apis/apiserver/v1alpha1/types.go

That's typically what people look at since that matches the config file they're writing (e.g. https://pkg.go.dev/k8s.io/apiserver/pkg/apis/apiserver/v1alpha1#EgressSelectorConfiguration)

staging/src/k8s.io/apiserver/pkg/endpoints/filters/traces.go
liggitt: interesting. is there any reason someone would not want that once this feature has graduated to GA and can't be disabled via gate?

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
liggitt: > It must adhere to https://github.com/grpc/grpc/blob/master/doc/naming.md, which allows unix, dns, and unix-abstract schemes. It defaults to dns as described in the grpc naming doc. Currently, we do not support TLS, but we could add that in the future.

should that be checked reflected in `validateEndpoint`? if something that parses as a URL but is malformed as a trace exporter endpoint is specified, does kube-apiserver fail to start or does it run in a degraded way?

liggitt:
thanks, just a couple questions left

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
lilic: Validation already happens when the connection tries to be established from what I can tell. But we could provide a more clear error log msg with a `validateEndpoint` helper function. And in either case we should not fail or be degraded if tracing is misconfigured, agreed 👍

staging/src/k8s.io/apiserver/pkg/endpoints/filters/traces.go
dashpole: Generally not, although there may be some users that want a different set of propagators, like the zipkin B3 propagators.

There is a related use-case in which the kube endpoint is public, and you don't trust the context from incoming requests.  For that case we wouldn't want to remove propagators, though.  We would want to use the [WithPublicEndpoint](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/config.go#L103) option on the handler to change the relationships with incoming SpanContext from being a Child to a Link.

staging/src/k8s.io/apiserver/pkg/apis/apiserver/types.go
dashpole: If the endpoint isn't valid (or not reachable), opentelemetry will periodically try to reconnect.  While disconnected, it will buffer (by default up to ~2k) spans, and then drop spans after the buffer is full.  Other than using memory for buffering spans, there won't be any degraded behavior since sending spans is async from the serving path.

I added some additional validation on the scheme.  Mostly, this will catch users that try and specify http(s) as the scheme.

liggitt: sounds good, those seem like likely errors it's good to catch early

logicalhan: Yeah I agree with @liggitt, let's catch this at boot rather than at runtime.

logicalhan: or CRD load time or whatever that point is..

lilic (commented):
/lgtm 🎉 

/hold

for a chance for others to have a look

staging/src/k8s.io/apiserver/pkg/server/config.go
```diff
@@ -775,6 +781,9 @@ func DefaultBuildHandlerChain(apiHandler http.Handler, c *Config) http.Handler {
 	handler = genericfilters.WithHTTPLogging(handler)
 	handler = genericapifilters.WithRequestInfo(handler, c.RequestInfoResolver)
 	handler = genericapifilters.WithRequestReceivedTimestamp(handler)
+	if utilfeature.DefaultFeatureGate.Enabled(genericfeatures.APIServerTracing) {
+		handler = genericapifilters.WithTracing(handler, c.TracerProvider)
+	}
```
liggitt, line 786: just to double check, from the KEP:

> For incoming requests, this would go below WithRequestInfo in the filter stack, as it must be after authentication and authorization, before the panic filter, and is closest in function to the WithRequestInfo filter.

above/below are a little ambiguous... this constructs the filters "inside out"... the last filter added here runs first. Is the intent to run this filter before almost all other filters, including authn/authz and request received timestamp?

staging/src/k8s.io/apiserver/pkg/endpoints/filters/traces.go
liggitt: > There is a related use-case in which the kube endpoint is public, and you don't trust the context from incoming requests. For that case we wouldn't want to remove propagators, though. We would want to use the WithPublicEndpoint option on the handler to change the relationships with incoming SpanContext from being a Child to a Link.

Given the filter is added so it runs before authn/authz, including on generally-accessible endpoints like /healthz, isn't that effectively already possible?

staging/src/k8s.io/apiserver/pkg/server/config.go
dashpole: I'm having trouble remembering the original conversation, but here is the comment I left: https://github.com/kubernetes/enhancements/pull/1458#discussion_r426961816.  The WithRequestInfo reference may not be as relevant anymore since it was recently moved in #102934.  I think the intention is that it is executed _after_ authentication, so I've placed it to be executed after (above in code) the authentication filter.

staging/src/k8s.io/apiserver/pkg/endpoints/filters/traces.go
dashpole: We should probably restrict tracing only to authenticated endpoints (and I moved the tracing filter so it only runs on authenticated requests).  Is there a good list somewhere of endpoints that are authenticated?  Or should I just start with  only tracing requests to `/api`?

liggitt: > I moved the tracing filter so it only runs on authenticated requests

for most clusters that permit anonymous requests to some endpoints, that filter can just decorate a request with no credentials with the "system:anonymous" user :)

what are the implications of an anonymous user (or authenticated-but-untrusted user, which is potentially all clusters) having control over Child relationships vs Link relationships?

dashpole: Ah, the important distinction is actually that a Link ignores the incoming span context (other than linking it), and doesn't use it for determining whether to sample or not.  If an anonymous user can create children, it can make all requests sampled.

In general, we want to be able to create children, rather than links, because debugging re-entrant API calls from webhook admission controllers is one of the core use-cases we want to solve.  Ideally we can do that without allowing anyone to generate a bunch of spans.  How do we solve that problem with audit logs?  Do we still audit unauthenticated requests?

brancz (commented):
Can't wait for this entire feature to be ready for use!

/lgtm
/approve
/hold 

(just a hold for @liggitt's comments to be resolved)

staging/src/k8s.io/apiserver/pkg/endpoints/filters/traces.go
liggitt: We do audit unauthenticated requests, though most unauthenticated requests would get stopped at the authorization layer.

liggitt: Putting this after the authorization check would mean you could only influence tracing for endpoints/verbs you were actually authorized to request.

I still don't have a clear picture of what "influencing tracing" means for an unauthenticated or authenticated-but-untrusted user.

If all users (anonymous or otherwise) are authorized to `GET /api/v1/namespaces/kube-public/configmaps/cluster-info` (as they are on kubeadm clusters), what info/headers/params could they pass in that would affect totally unrelated traces?

dashpole: It wouldn't affect completely unrelated traces, except that you might be able to DOS your tracing backend.

liggitt: > It wouldn't affect completely unrelated traces

does the info passed in via headers determine what is unrelated? is there anything an anonymous/untrusted requester could pass in in a header that would make tracing think this request was related to another request incorrectly?

logicalhan: they'd need access to that other request though.. otherwise how would they generate the trace id they shouldn't have?

dashpole: It is possible to append spans to an existing trace, if you know the trace ID, and one of the span IDs.  But spans are independent, so you wouldn't be able to alter any existing spans in the trace.

liggitt: > It wouldn't affect completely unrelated traces, except that you might be able to DOS your tracing backend. Is that something I should be worrying about?

Um... maybe? I see pros/cons the further back in the filter chain we put this:

* after authentication
  * pro: prevents requests with bad credentials from initiating traces (but not valid credentials or anonymous requests)
  * con: means tracing doesn't have visibility to costs of remote authentication methods like webhook authentication
* after priority-and-fairness
  * pro: prevents requests rejected because they exceed allowed priority-and-fairness threshholds from initiating traces
  * con: means tracing doesn't have visibility to requests delayed/rejected for this reason
* after authorization
  * pro: prevents unauthorized requests from initiating traces
  * con: means tracing doesn't have visibility to costs of remote authorization methods like webhook authorization

Even it we put it after all of those, it still sounds like an authorized request could saturate the tracing backend. What's the failure mode in that scenario?

> they'd need access to that other request though.. otherwise how would they generate the trace id they shouldn't have?

if the trace id is considered sufficiently obscure/secure to prevent abuse, that seems positive, and might limit them to influencing traces related to their own requests.

does a new trace id established for a request get returned to the user making that request? are there bad implications of them passing that into subsequent requests and establishing bogus child relationships (like the trace backend retaining the parent traces longer, running out of memory, etc)?

dashpole: The failure mode for saturation depends on the backend, but probably involves dropping trace data.

If a user sends a request without a trace id, it isn't ever returned to them.  But they can start a new trace (by creating a span without a parent themselves), and send it with the request to the apiserver.

The answer again will probably depend on the backend, but I dug into jaeger, and it looks like jaeger has a max duration for a trace (determined by the `max-span-age` flag) and it will ignore spans older than that age.  That would prevent creating traces that are ~infinitely long.  I haven't heard of that causing OOM, or anything like that.

dashpole: Having visibility into the costs of remote authentication would be really nice...

liggitt: > Having visibility into the costs of remote authentication would be really nice...

agreed... mostly trying to get my head around the risks this exposes. @lavalamp, you've thought about the abuse vectors more w.r.t. priority and fairness... do you have thoughts on this?

dashpole: I went ahead and added WithPublicEndpoint, which makes it ignore the incoming span context.  If that is acceptable, we can start with that and continue the discussion in an issue.

liggitt: sgtm, mind including a link to this thread (or hoisting relevant content) in the issue?

liggitt:
lgtm, looks like there might be a few commits that should be squashed down?

dashpole:
squashed

staging/src/k8s.io/apiserver/pkg/endpoints/filters/traces.go
```diff
@@ -0,0 +1,40 @@
...
+	opts := []otelhttp.Option{
+		otelhttp.WithPropagators(traces.Propagators()),
+		otelhttp.WithPublicEndpoint(),
+	}
+	if tp != nil {
+		opts = append(opts, otelhttp.WithTracerProvider(*tp))
+	}
+	// Even if there is no TracerProvider, the otelhttp still handles context propagation.
+	// See https://github.com/open-telemetry/opentelemetry-go/tree/main/example/passthrough
+	return otelhttp.NewHandler(handler, "KubernetesAPI", opts...)
```
liggitt, line 39: just realized this is in k8s.io/apiserver, not specific to kube-apiserver... if there are multiple servers built on this library, does each need to have a unique id, or is it ok if they all report `KubernetesAPI`

dashpole: This controls the name of the span, so it should be whatever would best identify the server, and it doesn't have to be unique.  https://github.com/kubernetes/kubernetes/pull/94942/files#diff-0e1453d49880a4f88f0353b9902c7e914513dd1274e3115100ea42c2d116136bR101-R106 is where we set the resource tags that should distinguish different instances of the apiserver.  Right now, we tag it with the service name, "kube-apiserver", and the API Server ID, which has the pattern kube-apiserver-<UUID>.

staging/src/k8s.io/apiserver/pkg/server/options/tracing.go
```diff
@@ -0,0 +1,124 @@
...
+	"google.golang.org/grpc"
+
+	"k8s.io/apiserver/pkg/server"
+	"k8s.io/apiserver/pkg/server/egressselector"
+	"k8s.io/apiserver/pkg/tracing"
+	"k8s.io/component-base/traces"
+	"k8s.io/utils/path"
+)
+
+const apiserverService = "kube-apiserver"
```
liggitt, line 38: things other than kube-apiserver use k8s.io/apiserver, this will need a follow-up to be able to be set by the particular binary consuming the k8s.io/apiserver library

staging/src/k8s.io/apiserver/pkg/endpoints/filters/traces.go
liggitt: ok... looks like APIServerID is overrideable by consumers of the k8s.io/apiserver library, so that's a reasonable thing to use... hard-coding kube-apiserver isn't good long-term, but can be adjusted in a follow-up

staging/src/k8s.io/apiserver/pkg/server/options/tracing.go
dashpole: Will rename to "apiserver".

PR #94950, opened 2020-09-21
Fix empty package license generation

dashpole:
**What type of PR is this?**

/kind cleanup

We currently skip vendor packages that don't have go files, but where subdirectories do have a go.mod file.  For example,
Skip `github.com/foo/bar` if `github.com/foo/bar/baz/go.mod` exists.
However, the current logic will not allow skipping `github.com/foo/bar` if `github.com/foo/bar/baz/abc/go.mod` exists.

This PR extends the logic added in https://github.com/kubernetes/kubernetes/pull/91366 to support that case.

Found while working on https://github.com/kubernetes/kubernetes/pull/94942.
I want to use `go.opentelemetry.io/contrib/instrumentation/net/http`, and it added `go.opentelemetry.io/contrib` to go.mod as well, but didn't put any files in vendor.

```release-note
NONE
```

/assign @liggitt @giuseppe

### DIFF
--- a/hack/update-vendor-licenses.sh
+++ b/hack/update-vendor-licenses.sh
@@ -197,18 +197,24 @@ for PACKAGE in $(go list -m -json all | jq -r .Path | sort -f); do
     continue
   fi
   # Skip a directory if 1) it has no files and 2) all the subdirectories contain a go.mod file.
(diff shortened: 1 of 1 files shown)


liggitt:
see also https://github.com/kubernetes/kubernetes/pull/92735

/cc @munnerz @dims
/unassign

PR #94964, opened 2020-09-22
Bug 1877793: Force releasing the lock on exit for KS

tnozicka:
I missed forcing release on cancel for KS in the hurry last time. Just noticed it while backporting.

/cc @soltysh @deads2k


tnozicka:
apologies, this was meant for our fork

PR #94968, opened 2020-09-22
fix session affinity related flaky cases in ipvs proxy mode

### LINKED ISSUE #94967: [Flaking Test][ipvs] should have session affinity timeout work for service with type clusterIP/NodePort
**Which jobs are flaking**:

https://k8s-testgrid.appspot.com/sig-network-gce#gci-gce-ipvs&width=5

Services should have session affinity timeout work for service with type clusterIP
Services should have session affinity timeout work for NodePort service

We have other flaky cases, like `ESIPP [Slow] should work from pods`, that should also bu resolved.
But `session affinity` cases also run in `pull-kubernetes-e2e-gci-gce-ipvs`, so I suggest we solve this problem first. 

**Testgrid link**:

https://k8s-testgrid.appspot.com/sig-network-gce#gci-gce-ipvs&width=5

**Reason for failure**:

In ipvs mode, we set session affinity timeout as 125s, the test case is:
1. Use run host command to send a request, mark returned hostName
2. sleep 125s
3. send another request, if return different hostName, test succed.

I think the reason is, in the second request, ipvs connection may still in sessionAffinity, send another request will result in seffinity restart count, eventually result in this test case flaky.

Lion-Wei:
**What type of PR is this?**
/kind failing-test
/kind flake

extend request interval to make session affinity cases stable

Fixes #94967

```release-note
NONE
```

### DIFF
--- a/test/e2e/network/service.go
+++ b/test/e2e/network/service.go
@@ -3408,7 +3408,19 @@ func execAffinityTestForSessionAffinityTimeout(f *framework.Framework, cs client
 			if hosts.Len() > 1 {
 				return
 			}
-			time.Sleep(time.Duration(svcSessionAffinityTimeout) * time.Second)
+			// In some case, ipvs didn't deleted the persistent connection after timeout expired,
+			// use 'ipvsadm -lnc' command can found the expire time become '13171233:02' after '00:00'
+			//
+			// pro expire state       source             virtual            destination
+			// TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
+			//
+			// pro expire state       source             virtual            destination
+			// TCP 13171233:02 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
+			//
+			// And 2 seconds later, the connection will be ensure deleted,
+			// so we sleep 'svcSessionAffinityTimeout+5' seconds to avoid this issue.
+			// TODO: figure out why the expired connection didn't be deleted and fix this issue in ipvs side.
+			time.Sleep(time.Duration(svcSessionAffinityTimeout+5) * time.Second)
 		}
 	}
 	framework.Fail("Session is sticky after reaching the timeout")


test/e2e/network/service.go
```diff
@@ -3456,7 +3456,7 @@ func execAffinityTestForSessionAffinityTimeout(f *framework.Framework, cs client
 			if hosts.Len() > 1 {
 				return
 			}
-			time.Sleep(time.Duration(svcSessionAffinityTimeout) * time.Second)
+			time.Sleep(time.Duration(svcSessionAffinityTimeout+1) * time.Second)
```
aojea, line 3459: is not better to change the variable directly?

Lion-Wei: `svcSessionAffinityTimeout` is the timeout second we set to `svc.Spec.SessionAffinityConfig`, and we'd better sleep `svcSessionAffinityTimeout+1` seconds to send another request.

aojea: ups, my fault, I've should read the whole context.
Can you add a comment explaining that so the next lazy person like me :sweat_smile:  don't ask the same question?

Lion-Wei: Done, PTAL. 😄

aojea:
/test pull-gci-gce-ipvs
let's see how it goes

aojea:
/lgtm
/hold
until the ipvs test pass, you can unhold when the test is good
/assign @andrewsykim

aojea:
ups
/lgtm cancel
```

Kubernetes e2e suite: [sig-network] Services should have session affinity timeout work for service with type clusterIP [LinuxOnly] [Conformance] expand_less | 22m0s
-- | --
test/e2e/framework/framework.go:629 Sep 23 08:10:58.522: Session is sticky after reaching the timeout test/e2e/network/service.go:3463

```

Lion-Wei:
Emmm, I have been tested many times in my local environment. I'll try to figure out the reason. 🤔

aojea:
> Emmm, I have been tested many times in my local environment. I'll try to figure out the reason.

can you add `+5` to see what happens and see if is a timing problem?
once we confirm it we can check for the best way to fix it

test/e2e/network/service.go
andrewsykim: Aren't we already accounting for a delay in the session affinity timeout?

https://github.com/kubernetes/kubernetes/blob/cf6a0868be9d684ae6e227f755444db1c2fd5b9f/test/e2e/network/service.go#L3403-L3408

andrewsykim:
Aren't we already accounting for a delay in the session affinity timeout?

https://github.com/kubernetes/kubernetes/blob/cf6a0868be9d684ae6e227f755444db1c2fd5b9f/test/e2e/network/service.go#L3403-L3408

aojea:
it seems it works now, will retest for double check
/test pull-kubernetes-e2e-gci-gce-ipvs

/lgtm

Lion-Wei:
@aojea Actually I still testing this, I'll convert to draft and take more time for local test, thanks.

aojea:
/lgtm cancel
ups sorry, I thought we were done and this was waiting for the ipvs PR you've sent

aojea:
@Lion-Wei I realised today that the test tries 10 times with a timeout interval to check if it receives 2 different pods,
https://github.com/kubernetes/kubernetes/blame/8058942aa2552bcaf13208d1ea678e9f3f5f4605/test/e2e/network/service.go#L3400-L3401

with current ipvs timeout the test can take too long ... maybe that is a clue to find the flakiness

Lion-Wei:
I tested add 5 more seconds, lvs can redirect the package completely follow the rr rule. So I'll `+5` and run more tests see what happened. Mean while, I'll learn more about lvs persistent logic, hopefully can find the root cause.

Lion-Wei:
Little update. I did some research, found out in some case, after the presistent timeout expired, the `ipvsadm -lnc` output of this connection is: 
```
TCP 00:01  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 13171233:02 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 13171233:02 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 13171233:02 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 13171233:02 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 13171233:02 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 13171233:01 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 13171233:01 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 13171233:01 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 13171233:01 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 13171233:01 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 13171233:01 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
TCP 13171233:01 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
```
This is the output of exec `ipvsadm -lnc` cyclically at 0.1s intervals.

Obviously, for some reason, lvs didn't delete this connection after timeout expired, and after about 2s, this connection be deleted.

So i'm sure `+5` will fix this problem, though I still can't understand why this happened, maybe if someone who familier with the lvs kernel will know why.

aojea:
> Little update. I did some research, found out in some case, after the presistent timeout expired, the `ipvsadm -lnc` output of this connection is:
> 
> ```
> TCP 00:01  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 13171233:02 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 13171233:02 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 13171233:02 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 13171233:02 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 13171233:02 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 13171233:01 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 13171233:01 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 13171233:01 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 13171233:01 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 13171233:01 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 13171233:01 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> TCP 13171233:01 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
> ```
> 
> This is the output of exec `ipvsadm -lnc` cyclically at 0.1s intervals.
> 
> Obviously, for some reason, lvs didn't delete this connection after timeout expired, and after about 2s, this connection be deleted.
> 
> So i'm sure `+5` will fix this problem, though I still can't understand why this happened, maybe if someone who familier with the lvs kernel will know why.

add this as a comment, so we can have the test working and if someone wants to keep investigating doesn't need to do all the research that you had to do :)

Lion-Wei:
Seems `Kubectl logs should be able to retrieve and filter logs` case failed in the last test.

I have run the `pull-kubernetes-e2e-gci-gce-ipvs` test many times, all passed, so I'll reopen this pr. Ready for another review, @aojea , @andrewsykim  : )

andrewsykim:
I'm still wondering about https://github.com/kubernetes/kubernetes/pull/94968#issuecomment-697933347

andrewsykim:
That timeout already adds 5 seconds to the 120s timeout, and this PR is already adding another 5s to that.

Lion-Wei:
@andrewsykim , sorry, I might missed that comment. The 5 more seconds in this pr, is because, when you set persistent time 125s to one virtual server, the persistence connection in lvs might exist more then 125s. In my test, if we wait 2s more after the 125s expired, will make sure session affinity invalid.

invidian:
How can I test this PR against my cluster? I'm affected by this too, so I'm happy to test it out!

andrewsykim:
> @andrewsykim , sorry, I might missed that comment. The 5 more seconds in this pr, is because, when you set persistent time 125s to one virtual server, the persistence connection in lvs might exist more then 125s. In my test, if we wait 2s more after the 125s expired, will make sure session affinity invalid.

So can we just increase the timeout to 130s in the original place instead of incrementing the timeout in twice?

aojea:
> How can I test this PR against my cluster? I'm affected by this too, so I'm happy to test it out!

You can build the e2e binary and run it against your cluster

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md#running-conformance-tests

invidian:
>You can build the e2e binary and run it against your cluster

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md#running-conformance-tests

Thanks, I'll test either today or on Friday.

Lion-Wei:
> So can we just increase the timeout to 130s in the original place instead of incrementing the timeout in twice?
I don't think so, the influence is, whatever session affinity time you set, the real affinity time might be 2s more.

aojea:
/lgtm
We are receiving lot of complaints with this test lately, adding an offset seems a good idea so the loadbalancer implementation can age the connection safely and avoid races, 

IIUIC right now the test does:
1. Check affinity (send 15 request and verify it goes to the same backend)
2. Enter a loop that polls 10 times the service every (timeout + 5) with a NEW connection
If only 1 backend is received it fails,

we don't want too strict with this, do we?

andrewsykim:
> I don't think so, the influence is, whatever session affinity time you set, the real affinity time might be 2s more.

Right, but the timeout here is hard coded timeout we use in the test anyways, we're not testing against dynamic affinity timeouts here:

https://github.com/kubernetes/kubernetes/blob/fb0e3400fd235587994e33953d241c3924e57bf9/test/e2e/network/service.go#L3354-L3361

andrewsykim:
Nvm, I see the problem now. We're increasing the time before checking the client IP so it needs to be a few seconds above the session affinity timeout we pass to Service, even if it's hardcoded.

test/e2e/network/service.go
```diff
@@ -3408,7 +3408,19 @@ func execAffinityTestForSessionAffinityTimeout(f *framework.Framework, cs client
 			if hosts.Len() > 1 {
 				return
 			}
-			time.Sleep(time.Duration(svcSessionAffinityTimeout) * time.Second)
+			// In some case, lvs didn't deleted the persistent connection after timeout expired,
```
andrewsykim, line 3411: we should probably say "ipvs" in the comments to be consistent with terminology in the rest of the repo.

test/e2e/network/service.go
```diff
@@ -3408,7 +3408,19 @@ func execAffinityTestForSessionAffinityTimeout(f *framework.Framework, cs client
...
+			// use 'ipvsadm -lnc' command can found the expire time become '13171233:02' after '00:00'
+			//
+			// pro expire state       source             virtual            destination
+			// TCP 00:00  NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
+			//
+			// pro expire state       source             virtual            destination
+			// TCP 13171233:02 NONE        10.105.253.160:0   10.105.253.160:80  10.244.1.25:9376
+			//
+			// And 2 seconds latter, the connection will be ensure deleted,
```
andrewsykim, line 3420: nit: s/latter/later/

Lion-Wei: Done. : )

aojea:
/retest
/lgtm
let's see if this is the good one :crossed_fingers: :smile:

andrewsykim (commented):
/approve

defer lgtm to @aojea

aojea:
/retest
    cacher_whitebox_test.go:1072: watcher is blocked by slower one (count: 22)

/lgtm

Thanks :crossed_fingers:

aojea:
:thinking: 
/retest

PR #94974, opened 2020-09-22
kubelet: add unit tests for imagePullSecrets keyring

andrewsykim:

**What type of PR is this?**
/kind cleanup

Add unit tests for generating kubelet's docker keyring for imagePullSecrets.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/credentialprovider/secrets/secrets_test.go
+++ b/pkg/credentialprovider/secrets/secrets_test.go
@@ -1,0 +1,286 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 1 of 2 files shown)


pkg/credentialprovider/secrets/secrets_test.go
```diff
@@ -0,0 +1,232 @@
...
+			authConfigs: []credentialprovider.AuthConfig{
+				{
+					Username: "user",
+					Password: "password",
+				},
+			},
+			found: false,
+		},
+		{
+			name:  "with pull secrets and has default keyring",
```
liggitt, line 179: what happens with default keyring and DockerConfigJsonKey or DockerConfigKey that doesn't match the given image?

andrewsykim: In this test case specifically, the default keyring would always return auth configs since the fake `Lookup()` call returns hardcoded auth configs. If we used a real keyring it would not return any auth configs if the image didn't match based on the check here:

https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/pkg/credentialprovider/keyring.go#L241-L245

liggitt: I was mostly looking for a test that exercised whether we fell back to the default or not for a DockerConfigJsonKey or DockerConfigKey that doesn't match the given image

andrewsykim: Ah, good point. Updated to include tests that only return credentials from the default keyring if DockerConfigJsonKey or DockerConfigKey didn't have a matching image URL passed into Lookup.

PR #94978, opened 2020-09-22
Add WithNamespace option to cli-runtime name printer

### LINKED ISSUE #94917: Support namespace printing option to NamePrinter

I'd like to add an option to [`k8s.io/cli-runtime/pkg/printers.NamePrinter`](https://github.com/kubernetes/kubernetes/blob/085beca6e137c13495996d82a2f8082fd378c85f/staging/src/k8s.io/cli-runtime/pkg/printers/name.go#L32-L41) in order to put namespace prefixes.

The output of the current [`NamePrinter.PrintObj`](https://github.com/kubernetes/kubernetes/blob/085beca6e137c13495996d82a2f8082fd378c85f/staging/src/k8s.io/cli-runtime/pkg/printers/name.go#L43-L87) is like:

```
configmap/test-cm-1 deleted
configmap/test-cm-1 deleted
configmap/test-cm-2 deleted
```

The output I'd like to opt-in is like:

```
my-namespace-1/configmap/test-cm-1 deleted
my-namespace-2/configmap/test-cm-1 deleted
my-namespace-2/configmap/test-cm-2 deleted
```

**Why is this needed**:

Some plugins sometimes do operations across all the namespaces. In such cases, printing namespaces would help the users to understand exactly what resources are operated.

micnncim:
**What type of PR is this?**

/kind feature

This PR adds an option to `k8s.io/cli-runtime/pkg/printers.NamePrinter` in order to put namespace prefixes.

The output of the current `NamePrinter.PrintObj` is like:

```
configmap/test-cm-1 deleted
configmap/test-cm-1 deleted
configmap/test-cm-2 deleted
```

The output I'd like to opt-in is like (if `WithNamespace` == `true`):

```
my-namespace-1/configmap/test-cm-1 deleted
my-namespace-2/configmap/test-cm-1 deleted
my-namespace-2/configmap/test-cm-2 deleted
```

Fixes #94917

```release-note
NONE
```

### DIFF
--- a/staging/src/k8s.io/cli-runtime/pkg/printers/name.go
+++ b/staging/src/k8s.io/cli-runtime/pkg/printers/name.go
@@ -40,7 +40,7 @@ type NamePrinter struct {
 	Operation string
 	// WithNamespace indicates whether a namespace should be
 	// printed along side the "resource/name" pair for an object.
-	// This flag effective useful when operations across all
+	// This flag is effective when operations across all
 	// namespace are done.
 	WithNamespace bool
 }
--- a/staging/src/k8s.io/cli-runtime/pkg/printers/name.go
+++ b/staging/src/k8s.io/cli-runtime/pkg/printers/name.go
@@ -40,6 +40,8 @@ type NamePrinter struct {
 	Operation string
 	// WithNamespace indicates whether a namespace should be
 	// printed along side the "resource/name" pair for an object.
+	// This flag effective useful when operations across all
+	// namespace are done.
 	WithNamespace bool
 }
 
--- a/staging/src/k8s.io/cli-runtime/pkg/printers/name.go
+++ b/staging/src/k8s.io/cli-runtime/pkg/printers/name.go
@@ -38,6 +38,9 @@ type NamePrinter struct {
 	// took place on an object, to be included in the
 	// finalized "successful" message.
 	Operation string
+	// WithNamespace indicates whether a namespace should be
+	// printed along side the "resource/name" pair for an object.
+	WithNamespace bool
 }
 
 // PrintObj is an implementation of ResourcePrinter.PrintObj which decodes the object
@@ -80,13 +83,21 @@ func (p *NamePrinter) PrintObj(obj runtime.Object, w io.Writer) error {
 	}
 
 	name := "<unknown>"
+	namespace := ""
 	if acc, err := meta.Accessor(obj); err == nil {
 		if n := acc.GetName(); len(n) > 0 {
 			name = n
 		}
+		if p.WithNamespace {
+			if ns := acc.GetNamespace(); len(ns) > 0 {
+				namespace = ns
+			} else {
+				namespace = "<unknown>"
+			}
+		}
 	}
 
-	return printObj(w, name, p.Operation, p.ShortOutput, GetObjectGroupKind(obj))
+	return printObj(w, name, namespace, p.Operation, p.ShortOutput, GetObjectGroupKind(obj))
 }
 
 func GetObjectGroupKind(obj runtime.Object) schema.GroupKind {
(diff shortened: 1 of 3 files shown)


staging/src/k8s.io/cli-runtime/pkg/printers/name.go
```diff
@@ -80,13 +83,21 @@ func (p *NamePrinter) PrintObj(obj runtime.Object, w io.Writer) error {
...
+	namespace := ""
 	if acc, err := meta.Accessor(obj); err == nil {
 		if n := acc.GetName(); len(n) > 0 {
 			name = n
 		}
+		if p.WithNamespace {
+			if ns := acc.GetNamespace(); len(ns) > 0 {
+				namespace = ns
+			} else {
+				namespace = "<unknown>"
```
knight42, line 95: There are cluster-scoped objects whose namespace is actually empty and I think it might be better to simply leave it empty.

micnncim: Thank you for your review. It makes sense so I've fixed that at [`45a69ab` (#94978)](https://github.com/kubernetes/kubernetes/pull/94978/commits/45a69ab7b93d60ec5fc98d2522f9801daaf9c263).

staging/src/k8s.io/cli-runtime/pkg/printers/name_test.go
```diff
@@ -0,0 +1,126 @@
...
+					},
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "test-pod",
+						Namespace: "test-namespace",
+					},
+				},
+			},
+			wantW:   "test-namespace/pod/test-pod\n",
+			wantErr: false,
+		},
```
knight42, line 99: Should we add a test about printing object without namespace(such as PersistentVolume)?

staging/src/k8s.io/cli-runtime/pkg/printers/name.go
```diff
@@ -38,6 +38,9 @@ type NamePrinter struct {
 	// took place on an object, to be included in the
 	// finalized "successful" message.
 	Operation string
+	// WithNamespace indicates whether a namespace should be
+	// printed along side the "resource/name" pair for an object.
+	WithNamespace bool
```
knight42, line 43: when would `WithNamespace` be true?

staging/src/k8s.io/cli-runtime/pkg/printers/name_test.go
micnncim: Added some additional test cases (https://github.com/kubernetes/kubernetes/pull/94978/commits/bbd5fe1fce2101d500c1ea2c51c74ad13b5fda16) 👍

staging/src/k8s.io/cli-runtime/pkg/printers/name.go
micnncim: Added a description about when to be enabled ([`4d46f00` (#94978)](https://github.com/kubernetes/kubernetes/pull/94978/commits/4d46f00b701309f920ff8fefd659a0a363a7cbba)).

micnncim:
@seans3 What do you think of this PR?

micnncim:
@seans3 @knight42 @soltysh @deads2k

Hi, could you take a look at this PR? If we need more discussion I'm willing to do so 🙂

soltysh (changes requested):
/hold
I don't think that printing resource as `ns/kind/name` is helpful. I'm worried this will only cause additional confusion. The reason we previously went with `kind/name` is that you can copy&paste that output to other kubectl commands. The command that allow cross-namespace functionality should be managing printing in such corner-cases manually. Have a look at how this is handled in `kubectl get pods --all-namespaces`, for example.

PR #94982, opened 2020-09-22
Revert "Merge pull request #93837 from jayunit100/DialFromContainerB"

aojea:
This reverts commit 61490bba46ee2ad39690c682db0bee425b21798e, reversing
changes made to 9ecab1b4b2a76aa998dea2bdfa9eb9d892103f42.

Some methods from the networking e2e tools are dialing from a
containter to another container, and failing the test if there was no
connectivity. This PR modified the methods to return an error instead of
failing the test.
However, these methods were used by other tests in the framework, and
they are not checking if the method returns an error, expecting that
the method fail the test. With this change, any connectivity problem
will go unnoticed on the tests that are not asserting the error, so we
need to revert to previous state.

/kind bug
/kind failing-test

xref: https://github.com/kubernetes/kubernetes/pull/93837#discussion_r492991061

```release-note
NONE
```

### DIFF
--- a/test/e2e/common/networking.go
+++ b/test/e2e/common/networking.go
@@ -18,7 +18,6 @@ package common
 
 import (
 	"github.com/onsi/ginkgo"
-	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2enetwork "k8s.io/kubernetes/test/e2e/framework/network"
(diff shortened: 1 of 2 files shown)


danwinship:
AIUI returning an error rather than failing directly from the helper function is the preferred style (both because it makes the ginkgo error show a more useful file+line, and because it allows the helper to be used both in tests where you expect success and tests where you expect failure). So ultimately, the right fix is to fix all the other users of these functions, not to revert the change. But if that's too much work and this is breaking things badly then reverting might be the right short-term fix.

(I don't have /approve bits here.)

aojea:
>  So ultimately, the right fix is to fix all the other users of these functions, not to revert the change. But if that's too much work and this is breaking things badly then reverting might be the right short-term fix.

I prefer to revert and merge it correctly, I'm not specially thrilled by the change, I prefer the tests to fail, this may hide flakes since basically is adding a retry inside the test.

 
> this is breaking things badly 

this is my concern, the tests are not going to fail and we can merge something broken ... and since the gates opened yesterday :shrug:

jayunit100:
Anyway I can take on finding which tests are erroneously expecting failure semantics ?  I was hoping to clean this up so that the error can be handled in a separate step , bc mixing the semantics makes it tricky to audit large cluster  test failures on a per node basis and other general code hygiene  :):)

jayunit100:
talked to Antonio on slack - im going to rebuild the commit on top of this which refactors the tests (other then intra-pod, which uses this error intelligently) to fail fast

aojea:
> talked to Antonio on slack - im going to rebuild the commit on top of this which refactors the tests (other then intra-pod, which uses this error intelligently) to fail fast

/close
Thanks Jay for the quick response

PR #94984, opened 2020-09-22
check readyz before adding endpoint for kube-apiserver

deads2k:
/healthz and /readyz don't mean the same thing.  this bit of code must not have been updated when we made the switch so long ago.


dims:
/release-note-none
/assign @liggitt 

LGTM

ialidzhikov:
@deads2k , does it make sense to file cherry-picks of this PR?

PR #94986, opened 2020-09-22
drop managed fields from audit entries

tkashem:
**What type of PR is this?**

/kind feature

Implements https://github.com/kubernetes/enhancements/pull/2982

This PR adds a new field `OmitManagedFields`  to both `audit.Policy` and `audit.PolicyRule` so cluster operators can opt in to omit managed fields of the request and response bodies from being written to the API audit log.

```release-note
A new field `omitManagedFields` has been added to both `audit.Policy` and `audit.PolicyRule` 
so cluster operators can opt in to omit managed fields of the request and response bodies from 
being written to the API audit log.
```


staging/src/k8s.io/apiserver/pkg/audit/request.go
```diff
@@ -153,9 +153,13 @@ func LogRequestObject(ae *auditinternal.Event, obj runtime.Object, gvr schema.Gr
 		return
 	}
 
+	copy := copyWithoutManagedFields(obj)
+	if copy == nil {
+		copy = obj
```
sttts, line 158: this is not a good pattern. If you have a func that cannot do its work, either have an error as return type, or at least a `ok` bool.

staging/src/k8s.io/apiserver/pkg/audit/request.go
```diff
@@ -244,3 +261,46 @@ func maybeTruncateUserAgent(req *http.Request) string {
...
+			return nil, false, err
+		}
+	}
+
+	if isList {
+		if err := meta.EachListItem(copy, removeManagedFields); err != nil {
+			return nil, false, err
+		}
+	}
+
```
tkashem, line 304: `Table` is not covered here.

sttts: put a todo

sttts: Does a table have managedfields at all? I don't think.

tkashem: Yes, I see `managedFields` in the elements under `rows`

staging/src/k8s.io/apiserver/pkg/audit/request.go
```diff
@@ -193,8 +201,17 @@ func LogResponseObject(ae *auditinternal.Event, obj runtime.Object, gv schema.Gr
 	if ae.Level.Less(auditinternal.LevelRequestResponse) {
 		return
 	}
-	// TODO(audit): hook into the serializer to avoid double conversion
+
 	var err error
+	copy, ok, err := copyWithoutManagedFields(obj)
+	if err != nil {
+		klog.Warningf("Audit failed while dropping managed fields from the response for %q error: %v", reflect.TypeOf(obj).Name(), err)
```
sttts, line 208: not failed, right?

staging/src/k8s.io/apiserver/pkg/audit/request.go
```diff
@@ -244,3 +261,46 @@ func maybeTruncateUserAgent(req *http.Request) string {
...
+	if _, err := meta.Accessor(obj); err != nil {
+		isAccessor = false
+	}
+	isList := meta.IsListType(obj)
+
+	if !isAccessor && !isList {
+		return nil, false, nil
+	}
+
+	copy := obj.DeepCopyObject()
```
sttts, line 291: would move the deepcopy into the two if statements

staging/src/k8s.io/apiserver/pkg/audit/request.go
```diff
@@ -153,8 +153,16 @@ func LogRequestObject(ae *auditinternal.Event, obj runtime.Object, gvr schema.Gr
 		return
 	}
 
-	// TODO(audit): hook into the serializer to avoid double conversion
 	var err error
+	copy, ok, err := copyWithoutManagedFields(obj)
```
sttts, line 157: do you need the var above?

staging/src/k8s.io/apiserver/pkg/audit/request.go
tkashem: from this point forward a deep copy is required, so thought to do it here once.

staging/src/k8s.io/apiserver/pkg/audit/request.go
tkashem: no, removed it.

staging/src/k8s.io/apiserver/pkg/audit/request.go
tkashem: changed it.

staging/src/k8s.io/apiserver/pkg/audit/request.go
```diff
@@ -244,3 +259,62 @@ func maybeTruncateUserAgent(req *http.Request) string {
...
+		isAccessor = false
+	}
+	isList := meta.IsListType(obj)
+
+	_, isTable := obj.(*metav1.Table)
+	if !isAccessor && !isList && !isTable {
+		return nil, false, nil
+	}
+
+	copy := obj.DeepCopyObject()
```
tkashem, line 290: Hi @wojtek-t, the deep copy is happening:
- only when the user is logging at `LevelRequest` and `LevelRequestResponse` level
- the deep copy is done only when it's necessary.

I would like your thoughts on the deep copy here, I understand that a deep copy can be expensive.

apelisse:
Do we want to always drop managed fields? I could see why we would want to also keep it under some circumstances.
I believe that audit logging can be configured, could this become an option?

tkashem:
> Do we want to always drop managed fields? I could see why we would want to also keep it under some circumstances.
> I believe that audit logging can be configured, could this become an option?

I think it's a good idea to make it an option, let me look into it.

sttts:
/assign @wojtek-t @liggitt 

Is this of general interest? We see this overhead of managedFields in (deep) audit logs of information that is pretty useless during audit.

staging/src/k8s.io/apiserver/pkg/audit/request.go
wojtek-t: This is once per-request, we are doing more copies/serializations/deserializations/etc. for a given request. That sounds fine to me (I mean, if we can avoid that somehow, let's avoid, but I don't see an easy way for it).

staging/src/k8s.io/apiserver/pkg/audit/request.go
```diff
@@ -244,3 +259,62 @@ func maybeTruncateUserAgent(req *http.Request) string {
...
 }
+
+// copyWithoutManagedFields will make a deep copy of the specified object and will discard
+// managed fields from the copy.
+// The specified object is expected to be a meta.Object or a "list". The specified object obj
+// is treated as readonly and hence not mutated.
+// On return, an error is set if the function runs into any error while removing the managed fields.
+// On return, the boolean value is true if the copy has been made successfully, otherwise false.
+func copyWithoutManagedFields(obj runtime.Object) (runtime.Object, bool, error) {
+	removeManagedFields := func(obj runtime.Object) error {
```
soltysh, line 270: Nit: can we move this to regular function. I'm not personally a fan of inner functions.

tkashem:
> Do we want to always drop managed fields? I could see why we would want to also keep it under some circumstances.
> I believe that audit logging can be configured, could this become an option?

@apelisse @sttts I added an option, by default, no managed fields will be dropped, keeping it compatible with current behavior. If an admin wants to drop managed fields the `Policy` object needs to have the following annotation:
```
    apiVersion: audit.k8s.io/v1beta1
    kind: Policy
    metadata:
      name: WriteRequestBodies
      annotations:
        audit.kubernetes.io/omit-managed-fields: true
```

tkashem:
Based on the conversation with @sttts, I am going to remove the annotation and add it as an apiserver flag `--audit-drop-managed-fields`

apelisse:
I haven't followed this quickly but I'd really like to have the option to select at runtime whether this field needs to be persisted or not. I do believe there are use-cases where this data is valuable to be saved and audited.

lavalamp:
An option in the policy object sounds more correct.

My first thought is that I'd split the "metadata" level into minimal-metadata and all-metadata; then we know user intent and can automatically classify new fields correctly in the future. Levels asking for more data than just metadata, e.g. request body, shouldn't be dropping any part of e.g. the request body.

lavalamp:
this is in my queue but I haven't gotten to it yet

staging/src/k8s.io/apiserver/pkg/audit/request.go
```diff
@@ -244,3 +284,62 @@ func maybeTruncateUserAgent(req *http.Request) string {
 
 	return ua
 }
+
+// copyWithoutManagedFields will make a deep copy of the specified object and will discard
```
lavalamp, line 288: why does the copy have to be deep? I'd expect shallow is fine for everything not a direct ancestor of managed fields?

staging/src/k8s.io/apiserver/pkg/audit/request_test.go
```diff
@@ -54,3 +60,166 @@ func TestMaybeTruncateUserAgent(t *testing.T) {
...
+				return
+			}
+
+			if err != nil {
+				t.Errorf("expected error to be nil, but got %v", err)
+			}
+			if !ok {
+				t.Error("expected ok to be true, but got false")
+			}
+			if !reflect.DeepEqual(test.expected, objectGot) {
```
lavalamp, line 220: We use `cmp.Equal` nowadays, I think.

This tests the "copy" part but not the "deep" part-- if that's important it should be verified. (And if it's not important it should not be done!)

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/create.go
```diff
@@ -130,7 +130,7 @@ func createHandler(r rest.NamedCreater, scope *RequestScope, admit admission.Int
 
 		ae := request.AuditEventFrom(ctx)
 		admit = admission.WithAudit(admit, ae)
-		audit.LogRequestObject(ae, obj, scope.Resource, scope.Subresource, scope.Serializer)
+		audit.LogRequestObject(req, obj, scope.Resource, scope.Subresource, scope.Serializer)
```
lavalamp, line 133: Since you have it here, why not pass ctx directly?

lavalamp (commented):
A few comments

staging/src/k8s.io/apiserver/pkg/audit/checker.go
```diff
@@ -0,0 +1,33 @@
...
+
+package audit
+
+import (
+	"k8s.io/apiserver/pkg/apis/audit"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+)
+
+// PolicyChecker exposes methods for checking the policy rules.
+type PolicyChecker interface {
```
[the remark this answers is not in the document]
tkashem: +1 to rename, how about `PolicyRuleEvaluator`?

staging/src/k8s.io/apiserver/pkg/audit/context.go
```diff
@@ -82,3 +85,14 @@ func auditAnnotationsFrom(ctx context.Context) []annotation {
...
 }
+
+// WithAuditPolicyChecker sets the audit policy checker interface for a given request.
+func WithAuditPolicyChecker(parent context.Context, checker PolicyChecker) context.Context {
+	return genericapirequest.WithValue(parent, auditPolicyCheckerKey, checker)
+}
+
+// AuditPolicyCheckerFrom returns the audit policy checker interface from the given context.
+func AuditPolicyCheckerFrom(ctx context.Context) PolicyChecker {
+	checker, _ := ctx.Value(auditPolicyCheckerKey).(PolicyChecker)
```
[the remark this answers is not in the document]
tkashem: yes, I am going with `(PolicyChecker, bool) `, this is more idiomatic.

staging/src/k8s.io/apiserver/pkg/audit/policy/checker.go
```diff
@@ -74,6 +69,14 @@ func (p *policyChecker) LevelAndStages(attrs authorizer.Attributes) (audit.Level
 	return DefaultAuditLevel, p.OmitStages
 }
 
+func (p *policyChecker) IsOmitManagedFieldsEnabled() bool {
+	if v, exists := p.Policy.Annotations["audit.kubernetes.io/omit-managed-fields"]; exists && v == "true" {
```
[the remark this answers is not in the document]
tkashem: This is from a `Policy` object:
```
apiVersion: audit.k8s.io/v1
    kind: Policy
    metadata:
      annotations: 
        audit.kubernetes.io/omit-managed-fields:  'true'
    rules:

```
It's enabled on a global scope, not at the `rule` scope. If we want this to be more granular then we would need to extend the Policy API.

staging/src/k8s.io/apiserver/pkg/audit/request.go
```diff
@@ -164,7 +181,13 @@ func LogRequestObject(ae *auditinternal.Event, obj runtime.Object, gvr schema.Gr
 }
 
 // LogRequestPatch fills in the given patch as the request object into an audit event.
-func LogRequestPatch(ae *auditinternal.Event, patch []byte) {
+func LogRequestPatch(req *http.Request, patch []byte) {
+	ctx := req.Context()
+	if ctx == nil {
```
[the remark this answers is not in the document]
tkashem: If this were to actually happen, we would not lose audit logs, I believe. We would see this entry in the audit logs but the `RequestObject.Raw` field of the audit event would be empty.

staging/src/k8s.io/apiserver/pkg/audit/request.go
```diff
@@ -193,6 +222,17 @@ func LogResponseObject(ae *auditinternal.Event, obj runtime.Object, gv schema.Gr
 	if ae.Level.Less(auditinternal.LevelRequestResponse) {
 		return
 	}
+
+	if policyChecker := AuditPolicyCheckerFrom(ctx); policyChecker != nil && policyChecker.IsOmitManagedFieldsEnabled() {
```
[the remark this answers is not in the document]
tkashem: Done

staging/src/k8s.io/apiserver/pkg/audit/request.go
tkashem: I am not certain if shallow copy is safe and has no side effect. If you think it's safe I can remove the deep copy.

staging/src/k8s.io/apiserver/pkg/audit/request_test.go
tkashem: +1 to `cmp.Equal` 
I added a step that verifies that deep copy is done.

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/create.go
tkashem: Done.

tkashem:
@lavalamp I have addressed your comments, please take a look when you have time.

staging/src/k8s.io/apiserver/pkg/audit/checker.go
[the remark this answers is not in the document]
lavalamp: Sure, that's better

staging/src/k8s.io/apiserver/pkg/audit/request.go
lavalamp: I can't think of a reason why a shallow copy would be problematic here. @wojtek-t for a second opinion, but I think avoiding a deep copy on every object that's audited would be really good.

staging/src/k8s.io/apiserver/pkg/audit/request.go
lavalamp: Why would this require a deep copy? it only exists for long enough to be serialized?

staging/src/k8s.io/apiserver/pkg/audit/policy/checker.go
```diff
@@ -74,6 +70,21 @@ func (p *policyChecker) LevelAndStages(attrs authorizer.Attributes) (audit.Level
...
 
+func (p *policyRuleEvaluator) IsOmitManagedFieldsEnabled() bool {
+	// If we can't decide, return false to maintain current behavior which is
+	// to retain the manage fields in the audit.
+	value, exists := p.Policy.Annotations["audit.kubernetes.io/omit-managed-fields"]
+	if !exists {
+		return false
+	}
+
+	if omitManagedFields, err := strconv.ParseBool(value); err == nil {
```
liggitt, line 81: is this re-parsing the policy annotation on every check of every request? can we do this once since the policy isn't dynamic?

staging/src/k8s.io/apiserver/pkg/audit/request.go
liggitt: do we have a way to do a shallow copy of an interface object reliably?

staging/src/k8s.io/apiserver/pkg/audit/policy/checker.go
tkashem: @liggitt thanks for the suggestion, I have moved the parsing logic to the construction of of the policy rule  evaluator interface.

tkashem:
> do we have a way to do a shallow copy of an interface object reliably?

not that I am aware of, didn't find any package in use today.

staging/src/k8s.io/apiserver/pkg/audit/request.go
tkashem: > I am not certain if shallow copy is safe and has no side effect

I spoke too soon, a shallow copy should be safe since we are dropping `managedFields` slice. I did not come across any utility in k8s to shallow copy an interface. There might be a way to use `reflect` (will introduce some cpu overhead) to build a shallow copy, I will look into it.

sttts:
First commit has refactoring and new feature mixed up. Can you first refactor without behaviour change and then add the feature (should be trivial then) ?

tkashem:
> First commit has refactoring and new feature mixed up. Can you first refactor without behaviour change and then add the feature (should be trivial then) ?

@sttts yes, refactor first, and then the changes, it is lot cleaner now.

staging/src/k8s.io/apiserver/pkg/audit/request.go
tkashem: I did not come across any utility in k8s to shallow copy from an interface object. One option is to use the `reflect` package, traverse all the fields of the type recursively and build a shallow copy. I am not sue if this is the route we want to go.

I don't know if the serializer can be tweaked to omit certain fields or one can provide custom `MarshalJSON` function. But such extension is not worth pursuing in my opinion, since it can lead to bugs in the system due to improper use.

@lavalamp @sttts @liggitt please let me know how you want me to move forward with this.

lavalamp: If it's excessively hard it's probably worth just doing the easier thing and measuring to see if it's actually a performance problem.

apelisse:
Hi @tkashem, would you be OK trying to get this merged somehow? I lost track because I thought it was going well, but apparently it's not. I'd love to help you finish if needed. Thanks!

tkashem:
> Hi @tkashem, would you be OK trying to get this merged somehow? I lost track because I thought it was going well, but apparently it's not. I'd love to help you finish if needed. Thanks!

@apelisse Yes, I am happy to work on it. I just reopened and rebased the PR. Right now it uses a deep-copy approach. we can merge it now and do some optimization later if there are any performance issues. I think deep copy is performed multiple times in different areas for a request, so we can come up with a way to measure the impact of deep copy in a separate PR.
Also, it can be disabled by removing the annotation and by default it is disabled. Thoughts?

apelisse:
OK Now is not the best time to do that because of code-freeze, but I'll try to give it a look.

apelisse:
What's the status on this? Let me know when I need to take a look, thanks!

tkashem:
@apelisse I think this is the only remaining topic - https://github.com/kubernetes/kubernetes/pull/94986#discussion_r587925572 we need to resolve.

if there was a utility that does shallow copy of a runtime object, I assume it would be based on `reflect` and it probably would be more expensive computationally as opposed to a deep copy.

apelisse:
Yeah, I don't know how blocking this is. Do we have any way to measure the impact of doing this? Also, dropping managed fields will have a positive impact on other parts of the system, so the trade-off is not so easy to compute. Also it looks like if the performance degrades too much, people can revert the option and we can fix it, so it doesn't sound like there wouldn't be a work-around. On the other hand, for people who struggle with the field, that can be positive. I see more positive than negative?

tkashem:
- It is disabled by default, so the cluster operator will need to opt in to enable this feature.
- if performance degrades, then the cluster operator can disable it.
So yes, I agree with you that it has more positive than negatives.

> Do we have any way to measure the impact of doing this?

deep copy is done 5 to 6 times in other places of the apiserver for a request in flight. Do we want to measure how much deep copy takes in general or go with a more local approach - add `Tracing` if `copyWithoutManagedFields` takes more than `25ms` (rough threshold)? If we are fine with a local approach, i can add the tracing.

apelisse:
I'd just move forward like this, but I'm not going to approve this PR. If this has the deep-copy in already, it sounds like it's ready for review then?

tkashem:
Yes, It was reviewed before by @sttts and @lavalamp. I had to do a rebase after that. Now, it's ready for another pass. Please take a look when you have time.

tkashem:
@apelisse The PR has gotten big - if it helps I can move the "rename only" commit https://github.com/kubernetes/kubernetes/pull/94986/commits/9c09b5c9a0e01d337eb7e46e005952f5c9beab67 into its own PR. Let me know.

apelisse:
Actually, I think it'd be great. I was a little confused to see that you had a refactoring along with the new feature in the same PR. That'd make your life of rebasing much easier too.

tkashem:
Sounds good, I am on it, I will ping you once it's ready.

tkashem:
Rename moved to a new PR: https://github.com/kubernetes/kubernetes/pull/104350. Please take a look.

tkashem:
@apelisse the refactor PR has merged, please take a look when you have time.

apelisse:
Looks good to me, I'd love to have another pair of eyeballs on this since I'm not familiar with this code area in general. Thanks!

tkashem:
> Looks good to me, I'd love to have another pair of eyeballs on this since I'm not familiar with this code area in general. Thanks!

@apelisse absolutely, do you have anyone in mind? @sttts and @lavalamp had reviewed this before. (it does not have any new changes since they reviewed)

wojtek-t:
> @apelisse absolutely, do you have anyone in mind? @sttts and @lavalamp had reviewed this before. (it does not have any new changes since they reviewed)

Given people are generally ok with it - let me take this and make a deeper double-check.

staging/src/k8s.io/apiserver/pkg/endpoints/filters/audit.go
```diff
@@ -141,6 +141,7 @@ func createAuditEventAndAttachToContext(req *http.Request, policy audit.PolicyRu
 	}
 
 	req = req.WithContext(request.WithAuditEvent(ctx, ev))
+	req = req.WithContext(audit.WithAuditPolicyRuleEvaluator(req.Context(), policy))
```
wojtek-t, line 144: WithContext is allocating a new http.Request object.

Instead of doing that twice we should do:
```
ctx = request.WithAuditEvent(ctx, ev)
ctx = audit.WidthAuditPolicyRuleEvaluator(ctx, policy)
req = req.WithContext(ctx)
``

File: staging/src/k8s.io/apiserver/pkg/audit/request.go (hunk 8 above)
Comment (wojtek-t) [reply staging/src/k8s.io/apiserver/pkg/audit/request.go:271]: 1. I agree that deep-copy isn't really needed here - shallow is enough.

2. I can live with deep-copy for now, but please add a TODO to figure out how to migrate to shallow copy here.
I don't have an easy way from the top of my head to that in a generic case.

File: staging/src/k8s.io/apiserver/pkg/audit/request.go [hunk 18]
```diff
@@ -242,3 +267,71 @@ func maybeTruncateUserAgent(req *http.Request) string {
...
+			return nil, false, err
+		}
+	}
+
+	if isTable {
+		table := copy.(*metav1.Table)
+		for i := range table.Rows {
+			rowObj := table.Rows[i].Object
+
+			if rowObj.Object == nil {
```
wojtek-t, line 307: Can we move checking 'obj != nil' to the removeManageFields function? [This way we would cover more potential corner cases]

staging/src/k8s.io/apiserver/pkg/audit/request.go
wojtek-t: As mentioned above - I don't think it has to be a deep-copy, but I don't know how to generically make a shallow copy here. I requested a TODO above.

staging/src/k8s.io/apiserver/pkg/audit/policy/checker.go
```diff
@@ -70,6 +76,32 @@ func (p *policyRuleEvaluator) LevelAndStages(attrs authorizer.Attributes) (audit
 	return DefaultAuditLevel, p.OmitStages
 }
 
+func (p *policyRuleEvaluator) IsOmitManagedFieldsEnabled() bool {
+	return p.omitManagedFields
+}
+
+func isOmitManagedFieldsEnabled(policy *audit.Policy) bool {
+	const annotationKey = "audit.kubernetes.io/omit-managed-fields"
```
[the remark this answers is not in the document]
tkashem: maybe it should go to the API, but for now I moved it to line 32. let's wait to hear from @lavalamp

staging/src/k8s.io/apiserver/pkg/endpoints/filters/audit.go
tkashem: done

staging/src/k8s.io/apiserver/pkg/audit/request.go
tkashem: done

staging/src/k8s.io/apiserver/pkg/audit/request.go
tkashem: yes, I have added a TODO

staging/src/k8s.io/apiserver/pkg/audit/request.go
```diff
@@ -242,3 +267,72 @@ func maybeTruncateUserAgent(req *http.Request) string {
...
+	if isList {
+		if err := meta.EachListItem(copy, removeManagedFields); err != nil {
+			return nil, false, err
+		}
+	}
+
+	if isTable {
+		table := copy.(*metav1.Table)
+		for i := range table.Rows {
+			rowObj := table.Rows[i].Object
```
tkashem, line 308: can `rowObj` ever be empty? (by general convention it should not be).  maybe add a check `if rowObj == nil { continue }`?

staging/src/k8s.io/apiserver/pkg/audit/policy/checker.go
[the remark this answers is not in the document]
tkashem: okay, I have added the constant to `v1/types`, i don't think I need to add it to all versions, right?

staging/src/k8s.io/apiserver/pkg/audit/policy/checker_test.go
```diff
@@ -345,3 +346,75 @@ func TestUnionStages(t *testing.T) {
 		}
 	}
 }
+
+func TestIsOmitManagedFieldsEnabled(t *testing.T) {
+	omitManagedFieldsKey := "audit.kubernetes.io/omit-managed-fields"
```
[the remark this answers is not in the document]
apelisse: I would actually hard-code the value in the tests rather than share the value through a global variable.

staging/src/k8s.io/apiserver/pkg/apis/audit/v1/types.go
```diff
@@ -31,6 +31,13 @@ const (
 	HeaderAuditID = "Audit-ID"
 )
 
+const (
+	// OmitManagedFieldsAnnotationKey is the name of the annotation for a Policy object, if
+	// set to 'true', the audit backend will omit 'managedFields' from the audit entries
+	// for both request and response.
+	OmitManagedFieldsAnnotationKey = "audit.kubernetes.io/omit-managed-fields"
```
wojtek-t, line 38: @lavalamp - I suggested moving this annotation key to the API, but I'm not 100% convinced.
But either way - I would appreciate your feedback (and if case it's needed review for it).

[I guess we want to add that in other versions too then]

wojtek-t:
@lavalamp - can you please advise on https://github.com/kubernetes/kubernetes/pull/94986#r708520133 ?

staging/src/k8s.io/apiserver/pkg/apis/audit/v1/types.go
lavalamp: I agree this would be a lot better as an actual field on the policy object.

tkashem: @lavalamp please let me know if this looks good to you

```
type Policy struct {
	// +optional
	OmitManagedFields bool `json:"omitManagedFields,omitempty" protobuf:"varint,4,opt,name=omitManagedFields"`
}

type PolicyRule struct {
	// +optional
	OmitManagedFields *bool `json:"omitManagedFields,omitempty" protobuf:"varint,4,opt,name=omitManagedFields"`
}
```

it's in both `Policy` and `PolicyRule` so the cluster operator has an option to use it in a more granular way:
- the `bool` in `Policy` dictates the global default policy.
- the `*bool` in `PolicyRule` can override the global default for a particular matching rule, with three interpretations:
  - `nil` will represent 'not specified by the operator', hence use the global default.
  - `{false|true}` will override the global default.

lavalamp: Yes, that sounds great, thanks!

ritazh:
/hold
After reviewing this in the sig auth triage meeting, we believe this requires a KEP. Feel free to discuss this in the sig auth community call or mailing list.

lavalamp:
> we believe this requires a KEP

@ritazh what is the primary problem which a KEP will solve? It seems like a lot of overhead. Perhaps a few lines can be added to the existing SSA KEP instead?

ritazh:
> Perhaps a few lines can be added to the existing SSA KEP instead?

I think that's reasonable. WDYT? @enj

tkashem:
API changes have been made according to the KEP - https://github.com/kubernetes/enhancements/pull/2982. This is ready for another pass.

@wojtek-t @apelisse @sttts

staging/src/k8s.io/apiserver/pkg/audit/request.go
liggitt: looks like this comment is outstanding

staging/src/k8s.io/apiserver/pkg/audit/request.go
liggitt: do we know the relative cost (cpu/allocs) of adding a deep copy into the audit pipeline by enabling the omitManagedFields option globally?

liggitt (commented):
audit policy config lgtm

staging/src/k8s.io/apiserver/pkg/audit/request.go
tkashem: I think it would be similar to the other deep copies we have in place today. I have not done any benchmarking to measure it. I opened this issue so we can track it - https://github.com/kubernetes/kubernetes/issues/106094

staging/src/k8s.io/apiserver/pkg/audit/request.go
tkashem: `rawObj` can't be `nil` - it's the `Object` field of type `runtime.RawExtension` from `TableRow`
```
type TableRow struct {
	Cells []interface{} `json:"cells"`
	Conditions []TableRowCondition `json:"conditions,omitempty"`
	Object runtime.RawExtension `json:"object,omitempty"`
}
```

I have added an empty  `TableRow` object to my test to ensure it does not panic.

tkashem:
@liggitt @sttts I added an empty `TableRow` object to my test to cover the `nil` case. that's the only change.

Issue #94487, opened 2020-09-03
test/e2e/upgrade tests are not running


liggitt:
**Which test(s) are failing**:

All tests in https://github.com/kubernetes/kubernetes/tree/master/test/e2e/upgrades

**Which jobs are failing**:

All jobs running those tests, e.g. https://prow.k8s.io/job-history/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-stable2-stable1-upgrade-cluster

**Since when has it been failing**:

1.18 (see https://github.com/kubernetes/release/pull/868#issuecomment-673727541)

**Reason for failure**:

The immediate cause is that the bringup scripts for the upgrade tests expect sha1 hash files that were removed in 1.18. If that was fixed, I don't know what the next blocker would be.

**Anything else we need to know**:

Lack of upgrade test signal puts the project at risk for introducing regressions. Examples of things these tests caught in the past or that we have open issues to add to our upgrade test suite:
- https://github.com/kubernetes/kubernetes/issues/64789#issuecomment-394883173
- https://github.com/kubernetes/kubernetes/issues/78904
- https://github.com/kubernetes/kubernetes/issues/78633

It also blocks ongoing dev/feature work:
- https://github.com/kubernetes/enhancements/pull/1912#discussion_r463692617
- https://github.com/kubernetes/kubernetes/pull/93455#issuecomment-686539615

cc @kubernetes/sig-testing @BenTheElder

vinayakankugoyal:
Any update on this? I don't see that this has been assigned to any one. PR #93455 has been blocked on this for 15 days.

liggitt:
@zshihang is working on parts of this (xref https://github.com/kubernetes/kubernetes/pull/94835)

BenTheElder:
@vinayakankugoyal I have a longer term goal to look at how we do upgrade-type testing of Kubernetes strategically, but we're not planning on investing in this test tooling further, and it's the cluster upgrade/deployment mechanism that is broken currently, not the tests. 

The lack of keeping after the signal for this / responsive ownership of failures caused the release team to refuse to monitor it further, leading us to the current state where it doesn't run at all. For starters someone will need to mitigate the dependency on deprecated binary release hashes in the older release branches.

>  I don't see that this has been assigned to any one. PR #93455 has been blocked on this for 15 days.

[...]

PR #78302, opened 2019-05-24
Add the ability to migrate CoreDNS configmap in kube-up

rajansandeep:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:
Following CoreDNS ConfigMap changes in #78030, it is required that the ConfigMap gets updated during an `upgrade`. 
If I understand correctly, `EnsureExists` doesn't update the ConfigMap during an upgrade and requires `Reconcile` for the ConfigMap to upgrade.

This is evident from the logs:

Logs from `ci-kubernetes-e2e-gce-new-master-upgrade-cluster`:
```
event for coredns-5cd44cdc5d-lzbmv: {kubelet bootstrap-e2e-minion-group-b8n6} Unhealthy: Readiness probe failed: Get http://10.64.2.20:8181/ready: dial tcp 10.64.2.20:8181: connect: connection refused
```
Logs from `ci-kubernetes-e2e-gce-master-new-downgrade-cluster`:
```
/etc/coredns/Corefile:4 - Error during parsing: Unknown directive 'ready'
```
In both cases, the CoreDNS deployment is updated but not the ConfigMap.
This PR aims to add the ability to migrate the CoreDNS ConfigMap in kube-up in the event of an upgrade/downgrade.

It also bumps the version of CoreDNS to 1.5.0

**Which issue(s) this PR fixes**:

Fixes #78123 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #94987, opened 2020-09-23
fix upgrade test

zshihang:
**What type of PR is this?**
/kind bug
/kind failing-test

partially address https://github.com/kubernetes/kubernetes/issues/94487

```release-note
NONE
```

### DIFF
--- a/cluster/gce/upgrade.sh
+++ b/cluster/gce/upgrade.sh
@@ -464,14 +464,20 @@ function update-coredns-config() {
   cleanup() {
     rm -rf "${download_dir}"
   }
-  trap cleanup EXIT
+  trap cleanup RETURN
 
   # Get the new installed CoreDNS version
-  echo "Waiting for CoreDNS to update"
-  until [[ $("${KUBE_ROOT}"/cluster/kubectl.sh -n kube-system get deployment coredns -o=jsonpath='{$.metadata.resourceVersion}') -ne ${COREDNS_DEPLOY_RESOURCE_VERSION} ]]; do
+  echo "== Waiting for CoreDNS to update =="
+  local -r endtime=$(date -ud "3 minute" +%s)
+  until [[ $("${KUBE_ROOT}"/cluster/kubectl.sh -n kube-system get deployment coredns -o=jsonpath='{$.metadata.resourceVersion}') -ne ${COREDNS_DEPLOY_RESOURCE_VERSION} ]] || [[ $(date -u +%s) -gt $endtime ]]; do
      sleep 1
   done
-  echo "Fetching the latest installed CoreDNS version"
+
+  if [[ $("${KUBE_ROOT}"/cluster/kubectl.sh -n kube-system get deployment coredns -o=jsonpath='{$.metadata.resourceVersion}') -ne ${COREDNS_DEPLOY_RESOURCE_VERSION} ]]; then
+    echo "== CoreDNS ResourceVersion changed =="
+  fi
+
+  echo "== Fetching the latest installed CoreDNS version =="
   NEW_COREDNS_VERSION=$("${KUBE_ROOT}"/cluster/kubectl.sh -n kube-system get deployment coredns -o=jsonpath='{$.spec.template.spec.containers[:1].image}' | cut -d ":" -f 2)
 
   case "$(uname -m)" in


liggitt:
/cc @johnbelamaric 
for ack of changes to coredns check on upgrade

liggitt:
with this fix, were you able to run the upgrade test from 1.19.x -> 1.20-alpha....? if so, a good next step would be to set up a test-infra job with that config so we can get signal on the follow-up changes around upgrade tests

zshihang:
> with this fix, were you able to run the upgrade test from 1.19...latest... -> 1.20-...latest...?

with this fix and the sha512 PR, we are able to run most master upgrade tests except the `ServiceUpgrade` which might requires some special env variables. It stuck at https://github.com/kubernetes/kubernetes/blob/0f332bae5daab8f9f0c3ecfacf7776c781b91add/test/e2e/upgrades/services.go#L109-L111.

right now, i suggest we cherrypick the upgrade test for BoundServiceAccountTokenVolume to 1.19 and then we will get signals after the config [PR](https://github.com/kubernetes/test-infra/pull/19353) is in.

cluster/gce/upgrade.sh
```diff
@@ -464,13 +464,8 @@ function update-coredns-config() {
   cleanup() {
     rm -rf "${download_dir}"
   }
-  trap cleanup EXIT
+  trap cleanup RETURN
 
-  # Get the new installed CoreDNS version
-  echo "Waiting for CoreDNS to update"
```
liggitt, line 470: this was added in https://github.com/kubernetes/kubernetes/pull/78302, want an ack from @MrHohn before removing it

MrHohn: Trying to understand - why is removing this step necessary for fixing the upgrade test?

cc @rajansandeep

rajansandeep: The check for `resourceVersion` is added for the detection of changes when CoreDNS is updating. 
But if there is no change in CoreDNS (eg: upgrading from 1.19 to 1.19), it might end up getting stuck here.

MrHohn: Thanks for the explanation. Does it mean we may get the old coredns version after removing this waiting step? Or if this is not concerning in practice then it looks fine to me.

liggitt: I can't tell if the upgrade/downgrade logic is load-bearing... are people using this for actual upgrades or is this just for tests?

if the config upgrade/downgrade is required to work, I'd expect that to be done in a more prescriptive way, like the etcd `cluster/images/etcd/migrate-if-needed.sh` approach, rather than here

MrHohn: > if the config upgrade/downgrade is required to work, I'd expect that to be done in a more prescriptive way, like the etcd cluster/images/etcd/migrate-if-needed.sh approach, rather than here

Agreed. My understanding is that this script is mostly used for tests but not in production at the moment.

MrHohn (approved):
/lgtm
/approve

cluster/gce/upgrade.sh
MrHohn: Another thought - instead of removing this step completely, could we set a max wait time here? If the version doesn't change for X minutes, we continue regardless. It may slow down the upgrade a bit but keeps the coredns version upgrade logic intact.

cc @mtaufen

mtaufen: Thanks @MrHohn. Ideally we also make it configurable. What's a sane default value?

MrHohn: Looking through the steps, the previous step `upgrade-master()` would wait for the new master to respond to API requests - then 3 minutes sounds more than enough. Alternatively we can make the step of waiting for coredns version updated configurable so tests can set it on-demand.

zshihang: the loop will timeout in 3 minutes now. PTAL

cluster/gce/upgrade.sh
```diff
@@ -464,11 +464,12 @@ function update-coredns-config() {
...
     rm -rf "${download_dir}"
   }
-  trap cleanup EXIT
+  trap cleanup RETURN
 
   # Get the new installed CoreDNS version
   echo "Waiting for CoreDNS to update"
-  until [[ $("${KUBE_ROOT}"/cluster/kubectl.sh -n kube-system get deployment coredns -o=jsonpath='{$.metadata.resourceVersion}') -ne ${COREDNS_DEPLOY_RESOURCE_VERSION} ]]; do
+  local -r endtime=$(date -ud "3 minute" +%s)
+  until [[ $("${KUBE_ROOT}"/cluster/kubectl.sh -n kube-system get deployment coredns -o=jsonpath='{$.metadata.resourceVersion}') -ne ${COREDNS_DEPLOY_RESOURCE_VERSION} ]] || [[ $(date -u +%s) -gt $endtime ]]; do
```
MrHohn, line 472: Thanks for making the change. Can we log whether we retrieved the updated version or timed out waiting for it so that it will be clear from outside?

zshihang: done

liggitt:
looks like the bot that removes the label was stuck

PR #95001, opened 2020-09-23
deprecate scheduler metrics

### LINKED ISSUE #94999: Deprecate old scheduler metrics
Mark the following metrics as deprecated `SchedulingAlgorithmPreemptionEvaluationDuration` and `BindingLatency`. Both of those metrics are now covered as part of `FrameworkExtensionPointDuration`, the former as a `PostFilter` the latter and a `Bind` plugin.

This can be done by setting `DeprecatedVersion: "1.20.0"` in each of the two metrics opts. Changing the name of the metric variable to be prefixed with "Deprecated" is useful as well.

/help
/sig scheduling

arghya88:
**What type of PR is this?**

Fixes #94999

```release-note
[ACTION REQUIRED] scheduler alpha metrics binding_duration_seconds and scheduling_algorithm_preemption_evaluation_seconds are deprecated, Both of those metrics are now covered as part of framework_extension_point_duration_seconds, the former as a PostFilter the latter and a Bind plugin. The plan is to remove both in 1.21

```

```docs

```

### DIFF
--- a/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
+++ b/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go
@@ -79,7 +79,7 @@ func (pl *DefaultPreemption) PostFilter(ctx context.Context, state *framework.Cy
 	preemptionStartTime := time.Now()
 	defer func() {
 		metrics.PreemptionAttempts.Inc()
-		metrics.SchedulingAlgorithmPreemptionEvaluationDuration.Observe(metrics.SinceInSeconds(preemptionStartTime))
+		metrics.DeprecatedSchedulingAlgorithmPreemptionEvaluationDuration.Observe(metrics.SinceInSeconds(preemptionStartTime))
 	}()
 
 	nnn, err := pl.preempt(ctx, state, pod, m)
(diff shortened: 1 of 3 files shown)


alculquicondor:
What was the number of releases before it gets removed? Worth putting that in the Release Note. Also `[ACTION REQUIRED]`

arghya88:
How about we say these metrics will be removed in the next release

ahg-g:
please update the release note as follows:

"[ACTION REQUIRED] scheduler alpha metrics binding_duration_seconds  and scheduling_algorithm_preemption_evaluation_seconds are deprecated, Both of those metrics are now covered as part of framework_extension_point_duration_seconds, the former as a PostFilter the latter and a Bind plugin. The plan is to remove both in 1.21"

> What was the number of releases before it gets removed?
Those metrics can be removed immediately since they are alpha, but we will remove them in the next cycle.

arghya88:
@ahg-g Updated release note

alculquicondor:
> Those metrics can be removed immediately since they are alpha, but we will remove them in the next cycle.

I remember that adding `DeprecatedVersion` caused other metrics to stop being exported at some point, even if we still put data from the scheduler code.

ahg-g:
> I remember that adding DeprecatedVersion caused other metrics to stop being exported at some point, even if we still put data from the scheduler code.

Right, the metric is alpha so it will stop exporting them directly (not sure if it will continue to export them if the metric is Beta), but admins can still pass a flag to the component to enable them.

ahg-g:
/lgtm 

will leave it to Aldo to approve

ahg-g:
@alculquicondor

PR #95008, opened 2020-09-23
Reverting antonios revert (and re-adding the original fix to breadth first polling which was lost in the original pr)

### LINKED ISSUE #93833: Breadth first polling in intra-pod networking Conformance verification

jayunit100:
Add back in the original test failures, and add in a revert that captures a hole which @aojea found in my original test refactoring logic.

**What type of PR is this?**

/kind bug
/kind regression

A few wks ago when refactoring side-effects out of the tests, introduced a loss of test coverage (many tests relied on side-effects of the dial functionality to fail)

This test reverts the original commit (to capture the logical history of this bug)  , and re-integrates it as an atomic commit with the fix which enables the original test coverage. 
 
Fixes #93833, again :)

```release-note
NONE
```

### DIFF
--- a/test/e2e/common/networking.go
+++ b/test/e2e/common/networking.go
@@ -18,7 +18,6 @@ package common
 
 import (
 	"github.com/onsi/ginkgo"
-	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2enetwork "k8s.io/kubernetes/test/e2e/framework/network"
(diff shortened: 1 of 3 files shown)


jayunit100:
@aojea I think this covers the issue you were mentioning

aojea:
/retest
lgtm, let's wait for the CI signal

jayunit100:
@SataQiu can you give this a quick look  ?  it is a complex but simple PR
1. per Antonio, explicit commit reverting the original pr
2. new commit adding it back in
3. final commit that adds back in the tests which were removed from

spiffxp:
Looks like https://github.com/kubernetes/kubernetes/pull/94982 was the original PR for the revert, just to make the trail of breadcrumbs easier to follow

spiffxp (approved):
/approve
/milestone v1.20

PR #95012, opened 2020-09-23
Make the creation of namespace using POST and PATCH consistent

### LINKED ISSUE #94168: patch-based (server-side-apply) and post-based Namespace creation requests to different endpoints populate different namespace attribute in AdmissionReview

nodo:
**What type of PR is this?**

/kind cleanup

PATCH verb is used when creating a namespace using server-side apply,
while POST verb is used when creating a namespace using client-side
apply.

The difference in path between the two ways to create a namespace led to
an inconsistency when calling webhooks. When server-side apply is used,
the request sent to webhooks has the field "namespace" populated with
the name of namespace being created. On the other hand, when using
client-side apply the "namespace" field is omitted.

This commit aims to make the behaviour consistent and populates the
"namespace" field when creating a namespace using POST verb (i.e.
client-side apply).

Fixes #94168

I have checked all the admission plugins as mentioned in https://github.com/kubernetes/kubernetes/issues/94168#issuecomment-690357036 and found that only one needed to be changed. Would be great to have another pair of eyes to check that though.

```release-note
AdmissionReview objects sent for the creation of Namespace API objects now populate the `namespace` attribute consistently (previously the `namespace` attribute was empty for Namespace creation via POST requests, and populated for Namespace creation via server-side-apply PATCH requests)
```


staging/src/k8s.io/apiserver/pkg/admission/plugin/resourcequota/admission.go
```diff
@@ -136,9 +137,13 @@ func (a *QuotaAdmission) Validate(ctx context.Context, attr admission.Attributes
...
-	// ignore all operations that are not namespaced
-	if attr.GetNamespace() == "" {
+	// ignore all operations that are not namespaced or creation of namespaces
+	if attr.GetNamespace() == "" || isNamespaceCreation(attr) {
 		return nil
 	}
 	return a.evaluator.Evaluate(attr)
 }
+
+func isNamespaceCreation(attr admission.Attributes) bool {
```
liggitt, line 147: nits:
1. put the cheap check (attr.GetOperation() == admission.Create) first
2. construct the comparison value once as a package var `var namespaceGVK = v1.SchemeGroupVersion.WithKind("Namespace").GroupKind()`

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/create.go
```diff
@@ -128,17 +127,22 @@ func createHandler(r rest.NamedCreater, scope *RequestScope, admit admission.Int
 		}
 		trace.Step("Conversion done")
 
+		// On create, get name from new object if unset
+		if len(name) == 0 {
+			_, name, _ = scope.Namer.ObjectName(obj)
+
+			if gvk.Kind == "Namespace" {
+				namespace = name
+			}
```
liggitt, line 136: 1. compare the full group/kind
2. shouldn't this be outside the `if len(name) == 0` block?

staging/src/k8s.io/apiserver/pkg/admission/plugin/resourcequota/admission.go
nodo: Thanks, will change both.

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/create.go
nodo: Thanks for spotting it, you are right, I will change both points.

liggitt:
cc @lavalamp @deads2k for visibility

I think populating the `namespace` admission attribute for creation of `Namespace` objects makes sense (it is already populated for update/patch/delete of Namespace objects, and for create of `Namespace` objects via apply), but wanted additional eyes on this

lavalamp:
Sounds good to me.

/ok-to-test

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/create.go
```diff
@@ -128,17 +129,21 @@ func createHandler(r rest.NamedCreater, scope *RequestScope, admit admission.Int
 		}
 		trace.Step("Conversion done")
 
+		// On create, get name from new object if unset
+		if len(name) == 0 {
+			_, name, _ = scope.Namer.ObjectName(obj)
+		}
+		if *gvk == namespaceGVK {
```
liggitt, line 136:
```suggestion
		if len(namespace) == 0 && *gvk == namespaceGVK {
```

liggitt (commented):
one nit, then lgtm

staging/src/k8s.io/apiserver/pkg/endpoints/handlers/create.go
nodo: Sounds good, I have committed the change 👍

test/integration/apiserver/admissionwebhook/admission_test.go
```diff
@@ -995,6 +998,20 @@ func testNamespaceDelete(c *testContext) {
 	}
 }
 
+func testNamespaceCreate(c *testContext) {
```
liggitt, line 1001: rather than adding a custom `testNamespaceCreate`, suggest broadening the existing special-case for namespaces in `expect`:

```diff
 func (h *holder) expect(gvr schema.GroupVersionResource, gvk, optionsGVK schema.GroupVersionKind, operation v1beta1.Operation, name, namespace string, object, oldObject, options bool) {
        // Special-case namespaces, since the object name shows up in request attributes for update/delete requests
-       if len(namespace) == 0 && gvk.Group == "" && gvk.Version == "v1" && gvk.Kind == "Namespace" && operation != v1beta1.Create {
+       if len(namespace) == 0 && gvk.Group == "" && gvk.Version == "v1" && gvk.Kind == "Namespace" {
                namespace = name
        }
```

nodo: Changed [here](https://github.com/kubernetes/kubernetes/commit/3cb510e33eecbdc37aad14f121396ccfbf5268cb#diff-706a54bca254b383e7104ed375dd6f54).

tnozicka:
> I think populating the namespace admission attribute for creation of Namespace objects makes sense (it is already populated for update/patch/delete of Namespace objects, and for create of Namespace objects via apply), but wanted additional eyes on this

Namespace is a cluster level resource and as such it's namespace field should be empty.

Why do you want to put a name of the instance there? This has broken our downstream admission that were ignoring cluster scoped resources (namespace=="") and let a namespace in, which created cyclic dependency on trying to list quota in a namespace that was in process of creation and couldn't be listed.

liggitt:
Namespaces are the edge-case between cluster-scoped and namespaces objects, so we had to decide whether the namespace's own name should be populated in admission. 

This is consistent with how authorization already worked for get/update operations on namespaces (get or update permission on a specific namespace can be granted via a rolebinding inside that namespace) and with other admission operations on the namespace:

> it is already populated for update/patch/delete of Namespace objects, and for create of Namespace objects via apply

liggitt:
> This has broken our downstream admission that was ignoring/skipping cluster scoped resources (namespace=="") and let a namespace in, which created cyclic dependency on trying to list quota in a namespace that was in process of creation and couldn't be listed.

cc @deads2k for visibility

tnozicka:
I see that this is already being set on other operations but I do not yet understand why. At least in my mind, a namespace can't be contained in itself. Namespaces live at cluster scope, same with etcd path. Namespace `foo` claiming it lives in namespace `foo` just creates cycle in my mind, or etcd storage and breaks a common distinction between namespaced and non-namespaced objects. When you ask the API to give you a namespace `foo` the response says `metadata.namespace: ""`. Maybe I am just not seeing the whole picture for admission and there is a good reason to act contrary to these assumptions. I may be biased because I've spent last 2 days chasing this down :)

tnozicka:
But `Attributes` are not the object. I don't want to be defensive here if the behaviour is really intentional, so let's try to steer the discussion: What is the recommended way to check if an object connected to `Attributes` is namespaced?
(Special casing Namespaces sounds a bit cumbersome and error prone.)

liggitt:
metadata.namespace on the object == "what namespace is this object contained within". It is only set for namespaced resources and is not set on Namespace objects themselves. 

namespace admission attribute == "what namespace is this request scoped to?". It is set from request attributes parsed from the URL, so it populated the namespace for update/delete/patch/create-via-apply requests (because those requests include "…/namespaces/$name" in the URL). This PR made the admission attribute consistent for the create-via-post request as well via a special case.

liggitt:
Until https://github.com/kubernetes/kubernetes/issues/88282 is improved, admission webhooks handling namespaced resources specially will likely need to special-case namespaces (or register their web hook to only receive namespaced objects with a `scope:       "Namespaced"` rule)

tnozicka:
> namespace admission attribute == "what namespace is this request scoped to?". It is set from request attributes parsed from the URL, so it populated the namespace for update/delete/patch/create-via-apply requests (because those requests include "…/namespaces/$name" in the URL). This PR made the admission attribute consistent for the create-via-post request as well via a special case.

Makes sense, although the special case not matching the URL is not great (although understandable), I think this isn't the issue. Our plugin should have probably looked directly at the object inside the attributes and not on the request itself.
(We aren't using admission webhooks, this is just hardwired admission to KAS.)

Thanks @liggitt and enjoy the holidays!

PR #95019, opened 2020-09-23
Fix e2e endpoint slice flake test

### LINKED ISSUE #94998: [Flaky Test][sig-network] EndpointSlice should create Endpoints and EndpointSlices for Pods matching a Service
**Which jobs are flaking**:
ci-kubernetes-e2e-gci-gce #1308692423005704192

[sig-network] EndpointSlice should create Endpoints and EndpointSlices for Pods matching a Service

**Testgrid link**:
https://testgrid.k8s.io/google-gce#gce-cos-master-default

**Reason for failure**:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce/1308692423005704192

```
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/network/endpointslice.go:167
Sep 23 09:23:35.932: Timed out waiting for EndpointSlices to match expectations: timed out waiting for the condition
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/network/endpointslice.go:324
```
https://storage.googleapis.com/k8s-gubernator/triage/index.html?test=EndpointSlice%20should%20create%20Endpoints%20

Flaking across jobs:

```


Time | Job | Test
-- | -- | --
21/09/2020, 09:36:22 | ci-kubernetes-e2e-gci-gce-kube-dns | [sig-network] EndpointSlice should create Endpoints and EndpointSlices for Pods matching a Service
21/09/2020, 09:16:32 | ci-kubernetes-e2e-gce-cos-k8sbeta-default | [sig-network] EndpointSlice should create Endpoints and EndpointSlices for Pods matching a Service
20/09/2020, 05:30:22 | ci-kubernetes-e2e-gci-gce-coredns-nodecache | [sig-network] EndpointSlice should create Endpoints and EndpointSlices for Pods matching a Service
20/09/2020, 02:12:30 | ci-cos-containerd-e2e-gci-gce | [sig-network] EndpointSlice should create Endpoints and EndpointSlices for Pods matching a Service
19/09/2020, 20:23:24 | ci-kubernetes-e2e-gci-gce-coredns | [sig-network] EndpointSlice should create Endpoints and EndpointSlices for Pods matching a Service



```

/sig network

aojea:
In some cases the EndpointSlice controller will create more
    EndpointSlices than necessary resulting in some duplication. This is
    valid and tests should only fail here if less EndpointSlices than
    expected are added.

/kind flake

```release-note
NONE
```

Fixes: https://github.com/kubernetes/kubernetes/issues/94998

### DIFF
--- a/test/e2e/network/endpointslice.go
+++ b/test/e2e/network/endpointslice.go
@@ -521,7 +521,11 @@ func hasMatchingEndpointSlices(cs clientset.Interface, ns, svcName string, numEn
 	for _, endpointSlice := range esList.Items {
 		actualNumEndpoints += len(endpointSlice.Endpoints)
 	}
-	if actualNumEndpoints != numEndpoints {
+	// In some cases the EndpointSlice controller will create more
+	// EndpointSlices than necessary resulting in some duplication. This is
+	// valid and tests should only fail here if less EndpointSlices than
+	// expected are added.
+	if actualNumEndpoints < numEndpoints {
 		framework.Logf("EndpointSlices for %s/%s Service have %d/%d endpoints", ns, svcName, actualNumEndpoints, numEndpoints)
 		return esList.Items, false
 	}


test/e2e/network/endpointslice.go
```diff
@@ -521,7 +521,11 @@ func hasMatchingEndpointSlices(cs clientset.Interface, ns, svcName string, numEn
 	for _, endpointSlice := range esList.Items {
 		actualNumEndpoints += len(endpointSlice.Endpoints)
 	}
-	if actualNumEndpoints != numEndpoints {
+	// In some cases the EndpointSlice controller will create more
+	// EndpointSlices than necessary resulting in some duplication. This is
+	// valid and tests should only fail here if less EndpointSlices than
+	// expected are added.
+	if actualNumEndpoints < numEndpoints {
```
robscott, line 528: Sadly I think we need something a bit more advanced here. As an example, we do need to ensure that we're not just creating endpoints for every pod in the cluster. Duplicate endpoints are fine, but endpoints referring to Pods that are not in this Service are not, I don't think we have sufficient coverage for that yet. I think we previously got around this with a check to ensure that all Pods we expected were referenced with Endpoints. That combined with this check to ensure there were no excess endpoints was sufficient, now it's not.

aojea: I think wojtek added some helper recently along this lines, will check tomorrow,

aojea: is not all deduplicationg and matchin addresses done in ?
https://github.com/kubernetes/kubernetes/blob/beef6ae51245515ed9c68090b9288d986a882c7b/test/e2e/network/endpointslice.go#L314-L317

IIUIC that functions calls this `hasMatchingEndpointSlices()` just to return the slices, and then it seems to do all dedup that you mention

robscott: Sorry for the delay on this! I think there's one more place that needs to change:

https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/endpointslice.go#L548-L555

Currently `numIPs` is based on `len(pods)`. If we changed that logic to instead count the number of unique IPs with something like a set I think that would fix the flakiness here.

aojea: but these are Endpoints, not Slices, len(pods) has to match the number of addresses, the Endpoints doesn't have duplicates, right?

robscott: 🤦 you're completely right. Maybe what we'd run into there would be one older EndpointSlice with fewer endpoints than the newer one. Even still, this should fix the majority of flaky tests.

robscott:
Thanks!

/hold cancel
/lgtm

aojea:
/assign @thockin 
for approval

freehan (commented):
/lgtm
/approve

PR #94875, opened 2020-09-18
Bump QEMU version to 5.1.0-2 to fix agnhost image on s390x

rajaskakodkar:
Signed-off-by: Rajas Kakodkar <rajas.kakodkar@ibm.com>

**What type of PR is this?**
/kind/bug

**What this PR does / why we need it**:
This PR updates QEMU version from 2.9.1 to 5.1.0-2 to resolve the following issue faced with agnhost image on s390x architecture.
```
Step 4/14 : RUN apk --update add bind-tools curl netcat-openbsd iproute2 iperf bash && rm -rf /var/cache/apk/*   && ln -s /usr/bin/iperf /usr/local/bin/iperf   && ls -altrh /usr/local/bin/iperf
 ---> Running in 87a9e760052b
ERROR: Unable to open root: Bad address
ERROR: Failed to open apk database: Bad address
?rwxrwxrwx    1 root     root          14 Sep 10 10:26 /usr/local/bin/iperf -> /usr/bin/iperf
Removing intermediate container 87a9e760052b
```
**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
None
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #94777, opened 2020-09-14
test images: Fixes echoserver s390x image

claudiubelu:
**What type of PR is this?**

/kind bug
/sig testing
/area conformance

**What this PR does / why we need it**:

When trying to build the s390x image, it would fail when running the apk command with the following error:

    ERROR: Unable to open root: Bad address
    ERROR: Failed to open apk database: Bad address

This can be fixed by updating the ``third_party/multiarch/qemu-user-static/register/register.sh`` and ``third_party/multiarch/qemu-user-static/register/qemu-binfmt-conf.sh`` scripts and their usage to a newer version [1].

Additionally, the packages ``nginx-mod-http-lua`` and ``nginx-mod-http-lua-upstream`` cannot be found in the regular ``http://dl-cdn.alpinelinux.org/alpine/v3.9/main/s390x/`` repository, but we can use an older one [2].

[1] https://github.com/qemu/qemu/blob/master/scripts/qemu-binfmt-conf.sh
[2] http://dl-cdn.alpinelinux.org/alpine/v3.8/main

**Which issue(s) this PR fixes**:

Fixes #84158

**Special notes for your reviewer**:

Image building logs with the current changes: https://paste.ubuntu.com/p/zrwVfbbS75/

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #95032, opened 2020-09-24
test images: Removes -p yes flag from qemu-user-static script call

claudiubelu:
**What type of PR is this?**

/kind bug

/sig testing
/priority important-soon

Currently, the Image Builder job is failing as it cannot build images for other architecture types [1]. This happens because the Image Builder image (``gcr.io/k8s-testimages/gcb-docker-gcloud:v20190906-745fed4``) does not have any of the expected ``qemu-*`` binaries in ``/usr/bin/``, so the command ``${sudo} "${KUBE_ROOT}/third_party/multiarch/qemu-user-static/register/register.sh --reset -p yes"`` is not working properly. This can be reproduced locally:

    docker run -ti --rm --privileged -v /var/run/docker.sock:/var/run/docker.sock -v /home/atuvenie/kubernetes:/workspace/kubernetes -v /home/atuvenie/.docker:/root/.docker gcr.io/k8s-testimages/gcb-docker-gcloud:v20190906-745fed4 -c "cd kubernetes && make all-container WHAT=agnhost``

The resulting logs are: [2]. Because the ``qemu-*`` binaries are missing, errors like this can be observed in the logs:

    Registering qemu-*-static binaries in the kernel
    Setting /usr/bin/qemu-alpha-static as binfmt interpreter for alpha
    sh: write error: No such file or directory
    Setting /usr/bin/qemu-arm-static as binfmt interpreter for arm
    sh: write error: No such file or directory
    Setting /usr/bin/qemu-armeb-static as binfmt interpreter for armeb
    sh: write error: No such file or directory
    Setting /usr/bin/qemu-sparc-static as binfmt interpreter for sparc
    sh: write error: No such file or directory
    Setting /usr/bin/qemu-sparc32plus-static as binfmt interpreter for sparc32plus
    sh: write error: No such file or directory

The command ``docker run --rm --privileged multiarch/qemu-user-static --reset -p yes`` can be used instead for the same effect. This will result in images being buildable inside a ``gcr.io/k8s-testimages/gcb-docker-gcloud:v20190906-745fed4`` container: [3]

Note that this wasn't an issue for hosts that already have qemu installed, as the image ``multiarch/qemu-user-static`` contains the ``qemu-*`` binaries mentioned before.

The images can also be built inside the container if the ``qemu-*`` binaries are included in it: [4] (``find /usr/bin -type f -name 'qemu-*' -exec sh -c 'docker cp {} containerid:/usr/bin/' \;``)

Alternatively, we can remove the ``-p yes`` flag.

The QEMUVERSION in ``test/images/Makefile`` is also bumped to 5.1.0-2, which is the same version as the ``third_party/multiarch/qemu-user-static/register`` scripts.

[1] http://storage.googleapis.com/k8s-staging-e2e-test-images-gcb/logs/log-77718177-fddb-4a8e-980b-0a38e76f4907.txt
[2] https://paste.ubuntu.com/p/Ncx2jrb8MH/
[3] https://paste.ubuntu.com/p/xBN5fXSbZX/
[4] https://paste.ubuntu.com/p/SX7Zmd75bY/

Fixes #

This issue started appearing since https://github.com/kubernetes/kubernetes/pull/94777 merged.

Included the commit from: https://github.com/kubernetes/kubernetes/pull/94875

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/images/image-util.sh
+++ b/test/images/image-util.sh
@@ -149,7 +149,7 @@ build() {
         if [[ $(id -u) != 0 ]]; then
           sudo=sudo
         fi
-        ${sudo} "${KUBE_ROOT}/third_party/multiarch/qemu-user-static/register/register.sh" --reset -p yes
+        ${sudo} "${KUBE_ROOT}/third_party/multiarch/qemu-user-static/register/register.sh" --reset
         curl -sSL https://github.com/multiarch/qemu-user-static/releases/download/"${QEMUVERSION}"/x86_64_qemu-"${QEMUARCHS[$arch]}"-static.tar.gz | tar -xz -C "${temp_dir}"
         # Ensure we don't get surprised by umask settings
         chmod 0755 "${temp_dir}/qemu-${QEMUARCHS[$arch]}-static"
--- a/test/images/Makefile
+++ b/test/images/Makefile
@@ -15,7 +15,7 @@
 REGISTRY ?= gcr.io/kubernetes-e2e-test-images
 GOARM ?= 7
 DOCKER_CERT_BASE_PATH ?=
-QEMUVERSION=v2.9.1
+QEMUVERSION=v5.1.0-2
 GOLANG_VERSION=1.15.2
 export
 
--- a/third_party/multiarch/qemu-user-static/README.kubernetes
+++ b/third_party/multiarch/qemu-user-static/README.kubernetes
@@ -1,1 +1,2 @@
-Files copied from https://github.com/multiarch/qemu-user-static at commit 22b0013668d2aed4a2cfd21650e85c664b1f21c6.
+qemu-binfmt-confg.sh comes from https://github.com/qemu/qemu/blob/7c81570d932268a9626457a662f1c5046ebc455e/scripts/qemu-binfmt-conf.sh
+register.sh comes from https://github.com/multiarch/qemu-user-static/blob/9898439ca975c9c4c0db2bdc4b3707d382559d09/containers/latest/register.sh


dims:
@claudiubelu do we want to remove third_party/multiarch/qemu-user-static/register/register.sh ?

claudiubelu:
> 
> 
> @claudiubelu do we want to remove third_party/multiarch/qemu-user-static/register/register.sh ?

After a local grep, it seems like ``./cluster/addons/addon-manager/Makefile`` still uses that script. I'm wondering if there's a job building that image, and if it has the same issue as we are.

I'm thinking of pinning version for ``multiarch/qemu-user-static``, so we don't start having issues if the latest images gets changed to something else that might generate other issues.

claudiubelu:
Building a test image inside a ``gcr.io/k8s-testimages/gcb-docker-gcloud:v20200422-b25d964`` container with this change: https://paste.ubuntu.com/p/yF4Pw3bpPG/

spiffxp:
/cc @ixdy 

I'm not sure I have enough historical context to know if we care about continuing to use the third_party version

ixdy:
I probably won't have time to review this before Thursday, but the main reason we switched from running a docker image to a script in `third_party` is that we felt that running an unaudited privileged docker container every time any developer ran `make` was a notable security concern.

test/images/image-util.sh
```diff
@@ -136,11 +136,7 @@ build() {
         ${SED} -i "s|QEMUARCH|${QEMUARCHS[$arch]}|g" Dockerfile
         # Register qemu-*-static for all supported processors except the current one
         echo "Registering qemu-*-static binaries in the kernel"
-        local sudo=""
-        if [[ $(id -u) != 0 ]]; then
-          sudo=sudo
-        fi
-        ${sudo} "${KUBE_ROOT}/third_party/multiarch/qemu-user-static/register/register.sh" --reset -p yes
+        docker run --rm --privileged multiarch/qemu-user-static:5.1.0-2 --reset -p yes
```
BenTheElder, line 139: hmm why are we both curling from their github and pulling from dockerhub?

it makes me very nervous to not pin these.

test/images/image-util.sh
```diff
@@ -136,11 +136,7 @@ build() {
...
         curl -sSL https://github.com/multiarch/qemu-user-static/releases/download/"${QEMUVERSION}"/x86_64_qemu-"${QEMUARCHS[$arch]}"-static.tar.gz | tar -xz -C "${temp_dir}"
```
BenTheElder, line 140: I didn't realize this wasn't checking a hash at least. yikes

test/images/image-util.sh
claudiubelu: Good point. I've changed this back and simply removed the ``-p yes``. That seems to be working too.

So, it seems that it is like this: that ``-p yes`` is the persistent flag. Only checked this briefly, but if we are using the ``-p yes`` flag then we won't need the curl below or include ``/usr/bin/qemu-$arch-static`` in the images. But that ``-p yes`` is causing the errors:

```Registering qemu-*-static binaries in the kernel
Setting /usr/bin/qemu-alpha-static as binfmt interpreter for alpha
sh: write error: No such file or directory
Setting /usr/bin/qemu-arm-static as binfmt interpreter for arm
sh: write error: No such file or directory
```

errors we're seeing inside the ``gcr.io/k8s-testimages/gcb-docker-gcloud:v20200422-b25d964`` container. In order to use the ``-p yes`` flag, we'd have to use:

```
docker run --rm --privileged multiarch/qemu-user-static:5.1.0-2 --reset -p yes
```

instead of:

```
${sudo} "${KUBE_ROOT}/third_party/multiarch/qemu-user-static/register/register.sh" --reset
```

test/images/image-util.sh
claudiubelu: at least it's versioned, it's using ``QEMUVERSION`` release.

spiffxp:
>The QEMUVERSION in test/images/Makefile is also bumped to 5.1.0-2, which is the same version as the third_party/multiarch/qemu-user-static/register scripts

We should have updated third_party/multiarch/qemu-user-static/README.kubernetes in https://github.com/kubernetes/kubernetes/pull/94777, since it still says

> Files copied from https://github.com/multiarch/qemu-user-static at commit 22b0013668d2aed4a2cfd21650e85c664b1f21c6.

which corresponds to https://github.com/multiarch/qemu-user-static/commit/22b0013668d2aed4a2cfd21650e85c664b1f21c6 which is 3.0.0

Can you update it in this PR?

claudiubelu:
> 
> 
> > The QEMUVERSION in test/images/Makefile is also bumped to 5.1.0-2, which is the same version as the third_party/multiarch/qemu-user-static/register scripts
> 
> We should have updated third_party/multiarch/qemu-user-static/README.kubernetes in #94777, since it still says
> 
> > Files copied from https://github.com/multiarch/qemu-user-static at commit 22b0013668d2aed4a2cfd21650e85c664b1f21c6.
> 
> which corresponds to [multiarch/qemu-user-static@22b0013](https://github.com/multiarch/qemu-user-static/commit/22b0013668d2aed4a2cfd21650e85c664b1f21c6) which is 3.0.0
> 
> Can you update it in this PR?

Good point. But I'll update it to:

```
Files copied from the Docker image multiarch/qemu-user-static:5.1.0-2.
```

And that's because the ``multiarch/qemu-user-static`` image's Dockerfile has this line [1]:

```
ADD https://raw.githubusercontent.com/qemu/qemu/master/scripts/qemu-binfmt-conf.sh /qemu-binfmt-conf.sh
```

Since the script is copied from master, I think it's more correct to get the script from the tagged image instead.

[1] https://github.com/multiarch/qemu-user-static/blob/v5.1.0-2/containers/latest/Dockerfile#L4

spiffxp:
> Files copied from the Docker image multiarch/qemu-user-static:5.1.0-2.

That tag is still potentially mutable without a digest to go along with it.  I would rather track it down to the git SHA, it's probably near the tag in git.

spiffxp:
qemu-binfmt-confg.sh comes from https://github.com/qemu/qemu/blob/7c81570d932268a9626457a662f1c5046ebc455e/scripts/qemu-binfmt-conf.sh

register.sh comes from https://github.com/multiarch/qemu-user-static/blob/9898439ca975c9c4c0db2bdc4b3707d382559d09/containers/latest/register.sh

claudiubelu:
> 
> 
> qemu-binfmt-confg.sh comes from https://github.com/qemu/qemu/blob/7c81570d932268a9626457a662f1c5046ebc455e/scripts/qemu-binfmt-conf.sh
> 
> register.sh comes from https://github.com/multiarch/qemu-user-static/blob/9898439ca975c9c4c0db2bdc4b3707d382559d09/containers/latest/register.sh

Done

spiffxp (approved):
/approve
/lgtm

spiffxp:
/assign @dims 
for the /third_party approval

PR #94902, opened 2020-09-18
proxy: Restructure config validation tests to check errors

cmluciano:
Signed-off-by: Christopher M. Luciano <cmluciano@us.ibm.com>

**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
The tests for most functions have also been revised to check the errors
explicitly upon validating. This will properly catch occasions
where we should be returning multiple errors if more error occurs or if just one block is failing.

**Which issue(s) this PR fixes**:
xref: #94732

**Special notes for your reviewer**:
Accompanying PR for validateProxyNodePortAddress and validateExcludeCIDRS #95036

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

Issue #94732, opened 2020-09-11
kube proxy configuration api tests are not validating errors correctly


cmluciano:
**What happened**:
The [kube-proxy config api validation tests](https://github.com/kubernetes/kubernetes/blame/master/pkg/proxy/apis/config/validation/validation_test.go) only appear to check for the existence of an error with respect to a small message. This means that the msg field can be blank and the test will still pass. It will only fail if the error is something arbitrary.

**What you expected to happen**:
Validation error tests should match the actual errors returned from the API or we should just be checking for if errors are expected.

**How to reproduce it (as minimally and precisely as possible)**:
```
cmluciano@starlord:~/go/src/k8s.io/kubernetes$ git d
diff --git a/pkg/proxy/apis/config/validation/validation_test.go b/pkg/proxy/apis/config/validation/validation_test.go
index 35bf9d77e0b..6db346d4f0a 100644
--- a/pkg/proxy/apis/config/validation/validation_test.go
+++ b/pkg/proxy/apis/config/validation/validation_test.go
@@ -447,7 +447,7 @@ func TestValidateKubeProxyConfiguration(t *testing.T) {
                                        TCPCloseWaitTimeout:   &metav1.Duration{Duration: 5 * time.Second},
                                },
                        },
-                       msg: "must be greater than 0",
+                       msg: "",
                },
        }
 
cmluciano@starlord:~/go/src/k8s.io/kubernetes$ go test ./pkg/proxy/apis/config/validation/...
ok  	k8s.io/kubernetes/pkg/proxy/apis/config/validation	0.021s


cmluciano@starlord:~/go/src/k8s.io/kubernetes$ git d
diff --git a/pkg/proxy/apis/config/validation/validation_test.go b/pkg/proxy/apis/config/validation/validation_test.go
index 35bf9d77e0b..92e10ee927a 100644
--- a/pkg/proxy/apis/config/validation/validation_test.go
+++ b/pkg/proxy/apis/config/validation/validation_test.go
@@ -447,7 +447,7 @@ func TestValidateKubeProxyConfiguration(t *testing.T) {
                                        TCPCloseWaitTimeout:   &metav1.Duration{Duration: 5 * time.Second},
                                },
                        },
-                       msg: "must be greater than 0",
+                       msg: "bleh",
                },
        }
 
cmluciano@starlord:~/go/src/k8s.io/kubernetes$ go test ./pkg/proxy/apis/config/validation/...
[...]

PR #95036, opened 2020-09-24
proxy: validate each CIDR config seperately and check for errors

cmluciano:
Signed-off-by: Christopher M. Luciano <cmluciano@us.ibm.com>

**What type of PR is this?**
/kind cleanup

This commit revises validateProxyNodePortAddress and
validateExcludeCIDRS to report on the exact CIDR that is
invalid within the array of strings. Previously we would just return
the whole block of addresses and now we identify the exact address
within the block to eliminate confusion. I also removed the break from
validateProxyNodeAddress so that we can report on all addresses that
may not be valid.

The tests for each function have also been revised to check the errors
explicitly upon validating. This also will properly catch occasions
where we should be returning multiple errors if more than one CIDR is invalid.

xref: #94732

I split this out from my WIP PR #94902 since it involved changing the validation.go 

-->
```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/proxy/apis/config/validation/validation.go
+++ b/pkg/proxy/apis/config/validation/validation.go
@@ -276,8 +276,7 @@ func validateKubeProxyNodePortAddress(nodePortAddresses []string, fldPath *field
 
 	for i := range nodePortAddresses {
 		if _, _, err := net.ParseCIDR(nodePortAddresses[i]); err != nil {
-			allErrs = append(allErrs, field.Invalid(fldPath, nodePortAddresses, "must be a valid IP block"))
-			break
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(i), nodePortAddresses[i], "must be a valid CIDR"))
 		}
 	}
 
@@ -307,7 +306,7 @@ func validateIPVSExcludeCIDRs(excludeCIDRs []string, fldPath *field.Path) field.
 
 	for i := range excludeCIDRs {
 		if _, _, err := net.ParseCIDR(excludeCIDRs[i]); err != nil {
-			allErrs = append(allErrs, field.Invalid(fldPath, excludeCIDRs, "must be a valid IP block"))
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(i), excludeCIDRs[i], "must be a valid CIDR"))
 		}
 	}
 	return allErrs
--- a/pkg/proxy/apis/config/validation/validation_test.go
+++ b/pkg/proxy/apis/config/validation/validation_test.go
@@ -924,52 +924,51 @@ func TestValidateKubeProxyNodePortAddress(t *testing.T) {
 		}
 	}
 
-	errorCases := []struct {
-		addresses []string
-		msg       string
+	testCases := map[string]struct {
+		addresses    []string
+		expectedErrs field.ErrorList
 	}{
-		{
-			addresses: []string{"foo"},
-			msg:       "must be a valid IP block",
+		"invalid foo address": {
+			addresses:    []string{"foo"},
+			expectedErrs: field.ErrorList{field.Invalid(newPath.Child("NodePortAddresses[0]"), "foo", "must be a valid CIDR")},
 		},
-		{
-			addresses: []string{"1.2.3"},
-			msg:       "must be a valid IP block",
+		"invalid octet address": {
+			addresses:    []string{"10.0.0.0/0", "1.2.3"},
+			expectedErrs: field.ErrorList{field.Invalid(newPath.Child("NodePortAddresses[1]"), "1.2.3", "must be a valid CIDR")},
 		},
-		{
-			addresses: []string{""},
-			msg:       "must be a valid IP block",
+		"address cannot be 0": {
+			addresses:    []string{"127.0.0.1/32", "0", "1.2.3.0/24"},
+			expectedErrs: field.ErrorList{field.Invalid(newPath.Child("NodePortAddresses[1]"), "0", "must be a valid CIDR")},
 		},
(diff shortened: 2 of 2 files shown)


pkg/proxy/apis/config/validation/validation_test.go
```diff
@@ -924,52 +924,51 @@ func TestValidateKubeProxyNodePortAddress(t *testing.T) {
...
-		{
-			addresses: []string{"::1"},
-			msg:       "must be a valid IP block",
+		"missing ipv6 subnet ranges": {
+			addresses: []string{"::0", "::1", "2001:db8::/32"},
+			expectedErrs: field.ErrorList{field.Invalid(newPath.Child("NodePortAddresses"), "::0", "must be a valid CIDR"),
+				field.Invalid(newPath.Child("NodePortAddresses"), "::1", "must be a valid CIDR")},
 		},
-		{
-			addresses: []string{"2001:db8:1"},
```
aojea, line 952: why removing this testcase?

pkg/proxy/apis/config/validation/validation_test.go
```diff
@@ -996,45 +995,46 @@ func TestValidateKubeProxyExcludeCIDRs(t *testing.T) {
...
-		{
-			addresses: []string{"::1"},
-			msg:       "must be a valid IP block",
+		"missing ipv6 subnet ranges": {
+			addresses: []string{"::0", "::1", "2001:db8::/32"},
+			expectedErrs: field.ErrorList{field.Invalid(newPath.Child("ExcludeCIDRS"), "::0", "must be a valid CIDR"),
+				field.Invalid(newPath.Child("ExcludeCIDRS"), "::1", "must be a valid CIDR")},
 		},
-		{
-			addresses: []string{"2001:db8:1"},
```
aojea, line 1024: why removing this?

pkg/proxy/apis/config/validation/validation_test.go
cmluciano: This is missing the subnet range at the end right? If so, it should be invalid similar to the test above and I cover that one in https://github.com/kubernetes/kubernetes/pull/95036/files#diff-2b4676e3886ea440d4f0360b58d391ddR947-R951 .

pkg/proxy/apis/config/validation/validation_test.go
cmluciano: This is missing the subnet range at the end right? If so, it should be invalid similar to the test above and I cover that one in https://github.com/kubernetes/kubernetes/pull/95036/files#diff-2b4676e3886ea440d4f0360b58d391ddR947-R951

cmluciano (commented):
feedback comments

pkg/proxy/apis/config/validation/validation_test.go
aojea: yep, didn't realize sorry

pkg/proxy/apis/config/validation/validation_test.go
aojea: yep, didn't realize sorry

pkg/proxy/apis/config/validation/validation.go
```diff
@@ -276,8 +276,7 @@ func validateKubeProxyNodePortAddress(nodePortAddresses []string, fldPath *field
 
 	for i := range nodePortAddresses {
 		if _, _, err := net.ParseCIDR(nodePortAddresses[i]); err != nil {
-			allErrs = append(allErrs, field.Invalid(fldPath, nodePortAddresses, "must be a valid IP block"))
-			break
+			allErrs = append(allErrs, field.Invalid(fldPath, nodePortAddresses[i], "must be a valid CIDR"))
```
thockin, line 279: Shouldn't this be `fldPath.Index(i)` ?

pkg/proxy/apis/config/validation/validation.go
```diff
@@ -307,7 +306,7 @@ func validateIPVSExcludeCIDRs(excludeCIDRs []string, fldPath *field.Path) field.
 
 	for i := range excludeCIDRs {
 		if _, _, err := net.ParseCIDR(excludeCIDRs[i]); err != nil {
-			allErrs = append(allErrs, field.Invalid(fldPath, excludeCIDRs, "must be a valid IP block"))
+			allErrs = append(allErrs, field.Invalid(fldPath, excludeCIDRs[i], "must be a valid CIDR"))
```
thockin, line 309: same as above

thockin (commented):
Small points, then LGTM - sorry to have missed this before

pkg/proxy/apis/config/validation/validation.go
cmluciano: I think this needs to remain fldPath because nodePortAddresses, will not return the full CIDR and it splits after the colon.

Ex. with the change 
```
   validation_test.go:965: Expected error: KubeProxyConfiguration.NodePortAddresses: Invalid value: "::0": must be a valid CIDR, got KubeProxyConfiguration.NodePortAddresses[0]: Invalid value: "::0": must be a valid CIDR
```

thockin: I'm sorry, I don't understand your reply.  The test is failing because the literal error it is expecting references the field "NodePortAddresses", but we almost always want to flag WHICH address so it should be "NodePortAddresses[0]".

cmluciano: That's true, I totally read this incorrectly in my test output, fixing now

thockin (commented):
Thanks!

/lgtm
/approve

PR #95040, opened 2020-09-24
move node performance tests to separate job

MHBauer:
- Node Performance Tests require resources above and beyond normal
   running. Have them skip running instead of fail to run.

 - Take them out of the Flaky suite, as they have been consistently
   running successfully in an evnironment with appropriate resources.

 - A test-infra PR will create a job specific to these tests, and free
   up resources from the regular flaky job.

**What type of PR is this?**
/kind cleanup
/kind flake
/kind regression

Flaky test job successfully runs this test, but has accidentally only ran this test instead of all flaky tests.

See associated PR in test-infra after it is created.

```release-note
NONE
```

### DIFF
--- a/test/e2e_node/node_perf_test.go
+++ b/test/e2e_node/node_perf_test.go
@@ -21,11 +21,13 @@ import (
 	"time"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubeletconfig "k8s.io/kubernetes/pkg/kubelet/apis/config"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
 	"k8s.io/kubernetes/test/e2e_node/perf/workloads"
 
 	"github.com/onsi/ginkgo"
@@ -57,7 +59,7 @@ func setKubeletConfig(f *framework.Framework, cfg *kubeletconfig.KubeletConfigur
 
 // Serial because the test updates kubelet configuration.
 // Slow by design.
-var _ = SIGDescribe("Node Performance Testing [Serial] [Slow] [Flaky]", func() {
+var _ = SIGDescribe("Node Performance Testing [Serial] [Slow]", func() {
 	f := framework.NewDefaultFramework("node-performance-testing")
 	var (
 		wl     workloads.NodePerfWorkload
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev (approved):
/lgtm

test/e2e_node/node_perf_test.go
```diff
@@ -111,6 +113,21 @@ var _ = SIGDescribe("Node Performance Testing [Serial] [Slow] [Flaky]", func() {
 		framework.Logf("Time to complete workload %s: %v", wl.Name(), perf)
 	}
 
+	ginkgo.BeforeEach(func() {
+		ginkgo.By("ensure environment has enough CPU + Memory to run")
+		minimumRequiredCPU := resource.MustParse("15")
+		minimumRequiredMemory := resource.MustParse("48Gi")
```
oomichi, line 119: just a question: Where this requirements (15 CPU and 48GB RAM) come from?
I cannot find the reason from the PR message.

oomichi (commented):
/cc @oomichi

test/e2e_node/node_perf_test.go
MHBauer: @oomichi from the workload descriptions themselves
https://github.com/kubernetes/kubernetes/blob/f77de02aa30994dd082a60b018796970bf8de892/test/e2e_node/perf/workloads/npb_ep.go#L47-L55
https://github.com/kubernetes/kubernetes/blob/f77de02aa30994dd082a60b018796970bf8de892/test/e2e_node/perf/workloads/npb_is.go#L45-L54
https://github.com/kubernetes/kubernetes/blob/f77de02aa30994dd082a60b018796970bf8de892/test/e2e_node/perf/workloads/tf_wide_deep.go#L47-L55
The last one is only 16GiB, so please check all of them.

oomichi: Sorry for my late response, and thanks for your explanation.
I felt it would be better to put different memory condition for 16Gi but such balance (CPU core number vs. RAM size) could not be matter. At least, this addition makes better than current one.

oomichi (commented):
/lgtm

test/e2e_node/node_perf_test.go
```diff
@@ -111,6 +113,21 @@ var _ = SIGDescribe("Node Performance Testing [Serial] [Slow] [Flaky]", func() {
...
+		localNodeCap := getLocalNode(f).Status.Capacity
```
sjenning, line 120: shouldn't this be `Allocatable` not `Capacity`

sjenning (changes requested):

test/e2e_node/node_perf_test.go
MHBauer: I don't think it makes much difference, but it does seem slightly more correct. Updating.

MHBauer:
weird. got my commit wires crossed and pushed something old first. hopefully it's fixed now...

oomichi (commented):
/lgtm

oomichi:
I forgot I can approve..

/approve

Issue #87325, opened 2020-01-17
k8s.gcr.io/pause:3.1 non-amd64 images claim to be for amd64


vrothberg:
**What happened**:

Non-amd64 images of k8s.gcr.io/pause:3.1 claim to be for amd64. Obviously, those images are executing just fine but the image configs claim to be for the amd64 architecture. We just started noticing the issue as the github.com/containers/image library has some more rigorous checks now.

**What you expected to happen**:

Each image config should contain the correct architecture.

**How to reproduce it (as minimally and precisely as possible)**:

```
$ podman pull --override-arch=arm  k8s.gcr.io/pause:3.1
Trying to pull k8s.gcr.io/pause:3.1...
  Image architecture mismatch: image uses "amd64", expecting "arm"
Error: error pulling image "k8s.gcr.io/pause:3.1": unable to pull k8s.gcr.io/pause:3.1: unable to pull image: Image architecture mismatch: image uses "amd64", expecting "arm"
```

A more detailed analysis can be found here: https://github.com/containers/libpod/issues/4877#issuecomment-575637147

**Anything else we need to know?**:

I suggest to update the 3.1 images. Although most runtimes/container engines are happily eating the image, the configs still claim to be for a different architecture. We haven't noticed the issue since the os/arch validity checks are fairly new in the containers/image library.

**Environment**:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

rphillips:
cc @ixdy  @kubernetes/release-engineering

dims:
cc @mkumatag

mkumatag:
I think this is happening because we cross-build these docker images x86 architecture and inspect shows am64 for that particular image it is trying to pull:

```
[root@flexmk3 pause]# docker inspect gcr.io/google-containers/pause-arm:3.1
[
    {
        "Id": "sha256:e11a8cbeda8688fdc3a68dc246d6c945aa272b29f8dd94d0ea370d30c2674042",
        "RepoTags": [
            "gcr.io/google-containers/pause-arm:3.1"
        ],
        "RepoDigests": [
            "gcr.io/google-containers/pause-arm@sha256:c84b0a3a07b628bc4d62e5047d0f8dff80f7c00979e1e28a821a033ecda8fe53"
        ],

[...]

PR #95045, opened 2020-09-24
feat(pause): move to using distroless/static:nonroot instead of scratch

### LINKED ISSUE #95038: Run Linux Pause container as a non-root user
Run the linux pause container as non-root user

**Why is this needed**:
The image is being flagged by scanning tools as not setting a nonroot user in the image definition. While the image is based on scratch which provides some coverage from exploiting vulnerabilities in the image that may escalate privileges it is still a good practice to explicitly set the user in the Dockerfile. There are two options - 

1 - Use a base image that can add a user and then copy /etc/passwd and set the user 
```Dockerfile
FROM ubuntu:20.04 as builder
RUN useradd -u 10001 nonroot
FROM scratch
...
COPY --from=builder /etc/passwd /etc/passwd
USER nonroot
ENTRYPOINT ["/pause"]
```

2 - Switch the image to a base image like distroless/static:nonroot

jchauncey:
**What type of PR is this?**
/kind bug

This pr moves the pause image from scratch to distroless. 

Fixes #95038

```release-note
NONE
```

```docs
```

### DIFF
--- a/build/dependencies.yaml
+++ b/build/dependencies.yaml
@@ -152,7 +152,7 @@ dependencies:
       match: tag =
 
   - name: "k8s.gcr.io/pause"
-    version: 3.3
+    version: 3.4
     refPaths:
     - path: build/pause/Makefile
       match: TAG =
--- a/build/pause/Dockerfile
+++ b/build/pause/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM scratch
+FROM gcr.io/distroless/static:nonroot
 ARG ARCH
 ADD bin/pause-${ARCH} /pause
 ENTRYPOINT ["/pause"]
--- a/build/pause/Makefile
+++ b/build/pause/Makefile
@@ -18,7 +18,7 @@ REGISTRY ?= staging-k8s.gcr.io
 IMAGE = $(REGISTRY)/pause
 IMAGE_WITH_ARCH = $(IMAGE)-$(ARCH)
 
-TAG = 3.3
+TAG = 3.4
 REV = $(shell git describe --contains --always --match='v*')
 
 # Architectures supported: amd64, arm, arm64, ppc64le and s390x


tanjunchen (commented):
/ok-to-test

BenTheElder:
https://github.com/kubernetes/kubernetes/blob/f34d7b5c835d789fd5c89cd318f2051da21fd117/build/pause/Makefile#L21 needs to be changed

then you'll need to file an image promotion PR to the production registry after this PR merges and the post-submit CI builds the image to https://github.com/kubernetes/k8s.io/tree/master/k8s.gcr.io

then we'll need to update kubeadm, dockershim, and some things under test/ to pick up the new version once it's promoted.
we'll also want to send a note at least, if not a PR to cri-o and containerd each.

BenTheElder:
Please squash these commits, otherwise lgtm

Can you confirm that this doesn't affect the multi-arch status of the image? xref: https://github.com/kubernetes/kubernetes/issues/87325, 3.3 had some fun working around this, but I've lost the context since then.

BenTheElder:
ah, you'll also need to update the build/dependencies.yaml entry ..

jchauncey:
Looks like the arch is set correctly for all of the images using `make all-container` to build

```shell
docker images | grep staging-k8s.gcr.io/pause | grep 3.4 | awk '{print $3}' | xargs docker inspect | jq -r '.[] | .Os + " " + .Architecture'
linux s390x
linux ppc64le
linux arm64
linux arm
linux amd64
```

jchauncey:
Alright tests passed let me know if there is anything else I need to do here.

tallclair:
What's the point of moving this image to distroless:static? I'm generally an advocate that distroless:static should always be used over scratch, but the pause image seems like the rare exception tot he rule.

Will running pause as non-root affect orphan process reaping? @verb

jchauncey:
With the recent changes to the build process to include windows we can now override the base image at runtime. My team can just rely on that process to build images based on a nonroot base image.

PR #83647, opened 2019-10-09
tag test that taints a node as disruptive

BenTheElder:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug

/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #95046, opened 2020-09-24
RuntimeHandler test - mark as serial instead of disruptive and provide non-serial alternative

SergeyKanzhelev:
**What type of PR is this?**

/kind cleanup
/sig node
/priority important-longterm

As part of [GA of RuntimeClass](https://docs.google.com/document/d/17nROj6ayPsUpx09mhLrzOkRLgo-8sn6Vgfyy-gmyuo4/edit#) checking on e2e test coverage. This test was marked disruptive https://github.com/kubernetes/kubernetes/pull/83647 with the intention to provide less disruptive alternative.

I wonder if Serial can/should be used instead of Disruptive? Test clears up taints after execution.

I also created a copy of the test the labels selector. I wonder if adding labels considered disruptive as well? This is why it is WIP.

Relates to #83647

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/test/e2e/node/runtimeclass.go
+++ b/test/e2e/node/runtimeclass.go
@@ -24,6 +24,7 @@ import (
 	nodev1beta1 "k8s.io/api/node/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	runtimeclasstest "k8s.io/kubernetes/pkg/kubelet/runtimeclass/testing"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
@@ -38,9 +39,11 @@ var _ = ginkgo.Describe("[sig-node] RuntimeClass", func() {
 	f := framework.NewDefaultFramework("runtimeclass")
 
 	ginkgo.It("should reject a Pod requesting a RuntimeClass with conflicting node selector", func() {
+		labelFooName := "foo-" + string(uuid.NewUUID())
+
 		scheduling := &nodev1beta1.Scheduling{
 			NodeSelector: map[string]string{
-				"foo": "conflict",
+				labelFooName: "conflict",
 			},
 		}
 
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev:
CC: @tallclair

test/e2e/node/runtimeclass.go
```diff
@@ -115,6 +115,44 @@ var _ = ginkgo.Describe("[sig-node] RuntimeClass", func() {
...
+
+		ginkgo.By("Trying to create runtimeclass and pod")
+		runtimeClass := newRuntimeClass(f.Namespace.Name, "non-conflict-runtimeclass", framework.TestContext.ContainerRuntime)
+		runtimeClass.Scheduling = scheduling
+		rc, err := f.ClientSet.NodeV1beta1().RuntimeClasses().Create(context.TODO(), runtimeClass, metav1.CreateOptions{})
+		framework.ExpectNoError(err, "failed to create RuntimeClass resource")
+
+		pod := e2enode.NewRuntimeClassPod(rc.GetName())
+		pod.Spec.NodeSelector = map[string]string{
+			"foo": "bar",
```
tallclair, line 144: foo bar is pretty common, so it's possible there could be a conflict with another test. Probably best to generate some random label. Maybe use the test namespace as the label?

SergeyKanzhelev: fixed. is it what did you have in mind?

SergeyKanzhelev: decided to go with UUID - so tests wouldn't interfere

test/e2e/node/runtimeclass.go
```diff
@@ -115,6 +121,47 @@ var _ = ginkgo.Describe("[sig-node] RuntimeClass", func() {
 		framework.ExpectEqual(nodeSelector, pod.Spec.NodeSelector)
 		gomega.Expect(pod.Spec.Tolerations).To(gomega.ContainElement(tolerations[0]))
 	})
+
+	ginkgo.It("should run a Pod requesting a RuntimeClass with scheduling without taints [NodeFeature:RuntimeHandler] ", func() {
```
tallclair, line 125: Does this actually need the `[NodeFeature:RuntimeHandler]`? If it doesn't have the runtime handler properly configured, the pod will fail, but the test doesn't actually check that the pod is running, so that might be ok?

SergeyKanzhelev: hm.. this is mostly a copy of a test above that has the same logic with the addition of taints. so I didn't put too much thoughts to it.

Looking at code we can replace `WaitForPodNotPending` with `WaitTimeoutForPodReadyInNamespace` to ensure that pod will actually run. But maybe a better solution would be to change `[NodeFeature:RuntimeHandler]` to `[NodeFeature:RuntimeClass]` as the test actually validates this part of the logic. Since RuntimeClass is about to GA, no need to prefix it with the test tag so I will just remove the mention of any feature.

test/e2e/node/runtimeclass.go
```diff
@@ -115,6 +121,47 @@ var _ = ginkgo.Describe("[sig-node] RuntimeClass", func() {
...
+		for key, value := range nodeSelector {
+			framework.AddOrUpdateLabelOnNode(f.ClientSet, nodeName, key, value)
+			framework.ExpectNodeHasLabel(f.ClientSet, nodeName, key, value)
+			defer framework.RemoveLabelOffNode(f.ClientSet, nodeName, key)
+		}
+
+		ginkgo.By("Trying to create runtimeclass and pod")
+		runtimeClass := newRuntimeClass(f.Namespace.Name, "non-conflict-runtimeclass", framework.TestContext.ContainerRuntime)
+		runtimeClass.Scheduling = scheduling
+		rc, err := f.ClientSet.NodeV1beta1().RuntimeClasses().Create(context.TODO(), runtimeClass, metav1.CreateOptions{})
```
tallclair, line 148: note: depending on which PR merges first, remember to update this to `NodeV1()`

SergeyKanzhelev:
/hold 

I want RuntimeClass GA to be merged first. Easier to update this PR

Issue #93718, opened 2020-08-05
Scheduler should log processed component config


damemi:
**What would you like to be added**:
When setting a non-default component config (especially custom profiles/plugin configurations), the scheduler should at least log what plugins have been enabled at startup.

**Why is this needed**:
Debugging the scheduler almost always requires knowing what plugins are enabled. Usually, it is safe to assume that the default registry is being used but this is obviously not always the case. Logging the plugin configuration would also match the current behavior which logs custom Policy settings:

```
I0805 14:15:30.692300       1 factory.go:232] Creating scheduler from configuration: {{ } [] [{ImageLocalityPriority 1000 <nil>}] [] 0 false}
```

/sig scheduling

alculquicondor:
I would expand it to print the whole configuration.

We print all command line args, so it makes sense.

damemi:
@alculquicondor good point it should print the whole config

ahg-g:
+1 to printing the whole configuration once initialized.

alculquicondor:
Actually, the hardest part of this task is to print the whole plugins configuration, as the determination of the default plugins happen kind of late in the startup sequence.

Can we make an initial PR that just prints the configuration as produced by API defaulting today?

alculquicondor:
/retitle Scheduler should log processed component config

damemi:
@alculquicondor I can pick this back up now if you'd like. Any specific timeline or blockers for this?

alculquicondor:
The only partial blocker is that we don't generate the `.profiles[*].plugins` or `profiles[*].pluginConfig` values until much later after API defaulting.

But having something is better than nothing, which is what we have today :)

damemi:
PR opened here: https://github.com/kubernetes/kubernetes/pull/96426, sorry for the delay

Issue #94008, opened 2020-08-14
Deprecate NodeLabel, ServiceAffinityand NodePreferAvoidPods from scheduler's CC api


ahg-g:
NodeLabel and ServiceAffinity are what used to be called customer predicates and priorities. Since we have GA alternatives to those (NodeAffintiy and inter-pod affinity), I believe we should deprecate them from the plugin's API in CC before it gets to GA.

This is important because if CC makes it to GA, we can deprecate those plugins only by going to v2. 

Basically we should:
1) Deprecate their args from CC (requires going to v1Beta2)
2) Keep the logic, and allow configuring them only through the policy api
3) Remove the logic once the policy api is deprecated

/sig scheduling
/cc @mikedanese @Huang-Wei @alculquicondor

hprateek43:
@ahg-g Can work on this. Let me know if we have finalized an approach around this.

ahg-g:
@hprateek43 are you still interested in doing this one? for this release we need to 

- PR1: introduce v1beta2 
  - copy `https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/kube-scheduler/config/v1beta1` to `https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/kube-scheduler/config/v1beta2` and modify the package name to v1beta2 and update the [SchemeGroupVersion](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kube-scheduler/config/v1beta1/register.go#L28)
  - similarly copy `https://github.com/kubernetes/kubernetes/tree/master/pkg/scheduler/apis/config/v1beta1` to `https://github.com/kubernetes/kubernetes/tree/master/pkg/scheduler/apis/config/v1beta2` and make the necessary adjustments.
 - update https://github.com/kubernetes/kubernetes/blob/c87dd9c5121edc26f654f569002d69daee1e1fc4/pkg/scheduler/framework/runtime/framework.go#L330 to v1beta2
- PR2: remove `NodeLabelArgs` and `ServiceAffinityArgs` arguments from the new v1beta2 api

[...]

PR #95048, opened 2020-09-24
Add DefaultingType to PodTopologySpreadArgs

alculquicondor:
/sig scheduling

**What type of PR is this?**

/kind feature

This new enum (`System`, `List`) allows to distinguish between k8s defined and operator provided `defaultConstraints` for `PodTopologySpread` plugin.

Ref #94863

- This is compliant with best API practices of not distinguishing between `nil` and `[]`.
- The system default constant was moved to the plugin, as per API reviewer suggestion.
- For backwards compatibility, the `.defaultingType` is set to List when the `.defaultConstraints` has a non empty value. This will be removed in component config `v1beta2` #94008.

Follow up to #94864

```release-note
New parameter `defaultingType` for `PodTopologySpread` plugin allows to use k8s defined or user provided default constraints
```

```docs
- [KEP]: https://git.k8s.io/enhancements/keps/sig-scheduling/1258-default-pod-topology-spread
- [Usage]: https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/#cluster-level-default-constraints
```

### DIFF
--- a/pkg/scheduler/apis/config/types_pluginargs.go
+++ b/pkg/scheduler/apis/config/types_pluginargs.go
@@ -64,6 +64,17 @@ type NodeResourcesFitArgs struct {
 	IgnoredResourceGroups []string
 }
 
+// PodTopologySpreadConstraintsDefaulting defines how to set default constraints
+// for the PodTopologySpread plugin.
+type PodTopologySpreadConstraintsDefaulting string
+
+const (
+	// SystemDefaulting instructs to use the kubernetes defined default.
+	SystemDefaulting PodTopologySpreadConstraintsDefaulting = "System"
+	// ListDefaulting instructs to use the config provided default.
+	ListDefaulting PodTopologySpreadConstraintsDefaulting = "List"
+)
+
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // PodTopologySpreadArgs holds arguments used to configure the PodTopologySpread plugin.
@@ -71,12 +82,24 @@ type PodTopologySpreadArgs struct {
 	metav1.TypeMeta
 
 	// DefaultConstraints defines topology spread constraints to be applied to
-	// pods that don't define any in `pod.spec.topologySpreadConstraints`.
-	// `topologySpreadConstraint.labelSelectors` must be empty, as they are
-	// deduced the pods' membership to Services, Replication Controllers, Replica
-	// Sets or Stateful Sets.
-	// Empty by default.
+	// Pods that don't define any in `pod.spec.topologySpreadConstraints`.
+	// `.defaultConstraints[*].labelSelectors` must be empty, as they are
+	// deduced from the Pod's membership to Services, ReplicationControllers,
+	// ReplicaSets or StatefulSets.
+	// When not empty, .defaultingType must be "List".
 	DefaultConstraints []v1.TopologySpreadConstraint
+
+	// DefaultingType determines how .defaultConstraints are deduced. Can be one
+	// of "System" or "List".
+	//
+	// - "System": Use kubernetes defined constraints that spread Pods among
+	//   Nodes and Zones.
+	// - "List": Use constraints defined in .defaultConstraints.
+	//
+	// Defaults to "List" if feature gate DefaultPodTopologySpread is disabled
+	// and to "System" if enabled.
+	// +optional
+	DefaultingType PodTopologySpreadConstraintsDefaulting
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/scheduler/apis/config/v1beta1/defaults.go
+++ b/pkg/scheduler/apis/config/v1beta1/defaults.go
@@ -195,23 +195,18 @@ func SetDefaults_VolumeBindingArgs(obj *v1beta1.VolumeBindingArgs) {
 }
 
 func SetDefaults_PodTopologySpreadArgs(obj *v1beta1.PodTopologySpreadArgs) {
-	if !feature.DefaultFeatureGate.Enabled(features.DefaultPodTopologySpread) {
-		// When feature is disabled, the default spreading is done by legacy
-		// SelectorSpread plugin.
+	if feature.DefaultFeatureGate.Enabled(features.DefaultPodTopologySpread) {
+		if obj.DefaultingType == "" {
+			// TODO(#94008): Always default to System in v1beta2.
+			if len(obj.DefaultConstraints) != 0 {
+				obj.DefaultingType = v1beta1.ListDefaulting
+			} else {
+				obj.DefaultingType = v1beta1.SystemDefaulting
+			}
+		}
 		return
 	}
-	if obj.DefaultConstraints == nil {
-		obj.DefaultConstraints = []corev1.TopologySpreadConstraint{
-			{
-				TopologyKey:       corev1.LabelHostname,
-				WhenUnsatisfiable: corev1.ScheduleAnyway,
-				MaxSkew:           3,
-			},
-			{
-				TopologyKey:       corev1.LabelZoneFailureDomainStable,
-				WhenUnsatisfiable: corev1.ScheduleAnyway,
-				MaxSkew:           5,
-			},
-		}
+	if obj.DefaultingType == "" {
+		obj.DefaultingType = v1beta1.ListDefaulting
 	}
 }
--- a/pkg/scheduler/apis/config/validation/validation_pluginargs.go
+++ b/pkg/scheduler/apis/config/validation/validation_pluginargs.go
@@ -75,6 +75,9 @@ func validateNoConflict(presentLabels []string, absentLabels []string) error {
 // with an additional check for .labelSelector to be nil.
 func ValidatePodTopologySpreadArgs(args *config.PodTopologySpreadArgs) error {
 	var allErrs field.ErrorList
+	if err := validateDefaultingType(field.NewPath("defaultingType"), args.DefaultingType, args.DefaultConstraints); err != nil {
+		allErrs = append(allErrs, err)
+	}
 	path := field.NewPath("defaultConstraints")
 
 	for i, c := range args.DefaultConstraints {
@@ -99,6 +102,16 @@ func ValidatePodTopologySpreadArgs(args *config.PodTopologySpreadArgs) error {
 		return nil
 	}
 	return allErrs.ToAggregate()
+}
+
+func validateDefaultingType(p *field.Path, v config.PodTopologySpreadConstraintsDefaulting, constraints []v1.TopologySpreadConstraint) *field.Error {
+	if v != config.SystemDefaulting && v != config.ListDefaulting {
+		return field.Invalid(p, v, fmt.Sprintf("must be one of {%q, %q}", config.SystemDefaulting, config.ListDefaulting))
+	}
+	if v == config.SystemDefaulting && len(constraints) > 0 {
+		return field.Invalid(p, v, "when .defaultConstraints are not empty")
+	}
+	return nil
 }
 
 func validateTopologyKey(p *field.Path, v string) field.ErrorList {
--- a/pkg/scheduler/framework/plugins/podtopologyspread/plugin.go
+++ b/pkg/scheduler/framework/plugins/podtopologyspread/plugin.go
@@ -19,6 +19,7 @@ package podtopologyspread
 import (
 	"fmt"
 
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/informers"
 	appslisters "k8s.io/client-go/listers/apps/v1"
@@ -35,14 +36,27 @@ const (
 	ErrReasonNodeLabelNotMatch = ErrReasonConstraintsNotMatch + " (missing required label)"
 )
 
+var systemDefaultConstraints = []v1.TopologySpreadConstraint{
+	{
+		TopologyKey:       v1.LabelHostname,
+		WhenUnsatisfiable: v1.ScheduleAnyway,
+		MaxSkew:           3,
+	},
+	{
+		TopologyKey:       v1.LabelZoneFailureDomainStable,
+		WhenUnsatisfiable: v1.ScheduleAnyway,
+		MaxSkew:           5,
+	},
+}
+
 // PodTopologySpread is a plugin that ensures pod's topologySpreadConstraints is satisfied.
 type PodTopologySpread struct {
-	args             config.PodTopologySpreadArgs
-	sharedLister     framework.SharedLister
-	services         corelisters.ServiceLister
-	replicationCtrls corelisters.ReplicationControllerLister
-	replicaSets      appslisters.ReplicaSetLister
-	statefulSets     appslisters.StatefulSetLister
+	defaultConstraints []v1.TopologySpreadConstraint
+	sharedLister       framework.SharedLister
+	services           corelisters.ServiceLister
+	replicationCtrls   corelisters.ReplicationControllerLister
+	replicaSets        appslisters.ReplicaSetLister
+	statefulSets       appslisters.StatefulSetLister
 }
 
 var _ framework.PreFilterPlugin = &PodTopologySpread{}
@@ -73,10 +87,13 @@ func New(plArgs runtime.Object, h framework.FrameworkHandle) (framework.Plugin, 
 		return nil, err
 	}
 	pl := &PodTopologySpread{
-		sharedLister: h.SnapshotSharedLister(),
-		args:         args,
+		sharedLister:       h.SnapshotSharedLister(),
+		defaultConstraints: args.DefaultConstraints,
 	}
-	if len(pl.args.DefaultConstraints) != 0 {
+	if args.DefaultingType == config.SystemDefaulting {
+		pl.defaultConstraints = systemDefaultConstraints
+	}
+	if len(pl.defaultConstraints) != 0 {
 		if h.SharedInformerFactory() == nil {
 			return nil, fmt.Errorf("SharedInformerFactory is nil")
 		}
--- a/staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
+++ b/staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
@@ -70,6 +70,17 @@ type NodeResourcesFitArgs struct {
 	IgnoredResourceGroups []string `json:"ignoredResourceGroups,omitempty"`
 }
 
+// PodTopologySpreadConstraintsDefaulting defines how to set default constraints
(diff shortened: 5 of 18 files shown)


pkg/scheduler/apis/config/validation/validation_pluginargs.go
```diff
@@ -75,6 +75,10 @@ func validateNoConflict(presentLabels []string, absentLabels []string) error {
 // with an additional check for .labelSelector to be nil.
 func ValidatePodTopologySpreadArgs(args *config.PodTopologySpreadArgs) error {
 	var allErrs field.ErrorList
+	if args.DisableDefaultConstraints && len(args.DefaultConstraints) > 0 {
+		p := field.NewPath("disableDefaultConstraints")
+		allErrs = append(allErrs, field.Invalid(p, args.DisableDefaultConstraints, "can not be true when there .defaultConstraints are not empty"))
```
Huang-Wei, line 80:
```suggestion
		allErrs = append(allErrs, field.Invalid(p, args.DisableDefaultConstraints, "cannot be true when .defaultConstraints are not empty"))
```

Huang-Wei (commented):
A nit. LGTM otherwise.

pkg/scheduler/apis/config/types_pluginargs.go
```diff
@@ -71,12 +71,17 @@ type PodTopologySpreadArgs struct {
...
 
 	// DefaultConstraints defines topology spread constraints to be applied to
-	// pods that don't define any in `pod.spec.topologySpreadConstraints`.
-	// `topologySpreadConstraint.labelSelectors` must be empty, as they are
-	// deduced the pods' membership to Services, Replication Controllers, Replica
-	// Sets or Stateful Sets.
-	// Empty by default.
+	// Pods that don't define any in `pod.spec.topologySpreadConstraints`.
+	// `.defaultConstraints[*].labelSelectors` must be empty, as they are
+	// deduced the Pod's membership to Services, ReplicationControllers,
```
adtac, line 76:
```suggestion
	// deduced from the Pod's membership to Services, ReplicationControllers,
```

adtac (commented):
minor nit, LGTM otherwise

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
```diff
@@ -77,14 +77,18 @@ type PodTopologySpreadArgs struct {
...
-	// deduced the pods' membership to Services, Replication Controllers, Replica
-	// Sets or Stateful Sets.
-	// Empty by default.
-	// +optional
+	// Pods that don't define any in `pod.spec.topologySpreadConstraints`.
+	// `.defaultConstraints[*].labelSelectors` must be empty, as they are
+	// deduced the Pod's membership to Services, ReplicationControllers,
+	// ReplicaSets or StatefulSets.
+	// If empty, the default constraints prefer to spread Pods across Nodes and Zones
 	// +listType=atomic
```
chendave, line 85: add back "// +optional" for this config and `DisableDefaultConstraints` below.

alculquicondor: What was that for again? What's the documentation for it?

I arrived to this, which doesn't help https://github.com/kubernetes/kube-openapi/blob/master/pkg/idl/doc.go#L109

chendave: We'd better to keep the still valid comment directive here. I don't have a better documentation for it either, seems like this will help to generate a better API documentation? I am not sure.

chendave:
LGTM, seem like this change still needs review from API owner.

pkg/scheduler/apis/config/validation/validation_pluginargs.go
```diff
@@ -75,6 +75,10 @@ func validateNoConflict(presentLabels []string, absentLabels []string) error {
...
+	}
```
soulxu, line 81: not sure whether we should return here, to avoid return too much error info to confuse the user.

Huang-Wei: We should gather all errs and then return

alculquicondor:
/assign @liggitt 
for API review

alculquicondor:
/unassign @liggitt 

/assign @lavalamp 

for API review

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
```diff
@@ -77,14 +77,20 @@ type PodTopologySpreadArgs struct {
...
-	// pods that don't define any in `pod.spec.topologySpreadConstraints`.
-	// `topologySpreadConstraint.labelSelectors` must be empty, as they are
-	// deduced the pods' membership to Services, Replication Controllers, Replica
-	// Sets or Stateful Sets.
-	// Empty by default.
+	// Pods that don't define any in `pod.spec.topologySpreadConstraints`.
+	// `.defaultConstraints[*].labelSelectors` must be empty, as they are
+	// deduced from the Pod's membership to Services, ReplicationControllers,
+	// ReplicaSets or StatefulSets.
+	// If empty, the default constraints prefer to spread Pods across Nodes and Zones
```
lavalamp, line 84: And if you don't want that behavior either, you use the next field? (add comment)

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
```diff
@@ -77,14 +77,20 @@ type PodTopologySpreadArgs struct {
...
 	// +optional
 	// +listType=atomic
-	DefaultConstraints []corev1.TopologySpreadConstraint `json:"defaultConstraints"`
+	DefaultConstraints []corev1.TopologySpreadConstraint `json:"defaultConstraints,omitempty"`
+
+	// DisableDefaultConstraints allows to disable .defaultConstraints.
+	// Defaults to false. When set to true, DefaultConstraints must be empty
+	// or nil.
+	// +optional
+	DisableDefaultConstraints bool `json:"disableDefaultConstraints,omitempty"`
```
lavalamp, line 93: Can you flip the polarity, e.g. `UseDefaultConstraints` defaulted true?

pkg/scheduler/apis/config/validation/validation_pluginargs.go
```diff
@@ -75,6 +75,10 @@ func validateNoConflict(presentLabels []string, absentLabels []string) error {
 // with an additional check for .labelSelector to be nil.
 func ValidatePodTopologySpreadArgs(args *config.PodTopologySpreadArgs) error {
 	var allErrs field.ErrorList
+	if args.DisableDefaultConstraints && len(args.DefaultConstraints) > 0 {
+		p := field.NewPath("disableDefaultConstraints")
+		allErrs = append(allErrs, field.Invalid(p, args.DisableDefaultConstraints, "cannot be true when .defaultConstraints are not empty"))
```
lavalamp, line 80:
```suggestion
		allErrs = append(allErrs, field.Invalid(p, args.DisableDefaultConstraints, "is true, but .defaultConstraints are not empty"))
```

(obviously s/true/false/ after switch polarity :) )

pkg/scheduler/apis/config/v1beta1/defaults.go
```diff
@@ -200,7 +200,7 @@ func SetDefaults_PodTopologySpreadArgs(obj *v1beta1.PodTopologySpreadArgs) {
 		// SelectorSpread plugin.
 		return
 	}
-	if obj.DefaultConstraints == nil {
+	if len(obj.DefaultConstraints) == 0 && !obj.DisableDefaultConstraints {
```
lavalamp, line 203: I guess people were providing an empty list to avoid getting the default? Yeah, that's bad.

This conditional defaulting is going to give you other problems though.

I actually recommend deleting this defaulting completely and making the consumer of the config do this logic.

In particular I think you should avoid mixing system defaults and user provided defaults in a single field in this layer.

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
lavalamp: After reading your implementation, I think if we were starting from scratch, you'd actually want a union/one-of?

E.g. the discriminator (this field) is "DefaultingType" and can be one of "SystemDefault", "List", or "None". The DefaultConstraints list may only be set if this field is "List".

Since there's existing files we want to not break (right?), I think we can default this discriminator to "List" if the list has items in it, or to "SystemDefault" otherwise. I hate to do this but I think it's OK since it's for backwards compatibility, and this is just a config file.

pkg/scheduler/apis/config/v1beta1/defaults.go
alculquicondor: > I guess people were providing an empty list to avoid getting the default? Yeah, that's bad.

Yeah, my bad. OTOH, the system defaulting is an alpha feature.

> I actually recommend deleting this defaulting completely and making the consumer of the config do this logic.

Even though this code is within pkg/scheduler, i.e. the consumer?

> This conditional defaulting is going to give you other problems though.

Could you explain?

Also, having this in defaulting makes it easier to share what configuration is the plugin actually using, once we resolve #93718

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
alculquicondor: I like DefaultingType. Is this just a regular enum that you called union because of its interpretation?

> Since there's existing files we want to not break (right?), I think we can default this discriminator to "List" if the list has items in it, or to "SystemDefault" otherwise.

It's alpha, so I assume it's fine to break (with the appropriate release note). However, is it normal to have to be this verbose?

lavalamp: Yeah, we're eventually going to get unions as first-class citizens of the API.

It's a little verbose but it makes it very clear what you want the system to do.

pkg/scheduler/apis/config/v1beta1/defaults.go
lavalamp: > Even though this code is within pkg/scheduler, i.e. the consumer?

The defaulting function is called during deserialization, which means original user intent is gone by the time the eventual actual consumer gets the go structures.

> > This conditional defaulting is going to give you other problems though.

> Could you explain?

We're adding support for static defaulting, but it won't be able to replace conditional defaulting which depends on multiple fields. I don't know if I can make a great argument for this off the top of my head, but we've found that complicated (e.g. conditional or multi-field) defaulting logic is almost always regretted later.

It's less important because this is a config file, but I'd still try to avoid it.

In my mind there's two concepts that are mixed here, "what value does this field have if it's not specified", and "what does the application do given certain input". This distinction isn't imposed on us by the laws of the universe, but imposing it on ourselves makes the codebase easier to understand, e.g. less spooky action at a distance.

If the system were not ever going to evolve then I'd be fine with the first version of this I saw. But given that it will evolve, I think it's easier to evolve this logic when it's in the application layer and not the defaulting (deserialization) layer.

I hope that makes sense, sorry :)

> easier to share what configuration is the plugin actually using

I'm not sure how? It sounds like stuff that'd go in a "status" section except we don't have those for config files.

pkg/scheduler/apis/config/v1beta1/defaults.go
```diff
@@ -195,23 +195,13 @@ func SetDefaults_VolumeBindingArgs(obj *v1beta1.VolumeBindingArgs) {
 }
 
 func SetDefaults_PodTopologySpreadArgs(obj *v1beta1.PodTopologySpreadArgs) {
-	if !feature.DefaultFeatureGate.Enabled(features.DefaultPodTopologySpread) {
-		// When feature is disabled, the default spreading is done by legacy
-		// SelectorSpread plugin.
+	if feature.DefaultFeatureGate.Enabled(features.DefaultPodTopologySpread) {
+		if obj.DefaultingType != "" {
+			obj.DefaultingType = v1beta1.SystemDefaulting
```
alculquicondor, line 200: @lavalamp help me out here.

Even though "system defaulting" is alpha (graduating to beta in 1.20), "list defaulting", i.e. the slice, is a beta API.
Once we flip the feature gate, users of "list defaulting" would break because of this line.

Should I still have the "if slice is not empty, set to ListDefaulting" marked for removal in v1beta2, which we are indeed planing.

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
alculquicondor: Actually, isn't "None" unnecessary? "List" + empty slice should be enough.

alculquicondor:
Sorry for the failed tests spam. I didn't fix them before getting some resolution on the approach.

pkg/scheduler/apis/config/v1beta1/defaults.go
lavalamp: We shouldn't do something that would cause a beta feature to change behavior. It's OK for the next API version to do something different, but an existing config file shouldn't change behavior, I think. Does that help?

alculquicondor:
/hold
for squash

I need another sig-scheduling pass @Huang-Wei

Huang-Wei (commented):
LGTM. One nit.

pkg/scheduler/apis/config/types_pluginargs.go
```diff
@@ -64,19 +64,38 @@ type NodeResourcesFitArgs struct {
...
 	DefaultConstraints []v1.TopologySpreadConstraint
+
+	// DefaultingType determines how .defaultConstraints are deduced. Can be one
+	// of "System" or "List".
+	//
+	// - "System": Use kubernetes defined constraints that spread Pods among
+	//   Nodes and Zones.
+	// - "List": Use constraints defined in .defaultConstraints.
+	// +optional
+	DefaultingType PodTopologySpreadConstraintsDefaulting `json:"defaultingType,omitempty"`
```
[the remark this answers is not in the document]
alculquicondor: copy-paste mistake :)

alculquicondor:
@lavalamp ready for API review again :)

pkg/scheduler/apis/config/validation/validation_pluginargs.go
```diff
@@ -101,6 +104,16 @@ func ValidatePodTopologySpreadArgs(args *config.PodTopologySpreadArgs) error {
 	return allErrs.ToAggregate()
 }
 
+func validateDefaultingType(p *field.Path, v config.PodTopologySpreadConstraintsDefaulting, constraints []v1.TopologySpreadConstraint) *field.Error {
+	if v != config.SystemDefaulting && v != config.ListDefaulting {
+		return field.Invalid(p, v, fmt.Sprintf("must be one of {%q, %q}", config.SystemDefaulting, config.ListDefaulting))
+	}
+	if v == config.SystemDefaulting && len(constraints) > 0 {
+		return field.Invalid(p, v, "is System, but .defaultConstraints are not empty")
```
adtac, line 112:
```suggestion
		return field.Invalid(p, v, "is %q, but .defaultConstraints is not empty", config.SystemDefaulting)
```

pkg/scheduler/framework/plugins/podtopologyspread/plugin.go
```diff
@@ -35,14 +36,27 @@ const (
 	ErrReasonNodeLabelNotMatch = ErrReasonConstraintsNotMatch + " (missing required label)"
 )
 
+var systemDefaultConstraints = []v1.TopologySpreadConstraint{
```
adtac, line 39: why do we define this here instead of defaults.go?

adtac (commented):
just a couple of minor nits

pkg/scheduler/framework/plugins/podtopologyspread/plugin.go
alculquicondor: See https://github.com/kubernetes/kubernetes/pull/95048#discussion_r500580974

alculquicondor:
Ping @lavalamp

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
```diff
@@ -70,21 +70,40 @@ type NodeResourcesFitArgs struct {
...
 	// DefaultConstraints defines topology spread constraints to be applied to
-	// pods that don't define any in `pod.spec.topologySpreadConstraints`.
-	// `topologySpreadConstraint.labelSelectors` must be empty, as they are
-	// deduced the pods' membership to Services, Replication Controllers, Replica
-	// Sets or Stateful Sets.
-	// Empty by default.
+	// Pods that don't define any in `pod.spec.topologySpreadConstraints`.
+	// `.defaultConstraints[*].labelSelectors` must be empty, as they are
+	// deduced from the Pod's membership to Services, ReplicationControllers,
+	// ReplicaSets or StatefulSets.
```
lavalamp, line 94: Mention that this only applies if .defaultingType is "List".

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
```diff
@@ -70,21 +70,40 @@ type NodeResourcesFitArgs struct {
...
-	DefaultConstraints []corev1.TopologySpreadConstraint `json:"defaultConstraints"`
+	DefaultConstraints []corev1.TopologySpreadConstraint `json:"defaultConstraints,omitempty"`
+
+	// DefaultingType determines how .defaultConstraints are deduced. Can be one
+	// of "System" or "List".
+	//
+	// - "System": Use kubernetes defined constraints that spread Pods among
+	//   Nodes and Zones.
+	// - "List": Use constraints defined in .defaultConstraints.
+	// +optional
```
lavalamp, line 105: Mention the default of this field?

lavalamp:
/approve

... and so you can fix the comments:
/hold

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
alculquicondor: It depends on the feature gate. I guess that's fair to mention until it gets deleted?

lavalamp: Yeah, I think so.

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
alculquicondor: Done

staging/src/k8s.io/kube-scheduler/config/v1beta1/types_pluginargs.go
alculquicondor: Done

PR #95063, opened 2020-09-25
[WIP] Add more dual stack tests

aojea:
Add e2e and integration tests for the new dual stack feature that pluralize cluster IPs

### DIFF
--- a/test/e2e/framework/network/utils.go
+++ b/test/e2e/framework/network/utils.go
@@ -97,6 +97,14 @@ func NewNetworkingTestConfig(f *framework.Framework, hostNetwork, SCTPEnabled bo
 	return config
 }
 
+// NewDualNetworkingTestConfig creates and sets up a new test config helper.
+func NewDualNetworkingTestConfig(f *framework.Framework, hostNetwork, SCTPEnabled bool) *NetworkingTestConfig {
+	config := &NetworkingTestConfig{f: f, Namespace: f.Namespace.Name, HostNetwork: hostNetwork, SCTPEnabled: SCTPEnabled, DualStackEnabled: true}
+	ginkgo.By(fmt.Sprintf("Performing setup for networking test in namespace %v", config.Namespace))
+	config.setup(getServiceSelector())
+	return config
+}
+
 // NewCoreNetworkingTestConfig creates and sets up a new test config helper for Node E2E.
 func NewCoreNetworkingTestConfig(f *framework.Framework, hostNetwork bool) *NetworkingTestConfig {
 	config := &NetworkingTestConfig{f: f, Namespace: f.Namespace.Name, HostNetwork: hostNetwork}
@@ -127,6 +135,8 @@ type NetworkingTestConfig struct {
 	// if the test pods are listening on sctp port. We need this as sctp tests
 	// are marked as disruptive as they may load the sctp module.
 	SCTPEnabled bool
+	// DualStackEnabled enables dual stack on services
+	DualStackEnabled bool
 	// EndpointPods are the pods belonging to the Service created by this
 	// test config. Each invocation of `setup` creates a service with
 	// 1 pod per node running the netexecImage.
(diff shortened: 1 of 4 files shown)


test/integration/dualstack/dualstack_test.go
```diff
@@ -0,0 +1,1386 @@
...
+
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
+	featuregatetesting "k8s.io/component-base/featuregate/testing"
+
+	"k8s.io/kubernetes/pkg/features"
+	"k8s.io/kubernetes/test/integration/framework"
+	netutils "k8s.io/utils/net"
+)
+
+// TestCreateServiceSingleStackIPv4 test the Service dualstackness in an IPv4 SingleStack cluster
```
danwinship, line 45: can't really test "dualstackness" in a single-stack cluster... "test the dual-stack Service APIs"?

test/integration/dualstack/dualstack_test.go
```diff
@@ -0,0 +1,1386 @@
...
+		t.Fatalf("bad cidr: %v", err)
+	}
+	cfg.ExtraConfig.ServiceIPRange = *cidr
+	_, s, closeFn := framework.RunAMaster(cfg)
+	defer closeFn()
+
+	client := clientset.NewForConfigOrDie(&restclient.Config{Host: s.URL})
+
+	// Wait until the default "kubernetes" service is created.
+	if err = wait.Poll(250*time.Millisecond, time.Minute, func() (bool, error) {
```
danwinship, line 65: If this was an e2e test I'd say you shouldn't use hardcoded timeouts and should use some constant from `tests/e2e/framework/...` instead. I'm not sure about integration tests...

test/integration/dualstack/dualstack_test.go
```diff
@@ -0,0 +1,1386 @@
...
+		{
+			name:           "Type ClusterIP - Server Allocated IP - IPv4 Family - Policy Prefer Dual Stack",
+			serviceType:    v1.ServiceTypeClusterIP,
+			clusterIPs:     []string{},
+			ipFamilies:     []v1.IPFamily{v1.IPv4Protocol},
+			ipFamilyPolicy: v1.PreferDualStack,
+			expectError:    false,
+		},
+		{
+			name:           "Type ClusterIP - Server Allocated IP - IPv4 Family - Policy Required Dual Stack",
```
danwinship, line 124: "Require" (no "d")
(everywhere)

test/integration/dualstack/dualstack_test.go
```diff
@@ -0,0 +1,1386 @@
...
+				ipFamilies = []v1.IPFamily{defaultIPFamily}
+			}
+			if err := validateServiceAndClusterIPFamily(svc, ipFamilies); err != nil {
+				t.Errorf("Unexpected error validating the service %s %v", svc.Name, err)
+			}
+		})
+	}
+}
+
+// TestCreateServiceDualStackIPv4Only test the Service dualstackness in an IPv4 only DualStack cluster
```
danwinship, line 465: "in a cluster with the dual-stack feature gate enabled but an IPv4-only configuration"

test/integration/dualstack/dualstack_test.go
```diff
@@ -0,0 +1,1386 @@
...
+	if svc.Spec.IPFamilies == nil {
+		return fmt.Errorf("service ip family nil for service %s/%s", svc.Namespace, svc.Name)
+	}
+	if !reflect.DeepEqual(svc.Spec.IPFamilies, expectedIPFamilies) {
+		return fmt.Errorf("ip families mismatch for service: %s/%s, expected: %s, actual: %s", svc.Namespace, svc.Name, expectedIPFamilies, svc.Spec.IPFamilies)
+	}
+
+	for j, ip := range svc.Spec.ClusterIPs {
+		if ip == v1.ClusterIPNone {
+			// ???
```
danwinship, line 1371: I forget if we decided this is allowed or not, but if it is, it would be an error in these tests

danwinship (commented):
Good first draft. There is a ton of duplication here that can be reduced:

1. I'm pretty sure you only need a single "test-runner" function, that takes arguments indicating the configured CIDR(s), feature gate status, etc.
2. The set of test cases is identical in each case; it's only the expected results that differ. Instead of having a single `expectError`, you could have `singleStackIPv4OK`, `singleStackIPv6OK`, `dualStackOK` and then you'd only need one copy of the array.
3. (Related to point 2) "Single stack IPv4 with the feature gate disabled" and "single stack IPv4 with the feature gate enabled" have exactly the same expected results, and likewise for the two single-stack IPv6 versions. The two dual stack versions (v4v6 / v6v4) have different expected `IPFamilies`/`ClusterIPs`, but the same expected success/failure.

PR #11056, opened 2015-07-10
Rewrite services shell test in Go.

wojtek-t:
Ref #4184

PR #95065, opened 2020-09-25
refector service some e2e cases to make it runing in multi providers

### LINKED ISSUE #94703: Replace e2essh on e2e service tests

JornShen:
**What type of PR is this?**
/kind bug
/kind cleanup

1. Background
    There are some cases still can not running in multi provider. These PR fix the cases are as fellow:
    - "should work after restarting kube-proxy"
    - "should work after restarting apiserver"

 2. changes 
     refector the approach of restarting kube-proxy and kube-apiserver. Before the case kill the process and make the kubelet to restart the static pod. This PR change to kill the pod, same with the effect of restaring.

Fixes  https://github.com/kubernetes/kubernetes/issues/94703
part with https://github.com/kubernetes/kubernetes/pull/94822#discussion_r493476807 

```release-note
NONE
```

### DIFF
--- a/test/e2e/network/service.go
+++ b/test/e2e/network/service.go
@@ -60,7 +60,6 @@ import (
 	e2erc "k8s.io/kubernetes/test/e2e/framework/rc"
 	e2eservice "k8s.io/kubernetes/test/e2e/framework/service"
 	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
-	e2essh "k8s.io/kubernetes/test/e2e/framework/ssh"
 	"k8s.io/kubernetes/test/e2e/storage/utils"
 	testutils "k8s.io/kubernetes/test/utils"
 	imageutils "k8s.io/kubernetes/test/utils/image"
@@ -83,6 +82,12 @@ const (
 	// AffinityConfirmCount is the number of needed continuous requests to confirm that
 	// affinity is enabled.
 	AffinityConfirmCount = 15
+
+	// label define which is used to find kube-proxy and kube-apiserver pod
+	kubeProxyLabelName     = "kube-proxy"
+	clusterAddonLabelKey   = "k8s-app"
+	kubeAPIServerLabelName = "kube-apiserver"
+	clusterComponentKey    = "component"
 )
 
 var (
@@ -703,45 +708,8 @@ func getServeHostnameService(name string) *v1.Service {
 	return svc
 }
 
-// restartKubeProxy restarts kube-proxy on the given host.
-func restartKubeProxy(host string) error {
-	// TODO: Make it work for all providers.
-	if !framework.ProviderIs("gce", "gke", "aws") {
-		return fmt.Errorf("unsupported provider for restartKubeProxy: %s", framework.TestContext.Provider)
-	}
-	// kubelet will restart the kube-proxy since it's running in a static pod
-	framework.Logf("Killing kube-proxy on node %v", host)
-	result, err := e2essh.SSH("sudo pkill kube-proxy", host, framework.TestContext.Provider)
-	if err != nil || result.Code != 0 {
-		e2essh.LogResult(result)
-		return fmt.Errorf("couldn't restart kube-proxy: %v", err)
-	}
-	// wait for kube-proxy to come back up
-	sshCmd := "sudo /bin/sh -c 'pgrep kube-proxy | wc -l'"
-	err = wait.Poll(5*time.Second, 60*time.Second, func() (bool, error) {
-		framework.Logf("Waiting for kubeproxy to come back up with %v on %v", sshCmd, host)
-		result, err := e2essh.SSH(sshCmd, host, framework.TestContext.Provider)
-		if err != nil {
-			return false, err
-		}
-		if result.Code != 0 {
-			e2essh.LogResult(result)
-			return false, fmt.Errorf("failed to run command, exited %d", result.Code)
-		}
-		if result.Stdout == "0\n" {
-			return false, nil
-		}
-		framework.Logf("kube-proxy is back up.")
-		return true, nil
-	})
-	if err != nil {
-		return fmt.Errorf("kube-proxy didn't recover: %v", err)
-	}
-	return nil
-}
-
-// waitForApiserverUp waits for the kube-apiserver to be up.
-func waitForApiserverUp(c clientset.Interface) error {
+// waitForAPIServerUp waits for the kube-apiserver to be up.
+func waitForAPIServerUp(c clientset.Interface) error {
 	for start := time.Now(); time.Since(start) < time.Minute; time.Sleep(5 * time.Second) {
 		body, err := c.CoreV1().RESTClient().Get().AbsPath("/healthz").Do(context.TODO()).Raw()
 		if err == nil && string(body) == "ok" {
@@ -1101,10 +1069,10 @@ var _ = SIGDescribe("Services", func() {
 	})
 
 	ginkgo.It("should work after restarting kube-proxy [Disruptive]", func() {
-		// TODO: use the ServiceTestJig here
-		e2eskipper.SkipUnlessProviderIs("gce", "gke")
-		e2eskipper.SkipUnlessSSHKeyPresent()
+		kubeProxyLabelSet := map[string]string{clusterAddonLabelKey: kubeProxyLabelName}
+		e2eskipper.SkipUnlessComponentRunsAsPodsAndClientCanDeleteThem(kubeProxyLabelName, cs, metav1.NamespaceSystem, kubeProxyLabelSet)
 
+		// TODO: use the ServiceTestJig here
 		ns := f.Namespace.Name
 		numPods, servicePort := 3, defaultServeHostnameServicePort
 
@@ -1127,18 +1095,10 @@ var _ = SIGDescribe("Services", func() {
 			framework.Failf("VIPs conflict: %v", svc1IP)
 		}
 
-		hosts, err := e2essh.NodeSSHHosts(cs)
-		framework.ExpectNoError(err, "failed to find external/internal IPs for every node")
-		if len(hosts) == 0 {
-			framework.Failf("No ssh-able nodes")
-		}
-		host := hosts[0]
-
 		framework.ExpectNoError(verifyServeHostnameServiceUp(cs, ns, podNames1, svc1IP, servicePort))
 		framework.ExpectNoError(verifyServeHostnameServiceUp(cs, ns, podNames2, svc2IP, servicePort))
 
-		ginkgo.By(fmt.Sprintf("Restarting kube-proxy on %v", host))
-		if err := restartKubeProxy(host); err != nil {
+		if err := restartComponent(cs, kubeProxyLabelName, metav1.NamespaceSystem, kubeProxyLabelSet); err != nil {
 			framework.Failf("error restarting kube-proxy: %v", err)
 		}
 		framework.ExpectNoError(verifyServeHostnameServiceUp(cs, ns, podNames1, svc1IP, servicePort))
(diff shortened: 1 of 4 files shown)


test/e2e/network/service.go
```diff
@@ -3733,84 +3687,64 @@ func validateEndpointsPorts(c clientset.Interface, namespace, serviceName string
...
 
-func sshRestartMaster() error {
-	if !framework.ProviderIs("gce", "aws") {
-		return fmt.Errorf("unsupported provider for sshRestartMaster: %s", framework.TestContext.Provider)
-	}
-	var command string
-	if framework.ProviderIs("gce") {
-		command = "pidof kube-apiserver | xargs sudo kill"
-	} else {
-		command = "sudo /etc/init.d/kube-apiserver restart"
```
aojea, line 3769: this is the change that worries me, and we need people from sig-testing to chime in.
With the new approach my understanding is that this test will only work when the apiserver runs inside a pod, and I don't thinik this is backwards compatible

test/e2e/network/service.go
```diff
@@ -703,45 +708,8 @@ func getServeHostnameService(name string) *v1.Service {
 	return svc
 }
 
-// restartKubeProxy restarts kube-proxy on the given host.
-func restartKubeProxy(host string) error {
-	// TODO: Make it work for all providers.
-	if !framework.ProviderIs("gce", "gke", "aws") {
-		return fmt.Errorf("unsupported provider for restartKubeProxy: %s", framework.TestContext.Provider)
-	}
-	// kubelet will restart the kube-proxy since it's running in a static pod
```
aojea, line 712: ok, this assumes that kube-proxy runs in a pod, cool

test/e2e/network/service.go
```diff
@@ -1102,9 +1070,6 @@ var _ = SIGDescribe("Services", func() {
 
 	ginkgo.It("should work after restarting kube-proxy [Disruptive]", func() {
 		// TODO: use the ServiceTestJig here
-		e2eskipper.SkipUnlessProviderIs("gce", "gke")
-		e2eskipper.SkipUnlessSSHKeyPresent()
```
spiffxp, line 1106: These checks are a poor proxy for asking "is kube-proxy running in this very specific manner and are we capable of restarting it?"

I might be in favor of abstracting / being more specific about what we're checking for e.g. `skipUnlessClusterComponentRunsAsPodsAndE2ETestClientIsAuthorizedToDeleteThem(component)` or `skipUnlessClusterProviderSupportsComponentRestart(component)` etc.

But I don't think I'm in favor of dropping the up-front check

spiffxp (changes requested):
I'm in favor of supporting more clusters, but I'm not in favor of dropping the capability check

test/e2e/network/service.go
JornShen: From the previous version,  the case alreay treat kube-proxy as pod. 

https://github.com/kubernetes/kubernetes/blob/fa15799f7409b8644221c2b43a494c74aac0c5ee/test/e2e/network/service.go#L712

if kube-proxy running as process, no one will restart it. And case can't run.Here are the same. 

Now if no find component pod in function `getComponentPod`, it wil return  error and the case will end with error.
Of cause,  in the front of case,  checking whether component is deploy as pod will be more nice. I will give it a try :)

JornShen:
@wojtek-t here are some problems need help :)

I find that the case(should work after restarting apiserver) treat apiserver as process before and use shell command to restart it if provider is AWS. 

The PR is merged in https://github.com/kubernetes/kubernetes/pull/11056/files.(seems a bit too long 😂 ). 
and more detail line in https://github.com/kubernetes/kubernetes/blob/fa15799f7409b8644221c2b43a494c74aac0c5ee/test/e2e/network/service.go#L3769

In AWS, kube-apiserver must be restart using `sudo /etc/init.d/kube-apiserver restart` ? Can we all treat apiserver as container(pod)? where or how  can I  confirm them?

test/e2e/network/service.go
JornShen: will turn the previous auth for help to confirm it. look below.

test/e2e/network/service.go
```diff
@@ -3733,84 +3687,64 @@ func validateEndpointsPorts(c clientset.Interface, namespace, serviceName string
...
-	v, err := cs.Discovery().ServerVersion()
+// restartComponent restarts component static pod
+func restartComponent(cs clientset.Interface, cName, ns string, labelSet labels.Set) error {
+	pods, err := getComponentPod(cs, ns, labelSet)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get %s's pods, err: %v", cName, err)
 	}
-	return framework.MasterUpgradeGKE(namespace, v.GitVersion[1:]) // strip leading 'v'
+	if err := killComponentPod(cs, ns, pods); err != nil {
```
wojtek-t, line 3696: This won't work for GKE (whereas the previous MasterUpgradeGKE works).

JornShen: yeah, we can not see kube-apiserver int gke, but we still can see kube-proxy pod.

wojtek-t: but this changes breaks things that works before this PR

JornShen: GKE this scene should be considered  in this PR.

JornShen: So if in AWS, can we use the approach of `killComponentPod ` to restart kube-apiserver. I find that case calculate kube-apiserver count by search as pod.

https://github.com/kubernetes/kubernetes/blob/ac10e4043af29c651407add4ead61975471f2fbd/test/e2e/network/service.go#L3730

https://github.com/kubernetes/kubernetes/blob/ac10e4043af29c651407add4ead61975471f2fbd/test/e2e/network/service.go#L3784-L3793

And if no pod, it will return err. And the case will failed.

wojtek-t: Let me clarify what I meant one more time.

Currently (without your PR), if run on GKE it would work, because we call MasterUpgradeGKE.

After your change - this will no longer work for GKE.

So this is a regression.
Please fix to not regress for the GKE usecase.

JornShen: All right, I fix it first.

JornShen: @wojtek-t ,thanks for reminding! In GKE  scene, it will do as before.

wojtek-t: Thanks!

test/e2e/network/service.go
BenTheElder: > if kube-proxy running as process, no one will restart it. And case can't run.Here are the same.

init(1) ?
I would think anyone running kube-proxy as a process has it under something like systemd.

BenTheElder:
In other places we've preformed actions on nodes by using a priv pod + exec + nsenter etc.
That approach ought to let us kill a process w/o relying on it being in a pod? Essentially you drop-in replace executing commands over ssh with using a special pod to execute them. The storage tests are an example.

AFAIK It's valid to run kube-proxy as a binary under e.g. systemd instead of as a pod.

test/e2e/network/service.go
JornShen: yeah, if kube-proxy run as process, systemd maybe manage it.

JornShen:
@spiffxp add `SkipUnlessComponentRunsAsPodsAndClientCanDeleteThem` up-front check to check the component run as pod.

JornShen:
> In other places we've preformed actions on nodes by using a priv pod + exec + nsenter etc.
> That approach ought to let us kill a process w/o relying on it being in a pod? Essentially you drop-in replace executing commands over ssh with using a special pod to execute them. The storage tests are an example.
> 
> AFAIK It's valid to run kube-proxy as a binary under e.g. systemd instead of as a pod.

which scene ?  I have see many k8s cluster, like cluster using kubeadm to install , kind, gke. They all run kube-proxy as pod.

test/e2e/network/service.go
```diff
@@ -3733,84 +3695,77 @@ func validateEndpointsPorts(c clientset.Interface, namespace, serviceName string
...
+			return fmt.Errorf("couldn't restart pod:%v, err: %v", pod.Name, err)
+		}
 	}
 	return nil
 }
 
-// waitForApiserverRestarted waits until apiserver's restart count increased.
-func waitForApiserverRestarted(c clientset.Interface, initialRestartCount int32) error {
+// waitForComponentRestarted waits until component's pod restart count increased.
+func waitForComponentRestarted(c clientset.Interface, cName, ns string, labelSet labels.Set, initialRestartCount int) error {
```
spiffxp, line 3741: I would get rid of this, I think it can be replaced with a call to e2epod (framework/pod) `PodsCreatedByLabel`

spiffxp (changes requested):
Some suggestions

test/e2e/network/service.go
```diff
@@ -3733,84 +3695,77 @@ func validateEndpointsPorts(c clientset.Interface, namespace, serviceName string
...
-	framework.Logf("Restarting master via ssh, running: %v", command)
-	result, err := e2essh.SSH(command, net.JoinHostPort(framework.APIAddress(), e2essh.SSHPort), framework.TestContext.Provider)
-	if err != nil || result.Code != 0 {
-		e2essh.LogResult(result)
-		return fmt.Errorf("couldn't restart apiserver: %v", err)
+	return waitForComponentRestarted(cs, cName, ns, labelSet, len(pods))
+}
+
+// killComponentPod will kill the pods
+func killComponentPod(cs clientset.Interface, ns string, pods []v1.Pod) error {
```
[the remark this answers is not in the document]
spiffxp: This could also be replaced with a call to `e2epod.DeletePodsWithGracePeriod` if you know what the default is for grace period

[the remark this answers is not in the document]
JornShen: I think we can use zero grace period, immediate delete it as before.

test/e2e/network/service.go
```diff
@@ -3733,84 +3695,77 @@ func validateEndpointsPorts(c clientset.Interface, namespace, serviceName string
...
+		framework.Logf("waiting for %v restart count to increase", cName)
 	}
-	return fmt.Errorf("timed out waiting for apiserver to be restarted")
+	return fmt.Errorf("timed out waiting for %v to be restarted", cName)
 }
 
-func getApiserverRestartCount(c clientset.Interface) (int32, error) {
-	label := labels.SelectorFromSet(labels.Set(map[string]string{"component": "kube-apiserver"}))
+// getComponentPod get all the satisfied pods
+func getComponentPod(c clientset.Interface, ns string, labelSet labels.Set) ([]v1.Pod, error) {
```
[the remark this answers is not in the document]
JornShen: Add a new func `GetPods` in framework/pods/resource.go.

test/e2e/network/service.go
```diff
@@ -3733,84 +3695,77 @@ func validateEndpointsPorts(c clientset.Interface, namespace, serviceName string
...
-		command = "sudo /etc/init.d/kube-apiserver restart"
+	if err := killComponentPod(cs, ns, pods); err != nil {
+		return fmt.Errorf("failed to restart component: %s, err: %v", cName, err)
 	}
-	framework.Logf("Restarting master via ssh, running: %v", command)
-	result, err := e2essh.SSH(command, net.JoinHostPort(framework.APIAddress(), e2essh.SSHPort), framework.TestContext.Provider)
-	if err != nil || result.Code != 0 {
-		e2essh.LogResult(result)
-		return fmt.Errorf("couldn't restart apiserver: %v", err)
+	return waitForComponentRestarted(cs, cName, ns, labelSet, len(pods))
```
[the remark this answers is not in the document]
JornShen: Thanks for guiding!

spiffxp:
flake: https://github.com/kubernetes/kubernetes/issues/95832
flake: https://github.com/kubernetes/kubernetes/issues/95920

spiffxp (approved):
/approve
/lgtm
Thank you for working through this

Issue #94900, opened 2020-09-18
Update test/integration to use the term "Control Plane"


spiffxp:
There are a number of tests that can probably updated to refer to a "control plane" or a "primary" instead of a "master".

Some of these updates can be done outright (variable names, comments), while others may require more refactoring.

It would probably be a good idea to start with the test/integration/framework package, and work outwards from there.
```
framework/master_utils.go
102:func (h *MasterHolder) SetMaster(m *controlplane.Master) {
127:func startMasterOrDie(masterConfig *controlplane.Config, incomingServer *httptest.Server, masterReceiver MasterReceiver) (*controlplane.Master, *httptest.Server, CloseFunc) {
252:func NewIntegrationTestMasterConfig() *controlplane.Config {
258:func NewIntegrationTestMasterConfigWithOptions(opts *MasterConfigOptions) *controlplane.Config {
285:func NewMasterConfig() *controlplane.Config {
290:func NewMasterConfigWithOptions(opts *MasterConfigOptions) *controlplane.Config {
332:// CloseFunc can be called to cleanup the master
336:func RunAMaster(masterConfig *controlplane.Config) (*controlplane.Master, *httptest.Server, CloseFunc) {
345:func RunAMasterUsingServer(masterConfig *controlplane.Config, s *httptest.Server, masterReceiver MasterReceiver) (*controlplane.Master, *httptest.Server, CloseFunc) {
```

Full list of files in test/integration that may be affected
```
$ ag -l master | sort
BUILD
apimachinery/watch_restart_test.go
apimachinery/watch_timeout_test.go
apiserver/admissionwebhook/admission_test.go
apiserver/admissionwebhook/client_auth_test.go
apiserver/admissionwebhook/load_balance_test.go
apiserver/admissionwebhook/reinvocation_test.go
apiserver/admissionwebhook/timeout_test.go
apiserver/apiserver_test.go
apiserver/apply/apply_test.go
apiserver/certreload/certreload_test.go
apiserver/flowcontrol/concurrency_test.go
auth/accessreview_test.go
auth/auth_test.go
auth/bootstraptoken_test.go
auth/node_test.go
auth/rbac_test.go
auth/svcaccttoken_test.go
certificates/admission_subjectrestriction_test.go
certificates/defaulting_test.go
certificates/field_selector_test.go
client/cert_rotation_test.go
configmap/configmap_test.go
cronjob/cronjob_test.go
daemonset/daemonset_test.go
[...]

PR #95071, opened 2020-09-25
Rename master to control plane in test/integration

spiffxp:
**What type of PR is this?**

/kind cleanup

Renames "master" to "control plane" in test/integration

Explicitly left untouched are occurrences that might be api, flag, or
package changes that are out of scope for test/integration, e.g.:
- `system:masters`
- `MasterCount` (pkg/controlplane.Instance)
- `--endpoint-reconciler-type master-count`

In most other cases, the replacements were of the form:
- master -> control plane (in comments/quotes not referencing vars)
- master -> controlplane (files/packages)
- master -> controlPlane (vars/funcs)
- Master -> ControlPlane (vars/funcs)

Part of https://github.com/kubernetes/kubernetes/issues/94900

Doesn't completely fix, but brings the list down to 
```
apiserver/admissionwebhook/admission_test.go
apiserver/admissionwebhook/client_auth_test.go
apiserver/admissionwebhook/load_balance_test.go
apiserver/admissionwebhook/reinvocation_test.go
apiserver/admissionwebhook/timeout_test.go
apiserver/certreload/certreload_test.go
auth/node_test.go
auth/rbac_test.go
certificates/admission_subjectrestriction_test.go
client/cert_rotation_test.go
controlplane/kube_apiserver_test.go
framework/controlplane_utils.go
ipamperf/README.md
```

I am open to suggestions if there are better names. I'll post the
specific replacements I used in a separate comment.

```release-note
NONE
```


spiffxp:
Script used.  Tries to care of specific cases before doing the more general cases, then fixes up exceptions afterwards.
```bash
#!/bin/bash

# func/var/comment renames
find . -type f -name '*.go' | while read -r f; do
  gsed -i -E \
    -e 's/Congfig/Config/g' \
    -e 's/SetMaster/SetInstance/g' \
    -e 's/Master metrics/Control plane metrics/g' \
    -e 's/Master/ControlPlane/g' \
    -e 's/master-token/controlplane-token/g' \
    -e 's/package master/package controlplane/g' \
    -e 's/masterCount/controlPlaneCount/g' \
    -e 's!(//.*)master!\1control plane!g' \
    -e 's!(".*)master!\1control plane!g' \
    -e 's/master/controlPlane/g' \
    -e 's/system:(not)?control planes/system:\1masters/g' \
    -e 's/ControlPlaneCount/MasterCount/g' \
    -e 's/control plane-count/master-count/g' \
    -e 's/control plane count/controlPlaneCount/g' \
    -e 's/control plane.G/controlPlane.G/g' \
    -e 's/((a)n( ))?API control plane/\2\3control plane/g' \
    "${f}"
done 

# file/package renames
renamed=false

if [[ -f ./framework/master_utils.go ]]; then
  git mv ./framework/{master,controlplane}_utils.go
  renamed=true
fi

if [[ -d ./master ]]; then
  git mv ./{master,controlplane}
  gsed -i \
    -e 's/package master/package controlplane/' \
    ./controlplane/*.go
  git mv ./controlplane/synthetic_{master,controlplane}_test.go
  renamed=true
fi

if ${renamed}; then
  pushd ../../
  ./hack/update-bazel.sh
  popd
fi
```

spiffxp:
/cc @liggitt @lavalamp 
if you have opinions or guidance here
/wg naming
/hold
for comment

lavalamp:
IMO "control plane" is the plural, "instance" is the singular. It will be very confusing to use the word "control plane" to refer to only one apiserver if there are multiple. Unfortunately that means that mass find-replacing is going to be hard to verify.

I was taking the other approach, the rename script I was using parses go files and finds references (but currently misses *_test.go references). I was going to do one symbol at a time.

spiffxp:
@laval
>IMO "control plane" is the plural, "instance" is the singular

That makes sense.  My concern is `instance` is pretty generic without some kind of qualifier.  e.g.
- `masterConfig` -> `instanceConfig`
- `NewIntegrationTestMasterConfig` -> `NewIntegrationTestInstanceConfig`

The best I can come up with given that opinion is to fully qualify, but it feels super verbose, e.g.
- `masterConfig` -> `controlPlaneInstanceConfig`
- `NewIntegrationTestMasterConfig` -> `NewIntegrationTestControlPlaneInstanceConfig`

Other things I considered
- `CPInstace/cpInstance`, but that could easily be confused with any of "cluster provider", "credentials provider", "cloud provider"
- `apiserver`, but there are tests in here that deal with apiservers separate from this, it seems like an instance of control plane is more than just an apiserver

WDYT?

lavalamp:
Yeah it's clear from the package for my cases. I agree it's verbose to fully specify. I don't have a better idea, though. "CP" is jargony but might be worth it.

This is probably why the existing term persists, honestly.

spiffxp:
/retest
flake (https://github.com/kubernetes/kubernetes/issues/93902)

neolit123 (commented):
looking at the tests / diff, "master" seems to be mostly a shorthand for a "kube-apiserver instance".
i think of "control-plane" as the umbrella, singular entity that includes all instances of the control-plane components.

spiffxp:
Ack.  I'll take another pass.  I'm coming at it from tests just needing a control plane to talk to, and they don't care how many instances it's composed of.  But it may be easier to just have everything talk to an instance.

spiffxp:
I will take a crack at this via smaller scope PRs, followup in linked issue

PR #95078, opened 2020-09-25
Disable one subpath test for windows

jingxu97:
Disable subpath test "should be able to unmount after the subpath
directory is deleted" for windows because the test will fail when
deleting a dir while another container still use it.

**What type of PR is this?**

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/test/e2e/storage/testsuites/subpath.go
+++ b/test/e2e/storage/testsuites/subpath.go
@@ -436,7 +436,9 @@ func (s *subPathTestSuite) DefineTests(driver TestDriver, pattern testpatterns.T
 		testPodFailSubpath(f, l.pod, true)
 	})
 
-	ginkgo.It("should be able to unmount after the subpath directory is deleted", func() {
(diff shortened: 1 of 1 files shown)


test/e2e/storage/testsuites/subpath.go
```diff
@@ -436,7 +436,7 @@ func (s *subPathTestSuite) DefineTests(driver TestDriver, pattern testpatterns.T
 		testPodFailSubpath(f, l.pod, true)
 	})
 
-	ginkgo.It("should be able to unmount after the subpath directory is deleted", func() {
+	ginkgo.It("should be able to unmount after the subpath directory is deleted [LinuxOnly]", func() {
```
msau42, line 439: Can you add a comment explaining why this doesn't work on Windows

jingxu97: added. PTAL

PR #95080, opened 2020-09-26
Add scheduler metrics to track pod with priority

### LINKED ISSUE #94659: Add a metric tracking pod priorities used in a cluster
A histogram that tracks the priorities used in a cluster.

**Why is this needed**:
Helps platform operators track feature usage. Priorities in particular are important since they trigger preemption in the scheduler, and so understanding how they are used helps inform optimizations related to preemption.

/sig scheduling

arghya88:
**What type of PR is this?**
/kind feature
/sig scheduling

Fixes #94659

```release-note
Add histogram metric `PodPriority` to track pod priorities used in a cluster
```

```docs

```

### DIFF
--- a/pkg/scheduler/metrics/metrics.go
+++ b/pkg/scheduler/metrics/metrics.go
@@ -187,6 +187,15 @@ var (
 			StabilityLevel: metrics.ALPHA,
 		}, []string{"type"})
 
+	PodPriority = metrics.NewHistogram(
+		&metrics.HistogramOpts{
+			Subsystem:      SchedulerSubsystem,
+			Name:           "pod_priority",
+			Help:           "Pod priority number",
+			Buckets:        metrics.ExponentialBuckets(1, 2, 15),
+			StabilityLevel: metrics.ALPHA,
+		})
+
 	metricsList = []metrics.Registerable{
 		scheduleAttempts,
 		e2eSchedulingLatency,
@@ -204,6 +213,7 @@ var (
 		SchedulerGoroutines,
 		PermitWaitDuration,
 		CacheSize,
+		PodPriority,
 	}
 )
 
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -419,6 +419,8 @@ func (sched *Scheduler) finishBinding(prof *profile.Profile, assumed *v1.Pod, ta
 // scheduleOne does the entire scheduling workflow for a single pod.  It is serialized on the scheduling algorithm's host fitting.
 func (sched *Scheduler) scheduleOne(ctx context.Context) {
 	podInfo := sched.NextPod()
+	podPriority := podutil.GetPodPriority(podInfo.Pod)
+	metrics.PodPriority.Observe(float64(podPriority))
 	// pod could be nil when schedulerQueue is closed
 	if podInfo == nil || podInfo.Pod == nil {
 		return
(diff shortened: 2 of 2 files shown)


pkg/scheduler/metrics/metrics.go
```diff
@@ -187,6 +187,15 @@ var (
 			StabilityLevel: metrics.ALPHA,
 		}, []string{"type"})
 
+	PodPriority = metrics.NewHistogram(
+		&metrics.HistogramOpts{
+			Subsystem:      SchedulerSubsystem,
+			Name:           "pod_priority",
```
adtac, line 193: while I don't have any strong opinions on whether this should be a new metric on its own, perhaps we could explore the possibility of adding pod priority as a label to existing metrics such as `pod_scheduling_attempts` or `pod_scheduling_duration_seconds`. However, one concern will be the high cardinality of pod priorities.

pkg/scheduler/metrics/metrics.go
```diff
@@ -187,6 +187,15 @@ var (
...
+			Help:           "Pod priority number",
+			Buckets:        metrics.ExponentialBuckets(1, 2, 15),
```
adtac, line 195: this would create buckets [1, 2), [2, 4), [4, 8)..., [16k, inf), right? In practice, I've seen pod priorities to be usually in the 1000s, which wouldn't give this much granularity (e.g. priority 5000 and 6000 cannot be differentiated). Also, would this capture pod priority 0 (default value)?

pkg/scheduler/scheduler.go
```diff
@@ -419,6 +419,8 @@ func (sched *Scheduler) finishBinding(prof *profile.Profile, assumed *v1.Pod, ta
 // scheduleOne does the entire scheduling workflow for a single pod.  It is serialized on the scheduling algorithm's host fitting.
 func (sched *Scheduler) scheduleOne(ctx context.Context) {
 	podInfo := sched.NextPod()
+	podPriority := podutil.GetPodPriority(podInfo.Pod)
+	metrics.PodPriority.Observe(float64(podPriority))
```
adtac, line 423: do we want to record the priority for the same pod if it goes through multiple attempts in multiple scheduling cycles? that seems counter-intuitive.

pkg/scheduler/scheduler.go
```diff
@@ -419,6 +419,8 @@ func (sched *Scheduler) finishBinding(prof *profile.Profile, assumed *v1.Pod, ta
 // scheduleOne does the entire scheduling workflow for a single pod.  It is serialized on the scheduling algorithm's host fitting.
 func (sched *Scheduler) scheduleOne(ctx context.Context) {
 	podInfo := sched.NextPod()
+	podPriority := podutil.GetPodPriority(podInfo.Pod)
```
adtac, line 422: this should come after the klog.V(3) statement that announces that we're attempting to schedule this pod

pkg/scheduler/metrics/metrics.go
arghya88: Thinking a bit more on this I think a counter metrics to keep count of how many pods have priority>0 maybe more useful..what's your thoughts @ahg-g ?

ahg-g:
we may get this using the new pod resources metric: https://github.com/kubernetes/enhancements/pull/1916#discussion_r501147210

pkg/scheduler/metrics/metrics.go
brancz: I fear that adding pod priority to existing metrics would make the cardinality explode. I think having this separate metric makes sense.

pkg/scheduler/metrics/metrics.go
brancz: Yes, I think in this case defining buckets manually is probably most reasonable. Exponential buckets are really meant for buckets of request latencies of a network service.

ahg-g:
We agreed to add priority to the pod resources metric https://github.com/kubernetes/enhancements/pull/1916#discussion_r501147210 so this PR is not necessary anymore. Thanks for looking into it.

/close

PR #95081, opened 2020-09-26
Log more url metadata in requestinfo.go

### LINKED ISSUE #95073: Log more url metadata in requestinfo.go when failures happen

requestinfo.go should log the url , as well as the parameters that are failing in a request:

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go#L214

However only logs

```
 req.URL.Query()
```

**Why is this needed**:

otherwise, its not easy to know what component is failing due to label selectors,   

```
E0925 19:23:52.079668       1 requestinfo.go:214] Couldn't parse request url.Values{"labelSelector":[]string{"name=kube-controller-manager-ip-10-0-4-229.us-west-2.compute.internal"}}: unable to parse requirement: invalid label value: "kube-controller-manager-ip-10-0-4-229.us-west-2.compute.internal": at key: "name": must be no more than 63 characters
```

wawa0210:
**What type of PR is this?**

/kind bug
/kind feature

Fixes #95073

```release-note
Log more url metadata in requestinfo.go when failures happen
```

```docs

```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go
+++ b/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go
@@ -211,7 +211,7 @@ func (r *RequestInfoFactory) NewRequestInfo(req *http.Request) (*RequestInfo, er
 		opts := metainternalversion.ListOptions{}
 		if err := metainternalversionscheme.ParameterCodec.DecodeParameters(req.URL.Query(), metav1.SchemeGroupVersion, &opts); err != nil {
 			// An error in parsing request will result in default to "list" and not setting "name" field.
-			klog.Errorf("Couldn't parse request %#v: %v", req.URL.Query(), err)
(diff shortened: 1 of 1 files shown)


lavalamp:
Sorry for delay.

Logging isn't the right way to expose errors like this. They need to be returned to the user and the request probably should be failed.

wawa0210:
> Sorry for delay.
> 
> Logging isn't the right way to expose errors like this. They need to be returned to the user and the request probably should be failed.

You are right, the code just log error not return the error https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go#L214

There is an error  judgment at the call of the function, should we return error directly instead of continuing to process it?

lavalamp:
> we return error directly instead of continuing to process it?

It's not 100% clear how to do that in this function, but more or less,
yeah, this should probably abort the request, especially if it could lose
things like dry run markers. I'm swamped today so won't be able to help
with the mechanics, sorry.

PR #62487, opened 2018-04-12
wire print flags through apply cmd

juanvallejo:
**Release note**:
```release-note
NONE
```

Depends on https://github.com/kubernetes/kubernetes/pull/62300
Adds PrintFlags to `apply` command.

cc @soltysh @deads2k

PR #95082, opened 2020-09-26
Do Not Show Deletion Flags Under kubectl apply Options

### LINKED ISSUE #928: Do Not Show Deletion Flags Under kubectl apply Options
I was recently looking into if `kubectl apply` offered a similar shorthand as `kubectl delete` does for `--wait` to avoid needing to run a `kubectl wait` command after to verify all pods become available in a namespace after an apply is run.

Looking further at the documentation for `kubectl apply -h`, it displays options like `--wait` and `--timeout`, and the descriptions seem to be in the context of `kubectl delete`:

```
--timeout=0s: The length of time to wait before giving up on a delete, zero means determine a timeout from the
size of the object
...
--wait=false: If true, wait for resources to be gone before returning. This waits for finalizers.
```

Not sure what the reasoning is for displaying these options with apply since they aren't supported options, but was wondering if there are any plans in place to either support these options with apply or to remove the non-applicable options. 

**What happened**:

Delete options like `--timeout` and `--wait` are displayed for `kubectl apply -h`:

```
--timeout=0s: The length of time to wait before giving up on a delete, zero means determine a timeout from the
size of the object
...
--wait=false: If true, wait for resources to be gone before returning. This waits for finalizers.
```

These options should not be displayed with `kubectl apply`.

Run `kubectl apply -h`

**Environment**:

- Kubernetes client and server versions (use `kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T23:35:15Z", GoVersion:"go1.14.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:50:46Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
```

N/A

N/A

lianghao208:
**What type of PR is this?**
/kind bug

Do Not Show Deletion Flags Under kubectl apply Options

Fixes [https://github.com/kubernetes/kubectl/issues/928](https://github.com/kubernetes/kubectl/issues/928)

just remove kubectl Deletion Flags when using apply Options.

```release-note
NONE
```

```docs
None
```

### DIFF
--- a/staging/src/k8s.io/kubectl/pkg/cmd/apply/apply.go
+++ b/staging/src/k8s.io/kubectl/pkg/cmd/apply/apply.go
@@ -180,7 +180,6 @@ func NewCmdApply(baseName string, f cmdutil.Factory, ioStreams genericclioptions
 	}
 
 	// bind flag structs
-	o.DeleteFlags.AddFlags(cmd)
 	o.RecordFlags.AddFlags(cmd)
 	o.PrintFlags.AddFlags(cmd)
 
--- a/hack/.golint_failures
+++ b/hack/.golint_failures
@@ -67,7 +67,7 @@ pkg/controller/job/config/v1alpha1
 pkg/controller/namespace
 pkg/controller/namespace/config/v1alpha1
 pkg/controller/nodeipam/config/v1alpha1
-pkg/controller/nodelifecycle/config/v1alpha1
+pkg/controller/nodelifecycle/config/v1alpha1 # only 'don't use underscores in Go names' due to auto generated functions
 pkg/controller/podautoscaler/config/v1alpha1
 pkg/controller/podautoscaler/metrics
 pkg/controller/podgc


lianghao208:
I signed it

staging/src/k8s.io/kubectl/pkg/cmd/apply/apply.go
```diff
@@ -180,7 +180,6 @@ func NewCmdApply(baseName string, f cmdutil.Factory, ioStreams genericclioptions
 	}
 
 	// bind flag structs
-	o.DeleteFlags.AddFlags(cmd)
```
eddiezane, line 183: The fix here is a bit more complicated than this because the `DeleteFlags` are where `-f` and `-k` come from.

https://github.com/kubernetes/kubernetes/blob/c8ebc8ab75a9c36453cf6fa30990fd0a277d856d/staging/src/k8s.io/kubectl/pkg/cmd/delete/delete_flags.go#L35

https://github.com/kubernetes/kubernetes/blob/c8ebc8ab75a9c36453cf6fa30990fd0a277d856d/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/filename_flags.go#L34-L36

eddiezane (changes requested):
Not entirely sure the correct path forward on this one. We may want to implement the missing options if they make sense otherwise we'll most likely need a refactor to separate the flags.

staging/src/k8s.io/kubectl/pkg/cmd/apply/apply.go
lianghao208: I agree, it needs a refactor to separate the flags.

zhouya0 (commented):
Same with @eddiezane.
And by the way, actually `--timeout`, `--wait` flag seems work with `apply`. You could find the explaination for `--timeout` is:
```
cmd.Flags().DurationVar(&o.Timeout, "timeout", o.Timeout, "Only relevant during a force apply. The length of time to wait before giving up on a delete of the old resource, zero means determine a timeout from the size of the object. Any other values should contain a corresponding time unit (e.g. 1s, 2m, 3h).")
```
Ref: https://github.com/kubernetes/kubernetes/pull/62487

danielhelfand:
> And by the way, actually `--timeout`, `--wait` flag seems work with `apply`. You could find the explaination for `--timeout` is:
> 
> ```
> cmd.Flags().DurationVar(&o.Timeout, "timeout", o.Timeout, "Only relevant during a force apply. The length of time to wait before giving up on a delete of the old resource, zero means determine a timeout from the size of the object. Any other values should contain a corresponding time unit (e.g. 1s, 2m, 3h).")
> ```
> 

Seems the pull request referenced removed a more apt description for the options in the context of apply.

staging/src/k8s.io/kubectl/pkg/cmd/apply/apply.go
soltysh: Additionally, since we've added these flags here (even if by mistake) we need to ensure we don't break people suddenly.  So you'd need to copy these flags directly here, deprecate them, which will also hide them. This way ppl won't be broken when using these flags, especially in scripts, but will get clear information about its deprecation and that they don't do anything.

soltysh (changes requested):

PR #95090, opened 2020-09-27
chore: use errors.Is in PluginArgs validation tests

### LINKED ISSUE #95047: Use errors.Is in PluginArgs validation tests
/sig scheduling
/kind cleanup

Files:

- https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/apis/config/validation/validation_test.go
- https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/apis/config/validation/validation_pluginargs_test.go

We currently compare the error strings instead of leveraging `errors.Is`, which will ensure the correct types are used.

krol3:
**What type of PR is this?**
/kind cleanup

Use errors.Is in PluginArgs validation tests

Fixes https://github.com/kubernetes/kubernetes/issues/95047

```
NONE
```

```docs
None
```


hwdef:
/release-note-none
/lgtm

thanks for your contribution!

pkg/scheduler/apis/config/validation/validation_pluginargs_test.go
```diff
@@ -483,7 +484,7 @@ func assertErr(t *testing.T, wantErr string, gotErr error) {
 		if gotErr == nil {
 			t.Fatalf("\nwant err to be:\n\t%s\ngot:\n\tnil", wantErr)
 		}
-		if gotErr.Error() != wantErr {
+		if !errors.Is(gotErr, errors.New(wantErr)) {
```
pancernik, line 487: @krol3 the tests are failing, because the errors returned by the validation functions (`k8s.io/apimachinery/pkg/util/validation/field`) do not match (according to `errors.Is` spec) ones created here with `errors.New`.

@alculquicondor I'm not sure if we can leverage `errors.Is` here to check the type of the error, because `field.Error` doesn't implement  `interface{ Is(error) bool }`.  In this scenario `errors.Is` uses equality operator to compare pointers to `field.Error` structs. They are different as they are created dynamically and ultimately point to different variables.

changshuchao: How about this : https://github.com/kubernetes/kubernetes/pull/95092

alculquicondor: > I'm not sure if we can leverage errors.Is here to check the type of the error, because field.Error doesn't implement interface{ Is(error) bool }

Then we can implement it :) I'm sure api-machinery folks will appreciate.

> How about this : #95092

I don't think so

staging/src/k8s.io/apimachinery/pkg/util/validation/field/errors.go
```diff
@@ -270,3 +270,8 @@ func (list ErrorList) Filter(fns ...utilerrors.Matcher) ErrorList {
 	// FilterOut takes an Aggregate and returns an Aggregate
 	return fromAggregate(err.(utilerrors.Aggregate))
 }
+
+// Implementing Is error method
+func (m Error) Is(target error) bool {
+	return target.Error() == m.Error()
```
pancernik, line 276: This should check the type of the error and compare the actual fields in `target`. Please have a look at https://blog.golang.org/go1.13-errors. It has an excellent example of Is implementation.

Please implement it on the pointer receiver similarly to other methods for this type.

The comment should also reflect what the method is doing using Godoc convention: https://blog.golang.org/godoc

alculquicondor:
Please rebase and squash

PR #95091, opened 2020-09-27
wrap errors in selectorspread and podtoplogyspread plugin

arghya88:
**What type of PR is this?**
/kind cleanup

Wrap errors in selectorspread and podtoplogyspread plugin

Ref #94695

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/scheduler/framework/plugins/podtopologyspread/filtering_test.go
+++ b/pkg/scheduler/framework/plugins/podtopologyspread/filtering_test.go
@@ -18,6 +18,7 @@ package podtopologyspread
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"testing"
 
@@ -1651,7 +1652,7 @@ func TestPreFilterDisabled(t *testing.T) {
 	p := &PodTopologySpread{}
 	cycleState := framework.NewCycleState()
 	gotStatus := p.Filter(context.Background(), cycleState, pod, nodeInfo)
-	wantStatus := framework.NewStatus(framework.Error, `error reading "PreFilterPodTopologySpread" from cycleState: not found`)
+	wantStatus := framework.AsStatus(fmt.Errorf(`reading "PreFilterPodTopologySpread" from cycleState: not found`))
 	if !reflect.DeepEqual(gotStatus, wantStatus) {
 		t.Errorf("status does not match: %v, want: %v", gotStatus, wantStatus)
 	}
--- a/pkg/scheduler/framework/plugins/podtopologyspread/scoring.go
+++ b/pkg/scheduler/framework/plugins/podtopologyspread/scoring.go
@@ -113,7 +113,7 @@ func (pl *PodTopologySpread) PreScore(
 ) *framework.Status {
 	allNodes, err := pl.sharedLister.NodeInfos().List()
 	if err != nil {
-		return framework.NewStatus(framework.Error, fmt.Sprintf("error when getting all nodes: %v", err))
+		return framework.AsStatus(fmt.Errorf("getting all nodes: %w", err))
 	}
 
 	if len(filteredNodes) == 0 || len(allNodes) == 0 {
@@ -127,7 +127,7 @@ func (pl *PodTopologySpread) PreScore(
 	}
 	err = pl.initPreScoreState(state, pod, filteredNodes)
 	if err != nil {
-		return framework.NewStatus(framework.Error, fmt.Sprintf("error when calculating preScoreState: %v", err))
+		return framework.AsStatus(fmt.Errorf("calculating preScoreState: %w", err))
 	}
 
 	// return if incoming pod doesn't have soft topology spread Constraints.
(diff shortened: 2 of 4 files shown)


pkg/scheduler/framework/plugins/podtopologyspread/filtering_test.go
```diff
@@ -1651,7 +1652,7 @@ func TestPreFilterDisabled(t *testing.T) {
 	p := &PodTopologySpread{}
 	cycleState := framework.NewCycleState()
 	gotStatus := p.Filter(context.Background(), cycleState, pod, nodeInfo)
-	wantStatus := framework.NewStatus(framework.Error, `error reading "PreFilterPodTopologySpread" from cycleState: not found`)
+	wantStatus := framework.AsStatus(fmt.Errorf(`error reading "PreFilterPodTopologySpread" from cycleState: %w`, fmt.Errorf("not found")))
```
chendave, line 1655: what's the difference with this? anything more bring us?
```golang
framework.AsStatus(fmt.Errorf(`error reading "PreFilterPodTopologySpread" from cycleState: not found`))
```

arghya88: I was following it up from another PR..but I agree and will revert this back

pkg/scheduler/framework/plugins/podtopologyspread/filtering_test.go
```diff
@@ -1651,7 +1652,7 @@ func TestPreFilterDisabled(t *testing.T) {
 	p := &PodTopologySpread{}
 	cycleState := framework.NewCycleState()
 	gotStatus := p.Filter(context.Background(), cycleState, pod, nodeInfo)
-	wantStatus := framework.NewStatus(framework.Error, `error reading "PreFilterPodTopologySpread" from cycleState: not found`)
+	wantStatus := framework.AsStatus(fmt.Errorf(`error reading "PreFilterPodTopologySpread" from cycleState: not found`))
```
alculquicondor, line 1655: remove the word "error" from the message. Same for the rest of the PR.

pkg/scheduler/framework/plugins/podtopologyspread/scoring.go
```diff
@@ -174,13 +174,13 @@ func (pl *PodTopologySpread) PreScore(
 func (pl *PodTopologySpread) Score(ctx context.Context, cycleState *framework.CycleState, pod *v1.Pod, nodeName string) (int64, *framework.Status) {
 	nodeInfo, err := pl.sharedLister.NodeInfos().Get(nodeName)
 	if err != nil || nodeInfo.Node() == nil {
-		return 0, framework.NewStatus(framework.Error, fmt.Sprintf("getting node %q from Snapshot: %v, node is nil: %v", nodeName, err, nodeInfo.Node() == nil))
+		return 0, framework.AsStatus(fmt.Errorf("getting node %q from Snapshot: %w, node is nil: %v", nodeName, err, nodeInfo.Node() == nil))
```
alculquicondor, line 177: `%w` should be at the end of the message.

IIRC, `nodeInfo.NodeI()` can no longer be nil, as we do the check when we create the Snapshot. Would you mind verifying and removing that check?

pkg/scheduler/framework/plugins/selectorspread/selector_spread.go
```diff
@@ -86,17 +86,17 @@ func (pl *SelectorSpread) Score(ctx context.Context, state *framework.CycleState
...
 	c, err := state.Read(preScoreStateKey)
 	if err != nil {
-		return 0, framework.NewStatus(framework.Error, fmt.Sprintf("Error reading %q from cycleState: %v", preScoreStateKey, err))
+		return 0, framework.AsStatus(fmt.Errorf("Error reading %q from cycleState: %w", preScoreStateKey, err))
 	}
 
 	s, ok := c.(*preScoreState)
 	if !ok {
-		return 0, framework.NewStatus(framework.Error, fmt.Sprintf("%+v convert to tainttoleration.preScoreState error", c))
+		return 0, framework.AsStatus(fmt.Errorf("%+v convert to tainttoleration.preScoreState error", c))
```
alculquicondor, line 94: we shouldn't reach this line, but "%+v" is very unreadable. Just remove it.

"cannot convert saved state to tainttoleration.preScoreState"

pkg/scheduler/framework/plugins/podtopologyspread/scoring.go
arghya88: There is a check here
 https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/podtopologyspread/scoring.go#L142

pkg/scheduler/framework/plugins/podtopologyspread/filtering_test.go
arghya88: Done

pkg/scheduler/framework/plugins/selectorspread/selector_spread.go
arghya88: Done

pkg/scheduler/framework/plugins/podtopologyspread/scoring.go
arghya88: Done

pkg/scheduler/framework/plugins/podtopologyspread/filtering_test.go
arghya88: Done

pkg/scheduler/framework/plugins/podtopologyspread/scoring.go
alculquicondor: /hold

The source of truth is actually the scheduler cache https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/internal/cache/cache.go#L201

And it looks like there might actually be a bug there were we consider some Nodes that are nil. I'm evaluating.

arghya88:
remove do-not-merge/hold

alculquicondor:
/hold

Hold was intended

alculquicondor:
/hold cancel
because #95130 is merged

